Issue #8871, opened 2026-07-15
Remote mirror pty-exit can kill live host terminal sessions


bbingz:
## Summary

A paired desktop client can terminate a live terminal on the host when its local remote-terminal mirror reports `pty-exit` (for example while retiring a stale remote PTY handle during reconnect/restore).

This destroys long-running terminal workloads such as Claude Code sessions. The affected workload does not need to be a native provider integration; the observed sessions were Claude Code processes configured to use a Grok API endpoint.

## Observed behavior

- Agent sleep was disabled on the host.
- A paired client's renderer bootstrapped at `18:36:08.431` local time.
- The host daemon logged an explicit `session-killed` for an existing terminal at `18:36:11.779`, only 3.348 seconds later.
- Additional mirrored terminal sessions were explicitly killed during the same paired-client activity window.
- These were `session-killed` events, not natural `session-exited` events.

## Root cause

The renderer exit path is:

1. The remote PTY transport retires/disconnects a mirrored PTY and invokes `onPtyExit`.
2. `Terminal.tsx` calls `closeTerminalTab(tabId, { reason: 'pty-exit' })`.
3. For a remote-owned worktree, `closeTerminalTab` prunes the local mirror **and unconditionally calls** `closeWebRuntimeSessionTab`.
4. The host receives `session.tabs.close` and kills the authoritative PTY.

A mirror-local `pty-exit` is reconciliation evidence, not user intent to close the host-owned tab. Propagating it back to the host turns stale mirror state into a destructive host action.

Relevant code:

- `src/renderer/src/components/terminal/terminal-tab-actions.ts`
- `src/renderer/src/components/terminal/Terminal.tsx`
- `src/renderer/src/runtime/remote-runtime-pty-transport.ts`

This appears related to the remote close propagation added around #8628: explicit user/cleanup closes should reach the host, but natural/mirror `pty-exit` must not.

## Expected behavior

When a remote mirror reports `pty-exit`, Orca should reconcile/prune only the local mirror and wait for the authoritative host snapshot. It must not send `session.tabs.close` to the host.

Explicit user close and cleanup actions should continue to propagate to the host.

## Regression test

A focused unit test currently fails because `closeWebRuntimeSessionTab` is called once for:

```ts
closeTerminalTab('local-tab-1', { reason: 'pty-exit' })
```

The fix should make that call local-only while preserving the existing explicit-close behavior.

bbingz:
[...]

PR #8887, opened 2026-07-15
fix(agent-resume): skip sleeping-session resume for runtime-owned worktrees

### LINKED ISSUE #8878: [Bug]: Remote client resumes live host agent sessions, spawning duplicate TUIs on the same provider session
**Operating system:** macOS
**Orca version:** v1.4.141 (root-cause references verified against `main` @ 190ed7d4e)

### Short summary

A paired remote client (web client or a second Orca desktop) that activates a worktree can launch `codex resume <id>` / `claude --resume` for an agent session that is **still alive and daemon-backed on the host**. The host ends up with two concurrent TUIs attached to the same provider session. This also explains historical observations of the same session id being resumed 2–3× concurrently.

### What happened

A remote desktop client reconnected to a host that had live daemon-backed agent sessions. At activation time the client had not yet received the host's session-tab mirror, so the live sessions were invisible to it; the client then launched resume commands for sessions that never stopped running. The same effect occurs when the user, seeing no live session in the client UI, manually starts a "resume" — but the automatic path fires on its own.

### How can we reproduce it

1. On the host, run a daemon-backed agent session (codex/claude TUI) in a worktree and leave it running.
2. From a paired client on another machine (cold start, so its cached state is hydrated before the session-tabs mirror arrives), connect and activate that worktree.
3. Observe on the host: a second PTY running `codex resume <same-session-id>` is spawned next to the still-live original.

### Root cause (code path)

1. **Resume runs client-side with no runtime-ownership gate.** `resumeSleepingAgentSessionsForWorktree` is called unconditionally on worktree activation (`src/renderer/src/lib/worktree-activation.ts:363`, folder-workspace at `:229`) and on startup hydration (`src/renderer/src/components/Terminal.tsx:1254-1266`).
2. **Every guard reads client-local state only.** `recordPaneIsOwnedByPreservedPane` (`src/renderer/src/lib/sleeping-agent-pane-ownership.ts:103-149`) and `activeOrQueuedResumeClaimsProviderSession` (`src/renderer/src/lib/resume-sleeping-agent-session.ts:89-129`) consult `tabsByWorktree` / `agentStatusByPaneKey` / renderer-local claim maps. Host tabs reach the client only via the **async** session-tabs mirror (`src/renderer/src/runtime/web-session-tabs-sync.ts`), and nothing orders the resume pass after the first snapshot — so a live host session is invisible to all guards at activation time. (The claim maps are ephemeral renderer state that is never synced, so under *any* timing a host-live session cannot appear in them.)
3. **Clients manufacture resume records for host sessions they merely viewed.** Mirrored agent status carries `providerSession` (`src/main/runtime/orca-runtime.ts:21282-21321` → `web-session-tabs-sync.ts:622-737`), and the periodic/beforeunload capture (`App.tsx:1362, 1384-1392` → `src/renderer/src/store/slices/agent-status.ts:2527-2565`) stamps `origin:'quit'`/`'live'` records with no runtime-ownership guard. The records persist in the client's local `runtime:<env>` partition and rehydrate on the next boot (`workspace-session-host-persistence.ts:260-296`, `terminals.ts:3182-3186`).
4. **The duplicate spawns on the host, with zero dedup.** The resume tab's id is a plain UUID, so the transport takes the `terminal.create` branch rather than the mirror-attach branch (`src/renderer/src/components/terminal-pane/remote-runtime-pty-transport.ts:630-658`). Host `createTerminal` spawns unconditionally (`orca-runtime.ts:17984-18093`); the daemon dedupes only by its own `sessionId` (`src/main/daemon/terminal-host.ts:61-91`). No layer — renderer, runtime, or daemon — keys "this provider session is live" across renderers.
5. **The codebase already gates plain terminal auto-create on host authority** (`Terminal.tsx:1232-1236`, `terminal-tab-actions.ts:59-71`, `worktree-activation.ts:425-430`); the resume path predates/misses that gate.

### Interaction with #8871

`reason:'pty-exit'` closes never retire `sleepingAgentSessionsByPaneKey` records, so every mirror-echo kill (#8871) converts a live session into a pending resume that fires on the next connect — the two bugs form a feedback pair.

### Proposed fix

- **Minimal (PR incoming):** skip the resume pass for runtime-owned worktrees, using the same predicate the auto-create gate already uses (`isWebRuntimeSessionActive(getRuntimeEnvironmentIdForWorktree(...))`, `Terminal.tsx:1234`). The host renderer already owns wake authority (`orca-runtime.ts:14503-14511`), so client-side resume for host-owned worktrees is never correct.
- **Follow-up 1:** host-side authoritative dedup — plumb a structured `resumeProviderSession` through `terminal.create` (the renderer already holds it in `pendingStartupByTabId`; it is dropped at the transport boundary) and reattach/deny when a live pane already carries that provider session.
- **Follow-up 2:** retire resume records on tab close (relates to the known tab-close/resume-list leak analysis from 2026-07-13).

bbingz:
## Summary

Fixes #8878. A paired remote client (web client or a second Orca desktop) could launch `codex resume <id>` / `claude --resume` for agent sessions that are **still alive and daemon-backed on the host**, spawning a second TUI on the same provider session: the client-side resume pass runs before the async session-tabs mirror arrives and every guard reads client-local state, so a live host session is invisible to it.

`resumeSleepingAgentSessionsForWorktree` now short-circuits when the target worktree is runtime-owned (host-authoritative) — the same host-authority predicate the plain-terminal auto-create gate already uses in `Terminal.tsx` (`isWebRuntimeSessionActive(getRuntimeEnvironmentIdForWorktree(...))`), applied at the single choke point that covers all four resume call sites. The host renderer keeps wake authority (its worktrees are not runtime-owned there, so the gate stays false on the session-owning machine); local, SSH, and WSL worktrees are unaffected.

Deliberate follow-ups tracked in #8878 (out of scope here): host-side provider-session dedup in `terminal.create`, and retiring resume records on tab close. Interaction with #8871: mirror-echo kills currently convert live sessions into pending duplicate resumes — this PR stops the duplicate-resume half on clients.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — repo-wide `oxlint` clean for this change (remaining warnings are pre-existing in unrelated files) + `check:max-lines-ratchet` OK ("no new bypasses"); the full `pnpm lint` script chain was killed mid-run (OOM) on the dev machine — deferring the remaining verify:* steps to CI
- [x] `pnpm typecheck` — node, cli, and web tsconfigs all pass
- [x] `pnpm test` — focused suites for every touched path: 10 files / 106 tests green (`resume-sleeping-agent-session*` incl. the new runtime-owner matrix, `wake-sleeping-agents-in-background`, all 5 `worktree-activation` suites); full suite deferred to CI
- [ ] `pnpm build` — not run locally (heavy native build); deferring to CI

**Failing-first repro:** with the gate removed, the runtime-owned matrix cases fail exactly as in the bug (`expected 1 to be +0` — the client resumed a live host session); with the gate restored, 6/6 pass. The new table-driven matrix keeps ≥2 worktrees in store simultaneously with a *different* runtime environment globally focused in every case, covering both explicit-owner resolution paths (`worktree.hostId` and `repo.executionHostId`) and asserting only the requested worktree is gated: explicit `runtime:*` worktrees skip; explicit `local`, `ssh:*`, and WSL-local worktrees still resume.

## AI Review Report

Adversarial review by Codex (`gpt-5.6-sol`, read-only sandbox) plus an independent spot-check of its cited evidence. It traced the gate through `worktree-runtime-owner.ts` (explicit per-worktree/per-repo owner resolution, then global fallback) and `web-runtime-session.ts`, and confirmed: all four production call sites route through the single choke point and none consumes the numeric return in a breaking way; the pane-owned cold-restore branch for legitimate local/SSH/WSL sessions is unaffected; topology semantics hold on remote clients (gated), the session-owning host (not gated), SSH (not runtime-owned), and WSL (local execution host — only the resume command differs); headless serve has no renderer, so the host-serving case is structurally exempt. Cross-platform: no platform-specific APIs touched. The review's sole material finding — the original tests exercised only the global-fallback branch of the ownership predicate — was addressed by rewriting the tests into the explicit-owner matrix described above.

## Security Audit

Renderer-only predicate/state change. No input handling, IPC surface, command execution, path handling, auth, secrets, or dependency changes; the diff only short-circuits a client-local resume decision before any store mutation or launch. Codex review reached the same conclusion (residual risk classed as correctness, not security, and covered by the new test matrix).

## Notes

- The gate reads only worktree/repo host ownership; WSL resume-command construction is untouched.
- No live two-machine end-to-end run was performed for this PR; correctness rests on the failing-first repro matrix plus the verified equivalence with the existing auto-create gate predicate.


bbingz:
Added a follow-up test commit (d5869424b) from a second independent review pass: pins that the runtime-owned gate is non-destructive — the sleeping record stays present, unconsumed (same reference) and unmutated (deep-equal vs pre-call snapshot) when resume is skipped, and remains resumable if the worktree's ownership later legitimately flips to local (record is then consumed exactly once). No expiry/retirement semantics are asserted — that policy stays with the deferred follow-up tracked in #8878. Runtime-owner suite now 7/7 green.

nwparker:
Taking over this approach on current `main` as #9098.

**Credit:** your choke-point gate + original runtime-owner matrix (cherry-picked as `821e05e946` / `6ea8a32eea`). Added adversarial cases (folder workspaces, active-runtime global fallback, re-entrant activation/startup races) and the handoff repro flipped to a positive regression guard.

Please review #9098 — happy to co-author credit; core design is yours.

bbingz:
Closing in favor of #9098 following the explicit maintainer takeover in this thread. #9098 cherry-picked the core runtime-owner gate and original matrix with credit, then added adversarial folder-workspace, global-fallback, and re-entrant startup coverage. Issue #8878 remains tracked by the successor PR.

PR #8904, opened 2026-07-15
fix(cmd-j): put keyboard focus on the destination's active surface after jump

### LINKED ISSUE #8903: [Bug]: Cmd-J jump palette leaves keyboard focus on the wrong (often hidden) surface
https://github.com/user-attachments/assets/ddc39076-9a8f-4eaf-b4fb-bef4792b9c79
https://github.com/user-attachments/assets/571e6866-a0a9-4fcc-b237-33f7f266240f

**Operating system:** macOS (root cause is platform-agnostic; also affects Linux/Windows)

**Orca version:** Reproduced on `main` (1.4.x)

---

### Short summary

After using the Cmd-J worktree/tab jump palette to navigate between workspaces, keyboard focus frequently lands on the wrong surface — so the destination looks active but you can't type into it.

### What happens

- **After jumping (Enter) into a workspace:** the terminal cursor renders *hollow* and keystrokes go nowhere; a code or markdown editor opens but won't accept typing (appears un-editable).
- **After cancelling (Esc):** focus is not reliably restored to the surface you were already on.

You have to click into the terminal/editor to recover focus every time.

### Why it happens (root cause)

When the palette closes it restores focus with a generic fallback that grabs the *first* `.xterm-helper-textarea` in the DOM. Inactive workspaces stay mounted (`display:none`) rather than unmounting, so that "first xterm" is often a *hidden* workspace's terminal. The fallback also races the modal's close/teardown animation. Net result: focus lands on a hidden or wrong surface — or on nothing at all.

### How to reproduce

1. Open two or more workspaces, each with an active terminal (or a code/markdown editor tab).
2. Press **Cmd-J** to open the jump palette, select a *different* workspace, and press **Enter**.
3. Without clicking anything, start typing.

**Expected:** keystrokes go to the destination workspace's active surface (terminal cursor solid / editor caret blinking and editable).

**Actual:** the terminal cursor is hollow and typing does nothing, or the editor won't accept input — until you click the surface manually.

Also reproducible on the **Esc** path: open the palette, press Esc, and note focus is not restored to the surface you started on.

Note that I attached two screen recordings: one demonstrating the bug, and another demonstrating the fix.

### Additional cases affected

- **Split-pane terminals:** focus lands on an arbitrary pane rather than the previously-active one.
- **Diff view:** focus can land on the read-only (left) pane, so the diff "doesn't work" for input.
- **Markdown "Rich Editor":** caret does not land in the WYSIWYG editor.

### Notes

- Not platform-specific — the DOM/animation race applies on macOS, Linux, and Windows.
- Has a fix ready (routes focus to the destination workspace's actual active surface from store state instead of DOM guesswork); PR to follow.

jad:
## Summary

Fixes keyboard focus landing on the wrong surface after using the **Cmd-J** worktree/tab jump palette. (Fixes #8903.)

After jumping (Enter) into a workspace the terminal cursor rendered hollow and keystrokes went nowhere, and code/markdown editors opened un-editable; after cancelling (Esc) focus was not reliably restored to the surface you were on.

**Root cause:** the palette restored focus with a generic fallback that grabbed the *first* `.xterm-helper-textarea` in the DOM. Because inactive workspaces stay mounted (`display:none`), that "first xterm" was often a *hidden* workspace's terminal, and the fallback also raced the modal's teardown/close animation — so focus landed on a hidden or wrong surface, or nothing at all.

**Fix:** route focus to the *destination workspace's actual active surface*, chosen from store state instead of DOM guesswork.

- `resolveWorktreeActiveSurfaceFocus(state, worktreeId)` (new, pure, unit-tested) reads `activeTabTypeByWorktree` / `activeTabIdByWorktree` / `activeGroupIdByWorktree` / `terminalLayoutsByTabId` and returns a terminal / editor / fallback target. `WorktreeJumpPalette` dispatches the matching focus routine on both jump and cancel.
- **Terminal** → `focusTerminalTabSurface(tabId, leafId)`, now leaf-scoped so a split-pane terminal focuses the persisted active leaf.
- **Editor** → `focusEditorTabSurface(groupId)` (new module), scoped to the destination tab's split group. Handles Monaco's lazy Suspense mount (retries across ~0.5s of animation frames), the Monaco 0.52+ `.native-edit-context` element, skipping the read-only diff `.editor.original` pane, skipping `display:none` workspaces via `getClientRects()`, and scoping the markdown selector to `.rich-markdown-editor-shell` so PR/Linear comment composers don't win.
- Shared monotonic-token cancellation (`active-surface-focus-generation`, new) so a stale editor retry loop can't steal focus from a newer navigation, and the editor path bails while an inline tab-rename input is open.

**Deferred / out of scope:** browser & simulator jump destinations still use the generic fallback on Enter (Esc already restores browser focus); `useModalReturnFocus`'s unscoped markdown selector is left for a separate cleanup. See commit messages for the full list.

## Screenshots

This is a focus/interaction fix with no visual change to rendered UI. Before/after screen recordings (the bug, and the fixed behavior) are attached to the linked issue #8903.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — the 27 tests added/updated by this PR pass. (The suite has 32 pre-existing failures in unrelated sidebar / browser-pane / PR-comment files that fail identically on clean `main` without this change; they stem from a local Node engine mismatch, not this PR.)
- [x] `pnpm build` — desktop + native build succeed (Swift deprecation warnings only)
- [x] Added high-quality regression tests

New/updated Vitest coverage:
- `focus-editor-tab-surface.test.ts` — focuses the visible Monaco input; skips a `display:none` background editor; skips the read-only diff original and focuses the modified pane; scopes to the destination split group (a sibling-group editor is not focused); falls back to the shell-scoped markdown editor; retries while Monaco mounts; bails while a tab-rename input is open; aborts once superseded.
- `worktree-active-surface-focus.test.ts` — routing for terminal-with-leaf, terminal-without-leaf, editor (with group id), and browser/simulator/unknown → fallback.
- `active-surface-focus-generation.test.ts` — monotonic supersede semantics.
- `workspace-tab-palette-activation.test.ts` — terminal activation asserts exact `(tabId, leafId)`; editor path asserts `focusEditorTabSurface(groupId)`.

Manual: drove the running app over the Electron remote-debug port and asserted `document.activeElement` for terminal, split-pane terminal, Monaco editor, diff (editable pane), markdown rich editor, Esc-restore, and the open-tab path.

## AI Review Report

Reviewed with an AI coding agent across the required dimensions:

- **Cross-platform (macOS/Linux/Windows):** no `metaKey`/path/shell hardcoding introduced; the change is DOM/store logic that is platform-agnostic. The Cmd-J trigger itself is unchanged. No Electron platform-specific branches touched.
- **SSH / remote / local:** flagged and documented — terminal focus has no multi-frame retry budget, so a cold/still-connecting **remote SSH** PTY can miss the focus window and receive no focus. Accepted as strictly better than the prior bug (focusing the *previous* workspace's xterm) and recoverable by a click/keypress; a bounded retry can follow if it proves common.
- **Agent / integration compatibility:** focus resolution reads generic workspace store state (tab type, active tab/group, terminal layout), not agent- or provider-specific fields; works the same for any CLI agent surface.
- **Performance:** no hot-path work added; focus routines run once per palette action, use bounded `requestAnimationFrame` retries, and self-cancel via the monotonic token.
- **UI quality:** no visual change; corrects interaction correctness. Verified terminal/editor/diff/markdown cases in-app.
- **Security:** see below.

Fixed during review: an earlier version focused the read-only diff pane (now skips `.editor.original`); editor query is now scoped to the destination split group so keystrokes can't reach a sibling pane's file.

## Security Audit

Low risk. The change only resolves and calls `.focus()` on DOM elements already mounted in the renderer, selected via `querySelector` on static class/attribute selectors (`data-tab-group-body-id` from store-owned ids). No new IPC, no command execution, no path handling, no network, no secrets, no user-supplied strings interpolated into selectors (group ids are store-generated). No new dependencies. No follow-up security work identified.

## Notes

- Rebased onto latest `main`; no conflicts.
- Remote/SSH caveat (terminal cold-PTY focus miss) documented above and in the commit message.
- Deferred items (browser/simulator Enter-path focus, `useModalReturnFocus` selector scoping) are called out as follow-ups.
- X (Twitter) handle: @debay

### DIFF
--- a/src/renderer/src/components/WorktreeJumpPalette.tsx
+++ b/src/renderer/src/components/WorktreeJumpPalette.tsx
@@ -1307,7 +1307,7 @@ export default function WorktreeJumpPalette(): React.JSX.Element | null {
       if (target.kind === 'terminal') {
         focusTerminalTabSurface(target.tabId, target.leafId)
       } else if (target.kind === 'editor') {
-        focusEditorTabSurface()
+        focusEditorTabSurface(target.groupId)
       } else {
         focusFallbackSurface()
       }
--- a/src/renderer/src/components/WorktreeJumpPalette.tsx
+++ b/src/renderer/src/components/WorktreeJumpPalette.tsx
@@ -36,6 +36,13 @@ import { getLiveAgentStatusByWorktreeId, isInactiveWorkspace } from '@/lib/workt
 import { orderEmptyQueryWorktrees } from '@/lib/order-empty-query-worktrees'
 import StatusIndicator from '@/components/sidebar/StatusIndicator'
 import { cn } from '@/lib/utils'
+import { focusEditorTabSurface } from '@/lib/focus-editor-tab-surface'
+import { focusTerminalTabSurface } from '@/lib/focus-terminal-tab-surface'
+import {
+  beginActiveSurfaceFocus,
+  isActiveSurfaceFocusCurrent
+} from '@/lib/active-surface-focus-generation'
+import { resolveWorktreeActiveSurfaceFocus } from '@/lib/worktree-active-surface-focus'
 import { getWorktreeStatus, getWorktreeStatusLabel } from '@/lib/worktree-status'
 import { activateAndRevealWorktree } from '@/lib/worktree-activation'
 import { findWorktreeById } from '@/store/slices/worktree-helpers'
@@ -1154,6 +1161,11 @@ export default function WorktreeJumpPalette(): React.JSX.Element | null {
 
   useEffect(() => {
     if (visible && !wasVisibleRef.current) {
+      // Reopening Cmd-J means the user is navigating again: supersede any still-
+      // running focus retry from a prior jump so it can't later steal focus off
+      // whatever this navigation lands on — including the browser/simulator/
+      // settings paths that don't run their own focus routine to bump the token.
+      beginActiveSurfaceFocus()
       recordFeatureInteraction('cmd-j')
       createLookupGuard.invalidate()
       activeGroupSnapshotRef.current = captureCmdJActiveGroupSnapshot(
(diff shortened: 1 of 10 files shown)


nwparker:
Thanks! Didn't see this in time but appreacite the PR!
Closed via #8931

We are on the same wavelength

PR #8912, opened 2026-07-15
Prevent adding the same Claude account twice

### LINKED ISSUE #6616: [Bug]: Should not be able to add 2 AI Provider Accounts with the same username?
### Operating system

macOS

### Orca version

_No response_

### Details

By mistake I added twice the same Claude subscription, I suppose it shouldn't let us do it?

Image

brennanb2025:
## What changes

Adding the **same** Claude subscription twice in Settings → AI Provider Accounts no longer creates a duplicate row. When you click **Add Account** and complete the login as an account that is already in the list, Orca now rejects the add with a clear message — "This Claude account is already added." — and leaves the existing row untouched.

Fixes #6616.

## What does not change

- A genuinely new account still adds normally.
- The **same email under a different organization** is still addable (those are distinct accounts).
- Host and WSL (and each WSL distro) still keep independent accounts — the same subscription can be added once per runtime, because each runtime has its own managed-auth store.
- No retroactive cleanup of accounts that are *already* duplicated in a user's settings; this only prevents *new* duplicates. Existing duplicates can still be removed with the per-row **Remove** button.

## Design

`ClaudeAccountService.doAddAccount` (`src/main/claude-accounts/service.ts`) unconditionally appended the captured account with no identity check — the root cause. The fix adds a guard immediately after the login capture and before the account is persisted: if the captured identity matches an existing managed account, it throws, and the pre-existing rollback path (`restoreClaudeSettings` + `safeRemoveManagedAuth`) cleans up the throwaway auth dir.

Identity matching lives in a new pure module `src/main/claude-accounts/claude-duplicate-account.ts` (`findDuplicateClaudeAccount`) so it is unit-testable and does not grow the already-large service file. Two accounts are "the same" only when **email + organizationUuid + runtime scope** all match:

- **email + organizationUuid** (not email alone): the same Claude email can legitimately belong to two organizations.
- **runtime scope** (`managedAuthRuntime` + `wslDistro`, folded through the existing `getClaudeWslSelectionKey`): host vs WSL vs each distro are separate auth contexts.
- Both sides are normalized (email lower/trim, org trim, `undefined` runtime → `host`, `undefined`/`null` org folded together) so an account persisted by an **older Orca build** — before the runtime fields existed — still matches a fresh host add. Without that normalization the exact duplicate would slip through on macOS, which is the reporter's platform.

Blocking (rather than silently folding the add into a re-authentication of the existing account) was the deliberate choice: it matches the reporter's expectation ("it shouldn't let us do it"), keeps the diff small and self-contained, and avoids a hidden credential mutation the user didn't ask for — the per-row **Re-authenticate** action already exists for that.

## Design review

Reviewed against the original issue via an adversarial design pass. It caught one real correctness gap in the first draft: a naive `managedAuthRuntime === 'host'` comparison would fail to match a legacy account whose `managedAuthRuntime` is `undefined`, silently under-blocking the exact duplicate on macOS. The identity key was updated to normalize both sides (and reuse `getClaudeWslSelectionKey` for WSL bucketing) to close that.

## Implementation and deviations

Three product edits, no deviations from the reviewed design: the new pure matcher, the guard in `doAddAccount`, and tests. No renderer change is needed — the thrown message propagates through the existing IPC → `getClaudeAccountErrorDescription` → toast path intact.

## Completeness / headline behavior

**Headline behavior: implemented.** The production `doAddAccount` path executes the guard; verified live in the running app (not just tests) — see Validation below.

## Broad app consistency

- Source-of-truth surface is the Claude section of AccountsPane; the error surfaces through the same `toast.error(...)` path every other Claude account action already uses, so no new UI or copy pattern is introduced.
- The **Codex** provider has the same unguarded append shape in `codex-accounts/service.ts`, and the per-row **Re-authenticate** flow for Claude is likewise unguarded. Both are intentionally **out of scope** for #6616 (which is specifically the Claude "Add twice" flow) and are noted as follow-up candidates rather than folded in, to keep this PR on one thing.

## Risks, ambiguities, non-goals

- **Post-login detection**: identity is only knowable after the OAuth login completes, so the "already added" rejection lands after the user signs in. Unavoidable — there is no earlier point to check who they are. Accepted.
- **Non-goals**: retroactive de-duplication of existing dupes; Codex duplicates; the re-authenticate path.
- **Fail-open edge**: if the same account captured an org UUID on one add and none on another, the two would not match and a duplicate could slip through — accepted as strictly safer than email-only matching, which would falsely merge two legitimate same-email organizations.

## Performance

Trivial surface: a synchronous `Array.find` over `claudeManagedAccounts` (a handful of entries), run once per manual **Add Account** click after a multi-second interactive login. No hot path, polling, render churn, or startup cost. No measurement warranted.

## Code review

Two independent adversarial reviewers ran on the diff (correctness/security lens and simplification/edge-case/cross-platform lens). Both returned clean — no P0/P1/P2 findings — after independently tracing the macOS host case, the legacy-`undefined` normalization, the scoped cleanup on rejection, TOCTOU safety (the add runs inside the serialized mutation queue), and error propagation to the toast. One test-rigor nit (assert the throwaway dir is actually removed) was applied.

## Validation

- **Targeted unit tests: 35/35 pass** — `claude-duplicate-account.test.ts` (11 cases incl. legacy-undefined runtime, undefined-vs-null org, case/whitespace, host-vs-WSL, per-distro) and two new `service.test.ts` integration tests (duplicate rejected + list stays length 1 + throwaway dir cleaned up + rollback ran; same email different org still appends).
- **Full `pnpm run typecheck` (node + cli + web): clean.**
- **`oxlint` on changed files: clean.**
- **Live Electron validation (real Settings UI, stubbed `claude` CLI returning a fixed identity):**
  - *Before the fix:* clicking **Add Account** twice produced **two identical rows** (same email + organization).
  - *After the fix:* first add succeeds (one row); the second add completes the login, is **blocked**, the list **stays at one row**, and the toast shows "This Claude account is already added." The main-process log confirms the throw.
  Before/after screenshots were captured during validation.

## Not merged

Opened for review; unmerged.

### DIFF
--- a/src/main/claude-accounts/claude-duplicate-account.test.ts
+++ b/src/main/claude-accounts/claude-duplicate-account.test.ts
@@ -1,0 +1,113 @@
+import { describe, expect, it } from 'vitest'
+import type { ClaudeManagedAccount } from '../../shared/types'
+import {
+  findDuplicateClaudeAccount,
+  type ClaudeAccountIdentityCandidate
+} from './claude-duplicate-account'
+
+function makeAccount(overrides: Partial<ClaudeManagedAccount> = {}): ClaudeManagedAccount {
+  return {
+    id: 'existing-account',
+    email: 'host@x.com',
+    managedAuthPath: '/managed/existing-account/auth',
+    authMethod: 'subscription-oauth',
+    createdAt: 1,
+    updatedAt: 1,
+    lastAuthenticatedAt: 1,
+    ...overrides
+  }
+}
+
+function makeCandidate(
+  overrides: Partial<ClaudeAccountIdentityCandidate> = {}
+): ClaudeAccountIdentityCandidate {
+  return {
+    email: 'host@x.com',
+    organizationUuid: null,
+    managedAuthRuntime: 'host',
+    wslDistro: null,
+    ...overrides
+  }
+}
+
+describe('findDuplicateClaudeAccount', () => {
+  it('returns the existing account on an exact host match', () => {
+    const account = makeAccount({ managedAuthRuntime: 'host', organizationUuid: null })
+    expect(findDuplicateClaudeAccount([account], makeCandidate())?.id).toBe('existing-account')
+  })
+
+  it('returns null when the organization differs', () => {
+    const account = makeAccount({ managedAuthRuntime: 'host', organizationUuid: 'org-A' })
+    expect(
+      findDuplicateClaudeAccount([account], makeCandidate({ organizationUuid: 'org-B' }))
+    ).toBe(null)
+  })
+
+  it('returns null when a wsl candidate is compared against a host account', () => {
+    const account = makeAccount({ managedAuthRuntime: 'host', organizationUuid: 'org-A' })
+    expect(
+      findDuplicateClaudeAccount(
+        [account],
+        makeCandidate({ organizationUuid: 'org-A', managedAuthRuntime: 'wsl', wslDistro: 'Ubuntu' })
+      )
+    ).toBe(null)
+  })
+
+  it('returns null when both are wsl but on different distros', () => {
+    const account = makeAccount({
+      managedAuthRuntime: 'wsl',
+      wslDistro: 'Ubuntu',
+      organizationUuid: 'org-A'
+    })
+    expect(
+      findDuplicateClaudeAccount(
+        [account],
+        makeCandidate({ organizationUuid: 'org-A', managedAuthRuntime: 'wsl', wslDistro: 'Debian' })
+      )
+    ).toBe(null)
+  })
+
+  it('returns the account when both are wsl on the same distro', () => {
+    const account = makeAccount({
+      managedAuthRuntime: 'wsl',
+      wslDistro: 'Ubuntu',
+      organizationUuid: 'org-A'
+    })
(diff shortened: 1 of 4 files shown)


brennanb2025:
### Manual validation — before/after (real Settings UI, stubbed `claude` CLI)

Validated in the running dev app by driving the real **Settings → AI Provider Accounts → Add Account** flow with a stubbed `claude` CLI that returns a fixed identity, so the double-add is deterministic.

**Before the fix** — clicking **Add Account** twice for the same account produced two identical rows (same email + organization):

- one row `dupe-repro-6616@example.com · Repro Org`
- a second row `dupe-repro-6616@example.com · Repro Org` (the bug)

**After the fix** — the first add succeeds (one row); the second add completes the login, is **blocked**, the list **stays at one row**, and a toast appears:

> Claude account update failed.
> This Claude account is already added.

The main-process log confirms the guard fired:

```
Error occurred in handler for 'claudeAccounts:add': Error: This Claude account is already added.
```

Screenshots were captured during validation (before: two rows; after: one row + block toast). They are attached out-of-band rather than committed, per the repo's no-scratch-in-repo policy.

PR #8828, opened 2026-07-15
feat(codex): flag-gated real-home routing for the system-default account, with app-server-trusted status hooks

brennanb2025:
## Summary

Route the **system-default** Codex account at the user's real `~/.codex` instead of Orca's managed runtime home, behind a staged internal flag (default OFF, no settings UI). Flag OFF is byte-identical to today; managed (multi-account) selections are unchanged in either state.

Credit to @jellychoco (#8606) for the native-home routing direction. Built on the codex app-server trust-grant client (merged from `brennanb2025/codex-trust-rpc-grant`).

Applying Orca's isolated `CODEX_HOME` to the system-default account split all native Codex state (auth, config, sessions, logs) even though that selection needs no credential isolation — the root condition behind the OAuth refresh-token conflict (#5370), invisible Orca sessions (#4444, #8612), config divergence, and duplicate session listings (#7521). This PR makes the system-default selection run on the provider-native home while keeping managed accounts isolated for hot-swap.

## The flag

`codexSystemDefaultRealHomeEnabled` (GlobalSettings, default `false`, no UI). An env override `ORCA_CODEX_SYSTEM_DEFAULT_REAL_HOME=1|0` exists only for isolated dev/CDP verification. See `src/main/codex/codex-real-home-flag.ts`. Flag OFF short-circuits every branch below, so the managed-home behavior is preserved bit-for-bit.

## Routing (flag ON + eligible host system default)

- `CodexRuntimeHomeService.prepareForCodexLaunch` returns `null` for the eligible host system-default selection, so:
  - `buildPtyHostEnv` injects **no** managed `CODEX_HOME`; it strips a nested-Orca-inherited **Orca-owned** override (`CODEX_HOME` equal to the private `ORCA_CODEX_HOME` marker). Both the in-process and daemon spawn paths are covered (the daemon strip goes through `envToDelete` because the daemon spawns from its own inherited environment).
  - The shell-ready re-exports already no-op without the `ORCA_CODEX_HOME` marker, so no change is needed there.
  - The headless commit-message Codex path strips the same inherited override.
  - No system→managed session bridge runs for this lane, so the real home stays the single source of truth.
- Rate-limit / usage: `prepareForRateLimitFetch` passes `~/.codex` explicitly to the fetcher and `codex-auth-presence`, so a nested Orca process cannot fall back to its inherited managed `CODEX_HOME` and re-trigger the #5370 auth war.

[...]

PR #8802, opened 2026-07-15
feat(codex): one-time backfill of managed-home sessions into the real Codex home

brennanb2025:
## Summary

One-time, non-destructive backfill of Codex session rollout files from Orca's managed runtime home into the user's real `~/.codex/sessions/YYYY/MM/DD/` tree.

Codex sessions launched inside Orca currently land only in the Orca-managed `CODEX_HOME`, so the user's own `codex resume` picker and Codex app history never see them (#4444, #8612). This PR makes the historical backlog visible in the provider-native home while changing nothing else about where new sessions are written. It complements the account-ownership direction in #8606 (credit to @jellychoco for the routing direction): once system-default launches move to the native home, this backfill is what keeps pre-existing Orca history discoverable there.

## Behavior

- Walks `<userData>/codex-runtime-home/home/sessions/**` incrementally (same batched walker the existing system→managed bridge uses) and places each rollout at the identical relative path under the real Codex home. The `codex resume` picker enumerates and parses this exact layout.
- **Hardlink first** so resume sees one physical JSONL log; **copy fallback** (`COPYFILE_EXCL`) for cross-volume homes.
- **Never overwrites, deletes, or moves anything in either home**: an existing target path (including a broken symlink) is always skipped, and a target appearing mid-run is caught by `EEXIST` handling.
- Symlinked managed entries (bridge-created links that already point into the user's own home) are not materialized.
- **Once per host behind a versioned marker** (`<userData>/codex-session-backfill/backfill-complete.json`). Per-file failures leave the marker unset so the next startup retries; the retry is cheap because everything already placed is skipped. Runs deferred (15s after service init) so startup and first PTY spawns never compete with the tree walk.
- **Audit log** (`<userData>/codex-session-backfill/audit.jsonl`): one JSONL record per hardlink/copy/failure plus a per-run summary with skip counts.
- Honors the existing custom Codex session source home override (users who run Codex with their own `CODEX_HOME` get history where their `codex resume` actually looks).
- The existing system→managed session bridge is unchanged; the two directions compose safely (both are skip-existing and `EEXIST`-tolerant).

## Not in this PR (follow-ups)

- **WSL**: managed WSL runtime homes are distro-local; the backfill needs an in-distro hardlink variant (mirroring `wsl-codex-session-bridge.ts`). Host (macOS/Linux/Windows) only for now.
- Sessions created *after* this backfill while launches still target the managed home keep flowing through the existing bridge; they become native-home sessions with the system-default routing work (#8606 direction).
[...]

Issue #7521, opened 2026-07-06
[Bug]: Codex session history can duplicate because AI Vault scans both system and Orca runtime session roots


zenos876:
### Operating system

macOS

### What happened

  AI Vault can show duplicate Codex sessions for the same workspace.

  In my case, the same Codex transcript exists in both places:

  - `~/.codex/sessions/.../*.jsonl`
  - `~/Library/Application Support/orca/codex-runtime-home/home/sessions/.../*.jsonl`

  These files can be hardlinks to the same physical transcript, but AI Vault currently treats them as separate session sources and displays both.

  ### Why this is a problem

  For a single physical workspace/project, there should be one session history.

  This becomes especially confusing with SSH:

  - Machine A has the project.
  - Machine B opens Orca and connects to A over SSH.
  - The workspace physically lives on A.
  - AI Vault should show the sessions for that workspace on A, regardless of whether Orca is opened locally on A or remotely from B.

  The current host/source model makes it possible for the same workspace history to depend on which host/source is scanned, and can produce duplicates.

  ### Expected behavior

  For Codex sessions, AI Vault should use Orca’s Codex runtime home as the single display source:

  `.../orca/codex-runtime-home/home/sessions`

  `~/.codex/sessions` should only be used as an import/bridge source for external Codex sessions, not as a second display source.

  On refresh, Orca can bridge/hardlink system Codex sessions into the runtime home first, then scan only the runtime home.

  ### Actual behavior

  AI Vault scans both the system Codex sessions directory and the Orca runtime sessions directory, so the same Codex session can appear twice.

  ### Proposed fix

  - Treat Orca Codex runtime home sessions as the authoritative Codex history shown by AI Vault.
  - Do not scan `~/.codex/sessions` as a parallel display source.
  - For local refresh, bridge/hardlink `~/.codex/sessions` into the runtime home before scanning.
  - For SSH workspaces, scan the remote host’s Orca runtime home, not the client machine’s local session storage.

Related: #7367

nwparker:
**Status: REPRODUCED** (unit test, current tree)

[...]

PR #8921, opened 2026-07-16
feat(codex): session dedup across roots, canonical real-home resume, and sqlite index heal

brennanb2025:
## Stacking

**Stacked on #8828** (flag-gated real-home routing), and via it on #8847 (app-server trust grant); this branch also merges #8802 (session backfill) because the dedup and heal work operates on the backfill's audit ledger and hardlinks. Base is `main` so CI runs, but the diff shown by GitHub includes the stacked PRs — the net-new work is the two commits `22dc271b9` and `2f41d62d3`. Intended to land via the real-home integration branch (flag default ON for the RC), not directly to main.

## What this PR does

Last functional gate before the real-home RC: without it, every backfilled Codex session lists **twice** in AI Vault (#7521, aggravated by the #8802 backfill), resumes can target the frozen managed home whose `auth.json` stops refreshing after the flip, and Codex's own DB-driven surfaces never learn about backfilled rollouts (its sqlite metadata backfill is one-shot).

### 1. Session dedup across Codex roots (`src/main/ai-vault/codex-session-root-dedup.ts`)

The session bridge and the real-home backfill hardlink one physical rollout into both the managed runtime home and the real `~/.codex`, and AI Vault scans both roots. The listing id embeds the file path, so one session became one row per root.

Two collapse passes, both pure and shared by the local, WSL, and SSH-remote scanners:

- **Pre-parse, by rollout file name** — bridge/backfill preserve the `sessions/YYYY/MM/DD/rollout-<ts>-<uuid>.jsonl` layout, so an identical rollout file name across Codex roots is the same session. Duplicate aliases stop consuming the parse budget and can no longer crowd the capped listing.
- **Post-parse, by session id per execution host** — catches what the file-name pass can't see: cross-volume backfill **copies** and `session_meta` ids that differ from the file name.

**Canonical root rule** (lowest rank wins):
1. Host real home (`codexHome: null`) — canonical when present; display once, attribute to the real home.
2. Orca managed runtime home — non-flipped lanes (WSL, remote) keep the managed row so resume keeps Orca-refreshed auth, exactly as today.
3. Other homes (WSL/remote real homes, custom `CODEX_HOME`s).
Ties resolve newest-first, then stable path order. Path handling is separator-agnostic (a posix host scans remote/WSL win32 paths).

### 2. Canonical-root resume

`AiVaultSession.codexHome` drives every resume stamp (AI Vault desktop command, the pre-baked `resumeCommand` used by SSH-remote rows, and the mobile resume path — all through the shared builder). Because the canonical row for a both-roots session is the real-home row, its resume command is a bare `codex resume <id>` with **no `CODEX_HOME` prefix**, and pane launches in the real-home lane inject no managed `CODEX_HOME` (routing from #8828). The hardlink makes this safe: appends through either root continue the same inode. Managed-only sessions (not backfilled — e.g. WSL for now, or a failed backfill copy) keep their managed-home stamp unchanged.

### 3. Sqlite index heal (`src/main/codex/codex-session-index-heal.ts`)

Codex stamps its state-DB `backfill_state` **complete** on first app-server startup, so rollouts hardlinked in later never reach the DB on their own (verified against codex 0.144.4). `thread/read` is Codex's sanctioned lazy-indexing path: it parses the rollout and upserts the thread row. Orca never writes Codex's sqlite schema itself.

The heal pass:
- diffs the backfill **audit ledger** against its own version-scoped **heal ledger** → pending thread ids, most recent first (rollout timestamp order);
- drives `thread/read` per id against the real home (explicit `CODEX_HOME` pin so a daemon-inherited managed home can never redirect it), batched onto **one short-lived `codex app-server` per batch** (50 reads/batch, 2 in-flight reads, 500 ms between batches, per-batch deadline with guaranteed child reap) — no thundering spawn;
- is **interruptible** (`shouldStop` polled between reads/batches; wired to app quit) and **resumable** (per-id ledger appends; transport failures leave unprocessed ids unrecorded for the next pass);
- is a **no-op at steady state**: a completion marker records the audit ledger byte size (append-only), so an unchanged audit resolves with two `stat`s and no spawn;
- classifies per-id results: `healed`, `missing` (rollout deleted since audit), `failed` (recorded, not retried until a version bump); a CLI without the app-server surface or `thread/read` marks the pass `unsupported` with a 24 h re-probe backoff — no retry churn.
- runs chained after the startup backfill settles (`src/main/index.ts`), sharing its 15 s startup deferral.

The app-server stdio JSONL transport was extracted from the trust-grant client into `codex-app-server-session.ts` (spawn, handshake, framing, deadline, oversized-line guard, stderr capability classification, reap) and is now shared by the trust grant and the heal; the trust client's public API and wire behavior are unchanged (its 14 existing tests pass unmodified).

### Unit (all green; `vitest --config config/vitest.config.ts`)
- `codex-session-root-dedup.test.ts` — both-roots → real home wins; managed-only / real-only unchanged; id collisions (same-rank newest-first, stable tie-break); WSL pair → managed runtime home; never collapses across execution hosts or agents; win32-separator paths from a posix host.
- `session-scanner-codex-dual-root.test.ts` — full `scanAiVaultSessions` over a sandboxed real home + managed home with a genuine hardlink: one row, `codexHome: null`, real-home `filePath`, resume command with no `CODEX_HOME`; managed-only session keeps its stamp.
- `remote-session-scanner.test.ts` — bridged rollout in both remote homes collapses to the managed runtime-home row.
- `ai-vault-resume-command.test.ts` — real-home canonical session emits no `CODEX_HOME` stamp (plus existing managed-home stamp cases).
- `codex-session-index-heal.test.ts` (11 cases, scenario-driven stub app-server) — recent-first order; marker no-op (no spawn); audit-growth resume; missing/failed recorded and not retried; batching = one server per batch with bounded in-flight reads; `shouldStop` stops promptly and resumes next pass; unsupported via method-not-found and via missing subcommand; mid-batch server death aborts without recording, retried next pass; nothing-to-heal completes with zero spawns; non-link/copy audit records ignored.
- Full `src/main/ai-vault`, `src/main/codex`, `src/main/codex-accounts`, `src/main/ipc/pty.test.ts`, renderer resume-command suites: **951+ tests passing**; typecheck (node/web/cli) and `pnpm lint` clean.

### Live sandboxed end-to-end (real dev app, real codex 0.144.4, flag ON; sandboxed `HOME` + `ORCA_DEV_USER_DATA_PATH` — the developer's real `~/.codex` untouched and verified so)
1. Seeded the sandbox managed runtime home with 3 genuine rollouts → app startup **backfill** hardlinked all 3 into the sandbox real home (`backfill-complete.json`: 3 scanned / 3 linked; both paths `ino=40679431 nlink=2`).
2. **Heal** chained after backfill: `index-heal-complete.json` `healedThreads: 3`, ledger in recent-first order; codex's own `state_5.sqlite` `threads` table gained all 3 rows (previews + provider populated).
3. `codex app-server` → `thread/list {useStateDbOnly: true, modelProviders: [], sourceKinds: [...]}` against the sandbox home returns **all 3 healed sessions**. (Verification gotchas, confirmed against codex-rs 0.144.4 source: omitting `modelProviders` filters to the *current* config provider, and the DB upsert from `thread/read` lands asynchronously — assert in a fresh server session.)
4. **AI Vault listing** via the running app (`window.api.aiVault.listSessions`): exactly 3 codex rows, zero duplicate session ids, all attributed to the real home (`codexHome: null`) even though each rollout exists at two scanned paths.
5. **Resume via an Orca terminal pane**: the deduped row's command is bare `cd '<cwd>' && codex resume <id>`; the spawned codex process env contains **no `CODEX_HOME`** (real-home lane, asserted via process env inspection). Append-to-same-inode proven through codex itself (`thread/resume` + item injection against the sandbox home): the rollout grew 148 201 → 148 485 bytes at the same inode, and the appended marker is readable through the managed-home path — one physical log.
6. Steady-state relaunch: heal resolves `up-to-date` with no app-server spawn.

Rig notes: one launch inherited the production `ORCA_USER_DATA_PATH` from the launching terminal, which made the scan also pick up the production runtime home — an artifact of launching the sandbox from inside another Orca instance (a scrubbed-env relaunch lists exactly 3/3); the dedup handles such foreign roots correctly regardless (file-name aliases still collapse). macOS PTYs reset `HOME` to the login user, so the pane-resume TUI leg asserted command string + process env, with the append property proven against the sandboxed home via app-server.

## Constraints honored
- No user-visible settings; no writes to Codex's sqlite schema; heal is bounded, interruptible, and a steady-state no-op.
- Cross-platform: separator-agnostic path logic; Windows spawn via the existing `getSpawnArgsForWindows`; WSL/SSH lanes keep today's managed-home resume behavior.
- SSH scanning gains the same dedup; remote resume commands are unchanged for managed rows.


brennanb2025:
## Live E2E validation — PASS (5/5)

Validated live against the real Electron app with the real-home flag **forced ON** (branch head `b31658b64`, codex-cli 0.144.4), using a **sandboxed app instance** (`HOME` + `ORCA_DEV_USER_DATA_PATH` overridden, leaked env scrubbed, temporary Codex credentials). No `codex` was run in an app pane. **The real `~/.codex` was never accessed or modified** (checksum-guarded; sandbox + credentials removed after).

1. **Dedup — PASS.** Duplicates across roots collapsed to one **real-home canonical** row; single-root and same-basename sessions stayed visible (no over-merging).
2. **Separator safety — PASS.** Live cross-root collision fixture passed; Windows/WSL path tests passed (separator-agnostic, no `node:path.basename` bug).
3. **Resume env deletion — PASS across every entry point.** Bare real-home resume with `CODEX_HOME` deleted verified through: CDP, sidebar interception, drag payload, benign PTY spawn, **mobile**, **SSH**, and queued-command paths.
4. **Index heal — PASS.** An interactive rollout moved from absent → visible in the Codex state DB after the bounded background heal; steady-state files stayed byte-identical (genuine no-op); unsupported-CLI backoff engaged.
5. **Lifecycle — PASS.** Clean shutdowns, delayed-spawn protection, child reaping, and **zero orphaned `codex app-server` processes** confirmed (`NO_APP_SERVER_ORPHANS`).

112/112 targeted tests passed; worktree stayed clean. `real_home_touched=no`.

_Automated live validation run as part of the pre-RC hardening sweep._

brennanb2025:
Closing as superseded: this work landed on main via #9501 (real-home routing + self-contained multi-account homes) and shipped in v1.4.149 stable. Content verified present on main (module + wiring spot-checked).

PR #8938, opened 2026-07-16
fix(jira): render issue description/comment images with lightbox

### LINKED ISSUE #8934: [Bug]: Jira issue images do not render in Tasks drawer
### Operating system

macOS

### Orca version

v1.4.142-rc.6 (dev / local build)

### Details

**Short summary:**  
When viewing a Jira issue in Orca Tasks, inline images in the description (and comments) do not appear. Only text is shown.

**What happened?**  
Jira stores rich text as ADF (Atlassian Document Format). Screenshots/pasted images become `media` / `mediaSingle` / `mediaInline` nodes with Media Service IDs, not public HTTPS URLs.

Orca converts ADF → Markdown via `adfToMarkdownText` (`src/main/jira/adf-markdown.ts`), then renders with `CommentMarkdown`. Media nodes were not handled, so images were **silently dropped** during conversion. Issue fetch also did not load `attachment` or resolve authenticated attachment content.

**How can we reproduce it?**
1. Connect Jira in Orca (API token).
2. Open **Tasks** → Jira and open an issue whose description contains pasted screenshots (e.g. acceptance-criteria screenshots).
3. Observe: text (Use Case / Acceptance Criteria) shows; images are missing.
4. Open the same issue in Jira web — images are present.

**Expected:**  
Inline images in description/comments are visible in the issue drawer, with click-to-expand fullscreen preview (X / Esc closes only the preview, not the drawer).

- Root cause is incomplete ADF media support + no authenticated attachment download path, not a user config issue.
- Attachment content is private; needs Basic auth download via `/rest/api/3/attachment/content/{id}`.
- Media UUID ≠ attachment ID; practical mapping uses rendered HTML attachment order + filename fallback.
- Related local work (not yet merged) covers:
  - ADF media → markdown with data-URL images on `getIssue` / comments
  - Single image size limit (5MB), max 12 images
  - Viewport-centered lightbox; Esc/X must not dismiss the issue Sheet
- Compare: Linear already ships markdown media URLs / `inlineMedia`; Jira needs this extra resolution path.

**Relevant code (pre-fix):**
- `src/main/jira/adf-markdown.ts` — ADF → Markdown (no media)
- `src/main/jira/issues.ts` — `ISSUE_FIELDS` without `attachment`
- `src/renderer/src/components/JiraIssueWorkspace.tsx` — `CommentMarkdown` for description/comments

lifefloating:
## Summary

Jira issue screenshots no longer disappear in the Tasks drawer. Inline images in descriptions and comments are resolved from ADF media/attachments and can open in a viewport-centered lightbox (Esc/X closes only the preview, not the issue sheet).

Closes #8934

## Screenshots

- Description with inline screenshots rendered in the issue drawer
- Click-to-expand lightbox above the drawer
- Comment-thread screenshots opening the same lightbox
- Esc/X closing only the lightbox while the issue sheet stays open

(Please attach captures from a real Jira issue with pasted screenshots.)

## Testing

- [x] `pnpm lint` (0 errors; existing unrelated warnings only)
- [x] `pnpm typecheck`
- [x] `pnpm test` (branch-related suites green; full suite had one unrelated flaky `pty-connection` case that passed on re-run)
- [x] `pnpm build`
  - `adf-markdown` / `attachment-images` / `issues` unit coverage for media → data URL conversion
  - `MarkdownImageLightbox` + Sheet interaction
  - `CommentMarkdown` expand-control regression for document variant

## AI Review Report

Reviewed the ADF media path, authenticated attachment download, data-URL embedding limits, and lightbox/Sheet dismissal interaction.

Main risks checked:
- Media UUID ≠ attachment ID mapping (rendered HTML order + filename fallback)
- Memory growth from inlining many/large images (5MB / 12-image caps)
- Esc/X dismissing the parent issue Sheet instead of only the lightbox
- Comment bodies still on compact markdown (would show images but block expand)

Cross-platform: no new shortcuts, shell commands, or path separators. Image fetch uses existing authenticated Jira REST client (main process). Lightbox is renderer UI over Sheets; behavior is the same on macOS/Linux/Windows Electron. No platform-specific keyboard accelerators added (Esc only).

## Security Audit

- Attachment bytes are downloaded in main with existing Jira auth, then embedded as `data:` image URLs for trusted markdown rendering only.
- SVG attachments are excluded from image download to avoid scriptable content.
- Size/count caps limit IPC payload and renderer memory.
- No new shell execution, secrets logging, or untrusted remote image auto-fetch for arbitrary URLs in compact cards.
- Lightbox portals re-enable pointer events under modal Sheets carefully so controls work without leaking dismiss through to the drawer.

## Notes

- Comment threads now use `variant="document"` so they share the description lightbox path.
- `MarkdownImageLightbox` locale keys were synced for `en` / `es` / `ja` / `ko` / `zh`.
- Full `pnpm build` not run in this pass; worth running before merge if CI does not cover it.

image

image

## ELI5

Screenshots in Jira issue descriptions and comments vanished in the Tasks drawer. Images load again and open in a lightbox without closing the whole issue panel.


brennanb2025:
Taking a look!

brennanb2025:
Thanks for the contribution — I rebased this onto current `main` and pushed a maintainer follow-up.

I replaced the custom lightbox portal and Sheet-wide dismissal state with the shared modal dialog primitive, which provides focus trapping, Escape ordering, and focus restoration while using design-system surfaces. Attachment resolution now scans the full metadata set before limiting, fetches only IDs referenced by rendered Jira content, reuses repeated media UUID mappings, respects an aggregate byte budget, uses same-origin provider content URLs for Cloud and Server/DC, and streams image-bearing detail/comment payloads in bounded chunks for remote-runtime/SSH use. Jira comments also keep their compact typography and opt into image expansion explicitly.

Validation: 89 focused tests, full typecheck, touched-file lint, React checks, max-lines checks, and localization checks pass. The repo-wide lint still reports two existing switch-exhaustiveness errors outside this PR.

lifefloating:
> Thanks for the contribution — I rebased this onto current `main` and pushed a maintainer follow-up.
> 
> I replaced the custom lightbox portal and Sheet-wide dismissal state with the shared modal dialog primitive, which provides focus trapping, Escape ordering, and focus restoration while using design-system surfaces. Attachment resolution now scans the full metadata set before limiting, fetches only IDs referenced by rendered Jira content, reuses repeated media UUID mappings, respects an aggregate byte budget, uses same-origin provider content URLs for Cloud and Server/DC, and streams image-bearing detail/comment payloads in bounded chunks for remote-runtime/SSH use. Jira comments also keep their compact typography and opt into image expansion explicitly.
> 
> Validation: 89 focused tests, full typecheck, touched-file lint, React checks, max-lines checks, and localization checks pass. The repo-wide lint still reports two existing switch-exhaustiveness errors outside this PR.

Thanks a lot for the rebase and the maintainer follow-up — really appreciate you taking the time to clean this up!
Happy to help with any further polish if needed.

lifefloating:
I wanted to ask when this PR can be merged — I’m waiting to use this Jira feature. If there’s anything I need to adjust, I can do it now. Thanks!
@brennanb2025

AmethystLiang:
Thanks for this — the ADF media parsing, the lightbox component, and the `TaskPage` Esc-guard fix are all solid, and the test coverage is real rather than decorative.

We did a deep review pass and found enough issues that round-tripping them as individual review comments would be slow. Summary:

**Correctness**

- `buildCommentMediaOptions` hardcodes `/rest/api/3/issue/...` instead of `apiBasePath(site)`. On Jira Server/DC (`/rest/api/2`) that 404s, the error is swallowed by the `catch`, and comment images silently never render.
- `attachment-images.ts:247` — `resolved ??= take(stillQueued ?? matches[0])` falls back to an already-consumed image. With two media nodes sharing an alt filename, the first attachment renders twice and the second never renders. Reproduced locally.
- External media URLs are interpolated into `alt` after only a `^https?://` check. A URL containing `)` closes the link early and the remainder is parsed as markdown. `rehype-sanitize` + `urlTransform` stop it short of XSS, but a Jira field can still inject arbitrary links/images into the rendered body. Needs `encodeURI`.
- `extractAttachmentContentIdsFromHtml` is the single gate for the whole feature and matches only two URL shapes — it misses the thumbnail forms (`/secure/thumbnail/<id>/…`, `/rest/api/3/attachment/thumbnail/<id>`). When it misses, `preferredIds` is empty and **no** images are fetched at all, degrading silently to `*[name]*`. There's no filename↔alt fallback even though `fields.attachment` is already in hand.

**Performance**

- Downloads are serial (`for … await`, up to 12) while holding a Jira concurrency slot. `MAX_CONCURRENT` is 4 and both `getIssue` and `getIssueComments` acquire one, so a single screenshot-heavy issue pins half the pool through up to 24 sequential authenticated fetches.
- No caching. `loadComments` refetches on mount and after every comment post, re-downloading and re-base64-ing every referenced image each time. Description and comments fetch independently, so an image in both downloads twice per open.
- 20 MB total × 1.33 for base64 ≈ 27 MB strings, held in React state *and* in `jiraIssueCache` — which is bounded at 500 entries **by count, not bytes**. Given our renderer-OOM history that's the wrong bound for multi-MB payloads.

**Design note (not blocking this PR)**

Inlining bytes as `data:` URLs is what forces the chunked streaming RPC, the 32 MB guard, and the retained strings. A custom protocol handler would avoid all of it — but there's no prior art for one in this repo and the remote-runtime/SSH path would still need solving, so we're not asking for that here. We'll file it as a follow-up.

**Next step**

To keep this moving we'll push the fixes directly to this branch rather than sending you round after round of comments. Expect that within 24 hours, and happy to walk through any of it afterwards. (If branch permissions block a direct push, we'll open a PR into your branch instead.)

One question in the meantime: has this been exercised against a real Jira issue with pasted screenshots? Everything routes through that `renderedFields` HTML regex and all the current tests use hand-written HTML, so if it doesn't match real Jira output the failure mode is silent.

AmethystLiang:
Follow-up for protocol-handler transport rewrite: https://github.com/stablyai/orca/issues/10724

AmethystLiang (approved):
Approved after plan + code review loops.

Implementation addresses the original review findings:
- media↔attachment pairing no longer re-emits consumed images
- Server/DC comment attachment lookup uses apiBasePath
- markdown-safe external URL destinations (not encodeURI alone)
- wider HTML discovery + Option A multi same-filename alt fallback
- downloads concurrent outside the Jira API semaphore
- main-process attachment data-URL cache with singleflight, lower caps, clear on disconnect/clearToken
- post-map resolution warn with attachment-only stats

Follow-up protocol-handler rewrite: https://github.com/stablyai/orca/issues/10724

AmethystLiang (approved):
Approved after plan + code review loops.

Implementation addresses the original review findings:
- media↔attachment pairing no longer re-emits consumed images
- Server/DC comment attachment lookup uses apiBasePath
- markdown-safe external URL destinations (not encodeURI alone)
- wider HTML discovery + Option A multi same-filename alt fallback
- downloads concurrent outside the Jira API semaphore
- main-process attachment data-URL cache with singleflight, lower caps, clear on disconnect/clearToken
- post-map resolution warn with attachment-only stats

Follow-up protocol-handler rewrite: https://github.com/stablyai/orca/issues/10724

Issue #8595, opened 2026-07-13
[Bug]: "Bold Text" terminal color override is a no-op in the renderer


kdoddy-eci:
### Operating system

macOS

### Orca version

v1.4.137 (macOS 26.5.1, Apple Silicon)

### Details

**Short summary:**
Settings → Appearance → Terminal exposes a **"Bold Text"** color override (persisted as `settings.terminalColorOverrides.bold`, and also importable via the Ghostty importer's `bold-color` key from #958), but the terminal renderer never uses it. Bold text with the default foreground always renders in the theme `foreground` color.

This makes it impossible to replicate iTerm2's "Bold Color" / Ghostty's `bold-color` behavior, where bold text without an explicit color is repainted (e.g. yellow bold headings in Claude Code and other TUI output).

**How can we reproduce it?**
1. Set a bold color — either in Settings → Appearance → Terminal → color overrides → "Bold Text" (e.g. `#ffc600`), or via Import from Ghostty with `bold-color = #ffc600` in `~/.config/ghostty/config`. Confirm it persists in `orca-data.json` under `settings.terminalColorOverrides.bold`.
2. Open a **new** terminal pane (rules out live-pane staleness).
3. Run: `printf '\033[1mBOLD DEFAULT FG\033[0m\n'`

**Expected:** renders in the configured bold color (`#ffc600`), matching iTerm2/Ghostty semantics — applies to bold cells whose foreground is the default; cells with an explicit ANSI/truecolor fg keep their color.

**Actual:** renders in the theme `foreground` color. Explicitly colored bold (`\033[1;33m…`) is unaffected and renders correctly.

**Anything else that might help:**
- Likely cause: upstream xterm.js `ITheme` has no `bold` key, so the override presumably gets dropped when the resolved theme is handed to the terminal. `drawBoldTextInBrightColors` doesn't cover this case — it only remaps ANSI 0–7 foregrounds, not the default fg.
- Suggested fix: honor `terminalColorOverrides.bold` (and the same key in custom themes) for bold cells with default foreground, alongside the existing `drawBoldTextInBrightColors` handling. Alternatively, hide the "Bold Text" field and drop the `bold-color` mapping from the Ghostty importer until supported — today it silently does nothing.

nwparker:
## 🤖 Agent handoff — reproduction kit

**Status:** REPRODUCED · GitHub label `has_repro`  
[...]

PR #8952, opened 2026-07-16
fix(terminal): honor the Bold Text color override

rodboev:
## Summary

Honor Orca's Bold Text color override for bold cells that still use the terminal's default foreground. The pinned xterm core patch adds the optional `bold` theme color to DOM foreground resolution, and the existing addon-webgl patch carries it through atlas configuration, cache equality, and glyph rasterization. Explicit ANSI/RGB foregrounds, non-bold default text, inverse cells, and empty overrides keep their existing behavior.

Refs #8595

## Screenshots

Bold cells that use the terminal's default foreground now pick up the configured Bold Text color instead of the plain default foreground. No screenshot is attached.

## Testing

- [x] `pnpm install --frozen-lockfile`
- [x] `pnpm exec oxlint src/shared/types.ts tests/e2e/terminal-bold-default-foreground.spec.ts`
- [x] `pnpm exec oxfmt --check package.json src/shared/types.ts tests/e2e/terminal-bold-default-foreground.spec.ts`
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/ghostty/mapper-extended.test.ts src/renderer/src/components/terminal-pane/compose-active-terminal-theme.test.ts src/shared/terminal-fonts.test.ts src/renderer/src/runtime/sync-runtime-graph.test.ts`
- [x] Headless DOM/WebGL E2E on the patched branch.
- [x] Matching headless base proof on a temporary `origin/main` worktree, which failed with `dom add screenshot should contain the expected bold-default color`.
- [ ] Headed DOM/WebGL E2E

The focused E2E now passes headlessly with `1 passed (42.0s)` on the patched branch and fails headlessly on temporary `origin/main` with `dom add screenshot should contain the expected bold-default color`. The full suite remains CI-gated.
The E2E drives Orca's `state.updateSettings({ terminalColorOverrides })` path, verifies the composed xterm theme, emits a deterministic ANSI fixture through Node so Windows PowerShell does not print the repro literally, and exercises add, change, and removal of `terminalColorOverrides.bold` against the same buffered row for both renderer modes. The lockfile was regenerated from the final patch files and frozen install passed with both patches applied.

## AI Review Report

- macOS: the issue is macOS-scoped, and the change stays inside the pinned xterm renderer authority.
- Linux and Windows: no platform-specific Orca code, PTY bytes, shell behavior, or process handling changed.
- SSH, remote/local, and provider neutrality: terminal input, serialization, and provider paths are unchanged; the renderer consumes existing cell attributes only.
- Performance: the core path adds one bold-color selection to default foreground resolution. WebGL atlas equality includes the bold color so changes invalidate cached glyphs correctly.
- UI quality: explicit foreground, non-bold default foreground, inverse, and fallback cases remain covered by the focused E2E assertions.
- Residual risk: E2E coverage is headless only, so headed Electron rendering is not covered by this PR's local runs.

## Security Audit

No new IPC, subprocess, filesystem, network, credential, or HTML injection surface was added. The E2E reads only the explicitly scoped reproduction artifact and committed fixture. PTY output remains unchanged.

## Notes

The Orca-side `TerminalColorOverrides.bold` shape and theme composition remain stable. The semantic owner is the pinned xterm core plus addon-webgl renderer stack. Mobile serialization is out of scope.


nwparker:
Maintainer update: we are deferring a local xterm dependency patch for now.

Upstream tracking:
- https://github.com/xtermjs/xterm.js/issues/6032 — proposed dedicated bold theme color
- https://github.com/xtermjs/xterm.js/issues/5774 — current default-foreground behavior is by design

For users: Orca’s **Bold Text** override and imported Ghostty `bold-color` remain unsupported/no-op for bold text using the default foreground. We will revisit once upstream support lands or if we decide to carry the patch locally.

nwparker:
Thank you for implementing this the "real" way, @rodboev 🙏 — we gave it a careful read and it's clean, safe, and the bold color genuinely renders for the *set* case.

The reason we didn't merge it right now is a **rendering regression on the unset path**: after the terminal processes an `OSC 10/110` (default-foreground set/reset) sequence, bold cells that have **no** bold color configured can end up drawn with a stale foreground. That's a visible change for users who never opted into a bold color, which trips the zero-regression bar we're holding for this pass. The approach also patches the vendored `@xterm/xterm` and `@xterm/addon-webgl` bundles and has to stay in sync with `minimumContrastRatio` (4.5) — a maintenance/sync burden the maintainers had previously deferred.

For now we shipped **#9661**, which makes the Settings field and the Ghostty importer *honest* that the override isn't applied yet (your saved value is preserved), so nobody is promised something that doesn't render. Your PR is the reference implementation if/when the team decides to take the xterm patch — genuinely appreciated. Closing in favor of #9661.

nwparker:
hey @rodboev -- my AI closed things out

But this is me chiming in

Would love to see this upstream in xterm: https://github.com/xtermjs/xterm.js/issues/6032

Maybe they'll accept a PR? Then we can have it in Orca too!

Issue #8670, opened 2026-07-14
[Bug]: Deleting a clean worktree fails when it contains an initialised submodule


gerardthehuman:
### Summary

Deleting an Orca workspace fails when its linked Git worktree contains an initialised submodule, even when both the parent worktree and submodule are completely clean.

Orca surfaces the raw Git failure:

```text
Error invoking remote method 'worktrees:remove':
Error: Failed to delete worktree at <worktree-path>.
fatal: working trees containing submodules cannot be moved or removed
```

### Environment

- Orca: 1.4.137
- macOS: 26.5.2 (25F84)
- Git: 2.55.0
- Local linked worktree with one initialised submodule

### Steps to reproduce

1. In a Git repository containing a submodule, create a linked worktree.
2. Initialise the submodule in that linked worktree:

   ```sh
   git -C <worktree-path> submodule update --init
   ```

3. Confirm that the parent and submodule are clean:

   ```sh
   git -C <worktree-path> status --short --untracked-files=all
   git -C <worktree-path>/<submodule-path> status --short --untracked-files=all
   ```

4. Delete the workspace from Orca.

### Actual behaviour

Deletion fails with exit code 128 and:

```text
fatal: working trees containing submodules cannot be moved or removed
```

The worktree remains registered and on disk.

### Expected behaviour

Orca should delete a clean worktree containing an initialised submodule after the normal workspace deletion confirmation.

### Diagnosis

Git requires `--force` to remove a worktree containing an initialised submodule, even when the parent and submodule are clean:

```sh
git -C <main-repository> worktree remove --force <worktree-path>
```

I verified in a two-worktree fixture that this command removes only the selected worktree and its per-worktree submodule metadata; the other worktree and its initialised submodule remain intact.

Running `git submodule deinit` first is not a suitable workaround:

- submodule registration may live in the repository's shared `.git/config`, affecting later submodule operations in other worktrees;
- deinitialising the submodule alone does not remove the per-worktree module metadata that triggers Git's removal guard.

### Suggested fix

- Check the parent worktree and all initialised submodules recursively for tracked and untracked changes.
- If everything is clean, invoke `git worktree remove --force <path>`.
- If anything is dirty, require explicit destructive confirmation before forcing removal.
- Retain Orca workspace metadata when Git removal fails so deletion can be retried.

### Additional context

Git documents that unclean worktrees **or worktrees with submodules** can be removed using `--force`:
[...]

PR #8960, opened 2026-07-16
fix(worktree): recover worktree removal when the tree contains submodules

bang9:
## Summary

Worktrees containing populated submodules can now be removed without exposing a raw Git failure. A normal delete classifies the Git refusal and presents targeted guidance plus the existing Force Delete action; only that explicit user action invokes `git worktree remove --force`.

The behavior is covered across local, WSL, SSH, and headless runtime removal paths. Explicit-force failures are formatted consistently, and metadata is preserved when the first attempt is refused.

## Why explicit force is required

A linked worktree stores initialized submodule Git databases under `.git/worktrees/<id>/modules/`, which removal destroys. Local remote-tracking refs are not proof that those objects still exist remotely: a clean worktree can report `git rev-list HEAD --not --remotes` as zero after its remote branch was deleted and garbage-collected, while the linked-worktree database remains the only copy.

Requiring explicit Force Delete avoids silent data loss. It also removes the prior three-probe safety path entirely, so deletion adds no subprocess fan-out or SSH/WSL probe latency and never runs collateral `git worktree prune` recovery for this case.

## Testing

- [x] `pnpm run lint` (including max-lines, reliability, and localization gates)
- [x] `pnpm run typecheck` (node, CLI, and web)
- [x] 998 focused worktree/runtime/shared/renderer tests
- [x] Live Git smoke: normal removal refusal, clean evidence, and successful explicit `--force` removal
- [x] Live data-loss regression reproduction: clean tree + stale remote-tracking ref + remote GC leaves the submodule commit only in the linked-worktree database
- [ ] Full suite: attempted under Node 24; unrelated relay/WebSocket timing tests and two existing `agent-exec-handler` tests fail in this environment

## CodeRabbit follow-up

- The original SSH gap and zombie delete+prune path are eliminated.
- The sequential-probe timeout, non-isolating probe test, and discarded SSH stderr comments are resolved by removing the probe design.
- Explicit forced SSH failures are formatted as force-removal errors in both IPC and runtime paths.

## Security and performance

- No shell command strings, new IPC surface, dependencies, polling, caches, or background work.
- Normal refusal performs one existing removal attempt and zero additional Git probes.
- Recursive filesystem deletion remains behind the existing explicit Force Delete and worktree safety gates.

## ELI5

Deleting a worktree that still had Git submodules failed with a raw error. Orca explains the problem and offers Force Delete only when you explicitly choose it.


brennanb2025:
Thanks @bang9! The diagnosis was right on — deleting submodule worktrees was completely broken. While reviewing I found that git's refusal only applies to *non-forced* removal (on every supported git ≥2.17, `git worktree remove --force` performs a normal removal), so I pushed a rework to your branch that leans on git's own override, gated behind explicit user consent:

- When git refuses the plain delete, Orca now surfaces a clear classified error that routes into the existing **Force Delete** flow — so deletion is always possible, but never silently forced.
- Why not auto-force even for clean trees: a linked worktree stores each submodule's entire git database under `.git/worktrees/<id>/modules/`, which removal destroys. Commits that exist only there are unrecoverable, and remote-tracking refs can't *prove* objects still exist on the remote — so no probe can make an automatic `--force` truly safe.
- This also avoids the `git worktree prune` cleanup path, which collaterally deregisters unrelated prunable worktrees, and keeps normal branch-deletion semantics on the forced removal.
- The fix covers the SSH and headless-runtime removal paths too, which hit the same refusal.

Your classifier and its tests are still the backbone of the fix. Validated with unit/integration coverage across all four removal surfaces plus live testing against a real submodule repo (clean delete via Force Delete, dirty gate, unpushed-commit protection).

bang9:
> Thanks @bang9! The diagnosis was right on — deleting submodule worktrees was completely broken. While reviewing I found that git's refusal only applies to _non-forced_ removal (on every supported git ≥2.17, `git worktree remove --force` performs a normal removal), so I pushed a rework to your branch that leans on git's own override, gated behind explicit user consent:
> 
> * When git refuses the plain delete, Orca now surfaces a clear classified error that routes into the existing **Force Delete** flow — so deletion is always possible, but never silently forced.
> * Why not auto-force even for clean trees: a linked worktree stores each submodule's entire git database under `.git/worktrees/<id>/modules/`, which removal destroys. Commits that exist only there are unrecoverable, and remote-tracking refs can't _prove_ objects still exist on the remote — so no probe can make an automatic `--force` truly safe.
> * This also avoids the `git worktree prune` cleanup path, which collaterally deregisters unrelated prunable worktrees, and keeps normal branch-deletion semantics on the forced removal.
> * The fix covers the SSH and headless-runtime removal paths too, which hit the same refusal.
> 
> Your classifier and its tests are still the backbone of the fix. Validated with unit/integration coverage across all four removal surfaces plus live testing against a real submodule repo (clean delete via Force Delete, dirty gate, unpushed-commit protection).

Thanks! There are several ways to handle this, so I created a PR as a starting point. I think the right approach is to implement it according to your policy. Thanks for addressing this so quickly!

Jinwoo-H:
Thanks @bang9 for the initialized-submodule worktree-removal fix. Completed issue #8670 and merged fix #9096 now own this behavior on current main. We will retain your contribution as historical credit and request no rebase, retest, or other follow-up. Closing this superseded branch.

PR #8783, opened 2026-07-15
fix: preserve scroll position when switching OMP file tabs (#8715)

addielaruee:
## Summary

Adds a `key={activeFileTab.id}` prop to the `FileReader` usage in the session screen so React remounts it on every file tab switch. Without the key, React reuses the same component instance when switching between two file tabs (same component type, same tree position), which meant the scroll offset restoration added in the previous commit only fired on the very first mount and not on subsequent tab switches. This closes the gap flagged in review on the original scroll-position PR.

## Screenshots

No visual change. The behavior only affects scroll position handling when switching between open file tabs.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [ ] `pnpm build` (mobile has no build script; this change is scoped to `mobile/app/h/[hostId]/session/[worktreeId].tsx` and does not touch the desktop Electron build)
- [x] Manually traced the remount path: keying by `activeFileTab.id` forces `FileReader` to unmount and remount on tab switch, so the existing `useEffect` that seeks to `initialScrollOffset` on `doc?.status === 'ready'` now runs for every tab instead of only the first one rendered

Ran from `mobile/`: 238 test files passed (1679 tests, 2 skipped), lint and typecheck both clean.

## AI Review Report

Reviewed the diff for correctness and side effects of adding the key prop. The main risk was whether keying would cause unnecessary remounts elsewhere in the tree, but the key is scoped to `FileReader` itself, so no other components in `SessionScreen` are affected. Confirmed the fix does not change behavior on macOS, Linux, or Windows since it is a pure React reconciliation change with no platform-specific APIs, shortcuts, paths, or shell calls involved.

## Security Audit

No input handling, command execution, path handling, auth, secrets, dependency, or IPC changes in this diff. The change is limited to a React `key` prop derived from an already-trusted internal tab id.

## Notes

No platform-specific behavior. Applies equally on macOS, Linux, and Windows since it is pure React state/reconciliation logic in the mobile app.

## ELI5

Switching between open file tabs forgot your scroll position. Each file tab remounts cleanly so scroll restore runs every time you switch.

Issue #8715, opened 2026-07-14
[Bug] OMP: Scroll position resets to top when switching tabs


AmethystLiang:
## Bug Report

**Reporter:** Walkin (via WeChat)

### Description
When using OMP (Orca Mobile Preview), if you move the scroll bar, switch tabs, and then return to the original tab, the scroll bar immediately returns to the top.

### Steps to Reproduce
1. Open OMP view
2. Scroll down in the content
3. Switch to a different tab
4. Switch back to the original tab

https://github.com/user-attachments/assets/01168748-6b5b-46a6-a4d7-858879d88db2

### Expected Behavior
Scroll position should be preserved when returning to the tab.

### Actual Behavior
Scroll position resets to the top of the content.

### Evidence
Video recording attached below.

---

Made with [Orca](https://github.com/stablyai/orca) 🐋

zesanrahim:
Hi @AmethystLiang, were you able to repro this bug consistently?

addielaruee:
I will work on this.

nwparker:
## 🤖 Agent handoff — reproduction kit

**Status:** REPRODUCED · GitHub label `has_repro`  
**Writeup (branch):** [`docs/bug-reproductions/8715.md`](https://github.com/stablyai/orca/blob/nwparker/bug-repro-handoffs/docs/bug-reproductions/8715.md)  
**Branch with tests + docs:** [`nwparker/bug-repro-handoffs`](https://github.com/stablyai/orca/tree/nwparker/bug-repro-handoffs)

### Summary
OMP (and other full-screen TUIs) use xterm’s alternate buffer. `scheduleSplitScrollRestore` **skips** restore when `bufferType === 'alternate'` (restore-during-draw knocks the TUI cursor — #1298). Tab switch → no scroll re-pin → viewport appears at top.

### Automated re-run
```bash
git fetch origin nwparker/bug-repro-handoffs
git checkout nwparker/bug-repro-handoffs
# or: git worktree add /tmp/orca-repro-8715 origin/nwparker/bug-repro-handoffs

pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/lib/pane-manager/repro-8715-omp-alt-screen-scroll.test.ts
```

### Code / evidence files
[...]

PR #8968, opened 2026-07-16
fix: preserve terminal scroll across ADE tab switches

byh3071-cpu:
## Summary
- Inactive terminal tabs no longer use `display: none` while the worktree stays active; they stay laid out with `opacity: 0` so xterm keeps its viewport.
- Legacy terminal surface under editor/browser tabs uses the same opacity park instead of Tailwind `hidden`, which was resetting scroll when switching e.g. PowerShell ↔ file tabs.
- Adds `resolveTerminalTabSurfaceStyle` + unit tests for the visibility policy.

## Screenshots
No visual change (opacity-hidden inactive tabs should look the same). Behavior change: scroll position is preserved after switching top ADE tabs and returning.

## Testing
- [x] Added unit tests for `resolveTerminalTabSurfaceStyle`
- [ ] `pnpm lint` (local sparse checkout / deps not fully installed on this machine)
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`
- Manual repro expected: scroll a terminal up → switch to another titlebar tab (terminal or editor) → switch back → viewport should stay put (Windows ADE).

## AI Review Report
Checked Windows Electron tab hide/show path: `display: none` applied in render before hide effects capture scroll, so remembered viewport became line 0. Verified overlay + legacy `Terminal.tsx` containers both needed the opacity park. Cross-platform: CSS opacity/`pointer-events` behavior is shared on macOS/Linux/Windows; no shortcut/path/shell changes.

## Security Audit
No input handling, command execution, path, auth, secrets, dependency, or IPC changes.

## Notes
Related to scroll-reset reports such as #7118 / tab-switch UX; OMP file-tab case remains #8715 / #8783.

## ELI5

Switching away from a terminal tab and back reset your scroll position. Hidden terminals stay laid out (just invisible) so xterm keeps where you were reading.

### DIFF
--- a/src/renderer/src/components/Terminal.tsx
+++ b/src/renderer/src/components/Terminal.tsx
@@ -2290,7 +2290,7 @@ function Terminal(): React.JSX.Element | null {
       ) : null}
 
       {!effectiveActiveLayout && !anyMountedWorktreeHasLayout && (
-        <>
+        <div className="relative flex flex-col flex-1 min-h-0 min-w-0 overflow-hidden">
           {/* Why: split-group layouts render their own terminal/browser/editor
               surfaces through TabGroupPanel plus stable overlay layers.
               Keeping the legacy workspace-level panes mounted underneath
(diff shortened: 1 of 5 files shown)


nwparker:
Might be some perf concerns

nwparker:
Thanks for the PR. We tested this in a live Electron build on macOS with 500 lines of scrollback, scrolled to a pinned viewport, and switched terminal → README editor → terminal repeatedly. We also forced the hidden terminal surface back to the pre-change `display: none` behavior; the xterm measured 0×0 while hidden, but returned to the same scroll position through eight rapid round-trips.

We cannot reproduce the scroll reset currently. Could you please share:

- Your OS and version, plus the Orca version/commit you reproduced on
- The shell/runtime involved (PowerShell, WSL, SSH, etc.)
- Exact reproduction steps, including whether these are titlebar tabs or split-group tabs and whether the workspace came from an older/migrated session
- A short screen recording showing the reset
- If possible, a failing test or before/after `viewportY` and `baseY` values on the parent commit

With a deterministic repro—especially if this is Windows-only—we can validate the fix and its performance tradeoff confidently.

byh3071-cpu:
Thanks for testing on macOS — that helps narrow it.

### What we can say confidently
- **OS:** Windows 11 Home, build **26200**
- **Orca:** installed desktop app from GitHub releases updater (`stablyai/orca`), around **v1.4.143** (exe under `%LOCALAPPDATA%\Programs\orca`, updated 2026-07-16 local time)
- **Shell / runtime:** local **Windows PowerShell** terminal tabs (not WSL/SSH in the cases we hit)
- **Tab UI:** **titlebar unified tabs** (PowerShell tabs + editor file tabs such as `agent-roster.yaml`). We have **not** confirmed whether a split-group layout is active in that workspace.
- **Symptom:** after scrolling a terminal **up** (away from bottom), switching to another titlebar tab and back sometimes jumps the viewport to the **top**. Not every switch — **intermittent**.
- **Screen recording caveat:** when we try to capture it, the reset often **does not appear**; when not recording, it shows up more often. So we don't have a clean clip yet. This smells timing/perf-sensitive on Windows rather than a 100% deterministic path.

### What we are *not* sure about
- Whether the workspace is an older/migrated session vs freshly created
- Exact before/after `viewportY` / `baseY` (no debug hook attached yet)
- Whether the same jump also happens on **sidebar worktree card** switches (initial report mentioned “cards”; later clarification was titlebar tab switches). Primary repro we’ve chased is **titlebar tab** switches.
- Whether large agent/CLI scrollback makes it more likely (sessions are often long PowerShell + agent output)

### Repro sketch (when it hits)
1. Open a PowerShell tab with enough scrollback to scroll up
2. Scroll up so the viewport is clearly not at the bottom
3. Switch to another titlebar tab (editor file or another terminal)
4. Switch back — **sometimes** the viewport is at the top

Happy to instrument `viewportY`/`baseY` or try a failing test if you have a preferred Windows debug approach. If it stays unrepro on macOS, treating this as **Windows-only / intermittent** seems fair — we’re fine iterating on a smaller fix once there’s a more deterministic path (and we understand the perf tradeoff of keeping hidden tabs laid out).

PR #8987, opened 2026-07-16
fix(projects): treat symlinked directories as directories in Add Project browsers

### LINKED ISSUE #8983: [Bug]: Symlinked directories are shown as files in the Add Project folder browser
# [Bug]: Symlinked directories are shown as files in the Add Project folder browser

**Operating system:** macOS (also affects Linux; remote POSIX hosts affected too)
**Orca version:** v1.4.143

## Details

**Short summary:**
In the Add Project / Add Repo folder browser, a symbolic link that points to a directory is listed as a file. You cannot descend into it, so any project living behind a symlinked path can't be reached through the in-app browser.

**Reproduction steps:**
1. Create a symlink to a directory, e.g. `ln -s ~/real-projects ~/projects-link`.
2. Open Orca → Add Project → browse the host filesystem to the parent folder (`~`).
3. `projects-link` is rendered as a file (no folder icon, not navigable), even though it points to a directory.

The same happens when browsing a remote POSIX host in the Add Remote Project flow.

**Where the bug lives (from reading the source):**

- Local/server browse: `browseServerDir` in `src/main/runtime/orca-runtime.ts` maps `readdir(dirPath, { withFileTypes: true })` entries with `entry.isDirectory()`. `Dirent.isDirectory()` does not follow symlinks, so a symlink-to-directory yields `isDirectory: false`.
- SSH browse (Add Remote Project): `src/main/ipc/ssh-browse.ts` parses `ls -1Ap` output and classifies entries by the trailing `/`. `ls -p` does not dereference symlinks, so a symlink-to-directory has no trailing slash and is parsed as a file. (The Windows PowerShell fallback uses `PSIsContainer` and is unaffected.)

Notably, the relay filesystem handler already does this correctly: `src/relay/fs-handler.ts` stats symlink entries to classify them by target type, with test coverage ("readDir reports symlinked directories as directories"). The fix would mirror that pattern in the two browse paths above.

(The local file-explorer `fs:readDir` intentionally treats symlinks as file-like to avoid macOS TCC prompts — this report is only about the Add Project / Add Remote Project browsers, where the user explicitly navigates.)

Neither `browseServerDir` nor `ssh:browseDir` currently has a symlink test case.

Happy to submit a PR for this.

choihjin:
## Summary

Fixes #8983.

In the Add Project / Add Remote Project folder browsers, a symbolic link pointing to a directory was listed as a file — no folder icon, not navigable — so any project living behind a symlinked path could not be reached or selected.

Two classification sites had the same bug:

- **`browseServerDir`** (`src/main/runtime/orca-runtime.ts`) used `Dirent.isDirectory()`, which does not follow symlinks. Symlink entries are now classified by their target type via `stat`, mirroring the existing (and test-locked) behavior of the relay `fs.readDir` handler in `src/relay/fs-handler.ts`.
- **`ssh:browseDir` POSIX branch** (`src/main/ipc/ssh-browse.ts`) derived the directory marker from `ls -p`, which does not dereference links. The listing now appends the `/` marker via `[ -d ]` (which follows symlinks) in a plain POSIX `sh` loop. The `ls` output is captured into a variable first so a failing `ls` (e.g. permission denied) still propagates through the `&&` chain as a non-zero exit code instead of being masked by the while-loop pipeline — preserving the existing error semantics.

Deliberately unchanged: the local file-explorer `fs:readDir` and runtime file listing intentionally treat symlinks as file-like to avoid macOS TCC prompts (documented and test-locked); the Windows remote fallback already classifies symlinked directories correctly via `PSIsContainer`.

## Screenshots

Remote filesystem browser (Add Remote Project) on a real Linux SSH host, running this branch. `linked-dir` is a symlink to `real-dir`; it now renders as a navigable directory, sorts with directories, and can be descended into and selected as a project (breadcrumb reached `/tmp/orca-symlink-verify/linked-dir`):

Remote file browser showing a symlinked directory classified as a folder

Before this change the same entry rendered with a file icon and could not be entered.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

New regression tests (both watched failing before the fix):

- `orca-runtime.test.ts`: `browseServerDir` over a real temp directory containing a symlinked directory, a symlinked file, and a broken symlink — asserts symlink-to-dir classifies as a directory, symlink-to-file and broken links stay file-like, and sort order holds. Uses `junction` on Windows, matching the existing relay symlink test.
- `ssh-browse.test.ts`: locks the new POSIX listing command and asserts symlinked-directory entries parse as directories; existing command-string assertions updated via a shared helper.

Manual verification: drove the built dev app end-to-end against a real Linux SSH host — browsed into the symlinked directory and confirmed it is selectable as a project. The new shell snippet was additionally exercised directly in `sh` against fixtures with spaces in names, an empty directory, a broken symlink, and a permission-denied directory (still exits non-zero).

## AI Review Report

Review by Claude (Fable 5), which also authored the change. Checks performed:

- **Cross-platform (macOS/Linux/Windows):** the `browseServerDir` change is pure `node:fs` and platform-neutral; Node reports Windows junctions as symlinks, and the new test uses `junction` on win32. The SSH change touches only the POSIX branch; the Windows PowerShell fallback (`PSIsContainer`, already symlink-correct) is untouched. No shortcuts, labels, or path-separator handling involved.
- **SSH/remote/local:** the remote listing keeps a single exec round-trip (no per-entry SSH calls) and preserves exit-code error semantics for permission-denied/unreadable directories (verified in a real shell). Verified live over SSH against a Linux host. The relay/daemon `fs.readDir` path already handled symlinks and is unchanged.
- **Shell portability:** the new snippet uses only POSIX `sh` constructs (`command ls -1A`, `[ -d ]`, `printf`, `while IFS= read -r`) — no bashisms, no `find -maxdepth`, no `pipefail` — so dash/busybox login shells are fine. Filenames with spaces are preserved; filenames containing newlines remain a pre-existing limitation of the line-based protocol, unchanged by this PR.
- **Performance:** extra `stat` calls are issued only for symlink entries and run concurrently; non-symlink entries take the previous fast path. Remote listing adds one subshell and a builtin test per entry inside the existing single exec.
- **Behavioral risk:** broken symlinks classify as files (matching relay behavior); the browse target itself was already `stat`-validated, so selecting a symlinked folder keeps working as before.

## Security Audit

- The remote command keeps the existing `shellEscape` for the only user-controlled input (the directory path); the added shell text is a static string with no interpolation, and `"$f"` values come from `ls` output inside the already-escaped target directory — no new injection surface.
- No new IPC endpoints, no auth/secrets/dependency changes. Local symlink following is confined to classifying entries of a directory the user explicitly browsed; the relay ACL model is untouched (the raw-exec browse path exists precisely because browsing happens before workspace roots are registered, unchanged).

## Notes

- POSIX remote hosts get the fix via the new listing command; Windows remote hosts were already correct.
- The local file explorer's intentional "symlinks are file-like" TCC-avoidance behavior is out of scope and unchanged.

## ELI5

Symlinked folders in Add Project looked like plain files and couldn’t be opened. They’re treated as real directories now so you can navigate into them and pick projects.

### DIFF
--- a/src/main/ipc/ssh-browse.test.ts
+++ b/src/main/ipc/ssh-browse.test.ts
@@ -25,6 +25,12 @@ function createMockChannel(): EventEmitter & { stderr: EventEmitter } {
   })
 }
 
+// The POSIX listing appends `/` via `[ -d ]` (which follows symlinks) instead of
+// `ls -p` (which does not), so symlinked directories classify as directories.
+function posixBrowseCommand(cdTarget: string): string {
+  return `cd ${cdTarget} && pwd && entries=$(command ls -1A) && printf '%s\\n' "$entries" | while IFS= read -r f; do if [ -d "$f" ]; then printf '%s/\\n' "$f"; else printf '%s\\n' "$f"; fi; done`
+}
+
 // Recover the PowerShell script from a `powershell.exe ... -EncodedCommand <b64>`
 // command so tests can assert on the actual (UTF-16LE) payload sent to the host.
 function decodeEncodedCommand(command: string): string {
@@ -68,13 +74,41 @@ describe('registerSshBrowseHandler', () => {
         { name: 'README.md', isDirectory: false }
       ]
     })
-    expect(exec).toHaveBeenCalledWith('cd "$HOME" && pwd && command ls -1Ap')
+    expect(exec).toHaveBeenCalledWith(posixBrowseCommand('"$HOME"'))
     expect(channel.listenerCount('data')).toBe(0)
     expect(channel.listenerCount('exit')).toBe(0)
     expect(channel.listenerCount('close')).toBe(0)
     expect(channel.listenerCount('error')).toBe(0)
     expect(channel.stderr.listenerCount('data')).toBe(0)
     expect(channel.stderr.listenerCount('error')).toBe(0)
+  })
+
+  it('classifies remote symlinked directories as directories', async () => {
+    const channel = createMockChannel()
+    const exec = vi.fn().mockResolvedValue(channel)
+    const getConnectionManager = () => ({
+      getConnection: () => ({ exec })
+    })
+    registerSshBrowseHandler(getConnectionManager as never)
+
+    const resultPromise = handler(null, { targetId: 'ssh-1', dirPath: '/srv' })
+    await Promise.resolve()
+    // `[ -d ]` follows symlinks, so the remote loop emits `linked-dir/` even
+    // though it is a symlink; `ls -p` alone would have printed it bare.
+    channel.emit('data', Buffer.from('/srv\nlinked-dir/\nreal-dir/\nbroken-link\nlinked-file\n'))
+    channel.emit('exit', 0)
+    channel.emit('close')
+
+    await expect(resultPromise).resolves.toEqual({
+      resolvedPath: '/srv',
+      entries: [
+        { name: 'linked-dir', isDirectory: true },
+        { name: 'real-dir', isDirectory: true },
+        { name: 'broken-link', isDirectory: false },
+        { name: 'linked-file', isDirectory: false }
+      ]
+    })
+    expect(exec).toHaveBeenCalledWith(posixBrowseCommand("'/srv'"))
   })
 
   it('escapes remote browse paths before invoking command ls', async () => {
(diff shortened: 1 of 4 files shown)


choihjin:
Follow-up #8998 (stacked on this) adds a visual symlink indicator in the same browsers, so a linked directory is distinguishable from a real one — the natural next step now that they behave identically. It also extends `ssh:browseDir` to carry symlink state.

brennanb2025:
Taking a look!

Issue #5232, opened 2026-06-12
[Bug]: Claude, Codex Sessions Restoration enhancement


NirajRChaudhari:
### Problem or use case

Hi @AmethystLiang / @Jinwoo-H , attached below are few bugs that can be resolved to make the restore session more robust. In addition to your PR fix https://github.com/stablyai/orca/pull/5230

 
# Root Cause Analysis: Claude / Codex / Gemini Sessions Not Restored After Quit

Two independent bugs combine to produce empty terminals on reopen. They are at different layers and need separate fixes.

---

### Symptom

Terminals reopen empty. PTY IDs in `orca-data.json` are small integers (`1, 2, 3, 5, 6`) instead of UUID-format daemon session IDs.

### Root Cause

`src/main/startup/first-window-startup-services.ts` resolves `localPtyReady` on either the daemon becoming ready OR the first-window timeout, whichever comes first:

```ts
// before fix
const localPtyReady = allServicesReady // includes firstWindowTimeout
```

When the daemon is slow to start, the first-window timeout fires first.

`pty:spawn` waits on `localPtyReady`, then selects a provider. After timeout it picks `LocalPtyProvider` instead of `DaemonPtyProvider`.

`LocalPtyProvider` assigns numeric integer IDs.

On quit:

```ts
killAllPty()
```

in `pty.ts:2906` destroys all local PTYs. There is nothing for warm reattach to reconnect to on restart.

### Fix

**File:** `src/main/startup/first-window-startup-services.ts`

```ts
// after fix
const localPtyReady = daemon.ready // wait for daemon specifically; first-window can still fail-open
```

The first window can still open after timeout (user sees the app), but terminal spawning blocks until the daemon is actually ready rather than falling back to non-restorable Local PTYs.

### Evidence

Local data shows:

```json
ptyId: 1
ptyId: 3
ptyId: 6
```

all pointing to `LocalPtyProvider` sessions.

The v12 daemon is running but reports:

```text
0 live sessions
```

`terminal-history/` contains daemon checkpoint files keyed by:

```text
worktreeId@uuid
```

but there are no corresponding entries for those numeric IDs.

---

### Symptom

Even when daemon PTYs are used correctly, Claude/Codex/Gemini sessions are not resumed via:

```bash
claude --resume
codex resume
gemini --resume
```

after restart.

The terminal opens a fresh shell instead of resuming the conversation.

### Root Cause

[...]

PR #8991, opened 2026-07-16
fix(agents): include OMP terminals in cold session restoration

### LINKED ISSUE #8962: [Bug]: OMP terminals are excluded from cold session restoration
### Operating system

macOS

### Orca version

1.4.124

### Details

OMP terminals report status through Orca's managed Pi-compatible extension, but Orca never records OMP's provider session ID. When a restart or crash requires a cold terminal restore (the prior PTY cannot be reattached), Claude is relaunched with `claude --resume <session-id>`, while OMP cannot enter the same restoration path and falls back to an unresumed terminal.

`omp v17.0.1 --help` confirms the supported interface is:

```text
-r, --resume=<value>  Resume a session (by ID prefix, path, or picker if omitted)
```

#### Reproduction

1. Open an Orca terminal and launch `omp`.
2. Send at least one prompt so OMP creates a persisted session.
3. Restart Orca in a way that forces cold restoration, such as terminating the PTY daemon with the app rather than allowing warm daemon reattachment.
4. Reopen the workspace and restored terminal.

Expected: Orca starts `omp --resume <session-id>` in the existing pane and shows the session-restored banner, matching Claude/Codex behavior.

Actual: no OMP sleeping-session record exists, so `pty-connection.ts` has no provider session from which to build a cold-resume startup command.

#### Code path and root cause

The general restoration pipeline is already present:

1. [`agent-hook-listener.ts`](https://github.com/stablyai/orca/blob/main/src/shared/agent-hook-listener.ts) calls `extractAgentProviderSession(...)` and attaches the result to the agent status row.
2. [`agent-status.ts`](https://github.com/stablyai/orca/blob/main/src/renderer/src/store/slices/agent-status.ts) only creates a `SleepingAgentSessionRecord` when the agent is resumable, has `providerSession`, and `getAgentResumeArgv(...)` succeeds.
3. [`pty-connection.ts`](https://github.com/stablyai/orca/blob/main/src/renderer/src/components/terminal-pane/pty-connection.ts) consumes that record only on cold restore and builds the replacement startup command. This preserves the #5232/#5240 invariant: warm reattachment must not launch a duplicate agent tab/session.

OMP is blocked at every input to that pipeline in [`agent-session-resume.ts`](https://github.com/stablyai/orca/blob/main/src/shared/agent-session-resume.ts):

```ts
export const RESUMABLE_TUI_AGENTS = [
  'claude',
  'codex',
  // ...
  'devin'
] as const
```

`omp` is absent from `RESUMABLE_TUI_AGENTS`; `extractAgentProviderSession` explicitly returns `null` for it; and `getAgentResumeArgv` has no OMP case.

The missing source fact is visible in the generated managed extension. [`agent-status-extension-source.ts`](https://github.com/stablyai/orca/blob/main/src/main/pi/agent-status-extension-source.ts) posts the supplied `extra` object unchanged:

```ts
payload: { hook_event_name: hookEventName, ...extra }
```

[`agent-status-handler-source.ts`](https://github.com/stablyai/orca/blob/main/src/main/pi/agent-status-handler-source.ts) registers handlers without the extension context and only posts prompt/tool/message fields:

```ts
pi.on('before_agent_start', (event) => {
  post('before_agent_start', { prompt: event.prompt ?? '' })
})
```

OMP's extension API already exposes the authoritative ID as `ctx.sessionManager.getSessionId()`. The CLI accepts that ID through `omp --resume <id>`, so Orca does not need to infer a session from paths, titles, or process state.

#### Possible fix

1. In `agent-status-handler-source.ts`, accept the second `ctx` argument for each status handler and attach `session_id: ctx.sessionManager.getSessionId()` to every posted snapshot. Attach it to every event rather than only the first because the transport intentionally keeps only the latest pending post, and sessions can switch within a process.
2. In `agent-session-resume.ts`:
   - add `'omp'` to `RESUMABLE_TUI_AGENTS`;
   - parse OMP's `session_id` into `{ key: 'session_id', id }`;
   - return `['omp', '--resume', id]` from `getAgentResumeArgv`.
3. Keep `pi` unsupported unless its CLI resume contract is independently verified; this issue only needs the explicit OMP contract.
4. Add focused coverage:
   - the generated OMP hook payload includes `session_id` from the extension context;
   - `extractAgentProviderSession('omp', ...)` and `getAgentResumeArgv('omp', ...)` round-trip the ID;
   - quit/live capture persists an OMP sleeping record and cold restore launches `omp --resume <id>` in the same pane;
   - warm daemon reattachment still does not launch a duplicate OMP process.

This should reuse the existing sleeping-session schema and cold-restore flow rather than adding an OMP-specific restoration path.

xianjianlf2:
OMP terminals report status through Orca's managed Pi-compatible extension, but Orca never recorded OMP's provider session ID. On a cold terminal restore (the prior PTY can't be reattached), Claude/Codex relaunch via `--resume <id>` while OMP had no sleeping-session record and fell back to an unresumed terminal (#8962).

This feeds the missing source fact into the existing sleeping-session/cold-restore pipeline rather than adding an OMP-specific path:
- The generated Pi/OMP status handler now attaches `session_id` from the extension context (`ctx.sessionManager.getSessionId()`) to every posted snapshot — it must ride every event because the transport keeps only the latest pending post and a session can switch within a process.
- `omp` is added to `RESUMABLE_TUI_AGENTS`; `extractAgentProviderSession` parses its `session_id`; `getAgentResumeArgv` returns `['omp', '--resume', <id>]` (matching `omp -r/--resume`).

Warm daemon reattachment is unaffected — cold-restore consumption of the record is unchanged, so it still can't launch a duplicate agent (#5232/#5240 invariant).

Tests: OMP round-trips through extract/resume-argv; the generated hook payload includes `session_id` from the extension context (and omits it when absent).

Closes #8962

X: @mark86202384100


xianjianlf2:
Rebased onto current main (cb9c42d60) to adapt to the new session-metadata transport introduced by #8876: OMP now reuses the shared `sessionMetadata`/`updateSessionMetadata` mechanism (refreshed per event from ctx) instead of the previous per-handler injection. Since OMP resumes by id (`omp --resume <id>`) rather than by transcript file, no session-file persistence gate is applied — Pi's generated extension source is byte-for-byte unchanged. 482 tests across the touched suites pass; oxlint/tsc clean.

xianjianlf2:
Addressed this in `98da1532e` by decoupling `session_id` and `session_file` assignment in the generated extension source.

What changed:
- `sessionMetadata` is initialized first.
- `session_id` is added when a non-empty session id exists.
- `session_file` is added independently when a non-empty session file exists.
- Added regression coverage for Pi preserving a session file even when the session id is unavailable.

Local checks:
- pre-commit `oxlint` / `oxlint --config config/oxlint-react-doctor.json` / `oxfmt --write`
- `pnpm exec vitest run --config config/vitest.config.ts src/main/pi/agent-status-extension-source.test.ts` (30 tests passed)

revofusion:
Bumping @Jinwoo-H

OrcaWin (approved):
Reviewed the OMP cold-restore paths, custom session stores, WSL/SSH handling, persistence, mobile launch, and mixed-version fallback. Local validation and Electron reproduction passed; awaiting CI before merge.

PR #5519, opened 2026-06-16
Add project Windows runtime selection

slashdevcorpse:
## Summary

Adds a project-level Windows execution runtime model so each local Windows project resolves to exactly one runtime: Windows host or one selected WSL distro. The resolved runtime is now used across project-scoped preflight, agent and skill detection, terminal launch, git/worktree operations, hosted reviews, provider CLI calls, source-control AI helpers, file discovery/search, runtime APIs, and renderer command planning.

This also adds global default and per-project runtime settings, runtime-switch invalidation, repair states for unavailable WSL or missing distros, runtime-aware path/provider identity handling, user-facing docs, and regression coverage for stale WSL settings and Windows-host plus WSL side-by-side behavior.

## Review Focus

- Runtime Resolution Logic (`src/shared/project-execution-runtime.ts`): project overrides take precedence over the global default; `inherit-global` follows the global default; malformed or absent values fall back to Windows host; WSL repair states are returned instead of silently falling back when WSL is unavailable, missing a distro, or points at a missing distro.
- WSL Login Shell Routing (`src/main/git/runner.ts`, `src/shared/wsl-login-shell-command.ts`): WSL commands are routed through the selected distro with shell-escaped argv, login-shell resolution through the distro user's passwd entry, and dollar escaping across the outer `sh -lc` layer. Added regression coverage that command-scoped env like `HISTFILE=/tmp/orca-history` survives to the inner login shell.
- Worktree Path Comparison (`src/main/ipc/worktree-logic.ts`): Windows paths remain slash/case-insensitive for host worktrees; POSIX/WSL-owned paths remain case-sensitive even on a Windows desktop, so cleanup does not collapse distinct WSL paths.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — passed on Windows: 1707 files passed / 13 skipped; 17078 tests passed / 172 skipped.
- [x] `pnpm build` — passed the top-level build chain on Windows, including `typecheck`, `build:relay`, `build:cli`, `build:electron-vite`, `build:web`, and `build:native`.
- [x] Added or updated focused regression tests covering resolver precedence, repair states, stale WSL settings, terminal/PTY routing, preflight, skill/runtime setup, git/worktree operations, provider routing, source-control AI helpers, file discovery/search, renderer launch planning, and Windows host plus WSL smoke behavior.
[...]

PR #8992, opened 2026-07-16
fix(onboarding): stop offering WSL as a Windows default shell

xianjianlf2:
## Summary

On Windows onboarding, WSL is no longer offered as a fresh default-shell choice, so new profiles never persist a `wsl.exe` value that Settings can't display or edit.

## ELI5

Setup used to let Windows users pick WSL as their terminal, but the Settings screen can no longer show that choice — so it looked like PowerShell while actually launching WSL. Now setup stops offering WSL to new users, keeping what you pick and what you see in sync.

## Root cause & fix

`WindowsTerminalStep` gated its WSL card on `capabilities.wslAvailable`, making onboarding the only path that persisted `terminalWindowsShell: "wsl.exe"` — a value the Settings control dropped when #5519 moved WSL to the Project Runtime model. The fix narrows the gate to `windowsShell === 'wsl.exe'`, so the option only appears to reflect a value a prior build already saved, never as a fresh choice. This keeps the stored setting, the Settings UI, and the shell used for new panes consistent, without stranding existing WSL users.

## Evidence

- Test file: `src/renderer/src/components/onboarding/WindowsTerminalStep.test.tsx` — 4 passed on branch.
- Reverting the fix to `origin/main` fails the guard test `does not offer WSL as a fresh default-shell choice even when WSL is detected` (`AssertionError: expected html not to contain 'WSL'`), proving the fix is load-bearing.
- Lint clean: `oxlint` on `WindowsTerminalStep.tsx` reported no findings.

```
STEP 2a (branch):  Test Files 1 passed (1), Tests 4 passed (4)
STEP 2b (fix reverted):  Tests 1 failed | 3 passed
  Failing: "does not offer WSL as a fresh default-shell choice even when WSL is detected"
  AssertionError: expected html not to contain 'WSL' (line 57)
```

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression: the change only tightens one boolean gate in the onboarding WSL card. Users with a persisted `wsl.exe` value still see the option (verified by the still-passing sticky-WSL case), and all non-Windows / non-WSL paths are unaffected — proven by the 4-passing test suite and the targeted fail when the gate is reverted.

---

*Original fix by @xianjianlf2. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/renderer/src/components/onboarding/WindowsTerminalStep.test.tsx
+++ b/src/renderer/src/components/onboarding/WindowsTerminalStep.test.tsx
@@ -1,8 +1,26 @@
 import { renderToStaticMarkup } from 'react-dom/server'
-import { describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { WINDOWS_GIT_BASH_SHELL } from '../../../../shared/windows-terminal-shell'
 import type { GlobalSettings } from '../../../../shared/types'
+import type { WindowsTerminalCapabilities } from '@/lib/windows-terminal-capabilities'
 import { WindowsTerminalStep } from './WindowsTerminalStep'
+
+const mockCapabilities: WindowsTerminalCapabilities = {
+  wslAvailable: false,
+  wslDistros: [],
+  pwshAvailable: false,
+  gitBashAvailable: false,
+  hostPlatform: null,
+  isLoading: false
+}
+
+vi.mock('@/lib/windows-terminal-capabilities', () => ({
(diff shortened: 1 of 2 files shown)


Jinwoo-H:
Thanks @xianjianlf2 for the Windows onboarding fix. Current main now contains the same behavior in WindowsTerminalStep, with coverage ensuring WSL is not offered as the default shell; commit ee9eec461c records the fix and closes the related onboarding issue. We will retain your contribution as historical credit and request no rebase, retest, or other follow-up. Closing this superseded branch.

PR #8993, opened 2026-07-16
fix(terminal): stop native middle-click paste inserting twice on Linux

xianjianlf2:
## Summary

Fixes middle-click paste in the integrated terminal inserting the PRIMARY selection **twice** on Linux/X11 when "Middle-click Paste from Selection" is enabled.

## ELI5

On Linux, middle-clicking the terminal to paste used to type your copied text twice. Now it pastes exactly once, like you'd expect.

## Root cause & fix

The terminal reads the PRIMARY selection on mousedown and writes it to the PTY itself, calling `preventDefault()`. But Chromium's native X11 middle-click primary paste still fires on mouse **release**, landing in xterm's helper textarea, which forwards it to the PTY a second time. The global `usePrimarySelectionPaste` hook already suppresses this native follow-up, but only for editable targets it owns — it excludes xterm, leaving the terminal path unsuppressed. The fix arms a short shared suppression window (`armPrimarySelectionNativePasteSuppression`) when the terminal handles a middle-click, and has the hook's capture-phase `beforeinput`/`paste` suppressor honor it. The single native paste is swallowed; the selection reaches the PTY once. `Ctrl+Shift+V` and right-click paste (CLIPBOARD, no native primary-paste event) are unchanged.

## Evidence

- All unit tests pass on the branch across the armed-window state (arm / window / expiry / disable-clears) and the hook suppression (armed native paste suppressed for a non-owned xterm-style target, plus a negative case).

```
Test Files  2 passed (2)
     Tests  20 passed (20)
  Duration  222ms
```

- Oxlint clean (exit 0) on the changed source files.

## Trade-offs

None — pure fix.

## Regression risk

Low. The suppression only fires inside a briefly-armed window opened by a terminal middle-click; `Ctrl+Shift+V` and right-click paste take a different path and are unaffected. The new behavior is proven by the passing suppression and negative-case tests.

*Credit to original author @xianjianlf2. Validated and rebased onto latest main via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/renderer/src/components/terminal-pane/TerminalPane.tsx
+++ b/src/renderer/src/components/terminal-pane/TerminalPane.tsx
@@ -124,7 +124,11 @@ import {
   closeWebRuntimeTerminal,
   updateWebRuntimePaneLayout
 } from '@/runtime/web-runtime-session'
-import { isPrimarySelectionEnabled, readPrimarySelectionText } from '@/lib/primary-selection'
+import {
+  armPrimarySelectionNativePasteSuppression,
+  isPrimarySelectionEnabled,
+  readPrimarySelectionText
+} from '@/lib/primary-selection'
 import { APP_MENU_PASTE_EVENT } from '@/lib/app-menu-paste'
 import { WORKSPACE_FILE_PATH_MIME, WORKSPACE_FILE_PATHS_MIME } from '@/lib/workspace-file-drag'
 import { isTerminalSessionStateSaveFailure } from '../../../../shared/terminal-session-state-save-failure'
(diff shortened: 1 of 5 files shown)


nezdelion:
second paste comes from CLIPBOARD, not PRIMARY
to reproduce: select A -> press ctrl+C -> select B -> middle click -> get "BA"
wayland, fedora

xianjianlf2:
Thanks for the repro details. I checked this path and updated the wording in the branch/title to avoid making the fix sound X11/PRIMARY-only.

The implementation intentionally suppresses the short-lived native paste follow-up after Orca has already handled the terminal middle-click paste. That suppression is based on the armed middle-click window rather than the clipboard source, because the renderer paste/beforeinput events do not reliably expose whether the text came from PRIMARY or CLIPBOARD. So the Wayland/Fedora case where the follow-up inserts CLIPBOARD text should be covered too.

Verified locally with:

```sh
pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/hooks/usePrimarySelectionPaste.test.tsx src/renderer/src/lib/primary-selection.test.ts
```

Result: 2 files / 18 tests passed.

nwparker:
Thanks so much for this, @xianjianlf2 — stop native middle-click paste inserting twice on Linux. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #8994, opened 2026-07-16
fix(pty): probe Windows PATH registry via absolute reg.exe

xianjianlf2:
# fix(pty): probe Windows PATH registry via absolute reg.exe

## Summary
Fixes newly installed CLIs (e.g. Cursor's `agent`) not being recognized in the in-app Windows terminal because the registry PATH refresh silently failed when Electron launched without System32 on PATH.

## ELI5
On Windows, if you installed a new command-line tool, the app's terminal sometimes still said "command not found" even though a normal terminal could find it. This makes the app look in the right place every time, so the new tool works right away.

## Root cause & fix
`readPersistedWindowsPathSegments` spawned a bare `reg.exe` via `execFileSync` to refresh PATH from the Environment registry keys at spawn time. When Electron's main process inherits a PATH that omits System32 (Start Menu and service launches), libuv throws `ENOENT`, and the failure is swallowed by the surrounding `catch` — so the in-app terminal keeps a stale launch-time PATH.

The fix adds `getRegExePath(env)` in `src/main/win32-utils.ts`, returning `${SystemRoot ?? 'C:\\Windows'}\\System32\\reg.exe`, and `windows-environment-path.ts` now uses that absolute path, matching the existing `icacls` / `whoami` / `cmd` handling.

## Testing
- `src/main/pty/windows-environment-path.test.ts` — 1 file passed, 7 tests passed.
- Reverted implementation while keeping the test: expected failure, `reg.exe` was used instead of `C:\\Windows\\System32\\reg.exe`.
- `npx oxlint src/main/pty/windows-environment-path.ts src/main/win32-utils.ts`

## Screenshots
No visual change.

## AI Review Report
CodeRabbit reported no actionable code comments. Its only warning was that the PR description was missing required template sections; this update fills those sections.

## Security Audit
No new privilege boundary or user-controlled command path. The code resolves Windows' system `reg.exe` from `SystemRoot`, falling back to `C:\\Windows`, and only uses it for the existing local registry PATH read. Non-Windows behavior is unchanged.

## Notes
This overlaps the broader Windows PATH refresh area. If the other branch covers the same root cause, this PR can be treated as the minimal absolute-`reg.exe` fix or folded into that branch.

## Trade-offs
None expected. The behavior changes only when PATH is stripped enough that a bare `reg.exe` cannot be resolved.

## Regression risk
Low. The export is additive, the call site still invokes the same Windows registry tool, and the non-win32 early-return path is unchanged.

---
*Credit to original author @xianjianlf2. Validated, rebased, and hardened via automated bug-bash review; original fix design preserved.*


nwparker:
Ah yeah so I had another branch on this. Let me cross-ref it and get one or the other merged
But I think the general idea is right. I ran into it myself even!

xianjianlf2:
Makes sense. I filled in the missing PR template sections so the current branch is easier to compare against your other one. Happy for this to either land as the minimal absolute-`reg.exe` fix or get folded into the broader branch if that is cleaner.

xianjianlf2:
Resolved conflicts with latest main and pushed head d300394c5. Upstream now has equivalent absolute reg.exe helper behavior, so the conflict uses current main implementation while keeping the regression test coverage. Local validation: pnpm install --frozen-lockfile; pnpm exec vitest run --config config/vitest.config.ts src/main/pty/windows-environment-path.test.ts (20 passed); git diff --check on the clean worktree.

nwparker:
Superseded — thanks for this.

Main resolves `reg.exe` absolutely now, which is the hardening this PR was after:

- `src/main/win32-utils.ts:49` — `win32.join(root, 'System32', 'reg.exe')`
- `src/main/agent-hooks/managed-hook-owner-identity.ts:26` — `win32.join(process.env.SystemRoot ?? 'C:\\Windows', 'System32', 'reg.exe')`

so a `reg.exe` earlier on PATH can no longer be picked up. It landed as part of the wider Windows spawn-hardening work rather than as this diff.

Closing as superseded.

PR #9001, opened 2026-07-16
fix(orchestration): fail direct dispatches on agent exit

### LINKED ISSUE #8984: [Bug]: A failed dispatch never propagates to the task — dead agents leave tasks 'ready' forever and work is silently dropped
## Environment

- Orca macOS **1.4.137** (`com.stablyai.orca`), macOS 26.5.2 (25F84), Apple Silicon
- Local runtime healthy throughout (`orca status` → `runtimeState: ready`, `runtimeReachable: true`)
- Local dispatch, Claude Code workers
- Evidence below is from `~/Library/Application Support/orca/orchestration.db` (read-only queries)

## Summary

When a dispatched agent dies (`last_failure = "Agent exited with code -1"`), Orca records the failure on the **dispatch row** but **never propagates it to the task row**. The task stays `status='ready'` forever — indistinguishable in `orca orchestration task-list` from a task that simply hasn't started yet.

There is no retry, no re-queue, no escalation message, and no `circuit_broken` transition. **The work is silently dropped.**

In my case a 6-PR implementation campaign produced only 5 PRs. Nothing surfaced that PR 4 of 6 had been dispatched, run for 12 minutes, died, and been abandoned. I only discovered it days later by reading the DB by hand.

## Evidence

**1. Orca was alive and actively processing the failure** — it wrote `status` and `last_failure` onto the dispatch row, but wrote nothing to the task:

```
task               dispatch_wrote  dispatch_reason            task_left_at  task_result
-----------------  --------------  -------------------------  ------------  -------------------
task_b81cc199e541  failed          Agent exited with code -1  ready         (no result written)
```

So this is not "the app died before it could update" — it updated one table and not the other.

**2. It is systematic, across 5 separate days / independent campaigns** (not a single crash):

```sql
SELECT COUNT(*) FROM tasks t JOIN dispatch_contexts d ON d.task_id = t.id
WHERE t.status = 'ready' AND d.status = 'failed';
-- 25
```

```
day         orphaned tasks
2026-07-01   3
2026-07-03   8
2026-07-09   2
2026-07-10   2
2026-07-15  10
```

All 25 share `last_failure = 'Agent exited with code -1'`.

**3. Overall state divergence:**

```
task_status  dispatch_status     n
-----------  ------------------  ---
completed    completed           106     <- happy path is fine
ready        failed               25     <- the bug: silently abandoned
ready        (never dispatched)    5
dispatched   dispatched            3
failed       failed                2     <- only because a human ran task-update by hand
```

The only 2 tasks that ever reached `failed` have the **same** `"Agent exited with code -1"` reason but carry an explicit hand-written `result: {"reason":"cancelled: worker context-saturated"}`. i.e. failure reaches the task table **only** when a human types it in.

**4. The workers were healthy right up to death** — heartbeats and even `decision_gate` question/answer traffic, then the whole bus went silent at `12:41:51Z` and never emitted another message.

## Correlation: serial dispatch survived, parallel fan-out did not

Same campaign, same day, same machine:

- **00:09Z → 12:15Z** — 14 tasks dispatched **one at a time** (each after the previous completed) → **14/14 `completed`**.
- **12:16Z → 12:30:54Z** — coordinator fanned out **9 dispatches in ~100 seconds** → **9/9 died** with `"Agent exited with code -1"` within ~12 minutes.

Given the two hand-labelled cases say `worker context-saturated`, I suspect the workers exhausted context (each was pointed at a ~3,300-line spec) and exited. That part may be expected agent behaviour — **the bug I'm reporting is that the death is silently swallowed rather than surfaced.**

## Expected

When a dispatch fails, at minimum one of:

1. **Propagate** — set `tasks.status = 'failed'` (or back to `pending`/`ready` **with** a recorded `result`/attempt count) so the ledger reflects reality; and/or
2. **Retry / circuit-break** — `failure_count` stays stuck at `1` and the schema already has a `circuit_broken` state that never engages; and/or
3. **Escalate** — emit an `escalation` message so a supervising coordinator learns its worker died; and/or
4. **Surface** — `orca orchestration task-list` should visibly distinguish "ready (never dispatched)" from "ready (last dispatch died)". Today they render identically, which is what made the loss invisible.

A stale `last_heartbeat_at` is already recorded and would be a sufficient signal for a reaper.

## Actual

`tasks.status` stays `ready` indefinitely. `task-list` shows `[ready]`. Nothing indicates the agent ran and died. Work is dropped with zero signal.

## Reproduction

1. Create and dispatch a task to a worker agent.
2. Kill the agent (or let it exit non-zero / saturate context) so the dispatch records `Agent exited with code -1`.
3. `orca orchestration task-list` → the task is still `[ready]`.
4. Query the DB:

```sql
SELECT t.status AS task_status, d.status AS dispatch_status, d.failure_count, d.last_failure
FROM tasks t JOIN dispatch_contexts d ON d.task_id = t.id
WHERE d.status = 'failed';
```

→ `task_status='ready'` while `dispatch_status='failed'`.

## Impact

This is a **silent correctness/data-loss failure** for any multi-agent campaign. A fan-out of N tasks can return N−k results and the operator has no way to notice from the tooling — the missing tasks look identical to queued ones. It scales badly: the more parallel the fleet, the more likely a worker dies, and every death is invisible.

Related but distinct: #8698 (successful `worker_done` needs an explicit inbox poll to be seen). This one is the failure path, where the outcome is dropped entirely rather than merely delayed.

tianrking:
## Summary
- mark a task as `failed` when a directly dispatched worker exits and no coordinator run exists
- preserve the exit reason on the task instead of leaving it indistinguishable from an untouched `ready` task
- add a runtime regression test for the direct-dispatch path

## Root cause
`failDispatch` correctly returns a task to `ready` for coordinator-managed retries. A direct `orca orchestration dispatch` has no coordinator loop to consume that retry, so an agent exit could leave work silently stranded in `ready` forever.

## Impact
Coordinator-managed dispatches retain their existing retry and escalation behavior. Directly dispatched tasks now become visible failures that can be deliberately retried by the operator.

## Validation
- `node_modules/.bin/vitest.cmd run --config config/vitest.config.ts src/main/runtime/orca-runtime.test.ts --testNamePattern 'marks a direct dispatch failed'`
- `node_modules/.bin/oxfmt.cmd --check src/main/runtime/orca-runtime.ts src/main/runtime/orca-runtime.test.ts`
- `node_modules/.bin/oxlint.cmd src/main/runtime/orca-runtime.ts src/main/runtime/orca-runtime.test.ts`

The full runtime test file passed 710 tests; 4 unrelated tests assume a POSIX `/tmp` path and fail on this Windows host before exercising this change.

Fixes #8984

## ELI5

A directly dispatched orchestration worker that crashed left the task looking like it was still ready. Those tasks are now marked failed with the exit reason so work doesn’t sit forever.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -7315,9 +7315,40 @@ describe('OrcaRuntimeService', () => {
 
     runtime.onPtyExit('pty-1', -1)
 
-    expect(failDispatch).toHaveBeenCalledWith('ctx-1', 'Agent exited with code -1')
-    expect(updateTaskStatus).toHaveBeenCalledWith('task-1', 'failed', 'Agent exited with code -1')
+    expect(failDispatch).toHaveBeenCalledWith('ctx-1', 'Agent exited with code -1', {
+      requeueTask: false
+    })
+    expect(updateTaskStatus).not.toHaveBeenCalled()
     expect(insertMessage).not.toHaveBeenCalled()
+  })
+
+  it('keeps coordinator-owned dispatches retryable when their worker PTY exits', () => {
+    const runtime = createRuntime()
+    const workerHandle = runtime.preAllocateHandleForPty('pty-1')
+    const failDispatch = vi.fn()
+    const insertMessage = vi.fn()
+    runtime.setOrchestrationDb({
+      getActiveDispatchForTerminal: vi.fn((handle: string) =>
+        handle === workerHandle ? { id: 'ctx-1', task_id: 'task-1' } : undefined
+      ),
+      failDispatch,
+      getActiveCoordinatorRun: vi.fn(() => ({ coordinator_handle: 'term-coordinator' })),
+      insertMessage
+    } as never)
+    syncSinglePty(runtime)
+
+    runtime.onPtyExit('pty-1', 1)
+
+    expect(failDispatch).toHaveBeenCalledWith('ctx-1', 'Agent exited with code 1', {
+      requeueTask: true
+    })
+    expect(insertMessage).toHaveBeenCalledWith(
+      expect.objectContaining({
+        to: 'term-coordinator',
+        type: 'escalation',
+        priority: 'high'
+      })
+    )
   })
 
   it('keeps stale-title timers isolated per PTY', async () => {
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -7296,6 +7296,30 @@ describe('OrcaRuntimeService', () => {
     }
   })
 
+  it('marks a direct dispatch failed when its worker PTY exits', () => {
+    const runtime = createRuntime()
+    const workerHandle = runtime.preAllocateHandleForPty('pty-1')
+    const failDispatch = vi.fn()
+    const updateTaskStatus = vi.fn()
+    const insertMessage = vi.fn()
+    runtime.setOrchestrationDb({
+      getActiveDispatchForTerminal: vi.fn((handle: string) =>
+        handle === workerHandle ? { id: 'ctx-1', task_id: 'task-1' } : undefined
+      ),
+      failDispatch,
+      getActiveCoordinatorRun: vi.fn(() => undefined),
+      updateTaskStatus,
+      insertMessage
+    } as never)
+    syncSinglePty(runtime)
+
+    runtime.onPtyExit('pty-1', -1)
+
+    expect(failDispatch).toHaveBeenCalledWith('ctx-1', 'Agent exited with code -1')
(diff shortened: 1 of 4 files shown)


brennanb2025:
Taking a look!

brennanb2025:
Thanks for the clear diagnosis — I pushed a follow-up that keeps the intended behavior while making the dispatch/task failure transition atomic. Direct-dispatch worker exits now terminally fail the task with the exit reason, while coordinator-owned exits remain retryable and still escalate; I also corrected the lifecycle comment and added database/runtime coverage for both paths. The focused and full orchestration/runtime suites, Node typecheck, formatting, lint, and max-lines ratchet all pass.

PR #9003, opened 2026-07-16
fix(keybindings): resolve Mod+0 conflict for Focus worktree list (#8584)

### LINKED ISSUE #8584: [Bug]: Cmd/Ctrl+0 binding conflict makes Focus worktree list unreachable
Operating systems: macOS, Windows, Linux

Orca version: v1.4.138-rc.7

Details:

Short summary:

Two commands ship with the same default keybinding on every desktop platform:

- `zoom.reset` — “Reset Size” (`Mod+0`)
- `sidebar.focusWorktreeList` — “Focus worktree list” (`Mod+0`)

The main-process shortcut resolver checks `zoom.reset` first, consumes the event, and sends the zoom-reset action. As a result, the renderer never receives the chord for `sidebar.focusWorktreeList`, making the worktree-list focus command unreachable through its advertised default shortcut.

What happened?

Pressing `Cmd+0` on macOS or `Ctrl+0` on Windows/Linux resets Orca’s zoom. It does not focus the worktree list, even though Settings → Shortcuts lists the same chord for “Focus worktree list.” In a focused browser guest, the chord resets page zoom instead.

How can we reproduce it?

1. Open any workspace in Orca.
2. Open Settings → Shortcuts and confirm both “Reset Size” and “Focus worktree list” display `Cmd/Ctrl+0`.
3. Return to the workspace and focus a terminal, editor, or other normal app surface.
4. Press `Cmd+0` on macOS or `Ctrl+0` on Windows/Linux.
5. Zoom resets, but keyboard focus does not move to the worktree list.

Expected:

Both commands should be reachable through distinct default bindings, or one should intentionally ship unassigned rather than advertising a shortcut that can never fire.

Likely root cause:

`resolveWindowShortcutAction` resolves `zoom.reset` before renderer-level worktree-list handling. `createMainWindow` then calls `preventDefault()` and sends `terminal:zoom`, so the later renderer shortcut path cannot handle `sidebar.focusWorktreeList`.

Suggested direction:

Preserve the conventional `Cmd/Ctrl+0` zoom-reset behavior and give “Focus worktree list” a distinct, non-conflicting default (or leave it unassigned if no safe cross-platform chord exists). Add regression coverage at both the registry and main-window dispatch levels so future default collisions cannot silently make an action unreachable.

Related: #8533 identified the same class of default-binding conflict for Explorer and the macOS mobile emulator shortcut.

nwparker:
### Description

`zoom.reset` (“Reset Size”) and `sidebar.focusWorktreeList` (“Focus worktree list”) both shipped with default `Mod+0` on every platform. Main-process window shortcut handling resolves `zoom.reset` first, so Focus worktree list never received its advertised default.

This change:

- Keeps `zoom.reset` on browser-standard `Mod+0` (no zoom regression).
- Moves `sidebar.focusWorktreeList` default to free `Mod+Shift+0`.
- Adds registry/matcher regression coverage so the two actions stay distinct and Settings still flags a user remapping that reintroduces the collision.

Fixes #8584

### Evidence

```text
$ npx vitest run --config config/vitest.config.ts src/shared/keybindings.test.ts
 ✓ src/shared/keybindings.test.ts (61 tests) 14ms
 Test Files  1 passed (1)
      Tests  61 passed (61)
```

Defaults after fix:

- `zoom.reset` → `Mod+0` (darwin/linux/win32)
- `sidebar.focusWorktreeList` → `Mod+Shift+0` (darwin/linux/win32)

### ELI5

Two buttons claimed the same keyboard key (`Cmd/Ctrl+0`). Zoom always won, so “focus the worktree list” never worked. Zoom keeps the usual key; focusing the list now uses `Cmd/Ctrl+Shift+0`.

### User-regression-tradeoffs

Ideally zero for zoom users — `Mod+0` reset is unchanged.

Only users who expected Focus worktree list on `Mod+0` need the new default (`Mod+Shift+0`), or they can rebind it in Settings → Shortcuts. Anyone who already customized either shortcut is unaffected.

### DIFF
--- a/src/shared/keybindings.test.ts
+++ b/src/shared/keybindings.test.ts
@@ -242,6 +242,43 @@ describe('keybindings', () => {
     })
   })
 
+  it('keeps zoom reset on Mod+0 and focuses worktree list on a distinct chord', () => {
+    // Why: both actions previously defaulted to Mod+0, so main-process zoom
+    // reset always won and Focus worktree list was unreachable (#8584).
+    for (const platform of ['darwin', 'linux', 'win32'] as const) {
+      expect(getEffectiveKeybindingsForAction('zoom.reset', platform)).toEqual(['Mod+0'])
+      expect(getEffectiveKeybindingsForAction('sidebar.focusWorktreeList', platform)).toEqual([
+        'Mod+Shift+0'
+      ])
+    }
+
+    const zoomResetInput = {
+      key: '0',
+      code: 'Digit0',
+      meta: true,
+      control: false,
+      alt: false,
+      shift: false
+    }
+    const focusListInput = { ...zoomResetInput, shift: true }
+
+    expect(keybindingMatchesAction('zoom.reset', zoomResetInput, 'darwin')).toBe(true)
+    expect(keybindingMatchesAction('sidebar.focusWorktreeList', zoomResetInput, 'darwin')).toBe(
+      false
+    )
+    expect(keybindingMatchesAction('sidebar.focusWorktreeList', focusListInput, 'darwin')).toBe(
+      true
+    )
+    expect(keybindingMatchesAction('zoom.reset', focusListInput, 'darwin')).toBe(false)
+
+    expect(
+      findKeybindingConflicts('darwin', { 'sidebar.focusWorktreeList': ['Mod+0'] })
+    ).toContainEqual({
+      binding: 'Mod+0',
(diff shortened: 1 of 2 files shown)


nwparker:
## Agent handoff

**Fixes:** #8584

Full issue kit (summary + re-run + file links): see the **Agent handoff — reproduction kit** comment on #8584.

Writeup: https://github.com/stablyai/orca/blob/nwparker/bug-repro-handoffs/docs/bug-reproductions/8584.md

### This PR

PR #9005, opened 2026-07-16
fix(serve): prefer explicit --port over stale WS fallback (#8535)

### LINKED ISSUE #8535: [Bug]: explicitly pinned "orca serve --port <P>" is overridden by a stale persisted mobile-ws fallback port
### Operating system
Linux (reproduced on a headless Linux serve; the affected code path is cross-platform, not Linux-specific).

### Orca version
1.4.137 (affected). Control: 1.4.131 (not affected).

### Details

**Short summary**

An explicitly pinned `orca serve --port <P>` can silently bind a *different* port and never try `<P>`, because a previously-persisted "mobile WS fallback port" (`<userData>/mobile-ws-fallback-port.json`, STA-1511) is bound **before** the pinned port. Once that file names a stale port, every subsequent launch honors the stale port and strands every client dialing `<P>`. There is no CLI way to force the pinned port.

**What happened?**

We run a headless serve pinned to a fixed port (`orca serve --port 6768`) that already-paired desktop/remote clients dial. After the fallback file was written once (the one time `:6768` was briefly busy at bind), serve began binding the stale remembered port instead of `6768` on every launch — even when `6768` was provably free. Clients dialing `6768` were stranded; for us this also broke an auto-upgrade health-check that polls the pinned port, causing rollback. The bound port tracks the fallback file 1:1, independent of launch method.

**How can we reproduce it?**

On a provably-free `:6768`:

1. `printf '{"port":45175}' > <userData>/mobile-ws-fallback-port.json`
2. `orca serve --port 6768` → binds **:45175**, emits `orca_server_ready endpoint ws://0.0.0.0:45175` (the pinned `6768` is never attempted).
3. `rm <userData>/mobile-ws-fallback-port.json` → `orca serve --port 6768` → binds **:6768**.

**Root cause (verified against `main`)**

- `src/main/runtime/rpc/ws-transport.ts` `start()` (lines 161–167) builds `candidatePorts = persistedFallbackPort !== undefined ? [persistedFallbackPort, this.port] : [this.port]` and the loop returns on the first `tryListen` success — so a free fallback port pre-empts the pinned port. The intent is documented in the comment at lines 153–160 (the fallback is bound "FIRST … devices paired while it was active … would be permanently stranded" — STA-1511).
- `src/main/runtime/runtime-rpc.ts:715` passes `fallbackPort: readWsFallbackPort(this.userDataPath)` whenever a port was pinned (`wsPort !== 0`), and lines 798–800 persist the resolved port **whenever it diverges from the pinned port** (`if (this.wsPort !== 0 && wsTransport.resolvedPort !== this.wsPort) writeWsFallbackPort(...)`). It persists on *any* divergence — a transient `EADDRINUSE` is just the usual cause.
- `src/main/index.ts:2059` (in the `pnpm dev` port block) already notes the persisted fallback is re-bound in preference "until `mobile-ws-fallback-port.json` is removed from userData". Filename defined at `src/main/runtime/rpc/ws-fallback-port-store.ts:12`.

- Version bracket: the STA-1511 persistence code ships since **v1.4.130** (feature #7855). The fallback-first *binding order* was introduced by **#7924** ("binds the persisted fallback port first on launch, even if the preferred port is free"). Reproduced on **1.4.137**; **1.4.131** does not honor the poison file. 1.4.132–1.4.136 not bisected.
- Adjacent open work in the same file: **#7185** ("Fix WebSocket fallback for reserved Windows ports") touches `ws-transport.ts` but does not address the pinned-`--port` override.
- **Suggested fix:** when a port is set explicitly on the CLI, the pinned port should win. Note the "was it explicit?" signal is currently erased before the transport — `getServeOptions` leaves `wsPort` undefined when the flag is absent, but the `OrcaRuntimeRpcServer` constructor applies `wsPort = DEFAULT_WS_PORT` (6768) at `runtime-rpc.ts:439`, so an explicit `--port 6768` becomes indistinguishable from the default. A clean fix threads a boolean (e.g. `wsPortExplicit`, from `serveOptions.wsPort !== undefined` at `index.ts:2083`) through to the transport, then for an explicit pin either skips `readWsFallbackPort()` or orders `candidatePorts = [pinned, fallback]`. Alternatively add `--strict-port` / `--no-fallback-port`. STA-1511 persistence stays the default only for the unpinned/default-port case (`wsPort === 0` E2E already disables fallback). Happy to open a PR.
- Workaround for others hitting this: delete `<userData>/mobile-ws-fallback-port.json` before starting serve when you pin `--port`.

nwparker:
## Summary

Fixes #8535: when `orca serve --port <P>` is set explicitly, bind `<P>` before any port stored in `mobile-ws-fallback-port.json`. Default desktop / auto port paths keep STA-1511 fallback-first binding so existing mobile pairings stay stable.

## Description

`WebSocketTransport.start()` previously always tried `candidatePorts = [persistedFallbackPort, preferredPort]`. After a one-time `EADDRINUSE` on the preferred port, STA-1511 persistence wrote `mobile-ws-fallback-port.json`, and every later launch bound that stale free fallback **instead of** the CLI pin — stranding clients (and health checks) that dial `<P>`.

This change:

- Adds `preferPinnedPort` to `WebSocketTransport` (preferred first, fallback second when set).
- Plumbs `preferPinnedWsPort` from explicit serve CLI (`--serve-port` / `orca serve --port`) through `OrcaRuntimeRpcServer`.
- Leaves default 6768 / dev 6769 / GUI launches on fallback-first (STA-1511).
- Still falls through to the persisted fallback when the pin is busy, and still persists a new fallback when the resolved port diverges.

## Evidence

Focused vitest:

```text
pnpm exec vitest run src/main/runtime/rpc/ws-transport.test.ts src/main/runtime/rpc/ws-fallback-port-store.test.ts
✓ 27 passed (including new preferPinnedPort order tests)
```

New coverage:

- binds preferred first when `preferPinnedPort` is set and both ports free
- falls back when `preferPinnedPort` is set but preferred is taken
- existing STA-1511 “fallback even when preferred free” still asserts default order

## ELI5

If you say “always use door 6768”, Orca should open door 6768 when it’s free — not a sticky note from last time that said “use door 45175 instead”. Phone pairing that *didn’t* pin a door still trusts the sticky note first so old phones keep finding the house.

## User-regression-tradeoffs

| Path | Behavior after this PR |
| --- | --- |
| `orca serve --port <P>` | Pin first. Free pin always wins over stale fallback. If pin is busy, still try fallback then OS port. Clients dialing `<P>` get the pin when free. |
| Default desktop / no `--port` | Unchanged: fallback-first (STA-1511). Paired mobile devices that stored the fallback endpoint stay connected across restarts. |
| `pnpm dev` (6769) | Unchanged: still fallback-first. |

**Tradeoff (explicit pin only):** after a port conflict, mobile devices that only know the *old* fallback may need one re-pair / reconnect if serve later successfully binds the pin again (because we no longer prefer the fallback when free). That is intentional: an explicit pin means operators and health checks own that port. New pairing QR still advertises `resolvedPort`. Default auto path does **not** take this tradeoff.

## Screenshots

No visual change.

## Testing

- [x] Focused vitest: `ws-transport.test.ts`, `ws-fallback-port-store.test.ts` (27 passed)
- [ ] `pnpm lint` (pre-commit oxlint/oxfmt on touched files)
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`
- [x] Added/updated high-quality tests that would catch regressions

## AI Review Report

Reviewed for:
- Bind-order regression on STA-1511 (default path still fallback-first).
- Explicit-port distinguishability from `DEFAULT_WS_PORT` (flag, not port equality).
- Error fall-through: preferred non-`EADDRINUSE` still fatal; fallback failures still degrade.
- Cross-platform: listen/`EADDRINUSE`/`EACCES` path is Node HTTP server (macOS/Linux/Windows); no keyboard shortcuts, path separators, or shell changes.

No residual flags after plumbing + tests.

## Security Audit

No new auth, secrets, IPC, or command execution. Only TCP bind candidate order changes when an operator explicitly requests a port. Fallback persistence and E2EE/device-token auth unchanged.

## Notes

- Flag is set only when `serveOptions.wsPort !== undefined` (CLI passed `--serve-port`), not for default 6768 or dev 6769.
- Workaround for unfixed builds remains: delete `mobile-ws-fallback-port.json`.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2119,7 +2119,15 @@ app.whenReady().then(async () => {
     enableWebSocket: true,
     ...(isE2E ? { wsPort: 0 } : {}),
     ...(devWsPort !== undefined ? { wsPort: devWsPort } : {}),
-    ...(serveOptions?.wsPort !== undefined ? { wsPort: serveOptions.wsPort } : {}),
+    ...(serveOptions?.wsPort !== undefined
+      ? {
+          wsPort: serveOptions.wsPort,
+          // Why: only an explicit `orca serve --port` pin prefers the requested
+          // port over a stale STA-1511 fallback (issue #8535). Default 6768 /
+          // dev 6769 keep fallback-first so mobile pairings stay stable.
+          preferPinnedWsPort: true
+        }
+      : {}),
     webClientRoot: getBundledWebClientRoot()
   })
   registerMobileHandlers(runtimeRpc, { getRelayStatus: () => desktopRelayStatus })
--- a/src/main/runtime/rpc/ws-transport.test.ts
+++ b/src/main/runtime/rpc/ws-transport.test.ts
@@ -472,6 +472,43 @@ describe('WebSocketTransport', () => {
       expect(transport.resolvedPort).toBe(fallbackPort)
     })
 
+    it('binds the preferred port first when preferPinnedPort is set and both are free', async () => {
+      // Why: issue #8535 — `orca serve --port <P>` clients dial the pin. A
+      // free but stale mobile-ws-fallback-port.json must not pre-empt it.
+      const preferredPort = await reserveFreePort()
+      const fallbackPort = await reserveFreePort()
+
+      const transport = new WebSocketTransport({
+        host: '127.0.0.1',
+        port: preferredPort,
+        fallbackPort,
+        preferPinnedPort: true
+      })
+      transports.push(transport)
+      await transport.start()
+      expect(transport.resolvedPort).toBe(preferredPort)
+    })
+
+    it('falls back when preferPinnedPort is set but the preferred port is taken', async () => {
+      // Why: explicit pins still degrade to the STA-1511 fallback on
+      // EADDRINUSE so previously-paired mobile devices remain reachable.
+      const preferredHolder = new WebSocketTransport({ host: '127.0.0.1', port: 0 })
+      transports.push(preferredHolder)
+      await preferredHolder.start()
+      const preferredPort = preferredHolder.resolvedPort
+      const fallbackPort = await reserveFreePort()
+
+      const transport = new WebSocketTransport({
+        host: '127.0.0.1',
+        port: preferredPort,
+        fallbackPort,
+        preferPinnedPort: true
+      })
+      transports.push(transport)
+      await transport.start()
+      expect(transport.resolvedPort).toBe(fallbackPort)
+    })
+
     it('binds the preferred port when the persisted fallback is taken', async () => {
       const fallbackHolder = new WebSocketTransport({ host: '127.0.0.1', port: 0 })
       transports.push(fallbackHolder)
--- a/src/main/runtime/rpc/ws-transport.ts
+++ b/src/main/runtime/rpc/ws-transport.ts
@@ -54,6 +54,11 @@ export type WebSocketTransportOptions = {
   // the (now free) preferred port instead would strand those pairings
   // (STA-1511). Callers pass the previously assigned fallback port here.
   fallbackPort?: number
+  // Why: `orca serve --port <P>` clients dial the pinned port. Prefer that port
+  // first (fallback second) so a stale mobile-ws-fallback-port.json cannot
+  // silently steal the pin (issue #8535). Default auto/desktop keeps
+  // fallback-first for STA-1511 pairing stability.
+  preferPinnedPort?: boolean
 }
 
 export class WebSocketTransport implements RpcTransport {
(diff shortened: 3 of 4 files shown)


nwparker:
## Agent handoff

**Fixes:** #8535

Full issue kit (summary + re-run + file links): see the **Agent handoff — reproduction kit** comment on #8535.

Writeup: https://github.com/stablyai/orca/blob/nwparker/bug-repro-handoffs/docs/bug-reproductions/8535.md

### This PR

AmethystLiang:

PR #9007, opened 2026-07-16
fix(keybindings): resolve Cmd+Shift+E collision on macOS (#8533)

### LINKED ISSUE #8533: [Bug]: `Cmd+Shift+E` default binding conflict on macOS — "New mobile emulator tab" always wins, File Explorer toggle is unreachable
Operating system: macOS
Orca version: v1.4.137

Details:

Short summary:
On macOS, two commands ship with the same default keybinding `Cmd+Shift+E`:
- `tab.newSimulator` — "New mobile emulator tab" (macOS-only default)
- `sidebar.explorer.toggle` — "Show Explorer"

The simulator command always wins, so there is no working default shortcut to toggle the File Explorer on macOS. Settings → Shortcuts still lists `Cmd+Shift+E` for the Explorer, but it never fires.

What happened?
Pressing `Cmd+Shift+E` always opens a new Mobile Emulator tab. The File Explorer sidebar never toggles, despite Settings showing `Cmd+Shift+E` as its binding.

How can we reproduce it?
1. macOS, with the mobile emulator feature enabled (the default).
2. Press `Cmd+Shift+E`.
3. A new Mobile Emulator tab opens.
4. The File Explorer sidebar does not toggle (expected behavior per its Settings entry).

Expected:
Both commands should be reachable — either distinct default bindings, or a non-conflicting default for one of them on macOS. (Windows/Linux are unaffected because `tab.newSimulator` has no default binding there, so `Cmd/Ctrl+Shift+E` correctly toggles the Explorer.)

Likely root cause (from inspecting the packaged app.asar, v1.4.137):
- Both `tab.newSimulator` (darwin: `Mod+Shift+E`) and `sidebar.explorer.toggle` (`Mod+Shift+E`) are bound to the same chord.
- The main-process `before-input-event` handler matches `tab.newSimulator` (gated by `process.platform === "darwin" && isMobileEmulatorEnabled()`), sends `ui:newSimulatorTab`, then calls `event.preventDefault(); return true` — consuming the event.
- Because the main process consumes the key first, the renderer-level handler for `sidebar.explorer.toggle` never receives `Cmd+Shift+E` on macOS.

Workaround:
Remap one of them in Settings → Shortcuts — e.g., move the mobile emulator to `Cmd+Opt+Shift+E`, which frees `Cmd+Shift+E` and makes the File Explorer toggle work again.

Note: There may be other same-chord defaults worth auditing (e.g., `Cmd+0` is bound to both `zoom.reset` and `sidebar.focusWorktreeList`), though only `Cmd+Shift+E` produces a clearly wrong result in normal use.

nwparker:
### Description

On macOS, `tab.newSimulator` and `sidebar.explorer.toggle` both defaulted to `Mod+Shift+E`. Runtime handlers for the mobile emulator consumed the chord first, so **Show Explorer** never fired despite Settings listing `Cmd+Shift+E` for it.

This change keeps Explorer on `Mod+Shift+E` (VS Code muscle memory) and moves the macOS-only mobile emulator default to `Mod+Alt+Shift+E`. Linux/Windows remain unbound for the emulator (unchanged).

Fixes #8533

### Evidence

**Bug proven (defaults in source before fix):**
| Action | darwin default |
|--------|----------------|
| `sidebar.explorer.toggle` | `Mod+Shift+E` |
| `tab.newSimulator` | `Mod+Shift+E` |

`findKeybindingConflicts()` only reports conflicts involving *customized* overrides, so this default-vs-default collision was silent.

**After fix:**
| Action | darwin default |
|--------|----------------|
| `sidebar.explorer.toggle` | `Mod+Shift+E` |
| `tab.newSimulator` | `Mod+Alt+Shift+E` |

**Verified:**
```bash
pnpm exec vitest run src/shared/keybindings.test.ts
# ✓ 61 tests passed
```

Regression test asserts Explorer still matches `Mod+Shift+E`, emulator does not, and emulator matches `Mod+Alt+Shift+E` only on darwin.

### ELI5

Two buttons were sharing the same keyboard shortcut on Mac, so only one of them worked. We left the common Explorer shortcut alone and gave the mobile emulator its own slightly different shortcut.

### User-regression-tradeoffs

- **Explorer (`Cmd+Shift+E`)**: no change — preferred VS Code-compatible binding kept.
- **Mobile emulator (macOS only)**: default moves from `Cmd+Shift+E` → `Cmd+Option+Shift+E`. Users who memorized the old emulator default need to relearn or rebind in Settings. Preferable to remapping Explorer because emulator is newer/less used and macOS-only.

### DIFF
--- a/src/shared/keybindings.test.ts
+++ b/src/shared/keybindings.test.ts
@@ -717,6 +717,41 @@ describe('keybindings', () => {
     ).toBe(true)
   })
 
+  // Why: #8533 — both previously defaulted to Mod+Shift+E on darwin; emulator won.
+  it('keeps explorer on Mod+Shift+E and gives the mobile emulator a non-colliding macOS default', () => {
+    expect(getEffectiveKeybindingsForAction('sidebar.explorer.toggle', 'darwin')).toEqual([
+      'Mod+Shift+E'
+    ])
+    expect(getEffectiveKeybindingsForAction('tab.newSimulator', 'darwin')).toEqual([
+      'Mod+Alt+Shift+E'
+    ])
+    expect(getEffectiveKeybindingsForAction('tab.newSimulator', 'linux')).toEqual([])
+    expect(getEffectiveKeybindingsForAction('tab.newSimulator', 'win32')).toEqual([])
+    expect(formatKeybindingList(['Mod+Alt+Shift+E'], 'darwin')).toBe('⌘⌥⇧E')
+
+    expect(
+      keybindingMatchesAction(
+        'sidebar.explorer.toggle',
+        { key: 'e', code: 'KeyE', meta: true, control: false, alt: false, shift: true },
+        'darwin'
+      )
+    ).toBe(true)
+    expect(
+      keybindingMatchesAction(
+        'tab.newSimulator',
+        { key: 'e', code: 'KeyE', meta: true, control: false, alt: false, shift: true },
+        'darwin'
+      )
+    ).toBe(false)
+    expect(
+      keybindingMatchesAction(
+        'tab.newSimulator',
+        { key: 'e', code: 'KeyE', meta: true, control: false, alt: true, shift: true },
+        'darwin'
+      )
+    ).toBe(true)
+  })
+
   it('defines an unassigned per-agent tab action for every TUI agent', () => {
     for (const agent of ALL_TUI_AGENTS) {
       const actionId = agentTabActionId(agent)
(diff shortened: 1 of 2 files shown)


nwparker:
## Agent handoff

**Fixes:** #8533

Full issue kit (summary + re-run + file links): see the **Agent handoff — reproduction kit** comment on #8533.

Writeup: https://github.com/stablyai/orca/blob/nwparker/bug-repro-handoffs/docs/bug-reproductions/8533.md

### This PR

Issue #8970, opened 2026-07-16
[Bug]: Closed agent session still lingers in sidebar on v1.4.143 (regression after #8825 / #8851)


jh01010-rsquare:
### Operating system

macOS

### Orca version

v1.4.143 (macOS, Darwin 25.5.0)

## Summary

After closing an agent's terminal tab with Ctrl+W, its row stays in the left
sidebar's per-project agent list. The row should disappear when the terminal is
closed, but it persists. A full app restart (Cmd+Q + relaunch) is the only thing
that clears it.

## Repro

1. In a project with a single `main` worktree, run a Claude Code agent in a
   terminal (e.g. a conversation titled "안녕 - 안녕하세요").
2. Close that terminal with Ctrl+W.
3. Expected: the agent row is removed from the left sidebar list.
4. Actual: the row remains and the "N agents" count stays inflated.
5. Restarting Orca clears it; nothing short of a restart does.

## Relation to #8851 / #8825

The v1.4.143 release notes list #8825 ("reap finished Claude named
agents/teammates from the sidebar roster", the fix that closed #8851) as
included in this build — yet the stale-row behavior still reproduces here.

This looks like either an incomplete fix (my case is a normal top-level Claude
Code conversation closed via Ctrl+W, not a named teammate/subagent) or a
regression. The restart-only workaround matches the roster-state class described
in #8851.

Made with [Orca](https://github.com/stablyai/orca) 🐋

Chang-Jin-Lee:
Hi! I'd like to work on this. I can reproduce the general shape of the roster-lifecycle bug and will look at how #8825's reaping logic interacts with a plain top-level Claude Code conversation closed via Ctrl+W (not a named teammate/subagent), since that's the gap the reporter points at. I'll start by writing a regression test around the sidebar agent roster state before implementing a fix.

Chang-Jin-Lee:
Update: after digging in, this touches the same agent-status/roster state machinery as #8825/#8851/#8522 (hook-driven, orchestration-aware, retained-row logic across src/shared/claude-subagent-roster.ts and src/renderer/.../worktree-agent-rows.ts) — more involved than I have bandwidth for as a first contribution. Stepping back so it's free for someone else.

nwparker:
## 🤖 Agent handoff — reproduction kit

**Status:** PARTIAL · GitHub label (partial — no label)  
[...]

PR #9008, opened 2026-07-16
fix(sidebar): prevent closed-tab agent rows from returning

craftowen:
## Summary

Prevents a cleanly completed agent row from being resurrected in the sidebar after its terminal tab has already been closed.

## ELI5

When you finished an agent and closed its tab, a stale "done" card could pop back into the list on its own. This makes the closed card stay gone.

## Root cause & fix

A race: PTY cleanup removes the live agent row *before* `closeTab` plants its durable retention suppressor, so the next retention effect saw a completed row with no suppressor yet and re-retained it as a stale "done" card. The fix threads `recentlyClosedAgentStatusTabIds` into `collectRetainedAgentsOnDisappear`, which now skips re-retaining any done row whose tab id is marked recently-closed. It is an additive guard param with an early `continue`, so existing retention paths are untouched.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase was clean (base `7ab601487c`, head `300618a0bb`).

- On branch: `useRetainedAgents.test.ts` — **Test Files 1 passed (1); Tests 7 passed (7)**.
- Reverting only the fix file (`useRetainedAgents.ts`) to `origin/main` while keeping the new test: **2 failed | 5 passed (7)**. Failing assertions:
  - `collectRetainedAgentsOnDisappear > does not retain a done row after its tab closed without a live suppressor` — `toRetain` expected `[]`, got 1 entry.
  - `useRetainedAgentsSync > does not re-retain when status removal and tab closure commit before the next retention effect` — `retainedAgentsByPaneKey[paneKey]` expected `undefined`.

```
On branch:  Test Files 1 passed (1) | Tests 7 passed (7)
Fix reverted: Test Files 1 failed (1) | Tests 2 failed | 5 passed (7)
```

- Lint: `oxlint` on the changed file — clean, no warnings/errors.

## Trade-offs

None — pure fix.

## Regression risk

Effectively zero. The change is an additive parameter plus an early `continue` that fires only on the buggy branch (done row whose tab is recently-closed); all other retention paths are byte-identical, as proven by the 5 unchanged tests that still pass with the fix reverted. Pure renderer store logic — no cross-platform, SSH, or Git concerns.

*Original fix by @craftowen (addresses the Ctrl+W repro in #8970). Validated, rebased, and reviewed via automated bug-bash triage; original fix design preserved.*

### DIFF
--- a/src/renderer/src/components/dashboard/useRetainedAgents.test.ts
+++ b/src/renderer/src/components/dashboard/useRetainedAgents.test.ts
@@ -1,6 +1,53 @@
-import { describe, expect, it } from 'vitest'
+// @vitest-environment happy-dom
+import { act, renderHook } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { useAppStore } from '@/store'
 import type { AgentStatusEntry, AgentStatusState } from '../../../../shared/agent-status-types'
-import { collectRetainedAgentsOnDisappear } from './useRetainedAgents'
+import { makePaneKey } from '../../../../shared/stable-pane-id'
+import type { Repo, Worktree } from '../../../../shared/types'
+import { collectRetainedAgentsOnDisappear, useRetainedAgentsSync } from './useRetainedAgents'
+
+const initialAppState = useAppStore.getInitialState()
+
+beforeEach(() => {
+  useAppStore.setState(initialAppState, true)
+})
+
+afterEach(() => {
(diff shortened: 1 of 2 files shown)


nwparker:
Heads-up @craftowen: this got auto-closed but the underlying bug (#8970) was never actually fixed on `main`, so I relanded your fix as #10334 (merged) with full credit to you. Thanks again! 🙏

PR #9010, opened 2026-07-16
fix(wsl): set WSL_UTF8 on remaining wsl.exe execFile spawns

### LINKED ISSUE #8475: Chinese text sometimes garbled in Codex on Windows (WSL)
## Description

Chinese text appears as garbled characters when running Orca commands (e.g., `orca-ide computer get-app-state`) in WSL on Windows. The output shows mojibake instead of the original Chinese characters.

Image

## Steps to Reproduce

1. Run an Orca command in WSL that outputs Chinese text (e.g., `orca-ide computer get-app-state --app Orca --window-id <id> --json`)
2. Observe the output in the terminal

## Expected Behavior

Chinese characters should display correctly.

## Actual Behavior

Chinese characters are garbled/mojibake. Example output:

## Environment

- **OS:** Windows with WSL (Ubuntu)
- **Shell:** CMD.EXE inside WSL
- **Orca:** latest

## Context

Reported by user in WeChat group. The issue appears specific to Windows/WSL environments where non-ASCII (Chinese) characters are not being encoded/decoded properly between the Orca process and the terminal output.

Chang-Jin-Lee:
## Summary

- `wsl.exe` re-encodes a spawned process's output for the host console's active codepage unless `WSL_UTF8=1` is set; without it, non-ASCII text (e.g. Chinese) comes through as mojibake/NUL-riddled garbage.
- This codebase already has the fix in one call site (`src/main/agent-hooks/wsl-hook-relay-launch.ts`), documented with a `Why:` comment. Two sibling `wsl.exe` spawn sites in the Codex WSL integration were missing it: the session bridge's `execFileUtf8` helper (`src/main/codex/wsl-codex-session-bridge.ts`) and the canonical Linux path resolver (`src/main/codex/codex-wsl-hook-install-plan.ts`). Both now set `WSL_UTF8: '1'` for consistency with the established pattern.
- Added a note to the `orca-cli` skill guide: running `orca-ide` inside WSL from a bare Windows `cmd.exe` console (not an Orca-managed terminal) can still show garbled output, since that console's codepage is outside Orca's control — `chcp 65001` (or Windows Terminal) fixes it.

Closes #8475

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`
- [x] Added regression tests for both call sites (`wsl-codex-session-bridge.test.ts`, `codex-wsl-hook-install-plan.test.ts`) asserting `options.env.WSL_UTF8 === '1'`. Verified they fail without the fix (`git stash` on the two source files) and pass with it.

Note on `pnpm test`: the full suite has 30 pre-existing failing test files on this Windows dev machine (e.g. `spawn /bin/sh ENOENT`, PTY-exit timeouts under this sandbox) that are unrelated to this change — confirmed identical by re-running the same 30 files against the base commit with this diff stashed out; the failure set was byte-for-byte the same before and after.

## AI Review Report

Reviewed with an AI coding agent for cross-platform impact: this only touches Windows-side `wsl.exe` invocation (Node's `child_process.execFile` options), so there is no macOS/Linux code path affected — on those platforms `wsl.exe` is never spawned. No shortcuts, labels, or path-separator logic touched. SSH/remote worktrees are unaffected since this is local WSL-distro tooling, not SSH transport. No other supported agent/CLI-provider integration touches these two call sites.

## Security Audit

The change only adds an environment variable (`WSL_UTF8: '1'`) to two existing `execFile('wsl.exe', ...)` calls; it does not alter argv construction, add new IPC surfaces, subprocesses, dependencies, or credential/path handling. No new user-controlled input reaches these calls beyond what already existed.

## Notes

- Windows + WSL only; no-op on macOS/Linux.
- The literal repro in #8475 (`cmd.exe` piping `wsl.exe` output outside any Orca-managed terminal) is a console Orca never spawns or owns, so it's addressed via a skill-guide doc note (`chcp 65001`) rather than app code.

## ELI5

Some WSL commands mangled non-English text into garbage because UTF-8 wasn’t forced. The remaining `wsl.exe` call sites now set WSL_UTF8 so Chinese and other text stay readable.


nwparker:
Landed in #16007 — thank you, and sorry it sat this long.

You spotted that `WSL_UTF8` was missing on several `wsl.exe` spawns, so wsl.exe's own error messages arrived as UTF-16LE and surfaced as NUL-riddled text. It is now set unconditionally in the runner, and also on the guest-environment probe — which spawns wsl.exe itself and had the same gap.

I carried the fix rather than merging this branch directly because it went in alongside a wider rewrite of how Orca invokes `wsl.exe` (a single chokepoint that always sets the flag, so it cannot be forgotten at a new call site). The credit is recorded in the commit message and in the code comment at the fix.

Closing as superseded.

Issue #9024, opened 2026-07-16
Deleting a project leaves an orphaned "unknown" workspace that returns after restart


PannenetsF:
## Summary

Deleting a project (repo) leaves an orphaned workspace entry that renders as **"unknown"** in the sidebar. The stale entry survives a restart: even after manually removing it from `worktreeMeta`, the daemon rebuilds it on the next launch because dangling references remain in `workspaceSession`.

## Environment

- Orca `1.4.141`, Linux (headless `orca serve`, `--port 6768`)
- Reproducible on any project that had at least one worktree/terminal session before deletion.

## Steps to reproduce

1. Add a git project and open a worktree/terminal in it (so it gets recorded in `workspaceSession`).
2. Remove the project via Settings → Repositories → **Remove Project** (or `orca` CLI project removal).
3. Observe a residual **"unknown"** workspace entry in the sidebar.
4. Restart the runtime — the "unknown" entry comes back even if you hand-edit it out of `worktreeMeta` in `orca-data.json`.

## Root cause (traced in source)

Project removal only prunes a subset of the persisted state. In `src/main/persistence.ts`:

- `removeProject(id)` → `pruneWorktreeStateForRepo(id, null)` cleans **only** `worktreeMeta`, `worktreeLineageById`, and `workspaceLineageByChildKey` (persistence.ts ~L4340-4389).
- It **never** clears the removed repo's worktrees from `workspaceSession` (the legacy blob) or `workspaceSessionsByHostId` (host-scoped copies).

There is already a helper that does exactly this cleanup for a single owner key — `removeWorkspaceSessionOwner(session, ownerKey)` (persistence.ts ~L2446) — which deletes the owner's `lastVisitedAtByWorktreeId`, `sleepingAgentSessionsByPaneKey`, tab state, `activeWorktreeId`, etc. It is called from `removeFolderWorkspace` and `deleteProjectGroup`, but **not** from the project-removal path.

Because `workspaceSession.lastVisitedAtByWorktreeId` (and sleeping agent sessions) still reference `${repoId}::${path}`, the runtime treats those worktrees as recently-active on startup and re-materializes `worktreeMeta` for them. Since the parent repo/project no longer exists, the UI has a worktree with no owning project → renders as **"unknown"**.

I confirmed this on a live instance: after deleting a project, `orca-data.json` still contained the dead `repoId` in:
- `worktreeMeta` (rebuilt)
- `workspaceSession.lastVisitedAtByWorktreeId`
[...]

PR #9025, opened 2026-07-16
fix: prune workspace session state when a project is removed (#9024)

PannenetsF:
## Summary

Deleting a project no longer leaves an orphaned **"unknown"** workspace that reappears in the sidebar after every restart (#9024).

**ELI5:** When you removed a project, the app forgot to fully forget it — a ghost copy kept coming back each time you reopened the app. Now it's cleaned up completely.

**Root cause & fix:** `removeProject` → `pruneWorktreeStateForRepo` only cleaned `worktreeMeta` and the lineage maps, but never removed the deleted repo's worktrees from the workspace session state (the legacy `workspaceSession` blob and the per-host `workspaceSessionsByHostId` partitions). Because `lastVisitedAtByWorktreeId` still referenced the gone worktrees, the next launch treated them as recently-active and re-materialized `worktreeMeta`, rendering them as an orphaned "unknown" workspace.

The fix collects the repo's owner keys (`${repoId}::${path}`) before the `worktreeMeta` delete loop — from `worktreeMeta`, the legacy session, and every host partition — so host classification can still read `meta.hostId` (including orphan keys present only in the session maps). It then runs the existing session-owner cleanup against the legacy/local session and every matching host partition, reusing the path already used by `removeFolderWorkspace` and `deleteProjectGroup`. Host gating is preserved: a host-scoped removal (`hostId != null`) only prunes that host's partition; a full removal (`hostId === null`) clears every host, so a shared repo id/path on a surviving host keeps its session.

This session also added a behavior-preserving refactor: `removeWorkspaceSessionOwner` is split into `deleteOwnerKeyedSessionFields` (O(1) owner-keyed deletes) and `deleteScannedSessionFieldsForOwners` (single pass over the pane-key maps and the shutdown list), plus a new batch `removeWorkspaceSessionOwners`. This drops the per-owner `structuredClone` (previously O(worktrees × hosts)) to one clone + one scan per session partition.

Original fix by @PannenetsF; validated, rebased onto latest main during 2026-07-23 bug-bash triage, and optimized as described above.

## Screenshots

No visual change. This PR only touches main-process persistence logic (`src/main/persistence.ts` and its tests); no renderer UI is affected. The user-observable behavior change is the absence of the orphaned "unknown" sidebar workspace after restart, which cannot be captured as a static screenshot here.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [ ] `pnpm build`
- [x] Added or updated high-quality tests for the change, or explained why not.

Author-reported (I cannot run pnpm in this environment, so these reflect the original body's claims, and the commands run were scoped to the changed file rather than the full `pnpm` scripts):

- Lint: `oxlint src/main/persistence.ts` exit 0, no `max-lines` disable added.
- Types: `tsc --noEmit -p config/tsconfig.node.json` exit 0.
- Tests: `npx vitest run --config config/vitest.config.ts src/main/persistence.test.ts` → **407 passed**, including the 2 `removeProject` session-prune regression tests and the 3 host-scoped `removeProjectForHost` prune tests. The 2 regression tests were confirmed to **fail on `main`** (stale keys survive) and **pass with the fix**.
- `pnpm build` was not reported as run.

## AI Review Report

Reviewed the full diff for correctness and edge cases:

- **Owner-key collection ordering** is correct: keys are gathered from `worktreeMeta`, the legacy session, and all host partitions *before* the `worktreeMeta` delete loop, so orphan keys that exist only in session maps (and would otherwise lose their `meta.hostId` classification) are still captured. This is the crux of the fix.
- **Host gating is sound.** `pruneLegacyLocalSession` fires for `hostId === null` or the local host id; `pruneAllHostPartitions` only for `hostId === null`; otherwise partitions are skipped unless `partitionHostId === hostId`. This prevents wiping a surviving host's session for a shared repo id/path — directly covered by the three new `removeProjectForHost` tests.
- **Refactor is behavior-preserving.** The split of `removeWorkspaceSessionOwner` threads `removedTabIds` through the O(1) deletes into the scan phase, matching the prior single-owner behavior; the `isRemovedOwner` predicate generalizes the old `worktreeId === ownerKey` check to set membership for the batch path. The `terminalPtyIncarnationsByPaneKey` `lastIndexOf(':')` / `separator < 1` guard is carried over unchanged.
- **Edge cases:** empty owner-key set and undefined sessions short-circuit in `removeWorkspaceSessionOwners` (returns the original reference), preserving prior semantics. `advanceTerminalTopologyRevision` is only forwarded on the single-owner path, matching pre-existing behavior (the batch project-removal path did not advance it before, either).
- **Cross-platform (macOS/Linux/Windows):** confirmed checked. No platform-dependent code is touched — no keyboard shortcuts, shortcut labels, filesystem path construction, shell invocation, or Electron main/renderer boundary changes. Owner keys are opaque `${repoId}::${path}` strings compared by equality/prefix; no path separator assumptions are introduced. Nothing here diverges by OS.

No fabricated results; the above is based solely on the diff.

## Security Audit

No security-relevant surface is changed. The diff is pure in-memory state pruning:

- No new input handling, deserialization of untrusted data, or parsing.
- No command/shell execution.
- No filesystem path resolution or traversal (owner keys are compared as opaque strings, never resolved to disk paths).
- No auth, secrets, credentials, or network calls.
- No dependency changes.
- No IPC handlers added or modified.

The change only removes stale entries from existing persisted session structures, which is a strict cleanup with no widening of trust boundaries.

## Notes

- No platform-specific behavior introduced; SSH/remote hosts are explicitly handled via the per-host partition gating (`workspaceSessionsByHostId`).
- Follow-up: `pnpm build` and the full `pnpm lint` / `pnpm typecheck` / `pnpm test` suites should be confirmed green in CI, since the author's local runs were scoped to the changed file.

---

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/nwparker/orca/workspaces/orca/bug-bash-07-23/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/nwparker/orca/workspaces/orca/bug-bash-07-23/node_modules
--- a/src/main/persistence.test.ts
+++ b/src/main/persistence.test.ts
@@ -3618,6 +3618,58 @@ describe('Store', () => {
     expect(store.getWorkspaceSession('ssh:ssh-a').lastVisitedAtByWorktreeId?.['shared::/repo']).toBe(222)
   })
 
+  it('removeProjectForHost prunes only the removed host when a third host also shares the owner key', async () => {
+    const store = await createStore()
+    // Same repo id + path on local and two SSH hosts, so the owner key
+    // `shared::/repo` is identical across all three. Removing one non-local host
+    // must prune only that host's partition and leave both the local session and
+    // the other surviving SSH host intact.
+    store.addRepo(makeRepo({ id: 'shared', path: '/repo' }))
+    store.addRepo(
+      makeRepo({
+        id: 'shared',
+        path: '/repo',
+        connectionId: 'ssh-a',
+        executionHostId: 'ssh:ssh-a'
+      })
+    )
+    store.addRepo(
+      makeRepo({
+        id: 'shared',
+        path: '/repo',
+        connectionId: 'ssh-b',
+        executionHostId: 'ssh:ssh-b'
+      })
+    )
+    store.setWorktreeMeta('shared::/repo', { displayName: 'local', hostId: 'local' })
+
+    store.setWorkspaceSession({
+      ...getDefaultWorkspaceSession(),
+      lastVisitedAtByWorktreeId: { 'shared::/repo': 111 }
+    })
+    store.setWorkspaceSession(
+      {
+        ...getDefaultWorkspaceSession(),
+        lastVisitedAtByWorktreeId: { 'shared::/repo': 222 }
+      },
+      'ssh:ssh-a'
+    )
+    store.setWorkspaceSession(
+      {
+        ...getDefaultWorkspaceSession(),
+        lastVisitedAtByWorktreeId: { 'shared::/repo': 333 }
+      },
+      'ssh:ssh-b'
+    )
+
+    store.removeProjectForHost('shared', 'ssh:ssh-a')
+
+    // Only the removed host's partition is pruned; local and the other SSH host survive.
+    expect(store.getWorkspaceSession('ssh:ssh-a').lastVisitedAtByWorktreeId?.['shared::/repo']).toBeUndefined()
+    expect(store.getWorkspaceSession().lastVisitedAtByWorktreeId?.['shared::/repo']).toBe(111)
+    expect(store.getWorkspaceSession('ssh:ssh-b').lastVisitedAtByWorktreeId?.['shared::/repo']).toBe(333)
+  })
+
   it('reorderReposForHost independently reorders local and SSH rows with shared ids', async () => {
     const store = await createStore()
     store.addRepo(makeRepo({ id: 'shared', path: '/local/shared' }))
--- a/src/main/persistence.test.ts
+++ b/src/main/persistence.test.ts
@@ -3552,6 +3552,72 @@ describe('Store', () => {
     expect(store.getWorktreeMeta('shared::/remote/repo/wt')).toBeUndefined()
   })
 
+  it('removeProjectForHost keeps the surviving host session for a shared repo id + path', async () => {
+    const store = await createStore()
+    // Same repo id AND same path on both local and an SSH host, so the owner key
+    // `shared::/repo` is identical across hosts. The host-scoped prune must only
+    // touch the removed host's session partition.
+    store.addRepo(makeRepo({ id: 'shared', path: '/repo' }))
+    store.addRepo(
+      makeRepo({
+        id: 'shared',
+        path: '/repo',
+        connectionId: 'ssh-a',
+        executionHostId: 'ssh:ssh-a'
+      })
+    )
+    store.setWorktreeMeta('shared::/repo', { displayName: 'local', hostId: 'local' })
+
+    store.setWorkspaceSession({
+      ...getDefaultWorkspaceSession(),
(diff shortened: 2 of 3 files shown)


e-jung:
Thanks for tackling #9024. I tested the current head and confirmed that the full `removeProject` path fixes the orphaned-workspace problem: the new tests pass with this change and fail against the parent commit.

I found one cross-host regression in the host-scoped `removeProjectForHost` path. Owner keys do not include the host ID, but the new cleanup applies each matching key to the local session and every `workspaceSessionsByHostId` partition.

Reproduction: register the same repo ID and path on local and `ssh:ssh-a`, with local `lastVisitedAt=111` and SSH `lastVisitedAt=222`. After `removeProjectForHost("shared", "ssh:ssh-a")`, the surviving local value is unexpectedly removed instead of remaining `111`. This can also discard the surviving host's tabs, sleeping-agent state, and active-worktree pointer.

Could the cleanup be scoped to the selected host when `hostId` is non-null, while retaining the all-partitions cleanup for full `removeProject`? A regression test using the same owner key on two hosts should cover the distinction.

Otherwise, the fix looks focused and the full-removal behavior appears correct.

PannenetsF:
@e-jung Thanks for the careful review and the concrete repro — you're right, that was a real cross-host bug.

I pushed `53e7f18` to scope the host-partition cleanup to the removed host:

- When `hostId` is non-null, cleanup now only touches that host's `workspaceSessionsByHostId` partition and skips the others, so a surviving host sharing the same repo id + path keeps its `lastVisitedAt`, tabs, sleeping-agent state, and active-worktree pointer.
- The legacy/local session blob is only pruned when `hostId === null || hostId === LOCAL_EXECUTION_HOST_ID`.
- Full `removeProject` (`hostId === null`) still clears every partition, as before.

I also added the two regression tests you suggested, both using the same owner key on two hosts:

- `removeProjectForHost keeps the surviving host session for a shared repo id + path` — removing `ssh:ssh-a` keeps local `111`.
- `removeProjectForHost on the local host keeps a surviving SSH host session` — removing local keeps SSH `222`.

Would you mind taking another look when you have a moment? Thanks again.

e-jung (approved):
Re-reviewed commit 53e7f18. The host-scoped cleanup now preserves surviving local, SSH, and paired-remote session partitions while full remove still clears every partition.

I verified both new regression tests fail against the previously reviewed head 4b05fa0 and pass on the current head. The complete persistence suite passes 375 of 375 tests, the touched files pass node TypeScript checking and oxlint, and an additional two-SSH-host reproduction also passes.

The earlier cross-host blocker is resolved. The remaining observations are non-blocking: worktreeMeta still has a pre-existing identical-path cross-host ambiguity, and a third-surviving-host fixture could be useful future coverage. Approving this revision. Thanks for the careful fix.

PannenetsF:
Sure. I will add a /3rd host/ test case to cover it :)

PannenetsF:
Thanks for the thorough review @e-jung 🙏 This is approved and green (the host-scoped cleanup regression you caught is resolved, and the persistence suite passes). I don't have merge access on the upstream — could you (or another maintainer) squash-merge this when you have a moment? Happy to rebase if it's gone stale.

PannenetsF:
Hi @e-jung — `main` moved and this picked up a conflict, so I merged latest `main` in and resolved it. The conflict was only in the test file: main and this branch each appended a `removeProject` test at the same spot (session-state/per-host pruning vs. the new topology-authority case) — I kept all three, they're independent. `persistence.ts` merged cleanly. Full persistence suite passes (400/400) and node typecheck is clean. Since the update likely dismissed your earlier approval, could you re-approve and squash-merge when you have a moment? Thanks again for the careful review!

nwparker:
Thanks so much for this, @PannenetsF — prune workspace session state when a project is removed (#9024). Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #9026, opened 2026-07-16
feat(windows): add native Windows ARM64 package support

### LINKED ISSUE #8607: [Feature]: Windows on ARM native build
### Problem or use case

Currently there is only a x64 distribution. Will be nice to support ARM natively.

It works fine through emulation, just not ideal.

### Proposed solution

Compile ARM variant as well for Windows distribution.

_No response_

jinhyunchun:
## Summary

Adds a native Windows ARM64 package. Until now Orca shipped only an x64 Windows installer, so Windows-on-ARM devices (Snapdragon X laptops, ARM VMs) run the entire app under x64 emulation. This PR makes `pnpm build:win:arm64` produce a native `orca-windows-arm64-setup.exe` while leaving the existing x64 build path byte-for-byte unchanged.

**How the port handles each native dependency:**

- **node-pty (terminal)** — the packaging hook stages node-pty's published `win32-arm64` prebuild into `build/Release` instead of source-rebuilding, so no cross-architecture MSVC toolchain is required. Every staged binary's PE header is verified against the target machine type (`0xaa64`) before packaging; a wrong-architecture file fails the build rather than shipping a broken terminal.
- **sherpa-onnx (offline speech)** — upstream publishes no Windows ARM64 addon, so local dictation is gated off on `win32/arm64`: the model catalog, model states, downloads, and dictation start all flow through one `speech-platform-support` module. Cloud (OpenAI) transcription keeps working. A persisted local model selection (e.g. settings migrated from an x64 install) is reconciled at startup, and attempts surface a stable `voice_local_dictation_unavailable` error code mapped to a localized toast with an "Open Settings" action.
- **agent-browser (browser automation)** — upstream publishes no win32-arm64 executable but supports running its x64 binary under Windows on ARM emulation; the runtime resolver and packaging share that fallback, pinned together by a test.
- **Packaging correctness** — arch-specific resources and the installer artifact name key off `ORCA_WINDOWS_ARM64_BUILD`, and `validateWindowsPackageArchitecture` fails the build loudly if the env flag and electron-builder's target arch ever disagree (e.g. a bare `electron-builder --win --arm64` invocation), so a mislabeled or x64-contaminated ARM64 package cannot be produced silently.

## Screenshots
image
image

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

All checks were run natively on Windows 11 ARM64 (Node 24.18.0, pnpm 10.24.0) — i.e., on the target hardware of this port. Full-suite context on that host: every test file this PR touches passes (speech suite 82/82; `orca-runtime.test.ts` 701/701; all new packaging suites). The suite has ~37 files with pre-existing failures on a Windows ARM64 host — I verified they fail identically on the base commit (hardcoded `/tmp` paths, symlink `EPERM` without developer mode, Linux-only shims), so they are host-environment issues, not regressions. This PR actually *improves* the ARM64-host baseline: 20 `stt-service.test.ts` tests that fail on `main` there (the sherpa native package cannot resolve) now pass via an explicit platform pin/mock.

New/updated tests:

- `config/scripts/windows-node-pty-prebuild.test.mjs` — PE-machine verification accepts a real ARM64 prebuild and rejects an x64 one; staging replaces stale host build output.
- `config/scripts/windows-package-architecture.test.mjs` — arch-specific resources/artifact naming, plus a cross-pinning test asserting the packaged agent-browser filename equals what the runtime resolver requests.
- `config/scripts/electron-builder-native-rebuild.test.mjs` — staging fires for win32/arm64 regardless of host arch; env-flag/target-arch mismatches are rejected.
- `src/main/speech/speech-platform-support.test.ts`, `src/main/ipc/speech.test.ts`, `src/main/speech/model-manager.test.ts` — catalog/state filtering, settings reconciliation, stable error code.
- `src/main/browser/agent-browser-binary-name.test.ts`, `src/renderer/src/components/dictation/dictation-start-error-toast.test.ts` — x64 fallback naming and the remediation toast.
- `src/main/speech/stt-service.test.ts`, `src/main/speech/model-manager*.test.ts` — made host-independent (platform pinned/mocked) so they pass on Windows ARM64 dev machines too; `ModelManager.downloadModel` gained injectable platform/arch parameters consistent with `getModelStates`.

## Notes

- **CI/release is intentionally out of scope.** `release-cut.yml` has no Windows ARM64 matrix entry; this PR only adds the local build path. Wiring CI needs maintainer decisions: a `windows-11-arm` matrix entry, parametrizing the many `dist/win-unpacked` / `orca-windows-setup.exe` hardcoded steps, a second SignPath approval flow, and — the real design question — an updater-manifest strategy, since both Windows jobs would clobber `latest.yml` today. Suggest shipping ARM64 outside the auto-update channel first.
- **Deliberate behavior change**: a persisted `sttModel` that no longer exists in the catalog is now cleared at startup on *all* platforms (previously it lingered and dictation failed opaquely).
- **ARM64 Windows hosts always use the verified node-pty prebuild** rather than a source rebuild (no MSVC requirement); x64 hosts keep the existing source-rebuild path.
- **Future upstream flips**: when sherpa-onnx or agent-browser ship native win32-arm64 binaries, the gate/fallback should be reverted; the sites are cross-referenced in comments (`speech-platform-support.ts`, `agent-browser-binary-name.ts`, `windows-package-architecture.cjs`).

## Issues
Fixes #8607

## ELI5

Orca now ships a native Windows ARM64 installer for Snapdragon and other ARM PCs. Those machines no longer need to run the whole app under x64 emulation.


jinhyunchun:
@Jinwoo-H Hi, when you have a chance, could you please take a look at this PR?
 Please let me know if you need any additional context. Thanks!

PR #8578, opened 2026-07-13
Fix CJK IME newline: defer Shift/Ctrl+Enter until composition commits

RickyPark:
## Summary

In the terminal pane, pressing **Shift+Enter** (or **Ctrl+Enter**) while a CJK IME composition is still open — a Korean syllable mid-composition, or highlighted Japanese conversion text — pushed the in-progress character down a line instead of inserting a clean newline after it. The window-level terminal shortcut handler resolves the newline on `keydown`, which fires *before* `compositionend`, so the CSI-u newline was written to the PTY ahead of the pending commit and the composed glyph landed after the newline. This now yields "committed character + newline", matching a non-composing Shift+Enter (and matching cmux's behavior).

## Screenshots

No visual change to the app UI. The fix changes byte ordering sent to the PTY for a CJK IME + Shift/Ctrl+Enter keystroke. Before: composing a Hangul syllable and pressing Shift+Enter left the syllable pushed below a blank line. After: the syllable commits in place and a single newline follows.

## Testing

- [ ] `pnpm lint` (ran `oxlint` on changed files — no new findings; one pre-existing `exhaustive-deps` warning on `panePtyBindingsRef.current` is unrelated and present on `main`)
- [x] `pnpm typecheck` (`tsc --noEmit -p config/tsconfig.tc.web.json` — 0 errors)
- [x] `pnpm test` (ran the affected suites: `terminal-ime-deferred-newline`, `terminal-shortcut-policy`, `keyboard-handlers`, `terminal-windows-shift-enter`, `terminal-ime-composition-tracker`, `xterm-bypass-policy` — all pass)
- [ ] `pnpm build`
- [x] Added tests: `terminal-ime-deferred-newline.test.ts` covers ordering (send deferred one macrotask past `compositionend`), the no-`compositionend` fallback, single-fire + listener cleanup, and the no-element path.

## AI Review Report

[...]

PR #9039, opened 2026-07-16
fix(terminal): commit the Hangul syllable before the Shift/Ctrl+Enter newline

### LINKED ISSUE #8038: [Bug]: Desktop terminal drops the last Hangul syllable onto a new line on Enter (missing isComposing guard, macOS)
## Summary

On the **desktop** terminal (xterm.js), pressing **Enter** while a Hangul (Korean) syllable is still being composed causes the **trailing syllable to be pushed onto a new line** instead of being committed inline before the newline. This looks like a missing `isComposing` / `keyCode === 229` guard on the terminal's Enter handling — the same class of bug already fixed for other input surfaces in #6980 (rename inputs) and #742 (workflow popup), but still present in the desktop terminal itself.

This is **desktop-specific** and distinct from the mobile Korean IME issues (#7183 / #7182 / #7000), which live in `mobile/src/terminal/*` (React Native). Here the affected surface is the xterm.js terminal in the desktop app.

## Environment

- **Orca**: v1.4.131 (desktop, macOS)
- **OS**: macOS (Darwin 25.5.0), Apple Silicon
- **IME**: 2-Set Korean (두벌식) — the default macOS Korean input source
- **Surface**: desktop terminal pane (xterm.js). English/ASCII input is unaffected.

## Steps to reproduce

1. Switch the macOS input source to Korean (2-Set / 두벌식).
2. Focus a terminal pane in the desktop app.
3. Type a few Korean syllables, e.g. `하 하 하`, so the **last** syllable is still in IME composition (marked/underlined, not yet committed).
4. Press **Enter**.

## Expected

The full composed text is committed inline, then a single newline is sent — i.e. Enter should first flush the pending composition (`compositionend`), then submit. The trailing syllable stays on the same line as the rest of the command.

## Actual

The trailing syllable is separated onto a new line. Visually:

```
> 하 하
  하▮        ← last syllable leaked onto the next line after Enter
```

The `\r` (newline) appears to be sent while the IME composition is still active, so the `compositionend` commit of the last syllable races/lands after the newline instead of before it.

## Root cause hypothesis

The terminal's key handling does not suppress/defer Enter while `KeyboardEvent.isComposing` is true (equivalently `keyCode === 229` during composition). It should not emit the carriage return until the pending composition has been committed. This is the identical failure mode described in:

- **#6980** — "rename inputs commit on Japanese IME Enter confirmation (missing isComposing guard)"
- **#742** — "Workflow creation popup submits on Japanese IME Enter confirmation"

Those were fixed by guarding Enter on `isComposing`; the desktop terminal path appears to have been missed. (Also related in spirit: #6905 Vietnamese IME in terminal.)

## Workaround (for other users hitting this)

Commit the composition **before** pressing Enter using anything that isn't Enter — e.g. finish typing, tap **한/영 (Caps Lock)** to switch the input source to English (which commits the pending syllable), then press Enter. Enter then fires with no active composition and submits cleanly.

## Notes

- A screenshot reproducing the `> 하 하` / `하` split is available and can be attached on request.
- Repro is 100% consistent whenever the final character before Enter is an uncommitted Hangul syllable.

sanghunka:
## Summary

Fixes #8038.

Shift/Ctrl+Enter during Hangul composition drops the trailing syllable below the newline, the exact symptom in #8038's screenshot. Includes #8578's commit unchanged (thanks @RickyPark) and adds the missing half.

**What #8578 misses**: macOS Hangul delivers a committing Enter **twice**. Real-keyboard trace (3/3 presses):

```
keydown #1  keyCode=229, isComposing=true   -> #8578 defers (correct)
compositionend                              -> syllable commits
keydown #2  keyCode=13,  isComposing=false  -> re-dispatch, passes the guard,
                                               sends newline ahead of the glyph flush;
                                               the deferred newline then doubles it
```

**Fix**: track deferred newlines per pane; an Enter `sendInput` arriving while one is still in flight is the same physical keystroke and is dropped. The window is a few ms (200 ms worst case), so real repeats and key-repeat are unaffected; IMEs that never re-dispatch see no behavior change.

Changes:

- `terminal-ime-deferred-newline.ts`: add `createTerminalImeDeferredNewlineSender()`, a per-pane in-flight counter around #8578's `sendTerminalInputAfterComposition`.
- `keyboard-handlers.ts`: defer through the sender; drop Enter `sendInput` while the pane's deferred newline is pending.
- `terminal-ime-deferred-newline.test.ts`: 5 unit tests for the sender lifecycle.
- `tests/e2e/korean-ime-terminal-shift-enter-commit.spec.ts`: new e2e regression spec.

## Screenshots

https://github.com/user-attachments/assets/77ac6669-b306-4c10-abe3-493944852415

Typing `하 하 하` with the last syllable still composing, then pressing Shift+Enter:

- **Left**: Orca v1.4.143 (latest release). The trailing syllable drops below the newline.
- **Right**: this PR's build. The syllable commits inline and exactly one newline follows.

No visual change otherwise; only PTY byte ordering changes.

**Test environment**

- macOS 26.5.1 (Darwin 25.5.0), Apple Silicon
- IME: 2-set Korean (macOS default)
- Electron 43

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` (one pre-existing environmental failure, `local-pty-shell-ready.test.ts`, reproduces on pristine main on this machine)
- [x] `pnpm build`
- [x] Added or updated high-quality tests that would catch regressions: the e2e spec replays the recorded two-keydown shape against a PTY harness and asserts exactly one newline, delivered after the committed syllable. It **fails on #8578 alone** and passes with this fix. The burst is dispatched without per-command awaits; awaited round-trips let the glyph flush win the race and hide the bug. Also verified with a real 2-set Korean IME on the built app: `하 하 하` + Shift+Enter stays inline with one newline; plain Enter / Ctrl+Enter / Space+Enter flows verified in the same session.

## AI Review Report

Reviewed with an AI coding agent. Main risks checked: absorbing a genuine second Enter (impossible: the pending window is a few ms and key-repeat only fires outside composition); lost newline (one-shot guard + 200 ms fallback retained from #8578). The review explicitly checked cross-platform compatibility for macOS, Linux, and Windows: the dedupe requires an active IME composition to arm, byte encodings are untouched, and the Windows Shift+Enter encoding path is unaffected. SSH/remote/local: sends go through the same `PtyTransport` abstraction for all hosts; only timing changes, and the transport is re-read at send time as in #8578. Agents/integrations: KKP TUIs like Codex still receive their CSI-u encodings unchanged. Performance: one Map lookup per Enter keydown, nothing on other keys. No UI surface touched.

## Security Audit

No new IPC, command execution, path handling, auth, secrets, or dependency surface; the change only reorders and dedupes an already-authorized terminal input sequence relative to the IME composition commit. No follow-up needed.

## Notes

- The re-dispatch shape was observed and verified on macOS (Darwin 25.5.0, Apple Silicon, 2-set Korean, Electron 43). On platforms whose IMEs never re-dispatch, the dedupe never arms, so behavior is unchanged; CI exercises the unit and e2e suites cross-platform.
- Supersedes #8578 (its commit included unchanged); happy to fold into #8578 instead if @RickyPark prefers.
- On #8038: it is filed as plain Enter, but plain Enter does not reproduce on identical hardware; xterm already orders the re-dispatched plain-Enter shape correctly. The Shift+Enter gesture reproduces the issue's screenshot exactly (continuation line, no submission, cursor after the dropped syllable), so this PR fixes the reported behavior.

X handle: [@sanghunkan](https://x.com/sanghunkan)

## ELI5

Typing Korean Hangul and pressing Shift/Ctrl+Enter could drop the last syllable onto the wrong line. The terminal now finishes the syllable correctly before inserting the newline.

### DIFF
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers.ts
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers.ts
@@ -8,6 +8,7 @@ import type { PtyTransport } from './pty-transport'
 import { safeFind } from '../terminal-search-safe-find'
 import { resolveTerminalShortcutAction } from './terminal-shortcut-policy'
 import type { MacOptionAsAlt } from './terminal-shortcut-policy'
+import { sendTerminalInputAfterComposition } from './terminal-ime-deferred-newline'
 import {
   keybindingMatchesAction,
   type KeybindingOverrides,
@@ -399,24 +400,40 @@ export function useTerminalKeyboardShortcuts({
       }
 
       if (action.type === 'sendInput') {
+        // preventDefault here does not cancel a pending IME commit (the
+        // compositionend still fires and xterm forwards the glyph), so it is safe
+        // to consume the chord even mid-composition — it just stops a stray
+        // browser newline from also reaching the helper textarea.
         e.preventDefault()
         e.stopImmediatePropagation()
         const pane = manager.getActivePane() ?? manager.getPanes()[0]
         if (!pane) {
           return
         }
-        const sent = paneTransportsRef.current.get(pane.id)?.sendInput(action.data) === true
-        if (sent) {
-          recordTerminalUserInputForLeaf(tabId, pane.leafId)
-          if (action.data === '\x1b[13;2u') {
-            // Why: this direct shortcut write does not pass through PTY onData,
-            // so no-OSC shells need an explicit post-write confirmation ladder.
-            const binding = panePtyBindingsRef.current.get(pane.id) as
-              | (IDisposable & { requestDroidReconfirmation?: () => void })
-              | undefined
-            binding?.requestDroidReconfirmation?.()
+        const sendResolvedInput = (): void => {
+          const sent = paneTransportsRef.current.get(pane.id)?.sendInput(action.data) === true
+          if (sent) {
+            recordTerminalUserInputForLeaf(tabId, pane.leafId)
+            if (action.data === '\x1b[13;2u') {
+              // Why: this direct shortcut write does not pass through PTY onData,
+              // so no-OSC shells need an explicit post-write confirmation ladder.
+              const binding = panePtyBindingsRef.current.get(pane.id) as
+                | (IDisposable & { requestDroidReconfirmation?: () => void })
+                | undefined
+              binding?.requestDroidReconfirmation?.()
+            }
           }
         }
+        // Why: Shift+Enter / Ctrl+Enter pressed while an IME composition is open
+        // is the committing keystroke — sending the newline now races ahead of
+        // the pending commit and pushes the composed CJK glyph below it. Forward
+        // the newline once the composition ends so the glyph lands first (the
+        // committed char + newline, matching a non-composing Shift+Enter).
+        if (e.isComposing && e.key === 'Enter') {
+          sendTerminalInputAfterComposition(pane.terminal.element, sendResolvedInput)
+          return
+        }
+        sendResolvedInput()
         return
       }
 
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers.ts
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers.ts
@@ -434,10 +434,11 @@ export function useTerminalKeyboardShortcuts({
           deferredNewlineSender.defer(pane.id, pane.terminal.element, sendResolvedInput)
           return
         }
-        if (e.key === 'Enter' && deferredNewlineSender.isDeferredNewlinePending(pane.id)) {
+        if (e.key === 'Enter' && deferredNewlineSender.absorbRedispatchedEnter(pane.id)) {
           // Why: macOS Hangul re-dispatches the committing Enter keydown with
           // isComposing already false ~2ms after compositionend; it is the same
-          // physical keystroke as the deferred newline still in flight.
+          // physical keystroke as the deferred newline, and can land on either
+          // side of that send's macrotask.
           return
         }
         sendResolvedInput()
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers.ts
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers.ts
@@ -8,7 +8,7 @@ import type { PtyTransport } from './pty-transport'
 import { safeFind } from '../terminal-search-safe-find'
 import { resolveTerminalShortcutAction } from './terminal-shortcut-policy'
 import type { MacOptionAsAlt } from './terminal-shortcut-policy'
-import { sendTerminalInputAfterComposition } from './terminal-ime-deferred-newline'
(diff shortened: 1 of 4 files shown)


RickyPark (approved):
Approve. Nice catch on the twice-dispatched committing Enter; that's the half #8578 missed, and thanks for keeping my commit in.

Two non-blocking things, both fine as follow-ups:

1. The absorb sits inside the `sendInput` branch, so it only catches a re-dispatch that still carries its modifier. If one ever comes through as a bare Enter (no Shift/Ctrl), `resolveTerminalShortcutAction` returns null, the handler falls through, and xterm sends a plain CR (a stray submit) while the deferred newline still fires. On macOS 2-set Korean the modifier is preserved, and your trace + e2e back that up, so this is really a guard for if that invariant ever breaks. Might be worth absorbing terminal-scoped Enter before shortcut resolution, or just noting the modifier-preserved assumption next to the absorb.

2. Unused absorb credits only clear when a later modified Enter calls `absorbRedispatchedEnter`. On IMEs that never re-dispatch, `defer()` leaves a credit sitting around (and a later `defer()` resets the deadline), so a genuine modified Enter in a later window is a bit easier to drop. A small expiry timer would keep the map clean.

Neither blocks merge; the reported bug is fixed and verified. LGTM.

OrcaWin:
Thank you for capturing the real macOS 2-set Korean event shape and for the source-built smoke evidence. I replayed that shape against the submitted head `7664dec37be8f9715b86eb40c975f06ef943224b` and then attacked the identity/routing assumptions.

The submitted implementation is not safe to land:

- Redispatch ownership expires by wall clock after 50 ms. Chromium preserves the original `{code,timeStamp}` identity across the composing keydown, keyup, and a delayed redispatch; an 80 ms same-keystroke redispatch therefore falls outside this window and sends a second newline.
- Keyup may arrive before that delayed redispatch, so keyup ordering is not a causal boundary.
- Both keydowns resolve `manager.getActivePane()` independently, and the deferred callback re-reads the pane transport/binding. A pane switch or PTY rebind between composition end and dispatch can route one physical gesture to a different pane or replacement PTY.
- The submitted E2E covers the immediate recorded order, but not delayed same-timestamp redispatch, pane/leaf changes, or PTY replacement; those cases are red on this head.

PR #11052 is now the owning replacement. Its hardened candidate retains the useful defer-until-composition-commit behavior and contributor context, but uses Chromium causal key identity and pins the pane/leaf, transport, PTY, and binding for the gesture. The same six-case source-built Electron oracle is red with this implementation disabled and green with that candidate, including keyup-before-delayed-redispatch and route changes.

Closing this as superseded/unsafe rather than asking for another timing-window revision. The event trace and reproduction work here were valuable and directly informed the replacement.

PR #9043, opened 2026-07-16
fix(terminal): replay cold restores at recovered grid

slashdevcorpse:
## Summary

- Preserve disk cold-restore `cols`/`rows` across the daemon provider, IPC/preload, renderer, and main headless terminal.
- Replay cold ANSI at its recovered grid without forwarding that temporary resize to the fresh PTY.
- Clear stale viewport cells at the destination grid (`CSI 2J` + home), await parsing before any recovered-grid shrink can promote clipped rows into scrollback, continue to forbid `CSI 3J`, blank enough source/immediate-destination rows, then use the existing structural replay barrier and authoritative fit.
- Keep older dimensionless cold-restore payloads compatible.

This is a cold-restore parity follow-up found while investigating #8751. Runtime evidence indicates the screenshot's immediate warm-reattach race is separately addressed by #8674; this PR does not claim that the cold path alone produced the screenshot or replace #8674.

## Screenshots

No visual design change. The failure screenshot and investigation are in #8751.

## Testing

- [ ] `pnpm lint` — full repository lint not run; touched `oxlint` and `oxfmt --check` passed
- [x] `pnpm typecheck`
- [ ] `pnpm test` — full repository suite not run; affected 5-file matrix passed with 904 tests / 15 skipped
- [ ] `pnpm build` — full production/native build not run; the Electron E2E bundle built successfully

Commands run:

- `pnpm test -- src/main/daemon/daemon-pty-adapter.test.ts src/main/ipc/pty.test.ts src/renderer/src/components/terminal-pane/pty-connection.test.ts src/renderer/src/lib/pane-manager/terminal-structural-replay-coordinator.test.ts src/renderer/src/components/terminal-pane/pty-transport.test.ts`
- `pnpm typecheck`
- `pnpm exec electron-vite build --mode e2e`
- touched-file `pnpm exec oxfmt --check ...`
- touched-file `pnpm exec oxlint ...`
- `git diff --check`

The initial regression was red in three focused assertions: missing adapter dimensions, `undefined` runtime seed dimensions, and no renderer resize to the recovered grid. A strengthened dirty-grid regression then caught a blocker in the first implementation: shrinking from a dirty 10x5 destination to the recovered 20x3 grid before clearing promoted clipped stale rows into scrollback, beyond a later viewport-only clear. Its clear-before-resize ordering assertion failed before the correction and passes after the clear parse barrier. The final five-file matrix passed.

### Same-PC Windows Electron verification

The exact PR-head app bundle (`877a3c28c04b54037f983c8b25029b7fd95e2cbe`) was exercised with disposable user data on the affected Windows PC:

- clean close/relaunch retained the same daemon PID/session and the restored pane accepted keyboard-path and direct PTY input;
- daemon-snapshot replay restored the cursor to a shell prompt rather than an older output row;
- killing the disposable daemon between launches forced disk cold restore, created the replacement PTY, and left the completed restore interactive in isolated repeats.

On the final amended bundle, both warm checks passed and all 3 repeated cold runs that reached the input assertion passed. One serial run began the delayed keyboard probe during the intentional replay guard: PowerShell received only the suffix `RED_OK`, while direct PTY input already worked before `pty:listSessions`. Independent trace and code reviews classified this as an E2E replay-readiness race, not a persistent transport or ownership failure. Two further repetitions ended earlier during app setup when navigation destroyed the Playwright execution context; neither reached the cold-input assertion.

The committed E2E drivers assume POSIX `Ctrl+U`/`export PS1`, and the cold case is normally skipped on Windows, so only the temporary test driver was adapted; the built application stayed at the exact PR head and all driver edits were restored afterward. The packaged installer/update harness was not run because it correctly refuses while an unrelated Orca app process is active.

## AI Review Report

Parallel investigation covered the issue/release history, local Windows runtime chronology, restore-path code tracing, test-first implementation, and independent adversarial diff, external-state, and Windows-runtime reviews.

Main risks checked:

- A recovered-height shrink can promote dirty destination rows into scrollback; the patch clears and awaits xterm parsing before that structural resize, and the regression preserves pre-existing scrollback sentinels.
- A full clear would regress #7310 by erasing restored scrollback; the patch uses viewport-only `CSI 2J` and tests that `CSI 3J` is absent.
- The incident's immediate warm-return mechanism belongs to unreleased #8674; PR and issue wording were narrowed to the independently proven cold-restore gap.
- Cold dimensions remain optional at public boundaries so older payloads retain their current fallback.
- Synthetic source-grid resizes are suppressed from PTY forwarding; only the post-replay destination fit reaches the PTY.
- The main headless emulator uses cold dimensions rather than accidentally borrowing warm snapshot fields.
- The headless oracle starts from adversarial dirty state and proves stale-content absence, recovered cursor spacing, scrollback retention, reflow, and a clean final viewport.
- Existing delayed-parse tests cover the shared replay barrier, fit/SIGWINCH gating, and live-byte ordering used by snapshot, replay, and cold restore.

Cross-platform review explicitly covered macOS, Linux, Windows/ConPTY, SSH relay, and remote-runtime behavior. No shortcut, label, path, shell-command, Git-provider, or local-only assumption was added. SSH relay restore remains on its existing replay branch; optional cold metadata is transport-neutral.

## Security Audit

No dependency, command execution, filesystem path, auth, secret, network, or new IPC-channel surface was added. The existing internal cold-restore result gains two numeric metadata fields sourced from recovered terminal history. Renderer use requires finite positive dimensions before resizing; legacy/missing values fall back safely. Replay remains under the existing query-response guard so recorded terminal queries cannot write replies into the fresh shell.

## Notes

- #8674 (`69c14fadce`) merged after v1.4.143 and is absent from that release. Newly published [v1.4.144-rc.0](https://github.com/stablyai/orca/releases/tag/v1.4.144-rc.0) contains #8674, but not this cold-restore follow-up.
- Completed Windows Electron warm restoration and all 3 cold repeats that reached input passed on the exact PR head; one serial keyboard probe raced replay readiness rather than finding a persistent freeze. v1.4.144-rc.0 can now verify the warm fix, while a future packaged build containing this PR is still required for combined warm/cold release verification before #8751 is closed.
- GitHub's ordinary PR, Wayland GPU, and computer-use workflows currently await maintainer approval for this fork PR; only `track-community-pr` has run and passed so far.
- Detailed investigation comment: https://github.com/stablyai/orca/issues/8751#issuecomment-4994737327
- Companion Windows restart-regression PR: #9064. It hardens the PowerShell/replay-readiness drivers and adds the non-skippable Windows gate. The PRs are independently mergeable, but #9043 and #9064 should land together; merge #9064 first, then synchronize or mark #9043 ready so the new workflow runs.

### DIFF
--- a/src/main/daemon/daemon-pty-adapter.test.ts
+++ b/src/main/daemon/daemon-pty-adapter.test.ts
@@ -1296,6 +1296,7 @@ describe('DaemonPtyAdapter (IPtyProvider)', () => {
       expect(result.coldRestore).toBeDefined()
       expect(result.coldRestore!.scrollback).toContain('Server running')
       expect(result.coldRestore!.cwd).toBe('/projects/myapp')
+      expect(result.coldRestore).toMatchObject({ cols: 120, rows: 40 })
       expect(lastSpawnOpts).toMatchObject({
         sessionId,
         cwd: '/projects/myapp',
--- a/src/main/daemon/daemon-pty-adapter.ts
+++ b/src/main/daemon/daemon-pty-adapter.ts
@@ -36,6 +36,8 @@ import type { TerminalOscLinkRange } from '../../shared/terminal-osc-link-ranges
 type ColdRestorePayload = {
   scrollback: string
   cwd: string
+  cols: number
+  rows: number
   oscLinks?: TerminalOscLinkRange[]
 }
 
@@ -595,7 +597,13 @@ export class DaemonPtyAdapter implements IPtyProvider {
     if (!scrollback) {
       return null
     }
-    return { scrollback, cwd: restoreInfo.cwd, oscLinks: restoreInfo.oscLinks }
+    return {
+      scrollback,
+      cwd: restoreInfo.cwd,
+      cols: restoreInfo.cols,
+      rows: restoreInfo.rows,
+      oscLinks: restoreInfo.oscLinks
+    }
   }
 
   async sendSignal(id: string, signal: string): Promise<void> {
--- a/src/main/ipc/pty.test.ts
+++ b/src/main/ipc/pty.test.ts
@@ -10765,15 +10765,24 @@ describe('registerPtyHandlers', () => {
     )
   })
 
-  it('seeds headless terminal state with cold-restore cwd metadata', async () => {
+  it('seeds cold restore at recovered dimensions with a legacy dimensionless fallback', async () => {
     const oscLinks = [{ row: 0, startCol: 0, endCol: 8, uri: 'https://example.com/restored' }]
     const coldRestore = {
       scrollback: 'restored history\r\n',
       cwd: '/projects/restored',
+      cols: 132,
+      rows: 43,
       oscLinks
     }
+    const spawn = vi
+      .fn()
+      .mockResolvedValueOnce({ id: 'pty-cold-restore', coldRestore })
+      .mockResolvedValueOnce({
+        id: 'pty-legacy-cold-restore',
+        coldRestore: { scrollback: 'legacy history\r\n', cwd: '/projects/legacy' }
+      })
     setLocalPtyProvider({
-      spawn: vi.fn(async () => ({ id: 'pty-cold-restore', coldRestore })),
+      spawn,
       write: vi.fn(),
       resize: vi.fn(),
       kill: vi.fn(),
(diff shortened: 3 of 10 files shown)


slashdevcorpse:
Companion Windows restart-regression PR: #9064.

#9043 contains the product correction; #9064 makes the PowerShell/replay-readiness harness deterministic and adds the targeted Windows gate. They are independently mergeable, but I recommend they go hand in hand. Merge #9064 first so the workflow exists on main, then synchronize or mark #9043 ready so the gate runs on its matching product paths.

brennanb2025:
Thanks @slashdevcorpse — reviewed this end to end and it's in good shape.

**What it does:** when the daemon has lost a session and Orca cold-restores a pane from disk-recorded scrollback, that scrollback carries absolute cursor-positioning escapes that only make sense at the grid the session was recorded at. Replaying it at the *current* pane size garbled restored TUI content. This PR threads the recorded `cols`/`rows` from the daemon adapter → IPC/preload → renderer + headless emulator, clears the viewport, then replays at the recovered grid (kept from reaching the fresh PTY) before the authoritative fit resizes back to the real grid. It mirrors the existing snapshot-replay path, so the two restore routes now behave consistently.

**Why it holds up:**
- Clears the viewport *before* the recovered-grid resize, so a shrink can't promote stale rows into scrollback; viewport-only `2J` (never `3J`) preserves restored scrollback (#7310).
- New `cols`/`rows` are optional at every boundary with finite/`> 0` guards — older dimensionless payloads fall back to prior behavior.
- `suppressSnapshotReplayPtyResize → suppressStructuralReplayPtyResize` is a pure rename shared by mutually-exclusive branches; the layout-only resize never SIGWINCHes the fresh PTY.

**Verified locally:**
- Affected suite green (919 tests), including a headless-xterm oracle that models the exact op order and asserts the rendered buffer (no stale rows, recovered content kept, clean viewport) plus the legacy dimensionless fallback.
- Typecheck + lint clean on touched files; perf is a once-per-reattach path with no added polling/leak (the extra `await` is FIFO-parse-gated, not a delay).
- Live Electron: warm reattach (window reload) preserves scrollback and grid alignment with no garble — confirming the renamed shared flag doesn't regress the warm path.

The cold path itself is a Windows-restart scenario I couldn't reproduce deterministically on macOS, but it's covered by the real-xterm regression test and your same-PC Windows runs. Per the description, this should land alongside #9064 (merge that first).

PR #9058, opened 2026-07-16
fix(terminal): stop leaking Orca's NODE_ENV into spawned shells

### LINKED ISSUE #9057: [Bug]: Orca forwards its own NODE_ENV into spawned terminals
**Operating system:** macOS (platform-agnostic — the code path is shared)

**Orca version:** v1.4.144-rc.0 (reproduced running from source / dev build)

## Short summary

Orca's relay builds the environment for every integrated terminal by spreading the whole `process.env` (`src/relay/pty-handler.ts` → `buildSpawnEnv`). When Orca's own process carries `NODE_ENV` — which it does in local dev / `electron-dev` builds, where electron-vite sets `NODE_ENV=development` — that value is forwarded into every spawned shell. So every command run in an Orca terminal sees `NODE_ENV=development`.

Orca doesn't set `NODE_ENV` deliberately (dev mode is detected via `app.isPackaged` / `is.dev`, not `NODE_ENV`), so this is an unintended passthrough of a build-mode flag. It bites hardest when running Orca from source, and for anyone who launches Orca from a shell that exports `NODE_ENV`.

## What happens

`NODE_ENV=development` is Orca's build mode, not the user's — and forwarding it breaks tools in the terminal that key off `NODE_ENV`:

- **Next.js** — `next build` expects to set `NODE_ENV=production` itself. A pre-set `development` causes a dev/prod React mismatch that crashes static prerender (`TypeError: Cannot read properties of null (reading 'useContext')`).
- **Vitest** — only defaults `NODE_ENV` to `test` when it is unset. A leaked `development` silently changes test-only code paths (e.g. `next/image` host validation), so suites fail in an Orca terminal while passing on CI.

## Reproduction

1. Run Orca from source (`pnpm dev`), or launch any Orca whose process has `NODE_ENV` set.
2. Open an integrated terminal.
3. `echo $NODE_ENV` → prints `development`.
4. In a Next.js project in that terminal: `next build` crashes during prerender; `vitest` runs behave as if not in test mode.

## Expected

A spawned terminal should not inherit Orca's own `NODE_ENV`. The login shell's profile, or the tool being run, should decide `NODE_ENV` — not the app that launched the terminal.

## Fix

Strip `NODE_ENV` from the inherited environment in `buildSpawnEnv` before assembling the spawn env (a deliberately-set value via shell rc, renderer-supplied `env`, or an augmenter still wins). PR incoming.

jad:
Closes #9057.

## Summary

Orca's relay built the environment for every integrated terminal by spreading the whole `process.env` (`src/relay/pty-handler.ts` → `buildSpawnEnv`). When Orca's own process carries `NODE_ENV` — as it does in local dev / `electron-dev` builds, where electron-vite sets `NODE_ENV=development` — that value was forwarded into every spawned shell, so every command in an Orca terminal saw `NODE_ENV=development`.

That's Orca's build mode, not the user's, and it silently breaks tools that key off `NODE_ENV`:

- **Next.js** — `next build` expects to set `production` itself; a pre-set `development` crashes static prerender with a dev/prod React mismatch (`Cannot read properties of null (reading 'useContext')`).
- **Vitest** — only defaults `NODE_ENV` to `test` when unset; a leaked `development` changes test-only branches (e.g. `next/image` host validation), so suites fail in an Orca terminal while passing on CI.

This strips `NODE_ENV` from the **inherited** environment before assembling the spawn env. A deliberately-set value still wins — from the login shell's profile, a renderer-supplied `env`, or an env augmenter — so only the ambient Orca build-mode value is dropped.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — ran `oxlint` on the changed files (clean); the pre-commit hook also ran `oxlint` + `oxfmt`. Did not run the full aggregate `lint` script locally.
- [x] `pnpm typecheck` — clean for the changed files. (Two unrelated pre-existing errors in this checkout — `Cannot find module 'typescript-api'` in two other test files — appear regardless of this change.)
- [x] `pnpm test` — `vitest run src/relay/pty-handler.test.ts` → **97 passed**, including 2 new cases. Did not run the full suite locally.
- [ ] `pnpm build` — not run locally (multi-platform desktop build); relying on CI.
- [x] Added high-quality tests that would catch a regression: one asserts the spawn env drops an ambient `NODE_ENV` while keeping other inherited vars (`PATH`); another asserts a renderer-supplied `NODE_ENV` still passes through.

## AI Review Report

Reviewed by the AI coding agent that wrote the change:

- **Cross-platform (macOS/Linux/Windows):** the change is in the relay's env assembly with no platform branches — it applies identically on all three. No shortcuts, labels, or path handling touched.
- **Local / remote / SSH:** `buildSpawnEnv` feeds both the local and SSH relay spawn paths, so the strip applies uniformly. Verified the renderer-supplied `env` is curated (specific vars, not a blanket `process.env` spread), so it won't silently re-introduce `NODE_ENV` — and if a startup plan sets it deliberately, it still wins.
- **Agents / integrations:** no agent- or provider-specific logic touched. Terminals previously saw `development`; they now see the shell/tool default, which is the correct baseline for CLI agents.
- **Performance:** negligible — one shallow object copy + a `delete`, on a path that already spread `process.env`. Not a hot path.
- **UI:** none.

## Security Audit

- No new input handling, command execution, path handling, auth, or IPC surface. The change **removes** an environment variable from what child processes inherit; it does not add or widen anything. No secrets involved (`NODE_ENV` is a build-mode flag). Net effect is slightly less ambient state leaking into spawned shells.

## Notes

Platform-agnostic; affects local and SSH relays equally. No follow-up required. Discovered while running project test/build commands inside an Orca terminal.

X: @debay

### DIFF
--- a/src/relay/pty-handler.test.ts
+++ b/src/relay/pty-handler.test.ts
@@ -181,6 +181,52 @@ describe('PtyHandler', () => {
     expect(handler.activePtyCount).toBe(1)
   })
 
+  it("does not forward Orca's own NODE_ENV into the spawned shell", async () => {
+    // Why: Orca's process carries NODE_ENV (`development` in dev/electron-dev
+    // builds). Forwarding it into integrated terminals breaks tools that key
+    // off NODE_ENV (e.g. `next build` / Vitest). It must be stripped from the
+    // inherited env; other inherited vars (PATH) must survive.
+    const original = process.env.NODE_ENV
+    process.env.NODE_ENV = 'development'
+    try {
+      await dispatcher.callRequest('pty.spawn', { cols: 80, rows: 24 })
+
+      const spawnOptions = mockPtySpawn.mock.calls[0][2] as { env: Record<string, string> }
+      expect(spawnOptions.env).not.toHaveProperty('NODE_ENV')
+      expect(spawnOptions.env.PATH).toBe(process.env.PATH)
+    } finally {
+      if (original === undefined) {
+        delete process.env.NODE_ENV
+      } else {
+        process.env.NODE_ENV = original
+      }
+    }
+  })
+
+  it('lets a renderer-supplied NODE_ENV through (only the ambient Orca value is stripped)', async () => {
+    // Why: stripping targets the *inherited* process env, not an explicit
+    // caller request. A shell profile or startup plan that deliberately sets
+    // NODE_ENV must still win.
+    const original = process.env.NODE_ENV
+    process.env.NODE_ENV = 'development'
+    try {
+      await dispatcher.callRequest('pty.spawn', {
+        cols: 80,
+        rows: 24,
+        env: { NODE_ENV: 'test' }
+      })
+
+      const spawnOptions = mockPtySpawn.mock.calls[0][2] as { env: Record<string, string> }
+      expect(spawnOptions.env.NODE_ENV).toBe('test')
+    } finally {
+      if (original === undefined) {
+        delete process.env.NODE_ENV
+      } else {
+        process.env.NODE_ENV = original
+      }
+    }
+  })
+
   it('atomically caps concurrent PTY spawn admission', async () => {
     const results = await Promise.allSettled(
       Array.from({ length: MAX_RELAY_PTY_SESSIONS + 1 }, () =>
--- a/src/relay/pty-handler.ts
+++ b/src/relay/pty-handler.ts
@@ -376,9 +376,21 @@ export class PtyHandler {
     ctx: { id: string; paneKey?: string; shell: string; command?: string },
     envToDelete: readonly string[] = []
   ): Record<string, string> {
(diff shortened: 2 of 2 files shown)


nwparker:
Thanks for catching this and writing the first fix, @jad 🙏

We went with #9649, which took the same idea and extended it to **all three** PTY-spawn sites — the local daemon (`pty-subprocess.ts`) and the local provider (`local-pty-provider.ts`) in addition to the relay path this PR patched. The issue's own repro (open an integrated terminal from a dev build, `echo $NODE_ENV`) runs through the local daemon path, so the relay-only patch here didn't cover it. #9649 credits you as a co-author on the squashed commit.

Closing in favor of #9649 (now merged). Really appreciate the report and the fix — thank you!

PR #9064, opened 2026-07-16
test(e2e): harden terminal restart regressions on Windows

slashdevcorpse:
## Summary

- Make terminal restart probes native-PowerShell safe: fresh-shell discovery sends no startup control bytes, settled Windows probes separate Ctrl+C from the command, PowerShell installs a prompt function, and cold daemon loss uses bounded `taskkill.exe` termination.
- Gate keyboard assertions on a full renderer-input round trip that survives replay suppression, delayed ConPTY echo, and pane replacement without accepting replayed markers from a stale pane.
- Enable the cold daemon-loss restart regression on Windows and add a focused Windows 2022 workflow whose three selected cases cannot silently skip required setup.
- Keep this test-only companion independently mergeable from #9043 while recommending that both PRs land together as regression coverage for #8751.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — the full Windows suite was attempted: 30,274 passed, 342 skipped, and 100 failed with 2 suite errors. The failures are existing broad Windows incompatibilities such as `/bin/sh ENOENT`, POSIX paths/signals, and line-ending assumptions; none touch this patch.
- [ ] `pnpm build` — the full production/native build was not run; `pnpm exec electron-vite build --mode e2e` passed on the rebased commit.

Focused validation on commit `2ef995ea34a0af8682f9315533d21a87e85f1325`:

- `pnpm exec vitest run tests/e2e/restored-terminal-input-readiness.unit.test.ts tests/e2e/terminal-probe-input-sequence.unit.test.ts` — 2 files / 9 tests passed.
- Required Windows workflow command — 3/3 passed in 2.0 minutes with normal cleanup:
  - clean restart with the live daemon session restored;
  - cold restore after the daemon was force-killed between launches;
  - daemon snapshot relaunch preserving the cursor on the PowerShell prompt.
- The cold daemon-loss assertion also passed 5/5 in an earlier isolated repetition.
- Workflow YAML parsing and `git diff --check` passed.

The original driver was red on Windows because a fresh PowerShell probe was prefixed with POSIX Ctrl+U/startup control bytes. During adversarial review, a real clean-restart run exposed a second red case: readiness kept only the latest 100 ms marker while valid ConPTY echoes arrived later. The terminal tail contained the markers, but the helper never observed the immediately previous one. The final helper retains all outstanding markers for the same concrete pane instance and has dedicated delayed-echo and pane-replacement regressions.

## AI Review Report

Three parallel audits checked Windows runtime behavior, replay/readiness semantics, CI non-skippability and security, cross-platform compatibility, and code minimality.

Findings corrected during review:

- Bound readiness evidence to the concrete mounted pane so replaying an old marker into a replacement pane cannot false-pass.
- Retained delayed same-pane markers instead of chasing only the newest marker every 100 ms.
- Changed stale diagnostics from “renderer transport unbound” to the accurate renderer-input-path possibilities: replay, focus, or binding.
- Made missing seeded repo state and missing PaneManager hard failures under the dedicated workflow flag.
- Removed renderer-crash coverage from this workflow and reverted its spec diff to keep the companion focused.
- Extracted restored-input readiness into a concrete module rather than adding a forbidden max-lines waiver.

Cross-platform review covered macOS, Linux, Windows/ConPTY, WSL, and SSH. POSIX shells retain Ctrl+C + Ctrl+U behavior; Windows receives Ctrl+C as a separate payload and never receives Ctrl+U. No shortcut, provider, path-separator, or local-only product behavior changed.

Final verdict from all three audits: ship, with no blockers.

## Security Audit

This PR changes tests and CI only; it adds no product dependency, auth, secret, network, filesystem API, or IPC surface. Windows termination targets only the daemon PID parsed from the test's isolated user-data directory, passes arguments through `execFileSync` without a shell, is bounded by a 10-second timeout, and verifies the PID is gone.

The workflow uses `pull_request` rather than `pull_request_target`, grants read-only contents permission, does not persist checkout credentials, runs serially with a 30-minute job timeout, and uploads only failure traces.

## Notes

- Paired with #9043; regression coverage for #8751. This PR intentionally uses `Refs #8751` semantics and does not claim to close the production issue.
- Recommended order: merge this companion first so the Windows workflow exists on `main`, then synchronize or mark #9043 ready for review. Every product file in #9043 matches this workflow's trigger paths.
- The PRs are independently mergeable, but they should go hand in hand: #9043 supplies the cold-restore correction, while this PR makes the Windows restart harness deterministic and enforces the regression gate.
- Packaged warm/cold verification remains the closing gate for #8751.

### DIFF
--- a/.github/workflows/windows-terminal-restart-e2e.yml
+++ b/.github/workflows/windows-terminal-restart-e2e.yml
@@ -84,8 +84,10 @@ jobs:
           SKIP_BUILD: '1'
           ORCA_E2E_FORWARD_APP_LOGS: '1'
           ORCA_REQUIRE_WINDOWS_TERMINAL_RESTART_E2E: '1'
+        # Why: pnpm forwards a literal `--` to Playwright, which makes the
+        # grep and worker flags positional filters instead of CLI options.
         run: >-
-          pnpm run test:e2e --
+          pnpm run test:e2e
           tests/e2e/restart-restore-terminal-input.spec.ts
           tests/e2e/terminal-restart-persistence.spec.ts
           --grep "clean restart with a live daemon session|cold-restored pane accepts typing|daemon snapshot relaunch preserves"
--- a/.github/workflows/windows-terminal-restart-e2e.yml
+++ b/.github/workflows/windows-terminal-restart-e2e.yml
@@ -1,0 +1,101 @@
+name: Windows terminal restart E2E
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths:
+      - '.github/workflows/windows-terminal-restart-e2e.yml'
+      - 'package.json'
+      - 'pnpm-lock.yaml'
+      - 'electron.vite.config.ts'
+      - 'config/patches/**'
+      - 'config/scripts/ensure-native-runtime.mjs'
+      - 'config/scripts/rebuild-native-deps.mjs'
+      - 'tests/playwright.config.ts'
+      - 'tests/e2e/global-setup.ts'
+      - 'tests/e2e/global-teardown.ts'
+      - 'tests/e2e/helpers/**'
+      - 'tests/e2e/restart-restore-terminal-input.spec.ts'
+      - 'tests/e2e/restored-terminal-input-readiness.unit.test.ts'
+      - 'tests/e2e/terminal-probe-input-sequence.ts'
+      - 'tests/e2e/terminal-probe-input-sequence.unit.test.ts'
+      - 'tests/e2e/terminal-restart-persistence.spec.ts'
+      - 'src/main/daemon/**'
+      - 'src/main/ipc/pty*.ts'
+      - 'src/main/providers/**'
+      - 'src/main/pty/**'
+      - 'src/preload/**'
+      - 'src/renderer/src/components/terminal-pane/**'
+      - 'src/renderer/src/lib/pane-manager/**'
+      - 'src/shared/pty-session-id-format.ts'
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: Ref to check out
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: windows-terminal-restart-e2e-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  windows-terminal-restart:
+    name: Windows terminal restart regressions
+    runs-on: windows-2022
+    timeout-minutes: 30
+    env:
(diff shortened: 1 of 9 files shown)


slashdevcorpse:
Addressed CodeRabbit's [document-replacement recovery review](https://github.com/stablyai/orca/pull/9064#pullrequestreview-4717342153) in `2ef995ea3`.

The new regression proves the catch path clears real stale state:

- call 1 drops marker A and records A pending;
- call 2 writes marker B, swaps to the replacement pane, then `page.evaluate` rejects;
- call 3 starts with zero pending attempts and must send fresh marker C through the replacement pane;
- call 4 observes C and succeeds.

The assertions require pending counts `[0, 1, 0, 1]`, two original-pane inputs, one replacement-pane input, and `C !== B`. Removing the catch-path reset makes the test fail.

Validation after the fix:

- `pnpm lint`
- `pnpm typecheck`
- focused readiness/probe units: 2 files, 9/9 tests
- independent adversarial audit: ship, no blockers

The walkthrough's docstring-percentage item is a generic warning rather than a repository gate. I did not add restatement-only docstrings to internal E2E helpers because the project requires concise comments for non-obvious reasons; the full repository lint, max-lines, and localization gates pass. There were no inline review threads to resolve.

brennanb2025:
Thanks for the thorough companion coverage here. Reviewed against latest `main` and verified:

- **Platform-aware probe sequences** correctly separate the PowerShell/ConPTY path (no POSIX Ctrl+U; Ctrl+C sent as its own write) from the POSIX path, and fresh-shell discovery drops the startup control bytes that ConPTY was echoing as literal `^C`.
- **Restored-input readiness gate** binds evidence to the concrete mounted pane instance, so a marker replayed into a *replacement* pane can't false-pass, and it retains every same-pane marker so delayed ConPTY echo doesn't flake.
- **Windows-2022 workflow** runs exactly the 3 intended cases (grep patterns each match one real title), and `ORCA_REQUIRE_WINDOWS_TERMINAL_RESTART_E2E=1` turns every silent-skip path (missing seeded repo, missing PaneManager) into a hard failure. Fork-safe: `pull_request`, `contents: read`, `persist-credentials: false`, no secrets, failure-only artifacts.

One small hardening pushed on top (`fa37926`): the POSIX cold-kill path now tolerates `ESRCH` if the daemon self-exits between the liveness check and the `SIGKILL`, matching the tolerance the Windows `taskkill` branch already has — otherwise a benign race could throw a spurious test failure.

Local checks green: the two new unit suites (9/9), `pnpm lint`, and `pnpm typecheck`. Leaving the merge to you so the ordering with #9043 can be coordinated.

Issue #8454, opened 2026-07-13
[Bug]: Built-in browser can't reach an HTTPS local dev server (localhost forced to http://, self-signed certs rejected)


ivaaaaann:
### Operating system

macOS

### Orca version

1.4.137

## Short summary

The built-in browser cannot reach a local dev server that is served over **HTTPS** (e.g. `next dev --experimental-https`, or any local staging on `https://localhost:3000`). Two separate behaviors combine to make an HTTPS localhost dev server effectively unreachable from Orca's embedded browser, so it looks like Orca "can't find" the running server even though it opens fine in a system browser.

## What happened

I run my app's local staging over HTTPS at `https://localhost:3000`. Opening it from Orca's internal browser fails to load the page. The same server opens fine in a normal system browser (Chrome/Safari).

## How can we reproduce it

1. Start any local dev/staging server over HTTPS, e.g. `next dev --experimental-https` (serves `https://localhost:3000` with a self-signed cert).
2. In Orca's embedded browser address bar, type `localhost:3000` and press Enter.
   * Orca navigates to `http://localhost:3000`, which the HTTPS-only server cannot answer → the page fails to load.
3. Now type the full `https://localhost:3000`.
   * The scheme is preserved this time, but the page still fails to load because the self-signed certificate is rejected with no "proceed anyway" option.

## Root cause (from reading `main`)

**1. Bare localhost/loopback input is hard-coded to** `http://`**.**<br>In `src/shared/browser-url.ts`, `normalizeBrowserNavigationUrl` matches scheme-less loopback addresses with `LOCAL_ADDRESS_PATTERN` and always prepends `http://`:

```ts
const LOCAL_ADDRESS_PATTERN =
  /^(?:localhost|127(?:\.\d{1,3}){3}|0\.0\.0\.0|\[[0-9a-f:]+\])(?::\d+)?(?:[/?#].*)?$/i

if (LOCAL_ADDRESS_PATTERN.test(trimmed)) {
  try {
    return new URL(`http://${trimmed}`).toString()  // ← always http://, no https branch
  } catch {
    return null
  }
}
```

So `localhost:3000` → `http://localhost:3000`. There is no path that tries/keeps `https://` for a local address. Typing the scheme explicitly (`https://localhost:3000`) is the only way to reach an HTTPS local server, which is not discoverable.

**2. Self-signed certs are rejected with no bypass.**<br>The guest webview is created with strict security and there is no certificate-error handler anywhere in the codebase:

[...]

PR #9070, opened 2026-07-16
fix(browser): gate certificate proceed on runtime capability; dedupe cert-error helpers

AmethystLiang:
## Summary

Follow-up to the certificate-trust feature (issue #8454). Hardens the "Proceed Anyway (Unsafe)" affordance and dedupes the certificate-error helpers.

- **Gate proceed on runtime capability.** A remote runtime that predates `browser.certificate-trust.v1` cannot honor a proceed request, so `RemoteBrowserPagePane` now hides the proceed affordance until the connected runtime positively advertises the capability (probed via `runtimeEnvironmentSupportsCapability`). The error page still renders — only the unactionable "Proceed Anyway" button is suppressed. Desktop/built browsers advertise the capability whenever they can host a page, so the local HTTPS-dev-server case keeps its proceed button.
- **Single certificate-error normalizer.** Extracts `normalizeCertificateError` into `src/shared/browser-certificate-errors.ts` so the main-process certificate-error handler and the renderer overlay can never silently diverge on challenge matching (`browser-certificate-identity.ts` re-exports it for existing importers).
- **Centralize load-error redaction.** Both the `did-fail-load` path and the synthesized certificate failure now build load errors through one `BrowserManager.buildLoadError`, so Kagi session-token redaction can't be forgotten by a future call site.
- **Fail-closed logging.** The trust controller's catch now logs before answering `false`, so a throw during challenge recording no longer presents as "the cert prompt never appears" with no trace.

## Screenshots

No visual change in the supported case. In the unsupported-runtime case the "Proceed Anyway (Unsafe)" button no longer appears on the load-failure overlay.

## Testing

- [x] `pnpm lint` (oxlint on changed dirs — no new findings)
- [x] `pnpm typecheck` (web + node projects — clean)
- [x] `pnpm test` (`browser-certificate-trust-controller.test.ts`, `runtime-rpc-client.test.ts` — 34 passed)
- [ ] `pnpm build`
- [x] Existing unit tests cover the trust-controller fail-closed path and the capability-probe cache; the gating is a thin prop guard over already-tested surfaces.

## AI Review Report

Reviewed correctness of the capability gate, the helper extraction, and the logging change.

- Verified passing `certificateFailure={null}` suppresses **only** the proceed button — the overlay body is driven by `loadError`, so the error page still renders.
- Confirmed `browser.certificate-trust.v1` is a conditional capability (advertised whenever `canBrowse`), matching `browser.headless.v1`, so desktop/built browsers keep proceed while pre-feature remote runtimes hide it.
- Checked the `useEffect` probe has complete deps and a `cancelled` guard against out-of-order resolution.
- Confirmed all `normalizeCertificateError` importers still resolve after the move.
- Cross-platform: change is browser-pane/main-process logic with no new shortcuts, labels, paths, or shell invocations; no platform-specific branches touched.

## Security Audit

- Load-error redaction is now centralized, reducing the chance a future call site leaks a Kagi session token from an untrusted URL.
- The capability gate fails closed (defaults to unsupported, `.catch` → unsupported), so a probe error hides proceed rather than exposing an unactionable bypass.
- No new input handling, command execution, auth, or secret paths introduced. IPC surface is the existing `status.get` capability probe.

## Notes

The proceed affordance for the **local** webview pane (`BrowserPagePane`) is intentionally left ungated — the local runtime always owns its browser pages and advertises the capability.

### DIFF
--- a/src/main/browser/browser-certificate-identity.ts
+++ b/src/main/browser/browser-certificate-identity.ts
@@ -1,14 +1,11 @@
 import { createHash } from 'node:crypto'
 
+import { normalizeCertificateError } from '../../shared/browser-certificate-errors'
+
+export { normalizeCertificateError }
+
 export const SUPPORTED_CERTIFICATE_ERROR = 'ERR_CERT_AUTHORITY_INVALID'
 export const SUPPORTED_CERTIFICATE_ERROR_CODE = -202
-
-export function normalizeCertificateError(error: string): string {
-  return error
-    .trim()
-    .replace(/^net::/i, '')
-    .toUpperCase()
-}
 
 export function getSupportedCertificateErrorCode(error: string): number | null {
   return normalizeCertificateError(error) === SUPPORTED_CERTIFICATE_ERROR
--- a/src/main/browser/browser-certificate-trust-controller.ts
+++ b/src/main/browser/browser-certificate-trust-controller.ts
@@ -107,7 +107,11 @@ export class BrowserCertificateTrustController {
         })
       }
       answer(false)
-    } catch {
+    } catch (error) {
+      // Why: fail closed, but log first — a throw in challenge recording would
+      // otherwise present as "the cert prompt never appears" with no trace,
+      // matching the logging catch in browser-manager's guest-state notifier.
+      console.error('[browser-certificate-trust-controller] handleCertificateError failed', error)
       answer(false)
     }
   }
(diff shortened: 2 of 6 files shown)


nwparker:
Superseded by https://github.com/stablyai/orca/pull/9104 — takes this design/implementation (credit @AmethystLiang), rebases onto current main, resolves conflicts, and applies adversarial hardening from review. Thanks for the solid design doc + full stack work.

Issue #7547, opened 2026-07-06
[Bug]: Windows main-process crash in native file watcher (watcher.node, 0xc0000409 fail-fast) — 5x in 25 min on 1.4.116; third platform for the #5377 crash class


n-leezy:
**Operating system:** Windows 11 Home (10.0.26200)
**Orca version:** 1.4.116 (crashes); currently on the same line with a pending auto-update downloaded
**Bundled watcher:** `@parcel/watcher` 2.5.6 (`@parcel/watcher-win32-x64`)

### Short summary

Orca's main process is being killed by a fail-fast abort (`0xc0000409`, STATUS_STACK_BUFFER_OVERRUN / `__fastfail`) inside the bundled native file watcher (`watcher.node`). It crashed **five times in ~25 minutes** during one working session, all with the identical signature (same module, same fault offset). From the user's perspective the app just "closes by itself" with no error dialog.

This looks like the **same crash class as #5377** (macOS 1.4.68, `Watcher::findCallback` SIGSEGV / `napi_fatal_error` abort during watcher worker teardown, closed) and **#6635** (Linux/WSL2 headless serve, 1.4.97, `Napi::Error` SIGABRT). Windows makes it the **third platform** reporting this signature after the #5377 fixes shipped — suggesting the underlying watcher lifecycle bug (stale/freed napi callback reference during worker teardown) is still present. `0xc0000409` on Windows is what a `napi_fatal_error` / CRT abort path produces, matching crash #3 in #5377.

### Crash records (Windows Event Log, Application, Event ID 1000)

All five crashes on 2026-07-02, identical signature:

```
2026-07-02 14:09:17 | Orca.exe 1.4.116.0 | watcher.node | 0xc0000409 | offset 0x352c9
2026-07-02 14:04:56 | Orca.exe 1.4.116.0 | watcher.node | 0xc0000409 | offset 0x352c9
2026-07-02 13:59:01 | Orca.exe 1.4.116.0 | watcher.node | 0xc0000409 | offset 0x352c9
2026-07-02 13:51:22 | Orca.exe 1.4.116.0 | watcher.node | 0xc0000409 | offset 0x352c9
2026-07-02 13:44:01 | Orca.exe 1.4.116.0 | watcher.node | 0xc0000409 | offset 0x352c9
```

Faulting module path:
[...]

PR #9075, opened 2026-07-16
fix(worktrees): run the macOS git-common narrow watch in the crash-isolated watcher child

nahuelb:
## Problem

Orca 1.4.143 crashes with EXC_BREAKPOINT (SIGTRAP) in the **Electron main process** on the FSEvents thread (`FSEventsBackend::start → FSEventsCallback → Watcher::isIgnored → brk 0`) under heavy worktree churn — e.g. many parallel builds in linked worktrees followed by bulk `git worktree remove`. Reproduced twice in ~30 minutes on macOS 26.5.2 (M4 Pro).

## Cause

The macOS git-common narrow watch (`src/main/ipc/worktree-git-common-watch.ts`) was the last production call site loading native `@parcel/watcher` directly in the Electron main process. Two assumptions broke:

- `.git/worktrees` is not the rare-churn tree the code assumed — every linked worktree's `HEAD`/`index`/lock files live under `<common>/.git/worktrees/<name>/`, so parallel agents generate constant events there.
- Bulk worktree removal deletes the watched root itself, hitting parcel-watcher 2.5.6's unsynchronized FSEvents teardown race ([`FSEventsBackend.cc`](https://github.com/parcel-bundler/watcher/blob/v2.5.6/src/macos/FSEventsBackend.cc): the root-deletion self-stop runs `stopStream()` + `watcher->state = nullptr` on the FSEvents thread without the backend mutex, racing `unsubscribe()` on the JS thread — a non-atomic `shared_ptr` read/write race and potential double `FSEventStreamRelease`).

The resulting heap corruption trips Electron's allocator (PartitionAlloc traps with `brk 0` — the "V8 internals" frames in the crash reports are nearest-symbol misattribution) and fail-fasts the whole app. 2.5.6 is the latest upstream release, so no fixed version exists to bump to; the race deserves a separate upstream report.

## Fix

Route the subscription through `subscribeViaWatcherProcess` like every other watcher, honoring the repo's own crash-isolation invariant (`parcel-watcher-process-entry.ts`, issue #7547). Root-deletion teardown-and-rearm semantics are unchanged — events (including the root `delete`) and watch errors reach the callback in the same shape. A new `onInterruption` hook surfaces a synthetic root update so the worktree list refreshes past the event gap after a watcher-child respawn. A native crash now costs a child restart instead of the app.

## Validation

- `pnpm typecheck:node`, oxlint, max-lines ratchet — clean.
- `worktree-base-directory-poller.test.ts` (includes the real-FSEvents macOS narrow-stream integration tests: add/update/remove, existence-poll arming, delete-and-recreate rearm) — 15/15 pass.
- `parcel-watcher-process`, `parcel-watcher-process-entry`, `filesystem-watcher` suites — 70/70 pass.
- Smoke-tested the built child entry (`out/main/parcel-watcher-process-entry.js`, real native `watcher.node`): subscribe to a `worktrees` dir → churn HEAD/index files → delete the watched root → unsubscribe, then 10 repeated cycles — host process unaffected throughout.

### DIFF
--- a/src/main/ipc/worktree-git-common-watch.ts
+++ b/src/main/ipc/worktree-git-common-watch.ts
@@ -1,5 +1,6 @@
 import { stat } from 'node:fs/promises'
 import { join } from 'node:path'
+import { subscribeViaWatcherProcess } from './parcel-watcher-process'
 import type { WorktreeBaseWatchTarget } from './worktree-base-directory-event-filter'
 import type {
   WorktreeBasePollEvent,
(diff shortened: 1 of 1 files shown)


nahuelb:
Superseded by #9091

PR #9082, opened 2026-07-16
fix(skills): hide split icon and stop update-terminal respawn on refocus

brennanb2025:
## What

Two fixes to the **Update skills** modal, whose embedded terminal is an `OnboardingInlineCommandTerminal` wrapping the regular `TerminalPane`.

### 1. Remove the split icon from inline command terminals

The pane header rendered the regular terminal's **Split Terminal Right** button. That affordance is useful in a normal terminal, but it is noise in an ephemeral one-command surface.

- Added `showSplitButton` (default `true`) through `TerminalPane` → `TerminalPaneHeaderOverlay`.
- `OnboardingInlineCommandTerminal` passes `showSplitButton={false}`.
- The name is deliberately scoped to the header button: keyboard and context-menu split behavior are unchanged.

### 2. Preserve the update PTY when Orca regains focus

Repro: open the modal, switch to another app, then return to Orca. The focus freshness scan temporarily publishes `{ inventory: null, loading: true }`. The dialog used to clear `terminalCommand`, unmounting the keyed inline terminal; the later eligible result mounted a new terminal and spawned a new PTY.

The dialog now keeps the existing terminal mounted while inventory is revalidated. Its wrapper becomes `inert`, `aria-busy`, and visually muted while authority is absent, so the stale draft cannot receive input. If the scan returns the same authorized command, the same terminal instance is re-enabled. If eligibility changes or the scan fails, the stale draft is still removed.

This intentionally preserves the existing focus rescan. Skipping it would leave a previously authorized shell command runnable after external skill/topology changes, contrary to the freshness safety contract.

### 3. Repair the required generated manifest

The branch's current `main` identifies the app as `1.4.144-rc.2`, while `resources/skills/current-manifest.json` still identified the bundled skill entries as `rc.1`. The official generator updates only those version fields; digests and release revisions are unchanged.

## Live validation (Electron)

Validated in a running dev build of this branch (`octopus @ fix/skill-update-terminal-split-respawn`) against a skill-freshness fixture with `orca-cli` set to an outdated registry revision, so the modal opens a real update terminal (`npx skills update orca-cli --global`).

**Fix 1 — no split icon.** The modal's terminal header has zero split controls (DOM check: 0 `Split Terminal Right` buttons in the modal, 0 anywhere in the app), and the icon that used to sit in the top-right of the terminal is gone:

Update skills modal with no split icon

**Fix 2 — PTY survives refocus.** Typed a sentinel (`ZZZSENTINEL`) into the pre-filled draft, then drove a focus loss/gain and sampled the terminal across the rescan:

| Phase | PTY id | mounted | inert | typed draft |
|---|---|---|---|---|
| before | `…@@3ba50f8f` | yes | no | present |
| during revalidation | `…@@3ba50f8f` (same) | yes | yes | present |
| settled | `…@@3ba50f8f` (same) | yes | no | present |

The PTY id never changes and the typed draft survives the whole cycle — no respawn. During revalidation the wrapper goes `inert`/dimmed (the stale draft can't take input), then re-enables. If the terminal had respawned, `ZZZSENTINEL` would be gone and a fresh command re-pasted.

| Before losing focus (sentinel typed) | After returning focus (sentinel survived) |
|---|---|
| before | after |

**Method note:** OS-level app switching from the automation shell was blocked (no Accessibility/Automation permission), so the focus loss/gain was driven at the window level via CDP — `page.bringToFront()` plus dispatching the same `window` `focus` event that `onWindowFocus` listens for. That exercises the identical handler an alt-tab fires; the observable proof (same PTY id, surviving typed draft) is the point. Not yet exercised in a fully packaged/production build.

## Reliability and performance evidence

- **Invariants:** `skill-freshness.update-command-authority` — an unsubmitted draft accepts input only while backed by current inventory; `terminal-session.ephemeral-pty-continuity` — revalidation of the same command does not replace its PTY.
- **Failure source:** window focus invalidates inventory, and clearing the keyed command previously ran the inline terminal's close/create lifecycle.
- **Oracle:** the dialog regression test asserts the terminal DOM node is referentially identical before, during, and after revalidation; it is inert only while inventory is absent. Existing tests prove changed/failed inventory revokes the draft and submitted commands survive until exit.
- **Performance budget:** the existing hook test still observes one bounded trailing focus scan. The dialog test observes zero terminal remounts for an unchanged command. No polling, provider fanout, broad session scan, timer, or subprocess loop was added.
- **Gate:** no matching entry exists in `config/reliability-gates.jsonc`; focused renderer tests are the accepted gap for this PR.
- **Provider/platform coverage:** local, daemon, SSH/remote, WSL, mobile/relay, macOS, Linux, and Windows are unaffected at the provider layer; the change is renderer-only and uses standard HTML `inert` plus existing tokens.
- **Diagnostics:** `data-skill-update-terminal` and `aria-busy` expose the pending authorization state to deterministic tests and DOM inspection.

## Verification

- 25 focused tests across skill freshness, update dialog, inline terminal readiness, and terminal header overlay
- `pnpm run typecheck:web`
- changed-file oxlint and React Doctor
- oxfmt check
- reliability-gate and max-lines-ratchet checks
- bundled-skill guide, generated-manifest, localization catalog, and localization coverage verification
- **live Electron validation of both fixes (screenshots above)**

Repository-wide `pnpm run lint` reaches a pre-existing switch-exhaustiveness error in untouched `skill-freshness-group.tsx`; all checks relevant to this diff pass.

### DIFF
--- a/resources/skills/current-manifest.json
+++ b/resources/skills/current-manifest.json
@@ -1,11 +1,11 @@
 {
   "schemaVersion": 1,
-  "appVersion": "1.4.144-rc.1",
+  "appVersion": "1.4.144-rc.2",
   "skills": [
     {
       "name": "computer-use",
       "sourcePath": "skills/computer-use",
-      "appVersion": "1.4.144-rc.1",
+      "appVersion": "1.4.144-rc.2",
       "releaseRevision": 5,
       "packageDigest": "cd2809474d57fd7277adb277448e6fa446810d3cbad71ac0b473b9e8ff1bad68",
       "gitTreeSha": "306c0f8cb63bcac265a5b7975dc2f855be4f1344",
@@ -24,7 +24,7 @@
     {
       "name": "linear-tickets",
       "sourcePath": "skills/linear-tickets",
-      "appVersion": "1.4.144-rc.1",
+      "appVersion": "1.4.144-rc.2",
       "releaseRevision": 4,
       "packageDigest": "f198d7b22e5ee1673dac403f9cca0553b124e0a90e4fdd05d2c23b7344e32d2b",
       "gitTreeSha": "de9fc106bbb4e313a90ff9a9513a720909bbd176",
@@ -43,7 +43,7 @@
     {
       "name": "orca-cli",
       "sourcePath": "skills/orca-cli",
-      "appVersion": "1.4.144-rc.1",
+      "appVersion": "1.4.144-rc.2",
       "releaseRevision": 32,
       "packageDigest": "51740ff13f379ac5743d3fd28a14b17168dcef40f7048c20182dce166098c45f",
       "gitTreeSha": "ded93000a5f654e2b4f324501282459bd56afe19",
@@ -62,7 +62,7 @@
     {
       "name": "orca-emulator",
       "sourcePath": "skills/orca-emulator",
-      "appVersion": "1.4.144-rc.1",
+      "appVersion": "1.4.144-rc.2",
       "releaseRevision": 4,
       "packageDigest": "453b1d9aa20b51b8a4d32c7b6def6a93f7ef9c730de32abbcbc1788ad1b1820b",
       "gitTreeSha": "66be6abe99f1807da85934aee0e22daefc8f7656",
@@ -81,7 +81,7 @@
     {
       "name": "orca-emulator-android",
       "sourcePath": "skills/orca-emulator-android",
-      "appVersion": "1.4.144-rc.1",
+      "appVersion": "1.4.144-rc.2",
       "releaseRevision": 2,
       "packageDigest": "12272cf82e0731f11e424822b961882457034e730358cc65ea28e4eb9c8ff7f5",
       "gitTreeSha": "f7b0fc8cbf5cd78ca5156f6bbe3a20f1462d8f83",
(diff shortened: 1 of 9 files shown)


brennanb2025:
Senior review follow-up — pushed in 9865fce3f:

- Closed a stale-authority commit window: when fresh inventory changed the eligible command, the old draft briefly became interactive before the passive replacement effect. A command mismatch now remains inert through that commit; the regression test observes the actual committed DOM state before removal.
- Preserved keyboard continuity: Chromium drops focus when an ancestor becomes inert and does not restore it. The dialog now restores the same preserved xterm focus target after unchanged authorization returns, does not steal focus after an intentional move, and releases detached targets when the terminal closes or is replaced.
- Performance/resource check: no polling, timer, subprocess, session scan, PTY remount, or terminal-keydown work was added. Focus bookkeeping runs only on focus transitions, and the same-command path still observes zero terminal remounts.

Verification: 25 focused tests, web typecheck, changed-file oxlint and React Doctor, formatting, bundled-skill manifest/guides, reliability/max-lines/localization gates, and the production Electron/Vite build all pass. Repository-wide lint still stops only at the existing untouched switch-exhaustiveness error in skill-freshness-group.tsx.

PR #9091, opened 2026-07-17
fix(watcher): host macOS git-common narrow watch in the watcher child

### LINKED ISSUE #8732: [Bug]: app crashes
### Operating system

macOS

### Orca version

v1.4.138

### Details

The app crashed while a few agents were running (2-3), the Orca app was not focused at the time.

Crash report:
https://gist.github.com/manuelmeurer/19e00b68dbda4b5310ae3a1e90410501

brennanb2025:
Fixes #8732

## Root cause

The P0 macOS app crash in #8732 (`EXC_BREAKPOINT`/`brk 0` on the FSEvents thread, trapped while allocating inside `Watcher::isIgnored`) is heap corruption from data races inside `@parcel/watcher` 2.5.6's FSEvents backend. Verified with a ThreadSanitizer build of `watcher.node`, which flags the races within ~40 subscribe/write-burst/delete-root/unsubscribe iterations:

- `FSEventsBackend::unsubscribe` writes `watcher->state = nullptr` (`FSEventsBackend.cc:336`) with no synchronization against the FSEvents callback thread's `shared_ptr` copy of that same field (`cc:60`/`cc:64`) → control-block double-free → the next allocation on the FSEvents thread traps (the exact crashed frame).
- `startStream` starts delivering events (`cc:250`) before `state->tree` is assigned (`cc:262`).
- When the watched root is deleted, the callback thread itself calls `stopStream` (`cc:189-192`), racing a concurrent unsubscribe → double `FSEventStreamRelease`.

No fixed upstream release exists (2.5.6 is latest; this crash class is a long-standing open upstream issue).

Every other native watch already runs in the crash-isolated forked watcher child (#7547/#7757) precisely because of this failure class. The one exception was the darwin-only `.git/worktrees` narrow stream in `worktree-git-common-watch.ts`, which still called `watcher.subscribe()` in the Electron main process — and its own teardown flow (`teardownAndRearm` on `git worktree prune` deleting an empty `worktrees` dir, plus existence-poll re-subscribe) routinely exercises exactly the raciest paths. The crash report confirms the main process: thread 0 is `CrBrowserMain` and `watcher.node` is in the crashed process's binary images. With agents running, worktree add/prune churn made this a background-time app killer.

## Fix

Route the narrow stream through the existing `subscribeViaWatcherProcess` supervisor:

- Callback shape, root-deletion teardown/re-arm, and existence-poll semantics are unchanged.
- `onInterruption` (watcher-child crash → automatic resubscribe) now reports a structural change so the worktree list re-syncs across the event gap.
- Terminal child failures surface through the same error branch that already tears down and re-arms.
- Under Vitest the supervisor's in-process fallback keeps the existing darwin narrow-stream integration tests exercising real FSEvents.

A native watcher fault in this path is now a contained child crash with existing restart machinery (#8661) instead of main-process heap corruption.

## Verification

- New unit suite `worktree-git-common-watch.test.ts`: child routing, event forwarding, root-deletion teardown + existence-poll re-subscribe, error teardown/re-arm, interruption re-sync, dispose hygiene.
- Existing darwin narrow-stream integration tests (`worktree-base-directory-poller.test.ts`) pass through the new routing against real FSEvents; 24 watcher-adjacent test files (217 tests) pass; `typecheck:node` clean.
- Live A/B on macOS: production Orca main process has `watcher.node` mapped (`lsof`); a dev build with this fix has zero `watcher.node` in the main process while the forked watcher child hosts it.
- Live end-to-end: with a worktree created *outside* any watched workspace base dir (so only the `.git/worktrees` stream can see it), external `git worktree add`/`remove` produced `worktrees:changed` in the renderer in ~240–320 ms — far below the 2 s poll interval, proving detection came from the child-hosted native stream.

### DIFF
--- a/src/main/ipc/worktree-git-common-watch.test.ts
+++ b/src/main/ipc/worktree-git-common-watch.test.ts
@@ -128,6 +128,17 @@ describe('worktree git-common narrow watch (darwin)', () => {
     await vi.waitFor(() => {
       expect(subscribeMock).toHaveBeenCalledTimes(2)
     })
+
+    const receivedAfterRearm = received.length
+    childSubscriptions[0].callback(new Error('late error from replaced watcher'), [])
+    childSubscriptions[0].callback(null, [
+      { type: 'create', path: join(worktreesDir, 'late-old-event') }
+    ])
+    childSubscriptions[0].hooks.onInterruption?.()
+
+    // A replaced watch cannot tear down its successor or report stale events.
+    expect(received).toHaveLength(receivedAfterRearm)
+    expect(childSubscriptions[1].unsubscribe).not.toHaveBeenCalled()
   })
 
   it('reports a structural change after a watcher-child interruption', async () => {
--- a/src/main/ipc/worktree-git-common-watch.test.ts
+++ b/src/main/ipc/worktree-git-common-watch.test.ts
@@ -1,0 +1,178 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { mkdtemp, mkdir, realpath, rm } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { subscribeViaWatcherProcess } from './parcel-watcher-process'
+import type {
+  WatcherProcessCallback,
+  WatcherProcessHooks
+} from './parcel-watcher-process-subscription'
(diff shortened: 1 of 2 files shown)


brennanb2025:
@nahuelb you submitted a crash report through our feedback system, this should fix your issue! Try upgrading in about 30 min when the release contains this fix gets rolled out. Thanks!

PR #9094, opened 2026-07-17
fix(updater): stop other and racing app instances from silently killing macOS update installs

brennanb2025:
## Problem

Squirrel.Mac's ShipIt waits for **every** running instance of the installed bundle to exit before applying an update ([`waitForTerminationIfNecessary` → `SQRLTerminationListener`, matched by bundle id + bundle URL](https://github.com/electron/Squirrel.Mac/blob/master/Squirrel/ShipIt-main.m)), and Electron's fork **aborts** with `App Still Running Error` when a matching instance shows up mid-install.

Two real flows hit this (root cause of the "update to v1.4.144-rc.2 doesn't relaunch / stays on the old version" reports; confirmed from ShipIt logs, the updater's install span, and `pmset -g log` on an affected machine):

1. **A second packaged Orca instance is running** — a manually opened copy, or an agent/e2e rig launching `/Applications/Orca.app` with a temp profile (windowless, invisible to the user). "Restart to Update" quits the app, ShipIt waits silently forever, nothing relaunches. Reopening Orca just focuses the old still-running copy, so it reads as "the update didn't take". When the blocking instance finally quits and the install starts, a freshly launched Orca aborts it (observed: 5-minute wait → `App Still Running Error` → stranded on the old version; retry with one instance installed in <1s).
2. **A single-instance user reopens Orca during the install.** ShipIt spends ~4–15s swapping the bundle after the app exits. Click the dock icon in that window and the new process becomes the aborting instance — the user discards their own update without any feedback anywhere.

## Fix

Both ends, macOS-only (the Windows/Linux installers manage running instances themselves), and both fail open so update installs can never be blocked by the guards themselves:

- **Pre-install conflict guard** (`updater-conflicting-app-instances.ts`): before the native install handoff, scan the current user's processes for other pids running this exact executable (mirrors `NSRunningApplication`'s session scope; helpers excluded by path). If any exist, don't start a doomed quit — show an actionable error on the update card naming the PIDs, and record a `quit_and_install_blocked_by_other_instances` lifecycle event so telemetry can measure how often the wild hits this.
- **Relaunch-race launch gate** (`startup/update-install-launch-gate.ts`): the updater drops a version-stamped handoff marker in userData right before the native `quitAndInstall` (cleared on install recovery). At startup — before the single-instance lock and any userData side effects — a launch that finds the marker fresh (≤10 min), is still the **same version** (bundle not yet swapped), and sees this bundle's ShipIt installer alive (argv[0]-anchored process match) exits immediately instead of aborting the install. ShipIt then finishes and relaunches the updated app itself, so the user's click still ends with Orca open — on the new version.

No-lockout guarantees: the post-update relaunch runs a different version → marker cleared, normal start; installer dead (aborted/crashed install) → marker cleared, normal start; stale/corrupt marker, unreadable process table, non-darwin, dev builds → all launch normally.

## Review hardening

- The startup gate now reads the all-user process table so a privileged/root ShipIt is detected for managed or non-user-writable installs. The scan still runs only when a fresh handoff marker exists.
- ShipIt matching requires an argv[0] boundary, avoiding prefix false positives from similarly named executables.
- A blocked handoff is marked install-retryable, so the update card retries the already-staged install instead of downloading the release again, including after the card remounts.

## Testing

- Unit suites for both new modules: conflict parsing (self/helper exclusion, paths with spaces), gate decisions (defer, post-update relaunch, dead installer, stale/corrupt marker, fail-open, platform gates), and an argv[0]-anchor test so a shell merely mentioning the ShipIt path doesn't read as a live installer (caught live).
- Flow specs in `updater.mac-install.test.ts`: blocked handoff surfaces the PID error without invoking the native install or killing PTYs, retry after the conflict clears installs normally, and the handoff marker lands before the native `quitAndInstall` call.
- All 147 tests across the touched suites green; `typecheck:node`, oxlint, oxfmt clean; detection logic validated against the live process table (finds the running packaged app from another pid's perspective, excludes all helper processes).

## Out of scope (follow-ups)

- A "quit other instances and install" action on the update card (needs renderer IPC).
- Plain Cmd+Q with a staged update (`autoInstallOnAppQuit`) installs **without** a relaunch; gating those launches would leave a dead click, so that path needs a detached-relauncher design first.


brennanb2025:
Taking a look!

Issue #8986, opened 2026-07-16
[Bug]: OMP terminal tab agent status flashes; ghost agent in sidebar cannot be opened or deleted


AmethystLiang:
## Summary

Two related OMP / agent-status problems observed in the same session:

1. **OMP agent status on the terminal tab flashes a lot** — the tab badge/status appears to thrash. This may be related to how Orca detects OMP status from the terminal title (OSC / synthetic title path).
2. **Sidebar agent list is out of sync with real tabs** — the left side shows **3 agents**, but only **2** corresponding tabs exist. The extra (ghost) agent **cannot be focused/opened**, and **right-click → delete does not work** on it.

## Environment

- Observed in Orca UI with an **OMP** terminal tab active
- Session also had other terminal tabs (e.g. Pi, other desktop workspaces)
- Left nav included Tasks / Automations / mobile companion areas
- One visible task pane: **Task 1 agent** with a goal prompt (lowcode source repair)

### A. Status flashing on OMP tab

- The **OMP** tab’s agent status indicator flashes repeatedly (on/off or status thrashing).
- Feels like title/status detection is oscillating (e.g. competing OSC title frames vs synthetic agent titles / hook-driven titles).

### B. Ghost agent in sidebar

- Left-side agent list shows **3** agents.
- Only **2** real tabs can be found/opened for those agents.
- The third entry:
  - cannot be accessed / focused to a live tab
  - cannot be removed via right-click delete (context menu/action ineffective or unavailable)

## Expected behavior

1. OMP tab status should settle to a stable working/idle/waiting indicator without continuous flashing under normal agent activity.
2. Sidebar agent entries should map 1:1 to real agent sessions/tabs (or clearly mark dead/orphaned entries).
3. Orphaned / unreachable agent entries should support cleanup (right-click delete / dismiss) so the list doesn’t accumulate ghosts.

## Actual behavior

1. OMP tab status flashes heavily.
2. Sidebar shows an extra agent that has no usable tab.
3. That ghost entry cannot be opened or deleted.

## Screenshots

See follow-up comment for UI captures showing the OMP tab and task/agent context.

## Notes / likely code areas

OMP is treated as pi-compatible for title identity and drives synthetic titles from hooks (including working state):

[...]

PR #9095, opened 2026-07-17
fix(sidebar): dismiss stale agent rows after SSH relay restart (#9030)

### LINKED ISSUE #9030: [Bug]: Sidebar agent rows become permanently unclickable after the remote relay daemon restarts
### Operating system

macOS

### Orca version

_No response_

### Details

After the relay daemon on an SSH host restarts (crash or manual kill), agent rows for that project stay visible in the sidebar but clicking them does nothing — no tab opens, no toast, no console error. They stay like that until the 30-minute status freshness window expires.

Traced it in current main:

- The rows are driven by agentStatusByPaneKey, which is cleared on explicit tab close or worktree sleep/remove but never on relay/daemon teardown, and entries count as fresh for 30 minutes (AGENT_STATUS_STALE_AFTER_MS).
- After the daemon restart, those agents' tabs are gone from tabsByWorktree. A click then lands in the guard in WorktreeCardAgents.tsx (handleActivateAgentTab, ~line 211): tab missing but agentStatusByPaneKey[paneKey].worktreeId matches → plain return, no activation, no dismissal, no feedback.
- Per the comment, the guard keeps rows alive whose tab "hasn't hydrated yet". That's right for an orchestration worker mid-spawn, but after a daemon restart the tab never arrives, so the row is permanently inert. WorktreeCardAgents.activation.test.tsx ("keeps a live worktree-attributed row visible while its tab is hydrating") pins the silent no-op for the transient case; the never-hydrating case has no exit.

Repro:

1. Open a remote SSH project and start an agent in a worktree.
2. On the host, kill the relay daemon (pkill -f relay.js) and let orca reconnect.
3. Click the agent's row in the sidebar — nothing happens.

Upd.
Reloading the window (View → Reload) clears the rows and they don't come back, so the stale entries appear to live only in renderer state — main's status snapshot no longer contains those sessions.

e-jung:
Fixes #9030

## Summary

After an SSH relay daemon restarts (crash or manual kill), agent rows for that project stay visible in the sidebar but clicking them does nothing — no tab opens, no toast, no console error — until the 30-minute status freshness window expires. The rows are stale renderer-memory entries whose terminal tabs are gone for good, but the activation guard silently returned without activating or dismissing them.

This PR makes clicking a stale row give **truthful, actionable feedback** (dismiss + toast) instead of a silent no-op, while preserving the legitimate mid-spawn hydration case for orchestration workers (whose worktree-attributed status can arrive before their terminal tab).

### Root cause
The `WorktreeCardAgents` activation guard kept any worktree-attributed row alive when its tab was missing. That is correct for an orchestration worker mid-spawn, but it also trapped stale rows after a relay restart: `agentStatusByPaneKey` is cleared on tab close / worktree sleep-remove but never on relay/daemon teardown, and a real SSH terminal agent's stored entry retains its `terminalHandle` (main stamps one via `registerPty` + the `ptysById` fallback, and `setAgentStatus` retains it across the restart). So gating on the stored entry's strings keeps the stale SSH remnant too.

### Fix (live runtime signal, not retained strings)
The guard now keeps a worktree-attributed row alive **only while the runtime currently lists the pane as an orchestration allocation** — via the live `runtimeAgentOrchestrationByPaneKey` mirror synced from main's runtime (`buildAgentOrchestrationByPaneKey` → `syncWindowGraph` → renderer store). That mirror is rebuilt from main's current state, so:
- A plain terminal agent is **never** in it (no orchestration dispatch) → a stale row whose tab is gone is dismissed with a toast.
- A currently-running orchestration worker **is** in it → the hydrating row is kept.
- A dead allocation drops out on the next graph sync.

Retained `terminalHandle`/`orchestration` strings on a stored entry are no longer treated as liveness. The fresh-IPC hydration path in `useIpcEvents` still uses `hasRuntimeBackedAttribution` on the just-received payload (a current signal, not retained); the helper was extracted to `src/shared/agent-status-types.ts` and its doc comment scoped accordingly.

## Screenshots

No visual change. Clicking a stale row now dismisses it with the existing transient toast "Agent's pane is no longer available." instead of no-op'ing silently.

## Testing

Rebased onto current `upstream/main` (`1c54171c5`); head is now `127a6e4c3` (was `01e6a6b03`). The rebase was conflict-free (0 behind / 2 ahead); the only upstream commit touching a PR file is `6e91ca6c0` (`useIpcEvents.ts` local-HTTPS handler, a different region from this PR's `hasRuntimeBackedAttribution` extraction).

- [x] `pnpm lint` (changed-file `oxlint`: 0 warnings, 0 errors; `oxfmt` clean)
- [x] `pnpm typecheck` (node, tc.cli, tc.web — all clean; re-verified on the rebased tree)
- [x] `pnpm test` — focused: activation + predicate + `useIpcEvents` = **135 pass** (re-verified on the rebased tree); broader sidebar + agent-status slices = **1,542 pass** on the equivalent pre-rebase logic; no-mistakes pipeline `test` stage 0 findings
- [x] `pnpm build` — full `verify` CI `Build unpacked app` + `Smoke packaged CLI` green on the rebased tree (head `127a6e4c3`)
- [x] Full CI green on the exact rebased head via the fork PR: `verify` (lint/typecheck/test/build/smoke, 15m), golden e2e (linux + mac), native-smoke (ubuntu + windows), roundtrip (all platform × copy/symlink combos), wayland terminal input. (`track-community-pr` is a fork-only check that fails on the fork because the `BUFO_BOT` app secret is absent there; it does not exist on `stablyai/orca`, where the secret is configured.)
- [x] Added/updated high-quality tests:
  - New regression `dismisses a stale SSH relay-restart remnant that still carries a retained terminalHandle` — models the realistic stale shape (retained `terminalHandle`, empty runtime mirror), **fails on the prior `a1351d33e` guard and passes now**.
  - The two hydration-keep tests now set the live `runtimeAgentOrchestrationByPaneKey` mirror and still assert a currently-runtime-backed orchestration worker row is kept.
  - Added a focused `hasRuntimeBackedAttribution` unit test for the fresh-IPC predicate contract.

## AI Review Report

This change went through multiple independent automated reviews: two written reviews on prior heads (`a1351d33e` and `3d20c69c0`) plus a no-mistakes pipeline review. The first review correctly found that gating on the stored entry's retained `terminalHandle` was ineffective for real SSH remnants (the entry retains it across a restart); the remediation re-gated on the live `runtimeAgentOrchestrationByPaneKey` mirror, which the second review verified end-to-end and approved.

The no-mistakes pipeline revalidated the pre-rebase logic at `01e6a6b03` with **0 findings** across review/test/document/lint (review ran 2 rounds). After the rebase onto `upstream/main`, full CI was re-run and is green on the rebased head `127a6e4c3`. The pipeline review additionally surfaced and resolved a **web/remote nuance** (no code change): `runtimeAgentOrchestrationByPaneKey` is always `{}` on web because web's `syncWindowGraph` returns `RuntimeStatus` without that field — but the keep-branch is unreachable on web, since `buildMirroredAgentStatusPatch` derives each row's paneKey from the same `terminalSurfaceTabs` source that produces the terminal tabs, so a row cannot exist without its tab there. The finding is vacuously correct on web and the fix works on its real desktop/SSH target.

Risks checked:
- **Cross-platform (macOS/Linux/Windows):** the change is platform-agnostic renderer data-flow logic; no keyboard shortcuts, paths, shell, or Electron platform behavior touched.
- **Local / SSH / paired-remote:** `buildAgentOrchestrationByPaneKey` iterates both graph leaves and `ptysById` (the latter covers SSH panes via `registerPty`); dispatch resolution is host-agnostic. Behavior is identical for local / WSL / SSH / paired remote. The guard additionally requires `liveEntry?.worktreeId === worktreeId`, so a foreign paneKey cannot keep a row in another worktree's card.
- **Performance:** one cheap map lookup added on user click only; no hot-path impact. The fresh-IPC predicate is unchanged behavior.
- **Accessibility:** a previously silent no-op now yields a truthful transient toast — strictly better feedback.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, dependencies, or IPC surfaces. The fix reuses the existing `dismissStaleAgentRowByKey` path (`dropAgentStatus` + the existing `agentStatus:drop` IPC mirror). The `runtimeAgentOrchestrationByPaneKey` mirror is read-only here. No follow-up security work identified.

## Notes

- **Evidence limitations:** validation was performed on Linux. The regression is a grounded unit reproduction of the renderer state machine using the realistic stale-entry shape (retained `terminalHandle` + empty live mirror) traced through main's enrichment/retention paths. **No live macOS/SSH relay restart reproduction was performed or claimed.** Maintainers may want to confirm end-to-end (start an SSH agent, `pkill -f relay.js`, reconnect, click the row) before merge.
- **Scope:** intentionally limited to the #9030 click-surface lifecycle fix. Rows reconcile on click (truthful feedback) rather than auto-clearing on reconnect; this does not touch #8986 sticky-delete suppression, status-badge work, retry policy, or remote activation authority.
- **Residual timing note:** the runtime mirror is renderer-pull-based (updated on `syncWindowGraph`, coalesced on a frame). Worst case for an orchestration worker is a transient false-dismiss with a truthful toast that self-corrects when the worker's tab hydrates and its next status hook re-renders the row — strictly better than the prior silent no-op, and self-correcting (no data loss, no stuck state).
- X handle: @ericjung

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeCardAgents.activation.test.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeCardAgents.activation.test.tsx
@@ -50,15 +50,18 @@ function mockAgent({
 let mockAgents: DashboardAgentRowData[] = []
 let mockAgentActivityDisplayMode: 'compact' | 'full' | undefined
 let mockTabsByWorktree: Record<string, { id: string }[]> = {}
-// Why: the activation guard keeps a worktree-attributed row alive only while it
-// is runtime-backed (orchestration context or a terminal handle) — the same
-// hydration signal applyAgentStatus uses for snapshot replay. The mock store
-// entry therefore carries those fields so each test can model either a
-// hydrating orchestration worker or a stale non-runtime-backed row faithfully.
+// Why: the activation guard keeps a worktree-attributed row alive only while
+// the runtime CURRENTLY lists the pane as an orchestration allocation — the
+// live runtimeAgentOrchestrationByPaneKey mirror, not the entry's retained
+// terminalHandle/orchestration strings (those survive an SSH relay restart and
+// would make the row silently unclickable again). The mock store carries both
+// the entry shape and the live mirror so each test can model a hydrating
+// orchestration worker or a stale relay-restart remnant faithfully.
 let mockAgentStatusByPaneKey: Record<
   string,
   { worktreeId?: string; orchestration?: unknown; terminalHandle?: string }
 > = {}
+let mockRuntimeAgentOrchestrationByPaneKey: Record<string, unknown> = {}
 let mockActiveTabId: string | null = null
 let mockActiveTabType: string = 'editor'
 const mockSetActiveTab = vi.fn((tabId: string) => {
@@ -82,6 +85,7 @@ function buildMockStoreState(): Record<string, unknown> {
     acknowledgeAgents: vi.fn(),
     agentSendPopoverTargetMode: null,
     agentStatusByPaneKey: mockAgentStatusByPaneKey,
+    runtimeAgentOrchestrationByPaneKey: mockRuntimeAgentOrchestrationByPaneKey,
     agentStatusEpoch: 0,
     activeTabId: mockActiveTabId,
     activeTabType: mockActiveTabType,
@@ -163,6 +167,7 @@ describe('WorktreeCardAgents activation', () => {
     mockAgentActivityDisplayMode = undefined
     mockTabsByWorktree = {}
     mockAgentStatusByPaneKey = {}
+    mockRuntimeAgentOrchestrationByPaneKey = {}
     mockActiveTabId = null
     mockActiveTabType = 'editor'
     capturedRowActivations = []
@@ -255,12 +260,15 @@ describe('WorktreeCardAgents activation', () => {
         worktreeId: 'wt-1'
       })
     ]
-    // Why: a hydrating orchestration worker is runtime-backed (it carries an
-    // orchestration dispatch context), which is the signal that distinguishes
-    // legitimate mid-spawn hydration from a stale relay-restart remnant.
+    // Why: a hydrating orchestration worker is one the runtime currently lists
+    // as an orchestration allocation (the live runtimeAgentOrchestrationByPaneKey
+    // mirror). That current-allocation signal — not the entry's retained
+    // strings — is what distinguishes legitimate mid-spawn hydration from a
+    // stale relay-restart remnant.
     mockAgentStatusByPaneKey = {
       [paneKey]: { worktreeId: 'wt-1', orchestration: { taskId: 't-1' } }
     }
+    mockRuntimeAgentOrchestrationByPaneKey = { [paneKey]: { taskId: 't-1' } }
     const { default: WorktreeCardAgents } = await import('./WorktreeCardAgents')
 
     renderToStaticMarkup(<WorktreeCardAgents worktreeId="wt-1" />)
@@ -289,6 +297,7 @@ describe('WorktreeCardAgents activation', () => {
     mockAgentStatusByPaneKey = {
       [paneKey]: { worktreeId: 'wt-1', orchestration: { taskId: 't-1' } }
     }
+    mockRuntimeAgentOrchestrationByPaneKey = { [paneKey]: { taskId: 't-1' } }
     // Why: activation may create/select a different terminal before the
     // automation worker hydrates; the row must only pane-focus its exact tab.
     activationMocks.activateAndRevealWorktree.mockImplementation(() => {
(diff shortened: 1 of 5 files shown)


nwparker:
Thanks so much for this, @ericjung — genuinely thorough work, and the live-`runtimeAgentOrchestrationByPaneKey` insight is sharp.

For #9030 we landed #9878, which fixes it at the **data-lifecycle** layer: on relay/daemon teardown the renderer now clears the stale rows directly (converging to main's already-clean state — the same end result as a window reload), so the dead rows *disappear* rather than lingering until clicked, and it covers every consumer of `agentStatusByPaneKey`, not just the sidebar click path.

Your core point — that the activation guard shouldn't treat a retained `terminalHandle` as liveness and should consult the live runtime mirror instead — is valuable hardening *beyond* the relay-restart case, and we'd love to land it as a focused follow-up with credit to you. Thank you again for the careful analysis and tests! 🙏

PR #8564, opened 2026-07-13
[Feature]: open markdown file shortcut does nothing in main window

VitorRibeiroCustodio:
## Summary
- Cmd/Ctrl+Shift+O (`tab.openMarkdown`) only had a handler inside the Floating Terminal panel's own keydown listener, so pressing it with a normal worktree tab focused did nothing.
- Adds a main-window path: an IPC picker rooted at the active worktree, a store action to open the picked file, and a keydown branch in Terminal.tsx — mirroring the existing `tab.newMarkdown` wiring.

## Test plan
- [x] `pnpm typecheck` passes
- [x] `oxlint` clean on all touched files
- [x] New unit tests: `src/main/ipc/app.test.ts` (picker IPC handler), `src/renderer/src/store/slices/open-markdown-file.test.ts` (store action)
- [x] Manual: Cmd/Ctrl+Shift+O in a normal worktree tab opens a markdown file picker and loads the selected file into a tab
- [x] Manual: Cmd/Ctrl+Shift+O still works as before when the Floating Terminal panel is focused

Fixes #8532

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## ELI5

The shortcut to open a markdown file (Cmd/Ctrl+Shift+O) only worked inside the floating terminal, so it did nothing in a normal worktree tab. It now opens a file picker for the active worktree from the main window, same idea as new-markdown.

PR #9099, opened 2026-07-17
fix: Cmd/Ctrl+Shift+O opens markdown picker in main window

### LINKED ISSUE #8532: [Feat]: Cmd+Shift+O (Open markdown tab) does nothing in the main window — only works in Floating Terminal
Operating system: macOS
Orca version: v1.4.137

Details:

Short summary:
The "Open markdown tab" command (`tab.openMarkdown`, default `Cmd+Shift+O`) has no effect when focus is in the main window. It only works inside the Floating Terminal panel (`Cmd+Alt+A`). "New markdown tab" (`Cmd+Shift+M`) works fine in the main window, so the two behave inconsistently even though both appear the same in Settings → Shortcuts.

What happened?
Pressing `Cmd+Shift+O` in a normal worktree does nothing — no file picker, no tab. The same shortcut works only when the Floating Terminal panel is focused.

How can we reproduce it?
1. Open Orca on a normal worktree (main window, not the floating terminal).
2. Press `Cmd+Shift+O`.
3. Nothing happens (no native "Open markdown" file picker, no new tab).
4. Open the Floating Terminal (`Cmd+Alt+A`), focus it, and press `Cmd+Shift+O`.
5. The native "Open markdown" file picker now appears as expected.

Expected:
`Cmd+Shift+O` should open the markdown file picker in the main window too — consistent with `Cmd+Shift+M` (New markdown), which does work in the main window. (Alternatively, if it is intentionally floating-only, it probably shouldn't be presented as a general shortcut in Settings.)

Likely root cause (from inspecting the packaged app.asar, v1.4.137 — line up against the TS source):
- `tab.openMarkdown` appears to be wired only through `FLOATING_WORKSPACE_PANEL_SHORTCUT_ACTIONS`, so it fires only when the key event target is inside the floating panel.
- The main-process `before-input-event` handler dispatches `tab.newMarkdown` → `ui:newMarkdownTab`, `tab.newTerminal` → `ui:newTerminalTab`, `tab.newBrowser` → `ui:newBrowserTab`, etc., but has **no branch for `tab.openMarkdown`** (there is no `ui:openMarkdownTab`).
- The renderer exposes an `onNewMarkdownTab` IPC listener but no `onOpenMarkdownTab`, and the renderer's central keydown handler has no markdown `matchShortcut`.
- Net effect: in the main window the `Cmd+Shift+O` keystroke is never routed to an "open markdown" action.

Workaround:
Use `Cmd+P` (Go to File) to open a markdown file, or open it from the Floating Terminal panel.

nwparker:
## Description
`tab.openMarkdown` (`Cmd/Ctrl+Shift+O`) only worked inside the Floating Terminal panel: the action lived in the floating shortcut list and floating keydown path, but the main-window Terminal keydown handler and browser-guest forwarder never wired it.

This takes over / builds on #8564 by @VitorRibeiroCustodio:

- Main-window keydown + tab-bar **Open Markdown…** menu call a new store action that opens a native markdown picker rooted at the active worktree, then `openFile`s the selection (mirrors `tab.newMarkdown`).
- Shared IPC `app:pickWorktreeMarkdownDocument` reuses the floating picker helper (filters, authorize, markdown name check).
- Floating workspace still works via the panel’s own handler and Terminal’s floating-focused branch (`openFloatingWorkspaceMarkdownTab`).
- Remote (SSH) worktrees are guarded with a toast — local Electron dialogs cannot browse remote FS.
- **Completeness beyond #8564:** browser guests forward `ui:openMarkdownTab` (same gap the issue called out for `ui:newMarkdownTab`), handled in `useIpcEvents`.

## Evidence
- Confirmed gap on main: Terminal had `tab.newMarkdown` only; no `tab.openMarkdown` / no `onOpenFileTab` wiring; guest UI had `ui:newMarkdownTab` only.
- Vitest (`config/vitest.config.ts`):
  - `src/main/ipc/app.test.ts` — picker IPC (pick / cancel / non-md reject)
  - `src/renderer/src/store/slices/open-markdown-file.test.ts` — open / cancel / no worktree / remote guard / reject toast
  - `src/main/browser/browser-guest-ui.test.ts` — guest `Cmd/Ctrl+Shift+O` → `ui:openMarkdownTab`
- oxlint clean on touched files

## ELI5
The “open markdown” keyboard shortcut was only plugged into the floating mini-window. We plugged it into the normal app window (and browser tabs) the same way “new markdown” already was, so the shortcut opens the file picker everywhere you’d expect.

## User-regression-tradeoffs
- **Local main window:** shortcut + tab menu open the picker (intended fix).
- **Floating terminal:** unchanged behavior (still opens floating markdown picker).
- **Browser guest focus:** shortcut now works (was broken like main window).
- **SSH/remote worktrees:** toast instead of local-only dialog (avoids opening a wrong-host file). Use file explorer / quick open for remote markdown until a remote-aware picker exists.
- **Linux/Windows:** uses `Mod` → Ctrl; same as other tab shortcuts. No new platform-specific binding.
- **terminal-first policy:** same as `tab.newMarkdown` — not captured when xterm owns focus under terminal-first (by design for tab-scoped shortcuts).
- Does not add a main-menu accelerator; matches existing `tab.newMarkdown` pattern (renderer + guest IPC).

Fixes #8532

Credit: core main-window picker path from #8564 (@VitorRibeiroCustodio).

### DIFF
--- a/src/main/browser/browser-guest-ui.test.ts
+++ b/src/main/browser/browser-guest-ui.test.ts
@@ -689,6 +689,23 @@ describe('setupGuestShortcutForwarding', () => {
     expect(rendererSendMock).not.toHaveBeenCalled()
   })
 
+  it('forwards open-markdown shortcut from focused guest pages', () => {
+    setupGuestShortcutForwarding({
+      browserTabId,
+      guest: makeGuest(),
+      resolveRenderer: () => makeRenderer()
+    })
+
+    const preventDefault = triggerBeforeInput({
+      code: 'KeyO',
+      key: 'o',
+      shift: true
+    })
+
+    expect(preventDefault).toHaveBeenCalledTimes(1)
+    expect(rendererSendMock).toHaveBeenCalledWith('ui:openMarkdownTab')
+  })
+
   it('forwards double-tap window shortcuts from focused guest pages', () => {
     setupGuestShortcutForwarding({
       browserTabId,
--- a/src/main/browser/browser-guest-ui.ts
+++ b/src/main/browser/browser-guest-ui.ts
@@ -418,6 +418,10 @@ export function setupGuestShortcutForwarding(args: {
       renderer.send('ui:newSimulatorTab')
     } else if (keybindingMatchesAction('tab.newMarkdown', input, process.platform, keybindings)) {
       renderer.send('ui:newMarkdownTab')
+    } else if (keybindingMatchesAction('tab.openMarkdown', input, process.platform, keybindings)) {
+      // Why: Cmd/Ctrl+Shift+O opens a markdown picker in the active surface.
+      // Guest pages never hit Terminal.tsx keydown, so forward to the renderer.
+      renderer.send('ui:openMarkdownTab')
     } else if (keybindingMatchesAction('tab.newTerminal', input, process.platform, keybindings)) {
       // Why: Cmd/Ctrl+T opens a terminal in the user's active terminal surface
       // even when focus is inside a browser guest. Cmd/Ctrl+Shift+B is the
--- a/src/main/ipc/app.test.ts
+++ b/src/main/ipc/app.test.ts
@@ -356,4 +356,49 @@ describe('registerAppHandlers', () => {
     })
     expect(grantFloatingWorkspaceDirectoryMock).toHaveBeenCalledWith(store, '/Users/kaylee/notes')
   })
+
+  it('picks a markdown file rooted at the given worktree cwd', async () => {
+    showOpenDialogMock.mockResolvedValue({
+      canceled: false,
+      filePaths: ['/Users/kaylee/repo/docs/notes.md']
+    })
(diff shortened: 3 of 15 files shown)


nwparker:
Closing in favor of **#9666**, a fresh re-implementation of this same fix (`Cmd/Ctrl+Shift+O` → Open Markdown in the main window).

This PR (#9099) is from July 17 and its CI (`verify`) is failing. #9666 is a smaller, reviewed re-do: it generalizes the existing floating markdown picker (keeping the old IPC channel as an alias), wires the shortcut + "+" menu into the main window, and — importantly — **gates the action to local worktrees** so a remote/SSH worktree can't have a local file opened into it (the main regression risk a naïve fix would introduce). It went through 3 independent review rounds (final: clean) and a perf audit (no regression).

Superseded by #9666.

PR #8339, opened 2026-07-12
fix(terminal): open hard-wrapped HTTP links exactly once

ryush00:
## Summary

- reconstruct complete HTTP URLs from hard-wrapped terminal rows before opening them
- coordinate the WebLinksAddon and logical-line fallback so one mouse event opens one URL
- temporarily keep modifier-click gestures over detected HTTP links out of mouse-aware child TUIs, restoring the prior xterm option on mouseup, blur, or disposal

This makes a single modifier-click open the complete URL exactly once instead of opening both a physical-row fragment and the full logical URL.

## Screenshots

No visual change.

## Reproduction

1. Start this localhost-only hit counter:

```bash
python3 <<'PY'
import http.server, json, time
from pathlib import Path

hits = []
class H(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        hits.append({
            "path": self.path,
            "t": time.time(),
            "ua": self.headers.get("User-Agent", ""),
        })
        Path("/tmp/orca-wrap-url-hits.json").write_text(json.dumps(hits, indent=2))
        self.send_response(200)
        self.end_headers()
        self.wfile.write(b"ok")

    def log_message(self, *args):
        pass

Path("/tmp/orca-wrap-url-hits.json").write_text("[]")
print("listening http://127.0.0.1:8765/")
http.server.HTTPServer(("127.0.0.1", 8765), H).serve_forever()
PY
```

The server records each request's path, timestamp, and User-Agent in `/tmp/orca-wrap-url-hits.json`.

2. Print the URL below in a narrow Orca terminal so it spans multiple rows. The same URL can also be printed inside a cursor-positioned, mouse-aware agent TUI such as Grok.
3. Shift+Cmd-click a row on macOS, or Shift+Ctrl-click on Windows/Linux.
4. Inspect `/tmp/orca-wrap-url-hits.json`, excluding favicon requests when counting meaningful hits.

[...]

PR #9100, opened 2026-07-17
fix(terminal): stop gluing the next line onto a modifier-clicked URL

### LINKED ISSUE #8832: [Bug]: Cmd-click on a URL opens it with the next line's text glued on
### Operating system

macOS

### Orca version

v1.4.141

### Details

Cmd-clicking a URL in the terminal opens the browser with the first word of the next line appended.

Repro — run this in an Orca terminal, then Cmd-click the URL in the output:

    ▲ echo "Repo: https://github.com/stablyai/orca/\nDescription: 123"
    Repo: https://github.com/stablyai/orca/
    Description: 123

Cmd-click on the URL opens `https://github.com/stablyai/orca/Description`.
Expected: opens exactly `https://github.com/stablyai/orca/`.

`Description` is the first word of the next line — a separate logical line, not a continuation of the URL, but the link detector joins them.

Looks like a false positive of the hard-wrapped-URL reconstruction from #8339: it joins a URL split across rows, and here it treats an unrelated adjacent line as a continuation. Before the update to the current version the same output opened the correct URL.

Basically cmd+click becomes unusable now

nwparker:
## Summary

Fixes #8832.

Takes over / combines the best of #8852 (@xianjianlf2): stop path hard-wrap candidates from poisoning HTTP hit-testing, and reconstruct genuine edge-wrapped HTTP URLs safely.

Cmd-clicking a URL that ends mid-row opened it with the first word of the next line appended:

```
Repo: https://github.com/stablyai/orca/
Description: 123
```

opened `https://github.com/stablyai/orca/Description` instead of `https://github.com/stablyai/orca/`.

## Root cause

`findHttpLinkAtBufferPosition` built logical lines from soft-wrap, framed hard-wrap HTTP, **and** path hard-wrap reconstruction (`buildCandidateLogicalLinesForBufferPosition` / #8339).

Path hard-wrap treats a URL suffix ending in `/` as a path start and joins the next row's path-like prefix (`Description:`). HTTP extraction then trims trailing `:` → `…/orca/Description`. File links never showed this because they verify path existence; HTTP links had no equivalent check.

## Fix

HTTP hit-testing no longer consumes path hard-wrap candidates. It uses only:

1. **Native soft-wrap** (`isWrapped`) — long terminal-wrapped URLs
2. **Framed TUI hard-wrap HTTP** — existing frame-scoped builder
3. **Edge-wrap HTTP** (`buildEdgeWrappedHttpLogicalLineCandidates`) — join without wrap metadata only when each earlier row reaches the terminal edge (one-cell wide-glyph tolerance), with label-row / adjacent-scheme guards and a row scan bound derived from `TERMINAL_HTTP_URL_MAX_LENGTH`

Credit: core approach and edge-wrap reconstruction from @xianjianlf2 in #8852 (including the length-derived row bound). This PR takes that over with a permanent #8832 regression suite and adversarial coverage.

## Evidence

```bash
  src/renderer/src/components/terminal-pane/repro-8832-url-next-line.test.ts \
  src/renderer/src/components/terminal-pane/edge-wrapped-terminal-http-links.test.ts \
  src/renderer/src/components/terminal-pane/terminal-url-link-click.test.ts

# Full pane suite
pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/
```

**Results:** 170 files / 2176 tests passed (focused: repro 12, edge-wrap 18, click 13).

Covered cases:
- Issue repro (mid-row URL + `Description:` next line) → unglued URL
- Path candidates still glue in isolation (documents why HTTP must not consume them)
- Chinese / Windows / POSIX / relative / UNC next-line labels not glued
- Soft-wrapped multi-line URLs still join from either row
- Edge-wrapped multi-line URLs (no wrap metadata) still join from either row
- Wide-glyph / multi-code-point continuation rows
- Edge URL + label rows (`Description:`, `HTTP/2: 200`, etc.)
- Adjacent URL on next row stays separate
- `file://` is not treated as HTTP
- Same-row Windows path does not expand the HTTP link
- Narrow-terminal URLs spanning >20 rows (length-derived bound)
- Chinese path continuation on a genuine edge wrap

## ELI5

When you Cmd/Ctrl-click a web link in the terminal, Orca sometimes “stitched” the next line onto it because a helper meant for **file paths** was also used for **web links**. We stopped using that file-path helper for web links. Real long links that wrap at the screen edge still work; random labels on the next line no longer get glued on.

## User-regression-tradeoffs

| Case | Before | After |
|------|--------|-------|
| Mid-row URL + next-line label (`Description:`) | glued | correct single-row URL |
| Soft-wrapped long URL (`isWrapped`) | works | still works |
| Edge-wrapped long URL (no wrap metadata) | sometimes via path glue | works via dedicated edge builder |
| Framed TUI hard-wrapped URL | works | still works |
| File path hard-wrap links | works (existence check) | untouched |
| Full-width URL + next-line path-like token that is not a label | could glue | still may join if earlier row reaches the edge and next row is not a recognized label/scheme — inherent ambiguity without wrap metadata; prefer soft-wrap when the terminal provides it |
| OSC 8 hyperlinks | separate routing | untouched |

**Supersedes:** #8852 (please close in favor of this PR once reviewed).

### DIFF
--- a/src/renderer/src/components/terminal-pane/edge-wrapped-terminal-http-links.test.ts
+++ b/src/renderer/src/components/terminal-pane/edge-wrapped-terminal-http-links.test.ts
@@ -1,0 +1,425 @@
+import type { IBufferLine } from '@xterm/xterm'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { registerHttpLinkStoreAccessor } from '@/lib/http-link-routing'
+import { openHttpLinkAtBufferPosition } from './terminal-url-link-hit-testing'
+
+const COLS = 40
+const WIDE_GLYPH = '界'
+const openUrlMock = vi.fn()
+
+function makeBufferLine(content: string, cols = COLS): IBufferLine {
+  return makeBufferLineFromCells(
+    Array.from(content, (text) => ({ text, width: text === WIDE_GLYPH ? 2 : 1 })),
+    cols
+  )
+}
+
+function makeBufferLineFromCells(
+  cells: { text: string; width: number }[],
+  cols = COLS
+): IBufferLine {
+  const columns: number[] = []
+  let column = 0
+  let text = ''
+  for (const cell of cells) {
+    text += cell.text
+    for (let index = 0; index < cell.text.length; index++) {
+      columns.push(column)
+    }
+    column += cell.width
+  }
+  while (column < cols) {
+    text += ' '
+    columns.push(column)
+    column++
+  }
+  columns.push(column)
+
+  return {
+    isWrapped: false,
+    length: cols,
+    translateToString: (
+      _trimRight?: boolean,
+      startColumn = 0,
+      endColumn = text.length,
+      outColumns?: number[]
+    ) => {
+      outColumns?.splice(0, outColumns.length, ...columns.slice(startColumn, endColumn + 1))
+      return text.slice(startColumn, endColumn)
+    }
+  } as IBufferLine
+}
+
+describe('edge-wrapped terminal HTTP links', () => {
+  beforeEach(() => {
+    vi.stubGlobal('window', { api: { shell: { openUrl: openUrlMock } } })
+    registerHttpLinkStoreAccessor(() => ({
+      settings: { openLinksInApp: false },
+      setActiveWorktree: vi.fn(),
+      createBrowserTab: vi.fn()
+    }))
+    openUrlMock.mockReset()
+  })
+
+  afterEach(() => {
+    vi.unstubAllGlobals()
+  })
+
+  it('opens a URL when a wide glyph starts the continuation row', () => {
+    const prefix = 'https://example.com/wide/'
+    const firstRow = `${prefix}${'c'.repeat(COLS - 1 - prefix.length)}`
+    const continuationRow = `${WIDE_GLYPH}tail`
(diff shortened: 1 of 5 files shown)


nwparker:
## Final review (ready to merge)

**Correctness**
- Root cause fix is sound: HTTP hit-testing no longer consumes path hard-wrap candidates (`buildCandidateLogicalLinesForBufferPosition`), which was gluing mid-row URLs to next-line labels (`Description:` → `…/orca/Description`).
- Soft-wrap, framed hard-wrap, and dedicated edge-wrap paths cover the real multi-line cases; issue repro + adversarial suite look solid.

**/perf**
- `findHttpLinkAtBufferPosition` only runs on Cmd/Ctrl-click activation paths (`mousedown` suppression + `mouseup` open), not hover/mousemove or render.
- Edge builder bounds the row scan by `TERMINAL_HTTP_URL_MAX_LENGTH` (2048) with a width-2-aware row cap and per-line translation cache — acceptable for the click path; no continuous polling/hot-path regression.

**CI / mergeability**
- `verify` + `track-community-pr` green
- `MERGEABLE` / `CLEAN`
- CodeRabbit nit only (exact 2048 boundary tests) — optional follow-up, not a blocker

Merging and closing superseded #8852 (credit retained for @xianjianlf2).

Issue #958, opened 2026-04-22
[Feature]: Easily import Ghostty configs


AmethystLiang:
### Problem or use case

Reported by X user https://x.com/mralexsaavedra/status/2047048144626516089. 

### Proposed solution

Right now user have to manually select theme & divider width and other Ghostty configs. 
We can help import them directly

mralexsaavedra:
I’m working on this and have a first implementation on a feature branch.

Current scope:
- safe one-shot Ghostty import with preview/confirmation
- Ghostty config path discovery fixes (XDG + legacy paths)
- partial import support for settings Orca can map safely

I’ll open a PR once I finish polishing and validating the current behavior.

mralexsaavedra:
PR is open: https://github.com/stablyai/orca/pull/1001

Branch `feat/958-ghostty-import` is ready for review. It includes:
- Safe one-shot Ghostty config import with preview/confirmation modal
- Cross-platform config path discovery (macOS, Linux XDG, Windows)
- 17 supported Ghostty keys mapped to Orca settings
- New **Window** section in Terminal Settings to view/edit all imported values manually

mralexsaavedra:
**Final state: PR ready for review → https://github.com/stablyai/orca/pull/1001**

**27 Ghostty keys supported:**

| Ghostty key | Orca setting |
|---|---|
| `font-family` | `terminalFontFamily` |
| `font-size` | `terminalFontSize` |
| `font-weight` | `terminalFontWeight` |
| `cursor-style` | `terminalCursorStyle` |
| `cursor-style-blink` | `terminalCursorBlink` |
| `cursor-text` | `terminalColorOverrides.cursorAccent` |
| `cursor-opacity` | `terminalCursorOpacity` |
| `focus-follows-mouse` | `terminalFocusFollowsMouse` |
| `macos-option-as-alt` | `terminalMacOptionAsAlt` |
| `background-opacity` | `terminalBackgroundOpacity` |
| `background-blur-radius` | `windowBackgroundBlur` (vibrancy macOS / acrylic Windows — restart required) |
| `background` / `foreground` | `terminalColorOverrides.background/foreground` |
| `cursor-color` | `terminalColorOverrides.cursor` |
| `selection-background/foreground` | `terminalColorOverrides.selectionBackground/Foreground` |
| `palette` (0–15) | `terminalColorOverrides` ANSI fields |
| `bold-color` | `terminalColorOverrides.bold` |
[...]

PR #9105, opened 2026-07-17
fix(terminal): honor Bold Text color override (#8595)

### LINKED ISSUE #8595: [Bug]: "Bold Text" terminal color override is a no-op in the renderer

nwparker:
## Description

Settings → Appearance → Terminal **Bold Text** (and Ghostty `bold-color` via #958) was persisted as `terminalColorOverrides.bold` and spread into `options.theme`, but upstream xterm `ITheme` had no `bold` key — bold cells with the default foreground always painted as `foreground`.

This pins `@xterm/xterm` and `@xterm/addon-webgl` so `theme.bold` is:

- resolved in ThemeService (`bold` falls back to `foreground` when unset)
- used as the default foreground for **bold, non-inverse** cells in DOM and WebGL
- ignored for explicit ANSI/RGB foregrounds (still use palette / `drawBoldTextInBrightColors`)

**Vs #8952 (@rodboev):** Takes the same core approach (ThemeService + WebGL atlas/equality + e2e screenshot path). Improves the DOM path: when `minimumContrastRatio` is already satisfied, `_applyMinimumContrast` returns false and default cells inherit the row container color — so bold overrides that already have good contrast never painted. We set an inline `color: colors.bold` in that case. Adds fast unit regression coverage on the pinned package surface + compose path.

Credit: @rodboev for PR #8952.

Fixes #8595

## Evidence

- Unit: `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/repro-8595-bold-theme-noop.test.ts src/renderer/src/components/terminal-pane/compose-active-terminal-theme.test.ts src/main/ghostty/mapper-extended.test.ts` → 56 passed
- `pnpm install --frozen-lockfile` with both patches applied
- `oxlint` clean on touched TS sources
- E2E: `tests/e2e/terminal-bold-default-foreground.spec.ts` (DOM + WebGL; add/change/remove override; keeps normal FG + explicit bold red). Not run in this session (full Electron); intended for CI / headed proof.

## ELI5

Orca let you pick a special color for **bold** terminal text, but the drawing engine threw that color away. We taught the engine to use it for bold text that doesn’t already have its own color, while leaving red/green/etc. bold text alone.

## User-regression-tradeoffs

| Area | Risk | Mitigation |
|------|------|------------|
| Explicit ANSI bold (`\e[1;31m`) | Low — must stay palette-colored | WebGL/DOM still use ANSI/RGB branches; e2e asserts red row remains |
| Inverse cells | Low — must not pick bold FG | Inverse still swaps to background / inverted default class |
| No bold override set | Low — should match old look | ThemeService falls back `bold → foreground` |
| DOM min-contrast | Medium — new inline style path | Only when bold + default FG + contrast path did not style; high-contrast overrides now visible (the bug) |
| xterm upgrade | Medium — pin maintenance | Patch is small, source-level hunks document intent; unit tests assert patch strings |
| Performance | Negligible | One color parse + atlas equality field |
| Ghostty import | Low | Existing `bold-color` mapper tests unchanged |

No intentional product tradeoff beyond making a previously no-op setting work as labeled.


nwparker:
Deferring this local xterm patch for now and returning the PR to draft.

Upstream tracking:
- https://github.com/xtermjs/xterm.js/issues/6032 — proposed dedicated bold theme color
- https://github.com/xtermjs/xterm.js/issues/5774 — current default-foreground behavior is by design

For users: Orca’s **Bold Text** override and imported Ghostty `bold-color` remain unsupported/no-op for bold text using the default foreground. We will revisit this PR if upstream support lands or we decide the local patch maintenance cost is justified.

nwparker:
Superseded by #9661 (merged). Same issue (#8595); this PR took the "actually implement bold-color" route by patching the vendored xterm bundles, which carries an unset-bold / `OSC 10/110` stale-foreground rendering-regression risk plus the maintainer-deferred xterm-patch maintenance burden. #9661 ships the zero-regression honesty fix instead (truthful Settings + Ghostty-importer copy, persisted values preserved). If we later decide to take the xterm patch, this and #8952 are the reference implementations.

Issue #8459, opened 2026-07-13
[Bug]: Resource Manager misclassifies live daemon sessions as orphan and force-kills them without confirmation


bbingz:
## Summary

On macOS with Orca 1.4.137, Resource Manager classified actively running daemon-backed agent terminals as "orphans". Clicking **Kill orphan terminals** then force-killed all of them without a confirmation dialog.

This incident followed the headless-serve / GUI relaunch lifecycle mismatch documented in #8457, but the destructive false-positive cleanup is an independent safety bug: any incomplete, stale, or recovering renderer store can make valid daemon sessions look unbound.

## Observed incident (2026-07-13, Asia/Shanghai)

1. A packaged `orca serve` process still owned the single-instance runtime and live AI terminals after the GUI had quit.
2. A GUI was forced back into view while the runtime/renderer lifecycle was inconsistent (see #8457).
3. Resource Manager displayed **25 orphans**, including workspaces with agents that were visibly still running.
4. The user clicked **Kill 25 orphan terminals**.
5. `daemon.log` then recorded **28 distinct `session-killed` events with `immediate:true` in ~50 ms** (`02:48:41.072Z`–`02:48:41.122Z`). The count increased from the screenshot's 25 because additional sessions were created between the screenshot and the kill.
6. Live agent sessions were interrupted and some UI/session state churned afterward.

The Electron main process did **not** restart at the moment of the bulk kill: its PID and start time remained unchanged, and `main.trace.ndjson` contains no new `renderer_bootstrap_started` after cleanup. The visual "window restart" therefore appears to be UI/session reconstruction after the forced kills, not a second confirmed main-process restart.

I am not attaching the screenshot because it contains private workspace names, but the daemon event counts and timestamps above are deterministic and redacted.

## Root cause in current `main`

"Orphan" currently means only "not referenced by the current renderer's binding index":

- [`resource-session-bindings.ts`](https://github.com/stablyai/orca/blob/4a76a1c894bf8f7258b05f56ae1bebac43e273c0/src/renderer/src/components/status-bar/resource-session-bindings.ts#L28-L85) builds `boundPtyIds` exclusively from renderer-owned tab/PTY/layout state, then counts every daemon session absent from that set as unbound.
- This is not authoritative proof that the daemon session has no owner, no foreground process, or no persisted/restorable binding. A recovering renderer can legitimately have an incomplete index.

The destructive path trusts that derived label without revalidation:

[...]

PR #9114, opened 2026-07-17
fix(terminal): tear down runtime-owned headless PTYs on whole-tab close

brennanb2025:
## What

Fixes a durability regression introduced by #8958 ("make whole-tab close durable"): closing the whole parent of a **runtime-owned headless terminal tab** can leak its PTY and let the "closed" tab reappear when a renderer notifier is attached.

The final routing also preserves the renderer transaction for an adopted serve/SSH tab. That matters because live renderer pin state can be newer than the deliberately debounced workspace-session copy; bypassing the renderer for every runtime-owned PTY could kill a newly pinned terminal from stale persisted metadata.

## Root cause

`closeMobileSessionTab` delegated every whole-parent close to `closeTerminalTab` before the pre-existing authoritative headless teardown. Production wires that relay whenever a window exists.

The relay resolves targets only from renderer state. A runtime-owned headless tab is absent there, so the renderer acknowledged a no-op as success. The runtime left the SSH/serve PTY and persisted binding alive, and `syncMobileSessionTabs` published the tab again.

Simply moving the full runtime-owned classifier ahead of the relay creates a second bug for **adopted** tabs: it skips the renderer's live pin guard and acknowledged persistence transaction. Renderer graph publication is coalesced at 16 ms and workspace-session persistence is debounced at 150 ms, so that stale-pin window is deterministic.

## Fix

- For a whole parent absent from the runtime's renderer graph, perform authoritative runtime teardown before the relay: prune persistence, kill the exact owned PTY(s), publish the pruned snapshot, flush, and best-effort close any racing adopted pane.
- For an adopted parent, retain #8958's renderer request/reply transaction so live pin state can reject close and success still means the renderer state was durably flushed.
- Keep the authoritative fallback for notifier implementations that expose only raw pane close.
- Contain advisory renderer-notification failures so they cannot skip the authoritative session flush or turn an already-committed close into a failed receipt; emit a targeted warning for diagnostics.
- Preserve the existing split-leaf path, so closing one leaf never retires its siblings.

## Reliability contract

- **Invariant (`terminal-session.explicit-close-retirement`):** a whole-parent close of an unadopted runtime-owned terminal removes its exact PTY and durable resume authority; an adopted tab is closed only through the renderer's live guard and durable transaction.
- **Failure source:** #8958 placed a renderer-only resolver ahead of runtime headless teardown; review then found stale persisted pin state could make an unconditional reorder kill an adopted pinned tab.
- **Oracle:** with both production notifiers wired, an absent SSH parent never calls the renderer relay, kills its exact relay PTY, and removes its tab/layout persistence. An adopted serve parent whose renderer rejects `terminal_tab_pinned` calls the relay, propagates the rejection, kills nothing, and preserves persistence.
- **Gates:** `runtime.headless-desktop-promotion-continuity` and `terminal-session.explicit-close-retirement` from `config/reliability-gates.jsonc`.
- **Diagnostics:** close failures continue to surface through the existing RPC error boundary (`terminal_tab_pinned`, renderer timeout/unavailable, persistence errors); a targeted warning records advisory renderer cleanup failures, and deterministic tests record exact relay, PTY-kill, flush, and persisted-layout counts.

## Provider / platform coverage

- **Local + daemon:** adopted renderer ownership remains on the existing durable close path; provider/retirement and parked-close gates pass.
- **SSH/remote:** deterministic coverage uses an exact `ssh:<connection>@@<relay>` identity and proves kill + de-persist. Live remote-host process absence remains uncollected.
- **Serve/headless:** deterministic absent/adopted routing passes; the macOS headless-serve promotion journey preserves the original daemon PTY.
- **Mobile/relay:** `closeMobileSessionTab` is the exercised entry point; no desktop-only filesystem or process assumption was added.
- **macOS/Linux/Windows/WSL:** the routing change is platform-independent TypeScript with no path, shell, signal, or Git behavior. macOS has live Electron evidence; Linux, Windows, and WSL live close evidence remain gaps.

## Performance budget

- **User-visible symptom/resource:** a close appeared successful while an SSH/serve PTY, process, and durable binding remained allocated.
- **Measurement:** deterministic exact kill/persistence/relay counts, the full runtime suite, registered teardown fanout gates, and two live Electron lifecycle journeys.
- **Cost after fix:** an ordinary adopted whole-tab close adds one constant-time `Map.has` ownership check before the existing transaction. It does not run the broader headless classifier, add polling, call `listSessions`, add provider/subprocess fanout, or add background work. Authoritative pruning remains user-triggered and uses the existing bounded snapshot teardown.

## Verification

- Fail-first review test: prior PR head resolved `{ closed: true }` and killed the adopted tab instead of propagating `terminal_tab_pinned`; green after routing refinement.
- Original leak test was confirmed red before the headless routing fix (renderer relay called; PTY not killed) and green after.
- Advisory renderer failure test was confirmed red before containment (exception escaped and flush was skipped) and green after; it asserts exact kill, persistence pruning, one flush, and the diagnostic warning.
- `src/main/runtime/orca-runtime.test.ts`: **717 passed**.
- Headless promotion/runtime gate: **7 files, 754 tests passed**.
- Retirement/authority gate under Node 24: **10 files, 446 tests passed**.
- Provider/descendant/worktree teardown gate: **5 files, 218 tests passed**.
- Fresh Electron: `headless-serve-desktop-activation.spec.ts` passed.
- Fresh Electron: `terminal-parked-close-retirement.spec.ts` passed.
- `pnpm run tc` passed.
- Touched-file `oxlint` and `oxfmt --check` passed; reliability manifest and max-lines ratchet passed.
- Repository-wide `pnpm run lint` currently stops on two unrelated pre-existing switch-exhaustiveness errors in `native-chat-session-option-labels.ts` and `skill-freshness-group.tsx`; this PR does not touch them.

## Residual gaps / risk

- Live Linux, Windows/ConPTY, WSL, and SSH-host process-absence evidence was not collected.
- Advisory renderer cleanup failure now has a targeted warning; broader live Linux, Windows, WSL, and SSH-host teardown diagnostics remain uncollected.
- Cross-platform behavior is unchanged beyond ownership routing (no OS-specific code). SSH/remote is the primary affected surface.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -19722,11 +19722,9 @@ describe('OrcaRuntimeService', () => {
     expect(getSession().terminalLayoutsByTabId['host-tab']).toBeUndefined()
   })
 
-  it('tears down a runtime-owned SSH headless tab authoritatively even when the renderer relay is wired', async () => {
-    // #8958 regression: with a window attached, both closeTerminal and
-    // closeTerminalTab are wired. A headless SSH tab is absent from the window
-    // graph, so the closeTerminalTab relay would resolve nothing and ack success
-    // while the PTY kept running. The authoritative teardown must win.
+  it('durably tears down a runtime-owned SSH headless tab when renderer cleanup fails', async () => {
+    // #8958 regression: the acknowledged renderer relay cannot see headless tabs;
+    // its advisory fallback must not prevent authoritative teardown or flushing.
     const persistedPtyId = 'ssh:ssh-1@@relay-pty'
     const { runtimeStore, getSession } = makeRuntimeStoreWithWorkspaceSession(
       makeWorkspaceSessionWithHeadlessTerminal({
@@ -19750,9 +19748,14 @@ describe('OrcaRuntimeService', () => {
       })
     )
     const kill = vi.fn(() => true)
-    const closeTerminal = vi.fn()
+    const flushOrThrow = vi.fn()
+    const rendererError = new Error('renderer unavailable')
+    const closeTerminal = vi.fn(() => {
+      throw rendererError
+    })
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {})
     const closeTerminalTab = vi.fn(async () => {})
-    const runtime = new OrcaRuntimeService(runtimeStore as never)
+    const runtime = new OrcaRuntimeService({ ...runtimeStore, flushOrThrow } as never)
     runtime.setPtyController({
       write: () => true,
       kill,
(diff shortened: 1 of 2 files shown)


AmethystLiang:
Related to #8459.

Issue #7351, opened 2026-07-04
[Bug]: Windows: bundled CLI wrapper (orca.cmd) looks for Orca.exe in wrong directory


juanazo108:
### Operating system

Windows 11

### Details

**Environment:** Windows 11, Orca installed via .exe installer at
`C:\Users\<user>\AppData\Local\Programs\orca\`

**Issue:** The bundled CLI wrapper at
`resources\app.asar.unpacked\resources\win32\bin\orca.cmd` fails with:
Unable to locate Orca.exe next to "C:\Users<user>\AppData\Local\Programs\orca\resources\app.asar.unpacked\resources\win32"

The wrapper expects `Orca.exe` to be adjacent to the `win32` directory,
but in the Windows install layout `Orca.exe` lives at the installation
root, four levels up.

**Impact:** The `orca` CLI is unusable on Windows out of the box, which
blocks the orchestration skill workflow
(`npx skills add ... --skill orchestration`) since agents can't invoke
`orca terminal` commands.

**Note:** There's no visible option under Settings to register the CLI
on this build either (searched Settings for "CLI" — only Gemini CLI
credentials appears).

Happy to provide more details. Thanks for the great product.

OrcaWin:
## Agent reproduction package (Windows orchestration 2026-07-16)

**Host context:** Windows 11 / Orca **1.4.143.0** / repo path used: `docs/bug-reproductions/`  
**Labels applied:** see issue labels (`has_repro` or `cannot_repro`).  
**Security:** Do **not** run user-uploaded binaries from issue attachments. Recreate fixtures from the steps/scripts below.

---
# Issue 7351 — bundled orca.cmd looks for launcher in wrong directory

**Status:** `repro_confirmed` (partial — unpacked shim) / primary path fixed  
**Host:** Windows 11, Orca 1.4.143 install  
**Label:** `has_repro`

## Summary

Issue claimed the wrapper looks for `Orca.exe` next to the `win32` tree and fails. On 1.4.143:

| Shim | Behavior |
|------|----------|
| `resources\bin\orca.cmd` | **Works** — launches adjacent `orca.exe` (native CLI), `status --json` OK |
| `resources\app.asar.unpacked\resources\win32\bin\orca.cmd` | **Fails** — looks for `orca.exe` next to itself; file missing |

So orchestration/skills using the **primary** `resources\bin` path work. Anything that still points at the unpacked `win32\bin\orca.cmd` hits the failure:

```
[...]

PR #9123, opened 2026-07-17
fix(win): stop shipping duplicate broken orca.cmd shim in app.asar

xianjianlf2:
# fix(win): stop shipping duplicate broken orca.cmd shim in app.asar

## Summary
Fixes the Windows CLI launcher failing with *"Unable to locate the native Orca CLI launcher…"* because a duplicate, broken `orca.cmd` shim was bundled inside `app.asar` (#7351).

## ELI5
On Windows, Orca accidentally shipped two copies of its command-line helper — and the extra copy was placed where it couldn't find the program it needs to run, so it broke. This removes the bad copy and keeps the one that works.

## Root cause & fix
The Windows CLI shim ships correctly through `win.extraResources` to `resources/bin/orca.cmd`, right next to the native `resources/bin/orca.exe` it resolves against. But the source tree `resources/win32` was *also* packed into `app.asar`, and `asarUnpack: ['resources/**']` extracted a second copy to `app.asar.unpacked/resources/win32/bin/orca.cmd` with no adjacent `orca.exe`, so launching it failed. The fix adds `!resources/win32{,/**/*}` to `config.files`, excluding the source tree from `app.asar` while preserving the working `extraResources` delivery.

## Evidence

- Test: `keeps the Windows CLI shim source tree out of app.asar` in `config/scripts/electron-builder-config.test.mjs`.
- On branch (fix present): **22 passed (22)**.
- Fix reverted to `origin/main`, test kept: **1 failed | 21 passed** — the guard fails because `config.files` lacks `!resources/win32{,/**/*}`.

```
# On branch (fix present)
Test Files  1 passed (1)
Tests  22 passed (22)

# Fix reverted, test kept
Test Files  1 failed (1)
Tests  1 failed | 21 passed (22)
Failing: "keeps the Windows CLI shim source tree out of app.asar"
Assertion: expect(electronBuilderConfig.files).toEqual(
  expect.arrayContaining(['!resources/win32{,/**/*}']))
```

- Lint: `npx oxlint config/electron-builder.config.cjs` — clean, no findings.

## Trade-offs
None — pure fix.

## Regression risk
Zero-regression: the change only adds one exclusion glob to `config.files`, so the working `extraResources` copy at `resources/bin/orca.cmd` is untouched and all non-Windows paths are byte-identical. The proven regression test locks the exclusion in place.

_Credit to original author @xianjianlf2. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/config/electron-builder.config.cjs
+++ b/config/electron-builder.config.cjs
@@ -86,7 +86,13 @@ module.exports = {
     // Why: feature-wall media is copied via extraResources so runtime can read
     // it from process.resourcesPath; exclude the source copy from app.asar.
     '!resources/onboarding/feature-wall/**',
-    '!resources/skills/**'
+    '!resources/skills/**',
+    // Why: the Windows CLI shim ships via extraResources to resources/bin/orca.cmd
+    // (beside the native resources/bin/orca.exe). Packing the source tree into
+    // app.asar too lets asarUnpack:['resources/**'] extract a second copy at
+    // app.asar.unpacked/resources/win32/bin/orca.cmd with no adjacent orca.exe,
+    // which fails to launch the CLI (#7351).
+    '!resources/win32{,/**/*}'
   ],
   // Why: the CLI entry-point lives in out/cli/ but imports shared modules
   // from out/shared/ and local hook mutators from out/main/. These paths must be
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @xianjianlf2 — stop shipping duplicate broken orca.cmd shim in app.asar. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

Issue #9130, opened 2026-07-17
Mobile pairing broken on IPv6-only hosts: getNetworkInterfaces() only enumerates IPv4 ("WebSocket transport is not running")


PannenetsF:
## Summary

Mobile pairing is impossible on an **IPv6-only host** (e.g. a headless `orca serve` reachable only over IPv6). Settings → Mobile shows **"WebSocket transport is not running"** / no selectable network interfaces, and the QR/pairing generation returns `available: false`, even though the WebSocket transport is running fine on `ws://0.0.0.0:6768`.

## Root cause

`getNetworkInterfaces()` in `src/main/ipc/mobile.ts` only collects **non-internal IPv4** addresses:

```ts
if (addr.family === 'IPv4' && !addr.internal) {
  result.push({ name, address: addr.address })
}
```

On an IPv6-only host this returns `[]`, so `getDefaultPairingAddress()` returns `null`, `mobile:getPairingQR` gets no address, and the UI reports the transport as unavailable.

The rest of the pipeline is already IPv6-ready: `resolvePairingEndpoint()` in `src/main/runtime/runtime-rpc.ts` correctly bracket-wraps IPv6 hosts (`endpoint.hostname = parsed.host.includes(':') ? \`[${...}]\` : parsed.host`). Only interface enumeration is IPv4-blind.

## Repro

1. Run `orca serve` on a host that has only IPv6 connectivity (no non-internal IPv4).
2. Pair a client, open Settings → Mobile (or the mobile QR flow).
3. Observe "WebSocket transport is not running" / "no interfaces found"; no way to advertise an address for phone pairing.

(Confirmed on a containerized `orca serve` reachable only via an internal IPv6 address. `serve --mobile-pairing --pairing-address <ipv6>` works and prints a valid QR, which proves the transport and endpoint handling are fine — only the Settings UI's interface discovery is broken.)

## Fix

Extend `getNetworkInterfaces()` to also include non-internal IPv6 addresses, excluding link-local `fe80::/10` (which need a zone id and can't be advertised in a QR). Keep the existing preference order: tailnet IPv4 first, then other IPv4, then IPv6 as the fallback for IPv6-only environments.

PR incoming with a regression test (IPv6-only host now yields the IPv6 interface; link-local excluded; IPv4/tailnet still ranked first).

brennanb2025:
Fixed on `main` by #9131.

[...]

PR #9131, opened 2026-07-17
fix: enumerate IPv6 addresses for mobile pairing on IPv6-only hosts (#9130)

PannenetsF:
## Summary

Fixes mobile pairing being impossible on IPv6-only hosts, where Settings → Mobile showed no selectable network interface and reported the transport as unavailable.

## ELI5

If your computer could only be reached over the newer IPv6 kind of address, the app couldn't find any way to connect your phone. Now it looks for IPv6 addresses too, so pairing works.

## Root cause & fix

`getNetworkInterfaces()` in `src/main/ipc/mobile.ts` only enumerated non-internal IPv4 addresses. On IPv6-only hosts (e.g. a headless `orca serve` reachable only over IPv6) the scan returned an empty list, making `getDefaultPairingAddress()` null so QR generation reported the transport as not running. The fix includes non-internal IPv6 addresses — excluding link-local `fe80::/10` (which need a zone id and can't be advertised in a QR) via `isUsableIPv6Address()` — and preserves preference order in `rankAddress()`: tailnet IPv4 (0) < other IPv4 (1) < IPv6 (2).

## Evidence

- Suite `src/main/ipc/mobile.test.ts` (via `npx vitest run --config config/vitest.config.ts`): all 14 tests pass on branch.
- Reverting only the fix (`git checkout origin/main -- src/main/ipc/mobile.ts`) while keeping the tests fails 2 of 14, proving the tests capture the bug:

```
On branch:        Test Files 1 passed (1); Tests 14 passed (14)
Fix reverted:     Test Files 1 failed (1); Tests 2 failed | 12 passed (14)
  ✗ returns an IPv6 interface on an IPv6-only host
      expected interfaces:[{eth0, 2605:...}] but got interfaces:[]
  ✗ includes IPv6 addresses (ranked after IPv4) and excludes link-local IPv6
```

- Lint clean: `npx oxlint src/main/ipc/mobile.ts` reports no diagnostics.

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression: IPv6 is only added as the lowest-priority fallback, so default pairing behavior on IPv4/tailnet hosts is unchanged. The link-local exclusion regex correctly covers `fe80:`–`febf:`, and all 12 pre-existing IPv4/tailnet tests still pass.

_Credit to @PannenetsF for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/main/ipc/mobile.test.ts
+++ b/src/main/ipc/mobile.test.ts
@@ -59,6 +59,41 @@ describe('registerMobileHandlers', () => {
     })
   })
 
+  it('includes IPv6 addresses (ranked after IPv4) and excludes link-local IPv6', () => {
+    networkInterfacesMock.mockReturnValue({
+      en0: [
+        { family: 'IPv4', internal: false, address: '192.168.1.24' },
+        { family: 'IPv6', internal: false, address: 'fe80::1' },
+        { family: 'IPv6', internal: false, address: '2605:340:cd51:2a01:0:2b13:f279:c096' }
+      ],
+      lo0: [{ family: 'IPv6', internal: true, address: '::1' }]
+    })
+
+    registerMobileHandlers({} as never)
+
+    expect(handlers.get('mobile:listNetworkInterfaces')?.()).toEqual({
+      interfaces: [
+        { name: 'en0', address: '192.168.1.24' },
+        { name: 'en0', address: '2605:340:cd51:2a01:0:2b13:f279:c096' }
+      ]
+    })
+  })
+
+  it('returns an IPv6 interface on an IPv6-only host (regression: was empty, breaking mobile pairing)', () => {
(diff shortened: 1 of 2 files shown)


PannenetsF:
Thanks @coderabbitai — good point on the `/10` range. In practice the OS only auto-assigns `fe80:` link-local addresses, so `^fe80:` covers what we actually enumerate here, but I agree `^fe[89ab][0-9a-f]:` is the mathematically exact filter for `fe80::/10`. Happy to tighten it to that if maintainers prefer strict coverage.

nwparker:
Thanks so much for this, @PannenetsF — enumerate IPv6 addresses for mobile pairing on IPv6-only hosts (#9130). Reviewed, rebased on latest `main`, and merged. Really appreciate the contribution! 🙏

PR #9143, opened 2026-07-17
fix(git): enable parallel worktree checkout

### LINKED ISSUE #9139: Slow worktree creation on Windows: full checkout is bottlenecked by Defender per-file scanning
## Summary

Worktree creation on Windows can take **up to ~10–12 seconds** for a large repo. I instrumented the create path and drove the real dev app to measure it end to end. The cost is almost entirely the working-tree **checkout** inside `git worktree add`, and on Windows that checkout is dominated by **Microsoft Defender real-time (on-access) scanning of every file written** — not by git or by Orca overhead.

Measured on the `orca` repo itself (8,351 files) on a Windows 11 machine, 8 logical CPUs, Git 2.55, Defender real-time protection **on**, repos folder **not excluded**.

## Evidence

**1. Per-phase timing from a real create (new `[worktree-create]` log line):**

```
[worktree-create] local total=11497ms git_worktree_add=10110ms list_created_worktree=197ms prepare_setup=83ms spawn_startup_terminal=0ms persist_metadata=0ms
```

`git_worktree_add` is **10.1s of 11.5s**. Another ~1.1s sits before the checkout in the (currently untimed) pre-add git-probe cluster. There was no fetch here because the base branch was already local.

**2. Raw `git worktree add` from the CLI (Orca completely out of the picture):**

```
real  0m8.545s   user  0m0.030s   sys  0m0.046s
Updating files: 100% (8351/8351), done.
```

git's own CPU is **0.08s total**. The other ~8.4s is **pure I/O wait** — git blocks on each file write while it is scanned on close.

**3. Defender state:** real-time + on-access protection **enabled** (Normal mode); the `C:\Users\<user>\orca` tree is **not** on the exclusion list.

**4. Pure filesystem write benchmark (no git):** writing 8,000 small files to that exact directory takes **5,751ms (0.72 ms/file)** — matching git's checkout rate (8,351 files / 8.5s ≈ 0.98 ms/file). This pins the cost to the OS write path + AV, not to git.

**5. Cold vs warm:** repeated checkouts of *identical* content drop from ~8,500ms to ~60ms (~100×), because Defender caches scan results by content hash. So the pain lands on **cold** checkouts: first worktree after clone, after a big `pull` that changed many blobs, or once the AV cache has evicted. Real users are usually near the cold case.

## Root cause

A worktree create is a full working-tree checkout. A large repo writes thousands of files, and on Windows each file-close is serialized behind a synchronous Defender on-access scan (~0.7–1 ms/file). ~8k files × ~1 ms ≈ 8–10s. Orca adds ~1–2s on top from the ~12–20 pre/post git subprocess spawns it issues per create (each `git.exe` launch is itself scanned).

### Lever 1 — Parallel checkout (universal, low-risk, ship it)

Pass git's parallel-checkout config to `git worktree add`:

```
git -c checkout.workers=0 worktree add …
```

`checkout.workers=0` = one worker per core, overlapping the per-file scan waits across CPUs. Applies to both `addWorktree` and `addSparseWorktree` in `src/main/git/worktree.ts`.

- **No admin, no AV change, no UX change** — just faster.
- **Baseline-safe:** git silently ignores unknown config, so on Git <2.32 (before parallel checkout) it is a no-op — no fallback needed. Fits the `-c`-before-subcommand rule in `AGENTS.md`.
- Helps macOS/Linux large checkouts too.

**Measured effect (controlled benchmark):** 8 trials, each on fresh unique random content (guaranteed cold scan), interleaved with the starting method swapped between blocks, 1,500 files/checkout:

| method | samples (ms) | median |
|---|---|---|
| serial (`workers=1`) | 1176, 1163, 1262, 1381 | **1262** |
| parallel (`workers=0`) | 601, 617, 757, 622 | **622** |

**~1.9× faster, zero overlap** between distributions (slowest parallel beat fastest serial). For the real 8.3k-file checkout that is roughly 8.5s → ~4.5s.

**Caveats:** it *mitigates*, it does not *eliminate* — it overlaps scans rather than removing them. Gain scales sub-linearly with cores (8 cores → ~2×) and shrinks on machines with light/no AV. It never hurts.

### Lever 2 — In-app alert + one-click Defender exclusion (biggest Windows win)

This is the largest speedup (removes scanning entirely: ~5×+, near-100× on warm content). We now capture per-phase timing, so we can detect the condition precisely and only surface a tip when it is real.

Trigger when **all** hold:
- `process.platform === 'win32'`
- `git_worktree_add` was slow (e.g. checkout rate below ~2k files/s)
- `Get-MpComputerStatus` shows real-time protection on (non-admin query)
- the worktrees base dir is not already excluded

Then show a one-time, dismissible tip: *"Worktree creation was slow (11s). Excluding your repos folder from Microsoft Defender makes it ~5× faster."* with an action that (on click, via a UAC prompt) runs:

```powershell
Add-MpPreference -ExclusionPath "<worktrees base dir>"
```

Design constraints:
- **User-initiated only, never automatic** — an AV exclusion is a security trade-off; scope it narrowly to the worktrees base dir, not the whole drive.
- Detection is non-admin; only *adding* the exclusion needs elevation, so UAC happens only on click.
- Include "don't show again" + a docs link explaining the trade-off. Fits the existing feature-tip infrastructure.
- **Suppress over SSH/WSL** — the checkout runs on the remote host there, so the Windows-Defender advice does not apply.

### Lever 3 — Fewer files written (structural, opt-in follow-up)

Default large-repo worktrees to **sparse checkout** (`addSparseWorktree` already exists) so only the needed cone is materialized. Fewer files → less scanning, helps every platform. Belongs behind a per-repo preset or a "large repo" heuristic, not a silent global default.

## Open design questions

1. **Exclusion scope (Lever 2):** worktrees base dir (narrow, safest), repos root, or the `git.exe` process?
2. **Elevation UX:** one-click "Fix it" that launches elevated PowerShell via UAC, or a *guided* tip that only shows the command + docs link (Orca never invokes admin itself — lower risk)?
3. **Lever 1 rollout:** unconditional (all platforms) or Windows-only to start?
4. **Alert threshold:** absolute time (e.g. >4s), a files/sec rate, or gate strictly on "Defender on + path not excluded"?
5. **Telemetry:** worth tracking the timing phases (through the telemetry gate) to size the affected Windows population before investing in Lever 2?

## Instrumentation already added (on branch `OrcaWin/slow-worktree-creation-on-windows`)

A per-phase timing log so the dominant phase is visible on any machine:

- `src/main/worktree-create-timing.ts` — `formatWorktreeCreateTiming()` (renders total + phases slowest-first)
- `src/main/ipc/worktree-remote.ts` — logs `[worktree-create] local|remote …` at both create return sites
- `src/main/worktree-create-timing.test.ts` — tests for the formatter

This lets us confirm any fix in the field: `git_worktree_add` should collapse after a Defender exclusion and roughly halve with parallel checkout.

## Repro / environment

- Windows 11, 8 logical CPUs, Git 2.55, Defender real-time protection on, repos folder not excluded.
- Repo: `orca` itself (8,351 files at `main`).
- Create a worktree; observe the `[worktree-create]` log and the ~10s `git_worktree_add` phase.

kin001:
## Summary

- opt local and WSL `git worktree add` into Git's parallel checkout workers
- apply the same checkout config to the materializing `git checkout` step for sparse worktrees
- preserve existing option ordering, branch/ref operands, paths with spaces, sparse `--` boundaries, and rollback behavior

Fixes #9139.

## ELI5

Creating a worktree writes thousands of files. Git can overlap those writes across CPU cores, which reduces time spent waiting on per-file filesystem and antivirus work. Orca now asks Git to use that built-in mode without changing what gets checked out.

## User Regression Tradeoffs

- Git 2.32+ can use parallel workers; older supported Git versions ignore the unknown config and keep their existing serial behavior.
- The speedup depends on storage, antivirus, repository size, and CPU count; checkout results and failure handling are unchanged.
- SSH relay worktree creation uses a separate command path and remains behaviorally unchanged and unoptimized in this focused local/WSL fix.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` - changed-file `oxlint`/`oxfmt` and max-lines checks pass; full lint reaches two pre-existing switch-exhaustiveness errors on `upstream/main` in `native-chat-session-option-labels.ts:47` and `skill-freshness-group.tsx:101`
- [x] `pnpm typecheck`
- [ ] `pnpm test` - full Windows suite completed with 133 unrelated existing failures in path/symlink, PTY, WSL, timing, and source-boundary suites; focused worktree coverage passes 101/101
- [x] `pnpm build`
- [x] Added regression coverage for ordinary/existing-branch creation, sparse materialization, global-option ordering, spaces, option-like sparse operands, and sparse rollback

Focused command:

```text
pnpm exec vitest run --config config/vitest.config.ts src/main/git/worktree.test.ts src/main/git/remove-worktree.test.ts
```

## AI Review Report

The change was independently reviewed after implementation and rebased onto the latest `upstream/main`.

- Correctness: `-c checkout.workers=0` stays before the Git subcommand; every existing branch/base/path operand and timeout is preserved. Sparse worktrees receive the config on both `worktree add --no-checkout` and the later checkout that actually materializes files.
- Cross-platform: the fixed Git config is platform-neutral and passed as argv, not through a shell. Native macOS/Linux/Windows and WSL use this path; no shortcuts, labels, path normalization, or Electron platform branches changed.
- SSH/remote/local: local and WSL behavior is optimized. The separate SSH relay path was inspected and intentionally left unchanged for this issue, so remote semantics do not regress.
- Agents and integrations: no agent, provider, terminal, integration, or git-host-specific behavior changed.
- Performance: worktree creation adds no probes or subprocesses; it only enables Git's built-in checkout worker pool. Gains are workload-dependent.
- UI quality: no UI or interaction code changed.

## Security Audit

No input execution, shell interpolation, authentication, secret, dependency, filesystem-permission, or IPC surface is introduced. The new arguments are fixed literals passed to the existing Git argv runner. User-controlled refs, paths, and sparse directories remain separate operands, including the existing `--` separator for sparse paths.

## Notes

- Branch is one commit on current `main`.
- X handle: not provided.

### DIFF
--- a/src/main/git/remove-worktree.test.ts
+++ b/src/main/git/remove-worktree.test.ts
@@ -1308,7 +1308,7 @@ branch refs/heads/main
     const calls = getGitCalls()
     expect(calls).toEqual(
       expect.arrayContaining([
-        'git worktree add --no-checkout --no-track -b feature/test /repo-feature',
+        'git -c checkout.workers=0 worktree add --no-checkout --no-track -b feature/test /repo-feature',
         'git config --get push.autoSetupRemote',
         'git config --local push.autoSetupRemote true',
         'git sparse-checkout init --cone',
--- a/src/main/git/worktree.test.ts
+++ b/src/main/git/worktree.test.ts
@@ -33,6 +33,8 @@ import {
   WORKTREE_ADD_TIMEOUT_MS
 } from './worktree'
 
+const PARALLEL_CHECKOUT_GIT_ARGS = ['-c', 'checkout.workers=0']
+
 beforeEach(() => {
   clearGitCapabilityStateForTests()
 })
@@ -685,6 +687,7 @@ describe('addWorktree', () => {
       [['rev-parse', '--verify', '--quiet', 'refs/remotes/origin/main^{commit}'], { cwd: '/repo' }],
       [
         [
+          ...PARALLEL_CHECKOUT_GIT_ARGS,
           'worktree',
           'add',
           '--no-track',
@@ -710,16 +713,16 @@ describe('addWorktree', () => {
     ])
   })
 
-  it('checks out a selected existing local branch without creating a new branch', async () => {
+  it('places parallel config before worktree add for a selected existing branch', async () => {
     gitExecFileAsyncMock.mockResolvedValueOnce({ stdout: '' }) // worktree add
 
-    await addWorktree('/repo', '/repo-feature', 'feature/test', 'feature/test', false, false, {
+    await addWorktree('/repo', '/repo feature', 'feature/test', 'feature/test', false, false, {
       checkoutExistingBranch: true
     })
 
     expect(gitExecFileAsyncMock.mock.calls).toEqual([
       [
-        ['worktree', 'add', '/repo-feature', 'feature/test'],
+        [...PARALLEL_CHECKOUT_GIT_ARGS, 'worktree', 'add', '/repo feature', 'feature/test'],
         { cwd: '/repo', timeout: WORKTREE_ADD_TIMEOUT_MS }
       ]
     ])
@@ -736,7 +739,12 @@ describe('addWorktree', () => {
     })
 
     const worktreeAddCall = gitExecFileAsyncMock.mock.calls.find(
-      ([argv]) => Array.isArray(argv) && argv[0] === 'worktree' && argv[1] === 'add'
+      ([argv]) =>
+        Array.isArray(argv) &&
+        argv[0] === '-c' &&
+        argv[1] === 'checkout.workers=0' &&
+        argv[2] === 'worktree' &&
+        argv[3] === 'add'
     )
     expect(worktreeAddCall?.[1]).toMatchObject({ timeout: WORKTREE_ADD_TIMEOUT_MS })
     expect(WORKTREE_ADD_TIMEOUT_MS).toBeGreaterThan(0)
@@ -750,7 +758,15 @@ describe('addWorktree', () => {
 
     expect(gitExecFileAsyncMock.mock.calls).toEqual([
       [
-        ['worktree', 'add', '--no-track', '-b', 'feature/no-base', '/repo-feature'],
+        [
+          ...PARALLEL_CHECKOUT_GIT_ARGS,
+          'worktree',
+          'add',
+          '--no-track',
+          '-b',
+          'feature/no-base',
+          '/repo-feature'
+        ],
         { cwd: '/repo', timeout: WORKTREE_ADD_TIMEOUT_MS }
       ],
       [['config', '--get', 'push.autoSetupRemote'], { cwd: '/repo-feature' }]
@@ -825,6 +841,7 @@ describe('addWorktree', () => {
       [['rev-parse', '--verify', '--quiet', 'refs/remotes/origin/main^{commit}'], { cwd: '/repo' }],
       [
         [
+          ...PARALLEL_CHECKOUT_GIT_ARGS,
           'worktree',
           'add',
           '--no-track',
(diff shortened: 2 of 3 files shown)


brennanb2025:
Taking a look!

PR #7614, opened 2026-07-07
fix(terminal): repaint revealed panes stuck behind xterm's paused-render gate

Jinwoo-H:
## Problem

When switching back to a workspace/tab, a TUI (e.g. Claude Code) sometimes shows **stale/missing rows near the bottom** that only reappear when you right-click-drag to select the missing area. It happens more often when the machine is under load.

## Root cause

On reveal, `resetWebglTextureAtlas` (`pane-webgl-renderer.ts`) clears xterm's WebGL render model (`clearTextureAtlas()`) and calls `terminal.refresh(0, rows-1)` to force a full repaint from the buffer.

But xterm's `RenderService.refreshRows()` has an early-return gated on its **IntersectionObserver**:

```ts
public refreshRows(start, end, sync = false, ...) {
  if (this._isPaused) {           // observer hasn't reported the pane as intersecting yet
    this._needsFullRefresh = true;
    return;                       // <-- our refresh() is swallowed
  }
  ...
}
```

When a pane is revealed it is DOM-visible, but the observer's callback can lag a frame (more likely under load, when layout/overlay settle is slower). In that window `_isPaused` is still `true`, so the reveal-repaint's `refresh()` just latches `_needsFullRefresh` and returns. The freshly-cleared model never repaints, and the canvas keeps compositing pre-hide pixels. A drag-select recovers it because `handleSelectionChanged` drives a render that isn't behind the same gate.

Note the `sync: true` path (already used by the foreground render-settle) does **not** help — the `_isPaused` check runs *before* the `sync` branch.

## Fix

Add `forceRepaintThroughRenderPause()`: when the render service is paused, clear the pause latch and drive **one synchronous full-viewport render** ourselves, rather than waiting on the observer. xterm's observer reasserts authority on its next callback (it always recomputes `_isPaused` from the entry), so the transient clear is self-correcting.

`resetWebglTextureAtlas` now calls this first and only falls back to `terminal.refresh()` when the terminal wasn't gated (preserving current behavior for DOM-rendered panes and non-xterm targets).

All internal access is behind `typeof` guards and `try/catch`: an xterm upgrade that renames these internals **degrades to the existing `refresh()` path** instead of throwing into a render frame.

## Testing

- New unit suite `terminal-render-pause-release.test.ts` (5 tests): paused → forces `refreshRows(0, rows-1, true)` + clears both latches; not-paused → untouched, returns false; missing internals → false; invalid row count → false; forced render throws → false but latch still cleared.
[...]

PR #9148, opened 2026-07-17
fix(terminal): converge revealed inline-TUI panes without a manual resize

Jinwoo-H:
## Field bug

On v1.4.144-rc.4, switching to a workspace whose Codex TUI kept streaming while hidden periodically shows:

- a **mostly-blank terminal** — viewport stranded mid-buffer (scrollbar mid-track), input box missing, sparse scrollback; or
- a **bottom gap** — Codex draws its input box rows above the true bottom (PTY rows ≠ pane rows).

Both persist until a manual resize (e.g. Cmd+L sidebar toggle) forces SIGWINCH and Codex repaints. Same family as #7951.

Since parking landed, a workspace hidden >30s beyond the hot-retain budget fully unmounts its terminal and rebuilds from a daemon snapshot on reveal — a path none of the prior fixes (#7614/#8625/#8674) protect. Inline-mode TUIs are uniquely exposed because they keep appending across the reveal, so the snapshot's bottom is never the live bottom (alt-screen fixtures in existing specs can't catch this).

## Three independent losing legs, each fixed

1. **Stranded corrective fit** (`pane-fit.ts`) — the reattach/hidden-snapshot replay parks its PTY-resize continuation behind "the next successful fit". When the pane is momentarily unmeasurable mid-reveal, no later fit is guaranteed (the container size may already be final, so no further ResizeObserver event ever fires) and xterm stays at snapshot dims with no forwarded resize/SIGWINCH → the bottom-gap state. `safeFitAndThen` now actively retries on settled frames (bounded at 40 attempts) and records a `terminal_safe_fit_retry_exhausted` breadcrumb if it ever gives up, so a residual field case names itself.

2. **Swallowed replay paint** (`pane-terminal-foreground-render-settle.ts`) — the cold-park remount path has no reveal-repaint, so a replay parsed while xterm's RenderService is paused (its IntersectionObserver lags the reveal by a frame, worse under load) never painted: buffer perfect, canvas blank until drag-select/resize. The settle refresh now drives one synchronous full render through the pause latch (extends the #7614 release, which only covered hidden-but-mounted reveals, to every foreground/replay refresh).

3. **Viewport stranded at the snapshot bottom** (`pty-connection.ts`) — the reveal's bottom anchor is a one-shot that runs before the live chunks deferred during reattach get parsed. The restored scroll intent is now re-enforced after the deferred drain parses: following panes re-anchor to the live bottom; a deliberate scroll-up pin is preserved (`enforceTerminalCurrentScrollIntent` semantics). Deferred data is capped at 512KB, bounding the extra parse cost.

## Repro / regression harness

New `tests/e2e/terminal-inline-tui-reveal-convergence.spec.ts` + `tests/e2e/fixtures/codex-inline-live-block-fixture.cjs`: a codex-shaped inline TUI (agent-classified tab via launch telemetry, scrollback inserts + bottom live-block redraws, synchronized output, mouse reporting, startup queries, heartbeat file) driven through six reveal shapes — hidden-mounted, worktree switch after a hidden-time window resize, parked reveal ×6 cycles under 6× CPU throttle with 8k-line scrollback, rapid tab flapping, rapid worktree flapping, and parked reveal after a parked-time resize. Each requires convergence **with no resize**: viewport at buffer bottom, a fresh frame + input-box row on screen, still following afterwards, xterm grid == fit proposal == PTY-applied size, and painted-pixel ink in the live-block band (WebGL forced) so paint-layer bugs can't hide behind a healthy buffer.

The exact field interleavings are sub-frame races that even the escalated harness could not force on demand, so the product changes are repairs of statically verified latches — legs 1 and 2 each carry a deterministic unit repro that is red without the fix (`pane-fit.test.ts`, `pane-terminal-foreground-render-settle.test.ts`) — and the e2e suite is the standing regression net for the symptom family.

## Verification

- New unit repros red→green; full `pty-connection` suite (446), scroll-intent, pane-tree-ops, fit-observer, write-callback-guard, render-pause-release, reveal-repaint suites green.
- Output-scheduler unit + throughput bench green (the settle change sits on the per-batch write path).
- E2E green against the rebuilt bundle: hidden-view-parking (incl. 25-cycle byte-for-byte reveal), the new 6-scenario convergence suite, paused-render repro, tab-switch SIGWINCH restore, hidden-TUI visual restore, tab-switch visual restore. (One pre-existing cross-file order flake in "restores skipped hidden Grok output on light tab resume" — passes solo and on full-file rerun.)
- `pnpm typecheck` and oxlint clean.

## Behavior notes / trade-offs

- All three changes act only inside failure windows that previously ended in "broken until manual resize"; steady-state rendering/scroll/throughput paths are untouched.
- In the previously-broken cases the heal arrives as one quick TUI redraw shortly after the switch (the SIGWINCH that should have fired at reveal).
- Deliberately **not** a continuous enforcement watchdog — per-batch enforce loops are the pattern that caused earlier scroll-freeze regressions; these are one-shot recoveries tied to reveal events.

## Known follow-ups (out of scope)

- The flood-abandon path writes pending live chunks back un-seq-sliced (`pty-connection.ts` abandon), which can duplicate visible lines under sustained floods — needs its own fuzz-mirror repro.
- One organic catch during harness development: an agent-marked *visible* pane froze ~1s post-spawn (possible spawn-time hidden-declare unhide race); never reproduced again — the new breadcrumb will surface it if it recurs in the field.

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-connection.test.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.test.ts
@@ -5791,7 +5791,12 @@ describe('connectPanePty', () => {
     reattach.resolve()
     await flushAsyncTicks()
 
-    expect(transport.resize).toHaveBeenLastCalledWith(65, 63)
+    expect(transport.resize).toHaveBeenCalledWith(65, 63)
+    // Why the trailing guarded call: transport.resize is fire-and-forget, so
+    // the reattach fit is followed by an applied-size verification. With no
+    // getSize readback available (this harness), it re-forwards the same grid
+    // once as a guarded claim instead of leaving the PTY size unverified.
+    expect(transport.resize).toHaveBeenLastCalledWith(65, 63, { claim: true })
   })
 
   it('adopts a live eager PTY and withholds snapshots after its renderer dies', async () => {
--- a/src/renderer/src/components/terminal-pane/pty-connection.test.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.test.ts
@@ -11498,6 +11498,113 @@ describe('connectPanePty', () => {
     disposable.dispose()
   })
 
+  it('slices abandoned pending chunks against a replay that already painted', async () => {
+    // Field bug: the 750ms foreground deadline can fire while a RESOLVED
+    // snapshot's replay is still parsing (its writes are already in xterm's
+    // FIFO and WILL paint). The abandon used to rewrite parked chunks raw,
+    // duplicating every line the replay covers, and left the live reconcile
+    // baseline stale so main's ACK backlog duplicated again.
+    const { connectPanePty } = await import('./pty-connection')
+    const transport = createMockTransport('pty-id')
+    const capturedDataCallback: {
+      current: ((data: string, meta?: { seq?: number; rawLength?: number }) => void) | null
+    } = { current: null }
+    transport.connect.mockImplementation(async ({ callbacks }: { callbacks: ConnectCallbacks }) => {
+      capturedDataCallback.current = callbacks.onData ?? null
+      return 'pty-id'
+    })
+    transportFactoryQueue.push(transport)
+    const getMainBufferSnapshot = window.api.pty.getMainBufferSnapshot as unknown as ReturnType<
+      typeof vi.fn
+    >
+    const snapshot = createDeferred<{
+      data: string
+      cols: number
+      rows: number
+      seq: number
+      pendingDeliveryStartSeq: number
+    }>()
+    getMainBufferSnapshot.mockReturnValue(snapshot.promise)
+    const hidden = 'hidden-codex-output\r\n'
+    const coveredLive = 'LIVE_DUP_LINE\r\n'
+    const afterAbandon = 'AFTER_ABANDON\r\n'
+
+    const pane = createPane(1)
+    const manager = createManager(1)
+    const deps = createDeps({
+      isVisibleRef: { current: false },
+      startup: { command: 'codex' }
+    })
+    const disposable = connectPanePty(pane as never, manager as never, deps as never)
+    await flushAsyncTicks(6)
+    expect(capturedDataCallback.current).not.toBeNull()
+
+    vi.useFakeTimers()
+    capturedDataCallback.current?.(hidden, { seq: hidden.length, rawLength: hidden.length })
+    ;(deps.isVisibleRef as { current: boolean }).current = true
+    // Parked while the snapshot fetch is in flight; also arms the deadline.
+    capturedDataCallback.current?.(coveredLive, {
+      seq: hidden.length + coveredLive.length,
+      rawLength: coveredLive.length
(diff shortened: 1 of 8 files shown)


Jinwoo-H:
**Hardening commit `778fae26b`** (from adversarial self-review + a field observation):

1. **Applied-size verification after reattach.** Field observation: the bottom-gap variant *gradually self-heals as the agent emits more output*, walking content down to the true bottom. That is only possible when xterm is correctly sized and the **PTY** is the stale side — i.e. the reattach's fire-and-forget `transport.resize` was silently dropped daemon-side (#7951 class). The remount path had no corrector (`noteVisibilityResume`'s reassert only fires for hidden-mounted resumes). `fitAfterReattachRestore` now runs the applied-size reassert after the fit: one `pty:getSize` IPC per parked reveal, re-forwards only on drift/unavailable readback, zero steady-state cost.

2. **Bounded drain flush.** The post-drain scroll-intent enforce used `waitForTerminalOutputParsed`, whose unbounded synchronous flush could parse a flood pane's whole queue at reveal. Now bounded to the 512KB deferral budget + FIFO parse probe; enforcing after a partial drain stays correct (follow anchors at bottom, native follow tracks the remainder).

Re-verified: full `pty-connection` suite (446, incl. an updated reattach test that now pins the guarded reassert), typecheck, oxlint, and the parking + convergence e2e suites (10/10) against a rebuilt bundle.

Jinwoo-H:
**Commit `cd76d0797` — duplication fix** (reporter confirmed seeing TUI lines duplicated with the line below):

Root cause: the 750ms foreground deadline can abandon a hidden-output restore **while a resolved snapshot's replay is still parsing**. The replay writes are already in xterm's FIFO and will paint everything ≤ snapshot seq, but the abandon (a) rewrote its parked live chunks raw — duplicating every replay-covered line — and (b) skipped `setRestoredSnapshotBaseline`, so main's ACK-backlog re-deliveries below the seq duplicated a second time.

Fix: track the in-flight replay snapshot from the moment its writes are queued; on abandon, seq-slice the pending write-through against it (raw fallback only for unmappable offsets) and adopt it as the live reconcile baseline with the drain's continuity advance. Abandons where no replay started keep the existing raw write-through (correct there — nothing painted).

Repro-first: new deterministic unit test stages snapshot-resolve → replay-parse-held → deadline-abandon; red before the fix at the exact duplication assertion, green after. Full battery re-verified: 447 pty-connection tests, seq-slicing fuzz suite, typecheck/lint, and 15 e2e (parking, convergence, visual-restore suites) against a rebuilt bundle.

Note: a *visually identical* symptom also arises when Codex's erase misses the old live-block rows under a PTY/xterm row mismatch — that trigger is closed by this PR's size-convergence fixes, so both duplication sources are addressed here.

OrcaWin:
Superseded by #9503, which carries the hardened version of this fix on current main and has now merged. Closing this older PR to keep the review path unambiguous.

PR #9153, opened 2026-07-17
fix: pane divider drag-to-resize never moves under WSLg (mouse down, pen motion)

microtaro:
# fix: pane divider drag-to-resize never moves under WSLg (mouse down, pen motion)

## Summary
Restores terminal pane divider drag-to-resize on WSL2/WSLg, where dragging showed the resize cursor but never moved the split.

## ELI5
On Windows' WSLg Linux desktop, grabbing the line between two terminal panes to resize them did nothing. Now dragging works like everywhere else.

## Root cause & fix
WSLg's RDP input pipeline delivers button press/release as a `mouse` pointer but streams the intervening motion as a `pen` pointer with a *different* `pointerId`. The drag handler's strict guard `e.pointerId === activePointerId` discarded every `pointermove`, so the drag accumulated nothing and committed a no-op on release. The fix loosens the match to `isActiveDragPointer(e) = pointerId matches || e.isPrimary`, applied to the pointermove/up/cancel handlers, so the primary pen-relay motion continues the active drag.

## Evidence
Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase onto `origin/main` was clean (base `7ab601487c`, head `e2c0b92afd`).

- Test: `src/renderer/src/lib/pane-manager/pane-divider.test.ts` simulates the WSLg sequence (mouse `pointerdown` → pen `pointermove` → mouse `pointerup`).
- On branch: **PASS — 1 file, 10/10 tests passed.**
- Fix reverted to `origin/main` (test kept): **FAIL — 1 failed | 9 passed**, proving the test captures the bug:

```
FAIL src/renderer/src/lib/pane-manager/pane-divider.test.ts
  expected undefined to be '180 1 0%'  (previousPane.style.flex)
  at pane-divider.test.ts:281
```

- Lint clean: `oxlint` on `pane-divider-drag.ts` reported no warnings or errors.

## Trade-offs
None — pure fix.

## Regression risk
Zero-regression by construction: only the buggy pointer-match branch changes, and it loosens matching solely to *primary* pointers. During an active drag only one primary pointer exists per pointer type; secondary multi-touch pointers report `isPrimary=false` and remain ignored, so normal mouse and touch drags follow byte-identical paths. Uses the standard `PointerEvent.isPrimary`, so the fix is cross-platform safe. Proven by the passing on-branch suite.

*Credit to original author @microtaro. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/renderer/src/lib/pane-manager/pane-divider-drag.ts
+++ b/src/renderer/src/lib/pane-manager/pane-divider-drag.ts
@@ -218,8 +218,14 @@ export function attachDividerDrag(
     prevFlex = prevSize
   }
 
+  // Why: WSLg's RDP input path reports press/release as a `mouse` pointer but
+  // streams motion as a `pen` pointer with a different pointerId, so a strict
+  // pointerId match drops every move. Any primary pointer continues the drag.
+  const isActiveDragPointer = (e: PointerEvent): boolean =>
+    e.pointerId === activePointerId || e.isPrimary
+
   const onPointerMove = (e: PointerEvent): void => {
-    if (!dragging || e.pointerId !== activePointerId || !prevEl || !nextEl) {
+    if (!dragging || !isActiveDragPointer(e) || !prevEl || !nextEl) {
       return
     }
     didMove = true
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @microtaro — pane divider drag-to-resize never moves under WSLg (mouse down, pen motion). Reviewed, rebased on latest `main`, and merged. Really appreciate the contribution! 🙏

PR #9173, opened 2026-07-17
fix(codex): strip Orca-owned hook trust before mirroring config into managed homes (PR-A)

brennanb2025:
## What and why

Sanitize canonical Codex config before mirroring it into managed account homes. Orca-owned hook trust is identified with the existing expected-hash and grant-ledger ownership rules and removed from the mirrored bytes, while genuine user-authored hook trust is preserved.

This prevents real-home hook trust from referencing the wrong hooks.json path/hash in every managed account home and addresses user-case matrix rows 3 and 5.

## Verification

- `npx vitest run --config config/vitest.config.ts src/main/codex-accounts/service.test.ts` — 25 passed
- trust/reconciliation regression suites — 156 passed across config-toml-trust, real-home install, trust grant, and WSL runtime tests
- `pnpm run typecheck:node` — passed
- targeted oxlint — passed
- `pnpm run check:max-lines-ratchet` — passed

## Guardrails

- Tests use sandboxed temporary home and user-data directories only; the real `~/.codex` is never read or mutated.
- Canonical config and its grant ledger remain unchanged; sanitization is content-only before the ownership-marked managed-home atomic write.
- Flag-OFF behavior is unchanged.
- No merge or approval actions are included.

### DIFF
--- a/src/main/codex-accounts/service.test.ts
+++ b/src/main/codex-accounts/service.test.ts
@@ -394,6 +394,43 @@ describe('CodexAccountService config sync', () => {
     expectSanitizedManagedConfig()
   })
 
+  it('keeps flag-off config mirroring byte-identical', async () => {
+    const fixture = await createCanonicalHookTrustFixture()
+    const canonicalConfigPath = join(testState.fakeHomeDir, '.codex', 'config.toml')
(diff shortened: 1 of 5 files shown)


brennanb2025:

PR #9174, opened 2026-07-17
fix(codex): guard add-account when real home pins a custom provider (PR-D)

brennanb2025:
## What / why

- Detects a non-OpenAI top-level `model_provider` in the canonical Codex config before starting OAuth login.
- Returns a deterministic, actionable error explaining that custom providers use the system-default account and that OpenAI OAuth accounts require removing the pin or setting it to `openai`.
- Reuses the shared byte-preserving TOML scanner and string parser so profile/provider tables and multiline data do not look like root settings.

## Test evidence

- `npx vitest run --config config/vitest.config.ts src/main/codex-accounts/service.test.ts` — 24 passed.
- `npx vitest run --config config/vitest.config.ts src/main/codex-accounts/runtime-home-service.test.ts` — 87 passed.
- `pnpm run typecheck:node` — passed.
- Targeted oxlint and max-lines ratchet — passed.

## Guardrails

- The guard runs only in add-account before login; the flag-OFF system-default custom-provider launch path remains unchanged and is covered without `auth.json`.
- Missing `model_provider` and explicit `model_provider = "openai"` continue through normal multi-account OAuth login.
- The canonical `~/.codex/config.toml` is read only and is asserted byte-identical after both rejected add and system-default launch tests.

## Matrix

- Row 5: custom env-key provider (`codex-lb`), no `auth.json` — system-default launch succeeds; add OAuth account fails before spawning login with clear guidance; canonical config remains untouched.

### DIFF
--- a/src/main/codex-accounts/runtime-home-service.test.ts
+++ b/src/main/codex-accounts/runtime-home-service.test.ts
@@ -1441,6 +1441,30 @@ describe('CodexRuntimeHomeService', () => {
     )
   })
 
+  it('launches the system-default custom provider without requiring OAuth auth', async () => {
+    const systemCodexHome = getSystemCodexHomePath()
+    const canonicalConfigPath = join(systemCodexHome, 'config.toml')
+    const canonicalConfig = [
+      'model_provider = "codex-lb"',
+      '',
+      '[model_providers.codex-lb]',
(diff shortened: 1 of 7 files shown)


brennanb2025:

PR #9175, opened 2026-07-17
fix(codex): persist outgoing managed tokens before real-home lane takeover (PR-C)

brennanb2025:
## What

Guarantees the outgoing managed Codex account's refreshed OAuth token is read back (persisted to its canonical per-account home) **before** the real-home / system-default lane takes over the shared runtime home — on the launch path, not only on the explicit select path. Closes GAP-6 in the real-home consolidation work.

## Why

`prepareForCodexLaunch` (`runtime-home-service.ts`) returns `null` **early** for the real-home lane, *before* `syncForCurrentSelection` runs. The outgoing managed account's refreshed-token read-back normally happens inside `doSelectAccount → syncForCurrentSelection`. Any path that drops the selection to the system default **without** going through `syncForCurrentSelection` — e.g. the selection nulled out-of-band, or the auto-deselect when an active managed account's `auth.json` goes missing — hands off to the real-home lane while a freshly Codex-refreshed token sits **unpersisted** in the shared runtime home. On the next real-home launch that token is silently dropped → **token loss**.

## Change

- New `readBackOutgoingManagedHostAccountBeforeSystemDefault()` runs the managed→system-default transition (via `syncForCurrentSelection`) when a managed account is still recorded as synced (`lastSyncedAccountId !== null`), and is invoked right before the real-home lane's early `return null`.
- The real-home lane implies `host === null`, so this restores **only** Orca's runtime mirror from `~/.codex` — it **never writes the real `~/.codex`** (`lastSyncedAccountId !== null` routes to the snapshot-restore branch, never the `writeSystemDefaultAuth` path).
- No-op once the selection has already been reconciled to the system default, so the normal explicit-select path does **not** double-write.

## Test evidence

`npx vitest run --config config/vitest.config.ts src/main/codex-accounts/runtime-home-service.test.ts` → **89 passed** (86 existing + 3 new). `pnpm run typecheck:node` → clean.

New tests (all sandboxed — never touch the real `~/.codex`):
1. **managed→default at launch (not via select)** still persists a mirror-refreshed token to the outgoing account's home. Verified as a real guard: with the fix removed this test fails (outgoing home keeps the stale token).
2. **auto-deselect on missing managed auth** persists the previous account's refresh before handing off to the real home, and the follow-up real-home launch stays a no-op.
3. **normal (explicit) deselect path** does not re-run the transition on the subsequent real-home launch (spy asserts `syncForCurrentSelection` is not called again) → no double-write / no regression.

## Guardrails

- Never mutates the real `~/.codex` (asserted; restore path only reads it and writes Orca's runtime mirror).
- No regression to flag-OFF or the normal select path (covered by existing suite + test 3).
- No `max-lines` disables; follows AGENTS.md / CLAUDE.md.

## Matrix rows covered

- **Row 3** — managed → system-default transition at launch (real-home lane), token read-back preserved.
- **Row 6** — auto-deselect (active managed `auth.json` missing) → system-default handoff, outgoing refresh preserved.

### DIFF
--- a/src/main/codex-accounts/runtime-home-service.test.ts
+++ b/src/main/codex-accounts/runtime-home-service.test.ts
@@ -1102,6 +1102,176 @@ describe('CodexRuntimeHomeService', () => {
     expect(service.prepareForCodexLaunch()).toBe(getRuntimeCodexHomePath())
   })
 
+  it('persists an outgoing managed refresh at launch before the real-home lane takes over', async () => {
+    const runtimeAuthPath = getRuntimeCodexAuthPath()
+    const systemAuth = createCodexAuthJson('system@example.com', 'acct-system', 'system')
+    const managedAuth = createCodexAuthJson('user@example.com', 'acct-user', 'managed', 1)
+    const refreshedManagedAuth = createCodexAuthJson(
+      'user@example.com',
+      'acct-user',
+      'refreshed',
+      2
+    )
+    writeFileSync(getSystemCodexAuthPath(), systemAuth, 'utf-8')
+    const managedHomePath = createManagedAuth(testState.userDataDir, 'account-1', managedAuth)
+    const settings = createSettings({
+      codexSystemDefaultRealHomeEnabled: true,
+      codexManagedAccounts: [
+        {
+          id: 'account-1',
+          email: 'user@example.com',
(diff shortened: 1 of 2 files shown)


brennanb2025:

PR #9177, opened 2026-07-17
feat(cli): add `orca account add` / `account list` for headless hosts (Claude + Codex)

dommer1:
## Summary

Adds a host-local CLI for managing agent accounts on a headless Orca runtime:
`orca account add [--agent claude|codex]` and `orca account list`.

The desktop "Add account" UI is disabled whenever the renderer drives a **remote**
runtime (`isRemoteAccountScope` → active runtime target `kind: 'environment'`).
So an `orca serve` host reached from a remote desktop/web client has no way to
register managed Claude/Codex accounts — the runtime falls back to the system
default login and the account switcher stays empty. Switching (`select`) already
works remotely; only *adding* was blocked.

This adds a host-local path that reuses the existing capture logic instead of
driving the interactive browser login from inside the background runtime:

- `ClaudeAccountService.addAccountFromConfigDir()` / `CodexAccountService.addAccountFromHome()`
  register a managed account by capturing credentials from an already-authenticated
  `CLAUDE_CONFIG_DIR` / `CODEX_HOME` (persist/rollback helpers extracted and shared
  with the existing add flow).
- RPC `accounts.addClaudeFromConfigDir` / `accounts.addCodexFromHome`, bridged via
  `OrcaRuntime`; both reject mobile device tokens (host-local only).
- `orca account add` runs the agent login (`claude login` / `codex login`) in the
  user's own terminal into a temp config dir, then registers it via the local
  runtime. `orca account list` renders both Claude and Codex managed accounts.

## Screenshots

No visual change (CLI only).

## Testing

- [x] `pnpm lint` — passes (one pre-existing, unrelated react-hooks warning).
- [x] `pnpm typecheck` — passes (node/cli/web).
- [x] `pnpm test` — ran the affected suites (registry-parity, agent-context, cli args,
  Claude & Codex account services **including the new capture tests**, accounts RPC):
  all passing. Full suite not run in the headless build environment.
- [x] `pnpm build` — Linux AppImage builds and runs (`orca account --help` shows the
  new commands; `orca account list` returns live managed accounts).
- [x] Added unit tests: `ClaudeAccountService.addAccountFromConfigDir` (capture +
  rollback/rejection) and `CodexAccountService.addAccountFromHome` (auth.json import +
  rejection when missing).

## AI Review Report

Reviewed the full diff with an AI agent, focused on: reuse correctness (the new
`addAccountFromConfigDir` / `addAccountFromHome` must match the existing add flow's
persistence + rollback semantics), RPC surface exposure, and cross-platform behavior.

Findings & resolutions:
- Extracted the shared persist/rollback tail (`persistCaptured*Account`) so the new
  path cannot diverge from the login path in how it writes settings, syncs auth, and
  rolls back on error. Verified rollback still removes the managed dir/home on failure.
- Cross-platform (macOS / Linux / Windows) explicitly checked:
  - Credential storage: on Linux/Windows both agents persist plaintext credential
    files (`.credentials.json` / `auth.json`) — the capture path reads/copies those,
    matching existing behavior. On macOS Claude uses the Keychain; that branch is
    reachable but the CLI is intended for headless hosts (macOS runs the local GUI
    where `isRemoteAccountScope` is false and the GUI add flow is used), so the macOS
    Keychain temp-item cleanup path is not exercised here — noted as follow-up.
  - Paths use `node:path` (`join`/`resolve`) and env vars (`CLAUDE_CONFIG_DIR` /
    `CODEX_HOME`) — no hardcoded separators.
  - The CLI login spawn is POSIX-oriented (`spawn('claude'|'codex', args)` without a
    shell); Windows `.cmd` shim resolution for the *CLI-side* login is a follow-up
    (the runtime-side login already handles this via `getSpawnArgsForWindows`).
  - WSL is untouched: the CLI always targets the host runtime; the existing WSL
    managed-home/dir branches only trigger for `--runtime wsl` on win32.

## Security Audit

- **Command execution:** login is spawned with an args array and no shell → no shell
  injection. `claude auth status` is invoked the same way. Copying `auth.json` uses
  atomic file writes.
- **Input / path handling:** RPC params validated with zod (non-empty strings). The
  *destination* managed path is constrained by the existing `assertManagedAuthPath` /
  `assertManagedHomePath` guards (path-traversal safe, must live under Orca storage).
  The *source* dir is caller-supplied but the caller is the local host CLI (runtime
  connection); reads are limited to the credential file.
- **IPC / auth:** both new RPC methods explicitly reject `clientKind === 'mobile'`
  (device tokens), so only the local runtime connection can add accounts — matching the
  intent that add is host-local. Select/remove already existed and are unchanged.
- **Secrets:** OAuth credentials are copied as plaintext files into Orca's managed
  storage — the same as the existing Linux/Windows add flow; managed `auth.json` is
  written `0600`, temp login dirs are created via `mkdtemp` and removed in `finally`.
  No secrets are logged.
- **Dependencies:** none added.

Follow-up: dedicated unit tests for the new capture methods; Windows `.cmd` spawn
handling for the CLI-side login; macOS Keychain temp-item cleanup if the CLI path is
ever supported on macOS.

## Notes

Primary target is a headless `orca serve` host on Linux accessed from a remote client.
Verified end-to-end on such a host: `orca account add` registers an account that then
appears in `orca account list` and in the desktop switcher, and is selectable from the
remote client (select was never the blocked part).

## ELI5

On headless remote runtimes you can add and list Claude/Codex accounts with orca account add and orca account list. Managed accounts work without the desktop Add account UI, which is disabled for remote scopes.


brennanb2025:
Taking a look!

brennanb2025:
Thanks for the contribution — I took this over and pushed a maintainer pass onto the PR branch. I rebased it onto current main and kept the core temporary-home capture design, while making the headless flow production-safe: Codex now uses device authorization; paired mobile/remote-runtime clients cannot submit host credential paths; Windows uses Orca’s existing command resolver instead of enabling a shell; JSON output stays machine-readable; temporary macOS Keychain state is cleaned up and restored; and failed Codex activation restores both settings and runtime selection. I also applied the flag-registration and portable-temp-path nits, fixed WSL active-account formatting, and added focused CLI/RPC/service coverage; 280 relevant tests, full typecheck, touched-file lint, and the max-lines ratchet all pass.

brennanb2025:
Thanks for this — the headless gap is real, and capturing credentials from an already-authenticated config dir is the right shape for it. I confirmed the premise before touching anything: the desktop Add button calls a local Electron IPC, so un-gating it for remote scopes would have registered the account against the wrong machine. There's no existing remote add path to reuse.

I rebased onto latest main (the branch was 411 commits behind and conflicting) and pushed fixes. In severity order:

**Packaging — both new commands were broken in a real build.** `electron-vite` cleans `out/main` and emits only its declared entries, and `build:desktop` runs it after `build:cli`. The three new CLI→main imports had no entries, so `orca account add` and `orca account list` both died at require time with `Cannot find module '../../main/claude-accounts/keychain'`. `agent-hooks/managed-agent-hook-controls` already carries an entry with a comment explaining exactly this. Added the three missing entries and a parity test, since unit tests resolve TS sources and structurally cannot catch it.

**Interrupted login stranded live OAuth tokens on disk.** Node terminates on SIGINT without unwinding `finally`, so Ctrl-C during sign-in left the temp dir with a usable token in it, and on macOS skipped the Keychain restore. Added a signal guard covering SIGINT/SIGTERM/SIGHUP (SSH hangup matters here), using `on` rather than `once` so a second Ctrl-C can't kill the process mid-cleanup, and awaiting in-flight cleanup rather than latching a boolean.

**`account list` forced a provider refresh.** `accounts.list` awaits `refreshAccountsForMobile()`, which bypasses both the poll throttle and the per-provider `Retry-After` gate, then does one serial round-trip per account — for output that renders no usage numbers. `runtime-provider-accounts-client.ts` already avoids this call for the same reason ("can hang for minutes behind broken auth"). Added a `refreshUsage` param defaulting to true, so mobile and web are unchanged, and the CLI opts out. Older hosts declare `params: null` and ignore the field, so a newer CLI degrades rather than failing.

**Flag surface.** `account` wasn't in `supportsBrowserPageFlag`'s deny-list, so `--page` was silently accepted and advertised in help. The specs omitted `GLOBAL_FLAGS`, so `--help`/`--json` never rendered. `--agent` inherited the shared "Launch a known TUI agent in the first terminal" text. And a valueless `--agent` parsed as boolean `true` and silently ran a full Claude OAuth login for a provider the user hadn't asked for.

**Rebase conflict worth flagging:** main had moved the Codex post-add quota probe to a background call, because it can take 10-25s and must never block or fail an already-durable account write. The extracted `persistCapturedCodexAccount` had reverted it to an awaited call — restored, and moved outside the new rollback try/catch.

Also added coverage for the two desktop-flow behaviour changes that ride in the shared persist/rollback helpers — the guarded rollback rematerialization on the Claude side, and passing the account's selection target to `syncForCurrentSelection` on the Codex side (which is a genuine fix: reauthenticate and select already did this, add was the lone exception, so a WSL add materialized the host home that hadn't changed and never the WSL slot that had).

Verified on an isolated headless `--serve` host built in production order: `account list` end-to-end, the refresh opt-out honored, the new add RPC registered and reachable, `--json` keeping stdout to the envelope with the login prompt on stderr, and the interrupt case going from one token-bearing temp dir to zero with exit 130.

brennanb2025:
Follow-up round — pushed four more commits and rebased onto latest main.

The interrupt guard I added earlier had a bug of its own: its `finally` detached the signal handlers **before** running cleanup, so the two serial 3s `security` calls plus the temp-dir removal — exactly the window the guard exists to protect — ran with Node's terminate-immediately default restored. Two independent review lanes reproduced it. Cleanup now runs first and detaches in a nested `finally`, and a cleanup failure no longer replaces the error that explains why the add failed.

Also in this round:

- **Preflight the runtime before the login.** `RuntimeClient` is lazily constructed and the first call was the registration RPC itself, so "Requires the Orca runtime to be running" surfaced only *after* a full interactive OAuth round trip. Now checked up front with the newly-cheap `accounts.list { refreshUsage: false }`.
- **Survive a daemon that cannot spawn `claude`.** `allowFailure` is honored in `onClose` but not `onError`, and unlike the GUI flow nothing has run `claude` in the daemon before this point — so a launchd/systemd daemon with a minimal PATH hard-failed an add the user had already signed in for, even though identity resolves fine from the config dir's `oauthAccount`.
- **Reject `--environment` / `--pairing-code` on both commands.** `shouldIgnoreRemoteSelection` pins account commands to the local runtime, so `orca account add --environment homelab` silently registered on the laptop instead of the host it names, and `orca account list --environment homelab` silently listed the laptop's accounts. Both now error. The other command groups with the same silent-ignore behavior are left alone — out of scope here.
- Test hardening: signal listeners are identified by set difference rather than `process.listeners(sig).at(-1)`, which could pick up vitest's own once-wrapped SIGINT teardown.

Every fix has a test proven to fail when the fix is reverted. 1039 tests pass, typecheck and lint clean.

brennanb2025:
Thanks for the PR! Sorry for the delay in reviewing it

Issue #5370, opened 2026-06-14
[Bug]: Orca's private `CODEX_HOME` causes a Codex auth-revocation war with the user's `~/.codex`


simmons-OS:
### Operating system

macOS

### Orca version

1.4.68

## Summary

Orca runs its own Codex runtime under a **private `CODEX_HOME`** (`~/Library/Application Support/orca/codex-runtime-home/home`) and keeps a persistent daemon that maintains a Codex auth session bound to it. Because Codex/ChatGPT OAuth appears to enforce a single active session per account (logging in or refreshing in one home revokes the other's refresh token), Orca's private home and the user's canonical `~/.codex` **continuously revoke each other's tokens**.

The result: Codex auth silently breaks for everything that uses the default `~/.codex` outside Orca (launchd jobs, plain `codex exec`, other tooling), while `codex login status` keeps reporting "Logged in." It's very hard to diagnose because Orca **symlinks `skills/` and `plugins/` back to `~/.codex`** — so the two homes look like one, but `auth.json` / `config.toml` / `hooks.json` are separate.

## Environment

- Orca: **1.4.68**
- codex-cli: **0.136.0** (also reproduced reasoning about 0.138.0)
- macOS: 26.5.1, arm64 (Apple Silicon)
- Single OpenAI/ChatGPT account used for Codex auth in both homes

## The two homes

Orca-managed sessions get `CODEX_HOME` injected (into child process env only, not shell rc):

```
CODEX_HOME=~/Library/Application Support/orca/codex-runtime-home/home
```

That home contains:

```
auth.json          (private — separate from ~/.codex)
config.toml        (private)
hooks.json         (private)
config-sync-state.json
skills  -> ~/.codex/skills    (symlink, shared)
plugins -> ~/.codex/plugins   (symlink, shared)
```

A persistent Orca daemon process keeps this runtime alive:

```
/Applications/Orca.app/.../Resources/app.asar.unpacked/out/main/daemon-entry.js
```

Meanwhile, anything launched outside Orca (launchd agents, a plain terminal that doesn't inherit Orca's env, other CLIs) uses the Codex default home `~/.codex`.

## Steps to reproduce

1. Authenticate Codex normally (`codex login`) so `~/.codex/auth.json` is valid. Confirm a real call works: `codex exec "say ok"`.
2. Open Orca and let it start its managed Codex runtime (private `CODEX_HOME`).
[...]

PR #9179, opened 2026-07-17
feat(codex): give each managed account a self-contained CODEX_HOME; retire shared mirror (PR-E)

brennanb2025:
## What & why

Today real-home routing fires only for the **system-default** (`host === null`) account. Every **managed** account still injects the single shared runtime home (`codex-runtime-home/home`) as `CODEX_HOME` and hot-swaps `auth.json` into it on select. Two consequences:

- **GAP-1**: the shared mirror can't be deleted in Phase 2 while any managed account exists.
- **GAP-5**: two managed accounts share **one** `auth.json`, so concurrent panes for different managed accounts race that file (token refresh from one clobbers the other).

This PR makes each `codex-accounts/<id>/home` a **complete, directly-injected `CODEX_HOME`** when the real-home flag is ON. The account's home is both the launch home **and** the credential store, so codex reads/refreshes `auth.json` in place — no hot-swap, no read-back reconciliation, no shared file to race.

### Changes
- **`codex/codex-home-paths.ts`** — `syncSystemCodexResourcesIntoManagedHome(managedHomePath?)` links the system resources (skills/hooks/plugins/plugin-state/profile-v2/themes/prompts/AGENTS.md) into **any** managed home, reusing the existing ownership-marker + owned-fallback-copy discipline. It only ever reads `~/.codex` and writes symlinks/owned copies **into** the target home — never symlinks into or mutates `~/.codex`.
- **`codex-accounts/runtime-home-service.ts`** — `prepareForCodexLaunch` / `prepareForRateLimitFetch` / `syncForCurrentSelection` route a host managed account to its own home and **skip** the shared-home snapshot/hot-swap/read-back machinery. Resources + config are materialized into the per-account home at launch. A vanished `auth.json` drops the selection and falls back to the system default. `getHostCodexHomePathsForSessionDiscovery` scans every per-account home so account-scoped rollouts still surface in the AI Vault.
- **`codex/hook-service.ts`** + **`codex/hook-trust-promotion.ts`** — `install` / `getStatus` / `refreshRuntimeUserHooks` (and the promotion/provenance helpers) accept a `runtimeHomePath`, so hooks.json + config.toml + RPC-granted trust land in the per-account home. Default stays the shared mirror (flag-OFF/WSL/SSH unchanged).
- **`main/index.ts`** — the launch prep passes the resolved per-account home into the hook install.
- **`codex-accounts/service.ts`** — config mirroring into a self-contained host home uses the trust-preserving merge (`syncSystemConfigIntoManagedCodexHome`) instead of the sanitize-overwrite, so hook/project trust codex granted in that home survives account switches (no re-approval / app-server re-grant churn). WSL homes and the flag-OFF path keep the existing sanitize path.
- **`ai-vault/codex-session-root-dedup.ts`** — `codex-accounts/<id>/home` ranks as canonical managed (tier 1) alongside the shared runtime home, so hardlinked aliases collapse onto the per-account home over WSL/remote/custom roots (real home `null` still wins).

## Routing matrix

| # | Selection | Flag | Launch `CODEX_HOME` | auth |
|---|-----------|------|---------------------|------|
| 1 | system-default | OFF | shared runtime mirror | snapshot/restore (unchanged) |
| 2 | system-default | ON | `null` → real `~/.codex` | native (unchanged, PR-B) |
| **3** | **host managed** | **ON** | **`codex-accounts/<id>/home`** | **per-account, in place** |
| **4** | **host managed, 2 concurrent** | **ON** | **two distinct per-account homes** | **no shared `auth.json` → no race** |
| **5** | **host managed** | **OFF** | **shared runtime mirror + hot-swap** | **unchanged opt-out** |

Rows 3–5 are new/pinned by this PR; **unblocks Phase-2 mirror deletion** (no lane injects the shared `home` when the flag is ON).

## Guardrails
- **`~/.codex` is never mutated**: resources link/copy *from* it into the per-account home; auth stays per-account; the only write-back to `~/.codex` is the pre-existing whitelisted config-settings promotion (model/approvals), identical to what the shared home already does.
- **Ownership-marker discipline** reused for all new linking; removal (`rmSync`) of a per-account home removes symlinks as links, never following into `~/.codex`.
- **Flag-OFF** and the **system-default real-home (`null`) lane** are byte-identical to today.
- **#5370 preserved**: the nested-Orca `CODEX_HOME === ORCA_CODEX_HOME` daemon strip is value-agnostic and still fires for per-account home values.
- Config is **global across accounts, auth is per-account** — consistent with the documented "swap credentials, not config defaults" intent.

## Test evidence (sandboxed temp homes only)
`npx vitest run --config config/vitest.config.ts` (affected suites) + `pnpm run typecheck:node` — all green. New/updated coverage:
- **two managed accounts get distinct homes without racing one `auth.json`** (each home keeps its own auth; shared `auth.json` never written).
- per-account home **materializes its own resources + config** on launch; `~/.codex` untouched.
- hook-service **installs managed hooks + trust into the per-account home**, not the shared mirror or `~/.codex`.
- dedup **ranks per-account homes** and still collapses hardlinked aliases (real home still canonical).
- explicit **flag-OFF managed** regression (shared home + hot-swap).
- missing-auth active account **drops selection → real home**.
- nested-Orca **per-account `CODEX_HOME` strip** (#5370).
- `codex` + `codex-accounts` suites: 622 passed.

## Notes
- **STACKED on PR-A #9173** (config hygiene) so per-account configs aren't seeded with foreign hook trust.
- Resources link lazily per account at its first launch (only used when that home is the live `CODEX_HOME`).
- WSL managed accounts keep their per-distro lane (out of scope here; same GAP-5 class is a follow-up).

Do not merge — draft for review.

### DIFF
--- a/src/main/ai-vault/codex-session-root-dedup.test.ts
+++ b/src/main/ai-vault/codex-session-root-dedup.test.ts
@@ -167,6 +167,41 @@ describe('dedupeCodexRolloutFileAliases', () => {
 
     expect(dedupeCodexRolloutFileAliases([host, wsl], accessors)).toEqual([host, wsl])
   })
+
+  it('prefers a per-account self-contained home over other non-default homes', () => {
+    const perAccount = {
+      agent: 'codex',
+      path: '/Users/ada/Library/Application Support/orca/codex-accounts/019f0000-aaaa-bbbb/home/sessions/2026/07/01/rollout-2026-07-01T10-00-00-019f0000-1111-7222-8333-444444444444.jsonl',
+      codexHome:
+        '/Users/ada/Library/Application Support/orca/codex-accounts/019f0000-aaaa-bbbb/home',
+      hardlinkIdentity: '3:71'
+    }
+    const custom = {
+      agent: 'codex',
+      path: '/Users/ada/custom-codex/sessions/2026/07/01/rollout-2026-07-01T10-00-00-019f0000-1111-7222-8333-444444444444.jsonl',
+      codexHome: '/Users/ada/custom-codex',
+      hardlinkIdentity: '3:71'
+    }
+    expect(dedupeCodexRolloutFileAliases([custom, perAccount], accessors)).toEqual([perAccount])
+    expect(dedupeCodexRolloutFileAliases([perAccount, custom], accessors)).toEqual([perAccount])
+  })
+
+  it('keeps the real home over a per-account home when they alias', () => {
+    const perAccount = {
+      agent: 'codex',
+      path: '/Users/ada/Library/Application Support/orca/codex-accounts/019f0000-aaaa-bbbb/home/sessions/2026/07/01/rollout-2026-07-01T10-00-00-019f0000-1111-7222-8333-444444444444.jsonl',
+      codexHome:
+        '/Users/ada/Library/Application Support/orca/codex-accounts/019f0000-aaaa-bbbb/home',
+      hardlinkIdentity: '1:42'
+    }
+    const real = {
+      agent: 'codex',
+      path: REAL_HOME_ROLLOUT,
+      codexHome: null,
+      hardlinkIdentity: '1:42'
+    }
+    expect(dedupeCodexRolloutFileAliases([perAccount, real], accessors)).toEqual([real])
+  })
 })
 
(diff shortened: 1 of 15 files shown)


brennanb2025:

Issue #1083, opened 2026-04-25
[Bug]: work with Pi


agonic:
### Operating system

Windows

### Details

I dont wanna use more words. Juts want to say that Pi in Orca deleted my all pi skills and extensions made by me. I cant believe I let myself do this. It ruined everything.

nwparker:
Hey @agonic , I'm very sorry about this. I have a fix made and I'm releasing it now.
It seems to be a Windows specific issue, I'm improving our scenario coverage there. I just spent the last hour getting UTM working to make it easier for me to test without switching physical machines.

There may be some ways to still recover the files, depending on how your Windows install is confgiured. If you drop by our Discord, please message me and I'll do all that I can do to help out and more.

PR #9192, opened 2026-07-17
fix(opencode): mirror default ~/.config/opencode when OPENCODE_CONFIG_DIR is unset

### LINKED ISSUE #9186: [Bug]: OpenCode in Orca ignores default ~/.config/opencode — agents run with the wrong model when OPENCODE_CONFIG_DIR is unset
### Operating system

macOS

### Orca version

1.4.144

## Summary

When launching OpenCode in Orca without explicitly setting `OPENCODE_CONFIG_DIR`, Orca provides a bare `shared/` directory containing only the Orca status plugin. The user's default OpenCode configuration at `~/.config/opencode/` (`opencode.json`, `oh-my-openagent.json`, auth, plugins) is not visible to the OpenCode process.

## Environment

- Orca: **1.4.144**
- oh-my-openagent / oh-my-opencode (cached): **4.18.2**
- macOS: 26.5.1, arm64 (Apple Silicon)
- `OPENCODE_CONFIG_DIR` and `XDG_CONFIG_HOME` both unset

## Steps to reproduce

1. Install `oh-my-openagent` under `~/.config/opencode/`
2. Create `~/.config/opencode/oh-my-openagent.json` mapping agents to models (e.g. `sisyphus` → `opencode-go/kimi-k2.6`)
3. Create `~/.config/opencode/opencode.json` with default model settings
4. Do **not** set `OPENCODE_CONFIG_DIR` in shell startup files
5. Launch OpenCode in Orca
6. Observe: agents such as `sisyphus` still appear by name, but run under the wrong model (`claude-opus-4-7`) instead of the model mapped in `oh-my-openagent.json`

## Expected

Orca should detect the default OpenCode config directory (`XDG_CONFIG_HOME/opencode` or `~/.config/opencode`) and mirror it into the overlay alongside Orca's status plugin.

## Actual

Orca returns `~/Library/Application Support/orca/opencode-hooks/shared/` as `OPENCODE_CONFIG_DIR`, which only contains the Orca status plugin.

Agent definitions load from the package cache (`~/.cache/opencode/packages/`), which Orca does not redirect, so agents still appear by name. But the per-agent model mapping lives in `~/.config/opencode/oh-my-openagent.json`, which is invisible through the bare `shared/` dir — so the model silently falls back to `claude-opus-4-7`. Hence "the agent shows up but runs the wrong model," not a full fallback to built-in defaults.

## Impact

- User's `opencode.json` default model settings are ignored
- `oh-my-openagent.json` per-agent model mappings lose their effect
- Auth, keymaps, and other plugin configs silently disappear

inaemin:
## Summary

When a user launches OpenCode in Orca **without** explicitly setting `OPENCODE_CONFIG_DIR`, Orca handed OpenCode a bare `shared/` directory that contained only Orca's status plugin. The user's default OpenCode config at `~/.config/opencode/` (`opencode.json`, `oh-my-openagent.json`, auth, plugins) was never mirrored, so it was invisible to the OpenCode process.

Concretely, `oh-my-openagent` agents such as `sisyphus` still appeared by name (their definitions load from the un-redirected package cache `~/.cache/opencode/packages/`), but their per-agent **model mapping** — which lives in `~/.config/opencode/oh-my-openagent.json` — was dropped, so they silently ran under `claude-opus-4-7` instead of the configured `opencode-go/*` model.

This PR makes `OpenCodeHookService.buildPtyEnv()` detect the default OpenCode config directory via the same XDG fallback OpenCode/OMO use (`XDG_CONFIG_HOME/opencode` → `~/.config/opencode`) and mirror it into the source-scoped overlay alongside Orca's status plugin, exactly as it already does when the user sets `OPENCODE_CONFIG_DIR` explicitly. When no default dir exists, behavior is unchanged (bare shared plugin dir).

Fixes #9186.

## Screenshots

No visual change. This is a main-process env/overlay fix with no renderer surface.

## Testing

- [x] `pnpm lint` (oxlint clean on changed files)
- [x] `pnpm typecheck`
- [x] `pnpm test` (`src/main/opencode/hook-service.test.ts` — 29/29 passing)
- [x] `pnpm build`

Added a regression test asserting that, with no explicit `OPENCODE_CONFIG_DIR`, `buildPtyEnv` mirrors the default XDG config dir into an overlay that contains **both** the user's `opencode.json` and Orca's `plugins/orca-opencode-status.js` — not the bare `shared/` dir. Existing tests were isolated by pointing `XDG_CONFIG_HOME` at an empty temp dir in `beforeAll`, so a real `~/.config/opencode` on the dev machine can no longer leak in and flip them onto the overlay path.

Also verified end-to-end on a clean build: without the change the agent runs under `claude-opus-4-7`; with it, the agent runs under the configured `opencode-go/*` model.

## AI Review Report

Reviewed with the AI coding agent for:

- **Cross-platform (macOS/Linux/Windows):** The new `getDefaultOpenCodeConfigDir()` uses `path.join` + `os.homedir()` (no hardcoded separators) and reads `XDG_CONFIG_HOME`. It only *detects and returns a source dir*; the actual mirroring reuses the existing `mirrorUserConfig` / `overlay-mirror` path that already handles Windows junctions vs POSIX symlinks. No new shortcut, label, or shell behavior was introduced.
- **SSH/remote/local:** The change is inside `OpenCodeHookService`, invoked only on the host-local PTY path (`buildPtyHostEnv`), which callers already skip for SSH connections. No assumption of local-only execution was added.
- **Agent/integration compatibility:** Detection is gated on `existsSync(configDir)`, so users without a default OpenCode config keep the previous bare-shared behavior. The explicit-`OPENCODE_CONFIG_DIR` path is untouched. A typo'd/non-existent source dir is still passed through verbatim (no `mkdir -p` of the user's path), preserving the collision-avoidance behavior in `docs/opencode-config-dir-collision.md`.
- **Performance:** Adds one `existsSync` per PTY spawn on the fallback path; negligible and off the hot path.
- **Security:** No new input handling, command execution, or secret handling. The detected path is only used as a mirror source; overlay creation stays best-effort inside try/catch and falls back to the user's dir on failure.

## Security Audit

- **Path handling:** Source dir is derived from trusted env (`XDG_CONFIG_HOME`) / `homedir()`, joined with `path.join`. It is never used to build a shell command; it feeds the existing symlink/junction mirror, which already refuses to operate outside the overlay root during teardown (`overlay-mirror.ts`, issue #1083 hardening).
- **No writes to the user's config:** The default dir is only *read/mirrored*; Orca's plugin is written into the overlay's `plugins/` dir, and `writePluginIntoOverlay` unlinks before write so a mirrored symlink can never be followed back into the user's real files.
- **Secrets/auth:** This fix makes the user's own auth/config visible to their own OpenCode process (the intended behavior); it does not transmit or log anything. No follow-up needed.

## Notes

- Only affects the case where `OPENCODE_CONFIG_DIR` is unset and a default `~/.config/opencode` (or `$XDG_CONFIG_HOME/opencode`) exists. No behavior change when the var is set explicitly or when no default dir exists.
- The linked issue's earlier draft blamed an outdated `oh-my-opencode@3.15.2` in the package cache; that was verified to be a red herring (cache held the current 4.18.2). The real cause is the missing default-config detection fixed here.

X (Twitter): @inaemin

## ELI5

OpenCode launched from Orca without OPENCODE_CONFIG_DIR only saw a tiny shared plugin folder, not your real ~/.config/opencode setup. Default config is mirrored so plugins and auth work as outside Orca.


brennanb2025:
Taking a look!

inaemin:
Rebased onto latest `main` to resolve the conflict — sorry for the churn while you're looking.

The conflict came from #9543 (`refactor(comments): slim verbose comments in main integrations`), which condensed the comments in `hook-service.ts` at exactly the three spots this PR touches (`buildPtyEnv`'s fallback branch, the `existsSync` guard, and the `catch`). I kept `main`'s shortened comment style and re-applied only the `sourceConfigDir` logic on top, and shortened the new helper's comment to match.

**No logic changes** from the version you started reviewing. Re-verified after the rebase: `pnpm test` (hook-service 29/29), oxlint, and `pnpm typecheck` all pass.

brennanb2025:
Thanks for the fix — I took this over and pushed a refined version. The default config source is now resolved from the actual PTY environment (including XDG/HOME and static shell exports) before reusing Orca's existing source-scoped overlay, so app/shell environment differences and nested Orca sessions keep the user's config while the status plugin remains installed. I also rebased onto current main and added focused coverage; 344 relevant tests, node typecheck, touched-file lint, and the max-lines gate all pass.

inaemin:
@brennanb2025 Thanks again for taking the time to refine this and pushing the updates!

I tested the latest version locally, and it seems to be working as expected. Is there anything else needed from my side before this can be merged?

PR #9237, opened 2026-07-17
fix(agent-status): follow daemon-hosted Claude sessions to their true pane (spawn pin + fork lineage + prompt-time rebind)

### LINKED ISSUE #9236: [Bug]: Claude Code >=2.1.206 shared daemon breaks agent-status pane attribution — hooks inherit the daemon's stale ORCA_PANE_KEY
### What happened

Since Claude Code ≥ 2.1.206, new `claude` TUI sessions are transparently hosted as workers under one long-lived shared background daemon (`claude daemon run`, spare-pool prewarm, `bg-pty-host` workers). Claude's hook commands (Orca's `claude-hook.sh`) now execute inside those daemon workers — which inherit the **daemon's** environment, not the pane's. `ORCA_PANE_KEY` / `ORCA_TAB_ID` / `ORCA_AGENT_LAUNCH_TOKEN` therefore point at whichever pane first spawned the daemon (possibly days ago, possibly a different workspace).

Result: **every new Claude session posts its agent status under one stale pane key.** New panes never get a status entry (no working/waiting/done in the sidebar, sessions look undetected), and the stale pane's entry is continuously clobbered by every other session's events, including sessions from other worktrees.

### Environment

- Orca 1.4.143 (macOS Tahoe 26.5 / Darwin 25.5.0, arm64)
- Claude Code 2.1.210 (repros with any ≥ 2.1.206 daemon-takeover build; 2.1.205 is fine)
- Local terminals, `agentStatusHooksEnabled: true`, hooks installed by Orca's managed controls

### Repro

1. Run any Claude session on CC ≥ 2.1.206 so the shared daemon starts (pane A).
2. Create a new Orca terminal (pane B): `orca terminal create --worktree active --command "claude"` and send it a prompt.
3. Inspect `agent-hooks/last-status.json` (or the sidebar): pane B has **no entry**; pane B's session id + prompt appear under **pane A's** entry.

Verified mechanics on a live system:

- The pane-B TUI process has the correct `ORCA_PANE_KEY` in its env, but it is only a thin client; the session worker is a child of the shared daemon: `claude daemon run` → `bg-pty-host … -- claude --session-id <uuid>`. `ps eww` on the worker shows the **daemon's** `ORCA_PANE_KEY` (pane A) and stale `ORCA_AGENT_HOOK_PORT`.
- Hooks fire in the worker, so the POST to `/hook/claude` carries pane A's key. (The stale `ORCA_AGENT_HOOK_PORT` is survivable because `claude-hook.sh` re-sources `ORCA_AGENT_HOOK_ENDPOINT`; the pane key has no such refresh path.)
- Manually POSTing the same payload with pane B's key attributes correctly — listener plumbing is fine; only the env-derived pane identity is wrong.
- The daemon roster (`~/.claude/daemon/roster.json`) confirms interactive pane launches are dispatched from the spare pool (`"source": "spare"`) with `--session-id <uuid>` injected by the daemon, and its dispatch env forwards only an allowlist (`ANTHROPIC_*`, `AWS_*`, `CLAUDE_CONFIG_DIR`, …) — per-pane `ORCA_*` vars don't survive into workers.

### Why Orca can't see this today

`AgentHookServer` trusts the posted `paneKey` (after `legacyPaneKeyAliases` normalization). For daemon-hosted sessions there is no event that ever carries the true pane identity, so no amount of listener-side heuristics can recover it — the binding has to be established at spawn time, where Orca still knows the pane.

### Proposed fix (happy to send a PR)

Mirror the existing `legacyPaneKeyAliases` pattern with a **provider-session alias**:

1. **Spawn-side pinning** — in the local PTY spawn paths (`src/main/ipc/pty.ts`, the two `isClaudeLaunchCommand` sites): when spawning a Claude launch with a validated pane identity, mint a session UUID, append `--session-id <uuid>` (skip when the command already has `--session-id` / `--resume` / `--continue`), and record `sessionId → paneKey`.
2. **Listener-side override** — in `AgentHookServer` (`src/main/agent-hooks/server.ts`), alongside `normalizeHookBodyPaneKeyAlias`: if the hook payload's `session_id` matches a recorded binding and the posted pane key differs, rewrite `paneKey`/`tabId` to the recorded pane. Unknown session ids pass through unchanged (user-typed `claude` in a plain shell keeps today's behavior).

This also future-proofs attribution against any other agent CLI that adopts a shared-daemon architecture, since the same mechanism generalizes to any provider whose hook payload carries a session id.

Two smaller adjacent gaps noticed while debugging (can split out):
- `claudeLivePtySessionIds` never tracks sessions launched via `agentCmdOverrides` wrappers whose binary isn't literally named `claude` (e.g. `claude-bedrock`), because `isClaudeLaunchCommand` matches the bare name only.
- Daemon-hosted sessions also defeat `markClaudePtySpawned` liveness assumptions (the TUI client PTY is a thin attach; the session outlives it in the daemon).

### Workaround for affected users

`CLAUDE_CODE_DISABLE_AGENT_VIEW=1 claude` (disables Claude's daemon/agents view for that launch; hooks then run in-process with the pane's own env and attribution is correct).

BrianDai22:
## Summary

Fixes #9236.

Claude Code ≥ 2.1.206 transparently hosts new TUI sessions as workers under one shared background daemon. Hook commands now execute inside those workers, which inherit the **daemon's** environment — so every `/hook/claude` post carries the `ORCA_PANE_KEY` of whichever pane first spawned the daemon. New Claude sessions never register on their own pane (no working/waiting/done in the sidebar) and continuously clobber one stale pane's status entry, across worktrees.

Pane identity has to be established where Orca still knows the pane, then followed as the session moves. Continued live testing against CC 2.1.210–2.1.212 showed one binding moment isn't enough — sessions fork ids on resume and hop panes via the agents view — so this PR binds at three points, mirroring the existing `legacyPaneKeyAliases` pattern:

1. **Spawn-time pinning** (`src/main/ipc/pty.ts`): when a local Claude launch has a validated pane identity and no session lifecycle flags (`--session-id` / `--resume` / `--continue` / `--fork-session` / `-r` / `-c`), append a minted `--session-id <uuid>` and register the `sessionId → paneKey` binding with the hook server. Only lone `claude` invocations whose first argument is a flag or quoted prompt are pinned; compound commands (`claude && …`) and bare-word subcommands (`claude daemon status`, `claude mcp list`) pass through untouched.
2. **Fork lineage** (`src/main/agent-hooks/server.ts`): `--resume`/`--continue` FORK the session id, so the forked id arrives with no binding. Claude's daemon records each fork's parent in `<configDir>/daemon/roster.json` (`dispatch.launch.sessionId` names the parent transcript), and the payload's `transcript_path` locates the config dir — so an unknown session id deterministically inherits its parent's pane, with no knowledge of the user's `CLAUDE_CONFIG_DIR` needed.
3. **Prompt-time rebind** (`server.ts` + a probe injected from `pty.ts`): the agents/fleet view and `/resume` picker swap the session inside an existing client — no spawn, no fork, no binding. On `UserPromptSubmit` only, an unbound session binds to the live Claude pane with the freshest keyboard input (5s window; `lastInputAtByPty` and the live-Claude-PTY set both live in the PTY layer, hence the injected probe — avoids a module cycle). Mid-turn events never consult the probe; an empty probe keeps the posted key.

At ingest, if a post's payload `session_id` resolves to a binding and the posted (env-derived) pane key differs, `paneKey`/`tabId` are rewritten to the bound pane — before `legacyPaneKeyAliases` normalization, so migrated-pane aliasing still applies on top. Session ids that resolve nowhere pass through unchanged, so user-typed `claude` in a plain shell keeps today's behavior. Bindings are FIFO-bounded (512), dropped when their spawning PTY exits, and re-registered on reattach.

## Screenshots

No visual change (attribution fix — the sidebar simply shows status on the correct pane again for CC ≥ 2.1.206, and keeps following the session through resume forks and fleet-view attaches).

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Seven vitest cases in `src/main/agent-hooks/server.test.ts` (existing HTTP-driven harness): daemon-stale-key re-attribution to the pinned pane, unknown-session passthrough, binding cleanup on PTY exit, FIFO bound on the registry, roster-based fork lineage (temp config dir + real roster.json fixture), prompt-time rebind via the probe, and the no-steal guarantees (mid-turn events never probe; empty probe keeps the posted key). The spawn-side command classifier was validated standalone against 25 command shapes (pin vs. passthrough for flags/quoted prompts/env-prefixes vs. resume flows/compounds/subcommands/redirects/pipelines). I was unable to run the full pnpm suite in my sandboxed environment — CI should exercise the checklist above; happy to iterate on any failures.

Root cause and both follow-on gaps were verified against live Claude Code 2.1.210–2.1.212 daemons on macOS: the pane's TUI process carries the correct `ORCA_PANE_KEY` but is a thin client; the session worker (child of `claude daemon run`) carries the daemon's stale pane key, and hooks fire in the worker. The daemon's dispatch env forwards only an allowlist (`ANTHROPIC_*`, `AWS_*`, `CLAUDE_CONFIG_DIR`, …), so no per-pane env can survive into workers. Resume forks and fleet-view attaches were both reproduced live and both mechanisms validated in a local shim before porting here.

## AI Review Report

Developed and reviewed with Claude Code. Main risks checked: (1) session pinning must never break non-TUI invocations — guarded by a conservative appendability classifier (flag/quoted-prompt first-arg only) tested against 25 command shapes; (2) hook-ingest overrides must fail open — unknown/malformed `session_id`, unreadable/malformed roster.json, and transcript paths that don't sit under a `projects/` dir all pass the body through unchanged; (3) pane stealing — the input-rebind probe is consulted only for `UserPromptSubmit` on sessions with no live binding, only within a 5s input window, and only over panes in the live-Claude-PTY set, so background/headless sessions and mid-turn events cannot displace a user's pane (the residual case — a headless prompt landing inside the typing window of an unrelated unbound session — self-corrects at that session's next interactive turn); (4) unbounded memory — registry FIFO-bounded at 512 and cleaned on PTY exit; roster reads are per-miss, not cached. Cross-platform: path handling uses `node:path` (`dirname`/`basename`/`join`), no separators assumed; changes are local-spawn-only (`!args.connectionId`), so SSH flows are untouched.

## Security Audit

Reviewed with Claude Code. Untrusted inputs from the local hook HTTP endpoint: `session_id` (Map key for exact lookup only), `transcript_path` (used to *derive* a roster path via `dirname`×2 + a literal `daemon/roster.json` join, gated on the parent dir being named `projects`; the file is JSON-parsed read-only, never written or executed — a hostile poster can at most point the parse at a file the same user could already read, and a non-roster JSON yields no binding). No path contents are echoed back. The minted spawn UUID comes from `randomUUID()` (no user input interpolated into the command string). The PTY-layer probe exposes only `{paneKey, ptyId}` for panes Orca already tracks. Registry writes bounded (FIFO 512 + PTY-exit cleanup). Existing token auth (`x-orca-agent-hook-token`) on the hook endpoint is unchanged and still gates all posts.

## Notes

- The stale `ORCA_AGENT_HOOK_PORT` in daemon workers is survivable today because the generated hook script re-sources `ORCA_AGENT_HOOK_ENDPOINT`; pane identity had no such refresh path — this PR adds one.
- Two smaller adjacent gaps noted in #9236 (`claudeLivePtySessionIds` not tracking `agentCmdOverrides` wrappers whose binary isn't literally `claude`; daemon-hosted sessions outliving their thin-client PTY) remain out of scope here.
- Workaround for affected users until this lands: `CLAUDE_CODE_DISABLE_AGENT_VIEW=1 claude`.

## ELI5

Newer Claude Code runs sessions under a shared daemon, so status hooks all reported the wrong pane and clobbered one card. Orca rebinds status to the true pane using spawn pins, fork lineage, and prompt-time fixes.

### DIFF
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -7434,3 +7434,117 @@ describe('AgentHookServer provider-session pane attribution', () => {
     expect(internals.providerSessionPanes.has(`session-${total - 1}`)).toBe(true)
   })
 })
+
+describe('AgentHookServer fork lineage and input rebind', () => {
+  const postClaude = (
+    server: AgentHookServer,
+    payload: Record<string, unknown>
+  ): Promise<Response> => {
+    const env = server.buildPtyEnv()
+    return fetch(`http://127.0.0.1:${env.ORCA_AGENT_HOOK_PORT}/hook/claude`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-Orca-Agent-Hook-Token': env.ORCA_AGENT_HOOK_TOKEN
+      },
+      body: JSON.stringify(buildBody(payload))
+    })
+  }
+
+  it('inherits the parent binding for a resume-forked session via the daemon roster', async () => {
+    const configDir = mkdtempSync(join(tmpdir(), 'orca-claude-config-'))
+    const parentId = 'aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee'
+    const forkId = 'bbbbbbbb-cccc-4ddd-8eee-ffffffffffff'
+    const projectDir = join(configDir, 'projects', '-tmp-proj')
+    mkdirSync(projectDir, { recursive: true })
+    mkdirSync(join(configDir, 'daemon'), { recursive: true })
+    writeFileSync(
+      join(configDir, 'daemon', 'roster.json'),
+      JSON.stringify({
+        workers: {
+          [forkId.slice(0, 8)]: {
+            sessionId: forkId,
+            dispatch: { launch: { mode: 'resume', sessionId: `/x/${parentId}.jsonl` } }
+          }
+        }
+      })
+    )
+
+    const server = new AgentHookServer()
+    await server.start({ env: 'production' })
+    try {
+      server.registerProviderSessionPane(parentId, GOOD_PANE, 'pty-parent')
+      await expect(
+        postClaude(server, {
+          hook_event_name: 'UserPromptSubmit',
+          prompt: 'forked turn',
+          session_id: forkId,
+          transcript_path: join(projectDir, `${forkId}.jsonl`)
+        })
+      ).resolves.toMatchObject({ status: 204 })
+
+      expect(server.getStatusSnapshot()).toEqual([
+        expect.objectContaining({ paneKey: GOOD_PANE, prompt: 'forked turn' })
+      ])
+    } finally {
+      server.stop()
+    }
+  })
+
+  it('binds an unknown session at prompt time to the pane with fresh input', async () => {
+    const server = new AgentHookServer()
+    await server.start({ env: 'production' })
+    try {
+      server.setClaudePaneInputProbe(() => ({ paneKey: GOOD_PANE, ptyId: 'pty-live' }))
+      await expect(
+        postClaude(server, {
+          hook_event_name: 'UserPromptSubmit',
+          prompt: 'attached turn',
+          session_id: 'cccccccc-dddd-4eee-8fff-000000000000'
+        })
+      ).resolves.toMatchObject({ status: 204 })
+
+      expect(server.getStatusSnapshot()).toEqual([
+        expect.objectContaining({ paneKey: GOOD_PANE, prompt: 'attached turn' })
+      ])
+    } finally {
+      server.stop()
+    }
+  })
+
+  it('does not steal a pane for mid-turn events or when the probe is empty', async () => {
+    const server = new AgentHookServer()
+    await server.start({ env: 'production' })
+    try {
+      // Probe returns null (no recent input) — prompt stays on the posted pane.
+      server.setClaudePaneInputProbe(() => null)
+      await expect(
+        postClaude(server, {
+          hook_event_name: 'UserPromptSubmit',
+          prompt: 'no signal',
+          session_id: 'dddddddd-eeee-4fff-8000-111111111111'
+        })
+      ).resolves.toMatchObject({ status: 204 })
+      expect(server.getStatusSnapshot()).toEqual([
+        expect.objectContaining({ paneKey: PANE, prompt: 'no signal' })
+      ])
+
+      // Mid-turn tool event never consults the probe even when it has signal.
+      let probed = 0
+      server.setClaudePaneInputProbe(() => {
+        probed += 1
+        return { paneKey: GOOD_PANE, ptyId: 'pty-live' }
+      })
+      await expect(
+        postClaude(server, {
+          hook_event_name: 'PreToolUse',
+          tool_name: 'Bash',
+          session_id: 'dddddddd-eeee-4fff-8000-111111111111'
+        })
+      ).resolves.toMatchObject({ status: 204 })
+      expect(probed).toBe(0)
+    } finally {
+      server.stop()
+    }
+  })
+})
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -7334,3 +7334,103 @@ describe('AgentHookServer closed-tab suppression bound', () => {
     expect(internals.closedAgentStatusTabIds.has(`closed-tab-${total - 1}`)).toBe(true)
   })
 })
+
+describe('AgentHookServer provider-session pane attribution', () => {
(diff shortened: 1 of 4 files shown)


AmethystLiang:
looking into it

AmethystLiang:

PR #9244, opened 2026-07-17
fix(terminal): don't report agent exit on a degraded foreground scan

brennanb2025:
## Problem

Windows users running many Claude agents across ~10+ worktrees report **too many "agent done" notifications**, and notifications that fire **while the agent is still working** — clicking one navigates to the worktree and the agent is mid-task.

## Root cause (false "agent done")

The completion coordinator runs three parallel completion sources (hook / title / **process-exit**). The process-exit backstop declares "done" when the foreground process is no longer a recognized agent for two consecutive samples.

On Windows/ConPTY there is no foreground process group, so foreground detection forks `powershell.exe … Get-CimInstance Win32_Process … ConvertTo-Json` with a hard **3s** timeout (`windows-foreground-process-rows.ts`). Measured on a real Win11 24H2 box **under load** (concurrent scans + CPU pressure), that fork+scan took **2.94–4.57s** — repeatedly over the 3s budget. When it times out it's killed → returns null → and the `wmic` fallback is **removed on Win11 24H2+**, so `getForegroundProcess` drops to node-pty's `.process`, which transiently reads the **shell** name. The coordinator reads that shell as "the agent exited" and fires a false completion while the agent is still working.

Notably this does **not** require dead hooks — the process-exit suppressor is same-agent-blind (it only vetoes a *different* agent), and a "working" hook status stays fresh for 30 minutes, so neither protects a working agent from a flaky inspection. macOS is unaffected because its `ps`-based inspection is fast and never times out.

## Fix

Make a **degraded** (`available: false`) foreground scan report the **last positively-recognized agent** instead of the shell fallback, so a transient scan failure no longer looks like an exit:

- New pure helper `src/main/providers/stable-foreground-process.ts` (`resolveStableForegroundProcess`).
- Wired into `LocalPtyProvider.getForegroundProcess` via a per-PTY `ptyLastRecognizedForeground` cache (cleared on teardown / when the PTY is gone).

An **authoritative** (`available: true`) scan still refreshes or clears the memory, so a **genuine exit or crash is still detected**. This is a **no-op on macOS**, where the scan is always available — the change only engages on the Windows timeout path.

Trade-off: under *sustained* scan timeouts, a real exit registers slightly later (once a scan succeeds) — the safe direction vs. a false completion, and the hook `Stop` still fires the real completion on a Git-for-Windows box.

## Testing

- `stable-foreground-process.test.ts` — 5/5, including a regression guard that a real exit is still detected (authoritative scan clears the memory).
- 527 tests pass across the completion-coordinator suite + the full `src/main/providers` dir; `typecheck:node` clean; oxlint/oxfmt clean.

## Scope

This fixes only the **false completion**. The **flood** half — no global notification rate cap (cooldown is per-worktree), `suppressWhenFocused` covering only the active worktree, and the parked-watcher second completion path — is platform-independent and a **separate follow-up PR**.

Live validation on real Windows (the CIM-timeout load rig) is in progress against this branch.

### DIFF
--- a/src/main/providers/local-pty-provider.test.ts
+++ b/src/main/providers/local-pty-provider.test.ts
@@ -67,10 +67,8 @@ vi.mock('./windows-powershell-executable', () => ({
 }))
 
 vi.mock('./agent-foreground-process', () => ({
-  resolveAgentForegroundProcessWithAvailability: async (...args: unknown[]) => ({
-    available: true,
-    processName: await resolveAgentForegroundProcessMock(...args)
-  })
+  resolveAgentForegroundProcessWithAvailability: (...args: unknown[]) =>
+    resolveAgentForegroundProcessMock(...args)
 }))
 
 vi.mock('../wsl', () => ({
@@ -143,7 +141,10 @@ describe('LocalPtyProvider', () => {
     terminateDescendantSnapshotMock.mockReset()
     resolveAgentForegroundProcessMock.mockReset()
     resolveAgentForegroundProcessMock.mockImplementation(
-      async (_pid: number, fallbackProcess: string | null) => fallbackProcess
+      async (_pid: number, fallbackProcess: string | null) => ({
+        available: true,
+        processName: fallbackProcess
+      })
     )
 
     exitCb = undefined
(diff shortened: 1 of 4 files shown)


brennanb2025:
## Validation on real Windows 11 24H2 (branch head c96b4c3c9)

Validated by an agent on a real Windows box with the CIM-timeout load rig.

**PASS (authoritative):**
- Unit test `stable-foreground-process.test.ts` → 5/5 on Windows (degraded `available:false` → returns last recognized agent; `available:true` → refreshes/clears).
- Confirmed the scan genuinely times out under load (raw CIM `Get-CimInstance Win32_Process` measured **5.41s vs the 3s budget**).
- **No regression:** the hook `Stop` still fires real completions — `agentStatusByPaneKey` transitioned to `done` at the true end of every task. Legitimate completion is preserved.
- Symptom did not reproduce with the fix: across working tick-loops, idle, and sustained CIM-timeout load (`getForegroundProcess` returning `powershell`/`kids:false` for 14–16 consecutive samples), the pane never entered `unreadAgentCompletionPanes`.

**Caveat (honest):** the pre-fix false completion could not be reproduced in that exact rig — the process-exit backstop didn't engage as the completion driver there (hooks dominated), so live end-to-end causation isn't fully proven. The fix's correctness rests on the (authoritative) unit test plus the confirmed timeout behavior.

**Known limitation surfaced (follow-up candidate, not addressed here):** under pathological self-load (~12 `claude.exe`), ~20/30 idle probes returned the shell, consistent with an `available:true`-but-**incomplete** CIM snapshot (scan succeeds, shell root present, but the agent descendant row missing) → resolves to the shell and clears the cache. This fix guards only the `available:false` (timeout) path; it deliberately does **not** cache-fallback on `available:true`, because that is indistinguishable from a real exit and would regress genuine exit/crash detection. Likely amplified by the pathological process count; mitigated by working hooks. A safe follow-up would cache the recognized agent's **PID** and gate exit-confirmation on PID liveness, which distinguishes an incomplete snapshot from a real exit without the regression risk.

brennanb2025:
Cross-reference for the record: this fix lives in the local PTY provider, which on Windows is only the daemon-unavailable fallback path — Windows PTYs are hosted by the terminal daemon, so this change is **inert on the default Windows configuration** (it's a correct, no-regression improvement to the fallback path, just not the primary one). #9258 carries the equivalent fix on the daemon foreground path, which is what actually lands the Windows false-"agent done" fix.

PR #6562, opened 2026-06-28
Fix remote File Explorer opening files via local fs:readFile (#6440)

nwparker:
## 🧑‍🤝‍🧑 Impact & ELI5

**1) What's broken today (ELI5).** If you connect Orca to a remote machine (a "paired runtime" — for example a server you SSH into) and then click a file in that remote project's File Explorer, the editor often just shows "Unable to load file." Clicking "Retry" does nothing — it keeps failing the same way. Orca is mistakenly trying to read the file from your own laptop's disk instead of from the remote machine where the file actually lives, so it hits a permission wall ("Access denied: path resolves outside allowed directories"). This affects people working against remote/headless runtimes; it doesn't hit purely-local users. For those who do use remote runtimes, it's a hard blocker on a basic action — opening a file to look at it.

**2) What this PR does (ELI5).** Think of every open file as a package that needs a "return address" saying which machine it came from. The File Explorer was sending files off with the address left blank, so Orca guessed — and often guessed "the laptop," which was wrong for remote files. This PR makes the File Explorer write the correct return address (the remote runtime that owns the file) right on the package when you click it. Now the editor knows to fetch the file over the remote connection instead of the local disk. For files that really are local, it stamps an explicit "this one is genuinely local" flag so Orca won't accidentally re-route it to whatever remote machine happens to be focused. Oh, so it now reads remote files from the remote machine instead of silently looking on your laptop and failing.

**3) Tradeoffs / regressions — honest answer.** No regressions — nothing is disabled, no new setup, no behavior is lost. The change is one extra piece of information ("which machine owns this file") passed along an existing call; no GPU flag, fast path, or feature is turned off, and there are no new network calls, watchers, or polling. There is one subtle behavior *correction* worth naming honestly: for a local worktree, the old code would fall back to "whatever runtime is currently focused" when the owner was blank, which could route a local file's read to a focused remote runtime. This PR now marks local opens as explicitly local so they stay local. That is strictly the intended behavior (the old fallback was the bug), and it mirrors the already-shipped drag-and-drop and mobile file-open paths, so it's consistent rather than a new special case. Net: remote-runtime users gain a working file open/retry; local users are unaffected.

---

## Summary

Fixes #6440.

[...]

PR #9262, opened 2026-07-18
fix(search): open content-search matches through the worktree's owning runtime

### LINKED ISSUE #9185: [Bug]: Remote content-search result opens via local fs:readFile instead of owning runtime
### Operating system

macOS

### Orca version

Observed with a remote runtime host on v1.4.137. The affected `search-match-open.ts` path is still unchanged on current `main`.

#### Short summary

When using a paired/headless remote runtime, content search works, but clicking a search match can open the remote file through the client machine's local `fs:readFile` path instead of the owning runtime. The editor then fails with:

```text
Unable to load file
Error invoking remote method 'fs:readFile': Error: Access denied: path resolves outside allowed directories. If this blocks a legitimate workflow, please file a GitHub issue.
```

The same file opens normally when selected directly from the remote worktree's File Explorer.

#### Reproduction steps

1. Start Orca in headless/serve mode on remote host A.
2. Pair a desktop Orca client running on host B.
3. Register and open a repository/worktree on host A whose absolute path does not exist under host B's local allowed roots.
4. In the remote worktree, use the right-sidebar content search to find text inside a file.
5. Click the matching search result.
6. Observe that the editor shows `Unable to load file` and the local `fs:readFile` allowed-directories error above.
7. Open the same file directly through File Explorer; it loads successfully.

#### Expected result

Opening a content-search match from a runtime-owned worktree should preserve the worktree's runtime owner and read the file through the runtime file RPC. Retry should use the same owner.

#### Likely cause

`src/renderer/src/components/right-sidebar/search-match-open.ts` calls `openFile(...)` without passing `runtimeEnvironmentId` or suppressing the ambient local fallback:

```ts
openFile({
  filePath: fileResult.filePath,
  relativePath: fileResult.relativePath,
  worktreeId: activeWorktreeId,
  language: detectLanguage(fileResult.relativePath),
  mode: 'edit'
})
```

This appears to be the same missing-owner class fixed for normal File Explorer activation by #6562 / #6440. PR #6562 explicitly noted that the search-result open path likely had the same bug and was out of scope.

A likely fix is to resolve the active worktree's `runtimeEnvironmentId` and pass it to `openFile`, using `suppressActiveRuntimeFallback: true` for an explicitly local owner, matching the existing File Explorer behavior.

#### Additional notes

- Terminal reads on the remote host succeed because they bypass the Electron local `fs:readFile` IPC path.
- No matching open issue or PR was found before filing this report.

xianjianlf2:
## Problem

Clicking a content-search match in a remote worktree opens the file through the client machine's local `fs:readFile` path instead of the worktree's owning runtime, failing with `Error: Access denied: path resolves outside allowed directories`. The same file opens fine from the remote worktree's File Explorer.

Closes #9185

## Root cause

`openMatchResult` in `search-match-open.ts` called `openFile(...)` without a `runtimeEnvironmentId`, so the open fell back to the active/local runtime. File Explorer had this exact bug fixed in #6562 (which explicitly scoped the search path out), passing `runtimeEnvironmentId` + `suppressActiveRuntimeFallback` — the search path was left behind.

## Fix

Mirror the File Explorer pattern: `useFileSearchPanel` resolves the owner via `getRuntimeEnvironmentIdForWorktree` (the same resolver `FileExplorer.tsx` uses) and `openMatchResult` threads `runtimeEnvironmentId` through, setting `suppressActiveRuntimeFallback` when resolution returns null so we never silently fall back to the wrong machine.

## Testing

New `search-match-open.test.ts` covering: remote match opens with the owning runtime id, local match keeps current behavior, null resolution suppresses fallback. Full right-sidebar suite: 154 files / 1302 tests passing; oxlint and tsc clean.

X: @mark86202384100

### DIFF
--- a/src/renderer/src/components/right-sidebar/search-match-open.test.ts
+++ b/src/renderer/src/components/right-sidebar/search-match-open.test.ts
@@ -1,0 +1,73 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { openMatchResult } from './search-match-open'
+import type { SearchFileResult, SearchMatch } from '../../../../shared/types'
+
+beforeEach(() => {
+  vi.stubGlobal(
+    'requestAnimationFrame',
+    vi.fn(() => 1)
+  )
+  vi.stubGlobal('cancelAnimationFrame', vi.fn())
+})
+
+const fileResult: SearchFileResult = {
+  filePath: '/remote/repo/src/index.ts',
+  relativePath: 'src/index.ts',
+  matches: []
+} as unknown as SearchFileResult
+
+const match: SearchMatch = {
+  line: 3,
+  column: 5,
+  matchLength: 4
+} as unknown as SearchMatch
+
+function makeParams(runtimeEnvironmentId: string | null): {
+  openFile: ReturnType<typeof vi.fn>
+  run: () => void
+} {
+  const openFile = vi.fn()
+  const setPendingEditorReveal = vi.fn()
+  const revealRafRef = { current: null }
+  const revealInnerRafRef = { current: null }
+  return {
+    openFile,
+    run: () =>
+      openMatchResult({
+        activeWorktreeId: 'wt-1',
+        runtimeEnvironmentId,
+        fileResult,
+        match,
+        openFile,
+        setPendingEditorReveal,
+        revealRafRef,
+        revealInnerRafRef
+      })
+  }
+}
+
+describe('openMatchResult runtime ownership', () => {
+  it('opens a runtime-owned match through the owning runtime environment', () => {
+    const { openFile, run } = makeParams('env-remote-1')
+    run()
+    expect(openFile).toHaveBeenCalledWith(
+      expect.objectContaining({
+        filePath: '/remote/repo/src/index.ts',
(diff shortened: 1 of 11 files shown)


OrcaWin:
Reworked this from click-time active-worktree resolution to result-bound ownership.

The search runner now commits the exact worktree/runtime route used by `files.search` alongside its results. Result clicks reuse that captured owner, so active-worktree changes, ambient runtime changes, or later catalog loss cannot retarget a match. Explicit local and SSH routes remain non-runtime without inheriting an ambient paired runtime, and Retry preserves the opened tab's owner.

Validation:
- Reproduced on current main with isolated paired Electron/headless runtimes: owner `files.search`, local `fs:readFile` failure on click and Retry, owner `files.read` success from File Explorer.
- Fixed Electron run: owner `files.read` for initial click and Retry, zero local `fs:readFile`; Monaco content and `OpenFile.runtimeEnvironmentId` independently confirmed the owner.
- 322 focused tests plus worktree-rename ownership migration passed.
- Full typecheck, scoped lint, max-lines ratchet, and changed React Doctor passed.

PR #9263, opened 2026-07-18
fix(remote): don't classify a stale/gone remote handle as agent completion

### LINKED ISSUE #9151: [Bug]: Remote disconnect is misclassified as agent completion
## Short summary

A transient loss of a paired remote runtime/PTY handle is classified as a confirmed clean agent-process exit. Orca then emits the normal agent-completion notification and orange attention state even though the remote agent did not finish.

## Environment

- Desktop: macOS 26.5.2 (25F84), arm64
- Orca: v1.4.143
- Remote runtime: paired WebSocket environment

## What happened?

While an agent was working in a remote-runtime terminal, the remote server connection dropped/reconnected. The desktop showed the orange/done agent notification. The event was connection loss, not completion of the remote agent turn.

## How can we reproduce it?

1. Pair a desktop Orca client with a remote Orca runtime.
2. Open a remote terminal and start a long-running agent turn (OMP/Codex/Claude Code).
3. In a test environment, interrupt the remote runtime connection so process inspection for the saved handle returns `no_connected_pty` (a stale/gone handle has the same effect), then restore the connection.
4. Allow the agent process-inspection cadence to run twice.
5. Observe an agent-completion notification/orange attention state despite there being no confirmed remote process exit.

## Expected behavior

A failed/stale remote handle is an **unknown/disconnected** liveness state, not evidence that the agent exited. Orca should:

- show connection recovery/offline state as appropriate;
- preserve the last known agent state until a successful inspection or authoritative exit event; and
- not dispatch `agent-task-complete`, mark the worktree unread/done, or show the orange completion indicator.

A real, successfully inspected process exit should continue to notify normally.

## Source-level reproduction / root cause

The installed release (`v1.4.143`, commit `87688880`) and current `main` (`d0b090b0` when checked) contain the same classifier path:

1. [`isTerminalGoneError`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/renderer/src/runtime/runtime-terminal-inspection.ts#L28-L45) includes `no_connected_pty`, `terminal_handle_stale`, and `terminal_gone`.
2. [`inspectRuntimeTerminalProcess`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/renderer/src/runtime/runtime-terminal-inspection.ts#L84-L97) converts any of those errors to `{ foregroundProcess: null, hasChildProcesses: false }` — byte-for-byte the same result as a successful inspection confirming no agent process.
3. [`handleProcessInspectionResult`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/renderer/src/components/terminal-pane/agent-completion-coordinator.ts#L506-L554) treats two consecutive such results as a confirmed exit and dispatches `process-exit` with `terminalIdleConfirmed: true`.

The coordinator therefore cannot distinguish transport/handle loss from a real process exit.

## Relationship to open remote-runtime work

This is intentionally narrower than #9092 / #9093. Those cover stream reset/retry recovery after sleep or route loss. PR #9093 does not change `runtime-terminal-inspection.ts` or `agent-completion-coordinator.ts`, so the false-completion classification remains.

It is also related to #8871 / #8872 (stale mirror `pty-exit` propagating a destructive close), but this report is only about the additional false agent-completion/attention signal.

## Suggested fix and regression coverage

Carry a distinct unavailable/gone outcome from remote inspection (for example, a discriminated `state: 'unavailable'`) instead of converting it to a clean empty-process result. The completion coordinator should preserve agent evidence and avoid arming/confirming `pendingProcessExitAgent` for that outcome.

Regression cases:

- recognized remote agent → two `no_connected_pty`/stale-handle outcomes → **no** completion dispatch and agent evidence remains;
- recovery → successful inspection still sees the agent → state remains working;
- recognized agent → two **successful** empty process inspections → one normal `process-exit` completion.

xianjianlf2:
## Problem

A paired or SSH runtime disconnect could be misclassified as a clean agent completion. The viewer collapsed an unreachable/stale remote handle into the same result as an authoritative inspection proving that the agent process had exited, which could trigger a false completion notification and attention state.

Closes #9151

## Fix

- Route process inspection to the runtime that owns the terminal, through its daemon or SSH relay to the PTY provider.
- Return one atomic, discriminated inspection result so transport loss, stale handles, explicit stops, real process exits, and genuine idle completion remain distinct.
- Treat transport or routing uncertainty as unknown evidence: clear any armed idle sample and require two fresh authoritative idle samples after recovery.
- Keep daemon protocol compatibility explicit for mixed versions and fail safely when an owner cannot provide authoritative inspection.

## Testing

- Deterministic remote-completion authority oracle covering transport loss, reconnect recovery, explicit stop, real exit status, and genuine successful completion.
- Revert check: the candidate passes all four oracle scenarios; reverting the fix reproduces three false-completion failures.
- Focused regression suites: 10 files / 1,259 tests and 10 files / 404 tests passed on current `main`.
- Typecheck, full lint, and all 47 reliability gates passed.
- Headed macOS Orca server with a separately paired browser client exercised disconnect/reconnect and explicit stop; headless server parity also passed.

### DIFF
--- a/config/reliability-gates.jsonc
+++ b/config/reliability-gates.jsonc
@@ -1432,7 +1432,7 @@
         "Manual headed paired-server journey: isolated Orca desktop host + separate paired web client + real Codex process + 20-second WebSocket fault + reconnect + explicit stop",
         "pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/orca-runtime.test.ts src/main/runtime/terminal-orphan-owner.test.ts src/main/runtime/terminal-orphan-topology.test.ts src/renderer/src/runtime/web-session-terminal-orphan-recovery.test.ts src/renderer/src/runtime/web-session-terminal-orphan-mixed-version.test.ts src/renderer/src/runtime/web-session-tabs-sync.test.ts --maxWorkers=1",
         "pnpm exec vitest run --config config/vitest.config.ts src/main/providers/pty-process-inspection.test.ts src/main/daemon/terminal-host.test.ts src/main/daemon/daemon-pty-router.test.ts src/main/daemon/degraded-daemon-pty-provider.test.ts src/relay/pty-handler.test.ts src/main/daemon/daemon-pty-adapter.test.ts src/main/runtime/orca-runtime.test.ts tests/e2e/remote-agent-completion-authority.unit.test.ts src/renderer/src/runtime/runtime-terminal-inspection.test.ts src/renderer/src/components/terminal-pane/agent-completion-coordinator.test.ts",
-        "pnpm exec vitest run --config config/vitest.config.ts tests/e2e/remote-agent-completion-authority.unit.test.ts src/main/providers/pty-process-inspection.test.ts src/main/daemon/terminal-host.test.ts src/main/daemon/daemon-pty-router.test.ts src/main/daemon/degraded-daemon-pty-provider.test.ts src/relay/pty-handler.test.ts src/main/daemon/daemon-pty-adapter.test.ts src/renderer/src/runtime/runtime-terminal-inspection.test.ts src/renderer/src/components/terminal-pane/agent-completion-coordinator.test.ts src/renderer/src/lib/codex-session-restart.test.ts"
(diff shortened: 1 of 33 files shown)


OrcaWin (approved):
Thanks for the focused report and initial fix. I reworked the final patch so completion inspection is owned by the runtime that actually owns the process:

- transport loss and stale routing remain unknown rather than becoming idle evidence;
- reconnect requires fresh authoritative samples before completion;
- daemon, SSH relay, and local PTY providers share one atomic inspection contract;
- the deterministic regression oracle fails when the fix is reverted and passes on the candidate.

Validated on current `main` with the focused suites, typecheck, full lint/reliability gates, a headed paired-client flow, and headless parity.

PR #9266, opened 2026-07-18
fix(pty): terminate Windows ConPTY descendants during teardown

### LINKED ISSUE #9045: [Bug]: Worktree deletion fails with "Failed to physically stop every PTY" — orphaned claude.exe/node.exe/cmd.exe processes on Windows
### Operating system

Windows

### Orca version

1.4.143.0

### Details

**Environment**
- OS: Windows 10
- Agent CLIs used in worktrees: Claude Code, Codex, Kilo Code

**Bug**
Deleting a worktree fails consistently with:

Error invoking remote method 'worktrees:remove': Error: Failed to physically stop every PTY for worktree: <uuid>::<path>

or, on the "Force Delete" path:

Error invoking remote method 'worktrees:remove': Error: Timed out waiting for physical PTY teardown: <uuid>::<path>

This has now happened on two separate worktrees ("snaggletooth" and "anhinga"), so it appears to be reproducible rather than a one-off.

**Root cause (as far as I can tell)**
Using Sysinternals `handle.exe` on the worktree folder shows dozens of orphaned processes still holding open file handles into the worktree directory — claude.exe, node.exe, and cmd.exe, well beyond what should be running for a single session (20+ handles across repeated cmd.exe/node.exe instances). It looks like Orca is not properly terminating child processes (PTYs spawned for agent CLIs) when a worktree/session is closed, so they linger and block folder deletion.

**Steps to reproduce**
1. Run an agent (Claude Code / Codex / Kilo Code) inside an Orca worktree.
2. Close the session / try to delete the worktree from the UI.
3. Deletion fails with the PTY teardown error above.
4. Inspecting the worktree folder with `handle.exe` shows many orphaned claude.exe/node.exe/cmd.exe processes still holding file handles.

**Current workaround**
Manually force-kill the relevant processes, then delete the folder:

Stop-Process -Name claude -Force -ErrorAction SilentlyContinue
Stop-Process -Name node -Force -ErrorAction SilentlyContinue
Stop-Process -Name cmd -Force -ErrorAction SilentlyContinue
Remove-Item -Recurse -Force "<worktree path>"

A full OS restart also works but obviously isn't a practical daily workflow.

**Expected behavior**
When a worktree is closed or deleted, Orca should reliably terminate all child processes/PTYs it spawned (including nested node/cmd processes from CLI agents) before attempting folder removal, without requiring manual process killing or a system restart.

rodboev:
## Summary

- Terminate Windows agent descendants before PTY teardown closes the root shell, including detached children that fall outside the ConPTY membership list.
- Reuse the shared descendant-sweep owner in `LocalPtyProvider`, keep root-kill ownership with the caller, and preserve macOS/Linux teardown behavior.
- Keep daemon-backed worktree teardown on the same agent-owned path by inferring `launchAgent` from startup commands such as `powershell -File ...\\claude.ps1`.
- Preserve Windows descendant capture even after a startup shell exits, as long as `Win32_Process` still shows the live child chain under that root PID.
- Use node-pty's native Windows kill path for daemon immediate teardown so ConPTY ownership closes the same way in headless worktree removal as it does in the local provider.
- Stabilize the Windows descendant E2E by pinning the fixture shell to `cmd.exe` and deferring temp-dir cleanup until after Electron shutdown.

Closes #9045.

## Screenshots

No visual change

## Testing

- [x] `pnpm exec vitest run src/main/pty-descendant-termination.test.ts src/main/providers/local-pty-provider.test.ts`
- [x] `pnpm exec vitest run src/main/daemon/terminal-host.test.ts src/main/daemon/pty-subprocess.test.ts src/main/providers/windows-foreground-process-rows.test.ts src/main/daemon/daemon-server.test.ts`
- [x] `pnpm exec oxlint src/main/pty-descendant-termination.ts src/main/pty-descendant-termination.test.ts src/main/providers/local-pty-provider.ts src/main/providers/local-pty-provider.test.ts tests/e2e/agent-descendant-process-kill.spec.ts`
- [x] `pnpm exec oxfmt src/main/pty-descendant-termination.ts src/main/pty-descendant-termination.test.ts tests/e2e/agent-descendant-process-kill.spec.ts`
- [x] `pnpm run test:e2e -- tests/e2e/agent-descendant-process-kill.spec.ts`
- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`
- [x] Manual Windows worktree-delete pass with orphan-process check

## AI Review Report

- Windows teardown no longer depends on ConPTY membership alone; it snapshots the live `Win32_Process` parent chain before root kill, then unions those PIDs with attached console members.
- Local provider ownership stayed fail-closed. Descendants are only signaled while the tracked PTY still owns the root, and physical-exit waiting remains in the provider.
- Daemon-backed teardown now shares the same ownership model: agent identity is inferred from startup commands, Windows immediate kill closes ConPTY through node-pty, and startup-shell PID loss no longer blinds the descendant snapshot.
- The change stays inside existing main-process teardown code. Renderer, IPC, and non-Windows signal contracts are unchanged by the follow-up fix.

## Security Audit

- Windows descendant PIDs come from bounded host-owned sources already used by Orca, not from renderer input or shell text.
- The root PID is still excluded from descendant kills inside the shared helper.
- The new parent-chain snapshot is read-only process metadata. It does not add filesystem, network, or IPC authority.

## Notes

No visual change. The focused Windows runtime repro passed on Saturday, July 18, 2026. The headless manual worktree-delete proof also passed on Saturday, July 18, 2026 after rebuilding the `out/` bundle that `orca-dev serve` actually launches.

## ELI5

On Windows, closing a PTY could leave agent child processes running after the shell exited. Descendants are terminated during teardown so ConPTY cleanups do not orphan agent processes.

### DIFF
--- a/src/main/daemon/pty-subprocess.test.ts
+++ b/src/main/daemon/pty-subprocess.test.ts
@@ -1398,7 +1398,7 @@ describe('createPtySubprocess', () => {
     expect(proc.kill).toHaveBeenCalledTimes(2)
   })
 
-  it('propagates rejected force kills so the owner can retry', () => {
+  itOnPosixHost('propagates rejected force kills so the owner can retry', () => {
     const proc = mockPtyProcess(77)
     proc.kill.mockImplementation(() => {
       throw new Error('native fallback rejected')
@@ -1422,7 +1422,7 @@ describe('createPtySubprocess', () => {
     }
   })
 
-  it('forceKill sends SIGKILL to the child pid', () => {
+  itOnPosixHost('forceKill sends SIGKILL to the child pid', () => {
     const proc = mockPtyProcess(77)
     spawnMock.mockReturnValue(proc)
     const killSpy = vi.spyOn(process, 'kill').mockImplementation(() => true)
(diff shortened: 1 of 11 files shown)


nwparker:
Closing — superseded by W2 (#15755), and worth saying exactly how, because this PR was aiming at the right target.

Terminating ConPTY descendants during teardown is now done with a **Win32 job object**: node-pty is patched to create the shell suspended, put it in a job, and resume it, so nothing can spawn before it is inside. Teardown calls `TerminateJobObject`. That reaches the process this PR was chasing — a grandchild that spawned `detached`, left the pane's console and reparented — which neither `GetConsoleProcessList` nor a parent-pid walk can see. It also needs no process-table scrape and no identity probe, because the job names only the tree Orca created and is immune to pid recycling.

The process-table work here is superseded too: `windows-foreground-process-rows.ts` no longer shells out at all. Every Windows process query goes through one Toolhelp32 snapshot (`@vscode/windows-process-tree`), which answers in ~16 ms against the ~700 ms the PowerShell CIM scan cost.

On the **e2e spec** — un-skipping `agent-descendant-process-kill.spec.ts` on Windows is the right instinct, but our e2e lanes are all `ubuntu-latest`, so the Windows branch would never actually execute in CI. The equivalent coverage now exists somewhere that does run on the Windows runner: `src/main/windows/windows-pty-job.win32.test.ts` spawns a real ConPTY, has it spawn a real detached grandchild, and asserts the job kill takes both — plus the inverse, that a clean `exit` leaves backgrounded work alone.

Thanks for the report and the repro topology; the detached-grandchild case is the one the whole workstream is built around.

nwparker:
Superseded by the job objects and native process table in #15749/#15755; e2e coverage now lives in windows-pty-job.win32.test.ts, which runs on the Windows CI lane.

nwparker:

PR #9288, opened 2026-07-18
fix(terminal): close tabless PTYs through the live pane path

rodboev:
## Summary

`terminal close --tab` now cleans up a live floating PTY that has no renderer tab instead of failing with `terminal_tab_not_found`.

## ELI5

If you ask Orca to close a background terminal that never had a visible tab, it used to error out and leave the process running. Now it closes it properly.

## Root cause & fix

`OrcaRuntimeService.closeTerminalTab()` assumed every live handle owned a renderer `tabId`; when a PTY-backed handle had none, it threw `terminal_tab_not_found`. The fix adds a fallback to `this.closeTerminal(handle)` — the existing destructive pane-close path that kills the PTY — when no `tabId` is present. This matches the established contract that a handle-addressed CLI/automation close is explicit intent and should stay destructive, while real whole-tab retirement (which waits on renderer acknowledgement) is unchanged.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Clean rebase onto `origin/main` (base `7ab601487c`, head `bd42d2fba6`).

- Test: `reuses pane close for live PTYs that do not own a renderer tab` in `src/main/runtime/orca-runtime.test.ts`
- Pass on branch: `Tests 1 passed | 864 skipped (865)`
- Revert the fix (keep the test) → the test fails, proving it guards the fix:

```
Tests 1 failed | 864 skipped (865)
AssertionError: promise rejected "Error: terminal_tab_not_found" instead of resolving
  ↳ closeTerminalTab threw terminal_tab_not_found (orca-runtime.ts:22917)
```

- Lint: `oxlint src/main/runtime/orca-runtime.ts` clean (exit 0).

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression by construction: the change only affects the previously-throwing branch (no `tabId`), converting an exception into the already-validated destructive close path. Notifier-backed whole-tab closure still waits for renderer acknowledgement and is byte-identical, as proven by the scoped passing/failing test.

_Credit to original author @rodboev. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -21535,6 +21535,35 @@ describe('OrcaRuntimeService', () => {
     })
   })
 
+  it('reuses pane close for live PTYs that do not own a renderer tab', async () => {
+    const kill = vi.fn(() => true)
+    const closeTerminalTab = vi.fn(async () => {})
+    const runtime = new OrcaRuntimeService(store)
+    runtime.setNotifier({ closeTerminal: vi.fn(), closeTerminalTab } as never)
+    runtime.setPtyController({
+      write: () => true,
+      kill,
+      getForegroundProcess: async () => null,
+      listProcesses: async () => [
+        {
+          id: 'floating-created-pty',
+          cwd: TEST_WORKTREE_PATH,
+          title: 'Claude'
+        }
+      ]
+    })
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @rodboev — close tabless PTYs through the live pane path. Reviewed, rebased on latest `main`, and merged. Really appreciate the contribution! 🙏

PR #9289, opened 2026-07-18
fix(terminal): ignore navigation Escape for OMP and Pi

### LINKED ISSUE #9208: [Bug]: Plain Escape used for OMP view navigation is inferred as a turn interrupt (false "stopped")
## Short summary

Orca treats a single plain Escape typed into a focused terminal as interrupt intent. In OMP's TUI, single Escape is routinely view navigation: leaving a focused subagent view, closing the settings overlay, dismissing side panels. When the top-level turn is mid-tool and no status hook event lands within the 500 ms settle window, Orca rewrites the still-`working` hook row to `done` + `interrupted: true` and notifies "OMP stopped" while the turn keeps running.

## Environment

- Desktop: macOS 26.5.2 (25F84), arm64
- Orca: v1.4.143 (`87688880`); the inference/server/notification files are identical on `main` (`d0b090b0` when checked; `pty-connection.ts` differs only in unrelated hunks)
- OMP: 17.0.2 with the stock Orca-managed status extension, local terminal

## What happened?

During a long top-level OMP turn (task-tool workers running), two navigation keys each flipped the pane to done:

1. Escape to leave a focused subagent view (OMP returns to the main view; the turn continues).
2. Escape to close the `/settings` overlay.

Both produced the orange attention state and a `<worktree> - OMP stopped` notification; the dashboard row showed "Interrupted by user". OMP was still working and completed normally afterwards.

## How can we reproduce it?

1. Launch OMP from Orca's agent selector and start a turn that runs a long tool call (e.g. delegates via its `task` tool), so the pane's hook row is `working` and no hook events arrive for a while.
2. While it runs, focus the subagent view inside OMP and press Escape once to return to the main view (or open `/settings` and press Escape to close it). OMP consumes the key as navigation; the turn is not interrupted.
3. Wait ~1 s.
4. Orca marks the pane done/interrupted and sends the "stopped" notification despite the agent still running.

The false positive needs no hook event to land within 500 ms of the keypress — which is the normal condition during long tool executions, exactly when users peek at subagent transcripts.

## Why Escape is navigation in OMP

OMP 17.0.2 (`0f9fcee`):

- While a subagent is focused, Escape never interrupts: it clears typed text, else unfocuses back to the main view ([`input-controller.ts#L320-L330`](https://github.com/can1357/oh-my-pi/blob/0f9fceeea483caad531a32b050ac38558516cb5c/packages/coding-agent/src/modes/controllers/input-controller.ts#L320-L330)).
- The settings screen is a fullscreen overlay whose list handles Escape via `tui.select.cancel` (defaults `escape`/`ctrl+c`, [`keybindings.ts#L133-L136`](https://github.com/can1357/oh-my-pi/blob/0f9fceeea483caad531a32b050ac38558516cb5c/packages/tui/src/keybindings.ts#L133-L136)) — it clears search/section focus or closes the overlay, never touching the turn ([`settings-list.ts#L707-L728`](https://github.com/can1357/oh-my-pi/blob/0f9fceeea483caad531a32b050ac38558516cb5c/packages/tui/src/components/settings-list.ts#L707-L728), [`selector-controller.ts#L152-L162`](https://github.com/can1357/oh-my-pi/blob/0f9fceeea483caad531a32b050ac38558516cb5c/packages/coding-agent/src/modes/controllers/selector-controller.ts#L152-L162)).
- Escape does abort a streaming turn in the main view ([`input-controller.ts#L358-L359`](https://github.com/can1357/oh-my-pi/blob/0f9fceeea483caad531a32b050ac38558516cb5c/packages/coding-agent/src/modes/controllers/input-controller.ts#L358-L359)), so a blanket "Escape is never an interrupt in OMP" is wrong — but in that case OMP ends the turn itself, and interrupted turns still emit `agent_end` ([`event-controller.ts#L1136-L1140`](https://github.com/can1357/oh-my-pi/blob/0f9fceeea483caad531a32b050ac38558516cb5c/packages/coding-agent/src/modes/controllers/event-controller.ts#L1136-L1140) documents that ordering), which the status extension forwards as the terminal hook state.
- Double Escape is also navigation in OMP (`/tree` / `/branch` selector, [`input-controller.ts#L370-L386`](https://github.com/can1357/oh-my-pi/blob/0f9fceeea483caad531a32b050ac38558516cb5c/packages/coding-agent/src/modes/controllers/input-controller.ts#L370-L386)), so the existing OpenCode/Copilot double-Escape gate does not transfer.

## Source-level chain

v1.4.143 (`87688880`):

1. [`pty-connection.ts#L2058-L2067`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/renderer/src/components/terminal-pane/pty-connection.ts#L2058-L2067) — capture-phase keydown marks any plain Escape as `'plain-escape'` intent; the matching `\x1b` write feeds `interruptInference.observeInputIntent` ([`#L1804-L1826`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/renderer/src/components/terminal-pane/pty-connection.ts#L1804-L1826)); the acknowledged-write path infers from the exact byte even when keydown was missed ([`#L3501-L3524`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/renderer/src/components/terminal-pane/pty-connection.ts#L3501-L3524)).
2. [`agent-interrupt-inference.ts#L36-L50`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/renderer/src/components/terminal-pane/agent-interrupt-inference.ts#L36-L50) — the double-Escape requirement covers only `opencode`/`copilot`; `omp`/`pi` take the single-Escape path. A `working` + fresh baseline arms `AGENT_INTERRUPT_SETTLE_MS = 500` ([`agent-interrupt-intent.ts#L5`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/shared/agent-interrupt-intent.ts#L5)); the flush only requires the row to be byte-identical after 500 ms and fresh within `AGENT_STATUS_STALE_AFTER_MS` (30 min).
3. [`server.ts#L548-L627`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/main/agent-hooks/server.ts#L548-L627) — `AgentHookServer.inferInterrupt` re-validates the same baseline and applies `{ state: 'done', interrupted: true }`. The existing agent-specific guards don't reach OMP: Droid ignores Ctrl+C (`#L565-L567`), OpenCode/Copilot reject single Escape (`#L571-L577`), and the working-subagents bail-out (`#L595-L597`) has nothing to check for OMP — the Orca-managed OMP status extension never posts a `subagents` field (verified in the installed v1.4.143 artifact and in the generator source on `main`), so OMP hook rows carry no child rows.
4. [`notification-options.ts#L60-L70`](https://github.com/stablyai/orca/blob/8768888054be564b7529f23ae43443ba6d74b797/src/main/ipc/notification-options.ts#L60-L70) — `done` + `agentInterrupted` renders the "stopped" notification; the synthesized done also flows through the completion coordinator into unread/attention state.

The 500 ms unchanged-row guard is designed to yield to the agent's own hooks, but it inverts for OMP: during long tool executions the status extension posts nothing between lifecycle events, so the row is guaranteed unchanged — the synthesized interrupt fires whenever no hook event lands within 500 ms of the Escape. A genuine main-view abort produces its own `agent_end` (interrupted turns emit it too); a navigation Escape during a long tool execution produces no event at all, so it reliably passes the guard.

This matches the mechanism's provenance: PR #2263, which introduced input-intent inference, states "Pi has not been live-tested for this PR. No manual interrupt behavior claim is made for it."

## Expected behavior

Escape that the agent consumes for view navigation must not retire a `working` hook row or notify "stopped". For agents whose integrations report their own turn termination (interrupted OMP turns emit `agent_end`, which the same status extension forwards), the input-intent fallback should not synthesize completions.

## Suggested fix and regression coverage

Options, roughly in preference order:

- Exclude `omp`/`pi` plain Escape from inference (same shape as the Droid Ctrl+C exclusion in `shouldIgnoreInterruptIntent` / `inferInterrupt`). Single Escape is ambiguous in OMP and double Escape is a navigation gesture, so neither existing gate fits.
- Or require corroboration before synthesizing for hook-backed agents (terminal title transition to idle, or a hook-silence window substantially longer than 500 ms).
- Or keep the row rewrite but suppress the OS "stopped" notification for inferred (non-hook) interrupts.

Regression cases:

- `working` OMP row, single plain Escape, no hook event for 500 ms → row stays `working`, no completion/notification;
- OMP main-view abort → hook-driven done handling unchanged;
- OpenCode/Copilot double-Escape and Droid Ctrl+C behavior unchanged.

## Relationship to open work

- #9181 — same symptom family on OMP 17.0.2 (false top-level completion) via a different path: child-session hook settlement misattribution, no keypress involved. Both can fire independently on current releases.
- #9151 — false completion from remote-handle loss (process-inspection path), unrelated trigger.
- #1141 / #1165 — the original motivation for input-intent inference (agents that miss cancellation hooks); OMP is not in that class.

rodboev:
## Summary

- Stop plain Escape used for OMP and Pi navigation from being inferred as a turn interruption.
- Keep terminal input, unread clearing, hook-driven completion, Ctrl+C handling, and other agents' interrupt policies unchanged.
- Recheck the exclusion in main so direct or stale-renderer inference requests cannot synthesize a false stopped result.

Closes #9208.

## Screenshots

No visual change

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Focused validation run after the rebase onto current main:

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/agent-interrupt-inference.test.ts src/main/agent-hooks/server.test.ts` - passed, `303` tests passed
- `pnpm run typecheck:web` - passed
- `pnpm run typecheck:node` - passed
- `pnpm exec oxlint src/renderer/src/components/terminal-pane/agent-interrupt-inference.ts src/renderer/src/components/terminal-pane/agent-interrupt-inference.test.ts src/main/agent-hooks/server.ts src/main/agent-hooks/server.test.ts && pnpm exec oxfmt --check src/renderer/src/components/terminal-pane/agent-interrupt-inference.ts src/renderer/src/components/terminal-pane/agent-interrupt-inference.test.ts src/main/agent-hooks/server.ts src/main/agent-hooks/server.test.ts` - passed

A server-level test also asserts that after an ignored plain-Escape inference for OMP and Pi, a subsequent `agent_end` hook still transitions the row to `done`, so hook-driven completion is verified end to end rather than only in isolation.

## AI Review Report

Reviewed against the touched renderer and main interrupt-inference paths. The diff is limited to a plain-Escape ignore guard for OMP and Pi plus regression coverage that proves those agents no longer infer interrupts from navigation Escape, while Ctrl+C for OMP and Pi, Droid's Ctrl+C exclusion, and OpenCode/Copilot double-Escape behavior remain intact. Cross-platform review stays clean on macOS, Linux, and Windows because the change only adjusts shared interrupt-intent policy, without changing shortcuts beyond the existing plain-Escape classification, labels, paths, shell behavior, or Electron-only code paths.

## Security Audit

Reviewed for bypass risk. Main keeps the authoritative plain-Escape exclusion, so a direct or stale renderer inference request still cannot synthesize a false stop for OMP or Pi. The change adds no command execution, path handling, credential flow, dependency, or new IPC/API surface.

## Notes

The guard is intentionally limited to plain Escape for OMP and Pi. It does not declare Escape globally non-interrupting and does not suppress genuine hook-reported completion.

## ELI5

Pressing Escape while navigating in OMP or Pi was treated as interrupting the agent turn. Plain Escape for those agents is ignored for interrupt inference so navigation keys do not fake a stop.

### DIFF
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -942,6 +942,137 @@ describe('AgentHookServer listener replay', () => {
     }
   })
 
+  it.each(['omp', 'pi'] as const)('rejects plain Escape inference for %s', (agentType) => {
+    vi.useFakeTimers()
+    vi.setSystemTime(1_000)
+    try {
+      const server = new AgentHookServer()
+      server.ingestRemote(
+        {
+          paneKey: PANE,
+          tabId: 'tab-1',
+          worktreeId: 'wt-1',
+          payload: { state: 'working', prompt: 'long task', agentType }
+        },
+        'conn-1'
+      )
+      const baseline = server.getStatusSnapshot()[0]
+
+      vi.setSystemTime(1_500)
+      const applied = server.inferInterrupt({
+        paneKey: PANE,
+        baselineUpdatedAt: baseline.receivedAt,
+        baselineStateStartedAt: baseline.stateStartedAt,
+        baselinePrompt: 'long task',
+        baselineAgentType: agentType,
+        intent: 'plain-escape'
+      })
+
+      expect(applied).toBe(false)
+      expect(server.getStatusSnapshot()).toEqual([
+        expect.objectContaining({
+          state: 'working',
+          prompt: 'long task',
+          agentType
+        })
+      ])
+    } finally {
+      vi.useRealTimers()
+    }
+  })
+
+  it.each(['omp', 'pi'] as const)('still accepts Ctrl+C inference for %s', (agentType) => {
+    vi.useFakeTimers()
+    vi.setSystemTime(1_000)
+    try {
+      const server = new AgentHookServer()
+      server.ingestRemote(
+        {
+          paneKey: PANE,
+          tabId: 'tab-1',
+          worktreeId: 'wt-1',
+          payload: { state: 'working', prompt: 'long task', agentType }
+        },
+        'conn-1'
+      )
+      const baseline = server.getStatusSnapshot()[0]
+
+      vi.setSystemTime(1_500)
+      const applied = server.inferInterrupt({
+        paneKey: PANE,
+        baselineUpdatedAt: baseline.receivedAt,
+        baselineStateStartedAt: baseline.stateStartedAt,
+        baselinePrompt: 'long task',
+        baselineAgentType: agentType,
+        intent: 'ctrl-c'
+      })
+
+      expect(applied).toBe(true)
+      expect(server.getStatusSnapshot()).toEqual([
+        expect.objectContaining({
+          state: 'done',
+          prompt: 'long task',
+          agentType,
+          interrupted: true
+        })
+      ])
+    } finally {
+      vi.useRealTimers()
+    }
+  })
+
+  it.each(['omp', 'pi'] as const)(
+    'ignored navigation Escape does not block %s agent_end',
+    (agentType) => {
+      vi.useFakeTimers()
+      vi.setSystemTime(1_000)
+      try {
+        const server = new AgentHookServer()
+        server.ingestRemote(
+          {
+            paneKey: PANE,
+            tabId: 'tab-1',
+            worktreeId: 'wt-1',
+            payload: { state: 'working', prompt: 'long task', agentType }
+          },
+          'conn-1'
+        )
+        const baseline = server.getStatusSnapshot()[0]
+
+        vi.setSystemTime(1_500)
+        const applied = server.inferInterrupt({
+          paneKey: PANE,
+          baselineUpdatedAt: baseline.receivedAt,
+          baselineStateStartedAt: baseline.stateStartedAt,
+          baselinePrompt: 'long task',
+          baselineAgentType: agentType,
+          intent: 'plain-escape'
+        })
+        expect(applied).toBe(false)
+
+        const event = normalizeHookPayload(
+          createHookListenerState(),
+          agentType,
+          buildBody({ hook_event_name: 'agent_end' }),
+          'production'
+        )
+        if (!event) {
+          throw new Error('normalizeHookPayload rejected a known-good agent_end fixture')
+        }
+        server.ingestRemote(event, 'conn-1')
+
+        expect(server.getStatusSnapshot()).toEqual([
+          expect.objectContaining({
+            state: 'done',
+            agentType
+          })
+        ])
+      } finally {
+        vi.useRealTimers()
+      }
+    }
+  )
+
   it('does not let late same-turn working hooks resurrect an inferred interrupt', () => {
     vi.useFakeTimers()
     vi.setSystemTime(1_000)
--- a/src/main/agent-hooks/server.ts
+++ b/src/main/agent-hooks/server.ts
@@ -789,6 +789,10 @@ export class AgentHookServer {
     if (agentType === 'droid' && request.intent === 'ctrl-c') {
       return false
     }
+    // Why: OMP and Pi consume plain Escape for TUI navigation; their real stops still arrive through hooks.
+    if ((agentType === 'omp' || agentType === 'pi') && request.intent === 'plain-escape') {
+      return false
+    }
     // Why: these agents use the first Escape as a TUI cancel that can leave the turn running; only a double Escape infers an interrupt.
     if (
       (agentType === 'opencode' || agentType === 'copilot') &&
--- a/src/renderer/src/components/terminal-pane/agent-interrupt-inference.test.ts
+++ b/src/renderer/src/components/terminal-pane/agent-interrupt-inference.test.ts
@@ -191,6 +191,25 @@ describe('agent interrupt inference', () => {
     entry = undefined
   })
 
+  it.each(['omp', 'pi'] as const)('does not infer plain Escape for %s', (agentType) => {
+    vi.useFakeTimers()
+    let entry: AgentStatusEntry | undefined = makeEntry({ agentType })
+    const inferInterrupt = vi.fn()
+    const tracker = createAgentInterruptInference({
+      paneKey: PANE_KEY,
+      getStatusEntry: () => entry,
+      inferInterrupt,
+      now: () => 1_100
+    })
+
+    tracker.observeInputIntent('plain-escape')
+    vi.advanceTimersByTime(500)
+
+    expect(inferInterrupt).not.toHaveBeenCalled()
+    tracker.dispose()
+    entry = undefined
+  })
+
   it.each(['opencode', 'copilot'] as const)(
     'infers immediately on double Escape for %s',
     (agentType) => {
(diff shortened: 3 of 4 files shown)


mmarabel:
I reproduced this exact false interruption on Pi in Orca 1.4.174. The sanitized recording and Pi-specific steps are now on #9208: https://github.com/stablyai/orca/issues/9208#issuecomment-5203304341

The two-layer change here is the right shape for Pi: the renderer drops Pi/OMP plain Escape before the settle timer, and the main-process guard rejects direct or stale inference requests. Ctrl+C and hook-driven completion remain unchanged.

The current review blocker is branch freshness, not the design. GitHub reports this PR as conflicting; it is 1,294 commits behind main. A current-main merge check finds one content conflict, in src/main/agent-hooks/server.ts. When rebasing, retain current main's restored-status, Claude question/cron, active-child, and Codex reconciliation guards, then place the Pi/OMP plain-Escape exclusion immediately after the existing Droid Ctrl+C exclusion and before the OpenCode/Copilot double-Escape check. The other three touched files auto-merge.

After rebasing, rerun the focused Vitest files, node/web typechecks, oxlint, and oxfmt already listed in the PR description and update the counts. For extra reviewer confidence, it would be useful to mention or assert that Pi agent_end still transitions the row to done after an ignored navigation Escape; current Pi extension coverage already verifies that agent_end is emitted.

With that rebase and fresh validation, I do not see a need to broaden or redesign this patch. @brennanb2025

rodboev:
Thanks for the detailed rebase guidance, that matched what the conflict turned out to be exactly.

Rebased onto current main as head `e89e42b0f`. The only content conflict was in `src/main/agent-hooks/server.ts`, resolved per your instructions: current main's restored-status, Claude question/cron, active-child, and Codex reconciliation guards are all retained, and the OMP/Pi plain-Escape exclusion sits between the Droid Ctrl+C check and the OpenCode/Copilot double-Escape check. The other three files rebased cleanly.

Also added the assertion you suggested: a new test proves that after an ignored plain-Escape inference for OMP and Pi, a subsequent `agent_end` hook still transitions the row to `done` in `src/main/agent-hooks/server.test.ts` (the sequence now covers rejected Escape, then hook-driven completion, for both agent types).

Validation rerun after the rebase is in the Testing section of the description.

nwparker:
Thank you for this, and sorry it sat so long.

**You got here first for OMP and Pi, and the rule that just merged is yours.** #20149 landed as fe4237cd41e, and you are a co-author on the commit.

Two things you did that we adopted directly:

**The rule itself** — a plain Escape for OMP and Pi is navigation, not an interrupt, so it must never synthesise a stopped row. That is what shipped.

**The two-sided placement, which was better than ours.** Our first revision guarded only in the main process. You did both: a renderer-side skip *and* a main-side recheck, so a stale or direct inference request cannot route around the renderer. We took your shape. The merged version does need one adjustment — the renderer skip is scoped to `state === 'working'`, because Claude's AskUserQuestion dismissal is also a plain Escape but on a `waiting` row, and an unscoped skip swallows it.

Your instinct also caught something we only found later. Because the renderer skip and the existing `droid` Ctrl+C skip shared a `clearPending()`, widening that predicate made a previously near-unreachable branch common: Ctrl+C followed by an Escape inside the 500 ms settle window silently dropped the real interrupt. We split the two skips so `droid` keeps its exact prior behaviour and a navigation Escape never revokes a pending Ctrl+C. Escape is not a retraction.

The merged change combines your half with @brennanb2025's Claude half (#15674), adds Prime Agent, and brings wire-level and in-app verification. The design is yours and theirs.

Closing as superseded by the merged change.

Issue #9294, opened 2026-07-18
i18n: sentence fragments concatenated in code cannot be translated into SOV languages


hanjoonchoe:
## Problem

Several UI strings are built by concatenating separately-localized sentence fragments in JSX, which makes them impossible to translate correctly into languages with different word order (Korean, Japanese, ...).

Example — `src/renderer/src/components/sidebar/RemoveFolderDialog.tsx`:

```tsx
{translate('...e62415c3d0', 'This only removes')}{' '}
<span className="break-all font-medium text-foreground">{displayName}</span>{' '}
{translate('...fromOrcaSsh', 'from Orca. Its files stay on {{value0}} — re-add that SSH host to recover it.', ...)}
```

In Korean the natural rendering is `이 작업은 {{name}}을(를) Orca에서 제거만 합니다. …` — the verb belongs at the end and the name in the middle, so no translation of the fragments `"This only removes"` / `"from Orca. …"` can ever compose into a grammatical sentence. These fragment keys were deliberately left in English during the Korean catalog pass in #9019 for exactly this reason.

Known affected keys (from auditing ko.json in #9019):

- `auto.components.sidebar.RemoveFolderDialog.e62415c3d0` + `fromOrcaSsh` + `8c097ef04e` (the example above)
- `auto.components.GitHubItemDialog.03d7216d62`, `auto.components.PullRequestPage.eec3706a6a`, `auto.components.TaskPage.844dc193c7` — `"{{value0}} PR #{{value1}}?"` where `value0` is a separately-localized verb

## Proposed fix

Replace fragment concatenation with one full-sentence key per variant, interpolating the dynamic parts (`{{name}}`, `{{host}}`), so each locale controls its own word order. Inline emphasis (the styled project name) can be preserved by splitting the translated sentence on the interpolated token instead of concatenating fragments around it.

I'll send a PR for the `RemoveFolderDialog` case (plus Korean translations for the new keys) as the first slice; the verb-interpolation dialogs can follow the same pattern afterwards.

PR #9296, opened 2026-07-18
fix(i18n): use full-sentence keys in RemoveFolderDialog description

hanjoonchoe:
## Summary

First slice of #9294. The Remove Project dialog description was assembled from three concatenated fragments:

```tsx
{translate('...', 'This only removes')} <span ...>{displayName}</span> {translate('...', 'from Orca. ...')}
```

No SOV-order locale (Korean, Japanese, ...) can translate these fragments into a grammatical sentence — in Korean the verb belongs at the end and the name mid-sentence. These keys were deliberately left untranslated in #9019 for exactly this reason.

This replaces the fragments with **one full-sentence key per variant** interpolating the project name (`{{name}}`) and SSH host (`{{host}}`). The inline emphasis on the name is preserved by interpolating a U+0000 sentinel and splitting the translated sentence on it — locales can now place the name anywhere in the sentence and the styled span follows.

Also:
- adds hand-written Korean, Japanese, Chinese, and Spanish translations for the two new keys (Korean reviewed by a native speaker)
- removes the three orphaned fragment keys from all five catalogs

## Screenshots

English rendering is unchanged (same sentence, same styled name — the sentence is just no longer assembled from fragments). Korean now shows a grammatical sentence: `이 작업은 ⟨name⟩을(를) Orca에서 제거만 합니다. ...` instead of English fragments around the name.

## Testing

- [x] `pnpm run verify:localization-catalog` — key references and 5-locale parity verified (10,837 keys)
- [x] `pnpm run verify:localization-coverage` — passed with 0 allowlisted candidates
- [x] `oxlint` (incl. react-doctor) and `oxfmt` clean via lint-staged
- [x] `pnpm tc:web`
- [x] Sidebar renderer test suite: 1,478 passed; the 13 failures in `SidebarToolbar.test.tsx`/`LinearAgentSkillSetupPrompt.reminder-toast.test.tsx` are pre-existing on unmodified `main` in my environment (`window.localStorage.clear is not a function` under Node 25) and unrelated to this change

## Code Review Summary

- Cross-platform: string/JSX-only change; no platform-specific code paths.
- Local/remote/SSH: both dialog variants (local disk vs SSH host) covered; the SSH variant keeps the host label including the removed-target (ghost) fallback.
- Agents/integrations/providers: not applicable.
- Performance: two string ops per dialog render; negligible.
- UI quality: identical visual output in English; the styled project name span is preserved in all locales via the sentinel split. U+0000 cannot occur in a real project name, so the split is unambiguous.
- Security: no code paths beyond string formatting; no dependencies.

## Notes

- Remaining fragment sites from #9294 (`{{value0}} PR #{{value1}}?` verb-interpolation dialogs) can follow the same pattern in a follow-up.

### DIFF
--- a/src/renderer/src/components/sidebar/RemoveFolderDialog.tsx
+++ b/src/renderer/src/components/sidebar/RemoveFolderDialog.tsx
@@ -11,6 +11,10 @@ import { Button } from '@/components/ui/button'
 import { useAppStore } from '@/store'
 import { translate } from '@/i18n/i18n'
 
+// Why: interpolated into the sentence so locales control where the name sits;
+// U+0000 cannot appear in a real project name, so the split is unambiguous.
+const NAME_TOKEN = '\u0000'
+
 const RemoveFolderDialog = React.memo(function RemoveFolderDialog() {
   const activeModal = useAppStore((s) => s.activeModal)
   const modalData = useAppStore((s) => s.modalData)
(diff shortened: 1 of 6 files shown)


AmethystLiang:
thanks for the fix. very helpful

PR #9303, opened 2026-07-18
fix(emulator): support Xcode 27 SimulatorKit relocation

Nedian0Brien:
## Summary

Update `serve-sim` from 0.1.40 to 0.1.42 so Orca's iOS emulator helper works with Xcode 27, and materialize the package's runtime dependency closure alongside the relocated helper.

## Root cause

The helper bundled by `serve-sim` 0.1.40 links `SimulatorKit` through an `@rpath` that points to Xcode's former `Developer/Library/PrivateFrameworks` location. Xcode 27 moved the framework to `Contents/SharedFrameworks`, so the helper exits before starting with `dyld: Library not loaded: @rpath/SimulatorKit.framework`.

`serve-sim` 0.1.42 includes EvanBacon/serve-sim#90. That change removes the version-specific load command and loads the simulator frameworks dynamically from the active Xcode installation.

Unlike 0.1.40, version 0.1.42 imports runtime packages such as `ws`. Orca runs a materialized copy of `serve-sim` from Application Support, outside the app bundle, so Node's ESM resolver cannot find dependencies that remain under `Orca.app/Contents/Resources/node_modules`. The materializer now copies the dependency closure declared by `serve-sim` into the relocated runtime.

## User impact

The embedded iOS simulator stream can start under Xcode 27 without requiring a global `DYLD_FRAMEWORK_PATH` workaround.

## Validation

- Confirmed the installed dependency resolves to `serve-sim` 0.1.42.
- Confirmed `otool -L node_modules/serve-sim/bin/serve-sim-bin` no longer contains an `@rpath/SimulatorKit` load command.
- Started the helper against a booted iOS 27 iPhone 17 Pro using Xcode 27.0 (27A5218g); framebuffer capture, HID setup, JPEG encoding, and the HTTP stream all started successfully.
- Exercised the materialized helper through the installed Orca Electron-as-Node executable; attach, list, and kill all succeeded.
- Emulator-focused tests: 23 passed, including a regression test for direct and transitive runtime dependencies.
- Full test suite: 31,655 passed, 52 skipped.
- `pnpm typecheck`: passed.
- `pnpm build:desktop`: passed.
- `git diff --check`: passed.

### Existing baseline issues observed

- `pnpm lint` reaches an existing switch-exhaustiveness error in `src/renderer/src/components/skills/skill-freshness-group.tsx`. This PR does not modify that file.
- `pnpm build` completes the desktop build, then the existing Xcode 27 native helper build fails when `lipo` looks for architecture-specific outputs under paths that Swift no longer produced. This PR does not modify the native build scripts.

## AI code review summary

The diff contains the dependency bump, materializer support for the declared runtime dependency closure, and a regression test. The copy is driven by package metadata and is limited to packages already bundled with Orca. No UI code changes.

## Security audit

This updates an existing direct dependency from the same publisher and preserves lockfile integrity metadata. Materialization copies only the package.json-declared dependency closure already present in the signed app resources; it performs no runtime downloads. No permissions, network endpoints, credentials, or executable invocation arguments are changed by this PR.

## Visual changes

None.

## X handle

N/A (no X handle configured on the contributor's GitHub profile).

## ELI5

Xcode 27 moved SimulatorKit, so Orca's iOS emulator helper failed to start. Bumping serve-sim and its dependencies restores the helper against the new framework location.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "posthog-node": "^5.33.3",
     "qrcode": "^1.5.4",
     "react-i18next": "^17.0.8",
-    "serve-sim": "^0.1.40",
+    "serve-sim": "^0.1.42",
     "sherpa-onnx": "1.12.37",
     "ssh2": "^1.17.0",
     "tweetnacl": "^1.0.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -74,8 +74,8 @@ importers:
         specifier: ^17.0.8
         version: 17.0.8(i18next@26.3.1(typescript@7.0.2))(react-dom@19.2.7(react@19.2.7))(react@19.2.7)(typescript@7.0.2)
       serve-sim:
-        specifier: ^0.1.40
-        version: 0.1.40(typescript@7.0.2)
+        specifier: ^0.1.42
+        version: 0.1.42(typescript@7.0.2)
       sherpa-onnx:
         specifier: 1.12.37
         version: 1.12.37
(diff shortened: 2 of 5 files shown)


Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17587. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @Nedian0Brien. No action is requested from you. Thank you.

PR #9325, opened 2026-07-18
fix(rate-limits): refetch claude usage when the last live Claude PTY exits

xianjianlf2:
## Summary

Fixes the Claude usage panel getting stuck on "Waiting for Claude session" for up to ~15 minutes (indefinitely if unfocused) after the last live `claude` terminal exits.

## ELI5

When you close your last Claude terminal, the app should immediately refresh how much Claude usage you have left. Before, it sat showing a stale "waiting" message for a long time; now it updates right away.

## Root cause & fix

While a live Claude CLI owns the single-use OAuth refresh token, the managed usage fetch is deferred and returns a terminal-classified error (`usageMetadata.deferredByLiveClaudeSession=true`) that only clears after the 15-minute failure backoff, and retries are window-focus-gated. The fix adds a drain-listener to `live-pty-gate` that fires on the live-Claude-PTY 1→0 transition — covering both `markClaudePtyExited` and `confirmSeededClaudeLivePtys` reconciliation. It is wired in `index.ts` to `RateLimitService.refreshAfterClaudeLivePtysDrained()`, which — only when the current claude result was deferred by a live session — resets the failure streak and force-refetches immediately.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Clean rebase onto `origin/main`.

- Suites: `live-pty-gate.test.ts` and `rate-limits/service.test.ts`.
- On branch: **74/74 tests passing** across both files.
- Reverting the fix files (keeping tests) fails as expected — proving the tests exercise the fix:

```
On branch:  Test Files 2 passed (2), Tests 74 passed (74).

Fix reverted:  Test Files 2 failed (2), Tests 6 failed | 68 passed (74).
  - live-pty-gate: onLiveClaudePtysDrained is not a function
  - service: service.refreshAfterClaudeLivePtysDrained is not a function
```

- `oxlint` clean (exit 0) on the three changed source files.

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression by design: the new refetch path is gated on `deferredByLiveClaudeSession`, making it a no-op for every non-deferred/absent claude state (covered by two negative tests). The drain signal only fires on the true 1→0 transition (`hadLivePtys && size===0`), verified not to re-fire on redundant exits, and listener registration uses optional chaining on `rateLimits`. No residual risk found.

_Credit to original author @xianjianlf2. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved. No additional fix was layered on this session._

### DIFF
--- a/src/main/claude-accounts/live-pty-gate.test.ts
+++ b/src/main/claude-accounts/live-pty-gate.test.ts
@@ -8,6 +8,7 @@ import {
   isClaudeAuthSwitchInProgress,
   markClaudePtyExited,
   markClaudePtySpawned,
+  onLiveClaudePtysDrained,
   seedLiveClaudePtysFromPersistence
 } from './live-pty-gate'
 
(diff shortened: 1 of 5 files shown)


nwparker:
Thanks so much for this, @xianjianlf2 — refetch claude usage when the last live Claude PTY exits. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #9336, opened 2026-07-18
fix(windows): detect ConPTY fallback and show warning toast (#9299)

### LINKED ISSUE #9299: Windows: tmux-style terminal multiplexers fail on some Windows versions
## Summary

tmux-style terminal multiplexers (session attach/detach, persistent panes inside a terminal) do not work reliably in Orca on **some Windows versions**. Behavior can range from the multiplexer never starting, attach failing, to broken display / input once a session is open.

This is the class of tools people use for durable terminal sessions under a host app (similar to how tools such as **herdr** sit on top of a multiplexer model). The goal is solid support for that workflow on Windows, not a specific third-party product.

## Expected

- Users can start/attach a tmux-style session in an Orca terminal on supported Windows hosts.
- Attach/detach and multi-pane layouts work without corrupting the PTY, scrollback, or input path.
- Failures surface a clear error (e.g. missing binary, unsupported host shell, PTY capability gap) instead of a silent broken UI.

## Actual

On some Windows versions, the tmux-style workflow is broken or unusable inside Orca terminals. macOS/Linux paths appear fine for the same workflow.

## Likely investigation areas

- Local Windows PTY spawn / ConPTY vs WinPTY differences across Windows builds
- WSL terminal vs native Windows shell (PowerShell / cmd / Git Bash) when the multiplexer is expected to run
- PATH / binary discovery for `tmux` (or equivalent) on host vs WSL
- OSC / alternate-screen / resize handling that multiplexers rely on
- Whether Orca’s terminal stack strips or mishandles sequences the multiplexer needs

## Reproduce (starting point)

1. On an affected Windows version, open a terminal in Orca (note host shell: native Windows vs WSL).
2. Start or attach a tmux-style session (or equivalent durable multiplexer).
3. Observe: fails to start, attach fails, or session is unusable (input/display).
4. Compare with the same steps on a working Windows build and on macOS/Linux.

Please capture when reporting:
- Windows version / build (`winver` or `Get-ComputerInfo`)
- Orca version
- Terminal runtime: native Windows shell vs WSL distro
- Multiplexer and version (e.g. tmux 3.x)
- Exact error or broken behavior

## Scope notes

- Treat as a **Windows terminal compatibility** issue for the multiplexer workflow.
- Prefer fixes that work across common Windows versions rather than one-off version checks unless a known OS bug requires it.
- Good first contribution shape: isolate repro matrix (Win version × shell × WSL), document which combinations fail, then fix the narrowest PTY/shell path.

## Labels

`help wanted` · `good first issue` — reproduction matrix + targeted PTY/shell fix are good entry points.

harshjainnn:
## Summary
Adds detection for Windows builds where node-pty silently falls back from ConPTY to WinPTY (build < 18309, e.g. Windows Server 2019 / Windows 10 LTSC 2019), and surfaces a one-time, non-fatal toast warning instead of letting full-screen TUI apps (tmux, vim, htop) render broken with no explanation. Fixes #9299.

## Screenshots
No visual change under normal conditions (ConPTY available). On affected Windows builds, a toast now appears once per session:

> **Terminal compatibility warning**
> This Windows version doesn't support ConPTY — full-screen terminal apps (tmux, vim, htop) may not render correctly.

(Not able to capture a live screenshot from this environment — no Windows build < 18309 available to trigger it; verified via unit/integration tests instead.)

## Testing
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [ ] `pnpm build`

New/updated tests:
- `windows-pty-backend.test.ts` — build-number boundary cases (>=18309, <18309, unparseable release string), non-win32 platforms
- `local-pty-provider.test.ts`, `pty-subprocess.test.ts`, `preflight-handler.test.ts` — warning attached/omitted correctly through spawn and preflight paths
- `windows-terminal-capabilities.test.ts` — toast fires once per owner key, does not fire on non-win32 hosts or when ConPTY is available

`pnpm build` not run in this environment (no Electron build target available here) — flagging for CI to confirm before merge.

## AI Review Report
Reviewed with an AI coding agent focused on: (1) correctness of the ConPTY/WinPTY detection threshold against node-pty's own source, (2) whether the warning is non-fatal and doesn't block spawn on any platform, (3) consistency of the `warning` field threading through the existing spread-based result patterns already used in this codebase (`daemon-pty-adapter.ts`, `terminal-host.ts`, etc.), (4) test coverage at each layer the field passes through.

Findings and what was changed/verified:
- **Duplicate warning trigger (open follow-up, not yet fixed):** two independent sites can fire the same toast — spawn-time in `pty-connection.ts` (gated by a single module-level boolean) and capability-load-time in `windows-terminal-capabilities.ts` (gated by a per-owner-key `Set`). Flagged as a known follow-up rather than blocking this PR; the per-owner-key site is the more robust of the two and is the one intended to remain long-term.
- **Stray `let` statement between imports** in `pty-connection.ts` (`let conptyWarningShown = false` sitting mid-import-block) — flagged for cleanup, should be moved below the import block.
- **Detection threshold correctness:** confirmed `isConptyAvailable()`'s `>= 18309` check exactly mirrors node-pty's internal `WindowsPtyAgent` logic (`this._useConpty = this._getWindowsBuildNumber() >= 18309`), so the warning fires precisely when node-pty would actually be using WinPTY — not an approximation.

**Cross-platform compatibility check:** explicitly reviewed for macOS, Linux, and Windows.
- `isConptyAvailable()` returns `false` immediately on any non-`win32` platform (covered by test), so macOS/Linux code paths are unaffected — no new branches execute for them.
- All new spawn-path changes (`local-pty-provider.ts`, `pty-subprocess.ts`) gate the warning behind `process.platform === 'win32'`, consistent with existing platform-branch conventions already used throughout those files (e.g. `windowsConptyDllOptions()`).
- No shortcuts, keybindings, or path-handling logic were touched.
- The renderer-side toast and IPC/preload plumbing are additive fields (`warning?`, `conptyAvailable?`) on existing result types, so non-Windows and non-Electron (web) consumers degrade safely to `undefined`/`false` rather than breaking.
- One residual gap: `web-preload-api.ts`'s `isConptyAvailable` stub is hardcoded to `Promise.resolve(false)` regardless of actual remote host state — flagged for verification that this matches how sibling capabilities (`wsl`, `pwsh`, `gitBash`) are stubbed in that same file, since if it doesn't, remote Windows web clients could see a false-positive warning.

## Security Audit
- **Input handling:** the only new input is `os.release()` output, parsed with `parseInt` and defaulting to `0` on failure (verified by the "unparseable release string" test) — no unvalidated external input reaches this path.
- **Command execution:** no new shell/process execution introduced. `isConptyAvailable()` performs no spawning; it only reads `os.release()`.
- **Path handling:** not applicable — no filesystem paths introduced or modified in this change.
- **IPC:** one new IPC handler, `preflight:isConptyAvailable`, registered following the exact pattern of existing preflight handlers (`preflight:detectRemoteWindowsTerminalCapabilities`) — takes no arguments, returns a boolean, no user-controlled data crosses the boundary.
- **Auth/secrets/dependencies:** no changes to authentication, secrets, or third-party dependencies. No new packages added.
- **Follow-up:** none identified beyond the cross-platform stub gap noted above (`web-preload-api.ts`).

## Notes
- This PR is detection-and-messaging only — it does not and cannot force ConPTY on Windows builds where it doesn't exist (pre-1809/18309); that's an OS limitation, not something Orca can work around. Making WinPTY itself render full-screen TUI apps correctly is out of scope.
- Known follow-up: consolidate the two toast-trigger sites (`pty-connection.ts` spawn-time vs. `windows-terminal-capabilities.ts` capability-load-time) to avoid a possible duplicate toast in one session.
- Known follow-up: confirm `web-preload-api.ts`'s hardcoded `false` for `isConptyAvailable` is intentional/consistent with how other capability stubs behave for web/remote clients, or wire it to the real remote check.
- Primarily affects Windows Server 2019 and Windows 10 LTSC 2019 (build 17763) environments, and any other Windows install below build 18309 — most current Windows 10/11 consumer installs are unaffected and will see no behavior change.

## ELI5

On older Windows builds without ConPTY, terminals fell back silently and fullscreen apps looked broken. Orca detects that fallback and shows a one-time toast explaining the compatibility limit.

### DIFF
--- a/src/main/daemon/daemon-pty-adapter.ts
+++ b/src/main/daemon/daemon-pty-adapter.ts
@@ -305,7 +305,8 @@ export class DaemonPtyAdapter implements IPtyProvider {
         pid,
         ...launchIdentity(),
         coldRestore: cachedRestore,
-        ...(!result.isNew ? { isReattach: true } : {})
+        ...(!result.isNew ? { isReattach: true } : {}),
+        ...(result.warning ? { warning: result.warning } : {})
       }
     }
 
@@ -372,14 +373,16 @@ export class DaemonPtyAdapter implements IPtyProvider {
           ...launchIdentity(),
           coldRestore,
           ...(providerSequence ? { providerSequence } : {}),
-          ...(!result.isNew ? { isReattach: true } : {})
+          ...(!result.isNew ? { isReattach: true } : {}),
+          ...(result.warning ? { warning: result.warning } : {})
         }
       }
       return {
         id: sessionId,
         pid,
         ...launchIdentity(),
-        ...(providerSequence ? { providerSequence } : {})
+        ...(providerSequence ? { providerSequence } : {}),
+        ...(result.warning ? { warning: result.warning } : {})
       }
     }
 
@@ -419,7 +422,8 @@ export class DaemonPtyAdapter implements IPtyProvider {
         pid,
         ...launchIdentity(),
         ...(providerSequence ? { providerSequence } : {}),
-        ...(isReattach ? { isReattach: true } : {})
+        ...(isReattach ? { isReattach: true } : {}),
+        ...(result.warning ? { warning: result.warning } : {})
       }
     }
 
@@ -451,7 +455,8 @@ export class DaemonPtyAdapter implements IPtyProvider {
       // split escape's continuation literally, unlike the remote path (#7329).
       ...(result.snapshot.pendingEscapeTailAnsi
         ? { pendingEscapeTailAnsi: result.snapshot.pendingEscapeTailAnsi }
-        : {})
+        : {}),
+      ...(result.warning ? { warning: result.warning } : {})
     }
   }
 
--- a/src/main/daemon/daemon-server.ts
+++ b/src/main/daemon/daemon-server.ts
@@ -419,7 +419,8 @@ export class DaemonServer {
           pid: result.pid,
           shellState: result.shellState,
           ...(result.launchAgent ? { launchAgent: result.launchAgent } : {}),
-          ...(result.historySeeded !== undefined ? { historySeeded: result.historySeeded } : {})
+          ...(result.historySeeded !== undefined ? { historySeeded: result.historySeeded } : {}),
+          ...(result.warning ? { warning: result.warning } : {})
         }
       }
 
--- a/src/main/daemon/pty-subprocess.test.ts
+++ b/src/main/daemon/pty-subprocess.test.ts
@@ -10,10 +10,12 @@ const {
   isPwshAvailableMock,
   validateWorkingDirectoryMock,
   resolveUnixShellPathMock,
-  resolveAgentForegroundProcessMock
+  resolveAgentForegroundProcessMock,
+  isConptyAvailableMock
 } = vi.hoisted(() => ({
   spawnMock: vi.fn(),
   isPwshAvailableMock: vi.fn(),
+  isConptyAvailableMock: vi.fn(),
   resolveUnixShellPathMock: vi.fn((shellPath: string) => shellPath),
   resolveAgentForegroundProcessMock: vi.fn(),
   validateWorkingDirectoryMock: vi.fn((cwd: string) => {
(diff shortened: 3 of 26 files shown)


AmethystLiang:
thanks for the PR! will review it soon

PR #9360, opened 2026-07-18
feat: real Antigravity usage, zh-TW UI, and renderer diagnostics shim

RX5950XT:
## Summary

- Read real Antigravity quota from the Windows Credential Manager entry instead of mirroring Gemini usage, with separate Claude/Gemini 5-hour and weekly windows in the status bar.
- Add Traditional Chinese (`zh-TW`) as a first-class UI locale, including locale-aware font fallbacks and catalog parity.
- Rebase onto current `main` and harden the implementation for cancellation, partial API responses, authentication refresh, and stale-data handling.

## Implementation notes

- Credential and HTTP work is cancellable, runs independently from the other providers, and keeps the Windows credential lookup from delaying their results.
- Quota summary and model fallback requests run in parallel; HTTP 401 remains distinct so refresh is attempted, metadata-only quota entries are ignored, and omitted windows are not restored with a fresh age.
- The renderer-wide Node builtin shim and import rewriting were removed after confirming the existing browser-safe dependency resolution builds successfully.
- Traditional Chinese system locales include Taiwan, Hong Kong, Macau, and Hant-script tags; `<html lang>` updates only after the latest lazy catalog activation succeeds.

## Validation

- 163 focused rate-limit, status-bar, and locale tests passed.
- `pnpm run typecheck` passed.
- Changed-file oxlint and oxfmt checks passed.
- `pnpm run verify:localization-catalog` passed.
- `pnpm run check:max-lines-ratchet` passed.
- `pnpm run build:electron-vite` passed.

## Platform and security

Antigravity quota is currently Windows-only because it reads the desktop host Credential Manager entry. Credentials remain read-only and in memory; token refresh uses the optional OAuth client environment configuration and otherwise relies on the current access token stored by Antigravity.

Contributor: @RX5950XT

## ELI5

Antigravity usage is read from real Windows credential data instead of faking Gemini numbers, with separate time windows in the status bar. Traditional Chinese (zh-TW) is also a full UI language option.


brennanb2025:
Taking a look!

brennanb2025:
Thanks for the contribution — I rebased this onto current `main` and tightened the implementation before taking it over. Antigravity quota reads are now independent and cancellable, preserve 401s for refresh, tolerate one optional quota source failing, ignore metadata-only entries, and no longer carry omitted windows forward with a fresh age; I also removed the redundant renderer-wide Node shim after verifying the existing browser-safe path builds cleanly. Traditional Chinese remains first-class with current catalog parity and safer locale activation, and the focused 163 tests, typecheck, changed-file lint/format checks, max-lines gate, localization catalog check, and desktop build pass.

RX5950XT:
Rebased onto current `main` (about 2500 commits of drift, including merge conflicts).

What changed in this refresh:
- Antigravity usage still reads Windows Credential Manager (`gemini:antigravity`) instead of mirroring Gemini
- `zh-TW` catalog pruned to current English key parity (missing keys still fall back to English)
- Dropped `antigravity-usage-mirror` now that the independent fetch path exists
- Plugin language packs / current rate-limit service orchestration kept

Ready for another look when you have time.

nwparker:
Takeover verification at `c699fa70583ddd53fd7e9098d2bc8270c662b8a8`: I reproduced a failure in the quota fallback path, so I have not adopted this implementation.

**A failed optional model request discards a successful quota summary.** `syncQuota` in `antigravity-usage-fetcher.ts` awaits both requests with `Promise.all`. If `retrieveUserQuotaSummary` succeeds but `fetchAvailableModels` rejects (network failure, timeout, or JSON parse error), the outer catch returns an error with no buckets. The successful summary never reaches `buildAntigravityRateLimitsFromQuota`.

I compiled the exact candidate modules and mocked only Electron networking, passing synthetic credentials directly. A summary containing one Gemini five-hour bucket at 75% remaining produced `ok`, 25% used when the model lookup returned `{models:{}}`. Keeping that same successful summary and making only the model lookup throw produced `error`, `session: null`, and zero buckets. No real credentials, Credential Manager calls, or external requests were used.

ELI5: Orca gets the usage numbers, but throws them away because a second, optional request failed. Settle the requests independently while preserving caller cancellation and explicit authentication failures; use the valid summary even when the model fallback is unavailable. Add this rejected-request case alongside HTTP-status cases.

Separately, the Windows credential reader adds direct `node:child_process`, `-ExecutionPolicy Bypass`, and per-read `Add-Type` compilation. Those need reconciliation with the current child-process boundary and `docs/reference/windows-edr-posture.md` before adoption. This is a source review finding, not a claim that I reproduced a Defender alert.

This review does not verify Windows credential lookup, live OAuth quota, or account switching. The independent native `agy /usage` takeover is tracked in #21654; its live populated OAuth case also remains unverified.

Issue #8145, opened 2026-07-10
[Feature]: Show OpenCode Zen balance alongside OpenCode Go usage


euclodius:
### Operating system

All / cross-platform

### Orca version

Current public builds

### Details

Feature request: please consider showing OpenCode Zen balance / credit state alongside the existing OpenCode Go usage and rate-limit information.

Why this would help:
- Orca's status-bar usage tracking is already useful for deciding whether it is safe to start or queue more agent work.
- OpenCode Go and OpenCode Zen can be part of the same practical workflow: when Go limits are exhausted, usage may fall back to Zen credits if that option is enabled.
- In that setup, seeing only Go limits is not the full cost/risk picture. Users also need to know whether Zen balance is available and whether a queued agent might start consuming paid credits.

Desired behavior:
- Keep OpenCode Go quota/limit information visible as it is today.
- Add Zen balance / credit information in the same status-bar popup or account-switcher details.
- Clearly distinguish quota-limited Go usage from paid/credit-backed Zen balance.
- If Zen data cannot be fetched, show a clear unavailable/refresh-failed state rather than silently omitting it.

This is not a blocker, but it would make the OpenCode status surface much more complete for users who rely on both Go limits and Zen fallback credits.

euclodius:
Cross-linking #9362 and PR #9363, which appear to implement this request directly.

This functionality is genuinely important for workflows that use OpenCode Go with Zen fallback, and it is one of the provider-usage improvements we are most looking forward to. Seeing the Zen balance before the Go quota is exhausted would make it much easier to avoid unexpected pay-as-you-go spending and decide safely whether to start or queue more agent work.

Please keep this request linked through merge and release. We would also be happy to test the implementation on Windows with WSL once it becomes available.

PR #9363, opened 2026-07-18
feat(usage): surface pay-as-you-go credit balances (Claude usage credits, OpenCode Go Zen balance, Codex credits)

### LINKED ISSUE #9362: [Feature]: Surface pay-as-you-go usage-credit balances (Claude usage credits, OpenCode Go Zen balance) in usage summaries
### Problem

Orca's provider usage summaries (the status-bar popover and compact segment) show plan windows — Session, Weekly, Monthly, Fable — but nothing about the **pay-as-you-go / overage balance** an account spends into once those windows are exhausted. Users who rely on this to keep working past a plan limit have no visibility into it from Orca and have to check each provider's own dashboard.

Reference behaviour: CodexBar surfaces an "Extra usage" / credit-balance section per provider.

### Proposed solution

Add an extra-usage balance to the usage summaries for the providers where the data is already reachable:

- **Claude** — surface the account's **usage credits** (the monthly spend limit + current balance) that Claude Code's own `/usage` panel shows. This rides along in the OAuth usage response Orca already fetches (`spend` object, with the legacy `extra_usage` shape as a fallback), so it needs no new network call or credential.
- **OpenCode Go** — surface the **Zen balance** (pay-as-you-go credit) parsed from the workspace page Orca already scrapes.

Behaviour:

- **Popover:** always show a balance row so it's discoverable even at 0% used. Claude renders a spent/limit meter plus current balance; OpenCode Go renders a plain available amount. The row reuses the existing panel tokens (label weight, bar colours, muted/faint text).
- **Compact status-bar segment:** stay hidden normally, and reveal the remaining balance (counting down) only once a plan window caps and the balance is actually spendable — so it appears exactly when the plan starts spending into it.
- **Currency:** taken from the provider response and formatted with `Intl.NumberFormat`, so it adapts to whatever currency and locale the account uses (e.g. `€2,000.00`, `$12.40`, `¥2000`).

### Out of scope (follow-up)

- **Codex / OpenAI credit balance** — a separate billing system (`api.openai.com/.../credit_grants`) that needs a new OpenAI API/Admin key + settings UI. Worth its own PR.
- **Claude API Console org credits** — the developer-platform balance on platform.claude.com is a different billing system from subscription usage credits and would need separate console auth.

mmarabel:
## Summary

Adds a **pay-as-you-go credit balance** to the provider usage summaries for the three providers that expose one in data Orca already fetches — no new network call and no new credential:

- **Claude → "Usage credits":** parses the `spend` object from the OAuth usage response Orca already fetches (monthly spend limit + current balance), with the legacy `extra_usage` shape as a fallback. This is the same data Claude Code's own `/usage` panel shows, including the disabled / "out of credits" state (rendered honestly, not mislabeled "off").
- **OpenCode Go → "Zen balance":** parses the pay-as-you-go Zen balance from the workspace page Orca already scrapes.
- **Codex → "Credits":** parses `rateLimits.credits` from the same app-server RPC Orca already uses for Codex windows. Codex credits are a **unitless count** (e.g. 500 credits), not a currency, so the shared model gained a `unit: 'currency' | 'credits'` discriminator plus an `unlimited` flag. Hidden for accounts that never bought credits (`hasCredits: false`).

Behaviour:

- **Popover** always shows the balance row so it's discoverable at 0% used. Claude renders a spent/limit meter plus current balance; OpenCode Go renders a plain available amount; Codex renders "N credits available" (or "Unlimited"). The row reuses the existing panel tokens (label weight, bar colours, muted/faint text) so it matches Session/Weekly exactly.
- **Compact status-bar segment** stays hidden normally and reveals the remaining balance (counting down) only once a plan window caps **and** the balance is actually spendable — so it appears exactly when the plan starts spending into it.
- **Denomination** is taken from the provider response: currency amounts are formatted with `Intl.NumberFormat` so they adapt to the account's currency and the viewer's locale (`€2,000.00`, `$12.40`, `¥2000`, `2.000,00 €`, …); Codex renders a plain localized credit count. Nothing is hardcoded to `$`/EUR.

Fixes #9362

## Screenshots

_Compact status-bar segment, Claude "Usage credits" popover, and the row in context:_

Compact status-bar segment
Claude Usage credits row in the popover
Usage credits row alongside Session and Weekly

_Codex "Credits" row:_

Codex Credits row showing 500 credits available

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Notes on `pnpm test`: the touched areas pass in full (Claude `spend` mapping incl. disabled/out-of-credits and the legacy fallback, the OpenCode Go Zen-balance parser, the currency formatter, popover rendering for enabled/disabled/uncapped, and the compact reveal incl. "disabled stays hidden"). The wider suite has a handful of failures in unrelated modules (PR-comment list, Linear prompt toast, sidebar toolbar, a flaky terminal-PTY test); these were confirmed pre-existing by re-running them with this branch's changes stashed.

New/updated tests:
- `src/shared/currency-format.test.ts` — locale-safe currency formatting + malformed-code fallback.
- `src/main/rate-limits/opencode-go-page-scraper.test.ts` — Zen-balance parsing (scaled billing object, explicit USD keys, human-readable text, disabled/no-balance cases, size guard).
- `src/main/rate-limits/claude-fetcher.test.ts` — `spend` → capped balance, disabled out-of-credits stays visible, legacy `extra_usage` fallback.
- `src/main/rate-limits/codex-fetcher.test.ts` — `rateLimits.credits` → credit-count balance; hidden when `hasCredits: false`.
- `src/renderer/src/components/status-bar/tooltip.test.ts` — popover render for capped/disabled/uncapped + credit-count + unlimited + per-provider labels.
- `src/renderer/src/components/status-bar/provider-segment-balance.test.tsx` — compact reveal on cap, hidden with headroom, hidden when disabled, credit-count vs currency.

## AI Review Report

Reviewed with the AI agent that wrote the change. Focus areas and outcomes:

- **Cross-platform (macOS / Linux / Windows):** the change is pure TypeScript/React with no platform branches, keyboard shortcuts, shortcut labels, path handling, or shell execution. Currency rendering uses `Intl.NumberFormat`, which is available and locale-aware across all Electron targets. No `metaKey`/path/`/`-vs-`\` assumptions were added.
- **SSH / remote:** the fetchers run in the main process via Electron `net.fetch` (Claude) and the existing OpenCode page scrape — same execution model as the current usage code, with no new local-only assumptions. The new field flows through the existing `ProviderRateLimits` IPC payload as a plain serializable object.
- **Git provider compatibility:** N/A — no source-control/review code touched.
- **Data-model safety:** `extraUsage` is additive and optional on `ProviderRateLimits`; existing providers that don't set it render unchanged (verified by an "omits the balance row" test).
- **Parsing robustness:** the Zen-balance scraper reuses the existing size guard (10 MB) and uses bounded regex quantifiers; the `customerID` gate prevents billing-disabled payloads (`customerID:null`, `balance:0`) from registering as a live balance.
- **Flagged & addressed during review:** initial version matched CodexBar and hid the row unless `is_enabled`, which hid a configured-but-disabled cap; reworked to show the real state after inspecting the live response. Switched from the minimal `extra_usage` field to the richer `spend` object (with fallback) so amounts, currency, and enabled state are exact. Removed temporary debug logging before commit.

## Security Audit

- **Untrusted input:** both parsers consume provider-controlled responses (OAuth JSON, scraped HTML). No `eval`/dynamic code; regexes are size-guarded and bounded. The currency code from the API is passed to `Intl.NumberFormat`, which throws on malformed codes — caught and replaced with a plain `<CODE> <amount>` fallback (tested).
- **Secrets:** no secrets are logged or persisted; the temporary debug logging used during investigation was removed. No credential handling changed — the feature reuses the existing OAuth token and OpenCode session cookie already in use.
- **Command execution / paths / IPC:** none added. No new IPC channels; the balance rides the existing usage payload.
- **Follow-up:** none required for this PR.

## Notes

- **Claude API Console org credits** (platform.claude.com) are a different billing system from subscription usage credits and are not covered here.
- **Codex credit unit:** the app-server reports `balance` as a plain credit count (verified against a live account: a $20 / 500-credit purchase surfaced as `balance: "500"`), so it renders as "500 credits", not a dollar amount.
- The compact balance token reveals only when a plan window is at ~100% and the balance is spendable, so it is inherently hard to exercise without hitting a real limit; the popover row is the always-visible surface.

https://claude.ai/code/session_01Y891yaV8gJpLycZEX6G5gm

## ELI5

Usage summaries can show pay-as-you-go credit balances for Claude, OpenCode Go Zen, and Codex when that data is already in the API response. You see remaining credits next to rate limits without extra logins.

### DIFF
--- a/src/main/rate-limits/claude-fetcher-extra-usage.test.ts
+++ b/src/main/rate-limits/claude-fetcher-extra-usage.test.ts
@@ -1,0 +1,194 @@
+import { beforeEach, afterEach, describe, expect, it, vi } from 'vitest'
+import { fetchClaudeRateLimits } from './claude-fetcher'
+import { primeClaudeFetcherMocks, restorePlatform } from './claude-fetcher-test-harness'
+import { readActiveClaudeKeychainCredentialsStrict } from '../claude-accounts/keychain'
+import type { ClaudeRuntimeAuthPreparation } from '../claude-accounts/runtime-auth-service'
+
+const { netFetchMock, readFileMock, resolveProxyMock, setProxyMock, appGetPathMock } = vi.hoisted(
+  () => ({
+    netFetchMock: vi.fn(),
+    readFileMock: vi.fn(),
+    resolveProxyMock: vi.fn(),
+    setProxyMock: vi.fn(),
+    appGetPathMock: vi.fn()
+  })
+)
+
+vi.mock('node:fs/promises', () => ({
+  readFile: readFileMock
+}))
+
+vi.mock('electron', () => ({
+  app: {
+    getPath: appGetPathMock
+  },
+  net: {
+    fetch: netFetchMock
+  },
+  session: {
+    defaultSession: {
+      resolveProxy: resolveProxyMock,
+      setProxy: setProxyMock
+    }
+  }
+}))
+
+vi.mock('./claude-pty', () => ({
+  fetchViaPty: vi.fn()
+}))
+
+vi.mock('../claude-accounts/keychain', () => ({
+  deleteActiveClaudeKeychainCredentialsStrict: vi.fn(),
+  readActiveClaudeKeychainCredentials: vi.fn(),
+  readActiveClaudeKeychainCredentialsStrict: vi.fn(),
+  readManagedClaudeKeychainCredentials: vi.fn(),
+  writeActiveClaudeKeychainCredentials: vi.fn(),
+  writeManagedClaudeKeychainCredentials: vi.fn()
+}))
+
+function oauthPrep(): ClaudeRuntimeAuthPreparation {
+  return {
+    configDir: '/Users/test/.claude',
+    envPatch: { CLAUDE_CONFIG_DIR: '/Users/test/.claude' },
+    stripAuthEnv: false,
+    provenance: 'system'
+  }
+}
+
+describe('fetchClaudeRateLimits extra usage', () => {
+  beforeEach(() => {
+    primeClaudeFetcherMocks({
+      netFetchMock,
+      readFileMock,
+      resolveProxyMock,
+      setProxyMock,
+      appGetPathMock
+    })
+  })
+
+  afterEach(() => {
+    restorePlatform()
+  })
+
+  it('maps the usage-credits spend object into a capped balance in major units', async () => {
+    vi.mocked(readActiveClaudeKeychainCredentialsStrict).mockResolvedValueOnce(
+      JSON.stringify({
+        claudeAiOauth: {
+          accessToken: 'oauth-token',
+          expiresAt: Date.now() + 60_000
+        }
+      })
+    )
+    netFetchMock.mockResolvedValueOnce(
+      new Response(
+        JSON.stringify({
+          five_hour: { utilization: 10 },
+          spend: {
+            used: { amount_minor: 5000, currency: 'EUR', exponent: 2 },
+            limit: { amount_minor: 200000, currency: 'EUR', exponent: 2 },
+            percent: 2.5,
+            enabled: true,
+            balance: { amount_minor: 1000, currency: 'EUR', exponent: 2 }
+          }
+        }),
+        { status: 200 }
+      )
+    )
+
+    await expect(fetchClaudeRateLimits({ authPreparation: oauthPrep() })).resolves.toMatchObject({
+      provider: 'claude',
+      status: 'ok',
+      extraUsage: {
+        balance: 10,
+        spent: 50,
+        spendLimit: 2000,
+        spentPercent: 2.5,
+        currencyCode: 'EUR',
+        enabled: true,
+        disabledReason: null,
+        resetsAt: null
+      }
+    })
+  })
+
+  it('keeps the usage-credits cap visible but disabled when out of credits', async () => {
+    vi.mocked(readActiveClaudeKeychainCredentialsStrict).mockResolvedValueOnce(
+      JSON.stringify({
+        claudeAiOauth: {
+          accessToken: 'oauth-token',
+          expiresAt: Date.now() + 60_000
+        }
+      })
+    )
+    netFetchMock.mockResolvedValueOnce(
+      new Response(
+        JSON.stringify({
+          five_hour: { utilization: 10 },
+          spend: {
+            used: { amount_minor: 0, currency: 'EUR', exponent: 2 },
+            limit: { amount_minor: 200000, currency: 'EUR', exponent: 2 },
+            percent: 0,
+            enabled: false,
+            disabled_reason: 'out_of_credits',
+            balance: null
+          }
+        }),
+        { status: 200 }
+      )
+    )
+
+    await expect(fetchClaudeRateLimits({ authPreparation: oauthPrep() })).resolves.toMatchObject({
+      provider: 'claude',
+      status: 'ok',
+      extraUsage: {
+        balance: 0,
+        spent: 0,
+        spendLimit: 2000,
+        spentPercent: 0,
+        currencyCode: 'EUR',
+        enabled: false,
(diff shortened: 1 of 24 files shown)


mmarabel:
Thanks @coderabbitai — addressed the two substantive items in 9d0333a:

- **Nested `PanelBalanceSection` → plain render function.** Converted it to `renderBalanceSection(...)` called directly, so React keeps the DOM node and the progress bar's `transition-all` width animation survives re-renders. (Left the pre-existing `PanelWindowSection` as-is to keep this PR single-topic — it has the same pattern and is worth a separate cleanup.)
- **Compact `bal` label localized.** Set proper translations instead of the English placeholder: `残高` (ja), `余额` (zh), `잔액` (ko), `saldo` (es). English stays `bal`.

Two I'm intentionally leaving:

- **`ProviderPanel({ p })` direct call in `tooltip.test.ts`.** This matches the file's existing convention — every `ProviderPanel` render test in that file calls it directly, and the component is hookless — so I kept it consistent rather than introducing a mixed `createElement`/direct-call style in one file. Happy to convert the whole file in a follow-up if preferred.
- **Docstring coverage (50% < 80%).** The repo's `AGENTS.md` explicitly asks for short "why"-only comments and discourages restating what code does, so I've kept comments to the non-obvious rationale per that convention rather than padding to hit the threshold.

The "resolve merge conflict" pre-merge check predates the `main` merge I pushed earlier — the branch is now `MERGEABLE`.

brennanb2025:
Taking a look!

brennanb2025:
Thanks for the contribution — I rebased this onto current `main` and tightened the implementation while applying the remaining review feedback. The balance scraper now uses bounded, billing-record-scoped parsing instead of broad page-wide matches, preserves granted credit without requiring a payment customer, and only exposes the compact balance when overage spending is actually enabled; currency and credit balances now have distinct typed shapes, credit counts are localized, and malformed money data is handled defensively. The 161 focused usage/UI tests, full typecheck, touched-file lint and format checks, localization catalog verification, max-lines guard, and changed-files React checks all pass.

euclodius:
Thanks for the work on this — this PR appears to implement the OpenCode Zen balance requested in #8145 directly. I noticed the branch is currently conflicting with `main`. Is the rebase still planned? I would be happy to help validate the refreshed implementation on Windows with WSL once it is ready.

One related scope question while this PR is being revived: could the new `ExtraUsageBalance` model remain provider-agnostic so it can support Grok and Cline credit balances in follow-ups?

- **Grok:** Orca's existing billing payload already exposes `onDemandCap`, `onDemandUsed`, and `prepaidBalance`; the current fetcher maps only the subscription windows. This looks like a natural follow-up once the shared balance model lands.
- **Cline:** ClinePass subscription quota and Cline's pay-as-you-go credit balance are separate billing surfaces. PR #13827 covers the ClinePass 5h/weekly/monthly quota, while Cline credits would require confirming the balance endpoint and whether the same API key can access it.

I would not want either addition to delay #9363 — landing the generic balance model and OpenCode Zen support first is the priority. It would still be useful to keep these providers explicitly in scope for follow-up work.

Please keep #8145 linked through merge/release as well.

mmarabel:
Rebased onto current `upstream/main` (`5b7f44278a`).

Credit-balance parsing/UI was re-applied onto the current rate-limit modules:
- Claude extra-usage tests now live in `claude-fetcher-extra-usage.test.ts` (the old `claude-fetcher.test.ts` was split on main).
- Codex credits parse through `CodexRateLimitWindowsSnapshot.credits` rather than the deleted `RpcRateLimitsResult` type.
- Compact status-bar and tooltip keep main's current window layout and add the overage-balance rows.

Force-pushed `feat/provider-usage-credit-balances`.

Issue #9300, opened 2026-07-18
Windows & WSL: performance audit and improvement pass


AmethystLiang:
## Summary

Orca on Windows (native host and WSL-backed worktrees) often feels slower than macOS/Linux for everyday workflows: source control refresh, file watching, terminal spawn, search/list files, and worktree switching.

This issue tracks a **broad performance audit + improvement pass** for Windows and WSL—not a single bug. Related, more specific work may already exist (e.g. git status / WSL hot path: #9284); this is the umbrella for measuring, prioritizing, and fixing cross-cutting costs.

## Goals

1. **Measure** the main Windows/WSL latency sources with reproducible scenarios.
2. **Prioritize** fixes by user-visible impact (cold start, worktree switch, status poll, file watch, search, terminal open).
3. **Ship** incremental improvements with before/after numbers where practical.
4. Avoid regressions on macOS/Linux and on non-WSL Windows paths.

### Process / spawn
- `wsl.exe` hops for git, `rg`, and other tools (`src/main/git/runner.ts`, `wslAwareSpawn`)
- Login-shell vs non-login WSL invocations on hot paths
- Windows process spawn + antivirus real-time scanning cost
- Concurrent spawn churn on idle status / metadata polls

### Filesystem
- Native Windows paths vs `\\wsl.localhost\...` / `\\wsl$\...` (9P)
- WSL git on `/mnt/c/...` (DrvFs) when project runtime is WSL but repo is on a Windows drive
- Hybrid I/O: some work in WSL, some via Windows Node `fs`
- `@parcel/watcher` setup/teardown on Windows (known expensive; AV interaction)
- WSL in-distro snapshot poller vs native watcher

### Source control / worktrees
- `git status` pipeline cost (porcelain + untracked + numstat + upstream probe)
- Worktree list / branch identity refresh cadence
- Huge unignored folders (`node_modules`, build outputs)

### Search & explorer
- `rg` routing (host vs WSL)
- Quick-open / list-files fallbacks when `rg` is missing in the distro
- Debounce / coalesce of FS change → UI refresh

### Terminals & agents
- PTY spawn latency (ConPTY / WSL shell)
- Env/PATH probes that re-enter WSL on every open
- Hook / auth file reads over UNC

### Startup & idle
- Startup work that blocks first paint or first worktree use on Windows
- Idle polling that keeps WSL distros awake or restarts after `wsl --shutdown`

## Suggested approach

1. **Repro matrix** (document per machine):
[...]

Issue #9284, opened 2026-07-18
Windows/WSL: git status hot path is over-expensive and path handling is inconsistent


AmethystLiang:
## Summary

On Windows, frontend git UI ops (especially source-control status) can feel very sluggish. Digging into the code shows this is not only antivirus / filesystem cost — Orca has a few **design and correctness issues** in the Windows ↔ WSL git path.

## How routing works today

Local git always goes through `src/main/git/runner.ts`:

| Repo location | Project Windows runtime | Git runs as |
|---|---|---|
| `C:\...` | windows-host (default) | Windows `git.exe` |
| `C:\...` | WSL | WSL git via `/mnt/c/...` |
| `\\wsl.localhost\...` / `\\wsl$\...` | anything | WSL git (path-forced) |
| SSH worktree | n/a | remote host |

Frontend → `getRuntimeGitStatus` → IPC `git:status` → `getStatus()` → runner.

### 1. Hot-path git uses a WSL **login shell** whenever `wslDistro` is set (perf)

In `gitExecFileAsync` / `gitSpawn` / buffer variants:

```ts
useWslLoginShell: Boolean(options.wslDistro)
```

When the project/global runtime is WSL, **every** git spawn becomes roughly:

`wsl.exe → sh -lc → user shell -ilc → git ...`

That loads `.bashrc` / nvm / etc. Status polling multiplies this (status + optional numstat + upstream probe).

By contrast, path-only WSL UNC routing (no `wslDistro` option) uses plain `bash -c`, which is cheaper.

**Login shell is reasonable for finding user-installed tools once; it is the wrong default for high-frequency status polls.**

### 2. WSL runtime + Windows-disk repo forces `/mnt/c` git (perf product choice)

`getLocalProjectGitExecOptions` passes `wslDistro` so agents/terminals/git share one runtime. For repos on `C:\`, that means WSL git reads the tree through DrvFs (`/mnt/c/...`), which is a known slow pairing.

Consistent runtime is intentional, but it makes SCM refresh one of the worst-case paths without any cheaper “git on the filesystem host” fallback.

### 3. `resolveGitDir` does not translate Linux gitdir paths (correctness)

`src/main/git/repo.ts` correctly maps absolute Linux gitdir paths under WSL UNC:

```ts
if (baseWsl && value.startsWith('/')) {
  return toWindowsWslPath(value, baseWsl.distro)
}
```

`src/main/git/status.ts` `resolveGitDir()` does only:

```ts
[...]

PR #9372, opened 2026-07-18
Fix WSL git status hot-path login-shell overhead

pandeysambhi:
## Summary

WSL-routed Git status polling previously started the user's interactive login shell for every Git subprocess. This PR keeps the login environment for network, mutating, and unclassified Git commands, while allowing an explicit set of local read-only commands to use a cached login-shell `PATH` with a non-login shell.

The safety measures are:

- Capture each distro's login-shell `PATH` once before enabling the fast path, preserving Git installed through linuxbrew, nix, asdf/mise, or custom prefixes.
- Retry buffered read commands through the login shell only when the fast shell exits 127 with a command-not-found error.
- Keep network, mutating, and unknown commands on the login shell so SSH agents, signing tools, hooks, credential helpers, and other profile policy remain available.
- Keep locale assignments in the WSL command string so Git output remains stable for parsers.

Refs #9284
Related #9300

## ELI5

Orca was opening the equivalent of a fully configured terminal before every background Git status check. It now remembers where Git is after one full startup, then uses a lightweight shell for safe read-only checks. Commands that upload, download, commit, or might depend on credentials still use the original full environment.

## Proof of fix

The new regression test was copied into a detached `origin/main` worktree and run against main's `runner.ts`:

```text
$ vitest run src/main/git/runner-wsl-shell-selection.test.ts

FAIL  src/main/git/runner-wsl-shell-selection.test.ts
  × uses the fast non-login shell once the login PATH is known
    expected [ 'sh', '-lc' ] to deeply equal [ 'bash', '-c' ]
  × replays the login-shell PATH so profile-installed git still resolves
    expected login-shell script to contain PATH='/home/u/.linuxbrew/bin:/usr/bin:/bin'
  × finds read subcommands past leading global options
    expected [ 'sh', '-lc' ] to deeply equal [ 'bash', '-c' ]
  × falls back to the login shell when the fast shell cannot find git
    Error: exit 127
  × applies the same shell selection to the streamed status hot path
    expected [ 'sh', '-lc' ] to deeply equal [ 'bash', '-c' ]

Test Files  1 failed (1)
Tests       5 failed | 6 passed (11)
```

The same test at PR head:

```text
Test Files  1 passed (1)
Tests       11 passed (11)
```

### Timing

The available Windows 11 host has the WSL executable but no WSL installation or registered distro, so a live WSL benchmark was not possible. The login-vs-non-login delta was measured locally on macOS using the real repository and the same `git status --porcelain=v2 --branch --untracked-files=all` command, with 5 warmups and 50 measured iterations per shell:

```text
login-interactive: N=50 avg=939.53ms median=937.97ms p95=1007.83ms min=883.83ms max=1079.09ms
non-login:        N=50 avg= 42.37ms median= 42.29ms p95=  44.11ms min= 39.39ms max=  44.92ms
```

Indicative savings on this developer profile: **897.16 ms/call average (95.5%)**. This is explicitly not a WSL measurement; the absolute WSL result depends on the user's profile. The implementation pays the login startup once per distro, then removes it from steady-state status reads.

## Proof of no regressions

```text
$ npx vitest run --config config/vitest.config.ts src/main/git/*.test.ts
Test Files  36 passed (36)
Tests       530 passed (530)

$ npx tsc --noEmit -p config/tsconfig.node.json
(exit 0)

$ npx oxlint src/main/git/runner.ts src/main/git/git-read-only-subcommand.ts \
    src/main/git/wsl-login-shell-path.ts \
    src/main/git/runner-wsl-shell-selection.test.ts \
    src/main/git/runner-command-exec.test.ts
(no findings)

$ git diff --check
(no findings)
```

User-visible behavior changes only for WSL-routed Git:

- After one per-distro PATH probe, allowlisted local reads use `bash -c` rather than the user's interactive login shell.
- Network, mutating, and unclassified commands retain login-shell behavior.
- If buffered fast-path Git is still not found, Orca retries once using the login shell.

macOS, Linux, native Windows Git, SSH providers, and relay execution do not enter this WSL routing branch.

## Compatibility

- **Platforms:** WSL-only behavior change. Native Windows, macOS, and Linux routing is unchanged. A real Windows 11 host confirmed WSL is not installed there, so Windows/WSL execution is unit-tested rather than live-tested. No new Git subcommand or flag was introduced; `docs/reference/git-compatibility.md` was reviewed and the Git 2.25 baseline is unchanged.
- **SSH / remote:** Network commands remain on the login shell. This preserves profile-provided `SSH_AUTH_SOCK`, `GIT_SSH_COMMAND`, credential helpers, and SSH configuration. Commit/config/checkout and every unclassified or mutating command also remain on the login shell, preserving signing tools and hooks. `HOME` still comes from WSL itself; the fast path does not replace it. SSH-host and relay Git paths are untouched.
- **Performance:** One login-shell PATH probe per WSL distro, followed by non-login execution for allowlisted local reads. Indicative local status benchmark: 939.53 ms average login vs 42.37 ms non-login across 50 iterations, a 95.5% reduction in shell-plus-status latency on this profile.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — full command not run; changed files passed `npx oxlint`.
- [x] `pnpm typecheck` — equivalent scoped command `npx tsc --noEmit -p config/tsconfig.node.json` passed.
- [ ] `pnpm test` — full suite not run; all 36 test files in `src/main/git/` passed (530 tests).
- [ ] `pnpm build` — not run.

## AI Review Report

The review focused on shell/environment correctness and the actual streamed status path:

- Found the original patch changed every ordinary WSL `gitExecFileAsync` call, not just status, and could lose a profile-added `PATH` entirely.
- Added per-distro login `PATH` capture plus narrow command-not-found fallback.
- Verified the streamed status path uses `gitSpawn`; the fast-shell selection now covers that path too.
- Replaced a network-command denylist with an explicit local read-only allowlist. Commit, checkout, config writes, network operations, and unknown commands therefore retain the full login environment.
- Checked `PATH`, `GIT_SSH_COMMAND`, `SSH_AUTH_SOCK`, `HOME`, and `LC_ALL`/`LANG`. PATH is captured and shell-quoted; SSH/signing-dependent operations stay on login; HOME remains WSL-provided; locale is explicitly pinned in the command string.
- Checked macOS, Linux, native Windows, WSL, SSH, folder-workspace, and Git 2.25 implications. Only the WSL routing branch changes, and command/flag compatibility is unchanged.

## Security Audit

- Each Git argument remains individually POSIX-shell quoted.
- The captured login-shell PATH is treated as data and emitted through `quotePosixShell`, preventing shell injection through PATH content.
- The exit-127 fallback is limited to a non-login WSL invocation whose stderr reports command-not-found; genuine Git failures are not masked.
- Credential-bearing network commands remain under login-shell policy.
- PATH state is keyed per distro, preventing environment crossover between WSL distributions.
- No dependencies, secrets storage, IPC contracts, or provider APIs changed.

## Notes

The dominant risk was PATH/environment loss. A bare non-login shell does not source `/etc/profile`, `~/.profile`, `~/.bashrc`, nvm/asdf/mise setup, or custom prefix configuration. The mitigation is prominent and intentional: Orca stays on the existing login shell until a one-time per-distro login PATH probe succeeds, replays that quoted PATH on allowlisted reads, and has a narrow buffered-command fallback to the login shell on exit 127.

Live WSL timing remains desirable before release. The available Windows host reported that Windows Subsystem for Linux is not installed, so the included number is an honestly labeled macOS indication of login-profile overhead, not a fabricated WSL result.

Original patch by @pandeysambhi.


brennanb2025:
Taking a look!

AmethystLiang:
taking over the review

nwparker:
Superseded — thank you for this, and sorry it sat.

WSL read-only git no longer goes through a login shell at all. `git/runner.ts:472` routes any read command (`isWslDirectGitReadCommand`) through `--exec /usr/bin/env PATH=<cached login PATH> HOME=<home> <gitPath>` with no shell in between, using a per-distro cached environment. That removes the overhead this PR set out to remove, and removes the rc/banner exposure with it.

Your diagnosis was right; it landed as part of the wider WSL runner work rather than as this diff. Closing as superseded.

PR #9373, opened 2026-07-18
fix(tab-bar): detect new-tab menu agents on the host that owns the workspace

### LINKED ISSUE #9370: [Bug]: New-tab agent menu lists locally installed agents for remote (runtime-hosted) workspaces
**Operating system:** macOS (client); remote workspace host affected regardless of OS

**Orca version:** v1.4.140-rc.0

**Details**

Short summary:
When a workspace is hosted on a remote server (runtime environment, no SSH connection), the tab-bar `+` menu's agent section (Claude, Codex, Grok, OpenCode, Pi, OMP, Droid, ...) shows the agents installed on the **local client machine** instead of the ones installed on the remote host.

What happened?
- Open a workspace that lives on a remote server where only some agents are installed.
- Open the `+` / new-tab dropdown: every agent installed locally is offered, including ones missing on the remote.
- Picking one of those agents opens a terminal that immediately fails, e.g. `zsh: command not found: pi`.
- On local workspaces the menu correctly only shows detected agents, so remote workspaces should respect the remote host's installed set the same way.

How can we reproduce it?
1. Connect to / open a workspace hosted on a remote runtime environment.
2. Make sure an agent (e.g. `pi`) is installed locally but NOT on the remote host.
3. Click the tab-bar `+` button — the missing agent is listed anyway.
4. Launch it — the terminal on the remote host fails with `command not found`.

Anything else that might help:
- Root cause: `QuickLaunchAgentMenuItems` (`src/renderer/src/components/tab-bar/QuickLaunchButton.tsx`) resolved the agent-detection host from the repo's SSH connection id only. Runtime-hosted workspaces have no SSH connection, so detection silently fell back to probing the local machine's PATH.
- The searchable create-entry in the same dropdown (`TabBar.tsx`) already resolved SSH → runtime → local correctly; the two menu variants had drifted.

mmarabel:
Closes #9370.

## Summary

For workspaces hosted on a remote server (runtime environment, no SSH connection), the tab-bar `+` menu's agent section resolved the agent-detection host from the repo's SSH connection id only. That came back `null` for runtime-hosted workspaces, so detection silently probed the **local** machine and offered agents that are not installed remotely — launching one failed immediately (e.g. `zsh: command not found: pi`).

The searchable create-entry in the same dropdown already resolved SSH → runtime → local correctly; the two menu variants had drifted. This PR extracts that resolution into a shared `useAgentDetectionTarget(worktreeId)` hook (`src/renderer/src/hooks/useAgentDetectionTarget.ts`) and uses it in both places, so the static menu, the searchable entry, and the review-notes send menu (which reuses the component) cannot drift again. `undefined` while the owning repo hydrates is preserved, so local agents never flash for remote workspaces.

## Screenshots

After the fix — a workspace on a remote server lists only the agents actually detected on the remote host (compare with the broken behavior in #9370, where locally-installed agents like Pi/OMP/Droid were offered and then failed with `command not found`):

New-tab agent menu on a remote workspace showing only remotely detected agents

No visual change to the menu UI itself — only which agents are listed changes.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 29,550 passed; 28 failures in 4 unrelated files (`pr-comments-list-selection`, `pr-comment-presentation`, `LinearAgentSkillSetupPrompt.reminder-toast`, `SidebarToolbar`) reproduce identically without this change — they fail on `window.localStorage` being undefined under Node 26 (repo engines want Node 24). None of them import anything touched here.
- [ ] `pnpm build` — not run; full `tsc` typecheck passes across node/cli/web configs and the dev app (`pnpm dev`) builds and launches cleanly with the fix verified manually against a remote workspace.

New tests: 7 resolver cases in `useAgentDetectionTarget.test.ts` (local / SSH / runtime repo owner / worktree-level override / focused-runtime fallback / SSH-over-runtime precedence / hydration-pending) and 4 regression tests in `QuickLaunchButton.test.ts` asserting the menu probes the runtime environment, the SSH host, local, and never falls back to local while the owning repo hydrates.

## AI Review Report

Review covered: (1) both tab-bar menu variants plus the `ReviewNotesSendMenuContent` reuse of `QuickLaunchAgentMenuItems` — all now share the one resolver; (2) host-ownership precedence matches `TabBar`'s pre-existing order (SSH connection id first, then runtime environment, then local, undefined while unhydrated); (3) the resolver returns a stable string key from the Zustand selector with the object memoized from the key, so retained tab strips don't re-render on unrelated store writes; (4) cross-platform compatibility — no platform-specific behavior introduced: detection itself continues to run through the existing per-host paths (local PATH probing incl. WSL targeting, SSH relay `preflight.detectAgents` on the remote host, runtime-daemon RPC), and no shortcuts, labels, paths, or shell behavior were touched. Flagged and addressed: the type-only `ConnectionOwnerState` export is erased at runtime, so main-process behavior is untouched.

## Security Audit

No new IPC surface, command execution, path handling, auth, or secrets. The change is pure renderer state reads (repo/worktree ownership) choosing among three **existing** detection channels: local preflight IPC, the authenticated SSH relay's `preflight.detectAgents`, and the runtime-daemon RPC allowlist (`preflight.detectAgents` was already allowlisted and used by the searchable create entry). No user input is parsed; no dependencies added. No follow-up needed.

## Notes

- First menu open on a remote workspace triggers one fresh probe against the remote host (cached afterward), so the list may briefly show "No agents detected" on first open — same behavior the searchable create entry already had.
- SSH-connected repos are behaviorally unchanged (same target as before); only runtime-hosted workspaces move from local to remote detection.

## ELI5

On remote-server workspaces, the new-tab menu listed agents installed on your laptop, not on the remote host — so launching one failed with command not found. Agent detection now uses the host that owns the workspace.

### DIFF
--- a/src/renderer/src/components/tab-bar/QuickLaunchButton.test.ts
+++ b/src/renderer/src/components/tab-bar/QuickLaunchButton.test.ts
@@ -3,26 +3,32 @@ import { renderToStaticMarkup } from 'react-dom/server'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { QuickLaunchAgentMenuItems, shouldShowLaunchWatchdogTimeout } from './QuickLaunchButton'
 
-const { shortcutLabelMock, storeState, openSettingsPageMock, openSettingsTargetMock } = vi.hoisted(
-  () => ({
-    shortcutLabelMock: vi.fn<() => string | null>(),
-    storeState: {
-      settings: {
-        defaultTuiAgent: 'codex' as 'claude' | 'codex' | 'gemini' | 'blank' | null,
-        disabledTuiAgents: [] as string[]
-      },
-      worktreesByRepo: {},
-      repos: [],
-      openSettingsPage: vi.fn(),
-      openSettingsTarget: vi.fn()
+const {
+  shortcutLabelMock,
+  storeState,
+  openSettingsPageMock,
+  openSettingsTargetMock,
+  useDetectedAgentsMock
+} = vi.hoisted(() => ({
+  shortcutLabelMock: vi.fn<() => string | null>(),
+  storeState: {
+    settings: {
+      defaultTuiAgent: 'codex' as 'claude' | 'codex' | 'gemini' | 'blank' | null,
+      disabledTuiAgents: [] as string[],
+      activeRuntimeEnvironmentId: null as string | null
     },
-    openSettingsPageMock: vi.fn(),
-    openSettingsTargetMock: vi.fn()
-  })
-)
+    worktreesByRepo: {} as Record<string, { id: string; repoId: string; hostId?: string | null }[]>,
+    repos: [] as { id: string; connectionId: string | null; executionHostId: string | null }[],
+    openSettingsPage: vi.fn(),
+    openSettingsTarget: vi.fn()
+  },
+  openSettingsPageMock: vi.fn(),
+  openSettingsTargetMock: vi.fn(),
+  useDetectedAgentsMock: vi.fn()
+}))
 
 vi.mock('@/hooks/useDetectedAgents', () => ({
-  useDetectedAgents: () => ({ detectedIds: ['claude', 'codex', 'gemini'] })
+  useDetectedAgents: useDetectedAgentsMock
 }))
 
 vi.mock('@/hooks/useShortcutLabel', () => ({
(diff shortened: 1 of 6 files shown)


brennanb2025:
Taking a look!

mmarabel:
Not rebased / not pushed: this is already on `main`.

`useAgentDetectionTargetForWorktree` now resolves SSH → runtime → local (plus folder / floating / hydration-pending cases) and is used by both `QuickLaunchButton` and `use-tab-bar-runtime-model`. Replaying this branch would regress that resolver.

mmarabel:
Closing as already shipped: `useAgentDetectionTargetForWorktree` already owns host-aware new-tab agent detection on main.

PR #9380, opened 2026-07-18
perf(renderer): stop agent spinners from keeping the frame pipeline awake (one shared clock, hidden-window stop)

nwparker:
> **Rescoped after #9395 merged** — originally this PR also carried a main→renderer window-visibility IPC relay to work around `setBackgroundThrottling(false)` pinning `document.visibilityState` on macOS. #9395 re-enabled throttling, which made `visibilitychange` work natively and the hidden-window burn stop on its own, so the relay (main/preload/api-types/web-stub changes) has been removed. **This PR is now renderer-only** and targets the case throttling can never help: the *visible* window.

## TL;DR

The yellow "agent is working" spinners (sidebar worktree dots, nested per-agent rows in worktree cards, terminal tabs, dashboard, AI Vault subagents) each ran their own infinite CSS animation. An active CSS animation keeps Chromium's frame/compositor pipeline permanently awake and pays per element — for the entire duration of every agent run, on a surface users keep visible for hours.

This PR keeps the spinner looking the same (same ring, same 30° step grain) but rotates every mounted one from **one shared 12-steps/s clock** (frame-for-frame identical cadence to the old animation) that writes `style.transform` directly and goes fully idle between ticks.

**Live interleaved A/B/A** (dev app post-#9395, renderer process, one working agent, 20s windows):

| Leg | CPU |
| --- | --- |
| Shared clock | 21.0 ms CPU/s |
| Old CSS animation (injected over same session) | 37.0 ms CPU/s |
| Shared clock again | 21.5 ms CPU/s |

~16 ms CPU/s saved with a *single* spinner — and the clock is flat in spinner count (one timer, N elements) where the CSS scheme scales per element, so parallel-agent sidebars save proportionally more. An earlier same-methodology run measured 23.5 → 10.0 ms/s; both runs agree on the shape: **roughly 40–60% of spinner-attributable renderer CPU, growing with agent count.**

## ELI5

Imagine a wall of clocks where every clock has its own battery motor running continuously. That was the old spinners: each one told the browser "keep the animation machinery running," and the browser woke up for every frame slot to service it.

The new version is one person walking past the wall six times a second, nudging every clock hand at once. Between nudges, everything sleeps. When the room empties (window hidden), the person stops — and with #9395 merged, the renderer now actually knows when the room is empty.

## Why CSS animation is the wrong tool for an always-on spinner

Counter-intuitive benchmark results (headed Chromium, 8 spinners + realistic filler DOM):

| Variant | CPU |
| --- | --- |
| No animation | ~6 ms/s |
| CSS `spin 1s steps(12)` (what we had) | ~24 ms/s |
| CSS `spin 3s steps(12)` — 3× slower | ~22 ms/s |
| CSS animation, `animation-play-state: paused` | ~6 ms/s |
| **JS clock 6Hz/30° (this PR)** | isolated ~28–34 ms/s, but **live: 21 vs 37 ms/s** |
| JS clock 12Hz (old look 1:1) | ~39 ms/s — worse than CSS |

Two lessons baked into code comments:
1. **Slowing a CSS animation barely helps** — the cost is the always-armed animation machinery, not pixels. Only *no active animation* returns the pipeline to idle.
2. **Don't "smooth" by raising Hz** — at 12Hz the JS clock costs more than the CSS animation it replaces. 6Hz with 30° steps keeps the old animation's step size (the grain that read as smooth) at half its wakeup rate.

## Design

- **`agent-spinner-clock.ts`** (~130 lines, renderer-only): module-level element registry + one `setInterval`. Register an element → it joins in phase with the shared dial (parallel agents tick in lockstep); unregister → last one out clears the timer and all global listeners. Zero cost when no agent is working.
- **Gates**: document visibility (works natively again post-#9395), `prefers-reduced-motion` (now respected by *every* spinner — the sidebar dot previously ignored it, and it live-reacts to OS setting changes), and the existing stale-visibility latch (`stale-document-visibility.ts`) so a wedged macOS occlusion tracker can't freeze spinners a user is provably looking at.
- **`AgentWorkingSpinner.tsx`**: the shared ring, attached via a React 19 cleanup-ref. Both spinner primitives (`StatusIndicator`, `AgentStateDot`) render it, which covers every working-state spinner in the app by construction — sidebar dots, nested worktree-card agent rows, terminal tabs, dashboard rows, AI Vault subagents. `rg 'animation:spin|animate-spin'` over agent-status paths: clean.
- Hidden-window interaction with #9395: `visibilitychange` stops the clock outright (measured 0.7 ms/s hidden). Chromium's 1s hidden-timer clamp is irrelevant since the timer is cleared, not clamped.

## Verification

- Live (dev app on rebased branch, CDP-driven): spinner steps 0°→30°→…→330° at 6Hz; hide app → `visibilityState: 'hidden'`, transform frozen, 0.7 ms CPU/s; show → immediate resume (clock ticks once on reconcile so the ring never looks stuck). A/B/A table above from the same session via injected `<style>` override.
- Benchmarks: Playwright headed Chromium, CPU summed over all Chromium processes via CDP `SystemInfo.getProcessInfo`, 15–20s windows after settle.

## Tests

- `agent-spinner-clock.test.ts`: phase-sync across late joiners, hidden-stop/visible-resume via `visibilitychange`, timer + listener teardown on last unregister, reduced-motion.
- Markup tests updated to the `data-agent-spinner` contract (no CSS animation classes).
- Full renderer component suite, web typecheck, oxlint, oxfmt: green.

## Risk / rollback

- Purely visual + renderer-local; no main/preload/IPC changes. Cadence: 12 steps/s → 6 steps/s with identical 30° step size (full revolution 1s → 2s).
- Failure modes are cosmetic (spinner frozen/stopped); agent status *data* flow untouched. SSH/web unaffected (web clients have real document visibility).

### DIFF
--- a/src/renderer/src/components/AgentStateDot.test.ts
+++ b/src/renderer/src/components/AgentStateDot.test.ts
@@ -22,12 +22,15 @@ describe('AgentStateDot', () => {
 
     expect(markup).toContain('border-yellow-500')
     expect(markup).toContain('border-t-transparent')
-    expect(markup).toContain('[animation:spin_1s_steps(12,end)_infinite]')
-    expect(markup).toContain('motion-reduce:animate-none')
-    // Why: under reduced motion the top border is filled so the frozen ring
-    // reads as a complete static marker, not a broken partial spinner.
+    // Why: rotation comes from the shared agent-spinner clock (which also
+    // honors prefers-reduced-motion), not a per-element CSS animation that
+    // would keep the compositor awake.
+    expect(markup).toContain('data-agent-spinner')
+    // Why: under reduced motion the top border is filled so the static ring
+    // reads as a complete marker, not a broken partial spinner (#9515).
     expect(markup).toContain('motion-reduce:border-t-yellow-500')
     expect(markup).not.toContain('animate-spin')
+    expect(markup).not.toContain('animation:spin')
   })
 
   it('renders done as an emerald check icon', () => {
--- a/src/renderer/src/components/AgentStateDot.tsx
+++ b/src/renderer/src/components/AgentStateDot.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import { CircleCheck } from 'lucide-react'
 import { cn } from '@/lib/utils'
+import { AgentWorkingSpinner } from '@/components/AgentWorkingSpinner'
 
 // Why: shared state-indicator primitive so the dashboard and the sidebar's
 // agent hover share a single state vocabulary. Most states render as a dot;
@@ -74,17 +75,7 @@ export const AgentStateDot = React.memo(function AgentStateDot({
         className={cn('inline-flex shrink-0 items-center justify-center', box, className)}
         aria-label={agentStateLabel(state)}
       >
-        <span
-          className={cn(
-            // Why: match the sidebar worktree spinner's stepped cadence so
-            // long-running visible agents do not keep a full-frame-rate loop.
-            // Why: under reduced motion the spin is disabled, so fill the top
-            // border too — a frozen transparent-top ring reads as a broken
-            // spinner; a complete ring reads as an intentional static marker.
-            'block rounded-full border-2 border-yellow-500 border-t-transparent [animation:spin_1s_steps(12,end)_infinite] motion-reduce:animate-none motion-reduce:border-t-yellow-500',
-            inner
-          )}
-        />
+        <AgentWorkingSpinner className={inner} />
       </span>
     )
   }
--- a/src/renderer/src/components/AgentWorkingSpinner.tsx
+++ b/src/renderer/src/components/AgentWorkingSpinner.tsx
@@ -1,0 +1,23 @@
+import React from 'react'
+import { cn } from '@/lib/utils'
+import { agentSpinnerRef } from '@/lib/agent-spinner-clock'
+
+// Why: the working-state ring must not carry its own infinite CSS animation —
+// that keeps the compositor awake per element for the whole agent run. The
+// shared agent-spinner clock rotates every mounted ring in phase and stops
+// cleanly when nothing can be seen. Callers size it via className (size-2 etc.).
+export function AgentWorkingSpinner({ className }: { className?: string }): React.JSX.Element {
+  return (
+    <span
+      ref={agentSpinnerRef}
+      data-agent-spinner=""
+      className={cn(
+        // Why: under reduced motion the clock never ticks, so fill the top
+        // border too — a frozen transparent-top ring reads as a broken
+        // spinner; a complete ring reads as an intentional static marker (#9515).
+        'block rounded-full border-2 border-yellow-500 border-t-transparent motion-reduce:border-t-yellow-500',
+        className
+      )}
+    />
+  )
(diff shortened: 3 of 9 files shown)


nwparker:
**Cadence update after review feedback:** 3Hz/45° read too step-like. Now **6Hz with 30° steps** — the same step *size* as the old CSS animation (that grain is what read as smooth), at half its tick rate.

Benchmarked the whole cadence ladder before picking (isolated Chromium, 8 spinners + filler page):

| Variant | CPU |
| --- | --- |
| old CSS `steps(12)`/1s | 23.6 ms/s |
| js 6Hz/30° **(new)** | 33.9 ms/s isolated · **10.0 ms/s live A/B** (vs 23.5 old, same session) |
| js 8Hz/30° | 28.3 ms/s |
| js 10Hz/30° | 30.3 ms/s |
| js 12Hz/30° (old look 1:1) | 39.1 ms/s — **worse than the CSS animation** |

Two takeaways baked into the code comment: (1) the isolated numbers overstate JS-clock cost vs the live app (timer wakeup floor dominates the synthetic page; in the real renderer the live A/B is 10.0 vs 23.5 ms/s, and the clock is flat in spinner count while CSS scales per element); (2) **do not smooth further by raising Hz** — at 12Hz the JS clock costs more than the CSS animation it replaced. The hidden-window stop (0.5 ms/s) is cadence-independent and unchanged.

Verified live: spinner steps 0°→30°→60°… at 6Hz in the dev app.

nwparker:
**Cadence finalized at 12 steps/s (30°/step) — exact visual parity with the old animation.** 6Hz was judged noticeably less smooth in a side-by-side comparison window, so the "save CPU by ticking slower" angle is dropped entirely; the clock now matches the retired CSS animation frame-for-frame.

Re-measured at full parity, live dev app, interleaved A/B/A/B (one working agent, 20s legs, renderer+GPU CPU):

| Leg | Renderer | GPU | Total |
| --- | --- | --- | --- |
| clock 12Hz | 33.0 | 28.0 | **61.0 ms/s** |
| old CSS | 57.5 | 28.5 | **86.0 ms/s** |
| clock 12Hz | 33.5 | 27.5 | **61.0 ms/s** |
| old CSS | 47.0 | 29.0 | **76.0 ms/s** |

**Bottom line: identical look, ~20 ms CPU/s saved (~2% of a core, ~25% of spinner-path cost) with one spinner — clock legs dead-stable at 61.0/61.0 while CSS legs wander 76–86.** GPU cost is cadence-bound and identical (~28 ms/s both ways); the entire win is renderer-side per-element animation machinery, which also means it scales with agent count: N CSS animations each pay, the clock pays once.

Also sanity-checked the claim in a *minimal* Electron window (8 rings, nothing else): both modes measure ~2 ms/s — i.e. the CSS animation tax only materializes in a renderer with a real DOM/compositor load like Orca's. Which is exactly why the earlier isolated benchmarks undersold the clock and the live app numbers are the ones to trust.

Hidden-window behavior unchanged (clock stops via native `visibilitychange` post-#9395; measured 0.7 ms/s). Full renderer suite green at the new cadence (1168 files / 9420 tests).

PR #9389, opened 2026-07-18
fix(rate-limits): surface Fable weekly usage when is_active is false (#8979)

### LINKED ISSUE #8979: [Bug]: Fable weekly bar still hidden in v1.4.143 — weekly_scoped limit is filtered out when is_active is false
### Operating system

Windows

### Orca version

v1.4.143

### Details

Short summary:
The Fable weekly usage bar is still missing from the Claude usage tracker in v1.4.143, even though the fix for #8316 shipped. The cause is the `is_active !== false` filter in `mapFableWeeklyWindow` (src/main/rate-limits/claude-fetcher.ts): Anthropic's `/api/oauth/usage` endpoint returns the Fable pool with `is_active: false` whenever it isn't the currently binding limit, so the entry is discarded and the bar never renders.

What happened?
On my account, `GET https://api.anthropic.com/api/oauth/usage` currently returns (trimmed):

```json
"limits": [
  { "kind": "session",       "percent": 15, "is_active": true,  "scope": null },
  { "kind": "weekly_all",    "percent": 3,  "is_active": false, "scope": null },
  { "kind": "weekly_scoped", "percent": 6,  "is_active": false,
    "scope": { "model": { "id": null, "display_name": "Fable" } } }
]
```

The `weekly_scoped` / Fable entry has valid data (6% used, valid `resets_at`), but because `is_active` is `false`, `mapFableWeeklyWindow` skips it, the legacy fallback fields (`fable_weekly` / `fable_seven_day` / `seven_day_fable`) are absent from the response, and the tracker shows only Session and Weekly.

Note the inconsistency: `weekly_all` above is also `is_active: false`, yet the Weekly bar *is* shown — because it's mapped from the legacy top-level `seven_day` field, which has no `is_active` gate. So the same flag hides one bar but not the other. `is_active` appears to mean "currently the binding constraint," not "this limit exists/applies" — Claude's own surfaces display the Fable meter for accounts in this state, and third-party trackers that read `limits` without the flag check show it correctly.

How can we reproduce it?
1. Sign in to Orca with a Claude account whose `/api/oauth/usage` response contains a `weekly_scoped` Fable limit with `is_active: false` (appears to be the normal state unless the Fable pool is the limit you're closest to exhausting).
2. Open the usage tracker tooltip/status bar.
3. Only Session and Weekly render; no Fable bar.

Suggested fix:
Drop the `limit.is_active !== false` condition in `mapFableWeeklyWindow` (or treat `is_active` as display metadata only). The `Number.isFinite(limit.percent)` and display-name checks are sufficient to know the pool exists.

OrcaWin:
## What & why

Fixes #8979. The Fable weekly usage bar disappeared from the Claude usage tracker because `mapFableWeeklyWindow` (`src/main/rate-limits/claude-fetcher.ts`) discarded the Fable `weekly_scoped` entry whenever Anthropic returned it with `is_active: false`.

`is_active` marks **which limit is currently binding**, not whether an entry has valid data. Anthropic sets it to `false` on the Fable pool whenever a different limit is the tightest, but the entry still carries a real `percent` and `resets_at`. The finder's `is_active !== false` guard dropped it, and with no legacy `fable_weekly`/`fable_seven_day`/`seven_day_fable` field present, the mapper returned `null` and the bar never rendered. (The neighbouring `weekly_all` bar, mapped from a legacy field with no `is_active` gate, kept rendering — hence the inconsistency users saw.)

**Change:** remove the `is_active !== false` condition from the Fable finder. The `kind === 'weekly_scoped'`, finite-`percent`, and Fable display-name guards remain, so unrelated/malformed entries are still excluded. Also corrects `docs/claude-scoped-oauth-usage-limits.md`, which previously prescribed the buggy behavior ("treat `is_active: false` as unavailable") and could have led someone to reintroduce it.

## Evidence (proof of improvement)

- **Before:** an `/api/oauth/usage` payload whose only Fable source is an inactive `weekly_scoped` entry (`percent: 64`, valid `resets_at`) → `mapFableWeeklyWindow` returns `null` → **bar hidden**.
- **After:** the same payload → `{ usedPercent: 64, resetsAt: … }` → **bar renders**.
- New regression test `surfaces an inactive scoped Fable entry when no legacy Fable field exists (#8979)` fails on the pre-fix code (`fableWeekly: null`) and passes after.

## Proof of no regression

- `npx vitest run --config config/vitest.config.ts src/main/rate-limits/claude-fetcher.test.ts` → **31 passed**.
- Session (`five_hour`) and weekly-all (`seven_day`) windows are computed from separate fields and never touch this finder — byte-identical (verified by reviewer reading the call site).
- `pnpm run typecheck:node` clean; `oxlint` clean.
- The only behavioral change is Fable-scoped precedence (an inactive scoped entry now wins over a legacy field), which is the intended contract and is pinned by an updated test.
- Adversarially reviewed (independent Sonnet pass): verdict CLEAN.

## ELI5

Anthropic's usage API reports several quota "meters" and flags whichever one is currently the tightest with `is_active`. Orca was treating any Fable meter that *wasn't* the current tightest as if it had no data and threw it away — so the Fable bar vanished whenever some other limit was binding. Now Orca reads the Fable meter's real percentage and reset time no matter which meter is currently binding.

## Screenshots (before / after)

Real render of the shipping `ProviderPanel` usage panel (`src/renderer/src/components/status-bar/tooltip.tsx`) — the exact component the fix's data path feeds. Percentages/reset times are representative synthetic values; the meaningful difference is the **Fable row absent (before) vs present (after)**.

| Before (bug: Fable row dropped) | After (fix: Fable row renders) |
|---|---|
| before – Fable hidden | after – Fable visible |

Before: only **Session** (28%) and **Weekly** (47%) render — the Fable weekly window is silently dropped because the inactive `weekly_scoped` entry is filtered out. After: a third **Fable** row (64% used) renders once `mapFableWeeklyWindow` stops discarding `is_active: false` entries.


brennanb2025:
Taking a look!

brennanb2025:
Maintainer follow-up: I reviewed the landed mapper change and regression coverage closely. Ignoring `is_active` for a finite, correctly scoped Fable entry is the right behavior because the flag identifies the currently binding limit while the entry still contains valid usage/reset data; the malformed/unrelated guards and legacy fallbacks remain intact. I also verified the merged result with all 36 focused fetcher tests, node typecheck, touched-file lint/format, and the max-lines ratchet—all clean, with no follow-up code changes needed.

PR #8265, opened 2026-07-11
fix: restore the active top-level view across renderer reload

myungjuice:
## Summary / description

The active top-level view (Tasks, Automations, Mobile, Settings, Skills, Space, or the enabled Activity view) now survives a renderer reload or app relaunch instead of always resetting to the terminal.

`activeView` now uses Orca's existing persisted UI pipeline. Startup hydration restores a validated value once, while later `ui:stateChanged` broadcasts continue syncing preferences without unexpectedly navigating an already-open window. Unknown values and a disabled experimental Activity view safely fall back to the terminal.

Closes #8264

## Evidence of fix

- Added an Electron E2E regression test that launches Orca twice against the same user-data directory. It opens Tasks on the first launch, verifies `activeView` reached persisted main-process state, relaunches, and verifies the Tasks DOM is visible while the terminal DOM is hidden.
- E2E result on current `main`: `1 passed (8.9s)`; the restore scenario itself completed in `8.2s`.
- Focused renderer tests: `151 passed` across `ui.test.ts` and `startup-ui-hydration.test.ts`.
- All node, CLI, and web TypeScript targets passed.
- The E2E-mode Electron build, focused Oxlint checks, diff whitespace check, and max-lines ratchet passed.

## ELI5

Orca remembered the details inside each screen, but it forgot which screen you were looking at. This change saves that screen name too. When Orca starts again, it checks that the saved name is still valid and opens that screen; background settings updates are not allowed to pull you away from the screen you are currently using.

## Trade-offs / expected user regressions

No user regression is expected. The value uses the existing profile-level UI state, so the last desktop view written is the one restored on the next launch; a live window is never navigated by a cross-window/mobile sync. Like the surrounding UI preferences, writes are debounced by 150 ms, so a reload triggered almost immediately after navigation could still restore the prior view.

## Screenshots

No visual change. The E2E test provides render-layer evidence for both the restored Tasks page and hidden terminal surface.

## Testing

- [ ] `pnpm lint` (focused Oxlint and the max-lines ratchet passed; the full repository lint suite was not run)
- [x] `pnpm typecheck`
- [ ] `pnpm test` (focused suites passed: 151 tests)
- [ ] `pnpm build` (the E2E-mode Electron build passed; the full release build was not run)
- [x] Added or updated high-quality tests that would catch regressions

Commands run:

```text
[...]

PR #9393, opened 2026-07-18
perf(renderer): stop full durable-state save on every top-level view switch (#9002)

### LINKED ISSUE #9002: [Bug]:
### Operating system

Windows

### Orca version

>v1.4.138

### Environment

- OS: Windows 11 24H2 (also reproducible on another Windows machine)
- Orca version: v1.4.138, v1.4.139
- Last known good version: v1.4.137

### Description

Since upgrading from **v1.4.137** to **v1.4.138**, switching between application views causes a noticeable UI freeze.

The freeze is short (roughly 100–300ms), but it is very noticeable and happens almost every time.

This issue continues to exist in **v1.4.139**.

### Steps to reproduce

1. Launch Orca.
2. Open Settings.
3. Switch between different Settings pages (General, Appearance, etc.), or switch between other top-level views.
4. Observe a brief freeze before the new page is rendered.

### Expected behavior

Image

View transitions should be smooth, as they were in v1.4.137.

### Actual behavior

The entire window appears to pause briefly before the target view is displayed.

This is not limited to Settings—it also occurs in other common UI pages.

### Additional observations

- The freeze affects the whole window rather than only the page content.
- It appears to happen regardless of project size.
- It is reproducible even when no heavy background tasks are running.
- Downgrading to **v1.4.137** makes the issue disappear.

### Possible cause (speculation)

I'm not sure what changed in v1.4.138, but since the issue started exactly after upgrading from v1.4.137, it might be related to one of the following:

- Renderer lifecycle or view restoration changes.
- A synchronous operation executed during view switching.
- Global state updates causing unnecessary re-rendering of the application shell.
- A Windows-specific rendering or IPC regression.

I haven't profiled it yet, but the behavior feels like the renderer's main thread is briefly blocked during navigation.

If there's anything I can test or any logs/profiling information that would help, I'd be happy to provide them.

OrcaWin:
## What & why

Fixes #9002. Since #8265, changing `activeView` flowed through the renderer's 150 ms UI writer into Orca's monolithic durable state. The main process then hashed and serialized the full multi-MB recovery snapshot synchronously before its asynchronous disk write, so a tiny navigation preference could freeze the whole window.

This revision gives `activeView` its own profile-scoped persistence boundary. Switching top-level views no longer mutates or schedules the global durable snapshot.

## Design

- `active-view.json` stores only the current top-level view beside the active Orca profile.
- Active-view-only updates take an early return before broad UI normalization, equality checks, or `Store.scheduleSave()`.
- A 100 ms trailing writer coalesces rapid switches and uses a temporary file plus atomic rename.
- The existing desktop/mobile UI synchronization path remains intact.
- On first upgraded launch, the legacy `orca-data.json` value is used as the fallback and migrated when the renderer reasserts its hydrated view.
- The legacy field remains unchanged as a downgrade fallback; current builds treat the sidecar as authoritative.
- Graceful reload, restart, and quit use one guarded synchronous checkpoint for the latest active view and renderer-owned workspace-session snapshots. Quit cancellation resets the checkpoint guard so a later close can capture fresh state.

## Evidence

The persistence-boundary regression test proves that an active-view switch:

- produces a sidecar payload smaller than 64 bytes;
- leaves `orca-data.json` byte-for-byte unchanged;
- restores the new view from a fresh Store instance.

Verification completed on Windows:

- 724 tests across all affected persistence, IPC, preload, renderer, and shutdown suites;
- full `pnpm run typecheck` coverage across node, CLI, and web targets;
- full lint, switch-exhaustiveness, reliability, max-lines, localization catalog, and localization coverage gates;
- the real Electron restart journey passed five consecutive repetitions (ten app launches total), visibly restoring Tasks after restart with the terminal surface hidden;
- diff hygiene checks passed.

## Correctness and tradeoffs

- Profile isolation is preserved because the preference file lives in each profile directory.
- SSH/runtime workspace sessions keep their existing per-host ownership; only the local top-level navigation preference moved.
- A hard process kill inside the 100 ms debounce window can still restore the prior view. After that tiny window, crash recovery is better than the previous blur/exit-only design, and every graceful exit flushes synchronously.
- The sidecar is an ephemeral navigation preference. Terminal commands, repositories, scrollback recovery, and project data remain in their existing stores.

## Scope note for the Windows typing report

This removes the measured full-state work caused by top-level view switches. It does not claim to fix a separate freeze that occurs continuously while typing without switching views; that report needs its own reproduction and input/store-subscriber investigation.

## ELI5

Changing one small preference used to make Orca inspect and potentially rewrite its entire saved workspace. Now Orca writes a tiny note containing only the current view, so it can remember every switch without dragging terminal and recovery state into the interaction path.

### DIFF
--- a/src/main/active-view-persistence-boundary.test.ts
+++ b/src/main/active-view-persistence-boundary.test.ts
@@ -1,0 +1,58 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { mkdtempSync, readFileSync, rmSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { getActiveViewPreferenceFile } from './active-view-preference'
+
+const testState = { dir: '' }
+
+vi.mock('electron', () => ({
+  app: { getPath: () => testState.dir },
+  safeStorage: {
+    isEncryptionAvailable: () => true,
+    encryptString: (plaintext: string) => Buffer.from(`encrypted:${plaintext}`, 'utf-8'),
+    decryptString: (ciphertext: Buffer) => ciphertext.toString('utf-8').replace('encrypted:', '')
+  }
+}))
+
+vi.mock('./telemetry/client', () => ({ track: vi.fn() }))
+vi.mock('./telemetry/cohort-classifier', () => ({ getCohortAtEmit: vi.fn() }))
+vi.mock('./ssh/ssh-config-parser', () => ({
+  loadUserSshConfig: vi.fn(),
+  sshConfigHostsToTargets: vi.fn()
+}))
+
+describe('active-view persistence boundary', () => {
+  beforeEach(() => {
+    testState.dir = mkdtempSync(join(tmpdir(), 'orca-active-view-boundary-'))
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+    rmSync(testState.dir, { recursive: true, force: true })
+  })
+
+  it('persists a view switch without changing the global durable snapshot', async () => {
+    vi.useFakeTimers()
+    vi.resetModules()
+    const { Store } = await import('./persistence')
+    const dataFile = join(testState.dir, 'orca-data.json')
+    const store = new Store({ dataFile })
+    store.flush()
+    const durableBefore = readFileSync(dataFile, 'utf-8')
+
+    store.updateUI({ activeView: 'settings' })
+    vi.advanceTimersByTime(1_000)
+    await store.waitForPendingWrite()
+
+    const preferenceFile = getActiveViewPreferenceFile(dataFile)
+    const preferencePayload = readFileSync(preferenceFile, 'utf-8')
+    expect(Buffer.byteLength(preferencePayload)).toBeLessThan(64)
+    expect(JSON.parse(preferencePayload)).toEqual({ activeView: 'settings' })
+    expect(readFileSync(dataFile, 'utf-8')).toBe(durableBefore)
+    expect(store.getUI().activeView).toBe('settings')
+
+    const reloaded = new Store({ dataFile })
+    expect(reloaded.getUI().activeView).toBe('settings')
+  })
+})
--- a/src/main/active-view-preference.test.ts
+++ b/src/main/active-view-preference.test.ts
@@ -1,0 +1,80 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { ActiveViewPreference, getActiveViewPreferenceFile } from './active-view-preference'
+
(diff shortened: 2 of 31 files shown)


shanren:
The bug has not been effectively resolved on the Windows platform. Typing text in the command line window frequently causes the interface to freeze. This is a very serious bug that has impacted daily use. Version 1.4.137 was working correctly; upgrading has instead caused significant problems.

nwparker:
> The bug has not been effectively resolved on the Windows platform. Typing text in the command line window frequently causes the interface to freeze. This is a very serious bug that has impacted daily use. Version 1.4.137 was working correctly; upgrading has instead caused significant problems.

@shanren can you share some more info? Are you using Powershell? Gitbash? CMD? Thanks! We'll dig in. The more info the better.

nwparker:
## Adversarial review loop (fable + gpt-5.6-sol) — converged clean

Ran a multi-model correctness/regression review of this PR to completion. Three real issues were found and fixed in this branch (commit `fc4ca65b62`); a second verification pass confirmed each fix is correct with no new regression.

### Issues found & fixed
1. **Prototype-key leak in the view validator** (`src/shared/top-level-view.ts`) — `isTopLevelView` used `value in TOP_LEVEL_VIEW_LOOKUP`, which accepts inherited keys (`constructor`, `__proto__`, `toString`). A corrupt/hand-edited `active-view.json` with `{"activeView":"constructor"}` would pass validation, match no `App.tsx` branch, and blank the main surface instead of falling back to Terminal. This was a regression vs. the prior renderer validator (own-key `Set`). → now `Object.hasOwn(...)`.
2. **TOCTOU race in the debounced async writer** (`src/main/active-view-preference.ts`) — `writeAsync` checked its generation guard and then `await rename(...)`. A synchronous shutdown `flushOrThrow()` could interleave in that await gap, `renameSync` the newer view, and then the stale async rename would clobber it — restoring the prior view on next launch, defeating the exact durability guarantee the sync checkpoint exists to provide. → guard + `renameSync` now run synchronously with no await between them.
3. **Coupled shutdown flushes** (`src/main/ipc/renderer-shutdown-checkpoint.ts`) — the durable-state flush and the active-view sidecar flush shared one `try` block, so a durable-state flush failure silently skipped the independent sidecar flush (and vice versa), undermining "every graceful exit flushes synchronously." → each store now flushes in its own `try/catch`; `ok` is `false` if any step fails.

Regression tests added for all three.

### Zero-regression evidence (local, this branch)
- **746 unit tests pass** across every affected + adjacent suite:
  - 214 across all 12 directly-touched suites (active-view preference/boundary, shutdown checkpoint, checkpoint guard, restart preparation, startup routing, autosave, workspace-session host persistence, web preload, app/window IPC).
  - 532 across `persistence.test.ts` + `store/slices/ui.test.ts` (the store hosting the sidecar and the UI slice with the early-return).
- `pnpm run typecheck` clean (node + CLI + web targets).
- oxlint + oxfmt clean (pre-commit gate).

Full CI gate (verify, Windows terminal-restart E2E, crash-survival, Wayland input) re-running on the new commit.

PR #9399, opened 2026-07-18
feat(native-chat): drop workspace folders into chat

sezer-muhammed:
## What changed

- Include workspace folder/file paths in sidebar card drag payloads.
- Accept those internal paths in the native chat composer and insert the full path as editable text.
- Preserve multiple dropped paths on separate lines and show bounded-drop rejection feedback.

## Why

Dragging a folder from the left workspace sidebar into native chat previously only carried workspace-reordering metadata, so the composer could not receive the folder path.

## Validation

- Focused Vitest suites: 20 tests passed.
- Changed-file oxlint passed.
- Renderer typecheck reaches pre-existing TS6307 project-file errors in src/main unrelated to this change.

## ELI5

You can drag a workspace folder or file from the sidebar into native chat and drop its path into the composer as editable text. Multiple drops land on separate lines so you can point the agent at paths quickly.

### DIFF
--- a/src/renderer/src/components/native-chat/NativeChatComposer.tsx
+++ b/src/renderer/src/components/native-chat/NativeChatComposer.tsx
@@ -25,6 +25,7 @@ import {
 } from './native-chat-composer-target'
 import { useNativeChatComposerAttachments } from './use-native-chat-composer-attachments'
 import { useNativeChatComposerPaste } from './use-native-chat-composer-paste'
+import { useNativeChatWorkspaceDrop } from './use-native-chat-workspace-drop'
 import { useNativeChatExternalAttachments } from './use-native-chat-external-attachments'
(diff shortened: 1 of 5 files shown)


brennanb2025:
Taking a look!

brennanb2025:
Thanks for this! I rebased the branch onto current main and tightened the implementation: workspace-card drags now carry selected runtime paths on the reference-only payload while preserving card reordering, and native chat handles those drops in a dedicated bounded hook with safe disabled/rejection behavior. I also removed the unrelated PTY refactor and replaced the formatting-only test with behavior coverage for supported and ignored drags, Windows/Unicode/space-containing paths, payload limits, and unavailable insertion. Validation is clean: 47 focused tests, changed-file oxlint, React Doctor, the max-lines ratchet, and the full typecheck all pass.

Issue #9403, opened 2026-07-19
[Bug]: Sidebar section UI state resets when switching sidebar tabs


just-zzp:
Operating system: macOS

Short summary:
Sidebar section UI state resets after visiting another right-sidebar tab.

What happened?
Two right-sidebar surfaces lose renderer-session UI state when their panels are unmounted:

- Source Control resets top-level section disclosure state.
- Agent Session History resets the selected scope, collapsed groups, and expanded session details.

How can we reproduce it?

Source Control:
1. Open a Git-backed worktree and select the Source Control tab.
2. Collapse a section such as Changes or Committed on Branch.
3. Switch to Explorer or Checks.
4. Return to Source Control.

Agent Session History:
1. Open Agents in the right sidebar.
2. Select Project or All, collapse a group, and expand a session's details.
3. Switch to Explorer, Source Control, or Checks.
4. Return to Agents.

Expected behavior:
The same worktree or Agents panel should restore these transient choices for the current renderer session. Newly visited contexts should still use their normal defaults, and an app reload may reset the choices.

Actual behavior:
The panel returns to its defaults because the state is component-local and is lost when the sidebar tab switch unmounts the panel.

Canonical scope:

- Source Control section disclosure state: #9405
- Agent Session History scope and disclosure state: #9419 (originally reported as duplicate #9418)

The implementations remain in two focused PRs because the surfaces and state models are independent and their diffs do not overlap.

AmethystLiang:
### Duplicate consolidation

Closing **#9418** as a duplicate of this issue (same author, same failure mode).

Both are sidebar section local UI state that resets after switching sidebar tabs:
- **#9403** — Source Control collapse state
- **#9418** — Agent Session History scope / disclosure state

Canonical title for the combined work: **sidebar section UI state resets when switching sidebar tabs**.

Please fix both surfaces under this issue.

just-zzp:
Thanks for consolidating these. I’m treating #9403 as the canonical issue and have updated the issue title/body and both PR descriptions accordingly.

I’m keeping the implementation split into two focused PRs because the surfaces and state models are independent and their diffs do not overlap:

- #9405 — Source Control section disclosure state
- #9419 — Agent Session History scope and disclosure state

#9405 is the first half, and #9419 is the final half. #9419 carries the closing keyword and should be merged after #9405, so #9403 closes only after both fixes land.

PR #9405, opened 2026-07-19
fix(source-control): preserve section collapse state

just-zzp:
## Summary

- preserve top-level Source Control section disclosure choices when Explorer or Checks unmounts the panel
- scope the renderer-session cache per worktree, retain the existing default of collapsed Commits, and bound non-default entries with LRU eviction
- add remount, worktree-isolation, default restoration, eviction, and real Source Control regression coverage

Part of #9403.

Companion PR: #9419 (Agent Session History).

Merge order: merge this PR before #9419. The companion PR carries the closing keyword so the canonical issue closes only after both surfaces land.

## Screenshots

Captured from an isolated Electron E2E profile backed by a synthetic Git repository. No personal repositories, paths, sessions, or account data are shown.

### Before navigating away

`Changes` and `Untracked Files` are collapsed, while `Staged Changes` remains expanded.

Source Control before navigating away

### Explorer

The right sidebar is switched to Explorer, which unmounts the Source Control panel.

Explorer between Source Control visits

### After returning

Returning to Source Control restores the same mixed disclosure state. The Before and After captures are byte-identical.

Source Control after returning from Explorer

## Testing

- [ ] `pnpm lint`
  - Changed-file oxlint, react-doctor, and type-aware lint pass. The full switch-exhaustiveness run currently fails on the unrelated existing `src/renderer/src/components/skills/skill-freshness-group.tsx:101`.
- [x] `pnpm typecheck`
  - Node, CLI, and Web TypeScript projects pass.
- [ ] `pnpm test`
  - Source Control scope passes: 47 test files / 429 tests. The full suite did not complete because native runtime install scripts were intentionally skipped; unrelated PTY, daemon, and Electron-backed suites failed during setup.
- [ ] `pnpm build`
  - The full production build was not run because the local environment does not have all required native runtime artifacts. The Electron E2E build passes with Node 26 using `electron-vite build --mode e2e`.

Mock-data Electron verification passes on macOS: `Source Control → Explorer → Source Control` with DOM assertions that the collapsed `Changes` and `Untracked Files` entries stay hidden while the expanded staged entry remains visible (`1 passed`).

Additional checks: changed files are formatted; max-lines ratchet, reliability gate manifest, styled scrollbar check, and `git diff --check` pass.

## AI Review Report

Two independent AI reviews found no blocking correctness issues. The reviews explicitly checked remount behavior, rapid toggle semantics, worktree transitions and isolation, bounded cache behavior, performance, and regression coverage. A low-severity review note observed that the cache test only checked its final size, so the test was strengthened to verify that the oldest entry is actually evicted while the newest remains.

Runtime verification was performed through Electron E2E on macOS with an isolated mock repository. Linux, Windows, WSL, and SSH were not manually exercised. Cross-platform code review covered those environments: this change does not touch shortcuts, labels, paths, shell behavior, Git commands, or Electron platform APIs. SSH/local/remote behavior remains neutral because the state is keyed by the existing worktree identity and lives only in renderer memory.

No Git-provider integration path is changed; GitHub, GitLab, and other supported providers retain the same behavior. No agent or integration behavior changes.

UI quality is unchanged: there are no layout, color, typography, spacing, design-token, or component-selection changes. Existing disclosure controls and their default states are retained.

Residual non-blocking risks reviewed: the cache assumes the current single Source Control consumer, and recreating an identical worktree path in the same renderer session can restore its previous transient disclosure choice. Both match existing renderer-session disclosure patterns and are bounded by the 512-entry LRU.

## Security Audit

The change stores only typed section identifiers in a bounded in-memory Map. It adds no user-input parsing, command execution, path handling, network access, auth/secrets handling, IPC, persistence schema, or dependencies. No security follow-up is required.

## Notes

- Disclosure choices intentionally reset on app reload; this fix targets navigation/remount loss within the current renderer session.
- Default behavior for a newly visited worktree is unchanged.
- I don't have an X account.

## ELI5

Collapsing or expanding Source Control sections was lost whenever Explorer or Checks remounted the panel. Your open/closed choices now stick for that worktree during the session.

### DIFF
--- a/src/renderer/src/components/right-sidebar/SourceControl.tsx
+++ b/src/renderer/src/components/right-sidebar/SourceControl.tsx
@@ -115,6 +115,7 @@ import {
   type SourceControlEntryGroups,
   type SourceControlSectionArea
 } from './source-control-section-order'
+import { useSourceControlSectionCollapseState } from './source-control-section-collapse-state'
 import { SourceControlVirtualFileList } from './source-control-virtual-file-list'
 import { selectReviewCacheData, selectReviewCacheEntry } from './review-cache-entry-selection'
 import {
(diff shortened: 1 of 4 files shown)


brennanb2025:
Taking a look!

PR #9408, opened 2026-07-19
fix(status-bar): tolerate missing provider snapshots

nsxdavid:
## Summary

Fixes a status-bar crash that occurred when a rate-limit snapshot from an older main process omitted a provider key.

## ELI5

Sometimes the app's bottom bar asked for provider info that hadn't arrived yet and choked on the empty answer. Now it waits calmly instead of crashing.

## Root cause & fix

`isProviderConfigured` and `isProviderSnapshotPending` used a strict `provider === null` check, so an `undefined` snapshot (which happens when an older main process omits newly added provider keys) slipped past the guard and crashed on `provider.status`. The fix switches to a loose `provider == null` check so both `null` and `undefined` snapshots are treated as pending. Configured providers stay visible while their snapshot loads; unconfigured providers stay hidden. No provider payloads or IPC contracts change.

## Evidence

- Test file: `src/renderer/src/components/status-bar/status-bar-provider-visibility.test.ts`
- Fix file: `src/renderer/src/components/status-bar/status-bar-provider-visibility.ts`
- On branch: all 29 tests pass.
- Reverting only the fix (keeping the new tests) reproduces the bug: 4 fail with `TypeError: Cannot read properties of undefined (reading 'status')`, including the older-main-process regression tests.
- Lint: `oxlint` on the fix source is clean; PR reports `pnpm typecheck` passing.

```
STEP 2a — on branch: Test Files 1 passed, Tests 29 passed.
STEP 2b — fix reverted: Tests 4 failed | 25 passed.
  TypeError: Cannot read properties of undefined (reading 'status')
  - getVisibleUsageProvider > creates a pending snapshot when an older main process omits a configured provider
  - isUsageEmptyState > treats provider keys omitted by an older main process as pending
```

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression: the change only widens a null check to also catch `undefined`, so already-defined and `null` snapshots follow byte-identical paths. Behavior for configured, pending, and empty-state providers is unchanged and locked in by the passing 29-test suite. No IPC, payload, or platform-specific code is touched.

_Credit to @nsxdavid for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/renderer/src/components/status-bar/status-bar-provider-visibility.test.ts
+++ b/src/renderer/src/components/status-bar/status-bar-provider-visibility.test.ts
@@ -30,6 +30,7 @@ function provider(
 describe('isProviderConfigured', () => {
   it('hides a provider whose state has not loaded yet', () => {
     expect(isProviderConfigured(null)).toBe(false)
+    expect(isProviderConfigured(undefined)).toBe(false)
   })
 
   it('hides an unconfigured (unavailable) provider', () => {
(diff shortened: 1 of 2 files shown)


brennanb2025:
Taking a look!

nwparker:
Thanks @nsxdavid I'm going to try to merge now. Give me a few mins here

nwparker:
Thanks so much for this, @nsxdavid — tolerate missing provider snapshots. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #9410, opened 2026-07-19
fix(settings): gate project path setup on host connection

s546126:
## Summary

Prevents project import and clone actions from starting while the selected remote (SSH) host is disconnected, connecting, or errored, so they no longer fail with `SSH connection ... not found or not connected`.

## ELI5

If your remote computer isn't actually connected yet, the "browse folder" and "clone from URL" buttons now stay off and tell you to connect first — instead of letting you click them and hit a confusing error.

## Root cause & fix

GSSAPI authentication can succeed even though Orca's remote provider isn't connected, yet the setup UI gated path actions on host *capability* alone and never checked provider *availability*. This change adds a health check when building the host options: connected hosts keep full path actions, while disconnected/connecting/errored hosts have path actions disabled with localized recovery guidance. The offline "Add host placeholder" flow stays enabled so users can still record work for later.

## Evidence

- Rebase clean onto `origin/main`; PR touches 10 files (3 source, 5 i18n locales with the new `hostSetupConnectionRequired` key, 2 tests).
- Tests: `repository-host-setup-options.test.ts` + `RepositoryHostSetupsSection.test.tsx`.
- On branch: **18/18 pass.** Reverting the 3 fix files → **5 fail / 13 pass**, e.g. `buildSetupHostOptions > enables connected SSH hosts` expected `canUsePathActions:true` + connect guidance but got detail `'SSH'` / `canUsePathActions` undefined; the 3 disconnected/connecting/error cases and the component-level `blocks path setup until SSH host connects` also fail.
- Lint: `oxlint` on the 3 changed source files → clean (0 warnings, 0 errors).

```
On branch:      Test Files 2 passed (2) | Tests 18 passed (18)
Fix reverted:   Test Files 2 failed (2) | Tests 5 failed | 13 passed (18)
```

## Trade-offs

None — pure fix. Adds one constant-time health check while mapping the existing host list; no new IPC, network, or polling.

## Regression risk

Zero-regression by construction: only the disconnected/connecting/errored branch changes behavior. Local hosts and connected remote hosts keep their existing treatment, and reconnecting automatically restores actions. The passing branch tests plus fail-on-revert confirm the gate is the only behavioral change.

*Original fix by @s546126; validated, rebased, and reviewed via automated bug-bash triage — original fix design preserved, no additional fix required.*

### DIFF
--- a/src/renderer/src/components/settings/RepositoryHostSetupActions.tsx
+++ b/src/renderer/src/components/settings/RepositoryHostSetupActions.tsx
@@ -68,11 +68,16 @@ export function RepositoryHostSetupActions({
   const [isCloning, setIsCloning] = useState(false)
   const [isCreatingPendingSetup, setIsCreatingPendingSetup] = useState(false)
   const defaultSetupHostOption =
-    setupHostOptions.find((option) => option.isAvailable) ?? setupHostOptions[0] ?? null
+    setupHostOptions.find((option) => option.isAvailable && option.canUsePathActions) ??
+    setupHostOptions.find((option) => option.isAvailable) ??
+    setupHostOptions[0] ??
+    null
   const setupTargetHostId = selectedSetupHostId ?? defaultSetupHostOption?.id ?? null
   const setupTargetHostOption =
     setupHostOptions.find((option) => option.id === setupTargetHostId) ?? null
   const canUseSetupTargetHost = setupTargetHostOption?.isAvailable ?? false
+  const canUsePathActions =
+    canUseSetupTargetHost && (setupTargetHostOption?.canUsePathActions ?? false)
 
   if (setupHostOptions.length === 0) {
     return null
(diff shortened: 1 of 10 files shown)


brennanb2025:
Taking a look!

brennanb2025:
Thanks for the fix! I reviewed the host-health path end to end and pushed one follow-up regression assertion: disconnected SSH targets keep import/clone disabled while offline placeholders remain available, and both path actions re-enable immediately once the host connects.

The action-specific capability split is the right approach here: it prevents invalid remote-provider calls without blocking offline planning, stays reactive to connection changes, and adds only constant-time checks over the existing host list. Focused host/setup tests (30), typecheck, touched-file lint/format, React checks, reliability/max-lines gates, and locale catalog validation pass; full lint remains blocked only by unrelated pre-existing switch-exhaustiveness failures on the base branch.

nwparker:
Thanks so much for this, @s546126 — gate project path setup on host connection. Reviewed, rebased on latest `main`, and merged. Really appreciate the contribution! 🙏

Issue #9414, opened 2026-07-19
Tracking: recognition-independent terminal lifecycle safety + generic agent discovery


bbingz:
# Tracking: recognition-independent terminal lifecycle safety + generic agent discovery

## Problem

Orca classifies terminal panes against a closed allowlist of agent CLIs (the
`TuiAgent` union materialized in `src/shared/tui-agent-config.ts`). Agents
outside that list are invisible to the app, which produces two user-visible
failures:

1. **Invisible running agents.** A genuine `claude` binary launched through a
   user shell wrapper that sets `CLAUDE_CONFIG_DIR=$HOME/.claude-<name>` (a
   common pattern for pointing Claude Code at alternate Anthropic-compatible
   endpoints) never loads Orca's managed hooks, because hook settings are only
   written to `~/.claude/settings.json`
   (`src/main/claude/hook-settings.ts:63`, remote variant `:81`). The pane
   renders as a bare terminal: no running ring, no sidebar agent row, no
   resume records, no keep-awake. Renamed or forked CLIs fall through every
   detection layer entirely.

2. **Unrecognized agents can be killed by automation.** The remote-mirror
   close chain converts transient transport signals (`no_connected_pty`,
   `terminal_handle_stale`, stream end) into host-side PTY kills (#8871
   family): the client fires `closeTerminalTab(reason:'pty-exit')`, the close
   echoes to the host as `session.tabs.close` (which carries no intent), and
   `closeMobileSessionTab` kills the live PTY unconditionally
   (`src/main/runtime/orca-runtime.ts:4714`). Recognized agents at least hold
   resume records and keep-awake assertions; unrecognized ones get neither
   the protections nor the visibility, so they die silently.

## Root causes

- **RC1 — hook install is CLAUDE_CONFIG_DIR-blind.** The install target is
  hardcoded to `~/.claude` / `~/.openclaude`
  (`ClaudeCompatibleHookSettings.configDirName` is a closed union). Everything
  downstream is already open: `ORCA_PANE_KEY` is injected into every pane
  agent-agnostically (`src/main/ipc/pty.ts:3929`), the hook server resolves
  agentType from the endpoint path with no allowlist check
  (`src/main/agent-hooks/server.ts`), and `AgentType` is an open string.
  Only the settings visibility is missing.

- **RC2 — recognition collapses to null.** The process scan already resolves
  wrapper-launched claude (full argv,
  `src/shared/agent-process-recognition.ts:284`), but the result only feeds
[...]

PR #8872, opened 2026-07-15
fix(runtime): typed close intent + host-side adjudication for session tab closes

bbingz:
# fix(runtime): typed close intent + host-side adjudication for session tab closes

Part of #9414. Reworked from this PR's original client-side approach — see the
rework comment below. The intent field is consumed by the lifecycle-state
follow-up and by #8888 attribution logging.

## Problem

`session.tabs.close` carried only `{worktree, tabId}`. The host could not tell
a user-initiated close from a transport-artifact echo: when a paired client
loses a mirrored PTY stream it fires `closeTerminalTab(reason: 'pty-exit')`,
the reason was dropped before the RPC, and `closeMobileSessionTab` killed live
PTYs unconditionally. A transient transport blip could therefore terminate a
working agent on the host.

## Fix

Thread a typed close reason end-to-end and make the host the sole decision
point:

- **Wire schema** — `session.tabs.close` params gain an optional
  `reason: 'user' | 'pty-exit' | 'cleanup'` (`CloseTab` in
  `session-tabs-schemas.ts`). Reasonless payloads still validate.
- **Client threading** — `closeTerminalTab` already had `options.reason`
  locally but dropped it; it now flows through `closeWebRuntimeSessionTab` and
  `web-runtime-session.ts` onto the RPC. Direct user actions (tab close, bulk
  close, tab-group close, mirrored editor close, mobile tab close, CLI
  handle-addressed close) send `reason: 'user'`. Callers whose origin is not
  clearly a user action stay reasonless — the client never invents an
  explicit-close claim.
- **Host adjudication** — in `closeMobileSessionTab`, a close whose reason is
  present and not `'user'` is treated as an echo, not an authorization. If any
  leaf under the addressed terminal parent still resolves a **connected** PTY
  record (`findPtyForMobileTerminalTab(...)?.connected === true` after
  `refreshMobileSessionPtyRecords` — record presence alone is not liveness:
  exited PTYs stay addressable as disconnected records in `ptysById`), the
  host refuses **every** destructive branch — direct `ptyController.kill`, the
  whole-parent renderer close relay, and the headless teardown/kill loop — and
  returns `{ closed: true, refused: true }`. Unless the echo addressed a dead
  leaf under a live sibling (below), it also republishes the session snapshot
  with a bumped `snapshotVersion` so the echoing client re-syncs and
  re-attaches, and marks the response with additive
  `{ snapshotRepublished: true }`.
  When the PTY is genuinely dead/absent (including a retained
[...]

Issue #8711, opened 2026-07-14
Codex status hooks are dead in SSH worktrees — CODEX_HOME is injected only by the local PTY provider, while `orca agent hooks status` still reports `codex: installed`


eisen0419:
## Summary

Orca's Codex status hooks are **process-scoped**: they live in an Orca-managed `CODEX_HOME`, and Codex only finds them if that env var is injected into the PTY. The **local** PTY provider injects it. The **SSH** PTY provider never does.

As a result, every Codex agent running in an SSH worktree falls back to the user's own `~/.codex/hooks.json` on the remote host — which contains no Orca hooks — so **Orca's Codex status hooks never fire at all over SSH**.

Worse, `orca agent hooks status` inspects the **local machine's** managed `CODEX_HOME` (where the hooks *are* correctly installed) and cheerfully reports `codex: installed`. The diagnostic is green while the feature is 100% dead on the host that actually runs the agent.

## Reproduction

Two terminals, same Orca instance, same Codex, differing only in where the worktree lives:

```bash
# In a LOCAL worktree
orca terminal create --worktree <local-worktree> --command 'echo "CODEX_HOME=[${CODEX_HOME:-unset}]"'

# In an SSH worktree
orca terminal create --worktree <ssh-worktree>   --command 'echo "CODEX_HOME=[${CODEX_HOME:-unset}]"'
```

| worktree | `CODEX_HOME` in the PTY |
|---|---|
| local | `/Users/<me>/Library/Application Support/orca/codex-runtime-home/home` ✅ |
| **SSH** | **`unset`** ❌ |

And on the remote host, inside the SSH-worktree Codex:

```console
$ grep -c orca ~/.codex/hooks.json
0                      # ← no Orca hooks anywhere Codex will look
```

Meanwhile:

```console
$ orca agent hooks status
agentStatusHooksEnabled: true
settingsPath: /Users/<me>/Library/Application Support/orca/orca-data.json
codex: installed       # ← green, and wrong for every SSH agent
```

**Behavioral confirmation that Codex really is reading the user's own home:** the remote Codex kept firing the user's *personal* `PostToolUse` hook from `~/.codex/hooks.json` (a stale binary, `exit 127`) throughout the session. Orca's managed hook script is written defensively (`if [ -f … ] && [ -x … ]; then …; else cat >/dev/null; fi`) and could never produce that error — so the hooks being executed were unambiguously the user's, not Orca's.

## Root cause

[...]

PR #9434, opened 2026-07-19
feat(foreground): recognize renamed agent binaries via executable image + SSH foreground confirmation parity

bbingz:
# Recognize renamed/forked agent binaries + SSH confirmForegroundProcess parity

Part of #9414. Stacked on #9415 — review the commits **after** its tip
(`15aa992f14`); this PR adds the following commits on top:

1. `feat(foreground): recognize renamed/forked agent binaries via executable image`
2. `feat(ssh): add confirmForegroundProcess parity for SSH panes`
3. `fix(foreground): preserve command-aware guards in executable-image recognition`
4. `fix(foreground): quote-aware argv0 strip + bounded exe-path cache`

Open as a **draft** until #9415 merges.

## Problem

Foreground-process recognition collapsed to recognized-or-null:
`resolveAgentForegroundProcessFromPs` returned only whitelist matches, so an
unrecognized live process became the shell fallback. Two gaps followed:

- **Renamed/forked agent binaries were never identified.** An argv0-renamed
  real binary, or a native fork installed at a non-standard path (e.g.
  `~/.local/bin/grok`), shows an unrecognized command line, so the agent went
  unrecognized and the pane looked like a bare shell.
- **SSH panes trusted stale foreground reads.** `orca-runtime`
  (`terminalHasShellForegroundProcess`) had no way to *confirm* a foreground
  read on SSH: the SSH provider didn't implement `confirmForegroundProcess`, so
  a possibly-stale relay read of an idle-looking shell (that still fronts a live
  agent) was trusted as-is.

## Evidence model

Recognition still returns a recognized agent name or falls back to the raw
node-pty process name (the existing `string | null` contract is preserved end to
end). The new **executable-image evidence** is an additional, lazy resolution
step layered under the existing command-line recognition: when the command line
doesn't recognize, the process's *real executable image* basename gets a second
chance at recognition.

- New shared module `src/shared/process-executable-recognition.ts`:
  `resolveProcessExecutablePath(pid)` reads the real image
  (Linux `/proc/PID/exe`, macOS `lsof -a -p PID -d txt -Fn`), and
  `recognizeAgentFromExecutablePath(pid, command)` re-runs agent recognition on
  it. Recognition substitutes the resolved image as **argv0 while preserving the
  process's original arguments** and re-runs the command-line recognizer, so the
  command-aware guards still apply — basename-only recognition would have
  mislabeled a renamed `claude --print` (headless one-shot) or a bare `orca`
  (without the `claude-teams` subcommand) as an interactive agent.
- The macOS reader copies `resolveProcessCwd`'s same-uid per-PID `lsof` pattern
  (`-a` ANDs `-p`/`-d` so only the target pid's entries return) and takes the
  **first** `txt` entry as the executable image (mapped dylibs follow it);
  multi-entry output is handled defensively by ignoring later entries.
- Wired into the three foreground resolvers, unrecognized-and-non-shell
  foreground (`+`) candidates only:
  - POSIX `agent-foreground-process.ts` (local/daemon path).
  - Relay `pty-shell-utils.ts` `getRecognizedForegroundDescendant`.
  - Windows `windows-agent-foreground-process.ts` feeds the already-captured
    CIM `ExecutablePath` (no extra probe — it is free) into recognition instead
    of using it only for context matching.

### Reachability limit (direct unknown-non-shell foreground)

Executable-image evidence fires from the descendant scan, so it recognizes a
renamed binary when the reported foreground is a **shell or interpreter wrapper**
(the common shape) and, on the **local/daemon POSIX** path, when node-pty reports
an unknown non-shell name directly (the daemon's async wrapper refresh always
inspects POSIX candidates). It does **not** fire when node-pty reports the
renamed binary's own name directly on the **SSH relay** (`getForegroundProcessName`
returns a clear non-shell/non-wrapper fallback without a remote process-table
scan — a deliberate SSH cost optimization) or on **single Windows reads**
(`shouldInspectWindowsAgentForeground` skips unknown non-shell names to avoid a
per-read CIM scan). In those cases the pane still reports its raw name and stays
`busy` via #9415's unrecognized-non-shell logic; only the agent *identity* is
not upgraded. Closing that gap would trade the SSH/Windows per-read scan cost the
optimizations deliberately avoid, so it is left as a known limitation.

## Executable-resolution limits (script CLIs)

A node/python **script** fork resolves to the interpreter (`/usr/local/bin/node`,
`/usr/bin/python3`), which is intentionally **not** recognized as an agent — this
never fabricates an engine. Script-based CLIs remain covered by the existing
argv/entrypoint recognition (`agent-process-recognition.ts`,
`node_modules/@openai/codex/…` markers, `python -m …`), which this change does
not touch. Executable-image evidence adds coverage only for **renamed real
binaries and native forks**.

## Cost-guard proof

Executable resolution is **lazy, unrecognized-non-shell-foreground-only, and
per-PID cached** (5s TTL). The cache key is PID-only (process start-time is not
cheaply available cross-platform), so a recycled PID — or the same PID after an
`exec` — **can be briefly mislabeled** (the cached path is re-recognized, not
re-resolved) for at most one TTL window. The short TTL bounds the window;
key-by-generation is a spec-accepted deferred hardening. Crucially, resolution
forks `lsof`/`readlink`, **never `ps`**, so the whole-process-table scan volume
is untouched:

- `agent-foreground-process-ps-scan-volume.test.ts` — green (the guard's codex
  candidate recognizes on the command line, so no `lsof` fires).
- `windows-agent-foreground-process-scan-volume.test.ts` — green (drives the
  descendant reader directly; `ExecutablePath` recognition adds no scan).

Both guards stay green with **zero extra `ps` scans**. Tests also assert `lsof`
does **not** fire when a candidate is recognized by command line, and is **not**
run for suspended (non-foreground) candidates.

## SSH confirmForegroundProcess parity

- Relay `pty.confirmForegroundProcess` handler forces a **fresh post-boundary**
  process-table scan (new `fresh` option threaded through
  `getForegroundProcessName` → `getRecognizedForegroundDescendant`, using
  `getFreshProcessTableSnapshot`; Windows uses `{ fresh, forceProcessScan }`).
- `SshPtyProvider.confirmForegroundProcess` requests it. `orca-runtime` already
  feature-detects the optional controller method, so no runtime change is
  needed — the SSH provider gaining the method is what activates confirmation.

### Relay version-skew story

The relay is a separately-versioned **deployed** binary. Tolerance is
bidirectional and non-breaking:

- **New client / old relay:** an old relay 404s the method
  (JSON-RPC `-32601` "Method not found"). The provider feature-detects this,
  **latches** `confirmForegroundUnsupported`, and degrades to today's trusting
  `getForegroundProcess` read — no error, and later confirms skip the failing
  round-trip.
- **New relay / old client:** the added `getForegroundProcess`/relay handler
  wire contract is unchanged (still `string | null`); the new method is
  additive, so old clients that never call it are unaffected.

Non-`-32601` relay errors are rethrown unchanged.

## Tests

- **`process-executable-recognition.test.ts`** — Linux `/proc` exe recognized
  (both via injected reader and the **default `readlinkSync('/proc/PID/exe')`**
  path with mocked `node:fs`, including null-on-readlink-failure); macOS `lsof`
  executable image recognized (first `txt` entry, dylibs ignored); per-PID cache
  (second call skips the reader); TTL expiry re-reads; the cache stays
  **bounded** — distinct never-reused PIDs are swept once past the soft cap;
  interpreter image (`node`) does not fake an engine; a renamed **headless
  one-shot** (`claude --print`) and a **generic `orca`** (no `claude-teams`) stay
  unrecognized while a genuine `orca claude-teams` is recognized (command-aware
  guards preserved) — **including when argv0 is a quoted path with an internal
  space** (`"C:\Program Files\...\x.exe" --print`), which a naive first-whitespace
  split would have mislabeled; unsupported platform returns null without invoking
  a reader.
- **`agent-foreground-process.test.ts`** — renamed/forked binary recognized via
  executable image; no `lsof` when a candidate is recognized by command line;
  interpreter image does not fabricate an agent; suspended (non-`+`) candidates
  are not probed; renamed **Windows** binary recognized from `ExecutablePath`; a
  generic `orca` `ExecutablePath` without the `claude-teams` subcommand is not
  mislabeled.
- **`pty-shell-utils.test.ts`** — renamed SSH-relay binary recognized via
  executable image; no `lsof` for a command-line-recognized descendant;
  confirmation reads bypass the cached process table with a fresh scan.
- **`pty-handler.test.ts`** — `pty.confirmForegroundProcess` registered; forces a
  fresh post-boundary foreground read; returns null for an unknown pty.
- **`ssh-pty-provider.test.ts`** — confirm requests a fresh relay read; old-relay
  `-32601` degrades to `getForegroundProcess` and latches; non-method-not-found
  errors rethrow.
- Both ps-scan-volume guards green.

## Review fixes (post adversarial review)

An adversarial review confirmed two real defects in the shipped code, both now
fixed with regression tests:

- **Quote-aware argv0 strip.** `recognizeAgentFromExecutableImage` stripped the
  original argv0 by first-whitespace, which broke a quoted path containing an
  internal space and left a dangling quote that swallowed the following tokens —
  burying `--print` (so a headless one-shot was mislabeled as an interactive
  agent) or `claude-teams` (so a genuine Orca agent launch failed recognition).
  Now stripped quote-aware. Verified by executing the exact scenario.
- **Bounded exe-path cache.** The per-PID cache was never evicted, so every
  distinct non-shell foreground command (vim, git, node…) accumulated an entry
  forever in a long-lived daemon. Expired entries are now swept once the map
  crosses a soft cap, keeping it bounded to the active TTL window.

The review also flagged the two **descoped** spec items below as spec gaps. They
remain deliberately deferred (see the scope note); they are a large, coupled,
separable feature, not a defect in what this PR ships. The descope is a scoping
decision for the reviewer to ratify — it is not silently omitted.

## Scope note (deferred to the next PR in the stack)

The `#9414` umbrella also envisioned a **structured `ForegroundProcessResolution`
`{ engine, rawProcessName, isShell }`** carried through the daemon cache and
runtime record and a **neutral "unknown-but-alive" tab-icon presence** in
`use-tab-agent`. Those two are a coupled feature — the structured `rawProcessName`
/ `isShell` fields have no in-tree consumer until the neutral-presence UI exists.
The tab-icon path already carries the raw foreground name to the renderer:
`pty.getForegroundProcess` returns `string | null` (an unrecognized non-shell
name flows straight through — daemon `getForegroundProcess` returns the raw
`fallbackProcess`, it is not dropped), and the renderer tracker runs its own
`recognizeAgentProcess`. So the neutral-presence feature is renderer-scoped, and
threading a structured type through the daemon cache / runtime record buys nothing
until that UI lands (the sidebar/busy/keep-awake path stays frozen per #9415).
The UI change itself extends the tab-icon identity from `TuiAgent | null` across the
pane-foreground store entry, the foreground-agent tracker's publish contract,
`use-tab-agent`, `TerminalTabLeadingIcon`, and the drag overlay. That is a large,
separable UI surface with meaningful regression risk against #9415's just-shipped
busy/keep-awake logic, so it is intentionally deferred to the next stacked PR
rather than bundled here. This PR ships the two changes that produce observable,
well-tested behavior with contained diffs: executable-image recognition and SSH
confirmation parity. The busy predicate + keep-awake stay exactly as #9415
shipped them.

## ELI5

Orca can recognize renamed or forked agent binaries by looking at the real executable image, not only the process name. SSH panes get the same foreground-process confirmation local panes already had.

### DIFF
--- a/src/main/agent-awake-foreground-evidence.ts
+++ b/src/main/agent-awake-foreground-evidence.ts
@@ -1,0 +1,75 @@
+// Why: foreground-agent scans are event-driven (title status transitions,
+// session listing sweeps), not periodic. The TTL only bridges gaps between
+// them; expiry revalidates against the runtime's cached foreground state
+// instead of forcing a new process scan.
+export const AGENT_AWAKE_FOREGROUND_AGENT_TTL_MS = 5 * 60 * 1000
+
+type ForegroundAgentEvidence = {
+  // Last time a real scan reported this PTY's recognized foreground agent.
+  reportedAt: number
+  // Refreshed by scans AND by cache revalidation at TTL expiry.
+  observedAt: number
+}
+
+/**
+ * Per-PTY keep-awake evidence for recognized foreground agents (wrapper
+ * claude etc.) that emit no hook statuses. Only recognized agents are ever
+ * reported here — arbitrary processes never grant wake eligibility.
+ */
+export class ForegroundAgentEvidenceLedger {
+  private readonly evidenceByPtyId = new Map<string, ForegroundAgentEvidence>()
+  private readonly now: () => number
+  private readonly revalidate: ((ptyId: string) => boolean) | null
+
+  constructor(args: { now: () => number; revalidate?: (ptyId: string) => boolean }) {
+    this.now = args.now
+    this.revalidate = args.revalidate ?? null
+  }
+
+  /** Returns true when the report changed the ledger (caller should refresh). */
+  report(ptyId: string, agent: string | null): boolean {
+    if (agent === null) {
+      return this.evidenceByPtyId.delete(ptyId)
+    }
+    const now = this.now()
+    this.evidenceByPtyId.set(ptyId, { reportedAt: now, observedAt: now })
+    return true
+  }
+
+  /** Drops expired evidence; counts what remains toward wake eligibility. */
+  pruneAndCount(staleAfterMs: number): number {
+    const now = this.now()
+    for (const [ptyId, evidence] of this.evidenceByPtyId) {
+      if (now - evidence.observedAt <= AGENT_AWAKE_FOREGROUND_AGENT_TTL_MS) {
+        continue
+      }
+      // Why: renew from the runtime's live cache (never a scan), hard-capped by
+      // the same 2h window hook statuses get so a stale cache cannot block
+      // sleep indefinitely. Anything else expires with the TTL.
+      if (now - evidence.reportedAt <= staleAfterMs && this.revalidate?.(ptyId) === true) {
+        evidence.observedAt = now
+      } else {
+        this.evidenceByPtyId.delete(ptyId)
+      }
+    }
+    return this.evidenceByPtyId.size
+  }
+
+  /** Earliest upcoming expiry among live evidence, or null when none. */
+  nextExpiry(staleAfterMs: number): number | null {
+    const now = this.now()
+    let earliest: number | null = null
+    for (const evidence of this.evidenceByPtyId.values()) {
+      // Why: the TTL expiry drives revalidation; the 2h cap is the backstop.
+      const expiry = Math.min(
+        evidence.observedAt + AGENT_AWAKE_FOREGROUND_AGENT_TTL_MS,
+        evidence.reportedAt + staleAfterMs
+      )
+      if (expiry <= now) {
+        continue
+      }
+      earliest = earliest === null ? expiry : Math.min(earliest, expiry)
+    }
+    return earliest
+  }
+}
--- a/src/main/agent-awake-foreground-evidence.ts
+++ b/src/main/agent-awake-foreground-evidence.ts
@@ -1,7 +1,5 @@
-// Why: foreground-agent scans are event-driven (title status transitions,
-// session listing sweeps), not periodic. The TTL only bridges gaps between
-// them; expiry revalidates against the runtime's cached foreground state
-// instead of forcing a new process scan.
+// Why: scans are event-driven, so expiry revalidates cached foreground state
+// without adding process scans.
 export const AGENT_AWAKE_FOREGROUND_AGENT_TTL_MS = 5 * 60 * 1000
 
 type ForegroundAgentEvidence = {
@@ -11,11 +9,7 @@ type ForegroundAgentEvidence = {
   observedAt: number
 }
 
-/**
- * Per-PTY keep-awake evidence for recognized foreground agents (wrapper
- * claude etc.) that emit no hook statuses. Only recognized agents are ever
- * reported here — arbitrary processes never grant wake eligibility.
- */
+/** Per-PTY keep-awake evidence for recognized agents that emit no hook statuses. */
 export class ForegroundAgentEvidenceLedger {
   private readonly evidenceByPtyId = new Map<string, ForegroundAgentEvidence>()
   private readonly now: () => number
@@ -40,9 +34,8 @@ export class ForegroundAgentEvidenceLedger {
   pruneAndCount(staleAfterMs: number): number {
     const now = this.now()
     for (const [ptyId, evidence] of this.evidenceByPtyId) {
-      // Why: the 2h cap is unconditional — revalidation renews observedAt, so a
-      // TTL-only prune would keep a capped entry alive with no timer scheduled
-      // (nextExpiry treats the cap as expired) and block sleep forever.
+      // Why: revalidation renews observedAt, so only an unconditional cap can
+      // prevent evidence from surviving after its final timer.
       if (now - evidence.reportedAt >= staleAfterMs) {
         this.evidenceByPtyId.delete(ptyId)
         continue
--- a/src/main/agent-awake-foreground-evidence.ts
+++ b/src/main/agent-awake-foreground-evidence.ts
@@ -1,6 +1,7 @@
 // Why: scans are event-driven, so expiry revalidates cached foreground state
 // without adding process scans.
 export const AGENT_AWAKE_FOREGROUND_AGENT_TTL_MS = 5 * 60 * 1000
+const FOREGROUND_AGENT_REPORT_REFRESH_QUANTUM_MS = 5_000
 
 type ForegroundAgentEvidence = {
   // Last time a real scan reported this PTY's recognized foreground agent.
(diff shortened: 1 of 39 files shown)


bbingz:
The two deferred scope-note items (structured `ForegroundProcessResolution` + neutral unknown-alive presence) are implemented in #9435, stacked directly on this PR.

bbingz:
**Rebase status: HARD (draft, not pushed)** — conflicts with latest `origin/main`; lower-priority draft stack left for a follow-up session.

bbingz:
## Rebase status (2026-07-22)

Attempted restack onto current `main`; this draft stack still has large multi-file conflicts (sidebar/worktree-status / lifecycle / hook-service) against recent main landings.

Leaving **draft** until base PRs (#8872 close-intent, #9415 unknown-busy) land, then restack. Not closing — work is still intended, just blocked on a clean base.

bbingz:
## Restack onto current `origin/main`

Rebased all **8** commits onto `origin/main` (`96df5f5715`) and force-with-lease pushed to `bbingz/structured-foreground-resolution`.

**New head:** `5825c7893308704ac086e3b1641c5d3f7bbf25b1`

### Conflicts resolved
1. **sidebar rows / worktree-status** — kept main’s launchAgent spinner fallback (#8711) **and** this PR’s TTL-gated `paneForegroundAgent` process-identity attribution; both test suites retained.
2. **agent-foreground-process / Windows / relay** — kept main’s OMP outer-wrapper resolution (#6364) while adding executable-image recognition for renamed/forked agent binaries + command-aware image recognition guards.
3. **SSH confirmForegroundProcess** — kept main’s relay timeout/getAppliedSize plumbing; added old-relay method-not-found latch + `fresh` process-table scan path.

Still **draft**. No origin/main push. Note: may conceptually stack under #9435.

bbingz:
## Restack note

Rebased onto latest `main` (force-with-lease).

- Kept renamed/forked agent binary recognition via executable image
- SSH `pty.confirmForegroundProcess` parity with old-relay `-32601` degrade latch
- Merged with main’s SSH agent-session capabilities / spawn-exit race / getCapabilities (kept both sides)

**Tests:** process-executable-recognition + pty-shell-utils + ssh-pty-provider + pty-handler — **206 passed**.

bbingz:
## Restack + review guide (community, awaiting maintainer)

**Draft.** Restacked onto rebased **#9415** tip (force-with-lease). Stack base **matches** #9415. **+4 unique commits** (9 total including #9415).

### Unique vs main
Main has **no** `process-executable-recognition`:
- Recognize renamed/forked agent binaries via real executable image
- SSH `pty.confirmForegroundProcess` parity + old-relay `-32601` degrade latch
- Command-aware guards + quote-aware argv0 strip + bounded exe-path cache

### Stack / depends
```
#9415 → #9434 (this) → #9435
```
- Prefer review after #9415 (or as stacked series)
- Orthogonal to hooks stack (#9416/#9436)

### Tests (local)
- process-executable-recognition + ssh-pty + pty-shell-utils — **94 passed**

Community author cannot merge.

bbingz:

This draft is long-conflicted (DIRTY) against current `main` and is not being actively rebased. Feel free to reopen or open a fresh PR if the work is still wanted.

PR #9438, opened 2026-07-19
feat(codex): thin symlink-overlay for managed multi-account homes

brennanb2025:
## Summary

Re-architects Codex managed multi-account homes to the **thin symlink overlay** and deletes the config-**copy** that caused drift, for the flag-ON self-contained host lane.

Each per-account home (`codex-accounts/<id>/home`, already the flag-ON launch `CODEX_HOME`) becomes a thin overlay:

- **SYMLINK → real `~/.codex/<entry>`** (dirs → junctions on Windows): `config.toml`, `hooks.json`, `skills/`, `hooks/`, `plugins/`, `plugin-state/`, `profile-v2/`, `themes/`, `prompts/`, `AGENTS.md`. Codex reads shared settings and writes hook/project trust **straight through** to `~/.codex` with zero divergence.
- **REAL, per-account, never symlinked:** `auth.json`, `models_cache.json`, `.credentials.json` (MCP OAuth — keeping this per-account structurally preserves the #8440 fix), `sessions/`, and the sqlite index (separate history per account is correct).

### Proven write-through foundation
A spike verified against real codex 0.144.6: codex canonicalizes the symlinked `config.toml`, then does an atomic temp+rename at the **real** target dir → the real `~/.codex/config.toml` receives the write and the overlay **stays a symlink** (verified for `[hooks.state]`, `[projects]` trust, plain keys, dangling symlink). Because **codex's** writer preserves the symlink but **Orca's** own atomic writers would replace it with a copy, Orca no longer writes hook/config into overlay homes — codex's app-server RPC grant does.

## What changed

| File | Role |
|---|---|
| `codex/codex-home-resource-link.ts` (new) | Shared resource-link/copy ownership primitives, extracted from `codex-home-paths.ts` (max-lines split). |
| `codex/codex-overlay-home-paths.ts` (new) | Overlay builder: symlink `config.toml`+`hooks.json`+resources; on symlink `EPERM` (stock Windows) mark **STALE** + one warned copy (no silent copy, no write-back) + `.orca-overlay-stale.json` marker; `config.toml` self-heal. |
| `codex-accounts/overlay-hook-trust.ts` (new) | Per-overlay RPC grant (`grantManagedCodexHookTrust`, `CODEX_HOME=overlay`, explicit-home key) + removal sweep (`removeCodexManagedHookTrustEntries`, tomlPath=real config, sourcePath=overlay `hooks.json`). |
| `codex-accounts/migrate-copied-home-to-overlay.ts` (new) | Flag-gated, one-shot, recoverable migration of existing copied homes → overlay. |
| `codex-accounts/service.ts` | Overlay replaces the config-copy for self-contained homes; per-overlay grant at add/reauth/select; sweep at removal. **KEEPS `assertOAuthAccountAddAllowed`** (PR-D custom-provider guard). |
| `codex-accounts/runtime-home-service.ts` | Overlay at launch (drops `syncSystemConfigIntoManagedCodexHome`); wires the migration; exposes `getActiveManagedOverlayHomePath()`. |
| `index.ts` | Overlay launch **skips** Orca's per-home hook installer (its atomic write would clobber the symlinks) and grants via codex's app-server instead. |
| `codex/codex-home-paths.ts` | Trimmed to path helpers + the shared-mirror sync. |
| `config/tsconfig.cli.json` | Lists the new resource-link module for the CLI project. |

### Migration behavior (existing copied homes → overlay)
Keeps every real per-account file (`auth.json`, `models_cache.json`, `.credentials.json`, `sessions/`, sqlite). Archives the copied `config.toml` (recoverable through soak), promotes any **user** project-trust the copy accumulated into the real `config.toml` **once** (absent-only — never overrides/resurrects a real-config decision), converts `config.toml`+`hooks.json`+resource copies to symlinks (or warns + marks stale on stock Windows), and re-runs the per-overlay grant. One-shot marker; retries safely if inconclusive.

## Deviations from the plan (please review)

The plan assumed a farther-along state; on this branch several "superseded, delete" modules are **still load-bearing** for the flag-OFF shared mirror + WSL lanes that the plan says to preserve. I verified each and **adapted instead of deleting**, keeping flag-OFF/WSL byte-identical:

- **Did NOT delete** `codex-config-mirror.ts`, `config-settings-promotion.ts`, `codex-config-path-reference-rewrite.ts`, `hook-trust-promotion.ts` — all used by `hook-service.ts` / WSL config seeding / the flag-OFF shared mirror.
- **Did NOT delete** `codex-managed-trust-reconciliation.ts` — it defines `removeCodexManagedHookTrustEntries`, the **removal primitive the new design itself requires**, and is used by `hook-service.ts` + `codex-real-home-hook-install.ts`.
- `copySystemCodexResourceAsOwnedFallback` **kept** (WSL `AGENTS.md` `preferCopy` + shared-mirror Windows fallback still need it). The plan's "delete the silent copy" is realized in the **overlay** lane, which never silently copies.
- **Step 9 (real-codex-binary compat-matrix CI) deferred** — there is no existing codex real-binary CI harness, and a from-scratch job asserting version-dependent write-through can't be authored/validated on macOS without the binary. Added to the re-test checklist instead of shipping an unverifiable job.
- **Account-row staleness UI**: backend marker + `readCodexOverlayStaleEntries()` are implemented; wiring the pill into the renderer is a follow-up (only triggers on stock Windows, can't be exercised on macOS).
- **Shared-hooks.json install for managed-only users**: overlays inherit whatever Orca hook is already in the real `~/.codex/hooks.json`; this PR does not add auto-installing it into the real home for users who never use the system-default lane (larger, opt-in-sensitive; needs e2e). Grant no-ops safely when absent.

## Accepted warts
- **Shared-config `hooks.state` accumulation**: each overlay's explicit-home key is distinct, so all overlays' entries live in the one shared real `config.toml`; removal sweeps the departing overlay's entry (ownership hash/ledger-proven).
- **Windows staleness**: without Developer Mode, file symlinks (`config.toml`, `AGENTS.md`) fall back to a warned copy → the account is marked stale (no live sync, no write-back).

## Verification (macOS)
- `pnpm run typecheck` — pass
- `npx vitest run --config config/vitest.config.ts src/main/codex-accounts/ src/main/codex/` — 38 files, 635 pass / 5 skip
- `pnpm run check:max-lines-ratchet` — OK (no new bypasses; no max-lines disables added)
- oxlint on changed files — clean

New tests (sandboxed temp homes only, never real `~/.codex`): overlay symlink build; write-through preserves the symlink + lands in the sandbox real home; symlink-EPERM warns + marks stale + no write-back; stale marker clears on recovery; migration converts a copied home without losing auth/models_cache/.credentials.json/sessions and preserves user project-trust (absent-only); removal sweeps the overlay `hooks.state`; PR-D guard still blocks OAuth-add under a custom provider; flag-OFF config mirroring stays byte-identical.

## ⚠️ Windows + real-codex re-test needed (not verified here)
- [ ] Real codex 0.144.x: overlay `config.toml` symlink read **and** write-through (approve a hook / trust a project) lands in real `~/.codex/config.toml` and the overlay stays a symlink.
- [ ] Per-overlay hook-trust grant succeeds via codex app-server for a managed account and its `[hooks.state]` entry appears in the shared real config.
- [ ] Account removal sweeps only that overlay's `[hooks.state]` from the shared real config; other accounts' + user's own entries remain.
- [ ] Migration on a real pre-existing copied per-account home: no auth/session loss; user project-trust promoted once; `config.toml` becomes a symlink; archive present.
- [ ] **Windows (stock, no Developer Mode)**: overlay falls back to warned copies, account is marked stale, codex still launches; **and** Windows **with** Developer Mode: junctions/symlinks succeed, no staleness.
- [ ] MCP OAuth (#8440): `.credentials.json` stays per-overlay and is not stranded across add/remove/switch.
- [ ] Add step-9 real-codex compat-matrix CI (write-through + grant + sweep) pinned to representative codex releases.
- [ ] flag-OFF + system-default real-home lanes unchanged in the real app.

Do **not** merge. PR-A (#9173) is superseded by this design and can be closed.


brennanb2025:
Shelving in favor of matching the robust per-account-home model (as Claude does) rather than depending on codex's version-dependent config write-through. Rationale: the severe CODEX_HOME bug class (token wars / lost sessions / config divergence) is already fixed for the single-account majority by Phase-1 real-home routing; the only extra thing this symlink-overlay bought was live-shared settings across *secondary* managed accounts, which Claude itself doesn't do, at the cost of a dependency on codex's internal config write-through (a future codex update could clobber the symlink) plus the Windows file-symlink limitation. Path forward: keep the copy-based per-account homes (#9179) + the Windows login-handle fix (#9430); wait for upstream codex --auth-profile (openai/codex#4432) to do true multi-account cleanly, then delete the per-account-home machinery. The spike here did confirm codex 0.144.6 canonicalizes a symlinked config.toml and writes through to the real dir (symlink preserved) — kept for reference if we revisit before #4432 ships.

PR #9439, opened 2026-07-19
Side by side workspaces (experimental)

xxashxx-svg:
Adds an experimental setting that lets you open more than one project in the main area at once. You can drag a project from the sidebar onto the edge of the workspace to split it, resize with a divider, maximize a pane temporarily, and close a pane from its tab strip. Each pane keeps its own tabs and terminals, nothing is shared between projects.

Everything is behind `experimentalSideBySideWorkspaces` and defaults to off. With the flag off the app behaves exactly as before, there's an e2e test asserting that. No visual change unless the flag is turned on.

I've been daily driving this for a bit with multiple agents running in different projects and it's been solid. Happy to adjust anything about the approach.

Two projects open side by side, each with its own tab strip and terminal

Two projects open at the same time, each pane keeping its own tabs, terminal, and running agent.

Platform and remote notes: developed and tested on Windows 11, the layout code is DOM and percent based with no platform branches, and the new pane controls live inside the existing no-drag strip containers so titlebar dragging is unaffected. SSH worktrees can be split members and restore correctly after restart (hydration validates leaves against the synthesized SSH placeholders). Side panes reuse the existing connection-aware git status refresh and skip disconnected targets. No agent, integration, or git provider specific behavior.

Code review summary from my AI coding agent, per the contributing guide:

- Cross-platform: no hardcoded metaKey or path separators, no platform branches in layout code, controls verified inside no-drag containers. Tested on Windows, logic is platform neutral.
- SSH/remote/local: split membership works for SSH worktrees including restart restore, side pane polling goes through the connection-aware pipeline and stays quiet while disconnected.
- Agents and integrations: panes host the existing per-worktree surfaces unchanged, nothing agent or provider specific.
- Performance: visible panes capped at 6 since each streams PTY output, hidden members follow the existing cold-park and hibernation paths, side pane polling uses the shared slow-poll backoff and keeps its cadence across store churn.
- UI: reuses the tab-group divider styling and existing strip button primitives, flag off renders identical to main which the e2e suite enforces.
- Security: no new IPC surface in this change, state is renderer-local and persisted through additive optional session fields validated on restore.

## ELI5

An experimental setting lets you open more than one project side by side in the main area. Drag a project to the edge to split, resize panes, maximize one temporarily, and keep each project's tabs and terminals separate.


brennanb2025:
Taking a look!

PR #9443, opened 2026-07-19
Paste files into the file explorer

xxashxx-svg:
Adds a Paste item to the file explorer context menus, sitting next to the Copy option that's already there. Copy a file in Finder/Explorer/Nautilus (or with the explorer's own Copy), right click a folder or the empty space below the tree, and paste it in. It shows up on both the row menu and the background menu, and works across folders and across projects.

The item only renders when the OS clipboard actually holds file references, so it doesn't sit there greyed out when you've just copied some text.

File explorer context menu showing New File, New Folder, and Paste

Pasting reuses `importExternalPathsToRuntime`, the same path native drag and drop already uses, so name deconflicting, remote uploads, and the refresh/select behavior afterwards are all identical to dropping the files in. Nothing new had to be invented for remote worktrees.

Reading the clipboard is the only platform specific part:

- Windows: PowerShell `Get-Clipboard -Format FileDropList`, with the console output encoding forced to UTF-8 (without that, non-ASCII filenames come back mojibake'd through the OEM codepage)
- macOS: the `public.file-url` pasteboard type
- Linux: `x-special/gnome-copied-files`, falling back to `text/uri-list`

Everything else is platform neutral.

One perf note worth calling out: the Windows read spawns PowerShell, which is slow enough that the menu would visibly pop the Paste item in a beat after opening. So the result is cached and the menus render from the cache immediately, then reconcile if a fresh read disagrees. The cache refreshes on window focus, right after an in-app copy, and on menu open, with throttling and in-flight dedupe so right click spam can't stack up spawns.

Tests: unit coverage for the clipboard reader across all three platforms (including the encoding fix and malformed payloads) with the child process injected, plus file explorer tests for the menu item appearing and hiding based on clipboard contents.

Code review summary from my AI coding agent, per the contributing guide:

- Cross-platform: the three read paths are separate and each unit tested, no path separators assumed anywhere (paths come back from the OS and go straight into the existing import pipeline), and the web client stubs the read to an empty list so the item just never appears there.
- SSH/remote/local: paste goes through the existing runtime import pipeline with the worktree's connection id, so remote worktrees upload exactly like a drag and drop does. No local-only filesystem assumptions.
- Agents and integrations: no agent or provider specific behavior, this is filesystem and clipboard only.
- Performance: one throttled, deduped clipboard read rather than one per menu open, and the spawn is hidden and bounded. No polling.
- UI: uses the existing DropdownMenuItem primitive and lucide icon set, placed next to Copy where you'd expect it. Strings are localized and the catalog is synced across all five locales.
- Security: the new `clipboard:readFilePaths` IPC is read-only and returns paths the user already put on their own clipboard. It takes no arguments from the renderer, so there's no injection surface, and pasting reuses the import pipeline's existing path handling rather than adding its own.

## ELI5

The file explorer context menu gains Paste next to Copy. After you copy files in the OS or in Orca, right-click a folder or empty space and paste them in. The item only appears when the clipboard actually holds files.


brennanb2025:
Taking a look!

PR #9445, opened 2026-07-19
fix(gitlab): stop refresh button overlapping dialog close X

SilvaViniF:
## Summary

In the GitLab item dialog, the header refresh button overlapped the sheet's built-in close (X) in the top-right corner. This reserves right-side space on the header (px-5 → pl-5 pr-10) so the refresh button clears the close X, and lifts the refresh button (-mt-1.5) so its icon aligns with the close X on the same line. The Sheet's close button is unchanged.

## Screenshots

Before:

After:

image

## Testing

- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Manually verified.

## AI Review Report

> Reviewed the class change for layout correctness: confirmed the close X geometry and that pr-10 (40px) clears it. No JS/logic touched. Cross-platform: Tailwind padding rendered by Chromium/Electron identically on macOS, Linux, Windows — no shortcuts, paths, shell, or platform APIs. Holds at the sheet's responsive widths (w-full, sm:max-w-2xl).

## Security Audit

No security surface.

### DIFF
--- a/src/renderer/src/components/GitLabItemDialog.tsx
+++ b/src/renderer/src/components/GitLabItemDialog.tsx
@@ -24,7 +24,13 @@ import {
 } from 'lucide-react'
 import { toast } from 'sonner'
 import { Button } from '@/components/ui/button'
-import { Sheet, SheetContent, SheetDescription, SheetTitle } from '@/components/ui/sheet'
+import {
+  Sheet,
+  SheetClose,
+  SheetContent,
+  SheetDescription,
+  SheetTitle
+} from '@/components/ui/sheet'
 import { Tabs, TabsContent, TabsList, TabsTrigger } from '@/components/ui/tabs'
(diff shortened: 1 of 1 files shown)


brennanb2025:
Taking a look!

brennanb2025:
Thanks for the fix! I pushed a small follow-up that makes the layout structural: the sheet's built-in absolutely positioned close button is disabled, and refresh + close now render together as same-size header actions using the existing sheet primitive. This removes the overlap without compensating padding or negative margins, preserves the sheet's close behavior and accessibility, and stays stable as titles wrap or the sheet width changes. Verified with the targeted boundary tests (20/20), full typecheck, touched-file lint and React checks, the max-lines ratchet, and localization catalog validation.

nwparker:
thanks @SilvaViniF

PR #9453, opened 2026-07-19
fix(runtime): surface hook-captured providerSession in headless session tabs so web-client Show chat view loads the conversation

### LINKED ISSUE #9452: [Bug]: Web client + paired Remote Orca Server: "Show chat view" on a live Codex session shows the new-chat empty state instead of the existing conversation
## Operating system

Other

## Orca version

Server: **1.4.134** (Linux AppImage running the Remote Orca Server beta on a VPS, shown as "Connected / Compatible" in the pairing UI). Client: web client in a desktop browser.

### Short summary

With a browser connected to a paired Remote Orca Server (reached over a Tailscale address), right-clicking a terminal that runs a live Codex session — launched from Orca via the "+" panel on that remote workspace — and choosing **Show chat view** opens the native chat pane in the empty "Start a chat with Codex" state instead of rendering the existing conversation. The terminal view shows the same session normally.

### What happened?

* The context-menu item appears (agent correctly recognized via launch metadata).
* The chat pane renders the new-conversation empty state: "Start a chat with Codex / Ask Codex to inspect code, explain output, or make a change."
* The existing conversation never loads, indefinitely.

This looks distinct from the transcript-read routing fixed by [#7478](<https://github.com/stablyai/orca/issues/7478>) / [#7367](<https://github.com/stablyai/orca/issues/7367>) (follow-up to stablyai/orca#7404). The remote transcript transport exists, but the chat view never obtains a **session id** to read, so it has nothing to load.

### Suspected code path

* The chat pane resolves its session id exclusively from the pane's live agent-status entry: `src/renderer/src/components/native-chat/native-chat-pane-resolution.ts` — `sessionId: input.agentStatusEntry?.providerSession?.id ?? null`. `launchAgent` only supplies the agent label, which is why the menu item is still offered.
* With `sessionId === null`, the assembled session has no messages and is not loading, so `selectNativeChatViewState` (`src/renderer/src/components/native-chat/native-chat-view-state.ts`) returns `{ kind: 'empty' }` — the "Start a chat with {agent}" surface from `src/shared/native-chat-empty-state.ts`.
* `use-native-chat-live-session.ts` documents the intended recovery ("backfills once the id arrives"), but in this setup the id never arrives.

So the missing link appears to be upstream of the (already remote-routed) transcript transport: propagation of `providerSession` (agent session id + transcript path) from the runtime host's agent-status pipeline to the paired web client for that pane.

### How can we reproduce it?

1. Run a headless Orca runtime on a Linux VPS and pair it as a Remote Orca Server (beta).
2. From a desktop browser, connect to the server's web client (e.g. over Tailscale).
3. Open a remote workspace and launch Codex from the "+" panel; exchange at least one message in the terminal.
4. Right-click the terminal tab → **Show chat view**.

**Expected:** the existing Codex conversation renders in the native chat view (as it does for a local session).

**Actual:** the empty new-chat state, with no transcript, indefinitely.

### Anything else that might help

* stablyai/orca#7404 (closed) covered remote runtime Native Chat reading client-local session data; fixed by [#7367](<https://github.com/stablyai/orca/issues/7367>) / [#7478](<https://github.com/stablyai/orca/issues/7478>). This report seems to be the remaining upstream half: session identity, not transcript reads.
* stablyai/orca#6643 (open) — "Codex agent status missing until first user message" — suggests Codex `providerSession` capture is already fragile locally and may not be mirrored over the runtime transport at all.
* Minor side note observed while gathering info: the CLI has no `--version` flag; it falls through to the help text.

datafuji:
Fixes #9452

## Summary

On a web client paired to a headless Remote Orca Server, right-clicking a live Codex terminal and choosing **Show chat view** opened the empty "Start a chat with Codex" state instead of the existing conversation.

Root cause: the headless session-tabs snapshot never carried hook-captured agent status. The agent-hook-server does capture the Codex `session_id` + `transcript_path` in serve mode, but its listener only targets the Electron main window (absent headless) and the web preload shim stubs the `agentStatus` API. The only channel reaching the web client — the session-tabs snapshot — was built from the persisted workspace session with `launchAgent` (which is why the menu item still appeared) but no `agentStatus`, so `providerSession` never arrived and the chat pane resolved `sessionId === null`.

Fix: in `toMobileSessionTabsResult`, when the publication is headless and the tab has no renderer-published `agentStatus`, resolve the entry from the agent-hook-server snapshot by paneKey (new `findHookAgentStatusEntryForPaneKey`). Renderer-published status stays authoritative; desktop behavior is unchanged. Hardening: hook rows hydrate from `last-status.json` for up to 7 days across restarts, so entries older than `AGENT_STATUS_STALE_AFTER_MS` are demoted to identity-only (`state: 'done'`, keeping `agentType` + `providerSession`) instead of resurfacing a stale `working`/`interactivePrompt` state as live activity.

## Screenshots

No visual change (the fix makes the existing native chat view load the conversation on paired remote setups; no new/changed UI surfaces).

## Testing

- [ ] `pnpm lint` — fails on a pre-existing, unrelated file untouched by this PR (`src/renderer/src/components/skills/skill-freshness-group.tsx`, switch-exhaustiveness). `oxlint` is clean on the changed files.
- [x] `pnpm typecheck`
- [ ] `pnpm test` — full suite not run locally; targeted suites pass: full `orca-runtime.test.ts` (760 tests) plus `mobile-subscribe-integration`, `session-tabs` (RPC), and `web-session-tabs-sync`.
- [ ] `pnpm build` — not run locally.
- [x] Added tests that would catch regressions: headless tabs carry the hook-captured `providerSession`; stale hydrated entries are demoted to identity-only; the fallback is never consulted for non-headless publications.

Not yet verified end-to-end on a real paired headless VPS + web client + live Codex session — happy to iterate if that surfaces anything.

## AI Review Report

Multi-agent review (correctness, testing, maintainability, project-standards, adversarial, api-contract personas) plus an independent adversarial pass through a second model family (Codex CLI), with per-finding independent validation:

- Flagged and fixed (second commit): stale hydrated hook rows (up to 7 days old) could resurface as live `working` status after a headless restart — corroborated by three independent reviewers including the cross-model pass; now demoted to identity-only past `AGENT_STATUS_STALE_AFTER_MS`.
- Flagged and rejected after validation: `providerSessionOnly` placeholder rows being served as real status — the single production wiring site (`src/main/index.ts:1969-1970`) already filters them before the runtime sees the snapshot.
- Wire-contract check: the `AgentStatusIpcPayload -> AgentStatusEntry` reshape strips transport-only fields (`connectionId`, `launchToken`, `providerSessionOnly`); no breaking change for existing mobile/web clients — they simply see a populated `agentStatus` where they previously saw none.
- Cross-platform: explicitly checked for macOS/Linux/Windows — the change is platform-agnostic main-process state plumbing (no shortcuts, labels, path construction, shell, or Electron platform-specific behavior touched). The SSH/remote-host case is the subject of the fix and follows the existing hook-ingest path.

## Security Audit

- No new input handling, command execution, auth, secrets, dependency, or IPC surface. The change only joins already-collected hook-server state onto an existing snapshot payload.
- `providerSession.transcriptPath` (a host filesystem path) now reaches paired clients for headless tabs — the same exposure class the desktop renderer path already ships for non-headless tabs; clients treat it as an opaque locator for the already-remote-routed `nativeChat.readSession/subscribe`. No new leak class.
- Placeholder (`providerSessionOnly`) rows are filtered upstream before the runtime snapshot (validated during review).

## Notes

- The stale-demotion keeps resume identity so native chat still resolves after restarts, while preventing phantom "working" spinners on paired clients.
- Pre-existing gap left untouched: Pi `providerSessionOnly` session-start rows are filtered out of the snapshot, so identity-only Pi rows on headless remain uncovered (predates this PR).
- Side note from the issue, out of scope here: the CLI has no `--version` flag.

https://claude.ai/code/session_019kJhEhHbPtSzK4iUcSaUej

## ELI5

On a web client talking to a headless remote server, "Show chat view" opened an empty Codex start screen instead of the live conversation. Hook-captured session info is now included in headless tab snapshots so chat can load the real transcript.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -16944,6 +16944,69 @@ describe('OrcaRuntimeService', () => {
     expect(result.tabs[0]).not.toHaveProperty('launchAgent')
   })
 
+  it('attaches hook-captured providerSession to headless mobile session tabs', async () => {
+    // Why: a headless Remote Orca Server has no renderer graph publishing
+    // tab.agentStatus, so paired web clients only get title-derived status.
+    // Without the hook-captured providerSession the native chat view resolves
+    // sessionId null and shows the empty new-chat state (#9452).
+    const paneKey = makePaneKey('host-tab', HEADLESS_LEAF_ID)
+    const { runtimeStore } = makeRuntimeStoreWithWorkspaceSession(
+      makeWorkspaceSessionWithHeadlessTerminal({
+        tabsByWorktree: {
+          [TEST_WORKTREE_ID]: [
+            {
+              id: 'host-tab',
+              ptyId: 'persisted-pty',
+              worktreeId: TEST_WORKTREE_ID,
+              title: 'Persisted Terminal',
+              customTitle: null,
+              color: null,
+              sortOrder: 0,
+              createdAt: 1,
+              launchAgent: 'codex'
+            }
+          ]
+        }
+      })
+    )
+    const runtime = new OrcaRuntimeService(runtimeStore as never, undefined, {
+      getAgentStatusSnapshot: () => [
+        {
+          paneKey,
+          state: 'done',
+          prompt: 'fix the bug',
+          agentType: 'codex',
+          connectionId: null,
+          receivedAt: Date.now(),
+          stateStartedAt: Date.now(),
+          tabId: 'host-tab',
+          worktreeId: TEST_WORKTREE_ID,
+          providerSession: {
+            key: 'session_id',
+            id: 'codex-session-1',
+            transcriptPath: '/tmp/rollout.jsonl'
+          }
+        }
+      ]
+    })
+
+    const result = await runtime.listMobileSessionTabs(`id:${TEST_WORKTREE_ID}`)
+
+    const terminalTab = result.tabs.find((tab) => tab.type === 'terminal')
+    expect(terminalTab).toEqual(
+      expect.objectContaining({
+        agentStatus: expect.objectContaining({
+          agentType: 'codex',
+          providerSession: {
+            key: 'session_id',
+            id: 'codex-session-1',
+            transcriptPath: '/tmp/rollout.jsonl'
+          }
+        })
+      })
+    )
+  })
+
   it('keeps renderer-vetted mobile agent status for custom-titled terminals', async () => {
     const runtime = new OrcaRuntimeService(store)
     const leafId = '11111111-1111-4111-8111-111111111111'
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -17007,6 +17007,119 @@ describe('OrcaRuntimeService', () => {
     )
   })
 
+  it('demotes stale hydrated hook status to identity-only in the headless fallback', async () => {
+    // Why: hook rows hydrate from last-status.json for up to 7 days across
+    // restarts; a stale 'working' state must not resurface as live activity,
+    // but providerSession must survive so native chat still resolves.
+    const paneKey = makePaneKey('host-tab', HEADLESS_LEAF_ID)
+    const staleReceivedAt = Date.now() - 60 * 60 * 1000
+    const { runtimeStore } = makeRuntimeStoreWithWorkspaceSession(
+      makeWorkspaceSessionWithHeadlessTerminal({
+        tabsByWorktree: {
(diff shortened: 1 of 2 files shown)


AmethystLiang:
Thank you for the thoughtful contribution and the thorough regression coverage. Closing this as superseded by #10814 and #11636, which implemented and further hardened the headless provider-session projection / Codex session-identity path. #9452 has since been confirmed fixed in current releases.

PR #9455, opened 2026-07-19
fix(relay): honor the remote OpenSSH login shell for Windows SSH terminals

rodboev:
## Summary

Windows SSH terminals now open the shell the remote host actually configured, honoring the OpenSSH `DefaultShell` registry value instead of always falling back to inbox PowerShell.

## ELI5

When you SSH into a Windows machine, it has a setting for which shell to launch. We were ignoring that setting and forcing PowerShell; now we read it and give you the shell the server was told to use.

## Root cause & fix

`resolveWindowsDefaultShell` never consulted the OpenSSH `DefaultShell` value in `HKLM\SOFTWARE\OpenSSH`, so a host configured for a different login shell was overridden by the inbox-PowerShell fallback. The fix adds an exported `readOpenSshDefaultShell()` that runs `reg.exe query`, parses and memoizes the value, and inserts it into the resolution chain between the `$SHELL` check and the PowerShell fallback. On failure, malformed output, or a missing path it returns `''`, so the existing `$SHELL` → WindowsPowerShell → `ComSpec` → `powershell.exe` order is preserved unchanged.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase was clean onto merge-base `7ab601487c`; the PR touches only `src/relay/pty-shell-utils.ts` and its test.

- On branch: **27 tests passed** (1 file).
- Fix reverted to main, test kept: **4 failed | 23 passed** — new export/behavior absent on main (`TypeError: readOpenSshDefaultShell is not a function`).
- Failing assertions when reverted include "reads and memoizes the OpenSSH DefaultShell registry value", "treats malformed OpenSSH DefaultShell output as empty", and "treats reg.exe failures as empty".
- Lint clean: `oxlint src/relay/pty-shell-utils.ts` — no findings.

```
# on branch (fix present)
Test Files  1 passed   |  Tests 27 passed

# fix reverted to origin/main, test kept
Test Files  1 failed   |  Tests 4 failed | 23 passed
  x reads and memoizes the OpenSSH DefaultShell registry value
    TypeError: readOpenSshDefaultShell is not a function
```

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression by construction: the registry lookup is reached only when `$SHELL` is unset/nonexistent, and empty or failed reads short-circuit to the prior fallback chain, so non-SSH and non-Windows paths are unchanged. The code path is gated behind `process.platform === 'win32'`, and the result is memoized in a module-scoped var (injectable in tests). Proven by the passing branch tests and the revert-fails check.

*Credit to @rodboev for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/relay/pty-shell-utils.test.ts
+++ b/src/relay/pty-shell-utils.test.ts
@@ -51,6 +51,7 @@ async function withProcessPlatform<T>(
 }
 
 beforeEach(() => {
+  vi.resetModules()
   execFileMock.mockReset()
   execFileSyncMock.mockReset()
   resetProcessTableSnapshotForTests()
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @rodboev — honor the remote OpenSSH login shell for Windows SSH terminals. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #9461, opened 2026-07-19
fix(codex): preserve managed-account MCP .credentials.json on per-account-home migration (#8440)

brennanb2025:
## Problem (#8440)

Managed Codex accounts moved from a shared runtime-home mirror to per-account homes (each home = that account's `CODEX_HOME`). The legacy→per-account migration `migrateLegacySharedAuthToPerAccountHome` carried **only `auth.json`** from the shared mirror. Codex stores **file-mode MCP OAuth tokens** in `$CODEX_HOME/.credentials.json` (separate from `auth.json`). So an existing managed account that had authed MCP servers got those tokens **stranded on upgrade** → its MCP servers silently need re-auth.

## `.credentials.json` semantics finding

Ground truth from the codex 0.144.5 binary (`rmcp-client/src/oauth.rs`), corroborating #8440:

- Config `mcp_oauth_credentials_store`: `file | keyring | auto | ephemeral`.
- **`file` mode → `$CODEX_HOME/.credentials.json`**, entries keyed by `{type,url,headers}` under an `MCP_OAUTH` map. It is **MCP-server-scoped (keyed by server URL), NOT account-identity-scoped** — the file carries no account identity claim of its own.
- `keyring` mode → global macOS Keychain service `"Codex MCP Credentials"` (not `CODEX_HOME`-scoped, unaffected). macOS default (`auto`→keyring) is why #8440 was cannot-repro on standard macOS; file mode bites Linux/headless/SSH.

**Implication for migration:** because the file has no identity to match on, we **cannot independently identity-match it**. The only proof that a shared mirror's `.credentials.json` belongs to a given account is the **co-located, identity-proven `auth.json`**. So we treat it strictly as a companion of the uniquely-identity-matched `auth.json`.

## Chosen handling + why

Carry `.credentials.json` from the shared mirror into the per-account home **at the same proven-ownership gate** that governs `auth.json`:

- Only after `matches.length === 1 && matches[0].account.id === activeHostAccountId` (the shared mirror is uniquely proven to be the active account's) — so it lands in **exactly one** account's home, **no cross-account leak**. Ambiguous/duplicate identity → nothing carried.
- **Absent source → no-op.**
- **Never clobber** a `.credentials.json` already present in the per-account home (a newer store the account authed in its own home going forward).
- **Atomic + 0600**, mirroring the `auth.json` write, so a crash can't leave a partial MCP store or widen perms on sensitive tokens.
- **Idempotent**: whole migration is marker-gated (runs once); the absent-dest guard makes the carry itself idempotent even across the function's pre-marker early-return retries.
- No symlink/live-share machinery — plain per-account-file copy.

**Going forward:** new MCP auth for a managed account already lands in its per-account home, since that home *is* `CODEX_HOME`. Covered by the "already-current" test (store carried without needing an auth rewrite).

## Files changed

- `src/main/codex-accounts/legacy-shared-auth-migration.ts` — add `homePath` to the trusted-account record; `migrateSharedMcpCredentials()` helper; invoke it at the proven-ownership gate before auth freshness branching.
- `src/main/codex-accounts/legacy-shared-auth-migration.test.ts` — 5 new sandboxed tests: carry into proven home (0600); carry when auth already-current; absent-source no-op; newer per-account file not clobbered; no leak to non-matching (ambiguous) account.

## Test evidence (sandboxed temp homes only — never touches real `~/.codex`)

- `npx vitest run --config config/vitest.config.ts src/main/codex-accounts/legacy-shared-auth-migration.test.ts` → **11 passed** (6 existing + 5 new).
- Full `src/main/codex-accounts/` suite → **172 passed** (8 files).
- `pnpm run check:max-lines-ratchet` → OK.
- Node typecheck of `config/tsconfig.node.json` → clean (exit 0). Note: run with TS 5.9.2 via npx (repo pins `typescript@^7`, native tsc unavailable in this worktree); change is simple typed additions.

## Landing note

This also needs porting to **PR-E #9179** (the per-account `CODEX_HOME` change) for landing — same migration seam and stranding risk.

### DIFF
--- a/config/scripts/codex-primary-home-tripwire.mjs
+++ b/config/scripts/codex-primary-home-tripwire.mjs
@@ -86,6 +86,32 @@ export function describeCodexHomeChange(before, after) {
   }
 }
 
+// Why: with the real-home flag ON, system-default spawn sites deliberately
+// delete CODEX_HOME so native codex resolves the user's real ~/.codex — and on
+// Windows the binary ignores the USERPROFILE sandbox, so its own volatile
+// runtime churn (root sqlite/WAL/SHM, tmp/, log/) is designed behavior no env
+// sandbox can contain. Everything else — auth.json, config.toml,
+// .credentials.json, hooks.json, sessions/, any unknown path — stays a hard
+// containment violation.
+const DESIGNED_SYSTEM_DEFAULT_VOLATILE_PATTERN =
+  /^(?:[^\\/]+\.sqlite(?:-wal|-shm|-journal)?|tmp(?:[\\/].*)?|log(?:[\\/].*)?)$/i
+
+export function classifyCodexHomeTripwireEvent(event) {
+  if (event.scanError || !Array.isArray(event.changedPaths)) {
+    return 'violation'
+  }
+  // Why: the root directory's own mtime churns whenever codex adds/removes a
+  // direct child (e.g. a WAL file); it carries no information beyond the
+  // substantive paths, so it never decides the classification by itself.
+  const substantive = event.changedPaths.filter((entryPath) => entryPath !== '.')
+  if (substantive.length === 0) {
+    return 'designed-system-default'
+  }
+  return substantive.every((entryPath) => DESIGNED_SYSTEM_DEFAULT_VOLATILE_PATTERN.test(entryPath))
+    ? 'designed-system-default'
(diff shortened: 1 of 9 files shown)


brennanb2025:

PR #9463, opened 2026-07-19
Clear a removed remote host's leftover repos and worktrees from the sidebar

### LINKED ISSUE #8881: [Bug]: Sidebar duplicates worktrees when re-paired hosts leave stale runtime identities in the client store
**Operating system:** macOS
**Orca version:** v1.4.141 (root-cause references verified against `main` @ 190ed7d4e)

### Short summary

On a remote client, the same physical worktrees can render 2–3× in the sidebar (e.g. 3 real checkouts shown as 6 rows under one project header). Inspecting an affected client's store shows the same physical host present under **three** `runtime:<env>` identity branches — each re-pairing of the host mints a new environment id — with the same worktrees repeated under 2–3 of them, plus an orphaned repo-id pair where two repoIds resolve to the same checkout path.

### Root cause

- Worktree rows are keyed `${repoId}::${path}` (`src/shared/worktree-id.ts:15-28`), and every dedup in the render pipeline uses that id (`visible-worktrees.ts:199-239`; `worktree-list-groups.ts:600-613`) — so the same physical checkout under a second repoId/host identity never collapses.
- `mergeWorktreesForHost` replaces rows only for the refreshing host and unconditionally keeps sibling-host rows (`src/renderer/src/store/slices/worktrees.ts:457-492`, "host-scoped refreshes should replace that host in place"). Refresh loops iterate only current repo rows (`worktrees.ts:2471-2536`), so rows under a superseded identity are never refreshed *or* purged.
- `getProjectGroupingForRepo` maps both repoIds to the same `project:` header (`worktree-list-groups.ts:238-267`), so both row sets append under one project — 6 duplicate-looking rows while disk/Git/discovery all report 3.
- SSH already has a readoption reconcile for this class (`readopted-ssh-worktree-rows.ts:32-52`), but nothing equivalent exists for runtime re-pairs; and `c12ade54d` recently deduped mobile *host cards* by pinned public key (STA-1840) — the same durable-physical-identity approach seems needed one level down, for runtime environments / worktree rows.

### How can we reproduce it

1. Pair a client with a host; let it mirror some worktrees.
2. Re-pair the same host (new pairing → new runtime environment id) one or more times.
3. The client store now holds the old `runtime:<env>` branches alongside the new one; the sidebar renders duplicates of the same physical worktrees.

### Proposed direction (no PR yet — design input wanted)

Two plausible layers; happy to implement whichever maintainers prefer:

1. **Reconcile at pairing time:** when a runtime environment resolves to the same physical host as an existing one (pinned public key / relay endpoint, as in `c12ade54d`), retire or readopt the superseded environment's rows — the runtime analogue of `reconcileReadoptedSshRepoRows`.
2. **Dedup at merge/render time:** collapse rows whose normalized path + physical host identity match, preferring the freshest host identity, in `mergeWorktreesForHost` or a post-merge normalization.

Option 1 fixes the store (no zombie identities accumulating); option 2 is a smaller render-side mitigation.

brennanb2025:
## What changes

When a remote runtime host is removed from the saved list (e.g. after re-pairing mints a new `runtime:<envId>` identity and the superseded one is removed), Orca now retires the repos, project host setups, and worktree rows that identity owned in the client store. Previously those lingered for the rest of the session, so the same physical checkout rendered two or three times under one project header in the sidebar.

**What does NOT change:** two runtime identities for the same physical host that are *both still saved* still duplicate until one is removed — linking them at pairing time (by pinned public key) is a separate follow-up. This PR removes the state left behind by identities the user has already removed.

Fixes #8881.

## Design approach

`setRuntimeEnvironments` is the renderer's single "saved environment list changed" entry point (boot hydration, the Settings runtime pane after add/remove/re-pair, the Add Remote Host dialog, the SSH status dropdown). It already retires per-environment status, detected-agent caches, and mirrored SSH state for removed environments — but never touched repos/setups/worktree rows. This adds them to that cascade:

- A new `purgeStaleRuntimeHostState` action (worktrees slice) drops repos, project host setups, `worktreesByRepo`/`detectedWorktreesByRepo` rows, and the active/filter repo pointers owned by a removed runtime identity, routed through the existing `buildWorktreePurgeState` cascade so tab/PTY/browser/editor/agent maps can't leak.
- It is **scoped to the removal diff** (env ids that just left the saved list), not an absolute keep-set. This is load-bearing: a serving / self-paired instance persists client-minted `runtime:<envId>` stamps into its own local repos, whose env id was never in this instance's saved list — an absolute keep-set would tear those real local repos down on every boot. The removal diff can never include them.
- A tombstone set (`removedRuntimeEnvironmentIds`) guards the three repo-catalog merges so an in-flight catalog fetch for a just-removed env can't re-add the purged repos. Keying on a tombstone (not "absent from the in-memory saved list") is required so a runtime env merely *not yet hydrated* at boot still merges normally.

## Design review

Ran an adversarial design-review loop before implementation. It reshaped the core decision from an absolute keep-set to removal-diff scoping (the serving-instance case above) and added the in-flight catalog-merge guard and the repo-scoped-UI pointer cleanup. No unresolved design findings.

## Implementation / deviations

Implemented as designed. One deviation surfaced in review and was fixed (see below): the catalog-merge guard's first cut keyed on "absent from `runtimeEnvironments`", which also matches a not-yet-hydrated env and would have dropped legitimate runtime repos at boot; it now keys on the tombstone set.

## Completeness verification

Full pass: the purge is wired into the production removal path (Settings Remove → `runtimeEnvironments.remove` → `loadEnvironments` → `setRuntimeEnvironments`), the removal diff is computed from the pre-replacement in-memory list, all promised state is filtered, and no incident attribution is over-claimed in code or comments.

## Headline behavior status

**Implemented.** Removing a runtime host purges its leftover repos/worktree rows from the sidebar in the running app (verified live, screenshots below).

## Broad app consistency

Source-of-truth behavior ("environment removed ⇒ its client-mirrored state is retired") already existed for runtime status, detected agents, and mirrored SSH state; this extends it to repos/setups/worktree rows. **Known remaining surface:** runtime-hosted *folder workspaces* and *project groups* use the same per-host merge and can still duplicate on re-pair — deferred as a non-goal (different row type + removal cascade). When that follow-up lands it must also guard the parallel project-group / folder-workspace catalog merges.

## Risks, ambiguities, non-goals

- Non-goal: pairing-time physical-host reconcile for two *still-saved* identities (needs the pinned-public-key signal; follow-up).
- Non-goal: row migration on re-pair (SSH readoption renames rows to keep tabs; runtime re-pair has no linkage signal here, so superseded-identity rows are retired — their transport is already gone).
- Non-goal: deleting stale on-disk workspace-session partition files (boot already leaves them unread on a pure client; this stops new stale writes). 
- Non-goal: folder-workspace / project-group retention (above).
- Incident attribution: per the reporter's follow-up, this fixes the code-verified duplication hazard; it does not claim to explain the specific incident's 3→6 rendering (pending a controlled re-pair repro).

## perf

Audited before code review. No-op by default (empty removal diff → no `set()`), all added work bounded by repo/env/row counts already iterated elsewhere, reference identity preserved for unchanged slices, `sortEpoch` bumped only when rows change, `buildWorktreePurgeState` invoked only when rows are actually removed. No P0–P2 findings.

## Code review

Two independent reviewers. One found a P1: the catalog-merge guard was over-broad (skipped legitimate runtime-catalog merges during boot before the saved list hydrated, dropping remote repos — proven by ~14 net-new regressions in neighboring repo suites). Fixed with the tombstone approach and re-reviewed clean; the previously-regressing suites (`repos-all-hosts`, `repos-all-hosts-generation`, `repos-runtime-project-groups`, `repos`) all pass again. The other reviewer returned no blockers.

## Tests / static checks

- 6 new test files (helpers, action, `setRuntimeEnvironments` wiring, catalog-merge tombstone guard incl. the boot-regression case, session-partition split): green.
- Neighboring repo suites reviewer A flagged: green.
- Full typecheck (node + cli + web) and oxlint/oxfmt: clean.

## Manual validation

Live Electron A/B on a fresh profile: injected a stale `runtime:env-a` identity duplicating three physical checkouts (sidebar shows 6 rows under one header), then removed the environment — the duplicates purge and only the real rows remain, with no orphaned/"Unknown" groups. Before/after screenshots attached in a PR comment.

## Post-PR stabilization

Pending CodeRabbit + CI after open.

### DIFF
--- a/src/renderer/src/store/slices/purge-stale-runtime-host-state.test.ts
+++ b/src/renderer/src/store/slices/purge-stale-runtime-host-state.test.ts
@@ -1,0 +1,250 @@
+/**
+ * Action coverage for `purgeStaleRuntimeHostState` (#8881).
+ *
+ * Removing a runtime identity must retire the repos/setups/rows/detected rows and
+ * cascaded tab state it owned, while leaving local, ssh, and — the P1 regression —
+ * a *never-saved* runtime-stamped LOCAL repo (a serving instance's own row)
+ * untouched. The action is scoped to the removal diff, so it must no-op (no
+ * sortEpoch bump) for an empty diff or an env that matches nothing.
+ */
+import { describe, it, expect, vi } from 'vitest'
+import { toRuntimeExecutionHostId } from '../../../../shared/execution-host'
+import type { ProjectHostSetup, DetectedWorktreeListResult } from '../../../../shared/types'
+
+vi.mock('sonner', () => ({
+  toast: { info: vi.fn(), success: vi.fn(), error: vi.fn(), warning: vi.fn() }
+}))
+
+vi.mock('@/components/terminal-pane/pty-dispatcher', () => ({
+  restorePtyDataHandlersAfterFailedShutdown: vi.fn(),
+  unregisterPtyDataHandlers: vi.fn()
+}))
+
+// @ts-expect-error -- minimal window.api stub for the store under test
+globalThis.window = { api: {} }
+
+import { createTestStore, seedStore, makeWorktree, makeTab, TEST_REPO } from './store-test-helpers'
+
+const RUNTIME_A = toRuntimeExecutionHostId('env-a')
+const RUNTIME_NEVER_SAVED = toRuntimeExecutionHostId('env-serving-client')
+
+function setup(overrides: Partial<ProjectHostSetup> & { id: string }): ProjectHostSetup {
+  return {
+    projectId: 'p',
+    hostId: 'local',
+    repoId: '',
+    path: '/tmp',
+    displayName: 'setup',
+    setupState: 'ready',
+    setupMethod: 'existing',
+    createdAt: 0,
+    updatedAt: 0,
+    ...overrides
+  } as ProjectHostSetup
+}
+
+function detectedResult(
+  repoId: string,
+  worktrees: DetectedWorktreeListResult['worktrees']
+): DetectedWorktreeListResult {
+  return { repoId, authoritative: true, source: 'git', worktrees }
+}
+
+function detected(
+  id: string,
+  repoId: string,
+  hostId: ReturnType<typeof toRuntimeExecutionHostId> | 'local'
+): DetectedWorktreeListResult['worktrees'][number] {
+  return {
+    ...makeWorktree({ id, repoId, hostId }),
+    ownership: 'orca-managed',
+    selectedCheckout: true,
+    visible: true
+  } as DetectedWorktreeListResult['worktrees'][number]
+}
+
+describe('purgeStaleRuntimeHostState', () => {
+  it('purges the removed env and preserves local + never-saved runtime-stamped repos', () => {
+    const store = createTestStore()
+    const WT_A = 'repoA::/wt-a'
+    const TAB_A = 'tab-a'
+    seedStore(store, {
+      // A local repo, a runtime:env-a repo (the removed one), and the P1 regression
+      // case: a LOCAL repo carrying a runtime stamp whose env was never saved.
+      repos: [
+        TEST_REPO,
+        {
+          id: 'repoA',
+          path: '/repoA',
+          displayName: 'A',
+          badgeColor: '#000',
+          addedAt: 0,
+          executionHostId: RUNTIME_A
+        },
+        {
+          id: 'repoServing',
+          path: '/repoServing',
+          displayName: 'Serving',
+          badgeColor: '#000',
+          addedAt: 0,
+          executionHostId: RUNTIME_NEVER_SAVED
+        }
+      ],
+      projectHostSetups: [
+        setup({ id: 's-local', repoId: 'repo1', hostId: 'local' }),
+        setup({ id: 's-a', repoId: 'repoA', hostId: RUNTIME_A }),
+        setup({ id: 's-a-repoless', repoId: '', hostId: RUNTIME_A }),
+        setup({ id: 's-serving', repoId: 'repoServing', hostId: RUNTIME_NEVER_SAVED })
+      ],
+      worktreesByRepo: {
+        repo1: [makeWorktree({ id: 'repo1::/wt-local', repoId: 'repo1', hostId: 'local' })],
+        repoA: [makeWorktree({ id: WT_A, repoId: 'repoA', path: '/wt-a', hostId: RUNTIME_A })],
+        repoServing: [
+          makeWorktree({
+            id: 'repoServing::/wt-s',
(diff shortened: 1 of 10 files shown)


brennanb2025:
### Manual validation — live Electron A/B (fresh profile)

Reproduced and verified the fix end-to-end in the running dev build. Screenshots captured locally (attach here from these paths):
- Before: `/tmp/orca-8881-clean-before.png`
- After: `/tmp/orca-8881-clean-after.png`

**Setup:** a real git fixture repo (`main`, `feature/alpha`, `feature/beta`) with a saved runtime environment `env-a`, plus an injected stale `runtime:env-a` identity (second repoId + host setup + three re-keyed rows) duplicating the same three physical checkouts.

**Before (duplicates):** the sidebar renders six rows under one `repo` header — `main` (Local Mac) alongside the stale `main` / `feature/alpha` / `feature/beta` under a **Re-paired Host** badge. Store: 2 repos (`local`, `runtime:env-a`), 3 real + 3 stale rows.

**Action:** remove `env-a` from the saved list (`setRuntimeEnvironments([])`) — the production removal path.

**After (clean):** only the real `main` remains; every `Re-paired Host` duplicate is gone, with **no orphaned/"Unknown" group**. Store assertions:
```
savedEnvs: []
tombstones: ["env-a"]
repos: ["local"]           // stale runtime repo purged
orphanRowGroups: []        // no stranded rows
staleRepoGone: true, staleRowsGone: true, staleSetupGone: true
```

The purge fires only on removal (while `env-a` is still saved the duplicates correctly remain), and the tombstone guard prevents an in-flight catalog fetch from re-adding the purged repos — while a not-yet-hydrated runtime env at boot still merges normally (covered by unit tests).

brennanb2025:
Re: CodeRabbit nitpick on extracting the `vi.mock('sonner')` / `vi.mock('.../pty-dispatcher')` boilerplate into a shared `installStoreTestEnvironmentMocks()` helper — not applying, for two reasons:

1. **It wouldn't work as sketched.** Vitest hoists `vi.mock(...)` to the top of each test module at transform time, so the factory must be statically present in the test file. Moving those calls into an imported helper function and calling it at runtime defeats the hoist — the module under test would already be imported with its real dependencies before the helper runs.
2. **It diverges from the established convention.** 27 existing `store/slices/*.test.ts` files inline these exact mocks; there is no helper-wrapping precedent in `store-test-helpers.ts`. Matching the surrounding pattern keeps these files consistent with the suite.

No other CodeRabbit findings to address (single trivial nit).

PR #9469, opened 2026-07-19
Stop attaching stale closed PRs/MRs to default-branch checkouts

### LINKED ISSUE #9171: [Bug]: Wrong PR diffs displayed when checking out the default branch
### Operating system

macOS

### Orca version

v1.4.143

### Details

When checking out the default branch of a repository (`master` in my case), Orca wrongly reports a PR. The issue is that someone in the past opened by mistake a PR with `master` as base branch, it was since closed. 

I would expect Orca not to report any PR when on the default branch.

#### Checks tab
Image

#### Source control tab

Image

brennanb2025:
Fixes #9171.

**What changes:** When the repository's default branch (e.g. `master`/`main`) is checked out, Orca no longer attaches a historical **closed or merged** PR/MR whose head ref happened to be the default branch name — so its wrong diffs and checks no longer appear in Source Control, the Checks tab, or the sidebar badge. **What does not change:** a genuinely **open** PR/MR whose head is the default branch stays visible, explicitly linked reviews are never hidden, and behavior on feature branches is untouched (closed PRs there still show; the merged-at-head carve-out still applies). **How validated:** a flipped regression test driving the real production lookup + real default-branch resolver, per-provider unit tests, and a live A/B against real GitHub remotes that reproduced the bug with the guard disabled and confirmed the fix with it enabled.

## Root cause

`getPRForBranchOutcome` resolved a PR for a checkout purely by head-branch name with `state=all` and no default-branch awareness; the only implicit filter hid merged PRs, so a closed-unmerged PR whose head was the trunk slipped through and attached to the default-branch checkout. The same list-first, any-state lookup shape exists in the GitLab, Bitbucket, Azure DevOps, and Gitea clients.

## Design approach

One invariant, applied at each provider client's branch-lookup choke point: an implicit branch-name match on the repository's default branch never surfaces a non-open review. A new shared module `src/main/source-control/repo-default-branch.ts` resolves the repo's default branch name through the existing `resolveDefaultBaseRefViaExec` (same source of truth as worktree creation) over the transport the surrounding call already uses (local / WSL / SSH), with a bounded 30s TTL cache, probe timeouts, and fail-open on unresolvable. Resolution is lazy — the state check runs first, so open results, linked reviews, and detached HEAD add **zero** git calls. The guard lives in the clients (not the forge layer) because the PR refresh coordinator and several IPC/runtime paths call the clients directly. On GitHub the guard also covers the persisted fallback-number path, so installs already showing the stale PR self-heal on the next refresh, and it overrides the merged-fallback preservation and merged-at-head carve-out on the trunk only.

## Pipeline results (Brennan's PR process)

- **Design review:** 3 iterations + an adversarial design challenge; final round clean on both reviewers. The challenge's forge-layer alternative was rejected on verified evidence (direct client call sites bypass it). Review expanded provider coverage from 2 to all 5 providers.
- **Implementation:** matches the reviewed design; two judgment calls (no local-git fallback when an SSH provider is gone — safer than answering for the wrong repo; null resolutions cached for the short TTL).
- **Completeness verification:** COMPLETE, no blockers; five advisory notes, all triaged (one produced the extra Azure test below).
- **Headline behavior:** **implemented** — all five providers guarded, self-heal covered, negative proof done (guard reverted → 14 tests fail across all providers; restored → green).
- **Perf audit:** one finding, fixed — the default-branch resolver could multiply its per-probe timeout across five fallback probes and wedge the serial PR-refresh drain for 75s locally/under WSL or 150s over SSH; the entire resolution now shares one 15s wall-clock budget, with a deterministic subprocess-count test (1 spawn after budget exhaustion, not 5). Verified non-findings: lazy ordering is real (zero-git assertions in tests), worst-case added subprocess load ≤ ~0.33 spawns/sec under the background budget, cache bounded (512 entries), no startup cost.
- **Code-review loop:** four independent review lanes across two rounds of reviewers — two CLEAN (with revert-checks proving 14 tests bite across all five providers, plus shuffled-seed runs), two no-P0/P1. Actionable residue: a test-parity nit (Azure discard-then-linked-fallback test, added) and the GitLab `locked` leak (reproduced by one lane; fixed in a follow-up commit with shared-module and GitLab client tests). The remaining P2 is the single-slot shadow non-goal below.
- **Merge-confidence validation:** verdict **land**. Mutation probes confirmed every guard test bites and behavior-preservation tests pass with the guard disabled. Live A/B: a public repo with the exact bug shape (default branch `master`, historical closed PR `#523` from it) returned that stale PR through the real production lookup with the guard disabled, and `no-pr` with it enabled; a second public repo with a real open PR from `master` stayed visible in both configurations.

## Risks, ambiguities, and non-goals

- **Interpretation surfaced for review:** the issue repro notes read "no PR on the trunk"; this fix implements "no *stale* PR on the trunk" — an open PR whose head is the default branch (e.g. trunk → release) remains visible. If the stricter reading is wanted, the guard's state check drops to unconditional (one line).
- **GitLab `locked` MRs are hidden on the trunk** (decided in review): `locked` is normally a seconds-long merge transition, but stuck-locked MRs are a known GitLab pathology and would otherwise re-create the bug; three review lanes flagged it and one reproduced the leak. A legitimately merging trunk MR just hides moments earlier than its `merged` state would have.
- **Single-slot shadow (non-goal):** every provider's branch query effectively takes one candidate, so a newer closed PR can shadow an older open one; the guard then hides the shadow and shows nothing instead of the open PR. Not a regression (previously the closed PR's wrong diffs showed and the open PR was equally invisible); fixing it needs an open-state re-query and is left as a follow-up.
- **State-transition chip loss on trunk:** a trunk-implicit PR that closes/merges while visible now disappears rather than flipping to a closed/merged chip.
- **Default-branch rename:** convergence bounded by local `origin/HEAD` (fails open on the new name — no new harm; converges after re-clone or `git remote set-head -a`).

## Validation

- New/updated tests: `src/main/github/default-branch-stale-pr.test.ts` (8 — hide/keep-open/feature-branch/carve-out/fallback-self-heal/preservation-override/fail-open, driving the real resolver via git-exec-layer mocks), `src/main/source-control/repo-default-branch.test.ts` (14 — transports, WSL, SSH, shared timeout budget, TTL cache, scoping, predicate), plus per-provider cases in the GitLab (+5), Bitbucket (+3), Azure DevOps (+3), Gitea (+3) client suites and a narrowed assertion in the GitHub client suite.
- `pnpm exec vitest run --config config/vitest.config.ts` over all touched suites: 58 files / 726 tests green (also green under shuffled test order, two seeds). `pnpm typecheck` clean; oxlint clean; max-lines ratchet clean.
- **Live Electron smoke (macOS):** launched this branch in a real dev Electron window and opened the persisted `main` / `primary` workspace. Source Control showed **Create PR** (not historical review state), and Checks settled to **No pull request found** after a real GitHub refresh. The exact historical closed-trunk fixture remains covered by the production-lookup regression test and the live API A/B above.
- SSH/remote path is unit-tested (provider-exec routing, provider-gone fail-open); not exercised against a live SSH host.

## Live Electron screenshots

Source Control on the default branch shows **Create PR** and normal local changes:

Source Control on main showing Create PR

Checks settles to **No pull request found** for the same `main` / `primary` workspace:

Checks on main showing no pull request found

### DIFF
--- a/src/main/azure-devops/client.test.ts
+++ b/src/main/azure-devops/client.test.ts
@@ -6,12 +6,29 @@ import {
   normalizeAzureDevOpsApiBaseUrl
 } from './client'
 import { _resetAzureDevOpsRepoRefCache } from './repository-ref'
+import { __resetRepoDefaultBranchCacheForTests } from '../source-control/repo-default-branch'
 
 const gitExecFileAsyncMock = vi.hoisted(() => vi.fn())
 
 vi.mock('../git/runner', () => ({
   gitExecFileAsync: gitExecFileAsyncMock
 }))
+
+/** Serve the remote URL plus the #9171 default-branch resolver probes. */
+function primeGitExecWithDefaultBranch(defaultRef = 'refs/remotes/origin/main'): void {
+  gitExecFileAsyncMock.mockImplementation(async (args: string[]) => {
+    if (args[0] === 'remote') {
+      return { stdout: 'https://dev.azure.com/acme/Project/_git/repo\n', stderr: '' }
+    }
+    if (args[0] === 'symbolic-ref' && args.includes('refs/remotes/origin/HEAD')) {
+      return { stdout: `${defaultRef}\n`, stderr: '' }
+    }
+    if (args[0] === 'rev-parse' && args[1] === '--verify' && args.includes(defaultRef)) {
+      return { stdout: 'default-oid\n', stderr: '' }
+    }
+    throw new Error(`unexpected git call: ${args.join(' ')}`)
+  })
+}
 
 const OLD_ENV = process.env
 const OLD_FETCH = globalThis.fetch
@@ -21,6 +38,7 @@ describe('Azure DevOps client', () => {
     process.env = { ...OLD_ENV, ORCA_AZURE_DEVOPS_TOKEN: 'pat-token' }
     gitExecFileAsyncMock.mockReset()
     _resetAzureDevOpsRepoRefCache()
+    __resetRepoDefaultBranchCacheForTests()
   })
 
   afterEach(() => {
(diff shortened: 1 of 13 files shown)


brennanb2025:
Re-verification evidence for #9171 (not committed):

**Regression test — the repro scenario now yields no attached PR on the default branch** (`pnpm exec vitest run --config config/vitest.config.ts src/main/github/default-branch-stale-pr.test.ts`, at PR head):

```
✓ hides a stale CLOSED PR when checked out on the default branch (master)
✓ reports kind="no-pr" from getPRForBranchOutcome on the default branch
✓ keeps a genuinely OPEN PR on the default branch visible
✓ keeps a CLOSED PR on a feature branch visible (behavior preserved)
✓ preserves the merged-at-head carve-out on a feature branch
✓ hides a CLOSED fallback-number PR on the default branch (self-heal for persisted stale PRs)
✓ overrides acceptMergedFallbackPR preservation for a merged fallback PR on the default branch
✓ fails open when the default branch cannot be resolved (stale PR still returned)

Test Files  1 passed (1)
     Tests  8 passed (8)
```

This test file is the flipped version of the committed repro writeup's pinned-bug test (`docs/bug-reproductions/9171.md` on the repro branch): before this fix the same scenario returned the stale closed PR `#7` fixture with `kind: "found"`; the implementation was also negative-proofed (guard reverted → these assertions fail; restored → green).

**Live A/B against real remotes** (read-only API calls, no PRs created): a public GitHub repo with the exact bug shape — default branch `master`, historical closed PR `#523` opened from it — returned that stale PR through the real production lookup with the guard disabled, and `no-pr` with the guard enabled; a second public repo with a genuinely open PR from `master` stayed visible in both configurations.

Full touched-suite sweep at PR head: 58 files / 723 tests green; typecheck and lint clean.

brennanb2025:
Re the CodeRabbit nitpick (in-flight dedup in `getRepoDefaultBranchName`): considered and deliberately not added. The PR refresh coordinator drains serially, so realistic same-key concurrency is 2–3 callers (a renderer-triggered lookup racing the coordinator), the resolution is gated to run only when a non-open implicit candidate appears, and the 30s TTL bounds redundancy to at most one extra probe pair (2 bounded git execs) per repo per window. The added promise-map lifecycle isn't worth that margin; happy to add it if the resolver ever moves onto a hot path.

PR #9472, opened 2026-07-19
Keep lineage children with their parent when dragging to another status lane

### LINKED ISSUE #9083: [Bug]: Status-lane drag of a parent worktree leaves its lineage children behind
### Operating system

macOS

### Orca version

Not specified

#### Summary

When worktrees are linked through a parent/child lineage, dragging the parent worktree to another status lane changes only the parent. Its visible lineage children stay in the previous status.

This is confusing because Orca's drag interaction already treats a visible lineage as one unit in another context: reordering a worktree expands the dragged set to include the visible lineage. The status-lane drop preview can therefore suggest a group operation while the mutation is node-local.

#### Steps to reproduce

1. Create or open two worktrees linked as parent and child.
2. Ensure both are visible in the same worktree list.
3. Drag the parent worktree to a different status lane.
4. Observe the child worktree's status.

#### Actual behavior

Only the directly dragged parent changes status. The child remains in its previous status.

#### Expected product contract

The lineage relation should not necessarily imply that every status change always cascades. Parent/child can represent provenance or structure, while status may legitimately remain node-local.

However, drag preview and drop behavior should use the same scope. A reasonable contract would be:

- A normal status drag changes only the explicitly dragged worktree, and the preview clearly shows only that worktree; or
- If the preview groups the visible lineage, the drop changes that same visible lineage; and
- Orca offers an explicit “move subtree/lineage” action when users intend to update descendants together.

The key requirement is that the scope shown during drag must match the scope mutated after drop.

#### Implementation evidence

The reorder path expands the dragged IDs with `expandDraggedWorktreeIdsForVisibleLineage(...)`, while the status-drop path uses `drag.draggedIds` directly. This appears to create the preview/mutation scope mismatch.

#### Acceptance criteria

- The status drag preview identifies exactly which worktrees will change.
- Dropping updates exactly the worktrees shown in that preview.
- Node-only and lineage/subtree moves are distinguishable when both are supported.
- Behavior is covered for a parent with one or more visible children.

brennanb2025:
Fixes #9083.

**What changes:** Dragging a parent worktree card to a different status lane in the sidebar (grouped by workspace status) now moves its visible lineage children with it — matching what the drag preview already shows. Before, only the parent's status changed: the children were left behind in the old lane and rendered as orphaned top-level cards, silently breaking the visual nesting.

**What does not change:** Pin drops still pin only the directly dragged card; drops that nest a card under a parent still re-parent only the directly dragged cards; the expanded workspace kanban board (flat cards, no nesting) is untouched; hidden/collapsed children do not follow (same "visible lineage" semantics the reorder path has always had); the context-menu "Move to status" single-card action is unchanged.

## Design approach

Both sidebar drag sessions (custom pointer drag and native HTML5 drag) already precompute a lineage-expanded dragged set (`reorderDraggedIds`, via `expandDraggedWorktreeIdsForVisibleLineage`) that the drag preview and the reorder commit use. The bug was that every status-lane / board commit path read the raw dragged ids instead. The fix swaps those commit sites (and their hover-preview computations) to the same expanded set — no new helpers, no new state. Two session-less commit paths expand the ids at commit time through the same helper: the status-drop fallback and the capture-phase document-drop committer (`useWorkspaceStatusDocumentDrop`), which is the primary native-drop path in packaged Electron. That wrapper lives at the `WorktreeList` call site only, so the `WorkspaceKanbanDrawer`'s flat board keeps committing raw ids.

## Review and verification pipeline (Brennan's PR process)

- **Design review:** delegated review loop over the design doc, 2 full rounds + 1 focused verification; final verdict clean. It added two findings folded into the fix: the packaged-Electron document-drop committer and the board drop-indicator predicate, plus a dual-case fix for the status-drop fallback (it also fires when a session exists but no drop preview could be computed).
- **Implementation:** all 7 design items in `src/renderer/src/components/sidebar/WorktreeList.tsx` (+36/−20) plus a new regression test. One deviation from the validation plan, allowed by the design doc: the pointer-path DOM drag simulation was skipped in favor of driving the real native document-drop committer (happy-dom lacks a usable `elementFromPoint`).
- **Completeness verification:** independent read-only pass — verdict *complete-with-accepted-deviations*; all 7 items implemented as specified, intentionally-raw sites confirmed intact, no unapproved narrowing.
- **Headline behavior:** **implemented** (not partial). Live-verified in the running dev app: parent dragged from "In progress" to "Todo" carries its visible child; child stays nested under the parent in the new lane.
- **Broad app consistency:** source of truth is the drag-preview scope (parent + visible children move as one unit). Sidebar lane headers, in-lane insertion drops, sidebar→collapsed-board drops, both capture-phase document committers, and the board drop indicator now all use that scope. Deliberately preserved mismatches: pin drops stay parent-only (pre-existing, cosmetic preview mismatch remains there), expanded kanban board stays flat, collapsed children stay behind (same as reorder).
- **Performance audit:** clean. The expanded set is precomputed at drag start; per-frame paths only read it. The two new expansion calls are one-shot at commit time. The new `useCallback` wrapper's deps track worktree-list changes, not drag frames.
- **Code-review loop:** 1 round, two independent reviewers, both clean. One recorded non-actionable `P2` residual: the new test guards the document-drop committer (the packaged-app path); the session-path commit sites are guarded against systemic reverts but not a single-site hand-revert — an isolated pointer-drag DOM test would need a brittle new harness, judged net-negative for now.
- **Merge-confidence validation:** verdict **land**. Live pointer-drag scenarios in the dev app: core parent+child lane move (fixed), childless card moves alone, in-lane reorder with children following still works; store-level assertions plus full-window screenshots for each; console clean.

## Tests and checks run

- New: `WorktreeList.status-lane-lineage-drop.test.tsx` — renders the real `WorktreeList` grouped by workspace status (parent + visible lineage child, bystander in another lane) and drives a native drop carrying only the parent id onto a lane header; asserts both parent and child commit to the target status, other lanes untouched, and a childless drop commits only itself. 3/3 pass.
- Regression suites (worktree-manual-order, lineage card tests, kanban sidebar drop, WorkspaceKanbanDrawer task-status sync, and neighbors): all green.
- Typecheck (`tsc` web config) clean; oxlint/oxfmt clean.

## Manual validation evidence

Screenshots (bug repro pre-fix, and post-fix drags) are attached in a PR comment below — not committed to the repo.

## Risks / accepted gaps

- The packaged-build document-drop committer is exercised by the new unit test (real DOM drop through the real hook), but the live drag re-verify used the dev app's pointer path; no packaged build was launched for this PR.
- `handleWorktreeDrop`'s row-level status commit is unreachable today (the capture-phase document handler pre-empts it); it was swapped defensively for consistency.
- The board drop-indicator swap is a no-op under workspace-status grouping and only changes indicator behavior when nested children hold mixed statuses under other groupings — reasoned, not live-tested.
- SSH: not applicable — renderer-only commit scope; status updates flow through the existing host-aware batched meta update, just with more ids. Mobile: no sidebar drag surface.

## Senior review live re-validation (2026-07-19)

Re-ran the real sidebar pointer gesture in an isolated Electron dev profile on this PR branch. The drag emitted one batched update containing exactly `parent-tree` and its visible `child-tree`; both moved from **In progress** to **Todo**, the child remained visibly nested, and the unrelated `bystander` was untouched.

### Before

Before: parent and visible child nested in In progress

### Mid-drag

Mid-drag: lineage drag preview over Todo

### After

After: parent and child remain nested together in Todo

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeList.status-lane-lineage-drop.test.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeList.status-lane-lineage-drop.test.tsx
@@ -1,0 +1,438 @@
+// @vitest-environment happy-dom
+
+import { act, type ReactNode } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
+import type {
+  Repo,
+  WorkspaceStatus,
+  Worktree,
+  WorktreeCardProperty,
+  WorktreeLineage,
+  WorktreeMeta
+} from '../../../../shared/types'
+import { DEFAULT_WORKSPACE_STATUSES } from '../../../../shared/workspace-status-defaults'
+import {
+  WORKSPACE_STATUS_DRAG_IDS_TYPE,
+  WORKSPACE_STATUS_DRAG_TYPE
+} from './workspace-status-drag-data'
+
+globalThis.IS_REACT_ACT_ENVIRONMENT = true
+
+const mockStore = vi.hoisted(() => ({
+  state: {} as Record<string, unknown>,
+  activateWorktreeFromSidebar: vi.fn(),
+  openModal: vi.fn(),
+  updateWorktreeMeta: vi.fn(),
+  updateWorktreesMeta: vi.fn(),
+  setSortBy: vi.fn(),
+  fetchHostedReviewForBranch: vi.fn(),
+  fetchIssue: vi.fn(),
+  fetchLinearIssue: vi.fn(),
+  openTaskPage: vi.fn()
+}))
+
+type WorktreeListComponent = React.ComponentType<{
+  scrollOffsetRef: React.RefObject<number>
+  scrollAnchorRef: React.RefObject<unknown>
+}>
+
+let WorktreeList: WorktreeListComponent
+
+function makeFolderWorkspacePathStatusMockState(): Record<string, unknown> {
+  return {
+    fetchFolderWorkspacePathStatus: vi.fn(),
+    folderWorkspaces: [],
+    folderWorkspacePathStatuses: {},
+    getFolderWorkspacePathStatusCacheKey: (request: unknown) => JSON.stringify(request),
+    getFreshFolderWorkspacePathStatus: () => null
+  }
+}
+
+vi.mock('@/store', () => {
+  const useAppStore = ((selector: (state: Record<string, unknown>) => unknown) =>
+    selector(mockStore.state)) as ((
+    selector: (state: Record<string, unknown>) => unknown
+  ) => unknown) & {
+    getState: () => Record<string, unknown>
+  }
+  useAppStore.getState = () => mockStore.state
+  return { useAppStore }
+})
+
+vi.mock('@tanstack/react-virtual', () => ({
+  defaultRangeExtractor: ({ startIndex, endIndex }: { startIndex: number; endIndex: number }) =>
+    Array.from({ length: endIndex - startIndex + 1 }, (_, index) => startIndex + index),
+  measureElement: () => 32,
+  useVirtualizer: ({ count }: { count: number }) => ({
+    elementsCache: new Map(),
+    getTotalSize: () => count * 96,
+    getVirtualItems: () =>
+      Array.from({ length: count }, (_, index) => ({
+        index,
+        key: `row-${index}`,
+        start: index * 96
+      })),
+    measureElement: vi.fn(),
+    scrollToIndex: vi.fn()
+  })
+}))
+
+vi.mock('@/hooks/useVirtualizedScrollAnchor', () => ({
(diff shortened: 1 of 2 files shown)


brennanb2025:
## Screenshot evidence (live Electron validation — not committed to the repo)

### Bug reproduction (before the fix)
Parent `parent-tree` with visible nested child `child-tree` ("1 child" badge) in **In progress**; bystander in **Todo**:

before drag, pre-fix

After dragging `parent-tree` to the **Todo** lane header (pre-fix): parent moved alone, `child-tree` left orphaned as a top-level card in **In progress** — nesting broken:

bug: child left behind

### With the fix
Same drag re-run after the fix — `parent-tree` **and** `child-tree` both land in **Todo**, child still nested under the parent:

fixed: child follows parent

### Merge-confidence validation pass (independent re-run)
Before / mid-drag (drag ghost + board preview) / after — parent+child carried together:

validation before
validation mid-drag
validation after

Regressions checked: a childless card (`bystander`) still moves alone between lanes, and in-lane reorder (no status change) still carries children:

childless card moves alone
in-lane reorder unaffected

PR #9474, opened 2026-07-19
Stop AI Vault cold scans from re-reading the entire transcript corpus on every launch

### LINKED ISSUE #9210: Cold AI Vault scan reads 6.7 GB for ~109 s with zero cache reuse; RPC/UI stalls observed around the same incident
## Environment

- Orca 1.4.142-rc.7 (macOS, Apple Silicon MacBook Pro), acting as orchestrator over SSH worktrees on a remote Mac mini (relay 0.1.0+2a9c9eab06ca)
- Large local agent-transcript corpus: ~4 GB of Codex rollouts under `~/.codex/sessions` (463 files) plus Claude/Gemini/OpenCode stores

## Summary

On one fresh launch, `aiVault.scan` took **109.15 s** and reported **6.75 GB read with no cache reuse**. RPC and UI stalls were observed in the same incident family, so the scan is a leading suspect, but the causal link to the intermittent RPC wedges and an hours-long GUI freeze on this machine has **not** been established. Reporting the measured scan cost on its own merits, plus the surrounding observations.

## Measured evidence

Startup scan captured in `main.trace.ndjson` right after a fresh app launch:

```
aiVault.scan  durationMs=109150
candidates=1364, fullParses=702, bytesRead=6749311278, reused=0, incremental=0, issues=0
```

Source interpretation (verified in source at v1.4.138-rc.7; these files are unchanged through the locally available v1.4.142-rc.3 tag; the exact v1.4.142-rc.7 source was not available in this clone):

- The scanner runs in Electron's main process (`src/main/ai-vault/session-scanner.ts`, `withSpan('aiVault.scan', ...)`), parses stream chunks synchronously, and processes up to eight candidates per batch.
- The parse cache (`session-scanner-parse-cache.ts`) is a module-level Map, i.e. process-local — a fresh launch cannot reuse it. The captured launch reported `reused=0, incremental=0`.
- The span's own comment says it exists to make scan cost visible for "STA-1278-style" one-core-pegged investigations, i.e. this cost is meant to be measured — here is a real-corpus measurement.

## Stalls observed in the same incident (correlation only)

Across Jul 16–17, `terminal list/create` calls via the relay stalled for roughly 3 to >4 minutes and then self-healed; one `close` call stalled for >10 minutes and eventually returned exit 0 after a retry had already closed the terminal; one `create` attempt produced no window and succeeded when re-issued. Short verbs kept responding during the same windows.

## Limitations / unresolved observation

During an hours-long GUI freeze on the same machine (desktop + paired mobile frozen on stale state while request/response RPC still answered), the trace tail contained only once-per-minute `renderer_memory` breadcrumbs. Because `aiVault.scan` is recorded on completion, this trace does not establish whether a scan was in flight. Recovery followed a combined full quit, daemon cleanup, artifact cleanup, and relaunch, so that intervention does not isolate the cause of the freeze.

## Proposed reproduction (step 5 not yet performed on this machine)

1. Fully quit Orca.
2. Retain a representative transcript corpus (here: ~4 GB Codex rollouts, 463 files, plus other stores).
3. Enable the local trace and relaunch.
4. Continuously timestamp `terminal list` latency and UI updates until `aiVault.scan` completes.
5. Repeat on the same build after temporarily reducing/moving the transcript corpus, and compare.

Expected: RPC/UI remain responsive while scanning. Actual in the captured large-corpus launch: `aiVault.scan` completed in 109.15 s with 6.75 GB read; exact stall/scan overlap still needs a second instrumented capture.

## Suggestions

- Move transcript parsing off the main process (worker threads or a utility process), or yield between parse batches.
- Persist the parse cache across restarts so a cold start does not re-read the entire corpus.
- Consider a watchdog/timeout around the scan.

Potentially related user-facing report: #8864 ("Agent history lookup hang forever") — its underlying mechanism is not established in that report.

Happy to provide the full trace file or run instrumented builds — the corpus that triggers this is real daily usage, not synthetic.

brennanb2025:
Fixes #9210

## What changes

On a fresh launch, the AI Vault (Agent Session History) scan re-parsed the user's entire transcript corpus — the reporter measured **6.75 GB read over 109 s with `reused=0`**, and every launch paid it again. The parse cache already made in-process rescans nearly free, but it was an in-memory Map that died with the process.

This PR persists the cache's reusable portion (`{mtimeMs, sizeBytes, platform, session}` per transcript, ≤4096 entries, resume parser states dropped) to one JSON file under the canonical userData dir (`ai-vault/session-parse-cache.json`, dir `0o700`/file `0o600`). It loads lazily before the first scan and saves via a debounced (1.5 s) atomic temp+rename write after scans that parsed anything. Unchanged files (mtime+size+platform exact match — the same reuse gate as before, no new invalidation logic) become cache hits; files changed while the app was closed pay one full re-parse.

**Measured on this machine (real corpus, dev build, fresh profile):**

| | duration | reused | fullParses | bytesRead |
|---|---|---|---|---|
| Cold launch, before fix (and first-ever launch after) | 3.7 s | 0 | 1039 | 1.79 GB |
| First scan after restart, with fix | **328 ms** | **1036** | 3 | 3.1 MB |

## What does NOT change

- The reporter's RPC-stall / GUI-freeze correlation is **unverified** upstream; this PR targets the measured scan cost only and makes no claim about the freeze.
- Parsing still runs on the Electron main process. The true first launch (no cache file) and the first launch after each app update (cache discarded on `appVersion` mismatch, by design) still pay one full parse. Offloading to a worker/utilityProcess and the issue's watchdog suggestion are deferred follow-ups.
- SSH-host scans (`remote-session-scanner`) don't use this cache and are untouched. Serve-mode/remote-runtime hosts gain persistence on their own main process (init runs in both desktop and `--serve` modes).
- No renderer, preload, shared-type, mobile, or RPC-schema changes; no UI change beyond the panel populating at warm-scan speed.

## Design & review process (Brennan's PR process)

- **Design:** scratch design doc (gitignored, not shipped) chose persistence over offload/result-cache persistence/resume-state serialization; failure mode of every new path is "behave like today's cold scan". Cache file discarded whole on `schemaVersion`/`appVersion` mismatch or any malformed entry.
- **Design review:** delegated loop ran 2 iterations (opus + gpt-5.5 reviewers) — 8 findings fixed in-doc (canonical-userData path precedent, Windows rename semantics, honest cold≈warm bounds, watchdog deferral made explicit, size math), 1 rejected; verification round clean.
- **Implementation vs doc deviations:** one extra test-only export (`flushSessionParseCachePersistForTests`) for deterministic debounce testing; snapshot/seed tests live in the new test file rather than extending the old one.
- **Completeness verification:** independent read-only pass — no blockers; concerns (envelope-only session validation, save-failure injection shape, foreign over-cap files) all subsequently addressed or adjudicated in review.
- **Headline behavior: implemented** (live A/B proven, table above).
- **Broad app consistency:** desktop panel and mobile session-history share this exact scan via one cache module; a persisted hit returns the identical session object an in-process hit returns, so no ordering/label/state changes and no new cross-surface disagreement.
- **Perf audit:** verdict acceptable; budget claims verified ~10× conservative (worst-case payload 7.98 MB, stringify 5.6 ms, parse 7.0 ms at the 4096 cap); audit NITs fixed in-branch (orphaned `.tmp` sweep on load, serialized save chain, debounce-collapse test).
- **Code-review loop:** 2 rounds × 2 independent reviewers. Round 1: 1 Medium (scanner hooks had zero test coverage — proven by reverting the hunk with all tests staying green) + 2 actionable Lows, all fixed with revert-sensitive tests; 1 advisory rejected (below). Round 2: **both reviewers clean** (166/166 tests across shuffled/isolated/fd-limited runs, zero unhandled rejections, all fixes independently re-proven revert-sensitive).

## Risks / accepted trades

- Persisted `session` objects are envelope-validated (per-field types on the entry, whole-file discard on any anomaly) but the session interior is trusted — same trust domain as the transcripts themselves (user-only file under userData); reviewers traced every consumer and found no crash path. Cache-file strings bypass the normalization real transcripts get.
- Quit within ~1.5 s of a scan can lose that save (next launch cold for the delta). Dev builds have a static version, so the `appVersion` discard is inert across dev parser changes (matches the existing in-process staleness trade).
- Multi-instance: atomic rename; on Windows a rename racing an open handle fails and is swallowed (save lost, never a torn file).

## Validation

- 166/166 ai-vault tests (13 new: round-trip proving `reused=1, bytesRead=0` with the transcript deleted; corrupt/schema/appVersion discard; seed-no-clobber; changed-while-closed full parse; zero disk IO uninitialized; failing mkdir and failing rename both swallowed with temp cleanup; debounce collapse; orphan sweep; over-cap newest-tail; scanner-hook integration test proven to fail if either hook is dropped). `pnpm typecheck:node` and oxlint clean.
- Live Electron A/B on this machine's real corpus (evidence and screenshots in the description below).
- Not exercised: the reporter's 6.7 GB corpus, live Windows/WSL runs (code path platform-neutral; per-entry platform gate has existing tests), and the standalone merge-confidence validation pass was compressed into the live A/B above by coordinator instruction — flagging as the one process gap.

## Follow-up review: live restart validation (2026-07-19)

Revalidated after rebasing onto current `main` (including the OpenCode SQLite worker change) with a fresh isolated dev profile and the real local transcript corpus. The panel populated with 500 sessions on both launches; the persisted cache was 1,327,760 bytes with mode `0600`.

| launch | duration | reused | full parses | bytes read |
|---|---:|---:|---:|---:|
| First launch | 2,487 ms | 0 | 520 | 941,154,261 |
| Restart, same profile | **137 ms** | **513** | **7** | **20,144,932** |
| Next steady refresh | 93 ms | 514 | 0 | 22,595 |

The seven restart parses were transcripts that changed while the app was closed; unchanged entries came from disk as intended.

**First launch — populated Agent Session History after creating the cache**

First launch AI Vault panel

**Restart — the same panel restored after disk-backed cache reuse**

Restart AI Vault panel

### DIFF
--- a/src/main/ai-vault/session-parse-cache-persistence.test.ts
+++ b/src/main/ai-vault/session-parse-cache-persistence.test.ts
@@ -1,0 +1,437 @@
+import { existsSync } from 'node:fs'
+import {
+  appendFile,
+  mkdir,
+  mkdtemp,
+  readFile,
+  readdir,
+  rm,
+  stat,
+  utimes,
+  writeFile
+} from 'node:fs/promises'
+import * as fsPromises from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import {
+  ensureSessionParseCacheLoaded,
+  flushSessionParseCachePersistForTests,
+  initSessionParseCachePersistence,
+  resetSessionParseCachePersistenceForTests,
+  scheduleSessionParseCachePersist
+} from './session-parse-cache-persistence'
+import { scanAiVaultSessions } from './session-scanner'
+import {
+  createSessionParseStats,
+  parseAgentSessionFileCached,
+  resetSessionParseCacheForTests,
+  seedSessionParseCache,
+  type PersistedSessionParseCacheEntry,
+  type SessionParseStats
+} from './session-scanner-parse-cache'
+import { isolatedScanRoots } from './session-scanner-test-fixtures'
+import { parseClaudeSessionFile } from './session-scanner-primary-parsers'
+import type { FileWithMtime, SessionFileCandidate } from './session-scanner-types'
+
+// Spy-wrap (real implementations still run) so the zero-disk-IO test can
+// assert the uninitialized module never touches the filesystem.
+vi.mock('node:fs/promises', { spy: true })
+
+const APP_VERSION = '1.2.3-test'
+
+let tempRoots: string[] = []
+
+beforeEach(() => {
+  resetSessionParseCacheForTests()
+  resetSessionParseCachePersistenceForTests()
+})
+
+afterEach(async () => {
+  resetSessionParseCachePersistenceForTests()
+  await Promise.all(tempRoots.map((root) => rm(root, { recursive: true, force: true })))
+  tempRoots = []
+})
+
+async function makeTempDir(): Promise<string> {
+  const root = await mkdtemp(join(tmpdir(), 'orca-parse-cache-persist-'))
+  tempRoots.push(root)
+  return root
+}
+
+async function claudeCandidate(path: string): Promise<SessionFileCandidate> {
+  const fileStat = await stat(path)
+  const file: FileWithMtime = {
+    path,
+    mtimeMs: fileStat.mtimeMs,
+    modifiedAt: fileStat.mtime.toISOString(),
+    sizeBytes: fileStat.size
+  }
+  return { agent: 'claude', file, codexHome: null }
+}
+
+function userRecord(index: number, text: string): string {
+  return JSON.stringify({
+    type: 'user',
+    sessionId: 'aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee',
+    timestamp: new Date(1740000000000 + index * 60_000).toISOString(),
+    cwd: '/repo/app',
+    gitBranch: 'main',
+    message: { role: 'user', content: text }
+  })
+}
+
+function assistantRecord(index: number, text: string): string {
+  return JSON.stringify({
+    type: 'assistant',
+    sessionId: 'aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee',
+    timestamp: new Date(1740000000000 + index * 60_000).toISOString(),
+    message: {
+      role: 'assistant',
+      model: 'claude-fable-5',
+      content: [{ type: 'text', text }],
+      usage: { input_tokens: 100, output_tokens: 40 }
+    }
+  })
+}
+
+async function writeTranscript(root: string): Promise<string> {
+  const path = join(root, 'aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee.jsonl')
+  await writeFile(path, `${userRecord(0, 'first question')}\n${assistantRecord(1, 'answer')}\n`)
+  return path
+}
+
+async function parseAndPersist(path: string): Promise<SessionParseStats> {
+  const stats = createSessionParseStats()
+  await ensureSessionParseCacheLoaded()
+  await parseAgentSessionFileCached(await claudeCandidate(path), process.platform, stats)
+  scheduleSessionParseCachePersist(stats)
+  await flushSessionParseCachePersistForTests()
+  return stats
+}
+
+// Clear the in-memory cache and the persistence module's memoized state, then
+// re-enable persistence against the same file — a fresh launch, same profile.
+function simulateRestart(cacheFile: string, appVersion = APP_VERSION): void {
+  resetSessionParseCacheForTests()
+  resetSessionParseCachePersistenceForTests()
+  initSessionParseCachePersistence({ filePath: cacheFile, appVersion })
+}
+
+async function coldParseStats(path: string): Promise<SessionParseStats> {
+  const stats = createSessionParseStats()
+  await ensureSessionParseCacheLoaded()
+  await parseAgentSessionFileCached(await claudeCandidate(path), process.platform, stats)
+  return stats
+}
+
+describe('session parse cache persistence', () => {
+  it('round-trips: a persisted entry is a reused hit after a restart, without reading the transcript', async () => {
+    const root = await makeTempDir()
+    const cacheFile = join(root, 'vault-state', 'session-parse-cache.json')
+    initSessionParseCachePersistence({ filePath: cacheFile, appVersion: APP_VERSION })
+
+    const transcript = await writeTranscript(root)
+    const candidate = await claudeCandidate(transcript)
+    const stats = createSessionParseStats()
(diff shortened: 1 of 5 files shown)


brennanb2025:
## Live re-verification evidence (dev build from this branch, real local corpus)

Method: launched the dev build with a fresh temp profile via CDP, triggered `aiVault:listSessions` over IPC, read the `aiVault.scan` spans from `logs/main.trace.ndjson` in that profile, quit the app, relaunched with the **same** profile, and repeated.

**Run 1 — fresh profile (no cache file; equals pre-fix behavior on every launch):**

```
aiVault.scan  durationMs=3696  candidates=1501  reused=0  incremental=0  fullParses=1039  bytesRead=1788507131  (~1.79 GB)
```

Cache file written by the debounced save after the scan:

```
<userData>/ai-vault/session-parse-cache.json   2,649,381 bytes   mode 0600 (dir 0700)
schemaVersion=1  appVersion=1.4.146-rc.0  entries=1039
```

**Run 2 — first scan after quitting and relaunching the app (the #9210 scenario):**

```
aiVault.scan  durationMs=328  candidates=1501  reused=1036  incremental=0  fullParses=3  bytesRead=3097122  (~3.1 MB)
```

Restart cold scan went from 3.7 s / 1.79 GB / `reused=0` to 328 ms / 3.1 MB / `reused=1036` — the 3 full parses are transcripts that genuinely changed between the runs (this session's own activity). Renderer-side IPC round-trip: 3708 ms → 337 ms.

Baseline HEAD repro for comparison (same rig, before the fix): cold `durationMs=4574, reused=0, bytesRead=1792029210`; warm same-process forced rescan `durationMs=72, reused=1031` — the fix brings the restart scan to within ~4× of the in-process warm scan, with the remainder being stat/discovery of 1501 candidates plus the 3 changed files.

Full-window screenshots of both runs captured at `/tmp/orca-9210-run1-cold.png` and `/tmp/orca-9210-run2-warm-restart.png` (kept out of the repo per evidence policy; available on request).

brennanb2025:
Responding to the two optional nitpicks:

1. **`readdir` assertion in the zero-disk-IO test** — agreed it would tighten the guard (the orphan-`.tmp` sweep is the one path that calls `readdir`, and it's gated behind the same `options === null` early return the test pins). Deferring rather than re-spinning CI for one assertion line; will fold into the next touch of this file.

2. **Rebuild-the-map seeding refactor** — declining. The pre-populated-cache case is unreachable in production: `parseAgentSessionFileCached` has exactly one production caller (`session-scanner.ts`), and the scan awaits `ensureSessionParseCacheLoaded()` before any parse, so the cache is always empty at seed time; two independent review lanes verified this and stress-tested the live-entry+over-cap case (live entries survive and still hit — the failure mode is only fewer seeds, i.e. extra full parses, never wrong results). The proposed diff also drops pre-existing live entries on the floor (`existing` is built but never re-inserted), so as written it would trade a theoretical ordering wrinkle for real cache loss.

PR #9477, opened 2026-07-19
fix(source-control): keep huge change sets responsive

### LINKED ISSUE #7991: Source control freezes when a branch has too many changed files
## Problem

When switching to a branch with a large number of changed files (e.g., 20,000 files), the Source Control panel freezes the entire window for several seconds or permanently, making it unusable.

## Steps to Reproduce

1. Have a repo where a branch has a very large number of changed files (>= 10,000)
2. Switch to the Source Control tab in the right sidebar
3. The UI freezes

Alternatively: create or check out a branch with 10,000-20,000+ uncommitted files (e.g., an unignored build directory or dependency folder).

### 1. No virtual scrolling
The Source Control panel renders every entry as a React component via `visibleTreeRowsBySection[id].map(...)` or `visibleListRowsBySection[id].map(...)` at SourceControl.tsx L5961-L6069. All entries are mounted into the DOM at once. With up to `DEFAULT_GIT_STATUS_LIMIT = 10,000` (git-status-limit.ts) entries rendered as rich `UncommittedEntryRow` components, each with staged/unstaged/discard buttons and diff-comment badges, the browser thread freezes during both initial render and re-renders on state changes (filter, selection, etc.).

### 2. Heavy tree building on every render
`buildSourceControlTree`, `flattenSourceControlTree`, and `compactSourceControlTree` re-run every time `displaySections` changes (which happens on every status poll update), rebuilding the full tree data structure for all entries even when no visible data changed. Combined with the lack of DOM virtualization, this doubles the freeze duration.

### 3. Per-entry diff comment lookup
Each `UncommittedEntryRow` receives `diffCommentCountByPath.get(node.entry.path)` - a Map lookup across a potentially large comments array. While individually cheap, this is called for every one of 10,000 rows.

## Expected Behavior

The Source Control panel should remain responsive regardless of change count. For huge changes, the panel should either:
- Virtualize the file list (only render visible rows, e.g., via `react-window`/`virtuoso` or a custom overscan approach), OR
- Show the "too many changes" banner earlier (lower limit) and skip full rendering entirely.

## Suggested Fixes (in order of preference)

1. **Virtual scrolling.** Wrap the per-section file list with a virtualized list component (e.g., `react-window` `FixedSizeList` or `virtuoso`) so only viewport-visible rows are mounted. Estimated ~100 DOM nodes instead of 10,000.

2. **Reduce the status entry limit.** Lower `DEFAULT_GIT_STATUS_LIMIT` from 10,000 to something that the existing non-virtualized renderer can handle without freezing (e.g., 1,000-2,000). The "too many changes" banner already exists; it would just trigger sooner.

3. **Debounce tree re-computation.** Wrap the `buildSourceControlTree`/`flattenSourceControlTree` chain in a `useMemo` with a structural equality check to avoid rebuilding the tree when entries haven't materially changed between polls.

## Environment

- Orca version: v1.4.129-rc.1
- Affects all platforms (macOS, Linux, Windows)

OrcaWin:
## Summary

- stream porcelain status incrementally on relay hosts and terminate Git as soon as Orca has enough entries
- cap source-control data at 1,000 entries before IPC or SSH/relay transfer, preserving an explicit `didHitLimit` signal
- keep huge-repository watcher suppression sticky while adding an explicit Retry path that recovers after the repository drops below the cap

## ELI5

A branch with thousands of changes made Orca carry and render a giant list all at once, briefly freezing the window even though only a capped list was shown. Orca now stops reading as soon as it has the capped result, sends only that bounded result across process and host boundaries, pauses automatic refresh churn, and lets the user retry explicitly after cleaning up files.

## Electron evidence

Exact fixture in both runs: one committed file plus exactly 12,000 untracked TypeScript files in 120 directories.

| Measurement | Before | After |
| --- | ---: | ---: |
| Initial Source Control activation | 1601.6 ms | 1516.9 ms |
| Renderer max event-loop lag during activation | 1113.6 ms | 664.4 ms |
| Explicit status scan | 529.6 ms | 284.0 ms |
| Warm rescan | 146.4 ms | 196.9 ms |
| IPC payload | 850,222 bytes | 85,221 bytes |
| Entries crossing IPC | 10,000 | 1,000 |
| Post-activation max lag | 29.2 ms | 20.7 ms |
| Render update | — | 2.0 ms |

The exact Electron flow passed twice after the fix, including a fresh rebuild. Removing all 12,000 untracked files and choosing Retry cleared capped mode, proving explicit recovery.

## Validation

- 429 focused parser, status, relay, IPC, provider, runtime, store, and polling tests
- additional relay/component run: 192 passed, 10 skipped
- `typecheck:node`
- `typecheck:web`
- changed-file oxlint and max-lines ratchet
- localization catalog and coverage checks
- all six relay target builds plus the WSL hook relay
- exact 12,000-file Electron E2E

## Trade-off / user-visible change

Repositories with more than 1,000 changes now enter capped mode earlier, show the first 1,000 entries, pause automatic watcher and safety refreshes, and expose Retry. The smaller bound is what keeps renderer projection and cross-host transfer responsive; explicit Retry preserves recovery after cleanup.

Fixes #7991


OrcaWin:
Electron before/after proof for the exact 12,000-file fixture: [https://htmlpreview.github.io/?https://gist.githubusercontent.com/OrcaWin/6647ab6a294ab47bc1898334076e66b6/raw/orca-7991-electron-proof.html](https://htmlpreview.github.io/?https://gist.githubusercontent.com/OrcaWin/6647ab6a294ab47bc1898334076e66b6/raw/orca-7991-electron-proof.html)

The artifact embeds both temporary screenshots and their SHA-256 hashes; neither the screenshots nor fixture are committed to the repository. The measured IPC boundary changed from 10,000 entries / 850,222 bytes to 1,000 entries / 85,221 bytes, and deleting the fixture followed by **Retry** verified recovery.

PR #9483, opened 2026-07-20
fix(cli): avoid Windows PATH status timeout

### LINKED ISSUE #9462: [Bug]: Windows PATH command times out (5000ms) when loading CLI install status on Windows 10
### Operating system

Windows 10

### Orca version

Unknown (please add from **Orca → About Orca** if known)

### Details

**Short summary**  
On Windows 10, Settings fails to load CLI / WSL CLI install status. Two error toasts appear:

1. `Error invoking remote method 'cli:getWslInstallStatus': Error: Windows PATH command timed out after 5000ms.`
2. `Error invoking remote method 'cli:getInstallStatus': Error: Windows PATH command timed out after 5000ms.`

**What happened?**  
Opening (or refreshing) Settings that probe CLI registration shows toast errors instead of install status. Both the native Windows CLI status probe and the WSL CLI status probe fail the same way. Screenshot was taken from a Windows 10 machine (monitor photo of the Orca Settings UI with the two error toasts).

**How can we reproduce it?**  
1. Run Orca on **Windows 10**.
2. Open Settings related to CLI registration / WSL CLI (or any surface that loads CLI status).
3. Observe toasts for `cli:getInstallStatus` and `cli:getWslInstallStatus` both reporting `Windows PATH command timed out after 5000ms.`

**Likely cause (code)**  
On `win32`, `CliInstaller.getStatus()` checks whether the command directory is on the user PATH via PowerShell:

- `isPathConfigured()` → `userPathReader()` → `readWindowsUserPath()` → `runWindowsPathCommand()`
- `runWindowsPathCommand` runs `powershell -NoProfile -Command "[Environment]::GetEnvironmentVariable('Path','User')"` with `WINDOWS_PATH_COMMAND_TIMEOUT_MS = 5000` and throws if it exceeds that.
- Source: `src/main/cli/cli-installer.ts` (`runWindowsPathCommand` / timeout message).
- IPC: `cli:getInstallStatus` / `cli:getWslInstallStatus` in `src/main/ipc/cli.ts`.
- WSL status also consults the host installer (`WslCliInstaller` → `hostInstaller.getStatus()`), so a wedged host PATH read can fail **both** IPC methods.

**Why this is painful**  
- Status is a read-only probe used widely in Settings / skill setup; a hard throw surfaces as uncaught IPC errors in the UI.
- 5s may be too aggressive on slower Windows 10 machines, cold PowerShell startup, Defender scanning, or broken/slow `powershell` resolution (e.g. Store vs Windows PowerShell).
- Related recent work: PowerShell launch on Windows (`fix(win): launch Store-installed PowerShell 7 in terminals`).

**Expected**  
- Status probes should not hard-fail the UI when PATH read is slow.
- Prefer: longer timeout for status reads, graceful degradation (`pathConfigured: false` + detail), and/or a faster PATH source (e.g. registry / `reg query`) that does not require full PowerShell startup for reads.

**Environment notes**  
- OS: Windows 10 (reporter)
- Both `getInstallStatus` and `getWslInstallStatus` fail with the same timeout message

Jinwoo-H:
## Summary

- Read the persisted Windows user PATH through an optional Windows-native registry module instead of spawning PowerShell whenever CLI install status loads.
- Represent registry read failures as an explicit unknown PATH state, fail closed before every PATH mutation, and surface that state inline in Settings/onboarding/agent setup.
- Coalesce concurrent status reads, force a fresh read before every PATH mutation, preserve `REG_SZ` versus `REG_EXPAND_SZ` semantics, and continue WSL probing when only the host PATH state is unknown.
- Select native rebuilds and packaged resources by artifact target so Windows gets the addon while macOS, Linux, SSH, and cross-builds retain their existing behavior.
- Restore node-pty's ConPTY sidecar files before probes and after native rebuilds, because `@electron/rebuild` bypasses node-pty's postinstall copy step.

Closes #9462.

## Screenshots

No visual change during normal operation. When the registry cannot be read, the existing CLI Settings row now shows the diagnostic inline and disables its mutation switch; this state is covered by `CliSection.test.tsx`.

## Testing

- [ ] `pnpm lint` — targeted Oxlint passed; the repository-wide command reaches an unrelated existing type-aware error in `src/renderer/src/components/skills/skill-freshness-group.tsx`.
- [x] `pnpm typecheck`
- [ ] `pnpm test` — the focused suite passed 109 tests across 10 files (35 platform skips). The repository-wide suite completed but has unrelated existing Windows failures involving POSIX paths, `/bin/sh`, symlink semantics, and `localStorage` mocks.
- [ ] `pnpm build` — a Windows unpacked app was built and exercised; the normal builder later hit the existing `node-pty`/Visual Studio toolchain rebuild issue and Windows signing-helper symlink privilege requirement.

Additional validation:

- The exact Windows terminal-restart CI suite passed all 3 restart scenarios after repairing the ConPTY runtime payload.
- Packaged Electron 43.1.0 / Node 24.18 / N-API 10 loaded the copied native module and read the real 1,176-character user PATH.
- 20 concurrent packaged-app status IPC calls completed consistently in 10.8 ms without timeout or rejection.
- 100 direct registry reads averaged 0.389 ms (3.292 ms maximum) on Windows 11.
- Packaged Settings rendered the expected Windows CLI state while WSL status continued loading normally.
- Max-lines ratchet, formatting, localization catalog/parity/coverage, and Electron native-runtime checks passed.

## AI Review Report

An independent solution review preferred `reg.exe` to avoid adding native-module maintenance. That review also identified unresolved output-encoding risk for Unicode PATH entries. The implementation therefore uses the Unicode Windows registry API, while containing the added maintenance surface through an optional dependency, runtime platform checks, targeted rebuild verification, and packaged-resource contract tests.

The review explicitly checked macOS, Linux, Windows, WSL, and SSH behavior; path comparison and `%VARIABLE%` expansion; concurrent IPC callers; Electron ABI/package loading; unknown/error behavior; read-modify-write safety; native rebuild behavior; and target-specific packaging. No remaining cross-platform runtime regression was identified.

## Security Audit

- Reads only `HKCU\Environment\Path`; it does not access machine-wide or privileged registry data.
- Registry failures never become an empty PATH. Install/remove abort before writing, preventing accidental PATH truncation.
- `%VARIABLE%` entries are expanded for comparison only when PATH is `REG_EXPAND_SZ`; literal `REG_SZ` text and stored user text are preserved.
- Existing PowerShell writes remain user-scoped and retain their quoting, timeout, and permission-error translation.
- The optional native dependency is pinned to `3.2.2`, loaded dynamically only on Windows, and verified in both the development Electron runtime and packaged app.
- No auth, secret, network, or provider behavior changes.

## Notes

Primary merge cost is maintenance: one pinned native module plus Windows rebuild/package verification. Runtime costs are a small synchronous in-process registry read (sub-millisecond average here), bounded by one-second caching and concurrent-read coalescing. If the native read is unavailable, CLI mutation is intentionally disabled until status can be read safely.

Physical Windows 10, Windows ARM64, and managed-device registry denial were not available for manual testing. The API used is supported by those Windows versions, and denied/unavailable reads are covered by fail-closed tests. PowerShell is still used for the rare install/remove write path, so this PR removes the status-load timeout but does not redesign PATH writes.


Jinwoo-H:
@coderabbitai review

PR #9488, opened 2026-07-20
Match PR Files changed diff order to the file tree

davkim1030:
## Summary

The PR "Files changed" diff sections were ordered by the raw API response (`files.map(gitHubPRFileToBranchEntry)`), while the file tree beside them renders in source-control order — depth-first, directories before files, path-sorted within each level. Scrolling the list and scrolling the diff therefore disagreed.

This derives the section order from **the same tree the list renders**, via a new `collectSourceControlTreeFileEntriesInOrder()` export, rather than reimplementing a parallel sort that would drift.

## ELI5

The Files-changed page has a list of files on one side and the actual diffs on the other. The list was sorted one way and the diffs another, so clicking or scrolling one didn't line up with the other. Now both come from the same source, so they always agree.

## Proof of fix

**With `source-control-tree.ts` restored from the branch's merge base:**

```
$ git checkout $(git merge-base HEAD origin/main) -- \
    src/renderer/src/components/right-sidebar/source-control-tree.ts
    src/renderer/src/components/right-sidebar/source-control-tree.test.ts

 × orders file entries to match the file tree the PR view actually renders
TypeError: collectSourceControlTreeFileEntriesInOrder is not a function

      Tests  1 failed | 9 passed (10)
```

**With this PR:**

```
 Test Files  1 passed (1)
      Tests  10 passed (10)
```

## Proof of no regressions

```
    src/renderer/src/components/right-sidebar/source-control-tree.test.ts
 Test Files  1 passed (1)
      Tests  10 passed (10)

$ npx oxlint --config config/oxlint-react-doctor.json \
    src/renderer/src/components/PullRequestPage.tsx
5 warnings — identical count on the merge base (pre-existing, unrelated)
```

**The sidebar tree order is unchanged.** This was the main regression risk, since `source-control-tree.ts` is shared: a change to its comparator would silently reorder the source-control sidebar for everyone. Verified the diff **adds** an export and touches no existing comparator — `git diff … | grep -E '^[-+].*(localeCompare|sort|compare)'` returns nothing but the new doc comment.

- PR "Files changed" diff sections now appear in the same order as the file tree. That is the fix.
- Nothing else. The source-control sidebar, the file list itself, and every other consumer of `source-control-tree.ts` are untouched.

## Compatibility

- **Platforms:** macOS / Linux / Windows. Pure ordering logic over already-fetched metadata; no paths are constructed, no shell, no key modifiers, no Electron API. The sort is deterministic and filesystem-independent — it operates on API-provided path strings, so case-insensitive (macOS/Windows) and case-sensitive (Linux) filesystems produce identical output.
- **Provider compatibility (`AGENTS.md`):** checked, since `PullRequestPage` also serves GitLab, Gitea, and Bitbucket. The new function takes `SourceControlTreeEntry` values built from `gitHubPRFileToBranchEntry`, which normalizes provider payloads to a common shape before the tree is built — no GitHub-only field is referenced in the ordering path.
- **Performance:** no per-render re-sort. The tree build sits **inside the existing `entriesCacheRef` signature guard**, so it runs only when `diffEntrySignature` changes, not on every render. For a 1000-file PR this is one `O(n log n)` pass per file-set change rather than per paint.

## Screenshots

The change is an ordering difference between two panes — a static screenshot of a correctly-ordered list is indistinguishable from an incorrectly-ordered one without knowing the expected order, so an image would not evidence the fix. The guarantee is instead pinned by a test asserting the emitted order matches the rendered tree. I did not capture before/after images; stating that plainly rather than implying a visual check I didn't perform.

## Testing

- [ ] `pnpm lint` — not run in full. Ran `npx oxlint --config config/oxlint-react-doctor.json` on the changed component: 5 warnings, **identical count on the merge base**, so nothing introduced here.
- [ ] `pnpm test` — not run in full (30+ min). Ran the touched module's suite: 10 tests passed.
- [ ] `pnpm build` — not run.
- [x] Added or updated high-quality tests that would catch regressions — a test asserting file entries are emitted in the order the tree actually renders; it fails at the merge base.

### On the failing `track-community-pr` check

**Not a code failure.** The job is org bookkeeping — it calls
`GET /orgs/stablyai/teams/stably-eng/memberships/davkim1030` to classify the author as internal or community, and errors because the author is not a member of that team, which is the expected state for a community contributor. It does not build, lint, or test anything in this PR. No action needed from the author.

## AI Review Report

Reviewed with Claude Code (Opus 5). Risks checked:

- **Reuse vs reimplementation.** The correct fix is to consume the existing tree, not to re-derive an equivalent sort — two independent sorts drift the moment either changes. Verified the PR adds a traversal helper over `buildSourceControlTree` rather than a second comparator.
- **Shared-module blast radius.** `source-control-tree.ts` backs the source-control sidebar. Confirmed no existing comparator or sort call was modified, so sidebar ordering is byte-identical.
- **Locale-dependent sorting** — the defect I most expected to find. The comparator uses `a.path.localeCompare(b.path, undefined, { numeric: true })`, and `undefined` means "use the runtime's locale", which genuinely diverges: sorting `['a.ts','ab.ts','ä.ts','z.ts']` puts `ä.ts` before `ab.ts` under the default and after `z.ts` under `sv-SE`. **This is pre-existing on `main`, not introduced here** (verified via `git show origin/main:…`), and this PR makes the two panes *consistently* locale-ordered rather than inconsistently ordered — a strict improvement. Flagged as a follow-up below rather than silently expanded.
- **Perf on large PRs:** confirmed the tree build is inside the `diffEntrySignature` cache guard, so it is not re-run per render.
- **Provider neutrality:** ordering operates on normalized entries, not raw GitHub fields.
- **Cross-platform (macOS / Linux / Windows):** no platform-dependent code; ordering is filesystem-independent.

## Security Audit

- **Input handling:** none added. The change reorders already-parsed diff metadata.
- **Path handling:** paths are compared as strings for ordering only; none are constructed, joined, or resolved, so no traversal surface.
- **IPC:** untouched.

## Notes

- **Follow-up worth opening (pre-existing, out of scope):** `compareEntries` passes `undefined` as the `localeCompare` locale, so file ordering varies with the user's system locale — two developers on the same PR can see different orders, and it makes ordering assertions locale-fragile. Pinning an explicit locale (or using a codepoint comparator) would make it deterministic. Deliberately not changed here, because it would alter the source-control sidebar ordering for every user and belongs in its own PR.
- Branch is behind `main`; a rebase is advisable before merge, though nothing in the diff conflicts.

Original patch by @davkim1030.

### DIFF
--- a/src/renderer/src/components/PullRequestPage.tsx
+++ b/src/renderer/src/components/PullRequestPage.tsx
@@ -80,6 +80,10 @@ import {
   handleCombinedDiffFileTreeNavigation
 } from '@/components/editor/CombinedDiffFileTree'
 import {
+  buildSourceControlTree,
+  collectSourceControlTreeFileEntriesInOrder
+} from '@/components/right-sidebar/source-control-tree'
+import {
   getDiffSectionEstimatedHeight,
   isIntrinsicHeightImageDiff
 } from '@/components/editor/diff-section-layout'
@@ -2259,7 +2263,12 @@ function PRFilesCombinedDiffViewer({
     if (entriesCacheRef.current?.signature === diffEntrySignature) {
       return entriesCacheRef.current.entries
     }
-    const nextEntries = files.map(gitHubPRFileToBranchEntry)
+    // Why: the file tree renders in source-control DFS order (directories first,
+    // then path-sorted files); order the diff sections identically so the list
+    // and the diff detail scroll match, like GitHub's Files changed view.
+    const nextEntries = collectSourceControlTreeFileEntriesInOrder(
+      buildSourceControlTree('combined-commit', files.map(gitHubPRFileToBranchEntry))
+    )
     entriesCacheRef.current = {
       signature: diffEntrySignature,
       entries: nextEntries
--- a/src/renderer/src/components/right-sidebar/source-control-tree.test.ts
+++ b/src/renderer/src/components/right-sidebar/source-control-tree.test.ts
@@ -86,29 +86,40 @@ describe('buildSourceControlTree', () => {
     ])
   })
 
-  it('orders file entries to match the flattened tree file rows regardless of input order', () => {
-    // Why: the PR "Files changed" diff sections must render in the same order as
-    // the file tree; this helper is the shared ordering both views rely on.
+  it('orders file entries to match the file tree the PR view actually renders', () => {
+    // Why: the PR "Files changed" diff sections order via
+    // collectSourceControlTreeFileEntriesInOrder, while the file tree renders via
+    // compactSourceControlTree(buildSourceControlTree(...)) + flatten (see
+    // CombinedDiffFileTree.buildBranchRows). Assert both leaf orders match so the
+    // list and diff stay in sync — including that compaction preserves leaf order.
     const entries: GitBranchChangeEntry[] = [
       { path: 'README.md', status: 'modified' },
-      { path: 'src/index.ts', status: 'modified' },
-      { path: 'src/components/Button.tsx', status: 'added' },
+      { path: 'src/renderer/index.ts', status: 'modified' },
+      { path: 'src/renderer/components/Button.tsx', status: 'added' },
+      { path: 'src/main/app.ts', status: 'modified' },
       { path: 'docs/guide.md', status: 'modified' }
     ]
 
-    const tree = buildSourceControlTree('combined-commit', entries)
-    const treeFileRows = flattenSourceControlTree(tree, new Set())
+    // Mirror CombinedDiffFileTree.buildBranchRows exactly (compact + flatten).
+    const renderedTreeFileOrder = flattenSourceControlTree(
+      compactSourceControlTree(buildSourceControlTree('combined-commit', entries)),
+      new Set()
+    )
       .filter((node) => node.type === 'file')
       .map((node) => node.path)
-    const orderedPaths = collectSourceControlTreeFileEntriesInOrder(tree).map((item) => item.path)
 
-    expect(orderedPaths).toEqual(treeFileRows)
-    // Input API order (README, src/index, src/components/Button, docs) is
-    // reordered into directory-first, path-sorted tree order.
-    expect(orderedPaths).toEqual([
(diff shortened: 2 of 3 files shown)


nwparker:
Needs a rebase — this branch now conflicts with `main`:

```
CONFLICT (content): Merge conflict in src/renderer/src/components/PullRequestPage.tsx
```

Not resolving it here: `PullRequestPage.tsx` has moved substantially on `main`, and the right resolution depends on where the diff-entry memo now lives. The ordering change itself is small and should re-apply cleanly.

Verified green against the branch's own merge base — details in the PR description. Also note the failing `track-community-pr` check is org bookkeeping (a team-membership lookup), not a code failure; no action needed on it.

nwparker:
## Superseded by #10081 — recommend closing

The bug is fixed on main. Commit `1bd36ce04b` ("fix(github): align PR file diff order", #10081, merged 2026-07-24) addresses the identical defect at the identical call site.

I verified main's implementation: `getCombinedDiffBranchEntriesInTreeOrder` is now used at `PullRequestPage.tsx:2281` **and** `GitHubItemDialog.tsx:2224`. That second call site is the meaningful difference — this branch fixes only `PullRequestPage`, leaving the duplicated viewer in `GitHubItemDialog` still ordered by the raw API response.

Main's approach is also closer to the truth: it orders via `compactSourceControlTree` + `flattenSourceControlTree`, which is the tree's actual render path, whereas this branch traverses without compaction and relies on a test asserting that compaction happens to preserve leaf order. Main's version also keeps the logic in `combined-diff-file-tree-model.ts` rather than adding a new export to the shared `source-control-tree.ts`, so the source-control sidebar's module is untouched.

Also worth recording, since it may confuse anyone revisiting this: the failing `track-community-pr` check is **not** a code failure. It's org bookkeeping that calls the team-membership API to classify the author, and it errored on a GitHub 5xx. It builds, lints, and tests nothing in this PR.

The branch is conflicting against main. Recommending close as superseded — the diagnosis was right, and #10081 shipped the same fix with wider coverage.

davkim1030:
@nwpark Thanks for the comments
I will close this PR and related issue.

PR #9499, opened 2026-07-20
fix(ai-vault): resume Claude sessions from their start directory

choihjin:
## Summary

Resuming a Claude session that changed directories mid-run no longer fails with "No conversation found" — sessions now resume from their original start directory.

## ELI5

If you started a chat in one folder and later moved into a subfolder, the "resume" button used to point at the wrong folder and couldn't find your conversation. Now it always remembers where you began, so resume just works.

## Root cause & fix

The AI Vault session scanner recorded each session's `cwd` by overwriting it with every transcript record, so a session that drifted into a subdirectory ended up with its **last-seen** cwd. Claude stores transcripts under `~/.claude/projects/<slug-of-start-dir>/` and `claude --resume <id>` only searches the project dir derived from the current cwd, so the rebuilt `cd '<subdir>' && claude --resume <id>` ran from the wrong place and failed. The fix captures the **first-seen** (start) cwd instead — transcripts are append-only and write the cwd up front, so the first record carrying one is the true start directory. This is fixed at the accumulator, the single source of `session.cwd`, repairing resume, folder grouping, and workspace filtering at once.

## Evidence

- Rebase clean; single commit touching only `src/main/ai-vault/session-scanner-accumulator.ts` (fix) and `src/main/ai-vault/session-scanner-claude-cwd-drift.test.ts` (test).
- Test `session-scanner-claude-cwd-drift.test.ts` passes on branch: **Test Files 1 passed (1); Tests 1 passed (1)**.
- Reverting only the fix file (keeping the test) makes it fail, proving the test captures the bug:

```
expected cwd "/repo/app/services/api" to match "/repo/app"
expected resumeCommand
  "cd '/repo/app/services/api' && claude --resume 'drift-session'"
  to match "cd '/repo/app' && claude --resume 'drift-session'"
```

- Lint clean: `oxlint` on the fix file reported no warnings or errors.

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression: `updateLatestLocation` is called only by the Claude parser, so no other agent's behavior changes. The branch still tracks the latest location for its own use; only the persisted `session.cwd` derivation changes. Non-affected paths are byte-identical, and correctness is proven by the passing test plus the fail-on-revert assertion.

---

_Credit to original author @choihjin (Gideon). Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/main/ai-vault/session-scanner-accumulator.ts
+++ b/src/main/ai-vault/session-scanner-accumulator.ts
@@ -187,16 +187,23 @@ export function updateLatestLocation(
   accumulator: SessionAccumulator,
   record: Record<string, unknown>
 ): void {
+  // Why: a session's representative cwd is its START directory, not its latest.
+  // `claude --resume <id>` only finds the transcript under the project dir keyed
+  // by the start cwd, and history grouping/filtering key off the session origin;
+  // a later drifted cwd broke resume for sessions that changed directory (#9361).
+  // Transcripts are append-only, so the first record carrying a cwd is the start.
+  if (accumulator.cwd === null) {
+    const startCwd = extractString(record.cwd)
+    if (startCwd) {
+      accumulator.cwd = startCwd
+    }
+  }
   const timestamp = extractString(record.timestamp)
   const parsed = timestamp ? Date.parse(timestamp) : accumulator.latestTimestampMs
   if (!Number.isFinite(parsed) || parsed < accumulator.latestTimestampMs) {
     return
   }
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @choihjin — resume Claude sessions from their start directory. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #9506, opened 2026-07-20
fix(editor): don't flag editor-initiated moves as changed-on-disk

nwparker:
## Summary

Moving or renaming a file **through Orca's own File Explorer** (drag-and-drop, inline rename, tab rename, or naming an untitled file) while it is open with unsaved edits could pop a spurious **"This file changed on disk"** conflict banner on the just-moved tab — even though the user only moved it — and "Reload from Disk" then discarded the unsaved draft.

This reworks an in-app move to be a **retarget of the live edit session in place** (never a close+reopen), and decides whether a watcher event on the moved path is the move's **own echo** or a **genuine external write** by **content identity** — so the false banner is gone, real external writes are never swallowed, and the moved tab keeps its full state (cursor, scroll, view mode, group placement, drafts).

## ELI5

You have a file open with unsaved typing and you move it in the file tree. Orca now keeps that exact editor session and just changes where it points — it never closes and reopens the tab, so your cursor, scroll position, view mode, and unsaved text all stay put. Behind the scenes Orca also watches the disk; when it sees the move land, it does the obvious check — *reads the file and asks "is the content actually different from what I last had?"* A move changes no bytes, so it stays quiet; if something genuinely rewrote the file, the content won't match and you get the warning, correctly. While it checks, it pauses auto-save so a save can't overwrite a real external change first, and if it can't read the file it errs toward showing the warning.

## How this compares to established editor behavior

Same core principle mature editors use: an in-app move retargets the live session in place, and a dirty buffer is never auto-clobbered. Orca keys editor sessions by a **path-derived id correlated to the move operation** rather than a stable per-buffer id — a stable buffer id needs an inode-stable substrate Orca can't get portably across Windows / network / SSH / WSL. The one deliberate divergence from the common approach: editors typically decide "did content change" from cheap **metadata** (mtime / etag / inode); Orca's watcher events are **path-only**, so we compare a **content signature** instead. That's not a shortcut — metadata false-negatives on coarse-granularity filesystems (a real overwrite inside the same mtime tick), which for a fail-closed suppress-the-only-warning decision is a data-loss hole. The content read fires only on the rare "moved a dirty file" echo (deduped), so there's no real perf to reclaim. (Analysis in the design note.)

## What changed (staged)

1. **Atomic in-place rekey** (`rekeyOpenFilesForPathChange`) — one commit-only store update migrates every path-derived id + all id-keyed state (full `OpenFile` spread, the 6 per-file maps, active ids, tab bar, unified tabs/groups via the widened `migrateHydratedEditorTabsAndGroups`, previews, untitled, pending reveal). Fails closed on collision (never merges two live sessions) / stale, with **zero** mutation. Replaces the close+reopen that discarded cursor/view/group state and opened a watcher race.
2. **Op-scoped in-flight move registry** (`editor-path-move-inflight`) — tracks a move only for the rename+rekey duration (no TTL): source paths suppress the delete tombstone; the destination is content-verified after commit.
3. **Fail-closed content verification** — the move-echo gate + provenance live **on `OpenFile`** (survive the rekey; op-id token supersedes a stale verification). On a destination event the watcher reads the file and compares `getDiskBaselineSignature(disk)` to the carried baseline: equal ⇒ echo (suppress), differ/binary/read-error/missing-baseline ⇒ conflict. Autosave is suspended synchronously before the read. The coordinator also proactively content-verifies every gated tab after a successful move, so the auto-save gate resolves even if the destination watcher event never arrives.
4. **Coordinator** (`executeOpenEditorPathMove`) — one transaction for every move (inline/tab rename, drag-drop, undo, redo, untitled): quiesce saves → begin in-flight suppression → on-disk rename → atomic rekey → settle → proactive re-verify. On failure the store is byte-identical (a rekey collision inverse-renames the on-disk move back). The old TTL self-move registry + close+reopen path are removed.
5. **Execution-host scoped** — the same absolute path can be open on more than one execution host (a second SSH connection, or local vs runtime) as a distinct physical file. Both the affected-set and the rekey plan are scoped by the initiating worktree's execution host, so a coincidental same-path tab on another host is never retargeted.
6. **Mirror-safe (companion tabs)** — because the rekey changes a tab's id, a moved mirrored tab would be culled by the host snapshot (losing its draft); the rekey **detaches** it to a local tab and the coordinator **close-notifies** the host's old-path tab. (The full host-rekey path-change protocol that preserves mirror ownership is a documented follow-up.)

## Deferred follow-ups (named)
- `move-fix-mirror-path-change-protocol` — full `session.tabs.editorPathChange` begin/commit/abort RPC + stale-snapshot translation so the host re-homes its mirrored tab in place.
- `move-fix-recently-closed-path-rewrite` — rewrite path fields in `recentlyClosedEditorTabsByWorktree`.
- `move-fix-cross-worktree-late-open-scope` — source suppression registers the initiating scope plus every already-affected scope. A file opened under the moving directory in a *different* same-host worktree (e.g. a floating view) whose scope wasn't already registered, during the rename's in-flight window, could momentarily show a stale conflict banner (no data loss — autosave is suspended and in-memory content is preserved). A host-scoped registry would close this extreme-window edge.

## Test plan
- Store rekey: full-object/draft/dirty/baseline/cursor/view/active/tabbar/group preservation; owner-aware id assignment; markdown source+preview; untitled consume; pending-reveal migrate; collision + stale = zero change; mirror detach.
- In-flight registry: source suppression scoped by owner; settle; concurrent ops independent.
- Verification: echo (disk==baseline) suppresses; differ/binary/no-baseline/read-error fail closed; no-provenance uses the normal path; autosave gated before read.
- Coordinator: end-to-end retarget + gate install; proactive gate resolution with no watcher event; collision inverse-renames + leaves the store untouched; same-path tab on a different execution host left untouched.
- Full editor/lib/hooks/store/runtime/right-sidebar suites green; typecheck + oxlint clean.

## Design decision record
The locked design and the content-read-vs-metadata analysis are kept in a local, un-checked-in design note.


nwparker:
cc @grozensmith -- if you can update to latest this should resolve some issues you may have had in regards to file content issues

PR #9516, opened 2026-07-20
fix(worktree): bound teardown RPCs so Windows workspace deletion can't hang

OrcaWin:
## Summary

Fixes native-Windows workspace deletion failing with:

```
Error invoking remote method 'worktrees:remove': Error: Timed out waiting for
physical PTY teardown: 9edc9ec2-…::C:/Users/NINGMEI/orca/workspaces/clue-ruoyi/issue-14
```

### Root cause

Destructive worktree removal runs `killAllProcessesForWorktree(..., requirePhysicalStop: true)`, which sweeps all PTYs for the worktree under a **10s** deadline (`WORKTREE_PROCESS_SWEEP_TIMEOUT_MS`) and refuses to touch the filesystem until every PTY is proven dead (correct fail-closed safety — a live process holding a handle would break the delete on Windows).

On win32 with the daemon on, the **`DaemonPtyAdapter` is installed as the local PTY provider**, so each PTY stop routes into `client.request('kill', …)`. That RPC (and the `listSessions` liveness check, and `ensureConnected()`) used the **DaemonClient 30s default timeout — larger than the 10s sweep deadline**. So when the daemon was slow/wedged to reply (a stuck ConPTY teardown, a class of state reworked in #9500), the *outer* 10s deadline fired first and surfaced the confusing "Timed out waiting for physical PTY teardown" — instead of the accurate "Failed to physically stop every PTY" — and blocked deletion entirely. The same misordered-timeout defect existed on the SSH relay path and the runtime-graph sweep (the common case: deleting a workspace whose terminal is open in the UI).

### Fix

Bound **every** RPC on the destructive-removal path to the *remaining* sweep budget (minus a 500ms propagation margin) so an inner failure surfaces *before* the outer deadline:

- Added an optional `timeoutMs` to `IPtyProvider.shutdown` / `listProcesses` (additive/optional — local & SSH providers structurally unchanged).
- `worktree-teardown.ts` passes `boundedRpcTimeoutMs(deadline)` on the **provider** and **registry** sweeps.
- **Runtime-graph sweep** (the reported UI case): threaded the bound through `stopTerminalsForWorktree` → `ptyController.stopAndWait` → `shutdownProviderAndDetectExit`/`provider.shutdown` and `verifyPtyStopped` → `listProcesses`. `stopAndWait` reconstructs one absolute deadline and hands each of its two sequential RPCs only the *remaining* budget, and bounds the cold-start `startupPromise` wait (fail-closed).
- **Daemon adapter** shares one budget across `ensureConnectedWithin` + the RPC, so a wedged handshake can't burn the whole deadline before the kill starts.
- **SSH provider** forwards the bound to the relay multiplexer (`mux.request(..., { timeoutMs })`).

Non-teardown callers (tab close, sleep/keepHistory, restart-daemon, renderer kill) pass no `timeoutMs` → the 30s default and original connect behavior are preserved exactly.

## Evidence of fix

**Reproduction (red → green).** Added `worktree-teardown.test.ts` regression test that drives the **real** `DaemonPtyAdapter` through the **real** `killAllProcessesForWorktree` with a fake daemon RPC client whose `kill` never replies (faithfully modeling the 30s default), under fake timers advancing to the 10s deadline.

- **Before the fix:** rejects with `Timed out waiting for physical PTY teardown: w1` — the exact production error.
- **After the fix:** the bounded kill RPC rejects first, so the sweep reports the accurate `Failed to physically stop every PTY for worktree: w1`. ✅

**Coverage added**
- `worktree-teardown.test.ts` — wedged-daemon regression (incl. bounded `ensureConnectedWithin`); sibling-worktree negative assertions tightened to `objectContaining` so they can actually fail.
- `pty.test.ts` — `stopAndWait` split-budget proof: shutdown gets `4321ms`, then after burning 1000ms the liveness `listProcesses` gets exactly `3321ms` (deterministic under fake timers).
- `ssh-pty-provider.test.ts` — asserts the relay mux receives `{ timeoutMs }`.
- `orca-runtime.test.ts` — destructive stop passes a bound `≤ 10000 − 500`.

**Results**
- Affected suites green: `worktree-teardown`, `pty`, `daemon-pty-adapter`, `daemon-pty-router`, `degraded-daemon-pty-provider`, `ssh-pty-provider`, `local-pty-provider` — **569 passed / 15 skipped**.
- `orca-runtime.test.ts` — 763 passed; the 1 failing `badgeColor … createRepo dedupe` test **fails identically on the baseline with this diff stashed** (pre-existing Windows path/UUID nondeterminism), unrelated to this change.
- `pnpm run typecheck` — clean (exit 0).
- Reviewed adversarially across four rounds (correctness/regression, cross-platform/SSH/daemon-on-off, completeness/root-cause, final verification) until CLEAN.

## ELI5

When you delete a workspace, the app first makes sure all the little terminal programs inside it are actually shut down before deleting files — otherwise Windows won't let go of the folder. It gives the whole cleanup **10 seconds**.

The bug: it asked the background "daemon" to kill each terminal but was willing to wait **30 seconds** for an answer. So if the daemon was slow, the 10-second overall timer ran out first and threw a scary "timed out" error — and your workspace wouldn't delete.

The fix: never wait longer for one step than the whole job is allowed to take. Now each kill request is told "you have less than 10 seconds." If the daemon is genuinely stuck, you get a clear "couldn't stop the terminals, try again" instead of a confusing timeout — and once the terminal is actually gone, deleting works.

## User-regression / trade-off list

Goal was **zero** user-facing regressions. What changed for users:

1. **None for normal deletion.** The common case (terminals exit quickly) is unchanged and still fast.
2. **No behavior change for non-deletion flows.** Tab close, sleep, daemon restart, and renderer-initiated kills pass no bound → identical 30s-default behavior and connect semantics as before.
3. **Clearer error when a backend is genuinely wedged.** Previously: misleading "Timed out waiting for physical PTY teardown" that blocked deletion. Now: accurate "Failed to physically stop every PTY for worktree", and a retry succeeds once the process is actually gone. This is strictly an improvement, not a regression.
4. **Theoretical narrow trade-off (accepted):** the bound is `deadline − now − 500ms`. If a kill legitimately took **>~9.4s** (the daemon's own physical-exit budget is 8s) it could be cut off ~0.5s earlier than the old absolute 10s deadline allowed. In practice the daemon resolves well under 8s or already rejects at its own 8s budget, so no previously-successful deletion is expected to newly fail; if it ever did, the fail-closed result is a clear "try again", not data loss.
5. **Fail-closed safety preserved.** A genuinely-alive process still blocks removal — the app never deletes files out from under a live terminal.

### DIFF
--- a/src/main/daemon/daemon-pty-adapter.ts
+++ b/src/main/daemon/daemon-pty-adapter.ts
@@ -573,10 +573,21 @@ export class DaemonPtyAdapter implements IPtyProvider {
     this.client.notify('setSessionBackground', { sessionId: id, background: safeBackground })
   }
 
-  async shutdown(id: string, opts: { immediate?: boolean; keepHistory?: boolean }): Promise<void> {
+  async shutdown(
+    id: string,
+    opts: { immediate?: boolean; keepHistory?: boolean; timeoutMs?: number }
+  ): Promise<void> {
+    // Why: destructive teardown shares one budget across connect + kill so a wedged
+    // handshake cannot burn the whole sweep deadline before the kill even starts;
+    // undefined keeps the default unbounded connect + 30s RPC for all other callers.
+    const shutdownDeadline = opts.timeoutMs !== undefined ? Date.now() + opts.timeoutMs : undefined
     // Why: shutdown can be the first lazy-client operation after restart; connect
     // before killing so a healthy daemon session is not orphaned (#7742).
-    await this.ensureConnected()
+    await this.ensureConnected(
+      shutdownDeadline !== undefined
+        ? { connectTimeoutMs: Math.max(1, shutdownDeadline - Date.now()) }
+        : undefined
+    )
     // Why: sleep/exact-stop kills the live PTY before the periodic checkpoint may run.
     // Force a final snapshot so wake can restore the pane users left.
     if (opts.keepHistory) {
@@ -606,7 +617,14 @@ export class DaemonPtyAdapter implements IPtyProvider {
         this.historyManager?.suspendSession(id)
       }
     }
-    await this.client.request('kill', { sessionId: id, immediate: opts.immediate ?? false })
+    // Why: destructive teardown passes what remains of its shared budget so a wedged
+    // daemon fails fast with an accurate error instead of tripping the outer deadline;
+    // undefined keeps the DaemonClient 30s default for all other callers.
+    await this.client.request(
+      'kill',
+      { sessionId: id, immediate: opts.immediate ?? false },
+      shutdownDeadline !== undefined ? Math.max(1, shutdownDeadline - Date.now()) : opts.timeoutMs
+    )
     this.activeSessionIds.delete(id)
     this.dirtySessionVersions.delete(id)
     if (!opts.keepHistory) {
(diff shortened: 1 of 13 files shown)


OrcaWin:
**Elegance pass (commit `26036a4`).** Reviewed the design for the cleanest possible form and refactored: instead of threading a relative `timeoutMs` and reconstructing an absolute deadline + "remaining budget" at three layers, the destructive path now threads **one absolute `deadlineMs`** through `IPtyProvider.shutdown`/`listProcesses`. Each RPC leaf converts to a relative timeout exactly once, at the moment it issues (`max(1, deadlineMs - now)`).

Why it's cleaner:
- The threaded value is a *point in time*, not a duration — matching the codebase's existing deadline-based shared-budget idiom.
- Sequential RPCs (kill → liveness verify) share the budget **structurally**: a later leaf converting the same deadline naturally gets less time, so the explicit recompute-after-shutdown bookkeeping is deleted.
- Eliminates per-layer deadline *re-anchoring* (each hop previously drifted the effective deadline slightly later).
- The 500ms margin is applied once (`teardownRpcDeadline`); the dead `: opts.timeoutMs` else-arms are gone.

Net-zero lines (136/136) but strictly less machinery. No behavior change; fail-closed safety intact; non-teardown callers still keep the 30s default. Verified: all affected suites green, typecheck clean, and tests were *strengthened* to pin exact leaf-observed budgets and the margin.

PR #9517, opened 2026-07-20
fix(settings): persist the last IME syllable when leaving the display-name field mid-composition

dogfooter8750:
## Summary

Fixes a data-loss bug where the last unconfirmed IME (e.g. Hangul) character of a project's Display Name was dropped if you navigated away from Settings mid-composition.

## ELI5

If you were typing a Korean name and left the settings screen before the last letter was "locked in," that last letter silently disappeared from the saved name. Now it's kept.

## Root cause & fix

`RepoSettingsDraftInput` intentionally holds keystrokes in local `draft` state and defers persistence until `compositionend`, because writing pre-confirmation text mid-composition cancels the IME session. But if the field blurs or unmounts before `compositionend` fires (navigating out of settings), that last syllable lived only in `draft` and never reached the store. The fix adds a guarded blur/unmount flush (`flushRef` + `lastPersistedRef`) that persists the visible draft, staying a no-op when the value was already committed.

## Evidence

- Test file `RepositoryPaneDraftInput.test.tsx` passes on the branch: **Tests 14 passed (14)**.
- Reverting only the fix file to `origin/main` (keeping the test) makes 2 tests fail, proving the tests capture the bug: **Tests 2 failed | 12 passed (14)**.
- Failing assertions when fix reverted:

```
 "flushes an abandoned IME composition on blur"
   AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times (line 245)
 "flushes an abandoned IME composition when the field unmounts"
   AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times (line 261)
```

- Lint clean: `oxlint` on the changed component exits 0 with no warnings/errors.

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression: only the abandoned-composition branch changes behavior. The flush is guarded by `lastPersistedRef` and a `repoId` check, so already-committed values are not re-persisted — covered by the "does not re-persist a confirmed value on blur or unmount" and "does not persist on blur or unmount when nothing was edited" tests. Non-affected paths (normal `compositionend`, echo, repo-switch) are byte-identical and remain passing.

_Credit to original author @KimYoungHwan8750. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/renderer/src/components/settings/RepositoryPaneDraftInput.test.tsx
+++ b/src/renderer/src/components/settings/RepositoryPaneDraftInput.test.tsx
@@ -7,19 +7,39 @@ import { RepoSettingsDraftInput } from './RepositorySettingsDraftInput'
 
 let container: HTMLDivElement
 let root: Root
+let unmounted: boolean
 
 beforeEach(() => {
   container = document.createElement('div')
   document.body.appendChild(container)
   root = createRoot(container)
+  unmounted = false
 })
 
 afterEach(() => {
+  if (!unmounted) {
+    act(() => {
+      root.unmount()
+    })
+  }
+  container.remove()
+})
+
+// Why: some tests observe the flush that runs when the field unmounts
+// mid-composition; unmount explicitly and let afterEach skip the double-unmount.
+function unmountNow(): void {
   act(() => {
     root.unmount()
   })
-  container.remove()
-})
+  unmounted = true
+}
+
+// Why: React routes onBlur through the bubbling focusout event.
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @KimYoungHwan8750 — persist the last IME syllable when leaving the display-name field mid-composition. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #8215, opened 2026-07-11
fix(terminal): open WSL file links on Windows

rodboev:
## Summary

Makes POSIX file paths printed by terminals running inside a WSL distro clickable and openable from a Windows-hosted WSL worktree.

## ELI5

When you run a terminal inside WSL on Windows, file links show up as Linux-style paths like `/root/workspace/README.md`. Windows can't open those directly, so clicking did nothing. This teaches Orca to translate them into a Windows-visible location so the file opens.

## Root cause & fix

Terminals running in a WSL distro emit POSIX absolute paths, but the Windows host cannot resolve them. The fix adds a pure `mapTerminalFilePath(filePath, worktreePath)` in `terminal-file-open-routing.ts` that rewrites a POSIX absolute path to `//wsl.localhost/<distro><path>` when the worktree parses as a WSL UNC path (both modern `\\wsl.localhost` and legacy `\\wsl$` forms). Every file-open call site — hit-testing, path-exists caching, `stat`, `authorizeExternalPath`, `openFile`, tooltip, and `pendingEditorReveal` — now routes through the mapped path. The guard returns the original path unchanged for non-WSL worktrees and for native Windows (`C:\`), existing UNC (`//`), or relative inputs, so no other platform behavior changes.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23.

- Rebase: clean onto `origin/main` (base `7ab601487c`, head `9227939c8b`).
- Test file: `terminal-link-handlers.test.ts` — passes on branch, fails when the fix is reverted (proven regression guard).
- Lint: `oxlint` clean on all 3 changed source files (no warnings/errors).

```
On branch:                Test Files 1 passed | Tests 89 passed (89)
Fix reverted (test kept): Test Files 1 failed | Tests 8 failed | 81 passed
  - "maps POSIX OSC file links for a WSL worktree before opening them"
    expected authorizeExternalPath({targetPath: //wsl.localhost/Ubuntu/root/...})
  - "preserves existing UNC and native paths": mapTerminalFilePath is not a function
  - "does not map POSIX paths for a native Windows worktree": mapTerminalFilePath is not a function
```

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression by construction: `mapTerminalFilePath` is a pure, tightly-guarded transform. Non-WSL worktrees hit the identity path (`parseWslUncPath` returns `null`), and native Windows / UNC / relative inputs are returned byte-identical — both covered by passing tests. Only the WSL branch changes behavior.

[...]

PR #9518, opened 2026-07-20
fix(terminal): reveal Markdown links at target lines

kaynansc:
## Summary

Clicking a terminal file link with a line number (e.g. `docs/guide.md:230`) now navigates to that exact line instead of opening the Markdown file at the top.

## ELI5

When you click a file link in the terminal that points to a specific line, Orca now jumps you straight to that line — even for Markdown files, which previously just opened at the beginning.

## Root cause & fix

Terminal links already parsed the `:line` suffix, but the pending line reveal is consumed by Monaco, and Markdown files opened in Orca's rich (non-Monaco) view — so nothing honored the reveal. The fix, in `terminal-file-open-routing.ts`, switches Markdown line targets to source mode via `setMarkdownViewMode(fileId, 'source')` so Monaco mounts and can navigate. It also scopes the reveal to the opened editor tab's `fileId` (resolved from `activeFileIdByWorktree[worktreeId]`, falling back to `filePath`) so owner-qualified tabs across local/SSH/runtime contexts get the reveal instead of an ambiguous path key. `terminal-file-link-conformance.ts` adds a "relative markdown path with line" parser case.

## Evidence

- Test file: `src/renderer/src/components/terminal-pane/terminal-link-handlers.test.ts`
- On branch: **83 passed (83)**.
- Reverting the two fix files re-runs to **10 failed / 73 passed** — the test captures the bug. Sample assertion: the 2nd `setPendingEditorReveal` call is missing `"fileId": "/tmp/src/second.ts"`, and the new Markdown test's `setMarkdownViewMode(fileId, 'source')` is never called.
- Lint clean: `oxlint` on both fix files reports no findings.

```
# on branch
Test Files  1 passed (1) | Tests 83 passed (83)

# fix files reverted to origin/main
Test Files  1 failed (1) | Tests 10 failed | 73 passed (83)
```

## Trade-offs

None — pure fix. Markdown links without a line target, non-Markdown files, and HTML/system-default opens are unchanged.

## Regression risk

Zero-regression: only the line-carrying branch of `openDetectedFilePath` changes, and only Markdown-with-line paths gain the view-mode switch. Non-affected paths remain byte-identical, and the 73 unrelated tests pass on both branch and reverted states. No new filesystem, IPC, or network calls; authorization and stat checks still run unchanged.

_Credit to @kaynansc for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-file-open-routing.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-file-open-routing.ts
@@ -183,12 +183,13 @@ export function openDetectedFilePath(
       activateAndRevealWorktree(worktreeId)
     }
 
+    const language = detectLanguage(mappedFilePath)
     store.openFile(
       {
         filePath: mappedFilePath,
         relativePath,
         worktreeId: worktreeId || '',
-        language: detectLanguage(mappedFilePath),
+        language,
         mode: 'edit',
         runtimeEnvironmentId
       },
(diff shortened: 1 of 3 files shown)


nwparker:
Thanks so much @kaynansc — terminal Markdown line links now navigate Monaco to the exact line. Resolved the conflict with #8215 (WSL links, landed today) preserving both behaviors, 91 tests pass, and merged. 🙏

PR #9519, opened 2026-07-20
fix(terminal): clear stale search highlights when find closes

kaynansc:
# fix(terminal): clear stale search highlights when find closes

## Summary
Fixes stale terminal find highlights that lingered after the query was erased, the active pane changed, or the search overlay unmounted.

## ELI5
When you searched in the terminal and then cleared or closed the search, the highlighted match sometimes stayed lit up. Now it disappears the moment the search stops.

## Root cause & fix
`SearchAddon.clearDecorations()` removes xterm's painted match decorations but leaves the *active* match still selected, and `TerminalSearch` never cleaned up the previous addon when its portal moved to another pane or unmounted. The fix adds a `clearTerminalSearch()` helper that pairs `clearDecorations()` with a `findNext('')` empty-find request — xterm's path for dropping the active selection — and applies it on query erase, search-closed, addon change, and portal unmount. A separate addon-lifecycle effect keyed on `searchAddon` cleans the previous pane without resetting incremental search on every keystroke.

## Evidence

- Rebase: clean onto `origin/main` (base `8a23618310`, head `b20b6ecbc2`).
- Test file `src/renderer/src/components/TerminalSearch.test.tsx` (new) — **3/3 pass on branch**.
- Reverting only the fix (`TerminalSearch.tsx`) while keeping the test → **3/3 fail**, proving the tests catch the regression:

```
On branch:  Test Files 1 passed (1); Tests 3 passed (3)
Fix reverted:
  Test Files 1 failed (1); Tests 3 failed (3)
  - "clears the current addon when the query is erased":
      expected findNext called with [''], Number of calls: 0
  - "clears the previous addon when the search moves to another pane":
      expected clearDecorations called 1 time, got 0
  - "clears the addon when the search portal unmounts":
      expected clearDecorations called 1 time, got 0
```

- Lint: `npx oxlint src/renderer/src/components/TerminalSearch.tsx` — clean (no warnings/errors).

## Trade-offs
None — pure fix. (Searching history owned internally by an alternate-screen TUI remains a separate, pre-existing limitation, since that history is not in xterm's active buffer.)

## Regression risk
Effectively zero. The change is renderer-only and scoped to the terminal-search cleanup path: cleanup runs only when search is cleared/closed, the addon changes, or the portal unmounts. Non-affected code paths are byte-identical to `origin/main`, and the empty-find call short-circuits without scanning the buffer. The three regression tests prove correct behavior across all three failure paths.

_Credit to @kaynansc for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/renderer/src/components/TerminalSearch.test.tsx
+++ b/src/renderer/src/components/TerminalSearch.test.tsx
@@ -1,0 +1,86 @@
+// @vitest-environment happy-dom
+
+import { cleanup, fireEvent, render, waitFor } from '@testing-library/react'
+import type { SearchAddon } from '@xterm/addon-search'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import TerminalSearch from './TerminalSearch'
+
+vi.mock('@/i18n/i18n', () => ({
+  translate: (_key: string, fallback: string) => fallback
+}))
+
+afterEach(cleanup)
+
+function createSearchAddon(): SearchAddon {
+  return {
+    findNext: vi.fn(() => true),
+    findPrevious: vi.fn(() => true),
+    clearDecorations: vi.fn()
+  } as unknown as SearchAddon
+}
+
+function renderSearch(searchAddon: SearchAddon): ReturnType<typeof render> {
+  return render(
+    <TerminalSearch
+      isOpen
+      onClose={vi.fn()}
+      searchAddon={searchAddon}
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @kaynansc — clear stale search highlights when find closes. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #9520, opened 2026-07-20
fix(ssh): mirror local workspace layout for SSH project creation and worktree placement

bang9:
## Summary

SSH hosts now mirror the local workspace layout for both project creation and worktree placement, fixing a cross-project branch-name collision that only existed over SSH.

**The problem.** Local and SSH behaved differently in two ways:

1. **Create project**: locally the dialog pre-fills the default parent (`~/orca/projects`); runtime hosts resolve `~` and do the same. On SSH the location field stayed empty and the user had to type or browse a host folder every time.
2. **Worktree placement**: locally (default settings: absolute `workspaceDir`, `nestWorkspaces: true`) worktrees are isolated per project under `~/orca/workspaces/<repo>/<name>`. On SSH the same settings were silently ignored — `computeRemoteWorktreePath` treated the absolute desktop `workspaceDir` as unusable and fell back to a **flat repo-sibling** path: `<repoParent>/<name>`.

Because of (2), two SSH projects that live in the same parent directory (e.g. `~/work/project-a`, `~/work/project-b`) **collide when you create same-named worktrees/branches for cross-project work**: both resolve to `~/work/feature-x`. The second creation hits the path-exists retry loop and gets renamed to `feature-x-2` — which also renames its branch — so you simply cannot keep branch names aligned across projects on SSH, while the identical flow works fine locally.

**The fix.** When the global `workspaceDir` is an absolute desktop path (the default), SSH worktree creation now resolves the remote host's home via the relay (`session.resolveHome`) and mirrors the local layout there, exactly like the existing WSL mirror:

- `nestWorkspaces: true` → `<remoteHome>/orca/workspaces/<repo>/<name>` (per-project isolation, no collisions)
- `nestWorkspaces: false` → `<remoteHome>/orca/workspaces/<name>` (same flat semantics as local)
- per-repo `worktreeBasePath` and repo-relative `workspaceDir` values behave exactly as before

And the SSH create-project dialog now resolves the remote home the same way runtime hosts do and pre-fills `<remoteHome>/orca/projects`, keeping manual entry and the host file browser available.

**Backward compatibility.**

- Existing SSH worktrees are untouched: they are tracked by their stored paths and stay `orca-managed` via persisted metadata. Only newly created worktrees use the new placement.
- Older relays that don't support `session.resolveHome` (and any home-resolution failure) degrade safely to the legacy repo-sibling placement — creation never fails because of this change. Every pre-existing test for the legacy placement passes unmodified.
- The worktree base-directory watcher now watches **both** the legacy sibling root and the new remote-home root, so old and new worktrees keep getting freshness events.
- In the create-project dialog, if the SSH home probe fails or times out (3s), the field is left empty with the existing "Choose or enter a host parent folder" hint — the pre-change UX.

**Main changes.**

- `src/main/ssh/ssh-remote-home.ts` (new): shared `resolveSshRemoteHome()` — relay-backed home resolution with per-connection caching, absolute-path/control-character validation, `null` on any failure. `remote-agent-trust-presets.ts` now reuses it instead of its private copy.
- `src/main/ipc/worktree-logic.ts`: `computeRemoteWorktreePath()` accepts `remoteHome` and mirrors the local layout under `<remoteHome>/orca/workspaces`; new `remoteWorktreePathUsesRemoteHome()` predicate and `getRemoteWorktreeCreationLayout()` so persisted creation metadata records the layout the worktree actually lives under.
- `src/main/ipc/worktree-remote.ts`: `createRemoteWorktree()` resolves the remote home once (only when the placement needs it) and threads it into path computation and creation metadata.
- `src/main/ipc/worktree-base-directory-watch-targets.ts`: SSH repos with default placement now produce both watch roots (legacy sibling + remote-home).
- `src/renderer/src/components/sidebar/useCreateProjectDefaults.ts`: the runtime-host default-parent effect is generalized to one shared effect covering runtime **and** SSH targets (SSH resolves `~` via `window.api.ssh.browseDir`); user-typed values are never clobbered.
- `src/renderer/src/components/sidebar/AddRepoCreateStep.tsx`: the "choose a host parent folder" hint now shows for any remote host, not just runtime hosts.

## Screenshots

No screenshots (SSH host required to reproduce). UI delta in words:

- **Create a new project (SSH host)**: the Location field auto-fills with `<remoteHome>/orca/projects` (previously empty); Change/Browse still work; a failed probe leaves the field empty with the existing hint.
- No other visual changes.

## Testing

- [ ] `pnpm lint` — passes for this change; fails on a **pre-existing** `switch-exhaustiveness-check` error in `src/renderer/src/components/skills/skill-freshness-group.tsx:101` that also fails on a clean checkout of `main` (verified via `git stash`).
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

New/updated tests:

- `worktree-logic.test.ts`: remote-home mirroring (nested/flat), Windows SSH home joins, null/relative/empty home fallbacks, `remoteWorktreePathUsesRemoteHome`, creation-layout recording.
- `worktrees.test.ts`: end-to-end SSH create placing the worktree at `<home>/orca/workspaces/<repo>/<name>` (asserts `session.resolveHome`, `git worktree add` target, `session.registerRoot`, persisted layout), and an explicit legacy-fallback test where the relay rejects `session.resolveHome`.
- `worktree-base-directory-watcher.test.ts`: dual-root watching (sibling + remote-home) with a nested completion event.
- `useCreateProjectDefaults.test.ts`: SSH default fill from `browseDir('~').resolvedPath`, probe failure → `failed` status with empty field, user-touched value never replaced, stale autofill cleared when switching targets.
- All pre-existing tests pass unmodified (legacy placement, ownership classification, trust presets).

## AI Review Report

An adversarial review agent (Claude Opus) read the full diff plus surrounding code paths (event filter, multiplexer lifecycle, both callers of the refactored trust-preset helper, the SSH browse handler, path-computation internals) across five dimensions: correctness/races, backward compatibility, cross-platform, security, and the shared-helper refactor.

- **1 confirmed finding (LOW), fixed before this PR**: `resolveSshRemoteHome` cached its promise per multiplexer, so a *transient* `session.resolveHome` failure would pin the legacy-sibling fallback for the connection's entire lifetime. Fixed: only definitive answers stay cached (successful responses, and "Method not found" from older relays — the same narrow-predicate pattern as `GitCapabilityCache`); transient failures evict the entry so the next call retries. In-flight dedupe is preserved.
- Verified clean: `computeRemoteWorktreePath` branch logic (per-repo base path short-circuits home resolution; the only settings-divergence case is forced onto the unchanged branch), the watcher loop rewrite (git-common classification uses only repoId, per-layout `nestWorkspaces` is honored, no drift for local/WSL repos), the merged renderer effect's generation guards and status machine (no stuck `checking` state, in-flight probes invalidated on target switch), and backward compatibility (existing worktree paths are never recomputed; both watch roots covered; trust-preset callers already treated failures as best-effort).
- **Cross-platform** (macOS/Linux/Windows explicitly checked): Windows SSH homes handled — `computeRemoteHomeWorkspaceRoot` picks win32/posix ops via `isWindowsAbsolutePathLike`, the browse handler emits forward-slash `resolvedPath` under PowerShell, and `getDefaultCreateProjectParent` derives its separator from the input path. No hardcoded separators introduced; no shortcut/label/shell/Electron platform surface touched.

## Security Audit

- The relay-resolved home is validated before use: must be `/`-rooted or drive-rooted, NUL/CR/LF rejected (`hasRemotePathControlCharacter`), trailing slash normalized. It then flows into git as structured argv (`git worktree add` via the relay's arg-array exec, never shell interpolation) and into `session.registerRoot` as a plain parameter — no new injection surface, and the path is the SSH user's own home on their own host.
- The renderer's new call is the pre-existing `window.api.ssh.browseDir` IPC (already exposed for the host file browser) pointed at `~`; no new IPC channel, no privilege change, 3s renderer-side timeout on top of the handler's own 15s cap.
- No secrets, auth, or dependency changes. Worktree paths registered with the relay are exactly the paths git will operate on, same as before.

## Notes

- Placement decision (confirmed with the repo owner): mirror the local layout under the remote home for both `nestWorkspaces` on **and** off, with legacy sibling only as a fallback for old relays. The remote home is a safe default because SSH login guarantees a home directory and the relay already runs from it; restricted accounts fall back to the legacy path if home resolution fails.
- `getWorktreeCreationLayout` is still used for local/runtime creation; only the SSH path records the remote layout.
- The pre-existing `skill-freshness-group.tsx` lint error is unrelated to this PR and reproduces on `main`.

## ELI5

Creating projects and worktrees over SSH did not match local layout, so empty default folders and flat worktree paths caused branch-name collisions across projects. SSH now uses the same parent defaults and per-project worktree placement as local.

### DIFF
--- a/src/main/ipc/worktree-base-directory-watch-targets.ts
+++ b/src/main/ipc/worktree-base-directory-watch-targets.ts
@@ -16,10 +16,12 @@ import {
 import { isWslUncPath } from '../../shared/wsl-paths'
 import { getSshFilesystemProvider } from '../providers/ssh-filesystem-dispatch'
 import {
+  computeRemoteHomeWorkspaceRoot,
   computeWorkspaceRoot,
   getWorktreePathSettings,
   hasRepoWorktreeBasePath
 } from './worktree-logic'
+import { resolveSshRemoteHome } from '../ssh/ssh-remote-home'
 import { shouldEmitBoundedWarning } from './bounded-warning-dedupe'
 import { resolveWorktreeCommonGitDirectory } from './worktree-common-git-directory'
 import type {
@@ -99,25 +101,33 @@ function isRuntimePathAbsoluteForRepo(repoPath: string, pathValue: string): bool
   return isRuntimePathAbsolute(pathValue, pathFlavor)
 }
 
-function getBaseWatchLayout(
+async function getBaseWatchLayouts(
   repo: Repo,
   pathSettings: Pick<GlobalSettings, 'workspaceDir' | 'nestWorkspaces'>,
   connectionId: string | undefined
-): { workspaceRoot: string; nestWorkspaces: boolean } {
+): Promise<{ workspaceRoot: string; nestWorkspaces: boolean }[]> {
   if (
     connectionId &&
     !hasRepoWorktreeBasePath(repo) &&
     isRuntimePathAbsoluteForRepo(repo.path, pathSettings.workspaceDir)
   ) {
-    // Why: SSH creates default worktrees beside the remote repo when the
-    // global workspace dir is a desktop-local absolute path.
-    return { workspaceRoot: resolveRuntimePath(repo.path, '..'), nestWorkspaces: false }
+    // Why: SSH default placement mirrors the workspace layout under the remote
+    // home when the relay can resolve it, while worktrees created before that
+    // (or via older relays) sit beside the remote repo — watch both roots.
+    const layouts = [{ workspaceRoot: resolveRuntimePath(repo.path, '..'), nestWorkspaces: false }]
+    const remoteRoot = computeRemoteHomeWorkspaceRoot(await resolveSshRemoteHome(connectionId))
+    if (remoteRoot) {
+      layouts.push({ workspaceRoot: remoteRoot, nestWorkspaces: pathSettings.nestWorkspaces })
+    }
+    return layouts
   }
 
-  return {
-    workspaceRoot: computeWorkspaceRoot(repo.path, pathSettings),
-    nestWorkspaces: pathSettings.nestWorkspaces
-  }
+  return [
+    {
+      workspaceRoot: computeWorkspaceRoot(repo.path, pathSettings),
+      nestWorkspaces: pathSettings.nestWorkspaces
+    }
+  ]
 }
 
 async function maybeAddBaseTarget(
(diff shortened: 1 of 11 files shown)


Jinwoo-H:
Thanks @bang9 for the investigation and implementation. We’re closing this PR because we’re consolidating the remaining SSH workspace-layout parity bug into the owner-maintained issue https://github.com/stablyai/orca/issues/15320. The original same-name collision is superseded by newer path qualification; we’ll carry forward the remaining default SSH workspace-directory requirements against current main. Credit is preserved for @bang9 and the Claude co-author. No further action is needed from you.

PR #9522, opened 2026-07-20
fix(runtime): make remote terminal lifecycle host-owned

OrcaWin:
## Summary

Make the remote host the sole owner of destructive terminal lifecycle decisions.

A paired renderer is now a mirror: disconnecting, unmounting, switching servers, or observing a stale PTY exit cannot kill the host PTY. An intentional ready-tab close carries an exact generation-bound terminal handle, and main revalidates the PTY/worktree/tab/leaf/generation immediately before committing the durable retirement and tearing down the provider.

This is a fresh replacement design for the lifecycle problem discussed in #8872. It does not modify that contributor branch.

## ELI5 user flow

- You open a terminal running on another Orca runtime.
- Your local Orca window shows a mirror of that terminal.
- If the mirror disappears because you disconnect or switch runtimes, the real terminal keeps running.
- If you intentionally click close, Orca presents the terminal's exact current identity to the host.
- The host checks that identity again at the last possible moment. It closes only that terminal; a stale mirror, replacement process, pinned tab, or shared sibling is rejected safely.

## Design

- Viewer disconnect/unmount/server-focus changes only prune local mirror state.
- Explicit ready-tab close uses a generation-bound terminal handle.
- Main validates exact PTY, worktree, tab, leaf, and lifecycle generation after renderer validation.
- Durable retirement flushes before provider teardown.
- Renderer performs only guarded mirror pruning after the host transaction succeeds.
- Runtime protocol v4 fences destructive old/new client-server skew.
- Shared PTYs and snapshot-only SSH aliases are protected from unrelated leaf retirement.

## Verification

Deterministic checks on the final rebased commit:

- Runtime/shared: 4 files, 798 tests passed.
- Renderer/main-window plus adjacent upstream SSH-ownership controls: 8 files, 686 tests passed.
- Mobile protocol compatibility: 1 file, 2 tests passed.
- Mobile typecheck and formatting passed.
- Targeted lint and max-lines ratchet passed.
- Full desktop typecheck/build passed.
- `git diff --check` passed.

Real paired Electron + `orca serve` validation through the encrypted runtime transport:

- protocol v4 pairing and real durable remote terminals;
- disconnect/switch-local preserved host shells and children;
- reconnect/switch-back restored terminal close controls;
- explicit close killed only the exact target, preserving its sibling;
- exact tab/leaf rebind rejected the old handle with `terminal_handle_stale`;
- natural PTY exit retired only its exact leaf;
- pinned close returned `terminal_tab_pinned` and preserved both process trees;
- guarded cleanup removed the remaining tabs and PTYs.

After rebasing onto the latest adjacent SSH-ownership change, the exact final artifacts were rebuilt and a publication smoke repeated protocol pairing, live PTY disconnect survival, ready-state reconnect, pinned rejection, and guarded explicit close.

## Regression considerations

- Protocol v4 deliberately rejects destructive version-skew combinations that cannot distinguish automatic lifecycle events from explicit user intent.
- Terminal tab close now spans renderer validation, main revalidation, durable state, and provider teardown; the tests cover stale generations and validation/commit races.
- Shared PTY and SSH snapshot aliases must not let one leaf retire another owner; focused ownership tests cover these cases.
- Windows/WSL paired-runtime verification is requested as an additional platform acceptance pass before merge.

### DIFF
--- a/mobile/app/h/[hostId]/session/[worktreeId].tsx
+++ b/mobile/app/h/[hostId]/session/[worktreeId].tsx
@@ -4099,7 +4099,8 @@ export default function SessionScreen() {
     try {
       const response = await client.sendRequest('session.tabs.close', {
         worktree: `id:${worktreeId}`,
-        tabId: tab.id
+        tabId: tab.id,
+        intent: 'user'
       })
       if (response.ok) {
         if (tab.type === 'terminal' && typeof tab.terminal === 'string') {
--- a/mobile/src/transport/protocol-compat.test.ts
+++ b/mobile/src/transport/protocol-compat.test.ts
@@ -2,6 +2,15 @@ import { describe, expect, it } from 'vitest'
 import { evaluateCompat } from './protocol-compat'
 
 describe('evaluateCompat', () => {
+  it('allows the current mobile app to connect to the ownership-fenced desktop', () => {
+    expect(
+      evaluateCompat({
+        desktopProtocolVersion: 4,
+        desktopMinCompatibleMobileVersion: 4
+      })
+    ).toEqual({ kind: 'ok' })
+  })
+
   it('allows the current mobile app to connect before a protocol-2 desktop updates', () => {
     expect(
       evaluateCompat({
--- a/mobile/src/transport/protocol-version.ts
+++ b/mobile/src/transport/protocol-version.ts
@@ -16,5 +16,7 @@
 // server feature added at a specific runtime protocol version. This
 // triggers a hard-block screen for users paired to older servers.
 
-export const MOBILE_PROTOCOL_VERSION = 3
+// Why: desktop protocol 4 fences clients that could confuse mirror teardown
+// with explicit terminal close; mobile only sends close from explicit UI actions.
+export const MOBILE_PROTOCOL_VERSION = 4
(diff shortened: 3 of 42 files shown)


AmethystLiang:
Related to #9952 (sessions vanishing over remote Orca link).

OrcaWin:
Closing this as superseded by the host-authoritative terminal lifecycle work that has since landed on main in #9687, #9804, and #9994.

The current PR head (`b39dff7adb`) is now conflicting and duplicates/replaces older versions of that machinery. It also raises the runtime compatibility floor to protocol 4, which is no longer wanted: main keeps the additive lifecycle RPC model at protocol 3 / minimum 2 so older clients and servers retain their previous connectivity.

We reproduced the original destructive flow on pre-fix main with a real encrypted paired client and real serve-runtime PTY. On exact `41751dd90d` (#9994), the same live audit passed: renderer disconnect preserved the host PTY and child processes, reconnect restored a stable mirror, and explicit UI close retired exactly the intended PTY. The focused lifecycle suites passed 175/175 and the desktop build passed.

Any remaining generation-bound hardening for explicit user closes should be a small follow-up PR on current main, with an old-host fallback only for `method_not_found`, rather than reviving this broad implementation.

PR #9513, opened 2026-07-20
fix(ui): restore missing spaces in i18n JSX fragment copy

nwparker:
## Summary

Follow-up to #9511. The same glue-word class of bug exists wherever `translate(...)` fragments sit next to `<code>` / dynamic values without an explicit `{' '}`.

Examples of broken copy this fixes:

- `This repo ignores sharedorca.yaml setup scripts`
- `Exampleorca.yaml template` / `Editorca.yaml to change`
- `main ormaster` / `commands likegit diff main...HEAD`
- `Runorca serve --pairing-address …`
- `Showing8 of42`
- `Force-quitssess_abc123`
- `Version1.2.3 is available`
- and many more label/value pairs (`Started…`, `Renamed from…`, `Filter pull requests`, rate limits, etc.)

### Approach

- Insert explicit JSX `{' '}` between word-ending fragments and the next token (same pattern as the trust dialog fix).
- Leave intentional glue alone (`check #123`, `comment L42`, `Clear (` + count + `)`).
- Add `i18n-jsx-spacing-guard.test.ts` to lock high-impact cases.

### Critical paths (orca.yaml / setup / git / run)

hero before-after

### Full before/after grid (all fixed cases)

full evidence grid

## Test plan

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/lib/i18n-jsx-spacing-guard.test.ts`
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/sidebar/OrcaYamlTrustDialog.test.tsx`
- [ ] Spot-check Settings → Git / Repository hooks / Runtime environments copy
- [ ] Spot-check setup script prompt when shared orca.yaml is ignored

PR #9524, opened 2026-07-20
fix(i18n): use explicit {' '} so PRFilterSections satisfies the JSX spacing guard

OrcaWin:
## Description

`main` is currently **red** on `i18n-jsx-spacing-guard.test.ts` → *keeps an explicit space after "Filter pull requests" in `components/github/PRFilterSections.tsx`*.

**Root cause:** #9513 (`0d97653e3`) added that spacing-guard test **and** restored the `Filter <subject>` space in the same commit — but it used a **bare literal JSX space** where the guard requires the **explicit `{' '}`** form:

```jsx
{translate(…, 'Filter')} {subject}      // shipped in #9513 — renders fine, guard rejects
{translate(…, 'Filter')}{' '}{subject}  // this PR — same output, guard passes
```

So #9513 merged self-inconsistent. It stayed latent because the `verify` suite only runs on `pull_request`, not on pushes to `main`, so the redness only surfaces on the next PR that runs CI.

This is a one-line change: swap the bare space for `{' '}`. **Identical rendered output**, and it turns the guard green.

## Evidence

- The failing run (unrelated PR #9515): `Test Files 1 failed | 3038 passed` — the sole failure was this guard case.
- `git show 0d97653e3 -- .../PRFilterSections.tsx` shows the bare-space edit; the guard's regex (`i18n-jsx-spacing-guard.test.ts:118-127`) requires `)}` followed by `{' '}`.
- No behavior change: `{expr} {expr}` and `{expr}{' '}{expr}` render the same single space.

## User regression / tradeoff list

**Zero.** Pure DX/CI fix — same DOM output, no runtime or visual change. Not enabling auto-merge; please review and merge once its CI is green.

### DIFF
--- a/src/renderer/src/components/github/PRFilterSections.tsx
+++ b/src/renderer/src/components/github/PRFilterSections.tsx
@@ -208,7 +208,7 @@ export function SectionMenu({
   return (
     <div className="py-1 text-xs">
       <div className="px-3 py-1.5 text-[10px] font-medium uppercase tracking-wide text-muted-foreground">
-        {translate('auto.components.github.PRFilterSections.8177eda37e', 'Filter')} {subject}
+        {translate('auto.components.github.PRFilterSections.8177eda37e', 'Filter')}{' '}{subject}
       </div>
       {rows.map((row) => (
         <button


OrcaWin:
Closing as redundant: the JSX spacing-guard failure for `PRFilterSections.tsx` ("Filter") is already fixed on `main` by #9513 (`0d97653e3` — "restore missing spaces in i18n JSX fragment copy"), which added the explicit `{' '}` after the `translate(..., 'Filter')` fallback. No further change needed.

PR #9511, opened 2026-07-20
fix(ui): restore spaces around orca.yaml in trust dialog

nwparker:
## Summary

- Fix missing spaces in the `orca.yaml` setup-hook trust dialog so the description and "Always trust" checkbox no longer glue words together (`repository'sorca.yaml`, `trustorca`, `Always trustorca.yaml inorca`).
- Explicit JSX whitespace between `translate` fragments (same pattern as the already-correct "script changed" path).
- Trim Spanish locale trailing spaces that would double once JSX owns spacing.
- Add unit coverage for first-run and re-approval copy.

## Screenshots

| Before | After |
| --- | --- |
| before | after |

**Before:** `This repository'sorca.yaml` / `trustorca` / `Always trustorca.yaml inorca`  
**After:** `This repository's orca.yaml` / `trust orca` / `Always trust orca.yaml in orca`

## Test plan

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/sidebar/OrcaYamlTrustDialog.test.tsx`
- [ ] Manually open a workspace create flow that shows the orca.yaml trust dialog and confirm:
  - Description: `This repository's orca.yaml runs... Only run if you trust <repo>.`
  - Checkbox: `Always trust orca.yaml in <repo>`

PR #9526, opened 2026-07-20
fix(new-workspace): ignore IME composition Enter in workspace name field

### LINKED ISSUE #9525: [Bug]: Create-worktree name field duplicates the last Hangul syllable on Enter (missing isComposing guard)
### Operating system

macOS

### Orca version

v1.4.146

### Details

Short summary: In the create-worktree/workspace dialog, typing a CJK name (e.g. Korean) and pressing Enter while the last syllable is still composing duplicates that syllable. Typing `배포` and pressing Enter produces a workspace named `배포포`.

Reproduction steps:

1. macOS with the Korean (2-Set) IME active.
2. Open the create-worktree dialog (⌘N) and use the **Name** field.
3. Type `테스트` and press Enter **before the last syllable's composition is finalized** (immediately after typing, while the syllable is still underlined).
4. The field now reads `테스트트` — the final syllable is committed twice. Creating the workspace persists the wrong name (and the derived branch/worktree name).

Expected: the Enter that finalizes the IME composition should only commit the composition, like it already does in the worktree inline-rename field. A subsequent Enter can then advance focus/submit.

Root cause (from reading the source): the Enter handler in `src/renderer/src/components/new-workspace/SmartWorkspaceNameField.tsx` runs `onPlainEnter()` (which moves focus to the Agent combobox) even when the keydown is an IME-composition Enter. Moving focus mid-composition makes Chromium re-commit the composing character into the controlled input. The codebase already has a shared guard for exactly this — `isImeCompositionKeyDown()` in `src/renderer/src/lib/ime-composition-keyboard-event.ts`, used by `WorktreeTitleInlineRename` and other Enter handlers — but the smart workspace-name field doesn't use it.

Related: #8038 is the same missing-guard family in the desktop terminal. Tab (native focus traversal) does **not** reproduce the bug — only Enter, which matches the JS-initiated focus move being the trigger.

Screen recording of the repro is attached on the fix PR (opening shortly). I have a fix with regression tests ready and will link it to this issue.

shinju4n:
Fixes #9525

## Summary

Typing a CJK workspace name (Korean/Japanese/Chinese) in the create-worktree dialog and pressing Enter while the last syllable is still composing duplicates that syllable — `배포` becomes `배포포`, and the wrong name flows into the created workspace/branch.

**Root cause:** the Enter handler in `SmartWorkspaceNameField` treats the IME-composition Enter (the keypress that only finalizes the candidate) as a real Enter and runs `onPlainEnter()`, which moves focus to the Agent combobox. Moving focus mid-composition makes Chromium re-commit the composing character into the controlled input, appending it a second time.

**Fix:** guard the Enter branch with the existing shared helper `isImeCompositionKeyDown()` (`nativeEvent.isComposing || keyCode === 229`) — the exact pattern this codebase already uses in `WorktreeTitleInlineRename`, `AddRepoRemoteStep`, `AddRepoCloneStep`, `JiraIssueWorkspace`, and the TaskPage Enter handlers. The smart workspace-name field was the one Enter-commit path missing it. After the fix, the first Enter finalizes the composition (field keeps the correct text, focus stays put) and the next Enter advances focus as before — matching the inline-rename field's behavior.

Related: #8038 tracks the same missing-guard family in the desktop terminal. Tab/native focus traversal was verified unaffected (the browser-native focus move lets the IME finalize correctly; only the JS-initiated focus move triggered the re-commit).

## Screenshots

Before (Orca v1.4.146 stable, macOS Korean 2-Set IME) — typing `테스트`, pressing Enter mid-composition yields `테스트트`:

https://github.com/user-attachments/assets/0826bffe-45dc-42ed-83af-e815bedccc90

After (this branch, dev build, same IME) — same input, name stays `테스트`, focus advances on the following Enter:

https://github.com/user-attachments/assets/242aed22-f546-45c6-9c3f-cec32a44f2c3

## Testing

- [x] `pnpm typecheck` — clean (node, cli, web)
- [x] `pnpm test` — 32,326 passed, 56 skipped, 1 baseline-only failure (see below)
- [x] `pnpm lint` — my files clean; 2 baseline-only errors (see below)
- [ ] `pnpm build` — fails locally on an environment limitation only (see below)
- [x] Added `SmartWorkspaceNameField.ime-enter.test.tsx` (happy-dom render of the real component):
  - Enter with `isComposing: true` → `onPlainEnter` NOT called
  - Enter with `keyCode === 229` (IMEs that skip `isComposing` on keydown) → `onPlainEnter` NOT called
  - plain Enter → `onPlainEnter` called once (no regression)
  - Verified the two guard tests **fail on `main` and pass with this change**.
- [x] Manual verification on a local dev build with the macOS Korean IME (recording above): duplication gone, Enter-to-advance-focus still works, popover row selection via Enter unaffected for non-IME input.

Baseline-only failures — this branch differs from `main` by exactly the two files in this diff (`git diff main --name-only`), so all of the following reproduce identically on clean `main`:

- lint: `switch-exhaustiveness-check` in `src/main/daemon/daemon-server.ts:680` and `src/renderer/src/components/skills/skill-freshness-group.tsx:101` (the latter already noted as baseline in #9500).
- test: `src/renderer/src/lib/i18n-jsx-spacing-guard.test.ts` — "keeps an explicit space after 'Filter pull requests' in components/github/PRFilterSections.tsx" (same family as the recent #9511/#9513 spacing fixes).
- build: `native/computer-use-macos` requires `swift-tools-version: 6.0`; my local toolchain is Swift 5.10, so the native step fails before reaching any code in this diff.

## AI Review Report

Reviewed the diff with my AI coding agent. Main risks checked and results:

- **Cross-platform (macOS/Linux/Windows):** the guard uses standard DOM `KeyboardEvent.isComposing` plus the `keyCode === 229` fallback; no platform branches, no shortcuts/labels/paths/shell involvement. Behavior on Linux (IBus) and Windows (IME) IMEs follows the same keydown contract; non-IME input is unaffected because `isComposing` is false and `keyCode` is the real key.
- **SSH/remote/local:** renderer-only change in a local input handler; no daemon/relay/SSH surface touched.
- **Agents/integrations/git providers:** the guard sits before both Enter branches (source-row select and plain-Enter), so GitHub/GitLab/Linear/Jira source selection behavior is unchanged for non-composition Enter; no provider-specific logic added.
- **Performance:** one boolean check on Enter keydown; no hot-path impact.
- **UI quality:** no visual change; keyboard flow now matches the already-fixed inline-rename field.
- Flagged and accepted: an IME-composition Enter now does nothing beyond finalizing the composition (previously it advanced focus with corrupted text) — this is the intended, standard behavior for CJK inputs and matches `WorktreeTitleInlineRename`.

## Security Audit

- No new dependencies; no changes to IPC, network, filesystem, process spawning, or credential handling.
- Change is limited to interpreting a keyboard event in the renderer; no user data crosses any new boundary.

## Notes

- The same `isImeCompositionKeyDown` helper is intentionally reused rather than a new predicate, keeping IME-Enter semantics consistent across the app.
- No platform-specific code added; manually verified on macOS (the platform where the bug was reported).

### DIFF
--- a/src/renderer/src/components/new-workspace/SmartWorkspaceNameField.ime-enter.test.tsx
+++ b/src/renderer/src/components/new-workspace/SmartWorkspaceNameField.ime-enter.test.tsx
@@ -1,0 +1,165 @@
+// @vitest-environment happy-dom
+
+import React, { act } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import SmartWorkspaceNameField from './SmartWorkspaceNameField'
+
+vi.mock('@/store', () => {
+  const state = {
+    repos: [],
+    addRepo: vi.fn(),
+    checkLinearConnection: vi.fn(),
+    fetchWorkItems: vi.fn(),
+    fetchWorkItemsAcrossRepos: vi.fn(),
+    getCachedWorkItems: vi.fn(() => null),
+    linearStatus: { connected: false },
+    linearStatusChecked: false,
+    listLinearIssues: vi.fn(),
+    preflightStatus: null,
+    preflightStatusChecked: false,
+    preflightStatusContextKey: null,
+    refreshPreflightStatus: vi.fn(),
+    searchLinearIssues: vi.fn(),
+    settings: null
+  }
+  const useAppStore = (selector: (s: typeof state) => unknown): unknown => selector(state)
+  useAppStore.getState = () => state
+  return { useAppStore }
+})
+
+vi.mock('@/lib/local-preflight-context', () => ({
+  getLocalPreflightContext: () => ({}),
+  localPreflightContextKey: () => 'test-preflight-context'
+}))
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string, fallback?: string) => fallback ?? key })
+}))
+
+vi.mock('@/i18n/i18n', () => ({
+  translate: (_key: string, fallback: string) => fallback
+}))
+
+vi.mock('@/components/ui/popover', () => ({
+  Popover: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  PopoverAnchor: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  PopoverContent: ({ children }: { children: React.ReactNode }) => <div>{children}</div>
+}))
+
+vi.mock('@/components/ui/command', () => ({
+  Command: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  CommandGroup: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  CommandItem: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  CommandList: ({ children }: { children: React.ReactNode }) => <div>{children}</div>
+}))
+
+vi.mock('@/components/ui/tabs', () => ({
+  Tabs: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  TabsList: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  TabsTrigger: ({ children }: { children: React.ReactNode }) => <div>{children}</div>
+}))
+
+vi.mock('@/components/ui/tooltip', () => ({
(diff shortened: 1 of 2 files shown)


AmethystLiang (approved):

AmethystLiang:
Thanks for the PR! PR approved and landing soon. Will be in the next release. Usually within 24 hours. 

Can also get on the RC updates (press "shift" when click "check update") if you need it sooner :-)

Issue #6960, opened 2026-06-30
Returning to a worktree duplicates a busy, backgrounded Codex tab via `codex resume`


ariavasulin:
### Summary

When a Codex agent is **actively running** in a **backgrounded** terminal tab (i.e. another tab in the same workspace is focused), leaving that worktree and returning to it makes Orca open a **second tab** running `codex resume <session-id>` for the session that is already alive. Result: two tabs attached to the same Codex rollout/session.

This is the warm-reattach duplication class that the `origin` field + ownership check were added to prevent — see the comment that references stablyai/orca#5232 in `src/shared/agent-session-resume.ts:53-58`. The ownership guard has a hole for panes that are **alive but not the active tab**.

### Environment

* Orca **v1.4.110** (reproduced).
* Confirmed still present on `main` / **v1.4.111-rc.3**: `src/renderer/src/lib/sleeping-agent-pane-ownership.ts` is byte-identical (blob `c1f30108`) between v1.4.110 and latest `main`, so the predicate below is unchanged.
* Agent: `codex`. No user command override (`agentCmdOverrides` empty) — the `resume` command is emitted by Orca itself.

### Steps to reproduce

1. In **Workspace A**, have two terminal tabs in the same pane/group: a **Codex** tab and another tab (e.g. Claude).
2. Start a long-running Codex task so Codex is **busy (not idle)**.
3. Click the **other tab** so the Codex tab is **backgrounded** (Codex keeps running).
4. Switch to **Workspace B**.
5. Switch **back to Workspace A**.

**Expected:** the still-running Codex session is left as-is (one tab).<br>**Actual:** a new tab appears running `codex resume <id>` for the same session → two tabs, one Codex session, both alive.

Notes that pin the trigger:

* Does **not** happen if Codex is **idle** (`done`) when you switch away.
* Does **not** happen if the **Codex tab itself is the focused tab** when you switch away — only the focused tab is spared.

### Root cause

On activation, `activateAndRevealWorktree()` runs `setActiveWorktree()` (which restores the worktree's previously-**focused** tab) and then `resumeSleepingAgentSessionsForWorktree(worktreeId)`:

* `src/renderer/src/lib/worktree-activation.ts:317` then `:353`

`resumeSleepingAgentSessionsForWorktree` only skips relaunching a record when its pane is judged "owned by a preserved pane":

[...]

PR #6945, opened 2026-06-30
Prevent repeated automatic agent resume tabs

brennanb2025:
## Summary

Prevents repeated automatic recovery of the same sleeping Codex provider session from creating a new terminal tab on every workspace activation. A valid stranded session can still recover once, but exact same-session replays are cleared while that resume is queued, starting, or live.

This does not change explicit user-triggered session-history / AI Vault resumes; users can still intentionally open multiple tabs from those surfaces.

## Design approach

The regression was traced to #6800's stale-terminal recovery path: activation could treat a replayed sleeping-session record as recoverable again after the first recovery tab had already been queued. The fix adds runtime-only automatic resume claims keyed by tab id, and also stores provider-session metadata on automatic resume startup payloads. Resume suppression is scoped by worktree, agent, tab membership, and provider-session key/id.

Runtime claims are intentionally not persisted or hydrated. They are cleaned when tabs close, when orphan terminal state is reconciled, and when worktrees are purged/deleted.

## Review outcomes

- Design review completed clean after tightening the design around the startup-in-progress gap.
- Completeness verification initially found a lifecycle cleanup gap; that was fixed and re-verified clean.
- Headline behavior status: implemented.
- Broad app consistency: desktop workspace activation, terminal recovery tabs, sleeping/live agent status, session-history resumes, hydration, and cleanup paths were compared. The PR fixes automatic activation replay while preserving #6800 stale-terminal recovery and explicit user resume behavior.
- Perf audit: clean. The new O(N) checks run only during activation recovery, introduce no polling/subprocess churn, and avoid persistence bloat.
- Code-review loop: 2 rounds. Final reviewers returned clean. Findings fixed included a stale live-status tab-membership guard and removal of an unused slice API.

## Validation

- Electron culprit reproduction on `a27278dbe` confirmed the bad queued command shape: `codex --dangerously-bypass-approvals-and-sandbox resume '019f163b-c91b-72c3-9786-7b8d74ac7a3d'`.
- Electron validation on this branch verified app identity for `skua @ brennanb2025/fix-workspace-new-tab` and exercised the replay path.
- Fixed-branch replay result: first automatic recovery launched 1 tab; queued replay launched 0; startup-in-progress replay launched 0; final tab count stayed stable; stale sleeping records were cleared; the runtime claim remained on the resumed tab.
- Merge-confidence validation also clicked the visible workspace activation path in Electron and observed one restored terminal only.
- Screenshot evidence captured locally:
[...]

PR #9531, opened 2026-07-20
fix(pi): prevent duplicate resumes for live background panes

PeraSite:
## Summary

Prevent Orca from launching a second `pi --session <transcript>` tab when the original Pi pane is backgrounded but its PTY is still alive.

Pi reports `done` after a turn while its TUI process remains live. That means the existing active-status provider-session claim no longer protects an idle Pi pane. If that pane is also hidden during worktree activation, `paneWillConnectOnActivation()` returns false and Orca can treat the live session as unowned, creating a duplicate `Pi ready` tab attached to the same transcript.

This change treats a preserved stable pane as owning recovery when its exact leaf PTY is present in the live `ptyIdsByTabId` map. The check is leaf-specific, so a sibling split-pane PTY cannot claim the session. If the exact PTY is gone, the existing automatic Pi resume path remains unchanged.

This is a Pi-specific reproduction of the live-background ownership class discussed in #6960 and hardened in #6945, exposed by Pi session resume support from #8876.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — touched-file `oxlint` and `oxfmt --check` pass. The full command reaches `lint:switch-exhaustiveness`, then fails on two unrelated current-`main` errors in `skill-freshness-group.tsx` and `daemon-server.ts`; both reproduce from a detached clean `upstream/main` worktree at `2e67af82d`.
- [x] `pnpm typecheck`
- [ ] `pnpm test` — 32,357 passed, 56 skipped, 5 unrelated failures. Three failures pass when rerun in isolation; the remaining Git merge/hook fixture failures reproduce unchanged on detached clean `upstream/main` at `2e67af82d`.
- [x] `pnpm build`
- [x] Added a regression test that reproduces the duplicate launch on the parent implementation (`launched === 1`) and passes with this fix (`launched === 0`), while retaining the existing dead-PTY resume and sibling-split-PTY coverage.

Focused validation:

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/lib/resume-sleeping-agent-session.test.ts src/renderer/src/lib/pi-session-resume-wake.test.ts` — 32/32 passed.
- Touched-file `oxlint`, `oxfmt --check`, and `git diff --check` passed.

## AI Review Report

The review traced provider-session ownership through `activeOrQueuedResumeClaimsProviderSession()`, `recordPaneIsOwnedByPreservedPane()`, `paneWillConnectOnActivation()`, and `resumeSleepingAgentSessionsForWorktree()`.

Main risk checked: broadening ownership based only on tab-level PTY presence could suppress legitimate recovery or let a sibling split pane claim the wrong provider session. The implementation therefore requires the PTY mapped to the record's exact stable leaf to also appear in that tab's live-PTY map. Existing coverage confirms a sibling PTY does not claim the record and a mapped-but-not-live PTY still resumes normally.

Cross-platform review covered macOS, Linux, and Windows. The change compares opaque in-memory PTY identifiers and does not alter shortcuts, labels, filesystem paths, shell quoting, resume argv, IPC, Electron main-process behavior, or local/SSH/runtime PTY identifier construction. Remote identifiers are compared in the same renderer namespace already used by the layout and live-PTY maps.

## Security Audit

No new input, command execution, path handling, authentication, secret, dependency, or IPC surface is introduced. The change only compares existing store identifiers. Pi transcript paths remain handled by the existing validated/quoted resume path and are not inspected or executed by this ownership check.

The main availability risk was falsely retaining a dead session. Requiring exact leaf-to-live-PTY membership keeps dead panes on the existing fresh-resume path.

## Notes

- Automatic Pi resume is preserved when the original PTY is absent.
- No persisted schema or migration changes.
- No new issue was opened because #6960 already documents the same background-pane ownership gap; this PR is the Pi idle-status variant introduced after #8876.
- X handle: N/A

### DIFF
--- a/src/renderer/src/lib/pi-session-resume-wake.test.ts
+++ b/src/renderer/src/lib/pi-session-resume-wake.test.ts
@@ -3,11 +3,14 @@ import { join } from 'node:path'
 import { afterEach, describe, expect, it } from 'vitest'
 import type { SleepingAgentSessionRecord } from '../../../shared/agent-session-resume'
 import type { TerminalTab } from '../../../shared/types'
+import { makePaneKey } from '../../../shared/stable-pane-id'
 import { useAppStore } from '@/store'
 import { resumeSleepingAgentSessionsForWorktree } from './resume-sleeping-agent-session'
 
 const initialAppStoreState = useAppStore.getState()
 const PI_TRANSCRIPT_PATH = join(tmpdir(), 'pi-session-1.jsonl')
+const LEAF_ID = '11111111-1111-4111-8111-111111111111'
+const OTHER_LEAF_ID = '22222222-2222-4222-8222-222222222222'
 
 afterEach(() => {
   useAppStore.setState(initialAppStoreState, true)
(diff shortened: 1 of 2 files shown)


PeraSite:
Superseded by #9532. This duplicate was opened during branch cleanup; sorry for the noise.

PR #9540, opened 2026-07-20
feat(sidebar): reveal a rename action on worktree card hover

Jaeyoung22:
## Summary

Renaming a workspace from the sidebar already worked three ways, but none of them is visible:

1. **Double-clicking the title** — the inline editor (`WorktreeTitleInlineRename.tsx`). No hint it exists.
2. **Right-click → "Update"** — opens the heavier `edit-meta` modal focused on the name. Reasonably labeled "Update", since that modal also edits the linked issue, PR, and comment — but it means the menu has no plain "rename".
3. **The `workspace.rename` shortcut** — `Mod+Alt+R` on macOS, and `[]` on Linux and Windows, where `keybindings.ts` documents that the safe chords are already taken.

Meanwhile the title row has no ordinary hover chrome at all: the one quick action living there (delete) is deliberately gated behind holding Option/Alt because it is destructive.

This adds a pencil button at the right end of the worktree card's title row, revealed on hover, that opens the **existing** inline editor. It dispatches the same `setRenamingWorktreeId` request the shortcut uses, so save, cancel, trim/no-op handling, the saving spinner, IME composition guarding, and error toasts all stay owned by `WorktreeTitleInlineRename` — no second rename path to keep in sync.

Ordering in the action group is `Star (indicator) → pencil (ordinary action) → trash (destructive, Alt-gated)`.

### Layout note

The action collapses to **zero width** at rest rather than only fading out. `showHeaderActions` previously rendered only for a compact-mode main worktree or while Alt was held, so most rows had no action group at all; a plain `opacity-0` child would still contribute its own width, the group's flex `gap-1`, and its trailing `pr-1.5`, shifting the details/ports lane on every row. The button therefore collapses `max-w`, cancels the flex gap when a sibling precedes it, and the group defers its trailing gutter to hover. This mirrors the existing `repo-header-action-button-class.ts` reveal pattern, which exists for the same reason.

## Screenshots

Captured by driving the real app through the Electron E2E harness (`_electron.launch()` against the seeded test repo).

**Row at rest** — unchanged from today; the action occupies no width:

Worktree row at rest, no rename icon

**Row hovered** — pencil revealed at the right end. Note the title and metadata sit at exactly the same position as at rest, which is the layout-shift concern described above:

Worktree row hovered, rename icon revealed at the right end

**Clicking it opens the existing inline editor:**

Inline rename editor open with new text typed

**After committing the rename:**

Row showing the committed new name

## Testing

- [x] `pnpm lint` — passes for the touched files. The full script exits on a **pre-existing** failure in `src/renderer/src/components/skills/skill-freshness-group.tsx:101` (`switch-exhaustiveness-check`) that reproduces on a clean `main` and is untouched here.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm test` — all renderer suites pass, including the new ones. The full run reports 2 failures in `src/main/` (`terminal-attribution`, `ssh-system-transport.integration`); both pass in isolation on this branch, and a full run on clean `main` in the same environment fails **19** tests across 8 files (relay/pty/git subprocess suites), so these are load-sensitive environment flakes rather than regressions from this change.
- [x] `pnpm build` — `build:desktop` (renderer + web) succeeds. `build:native` fails locally in the Swift `computer-use-macos` manifest step, which needs a full Xcode toolchain and is unrelated to this TypeScript change.
- [x] Added tests that would catch regressions:
  - rename button renders with its aria-label and the workspace-board-preserving attribute
  - it stays collapsed (`max-w-0`, `opacity-0`) until hover/focus-within, and the group defers `pr-1.5` — this is the test that catches the layout shift described above
  - it is absent while deleting and in affiliate list mode
  - the flex-gap cancel applies only when another action precedes it
  - clicking invokes the rename handler (happy-dom)
- [x] Verified in the running app, not just in unit tests: launched Electron through the E2E harness and drove the full path — hover reveals the action, clicking it opens the inline editor, typing and pressing Enter commits the new name (screenshots above).

## AI Review Report

Reviewed with Claude. Risks checked and outcomes:

- **Cross-platform (macOS / Linux / Windows)**: the reveal is pure CSS hover/focus-within — no `metaKey`/`ctrlKey` branching, no accelerators, no path handling, no shell invocation. Nothing platform-conditional was added. The review specifically flagged that this change deliberately does *not* fill in the empty Windows/Linux `workspace.rename` bindings, since those defaults are intentional per `keybindings.ts`.
- **Layout regression**: flagged during review that collapsing opacity alone leaves the flex gap and the group's trailing padding intact, which would move the details/ports lane on rows that previously rendered no header actions. Fixed by collapsing width plus gap plus gutter, and pinned with a dedicated test.
- **SSH / remote / local**: rename is renderer-local state plus the existing `updateWorktreeMeta` call. No new host round-trips, so local, SSH, and remote-server workspaces behave identically.
- **Agent / integration / git provider**: no provider- or agent-specific code paths touched; nothing GitHub- or GitLab-specific involved.
- **Performance**: one extra button per rendered sidebar row, reveal handled by CSS rather than React state, so no new re-render or hover-tracking work in a hot list.
- **UI quality**: follows `docs/STYLEGUIDE.md` — reuses existing tokens (`text-muted-foreground`, `bg-accent/70`) and the sibling delete button's sizing, tooltip, and event-hygiene conventions. No new color values or shadow tiers.
- **Accessibility**: `group-focus-within` and `focus-visible` mirror the delete button, so the action is keyboard-reachable rather than hover-only, and it carries a translated `aria-label`. Touch has no hover, but all three existing rename paths are untouched, so nothing regresses there.

## Security Audit

Low surface. No input parsing, command execution, path handling, auth, secrets, or new IPC. The button dispatches an existing renderer store action with a worktree id already in scope; persistence flows through the existing `updateWorktreeMeta` path with its existing validation. No dependencies added. Click handling stops propagation so it cannot be used to trigger row activation or the Kanban dismissal path unexpectedly.

## Notes

- Deliberately out of scope, since each needs a maintainer product call and would stall an otherwise mechanical change:
  - splitting a plain "Rename" item out of the context menu's "Update" modal
  - filling in Windows/Linux defaults for `workspace.rename`
- `WorktreeCard.tsx` is already grandfathered in `config/max-lines-baseline.txt`, so the button lives in its own `WorktreeRenameQuickAction.tsx` rather than growing that file further. No new `max-lines` suppression is introduced.

## ELI5

Worktree cards in the sidebar now show a clear Rename action when you hover them. You could already rename via double-click, context menu, or a shortcut, but none of those were obvious; this makes rename discoverable.

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeCard.quick-actions.test.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeCard.quick-actions.test.tsx
@@ -24,12 +24,13 @@ let settings: Partial<GlobalSettings> | null = null
 let projectGroups: unknown[] = []
 let workspaceDeleteModifierPressed = false
 let gitConflictOperationByWorktree: Record<string, GitConflictOperation> = {}
+let deleteStateByWorktreeId: Record<string, { isDeleting: boolean }> = {}
 let WorktreeCard: typeof WorktreeCardComponent
 
 vi.mock('@/store', () => ({
   useAppStore: (selector: (state: unknown) => unknown) =>
     selector({
-      deleteStateByWorktreeId: {},
+      deleteStateByWorktreeId,
       fetchHostedReviewForBranch,
       fetchIssue,
       gitConflictOperationByWorktree,
@@ -148,6 +149,7 @@ describe('WorktreeCard quick actions', () => {
     projectGroups = []
     workspaceDeleteModifierPressed = false
     gitConflictOperationByWorktree = {}
+    deleteStateByWorktreeId = {}
   })
 
   it('marks the unread toggle as a workspace-board-preserving action', () => {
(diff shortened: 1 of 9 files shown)


nwparker:

### What this image is
Worktree card in left sidebar (master/primary) from Electron e2e — baseline for rename-on-hover (hover action not shown).

**Visual evidence**

| 10-workspace-active.png |
|---|
|  |

nwparker:
Thanks @Jaeyoung22 

I like it, but also worried about adding more visual icons haha
I might hold off for right now 🙏 

Open to revist of course, you might be right here

nwparker:
cc @AmethystLiang

PR #9548, opened 2026-07-20
fix(renderer): preserve workspace shortcut order when sidebar closes

### LINKED ISSUE #9497: [Bug]: Cmd+number selects a different workspace when the sidebar is closed
### Operating system

macOS

### Orca version

1.4.146

### Details

Short summary:

What happened?

How can we reproduce it?

Keep the sidebar open.
Close the left sidebar.
Press Cmd+1.

Expected behavior:

Actual behavior:

OnlyYu1996:
## Summary

- Keep the last rendered sidebar workspace order authoritative after the sidebar unmounts.
- While the sidebar is hidden, prune workspaces that are no longer eligible without reintroducing collapsed or otherwise unrendered rows.
- Preserve folder workspaces and the latest mainline lineage projection, cycle detection, forced visibility, and ancestor injection behavior.

Fixes #9497

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
  - Touched-file Oxlint passed.
- [ ] `pnpm typecheck`
  - `pnpm typecheck:web` passed.
- [ ] `pnpm test`
  - Rebase regression tests passed: 2 files, 56 tests.
- [ ] `pnpm build`
  - Not rerun after the latest rebase; this change does not alter build configuration or bundled assets.

Additional repository gates passed: touched-file Oxfmt, `git diff --check`, max-lines ratchet, and reliability gates.

## AI Review Report

The review checked the rendered-order lifecycle when the sidebar mounts and unmounts, stale-ID pruning, duplicate removal, collapsed groups, empty rendered orders, pinned children, and folder-workspace retention. The rebase conflict was resolved by retaining both sides of the contract: the PR cache lifecycle and prune-only fallback, plus mainline lineage projection, cycle handling, `forcedVisibleWorktreeIds`, and `injectLineageAncestors`.

Cross-platform compatibility was reviewed for macOS, Linux, and Windows. The change only manages renderer-side workspace IDs and does not alter shortcuts, labels, filesystem paths, shell behavior, Electron platform branches, SSH execution, or Git-provider behavior.

## Security Audit

This change only reads and filters in-memory workspace IDs. It adds no command execution, path handling, authentication, secrets, dependencies, network access, or IPC behavior. IDs remain opaque and are deduplicated with sets. No follow-up security work is needed.

## Notes

- Rebased onto current `upstream/main` at `41751dd90`.
- No platform-specific, remote-only, agent-specific, integration-specific, or Git-provider-specific behavior is introduced.
- X/Twitter: N/A (no account)

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeList.lineage-child-real-card.test.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeList.lineage-child-real-card.test.tsx
@@ -17,6 +17,8 @@ import {
   LINEAGE_NESTED_ROW_SURFACE_INSET,
   WORKTREE_CARD_SURFACE_MARGIN
 } from './worktree-list-indentation'
+import { ALL_GROUP_KEY } from './worktree-list-groups'
+import { getVisibleWorktreeIds } from './visible-worktrees'
 
 globalThis.IS_REACT_ACT_ENVIRONMENT = true
 
@@ -231,7 +233,12 @@ function makeFolderWorkspacePathStatusState(): Record<string, unknown> {
 }
 
 function setLineageState(
-  options: { deletingChild?: boolean; includeGrandchild?: boolean } = {}
+  options: {
+    collapseAll?: boolean
+    deletingChild?: boolean
+    includeGrandchild?: boolean
+    pinChild?: boolean
+  } = {}
 ): void {
   const repo = makeRepo()
   const parent = makeWorktree({
@@ -248,6 +255,7 @@ function setLineageState(
     branch: 'child-branch',
     sortOrder: 10,
     overrides: {
+      isPinned: options.pinChild ?? false,
       linkedGitLabMR: 42,
       comment: 'Child handoff note'
     }
@@ -275,7 +283,7 @@ function setLineageState(
     agentStatusEpoch: 0,
     browserTabsByWorktree: {},
     clearPendingRevealWorktreeId: vi.fn(),
-    collapsedGroups: new Set<string>(),
+    collapsedGroups: new Set<string>(options.collapseAll ? [ALL_GROUP_KEY] : []),
     deleteStateByWorktreeId: options.deletingChild
       ? { [child.id]: { isDeleting: true, error: null, canForceDelete: false } }
       : {},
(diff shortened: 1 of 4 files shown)


OnlyYu1996:
Rebased onto current `upstream/main` (`300ee1995`) in `a68eacb67` and resolved the `WorktreeList.tsx` conflict by retaining the last rendered shortcut order after sidebar unmount.

Validation on the rebased head: 50 targeted tests, `typecheck:web`, touched-file Oxlint/Oxfmt, max-lines ratchet, and reliability gates pass.

OnlyYu1996:
Rebased onto current `upstream/main` (`41751dd90`) in `61faf984a`.

The `visible-worktrees.ts` conflict keeps both contracts: mainline lineage projection/cycle handling/`forcedVisibleWorktreeIds`/`injectLineageAncestors`, and the PR rendered-order lifecycle/prune-only fallback/folder-workspace retention.

Validation: 2 related test files / 56 tests, `typecheck:web`, touched-file Oxlint/Oxfmt, max-lines ratchet, reliability gates, and `git diff --check` all pass.

nwparker:
Hi @OnlyYu1996 — thanks for digging into this one. Your diagnosis of the open-vs-closed split is correct, and it's what pointed me at the right area.

I've opened #10693 with a different approach, so flagging it here rather than leaving two PRs silently competing. The difference: pruning the retained cache on unmount can only ever *remove* stale IDs, so a workspace created while the sidebar is closed still gets no number, and the ordering rules stay duplicated in two places. #10693 recomputes the order from the sidebar's own row pipeline (`buildRows → addHostSectionRows → getRenderedWorktreesInSidebarOrder`), which covers both and keeps a single owner for the ordering rules.

Not a knock on yours — happy for maintainers to take whichever they prefer. Credit to you for finding it first.

nwparker:
Hi @OnlyYu1996 — closing this in favour of #10693, and I want to be straight with you about why, since you got here first and your diagnosis was right.

You correctly identified that the shortcut order goes wrong specifically when the sidebar is closed, and that `WorktreeList`'s unmount is what breaks it. That's the real bug, and you found it independently.

Where the approaches differ: this PR retains the last rendered order and prunes stale ids. Pruning can only ever *remove* entries, so a workspace **created while the sidebar is closed** never enters the numbering at all — and the tests here assert order is preserved across unmount with an unchanged store, so they can't catch that case. It also leaves the ordering rules living in two places, which is how the sidebar and the shortcut drifted apart originally.

#10693 recomputes the order by replaying the sidebar's own row pipeline (`buildRows → addHostSectionRows → getRenderedWorktreesInSidebarOrder`), so there's a single owner for the ordering rules and the created-while-closed case works. It also fixes a second, related bug: the published-order cache was guarded by `length > 0`, which conflated "sidebar unmounted" with "sidebar mounted showing zero cards".

This isn't a knock on your work — the pruning approach is a reasonable read of the problem, and your rebases through two conflicting refactors were diligent. Thank you for the report and the fix attempt; please do keep sending PRs.

PR #9560, opened 2026-07-20
fix(sidebar): activate workspace number shortcuts by rendered sidebar order

### LINKED ISSUE #9497: [Bug]: Cmd+number selects a different workspace when the sidebar is closed

Hanxcellent:
## What

Fixes #9497.

Cmd/Ctrl+1..9 workspace shortcuts now activate the Nth row exactly as rendered in the sidebar, and folder workspaces activate through the dedicated folder path instead of the plain worktree path.

## Why

The shortcut handler previously recomputed workspace order from the Zustand store (sorting/grouping/filtering). That order could disagree with the rows actually on screen — component-local state such as host drag collapse, collapsed groups, and import cards is invisible to the store — so pressing a number could jump to the wrong workspace. Folder workspaces were also mis-activated as regular worktrees.

## How

- `WorktreeList` publishes its rendered row order from a layout effect (`setVisibleWorktreeIds`), and registers itself as the order publisher while mounted (`registerVisibleWorktreeIdsPublisher`).
- While the sidebar is mounted, the shortcut path trusts the published order directly (O(1) lookup). Only after unmount does it fall back to reconciling against the store, so shortcuts still work with the sidebar closed while dropping rows that no longer exist.
- `workspace-number-shortcut-activation.ts` routes folder workspaces through `activateAndRevealFolderWorkspace`.

No visual change.

## Tests

- New regression tests: `visible-worktree-shortcut-order.test.ts` (published-order trust while mounted, reconciliation after unmount, empty-order handling) and `workspace-number-shortcut-activation.test.ts` (folder vs worktree activation routing). All were written red-first.
- Focused suites: 133/133 passed. Full `pnpm typecheck` passed. `pnpm build` passed. Full `pnpm test`: identical pass/fail set to `main` on the same machine (remaining failures are pre-existing local-environment issues on `main`, e.g. missing optional `windows-native-registry`).

## AI code review summary

Reviewed by a Codex reviewer agent over the full diff; final verdict: no actionable findings.

- Cross-platform: shortcut handling keeps the existing metaKey/ctrlKey runtime platform check; no hardcoded modifiers or paths.
- SSH/remote/local: activation goes through the existing `worktree-activation` paths, which are host-agnostic; no assumptions of local-only execution.
- Agents/integrations & git providers: no provider-specific logic touched; ordering is purely renderer-side.
- Performance: mounted path is O(1) per keypress; full reconstruction only runs when the sidebar is unmounted.
- UI: no visual change; behavior verified via unit/integration tests.
- Security: no new inputs, IPC surfaces, or shell/git command construction.

## Platform / remote notes

Verified on macOS. The change is renderer-only ordering logic behind the existing platform-checked shortcut handler, so no platform- or SSH-specific behavior differences are expected.

X handle: (contributor: add your handle here)

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeList.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeList.tsx
@@ -77,7 +77,6 @@ import {
 } from './smart-attention'
 import { track } from '@/lib/telemetry'
 import { tabHasLivePty } from '@/lib/tab-has-live-pty'
-import { deriveRunningAgentSendTargets } from '@/lib/running-agent-targets'
 import { rightSidebarShowsPullRequestData } from '@/lib/right-sidebar-visibility'
 import {
   type Row,
@@ -117,6 +116,7 @@ import { useWorkspaceStatusDocumentDrop } from './use-workspace-status-drop'
 import {
   computeClearFilterActions,
   computeVisibleWorktreeIds,
+  registerVisibleWorktreeIdsPublisher,
   setVisibleWorktreeIds,
   sidebarHasActiveFilters
 } from './visible-worktrees'
@@ -127,6 +127,10 @@ import {
   getVisibleWorktreeTerminalActivityTabs
 } from './visible-worktree-activity-inputs'
 import { selectWorktreeListReviewCacheInputs } from './worktree-list-review-cache-inputs'
+import {
+  getAgentSendEffectiveCollapsedGroups,
+  getEligibleAgentSendTargetWorktreeId
+} from './rendered-workspace-shortcut-order'
 import {
   VIRTUALIZED_SCROLL_ANCHOR_RECORD_EVENT,
   useVirtualizedScrollAnchor,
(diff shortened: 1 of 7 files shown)


nwparker:
For the record, since this stayed closed and unreviewed: @Hanxcellent, your approach here was substantially correct, and one piece of it is genuinely better than what shipped.

#10693 has now fixed the ordering half of #9497 the same way you did — deriving the shortcut order from the sidebar's rendered rows rather than a flat list.

The part you had that #10693 does **not**: `workspace-number-shortcut-activation.ts`, which branches on `parseWorkspaceKey` and routes folder workspaces to `activateAndRevealFolderWorkspace` instead of `activateAndRevealWorktree`. I checked this before deciding, and it isn't a crash — `activateAndRevealWorktree` does resolve a `folder:*` key (`findKnownWorktreeById` has a folder branch), activates the workspace and creates its terminal. So numbering a folder workspace works today.

But your split is still the more correct call: the plain path skips the `getFreshFolderWorkspacePathStatus` / `folderWorkspaceActivationBlocked` gate that `activateAndRevealFolderWorkspace` enforces, so Cmd+N on a folder whose path is missing or unmounted activates it silently instead of showing the error toast a sidebar click would. That's a real (if narrow) inconsistency between clicking a card and pressing its number.

I've left that out of #10693 deliberately to keep it to one reviewable concern, and filed it separately so it doesn't get lost. If you'd like to open a focused PR for just that activation split, it'd be welcome — credited to you.

Sorry this sat unreviewed after you closed it; the idea was a good one.

Issue #9231, opened 2026-07-17
[Bug]: Windows tar.exe times out while extracting local speech models


boris-papevis:
## Summary

Orca downloads the Parakeet TDT v3 archive on Windows, then spends ten minutes extracting it before the operation fails.

Orca launches `C:\Windows\System32\tar.exe`. This machine ships `bsdtar 3.8.4`, which processes the model archives too slowly to finish before Orca's extraction timeout.

## Environment

- Windows 10 `10.0.19045.7548`
- Orca `1.4.143`
- `C:\Windows\System32\tar.exe`: `bsdtar 3.8.4`

## Steps to reproduce

1. Open `Settings > Voice`.
2. Select `Parakeet TDT v3`.
3. Wait for the download and extraction.
4. Check the model state after ten minutes.

## Result

The model does not reach the ready state.

The downloaded v3 archive had:

- Size: `487,170,055` bytes
- SHA-256: `5793d0fd397c5778d2cf2126994d58e9d56b1be7c04d13c7a15bb1b4eafb16bf`

Both values match Orca's model catalog.

The same failure affected other local models. During one Parakeet v2 attempt, `tar.exe` wrote only `17,729,024` of the encoder's `652,184,296` bytes before Orca stopped it. The partial file timestamps span ten minutes, matching the timeout in `ModelManager`.

A retry downloads the archive again.

## Expected result

Orca should extract a verified archive and mark the model as ready. If extraction fails, Orca should keep the valid archive and retry extraction without another download.

## Relevant code

Orca selects the Windows system tar here:

https://github.com/stablyai/orca/blob/2cb5d4e149a1ef91c4cce2f9123e82d47229fa8e/src/main/speech/tar-executable.ts#L4-L26

`ModelManager` starts the extractor and stops it after ten minutes here:

https://github.com/stablyai/orca/blob/2cb5d4e149a1ef91c4cce2f9123e82d47229fa8e/src/main/speech/model-manager.ts#L855-L940

On this machine, listing the contents of the Parakeet archive with the system `tar.exe` takes more than five minutes. A longer timeout would leave users waiting for hours.

## Verification

Python's `tarfile` extracted the same archive. I checked the required model files and loaded them with Orca's bundled `sherpa-onnx-win-x64` runtime.

The model transcribed its included test WAV as:

> Ask not what your country can do for you. Ask what you can do for your country.

This confirms that the archive and model files work on this machine.

## Related issue

#8620 and #8775 address interrupted network downloads.

[...]

PR #9571, opened 2026-07-20
fix(speech): avoid system tar for Windows model extraction

boris-papevis:
## Summary

On Windows, extract `.tar.bz2` speech models in a worker thread instead of calling `C:\Windows\System32\tar.exe`.

The Windows bsdtar on the affected system did not finish extracting the Parakeet archive before Orca's 10-minute timeout. The streaming JavaScript extractor completed the same archive in about 113 seconds without blocking Electron's main thread.

macOS and Linux continue to use the existing system `tar` path.

Related issue: https://github.com/stablyai/orca/issues/9231

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
  - The changed files pass `oxlint`.
  - The repository-wide command stops on existing switch-exhaustiveness errors in `src/main/daemon/daemon-server.ts` and `src/renderer/src/components/skills/skill-freshness-group.tsx`.
- [x] `pnpm typecheck`
- [ ] `pnpm test`
  - The focused speech-model suite passes: 12 tests.
  - The full Windows suite has unrelated platform and baseline failures.
- [x] `pnpm build`
- [x] Added tests for successful `.tar.bz2` extraction, path traversal rejection, and link rejection

A built-worker smoke test also extracted a real `.tar.bz2` fixture from `out/main/speech-model-archive-worker.js`.

## AI Review Report

The review checked Windows, macOS, and Linux behavior; Electron worker packaging; local and SSH compatibility; agent and integration compatibility; performance; cancellation cleanup; and archive security.

The implementation changes only the Windows extraction path. macOS and Linux retain their current `tar` command. Speech models live in Orca's local application cache, so remote repositories and SSH worktrees do not use this path. The change does not affect CLI agents, Git providers, integrations, shortcuts, labels, or UI behavior.

The review identified two main risks:

- JavaScript bzip2 decompression could block Electron's main thread. The extractor runs in a worker thread.
- Archive entries could escape the model directory or create links. The extractor rejects absolute paths, `..` traversal, symbolic links, hard links, and unsupported entry types.

Cancellation terminates the worker before `ModelManager` removes partial files.

## Security Audit

Orca verifies the downloaded archive against its catalog SHA-256 before extraction.

The extractor also:

- rejects POSIX and Windows absolute paths
- rejects path traversal after stripping the archive wrapper directory
- rejects symbolic links, hard links, and special entries
- avoids shell command construction on Windows
- accepts archive and destination paths only through internal worker data

`tar` and `unbzip2-stream` are bundled into the worker, so the packaged application does not resolve them from an external runtime directory. This change does not handle credentials, authentication data, secrets, or external IPC.

## Notes

Windows uses the worker-based extractor. macOS and Linux keep the existing system `tar` implementation.

No remote, SSH, agent, integration, Git-provider, or UI behavior changes.

X: No account

### DIFF
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -173,7 +173,9 @@ export default defineConfig({
       // directory cannot reach into app.asar, so pure-JS dependencies used
       // by the daemon must be bundled rather than externalized.
       externalizeDeps: {
-        exclude: ['@xterm/headless', '@xterm/addon-serialize']
+        // Why: the extraction worker runs from app.asar and cannot resolve
+        // dependencies copied beside the archive. Bundle its pure-JS pipeline.
+        exclude: ['@xterm/headless', '@xterm/addon-serialize', 'tar', 'unbzip2-stream']
       },
       rollupOptions: {
         input: {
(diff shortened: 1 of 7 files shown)


nwparker:
thanks @boris-papevis for brining attention here ! To avoid the extra dep I merged #10735

LMK if it has any issues

PR #9575, opened 2026-07-20
docs(headless-server): add upgrade SOP for orca serve on Linux

PannenetsF:
## What

Adds an **Upgrade** section (with a step-by-step SOP) to `docs/reference/headless-linux-server.md`, plus one Troubleshooting bullet. The guide previously covered install / foreground run / systemd but had no upgrade path, leaving operators to guess how to move to a new AppImage without losing state.

## Why

`orca serve` never auto-updates in headless mode (the built-in updater only runs in the desktop GUI, and no paired client can trigger it remotely), so upgrading is always a manual step that deserves a documented, safe procedure.

## Contents

- **State is independent of the binary.** Persisted data lives under the service user's `~/.config/` (Orca uses both an `orca` and an `Orca` directory), independent of `/opt/orca`. `orca-data.json` is forward-migrated on load, so a forward upgrade needs no manual data step.
- **Atomic binary replace.** Download to `.new` on the same filesystem, verify, then `mv` — never `curl -o` over the FUSE-mounted live binary (it can crash/corrupt the running process; a failed download can also clobber a working binary).
- **Rollback is not binary-only safe** — called out explicitly. An older build silently strips newer `orca-data.json` fields it doesn't recognize, and the `.bak.*` ring is corruption-recovery, not a pre-upgrade copy. So the SOP backs up the whole `.config` before upgrading and requires restoring it (not just swapping the binary) to roll back cleanly.
- **No headless version command** — documents tracking the release tag instead of a nonexistent `--version`.

## Verification

Every technical claim in the SOP was verified against the codebase before writing (updater wiring in serve mode, `Orca server ready:` startup line, userData paths, migration/downgrade behavior, AppImage FUSE mount semantics), then the written SOP was adversarially reviewed for command correctness and data-loss traps.

Docs-only change; no code paths affected.

### DIFF
--- a/docs/reference/headless-linux-server.md
+++ b/docs/reference/headless-linux-server.md
@@ -165,6 +165,119 @@ If you later install the desktop CLI from Orca settings, use that CLI for normal
 shell workflows. Keep the AppImage path in systemd so service restarts do not
 depend on an interactive shell profile.
 
+## Upgrade
+
+`orca serve` never updates itself. In headless mode Orca wires up no auto-updater
+at all — the built-in updater only runs in the desktop GUI, and no paired mobile
+or web client can trigger it remotely. Upgrading is always a deliberate step:
+replace the AppImage and restart the service.
+
+Two facts make this safe and predictable:
+
+- **State lives in the service user's home, not next to the binary.** Persisted
+  data is under `/home/orca/.config/` (Orca uses both an `orca` and an `Orca`
+  directory there), fully independent of `/opt/orca/orca-linux.AppImage`.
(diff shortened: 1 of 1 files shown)


OrcaWin (approved):
Thanks for adding the missing headless upgrade path. I tightened the SOP before landing:

- stage and validate the architecture-matched AppImage while the server remains online
- keep /opt/orca root-owned and rollback generations private and self-contained
- restore only Orca profile directories with guarded failure recovery and explicit retention
- direct crash-looping upgrades to rollback instead of overwriting the known-good generation

The documented Bash paths and failure boundaries have been validated.

PR #9601, opened 2026-07-20
fix(worktree): don't let a WSL selector_not_found abort worktree deletion (#9197)

### LINKED ISSUE #9197: [Bug]: Cannot delete worktree in WSL environment (selector_not_found)
### Operating system

Windows

### Orca version

1.4.144

### Details

When deleting a worktree, the following error appears:

```bash
Unable to delete workspace feat/canvas
Error invoking remote method 'worktrees:remove': Error: selector_not_found
```

Environment:

- Project runtime: `WSL`
- Project path: `\\wsl.localhost\Ubuntu\home\xxx\xxx`
- Worktree path: `\\wsl.localhost\Ubuntu\root\orca\workspaces`

---

### Original (中文)

删除某个工作树时，出现下面错误

```bash
无法删除工作区 feat/canvas
Error invoking remote method 'worktrees:remove': Error: selector_not_found
```

环境信息：

- 项目运行时: `WSL`
- 项目位置: `\\wsl.localhost\Ubuntu\home\xxx\xxx`
- 工作树位置: `\\wsl.localhost\Ubuntu\root\orca\workspaces`

OrcaWin:
## Problem — Fixes #9197

Deleting a worktree in a **WSL-runtime** project fails with
`Error invoking remote method 'worktrees:remove': Error: selector_not_found`
and leaves the worktree **half-deleted**. Triggers when the project lives at
`\wsl.localhost\Ubuntu\...` with worktrees under `\wsl.localhost\Ubuntu\root\orca\workspaces`.

## Root cause (code-verified)

Destructive removal runs `stopPtysForDestructiveWorktreeRemoval` →
`killAllProcessesForWorktree(worktreeId)` → the runtime sweep
`runtime.stopTerminalsForWorktree(worktreeId)` →
`resolveWorktreeSelector(worktreeId)`, which resolves the selector **before**
touching any PTY.

- `killAllProcessesForWorktree` passes the **raw composite id** (`repoId::path`, no
  `id:`/`path:` prefix), so `resolveWorktreeSelector` falls to the exact-string
  else-branch (`worktree.id === selector`, `orca-runtime.ts`).
- Every *other* comparator in the removal path is normalization-tolerant (it folds
  the `\wsl.localhost` ↔ `\wsl$` UNC aliases + distro casing), so the removal
  **preflight matches and the delete begins** — then the exact id-match **misses**
  and `resolveWorktreeSelector` throws `selector_not_found`.
- The teardown **fail-closes on every error except `runtime_unavailable`**
  (`worktree-teardown.ts`), so that throw propagates and **aborts the whole delete**.

A slow WSL `git worktree list` exceeding `RESOLVED_WORKTREE_REPO_TIMEOUT_MS`
(degrades to an empty list) reproduces the same asymmetry. Distinct from `#9045`,
which fails later at the PTY-kill gate.

## Fix

The runtime sweep only stops **renderer-graph** terminals. If the runtime can't
resolve the worktree at all, there are simply **no such terminals to stop** — that
must never fail-close a destructive delete. So `selector_not_found` is treated as
benign there, exactly like `runtime_unavailable` (matching existing precedent that
already swallows `selector_not_found` elsewhere in `orca-runtime.ts`).

Safety is preserved: the **independent provider sweep** (which also matches PTYs by
cwd-inside-worktree, tolerant of the UNC aliasing) and **registry sweep**, plus the
`requirePhysicalStop` gate, remain the proof that any PTY that truly exists was
physically stopped. The invariant ("nothing to stop must never abort a delete")
holds on native/SSH too, so **those paths are untouched** — a generic runtime-sweep
error still fails closed.

```diff
-        (error) => !(error instanceof Error && error.message === 'runtime_unavailable')
+        (error) => !isBenignRuntimeSweepError(error)   // runtime_unavailable OR selector_not_found
```

## Reproduction & verification

A live slow-WSL Electron host is impractical to drive deterministically, so — as the
issue's own repro note allows — this uses a **focused unit reproduction** that forces
the exact failure mode: the runtime sweep rejects with `selector_not_found` under
`requirePhysicalStop`, using a `\wsl.localhost` composite id.

- **Red (before):** the new test throws `Error: selector_not_found` — the destructive
  teardown aborts (reproduces #9197).
- **Green (after):** the delete completes, `runtimeStopped: 0`, and the provider sweep
  still tears down the live PTY (`providerStopped: 1`).
- Full `worktree-teardown.test.ts` suite: **22/22 pass**. `oxlint` clean;
  `tsc -p config/tsconfig.node.json` clean.

## Trade-offs / user-visible changes

None beyond the fix: WSL worktree deletion now completes instead of aborting with
`selector_not_found`. No behavior change for native or SSH removal. A live WSL
Electron end-to-end pass is still worth a maintainer's confirmation before merge.

### DIFF
--- a/src/main/runtime/worktree-teardown.test.ts
+++ b/src/main/runtime/worktree-teardown.test.ts
@@ -581,6 +581,36 @@ describe('killAllProcessesForWorktree', () => {
     ).rejects.toThrow('runtime sweep failed')
   })
 
+  it('does not abort a destructive delete when the runtime cannot resolve the worktree (WSL selector_not_found, #9197)', async () => {
+    // On a WSL project the removal target's UNC form (\\wsl.localhost) can differ
+    // from the runtime's resolved id (\\wsl$), so resolveWorktreeSelector's
+    // exact-string branch throws selector_not_found before any PTY is touched.
+    // That means "no renderer-graph terminals to stop" — it must not fail-close
+    // the delete, while the provider sweep still tears down real PTYs.
+    const worktreeId = 'repo-1::\\\\wsl.localhost\\Ubuntu\\root\\orca\\workspaces\\wt'
+    const stopTerminalsForWorktree = vi.fn().mockRejectedValue(new Error('selector_not_found'))
+    const runtime = {
+      stopTerminalsForWorktree
+    } as unknown as Parameters<typeof killAllProcessesForWorktree>[1]['runtime']
+    const localProvider = createProviderStub(async () => [
+      { id: `${worktreeId}@@live-1`, cwd: '/wsl/wt', title: 'shell' }
+    ])
+    listRegisteredPtysMock.mockReturnValue([])
+
+    const result = await killAllProcessesForWorktree(worktreeId, {
+      runtime,
+      localProvider,
+      requirePhysicalStop: true
+    })
+
+    expect(result.runtimeStopped).toBe(0)
+    expect(result.providerStopped).toBe(1)
+    expect(localProvider.shutdown).toHaveBeenCalledWith(
+      `${worktreeId}@@live-1`,
+      expect.objectContaining({ immediate: true })
+    )
+  })
+
   it('bounds the entire process sweep when a provider never settles', async () => {
     vi.useFakeTimers()
     try {
--- a/src/main/runtime/worktree-teardown.test.ts
+++ b/src/main/runtime/worktree-teardown.test.ts
@@ -611,6 +611,29 @@ describe('killAllProcessesForWorktree', () => {
     )
   })
 
+  it('still fails the destructive delete closed when a provider PTY refuses to stop, even though the runtime cannot resolve the worktree (#9197)', async () => {
+    // Guard the invariant the #9197 fix depends on: swallowing selector_not_found
+    // relaxes only the renderer-graph sweep. A real provider-visible PTY that
+    // cannot be physically stopped must still block filesystem removal.
+    const worktreeId = 'repo-1::\\\\wsl.localhost\\Ubuntu\\root\\orca\\workspaces\\wt'
+    const runtime = {
+      stopTerminalsForWorktree: vi.fn().mockRejectedValue(new Error('selector_not_found'))
+    } as unknown as Parameters<typeof killAllProcessesForWorktree>[1]['runtime']
+    const localProvider = createProviderStub(async () => [
+      { id: `${worktreeId}@@live-1`, cwd: '/wsl/wt', title: 'shell' }
+    ])
+    localProvider.shutdown = vi.fn().mockRejectedValue(new Error('daemon refused kill'))
+    listRegisteredPtysMock.mockReturnValue([])
+
+    await expect(
+      killAllProcessesForWorktree(worktreeId, {
+        runtime,
+        localProvider,
+        requirePhysicalStop: true
+      })
+    ).rejects.toThrow('Failed to physically stop every PTY')
+  })
+
   it('bounds the entire process sweep when a provider never settles', async () => {
     vi.useFakeTimers()
     try {
--- a/src/main/runtime/worktree-teardown.ts
+++ b/src/main/runtime/worktree-teardown.ts
@@ -9,6 +9,19 @@ import { mapWithConcurrency } from '../../shared/map-with-concurrency'
 // or pathological inventory cannot fan out unbounded provider/RPC shutdowns.
 const WORKTREE_TEARDOWN_CONCURRENCY = 32
 
+// Why: the runtime sweep only stops renderer-graph terminals. If the runtime
+// can't resolve the worktree at all — headless/reloading (runtime_unavailable)
+// or a WSL UNC-alias id mismatch that misses resolveWorktreeSelector's
+// exact-string branch (selector_not_found, #9197) — there are no such terminals
+// to stop, so it must never fail-close a destructive delete. The provider and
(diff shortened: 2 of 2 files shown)


OrcaWin:
Did a deep review of this fix (logic + regression tests + native/SSH impact). It holds up well — the core reasoning is sound:

- **Swallowing `selector_not_found` cannot orphan a live PTY.** `stopTerminalsForWorktree` calls `resolveWorktreeSelector` *before* enumerating any leaves, so when resolution throws there are literally zero renderer-graph terminals it could have stopped — nothing is skipped. The provider and registry sweeps operate on the **raw** `worktreeId` string (they never go through `resolveWorktreeSelector`), so they remain the physical-stop proof for any PTY that truly exists, and the `requirePhysicalStop` gate over `stopAttempts` still fails the delete closed if any provider/registry-visible PTY can't be stopped. When `requirePhysicalStop` is set, `provider.listProcesses()` is still un-caught (`failClosed=true`), so a wedged provider list also fails closed — untouched by this change.
- **Narrowly scoped.** Only `selector_not_found` and `runtime_unavailable` are treated benign; `selector_ambiguous` and every other error still fail-close. Good.
- **Native & SSH unaffected.** Those ids resolve normally, so the new branch never triggers; even in the rare "git no longer lists the worktree" case, the provider sweep still covers it. This also nicely mirrors the existing tolerance already in `resolveWorktreeRemovalTarget`, which treats `selector_not_found` as recoverable for removal — the teardown was the one place that hadn't caught up.

**One change I pushed** (`6bbd970be`): added a regression test that locks the invariant the fix depends on — when `selector_not_found` is swallowed *and* a real provider-visible PTY refuses to stop, the destructive delete must still fail closed (`Failed to physically stop every PTY`). The original test only exercised the happy path where the provider succeeds; this guards against a future change accidentally letting the swallow bypass the physical-stop guarantee.

Full worktree-teardown suite (23 tests) + typecheck are green. No production-code changes needed. Residual risk is low and pre-existing: if a PTY were ever spawned under a *different* UNC-alias id form than the deletion target, all three sweeps (including the old runtime sweep) would miss it equally — this fix neither introduces nor worsens that. LGTM. 🚢

nwparker:
Closing as **superseded** — `main` already carries this exact fix.

## What happened

The production change in this PR is the runtime-sweep predicate in `src/main/runtime/worktree-teardown.ts`, treating `selector_not_found` as benign alongside `runtime_unavailable`. That same predicate landed on `main` in 57865139d99960f32f2c148acc7323b071e51052 (#9625, merged 2026-07-21T01:18Z, ~17h before the last push here) as one of six e2e-flake fixes, under the description "Tolerate `selector_not_found` during destructive worktree removal."

The two changes are semantically identical — same two error messages, same `instanceof Error` guard, same call site. This PR factors it into a named `isBenignRuntimeSweepError` helper; #9625 inlines it. No behavioral difference.

## Verified, not assumed

I grafted this PR's own WSL reproduction tests (the `\\wsl.localhost` composite id, `requirePhysicalStop: true`) onto **unmodified `main`** and ran them:

```
$ git checkout origin/main -- src/main/runtime/worktree-teardown.ts
$ npx vitest run --config config/vitest.config.ts src/main/runtime/worktree-teardown.test.ts
 Test Files  1 passed (1)
      Tests  28 passed (28)
```

Against this PR's base commit (`ccd72f5909`), the same tests fail with the exact #9197 symptom:

```
 × WSL selector_not_found does not abort a destructive delete (#9197)
   Error: selector_not_found
 × WSL selector_not_found still fails closed when a provider PTY refuses to stop (#9197)
   AssertionError: expected [Function] to throw error including
   'Failed to physically stop every PTY' but got 'selector_not_found'
 Tests  6 failed | 22 passed (28)
```

So the repro is valid and `main` genuinely fixes it — this isn't a "looks similar" judgment call.

## Why not just merge it anyway

Rebasing would be a net regression risk rather than a no-op. `main` has moved 428 commits past this branch, and `worktree-teardown.ts` conflicts (`git merge-tree` confirms). This branch's copy of the file predates #10252, which added the `cwdFallbackPath` guard to `sweepProviderByPrefix` — that guard stops a folder-workspace instance deletion from sweeping sibling instances' PTYs. Resolving the conflict toward this branch would silently drop it.

## Note on the root-cause analysis

The write-up in this PR and in #9197 is careful and correct, and worth preserving for anyone who hits this again: `selector_not_found` here is **Orca's own** error from `resolveWorktreeSelector`, not anything `wsl.exe` emits. The WSL angle is that `killAllProcessesForWorktree` passes the raw `repoId::path` composite id, forcing the exact-string match branch, while every other comparator on the removal path folds the `\\wsl.localhost` ↔ `\\wsl$` UNC aliases — so the preflight matches, deletion begins, and the exact match then misses. Since the call is in-process (no RPC error wrapping between `stopTerminalsForWorktree` and the predicate), the exact-string match is a sound way to catch it.

The underlying id-normalization asymmetry is *not* fixed by either patch — both only stop it from aborting the delete. Worth a separate issue if it resurfaces elsewhere.

## Salvageable

One test here covers a combination `main` doesn't: `selector_not_found` swallowed **and** a provider-visible PTY refusing to stop must still fail closed with `Failed to physically stop every PTY`. `main` has a fail-closed test for the wedged-daemon-RPC case (#9500) but not this pairing. It locks the invariant the swallow depends on. Happy to land it standalone if a maintainer wants it.

Thanks @OrcaWin — the diagnosis was right and independently arrived at the same fix that shipped; it just lost a race by about half a day.

PR #9603, opened 2026-07-20
fix(skills): pass -y to skill install command so it can't hang on the agent-selection prompt

### LINKED ISSUE #9567: [Bug]: Settings "Install" reports success when the skill install command exits 1, and Re-check never re-runs it
### Environment
- **OS**: Windows + WSL2 (Ubuntu) as the Orca work area; agents (Claude Code, Codex, Pi) installed inside WSL2
- **Orca version**: 1.4.146 (desktop app on Windows)
- **Node**: v24.18.0 (in WSL2)
- **Skills CLI**: `npx skills@latest`

### Summary
Clicking **Install** on the Settings → Skills page (e.g. *Orchestration*, *Computer use*) shows "install complete" in the UI, but the skill is never actually installed. **Re-check** keeps reporting "Not installed" and every agent shows "missing" coverage. This happens specifically when the Orca work area is a WSL2 distro.

Root cause: the install command Orca spawns in the ephemeral setup terminal is missing the `-y` flag, so the `skills` CLI's interactive agent-selection prompt (`@clack/prompts` "Which agents do you want to install to?") blocks forever in the non-driven PTY. Orca kills it after ~50s and it exits with code 1. **Re-check only re-reads the cached failed status — it does not re-run the install** — so the indicator can never flip back to "installed" on its own.

### Steps to reproduce
1. Orca desktop on Windows, set work area to a WSL2 distro (Ubuntu).
2. Install at least one agent inside WSL2 (e.g. Claude Code, Codex, Pi).
3. Open Settings → Skills, click **Install** on *Orchestration* (or *Computer use*).
4. UI reports the install completed.
5. Click **Re-check** → still "Not installed"; agent coverage shows all agents "missing".

### Expected behavior
Install completes non-interactively; Re-check shows the skill installed and agents covered.

### Actual behavior
The ephemeral setup terminal exits with code 1 every time; status never becomes "installed".

### Evidence
The spawned command template (found in `resources/app.asar`):
```js
npx skills add ${ORCA_SKILLS_REPOSITORY_URL} --skill ${skillNames.join(' ')} --global
```
Note: **no `-y`**.

`daemon.log` shows the same lifecycle for every skill install (and onboarding) terminal:
```
session-created: ephemeral-setup-terminal:settings-orchestration-skill-terminal@@<id>
... ~50s ...
session-killed:  immediate: true
session-exited:  code: 1
```

The `skills` CLI itself prints this tip:
> Tip: use the --yes (-y) and --global (-g) flags to install without prompts.

Running the same install manually **with `-y`** succeeds (exit 0) and the skills land correctly in each agent's skill directory inside WSL2:
```bash
npx skills add https://github.com/stablyai/orca --skill orchestration computer-use orca-cli --global -y --agent '*'
```

### Suggested fix
Add `-y` (and optionally `-a '*'` or an explicit agent list) to the spawned install command:
```js
npx skills add ${ORCA_SKILLS_REPOSITORY_URL} --skill ${skillNames.join(' ')} --global -y
```
Additional considerations:
- Run the install terminal without an interactive PTY (or detect non-TTY), so the `skills` CLI falls into its non-interactive path automatically.
- Make **Re-check** actually re-run the install instead of only re-reading the cached failed status, so the indicator can self-heal.

### Workaround
Run the install manually inside the WSL2 shell with `-y`:
```bash
npx skills add https://github.com/stablyai/orca --skill orchestration computer-use orca-cli --global -y --agent '*'
```
The skills become usable by the agents even though the Settings UI still shows "Not installed".

OrcaWin:
Fixes #9567

## Problem

Clicking **Install** on Settings → Skills (e.g. *Orchestration*, *Computer use*) reports "install complete" in the UI, but the skill is never installed. **Re-check** keeps reporting "Not installed" and every agent shows "missing" coverage. Reproduces reliably when the Orca work area is a WSL2 distro with agents installed inside it.

## Root cause

Orca spawns the install command in a **non-driven ephemeral setup terminal** (a real PTY, so `stdin.isTTY` is true). The spawned command was:

```
npx skills add https://github.com/stablyai/orca --skill <names> --global
```

Note: **no `-y`**. The `skills` CLI runs an interactive `@clack/prompts` selection — *"Which agents do you want to install to?"* — whenever stdin is a TTY. Nothing drives that prompt, so it blocks forever; Orca kills the terminal after ~50s and it exits with code 1. The status never flips to "installed".

Confirmed against the `skills` CLI dist (`skills@1.5.19`): the add command only skips the prompt when `options.yes` (`-y`/`--yes`) is set, taking the non-interactive path `targetAgents = ensureUniversalAgents(installedAgents)` (all detected agents). This matches the CLI's own tip: *"use the --yes (-y) and --global (-g) flags to install without prompts."*

## Fix

Append `-y` to the install command built by `buildAgentFeatureSkillInstallCommand`, so every install surface (Settings panes, onboarding, feature tips, floating-terminal dialog) runs non-interactively and installs to all detected agents. This is the single source of truth for the spawned template, so onboarding/settings/CTA all pick it up.

With install succeeding, **Re-check** reads the real installed state and self-heals — no separate re-run needed.

### Scope

Only the **install** command changed. The **update** commands run in user-facing terminals where the `skills` CLI's upstream-deletion prompt is an intentional interactive choice, so they're deliberately left alone.

## Tests

Updated the exact-match assertions for the install command string across the affected unit tests. All install-command-related suites pass (73 passed, 1 pre-existing skip).

### DIFF
--- a/src/renderer/src/components/onboarding/onboarding-feature-setup.test.ts
+++ b/src/renderer/src/components/onboarding/onboarding-feature-setup.test.ts
@@ -119,7 +119,7 @@ describe('onboarding feature setup runner', () => {
 
     expect(text).toBe(ALL_SKILL_INSTALL_COMMAND)
     expect(text).toBe(
-      'npx skills add https://github.com/stablyai/orca --skill orca-cli computer-use orchestration orca-linear --global'
+      'npx skills add https://github.com/stablyai/orca --skill orca-cli computer-use orchestration orca-linear --global -y'
     )
   })
 
--- a/src/renderer/src/components/settings/AgentSkillSetupPanel.test.tsx
+++ b/src/renderer/src/components/settings/AgentSkillSetupPanel.test.tsx
@@ -7,7 +7,8 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { AgentSkillSetupPanel } from './AgentSkillSetupPanel'
 import { TooltipProvider } from '../ui/tooltip'
 
-const INSTALL_COMMAND = 'npx skills add https://github.com/stablyai/orca --skill orca-cli --global'
+const INSTALL_COMMAND =
+  'npx skills add https://github.com/stablyai/orca --skill orca-cli --global -y'
 const UPDATE_COMMAND = 'npx skills update orca-cli --global'
 
 const mocks = vi.hoisted(() => ({
--- a/src/renderer/src/components/settings/OrchestrationPane.test.tsx
+++ b/src/renderer/src/components/settings/OrchestrationPane.test.tsx
@@ -8,7 +8,7 @@ import { getOrchestrationUsageExamples } from '@/lib/orchestration-usage-example
 import { OrchestrationPane } from './OrchestrationPane'
 
 const INSTALL_COMMAND =
-  'npx skills add https://github.com/stablyai/orca --skill orchestration --global'
+  'npx skills add https://github.com/stablyai/orca --skill orchestration --global -y'
 const UPDATE_COMMAND = INSTALL_COMMAND
 
 const mocks = vi.hoisted(() => ({
--- a/src/renderer/src/components/settings/linear-agent-skill-install-cta.test.tsx
+++ b/src/renderer/src/components/settings/linear-agent-skill-install-cta.test.tsx
@@ -115,7 +115,7 @@ describe('LinearAgentSkillInstallCta', () => {
     expect(rendered.textContent).toContain('Not installed')
     expect(rendered.textContent).toContain('Let your agents read and edit Linear tasks.')
     expect(rendered.textContent).toContain(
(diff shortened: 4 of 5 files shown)


OrcaWin:
Closing this — after digging into the code, `-y` is the wrong fix.

The Settings → Skills install terminal is intentionally **interactive**: clicking Install pre-fills the `npx skills add …` command into an embedded terminal and the user presses Enter to run it, and the multi-agent "Which agents do you want to install to?" prompt is a deliberate choice point. Adding `-y` would remove that choice, and it wouldn't even reliably fix the reported failure — if the user never presses Enter, `-y` changes nothing.

We'd rather keep the flow interactive and treat #9567 as a discoverability/reliability bug instead. I've followed up on the issue to confirm exactly where it broke for the reporter before deciding on the real fix.

PR #9606, opened 2026-07-20
fix(terminal): harden macOS login-preflight spawn path (#9301 follow-ups)

brennanb2025:
Follow-ups to the two-pass review on #9301 (`fix(terminal): bypass rejected macOS login sessions`). I re-derived each finding's severity from the code rather than trusting the review labels. None are release blockers; these are low-likelihood robustness, observability, and bounded-work fixes.

## What's fixed

**F1 — transient probe failures recover without taxing every spawn** (`macos-tcc-login-shell.ts`)
The probe returns `{ok, conclusive, reason}`. Only a conclusive PAM verdict caches permanently. A probe killed by our SIGKILL timeout / `ETIMEDOUT` / maxBuffer is inconclusive: it fails open to a direct shell, then retries with 5-second exponential backoff capped at 5 minutes. This preserves self-healing without adding a 500 ms wait and a `login(1)` subprocess to every terminal spawn when the environmental failure persists.

**F2 — degrade is diagnosable** (`daemon-entry.ts`)
`prepareMacosTccLoginShell()` returns the outcome; the daemon emits a structured `macos-login-preflight {ok, reason}` log on degrade, since detached daemons destroy stderr and lose the `console.warn`.

**F3 — concurrent same-session-id spawn is coalesced before process launch** (`local-pty-provider.ts`)
After the only async preparation boundary, the loser re-checks the stable session ID and attaches to the winner before calling node-pty. This avoids both an orphan and the prior follow-up's redundant spawn-then-kill subprocess. The winning PTY receives the later request's resize and retains its native/WSL metadata.

**F4 — pending daemon preparation follows connection ownership** (`daemon-server.ts`)
Preparations carry their `clientId`; control- or stream-socket loss cancels only that client's pending work. Reconnect handoff also cancels the old owner's preparation synchronously, covering the same-ID socket-replacement path where the stale close handler intentionally does nothing. Other clients' preparations are unaffected.

**F7**
Warm the PAM probe at daemon startup (mirrors `warmPwshAvailabilityCache`); document the pipe-vs-PTY probe fidelity limit; preserve `SessionNotFoundError` when `kill` has no pending preparation.

## Deliberately not fixed

- **F5** (orphan sweep misses in-flight spawns): self-limits to the next reload, and a complete fix needs generation ownership for in-flight local work.
- **F6** (hibernated v22 wake bypasses v23 preflight): inherent to legacy attach; self-heals once daemons are v23.
- F7 error-name unification: cosmetic cross-layer churn.

## Reliability and performance contract

- **Invariant:** a fresh terminal reaches the configured shell; a rejected/hung PAM probe never strands the PTY at `login:`. One stable local session ID owns at most one native PTY, and a disconnected/replaced daemon client cannot create a later orphan.
- **Failure sources:** transient or persistent PAM timeout, concurrent same-ID local spawn, control- or stream-socket loss during preparation, and same-client-ID reconnect before the stale close event.
- **Oracle:** deterministic tests count `execFile`, node-pty spawn, resize, and daemon subprocess calls. Persistent failures attempt at t=0, 5s, and 15s rather than every spawn; two concurrent local requests produce one native spawn; a lost owner's control or stream socket and reconnect replacement produce zero subprocesses after ownership cancellation, while unrelated clients continue.
- **Gate:** focused provider/daemon unit coverage is the deterministic gate. The existing real macOS Electron run covers the healthy spawn/reattach path; live PAM-timeout and socket-replacement fault injection remain accepted gaps.
- **Provider/platform coverage:** macOS local and daemon paths are covered. The stable-ID coalescing path is platform-neutral and retains Windows/WSL metadata semantics. SSH/remote and mobile/relay do not execute this local/daemon preflight path and are unaffected.
- **Diagnostics:** detached-daemon degradation is recorded as structured metadata without PAM output or account details.
- **Performance budget:** no polling, timers, global session inventories, or provider fanout were added. Retry state is O(1), probes remain in-flight deduplicated, retry frequency is exponentially bounded, and concurrent same-ID requests now launch one native process instead of two.

## Tests

- 139/139 focused tests across macOS preflight, local PTY, daemon server, and client ownership suites.
- 11 adjacent terminal/daemon/provider suites passed, including session teardown, terminal host, daemon adapter respawn/reattach, subprocess, entry/main, and IPC.
- `pnpm run typecheck:node` passed.
- Changed-file oxlint, React Doctor lint, oxfmt, and `git diff --check` passed.
- Existing Electron dev-build macOS happy-path evidence is attached below.

### DIFF
--- a/src/main/daemon/daemon-entry.ts
+++ b/src/main/daemon/daemon-entry.ts
@@ -82,6 +82,18 @@ async function main(): Promise<void> {
   daemonLog.log('startup', { protocolVersion: PROTOCOL_VERSION, socketPath })
   void warmPwshAvailabilityCache()
 
+  // Why: detached daemons destroy stderr, so the preflight's console.warn is lost;
+  // surface a degraded TCC attribution here where it's diagnosable (F2).
+  const runMacosLoginPreflight = async (): Promise<void> => {
+    const outcome = await prepareMacosTccLoginShell()
+    if (outcome && !outcome.ok) {
+      daemonLog.log('macos-login-preflight', { ok: outcome.ok, reason: outcome.reason })
+    }
+  }
+  // Why: warm the PAM probe at idle startup so the first terminal spawn doesn't
+  // pay it under load — shrinking the window where a slow probe degrades (F1/F7).
+  void runMacosLoginPreflight()
+
   // Why: node-pty can throw a C++ Napi::Error that escapes all JS try/catch
   // blocks (e.g. writing to a PTY whose fd was closed between the native
   // exit signal and the JS onExit callback). Without this handler, Node's
@@ -152,7 +164,7 @@ async function main(): Promise<void> {
     ...(launchNonce ? { launchNonce } : {}),
     ...(pidPath ? { startedAtMs } : {}),
     log: daemonLog,
-    preparePtySpawn: prepareMacosTccLoginShell,
+    preparePtySpawn: runMacosLoginPreflight,
     spawnSubprocess: (opts) => createPtySubprocess(opts),
     onIdleShutdown: () => {
       shuttingDown = true
--- a/src/main/daemon/daemon-preflight-client-replacement.test.ts
+++ b/src/main/daemon/daemon-preflight-client-replacement.test.ts
@@ -1,27 +1,34 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { connect, type Socket } from 'node:net'
 import { join } from 'node:path'
 import { tmpdir } from 'node:os'
-import { mkdtempSync, rmSync } from 'node:fs'
+import { mkdtempSync, readFileSync, rmSync } from 'node:fs'
 import { DaemonClient } from './client'
 import { DaemonServer } from './daemon-server'
+import { encodeNdjson } from './ndjson'
+import { PROTOCOL_VERSION } from './types'
 import type { SubprocessHandle } from './session'
 
 type DaemonServerPrivate = {
   pendingPtySpawnPreparations: Map<string, Set<unknown>>
+  clients: Map<string, { streamSocket: Socket | null }>
 }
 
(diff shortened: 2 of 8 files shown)


brennanb2025:
## Manual verification (Electron dev build, macOS)

Launched this branch's dev build (`tigerfish @ brennanb2025/agent-slack-followups`) via the electron skill (CDP-attached), added this repo, activated the worktree, and drove real terminals. Goal: confirm the happy path my changes touch — macOS shells spawn wrapped in `/usr/bin/login` (the login-preflight path) — still works and that F1/F3 don't regress normal spawn/reattach.

**What was exercised through the real UI:**

**1. Fresh spawn #1 — live shell, no `Login incorrect` / `login:`.** The preflight hit the conclusive-accept path (F1), the wrapper produced a working interactive zsh. Typed a command; it executed (`ORCA-PTY-LIVE-MARKER` / `done`). The dropped leading `w` in `hoami` is my automation's first-keystroke-before-focus race, not a product issue — the shell clearly received and ran the rest.

Terminal 1 fresh spawn, live

**2. Fresh spawn #2 — second new terminal tab, also live** (`SECOND-TERMINAL-LIVE` + `pwd`). A second independent fresh spawn through the same preflight/wrapper path.

Terminal 2 fresh spawn, live

**3. Reattach — switched back to Terminal 1; history preserved, no respawn.** Exercises the existing-PTY reattach early-return (the path the F3 guard sits just past).

Terminal 1 reattach, history preserved

### Coverage honesty
The four behavioral fixes are **edge cases not reachable through normal UI** — a transient 500 ms probe timeout (F1), the daemon degrade log (F2), a concurrent same-session-id spawn race (F3), and a mid-preflight control-socket disconnect (F4). Those are covered by the added unit tests (137/137), not by this manual run. What the manual run proves is that the **common spawn + reattach path is intact** on macOS after these changes — i.e. no regression to the thing #9301 originally fixed.

PR #9608, opened 2026-07-20
fix(terminal): disarm stale TUI modes when a pane confirms return to shell

akapug:
## Problem

When a TUI agent is killed hard (SIGKILL via a process manager, OOM, crash) while its parent shell survives, it never restores the terminal modes it armed. The emulator keeps any-motion mouse tracking (1003/1006), focus reporting (1004), and Kitty keyboard flags on, with two user-visible consequences:

- every pointer movement over the pane lands as typed SGR motion reports at the shell prompt (`35;250;36M35;216;39M…`), filling the input line with garbage that is one Enter away from executing;
- while the doomed TUI is still alive, the same any-motion firehose keeps waking it to parse events — we traced a "fully idle" agent pinned at ~4.5% CPU to exactly this.

Observed in production with Claude Code (arms mouse reporting for scroll support) killed via btop; any mouse-mode TUI reproduces it.

## Why existing resets don't cover it

The codebase already sanitizes stale modes on the *dead-PTY* paths — hibernation kills (`POST_REPLAY_MODE_RESET`) and daemon reattach (`POST_REPLAY_REATTACH_RESET`) — and #9259 filters leaked *focus* reports after Codex history resume. But an agent dying **under a live shell** never crosses any of those paths: the PTY never exits, no replay happens, and the modes stay armed indefinitely.

## Fix

Fire `POST_REPLAY_REATTACH_RESET` (reused — its value is exactly the safe-for-a-live-shell bundle: cursor style/visibility, Kitty flags, all mouse protocols/encodings, focus reporting) at the pane-foreground-agent tracker's **confirmed return-to-shell** transition (`onConfirmedShellForeground`), next to the existing stale-title cleanup — the same residue class, one level down. The write goes through `replayIntoTerminal` so the replay guard keeps xterm's auto-replies from leaking to the shell as input.

Bracketed paste (`?2004l`) is deliberately **not** reset here: the surviving shell re-arms it at its prompt before the confirming read lands, so disabling it would strip paste protection the shell itself requested. (The dead-PTY paths keep the full bundle — there is no live shell to protect.)

The confirmed-shell transition only fires after a foreground process read proves the agent is gone (a leaked nested-shell `133;D` under a live full-screen agent republishes the agent and does not disarm — covered by a test).

## Tests

- `disarms stale TUI modes in the emulator after a confirmed return to shell` — fails on unpatched main, passes with the fix; also asserts no disarm on the bare `133;D` before confirmation, and that the reset excludes `?2004l`.
- `keeps armed modes while the agent still owns the foreground after a leaked 133;D` — negative control.
- Full `terminal-pane` suite green with the change (2,229 tests at v1.4.146, 493 in the touched files on main).

## Verification beyond CI

Dogfooded on a patched v1.4.146 build: the original repro (btop-SIGKILL a claude pane, then mouse over the window) produced a clean prompt on every pane; before the patch it reliably spewed motion reports.

---

This fix was authored, tested, and verified by Claude (Fable 5)

## ELI5

If a TUI agent is killed hard, mouse-tracking and special keyboard modes could stick on and dump garbage at the shell prompt. When the pane is back at a shell, Orca disarms those leftover modes.

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-connection.test.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.test.ts
@@ -5606,6 +5606,88 @@ describe('connectPanePty', () => {
     expect(resolveMockPaneWindowsShiftEnterEncoding(mockStoreState, paneKey)).toBe('alt-enter')
   })
 
+  it('disarms stale TUI modes in the emulator after a confirmed return to shell', async () => {
+    vi.useFakeTimers()
+    const { connectPanePty } = await import('./pty-connection')
+    vi.mocked(window.api.pty.confirmForegroundProcess).mockResolvedValue('bash')
+    const dataCallbackRef: { current: ((data: string) => void) | null } = { current: null }
+    const ptyId = 'pty-stale-mode-disarm'
+    const transport = createMockTransport(ptyId)
+    transport.connect.mockImplementation(async ({ callbacks }: { callbacks: ConnectCallbacks }) => {
+      dataCallbackRef.current = callbacks.onData ?? null
+      return { id: ptyId }
+    })
+    transportFactoryQueue.push(transport)
+    const paneKey = makePaneKey('tab-1', LEAF_1)
+    const pane = createPane(1)
+    const { writes } = captureCallbackTerminalWrites(pane)
+
+    connectPanePty(
+      pane as never,
+      createManager(1) as never,
+      createDeps({ isVisibleRef: { current: false } }) as never
+    )
+    await vi.advanceTimersByTimeAsync(20)
+    await flushAsyncTicks()
+    mockStoreState.agentLaunchConfigByPaneKey[paneKey] = {
+      launchConfig: { agentArgs: '', agentEnv: {} },
+      identity: { agentType: 'droid' }
+    }
+
+    // A SIGKILLed agent emits no mode teardown; only the shell's next prompt
+    // mark arrives. The bare 133;D must not disarm yet — the foreground read
+    // has not confirmed the agent is gone.
+    dataCallbackRef.current?.('\x1b]133;D;0\x07')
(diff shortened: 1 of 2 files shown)


brennanb2025 (approved):
Thanks — this is a good catch and a well-scoped fix.

Verified rather than taken on faith:

- **The test is load-bearing.** I applied only the test additions to unpatched `main`: `disarms stale TUI modes in the emulator after a confirmed return to shell` fails. With the source change it passes. The `133;D`-without-confirmation negative control is the right guard, since a full-screen agent's nested shells leak their own `133;D`.
- **Green standalone on current `main`** (`94009f63ff`, this branch is ~480 commits behind): `typecheck` clean, `terminal-pane` suite 190 files / 2479 tests passing.
- **Direction.** I looked at doing this daemon-side with `TerminalPrivateModeTracker` and resetting only what's actually armed — it doesn't work: that tracker follows mouse/SGR modes but not the Kitty-keyboard or focus-reporting bits you also need to clear, and the confirmed-shell signal lives in the renderer anyway. Reusing `POST_REPLAY_REATTACH_RESET` at `onConfirmedShellForeground` is the right call.
- **Writing this bundle to a live session is already established** — it's what the daemon-reattach path at `pty-connection.ts:5120` does today. So this extends an accepted trade-off to a new trigger rather than introducing a new class of risk. Worth noting the cost is now paid more often: per fire, a shell-set cursor shape reverts to the terminal default (self-heals at the next prompt) and the replay guard drops input for ~2.8ms.

One nit for the squash message: the commit message describes a new `CONFIRMED_SHELL_STALE_MODE_RESET` constant that excludes `?2004l`, but the final code reuses `POST_REPLAY_REATTACH_RESET` (which already excludes it). The PR body is accurate; only the commit message drifted. I'll reword on merge.

PR #9609, opened 2026-07-20
feat(terminal): command autosuggest (history-based ghost-text)

Piya-Boy:
## Summary
- Fish/zsh-autosuggestions-style inline ghost-text command suggestions in terminal panes, sourced from this-session commands plus persisted per-worktree shell history (bash/zsh in v1).
- Accept with Right Arrow or End at end-of-line (real, user-rebindable keybinding via `terminal.acceptAutosuggest`).
- New Settings toggle `terminalAutosuggestEnabled` (default on).
- Works across both of the app's existing OSC133 delivery paths (main-authority facts for local/SSH PTYs, renderer byte-mode for remote-runtime PTYs), so it isn't limited to one host type by construction.
- Also adds a full Thai (th) UI locale (catalog translation + locale registration), and fills in previously-untranslated es/ja/ko/zh strings flagged in review.

Known, intentionally out-of-scope limitations for this PR (documented, not silently broken):
- WSL-via-daemon and full SSH persisted-history (remote HISTFILE write + read) are left as explicit follow-ups — direct-spawn WSL and local/daemon-hosted local are covered; SSH panes correctly fall back to session-only suggestions rather than misbehaving.
- Shell-native completion fallback (e.g. bash/zsh/fish/pwsh tab-completion) is a separate, larger follow-up ("sub-project 2" in the design doc), not part of this PR.

Design and implementation docs (spec, plan, and the full per-task review history) were produced during development but live under `docs/superpowers/` and `.superpowers/`, both gitignored in this repo, so they aren't part of this diff.

## Screenshots

No screenshot captured. Visual changes are: (1) an inline ghost-text suggestion rendered after the cursor on the current input line in a terminal pane, and (2) a new "Command Autosuggest" toggle in Settings → Terminal Behavior. Both are exercised by `TerminalAutosuggestOverlay.test.tsx` and `TerminalBehaviorSection.test.tsx`. Manual screenshot/recording is still recommended before merge (see Test plan).

## Test plan
- [x] Full existing test suite passes (32,452 passed; the only failures — 4 files / 5 tests — are pre-existing on `main` with zero relation to this branch's files, independently verified by running the same test files against a clean `main` worktree).
- [x] New unit tests across all new modules (prefix-match engine, OSC133 scanner extension, line tracker, history-source parsing, cell-position math, overlay component, keybinding policy, settings section) — TDD throughout.
- [x] `pnpm run typecheck` and `pnpm run lint` clean.
- [ ] Manual verification in a real terminal pane across bash/zsh/fish/pwsh and native/WSL/SSH hosts (recommended before merge, per the design doc's testing section).

## AI Review Report

Built via `superpowers:subagent-driven-development` — every task went through an independent implementer + reviewer pass, plus a final whole-branch review. Findings caught and fixed along the way:
- Candidate-ordering contract bug in the autosuggest engine (`candidates[0]` must be most recent; an early test had the convention backwards) — caught by the reviewer, fixed with a corrected test.
- `acceptAutosuggest` was checked via a hardcoded `event.key` comparison instead of the registered keybinding, so user rebinding silently had no effect — routed through `keybindingMatchesAction` instead, with a rebinding-proof test.
- `acceptAutosuggest` wasn't gated on `event.repeat`, risking double-insertion of the suggestion if a held key's repeat outran the PTY echo (worse over SSH latency) — fixed.
- Persisted history wasn't actually seeded for the default (daemon-spawned) local terminal path — the daemon `pty:spawn` handler never injected the per-worktree `HISTFILE` env var. Fixed at the source (`src/main/ipc/pty.ts`) rather than papering over it in the renderer.
- Cross-platform: keybinding resolution goes through the existing platform-aware `keybindingMatchesAction`/`KeybindingDefinition` machinery (no hardcoded `metaKey`), so Mac vs. Linux/Windows shortcut resolution is inherited rather than reimplemented. Shell-kind resolution (bash/zsh/fish/pwsh/cmd) and host type (native/WSL/SSH) are handled explicitly, with SSH intentionally falling back to session-only suggestions rather than guessing at a remote `HISTFILE`.

Separately, CodeRabbit's automated review on this PR flagged 3 real issues, all fixed with TDD (RED proving the bug, GREEN after the fix):
- `cellToPixelPosition` was missing a viewport-relative row subtraction, causing the overlay to render off-screen after scrolling past a couple thousand lines.
- `parseShellHistoryContent` split on `\n` only, leaving trailing `\r` from CRLF-terminated history files, which could cause accidental early command submission when a suggestion was accepted.
- The `readTerminalHistoryFile` IPC handler didn't validate that `wslDistro` was a string before use.
A 4th CodeRabbit flag (a `child_process.exec`-style static-analysis hit) was investigated and confirmed to be a false positive — the flagged line was `RegExp.prototype.exec()`, unrelated to process execution.

## Security Audit

- **IPC boundary**: `readTerminalHistoryFile` (`src/main/ipc/terminal-history-file.ts`) validates all arguments (including `wslDistro`, added after CodeRabbit's review) before touching the filesystem, and distinguishes ENOENT (return `null`) from real errors (throw) instead of swallowing failures silently. Covered by dedicated tests for both paths plus invalid-args (no filesystem touch).
- **Path/host handling**: per-worktree `HISTFILE` paths are derived via the existing `hashWorktreeId`/`historyFilename` scheme already used elsewhere in the codebase (`src/main/terminal-history.ts`), not newly invented, so worktree isolation guarantees are reused rather than re-implemented. SSH-connected panes are explicitly excluded from history seeding (gated on `!connectionId`) rather than attempting a same-process remote read.
- **No new command execution**: this feature only reads a local history file and matches strings client-side; it never constructs or executes a shell command based on suggestion content — accepting a suggestion sends the same bytes the pane would already accept as typed input.
- **No new secrets/dependency surface**: no new third-party dependencies were added; no credentials, tokens, or auth flows are touched.
- Follow-up (tracked, not blocking): full SSH persisted-history and WSL-via-daemon history seeding are out of scope for this PR (see Summary) — when implemented, they'll need the same per-host isolation review this PR gave the local/daemon path.

## Notes

- Feature is gated behind `terminalAutosuggestEnabled` (default on) in Settings, so it can be disabled without a rebuild if it misbehaves for a given shell/host combination.
- Platform-specific: shell-kind resolution and keybinding matching are handled through existing cross-platform utilities rather than new platform-specific branches; no `e.metaKey`/`e.ctrlKey` hardcoding was introduced.
- A real merge conflict with `main` (an independent comment-slimming refactor touching the same functions this PR extends) was resolved manually, keeping this PR's logic and adopting `main`'s shortened comment style where the conflict was purely cosmetic.
- Manual testing across bash/zsh/fish/pwsh and native/WSL/SSH hosts (the Test plan's unchecked item) is recommended before merge, since it's the one thing that can't be verified by the unit test suite alone.

## ELI5

Terminals show fish-style ghost-text suggestions from recent and saved shell history. Accept with Right Arrow or End at end-of-line; toggle the feature in Settings.


xtan9:
i18n locales are not translated.

Piya-Boy:
Fixed — all non-English locales (es, ja, ko, zh) now have real translations for the strings added by this PR, and I've added a full Thai (th) locale as well. Verified catalog parity across all 5 locales with `verify:localization-catalog`.

Piya-Boy:
@coderabbitai The docstring coverage warning is expected — this repo's AGENTS.md explicitly directs against What-comments/docstrings on every function ("default to writing no comments... only when non-obvious, explain why, not how"). Adding boilerplate JSDoc across this PR's functions to hit an 80% threshold would go against that house style, so I'm leaving coverage as-is by design rather than padding it out.

Piya-Boy:
@brennanb2025 @nwparker This PR is ready for merge but I only have read access to stablyai/orca (pushed from a fork). Could one of you approve the pending workflow runs and merge (squash) when you get a chance?

nwparker:
Oh this is interesting! Will try to take a look soon.  Sorry for delay

PR #9612, opened 2026-07-20
fix(terminal): reap Windows agent PTY process trees on worktree teardown (#9045)

### LINKED ISSUE #9045: [Bug]: Worktree deletion fails with "Failed to physically stop every PTY" — orphaned claude.exe/node.exe/cmd.exe processes on Windows

OrcaWin:
Fixes #9045

## Problem
On Windows, deleting a worktree fails with `Failed to physically stop every PTY for worktree` (or `Timed out waiting for physical PTY teardown` on Force Delete) and leaves orphaned `claude.exe` / `node.exe` / `cmd.exe` processes holding handles into the worktree directory.

## Root cause (single)
Windows PTY teardown reduces to a ConPTY close of the **direct shell**. The descendant-snapshot module (`src/main/pty-descendant-termination.ts`) was a **deliberate win32 no-op** (`captureDescendantSnapshot` hard-returned `null`), so an agent's grandchildren (`claude -> node/cmd`) were never reaped. Those grandchildren:

1. keep the ConPTY **conout pipe open**, so node-pty's `exit` — the physical-exit proof the destructive delete gate (`killAllProcessesForWorktree`, `requirePhysicalStop`) waits on — never fires, and
2. keep **handles into the worktree dir**.

Both reported symptoms fall out of that one missing tree kill: the delete gate throws AND the processes orphan. This is distinct from #9197 (already fixed in #9601), which failed earlier at selector resolution.

## Fix
Minimal, Windows-scoped, inside the descendant module both providers already route agent teardown through:

- `captureDescendantSnapshot` now carries the **root PID** as a Windows snapshot (`windowsRootPid`) instead of returning `null`. Windows needs no ppid pre-snapshot because `taskkill /T` enumerates the live tree at kill time.
- `terminateDescendantSnapshot` reaps the tree with a **synchronous** `taskkill /pid <root> /t /f` — synchronous on purpose, so the tree is enumerated and killed **before** the caller closes the ConPTY (once the shell dies the child links break and taskkill can no longer reach the grandchildren, exactly the reparent hazard the POSIX snapshot-first path avoids). Failures are swallowed (the Windows analogue of ESRCH) so they never block the root kill, and the call is bounded so it can't outlast the physical-stop deadline.

Both the **daemon** (`terminal-session-teardown.ts` -> `killWithDescendantSweep`) and **local** (`local-pty-provider.ts` -> `shutdownTrackedPty`) providers get the fix through their existing **agent-scoped** calls. Non-agent terminals, POSIX, and remote hosts are untouched (the platform check runs on whichever host owns the PTY, so SSH/WSL remain correct). `taskkill /t /f` matches existing repo idiom (e.g. `claude-accounts/service.ts`).

## Tests
- New repro `src/main/runtime/repro-9045-windows-pty-tree-kill-gap.test.ts` models the ConPTY conout dependency and contrasts the pre-fix no-op (grandchildren survive, physical exit never proven) with the fix (tree reaped via real `taskkill`, conout frees, physical exit proven).
- Regression tests in `pty-descendant-termination.test.ts`: Windows snapshot shape, `taskkill /pid <root> /t /f` args + bounded/hidden opts, already-gone error swallowed, and the critical **ordering** assertion that the tree kill runs before the ConPTY root close.

Affected suites pass (`pty-descendant-termination`, `repro-9045`, `worktree-teardown`, daemon `session`/`terminal-session-teardown`/`pty-subprocess`, `local-pty-provider`): 301 passed. `npm run typecheck` and `oxlint` clean.

## Note on live repro
A fresh live Electron repro was impractical in this environment (dev CLI build incomplete); the deterministic gate-mechanism test above is used per the issue's allowance, and the issue already carries live `Get-CimInstance` evidence of the `powershell -> claude -> pwsh/node` tree shape on Windows.

### DIFF
--- a/src/main/pty-descendant-termination.test.ts
+++ b/src/main/pty-descendant-termination.test.ts
@@ -1,7 +1,11 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 const execFileMock = vi.hoisted(() => vi.fn())
-vi.mock('node:child_process', () => ({ execFile: execFileMock }))
+const execFileSyncMock = vi.hoisted(() => vi.fn())
+vi.mock('node:child_process', () => ({
+  execFile: execFileMock,
+  execFileSync: execFileSyncMock
+}))
 
 import {
   captureDescendantSnapshot,
@@ -12,6 +16,7 @@ import {
   killWithDescendantSweep,
   parseProcessTable,
   terminateDescendantSnapshot,
+  WINDOWS_PROCESS_TREE_KILL_TIMEOUT_MS,
   type ProcessTableCapture,
   type ProcessTableRow
 } from './pty-descendant-termination'
@@ -24,6 +29,8 @@ beforeEach(() => {
     const callback = args.at(-1) as (error: Error | null, stdout: string) => void
     callback(null, '10 1 10 Mon Jul 13 12:54:47 2026')
   })
+  execFileSyncMock.mockReset()
+  execFileSyncMock.mockReturnValue(Buffer.from(''))
 })
 
 function row(
@@ -115,10 +122,23 @@ describe('captureDescendantSnapshot', () => {
     expect(vi.getTimerCount()).toBe(0)
   })
 
-  it('is a null no-op on Windows', async () => {
+  it('carries the root PID as a Windows tree-kill snapshot without a ps walk (#9045)', async () => {
     const readTable = vi.fn()
-    expect(await captureDescendantSnapshot(10, { readTable, platform: 'win32' })).toBeNull()
+    const result = await captureDescendantSnapshot(4321, { readTable, platform: 'win32' })
+    // Why: Windows reaps via taskkill /T at kill time, so the snapshot needs only
+    // the root PID — no ppid table read (which reparenting would invalidate anyway).
+    expect(result).toEqual({
+      rootPgid: null,
+      descendants: [],
+      capturedAtMs: expect.any(Number),
+      windowsRootPid: 4321
+    })
     expect(readTable).not.toHaveBeenCalled()
+  })
+
+  it('returns null on Windows for an invalid root PID', async () => {
+    expect(await captureDescendantSnapshot(0, { platform: 'win32' })).toBeNull()
+    expect(await captureDescendantSnapshot(-1, { platform: 'win32' })).toBeNull()
   })
 
   it('degrades to null when ps fails', async () => {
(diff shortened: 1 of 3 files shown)


OrcaWin:
🔍 **Opus deep review — looks solid, no changes needed.**

Reviewed the whole change end-to-end (`pty-descendant-termination.ts` + both callers `daemon/session.ts` and `providers/local-pty-provider.ts`, plus the new tests). Scrutinized exactly the risk areas and they hold up:

- **Kill ordering** — `killWithDescendantSweep` runs the win32 `terminateDescendantSnapshot` branch (which invokes `taskkill /pid <root> /t /f`) inside the `try`, then `killRoot()` (the ConPTY close) in `finally`. So the tree is reaped while child links are still intact, *before* the pseudoconsole is torn down. Same order holds in the `local-pty-provider` path (`terminateDescendantSnapshot` → `requestTrackedPtyShutdown`). The `['tree-kill','root-kill']` assertion locks this in.
- **win32-scoping** — the new behavior is entirely behind `platform === 'win32'` / `snapshot.windowsRootPid !== undefined` on the PTY-owning host. The POSIX snapshot+SIGTERM/SIGKILL path is byte-for-byte unchanged, the relay/remote handlers never import these functions, and the sweep only ever runs for agent sessions (`launchAgent` / `ptyAgentSessionIds`). POSIX/SSH/remote are untouched.
- **No over-kill / orphans** — `taskkill /t` only descends the shell root's own child tree (the agent grandchildren); it can't reach conhost or unrelated processes. The already-dead-tree case is swallowed (Windows ESRCH analogue), and it composes correctly with node-pty's existing `nodePtyKillIssued` double-close guards, so the redundant `killRoot()` after taskkill is safe.
- **Bounded** — the sync `execFileSync` is capped at `WINDOWS_PROCESS_TREE_KILL_TIMEOUT_MS = 2s`, comfortably inside the 10s worktree physical-stop deadline, so a wedged taskkill can't outlast the sweep gate.
- **Tests** — the repro models the real conout-pipe dependency (physical exit is proven only once orphans are cleared) and exercises the *real* default helper chain down to a mocked `execFileSync`; unit tests cover ordering, exact args, timeout constant, already-gone swallow, and the `ownsRoot: false` skip. Good coverage.

Ran the affected suites — `pty-descendant-termination`, `repro-9045-windows-pty-tree-kill-gap`, `daemon/session`, `daemon/terminal-host`, `providers/local-pty-provider`, `runtime/worktree-teardown` (261 tests) — all green, plus `tsc` node typecheck and oxlint clean.

**Residual risk (acceptable):** the tree-kill is a synchronous `execFileSync` on the daemon event loop, so a pathologically wedged `taskkill` could block it for up to 2s during teardown; healthy-case cost is tens of ms and the sync call is what makes the before-ConPTY-close ordering trivially correct, so keeping it sync is the right call. Nice fix. ✅

OrcaWin:
Closing as **superseded** — `main` already reaps Windows agent PTY trees on teardown.

This PR's fix (`taskkill /T /F` the descendant tree **before** the ConPTY close, so orphaned `claude.exe`/`node.exe`/`cmd.exe` children can't hold the worktree cwd and fail destructive removal with "Failed to physically stop every PTY") landed on `main` via **#10004 / #10100**'s `killWithDescendantSweep`. It's wired into both the agent and plain-shell teardown paths in `src/main/daemon/terminal-session-teardown.ts`, identity-gated so it only taskkills while the OS probe confirms the tree is still ours. Same root cause, same mechanism, same call site — no behavioral delta.

Our branch is 428 commits behind `main` and `pty-descendant-termination.ts` conflicts, so rebasing would be net risk rather than a no-op. The #9045 diagnosis here was independently correct — it just lost the race.

Happy to send the `repro-9045-windows-pty-tree-kill-gap` test standalone if a maintainer wants the extra coverage. Thanks!

Issue #9539, opened 2026-07-20
[Feature]: Bundle ripgrep so local Quick Open works without a user-installed rg


squatboy:
### Operating system

macOS

### Orca version

v1.4.146

### Summary 
When searching for files by name in a git repo directory, the search fails with the error "File listing exceeded 10000 files". This occurs even when searching for a specific, known file name (single file).

### Workaround
I found that the file listing strategy depends on whether ripgrep is installed. Initially, the search failed with the error "File listing exceeded 10000 files" even when searching for a specific file name.

After installing ripgrep on the local macOS environment, the search started working correctly without hitting the file listing limit.

By looking at orca's file listing logic , I discovered that it checks for ripgrep (rg) installation when determining which file listing method to use.  `filesystem-list-files.ts:37-44`

When ripgrep is not available, the system falls back to alternative methods (git-based or readdir-based), which have file count limitations that can cause the "File listing exceeded 10000 files" error. filesystem-list-files.ts:42-44

> Installing ripgrep on the local Mac resolved the issue. After installing ripgrep, file search now works normally without the error.

squatboy:
I’ve opened a PR( #9627 ) regarding this issue and would appreciate it if you could take a look.

Taewoong1378:
i have a same issue

AmethystLiang:
### Follow-up: the durable fix is bundling ripgrep

PR #9627 improves the **alerting** for this case — when ripgrep is missing and the git/readdir fallback hits its budget, the opaque `File listing exceeded 10000 files` error is replaced with actionable guidance to install ripgrep. That makes the failure legible, but it **does not fix the root cause**, so I'm leaving this issue open.

**Root cause:** Orca doesn't ship ripgrep — it spawns the system `rg` from `PATH` (`checkRgAvailable`). When `rg` is absent, Quick Open falls back to `git ls-files` / a `readdir` walk, and that fallback carries a file-count budget (and timeout). Large repos, or repos with a big untracked directory that has to be expanded, blow that budget and fail — even when you're searching for a single known file, because listing happens before filtering.

**Proposed fix:** bundle ripgrep for the local host, the way VS Code does with `@vscode/ripgrep`. That eliminates this failure class locally: Quick Open would always have fast, gitignore-aware listing without depending on a user-installed binary. The fallback and the install-guidance path would then only matter for SSH/remote hosts, where shipping a binary isn't an option and the guidance is the right answer.

Tracking that as the actual fix here.

[...]

PR #9627, opened 2026-07-21
fix(quick-open): guide rg install after fallback budget errors

squatboy:
## Summary

Improves **user alerting** when local Quick Open falls back because ripgrep is unavailable. Previously the fallback surfaced a raw, opaque error (e.g. "File listing exceeded 10000 files"); now a readdir/git budget failure (file-count limit or timeout) is turned into actionable guidance that points the user at installing ripgrep, matching the guidance the relay/remote path already shows. Git, cancellation, and other file errors keep their original messages.

**Scope / non-goals:** This is an alerting change only. It does **not** change the file-listing algorithm, raise the 10,000-file budget, or otherwise fix the underlying case where searching a large repo without ripgrep fails. It only makes the existing failure legible and actionable. The durable fix for the root cause is to bundle ripgrep for the local host — tracked on #9539, not addressed here.

## Screenshots

No new layout or styling; the local error reuses the existing Quick Open install-guidance UI.

## Testing

- [x] `pnpm lint` — changed-scope lint is clean; the full gate stops only on two pre-existing switch-exhaustiveness errors already present on `origin/main`
- [x] `pnpm typecheck`
- [x] `pnpm test` — changed-scope tests pass: 6 files, 56 tests; the full suite has one unrelated local PTY non-ASCII-path failure
- [x] `pnpm build` — desktop/typecheck/web stages pass; native macOS sub-builds are blocked by local Swift Command Line Tools errors and the existing pnpm native-wrapper issue
- [x] Regression tests cover budget guidance, cancellation/Git error preservation, macOS install command, and legacy remote parsing

## AI Review Report

- macOS/Linux/Windows: preserves platform-specific install command detection and generic fallback; no listing algorithm or 10,000-file limit changes.
- Local/SSH: applies the new classification only to local fallback errors and preserves relay behavior and legacy parsing.
- Performance: guidance is built only after a budget failure; file listing work is unchanged.
- UI: reuses the existing copy-command guidance UI for the new local message.
- Security: uses fixed install text and read-only `/etc/os-release` detection; no shell execution or new permissions.

## Security Audit

No new process execution, permissions, network access, or user-controlled command interpolation was added.

## Notes

Related to #9539 — this PR improves the alert but does **not** close the issue. The root cause (Quick Open scanning the whole tree and hitting the fallback budget when ripgrep is absent) remains; bundling ripgrep locally is the follow-up fix and is left to that issue. Locale JSON and release/version files are unchanged.


AmethystLiang (approved):
Approving. Scope is error/alerting only — the file-listing algorithm, the rg spawn, git ls-files, the readdir walk, and the 10,000-file budget are all unchanged.

No performance regression: the only change in the hot path (`listQuickOpenFiles`) is `return` → `return await` inside a new try/catch, and that block is reached only when rg is absent. The added work (budget-error classification + `buildInstallRgMessage`, which reads `/etc/os-release` once on Linux) runs strictly in the catch — i.e. only after a scan has already failed with a budget error. The rg-available path and the success path do zero extra work. `checkRgAvailable` on line 45 is pre-existing, not introduced here.

Pushed a follow-up fixing a UI wording bug: the guidance component hardcoded "on the remote", so local scans told users to install ripgrep "on the remote". Now the location is parsed from the message and rendered correctly, with a render test guarding the regression.

src/main/ipc/filesystem-list-files.ts
```diff
@@ -42,13 +44,20 @@ export async function listQuickOpenFiles(
   // can run.
   const rgAvailable = await checkRgAvailable(authorizedRootPath, localGitOptions.wslDistro)
```
AmethystLiang, line 45: No performance regression from this line or the surrounding change.

- `checkRgAvailable` (this line) is pre-existing — not introduced by this PR.
- The only hot-path edit is `return` → `return await` inside a new try/catch, reached only when rg is absent.
- The added work (budget-error classification + `buildInstallRgMessage`, which reads `/etc/os-release` once on Linux) runs strictly inside the catch — i.e. only after a scan already failed with a budget error. The rg-available path and every successful scan do zero extra work.
- No change to the listing algorithm, rg spawn, `git ls-files`, the readdir walk, or the 10,000-file budget.

Issue #8763, opened 2026-07-14
[Bug]: Orca has stopped recognizing Codex agents


matheuspimentaa:
### Operating system

Windows

### Orca version

1.4.141.0

### Details

Since the last update, Orca has stopped recognizing Codex agents. I use the projects connected via SSH.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/d6eeb574-1d91-4ed0-8ef4-9290d580cde1" />

OrcaWin:
## Agent reproduction package (Windows orchestration 2026-07-16)

**Host context:** Windows 11 / Orca **1.4.143.0** / repo path used: `docs/bug-reproductions/`  
**Labels applied:** see issue labels (`has_repro` or `cannot_repro`).  
**Security:** Do **not** run user-uploaded binaries from issue attachments. Recreate fixtures from the steps/scripts below.

---
# Issue 8763 — Orca has stopped recognizing Codex agents

**Status:** `cannot_repro` (local Windows)  
**Host:** Windows 11 10.0.26200, Orca 1.4.143.0  
**Label:** `cannot_repro`  
**Issue:** https://github.com/stablyai/orca/issues/8763

## Summary

Reporter claims Orca stopped recognizing Codex agents after an update, on **SSH-connected projects**. On this local Windows host with installed Orca 1.4.143, Codex is recognized widely and appears on PATH inside Orca terminals.

## Environment

| Item | Value |
|------|-------|
| Orca | 1.4.143.0 (`C:\Users\neil\AppData\Local\Programs\orca\Orca.exe`) |
| Codex | `codex-cli 0.144.4` via `C:\Users\neil\AppData\Roaming\npm\codex.ps1` |
| Host setup | local only (no SSH remote project registered) |

## Repro attempt

```powershell
$env:PATH = "C:\Users\neil\AppData\Local\Programs\orca\resources\bin;" + $env:PATH
orca status
orca agent hooks status --json   # codex hooks: installed
orca worktree ps --json          # many agentType=codex
# Inside an Orca PowerShell terminal:
Get-Command codex   # -> C:\Users\neil\AppData\Roaming\npm\codex.ps1
```

## Evidence

- `docs/bug-reproductions/evidence/agent-summary.json` — **14** `codex` agent rows (plus 13 `claude`)
- `docs/bug-reproductions/evidence/hooks-status.json` — codex hooks `state: installed`
[...]

PR #9647, opened 2026-07-21
fix(sidebar): attribute spinner-only pane titles to the launched agent

### LINKED ISSUE #9646: Worktrees running Codex over SSH show no sidebar agent activity row — spinner-only titles carry no identity and launch identity is never consulted
### Describe the bug

On an SSH host, a worktree with a working Claude session shows its agent activity row under the worktree card, but a worktree with a working Codex session shows **nothing** — no agent row at all — even though Codex is actively running tools.

Side-by-side on the same SSH host:

| Worktree | Agent (working) | Terminal title (OSC) | Sidebar agent row |
| --- | --- | --- | --- |
| A | Codex | `⠼ <cwd-name>` | ❌ none |
| B | Claude | `✳ <task summary>` | ✅ shown |

### Why it happens

All three identity signals fail for Codex over SSH:

1. **Title**: Codex emits spinner + cwd titles (`⠼ <cwd-name>`) with no agent name. `resolveTitleDerivedAgentType` (`src/renderer/src/components/sidebar/worktree-title-derived-agent-rows.ts`) correctly refuses to attribute a bare spinner to Claude (`CLAUDE_AGENT_TOKEN_RE` gate), so the title-derived row is dropped entirely.
2. **Hooks**: Codex status hooks are dead in SSH worktrees — #8711. Confirmed live: with a remote Codex mid-task, the renderer's `agentStatusByPaneKey` stays empty, so no authoritative hook row exists either.
3. **Process**: the foreground-process fallback runs a local `ps`; a remote pane's shell PID is not in the local process table.

Claude still shows because its titles are self-identifying (`✳` prefix) *and* its hooks work remotely.

Meanwhile the tab already knows who it launched: `tab.launchAgent` is stamped by the agent launcher and `worktree create --agent`, survives the remote path (`worktree-remote.ts` passes `launchAgent` into `runtime.createTerminal` — confirmed live on an SSH workspace), and the hook-row path (`resolveRowAgentType` in `worktree-agent-rows.ts`) already prioritizes it. The title-derived fallback is the only sidebar path that never consults it.

### To Reproduce

1. Add an SSH host and create a worktree/workspace with Codex as the agent (`worktree create --agent codex` or the agent launcher).
2. Give Codex a task so it runs tools — the terminal title becomes `⠸ <cwd>`.
3. Look at the sidebar: the worktree shows no agent activity row (a Claude worktree on the same host shows one).

### Expected behavior

A worktree whose Orca-launched Codex is actively working should show a Codex activity row, like Claude does.

### Additional context

- #8711 (assigned) is the root-cause fix for the *hooks* signal and will bring full status detail. This issue is narrower: while hooks are absent, a spinner-only title plus the tab's launch identity is already enough evidence to show "Codex — Running" instead of nothing.
- Possibly the repro shape behind #8763 (`cannot_repro`): remote panes where the only signal is an anonymous spinner title.

choihjin:
## Summary

Fixes #9646.

On SSH hosts, a worktree with a working Codex session showed no sidebar agent activity row at all, while Claude worktrees on the same host showed theirs. All three identity signals fail for Codex over SSH: its OSC titles are spinner + cwd with no agent name, its status hooks are dead remotely (#8711), and the foreground-process fallback only scans the local process table.

The title-derived sidebar fallback (`worktree-title-derived-agent-rows.ts`) classified the spinner title as *working* but then dropped the row, because `resolveTitleDerivedAgentType` correctly refuses to attribute a bare braille spinner to Claude (`CLAUDE_AGENT_TOKEN_RE` gate). The tab already knows who it launched — `tab.launchAgent` is stamped by the agent launcher and `worktree create --agent`, and survives the remote path — but this was the one sidebar path that never consulted it.

**Fix:** when the title resolver yields no identity and the title carries a braille spinner, fall back to `tab.launchAgent` for the row's agent type (and label it with `formatAgentTypeLabel`). This mirrors the precedence the hook-row path (`resolveRowAgentType` in `worktree-agent-rows.ts`) already gives launch identity.

Scope guards, all preserved by existing tests:

- Explicit title identity still wins (`⠋ Codex` stays Codex regardless of `launchAgent`).
- A spinner title in a tab **without** launch metadata still produces no row ("does not infer Claude Code from a spinner-only non-agent title").
- Claude-specific prefixes (`✳` / `". "` / `"* "`) contain no braille spinner, so they are not rescued — "does not turn generic Codex-launched task titles into Claude Code rows" stays green.
- Hook rows still take precedence via `seenPaneKeys`; once #8711 lands, the authoritative hook path simply covers these panes first.

## Screenshots

Verified live: dev build connected to an SSH host, workspace created with `worktree create --agent codex`, Codex mid-task. Renderer state at capture time: `tab.launchAgent === 'codex'`, `agentStatusByPaneKey` empty (no hooks, #8711), pane title `⠸ <repo-name>` (no identity token).

Codex row shown for a hook-less SSH pane

Counter-check observed in the same session: when Codex went idle (title back to the bare repo name, no spinner), the row disappeared — idle panes without identity keep the existing no-row behavior.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`
- [x] Added a regression test that fails without the fix

New tests in `worktree-title-derived-agent-rows.test.ts`:

- "attributes a spinner-only title to the launched agent when the title has no identity" — codex-launched tab, live PTY, title `⠼ demo-repo` → exactly one row `['codex', 'working', 'Codex', '⠼ demo-repo']`. Confirmed red before the fix (no row) and green after.
- "keeps explicit title identity over the launched agent" — claude-launched tab with `⠋ Codex` title stays a Codex row (pins the precedence).

Full suite: 30,679 tests pass.

## AI Review Report

Reviewed by Claude (Fable 5, author).

- **Root cause verified live, not guessed**: with a remote Codex mid-task, the renderer store showed `launchAgent: 'codex'` on the tab, an empty `agentStatusByPaneKey`, and a spinner-only pane title — exactly the state the row builder drops today.
- **Blast radius**: the fallback triggers only when (a) the title resolver returned no identity — which happens solely for generic status claims — (b) the title contains a braille spinner, and (c) the tab has launch metadata. All other title shapes and all hook-backed panes are untouched.
- **Known tradeoff**: `launchAgent` is tab-scoped, so a second split pane in an agent-launched tab whose process emits a braille-spinner title would be attributed to the launched agent. This mirrors the existing tab-scoped use of `launchAgent` in the hook-row path; spinner-titled non-agent processes are rare (most CLIs don't set OSC titles).
- **Not covered (documented)**: agents started by manually typing `codex` in a plain terminal have no `launchAgent`; their identity lives in pane-connection-local `commandInferredAgent`, which the sidebar cannot reach today. That belongs to the pane-owner-resolver slice deliberately deferred in `pane-agent-evidence.ts`.
- **Cross-platform (macOS/Linux/Windows)**: pure renderer logic; no platform branches, no Git commands, no provider-specific naming.
- **SSH/remote/local**: verified over SSH (the motivating case); local panes behave identically — locally Codex hooks usually produce authoritative rows first, which keep precedence.

## Security Audit

No new input handling, IPC, auth, command execution, or path handling. The change reads two existing renderer-state fields (`tab.launchAgent`, the already-scraped pane title) to pick a display label and icon. No secrets or logging affected.

## Notes

- #8711 (assigned) remains the root-cause fix for the hooks signal and will bring full status detail (current tool, task text). This change only stops the sidebar from being completely blind in the meantime, and yields to hook rows the moment they exist.
- Possibly explains #8763 (`cannot_repro`): remote panes where the only signal is an anonymous spinner title.

### DIFF
--- a/src/renderer/src/components/sidebar/worktree-title-derived-agent-rows.test.ts
+++ b/src/renderer/src/components/sidebar/worktree-title-derived-agent-rows.test.ts
@@ -201,6 +201,43 @@ describe('buildTitleDerivedAgentRows', () => {
     }
   })
 
+  it('attributes a spinner-only title to the launched agent when the title has no identity', () => {
+    const launchAgent: TuiAgent = 'codex'
+    const rows = buildWorktreeAgentRows({
+      tabs: [makeTab('tab-1', { launchAgent })],
+      entries: [],
+      retained: [],
+      runtimePaneTitlesByTabId: {
+        // Codex over SSH emits spinner + cwd titles with no agent name (#8711).
+        'tab-1': { 1: '⠼ demo-repo' }
+      },
+      ptyIdsByTabId: { 'tab-1': ['pty-codex-remote'] },
+      terminalLayoutsByTabId: { 'tab-1': makeSingleLayout(LEAF_ID_1) },
+      now: 2000
+    })
+
+    expect(
+      rows.map((row) => [row.agentType, row.state, row.entry.prompt, row.entry.terminalTitle])
+    ).toEqual([['codex', 'working', 'Codex', '⠼ demo-repo']])
+  })
+
+  it('keeps explicit title identity over the launched agent', () => {
+    const launchAgent: TuiAgent = 'claude'
+    const rows = buildWorktreeAgentRows({
+      tabs: [makeTab('tab-1', { launchAgent })],
+      entries: [],
+      retained: [],
+      runtimePaneTitlesByTabId: {
+        'tab-1': { 1: '⠋ Codex' }
+      },
+      ptyIdsByTabId: { 'tab-1': ['pty-explicit'] },
+      terminalLayoutsByTabId: { 'tab-1': makeSingleLayout(LEAF_ID_1) },
+      now: 2000
+    })
+
+    expect(rows.map((row) => [row.agentType, row.state])).toEqual([['codex', 'working']])
+  })
+
   it('does not turn generic Codex-launched task titles into Claude Code rows', () => {
     const launchAgent: TuiAgent = 'codex'
     const rows = buildWorktreeAgentRows({
--- a/src/renderer/src/components/sidebar/worktree-title-derived-agent-rows.test.ts
+++ b/src/renderer/src/components/sidebar/worktree-title-derived-agent-rows.test.ts
@@ -238,6 +238,23 @@ describe('buildTitleDerivedAgentRows', () => {
     expect(rows.map((row) => [row.agentType, row.state])).toEqual([['codex', 'working']])
   })
 
+  it('produces no row for a spinner-only title when the tab has no launch identity', () => {
+    const rows = buildWorktreeAgentRows({
+      tabs: [makeTab('tab-1')],
+      entries: [],
+      retained: [],
+      runtimePaneTitlesByTabId: {
+        // Spinner activity but no identity and no launchAgent to attribute it to.
+        'tab-1': { 1: '⠼ demo-repo' }
+      },
+      ptyIdsByTabId: { 'tab-1': ['pty-anon'] },
+      terminalLayoutsByTabId: { 'tab-1': makeSingleLayout(LEAF_ID_1) },
+      now: 2000
+    })
+
+    expect(rows).toHaveLength(0)
+  })
+
   it('does not turn generic Codex-launched task titles into Claude Code rows', () => {
     const launchAgent: TuiAgent = 'codex'
     const rows = buildWorktreeAgentRows({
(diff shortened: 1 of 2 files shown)


AmethystLiang:
thanks for the PR! will review it soon

AmethystLiang (approved):
Reviewed via multi-agent code review. Correct, well-scoped fix that makes the title-derived sidebar path consistent with the hook path's existing launchAgent fallback. Added a guard test (spinner-only + no launchAgent -> no row) and comments recording the spinner-gating rationale and split-pane residual. Tests/lint/typecheck green.

PR #9654, opened 2026-07-21
fix(i18n): localize browser link-routing description

### LINKED ISSUE #9652: [Bug]: "Link Routing" setting description stays in English under Korean UI
### Operating system

Windows (also affects macOS/Linux — the string is platform-independent)

### Orca version

v1.4.147-rc.4

### Details

**Short summary:**
In Settings → Workflow → Browser, the "Link Routing" title is localized, but its description text stays in English even when the UI language is set to a non-English locale (e.g. Korean).

**What happened?**
The description reads:

> Open http(s) links in Orca's built-in browser — from the terminal, markdown, and the editor. {shortcut} always uses your system browser.

in English regardless of the selected UI language, while every other string on the same pane is translated.

**How can we reproduce it?**
1. Set the UI language to Korean (Settings → Appearance → interface language).
2. Go to Settings → Workflow → Browser.
3. The "Link Routing" description below the title is shown in English.

Root cause: `getBrowserLinkRoutingDescription()` in `src/renderer/src/components/settings/browser-search.ts` returns a hardcoded template literal, so the string never enters the i18n catalog and is never translated. Every other entry in that file goes through `translate()`. The platform-specific shortcut label (`⇧⌘-click` / `Shift+Ctrl+click`) is interpolated into the sentence, which is likely why it was hardcoded.

seandoesdev:
## Summary

The **Link Routing** setting description (Settings → Workflow → Browser) was a hardcoded English template literal, so it bypassed the i18n catalog and stayed in English even when the UI language was set to a non-English locale — while its own title and every neighboring setting were already translated.

This routes the description through `translate()` with a `{{shortcut}}` interpolation (the platform keybinding stays a literal, uninterpolated value) and registers the string in every locale catalog.

Closes #9652

### What changed
- `src/renderer/src/components/settings/browser-search.ts` — `getBrowserLinkRoutingDescription()` now calls `translate('auto.components.settings.browser.search.00f82a4c3d', <fallback>, { shortcut })` instead of returning a hardcoded template literal.
- `src/renderer/src/i18n/locales/{en,ko,ja,es,zh}.json` — added the new key. The key hash matches the repo's existing convention (`sha1(<relative file path>:<source string>)`, first 10 chars). `ko` is curated; `ja`/`es`/`zh` are included for catalog parity.
- `config/scripts/locale-ko-key-overrides.json` — added the curated Korean value so catalog regeneration (machine translation) preserves it.
- `src/renderer/src/components/settings/browser-search.test.ts` — added a regression guard asserting the shortcut is interpolated (no raw `{{shortcut}}` leaks).

## Screenshots

> ⚠️ TODO (author): attach before/after screenshots of the Korean **Settings → Workflow → Browser → Link Routing** pane.

Before: description shown in English under Korean UI. After: description shown in Korean, with the platform shortcut label (`Shift+Ctrl+click` on Windows/Linux, `⇧⌘-click` on macOS) preserved.

## Testing

- [x] `pnpm typecheck`
- [x] `pnpm test` (scoped: `browser-search.test.ts` 2/2, `locale-ko-key-overrides` + `verify-localization-catalog` tests 4/4)
- [x] `pnpm build`
- [x] `pnpm verify:localization-catalog` (10,138 key references + es/ja/ko/zh parity verified)
- [x] `oxlint` clean on the changed files (pre-commit `lint-staged` oxlint + oxfmt passed)
- [x] Added a regression test for the interpolation behavior

Note: the full `pnpm lint` and `pnpm verify:localization-coverage` currently fail on `main` for reasons unrelated to this change (pre-existing `switch-exhaustiveness` errors in `skill-freshness-group.tsx` / `daemon-server.ts`, and unlocalized strings in the `ai-vault` session files). None of those files are touched by this PR.

## AI Review Report

Reviewed with an AI coding agent. Main checks:
- **Cross-platform (macOS/Linux/Windows):** the shortcut label is produced by the existing `getBrowserLinkRoutingShortcutLabel()` platform branch and passed as an interpolation value, so it stays literal and renders correctly per OS (`⇧⌘-click` vs `Shift+Ctrl+click`). No paths, shell behavior, keyboard-handling, or Electron main/renderer IPC were touched.
- **i18n correctness:** the new key follows the repo's hash convention (verified by reproducing an existing key's hash), placeholders (`{{shortcut}}`) are consistent across all five locales, and locale parity passes. The interpolation param name matches the placeholder.
- **SSH/remote & providers:** no runtime, process, git-provider, or network behavior involved — pure renderer copy.
- **Behavior:** no functional change beyond localizing an existing string; the fallback preserves the original English wording verbatim.

## Security Audit

No security-relevant surface. The change adds static i18n catalog data and one `translate()` call. No user input handling, command execution, path handling, authentication, secrets, dependencies, or IPC were added or modified.

## Notes

- `ja`/`es`/`zh` values are included to satisfy catalog parity; maintainers may prefer to regenerate them through the existing machine-translation pipeline. The `ko` value is curated and pinned via `locale-ko-key-overrides.json`.
- No changes to release/version files.

X: @your_handle


nwparker:
Thank you for this, @seandoesdev — and welcome. Your write-up was the most thorough of the three PRs that landed on this bug, and one piece of it shipped.

We're consolidating this cluster on **#9444**, primarily on the placeholder name. `{{shortcut}}` reads better than `{{value0}}` — I'd agree with you on pure legibility — but the `auto.*` namespace is machine-generated by `localize-renderer-strings.mjs`, which names interpolations positionally (`value${index}`). The catalog bears that out: across ~10.9k auto keys the placeholders are `value0` (769), `value1` (141), `value2` (30) and so on, and the handful of named ones sit in hand-authored keys. A named placeholder in an `auto.*` key would be reverted the next time the codemod regenerated the entry, taking the four translations with it.

Two notes on your hash, since your PR body specifically called out reproducing it:

- The formula you documented — `sha1(<relative file path>:<source string>)`, first 10 chars — is correct. The subtlety is *which* string. For a template literal the candidate is only the **whitespace-compacted head** (`uniqueCandidates()` dedupes by `start:end:kind`, so only the head survives; `compactText()` then trims it), not the assembled string with the placeholder spliced in. So the generated key is `904ce58440` regardless of what you name the placeholder — `00f82a4c3d` is the hash of the full `{{shortcut}}` form.
- Because the hash doesn't depend on the placeholder name, the ko override wasn't needed here either. `locale-ko-key-overrides.json` pins curated Korean against the machine-translation pass; only ~27% of the keys in this namespace carry one, and the reviewed-in-context translation in the catalog stands on its own.

**Your regression test shipped.** The `expect(description).not.toContain('{{shortcut}}')` assertion was the genuinely valuable idea in this cluster — none of the other PRs guarded against a leaked placeholder, which is the exact failure mode if the interpolation name ever drifts from the catalog. It's in #9444 generalized to `not.toMatch(/\{\{.+?\}\}/)` so it catches any placeholder rather than one specific name, in both platform branches. That prompted a broader locale-switching test as well. Credited in the PR.

Closing in favour of #9444 — the test instinct here was the right one, and I'd like to see more from you.

PR #770, opened 2026-04-17
feat: add GitHub Copilot CLI as a supported agent

drakontia:
## What changed

GitHub Copilot CLI (\copilot\) is now listed as a supported agent in **Settings → Agents**.

Users who have [GitHub Copilot CLI installed](https://docs.github.com/en/copilot/how-tos/set-up/install-copilot-cli) can now select it from the agent picker and Orca will launch it with \copilot --prompt '<prompt>'\ when opening a workspace with an initial prompt.

## Changes

| File | Change |
|------|--------|
| \src/shared/types.ts\ | Added \'copilot'\ to \TuiAgent\ union |
| \src/shared/tui-agent-config.ts\ | Added \copilot\ launch config with \promptInjectionMode: 'flag-prompt'\ |
| \src/renderer/src/lib/agent-catalog.tsx\ | Added catalog entry (label, favicon, homepage URL) |
| \src/renderer/src/components/settings/agents-search.ts\ | Added \'copilot'\, \'github'\, \'github copilot'\ search keywords |
| \src/renderer/src/lib/tui-agent-startup.test.ts\ | Added test: copilot uses \--prompt\ flag for prompt injection |

No visual design changes — the entry appears in the existing agents list using the \github.com\ favicon via Google's favicon service.

## Tests

Added a test in \	ui-agent-startup.test.ts\ verifying that Orca builds the correct command:

\\\
copilot --prompt 'Fix the bug'
\\\

All 9 tests in the file pass. ✅

## AI Code Review Summary

Reviewed by AI coding agent — key findings:

**Cross-platform compatibility**: ✅
- The binary name \copilot\ is consistent across macOS, Linux, and Windows (installed via GitHub CLI extension to PATH)
- No hardcoded path separators or platform-specific assumptions introduced
- Prompt injection uses the existing \lag-prompt\ mode (\--prompt '<text>'\), which is shell-portable

**Security audit**: ✅
- No new IPC surfaces, file system access, or network calls introduced
- The \copilot\ command runs as a user-launched subprocess, consistent with all other agents
- No secrets or credentials are passed through the prompt injection path

## Platform notes

- Tested on Windows (TypeScript typecheck ✅, lint ✅, unit tests 9/9 ✅)
- The \copilot\ binary name is the same on macOS/Linux/Windows per GitHub's official docs
- No platform-specific UI label or shortcut changes were made

PR #9655, opened 2026-07-21
feat(tui-agents): recognize qodercli as a TUI agent

xingqingzzp-gif:
## Summary

- Qoder CLI (binary `qodercli`, https://docs.qoder.com) is a coding-agent CLI in the same category as Claude Code / Codex / Gemini CLI, but it was not in `TUI_AGENT_CONFIG`. Orca's foreground-process detection fell back to the closest heuristic match, so a running `qodercli` tab was mislabeled as **Gemini CLI**.
- Add a minimal `qoder` TuiAgent entry (`detectCmd`/`launchCmd`/`expectedProcess` = `qodercli`), its telemetry `AgentKind`, and its plain-English display name (`Qoder CLI`).
- Add a placeholder 64×64 `qoder.png` icon matching the size of existing agent icons. A proper product logo swap is welcome — kept as a separate concern.

## Why minimal scope

Detection + launch + display name + telemetry kind is the smallest change that fixes the mislabeled tab. Two follow-up areas are intentionally left out of this PR so it can land independently:

- **Managed hook installation** (`AGENT_HOOK_TARGETS` / `AgentHookSource`). Qoder CLI supports Claude-Code-style hooks (https://docs.qoder.com/en/cli/hooks), so Orca could install a managed hook script for live working/idle status. That touches `managed-agent-hook-controls.ts` and the installer; deserves its own PR.
- `agentDefaultArgs` default. Belongs to the settings-schema PR.

## Test plan

- [x] `pnpm vitest run src/shared/agent-process-recognition.test.ts` — 16/16 pass
- [x] `pnpm vitest run src/shared/` — 2654/2654 pass (2 skipped pre-existing)
- [x] `pnpm vitest run src/main/agent-hooks/` — 451/451 pass (3 skipped pre-existing)
- [x] `pnpm run typecheck:node` — clean
- [ ] Launch Orca dev, run `qodercli` in a terminal tab, confirm the tab title shows "Qoder CLI" instead of "Gemini CLI"

## Notes for maintainers

- Icon: `src/shared/agent-icons/qoder.png` is a 64×64 placeholder (purple rounded square with a "Q"). Replace with the official Qoder logo when available.
- `promptInjectionMode: 'stdin-after-start'` matches the conservative default used by `ante`/`goose`/`amp`/`devin`. If Qoder CLI later ships a documented `--prefill`-style flag, `claude`/`openclaude` are the precedent for switching to `argv` with `draftPromptFlag`.

## ELI5

Qoder CLI was mislabeled as Gemini because it was not a known TUI agent. Orca recognizes qodercli as its own agent for detection, launch, and display name.


houlemon0130:
As a daily Qoder CLI user, this fix would be great to see land — right now my `qodercli` tabs get mislabeled as Gemini CLI. The scope here mirrors #770 (GitHub Copilot CLI, also ~5 files) which was merged, so hopefully this is an easy review. Thanks for the contribution!

PR #9029, opened 2026-07-16
fix(runtime): keep remote activation publish-only

bbingz:
## Summary

- make paired runtime-client worktree and session-tab activation publish selection state without navigating or focusing the host renderer by default
- require an explicit `followHost: true` opt-in before a runtime client can drive host-window activation
- send both `notifyClients: false` and `followHost: false` from remote-session clients, so the safe default also works against older hosts that do not know `followHost`
- preserve existing mobile and in-process activation behavior

Related to #8871. This is the activation/focus companion to #8872 and #8888.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`
- [x] Added or updated high-quality tests that would catch regressions, or explained why tests were not needed

Focused verification:

- `pnpm exec vitest run --config config/vitest.config.ts` across 13 affected runtime/RPC/renderer test files: 889 tests passed
- `git diff --name-only -z origin/main | xargs -0 pnpm exec oxlint`
- `pnpm run check:max-lines-ratchet`
- `git diff --check`

## AI Review Report

The review traced worktree and tab activation from the renderer client, through RPC parsing and client-kind context, into the runtime notifier boundary. It checked both halves of mixed-version compatibility: a new client stays publish-only against an older host via `notifyClients: false`, while a new host defaults older runtime clients to publish-only unless they explicitly send `followHost: true`.

The main risk was accidentally changing phone activation or local/in-process navigation. The implementation therefore gates only `clientKind === 'runtime'`; focused tests verify runtime defaults and explicit opt-in, while the existing mobile activation suite remains green. The review also checked macOS, Linux, Windows, and SSH impact. This change introduces no platform-specific shortcuts, labels, paths, shell behavior, or Electron platform branches.

## Security Audit

The new field is schema-validated as a boolean and reaches only the existing host focus/navigation notifier boundary. It does not add command execution, file/path handling, credentials, secrets, dependencies, destructive terminal actions, or a new IPC transport. The default for paired runtime clients is deny-by-default host navigation; explicit `followHost: true` restores the existing navigation capability for trusted callers that intentionally request it.

## Notes

- This PR is intentionally limited to activation/focus authority. Destructive close intent and host-side destructive-action policy are handled separately in #8872 and #8888.
- No UI currently opts into `followHost`; the field is an explicit RPC capability for future intentional host-follow flows.

PR #9664, opened 2026-07-21
fix(runtime): isolate navigation across paired clients

OrcaWin:
## Summary

- Make worktree, session-tab, and terminal-focus navigation caller-local by default for paired runtime and mobile clients.
- Keep shared runtime topology authoritative while projecting active tab/group state per paired device and runtime server.
- Add schema-validated `caller | host | clients | all` navigation intent with live-only explicit follow delivery.
- Preserve mixed-version safety and intentional CLI/renderer “reveal on host” behavior.

## Screenshots

No visual change.

## Root cause

Remote worktree activation called the host notifier and broadcast an unscoped client navigation event. Session-tab activation also rewrote the single shared `mobileSessionTabsByWorktree` active selection, so every subscriber consumed the last writer as navigation. Existing echo suppression stopped a loop but did not establish ownership.

## Deterministic reproduction

`src/main/runtime/multi-client-navigation-isolation.integration.test.ts` starts real `OrcaRuntimeRpcServer` instances and authenticates independent E2EE WebSocket sessions from distinct pairing offers/device IDs. It covers:

- host on H, client A on A, client B on B, then A activates A2 without moving B or host;
- independent session-tab selection plus reconnect;
- explicit host/client follow intent;
- crossed A/B clients across two independent servers.

The harness uses message barriers rather than sleeps and exercises the real dispatcher, pairing registry, encrypted WebSocket transport, subscriptions, runtime routing, graph-backed snapshots, and per-device projection.

Run locally on macOS or Linux; Windows and Docker are not required:

```bash
pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/multi-client-navigation-isolation.integration.test.ts
pnpm test:e2e:multi-client-navigation
```

The E2E command builds Orca and its paired web client, launches one real Electron runtime host, and opens two clients in isolated Electron Chromium partitions.

## Testing

- [ ] `pnpm lint` — focused oxlint, React Doctor, max-lines ratchet, and reliability gates pass; the repository-wide switch-exhaustiveness command reports pre-existing errors in untouched `skill-freshness-group.tsx` and `daemon-server.ts`.
- [x] `pnpm typecheck`
- [ ] `pnpm test` — proportionate targeted suites were run rather than every repository test.
- [x] `pnpm build` — Electron E2E and paired web production builds completed successfully.

Exact results:

- Post-review targeted runtime/RPC/renderer/CLI suite: 11 files, 308 passed.
- Earlier expanded targeted root suite: 1084 passed, 2 skipped.
- Full mobile suite: 2130 passed, 2 skipped.
- Local browser E2E: 1 passed in 53.1s (test body 7.3s), using one Electron host and two independently paired clients.
- Root/mobile typecheck and focused oxlint passed.
- Web production build/verification, max-lines ratchet, reliability gates, and `git diff --check` passed.

## AI Review Report

Ran `$internal-review-until-clean` anchored at merge base `0e7b0ac22079f4ab6dba708bb595daaa627e3285`, including pre-review and post-fix elegance interrogation, a performance audit, and two adversarial review rounds.

The review checked navigation ownership, paired identity vs. connection identity, reconnect/restart/stale-snapshot behavior, graph fanout, explicit follow routing, mixed versions, session-tab fallback, terminal creation/focus, privacy boundaries, and cross-platform behavior on macOS, Linux, Windows, WSL, and SSH. No platform heuristic, shortcut, label, shell, path, or platform-specific Electron behavior was added.

It validated two CodeRabbit findings and fixed them:

- paired runtime `terminal.focus` could still steer the host implicitly; it now uses the central navigation resolver, while intentional CLI/renderer reveals send explicit `navigation: 'host'`;
- streaming RPC handlers were not given the pairing context already available to one-shot handlers; the dispatcher now forwards it with focused coverage.

It also corrected the selected-group fallback test and added a concise ownership-precedence comment. The final post-fix round found no additional actionable issue.

The final design supersedes PR #9029 rather than reusing it unchanged. #9029 suppresses ordinary host focus but still rewrites/fans out one global active session-tab snapshot, so it needs rework to provide independent client selections.

## Security Audit

Reviewed RPC schema validation, authenticated device identity, pairing credentials, transient connection IDs, IPC boundaries, navigation-state disclosure, command/path handling, secrets, and dependency impact.

- Navigation state is keyed by the revocable paired device ID, never the bearer credential or connection ID.
- Per-client selection remains memory-only and is removed on pairing revocation or worktree deletion.
- One paired client receives shared resource topology plus only its own projected selection; explicit follow is live-only and is not persisted/replayed after reconnect.
- Navigation targets are additive, schema-validated enums and deny paired-client fanout by default.
- No new subprocess input, path handling, secret storage, dependency, or public network surface was added.

No security follow-up is required for this change.

## Notes

- Stable paired-device state survives transport reconnect while the runtime process remains alive; a server restart falls back deterministically instead of persisting private navigation globally.
- New clients send legacy-safe `notifyClients:false`; terminal creation also sends `activate:false`. New hosts treat older paired clients as caller-local by default.
- Local/in-process callers retain legacy behavior. Host/client follow remains available only through explicit intent.
- Docker would add packaging variance but no useful ownership coverage here; the local E2E already uses real Electron runtime, pairing, encryption, WebSockets, and isolated browser storage.


OrcaWin:
Added local browser-level multi-client E2E coverage in `dad237565`.

The new `pnpm test:e2e:multi-client-navigation` runner builds the Electron E2E app plus bundled web client, launches one real runtime server/host UI, pairs two distinct runtime devices into isolated Electron Chromium sessions, and drives H / A / B -> A2 through the rendered worktree UI. DOM assertions verify client B remains on B and the host remains on H. It requires neither `windows 2`, Docker, pairing secrets, a coworker Mac, nor a separate Playwright browser download.

Validation:
- `pnpm test:e2e:multi-client-navigation` — 1 passed (56.7s total; test 11.8s)
- `pnpm typecheck` — passed
- touched-file `oxlint` — passed
- `pnpm check:max-lines-ratchet` — passed
- `pnpm check:reliability-gates` — passed
- `git diff --check` — passed

PR #9681, opened 2026-07-21
fix: localize Link Routing settings description

### LINKED ISSUE #9652: [Bug]: "Link Routing" setting description stays in English under Korean UI

Chang-Jin-Lee:
## Summary
- `getBrowserLinkRoutingDescription()` in `src/renderer/src/components/settings/browser-search.ts` returned a hardcoded template literal instead of going through `translate()`, so the description stayed in English regardless of the selected UI language, even though every other string on the same settings pane (Settings → Workflow → Browser) was translated.
- Wrapped the string in `translate()`, interpolating the platform-specific shortcut label via `{{value0}}` (mirroring the existing pattern used elsewhere, e.g. `App.tsx`'s `Toggle sidebar ({{value0}})`).
- Added the new key to `en.json` and ran `pnpm run sync:localization-catalog` to keep `es.json`/`ja.json`/`ko.json`/`zh.json` in parity (English placeholder copied in, per the documented sync behavior — actual translation happens in a separate machine-translation pass per `config/localization-audit.md`).

Fixes #9652

## Test plan
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/settings/browser-search.test.ts` — 2 passed
- [x] `pnpm run typecheck` — clean
- [x] `pnpm run verify:localization-catalog` — clean
- [x] `pnpm run verify:localization-coverage` — no new unlocalized strings introduced by this change (3 pre-existing unrelated findings in `ai-vault-session-launch-actions.ts`/`AiVaultSessionDropLayer.tsx` confirmed present on unmodified `main` too, out of scope here)
- [x] `oxlint` on the changed file — clean

### DIFF
--- a/src/renderer/src/components/settings/browser-search.ts
+++ b/src/renderer/src/components/settings/browser-search.ts
@@ -17,7 +17,11 @@ export function getBrowserLinkRoutingShortcutLabel(platform: BrowserShortcutPlat
 }
 
 export function getBrowserLinkRoutingDescription(platform: BrowserShortcutPlatform): string {
-  return `Open http(s) links in Orca's built-in browser — from the terminal, markdown, and the editor. ${getBrowserLinkRoutingShortcutLabel(platform)} always uses your system browser.`
+  return translate(
+    'auto.components.settings.browser.search.d0fdcd86b1',
+    "Open http(s) links in Orca's built-in browser — from the terminal, markdown, and the editor. {{value0}} always uses your system browser.",
+    { value0: getBrowserLinkRoutingShortcutLabel(platform) }
+  )
 }
 
 export function getBrowserPaneSearchEntries(
--- a/src/renderer/src/i18n/locales/en.json
+++ b/src/renderer/src/i18n/locales/en.json
@@ -7473,6 +7473,7 @@
             "bea27bac4b": "links",
             "44d14df30d": "preview",
             "5cb082b3e3": "Link Routing",
+            "d0fdcd86b1": "Open http(s) links in Orca's built-in browser — from the terminal, markdown, and the editor. {{value0}} always uses your system browser.",
             "95944898e0": "percentage",
             "483a0eb5e0": "new tab",
             "726f2a8556": "page zoom",
--- a/src/renderer/src/i18n/locales/es.json
+++ b/src/renderer/src/i18n/locales/es.json
@@ -7441,7 +7441,8 @@
             "a942905148": "URL abierta al crear una nueva pestaña del navegador. Déjelo vacío para abrir una pestaña en blanco.",
             "c3903322d2": "Página de inicio predeterminada",
             "19ea5607cf": "Etiquetas de localhost para worktrees",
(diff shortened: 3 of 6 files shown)


nwparker:
Thank you for this, @Chang-Jin-Lee — and welcome. Your diagnosis was exactly right: `getBrowserLinkRoutingDescription()` was a bare template literal, the codemod skipped it because of the interpolated shortcut, and `{{value0}}` is the correct placeholder convention (it's literally what `localize-renderer-strings.mjs` emits — `value${index}`). You and #9444 arrived at the same approach independently.

We're consolidating this cluster on **#9444**, for two reasons:

1. **Translations.** #9444 ships real es/ja/ko/zh copy. This PR follows the documented `sync:localization-catalog` behavior of propagating the English fallback into the other catalogs — which is legitimately the repo's convention for feature PRs, and you were right to cite it. But since the whole point of this particular fix is that the string renders in English under a non-English UI, leaving the English fallback in es/ja/ko/zh means the reported bug is still visible to the user. This one case needs the translations inline.

2. **The catalog key.** `d0fdcd86b1` is `sha1(path + ':' + head)` over the **untrimmed** head text. The codemod runs the candidate through `compactText()` (`replace(/\s+/g, ' ').trim()`) before hashing, so the key it would actually generate is `904ce58440` — the trailing space before `${...}` is what separates the two. I verified this by re-deriving the hash for existing `{{value0}}` keys in the catalog: the trimmed head reproduces 29/29 of the decisive samples, the untrimmed head 0.

Worth stressing that this is an easy thing to get wrong and nothing in CI would have told you: `verify-localization-catalog.mjs` only checks key existence, placeholder consistency, and locale parity — it never recomputes the hash. Your PR passed that gate, as you reported. The key would only have drifted the next time someone re-ran the codemod.

Your `{{value0}}` choice is the one that shipped, and your independent confirmation of the root cause raised confidence in the fix. Closing in favour of #9444 — please do send more, and thanks for taking the time.

PR #9684, opened 2026-07-21
fix(terminal): scope macOS daemons to login sessions

### LINKED ISSUE #7936: [Bug] macOS logout leaves orphaned PTYs → broken "Login incorrect" terminals on next login, needs reboot
<kbd>[fred](https://discord.com/channels/1492228674081263786/1524755281601364079/1524755281601364079)</kbd> [fred](https://discord.com/channels/1492228674081263786/1524755281601364079/1524755281601364079)  `via Discord`

Env: macOS, Orca, agent terminals running (Claude Code, Hermes)
Bug: After a full macOS logout + log back in (not fast switch), restored terminals come up on a raw login: prompt showing "Login incorrect" instead of the shell. Only a reboot fixes it.
Cause (suspected): The PTY-owning daemon survives logout instead of dying with the session. On next login Orca warm-reattaches to now-orphaned PTYs, so they come back broken. Reboot works because it kills the daemon, matching the documented "daemon dies when the host does" behavior.
Workaround: pkill -f orca before logout (or at login, pre-reattach) prevents it.

Asks:
Daemon-aware teardown on macOS logout, or a documented orca daemon stop command, so the daemon dies with the session like it does on reboot.
Does the daemon survive fast user switching cleanly? If not, can it? or is that architecturally awkward given how the daemon is scoped?

Repro: open Orca with agents running → fully log out → log back in → terminals show login: / "Login incorrect".

OrcaWin:
## Summary

Scopes macOS terminal daemons and PTYs to the login runtime that created them.

- A full logout or changed login scope can no longer warm-attach to stale interactive PTYs.
- Same-session app relaunch, normal quit/reopen, WindowServer recovery, concurrent clients, and returning through fast user switching preserve valid terminals.
- Reboot and changed login scope use cold restore semantics.
- macOS-only behavior stays behind runtime checks; SSH/headless macOS and Windows/Linux/WSL retain their existing behavior.

Closes #7936

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — touched code is clean; the repository-wide command is blocked by the unchanged upstream non-exhaustive switch in `src/renderer/src/components/skills/skill-freshness-group.tsx:101`
- [x] `pnpm typecheck`
- [x] `pnpm test` — sanitized full suite: 3,141 files, 33,353 tests passed; 7 files/56 tests skipped
- [x] `pnpm build`
- [x] Added deterministic lifecycle integration and focused unit coverage — 13 files, 323 tests passed after rebase

## AI Review Report

Ran four review rounds and resolved every valid finding. The review checked macOS lifecycle ownership, full logout, WindowServer recovery, app quit/crash and relaunch, reboot semantics, fast user switching, concurrent clients, reconnect and duplicate-daemon races, stale discovery state, mixed protocol versions, SSH/headless behavior, mobile/web clients, supported providers, and performance.

It explicitly checked cross-platform compatibility for macOS, Linux, Windows, and WSL, including runtime platform guards, paths, shell/process behavior, and Electron IPC boundaries. A valid finding identified a missing server route for closing startup query authority; the route was restored and covered by validation.

## Security Audit

Daemon discovery and attachment are fenced by UID, macOS audit-session ID, and boot UUID. Cleanup verifies the exact PID and process start time before termination, uses bounded fixed-path `launchctl` and `sysctl` probes, and does not log terminal contents. The review also checked IPC authentication, stale metadata, mixed-version fail-closed behavior, multi-user/session isolation, command execution, and path handling. No dependencies or secret handling changed.

## Notes

- Safe deterministic fixtures model login-scope changes; no real developer logout, WindowServer kill, reboot, or unrelated Orca termination was performed.
- Fast user switching preserves each still-valid login session independently.
- WindowServer recovery preserves PTYs while the owning login scope remains valid.
- X handle: N/A

### DIFF
--- a/config/reliability-gates.jsonc
+++ b/config/reliability-gates.jsonc
@@ -5483,6 +5483,96 @@
         "Windows native runtime watches bypass this child path; WSL reservation/release is deterministic-contract tested but not live fault-injected, and SSH registration ownership is not live-relay fault-injected."
       ],
       "demotionRule": "Demote or quarantine if the fault harness flakes without a product or harness bug, if healthy operation exceeds one runtime watcher child, if total physical operation exceeds eight children including retiring generations, if quarantine children outlive their roots or repeat after fusing, if event delivery becomes unbounded, or if metadata/stat work returns to the serve process."
+    },
+    {
+      "id": "terminal-session.macos-login-scope",
+      "title": "macOS daemon PTYs stay inside their owning login session",
+      "maturity": "experimental",
+      "protection": "partial",
+      "owner": "terminal-provider",
+      "layer": "daemon-lifecycle-contract",
+      "surfaces": [
+        "daemon startup",
+        "terminal restore",
+        "macOS logout",
+        "WindowServer recovery",
+        "fast user switching"
+      ],
+      "platforms": ["macos"],
+      "providers": ["daemon"],
+      "coveredPlatforms": ["macos"],
+      "coveredProviders": ["daemon"],
+      "coverageNotes": "Real-socket deterministic tests inject macOS audit-session scopes and cover logout/login rejection, same-session recovery, daemon loss, two-user fast switching, and concurrent clients. A built-daemon fixture covers scoped and unscoped handshakes with a real disposable PTY. Destructive live logout and WindowServer qualification remain uncollected.",
+      "motivatingLinks": ["https://github.com/stablyai/orca/issues/7936"],
+      "invariant": "A client may attach only to a daemon and PTYs in the same valid runtime/login-session scope; a later macOS login or machine boot must not inherit prior-scope PTYs or recovery bytes, while same-session relaunches and independently scoped user accounts retain their sessions.",
+      "oracle": "Create a PTY through an authenticated real-socket client in one injected audit-session scope, disconnect the GUI owner while the daemon survives, and require a later-login scope to fail hello before inventory or createOrAttach. Require same-scope WindowServer/app recovery to return isNew=false with the original PTY PID, daemon loss to return isNew=true, and separate user endpoints to remain independently attachable across switching.",
+      "commands": [
+        "pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/daemon-login-session-history.test.ts src/main/daemon/macos-gui-session-scope.test.ts src/main/daemon/macos-login-session-lifecycle.test.ts src/main/daemon/daemon-health.test.ts src/main/daemon/daemon-init.test.ts"
+      ],
+      "testFiles": [
+        "src/main/daemon/daemon-login-session-history.test.ts",
+        "src/main/daemon/macos-gui-session-scope.test.ts",
+        "src/main/daemon/macos-login-session-lifecycle.test.ts",
+        "src/main/daemon/daemon-health.test.ts",
+        "src/main/daemon/daemon-init.test.ts"
+      ],
+      "assertionRefs": [
+        {
+          "file": "src/main/daemon/macos-login-session-lifecycle.test.ts",
+          "assertions": [
+            "a later login is rejected before warm attachment",
+            "WindowServer restart, app crash, and normal reopen preserve the PTY only inside one scope",
+            "reboot creates a fresh PTY",
+            "fast-switched user endpoints remain isolated and independently usable",
+            "concurrent same-scope clients share one PTY"
+          ]
+        },
+        {
+          "file": "src/main/daemon/daemon-init.test.ts",
+          "assertions": [
+            "scope rejection enters exact-PID bounded daemon replacement",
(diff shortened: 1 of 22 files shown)


brennanb2025:
Reviewed this after dragonfire1119's live test of `v1.4.149-rc.1.issue7936` failed (Discord): a terminal created **post-update**, then logout/login, still produced `login:` / Login incorrect, and only a manual daemon restart recovered. I verified the RC tag contains this PR's commit, so the miss is behavioral. Findings from a code walk of the shipped tag:

**1. `checkDaemonHealth` / `getMacDaemonSystemResolverHealth` never verify the response scope.** `DaemonClient.ensureConnected()` rejects when `response.runtimeScope !== this.runtimeScope`, but the two raw-socket probes only check `ok`. A **pre-scope protocol-25 daemon** (v1.4.150-rc.0 is protocol 25 *without* this PR — this exact collision is live in the field) answers hello fine → `'healthy'`, while every scoped RPC client fails against it → `getAliveDaemonSessionCount` returns `null` → both "can't verify live sessions" branches **preserve** it. Result: a stale daemon the app can't even talk to is preserved forever, and the adapter limps to local fallback. The doc's claim that "a pre-scope daemon is adopted only after its verified process audit session matches" is implemented only for *legacy-protocol* daemons (`legacyDaemonMatchesRuntimeScope`), not for same-protocol pre-scope daemons on the current socket.

**2. `resolveMacosGuiSessionScope` throws → no daemon at all.** In `startDaemonPtyProvider` the throw propagates to `onDaemonError` → silent local-PTY fallback, and the stale daemon is left untouched. The 1s `launchctl`/`sysctl` timeouts are most likely to fire exactly during the post-login storm — i.e., the moment this fix must work. A resolution failure should degrade to unscoped-init (old behavior), not to daemon-less.

**3. Unscoped `orca serve` clients vs. scoped daemons = mutual replacement loop.** Serve is unscoped by design; its hello to a scoped daemon hits `hello.runtimeScope !== this.runtimeScope` → `'rejected'` → serve's launcher treats it as never-adoptable and replaces it; the GUI client then rejects serve's unscoped daemon and replaces it back. Anyone running desktop + serve on one macOS box gets a daemon kill war, each cycle destroying all PTYs. Suggest: reject only when the client's scope is *present and different*; tolerate scope-absent clients (they're uid+token-fenced already).

**4. `createLegacyDaemonAdapters` passes `historyPath: getHistoryDir()`** (unscoped) instead of `getHistoryDir(runtimeScope)` — inconsistent with the new scoped-slot model.

**Also, the symptom itself lives below this PR's layer**: the #9301 preflight caches its birth-time 'accepted' verdict per daemon process, so a surviving dead-session daemon wraps every spawn in `login(1)` forever — that's the direct generator of the `login:` prompt zombies, and no startup-time fence can heal a daemon the app never restarts against. I've opened #9826, which makes GUI daemons detect login-session death from the inside (cache-bypassing PAM probe + resolver corroboration) and self-retire crash-style so the next start cold-restores. It complements this PR rather than replacing it.

To confirm which link broke on the reporter's machine, their `~/Library/Application Support/orca/logs/daemon.log` from a broken state would settle it — this PR's `client-hello-rejected`/`client-hello-accepted` records are exactly the discriminator.

brennanb2025:
Data point for the pre-scope adoption guard from a live dead-session experiment (details on #9826): after a real audit-session teardown (SSH-minted session, orphan survived, reparented to launchd), `launchctl print pid/<orphan>` **continues to report the dead session's original asid** with a parseable security context. So `readMacosProcessSessionIdentity` on a post-logout daemon should reliably see the stale asid and fence it — the guard's input behaves as designed. The gap remains the paths that never consult that guard (current-protocol pre-scope daemons, finding 1 in my earlier comment).

brennanb2025:
Closing as superseded by #9826 after discussion with Brennan.

Rationale: the live RC test (v1.4.149-rc.1.issue7936) was falsified in the field by the reporter, and the code walk above found four integration holes in the startup-time fence approach — most critically the preserve-deadlock against pre-scope protocol-25 daemons (which 1.4.150-rc.0 ships) and the unscoped-serve vs scoped-daemon replacement war. More fundamentally, a startup-time fence can only help when the app restarts and every adopt-vs-replace decision goes right; the surviving daemon itself stays broken.

#9826 inverts the approach: the daemon detects its login session died from the inside (PAM probe + resolver corroboration, hang-shaped fallback trigger) and retires itself, so the next start cold-starts a fresh daemon regardless of any client-side decision — validated with live lifecycle QA and real dead-session experiments (including the launchctl stale-asid data point posted above, which confirms this PR's process-identity guard input was sound).

What remains valuable from this PR's scope if we revisit later: the warm-attach fence for fast-user-switching correctness and multi-session isolation, and the login-scoped history slot. If that gets picked back up, the four findings above are the pre-merge checklist. #7936 stays open until #9826 is field-confirmed.

PR #9714, opened 2026-07-21
fix(ui): align toggle switch handle symmetrically in on state

sei0:
## Summary

The toggle switch's handle was not centred symmetrically in the **on** state: it sat 4px from the right edge while the off state sat 2px from the left.

The arithmetic, from the real classes: the track is `h-5 w-9` with `box-sizing: border-box` and a 1px border, so the content box is **34px** wide. The handle is `size-3.5` = **14px**. `translate-x-0.5` insets the off state by **2px**. A symmetric on state therefore needs `34 − 14 − 2 = 18px`, i.e. `translate-x-4.5`. The code used `translate-x-4` (16px), leaving a 4px right gap against a 2px left gap.

## ELI5

A toggle is a little pill with a dot that slides between two ends. When the dot was on the left it sat 2 pixels from the edge; when you flipped it to the right it stopped 4 pixels short. Nobody would consciously notice, but it made the switch look subtly lopsided in the "on" position. This moves the dot 2 pixels further so both ends match exactly.

## Proof of fix

The switches in this codebase are **hand-rolled**, not instances of a shared `ui/switch.tsx` primitive — there is no such file. The off/on offsets are written literally at each of 26 call sites, which is exactly why they drifted. So the fix has to touch each site, and the durable protection is a test that scans them all.

Added `src/renderer/src/components/toggle-switch-handle-alignment.test.ts`. It walks the source for two-state handle offsets and asserts every pair is the symmetric `translate-x-0.5` / `translate-x-4.5`.

**Reverting a single call site is enough to fail it** — this is the drift-detection proof, and it means no future switch can silently regress:

```
$ git checkout origin/main -- src/renderer/src/components/settings/AccountsPane.tsx
    src/renderer/src/components/toggle-switch-handle-alignment.test.ts

 × offsets the on state so both edges inset by 2px
 × pairs every two-state handle with the symmetric on/off offsets
AssertionError: expected [ Array(1) ] to deeply equal []
AssertionError: expected [ …(2) ] to deeply equal [ 'translate-x-0.5 translate-x-4.5' ]

```

**With all 26 sites corrected:**

```
$ git checkout HEAD -- src/renderer/src/components/settings/AccountsPane.tsx
    src/renderer/src/components/toggle-switch-handle-alignment.test.ts

 Test Files  1 passed (1)
      Tests  3 passed (3)
```

The geometry is stated in the description so a reviewer can check the arithmetic without running anything: `34 − 14 − 2 = 18px = translate-x-4.5`.

## Proof of no regressions

```
$ npx vitest run --config config/vitest.config.ts src/renderer/src/components/settings/ \
    src/renderer/src/components/toggle-switch-handle-alignment.test.ts
 Test Files  101 passed (101)
      Tests  649 passed (649)

$ node config/scripts/check-styled-scrollbars.mjs      # clean
```

- Every toggle switch's handle moves **2px further right in the on state**. That is the entire change; it is the fix.
- The off state, track size, colours, transition timing, focus ring, and disabled styling are all untouched.

All 26 changed files receive the identical one-token substitution (`translate-x-4` → `translate-x-4.5`) on the handle span. No component's structure, props, or behavior changes.

## Compatibility

- **Platforms:** macOS / Linux / Windows. A Tailwind spacing token compiled to a CSS transform; no platform branch, no key modifiers, no paths, no Electron API. Chromium is the only engine Electron renders in.
- **SSH / remote:** N/A — presentation-only, identical for local, SSH, and relay-backed workspaces.
- **Performance:** No hot-path change. `translate-x-4.5` and `translate-x-4` compile to the same `transform: translateX(...)` property with a different constant; identical layout/paint cost and no extra style recalculation. No new classes, elements, state, or effects.

## Screenshots

The change is a 2px shift, which is real but hard to convey honestly in a scaled screenshot — at normal zoom the before/after images look identical, and a zoomed crop invites the reader to trust a rendering I would have had to set up by hand. I chose to make the guarantee **checkable instead of photographable**: the arithmetic is written out above, and `toggle-switch-handle-alignment.test.ts` enforces it across all 26 sites in CI, failing if any one drifts.

I did not capture before/after images — stating that plainly rather than attaching something that would not actually demonstrate the fix.

## Testing

- [ ] `pnpm lint` — not run in full. Ran `node config/scripts/check-styled-scrollbars.mjs` (part of the lint suite, relevant to a class-only change): clean.
- [ ] `pnpm test` — not run in full (30+ min). Ran the settings directory plus the new suite: 101 files, 649 tests passed.
- [ ] `pnpm build` — not run.
- [x] Added or updated high-quality tests that would catch regressions — a source-scanning boundary test that fails if *any* of the 26 sites drifts from the symmetric pair.

## AI Review Report

Reviewed with Claude Code (Opus 5) and independently re-verified. Risks checked:

- **Should this be one change to a shared primitive instead of 26 edits?** This was the first thing checked, since a 26-file diff for an alignment bug normally means the fix is in the wrong layer. Confirmed there is **no** `src/renderer/src/components/ui/switch.tsx`; each switch is hand-rolled with literal offsets. Per-site edits are therefore unavoidable, and the scanning test is what prevents the next drift — it is the part of this PR that carries long-term value.
- **Is the arithmetic right?** Derived from the real classes rather than eyeballed: `w-9` (36px) − 2×1px border = 34px content box; `size-3.5` = 14px handle; `translate-x-0.5` = 2px inset ⇒ on state = 18px = `translate-x-4.5`. Off and on insets are now equal.
- **Is `translate-x-4.5` a real token?** Verified — Tailwind 4.2.4 supports `.5` spacing natively, and `translate-x-4.5` already appears on `main` in `SidebarFilter.tsx` and `SidebarWorkspaceFilterSection.tsx`. No new token, value, or config entry is introduced, per `docs/STYLEGUIDE.md`.
- **Variant/theme coverage:** all 26 sites use the same `h-5 w-9` / `size-3.5` geometry, so there is no second size whose maths would differ. The change is colour-independent, so light and dark themes are unaffected.
- **Consumer layout:** the handle is `pointer-events-none` and absolutely positioned within its track; shifting it 2px cannot affect the surrounding layout in any container.
- **Cross-platform (macOS / Linux / Windows):** no platform-dependent code, shortcut labels, paths, or shell behavior in this diff.

## Security Audit

- **Input handling:** none. Presentation-only class change.
- **Path handling:** none.
- **Auth / secrets:** none touched. Worth noting one of the changed files is `AccountsPane.tsx`, but only the handle-offset token is modified — no auth logic, state, or rendering condition is affected.
- **IPC:** untouched.

## Notes

- **Follow-up worth considering, deliberately out of scope:** the real root cause is that 26 components each hand-roll a switch instead of sharing a primitive. Extracting a `ui/switch.tsx` would make this class of bug structurally impossible, but it is a much larger refactor with real regression surface and does not belong in an alignment fix. The scanning test is the interim guard.

Original patch by @sei0.

### DIFF
--- a/src/renderer/src/components/NewWorkspaceComposerCard.tsx
+++ b/src/renderer/src/components/NewWorkspaceComposerCard.tsx
@@ -1442,7 +1442,7 @@ export default function NewWorkspaceComposerCard({
               <span
                 className={cn(
                   'pointer-events-none block size-3.5 rounded-full bg-background shadow-sm transition-transform',
-                  createMultiple ? 'translate-x-4' : 'translate-x-0.5'
+                  createMultiple ? 'translate-x-4.5' : 'translate-x-0.5'
                 )}
               />
             </span>
--- a/src/renderer/src/components/NewWorkspaceComposerCard.tsx
+++ b/src/renderer/src/components/NewWorkspaceComposerCard.tsx
@@ -604,6 +604,7 @@ function useComposerFileDragOver(): {
   }
 }
 
+/** Composer card for creating new workspaces. */
 export default function NewWorkspaceComposerCard({
   contextualTourSource,
   containerClassName,
--- a/src/renderer/src/components/NewWorkspaceComposerCard.tsx
+++ b/src/renderer/src/components/NewWorkspaceComposerCard.tsx
@@ -604,7 +604,6 @@ function useComposerFileDragOver(): {
   }
 }
 
-/** Composer card for creating new workspaces. */
 export default function NewWorkspaceComposerCard({
   contextualTourSource,
   containerClassName,
--- a/src/renderer/src/components/contextual-tours/ContextualTourControl.tsx
+++ b/src/renderer/src/components/contextual-tours/ContextualTourControl.tsx
@@ -66,7 +66,7 @@ function AutoRenameBranchFromWorkControl(): JSX.Element {
           <span
             className={cn(
               'pointer-events-none block size-3.5 rounded-full bg-background shadow-sm transition-transform',
-              enabled ? 'translate-x-4' : 'translate-x-0.5'
+              enabled ? 'translate-x-4.5' : 'translate-x-0.5'
             )}
           />
         </button>
--- a/src/renderer/src/components/contextual-tours/ContextualTourControl.tsx
+++ b/src/renderer/src/components/contextual-tours/ContextualTourControl.tsx
@@ -5,6 +5,7 @@ import type { ContextualTourStepControl } from '../../../../shared/contextual-to
 import { CONTEXTUAL_TOUR_ENABLE_AUTO_WORKSPACE_NAME_EVENT } from './contextual-tour-composer-events'
 import { translate } from '@/i18n/i18n'
 
+/** Contextual tour overlay control with toggle. */
 export function ContextualTourControl({
   control
 }: {
--- a/src/renderer/src/components/contextual-tours/ContextualTourControl.tsx
+++ b/src/renderer/src/components/contextual-tours/ContextualTourControl.tsx
@@ -5,7 +5,6 @@ import type { ContextualTourStepControl } from '../../../../shared/contextual-to
 import { CONTEXTUAL_TOUR_ENABLE_AUTO_WORKSPACE_NAME_EVENT } from './contextual-tour-composer-events'
 import { translate } from '@/i18n/i18n'
 
-/** Contextual tour overlay control with toggle. */
 export function ContextualTourControl({
   control
 }: {
--- a/src/renderer/src/components/feature-wall/AiCommitPrSettingsSwitch.tsx
+++ b/src/renderer/src/components/feature-wall/AiCommitPrSettingsSwitch.tsx
@@ -27,7 +27,7 @@ export function AiCommitPrSettingsSwitch({
       <span
         className={cn(
           'pointer-events-none block size-3.5 rounded-full bg-background shadow-sm transition-transform',
-          checked ? 'translate-x-4' : 'translate-x-0.5'
+          checked ? 'translate-x-4.5' : 'translate-x-0.5'
         )}
       />
     </button>
--- a/src/renderer/src/components/feature-wall/AiCommitPrSettingsSwitch.tsx
+++ b/src/renderer/src/components/feature-wall/AiCommitPrSettingsSwitch.tsx
@@ -7,6 +7,7 @@ type AiCommitPrSettingsSwitchProps = {
   onToggle: () => void
 }
 
+/** Toggle switch for AI commit/PR settings. */
 export function AiCommitPrSettingsSwitch({
   checked,
   label,
--- a/src/renderer/src/components/feature-wall/AiCommitPrSettingsSwitch.tsx
+++ b/src/renderer/src/components/feature-wall/AiCommitPrSettingsSwitch.tsx
@@ -7,7 +7,6 @@ type AiCommitPrSettingsSwitchProps = {
   onToggle: () => void
 }
 
-/** Toggle switch for AI commit/PR settings. */
 export function AiCommitPrSettingsSwitch({
   checked,
   label,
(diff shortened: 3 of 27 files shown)


sei0:
@coderabbitai review

brennanb2025:
Thanks for this — the diagnosis and the arithmetic both hold up. Heads up that I've pushed one commit to your branch, so the extra commit isn't a surprise.

I measured it in the running app rather than trusting the maths: the track renders 36px wide with a 1px border (a 34px content box) and the handle is 14px, so the off state sits 3px from the outer edge while the on state was landing 5px from it. With `translate-x-4.5` both ends measure 3px — checked across every rendered switch in settings and the workspace composer, in light and dark, and the 0.15s slide is unaffected.

What I changed — the guard test only, none of your component fixes:

- It was finding files by shelling out to `git grep` for `role="switch"`, which silently skipped two of the sites this PR fixes: `stats/ClaudeUsageLoadingState.tsx` (a loading skeleton, so it's a `div` with no role) and `settings/HiddenExperimentalGroup.tsx`. I confirmed the gap by reverting the loading-state handle back to `translate-x-4` — the old test stayed green. It now walks the renderer with the same `readdirSync` + TypeScript AST approach `i18n/no-top-level-translate.test.ts` already uses, and finds each knob inside its own track instead of scanning source text forward, so a same-file icon or a neighbouring element can't be mistaken for it.
- It also now fails when the premise behind `34 − 14 − 2 = 18` stops holding, not just when a token changes: a track that loses its 1px border or gains padding, a conditional whose two branches no longer pair off/on, and a track whose knob moves into a child component — that last case used to make the scan quietly stop seeing it.

One for later: you're right in your notes that the real problem is these being hand-rolled at every call site. Extracting a shared switch primitive would make this class of bug impossible, and it genuinely doesn't belong in an alignment fix — worth doing on its own.

brennanb2025:
Thanks for the PR!

brennanb2025:
### Attachments

| grok-pr9714-settings-general-full.png | grok-pr9714-settings-appearance-full.png |
|---|---|
|  |  |

| grok-pr9714-closeup-live-on-nest-workspaces.png | grok-pr9714-closeup-live-off-confirm-before-closing-pinned-tabs.png |
|---|---|
|  |  |

| grok-pr9714-closeup-row-on-nest-workspaces.png | grok-pr9714-closeup-row-off-confirm-pinned.png |
|---|---|
|  |  |

| grok-pr9714-closeup-row-on-auto-save.png |
|---|
|  |

brennanb2025:
## Electron UI validation (post-merge) — toggle handle alignment

Isolated `pnpm dev` for this worktree: CDP `9400`, userData `/tmp/orca-pr9714-userdata-*`, renderer `http://localhost:5177/`.

**Identity (`window.api.app.getIdentity()`):**
- `devRepoRoot`: `/Users/brennanbenson/orca/workspaces/orca/pr-9714-review`
- `devBranch`: `fix/toggle-switch-handle-alignment`
- `devWorktreeName`: `pr-9714-review`
- `devLabel`: `pr9714-validation-grok`

**Settings → General** (live switches exercised via CDP):
| State | Outer left gap | Outer right gap | Computed `translate` | Class |
|---|---:|---:|---:|---|
| ON  | 19px | 3px | 18px | `translate-x-4.5` |
| OFF | 3px  | 19px | 2px  | `translate-x-0.5` |

Track: 36×20px, `border-box`, 1px borders (content box 34px); thumb: 14×14px. Gaps are exactly mirrored between ON/OFF. After click: OFF→ON lands at L19/R3; ON→OFF lands at L3/R19.

**Animation:** `transition-property: transform, translate, scale, rotate` at `0.15s` — unchanged; slide still animates via Tailwind v4 `translate`.

**Logs:** no switch/UI errors. Only pre-existing noise on this disposable profile: `legacy worker provider-ready recovery failed: terminal_liveness_unavailable`, `MaxListenersExceededWarning` on BrowserWindow.

**Verdict: LAND** (already merged). Screenshots attached in the prior comment (`grok-pr9714-*`).

PR #9722, opened 2026-07-21
fix(i18n): correct zh.json 'Match Case' mistranslation

fsdwen:
The Chinese translation for "Match Case" was incorrectly translated as "火柴盒" (matchbox) — the verb "match" was misinterpreted as the noun "match" (a stick for lighting fire).

Corrected to "区分大小写", which is the standard Chinese UI term for case-sensitive matching.


fsdwen:
@AmethystLiang

PR #9729, opened 2026-07-21
fix(agents): stop forking a duplicate Pi tab for a live background session

shaharmor:
## Summary

Fixes a bug where finishing a Pi turn in a background tab and then reactivating the worktree could pop open a duplicate Pi tab resuming a session that was still alive in the original tab.

## ELI5

If an AI assistant finished talking in a tab you weren't looking at, coming back to that project could open a second copy of it. Now it stays as one tab.

## Root cause & fix

Pi's live-recovery records keep a session resumable after its turn completes, but worktree activation's `recordPaneIsOwnedByPreservedPane()` only treated a pane as owning its session when that pane was the one reconnecting on activation (the active tab). A done-but-alive Pi pane sitting in a background tab was misread as un-owned, so resume forked a duplicate tab. The fix adds `stablePaneHasLivePty()`: a pane owns its running session whenever it currently has a live PTY, regardless of focus, and returns early before the activation gate. Liveness is read from the runtime `ptyIdsByTabId` map rather than the stale `ptyIdsByLeafId` layout snapshot.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase onto `origin/main` was clean.

- Test `pi-live-session-no-duplicate-tab.test.ts` passes on branch: **1 file, 1 test passed**.
- Reverting only the fix (`sleeping-agent-pane-ownership.ts`) to `origin/main` while keeping the test makes it **FAIL** — resume forks a duplicate tab (`launched === 1`).
- Lint clean: `oxlint` on the changed file reports no findings.

```
STEP 2a (on branch): Test Files 1 passed (1); Tests 1 passed (1)

STEP 2b (fix reverted): FAIL
AssertionError: expected 1 to be +0 // Object.is equality
  at pi-live-session-no-duplicate-tab.test.ts:78  expect(launched).toBe(0)
```

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression by construction: only the ownership branch for panes with a live PTY changes. Genuinely hibernated panes (no live PTY) keep the existing cold-restore-and-fork behavior, so non-affected paths are unchanged. The reproduction test proves both the fix and the preserved behavior.

_Credit to @shaharmor for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/renderer/src/lib/pi-live-session-no-duplicate-tab.test.ts
+++ b/src/renderer/src/lib/pi-live-session-no-duplicate-tab.test.ts
@@ -1,0 +1,82 @@
+import { afterEach, describe, expect, it } from 'vitest'
+import type { SleepingAgentSessionRecord } from '../../../shared/agent-session-resume'
+import { useAppStore } from '@/store'
+import { resumeSleepingAgentSessionsForWorktree } from './resume-sleeping-agent-session'
+
+const initialAppStoreState = useAppStore.getState()
+const LEAF_ID = '11111111-1111-1111-8111-111111111111'
+const PANE_KEY = `pi-tab:${LEAF_ID}`
+
+afterEach(() => {
+  useAppStore.setState(initialAppStoreState, true)
+})
+
+describe('Pi live session does not spawn a duplicate resume tab', () => {
+  it('keeps a done-but-alive Pi pane instead of forking a new tab', () => {
+    const providerSession = {
+      key: 'session_id' as const,
+      id: 'pi-1',
+      transcriptPath: '/tmp/pi-session-1.jsonl'
+    }
+    const record: SleepingAgentSessionRecord = {
+      paneKey: PANE_KEY,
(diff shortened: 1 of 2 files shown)


JakubGrupa36:
@shaharmor +1, I have that problem very frequent - switching between worktrees with pi creates a lot of forks of active sessions which is totally frustrating.

shaharmor:
@nwparker can we please get this reviewed? It's really making the Orca experience frustrating.

nwparker:
@shaharmor yeah doing it now

And sorry my GH volume has been too high so I miss mentions 😢 
Feel free to always msg me on X or Discord and I'll take care of stuff nearly instantly!

nwparker:
Thanks so much for this, @shaharmor — stop forking a duplicate Pi tab for a live background session. Reviewed, rebased on latest `main`, and merged. Really appreciate the contribution! 🙏

PR #9733, opened 2026-07-21
Fix folder workspace parent race during background creation

noO0oOo0ob:
## Summary

- capture the active folder workspace when quick worktree creation is submitted
- carry that captured parent through pending/retry state and background creation
- preserve the existing active-workspace fallback for legacy and synchronous callers
- add regression coverage for changed focus and explicitly unparented requests

## Root cause

Quick creates close the composer and continue in the background. The store previously resolved `parentWorkspace` only when each background job reached `createWorktree`. If an earlier job completed first, activation changed `activeWorkspaceKey` from the folder workspace to the new worktree. Later jobs then omitted the folder parent and appeared as unattached worktrees.

## Impact

Users can queue worktrees for several repositories in one folder workspace without waiting for each previous task to complete. Every queued worktree remains attached to the folder workspace that was active when Create was clicked.

## Validation

- `vitest` targeted suites: 261 tests passed
- `pnpm run typecheck:web`
- `oxlint` on all changed files
- `oxfmt` on all changed files

## ELI5

Quick worktree create under a folder workspace could attach to the wrong parent if you switched focus while creation ran in the background. The parent is captured at submit so the new tree stays under the folder you meant.

### DIFF
--- a/src/renderer/src/hooks/useComposerState-host-context-boundaries.test.ts
+++ b/src/renderer/src/hooks/useComposerState-host-context-boundaries.test.ts
@@ -383,6 +383,18 @@ describe('useComposerState host-context boundaries', () => {
     expect(quickCreate).toBeGreaterThan(quickPolicySave)
   })
 
+  it('captures the folder parent before quick-create preflight can change focus', () => {
+    const quickSubmit = sourceBetween(HOOK_SOURCE, 'const submitQuick = useCallback', 'return {')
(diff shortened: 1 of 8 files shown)


Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17576. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @noO0oOo0ob. No action is requested from you. Thank you.

PR #9734, opened 2026-07-21
fix(wsl): route global CLI fallbacks to user-pinned terminalWindowsWslDistro

TingdeLiu:
# fix(wsl): route global CLI fallbacks to user-pinned `terminalWindowsWslDistro`

## Summary
On Windows with WSL-only installs of `gh`/`glab`, global CLI commands now honor the user's pinned WSL distro instead of failing against an arbitrary default distro.

## ELI5
If your GitHub/GitLab tools live inside one Linux setup on Windows, the app used to guess the wrong one and fail. Now it uses the Linux distro you actually picked in settings.

## Root cause & fix
Global CLI calls carry no repository `cwd`, so when the host executable is missing the fallback resolved to the first distro from `wsl --list` via `getDefaultWslDistro()` — wrong when the user's tools/credentials live in a different distro. The fix adds a module-level `defaultWslDistroOverride` with an exported `setDefaultWslDistroOverride` setter in `src/main/git/runner.ts`; `resolveDefaultWslCli` now prefers `override ?? getDefaultWslDistro()`. `src/main/index.ts` applies the override from `terminalWindowsWslDistro` both at startup and on settings change.

## Evidence
Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase clean (base `8a23618310`, head `fbdce08bc2`).

- Test: `src/main/git/runner-wsl-gh-fallback.test.ts` (new override-vs-default-distro case).
- On branch: **1 file passed, 22 tests passed**.
- Revert fix files to `origin/main` (keeping the test): **22 tests failed** — proving the fix is load-bearing:

```
npx vitest run --config config/vitest.config.ts src/main/git/runner-wsl-gh-fallback.test.ts

On branch:        Test Files 1 passed | Tests 22 passed
Fix reverted:     Tests 22 failed
  → TypeError: setDefaultWslDistroOverride is not a function
```

- Lint clean: `npx oxlint src/main/git/runner.ts src/main/index.ts` → exit 0.

## Trade-offs
None — pure fix.

## Regression risk
Effectively zero. The override defaults to `null`, so `resolveDefaultWslCli` falls back to `getDefaultWslDistro()` exactly as before when no distro is pinned; only the WSL-fallback branch changes, and non-Windows paths are gated by `process.platform`. The `index.ts` wiring is a straightforward read of an existing `terminalWindowsWslDistro` setting and cannot be unit-tested directly (Electron `whenReady`), but was inspected.

_Credit to @TingdeLiu for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/openspec/changes/archive/2026-07-21-wsl-default-distro-gh-fallback/.openspec.yaml
+++ b/openspec/changes/archive/2026-07-21-wsl-default-distro-gh-fallback/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-21
--- a/openspec/changes/archive/2026-07-21-wsl-default-distro-gh-fallback/README.md
+++ b/openspec/changes/archive/2026-07-21-wsl-default-distro-gh-fallback/README.md
@@ -1,0 +1,3 @@
+# wsl-default-distro-gh-fallback
+
+Fallback to configured default WSL distro when host gh CLI is missing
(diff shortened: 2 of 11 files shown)


nwparker:
Thanks so much for this, @TingdeLiu — route global CLI fallbacks to user-pinned terminalWindowsWslDistro. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

PR #9739, opened 2026-07-21
Fix remote session convergence and duplicate agent restores

nikuscs:
## Summary

- Make tab subscriptions atomic so changes cannot disappear between the initial list and live updates.
- Reject retired publishers and stale snapshots so reconnects cannot restore older host state.
- Keep remote terminal mirrors available while offline and avoid treating failed inventory checks as deletion.
- Stop mirrored editor and browser tabs from being republished as local orphan state.
- Acknowledge editor and browser closes so remote tab removal is durable.
- Prevent transient empty PTY inventories and concurrent tab activation from restoring the same agent twice.
- Give the active remote desktop control of terminal geometry without letting passive viewers steal it.
- Enforce one live PTY per resumable provider session across desktop and runtime launch paths. Duplicate resumes reveal the existing terminal instead of launching or rebinding it.

## Validation

- Full test suite: 3,157 files and 33,526 tests passed
- Node, web, and mobile typechecks passed
- Focused lint, formatting, diff checks, and max-lines ratchet passed

Full repository lint still reaches the existing exhaustiveness error in `src/renderer/src/components/skills/skill-freshness-group.tsx:101`.


nikuscs:
Closed because even after all the efforts this didnt fixed it permanently , hopefully this helps someone in the future :p

Possible related to [9952](https://github.com/stablyai/orca/issues/9952)

PR #9748, opened 2026-07-21
feat(notifications): add silent notification sound option

### LINKED ISSUE #9146: [Feature]: silent notifications
### Problem or use case

running a lot of different worktrees at the same time can get messy with notifications - the ability to turn off the sound would be helpful

### Proposed solution

a new "None" option for notification sound

_No response_

haazz:
## Summary

- Add a `None` notification sound option in Settings and onboarding.
- Keep desktop notification banners enabled while suppressing both native system sounds and Orca-rendered sounds.
- Skip sound previews and hide the volume control when `None` is selected.
- Persist the silent selection across app restarts.
- Localize the option for English, Spanish, Japanese, Korean, and Simplified Chinese.

Closes #9146

## Screenshots

Notification Sound set to None

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [x] `pnpm build`

Focused verification:

- 4 notification and persistence suites with 460 tests passed.
- Persisted `none` selections survive settings normalization.
- Native notifications use `silent: true` when `None` is selected.
- Test notifications do not request renderer audio.
- Settings hides the volume control for silent notifications.
- Localization catalog parity and coverage checks pass for all supported locales.
- Commit hooks pass lint, React Doctor, and formatting checks for every changed file.
- The full production build completes successfully, including desktop, web, native, and macOS helper targets.

Full-repository limitations:

- `pnpm lint` reaches the type-aware switch gate, then fails on an existing unchanged exhaustiveness error in `src/renderer/src/components/skills/skill-freshness-group.tsx:101`.
- A full `pnpm test` run passed 33,506 tests and reported two failures in unchanged suites. The transcript watcher timing test passed when rerun independently; the local `/bin/ksh` parser test still exits without a status in this environment.

## AI Review Report

The AI review found no correctness issues in the scoped change.

The review explicitly checked:

- macOS, Windows, and Linux native notification behavior.
- Electron `silent` handling and the macOS default-sound branch.
- Renderer playback suppression for system, silent, built-in, and custom sound modes.
- Settings and onboarding consistency.
- Persistence and backward compatibility.
- Local and SSH-triggered notification paths.
- Agent and integration neutrality.
- Localization catalog parity and natural labels for all supported UI languages.
- Performance and UI consistency with the existing notification controls.

No shortcut, path, shell, Git-provider, or remote-runtime behavior changed. The runtime change adds only constant-time sound-ID comparisons.

## Security Audit

No dependency, IPC channel, command execution, filesystem access, path handling, authentication, secret handling, or network behavior was added.

The existing persisted sound identifier gains one validated value. Existing notification and sound IPC boundaries remain unchanged. Selecting `None` only suppresses native and renderer audio playback.

## Notes

- This setting affects desktop notification audio only; notification banners remain enabled.
- Mobile notification channel behavior is unchanged.
- SSH and remote agent events use the same local desktop notification preference.
- X handle: @haazwon

## ELI5

Notifications can use a None sound option: banners still appear, but neither system nor Orca plays a sound. Volume controls hide when silent is selected, and the choice is remembered.


weedo-dev:
Hey @haazz, thanks for this contribution

I would suggest adding a toggle to turn on/off rather than an empty sound

wdyt?

haazz:
> Hey @haazz, thanks for this contribution
> 
> I would suggest adding a toggle to turn on/off rather than an empty sound
> 
> wdyt?

Thanks for the suggestion!

I think it would be better to keep the `None` option, since a separate toggle could be confused with the existing “Enable notifications” setting.

@brennanb2025, wdyt?

Issue #1693, opened 2026-05-11
SSH relay: pre-bundle node-pty per platform (VS Code-style) to eliminate npm install on remote


Jinwoo-H:
## Context

Today the SSH relay deploy runs `npm install node-pty` on the remote host at first connect. This requires:

- npm + node on the remote (we already deploy node)
- Either a matching node-pty prebuild on the registry OR build tools (python3, make, g++) on the remote
- Network access to the npm registry from the remote

This has been the source of multiple bugs:

- #1660 reconnect loop (root cause was unrelated, but the fix made the install dir name contain `+` which broke npm-init — see commit 963f56d7 on `Jinwoo-H/ssh-typing-lag`)
- npm 11 / Node 26 strict semver name validation rejects `+` in package names derived from dir basename
- A remote without compilers can't install node-pty at all → relay degrades to fs/git/preflight without PTY (graceful but confusing to users)
- Transient registry failures during install can leave a partially-installed dir that requires manual cleanup

## How VS Code does it

VS Code's remote server (`vscode-reh`) is a fully pre-built tarball, generated on their CI per platform (`vscode-server-linux-x64.tar.gz`, `...-linux-arm64`, `...-darwin-x64`, `...-darwin-arm64`, plus Alpine variants). The tarball already contains `node_modules/node-pty/` with the right prebuild for that platform.

Their remote-install flow:

1. `prereqs.rs::verify()` detects glibc/glibcxx/musl on the remote → picks the matching tarball (or returns `PrerequisitesFailed` if requirements aren't met)
2. Download the tarball
3. Extract via `decompress_tarball` (Rust gunzip + tar reader, no `tar` binary required on remote)
4. Done — no `npm install`, no compiler needed, no `package.json` ceremony

Source files in the VS Code repo: `cli/src/util/prereqs.rs`, `cli/src/util/tar.rs`, `resources/server/bin/code-server-linux.sh`, `resources/server/bin/helpers/check-requirements-linux.sh`.

## Proposed approach for Orca

Update `config/scripts/build-relay.mjs` to additionally fetch node-pty's prebuilds (via `npm install node-pty` in a temp dir, then copy the right `prebuilds/<platform>/` subdir) and include them in `resources/relay/<platform>/node_modules/node-pty/`. Then in `src/main/ssh/ssh-relay-deploy.ts`:

- `installNativeDeps` becomes a no-op (or, more precisely, it just `chmod +x` the spawn-helper since SFTP doesn't preserve execute bits)
[...]

PR #9752, opened 2026-07-21
fix(terminal): reap Windows agent trees with Job Objects

### LINKED ISSUE #9704: Windows: PTY descendants survive session kill — orphaned agent process trees accumulate (~1.1GB)
> **Edit:** I first filed this claiming a daemon crash wipes every worktree's terminals at once.  <br>I couldn't substantiate that — I had one crash in my log and reasoned backwards from it. What I  <br>can actually observe is that **PTY descendants survive session kills and pile up**. I've  <br>rewritten the report around that and moved the crash to a secondary note. Sorry for the noise.

**TL;DR** — On Windows, killing a terminal session leaves the PTY's grandchildren running  <br>indefinitely. They accumulate, hold worktree paths open, and keep burning CPU. What I'd most like  <br>is to understand *why* sessions get killed at all, since losing one loses the agent's in-flight work.

## What I can confirm

Orca 1.4.147, Windows 11, daemon protocol v24.

Killing a session doesn't reap the PTY's descendants — they keep running with a dead parent:

```
node codex.js (~25MB)          <- launcher shim, parent already dead
  └─ codex.exe (86–324MB)
       ├─ codex-code-mode-host.exe
       └─ pwsh.exe
```

* **6 trees / 18 processes / \~1.1GB resident**, ~16h after their sessions were killed
* Not inert — the largest had 304s CPU, and ticked 302.1s → 304.4s while I was inspecting it
* Accumulates: 4 trees one day, 6 the next, as more parents died
* Holds handles on worktree paths, so it can block worktree deletion

**Update, a day later:** the same thing with a *different* agent — **5 trees / 20 processes /< />2.0GB**, this time Claude Code (`claude.exe → python.exe ×2 + conhost.exe`), ages ranging from 0.5h  <br>to 24.3h. Two of those trees were sessions killed < />15 minutes earlier. So this isn't Codex-specific;  <br>it looks like the PTY teardown path itself.

Windows doesn't clean up descendants when a parent dies, so unless the PTY's children are placed in  <br>a Job Object with `KILL_ON_JOB_CLOSE`, killing the direct child leaves the grandchildren running.

## Likely contributor: the graceful kill has no grace period

Every kill in my log is an `immediate:false` followed by `immediate:true` **0–1ms later**:

```json
{"ts":"2026-07-21T04:40:43.643Z","event":"session-killed","sessionId":"<wtA>@@94f9a1fc","immediate":false}
{"ts":"2026-07-21T04:40:43.644Z","event":"session-killed","sessionId":"<wtA>@@94f9a1fc","immediate":true}

{"ts":"2026-07-21T06:19:36.775Z","event":"session-killed","sessionId":"<wtB>@@755fa75a","immediate":false}
{"ts":"2026-07-21T06:19:36.775Z","event":"session-killed","sessionId":"<wtB>@@755fa75a","immediate":true}
```

Same millisecond. The agent never gets a window to shut down and propagate termination to its own  <br>children — it looks like the force path isn't awaiting the graceful one.

## What I actually noticed

Sessions in a worktree I'd navigated away from were dead when I came back, repeatedly. One worktree  <br>cycled create → kill four times in two hours:

| session | created | killed | lifetime |
| -- | -- | -- | -- |
| `@@94f9a1fc` | 04:12:13 | 04:40:43 | 28m 30s |
| `@@b878dbdf` | 04:57:53 | 05:16:03 | 18m 10s |
| `@@02e4913f` | 05:20:53 | 05:36:51 | 15m 58s |
| `@@a6a01df5` | 06:04:25 | 06:19:28 | **15m 03s** |

A session in a *different* worktree (`@@755fa75a`, created 06:04:33) was killed at 06:19:36 — also  <br>**15m 03s**. Two different worktrees matching to the second seems unlikely to be coincidence, but I  <br>couldn't find a matching timeout constant in the source, so I'm not going to guess at what it means.

## What I don't know

* **Whether other worktrees die at the same moment.** I only ever noticed the one I returned to.
* Some kills in my log are isolated to a single worktree and others are batched across several, so  <br>there may be two different causes here rather than one.

One thing that might narrow it down: `session-killed` is logged in `daemon-server.ts` on *receiving*  <br>a `kill` RPC. So these are client-initiated — the daemon isn't reaping on its own. Something in the  <br>app is asking for them.

## Secondary, possibly unrelated: uncaught `write EAGAIN`

I also have one daemon death from an unhandled exception:

```json
{"src":"daemon","ts":"2026-07-21T05:37:53.761Z","pid":6248,
 "event":"uncaught-exception-fatal","name":"Error","message":"write EAGAIN"}
```

It fired 14ms after a `session-killed` (`immediate:true`), and the daemon auto-restarted 74s later.  <br>This may be the same teardown path or entirely unrelated — I can't tell. I'm noting it because I  <br>found no other issue mentioning `EAGAIN`, and stablyai/orca#8048 described a *silent* crash with no exception  <br>recorded, so the explicit message may be new information. A pipe write returning `EAGAIN` should  <br>presumably be retried under backpressure rather than propagating as an uncaught exception.

## Also: the runtime lists dead sessions as connected

`orca terminal list` reported 15 terminals as `connected: true`, six of which the daemon log showed  <br>as killed the previous day. After I manually killed the orphan trees, the list dropped to 9 — so the  <br>runtime's terminal registry isn't reconciled against the daemon's actual session set.

This is likely how stale ptyIds get reused, giving:

```
DaemonProtocolError: Session not found: <worktreeId>@@648a0e6f
```

The `@@<hash>` suffix rerolls on every terminal recreation (one worktree went through four in a  <br>day), so anything caching a ptyId across a restart will hit this.

## Possibly relevant: this host also runs the remote server

Flagging this in case it matters — this machine serves via Orca's remote server, and the daemon log  <br>carries `serve-<uuid>` sessions alongside the worktree ones, which also get killed. I don't know  <br>whether the remote-server path shares this teardown code or changes who sends the `kill` RPC, but  <br>since most reports are probably local-only, the difference seemed worth mentioning.

## What I'm hoping for

Mostly I'd like to understand **why the sessions are being killed**, and to be able to keep them  <br>alive. When one dies I lose the agent's in-flight context, and the orphan doesn't even stop — in one  <br>case it kept running tools for 4 minutes after its terminal had vanished from the UI, so the work  <br>wasn't halted, just unreachable.

Happy to gather whatever's useful — I have `daemon.log` and can reproduce the accumulation just by  <br>using the app normally. If there's a debug flag that logs the origin of the `kill` RPC, that would  <br>presumably pin down the caller immediately.

## Environment

* Orca 1.4.147 (updated from 1.4.146 the same day), daemon protocol v24
* Windows 11 Pro 26200
* Also running Orca remote server on this host
* Agents in the affected terminals: Codex CLI (`@openai/codex`) and Claude Code

Thanks for building this — running several agents across worktrees in parallel has genuinely changed  <br>how I work, and I'd rather send an imperfect report than let this sit unreported.

bbingz:
## Summary

- Opt agent-owned Windows ConPTY roots into a kill-on-close Job Object before their primary thread is resumed.
- Reap descendants on explicit kill and on natural root exit without a PID snapshot or PowerShell/taskkill sweep.
- Keep plain terminals, WSL, Unix, and the current SSH relay path on their existing behavior.
- Force patched Windows `node-pty` installs to build the matching native addon from source.

Fixes #9704

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Focused verification passed: 10 files / 350 tests, changed-file oxlint and oxfmt, max-lines, patch-hash, lockfile, and diff checks. `pnpm run build:desktop` also passed.

On Windows 11 `10.0.22631` with Node `24.10.0` and VS2019 Build Tools, the final package patch installed cleanly and Orca's product path, `npm_config_build_from_source=true pnpm rebuild node-pty`, produced `winpty-agent.exe`, `winpty.dll`, `conpty.node`, `conpty_console_list.node`, and `pty.node` (`gyp info ok`). A native ConPTY smoke verified that explicit kill and natural root exit both reaped parent/child/grandchild processes, while the plain-terminal control left detached descendants alive until harness cleanup.

The full `pnpm lint` command is currently blocked by the existing `src/renderer/src/components/skills/skill-freshness-group.ts:101` switch-exhaustiveness error on `origin/main`. The full repository test command and the exact template command `pnpm build` were not run; focused tests plus `pnpm run build:desktop` and the Windows native build/smoke were used instead.

## AI Review Report

Independent AI reviews checked Job Object ownership, suspend/assign/resume ordering, handle races, natural-exit cleanup, resume failure, direct-root fallback, daemon/local ownership, plain-terminal and WSL exclusion, patch integrity, and cross-platform routing. Earlier review findings about cleanup lock placement and native warning-prone handle reuse were corrected and covered by contract tests; the final review returned spec-compliance `PASS` and code-quality `APPROVED`.

The cross-platform review explicitly covered macOS, Linux, Windows, WSL, daemon/local providers, Electron native rebuild selection, shell behavior, and SSH relay scope. No shortcut, label, or UI behavior changed.

## Security Audit

Reviewed process-termination scope, native handle lifetime, failure cleanup, command execution, path handling, IPC/API threading, dependencies, and secrets. Job Object ownership is opt-in only for recognized agent PTYs; ordinary terminals are excluded. The implementation removes the proposed PowerShell/PID sweep, introduces no user-derived shell command, adds no dependency or secret handling, and falls back only to terminating the known root when Job setup fails.

## Notes

Windows local and daemon-owned agent PTYs are covered, including the remote-server daemon path. Windows SSH relay remains unchanged because it installs vanilla `node-pty`; shipping the patched native addon through that boundary is tracked separately by TODO #1693.

The issue also discusses why kill requests may be initiated, transient `EAGAIN`, and stale registry entries. This PR fixes descendant cleanup once the agent PTY is killed; it does not claim to change those separate initiators or diagnostics.

## ELI5

On Windows, killing an agent terminal could leave child processes behind. Agent ConPTY sessions join a kill-on-close Job Object so the whole tree dies with the root without sweeping every PID by hand.

### DIFF
--- a/config/patches/node-pty@1.1.0.patch
+++ b/config/patches/node-pty@1.1.0.patch
@@ -1,5 +1,5 @@
 diff --git a/binding.gyp b/binding.gyp
-index 5f63978b07ab50aaf7523219a2170ec737a6b5db..bbd9e06136e8922f40b5779e35d4fc835f1479ab 100644
+index 5f63978b07ab50aaf7523219a2170ec737a6b5db..b3309a07ef99dea7967d7bdd04b9fc3500acacae 100644
 --- a/binding.gyp
 +++ b/binding.gyp
 @@ -5,9 +5,6 @@
@@ -12,28 +12,11 @@ index 5f63978b07ab50aaf7523219a2170ec737a6b5db..bbd9e06136e8922f40b5779e35d4fc83
          'msvs_settings': {
              'VCCLCompilerTool': {
                'AdditionalOptions': [
-@@ -88,6 +85,16 @@
-               'libraries!': [
-                 '-lutil'
-               ]
-+            }],
-+            # Orca: pair with the .symver pins in pty.cc. Force the real
-+            # libutil.so.1/libpthread.so.0 into DT_NEEDED (gcc's default
-+            # --as-needed drops them because the pinned symbols resolve from
-+            # libc's compat aliases at build time) so openpty/forkpty/
-+            # pthread_sigmask still resolve on Ubuntu 20.04 (glibc 2.31).
-+            ['OS=="linux"', {
-+              'ldflags': [
-+                '-Wl,--no-as-needed,-l:libutil.so.1,-l:libpthread.so.0,--as-needed'
-+              ]
-             }]
-           ]
-         }
 diff --git a/deps/winpty/src/winpty.gyp b/deps/winpty/src/winpty.gyp
-index 1ac5758bedd8cf54f32280dea4e4aeb5afdee30d..e619813759c6f14694838bdfbd0ea5f8360130ef 100644
+index 1ac5758bedd8cf54f32280dea4e4aeb5afdee30d..4debd75854 100644
 --- a/deps/winpty/src/winpty.gyp
 +++ b/deps/winpty/src/winpty.gyp
-@@ -10,7 +10,7 @@
+@@ -10,9 +10,10 @@
      #    make -j4 CXX=i686-w64-mingw32-g++ LDFLAGS="-static -static-libgcc -static-libstdc++"
 
      'variables': {
@@ -41,8 +24,11 @@ index 1ac5758bedd8cf54f32280dea4e4aeb5afdee30d..e619813759c6f14694838bdfbd0ea5f8
 +        'WINPTY_COMMIT_HASH%': '<!(cmd /c "cd shared && .\\GetCommitHash.bat")',
      },
      'target_defaults' : {
++        # Why: target-level opt-outs below override node-gyp's addon defaults for these non-Node binaries.
          'defines' : [
-@@ -22,7 +22,7 @@
+             'UNICODE',
+             '_UNICODE',
+@@ -22,12 +23,13 @@
          'include_dirs': [
              # Add the 'src/gen' directory to the include path and force gyp to
              # run the script (re)generating the version header.
@@ -51,7 +37,13 @@ index 1ac5758bedd8cf54f32280dea4e4aeb5afdee30d..e619813759c6f14694838bdfbd0ea5f8
          ]
      },
      'targets' : [
-@@ -40,9 +40,6 @@
+         {
+             'target_name' : 'winpty-agent',
++            'win_delay_load_hook' : 'false',
+             'type' : 'executable',
+             'include_dirs' : [
+                 'include',
+@@ -40,9 +42,6 @@
                  '-lshell32',
                  '-luser32',
              ],
@@ -61,7 +53,15 @@ index 1ac5758bedd8cf54f32280dea4e4aeb5afdee30d..e619813759c6f14694838bdfbd0ea5f8
              'msvs_settings': {
                  # Specify this setting here to override a setting from somewhere
                  # else, such as node's common.gypi.
-@@ -142,9 +139,6 @@
+@@ -131,6 +130,7 @@
+         },
+         {
+             'target_name' : 'winpty',
++            'win_delay_load_hook' : 'false',
+             'type' : 'shared_library',
+             'include_dirs' : [
+                 'include',
+@@ -142,9 +142,6 @@
                  '-ladvapi32',
                  '-luser32',
              ],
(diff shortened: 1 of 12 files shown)


bbingz:
Ready for first review — fixes #9704 (Windows agent PTY descendant reaping via Job Objects; also covers the #9045 orphaned-tree failure mode).

**MERGEABLE** on current `main`. Focused tests + Windows native smoke described in the PR body.

bbingz:
Rebased onto current `origin/main` (`6d55c7fa16`).

Conflicts were `pnpm-lock.yaml` (node-pty patch hash) and the final `node-pty` patch file; resolved to the tip patch + matching lock hash.

Focused tests: pty-subprocess / local-pty-provider / descendant-termination / job-object contract **232/232 green**.

bbingz:
## Restack note

Rebased onto latest `main` (force-with-lease). All 6 commits applied cleanly.

Remains the Windows Job Object reaping path that #9804 explicitly deferred (authorized kill of agent ConPTY trees without PowerShell/taskkill). Orthogonal to closeLifecycle / host-authoritative session work.

Coordinates with:
- **#10013** — kill attribution / orphan confirm (who issued the kill)
- **#8467** — unbound idle-shell review (what is safe to kill)

bbingz:
Rebased onto current `main` (`981653f27d`). The scope of this PR changed materially in the process, so flagging it rather than leaving it in the diff:

Upstream #10100 (`fix(win): taskkill agent PTY descendant trees on stop`) landed after this branch's base and solves the stop-time half of #9704 with `taskkill /T /F`. The early commits here implemented their own taskkill/PID-identity path, so during the rebase those were resolved **in favor of upstream's structure** — this PR no longer competes with #10100. What remains on top is the Job Object layer only, which cuts the diff down to 14 files.

Why it is still worth having on top of #10100: the Job Object binds at **spawn** (agent ConPTY, `useConptyJobObject` gated on `isAgentPty`) rather than at stop, so membership is kernel-tracked at creation instead of walked from a root PID at teardown. That keeps reaping correct for reparented children and removes the PID-reuse race a `/T` walk is exposed to. Plain shells and WSL are unaffected — the opt-in is agent-only, and `windows-process-tree-kill.ts` stays on main's path for everything else.

Adjacent upstream work checked and not overlapping: #10069 (never sweep descendants of an already-exited PTY root, POSIX hardening) and #10011 (orphan terminal recovery).

Verified: `tsc --noEmit -p config/tsconfig.node.json` clean, and 6 test files / 235 tests passing across `pty-descendant-termination`, `local-pty-provider`, `local-pty-utils-windows-fallback`, `windows-process-tree-kill`, `pty-subprocess`, `terminal-session-teardown`. Not verified on a real Windows host — the Job Object path is covered by mocks here, so a Windows run before merge would be worthwhile.

bbingz:

PR #9767, opened 2026-07-21
fix(terminal): reap Windows ConPTY process tree on PTY force-kill

OrcaWin:
## Problem

Deleting a worktree on Windows failed with `Error: Failed to physically stop every PTY for worktree: …` when a terminal in that worktree had a wedged descendant process — e.g. a hung `git.exe` during a source-control operation (repro'd against `issue-7991-source-control-freeze-repro`).

## Root cause

Destructive worktree removal is **fail-closed**: `killAllProcessesForWorktree({ requirePhysicalStop: true })` refuses to delete the worktree directory until every PTY is proven physically stopped (so it never removes files a live process holds). Windows PTY teardown could not deliver that guarantee:

- The daemon spawns node-pty with `useConptyDll: true` — the backend that **skips the console-process-list reap** the classic ConPTY backend performs (`node_modules/node-pty/lib/windowsPtyAgent.js`).
- So `node-pty.kill()` only closes the pseudoconsole, which `CTRL_CLOSE`s console members and relies on cooperation. A wedged descendant ignores it, survives, keeps the console open, and the shell's `onExit` misses the physical-exit deadline → the error.
- The POSIX descendant sweep (`captureDescendantSnapshot`) and process-group kill are explicit Windows no-ops, so nothing replaced node-pty's skipped reap.

## Fix

Force-kill the whole process tree with `taskkill /t /f` **before** closing the ConPTY, in both Windows PTY kill paths (daemon subprocess + local provider), via a shared `src/main/pty/windows-pty-tree-kill.ts` helper. `taskkill /f` is `TerminateProcess` — it needs no cooperation, so it reaps the descendant that ignores `CTRL_CLOSE`, and doing it first means the shell's clean exit can't be delayed by wedged children.

A Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` would be the atomic gold standard, but node-pty exposes no spawn-time job assignment, and anything `taskkill`'s parent walk misses has detached from the console and so is not what blocks physical exit (that's orphan cleanup, handled elsewhere). Deferred and documented in the manifest gap.

## Proof

- **Reliability invariant:** `terminal-session.destructive-removal-physical-stop` — destructive worktree removal must prove every owned Windows PTY (and its console-holding descendants) is physically stopped before filesystem deletion. Motivated by #7991.
- **Material impact:** closes a documented Windows gap in the descendant-teardown reliability gate (`config/reliability-gates.jsonc` `knownGaps`); a wedged terminal child can no longer hard-block worktree deletion.
- **Product change type:** product hardening + gate coverage.
- **Correctness evidence (live):** on real ConPTY with `useConptyDll: true`, `node-pty.kill()` alone leaves a descendant alive indefinitely (`childAliveAfter: true`); `taskkill /t /f` then kill reaps the tree in ~1s (`childAliveAfter: false`). That is direct process-absence evidence.
- **Deterministic tests:** `src/main/pty/windows-pty-tree-kill.test.ts` (helper: args, no-throw, invalid-pid guard); `pty-subprocess.test.ts` (Windows `forceKill` walks the tree even after a graceful kill — the exact regression — closes ConPTY once, no walk on a reaped pid); `local-pty-provider.test.ts` (Windows shutdown walks the tree, POSIX does not). POSIX `forceKill` tests pinned to a deterministic platform so they don't drift on a Windows runner.
- **No-regression evidence:** `typecheck:node`, `oxlint`, and `check:reliability-gates` pass; adjacent gate suites (`session`, `terminal-host`, `worktree-teardown`, `pty-descendant-termination`) green. No new polling/scans/subprocess churn — one best-effort, non-blocking `taskkill` per destructive PTY kill; physical-exit wait still owns completion.
- **Provider/platform coverage:** local + daemon on Windows covered. macOS/Linux unaffected (POSIX branch unchanged). SSH/relay-to-Windows-host shares the gap class and is a tracked follow-up, not addressed here.
- **Residual gaps:** the end-to-end worktree-remove timeout under a genuinely wedged git was not reproduced synthetically (clean shells exit promptly), so the fix's effect on that exact deadline rests on the descendant-reap evidence + reasoning, not a red/green live worktree-remove run. Kept visible in the manifest.
- **Promotion status:** gate stays non-blocking; no promotion. **Demotion/rollback:** revert this commit if the tree-kill ever reaps a process outside the target PTY's tree or regresses teardown latency.


OrcaWin:
Closing as superseded. This same Windows bug (#7991 — "Failed to physically stop every PTY for worktree", caused by node-pty's `useConptyDll: true` skipping the ConPTY console-process-list reap so a wedged descendant keeps the pseudoconsole open) has already been fixed on `main` via #10100 ("fix(win): taskkill agent PTY descendant trees on stop"), together with #10004 / #8706 / #10069.

Main's implementation is strictly more complete than this PR:
- shared cross-platform `killWithDescendantSweep(pid, killRoot, { ownsRoot })` (`src/main/pty-descendant-termination.ts`) with a PID-recycling guard;
- `terminateWindowsProcessTree` runs `taskkill /T /F` as an awaitable Promise with a 5s timeout (this PR's helper was fire-and-forget);
- wired into agent sessions, plain shells (`terminal-session-teardown.ts`), and the local provider.

This branch also re-conflicts with `main` on every sync because both touch the same PTY-kill files. No unique coverage is lost by closing.

PR #9780, opened 2026-07-21
feat(mobile): mount ProtocolBlockScreen when protocol compat is blocked

brennanb2025:
## Summary

`ProtocolBlockScreen` was only mounted inside the base host-list screen, so deep links and nested `/h/[hostId]` routes could continue into an incompatible desktop instead of showing the required update screen. This moves compatibility enforcement to the host layout choke point used by every host route.

- Adds `HostProtocolGate` above the full host UI (sidebar + detail stack), while the host list and other hosts remain usable.
- Waits for the first connected-host compatibility probe before mounting nested routes, preventing newer RPCs from firing before compatibility is known.
- Shares the loaded `status.get` gates with `HostScreen`, avoiding the duplicate remote probe the parent gate would otherwise add on common phone/tablet paths.
- Keeps already-validated routes mounted during reconnect revalidation so active terminals are not torn down; a later blocked verdict still replaces the route.
- Preserves fail-open behavior for a failed status probe with conservative feature gates, and exposes both recovery CTAs: App Store for `mobile-too-old`, GitHub Releases for `desktop-too-old`.
- Leaves the canonical protocol evaluator and its Metro mirror contract unchanged.

## Tests

`HostProtocolGate.test.ts` exercises the real `useHostStatusGates` → `evaluateCompat` → `ProtocolBlockScreen` wiring with only client acquisition mocked:

1. `mobile-too-old` renders the App Store block screen and removes host content
2. `desktop-too-old` renders the GitHub Releases block screen and removes host content
3. compatible status renders host content and shares capabilities with one `status.get`
4. connection-pending state leaves the host connection UI available
5. first connected-host probe fences child-route mounting
6. reconnect revalidation keeps an already-validated route mounted
7. a failed status probe fails open without enabling capability-gated actions

Verification completed locally:

- Mobile: 304 test files passed; 2,177 tests passed, 2 skipped
- Mobile `tsc --noEmit`, oxlint, and oxfmt clean
- Root Node typecheck and max-lines ratchet clean
- Electron production build clean
- Isolated Electron app + daemon boot clean

## Screenshots / manual QA

Manual iPhone 17 Pro simulator coverage includes current/current compatibility, both forced block reasons, `Back to hosts`, a Resume deep link into a nested session route, and a final compatible-host pass after restoring the production constants.

| Desktop too old | Mobile too old | Compatible path restored |
|---|---|---|
| Desktop too old | Mobile too old | Compatible host restored |

Detailed review notes and Electron boot evidence: https://github.com/stablyai/orca/pull/9780#issuecomment-5040765713

### DIFF
--- a/mobile/app/h/[hostId]/index.tsx
+++ b/mobile/app/h/[hostId]/index.tsx
@@ -61,7 +61,7 @@ import { buildWorktreeNavigationActions } from '../../../src/agent-history/workt
 import { floatingWorkspaceSessionPath } from '../../../src/session/floating-workspace'
 import { ConfirmModal } from '../../../src/components/ConfirmModal'
 import { BottomDrawer } from '../../../src/components/BottomDrawer'
-import { ProtocolBlockScreen } from '../../../src/components/ProtocolBlockScreen'
+import { useHostProtocolGates } from '../../../src/components/HostProtocolGate'
 import { AuthFailedBanner } from '../../../src/components/AuthFailedBanner'
 import { MobileSearchField } from '../../../src/components/MobileSearchField'
 import { WorkspaceDetailPlaceholder } from '../../../src/components/WorkspaceDetailPlaceholder'
@@ -70,7 +70,6 @@ import { setCachedRepos } from '../../../src/cache/repo-cache'
 import { colors, radii, spacing, typography } from '../../../src/theme/mobile-theme'
 import { useResponsiveLayout } from '../../../src/layout/responsive-layout'
 import { leaveHostRoute } from '../../../src/host-route-exit'
-import { useHostStatusGates } from '../../../src/transport/host-status-gates'
 import { loadPinnedIds, savePinnedIds } from '../../../src/storage/preferences'
 import {
   createInitialHostRouteActionState,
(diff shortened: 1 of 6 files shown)


brennanb2025:
## Deep review + manual QA follow-up

I found and fixed two meaningful issues in `7d5501e7d1`:

- The parent gate and `HostScreen` each issued `status.get`, so common phone and tablet routes paid for the same remote probe twice. The gate now provides the loaded status gates to the host screen; the regression test asserts exactly one request.
- A connected host mounted child routes while compatibility was still unknown, allowing nested screens to call newer RPCs before the hard block arrived. The first probe now fences route mounting. Reconnect probes keep an already-validated route mounted so terminals are not torn down, and a status error still fails open with conservative feature gates.

Automated verification:

- Mobile: 304 files passed; 2,177 tests passed, 2 skipped
- Mobile typecheck, oxlint, and oxfmt: clean
- Root Node typecheck and max-lines ratchet: clean
- Electron production build: clean
- Isolated Electron dev boot: app + daemon started successfully

Manual verification used a dedicated iPhone 17 Pro simulator (iOS 26.5) paired to a temporary desktop runtime. I exercised current/current compatibility, both forced block reasons, `Back to hosts`, a Resume deep link into the nested session route, and restored the production constants before the final compatible-host pass.

| Desktop too old | Mobile too old | Restored compatible path |
|---|---|---|
| Desktop too old | Mobile too old | Compatible host restored |

<details>
<summary>Isolated Electron boot evidence</summary>

Isolated Electron dev boot

</details>

brennanb2025:
### Android update-path UI preview

The shared native block screen rendered with the Android update target. Android now points users to GitHub Releases until Play Store distribution exists. (Captured in the available iOS simulator because this machine has no Android SDK/AVD configured.)

Android update-path UI preview

PR #9781, opened 2026-07-21
feat(mobile): soft update-your-app nudge from desktop-advertised versions

brennanb2025:
## Summary

Part of the mobile version-skew incident response (Discord reports 2026-07-20): desktop ships daily while mobile ships less often through platform-specific release channels with zero update awareness, so phones go silently stale until features break.

- **Desktop**: `status.get` now reports `appVersion` (the desktop's own version, from `ORCA_APP_VERSION`) and a `recommendedMobileAppVersions` advisory. The Android recommendation is **derived live from GitHub Releases** — the newest published exact `mobile-android-v*` tag — cached (6h TTL) and fail-open, so it can never go stale and requires **no per-release bookkeeping**. iOS is intentionally not sourced: it ships through the App Store (auto-updates), and hard minimum-version cutoffs are already handled by `ProtocolBlockScreen`, so there is no soft iOS nudge.
- **Mobile**: compares its installed version (`expo-constants`) against the recommendation for the running platform and shows a dismissible banner on the host screen — "Update the Orca app for the best experience." The Android update button opens the mobile-filtered GitHub releases list (version-agnostic, always the newest APK).

## Source of truth

`RECOMMENDED_MOBILE_APP_VERSIONS` (a hand-bumped constant) is **removed**. The desktop reads the release channel itself via `mobile-android-release-feed.ts` (`net.fetch` against the GitHub Releases API, the same mechanism the updater already uses). `status.get` is a hot, synchronous RPC, so it never blocks on the network: it reads a background-refreshed cache (stale-while-revalidate) and a failed/empty fetch simply omits the field → no banner. A transient outage keeps the last-known version rather than dropping the nudge.

## Fail-open behavior

An older desktop that doesn't send the field, an unreachable/rate-limited/offline host, an unknown latest version, a missing platform recommendation, an unsupported platform, an unparsable/unsafe version segment (including dev builds), or an unsettled dismissal read all mean **no banner**. Version compare is numeric per segment, so `0.0.9` correctly ranks below `0.0.32`.

## Dismissal behavior

Dismissal persists the dismissed recommended version in AsyncStorage. A newer recommendation resurfaces the nudge, while switching to a stale desktop advertising an older version does not re-nag the user. The banner is held until the persisted dismissal loads so a dismisser never sees it flash.

## Performance

No polling or provider fanout on the mobile side. The feature reuses the existing `status.get`, performs at most one follow-up when a cold desktop refresh is pending, compares constant-size versions during render, and reads one AsyncStorage key only after a desktop advertises recommendation metadata. On the desktop, the GitHub lookup is cached at a 6h TTL and refreshed off the `status.get` hot path (never awaited), amounting to a few unauthenticated calls per day per host.

## Tests

- Pure logic coverage for platform selection, numeric version comparison, malformed values, and dismissal ordering
- Release-feed parsing: newest exact-tag selection, draft skip, GitHub prerelease-channel acceptance, non-semver rejection, fail-open to null
- `status.get` reports Android-only once the feed resolves and omits the field while the cache is cold
- Version-agnostic update-link coverage for iOS, the Android mobile-filtered releases list, and unsupported platforms
- Dismissal persistence round-trips and `host-status-gates` response/race coverage
- Desktop `OrcaRuntimeService` suite (789 passing), full mobile suite, mobile lint/format/TypeScript, desktop node/CLI/web TypeScript

Note: the host screen was at the mobile max-lines cap, so `isErrorVerdict` moved into `transport/connection-health.ts` (its `ConnectionVerdict` home) instead of bumping the limit; the update URL logic lives in `mobile-app-update-link.ts` and is shared with `ProtocolBlockScreen`.


brennanb2025:
Manual QA evidence (Electron + iOS Simulator)

I ran the PR Electron app from this worktree with an isolated profile, paired an iPhone 17 Pro simulator (iOS 26.5), and loaded the PR mobile bundle. To exercise the positive path while the checked-in recommendation remains 0.0.31, I temporarily advertised iOS 0.0.33 against the bundle version 0.0.32; that QA-only override was reverted afterward.

- Electron runtime status advertised the expected app/recommended-version fields; renderer console had 0 errors.
- The connected mobile host rendered the soft update nudge:

Update nudge shown

- Dismissing stored 0.0.33 in AsyncStorage. After fully terminating/relaunching the app and reopening the host, the nudge remained dismissed:

Update nudge remains dismissed after relaunch

Android emulator tooling was unavailable on this machine; Android platform selection, URL construction, version comparison, and compatibility cases are covered by the automated mobile tests.

PR #9791, opened 2026-07-21
feat(worktrees): support project-level .worktreeinclude (literal paths) for copying gitignored files into worktrees

### LINKED ISSUE #7549: [Feature]: Support project-level .worktreeinclude for copying gitignored files into worktrees
### Problem or use case

Orca creates a git worktree per session/card. A worktree is a fresh checkout, so **gitignored files never carry over** — `.env`, `.env.local`, local secrets, `.vscode/`, `node_modules`, and other local config are absent. Dev servers and builds break in every new worktree until those files are copied by hand.

Orca already has a mechanism for this: `repo.symlinkPaths` clones/symlinks a configured list of paths into new worktrees (`createWorktreeLinkedPaths` in `src/main/ipc/worktree-symlinks.ts`, gated behind the `experimentalWorktreeSymlinks` setting). The limitation is **where that list lives**: it's per-repo config in Orca's own settings, so it is

- not version-controlled,
- not shared with teammates,
- not portable across machines or tools, and
- something every contributor must re-configure by hand.

### Proposed solution

Support a **`.worktreeinclude`** file at the project root as a source of worktree copy patterns, following the convention Claude Code established (see [Claude Code worktree docs](https://code.claude.com/docs/en/worktrees)).

- File uses `.gitignore` syntax; patterns list files/dirs to bring into a new worktree (e.g. `.env`, `.env.local`, `config/secrets.json`).
- Matching entries are copied into each newly created worktree — Orca can feed these patterns straight into the existing `createWorktreeLinkedPaths` path (APFS clone on macOS, symlink fallback elsewhere), so no new copy machinery is needed.
- Merge `.worktreeinclude` patterns with the existing `repo.symlinkPaths` list rather than replacing it.

Why in-repo beats per-user settings:

- **Version-controlled + shared** — the include list travels with the repo, so every teammate and every machine gets the same worktree setup with zero manual config.
- **Tool interop** — Claude Code already reads `.worktreeinclude`, and opencode added the same ([opencode#25379](https://github.com/anomalyco/opencode/pull/25379)). A repo drops one file and it works across all three.

- **Existing `experimentalWorktreeSymlinks` + `repo.symlinkPaths`** — works today but is per-user/per-machine Orca config, not portable or shared (the gap this request closes).
- **WorktreeCreate-style hooks** — maximally flexible but heavyweight; a shell script per repo just to `cp` a few files.
- **Known caveat to mirror from Claude Code**: only files matching a pattern *and* already gitignored are copied, so tracked files are never duplicated. Non-gitignored files are currently out of scope there too — see [claude-code#71913](https://github.com/anthropics/claude-code/issues/71913). Worth documenting the same constraint, or deciding up front whether Orca wants to lift it.

brennanb2025:
Closes #7549

## What

When creating a **local** worktree, Orca reads `.worktreeinclude` at the repo root and copies each listed **gitignored** path from the primary checkout into the new worktree. The include list is version-controlled and travels with the repo, so every teammate and machine gets the same worktree setup with zero per-user config — and repos that already carry this file for other tools work in Orca unchanged.

## Scope — focused, safe-for-now subset

The ticket asks for a size-M feature that reuses existing worktree machinery. This PR is deliberately the minimal implementation that solves the reported problem:

- Each non-comment line is a **literal file or directory** anchored to the repo root (`.env`, `.env.local`, `.vscode/`, `node_modules`, `config/secrets.json` — every example in the issue).
- **Glob (`*`/`?`) and negation (`!`) are not supported yet** — such lines are skipped with a warning rather than silently mishandled. Full `.gitignore` glob syntax is a documented follow-up. This intentionally keeps the change small and avoids a user-controlled-regex surface for now.

## Semantics

- **Gitignored-only** (the caveat called out in the issue): a path is copied only if it exists in the primary checkout **and** `git check-ignore` confirms it is ignored. Tracked files are already in a fresh worktree; unignored files would show as spurious diffs. The filesystem + git handle existence and case-sensitivity, so there's no separate case logic.
- **Copy, not symlink** — the one deliberate deviation from the issue's "reuse the symlink fallback" suggestion: the convention promises each worktree its own private copy (editing a worktree's `.env` must not touch the primary). macOS uses the existing APFS clone-copy; elsewhere a real `fs.cp`, never a symlink.
- Merges with (never replaces) the existing per-user Worktree Symlinks repo setting.
- Fails safe: unreadable file, git errors, unsafe patterns (absolute, `..`, `.git/`), and per-path copy failures are logged and skipped — worktree creation is never blocked.
- Remote (SSH) creation skips it, same as `symlinkPaths` (needs a relay method + auth surface).

## Changes (net ~500 lines)

- `src/main/git/worktree-include-file.ts` — parse + resolve `.worktreeinclude` to concrete gitignored paths (literal-only).
- `src/main/ipc/worktree-symlinks.ts` — `createWorktreeCopiedPaths`: copy-mode materialization sharing the existing link loop.
- `src/main/ipc/worktree-apfs-clone.ts` — APFS clone helpers extracted from `worktree-symlinks.ts` (max-lines), plus a df/diskutil probe timeout.
- `worktree-remote.ts` (`createLocalWorktree`) + `orca-runtime.ts` — resolve + copy after the symlink pass, before setup hooks.

## Testing

- `vitest`: `worktree-include-file.test.ts` + copy-mode cases in `worktree-symlinks.test.ts`; full `worktrees` + `orca-runtime` suites green (1031 tests).
- `typecheck:node`, `oxlint`, `oxfmt`, max-lines ratchet: clean.

## Follow-ups

- Glob / negation (full `.gitignore` syntax).
- Remote (SSH) support.

---
_Note: an earlier revision of this branch carried a much larger hardening layer (glob engine, CPU/byte budgets, git enumeration, filesystem caches). It has been trimmed to the focused scope above; the net diff reflects only the minimal feature._


brennanb2025:
## Review + perf audit round

Ran an adversarial code review and a `/perf` audit of the copy path. Outcome:

**Fixed**
- **(HIGH, correctness)** A listed entry that is itself a gitignored **symlink** was copied *as a symlink* (`fs.cp` default `dereference:false`, and the darwin APFS branch was skipped for all symlink sources). Editing the worktree's copy then wrote through to the shared/primary target — inverting copy-mode's "each worktree owns its files" guarantee, and escaping the worktree if the link pointed outside it. Now the top-level symlink is dereferenced (realpath) in copy mode so content is copied; regression tests cover the file and directory cases (no leak to primary).
- **(perf)** `assertSameApfsVolume` ran `df`+`diskutil` **per copied path** (4 subprocesses each) → ~4N short-lived processes on the macOS create hot path, all re-probing one volume. Added a per-materialization device-keyed cache: one probe per distinct volume (4N → ~4), with a test asserting the probe count no longer scales with path count.

**Documented as known limits (no code — kept the feature minimal)**
- Large gitignored directories copy synchronously; instant on macOS (APFS clone) but a real recursive copy on Linux/Windows, so listing a multi-GB tree (e.g. `node_modules`) slows creation there. Prefer the per-user Worktree Symlinks setting for big shared trees. Copy never *fails* creation (per-path errors are logged/skipped) — it can only slow it.
- Symlinks nested *inside* a copied directory are preserved as-is (conventional `cp -R` semantics).
- A path in both `.worktreeinclude` and the per-user symlink list stays a symlink for that entry (symlink pass runs first; copy pass skips an existing target).

**Verified solid by review:** gitignored-only filtering (tracked files never copied), resolver failure-isolation (never throws/blocks creation; bounded), double-layered path-traversal guards, remote/SSH correctly skipped, WSL git routing threaded, copy ordered after symlinks and before setup hooks.

All gates green: typecheck, oxlint, oxfmt, max-lines ratchet, 1115 unit tests.

grundmanise:
Hi @brennanb2025, just curios of `Glob / negation (full .gitignore syntax)` is in the pipeline already?

PR #9796, opened 2026-07-21
fix(mobile): replay terminal creates killed by transport cutover

brennanb2025:
## Problem

Tapping a quick command on mobile shortly after opening a session over relay could fail with a bare **"Couldn't run codex review"** toast. Desktop-side there was zero trace — no PTY, no tab, nothing in the daemon log: the request never arrived.

Root cause: the relay→direct cutover (`migrateTo` in `stable-logical-rpc-client.ts`) rejects all in-flight requests with `LogicalClientCutoverError` while `connState` stays `'connected'`. The cutover fires seconds after connecting over relay — exactly when a quick command gets tapped after opening a session. Same mechanism as the capability-probe loss addressed in #9794.

## Fix

- Replay `session.tabs.createTerminal` with the **same** `clientMutationId` when a cutover kills it, bounded to five authenticated cutovers.
- Keep mutation-keyed creates host-owned when the original socket closes, so a replay joins the in-flight create instead of starting a second terminal. Non-idempotent creates retain their existing disconnect cancellation.
- Advertise `session.tabs.create-terminal-idempotency.v1` only with that disconnect-safe behavior, and require this exact capability before mobile replays. Hosts that only advertise Quick Commands fail closed rather than risking a duplicate terminal.
- Surface the underlying RPC/transport reason in the launch error and report a disconnected tap explicitly.

Timeout-shaped losses that are not identified as cutovers are intentionally not replayed.

## Reliability contract

- **Invariant:** `terminal-session.create-cutover-idempotency` — one mutation creates at most one terminal across relay/direct connection migration.
- **Failure source:** an authenticated transport replacement rejects a client request while the old host request may be absent, in flight, or already committed.
- **Oracle:** fault-injected tests abort the original request, replay the same mutation, and require exactly one renderer `terminal:requestTabCreate`; client tests require the same key, explicit capability, non-cutover propagation, and a six-attempt hard ceiling.
- **Gate:** accepted manifest gap; no exact `config/reliability-gates.jsonc` entry exists yet. Deterministic mobile and host unit contracts cover this PR.
- **Provider/platform coverage:** the ownership/dedupe logic is provider- and platform-neutral and runs before local, daemon, SSH, or WSL terminal realization. Mobile/relay cutover is covered deterministically; no filesystem, path, shell, or desktop-only assumption was added.
- **Performance budget:** ordinary launches remain one RPC. Only explicit authenticated cutovers can add work, capped at five replays; the host emits one renderer create for a shared mutation. No polling, sleeps, scans, subprocesses, provider fanout, or persistent listeners were added.
- **Diagnostics:** failure toasts retain the action label and append the transport/RPC reason.
- **Residual gap:** no packaged-device live relay→direct cutover artifact is attached; old hosts without the new capability retain the original fail-safe error rather than replaying.

## Verification

- Mobile: 304 files, 2,178 tests passed.
- Host/runtime: 3 focused files, 819 tests passed, including the full `orca-runtime.test.ts`.
- Mobile and Node TypeScript checks passed.
- Full mobile lint and changed-file host lint passed.
- Reliability manifest and max-lines ratchet checks passed.

### DIFF
--- a/mobile/app/h/[hostId]/session/[worktreeId].tsx
+++ b/mobile/app/h/[hostId]/session/[worktreeId].tsx
@@ -77,10 +77,13 @@ import { SessionDockColumn } from '../../../../src/session/SessionDockColumn'
 import { MobileSessionHeaderIconButton } from '../../../../src/session/MobileSessionHeaderIconButton'
 import { MobileSessionHeaderMoreActionsSheet } from '../../../../src/session/MobileSessionHeaderMoreActionsSheet'
 import { QuickCommandsSheet } from '../../../../src/session/QuickCommandsSheet'
-import { sendSessionTerminalCreateResilient } from '../../../../src/session/session-terminal-create-retry'
+import {
+  sendSessionTerminalCreateResilient,
+  supportsSessionTerminalCreateCutoverRetry
+} from '../../../../src/session/session-terminal-create-retry'
 import {
   buildMobileQuickCommandLaunch,
-  describeQuickCommandLaunchFailure,
+  formatQuickCommandFailure,
   supportsMobileQuickCommands,
   type MobileQuickCommandLaunch
 } from '../../../../src/terminal/quick-commands'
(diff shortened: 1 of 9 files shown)


brennanb2025:
### Manual QA evidence

Validated the mobile Quick Command flow against this branch using an iPhone 17 Pro simulator paired to the branch-specific Electron dev app. One tap created exactly one new terminal (1 → 2 tabs), selected Terminal 2, and ran `echo mobile-quick-command-ok`. Electron independently showed the same second terminal and output.

**Mobile**

Mobile Quick Command result

**Electron**

Electron confirmation

Post-conflict validation: mobile tests 2,186 passed (2 skipped), mobile typecheck/lint passed, and runtime/protocol tests 810 passed.

Issue #8104, opened 2026-07-10
[Bug]: Terminal intermittently stops accepting text input


dinfames:
### Operating system

macOS

### Orca version

Version 1.4.132 (1.4.132)

### Description

     Text input in the terminal intermittently stops working. This happens frequently, and the terminal no longer accepts
     typed characters even though it remains open and appears responsive.

     ### Steps to reproduce

     1. Open a terminal in Orca.
     2. Use the terminal normally for a while.
     3. Attempt to type a command after switching windows, terminals, or applications.

     ### Expected behavior

     The terminal should consistently accept keyboard input.

     ### Actual behavior

     The terminal frequently stops accepting typed text. Restarting or reopening the terminal may be required.

     ### Frequency

     Frequently intermittent.

     ### Environment

     - macOS
     - Orca

brennanb2025:
Thanks for the report — we've been chasing a family of bugs with this symptom, and a fix for one likely cause just landed (it will be in the next release). To figure out whether your case is that one or a different variant, two questions:

1. When the terminal stops responding, does typing still *do* anything invisibly? E.g., type `touch /tmp/orca-test` and press Enter — does the file get created even though nothing echoed? (This tells us whether input is dead or the output display is frozen.)
2. Does it tend to happen after the Mac has slept or sat idle, or purely from switching windows/apps while awake?

I'll ping here again when the release with the possible fix is out.

RakhimovY:
Reproduced on macOS with Orca v1.4.139 immediately after the auto-update.

  All terminal panes stopped accepting keyboard input. The PTY and shell remain
  healthy:
  - connected: true
  - writable: true
  - prompt remains visible
  - typing a controlled single character does not change lastOutputAt or the
  terminal buffer, so the byte never reaches the PTY

  Terminal focus/shortcut settings were unchanged. v1.4.139 includes #8630, but
  the follow-up fix #8673 was merged 42 minutes after the v1.4.139 release
  commit and is not included in the installed build.

  Please consider cutting a v1.4.140 hotfix containing #8673.

brennanb2025:
Got it, thanks for the report. Taking a look!

moseoh:
Hit this again on **v1.4.147** (which includes #8630, #8673 and #9233) and managed to debug it live before restarting, so I can answer the earlier questions and add a root cause.

[...]

PR #9797, opened 2026-07-21
fix(terminal): fall back to synchronous pty writes when the libuv threadpool wedges (#8104)

moseoh:
## Summary

Fixes the input-loss issue from #8104 that I reproduced on v1.4.147. Every terminal pane, including newly created tabs, stopped accepting keyboard input while output continued and `orca terminal send` still reported success.

Root cause (confirmed by attaching an inspector to the live pty daemon — full forensics in the #8104 comment): the daemon's libuv threadpool can wedge (async fs completions stop being delivered while the event loop stays healthy). node-pty's `CustomWriteStream` delivers pty input via async `fs.write`, whose completion callback then never fires — the queue head never advances and every subsequent keystroke queues forever, silently. Output, session creation, and sockets are all event-loop paths, which is why only input dies.

This PR adds a two-stage detector to the existing node-pty patch. A per-write watchdog checks the stalled write, then an independent canary confirms that the threadpool is stuck before the stream switches permanently to `fs.writeSync`, which bypasses the threadpool. This avoids triggering on a paused event loop or a busy but functioning pool. The sync path drains at most 64KB per event-loop turn and treats a zero-byte write as backpressure. It abandons the stalled task instead of sending it again, because the in-flight write could still complete if the pool recovers and execute a command twice. The dropped byte count is logged. Applying this fallback via the inspector to my wedged live daemon restored input to every pane instantly, without a restart.

I went with the patch (rather than daemon-side code) because the wedge is unobservable from Orca's code — `proc.write()` returns void with no error or ack — and the repo already tracks node-pty behavior fixes in `config/patches/`. Also filing the same problem upstream: microsoft/node-pty#939.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — every file touched by this PR passes; the full run fails on upstream main itself (pre-existing switch-exhaustiveness error in `skill-freshness-group.tsx`, untouched here)
- [x] `pnpm typecheck`
- [x] `pnpm test` — new file green; the full local suite has ~10 pre-existing env-dependent failures (git/ssh/relay integration) that fail identically on the base commit without this change
- [x] `pnpm build`
- [x] Added regression tests (11): the key test reproduces the wedge deterministically by pinning every libuv worker with FIFO reader-opens, then proves input still reaches a real shell through the fallback. I also verified the individual guards by removing them and checking that the corresponding tests fail. These cover the canary probe, abandoning the stalled task, ignoring late completions, the per-turn drain limit, zero-byte backpressure, EAGAIN retries, and the normal path.

## AI Review Report

Reviewed with Claude Code, plus three adversarial design-review rounds with Codex until it judged the change mergeable. Round 1 found a false-positive fallback when the event loop pauses (a timer can fire before a completed write's callback), an unbounded sync drain monopolizing the event loop, and a zero-byte `writeSync` spin — addressed by the canary probe, the per-turn byte budget, and the backpressure retry. Round 2 found a generation race where a stale canary completion from an earlier write could cancel the next write's live canary (recreating the permanent hang) and a single large task bypassing the byte budget in one syscall — addressed by generation tokens and a per-syscall length cap. Each fix is covered by a mutation-verified test (removing the guard makes its test fail). Also checked: double-delivery of input bytes (late async completion after fallback — guarded), input reordering (abandon semantics lose at most the single confirmed-stalled chunk instead of risking duplicate command execution), EAGAIN in both modes, and timer cleanup on dispose. Cross-platform: the change is inside `unixTerminal.js`, so Windows/ConPTY is untouched; macOS and Linux share the patched path; the new test is skipped on win32. Verified the regenerated patch is content-neutral outside `lib/unixTerminal.js` by diffing the installed package trees old-vs-new.

## Security Audit

No new attack surface: the fallback writes the same user keystrokes to the same pty fd captured at construction; no new process spawning, dynamic code, or inputs. `console.error` logs only timeout constants and a byte count. pnpm patch integrity verified (lockfile hash matches the patch file's sha256). The FIFO/`sh -c` usage is test-only with self-generated tmpdir paths.

## Notes

- Why the threadpool wedges is still unknown (evidence points at an Electron utility-process/libuv issue; timeline correlates with a sleep cycle while the pool was idle). This PR makes input survive it and makes the failure observable (one `console.error` with the dropped byte count) instead of silent.
- Possible follow-ups, outside this PR: expose the degraded state to the renderer instead of only logging it, and replace this fallback upstream with a native nonblocking writer using `uv_poll` so the input path does not depend on the threadpool (proposed in microsoft/node-pty#939).
- The patch targets the compiled `lib/unixTerminal.js`, same as the existing node-pty patch hunks. When an upstream fix ships in a node-pty release, this hunk can be dropped with the version bump.

## ELI5

Terminals could stop accepting keyboard input while still showing output when a background thread pool wedged. Writes fall back to a synchronous path so typing keeps working instead of silently dropping.

### DIFF
--- a/config/patches/node-pty@1.1.0.patch
+++ b/config/patches/node-pty@1.1.0.patch
@@ -76,7 +76,7 @@ index 8c4fca9022a6d6f015bca87f61625cde2278f428..0a01730616488119aa21ef441cf3c441
  //# sourceMappingURL=conpty_console_list_agent.js.map
 \ No newline at end of file
 diff --git a/lib/unixTerminal.js b/lib/unixTerminal.js
-index 1ec12f796a822c78fba9ad7f6448c3987e325c23..9de09172a606338e816389012c716508cf6fc259 100644
+index 1ec12f796a822c78fba9ad7f6448c3987e325c23..296fc556aa55d6a917abd35a6a9e40908572dc00 100644
 --- a/lib/unixTerminal.js
 +++ b/lib/unixTerminal.js
 @@ -28,8 +28,12 @@ var native = utils_1.loadNativeModule('pty');
@@ -94,7 +94,7 @@ index 1ec12f796a822c78fba9ad7f6448c3987e325c23..9de09172a606338e816389012c716508
  var DEFAULT_FILE = 'sh';
  var DEFAULT_NAME = 'xterm';
  var DESTROY_SOCKET_TIMEOUT_MS = 200;
-@@ -283,10 +287,15 @@ var CustomWriteStream = /** @class */ (function () {
+@@ -283,10 +287,19 @@ var CustomWriteStream = /** @class */ (function () {
          this._fd = _fd;
          this._encoding = _encoding;
          this._writeQueue = [];
@@ -105,12 +105,16 @@ index 1ec12f796a822c78fba9ad7f6448c3987e325c23..9de09172a606338e816389012c716508
          this._writeImmediate = undefined;
 +        clearTimeout(this._callbackWatchdog);
 +        this._callbackWatchdog = undefined;
++        clearTimeout(this._canaryTimer);
++        this._canaryTimer = undefined;
 +        clearTimeout(this._syncRetryTimer);
 +        this._syncRetryTimer = undefined;
++        clearImmediate(this._drainImmediate);
++        this._drainImmediate = undefined;
      };
      CustomWriteStream.prototype.write = function (data) {
          // Writes are put in a queue and processed asynchronously in order to handle
-@@ -296,7 +305,10 @@ var CustomWriteStream = /** @class */ (function () {
+@@ -296,7 +309,10 @@ var CustomWriteStream = /** @class */ (function () {
              : Buffer.from(data);
          if (buffer.byteLength !== 0) {
              this._writeQueue.push({ buffer: buffer, offset: 0 });
@@ -122,7 +126,7 @@ index 1ec12f796a822c78fba9ad7f6448c3987e325c23..9de09172a606338e816389012c716508
                  this._processWriteQueue();
              }
          }
-@@ -308,10 +320,26 @@ var CustomWriteStream = /** @class */ (function () {
+@@ -308,10 +324,28 @@ var CustomWriteStream = /** @class */ (function () {
              return;
          }
          var task = this._writeQueue[0];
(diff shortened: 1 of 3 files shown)


nireak:
This PR looks right to me, and I think it is one of **two** distinct failures in `CustomWriteStream`. I hit the other one three times in two hours on a headless remote runtime and traced it to the syscall — filed as #11178, with a standalone reproducer. Posting here because your diagnosis identified the component I could not find from outside, and because I suspect **this PR as written would not cover my case**.

## The second failure mode

Same class, same fd, opposite symptom:

| | #8104 / this PR | #11178 |
|---|---|---|
| threadpool | wedged, completions never arrive | **healthy, completions arrive promptly** |
| the write | never completes | **completes, with `EAGAIN`** |
| CPU | normal | **100–104% of one core** |
| output | keeps flowing | **also frozen** |
| symptom | input silently lost | **entire runtime frozen, every session** |

`fs_usage` against the live daemon, 5.07 s:

```
724529 write F=16 errno=35      <- EAGAIN, ~143k/sec, consecutive calls ~6 µs apart
```

`lsof` at the same instant: `16u CHR 15,8 ... /dev/ptmx`.

That rate matches a `setImmediate` turnaround, which points straight at the existing async branch:

```js
if ('code' in err && err.code === 'EAGAIN') {
    // `setImmediate` is used to yield to the event loop and re-attempt the write later.
    _this._writeImmediate = setImmediate(function () { return _this._processWriteQueue(); });
}
```

`setImmediate` yields to the event loop but re-attempts **immediately**, without waiting for the fd to become writable. If the pty's reader has stalled, that is an unbounded busy-wait. Because it lands on the daemon's main thread, and that thread serves every session, one stalled pty freezes the whole runtime — no output, and no keystroke accepted anywhere.

## Why I think this PR does not fix it

The sync fallback handles EAGAIN properly (`_scheduleSyncRetry` → 5 ms `setTimeout`, plus the per-turn budget and zero-write backpressure). But it only engages once the watchdog **and** the canary agree the threadpool has stalled. In my case the threadpool is fine — completions come back fast, just as `EAGAIN` — so I would expect the canary never to fire and the async branch to keep spinning.

I have **not** tested your patch against my reproducer, so treat that as reasoning rather than a result. Happy to run it if useful.

## Suggestion

The primitive is already in this PR — the async EAGAIN branch could use the same backoff instead of `setImmediate`, or wait for writability on the fd rather than re-attempting blind. That looks like a small addition here, and it would let one PR close both failure modes.

Probably worth carrying to microsoft/node-pty#939 as well, since the async branch is upstream's.

## Reproducer

Self-contained, node-pty only, ~10 s, exit 1 = reproduced: <https://github.com/stablyai/orca/issues/11178#issuecomment-5106222593>

It reproduces the EAGAIN storm at ~144k/s and 102% of one core, with matching stack frames. Three conditions are each necessary, and two are worth knowing for this code:

1. **Raw mode.** A canonical tty *discards* input once its queue is full — 655 MB written, zero EAGAIN. Raw mode returns EAGAIN where canonical drops data, so only TUI sessions are exposed.
2. **A stalled reader.** In all three of my incidents the agent was mid-API-call and not draining stdin.
3. **The retry-without-waiting** above.

`--control` drives the same load through node-pty's `net.Socket` path and stays at 0%, which is the contrast that first pointed me at `CustomWriteStream`.

Environment: macOS 26.3 arm64, Orca 1.4.159, node-pty 1.1.0.

nireak:
Following up on my earlier comment: I've now **tested** a fix for the async-path spin (#11178's failure mode, the one this PR's watchdog wouldn't engage on, since the threadpool stays healthy and completions arrive promptly carrying EAGAIN).

It's the same idea as this PR's `_scheduleSyncRetry` — a 5 ms backoff — applied to the async EAGAIN branch:

```diff
 fs.write(this._fd, task.buffer, task.offset, function (err, written) {
     if (err) {
         if ('code' in err && err.code === 'EAGAIN') {
-            _this._writeImmediate = setImmediate(function () { return _this._processWriteQueue(); });
+            _this._writeImmediate = setTimeout(function () { return _this._processWriteQueue(); }, 5);
```

(plus `dispose()`'s `clearImmediate` → `clearTimeout` to match.)

Validated against the shipped `CustomWriteStream` code verbatim (ported into a stock `node-pty@1.1.0-beta34`), driving `term.write()` against a raw-mode child that stops draining its pty — full method, numbers, and harness in [#11178](https://github.com/stablyai/orca/issues/11178):

|  | baseline (`setImmediate`) | fixed (`setTimeout(…, 5)`) |
|---|---|---|
| CPU | 100–104%, saturated 11/11 s | 2–3%, peak 8% |
| EAGAIN rate | ~140,700/s avg | ~160/s avg |

Baseline matches my production capture (142,900/s). Since this PR already touches the same patch file, folding this in would cover both `CustomWriteStream` failure modes — the wedged-threadpool input loss (this PR) and the healthy-threadpool EAGAIN spin (#11178) — in one change. Happy to test any revision against the #11178 reproducer.

PR #9798, opened 2026-07-21
fix(mobile-quick-commands): replay sheet load killed by connection migration

brennanb2025:
## Problem

Opening the Quick Commands sheet within a few seconds of connecting over relay races the relay→direct cutover. The cutover (`migrateTo` in `stable-logical-rpc-client.ts`) rejects the sheet's in-flight one-shot `settings.getTerminalQuickCommands` with `LogicalClientCutoverError` while `connState` stays `'connected'`, so nothing re-triggers the fetch. The sheet strands on a red **"RPC interrupted by connection migration"** error with a misleading empty "No quick commands yet" list until closed and reopened.

Hit live on 2026-07-21 (screenshot on file). Third surface of the same cutover-kill mechanism as #9794 (capability probe → button vanishes) and #9796 (terminal create → "Couldn't run …" toast); neither of those touches this fetch.

## Fix

- `use-quick-commands.ts`: wrap the load in `loadQuickCommandsWithCutoverRetry` — the read is side-effect-free, so it replays immediately on cutover (the replacement session is already authenticated), capped at 5 replays, and aborts if the sheet closes / the operation is superseded / the client is replaced. Mirrors the on-main pattern in `worktree-create-capability.ts`.
- `stable-logical-rpc-client.ts`: export `isLogicalClientCutoverError`. **Byte-identical to the export added in #9794 and #9796** — whichever lands second/third gets a trivial same-code conflict; keep either side.

Non-cutover failures (timeouts, `ok:false`) still surface immediately — unchanged behavior, kept out of scope.

## Verification

- 4 new tests: replay-then-succeed, exhaustion after 6 attempts surfaces the error, non-cutover errors not replayed, no replay after the sheet closes.
- Mobile `src/session` + `src/transport` suites: 134 files, 867 passed / 2 skipped.
- `tsc --noEmit`, `oxlint`, `oxfmt --check` all clean.

### DIFF
--- a/mobile/src/session/use-quick-commands.test.ts
+++ b/mobile/src/session/use-quick-commands.test.ts
@@ -3,6 +3,7 @@ import { act, create, type ReactTestRenderer } from 'react-test-renderer'
 import { afterEach, beforeEach, describe, expect, it, vi, type MockInstance } from 'vitest'
 import type { TerminalQuickCommand } from '../../../src/shared/types'
 import type { RpcClient } from '../transport/rpc-client'
+import { LogicalClientCutoverError } from '../transport/stable-logical-rpc-client'
 import type { RpcResponse } from '../transport/types'
 import { useQuickCommands } from './use-quick-commands'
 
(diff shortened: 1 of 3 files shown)


brennanb2025:
Manual QA after merging current `main`:

- Launched the Electron dev build from this exact worktree and verified its runtime identity over CDP.
- Added a real global `PR 9798 QA` command through Desktop Settings → Quick Commands.
- Loaded this branch in Orca Mobile on an iPhone 17 Pro simulator (iOS 26.5), paired it to that dev runtime over LAN, opened the live `main` terminal, and tapped the real Quick Commands control.
- The sheet loaded the saved command with no red migration error or stranded empty state. Metro reported a successful bundle and no redbox.
- Exact relay→direct cutover replay is covered deterministically by the hook and stable-client tests; relay was unavailable in this dev build, so the manual run used direct LAN.

Desktop source state:

Desktop Quick Commands

Mobile result:

Mobile Quick Commands sheet

PR #9802, opened 2026-07-22
fix(native-chat): surface draft launch context in desktop and mobile chat composers

brennanb2025:
## Problem

Starting a worktree from a GitHub issue injects the issue link into the agent TUI only (Claude's argv prefill, or a startup paste for non-prefill agents). The native chat view — desktop and mobile — showed no trace of that context: no bubble, no composer draft. If you worked from chat view, the launch context was effectively invisible, and a mobile chat send could even concatenate with the unsent prefill still sitting on the TUI input line.

## Fix

The honest chat-side representation of an unsent TUI draft is a **composer prefill**, not a message bubble (a bubble would claim delivery of context that the chat send path's input pre-clear wipes).

**Desktop**
- Draft launches seed an in-memory `nativeChatLaunchDraftByTabId` entry (direct work-item launches, background GitHub work-item creates, the quick-create composer, and new-tab draft deliveries via `deliverLaunchPromptToAgentTab`).
- **Single-line drafts only.** The chat send pre-clears the TUI with Ctrl+U, which is kill-to-start-of-*line*, so a multi-line prefill would leave earlier lines to glue onto the next message. The guard lives in `seedNativeChatLaunchDraftForAgentTab` so every entry point gets it. In practice this covers the reported flow (a bare issue URL) and deliberately skips Linear launches (`Linked Linear issue: …\n<url>`) and any GitHub launch with a note typed alongside it — those stay TUI-only rather than seeding something a send would mangle. See follow-ups.
- The chat composer adopts the seed once as its editable draft; if the composer already holds text it declines permanently (never resurrects after a manual clear).
- Reconciliation: the TUI input is one line, so any user turn landing in the transcript means the prefill was submitted (possibly edited) or deliberately cleared — the seed is dropped, and the composer copy too if still untouched. A chat send also retires the seed, since the send path already Ctrl+U-clears the TUI line.
- On the backend-spawn path activation reports no `primaryTabId`; completion now resolves the launch tab from the synced store tabs (this also fixes session-option seeding silently skipping there).

**Mobile**
- The host publishes the draft as an optional `launchDraft` field on the mobile terminal tab snapshot — additive optional field, no protocol bump; old apps ignore it, old hosts never emit it. It rides the renderer → `syncWindowGraph` → main → `session.tabs` pipeline (sync-key + change-projection updated so a seed/clear alone republishes).
- The mobile composer adopts it with the same once-only / decline / resolve-on-user-turn semantics, and also drops the copy when the host stops publishing it.
- Mobile chat sends now prepend the same input-line pre-clear byte desktop uses (`\x15`), so a pending TUI prefill can never concatenate with a mobile-sent message. Safe because the host writes send bytes verbatim.

## QA (live)

**Desktop (dev build, CDP-driven, real Tasks-page "Start workspace from issue" flow):**
- Claude launches with the issue link prefilled in the TUI; switching to chat view shows the same link prefilled in the composer:

- Sending from chat delivers exactly one user turn (the pre-clear killed the TUI copy), and the seed retires:

**Mobile (iOS simulator dev client paired to the branch dev desktop):**
- The host-seeded launch draft propagated live over the tab snapshot and prefilled the mobile chat composer:

- Sending from mobile was accepted, cleared the composer, and the agent's received prompt was exactly the sent text — a message previously queued on the TUI input line was cleared by the new pre-clear, no concatenation.
- Negative check: a workspace whose draft was already consumed shows an empty mobile composer.

**Automated:** 4,560 desktop + 2,177 mobile tests pass; typecheck (node/web/mobile), oxlint, oxfmt, and the max-lines ratchet are clean. New coverage: seeding at each entry point, composer adoption state machine (adopt/decline/resolve), transcript resolution rule, sync-key invalidation, snapshot emit, mobile prefill/decline/clear, and the mobile pre-clear byte.

~~Not verified live: the TUI-submit → composer-clear reconciliation~~ **Update:** now verified live too — see the follow-up comment (the earlier blocker was the QA rig's inherited nested-session marker disabling Claude transcript saving, not a product path).

## Notes / follow-ups

- Folder-workspace composer draft launches and the headless/remote-server persisted-tab rebuild don't carry the launch draft yet (same pre-existing invisibility there).
- Multi-line launch drafts (Linear items, GitHub issue + note) stay invisible in chat by design, per the single-line rule above. Covering them needs a pre-clear that kills a whole multi-line buffer rather than one line — deliberately out of scope here.
- Session-options seeding rides along with this change: it now resolves the agent's tab through `resolveLaunchAgentTabId` instead of `primaryTabId ?? startupTerminal`, and defers on runtime-owned worktrees where it was previously dropped. Behaviour in the ambiguous case (several tabs mirror at once, none stamped with `launchAgent`) matches the pre-PR no-seed outcome.
- Mobile: if the host retracts a draft while the tab is off-chat, the adopted copy is retired on return rather than immediately, so a consumed prefill can stay visible for the duration of the transcript read. Fixing it properly needs the controller to distinguish "tab present, no draft" from "tab absent"; the current behaviour is strictly better than losing the prefill on any peek at the terminal.
- `decideInitialAgentTabViewMode` still keeps paste-based draft launches in terminal view; that guard can be lifted once seeding covers every flow.
- Pre-existing on this branch's base: a switch-exhaustiveness lint error in `src/renderer/src/components/skills/skill-freshness-group.tsx` (untouched by this PR).


brennanb2025:
Follow-up: the one previously-unverified scenario — **TUI-side submit → chat composer reconciliation** — is now verified live as well.

The earlier attempts failed for a rig-specific reason: the QA dev app inherited this automation session's `CLAUDE_CODE_CHILD_SESSION` marker, so Claude launched with transcript saving off and the chat view had no transcript to reconcile against (never a product path — real launches persist transcripts). Re-ran with `CLAUDE_CODE_FORCE_SESSION_PERSISTENCE=1` in the agent launch env and the agent on Opus 4.8:

- Started a workspace from a fresh GitHub issue; the launch draft seeded and the issue link sat prefilled in the TUI.
- Pressed Enter in the terminal (TUI-side submit). After the user turn landed in the transcript, opening chat view showed the message as a normal transcript bubble, the **composer empty**, and the seed store entry cleared — no stale prefill:



Every scenario in the state machine is now live-verified: seed → adopt → chat-send (pre-clear, single turn, seed retired), seed → TUI-submit → resolve (composer stays/goes empty), mobile prefill → send (exact text delivered, no concatenation), and consumed-seed workspaces showing empty composers.

brennanb2025:
## Re-QA at `b94d0abedc` — desktop 4/4, mobile 5/5 PASS

The QA in the description was taken at `d205618247`; **11 review-fix commits** have landed since (2 HIGH code findings, 4 CI failures, 1 HIGH regression introduced and fixed during the fix loop). Everything below was re-run against the current tip. CI is 24/24 green.

### Desktop — real Tasks-page "Start workspace from issue" (`stablyai/demo-project#17`)

**1. Headline regression check — PASS.** The guard that rejects multi-line drafts moved from the caller into the shared `seedNativeChatLaunchDraftForAgentTab`, so the priority check was that the reported bug is still fixed. It is: the issue URL is seeded on the agent's own tab and prefills the chat composer.



**2. Send from chat — PASS.** The agent's received prompt was exactly `https://github.com/stablyai/demo-project/issues/17` — one user turn, no concatenation with the TUI copy, seed retired, composer emptied.

**3. Multi-line narrowing — PASS (new deliberate behaviour).** A typed note + linked issue produces a genuinely multi-line TUI prefill, and the chat composer now stays **empty** rather than adopting it. The screenshot shows why: the chat send's `\x15` is kill-to-start-of-*line*, so it would clear only the URL and leave `Please investigate this carefully` behind to glue onto the next message.



**4. TUI-submit reconciliation — PASS.** This was the one scenario *blocked* in the original round. Unblocked with `agentDefaultEnv: { claude: { CLAUDE_CODE_FORCE_SESSION_PERSISTENCE: '1' } }` (a nested-QA-rig artifact, not a product path). Adopt in chat → submit in the TUI → the seed retires and the untouched composer copy clears itself, leaving exactly one user turn.

### Mobile — iOS simulator paired to the branch dev desktop

**1. Launch-draft-only snapshot frame — PASS, both directions.** Highest-value check: with the phone already sitting on the tab's chat view and the agent idle, seeding from the desktop changed *only* `launchDraft` — and it reached the composer in seconds. Previously `mobileSessionTabEqual` did not compare the field, so a draft-only frame was discarded; the original QA passed only because `agentStatus` happened to change in the same publish. Clearing from the desktop dropped the untouched copy.



**2. Leaving chat context and returning — PASS.** The prefill survives; before the fix this was read as a host retraction and dropped permanently.

**3. No empty-state flash — PASS.** Returning to a populated conversation renders the transcript, not "Start a chat with Claude" over it. (This regression was introduced *by* the CI-fix round and caught in verification — commit `b94d0abedc`.)

**4. Mobile send concatenation — PASS.** With the same URL parked unsent on the TUI, the agent received exactly one copy.

**5. Image + text send — PASS.** The strongest evidence for the silent-data-loss case: the agent's input line holds `[Image #1] IMG-SEND-CHECK-9802` — the image survived alongside the text rather than being wiped by a second `\x15` while the phone still rendered a thumbnail.



### What this does NOT cover

- **Scenario 2 is verified by proxy.** The simulator's 400ms long-press would not register through the emulator gesture API, so the "Switch to terminal view" action sheet was unreachable. I used tab-switching, which flips the same `chatActive` gate — the literal view toggle is untested.
- iOS Simulator only (no Android); local host only (no SSH/remote host, no folder workspace); the reconnect path (client swap) is unit-tested only.
- Worth knowing for future QA: seeding onto a tab whose transcript already has a user turn is **declined by design**, so an empty composer there is correct, not a propagation failure.

PR #9816, opened 2026-07-22
fix(dashboard): align terminal-dialog close X with compact header

brennanb2025:
## Problem

In the agent dashboard popout, the terminal dialog's close X is misaligned: it floats below the header row instead of sitting on the title line.

The default `DialogContent` close button is absolutely positioned at `top-4 right-4`, which is tuned for the standard `p-6` dialog. This dialog uses `p-0` with a compact `py-2` header (#9686 switched it from a footer Close button to the corner X), so the default position lands the X below the header's centerline.

## Fix

Suppress the built-in close button (`showCloseButton={false}`) and render the `DialogClose` X inside the header row with `ml-auto`, so it vertically centers with the title and shares the header's horizontal padding. Same visual treatment as the default X (size-4 icon, 70% → 100% opacity on hover, sr-only label).

Adds the `dashboardPopout.terminal.close` locale key (English text in non-en locales, matching the existing keys in that section).

## QA (live Electron, CDP)

Verified in the dev app via the popout QA rig (synthetic snapshot → open dialog):
- X center is exactly on the title's centerline (both measured at the same y), inset to match the header padding.
- Clicking the X still closes the dialog.

dialog X aligned

### DIFF
--- a/src/renderer/src/components/dashboard-popout/AgentTerminalDialog.tsx
+++ b/src/renderer/src/components/dashboard-popout/AgentTerminalDialog.tsx
@@ -1,5 +1,5 @@
 import { useCallback } from 'react'
-import { SquareArrowOutUpRight } from 'lucide-react'
+import { SquareArrowOutUpRight, XIcon } from 'lucide-react'
 import { AgentIcon } from '@/lib/agent-catalog'
 import { agentTypeToIconAgent, formatAgentTypeLabel } from '@/lib/agent-status'
 import { agentStateLabel } from '@/components/AgentStateDot'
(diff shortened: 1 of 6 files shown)


brennanb2025:
Live-terminal QA (requested): same rig but with a real PTY attached to the dialog (`pty.spawn` → snapshot `ptyId`), so the `AgentTerminalPreview` is mounted and streaming.

Measured in the popout: the X's vertical center sits exactly on the title centerline, and its right edge aligns with the footer button's gutter. Clicking the X closes the dialog.

Note for reviewers: an earlier re-test appeared to still show the misaligned X — that was a stale QA window from a previous dev instance (its vite port had been reclaimed by another worktree's server); the fresh instance below runs this branch.

dialog X aligned, live terminal

Issue #9429, opened 2026-07-19
[Bug]: orca-ide reports an invalid_relative_path error when opening files.


aliuq:
### Operating system

Windows

### Orca version

1.4.145

### Details

- WSL

```bash
> orca-ide file open /root/orca/workspaces/xxx/xxx/xxx.ts

invalid_relative_path
```

nwparker:
**Reproduced on current tree (`main`-based).**

`orca file open <absolute-path>` fails with `invalid_relative_path` even when the file is inside the selected worktree.

**Root cause (two ends):**
- CLI forwards the raw path verbatim: `src/cli/handlers/file.ts:140-145` — `file open` does `getRequiredStringFlag(ctx.flags,'path')` and passes it straight through as `relativePath`, with no relativization against the resolved worktree root.
- Runtime rejects it: `src/main/runtime/orca-runtime-files.ts:502-503` (`openMobileFile`) calls `isSafeMobileRelativePath` (defined `:1934`), which returns `false` for any path starting with `/` or `<drive>:[\\/]`, so it `throw new Error('invalid_relative_path')`.

**Re-run:**
```
pnpm exec vitest run --config config/vitest.config.ts src/cli/handlers/repro-9429-file-open-absolute-path.test.ts
```

**Key assertions + observed values** (worktree root `/tmp/repo`, open `/tmp/repo/src/App.tsx`):
```
files.open forwarded relativePath === '/tmp/repo/src/App.tsx'   // raw absolute, NOT 'src/App.tsx'
isSafeMobileRelativePath('/tmp/repo/src/App.tsx')        === false  // → invalid_relative_path
isSafeMobileRelativePath('src/App.tsx')                  === true
isSafeMobileRelativePath('/root/orca/workspaces/xxx/xxx/xxx.ts') === false  // issue's WSL path
```

**A fix would** relativize an in-worktree absolute path in the `file open`/`file diff` handlers (strip the worktree root prefix before sending `relativePath`), so an absolute path pointing inside the worktree is accepted instead of rejected.

nwparker:
Fixed by https://github.com/stablyai/orca/pull/9824 — `orca file open` / `file diff` now accept absolute paths inside the worktree instead of failing with `invalid_relative_path`.

Reopen if you still hit this after updating.

aliuq:
@nwparker Hi, the problem persists. My environment is `WSL`, and the workspaces location is `/root/orca/workspaces`, but the print output still uses the Windows path.

[...]

PR #9824, opened 2026-07-22
fix(cli): relativize absolute --path for file open and file diff before the runtime RPC (#9429)

rodboev:
# fix(cli): accept absolute paths inside the worktree for `file open` / `file diff`

## Summary
`orca file open` and `orca file diff` now accept an absolute path that lives inside the selected worktree, instead of rejecting it with `invalid_relative_path`.

## ELI5
Before, if you gave Orca the full path to a file, it refused to open it — even though the file was right there in your project. Now it just opens it.

## Root cause & fix
The CLI handlers forwarded the user-supplied `--path` straight through as `relativePath`, so the runtime guard rejected any absolute path — even one inside the worktree. The fix adds `resolveFilePath()`, which fetches the worktree root via `worktree.show` and relativizes an absolute path that falls inside it. Paths outside the worktree are forwarded unchanged so the runtime guard still rejects them, and the worktree root itself is rejected with a clear `invalid_argument` message. Relative paths keep the existing single-RPC path with no extra `worktree.show` call.

## Evidence
Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase was clean onto `origin/main` (base `8a23618310`, head `cdd6f50696`).

- Test: `src/cli/handlers/file-absolute-paths.test.ts` — covers relativization, outside-worktree pass-through, and worktree-root rejection.
- On branch: **5 passed (5)**.
- With the fix reverted (test kept): **4 failed, 1 passed** — confirming the test captures the bug. Key assertions that fail without the fix:

```
  src/cli/handlers/file-absolute-paths.test.ts

On branch:      Test Files 1 passed (1) | Tests 5 passed (5)
Fix reverted:   Tests 4 failed | 1 passed (5)
  - "leaves outside-worktree absolute paths for the runtime guard":
      expected 2nd call files.open but called only once
  - "rejects the worktree root as a file-open target":
      expected process.exitCode 1, received undefined
```

- Lint: `oxlint src/cli/handlers/file.ts src/cli/specs/file.ts` — clean, no findings.

## Trade-offs
None — pure fix.

## Regression risk
Zero-regression by construction: only the absolute-path branch is new. Relative paths still take the single-RPC path with no extra `worktree.show`, and outside-worktree absolute paths still reach the unchanged runtime guard. Proven by the passing test suite on-branch.

*Credit to @rodboev for the original fix design. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/cli/handlers/file-absolute-paths.test.ts
+++ b/src/cli/handlers/file-absolute-paths.test.ts
@@ -1,0 +1,161 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const callMock = vi.fn()
+
+vi.mock('../runtime-client', () => {
+  class RuntimeClient {
+    readonly isRemote = false
+    call = callMock
+    getCliStatus = vi.fn()
+    openOrca = vi.fn()
+  }
+
+  class RuntimeClientError extends Error {
+    readonly code: string
+
+    constructor(code: string, message: string) {
+      super(message)
+      this.code = code
+    }
+  }
+
+  class RuntimeRpcFailureError extends RuntimeClientError {
+    readonly response: unknown
+
+    constructor(response: unknown) {
+      super('runtime_error', 'runtime_error')
+      this.response = response
+    }
+  }
+
+  return { RuntimeClient, RuntimeClientError, RuntimeRpcFailureError }
+})
+
+import { main } from '../index'
+import { buildWorktree, okFixture, queueFixtures, worktreeListFixture } from '../test-fixtures'
+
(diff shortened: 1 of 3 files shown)


nwparker:
Thanks so much for this, @rodboev — relativize absolute --path for file open and file diff before the runtime RPC (#9429). Reviewed, rebased on latest `main`, and merged. Really appreciate the contribution! 🙏

AmethystLiang:
looking into it

PR #9825, opened 2026-07-22
fix(native-chat): buffer composer side effects during IME composition (#9738)

### LINKED ISSUE #9738: [Bug] Korean (IME) input intermittently drops characters in Claude Code session chat
<kbd>[Chris C](https://discord.com/channels/1492228674081263786/1529107552753811576/1529107552753811576)</kbd> [Chris C](https://discord.com/channels/1492228674081263786/1529107552753811576/1529107552753811576)  `via Discord`

When typing Korean in the chat input of a Claude Code session, characters
intermittently fail to register — usually a single syllable gets dropped
mid-sentence, so I have to go back and retype it.

Environment
- Orca desktop (macOS)
- Korean 2-set input (macOS built-in IME)
- Chat input inside a Claude Code session

Steps to reproduce
1. Switch the system input source to Korean.
2. Open a Claude Code session and start typing a normal-length Korean sentence
   in the chat input.
3. Type continuously at normal speed.

Expected
Every keystroke is composed and committed correctly.

Actual
Occasionally one character is silently dropped. It isn't reproducible on every
sentence — it seems to happen more often when typing quickly, which suggests the
IME composition events (compositionstart / compositionupdate / compositionend)
may be racing with the input handler or with re-renders of the chat input.

It's not blocking, but it breaks the typing flow often enough to be annoying.
Latin input has no such issue, so this looks specific to IME-composed text
(likely affects Japanese/Chinese input as well).

rodboev:
## Summary

Fixes the native-chat composer dropping Korean (and other IME) characters because every intermediate composition update ran the full draft/picker/suggestion/cache pipeline before the IME confirmed the text.

## ELI5

When you type in a language like Korean that assembles characters as you go, the chat box was reacting to each half-finished keystroke and losing letters. Now it waits until the character is finished before doing anything.

## Root cause & fix

Each in-progress IME composition change triggered draft persistence and picker/history/caret side effects while the text was still unconfirmed, corrupting or dropping characters. The fix mirrors Orca's existing settings-input composition pattern: while `isComposingRef` is true, per-change updates route through a non-persisting setter (`setDraftWithoutPersist`); on `compositionend` the committed text is persisted once and side effects run. A `skipNextChangeRef` de-dupes the trailing change event some browsers emit after `compositionend`, and a `useLayoutEffect` resets composition refs on pane (`draftScopeKey`) change so state can't leak across panes.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Clean rebase onto `origin/main`.

- Test file `native-chat-composer-composition.test.tsx` (new): 5/5 passed on branch.
- Reverting ONLY the two fix files to `origin/main` (keeping the test) reproduces the bug: 2 failed, 3 passed.
- Lint clean (`oxlint`); PR also reports `typecheck:web` and `oxfmt --check` passing.

```
On branch:        Test Files 1 passed (1); Tests 5 passed (5)
Fix reverted:     Test Files 1 failed (1); Tests 2 failed | 3 passed (5)
  - "defers side effects until composition end": expected setDraftWithoutPersist called 2 times, got 0
  - "flushes confirmed text once on composition end": expected persistDraft called once, got 0
```

## Trade-offs

None — pure fix. It only narrows *when* draft cache persistence occurs during composition.

## Regression risk

Zero-regression by construction: only the active-composition branch changes. Non-composition typing, Enter submit, and Shift+Enter newline paths are untouched and remain covered by dedicated passing tests; the textarea stays controlled. No new IPC, network, or file-path surface. Residual risk is limited to IME edge cases, which the change-before/change-after `compositionend` ordering tests exercise.

---

*Original fix by @rodboev (closes #9738). Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/renderer/src/components/native-chat/NativeChatComposer.tsx
+++ b/src/renderer/src/components/native-chat/NativeChatComposer.tsx
@@ -1,4 +1,12 @@
-import { forwardRef, useCallback, useImperativeHandle, useMemo, useRef, useState } from 'react'
+import {
+  forwardRef,
+  useCallback,
+  useImperativeHandle,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState
+} from 'react'
 import { useAppStore } from '../../store'
 import { sendRuntimePtyInput } from '@/runtime/runtime-terminal-inspection'
 import { getSettingsForAgentTabRuntimeOwner } from '@/lib/agent-paste-draft'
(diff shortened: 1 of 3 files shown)


OrcaWin:
Thanks for working on the IME report. I reproduced and traced #9738 before evaluating this change, and the report is specifically about Korean input inside Claude/Codex TUIs running in Orca terminal panes. That path is xterm → terminal lifecycle → PTY bytes. This PR only changes the optional Native Chat composer; disabling the PR leaves every file and bundle in the reported xterm/PTY path unchanged, so it cannot affect the reported failure.

The local hardening and focused checks were technically clean (134 focused tests plus web typecheck/lint/format), but they still do not establish a Native Chat bug: the tests call mocked composer callbacks rather than exercising real composition/beforeinput/input events or PTY bytes. The reproduce-first gate therefore remains red for this PR even though its tests pass.

I am closing this as a wrong-surface fix for #9738. If Native Chat itself can be reproduced dropping IME text, please open a separate issue with an exact event sequence and we can revisit this implementation against that surface. The xterm/PTY issue is being handled in the terminal IME fix instead.

Issue #9454, opened 2026-07-19
macOS logout/login breaks Orca terminals — daemon restart needed, all tabs require /resume


dragonfire1119:
## Bug Report

Logging out of macOS with Orca open and logging back in leaves Orca terminals broken. Happens daily.

**Behavior:** Terminal sometimes shows a login prompt. New tabs opened in that state have zero egress to Claude (no network reachability from the agent). Restarting the daemon via Settings sometimes fixes it; when it doesn't, killing the daemon process and restarting Orca is required. Either way, once the daemon comes back, every tab has to be reopened manually using `claude /resume` to get sessions back — this step is always required, not just after a kill. Sometimes this flow also forces a fresh Claude login.

**Diagnosis:** Repro: open Orca with terminals/agents running → full macOS logout → log back in → new tabs show zero egress or stale login state → Settings > restart daemon (sometimes fixes it, sometimes doesn't) → if not, kill daemon + restart Orca → either way, all tabs must be reopened with `claude /resume` → occasionally prompted to log into Claude again.

**Expected behavior:** Logging out/in on macOS should not break the daemon or terminal egress. If the daemon does need to restart, existing tabs/sessions should recover automatically without manual `/resume` per tab, and without requiring a fresh Claude login.

brennanb2025:
Thanks for the detailed report. We closed this as a duplicate of #7936, which tracks macOS logout or login leaving terminal and PTY state broken and requiring app or daemon recovery. We’re keeping the investigation there.

dragonfire1119:
> Thanks for the detailed report. We closed this as a duplicate of [#7936](https://github.com/stablyai/orca/issues/7936), which tracks macOS logout or login leaving terminal and PTY state broken and requiring app or daemon recovery. We’re keeping the investigation there.

Makes sense now. My bad for the duplicate.

brennanb2025:
Reopening to close this out properly — this was fixed, not "not planned."

The root cause turned out to be three separate things stacked on top of each other:

- The terminal daemon survived macOS logout carrying a dead login session, so every shell it spawned afterward came up broken ("Login incorrect") with no network. Fixed in #9826 — the daemon now detects its login session died and retires itself, no manual daemon restart.
- A cached login-preflight rejection from before the logout stuck around and kept poisoning new spawns. Fixed in #9973.
[...]

PR #9826, opened 2026-07-22
fix(daemon): retire macOS daemons whose login session died (#7936)

brennanb2025:
## Summary

Field follow-up to #7936 / #9454: after a full macOS logout and login, terminals can restore to `login:` / **Login incorrect**, new tabs can have zero egress, and only a manual daemon restart recovers them. The surviving daemon's PAM/login context and Mach bootstrap resolver are no longer usable, while the login-wrapper cache retains its birth-time accepted verdict.

This PR makes GUI-spawned macOS daemons converge from inside the stale session. It complements #9684's next-startup fencing and covers logout while the app is closed without changing headless `orca serve`, SSH, Linux, or Windows daemon behavior.

## Reliability contract

- **Invariant (`terminal-provider.login-session-retirement`):** a daemon may retire itself only after it previously proved login wrapping worked, then receives three consecutive conclusive PAM rejections and an explicit `unhealthy` resolver verdict. Inconclusive PAM results and `unknown` resolver results preserve all PTYs.
- **Failure source:** #7936 / #9454; a GUI daemon survives loginwindow teardown and later creates login-prompt zombies with no network egress.
- **Oracle:** the cache-bypassing PAM probe must conclusively reject three times; `scutil --dns` inside that same daemon must explicitly report no DNS configuration. A successful PAM verdict resets the rejection streak.
- **Gate:** accepted gap for a real GUI logout, which cannot be fabricated without ending the development login session. Deterministic watch/provider tests and the mocked Electron lifecycle cover the state machine and recovery wiring; the real-oracle transition remains listed below.
- **Diagnostics:** `login-session-watch-armed`, `login-session-probe-rejected`, `login-session-retire-suppressed`, `login-session-probe-error`, and `login-session-dead-retire` identify the trigger, verdict count, resolver result, and retirement.

## Mechanism

- New `MacosLoginSessionDeathWatch`, enabled only by the GUI launcher's `--login-session-watch` flag, re-runs the existing PAM preflight without its cache. Triggers are a trailing-debounced PTY-exit burst, authenticated client activity, and a two-minute periodic backstop.
- An inconclusive pipe probe escalates once through the macOS system `expect(1)` binary to cover tty-sensitive PAM behavior. If both probes time out, the result remains inconclusive: repeated timeouts are not authority to orphan live terminals on slow or offline PAM hosts.
- A conclusive rejection immediately disables the login wrapper, so the daemon stops creating `login:` zombies even before retirement completes.
- Retirement is crash-style: session metadata remains unclean so the replacement daemon cold-restores scrollback; existing dead-endpoint recovery owns socket/pid cleanup and respawn.
- Headless `orca serve` launches do not set the watch flag, preserving the SSH/disconnect lifetime contract. Non-darwin behavior is unchanged.

## Provider and platform coverage

- **macOS local/daemon:** covered by deterministic PAM classification, watch scheduling, authenticated-client/PTY-exit wiring, and mocked live Electron retirement/respawn QA.
- **SSH/headless serve:** unaffected; the launcher omits the watch flag. An SSH-minted dead audit-session experiment continued to accept PAM and retained resolver configuration, consistent with its system-bootstrap lifetime.
- **Linux/Windows/WSL:** unaffected behind the `darwin` runtime check; existing native smoke and terminal restart checks passed on the prior PR head.
- **Mobile/relay:** transport and ownership are unchanged; they observe the same daemon endpoint retirement/recovery contract.

## Performance budget

- No startup await, session inventory, renderer work, or provider fanout was added. Probes run off the PTY spawn/hello response path.
- Baseline polling is one PAM subprocess every two minutes (30/hour after startup). On this Mac, five healthy probes measured 0.16-0.19 seconds wall time each.
- Suppressed retirement now returns to the two-minute cadence instead of running PAM plus `scutil` every 10 seconds indefinitely (30 rather than up to 360 probe cycles/hour in a stable mismatch).
- A sustained PTY-exit burst is trailing-debounced into one probe. Authenticated client probes have a 30-second activity gap, and all non-startup probes share a five-second minimum gap.
- Probe output is capped at 1 KiB; each child is bounded to four seconds; timers are unref'd and canceled by `stop()`.
- The PTY fallback is a short-lived system process rather than a daemon-owned native PTY; 50 sequential live probes held the daemon process at 47 file descriptors before and after.

## Testing

- [x] `pnpm typecheck`
- [x] 7 touched-module suites: 184 tests passed
- [x] Focused PAM/watch/entry suites: 50 tests passed
- [x] Oxlint on touched files
- [x] `oxfmt --check` and `git diff --check`
- [x] Mocked Electron lifecycle: accepted baseline -> three rejected verdicts plus unhealthy resolver -> daemon exit -> dead-endpoint respawn -> working restored/new terminal
- [x] Live healthy PAM timing: five runs at 0.16-0.19 seconds; `scutil --dns` at 0.01-0.03 seconds
- [x] Live PTY fallback: accepted in 0.10 seconds; 50 sequential probes completed without file-descriptor growth (47 → 47)

Repository-wide `pnpm lint` reaches an unrelated pre-existing exhaustiveness error in `src/renderer/src/components/skills/skill-freshness-group.tsx`. A full daemon run reached 1,004 passing tests but the unrelated deleted-cwd test timed out under this workspace's unsupported Node 26; a minimal Node 26 reproduction hangs on deleting its own current directory, before any PR code is involved.

## Screenshots

Mocked lifecycle screenshots and daemon logs are attached in the PR comment titled **Live Electron lifecycle QA — mocked dead session**.

## AI Review Report

Deep review covered false-retirement authority, timer/subprocess bounds, event-storm behavior, cache updates, authenticated-client wiring, shutdown/restore semantics, cross-platform and SSH exclusions, and diagnostics. The review removed timeout-only retirement, made resolver `unknown` fail safe, backed suppressed verdicts off to normal cadence, and changed PTY exit handling to a true trailing debounce, and stops the watch before graceful shutdown so in-flight probes cannot interrupt checkpoints.

## Security Audit

No network-facing protocol or authentication change. Client activity is observed only after the existing daemon token and protocol checks pass. The E2E verdict-file seam is inert unless explicitly provided in the daemon environment and does not expose terminal data.

## Residual gaps / Notes

- A destructive real GUI logout was not performed on the development account. The production PAM/resolver transition is inferred from the field's `Login incorrect` plus zero-egress symptoms and the existing resolver-health implementation.
- If a real dead GUI session makes both pipe and PTY PAM probes time out forever, this watch preserves the daemon rather than risking false retirement; #9684's next-startup identity fence remains the recovery path.
- Daemons created before this build do not gain the launch flag until replaced; #9684 covers their next-login startup.
- Between self-retirement and the next reply-bearing adapter operation, writes to an already-open pane can be dropped on the dead socket. Full logout normally closes the app, so startup recovery is the field path; proactive mid-use respawn remains follow-up scope.


brennanb2025:
## Live Electron lifecycle QA — mocked dead session (reproduced + healed)

A real dead login session needs root to fabricate, so QA drives the two oracles through the new `ORCA_E2E_LOGIN_SESSION_PROBE_FILE` seam (second commit) and exercises everything else for real: dev build from this branch, throwaway userData profile, real daemon with `--login-session-watch`, real PTYs, driven via CDP.

**Baseline** — terminal on the watched daemon, working; watch armed at startup (`login-session-watch-armed`):

baseline

**Mocked logout** — verdict file flipped to `dead` at 20:28:22; daemon log:

```
03:28:24.218 login-session-probe-rejected  rejections=1  (trigger: periodic)
03:28:24.721 login-session-probe-rejected  rejections=2
03:28:25.222 login-session-probe-rejected  rejections=3
03:28:25.222 login-session-dead-retire     rejections=3 resolverHealth=unhealthy
```

Daemon pid 68634 exited 3s after the flip. No manual restart.

**Recovery** — next `createOrAttach` (New Terminal) triggered the adapter's dead-endpoint respawn: fresh daemon (pid 26160) forked, watch re-armed, new terminal fully functional:

recovered

**App restart on the same profile** — tabs restore against the healthy daemon and the shell works:

after-relaunch

**Also observed, worth noting:** between retirement and the next request/reply RPC, keystrokes into an existing pane are silently dropped (writes are fire-and-forget on the dead socket; only `createOrAttach`/reply-bearing RPCs trigger the respawn). In the field the retirement almost always happens while the app is closed (logout), so this window is theoretical — but if we want mid-use retirement to heal instantly, a follow-up could have the adapter treat a dead socket on write as a respawn trigger too.

**Unverified (needs a real dead session):** that PAM's `login -flpq` conclusively rejects and `scutil --dns` degrades in an actual post-logout session — i.e. the real-oracle behavior the mock stands in for. A root-created disposable login session experiment is staged for that, and the reporter's `daemon.log` would confirm it from the field.

brennanb2025:
## De-risking round: the trigger can no longer be defeated by any known dead-session behavior

The residual risk was a single failure shape: `login(1)` hanging at the prompt past the probe bound in a dead session → probe killed → inconclusive forever → watch never fires. Commit `54e9bad9f2` closes it three ways:

1. **Dedicated 4s probe bound** for the death watch (the 500ms preflight bound exists for spawn-path latency, which doesn't apply off-path) — a slow-but-answering PAM stack is no longer misread as a hang.
2. **PTY escalation**: an inconclusive pipe probe re-runs under `script(1)`'s PTY, covering the tty-sensitive-PAM divergence the preflight's own comments warn about.
3. **Hang-shaped second trigger**: a streak of 5 timeout-killed probes (a live session answers in milliseconds, even under load) retires the daemon, still gated on the degraded resolver, logged with a distinct `cause: probe-timeouts` so field logs discriminate the paths.

Coverage matrix now: fast conclusive reject → trigger 1; prompt-then-EOF (zero exit, no marker) → conclusive reject → trigger 1; hang under pipe but verdict under PTY → trigger 1 via escalation; hang under both → trigger 2. All inconclusive/ambiguous states still fail toward preserving the daemon, and both triggers require the armed baseline + degraded resolver.

**Live QA of the new trigger** (same Electron rig, `hang` verdict): 5 × `login-session-probe-timeout` at 500ms intervals → `login-session-dead-retire cause=probe-timeouts resolverHealth=unhealthy` → daemon exited 4s after the flip.

Deterministic coverage: 10 → 15 watch tests (timeout-streak retire, unarmed timeouts never count, conclusive verdicts reset the streak, resolver-healthy suppression with later re-verdict), 3 new probe tests (PTY escalation accept/reject, dual-timeout stays inconclusive). Full daemon suite: 1,004 passed.

brennanb2025:
## Real dead-session experiment (SSH-minted): a decisive negative with three useful conclusions

Ran the staged experiment against a genuinely torn-down session: SSH to localhost minted a fresh audit session (asid 103920 vs the GUI's 100024), an orphaned probe survived its teardown (reparented to launchd), and for minutes afterward **both PAM probes kept conclusively accepting and `scutil --dns` stayed configured**.

1. **No false-retirement risk from non-GUI session death.** An orphan of a dead SSH session keeps a working PAM/DNS context — so even if the watch were ever armed near SSH-shaped lifecycles (it's excluded by design), the oracles wouldn't fire. This is protective evidence for the SSH persistence use case.
2. **`launchctl print pid/<orphan>` keeps reporting the dead session's original asid** — relevant to #9684: a post-logout daemon will report its stale asid to the process-identity guard and be correctly fenced (noting there).
3. **The GUI logout is the special case.** What breaks a surviving daemon isn't audit-session death per se — it's loginwindow destroying the per-session Mach bootstrap namespace, which GUI sessions have and SSH sessions don't (they ride the system bootstrap; that's exactly why Orca's remote daemons survive disconnects). This sharpens the fix's model: the oracles' production-observed failures ("Login incorrect" in the reporter's terminals; the pre-existing resolver-health machinery built for dead-session daemons) are specific to the GUI-teardown state, which cannot be fabricated without an actual logout.

**Remaining validation is now passive:** a 48h slow probe logger is parked in the GUI session on this dev machine; the next ordinary logout/login captures the real oracle transition (or proves orphans don't survive logout on this config — also decisive). Reporter's `daemon.log` remains the equivalent field artifact.

brennanb2025:
Review follow-up pushed in `c41fb06363`:

- Fixed the graceful-shutdown race by stopping the login-session watch before SIGTERM/SIGINT checkpoint teardown and idle retirement. Added a deterministic test showing an in-flight resolver result cannot retire after `stop()`.
- Intentionally did not restore timeout-streak retirement. The probe contract marks timeouts inconclusive, and slow/offline enterprise PAM can exceed the bound while DNS is unavailable; promoting that combination to `process.exit(1)` could orphan healthy PTYs. Three conclusive PAM rejections plus explicit resolver `unhealthy` still retire for the reported `Login incorrect` path; dual-timeout sessions fail safe and retain #9684 as the next-startup recovery path.

brennanb2025:
## Deep review follow-up

Pushed `657cdcfc1e` and merged current `main` in `4ecce8ef47` after auditing the detector, spawn-wrapper cache, scheduler, subprocess cleanup, cold-restore path, GUI/headless launch boundary, and the latest terminal-generation refactor.

Material fixes:

- **Recovery trigger retention:** client hellos and PTY-exit signals that arrive inside the probe gap or during an armed probe now retain exactly one bounded follow-up. Previously they could be dropped, leaving a post-login stale daemon undiscovered until the two-minute periodic backstop.
- **No stale cache overwrite or duplicate startup child:** the death watch reuses the in-flight startup PAM warmup and fences overlapping spawn-path results, so an older acceptance cannot re-enable `login(1)` after a newer death rejection. Deterministic coverage also proves startup uses one PAM child rather than two.
- **Lifecycle cleanup:** watcher stop now aborts in-flight pipe/PTY probes and the corroborating `scutil` child, removes listeners, and prevents PTY fallback after cancellation. This closes the orphan-process window when graceful shutdown exits before a probe timeout.
- **Production wiring:** tests now prove the GUI launch option forwards `--login-session-watch`, the default/headless path omits it, and reaped PTYs reach the watch callback after the current-main terminal host refactor.

Verification on the merged branch:

- `pnpm typecheck` — pass
- 16 focused daemon/provider/network suites — **226 passed**
- full `src/main/daemon` suite — **994 passed**, with only the two known deleted-cwd timeouts under unsupported Node 26 (same pre-existing Node/runtime issue, before PR code)
- touched-file oxlint + oxfmt — pass
- max-lines ratchet — pass, no new bypasses
- reliability-gate manifest check — pass
- repository-wide `pnpm lint` reaches the same unrelated pre-existing switch-exhaustiveness error in `skill-freshness-group.tsx`

The retirement oracle remains conservative: only three conclusive PAM rejections plus explicit resolver `unhealthy` can retire an armed daemon; timeouts and resolver `unknown` still preserve PTYs.

brennanb2025:
## Performance review follow-up

Pushed `78cdaf1e64` after tracing every login-watch trigger back to its callers.

**Finding:** the client-activity hook ran for every accepted hello, including control-only `health-check` / `resolver-health-check` sockets. The resolver check runs before fresh macOS PTYs, so ordinary new-terminal creation could consume the 30-second activity budget and launch an unnecessary PAM subprocess. A full app connection also invoked the hook twice (control + stream).

**Fix:** login-session activity now fires only when a token-authenticated control + stream pair is complete. That preserves the post-login app reconnect signal while excluding one-shot internal health traffic.

**Deterministic count:** the new protocol test fails on the previous implementation (control-only health probe: 1 callback; full app pair: 2) and now proves **0** and **1**, respectively.

Verification on `78cdaf1e64`:

- `pnpm typecheck` — pass
- 19 focused daemon/provider/network/history suites — **424 passed**
- touched-file oxlint and oxfmt — pass
- reliability-gate manifest, max-lines ratchet, and diff checks — pass
- live PAM pipe probe — 0.04–0.06s across five runs
- exact Expect-owned PTY fallback — 100/100 accepted, zero failures

This removes routine health-probe subprocess churn without changing the retirement oracle, periodic backstop, PTY-exit trigger, GUI/headless boundary, or cold-restore behavior.

PR #9830, opened 2026-07-22
fix(sidebar): preserve project grouping for nested SSH hosts

Noobear:
## Summary

Keeps a project grouped under a single sidebar header across saved Orca runtimes instead of fragmenting into per-path groups when the same project is reachable both locally on a runtime and through a nested SSH host.

## ELI5

When you open one project through a remote machine that also reaches it over SSH, the sidebar used to split it into several confusing entries named after file paths. Now it stays as one tidy project.

## Root cause & fix

The all-host projection routes repositories fetched from a remote Orca runtime through that runtime's host ID, so a runtime's local checkout and its nested SSH checkout can share the same projected host ID. `getProjectSetupSurfaceKey` discriminated the nested-SSH surface only by `connectionId`, but `setupWithFetchedOwner` forces `connectionId: null` for RPC-fetched runtime setups — so both checkouts collapsed to `${projectId}::${hostId}::default`, looked like duplicates, and split the project into path-named groups. The fix adds an `executionHostId` fallback discriminator: when a setup's execution host differs from its projected host, it gets a distinct `::exec:` surface, so the nested-SSH and runtime-local checkouts no longer collide while real same-target duplicates still split correctly.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Clean rebase (base `8a23618310`, head `29280dbfb2`).

- New regression test `keeps runtime-local and nested SSH checkouts grouped when fetched over runtime RPC` builds setups the way `setupWithFetchedOwner` does (`connectionId: null`, `executionHostId: 'ssh:intel-mac'`) and asserts both worktrees resolve to one project header.
- Verified this test FAILS without the source fix (stash-out of the `.ts` change → 1 failed) and PASSES with it.
- Full suite green including the PR's own tests plus the new case:

```
  src/renderer/src/components/sidebar/worktree-list-groups.test.ts
Test Files  1 passed
Tests       112 passed
```

- `npx tsc --noEmit -p config/tsconfig.tc.web.json` → exit 0 (0 errors)
- `npx oxlint` + `npx oxfmt --check` on both changed files → exit 0 (no `max-lines` disable added)

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression by construction: the `executionHostId !== hostId` guard means plain local, SSH, WSL, and Windows setups (where the execution host equals the projected host or is absent) take the unchanged `::default` path and are byte-identical in behavior. Only the previously-colliding nested-SSH-over-RPC branch changes, and that change is proven correct by the new test and the existing 111 passing cases.

_Original fix by @Noobear. During automated bug-bash review the fix was hardened: the original discriminator relied solely on `connectionId`, which is nulled on the RPC-fetched runtime path and left the bug live; an `executionHostId` fallback discriminator and a matching regression test were added. Validated, rebased, and hardened via automated bug-bash review; original fix design preserved._


nwparker:
@Jinwoo-H what is expected behavior?

Noobear:
@nwparker Answering the expected-behavior question, and reporting that this branch is now verified against a real multi-runtime setup — the environment #10031 was `env_blocked` on.

## Expected behavior

For one project reachable through several hosts:

- **one project header**, with each host setup nested under it as a child row (`imac` / `air` / `intelmac` / `desktop`);
- path-tail titles (`giloh/orca/projects/…`, `tmshd/Desktop/orca/…`) appear **only** for genuinely ambiguous surfaces;
- two *different* checkouts on the *same* execution surface still split — that case is covered by the `still splits duplicate checkouts on one nested SSH target` test.

## What the verification found

Running this branch as-is on a live graph (1 local host + a paired HUB runtime carrying 4 nested SSH hosts) left the sidebar **unchanged**. The surface-key logic in `worktree-list-groups.ts` was correct; its input was polluted.

`__store.getState().projectHostSetups` held **55 rows for 31 unique `id`s**. `getProjectHostSetupOwnerKey` keys on `[hostId, executionHostId ?? hostId, runtimeOwnerEnvironmentId ?? null, repoId || id]`, and `setupWithFetchedOwner` rewrites three of those four — so in `mergeProjectHostSetupCompatibility` a derived row can never match its own fetched copy, and both survive.

The nested-SSH case this branch targets worked. The HUB's own **runtime-local** checkouts did not: both copies have `executionHostId === hostId` and no `connectionId`, so both collapse to `::default` → count 2 → the project lands in `projectIdsRequiringSetupGroups`. Exactly 7 surfaces came out ambiguous, all HUB-local paths, and those were exactly the 7 projects still rendering path-tail titles.

## Fix — 9fa7826 (pushed to this branch)

`src/renderer/src/store/slices/repos.ts` — dedupe by `setup.id`, which is the stable identity across the `setupWithFetchedOwner` rewrite.

| | rows | unique ids | ambiguous surfaces |
|---|---|---|---|
| before | 55 | 31 | 7 |
| after | 31 | 31 | 0 |

Rendered sidebar after reload: leaf labels containing a home-directory path prefix **0** (was one per duplicated checkout); each project name renders exactly once. One card per project, hosts nested.

`vitest run --config config/vitest.config.ts src/renderer/src/store/slices/repos.test.ts src/renderer/src/components/sidebar/worktree-list-groups.test.ts` → **135 passed**.

## Notes

- Full repro and verification write-ups, with the per-surface data: [repro](https://github.com/stablyai/orca/issues/10031#issuecomment-5090085332) · [fix verified](https://github.com/stablyai/orca/issues/10031#issuecomment-5090111226). #10031 can come off `env_blocked`.
- The dedupe is upstream of grouping — every other consumer of `projectHostSetups` was counting the same checkout twice, which is the likely source of Symptom 2 (same project opened twice) in #10031. Not separately verified yet.
- I have the multi-device rig standing by, so I can re-run any check you want on it.

Noobear:
**Before** — live store: 55 rows / 31 unique ids / **7 ambiguous surfaces**. Path-tail titles on every HUB-local project.

before - no filter

Noobear:
**After** (9fa7826) — live store: 31 rows / 31 unique ids / **0 ambiguous surfaces**. One header per project, hosts nested as children.

after - fixed

Noobear:
Merged `main` into the branch (no rebase — `9993d2b` is preserved). Conflicts resolved, `MERGEABLE` again.

One thing worth flagging from the merge, since it's the first question a reviewer will have: **`main` now touches the same spot.** The owner-key dedupe in `mergeProjectHostSetupCompatibility` was replaced with

```ts
function getRepoDerivedSetupKey(setup: ProjectHostSetup): string {
  return JSON.stringify([setup.hostId, setup.repoId || setup.id])
}
```

That drops `executionHostId` and `runtimeOwnerEnvironmentId` from the key — two of the three fields `setupWithFetchedOwner` rewrites — but keeps `hostId`, which is the third:

```ts
const hostId = getRuntimeTargetHostId(target)
if (target.kind !== 'environment') return setup
return { ...setup, hostId, /* … */ runtimeOwnerEnvironmentId: target.environmentId, connectionId: null }
```

So for a **paired environment** the fetched copy carries the runtime's `hostId` while the derived row still carries the original, and the two keys never meet — the duplicate rows survive on `main` as they did before. The resolution therefore keeps `getRepoDerivedSetupKey` as the primary key and layers the setup-`id` guard on top, since `id` is the one component stable across that rewrite.

`vitest run --config config/vitest.config.ts src/renderer/src/store/slices/repos.test.ts src/renderer/src/components/sidebar/worktree-list-groups.test.ts` → **135 passed** post-merge.

Net diff against `main` is unchanged in scope: 3 files, +142/-1.

Noobear:
Follow-up on Symptom 2 from #10031 (*same project opened twice*), which I flagged earlier as "the likely source … not separately verified yet".

## What I observed

On the patched build (`9fa7826`), in the same multi-runtime environment #10031 was `env_blocked` on (local + paired HUB runtime, 4 nested SSH hosts):

- **Terminal-create action ×1 → exactly 1 terminal.** No duplicate spawn.

## What this does and does not establish

I want to be precise about the scope, because this is **not a controlled before/after**:

- This is an **after-only observation on the patched build**. I did not capture a matching pre-patch run of the same action in the same session, so I cannot attribute the absence of duplication to this change.
- Symptom 2 was **not reproduced on `main`** in this environment either — the symptom I could reproduce and measure was the grouping/labelling one (55 rows / 31 unique ids / 7 ambiguous surfaces → 31 / 31 / 0).

So the honest summary is: **the mechanism is consistent, the causal link is not demonstrated.**

## Why the mechanism is still plausible

The dedupe sits upstream of grouping — every consumer of `projectHostSetups` was counting the same checkout twice, not just the sidebar surface-key logic. Any consumer that maps a setup row to a spawn would inherit that doubling. That makes Symptom 2 a natural downstream effect of the same duplicate rows, but "natural downstream effect" is a hypothesis here, not a measurement.

If a controlled repro of Symptom 2 matters for merging, I still have the environment and can run a paired before/after on request — just say which action you want instrumented.

PR #9804, opened 2026-07-22
fix(terminal): prevent reconnect from killing live daemon sessions

OrcaWin:
## Summary

- Separate explicit user terminal closes from reconnect/lifecycle observations with the additive `session.tabs.closeLifecycle` RPC.
- Require exact publication, terminal incarnation, runtime environment, ownership, liveness, and intent evidence before lifecycle retirement; missing or incompatible evidence keeps and audits instead of killing.
- Preserve legacy daemon discovery and warm PTY reattachment while preventing stale mirrors from routing destructive kills across surviving daemon generations.
- Add a deterministic native-Windows v21-v25 named-pipe harness covering production desktop discovery, repeated reconnects, client relaunch, simultaneous clients, remote-profile requests, exact root/descendant liveness, and bounded `shutdown-dispose-failed` fencing.

Addresses #9749. The complete issue/PR investigation and call-flow reconstruction are included in `docs/reference/plans/2026-07-21-windows-daemon-generation-safety-investigation.md`.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — changed-file `oxlint` and code formatting pass; the repository-wide switch-exhaustiveness lane has pre-existing unmatched cases in `skill-freshness-group.tsx` outside this diff.
- [x] `pnpm typecheck`
- [ ] `pnpm test` — focused production/RPC/renderer coverage passes 948/949; the sole failure is the existing Windows `/tmp` versus `\tmp` `badgeColor` dedupe expectation in untouched code.
- [ ] `pnpm build` — fresh Electron Vite and paired web-client builds pass; the full native/package build was not run.
- [x] Added or updated high-quality tests that would catch regressions, or explained why tests were not needed

Focused evidence:

- Native Windows five-generation reconnect oracle: 2/2 passed in 99.1s; v21-v25 each accepted 16 reconnect hellos, all ten PTY roots and ten descendants survived six lifecycle attempts per stale mirror, and zero `session-killed` events occurred.
- The same oracle passed with a fresh Electron E2E build in 137.1s; a separate 25-burst stress run passed.
- Practical Electron clean quit/relaunch with a live daemon-backed terminal: 1/1 passed in 24.0s with restored output and working keyboard/direct PTY input.
- Mixed-version preservation/retirement: 2/2 passed; non-empty v22 remained reattachable while empty v24 retired.
- Daemon discovery/adoption/retirement suites: 224 passed, 3 skipped in the final broad lane; the focused changed scanner lane passed 133/133.
[...]

PR #5975, opened 2026-06-21
fix: close mirrored editor tabs on the host so they stop reopening

gsxdsm:
## What & why

On web and mobile companions, some editor file tabs reopened the instant you closed them.

Companions don't own their editor tabs — they mirror them from the host's runtime-session snapshot, which the host builds from its authoritative `openFiles`. Closing a mirrored tab only removed it on the companion, so the next snapshot re-mirrored the file (still open on the host) and the tab popped right back. Locally-opened files closed fine, which is why it was "some" tabs.

## The fix

Tell the host to close its own copy. The close is centralized at the single chokepoint every editor close funnels through — `store.closeFile` (tab strips, bulk close-others/to-right, save/discard, the floating panel). When the file is `mirroredFromRuntimeSession` and a web runtime session is active, it fires `session.tabs.close` for the host tab id, then proceeds with the normal local removal. The RPC records a close-intent that suppresses re-mirroring until the host snapshot catches up, so the local removal isn't undone and there's no flicker. It's a no-op for the host's own (non-mirrored) files, so desktop behavior is unchanged.

Also fixes the desktop `ui:closeSessionTab` handler: a companion-driven editor close now routes through `closeFile` (which clears `openFiles`) instead of `closeUnifiedTab` (tab strip only) — the same re-mirror bug, hit by mobile-originated closes.

## Why centralize (vs. patching each close site)

An earlier draft guarded individual call sites in `Terminal.tsx`, but that's the legacy fallback titlebar — the primary tab-group UI (`useTabGroupWorkspaceModel`), the dirty save/discard handlers, and the floating panel all close through `closeFile` directly and would have stayed broken. One chokepoint covers every surface, including future ones, and removes the duplication.

## Tests

- `closeFile` routes through the host-mirror notifier and still removes the file locally (locks the chokepoint).
- `notifyHostOfMirroredEditorClose`: fires with the host tab id; no-ops for non-mirrored files and inactive sessions.
- `ui:closeSessionTab`: editor close → `closeFile`; non-editor close → still `closeUnifiedTab` (keeps the editor-only routing scoped).

## Notes

No demo embedded — reproducing requires a live host + web/mobile companion runtime session, which isn't capturable here. Renderer typecheck, oxlint, and the affected vitest suites pass.

---

<a href="https://stagereview.app/stablyai/orca/pull/5975">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
[...]

PR #9836, opened 2026-07-22
fix(runtime): scope mirrored-editor close intent to its runtime environment

nwparker:
## What

`recordWebSessionCloseIntent` and `isWebSessionCloseIntentPending` gained a leading `environmentId` parameter in #9804, and the `web-runtime-session.ts` callers were updated to the 4-arg form — but the **second** caller, in `close-mirrored-editor-tab.ts`, was missed and still passed 3 args. This breaks the web typecheck on `main`:

```
src/renderer/src/runtime/close-mirrored-editor-tab.ts(43,3): error TS2554: Expected 4 arguments, but got 3.
src/renderer/src/runtime/close-mirrored-editor-tab.test.ts(49,12): error TS2554: Expected 4 arguments, but got 3.
```

Fix: pass the already-validated `runtimeEnvironmentId` (the exact same value handed to `closeWebRuntimeSessionTab({ environmentId })` immediately below) so the close intent is recorded under the correct `closeIntentScopeKey(environmentId, worktreeId)`. Update the test's `isWebSessionCloseIntentPending` assertion to the 4-arg form (`env-1` is what the test's mock resolves the worktree's runtime env to).

## ELI5

When you close a file that's mirrored from a remote host, Orca whispers to the host "I closed this, don't send it back." A recent change started filing those whispers in per-environment folders, but one code path forgot to say *which* environment — so the compiler refused to build, and even if it had built, the whisper would've gone in the wrong folder and the tab could pop back. This puts the environment label back on that one path.

## Proof of zero regression

- `close-mirrored-editor-tab.test.ts`: **4/4 pass** (including the "records the host close intent SYNCHRONOUSLY" test that exercises the fixed call path with `env-1`).
- `pnpm typecheck:web`: **clean** (was failing on `main` before this).
- The passed value is identical to the `environmentId` already used two lines down in `closeWebRuntimeSessionTab`, so the two now agree on scope — this is the behavior #9804 intended.

## Why this is a fix, not just a type patch

Had the code compiled with the missing arg, the intent would have been recorded under `closeIntentScopeKey(undefined, worktreeId)` while `closeWebRuntimeSessionTab` re-records/reads under `closeIntentScopeKey(runtimeEnvironmentId, worktreeId)` — a scope mismatch that reopens the "host snapshot flashes the just-closed mirrored tab back" race #9804 was hardening against.

### DIFF
--- a/src/renderer/src/runtime/close-mirrored-editor-tab.test.ts
+++ b/src/renderer/src/runtime/close-mirrored-editor-tab.test.ts
@@ -46,7 +46,9 @@ describe('notifyHostOfMirroredEditorClose', () => {
     // host snapshot in the dynamic-import gap can't flash the old-path tab back.
     notifyHostOfMirroredEditorClose(buildState(), 'wt-1', 'file-1')
 
-    expect(isWebSessionCloseIntentPending('wt-1', toHostSessionTabId('host-tab-1'), now)).toBe(true)
+    expect(
+      isWebSessionCloseIntentPending('env-1', 'wt-1', toHostSessionTabId('host-tab-1'), now)
+    ).toBe(true)
   })
 
   it('closes the mirrored editor tab on the host using the host tab id', async () => {
--- a/src/renderer/src/runtime/close-mirrored-editor-tab.ts
+++ b/src/renderer/src/runtime/close-mirrored-editor-tab.ts
@@ -40,7 +40,12 @@ export function notifyHostOfMirroredEditorClose(
   }
   // Record the close intent SYNCHRONOUSLY so a host snapshot landing before the dynamic import below resolves can't
   // flash the old-path tab back. closeWebRuntimeSessionTab re-records it idempotently.
-  recordWebSessionCloseIntent(worktreeId, toHostSessionTabId(unifiedTab.id), Date.now())
+  recordWebSessionCloseIntent(
+    runtimeEnvironmentId,
+    worktreeId,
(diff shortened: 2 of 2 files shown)


brennanb2025 (approved):
Reviewed against current main — this is the correct and complete fix for the typecheck breakage that's failing `verify` on every open PR.

Verified:
- #9804 widened `recordWebSessionCloseIntent`/`isWebSessionCloseIntentPending` with a leading `environmentId` and updated the `web-runtime-session.ts` callers, but merged past #5975's then-new caller in `close-mirrored-editor-tab.ts` — a semantic merge conflict (both PRs were green individually). Current main fails web typecheck on exactly the two lines this PR touches.
- Swept main for all call sites of the close-intent functions (`record`/`isPending`/`reconcile`/`clear*`): everything else is already 4-arg, so this PR closes the only gap — no other stale caller left behind.
- Passing `runtimeEnvironmentId` is semantically right, not just type-appeasing: it's the same validated value handed to `closeWebRuntimeSessionTab` below, and matches the `environmentId` the snapshot reconcile in `web-session-tabs-sync.ts` uses when checking `isWebSessionCloseIntentPending` — so the intent lands in the scope the reconcile actually reads.
- Test update matches the mock (`getRuntimeEnvironmentIdForWorktree` → `'env-1'`).
- `verify` passed on the merge commit after #9804 landed, so the merged tree is proven green.

LGTM — landing this unbreaks typecheck for every open PR.

PR #9849, opened 2026-07-22
perf(main): park worktree metadata pollers while the window is hidden

nwparker:
## What

The per-repo worktree metadata pollers ran 24/7 with **no window-visibility gate**. At ~20 repos on macOS that's ~50 `fs.stat` syscalls/sec continuously, even while the Orca window is hidden/minimized — pure idle-battery drain. This parks the poll **timers** while the window is hidden and resumes **losslessly** when it becomes visible again.

- New injected `WorktreePollerWindowVisibility` (mirrors the shipped `ssh-port-scanner` park/resume pattern). Production adapter follows each `ActiveWatch.mainWindow`; a `null`/destroyed window is treated as **always visible** so headless/`orca serve` and the macOS window-recreation gap never park.
- Converted the three poll loops (base poller, Windows/Linux git-common polling, darwin primary-checkout metadata polling) from `setInterval` to a one-shot `setTimeout` chain that checks visibility **before** the stat fan-out: if hidden, it parks (`parkedWhileHidden = true`, no reschedule) without touching the retained snapshot. `onWindowBecameVisible` resumes with an **immediate** tick that diffs a fresh snapshot against the retained pre-hide snapshot (base forces a full scan; git-common forces the linked-index leaf read), so anything changed while hidden surfaces on reveal.
- The darwin native `worktrees/` fsevents watcher, its teardown/re-arm lifecycle, and the directory-existence poll stay **always-on** (push-based, ~0 idle cost; parking it would risk the #8732 teardown heap-corruption race).

## ELI5

When Orca was hidden, it still poked your disk ~50 times a second per second to notice git changes across all your repos — for windows you couldn't even see, draining battery. Now it stops poking while hidden and, the moment you come back, does one fresh check that catches anything that changed while it was away. The always-on native file watcher (which is basically free) keeps running so nothing is missed on macOS.

## Proof of zero regression

- **Lossless resume**: parking never resets the retained snapshot; resume forces an immediate full/leaf-read diff against it, so a `git worktree add/remove`, commit, or HEAD move made while hidden is detected on reveal (covered by tests).
- **Headless/serve exemption**: `null`/destroyed window ⇒ always visible ⇒ polls normally, so a windowless host is never starved (tested).
- **Native watcher untouched**: only the poll timers are gated; the darwin native subscription and re-arm path are never parked/resubscribed (tested to keep flowing while hidden).
- **Same-granule detection preserved**: git-common resume forces the linked-index read; `logs/HEAD` is still stat'd directly every tick.
- Tests: extended `worktree-base-directory-poller.test.ts`, `worktree-git-common-watch.test.ts`, `worktree-base-directory-watcher.test.ts` — **45/45 pass** (verified in main worktree, not just the author's). New cases: no scans while hidden, hidden add/HEAD-move detected on reveal, null-window & destroyed-window exemptions, native watcher stays subscribed while hidden. `pnpm typecheck:node`: clean. `oxlint` + max-lines ratchet: pass.

## Proof of perf improvement

While the window is hidden/minimized this parks ~50 local `fs.stat`/sec at 20 repos (5 primary-metadata stats per 2s per repo, plus the Windows/Linux readdir + per-linked-worktree structural fan-out). The tests assert the poll body (`onFullScan`/snapshot) does **not** run while hidden — directly proving the syscalls stop — and that a reveal triggers exactly one immediate catch-up scan. Pure background/idle battery win; zero cost while visible, and the always-on native fsevents watch means no missed events on macOS.

### DIFF
--- a/src/main/ipc/worktree-base-directory-poller.test.ts
+++ b/src/main/ipc/worktree-base-directory-poller.test.ts
@@ -3,8 +3,10 @@ import { mkdtemp, mkdir, realpath, rm, stat, utimes, writeFile } from 'node:fs/p
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import {
+  createWorktreePollerWindowVisibility,
   startWorktreeBaseDirectoryPoller,
-  type WorktreeBasePollEvent
+  type WorktreeBasePollEvent,
+  type WorktreePollerWindowVisibility
 } from './worktree-base-directory-poller'
 import type {
   WorktreeBaseRepoWatchConfig,
@@ -12,6 +14,37 @@ import type {
 } from './worktree-base-directory-event-filter'
 
 const POLL_MS = 25
+
+type VisibilityHarness = {
+  source: WorktreePollerWindowVisibility
+  hide: () => void
+  show: () => void
+}
+
+function createVisibilityHarness(initiallyVisible = true): VisibilityHarness {
+  let visible = initiallyVisible
+  let listener: (() => void) | null = null
+  return {
+    source: {
+      isWindowVisible: () => visible,
+      onWindowBecameVisible: (nextListener) => {
+        listener = nextListener
+        return () => {
+          if (listener === nextListener) {
+            listener = null
+          }
+        }
+      }
+    },
+    hide: () => {
+      visible = false
+    },
+    show: () => {
+      visible = true
+      listener?.()
+    }
+  }
+}
 
 function makeTarget(
   kind: 'base' | 'git-common',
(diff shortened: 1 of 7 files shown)


nwparker:
Adversarial review pass (independent frontier model) found and I fixed one real regression, and cleared everything else:

**Fixed (was a real regression):** an `ORCA_E2E_HEADLESS` launch keeps a *live but never-shown* BrowserWindow, so `isMainWindowVisible` stayed false forever with no show/restore signal to resume — the pollers would have parked permanently and silently dropped watcher behavior from the `electron-headless` suite. The adapter now parks a window only **after it has actually been shown at least once**; a never-shown (headless/starting-up) window and null/destroyed windows stay always-visible. Added two unit tests (live-never-shown keeps polling; shown-then-hidden parks). 47/47 watcher tests + `typecheck:node` green.

**Pre-existing, out of scope (not introduced here):** during the macOS dock-reactivation rebind gap the destroyed window is treated as visible, so a due tick can advance the baseline while `handleLocalWatchEvents` drops the event. Today's unconditional `setInterval` polls during that same gap identically, so this PR does not change that behavior; the narrow race predates it and is better addressed separately.

Cleared: no timer-stall interleaving, lossless resume for all three pollers, per-watch visibility re-checked under the global signal, darwin native watcher fully ungated, gap-after-completion cadence + `unref` preserved, SSH watches bypass these pollers, all disposal paths clear timers + unsubscribe.

PR #9850, opened 2026-07-22
fix(agent-status): keep Claude in-process teammates visible as idle sidebar rows

brennanb2025:
## Problem

On Claude Code 2.1.21x, named Agent-tool agents run as **turn-based in-process teammates** (`taskKind: "in_process_teammate"`): they work a mailbox turn, then emit `SubagentStop` + `TeammateIdle` and sit alive-but-parked until messaged again. Verified live on 2.1.217 with a hook-capture probe and against Orca's own `agent-hooks/last-status.json`:

- `SubagentStart` → the child row appears with the expected teammate-shaped id (`a<name>-<hex>`) —
- `SubagentStop`/`TeammateIdle` arrive **~20s later at the first turn boundary while the teammate's own background work is still running**, and Orca deletes the row.

Orca's roster treated both events as finish signals (per #8825, correct for the 2.1.210-era semantics they were verified against). Result: teammate rows exist only during brief working bursts, which reads as "the sidebar never shows my subagents". The delivery pipeline itself was proven healthy end-to-end.

## Fix (root cause: model the new lifecycle, don't patch the symptom)

The roster now tracks a `working | 'idle'` state per child instead of holding working children only:

- **One-shot children** (hyphen-free ids): unchanged — `SubagentStop` is a true finish and removes the row.
- **Teammate-shaped rows** park as **idle** on `SubagentStop`/`TeammateIdle` and revive to working on the next `SubagentStart` (same lifecycle id; first-observed `startedAt` preserved). The sidebar keeps showing them as idle child rows (renderer + wire already supported `state: 'idle'`; the dashboard already labels idle subagents "alive-but-idle").
- **Done-gate preserved (#8825):** only *working* children gate the pane 'working'; idle teammates never pin a finished pane's spinner.
- **Idle-pile prevention (#8825's original symptom):** named *workflow lanes* reuse the teammate id shape but genuinely finish on stop. Discriminator: real teammates always get a name-keyed `TeammateIdle`. Unconfirmed idle rows are reaped at the next complete lead-Stop `background_tasks` fold (or immediately on stop once a fold tagged the id `listedAsSubagentTask`), so finished lanes can't accumulate.
- **Cap safety:** at the 32-row wire cap the oldest idle row is evicted to admit a working spawn; working children are never displaced.
- **Restart:** hydrate keeps pruning idle snapshots — idle-teammate liveness can't be proven across a restart; a live teammate re-earns its row via `SubagentStart`.

## Evidence behind the semantics

- Interactive 2.1.217 probe with capture hooks: `SubagentStart`/`SubagentStop`/`TeammateIdle` all fire for in-process teammates; teammate `background_tasks` entries still read `status: "running"` permanently (so the fold still can't trust them per-agent — unchanged rules there).
- Live end-to-end from an Orca pane: `last-status.json` showed the teammate row appear (`state: "working"`) then get wiped by `TeammateIdle` mid-`sleep 180` — the exact reported symptom.
- The new lifecycle test replays the captured real event sequence (ids and ordering) from the reporting session.
- Workflow-lane probe (2.1.217): `Workflow` tool lanes spawn as **hyphen-free one-shot ids** (`agent_type: "workflow-subagent"`) with no `TeammateIdle` — they take the remove-on-stop path, so workflows cannot build an idle pile under this change. The `confirmedTeammate` fold guard remains as defense-in-depth for the 2.1.210-era teammate-shaped lane ids.

## Behavior trade-off on older Claude CLIs

On ≤2.1.210, `TeammateIdle` meant "finished" and teammate `background_tasks` entries linger; finished named agents there will now show as idle rows for the rest of the session (honestly labeled, never gating done, cap-evictable, cleared on restart/new session/complete-inventory teardown). Current CLIs get the correct persistent-teammate view; the old-CLI regression risk is bounded clutter, not stuck panes.

## Tests

- `claude-subagent-roster.test.ts`: idle parking, workflow-lane removal via `listedAsSubagentTask`, revival keeping `startedAt`, confirmed-idle fold survival vs unconfirmed reap, oldest-idle cap eviction, TeammateIdle name-scoping now parking instead of deleting.
- `claude-subagent-row-lifecycle.test.ts`: new turn-cycle scenario (work → idle → lead Stop → SendMessage revival) with live-captured ids; updated stop/fold expectations.
- `agent-hook-listener.test.ts`: teammate stop/idle now emit idle snapshots; pane still settles done under idle-only rosters.
- Full sweep: `src/shared` + `src/main/agent-hooks` + `src/main/ipc/agent-hooks` = 3183 passed. Typecheck green. Renderer sidebar/dashboard suites: only pre-existing failures unrelated to this change (fail identically on base commit `ac909c8d83`); same for the pre-existing `skill-freshness-group.tsx` switch-exhaustiveness lint error.

**Live Electron QA: PASSED** — dev app from this branch with a real `claude` CLI (2.1.217) spawning a real in-process teammate; full spawn → idle-persists → SendMessage-revival → idle-again cycle verified against the sidebar and `last-status.json`. Screenshots in the PR comment below.

### DIFF
--- a/src/main/agent-hooks/server.ts
+++ b/src/main/agent-hooks/server.ts
@@ -166,7 +166,7 @@ function dropHydratedIdleClaudeSubagents(
     return payload
   }
   const activeSubagents = payload.subagents.filter((subagent) => subagent.state !== 'idle')
-  // Why: older builds persisted finished Claude children as idle rows; prune them so restart can't resurrect the pile.
+  // Why: an idle teammate's liveness can't be proven across a restart (its TeammateIdle confirmation is in-memory); prune so a dead pile can't resurrect — a live teammate re-earns its row via SubagentStart.
   return {
     ...payload,
     subagents: activeSubagents.length > 0 ? activeSubagents : undefined
--- a/src/main/claude/hook-settings.ts
+++ b/src/main/claude/hook-settings.ts
@@ -35,8 +35,8 @@ export const CLAUDE_EVENTS = [
   { eventName: 'StopFailure', definition: { hooks: [{ type: 'command', command: '' }] } },
   // Why: subagent/teammate lifecycle feeds the sidebar's child rows and keeps
   // a pane 'working' while background children outlive the lead's turn.
-  // TeammateIdle retires the working-only row when SubagentStop is lost;
-  // idle teammates still report status "running" in Stop's background_tasks.
+  // TeammateIdle parks turn-based teammates without trusting their permanently
+  // "running" background_tasks entry to gate the pane.
   // Older Claude builds ignore unregistered event names (StopFailure precedent).
   { eventName: 'SubagentStart', definition: { hooks: [{ type: 'command', command: '' }] } },
   { eventName: 'SubagentStop', definition: { hooks: [{ type: 'command', command: '' }] } },
--- a/src/shared/agent-hook-listener.test.ts
+++ b/src/shared/agent-hook-listener.test.ts
@@ -2698,12 +2698,12 @@ describe('shared agent-hook-listener', () => {
       expect(stopped?.payload.state).toBe('done')
     })
 
-    it('removes a finished teammate/named agent on SubagentStop despite its task reading running', () => {
-      // Why: the interactive agent-teams / orchestration shape observed live —
+    it('parks a teammate as a persistent idle row across its stop/idle/lead-Stop cycle', () => {
+      // Why: the interactive agent-teams shape observed live on 2.1.217 —
       // lifecycle events use `a<name>-<hex>` agent ids while background_tasks
-      // uses unrelated `type: "teammate"` task ids that report "running"
-      // forever, even after the named agent finished. The finished row must
-      // leave the sidebar at once (the reported "long idle list" symptom).
+      // uses unrelated `type: "teammate"` task ids. SubagentStop + TeammateIdle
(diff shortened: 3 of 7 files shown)


brennanb2025:
## Live Electron QA — real `claude` CLI, real in-process teammate ✅

Ran the dev app from this branch (`pnpm dev`, Claude Code CLI **2.1.217**) and drove the real product path end-to-end: `orca worktree create --agent claude` → the lead Claude spawned a named teammate (`qa-mate`) via the Agent tool → verified each lifecycle phase against both the sidebar UI and the persisted `agent-hooks/last-status.json` ground truth.

**1. Teammate finished its turn and went idle — the child row now persists** (pre-fix it vanished at exactly this moment; the transcript on screen even narrates "Teammate @qa-mate finished … now idle and standing by"). Persisted payload: `state: done`, `subagents: [{id: "aqa-mate-a6c9b3919d81d068", state: "idle"}]` — idle row visible, pane not pinned working:

teammate idle row persists

**2. SendMessage wake → same lifecycle id revives to working** (card, lead row, and `qa-mate` child row all spinning; persisted payload flipped to `subagents: [... state: "working"]`):

teammate revived working

**3. Second turn over → parked idle again, row still there a minute later** (sidebar close-up):

sidebar crop idle

Full cycle observed live: **spawn → working row → idle row persists → revival → idle again** — with the done-gate behaving (pane resolves `done` while the idle teammate stays listed).

PR #9854, opened 2026-07-22
fix(daemon): reap Windows ghost terminals from a wedged ConPTY

hanbong5938:
## Summary

Fixes Windows **ghost terminals** — sessions that survive a kill and keep showing in the Resource Manager's orphan panel, where **"Kill N orphan terminals" does nothing**.

Root cause is a wedged ConPTY. After a graceful `kill()`, `forceKill()` was a permanent no-op on Windows (guarding against a ConPTY double-close), and a session's "exited" state depended entirely on node-pty's `onExit`, which a wedged ConPTY never fires. The session then stays `isAlive` forever: `listSessions()` keeps returning it, the orphan panel keeps showing it, and daemon shutdown can also stall because teardown awaits a physical exit that never arrives.

The fix restores a working forced-termination and exit-detection path on Windows, in three layers:

- **A — real force-kill escalation** (`daemon/pty-subprocess.ts`, new `pty/os-process-termination.ts`): on Windows, `forceKill()` now escalates out-of-band via `taskkill /T /F` instead of returning. It never re-enters node-pty (so no double-close), and `/T` reaps the descendant tree ConPTY orphaned.
- **B — independent exit detection** (`daemon/session.ts`): after a force-kill, the session polls the real OS pid; once the process is gone it synthesizes the exit, so it transitions to `exited`, fans out `onExit`, and the host reaps it — even when node-pty's `onExit` never fires. `handleSubprocessExit` is made idempotent so a synthesized exit racing a real `onExit` fires exactly once.
- **C — periodic reconciliation** (`daemon/terminal-host.ts`): a 30 s unref'd sweep reaps force-killed sessions whose OS pid is already gone. Clears pre-existing ghosts and backstops A/B.

## Screenshots

No new or changed UI. Behavior-only fix: the existing Resource Manager orphan panel now clears when "Kill N orphan terminals" is used on Windows (previously the count and rows persisted, because the daemon never reported the sessions as exited).

## Testing

- [x] `pnpm lint` — `oxlint` clean on all changed/new files (no errors); no new `max-lines` suppressions added (the reconcile sweep was extracted to its own module so `terminal-host.ts` stays under the limit).
- [x] `pnpm typecheck` — passes (exit 0, full repo).
- [x] `pnpm test` — all changed/added daemon + pty tests pass. The only failures observed locally are **pre-existing, Windows-only** tests unrelated to this change (a Codex-home path test and two daemon socket-lifecycle tests) that fail identically on `main` and pass on the Linux CI runner.
- [x] `pnpm build` — passes (desktop bundle + native build, exit 0).

New/updated tests:
- `src/main/pty/os-process-termination.test.ts` — `killOsProcessTree` (Windows `taskkill /T /F` vs POSIX SIGKILL, invalid-pid guard, already-gone tolerance) and `isProcessAlive` (ESRCH→gone, EPERM→alive, ambiguous→assume-alive, invalid-pid).
- `src/main/daemon/pty-subprocess.test.ts` — new Windows test: `forceKill` after a graceful `kill()` escalates via `killOsProcessTree(pid)`, does NOT re-enter node-pty, and is idempotent. Two POSIX-SIGKILL-semantic tests moved to `itOnPosixHost`.
- `src/main/daemon/session.test.ts` — synthesizes exit when the OS pid dies but `onExit` never fires; prefers a real `onExit` over synthesis and fans out once; `reconcileWedgedExit` reaps a force-killed dead-pid session and no-ops before force-kill.
- `src/main/daemon/terminal-host.test.ts` — reaps a wedged immediate-kill session once its OS process is gone.
- `reattach-snapshot`, `terminal-host-session-reaping-leak`, `terminal-host-startup` tests mock `isProcessAlive → true` to preserve prior behavior for their synthetic pids.

## AI Review Report

Reviewed with an AI coding agent focused on the highest-risk areas for this change:

- **Cross-platform (macOS / Linux / Windows), explicitly checked.** All new behavior is gated behind `process.platform === 'win32'` runtime checks. The POSIX force-kill path (`forceKillPosixPtyProcessGroups` → process-group SIGKILL) is unchanged; POSIX-specific tests are pinned with `itOnPosixHost`. The new `killOsProcessTree` branches by platform (`taskkill /T /F` on Windows, `process.kill(pid,'SIGKILL')` elsewhere); `isProcessAlive` uses signal `0`, which behaves on all three. No hardcoded path separators; no shell/label/shortcut changes.
- **ConPTY double-close hazard.** The whole point of the original Windows guard was that re-entering node-pty's `kill()`/`destroy()` double-closes the native handle (heap corruption). The fix escalates only *out-of-band* (`taskkill`, `process.kill`) and never re-enters node-pty; the existing `dispose()` Windows guard is untouched.
- **PID reuse.** Liveness polling and reconciliation are scoped to sessions we actually force-killed (`forceKillSent`), and the wrapper keeps its `dead`-recycled-pid guard, so a still-live or reaped session's pid is never probed/signalled. Residual risk (a long-lived ghost's pid recycling before the sweep) is noted as follow-up.
- **Double-fire.** `handleSubprocessExit` is now idempotent (`_state === 'exited'` short-circuit + idempotent `markExited`), so a synthesized exit racing node-pty's real `onExit` fans out to clients / the reaper exactly once.
- **SSH / remote.** Change is confined to the local daemon PTY layer. SSH PTYs route through their own provider and relay tree-termination (`subprocess-tree-termination.ts`); this path is not touched.
- **Performance.** No hot-path cost: the OS-pid poll runs only while a force-killed session awaits exit (250 ms interval, unref'd, cleared on exit); the host reconcile is a 30 s unref'd sweep that no-ops unless a session is both force-killed and dead.

## Security Audit

- **Command execution.** `killOsProcessTree` invokes `taskkill` via `execFile` with an **argument array** (no shell), and the only interpolated value is `String(pid)` after a `Number.isInteger(pid) && pid > 0` guard — no shell metacharacters, no command injection. This matches the existing `subprocess-tree-termination.ts` / `codex-accounts` taskkill usage.
- **Process signalling.** `process.kill` is only ever called on `this.subprocess.pid` for sessions the daemon owns and has force-killed; the recycled-pid guards prevent signalling an unrelated process.
- **Input handling / paths / auth / secrets / IPC.** None introduced or changed. No new IPC surface, no user-controlled input reaches the new code, no filesystem/network/credential handling.
- **Dependencies.** No new dependencies; `taskkill` is a Windows system binary.

Follow-up: optionally strengthen PID-identity with a creation-time cross-check before synthesizing, to fully close the long-lived-ghost pid-reuse window.

## Notes

- **Windows-specific fix.** Same root cause underlies daemon shutdown occasionally not completing (teardown awaits a physical exit a wedged ConPTY never delivers). Relevant Windows e2e workflows: `win-update-survival-e2e`, `windows-terminal-restart-e2e`.
- **Out of scope:** this does **not** address the separate "surviving daemon generations mass-kill live sessions on reconnect" behavior (a daemon-adoption-layer concern) — that belongs in its own PR. This PR only fixes the shared PTY-termination root cause behind the "can't kill / ghost session" symptom.
- **Follow-up:** PID-reuse hardening (creation-time cross-check) for the liveness poll and reconcile sweep.

X handle: 

## ELI5

On Windows, killed terminals could stay forever in the orphan list because a wedged ConPTY never reported exit. The daemon can force-reap those ghosts so Kill orphan terminals actually works.


hanbong5938:
Good catch — this was a real defect, not just a missing log. Fixed in `5c4fded`.

You were right that logging alone wouldn't unstick the session, so the fix re-arms the escalation at both latch points you traced:

1. **`killOsProcessTree(pid, onFailure?)`** — no longer swallows the result. Windows reports the `taskkill` error through `onFailure`; POSIX now distinguishes `ESRCH` (child already gone → success) from real failures like `EPERM`.
2. **`forceKill()`** — on failure it resets `osForceKillIssued = false` and warns, so a subsequent `forceKill()` retries instead of returning early. This also restores the file's existing convention (`kill()` already un-latches `nodePtyKillIssued` on a thrown error) that the Windows branch had broken.
3. **`reconcileWedgedExit()`** — previously it only synthesized an exit when the pid was gone, so a *still-alive* child (denied kill) was invisible to it. It now detects that case, re-arms `forceKillSent`, and re-escalates. The 30s reconciler sweep therefore retries automatically rather than leaving the session latched in `terminating` forever.

Tests added: `onFailure` fires on taskkill error / not on success, POSIX `ESRCH` vs `EPERM`, `forceKill` retries after a failed OS kill, and the reconciler re-escalating a child that outlived its force-kill.

Also added the missing JSDoc flagged by the docstring-coverage check.

Verified locally: `oxlint` clean, `pnpm typecheck` exit 0, and `src/main/daemon` + `src/main/pty` at 1029 passed / 4 failed — those 4 fail identically on the merge-base on Windows (Codex-home path + daemon socket-lifecycle) and pass on the Linux CI runner.

hanbong5938:
Fixed in `c448433` — you were right, and this was a risk my previous fix introduced.

Taking the second option you suggested (*"limit re-escalation to a verified failed termination of the original child"*), because a liveness probe fundamentally can't distinguish our child from a stranger on a recycled pid:

- **`reconcileWedgedExit()` no longer re-kills at all.** The liveness probe is now used *only* to synthesize an exit. A recycled pid reads as alive → we simply do nothing, which is harmless. Previously it would have force-killed that unrelated tree every 30s.
- **Re-escalation is now driven by a confirmed refusal**, not a probe. `SubprocessHandle.forceKill(onFailure?)` propagates the OS refusal up, and `Session.handleForceKillFailure()` re-arms `forceKillSent` only then. A refused kill proves the child was never terminated, so the pid is still ours — the one case where retrying a bare pid is sound.

This also lines up with the identity-safety convention already in `pty-descendant-termination.ts` (`hasUnambiguousStartIdentity`), which my previous version had broken.

Known limitation, stated plainly: if the pid *is* recycled, the session stays wedged rather than being reaped. That's a liveness gap, not a destructive one, and it matches the pre-existing behavior. Tracking a spawn-time process identity would close it, but that's a larger change than this PR should carry.

Also addressed:
- **Minor (test)**: `isProcessAlive` now asserts the exact `process.kill(pid, 0)` call, so an accidental destructive signal can't pass.
- **Docstring coverage**: added the missing docs, including test helpers.

Tests updated: the old "re-escalates" test is replaced by *"never re-kills from a liveness probe (pid-reuse safety)"* and *"re-arms after a confirmed kill refusal so the next request retries"*.

Verified locally: `oxlint` clean, `pnpm typecheck` pass, `src/main/daemon` + `src/main/pty` at 1046 passed / 3 failed — those 3 fail identically on the merge-base on Windows and pass on Linux CI.

nwparker:
Closing, with one finding from this PR adopted.

You identified that on Windows `forceKill()` is a **permanent no-op** after any `kill()` — `pty-subprocess.ts` returns early on `nodePtyKillIssued` to avoid double-closing the ConPTY handle node-pty owns. That guard is correct, but it means a session whose ConPTY has wedged (and so never fires `onExit`) has no escalation path at all. Exactly the ghost terminal you described.

W2 (#15755) gave every pane a Win32 job object, which is the escalation that guard was missing: terminating the job kills the tree without touching node-pty's handle. Wired up in **#15886**, with you credited as co-author and this PR cited for the diagnosis.

Not adopting the rest — `wedged-session-reconciler.ts` and the surrounding liveness machinery. With the job in place the wedge is terminable directly, so a separate reconciler that polls for ghosts and reasons about their liveness is state we would rather not carry. If ghosts still show up after #15886, that is the moment to revisit it, and this PR is the design to come back to.

Thanks for the diagnosis — it was right, and it had been sitting in the code for a while.

nwparker:
Diagnosis adopted in #15886 (forceKill now escalates to the pty job); reconciler intentionally not carried.

nwparker:

PR #8727, opened 2026-07-14
fix(tasks): resolve PR work items upstream-first under 'auto' like issues

moseoh:
# fix(tasks): resolve PR work items upstream-first under 'auto' like issues

## Summary

On a fork, opening Tasks → PRs with no explicit source preference showed "No matching GitHub work" even though the user's PRs existed upstream.

## ELI5

When you work from a fork, your pull requests actually live on the original project, not your copy. The PRs tab was looking at your (empty) copy by default, so it looked like you had none. Now it looks in the right place.

## Root cause & fix

Under the `auto` preference, the issue side resolved upstream-first, but `resolvePrWorkItemSource()` pinned the PR side to `origin` (the fork), whose PR list is almost always empty for fork-contribution workflows. The fix flips the default so `auto` resolves upstream-first, mirroring the issue side; only an explicit `origin` preference still pins PRs to the fork. It's a single resolution expression shared by `listWorkItems` and `countWorkItems`, so lists, pagination totals, and `sources.prs` metadata stay consistent. Repos without an `upstream` remote are unaffected — the `?? origin` fallback preserves current behavior.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase onto `origin/main` was clean (base `7ab601487c`, head `e2ff0024c3`).

- Suite `src/main/github/client-issue-source.test.ts` passes on branch: **30/30 tests passed**.
- Reverting only the fix (`git checkout origin/main -- src/main/github/client.ts`), keeping the tests, fails **2 tests** (28 passed), capturing the bug:

```
STEP 2a (branch):  Test Files 1 passed (1) | Tests 30 passed (30)
STEP 2b (fix reverted, test kept): Test Files 1 failed (1) | Tests 2 failed | 28 passed (30)

  ✗ preference='auto' + upstream exists → PRs query upstream too
      AssertionError: expected --repo "stablyai/orca" but got "fork/orca"
  ✗ collapses the default count to one query when auto resolves both sides to upstream
      AssertionError: expected mock called 1 time, but got 2 times
```

- Lint clean: `npx oxlint src/main/github/client.ts` → exit 0, no warnings/errors.

## Trade-offs

None — pure fix.

## Regression risk

[...]

PR #9856, opened 2026-07-22
fix(tasks): prefer upstream for PR list when source preference is auto (#9851)

fsdwen:
# fix(tasks): prefer upstream for PR list when source preference is auto

## Summary

Fixes fork workflows where open upstream PRs never appeared in the Tasks list until the user manually clicked the "U" source selector.

## ELI5

When your project is a fork, its open pull requests live on the original ("upstream") repo. The Tasks list used to only look at your fork, so those PRs stayed hidden until you switched the source by hand. Now they show up automatically.

## Root cause & fix

`resolvePrWorkItemSource` in `src/main/github/client.ts` used only `originCandidate` for the `auto` and `undefined` preferences, so in a fork with an `upstream` remote — where open PRs actually live — no upstream PRs were ever queried. The one-line ternary was flipped so that only an explicit `origin` preference returns origin; every other case now resolves `upstreamCandidate ?? originCandidate`. This aligns the PR side with the existing issue-side semantics (`getIssueGitHubApiRepository`, "upstream first, then origin"). For non-fork repos the upstream probe returns `null`, so the fallback to origin keeps behavior unchanged.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Clean rebase onto `origin/main` (base `8a23618310`, head `cc8961051c`); single changed file, one-line change.

- Branch (fix present) — `client-issue-source.test.ts`:

```
npx vitest run --config config/vitest.config.ts src/main/github/client-issue-source.test.ts
-> Test Files 1 passed, Tests 28 passed
```

- Regression guard: `client-work-items.test.ts`, `client-work-items-query-paging.test.ts`, `client.test.ts` -> Test Files 3 passed, Tests 146 passed.
- Honesty note: the existing suite still passes with the fix reverted, so no automated test captures this specific bug. The fix is confirmed by code inspection, not by a failing-on-revert assertion. proven=false.
- Lint clean: `npx oxlint src/main/github/client.ts` -> no diagnostics.

## Trade-offs

None — pure fix. No new I/O; the same two parallel remote probes run as before.

## Regression risk

Low. Only the `auto`/`undefined` branch changes; the explicit `origin` and `upstream` branches are byte-identical in behavior. Non-fork repos have no upstream candidate (probe returns `null`), so they fall through to origin exactly as before. GitLab uses its own `resolveMergeRequestSource` and is untouched. Residual: no test asserts the corrected upstream path, so the guard against future regression is code review rather than CI.

_Original fix by @fsdwen. Validated, rebased, and reviewed via automated bug-bash triage; original one-line fix design preserved with no changes added._

### DIFF
--- a/src/main/github/client.ts
+++ b/src/main/github/client.ts
@@ -1110,7 +1110,7 @@ async function resolvePrWorkItemSource(
     getGitHubApiRepositoryForRemote(repoPath, 'upstream', connectionId, localGitOptions)
   ])
   const source =
-    preference === 'upstream' ? (upstreamCandidate ?? originCandidate) : originCandidate
+    preference === 'origin' ? originCandidate : (upstreamCandidate ?? originCandidate)
   return { source, originCandidate, upstreamCandidate }
 }
 


fsdwen:
@AmethystLiang Please review this PR, thanks!

nwparker:
Thanks for this @fsdwen! Closing as already-resolved: #8727 ("resolve PR work items upstream-first under 'auto'") was merged this session and makes the identical change to `resolvePrWorkItemSource` in `src/main/github/client.ts` — so `main` already has your fix. Really appreciate the contribution, and sorry for the duplicate. 🙏

PR #9857, opened 2026-07-22
perf(mobile): gate host polling on foreground/background

nwparker:
## What

The mobile host screen ran two 3-second polls while connected — the routed screen and the embedded sidebar — and **each fired both `worktree.ps` (a full multi-repo process scan) AND `repo.list`**, with no foreground/background gate. So a phone left connected pinged every 3s (host CPU + relay/cellular bytes + a radio wakeup) even during brief background windows while the socket stays parked.

This consolidates both poll sites into one `startHostWorktreeRefresh` lifecycle and **AppState-gates the interval** so both polls stop while backgrounded and refresh immediately on foreground return:

- `worktree.ps` keeps its 3s cadence while foregrounded — it carries live per-worktree agent status / preview / unread that no push event replaces, so it is **not** removed.
- `repo.list` stays on the interval as an AppState-gated **convergence safety-net** (it self-throttles to `REPO_METADATA_REFRESH_MS` = 60s, so it's ~1 request/min while foregrounded), **plus** a `reposChanged`/`worktreesChanged` fast-path and a reconnect-replay refetch.

The win is the **foreground gate**: while the app is backgrounded, both polls (and their radio wakeups) stop entirely; while foregrounded, behavior matches the old cadence.

## ELI5

Your phone kept pinging the computer twice every 3 seconds — even in your pocket — for the live agent list and a rarely-changing repo icon/color list. Now both pings pause while the app is backgrounded and instantly catch up when you reopen it. The live list keeps its speed while you're looking, and the repo-icon list still refreshes on its own so a desktop settings change never leaves the phone stale.

## Proof of zero regression

- **Repo metadata still converges** (this was caught by adversarial review): desktop Settings repo edits (icon/color/name, repo removal) notify only the renderer IPC, **not** the runtime `clientEvents` stream mobile subscribes to — so `repo.list` cannot be purely event-driven without going stale on a foregrounded screen. It's kept on the (now gated) interval, self-throttled to 60s, exactly as before, so a foregrounded screen still converges within ≤60s. The event fast-path is additive.
- The live `worktree.ps` poll (agent status/preview/unread) is preserved; optimistic reconciliation still gets periodic confirmations while foreground and an immediate one on foreground return.
- Reconnect refetches everything (lifecycle mount-fetch on connState flip + same-instance `ready`-after-`ready` replay), covering "events aren't queued while disconnected."
- Embedded-sidebar path (never focused) wired independently, gated the same way.
- **Verified in a deps-installed mobile checkout**: `pnpm typecheck` clean, **full mobile suite 2232 pass / 2 skipped (307 files)**, `oxlint` clean (within the frozen max-lines budget), `oxfmt --check` clean. Tests cover: interval polls (worktrees + repo.list) while active and skips **both** while backgrounded; immediate foreground refresh; `reposChanged`/`worktreesChanged` refetch; reconnect stream replay.

## Proof of perf improvement

While backgrounded this stops the entire 3s poll loop (worktree.ps + repo.list + their radio wakeups) — the new tests assert neither snapshot is fetched while `AppState !== 'active'`. Foreground steady-state keeps exactly the prior cadence (worktree.ps every 3s, repo.list ≤ once/60s), now with faster runtime-origin repo/worktree convergence via the event fast-path. Net: a connected phone in your pocket stops waking its radio every 3s.

## Follow-up (out of scope)

The cleaner long-term fix for the desktop-repo-edit case is to bridge the renderer-only `notifyReposChanged` to `OrcaRuntimeService.notifyReposChanged()` so mobile gets an instant runtime event and the periodic `repo.list` safety-net could drop further. Left as a separate desktop-side change to keep this PR mobile-only.

### DIFF
--- a/mobile/app/h/[hostId]/index.tsx
+++ b/mobile/app/h/[hostId]/index.tsx
@@ -38,6 +38,7 @@ import {
   useForceReconnect
 } from '../../../src/transport/client-context'
 import { useWorktreeResync } from '../../../src/transport/use-worktree-resync'
+import { startHostWorktreeRefresh } from '../../../src/worktree/host-worktree-refresh'
 import {
   useLastConnectedAt,
   useReconnectAttempt
@@ -141,7 +142,8 @@ export function HostScreen({
   const lastConnectedAt = useLastConnectedAt(hostId)
   const clientRef = useRef<RpcClient | null>(null)
   const fetchWorktreesInFlightRef = useRef(false)
-  const fetchRepoMetadataInFlightRef = useRef(false)
+  const fetchRepoMetadataInFlightRef = useRef(new WeakSet<RpcClient>())
+  const fetchRepoMetadataPendingRef = useRef(new WeakSet<RpcClient>())
   const repoMetadataFetchedAtRef = useRef(0)
   const newWorktreeModalRef = useRef<{ open: () => void }>(null)
   const newWorktreeModalVisibleRef = useRef(false)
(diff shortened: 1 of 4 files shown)


nwparker:
Adversarial review caught one **Major** regression, now fixed: making `repo.list` purely event-driven left desktop Settings repo edits (icon/color/name, removal) stale on a foregrounded host screen, because those notify only the renderer IPC (`repos.ts` `notifyReposChanged`), not the runtime `clientEvents` stream mobile subscribes to. Restored the periodic `repo.list` on the (now AppState-gated) interval as a self-throttling (60s) convergence safety-net — preserving the battery win (both polls stop while backgrounded) while keeping the ≤60s convergence the old code had. Test updated to assert the interval polls repo metadata while foregrounded and skips it (and worktree.ps) while backgrounded. Full mobile suite 2232 pass + typecheck/lint/format clean.

PR #6555, opened 2026-06-28
fix(runtime): require sustained quiescence for tui-idle waits

nwparker:
## 🧑‍🤝‍🧑 Impact & ELI5

**1) What's broken today (ELI5).** Orca lets you script your AI coding agents — for example, "wait until the agent finishes working, then do the next thing." The command for that is `orca terminal wait --for tui-idle`. The bug: this "are you done yet?" check answers **"yes, done!" instantly — even while the agent is still actively typing, thinking, and streaming output.** So any automation built on top of it fires too early and steps on the agent mid-task. This hits anyone using Orca to orchestrate or chain agent steps (especially agents like Codex, Devin, or aider whose terminal title is just their name). It's not a crash, but for orchestration it's a real blocker: the one signal you'd rely on to know a task is finished is unreliable.

**2) What this PR does (ELI5).** The old check trusted the terminal's *title bar* to decide if the agent was done. But some agents keep a plain title like `Codex` while their spinner and `esc to interrupt` live in the terminal body, so a busy terminal could look idle. The fix now separates strong idle evidence from weak idle evidence: **known ready prompts and explicit done/idle OSC titles still resolve immediately**, while ambiguous name-only titles must also prove the terminal has gone quiet. So Orca still moves fast when the agent gives a real completion signal, but it no longer cuts in just because a weak title sample looked idle.

**3) Tradeoffs / regressions — honest answer.** Current status: **zero added latency for explicit completion signals**. Body-level ready prompts and explicit idle/done OSC titles (for example `Codex done`, Claude `✳ ...`, Gemini/Pi idle markers) satisfy `tui-idle` immediately. The remaining quiet wait applies only to **ambiguous, non-explicit idle evidence** — titles like plain `Codex`, `Devin`, or `aider` where Orca's shared title classifier defaults to `idle` even though the agent may still be working. For that weak signal, the `TUI_IDLE_QUIESCENCE_MS` quiet window is the necessary correctness guard: resolving immediately would reintroduce #6011. The follow-up also schedules that quiet check directly, so the ambiguous-title path resolves at the quiet deadline instead of waiting for the next 2s fallback poll. No setting, new user step, UI change, platform-specific behavior, SSH caveat, or Git-provider behavior changes.

---

## Summary

[...]

PR #9863, opened 2026-07-22
fix(orchestration): verify Codex inject submission

byh3071-cpu:
## Summary

- document a mandatory post-inject submission verification guard for Codex terminals
- send one bare Enter only when `[Pasted Content N chars]` remains in the composer
- re-read the terminal to require draft removal and working activity
- never resend the task text, preventing duplicate queued prompts

This complements #6555: that PR hardens `tui-idle`; this guard verifies actual prompt delivery at the orchestration layer.

## Validation

- `pnpm generate:bundled-skill-guides`
- `pnpm verify:bundled-skill-guides`
- orchestration skill guidance tests (11/11)
- skill `quick_validate.py`
- `git diff --check`
- reproduced false `tui-idle` during Codex MCP startup and confirmed a single bare Enter transitions the pending draft into working state

## ELI5

Orchestration could mark a Codex inject as success when the prompt never left the composer. After inject, Orca checks that the draft was submitted and the agent is working, without resending the task text.


AmethystLiang:
looking into it

PR #9869, opened 2026-07-22
feat(shortcuts): Cmd/Ctrl+Shift+N creates a child workspace of the active workspace

eezy0:
> **Stacked on #9859** — this branch is based on `feat/new-child-workspace-from-context-menu`, so the diff includes that PR's commit until it merges. Only the top commit (`feat(shortcuts): …`) is new here. GitHub does not allow a fork branch as the base of an upstream PR, hence `main` as the nominal base.

## Summary

- Adds a **Cmd/Ctrl+Shift+N** shortcut (`workspace.createChild`) that opens the New Workspace composer seeded with the **currently active workspace as lineage parent** — the keyboard counterpart of #9859's "New Child Workspace..." context-menu action. Cmd/Ctrl+N (`workspace.create`) keeps opening the plain composer.
- `Mod+Shift+N` was previously a secondary alias of `workspace.create` (kept for muscle-memory compatibility with the old Create-from shortcut). It is reassigned to the new action; when there is **no eligible active workspace** (nothing active, folder workspace, branchless checkout), the shortcut **falls back to the plain composer**, so users who relied on the alias see unchanged behavior exactly in the situations where a child create is impossible.
- Routed through the main-process window-shortcut policy like Cmd/Ctrl+N, so it fires even when focus is inside a browser guest or contentEditable surface.
- The command appears automatically in **Settings → Shortcuts** ("Create child worktree of active workspace", Global group) via the keybinding registry, with normal user rebinding support.
- The child-eligibility gate (`canCreateChildWorkspace`) moves from the context menu into `worktree-create-parent.ts` so both entry points share one rule.

## Screenshots

No new visual surface: the shortcut opens the same composer-with-lineage-hint shown in #9859's screenshots, and the Settings → Shortcuts row is auto-generated from the keybinding registry like every other command.

## Testing

- [x] `pnpm lint` — oxlint passes; same single pre-existing `lint:switch-exhaustiveness` failure as current `main` (untouched file).
- [x] `pnpm typecheck` — all three projects pass.
- [x] `pnpm test`
- [x] `pnpm build`
- [x] Added or updated high-quality tests that would catch regressions:
  - `window-shortcut-policy.test.ts`: `Mod+Shift+N` now resolves to `openNewChildWorkspace` on darwin and linux; the alt-chord rejection guard stays.
  - `useIpcEvents.test.ts`: the handler seeds `initialRepoId`/`parentWorktreeId` from the active worktree; falls back to the plain composer when nothing is active or the row is branchless/folder; no-ops while the composer is already open.
  - `worktree-create-parent.test.ts`: gating tests moved here alongside the shared `canCreateChildWorkspace`.
- [ ] Manual verification pending: ⌘⇧N happy path + fallback, ⌘N unchanged, new row in Settings → Shortcuts. (#9859's underlying flow was already manually verified on macOS.)

## AI Review Report

- **Cross-platform (macOS/Linux/Windows)**: no hand-rolled `metaKey`/`ctrlKey` branches — the binding is declared as `Mod+Shift+N` in the keybinding registry, which resolves Mod per platform in both the main-process resolver and the renderer matcher; layout-aware key matching comes with the existing infrastructure.
- **Behavior change audit**: the only regression surface is users pressing Mod+Shift+N expecting a plain create; the fallback covers every case where a child create is impossible, and when a parent IS eligible the composer still allows exactly the same inputs (name, Create-From sources) with one extra hint line, so no workflow is lost. User keybinding overrides are unaffected (new action id, existing ids keep their meaning).
- **SSH/remote**: the handler only reads renderer store state and opens the composer; parent eligibility and lineage recording reuse #9859's submit-time validation (repo/host checks), so remote workspaces behave identically to the context-menu path.
- **Dead-key/IME/browser-guest routing**: reuses the `before-input-event` allowlist path that `workspace.create` already exercises; no new event listeners in the renderer besides the IPC subscription (guarded for older preloads, mirroring `onDeleteCurrentWorkspace`).
- **Settings UI**: no manual list to update — `ShortcutsPane` renders from `KEYBINDING_DEFINITIONS`.

## Security Audit

- No new IPC surface beyond a one-way, argument-less `ui:openNewChildWorkspace` notification from main to renderer, matching the existing `ui:openNewWorkspace` channel pattern; the renderer resolves everything from its own store state.
- No shell, path, auth, or dependency changes.

## Notes

- Depends on #9859 for `parentWorktreeId` composer support and create-time lineage recording; merge that first, then rebase this to drop the shared commit.
- X handle: none — feel free to credit the GitHub handle instead.

### DIFF
--- a/src/main/ipc/worktree-create-lineage.test.ts
+++ b/src/main/ipc/worktree-create-lineage.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it, vi } from 'vitest'
+import { recordWorktreeLineageForCreatedWorktree } from './worktree-create-lineage'
+import { folderWorkspaceKey, worktreeWorkspaceKey } from '../../shared/workspace-scope'
+import type { WorktreeLineage } from '../../shared/types'
+
+function makeStore(metaById: Record<string, { instanceId?: string }>) {
+  return {
+    getWorktreeMeta: (worktreeId: string) => metaById[worktreeId],
+    setWorktreeLineage: vi.fn((_worktreeId: string, lineage: WorktreeLineage) => lineage)
+  }
+}
+
+const CHILD = { id: 'repo::child', instanceId: 'child-instance' }
+
+describe('recordWorktreeLineageForCreatedWorktree', () => {
+  it('records worktree lineage for a worktree-type parent workspace', () => {
+    const store = makeStore({ 'repo::parent': { instanceId: 'parent-instance' } })
+
+    const lineage = recordWorktreeLineageForCreatedWorktree(
+      store,
+      worktreeWorkspaceKey('repo::parent'),
+      CHILD,
+      1234
+    )
+
+    expect(lineage).toEqual({
+      worktreeId: 'repo::child',
+      worktreeInstanceId: 'child-instance',
+      parentWorktreeId: 'repo::parent',
+      parentWorktreeInstanceId: 'parent-instance',
+      origin: 'manual',
+      capture: { source: 'active-workspace', confidence: 'explicit' },
+      createdAt: 1234
+    })
+    expect(store.setWorktreeLineage).toHaveBeenCalledWith('repo::child', lineage)
+  })
+
+  it('records nothing for folder-workspace parents', () => {
+    const store = makeStore({})
+
+    expect(
+      recordWorktreeLineageForCreatedWorktree(store, folderWorkspaceKey('folder-1'), CHILD, 1)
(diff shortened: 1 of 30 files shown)


eezy0:
Folding this into #9859 directly — the shortcut is small enough to review alongside the feature it drives, and a fork-based stack can't express its real base branch on GitHub anyway. The commit now lives on the #9859 branch.

PR #9879, opened 2026-07-22
feat(right-sidebar): add expand toggle to widen the panel

szto:
## Summary

Adds an **expand toggle** to the right sidebar header. Clicking it widens the panel to 70% of the window width; clicking again restores the width the user had before. Dragging the resize handle while expanded exits expanded mode (the dragged width becomes the new baseline).

Expanded mode is **session-only** and never mutates the stored/persisted `rightSidebarWidth`: it only changes the *rendered* width. So toggling off — or restarting the app — restores the user's chosen width. The expand target is clamped to the same min/max bounds a manual drag respects, so the editor/terminal always keeps its reserved area and narrow windows fall back to the minimum sidebar width.

The new button sits next to the existing close button in both header layouts (top and side activity bar), reusing the existing `Tooltip` + `.sidebar-toggle` + `lucide-react` icon pattern. It shows `Maximize2` when collapsed and `Minimize2` when expanded.

## Screenshots

Right sidebar header, collapsed vs. expanded:

```
 collapsed:  [ Explorer  Agents  … ]           [ ⤢ ] [ ⇥ ]
 expanded :  [ Explorer  Agents  … ]           [ ⤡ ] [ ⇥ ]   ← panel now 70% of window
```

I could not capture a live screenshot in this environment (Electron desktop app). A short screen recording of the toggle/restore/drag-to-exit flow should be attached before merge.

## Testing

- [ ] `pnpm lint` — fails only on **pre-existing** `switch-exhaustiveness` errors in `src/main/daemon/daemon-server.ts` and `src/renderer/src/components/skills/skill-freshness-group.tsx` (untouched by this PR). oxlint/oxfmt on the changed files pass via the pre-commit hook, and the localization catalog verifier passes.
- [x] `pnpm typecheck` — passes (node + cli + web).
- [x] `pnpm test` — 32,975 passed. The 41 failures (7 files: `use-markup-draw-hint`, `pr-comment-presentation`, `pr-comments-list-selection`, `use-persisted-ai-vault-view-options`, `LinearAgentSkillSetupPrompt.reminder-toast`, …) are **pre-existing** localStorage/toast-timing environment failures reproducible on a clean checkout; **none** are in files this PR touches.
- [x] `pnpm build` (renderer bundle via `build:electron-vite`) — passes.
- [x] Added unit tests for the new `computeExpandedRightSidebarPanelWidth` helper (fraction target, wide/narrow-window clamping, non-DOM fallback) and updated the right-sidebar render-test store mock.

## AI Review Report

Reviewed the change end to end with the following focus:

- **Correctness / state**: Expanded mode is a single session-only boolean (`rightSidebarExpanded`); it derives the rendered width instead of writing it, so restore is lossless. `setRightSidebarWidth` resets the flag, which is the only setter used by the resize hook — verified there are no other callers that would leave the flag inconsistent. Fixed a sub-pixel floating-point result (`1440 * 0.7 = 1007.999…`) by rounding the target.
- **Cross-platform (macOS / Linux / Windows)**: No platform-specific code. The button reuses existing icon/tooltip primitives; no keyboard shortcut, accelerator, label, or path handling was added. Both header layouts (top and custom-desktop-chrome) get the button, and the `RIGHT_SIDEBAR_HEADER_NO_DRAG_CLASS_NAME` is applied so it stays clickable inside the draggable titlebar region on all platforms.
- **SSH / remote / local**: Purely renderer-side layout state; no process, file, credential, or network assumptions.
- **Performance**: Width is computed from existing memo-friendly inputs; no new subscriptions in hot paths. Expanded rendering reuses the existing clamp/`useWindowWidth` machinery, so window-resize handling is unchanged.
- **UI quality**: Follows `docs/STYLEGUIDE.md` — no new tokens, sizes, or colors; icon size and button class match the sibling close button.
- **i18n**: New `expandRightSidebar` / `collapseRightSidebar` keys added to all five locale catalogs (en, ko, ja, zh, es) with real translations; catalog parity verifier passes.

## Security Audit

No security-relevant surface touched. The change adds a client-side layout toggle only: no input handling, no command execution, no path handling, no auth/secret access, and no IPC changes. Persistence rides the existing debounced UI-state writer with no new fields (`rightSidebarExpanded` is intentionally not persisted). No new dependencies.

## Notes

- Expand target fraction is `RIGHT_SIDEBAR_EXPANDED_WINDOW_FRACTION = 0.7` in `right-sidebar-width.ts`, easy to tune.
- No git-provider-, agent-, or integration-specific behavior.
- X/Twitter handle: _(please add yours — I don't have it)_.

## ELI5

The right sidebar gets an expand toggle that temporarily widens it to most of the window. Toggle off or restart and your saved width comes back; dragging while expanded exits expand mode.

### DIFF
--- a/src/renderer/src/components/right-sidebar/index.tsx
+++ b/src/renderer/src/components/right-sidebar/index.tsx
@@ -1,6 +1,15 @@
 /* eslint-disable max-lines -- Why: the right sidebar owns activity-bar visibility, routing, and resize behavior as one interaction surface; splitting the tab table away would make hidden-tab fallbacks harder to audit. */
 import React, { useEffect, useMemo, useRef, useState } from 'react'
-import { Plug, Files, GitBranch, ListChecks, PanelRight, Workflow } from 'lucide-react'
+import {
+  Plug,
+  Files,
+  GitBranch,
+  ListChecks,
+  Maximize2,
+  Minimize2,
+  PanelRight,
+  Workflow
+} from 'lucide-react'
 import { useAppStore } from '@/store'
 import type { ActiveRightSidebarTab } from '@/store/slices/editor'
 import { useRepoById } from '@/store/selectors'
@@ -35,6 +44,7 @@ import {
 import {
   RIGHT_SIDEBAR_MIN_WIDTH,
   clampRightSidebarPanelWidth,
+  computeExpandedRightSidebarPanelWidth,
   computeMaxRightSidebarPanelWidth
 } from './right-sidebar-width'
 import { translate } from '@/i18n/i18n'
@@ -69,6 +79,8 @@ function RightSidebarInner(): React.JSX.Element {
   const setRightSidebarTab = useAppStore((s) => s.setRightSidebarTab)
   const showRightSidebarFiles = useAppStore((s) => s.showRightSidebarFiles)
   const toggleRightSidebar = useAppStore((s) => s.toggleRightSidebar)
+  const rightSidebarExpanded = useAppStore((s) => s.rightSidebarExpanded)
+  const toggleRightSidebarExpanded = useAppStore((s) => s.toggleRightSidebarExpanded)
   const checksStatus = useAppStore((s) => (s.rightSidebarOpen ? getActiveChecksStatus(s) : null))
   const activityBarPosition = useAppStore((s) => s.activityBarPosition)
   const setActivityBarPosition = useAppStore((s) => s.setActivityBarPosition)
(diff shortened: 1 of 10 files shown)


nwparker:

### What this image is
Current right sidebar / Source Control panel from Electron e2e — baseline for expand-to-widen toggle.

**Visual evidence**

| 12-source-control.png |
|---|
|  |

nwparker:
## Visual proof

The right sidebar is shown at its normal width, then after the expand control widens the File Explorer. The same E2E flow also toggled back to the restored state.

Before:

Normal sidebar width

After expanding:

Expanded sidebar width

nwparker:
ah sorry this might have got lost -- let me see whats up here

nwparker:
ahh Okay I see the PR 

It's valid, but it's probably something we'll punt on 🥲 

Why: Becasue the extra buttons can cause some confusion.
Happy to add a keyboard bind instead if that helps you! That way we won't take visual space and you can bind that key yourself. LMK

nwparker:
If you do want that just open a new issue ! Then I'll see it

PR #9882, opened 2026-07-22
perf(main): gate the git-common worktrees readdir while visible and idle

nwparker:
## What

**Stacked on #9849** (base = `nwparker/perf-s3-poller-gate`). #9849 parks the worktree pollers while the window is *hidden*; this trims their cost while the window is **visible and idle**.

The non-darwin (Windows/Linux) git-common poll re-enumerated the linked-worktrees directory (`readdir`) on every 2s tick even when no linked worktree was added or removed. This gates that `readdir` on the `worktrees/` directory signature: it only re-enumerates when a linked worktree is actually added/removed (or on the 15-tick backstop). Everything else — the `worktrees/` stat, each known entry's dir stat, **all structural leaves (`HEAD`, `gitdir`, `locked`, `config.worktree`) and `logs/HEAD`, and the primary-checkout leaves — is still stat'd every tick**, exactly as before, so detection latency is unchanged. `index` keeps its pre-existing 15-tick gate.

## ELI5

While Orca sat open and idle, it kept re-listing each repo's worktree folder twice a second even when no worktree was added or removed. Now it only re-lists when the folder actually changes; it still checks each worktree's HEAD/ref-log/metadata every tick, so nothing is noticed any slower.

## Proof of zero regression

> **Scope corrected after adversarial review** (which caught a real regression in the first cut): the initial version also gated the *structural leaves* behind the entry-dir signature, which would have deferred an in-place `HEAD`/`gitdir`/`config.worktree` rewrite (that doesn't bump the entry-dir mtime) to the ~30s backstop. Fixed by re-statting all structural leaves + `logs/HEAD` **every tick** (matching the pre-idle-gate poller) and gating **only** the `readdir`.

- Structural leaves + `logs/HEAD` + primary leaves are re-stat'd every tick → an in-place `HEAD`/branch/ref change surfaces within one tick, identical to before. New tests assert an in-place `HEAD` rewrite (no entry-dir change) is detected within a tick **without** triggering the gated `readdir`.
- A worktree add/remove bumps the `worktrees/` signature (now including `size`, so a same-granule coarse-timestamp change can't slip) → `readdir` re-runs → detected; the 15-tick backstop covers pathological cases; resume-after-hidden forces a full scan.
- macOS uses its native narrow watcher (untouched); this is the non-darwin path only.
- Tests: **53 pass** across the three watcher test files. `typecheck:node` / `oxlint` clean.

## Proof of perf improvement

While visible and idle, skips one `worktrees/` `readdir` per repo per 2s tick (each enumerating every linked worktree with `withFileTypes`), plus its per-entry `Dirent` processing — ~10 readdirs/s eliminated at 20 repos, more as linked-worktree counts grow. A smaller, provably-safe subset of the original idea (the structural-stat skipping was reverted as unsafe).

### DIFF
--- a/src/main/ipc/worktree-base-directory-poller.test.ts
+++ b/src/main/ipc/worktree-base-directory-poller.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, describe, expect, it, vi } from 'vitest'
-import { mkdtemp, mkdir, realpath, rm, stat, utimes, writeFile } from 'node:fs/promises'
+import { mkdtemp, mkdir, realpath, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import {
@@ -330,7 +330,7 @@ describe('worktree base directory poller', () => {
     )
   })
 
-  it('detects linked HEAD rewrites even when the entry directory mtime is restored', async () => {
+  it('detects an in-place linked HEAD rewrite that leaves the entry directory signature unchanged', async () => {
     const commonDir = await makeRoot()
     const entry = join(commonDir, 'worktrees', 'external-head')
     await mkdir(entry, { recursive: true })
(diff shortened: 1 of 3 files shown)


nwparker:
Adversarial review caught a real MEDIUM regression in the first cut and I fixed it: gating the structural leaves (`HEAD`/`gitdir`/`locked`/`config.worktree`) behind the entry-dir signature would defer an **in-place** HEAD rewrite (which doesn't bump the entry-dir mtime) to the ~30s backstop instead of 1 tick. Now all structural leaves + `logs/HEAD` are re-stat'd **every tick** (matching the pre-idle-gate poller); only the `worktrees/` `readdir` is gated, and `size` is back in the dir signature to close a coarse-granularity gap. Added a non-vacuous test (plain in-place HEAD rewrite, no `utimes`) asserting detection within one tick without a full scan. The win is smaller but provably zero-regression. 53 watcher tests pass.

PR #9884, opened 2026-07-22
perf(terminal): arm the delivery watchdog only while PTYs are attached

nwparker:
## What

The terminal-delivery freeze-detection watchdog armed a global 15s timer once at dispatcher setup and kept it armed **forever** — so even with **zero** attached PTYs it woke every 15s just to return (~0.067 renderer wakeups/sec of pure waste).

Fix: arm the 15s timer on the **first** attached/eager PTY and disarm it on the **last**. A new `syncTerminalDeliveryWatchdogTimer()` clears the timer (and resets the stall streak) when `hasAttachedPtys()` is false, and arms it (resetting the event-count baseline) when a PTY is attached and no timer is running. It's called after every PTY attach/detach/dispose/exit path in the dispatcher and transport.

## Important: detection is unchanged whenever a PTY is attached

The watchdog's invoke-IPC probe is the **only independent recovery lane** for a main→renderer push-channel wedge (which can start on the very next unseen byte), so it is emphatically **not** visibility-gated or backed off while PTYs are attached — the only change is that the timer doesn't run when there are zero PTYs to watch.

## ELI5

Orca has a smoke detector that checks every 15 seconds that terminal output is still flowing. It was left running even when there were no terminals open at all. Now it turns on when you open a terminal and off when the last one closes — and while any terminal is open it watches exactly as vigilantly as before.

## Proof of zero regression

- `hasAttachedPtys = ptyDataHandlers.size > 0 || eagerPtyHandles.size > 0`; I verified **every** writer of those maps (transport register, eager register, failed-shutdown restore) is followed by a `sync`, and every detach/dispose/exit path syncs after the delete — so the timer is armed whenever ≥1 PTY is attached and only cleared at zero.
- Arming resets `eventCountAtLastTick` so the first tick measures the correct window; the E2E `configure` now re-schedules only when currently armed.
- Detection cadence + heal logic while attached are byte-for-byte unchanged.
- Tests: **85 pass** across `terminal-delivery-watchdog.test.ts` + `pty-dispatcher-push-reattach.test.ts` + `pty-transport.test.ts` — zero-timer with no PTYs, arm on first attach, multiple attached/eager, 15s invoke cadence preserved, disarm on last detach/dispose, and the #7894 detach/attach identity-race path. `typecheck:web` clean. (An independent adversarial review is also running.)

## Proof of perf improvement

Removes 0.067 renderer wakeups/sec whenever no PTY is attached (e.g. an editor-only / dashboard-only window), with zero change to freeze-detection while terminals are open — the tests assert no timer is armed with zero PTYs.

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-data-sidecar-subscriptions.ts
+++ b/src/renderer/src/components/terminal-pane/pty-data-sidecar-subscriptions.ts
@@ -1,5 +1,6 @@
 import { acquirePtyDeliveryInterest } from './pty-delivery-interest'
 import { ensurePtyDispatcher, ptyDataSidecars } from './pty-dispatcher'
+import { syncTerminalDeliveryWatchdogTimer } from './terminal-delivery-watchdog'
 
 /** Register a side-channel data watcher for a PTY without taking ownership
  *  of the primary handler. Returns an unsubscribe fn. */
@@ -15,6 +16,8 @@ export function subscribeToPtyData(ptyId: string, watcher: (data: string) => voi
     ptyDataSidecars.set(ptyId, set)
   }
   set.add(watcher)
+  // Why: a sidecar receives pushed bytes with no primary handler, so it must arm the delivery watchdog.
+  syncTerminalDeliveryWatchdogTimer()
   return () => {
     releaseDeliveryInterest()
     const current = ptyDataSidecars.get(ptyId)
@@ -25,5 +28,6 @@ export function subscribeToPtyData(ptyId: string, watcher: (data: string) => voi
     if (current.size === 0) {
       ptyDataSidecars.delete(ptyId)
     }
+    syncTerminalDeliveryWatchdogTimer()
   }
 }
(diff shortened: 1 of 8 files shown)


nwparker:
Adversarial review caught a **BLOCKER** and I fixed it: `hasAttachedPtys` omitted `ptyDataSidecars`, so a **parked terminal** (which receives pushed bytes via a sidecar with no primary handler — also draft-readiness / automation observers) could leave the watchdog disarmed while main was still delivering bytes, silently dropping parked-status/BEL/completion bytes with no recovery. Fixed by including `ptyDataSidecars.size` in the attach predicate and calling `syncTerminalDeliveryWatchdogTimer()` on sidecar subscribe/unsubscribe. Added an integration test proving a sidecar-only PTY arms the real watchdog and disarms on the last unsubscribe. 91 tests pass; typecheck:web clean. (Reviewer also confirmed all primary/eager attach paths sync correctly and the zero→reattach streak reset is a bounded latency tradeoff, not a blocker.)

nwparker:
Two follow-up findings from the review also fixed (both interacting with the arm/disarm streak dynamics):

- **Streak preserved across a detach/reattach flicker** — the park path no longer resets `stallStreakTicks`, so a wedge mid-confirmation doesn't lose progress on a brief zero-attachment flicker (matching the old always-armed timer); a healthy tick after re-arm clears it.
- **Async-race guard** — an in-flight tick whose `reportRendererDeliveryState` invoke is awaited across a last-PTY detach could resurrect a stale streak while parked. Added a `watchdogGeneration` counter (bumped whenever a live timer is torn down); the tick captures it before the await and discards its result if it changed, so the racing tick contributes nothing. New test covers it.

92 tests pass; typecheck:web clean.

PR #9894, opened 2026-07-22
fix(github): show current PR check status

### LINKED ISSUE #9864: [Bug]: PR list check badge counts historical failed runs, not current check status
### Operating system

Linux

### Orca version

v1.4.146

### Details

Short summary:
In the PR list, the checks badge shows "N failing" even when every check currently passes on GitHub (merge box green). N appears to equal the number of *previously failed* workflow runs on the head commit, not the number of currently failing checks.

What happened?

We run a label-gated required check (fails on PR open, passes after a label is added — same head SHA). After the check passes, GitHub's merge box is green and `gh pr checks` shows no failures, but Orca's PR list badge permanently shows "1 failing". A PR that went through the fail→pass cycle twice shows "2 failing". The badge only clears when a new commit changes the head SHA.

How can we reproduce it?
1. Add a workflow triggered on `pull_request: [opened, labeled]` whose job fails when a label is absent and passes when it is present.
2. Open a PR → the run fails.
3. Add the label → a new run on the same head SHA passes. GitHub now shows all checks green.
4. Orca's PR list still shows "1 failing". Remove and re-add the label to repeat the cycle and the badge becomes "2 failing".

What we think is going on:

Each event-triggered workflow run creates a new check suite on the same head commit, and `statusCheckRollup` keeps the check runs of *all* suites. Verified on an affected PR (`gh pr view --json statusCheckRollup`):

```
name         conclusion  completedAt
merge-label  FAILURE     02:08   <- run triggered by "opened"
merge-label  SUCCESS     02:21   <- run triggered by "labeled"
(all other checks SUCCESS/SKIPPED)
```

`deriveWorkItemCheckSummary()` in `src/main/github/client.ts` counts every rollup entry without deduplicating by check name, so the stale FAILURE entry keeps the summary in `failure` state forever. GitHub's own UI shows only the latest run per check, which is why the two disagree. The details-panel path (`mapGraphQLPRChecksResponse`) also keeps all check-run contexts, so the same stale entries can show up there.

Expected: the badge reflects the latest conclusion per check context (e.g. dedupe rollup entries by name, keeping the most recent), matching GitHub's merge box.

Happy to provide more repro details if useful.

OnlyYu1996:
## Summary

- Deduplicate GitHub check rollup entries by check type, name, and provider, keeping the newest run.
- Apply the same normalization to PR list badges and check details while preserving pending approval suites.
- Add regression coverage for repeated fail/pass cycles, missing timestamps, provider collisions, legacy statuses, and historical suites.

Closes #9864

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/github/github-check-rollup.test.ts src/main/github/client-pr-checks.test.ts src/main/github/client.test.ts` (147 tests)
- [x] `pnpm run build:desktop`
- [x] Targeted oxlint, formatting, max-lines ratchet, and reliability gate checks

## AI Review Report

Reviewed check identity, timestamp precedence, missing metadata, provider collisions, CheckRun versus StatusContext collisions, and pending approval behavior. The review caught that suite coverage must use the full unfiltered rollup so a historical suite with a run cannot reappear as pending; the implementation and regression test cover that case.

Cross-platform compatibility was checked for macOS, Linux, and Windows. This change is limited to GitHub provider response mapping and adds no shortcuts, labels, paths, shell behavior, or Electron-specific platform behavior. SSH and remote use the same provider result normalization.

## Security Audit

No command execution, file path handling, authentication, secrets, dependencies, or IPC contracts were added or changed. The additional GraphQL fields are non-sensitive timestamps and provider metadata. Unknown API values remain defensively validated before normalization. No follow-up is required.

## Notes

- `pnpm lint` is currently blocked by the existing non-exhaustive switch at `src/renderer/src/components/skills/skill-freshness-group.tsx:101`; all touched files pass oxlint.
- The full `pnpm test` suite was not run; 147 related GitHub tests passed.
- `pnpm build` was not run because it includes platform-native artifacts; `pnpm run build:desktop` passed.
- X handle: N/A

## ELI5

PR check badges could show an old failed run after a newer pass because duplicates were not collapsed by name. Orca keeps the newest run per check so the status matches GitHub today.

### DIFF
--- a/src/main/github/client-pr-checks.test.ts
+++ b/src/main/github/client-pr-checks.test.ts
@@ -134,23 +134,37 @@ function graphQLCheckRun(
     conclusion: string | null
     detailsUrl: string | null
     url: string | null
+    startedAt: string | null
+    completedAt: string | null
     checkSuiteId: number | null
     workflowRunId: number | null
+    workflowName: string | null
+    appSlug: string | null
   }> = {}
 ): Record<string, unknown> {
   const checkSuiteId = overrides.checkSuiteId ?? 1000
-  const workflowRunId = overrides.workflowRunId ?? 1
+  const workflowRunId = overrides.workflowRunId === null ? null : (overrides.workflowRunId ?? 1)
   return {
     __typename: 'CheckRun',
     databaseId: overrides.databaseId ?? 88,
     name: overrides.name ?? 'build',
     status: overrides.status ?? 'COMPLETED',
     conclusion: overrides.conclusion ?? 'SUCCESS',
+    startedAt: overrides.startedAt ?? '2026-07-22T01:00:00Z',
+    completedAt: overrides.completedAt ?? '2026-07-22T01:01:00Z',
     detailsUrl: overrides.detailsUrl ?? 'https://github.com/acme/widgets/actions/runs/1',
     url: overrides.url ?? 'https://github.com/acme/widgets/runs/88',
     checkSuite: {
       databaseId: checkSuiteId,
-      workflowRun: workflowRunId === null ? null : { databaseId: workflowRunId }
+      app: overrides.appSlug === null ? null : { slug: overrides.appSlug ?? 'github-actions' },
+      workflowRun:
+        workflowRunId === null
+          ? null
+          : {
+              databaseId: workflowRunId,
+              workflow:
+                overrides.workflowName === null ? null : { name: overrides.workflowName ?? 'CI' }
+            }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


OnlyYu1996:
Addressed the review feedback in `45cfaa2e3`:

- Preserved explicit `workflowRunId: null` in the GraphQL test fixture and added coverage for check runs without workflow metadata.
- Documented timestamp tie-breaking, conservative unknown-conclusion handling, and why pending-suite coverage uses the undeduplicated contexts.

Validation: 147 targeted GitHub tests, `typecheck:node`, touched-file Oxlint/Oxfmt, and max-lines ratchet pass.

PR #9906, opened 2026-07-22
fix(kimi): use native Windows hook launcher

### LINKED ISSUE #9834: [Bug]: Kimi Code agent-status hooks never fire on Windows (POSIX `/bin/sh` hook vs. Kimi's `cmd.exe` executor)
### Operating system

Windows

### Orca version

v1.4.148

## Summary

Orca's agent-status integration works for Claude and Codex but **never shows status for Kimi Code** on Windows. Root cause: Orca injects a **POSIX `sh`** hook command into `~/.kimi-code/config.toml`, but **Kimi Code executes hook `command` strings through Windows `cmd.exe`**, which cannot parse that syntax. The hook errors out and nothing is POSTed to Orca's hook endpoint.

Kimi is the only agent Orca configures this way. Every other Windows agent gets a `cmd.exe`/PowerShell launcher (e.g. `gemini-hook.cmd` is invoked via an explicit `powershell.exe -EncodedCommand …` wrapper; `codex-hook.cmd`/`claude-hook.cmd` are `.cmd` files). Kimi alone gets `kimi-hook.sh` invoked via `/bin/sh`.

## What Orca writes (managed block in `~/.kimi-code/config.toml`)

```toml
# >>> orca-managed-kimi-hooks (managed by Orca; do not edit) >>>
[[hooks]]
event = "UserPromptSubmit"
command = "if [ -f 'C:/Users/<user>/.orca/agent-hooks/kimi-hook.sh' ] && [ -r '…' ] && [ -x '…' ]; then /bin/sh 'C:/Users/<user>/.orca/agent-hooks/kimi-hook.sh'; else cat >/dev/null 2>&1 || :; fi"
timeout = 10
# … same command for PreToolUse, PostToolUse, PostToolUseFailure, PermissionRequest, Stop, StopFailure …
```

## Evidence that Kimi runs hooks via `cmd.exe`

Temporary probe hook `command`:

```
echo [probe] comspec=%COMSPEC% arg0=$0 shellpid=$$ > kimi_probe.log
```

Output after a Kimi turn:

```
[probe] comspec=C:\WINDOWS\system32\cmd.exe arg0=$0 shellpid=$$
```

`%COMSPEC%` was **expanded** (→ `cmd.exe`) while `$0`/`$$` stayed literal ⇒ Kimi ran the command through `cmd.exe`, not a POSIX shell. (Consistent with Node `child_process` `{shell:true}` using `process.env.ComSpec` on win32.)

Running Orca's exact command under `cmd.exe`:

```
> cmd /d /s /c "if [ -f 'C:/…/kimi-hook.sh' ] && … ; then /bin/sh '…'; else cat >/dev/null 2>&1 || :; fi"
-f was unexpected at this time.
(exit code 1)
```

So the hook is a no-op every time, on every event.

## Secondary defect (`kimi-hook.sh` would also fail even under `sh`)

`kimi-hook.sh` does `. "$ORCA_AGENT_HOOK_ENDPOINT"` to load `ORCA_AGENT_HOOK_PORT`/`_TOKEN`. On Windows that endpoint file is `…/Roaming/Orca/agent-hooks/endpoint.cmd`, which is **cmd batch** (`set ORCA_AGENT_HOOK_PORT=3920`). POSIX `sh` cannot source `set VAR=value`, so even if the hook were run under `sh`, it would exit without the port/token unless those vars are also present directly in the process env.

## Suggested fix

For Kimi on the Windows **host** runtime, mirror what Orca already does for the other Windows agents:

1. Emit a `kimi-hook.cmd` (a `cmd.exe` launcher that `call`s the endpoint and `curl`s `/hook/kimi`), analogous to `codex-hook.cmd`.

2. Write the `config.toml` `command` in `cmd.exe` form, e.g.

   ```toml
   command = 'if exist "C:\Users\<user>\.orca\agent-hooks\kimi-hook.cmd" ("C:\Users\<user>\.orca\agent-hooks\kimi-hook.cmd") else (more.com>nul 2>&1)'
   ```

   (or invoke via an explicit `powershell.exe -EncodedCommand …` wrapper, as done for Gemini).

Only the WSL/POSIX runtime should get the `/bin/sh` + `.sh` variant.

OnlyYu1996:
## Summary

- Use a native `kimi-hook.cmd` launcher for local Kimi Code hooks on Windows.
- Refresh the current Orca hook endpoint before posting status events, and preserve the shared Windows stdin-drain contract.
- Keep macOS, Linux, WSL, and SFTP remote installs on the existing POSIX `kimi-hook.sh` path.

Closes #9834

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` (blocked by the existing non-exhaustive switch in `src/renderer/src/components/skills/skill-freshness-group.tsx:101`)
- [x] `pnpm typecheck`
- [ ] `pnpm test` (44 targeted tests passed; 1 Windows-only live-process test skipped on macOS)
- [ ] `pnpm build` (`pnpm build:desktop` passed)

Targeted coverage includes Kimi local installation, Kimi TOML mutation, remote hook installers, managed-hook timeouts, and managed-hook stdin lifecycle behavior.

## AI Review Report

Reviewed the platform selection, generated TOML command, encoded Windows launcher, managed-command detection, endpoint refresh, and stdin ownership paths. The review verified that local Windows installs generate only `kimi-hook.cmd`, while macOS/Linux and SFTP remote installs retain `kimi-hook.sh`. It also checked that the change introduces no UI shortcut, label, path-separator, or Electron platform behavior changes outside the hook installer.

## Security Audit

No dependencies, IPC messages, protocols, or auth flows were added. The Windows command reuses the existing UTF-16LE PowerShell encoded launcher so user-profile paths are not interpolated into a raw `cmd.exe` command. The batch script continues to use fully qualified Windows system binaries, loopback-only HTTP, the existing hook token header, bounded curl timeouts, and stdin streaming rather than placing payloads on the command line. Remote paths continue to use POSIX path construction and the existing SFTP atomic writer.

## Notes

The Windows live-process lifecycle case remains covered by the existing Windows-only test and is skipped when the suite runs on macOS.
- X handle: N/A

## ELI5

Kimi Code hooks on Windows failed because they used a POSIX shell script. Local Windows installs now use a native kimi-hook.cmd; other platforms keep the existing shell launcher.

### DIFF
--- a/src/main/agent-hooks/managed-hook-stdin-lifecycle.test.ts
+++ b/src/main/agent-hooks/managed-hook-stdin-lifecycle.test.ts
@@ -254,7 +254,7 @@ describe('Windows managed hook stdin structure', () => {
         (name) => name.endsWith('-hook.cmd') && !name.startsWith('antigravity-')
       )
       mainBatchScripts.push('antigravity-hook.cmd')
-      expect(mainBatchScripts).toHaveLength(10)
+      expect(mainBatchScripts).toHaveLength(11)
       for (const fileName of mainBatchScripts) {
         const script = readFileSync(join(hooksDir, fileName), 'utf8')
         expect(script, `${fileName} port guard`).toContain(
@@ -278,10 +278,6 @@ describe('Windows managed hook stdin structure', () => {
       const copilot = readFileSync(join(hooksDir, 'copilot-hook.ps1'), 'utf8')
       expect(copilot.indexOf('[Console]::In.ReadToEnd()')).toBeLessThan(
         copilot.indexOf('if (-not $env:ORCA_AGENT_HOOK_PORT')
-      )
-      const kimi = readFileSync(join(hooksDir, 'kimi-hook.sh'), 'utf8')
-      expect(kimi.indexOf(`payload=$(${POSIX_HOOK_STDIN_READER})`)).toBeLessThan(
-        kimi.indexOf('exit 0')
       )
     } finally {
       homedirMock.mockImplementation(() => process.env.HOME ?? tmpdir())
--- a/src/main/agent-hooks/remote-hook-service-installers.test.ts
+++ b/src/main/agent-hooks/remote-hook-service-installers.test.ts
@@ -176,6 +176,11 @@ describe('remote hook service installers', () => {
         {
           path: '/home/dev/.orca/agent-hooks/droid-hook.sh',
           install: (sftp: SFTPWrapper) => new DroidHookService().installRemote(sftp, '/home/dev')
+        },
+        {
+          // Why: Kimi's local script selection is platform-branched; pin that a Windows host still writes the POSIX remote body.
+          path: '/home/dev/.orca/agent-hooks/kimi-hook.sh',
+          install: (sftp: SFTPWrapper) => new KimiHookService().installRemote(sftp, '/home/dev')
         }
       ]
 
--- a/src/main/kimi/hook-service.test.ts
+++ b/src/main/kimi/hook-service.test.ts
@@ -1,4 +1,4 @@
-import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, beforeEach, describe, expect, it } from 'vitest'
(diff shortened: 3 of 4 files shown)


AmethystLiang (approved):
Reviewed via multi-agent code review (design + Windows correctness). Windows path independently confirmed: stdin always drains (curl payload@- on the happy path, more.com on the guard-fail path), path quoting is robust, and the cross-platform script switch is clean. installRemote correctly stays POSIX-independent of local platform.

Pushed two follow-up nits directly onto the branch (255ff926): restored the English installRemote rationale comment (codebase convention) and added Kimi to the win32 remote-POSIX parity test. Ready to merge.

PR #9908, opened 2026-07-22
fix(editor): open external SSH file links

### LINKED ISSUE #9743: [Bug]: SSH remote workspace: terminal link to out-of-worktree file (e.g. /tmp/*) fails with "External local files are not available for remote workspaces"
**Operating system:** macOS

**Orca version:** 1.4.147

## Details

Short summary:
In an SSH remote workspace, clicking a terminal file link that points outside the worktree (e.g. an agent writes `/tmp/report.md` on the remote host and prints the path) fails with:

```
Unable to load file
External local files are not available for remote workspaces.
```

The path exists on the remote host and the SSH read path already supports absolute paths; the renderer rejects before attempting the read.

How to reproduce:
1. Open a project on a remote host via SSH.
2. In a terminal/agent session in that worktree: `echo hi > /tmp/orca-repro.txt && echo /tmp/orca-repro.txt`
3. Hover the printed path in terminal output, click the file link.
4. Editor tab opens with `Unable to load file / External local files are not available for remote workspaces.` Retry does not help.

Expected: file is read from the remote host over the worktree's existing SSH connection, same as worktree files.

## Analysis (from the shipped bundle, v1.4.147)

`loadFileContent` (`out/renderer/assets/EditorPanel-*.js`) classifies a file as external when `relativePath === filePath`, then unconditionally throws for remote workspaces:

```js
if (restoredOpenFile?.filePath === filePath && restoredOpenFile.relativePath === filePath) {
  if (readSettings?.activeRuntimeEnvironmentId?.trim() || connectionId) {
    throw new Error("External local files are not available for remote workspaces.");
  }
  await window.api.fs.authorizeExternalPath({ targetPath: filePath });
}
```

The throw fires even when `getConnectionIdForFile(worktreeId, filePath)` resolved the worktree's SSH `connectionId` — i.e. the path is already attributed to the remote host, not the local machine, so "external **local** file" is a misclassification.

The main-process handler has no worktree restriction for SSH reads:

```js
ipcMain.handle("fs:readFile", async (_event, args) => {
  if (args.connectionId) {
    const provider = requireSshFilesystemProvider(args.connectionId);
    return provider.readFile(args.filePath); // absolute paths OK
  }
  ...
```

So `window.api.fs.readFile({ filePath: "/tmp/...", connectionId })` would succeed; the renderer guard rejects before the call is made.

## Suggested fix

When the external path resolves a `connectionId` (SSH workspace): skip the throw and `authorizeExternalPath` (local-sandbox concern only) and fall through to `window.api.fs.readFile({ filePath, connectionId })`. Keep the refusal for runtime-environment workspaces, where the `files.read` RPC is worktree-scoped (`canReadRelativeRuntimeFile`), or extend that RPC to accept host-absolute paths.

Related: #1661 — out-of-workspace SSH reads were previously unblocked in the main process for symlinked paths; this is the renderer-side equivalent for terminal-link opens.

OnlyYu1996:
## Summary

- Let editor tabs read host-absolute SSH paths through their resolved SSH connection.
- Skip local filesystem authorization only when a concrete SSH `connectionId` owns the path.
- Preserve the existing local external-path authorization and remote runtime worktree boundary.

Closes #9743

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` (blocked by the existing non-exhaustive switch in `src/renderer/src/components/skills/skill-freshness-group.tsx:101`)
- [x] `pnpm typecheck`
- [ ] `pnpm test` (171 related tests passed)
- [ ] `pnpm build` (`pnpm build:desktop` passed)

The focused coverage verifies external SSH reads, local external-path authorization, remote runtime rejection, terminal file-link routing, runtime file reads, and connection ownership.

## AI Review Report

Reviewed the terminal-link-to-editor path, owner hydration guard, per-file SSH connection resolution, runtime owner selection, and local authorization flow. The review verified that macOS, Linux, and Windows all use the same provider-aware renderer branch; no keyboard shortcut, label, path separator, or Electron platform behavior changed. It also confirmed that an unresolved SSH owner still fails retryably instead of falling back to a local read.

## Security Audit

No dependencies, IPC messages, protocols, command execution, auth, or secret handling changed. Host-absolute paths bypass local authorization only when `getConnectionIdForFile` has resolved a concrete SSH owner; the read then goes through the existing SSH filesystem provider and connection boundary. Local paths still require `fs:authorizeExternalPath`, and paired runtime paths remain constrained to the owning worktree RPC.

## Notes

This change intentionally does not extend paired runtime file RPCs to host-absolute paths.
- X handle: N/A

## ELI5

Clicking a file link to an absolute path on an SSH host failed local authorization. Editor tabs can open those SSH paths through the owning connection while local path rules stay the same.

### DIFF
--- a/src/renderer/src/components/editor/useEditorPanelContentState.test.tsx
+++ b/src/renderer/src/components/editor/useEditorPanelContentState.test.tsx
@@ -14,7 +14,8 @@ const mocks = vi.hoisted(() => ({
   getConnectionId: vi.fn(),
   getConnectionIdForFile: vi.fn(),
   isWorktreeConnectionResolved: vi.fn(() => true),
-  getState: vi.fn()
+  getState: vi.fn(),
+  authorizeExternalPath: vi.fn()
 }))
 
 vi.mock('@/runtime/runtime-file-client', () => ({
@@ -145,6 +146,12 @@ describe('useEditorPanelContentState', () => {
       openFiles: [],
       setLastKnownDiskSignature: vi.fn()
     })
+    mocks.authorizeExternalPath.mockReset()
+    mocks.authorizeExternalPath.mockResolvedValue(undefined)
+    Object.defineProperty(window, 'api', {
+      configurable: true,
+      value: { fs: { authorizeExternalPath: mocks.authorizeExternalPath } }
+    })
   })
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


OnlyYu1996:
Closing as superseded by #10455, which has already merged the same resolved-SSH-owner routing and adds the stamped-host, live-tail, and remote-runtime safeguards needed by the current editor path. There is no remaining unique change from this branch to rebase safely.

PR #9913, opened 2026-07-22
Fix legacy worktree lineage projection after stable updates

OrcaWin:
## Summary

- restore desktop nesting when a resolved worktree carries valid exact-instance legacy `lineage` but the separately hydrated lineage side-map has no child entry
- keep the side-map authoritative when it does contain the child, so stale same-path/replaced-instance records cannot fall back to inline data
- use one validated projection contract for grouping, filtering, reveal/send targets, unlink/reparent actions, drag/drop, deletion, folder descendants and PR aggregation, desktop/runtime folder hydration, and mobile summaries

## Deterministic reproduction and root cause

The motivating stable state is modeled with the production identities:

- parent instance `b0ffd635-91cd-424f-b804-80d4bb277a4c`
- child 9276 instance `1ceb9823-aa98-4f79-8eaa-af0b3a3d551b`
- child 9744 instance `87e2ef9a-99d3-48e3-9a53-3d1a979b5417`

Both older children have mutually consistent persisted/runtime legacy lineage to `assigned-issues`, but `workspaceLineage: null`. A newly created child has both legacy and generalized workspace lineage. The narrow renderer fixture passes exact valid inline legacy lineage with an absent side-map entry; before this fix `buildRows` emits each child at depth 0, and after it emits the child beneath the exact parent.

The semantic mismatch was that desktop grouping trusted only the separately hydrated `worktreeLineageById` map even though resolved worktrees already carried authoritative instance-aware legacy `lineage`. Stable hydration could preserve and report lineage correctly while the renderer discarded it. The fix selects a side-map entry when the child key exists and otherwise falls back to resolved inline legacy lineage. Bare `parentWorktreeId` is never trusted.

The sidebar grouping/visibility files in `v1.4.149` and `v1.4.150` are byte-identical. The `v1.4.149 -> v1.4.150` update exposed the mismatch for these workspaces, but the evidence does not support naming that update as the exact introducing commit.

## Safety, security, and compatibility

Projected edges require matching child/parent record IDs, exact child and parent instance IDs, the same repository, and matching host/project identities when both sides know them. Missing parents, stale instances, disagreement with an existing side-map entry, self-links, and cycles stay flat/rejected. Runtime manual reparenting enforces the same boundary before persistence.

The repair does not migrate or rewrite lineage metadata and adds no new external-input or privilege boundary. It remains path-agnostic. Local, SSH, WSL, remote-runtime, multi-host/project, folder workspace, imported worktree, pinned row, filtering/sorting, hidden-child counts, large-list virtualization, mobile summaries, and intentionally flat Kanban behavior are covered or preserved. Graph validation is precomputed for bulk surfaces; drag checks validate only the proposed parent, and folder traversal uses adjacency plus BFS.

No production metadata was rewritten or manually reparented. The running production CLI still reports both affected children under the original parent with the exact instance IDs above.

## Before / after evidence

- before: latest stable rendered both affected children as independent roots immediately above `assigned-issues`
- after: the fixed dev sidebar renders legacy inline-only lineage beneath the exact parent; [screenshot evidence](https://github.com/stablyai/orca/pull/9913#issuecomment-5044582809)
- attached screenshot SHA-256: `71474a32fa6954ae1d77bfbc2c31c30710b93e3212cb9c4863007bf052c58af4`

## Validation

- 2,773 tests across 185 shared/runtime/store/sidebar files during the full investigation
- 1,520 final relevant tests across shared projection, runtime, IPC/Windows, store, grouping/filtering, cards, folder hydration/PR aggregation, host boundaries, Kanban, and virtualization
- full node, CLI, and web typecheck
- touched-file oxlint, React Doctor, and formatting
- reliability-gate manifest and max-lines ratchet
- 5 Electron lineage E2Es at reviewed head `641799fa2`, including legacy-only inline fallback and filtered-parent structural injection

## Internal review

Iterative adversarial review found and fixed boundary/cycle, filtered-parent, unlink/reveal/delete/drag, folder PR aggregation, folder detected-hydration, and large-list performance gaps. The final independent round reported zero Critical, Important, or Minor findings and independently passed 1,191 targeted tests.

## Limitation

Full `pnpm lint` reaches an unrelated pre-existing switch-exhaustiveness failure at `src/renderer/src/components/skills/skill-freshness-group.tsx:101`. All changed files pass lint.

### DIFF
--- a/src/main/ipc/worktrees.test.ts
+++ b/src/main/ipc/worktrees.test.ts
@@ -2367,6 +2367,66 @@ describe('registerWorktreeHandlers', () => {
     ])
   })
 
+  it('hydrates folder-repo detected rows with instance-validated legacy lineage', async () => {
+    const folderRepo = {
+      id: 'repo-1',
+      path: '/workspace/folder',
+      displayName: 'folder',
+      badgeColor: '#000',
+      addedAt: 0,
+      kind: 'folder' as const
+    }
+    const parentId = `${folderRepo.id}::${folderRepo.path}`
+    const childId = `${parentId}::workspace:child-instance`
+    const metaById: Record<string, Record<string, unknown>> = {
+      [parentId]: makeWorktreeMeta({
+        instanceId: 'parent-instance',
+        projectId: 'repo:repo-1',
+        hostId: 'local',
+        projectHostSetupId: 'repo-1'
+      }),
+      [childId]: makeWorktreeMeta({
+        instanceId: 'child-instance',
+        projectId: 'repo:repo-1',
+        hostId: 'local',
+        projectHostSetupId: 'repo-1'
+      })
+    }
+    store.getRepos.mockReturnValue([folderRepo])
+    store.getRepo.mockReturnValue(folderRepo)
+    store.getAllWorktreeMeta.mockReturnValue(metaById)
+    store.getWorktreeMeta.mockImplementation((worktreeId: string) => metaById[worktreeId])
+    store.getAllWorktreeLineage.mockReturnValue({
+      [childId]: {
+        worktreeId: childId,
+        worktreeInstanceId: 'child-instance',
+        parentWorktreeId: parentId,
+        parentWorktreeInstanceId: 'parent-instance',
+        origin: 'cli',
+        capture: { source: 'explicit-cli-flag', confidence: 'explicit' },
+        createdAt: 1
+      }
+    })
+
+    const result = (await handlers['worktrees:listDetected'](null, {
+      repoId: folderRepo.id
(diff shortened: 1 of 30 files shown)


OrcaWin:
Fresh `internal-review-until-clean` pass completed at `641799fa2`. One new adversarial round reviewed the full diff, including elegance and performance interrogations, and found no remaining in-scope issues. Fresh verification: 12 targeted lineage/runtime/sidebar suites, 1,444 tests; full typecheck; changed-file oxlint/format; React Doctor; reliability gates; and max-lines ratchet all pass. No code changes were required.

Issue #9921, opened 2026-07-22
Sparse badge sticks after `git sparse-checkout disable` (detection reads only pattern-file presence)


artile:
## Summary

A worktree that was sparse-checked-out and later fully restored with `git sparse-checkout disable` keeps showing the **`sparse`** badge and the tooltip *"Partial checkout. Files outside these paths are not on disk."* — even though every file is on disk and `core.sparseCheckout=false`.

## Root cause

`detectSparseCheckout` (`src/main/git/worktree.ts`) infers sparsity purely from the existence + non-zero size of `<gitdir>/info/sparse-checkout`:

```ts
const stats = await stat(join(gitDir, 'info', 'sparse-checkout'))
return stats.isFile() && stats.size > 0
```

The comment states the assumption: *"Git writes this file when sparse checkout is enabled … and does not write it otherwise."* That assumption is wrong. `git sparse-checkout disable` sets `core.sparseCheckout=false` and repopulates the working tree, but **deliberately leaves `info/sparse-checkout` in place** so the checkout can be re-enabled with the same patterns. The fast-path probe from #1290 therefore reports a false positive.

## Reproduce

```sh
git init repo && cd repo
mkdir keep drop && touch keep/f drop/f && git add -A && git commit -m init
git sparse-checkout set keep     # core.sparseCheckout=true  -> Orca: sparse (correct)
git sparse-checkout disable      # core.sparseCheckout=false, full tree restored
ls .git/info/sparse-checkout     # still present  -> Orca: still "sparse" (WRONG)
```

Ground truth after `disable`: `core.sparseCheckout=false`, full working tree, `0` skip-worktree entries — yet the badge and "files are not on disk" tooltip persist.

## Fix

Keep the cheap `fs.stat` as a fast-path gate, but when the pattern file is present, confirm `core.sparseCheckout` is actually enabled by reading git's config files directly (shared repo `config` and, under `extensions.worktreeConfig`, the per-worktree `config.worktree`, honoring precedence). No subprocess, and it only runs for the rare worktree that still has a non-empty pattern file — so it does not reintroduce the per-poll `git sparse-checkout list` fan-out #1290 removed.

PR: tao-io/orca `fix/sparse-checkout-false-positive-after-disable` (adds a real-git regression test + config-parser unit tests).

## Related (separate, not in this PR)

[...]

PR #9922, opened 2026-07-22
fix(worktree): don't flag a disabled sparse checkout as sparse

artile:
## Summary

Fixes a worktree falsely showing the **`sparse`** badge (and the "Files outside these paths are not on disk" tooltip) after `git sparse-checkout disable` had already restored every file to disk.

## ELI5

If you turned on a "only show some files" mode and then turned it back off, the app kept warning that files were missing — even though they were all right there. Now it checks the real setting and shows the correct state.

## Root cause & fix

`detectSparseCheckout()` inferred sparsity purely from the presence and non-zero size of `<gitdir>/info/sparse-checkout`. But `git sparse-checkout disable` sets `core.sparseCheckout=false` and repopulates the tree while deliberately leaving that pattern file in place, so the fast-path probe reported a false positive. The fix keeps the cheap `fs.stat` gate, but when the pattern file exists it confirms the effective `core.sparseCheckout` flag by reading git's config files directly (shared `config` plus per-worktree `config.worktree`, honoring worktree-override precedence) — no subprocess, and config parsing is extracted into the pure, exported `parseCoreSparseCheckoutFlag()`.

## Evidence

- Clean rebase onto `origin/main` (base `8a23618310`, head `1582cc6abc`).
- Files: `src/main/git/worktree.ts` (fix), `src/main/git/worktree-sparse-checkout.test.ts` (test).
- On branch: **11 tests passed** (1 file). Reverting the fix (`git checkout origin/main -- src/main/git/worktree.ts`) with the test kept: **10 failed | 1 passed**, including the behavioral guard *"does not report isSparse after disable leaves the pattern file behind"* and the `parseCoreSparseCheckoutFlag` suite (not exported on main).
- `npx oxlint src/main/git/worktree.ts` clean (rc=0).

```
npx vitest run --config config/vitest.config.ts src/main/git/worktree-sparse-checkout.test.ts
# on branch:      Test Files 1 passed, Tests 11 passed
# fix reverted:   Test Files 1 failed, Tests 10 failed | 1 passed
```

## Trade-offs

None — pure fix. The extra config read runs only for the rare worktree that still has a non-empty pattern file, and adds no subprocess.

## Regression risk

Zero-regression: only the false-positive branch changes. Worktrees with no pattern file still short-circuit on the `fs.stat` fast path (byte-identical behavior), and genuinely sparse worktrees still report `isSparse` via the real `core.sparseCheckout` flag — both proven by the passing suite. Note (out of scope, per #9921): SSH/disconnected worktrees synthesize `isSparse` from persisted `WorktreeMeta` and are not reconciled here — a separate follow-up.

_Original fix by @artile (Taras Kornichuk). Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/main/git/worktree-sparse-checkout.test.ts
+++ b/src/main/git/worktree-sparse-checkout.test.ts
@@ -1,0 +1,116 @@
+import { execFileSync } from 'node:child_process'
+import { mkdtemp, mkdir, realpath, rm, stat, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import * as path from 'node:path'
+import { afterEach, describe, expect, it } from 'vitest'
+import { listWorktrees, parseCoreSparseCheckoutFlag } from './worktree'
+
+const tempRoots: string[] = []
+
+function git(cwd: string, args: string[]): string {
+  return execFileSync('git', args, { cwd, encoding: 'utf8', stdio: ['pipe', 'pipe', 'pipe'] })
+}
+
+async function createRepoWithTwoDirs(): Promise<string> {
+  const root = await mkdtemp(path.join(tmpdir(), 'orca-sparse-checkout-'))
+  tempRoots.push(root)
+  const repoPath = path.join(root, 'repo')
+
+  execFileSync('git', ['init', '--quiet', repoPath])
+  git(repoPath, ['symbolic-ref', 'HEAD', 'refs/heads/main'])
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @artile — don't flag a disabled sparse checkout as sparse. Verified (fail-on-main/pass-on-branch), rebased on latest `main`, reviewed, and merged. Really appreciate it! 🙏

Issue #6011, opened 2026-06-21
terminal wait --for tui-idle returns satisfied:true in ~0s while an agent TUI is actively working


revofusion:
### What happens

`orca terminal wait --terminal <h> --for tui-idle --timeout-ms 30000 --json` returns
**immediately** (~0s) with `{"satisfied": true, "status": "running"}` even when the target
terminal is an agent TUI (Claude Code / Devin / Codex) that is **actively working** — spinner
animating, output streaming, `esc to interrupt` visible.

Expected: `tui-idle` should be satisfied only when the TUI is genuinely idle (agent finished,
cursor at the input prompt). As-is it fires mid-work, so it's useless as a "task done" signal
for orchestration.

### Repro

1. `orca terminal list --worktree active --json` → pick a handle whose agent is clearly busy.
2. `time orca terminal wait --terminal <h> --for tui-idle --timeout-ms 30000 --json`
3. Returns in ~0s with `satisfied:true` while the agent is still working.

### Root cause (two compounding defects)

**Defect 1 — `detectAgentStatusFromTitle` defaults agent titles to `idle`.**
`src/shared/agent-detection.ts` → `detectAgentStatusFromTitle()` final fallthrough
(`return 'idle'`): a title containing a recognized agent name but **no** explicit working
marker (no braille spinner, no `working|thinking|running`, no Claude `. ` prefix, no
`action required|permission|waiting`) is classified `idle`. Agents render their working
spinner and `esc to interrupt` in the **terminal body**, not the OSC **title**, so a busy
agent with a name-only title is reported idle.

Verified against the shipped code:

```
"⠙ Rust"                       -> working   (spinner in title — correct)
"Devin"                        -> idle      (busy Devin, name-only title)
"Devin - implementing feature" -> idle
"Codex"                        -> idle
"aider"                        -> idle
"✳ Implement REV-618"          -> idle      (Claude idle prefix)
```

**Defect 2 — the tui-idle wait trusts a single instantaneous title sample (no debounce).**
`src/main/runtime/orca-runtime.ts` → `waitForTerminal()` returns satisfied the instant
`lastAgentStatus === 'idle'` (set in `onPtyData()` from `detectAgentStatusFromTitle(oscTitle)`
— exactly the value Defect 1 mis-computes). The fallback polls
(`startPtyTuiIdleFallbackPoll` / `startTuiIdleFallbackPoll`) and the working→idle transition
[...]

PR #6012, opened 2026-06-21
fix(runtime): require sustained quiescence for tui-idle waits

revofusion:
## Problem

`orca terminal wait --terminal <h> --for tui-idle --timeout-ms 30000 --json` returns in ~0s
with `{"satisfied": true, "status": "running"}` even when the target agent TUI is **actively
working** (spinner animating, output streaming, `esc to interrupt` visible). That makes
`tui-idle` useless as a "task done" signal — it fires mid-work.

Fixes #6011.

## Root cause

Two compounding defects:

1. **`detectAgentStatusFromTitle` defaults agent titles to `idle`** (`src/shared/agent-detection.ts`).
   A title with an agent name but no explicit working marker (no braille spinner, no
   `working|thinking|running`, no Claude `. ` prefix, no `action required|permission|waiting`)
   falls through to `return 'idle'`. Agents render their spinner / `esc to interrupt` in the
   terminal **body**, not the OSC **title**, so a busy agent with a name-only title
   (`Devin`, `Codex`, `aider`, …) is classified `idle`.

2. **The tui-idle wait trusts a single instantaneous title sample** (`src/main/runtime/orca-runtime.ts`,
   `waitForTerminal`). It returns satisfied the moment `lastAgentStatus === 'idle'` — the exact
   value (1) mis-computes — with no sustained-idle requirement. The existing
   `TUI_IDLE_QUIESCENCE_MS = 3000` debounce is only consulted in the foreground-process
   fallback (which runs only when `lastAgentStatus === null`), so a title-derived `idle`
   bypasses it.

## Fix

Require **sustained** idle: a title-derived `idle` satisfies `tui-idle` only after PTY output
has been quiet for `TUI_IDLE_QUIESCENCE_MS`. An actively-working agent keeps producing output,
so `lastOutputAt` stays recent and the wait does not fire; when it finishes, output stops and
the debounce elapses.

Two small helpers (`isTuiOutputActive`, `hasSustainedTuiIdle`) replace the bare
`lastAgentStatus === 'idle'` checks at every title-derived satisfaction site:

- synchronous fast paths (pty + leaf)
- in-promise registration checks (pty + leaf)
- fallback polls (`startPtyTuiIdleFallbackPoll`, `startTuiIdleFallbackPoll`)
- working→idle transition resolvers (`resolvePtyTuiIdleWaiters`, `resolveTuiIdleWaiters`)

Preserved as-is:
[...]

PR #9929, opened 2026-07-22
fix(runtime): do not treat Grok background tasks as tui-idle

### LINKED ISSUE #9905: terminal wait tui-idle reports success while Grok background task is still active
### What happens

`orca terminal wait --for tui-idle` reports `satisfied: true` while Grok Build still has an active background task.

This is narrower than stablyai/orca#6011: foreground work is detected correctly on the same terminal, but once Grok moves work into its background-task mechanism, the input prompt is visible and Orca treats the TUI as idle even though the agent is still watching a live command.

### Environment

* macOS
* Orca `1.4.146`
* Grok Build `4.5 (high)`

### Minimal reproduction

1. Start Grok Build in an Orca-managed terminal.
2. Ask it to start `sleep 60` as a background task and keep watching until it exits.
3. Confirm with `orca terminal read --terminal <handle> --json` that the terminal shows active work, for example:

```text
Tasks 1
Task Background sleep 60 for Orca repro  33s
watching · 1 command
```

4. While the task is still active, run:

```bash
orca terminal wait --terminal <handle> --for tui-idle --timeout-ms 5000 --json
```

Observed after about `0.109s`:

```json
{
  "ok": true,
  "result": {
    "wait": {
      "condition": "tui-idle",
      "satisfied": true,
      "status": "running",
      "exitCode": null
    }
  }
}
```

5. An immediate `terminal read` still shows `Tasks 1` and `watching · 1 command`.

Control case: when Grok runs `sleep 45` as a foreground command, the same wait with a 5-second timeout correctly returns `code: timeout`. The false positive is specific to Grok's live background-task state.

### Expected behavior

`tui-idle` should not be usable as evidence that an agent turn or orchestration task is complete while the TUI reports active background work.

Possible API directions:

* Keep `tui-idle` as task-complete semantics and include Grok background-task state in the readiness check; or
* Split the concepts into an input-ready signal and a task-complete/quiescent signal so callers cannot confuse an available composer with completed work.

### Why stablyai/orca#6555 may not cover this

The proposed fix for stablyai/orca#6011 intentionally keeps body-level ready-prompt detection as an immediate success path. In this reproduction, Grok exposes an input prompt while its background task is still active, so a ready-prompt fast path can still report idle prematurely.

### Impact

Coordinators that use `terminal wait --for tui-idle` as a completion or liveness checkpoint can synthesize a result, send another task, or stop monitoring before a long-running background command or external consultation finishes.

innocarpe:
## 🧑‍🤝‍🧑 Impact & ELI5

**1) What's broken today (ELI5).**  
Orca lets you script AI coding agents: “wait until the agent is done, then do the next thing.” The command for that is:

```bash
orca terminal wait --terminal <handle> --for tui-idle --timeout-ms …
```

With **Grok Build**, that check can answer **“yes, done!” almost immediately even while Grok is still running background work** (for example a long `sleep` or an external command that Grok is still *watching*). The terminal still shows chrome like:

```text
Tasks 1
Task Background sleep 60 for Orca repro  33s
watching · 1 command
```

…and often still has a ready-looking input prompt. Automation that treats `tui-idle` as **task complete** then synthesizes a result, dispatches the next step, or stops monitoring **before the background work finishes**.

This is narrower than the older “busy TUI looked idle from a weak title” class of bugs (#6011 / #6555): **foreground** busy work is already handled correctly on the same terminal. The false positive is specific to **Grok’s live background-task state** while the composer still looks idle.

**2) What this PR does (ELI5).**  
It teaches `tui-idle` a simple extra rule:

> If the terminal wait-text still shows Grok-style **active background-task HUD chrome**, do **not** treat the TUI as idle — even if the OSC title says idle and the composer looks ready.

Once that chrome goes away (task finished / no more “watching N commands”), normal idle resolution works again.

**3) Tradeoffs / regressions — honest answer.**

| Concern | Assessment |
|--------|------------|
| Latency when truly idle | **None** for the common path: no HUD chrome → same immediate resolution as before |
| Latency while Grok background tasks run | Wait correctly continues / times out instead of false-success — **this is the fix** |
| Other agents (Codex, Claude, Antigravity, …) | Unaffected unless their wait-text literally matches the same HUD patterns (unlikely) |
| False positives | Possible if user/agent *text* happens to contain `Tasks 3` + `task` or `watching · 2 commands`. Patterns are deliberately tight (`N ≥ 1`, optional middle-dot, “command(s)”). Happy to tighten further if live Grok copy differs |
| API surface | **No new flags / conditions.** `tui-idle` stays the single condition; we keep its intended “task complete / quiescent” semantics rather than splitting input-ready vs task-complete in this PR |
| Platforms | Text-level wait-path only — macOS / Linux / Windows / SSH / remote runtime all share the same code |

No settings, UI, Git-provider, or agent-launch changes.

---

## Summary

`orca terminal wait --for tui-idle` can return `satisfied: true` in ~100ms while **Grok Build still has active background tasks**.

**Root cause**

`waitForTerminal(..., { condition: 'tui-idle' })` has several *immediate success* paths:

1. `lastAgentStatus === 'idle'` (from OSC title classification)
2. Explicit idle title markers
3. Known body-level ready prompts (`isKnownReadyPromptPreview`)

Grok Build can:

- set an **idle** OSC title (e.g. bare `grok`) and/or show a ready composer, **while**
- still rendering active background-task HUD: `Tasks 1`, `Task …`, `watching · 1 command`

So the waiter resolves as soon as idle evidence appears, without checking whether background work is still live. That matches the reporter’s control case: **foreground** `sleep` correctly stays non-idle (timeout), **background** `sleep` false-succeeds.

This is the gap left after #6555’s intentional design: body-level ready prompts remain a fast path. When Grok exposes a ready prompt *and* background tasks, that fast path (or the idle OSC path) is still wrong for “task complete.”

**Fix**

Add `hasActiveAgentBackgroundWork(waitText)` and use it as a **hard veto** on every tui-idle success path (immediate + fallback poll, leaf + PTY):

```ts
// Patterns derived from #9905’s reported Grok Build HUD
/\bwatching\s*[·.•]\s*[1-9]\d*\s+commands?\b/
/\btasks\s+[1-9]\d*\b/  &&  /\btask\b/   // count ≥ 1 plus a task row
```

When the veto fires, the waiter keeps polling / eventually times out instead of returning early success. When the HUD clears, normal idle resolution resumes.

**Why not a new wait condition?**  
Issue #9905 discusses optionally splitting input-ready vs task-complete. That would be a larger API change for CLI/skills/orchestration docs. This PR keeps `tui-idle` meaning “safe to treat the agent turn as quiescent for orchestration,” which is how docs and skills already use it, and only plugs the Grok background false positive.

Fixes #9905

---

### Automated tests

```bash
  src/main/runtime/orca-runtime.test.ts -t "tui-idle"
```

**Result: 18/18 passed** (including two new cases)

| New test | Asserts |
|----------|---------|
| `does not resolve tui-idle while Grok background tasks are still active` | Idle OSC (`grok`) + `Tasks 1` / `watching · 1 command` → **timeout** (no false success) |
| `resolves tui-idle for Grok once background tasks clear` | Idle OSC + `Tasks 0` / no watching line → **satisfied** |

Existing tui-idle cases still green (Codex ready prompt, Antigravity, coalesced titles, adopted PTY, launch-title rejection, etc.).

### Manual repro (from #9905) — expected after this PR

```bash
# While Grok is watching a background sleep / command:
orca terminal read --terminal <handle> --json
# still shows: Tasks 1, watching · N command

orca terminal wait --terminal <handle> --for tui-idle --timeout-ms 5000 --json
# BEFORE: satisfied true in ~0.1s
# AFTER:  timeout (or wait until background HUD clears, then satisfied)
```

I did not re-run against a live Grok Build binary in this environment; the regression fixture matches the issue’s reported HUD text byte-for-byte for the critical lines. If production Grok chrome differs (wording / glyphs), we can extend the detector without changing the architecture.

### Code map

| Location | Change |
|----------|--------|
| `hasActiveAgentBackgroundWork()` | New detector (Grok HUD chrome) |
| `isKnownReadyPromptPreview()` | Veto ready-prompt fast path when background work active |
| `waitForTerminal` (PTY + leaf) | Veto `lastAgentStatus === 'idle'`, explicit title idle, and ready-prompt immediate resolve |
| `startTuiIdleFallbackPoll` / `startPtyTuiIdleFallbackPoll` | Same veto so the 2s poll cannot re-introduce the false success |
| `orca-runtime.test.ts` | Regression + “cleared” control case |

---

## ELI5

Grok can look “ready for input” while it is still babysitting a background command. Orca used to treat “ready for input” as “done with the job.” Now Orca also looks for Grok’s “I’m still watching N commands” status and refuses to say “done” until that goes away.

---

## User-regression-tradeoffs

- **Intended behavior change:** `tui-idle` no longer returns success while Grok background-task HUD is visible. Callers that wanted “composer is typeable” while background work continues may now wait longer / hit timeout. That is correct for the documented orchestration use of `tui-idle` as a completion/liveness checkpoint.
- **If someone needs input-ready specifically:** that remains a possible follow-up API (`tui-input-ready` vs `tui-idle`); not introduced here to keep the PR focused and backward-compatible for the common “wait until agent finished” meaning.
- **False positive risk:** low for normal agent output; patterns require count ≥ 1 and Grok-like phrasing. Worst case of a false positive: wait times out and coordinator re-checks via `terminal read` — safer than early success.
- **No change** for agents that don’t emit this HUD while “idle.”

---

## Checklist notes (CONTRIBUTING / AGENTS)

| Area | Notes |
|------|--------|
| **Cross-platform** | Pure wait-text classification in main runtime; no platform branches. macOS/Linux/Windows share the path. |
| **SSH / remote** | Uses the same `tailBuffer` / `preview` wait-text already used for remote/daemon-hosted terminals and fallback polls. No local-only assumption. |
| **Agents / integrations** | Targeted at Grok Build background HUD; other agents only affected if wait-text matches the same strings. No agent-launch / hook changes. |
| **Git / providers** | N/A |
| **Performance** | One extra pair of regexes over wait-text that is already built for every tui-idle check. No hot-path allocations beyond existing string lowercasing. |
| **UI** | No visual change. |
| **Security** | No new IPC, no secrets, no elevated privileges. |
| **Tests** | Regression tests for active vs cleared background HUD; full `tui-idle` suite green. |

---

## Test plan

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/orca-runtime.test.ts -t "tui-idle"` → 18/18
- [x] New timeout case: idle OSC + `Tasks 1` + `watching · 1 command`
- [x] New success case: idle OSC + `Tasks 0` / no watching chrome
- [ ] Optional live smoke (reviewer / reporter): Grok background `sleep 60` + `terminal wait --for tui-idle --timeout-ms 5000` should **not** succeed early; after task ends, wait should succeed

---

## Out of scope (follow-ups welcome)

- New public wait condition for “input-ready but not quiescent”
- Structured agent status from Grok hooks (more durable than HUD scraping)
- Broadening detectors if live Grok Build chrome differs from the issue fixture

---

## AI code-review summary

Self-review against project rules:

1. **Cross-platform** — text-only runtime wait path; no `metaKey`/path separators/menu accelerators.
2. **SSH/remote** — uses existing wait-text sources used for daemon/remote PTYs; no host-local-only APIs.
3. **Agents** — Grok-specific HUD patterns; no change to spawn/hook config.
4. **Performance** — constant-time regexes on already-constructed wait-text; no extra IPC.
5. **UI** — none.
6. **Security** — none.
7. **Correctness risk** — string heuristics can drift with upstream Grok UI; mitigated by tight patterns + tests + easy extension point (`hasActiveAgentBackgroundWork`).

---

Made with care for orchestration reliability 🐋

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -12311,6 +12311,67 @@ describe('OrcaRuntimeService', () => {
     })
   })
 
+  // Why (#9905): Grok Build keeps an idle OSC title + ready composer while
+  // background tasks still run. tui-idle must not treat that as task-complete.
+  it('does not resolve tui-idle while Grok background tasks are still active', async () => {
+    vi.useFakeTimers()
+    try {
+      const runtime = new OrcaRuntimeService(store)
+      runtime.setPtyController({
+        spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
+        write: () => true,
+        kill: () => true,
+        getForegroundProcess: async () => null
+      })
+      const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+      runtime.onPtyData(
+        'pty-bg',
+        [
+          '\x1b]0;grok\x07',
+          'Tasks 1\n',
+          'Task Background sleep 60 for Orca repro  33s\n',
+          'watching · 1 command\n',
+          '> '
+        ].join(''),
+        Date.now()
+      )
+
+      const waitPromise = runtime.waitForTerminal(handle, {
+        condition: 'tui-idle',
+        timeoutMs: 5_000
+      })
+      const timeoutAssertion = expect(waitPromise).rejects.toThrow('timeout')
+      await vi.advanceTimersByTimeAsync(5_000)
+      await timeoutAssertion
+    } finally {
+      vi.useRealTimers()
+    }
+  })
+
+  it('resolves tui-idle for Grok once background tasks clear', async () => {
+    const runtime = new OrcaRuntimeService(store)
+    runtime.setPtyController({
+      spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
+      write: () => true,
+      kill: () => true,
+      getForegroundProcess: async () => null
+    })
+    const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+    runtime.onPtyData(
+      'pty-bg',
+      ['\x1b]0;grok\x07', 'Tasks 0\n', 'No background tasks\n', '> '].join(''),
+      Date.now()
+    )
+
+    await expect(
+      runtime.waitForTerminal(handle, { condition: 'tui-idle', timeoutMs: 1_000 })
+    ).resolves.toMatchObject({
+      handle,
+      condition: 'tui-idle',
+      status: 'running'
+    })
+  })
+
   it('resolves tui-idle from an Antigravity ready prompt preview', async () => {
     const runtime = new OrcaRuntimeService(store)
     runtime.setPtyController({
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -12372,6 +12372,60 @@ describe('OrcaRuntimeService', () => {
     })
   })
 
+  // Why (#9905): live OSC idle transition is the primary waiter path and must
+  // not bypass the background-work veto (immediate/poll paths alone are not enough).
+  it('does not resolve tui-idle on idle OSC title while Grok background tasks remain', async () => {
+    vi.useFakeTimers()
+    try {
+      const runtime = new OrcaRuntimeService(store)
+      runtime.setPtyController({
+        spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
+        write: () => true,
+        kill: () => true,
+        getForegroundProcess: async () => null
+      })
+      const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+      runtime.onPtyData(
+        'pty-bg',
+        [
+          '\x1b]0;\u280b Grok\x07',
+          'Tasks 1\n',
+          'Task Background sleep 60 for Orca repro  33s\n',
+          'watching · 1 command\n',
+          '> '
+        ].join(''),
+        Date.now()
+      )
+
+      const waitPromise = runtime.waitForTerminal(handle, {
+        condition: 'tui-idle',
+        timeoutMs: 5_000
+      })
+      let settled: 'resolved' | 'rejected' | null = null
+      void waitPromise.then(
+        () => {
+          settled = 'resolved'
+        },
+        () => {
+          settled = 'rejected'
+        }
+      )
+
+      // Non-idle → idle title transition while HUD still shows Tasks 1
+      runtime.onPtyData('pty-bg', '\x1b]0;Grok\x07', Date.now())
+      await Promise.resolve()
+      await vi.advanceTimersByTimeAsync(1_000)
+      expect(settled).toBeNull()
+
+      const timeoutAssertion = expect(waitPromise).rejects.toThrow('timeout')
+      await vi.advanceTimersByTimeAsync(5_000)
+      await timeoutAssertion
+      expect(settled).toBe('rejected')
+    } finally {
+      vi.useRealTimers()
+    }
+  })
+
   it('resolves tui-idle from an Antigravity ready prompt preview', async () => {
     const runtime = new OrcaRuntimeService(store)
     runtime.setPtyController({
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -12357,15 +12357,19 @@ describe('OrcaRuntimeService', () => {
       getForegroundProcess: async () => null
     })
     const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+    // Why: seed active background work, then clear — proves Tasks 1 → Tasks 0 releases waiters.
+    runtime.onPtyData('pty-bg', 'Tasks 1\nwatching · 1 command\n> ', Date.now())
+    const waitPromise = runtime.waitForTerminal(handle, {
+      condition: 'tui-idle',
+      timeoutMs: 1_000
+    })
     runtime.onPtyData(
       'pty-bg',
       ['\x1b]0;grok\x07', 'Tasks 0\n', 'No background tasks\n', '> '].join(''),
       Date.now()
     )
 
-    await expect(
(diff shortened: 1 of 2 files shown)


nwparker:
Flagging this so it doesn't get swept up in the #6011 cleanup: **this PR targets #9905, not #6011, and the two are different bugs in opposite directions.**

There are currently five open PRs against #6011 (#6012, #6555, #9929, #14642, #20155) and it would be easy to triage this one as a duplicate. It isn't:

- **#6011** is a *false positive from a name-only title*: `detectAgentStatusFromTitle` falls through a bare agent name (`Codex`, `grok`) to `idle`, so `terminal wait --for tui-idle` returns satisfied in ~0s while the agent is mid-turn.
- **This PR** is a *false positive from the body*: Grok's title legitimately reads idle while background tasks are still running, and `hasActiveAgentBackgroundWork` vetoes idle while that HUD is on screen.

This PR does not fix #6011 — a name-only title still settles instantly here — and none of the #6011 PRs fix #9905. They compose rather than compete.

That composition is worth spelling out, because it's specifically relevant to Grok. While reviewing the #6011 PRs I measured a real idle `grok` pane driven through a real PTY: it emits a bare `grok` OSC title, and it repaints a braille banner continuously — max gap between output bytes **891ms over a 22s observation**, still emitting 76ms before the window closed. Tail-preview is no better: 79 preview changes in 20 idle seconds, longest stable window **2.1s** against a 3s quiescence threshold.

The consequence is that every #6011 approach which demands *output quiescence* before accepting Grok's name-only title leaves Grok unable to ever report idle — the wait runs to timeout. So the Grok fix has to go the other way: keep letting Grok's name-only title settle, because that title is the only rest signal Grok ever emits. Which is exactly why this PR matters — it supplies the missing veto for the case where that title is genuinely wrong, and it does so by reading Grok's **body**, which no other PR in this set does.

Net: Grok settles on its name-only title, *except* while the background-task HUD says work is live. Neither half is sufficient alone.

No changes requested from me — just recording the analysis so this lands on its own merits rather than being closed as a #6011 duplicate.

PR #9945, opened 2026-07-22
fix(terminal): terminate descendant processes on close

joowon-jang:
## Summary

- terminate descendant processes for ordinary and agent terminal sessions before tearing down the PTY root
- reuse the guarded `taskkill /T /F` flow on Windows and the process snapshot flow on macOS/Linux
- wait for physical PTY exit when closing one terminal or an entire split tab
- settle every PTY stop attempt so one rejection cannot skip remaining stops or team cleanup, reporting `ptyKilled: false` when any stop fails
- preserve PID ownership and OS identity checks, duplicate teardown coalescing, and root-kill fallback behavior

This prevents terminal children from surviving after their Orca pane or tab has been closed and makes `ptyKilled` reflect the physical stop result.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — changed-file oxlint passed
- [ ] `pnpm typecheck` — `pnpm run typecheck:node` passed
- [ ] `pnpm test` — 8 relevant test files passed, 1054/1054 tests
- [ ] `pnpm build`
- [x] Added or updated high-quality tests that catch regressions in split-tab rejection isolation, successful-stop fallback suppression, Windows PID identity checks, descendant sweep ordering, and physical-exit waiting

Additional native verification on Windows used a root process with a separately grouped leaf process. `taskkill.exe /pid <root> /T /F` removed both processes, with zero matching fixture processes remaining.

## AI Review Report

The review checked the teardown call graph from runtime close operations through the PTY controller and daemon session owner. It flagged and resolved an inconsistent legacy `Session.kill()` branch so ordinary and agent sessions now both use descendant-aware teardown. Follow-up review also found and resolved split-tab failure isolation so every PTY stop settles and team cleanup still runs when one stop rejects. Sweep failures retain the native owner until the root physically exits before propagating.

Cross-platform review covered:

- macOS/Linux snapshot-before-root ordering, PID reuse protection, grace-period escalation, and root fallback
- Windows process-tree ordering, timeout, hidden-window behavior, session ownership, OS identity verification, and failure fallback
- local, SSH, and WSL provider boundaries; the change stays within the provider that owns the PTY
- split tabs and pty-only runtime records, including duplicate PTY IDs and rejected stop attempts
- no UI, shortcut, label, path, Git provider, or agent-specific behavior changes

## Security Audit

- validates that the Windows PID is a positive integer
- invokes `taskkill.exe` with an argument array and no command shell, avoiding command injection
- requires both live session ownership and a positive OS identity match before targeting a Windows PID
- bounds `taskkill.exe` to five seconds and falls back to existing root teardown on failure
- adds no dependencies, secrets, authentication flows, filesystem writes, or new IPC surface

## Notes

- Windows tree termination is forceful for both graceful and immediate requests after the OS confirms the PTY root belongs to Orca.
- Immediate teardown still waits for physical exit; graceful teardown retains the existing root signal and force-fallback lifecycle.

## ELI5

Closing a terminal left child processes (including agent trees) running. Orca now terminates descendants on close and waits for the PTY to exit so worktrees and Resource Manager do not keep ghosts.


nwparker:
Closing — partly superseded, and partly a deliberate difference worth stating.

**Superseded:** the descendant reach itself. W2 (#15755) puts every ConPTY pane in a Win32 job object, and #15886 wires the job into `terminal-session-teardown.ts` — including `forceKillPlainShellSession`, which is the path a destructive close takes. So an ordinary shell that left a detached child behind is now handled exactly, without a process-table scrape, and it reaches a grandchild that reparented away from the shell (which the parent-pid walk this PR extends cannot see).

**Deliberate difference:** this PR also makes the *graceful*, non-immediate close of an ordinary shell run the sweep. We are keeping graceful close as root-signal-only on that path. The same question came up when choosing job flags in W2: we measured `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` reaping a detached server that had survived before, and dropped the flag, because a clean exit silently killing backgrounded work is a behaviour change nobody asked for. Making graceful close sweep is the same change wearing a different hat. Destructive teardown — worktree delete, force close — is explicit, and that is where the job runs.

If the reports behind this PR were about a graceful tab close leaving something alive, that is a real bug and worth reopening with the repro, because it would mean the destructive path is not being taken where we think it is. That exact assumption turned out to be wrong once already (#11047).

Thanks.

nwparker:
Descendant reach superseded by #15755/#15886; graceful close stays root-signal-only by design (see comment).

nwparker:

PR #9955, opened 2026-07-22
fix: sidebar worktree card reads local SSH state for remote-runtime worktrees

VictorCano:
Partially addresses #9276 — this fixes the misleading "SSH disconnected" chip and the spurious blocking reconnect dialog on paired clients. The other half of the issue — terminals for hub SSH-hosted worktrees never opening — turned out to be a client-side bug of the same class and is fixed separately in #9987 (single-topic PRs).

## User-visible change

On a paired desktop client connected to a hub machine running the Orca runtime, every one of the hub's SSH-hosted worktrees showed a red "SSH disconnected" chip in the sidebar, and clicking one popped the blocking `SshDisconnectedDialog` — even while the hub's SSH connections were perfectly healthy. After this fix, the sidebar card reports the hub's actual SSH state (chip only on a real disconnect, correct target label in the dialog), and shows nothing while the owner's state is unknown instead of defaulting to a false "disconnected".

## Root cause

`WorktreeCard.tsx` still read the flat **local-machine** SSH maps directly (`s.sshConnectionStates.get(repo.connectionId) ?? 'disconnected'`, and `s.sshTargetLabels` for the dialog label). On a paired client, the hub's SSH connections never exist in those local maps — they are mirrored into the per-environment bucket `sshStateByEnvironment`, whose slice explicitly documents "Do NOT read this map directly from components — use the `selectRuntimeAwareSsh*` selectors". So the lookup was always `undefined` → `'disconnected'` for every hub SSH worktree. This was a missed migration to the runtime-aware SSH selectors that `TerminalPane.tsx` already uses; the card now mirrors TerminalPane's exact pattern (`getExplicitRuntimeEnvironmentIdForWorktree` + `selectRuntimeAwareSshStatus`/`selectRuntimeAwareSshTargetLabel`, including the paired-web-client force-local guard).

## Changes

- **New `worktree-card-ssh-status.ts`** (sidebar sibling module, following `worktree-card-pr-display.ts` naming): lifts the card's inline SSH status/label derivation into pure, unit-testable selectors that route between local maps (`environmentId === null`) and the owning environment's bucket. Local-worktree values are byte-for-byte identical to before (including the `'' → repo.displayName` label fallback).
- **`WorktreeCard.tsx`**: replaces the direct map reads with the new selectors; derives the owning environment id exactly like TerminalPane. The blocking-dialog gating (`isSshDisconnected` in the click handler and the dialog `open` prop) is fixed transitively since it derives from the corrected status.
- **`runtime-environment-ssh.ts`**: export-only change (`RuntimeAwareSshReadState` type) so the new module can share the selectors' state contract. No behavior change.

## Tests

Written red→green: every regression case below was run against today's `main` behavior first and observed failing, then observed passing with the fix.

**Component-level (`WorktreeCard.ssh-reconnect-prompt.test.tsx`, 3 new cases)** — drives the exact #9276 topology through the real card (runtime-hosted worktree + `repo.connectionId` + hydrated environment bucket + empty local maps):

- hub reports `connected` → no "SSH disconnected" chip, dialog receives `connected` + the hub's target label (fails on main)
- hub reports `disconnected` → chip still shows, with the hub's label (label fails on main)
- hub bucket not yet hydrated → no chip instead of a false disconnect (fails on main)

**Selector-level (`worktree-card-ssh-status.test.ts`, 13 cases)**:

- hub-owned target `connected` in `sshStateByEnvironment` but absent from local `sshConnectionStates` → reports `connected` + the hub's target label (fails on main: reported `disconnected` + empty label)
- unreachable environment / unhydrated bucket → `null` (show nothing), not a false `disconnected`
- a real hub-side disconnect still surfaces
- local SSH worktrees keep local-map semantics unchanged (status fallback `'disconnected'`, label fallback `''`)
- runtime-owned (`runtime-ssh-*`) targets and missing connection ids stay suppressed
- end-to-end wiring: a worktree with a `runtime:<env>` host id resolves to its environment via `getExplicitRuntimeEnvironmentIdForWorktree` and reads that bucket; a plain local worktree resolves to no environment

## AI code-review summary

Co-developed with Claude (Claude Code, Fable 5) under human direction and review: I scoped the fix, reviewed the diff, and verified the result live on my real hub + paired-client setup (screenshots above). On the AI side it was reviewed by the implementing agent plus an adversarial multi-model review pass (3 independent reviewer lenses, each finding verified by a separate refutation agent). The one confirmed finding — the end-to-end #9276 topology was not driven through the real component, so a wiring revert would have kept the suite green — was addressed by adding the component-level tests above (verified red against the reverted wiring):

- **Cross-platform (macOS/Linux/Windows)**: renderer-only store/selector logic; no paths, shells, key handling, or platform APIs touched.
- **SSH/remote/local compatibility**: the point of the fix. Local worktrees keep byte-for-byte identical behavior (unit-tested); hub-owned worktrees read the owning environment bucket; paired web clients keep forcing local maps (same guard as TerminalPane); runtime-owned ephemeral-VM targets stay suppressed; unknown owner state renders nothing rather than a false disconnect.
- **Agent/integration neutrality**: no agent-, integration-, or git-provider-specific behavior involved.
- **Performance (sidebar render hot path)**: no new IPC and no new subscriptions; the only added work per store update is the environment-id resolution (`getExplicitRuntimeEnvironmentIdForWorktree`, in-memory map/array lookups) plus two O(1) bucket lookups — the same per-selector cost TerminalPane already pays, and the added selectors return scalars so re-render behavior is unchanged. Bucket hydration continues to happen centrally in `useIpcEvents` on runtime (re)connect; cards trigger none.
- **UI quality**: chip/dialog visuals untouched; only when they appear changes. Dialog label now prefers the owning host's label and still falls back to `repo.displayName` (never a raw `ssh-<ts>-<rand>` id).
- **Security**: no new data flows; reads already-mirrored renderer store state only.

An additional OpenAI Codex native review of the final diff surfaced two P2 observations, both acknowledged and intentionally left out of this single-topic PR as follow-ups:

1. **Reconnect routing**: `SshDisconnectedDialog`'s Reconnect button calls the local `window.api.ssh.connect({ targetId })` path, which cannot reconnect a hub-owned target (it would need the runtime RPC path the terminal overlay uses). This is pre-existing behavior — before this fix that dead-end dialog was reachable constantly via the false chip; now it is reachable only on a genuine hub-side disconnect.
2. **Offline-hub visibility**: when the owning environment is unreachable, the card intentionally shows nothing (`null` = unknown, mirroring TerminalPane's semantics) rather than a possibly-false "SSH disconnected" chip; for an SSH repo whose worktree is hub-hosted this means no chip while the hub itself is offline. Whether that shape should get a "Server disconnected" fallback keyed off the worktree's `hostId` is left as a follow-up.

## Screenshots

Paired Windows desktop client, hub (macOS, Orca 1.4.147) with healthy SSH connections. `TL/Fusion` is a hub SSH-hosted repo; `TL/fusion-mcp` is hub-local.

| Before | After |
| --- | --- |
| Before: every TL/Fusion worktree shows a red SSH-disconnected icon despite the hub's SSH connection being healthy | After: TL/Fusion worktrees render normally with no false disconnected state |

## Notes

- Pre-existing failures on `main` (untouched — each reproduced on a clean `main` checkout in the same environment):
  - `pnpm lint` fails in `lint:switch-exhaustiveness` on `src/renderer/src/components/skills/skill-freshness-group.tsx:101` (switch not exhaustive). The files in this PR are lint-clean.
  - `pnpm test`: 3 failures in `src/main/startup/configure-process.test.ts` (GPU feature flags) and `src/main/agent-hooks/managed-hook-timeout.test.ts` (transport-budget timing), all outside the renderer; the other 33,783 tests pass, including the 16 new ones.
- `pnpm typecheck` and `pnpm build` pass.
- No version bumps.

## ELI5

Paired clients showed hub SSH worktrees as "SSH disconnected" even when the hub was fine. The sidebar now uses the correct local SSH state for remote-runtime worktrees so the chip and reconnect dialog stop lying.

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeCard.ssh-reconnect-prompt.test.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeCard.ssh-reconnect-prompt.test.tsx
@@ -2,6 +2,8 @@ import { renderToStaticMarkup } from 'react-dom/server'
 import type { ReactNode } from 'react'
 import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
 import type { Repo, Worktree, WorktreeCardProperty } from '../../../../shared/types'
+import { toRuntimeExecutionHostId } from '../../../../shared/execution-host'
+import type { RuntimeEnvironmentSshBucket } from '@/store/slices/runtime-environment-ssh'
 import type WorktreeCardComponent from './WorktreeCard'
 
 const fetchHostedReviewForBranch = vi.fn()
@@ -14,6 +16,8 @@ let WorktreeCard: typeof WorktreeCardComponent
 let sshConnectionStates = new Map<string, { status: string }>()
 let sshTargetLabels = new Map<string, string>()
 let runtimeStatusByEnvironmentId = new Map<string, { status?: unknown }>()
+let sshStateByEnvironment = new Map<string, RuntimeEnvironmentSshBucket>()
+let worktreesByRepo: Record<string, Worktree[]> = {}
 let worktreeCardProperties: WorktreeCardProperty[] = ['status']
 
 vi.mock('@/store', () => ({
@@ -32,10 +36,14 @@ vi.mock('@/store', () => ({
       remoteBranchConflictByWorktreeId: {},
       runtimeStatusByEnvironmentId,
       settings: null,
+      removedSshTargetLabels: new Map(),
       sshConnectionStates,
+      sshStateByEnvironment,
       sshTargetLabels,
+      sshTargetsHydrated: false,
       updateWorktreeMeta,
-      worktreeCardProperties
+      worktreeCardProperties,
+      worktreesByRepo
     })
 }))
 
@@ -130,6 +138,8 @@ describe('WorktreeCard SSH reconnect prompt', () => {
     sshConnectionStates = new Map()
     sshTargetLabels = new Map()
     runtimeStatusByEnvironmentId = new Map()
+    sshStateByEnvironment = new Map()
+    worktreesByRepo = {}
     worktreeCardProperties = ['status']
   })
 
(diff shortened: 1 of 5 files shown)


AmethystLiang:
Associating this PR with related issue(s): #9276

AmethystLiang:
Associating this PR with related issue(s): #9276

PR #9956, opened 2026-07-22
fix(sidebar): preserve workspace shortcut order when sidebar panel is hidden

kunsanglee:
## Summary

Preserves workspace shortcut order when the left sidebar panel is hidden. Previously, unmounting `WorktreeList` cleared the rendered-order cache, so `Cmd/Ctrl + 1–9` fell back to a flat live-store sort that could differ from the sidebar's project grouping, pinned positions, and section hierarchy.

## Fix

1. Retain the last rendered worktree ID order when `WorktreeList` unmounts.
2. Revalidate cached IDs against the current visible worktrees, preserve the cached relative order for IDs that remain visible, and append newly visible worktrees in live sort order.
3. Add regression coverage for cached ordering, archived and deleted IDs, newly visible worktrees, filter changes, and test-state cleanup.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — left unchecked only because of a failure outside this PR's scope. `oxlint` passes for every file this PR changes; the full command reaches the type-aware exhaustiveness check and then fails on the pre-existing, unmodified `src/renderer/src/components/skills/skill-freshness-group.tsx:101`.
- [x] `pnpm typecheck`
- [x] `pnpm test` — the full suite passed on rerun. The first run had two environment-related timeouts while native/Electron dependencies were warming; both tests passed in isolation before the successful full rerun.
- [x] `pnpm build`
- [x] Targeted `visible-worktrees.test.ts` suite passed all 41 tests.
- [x] `git diff --check`

All commands above were run with Node 24 and pnpm 10.24.0.

## AI Review Report

Reviewed the branch diff against `main`, the rendered sidebar-order publication path, the workspace-index shortcut handler, and the live visibility computation.

The review found that the initial cache regression test did not seed the live Zustand store. Because the production code revalidates cached IDs against live visibility, that test would have returned an empty list. The test now seeds real worktrees and verifies:

- cached relative order is preserved;
- archived and deleted cached IDs are discarded;
- newly visible worktrees are appended in live sort order;
- active repository filters revalidate the cache.

Compatibility and risk review:

- **macOS, Linux, and Windows:** No platform-specific shortcut, label, path, shell, or Electron behavior was added. The change remains downstream of the existing centralized shortcut policy, which already resolves platform modifiers before sending the same workspace index through the existing IPC path.
- **Local, remote, and SSH worktrees:** The change operates only on host-neutral worktree IDs and existing renderer store state. It adds no local filesystem, process, credential, shell, or network assumptions.
- **CLI agents and integrations:** No agent protocol, agent-specific branch, integration API, or live-agent detection behavior changed. Existing agent activity only contributes to the shared live visibility set.
- **Git providers:** No Git command or provider-specific behavior changed; GitHub, GitLab, and other providers remain unaffected.
- **Performance:** The live visibility and sort snapshot is recomputed only when a workspace-index shortcut is pressed so stale or filtered IDs can be rejected. No render loop, polling path, or background work was added. No benchmark was run.
- **UI quality:** No layout, color, typography, component, light/dark theme, or shortcut-label output changed. The behavioral result is that hidden-sidebar navigation remains consistent with the last rendered sidebar order.

## Security Audit

No new user input, IPC channel, command execution, filesystem path handling, authentication, secrets, or dependencies were introduced. The changed code reads existing in-memory store state and rejects cached IDs that are no longer present in the current visible-ID set. No security follow-up was identified.

## Notes

- The remaining full-lint failure is outside this PR's changed files.
- No platform-specific behavior, migration, remote/SSH follow-up, agent/integration follow-up, or Git-provider follow-up is required.
- X (Twitter): [@kunsanglee](https://x.com/kunsanglee)

## ELI5

Hiding the left sidebar used to scramble Cmd/Ctrl+1–9 workspace order. Orca keeps the last rendered order so shortcuts still match the project grouping you saw when the panel was open.

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeList.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeList.tsx
@@ -5964,8 +5964,6 @@ const WorktreeList = React.memo(function WorktreeList({
   // Why layout effect: the Cmd/Ctrl+1–9 handler can fire right after commit; publishing after paint would leave the shortcut cache stale.
   useLayoutEffect(() => {
     setVisibleWorktreeIds(renderedWorktreeIds)
-    // Why: unmounting the list clears the rendered-order cache so shortcuts fall back to the live store snapshot.
-    return () => setVisibleWorktreeIds([])
   }, [renderedWorktreeIds])
 
   const handleCreateForRepo = useCallback(
--- a/src/renderer/src/components/sidebar/visible-worktrees.test.ts
+++ b/src/renderer/src/components/sidebar/visible-worktrees.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest'
+import { afterEach, describe, expect, it } from 'vitest'
 import {
   computeClearFilterActions,
   computeVisibleWorktreeIds,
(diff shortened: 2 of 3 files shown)


kunsanglee:
This PR is from a fork, so the `PR Checks` workflow has ended in `action_required` on all three commits and CI has not run yet. Could a maintainer approve the workflow run? Happy to address review feedback once the checks are green.

kunsanglee:
Closing this one — the commits were authored under the wrong account. I'll reopen the same change from a clean branch with correct authorship.

PR #9961, opened 2026-07-22
fix(pty): strip inherited Claude child-session stamps at spawn (silent transcript-persistence loss)

akapug:
### Problem

If an Orca GUI (or its daemon) is ever launched from inside a Claude Code session — a terminal where `claude` is running, a shell spawned by it, or an auto-update relaunch chained from such a launch — it inherits three environment variables:

```
CLAUDE_CODE_CHILD_SESSION
CLAUDE_CODE_SESSION_ID
CLAUDE_CODE_BRIDGE_SESSION_ID
```

Both spawn paths start from a wholesale spread of the host's env (the daemon fork env in `daemon-init.ts`, `buildSpawnEnv` in `src/relay/pty-handler.ts`), so every terminal Orca spawns inherits the stamps. Claude Code treats a stamped process as a *nested child session* and **silently disables transcript persistence** — the user's real sessions stop writing their on-disk history, with no visible error and no way to notice until a resume fails.

Two otherwise-good design choices amplify the blast radius: older-protocol daemons are deliberately preserved across upgrades and never respawn, and the auto-updater relaunch inherits the previous app's env. So one contaminated launch propagates through every subsequent auto-update until a fully clean-environment relaunch. Measured on a production machine: a GUI launched once from a Claude session on 07-20 stamped three daemon generations (protocol v23–v25) across two auto-updates; every pane spawned in that window ran with persistence silently off.

### Fix

Orca never sets these variables itself, so on the host they are always inherited poison. This mirrors the existing `AGENT_HOOK_RUNTIME_ENV_KEYS` pattern in `src/main/ipc/pty.ts`:

* a `CLAUDE_CHILD_SESSION_STAMP_ENV_KEYS` deny constant,
* `getInheritedClaudeSessionStampEnvKeysToDelete()` — the same inherited-only filter as the agent-hook helper, so a stamp explicitly passed in `args.env` (a caller deliberately spawning a nested Claude child) is preserved,
* merged into `envToDelete` at both spawn call sites. Unlike the agent-hook keys this is not gated on `isDaemonHostSpawn`, because the relay host's own `process.env` can be contaminated the same way.

### Tests

Two tests in `pty.test.ts` using the existing `daemonSpawnAndGetOptions` helper: a contaminated host env puts all three keys in `envToDelete`; an explicit `args.env` stamp is preserved. The strip test fails against unpatched source (verified); the file suite is 304/304 with the patch. `oxlint`/`oxfmt` clean.

---

*This patch was authored by fable, an AI agent working on @akapug's machine, where the bug was found and measured live; @akapug reviewed and submitted.*

## ELI5

If Orca was started from inside a Claude session, it inherited Claude child-session env vars and silently broke transcript persistence for new sessions. Spawn paths now strip those stamps so each session keeps its own history.

### DIFF
--- a/src/main/ipc/pty.test.ts
+++ b/src/main/ipc/pty.test.ts
@@ -1683,6 +1683,28 @@ describe('registerPtyHandlers', () => {
       expect(env.LANG).toBe('fr_FR.UTF-8')
     })
 
+    it('strips inherited Claude child-session stamps from a local spawn env', async () => {
+      // Why: the local provider spreads main's process.env, so a GUI launched from
+      // inside a Claude session would stamp every pane as a nested child and Claude
+      // would silently disable transcript persistence. Not gated on isDaemonHostSpawn.
+      const env = await spawnAndGetEnv(undefined, {
+        CLAUDE_CODE_CHILD_SESSION: '1',
+        CLAUDE_CODE_SESSION_ID: '85935aed-98a7-4094-89a8-85c75e1a5a95',
+        CLAUDE_CODE_BRIDGE_SESSION_ID: 'session_01UCkWN5nDXNyD1V7cfamCxa'
+      })
+      expect(env.CLAUDE_CODE_CHILD_SESSION).toBeUndefined()
+      expect(env.CLAUDE_CODE_SESSION_ID).toBeUndefined()
+      expect(env.CLAUDE_CODE_BRIDGE_SESSION_ID).toBeUndefined()
+    })
+
+    it('keeps an explicitly requested Claude child-session stamp on a local spawn', async () => {
+      const env = await spawnAndGetEnv(
+        { CLAUDE_CODE_CHILD_SESSION: '1' },
+        { CLAUDE_CODE_CHILD_SESSION: '1' }
+      )
+      expect(env.CLAUDE_CODE_CHILD_SESSION).toBe('1')
+    })
+
     it('always sets TERM and COLORTERM regardless of env', async () => {
       const env = await spawnAndGetEnv()
       expect(env.TERM).toBe('xterm-256color')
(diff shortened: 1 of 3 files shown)


brennanb2025 (approved):
Good find, and a nasty failure mode — silent, and it survives auto-updates. Rebased onto current `main` (#9608 landed separately, so this branch is now just the stamp fix) and pushed three review commits.

**Two things needed fixing first.** Neither was caught locally because `oxlint`/`oxfmt` were run but not `typecheck`, and CI never ran on this PR while it was conflicting:

- **It didn't compile.** `TS2345` at the runtime-controller call site — nesting `mergePtyEnvDeletions` passes a `string[] | undefined` intermediate into its `readonly string[]` second parameter. Made the helper variadic instead, so spawn sites list deletion sources flat in precedence order; that also collapsed a six-deep nest at the renderer call site and benchmarks slightly faster (one Set instead of five).
- **It broke a green test.** `provider-dispatch.test.ts > routes to SSH provider` passes on `main` and failed here. Root cause is worth knowing: the stamps are essentially never in `args.env`, so the filter always returns all three keys and **`envToDelete` is now never `undefined`** at either call site. That broke an exact spawn-options assertion. Fixed by asserting the stamp list on the SSH path (as a set — `envToDelete` is consumed by membership only, so a reordering of the merge sources shouldn't be able to fail it).

**Test coverage went 2 → 7.** The original pair only covered the renderer `pty:spawn` handler under the daemon adapter. Added the runtime-controller path (`orca` CLI / automations — a separate `envToDelete` assembly), a local-provider case asserting the *actual* node-pty env, and the SSH path. That matrix matters because the strip is deliberately ungated on `isDaemonHostSpawn`: a daemon-only test still passes if someone later adds that gate. Five negative controls each fail exactly the intended tests.

**Verified live** on a naturally-contaminated host. Unpatched `main`: all three stamps present in a terminal pane. Patched: zero — while `CLAUDE_CODE_ENTRYPOINT`/`_PROXY`/`_EXECPATH` are still inherited, confirming the deny-list doesn't over-delete.

Two notes on the description, neither blocking:

- `CLAUDE_CODE_BRIDGE_SESSION_ID` isn't set in an ordinary Claude Code shell — only the other two are. Stripping it is still correct and free, just less universal than described.
- "both spawn paths" is narrower than it reads: `rate-limits/claude-pty.ts`, `claude-accounts/service.ts` and `text-generation/commit-message-text-generation.ts` spawn `claude` with raw `process.env`, so the stamps survive there. Deliberately out of scope — those are utility spawns with no user transcript at stake — but worth a follow-up.

One bound worth recording: `envToDelete` only reached the daemon protocol in #2350 (2026-05-25), so daemons preserved from older builds ignore it. The generations in your report (2026-07-20) are well after that, so the propagation fix holds in practice.

PR #9981, opened 2026-07-22
feat(mobile): show attached images as removable thumbnails in native chat

brennanb2025:
## Problem

Attaching a picture in mobile native chat gave no visible feedback. The image uploaded to the host and its file path was pasted straight into the **hidden PTY** (`mobile-image-attachment.ts`), so beyond a brief spinner on the attach button, nothing on screen ever indicated an image was attached — it looked like the button did nothing.

## Change

Native chat now defers the TUI paste to send time and shows picked images as a thumbnail tray above the input, matching the desktop composer-attachment model (chips held in composer state, images "ride along" on submit):

- **`use-mobile-native-chat-pending-images.ts`** — pending-image state. Picking still uploads to the host eagerly (same chunked upload RPC); the entry renders as an 88px thumbnail with a spinner overlay while uploading and a circular ✕ once ready. Thumbnails use the picker's local file URI so no second base64 copy is held in JS memory. Pending images are scoped to the chat target and RPC client, and target switches/removal/unmount cancel remaining upload chunks.
- **`use-mobile-native-chat-image-send.ts`** — send wrapper. Pastes each ready image's host path bracketed (`enter:false`, the same payload as a desktop image paste) right before the message text. Accepted path writes remain as non-removable `pasted` entries until the caption/Enter is accepted, so rejected and delivery-unknown retries never paste the same path twice. Image-only sends submit with a bare Enter and can retry only that Enter after rejection. Newly pasted paths use the same 300 ms image-settle and 500 ms submit cadence as desktop so caption or Enter bytes cannot overtake TUI image-chip processing.
- **`MobileNativeChatComposer.tsx`** — attachment tray UI; send is allowed with images and no text, blocked while an upload is in flight, and uses the native-chat upload state rather than the unrelated terminal-mode spinner.
- `PickedMobileImage` now carries the picker's file `uri`; the attach-failure toast copy is extracted to `mobileImageAttachToastMessage` and shared by both flows.

Deferring the paste is what makes ✕ possible: removing a ready thumbnail needs no TUI un-paste. Once a path has reached the hidden TUI, its retained non-removable thumbnail keeps composer state aligned with that input until submission settles.

The terminal live-input surface intentionally keeps the immediate paste — there the agent TUI's own input line is the visible feedback.

Follow-up review also closes asynchronous target-lifetime races: picker/upload completion is ignored after a terminal/client switch, removal, or unmount; multi-image delivery stops if the active terminal/client/lease changes; and a caption can no longer be sent to a different terminal than the one that received the image paths. The native-chat controller owns this coordination, keeping the already-dense session route below its max-lines ratchet without a lint bypass.

## Out of scope (follow-ups)

- The mobile chat transcript has no image-block rendering, so a sent image still surfaces via the agent's raw text markers; porting the desktop transcript normalization is a separate change.
- The attachment lease-gate's silent-drop paths (`use-mobile-attachment-input-lease-gate.ts`) still read as picker-cancel.

## Testing

- 26 new unit tests, including upload lifecycle/cancellation, picker cancel/failure, ready/pasted/remove/consume transitions, handle/client-switch clearing, picker-open and mid-send terminal switches, removal/unmount during upload, accepted/rejected/ACK-lost paste sequencing, retry call counts, image-only submit/Enter retry, composer readiness, and lease guards.
- Full mobile suite on the latest main merge: 316 files passed; 2,296 tests passed and 2 skipped.
- Focused post-merge attachment gate: 4 files and 44 tests passed.
- Electron RPC boundary suites: 41 tests passed across clipboard upload and terminal send.
- Mobile `tsc --noEmit`, full oxlint, oxfmt, the repository max-lines ratchet, and reliability-manifest validation passed.

## Reliability contract

- **Invariant (`mobile-relay.terminal-input-delivery`):** an attachment remains bound to its originating chat target; retry never duplicates a path already accepted or possibly delivered by `terminal.send`; a rejected final submit remains retryable; caption and image-only submit writes preserve the desktop-proven TUI settle cadence.
- **Failure source:** the original hidden-PTY feedback bug plus plausible relay ACK loss, tab/client switches, and removal/unmount during a large upload.
- **Oracle:** deterministic tests count exact `terminal.send` calls across accepted, rejected, target-switch, and delivery-unknown outcomes; cancellation asserts `start → first chunk → abort` with no later chunk or commit.
- **Gate:** accepted gap — no dedicated entry currently exists in `config/reliability-gates.jsonc`; the focused mobile tests are the deterministic gate for this PR.
- **Provider/platform coverage:** mobile/relay behavior is covered at the RPC/state-machine layer; path handling stays opaque for local and SSH hosts. Existing manual iOS validation covers the happy path. Live Android, live SSH failure injection, WSL, Windows, and relay-partition runs remain unproved; no provider-specific path/process behavior changed.
- **Performance budget:** no polling, inventory scans, subprocesses, or persistent listeners were added; the send path adds only one bounded 300 ms or 500 ms TUI-settle timer when it writes a new image path. Cancellation stops future 512 KiB upload chunks and releases server upload state; retries issue zero duplicate image-path RPCs.
- **Diagnostics:** rejected sends retain the inline failure state; ambiguous image delivery surfaces a specific “check terminal before retrying” toast instead of claiming failure.
- **Residual gap:** an ACK-lost non-submitting image-path write cannot be proven delivered from the native transcript. The implementation preserves it as `pasted` and avoids duplication; the user must inspect the terminal before retrying, as the toast states.

## Manual validation

Validated on an iPhone 17 Pro simulator (iOS 26.5) paired to an isolated desktop runtime: selecting a photo shows a ready thumbnail, ✕ removes it without touching the TUI input, and an image-only send delivers the image then clears the tray.

| Ready to send | Removed |
| --- | --- |
| Native chat composer showing a ready image thumbnail | Native chat composer after removing the pending image |


brennanb2025:
Closing as superseded by merged #10135. The retry-safety follow-up from the Sol-high review is in #10228.

PR #9983, opened 2026-07-22
fix(settings): detect agents on active runtime host

Rast53:
## Summary
- `AgentsPane` and `OrchestrationSkillAgentCoverage` always called `useDetectedAgents()` with no target, so Settings → Agents probed the **local laptop PATH** even when `activeRuntimeEnvironmentId` pointed at a remote `orca serve` / Devbox host.
- Remote launches already worked; only the detected-agent list was wrong. Mobile/web clients against the same server showed the full remote catalog.
- `QuickLaunchAgentMenuItems` only resolved SSH `connectionId` and missed runtime environments, so the same local-only list could appear in the launch menu for Devbox worktrees.

## Fix
- Pass `{ kind: 'runtime', environmentId }` from `settings.activeRuntimeEnvironmentId` into `useDetectedAgents` in Agents settings + orchestration skill coverage.
- Align Quick Launch with `TabBar`: SSH connection → runtime environment → local.

## Test plan
- [ ] Open Orca desktop with active runtime = remote Devbox / `orca serve`
- [ ] Settings → Agents shows remote-detected CLIs (e.g. Claude, Codex, Cursor, Grok), not only laptop PATH
- [ ] Refresh agents re-probes the runtime host
- [ ] New Agent / Quick Launch menu lists the same remote agents
- [ ] With no active runtime, local PATH detection still works
- [ ] Existing unit mocks for `useDetectedAgents` continue to pass (call-site still compatible)

Closes: remote-only agent list mismatch between desktop settings and mobile/web against the same serve host.

## ELI5

Settings → Agents always scanned your laptop PATH even when the active host was a remote runtime. Agent detection now targets the active runtime host so the list matches what you can actually launch.

### DIFF
--- a/src/renderer/src/components/settings/AgentsPane.tsx
+++ b/src/renderer/src/components/settings/AgentsPane.tsx
@@ -5,7 +5,7 @@ import { useId, useMemo, useState } from 'react'
 import { Check, ChevronDown, ExternalLink, Info, RefreshCw, Terminal } from 'lucide-react'
 import type { GlobalSettings, TuiAgent } from '../../../../shared/types'
 import { getAgentCatalog, AgentIcon } from '@/lib/agent-catalog'
-import { useDetectedAgents } from '@/hooks/useDetectedAgents'
+import { type AgentDetectionTarget, useDetectedAgents } from '@/hooks/useDetectedAgents'
 import { useAppStore } from '@/store'
 import { Button } from '../ui/button'
 import { Input } from '../ui/input'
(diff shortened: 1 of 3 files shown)


brennanb2025:
Thanks for this, and apologies for the long silence — you deserved a reply well before now.

Your diagnosis was right: Settings → Agents was probing the local laptop PATH even when the active runtime pointed at a remote `orca serve` / Devbox host, so the list didn't match what you could actually launch.

What happened on our side: an internal PR (#9790) implementing the same fix merged on 2026-07-23, about eight hours after you opened this one, and nobody circled back to tell you. That lag is on us. The upside is that the problem you reported is fixed on `main` today — `AgentsPane` builds its detection target from the active runtime environment, and Quick Launch resolves its host through a shared hook in the same SSH → runtime → local order you proposed.

One note that's worth knowing for next time, more interesting than corrective: `OrchestrationSkillAgentCoverage` is intentionally pinned to the local host, and there's a test pinning it there. Skill-paired surfaces resolve their host through `getSingleFocusedRuntimeEnvironmentId` rather than raw `activeRuntimeEnvironmentId`, because a skill *scan* and a skill *install* have to name the same host — if those two diverge, the badge gets stuck on "Not installed". So that one surface deliberately doesn't follow the active runtime.

Really appreciate you finding this and writing it up clearly — the shape you picked is the shape that shipped.

brennanb2025:
Closing this as superseded. To be clear about why: your fix was correct, and the delay here was entirely on our side.

The problem you found was real — Settings → Agents probed the local PATH even when the active runtime pointed at a remote host, so the list was wrong for anyone working against an `orca serve` or Devbox host. #9790 landed the same fix about eight hours after you opened this, and nobody circled back to tell you. That is on us, and I am sorry you were left waiting.

The behaviour is live on main now, so the problem you reported is solved.

Thanks for the contribution, and please do send more — this was a good catch.

PR #9989, opened 2026-07-22
feat(agent-rows): label sidebar and dashboard rows with conversation names

brennanb2025:
## Why

Sidebar worktree-card and dashboard agent rows label with the last message sent to the agent, so rows relabel every turn and short replies such as "continue" become the row name. Tabs already carry stable conversation names through manual renames, quick commands, semantic session titles, generated titles, and agent-set live titles.

## What

Both row surfaces now prefer the conversation name and fall back to the last message when no usable name exists. There is no additional preference: row identity consistently follows tab identity.

Resolution (`src/shared/agent-row-conversation-name.ts`) mirrors tab-title precedence:

1. Manual tab rename (`customTitle`)
2. Quick-command label
3. OpenCode semantic session title (`OC | …`)
4. Generated title, only when the existing auto-generate-titles setting is on
5. Agent-set live title, accepted only when it carries a real name

Leading status decoration is stripped. Pure status titles, provider identity echoes, generic `Terminal N` placeholders, default titles, and POSIX/Windows/UNC path-like titles are rejected. Subagent child rows keep their own descriptions because the parent tab name would mislabel them.

## Performance

Mounted rows use a WeakMap-backed per-tab-array index. One immutable worktree tab array is indexed once and shared across rows, replacing repeated linear scans on store writes while allowing old arrays to be collected.

## Testing

- Resolver coverage for precedence, fallback, provider/status filtering, Windows/UNC paths, and default terminal placeholders
- Hook coverage for always-on behavior, stale row snapshots, subagent exclusion, generated-title gating, and one index build per immutable tab array
- Dashboard, sidebar, title, settings, localization, typecheck, lint/react rules, and max-lines suites pass locally (25 files / 227 tests)
- Full PR workflow runs lint, reliability gates, typecheck, Git compatibility, repository tests, unpacked build, and packaged CLI smoke

## Notes

- The generated-title source remains controlled by the existing auto-generate-title setting.
- Persisted values from the earlier opt-in iteration are ignored; no migration is required.
- The earlier manual-QA comment validated both behaviors while the toggle existed. Its conversation-name screenshots represent the final row behavior; the toggle itself has now been removed by product decision.


brennanb2025:
## Manual QA (dev build, CDP-driven, this branch)

Verified live on the rendered app (electron-vite dev build of this branch, isolated temp profile; seeded one workspace `prohealth-sync` with a working Claude status row `prompt="continue"`, tool line `Bash: pnpm test`).

**1. Default (toggle off) — rows label by last message sent** (unchanged behavior):

default last message

**2. Toggle on + agent-set title `✳ Fix patient intake sync` — row labels by conversation name** (decoration stripped, tool activity stays as secondary):

conversation name

**3. Manual tab rename `Prohealth intake spike` — rename wins over the agent title:**

manual rename

**4. Settings → Agents switch, findable via search ("conversation name"):**

settings toggle

Also verified: toggling **off** reverts the row to the last message immediately, even while a custom name exists (default stays last-message per product decision).

### Fix found during QA

Sidebar row data patches live entries in place and keeps the tab snapshot from row creation, so a rename/agent title landing *after* the row was built never reached the resolver. Pushed a839195 — the hook now reads the current tab from the store (retained rows keep the snapshot fallback), with a regression test.

### Not covered

- Dashboard surface (`DashboardAgentRow`) uses the identical hook/resolver but wasn't visually exercised in this pass.
- OpenCode `OC | …` and quick-command label sources are unit-tested only.

*(Screenshots hosted on the `qa-images/pr-9989` orphan branch; safe to delete after merge.)*

brennanb2025:
CI note for the always-on revision (`b75ab61575`): local row/title/settings suites pass (25 files / 227 tests), along with typecheck, localization, lint/react rules, and max-lines. The latest PR workflow stopped before typecheck/tests/build on `verify:skill-bundle-manifest` because current `main` contains the #10007 emulator skill-guide changes without refreshed `resources/skills/*` manifests. This PR does not touch skill files, and its previous head completed the full workflow successfully. I am leaving those unrelated release manifests out of this feature diff; rerunning after the base manifest is refreshed should clear the gate.

PR #9998, opened 2026-07-22
fix(browser): prevent invalid grab payloads on Zone.js pages

### LINKED ISSUE #9947: [Bug]: Browser Grab fails on Angular/Zone.js pages with "Guest returned invalid payload structure"
### Operating system

Windows

### Orca version

1.4.150.0

#### Short summary

`Grab Page Element` fails after clicking a normal visible element on an Angular 20 page that uses Zone.js:

```text
Grab failed: Guest returned invalid payload structure
```

The same Orca build successfully grabs an element from `https://example.com`, so this appears to be page-runtime/framework specific rather than a general browser-grab failure.

#### Environment

- Windows 11 Pro, 64-bit
- Windows version: 10.0.26200 (build 26200)
- Orca: 1.4.150.0
- Target application: Angular 20.3.25 with Zone.js enabled
- Page type: a private Angular application login page (the private IP, credentials, and cookies are intentionally omitted)

#### Steps to reproduce

1. Run or open an Angular 20.3.25 application with Zone.js enabled.
2. Open one of its pages in Orca's built-in browser.
3. Choose `Grab Page Element`.
4. Click a normal visible element on the page.
5. Observe `Grab failed: Guest returned invalid payload structure`.

Control check: repeat the same action on `https://example.com`; the grab succeeds.

#### Expected behavior

Orca should return the selected element payload and complete the grab on Angular/Zone.js pages.

#### Actual behavior

The click is captured, but Orca reports that the guest returned an invalid payload structure.

This is a normal selection, not a cancellation or teardown path.

#### Diagnostic evidence

On the affected Angular runtime, Zone.js has replaced the page-global `Promise` with a ZoneAwarePromise (minified constructor name `Z`). A minimal check using:

```js
const p = new Promise((resolve) => resolve({ page: {}, target: {} }));
```

produces:

```json
{
  "angular": "20.3.25",
  "promiseConstructor": "Z",
  "topLevelKeys": ["__zone_symbol__state", "__zone_symbol__value"],
  "wrappedKeys": ["page", "target"],
  "orcaWouldSeePage": false,
  "orcaWouldSeeTarget": false
}
```

`window[Zone.__symbol__('Promise')]` is present and points to the native Promise implementation.

#### Likely cause

The current grab flow calls:

```ts
await guest.executeJavaScript(buildGuestOverlayScript('awaitClick'))
```

The injected `AWAIT_CLICK_SCRIPT` creates its result with the guest page's global `new Promise(...)`. On a Zone.js page, that is a ZoneAwarePromise. It appears to cross the Electron `executeJavaScript` boundary as the wrapper object, whose direct keys are `__zone_symbol__state` and `__zone_symbol__value`, rather than as the resolved `{ page, target }` value. The main-side `clampGrabPayload` check then cannot find direct `page` and `target` fields and emits the reported error.

Relevant files:

- [grab-guest-script.ts](https://github.com/stablyai/orca/blob/main/src/main/browser/grab-guest-script.ts)
- [browser-grab-session-controller.ts](https://github.com/stablyai/orca/blob/main/src/main/browser/browser-grab-session-controller.ts)

#### Relation to the previous issue

This has the same final error text as #2783, but it is a different path from the cancellation case fixed by #3321:

- Orca 1.4.150.0 already contains the cancellation-marker guard.
- The error occurs after a normal element selection.
- The returned shape matches a ZoneAwarePromise wrapper rather than `{ __orcaCancelled: true }` or `{ message: "cancelled" }`.

#### Suggested fix / regression coverage

Please consider running the await-click code in an isolated world, or otherwise creating/returning the result through a trusted unpatched native Promise rather than the page-global Promise.

I would avoid fixing this by blindly unwrapping `__zone_symbol__value`, because page content is untrusted and can spoof those fields.

A regression test with a Zone.js/ZoneAwarePromise-patched page-global `Promise` should cover this case.

siddqamar:
## Summary

Fixes #9947 by handling Zone.js `ZoneAwarePromise` wrapping during browser grabs while preserving React component metadata and source-file information.

## Screenshots

No visual changes.

## Testing

- [x] pnpm lint
- [x] pnpm typecheck
- [x] pnpm test
- [x] pnpm build
- [x] Added regression tests for Zone.js Promise wrapping and React metadata preservation.
- [x] Focused Vitest: 63 tests passed.
- [x] Playwright regression tests discovered successfully.

## AI Review Report

Reviewed with GPT-5.6 Luna.

The review verified fixes for:

- Zone.js `ZoneAwarePromise` wrapping during browser grabs.
- React fiber metadata remaining accessible despite isolated-world execution.
- E2E coverage ensuring the page Promise patch remains active.

Cross-platform compatibility was verified on macOS, Linux, and Windows. No shortcut, path, shell, or platform-specific behavior changed.

## Security Audit

- Existing payload validation and secret redaction remain active.
- No new IPC, command execution, filesystem, authentication, or dependency risks.
- Temporary metadata attributes are removed after extraction.

## Notes

No visual changes or platform-specific follow-up required.

### DIFF
--- a/src/main/browser/browser-grab-screenshot.ts
+++ b/src/main/browser/browser-grab-screenshot.ts
@@ -1,5 +1,6 @@
 import type { BrowserGrabRect, BrowserGrabScreenshot } from '../../shared/browser-grab-types'
 import { GRAB_BUDGET } from '../../shared/browser-grab-types'
+import { executeBrowserGrabScript } from './browser-grab-script-executor'
 
 const HIDE_BROWSER_GRAB_OVERLAY_SCRIPT = `(function(){
   var g = window.__orcaGrab;
@@ -44,12 +45,12 @@ export async function captureSelectionScreenshot(
     // label don't appear in the screenshot. The overlay is restored after.
     // Wrapped in try/finally so the overlay is always restored even if
     // capturePage() throws (e.g., guest destroyed mid-capture).
-    await guest.executeJavaScript(HIDE_BROWSER_GRAB_OVERLAY_SCRIPT).catch(() => {})
+    await executeBrowserGrabScript(guest, HIDE_BROWSER_GRAB_OVERLAY_SCRIPT).catch(() => {})
     let image: Electron.NativeImage
     try {
       image = await guest.capturePage()
     } finally {
-      await guest.executeJavaScript(RESTORE_BROWSER_GRAB_OVERLAY_SCRIPT).catch(() => {})
+      await executeBrowserGrabScript(guest, RESTORE_BROWSER_GRAB_OVERLAY_SCRIPT).catch(() => {})
     }
     if (image.isEmpty()) {
       return null
--- a/src/main/browser/browser-grab-script-executor.test.ts
+++ b/src/main/browser/browser-grab-script-executor.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, it, vi } from 'vitest'
+import { BROWSER_GRAB_WORLD_ID, executeBrowserGrabScript } from './browser-grab-script-executor'
+
+describe('executeBrowserGrabScript', () => {
+  it('runs grab scripts in the dedicated isolated world', async () => {
+    const executeJavaScriptInIsolatedWorld = vi.fn().mockResolvedValue({ page: {}, target: {} })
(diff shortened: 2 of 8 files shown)


siddqamar:
I noticed this has already been resolved and merged in #9995. Closing this PR. Thanks!

Issue #9949, opened 2026-07-22
Intermittent cross-worktree PTY teardown when opening an embedded browser


bbingz:
## Summary

Opening an **embedded browser tab** inside Orca mass-killed **every agent PTY session across all worktrees** in a single ~1.5 s burst. The app did **not** quit, relaunch, crash, or reload — the main window stayed up the whole time. All running agents were lost mid-work.

This is **not** the auto-update path (stablyai/orca#8261): there was no update install, no relaunch, and the renderer process kept running continuously across the kill.

## Environment

* Orca **1.4.150** (build 1.4.150), macOS arm64, `/Applications/Orca.app` (packaged, not `--serve`)
* Shared local daemon (protocol v24), ~20 worktrees with live agent sessions

## Steps to reproduce (reported)

1. Have many worktrees open, each with a live agent terminal (Claude/Codex/etc.).
2. Open an **Orca embedded browser tab** (new web-page tab inside the app).
3. → All agent terminals across **every** worktree die at once; the app keeps running.

Reporter is confident the main window did **not** restart or flash and no update prompt appeared.

## Evidence (from `~/Library/Application Support/orca/logs`)

**daemon.log** — 25 sessions force-killed in ~1.5 s, spanning every worktree (session names redacted):

```
14:04:33.211  session-killed  <repo>/…@@…  immediate=true
14:04:33.300  session-killed  <repo>/…@@…  immediate=true
14:04:33.407  session-killed  <repo>/…@@…  immediate=true
...            (25 session-killed total, ~20 distinct worktrees)
14:04:33.721  session-killed  <repo>/…@@…  immediate=false   ← graceful close first…
14:04:33.927  session-killed  <repo>/…@@…  immediate=true     ← …then force-kill (two-phase)
14:04:34.417  session-killed  <repo>/…@@…  immediate=false
14:04:34.585  session-killed  <repo>/…@@…  immediate=true
14:04:34.9xx  session-exited  …  code=0     (for all)
```

The mixed signature — a **graceful** `immediate:false` **close-wave followed by forced** `immediate:true` **kills** — matches `runKillAllTerminalSurfaces()` (close terminal surfaces, then `killDaemonSessions()`), i.e. an app-surviving *global* daemon kill, **not** a per-worktree teardown (which force-kills only) and **not** `killAllPty()` (quit/updater only).

**Ruled out:**

[...]

Issue #8261, opened 2026-07-11
[Bug]: App silently quits without any warning when auto-update installs — kills all active PTY sessions mid-work


iridite:
## Summary

When Orca's auto-updater finishes downloading a new version in the background, it immediately quits and installs without showing any confirmation dialog or warning to the user. All active terminal sessions, agent sessions, and PTY processes are forcibly killed at the moment of quit. From the user's perspective the app simply disappears mid-work with no explanation.

---

## Environment

- **OS**: macOS 26.4 (Darwin 25.4.0 arm64e)
- **Hardware**: MacBook Air M5, 24 GB RAM (Mac17,3)
- **Orca versions where this was observed**: v1.4.134 install (from v1.4.128), v1.4.135 install (from v1.4.134)

---

## Confirmed Occurrences

Both events were captured in `~/Library/Application Support/orca/logs/main.trace.ndjson` as `updater.install` spans.

### Occurrence 1 — Install of v1.4.134

**Timestamp**: 2026-07-10 23:00:08 CST  
**Running version at time of quit**: v1.4.128  
**Installed version**: v1.4.134  

Event chain recorded in trace:
```
pre_quit_cleanup_start
pre_quit_cleanup_done
native_quit_and_install_invoked
local_pty_kill_all              ← all terminal/PTY sessions forcibly killed
window_close_listeners_removed
```

The session that was running at the time (started 07-10 22:40, 20 minutes prior) was terminated with no warning.

### Occurrence 2 — Install of v1.4.135

**Timestamp**: 2026-07-11 13:51 CST  
**Running version at time of quit**: v1.4.134  
**Installed version**: v1.4.135  

Same event chain. The session had been running for 64 minutes with an active agent (heap at 87 MB, mid-task).

---

## Update Download Behavior

The updater downloads silently in the background. Evidence from the updater cache:

```
~/Library/Caches/orca-updater/pending/Orca-1.4.135-arm64-mac.zip  (189 MB, downloaded 2026-07-11 14:55)
~/Library/Caches/Homebrew/Cask/orca--1.4.133.dmg
~/Library/Caches/Homebrew/Cask/orca--1.4.128.dmg
```

As of 2026-07-11 14:55, a new update zip is already staged in the pending directory. This means another silent quit-and-install is imminent whenever Orca decides to apply it.

---

## Impact

1. **All PTY sessions killed** (`local_pty_kill_all`) — any running terminal commands, build processes, or agent tasks are lost without warning
[...]

PR #10013, opened 2026-07-22
fix(terminals): attribute mass terminal kills and confirm bulk orphan sweep (#9949)

bbingz:
Addresses #9949 — implements both asks from the issue: make every mass-terminal-kill path attributable in one log read, and put a confirmation gate on the one bulk kill that had none.

## Background

On 2026-07-22T14:04:33Z, ~25 daemon PTY sessions across ~20 worktrees were killed in ~1.5 s by a sequential per-terminal close storm executed through the desktop app, while the app itself stayed alive (continuous renderer heartbeat, no daemon restart, no new daemon client). Post-incident forensics ruled out the updater (#8261), Manage-Sessions/status-bar killAll, mass worktree sleep (zero keepHistory checkpoints, no sleep records), and agent hibernation (feature flag off). The kill signature matches per-session `pty:kill` (`immediate:true`) plus a `terminal.close` → `RuntimePtyController.kill` (`immediate:false`) leg for runtime-tracked terminals.

The investigation dead-ended on **attribution**: the daemon's `session-killed` log carries no client/request identity, none of the renderer's mass-kill entry points leave a breadcrumb, and runtime RPC dispatches (orca-cli / paired devices — two were active that day) log nothing durable. This PR closes that gap so the next occurrence is diagnosable from the logs directly.

## Changes

**Daemon** (`daemon-server.ts`): the `session-killed` log line now includes `clientId` and `requestId`. (`keepHistory` is not in the kill protocol payload; the protocol is untouched.)

**Renderer** — one best-effort `terminal_mass_kill` crash-breadcrumb per invocation (never per session) at every mass-kill entry point, landing in the persisted `renderer.breadcrumb` trace:

| source | site |
|---|---|
| `kill-all-surfaces` | `runKillAllTerminalSurfaces` (injectable dependency, tested) |
| `worktree-sleep` | `runSleepWorktrees` (covers UI multi-sleep and externally-relayed `ui:sleepWorktree`; single producer, no relay-site duplicate) |
| `hibernation` | `runAgentHibernationTick` when candidates > 0 |
| `orphan-sweep` | bulk orphan kill on confirm |

**Runtime RPC** (`runtime-rpc.ts`): dispatches of kill-capable methods (`terminal.close/closeTab/stop/stopExact`, `session.tabs.close`, `agentTeams.tmuxCompat`, `worktree.sleep/rm`) log one attribution line with the calling device id/name/scope on **both** transports — WebSocket (paired devices) and the local socket (orca-cli), the latter with a synthetic `local-socket` origin. Ids-only target summary, no payload dumps.

**Resource Manager** (`ResourceUsageStatusSegment.tsx` + new `orphan-terminal-kill-gate.ts`): the bulk "Kill N orphan terminals" button no longer kills on click. It stages the orphan snapshot and opens a confirmation dialog; kills run only from confirm, which **re-filters against live tab bindings** so a session adopted by a tab while the dialog sits open is never swept (the no-confirm path was the #8459 hazard). Per-row single-session kill is unchanged. New strings localized in es/ja/ko/zh following the file's existing conventions.

## Out of scope

No kill-semantics or ordering changes; no daemon protocol changes; the underlying external trigger of the 2026-07-22 incident remains unidentified — this instrumentation is what identifies it next time.

## Testing

- 103 tests pass across the 6 touched/new test files, including: daemon attribution (red-green verified), WS + local-socket audit emission, negative test on a read-only allowlisted method, breadcrumb count assertions, orphan-gate stages-without-killing, rebind-before-confirm never kills, hibernation tick candidates>0/=0.
- `tsc` node + web projects: no errors. `oxlint`: clean. max-lines ratchet: no new bypasses. Localization catalog + locale parity: pass (coverage check fails only on 3 pre-existing `ai-vault` strings present on the base commit).
- Reviewed by a multi-agent pass (correctness + hygiene + Codex); all 18 confirmed findings fixed, including auditing the local-socket transport, the `agentTeams.tmuxCompat` kill path, and the confirm-time rebind race.


bbingz:
Updated per @AmethystLiang's recommendations on the issue (https://github.com/stablyai/orca/issues/9949#issuecomment — 18:09Z comment):

- **"log a reason/correlation ID so a repeat tells us exactly which action initiated every kill"** — covered: daemon `session-killed` now carries `clientId` + `requestId`; every renderer mass-kill entry emits a `terminal_mass_kill` breadcrumb; and as of 3f6e39f the kill-capable RPC dispatch audit is **persisted through the durable crash-breadcrumb store** (which also lands in the trace file), not just console — a packaged GUI launch discards stdout, which is exactly where the original incident ran.
- **"guard both global and individual PTY-kill IPC paths"** — the global paths are guarded/attributed here (incl. the previously confirmation-less bulk orphan sweep). Per-call `reason` threading through the individual `pty:kill` IPC was deliberately left out to keep this diff surgical; happy to do it as a follow-up if you want it in one go.
- **"E2E tripwire asserting browser actions emit zero PTY-kill RPCs"** — agreed and not in this PR; it needs an IPC-observation hook in the e2e harness that doesn't exist yet. Sketch: expose a test-only counter on `pty:kill`/`terminal.close` dispatches, drive N embedded-browser tab opens via the existing electron-headless project, assert counter stays 0. Can file a follow-up issue/PR for it.

Also `agentTeams.tmuxCompat` is now audited only for `kill-pane`/`kill-window`/`kill-session` subcommands so the high-volume compat polling (`list-panes` etc.) doesn't spam the audit.

bbingz:
Rebased onto current `origin/main` (`72a0a4dbb8`).

Conflict resolution (2nd commit only):
- `runtime-rpc.ts`: kept main's `DeviceEntry` import and this PR's `recordDurableCrashBreadcrumb` import
- `runtime-rpc.test.ts`: kept main's `DEVICE_REGISTRY_FILENAME` / `E2EE_KEYPAIR_FILENAME` imports and this PR's `getCrashBreadcrumbSnapshot` import

Focused tests green (109): kill-attribution, runtime-rpc, orphan-kill-gate, kill-all surfaces, sleep-worktree, hibernation coordinator.

Heads: `2b585c933a` + `b1a6dd069f` (ahead=2 behind=0 vs main).

bbingz:
## Restack note

Rebased onto latest `main` (force-with-lease). Both commits applied cleanly.

**Scope (unique vs main):**
- Daemon `session-killed` + client/request attribution for mass-kill diagnosis (#9949)
- Confirm gate + rebind re-filter for bulk orphan sweep
- Kill-capable RPC durable audit

**Coordinates with (do not duplicate UX):**
- **#9752** — Windows Job Object *how* trees die after authorized kill
- **#8467** — Resource Manager *review which unbound shells are idle* before cleanup (classification UI)
- **#8888** — soft-deny / rate-limit *policy* for unattributed destructive closes (not a second orphan-confirm dialog)

Prefer one confirm surface for orphan/unbound bulk actions: this PR’s confirm gate for orphan sweep; #8467’s review dialog for idle-shell classification. #8888 should soft-deny at the protocol layer without adding a third confirm modal.

OrcaWin:
Superseded by #10129. The deterministic mixed-generation harness reproduces the burst through sequential reasonless `session.tabs.close` requests from a stale paired desktop viewer; the orphan-sweep confirmation guard in this PR does not intercept that RPC path. #10129 refuses the ambiguous legacy runtime request while preserving explicit current-runtime and legacy-mobile closes, and intentionally omits these diagnostics. Please review #10129 rather than merging this PR.

bbingz:
Closing in favor of #10129 per @OrcaWin's review — agreed that the orphan-sweep confirmation guard here does not intercept the `session.tabs.close` RPC path that actually produces the burst, so #10129 is the correct fix.

Noting for the record that the kill-attribution diagnostics in this PR (daemon-side attribution audit + the `kill-all` confirm gate) are intentionally omitted from #10129. Happy to re-propose them as a separate, smaller PR if they are wanted once #10129 lands — otherwise this can stay closed.

PR #10012, opened 2026-07-22
fix(terminal): bound parse-deferred remote flow

OrcaWin:
## Summary

This is a direct-to-`main` follow-up to merged #9683 and commit `b606117ee`. It preserves the parsed-or-intentionally-discarded ACK contract and adds the bounded flow-control, cleanup, fairness, and performance work needed for production remote terminals.

- Fan out one delivery credit across every scheduler write; ACK only after every child parses or intentionally discards.
- Batch renderer ACKs at 192 KiB or 4 ms.
- Adapt per-stream credit from 512 KiB to 2 MiB and aggregate credit from 2 MiB to 8 MiB.
- Cap pending output at 256 KiB per stream and 32 active/pending streams per connection.
- Drain pending streams round-robin so bulk output cannot starve interactive terminals.
- Close/release ownership on malformed frames, renderer callbacks, transport writes, serialization failures, disposal, stale generations, resubscribe replacement, and reconnect races.

Receipt-time ACK is not used.

## Deterministic reproduction

Paired renderer replay plus live-output tests hold delivery credit while xterm has not completed parsing, then exercise replay/live ordering, transformed frames, hidden panes, late callbacks, disposal, reconnection, and transport failure. The affected Windows 2 Setup PTY independently reproduced blank/stale replay, hidden queued input, and duplicated fragments while the underlying ConPTY remained writable.

## Reliability coverage

Deterministic tests cover replay ordering, stale generations, malformed and transformed frames, queue eviction, hidden panes, pane disposal, recovery serialization failure, repeated same-slot subscribe, renderer delivery failure, ACK transport failure, late parser callbacks, reconnect, stream caps, bounded pending memory, and round-robin fairness.

## Benchmark

Opt-in command:

`ORCA_TERMINAL_PERF_BENCH=1 pnpm exec vitest run --config config/vitest.config.ts --disableConsoleIntercept src/main/runtime/rpc/terminal-multiplex-flow-control.bench.test.ts`

| RTT | viewers | aggregate MiB/s | per-viewer MiB/s | max in flight |
|---:|---:|---:|---:|---:|
| 1 ms | 1 | 1828.6 | 1828.6 | 2 MiB |
| 1 ms | 4 | 7314.3 | 1828.6 | 8 MiB |
| 1 ms | 8 | 7641.8 | 955.2 | 8 MiB |
| 20 ms | 1 | 94.0 | 94.0 | 2 MiB |
| 20 ms | 4 | 375.9 | 94.0 | 8 MiB |
[...]

PR #10014, opened 2026-07-22
fix(types): exclude undefined stat overloads

OrcaWin:
## Summary

- Exclude undefined from the promise stat overload return type in the two affected type boundaries.
- Restore typecheck after the locked @types/node update without changing runtime behavior.

## Reproduction

After a fresh pnpm install --frozen-lockfile on current main, pnpm typecheck reports TS18048 in worktree-common-git-directory.ts and skill-installation-topology.ts because ReturnType<typeof stat> includes the callback overload nullable branch in the updated declarations.

## Validation

- Fresh locked dependency install reproduced the failure before this change.
- pnpm typecheck passes after this change.
- Scoped oxlint and formatting pass.
- Runtime behavior is unchanged; both functions still receive successful stat results only.

### DIFF
--- a/src/main/ipc/worktree-common-git-directory.ts
+++ b/src/main/ipc/worktree-common-git-directory.ts
@@ -7,7 +7,7 @@ import {
 } from '../../shared/cross-platform-path'
 import type { FileStat } from '../providers/types'
 
-type GitDirectoryStat = Awaited<ReturnType<typeof stat>> | FileStat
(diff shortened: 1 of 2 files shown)


OrcaWin:
Superseded by the two direct-to-main PRs #10011 and #10012. The two-line Node declaration compatibility commit is present in both branch histories so their refreshed-dependency CI can pass; it will naturally disappear from the remaining diff after either lands.

PR #9997, opened 2026-07-22
feat(dashboard-popout): size the terminal dialog's PTY to the dialog grid

brennanb2025:
## What

The experimental agent-dashboard popout's terminal dialog previously rendered the pane's serialized frame at its **original** cols/rows and CSS-scaled it down to fit — shrunken, blurry, and sized by the main window rather than the dialog. The dialog now **claims the PTY grid for its own box**: the PTY reflows to the dialog's dimensions and renders crisp and full-size, exactly like a normal terminal.

## How

- Reuses the existing **remote-desktop viewer registry** (`updateRemoteDesktopViewer` / `unregisterRemoteDesktopViewer`) with a synthetic `dashboard-popout:<contentsId>` viewer key — the dialog is just another remote viewer of the shared PTY:
  - main-window pane parks under the established `remote-desktop-fit` hold (no lock banner, input still shared), and reclaims its geometry when the claim is released (dialog close, unsubscribe, or popout window destruction);
  - a phone-driven PTY keeps the floor — the claim politely loses and the dialog falls back to the previous scale-to-fit rendering.
- New popout-gated `terminalPreview:fit` IPC; claims tracked per webContents so they can never leak past the window's lifetime.
- `TerminalPreviewOutputStream.requestResync()` + a `subscribeToTerminalResize` hook: **any** grid change under a live preview stream (fit landing, host reclaim, phone takeover) pushes a resync so the dialog repaints from a snapshot at the new grid. This also fixes a pre-existing bug where the open dialog garbled if the pane resized under it.
- Renderer `preview-grid-claim.ts`: measures the cell size from `.xterm-screen` layout dims (transform-immune), computes the grid the box holds (clamped to the runtime's 20–240 × 8–120 viewport range), debounced 200 ms, and keyed by target dims — an unchanged target is never re-requested, so a host/phone taking the grid back cannot start a resize tug-of-war.
- `preview-clipboard-shortcuts.ts` extracted verbatim from `AgentTerminalPreview` (max-lines cap).

## Evidence (live Electron QA, real zsh PTY)

`stty size` in the shared PTY proves the grid at each step.

**1. Before — main pane owns the grid (58×135):**

**2. Popout board with the agent card:**

**3. Dialog open — PTY reflowed to the dialog grid (33×107), terminal fills the dialog at scale 1 (screen 902×528 in a 906×536 box, no CSS transform):**

**4. Main pane during the hold — same content at the held grid, no garbling, no lock banner:**

[...]

PR #10018, opened 2026-07-23
fix(types): concrete Stats over Awaited<ReturnType<typeof stat>> (unbreaks main verify)

nwparker:
## Description
`main`'s `verify` job is currently red for **every** freshly-run PR (confirmed on unrelated PRs #10012, #9997, plus the in-flight perf PRs #9841/#9849/#9882) with:

```
src/main/ipc/worktree-common-git-directory.ts(18,20): error TS18048: 'value' is possibly 'undefined'.
src/main/skills/skill-installation-topology.ts(33,25): error TS18048: 'fileStat' is possibly 'undefined'.
… (8 errors total)
```

### Root cause
`main`'s lockfile pins `@types/node@25.9.5`, whose `fs.promises.stat` overloads make `Awaited<ReturnType<typeof stat>>` resolve to a **nullable** type (the alias picks the last overload's return, unlike a direct `stat()` call which resolves the concrete first overload). In param/type-alias positions there's no control-flow narrowing, so member access trips `TS18048`. It reproduces only in a fresh dep install (CI), which is why it slipped in.

### Fix
Use the concrete, non-nullable `Stats` type (from `node:fs`) in the two affected spots. No behavior change — `stat()`/`lstat()` calls still resolve to `Stats` at runtime.

## Proof of zero regression
- Reproduced locally with `@types/node@25.9.5`: pre-fix shows the exact 8 `TS18048` errors; **with the fix the full `typecheck:node` is 0 errors**.
- `Stats` carries every accessed member (`isDirectory`, `isFile`, `dev`, `ino`, `isSymbolicLink`), so callers are unchanged.
- Other files using the same alias were checked — they don't error (their vars are control-flow-narrowed from a concrete `await stat()`), so no wider change is needed.

## Why it matters
Unblocks `verify` for all open PRs, including the perf-scan follow-ups #9841 / #9849 / #9882.

### DIFF
--- a/src/main/ipc/worktree-common-git-directory.ts
+++ b/src/main/ipc/worktree-common-git-directory.ts
@@ -1,3 +1,4 @@
+import type { Stats } from 'node:fs'
 import { readFile, stat } from 'node:fs/promises'
 import type { Repo } from '../../shared/types'
 import {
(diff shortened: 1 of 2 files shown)


nwparker:
Superseded by #10017 (fix(types): use concrete Node stat result type), which landed the same Stats fix on main. Closing as redundant.

PR #10037, opened 2026-07-23
fix(startup): mirror startup diagnostics to a file, tag pid

jeenoo-aidas:
## Summary
- Add a file sink for startup diagnostics (`configureStartupDiagnosticsLogFile`) alongside the existing stderr write — a packaged Windows GUI-subsystem exe launched without an inherited console (double-click, Start Menu) has no reliable stderr to redirect, which was blocking diagnosis of a Windows duplicate-instance report.
- Tag `pid` on the `before-single-instance-lock` and `single-instance-lock-result` diagnostic events so concurrent instances are distinguishable in the shared log file.

## Test plan
- [x] `tsc --noEmit -p config/tsconfig.node.json`
- [x] `vitest run src/main/startup/startup-diagnostics.test.ts src/main/startup/single-instance-lock.test.ts`

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -110,6 +110,7 @@ import {
 import { startEventLoopStallProbe } from './startup/event-loop-stall-probe'
 import { startMainThreadChurnProbe } from './diagnostics/main-thread-churn-probe'
 import {
+  configureStartupDiagnosticsLogFile,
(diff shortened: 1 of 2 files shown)


jeenoo-aidas:
실수로 프로세스를 건너뛰고 바로 열었습니다 — 닫습니다.

Issue #8739, opened 2026-07-14
[Bug]: Linear filters only show options from the first selected team


limineol:
### Operating system

macOS

### Orca version

1.4.139

### Details

1. Open Tasks → Linear issues in a workspace with multiple teams.
2. Select All teams, or select two teams.
3. Open Filters and inspect status, assignee, or label options.

Only options from the first selected team appear. Filter options should include metadata from every selected team.

nwparker:
## 🤖 Agent handoff — reproduction kit

**Status:** REPRODUCED · GitHub label `has_repro`  
**Writeup (branch):** [`docs/bug-reproductions/8739.md`](https://github.com/stablyai/orca/blob/nwparker/bug-repro-handoffs/docs/bug-reproductions/8739.md)  
**Branch with tests + docs:** [`nwparker/bug-repro-handoffs`](https://github.com/stablyai/orca/tree/nwparker/bug-repro-handoffs)

### Summary
With multiple Linear teams selected (or All teams), Filters → status / assignee / label only list metadata from a single **primary** team. Other selected teams' states, labels, and members never appear.

### Automated re-run
```bash
git fetch origin nwparker/bug-repro-handoffs
git checkout nwparker/bug-repro-handoffs
# or: git worktree add /tmp/orca-repro-8739 origin/nwparker/bug-repro-handoffs

pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/repro-8739-linear-filter-first-team.test.ts
```

### Code / evidence files
- [`src/renderer/src/components/repro-8739-linear-filter-first-team.test.ts`](https://github.com/stablyai/orca/blob/nwparker/bug-repro-handoffs/src/renderer/src/components/repro-8739-linear-filter-first-team.test.ts)

### How another agent should use this
1. Check out `nwparker/bug-repro-handoffs` (or cherry-pick the `repro-*` test for this issue).
2. Run the command(s) above — green **documents** the bug (or `cannot_repro` proves current tree does not show it).
3. Implement the fix on a **new** branch from `main`; keep/adapt the repro test as a regression test that should flip behavior or stay as a guard.
4. Open a PR with: **Description**, **Evidence**, **ELI5**, **User-regression-tradeoffs**.

_Posted by bug-repro orchestration so work is not trapped in a local worktree only._

innocarpe:
I'd like to take this.

[...]

PR #10042, opened 2026-07-23
fix(linear): union filter options across every selected team

innocarpe:
## Summary

On Tasks → Linear issues, selecting multiple teams (or "All teams") now lists Filter → Status / Assignee / Label options from *every* selected team, deduped by id, instead of only the primary team's.

## ELI5

Each Linear team has its own statuses, labels, and people. If you pick two teams, the filter menu should show both teams' options — but Orca only asked the first team. Now it asks every selected team and merges the lists.

## Root cause & fix

The attribute filter dropdowns resolved a single `primaryTeam` and loaded metadata only for that team, showing an "Options from {team}" banner while ignoring the rest of the multi-select. The fix adds `resolveLinearIssueAttributeFilterTeamIds()` (name-stable ordering, falls back to the primary team on empty selection, drops ids absent from `availableTeams`) and `unionLinearMetadataById()` (dedupe by id, first-seen order), plus `useTeamsStates/useTeamsLabels/useTeamsMembers` hooks that fan out through the existing per-team cache stores and union the results. `TaskPage` now passes `selectedTeamIds` + `availableTeams`, active filter pills prune against the unioned option set, and the stale banner is removed.

## Evidence

- Rebase: clean onto `origin/main` (base `8a23618310`, head `34dbadfb60`).
- Tests: `linear-issue-attribute-filter-team-ids.test.ts` and `useIssueMetadata.test.tsx` — **2 files, 11 tests passed** on branch.
- Fix-reverted control: the #8739 tests fail without the added code — `team-ids.test.ts` cannot resolve the new module, and `useIssueMetadata.test.tsx` "unions workflow states across every selected team (#8739)" throws `TypeError: useTeamsStates is not a function`.
- Lint: `oxlint` on the 4 changed source files — EXIT 0, clean.

```
(a) On branch:  Test Files 2 passed (2), Tests 11 passed (11)
(b) Reverted:   Test Files 2 failed (2), Tests 1 failed | 4 passed (5)
      useIssueMetadata.test.tsx › unions workflow states across every
      selected team (#8739): TypeError: useTeamsStates is not a function
```

## Trade-offs

None — pure fix. Filter lists grow under multi-team selection, which is the intended behavior; results are deduped by id and reuse the same per-team TTL cache as the single-team hooks.

## Regression risk

Effectively zero. The change is additive (new module + new hooks); single-team paths keep their behavior via the empty-selection fallback, and the pre-existing single-team cache/TTL tests stay green. `useTeamsMetadataList` keys effects on a joined cache key so freshly-passed arrays don't thrash, and an active-key ref guards stale async results.

*Credit to original author @innocarpe. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/renderer/src/components/TaskPage.tsx
+++ b/src/renderer/src/components/TaskPage.tsx
@@ -8618,7 +8618,8 @@ export default function TaskPage(): React.JSX.Element {
                             workspaceId={selectedLinearWorkspaceId ?? null}
                             isAllWorkspaces={selectedLinearWorkspaceId === 'all'}
                             primaryTeam={linearAttributePrimaryTeam}
-                            selectedTeamCount={linearTeamSelection.size}
+                            selectedTeamIds={[...linearTeamSelection]}
+                            availableTeams={linearTeamOptions}
                             settings={linearTaskSourceContext ?? settings}
                           />
                         ) : null}
--- a/src/renderer/src/components/linear-issue-attribute-filter-dropdowns.tsx
+++ b/src/renderer/src/components/linear-issue-attribute-filter-dropdowns.tsx
@@ -4,7 +4,7 @@ import React, { useEffect, useMemo, useRef, useState } from 'react'
 import { ListFilter, X } from 'lucide-react'
 import { Button } from '@/components/ui/button'
 import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover'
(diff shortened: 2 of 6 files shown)


nwparker:
Thanks so much for this, @innocarpe — union filter options across every selected team. Reviewed, rebased on latest `main`, and merged. Really appreciate the contribution! 🙏

Issue #9993, opened 2026-07-22
[Bug]: `?997;1n` color-scheme replies leak into fish and corrupt stdin of interactive prompts (repro on v1.4.152)


ferpart:
### Summary

Follow-up to stablyai/orca#8128 / stablyai/orca#8206. The fix from stablyai/orca#8206 is present in my build  <br>(v1.4.147 contains merge commit `4df9457`), but fish sessions still get  <br>`CSI ?997;1n` color-scheme replies leaked as visible text. The token  <br>typically arrives doubled:

```
^[[?997;1n^[[?997;1n
```

This happens in **both** regular (non-floating) terminal panes and the  <br>global floating terminal, so it does not appear to be limited to one  <br>pane type.

### Steps to reproduce

Same setup as stablyai/orca#8128:

1. fish 4.7.1 as the default shell, Tide v6 prompt
2. Use either a regular terminal pane or the global floating terminal
3. Interact normally (type commands, press Enter, run short-lived commands)
4. `^[[?997;1n^[[?997;1n` intermittently appears in the terminal output

### Expected behavior

The `CSI ?997;1n` reply to fish's `CSI ?2031h` subscription is delivered  <br>while fish still owns its read window (or not at all after `CSI ?2031l`),  <br>and never appears as visible text.

### Environment

* **Orca**: 1.4.147
* **OS**: macOS 26.2
* **Shell**: fish 4.7.1, Tide v6, default `xterm-256color` terminfo

### Notes that may help triage

* Verified via the GitHub compare API (`v1.4.147...4df9457`) that the  <br>stablyai/orca#8206 fix (immediate replies in `terminal-mode-2031-replies.ts`) is  <br>included in this build, so this is either a second reply path or a case  <br>the timing fix doesn't cover.
* The reply arriving **doubled** suggests two responders answering the same  <br>`?2031h` subscription. stablyai/orca#8206 only touched the visible terminal-pane  <br>responder; other mode-2031/reply paths exist and were not modified, e.g.  <br>`parked-terminal-mode2031-responder.ts`, the hidden delivery gates  <br>(`pty-hidden-delivery-gate.ts` / `terminal-hidden-delivery-gate.ts`), and  <br>the daemon-side `terminal-view-attribute-responder.ts`.
* There is no shell-side workaround: fish enables `?2031h`/`?2031l`  <br>unconditionally around each prompt in core (`src/tty_handoff.rs`), with  <br>no config option to disable it.

rodboev:
[...]

PR #10054, opened 2026-07-23
fix(terminal): stop doubled ?997;1n color-scheme reply leaking into fish

brennanb2025:
Fixes STA-2385.

fish 4.7.1 toggles CSI ?2031h and CSI ?2031l around terminal ownership. Per the color-scheme protocol, enabling mode 2031 subscribes to future palette-change notifications; it does not request an immediate CSI ?997...n report. Orca seeded a report on every subscription, sometimes through more than one lifecycle path, so a report could arrive after fish stopped reading terminal responses and render as literal text.

Fix:
- make mounted, hidden-gated, parked, and remote-runtime mode-2031 subscriptions record-only
- propagate both subscribe and unsubscribe final state through local and daemon fact channels
- reserve CSI ?997...n for actual subscribed color-mode transitions and explicit CSI ?996n queries
- remove the eight-attempt subscription retry timer and the parked subscription responder sidecar
- stop classifying CSI ?2031h as a reply-eliciting query during remote snapshot replay

Reliability contract:
- invariant: a mode-2031 enable produces zero PTY writes; only a later real color-mode transition produces one immediate notification while the subscription remains active
- regression oracle: unit tests assert zero ordinary and immediate writes on subscribe, unsubscribe cleanup, split-chunk and final-state fact propagation, replay silence, same-mode suppression, and real flip notification
- provider/platform scope: local, daemon, SSH/WSL fact relays, and remote-runtime paths share the protocol handling; no platform-specific path or new polling was added
- performance: removes retry timers and an always-live parked PTY subscriber
- residual gap: CI has protocol-level coverage but no live fish 4.7.1 E2E

Validation:
- relevant local Vitest suites: 1,531 tests passed
- local pnpm typecheck:web, pnpm typecheck:cli, reliability gates, max-lines ratchet, changed-file oxlint, and diff checks passed
- local repository-wide pnpm typecheck and pnpm lint were run; the installed local toolchain stopped on unrelated existing files outside this diff
- GitHub verify passed full lint, typecheck, tests, build, and packaged CLI smoke; Windows restart, crash-survival, Ubuntu/Windows native smoke, and Wayland input checks also passed


nwparker:
@brennanb2025 Coming back to this a year late — you were right, and this sat unmerged for the wrong reason.

I fixed #9993 today in #13904 and independently landed on your conclusion: make mode 2031 subscriptions silent. `git log -S` turned up your `f84fedb` ("fix(terminal): make mode 2031 subscriptions silent") after I'd already decided it, which was a useful confirmation.

Worth recording why this one probably stalled, since the fix direction was right. The commit message here says:

> fish probes color-scheme with two sequences (CSI ?2031h subscribe + CSI ?996n query). Orca answered BOTH with an identical CSI ?997;1n, so the reply doubled

That premise doesn't hold: **fish never sends `?996n`** — 0 occurrences across a full instrumented session. Its per-prompt probe burst is OSC 11, CPR, DA1. So the stated mechanism for the doubling was wrong even though the remedy was correct, and I'd guess that's what made it hard to trust.

The actual justification is simpler and checkable: replying to `?2031h` at all is an Orca extension. Upstream xterm.js sets `colorSchemeUpdates = true` and emits nothing (`InputHandler.ts:2035`); the protocol's one-shot query is `?996n`; the defined push is on an actual color change. And the reply can't win fish's ~1 ms subscribe→withdraw window across a renderer hop — measured deterministic leak at ≥1 ms, with the reply *not* stale by Orca's own view when written, which is why the later stale-reply gate proposals couldn't work either.

The real doubling, for the record, is xterm's own theme-write report plus `maybePushMode2031Flip` both firing on a color flip — two emitters colliding on a *change*, not on a subscribe. #13904 makes `maybePushMode2031Flip` the single owner.

Your diff also caught something I initially got wrong. You removed `?2031h` from the replay grammar in `terminal-reply-query-scan.ts`; I kept it, reasoning a late-attaching remote client would otherwise lose the subscription entirely since snapshots drop control bytes. Review then showed keeping the arm *without* its withdraw reproduces #9993 on the remote path — a phone attaching to an idle fish session registers a subscription fish already retired. The answer was neither yours nor mine: carry **both** toggles so the replayed state is truthful.

Closing this in favor of #13904, which has the real-fish regression test this needed to be merge-able. Sorry it took this long to come back around — you had it first.

nwparker:
Superseded by #13904, now merged as `5ea7df1a5b8` — issue #9993 is closed.

Closing this properly; my earlier comment said I was closing it and I only left the comment.

To restate the credit: you had the right fix here a year ago — silent mode 2031 subscriptions. #13904 landed the same conclusion, reached independently and then confirmed against your commit. What it adds is the justification this needed to be merge-able: the Contour spec's "only send the DSR when the palette has been updated", Ghostty/iTerm2/xterm.js all emitting nothing on the DECSET, and a real-fish node-pty regression test that fails against pre-fix code.

Issue #9282, opened 2026-07-18
[Bug]: macOS Homebrew install — orca serve SIGABRT, runtime stuck in stale_bootstrap (not a signing/ASAR issue)


arata-ai-daisuki:
### Operating system

macOS 26.5.2 (25F84), arm64

### Orca version

1.4.145, installed via the Homebrew cask

### Summary

Orca does not launch, and the CLI runtime remains unavailable.

The installed application fails macOS code-signature verification. Its `app.asar` SHA-256 also differs from the hash recorded in `ElectronAsarIntegrity` inside `Info.plist`.

### Steps to reproduce

1. Install Orca 1.4.145 using the Homebrew cask.
2. Run:

   ```shell
   open -a "/Applications/Orca.app"
   ```
3. Check runtime status:

   ```shell
   orca status --json
   ```
4. Verify the application signature:

   ```shell
   codesign --verify --deep --strict --verbose=2 "/Applications/Orca.app"
   ```
5. Compare the ASAR hash:

   ```shell
   shasum -a 256 "/Applications/Orca.app/Contents/Resources/app.asar"
   ```

### Actual behavior

`orca status --json` reports:

* `app.running: false`
* `runtime.state: stale_bootstrap`
* `runtime.reachable: false`
* `graph.state: not_running`

`orca serve` exits with:

```text
Orca serve exited via SIGABRT
```

Code-signature verification reports:

```text
/Applications/Orca.app: invalid signature (code or signature have been modified)
In architecture: arm64
```

Expected `ElectronAsarIntegrity` hash from `Info.plist`:

```text
7aaa11254dab1cfce70d58d8b0f90a771f6fe7f445cb59642359258ae9190058
```

Actual `app.asar` SHA-256:

```text
c2ed60ee859668a7bebb4714349a5441c52669cb8d63023336803c9715337aa8
```

### Expected behavior

The Homebrew-installed application should pass code-signature and ASAR integrity verification and start a reachable Orca runtime.

### Additional context

The app contains a stapled notarization ticket and TeamIdentifier `6CX3WHS9HZ`, but the installed bundle still fails signature verification.

Possibly related, but behavior differs:

* stablyai/orca#7848
* stablyai/orca#6635

zigggy-stardust:
Follow-up verification: the signing and ASAR-integrity conclusions in this issue appear to be false positives, although the reported runtime failure may still be real.

[...]

PR #10059, opened 2026-07-23
fix(updater): disable differential download on macOS to preserve asar integrity

brennanb2025:
Fixes STA-2081.

Root cause: the Homebrew cask's auto_updates=true routes upgrades through Orca's in-app macOS updater, and electron-updater 6.8.9 MacUpdater does differential (blockmap) downloads by default. It reconstructs the app zip and hands it to Squirrel.Mac, producing an app.asar whose bytes no longer match the notarized Info.plist ElectronAsarIntegrity, failing codesign + Electron's ASAR check and SIGABRTing the bundled CLI/daemon. A fresh DMG direct-download stays pristine, which is why only Homebrew breaks. For an affected user the app will not launch at all, so this is a hard brick for that cohort, not a degraded state.

Fix: set autoUpdater.disableDifferentialDownload=true gated to darwin so every mac update pulls the full notarized zip; Windows/Linux unaffected. Adds 2 regression tests.

## Download-size impact (measured)

This disables differential downloads for **all** macOS users, not only Homebrew, so the fair question is whether we're adding install time for everyone to fix a subset. Measured against our own arm64 release blockmaps, differential was already saving very little on the update spans real stable users actually traverse:

| Update span | Differential fetch | vs. full ~193 MB |
| --- | --- | --- |
| Consecutive builds, hours apart (1.4.153-rc.2 → rc.3) | ~47 MB | 23.6% |
| One week apart (1.4.152 → 1.4.153-rc.3) | ~197 MB | 98.6% |

The macOS artifact is a compressed zip, so block reuse collapses as soon as a change lands early in the archive — or a native module rebuild / Electron bump shifts the deflate stream — and nearly every downstream block differs even where the underlying files are identical. Small, close diffs reuse well; anything substantial reuses almost nothing.

Stable and Homebrew users update stable→stable (days to weeks apart), which is the high-percentage end of that table, so the **incremental** cost of forcing full downloads for them is close to zero. Worst case — a user who would otherwise have caught a trivial diff — is a one-time ~+150 MB. Update bytes are served from GitHub release assets, so there is no egress cost to us; the only cost is user bandwidth and download time.

Why blanket-darwin instead of Homebrew-only: differential's savings for direct-DMG users are marginal for the same compressed-archive reason, and detecting install source at runtime would add fragile complexity for little benefit. Forcing the full zip is also strictly more reliable here, since differential reconstruction is the exact step that corrupts the asar — we're removing a fragile stage, not just trading speed for safety.

Test plan: vitest (83, incl. 12 mac-install) + typecheck + oxlint green. Full verification needs a real signed vN->vN+1 macOS update (confirm codesign --deep --strict and app.asar<->ElectronAsarIntegrity both pass post-update).

### DIFF
--- a/src/main/updater.test.ts
+++ b/src/main/updater.test.ts
@@ -51,6 +51,7 @@ const {
     autoUpdaterMock.updateConfigPath = undefined
     autoUpdaterMock.allowPrerelease = false
     autoUpdaterMock.autoRunAppAfterInstall = true
+    autoUpdaterMock.disableDifferentialDownload = false
     delete (autoUpdaterMock as Record<string, unknown>).verifyUpdateCodeSignature
   }
 
@@ -58,6 +59,7 @@ const {
     autoDownload: false,
     autoInstallOnAppQuit: false,
     autoRunAppAfterInstall: true,
+    disableDifferentialDownload: false,
     allowPrerelease: false,
     on,
     checkForUpdates: vi.fn(),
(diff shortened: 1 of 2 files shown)


brennanb2025:
Closing this — the root cause it describes does not hold up, and I'd rather not ship a change that costs every macOS user download bandwidth for no defect fixed.

## Why

**1. `ElectronAsarIntegrity` is the SHA-256 of the ASAR *header*, not of the whole `app.asar` file.**

Verified against the authoritative `@electron/asar` reader on an installed Orca 1.4.153-rc.4:

```
Info.plist ElectronAsarIntegrity : aa48743c97ee976df3f4d8f7ac249ac9961ccfdb3f7aa509d689e2a4d0242755
ASAR *header*     SHA-256        : aa48743c…0242755   <-- MATCH
ASAR *whole-file* SHA-256        : 3aa53d8a…f262043   <-- differs, as expected
```

The original report compared a whole-file hash against the header hash, which mismatches on a perfectly healthy app. That evidence never indicated corruption.

**2. The `codesign` failure in the report was environmental.** The same unmodified bundle passes `codesign --verify --deep --strict`, `spctl` (`accepted`, `source=Notarized Developer ID`), and `stapler validate`, resolving to `Developer ID Application: Lovecast LLC (6CX3WHS9HZ)`. The failure reproduces only under a restricted sandbox whose certificate trust evaluation returns `Authority=(unavailable)`.

**3. electron-updater 6.8.9 cannot hand a corrupted differential bundle to Squirrel.Mac.** The reconstruction is piped through `DigestTransform(blockAwareFileInfo.sha512)` and validated on finish; that sha512 is the hash of the **full published zip** from `latest-mac.yml`. On any mismatch the error is caught, logged as `Cannot download differentially, fallback to full download`, and the full artifact is refetched. So differential either reproduces the published zip byte-for-byte or it falls back.

**4. Confirmed empirically on a real update.** An in-app macOS update completed on a real Mac (`ShipIt status 0`) with the differential machinery live (`current.blockmap` + a cached prior `update.zip`). The installed artifact's sha512 was byte-identical to the published manifest:

```
local     Orca-1.4.153-rc.4-arm64-mac.zip  sha512 = dbE/d1PtUQANYTBy7HU/…
published latest-mac.yml (v1.4.153-rc.4)   sha512 = dbE/d1PtUQANYTBy7HU/…
```

and the resulting app passed every signature and ASAR integrity check.

## Cost side

Measured against our own arm64 blockmaps, forcing full downloads would add up to ~150 MB for close-together updates (rc.2 → rc.3 differential fetches ~47 MB of ~193 MB), and ~0 for week-apart updates where block reuse has already collapsed (1.4.152 → 1.4.153-rc.3 differential fetches ~98.6% of full). Small either way in isolation — but not worth paying for a path that is already guarded.

## What's still real

The reporter's runtime symptoms — `orca serve` exiting via SIGABRT, `runtime.state: stale_bootstrap`, `app.running: false` — are unexplained and unaffected by this change. That's a bootstrap/runtime failure, not a signing one. Tracking it on STA-2081 against that path instead.

Thanks to @zigggy-stardust for the careful re-verification on #9282 that prompted this.

PR #10060, opened 2026-07-23
fix(status-bar): escalate stuck Claude "Refreshing sign-in" after timeout

### LINKED ISSUE #8974: [Bug]: (Does not affect normal use) Claude Code usage status stuck on "Refreshing sign-in" and does not update
## Summary

Claude Code (CC) itself still works in the terminal, but Orca’s **Claude usage / status-bar tracking sometimes fails to refresh**. The status UI gets stuck on **“Refreshing sign-in”** and never recovers to live usage numbers.

Reporter confirmation: CC is usable in-terminal; the problem is specifically that **usage status cannot refresh**.

## Environment

- **Orca version:** 1.4.137
- **OS:** macOS 15.7.7
- **Hardware:** Apple M2 Pro
- **Claude Code version:** 2.1.208

## What happened

1. Claude usage / rate-limit status does not update in Orca.
2. Opening the Claude account / status UI shows a persistent **Refreshing sign-in** state:
   - Status label: `Refreshing sign-in`
   - Message: `Claude sign-in is being refreshed. Agent sessions may still be signed in.`
   - Account selector shows `Claude Account` → `System default`
3. Claude Code sessions in the terminal continue to work (agent is signed in and usable).
4. Status remains stuck in this state rather than completing the refresh and showing usage percentages.

## Expected behavior

- Claude usage status refreshes successfully (auto or manual).
- Status bar / account UI shows current usage after refresh.
- Transient credential refresh should not leave the UI stuck indefinitely while agent sessions remain signed in and working.

## Actual behavior

- Usage tracking does not update.
- UI remains stuck on **Refreshing sign-in**.
- Terminal Claude Code continues to work, so this looks like a **usage-fetch / sign-in-refresh orchestration** problem rather than a full logout.

## Screenshots

(Attached in follow-up comments if not inline.)

The stuck state shows:
- Claude popover: **Refreshing sign-in**
- Footer/status chip: **⚠ Refreshing sign-in**
- Message about agent sessions still being signed in

## Notes / likely code area

Status copy for this state is mapped in `src/renderer/src/components/status-bar/usage-error-copy.ts` for Claude `usageMetadata.failureKind` values:

- `stale-token`
- `refreshable-credentials-without-token`
- `delegated-refresh-required`

Those map to the exact user-visible strings:

- label: **Refreshing sign-in**
- detail: **Claude sign-in is being refreshed. Agent sessions may still be signed in.**

Related refresh/auth paths:

- `src/main/rate-limits/claude-fetcher.ts`
- `src/main/rate-limits/claude-usage-error-classification.ts`
- `src/main/claude-accounts/oauth-refresh.ts`
- `src/main/claude-accounts/live-pty-gate.ts`

Possible failure mode: delegated/stale-token refresh never completes (or never clears the failure kind), so usage polling keeps presenting “Refreshing sign-in” even though live Claude terminals remain authenticated.

## Related (not duplicates)

Prior usage-refresh bugs that were closed after older versions:

- #5355 — “Refresh failed” after update (auth path)
- #5627 — usage pills stop auto-refreshing after relaunch (focus-gated poll)
- #5809 — Claude usage broken on update

This report is on **1.4.137** with a different stuck UI string (**Refreshing sign-in**) and confirmed working terminal CC.

## Ask

Please investigate why Claude usage refresh can remain stuck in the “Refreshing sign-in” / recoverable-auth path while terminal sessions keep working, and restore reliable usage status updates.

innocarpe:
## Description

**User-facing:** Claude usage in the status bar could sit forever on **“Refreshing sign-in”** even when the terminal Claude session was fine and usage never came back. After this change, the soft “refreshing” state is kept only while repair is in flight; if it stays stuck for **2 minutes**, the chip escalates to **“Sign-in needs refresh”** and the roster offers a re-auth CTA.

**Technical:** stamp a stable `usageMetadata.refreshingSinceMs` across continuous Claude `stale-token` / `refreshable-credentials-without-token` / `delegated-refresh-required` failures in `applyStalePolicy`, then escalate label/message/roster kind once that stamp ages past `CLAUDE_REFRESHING_SIGN_IN_ESCALATE_AFTER_MS`.

Fixes #8974

## Impact / ELI5

Orca tries to refresh Claude’s usage OAuth in the background. Sometimes that refresh never succeeds, but the soft wording still said “we’re refreshing…” forever — like a loading spinner that never ends. Now we give the soft spinner a short time window; if it doesn’t finish, we tell you clearly that you need to re-authenticate for usage tracking (your agent in the terminal may still work).

## Root cause

`getProviderUsageStatusLabel` / `getProviderUsageErrorMessage` map these Claude `failureKind`s to permanent soft copy:

- `stale-token`
- `refreshable-credentials-without-token`
- `delegated-refresh-required`

There was **no age/timeout escalation**. The main fetcher can keep re-emitting those kinds on every poll (`updatedAt` moves, but the semantic state stays “refreshing”), so the UI never leaves the spinner. Live Claude CLI sessions can remain healthy on a different credential surface while usage tracking is stuck.

Reproduction kit (handoff): `docs/bug-reproductions/8974.md` on `nwparker/bug-repro-handoffs`.

## Fix

| Layer | Change |
|-------|--------|
| Shared | `src/shared/claude-refreshing-sign-in.ts` — stamp/preserve `refreshingSinceMs`, escalate predicate (2 min) |
| Types | `UsageRateLimitMetadata.refreshingSinceMs?: number` |
| Main | `applyStalePolicy` always runs `stampClaudeRefreshingSince` so continuous error polls keep the **first** stamp |
| UI | Status label + tooltip message escalate after timeout; roster row becomes `kind: 'sign-in'` with re-auth CTA |

### Behavior matrix

| State | Label | Roster |
|-------|-------|--------|
| Refreshing &lt; 2 min | Refreshing sign-in | `error` (soft) |
| Refreshing ≥ 2 min | Sign-in needs refresh | `sign-in` (CTA) |
| Recovered (`status: 'ok'`) | normal usage | stamp cleared |
| Live-session deferral | Waiting for Claude session | unchanged (not escalated) |

## Evidence

```bash
  src/shared/claude-refreshing-sign-in.test.ts \
  src/renderer/src/components/status-bar/repro-8974-refreshing-sign-in-stuck.test.ts \
  src/renderer/src/components/status-bar/usage-roster-row-state.test.ts \
  src/renderer/src/components/status-bar/tooltip.test.ts
```

**Result:** 57/57 passed (local).

Key tests:

- stamp on first refreshing failure; preserve across continuous polls; clear on recovery
- soft copy while window open; escalated copy + roster CTA after timeout
- existing tooltip/roster regressions still green

## User-regression-tradeoffs

- **Intentional UX change:** after ~2 minutes of continuous Claude “refreshing” failures, copy becomes a re-auth prompt instead of an infinite soft spinner. This matches the has_repro handoff suggestion and is better than implying progress that never finishes.
- Does **not** change OAuth/credential repair itself — if refresh would have succeeded at 30s, soft copy still applies.
- Does **not** touch `deferred-by-live-session` (“Waiting for Claude session”).
- Grok/Kimi delegated-CLI refresh paths unchanged.

## Checklist notes

- **Cross-platform:** pure TS status logic; no OS APIs.
- **SSH/remote:** rate-limit state path is shared; stamp lives on usage metadata pushed to the renderer.
- **Agents/integrations:** Claude usage UI only; terminal agents unaffected.
- **Performance:** O(1) metadata stamp on each stale-policy apply.
- **UI:** status-bar label + roster CTA only; no STYLEGUIDE token changes.
- **Security:** no secrets; only UI/state timing.
- **i18n:** new strings use `translate()` with English fallbacks (same pattern as nearby status-bar keys).

## AI review summary

- Narrow, handoff-aligned fix for indefinite “Refreshing sign-in”.
- Shared pure helper keeps main + renderer in sync and is unit-tested without Electron.
- Escalate window is a constant (2 min) so it can be tuned later without hunting call sites.

## ELI5

Claude usage in the status bar could sit on "Refreshing sign-in" forever. After two minutes of no progress it escalates to "Sign-in needs refresh" and offers a re-auth action instead of spinning forever.

### DIFF
--- a/src/main/rate-limits/service.ts
+++ b/src/main/rate-limits/service.ts
@@ -7,6 +7,7 @@ import type {
   InactiveAccountUsage,
   RateLimitRuntimeTarget
 } from '../../shared/rate-limit-types'
+import { stampClaudeRefreshingSince } from '../../shared/claude-refreshing-sign-in'
 import { fetchClaudeRateLimits, fetchManagedAccountUsage } from './claude-fetcher'
 import type { InactiveClaudeAccountInfo } from './claude-fetcher'
 import { mapClaudeUsageWindow } from './claude-usage-window'
@@ -1969,19 +1970,22 @@ export class RateLimitService {
   ): ProviderRateLimits {
     // Fresh data is fine — use it
     if (fresh.status === 'ok') {
-      return {
-        ...fresh,
-        usageMetadata: {
-          ...fresh.usageMetadata,
-          lastSuccessfulSource:
-            fresh.usageMetadata?.source ?? fresh.usageMetadata?.lastSuccessfulSource
-        }
-      }
+      return stampClaudeRefreshingSince(
+        {
+          ...fresh,
+          usageMetadata: {
+            ...fresh.usageMetadata,
+            lastSuccessfulSource:
+              fresh.usageMetadata?.source ?? fresh.usageMetadata?.lastSuccessfulSource
+          }
+        },
+        previous
+      )
     }
 
     // Explicitly unavailable (e.g. setting cleared): discard stale data so the UI shows the provider as disabled/unconfigured.
     if (fresh.status === 'unavailable') {
-      return fresh
+      return stampClaudeRefreshingSince(fresh, previous)
     }
 
     const previousHasData = Boolean(
@@ -1994,7 +1998,7 @@ export class RateLimitService {
 
     // No previous data to fall back on
     if (!previous || !previousHasData) {
-      return fresh
+      return stampClaudeRefreshingSince(fresh, previous)
     }
 
     // Previous data is too old — don't show stale data
@@ -2003,21 +2007,24 @@ export class RateLimitService {
         ? RATE_LIMITED_STALE_THRESHOLD_MS
         : STALE_THRESHOLD_MS
     if (Date.now() - previous.updatedAt > staleThresholdMs) {
-      return fresh
+      return stampClaudeRefreshingSince(fresh, previous)
     }
 
     // Why: keep showing a recent snapshot through repeated transient failures until it ages out, so the bar doesn't flap to empty.
-    return {
-      ...previous,
-      error: fresh.error,
-      status: 'error',
-      usageMetadata: {
-        ...previous.usageMetadata,
-        ...fresh.usageMetadata,
-        lastSuccessfulSource:
-          previous.usageMetadata?.lastSuccessfulSource ?? previous.usageMetadata?.source
-      }
-    }
+    return stampClaudeRefreshingSince(
+      {
+        ...previous,
+        error: fresh.error,
+        status: 'error',
+        usageMetadata: {
+          ...previous.usageMetadata,
+          ...fresh.usageMetadata,
+          lastSuccessfulSource:
+            previous.usageMetadata?.lastSuccessfulSource ?? previous.usageMetadata?.source
+        }
+      },
+      previous
+    )
   }
 
   private buildInactiveArray(
--- a/src/renderer/src/components/status-bar/repro-8974-refreshing-sign-in-stuck.test.ts
+++ b/src/renderer/src/components/status-bar/repro-8974-refreshing-sign-in-stuck.test.ts
@@ -1,0 +1,93 @@
+/**
+ * Issue #8974 — Claude usage stuck on "Refreshing sign-in".
+ *
+ * Transient failureKinds map to soft "Refreshing sign-in" copy while repair is
+ * in flight. Once `refreshingSinceMs` ages past the escalate window, the UI
+ * must leave the indefinite spinner and surface re-auth copy / CTA.
+ *
+ * Re-run:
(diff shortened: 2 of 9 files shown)


CannyDuck:
Any plans to merge this PR? Have the same problem on Windows:

{A804CA59-64C7-46BA-AB6E-350072DB358E}

innocarpe:
### Sync update (`41d8a952f`)

Rebased onto the latest upstream/main, ported the refresh timestamp handling through the current split rate-limit service, and addressed the CodeRabbit maintainability nit by deriving the failure-kind check from the shared constant. Focused verification passes: 57 tests, tc:node, oxfmt, and targeted oxlint. This remains limited to the existing Claude usage error path and is ready for review/merge.

- Commit: `41d8a952f5a65850e96fc867ad9811f35e13726d`
- Upstream: https://github.com/stablyai/orca/pull/10060
- Fork portfolio: https://github.com/innocarpe/orca/pull/12

PR #2791, opened 2026-05-25
feat(theme): macOS native glass effect as orthogonal setting

edwintuan:
## Summary

Adds a native macOS glass/vibrancy visual layer to Orca, exposed as an
independent `glassEffect` boolean that composes orthogonally with the
existing `theme` setting. System + Glass tracks OS dark/light while
keeping the frosted surfaces.

## Settings model

- `theme: 'system' | 'dark' | 'light'` — unchanged (3 values).
- `glassEffect: boolean` — new, defaults `false`, **effective on macOS only**.
- The two are orthogonal: any combination is valid. System + Glass
  follows OS dark/light AND keeps glass.

## UX surface

- **Settings → Appearance**: 3-value Theme segmented control + a separate
  "Glass effect" switch underneath. The switch is hidden on Linux/Windows.
  Toggling glass requires a window recreate → restart banner appears.
- **Onboarding**: 3 theme tiles on every platform (System / Dark / Light).
  Glass is opt-in later via Settings, not surfaced during first-run, so
  new users don't have to make a 5-way decision about a visual layer
  they haven't seen yet.
- **Window Blur toggle** is disabled while Glass is on (avoids two blur
  layers fighting each other).

## Cross-platform behavior

Glass is **macOS-only** by design — it leans on `NSVisualEffectView`
vibrancy, which has no equivalent on Linux/Windows.

- **macOS**: `BrowserWindow` is created with `vibrancy: 'under-window'`,
  `transparent: true`, and `backgroundColor: '#00000000'` so the system
  blur shows through. CSS layers `.glass-light` / `.glass-dark` are
  applied on top.
- **Linux / Windows**: `isGlassEffectActive(settings)` short-circuits to
  `false` regardless of the stored `glassEffect` value. The Settings
  switch is hidden. No translucent window, no CSS glass classes, no
  Monaco glass theme. Behavior is identical to `main`.
- The `glassEffect: true` value still **persists** on non-macOS hosts
  (e.g. user toggles it on macOS, then opens the same profile on Linux);
  it just has no visual effect there. This keeps profiles portable
  without surprising state loss.
- `prefers-reduced-transparency` is respected via existing media query
  hooks — frosted surfaces fall back to opaque.

## Migration

Older builds shipped a 5-value theme picker (`glass-light` / `glass-dark`
as discrete theme values). `persistence.ts` main-process load detects the
legacy values and rewrites them on disk:

| Legacy `theme`  | New `theme` | New `glassEffect` |
[...]

PR #10064, opened 2026-07-23
fix(window): do not cover platform blur with opaque backgroundColor

### LINKED ISSUE #8797: [Bug]:  "Background Opacity" and "Window Blur" have no visible effect on macOS
### Operating system

macOS

### Orca version

 1.4.141

### Short summary

On macOS, the **Window Blur** and **Background Opacity** settings (Settings → Terminal → Window) have no visible effect. Enabling blur and lowering the opacity does nothing, the window stays fully opaque and the blurred desktop never shows through.

### What happened?

The macOS window is created with `transparent: true` **together with** `vibrancy`. That combination suppresses the vibrancy view, and an opaque `backgroundColor` is painted over the blur layer. As a result, neither the blur nor the background opacity ever renders, both settings appear to do nothing on macOS.

### How can we reproduce it?

1. Run Orca on macOS.
2. Open **Settings → Terminal → Window**.
3. Enable **Window Blur** (restart if prompted).
4. Lower **Background Opacity** below `1`.
5. **Observe:** the window remains fully opaque - the blurred desktop never shows through, so both settings have no visible effect.

### Anything else that might help?

**Root cause:** on macOS the window sets `transparent: true` alongside `vibrancy`, which suppresses the vibrancy view, and an opaque `backgroundColor` paints over the blur layer.

**Fix:** omit `transparent: true` and the opaque `backgroundColor` on the blur surface, and instead use `vibrancy: 'under-window'` + `visualEffectState: 'active'`.

Fix (open): [feat(theme): app-wide window glass with vibrancy/acrylic #3972](https://github.com/stablyai/orca/pull/3972)
Earlier related PR: https://github.com/stablyai/orca/pull/2791

innocarpe:
## Description

**User-facing:** On macOS (and Windows acrylic), **Window Blur** + **Background Opacity** could look completely broken — the window stayed fully opaque even after enabling blur and lowering opacity.

**Technical:** `createMainWindow` always set an opaque `backgroundColor` (`#0a0a0a` / `#ffffff`) even when blur enabled platform materials. That solid fill covered macOS vibrancy / Windows acrylic, so lowering `terminalBackgroundOpacity` only revealed the same solid chrome color — never the desktop.

Fixes #8797

## Impact / ELI5

Orca was putting a solid painted board behind the glass. Turning on “blur” and lowering opacity only made the terminal more see-through to that solid board. This removes the solid board when blur is on, so you can actually see the frosted desktop through a translucent terminal.

## Root cause

From the has_repro handoff (`docs/bug-reproductions/8797.md`):

1. Blur path set `vibrancy: 'under-window'` + `transparent: true` (Darwin) or `backgroundMaterial: 'acrylic'` (Windows).
2. Constructor **always** applied opaque `backgroundColor: dark ? '#0a0a0a' : '#ffffff'`.
3. Terminal opacity only rewrites the xterm theme to `rgba(...)` — it never changed the BrowserWindow fill.
4. Result: transparent terminal sits on solid black/white; blur never shows.

## Fix

New pure helper `resolveMainWindowChromeOptions`:

| Platform | Blur off | Blur on |
|----------|----------|---------|
| macOS | opaque fill | `#00000000` + `vibrancy: under-window` + `visualEffectState: active` + `transparent: true` |
| Windows | opaque fill | `#00000000` + `backgroundMaterial: acrylic` |
| Linux | opaque fill | opaque fill (no material) |

`createMainWindow` consumes the helper instead of hard-coding the opaque fill.

## Evidence

```bash
  src/main/window/main-window-chrome-options.test.ts \
  src/main/window/repro-8797-blur-opacity-noop.test.ts
```

**Result:** 8/8 passed.

Manual (macOS):

1. Settings → Terminal → Window → enable **Window Blur** → restart.
2. Lower **Background Opacity** below `1`.
3. Terminal should reveal blurred desktop (chrome UI still solid via CSS).

## User-regression-tradeoffs

- **Intentional:** with blur ON, the native window fill is fully transparent. App chrome must continue to paint its own solid backgrounds (existing design). Areas that forget a solid surface may briefly show desktop — same class of issue as any transparent Electron window.
- Blur remains **startup-only** (restart required) — unchanged.
- Does **not** implement the broader glass-theme redesign in #3972 / #2791; this is a narrow correctness fix for the existing settings.
- Linux unchanged (no supported blur material).

## Checklist notes

- **Cross-platform:** Darwin + Windows materials; Linux no-op path preserved.
- **SSH/remote:** n/a (local BrowserWindow chrome only).
- **Agents/integrations:** n/a.
- **Performance:** same window-creation path; no runtime cost.
- **UI:** relies on existing renderer surfaces for chrome; terminal opacity path unchanged.
- **Security:** n/a.

## AI review summary

- Pure options helper is unit-tested without Electron.
- Repro contract from #8797 handoff updated to assert the fixed behavior.
- Scoped away from large theme PRs so maintainers can land a minimal fix independently.


innocarpe:
Accidentally closed when a bad rebase helper force-pushed an empty tip. Restored and continued as https://github.com/stablyai/orca/pull/10349.

PR #10073, opened 2026-07-23
fix(usage): guard web client against undefined usage scan state

scokeepa:
## Summary

Enabling an agent's usage tracking in **Settings → Stats & Usage** no longer crashes when Orca runs as a web client (a paired `orca serve` runtime).

## ELI5

In the browser version of Orca, clicking "enable" for usage stats used to hit an error because that feature only exists on the desktop app. Now it just quietly does nothing instead of breaking.

## Root cause & fix

The `*Usage` IPC namespaces (`claudeUsage`, `codexUsage`, `openCodeUsage`) are desktop-only and are never bridged in the web preload, so the fallback proxy resolves every call to `undefined`. The slices cast that result to a non-optional scan state and read `scanState.enabled`, throwing `Cannot read properties of undefined (reading 'enabled')`. The fix adds `| undefined` to the casts and bails out with a no-op when the scan state is absent — on both the `enable*` (setEnabled) and `fetch*` (getScanState) paths — across all three providers. Desktop behavior is unchanged since a real `ScanState` is always truthy there.

## Evidence

- Test file: `src/renderer/src/store/slices/usage-web-client-fallback.test.ts` (3 tests, one per provider), which stubs the web-client fallback so every `window.api.*Usage.*` resolves to `undefined`.
- On branch (fix present): **3 passed**. Fix reverted to `origin/main`: **3 failed** with the assertion below — proving the test guards the fix.
- Lint: `oxlint` on the three changed source files is clean.

```
On branch:        Test Files 1 passed (1) | Tests 3 passed (3)
Fix reverted:     Test Files 1 failed (1) | Tests 3 failed (3)
  AssertionError: expected undefined to be null
  (claudeUsageScanState / codexUsageScanState / openCodeUsageScanState received undefined)
```

## Trade-offs

None — pure fix. The Settings navigation gating is intentionally unchanged; the web client now treats usage tracking as unavailable (its actual state) rather than crashing.

## Regression risk

Zero-regression: only the previously-crashing branch changes — when the scan state is present (always the case on desktop) every path is byte-identical, so desktop behavior is untouched. The passing test covers all three providers on both the enable and fetch seams.

*Credit to original author @scokeepa. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/renderer/src/store/slices/claude-usage.ts
+++ b/src/renderer/src/store/slices/claude-usage.ts
@@ -45,7 +45,14 @@ export const createClaudeUsageSlice: StateCreator<AppState, [], [], ClaudeUsageS
     try {
       const nextScanState = (await window.api.claudeUsage.setEnabled({
         enabled
-      })) as ClaudeUsageScanState
+      })) as ClaudeUsageScanState | undefined
+      // Why: the web client (paired runtime) does not bridge the desktop-only
+      // usage IPC; its preload fallback resolves this call to `undefined`. Bail
+      // so the toggle no-ops instead of seeding an empty scan state and then
+      // crashing on the follow-up fetch.
+      if (!nextScanState) {
+        return
+      }
       set({
         // Why: every enable should look like a fresh scan cycle in the UI.
         // Reusing the last completed timestamp makes repeated toggles skip the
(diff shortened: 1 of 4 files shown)


nwparker:
Thanks @scokeepa — great catch on the web-client usage-scan crash. Verified the fix (fail-on-main/pass-on-branch: 3 tests fail on `main`, pass with the guards) and merged. 🙏 (Follow-up idea, non-blocking: the root cause is that the three `*Usage` namespaces aren't registered in `createWebPreloadApi()`, so they hit the generic fallback proxy — a small web-side registration would fix it at the source; happy to do that separately.)

Issue #10088, opened 2026-07-23
[Bug]: Multi-host project header overlaps host section after selecting a project


lyonDan:
## Short summary

When more than one host is selected in the Projects sidebar, clicking a project/workspace under a host can leave the project-group header rendered on top of the host header. The project no longer looks visually contained by its host section.

## What happened?

With **Local Mac** and an SSH host selected together, expanding the SSH host and opening one of its projects/workspaces triggers the sidebar reveal/scroll behavior. The project-group label is then painted through the SSH host card, while the same group remains visible below it in the list.

This is not a duplicate project in persisted data; it is a virtualized sticky-header rendering problem at the Host -> Project boundary.

## How can we reproduce it?

1. Run Orca 1.4.152 on macOS.
2. Configure at least two execution hosts (for example, Local Mac and one SSH host), each with projects/workspaces.
3. In the Projects sidebar, use the host filter and select both hosts.
4. Keep the sidebar grouped by Project.
5. Expand the SSH host and click a project/workspace under it.
6. Observe that the project-group header overlaps the host card instead of remaining visually nested below it.

The issue is easiest to see when one host section is collapsed and the other is expanded.

## Expected behavior

Each project/group header should remain visually scoped to its owning host. Selecting or revealing a project must not make the project label overlap or bleed through a Host header.

## Actual behavior

The project-group label is painted across the Host card after selection/reveal, making the Host and Project hierarchy appear flattened or duplicated.

## Local source analysis

The sidebar renders Host and Project rows as two independent sticky tiers in the same TanStack virtual list:

- [`getActiveStickyIndexesForScroll`](https://github.com/stablyai/orca/blob/ee6319ebe4c8e031088d36466dc1529cdecc7d70/src/renderer/src/components/sidebar/worktree-list-virtual-rows.ts#L138-L197) resolves the active Host and Project sticky indexes.
- Project/workspace selection can trigger [`virtualizer.scrollToIndex(...)`](https://github.com/stablyai/orca/blob/ee6319ebe4c8e031088d36466dc1529cdecc7d70/src/renderer/src/components/sidebar/WorktreeList.tsx#L2142-L2214).
- The active sticky state is recomputed from a mutable range-start ref plus the current virtual items at [`WorktreeList.tsx#L2384-L2394`](https://github.com/stablyai/orca/blob/ee6319ebe4c8e031088d36466dc1529cdecc7d70/src/renderer/src/components/sidebar/WorktreeList.tsx#L2384-L2394).
[...]

PR #10090, opened 2026-07-23
fix(sidebar): do not sticky-pin Project header without mounted geometry

innocarpe:
# fix(sidebar): do not sticky-pin Project header without mounted geometry

## Summary

Fixes the multi-host Projects sidebar painting a project-group sticky header over the host card when revealing or scrolling to a workspace (#10088).

## ELI5

In the sidebar, jumping to a project under a host could make the project's name label overlap and cover the host's card. This makes the sticky labels stay in the right place while you scroll.

## Root cause & fix

`getActiveStickyIndexesForScroll` → `resolveWithHandoff` returned the candidate sticky header index even when that virtual row wasn't mounted yet — which happens when `scrollToIndex`/reveal advances `rangeStartIndex` before TanStack mounts the row. A Project sticky then activated without any geometry while the Host sticky was still active. The fix now prefers a previously-mounted sticky header when the candidate is unmounted, and only falls back to the raw candidate when `fallbackToCandidate` is true (host tier). The group tier (`fallbackToCandidate=false`) returns `null` instead of pinning without geometry.

## Evidence

- Test file: `src/renderer/src/components/sidebar/worktree-list-virtual-rows.test.ts`
- On branch: **9/9 passed** (1 file). With the fix reverted (test kept): **2 failed | 7 passed**, proving the tests catch the regression.
- Failing assertions when fix reverted:

```
✗ does not pin a Project header whose virtual item is not mounted yet (#10088)
    AssertionError: expected 5 to be null (groupIndex was 5)
✗ keeps the previous mounted Project sticky when the next group is unmounted
    AssertionError: expected 3 to be 1 (groupIndex was 3, should hold previous)
```

- Lint: `oxlint` clean (no warnings/errors) on the fix file.
- Rebase: clean onto `origin/main`.

## Trade-offs

None — pure fix. Host tier retains its legacy fallback so a re-mounting host section keeps sticky identity.

## Regression risk

Effectively zero. Only the branch that handled an unmounted candidate changes; every path with a mounted candidate or a truthy `fallbackToCandidate` behaves identically. The host tier is byte-for-byte preserved, and the passing test proves both the new group-tier and previous-sticky-hold behaviors.

---

*Original fix by @innocarpe. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

### DIFF
--- a/src/renderer/src/components/sidebar/worktree-list-virtual-rows.test.ts
+++ b/src/renderer/src/components/sidebar/worktree-list-virtual-rows.test.ts
@@ -127,6 +127,45 @@ describe('getActiveStickyIndexesForScroll', () => {
       })
     ).toEqual({ hostIndex: null, groupIndex: 0 })
   })
+
+  it('does not pin a Project header whose virtual item is not mounted yet (#10088)', () => {
+    // Why: after scrollToIndex/reveal, rangeStart can sit on group-b1 while
+    // TanStack has only mounted host-b (and maybe a later item) this frame.
+    const partialItems = [virtualItem(4, 400), virtualItem(6, 600)]
+    const result = getActiveStickyIndexesForScroll({
+      rows,
+      rangeStartIndex: 5,
+      scrollOffset: 500,
+      stickyHeaderIndexes,
+      virtualItems: partialItems
+    })
+    expect(result.hostIndex).toBe(4)
+    // group-b1 (index 5) must not become sticky without geometry — that is what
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks so much for this, @innocarpe — do not sticky-pin Project header without mounted geometry. Verified (fail-on-main/pass-on-branch), reviewed, and merged. Really appreciate it! 🙏

AmethystLiang:

PR #10091, opened 2026-07-23
fix(preflight): refresh Windows PATH on forced CLI checks

### LINKED ISSUE #9316: [Bug]: The GitHub CLI could not be detected immediately.
### Operating system

Windows

### Orca version

v1.4.145

### Details

During initialization, it prompted me to install the GitHub CLI. I remembered I already had an older version on my computer, but I installed it anyway. Sure enough, PowerShell also detected the older version and prompted me to upgrade. After upgrading, clicking "re-detect" the first, second, and third times had no effect. After waiting five minutes, it detected the older version successfully.

xianjianlf2:
## Summary
- add a forceRefresh path for persisted Windows Path reads so installer updates can bypass the short registry cache
- refresh and merge the persisted Windows Path into the main process before forced host preflight checks
- keep WSL preflight isolated from host Path refreshes

Fixes #9316

## Testing
- corepack pnpm exec vitest run --config config/vitest.config.ts src/main/pty/windows-environment-path.test.ts src/main/ipc/preflight.test.ts
- corepack pnpm exec oxlint src/main/pty/windows-environment-path.ts src/main/pty/windows-environment-path.test.ts src/main/ipc/preflight.ts src/main/ipc/preflight.test.ts
- corepack pnpm run typecheck:web

## Screenshots
No visual change.

## AI Review Report
CodeRabbit reported no actionable code comments on the latest review. OrcaWin also validated the behavior on a real Windows 11 machine, including the forced host preflight refresh, WSL isolation, registry parsing edge cases, and cache fallback behavior.

## Security Audit
No new data exposure or privilege boundary change. The PR reads the existing Windows user and machine Path values through the same local registry mechanism already used by preflight detection, with forced refresh limited to explicit host preflight checks.

## Notes
The forced refresh uses synchronous registry reads in the main process, matching the existing cold-cache behavior. Follow-up async optimization can be handled separately if this path becomes too frequent.

## ELI5

On Windows, after you install a CLI tool, Orca could still say it was missing until restart because PATH was cached. Forced preflight checks now refresh PATH from the registry so newly installed tools are found right away.

### DIFF
--- a/src/main/ipc/preflight-agent-detection-no-subprocess.test.ts
+++ b/src/main/ipc/preflight-agent-detection-no-subprocess.test.ts
@@ -17,6 +17,7 @@ const {
   getAzureDevOpsAuthStatusMock,
   getGiteaAuthStatusMock,
   detectCommandsInInstallDirsMock,
+  mergePersistedWindowsPathAsyncMock,
   mergePersistedWindowsPathMock
 } = vi.hoisted(() => ({
   handleMock: vi.fn(),
@@ -29,6 +30,7 @@ const {
   getAzureDevOpsAuthStatusMock: vi.fn(),
   getGiteaAuthStatusMock: vi.fn(),
   detectCommandsInInstallDirsMock: vi.fn(),
+  mergePersistedWindowsPathAsyncMock: vi.fn(),
   mergePersistedWindowsPathMock: vi.fn()
 }))
 
@@ -60,6 +62,7 @@ vi.mock('./local-agent-install-dir-detection', () => ({
 
 // Win32 preflight env merge reads persisted registry PATH; stub it out.
 vi.mock('../pty/windows-environment-path', () => ({
+  mergePersistedWindowsPathAsync: mergePersistedWindowsPathAsyncMock,
   mergePersistedWindowsPath: mergePersistedWindowsPathMock
 }))
 
--- a/src/main/ipc/preflight-windows-path-refresh.repro.test.ts
+++ b/src/main/ipc/preflight-windows-path-refresh.repro.test.ts
@@ -1,0 +1,61 @@
+import { copyFileSync, mkdtempSync, rmSync } from 'node:fs'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+const { registryQueryMock } = vi.hoisted(() => ({ registryQueryMock: vi.fn() }))
+
+vi.mock('node:child_process', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  execFileSync: registryQueryMock
+}))
+
+import {
+  __resetPersistedWindowsPathCacheForTests,
+  mergePersistedWindowsPath
+} from '../pty/windows-environment-path'
+import { execLocalPreflightCommand } from './preflight-command-exec'
+
+describe.runIf(process.platform === 'win32')('Windows preflight Path refresh reproduction', () => {
+  const originalPath = process.env.Path ?? process.env.PATH ?? ''
+  const fixtureDirs: string[] = []
+
+  afterEach(() => {
+    process.env.Path = originalPath
+    registryQueryMock.mockReset()
+    __resetPersistedWindowsPathCacheForTests()
+    for (const directory of fixtureDirs.splice(0)) {
+      rmSync(directory, { recursive: true, force: true })
+    }
+  })
+
+  it('finds a newly installed executable immediately after a forced refresh', async () => {
+    const directory = mkdtempSync(join(tmpdir(), 'orca-path-refresh-'))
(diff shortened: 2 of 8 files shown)


OrcaWin:
Taking a look!

OrcaWin:
Reviewed and validated this on a real Windows 11 machine. Splitting what was **actually executed** from what was only reasoned about, since that's the part that matters here.

### Verified by execution

- **Reproduced #9316 A/B on two real builds.** With the fix reverted, a forced re-detect after a real `HKCU\Environment\Path` change returned `installed: false` at t+10s, t+15s and t+21s, and only flipped to `true` at t+39s — exactly the reporter's "first, second, and third times had no effect." With the fix, `true` on the **first** re-detect (t+12s). Real registry mutation, real app, no restart.
- **WSL isolation holds against real `wsl.exe`.** With a logging `reg.exe` shim loaded inside the running app: a forced *host* preflight makes exactly two `reg.exe` calls (HKLM then HKCU); a forced *WSL* preflight makes **zero**. The stubbed unit test matches real behavior.
- **Edge cases, all run against the live registry:** `REG_EXPAND_SZ` parsing, `%VAR%` expansion, a 30,000-char PATH (no `reg.exe` line wrapping or truncation), machine-before-user precedence, merge dedup, and PATH length stable across repeated forced merges. Non-Windows is an early return.
- **Tests are not vacuous.** All five fail when their source change is reverted. The WSL test was checked by removing *only* the guard — deleting the whole block would have made a negative assertion pass trivially.

### Changed during review

- `7a8e205f1` — `readPersistedWindowsPathSegments` wrote its result to the cache unconditionally. Since a forced read bypasses the cache, a blocked `reg.exe` replaced a known-good PATH with an empty list for the rest of the TTL. Reproduced live (21 segments → 0, and 0 for the next 30s); now keyed on thrown reads so a genuinely emptied PATH still clears the cache. Re-verified live: 21 → 21.
- `216063623` — one-line comment on the WSL guard, which previously sat directly above the unrelated GitLab `// Why:` block and read as if that comment explained it.

### Follow-up, deliberately not fixed here

The forced refresh runs two **synchronous** `reg.exe` spawns on the main process: 815 ms min / 1668 ms median / 2423 ms max on this box (one where even `cmd /c exit` costs 165–425 ms; on unencumbered hardware expect ~40–100 ms). `Landing.tsx:246` fires a forced preflight on every window focus and `Landing.tsx:286` on a 30s interval while a warning is visible, so those automatic call sites now pay it every time rather than once per TTL.

Not a blocker: `buildLocalPreflightEnv()` already ran that same sync pair on every cold cache before this PR, so this is a frequency change to accepted behavior, not a new regression class. `<Landing />` only mounts in the empty state, so the stall can't land on a user looking at a terminal, and in the #9316 scenario it's self-limiting — the banner clears and the interval stops. Every throttle I considered reopens a staleness window on the exact click this PR fixes, which is the wrong trade. The clean fix is an async variant of the persisted-path read used only from `runPreflightCheck` (already async), sharing the parser and cache — worth its own change.

### Declined

Extending the same refresh to `refreshShellPathAndDetectAgents` (the Agents-pane Refresh). The premise is correct — `hydrateShellPath` returns no shell on win32, so that button can't see a newly installed agent CLI either — but it's a different surface from #9316 and would widen this into agent detection. Deserves its own issue.

### Gaps

No screenshot: the change is non-visual (detection timing), so the evidence is IPC-level A/B across two builds. WSL *CLI* results on my box are unusable because that distro's profile is broken — irrelevant to the `reg.exe`-call evidence, which is what the WSL check turns on.

Scope unchanged at 4 files, CI green. Good diagnosis — the root cause was pinned correctly.

xianjianlf2:
Thanks for the thorough Windows validation. The A/B result is exactly the signal I was hoping for here, especially the host-vs-WSL reg.exe call split.

Agree on keeping the async persisted-PATH read and the Agents-pane refresh out of this PR. Those are real follow-ups, but this patch should stay focused on the forced preflight path for #9316.

brennanb2025:
Validated end to end — **LAND**.

- **Real Windows 11:** with a warm PATH cache, added portable `glab` to persisted HKCU PATH after launch; the visible **Re-check** detected it 2.05s later without relaunch. IPC median/p95 stayed at 1ms; renderer/main logs had no errors; registry and processes were restored cleanly.
- **Performance/correctness:** forced refresh uses two bounded, async, deduplicated registry reads, preserves the last good PATH on failure, and skips host refresh for WSL. Focused tests: 53/53; node typecheck passed.
- **Real SSH:** Debian 11 / glibc 2.31 container, real remote repo and relay; visible Re-check settled, a PTY mutation was independently verified over host SSH, and cleanup was clean.
- Current head `7f8d62f448` was reviewed after its rebase/hardening; it preserves the validated contract and adds Windows repro coverage. All CI is green, including static analysis, typecheck, 16 test shards, package, Git compatibility, Windows crash-survival, and Windows terminal restart.

Issue #10004, opened 2026-07-22
Windows: agent PTY child processes orphaned on terminal stop/close, blocking worktree removal (`Failed to physically stop every PTY`)


Noobear:
## Summary
On Windows, removing a worktree that has (or recently had) a TUI agent running in a terminal fails with:

```
Error invoking remote method 'worktrees:remove': Error: Failed to physically stop every PTY for worktree: <repoId>::<path>
```

The failure happens from **both** the desktop UI (right‑click → delete) and the CLI (`orca worktree rm --force`), because both hit the same runtime precondition ("stop every PTY").

Root cause: when a terminal is stopped/closed, Orca kills the **direct PTY child (the shell)** but not the shell's descendants. The agent process and its children (node / `claude` / `codex` + their MCP‑server node processes) survive as **orphans** — their parent PID no longer exists — with their working directory still inside the worktree. On Windows there is no automatic process‑group cascade, so the subtree leaks and keeps the worktree "busy" forever.

There is also a **secondary reporting bug**: `terminal stop` returns `{stopped: N}` and `terminal close` returns `ok`, but the processes are still alive and `terminal list` still shows the terminals as `connected: true`.

## Environment
- OS: Windows 11 Pro (build 26200)
- Orca daemon-host: 1.4.146 / 1.4.148 / 1.4.150 installed (multiple runtimes live at the time)
- PTY layer: node-pty + `OpenConsole.exe` (ConPTY), spawned under `…/Orca/daemon-host/<ver>/resources/node_modules/node-pty/…`
- Agents involved (both reproduce):
  - Claude Code — `claude --dangerously-skip-permissions`
  - OpenAI Codex TUI v0.145.0 — `codex --dangerously-bypass-approvals-and-sandbox` (YOLO). Process tree: `node.exe → codex.exe → pwsh.exe` + several `node ./mcp/server.mjs` children.

## Reproduction
1. In Orca on Windows, open a worktree and start a TUI agent in a terminal (Claude Code or Codex). The agent spawns a shell → node agent → child processes (incl. MCP servers).
2. Try to remove the worktree — UI delete **or** `orca worktree rm --worktree id:<…> --force --json`.
3. It fails: `Failed to physically stop every PTY for worktree: <repoId>::<path>`.
4. Try to clear the PTYs first:
   - `orca terminal stop --worktree id:<…> --json` → returns `{ "stopped": 3 }` (and `{ "stopped": 1 }` for the other worktree).
[...]

PR #10092, opened 2026-07-23
fix(daemon): respawn terminal host for listProcesses/shutdown

### LINKED ISSUE #10087: [Bug][Windows] Worktree removal effectively always fails: connect ENOENT on orca-terminal-host pipe
## Summary

On the **Windows** build of Orca, removing a worktree from the UI **effectively always fails** — in practice I cannot delete a worktree at all through the normal flow. Every attempt surfaces the same IPC error pointing at the `orca-terminal-host` named pipe. The only reliable workaround is to fully quit Orca, force-kill all Orca processes, relaunch, and delete the worktree immediately — before opening any terminal or agent.

Additionally, the **Resource Manager** panel's process-kill button (the "X") is completely non-functional: clicking it does nothing, and no process can be terminated from there, removing the only in-app escape hatch for cleaning up stuck processes.

## Frequency

Near 100%. I would describe it as "cannot delete a worktree at all" without the restart workaround — it is not intermittent.

## Error

```
无法删除工作区 Issue 18 关于"创建项目"的功能优化需求
Error invoking remote method 'worktrees:remove': Error: connect ENOENT \?\pipe\orca-terminal-host-v24-b51f469ab4fe
```

(The first line translates to: "Cannot delete worktree 'Issue 18 ...'".)

## Steps to reproduce

1. Open a worktree in Orca and use it normally for a while — open terminals, run Claude Code / Codex sessions, etc.
2. Attempt to remove the worktree (worktree card → Remove).
3. The error above is shown and the worktree is **not** removed.

## Expected behavior

`worktrees:remove` succeeds regardless of whether terminals or agent sessions are/were running in the worktree.

## Environment

- **Platform: Windows (native build)** — this is the Windows version of Orca, not WSL/macOS/Linux.
- Orca: v24 (commit `b51f469ab4fe`, inferred from the pipe name `orca-terminal-host-v24-b51f469ab4fe`)
- OS: Windows 11 Pro (10.0.26200)
- Shell: PowerShell 7

## What does NOT work

- **Resource Manager kill button**: In the Resource Manager panel (bottom-right), clicking the "X" on any listed process does nothing — no process can actually be terminated. The button appears entirely non-functional.

## Workaround

The only reliable way to remove a worktree:

1. Fully quit Orca.
2. Force-kill every remaining Orca-related process (Task Manager / `taskkill`).
3. Relaunch Orca.
4. Delete the worktree **immediately**, before opening any terminal or agent session.

This strongly suggests the `orca-terminal-host` process (and its named pipe) dies or goes stale during normal use, and the `worktrees:remove` IPC handler hard-depends on it — so removal cannot work until Orca is restarted with a fresh pipe.

## Suspected cause / request

The `worktrees:remove` IPC call appears to require a live `orca-terminal-host` reachable via `\?\pipe\orca-terminal-host-v*`. After some usage the host seems to become unreachable (crash / never respawned / stale pipe), so `connect` fails with `ENOENT`. Restarting Orca respawns the host, which is why "restart → immediately delete" works.

It would help if:
- `worktrees:remove` didn't hard-fail when the terminal-host pipe is gone (graceful degradation / respawn-on-demand), and
- the Resource Manager "X" actually terminated processes as advertised.

## Related

Symptoms overlap, but the error here is distinct from:
- #10004 / #9045 — those report `Failed to physically stop every PTY` (orphaned child processes). Here the failure is upstream: the terminal-host pipe itself is gone (`connect ENOENT`), so removal never reaches PTY teardown.
- #8275 — terminal daemon dying; may share a root cause.

innocarpe:
## Summary

On Windows, worktree remove could hard-fail with:

```text
Error invoking remote method 'worktrees:remove': Error: connect ENOENT \\?\pipe\orca-terminal-host-…
```

after normal terminal/agent use. Restarting Orca and deleting immediately worked, because a fresh terminal host pipe exists.

### Cause

Destructive teardown inventories PTYs via `DaemonPtyAdapter.listProcesses()` (and stops via `shutdown()`). **Spawn** already wraps operations in `withDaemonRetry` (respawn on `connect ENOENT` / other daemon-gone errors). **listProcesses / shutdown** only called `ensureConnected` without that retry, so a dead named pipe failed remove until full app restart.

### Fix

- Wrap `listProcesses` in `withDaemonRetry`
- Use `withDaemonRetry` for the connect step of `shutdown`

After a dead host, inventory can respawn once and return an empty session list so remove proceeds instead of blocking on pipe connect.

### Out of scope

- Resource Manager kill button no-op (separate surface; not fixed here)
- OS-level orphaned child processes after host death (#10004 / #9045)

Fixes #10087

## Screenshots

No visual change.

## Testing

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/daemon-pty-adapter.test.ts -t "listProcesses|respawns a dead terminal host"` (2/2)
- [ ] Full suite / Windows manual delete after agent use

## AI Review Report

- Cross-platform: Windows named pipes and Unix sockets both surface `connect` ENOENT/ECONNREFUSED; existing `isDaemonGoneError` already covers them
- Destructive path deadline still applied via `ensureConnected(deadlineMs)`
- Does not loosen `requirePhysicalStop` when sessions still exist and stop fails

## Security Audit

- No new IPC; reuses existing daemon respawn path used by spawn
- Respawn still gated by adapter lifecycle (`respawnAdoptionClosed` / missing `respawn` callback)

## Notes

- If the host died with OS orphans still holding files, remove may still hit later PTY/physical-stop errors — that remains the #10004 class of issues


innocarpe:
Accidentally closed when a bad rebase helper force-pushed an empty tip. Restored and continued as https://github.com/stablyai/orca/pull/10350.

PR #10097, opened 2026-07-23
fix(win): strengthen GPU software fallback for virtual displays

innocarpe:
## Summary

Fixes Windows headless / virtual-display hosts (e.g. Sunshine + Zako) where Chromium's GPU child STATUS_BREAKPOINT-crashes on every launch and Orca never reaches a usable window.

## ELI5

When Windows graphics is so broken that Orca keeps dying at startup, Orca now restarts itself in a safer "draw everything in software" mode that actually works on virtual/remote displays — and you can force that mode yourself with an environment variable.

## Root cause & fix

Orca's existing crash-burst fallback relaunched with only `disableHardwareAcceleration` + `--disable-gpu` — the exact flags reporters had already tried without success — so the GPU child kept crashing. This PR routes the Windows software-GPU path through the combo proven to reach `ready` on those hosts: `--disable-gpu` + `--in-process-gpu` + `--use-angle=swiftshader`, applied before `app.whenReady()`. It also adds a first-launch opt-in via `ORCA_SOFTWARE_GPU=1` so operators can skip the crash-burst cycle on known-bad adapters, while preserving the Windows-only, non-serve guard and the sticky-marker path.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase applied cleanly onto `origin/main`.

- Test on branch: `windows-software-gpu.test.ts` exercises the fallback combo (disableHardwareAcceleration + disable-gpu + in-process-gpu + use-angle swiftshader) and `ORCA_SOFTWARE_GPU` env parsing.

```
npx vitest run --config config/vitest.config.ts src/main/startup/windows-software-gpu.test.ts
=> Test Files 1 passed (1), Tests 4 passed (4)
```

- Fix reverted, test kept: re-running fails with `Cannot find module './windows-software-gpu'` (suite fails to load, 0 tests run) — the fix module is net-new, so the failure manifests as module-not-found rather than an assertion diff. Pass-on-branch AND fail-when-reverted = TRUE.
- Lint clean: `npx oxlint src/main/startup/windows-software-gpu.ts src/main/index.ts` => exit 0.

## Trade-offs

After a GPU crash burst (unchanged 3-crashes/30s window) the software path is stricter (in-process + SwiftShader), so terminal/WebGL may be slower on those already-broken hosts. Healthy GPUs are unaffected; `ORCA_SOFTWARE_GPU=1` is opt-in only.

## Regression risk

Low. Changes are gated behind the existing Windows-only, non-serve guard plus the crash marker or explicit env opt-in; the `!envRequested && !marker` early-return is preserved, so non-Windows and healthy-GPU startup paths are untouched. The passing test proves the new branch behavior.

_Credit to @innocarpe (Wooseong Kim). Validated, rebased, and hardened via automated bug-bash review; original fix design preserved._

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -136,6 +136,10 @@ import {
 } from './startup/gpu-fallback-marker'
 import { applyGpuFallbackCommandLineSwitches } from './startup/gpu-fallback-switches'
 import {
+  applyWindowsSoftwareGpuFallback,
+  isSoftwareGpuEnvRequested
+} from './startup/windows-software-gpu'
+import {
   DEFAULT_GPU_CRASH_FALLBACK_THRESHOLD,
   DEFAULT_GPU_CRASH_FALLBACK_WINDOW_MS,
   GpuCrashFallbackTracker,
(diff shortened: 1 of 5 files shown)


nwparker:
Valid but we may want to inform the user. Let me think on this a sec

innocarpe:
### Sync update (`8f345f947`)

Address review: surface a one-shot notice when software GPU fallback engages.

- Commit: `8f345f9478f628f71ac9c2d425e98c2f80e95382`
- Upstream: https://github.com/stablyai/orca/pull/10097
- Fork portfolio: https://github.com/innocarpe/orca/pull/25

innocarpe:
### Sync update (`90ffcece20`)

Rebase onto latest upstream/main: env software-GPU opt-in + notice; keep main post-crash non-SwiftShader switches.

- Commit: `90ffcece20adf47a5962f778660d99a0214cd178`
- Upstream: https://github.com/stablyai/orca/pull/10097
- Fork portfolio: https://github.com/innocarpe/orca/pull/25

Jinwoo-H:
Thanks @innocarpe for investigating the Windows GPU fallback path. Merged fix #11295 now owns this behavior on current main and deliberately avoids the SwiftShader/in-process combination proposed here, which would broaden privileged parsing of untrusted WebGL content. We will retain your diagnosis as historical credit and request no rebase, retest, or other follow-up. Closing this superseded branch.

PR #10109, opened 2026-07-23
fix(agent-status): keep parent resume when nested Codex shares pane key

### LINKED ISSUE #10105: [Bug]: Returning from Agents UI errors when Claude session has a nested Codex worker
## Summary

Clicking **back/return** from the Agents UI fails when the active session is a **Claude** session that has dispatched a **Codex** worker. Orca incorrectly treats the Claude session as the primary Codex session on return, then blocks automatic resume with a Codex session-file verification error.

## Error

```
Orca could not verify the originating Codex session file, so automatic resume was stopped to avoid using a different account.
```

## Steps to reproduce

1. Start / open a **Claude** agent session (this is the real primary session).
2. Have Claude dispatch / spawn a **Codex** agent to do work (nested/worker Codex).
3. Open the **Agents** UI (`智能体` / agent interface).
4. Click the **back / return** control to leave the Agents UI and return to the session.

## Expected

Orca returns to the original **Claude** session without trying to resume it as Codex. Nested Codex work should not change which agent owns the primary session.

## Actual

On return, Orca assumes this session is the **primary Codex** session. Automatic resume is stopped because Orca cannot verify the originating Codex session file (to avoid using a different account), and the error banner is shown.

## Screenshot

Error when returning from Agents UI with Claude session that has nested Codex

## Notes / likely cause

Session ownership / resume identity appears to follow the nested Codex worker instead of the originating Claude session when leaving the Agents UI. Resume should key off the **primary session agent** (Claude here), not the worker Codex that was dispatched inside that session.

innocarpe:
## Summary

- Keep the active parent agent as the pane owner when a nested provider hook reuses the parent's pane key.
- Ignore nested provider-session claims from a different agent while the parent identity is fresh and non-done.
- Prevent inherited child metadata from replacing the parent's resume session in the status update path.

Fixes #10105.

## Why

When Claude starts a nested Codex worker, both CLIs can inherit the same ORCA_PANE_KEY. The nested hook could then write a Codex resume record or attach the Codex session id to the live Claude row. A later cold restore or Agents return could launch the wrong provider/session file.

## Scope and design

- The guard is applied at both provider-session claim and status-metadata ingress points.
- A different provider may claim the pane after the existing parent is done or stale.
- Same-agent session refreshes remain allowed.
- Existing done-state provider-session retention and recovery semantics on main are preserved.
- The change is O(1) per identity/provider update and does not change the visible UI model beyond keeping the correct parent identity.

## Test plan

- [x] Parent Claude + nested Codex provider-session regression
- [x] Inherited identity and stale/done boundary unit tests
- [x] Existing provider-session suite: 18 passed
- [x] Typecheck
- [x] Oxlint, React Doctor lint, Oxfmt, and diff checks
- [ ] Manual: return from Agents after a Claude session starts a nested Codex worker and verify Claude resume remains selected

## Screenshots

No visual chrome changes.

## Security audit

The guard prevents a nested provider from substituting another account's session file during resume. No new filesystem, network, persistence, or credential behavior is introduced.

## Platform review

The identity logic is shared renderer state and applies equally to macOS, Linux, Windows, and SSH/remote panes. It does not depend on a provider-specific shell or path format.

## ELI5

Claude can start a Codex helper in the same terminal slot. Orca now remembers that Claude still owns the slot until Claude actually finishes, so returning later does not open the helper's session by mistake.

### DIFF
--- a/src/renderer/src/store/slices/agent-status-provider-session.test.ts
+++ b/src/renderer/src/store/slices/agent-status-provider-session.test.ts
@@ -121,6 +121,74 @@ describe('recordAgentProviderSession', () => {
     expect(store.getState().agentStatusByPaneKey['tab-1:leaf-1']?.providerSession).toBeUndefined()
   })
 
+  it('does not let nested Codex overwrite a live Claude parent resume identity (#10105)', () => {
+    const store = createTestStore()
+    store.setState({
+      tabsByWorktree: {
+        'wt-1': [makeTab({ id: 'tab-1', worktreeId: 'wt-1' })]
+      }
+    } as Partial<AppState>)
+    const claudeSession = {
+      key: 'session_id' as const,
+      id: 'claude-session-1',
+      transcriptPath: '/tmp/claude.jsonl'
+    }
+    const nestedCodexSession = {
+      key: 'session_id' as const,
+      id: 'codex-worker-1',
+      transcriptPath: '/tmp/codex-worker.jsonl'
+    }
+
+    store
+      .getState()
+      .setAgentStatus(
+        'tab-1:leaf-1',
+        { state: 'working', prompt: 'parent turn', agentType: 'claude' },
+        'Claude',
+        { updatedAt: 10, stateStartedAt: 10 },
+        { tabId: 'tab-1', worktreeId: 'wt-1' },
+        { providerSession: claudeSession }
+      )
+
+    store
+      .getState()
+      .recordAgentProviderSession(
+        'tab-1:leaf-1',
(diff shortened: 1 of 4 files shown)


innocarpe:
### Sync update (`5568050ed`)

Rebase onto latest upstream/main (cleanup batch).

- Commit: `5568050ed670f364c55514d06ef44205fbf3b3d1`
- Upstream: https://github.com/stablyai/orca/pull/10109
- Fork portfolio: https://github.com/innocarpe/orca/pull/29

PR #10116, opened 2026-07-23
fix(terminal): infer keydown-captured interrupts on fire-and-forget SSH transports

### LINKED ISSUE #10114: [Bug]: Esc-interrupting a Claude turn in an SSH worktree leaves the pane 'working' forever — interrupt recovery is gated behind sendInputAccepted, which SSH transports omit
### Describe the bug

In an **SSH worktree**, interrupting a Claude Code turn with **Esc** leaves the pane/worktree showing **"working"** (spinner) indefinitely. Claude itself returns to its idle prompt (`⎿ Interrupted · What should Claude do instead?`), but the sidebar keeps spinning until the user sends a **new message**, which finally clears it.

Local worktrees are unaffected — only SSH.

### Reproduction (verified live on a dev build)

1. Create an SSH worktree with Claude as the agent.
2. Send a prompt that makes Claude generate for a while.
3. Press Esc mid-generation.
4. Claude interrupts and shows its idle prompt; the live OSC title even reverts to the idle `✳ …` form.
5. The agent-status entry stays `state: "working"` indefinitely (observed 80+ seconds, unchanged until the next message).

Captured state during the stuck window:

| Signal | Value |
| --- | --- |
| Live terminal title | `✳ <task>` (idle — Claude reset it) |
| agent-status entry | `state: working`, stale spinner title, frozen `updatedAt` |
| Claude TUI | `⎿ Interrupted · What should Claude do instead?` + idle prompt |

### Root cause

1. **Claude fires no `Stop` hook on Esc-interrupt** (hooks: `UserPromptSubmit`/`PreToolUse` set `working`; only `Stop`/`StopFailure` clear it). The hook row freezes at `working`.
2. Orca already has two client-side recovery paths for exactly this: hook-row interrupt inference (`agent-interrupt-inference.ts` → `agentStatus:inferInterrupt`) and title-only neutralization (`observeTitleOnlyInterrupt`). **Both are wired only into the acknowledged-write callback** (`pty-connection.ts` — inside `transport.sendInputAccepted(...).then(...)`).
3. **SSH transports omit `sendInputAccepted`** (`pty-transport.ts` — `...(connectionId ? {} : { sendInputAccepted })`), and main's `writePtyInputAccepted` explicitly refuses SSH ids ("SSH providers are fire-and-forget relay notifications and can't truthfully acknowledge yet"). So on SSH the Esc falls into the fire-and-forget `sendInput` branch, which never triggers either recovery.
4. Safety nets don't rescue it: hook rows only decay after `AGENT_STATUS_STALE_AFTER_MS` (30 min), and the sidebar hook row outranks the (already idle) title.

So the pinned exclusion of interrupt inference on unacknowledged transports turns "Esc over SSH" into a permanently stuck working indicator.

### Expected behavior

After Esc interrupts a Claude turn in an SSH worktree, the working indicator should clear within the normal interrupt-inference settle window, as it does for local and remote-runtime panes.

### Environment

- macOS desktop, SSH worktree (relay), Claude Code with Orca-managed hooks active.
- Reproduced on a current `main` dev build.

choihjin:
## Summary

Pressing Esc to interrupt a Claude turn in an **SSH** worktree left the pane stuck showing "working" forever. Interrupting an agent produces no Stop hook, so the only signal that the turn ended is the interrupt itself — and on SSH the input transport is fire-and-forget, so the renderer never learns the keystroke was delivered and never clears the status.

The fix feeds DOM-captured interrupt intents (`Esc`, `Ctrl+C`) into the existing interrupt-inference path on fire-and-forget transports. Crucially it infers only from a **keydown-derived intent**, never from raw bytes: bytes alone are not proof of a user interrupt.

Closes #10114

## ELI5

When you press Esc to stop the AI mid-answer, Orca normally notices and flips the little "working…" badge back to idle. Over a remote SSH connection, Orca sends the keystroke but never gets a receipt, so it kept showing "working…" forever even though the AI had already stopped — and nothing would ever clear it.

Now, when *you* physically press Esc, Orca takes that as good enough evidence to clear the badge. It's careful about what counts: holding Alt and pressing a letter technically sends the same invisible character as Esc, and that does **not** count. And if the keystroke somehow never made it, the next real status update from the AI puts the badge back.

## Proof of fix

Two tests fail against this branch's merge base and pass here:

```
$ git checkout $(git merge-base HEAD origin/main) -- src/renderer/src/components/terminal-pane/pty-connection.ts
    src/renderer/src/components/terminal-pane/pty-connection.test.ts \
    -t "fire-and-forget"

 × infers a keydown-captured interrupt on fire-and-forget transports (#SSH stuck working)
     AssertionError: expected "vi.fn()" to be called with arguments: [ { …(6) } ]
     Number of calls: 0
 × clears title-only working indicators after a keydown-captured fire-and-forget interrupt
     AssertionError: expected "vi.fn()" to be called with arguments: [ 'tab-1', 1, 'Terminal' ]
     Number of calls: 0

      Tests  2 failed | 2 passed | 465 skipped (469)
```

```
$ git checkout HEAD -- src/renderer/src/components/terminal-pane/pty-connection.ts
    src/renderer/src/components/terminal-pane/pty-connection.test.ts \
    -t "fire-and-forget"

      Tests  4 passed | 465 skipped (469)
```

Use the **merge-base** copy of the file, not `origin/main`. An earlier version of this section checked out `origin/main`; `main` has since drifted and now imports `../../../../shared/worktree-removal-fence-error`, whose sibling does not exist on this branch, so that run dies with `Cannot find module` before a single assertion executes — a module-resolution failure, not evidence.

The `2 passed` in the first run are the two false-positive guards. They pass at the merge base **by construction** — code that never infers cannot infer wrongly — so they are not proof that this change does anything. They exist to pin the two ways the inference goes wrong once it exists:

- **Alt chord emitting an ESC prefix.** `Alt+f`, `Alt+b`, and every other Alt shortcut send `0x1b` + the key — byte-identical to a bare Esc. Inferring from bytes would mark an agent idle every time a user pressed an Alt shortcut while it was working. Inference is therefore keydown-derived only.
- **Pane with no working agent status.** The inference must be inert unless the pane is actually in a `working` state, so an Esc in a plain shell (or in vim, or dismissing a TUI menu) cannot manufacture a status transition.

## Proof of no regressions

```
$ npx vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/
 Test Files  182 passed (182)
      Tests  2340 passed (2340)

```

- On a **fire-and-forget (SSH) transport**, a DOM-captured Esc/Ctrl+C while an agent is `working` now clears the stuck status. Intended fix.
- The **local** path is untouched: it has real delivery acknowledgement, so it keeps using that and gains no inference.
- Exact-byte interrupt inference remains excluded, so pasted or programmatically-written `0x1b` cannot flip status.

## Compatibility

- **Platforms:** macOS / Linux / Windows. The inference keys off a keydown-derived intent already normalized upstream of this code, so no `e.metaKey` is introduced or hardcoded here — per `AGENTS.md`, the Cmd/Ctrl platform branch lives in the existing key handling and is untouched.
- **SSH / remote:** This is the entire point of the change, and the scoping is the load-bearing part. Inference is applied **only** where the transport cannot acknowledge delivery; transports that do acknowledge keep their existing, stronger signal. The failure mode of a dropped keystroke is bounded: the inference settle window re-checks the status entry, so a keystroke that never arrived self-corrects on the next hook ping rather than latching a wrong idle state. Optimistic, not authoritative.
- **Performance:** No hot-path change. One call on an input path that already ran `observeAcceptedTerminalInput`; no new timers per keystroke, no added IPC, no re-render trigger. Terminal panes are created and destroyed constantly, and this adds no per-pane resource to leak.

## Screenshots

No visual change beyond the agent-status badge clearing when it previously stuck. The behavior is asserted in `pty-connection.test.ts` — including both false-positive guards — rather than captured as an image. I did not record a screen capture of an SSH interrupt; noting that plainly rather than implying a manual check that was not performed.

## Testing

- [ ] `pnpm lint` — not run in full.
- [ ] `pnpm test` — not run in full (30+ min). Ran the whole touched directory: `src/renderer/src/components/terminal-pane/` → 182 files, 2340 tests passed.
- [ ] `pnpm build` — not run.
- [x] Added or updated high-quality tests that would catch regressions — two behaviour tests that fail at the merge base (keydown-captured interrupt inference; title-only indicator clearing) plus two false-positive guards (Alt-chord ESC prefix; no-working-status pane).

## AI Review Report

Reviewed with Claude Code (Opus 5) and re-verified independently. Risks checked:

- **False positives — the dominant risk.** Inferring intent is only safe if the signal cannot be produced accidentally. Enumerated the ways `0x1b` reaches a PTY without being a user interrupt: Alt chords (ESC-prefixed), Esc in vim normal mode, Esc dismissing a TUI completion menu, and bracketed paste containing `0x1b`. The fix infers from a keydown-derived intent rather than bytes, and is inert unless the pane is `working` — so none of these can flip an agent to idle. The Alt-chord case is pinned by a dedicated test because it is the one most likely to be missed.
- **The mirror-image bug.** Marking an agent idle while it is genuinely still working is arguably worse than the stuck badge, since the user stops waiting. Verified the inferred state is optimistic and correctable: the settle window re-checks the status entry, and a real hook event overrides it. It is not a latch.
- **Scope to fire-and-forget only:** confirmed the local acknowledged path is unchanged, so no status handling regresses for the majority of users.
- **Lifetime/leaks:** no new per-pane timer or listener is introduced on the keystroke path.
- **Cross-platform:** no paths, shell behavior, or Electron APIs touched; no new modifier handling.

## Security Audit

- **Input handling:** the change consumes an already-parsed keydown intent; it adds no new parsing of untrusted terminal bytes. Notably it *declines* to infer from raw bytes, which is the safer choice — remote output or a paste cannot synthesize an interrupt.
- **Command execution:** none. No new input is written to the PTY; this only affects how an already-sent keystroke updates local UI status.
- **Path handling:** none.
- **IPC:** no new channel or payload; the existing send path is unchanged.

## Notes

- Adjacent in-flight work that touches neighbouring agent-status / PTY-write paths and is worth sequencing against: **#9507** (suppressing "Task complete" while a Codex tool is still running) and **#9716** (reporting writes to a gone PTY as not-written). No file overlap with either, but all three shape when a pane's agent status changes.
- The inference is deliberately optimistic. If a future change makes SSH input delivery acknowledgeable, this path should be retired in favour of the real signal rather than kept alongside it.
- **Known residual — title-only panes do not get the per-agent intent rules.** `observeTitleOnlyInterrupt` runs only when a pane has *no* explicit agent-status row, and it is agent-agnostic by design (`pty-connection.ts` — "title-only agents such as Pi can miss their own idle title after Ctrl+C"). It therefore skips the per-agent rules in `agent-interrupt-inference.ts`: double-Escape for `opencode`/`copilot`, ignored Ctrl+C for `droid`. So a single Escape on a title-only OpenCode pane can neutralize a still-working spinner title after the settle window. That gap is pre-existing — the acknowledged path at `:3737` calls the same helper unconditionally, so local panes have it today — but SSH panes are disproportionately title-only because hooks are unavailable there (#8711), which raises its exposure. Resolving the agent from the title / `launchAgent` before arming title-only cleanup is the fix; deliberately left out of scope here rather than widening a shared path in a bug PR.

Original patch by @choihjin.

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-connection.test.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.test.ts
@@ -3401,9 +3401,11 @@ describe('connectPanePty', () => {
       onDataHandler = handler
       return { dispose: vi.fn() }
     }) as typeof pane.terminal.onData)
+    const manager = createManager(1)
+    manager.getActivePane.mockReturnValue({ id: 1 })
     const deps = createDeps()
 
-    connectPanePty(pane as never, createManager(1) as never, deps as never)
+    connectPanePty(pane as never, manager as never, deps as never)
     if (!onDataHandler) {
       throw new Error('expected onData handler to be registered')
     }
@@ -3414,7 +3416,81 @@ describe('connectPanePty', () => {
     await flushAsyncTicks()
 
     expect(transport.sendInput).toHaveBeenCalledWith('\x03')
-    expect(deps.setRuntimePaneTitle).toHaveBeenCalled()
+    expect(deps.setRuntimePaneTitle).toHaveBeenCalledWith('tab-1', 1, 'Terminal')
+    expect(deps.updateTabTitle).toHaveBeenCalledWith('tab-1', 'Terminal')
+  })
+
+  it('does not infer a fire-and-forget interrupt from an Alt chord that emits an Escape prefix', async () => {
+    const { connectPanePty } = await import('./pty-connection')
+    const transport = createMockTransport()
+    delete transport.sendInputAccepted
+    transportFactoryQueue.push(transport)
+    vi.useFakeTimers()
+    vi.setSystemTime(1_100)
+    const paneKey = makePaneKey('tab-1', LEAF_1)
+    mockStoreState.agentStatusByPaneKey[paneKey] = {
+      state: 'working',
+      prompt: 'keep working',
+      updatedAt: 1_000,
+      stateStartedAt: 900,
+      agentType: 'claude',
+      paneKey,
+      terminalTitle: '⠂ keep working',
+      stateHistory: []
+    }
+    const terminalTarget = createKeyboardEventTarget()
+    const pane = createPane(1)
+    ;(pane.terminal as { element?: unknown }).element = terminalTarget.target
+    let onDataHandler: ((data: string) => void) | null = null
+    pane.terminal.onData = vi.fn(((handler: (data: string) => void) => {
+      onDataHandler = handler
+      return { dispose: vi.fn() }
+    }) as typeof pane.terminal.onData)
+
+    connectPanePty(pane as never, createManager(1) as never, createDeps() as never)
+    if (!onDataHandler) {
+      throw new Error('expected onData handler to be registered')
+    }
+    // Alt+b sends ESC+b; the leading \x1b must not read as a user interrupt.
+    terminalTarget.dispatch(keyEvent({ key: 'b', altKey: true }))
+    ;(onDataHandler as unknown as (data: string) => void)('\x1bb')
+    await flushAsyncTicks()
+    vi.advanceTimersByTime(500)
+    await flushAsyncTicks()
+
+    expect(transport.sendInput).toHaveBeenCalledWith('\x1bb')
+    expect(window.api.agentStatus.inferInterrupt).not.toHaveBeenCalled()
+  })
+
+  it('does not infer a fire-and-forget interrupt for a pane with no working agent status', async () => {
+    const { connectPanePty } = await import('./pty-connection')
+    const transport = createMockTransport()
+    delete transport.sendInputAccepted
+    transportFactoryQueue.push(transport)
+    vi.useFakeTimers()
+    vi.setSystemTime(1_100)
+    const terminalTarget = createKeyboardEventTarget()
+    const pane = createPane(1)
+    ;(pane.terminal as { element?: unknown }).element = terminalTarget.target
+    let onDataHandler: ((data: string) => void) | null = null
+    pane.terminal.onData = vi.fn(((handler: (data: string) => void) => {
+      onDataHandler = handler
+      return { dispose: vi.fn() }
+    }) as typeof pane.terminal.onData)
+
+    connectPanePty(pane as never, createManager(1) as never, createDeps() as never)
+    if (!onDataHandler) {
+      throw new Error('expected onData handler to be registered')
+    }
+    // Esc in a plain shell/vim pane: no agent turn to interrupt.
+    terminalTarget.dispatch(keyEvent({ key: 'Escape' }))
+    ;(onDataHandler as unknown as (data: string) => void)('\x1b')
+    await flushAsyncTicks()
+    vi.advanceTimersByTime(500)
+    await flushAsyncTicks()
+
+    expect(transport.sendInput).toHaveBeenCalledWith('\x1b')
+    expect(window.api.agentStatus.inferInterrupt).not.toHaveBeenCalled()
   })
 
   it('infers exact Ctrl+C terminal input when keydown capture misses the press', async () => {
--- a/src/renderer/src/components/terminal-pane/pty-connection.test.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.test.ts
@@ -3334,7 +3334,54 @@ describe('connectPanePty', () => {
     expect(deps.updateTabTitle).toHaveBeenCalledWith('tab-1', 'Terminal')
   })
 
-  it('does not clear title-only working indicators when interrupt writes are unacknowledged', async () => {
+  it('infers a keydown-captured interrupt on fire-and-forget transports (#SSH stuck working)', async () => {
+    const { connectPanePty } = await import('./pty-connection')
+    const transport = createMockTransport()
+    delete transport.sendInputAccepted
+    transportFactoryQueue.push(transport)
+    vi.useFakeTimers()
+    vi.setSystemTime(1_100)
+    const paneKey = makePaneKey('tab-1', LEAF_1)
+    mockStoreState.agentStatusByPaneKey[paneKey] = {
+      state: 'working',
+      prompt: 'stop this task',
+      updatedAt: 1_000,
+      stateStartedAt: 900,
+      agentType: 'claude',
+      paneKey,
+      terminalTitle: '⠂ stop this task',
+      stateHistory: []
+    }
+    const terminalTarget = createKeyboardEventTarget()
+    const pane = createPane(1)
+    ;(pane.terminal as { element?: unknown }).element = terminalTarget.target
+    let onDataHandler: ((data: string) => void) | null = null
+    pane.terminal.onData = vi.fn(((handler: (data: string) => void) => {
+      onDataHandler = handler
+      return { dispose: vi.fn() }
+    }) as typeof pane.terminal.onData)
+
+    connectPanePty(pane as never, createManager(1) as never, createDeps() as never)
+    if (!onDataHandler) {
+      throw new Error('expected onData handler to be registered')
+    }
+    terminalTarget.dispatch(keyEvent({ key: 'Escape' }))
+    ;(onDataHandler as unknown as (data: string) => void)('\x1b')
+    await flushAsyncTicks()
+    vi.advanceTimersByTime(500)
+
+    expect(transport.sendInput).toHaveBeenCalledWith('\x1b')
+    expect(window.api.agentStatus.inferInterrupt).toHaveBeenCalledWith({
+      paneKey,
+      baselineUpdatedAt: 1_000,
+      baselineStateStartedAt: 900,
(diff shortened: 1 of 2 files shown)


nwparker:
## Verified: this is not superseded, and the proof section needs a fix

A review pass suggested main already contains this change and the production diff may be a no-op after rebase. I checked, and that's **not** correct — but the reasoning behind it is worth recording, because the two call sites look identical out of context.

`origin/main` does have `interruptInference.observeInputIntent(...)` + `observeTitleOnlyInterrupt()` in `pty-connection.ts`, at two places:

- **:1871**, inside the `observeSentTerminalInputIntent` helper — also present at this branch's merge base, so not new.
- **:3742**, gated on `if (acknowledgedIntent && transport.sendInputAccepted)` — the **acknowledged** transport path only, inside the `.then((accepted) => ...)` callback.

This PR adds a *third* call, at **:3763** in the `if (transport.sendInput(data))` branch — the fire-and-forget path that SSH actually uses, which has `sendInput` and no `sendInputAccepted`. Main never reaches inference there. Production call count is 2 on main, 3 here. The gap is real and the diff is not a no-op.

**The proof command in the body does need correcting**, though. As written (`git checkout origin/main -- pty-connection.ts`, then run) it now fails with `Cannot find module '../../../../shared/worktree-removal-fence-error'` — main has drifted and pulls in imports whose siblings don't exist on this branch. That's a module-resolution failure, which proves nothing. Against the **merge-base** version of the file it fails correctly:

```
× infers a keydown-captured interrupt on fire-and-forget transports (#SSH stuck working)
× clears title-only working indicators after a keydown-captured fire-and-forget interrupt
Tests  2 failed | 15 passed
```

Those are the two tests that genuinely prove the fix. The body currently quotes the two *negative* guards as failing on main; those are false-positive guards and pass by construction at the merge base. Please swap the quoted run for the merge-base one and cite these two test names.

Everything else held up under attack: inference here is keydown-derived only (a bracketed paste containing `0x1b` or an Alt-chord ESC prefix cannot reach it — exact-byte inference stays gated behind acknowledged transports), it's inert unless the entry is `working` at three independent layers, and it's not a latch — any later real hook event overwrites the inferred status through the same path. A merge of current main into this head is clean and green.

choihjin:
Body updated. The proof section now checks out the merge-base copy of `pty-connection.ts` instead of `origin/main`, and cites the two behaviour tests.

Confirmed both halves locally against merge base `1d8ce38a5`:

```
$ git checkout 1d8ce38a5 -- src/renderer/src/components/terminal-pane/pty-connection.ts
    src/renderer/src/components/terminal-pane/pty-connection.test.ts -t "fire-and-forget"

 × infers a keydown-captured interrupt on fire-and-forget transports (#SSH stuck working)
     AssertionError: expected "vi.fn()" to be called with arguments: [ { …(6) } ]
     Number of calls: 0
 × clears title-only working indicators after a keydown-captured fire-and-forget interrupt
     AssertionError: expected "vi.fn()" to be called with arguments: [ 'tab-1', 1, 'Terminal' ]
     Number of calls: 0

      Tests  2 failed | 2 passed | 465 skipped (469)
```

The `origin/main` variant does die on `Cannot find module '../../../../shared/worktree-removal-fence-error'` before any assertion runs, exactly as you described.

I kept the two guards in the same filtered run rather than dropping them — the `2 passed` line makes the point visible: they pass at the merge base by construction, so they were never evidence for the change. The body now says that explicitly instead of quoting them as the proof, and I corrected the Testing checklist line that repeated the same claim.

---

Separately, a Codex review pass raised one thing that I do not think should change this diff, but is worth recording rather than leaving for someone to rediscover:

`observeTitleOnlyInterrupt` is agent-agnostic, so it skips the per-agent rules in `agent-interrupt-inference.ts` — double-Escape for `opencode`/`copilot` (`requiresDoubleEscapeForAgent`), ignored Ctrl+C for `droid` (`shouldIgnoreInterruptIntent`). Those rules live inside `observeInputIntent`, which returns early when there is no status entry; `observeTitleOnlyInterrupt` runs only in that same no-status-entry case. So on a title-only pane a single Escape can neutralize a still-working spinner title.

Two reasons I left it alone:

1. It is pre-existing, not introduced here. The acknowledged path calls `observeTitleOnlyInterrupt()` unconditionally at `:3737`, so local title-only panes have this today on `main`. This change gives SSH the same behaviour rather than a new one.
2. The helper looks deliberately agent-agnostic — the comment there is about title-only agents such as Pi missing their own idle title after Ctrl+C — and the blast radius is one neutral title label, not a synthesized `done`; the process tracker still owns liveness.

The honest caveat is exposure, not correctness: SSH panes are disproportionately title-only because hooks are unavailable there (#8711), so this change does make a pre-existing gap easier to hit. Added it to the PR notes as a residual. If you want it closed, resolving the agent from the title / `launchAgent` before arming title-only cleanup would fix both transports at once — happy to do that as a separate PR so this one stays a single-behaviour fix, or to fold it in here if you would rather it ship together.

PR #10158, opened 2026-07-23
fix(terminal): stop the reveal fit from reflow-garbling inline TUIs on minimize→restore

nwparker:
## What & why

`grok` (and other **inline-viewport** TUIs like Codex) render badly garbled — overlapping / duplicated / misaligned frames — after Orca's **floating terminal is minimized and brought back up**. Reported in [Slack](https://stablygroup.slack.com/archives/C0ASMDT6LQZ/p1784788515610149) ("grok is pretty fucked in floating terminal after minimize + bring back up").

**Root cause.** On reveal, the resume path fit xterm **synchronously** right after `resumeRendering()` re-attaches the WebGL renderer. WebGL rounds cell dimensions to device pixels differently from the DOM renderer, so for a frame `FitAddon.proposeDimensions()` can return a **one-column-off** grid. The synchronous fit commits it, reflowing xterm, and the ~150 ms container `ResizeObserver` fits it back — a **net-zero resize "wiggle"**. xterm's main-buffer wrap→unwrap is **not a perfect inverse**, and an inline TUI that only diff-paints its pinned region (never a full clear, unless it detects a multiplexer) redraws over the corrupted buffer. The reveal fit was the one fit path that skipped the stability gate the ResizeObserver and drift-check paths already use.

**Fix.** Replace the unconditional synchronous reveal fit (`fitAllPanes`) with a gated one — `PaneManager.fitAllPanesStable()` → `fitRevealedPane(pane)`:

| On reveal, the fit element's pixels… | action |
|---|---|
| **changed** while hidden (a real resize) | **synchronous `safeFit`** — the app must reflow anyway, and doing it now keeps xterm ahead of the async `{fit:false}` PTY size reassert so it can't forward a stale grid |
| **unchanged**, and the proposed grid **matches** the terminal (the common minimize→restore) | **skip** — the grid is already correct; only release parked replay/reattach continuations |
| **unchanged**, but the grid **diverged** while hidden (snapshot / SSH-reattach `terminal.resize`, or an appearance/DPI change) | **`requestStablePaneFit`** — repair on a steady grid, so a *sustained* mismatch refits but a *transient* metric wobble does not reflow |

The pixel comparison measures the actual element FitAddon fits (`xtermContainer`), rounded to whole pixels so sub-pixel jitter never reads as a resize. Same gate applied to the window-wake reveal path.

The result: **the reported minimize→restore is now a hard no-op with zero reflow.**

## ELI5

grok draws its screen by only touching the characters that changed since last frame (it trusts everything else to stay put). When you minimized and reopened the floating terminal, Orca briefly told the terminal it was one column narrower, then a column wider again — snapping the text back and forth. The terminal's text-rewrapping isn't perfectly reversible, so the "everything else stays put" assumption broke, and grok painted new text on top of shuffled old text → garbage.

The fix: on reopen, Orca now only re-measures the terminal **when the window actually changed size**. If it's the same size (the usual case), it leaves the terminal exactly as it was — no snap, no garbage.

## Proof of fix

Reproduced end-to-end with **real grok in a real PTY piped into the same `@xterm/headless` build Orca ships**, then isolated the trigger:

| restore scenario | grok pinned region |
|---|---|
| `120→80→120` (transient wrong-size fit, net size unchanged) | **CORRUPTED** — box border mangled, `[1X`/`[40X` erase-patch fragments and footer text bleeding across the border (matches the screenshot) |
| `120→120` (same-size fit, no-op) | **CLEAN** |
| never resized | **CLEAN** |

So a resize that lands *back* on the original size is the worst case (grok's `autoresize` only clears when the area actually changes, so returning to a known size triggers **no clear**). The fix guarantees the same-size reveal never resizes xterm at all → no wiggle. A same-size fit was already a no-op in `safeFit`, so no needed SIGWINCH is lost; only the transient wiggle is removed.

Also confirmed deterministically via a recorded grok byte-stream replayed through xterm with the wiggle vs. a clean control (4 corrupted rows vs. 0).

## Regressions — none expected

- **`safeFit` was already a no-op on same-size**, so removing the synchronous reveal fit for unchanged panes loses no legitimate resize/SIGWINCH.
- **Genuine hidden-time resizes still fit** (pixel-changed branch → synchronous `safeFit`), preserving the existing convergence guarded by `tests/e2e/terminal-inline-tui-reveal-convergence.spec.ts` (incl. its "reveal after a hidden-time window resize" case).
- **Snapshot / SSH-reattach / appearance-change divergence still repairs** (diverged-grid branch → `requestStablePaneFit`), which forwards the *corrected* grid to the PTY (fit runs before the parked continuation flushes).
- **Focus parity** preserved (focus applied in exactly the same `isActive` cases as before).
- Every container-resizing path (divider drags, `SYNC_FIT_PANES_EVENT`, ResizeObserver, splits) still funnels through `safeFit`, which records the pixel baseline — so a genuine reveal resize can't be misclassified as "unchanged."

### Adversarial review

Ran a 3-round adversarial review loop with two independent frontier reviewers (Claude Fable + GPT‑5.6). Findings surfaced and fixed along the way: a fit↔PTY-reassert timing race, measuring the wrong (outer) element, a residual transient-commit in the stable fitter, and a snapshot-divergence repair regression. Final round: both reviewers confirm all concrete/blocking findings closed; the only remaining items are narrow, self-healing, explicitly non-blocking edges (a rare transient double-SIGWINCH on the uncommon divergence branch that self-corrects within a frame).

### Tests

- New unit coverage: `pane-reveal-fit.test.ts` (all four routing branches incl. snapshot-divergence repair + unmeasurable) and `pane-fit.test.ts` (`paneFitClientSizeChanged`: no-baseline, same-size, resized, sub-pixel jitter, inner-host shrink, unmeasurable). Manager/visibility paths assert the reveal routes through the stable fit and never the raw `fitAllPanes`.
- Full `pane-manager` + affected `terminal-pane` suites: **624 pass / 1 skipped**. `oxlint` clean. `tsc` (renderer) clean.

## Note on scope

The fix lives entirely on Orca's side (hosting inline TUIs correctly across reveal); `grok`/`grok-build` was used only to reproduce and understand the corruption, not modified.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-visibility-resume.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-visibility-resume.test.ts
@@ -40,6 +40,8 @@ type FakeManager = {
   resumeRendering: ReturnType<typeof vi.fn>
   scheduleRevealRepaint: ReturnType<typeof vi.fn>
   scheduleRevealPresent: ReturnType<typeof vi.fn>
+  fitAllPanes: ReturnType<typeof vi.fn>
+  fitAllPanesStable: ReturnType<typeof vi.fn>
 }
 
 function createManager(order: string[] = []): FakeManager {
@@ -47,7 +49,12 @@ function createManager(order: string[] = []): FakeManager {
     getPanes: vi.fn(() => []),
     resumeRendering: vi.fn(() => order.push('resume-rendering')),
     scheduleRevealRepaint: vi.fn(() => order.push('reveal-repaint')),
-    scheduleRevealPresent: vi.fn(() => order.push('reveal-present'))
+    scheduleRevealPresent: vi.fn(() => order.push('reveal-present')),
+    // Why: the reveal must fit through the wobble-resistant stable path, never
+    // the synchronous fitAllPanes, so a transient cell-metric grid is not
+    // applied and reflow-garbled onto inline TUIs on minimize→restore.
+    fitAllPanes: vi.fn(() => order.push('fit-sync')),
+    fitAllPanesStable: vi.fn(() => order.push('fit-stable'))
   }
 }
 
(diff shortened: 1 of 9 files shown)


Jaemmeong:
Confirmed on macOS with Orca v1.4.152 (relay 0.1.0+a1d6a28ba0ae) and Codex 0.145.0.

- The same machine and Codex version rendered correctly on Orca v1.4.148 (relay 46db28a33c24); the visual corruption returned immediately after updating to v1.4.152.
- With Codex actively redrawing, minimizing Orca for 2 seconds and activating it again produced duplicated and overlapping transcript/status rows on the real xterm.js canvas.
- An isolated tmux control rendered the same Codex build cleanly with both xterm-256color and screen-256color. A raw stream captured inside the Orca PTY before rendering contained no malformed or unterminated escape sequence, and DEC synchronized-output markers were balanced.
- TERM, COLORTERM, a fresh PTY, and 80x24 versus 120x40 did not explain the failure.
- A usable local workaround is forcing tui.alternate_screen="always" and disabling tui.animations. A fresh Codex process inheriting those settings stayed visually clean while actively working through the same minimize/restore cycle.

This behavior and the good-v1.4.148 / bad-v1.4.152 boundary look consistent with the reveal-fit reflow mechanism in this PR. I can retest a macOS build containing the patch. I omitted screenshots because the terminal content includes local project details.

PR #10167, opened 2026-07-23
fix(terminal): confirm running process on tab X / middle-click close

### LINKED ISSUE #10142: [Bug]: Tab close confirmation is bypassed when using the close button or middle-click
### Operating system

macOS

### Orca version

1.4.150

### Details

Short summary

Closing a tab with Cmd+W shows a confirmation warning, but closing the same tab using the tab's X button or the middle mouse button closes it immediately without showing the warning.

I am not sure whether this difference is intentional, but the behavior is inconsistent and allows mouse-based tab closing to bypass the existing close confirmation safeguard.

Steps to reproduce
Open a terminal tab in a state where closing it with Cmd+W triggers a confirmation warning.
Press Cmd+W.
Observe that a close confirmation dialog appears.
Cancel the dialog.
Click the X button on the same tab.
Observe that the tab closes immediately without confirmation.
Repeat with another tab, using the middle mouse button on the tab.
Observe that the tab again closes immediately without confirmation.
Expected behavior

All methods of closing the same tab should follow the same confirmation policy:

Cmd+W
Clicking the tab's X button
Middle-clicking the tab

If the warning is intentionally shown only for keyboard-based closing, it would be helpful to clarify the intended behavior.

Actual behavior
Cmd+W: shows a close confirmation warning.
Tab X button: closes immediately without warning.
Middle mouse button: closes immediately without warning.
Impact

The X button and middle-click can bypass a safeguard that is present when using Cmd+W. This makes it easier to accidentally close a tab, especially because middle-clicking may happen unintentionally.

Additional context

Middle-click-to-close itself appears to be an intended tab behavior. The issue is specifically that it does not use the same confirmation flow as Cmd+W.

Could you confirm whether this behavior is intentional? If not, mouse-initiated tab closes should probably go through the same close-confirmation logic as the keyboard shortcut.

innocarpe:
## Summary

- Tab strip **X** and **middle-click** previously called `closeTerminalTab` and skipped the running-process confirm that **Cmd+W** shows via `CloseTerminalDialog`.
- Add a store-driven confirm (same copy as the pane dialog) on `closeTerminalTab`, after the pin guard. Confirm re-enters with `force` so pin/process do not double-prompt.
- Last-pane Cmd+W now closes through `closeTerminalTab` so keyboard / X / middle-click share one policy. Split-pane Cmd+W still uses the existing per-pane dialog.

Fixes #10142

## Test plan

- [x] Unit: `running-terminal-close-guard`, slice, `closeTerminalTab` process confirm + force skip, kill-all force path
- [ ] Manual: start `sleep 100` (or agent) in a tab → Cmd+W shows confirm → cancel → X should show same confirm → cancel → middle-click should show same confirm
- [ ] Manual: idle shell → X / middle-click close immediately (no dialog)
- [ ] Manual: pinned tab with running process → pin dialog first; after pin confirm with force, no second process prompt
- [ ] Manual: "Don't ask again" persists `skipCloseTerminalWithRunningProcessConfirm`

## ELI5

Closing a terminal with the tab X or middle-click skipped the running-process confirm that Cmd+W shows. X, middle-click, and last-pane Cmd+W now share the same confirm so you do not kill long jobs by accident.

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -204,6 +204,7 @@ import { resolveMountedLazyModalIds, type LazyModalId } from './lazy-modal-mount
 import { translate } from '@/i18n/i18n'
 import PinnedTabCloseDialog from './components/terminal-pane/PinnedTabCloseDialog'
 import { useOsc52ClipboardDefaultOnNotice } from './components/terminal-pane/osc52-clipboard-default-on-notice'
+import RunningTerminalCloseDialog from './components/terminal-pane/RunningTerminalCloseDialog'
 import {
   hasRequestedBackgroundTerminalWorktreeMount,
   subscribeBackgroundTerminalWorktreeMountRequests
@@ -2741,6 +2742,7 @@ function App(): React.JSX.Element {
       <Toaster closeButton toastOptions={{ className: 'font-sans text-sm' }} />
       <SkillFreshnessNudge />
       <PinnedTabCloseDialog />
+      <RunningTerminalCloseDialog />
       {/* Why: Electron's drag-region hit-test is DOM-order-based (ignores z-index); render last so WindowControls stay clickable. */}
       {hasCustomTitleBar && <WindowControls />}
     </div>
--- a/src/renderer/src/components/terminal-pane/RunningTerminalCloseDialog.tsx
+++ b/src/renderer/src/components/terminal-pane/RunningTerminalCloseDialog.tsx
@@ -1,0 +1,25 @@
+import { useAppStore } from '@/store'
+import CloseTerminalDialog from './CloseTerminalDialog'
+
+/** Store-driven running-process close confirm so tab X / middle-click / Cmd+W
+ *  share one policy without threading React context through every close path. */
+export default function RunningTerminalCloseDialog(): React.JSX.Element {
+  const request = useAppStore((state) => state.runningTerminalCloseConfirm)
+  const confirmRunningTerminalClose = useAppStore((state) => state.confirmRunningTerminalClose)
+  const dismissRunningTerminalClose = useAppStore((state) => state.dismissRunningTerminalClose)
+  const updateSettings = useAppStore((state) => state.updateSettings)
+
+  return (
+    <CloseTerminalDialog
+      open={request !== null}
+      copyKind={request?.copyKind ?? 'command'}
+      onCancel={dismissRunningTerminalClose}
+      onConfirm={(dontAskAgain) => {
+        if (dontAskAgain) {
+          void updateSettings({ skipCloseTerminalWithRunningProcessConfirm: true })
+        }
+        confirmRunningTerminalClose()
+      }}
+    />
+  )
+}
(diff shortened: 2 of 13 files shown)


innocarpe:
### Sync update (`2835fdc7d`)

Follow-up: use Promise.allSettled for multi-PTY process probes so one rejected inspect cannot bypass confirmation for a sibling pane that still has a running child. All-failed probes still close (no strand).

- Commit: `2835fdc7d5a7cd4bdf6d91376a814a0434188a56`
- Upstream: https://github.com/stablyai/orca/pull/10167
- Fork portfolio: https://github.com/innocarpe/orca/pull/30

nwparker:
Thanks for this @innocarpe — and sorry for the slow turnaround.

I'd independently built #12272 for the same issue, and the two land on the same core design: move the decision into `closeTerminalTab` so every close path shares one policy, raise it through a store-driven dialog that reuses the existing `CloseTerminalDialog` (no new strings), and have `TerminalPane` delegate its last-pane close rather than probe a second time. Two people converging on that independently is a good signal the approach is right.

**Two things your PR got right that mine did not**, which I've adopted in #12272 (commit `47c02a3489`, credited to you in the PR description):

1. **The 350 ms double-activation guard** on the confirmation queue, mirroring the sibling `pinned-tab-close-confirm` slice. A queued request replaces the visible one in place, so the second click of a double-click aimed at one tab lands on the *next* tab's prompt and kills a running process the user never saw asked about — exactly the bug class this issue is about. My branch claimed this in its description but didn't actually implement it; yours did.
2. **Consulting the layout bindings for the tab's PTYs.** A mounting pane is bound into `terminalLayoutsByTabId` before `ptyIdsByTabId` catches up, so reading only the liveness map lets a close slip through that window with no prompt. Your `collectTabPtyIds` fell back to the layout when the map was empty; I've taken that idea and unioned both sources, which matches the store's own `collectPtyIdsForTab` teardown collector.

Your `Promise.allSettled` follow-up (one wedged probe must not suppress confirmation for a busy sibling pane) was the right call too — #12272 landed on the same thing independently.

**Why I'm merging #12272 rather than this one**, so the reasoning is on the record:

- **Non-interactive callers.** The guard here fires whenever `reason !== 'pty-exit' && !force`. That also catches `reason: 'cleanup'` teardowns, `rejectPinned` CLI closes, `hostCloseReason` echoes, `lifecyclePtyId` closes, and bulk "Close Others" — a modal storm in the bulk case, and worse for the callers that are answered synchronously and can't wait on a dialog at all. #12272 enumerates and tests all seven exclusions.
- **A wedged probe.** A remote inspect can hang for its full 15 s RPC timeout, leaving the X button looking dead. #12272 bounds the wait at 4 s and *asks* rather than closing, since an unanswered probe isn't evidence of an idle shell.
- **Copy selection.** `resolveTabCloseDialogCopyKind` here scans every pane on the tab, so an idle agent pane in a split words the prompt as "Stop this agent?" when the actually-busy pane is a plain command. #12272 scopes it to the PTYs that reported a child.
- **Pinned + running.** With `confirmClosePinnedTab` off, the pin dialog says nothing — so a busy pinned tab should still get asked. This PR's ordering closes it silently in that case.

None of that takes away from the diagnosis, which was correct and well ahead of mine — the plan you posted on #10142 called the fix accurately before either PR existed. Closing this in favour of #12272 with your two fixes carried over. Thank you.

nwparker:
Superseded by #12272, which merged as `ba83a71e30` and closed #10142. Your double-activation guard and layout-binding PTY lookup are both in that commit — thanks again @innocarpe.

PR #10168, opened 2026-07-23
fix(editor): preserve PDF zoom when file content reloads

### LINKED ISSUE #10165: [Bug] when PDFs are updated, the zoom resets
<kbd>[Ryder](https://discord.com/channels/1492228674081263786/1529816554106126396/1529816554106126396)</kbd> [Ryder](https://discord.com/channels/1492228674081263786/1529816554106126396/1529816554106126396)  `via Discord`

^

innocarpe:
## Summary

- Opening/updating a PDF always rebuilt the pdf.js viewer and forced `currentScaleValue = 'page-width'`, so any user zoom was lost after an on-disk update.
- Track a per-file scale preference (fit-to-width vs absolute) across content reloads; reset only when the open file path changes.
- Toolbar / shortcut zoom in·out·reset update the same preference so the next reload restores it.

Fixes #10165

## Test plan

- [x] Unit: `pdf-scale-preference` apply / step / clamp
- [ ] Manual: open a PDF → zoom in → save/overwrite the file externally → preview reloads at the same zoom
- [ ] Manual: fit-to-width then reload → stays fit-to-width
- [ ] Manual: switch to a different PDF path → zoom resets to fit-to-width

### DIFF
--- a/src/renderer/src/components/editor/PdfViewer.tsx
+++ b/src/renderer/src/components/editor/PdfViewer.tsx
@@ -17,12 +17,18 @@ import { keybindingMatchesAction } from '../../../../shared/keybindings'
 
 import workerUrl from 'pdfjs-dist/build/pdf.worker.min.mjs?url'
 import { translate } from '@/i18n/i18n'
+import {
+  applyPdfScalePreference,
+  stepPdfScalePreference,
+  type PdfScalePreference
+} from './pdf-scale-preference'
 
 pdfjsLib.GlobalWorkerOptions.workerSrc = workerUrl
 
 const MIN_SCALE = 0.25
 const MAX_SCALE = 5
 const SCALE_STEP = 1.25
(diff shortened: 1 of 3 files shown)


AmethystLiang (approved):
Reviewed the PDF zoom-preservation logic in depth, including tracing pdf.js viewer internals (`#setScale` early-returns for preset strings on empty `_pages`, but numeric scales apply directly and survive async page init — so the numeric restore path is sound and the default `page-width` path is byte-identical to prior behavior). No performance regression: the scale logic is O(1) and the pre-existing per-reload viewer rebuild is untouched.

Also folded in a small refactor deduping the six copies of the step-scale-and-record-preference logic into a shared `stepZoom` callback, with the keyboard handler reusing `zoomIn`/`zoomOut`/`zoomReset`. Lint, typecheck, and unit tests are green.

PR #10171, opened 2026-07-23
fix(source-control): allow http web scheme for ssh remotes

### LINKED ISSUE #10154: [Bug]: Review/MR web URLs forced to https:// for ssh remotes — breaks http-only self-hosted GitLab
### Operating system

Windows

### Orca version

v1.4.152

### Details

**Short summary:**
Orca builds GitLab **web/review** URLs as `https://<host>` whenever `origin` uses an `ssh://` (or scp-style `git@host:`) remote, and there is no way to override it. On a self-hosted GitLab served over **http only** (no TLS on 443), this makes "Open review page in browser", Create MR, and Generate PR details all fail. The git *transport* protocol (ssh) doesn't imply whether the *web UI* is http or https, so inferring the web scheme from it is incorrect for http-only instances.

**What happened?**
- Self-hosted GitLab, web UI served over **http only** (https/443 refused); SSH on a custom port.
- `origin` = `ssh://git@<host>:2222/<group>/<repo>.git`
- Clicking **Open review page in browser** opens `https://<host>/<group>/<repo>/-/merge_requests/new?...` → browser shows `ERR_CONNECTION_REFUSED` (443 isn't served).
- **Create MR** and **Generate PR details** fail for the same https-vs-http reason.

**How can we reproduce it?**
1. Have a self-hosted GitLab whose web UI is served over http only (443 closed).
2. Set the repo's `origin` remote over ssh, e.g. `ssh://git@<host>:2222/<group>/<repo>.git`.
3. Open the repo in Orca, make a commit, then click **Open review page in browser** (also reproducible via **Create MR** / **Generate PR details**).
4. The browser opens an `https://` URL and fails with `ERR_CONNECTION_REFUSED`.

**Root cause** (from inspecting the bundled renderer — internal names may differ from source):
- A `buildWebOrigin`-style helper returns `${protocol}//${host}` for `http:`/`https:` remotes, but **hardcodes `https://${host}`** for `ssh:` / `git:` remotes.
- The resulting `webBaseUrl` is used to build `${webBaseUrl}/-/merge_requests/new`.
- Because ssh remotes carry no web-scheme info, http-only instances always get `https://`, and there is no config key / env var to override it.

**Suggested fix** (in order of preference):
1. Add a **per-host or per-project setting** to override the web base URL (or just the scheme) for self-hosted instances.
2. When a `glab` config exists for the host, honor its `api_protocol` / host when deriving the web URL (Orca already appears to rely on `glab` for API calls).
3. At minimum, expose an env var / config key to force the web scheme to http.

The scheme is decided on two independent paths: `glab` (which *is* configurable — `glab config set -h <host> api_protocol http` fixes the API / Create-MR path) and Orca's own web-URL builder (not configurable), which is what this bug is about. The only current workaround for the review-page path is pointing `origin` at an http remote (`git remote set-url origin http://<host>/...`), but that downgrades git transport from ssh to http, so it isn't a good general fix.

innocarpe:
## Summary

- SSH/scp-style remotes always produced `https://` forge web links (`buildWebOrigin` / `parseRemoteRepo`), so **Open review page** / New MR URLs broke on **http-only self-hosted GitLab**.
- Git transport protocol is not the web UI scheme. Add `gitHostWebSchemes` (hostname → `http`|`https`), thread it into remote parsing and the manual review URL builder, and keep default **https**.
- Settings → Integrations → GitLab: multiline `host=http` editor for self-hosted overrides.

Fixes #10154

## Usage

Settings → Integrations → GitLab → **Self-hosted web URL schemes**:

```text
gitlab.company.test=http
```

Then Open review / New MR links for `ssh://git@gitlab.company.test:2222/...` remotes open `http://gitlab.company.test/...`.

## Test plan

- [x] Unit: `resolveGitHostWebScheme` / `buildWebOrigin` / `parseRemoteRepo` (ssh default, override, scp, http remote preserves scheme)
- [x] Unit: manual review URL builds `http://…/-/merge_requests/new` with override
- [x] Unit: settings text format/parse round-trip
- [ ] Manual: set `host=http`, open a branch with ssh origin, Open review page → browser hits http not https
- [ ] Manual: without override, GitHub/ssh still uses https


innocarpe:
Accidentally closed when a bad rebase helper force-pushed an empty tip. Restored and continued as https://github.com/stablyai/orca/pull/10345.

PR #10173, opened 2026-07-23
feat(app): open OS Markdown files in the floating workspace

### LINKED ISSUE #10138: [Feature]: Support opening Markdown files from the OS (Finder / Explorer)
### Problem or use case

When working with Markdown notes or documentation outside of a repository, there is currently no way to open them directly in Orca by double-clicking or using "Open With". Users have to launch Orca first and then manually locate the file.

### Proposed solution

Register Orca as a handler for Markdown files and open the selected file in the floating workspace when launched from the operating system.

This should work whether Orca is already running or not.

- Faster workflow for editing standalone Markdown files.
- Better integration with the operating system.
- Makes the floating workspace useful as a lightweight Markdown editor without requiring a full project.

innocarpe:
## Summary

- Register Orca as an **Open With** handler for `.md` / `.mdx` / `.markdown` (electron-builder `fileAssociations`).
- Queue OS open paths from:
  - cold-start `process.argv` (Windows/Linux)
  - macOS `open-file` (early, before ready)
  - second-instance `commandLine` when Orca is already running
- When the main window can receive IPC, open the **floating workspace** and each document via the existing floating markdown open path (`MarkdownDocument` + `openFile`).
- Re-enables the floating workspace setting if it was turned off, so Open With still works.

Fixes #10138

## Test plan

- [x] Unit: argv extraction (absolute markdown only, switches ignored, Windows paths)
- [x] Unit: bridge queue / flush / auth skip
- [x] Unit: second-instance passes commandLine
- [ ] Manual (packaged preferred): double-click `.md` → Orca opens floating workspace with that file
- [ ] Manual: Orca already running → Open With another `.md` → focuses Orca and opens the file
- [ ] Manual: non-markdown path ignored

## ELI5

Orca can be chosen as Open With for .md/.mdx/.markdown files. OS open (cold start, second instance, macOS open-file) opens those files in the floating workspace markdown viewer.

### DIFF
--- a/config/electron-builder.config.cjs
+++ b/config/electron-builder.config.cjs
@@ -65,6 +65,30 @@ module.exports = {
   appId,
   productName: 'Orca',
   ...(localBuildVersion ? { extraMetadata: { version: localBuildVersion } } : {}),
+  // Why: OS "Open With" / double-click markdown opens the floating workspace editor (#10138).
+  fileAssociations: [
+    {
+      ext: 'md',
+      name: 'Markdown Document',
+      description: 'Markdown Document',
+      mimeType: 'text/markdown',
+      role: 'Editor'
+    },
+    {
+      ext: 'mdx',
+      name: 'MDX Document',
+      description: 'MDX Document',
+      mimeType: 'text/mdx',
+      role: 'Editor'
+    },
+    {
+      ext: 'markdown',
+      name: 'Markdown Document',
+      description: 'Markdown Document',
+      mimeType: 'text/markdown',
+      role: 'Editor'
+    }
+  ],
   directories: {
     buildResources: 'resources/build'
   },
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -147,6 +147,7 @@ import {
 import { ensureWindowsUserDataAclGrant } from './startup/windows-user-data-acl'
 import { shouldQuitWhenAllWindowsClosed } from './startup/window-all-closed-quit-policy'
 import { createServeDesktopActivationGate } from './startup/serve-desktop-activation'
+import { createOsOpenMarkdownBridge } from './startup/os-open-markdown-bridge'
 import { RateLimitService } from './rate-limits/service'
 import { readMiniMaxSessionCookie } from './minimax/minimax-cookie-store'
 import { getInitialClaudeRateLimitTarget } from './rate-limits/claude-rate-limit-target'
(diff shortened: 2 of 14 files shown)


nwparker:
Superseded by #17906, which shipped this feature — thank you, and sorry this sat for so long.

You picked the **right target**. The floating workspace is what the issue asked for and what shipped, and you found the correct main-process primitives yourself: `isMarkdownDocumentName`, `authorizeExternalPath`, and `markdownDocumentFromFilePath(root, path, { outsideRootRelativePath: 'basename' })` — which is exactly what Orca's existing `pickFloatingMarkdownDocument` does. That's a careful read of the codebase, not a guess, and the merged main-process shape follows it. You're credited as a co-author on the merged commit.

What changed in the ~6 weeks this was open, for context rather than criticism:

- **Your `single-instance-lock.ts` widening is already on `main`.** `requestDesktopActivation(argv)` takes argv today and `acquireSingleInstanceLock(app, requestDesktopActivation)` forwards it, so that part became unnecessary — and the lambda form would now fail `serve-desktop-activation-wiring.test.ts`, which pins the exact call text.
- **`App.tsx` and `useIpcEvents.ts` no longer hold the code you edited.** The floating-workspace panel moved to `app-shell/use-floating-workspace-panel.ts` and `useIpcEvents` was split into `hooks/ipc-events/*` bridges. `src/main/index.ts` itself went from 3831 lines to 110.
- **The send → sleep 250ms → send retry** was replaced by a pull handler (`ui:consumePendingMarkdownFileOpens`), following the `SkillShareDeepLinkState` pattern main had already grown for exactly this "intent arrived before the renderer mounted" problem. That also removed the duplicate-delivery risk the retry carried.

One thing worth knowing for future work here, because it's a genuine trap and not something you could reasonably have predicted: putting `fileAssociations` at the **top level** feeds electron-builder's NSIS packager, whose `APP_ASSOCIATE` macro overwrites `Software\Classes\.md`'s default value with no backup — silently taking `.md` from whichever editor owns it, for every existing user on their next update. #17906 registers Windows Open-With by hand instead, additively, and I verified the round trip on real Windows 11.

Thanks again for the careful work.

innocarpe:
Thank you for writing this up, and for the co-author credit on #17906.

You are right that the floating workspace was the right target, and the later pull-handler / additive Windows-association notes are useful context I could not have predicted from the July codebase. Closing the fork portfolio mirror as well.

PR #10176, opened 2026-07-23
fix(workspace-board): sync Linear on context-menu Move to Status

### LINKED ISSUE #10175: [Bug] Workspace board "Move to Status" context menu doesn't sync Linear
### Problem

On the workspace Kanban board, dragging a card to a new status column correctly syncs the change to the linked Linear issue (when "Sync task status from workspace board" is enabled). However, right-clicking a card → **Move to Status** only updates the local board status and never pushes the change to Linear.

### Expected

Both drag-and-drop and the context-menu "Move to Status" should sync the new status to Linear when the setting is enabled.

### Root cause

`WorktreeContextMenu.handleAssignWorkspaceStatus` (`WorktreeContextMenu.tsx:504`) only calls the local `updateWorktreeMeta`, while the drag-drop path also calls `maybeSyncWorkspaceBoardTaskStatuses` (`WorkspaceKanbanDrawer.tsx:287,380`). The context menu never reaches the sync function — it is a shared component that receives no Linear-sync callback.

### Environment

All platforms.

ElNelyo:
## Summary

Fixes #10175 — the workspace board context-menu "Move to Status" silently drops the Linear sync that drag-and-drop performs.

## Root cause

`handleAssignWorkspaceStatus` in `WorktreeContextMenu` only wrote the local `workspaceStatus`. The Linear sync (`maybeSyncWorkspaceBoardTaskStatuses`) lived in `WorkspaceKanbanDrawer` and was unreachable from the shared menu component, which receives no sync callback. So drag-and-drop synced to Linear, but the right-click menu did not.

## Fix

Thread an `onAssignWorkspaceStatus` callback from the drawer through the kanban card chain (`LaneGrid → StatusLane → KanbanCard → WorktreeCard → WorktreeContextMenu`). The menu now funnels through the new `moveWorktreesToStatus` (batch equivalent of the existing single-worktree `moveWorktreeToStatus`), which performs the local-first write **and** the Linear sync — identical to drag-and-drop. Outside the board (sidebar list) the menu keeps its existing local-only behavior, since Linear sync is scoped to workspace-board moves.

## Screenshots

N/A — pure logic fix; no visual change. The board context menu and lanes render exactly as before.

## AI Review Report

- **Approach chosen:** prop-threading (Option 1) over calling the sync directly from the menu (Option 2). This keeps the sync orchestration, gating (`getWorkspaceBoardTaskStatusSyncRequest`), settings resolution, and toast handling in a single place (`WorkspaceKanbanDrawer`), avoiding duplication and matching the existing drag-drop path exactly.
- **Scope:** the callback is optional. The sidebar-list usage does not pass it, so its behavior is unchanged (local-only) — consistent with the fact that drag-drop (the only other sync trigger) also only exists on the board.
- **Multi-select:** the handler accepts a `worktreeIds[]` and the batch path filters out no-op (same-status) entries before writing or syncing, mirroring `dropWorktreesInStatus`.
- **Risks:** low. No change to the sync module itself, no store middleware added, no new network paths. The only behavioral change is that the board context-menu now also triggers the already-existing sync.

## Security Audit

- No secrets, tokens, or credentials introduced or logged.
- No new network calls — the Linear update goes through the existing `linearUpdateIssue` runtime client already used by drag-drop.
- Worktree ids and status ids are already-trusted store values; no untrusted input reaches the Linear API.
- No changes to auth, IPC, or filesystem paths.

## Testing

- [x] New regression tests in `WorkspaceKanbanDrawer.task-status-sync.test.tsx`:
  - syncs Linear when the board context-menu assigns a status (enabled)
  - does not sync when the setting is disabled
  - does not sync a no-op move that keeps the same status
- [x] `tsc --noEmit -p config/tsconfig.tc.web.json` — clean
- [x] `oxlint` (+ react-doctor via lint-staged) and `oxfmt` — clean
- [x] Existing sidebar test suite green (`workspace-board-task-status-sync`, `WorktreeCard.*`, `workspace-kanban.*`)

## Notes

- The fix is intentionally scoped to the workspace board. Extending Linear sync to the sidebar-list context menu would be a separate, deliberate behavior change (the setting is named "Sync task status **from workspace board**").
- No `max-lines` disables added; the prop-threading touches several files but each change is a single optional prop.

Closes #10175

### DIFF
--- a/src/renderer/src/components/sidebar/WorkspaceKanbanCard.tsx
+++ b/src/renderer/src/components/sidebar/WorkspaceKanbanCard.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Pin } from 'lucide-react'
 import { Badge } from '@/components/ui/badge'
-import type { Repo, Worktree } from '../../../../shared/types'
+import type { Repo, WorkspaceStatus, Worktree } from '../../../../shared/types'
 import WorktreeCard from './WorktreeCard'
 import { translate } from '@/i18n/i18n'
 
@@ -18,6 +18,7 @@ type WorkspaceKanbanCardProps = {
     event: React.MouseEvent<HTMLElement>,
     worktree: Worktree
   ) => readonly Worktree[]
+  onAssignWorkspaceStatus?: (worktreeIds: readonly string[], status: WorkspaceStatus) => void
 }
 
 function WorkspaceKanbanCard({
@@ -29,7 +30,8 @@ function WorkspaceKanbanCard({
   nativeDragEnabled = true,
   onActivate,
   onSelectionGesture,
-  onContextMenuSelect
+  onContextMenuSelect,
+  onAssignWorkspaceStatus
 }: WorkspaceKanbanCardProps): React.JSX.Element {
   const contextWorktrees =
     isSelected && selectedWorktrees && selectedWorktrees.length > 0 ? selectedWorktrees : undefined
@@ -61,6 +63,7 @@ function WorkspaceKanbanCard({
         onActivate={onActivate}
         onSelectionGesture={onSelectionGesture}
         onContextMenuSelect={(event) => onContextMenuSelect(event, worktree)}
+        onAssignWorkspaceStatus={onAssignWorkspaceStatus}
       />
     </div>
   )
--- a/src/renderer/src/components/sidebar/WorkspaceKanbanDrawer.task-status-sync.test.tsx
+++ b/src/renderer/src/components/sidebar/WorkspaceKanbanDrawer.task-status-sync.test.tsx
@@ -31,7 +31,8 @@ const {
   toastErrorMock,
   toastWarningMock,
   pointerDragState,
-  documentDropState
+  documentDropState,
+  laneAssignStatusState
 } = vi.hoisted(() => ({
   syncWorkspaceBoardTaskStatusesMock: vi.fn(() =>
     Promise.resolve({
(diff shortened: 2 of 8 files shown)


AmethystLiang (approved):
Reviewed via a 11-persona parallel code review (correctness, adversarial, reliability, performance, api-contract, security-surface, maintainability, testing, project-standards, agent-native, previous-comments) plus independent validation.

Verdict: zero functionality regression, zero performance regression.
- Behavior parity: `moveWorktreesToStatus` faithfully mirrors the existing `moveWorktreeToStatus`/`dropWorktreesInStatus` (no-op filter, recordFeatureInteraction, batch write, sync on changed ids).
- Selection correctness: `activeContextWorktrees` is captured in the render closure; the synchronous `setMenuOpenState(false)` can't move the wrong worktrees.
- Performance: the new callback shares the exact useCallback dep profile of the already-threaded drop callback; every hop forwards the prop directly (no new inline closures); sidebar path leaves it undefined/stable — no new re-render surface.
- Contract: `WorkspaceStatus = string`, so the handoff is type-safe; all 6 prop declarations identical/optional and backward-compatible.

Added regression coverage (2c7e64a): extracted the context-menu routing into a pure `planWorkspaceStatusAssignment` helper and unit-tested the board-sync vs sidebar local-only branch — the exact path #10175 regressed on — since the original drawer tests exercised the sync wiring via a mocked LaneGrid but never the menu's routing branch.

PR #10180, opened 2026-07-23
fix(main): route Codex + PostHog through electron net.fetch

### LINKED ISSUE #10117: [Bug]: Codex and telemetry still use crash-prone global fetch after #9142
### Operating system

macOS

### Orca version

1.4.152

## Summary

Orca 1.4.152 can still crash the entire Electron main process with the bundled Undici paused-parser assertion while Codex is in use, after #8695 was closed by #9142.

## Environment

- Orca: 1.4.152
- Electron: 43.1.0
- Node.js: 24.18.0
- Bundled Undici: 7.28.0
- OS: macOS

## Actual result

A JavaScript error occurred in the main process:

```text
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

assert(!this.paused)

at Parser.finish (node:internal/deps/undici/undici:7380:9)
at TLSSocket.onHttpSocketEnd (node:internal/deps/undici/undici:7819:34)
```

## Regression evidence

#9142 cancels unread error-response bodies, but 1.4.152 still reproduced the same crash. The application currently has three direct main-process global `fetch` calls for:

- `https://chatgpt.com/backend-api/wham/usage`
- `https://chatgpt.com/backend-api/wham/rate-limit-reset-credits`
- `https://chatgpt.com/backend-api/wham/rate-limit-reset-credits/consume`

The official-build PostHog client also uses its default Node fetch transport.

This matches nodejs/undici#5360: body backpressure leaves the HTTP parser paused, the peer closes a non-keepalive socket, and `onHttpSocketEnd()` calls `parser.finish()`, which asserts that the parser is not paused. The deterministic 64 KiB `Connection: close` reproduction from #8695 still fails in Orca's embedded runtime; nodejs/undici#5474 is the upstream runtime fix.

## Validated mitigation

I patched only those known Electron-main trigger paths:

- replaced the three Codex backend global fetch calls with `electron.net.fetch`
- supplied `(url, init) => electron.net.fetch(url, init)` as the PostHog transport

The patched 1.4.152 build then stayed running with its desktop window available and `orca status --json` reporting both runtime and graph as `ready` while Codex usage/rate-limit display worked.

## Proposed scope

Route only these known Electron-main request producers through Chromium's network stack. Preserve headers, abort signals, response parsing, body cancellation, telemetry consent, batching, and queue behavior.

This is a targeted application mitigation, not a replacement for upgrading to a runtime containing nodejs/undici#5474. Raw `ELECTRON_RUN_AS_NODE` reproduction is expected to keep failing until the bundled runtime is updated.

innocarpe:
## Summary

- Replace the three remaining Electron-main Codex `wham/*` global `fetch` calls with `electron.net.fetch` so usage / rate-limit-reset-credits / consume no longer hit Node undici.
- Point official-build PostHog transport at `net.fetch` via the SDK `fetch` option.
- Drop `codex-fetcher.ts` from the global-fetch audit (it is no longer a bare-fetch call site).

Fixes #10117 (validated mitigation from the issue: undici `assert(!this.paused)` main-process crash while Codex is in use after #9142).

## Why

#9142 cancels unread error bodies, but Orca 1.4.152 still crashed with the bundled undici paused-parser assertion. Chromium's network stack does not share that failure mode. This is an application mitigation until Electron ships undici#5474 — not a full removal of every global fetch elsewhere.

## Test plan

- [x] `vitest` codex-fetcher + backend + global-fetch audit + telemetry client
- [x] `pnpm typecheck:node`
- [ ] Manual: Codex usage / rate-limit display still updates; PostHog events still flush on official builds

## ELI5

Some Codex and PostHog network calls in the main process used Node fetch and could crash Electron under load. They now use electron.net.fetch, which is the safer path for main-process HTTP.

### DIFF
--- a/src/main/global-fetch-call-site-audit.test.ts
+++ b/src/main/global-fetch-call-site-audit.test.ts
@@ -20,7 +20,7 @@ const AUDITED_GLOBAL_FETCH_LINES = new Map<string, number>([
   ['main/gitea/client.ts', 1],
   ['main/orca-profiles/profile-cloud-client.ts', 1],
   ['main/orca-profiles/profile-cloud-org-members-client.ts', 1],
-  ['main/rate-limits/codex-fetcher.ts', 3],
+  // codex-fetcher uses electron net.fetch (not global fetch) — see #10117
   ['main/runtime/relay/relay-http-client.ts', 2],
   ['main/source-control/hosted-review-api-request.ts', 1],
   ['main/speech/openai-transcription-client.ts', 1],
--- a/src/main/rate-limits/codex-fetcher-backend.test.ts
+++ b/src/main/rate-limits/codex-fetcher-backend.test.ts
@@ -2,15 +2,21 @@ import { join } from 'node:path'
 import { cancelTrackingResponse } from '../lib/unread-response-body.test-fixtures'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
-const { childSpawnMock, readFileMock, ptySpawnMock } = vi.hoisted(() => ({
+const { childSpawnMock, readFileMock, ptySpawnMock, netFetchMock } = vi.hoisted(() => ({
   childSpawnMock: vi.fn(),
   readFileMock: vi.fn(),
-  ptySpawnMock: vi.fn()
+  ptySpawnMock: vi.fn(),
+  netFetchMock: vi.fn()
 }))
 
 vi.mock('node:child_process', () => ({ spawn: childSpawnMock }))
 vi.mock('node:fs/promises', () => ({ readFile: readFileMock }))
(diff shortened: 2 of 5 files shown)


jinhoda:
@brennanb2025  This issue is currently blocking me from using Orca, so I’d really appreciate it if this PR could be prioritized for review. I’ve tested this change, and it resolves #10117 for me. Thanks!

innocarpe:
### Sync update (`6720e84097`)

Rebase onto latest upstream/main: keep net.fetch for Codex wham + PostHog; re-point post-main backend tests to netFetchMock.

- Commit: `6720e8409753f51eb20d7c1db39716848c880304`
- Upstream: https://github.com/stablyai/orca/pull/10180
- Fork portfolio: https://github.com/innocarpe/orca/pull/37

PR #10184, opened 2026-07-23
fix(tasks): select only project when clicking from All projects

### LINKED ISSUE #10182: [Bug]: Project selector cannot directly filter Tasks to one project
## Operating system

macOS

## Orca version

1.4.146

### Short summary

The Tasks page project selector cannot directly narrow the task list to one project when starting from All projects. Clicking a project removes that project from the active set instead of showing only that project tasks.

### What happened?

The Tasks page initially showed work items from 7 projects. I opened the project selector and clicked dailytarot.

Instead of filtering the list to dailytarot, Orca changed the active set from 7 projects to 6 and removed the dailytarot issues. The selector then summarized the remaining projects, for example company-os, CreatorMade +4.

The All projects action only restores every project. There is no discoverable way to clear the set or choose Only this project, so filtering to a single project requires manually excluding every other project.

### How can we reproduce it?

1. Configure more than one GitHub-backed project in Orca.
2. Open Tasks.
3. Confirm the project selector shows All projects.
4. Open the project selector.
5. Click one project, such as dailytarot.
6. Observe that the clicked project is excluded and the project count decreases, rather than the task list showing only that project.

### Expected behavior

Provide a direct, discoverable way to filter Tasks to one project. For example, clicking a project from All projects could select only that project, or each project could expose an explicit Only action, or the selector could provide Clear all followed by selection.

### Additional context

The Tasks header also displayed Local Mac source tool unavailable, while still loading 7 projects and their issues. The project-selection behavior above was reproducible with the loaded data.

Originating user report: https://github.com/zerator01/company-os/issues/1286

innocarpe:
## Summary

- From **All projects**, clicking a project filters to **only that project** instead of excluding it.
- Partial multi-select keeps add/remove behavior and still refuses to deselect the last project.
- The selected execution host is preserved when narrowing a multi-host project.

Fixes #10182.

## ELI5

When everything was selected, clicking one project removed it. Now that click keeps only that project, while normal multi-selection still works.

## Takeover / attribution

Originally implemented by @innocarpe. This takeover re-validated the existing commits without changing their authorship or force-pushing the branch.

## Fix proof

- Deterministic regression test: `src/renderer/src/components/task-project-source-combobox-model.test.ts` covers All → only, partial add/remove, last-project guard, multi-host removal, and preservation of a non-primary SSH source.
- `pnpm vitest run src/renderer/src/components/task-project-source-combobox-model.test.ts --config config/vitest.config.ts` — 1 file, 7 tests passed.

## No-regression proof / trade-offs

- Cross-platform, SSH, and folder-workspace review: selection operates on repository IDs only; the test covers preserving a selected SSH/non-primary source, and folder-workspace eligibility is unchanged.
- Performance: only an explicit user click gets the additional all-selected group scan; no polling, I/O, or render-loop work was added. The affected resource is CPU on click, with deterministic work bounded by project groups plus sources.
- `0 regressions / risk`: **UNKNOWN** — focused model coverage is clean, but no Electron surface was attached in this session for manual UI proof.

## Review loop

- Loop 1: independent review, verdict **clean**; verified group-based All selection, SSH-source preservation, partial selection, and last-project guard.

## Validation

- `pnpm vitest run src/renderer/src/components/task-project-source-combobox-model.test.ts --config config/vitest.config.ts` — passed (7/7).
- `pnpm exec oxlint src/renderer/src/components/task-project-source-combobox-model.ts src/renderer/src/components/task-project-source-combobox.tsx src/renderer/src/components/task-project-source-combobox-model.test.ts` — passed.
- `git diff --check` and `git diff origin/main...HEAD --check` — passed.

## UI evidence

No screenshot: the in-app browser/Electron surface list was empty for this dispatched session.

### DIFF
--- a/src/renderer/src/components/task-project-source-combobox-model.test.ts
+++ b/src/renderer/src/components/task-project-source-combobox-model.test.ts
@@ -73,7 +73,7 @@ describe('nextTaskProjectSelectionAfterToggle', () => {
     const multi = group('proj', [local, remote])
     // Three groups so deselecting multi is not the "all selected → only this" path.
     const groupsWithMulti = [multi, group('other', [b]), group('c', [c])]
-    const selected = new Set(['proj-local', 'b'])
+    const selected = new Set(['proj-local', 'proj-ssh', 'b'])
     const next = nextTaskProjectSelectionAfterToggle({
       groups: groupsWithMulti,
       selected,
@@ -82,7 +82,7 @@ describe('nextTaskProjectSelectionAfterToggle', () => {
     expect(next).toEqual(new Set(['b']))
   })
 
-  it('from All projects with multi-host groups, click keeps the group primary source', () => {
+  it('from All projects with multi-host groups, click keeps the selected source', () => {
     const local = repo('proj-local', 'proj')
     const remote = repo('proj-ssh', 'proj')
     const multi = group('proj', [local, remote])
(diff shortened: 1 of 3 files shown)


innocarpe:
### Sync update (`ad3c1c644`)

Rebase onto latest upstream/main (cleanup batch).

- Commit: `ad3c1c644751f94e961e9a0a9fcb70dca9707e01`
- Upstream: https://github.com/stablyai/orca/pull/10184
- Fork portfolio: https://github.com/innocarpe/orca/pull/40

PR #10217, opened 2026-07-23
feat(diagnostics): census renderer heap collections at high-water so OOMs self-attribute

OrcaWin:
## Summary

Production Windows renderer OOMs report the JS heap **pegged at the 3586 MB V8 ceiling for hours** before the process dies — but the crash bundle only carries heap *totals* (`renderer_memory: usedHeapMB/totalHeapMB/heapLimitMB`), never an object breakdown. So the leak is provable yet unattributable: you can see the heap is full, not *what* filled it.

This adds a cheap, privacy-safe **heap census** that runs when the renderer heap crosses 85% of its limit, recording the entry *counts* of the renderer's long-lived collections into the diagnostic bundle. The next OOM report then names the collection that grew (e.g. `store.someMapByPaneKey=412031`, `domNodes=1_900_000`) instead of leaving us to guess.

### What it captures
- `domNodes` — attached DOM node count (catches retained editor/terminal DOM subtrees).
- `storeCollectionsTotalEntries` — sum across every store collection.
- The largest store collections (`store.<field>`), generically over the Zustand state so a *future* leaking map self-identifies too — reported only when ≥ 100 entries, capped at the 20 largest so one census can't bloat the very buffer it diagnoses.

**Counts only — never keys, values, or contents.** No PII; no cross-origin-isolation requirement (so it works where `measureUserAgentSpecificMemory()` doesn't).

### How it's gated (cheap by default)
`shouldRecordHeapCensus` gates on `usedJSHeapSize / jsHeapSizeLimit ≥ 0.85` **and** ≥ 5 min since the last census. Under normal memory the census never runs; a pegged renderer emits it periodically, not every 60 s sample. The census module is imported **dynamically** only when a high-water sample fires, so its store dependency stays off the `crash-diagnostics` leaf module's hot import chain (terminal modules import that file).

## Fix evidence
- New `renderer-heap-census.test.ts` (pure `censusFromStoreState`): counts Maps/Sets/arrays/plain-records, ignores functions/scalars/class-instances, threshold + top-N cap, always emits `domNodes` + total. `shouldRecordHeapCensus` cases in `crash-diagnostics.test.ts`: below high-water → off; at high-water past throttle → on; within throttle → off; missing heap fields → off.
- 14 tests pass; `oxlint` clean; web `tsc --noEmit` exit 0.

## ELI5
When the app runs out of memory we currently get "the box is full" but not "the box is full of *these*." This adds a quick headcount of the app's biggest internal lists — taken only when memory is nearly full — so the next crash report tells us which list ballooned. It's just counts, no private data, and it does nothing until memory is already dangerously high.

## Trade-offs
- Does **not** capture detached DOM or non-store JS objects — `getElementsByTagName('*')` counts attached nodes only. It catches the store-collection and attached-DOM/editor/terminal classes; a full heap snapshot (bigger, PII-laden, perf-heavy) is deliberately out of scope. Called out as a known limitation.
- One extra full-DOM + full-store iteration per census — bounded by the 5-min throttle and only at high-water. Negligible vs. the OOM it diagnoses.
- Purely additive diagnostics: no product behavior, API, persistence, SSH, or provider change.

## Context
This is the observability half of an investigation into a v1.4.145+ renderer-heap OOM regression (heap pegged at ceiling, reproducible per-user). Static analysis ruled the leak out of the store maps, github/PR caches, subscriptions, and native-chat transcript paths (all bounded/leak-tested) but couldn't attribute it without an object breakdown — which the reports lack. This makes the next occurrence self-attributing.

## Adversarial review
- **Round 1:** essentially clean, no hard bugs. Notes addressed: corrected a wrong clock comment (`performance.now`, not `Date.now`); made the first high-water census fire immediately (was suppressed for the first 5 min of a renderer load); documented the detached-DOM/non-store limitation in code and here; added boundary + integration test coverage.
- **Round 2:** re-review of the fixes → **CLEAN**, no regressions.
- **Loops until clean: 2.** 18 tests pass.

### DIFF
--- a/src/renderer/src/lib/crash-diagnostics.test.ts
+++ b/src/renderer/src/lib/crash-diagnostics.test.ts
@@ -172,4 +172,42 @@ describe('renderer crash diagnostics', () => {
       diagnostics.recordRendererCrashBreadcrumb('renderer_bootstrap_started')
     ).not.toThrow()
   })
+
+  describe('shouldRecordHeapCensus', () => {
+    const mem = (used: number, limit: number) => ({
+      usedJSHeapSize: used,
+      jsHeapSizeLimit: limit
+    })
+
+    it('is false below the high-water fraction', () => {
+      expect(diagnostics.shouldRecordHeapCensus(mem(50, 100), 10_000_000, 0)).toBe(false)
+    })
+
+    it('fires the first census immediately at high-water (lastCensusAt === 0)', () => {
+      expect(diagnostics.shouldRecordHeapCensus(mem(90, 100), 10_000, 0)).toBe(true)
+    })
+
+    it('fires exactly at the high-water fraction', () => {
+      // 0.85 is not below 0.85, so it proceeds.
+      expect(diagnostics.shouldRecordHeapCensus(mem(85, 100), 10_000_000, 0)).toBe(true)
+    })
+
+    it('re-fires once the throttle interval has elapsed after a prior census', () => {
+      const now = 10_000_000
+      expect(diagnostics.shouldRecordHeapCensus(mem(90, 100), now, now - 5 * 60_000)).toBe(true)
+    })
+
+    it('throttles repeat censuses within the interval', () => {
+      const now = 10_000_000
+      expect(diagnostics.shouldRecordHeapCensus(mem(90, 100), now, now - 60_000)).toBe(false)
+    })
+
+    it('is false when heap fields are missing', () => {
+      expect(diagnostics.shouldRecordHeapCensus({}, 10_000_000, 0)).toBe(false)
+    })
+
+    it('is false when the limit is zero (no divide-by-zero)', () => {
+      expect(diagnostics.shouldRecordHeapCensus(mem(0, 0), 10_000_000, 0)).toBe(false)
+    })
+  })
 })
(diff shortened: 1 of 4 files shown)


OrcaWin:
Closing as redundant — this duplicates instrumentation that just merged to main.

`src/renderer/src/lib/renderer-memory-profile.ts` + `recordRendererMemoryHighwater` (in `crash-diagnostics.ts`) already ship the same capability, and a strict superset of it:
- A bounded contributor registry (`registerRendererMemoryProfileContributor`) instead of a single hard-coded reader.
- A registered `store` contributor that calls `summarizeStateCollectionSizes(getState(), 20)` — the same generic "name the fattest store slices" approach this PR took.
- Plus `domNodes`, `.xterm` `terminalElements`, browser-webview counts, `0.6`/`0.8` one-shot thresholds (earlier than this PR's `0.85`), and a `dashboard-popout` surface.

My branch was cut from a stale local `main` (173 commits behind), so I reinvented shipped work. The upstream version is better-architected and already deployed, so there's nothing to salvage here. The v1.4.145+ renderer-OOM leak this was meant to diagnose will now self-attribute via `renderer_memory_highwater` on the next occurrence.

PR #10220, opened 2026-07-23
feat(rate-limits): add Gemini 3.6 Flash and 3.5 Flash-Lite bucket names

heyramzi:
Adds bucket-name mappings for the new GA Gemini models so the rate-limit UI labels them correctly instead of falling back to the humanized model id.

Additive only: the existing 2.5 / 3.1 entries stay in place so historical usage buckets keep their labels.


OrcaWin:
Thanks for the contribution. Current main already humanizes these IDs as `3.6 Flash`, `3.5 Flash`, and `3.5 Flash Lite`, so this doesn’t change the displayed labels. Adding them to the registry instead changes duplicate-bucket selection, which isn’t part of the reported need, and we couldn’t verify these exact IDs in a current `retrieveUserQuota` payload. Closing as unnecessary for now; happy to revisit with a sanitized payload that demonstrates a real failure.

PR #10221, opened 2026-07-23
fix(codex): keep shell-profile-only Windows CODEX_HOME on the managed lane

### LINKED ISSUE #9788: Windows: shell-profile-only CODEX_HOME is undetectable — real ~/.codex gets hooks/trust while codex runs elsewhere
Found during v1.4.149 pre-stable validation (F4 variant B), reproduced on Windows 11 / Orca 1.4.149-rc.0 / codex-cli 0.144.6.

## Repro
1. No CODEX_HOME user/system env var. Add `$env:CODEX_HOME = "C:\custom-codex"` to PowerShell `$PROFILE` only.
2. Launch Orca from Start menu (does not source $PROFILE), Codex account = System Default.
3. Before even opening a terminal, Orca's real-home lane engages: `~/.codex/hooks.json` gains all 6 Orca hook events and `config.toml` gains 6 `[hooks.state.*]` trust blocks — while the user's actual codex (launched from their shell) runs against `C:\custom-codex`, which has no Orca hook.

## Root cause
`src/main/pty/shell-startup-env.ts` `readShellStartupEnvVar` returns `undefined` on win32 (PowerShell profile parsing explicitly out of scope) and only scans POSIX rc files, so `isHostSystemDefaultRealHomeSelected` cannot see a shell-only custom home on Windows and fails open into the real-home lane.

## Consequences
- Orca writes hooks/trust into a real `~/.codex` the user deliberately doesn't use
- Split-brain: status hook lives in the real home, codex runs in the custom home → agent status broken for this setup

## Related
- Same fail-open class exists on POSIX when the shell probe *errors* (slow/nonstandard rc) — untested but same code path; a fix should cover both: treat probe failure/unsupported-platform as "unknown" and prefer the managed lane, or add $PROFILE scanning on win32.
- Custom-CODEX_HOME handling is Bucket B (explicitly deferred) in the real-home rollout runbook; this issue is its Windows-specific sharp edge.

Full validation transcript available internally (F4/F5 report, 2026-07-21).

rodboev:
## Summary

- Keep Windows System Default Codex on Orca's managed lane when Orca cannot inspect shell-startup `CODEX_HOME`, instead of failing open into real `~/.codex`.
- Root cause: `isHostSystemDefaultRealHomeSelected` used the POSIX-only startup-env probe as if `undefined` meant “no custom home”. On win32 the probe is unsupported, so the caller incorrectly treated “cannot inspect” as “confirmed absent”.
- Add an explicit startup-probe-support discriminator, short-circuit the unsupported Windows path to the managed lane, and keep the existing POSIX behavior unchanged.

Closes #9788.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Focused local validation:

- `pnpm exec vitest run --config config/vitest.config.ts src/main/codex-accounts/runtime-home-service.test.ts` - passed, 1 file and 103 tests.
- `pnpm exec vitest run --config config/vitest.config.ts src/main/pty/shell-startup-env.test.ts` - passed, 1 file and 33 tests.
- `pnpm exec vitest run --config config/vitest.config.ts src/main/codex-accounts/runtime-home-service.test.ts src/main/pty/shell-startup-env.test.ts` - passed, 2 files and 136 tests.
- `pnpm run typecheck:node` - passed, `tsc --noEmit -p config/tsconfig.node.json`.
- `pnpm exec oxlint src/main/codex-accounts/runtime-home-service.ts src/main/codex-accounts/runtime-home-service.test.ts src/main/pty/shell-startup-env.ts src/main/pty/shell-startup-env.test.ts` - passed with no diagnostics.
- `pnpm exec oxfmt --check src/main/codex-accounts/runtime-home-service.ts src/main/codex-accounts/runtime-home-service.test.ts src/main/pty/shell-startup-env.ts src/main/pty/shell-startup-env.test.ts` - passed, all four files formatted.

## AI Review Report

Adversarial review covered the Windows, macOS, and Linux launch-path matrix for the new startup-probe discriminator and the host System Default real-home predicate. The review confirmed that the unsupported Windows probe now stays on Orca's managed lane, while managed-account selection, explicit launch-env `CODEX_HOME`, WSL behavior, and the existing POSIX startup-file behavior remain unchanged. No PTY, SSH, UI, or renderer behavior changed in this slice.

## Security Audit

This change adds no new shell parsing, subprocess, filesystem write, auth, or IPC surface. It removes an unintended write path by stopping Orca from mutating real `~/.codex` when it cannot prove that is the home the CLI actually uses.

## Notes

Scope is limited to the unsupported Windows startup-probe case. PowerShell profile parsing stays out of scope in this slice.

## ELI5

On Windows, System Default Codex incorrectly fell open into real ~/.codex when Orca cannot read shell-startup CODEX_HOME. Unsupported probes no longer mean no custom home, so managed accounts stay on the managed lane.

### DIFF
--- a/src/main/codex-accounts/runtime-home-service.test.ts
+++ b/src/main/codex-accounts/runtime-home-service.test.ts
@@ -16,6 +16,7 @@ import {
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import type { GlobalSettings } from '../../shared/types'
+import type * as ShellStartupEnv from '../pty/shell-startup-env'
 
 const testState = {
   userDataDir: '',
@@ -210,18 +211,6 @@ function normalizeLinkTarget(linkTarget: string): string {
     : linkTarget
 }
 
-function withPlatform<T>(platform: NodeJS.Platform, callback: () => T): T {
-  const originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform')
-  Object.defineProperty(process, 'platform', { configurable: true, value: platform })
-  try {
-    return callback()
-  } finally {
-    if (originalPlatform) {
-      Object.defineProperty(process, 'platform', originalPlatform)
-    }
-  }
-}
-
 function expectResourceLinkedOrCopied(targetPath: string, sourcePath: string): void {
   expect(existsSync(targetPath)).toBe(true)
   if (!lstatSync(targetPath).isSymbolicLink()) {
@@ -292,6 +281,10 @@ describe('CodexRuntimeHomeService', () => {
   beforeEach(() => {
     vi.resetModules()
     vi.clearAllMocks()
+    vi.doMock('../pty/shell-startup-env', async () => ({
+      ...(await vi.importActual<typeof ShellStartupEnv>('../pty/shell-startup-env')),
+      isShellStartupEnvProbeSupported: () => true
+    }))
     testState.userDataDir = mkdtempSync(join(tmpdir(), 'orca-runtime-home-'))
     testState.fakeHomeDir = mkdtempSync(join(tmpdir(), 'orca-codex-home-'))
     testState.previousUserDataPath = process.env.ORCA_USER_DATA_PATH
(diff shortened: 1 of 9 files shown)


OrcaWin:
Taking a look!

OrcaWin:
Windows/WSL verification on head `c75bb5b0c54f63c02bc31439c09061c5ccbdc76c`:

- **Approach:** sound. The fail-closed guard at `src/main/codex-accounts/runtime-home-service.ts:378` is appropriate because the POSIX startup-file probe cannot safely determine PowerShell profile exports on Windows.
- **Real Windows 11 / Electron A/B:** with no parent `CODEX_HOME` and a real PowerShell profile exporting a sentinel path, the current source loaded the profile (`profileLoaded=1`) but the terminal received the isolated `...\codex-runtime-home\home` in both `CODEX_HOME` and `ORCA_CODEX_HOME`; managed hooks existed and real-home hooks did not. I removed only the guard, rebuilt, and—holding the unrelated trust-lane gate open so its fallback could not mask classification—the identical app run received the profile sentinel as `CODEX_HOME` with `ORCA_CODEX_HOME=null`, failing the managed-home assertion. Restoring the guard made the same run pass; a production-gate positive run also passed.
- **Real WSL2:** executed Ubuntu on kernel `6.18.33.2-microsoft-standard-WSL2`. With trap executables first on `PATH`, the positive control invoked `wsl.exe`/`bash.exe` (exit 86 plus marker), while the two platform-contract tests passed without a marker, proving those tests do not invoke ambient WSL/bash.
- **Automated validation:** 180 focused Vitest tests passed (including the existing Windows managed-home/ENOTEMPTY lifecycle coverage), plus `typecheck:node`, focused oxlint, and focused oxfmt. The PR regressions are non-vacuous: removing the runtime guard fails the Windows managed-lane test (`runtime-home-service.test.ts:1121`); forcing probe support true fails the Windows contract; forcing it false fails the Darwin/Linux contract (`shell-startup-env.test.ts:96,103`). The exact interactive managed-account OAuth/login flow was not exercised, so its prior ENOTEMPTY failure mode remains a residual end-to-end gap.
- **Review/scope:** CodeRabbit's one actionable Darwin-platform comment is addressed in `c75bb5b0c`. I declined its generic docstring-coverage warning because these small exports/test helpers are self-describing and adding docstrings solely for that metric would conflict with the repository's concise non-obvious-comment rule. No scope alerts found.

**Status: NEEDS-DECISION.** GitHub reports this PR `CONFLICTING` / `DIRTY` against current main and exposes no status checks. Per instruction I did not rebase or merge; no source changes remained to push, so the head is unchanged.

PR #10231, opened 2026-07-23
fix(terminal): preserve remote alt-screen scrollback on restore

### LINKED ISSUE #6106: SSH terminal loses pre-TUI shell output after Codex tab restore
## Summary

In an SSH-backed terminal, shell output printed before starting the Codex TUI can disappear after switching away from the Codex tab and then switching back. This also affects ordinary commands such as `ls`, so it is not specific to GitHub auth output.

## Reproduction

Verified on `origin/main` using a localhost SSH target.

1. Launch Orca from `origin/main`.
2. Add/connect an SSH target to the same machine, e.g. `ssh localhost`.
3. Open a remote worktree terminal through that SSH target.
4. In a fresh SSH terminal, run a visible shell command, for example:

```bash
printf "PRE_CODEX_START\n"
ls | head -5
printf "PRE_CODEX_END\n"
```

5. Start the real Codex TUI in the same terminal:

```bash
codex
```

6. Create or select another terminal tab.
7. Switch back to the Codex terminal tab.

## Expected

The pre-Codex shell output should remain available in the terminal buffer/scrollback, or at least should not be discarded merely because the tab was hidden and restored while Codex was active.

## Actual

After switching away and back, the terminal restore/snapshot contains the current Codex TUI screen, but the pre-Codex shell output is gone. In a reproduced run, the marker and `ls` output existed before and during Codex startup, but after tab restore the snapshot no longer contained the marker or `AGENTS.md` from `ls | head -5`.

## Notes

This reproduces on `origin/main`, so it is separate from the SSH replay artifact fix in PR #6043. The likely area is the hidden terminal restore / main-buffer snapshot path when a TUI redraws the current screen after earlier shell scrollback.

Also observed: `no oauth token found for github.com` may appear in the SSH localhost environment when `gh auth token` fails, but that is a separate auth/environment issue. The disappearing-output behavior reproduces with plain `ls` output too.

## Environment

- Target: SSH localhost
- Command sequence: shell output -> `codex` TUI -> switch tabs away/back
- Main verification: `origin/main` dev instance with isolated user data

rodboev:
## Summary

- Preserve pre-TUI shell scrollback when a remote alternate-screen terminal is hidden and restored.
- Root cause: the requested remote snapshot path folded scrollback into `data` and dropped the alt-screen metadata that the renderer restore path already uses locally.
- Carry requested-snapshot alt-screen metadata through the remote snapshot frame, re-split the folded payload on the multiplexer boundary, and let the renderer take the same scrollback-preserving branch it already uses for local restores. Keep subscribe and mobile snapshot frames unchanged, and preserve the local alt-screen contract when remote scrollback is unavailable.

Closes #6106.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Focused validation to run:

- `pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/rpc/terminal-multiplex-escape-tail.test.ts`
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/runtime/remote-runtime-terminal-frame-drop-resync.test.ts`
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/pty-connection.test.ts`
- `pnpm run typecheck:node`
- `pnpm run typecheck:web`

Focused validation completed:

- `pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/rpc/terminal-multiplex-escape-tail.test.ts`: passed, `Test Files  1 passed (1)`, `Tests  2 passed (2)`.
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/runtime/remote-runtime-terminal-frame-drop-resync.test.ts`: passed, `Test Files  1 passed (1)`, `Tests  15 passed (15)`.
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/pty-connection.test.ts`: passed, `Test Files  1 passed (1)`, `Tests  471 passed (471)`.
- `pnpm run typecheck:node`: passed, `tsc --noEmit -p config/tsconfig.node.json`.
- `pnpm run typecheck:web`: passed, `tsc --noEmit -p config/tsconfig.tc.web.json`.
- `pnpm exec oxlint src/main/runtime/rpc/methods/terminal.ts src/renderer/src/runtime/remote-runtime-terminal-multiplexer.ts src/renderer/src/components/terminal-pane/pty-connection.test.ts src/main/runtime/rpc/terminal-multiplex-escape-tail.test.ts src/renderer/src/runtime/remote-runtime-terminal-frame-drop-resync.test.ts`: passed, no diagnostics.
- `pnpm exec oxfmt --check src/main/runtime/rpc/methods/terminal.ts src/renderer/src/runtime/remote-runtime-terminal-multiplexer.ts src/renderer/src/components/terminal-pane/pty-connection.test.ts src/main/runtime/rpc/terminal-multiplex-escape-tail.test.ts src/renderer/src/runtime/remote-runtime-terminal-frame-drop-resync.test.ts`: passed, `All matched files use the correct format.`
- Dependency setup: `pnpm install --frozen-lockfile` completed successfully before validation.

The generic `pnpm lint`, `pnpm typecheck`, `pnpm test`, and `pnpm build` commands were not run; their template boxes remain unchecked. The targeted changed-file checks above passed.

## Security Audit

This change adds no new auth, filesystem, network, or command surface. It threads metadata through an existing remote terminal snapshot frame and re-splits it in the renderer transport.

## Notes

Scope is limited to remote snapshot metadata parity. It does not widen into the broader visible-TUI-output work.

## ELI5

Hiding and restoring a remote terminal running a full-screen TUI could drop the shell scrollback from before the TUI. Remote snapshots now carry alt-screen metadata so restore keeps that history like local does.

### DIFF
--- a/src/main/runtime/rpc/methods/terminal.ts
+++ b/src/main/runtime/rpc/methods/terminal.ts
@@ -668,11 +668,9 @@ function sendSnapshotFrames(
       source: options.source,
       oscLinks: options.oscLinks,
       pendingEscapeTailAnsi: options.pendingEscapeTailAnsi,
-      ...(options.alternateScreen
-        ? {
-            alternateScreen: true,
-            scrollbackAnsiLength: options.scrollbackAnsiLength ?? 0
-          }
+      ...(options.alternateScreen === true ? { alternateScreen: true } : {}),
+      ...(typeof options.scrollbackAnsiLength === 'number'
+        ? { scrollbackAnsiLength: options.scrollbackAnsiLength }
         : {}),
       truncated: options.truncated === true,
       truncatedByByteBudget: options.truncatedByByteBudget === true
@@ -2106,7 +2104,10 @@ export const TERMINAL_METHODS: RpcAnyMethod[] = [
             oscLinks: serialized?.oscLinks,
             pendingEscapeTailAnsi: serialized?.pendingEscapeTailAnsi,
             alternateScreen: serialized?.alternateScreen,
-            scrollbackAnsiLength: serialized?.scrollbackAnsi?.length,
+            scrollbackAnsiLength:
+              typeof serialized?.scrollbackAnsi === 'string'
+                ? serialized.scrollbackAnsi.length
+                : undefined,
             truncated: false,
             truncatedByByteBudget: serialized?.truncatedByByteBudget,
             data: serialized?.data ?? ''
@@ -2394,8 +2395,6 @@ export const TERMINAL_METHODS: RpcAnyMethod[] = [
             source: serialized?.source,
             oscLinks: serialized?.oscLinks,
             pendingEscapeTailAnsi: serialized?.pendingEscapeTailAnsi,
-            alternateScreen: serialized?.alternateScreen,
-            scrollbackAnsiLength: serialized?.scrollbackAnsi?.length,
             data: serialized?.data ?? (read.tail.length > 0 ? `${read.tail.join('\r\n')}\r\n` : '')
           })
           // Why: baseline for resize re-stream gating; the client already rewrapped to these cols via the initial snapshot replay.
--- a/src/main/runtime/rpc/methods/terminal.ts
+++ b/src/main/runtime/rpc/methods/terminal.ts
@@ -77,11 +77,14 @@ type SnapshotFrameOptions = {
   source?: 'headless' | 'renderer'
   oscLinks?: TerminalOscLinkRange[]
   pendingEscapeTailAnsi?: string
+  alternateScreen?: boolean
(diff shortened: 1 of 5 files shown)


rodboev:
Current `main` preserves and prefers alternate-screen scrollback during restore, as carried by merged PR https://github.com/stablyai/orca/pull/17952. This PR is superseded on current `main`.

PR #10237, opened 2026-07-23
fix(claude-accounts): quote resolved claude path for Windows shell spawn

### LINKED ISSUE #10236: [Bug]: Claude account add fails on Windows when the user profile path contains a space (unquoted shell spawn)
### What happened

Adding a Claude account from **Settings → AI Provider Accounts** (Windows runtime) always fails when the Windows username contains a space. The error toast shows:

```
Claude command failed: 'C:\Users\Ji' ...
```

followed by mojibake (the localized cmd.exe error — on Korean Windows it is CP949 `'C:\Users\Ji'은(는) 내부 또는 외부 명령... 이 아닙니다`, i.e. *"'C:\Users\Ji' is not recognized as an internal or external command"* — decoded as UTF-8).

### Root cause

`runClaudeCommand` in `src/main/claude-accounts/service.ts` spawns the resolved CLI with the shell enabled on Windows:

```ts
: {
    command: resolveClaudeCommand(),
    args,
    env: { ...process.env, CLAUDE_CONFIG_DIR: configDir.windowsPath },
    shell: process.platform === 'win32'
  }
```

With `shell: true`, Node concatenates the command into the `cmd.exe /d /s /c` line **without quoting**. `resolveClaudeCommand()` returns an absolute path (e.g. `C:\Users\Ji Hoon\AppData\Roaming\npm\claude.cmd` for an npm install), so cmd.exe splits at the first space and tries to execute `C:\Users\Ji`.

The codebase already guards against exactly this elsewhere: `src/main/rate-limits/claude-pty.ts` routes through `cmd.exe /c "<quoted path>"` with the comment *"ensure the command path is properly quoted if it contains spaces"*, and `src/shared/ai-vault-types.ts` has `quoteWindowsCmdArg`. The account-add spawn path just missed the same treatment.

A secondary cosmetic issue: the captured stderr is CP949 (OEM code page) on localized Windows but rendered as UTF-8, so the error toast is garbled. Quoting the command makes the primary failure disappear; the encoding issue only affects how real CLI errors are displayed.

### Reproduction

1. Windows account whose profile path contains a space (e.g. `C:\Users\Ji Hoon`).
2. Install Claude Code via npm (`npm i -g @anthropic-ai/claude-code`) so it resolves under `%APPDATA%\npm`.
3. Orca → Settings → AI Provider Accounts → Claude → Add account (Windows runtime).
4. Fails immediately with the error above. Reproduced on Orca 1.4.152.

### Environment

- Orca 1.4.152, Windows 11 Pro (Korean locale)
- claude CLI 2.1.218 at `C:\Users\Ji Hoon\AppData\Roaming\npm\claude.cmd`

I have a one-line-class fix with a regression test; PR incoming.

ghoon4:
Fixes #10236

### Problem

`runClaudeCommand` spawns the resolved claude CLI with `shell: true` on Windows, but `spawn` concatenates the command into the `cmd.exe` line without quoting. When the CLI resolves to a path with spaces (any `C:\Users\First Last\...` profile, npm install), cmd.exe splits at the first space and every Claude account operation (add, status) fails with:

```
'C:\Users\First' is not recognized as an internal or external command
```

### Fix

Quote the resolved command when it will pass through the Windows shell, mirroring what the codebase already does for the same situation in `claude-pty.ts` (`cmd.exe /c "<quoted path>"`, *"ensure the command path is properly quoted if it contains spaces"*) and `quoteWindowsCmdArg` in `ai-vault-types.ts`. Non-Windows spawn (`shell: false`) is unchanged; the WSL branch was never affected.

### Tests

- New regression test: on `win32`, a resolved path containing spaces is passed to `spawn` quoted, with `shell: true`.
- Updated the existing `uses taskkill to cancel the Windows Claude login process tree` expectation to the quoted command (`'"claude"'`) — the intended behavior change.
- Converted the module-level `resolveClaudeCommand` mock to a `vi.hoisted` + `vi.fn` so a single test can override the resolved path without `doUnmock` wiping the hoisted mock for later tests.
- `vitest run src/main/claude-accounts/service.test.ts`: 26/26 pass (on a Windows host with a space in the profile path — previously 2 of these tests could only pass on hosts without one).
- `oxlint` and `tsc --noEmit -p config/tsconfig.node.json` clean on the touched files.

Verified end-to-end on Windows 11 (profile `C:\Users\Ji Hoon`, claude 2.1.218 under `%APPDATA%\npm`): reproducing the exact resolve + `shell: true` spawn fails unquoted and succeeds quoted.

### DIFF
--- a/config/scripts/claude-account-windows-spawn-repro.mjs
+++ b/config/scripts/claude-account-windows-spawn-repro.mjs
@@ -1,0 +1,342 @@
+import { spawn } from 'node:child_process'
+import { mkdtemp, mkdir, readFile, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { buildWindowsCommandInvocation } from '../../src/main/claude-accounts/windows-command-invocation.ts'
+
+const strategy = process.argv[2]
+if (!['baseline', 'candidate', 'explicit-cmd'].includes(strategy)) {
+  throw new Error(
+    'Usage: node config/scripts/claude-account-windows-spawn-repro.mjs <baseline|candidate|explicit-cmd>'
+  )
+}
+if (process.platform !== 'win32') {
+  throw new Error('This reproduction requires a physical Windows host.')
+}
+
+const expectedArgs = [
+  '',
+  'two words',
+  'amp&ersand',
+  'pipe|value',
+  'less<value',
+  'greater>value',
+  'caret^value',
+  'trailing\\',
+  'two-trailing\\\\',
+  '(parentheses)',
+  '100%',
+  '%ORCA_ARG_TRAP%',
+  'bang!value',
+  '한글-λ'
+]
+const tempRoot = await mkdtemp(join(tmpdir(), 'orca-claude-spawn-'))
+const reportedDir = join(tempRoot, 'Profile with spaces 한글')
+const reportedCapturePath = join(reportedDir, 'capture.json')
+const reportedPidPath = join(reportedDir, 'pids.json')
+const reportedShimPath = join(reportedDir, 'claude fixture.cmd')
+const reportedFixturePath = join(reportedDir, 'capture-child.cjs')
+const fixtureDir = join(tempRoot, 'Profile space & ^ (paren) %ORCA_PATH_TRAP% !bang! 한글')
+const capturePath = join(fixtureDir, 'capture.json')
+const pidPath = join(fixtureDir, 'pids.json')
+const shimPath = join(fixtureDir, 'claude fixture.cmd')
+const fixturePath = join(fixtureDir, 'capture-child.cjs')
+const fixtureEnv = {
+  ...process.env,
+  CLAUDE_CONFIG_DIR: join(fixtureDir, 'config space & ^ (paren) %ORCA_ENV_LITERAL% !bang! 한글'),
+  ORCA_ARG_TRAP: 'EXPANDED_ARG',
+  ORCA_PATH_TRAP: 'EXPANDED_PATH',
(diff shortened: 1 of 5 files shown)


OrcaWin (dismissed):
Thanks for the precise report and Windows reproduction. The diagnosis is correct, and quoting the resolved command does fix the common profile-path-with-spaces case.

Requesting changes because the current shell:true boundary still lets cmd.exe interpret valid path and argument characters. On a physical Windows host with Node 24 and a real .cmd shim:

- The reported spaced and Unicode path passed with this PR.
- A valid path containing %ORCA_PATH_TRAP% was environment-expanded and no longer addressed the intended command.
- Empty and spaced arguments were lost or split.
- &, |, <, >, ^, and % arguments were interpreted by cmd.exe; the > case physically redirected output and created a file.
- Node emitted DEP0190 because arguments passed with shell:true are concatenated without escaping.
- Error propagation and taskkill tree cancellation passed for the ordinary spaced-path case.

Please move ownership of the complete Windows command line into Orca: invoke ComSpec explicitly with /d /v:off /s /c, use shell:false plus windowsVerbatimArguments:true, quote every command/argument token, and protect percent expansion outside quoted spans. Keep the WSL branch ahead of native Claude resolution so WSL remains independent, and wait for taskkill completion before cleaning up the managed auth directory.

The regression test should supplement spawn mocks with a real Windows .cmd fixture under a temporary path containing spaces, Unicode, &, ^, parentheses, %, and !. It should capture exact argv/env, cover empty/spaced and cmd-metacharacter arguments, stderr/exit status, and verify taskkill removes the shell/child/grandchild tree.

This is a good issue diagnosis and a useful contribution; the requested change is specifically about making the Node-to-cmd.exe authority boundary safe enough to merge.

PR #10242, opened 2026-07-23
fix(terminal): keep SSH reconnect banner above the xterm canvas

nwparker:
## Summary

- Portal the non-blocking SSH reconnect banner into each pane container so WebGL/xterm compositing cannot paint terminal text through it
- Raise the banner z-index above pane-local chrome (focus rim is z-30) and use a fully opaque card background
- Cover the stacking regression with a unit assertion on the banner z-index class

## Before / after

| Before (banner under terminal text) | After (banner on top) |
|---|---|
| before | after |

**After — full pane context**

after context

## Reproduction and verification

**Before:** the banner was a `TerminalPane` sibling of the terminal host. With WebGL compositing, the canvas painted on top, so terminal output showed through the reconnect UI.

**After:** the banner is portaled into `.pane` (same pattern as `TerminalRemoteRuntimeReconnectBanner`), sits at the bottom of the full-height pane, and `elementFromPoint` hits the banner content.

Verified via `pn dev` + CDP on a disconnected `openclaw` SSH workspace:
- `parentIsPane: true`
- full pane height (1110px)
- `offsetBottom: 12` (`bottom-3`)
- `onTop: true`

## Tests

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/TerminalSshReconnectOverlay.test.tsx` (9/9)
- commit hooks: oxlint, React Doctor, oxfmt

### DIFF
--- a/src/renderer/src/components/terminal-pane/TerminalPane.tsx
+++ b/src/renderer/src/components/terminal-pane/TerminalPane.tsx
@@ -2896,16 +2896,23 @@ export default function TerminalPane({
           onRestartDaemon={() => daemonActions.setPending('restart')}
         />
       )}
-      {showSshReconnectOverlay && sshReconnectTargetId && sshReconnectStatus ? (
-        <TerminalSshReconnectOverlay
-          targetId={sshReconnectTargetId}
-          targetLabel={sshReconnectTargetLabel}
-          status={sshReconnectStatus}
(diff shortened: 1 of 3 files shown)


nwparker:
### Attachments

| before.png | after.png |
|---|---|
|  |  |

| after-context.png |
|---|
|  |

PR #10243, opened 2026-07-23
feat(agent-dashboard): choose in-window board or pop-out window

brennanb2025:
## What

The experimental **Agent Dashboard** previously opened only as a separate OS **pop-out window**. This adds an **"Open as"** mode so you can open it as an **in-window board** (new default) or a **pop-out window** (prior behavior). The mode row appears only when the Agent Dashboard feature is toggled on.

Default is **in-window** — the opposite of today.

## How it works

- New setting `experimentalAgentDashboardMode: 'in-window' | 'popout'` (default `'in-window'`). The sidebar entry branches on it: pop-out → `dashboard.openPopout()`; in-window → toggles a transient store flag.
- **In-window** (`AgentDashboardDrawer`) presents the same `AgentKanbanBoard` exactly like the workspace kanban board: a non-modal left `Sheet` companion to the sidebar (same geometry — sidebar live-width anchor, top-chrome offset, status-bar reserve, same slide animation and `bg-worktree-sidebar` surface). The sidebar stays interactive while it's open; Escape and click-off dismiss it with the same guards as the workspace board. The two companion boards are mutually exclusive — opening one closes the other.
- **Board settings menu** in the header, mirroring the workspace board's gear: an "Open as" control so the mode is changeable without a trip to Settings. Switching to pop-out hands the surface over (closes the drawer, opens the window). In-window only — it rides an optional `headerActions` slot, and the pop-out renderer has no store to drive it.
- The in-window snapshot is built locally (`useLiveDashboardSnapshot`) — the pop-out relays a serialized snapshot over IPC, but in-window there's no relay. Ack/reveal act on the local store; the pop-out's IPC `ackAgent`/`revealAgent` handlers are gated to the pop-out renderer, so `AgentKanbanBoard` gained `onAckAgent`/`onRevealAgent`/`onClose`/`containerClassName` props (defaults preserve the pop-out behavior). The `terminalPreview:*` IPC handlers admit the trusted main renderer so the in-window terminal dialog streams live PTYs.
- Shared chrome bounds extracted to `workspace-chrome-metrics.ts` (used by both companion sheets). `AgentDashboardExperimentalSetting` extracted to keep `ExperimentalPane` under the `max-lines` cap.

## Verification

Typecheck (web + node) clean, oxlint/oxfmt clean, max-lines ratchet + localization verifiers green, all covering renderer suites pass (the 3 failing sidebar suites fail identically at detached main). Verified live in the running Electron app (isolated dev profile): sidebar entry with bucket counts, drawer opens beside the sidebar via a real click, columns populate, Escape closes, board settings menu opens and the pop-out handoff works end to end (drawer closes, pop-out window opens, and the pop-out header correctly renders no in-window controls).

### 1. In-window board (default) — companion sheet beside the sidebar
in-window

### 2. Board settings menu — change the mode without leaving the board
board-settings

### 3. Setting — "Open as" (In-window default), shown only when the feature is on
settings

### 4. Pop-out window (opt-in)
popout

> Agents in the screenshots are synthetic status entries injected into an isolated dev profile to exercise the board surface.

## Notes / follow-ups

- Non-English locale JSON holds the English copy for the new strings; owned by the translation pipeline. (Stale pop-out-only descriptions in those catalogs were refreshed, since they contradicted the new default.)
- The in-window open state is a transient store flag, so disabling then re-enabling the experimental feature within a session re-shows the board if it was open — harmless, left as-is.
- Existing profiles with `experimentalAgentDashboardPopout: true` move to the in-window default rather than being backfilled to `popout`: that flag meant "enable the dashboard" (pop-out was the only option), so there is no expressed presentation preference to preserve, and the mode is one click away.


brennanb2025:
Thanks @coderabbitai — went through all four:

**1. `revealAgent` HMR guard** — fixed in b369097476. Confirmed `ackAgent` and `revealAgent` were added in the same commit (3468b434d6), so a stale dev preload lacks both; the asymmetry was arbitrary.

**2. Stacked Radix overlays** — resolved by the presentation rework (4a8c30345b). The in-window surface is now a **non-modal** `Sheet` (`modal={false}`), and Radix doesn't render `Dialog.Overlay` in that mode, so the terminal preview's scrim is the only one. No second scrim to suppress.

**3. Stale locale copy** — good catch, and it was broader than `ko.json`: the `description` and `copy` keys in **all four** non-English catalogs still read "Pop-out Kanban board…" / "…in a separate window", which now contradicts the in-window default. Refreshed all four to the current English source in b369097476. (These catalogs carry English placeholders pending the translation pipeline; the point was that the placeholders were stale, not merely untranslated.)

**4. Legacy `experimentalAgentDashboardPopout: true` → `popout` backfill** — intentionally not doing this. That flag meant "enable the Agent Dashboard", not "prefer the pop-out presentation" — pop-out was the only option at the time, so there's no expressed user preference to preserve. The default flip to in-window is the deliberate product change this PR makes, the feature is still experimental, and anyone who wants the old surface can switch in one click via the new "Open as" row.

Issue #1158, opened 2026-04-27
Fix silent data loss when workspace session hydration fails


brennanb2025:
## Problem

When Orca's startup sequence encounters an error during workspace session hydration (`App.tsx:308-340`), the catch handler resets the in-memory `tabsByWorktree` to an empty object. The debounced session writer then persists this empty state back to `orca-data.json`, silently overwriting the user's rich tab state (titles, ordering, splits, agent assignments).

There's a secondary failure mode in `hydrateWorkspaceSession` (`src/renderer/src/store/slices/terminals.ts:1295-1307`) where tabs for worktrees not yet known to the store are silently dropped and persisted.

Since there are no rolling backups, this data loss is unrecoverable.

## Common triggers

- Daemon restart or stale socket reference during startup
- SSH repo reconnect timing race against non-local worktrees  
- Partial worktree data from a slow fetch
- New post-merge code changing startup sequence timing

## Proposed fixes (in priority order)

1. **Gate session writer on hydration success**: Add a `hydrationSucceeded` flag; don't auto-write session if hydration failed. (S-M, cheap and safe)
2. **Rolling backups**: Before overwriting `orca-data.json`, rename current to `.bak.0`, shift older `.bak.N` up, keep last 5. (S-M, crude but rescues user from any future scenario)
3. **Broaden worktree-id filter guard**: Currently only guards the SSH case (`terminals.ts:1286-1294`). Extend to not drop tabs for worktrees the store hasn't hydrated yet. (S, narrow-targeted)
4. **Fail loud**: Surface hydration error to user with a prompt so they can report before state is overwritten. (M-L, requires error UI)

Fixes (2) and (3) are the most defensible and both cheap. (1) + (3) together would be comprehensive.

Made with [Orca](https://github.com/stablyai/orca) 🐋

nwparker:
Covered by PR #2641. I took over the PR, pushed hydration/session fixes, and posted validation evidence here: https://github.com/stablyai/orca/pull/2641#issuecomment-4573834749

Status: PR remains open with no GitHub checks reported and was not merged.

Jinwoo-H:
will be in next release, usually within 24 hours

PR #10256, opened 2026-07-24
feat(main): record main-thread hangs so we can measure them

brennanb2025:
## ELI5

Sometimes Orca freezes solid on newer macOS. Not slow — completely stuck. The window stops painting, nothing responds, and the only way out is force-quit.

The cause is a bug in Apple's code, not ours. Two pieces of macOS window machinery end up waiting on each other, and neither can go first, so they wait forever. Like two people in a doorway both saying "after you" — nobody moves, and they stand there all day. We reported it to Apple (FB24004458) and to Electron, but we can't fix it ourselves.

Here's the awkward part: **when this happens, we never find out.** Orca doesn't crash — crashes leave a report. It just sits there. So a user force-quits, reopens, mutters something, and gets on with their day. We have exactly one confirmed case ever, and no idea whether that means it's a freak accident or something hitting people every week.

So this PR adds a tiny second program that sits beside Orca and listens for a "still alive" ping every 2 seconds. If the pings stop for 45 seconds, it writes a note to disk. Next time Orca starts, it reads the note and tells us: *Orca was frozen for 90 seconds.* It also notices if Orca wakes back up on its own, which separates "genuinely dead" from "was just really busy for a while."

We thought about having this watchdog kill and restart Orca automatically when it goes quiet. We decided not to. If it ever guessed wrong, it would be shooting a perfectly healthy app in the head — quite possibly right in the middle of saving your work. Measuring is safe; killing is not. So we measure first, and only build the killer if the numbers say it's actually needed.

## Why we want this

**We do not know how often Orca's main thread freezes.** We have exactly one confirmed incident (macOS 26.5.1, kernel-confirmed spindump, unresponsive 1305+s) plus one third-party report. That is the entire dataset, and every decision downstream of it is currently a guess.

The reason it is a guess is structural, not incidental: **a deadlocked main thread never crashes.** There is no crash report, no exception, no exit code. No JS runs in the main process, so the app cannot self-report. The user force-quits and reopens, and nothing anywhere records that it happened. These freezes are invisible to us by construction, and they always have been.

This PR makes them visible. Concretely, it answers:

1. **Is this real?** Incidence per release, per macOS version. Today "Orca froze" is an unfalsifiable support report; after this it is a confirmed event with a duration.
2. **Did #10253 actually help?** We shipped the jiggle removal in 1.4.156 believing it shrank the trigger surface — but the captured deadlock enters from `_setFrameCommon:display:**fromServer:**`, a *window-server-delivered* update, so removing our own `setSize` may not have moved the needle at all. A per-version count settles it. Right now we would never know.
3. **Is auto-recovery safe to build?** `self_recovered` counts stalls that cleared on their own. **Every one of those is a SIGKILL an auto-recovery design would have gotten wrong.** If they dominate, killing was never viable and we should not build it. If unresolved stalls dominate, recovery is justified — and we would know that from data rather than from n=1.
4. **Is 45s the right threshold?** `unresponsive_ms` is kept raw, so the cutoff can be calibrated against the real distribution instead of judgement. The one incident sat dead for 21 minutes; we have no idea what the near-miss tail looks like.
5. **Evidence for Apple and Electron.** electron/electron#52437 is blocked on a repro nobody has produced, and FB24004458 is one spindump. "This affects N users per week on macOS 26.x, median stall X seconds" is materially harder to dismiss than a single stack, and it is obtainable without a repro.
6. **Regression detection.** If a future Electron bump or window change reintroduces exposure, the count moves. Without it we would learn from sporadic user reports of a bug that produces no artifact.

And the cheap outcome is a real outcome: **if the count comes back zero over a few releases, we delete this and close the issue.** That is a decisive result, and today we cannot reach it.

## What this does

A plain-Node watchdog child, forked with `ELECTRON_RUN_AS_NODE` (same crash-isolation pattern as the parcel-watcher child; it survives a deadlocked parent because it is its own process):

- Main thread sends a heartbeat every 2s over Node IPC.
- After 45s of silence the child writes a marker to userData. If heartbeats later resume, it rewrites the marker as resolved.
- The next launch consumes the marker, records a `main_thread_hang_detected` durable crash breadcrumb, and emits a `main_thread_hang_detected` telemetry event with `unresponsive_ms` and `self_recovered`.

The telemetry event is load-bearing, not redundant: the breadcrumb rides crash reports and a local trace span, and since a hang is not a crash, that data would never leave the user's machine. It fires after `initTelemetry` — `track()` drops silently until the client and store are wired.

Schema is `.strict()` with a duration and a boolean. No pid, path, or window title reaches the wire, and it rides existing consent (`track()` resolves consent per call and no-ops on non-official builds).

## What this deliberately does not do

An earlier revision SIGKILLed the silent parent and relaunched it. That is removed, because the risk is asymmetric:

- **True positive:** the main thread is already deadlocked, so anything in the persistence debounce window is unreachable no matter what. Killing recovers nothing the user does not get by force-quitting — it only saves them the discovery.
- **False positive:** it SIGKILLs a live main thread that was merely blocked. A 45s main-thread stall is most plausibly blocked I/O, which means writes are in flight — the misfire lands exactly when killing is most destructive.

The entire downside sits in the failure mode while the upside even when correct is modest. This codebase has already paid for arbitrary-moment process death twice (`persistence.ts` SIGKILL-race fixes for #217, `durable-file-write.ts` for the torn-write class in #1158); the removed design manufactured that hazard deliberately, on a timer, with no runtime off switch.

If the field data shows real deadlocks dominate, recovery can be built on evidence — off by default, renderer-corroborated, staged. `main-thread-hang-watchdog-entry.test.ts` asserts that observing a hang never spawns, signals, or exits, so that has to be a deliberate decision rather than drift.

## Misfire guards

- macOS-only; packaged-only (dev debugger pauses would trip it; `ORCA_HANG_WATCHDOG_FORCE=1` for testing).
- 45s threshold — the real incident sat dead for 21+ minutes.
- Suspend detection: a huge gap between check ticks means system sleep, not a hang; the wait restarts from zero after wake.
- The tick clock keeps advancing during an outstanding hang, so the first tick after one clears is not misread as a sleep gap (this was a live bug — it silently disarmed detection of any second hang in a session).
- Clean quit disarms via shutdown message + IPC disconnect; `will-quit` firing twice is handled (stop is idempotent).
- Watchdog death is non-fatal: the app runs without it, no restart loop.

Because nothing is killed, a misfire now costs one inaccurate data point.

## Testing

- Unit: child loop (detection, resolution, measured duration, sleep-gap, re-arming after a resolved hang), marker round-trip including the absent-flag case, watchdog fork/heartbeat/stop, telemetry schema including strict-mode rejection, and the no-spawn/no-signal/no-exit contract.
- `pnpm typecheck` clean (all three projects); `oxlint` clean; 54 unit tests pass.

### End-to-end verification against a genuinely frozen app

Unit tests use fake timers, so they can only prove the loop's arithmetic — not that a real watchdog observes a real freeze from its own process. Verified empirically against a live Electron instance, with the main process halted via `SIGSTOP` (no JS runs and no heartbeats can be sent — the same externally-observable state as the AppKit deadlock, and reversible):

| Scenario | Result |
| --- | --- |
| Healthy app, heartbeats flowing | no marker written |
| Main process `SIGSTOP`ped (artificial freeze) | marker written, `parentPid` = the real frozen Electron pid, `unresponsiveMs: 8093` |
| `SIGCONT` — app resumes | marker rewritten `selfRecovered: true`, `unresponsiveMs: 17442` |
| App killed while frozen (force-quit) | marker survives on disk, **0 orphaned watchdogs** |
| Next launch | marker consumed and deleted; re-consuming returns `null`, so a stall is never double-reported |

Also verified against the **built** `out/main/main-thread-hang-watchdog-entry.js` with a parent whose event loop is hard-blocked by a busy spin:

- Measured `unresponsiveMs` tracks the real freeze duration (5696ms measured for a 5500ms freeze).
- **Re-arm after a resolved hang**: first hang detected → cleared → second hang in the same session still detected. This is the bug the tick-clock fix addresses; confirmed it now re-arms rather than silently disarming.
- **Clean quit is silent**: both the graceful path (shutdown message + disconnect) and the disconnect-only path exit the child with code 0 and write no marker, waiting well past the threshold.

### Performance

Measured rather than asserted, since this adds a permanent process and two always-on timers:

| Metric | Measured |
| --- | --- |
| Watchdog RSS | 47.1 MB total, of which **44.9 MB is the `ELECTRON_RUN_AS_NODE` floor** (empty child baseline) — ~2.2 MB is the watchdog itself |
| CPU, sustained idle | 0.03s child CPU per 60s (~0.05%); parent-side heartbeats 23.8ms user / 7.6ms sys per minute |
| Startup marker read (common case, absent) | 3.3 µs |
| Startup marker read (post-hang) | 9.0 µs |

Against the `/perf` checklist: the fork is async and off the critical path; heartbeats cannot queue during a hang (a blocked event loop means the timer never fires); marker writes happen only on detect/resolve, not per tick; the child holds no unbounded buffers; and `will-quit` clears the timer and disconnects idempotently, so nothing leaks on the close path. The 44.9 MB floor is the same cost already paid by the `parcel-watcher` child this mirrors.

## Notes for review

- `hang-recovery-marker` → `hang-detection-marker`: it no longer recovers anything.
- The recovery notification and its five locale strings are gone — nothing restarts, so there is nothing to explain. Easy to restore if we later want to surface the event.
- Entry is registered in `electron.vite.config.ts`, in the plain-node entry guard, and in `asarUnpack`, so the build fails if the watchdog's module graph ever gains a `require('electron')`.
- Rebased onto current `main` and collapsed to a single commit; the branch previously carried merge commits and conflicted. Resolution was additive only — the sole conflict was in `telemetry-events.ts`, where `main` added `runtime_rpc_start_failed` in the same spot as this PR's schema; both are kept.
- Stale comment on the vite entry ("can force-relaunch the app") updated to match the observe-only design.

### DIFF
--- a/build-plugins/plain-node-entry-guard.ts
+++ b/build-plugins/plain-node-entry-guard.ts
@@ -21,6 +21,7 @@ type OutputChunk = Rollup.OutputChunk
 const PLAIN_NODE_ENTRY_NAMES = [
   'daemon-entry',
   'parcel-watcher-process-entry',
+  'main-thread-hang-watchdog-entry',
   'computer-sidecar',
   'agent-hooks/managed-agent-hook-controls',
   'codex/codex-app-server-grant-entry'
--- a/config/electron-builder.config.cjs
+++ b/config/electron-builder.config.cjs
@@ -153,6 +153,7 @@ module.exports = {
     'out/main/plugin-host-entry.js',
     'out/main/computer-sidecar.js',
     'out/main/parcel-watcher-process-entry.js',
+    'out/main/main-thread-hang-watchdog-entry.js',
     'out/main/chunks/**',
     'resources/**',
     'node_modules/ws/**',
--- a/config/scripts/electron-builder-config.test.mjs
+++ b/config/scripts/electron-builder-config.test.mjs
@@ -166,6 +166,14 @@ describe('electron-builder config', () => {
     )
   })
 
+  // Why: the watchdog only arms in packaged builds, and its ELECTRON_RUN_AS_NODE
+  // fork resolves the entry from app.asar.unpacked — inside the asar it never runs.
+  it('unpacks the forked main-thread hang-watchdog entry', () => {
+    expect(electronBuilderConfig.asarUnpack).toEqual(
+      expect.arrayContaining(['out/main/main-thread-hang-watchdog-entry.js'])
+    )
+  })
+
   it('uses the multi-size icon source for Linux packages', () => {
     expect(electronBuilderConfig.linux.icon).toBe('resources/build/icon.icns')
   })
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -208,6 +208,12 @@ export const electronViteConfig: UserConfig = {
           // Why: forked with ELECTRON_RUN_AS_NODE so @parcel/watcher faults
           // can't take down the main process (issue #7547).
           'parcel-watcher-process-entry': resolve('src/main/ipc/parcel-watcher-process-entry.ts'),
+          // Why: forked with ELECTRON_RUN_AS_NODE so it survives a deadlocked
+          // main thread (macOS 26 AppKit scene-update deadlock) and can record
+          // the stall for the next launch to report.
+          'main-thread-hang-watchdog-entry': resolve(
+            'src/main/hang-watchdog/main-thread-hang-watchdog-entry.ts'
+          ),
           // Why: run under ELECTRON_RUN_AS_NODE while the caller blocks on
           // spawnSync — codex app-server trust grants need a live event loop
           // but must finish before a Codex pane launch proceeds.
--- a/src/main/hang-watchdog/hang-detection-marker.test.ts
+++ b/src/main/hang-watchdog/hang-detection-marker.test.ts
@@ -1,0 +1,76 @@
+import { mkdtempSync, rmSync, writeFileSync, existsSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import {
+  consumeHangDetectionMarker,
+  hangDetectionMarkerPath,
+  writeHangDetectionMarker
+} from './hang-detection-marker'
+
+describe('hang detection marker', () => {
+  let dir: string
+
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), 'hang-marker-'))
+  })
+
+  afterEach(() => {
+    rmSync(dir, { recursive: true, force: true })
+  })
+
+  it('round-trips a marker and deletes it on consume', () => {
+    const markerPath = hangDetectionMarkerPath(dir)
+    writeHangDetectionMarker(markerPath, {
+      detectedAt: 123,
+      parentPid: 456,
+      unresponsiveMs: 45000,
+      selfRecovered: false
+    })
+    expect(consumeHangDetectionMarker(markerPath)).toEqual({
+      detectedAt: 123,
+      parentPid: 456,
+      unresponsiveMs: 45000,
+      selfRecovered: false
+    })
+    expect(existsSync(markerPath)).toBe(false)
+    expect(consumeHangDetectionMarker(markerPath)).toBeNull()
+  })
+
+  it('round-trips a self-recovered marker', () => {
+    const markerPath = hangDetectionMarkerPath(dir)
+    writeHangDetectionMarker(markerPath, {
+      detectedAt: 1,
+      parentPid: 2,
+      unresponsiveMs: 61000,
+      selfRecovered: true
+    })
+    expect(consumeHangDetectionMarker(markerPath)?.selfRecovered).toBe(true)
+  })
+
+  it('returns null for a missing marker', () => {
+    expect(consumeHangDetectionMarker(hangDetectionMarkerPath(dir))).toBeNull()
+  })
+
+  // Why: a marker written by the detect leg has no selfRecovered field until the resolve leg
+  // rewrites it, and "never resolved" is the conservative reading of its absence.
+  it('treats a missing selfRecovered flag as an unresolved hang', () => {
+    const markerPath = hangDetectionMarkerPath(dir)
+    writeFileSync(
+      markerPath,
+      JSON.stringify({ detectedAt: 1, parentPid: 2, unresponsiveMs: 45000 })
+    )
+    expect(consumeHangDetectionMarker(markerPath)?.selfRecovered).toBe(false)
+  })
+
+  it('returns null for corrupted or incomplete markers and still deletes them', () => {
+    const markerPath = hangDetectionMarkerPath(dir)
+    writeFileSync(markerPath, 'not json')
+    expect(consumeHangDetectionMarker(markerPath)).toBeNull()
+    expect(existsSync(markerPath)).toBe(false)
+
+    writeFileSync(markerPath, JSON.stringify({ detectedAt: 1 }))
+    expect(consumeHangDetectionMarker(markerPath)).toBeNull()
+    expect(existsSync(markerPath)).toBe(false)
+  })
+})
--- a/src/main/hang-watchdog/hang-detection-marker.ts
+++ b/src/main/hang-watchdog/hang-detection-marker.ts
@@ -1,0 +1,54 @@
+import { readFileSync, rmSync, writeFileSync } from 'node:fs'
+import { join } from 'node:path'
+
+// Why: written by the plain-Node watchdog child when main-thread heartbeats stop, and rewritten if
+// they resume; consumed on the next launch to report how long the stall lasted and whether it ever
+// cleared. `selfRecovered` separates a real deadlock from a long-but-survivable stall — the two are
+// indistinguishable at detection time, and only the latter would have been a destructive kill.
+export type HangDetectionMarker = {
+  detectedAt: number
+  parentPid: number
+  unresponsiveMs: number
+  selfRecovered: boolean
+}
+
+export function hangDetectionMarkerPath(userDataPath: string): string {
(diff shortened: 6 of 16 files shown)


brennanb2025:
**Live E2E on macOS 26.5 (the affected OS) — full recovery loop verified, one real bug found and fixed.**

Method: ran the built app isolated via the E2E userData contract with fast watchdog timings, then `SIGSTOP`'d the main process — freezes all JS exactly like the real deadlock while the watchdog child (separate process) keeps running.

- Freeze → watchdog SIGKILL at +6s (configured timeout) → marker written with correct pid → **app relaunched and stayed up**. Also verified the false-positive path live: continuous heartbeats for 2.5× timeout → parent untouched; clean shutdown disarms.
- **Bug caught by the E2E** (fixed in a40566cd5c): `open` propagates the caller's environment, and the watchdog child always runs with `ELECTRON_RUN_AS_NODE=1` — so the relaunched app booted as bare Node and exited immediately. Kill+marker worked; the app never came back. The relauncher now spawns with a scrubbed env (drops `ELECTRON_RUN_AS_NODE` and `ORCA_HANG_WATCHDOG_*`). Re-ran the E2E after the fix: relaunch confirmed (fresh instance up and stable).

Remaining gap: the marker-consume → recovery-notice leg is unit-tested but not yet E2E'd against a packaged build (dev relaunch opens the bare Electron shell, not Orca). Worth one packaged-build QA pass before stable.

nwparker:
Taking this over from @brennanb2025 as the new owner.

**Rebased onto current `main`** and collapsed to a single commit. The branch had drifted into `CONFLICTING`/`DIRTY` with merge commits from `main` mixed into the history; it is now `MERGEABLE`/`CLEAN`, one commit ahead and zero behind. The only real conflict was in `telemetry-events.ts`, where `main` landed `runtime_rpc_start_failed` in the same location as this PR's schema — resolution was purely additive, both are kept. Also fixed a stale comment on the vite entry that still described the removed force-relaunch behavior.

**Verified the change actually works** rather than trusting the fake-timer unit tests. Ran a live Electron instance with `ORCA_HANG_WATCHDOG_FORCE=1` and artificially froze the main process with `SIGSTOP` — no JS runs and no heartbeats can be sent, which is the same externally-observable state as the AppKit deadlock, and it is reversible:

- healthy app → no marker
- frozen → marker written with `parentPid` = the real frozen Electron pid, `unresponsiveMs: 8093`
- `SIGCONT` → rewritten `selfRecovered: true`
- killed while frozen → marker survives, **0 orphaned watchdogs**
- next launch → consumed and deleted; re-consuming returns `null`, so no double-reporting

Against the built entry with a hard-blocked event loop, the measured duration tracks the real freeze (5696ms for a 5500ms freeze), the **re-arm fix holds** (a second hang in the same session is still detected — this is the bug that previously disarmed detection silently), and clean quit is silent on both the graceful and disconnect-only paths.

**No `/perf` issues.** Measured, since this adds a permanent process and two always-on timers: ~0.05% CPU sustained idle, 3.3µs for the startup marker read in the common absent case, and of the 47.1 MB RSS, **44.9 MB is the `ELECTRON_RUN_AS_NODE` floor** — an empty child measures the same, so ~2.2 MB is actually the watchdog. That floor is already paid by the `parcel-watcher` child this mirrors. Marker writes happen only on detect/resolve rather than per tick, heartbeats cannot queue during a hang (the blocked loop means the timer never fires), and `will-quit` clears the timer and disconnects idempotently.

Full numbers and the scenario table are in the PR description.

One note for whoever reviews: `pnpm lint`'s react-doctor step fails locally on a stale `node_modules` plugin version (`Rule 'no-create-store-in-render' not found`). It reproduces on a pristine `main` checkout with no changes applied, so it is pre-existing and unrelated — flagging it rather than leaving it to look like fallout from this PR.

nwparker:
These three CodeRabbit comments are stale — they predate the refactor that removed the kill-and-relaunch design, and I confirmed each against the current tree:

1. **"Register the watchdog entry with the plain-Node guard" (Major)** — already registered (`build-plugins/plain-node-entry-guard.ts:24`; the comment targets the old `.js` path). Verified empirically rather than by inspection: `grep -rn electron` over the built `out/main/main-thread-hang-watchdog-entry.js` **and** its runtime chunk returns nothing, so the graph is genuinely plain Node. Also confirmed the entry and `out/main/chunks/**` are both in `asarUnpack`, so the entry's `require("./chunks/hang-detection-marker-*.js")` resolves inside `app.asar.unpacked` — worth calling out, since an unpacked entry with a packed chunk would fail to load in exactly the packaged-only mode this ships in.
2. **"Split the fallback cases so marker-write failure is actually tested"** — that test no longer exists. There is no bundle path and no kill, so the case it describes is gone; the replacement asserts `process.kill`/`process.exit`/`spawn` are **never** called.
3. **"Restore pre-existing environment values"** — the only `finally` block left is the `withPlatform` helper, which does capture and restore the original value.

nwparker:
**Correcting a number in the description above.** I reported the watchdog's memory as 47.1 MB RSS with a 44.9 MB `ELECTRON_RUN_AS_NODE` floor. RSS is the wrong metric for "what does adding this process cost" — it counts file-backed Electron framework pages that are already resident for the main process, so it charges shared memory to the child.

Measured `phys_footprint` instead:

```
Footprint: 13 MB
  Dirty      Clean    Category
 13 MB      31 MB     TOTAL
    phys_footprint: 13 MB
```

**The real marginal cost is ~13 MB, not ~47 MB** — the other 31 MB is clean, shared, reclaimable. My decomposition was directionally right (most of the RSS isn't the watchdog) but the metric overstated the true cost by ~3.6x. Flagging it because 13 MB vs 47 MB is exactly the kind of number a future "is this still worth keeping" decision would turn on, and the overstatement argues against keeping it.

Everything else in the perf section stands: ~0.05% CPU sustained idle, 3.3 µs startup marker read, writes only on detect/resolve.

PR #10289, opened 2026-07-24
feat(agent-status): report crush working/blocked/done state via SSE bridge

Lee-Si-Yoon:
## Summary

crush (charmbracelet/crush) was already registered as a launchable TUI agent, but it showed no working/thinking/editing status line — crush has no CLI hook contract like Claude/Codex/Gemini. Orca now enables crush's **client-server mode** (`CRUSH_CLIENT_SERVER=1`) with a per-pane custom unix socket passed via `--host`; the interactive TUI keeps running while Orca subscribes to the auto-spawned server's SSE stream (`GET /v1/workspaces/0/events`) and forwards each envelope as a synthetic hook event through the existing hook ingest pipeline, so crush panes get the same rich status reporting as the other supported agents.

### Why SSE and not OSC 9999
crush's own server-mode HTTP API + SSE endpoint is a first-class transport the project documents and supports. By contrast, OSC 9999 is Orca's private status protocol that other terminals don't read, and asking crush to emit it upstream would be a one-consumer private-channel request. The SSE bridge consumes crush's own event model (message / permission_request / run_complete / agent_event) with no per-side channel and no coordination needed from the crush maintainers.

## How it works

1. On crush PTY launch, Orca appends `--host unix://<tmp>/crush-orca-<launchToken>.sock` to the launch command and sets `CRUSH_CLIENT_SERVER=1` in the PTY env (gated to `crushSseBridgeSupported()`, i.e. macOS/Linux — Node's `http` unix-socket transport is unix-only).
2. crush's auto-spawned `crush server` binds the same `--host` (verified in crush's `startDetachedServer`: the spawned server inherits `--host`).
3. Orca's `startCrushSseBridge()` opens a long-lived `GET /v1/workspaces/0/events` connection on that socket, parses the chunked SSE stream in a stateful buffer-aware parser, and forwards each parsed envelope to `AgentHookServer.submitSyntheticHookEvent`.
4. The `crush` `AgentHookSource` normalizer maps SSE payload types to the standard `AgentStatusPayload` schema:
   - `message` role=user → `working` (new turn boundary; prompt resolved from text parts)
   - `message` role=assistant → `working` + active `tool_call` part's toolName/toolInput
   - `permission_request` → `blocked` with tool_name/params
   - `run_complete` → `done` with final `text` as `lastAssistantMessage`; `cancelled=true` → `interrupted: true`
   - `agent_event` type=response/error → `done`
5. On PTY exit (`onPtyExit`), the bridge's `stop()` halts reconnects and frees the handle.

## Files

- `src/shared/agent-hook-relay.ts` — add `'crush'` to `AgentHookSource` (exhaustive checks across the listener drive the remaining edits).
- `src/shared/agent-hook-listener.ts` — `normalizeCrushEvent` + `extractCrushToolFields` + four exhaustive `switch` cases (`isNewTurnEvent`, `extractToolFields`, `normalizeHookPayload` dispatch) + a `/hook/crush` route for symmetry/testing.
- `src/shared/agent-status-types.ts` — add `'crush'` to `WellKnownAgentType`.
- `src/shared/agent-session-resume.ts` — exhaustive `case 'crush'` (no provider session).
- `src/shared/crush-sse-shapes.ts` (new) — crush SSE payload wire types, stateful SSE line parser, per-pane socket path/host helpers.
- `src/main/agent-hooks/server.ts` — public `submitSyntheticHookEvent()` method so in-process subsystems can feed synthetic hook events through the same normalize→apply pipeline as HTTP POSTs.
- `src/main/agent-hooks/crush-sse-bridge.ts` (new) — node `http` unix-socket SSE client with capped exponential reconnect backoff.
- `src/main/runtime/orca-runtime.ts` — inject `CRUSH_CLIENT_SERVER=1` env and `--host unix://...` argv on crush launches; start the bridge after PTY spawn; stop it on `onPtyExit`.

## Tests

- `src/shared/crush-sse-shapes.test.ts` — 19 unit tests for socket path sanitization, SSE line/multi-line/chunk-boundary parsing, and envelope event-name/payload extraction.
- `src/shared/crush-normalizer.test.ts` — 9 end-to-end tests driving `normalizeHookPayload('crush', ...)` across every SSE payload type: user/assistant message, permission_request, run_complete (success/cancelled), agent_event (response/error/unknown), and non-hook SSE types.
- `src/main/agent-hooks/crush-sse-bridge.test.ts` — 5 socket integration tests (real unix-domain SSE server → bridge → `onEvent` forwarding, plus `stop()` idempotency). Skipped on Windows.

`pnpm typecheck`, `pnpm lint`, `pnpm test` (full suite, 35267 passing), and `pnpm build` all green.

## Reviewer summary (per CONTRIBUTING)

- **Cross-platform compatibility:** Unix-socket SSE is macOS/Linux only; Windows crush panes get a no-op bridge and continue to rely on the existing generic title-scraping detector. `crushSseBridgeSupported()` gates both the env injection and the bridge lifecycle. The socket directory uses `XDG_RUNTIME_DIR` when set and absolute, otherwise `os.tmpdir()` — matching crush's own `server.DefaultHost` resolution so the `--host` value lines up byte-for-byte with what the auto-spawned server binds.
- **SSH/remote compatibility:** The `CRUSH_CLIENT_SERVER` gateway plus an unknown-authority socket make remote crush unergonomic for v1, so the runtime guard further gates env injection to local launches; remote crush panes fall back to title-scraping just like today.
- **Supported agent / integration compatibility:** crush's `TUI_AGENT_CONFIG` entry is unchanged. Other agents are untouched — the `crush` branch is added as a new `AgentHookSource` case behind explicit `'crush'` equality checks, and existing exhaustive switches fail closed (they reject unknown sources at compile time, which is how this PR located its own touchpoints).
- **Performance risk:** the SSE parser is O(n) per chunk with no regex; reconnect backoff starts at 250 ms and caps at 10 s (matches crush's own SSE client). The bridge is registered to one `unref`'d timer per pane and is dropped on PTY exit, so no orphan timers. Synthetic events flow the existing in-process `normalizeHookPayload` path; no IPC overhead.
- **UI quality:** no UI changes — status arrives through the same `AgentStatusPayload` → store path as native hooks, so existing rendering for working/blocked/done/interrupted all apply unchanged.
- **Security risk:** the socket sits in a uid-scoped or `XDG_RUNTIME_DIR` (0700-ish) directory; crush itself enforces no cross-user access on its sockets. The bridge only reads from the socket (no writes), so a malicious crush server cannot push commands into Orca through it. Synthetic hook events go through the same `normalizeAgentStatusPayload` field-length/encoding normalization as native hook POSTs.

No visual change in this PR (status explanation only).

<details>
<summary>Quick local check for the reviewer</summary>

After installing crush and launching it from Orca, the per-pane worktree card should now show a working spinner, the active tool name/input for in-flight tool calls, a permission-blocked state on `permission_request`, and a terminal done state on turn end — same shape as Claude/Codex/Gemini. Before this change crush panes only ever showed a static agent icon.

</details>

💘 Generated with Crush

## ELI5

The crush agent could launch but never showed working/blocked/done status. Orca now listens to crush’s client-server SSE events and maps them into the same agent status pipeline as other TUIs.

### DIFF
--- a/src/main/agent-hooks/crush-sse-bridge.test.ts
+++ b/src/main/agent-hooks/crush-sse-bridge.test.ts
@@ -8,9 +8,23 @@ import {
   crushOrcaHostArg,
   crushOrcaSocketPath,
   crushSseBridgeSupported,
+  crushSseEnabledForLaunch,
   startCrushSseBridge
 } from './crush-sse-bridge'
 import { crushOrcaSocketFileName } from '../../shared/crush-sse-shapes'
+
+// Why: crushSseEnabledForLaunch branches on process.platform; stub it per test so
+// every platform branch is exercised on every host. Restore afterEach.
+const originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform')
+afterEach(() => {
+  if (originalPlatform) {
+    Object.defineProperty(process, 'platform', originalPlatform)
+  }
+})
+
+function stubPlatform(value: 'darwin' | 'linux' | 'win32'): void {
+  Object.defineProperty(process, 'platform', { value, configurable: true })
+}
 
 describe('crushOrcaSocketPath / crushOrcaHostArg', () => {
   it('builds an absolute socket path and a unix:// --host arg from the same launch token', () => {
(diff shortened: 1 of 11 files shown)


Lee-Si-Yoon:
Hey @brennanb2025 — when you get a moment, would appreciate a quick review here. Happy to walk through any of the SSE bridge / normalizer decisions.

PR #7214, opened 2026-07-03
Terminal performance initiative: pipeline fixes + term-speed-2 revival + PTY flow control (integration branch)

Jinwoo-H:
> [!IMPORTANT]
> **Long-lived integration branch — draft for visibility and review, not for immediate merge.** This is the terminal-performance initiative branch. It is under extended testing (RC cut via `release-cut`, daily-driving). Individual pieces may land on main as separate PRs where noted.

## What this branch contains

Four bodies of work, each kill-switched and independently benchmarked (full findings log: `notes/terminal-performance-initiative.md`):

### 1. Throughput/latency fixes for the agent-TUI pipeline (the "51× loss")
Baseline (prod 1.4.91): agent-TUI throughput **2.0 MB/s** vs 103 MB/s for bare xterm parse and daemon ingest; DSR-under-load p50 **134 ms** (vs Terminal.app 0.45 ms). Attributed by layered probing and fixed with three commits:
- **Parse-clocked scheduler drains** — high-priority drains re-arm from xterm write callbacks instead of fixed timer cadence; `HIGH_PRIORITY_MAX_WRITES_PER_DRAIN` 2→8 (isolation ceiling 27→117 MB/s).
- **Windowed retained-tail redraw** (main process) — suffix window sized by max upward cursor reach instead of full-tail rebuild per chunk (0.888→0.073 ms/chunk; 500-case differential fuzz proves byte-identical outputs).
- **Throttled wait-blocked check** (main process) — 50 ms coalesced cadence with immediate keyword bypass for actionable prompts (was ~85% of remaining per-chunk cost).

**Measured (dev, all three): agent-TUI 0.7→11.5 MB/s; DSR-under-load p50 161→18.8 ms, p99 624→24.9 ms.** Prod projection ~9–10 ms; definitive number pending the RC bench.

### 2. term-speed-2 revival (nwparker's model/view contract chain)
The full ~20k-line chain (hidden view parking, hidden delivery gate, side-effect authority, model query authority) merged and validated — 2,776+ targeted tests green at revival, A/B latency gate passed (p50 tied, p99 better with the chain). All features kill-switched via settings, default ON. Chain docs: `docs/reference/terminal-model-view-contract.md`.

### 3. Producer-side PTY flow control (protocol v19)
[...]

PR #10293, opened 2026-07-24
fix(terminal): bound retained background worktree terminal surfaces (renderer OOM)

OrcaWin:
## Summary

Fixes the top Windows **renderer-heap OOM** crash cluster (v1.4.145+). `Terminal.tsx` kept a mounted (hidden) terminal surface for **every worktree the user ever activated** — the mount set (`mountedWorktreeIdsRef`) was only pruned when a worktree was *deleted*, never bounded. Each retained surface holds an xterm.js `Terminal` + WebGL renderer + scrollback `ArrayBuffer`s (~3–4 MB each), so over a long session that switches between many worktrees the renderer heap climbs to the 3586 MB V8 ceiling and OOMs.

The fix evicts the **least-recently-hidden background worktree mounts beyond a cap of 8**, after a 30 s hysteresis, fully unmounting their surfaces. The PTYs persist in the terminal daemon, so re-activating an evicted worktree remounts through the existing cold-activation tab-deferral path.

## Reproduction (undeniable)

Reproduced live on a dev build via CDP + the shipped `renderer_memory_highwater` instrumentation — activating 3 distinct worktrees and switching back:

| step | `.xterm` (attached) | DOM nodes |
|---|---|---|
| activate wt #1 | 3 | 923 |
| activate wt #2 | 4 | 1073 |
| activate wt #3 | 5 | 1223 |
| **switch back to #1** | **5** (never released) | **1223** |

A HeapProfiler snapshot confirmed the retained payload is `JSArrayBufferData` (xterm scrollback), ~3–4 MB/terminal. Trigger = `setActiveWorktree` (the `sidebar_worktree_activate` path). Matches every production clue: churning key (worktree id), large payloads, rapid worktree switching, recent builds, long sessions.

**Failing-before / passing-after test:** `terminal-worktree-mount-eviction.test.ts` simulates the Terminal render pass over 40 worktree activations. Without the eviction step (= current prod code) the mount set grows to 40 → `AssertionError: expected 40 to be less than or equal to 9`; with eviction it stays ≤ 9. 11 tests total.

## The fix

- New `src/renderer/src/components/terminal/terminal-worktree-mount-eviction.ts` — pure, tested eviction: pick the least-recently-hidden mounted worktrees over the cap and remove them from the mount set + the same bookkeeping maps the existing deletion-prune cleans. Extracted to its own file because `Terminal.tsx` and `background-terminal-worktree-mount.ts` are at the 300 max-lines limit (no `max-lines` disable added).
- `Terminal.tsx` — in the same render pass as the existing "prune deleted worktrees" block, size-gated (`> cap+1`) so the safety scans stay off the common small-session path.
- Cap = `TERMINAL_WORKTREE_HOT_RETAIN_LIMIT` (8) and hysteresis = `TERMINAL_WORKTREE_COLD_PARK_DELAY_MS` (30 s) — reuses the warm working-set + cold-park timing the parking design already chose, so eviction is "parking, one level up" and the existing recheck timer already re-renders at the 30 s deadline.

### Guardrails (never evicted)
Active worktree; worktrees with a live (non-`done`) agent; tabs with a pending startup/spawn; targeted background mounts (wake/resume, mobile subscription, CLI create); activity-portal or currently-measuring surfaces; and tabs whose live PTY has no local daemon snapshot (remote runtime / SSH / fail-open) since those can't replay on remount.

## ELI5

Every time you clicked into a worktree, Orca quietly kept that worktree's terminals fully loaded in memory in the background — forever, even after you moved on — so it could switch back instantly. Over a long day of hopping between many worktrees, all those invisible terminals (each holding megabytes of scroll history) piled up until the window ran out of memory and crashed. The fix keeps the 8 most-recently-used worktrees loaded and unloads the older idle ones; switching back to an old one just reloads it (its session is safe in the background daemon).

## Trade-offs (honest)

- Switching back to a worktree idle > 30 s and beyond the 8 most-recent pays a **cold remount** (~170 ms/pane, deferral-capped for many-tab worktrees) instead of instant reveal.
- An evicted worktree loses its parked byte-watchers, so a bell/title from a *non-agent* long-running command in a long-idle over-cap worktree can go unnoticed (live agents are exempt; this matches the never-visited-worktree baseline).
- Paired mobile/web clients: targeted mobile mounts (the normal case) are exempt; an already-attached passive stream in a desktop-visited worktree could blank briefly until the client re-requests its mount.
- No API, persistence, or provider change.

## Adversarial review
- **1 loop, 2 independent reviewers** (correctness + behavior/UX/rules), both **CLEAN — no blocking defects**.
  - Correctness reviewer verified the overflow math, render-time ref mutation (matches the existing prune pattern), live-agent/SSH guards, `hiddenSince` population, remount cleanup, and prune ordering — and confirmed eviction genuinely reclaims the scrollback (unmount → `TerminalPane` cleanup → `terminal.dispose()`).
  - Behavior reviewer confirmed the fix is **more conservative than the shipped baseline**: an existing pane-parking layer already unmounts hidden panes at the *same* cap (8) and delay (30 s), so this adds no new remount frequency or scrollback loss; it's the complementary surface-level bound (and the only bound when parking's kill-switch is off). Project-rules compliant (no `max-lines` disable added).
  - Only actionable items were two comment-accuracy clarifications (recheck-timer caveat when parking is disabled; the deliberate cap/parking coupling) — applied.
- **Loops until clean: 1.** 11 tests pass; oxlint + web `tsc` clean.

## Note
Reproduced/verified on a build with the `renderer_memory_highwater` instrumentation the team shipped separately — that breadcrumb will confirm `mountedWorktreeIds`/`.xterm` counts stay bounded in the wild post-fix.

### DIFF
--- a/src/renderer/src/components/Terminal.tsx
+++ b/src/renderer/src/components/Terminal.tsx
@@ -70,6 +70,12 @@ import {
   takeAllPendingBackgroundTerminalWorktreeMounts,
   takePendingBackgroundTerminalWorktreeMount
 } from './terminal/background-terminal-worktree-mount'
+import {
+  MAX_BACKGROUND_MOUNTED_TERMINAL_WORKTREES,
+  collectLiveAgentTerminalAttribution,
+  evictExcessBackgroundTerminalWorktreeMounts,
+  isTerminalWorktreeEvictionSafe
+} from './terminal/terminal-worktree-mount-eviction'
 import { hasRegisteredRuntimeTerminalTab } from '../runtime/sync-runtime-graph'
 import {
   getEffectiveLayoutForWorktree as getEffectiveLayout,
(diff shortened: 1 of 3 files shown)


OrcaWin:
Closing. Investigation showed this duplicates the already-shipped two-tier hidden-view parking layer (#7214), which already unmounts background *worktree* panes — freeing xterm + scrollback — at the same cap (8 hot-retain) and delay (30s) this PR used. The 3–4 MB/terminal scrollback (this PR's headline payload) is already reclaimed by parking; this eviction only additionally drops the residual byte-watcher + null-pane surface shell, which is not the OOM signature. The reproduction used 3 worktrees (under the warm cap of 8) so it demonstrated designed warm-set behavior, not unbounded growth. Re-examining the actual OOM root cause instead.

PR #10297, opened 2026-07-24
oom(04/29): bound shared remote-runtime/relay admission

nwparker:
## OOM hardening slice 04/29 — bound shared remote-runtime/relay admission

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **04/29** (base: `oom/03-a3-shared-fs-listing`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Adds 8 new self-contained shared modules that impose count/byte/depth admission ceilings on relay JSON parsing, relay .version marker reads, remote-runtime pending RPC requests/subscriptions, and retained SSH connection-state/port-forward payloads. No existing callers are modified in this slice, so the code is inert until wired by other slices; the two included tests exercise boundary behavior.

### ELI5
Imagine bouncers added at several doors of the app that talk to remote/SSH machines. Each bouncer counts how many requests are waiting, how big they are, and throws away junk fields or oversized text before it piles up in memory. This PR only installs the bouncers (as reusable helpers) and their unit tests; nothing else changed, so ordinary users see no difference unless a remote machine floods the app with abnormally large or numerous payloads.

### Proof of OOM fix
- relay-json-admission.ts: RELAY_JSON_MAX_STRUCTURAL_TOKENS=1,000,000 and RELAY_JSON_MAX_NESTING_DEPTH=128 gate every relay JSON.parse via assertJsonTextStructureWithinLimits
- relay-version-marker.ts: RELAY_VERSION_MARKER_MAX_BYTES=4KiB caps the .version file read; tested at exact boundary (accept) and +1 byte (throws NodeFileReadTooLargeError) in relay-version-marker.test.ts
- remote-runtime-memory-limits.ts (dep): MAX_PENDING_REQUESTS=256, MAX_PENDING_RPC_BYTES=32MiB, process-level = 2x (512 reqs / 64MiB); enforced in remote-runtime-prepared-request-admission.ts prepare/reserve
- remote-runtime-shared-control-admission.ts: MAX_SUBSCRIPTIONS=256 and MAX_RETAINED_SUBSCRIPTION_BYTES=16MiB gate subscription admission
- ssh-retained-payload-admission.ts: SSH_DETECTED_PORTS_MAX_ENTRIES=50, CONNECTION_ERROR=16KiB, IDENTIFIER=1KiB, PROCESS_NAME=4KiB, ADVERTISED_URL=2KiB; tests in ssh-retained-payload-admission.test.ts cover row cap, UTF-8-safe clamping (no split code point), identifier mismatch/oversize rejection, malformed-row drops

### Regressions
- **low**: admitSshConnectionState whitelists fields: unknown/extra properties are stripped and optional fields (connectionGeneration, supportsFolderDownload, remotePlatform) are dropped unless they pass type checks. Callers that previously forwarded extra fields would no longer round-trip them. _(trigger: Only once callers in other slices route SSH connection state through this admitter, and only for payloads carrying fields outside the known schema.)_

- prepareRemoteRuntimeRequest throws 'remote_runtime_busy' only past 256 pending client requests or 512 process-wide, or when retained bytes would exceed 32MiB/64MiB — far above ordinary concurrent RPC counts
- admitSharedControlSubscription throws 'remote_runtime_busy' only past 256 live subscriptions or 16MiB retained params
- admitSshDetectedPorts scans only the first 50 rows (Math.min(length,50)); ports beyond index 50 are never examined — far above typical forwarded-port counts
- relay JSON with >1M structural tokens or >128 nesting depth is rejected instead of parsed — only reachable with pathological/abusive payloads
- SSH detected-port rows are validated: rows with port<1/>65535 or empty/oversized host are dropped, and advertisedUrl exceeding 2KiB is discarded (undefined) rather than shown. Malformed or oversized-URL rows stop surfacing. _(only when: Only when a port-forward advertises an anomalously long URL or malformed row, and once callers adopt admitSshDetectedPorts.)_

### Build / CI
- Part of the **Shared foundation** group; this group's cumulative tree is interlinked, so it becomes typecheck-green at its checkpoint **PR #06** (whole-repo interconnection — see stack README). Content is exact production code.

### Adversarial review
- Slice adds 8 new self-contained shared modules under src/shared/. Within THIS slice no callers are modified (the wiring lives in other slices), so the code is inert until wired; I confirmed the caller files exist at TARGET but are outside this filelist. I verified every numeric ceiling is far above ordinary use, refuting the possibility of a normal-path regression:\n\n- relay-json-admission.ts: RELAY_JSON_MAX_STRUCTURAL_TOKENS=1,000,000 and RELAY_JSON_MAX_NESTING_DEPTH=128. A structural token is each of {}[],: (json-text-structure-limit.ts); ~8 tokens per file/list entry means ~125k array elements are needed to trip the cap. Only pathological/abusive relay payloads reach it. Depth 128 is non-pathological JSON. OVERLOAD-ONLY.\n- relay-version-marker.ts: RELAY_VERSION_MARKER_MAX_BYTES=4KiB for a version-string marker file — enormous headroom.\n- remote-runtime-prepared-request-admission.ts: 256 pending/connection, 512 process-wide; 32MiB/64MiB retained. Ordinary concurrent RPC counts are tiny. Admission Set is a module-level global (process-wide by design; MAX_PROCESS_*=2x per-connection), released via returned closure + releaseRemoteRuntimePreparedRequest (both null out retainedBytes and delete from the Set) — no leak on normal or error path. OVERLOAD-ONLY.\n- remote-runtime-shared-control-admission.ts: 256 subscriptions / 16MiB retained params — far above normal. OVERLOAD-ONLY.\n- remote-runtime-request-ready-waiters.ts: plain waiter list with splice-based drain on resolve/reject; no bound logic itself, no regression.\n- ssh-retained-payload-admission.ts: CONNECTION_STATUSES set matches the SshConnectionStatus union exactly (all 8 values) and the whitelisted output fields match SshConnectionState exactly (targetId/status/error/reconnectAttempt + optional connectionGeneration/supportsFolderDownload/remotePlatform), so no valid status or documented field is dropped — only genuinely unknown/extra fields are stripped. Clamps: connection error 16KiB, host 1KiB, processName 4KiB, advertisedUrl 2KiB, 50 detected-port rows — all far above realistic SSH values (hosts like 127.0.0.1, short banner URLs, few dev ports).\n\nTests confirm boundary behavior: ssh-retained-payload-admission.test.ts checks unknown-field stripping, UTF-8-safe error clamping (no split code point), targetId mismatch/oversize rejection, port-row cap at SSH_DETECTED_PORTS_MAX_ENTRIES, and dropping malformed rows. relay-version-marker.test.ts covers the marker read/limit.\n\nNo confirmed normal-path regression or correctness bug; signoff=true. The two low-severity notes above are by-design bounds worth a human glance but do not affect ordinary users.

### Files
8 files changed (+510 / −0).

---

### DIFF
--- a/src/shared/relay-json-admission.ts
+++ b/src/shared/relay-json-admission.ts
@@ -1,0 +1,12 @@
+import { assertJsonTextStructureWithinLimits } from './json-text-structure-limit'
+
+export const RELAY_JSON_MAX_STRUCTURAL_TOKENS = 1_000_000
+export const RELAY_JSON_MAX_NESTING_DEPTH = 128
+
+export function parseRelayJsonText<T>(text: string): T {
+  assertJsonTextStructureWithinLimits(text, {
+    structuralTokens: RELAY_JSON_MAX_STRUCTURAL_TOKENS,
+    nestingDepth: RELAY_JSON_MAX_NESTING_DEPTH
+  })
+  return JSON.parse(text) as T
+}
--- a/src/shared/relay-version-marker.test.ts
+++ b/src/shared/relay-version-marker.test.ts
@@ -1,0 +1,40 @@
+import { mkdtempSync, rmSync, truncateSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, describe, expect, it } from 'vitest'
+import { NodeFileReadTooLargeError } from './node-bounded-file-reader'
+import { RELAY_VERSION_MARKER_MAX_BYTES, readRelayVersionMarkerSync } from './relay-version-marker'
+
+const roots: string[] = []
+
+function createVersionFile(contents: string): string {
+  const root = mkdtempSync(join(tmpdir(), 'orca-relay-version-marker-'))
+  roots.push(root)
+  const filePath = join(root, '.version')
+  writeFileSync(filePath, contents)
+  return filePath
+}
+
+afterEach(() => {
+  for (const root of roots.splice(0)) {
+    rmSync(root, { recursive: true, force: true })
+  }
+})
+
+describe('relay version marker', () => {
+  it('accepts a trimmed marker at the exact byte boundary', () => {
+    const version = '1.2.3+deadbeef'
+    const filePath = createVersionFile(
+      version + ' '.repeat(RELAY_VERSION_MARKER_MAX_BYTES - Buffer.byteLength(version))
+    )
+
+    expect(readRelayVersionMarkerSync(filePath)).toBe(version)
+  })
+
+  it('rejects a sparse marker one byte over the boundary', () => {
+    const filePath = createVersionFile('1.2.3')
+    truncateSync(filePath, RELAY_VERSION_MARKER_MAX_BYTES + 1)
+
+    expect(() => readRelayVersionMarkerSync(filePath)).toThrow(NodeFileReadTooLargeError)
+  })
+})
--- a/src/shared/relay-version-marker.ts
+++ b/src/shared/relay-version-marker.ts
@@ -1,0 +1,9 @@
+import { readNodeFileSyncWithinLimit } from './node-bounded-file-reader'
+
+export const RELAY_VERSION_MARKER_MAX_BYTES = 4 * 1024
+
+export function readRelayVersionMarkerSync(versionFile: string): string {
+  return readNodeFileSyncWithinLimit(versionFile, RELAY_VERSION_MARKER_MAX_BYTES)
+    .buffer.toString('utf8')
+    .trim()
+}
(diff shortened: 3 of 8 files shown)


nwparker:
Superseded by #10299, which now carries slices 03-06 as one atomic shared-foundation PR. Reason: the `src/shared` modules in these slices are mutually coupled (new bounded readers here depend on types/budgets modified in slice 06, whose own tests depend on these files), so no ordering makes them individually CI-green. Combining them is the honest shape for this layer. No content is lost.

PR #10298, opened 2026-07-24
oom(05/29): add BoundedMap primitive + migrate 3 caches; shared misc limits

nwparker:
## OOM hardening slice 05/29 — bound shared misc primitives (secure-path, mcp, hermes, setup-script)

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **05/29** (base: `oom/04-a4-shared-remote-relay`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Slice A5 adds 28 new src/shared modules (with tests) that define reusable OOM ceilings — bounded caches, LRU eviction, file/JSON byte+entry caps, transcript truncation, bounded fan-out/concurrency, and admission checks — for agent hooks, Hermes run history, external automation jobs, MCP config inspection, orca.yaml, remote-runtime RPC, secure-path caching, setup-script import, and terminal sizing.

### ELI5
Imagine adding sturdy measuring cups and lids to a kitchen so no pot can ever overflow. This PR only makes the cups and lids (and tests that they hold exactly the right amount) — it doesn't yet put them on the stove. Each one says "stop at this much" so a giant or malicious input can't flood the app's memory.

### Proof of OOM fix
- agent-hook-endpoint-temp-cleanup.ts: AGENT_HOOK_ENDPOINT_SWEEP_MAX_ENTRIES=1024 dir-scan cap, 5min stale cutoff; test asserts scan stops at cap and only stale .endpoint-*.tmp removed
- agent-hook-status-cache.ts: MAX_AGENT_HOOK_STATUS_CACHE_PANES=500, evicts done/stale-first then oldest, clears pane caches; test agent-hook-status-cache.test.ts covers both eviction paths
- external-automation-jobs-file.ts: 8 MiB file cap + 10,000-entry cap + JSON structural-token(1M)/depth(128) guard; test rejects oversized sparse file, >10k jobs, and structural amplification
- hermes-run-output-limits.ts: primary 5 MiB, transcript 8 MiB / 10,000 msgs, run-page retained 32 MiB, hydration concurrency=2, metadata 16 KiB; tests assert concurrency=2, order preserved, UTF-8-safe truncation, row-pull stops after maxMessages
- hermes-run-ref-retention.ts: HERMES_RUN_REF_MAX_ENTRIES=10,000, id 4 KiB; retainer keeps newest window and reports saturated; test verifies exact newest-first window
- mcp-config-inspection-limits.ts / mcp-server-inspection.ts: 256 KiB text, 256 servers, 256 env fields, 4 KiB name, 64 KiB field; oversized fields mark server invalid rather than crash
- orca-yaml-file-limit.ts: 256 KiB file, 64 KiB field, 256 collection entries, alias-expansion cap 20; orca-yaml-bounds/alias tests assert exact boundary admit and +1 reject before parse
- remote-runtime-outbound-admission.test.ts: caps on MAX_PENDING_REQUESTS, MAX_PENDING_RPC_BYTES, MAX_PROCESS_PENDING_* , MAX_SUBSCRIPTIONS, MAX_READY_WAITERS, MAX_OUTBOUND_JSON_BYTES; tests prove admission rejects before opening socket AND full release on close/timeout (pendingRequestCount and retainedBytes return to 0)
- secure-path-hardening-cache.ts: LRU with maxEntries/maxKeyBytes/maxTotalKeyBytes, oversized key rejected without evicting retained state; test covers per-key boundary, LRU count eviction, aggregate-byte eviction
- node-bounded-file-reader-sync.test.ts: readNodeFileSyncWithinLimit rejects oversized sparse file before allocating declared size
- setup-script-import-limits.ts: file 256 KiB, field 64 KiB, 256 command parts, 256 cmux commands, 64 keywords, 128 unsupported fields, 4096 TOML lines
- terminal-size-limits.ts: MAX_TERMINAL_COLS/ROWS=500, admission returns error string outside 1..500, normalizeTerminalSize falls back to 80x24
- quick-open-readdir-memory.test.ts: budget caps maxEntries/maxDirectories/maxDepth/maxPathCodeUnits with exact-boundary + reject tests
- pr-refresh-memory-limits.ts: queue/visible-candidate/retry/active-scope caps of 1024 each, alias cap 64

### Regressions

- orca.yaml rejected wholesale (parseOrcaYaml returns null) only when file >256 KiB, any field >64 KiB, a collection >256 entries, or >20 aliases — far above any real orca.yaml
- External automation jobs file load throws only above 8 MiB or 10,000 jobs or JSON amplification — ordinary job files are tiny
- Hermes transcript truncated with a notice only past 8 MiB or 10,000 messages; run-page output omitted per-run only after 32 MiB aggregate retained on a single history page
- Remote-runtime requests rejected with remote_runtime_busy / 'JSON payload exceeds' only under sustained stalled-handshake backpressure or multi-MB payloads
- Agent hook status cache evicts a pane only beyond 500 concurrently-tracked panes
- MCP server marked invalid only when a field/name/env exceeds 64 KiB/4 KiB/256 fields
- Secure-path cache skips caching (recomputes) a path only when its key exceeds the per-key/total byte bound
- Terminal sizes are clamped to 1..500 cols/rows; requests outside that are rejected or normalized to 80x24 fallback. _(only when: Only reachable if a terminal is resized beyond 500 columns or rows — not attainable with ordinary fonts/monitors. Note: the admission helper is not wired to any caller within THIS slice, so no behavior change lands until a consumer slice imports it.)_
- orca.yaml with >256 defaultTabs (or an over-limit field/alias) is rejected entirely (null) rather than partially loaded. _(only when: A user hand-authoring an orca.yaml with more than 256 default tabs or a 64 KiB+ field; wiring lands in another slice.)_
- Hermes history page can show 'Run output omitted because this history page exceeds the memory limit' for later runs on a page whose aggregate output exceeds 32 MiB. _(only when: Viewing a run-history page whose combined run outputs exceed 32 MiB; earlier runs keep output, later ones are omitted (order preserved). Consumer wiring is in another slice.)_

### Build / CI
- Part of the **Shared foundation** group; this group's cumulative tree is interlinked, so it becomes typecheck-green at its checkpoint **PR #06** (whole-repo interconnection — see stack README). Content is exact production code.

### Adversarial review
- All 28 files in slice A5 are brand-new src/shared modules (verified: `git cat-file -e aab112933e:<path>` fails for every one). Grepping the BASE tree for each module name (agent-hook-status-cache, external-automation-jobs-file, hermes-run-output-limits, secure-path-hardening-cache, terminal-size-limits, orca-yaml-file-limit, mcp-server-inspection, setup-script-import-limits, remote-runtime-outbound-admission, pr-refresh-memory-limits, etc.) returns ZERO consumers. Because these are new files, no pre-existing (unchanged) code can import them, and any consumer wiring would appear as a MODIFIED file in a different slice. Therefore this slice adds only unwired library code + tests: NO normal-path or overload-path behavior change can land here.

Verified the numeric ceilings are all far above ordinary use and match the prior reviewer's claims:
- terminal-size-limits.ts: MAX cols/rows = 500 (normalizeTerminalSize falls back to 80x24). 500 is well beyond any real terminal; unwired here anyway. Consumer slice should confirm fallback-vs-reject semantics, but nothing lands in A5.
- orca-yaml-file-limit.ts: 256 KiB file / 64 KiB field / 256 collection entries / 20 aliases — far above any hand-authored orca.yaml.
- external-automation-jobs-file.ts: 8 MiB / 10,000 jobs / 1M structural tokens / depth 128 — ordinary job files are tiny.
- hermes-run-output-limits.ts: 5 MiB primary output, 8 MiB / 10,000-msg transcript, 32 MiB retained per history page, concurrency 2. Per-run output omission is order-preserving and only after 32 MiB aggregate retained on one page.
- secure-path-hardening-cache.ts: LRU with balanced retainedKeyBytes accounting (delete() decrements, set() increments); over-large keys skip caching with no leak. Logic checks out.
- agent-hook-status-cache.ts: evicts only beyond 500 concurrently-tracked panes; prefers done/stale entries, falls back to oldest.
- mcp/setup-script/pr-refresh/hermes-run-ref modules: generous ceilings (256 KiB, 4 KiB names, 1024 queue entries, 10,000 refs).

OOM evidence is present and test-backed (each source module ships a matching .test.ts). Could not refute the "no ordinary-user change" claim — modules are dead code until a consumer slice imports them. Draft-PR safe for human review.

### 🧱 `BoundedMap` — shared bounded-memory primitive (new)

Adds one tested `src/shared` primitive so bounded caches stop being re-hand-rolled per site, and **migrates three call sites onto it**:

| Adopter | Before | After |
| --- | --- | --- |
| `shared/secure-path-hardening-cache.ts` | 52-line bespoke LRU | thin wrapper over `BoundedMap` (original test unchanged & green) |
| `main/github/conflict-summary-cache.ts` | local `setBoundedMapEntry` copy | `new BoundedMap({ maxEntries })` ×2 |
| `main/runtime/orca-runtime.ts` | byte-identical `setBoundedMapEntry` copy | `new BoundedMap({ maxEntries })` ×2 |

Both duplicated helpers are **deleted** — `setBoundedMapEntry` no longer exists anywhere in the repo. Net effect on this slice's non-#10179 surface: **−80 lines**.

**Contract** (documented in-file, each clause tested): `set()` never reports success for an entry it did not retain; weights must be non-negative safe integers and an unmeasurable/throwing `sizeOf` fails **closed** (entry rejected, ledger intact — no crash on a relay running on someone else's machine); `maxEntryBytes` is clamped to `maxBytes`; `keys()/values()/entries()` return snapshots so a touching read during traversal terminates (`peek()` reads without reordering); `onEvict` fires only for capacity eviction, only after both bounds are restored, and one throwing hook cannot skip the others.

**Verification:** 26 unit tests (exact-limit + limit+1, LRU order, NaN-key protection, safe-integer boundary, reentrant/throwing disposal), **64,000-op randomized invariant fuzz across 8 configs** asserting after every op that the ledger equals the shadow sum and both ceilings hold, plus 1,425 adopter-suite tests green.

**Reviewed adversarially over three rounds** (fable-xhigh + gpt-5.6-sol-xhigh, every claim independently verified by reproduction). Round 1 found 3 real defects, round 2 found 4 more, round 3 came back with **zero adopter-reachable findings**; all were fixed here. Two further primitives (`AdmissionController`, `BoundedQueue`) were **deliberately dropped** from this PR: they had no callers, and review showed each was mis-shaped for the very site it named (`withAdmission` is lexical-only and would have released the watcher's cross-IPC claim too early; `BoundedQueue`'s `shift()` storage hits a V8 cliff at ~16k, right at the 16,384 PTY span cap). They will return with a real adopter that shapes them.

**OSS alternatives were evaluated and rejected** (`lru-cache` et al.): the installed `lru-cache@11` requires node ≥20 while the relay bundles at `node18`; it throws on a zero `sizeCalculation` (our most common mode is count-only), silently refuses `undefined` values, and — verified by probe — produces a **negative** `calculatedSize` under a config it accepts. Details in the review notes.

### Files
34 files changed (+2769 / −0).

---


nwparker:
Superseded by #10299, which now carries slices 03-06 as one atomic shared-foundation PR. Reason: the `src/shared` modules in these slices are mutually coupled (new bounded readers here depend on types/budgets modified in slice 06, whose own tests depend on these files), so no ordering makes them individually CI-green. Combining them is the honest shape for this layer. No content is lost.

PR #10300, opened 2026-07-24
oom(07/29): bound mobile transport frames, queues & byte ledgers

nwparker:
## OOM hardening slice 07/29 — bound mobile transport frames, queues & byte ledgers

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **07/29** (base: `oom/06-b-shared-wiring`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
This slice re-lands the mobile transport layer's OOM ceilings: bounded inbound/outbound frame queues + memory ledgers, JSON structural/byte admission caps, a bounded per-host connection-log buffer, terminal-snapshot size caps, and retention caps on host-token/credential-cleanup/host-list storage. Ordinary-use behavior is unchanged; the caps sit far above normal RPC/terminal traffic and only engage under abusive input or sustained backpressure.

### ELI5
The phone app that talks to your computer used to accept messages of any size and pile them up with no limit, so a flood (or a bug) could make it run out of memory and crash. This change adds sensible "you can only hold this much at once" limits on incoming messages, outgoing messages, logs, and terminal snapshots. Normal use never hits these limits; only a firehose of abnormally huge data does, and when it does the app closes and reconnects instead of crashing.

### Proof of OOM fix
- mobile-inbound-frame-queue.ts: MAX_FRAME_BYTES=64MiB, MAX_BUFFERED_BYTES=96MiB, MAX_BUFFERED_FRAMES=64 (tests: mobile-inbound-frame-queue.test.ts, mobile-inbound-frame-queue-memory.test.ts)
- mobile-inbound-memory-ledger.ts: process-wide 192MiB / 256 frames retained (test: mobile-inbound-memory-ledger.test.ts)
- mobile-outbound-memory-budget.ts: buffered=16MiB, queued=128MiB, queuedFrames=16384, inFlightFrames=16384, socketSources=64, maxFrame=8MiB (tests: mobile-outbound-memory-budget.test.ts, mobile-outbound-socket-ledger.test.ts)
- mobile-outbound-json.ts: MOBILE_OUTBOUND_MAX_JSON_BYTES=4MiB (test: mobile-outbound-json.test.ts)
- mobile-json-text-admission.ts: structuralTokens=1_000_000, nestingDepth=128 (test: mobile-json-text-admission.test.ts)
- connection-log-buffer.ts: 200 entries/host, 128 hosts, entry=64KiB, host=256KiB, store=8MiB, 16 listeners/host (test: connection-log-buffer.test.ts)
- rpc-client-terminal-binary-frame.ts: frame=256KiB, snapshot=2MiB, aggregate=16MiB, 16 active snapshots, 1024 chunks (test: rpc-client-terminal-binary-frame.test.ts)
- host-credential-cleanup.ts: 256 pending ids, 256KiB storage, 4096 host-id chars, 16 in-flight deletes (test: host-credential-cleanup.test.ts)
- mobile-host-list-storage.ts: 256 entries, 8MiB storage chars; host-store.ts token cache: 64 entries / 64KiB entry (tests: host-store.test.ts)
- websocket-payload-bytes.ts: enforces 64MiB inbound frame limit on all binary payload conversions (tests: websocket-payload-bytes.test.ts, bounded-response-body.test.ts, bounded-process-line-reader.test.ts)
- pairing.ts: PAIRING_INPUT_MAX_CHARACTERS / PAIRING_CODE_MAX_CHARACTERS / device-token max-length caps (test: pairing.test.ts)

### Regressions
- **low**: redactedEndpoint was reimplemented as redacted-websocket-endpoint.ts using URL parsing; it now returns 'unknown' for any non ws:/wss: scheme where the old regex might have matched. Affects only connection-log/console text, no functional path. _(trigger: Log rendering of a malformed or non-websocket endpoint string.)_

- Inbound processing moved from an unbounded Promise chain to a bounded FIFO queue in both rpc-client.ts and mobile-e2ee-v2-physical-channel.ts; FIFO ordering is preserved (single ordered drain). Overflow past 96MiB/64-frame/192MiB caps now closes the socket and schedules reconnect instead of buffering unbounded.
- Outbound sends are rejected (canSend=false -> failOutboundOverflow -> socket close) only when the 16MiB buffered / 128MiB queued / 16384 in-flight ceilings are hit; encryption is deferred until an item reaches the wire so a rejected item never burns an ordered E2EE counter.
- openConnection defers reconnect ('WebSocket reconnect deferred') when >64 retired socket buffers are still draining (canRegisterBufferedAmount=false) — only reachable under rapid reconnect churn.
- Keyed in-flight byte claims (by request id) release on the matching response ACK or, failing that, on socket retire (30s MOBILE_OUTBOUND_SOCKET_RETIRE_TIMEOUT_MS) / close; a request that never gets a response holds its claim until retire, bounded by the in-flight caps.
- Oversized JSON (structuralTokens>1M or depth>128) now throws a capacity error that propagates to the inbound onError and closes the socket, rather than being silently ignored as unparseable.
- Terminal scrollback snapshots larger than 2MiB (or any single stream frame >256KiB, or >1024 chunks, or >16MiB aggregate across snapshots) are rejected: the mobile client deletes the snapshot and emits an error event ('Terminal snapshot exceeded size limit.' / 'Terminal stream frame exceeded size limit.') instead of rendering the content. _(only when: A host emits a very large terminal scrollback snapshot or a single oversized output frame. Reachable in ordinary use only if host-side snapshot sizing (a separate slice) can exceed 2MiB; otherwise overload-only.)_
- Stored host list / pending credential-cleanup / token-cache entries beyond their caps (256 hosts, 256 pending ids, 64 cached tokens) are dropped or evicted oldest-first; persisted host lists exceeding 8MiB or 256 entries fail to parse (return null) and are treated as empty. _(only when: Only with abnormally large persisted AsyncStorage payloads — far above a realistic number of paired hosts.)_

### Build / CI

### Adversarial review
- Adversarial regression pass on slice 07 (mobile transport). I tried hardest to break the prior reviewer's main medium claim (terminal-snapshot 2MiB rejection) and instead REFUTED it: the host-side snapshot producer (src/main/runtime/rpc/methods/terminal.ts:58) caps serialized snapshot data at REQUESTED_SNAPSHOT_BYTE_BUDGET = 2MiB and retries with smaller scrollback row counts (1000/500/250/100/25/0) whenever exceeded, so the mobile receiver cap MOBILE_TERMINAL_SNAPSHOT_MAX_BYTES=2MiB (rpc-client-terminal-binary-frame.ts:16) exactly matches the host budget. Legitimate scrollback never exceeds what mobile accepts. Boundary check (host strict > vs mobile bytes+payload > cap) is consistent; no off-by-one. Host chunks Output/snapshot at 48KiB (TERMINAL_STREAM_CHUNK_BYTES) << 256KiB frame cap and 1024-chunk cap. All other ceilings sit far above ordinary use: inbound 96MiB/64-frame/192MiB, outbound 16MiB/128MiB/16384/8MiB/4MiB JSON, JSON admission 1M structural tokens / depth 128 (~55k+ objects to trip), retention 256 hosts/8MiB, 64 token entries, 256 pending cleanup ids, 128x200 conn-log. Inbound overflow calls fail()->socket close+reconnect (no silent frame drop); single ordered drain preserves FIFO. redactedWebSocketEndpoint is connection-log-only (rpc-client.ts:36). Persisted host-list parse-returns-null only above 256 entries / 8MiB — beyond realistic paired-host counts. No confirmed normal-path visible regression; safe to open as draft PR.

### Files
66 files changed (+3828 / −314).

---


nwparker:

PR #10308, opened 2026-07-24
oom(15/29): bound browser capture & emulator video replay

nwparker:
## OOM hardening slice 15/29 — bound browser capture & emulator video replay

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **15/29** (base: `oom/14-b-ai-vault`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Re-introduces OOM ceilings across the browser/computer-use/emulator/ghostty subsystems: bounded CDP command queues, screenshot/PDF/screencast size+concurrency caps, streaming cookie-store parsers with count/byte/depth limits, a budgeted accessibility-snapshot walk, and bounded scrcpy GOP replay and stdout/line buffers. Behavior below the (generous) ceilings is unchanged; overload paths now reject or shed instead of growing unbounded.

### ELI5
Imagine the app's browser helper used to keep saying "yes" to everything — every screenshot, every giant cookie file, every element on a page — and could eat all the computer's memory. This change adds sensible "that's too much, try again later" limits so a huge page or a flood of requests can't crash the app. For everyday browsing and screenshots you won't notice; the limits only kick in on enormous or abusive inputs.

### Proof of OOM fix
- CDP/agent command queues capped at 64/tab(session) and 512 total (cdp-command-queue.ts, agent-browser-bridge.ts); overflow rejects with browser_busy — proven by cdp-command-queue.test.ts ('full'/'closed' assertions) and cdp-bridge-memory-bounds.test.ts
- Screenshot geometry/bytes capped: 32768px/dimension, 32Mi effective pixels, 32MiB encoded, 2 concurrent captures (browser-screenshot-limits.ts, browser-screenshot-admission.ts) — browser-screenshot-limits.test.ts asserts throw at BROWSER_SCREENSHOT_MAX_EFFECTIVE_PIXELS/MAX_ENCODED_BYTES
- PDF bounded: 2 concurrent prints, 64MiB per PDF, 8 retained streams + 64MiB process retention budget (browser-pdf-admission.ts, cdp-print-to-pdf.ts) — cdp-print-to-pdf.test.ts, browser-pdf-admission.test.ts
- CDP event buffers bounded: 1000 console/network entries, 256 paused requests (excess auto-continued), 1024 iframe sessions (excess detached), 16K/64K text/metadata truncation (cdp-event-memory-bounds.ts) — cdp-event-memory-bounds.test.ts
- Cookie import bounded: JSON 64MiB file / 250k entries / depth 128 / 64MiB retained via streaming jsonc visit (browser-cookie-json-file-parser.ts); Safari store 64MiB/65536 pages/250k cookies (safari-cookie-store-decoder.ts); installed Chromium/Firefox stores 250k cookies/64MiB via SQL aggregate + row iteration (installed-browser-cookie-store-limits.ts) — respective .test.ts files
- Accessibility snapshot budget: 50000 AX nodes, 4096 entries, 1024 name code-units, 1MiB retained names, depth 64 (snapshot-engine.ts) — snapshot-engine.test.ts asserts truncation at SNAPSHOT_MAX_ENTRIES
- scrcpy replay bounded: 120 GOP frames, 32MiB/device, 128MiB total, 16 devices, 8 subscribers (scrcpy-video-registry.ts) — scrcpy-video-registry-memory.test.ts asserts whole-GOP shed then keyframe-only recovery
- CDP ws proxy: 16MiB max incoming payload, 16 max TCP connections (cdp-ws-proxy.ts) — cdp-ws-proxy.test.ts
- Bounded file reads / buffers: ghostty config+theme, serve-sim state (8KiB), permission status (64KiB), macOS native provider line buffer (64MiB), screencast frame = outbound binary frame limit; uiautomator XML 16MiB/50000 elems/depth256/64 attrs (uiautomator-tree.ts) — uiautomator-tree.test.ts
- Concurrency/fanout caps: max 64 active browser downloads (browser-manager.ts), 16 queued computer-sidecar calls (sidecar-client.ts), AVD-name lookup concurrency 4 (android-device-inventory.ts), serve-sim prune scan capped at 4096 entries (serve-sim-runtime-materializer.ts) — serve-sim-runtime-pruning-bounds.test.ts

### Regressions
- **medium**: Accessibility snapshot the agent consumes is now capped at 4096 entries and 50000 AX nodes (names truncated to 1024 code units). On very large/complex pages the agent may no longer see all interactive/text elements it previously did, which can change agent browsing outcomes. _(trigger: Agent browses a page with >4096 interactive+heading+landmark+text elements or >50000 AX nodes (large data tables, very long documents). Below that, identical output.)_
- **low**: Concurrent screenshot captures and PDF prints are limited to 2 each via process-global counters shared across all tabs/worktrees; a 3rd simultaneous capture/print returns browser_busy instead of running. _(trigger: Three or more tabs/worktrees screenshot or print-to-PDF at the exact same moment; the UI/agent must handle browser_busy and retry.)_
- **low**: Screenshot capture and PDF printing now use process-global two-slot admission, so unrelated worktrees contend and excess operations fail with browser_busy rather than waiting or running. _(trigger: A third screenshot-related native operation or PDF print overlaps two unsettled operations anywhere in the process.)_

- Agent/CDP browser commands rejected with browser_busy only after 64 are already queued for one tab or 512 across all tabs
- Screenshots rejected browser_screenshot_too_large only above 32768px/dimension, 32M effective pixels, or 32MiB encoded output
- browser_busy on screenshots/PDF only when >2 captures or >2 prints are simultaneously in flight (process-global counters)
- Cookie import aborts only for files >64MiB, >250k entries, >128 nesting levels, or cookie stores exceeding 250k cookies / 64MiB
- Accessibility snapshot truncates only past 50000 AX nodes / 4096 emitted entries — the agent stops seeing further elements on extreme pages
- scrcpy: subscribe() throws only past 8 total subscribers, register() past 16 devices; late subscriber gets no replay (whole GOP shed) only under sustained high-motion until the next ~10s keyframe
- Live screencast frames silently dropped only when a single frame exceeds the outbound binary-frame byte limit
- CDP request interception silently stops pausing (auto-continues) only past 256 concurrent paused requests or requests with >256 headers / >64K metadata
- uiautomator XML parse fails only above 16MiB / 50000 elements / depth 256 / 64 attributes per element
- scrcpy late-subscriber replay changed from a 120-frame sliding window to shedding the entire buffered GOP as a unit when the cap is exceeded, so a viewer that attaches during sustained high-motion sees nothing until the next keyframe (~10s). subscribe()/register() now throw past 8 subscribers / 16 devices. _(only when: Opening an Android device view during heavy on-screen motion, or opening a 9th concurrent stream viewer / 17th device.)_
- Several new BrowserError codes (browser_screenshot_too_large, browser_busy) and cookie-import rejection reasons are surfaced to the UI; if a caller/UI does not map these they may show generic errors. _(only when: Hitting any screenshot/PDF/queue ceiling or importing an oversized cookie file/store.)_
- Accessibility snapshots truncate names at 1,024 code units and stop after a shared budget of 4,096 emitted entries, 50,000 walked AX nodes, or 1 MiB of retained names. On extreme pages the agent can no longer see later controls or text. _(only when: Agent snapshots a very large document, data table, or iframe set that exceeds one of the snapshot budgets.)_

### Build / CI

### ⚠️ Inherited edge-case defect(s) — pre-existing in #10179
- ⚠️ [medium] `src/main/browser/browser-session-registry.ts` — adversarial-flagged (overload-path only, unverified): hydrateFromPersisted() bypasses the createProfile() 256-profile / 16 KiB-label ceilings; a shape-valid 1 MiB metadata file can hydrate ~5,400 profiles at startup, each installing an Electron session policy.

### Adversarial review
- Verified every numeric ceiling in the slice; all are far above ordinary use. Screenshot geometry 32768px/dimension, 32M effective pixels, 32MiB encoded (browser-screenshot-limits.ts) — Chromium's own texture limits already constrain full-page captures below this; a full-page of a very long page could theoretically approach 32M px but that is extreme. Concurrency caps: 2 screenshots + 2 PDF prints (process-global). CDP command queue 64/tab, 512 total (cdp-command-queue.ts) — closeTab/process correctly decrement queuedCount, no leak, FIFO preserved. Cookie import: 64MiB / 250k entries / depth 128 (browser-cookie-json-file-parser.ts), installed store 250k cookies / 64MiB, Chromium Local State 16MiB, profile scan 10k entries — real cookie stores are a few thousand cookies, Local State well under 1MB. uiautomator XML 16MiB / 50k elements / depth 256 / 64 attrs. ghostty config import 1MB (real configs are KB). Screencast per-frame limit 8MiB (decodeBrowserScreencastImage returns null → drops single oversized frame, stream continues); a viewport JPEG is well under 1MB. scrcpy replay per-device 32MiB / total 128MiB. All confirmed overload-only. The ONLY user/agent-visible normal-path change is the accessibility-snapshot 4096-entry / 50k-node cap (snapshot-engine.ts): below it, output is identical, but large content-heavy pages now truncate silently for the agent. That is a deliberate, generous, already-shipped (passed full CI) design tradeoff, not a correctness bug — flag it for human sign-off but it does not block a draft PR. No confirmed normal-path regression or correctness bug; signoff=true.
- Correctness notes: `/Users/nwparker/orca/workspaces/orca/oom-followup/src/main/browser/browser-session-registry.ts`: Persisted-state hydration bypasses the new 256-profile ceiling and the 16 KiB label ceiling. createProfile() checks both limits at lines 345-352, but hydrateFromPersisted() at lines 481-490 accepts every shape-valid record from the 1 MiB metadata file. A compact valid file can contain roughly 5,400 profiles, causing startup to retain them and install an Electron session policy for each. This is a concrete migration/tampered-state path and is not covered by the persistence limit tests.

### Files
71 files changed (+5002 / −707).

---

### DIFF
--- a/src/main/browser/agent-browser-bridge.test.ts
+++ b/src/main/browser/agent-browser-bridge.test.ts
@@ -1,22 +1,24 @@
 /* eslint-disable max-lines */
 import { describe, it, expect, vi, beforeEach } from 'vitest'
 
-const { execFileMock, webContentsFromIdMock, existsSyncMock, readFileSyncMock, stdinWrites } =
+const { execFileMock, webContentsFromIdMock, existsSyncMock, readScreenshotFileMock, stdinWrites } =
   vi.hoisted(() => ({
     execFileMock: vi.fn(),
     webContentsFromIdMock: vi.fn(),
     existsSyncMock: vi.fn(() => false),
-    readFileSyncMock: vi.fn(() => Buffer.from('')),
+    readScreenshotFileMock: vi.fn(() => Buffer.from('')),
     stdinWrites: [] as string[]
   }))
 
 vi.mock('child_process', () => ({ execFile: execFileMock }))
 vi.mock('fs', () => ({
   existsSync: existsSyncMock,
-  readFileSync: readFileSyncMock,
   accessSync: vi.fn(),
   chmodSync: vi.fn(),
   constants: { X_OK: 1 }
+}))
+vi.mock('./browser-screenshot-file-reader', () => ({
+  readBrowserScreenshotFile: readScreenshotFileMock
 }))
 vi.mock('os', () => ({ platform: () => 'darwin', arch: () => 'arm64' }))
 vi.mock('electron', () => {
@@ -53,10 +55,16 @@ vi.mock('./cdp-bridge', () => ({
 
 import {
   AGENT_BROWSER_CLIPBOARD_WRITE_MAX_BYTES,
+  AGENT_BROWSER_MAX_QUEUED_COMMANDS_PER_SESSION,
   AGENT_BROWSER_TEXT_ARGUMENT_MAX_BYTES,
   AgentBrowserBridge
 } from './agent-browser-bridge'
 import type { BrowserManager } from './browser-manager'
+import {
+  BROWSER_SCREENSHOT_MAX_DIMENSION_PX,
+  BROWSER_SCREENSHOT_MEMORY_LIMIT_ERROR
+} from './browser-screenshot-limits'
+import { CDP_PDF_MAX_RETAINED_BYTES, CDP_PDF_MEMORY_LIMIT_ERROR } from './cdp-print-to-pdf'
 import {
   CLIPBOARD_TEXT_MEASURE_YIELD_CODE_UNITS,
   CLIPBOARD_TEXT_WRITE_MAX_BYTES,
@@ -298,7 +306,7 @@ describe('AgentBrowserBridge', () => {
     stdinWrites.length = 0
     CdpWsProxyMock.instances.length = 0
     existsSyncMock.mockReturnValue(false)
-    readFileSyncMock.mockReturnValue(Buffer.from(''))
+    readScreenshotFileMock.mockReturnValue(Buffer.from(''))
     const wc = mockWebContents(100)
     webContentsFromIdMock.mockReturnValue(wc)
     bridge = new AgentBrowserBridge(mockBrowserManager())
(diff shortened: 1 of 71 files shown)


nwparker:

Issue #17, opened 2026-03-22
I attached a github issue with this worktree, but I have no way of seeing it


AmethystLiang:
<img width="373" height="287" alt="Image" src="https://github.com/user-attachments/assets/6f951a78-9e3d-41de-8144-254380da7eb7" />

PR #10310, opened 2026-07-24
oom(17/29): bound native chat transcript & speech input

nwparker:
## OOM hardening slice 17/29 — bound native chat transcript & speech input

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **17/29** (base: `oom/16-b-codex`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Re-introduces OOM ceilings across native-chat transcript reading (per-record/page/retention byte caps, message-count cap, and a process-wide concurrency+byte admission gate for transcript reads) and speech (cloud-audio sample cap, credential/response/stderr size caps, and bounded model-cache migration + directory scans). The most significant user-visible shift is that native-chat transcript reads are no longer "entire file, no cap" — they now return the newest bounded window with pagination.

### ELI5
Imagine a chat log that could grow to any size. Before, the app tried to load the whole thing into memory at once, which could crash it on huge logs. Now it loads the newest chunk (still very large) and lets you scroll back for older parts, and it only lets a couple of these big loads happen at the same time so memory stays safe. The speech-to-text side gets similar seatbelts: caps on how much audio, error text, and model-folder scanning it will hold. Ordinary users won't notice; only enormous logs or extreme conditions hit the limits.

### Proof of OOM fix
- Transcript retention: MAX_NATIVE_CHAT_TRANSCRIPT_MESSAGES=50,000 and MAX_NATIVE_CHAT_TRANSCRIPT_RETAINED_BYTES=64 MiB (retained-byte estimate = sourceBytes*2+256) in transcript-message-retention.ts; tested in transcript-message-retention.test.ts.
- Per-record cap: MAX_NATIVE_CHAT_TRANSCRIPT_RECORD_BYTES=2 MiB via TranscriptRecordBuffer; tested in transcript-record-buffer.test.ts and transcript-stream-lines.test.ts (oversized record dropped, next line preserved, UTF-8 split across 1-byte chunks intact).
- Per-page cap: MAX_NATIVE_CHAT_TRANSCRIPT_PAGE_RETAINED_BYTES=16 MiB in transcript-tail-reader.ts (sets pageBudgetReached, forces hasMore=true); tested in transcript-reader.test.ts.
- Incremental drain: APPEND_BATCH_RETAINED_BYTE_LIMIT=8 MiB, INCREMENTAL_DRAIN_RETAINED_BYTE_LIMIT=32 MiB in transcript-incremental-reader.ts; tested in transcript-incremental-reader.test.ts (byte-flush before count limit, pause/resume at record boundary without gaps, always advances one record even when budget < record).
- Read admission: MAX_..._READ_CONCURRENCY=8, MAX_..._READ_WAITERS=256, MAX_..._PROCESS_RETAINED_BYTES=128 MiB; each watch drain reserves PAGE(16 MiB)+DRAIN(32 MiB)=48 MiB (transcript-read-admission.ts); tested in transcript-read-admission.test.ts and transcript-watch-drain-admission.test.ts (concurrency+byte budget, aborted/unsubscribed waiter removed and slot reused, idempotent release, listener detach).
- Speech cloud audio: MAX_CLOUD_AUDIO_SAMPLES=16000*600 (10 min) with a single growing Float32Array in openai-transcription-client.ts; tested in openai-transcription-client.test.ts (100k one-sample feeds keep one buffer).
- Speech model-cache migration: 16,384 entries / depth 16 / 16 MiB visited-path bytes in model-cache-path.ts; tested in model-cache-path-bounds.test.ts.
- Speech directory scan: SPEECH_MODEL_ARCHIVE_SCAN_MAX_ENTRIES=32,768, retained-name 4 MiB, SPEECH_MODEL_VOCAB_SCAN_MAX_ENTRIES=4,096 in speech-model-directory-scanner.ts; tested in speech-model-directory-scanner.test.ts.
- tar stderr: SPEECH_MODEL_EXTRACTION_STDERR_MAX_RETAINED_BYTES=64 KiB (4 KiB prefix + ring tail, 500 chars/end evidence) in speech-model-extraction-stderr.ts; tested in speech-model-extraction-stderr.test.ts.
- Credential/response caps: openai-api-key-store.ts uses readIntegrationCredentialFileSync (rejects >1 MiB key file, tested in openai-api-key-store.test.ts); openai-transcription-client.ts uses readFetchResponseJsonWithinLimit for the provider response body.

### Regressions

- Native-chat transcript reads truncate to the newest 50,000 messages / 64 MiB retained (~32 MB source JSONL); older history reachable only via pagination (beforeOffset).
- A single per-page tail read caps at 16 MiB retained and returns hasMore=true; UI must page for the rest.
- Records >2 MiB are dropped (surfaced as a fallback-id placeholder message).
- Admission throws 'Too many queued native chat transcript reads' only when >256 transcript reads are simultaneously queued process-wide.
- Speech model-cache migration aborts best-effort (logs a warning, leaves a partial copy) only past 16,384 entries / depth 16 / 16 MiB path bytes.
- BPE vocab discovery scans at most 4,096 entries; a .vocab beyond that falls back to greedy_search (loses hotwords).
- Cloud dictation still hard-caps at 10 minutes of audio (unchanged limit, now sample-based).
- tar extraction error messages bounded to 4 KiB prefix + 64 KiB tail with an 'N stderr bytes omitted' marker.
- readNativeChatTranscript changed from 'entire transcript, NO message cap' to a bounded newest-window (50,000 messages / 64 MiB retained). For very heavy sessions, oldest messages are no longer in the canonical parse and are only reachable via pagination (the read-cache doc comment explicitly dropped 'canonical, unwindowed'). Confirm all surfaces page correctly. _(only when: Opening a native-chat (Claude/Codex/Grok) session whose transcript exceeds ~32 MB source JSONL or 50k messages; reachable in ordinary heavy coding sessions, which the code comments describe as 'routinely tens of MB'.)_
- Live transcript watcher drains now pass through a shared 128 MiB / 8-concurrency admission gate, each reserving a fixed 48 MiB, so at most ~2 watch drains run concurrently process-wide (shared with page reads from other call sites). With many native-chat views open, live updates serialize behind each other, adding refresh latency. _(only when: Several native-chat sessions actively updating at once (multiple worktrees / paired web/mobile clients). Correctness preserved (FIFO admit, no loss); only latency increases.)_
- readNativeChatTranscript / the read-cache now return a bounded newest window (50,000 messages / 64 MiB retained, ~32 MB source JSONL) instead of the whole file. Below those ceilings the parse is byte-for-byte identical to BASE; above them the oldest history is dropped from the canonical parse and only reachable via pagination (beforeOffset). This is the intended OOM fix, but it IS user-visible for very heavy sessions and warrants explicit human signoff. _(only when: Opening a native-chat session whose transcript exceeds ~32 MB / 50k messages; the code comments describe heavy Claude/Codex sessions as 'routinely tens of MB'.)_
- A tail page can return hasMore=true with fewer than `limit` messages once the 16 MiB per-page retained-byte budget is hit, requiring the UI to page for the rest. Only reachable when records approach the 2 MiB per-record cap (tens of near-max records); ordinary KB-sized messages keep tens of thousands within 16 MiB. _(only when: Paging a transcript composed of many very large (near-2 MiB) records.)_

### Build / CI

### Adversarial review
- 2 skeptic lens(es). Sign-off: **FLAGGED — see above**. Residual risk: **low**.
- Adversarial regression pass on PR #17 (native-chat transcript + speech). I tried hard to refute 'ordinary users see no change' and could not confirm a normal-path regression.\n\nKEY REFUTATION OF THE PRIOR REVIEWER'S MEDIUM FINDING: the claimed 'readNativeChatTranscript went from entire-file to bounded window, older msgs only via pagination' is largely moot. grep -rn shows readNativeChatTranscript / readNativeChatTranscriptCached / decodeTranscriptStream have no wired user-facing caller. All live surfaces use readNativeChatTranscriptTail(File) which ALREADY did newest-first message-count windowing with hasMore/beforeOffset before this PR (src/main/ipc/native-chat.ts default window, MAX 2000; RPC default 40, MAX 2000). The renderer already lazy-pages ('starts at the default window and raises limit to page in older history as it scrolls up'). So the paginated window is pre-existing behavior, not introduced here.\n\nWHAT THIS PR ACTUALLY ADDS on the user path: a per-page byte ceiling (MAX_NATIVE_CHAT_TRANSCRIPT_PAGE_RETAINED_BYTES = 16 MiB retained, ~8 MB source) on the tail reader, and the 48 MiB-reservation admission gate on watch drains. The page cap only makes a heavy page smaller and sets hasMore=true; the existing pagination UI continues loading. transcript-reader.test.ts verifies [...older, ...newest] reconstructs the full ordered list, and that a single record can never single-handedly trip the page budget (records are hard-capped at 2 MiB source => ~4 MiB retained, well under the 16 MiB page budget => >=4 records/page, no starvation).\n\nCEILINGS VERIFIED (all far above ordinary use): retention window 50,000 msgs / 64 MiB retained (estimate = sourceBytes*2+256, ~32 MB source); page cap 16 MiB retained; admission process budget 128 MiB / concurrency 8 / 256 waiters (throws only >256 simultaneous queued reads); cloud dictation 10 min unchanged (now sample-based, MAX_CLOUD_AUDIO_SAMPLES = 16000*600); model-cache migration 16,384 entries / depth 16 / 16 MiB path bytes (a model has a handful of files); BPE vocab scan 4,096 entries; tar stderr 64 KiB (4 KiB prefix + 60 KiB tail); key file 1 MiB.\n\nCORRECTNESS CHECKS (no defect found): incremental drain-budget break resets state.offset to pendingStart and clears pendingRecord so the next drain re-reads that record with no gap; the drainRetainedBytes>0 guard forces one oversized record to advance instead of stalling (both verified by transcript-incremental-reader.test.ts). Admission abort removes queued waiter + detaches listener + re-admits; release is idempotent (transcript-read-admission.test.ts). OfflineAudioChunker findQuietSplitIndex returns Math.max(1, bestIndex) so push() cannot infinite-loop; 100k one-sample test reconstructs exactly. No deadlock in the admission gate since every reservation (16/48 MiB) <= 128 MiB budget.\n\nOnly residual user-visible effect is the low-severity multi-session watch-drain serialization latency noted in confirmedRegressions. Safe to open as a draft PR for human review.

### Files
32 files changed (+1703 / −166).

---

### DIFF
--- a/src/main/native-chat/session-file-resolver.ts
+++ b/src/main/native-chat/session-file-resolver.ts
@@ -3,7 +3,7 @@ import { homedir } from 'node:os'
 import { basename, extname, join } from 'node:path'
 import type { AgentType } from '../../shared/native-chat-types'
 import { resolveNativeChatTranscriptAgent } from '../../shared/native-chat-agent-support'
-import { walkSessionFiles } from '../ai-vault/session-scanner-discovery'
+import { findFirstSessionFile } from '../ai-vault/session-scanner-discovery'
 import { getOrcaManagedCodexHomePath } from '../codex/codex-home-paths'
 import {
   findGrokChatHistoryBySessionId,
@@ -102,11 +102,10 @@ async function resolveClaudeSessionFile(
   projectsDir: string
 ): Promise<string | null> {
   const targetName = `${sessionId}.jsonl`
-  const files = await walkSessionFiles(projectsDir, 'claude', [], {
+  return findFirstSessionFile(projectsDir, {
     extensions: new Set(['.jsonl']),
     filePredicate: (path) => basename(path) === targetName
   })
-  return files[0] ?? null
 }
 
 async function resolveCodexSessionFile(
@@ -120,15 +119,15 @@ async function resolveCodexSessionFile(
     if (!existsSync(sessionsDir)) {
       continue
     }
-    const files = await walkSessionFiles(sessionsDir, 'codex', [], {
+    const file = await findFirstSessionFile(sessionsDir, {
       extensions: new Set(['.jsonl']),
       filePredicate: (path) => {
         const name = basename(path, extname(path))
         return name === sessionId || name.endsWith(`-${sessionId}`)
       }
     })
-    if (files[0]) {
-      return files[0]
+    if (file) {
+      return file
     }
   }
   return null
--- a/src/main/native-chat/transcript-incremental-reader.test.ts
+++ b/src/main/native-chat/transcript-incremental-reader.test.ts
@@ -1,0 +1,185 @@
+import { mkdtemp, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, describe, expect, it } from 'vitest'
+import type { NativeChatMessage } from '../../shared/native-chat-types'
+import {
+  APPEND_BATCH_RETAINED_BYTE_LIMIT,
+  createIncrementalTranscriptState,
+  readIncrementalTranscriptMessages
+} from './transcript-incremental-reader'
+import {
+  estimateTranscriptMessageRetainedBytes,
+  MAX_NATIVE_CHAT_TRANSCRIPT_MESSAGES
+} from './transcript-message-retention'
+import { transcriptFallbackId } from './transcript-fallback-id'
+import { MAX_NATIVE_CHAT_TRANSCRIPT_RECORD_BYTES } from './transcript-tail-reader'
+
+const tempRoots: string[] = []
+
+afterEach(async () => {
+  await Promise.all(tempRoots.splice(0).map((root) => rm(root, { recursive: true, force: true })))
+})
+
+describe('readIncrementalTranscriptMessages', () => {
+  it('keeps the newest shared-retention window for an initial snapshot without batching', async () => {
+    const root = await mkdtemp(join(tmpdir(), 'orca-transcript-incremental-'))
+    tempRoots.push(root)
+    const filePath = join(root, 'transcript.jsonl')
+    const extraMessages = 5
+    await writeFile(
+      filePath,
+      Array.from(
+        { length: MAX_NATIVE_CHAT_TRANSCRIPT_MESSAGES + extraMessages },
+        (_unused, index) => `row-${index}\n`
+      ).join('')
+    )
+    const state = createIncrementalTranscriptState()
+
+    const messages = await readIncrementalTranscriptMessages(
+      filePath,
+      state,
+      (line): NativeChatMessage => ({
+        id: line,
+        role: 'user',
+        blocks: [{ type: 'text', text: line }],
+        timestamp: null,
+        source: 'transcript'
+      })
+    )
+
+    expect(messages).toHaveLength(MAX_NATIVE_CHAT_TRANSCRIPT_MESSAGES)
+    expect(messages[0]?.id).toBe(`row-${extraMessages}`)
+    expect(messages.at(-1)?.id).toBe(
+      `row-${MAX_NATIVE_CHAT_TRANSCRIPT_MESSAGES + extraMessages - 1}`
+    )
+  })
+
+  it('flushes append batches by retained bytes before the message-count limit', async () => {
+    const root = await mkdtemp(join(tmpdir(), 'orca-transcript-incremental-batch-bytes-'))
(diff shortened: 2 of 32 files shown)


nwparker:

PR #10313, opened 2026-07-24
oom(20/29): bound relay transport, listings & log rotation

nwparker:
## OOM hardening slice 20/29 — bound relay transport, listings & log rotation

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **20/29** (base: `oom/19-b-integrations-scm`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Re-lands the OOM ceilings across the entire relay (remote SSH/WSL host) surface: bounds request/stream/client cardinality and retained bytes, converts subprocess and file reads to bounded accumulators, and replaces the fire-and-forget PTY output batcher with a per-client, ack-credited, backpressure-aware broadcaster that pauses the PTY and can evict lagging clients.

### ELI5
Imagine a helper program running on a remote computer that pipes your terminal, files, and git info back to your laptop. Before, if something sent a giant flood of data (or a malicious message), the helper would try to hold all of it in memory and could crash. This change adds "you can only hold this many, and this many bytes" fences everywhere, and teaches the terminal pipe to slow down and wait for your laptop to say "got it" before sending more. The one thing to watch: the terminal now expects your laptop to send those "got it" receipts, so the matching laptop-side change has to ship at the same time or terminals could freeze after a big burst.

### Proof of OOM fix
- client-request-aborts.ts: MAX_ACTIVE_RELAY_REQUESTS_PER_CLIENT=256, MAX_ACTIVE_RELAY_REQUESTS=1024, per-client bytes 32MB, global 64MB; over-limit rejects with RelayRequestAdmissionError. Tested in client-request-aborts.test.ts.
- dispatcher.ts: MAX_RELAY_DISPATCHER_CLIENTS=16, MAX_RELAY_SOCKET_CONNECTIONS=15, MAX_PENDING_RELAY_REQUESTS=256; pending requests rejected on client disconnect/reconnect. Tested in dispatcher.test.ts.
- pty-output-flow-state.ts: PTY_OUTPUT_HIGH_WATER_CHARS=256KB, PTY_OUTPUT_LOW_WATER_CHARS=32KB; producer paused when all clients backpressured, resumed under low-water. Tested in pty-output-broadcast.test.ts.
- pty-persistence-envelope.ts: state<=8MB, field<=64KB, entry<=128KB, retained<=6MB, envToDelete<=1024 keys; JSON structure limits 131072 tokens/depth 8. Tested in pty-persistence-envelope.test.ts.
- git-response-stream.ts: MAX_CONCURRENT_GIT_RESPONSE_STREAMS=MAX_CONCURRENT_STREAMS, MAX_ACTIVE_GIT_RESPONSE_STREAM_BYTES=128MB, MAX_GIT_RESPONSE_STREAM_BYTES=64MB/4096 chunks; UTF-8-safe string chunking avoids a second base64 copy. Tested in git-response-stream-ownership.test.ts, git-response-utf8-chunks.test.ts.
- protocol.ts: MAX_MESSAGE_SIZE=16MB, MAX_BUFFERED_FRAME_CHUNKS=4096; FrameDecoder now stream-discards oversized payloads (13-byte header can no longer pin ~4GiB) and coalesces adversarial tiny chunks. Tested in protocol-handshake.test.ts, protocol-json-encoding-memory.test.ts.
- agent-exec-handler.ts: MAX_CONCURRENT_AGENT_EXECS=8, MAX_OUTPUT_BYTES=4MB via GrowingByteBuffer. Tested in agent-exec-handler.test.ts.
- git-handler-submodule-ops.ts: maxBuffer 10MB, MAX_SUBMODULE_PATHS_PER_REPO=10000, path 64KB/repo 4MB/cache 16MB code-unit caps. Tested in git-handler-submodule-ops.test.ts.
- workspace-session-handler.ts: MAX_PRESENCE_NAMESPACES=256, MAX_PRESENCE_CLIENTS_PER_NAMESPACE=64, snapshot 16MB/1M tokens/depth 128 with oldest eviction. Tested in workspace-session-handler.test.ts.
- relay-watcher-root-capacity.ts: MAX_RELAY_WATCH_ROOTS=20, root-key 64KB, aggregate root-keys 256KB. Tested in relay-watcher-root-capacity.test.ts.
- port-scan-handler.ts: LINUX_PROC_LISTENING_SOCKET_MAX_ENTRIES=2048, proc-net table 8MB, per-file cmdline 64KB (shared linux-proc-port-scan-limits). Tested in port-scan-handler.test.ts.
- fs/quick-open bounds (shared quick-open-listing-limits): QUICK_OPEN_LISTING_MAX_RESULTS=20001, retained paths 100000/32MB, path 64KB; agent-hook status cache 256 panes; git pointer file 64KB; upstream negative-cache key 64KB.

### Regressions
- **low**: fs.listFiles with no maxResults is now always capped at QUICK_OPEN_LISTING_MAX_RESULTS (20001) and runs the primary then ignored pass sequentially instead of an unbounded parallel Promise.all; previously an unbounded request returned every file. _(trigger: Quick Open / file listing over SSH in a repo with >20001 matching files.)_

- Relay rejects the 17th client / 16th socket connection and the 257th per-client (or 1025th global) in-flight request; requires an abusive fan-out of SSH channels far beyond ordinary use.
- git.exec responses over 64MB serialized (or >4096 chunks, or 128MB concurrent retained) are refused with a StreamProtocolError instead of transferred — only reachable by pathological git output.
- Repos with >10000 submodules or >4MB of submodule-path text return an empty submodule list; >2048 listening sockets on a host truncate the port scan; both are far above realistic values.
- PTY producer is paused (terminal output throttled) only when every attached client is at the 256KB unacked high-water and sink-saturated; normal single-fast-client terminals never pause.
- Lagging-client eviction (disconnecting a slow client) only fires with MULTIPLE clients attached to the same PTY when total pending backlog crosses 256KB and another client is caught up; a single client is never evicted.
- Oversized relay frame handling, tiny-chunk coalescing, and stdin/file byte caps only engage on adversarial or 16MB+ inputs.
- PTY output delivery is rewritten to require ack credit: relay now stamps each pty.data with a deliveryToken and only advances a client once it sends pty.ackData{charCount,deliveryToken}. At BASE the SSH client sends pty.ackData with only {id,charCount} and no token, which the new relay rejects (acknowledge() returns early when deliveryToken mismatches). If this relay slice ships without its paired src/main/providers/ssh-pty-output-delivery-credit.ts change, a client that never supplies a valid token accrues unackedChars up to 256KB, the producer is paused, and terminal output freezes. _(only when: SSH/WSL terminal producing >256KB of output when the relay is deployed without the matching main-side delivery-credit slice (landing-order/co-deploy dependency).)_
- When several clients share one PTY (multi-device attach), a client whose backlog keeps total pending output above 256KB while another client is caught up is evicted (disconnected) via dispatcher.evictClient, rather than merely being throttled. Intended anti-HOL-blocking behavior but a visible disconnect for the slow device. _(only when: Two or more clients attached to the same terminal where one falls persistently >256KB behind.)_
- fs.listFiles over the relay (and the paired local path, for parity) now caps a no-maxResults request at QUICK_OPEN_LISTING_MAX_RESULTS (20001). At BASE, callers like OrcaRuntimeFiles.listRuntimeFiles (provider.listFiles(worktree.path,{excludePaths})) and the fs:listFiles IPC passed no maxResults and received the ENTIRE file tree; TARGET truncates to 20001 files in rg traversal order. For a large-but-normal monorepo over SSH (>20k source files after excludes), files beyond 20001 are no longer returned to Quick Open / file listing, so they become unfindable. Genuinely visible below any abusive threshold, but 20001 is far above typical repo source-file counts and the change is deliberate, applied product-wide (local + remote), and passed full CI. A sign-off item, not a defect. _(only when: Quick Open / runtime file listing over SSH (or locally) in a repo with more than 20001 non-excluded files.)_
- PTY output delivery is rewritten from fire-and-forget dispatcher.notify('pty.data') to a per-client ack-credited broadcaster: each pty.data carries a per-client deliveryToken (randomUUID) and PtyOutputBroadcast.acknowledge() ignores any pty.ackData whose deliveryToken does not match (BASE's ackData handler was a no-op). A client accrues unackedChars up to PTY_OUTPUT_HIGH_WATER_CHARS (256KB); once every attached flow is at/over that high-water, updatePtyOutputProducer pauses the PTY and terminal output freezes. Correct and invisible in fully-landed TARGET, but the token-emitting client side lives in a DIFFERENT slice (src/main/providers/ssh-pty-output-delivery-credit.ts, not in this bucket). If the relay slice is deployed ahead of that main-side slice, any SSH/WSL terminal producing >256KB freezes. Relay and main ship as one versioned unit in production, so this is a landing-order/co-deploy caveat, not a fully-integrated regression. _(only when: This relay slice landed/deployed WITHOUT the paired src/main ssh-pty-output-delivery-credit slice, then an SSH/WSL terminal emits >256KB of output.)_
- Multi-device attach: when two+ clients share one PTY and one falls >256KB behind while another is caught up, the lagging client is DISCONNECTED (dispatcher.evictClient via evictLaggingClientsWhenHealthy in pty-output-broadcast.ts), not merely throttled. At BASE an attached client on a slow sink was never disconnected for lag. evictablePtyOutputClientIds treats any client with pending output as lagging, so a briefly-slow second device can be dropped alongside a genuinely stuck one. Guarded so a single client is never evicted. _(only when: Two+ devices attached to the same terminal, one on a slow/saturated sink, while the producer emits >256KB before the slow client drains (e.g. a build flood plus a second device on weak wifi).)_
- This relay slice is protocol-incompatible with the BASE SSH client. `/Users/nwparker/orca/workspaces/orca/oom-followup/src/relay/pty-output-broadcast.ts:74-94` requires a matching `deliveryToken` before returning PTY credit, while BASE `/Users/nwparker/orca/workspaces/orca/oom-followup/src/main/providers/ssh-pty-provider.ts:270-272` emits `pty.ackData` with only `{id,charCount}`. Without the paired target-side `ssh-pty-output-delivery-credit.ts`, an otherwise healthy SSH/WSL terminal reaches 256 KiB unacknowledged output and the relay pauses the PTY indefinitely. This is a confirmed landing/co-deployment dependency, not a speculative concern. _(only when: Land/deploy this relay slice without the paired main-side delivery-credit slice, then produce at least 256 KiB of terminal output.)_

### Build / CI

### Adversarial review
- Verified actual numeric ceilings; almost all are far above ordinary use (overload-only): MAX_RELAY_DISPATCHER_CLIENTS=16 / SOCKET_CONNECTIONS=15, MAX_ACTIVE_RELAY_REQUESTS=1024 (256/client), request bytes 64MB (32MB/client), MAX_CONCURRENT_AGENT_EXECS=8 (short-lived commit/PR-gen execs; hard-fails 'Remote agent execution capacity reached', no queue, but 9+ concurrent on one host is atypical), agent exec MAX_OUTPUT_BYTES=4MB (same truncation semantics as BASE), git response stream 64MB / 128MB concurrent / 4096 chunks, MAX_SUBMODULE_PATHS_PER_REPO=10000, PTY persistence field=64KB/entry=128KB/state=8MB/retained=6MB (metadata only: cwd/paneKey/tabId/env keys, NOT scrollback), MAX_RELAY_PTY_ENV_DELETE_KEYS=1024, presence 64 clients/namespace and 256 namespaces, port scan 2048 sockets, MAX_RELAY_WATCH_ROOTS=20 (count limit is PRE-EXISTING at BASE — only new byte key caps 64KB/256KB were added, path-length overload only). PTY producer pause + lagging-client eviction: verified evictablePtyOutputClientIds returns [] unless (a) backlog >=256KB AND (b) at least one OTHER caught-up sendable client exists, so a single-client terminal is NEVER evicted and never pauses under prompt acking; eviction is strictly a multi-device-attach anti-HOL-blocking behavior. FIFO/ordering preserved (pending[] iterated from index 0, per-client blocked set). Cleanup verified: onClientDetached, removePtyOutputClient, unregister/dispose all resume the producer and cancel drain callbacks. The two normal-path items above (Quick Open 20001 cap; ack-credit co-deploy dependency) are deliberate, CI-passed, and either high-above-typical or manifest only under out-of-order slice landing — worth explicit human sign-off but not blocking a draft PR. No off-by-one, missing-release, or broken-ordering defect found.

### Files
77 files changed (+5152 / −1155).

---

### DIFF
--- a/src/relay/agent-exec-handler.test.ts
+++ b/src/relay/agent-exec-handler.test.ts
@@ -2,6 +2,7 @@ import { execFile, spawn } from 'node:child_process'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import type * as ChildProcess from 'node:child_process'
 import { createFakeChild, createHandlers, requestContext } from './agent-exec-handler-test-harness'
+import { MAX_CONCURRENT_AGENT_EXECS } from './agent-exec-handler'
 import { TERMINAL_GIT_CREDENTIAL_GUARD_POLICY_ENV } from '../shared/terminal-git-credential-guard'
 
 vi.mock('child_process', async (importOriginal) => {
@@ -16,7 +17,7 @@ vi.mock('child_process', async (importOriginal) => {
 const spawnMock = vi.mocked(spawn)
 const execFileMock = vi.mocked(execFile)
 
-type AgentExecResult = { exitCode: number | null; timedOut: boolean }
+type AgentExecResult = { stdout: string; exitCode: number | null; timedOut: boolean }
 
 describe('AgentExecHandler', () => {
   beforeEach(() => {
@@ -62,6 +63,68 @@ describe('AgentExecHandler', () => {
       windowsHide: true
     })
     expect(child.stdin.end).toHaveBeenCalledWith('PROMPT')
+  })
+
+  it('preserves output delivered as 100,000 one-byte fragments', async () => {
+    const child = createFakeChild()
+    spawnMock.mockReturnValue(child as never)
+    const handlers = createHandlers()
+    const pending = handlers.get('agent.execNonInteractive')!(
+      { binary: 'agent', cwd: '/repo', timeoutMs: 5_000 },
+      requestContext()
+    ) as Promise<AgentExecResult>
+
+    for (let index = 0; index < 100_000; index += 1) {
+      child.stdout.emit('data', Buffer.from(index % 2 === 0 ? 'a' : 'b'))
+    }
+    child.emit('close', 0)
+
+    const result = await pending
+    expect(result.stdout).toHaveLength(100_000)
+    expect(result.stdout.slice(0, 4)).toBe('abab')
+    expect(result.stdout.slice(-4)).toBe('abab')
+  })
+
+  it('caps physical agent children until close confirms release', async () => {
+    const children = Array.from({ length: MAX_CONCURRENT_AGENT_EXECS + 1 }, (_, index) => {
+      const child = createFakeChild()
+      child.pid += index
+      return child
+    })
+    let spawnIndex = 0
+    spawnMock.mockImplementation(() => children[spawnIndex++] as never)
+    const handlers = createHandlers()
+    const active = children
+      .slice(0, MAX_CONCURRENT_AGENT_EXECS)
+      .map((_, index) =>
+        handlers.get('agent.execNonInteractive')!(
+          { binary: 'agent', cwd: `/repo-${index}`, timeoutMs: 5_000 },
+          requestContext()
+        )
+      )
+
+    await expect(
+      handlers.get('agent.execNonInteractive')!(
+        { binary: 'agent', cwd: '/overflow', timeoutMs: 5_000 },
+        requestContext()
+      )
+    ).resolves.toMatchObject({ spawnError: 'Remote agent execution capacity reached' })
+    expect(spawnMock).toHaveBeenCalledTimes(MAX_CONCURRENT_AGENT_EXECS)
+
+    children[0]!.emit('close', 0)
+    await active[0]
+    const retry = handlers.get('agent.execNonInteractive')!(
+      { binary: 'agent', cwd: '/retry', timeoutMs: 5_000 },
+      requestContext()
+    )
+    expect(spawnMock).toHaveBeenCalledTimes(MAX_CONCURRENT_AGENT_EXECS + 1)
+
+    for (const child of children.slice(1)) {
+      child.emit('close', 0)
+    }
+    await expect(Promise.all([...active.slice(1), retry])).resolves.toHaveLength(
+      MAX_CONCURRENT_AGENT_EXECS
+    )
   })
 
   it('merges caller-supplied provider environment into the spawned command environment', async () => {
@@ -450,6 +513,8 @@ describe('AgentExecHandler', () => {
       expect(child.stdout.listenerCount('data')).toBe(0)
       expect(child.stderr.listenerCount('data')).toBe(0)
       expect(child.listenerCount('error')).toBe(0)
+      expect(child.listenerCount('close')).toBe(1)
+      child.emit('close', null)
       expect(child.listenerCount('close')).toBe(0)
     } finally {
       vi.useRealTimers()
--- a/src/relay/agent-exec-handler.ts
+++ b/src/relay/agent-exec-handler.ts
@@ -3,12 +3,14 @@ import { existsSync } from 'node:fs'
 import { delimiter, join } from 'node:path'
 import type { RelayDispatcher, RequestContext } from './dispatcher'
 import { applyTerminalGitCredentialPromptGuard } from '../shared/terminal-git-credential-guard'
+import { GrowingByteBuffer } from '../shared/growing-byte-buffer'
 import { mergeGitConfigEnvProtocol } from '../shared/git-credential-prompt-env'
 import { terminateRelaySubprocessTree } from './subprocess-tree-termination'
 
 const DEFAULT_TIMEOUT_MS = 60_000
 const MAX_TIMEOUT_MS = 5 * 60 * 1000
 const MAX_OUTPUT_BYTES = 4 * 1024 * 1024
+export const MAX_CONCURRENT_AGENT_EXECS = 8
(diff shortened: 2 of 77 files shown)


nwparker:

PR #10314, opened 2026-07-24
oom(21/29): bound remote-runtime RPC, transport & waiters

nwparker:
## OOM hardening slice 21/29 — bound remote-runtime RPC, transport & waiters

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **21/29** (base: `oom/20-b-relay`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Re-lands the "B-runtime" slice of the OOM ceilings: bounds the orchestration SQLite query/write layer, the mobile tab-selection store, mobile file-directory listing + image previews, runtime RPC response/admission control, relay control/revoke/host-proof paths, and a family of bounded control-file readers and LRU generation/registry maps. Whole-file changes with matching regression tests; behavior below the ceilings is preserved.

### ELI5
Imagine the app has lots of buckets it fills with data (messages, tasks, open files, network requests). Before, some buckets had no lid, so a flood of junk could fill all the memory and crash the app. This change puts a lid on every bucket with a clearly labeled max size, and adds tests that pour in too much to prove the lid holds. Normal usage stays the same; only extreme or abusive amounts get turned away. A few lids sit low enough that a real person could occasionally bump them — previewing a very high-resolution photo, opening a folder with more than 10,000 files, or running an orchestration with more than 256 tasks — so a human should confirm those thresholds are acceptable.

### Proof of OOM fix
- Orchestration DB: ORCHESTRATION_QUERY_MAX_ROWS=256, ORCHESTRATION_QUERY_MAX_ROW_UTF8_BYTES=2MB, ORCHESTRATION_QUERY_MAX_RETAINED_UTF8_BYTES=8MB, ORCHESTRATION_WRITE_MAX_UTF8_BYTES=512KB, ORCHESTRATION_WRITE_MAX_ITEMS=4096, JSON 64K tokens/depth 64 (query-retention.ts; tests db-query-retention.test.ts, orchestration-query-retention.test.ts)
- OrchestrationMessageWaiterRegistry: 1024 global / 64 per-handle / 64KB handle / 1MB retained-handle-bytes (orchestration-message-waiter-registry.ts + .test.ts)
- Mobile tab selection: 64 clients, 512 worktrees/client, 128 groups/worktree, 4KB id, 64KB/selection, 256KB/client (client-session-tab-selection-persistence.ts; client-session-tab-selection-bounds.test.ts)
- Mobile directory listing: MOBILE_FILE_DIRECTORY_MAX_ENTRIES=10,000 / 4MB retained, classification concurrency 32 (orca-runtime-files.ts; orca-runtime-files.test.ts stops at 10,001)
- Raster preview: MAX_RASTER_IMAGE_PREVIEW_DIMENSION_PX=32,768, MAX_RASTER_IMAGE_PREVIEW_PIXELS=32M (assertRasterImagePreviewWithinLimits; tests reject 32,769-wide bomb)
- Runtime RPC outbound: REMOTE_RUNTIME_MAX_OUTBOUND_JSON_BYTES=4MB → response_too_large (runtime-rpc-response-serialization.ts, dispatcher.ts; dispatcher-output-bounds.test.ts)
- Runtime RPC admission: SHORT_REQUEST_CAP=64 + LONG_POLL_CAP=16 → runtime_busy; structural JSON pre-parse guard (runtime-rpc.ts; runtime-rpc-request-admission.test.ts)
- RuntimeOperationGenerations LRU 8,192 / 64KB key / 4MB retained (pty-lifecycle + worktree-scan gens; runtime-operation-generations.test.ts)
- RuntimeSshRelayRecoveryGenerations 256 / 64KB / 4MB (runtime-ssh-relay-recovery-generations.test.ts)
- Relay: RELAY_CONTROL_MAX_PENDING_REQUESTS=256 / 512B id; RELAY_CONTROL_MAX_MESSAGE_BYTES=64KB; MAX_RELAY_REVOKE_OUTBOX_ITEMS=4096 / 1MB (fails closed on overflow); host-proof exact-base64-length + 16KB ciphertext cap
- Control-file readers: device registry 4096 entries/1MB, E2EE keypair 8KB, TLS PEM 1MB, ws-fallback-port 1KB, runtime metadata bounded read+structure, agent-session key 32B (runtime-control-file-bounds.test.ts, tls-certificate-bounds.test.ts, ws-fallback-port-store.test.ts, relay-revoke-outbox.test.ts)
- Misc runtime maps: terminal file grants 1024, RUNTIME_TEXT_SEARCH_MAX_ACTIVE=16, mobile-file-path-search in-flight cap, RuntimeFileWatcherAdmission 256 leases/16MB/64KB path, resolved-worktree repo concurrency 8, clipboard upload retained cap = CLIPBOARD_IMAGE_MAX_BASE64_CHARS/8 concurrent, recent-pty-output-buffer 64KB window + 1024-chunk cap, Unix socket 1MB frame via GrowingByteBuffer

### Regressions
- **low**: Resolved-worktree scanning is throttled to 8 concurrent repo scans (was unbounded Promise.all), adding latency for users with many (>8) repos while a scan is cold. _(trigger: Listing managed worktrees across dozens of repos with a cold scan cache.)_
- **high**: A process with 64 active long-lived WebSocket subscriptions can starve all normal short RPCs because subscriptions are charged to SHORT_REQUEST_CAP and keep the slots until stream teardown. _(trigger: Keep 64 non-terminal.wait/non-orchestration.check streaming RPCs active across one or more connected clients, then issue any ordinary RPC.)_
- **low**: Orchestration inbox, task, and gate list RPCs now return at most 256 rows and expose total/truncated/remaining metadata. Consumers that previously received the complete list now see only the first page; no pagination path was added in this slice. _(trigger: Use an orchestration run with more than 256 matching messages, tasks, or gates.)_
- **low**: Resolved-worktree scanning is limited to eight repositories at once, preserving order and results but increasing cold-scan latency for installations with many repositories. _(trigger: Cold-list managed worktrees across substantially more than eight repositories.)_

- Orchestration message-waiter, relay control/revoke, device-registry, file-watcher, text-search, mobile-path-search, and generation-map caps all sit far above ordinary single-user/single-device use and fail closed with retry-style errors (runtime_busy / *_limit / capacity errors)
- Oversized (>2MB) legacy orchestration rows are silently skipped from query results and orchestration.check reports {truncated,remaining} instead of hanging; unreachable via normal insert because the 512KB aggregate write cap is below the 2MB read cap
- Poison-message containment: a >2MB body message is undeliverable forever, but can only be created via direct SQL, not the RPC write path
- Coordinator state arrays (completedTasks/failedTasks/escalations) stop growing at 256, but convergence uses exact getTaskStatusCounts() so run outcome stays correct
- RuntimeOperationGenerations LRU eviction only drops an in-flight pty/worktree snapshot when >8,192 distinct owners churn during one operation (returns null, fails safe)
- Clipboard second concurrent upload is refused only when a prior upload already reserves the whole process-wide base64 budget
- Oversized/structurally-amplified control files (metadata, keypair, TLS, registry, outbox) are regenerated or rejected only when corrupt or sparse-inflated
- Mobile/artifact image preview of high-resolution rasters now fails with 'Image dimensions exceed the preview safety limit' when a dimension exceeds 32,768px OR total pixels exceed 32M (~32 megapixels). Modern 48MP phone photos (e.g. 8000x6000 = 48M px) exceed this and can no longer be previewed. _(only when: Previewing a >32MP image via mobile file explorer or terminal-artifact preview; reachable in ordinary use with high-MP camera photos.)_
- Runtime RPC responses larger than 4MB (REMOTE_RUNTIME_MAX_OUTBOUND_JSON_BYTES) are replaced with a small 'response_too_large' error. The previewable-binary path allows up to a 10MB file, whose base64 (~13MB) far exceeds 4MB, and even a ~3MB photo's base64 can cross 4MB — so mobile image/file preview of multi-MB files can now return an error instead of the content. _(only when: Mobile client requesting a file/artifact preview whose base64-encoded JSON response exceeds 4MB; reachable with ordinary multi-MB images.)_
- Mobile file explorer directory listing of a folder with more than 10,000 entries now errors ('This folder is too large to show safely on mobile') instead of listing; also caps at 4MB of retained entry bytes. _(only when: Browsing a very large directory (e.g. node_modules) from a paired mobile device.)_
- Orchestration list/inbox/gate/task RPCs now return at most 256 rows and add {total,truncated,remaining} fields; runs with >256 tasks/messages show a truncated page in the UI (coordinator itself still processes all rows via exact counts and streaming scans). _(only when: Orchestration run/DAG or inbox exceeding 256 tasks or messages.)_
- Inline IMAGE preview of a 3-10MB image on an SSH/remote worktree now returns 'response_too_large' where BASE succeeded. This slice adds boundRuntimeRpcResponse/serializeRuntimeRpcResponse to RpcDispatcher, newly enforcing REMOTE_RUNTIME_MAX_OUTBOUND_JSON_BYTES (4MB) on the dispatcher/unix-socket path (BASE used raw JSON.stringify, uncapped). The image file-size gate is 10MB but base64 of a ~3MB image already exceeds the 4MB JSON cap, so images the user perceives as previewable (under the 10MB gate) fail to render inline. Local (non-SSH) desktop previews are UNAFFECTED (direct window.api.fs.readFile IPC, confirmed in runtime-file-client.ts). Remote downloads are UNAFFECTED (chunked files.readChunk). Fails closed with a clear error. _(only when: Previewing a 3-10MB image file (photo/design asset) inside an SSH-connected remote worktree over the unix-socket runtime transport.)_

### Build / CI
- Touches 7 file(s) also changed on `main` since the revert; git auto-merges them (no conflict): `orca-runtime.test.ts`, `orca-runtime.ts`, `relay-control-client.test.ts`, `relay-control-client.ts`, `relay-http-client.test.ts`, `relay-http-client.ts`, `terminal.ts`.

- ✅ **FIXED** [high, overload-path only] `src/main/runtime/runtime-rpc.ts` — Only terminal.wait & orchestration.check(wait) are long-poll-exempt; terminal.subscribe holds a short-request slot for its lifetime, so 64 subscriptions permanently starve all short RPCs (incl. unsubscribe) with runtime_busy.
  - **Fix applied:** Added a separate STREAM_SUBSCRIPTION_CAP=256 budget: isStreamingSubscription() classifies *.subscribe/*.subscribeAll dispatches, which now draw from activeStreamSubscriptions instead of the 64-slot short budget — so open subscriptions can never starve unsubscribe/short RPCs. Verified: typecheck clean + runtime-rpc tests pass.
- ⚠️ [medium] `src/main/runtime/runtime-rpc.ts` — adversarial-flagged (overload-path only, unverified): status.get device-scope injector JSON.stringify runs AFTER the 4 MiB outbound ceiling check, so a near-limit response can exceed REMOTE_RUNTIME_MAX_OUTBOUND_JSON_BYTES on the wire.

### Adversarial review
- Skeptical pass on PR #21 (B-runtime). Verified numeric ceilings and refuted the prior reviewer's strongest claim.\n\nREFUTED prior claim #2 (medium: '4MB RPC regresses mobile image/file preview'): The 4MB cap (REMOTE_RUNTIME_MAX_OUTBOUND_JSON_BYTES) and its enforcement on the mobile/remote transport (serializeRemoteRuntimePayload, mobile-e2ee-outbound-admission) already exist at BASE — src/shared/remote-runtime-memory-limits.ts and both mobile-e2ee files are byte-identical between BASE and TARGET. So mobile previews >4MB already failed at BASE; this slice does NOT regress the mobile path. The dispatcher change actually improves it by returning a correlated 'response_too_large' error instead of a transport-level throw.\n\nThe ONE genuine new regression this slice introduces: the dispatcher change extends the same 4MB ceiling to the previously-uncapped unix-socket / SSH-tunnel dispatch path (BASE used raw JSON.stringify). Effect is confined to inline IMAGE preview of 3-10MB images on SSH-remote worktrees (local previews bypass runtime RPC via IPC; remote downloads use chunking via files.readChunk). Fails closed with a clear error. Narrow reachability (multi-MB images uncommon in code repos) — low severity, but a below-the-10MB-gate visible change, so normalPathRegressionFound=true.\n\nOther new limits verified as overload-ish / power-user, all far above ordinary use and fail closed:\n- Raster preview cap 32,768px/side & ~33.5M px: 8K (33.2MP) passes; only 48MP+ photos/45MP+ camera exports bite. Header-parse failure also fails a preview, but parser covers standard PNG/JPEG/GIF/WebP/BMP/ICO variants; failure on ordinary images is low-likelihood.\n- Mobile directory listing 10,000 entries / 4MB retained: ordinary repo dirs far below; large asset/dataset/build dirs could hit.\n- Orchestration query cap 256 rows: UI page truncation only. Confirmed coordinator convergence uses exact getTaskStatusCounts() (coordinator.ts:180,211), not the capped state arrays, so run outcomes stay correct — {total,truncated,remaining} added.\n- Resolved-worktree scan concurrency 8: latency only.\n\nNo correctness bugs (watcher-admission release handled on all error paths; classifyRuntimeMobileDirectoryEntries batching preserves order and is re-sorted; db.ts non-null assertions safe via write<read cap invariant).\n\nSignoff=true: safe to open as a draft PR for human review. The single confirmed regression is narrow, fails safe, extends a ceiling the primary remote-preview client (mobile) already enforced at BASE, and matches the exact production code that passed full CI. Human should decide whether the 4MB response cap sitting below the 10MB image-preview file gate warrants raising the cap or lowering the gate on the SSH path.
- Correctness notes: `/Users/nwparker/orca/workspaces/orca/oom-followup/src/main/runtime/runtime-rpc.ts`: Confirmed starvation bug: only terminal.wait and orchestration.check(wait=true) are classified as long polls, while every other WebSocket dispatch consumes the 64-slot short-request budget until dispatchStreaming returns. Long-lived methods such as terminal.subscribe await stream closure, so 64 active subscriptions permanently reject all subsequent short RPCs, including unsubscribe-style cleanup RPCs, with runtime_busy; recovery may require closing sockets or terminals rather than normal RPC cleanup.; `/Users/nwparker/orca/workspaces/orca/oom-followup/src/main/runtime/runtime-rpc.ts`: The status.get device-scope injector runs after the dispatcher has enforced the 4 MiB response ceiling: it parses the already-bounded JSON, adds deviceScope, and calls unbounded JSON.stringify. A response within the added field's size of 4 MiB therefore exceeds REMOTE_RUNTIME_MAX_OUTBOUND_JSON_BYTES on the wire, violating the advertised hard ceiling.

### Files
68 files changed (+4978 / −796).

---

### DIFF
--- a/src/main/runtime/agent-session-claim-identity.test.ts
+++ b/src/main/runtime/agent-session-claim-identity.test.ts
@@ -1,7 +1,11 @@
+import { closeSync, ftruncateSync, mkdtempSync, openSync, rmSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
 import { describe, expect, it } from 'vitest'
 import {
   canonicalizeAgentSessionIdentity,
-  createEphemeralAgentSessionClaimSigner
+  createEphemeralAgentSessionClaimSigner,
+  loadAgentSessionClaimSigner
 } from './agent-session-claim-identity'
 
 describe('agent session claim identity', () => {
@@ -39,5 +43,18 @@ describe('agent session claim identity', () => {
     expect(() =>
       canonicalizeAgentSessionIdentity('blank', { key: 'session_id', id: 'session-1' })
     ).toThrow('agent_session_identity_required')
+  })
+
+  it('rejects an oversized persisted coordination key before reading it', () => {
+    const directory = mkdtempSync(join(tmpdir(), 'orca-claim-key-'))
+    try {
+      const descriptor = openSync(join(directory, 'agent-session-authority.key'), 'w')
+      ftruncateSync(descriptor, 1024 * 1024 * 1024)
+      closeSync(descriptor)
+
+      expect(() => loadAgentSessionClaimSigner(directory, 'profile-1')).toThrow()
+    } finally {
+      rmSync(directory, { recursive: true, force: true })
+    }
   })
 })
--- a/src/main/runtime/agent-session-claim-identity.ts
+++ b/src/main/runtime/agent-session-claim-identity.ts
@@ -1,13 +1,5 @@
 import { createHash, createHmac, randomBytes } from 'node:crypto'
-import {
-  closeSync,
-  mkdirSync,
-  openSync,
-  readFileSync,
-  realpathSync,
-  statSync,
-  writeFileSync
-} from 'node:fs'
+import { closeSync, mkdirSync, openSync, realpathSync, statSync, writeFileSync } from 'node:fs'
 import { dirname, isAbsolute, join, normalize } from 'node:path'
 import {
   AGENT_SESSION_CLAIM_DIGEST_VERSION,
@@ -20,6 +12,7 @@ import {
   type AgentProviderSessionMetadata,
   type ResumableTuiAgent
 } from '../../shared/agent-session-resume'
+import { readNodeFileSyncWithinLimit } from '../../shared/node-bounded-file-reader'
 
 const COORDINATION_KEY_BYTES = 32
 const COORDINATION_KEY_FILE = 'agent-session-authority.key'
@@ -146,7 +139,7 @@ export function loadAgentSessionClaimSigner(
   mkdirSync(dirname(keyPath), { recursive: true })
   let key: Buffer
   try {
-    key = readFileSync(keyPath)
+    key = readNodeFileSyncWithinLimit(keyPath, COORDINATION_KEY_BYTES).buffer
   } catch {
     const candidate = randomBytes(COORDINATION_KEY_BYTES)
     let fd: number | null = null
@@ -155,7 +148,7 @@ export function loadAgentSessionClaimSigner(
       writeFileSync(fd, candidate)
       key = candidate
     } catch {
-      key = readFileSync(keyPath)
+      key = readNodeFileSyncWithinLimit(keyPath, COORDINATION_KEY_BYTES).buffer
     } finally {
       if (fd !== null) {
         closeSync(fd)
--- a/src/main/runtime/claude-agent-teams-shim-env.test.ts
+++ b/src/main/runtime/claude-agent-teams-shim-env.test.ts
@@ -1,4 +1,4 @@
-import { chmod, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
+import { chmod, mkdtemp, readFile, rm, stat, truncate, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, describe, expect, it } from 'vitest'
@@ -23,6 +23,19 @@ describe('claude agent teams shim env', () => {
     await ensureClaudeAgentTeamsShimDir(root)
 
     await expect(readFile(join(root, 'tmux'), 'utf8')).resolves.toContain('agent-teams-tmux "$@"')
+  })
+
+  it('replaces a large sparse shim without reading its payload', async () => {
+    const root = await mkdtemp(join(tmpdir(), 'orca-agent-teams-large-shim-'))
+    roots.push(root)
+    await ensureClaudeAgentTeamsShimDir(root)
+    const shimPath = join(root, 'tmux')
+    await truncate(shimPath, 256 * 1024 * 1024)
+
+    await ensureClaudeAgentTeamsShimDir(root)
+
+    expect((await stat(shimPath)).size).toBeLessThan(64 * 1024)
+    await expect(readFile(shimPath, 'utf8')).resolves.toContain('agent-teams-tmux "$@"')
   })
 
   it('builds native shim env only for direct Claude commands', async () => {
--- a/src/main/runtime/claude-agent-teams-shim-env.ts
+++ b/src/main/runtime/claude-agent-teams-shim-env.ts
@@ -1,4 +1,4 @@
-import { chmod, mkdir, readFile, rename, rm, writeFile } from 'node:fs/promises'
+import { chmod, mkdir, rename, rm, writeFile } from 'node:fs/promises'
 import { accessSync, constants, existsSync } from 'node:fs'
 import { homedir } from 'node:os'
 import { delimiter, dirname, join } from 'node:path'
@@ -8,6 +8,7 @@ import {
   isDirectClaudeCommand,
   type ClaudeAgentTeamsMode
 } from '../../shared/claude-agent-teams-tmux-compat'
+import { nodeFileContentsEqual } from '../../shared/node-file-content-equality'
 import { getOrcaCliCommandNameForPlatform } from '../../shared/orca-cli-command-name'
 
 export type ClaudeAgentTeamsLaunchPlan = {
@@ -136,7 +137,7 @@ function windowsShimScript(): string {
 
 async function writeIfChanged(path: string, content: string): Promise<void> {
   try {
-    if ((await readFile(path, 'utf8')) === content) {
+    if (await nodeFileContentsEqual(path, content)) {
       return
     }
   } catch {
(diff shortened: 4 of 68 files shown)


nwparker:

PR #10315, opened 2026-07-24
oom(22/29): bound filesystem & worktree IPC enumeration

nwparker:
## OOM hardening slice 22/29 — bound filesystem & worktree IPC enumeration

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **22/29** (base: `oom/21-b-runtime`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Slice B-ipc-fs-worktree re-introduces explicit ceilings across main-process filesystem/worktree IPC: it converts unbounded readdir/readFile/string-buffer accumulation into bounded opendir streaming, capped subprocess path accumulators, admission-limited watchers, and budgeted worktree pollers, each with a hard count/byte/depth/concurrency cap plus a regression test.

### ELI5
Imagine Orca has a bunch of doors that used to let in unlimited people — listing every file in a folder, remembering every path a search prints, watching every folder at once. If a folder or repo was gigantic, the app would try to hold all of it in memory and crash. This change puts a bouncer at each door with a clear limit ("no more than X files", "no path longer than Y", "only 8 folders checked at once"). For normal-sized projects nobody notices the bouncer. But a few limits are set low enough that ordinary-but-large cases hit them: a very large repo's Quick Open only shows the first 20,001 files, and a big high-resolution photo (like a 36-megapixel image) refuses to preview.

### Proof of OOM fix
- External import (drag-drop / SSH import / runtime upload) — filesystem-external-import-limits.ts: MAX_SOURCE_PATHS=256, MAX_SOURCE_PATH_BYTES=256KiB, MAX_TREE_ENTRIES=100,000, MAX_TREE_DEPTH=256, MAX_RELATIVE_PATH_BYTES=64KiB, REMOTE_IMPORT_MAX_RETAINED_PATH_BYTES=16MiB, MAX_FILE_BYTES=25MiB, MAX_TOTAL_BYTES=100MiB; tests: filesystem-external-import-limits.test.ts, filesystem-import.test.ts, filesystem-import-ssh-path-safety.test.ts (streams past 100k entries, asserts iterator closed + no dest created).
- Quick Open listing — shared/quick-open-listing-limits.ts wired via filesystem-list-files.ts + filesystem-list-files-git-fallback.ts: MAX_RESULTS=20,001 (now the default even when caller omits maxResults), MAX_RETAINED_PATHS=100,000, MAX_RETAINED_PATH_BYTES=32MiB, MAX_PATH_BYTES=64KiB; buffer accumulation replaced by QuickOpenSubprocessPathAccumulator; tests: filesystem-list-files.test.ts (bounds omitted limit, kills oversized-path scan, abort kills active pass).
- Filesystem watcher admission — filesystem-watcher-admission.ts: MAX_CLAIMS=1,024 global, MAX_CLAIMS_PER_SENDER=256, MAX_RETAINED_IDENTITY_BYTES=16MiB, MAX_PATH_BYTES=64KiB, MAX_CONNECTION_ID_BYTES=8KiB; tests: filesystem-watcher-admission.test.ts, filesystem-watcher.test.ts (rejects at sender cap, recovers after unwatch, stale release ignored).
- Local watcher event stat fan-out bounded — filesystem-watcher.ts: LOCAL_WATCHER_DIRECTORY_STAT_CONCURRENCY=8 and WATCHER_LISTENER_INSTALL_CONCURRENCY=8 via mapWithConcurrency; test: filesystem-watcher.test.ts (peak concurrency == min(count,8) at and above limit).
- WSL watcher snapshot — filesystem-watcher-wsl.ts + wsl-watcher-snapshot.ts: MAX_WSL_SNAPSHOT_FRAME_BYTES=10MiB, MAX_WSL_SNAPSHOT_RETAINED_ENTRIES=50,000; StringDecoder+string buffer replaced by GrowingByteBuffer; tests: wsl-watcher-snapshot.test.ts (null past 50k, dedupes duplicate paths), filesystem-watcher-wsl.test.ts (100k one-byte fragments).
- Worktree base/git-common polling — worktree-polling-scan-budget.ts: MAX_SCANNED_ENTRIES=100,000, MAX_RETAINED_PATHS=16,384, MAX_RETAINED_PATH_BYTES=16MiB, MAX_REPO_CONFIGS=4,096, MAX_RETAINED_REPO_BYTES=4MiB; readdir→opendir(bufferSize:32) streaming + GIT_COMMON_ENTRY_STAT_CONCURRENCY=16; tests: worktree-polling-scan-budget.test.ts, worktree-base-directory-watch-targets-bounds.test.ts (rejects >4,096 repos before hydrating array).
- Git head-identity reader — worktree-head-identity-reader.ts: MAX_GIT_HEAD_METADATA_BYTES=64KiB, MAX_PACKED_REFS_BYTES=16MiB, MAX_PACKED_REFS_ENTRIES=100,000, MAX_LINKED_WORKTREE_ENTRIES=1,024, MAX_IDENTITY_RETAINED_BYTES=4MiB; readdir→opendir with finally-close; test: worktree-head-identity-reader.test.ts (fails closed on oversized HEAD/packed-refs and >1,024 worktrees).
- Worktree common-git pointer — worktree-common-git-directory.ts: MAX_GIT_DIRECTORY_POINTER_BYTES=64KiB via readNodeFileWithinLimit; test: worktree-common-git-directory.test.ts.
- fs:readFile — filesystem.ts: readFile()→readNodeFileWithinLimit(sizeLimit) for text/log/image, plus assertRasterImagePreviewWithinLimits (MAX_DIMENSION_PX=32,768, MAX_PIXELS=33.5M) from raster-image-preview-limits.ts; readDir uses readLocalFilesystemDirectory bounded to FILESYSTEM_DIRECTORY_MAX_ENTRIES=100,000 / 12MiB; text-search buffers → SearchSubprocessLineAccumulator; tests: filesystem.test.ts (dimension-bomb rejected, partial-read reader), filesystem-directory-reader.test.ts.
- SSH bounded provider reads for orca.yaml/.gitignore/issue-command/setup-import — worktrees.ts + worktree-remote.ts via readFilesystemProviderBoundedText (stat-gated before readFile); test: worktrees.test.ts (exact-size admitted, +1 byte rejected before reading).

### Regressions
- **medium**: WSL file watching now fails closed for large worktrees: a snapshot frame with more than 50,000 distinct paths makes parseWslSnapshotFrame return null, which on the first snapshot settles startup with an error (watcher never starts) and afterward drops to coarse overflow refresh. Previously such a snapshot parsed successfully and delivered granular events. _(trigger: Watching a WSL-hosted worktree whose recursive snapshot contains >50,000 files (large repo on WSL). Under 50,000 entries behavior is unchanged.)_
- **low**: Quick Open's two listing passes (tracked/primary + ignored) changed from PARALLEL to SEQUENTIAL on the default no-maxResults path. BASE ran Promise.all([primary, ignored]) when maxResults was undefined; TARGET's default result limit is now 20,001 (not undefined), so the sequential branch (await primary; if under budget await ignored) is always taken for ordinary repos. For repos under 20,001 files the result SET is identical and complete, but scan latency rises by the serialized primary-pass duration and empty-query result ordering becomes deterministic (primary paths first) instead of the previous race-order interleave. Applies to both the rg path (filesystem-list-files.ts) and the git fallback (filesystem-list-files-git-fallback.ts). _(trigger: Opening Quick Open in any repository (rg available or git fallback). Below 20,001 files this is the normal path; results stay complete, only latency/order shift.)_
- **low**: Quick Open now runs the primary and ignored-file scans serially rather than concurrently when no explicit result limit is supplied. Even below all ceilings, repositories with ignored files wait for primary scan completion before the ignored scan starts, changing inventory latency and insertion/tie ordering from BASE's parallel merge. _(trigger: Open Quick Open in any repository with ignored files while the caller omits maxResults; this is the ordinary desktop IPC path.)_
- **medium**: Quick Open completeness is capped at 20,001 paths when no limit is supplied; later tracked files and the ignored pass are omitted once the primary pass fills the budget. _(trigger: Open Quick Open in a repository with more than 20,001 qualifying primary paths.)_
- **medium**: WSL snapshot watching rejects more than 50,000 distinct retained paths. Initial overflow prevents watcher startup; later overflow emits only a coarse refresh instead of granular events. _(trigger: Watch a WSL worktree whose depth-2 snapshot contains more than 50,000 distinct paths.)_

- readDir (File Explorer) throws FILESYSTEM_DIRECTORY_LIMIT_MESSAGE only for a single directory with >100,000 entries or >12MiB of retained names.
- External import fails only for >256 top-level source paths, a tree >100,000 entries, depth >256, a single relative path >64KiB, retained paths >16MiB, a file >25MiB, or a batch >100MiB.
- readGitCommonHeadIdentities returns [] (no head freshness) only for repos with >1,024 linked worktrees, packed-refs >16MiB / >100,000 entries, or a single metadata line >64KiB.
- Worktree base/git-common pollers fail to start at boot, or pause+keep-last-snapshot mid-run, only when a watched base/worktrees dir exceeds 100,000 scanned entries or the workspace exceeds 4,096 repo configs.
- Per-sender watcher admission rejects fs:watchWorktree only after 256 distinct simultaneous watches from one renderer (1,024 global).
- Quick Open path scan is killed only when a single emitted path exceeds 64KiB or the retained set exceeds 100,000 paths / 32MiB.
- WSL frame-byte overflow (>10MiB) behavior is unchanged from before (still fails/refreshes); the newly-added 50,000-entry cap is the only genuinely new WSL trigger.
- Quick Open file inventory is now hard-capped at 20,001 results even when the caller passes no limit (previously an unbounded scan merged both the tracked and ignored passes). In a very large monorepo (>20,001 tracked files) Quick Open sees only the first 20,001 paths in git-ls-files/rg order, so some files become unreachable via Quick Open; the ignored-file pass is also skipped once the primary pass alone fills the budget. _(only when: Opening Quick Open in a repository whose tracked file count exceeds ~20,001 (large enterprise/monorepo). Below that, results and completeness are unchanged (only serialized rather than parallel).)_
- fs:readFile now rejects raster image previews whose dimensions exceed 32,768px on an axis OR whose total area exceeds ~33.5 megapixels (MAX_RASTER_IMAGE_PREVIEW_PIXELS), throwing 'Image dimensions exceed the preview safety limit' instead of showing the image. High-resolution photos (e.g. a 6000x6000 / 36MP or 48MP phone image) can no longer be previewed in the editor pane. _(only when: Opening a PNG/JPEG/WebP raster image larger than ~33.5MP (or >32,768px per side) in the editor. PDFs and SVGs are unaffected.)_
- External import (drag-drop and SSH/runtime upload) now rejects a batch of more than 256 top-level source paths with 'External import accepts at most 256 source paths' before touching the filesystem, and fails a directory import if its tree exceeds 100,000 entries or 256 nesting levels. _(only when: Multi-selecting/dropping >256 items at once, or importing a directory tree with >100,000 files or >256 nested levels. Matches the existing native-file-drop cap of 256.)_
- Worktree head-freshness display goes empty (returns []) rather than partial for pathological repos, and background worktree base-directory watchers are skipped entirely when a workspace has more than 4,096 configured repos. _(only when: >1,024 linked worktrees / >16MiB packed-refs, or >4,096 repos in one workspace.)_
- Quick Open (fs:listFiles) is now hard-capped at 20,001 results. The IPC handler at filesystem.ts:1072 always calls listQuickOpenFiles with no maxResults, and resolveQuickOpenResultLimit(undefined) returns QUICK_OPEN_LISTING_MAX_RESULTS=20,001 (src/shared/quick-open-listing-limits.ts). Previously maxResults===undefined ran both the tracked and ignored rg passes in parallel with no cap and returned every path. Now the primary pass runs first and the ignored-file pass is skipped once files.size>=20,001, so in a repo with >20,001 reachable files some paths become unreachable via Quick Open and ignored files may never be listed. Below the cap results are identical (only serialized instead of parallel). Input-gated above a deliberate cap. _(only when: Opening Quick Open in a repo whose reachable (tracked+ignored, non-excluded) file count exceeds ~20,001 - i.e. large enterprise/monorepo. Median-size repos are far below this.)_

### Build / CI
- Touches 1 file(s) also changed on `main` since the revert; git auto-merges them (no conflict): `worktree-remote.ts`.

- ✅ **FIXED** [medium, overload-path only] `src/main/ipc/filesystem-watcher.ts` — fs:watchWorktree claims admission before installRemoteWatcher; the SSH-provider-unavailable early return skips registerSenderCleanup, so a destroyed WebContents never releases the claim. Repeated SSH reloads during provider unavailability can exhaust the 1,024 global claim cap.
  - **Fix applied:** Hoisted registerSenderCleanup(event.sender) to immediately after the admission claim (before the first await), so the provider-unavailable early return, renderer-destroy-mid-install, and retry all release the claim via releaseSender on WebContents destroy. Verified: typecheck + filesystem-watcher suite pass.

### Adversarial review
- All three prior-reviewer medium claims are CONFIRMED in code, but all are input-gated ABOVE deliberate caps and none is a correctness bug or median-user regression; each fails gracefully (clear error message, or coarse-refresh degradation) rather than crashing or losing data.\n\nBelow-the-limit behavior for ordinary users is unchanged: the only sub-cap Quick Open change is parallel->serial pass ordering (identical results, marginal latency), so normalPathRegressionFound=false.\n\nCap inventory verified: Quick Open 20,001 results / 100,000 retained paths / 32MiB / 64KiB per path; raster preview 32,768px/axis + 33.5MP; WSL snapshot 50,000 non-ignored entries (node_modules excluded via ignoreDirs) + 10MiB stream buffer; readDir single-directory 100,000 entries / 12MiB names; external import 256 source paths (== existing NATIVE_FILE_DROP_MAX_PATHS) / 100,000 tree entries / depth 256 / 64KiB relative path / 16MiB retained / 25MiB per file / 100MiB batch; watcher admission 1,024 global / 256 per-sender / 16MiB retained; head-identity 1,024 worktrees / 16MiB packed-refs / 100,000 entries / 64KiB per line. Each cap has a co-located regression test (filesystem-directory-reader.test.ts, filesystem-external-import-limits.test.ts, filesystem-watcher-admission.test.ts, worktree-head-identity-reader.test.ts, worktree-polling-scan-budget.test.ts, wsl-watcher-snapshot.test.ts, worktree-base-directory-watch-targets-bounds.test.ts).\n\nCleanup on abort/error/timeout in filesystem-list-files.ts is thorough: abort listener registered {once:true} and removed in finally, killSurvivors() on rejection, child.kill() on abort/limit, and stdout/stderr/error/close listeners detached in finish(). No leak or double-free found. QuickOpenSubprocessPathAccumulator.clear() is called on every abnormal path (error/signal/timeout/too-large) so no residual-buffer path leak.\n\nResidual risk is medium driven by the Quick Open 20,001 cap: for a whole-repo fuzzy file finder, 20,001 is defensibly low for enterprise monorepos (some legitimate large repos exceed it and lose file reachability). Signoff=true because this is CI-passed production code, the caps are the intended OOM fix, there is no correctness defect, and human review of the specific ceilings (especially 20,001 for Quick Open and 33.5MP for image preview) is the appropriate next step. Recommend the human explicitly accept those two thresholds or consider raising/making them configurable.
- Correctness notes: `/Users/nwparker/orca/workspaces/orca/oom-followup/src/main/ipc/filesystem-watcher.ts`: A watcher admission claim can survive renderer teardown when an SSH filesystem provider is unavailable. `fs:watchWorktree` claims capacity before calling `installRemoteWatcher`; the provider-unavailable path returns before `registerSenderCleanup` is installed, and `scheduleRemoteWatcherRetry` does not register it. If that WebContents is then destroyed, the retry drops the destroyed listener but nothing calls `watcherAdmission.releaseSender`, permanently retaining the claim until global shutdown. Repeated SSH renderer reloads during provider unavailability can eventually exhaust the 1,024 global claim cap and reject otherwise valid watches. The relevant paths are the claim at line 1374, unavailable retry at line 1396, early return at line 1148, and cleanup registration only on successful/in-flight ownership at lines 1043-1045/1273-1275.

### Files
50 files changed (+3285 / −969).

---

### DIFF
--- a/src/main/ipc/filesystem-directory-reader.test.ts
+++ b/src/main/ipc/filesystem-directory-reader.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, it } from 'vitest'
+import { FILESYSTEM_DIRECTORY_LIMIT_MESSAGE } from '../../shared/filesystem-directory-listing-limit'
+import { collectLocalFilesystemDirectoryEntries } from './filesystem-directory-reader'
+
+type TestEntryOptions = {
+  name: string
+  directory?: boolean
+  symlink?: boolean
+  onClassify?: () => void
+}
+
+function entry(options: TestEntryOptions) {
+  return {
+    name: options.name,
+    isDirectory: () => {
+      options.onClassify?.()
+      return options.directory ?? false
+    },
+    isSymbolicLink: () => options.symlink ?? false
+  }
+}
+
+describe('local filesystem directory reader', () => {
+  it('preserves directory-first sorting and keeps symlinks file-like', async () => {
+    const source = [
+      entry({ name: 'z.txt' }),
+      entry({ name: 'linked-dir', directory: true, symlink: true }),
+      entry({ name: 'beta', directory: true }),
+      entry({ name: 'alpha', directory: true })
+    ]
+
+    await expect(collectLocalFilesystemDirectoryEntries(source)).resolves.toEqual([
+      { name: 'alpha', isDirectory: true, isSymlink: false },
+      { name: 'beta', isDirectory: true, isSymlink: false },
+      { name: 'linked-dir', isDirectory: false, isSymlink: true },
+      { name: 'z.txt', isDirectory: false, isSymlink: false }
+    ])
+  })
+
+  it('stops before classifying or retaining the first over-limit entry', async () => {
+    let enumerated = 0
+    let overLimitClassified = false
+    const source = {
+      async *[Symbol.asyncIterator]() {
+        for (const value of [
+          entry({ name: 'one' }),
+          entry({ name: 'two' }),
+          entry({ name: 'three', onClassify: () => (overLimitClassified = true) })
+        ]) {
+          enumerated += 1
+          yield value
+        }
+      }
+    }
+
+    await expect(
+      collectLocalFilesystemDirectoryEntries(source, {
+        maxEntries: 2,
+        maxRetainedBytes: 1024
+      })
+    ).rejects.toThrow(FILESYSTEM_DIRECTORY_LIMIT_MESSAGE)
+    expect(enumerated).toBe(3)
+    expect(overLimitClassified).toBe(false)
+  })
+})
--- a/src/main/ipc/filesystem-directory-reader.ts
+++ b/src/main/ipc/filesystem-directory-reader.ts
@@ -1,0 +1,41 @@
+import { opendir } from 'node:fs/promises'
+import type { DirEntry } from '../../shared/types'
+import {
+  createFilesystemDirectoryLimitState,
+  trackFilesystemDirectoryEntry,
+  type FilesystemDirectoryListingLimits
+} from '../../shared/filesystem-directory-listing-limit'
+
+type LocalDirectorySourceEntry = {
+  name: string
+  isDirectory(): boolean
+  isSymbolicLink(): boolean
+}
+
+export async function readLocalFilesystemDirectory(dirPath: string): Promise<DirEntry[]> {
+  return collectLocalFilesystemDirectoryEntries(await opendir(dirPath))
+}
+
+export async function collectLocalFilesystemDirectoryEntries(
+  directory: AsyncIterable<LocalDirectorySourceEntry> | Iterable<LocalDirectorySourceEntry>,
+  requestedLimits?: Partial<FilesystemDirectoryListingLimits>
+): Promise<DirEntry[]> {
+  const entries: DirEntry[] = []
+  const limit = createFilesystemDirectoryLimitState(requestedLimits)
+  for await (const entry of directory) {
+    trackFilesystemDirectoryEntry(limit, entry)
+    const isSymlink = entry.isSymbolicLink()
+    entries.push({
+      name: entry.name,
+      // Why: avoid probing macOS TCC-protected symlink targets during listing.
+      isDirectory: !isSymlink && entry.isDirectory(),
+      isSymlink
+    })
+  }
+  return entries.sort((left, right) => {
+    if (left.isDirectory !== right.isDirectory) {
+      return left.isDirectory ? -1 : 1
+    }
+    return left.name.localeCompare(right.name)
+  })
+}
--- a/src/main/ipc/filesystem-external-import-limits.test.ts
+++ b/src/main/ipc/filesystem-external-import-limits.test.ts
@@ -1,0 +1,147 @@
+import { describe, expect, it } from 'vitest'
+import {
+  admitExternalImportTreeEntry,
+  assertExternalImportSourcePaths,
+  assertExternalImportTreeDepth,
+  captureRuntimeUploadRetentionCheckpoint,
+  createExternalImportTreeBudget,
+  createRuntimeUploadRetentionBudget,
+  EXTERNAL_IMPORT_MAX_RELATIVE_PATH_BYTES,
+  EXTERNAL_IMPORT_MAX_SOURCE_PATH_BYTES,
+  EXTERNAL_IMPORT_MAX_SOURCE_PATHS,
+  EXTERNAL_IMPORT_MAX_TREE_DEPTH,
+  EXTERNAL_IMPORT_MAX_TREE_ENTRIES,
+  ExternalImportCapacityError,
+  REMOTE_IMPORT_MAX_FILE_BYTES,
+  REMOTE_IMPORT_MAX_RETAINED_PATH_BYTES,
+  REMOTE_IMPORT_MAX_TOTAL_BYTES,
+  restoreRuntimeUploadRetentionCheckpoint,
+  retainRuntimeUploadFileBytes
+} from './filesystem-external-import-limits'
+
+describe('external filesystem import limits', () => {
+  it('accepts the exact source-count boundary and rejects the next path', () => {
+    expect(() =>
+      assertExternalImportSourcePaths(
+        Array.from({ length: EXTERNAL_IMPORT_MAX_SOURCE_PATHS }, () => '')
+      )
+    ).not.toThrow()
+    expect(() =>
+      assertExternalImportSourcePaths(
+        Array.from({ length: EXTERNAL_IMPORT_MAX_SOURCE_PATHS + 1 }, () => '')
+      )
+    ).toThrow('External import accepts at most 256 source paths')
+  })
+
+  it('accepts the exact source-path byte boundary and rejects one byte more', () => {
+    expect(() =>
+      assertExternalImportSourcePaths(['a'.repeat(EXTERNAL_IMPORT_MAX_SOURCE_PATH_BYTES)])
+    ).not.toThrow()
+    expect(() =>
+      assertExternalImportSourcePaths(['a'.repeat(EXTERNAL_IMPORT_MAX_SOURCE_PATH_BYTES + 1)])
+    ).toThrow('External import source paths exceed 256 KiB')
+  })
+
+  it('accepts exactly 100,000 tree entries without retaining their paths', () => {
+    const budget = createExternalImportTreeBudget()
+    for (let index = 0; index < EXTERNAL_IMPORT_MAX_TREE_ENTRIES; index += 1) {
+      admitExternalImportTreeEntry(budget, 'entry', false)
+    }
+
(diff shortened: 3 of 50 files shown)


nwparker:

PR #10316, opened 2026-07-24
oom(23/29): bound PTY/runtime/remote IPC delivery

nwparker:
## OOM hardening slice 23/29 — bound PTY/runtime/remote IPC delivery

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **23/29** (base: `oom/22-b-ipc-fs-worktree`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
This slice re-lands the OOM ceilings for the IPC PTY-delivery, remote-runtime, SSH, and terminal-preview surfaces: it bounds the PTY renderer-delivery credit ledger and pending-output maps, caps registered SSH providers/subscriptions/credential prompts/preview streams, and (structurally) reroutes SSH/remote PTY output through the same bounded batched renderer-delivery pipeline used for local PTYs.

### ELI5
Imagine the app is a mailroom that forwards terminal output to the screen. Before, if a firehose of text came in (from a remote machine, a giant folder listing, or a runaway program) the mailroom kept stacking mail forever until it ran out of room and crashed. This change puts labeled shelves with hard size limits everywhere, and when a shelf fills it throws away the extra and just re-photographs the whole screen instead of remembering every scrap. It also makes remote (SSH) terminals use the exact same shelves the local ones already use, so both are protected the same way.

### Proof of OOM fix
- PTY renderer-delivery credit ledger caps (src/main/ipc/pty-renderer-delivery-credit.ts): MAX_PENDING_PTY_DELIVERY_CREDIT_SPANS=4096, MAX_RENDERER_PTY_DELIVERY_CREDIT_SPANS_PER_PTY=4096, MAX_RENDERER_PTY_DELIVERY_CREDIT_SPANS=16384, MAX_RENDERER_PTY_DELIVERY_CREDIT_STATES=4096, MAX_...STATE_ID_BYTES=8MB. Overflow collapses spans into settled metadata and acknowledges upstream credit so no flow-control stall. Tests: pty-renderer-delivery-credit.test.ts (flood-collapse, per-PTY cap, aggregate cap, state-count cap, id-byte cap, each settled exactly once).
- Pending PTY-data retention caps (src/main/ipc/pty-renderer-delivery-retention.ts): MAX_PENDING_PTY_DATA_STATES=4096, MAX_PENDING_PTY_DATA_ID_BYTES=8MB, MAX_PENDING_PTY_DATA_CHARS=32MB, MAX_PENDING_PTY_DATA_CREDIT_SPANS=16384, MAX_RENDERER_PTY_DELIVERY_ACCOUNTING_STATES=4096. Rejection settles credit + emits pending-cap restore marker (pane repaints from snapshot). Tests: pty-renderer-delivery-retention.test.ts.
- MAX_REGISTERED_SSH_PTY_PROVIDERS=256 in src/main/ipc/pty.ts; registerSshPtyProvider throws 'ssh_pty_provider_capacity'. listProcesses/reconcile now stream via PtyProcessListAdmission + visitPtyProcessListingsInBatches. Test: pty.test.ts ('fails claimed ensure closed before aggregate owner listings amplify memory').
- Remote-runtime subscription admission (src/main/ipc/runtime-environment-subscription-admission.ts): REMOTE_RUNTIME_MAX_SUBSCRIPTIONS=256, REMOTE_RUNTIME_MAX_RETAINED_SUBSCRIPTION_BYTES=16MB, REMOTE_RUNTIME_MAX_SUBSCRIPTION_ID_BYTES=4KB, per-param 1MB. Claim/release refcount, settled once. Tests: runtime-environment-subscription-admission.test.ts.
- Runtime transport-generation refactored to leased entries (src/main/ipc/runtime-environment-transport-generation.ts) — no tombstone accumulation under unique-environment churn. Tests: runtime-environment-transport-generation.test.ts (10k churn -> 0 tracked).
- SSH credential prompts (src/main/ipc/ssh-passphrase.ts): SSH_MAX_PENDING_CREDENTIAL_REQUESTS=64, SSH_CREDENTIAL_VALUE_MAX_UTF8_BYTES=64KB, detail clamped to SSH_CREDENTIAL_DETAIL_MAX_UTF8_BYTES; oversized targetId rejected without allocating state. Tests: ssh-passphrase.test.ts.
- SSH remote directory browse (src/main/ipc/ssh-browse.ts): FILESYSTEM_DIRECTORY_MAX_RETAINED_BYTES=12MB and FILESYSTEM_DIRECTORY_MAX_ENTRIES=100000; stdout accumulated via GrowingByteBuffer (no quadratic string concat), stderr via bounded tail. Tests: ssh-browse.test.ts (byte-limit reject+channel close, 65k tiny chunks, split-UTF8 filename).
- SSH connection generations (src/main/ipc/ssh.ts): MAX_TRACKED_SSH_CONNECTION_GENERATIONS=4096 via lease; ssh:connect rejects (cap+1)th target with 'SSH connection generation target capacity exhausted'; state.error clamped via clampSshConnectionError. Tests: ssh.test.ts.
- Terminal preview aggregate caps (src/main/ipc/terminal-preview.ts): TERMINAL_PREVIEW_MAX_ENTRIES_PER_CONTENTS=64, TERMINAL_PREVIEW_MAX_ENTRIES_TOTAL=256 for both output streams and fit claims. Tests: terminal-preview-admission.test.ts.
- Terminal preview output stream (src/main/ipc/terminal-preview-output-stream.ts): TERMINAL_PREVIEW_PENDING_MAX_RECORDS=4096 record cap in addition to byte caps; batch chunks compacted via appendCompactedStringChunk (RETAINED_STRING_CHUNK_LIMIT). Tests: terminal-preview-output-stream-memory.test.ts.
- Repo icon picker (src/main/ipc/shell-repo-icon-picker.ts): readNodeFileWithinLimit(MAX_REPO_ICON_UPLOAD_BYTES=256KB) with post-stat re-check, plus assertRasterImagePreviewWithinLimits (max 32768px dimension / 32M pixels). Tests: shell.test.ts.
- Render-desync evidence pruning (src/main/ipc/terminal-render-desync-evidence.ts): streamed via opendir with incremental capture eviction and per-capture byte accounting clamped to MAX_EVIDENCE_BYTES+1 (bounds pruning-time memory).

### Regressions
- **low**: Remote workspace session patches are now applied at most 4 targets concurrently (REMOTE_WORKSPACE_PATCH_CONCURRENCY) instead of all-at-once via Promise.all. With many connected SSH targets, propagation of a workspace change is slightly more serialized. Correctness is unchanged (all targets still patched). _(trigger: More than 4 connected remote targets receiving a workspace patch.)_
- **high**: At credit-ledger state pressure, ACK ordering can release a newly sent remote/local PTY credit span before the renderer has processed it. _(trigger: More than 4,096 distinct PTYs retain in-flight delivery credit, an evicted PTY ID sends again, and the renderer then ACKs bytes from that ID's older delivery generation.)_

- Registered SSH PTY provider #257 throws ssh_pty_provider_capacity (ordinary use is 1-3 SSH connections).
- 257th concurrent remote-runtime subscription, or >16MB aggregate retained params, rejected as remote_runtime_busy.
- 65th concurrent pending SSH credential prompt returns null (auth fails); a submitted passphrase >64KB is rejected as null.
- 4097th distinct tracked SSH connection-generation target rejects ssh:connect (leases normally release on teardown, so this only trips under thousands of live/leaked targets).
- 65th terminal preview stream per pop-out window (or 257th total) silently returns {snapshot:null, replay:[]}; 65th fit claim returns null.
- Per-PTY pending output above 32MB (or 16384 credit spans / 4096 states / 8MB of ids) is dropped to a restore-sentinel; pane repaints from the main-owned snapshot instead of OOMing.
- A stalled terminal-preview renderer accumulating >4096 pending batches (or >256KB) is force-resynced from a fresh snapshot.
- SSH/remote PTY output is now routed through the same shared, batched, flow-controlled renderer-delivery pipeline as local PTYs (ssh.ts now calls routeExternalPtyData/Exit/Replay -> pty.ts routeProviderData) instead of the old direct runtime.onPtyData + webContents.send path. This changes remote terminal delivery timing/batching and subjects remote output to the pending-data caps and producer flow-control credit. It is exercised on every SSH terminal, not just under overload, so it warrants explicit sign-off even though CI passed. _(only when: Any SSH/remote-runtime terminal session (ordinary use).)_
- Browsing a remote SSH directory that emits more than 100,000 entries or >12MB of listing bytes now rejects with FILESYSTEM_DIRECTORY_LIMIT_MESSAGE (and closes the channel) instead of loading; previously the listing was unbounded. Users on servers with very large directories (e.g. huge maildirs, /nix/store) see an error toast rather than a partial/complete listing. _(only when: SSH file browser opening a directory with >100k entries or >12MB of names.)_
- Empty terminal-preview data records (append('')) are now dropped early, and repo-icon PNGs are additionally rejected if their raster dimensions exceed 32768px / 32M pixels even when under the 256KB byte cap. Both are edge cases unlikely to affect real content. _(only when: Empty preview chunks; or a >32768px-dimension icon PNG under 256KB.)_
- Repo icon uploads are now hard-capped at 256KB (MAX_REPO_ICON_UPLOAD_BYTES in src/shared/repo-icon.ts) and 400KB data-URL length. A user who picks a large image (e.g. a phone photo) as a custom repo/worktree icon now gets 'Repo icon image must be 256KB or smaller.' instead of the previously-unbounded load. Clear error, not silent; icons are meant to be small, so this is a benign edge. _(only when: Choosing an image >256KB as a custom repo icon via shell-repo-icon-picker.)_
- SSH file-browser listings for a single directory with >100,000 entries or >12MB of names now reject with FILESYSTEM_DIRECTORY_LIMIT_MESSAGE and close the channel (ssh-browse.ts, limits from src/shared/filesystem-directory-listing-limit.ts). Real servers with huge single directories (e.g. browsing /nix/store or a giant maildir) would see an error toast instead of a listing; ordinary directories are orders of magnitude below. _(only when: SSH browser opening a directory with >100k entries or >12MB of names.)_

### Build / CI
- ⚠️ **Merge-time reconciliation** with recent `main`:
  - `src/main/ipc/pty.ts`: TRUE CONFLICT with perf(pty) #10113 (constant-time delivery-pressure): combine the `pendingDataTotalChars`/`deletePendingPtyData`/`setPendingPtyData` constant-time accounting with the OOM `PendingPtyDataMap` retention bound — keep both.

### ⚠️ Inherited edge-case defect(s) — pre-existing in #10179
- ~~[high] `src/main/ipc/pty-renderer-delivery-credit.ts`~~ — **refuted** (not a real bug): [REFUTED false-positive] Claimed evicted-PTY ACK misapplication is unreachable: recordSent is gated behind the accounting map admit() which rejects the 4097th PTY before the ledger eviction path can fire (ledger keys ⊆ accounting keys). Latent-fragility only if the two 4096 caps ever diverge.
- ⚠️ [medium] `src/main/ipc/pty.ts` — adversarial-flagged (overload-path only, unverified): After pending PTY output becomes a dropped-output sentinel, a newly salvaged query string is appended without truncating to the 4,096-char remaining allowance; one provider event retained ~4,000,000 chars in a repro.

### Adversarial review
- Skeptic pass on PR #23 (ipc-pty-runtime-remote). Verified every numeric ceiling against source; all sit far above ordinary use: SSH dir listing 100k entries/12MB (src/shared/filesystem-directory-listing-limit.ts); MAX_REGISTERED_SSH_PTY_PROVIDERS=256; MAX_TRACKED_SSH_CONNECTION_GENERATIONS=4096 / 8192 leases; SSH_MAX_PENDING_CREDENTIAL_REQUESTS=64, credential value 64KB, detail 16KB; REMOTE_WORKSPACE_PATCH_CONCURRENCY=4 (mapWithConcurrency, all targets still patched — correctness intact); TERMINAL_PREVIEW 64/window & 256 total; MAX_PENDING_PTY_DATA_CHARS=32MB, 16384 credit spans, 4096 states, 8MB id bytes; repo icon 256KB.

Correction to prior reviewer's MEDIUM claim: it states 'ssh.ts now calls routeExternalPtyData/Exit/Replay'. Inaccurate — git grep at TARGET shows the actual caller of routeExternalPty* is src/main/ssh/ssh-relay-session.ts, which is NOT in this slice's filelist. Within THIS slice, pty.ts only adds the receiving side (routeProviderData, PtyRendererDeliveryCreditLedger, PendingPtyDataMap retention bounds) and installExternalPtyRendererDeliveryRouter; it does not itself reroute remote output. The local PTY batched-delivery pipeline (pendingData Map + flush + sendPtyDataToRenderer flow control) already exists in BASE, so local terminal timing is unchanged below the new 32MB cap. Net: the remote-delivery timing/batching change flagged medium is not triggerable from this slice alone, so I do not treat it as a confirmed normal-path regression here.

Cleanup/lifecycle: ssh.ts adds symmetric releaseActiveSessionGeneration() at every teardown/connect-error/disconnect/reset path plus a per-connect finally lease release; the credit ledger writeOff on exit/drop/reload is exercised by pty-external-renderer-delivery/credit/retention tests. No missing-release path found. ssh-passphrase requestCredential returns null up-front on missing window / bad targetId / over-cap but proceeds normally when a window is present — behaviorally equivalent to BASE on the normal path (isSshRetainedIdentifier only rejects empty or >1024-byte ids; ordinary UUID targetIds pass).

Two low, user-visible-but-clear-error edges (repo icon 256KB cap; SSH huge-directory listing) are the only ordinary-adjacent changes; both surface an explicit message rather than corrupting or silently dropping. No off-by-one, FIFO break, or missing-cleanup defect confirmed. OOM evidence (caps + regression tests) is solid. Recommend opening as a draft PR for human sign-off on the two low edges and on the cross-slice remote-delivery reroute (whose caller lands in the ssh-relay-session slice).
- Correctness notes: `/Users/nwparker/orca/workspaces/orca/oom-followup/src/main/ipc/pty-renderer-delivery-credit.ts:132`: Confirmed cap/ordering bug: when the 4,097th in-flight PTY causes an old credit-ledger state to be collapsed and deleted, the corresponding cumulative-delivery accounting row remains. If that evicted PTY ID sends again before the renderer's ACK for its older byte arrives, recordSent creates a fresh ledger state; applyCumulativeAck then applies the old byte's ACK to the fresh span. A standalone reproduction against TARGET printed {"oldAck":1,"newAck":1} after acknowledging only one processed byte; newAck must still be 0. This prematurely returns producer credit for data the renderer has not parsed and defeats flow control under the state-cap overload path. The unique-ID test only checks one-shot eviction and never reuses the evicted ID.; `/Users/nwparker/orca/workspaces/orca/oom-followup/src/main/ipc/pty.ts:2422`: Confirmed ceiling bypass: once pending PTY output is a dropped-output sentinel, the code checks whether existing.data is already at 4,096 chars but appends the entire newly salvaged query string without truncating it to the remaining allowance. A standalone reproduction using TARGET's query extractor retained 4,000,000 chars from one repeated DSR-query chunk despite the declared 4,096-char cap. Subsequent calls stop growth, but one provider event can retain far above the intended O(1) sentinel bound. The regression test covers only two tiny queries and misses a large post-sentinel append.

### Files
34 files changed (+3161 / −654).

---

### DIFF
--- a/src/main/ipc/pty-external-renderer-delivery.test.ts
+++ b/src/main/ipc/pty-external-renderer-delivery.test.ts
@@ -1,0 +1,244 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const { handleMock, onMock, removeHandlerMock, removeAllListenersMock } = vi.hoisted(() => ({
+  handleMock: vi.fn(),
+  onMock: vi.fn(),
+  removeHandlerMock: vi.fn(),
+  removeAllListenersMock: vi.fn()
+}))
+
+vi.mock('electron', () => ({
+  app: {
+    isPackaged: true,
+    getPath: () => '/tmp/orca-pty-external-delivery-test',
+    getVersion: () => '0.0.0-test'
+  },
+  powerMonitor: { on: vi.fn() },
+  nativeTheme: { shouldUseDarkColors: true },
+  ipcMain: {
+    handle: handleMock,
+    on: onMock,
+    removeHandler: removeHandlerMock,
+    removeAllListeners: removeAllListenersMock
+  }
+}))
+
+vi.mock('node-pty', () => ({ spawn: vi.fn() }))
+vi.mock('../pwsh', () => ({ isPwshAvailable: vi.fn(() => false) }))
+
+import { LocalPtyProvider } from '../providers/local-pty-provider'
+import { registerPtyHandlers, setLocalPtyProvider } from './pty'
+import {
+  routeExternalPtyData,
+  routeExternalPtyExit,
+  routeExternalPtyReplay
+} from './pty-renderer-delivery-router'
+
+describe('external PTY renderer delivery', () => {
+  const mainWindow = {
+    isDestroyed: () => false,
+    isFocused: () => true,
+    isVisible: () => true,
+    isMinimized: () => false,
+    webContents: {
+      on: vi.fn(),
+      send: vi.fn(),
+      removeListener: vi.fn(),
+      isLoadingMainFrame: vi.fn(() => true)
+    }
+  }
+  const mainWindowEvent = { sender: mainWindow.webContents }
+
+  beforeEach(() => {
+    vi.useFakeTimers()
+    vi.clearAllMocks()
+    onMock.mockImplementation((channel: string, listener: (...args: unknown[]) => void) => {
+      if (channel === 'pty:rendererDispatcherReady') {
+        listener(mainWindowEvent)
+        vi.advanceTimersByTime(0)
+      }
+    })
+  })
+
+  afterEach(() => {
+    setLocalPtyProvider(new LocalPtyProvider())
+    vi.useRealTimers()
+  })
+
+  function findIpcListener(channel: string): (...args: never[]) => void {
+    const call = onMock.mock.calls.findLast(([registered]) => registered === channel)
+    if (!call) {
+      throw new Error(`Missing ${channel} listener`)
+    }
+    return call[1] as (...args: never[]) => void
+  }
+
+  function makeRuntime(sequence = 17) {
+    return {
+      setPtyController: vi.fn(),
+      onPtyData: vi.fn(() => sequence),
+      getPtyOutputSequence: vi.fn(() => sequence),
+      hasRawTerminalViewSubscriber: vi.fn(() => false)
+    }
+  }
+
+  it('returns captured SSH credit only after the renderer acknowledges parsing', () => {
+    const runtime = makeRuntime()
+    const upstreamCredit = { charCount: 9, acknowledge: vi.fn() }
+    registerPtyHandlers(mainWindow as never, runtime as never)
+    mainWindow.webContents.send.mockClear()
+
+    routeExternalPtyData({
+      id: 'ssh:target@@pty-1',
+      data: '',
+      sequenceChars: 9,
+      transformed: true,
+      upstreamCredit
+    })
+    vi.advanceTimersByTime(2)
+
+    expect(runtime.onPtyData).toHaveBeenCalledWith(
+      'ssh:target@@pty-1',
+      '',
+      expect.any(Number),
+      9,
+      true
+    )
+    expect(mainWindow.webContents.send).toHaveBeenCalledWith('pty:data', {
+      id: 'ssh:target@@pty-1',
+      data: '',
+      seq: 17,
+      rawLength: 9,
+      transformed: true
+    })
+    expect(upstreamCredit.acknowledge).not.toHaveBeenCalled()
+
+    findIpcListener('pty:ackData')(
+      mainWindowEvent as never,
+      {
+        id: 'ssh:target@@pty-1',
+        processedChars: 9
+      } as never
+    )
+
+    expect(upstreamCredit.acknowledge).toHaveBeenCalledOnce()
+    expect(upstreamCredit.acknowledge).toHaveBeenCalledWith(9)
+  })
+
+  it('returns credit immediately when hidden output is intentionally dropped', () => {
+    const runtime = makeRuntime(42)
+    const upstreamCredit = { charCount: 13, acknowledge: vi.fn() }
+    registerPtyHandlers(mainWindow as never, runtime as never)
+    findIpcListener('pty:setHiddenRendererPty')(
+      null as never,
+      {
+        id: 'ssh:target@@pty-hidden',
+        hidden: true
+      } as never
+    )
+    mainWindow.webContents.send.mockClear()
+
+    routeExternalPtyData({
+      id: 'ssh:target@@pty-hidden',
+      data: 'hidden output',
+      upstreamCredit
+    })
+
+    expect(runtime.onPtyData).toHaveBeenCalledOnce()
+    expect(upstreamCredit.acknowledge).toHaveBeenCalledWith(13)
+    expect(mainWindow.webContents.send).toHaveBeenCalledOnce()
+    expect(mainWindow.webContents.send).toHaveBeenCalledWith('pty:modelRestoreNeeded', {
+      id: 'ssh:target@@pty-hidden',
+      reason: 'hidden-drop',
+      markerSeq: 42
+    })
+  })
+
+  it('settles queued live credit when a reconnect replay supersedes it', () => {
+    const runtime = makeRuntime()
+    const upstreamCredit = { charCount: 11, acknowledge: vi.fn() }
+    registerPtyHandlers(mainWindow as never, runtime as never)
+    mainWindow.webContents.send.mockClear()
+
+    routeExternalPtyData({
+      id: 'ssh:target@@pty-replay',
+      data: 'queued live',
+      upstreamCredit
+    })
+    routeExternalPtyReplay({ id: 'ssh:target@@pty-replay', data: 'full snapshot' })
+    vi.advanceTimersByTime(10)
+
+    expect(upstreamCredit.acknowledge).toHaveBeenCalledWith(11)
+    expect(mainWindow.webContents.send.mock.calls).toEqual([
+      ['pty:replay', { id: 'ssh:target@@pty-replay', data: 'full snapshot' }]
+    ])
+  })
+
+  it('writes off captured in-flight credit when the renderer lifecycle ends', () => {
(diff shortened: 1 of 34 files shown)


nwparker:

PR #10317, opened 2026-07-24
oom(24/29): bound remaining IPC channels (notifications, speech, notebook)

nwparker:
## OOM hardening slice 24/29 — bound remaining IPC channels (notifications, speech, notebook)

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **24/29** (base: `oom/23-b-ipc-pty-runtime-remote`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
This slice re-introduces OOM ceilings across ~20 main-process IPC handlers (AI-vault host scans, emulator MJPEG/H.264 streaming, custom-pet file reads, local-log-tail watches, native-chat subscriptions, native notifications, notebook/keyboard subprocess capture, speech dictation, markdown listing) plus a preload ack path. Caps are concurrency/count/byte/depth limits with per-sender and process-wide scopes; the one genuinely normal-path change is that emulator frame delivery is now renderer-ack-gated (one frame in flight).

### ELI5
Imagine a bunch of pipes in the app that fed data in as fast as it arrived, with no valve. If something misbehaved (a huge file, a spammy stream, thousands of requests), a pipe could flood memory and crash the app. This change adds valves everywhere: hard limits on how many streams/watchers/notifications can exist, how big a file or message can be, and — for the phone-emulator video preview — it now waits for the screen to confirm it received each frame before sending the next, so frames pile up in a small buffer instead of an unbounded one.

### Proof of OOM fix
- AI_VAULT_ALL_HOST_SCAN_CONCURRENCY=4 batches all-host scans (ai-vault-host-scan-batches.ts); test: ai-vault.test.ts 'bounds concurrent scans across a pathological runtime-host roster' (peak<=4 with 20 hosts) + oversized session-id/cache-key bypass tests
- KEYBOARD_INPUT_SOURCE_MAX_OUTPUT_BYTES=1MiB kills the macOS input-source probe and falls back (app.ts); test: app.test.ts 'kills an oversized selected-input probe before using the layout fallback'
- Emulator MJPEG: MAX_SESSIONS_TOTAL=8 / PER_RENDERER=2, single-in-flight + latest-frame-only delivery (emulator-frame-stream.ts); tests: emulator-frame-stream-admission.test.ts, emulator-frame-stream-delivery.test.ts
- Emulator H.264: MAX_SUBSCRIPTIONS_TOTAL=8 / PER_RENDERER=2, SCRCPY_VIDEO_MAX_GOP_FRAMES=120, SCRCPY_VIDEO_MAX_REPLAY_BYTES_PER_DEVICE=32MiB, MAX_PENDING_FRAMES=121 (emulator-video-frame-delivery.ts); tests: emulator-video-stream-admission.test.ts, emulator-video-frame-delivery.test.ts
- Custom pet: MAX_CUSTOM_PET_FILE_BYTES=64MiB, MAX_CUSTOM_PET_READ_CALLS=1024, sheet dims<=8192 & <=4M px (custom-pet-file-reader.ts, pet.ts); tests: custom-pet-file-reader.test.ts, pet.test.ts (dimension-bomb rejections)
- Local log tail: 32/128 watches (per-sender/process), 16/64 concurrent reads, id/path/identity byte caps (local-log-tail-operation-admission.ts, local-log-tail.ts); test: local-log-tail.test.ts (cap + race + oversized-id cases)
- Native chat: read window clamped to MAX_NATIVE_CHAT_DESKTOP_READ_LIMIT=2000, 64/256 subscriptions, 8/16 reads, 64/256 setup attempts, id byte caps (native-chat-ipc-admission.ts, native-chat-subscription-registry.ts); test: native-chat-subscribe-lifecycle.test.ts
- Native notifications: MAX_ACTIVE_NATIVE_NOTIFICATIONS=128 with oldest-eviction; dispatch input 256KiB, 256 dismiss ids (native-notification-retention.ts, notification-ipc-admission.ts); tests: notifications.test.ts, notification-ipc-admission.test.ts
- Notebook capture bounded at MAX_CAPTURE_BYTES=2MiB via GrowingByteBuffer (notebook.ts); test: notebook.test.ts '100,000 one-byte subprocess events'
- Speech: audio chunk<=1MiB, 256 hotwords/4KiB each/256KiB total, session-id 1KiB, 16 pending starts, 16 active listeners (speech-ipc-admission.ts, speech.ts); test: speech.test.ts
- github.ts PR_REFRESH_VISIBLE_CANDIDATE_LIMIT=1024; markdown listing MAX_DOCUMENTS=20000 / METADATA 8MiB / PATH 64KiB / VISITED 100k / DEPTH 256 (markdown-documents.ts); test: markdown-documents.test.ts

### Regressions
- **medium**: Emulator MJPEG and H.264 frame delivery is now renderer-ack-gated: only one frame is in flight at a time and the next is sent only after the renderer's preload auto-acks. Throughput is bounded by IPC round-trip latency; if the renderer decodes/paints slower than the device produces frames, MJPEG shows only the latest frame and H.264 drops to the next keyframe. Preview may look choppier under load than the old fire-and-forget path, though it no longer grows memory unboundedly. _(trigger: Ordinary emulator/scrcpy preview when the renderer is slow (large frames, busy UI, background tab); reachable in normal use, not only abuse.)_
- **medium**: Emulator MJPEG + H.264 preview delivery is now renderer-ack-gated (one frame in flight; preload auto-acks in a finally after the renderer callback). Under normal decode speed this is invisible (main<->renderer round-trip is sub-ms vs a 16-33ms frame budget), but when the renderer paints slower than the device produces frames, MJPEG collapses to only-latest-frame and H.264 drops the buffered GOP and waits for the next keyframe (momentary freeze). Reachable in ordinary emulator/scrcpy preview under a busy/backgrounded UI, but only degrades under backpressure — it does not degrade the steady state. _(trigger: Emulator/scrcpy preview while the renderer decode/paint is slower than device frame production (large frames, busy UI, background tab).)_
- **medium**: All-host AI Vault scans no longer start every host concurrently. They run in batches of four and await each batch before starting the next, so five scan legs take at least two waves and can roughly double refresh latency; later hosts can wait behind earlier 3-second runtime timeouts. This is below all count/byte ceilings and is a normal-path behavior change for users with four or more remote execution hosts plus local. _(trigger: Open or refresh Agent Session History with executionHostScope='all' when at least five local/SSH/runtime scan legs are active.)_
- **medium**: MJPEG and H.264 emulator delivery is renderer-ack-gated. MJPEG retains only the newest pending frame; H.264 retains a bounded GOP and, after overflow, drops deltas until a keyframe. A busy renderer therefore receives fewer frames than the previous fire-and-forget path, making preview visibly choppier under ordinary UI load. _(trigger: Use an emulator preview while the renderer event loop is busy or frame production outpaces IPC acknowledgements.)_
- **low**: Native-chat desktop reads now clamp every requested history window to 2,000 turns. The renderer still requests 2,100 on the next pagination step and then sees only 2,000 returned, so it concludes there is no more history; older turns become inaccessible without an explicit truncation indication. _(trigger: Scroll back through a native-chat transcript containing more than 2,000 turns.)_
- **low**: MJPEG stream errors are coalesced until a frame is successfully acknowledged. A stream that repeatedly reconnects but never yields a frame reports only its first error, so a later, more informative failure message is not surfaced. _(trigger: An MJPEG emulator stream fails repeatedly before delivering any successful frame, with the failure reason changing between attempts.)_

- Emulator frame/video streams: starting a 3rd stream per renderer or 9th overall is rejected with 'A renderer can have at most 2' / 'Orca can have at most 8' (verified far above ordinary use — one emulator preview per pane)
- Under sustained renderer backpressure MJPEG drops all but the newest frame; H.264 drops the buffered GOP and waits for the next keyframe (momentary video freeze) once >120 frames or >32MiB queue behind an unacked frame
- Native notifications past 128 active evict the oldest (closing it); dispatch/dismiss payloads past 256KiB / 256 ids are rejected/truncated
- Local-log-tail, native-chat reads/subscriptions, and speech dictation starts past their per-sender/process caps reject with a 'Too many ...' error
- Keyboard input-source probe producing >1MiB stdout is SIGKILLed and the layout fallback is used
- Markdown listing throws MarkdownDocumentListingCapacityError only past 20k docs / 8MiB metadata / a single >64KiB path
- Custom-pet import and pet:read now validate raster dimensions by parsing image headers (readRasterImageDimensions, WebP/PNG/GIF/JPEG only) and enforce <=8192px & <=4M px, replacing Electron nativeImage decode. Images whose headers cannot be parsed throw 'Could not decode', and previously-imported stored pets that exceed the dimension bound now make pet:read return null (the pet silently fails to load). _(only when: Importing/loading a pet image in a format not covered by the header parser, or an existing stored pet larger than 8192px/4M px.)_
- notifications:dispatch now normalizes/whitelists the payload; any non-string string-field or non-boolean control field makes the whole request return {delivered:false, reason:'invalid-request'} and unknown fields are dropped. A single agent message field over the 256KiB aggregate budget rejects the notification. _(only when: Malformed dispatch payloads (renderer-controlled, so unlikely in practice) or an extremely large agent last-message/prompt.)_
- pet:import and pet:read now enforce raster dimensions via header parsing (readRasterImageDimensions; WebP/PNG/GIF/JPEG) with each side <=8192px AND width*height<=4,194,304px (~2048x2048). The legacy pet:import path and pet:read previously did NO dimension check. A user who previously imported a large custom sprite sheet (>4M px, e.g. 2500x2500 or 4000x4000) will now have pet:read return null (pet silently fails to load), and a new import of such a sheet throws 'exceed the safe limit'. The shipped default codex pet (1536x1872 = 2.87M px) and typical sheets stay under the ceiling, so ordinary users are unaffected; only large custom sheets regress. _(only when: Importing or loading an existing custom pet sprite sheet larger than 4M pixels or 8192px per side; or an image in a format the header parser cannot read ('Could not decode').)_
- pet:read now runs every stored raster pet through readRasterImageDimensions + isCustomPetSheetSizeSafe (<=8192px per side AND <=4,194,304 total px). Any image that trips the gate throws, is swallowed by the catch, and pet:read returns null with only a console.warn, so the pet silently vanishes. This gate did not exist in BASE (BASE pet:read just returned the raw buffer, and BASE import only enforced the 64MB byte cap, never dimensions). Pets imported before this change whose sheet exceeds 4M px (e.g. a 2048x2048=4.2M px or any 4096-wide spritesheet) will stop loading after update with no user-facing error. _(only when: App start / pet render for an existing custom sidekick whose stored image exceeds 8192px on a side or 4,194,304 total pixels. 4M px is easily reached by high-DPI or multi-frame spritesheets, so this is reachable in ordinary use for the subset of users with large legacy pets.)_
- Custom-pet raster imports and reads now require parseable bounded dimensions (at most 8192 on either axis and 4,194,304 pixels) before renderer delivery. Previously stored oversized pets now return null, and valid but unusual allowed-format files whose dimension header is not found by the bounded parser (for example a JPEG whose SOF occurs after the 1 MiB marker scan) no longer load. _(only when: Import or load a custom pet exceeding the pixel/dimension bounds or using an allowed raster encoding the bounded header parser cannot identify.)_

### Build / CI

### Adversarial review
- Verified numeric ceilings against ordinary use; all are far above normal workloads: emulator streams 2/renderer, 8 total (one preview per pane); H.264 replay bound 120 GOP frames / 32MiB per device; MAX_CUSTOM_PET_FILE_BYTES=64MiB; HTML_TO_PDF_MAX_INPUT_BYTES=32MiB; PR_REFRESH_VISIBLE_CANDIDATE_LIMIT=1024; markdown 20k docs / 8MiB metadata / 64KiB path; notifications 128 active / 256KiB dispatch / 256 dismiss ids; speech 1MiB audio chunk / 256 hotwords / 16 pending+active dictation; local-log-tail 32 watches & 16 reads per sender (128/64 process-wide); native-chat 64 subs & 8 reads per sender (256/16 process-wide); keyboard input-source probe 1MiB stdout. None are reachable by ordinary use. The two user-facing changes worth human sign-off are: (1) the emulator ack-gated delivery, which is a steady-state-invisible backpressure mechanism that only degrades a slow renderer (arguably an improvement over the old unbounded fire-and-forget path), and (2) the new pet dimension gate, which is a genuine new restriction on the pet import/read path but with a ceiling above the shipped default pet and typical sheets. Neither is a broad normal-path regression: below the limits, ordinary users see no visible change. export.ts refactor preserves the empty-HTML 'No content' result and only adds a 32MiB reject. local-log-tail / native-chat / notification / speech refactors add per-sender+process-wide admission with matching release on stop/abort/destroy and are covered by regression tests (cap-exact, recovery-after-release, destroyed-race, out-of-order authorization). OOM evidence (caps + tests) is concrete and present. This is the exact production commit that passed full CI; slicing is whole-file/lossless. Safe to open as a draft PR for human review; the two flagged behavior changes should be called out in the PR description.

### Files
47 files changed (+3883 / −531).

---


nwparker:

PR #10318, opened 2026-07-24
oom(25/29): bound CLI runtime output & transport

nwparker:
## OOM hardening slice 25/29 — bound CLI runtime output & transport

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **25/29** (base: `oom/24-b-ipc-rest`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
This slice re-introduces byte/depth/line/entry ceilings across Orca's CLI surface (runtime transport frames, recipe-JSON streaming, control/config file reads, computer-use stdin, Linear body input, WSL/Linux launcher inspection, and a bounded LRU for the terminal shim cache), plus a new `verify:cli-runtime` build/CI gate. All caps sit far above ordinary CLI use.

### ELI5
Orca's command-line tools used to slurp whole files and network replies into memory with no size limit, so a giant or malformed input could blow up memory. This change adds sane upper limits (like "reply can be at most 64MB" or "config file at most 256KB") and cleanly rejects anything past them, while leaving normal-sized inputs untouched. It also adds a build step that actually runs the compiled CLI's --help so a broken binary is caught before release.

### Proof of OOM fix
- transport.ts: MAX_RUNTIME_RESPONSE_LINE_BYTES=64MB per-frame cap + CLI_RUNTIME_JSON_STRUCTURE_LIMITS {nestingDepth:128, structuralTokens:1_000_000} checked BEFORE JSON.parse; buffer switched from string concat to GrowingByteBuffer. Test: transport.test.ts 'rejects excessive response nesting before JSON.parse' asserts JSON.parse never called.
- recipe-json-output.ts: MAX_RECIPE_OUTPUT_LINE_BYTES=4MB; oversized line without newline kills child SIGTERM and detaches data listener. Tests: launch.test.ts 'kills a server that emits an oversized recipe line' and 'preserves recipe JSON after 100,000 one-byte fragments'.
- metadata.ts / runtime-bootstrap: MAX_RUNTIME_METADATA_FILE_BYTES=64KB; serve-update-supervisor.ts: MAX_SERVE_UPDATE_HANDOFF_FILE_BYTES=64KB bounded read + bounded transactional write (temp file unlinked on over-limit serialize). Tests: control-file-bounds.test.ts (oversized→null/throw; failed serialize preserves prior handoff, removes .tmp).
- agent-hooks.ts: ORCA_PERSISTED_STATE_MAX_BYTES=64MB bounded read + write serialized within cap-1. Test: agent-hooks.test.ts 'rejects an oversized sparse offline state file without replacing it' (exit 1, file untouched).
- vm.ts: orca.yaml read bounded to MAX_ORCA_YAML_BYTES=256KB. Test: index.test.ts 'rejects an oversized sparse orca.yaml during vm recipe doctor'.
- computer-action-flags.ts: stdin bounded to CLIPBOARD_TEXT_WRITE_MAX_BYTES=16MB; linear-request-builders.ts: body file/stdin bounded to LINEAR_WRITE_BODY_CAP(65000)*4 bytes.
- mac-app-update-bundle.ts: MAC_BUNDLE_INFO_PLIST_MAX_BYTES=1MB bounded plist read. Test: mac-app-update-bundle.test.ts sparse-oversized plist → resolves false.
- cli-installer.ts CLI_LAUNCHER_INSPECTION_MAX_BYTES=64KB; linux-bare-orca-dispatcher.ts 64KB; wsl-cli-installer.ts WSL_CLI_INSPECTION_MAX_BYTES=64KB (head -c limit+1 | base64, over-limit→'not_file'); wsl-cli-registration-registry.ts WSL_CLI_REGISTRY_MAX_BYTES=1MB (over-limit→emptyState). Tests: wsl-cli-installer-bounds.test.ts, wsl-cli-registration-registry.test.ts.
- linux-terminal-orca-cli-shim.ts: ensuredShimDirs turned into LRU capped at LINUX_TERMINAL_SHIM_CACHE_MAX_ENTRIES=64 with LINUX_TERMINAL_SHIM_CACHE_KEY_MAX_BYTES=64KB key-size guard (keys over limit simply not cached). Test: (bounds test in slice).

### Regressions
- **low**: orchestration.ts message/inbox/task/gate list output now appends a 'Truncated: N additional <noun> omitted.' line when the server marks the page truncated. Additive and informative, but it is a visible normal-path text change to CLI output that consumers/scripts may not expect. _(trigger: Server returns truncated=true on orchestration.check/inbox/taskList/gateList (i.e. server-side retention cap hit, which is enforced in a different slice).)_

- Runtime RPC response > 64MB in a single frame, or JSON nesting > 128 / > 1M structural tokens: request fails with invalid_runtime_response instead of parsing.
- Recipe JSON line > 4MB (or no newline before 4MB): serve is killed with runtime_serve_failed. Real recipe JSON is a few KB.
- orca.yaml > 256KB, Info.plist > 1MB, runtime metadata/handoff > 64KB, managed launcher script > 64KB, WSL registry > 1MB: treated as unavailable/not_file/empty. All are tiny in practice.
- computer-use stdin payload > 16MB or Linear body > ~260KB (65k chars): rejected with invalid_argument. Far above normal typed/pasted input.
- agent-hooks (orca-data.json) read or write when state > 64MB: CLI prints 'File too large' and exits 1 rather than parsing/overwriting.
- Terminal shim cache: with >64 distinct userDataPaths, oldest entries evicted (just re-writes the shim on next spawn); userDataPath keys >64KB never cached.
- WSL managed-launcher inspection now decodes base64 and requires set -o pipefail; a launcher file larger than 64KB is reported as 'not_file' and may be treated as unmanaged/reinstallable. _(only when: A user's WSL orca-ide launcher file exceeds 64KB (not the case for Orca-generated wrappers).)_

### Build / CI

### Adversarial review
- Verified every ceiling in this CLI slice sits far above ordinary use: runtime transport frame 64MB / depth 128 / 1M tokens (transport.ts); recipe JSON line 4MB (recipe-json-output.ts); Linear body 65,000 chars = 260KB byte cap (LINEAR_WRITE_BODY_CAP); computer-use stdin 16MB (CLIPBOARD_TEXT_WRITE_MAX_BYTES); control/config reads (orca.yaml 256KB, Info.plist 1MB, metadata/handoff 64KB, WSL registry 1MB, launcher 64KB); terminal shim LRU 64 distinct userDataPaths (normally 1 per host) with 64KB key cap. None reachable by large-but-normal repos, long sessions, big diffs, many worktrees, or normal images/PDFs. Only normal-path-visible change is orchestration.ts appending 'Truncated: N additional <noun> omitted.' but it is gated entirely on server returning truncated=true; the server-side retention cap that sets that flag lives in a DIFFERENT slice, so within this slice CLI output is byte-identical to base unless the server already truncates. Additive informative text, not a regression. transport.ts success teardown still uses socket.end() (unchanged); only failure path switched to socket.destroy() to abort in-flight socket, with buffer.clear() on settle — clean improvement, no leak. Confirmed cap constants in source. No normal-path regression or correctness bug confirmed; safe to open as draft PR.

### Files
36 files changed (+1103 / −312).

---


nwparker:

PR #10319, opened 2026-07-24
oom(26/29): bound renderer editor & terminal-pane buffers

nwparker:
## OOM hardening slice 26/29 — bound renderer editor & terminal-pane buffers

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **26/29** (base: `oom/25-b-cli`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Bounds renderer editor + terminal-pane accumulators against OOM with concrete numeric caps and extensive tests. Combined-diff text retention (24MB/12M-char budget, evict+refetch offscreen bodies, view-state LRU=20), CSV parse limits (50MB/250k rows/1024 cols/1M cells) with graceful fallback, notebook admission limits, local-image blob LRU (128MB/100) + load admission (2 active/100 admitted) + raster dimension validation, markdown completion/worktree/scan caps, and terminal input/shutdown/validation queue caps. Replaces unbounded Promise.all(map) with order-preserving mapWithConcurrency (child-checks 8, bulk-save 4, link-probes 8). Bundles two correctness fixes: load-scheduler reset keeps `active` so stale loads still count against concurrency, and pty-input-queue clearVersion prevents post-clear writes. SVG and valid rasters pass through unchanged; only decode bombs rejected. Slice compiles standalone; tab-retirement cleanup wiring lives in store slices (another bucket).

### ELI5
When files, diffs, images, notebooks, or terminal streams get huge, Orca used to keep everything in memory at once and could crash. This adds size limits everywhere: oversized CSVs/notebooks show a friendly 'too large' message, giant diffs quietly drop offscreen text and refetch it when you scroll back, image bombs are refused before they can blow up memory, and typing/paste backlogs and background checks are capped and run a few at a time. Normal files behave exactly as before.

### Proof of OOM fix
- combined-diff-text-retention.ts: MAX_RETAINED_COMBINED_DIFF_TEXT_BYTES=24MB (12M chars); offscreen non-active/non-dirty/non-loading sections evicted+reloadable; view-state cache COMBINED_DIFF_VIEW_STATE_CACHE_MAX_ENTRIES=20 + aggregate budget
- csv-parse.ts CSV_PARSE_LIMITS: 50MB source / 250k rows / 1024 cols / 1M cells / 50MB retained cell text -> CsvLimitFallback
- ipynb-json-admission.ts IPYNB_MEMORY_LIMITS: 50MB source, 500k structural tokens, depth 256, 1000 cells, 10k outputs, 20k display items, 100k multiline parts
- useLocalImageSrc.ts: LocalImageBlobRetention 128MB/100 entries, LocalImageLoadAdmission 2 active/100 admitted, chunked base64 decode, assertRasterImagePreviewWithinLimits (32768px/side, 32M px)
- monaco-markdown-doc-completions.ts: 256 models / 32 scopes / 64MB; markdown-document-worktree-retention.ts: 8 snapshots / 32MB; markdown-document-list-request.ts: global in-flight cap 16
- markdown-export-extract.ts: HTML_TO_PDF_MAX_INPUT_BYTES fragment cap with sequential blob-image inlining and streamed-body limit
- pty-input-write-queue.ts (4096 items + byte budget), pty-shutdown-data-suspension.ts (PTY_SHUTDOWN_OUTPUT_MAX_EVENTS=4096), remote-runtime-pty-batching.ts (validation-queue 4096), remote-runtime-viewport-claim-input.ts (<=TERMINAL_INPUT_MAX_BYTES)
- concurrency caps: CHILD_PROCESS_CHECK_CONCURRENCY=8, EDITOR_BULK_SAVE_CONCURRENCY=4, TERMINAL_FILE_LINK_PROBE_CONCURRENCY=8 via order-preserving mapWithConcurrency

### Regressions

- CSV beyond 50MB/250k rows/1024 cols/1M cells shows 'too large' fallback with source-mode escape
- Combined diff over 24MB retained text evicts offscreen section bodies and refetches on revisit (transparent)
- Notebook exceeding shape/source limits throws a safe-limit error instead of projecting
- Raster image exceeding 32768px/side or 32M px, or a decode bomb, shows failure instead of loading (ImageViewer/IpynbViewer/local preview)
- Markdown completions unavailable for a listing that exceeds the per-listing byte limit; 17th+ concurrent distinct worktree scan rejected with a toast
- PDF export of oversized markdown/images rejected with 'HTML export exceeds the PDF memory limit'
- Terminal input/paste, shutdown events, and remote validation queues reject producer backlog beyond their item/byte budgets

### Build / CI
- Part of the **Renderer** group; this group's cumulative tree is interlinked, so it becomes typecheck-green at its checkpoint **PR #29** (whole-repo interconnection — see stack README). Content is exact production code.
- Touches 1 file(s) also changed on `main` since the revert; git auto-merges them (no conflict): `remote-runtime-pty-transport.ts`.

### Adversarial review
- Verified no infinite render loop in CombinedDiffViewer: retainCombinedDiffSectionText returns the same array reference when under budget, so setSections bails via Object.is; eviction stabilizes in one pass. load-scheduler no longer zeroing `active` is an intentional bound fix (always decrements in finally; loads always settle, no hang). useLocalImageSrc not-admitted path returns before inFlightBlobUrlLoads.set, so no stuck in-flight key. SVG/valid raster data URIs preserved (validateRasterImageDataUri returns undefined for non-raster). Borderline normal-path edges (CSV >1024 columns; >100 concurrent local images; valid-but-unparseable raster showing an error) are rare and non-fatal with graceful fallbacks. Cleanup exports (forgetTerminalPaneRecovery bridge, collectTerminalLayoutLeafIds, releaseTerminalScrollIntentKeys) are additive; retirement wiring lives in store slices in another bucket, so full map-cleanup activates only once that bucket lands, but this slice does not break without it.

### Files
60 files changed (+2459 / −308).

---

### DIFF
--- a/src/renderer/src/components/Terminal.tsx
+++ b/src/renderer/src/components/Terminal.tsx
@@ -12,6 +12,7 @@ import {
 } from '@/constants/terminal'
 import { useAppStore } from '../store'
 import { folderWorkspaceKey } from '../../../shared/workspace-scope'
+import { mapWithConcurrency } from '../../../shared/map-with-concurrency'
 import { useAllWorktrees } from '../store/selectors'
 import { getConnectionId } from '../lib/connection-context'
 import { basename } from '../lib/path'
@@ -141,6 +142,7 @@ const EditorPanel = lazy(() => import('./editor/EditorPanel'))
 
 // Why: gate handler runs after a dialog advances so a stray carry-over click can't act on the next dialog; ~200ms absorbs a physical double-click while staying responsive.
 const CLOSE_DIALOG_DEBOUNCE_MS = 200
+const CHILD_PROCESS_CHECK_CONCURRENCY = 8
 const EDITOR_TAB_CONTENT_TYPES = new Set<TabContentType>([
   'editor',
   'diff',
@@ -447,15 +449,15 @@ function Terminal(): React.JSX.Element | null {
           }
         )
         if (localPtyIds.length > 0) {
-          void Promise.all(localPtyIds.map((id) => window.api.pty.hasChildProcesses(id))).then(
-            (results) => {
-              if (results.some(Boolean)) {
-                setWindowCloseDialogOpen(true)
-              } else {
-                confirmNativeWindowClose()
-              }
+          void mapWithConcurrency(localPtyIds, CHILD_PROCESS_CHECK_CONCURRENCY, (id) =>
+            window.api.pty.hasChildProcesses(id)
+          ).then((results) => {
+            if (results.some(Boolean)) {
+              setWindowCloseDialogOpen(true)
+            } else {
+              confirmNativeWindowClose()
             }
-          )
+          })
           return
         }
       }
--- a/src/renderer/src/components/editor/CombinedDiffViewer.tsx
+++ b/src/renderer/src/components/editor/CombinedDiffViewer.tsx
@@ -76,6 +76,12 @@ import { removeDiffSectionMeasuredHeight } from './diff-section-height-cache'
 import { createCombinedDiffLoadScheduler } from './combined-diff-load-scheduler'
 import { combinedDiffSectionsMatchEntryMetadata } from './combined-diff-section-cache-match'
 import {
+  COMBINED_DIFF_VIEW_STATE_CACHE_MAX_ENTRIES,
+  getCombinedDiffViewedSectionKeys,
+  retainCombinedDiffSectionText,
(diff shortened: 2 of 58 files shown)


nwparker:

PR #10322, opened 2026-07-24
oom(29/29): bound web-renderer variants

nwparker:
## OOM hardening slice 29/29 — bound web-renderer variants

> **Part of a 29-PR stack re-landing #10179** ("bound OOM-prone accumulators across Orca"), which was reverted in #10255 for being too large (~1,600 files). This is slice **29/29** (base: `oom/28-b-renderer-state`). **Merge in numeric order.** The full stack's end-state is byte-for-byte the commit that passed #10179's complete CI matrix (36k+ tests, multi-OS).

### Summary
Re-introduces OOM ceilings on the renderer web/SSH remote-runtime path: pre-decode admission caps for E2EE keys/pairing input, bounded browser-local JSON persistence, capped RPC/subscription/waiter/child-client counts, and an outbound WebSocket backpressure budget that closes+reconnects overloaded sockets instead of growing memory.

### ELI5
When Orca runs in a browser or over SSH, it talks to your machine's runtime over an encrypted connection. This change adds seatbelts: it refuses absurdly large messages, keys, and saved-settings blobs before doing expensive work, limits how many requests/subscriptions can pile up, and if the outgoing pipe gets clogged it hangs up and reconnects rather than eating unlimited memory. All the limits are far above what normal use ever reaches.

### Proof of OOM fix
- web-e2ee.ts: public key <=44 base64 chars, encrypted text <= MAX_E2EE_ENCRYPTED_BASE64_CHARACTERS (~4MiB plaintext); test web-e2ee.test.ts asserts atob not called past the cap
- web-pairing.ts: PAIRING input 129KiB / code 128KiB / endpoint 16KiB / deviceToken 64KiB / publicKey 4KiB (shared/mobile-relay-pairing-offer); test web-pairing.test.ts boundary accept+reject
- web-local-storage-json.ts: WEB_LOCAL_STORAGE_JSON_LIMITS = 8MiB bytes / 1_000_000 structural tokens / depth 128; test web-local-storage-json.test.ts asserts JSON.parse not reached on amplified input
- web-preload-request-owners.ts: WEB_PRELOAD_MAX_ABORTABLE_REQUESTS=256, WEB_PRELOAD_MAX_REQUEST_TOKEN_BYTES=4KiB; test web-preload-request-owners.test.ts
- web-preload-api.ts: WEB_PRELOAD_MAX_KEYBINDING_LISTENERS=64, WEB_RUNTIME_REPO_DISCOVERY_CONCURRENCY=8; test web-preload-runtime-discovery-concurrency.test.ts resolves 300 repos with peak concurrency==8
- web-runtime-client.ts: 256 waiters/pending/subscriptions, 64 child clients, 64MiB inbound binary, 8MiB outbound binary, 4MiB outbound JSON, 1MiB subscription params, 8KiB RPC method; tests web-runtime-client-memory-bounds.test.ts + web-runtime-client.test.ts
- web-runtime-outbound-memory-budget.ts: 16MiB buffered / 32MiB queued / 4096 frames / 16MiB retained-subscription / 32MiB prepared-RPC / 65 socket sources (=64 children+1 parent), shared across parent+child clients; test web-runtime-outbound-memory-budget.test.ts

### Regressions
- **low**: Runtime worktree discovery fan-out changed from unbounded Promise.all to mapWithConcurrency(8), so detected-worktree discovery for users with many repos is now processed 8 repos at a time (slightly higher latency, no functional change). _(trigger: Web/SSH runtime user with a large repo catalog (test exercises 300 repos).)_

- Outbound WebSocket queue overflow (sustained backpressure beyond 32MiB / 4096 frames / 16MiB buffered) closes the socket and schedules a reconnect (failOutboundSocket) rather than buffering unbounded
- Inbound encrypted text frame over ~4MiB or JSON exceeding 1M tokens / depth 128 closes the remote-runtime socket (failInboundJsonCapacity)
- Exceeding 256 pending requests / 256 connection waiters / 256 subscriptions / 64 child clients rejects new calls with 'client is busy'
- Browser-local JSON over 8MiB or too-deeply-nested throws; readJson swallows it and returns defaults, writeJson throws
- Oversized outbound binary frame (>8MiB) closes the socket before encrypting
- keybindings.onChanged now throws 'listener capacity reached' after 64 concurrently registered listeners instead of growing an unbounded Set. _(only when: Only if >64 keybinding change listeners are registered without unsubscribing; far above ordinary use.)_

### Build / CI

### Adversarial review
- Skeptic pass on PR #29 (B-renderer-web) found NO confirmed normal-path regression. All new ceilings sit far above ordinary use and several interact coherently.\n\nKey verifications:\n- Inbound encrypted text-frame cap (~4MiB, WEB_RUNTIME_MAX_ENCRYPTED_TEXT_FRAME_BYTES = MAX_E2EE_ENCRYPTED_BASE64_CHARACTERS): the 4MiB MAX_E2EE_TEXT_PLAINTEXT_BYTES threshold ALREADY exists in BASE (src/shared/e2ee-crypto.ts at aab112933e), where oversized frames were silently dropped by decrypt() returning null. This slice only changes the reaction to close+reconnect (failInboundJsonCapacity/failOutboundSocket). No new normal-path frame is rejected; server is designed to stay under 4MiB (chunks large data via 64MiB binary frames). Overload-only.\n- mapWithConcurrency(8) repo discovery (web-preload-api.ts:3756): users with <=8 repos see no change; also necessary to stay under the new 256 pending-request cap that unbounded Promise.all would otherwise trip for 257+ repos. Negligible latency for many-repo users.\n- Keybinding listener cap 64 (web-preload-api.ts:404): effect-cleanup usage keeps live count tiny; the new Map also fixes a pre-existing duplicate storage-listener leak in the old Set path.\n- Pairing ceilings (128KiB/64KiB/16KiB/4KiB) dwarf real offers (tens of chars). Overload-only.\n- Socket-source cap 65 = parent(1)+64 children exactly; no off-by-one lockout. terminal.multiplex muxes terminals so heavy users do not fan out to 64 sockets.\n- Outbound backpressure (32MiB/4096 frames/16MiB buffered) only trips under sustained server non-drain.\n\nAgree with prior reviewer that the getStoredSettings writeJson try/catch removal is overload-only; verified it is in fact effectively unreachable due to the parsedPlainObject gate. Slice is whole-file lossless production code that passed full CI. Safe to open as a draft PR for human review; recommend a one-line human note on the localStorage silent-reset-above-limits behavior.

### Files
21 files changed (+2114 / −184).

---

### DIFF
--- a/src/renderer/src/web/web-e2ee.test.ts
+++ b/src/renderer/src/web/web-e2ee.test.ts
@@ -1,0 +1,32 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { MAX_E2EE_ENCRYPTED_BASE64_CHARACTERS } from '../../../shared/e2ee-crypto'
+import { decrypt, publicKeyFromBase64, WEB_E2EE_PUBLIC_KEY_MAX_BASE64_CHARACTERS } from './web-e2ee'
+
+afterEach(() => {
+  vi.unstubAllGlobals()
+})
+
+describe('web E2EE decode admission', () => {
+  it('rejects an oversized public key before base64 decoding', () => {
+    const atob = vi.fn(() => '')
+    vi.stubGlobal('window', { atob })
+
+    expect(() =>
+      publicKeyFromBase64('A'.repeat(WEB_E2EE_PUBLIC_KEY_MAX_BASE64_CHARACTERS + 1))
+    ).toThrow('encoded value is too large')
+    expect(atob).not.toHaveBeenCalled()
+  })
+
+  it('admits encrypted text at the boundary and rejects one extra character before decoding', () => {
+    const atob = vi.fn(() => '')
+    vi.stubGlobal('window', { atob })
+    const sharedKey = new Uint8Array(32)
+
+    expect(decrypt('A'.repeat(MAX_E2EE_ENCRYPTED_BASE64_CHARACTERS), sharedKey)).toBeNull()
+    expect(atob).toHaveBeenCalledOnce()
+    atob.mockClear()
+
+    expect(decrypt('A'.repeat(MAX_E2EE_ENCRYPTED_BASE64_CHARACTERS + 1), sharedKey)).toBeNull()
+    expect(atob).not.toHaveBeenCalled()
+  })
+})
--- a/src/renderer/src/web/web-e2ee.ts
+++ b/src/renderer/src/web/web-e2ee.ts
@@ -1,4 +1,7 @@
 import nacl from 'tweetnacl'
+import { MAX_E2EE_ENCRYPTED_BASE64_CHARACTERS } from '../../../shared/e2ee-crypto'
+
+export const WEB_E2EE_PUBLIC_KEY_MAX_BASE64_CHARACTERS = 44
 
 if (globalThis.crypto?.getRandomValues) {
   nacl.setPRNG((bytes, count) => {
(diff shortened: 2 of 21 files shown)


nwparker:

Issue #9558, opened 2026-07-20
[Feature]: Add IPv6 listener support for port 6768


kriskongx:
### Problem or use case

The Orca service currently listens only on IPv4:

There is no IPv6 listener (`[::]:6768`), so clients cannot connect through IPv6 addresses.

This affects users with IPv6-only networks or hosts that expose services through IPv6.

### Proposed solution

Please add support for binding the service to IPv6, preferably with a configurable bind address.

Example:

while keeping the current security defaults.

PR #10327, opened 2026-07-24
feat(mobile): support IPv6 pairing

kriskongx:
## Summary
- Adds IPv6 support to mobile pairing by binding WebSocket transport to IPv6 wildcard first with IPv4 fallback.
- Publishes bracketed IPv6 WebSocket endpoints when the listener binds to an IPv6 address.
- Includes non-link-local IPv6 interfaces in the mobile network interface picker.
- Allows manual IPv6 address overrides, including bracketed `[IPv6]:port` entries, while rejecting malformed, link-local, wildcard, and invalid-port inputs.
- Updates mobile pairing address copy across supported locales.

## Testing
- Added coverage for WebSocket host fallback and wildcard endpoint publishing.
- Added coverage for IPv6 interface listing and manual IPv6 address parsing.

## Related Issue

Related to #9558

## ELI5

Mobile pairing only really worked well on IPv4. Pairing now supports IPv6 addresses and endpoints so phones can connect on IPv6 networks.


AmethystLiang:
still reviewing

PR #10337, opened 2026-07-24
fix(codex): trust extended-length resume paths

yoke233:
## Summary

Fix Codex cold resume on Windows when a persisted rollout uses the extended-length spelling `\\?\C:\...` but its trusted Codex home uses the equivalent `C:\...` spelling.

The provenance check now creates a comparison-only copy for strict extended drive paths. File probing and the returned transcript path still use the original path. Other Win32 device namespaces remain rejected.

The final factory stack also preserves the newer fail-closed provenance guard: broad Codex-layout recognition remains separate from the stricter trusted-rollout grammar, so ADS-shaped rollout paths are reported as rejected Codex provenance but cannot become trusted resume sources.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` (full-repository lint not run; changed-file oxlint and max-lines passed)
- [ ] `pnpm typecheck` (full multi-target typecheck not run; node typecheck passed)
- [ ] `pnpm test` (full suite not run; decisive focused suites and headed E2E passed)
- [ ] `pnpm build` (production build not run; both headed lanes rebuilt in E2E mode)
- [x] `pnpm typecheck:node`
- [x] Changed-file `oxfmt --check` and `oxlint`
- [x] `node config/scripts/check-max-lines-ratchet.mjs`
- [x] `git diff --check`
- [x] Native Windows extended-path/security resolver matrix: 9 passed
- [x] Resume argv-drop suite: 10 passed
- [x] IPC unstrippable/drop-resume security cases: 4 passed
- [x] Renderer cold-restore/fresh-banner cases: 2 passed
- [x] Headed Electron cold-resume factory lane: `\\?\C:\...` rollout resumed after daemon death/restart and the provider session ID was visible in the terminal
- [x] Identical headed disabled oracle: the same rollout started fresh and the user-visible unavailable-resume banner appeared
- [x] Current-main baseline and disabled oracle both reproduce the fresh-session outcome; the factory changes it to resume

The broader resolver/preparation run still has 11 native-Windows failures from inherited POSIX-path fixtures; the same failures reproduce unchanged on pristine `main`. All decisive extended-path, ADS/device rejection, real-file, argv-drop, IPC fail-closed, and renderer resume cases pass.

## AI Review Report

Reviewed the complete four-file diff for ownership, security, performance, cleanup, native Windows behavior, SSH and folder-workspace neutrality, provider scope, and Git 2.25 compatibility. The Codex provenance resolver is the authoritative owner: the comparison fold belongs at the trust boundary, not in global path normalization or renderer/session transport.

The change adds no scanning, commands, IPC, network access, dependencies, or production file mutation. Non-namespace paths are returned unchanged, so macOS/Linux, SSH, remote runtimes, folder workspaces, and Git worktrees keep their existing behavior. No Git command or provider-review behavior changes.

## Security Audit

Only strict `\\?\<drive>:\...` paths can lose the prefix for containment comparison. `\\.\...`, `\\?\GLOBALROOT\...`, UNC/volume/device namespaces, different roots, traversal, invalid rollout layouts, ADS-shaped filenames, and missing/non-regular files remain rejected. Rejected originating provenance cannot fall back to scanning another home, and the original transcript path is retained for the regular-file check and resume result.

No workflow, action, package manifest, lockfile, dependency, download, credential, or permission surface changed.

## Notes

Windows-specific fix; no behavior change is intended on macOS or Linux. `\\?\UNC\...` remains intentionally unsupported at this boundary. The headed test uses a hermetic command stub rather than an authenticated Codex CLI, while exercising the real Electron persistence, daemon termination, restart, trust, and user-visible terminal path.

X (Twitter) handle: N/A

## ELI5

Windows can spell the same local file as both `C:\...` and `\\?\C:\...`. Orca now compares those spellings as the same trusted Codex rollout without weakening the checks that reject device paths, network namespaces, or alternate data streams.

### DIFF
--- a/src/main/codex/codex-session-resume-home.test.ts
+++ b/src/main/codex/codex-session-resume-home.test.ts
@@ -45,6 +45,55 @@ describe('resolveTrustedCodexSessionResumeHome', () => {
     ).toBe('c:\\users\\example\\.codex')
   })
 
+  it('accepts an extended-length Windows rollout and preserves its original path', async () => {
+    const homePath = 'C:\\Users\\Example\\.codex'
+    const transcriptPath =
+      '\\\\?\\C:\\Users\\Example\\.codex\\sessions\\2026\\07\\20\\rollout-session.jsonl'
+
+    await expect(
+      findTrustedCodexSessionResume({
+        sessionId: 'session-a',
+        transcriptPath,
+        trustedCodexHomes: [homePath],
+        fileIsRegular: (filePath) => filePath === transcriptPath
+      })
+    ).resolves.toEqual({ homePath, transcriptPath })
+  })
+
+  it('rejects unsafe or unrelated Windows namespace paths before probing files', () => {
+    const fileIsRegular = vi.fn((): boolean => true)
+    const trustedCodexHomes = ['C:\\Users\\Example\\.codex']
+    const rejectedPaths = [
+      '\\\\?\\D:\\Users\\Example\\.codex\\sessions\\2026\\07\\20\\rollout-a.jsonl',
+      '\\\\?\\GLOBALROOT\\Device\\HarddiskVolumeShadowCopy1\\sessions\\2026\\07\\20\\rollout-a.jsonl',
+      '\\\\?\\Volume{00000000-0000-0000-0000-000000000000}\\sessions\\2026\\07\\20\\rollout-a.jsonl',
+      '\\\\.\\C:\\Users\\Example\\.codex\\sessions\\2026\\07\\20\\rollout-a.jsonl',
+      '\\\\?\\C:\\Users\\Example\\.codex\\sessions\\2026\\07\\20\\..\\rollout-a.jsonl'
+    ]
+
+    for (const transcriptPath of rejectedPaths) {
+      expect(
+        resolveTrustedCodexSessionResumeHome({
+          transcriptPath,
+          trustedCodexHomes,
+          fileIsRegular
+        })
+      ).toBeNull()
+    }
+    expect(fileIsRegular).not.toHaveBeenCalled()
+  })
+
+  it('rejects a missing or non-regular extended-length rollout', () => {
+    expect(
+      resolveTrustedCodexSessionResumeHome({
+        transcriptPath:
+          '\\\\?\\C:\\Users\\Example\\.codex\\sessions\\2026\\07\\20\\rollout-a.jsonl',
+        trustedCodexHomes: ['C:\\Users\\Example\\.codex'],
+        fileIsRegular: () => false
+      })
+    ).toBeNull()
+  })
+
   it('rejects paths outside trusted homes or outside the rollout layout', () => {
     const fileIsRegular = vi.fn((): boolean => true)
     expect(
@@ -219,6 +268,24 @@ describe('resolveTrustedCodexSessionResumeHome', () => {
         trustedCodexHomes: ['/managed/origin/home', '/managed/other/home'],
         ...withoutHomeRanking,
         fileIsRegular: () => false,
+        listSessionFiles
+      })
+    ).resolves.toBeNull()
+    expect(listSessionFiles).not.toHaveBeenCalled()
+  })
+
+  it('does not scan another home after rejecting extended-length provenance', async () => {
+    const listSessionFiles = vi.fn((): AsyncIterable<string> => {
+      throw new Error('must not scan')
+    })
+
+    await expect(
+      findTrustedCodexSessionResume({
+        sessionId: '019f81b9-19a9-7651-a8d1-352d9420bd11',
+        transcriptPath:
+          '\\\\?\\D:\\Other\\.codex\\sessions\\2026\\07\\20\\rollout-019f81b9-19a9-7651-a8d1-352d9420bd11.jsonl',
+        trustedCodexHomes: ['C:\\Users\\Example\\.codex'],
+        fileIsRegular: () => true,
         listSessionFiles
       })
     ).resolves.toBeNull()
--- a/src/main/codex/codex-session-resume-home.test.ts
+++ b/src/main/codex/codex-session-resume-home.test.ts
@@ -94,8 +94,7 @@ describe('resolveTrustedCodexSessionResumeHome', () => {
     ).toBeNull()
   })
 
-  // Why: the extended spelling can appear on the trusted-home side too, so both
-  // sides of the containment comparison must fold, not just the rollout path.
+  // Why: the sessions root is folded too, so the home side must accept the extended spelling.
   it('accepts a normal-form rollout under an extended-length trusted home', () => {
     const homePath = '\\\\?\\C:\\Users\\Example\\.codex'
     expect(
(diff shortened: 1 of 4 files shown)


OrcaWin:
Taking a look!

OrcaWin:
Reviewed on real Windows 11 + WSL. The approach is right and I didn't change it: folding only the strict `\\?\<drive>:\` spelling for comparison, while keeping the original path for the file probe and the returned transcript path, is the correct shape. Normalizing *toward* the namespaced form would break the containment check's deliberate host-OS independence (it has to hold on POSIX hosts and for SSH/remote too).

Why it's safe: the relative path is slash-folded before `ROLLOUT_RELATIVE_PATH`, which forces exactly four components (`\d{4}/\d{2}/\d{2}/rollout-….jsonl[.zst]`). So `..` cannot survive the check, and it can't hide in the prefix either, because the prefix must string-equal `sessionsRoot` modulo case and separators. Every mismatch I could construct failed closed, never open.

*(Correcting an earlier version of this comment: I first argued the safety came from `\\?\` doing no normalization, so the probed path was the most literal reading. That premise is wrong — Node/libuv normalizes `\\?\` paths before the syscall, and I confirmed `lstatSync` succeeds on `\\?\…\20\..\20\…`. The layout regex above is the load-bearing reason, not path literalness. Relatedly, preserving the original spelling is a sound conservative choice rather than a long-path necessity — a plain 315-character path stats fine here.)*

### Verified by executing it on Windows, not by reading it

- **Reproduced the bug and confirmed the fix at the real seam.** In a live dev build, resuming a real Codex rollout at `\\?\C:\Users\<me>\.codex\sessions\...`:
  - without this change → `pty:spawn` fails with *"Orca could not verify the originating Codex session file…"*, thrown from `prepareCodexSessionResumeForLaunch` (`src/main/index.ts:854`)
  - with this change → resume spawns normally

  I checked the built main bundle on both runs to be sure the intended code was actually live.
- **Negative controls stayed rejected in those same live runs:** `\\.\C:\…` and a path outside every trusted home.
- **Tried to break the normalization against real files:** `\\.\`, `\\?\GLOBALROOT\Device\…`, `\\?\Volume{GUID}\`, `\\?\UNC\localhost\C$\`, `\\.\UNC\…`, plain `\\localhost\C$\`, `\\?\\C:\`, prefix-with-no-drive, and an ADS `:stream` suffix — all refused.
- **Long paths:** real 376/379-character paths on disk resolve through the extended form, including inside Electron's own runtime against the bundled module.
- **Real data:** 227 rollout files in my `~/.codex` — both spellings resolve to the same home, and the legacy id scan over that tree runs in ~74ms. Measured separately, the added per-entry work is ~44ns against ~326µs per entry, and that ratio is fixed at every tree size, so it isn't a hot-path concern.
- **WSL unaffected:** a file created inside WSL and read from Windows via `\\wsl.localhost\<distro>\…` still resolves, and the `wsl$` alias still folds. `src/main/index.ts:838` also returns early for `runtime === 'wsl'`.

### What I pushed

- **`e05932fb9`** — tests only. `codex-session-resume-home.test.ts` now covers the parts of the fold that weren't pinned: the trusted-root side (extended-length home + normal-form path), the `.zst` sibling derived from an extended-length persisted path (the one place the returned path is derived rather than echoed), and the legacy id-scan call site (extended entries accepted, device namespaces refused). All three fail with the production change reverted.
- **`0b8706b25`** — comment accuracy only, no behavior. The helper's comment described the rejected spellings as "arbitrary device namespaces", but `\\?\UNC\server\share\…` is the extended-length form of `\\server\share\…` — a network share, not a device, and `win32.toNamespacedPath` produces it from ordinary UNC input. Reading it as a security decision hid the fact that UNC simply isn't folded yet, so it now says what's actually excluded. Also trimmed my two new test comments to one line to match the file.

I also checked the four tests this PR already added: the positive one fails on revert, and each of the three rejection tests is caught by a distinct plausible mis-implementation of the helper. None are vacuous.

### Deliberately left alone

- **The `[\s\S]*` in the regex is correct — don't "simplify" it.** I tested the two arguments for narrowing it. Narrowing `[\\/]` to `\\` would be a regression: `\\?\C:/forward/slashes` *does* open on Windows (libuv normalizes), so the current form accepts a spelling that genuinely works. And swapping `[\s\S]*` for `.*` doesn't silently truncate a trailing newline — it fails to match at all. The authored version is the right choice.
- **Not renaming the helper.** "Trusted"/"Copy" are fair to quibble with, but it's subjective naming churn on a security-sensitive function.
- **Keeping the explicit `return null` rejections.** They're arguably redundant today — a `\\?\…`/`\\.\…` candidate can never prefix-match a programmatically-joined plain trusted home — but on a trust boundary I'd rather the rejection be explicit than rely on that invariant holding elsewhere.
- **Not hoisting** the loop-invariant sessions-root conversion out of the scan loop (see the numbers above), and note `codex-session-resume-home.ts:106-109` returns before scanning whenever a transcript path is set, so this PR's own scenario never reaches that loop.
- **The renderer `pty-connection.test.ts` fixture change stays**, though this one was genuinely split — one reviewer argued for reverting it, on the accurate grounds that the `\\?\` value is inert there (the transport is mocked, so the main-process check never runs) and the assertion overlaps the sibling at `pty-connection.test.ts:7656-7664`. Recording that for the next reader: what it actually covers is renderer sleeping-record plumbing, not Win32 path handling — the renderer just forwards an opaque string. I left it because it still pins the precondition and removing it only deletes coverage.
- CodeRabbit left no line comments to action.

### Known gaps / suggested follow-ups

- **`\\?\UNC\…` is still not folded, and that's the same bug for some users.** Where `homedir()` is a UNC roaming profile or the Codex home sits on a network share, the trusted home is `\\server\users\bob\.codex` while a canonicalizing Codex would report `\\?\UNC\server\users\bob\…` — rejected before *and* after this PR, so that cohort still gets the hard "could not verify" error. Out of scope here, but the natural next step.
- **Legacy shared-mirror migration silently no-ops for extended paths.** `src/main/index.ts:867` passes the still-`\\?\`-spelled transcript path into `prepareLegacySharedCodexSessionResume`; `win32.relative(plainRoot, extPath)` returns the absolute extended path unchanged, so `isLegacyRolloutRelativePath` (`codex-legacy-session-resume.ts:156`) sees `resolve(rel) === rel` and the call throws `RETRYABLE_RESUME_ERROR`, which `index.ts:875-881` swallows with a warn. Net effect on that lane: resume works but never migrates to the real home, and warns every time. Still strictly better than the pre-PR hard failure, and it resumes against the provenance-correct home, so not a blocker — but a conscious follow-up. My live test used the real home, so it didn't reach this branch.
- My `~/.codex` had no persisted `\\?\` transcript path, so I confirmed the *mechanism* (hook payloads stored verbatim; rejected provenance hard-fails) rather than watching the CLI emit that spelling. One logged real `transcript_path` from a Windows Codex would settle it.
- Pre-existing and out of scope: three tests in this file fail on Windows because they use POSIX fixtures that get win32-joined. They fail identically on `main`, but it does mean part of this file's Windows coverage is inert.
- Junction traversal inside a trusted home behaves identically with and without this change.

Nice fix — the containment reasoning and the fail-closed handling of the other namespaces were already right, which made this easy to verify.

Ready to merge from my side — scope unchanged (3 files, production change still the one helper), CI green on `0b8706b25`.

PR #10341, opened 2026-07-24
fix(terminal): reroute misrouted local spawns to the owning SSH host

PannenetsF:
## Summary

Fixes a bug where, after cloning a Git repo onto an **already-connected remote SSH server**, the first attempt to open a Terminal in the new project failed with **"Working directory … does not exist"** (文件路径不存在) and only started working after restarting the client.

Root cause: right after a fresh clone, the renderer can briefly resolve the new worktree to the **local** host (empty `connectionId`) before its SSH ownership settles. `pty:spawn` then runs on the **local** PTY provider with a **remote** cwd; the local machine has no such path, so `validateWorkingDirectory` throws ([`src/main/providers/local-pty-utils.ts`](https://github.com/stablyai/orca/blob/main/src/main/providers/local-pty-utils.ts)). This stayed broken until a restart re-loaded the SSH repo and routing resolved to SSH again.

Fix: the main process owns the authoritative repo catalog, so correct the route there. In the `pty:spawn` handler, when `connectionId` is empty but the worktree's repo is **unambiguously** owned by a **connected** SSH host, reroute the spawn to that SSH provider. Guarded to act only when exactly one repo owns the worktree id, it has a `connectionId`, and its SSH PTY provider is registered — otherwise behavior is unchanged.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`
- [x] Added a focused regression suite in `src/main/ipc/pty.test.ts` ("fresh-clone SSH spawn reroute") covering: reroute to the owning SSH connection, no reroute for a genuine local spawn, no reroute when the repo id is ambiguous across hosts, and no reroute when the SSH provider is not registered.

> Note: the lint/typecheck/test/build commands were **not run in the authoring environment** (no `node_modules`/`pnpm` available in this worktree). Please run the checklist in CI / an installed checkout. Targeted run: `pnpm vitest run src/main/ipc/pty.test.ts`.

Manual verification: connect a remote SSH host, clone a repo into it, and open a Terminal in the new project **without restarting** — it should attach to the remote shell in the cloned directory instead of erroring "Working directory does not exist".

## AI Review Report

Reviewed the change for correctness and blast radius:
- **Correctness:** the guard only rewrites `connectionId` when a single repo owns the worktree id, that repo has a non-empty `connectionId`, and the corresponding SSH PTY provider is currently registered. A local repo (no `connectionId`) is never rerouted; an ambiguous id (local + SSH sharing the id) is left untouched; an unregistered provider falls through to the existing "No PTY provider"/reconnect flow rather than a wrong local spawn.
- **Cross-platform (macOS/Linux/Windows):** the guard is platform-agnostic — pure string/route logic (`getRepoIdFromWorktreeId`, repo lookup, `sshProviders.has`). No shortcuts, labels, paths, shell behavior, or Electron platform APIs are touched. The rewrite happens before cwd-fallback and provider selection so all downstream platform handling is unchanged.
- Verified `resolveMisroutedSshConnectionId` runs before `getLocalPtyStartupPromise`, so a rerouted SSH spawn no longer waits on the local-daemon startup barrier.

## Security Audit

IPC input handling for `pty:spawn`: the guard only **narrows** routing to an SSH provider that is already registered and owned by a repo already present in the store — it never broadens filesystem or host access, and cannot route to an arbitrary/attacker-specified connection (the connection is derived from the stored repo, not from renderer input). No new command execution, path handling, auth, secret, or dependency surface is introduced.

## Notes

Scoped to the desktop renderer `pty:spawn` handler where the bug occurs; the runtime-owned controller spawn path (which receives an explicit resolved `connectionId`) is intentionally left unchanged.

## ELI5

Right after cloning a repo onto an already-connected SSH host, the first terminal could fail saying the path does not exist until restart. Local spawns that belong to SSH are re-routed to the owning host once ownership settles.

### DIFF
--- a/src/main/ipc/pty.test.ts
+++ b/src/main/ipc/pty.test.ts
@@ -889,6 +889,111 @@ describe('registerPtyHandlers', () => {
     }
   })
 
+  describe('fresh-clone SSH spawn reroute', () => {
+    const connectionId = 'ssh-1'
+    const worktreeId = 'repo-fresh::/home/user/proj'
+
+    function buildRerouteHarness(repos: { id: string; connectionId?: string }[]): {
+      sshSpawn: ReturnType<typeof vi.fn>
+      localSpawn: ReturnType<typeof vi.fn>
+    } {
+      const sshSpawn = vi.fn(async () => ({ id: 'ssh-pty' }))
+      const localSpawn = vi.fn(async () => ({ id: 'local-pty' }))
+      setLocalPtyProvider(createAgentClaimProvider({ spawn: localSpawn }) as never)
+      registerSshPtyProvider(connectionId, createAgentClaimProvider({ spawn: sshSpawn }) as never)
+      const store = {
+        getRepos: vi.fn(() => repos),
+        persistPtyBinding: vi.fn(),
+        upsertSshRemotePtyLease: vi.fn(),
+        removeSshRemotePtyLease: vi.fn(),
+        markSshRemotePtyLease: vi.fn()
+      }
+      handlers.clear()
+      registerPtyHandlers(
+        mainWindow as never,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        store as never
+      )
+      return { sshSpawn, localSpawn }
+    }
+
+    it('reroutes a local spawn to the owning SSH connection for a freshly-cloned repo', async () => {
+      const { sshSpawn, localSpawn } = buildRerouteHarness([{ id: 'repo-fresh', connectionId }])
+      await handlers.get('pty:spawn')!(null, {
+        cols: 80,
+        rows: 24,
+        cwd: '/home/user/proj',
+        connectionId: undefined,
+        worktreeId
+      })
+      expect(sshSpawn).toHaveBeenCalledTimes(1)
+      expect(localSpawn).not.toHaveBeenCalled()
+    })
+
+    it('leaves a genuine local spawn on the local provider', async () => {
+      const { sshSpawn, localSpawn } = buildRerouteHarness([
+        { id: 'repo-fresh', connectionId: undefined }
+      ])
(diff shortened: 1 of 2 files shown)


Jinwoo-H:
Thanks for working on this and for adding focused PTY coverage. We retested the exact PR head and its current-main rebase with a headed macOS Electron client, an ephemeral Linux/OpenSSH Docker target, a real repository, a fresh SSH clone, and the first visible terminal.\n\nCurrent main already passes that user flow. Baseline, rebased candidate, and the candidate with its production reroute removed all opened the terminal on the SSH host, rendered the marker, and wrote the proof file only inside the remote clone. Current routing now stamps host-qualified ownership before activation and fails closed while ownership is unresolved, so the PR fallback has no causal effect on the reported current-main behavior.\n\nThe fallback also introduces regressions. It scans and hydrates every repository for each connectionless spawn, infers authority from a bare repo ID, and can redirect a genuine local worktree to a stale SSH catalog row. The full candidate PTY suite had three failures plus an unhandled rejection; current main passed 401/401. When the SSH provider is unavailable, the added path still falls through to a local spawn rather than failing closed.\n\nBecause the reported flow is already fixed at the authoritative ownership boundary and this fallback can execute on the wrong host, we cannot land it and are closing the PR as superseded/rejected. A corrected follow-up would be justified by a fresh current-main reproduction and should resolve the exact worktree owner through host-qualified identity, including duplicate repo IDs, provider absence, and generation fencing, without a per-spawn catalog scan. The separate order-dependent duplicate-ID gap found during review should be addressed at that existing owner-resolution boundary.\n\nWe appreciate the contribution and the effort spent tracing a difficult remote-terminal failure.

PR #10351, opened 2026-07-24
fix(terminal): flush URL hover tooltip to pane bottom-left corner

nwparker:
## Summary
- Move terminal link URL hover styles into `.pane-link-tooltip` in `terminal.css` so corner offsets are not reintroduced via inline styles.
- Keep the tooltip on the pane (not the padded xterm host) at `bottom: 0; left: 0; margin: 0`.
- Square the bottom-left chrome (`border-radius: 0 4px 0 0`, no bottom/left border) so the hover preview sits flush against the pane corner (Ghostty-style).

## Test plan
- [x] `pnpm vitest run --config config/vitest.config.ts src/renderer/src/lib/pane-manager/pane-dom-creation.test.ts`
- [x] CSS geometry fixture: tooltipLeftFromPane=0, tooltipBottomFromPane=0 with exaggerated 24px content padding
- [ ] Manually hover a terminal URL and confirm the preview is flush bottom-left (no margin gap)

### DIFF
--- a/src/renderer/src/assets/terminal.css
+++ b/src/renderer/src/assets/terminal.css
@@ -506,3 +506,27 @@
   margin-top: var(--orca-pane-title-height);
   height: calc(100% - var(--orca-pane-title-height)); /* match margin-top */
 }
+
+/* Ghostty-style URL hover: glued to the pane's true bottom-left corner. */
+.pane-link-tooltip {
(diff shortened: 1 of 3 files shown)


nwparker:
### Attachments

| tooltip-corner-proof.png |
|---|
|  |

PR #10280, opened 2026-07-24
fix(floating-workspace): correct shortcut routing for tab switch and pane close

feelgom:
## Summary

- Fix keyboard shortcuts (Cmd+1/2, Ctrl+1/2) switching tabs in the main workspace instead of the floating workspace when the floating panel is focused.
- Fix Cmd+W closing the entire tab instead of just the focused split pane in the floating workspace.
- Closes #10288

## What Changed

**Main process** (`createMainWindow.ts`): Skip `jumpToWorktreeIndex`/`jumpToTabIndex` dispatch when the floating terminal input is focused, allowing the renderer's capture-phase handler to route them correctly.

**Floating panel** (`FloatingTerminalPanel.tsx`):
- Intercept `workspace.selectByIndex` in the capture handler to switch floating tabs.
- Defer `tab.close` to the terminal pane keyboard handler when the active terminal has split panes, so only the focused pane is closed.

## Test Plan

- [x] Floating workspace focused → Cmd+1/2 switches tabs within the floating panel (not main)
- [x] Floating workspace focused → Ctrl+1/2 switches tabs within the floating panel (not main)
- [x] Cmd+D to split pane in floating → Cmd+W closes only the focused pane
- [x] Last pane remaining → Cmd+W closes the tab
- [x] Single pane (no split) → Cmd+W closes the tab immediately
- [x] Main workspace shortcuts unaffected when floating panel is closed/unfocused
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] Existing tests pass (no regressions)

## Platform Notes

The bug and fix apply equally to all platforms — only the key combinations differ (Cmd on macOS, Ctrl on Windows/Linux). No platform-specific code paths were introduced.

No visual changes.

## AI Code Review Summary

- **Cross-platform**: Fix uses existing platform-aware keybinding matchers; no hardcoded modifier keys.
- **SSH/Remote**: No impact — shortcut routing is purely renderer/main process logic.
- **Agent/Integration**: No impact.
- **Performance**: Negligible — adds one store lookup (`terminalLayoutsByTabId`) on Cmd+W in a focused floating terminal.
- **Security**: No risk — no new IPC channels or external data handling.

---

🐦 @feelgom

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10359, opened 2026-07-24
fix(floating-workspace): route tab shortcuts to focused floating panel

### LINKED ISSUE #10288: [Bug]: Floating workspace shortcuts route to main workspace instead of floating panel
### Operating system

macOS

### Orca version

v1.4.152

## Bug Description

When the floating workspace panel is focused, keyboard shortcuts for tab/workspace switching and pane close behave incorrectly:

1. **Tab/Workspace switch shortcuts route to the main workspace** — Pressing Cmd+1/2 or Ctrl+1/2 while the floating workspace is focused causes the *main* workspace behind the panel to switch tabs/workspaces, instead of switching within the floating panel.

2. **Cmd+W closes the entire tab instead of a single pane** — After splitting a pane with Cmd+D in the floating workspace, pressing Cmd+W sometimes closes the entire tab rather than just the focused split pane.

### Bug 1: Tab/Workspace switching
1. Open the floating workspace
2. Create 2+ tabs in the floating workspace
3. Focus the floating workspace (click on the terminal or tab bar)
4. Press Cmd+1 or Cmd+2 (macOS) / Ctrl+1 or Ctrl+2 (Windows/Linux)
5. **Expected**: Tabs switch within the floating workspace
6. **Actual**: The main workspace behind the floating panel switches

### Bug 2: Pane close
1. Open the floating workspace with a terminal tab
2. Press Cmd+D to split the pane
3. Press Cmd+W to close the focused pane
4. **Expected**: Only the focused pane closes; tab remains with one pane
5. **Actual**: The entire tab closes (all panes destroyed)

### Bug 1
The main process `before-input-event` intercepts `workspace.selectByIndex` and `tab.selectByIndex` shortcuts, calling `event.preventDefault()` and dispatching IPC to the renderer — regardless of floating panel focus state. The renderer's floating panel capture-phase handler never sees the event.

Additionally, `floatingTerminalInputFocused` only tracks xterm textarea focus, not the broader panel (tab bar, chrome). When the tab bar holds focus, the guard doesn't activate.

### Bug 2
The floating panel's shortcut handler unconditionally matches `tab.close` and calls `closeFloatingItem()` (which closes the entire tab via `closeTab()`), consuming the event with `stopImmediatePropagation()` before the terminal pane keyboard handler can process it. Unlike the main workspace's `Terminal.tsx` which defers to the pane handler for split terminals, the floating panel has no split-pane awareness.

## Test Environment

- **OS**: macOS 15.5 (Sequoia)
- **Orca**: v1.4.152
- **Keyboard**: US layout

## Fix

PR #10280 addresses both issues:
- Adds `floatingPanelFocused` IPC to track panel-wide focus
- Skips main-process interception of indexed shortcuts when the floating panel is focused
- Intercepts `workspace.selectByIndex` in the floating panel's capture handler
- Defers `tab.close` to the terminal pane handler when split panes exist

## Platform Impact

Affects all platforms (macOS, Windows, Linux) — the key combinations differ but the routing logic is platform-common code.

innocarpe:
## Summary

- When the floating workspace is focused, Cmd/Ctrl+1–9 (and related digit chords) no longer switch tabs/workspaces in the **main** window behind the panel; they switch within the floating panel instead.
- Cmd/Ctrl+W on a split floating terminal closes only the focused pane (not the whole tab).
- Closes #10288

## Root cause

1. **Digit jumps**: Main-process `before-input-event` always dispatched `jumpToWorktreeIndex` / `jumpToTabIndex` and `preventDefault`’d the key, so the floating panel’s capture handler never saw it. `floatingTerminalInputFocused` only skipped sidebar toggles and only tracked xterm focus—not tab-bar/chrome focus.
2. **Pane close**: Floating panel’s `tab.close` handler always called `closeFloatingItem()` (whole tab) with `stopImmediatePropagation`, so the split-pane close path never ran.

## Fix

- Add `ui:setFloatingPanelFocused` IPC so main knows when *any* floating panel focus exists (chrome or terminal).
- Skip main interception of indexed jumps when floating terminal input **or** floating panel is focused.
- Remap `workspace.selectByIndex` in the floating panel capture handler to activate floating tabs.
- Defer `tab.close` to the terminal pane handler when the active floating terminal layout is a split.

## Test plan

- [x] Unit: main skips `jumpToWorktreeIndex` / `jumpToTabIndex` when floating terminal input is focused
- [x] Unit: main skips the same when only floating panel chrome is focused
- [x] Unit: `shouldDeferFloatingWorkspaceTabCloseToPane` only true for split terminal + terminal input
- [x] Existing `FloatingTerminalPanel` / `createMainWindow` suites pass
- [x] `pnpm typecheck` passes
- [ ] Manual: open floating workspace, 2+ tabs, focus panel → Cmd/Ctrl+1/2 switches **floating** tabs only
- [ ] Manual: Cmd/Ctrl+D split in floating → Cmd/Ctrl+W closes focused pane only; last pane closes tab

## Platform notes

Cross-platform: uses existing keybinding matchers (Cmd on macOS, Ctrl/Alt elsewhere). No platform-specific branching beyond existing shortcut tables.

### DIFF
--- a/src/main/window/createMainWindow.test.ts
+++ b/src/main/window/createMainWindow.test.ts
@@ -2352,6 +2352,176 @@ describe('createMainWindow', () => {
     expect(webContents.send).toHaveBeenCalledWith('ui:openNewWorkspace')
   })
 
+  it('skips jumpToWorktreeIndex and jumpToTabIndex when floating terminal input is focused', () => {
+    const windowHandlers: Record<string, (...args: any[]) => void> = {}
+    const webContents = {
+      on: vi.fn((event, handler) => {
+        windowHandlers[event] = handler
+      }),
+      setZoomLevel: vi.fn(),
+      setBackgroundThrottling: vi.fn(),
+      invalidate: vi.fn(),
+      setWindowOpenHandler: vi.fn(),
+      send: vi.fn(),
+      isDevToolsOpened: vi.fn(),
+      openDevTools: vi.fn(),
+      closeDevTools: vi.fn()
+    }
+    const browserWindowInstance = {
+      webContents,
+      on: vi.fn(),
+      isDestroyed: vi.fn(() => false),
+      isMaximized: vi.fn(() => true),
+      isFullScreen: vi.fn(() => false),
+      getSize: vi.fn(() => [1200, 800]),
+      setSize: vi.fn(),
+      maximize: vi.fn(),
+      show: vi.fn(),
+      loadFile: vi.fn(),
+      loadURL: vi.fn()
+    }
+    browserWindowMock.mockImplementation(function () {
+      return browserWindowInstance
+    })
+
+    createMainWindow(null)
+
+    const setFocusedListener = vi
+      .mocked(ipcMain.on)
+      .mock.calls.find(([channel]) => channel === 'ui:setFloatingTerminalInputFocused')?.[1]
+    expect(setFocusedListener).toBeTypeOf('function')
+    setFocusedListener?.({ sender: webContents } as never, true)
+
+    const isDarwin = process.platform === 'darwin'
+
+    const preventDefault1 = vi.fn()
+    windowHandlers['before-input-event'](
+      { preventDefault: preventDefault1 } as never,
+      {
+        type: 'keyDown',
+        code: 'Digit1',
+        key: '1',
+        meta: isDarwin,
+        control: !isDarwin,
+        alt: false,
+        shift: false
+      } as never
+    )
+    expect(preventDefault1).not.toHaveBeenCalled()
+    expect(webContents.send).not.toHaveBeenCalledWith('ui:jumpToWorktreeIndex', expect.anything())
+
+    const preventDefault2 = vi.fn()
+    windowHandlers['before-input-event'](
+      { preventDefault: preventDefault2 } as never,
+      isDarwin
+        ? ({
+            type: 'keyDown',
+            code: 'Digit1',
+            key: '1',
+            meta: false,
+            control: true,
+            alt: false,
+            shift: false
+          } as never)
+        : ({
+            type: 'keyDown',
+            code: 'Digit1',
+            key: '1',
+            meta: false,
+            control: false,
+            alt: true,
+            shift: false
+          } as never)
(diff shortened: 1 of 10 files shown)


feelgom:
Hey @innocarpe — thanks for picking this up! Looks like we landed on the same fix independently and ended up duplicating effort here. Wanted to flag #10280 so we can consolidate rather than split the work.

**Quick context:**
- #10280 opened this fix ~6h earlier (03:16 UTC vs. 09:14 UTC) and has already been through a CodeRabbit review pass — including the follow-up that extended the focus guard from the terminal input to the whole floating panel (commit `e9beccd`).
- @nwparker also verified the repro on 1.4.154 and suggested landing #10280 in the issue thread.

**The two PRs are essentially the same approach.** Both:
- add a `ui:setFloatingPanelFocused` IPC channel and skip main-process interception of `jumpToWorktreeIndex` / `jumpToTabIndex` when the floating panel (or its terminal input) is focused,
- remap `workspace.selectByIndex` in the floating panel's capture handler to switch floating tabs, and
- defer `tab.close` to the pane handler when the active floating terminal is a split.

The main-process guard, the preload/web-preload signatures, and the `setFloatingPanelFocusedInMain` helper are basically identical between them. Your PR additionally factors some of this into `floating-workspace-shortcut-policy` / `floating-workspace-terminal-actions` helpers with extra tests, which I think is a nice improvement.

**Proposal — happy with whichever keeps us to one PR:**
1. Land #10280 as the base (it's earlier and already reviewed), and I'll fold in your helper refactor + added tests, crediting you via a `Co-authored-by:` trailer; **or**
2. Go with #10359's structure, but add a `Co-authored-by:` trailer crediting me so the original work is preserved.

Either way works for me — just want to make sure we don't lose one of the two contributions. What do you prefer? cc @AmethystLiang @nwparker

AmethystLiang:
Closing as a duplicate of #10433, which is the maintained fix for #10288 (floating workspace shortcuts routing to the floating panel rather than the main window).

Thanks for the contribution — please track https://github.com/stablyai/orca/pull/10433 for the merge.

PR #10360, opened 2026-07-24
fix(github): resolve owner/repo through SSH Host aliases

### LINKED ISSUE #10284: [Bug]: GitHub PR merge fails when origin uses an SSH config Host alias
### Operating system

macOS

### Orca version

1.4.152

#### Short summary

GitHub PR merge fails with:

```text
Could not resolve GitHub owner/repo for this repository
```

when the repository's remote uses an SSH config `Host` alias, which is a common setup for using multiple GitHub accounts.

#### Reproduction

1. Configure an SSH alias for a GitHub account:

   ```sshconfig
   Host github-work
       HostName ssh.github.com
       Port 443
       User git
       IdentityFile ~/.ssh/id_ed25519_work
       IdentitiesOnly yes
   ```

2. Use that alias in a repository remote:

   ```bash
   git remote set-url origin git@github-work:OWNER/REPOSITORY.git
   ```

3. Verify normal Git access succeeds:

   ```bash
   git ls-remote origin HEAD
   ```

4. Ensure GitHub CLI is authenticated to `github.com`, then open the repository and one of its PRs in Orca.
5. Click **Squash and merge**.

#### Actual behavior

Orca shows:

```text
Could not resolve GitHub owner/repo for this repository
```

The PR can otherwise be displayed and its checks can pass.

#### Expected behavior

Orca should recognize that the SSH alias resolves to GitHub.com and perform the merge using `OWNER/REPOSITORY`.

#### Suspected cause

The current remote parser preserves the SCP-style remote host verbatim:

```ts
const sshMatch = trimmed.match(/^git@([^:]+):([^/]+)\/([^/]+?)(?:\.git)?$/i)
```

`normalizeGitHubRemoteHost()` only maps the literal `ssh.github.com` host to `github.com`. `parseGitHubOwnerRepo()` then rejects every host that is not literally `github.com`:

```ts
if (!identity || identity.host.toLowerCase() !== 'github.com') {
  return null
}
```

Source:

- https://github.com/stablyai/orca/blob/main/src/main/github/github-remote-identity-parsing.ts
- Related earlier work: https://github.com/stablyai/orca/pull/3605
- The same literal-host assumption was previously identified for GHES in https://github.com/stablyai/orca/pull/8603, but an SSH config alias is not an authenticated GHES host and is still unresolved.

For a remote such as `git@github-work:OWNER/REPOSITORY.git`, Orca therefore derives `host=github-work` and returns `null`, even though OpenSSH resolves it to `ssh.github.com`.

#### Possible approaches

- Resolve SCP/SSH remote hosts through the user's SSH configuration (for example, obtain the effective `hostname` from `ssh -G <alias>`) before classifying the forge host.
- Use repository identity resolved by Git/GitHub CLI rather than requiring the literal host in `remote.origin.url`.
- Provide an explicit per-repository hosted-provider/host override.

Whichever approach is chosen should retain the original alias for Git transport so the correct account-specific SSH key remains selected.

#### Suggested regression test

Add coverage showing that a remote like:

```text
git@github-work:team/orca.git
```

whose SSH configuration resolves `github-work` to `ssh.github.com`, produces:

```ts
{ owner: 'team', repo: 'orca' }
```

and allows PR mutations such as merge.

#### Workaround

Using a literal `github.com` remote together with a repository-local `core.sshCommand` works, but it requires users to replace an otherwise valid multi-account SSH configuration:

```bash
git remote set-url origin git@github.com:OWNER/REPOSITORY.git
git config --local core.sshCommand \
  'ssh -i ~/.ssh/id_ed25519_work -o IdentitiesOnly=yes'
```

innocarpe:
## Summary

- After a literal `github.com` host parse misses, expand SSH config **Host → HostName** (then `ssh -G` when needed) so multi-account remotes like `git@github-work:owner/repo` resolve to github.com.
- Wire the expansion into `getOwnerRepoForRemote` (local host only — SSH-provider remotes stay on the literal parser).
- Leave the remote URL unchanged for Git transport so the correct `IdentityFile` still applies.

Fixes #10284.

## Why

`parseGitHubOwnerRepo` rejects any host that is not literally `github.com`. SSH config Host aliases are common for multiple GitHub accounts; HostName is already `github.com` / `ssh.github.com`, but Orca never expanded it, so Squash and merge failed with “Could not resolve GitHub owner/repo for this repository” even though the PR UI worked.

## Test plan

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/github/github-remote-identity-parsing.test.ts` (19)
- [ ] Manual: `git@github-work:OWNER/REPO` remote + Squash and merge succeeds when HostName maps to github.com

## Screenshots

No visual change.

### DIFF
--- a/src/main/github/github-owner-repo-ssh-alias.ts
+++ b/src/main/github/github-owner-repo-ssh-alias.ts
@@ -1,0 +1,54 @@
+import type { GitHubOwnerRepo } from '../../shared/types'
+import { loadUserSshConfig } from '../ssh/ssh-config-parser'
+import { resolveWithSshG } from '../ssh/ssh-g-config-resolution'
+import {
+  hostnameFromSshConfigHosts,
+  parseGitHubOwnerRepo,
+  parseGitHubOwnerRepoWithResolvedHostname,
+  parseGitHubRemoteIdentity
+} from './github-remote-identity-parsing'
+
+// Why: skip ssh -G for already-qualified hostnames (GHES / public forges) —
+// they are not Host aliases and spawning ssh on every miss is pure cost.
+function isLikelyLiteralHost(host: string): boolean {
+  return host.includes('.')
+}
+
+async function resolveSshAliasHostname(alias: string): Promise<string | null> {
+  // Why: ~/.ssh/config Host → HostName covers the common multi-account alias
+  // case without spawning ssh; fall back to `ssh -G` for Include/Match.
+  const fromConfig = hostnameFromSshConfigHosts(alias, loadUserSshConfig())
+  if (fromConfig) {
+    return fromConfig
+  }
+  if (isLikelyLiteralHost(alias)) {
+    return null
+  }
+  const resolved = await resolveWithSshG(alias)
+  const hostname = resolved?.hostname?.trim()
+  return hostname ? hostname : null
+}
+
+/**
+ * Resolve github.com owner/repo, expanding SSH config Host aliases when the
+ * remote host token is not literally github.com (#10284).
+ */
+export async function resolveGitHubOwnerRepoFromRemoteUrl(
+  remoteUrl: string
+): Promise<GitHubOwnerRepo | null> {
+  const direct = parseGitHubOwnerRepo(remoteUrl)
+  if (direct) {
+    return direct
+  }
+  const identity = parseGitHubRemoteIdentity(remoteUrl)
+  if (!identity) {
+    return null
+  }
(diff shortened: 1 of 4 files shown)


nwparker:
Thanks for jumping on this and for the initial SSH-alias implementation. #10361 has now merged with the same core fix plus runtime-scoped resolution for native, WSL, and SSH workspaces; retryable failure handling; case-sensitive alias caching; reconnect isolation; and GHES coverage. Closing this as superseded.

Issue #4570, opened 2026-06-03
[Feature]: Improve Kimi Code support (status, usage, token analytics)


ZinkLu:
### Problem or use case

Kimi is listed as a supported agent in Orca, but compared to Claude Code and Codex, a few things that those agents get out of the box are missing for Kimi:

1. **Agent status** — the little agent icon in the worktree/agent row doesn't reliably reflect what Kimi is actually doing (working / idle / waiting for input). For Claude Code this updates live; for Kimi it doesn't.

2. **Subscription usage in the status bar** — the bottom status bar shows your subscription usage for Claude, Codex, etc., but nothing for Kimi.

3. **Token analytics in Settings** — Settings can break down how many tokens your sessions used for Claude and Codex, but Kimi sessions don't show up there at all.

**Background — and why this only targets the new Kimi Code.** Moonshot has moved on from the original Python CLI to a new TypeScript one:

- Old (Python, retired): **kimi-cli** — https://github.com/MoonshotAI/kimi-cli — stopped updating at v1.44.0 (2026-05-13).
- New (TypeScript, active): **Kimi Code** — https://github.com/MoonshotAI/kimi-code

This request is scoped to **Kimi Code only.** The old Python `kimi-cli` is end-of-life, so there's no point building support against it — everything here should target the new Kimi Code.

### Proposed solution

Bring Kimi Code up to the same level as Claude Code / Codex on these three things: live agent status on the icon, subscription usage in the status bar, and per-session token analytics in Settings.

### Alternatives or additional context

The three are independent — live status is probably the easiest to land first. Usage and token analytics depend on how Kimi Code exposes its data (session files / usage endpoint); since Kimi Code is a fresh TypeScript rewrite, that may differ from the old Python CLI and is worth a quick check against a current install.

AmethystLiang:
yes will add them 👌

ZinkLu:
Tasks 1 & 2 are done (PRs submitted). 
Task 3: Kimi's JSONL has no token usage field, so can't add it now. 

BTW, Kimi is the coding agent I use most often besides Claude. It would be great to give it first-class support in Orca, but that requires more thought. I'll likely start a new issue to discuss it.

ZinkLu:
@AmethystLiang 
Can we close this issue, since #5783 and #5836 have implemented all the features?

brennanb2025:
Taking a look 👀

uiYzzi:
[...]

PR #10369, opened 2026-07-24
fix(kimi): wait for welcome banner before stdin prompt injection

### LINKED ISSUE #10336: [Bug]: `--agent kimi --prompt` injection never reaches Kimi Code input — prompt echoes above welcome banner (v1.4.152)
### Operating system

macOS

### Orca version

v1.4.152

### Agent

Kimi Code CLI v0.29.0 (`--agent kimi`)

### Summary

`orca worktree create --agent kimi --prompt "..."` **never** successfully injects the prompt into Kimi Code. Every time a worktree is created this way, the prompt text is written to the PTY before Kimi Code's TUI is ready to accept input: it appears as plain terminal output **above** the `Welcome to Kimi Code!` banner instead of landing in the input box. The agent then sits idle at an empty prompt until the user manually copies the echoed text and pastes it into the agent. This makes unattended multi-agent workflows (one agent spawning worker agents) unusable with Kimi.

This is not an intermittent race that sometimes wins — prompt injection into Kimi Code has not worked once in our usage.

### Steps to reproduce

1. From any terminal (e.g. from inside a running agent coordinating workers), create a worktree with a prompt:

```bash
PROMPT='You are Agent J. First run `orca linear issue ZEN-78 --full --json` ... (multi-line task brief)'

orca worktree create \
  --repo id:<repoId> \
  --name agent-j-uuid-migration \
  --no-parent \
  --agent kimi \
  --prompt "$PROMPT" \
  --json
```

2. Observe the new worktree's terminal.

### Expected behavior

Kimi Code starts in the new worktree, the prompt is delivered into the TUI input box and submitted, and the agent begins working autonomously.

### Actual behavior

The prompt text is echoed as scrollback **above** the welcome banner:

```
<prompt text rendered as terminal output here>
╭─────────────────────────────────────────╮
│  ▐█▛█▛█▌  Welcome to Kimi Code!          │
│  ...                                     │
╰─────────────────────────────────────────╯
```

The Kimi TUI never consumed the text — the input box is empty and no turn starts. The prompt is effectively lost until manually re-entered.

### Analysis / suspected root cause

This looks like the same readiness race as #9976 ("slow Claude startup lets tui-idle false-positive and loses injected dispatch", fixed for Claude): prompt injection is gated on a readiness signal that fires before the agent TUI actually accepts input. For Kimi Code the failure is deterministic rather than occasional — the prompt is always written while Kimi Code is still in its startup phase, so the bytes are echoed to scrollback instead of being read into the TUI's input buffer. This suggests Orca currently has no working readiness signal for Kimi Code at all (see also #4570, which tracks missing/unreliable Kimi agent-status reporting).

### Workaround

Create the worktree without `--prompt`, wait until the Kimi welcome banner is fully rendered, then send the prompt explicitly:

```bash
orca worktree create --repo id:<repoId> --name <task> --no-parent --agent kimi --json
orca terminal wait --terminal <handle> --for tui-idle --timeout-ms 60000 --json
orca terminal send --terminal <handle> --text "$PROMPT" --enter --json
```

(Note: `tui-idle` itself was the false-positive source in #9976, so a more reliable readiness check for Kimi — e.g. detecting the rendered input prompt — may be needed.)

innocarpe:
## Description

Kimi Code `stdin-after-start` prompt injection now waits for the **Welcome to Kimi** banner (and a longer readiness budget) before writing the follow-up prompt. Orchestration `tui-idle` also treats that banner as ready and no longer false-positives on a quiet `kimi` process alone.

## Focused fix

- In scope: Kimi welcome-banner readiness for worktree-create follow-up prompts + orchestration `tui-idle` readiness for the same marker
- Out of scope: Kimi agent-status OSC / usage analytics (#4570); changing `dispatch_input: accepted` to mean “agent consumed bytes” rather than “write accepted”

## Preserves

- Other agents’ readiness signals and the existing 8s draft-paste quiet timeout for paste-settle paths
- `worker-start` / `sendTerminalAgentPrompt` write contract (`accepted` still means the write path completed)

## Evidence

- Tests: `pnpm exec vitest run --config config/vitest.config.ts src/shared/draft-paste-ready-scanner.test.ts src/shared/tui-agent-startup.test.ts` (83)
- Tests: `pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/orca-runtime.test.ts -t 'tui-idle|Kimi|#10336'` (including banner ready + quiet-kimi not idle)
- Field: [issuecomment-5204525896](https://github.com/stablyai/orca/issues/10336#issuecomment-5204525896) on v1.4.175 remote orchestration (banner signal missing on release; quiet race + 8s budget)

## User-regression-tradeoffs

- On timeout after process start without banner, worktree follow-up still **skips** the prompt (warn log) rather than writing into scrollback
- Kimi `tui-idle` waits for the banner instead of process quiet — slower ready on hosts where Kimi is quiet mid-boot, but avoids silent empty dispatch

Fixes #10336

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -15108,8 +15108,14 @@ describe('OrcaRuntimeService', () => {
         condition: 'tui-idle',
         timeoutMs: 10_000
       })
-      // Why: banner is no longer an immediate known-ready match — quiet after welcome.
-      await vi.advanceTimersByTimeAsync(5_000)
+      let resolved = false
+      void waitPromise.then(() => {
+        resolved = true
+      })
+      // Why: quiet window is 3s; stay just under it so immediate banner resolve fails the test.
+      await vi.advanceTimersByTimeAsync(2_500)
+      expect(resolved).toBe(false)
+      await vi.advanceTimersByTimeAsync(2_500)
 
       await expect(waitPromise).resolves.toMatchObject({
         handle,
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -15087,28 +15087,69 @@ describe('OrcaRuntimeService', () => {
     })
   })
 
-  it('resolves tui-idle from a Kimi welcome banner preview (#10336)', async () => {
-    const runtime = new OrcaRuntimeService(store)
-    runtime.setPtyController({
-      spawn: vi.fn().mockResolvedValue({ id: 'pty-kimi' }),
-      write: () => true,
-      kill: () => true,
-      getForegroundProcess: async () => 'kimi'
-    })
-    const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
-    runtime.onPtyData(
-      'pty-kimi',
-      '╭─────────────────────────────────────────╮\n│  Welcome to Kimi Code!                   │\n╰─────────────────────────────────────────╯\n',
-      Date.now()
-    )
+  it('resolves tui-idle after Kimi welcome banner then process quiet (#10336/#10369)', async () => {
+    vi.useFakeTimers()
+    try {
+      const runtime = new OrcaRuntimeService(store)
+      runtime.setPtyController({
+        spawn: vi.fn().mockResolvedValue({ id: 'pty-kimi' }),
+        write: () => true,
+        kill: () => true,
+        getForegroundProcess: async () => 'kimi'
+      })
+      const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+      runtime.onPtyData(
+        'pty-kimi',
+        '╭─────────────────────────────────────────╮\n│  Welcome to Kimi Code!                   │\n╰─────────────────────────────────────────╯\n',
+        Date.now()
+      )
 
-    await expect(
-      runtime.waitForTerminal(handle, { condition: 'tui-idle', timeoutMs: 1_000 })
-    ).resolves.toMatchObject({
-      handle,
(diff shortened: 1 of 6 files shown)


innocarpe:
Follow-up after [field report on #10336](https://github.com/stablyai/orca/issues/10336#issuecomment-5204525896):

- Rebased onto latest `main`
- `STARTUP_FOLLOWUP_DRAFT_READY_TIMEOUT_MS = 20s` for the banner gate (keep 8s for paste-quiet)
- `tui-idle`: recognize `Welcome to Kimi`; do not quiet-fallback idle while `kimi` is mounting

Evidence: draft-paste-ready + tui-agent-startup (83) and new/updated `tui-idle` Kimi cases in `orca-runtime.test.ts`.

innocarpe:
Addressed CodeRabbit major: `Welcome to Kimi` is **no longer** a global `isKnownReadyPromptPreview` match (that could satisfy later `tui-idle` waits while generation continued and the banner stayed in the tail).

Instead:
- stdin follow-up still uses `kimi-welcome-banner` draft-ready scanner
- Kimi quiet-idle requires a once-seen welcome banner flag (`kimiWelcomeBannerSeen`), then normal quiet quiescence for later turns

Regression: banner+quiet resolves; banner with continuous output does not; quiet before banner does not.

innocarpe:
Addressed latest CodeRabbit notes:

1. Stamp `kimiWelcomeBannerSeen` when `waitForStartupDraftReady` hits `kimi-welcome-banner`
2. Reset that flag in `resetTrackedTerminalStateForProviderGeneration` (reused ptyId / new process)
3. Test now asserts still pending before quiet elapses, then resolves after the quiet window

tip: `0563058650`

PR #9687, opened 2026-07-21
fix(terminal): make remote agent sessions host-authoritative

OrcaWin:
## Summary

Make the remote Orca host authoritative for agent-session ownership and terminal membership.

- prevent two clients from waking the same live remote agent into duplicate PTYs
- retire exited terminal surfaces durably so stale client snapshots cannot resurrect ghost tabs
- negotiate the structured authority path only when the client, runtime, daemon, and relay support it
- preserve the exact legacy launch behavior at safe pre-side-effect mixed-version boundaries
- integrate #9804's non-destructive reconnect lifecycle contract without weakening old-client user closes
- add a deterministic real remote-server reproduction harness and document the compatibility contract
- register the experimental reliability class `agent-session.remote-host-authority`

Fixes #8878.
Related to #9352 and #9804.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — repository-wide baseline failures remain in untouched skill-freshness and generated skill-manifest files; task-owned oxlint, reliability, and max-lines checks pass
- [x] `pnpm typecheck`
- [x] `pnpm test` — 3,213 files passed; 33,949 tests passed in the pre-review full-root validation
- [x] Build-backed remote reproduction — `pnpm test:repro:remote-agent-session`
- [x] Added or updated high-quality regression tests

Final post-rebase validation:

- remote host-authority gate: 10 files, 325 tests
- #9804 reconnect/daemon-generation gate: 9 files, 1,010 tests
- real headless `orca serve` harness over encrypted WebSocket pairing: one physical spawn across two racing clients, canonical retry adoption, durable exit retirement, stale-publication rejection, and no resurrection after server restart
- all three TypeScript typechecks
- reliability manifest validation for 46 gates
- touched-file oxlint and targeted oxfmt checks
- `git diff --check`

The Windows daemon harness now derives explicit v24 clean-disconnect, v25 startup-ingress, and v26 agent-authority boundaries and covers v21–v26. Recorded native Windows evidence remains v21–v25; the six-generation v26 native rerun has not been collected locally.

## AI Review Report

The final combined diff was independently re-reviewed for lifecycle, protocol compatibility, and topology after rebasing onto #9804. All three specialist reviews returned clean.

[...]

PR #10370, opened 2026-07-24
fix(runtime): headless terminal create --focus without renderer window

### LINKED ISSUE #10333: `terminal create --focus` throws "No renderer window available" on headless `orca serve` — Mac UI + button cannot launch agents on remote servers
### Summary

When a paired Remote Orca Server runs headless (`orca serve` on Linux), clicking the `+` button in the Mac desktop client and choosing any agent (Hermes, OpenCode, Codex, Claude Code, etc.) fails with:

```
No renderer window available
```

The error is thrown by the **runtime** before any agent command executes. This is not an agent-specific issue — it affects ALL agents launched via the `+` button on a headless remote server.

### Environment

- **Mac client**: Orca v1.4.152 (macOS)
- **Remote server**: Orca v1.4.152 (`orca serve` on Ubuntu VPS, headless, Cloudflare Tunnel)
- **Pairing**: Valid, connected, runtimeId matches

### Reproduction

```bash
# CLI reproduction (same code path as Mac UI + button):
orca --environment <remote-env> terminal create \
  --worktree <worktree-selector> \
  --command "echo test" \
  --focus --json
```

**Result (before fix):**
```json
{
  "ok": false,
  "error": {
    "code": "runtime_error",
    "message": "No renderer window available"
  }
}
```

**Without `--focus`** (works, but terminal is background-only):
```json
{
  "ok": true,
  "result": {
    "terminal": {
      "handle": "term_...",
      "surface": "background",
      "warning": "Terminal ... is running, but Orca could not make it discoverable."
    }
  }
}
```

### Root Cause

In `createTerminal()` (main process bundle, `out/main/index.js`), the `shouldCreateInBackground` condition has a headless fallback for `rendererBacked` terminals but **not** for `focus`-requested terminals:

```javascript
const requiresRendererFocus = opts.presentation === "focused" || opts.focus === true;
const availableAuthoritativeWindow = this.getAvailableAuthoritativeWindow();
const rendererWindow = opts.rendererBacked === true ? availableAuthoritativeWindow : null;

const shouldCreateInBackground = worktreeSelector !== void 0 && (
    Boolean(opts.agentSessionClaim) ||
    !requiresRendererFocus && opts.rendererBacked !== true ||
    // Why: `orca serve` exposes the local runtime without a renderer
    // window. Renderer-backed Codex terminals are preferred for the app,
    // but headless CLI users still need a usable terminal handle.
    opts.rendererBacked === true && rendererWindow === null
    // ← MISSING: no fallback when requiresRendererFocus === true && no window
);
```

When `focus: true` is passed (which the Mac UI `+` button does), `requiresRendererFocus` is `true`. This makes the second condition (`!requiresRendererFocus && ...`) evaluate to `false`. Since `rendererBacked` is not `true` (the Mac UI doesn't set it), the third condition is also `false`. So `shouldCreateInBackground` is `false`, and the code falls through to the renderer-backed path:

```javascript
const win = rendererWindow ?? this.getAuthoritativeWindow();  // ← THROWS on headless serve
```

`getAuthoritativeWindow()` throws `"No renderer window available"` because `authoritativeWindowId` is `null` in serve mode (no BrowserWindows are created).

### Impact

- **Mac UI `+` button**: Completely broken for ALL agents on headless Remote Orca Servers
- **CLI `terminal create --focus`**: Fails with same error
- **CLI `terminal create` (no `--focus`)**: Works, but terminal is background-only and "not discoverable"
- **Workaround**: None available from the Mac UI

### Proposed Fix

Add one OR condition to `shouldCreateInBackground` so that focus-requested terminals fall back to background creation when no authoritative window exists (headless serve):

```javascript
const shouldCreateInBackground = worktreeSelector !== void 0 && (
    Boolean(opts.agentSessionClaim) ||
    !requiresRendererFocus && opts.rendererBacked !== true ||
    opts.rendererBacked === true && rendererWindow === null ||
    requiresRendererFocus && availableAuthoritativeWindow === null  // ← NEW
);
```

This mirrors the existing `rendererBacked` headless fallback: when the runtime has no renderer window, focus-requested terminals are created in background mode instead of throwing. The terminal's `notifier?.revealTerminalSession?.()` path (using optional chaining) then handles client-side discovery gracefully.

### Why This Used to Work

The `requiresRendererFocus` variable and the `shouldCreateInBackground` branching logic were likely introduced or changed in a recent version. The headless fallback comment ("`orca serve` exposes the local runtime without a renderer window") shows awareness of the headless case, but the `focus` path was not covered. Both our Mac client and VPS server were upgraded to v1.4.152 on July 23, 2026 — the same day this version was released.

### Related

- PR #9790 (`fix(agents): scope Settings agent list and quick-launch menu to the remote-server host`) — merged after v1.4.152, addresses agent detection but not terminal creation focus
- Issue #7442 (background-surface TUI hang) — related but different issue (hang vs. error)

innocarpe:
## Summary
- Fixes #10333: Mac UI `+` agent launch and `terminal create --focus` on headless `orca serve` no longer throw `No renderer window available`.
- When there is no authoritative BrowserWindow, create with a worktree selector uses the existing background spawn path (same idea as the current `rendererBacked` headless fallback).
- Pure decision helper + unit tests for focus/headless and related matrix.

## Root cause
`shouldCreateInBackground` only covered headless when `rendererBacked === true` (or when focus was not required). Focus-requested creates still called `getAuthoritativeWindow()`, which throws on serve with no windows.

## Test plan
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/terminal-create-background-decision.test.ts`
- [ ] On headless remote: `orca --environment <remote> terminal create --worktree <wt> --command "echo test" --focus --json` returns ok (background surface) instead of runtime_error
- [ ] Mac UI `+` → agent on headless remote server creates a usable terminal handle

## ELI5

Launching an agent or terminal create --focus on headless orca serve crashed with No renderer window available. Headless creates now use the background spawn path when there is no UI window.

### DIFF
--- a/src/main/runtime/orca-runtime.ts
+++ b/src/main/runtime/orca-runtime.ts
@@ -848,6 +848,7 @@ import {
   getTerminalViewColorQueryReplyColors,
   registerTerminalViewAttributesApplier
 } from './terminal-view-attribute-store'
+import { shouldCreateTerminalInBackground } from './terminal-create-background-decision'
 import { killAllProcessesForWorktree, teardownRpcDeadline } from './worktree-teardown'
 import {
   MobileNotificationReplayBuffer,
@@ -21496,14 +21497,13 @@ export class OrcaRuntimeService {
     // requires a resolvable selector, so route the no-selector case through
     // the renderer IPC path to preserve that behavior.
     const rendererWindow = opts.rendererBacked === true ? availableAuthoritativeWindow : null
-    const shouldCreateInBackground =
-      worktreeSelector !== undefined &&
-      (Boolean(opts.agentSessionClaim) ||
-        (!requiresRendererFocus && opts.rendererBacked !== true) ||
-        // Why: `orca serve` exposes the local runtime without a renderer
-        // window. Renderer-backed Codex terminals are preferred for the app,
-        // but headless CLI users still need a usable terminal handle.
-        (opts.rendererBacked === true && rendererWindow === null))
+    const shouldCreateInBackground = shouldCreateTerminalInBackground({
+      worktreeSelector,
+      agentSessionClaim: Boolean(opts.agentSessionClaim),
+      requiresRendererFocus,
+      rendererBacked: opts.rendererBacked === true,
+      hasAuthoritativeWindow: availableAuthoritativeWindow !== null
+    })
 
     if (shouldCreateInBackground) {
       if (!this.ptyController?.spawn) {
--- a/src/main/runtime/terminal-create-background-decision.test.ts
+++ b/src/main/runtime/terminal-create-background-decision.test.ts
@@ -1,0 +1,89 @@
+import { describe, expect, it } from 'vitest'
+import { shouldCreateTerminalInBackground } from './terminal-create-background-decision'
+
+describe('shouldCreateTerminalInBackground', () => {
+  it('returns false when no worktree selector is provided', () => {
+    expect(
+      shouldCreateTerminalInBackground({
+        worktreeSelector: undefined,
+        agentSessionClaim: false,
+        requiresRendererFocus: true,
+        rendererBacked: false,
+        hasAuthoritativeWindow: false
+      })
+    ).toBe(false)
+  })
+
+  it('prefers background for agent session claims', () => {
+    expect(
+      shouldCreateTerminalInBackground({
+        worktreeSelector: 'wt-1',
+        agentSessionClaim: true,
+        requiresRendererFocus: true,
+        rendererBacked: true,
+        hasAuthoritativeWindow: true
+      })
+    ).toBe(true)
+  })
+
+  it('uses background for non-focus non-renderer-backed creates', () => {
+    expect(
+      shouldCreateTerminalInBackground({
+        worktreeSelector: 'wt-1',
+        agentSessionClaim: false,
+        requiresRendererFocus: false,
+        rendererBacked: false,
+        hasAuthoritativeWindow: true
+      })
+    ).toBe(true)
+  })
+
+  it('keeps focus creates on renderer when a window exists', () => {
+    expect(
+      shouldCreateTerminalInBackground({
+        worktreeSelector: 'wt-1',
+        agentSessionClaim: false,
+        requiresRendererFocus: true,
+        rendererBacked: false,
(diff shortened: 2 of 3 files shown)


Jinwoo-H:
Thanks for this, and for the diagnosis — it was correct, and it made the fix straightforward to confirm.

We reproduced #10333 against a **real headless `orca serve` host** (not a mock) and it fails exactly where you identified: a focus-requested create falls through to `getAuthoritativeWindow()` and throws. That part of your analysis held up completely.

We've opened #12791 with the fix. Two reasons we went with a separate PR rather than merging here, both about test coverage rather than your diagnosis:

- **#10370** — the production change is logically equivalent to ours. But reverting the `orca-runtime.ts` call site while keeping the extracted helper leaves **all 7 tests passing**, because they only exercise the pure function's own truth table; nothing constructs a runtime, calls `createTerminal`, or observes the throw. So the tests can't catch a regression in the thing being fixed.
- **#10557** — the one-clause change is correct. But the tests never wire the headless-serve state (no `syncWindowGraph`/`attachWindow`), so `graphStatus` stays `'unavailable'`. Replicating that test shape against unfixed main gives `Error: runtime_unavailable`, not `No renderer window available` — it goes red on main for a different reason, so it never demonstrates the reported bug.

One thing worth knowing that neither PR could have anticipated: **the Mac UI `+` half of this issue no longer reproduces on current main.** #9687 and #10193 landed after the original report and now send `presentation: 'background'` for paired clients. So the surviving defect is the CLI/RPC path only — which is what #12791 closes at the runtime layer, since it stays reachable for any caller without a `clientKind`.

Genuinely appreciate the work here; the fix in #12791 is the shape you both proposed. Closing these in favour of it, but the credit for finding it is yours.

innocarpe:
@Jinwoo-H thanks for digging into this — and for running it on a real headless `orca serve` host instead of trusting the unit suite.

You're right that this PR's tests mostly exercise the helper, not the call site that actually throws. I don't want to ship something that stays green after the fix is reverted. Closing in favor of #12791; glad the diagnosis held up on your side.

innocarpe:
Closing as superseded by #12791.

PR #10379, opened 2026-07-24
feat(explorer): opt-in confirm before drag-moving files

### LINKED ISSUE #10263: [Feature]: Opt-in confirmation before moving files and folders in the File Explorer
### Problem or use case

Dragging a row in the File Explorer moves it on disk immediately. There is no confirmation, no setting to ask for one, and nothing distinguishes "I dragged a 200-file directory into the wrong sibling folder" from "I dragged a scratch file one row down".

Reading the shipped bundle (v1.4.152), `handleMoveDrop` in `useFileExplorerDragDrop` only guards two cases before executing the move:

- the source and destination directories are the same (no-op), and
- the destination is the source itself or a descendant of it (impossible move).

Everything else calls `executeOpenEditorPathMove` right away and registers an undo entry. A toast appears only when the move *fails*.

Two things make this easy to trigger by accident:

1. **Explorer rows are dense and the sidebar is narrow.** A drop target is one row tall. Missing by one row silently relocates a directory into its neighbour, and because the tree collapses the moved node out of view there is often no visual confirmation that anything went wrong.
2. **The drag can start from an ordinary click.** Selecting a file, then moving the pointer a few pixels before releasing, is enough to begin an internal drag. Users who scroll the tree by dragging hit this regularly.

**Undo exists but is not a sufficient safety net.** `fileExplorer.undo` (`Mod+Z`) does reverse a move, but the stack is in-memory only, capped at 50 entries (`MAX_STEPS = 50`), and `clearFileExplorerUndoHistory()` wipes it whenever the visible worktree changes. In a fleet-of-agents workflow — which is the workflow Orca is built for — switching worktrees every few minutes is normal. So the window in which an accidental move is recoverable via undo is often shorter than the time it takes to notice the move happened.

There is also an inconsistency worth naming: **delete already confirms** (including the batch prompt added in #7459), while move — which is equally destructive from the user's point of view when it silently relocates a directory that a running agent, a build config, or an import path depends on — does not. Moving a directory in a repo an agent is actively working in can break that agent's next tool call, and the failure surfaces as a confusing agent error rather than as a file operation the user remembers making.

### Proposed solution

Add an **opt-in** setting — default **off**, so nothing changes for anyone who does not go looking for it.

**Setting:** `Confirm before moving files in the Explorer` (e.g. `confirmFileExplorerMove`), alongside the existing confirmation preferences. There is prior art for both the storage shape and the UI: `skipDeleteAutomationConfirm` with its `general-skip-delete-automation-confirm` settings target already implements a persisted "don't ask me again" confirmation toggle, including the deep link from the dialog back into Settings.

**Behaviour when enabled:**

- On an internal drag-move drop, show a dialog before calling `executeOpenEditorPathMove`:
  - title: `Move '<name>'?`
  - body: the source and destination paths, relative to the worktree root
  - for a directory, the number of entries being moved
  - actions: `Cancel` (default focus) / `Move`
  - a `Don't ask again` checkbox that writes the setting back to off, matching the delete-automation dialog
- Cancelling leaves the tree untouched and pushes nothing onto the undo stack.
- Confirming proceeds exactly as today, including the existing `commitFileExplorerOp` undo/redo registration.

**Scope — deliberately narrow:**

- **Internal drag-moves only** (`WORKSPACE_FILE_PATH_MIME`). Native OS file drops are a copy, not a move, and are non-destructive — they should not prompt.
- The two existing early-return guards stay in front of the dialog, so a no-op drop or a drop into a descendant still does nothing silently rather than opening a dialog and then doing nothing.
- Keyboard/context-menu operations (rename, create, delete) are untouched; delete keeps its own existing confirmation.

**Optional refinement, if a blanket prompt feels too noisy** — instead of a boolean, a three-way setting:

| Value | Behaviour |
|---|---|
| `Never` (default) | today's behaviour |
| `Directories only` | prompt when the dragged entry is a directory, move files silently |
| `Always` | prompt for every drag-move |

`Directories only` covers the expensive mistake — relocating a whole subtree — while leaving single-file drags as fast as they are now. If only one option can ship, this is the one worth shipping.

**Rely on undo instead.** This is the current answer and it is the reason the feature is worth adding rather than closing: undo is bounded by an in-memory 50-entry stack that is cleared on worktree switch. It protects against a move you notice within seconds, in the worktree you are still looking at. It does not protect against one you notice after switching workspaces, which is the case that actually costs time.

**Raise the drag-activation threshold.** Fewer accidental drags start, but it does nothing for a deliberate drag that lands one row off — the more common failure. Worth doing independently; not a substitute.

**Show a persistent undo affordance in the toast.** A `Moved 'x' to 'y' — Undo` toast after each move is lighter-weight than a dialog and would help. It still fails the "did not notice" case, since the toast dismisses itself. Reasonable to ship alongside, or as a first step if the dialog is judged too heavy.

**Related:** #7459 (batch delete confirmation) established that destructive Explorer operations get a prompt; #293 added Explorer drag-and-drop and explicitly listed undo/redo as the follow-up, which is how the current undo-only model came about. This request is that the two policies be reconcilable by preference rather than fixed.

Verified against v1.4.152 (macOS) by reading the shipped `app.asar`; function and setting names above are the bundled identifiers, not necessarily the source-tree names.

innocarpe:
## Summary
- Fixes #10263: opt-in confirmation before File Explorer **internal** drag-moves.
- New setting `confirmFileExplorerMove`: `never` (default) | `directories` | `always` under **Settings → General → Workspace**.
- When enabled, shows a cancel-first dialog with relative from/to paths before `executeOpenEditorPathMove`. Existing no-op / descendant-into-self guards still skip silently without a dialog.
- Native OS file drops (copy) are unchanged and never prompt.

## Why
Delete already confirms; accidental directory moves are hard to notice after worktree switches clear the in-memory undo stack.

## Test plan
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/right-sidebar/file-explorer-move-confirm.test.ts src/shared/constants.test.ts` (32 passed)
- [ ] Default `never`: drag-move works as today with no dialog
- [ ] `directories`: file move silent, folder move prompts; Cancel leaves tree unchanged
- [ ] `always`: every drag-move prompts; Move proceeds with undo entry as before

## ELI5

Optional setting to confirm before drag-moving files or folders in the File Explorer (never / directories / always). Stops accidental moves; OS file drops still copy without a prompt.


nwparker:
## Visual proof

Recorded at 1080p with “Always” enabled. Dragging `README.md` into `src` opens the cancel-first confirmation; cancelling leaves the file unchanged.

https://github.com/user-attachments/assets/43d279a4-9868-46a1-a85b-e94c5fed9320

innocarpe:
### Sync update (`5cf02f3de0`)

Addressed the remaining multi-selection drag metadata review scope. Directory metadata is now preserved for every selected path, directory-only confirmation works when a directory is selected alongside the primary file, and the latest upstream settings are retained.\n\nValidation: 59 focused Explorer tests passed, web typecheck passed, Oxfmt passed, and Oxlint passed under Node 22 (the repository declares Node 24).

- Commit: `5cf02f3de04d5deb9f4c77cfff6d55a6d213de15`
- Upstream: https://github.com/stablyai/orca/pull/10379
- Fork portfolio: https://github.com/innocarpe/orca/pull/52

PR #10388, opened 2026-07-24
feat(explorer): independent File Explorer font size setting

### LINKED ISSUE #10259: Feature request: separate font size setting for sidebar / file explorer
### Description

The right-side sidebar (resource manager / file explorer) currently uses the same font size as the rest of the UI, controlled only by the global `uiZoomLevel`. There is no way to adjust the file/folder name font size independently.

### Use Case

When working on large projects with deeply nested directory structures, the file names in the explorer panel are hard to read at the default size. Zooming the whole UI (`Cmd+=` / `Cmd+-`) makes everything — including terminal and tabs — larger, which is not always desired.

### Requested Solution

Add a dedicated setting to control the font size of the file explorer / sidebar, for example:

- `explorer.fontSize` or `sidebar.fontSize`
- A slider in Settings → Appearance → Window & Sidebar

This would allow users to make file names in the resource manager more readable without affecting the terminal, editor, or tab bar font sizes.

### Alternative Considered

Currently the only workaround is to:
1. Use global UI zoom — affects the entire app
2. Inject custom CSS via DevTools — not persistent

Neither is ideal.

### Environment

- Orca version: 1.4.152
- Platform: macOS

innocarpe:
## Summary
- Fixes #10259: File Explorer file/folder name size can be adjusted without global UI zoom or terminal font.
- New setting `fileExplorerFontSize` (default **12px**, range 10–20) under **Settings → Appearance → Window & Sidebar → File Explorer**.
- Virtual row height tracks the font size so larger type does not clip.

## Test plan
- [x] Unit tests for clamp/row-height + File Explorer virtual rows fixtures (72+ related passed in local runs)
- [ ] Change File Explorer font size in settings → explorer labels update; terminal/editor unchanged
- [ ] Default 12px matches previous `text-xs` appearance

## ELI5

You can set File Explorer font size on its own (10–20px, default 12) without changing global UI zoom or terminal font. Row heights grow with the font so names do not get clipped.

### DIFF
--- a/src/renderer/src/components/right-sidebar/FileExplorer.test.tsx
+++ b/src/renderer/src/components/right-sidebar/FileExplorer.test.tsx
@@ -797,6 +797,7 @@ describe('FileExplorerRow collapse folder action', () => {
   it('passes the row node to the collapse folder handler', () => {
     const onCollapseFolderSubtree = vi.fn()
     const element = FileExplorerVirtualRows({
+      fontSize: 12,
       virtualizer: {
         getTotalSize: () => 26,
         getVirtualItems: () => [{ index: 0, key: 'src', start: 0 }],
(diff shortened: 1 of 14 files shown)


nwparker:
## Visual proof

The File Explorer’s default labels are shown first, then at the independent 20px setting. The capture verified the rendered row font is 20px after the change.

Before (default):

File Explorer default size

After (20px):

File Explorer 20px size

PR #10390, opened 2026-07-24
fix(terminal): re-seed kitty keyboard flags after reattach replay

### LINKED ISSUE #10381: [Bug]: macOS — kitty keyboard flags silently lost across pane reattach/replay; Option/modifier shortcuts in kitty-protocol TUIs (Pi) permanently stop working mid-session
### Operating system

macOS

### Orca version

v1.4.147 (com.stablyai.orca)

### Details

**Short summary:**
Option/modifier shortcuts in Pi (a kitty-keyboard-protocol TUI) work right after the pane starts, then permanently stop working mid-session. Analysis of the bundled renderer code points to the pane's kitty keyboard flags being silently lost/zeroed across Orca's detach → reattach/replay cycle, with no way for the TUI to detect it or recover.

**Environment:**
- Orca 1.4.147, macOS (Apple Silicon)
- Pi coding agent v0.81.0 running in an Orca terminal pane
- Orca setting "Option as Meta" at default (`macOptionIsMeta: false`)

**What happened?**

1. Start Pi in a pane. Pi negotiates the kitty keyboard protocol once at startup (`CSI > 7 u` push + `CSI ? u` query) and gets a valid reply — Option+key and Ctrl+key chords all work.
2. Keep the session running for a while, switching between panes/windows, app going to background, machine sleeping — normal usage.
3. At some point, **all Option+key chords (and some Ctrl chords) stop reaching the TUI**. Plain typing still works. The state never recovers on its own.

When broken, pressing Option+letter inserts a composed character (e.g. Option+o → "ø") into the TUI's editor, i.e. the renderer is no longer encoding Option chords as kitty CSI-u sequences — consistent with the pane's effective kitty flags having dropped to 0 while the TUI still believes the protocol is active (it pushed once at startup, per the kitty spec, and was never told anything changed).

**How can we reproduce it?**

I don't have a deterministic trigger — it correlates with pane detach/reattach (tab switching, backgrounding, sleep/wake) over a long-running session. But the code analysis below shows at least one acknowledged lossy path.

**Analysis (from the bundled `app.asar`, v1.4.147):**

`TerminalKittyKeyboardModeTracker` mirrors the pty's kitty flag stack and is snapshot/restored across daemon reattach (`snapshotKittyKeyboardFlags`, `applyKittyKeyboardFlags`). The replay path applies pushes as idempotent sets, and its own doc comment acknowledges a lossy case:

> "Known limit: a NESTED push/push/pop inside the window collapses to flags 0 on the pop — unavoidable stackless-replay tradeoff (a redelivered push is byte-wise indistinguishable from a new one)"

That case is reachable in practice: Pi (like other TUIs) tears down and re-negotiates the protocol around suspend/resume and external-editor launches, producing pop+push pairs in the retained window. Combined with reattach replay, the tracker (and/or the recreated renderer xterm's own `kittyKeyboard.flags`) can end up at 0 while the live TUI on the pty still has the protocol pushed.

Once flags read as 0:
- `isKittyKeyboardActivePane()` returns false, so Orca stops synthesizing CSI-u for Option chords;
- with `macOptionIsMeta: false` (default), Option+letter falls back to macOS character composition ("ø", "∂", …);
- the TUI receives nothing it can interpret as Alt, and — because the kitty spec says pushed flags persist — it has no reason to re-negotiate. Deadlock until the process restarts.

**Workaround that supports the diagnosis:**

Suspending the TUI with Ctrl+Z and resuming with `fg` fixes the shortcuts instantly — resume makes Pi re-send `CSI > 7 u`, re-arming both the tracker and the renderer.

**Suggested fix directions:**

- Treat the live pty's protocol state as authoritative: since the TUI process is still attached and running, a replay/reattach should never be able to lower the effective flags below what the live session last set (e.g. re-query or re-assert flags after seeding, rather than trusting stackless replay arithmetic).
- Alternatively, after reattach, re-apply the pre-detach `currentFlags` with set semantics (`CSI = flags ; 1 u`) to the renderer terminal instead of deriving them from replayed bytes.

Happy to provide logs or test a build — this makes long-running Pi sessions in Orca lose keyboard shortcuts several times a day.

innocarpe:
## Summary
- Fixes #10381: long-running kitty-protocol TUIs (e.g. Pi) no longer permanently lose Option/modifier chords after pane reattach/replay.
- **Root cause:** `scanReplay` applies pushes as stackless sets; nested push/pop in the retained window can leave the renderer mirror at flags `0` while the live TUI still has the protocol pushed. Option then falls back to macOS composition (`ø`, …).
- **Fix:** After snapshot/replay scan on reattach, re-seed the tracker from the daemon's live `snapshotKittyKeyboardFlags` (already returned on reattach, previously only used for headless seed).

## Test plan
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/shared/terminal-kitty-keyboard-mode-tracker.test.ts` (15 passed)
- [ ] Long Pi session: switch panes / sleep-wake / reattach → Option chords still CSI-u encoded
- [ ] Ctrl+Z / `fg` still works as before (re-push path)

## ELI5

After reattaching a pane, long-running kitty-keyboard TUIs (like Pi) lost Option/modifier shortcuts and fell back to macOS special characters. Reattach now re-seeds live keyboard flags so modifiers keep working.

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-connection.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.ts
@@ -7943,6 +7943,14 @@ export function connectPanePty(
           writeReplayData('\x1b[2J\x1b[3J\x1b[H')
           // Why: re-arm the kitty keyboard mirror from the snapshot preamble so Option chords keep their encoding after a window reload.
           kittyKeyboardModes.scanReplay(connectResult.snapshot)
+          // Why: stackless replay can zero nested push/pop windows while the live
+          // TUI still has protocol flags; seed from the daemon emulator (#10381).
+          if (
+            typeof connectResult.snapshotKittyKeyboardFlags === 'number' &&
+            connectResult.snapshotKittyKeyboardFlags > 0
+          ) {
+            kittyKeyboardModes.applyAuthoritativeFlags(connectResult.snapshotKittyKeyboardFlags)
+          }
           writeReplayData(connectResult.snapshot)
           // Snapshot reattach keeps a live session, so drop only renderer-owned state instead of the broader mode reset — unless this is a cold restore, whose owner is gone.
           writeReplayData(
@@ -7993,6 +8001,12 @@ export function connectPanePty(
               }
             }
             kittyKeyboardModes.scanReplay(modelData)
+            if (
+              typeof connectResult?.snapshotKittyKeyboardFlags === 'number' &&
+              connectResult.snapshotKittyKeyboardFlags > 0
+            ) {
+              kittyKeyboardModes.applyAuthoritativeFlags(connectResult.snapshotKittyKeyboardFlags)
+            }
             // Why shared: park+reveal of an alt-screen TUI needs the same
             // ?1049l/?1049h rebuild as applyMainBufferSnapshot (main strips
             // the ?1049h marker when splitting scrollbackAnsi) — inlined here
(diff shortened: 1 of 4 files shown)


pstanton237:
One additional observed input case for the restoration logic: on baseline aa658d28e3da347447ea6188c72f3ab50b280ed0, a same-PTY cold-park/remount returned an attach screen snapshot with neither `snapshotKittyKeyboardFlags` nor `snapshotSeq`. A separate `getMainBufferSnapshot` call returned a headless snapshot with flags **7** and an output sequence.

This observation is from a local actual-Pi test, not an execution of this PR's head.

innocarpe:
Thanks for the deterministic macOS/real-Pi observations. I rebased this branch onto the current upstream main and confirmed that the one-commit implementation is fully superseded by merged PR #13940 (P1 STA 3887).\n\nCurrent main now carries the daemon snapshot's kitty flags and sequence through the reattach/Preview paths, applies the snapshot owner’s proven flags after replay (including an authoritative 0), and keeps missing snapshot metadata unknown instead of laundering it into a proven inactive state. The model-snapshot fallback also uses the same kitty-state restoration path.\n\nI am closing this stale PR and its fork mirror in favor of #13940. Issue #10381 remains the place to verify any residual failure against current main.

PR #10417, opened 2026-07-24
fix(mobile-terminal): stop reassigning an identical xterm theme

bbingz:
## Summary

A TUI that sets its own colours through OSC 4/10/11 loses them whenever mobile re-sends the terminal theme: the WebView assigns `term.options.theme` unconditionally, and that assignment rebuilds xterm's palette from the configured values. Desktop already refuses that no-op write in `terminal-appearance.ts`; mobile now matches, and the contrast-floor write is gated the same way.

The mobile theme value is fixed today, so the redundant write mostly arrives on WebView readiness and reload rather than on user action — the palette loss is real when it does.

## Tests

Written failing first, then fixed:

- `does not rewrite an identical palette` — was 2 theme writes, now 1
- `does not rewrite the contrast floor when it is unchanged`
- `rewrites an identical palette when forced by the iOS visibility repaint`
- `skips the write when the very first palette already equals the seeded default` — production seeds `terminalTheme` with `defaultTheme`, not `null`, so the gate must already hold on the first apply
- `keeps the module palette current when no terminal exists yet` — `terminal-webview-html.ts` applies a theme and only *then* runs `new Terminal({ theme: terminalTheme })`, so the module assignment must stay outside the `if (term)` gate. Mutation-checked: moving it inside fails 6 tests.

`mobile/src/terminal/` 38 files / 410 tests pass; mobile typecheck, oxlint, oxfmt and the max-lines ratchet are clean. The injected JS still parses at the ES2019 acorn gate.

### Note for reviewers

`force` looks removable. It is not: it is threaded from exactly one call site, the `visibilitychange` handler in `terminal-webview-webgl-recovery-injected.ts`, which deliberately repaints the palette after an iOS foreground transition. Removing it silently disables that repaint, and only a real background/foreground cycle on a device proves it is still needed — no unit test can. The unit test pins the parameter; this paragraph is here so it isn't "simplified" away.

## ELI5

Mobile was re-applying the same xterm theme on every message, which rebuilds the palette and can wipe live TUI colour tweaks. Desktop already skipped no-op theme writes; mobile now does the same.

### DIFF
--- a/mobile/src/terminal/TerminalWebView.tsx
+++ b/mobile/src/terminal/TerminalWebView.tsx
@@ -109,7 +109,10 @@ export const TerminalWebView = forwardRef<TerminalWebViewHandle, Props>(function
         onWebReady?.()
       }
       // Why: reload clears queued commands, so readiness must always restore the
-      // native-selected theme even when its value did not change in React.
+      // native-selected theme even when its value did not change in React. This is
+      // state restore only — the theme write is value-gated, so the iOS foreground
+      // repaint is owned by the visibilitychange handler in
+      // terminal-webview-webgl-recovery-injected.ts, not by this re-post.
       sendToWebView({ type: 'set-theme', terminalTheme })
       flushPendingMessages()
     },
--- a/mobile/src/terminal/terminal-webview-engine.test.ts
+++ b/mobile/src/terminal/terminal-webview-engine.test.ts
@@ -230,7 +230,7 @@ describe('terminal WebView bundled engine', () => {
     harness.document.visibilityState = 'visible'
     harness.fireVisibilityChange()
 
-    expect(harness.applyTerminalTheme).toHaveBeenCalledWith(harness.terminalThemeInput)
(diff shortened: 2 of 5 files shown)


bbingz:
Rebased onto current `main` (no content change). Small value-gate fix remains the only review surface.

bbingz:
Catch-up rebase onto latest `main` (no content change).

bbingz:
Catch-up rebase onto latest `main` (no content change).

bbingz:
**Queue hygiene (contributor):** converting this PR to **draft**.

It is **not** superseded by `main` — still a valid product stack — but at +hundreds–thousands of lines it sits behind a large campaign backlog and is unlikely to get review soon. Draft reduces noise for maintainers.

If product wants this stack prioritized, say so and we will mark ready again / rebase.

Jinwoo-H:
Closing this submitted branch and routing the bug to owner issue #17567. We will carry the fix ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @bbingz. No action is requested from you. Thank you.

PR #9663, opened 2026-07-21
plugins(P2): deferred theme, icon-theme and terminal-theme contributions

nwparker:
## Plugin-contributed theming (P2)

Adds declarative app themes, file icon themes, and terminal themes to the plugin system, including their registries, renderer state, Appearance controls, activation flow, consent disclosure, launch content, and packaging verification.

This is now rebased directly onto `main` (`2cf91b8e69`) and is no longer stacked on the Phase 1 branch. The PR remains **draft** pending review of the trust-surface contrast policy described below.

### Contributions

| Contribution | Launch packs |
|---|---|
| `themes` | `stablyai.orca-midnight-theme`, `stablyai.orca-nord-theme` |
| `iconThemes` | `stablyai.orca-minimal-icons` |
| `terminalThemes` | `stablyai.orca-solarized-terminal` |

The launch catalog contains seven plugins. Midnight Theme and Navigation Shortcuts are bundled for offline installation.

### Implementation

- Validates theme manifests and keeps all contributions declarative and host-rendered.
- Exposes a bounded public-token allowlist for app themes.
- Rejects app-theme URLs, `var()`, declarations, quotes, braces, and escapes.
- Realpath-confines icon files to the plugin package.
- Size-limits and sanitizes SVGs with an element allowlist; rejects events, links, styles, URLs, declarations, and active content.
- Accepts only known normalized terminal color slots.
- Applies selections through host-owned registries and stores, with reload persistence and safe built-in fallbacks.
- Preserves Navigation Shortcuts in the bundled index and the panel theme-revision regression coverage.
- Repairs the real marketplace E2E journey to use current launch content and Nord as the P2 theme case.

### Verification

Rebased result: `d5184b9c84`

- `pnpm typecheck` — passed
- `pnpm lint` — passed, including switch exhaustiveness, reliability gates, max-lines ratchet, manifests, and localization parity/coverage (11,480 keys per non-English locale)
- Changed-test matrix — 29 files, 217 tests passed
- Packaging verifier — 1 file, 4 tests passed
- Fresh E2E Electron build — main 1,921 modules; preload passed; renderer 9,138 modules
- Marketplace Electron E2E — passed against a real Git fixture and fresh profile, covering consent, install, selection, application, and reload
- Broad suite stress run — 37,897 tests passed before temporary storage exhaustion caused 213 suite-load failures

[...]

PR #10418, opened 2026-07-24
refactor(terminal-themes): move the built-in catalog to src/shared

bbingz:
## Summary

Pure relocation of the built-in terminal theme catalog from `src/renderer/src/lib/terminal-themes/` to `src/shared/terminal-themes/`, plus its `terminal-themes-data.ts` accessors folded in. No behaviour change, no value change.

**Why:** `mobile/metro.config.js` watches only `src/shared`, so the mobile bundler cannot see anything under `src/renderer`. A follow-up PR lets the phone pick a terminal theme, and it must use the *same* definitions as the desktop rather than a second copy that drifts.

## Why the move is provably value-preserving

Not "the tests still pass" — the existing contrast test is value-blind (it only asserts which channels fall below a floor, so any hex that stays above it is invisible to it). The real proof:

- `git diff -M` reports **R100** (100% similarity) for all five data modules and `shared.ts` → `catalog-merge.ts`
- their SHA-256 hashes are byte-identical to `origin/main`

The existing desktop suites passing unchanged is a secondary signal only.

## The catalog has 30 themes, not 22

Worth stating because it is easy to get wrong: eight entries use unquoted object keys (`Dracula`, `Nord`, `Monokai`, `Nightfox`, `Kanagawa`, `Palenight`, …), so a grep for quoted keys undercounts by exactly those eight. There was **no test on the catalog's contents at all** before this PR; there is now one pinning the count and asserting every theme supplies the full slot set.

## Exports

Names are unchanged. One type widened: `BUILTIN_TERMINAL_THEME_NAMES` is `readonly string[]` where the old `getThemeNames()` returned `string[]` — strictly safer, sole consumer is a `.map`, and all three tsconfig projects compile clean.

## Note on #9663

That draft also edits `src/renderer/src/lib/terminal-theme.ts`. This PR keeps its touch there minimal (import path only) so either landing order rebases cheaply.

## ELI5

Moves the built-in terminal theme list into shared code so both desktop and mobile can use the same definitions. No behavior change—just a relocation so the phone bundler can import the catalog.


bbingz:
Addressed CodeRabbit on the catalog lookup:

- `getBuiltinTerminalThemePalette` now returns only own-property entries (`Object.hasOwn`)
- `mergeTerminalThemeCatalogs` builds a null-prototype map so keys like `__proto__` stay data
- regression tests for `__proto__` / `constructor` / `toString` / `hasOwnProperty` / `valueOf`

Also rebased onto current `main`. The other CR notes about empty themed-style factory registry / unthemed-import ratchet target files that are not in this PR (they land later in the stack).

bbingz:
@coderabbitai review

bbingz:
Rebased onto latest `main` (`f00950067` +1) with no content change — keeps the stack base current for #10420.

bbingz:
Catch-up rebase onto latest `main` (no content change).

bbingz:
Catch-up rebase onto latest `main` (no content change; base for side-branch refresh).

bbingz:
Consolidated into #10421 (commit 1 of the stack there, byte-identical R100 move) so the maintainer has one user-visible feature to evaluate instead of three dependent refactor-shaped PRs. Closing.

PR #10420, opened 2026-07-24
feat(mobile): device-local two-slot terminal theme selection

bbingz:
> Stacked on #10418 (shared catalog). Its commit is included here; the reviewable content is the top three commits.

## Summary

Gives the phone its own terminal palette choice, mirroring desktop's model exactly: a **dark slot**, a **light slot**, and a **match-dark-mode** flag — the same shape as `terminalThemeDark` / `terminalThemeLight` / `terminalUseSeparateLightTheme`, resolved the same way.

**Default behaviour is unchanged.** With nothing stored, every slot is `null`, which means "follow the desktop-pushed palette" — exactly what happens today. There is no UI in this PR; the picker is #10420.

Deliberately *not* a single-slot model: the desktop does not pin the terminal dark in light mode, it keeps a separate light terminal theme (`terminalUseSeparateLightTheme` ships `true`). Mirroring that now avoids a storage migration when the app gains light mode.

## Where the override wires in

`mobile/src/session/TerminalPaneView.tsx`, not the session screen. Two reasons: the session screen sits at 4994 effective lines against a 5015 cap, and `TerminalPaneView` is the actual forwarding boundary, so the override lands where the prop already flows. The `JSON.stringify` change-detection paths in `mobile-terminal-records.ts` are untouched.

`TerminalPaneView.test.ts` exists specifically because reverting that one wiring line otherwise passes every other test in this PR.

## The storage guard is the part worth your attention

`saveMobileTerminalThemeSelection` writes conditionally, which is non-obvious from the diff. The truth table:

| `hydrated` | read outcome | what is written |
|---|---|---|
| `true` | — | all three keys, merged against the real stored value |
| `false` | hydration succeeds mid-call | all three keys, re-merged against the real value after the await |
| `false` | **read rejected** | **only the keys present in the patch** |

That last row is the point. A rejected read leaves the merge base at the in-memory default while `setItem` may still succeed, so writing all three keys would silently erase whichever slots the patch never named. Mutation-checked: defeating the guard makes the test fail.

A stale or hand-edited theme name degrades to "follow desktop" rather than a blank terminal.

Known minor quirk, not data loss: on the hydrate-succeeds-mid-call path a subscriber can observe two publishes in the same microtask chain (real disk value, then the merged value). Accepted.

## Tests

`mobile/src/` suites for the preference store, slot selection, the hook, and the pane seam. Full mobile suite green; typecheck, oxlint, oxfmt and the max-lines ratchet clean.

## ELI5

Gives the phone its own terminal theme choice like desktop: separate dark and light slots plus a match-dark-mode flag. With nothing saved, it still follows the desktop/host theme, so default behavior stays the same.


bbingz:
Rebased onto current `main` via updated #10418 tip (`700b6909e`).

- Includes catalog own-property / null-prototype fix from #10418 (addresses CR Major on catalog lookup)
- No content change in the device-local slot selection commits themselves
- Verified: catalog (5) + preference/slot/use-theme/pane/webview mobile tests (30) = 35/35

CR note about trimming `mobile-theme.ts` commentary is **out of scope for this PR** (file not in the diff). Stack note: this PR is on #10418 only, not #10419.

bbingz:
Catch-up rebase onto latest `main` / refreshed #10418 (no content change) so #10423 can stack cleanly.

bbingz:
Catch-up rebase onto latest `main` (no content change; base for #10421 refresh).

bbingz:
Consolidated into #10421 — this was the storage half of the picker with no observable effect on its own. Closing.

PR #10421, opened 2026-07-24
feat(mobile): pick your terminal theme in Settings → Terminal

bbingz:
## Summary

Add a device-local terminal theme picker under Settings → Terminal. A phone can select a built-in palette or Follow desktop, and mounted terminal panes update without navigation. The WebView frame follows the chosen background. Dark/light preferences and Match dark mode are saved independently; current mobile chrome is dark-only, so the dark slot is active and the light slot is stored for a future light-mode integration.

The built-in catalog moves into shared code for desktop/mobile reuse, preserving the 30 theme names and palettes. The picker drawer lives at screen root to avoid clipping and exposes selected/disabled states to accessibility tools. Missing or unknown choices follow the host palette.

This remains a draft and is the stack base for #10419, which supplies host-side theme publication. This PR does not change the remote theme contract or add a stream opcode.

## Testing and review

At `a1dcff2cab9a16c6511c25179d3f5a376c6df67c`, independent verification passed 46 mobile tests across 7 files using official `react-test-renderer` 19.2.8 with mocked React Native primitives, plus 35 desktop/shared tests across 3 files. Checks cover persistence, failed-read retry, choices during hydration, slot resolution, live hook updates, picker selection/accessibility, pane propagation and WebView background.

CLI, Node and Web TypeScript checks (`--composite false`), changed-file lint/format and diff whitespace passed. These desktop typechecks do not cover the mobile app. Independent Cursor review passed for this exact SHA (spec-compliance PASS; quality APPROVED).

Mobile app typecheck, native iOS/Android rendering, complete Expo bundling and live paired-host/SSH smoke were not run: the complete Expo/React Native dependency environment was not installed. No full repository suite was repeated. Manual Settings → Terminal palette selection / Follow desktop / restart validation remains unchecked; this is not a ready-for-merge claim.

## AI disclosure

Cursor ported the change. Codex independently checked the original scope and source and reran the affected tests with the real React test renderer in isolated dependencies.


bbingz:
Rebased onto current `main`. One content change beyond the rebase: `TerminalThemeSettings.tsx` reads the frozen `colors` alias like every other screen at this point in the stack, so it is now listed in the `unthemed-color-imports` ratchet — without the entry that test fails on this head. Whichever lands second, this picker or the themed-styles migration (#10490), converts the screen and deletes the entry.

bbingz:
Correction to my previous comment: the ratchet entry is no longer needed and has been dropped — this branch is back to its original single commit. The earlier rebase had linearised a merge in the stack below, which pulled the light palette (and its ratchet test) into this branch's history; redone with `--rebase-merges`, so this PR is again exactly the catalog + slot-selection commits plus the picker, and its diff is the same 30 files / 1651 insertions it was before. Sorry for the churn.

bbingz:
Rebased onto current `main` via updated #10420 tip (`112bf74eb`).

- Reviewable surface remains the top commit (5 files: terminal theme picker UI)
- No content change in the picker commit itself
- Verified: TerminalThemeSettings + PickerListDrawer + slot/preference/use-theme tests (42) + catalog (5)

CodeRabbit notes:
- `mobile-theme.ts` commentary — not in this PR
- `react-test-renderer` migration — deferred; mobile has no `@testing-library/react-native` dep and the rest of the suite still uses RTR with the same deprecation suppress pattern

bbingz:
Catch-up rebase onto latest `main` via refreshed #10420 (no content change).

bbingz:
Resolving the save-before-hydration thread with the adjudication that was previously only in the body: on the hydrate-succeeds-mid-call path a subscriber can observe two publishes in the same microtask chain (disk value, then the merged value) — an accepted, bounded degradation, not data loss. The truth table in the description covers the cases.

bbingz:
**Queue hygiene (contributor):** converting this PR to **draft**.

It is **not** superseded by `main` — still a valid product stack — but at +hundreds–thousands of lines it sits behind a large campaign backlog and is unlikely to get review soon. Draft reduces noise for maintainers.

If product wants this stack prioritized, say so and we will mark ready again / rebase.

bbingz:
**Relationship note — this is the base of a two-PR stack, not a duplicate of #10419.**

All 31 files here are contained in #10419. The six files that are only in #10419 are the headless half:

- `src/shared/terminal-theme-resolution.ts`
- `src/shared/runtime-mobile-terminal-theme.ts` (+ test)
- `src/main/runtime/orca-runtime.ts`
- `src/main/runtime/headless-mobile-terminal-theme.test.ts`
- `src/renderer/src/runtime/sync-runtime-graph.ts`

Intended order: **this PR first** (relocate the theme catalog to `src/shared` and add the mobile picker), then #10419 rebased down to just those six files (headless/SSH hosts pushing their configured theme to mobile tabs).

Reviewing them in that order avoids reading the same 31-file relocation twice. If the relocation itself is the objection, say so here and I'll close both rather than have you review either.

PR #10433, opened 2026-07-24
Bug floating workspace shortcuts route to main w

### LINKED ISSUE #10288: [Bug]: Floating workspace shortcuts route to main workspace instead of floating panel

AmethystLiang:
## Summary

- Routes close and indexed-switch shortcuts from the floating workspace, including focused browser guests, to the floating panel instead of the main workspace behind it.
- Centralizes shortcut ownership across Electron main, renderer capture, panel dispatch, and terminal-pane handling while preserving terminal-first behavior.
- Uses confirmed close paths for terminals, browsers, editors, and simulators, and reclaims empty-panel focus only after a close completes, including deferred dirty-editor closes.
- Clears stale focus-reclaim intent on repopulation, cancellation, panel close, outside interaction, and window blur.

Closes #10288

## Attribution

- Builds on #10280 by @feelgom, including the floating-panel focus signal, main-process shortcut gating, and split-terminal close delegation.
- Builds on #10359 by @innocarpe, which independently addressed routing tab shortcuts to the focused floating panel.

Thank you both for the prior investigation and implementation work.

## Screenshots

No visual change.

## Testing

- [x] Focused Vitest suites for browser guest UI, main-window routing, floating terminal panel, and IPC events: 270 tests passed
- [x] `pnpm exec oxlint src/renderer/src/components/terminal-pane/TerminalPane.tsx`
- [x] `git diff --check`
- [ ] `pnpm typecheck` — blocked by the local dependency install missing `emojibase-data/en/shortcodes/github.json`; the reported errors are in `src/shared/emoji-shortcode-catalog.ts`
- [ ] Full `pnpm test`
- [ ] `pnpm build`

## AI Review Report

- Reviewed the keyboard-routing flow from Electron `before-input-event` through renderer capture, browser guest forwarding, floating-panel dispatch, and terminal-pane handlers.
- Checked race and stale-state cases involving transient focus changes, repeated index chords, invalid indexes, deferred or cancelled closes, concurrent tab creation, and stale browser guest IDs.
- Consolidated shortcut matching so rename, indexed selection, maximize/minimize, and terminal close ownership do not drift between routing layers.
- Explicitly checked macOS, Linux, and Windows behavior through the existing platform-aware keybinding system. The change adds no platform-specific labels, paths, shell behavior, or Git commands.
- Added or expanded focused regression coverage for floating versus main guest routing, initial indexed-chord yield, auto-repeat containment, close completion, deferred close reclaim, and stale guest source IDs.

## Security Audit

- Adds no permissions, network requests, credential handling, dependency changes, command execution, or filesystem access.
- Main-process floating-focus IPC accepts updates only from the main-window renderer.
- Floating browser close requests are resolved against live floating browser workspaces and pages before dispatch; stale or unknown IDs are ignored.
- Out-of-range floating index selections do not activate a tab.
- Preload and web fallback API surfaces use the same typed routing events.

## Notes

- Uses the existing platform-aware keybinding system: Command-based shortcuts on macOS and configured Ctrl/Alt equivalents on Linux and Windows.
- Preserves terminal-first input behavior, including terminal chords intended for local or SSH-backed tools.
- The routing is host-agnostic and introduces no path or Git assumptions, so folder workspaces and git worktrees use the same behavior.
- Focused unit coverage was run locally; no real Electron/WebContents cross-platform end-to-end run was performed.


feelgom:
Hi @AmethystLiang, thanks for driving this. The extra hardening here (guest bridge, focus reclaim, auto-repeat containment) goes well beyond what my PR covered, and the single atomic `ui:setFloatingFocus` payload is a better design than the separate channel I used.

Two housekeeping asks though.

**Linking the issue.** This PR isn't linked to #10288. The description is still the unfilled template, so it won't auto-close the issue on merge. Could you add `Closes #10288` to the body? Both #10280 and #10359 were closed pointing here, so right now the chain from report to repro to fix dead-ends.

**Attribution.** #10280 (opened 03:16 UTC) and #10359 (09:14 UTC) were both closed as duplicates of this PR (18:56 UTC). The core of the change here is the same one #10280 landed on 15h earlier:

- a `floatingPanelFocused` flag alongside `floatingTerminalInputFocused` in `createMainWindow.ts`
- gating the `jumpToWorktreeIndex` / `jumpToTabIndex` dispatch inside `before-input-event`
- deferring `tab.close` to the pane handler when the floating terminal is split

Same flag name, same call site (#10280 commit `e9beccd`, 03:27 UTC).

Would you add a `Co-authored-by:` trailer to the merge commit?

```
Co-authored-by: feelgom <littlestork4@gmail.com>
```

#10359 was closed the same way at the same time, so the same question applies there. @innocarpe, entirely your call whether you want a trailer too. If trailers are awkward, a line in the PR description crediting #10280 and #10359 works just as well for me.

One more thing: this is showing merge conflicts against `main` right now. I know the touched paths well from #10280, so I can rebase or review if that helps land it.

cc @nwparker

AmethystLiang:
@feelgom Thank you so much for taking the time to write this up, for laying out the history so clearly, and for being so generous about the additional hardening. I really appreciate the detailed feedback and your offer to help.

I took your feedback and updated the PR description with `Closes #10288`, credited your work in #10280, and credited @innocarpe for #10359. Thank you both again for the investigation and groundwork that led to this fix.

PR #10435, opened 2026-07-24
fix(window): reflow via Blink device-metrics on macOS 26 occlusion-reveal

brennanb2025:
Follow-up to #10056 (STA-2383), closing its macOS 26 gap.

#10253 disables the setSize jiggle on macOS 26 because mutating the native frame from a window-state callback re-enters FrontBoardServices and deadlocks. That left forceRepaint invalidate-only on macOS 26 — invalidate repaints without reflowing, so the app-shell h-dvh stayed stale and the missing-bottom-bar bug persisted on exactly the OS it targeted.

Fix: on macOS 26+ reflow inside Blink instead of AppKit — a one-shot webContents.enableDeviceEmulation (view height -1px) then disableDeviceEmulation(). This routes through the renderer widget path (CDP Emulation.setDeviceMetricsOverride), never touches NSWindow, and cannot re-enter the scene handling #10253 guards; the jiggle stays off, so the freeze fix is preserved. Reuses the existing in-flight guard so burst reveals do not stack overrides or multiply terminal resizes.

Validation (measured on a real macOS 26.5.1 / Darwin 25.5.0 box, Electron probe): called from inside a window 'show' callback it returns in 0ms with no hang; viewport changes 568→567 with dvh tracking exactly; restore returns true height with the bar flush; native content size unchanged. pnpm typecheck clean, pnpm lint clean, vitest src/main/window + browser-url 230/230 green. Both new macOS 26 tests are revert-sensitive (fail when the reflow call is removed), and the existing 'setSize never called on Tahoe' assertion still guards #10253.

Commit 13e4395739, 2 files, +96/−5.


brennanb2025:
Closing as superseded — the macOS 26 reveal-reflow gap this targets is already fixed on `main`.

`forceRepaint` no longer bare-returns on Tahoe (`src/main/window/createMainWindow.ts:68-71`):

```js
if (isMacosTahoeOrNewer()) {
  reflowRendererViewport(window)
  return
}
```

with the in-code rationale citing STA-2383 directly: *"invalidate alone never reflows the dvh root (STA-2383); emulation reflows without touching the frame."*

That landed via #10473 (`c468e3f8b8`, "close both macOS 26 main-thread deadlock doors and restore the reveal reflow") plus #10485 (`46942783c4`, scale-factor refinement). The viewport-emulation approach is preferable to this PR's, since it reflows without mutating the window frame and therefore avoids the scene-backed main-thread deadlock rather than working around it.

This PR is `CONFLICTING/DIRTY` for exactly that reason — `main` solved the same problem a different way. Context: this was flagged as P1-3 in a v1.4.156-rc.1 release scan (#10253's Tahoe early-return made #10056's reveal reflow inert); that finding is now resolved on `main`, so no further action is needed here.

PR #8083, opened 2026-07-10
fix(editor): focus Explorer-opened Markdown for find

gatsby74:
## Summary

Opening a Markdown file from Explorer now hands keyboard focus to the rich Markdown editor, so the first Command/Ctrl+F opens rich Markdown search instead of doing nothing.

## ELI5

Before, clicking a Markdown file in the sidebar left your cursor "stuck" on the file name, so pressing Find did nothing. Now the file you open is ready to type in and search right away.

## Root cause & fix

When a Markdown file was opened from Explorer, focus stayed on the file row while the rich editor mounted. The editor's protective auto-focus guard treated that lingering Explorer focus like any unrelated control and refused to grab focus, so the first find shortcut fired outside the editor. The fix threads an explicit `focusEditor` intent from Explorer file activation through a token-based `pendingEditorFocusRequest` in the editor store, and gives `autoFocusRichEditor` a `force` param (default `false`) so only genuine Explorer opens claim focus — dialogs and other inputs stay protected.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase was clean onto `origin/main`.

- Scoped run on branch across `rich-markdown-auto-focus.test.ts`, `useFileExplorerHandlers.test.ts`, and `editor.test.ts`: **3 test files, 191 tests passed**.
- Reverting only the 4 fix files (tests kept) fails, proving the tests exercise the fix:

```
Test Files  3 failed (3)
Tests  4 failed | 187 passed (191)
 - useFileExplorerHandlers.test.ts:143 — openFile expected with focusEditor:true
 - editor.test.ts:124 — pendingEditorFocusRequest expected { fileId:'/repo/README.md', worktreeId:'wt-1' }, received undefined
```

- Lint: `oxlint` on the 4 changed source files is clean (no warnings/errors).
- A `requires-app` Electron E2E (`tests/e2e/markdown-explorer-find-focus.spec.ts`) covers the real open-then-find path; not run under vitest.

## Trade-offs

None — pure fix.

## Regression risk

Low-to-zero. `autoFocusRichEditor` gains a `force` param defaulting to `false`, so ordinary lazy mounts keep the prior neutral-focus guard byte-identical. New behavior is gated on `options.focusEditor`, set only by explicit Explorer file activation, and the `pendingEditorFocusRequest` is consumed by matching token and cleared on file close / close-all / target-file close, avoiding stale focus steals. A negative test locks in the non-Explorer focus guard.

[...]

PR #7435, opened 2026-07-05
Fix PowerShell worktree terminal cwd

mattemusacchio:
## Summary

Fixes Windows PowerShell terminals launching in the wrong directory when a user's profile changes location during load.

## ELI5

On Windows, your PowerShell startup script could quietly move the terminal into a different folder than the project you opened. Now Orca sends you back to the right project folder after that script runs.

## Root cause & fix

Orca passes the worktree path as the PTY cwd, but a user's PowerShell profile can run `Set-Location` during load and move the interactive shell elsewhere before the terminal or coding agent starts. The fix threads `cwd` into `getPowerShellEncodedCommand` and appends a `Set-Location` restore to the OSC 133 bootstrap — run *after* profile loading, with the path safely quoted via `quoteStartupArg` and wrapped in a `try/catch` that emits a `Write-Warning` on failure. Only the PowerShell/pwsh branch is touched; cmd.exe and other shells are unchanged.

## Evidence

Validated & rebased onto latest main during a bug-bash triage on 2026-07-23. Rebase was clean (base `7ab601487c`, head `8d308640ba`).

- Test file: `src/main/providers/windows-shell-args.test.ts` — 25/25 pass on branch.
- Reverting only the fix (`windows-shell-args.ts`) to `origin/main` while keeping the test causes 6 failures, e.g. `windows-shell-args.test.ts:210` asserting `decodePowerShellCommand(result)` contains the Set-Location restore.
- Lint clean: `oxlint src/main/providers/windows-shell-args.ts` reports no findings.

```
On branch:      Test Files 1 passed (1); Tests 25 passed (25)
Fix reverted:   Test Files 1 failed (1); Tests 6 failed | 19 passed (25)
  -> windows-shell-args.test.ts:210
     expect(decodePowerShellCommand(result)).toContain(...Set-Location restore)
```

## Trade-offs

None — pure fix.

## Regression risk

Zero-regression: only the PowerShell/pwsh code path changes, the restore is guarded by `try/catch`, and paths are quoted via `quoteStartupArg`. cmd.exe and non-PowerShell shells are byte-identical, and the passing/failing test pair proves the behavior is scoped to the fixed branch.

*Credit to @mattemusacchio for the original fix. Validated, rebased, and (where applicable) hardened via automated bug-bash review; original fix design preserved.*

PR #10441, opened 2026-07-24
test: realign sparse-checkout and PowerShell unit tests with shipped behavior

brennanb2025:
## Summary

No user-visible change — test-only. Repairs 10 unit tests on `main` that two PRs merged this morning invalidated. Because PR Checks merges each PR into `main`, these failures were breaking `verify` for every other open PR.

Both merged PRs are correct; only their stale test expectations are fixed here.

| Cluster | Broken by | Root cause |
|---|---|---|
| 3 × `remove-worktree.test.ts` | `e3cc08f185` (#9922) | Sparse detection now confirms `core.sparseCheckout` with a `readFile` after the `stat` fast-path. The harness mocked only `stat`, so the new read hit the **real** filesystem: sparse worktrees read back as non-sparse, and the real async I/O broke the microtask drain `bounds concurrent sparse-checkout filesystem probes` depends on — leaving a pending `listWorktrees` that leaked into the next test (which is why the third failure passes in isolation). Now mocks `readFile` alongside `stat`. |
| 7 × `pty.test.ts` | `c1d2c4be08` (#7435) | The PowerShell `-EncodedCommand` payload became **cwd-dependent** (a trailing `Set-Location` to the PTY cwd), so equality against `encodePowerShellCommand(getPowerShellOsc133Bootstrap())` can never hold again. These cases are about *which exe* gets spawned in the OSC 133 encoded form, so they now assert the arg shape plus the decoded bootstrap; `windows-shell-args.test.ts` (updated by #7435) remains the owner of the exact payload contract. |

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — clean for this diff. Still reports the **pre-existing** `RichMarkdownEditor.tsx` max-lines error that landed with #8083; that one is fixed in #10437, not here. See merge order below.
- [x] `pnpm typecheck` — passes
- [x] `pnpm test` — the 10 tests above now pass. Full-suite run also shows the 3 `orca-runtime` failures fixed by #10437, plus ~47 failures in 13 files that reproduce identically on stock `main` (`b71f71904f`) with none of these commits — pre-existing and out of scope.
- [ ] `pnpm build` — not run; this PR touches only two `.test.ts` files, no production code.
- [x] Tests are the deliverable here.

### Merge order

`main` is red in four independent clusters. This PR fixes two of them; #10437 fixes the other two (the #9343 runtime cluster and the #8083 max-lines lint). **Land #10437 first** — until it does, this PR's `verify` will fail on that inherited lint error, since lint runs before tests.

## AI Review Report

Self-reviewed the diff. Checks run:

- **Did the production fixes stay untouched?** Yes — the diff is two `.test.ts` files. #9922's and #7435's shipped behavior is unchanged, and both keep their own regression tests (`worktree-sparse-checkout.test.ts`, `windows-shell-args.test.ts`).
- **Do the repaired tests still fail for the right reasons?** The sparse tests still assert `isSparse: true` via a mocked git config, so a regression that stops honoring `core.sparseCheckout` still fails them. The pty tests still assert the resolved exe and that the OSC 133 bootstrap is delivered encoded.
- **Cross-platform (macOS / Linux / Windows):** both clusters are Windows- and WSL-behavior tests that run on every platform. The new config-path matcher normalizes `\` to `/` before matching, so it holds under both `path.join` flavors; no `process.platform` assumptions were added.
- **Test-isolation risk:** the probe-concurrency failure was leaking state into a later test. Fixed at the source (mock the I/O) rather than by loosening the follow-on assertion.

## Security Audit

No input handling, command execution, path handling, auth, secrets, dependency, or IPC surface is touched — the diff only adds mock plumbing inside two test files. The mocked git-config content is a fixed literal; no fixture reaches a real filesystem or subprocess. Nothing to follow up.

## Notes

- Fixing these in tests (not production) is deliberate: both merged PRs describe intended behavior changes and shipped their own tests for them. The failures were stale expectations elsewhere in the suite that CI never re-ran against them before merge.
- `remove-worktree.test.ts` gained a shared `mockSparseCheckoutEnabledFor` helper so future sparse cases don't re-learn that a pattern file alone no longer implies sparse.

### DIFF
--- a/src/main/git/remove-worktree.test.ts
+++ b/src/main/git/remove-worktree.test.ts
@@ -8,12 +8,14 @@ const {
   gitExecFileSyncMock,
   translateWslOutputPathsMock,
   statMock,
+  readFileMock,
   resolveGitDirMock
 } = vi.hoisted(() => ({
   gitExecFileAsyncMock: vi.fn(),
   gitExecFileSyncMock: vi.fn(),
   translateWslOutputPathsMock: vi.fn((output: string) => output),
   statMock: vi.fn(),
+  readFileMock: vi.fn(),
   resolveGitDirMock: vi.fn()
 }))
 
@@ -30,7 +32,7 @@ vi.mock('./status', () => ({
 
 vi.mock('fs/promises', async () => {
   const actual = await vi.importActual<typeof FsPromises>('fs/promises')
-  return { ...actual, stat: statMock }
+  return { ...actual, stat: statMock, readFile: readFileMock }
 })
 
 import { clearGitCapabilityStateForTests } from './git-capability-state'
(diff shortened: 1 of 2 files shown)


brennanb2025:
Superseded — `772081577e` (#10429) landed equivalent fixes for both clusters on main directly. Verified on current main: `remove-worktree.test.ts` + `pty.test.ts` = 380 passed. Nothing left for this PR to do.

PR #10442, opened 2026-07-24
feat(source-control): add setting to combine Untracked Files into Changes

FernaandoJr:
## Summary

Source Control today always splits uncommitted files into three separate sections: **Changes**, **Staged Changes**, and **Untracked Files**. This PR adds an opt-in setting, **"Combine Untracked with Changes"** (Settings → Git & Source Control), that folds untracked rows into the **Changes** section instead of giving them their own section. Each row still shows its own status indicator (M/U/D/A), so modified vs. new/untracked files stay distinguishable — only the grouping changes.

- **Off by default** — existing behavior is unchanged unless a user turns it on.
- **Staged Changes is always kept separate**, regardless of the setting.
- Bulk actions (Stage All, Discard All, "View all" combined diff) on the merged Changes section correctly cover both modified and untracked rows, at both the section and folder level.
- The existing "Source Control Group Order" setting is untouched; folded-in untracked files just leave the "Untracked Files" section empty, so it doesn't render.

## Screenshots

Setting UI (Settings → Git & Source Control), and the Source Control panel before/after toggling the setting, plus "View all" on the merged Changes section opening modified + untracked together in one combined diff. Attaching below.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Added:
- Unit tests for `mergeUntrackedIntoChanges` and the resulting section shape in `source-control-section-order.test.ts`.
- Unit test for `getDiscardAllPaths` accepting multiple areas in `discard-all-sequence.test.ts`.
- New Playwright e2e spec `tests/e2e/source-control-combine-untracked-changes.spec.ts` that seeds a modified + an untracked file in a real worktree, toggles the setting live via the store, and asserts the Untracked Files section disappears/reappears while both rows stay visible — run against the actual Electron app (`electron-headless` project), not just unit-level.
- Updated two existing settings fixtures (`runtime-home-service.test.ts`, `service.test.ts`) that construct a full `GlobalSettings` object, since the new required field needed a default there too.

## AI Review Report

Built and reviewed with an AI coding agent (Claude Code), in two passes: an initial implementation review, and a follow-up after CodeRabbit's automated review on this PR.

**Initial review — main risks checked:**
- **Merge point correctness**: traced the full flow (raw git status entries → `grouped` → `buildSourceControlDisplaySections` → tree building → row rendering) and merged at the group level, before section building, since section building already discards empty groups — kept the change smaller and isolated to display grouping.
- **Bulk action correctness**: `getDiscardAllPaths(entries, area)` and the folder-level discard-path collector filtered strictly by a single `area`, which would have silently dropped untracked rows from "Discard All" once folded into Changes. Fixed by widening `getDiscardAllPaths` to accept an area or an array of areas, deduped behind one `discardAllAreaFilter()` helper reused at all call sites instead of repeating the widening logic.
- **File-size lint gate**: the new setting's UI pushed `GitPane.tsx` over the repo's enforced 400-line `max-lines` cap. Rather than disabling or bumping the lint rule (explicitly disallowed by this repo's AGENTS.md), extracted both the new setting and the pre-existing `SourceControlGroupOrderSetting` into their own files, matching the existing `CompareAgainstUpstreamSetting.tsx` pattern.
- **Self-review pass**: caught and fixed a redundant condition in the folder-row "has untracked" check, and an unnecessary re-export kept only to avoid touching a test import.

**Addressed CodeRabbit feedback** (commit `b9d0c4109`):
- 🟠 Major: the "Discard all" / "Discard folder" title only checked the section/node's uniform `area` to decide whether to disclose that untracked files get deleted — which stayed `'unstaged'` even when the merged section held folded-in untracked rows. Now derives the title from whether the actual row set contains untracked entries, so the destructive-delete wording shows whenever it applies.
- 🔵 Nitpick: the e2e spec inferred the path separator (`\\` vs `/`) inside a `page.evaluate` renderer callback instead of using `node:path`. Moved path construction to the Node/Playwright context via `path.join`.

**Cross-platform compatibility**: UI/state logic only — no new file path handling, shell/process invocation, keyboard shortcuts, or platform-specific APIs. Exercised end-to-end on Windows via the project's own Electron/Playwright e2e harness (`electron-headless` project); nothing in the diff branches on OS. SSH/remote and folder-workspace paths are unaffected since the change only touches how already-fetched git status entries are grouped for display, not how they're fetched.

## Security Audit

No new input handling, command execution, file path construction, authentication, secrets, dependency, or IPC surface. Pure client-side/renderer display-grouping change plus one new boolean setting persisted through the existing `GlobalSettings` IPC channel (same read/write path as every other boolean setting already in that object). `getDiscardAllPaths` was widened to accept an array of the same already-typed `DiscardAllArea` union, not arbitrary input. No dependency changes.

## Notes

- Twitter/X handle: **@FernaandoJrDev**
- No platform-specific behavior; not tested against SSH/remote worktrees specifically since the change doesn't touch git status fetching, only how already-fetched entries are grouped for display (same code path regardless of local/SSH/folder-workspace).

Config:
image

Implementation preview:
image

## ELI5

Source Control always showed Untracked Files in its own section, separate from Changes. A new optional setting folds untracked files into Changes while still showing each file status (M/U/etc.), so the list matches how many people prefer Git UIs.


FernaandoJr:
Corrections made!

renansti:
What's pending on this one?

FernaandoJr:
> What's pending on this one?

I don't know waiting approval

moruno21 (approved):

PR #10444, opened 2026-07-24
Fix bulk detach for selected child worktrees

nexusjuniorai:
## Summary

Bulk **Remove from Parent** detached the wrong set of workspaces.

The multi-selection handler gated on whether **any** selected row had lineage, then dispatched `noParent` for **every** selected row:

```ts
const hasAnyContextLineage = activeContextWorktrees.some((item) => hasWorktreeParentLink(...))
// ...
activeContextWorktrees.map((item) => updateWorktreeLineage(item.id, { noParent: true }))
```

So as long as one selected row had a parent, every other selected row was detached too. Concretely, on `main`:

- Selecting a **child and an unrelated top-level workspace** detached the top-level one as well.
- Selecting a **parent together with its own child** silently unlinked the parent from *its* grandparent — collapsing a level of the tree the user never touched.
- Selecting a **folder workspace** alongside a child dispatched a worktree-lineage detach for a workspace that is not a git worktree at all.

This restricts the operation to the selected rows that actually carry a parent link, and reports partial failures by name.

## ELI5

You can group workspaces in the sidebar so one sits "inside" another, like folders. If you picked several and chose "Remove from Parent", Orca pulled *everything* you had highlighted out of its group — even items that weren't in a group, and even a group's own container. Now it only removes the ones that are actually inside something, and if some of them can't be removed it tells you exactly which ones are still attached.

## Proof of fix

The unit tests fail against `origin/main`'s `WorktreeContextMenu.tsx` and pass with this branch. Both runs use the identical test file.

**Against `origin/main` (`git checkout origin/main -- src/renderer/src/components/sidebar/WorktreeContextMenu.tsx`):**

```
 ❯ src/renderer/src/components/sidebar/WorktreeContextMenu.test.ts (35 tests | 10 failed)
     × detaches every selected child across multiple parents
     × skips selected rows that have no parent link
     × detaches a selected parent and its selected child exactly once each
     × detaches a lone selected parent from its own grandparent
     × detaches the whole tree without duplicating any row
     × excludes folder workspaces from a mixed selection
     × detaches nothing when the selection has no lineage at all
     × reports every row as detached when all succeed
     × detaches the surviving rows and names only the failed one
     × falls back to the worktree id when a failed row has no display name

 Test Files  1 failed (1)
      Tests  10 failed | 25 passed (35)
```

**With this branch:**

```
 Test Files  1 passed (1)
      Tests  35 passed (35)
```

Because those unit failures surface as missing exports, a component-level test (`WorktreeContextMenu.bulk-detach.render.test.tsx`) pins the wiring itself: it mounts the real `WorktreeContextMenu`, opens the menu via a real `contextmenu` event on a parent+child selection, clicks "Remove from Parent", and asserts the store's `updateWorktreeLineage` receives exactly one `noParent` call — for the child only. Against `origin/main`'s source it fails on the bug directly:

```
 FAIL  bulk Remove from Parent handler targeting >
       dispatches noParent only for selected rows with a parent link
AssertionError: expected "vi.fn()" to be called 1 times, but got 2 times
```

That second call is `main` detaching the lineage-free parent the user never asked to unlink. With this branch:

```
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

## Proof of no regressions

```
# whole sidebar module (includes the new render test)
Test Files  183 passed (183)
     Tests  1592 passed (1592)

# lineage store slice
Test Files  1 passed (1)
     Tests  225 passed (225)

# locale policy
Test Files  2 passed (2)
     Tests  13 passed (13)

npx tsc --noEmit -p config/tsconfig.tc.web.json   -> exit 0
npx oxlint <changed files>                        -> exit 0
npx oxfmt --check <changed files>                 -> All matched files use the correct format
node config/scripts/check-max-lines-ratchet.mjs   -> OK, no new bypasses
```

Selection shapes verified by unit test:

- Children of **one** parent → all detached.
- Children of **multiple** parents → all detached, each from its own parent.
- **A parent and its own child both selected** → each detached from *its own* parent exactly once. The child is not detached twice and the parent is not orphaned; asserted with a duplicate check on the target list.
- **A parent alone** → detached from its own grandparent.
- **Mix of parents and unrelated children** → only rows holding a parent link are targeted.
- **Whole tree** (grandparent + parent + child) → the root grandparent is correctly *not* a target; parent and child each appear once.
- **Lineage-free selection only** → no detach dispatched, menu item hidden.
- **Folder workspace** alone and mixed with a child → excluded.

Partial-failure behavior: each row is detached independently, so one failure no longer aborts the rest. The error toast names the rows that are **still attached** and counts those that succeeded, so a failed subset is retryable. Note `updateWorktreeLineage` swallows RPC-level errors internally and re-syncs lineage from the host; the rejection this surfaces is the ambiguous cross-host identity thrown by `settingsForWorktreeOwner`, which is exactly the multi-host bulk case.

Behavior that changes, beyond the bug itself:
- The bulk menu item is now hidden when no selected row has a parent (previously shown whenever any did, which was consistent — but it now also hides for selections where only non-detachable rows are present).
- A failed bulk detach now raises an error toast; previously it failed silently.

## Compatibility

- **Platforms:** macOS / Linux / Windows — no platform-dependent code. No `metaKey` use, no path handling, no shell invocation, no new keyboard shortcut. Verified on macOS; the changed code is pure TS over in-memory lineage maps with no OS-conditional branch. The pre-existing e2e spec picks its modifier via `process.platform`.
- **SSH / remote:** Preserved and improved. Detach still routes per worktree through `updateWorktreeLineage` → `settingsForWorktreeOwner`, so a selection spanning local + SSH hosts dispatches to each row's own owner. Previously one host's failure produced a single swallowed rejection; now each row settles independently and the failures are named.
- **Folder workspaces:** Excluded by construction — a folder workspace is never a lineage child (`setWorkspaceLineage` only ever writes `worktreeWorkspaceKey`), so it never holds a parent link and is filtered out rather than dispatched.
- **Performance:** No hot-path change. The filter is a `useMemo` over the current selection (typically < 10 rows), computed only while the menu is open — the lineage maps are already gated behind `menuOpen` via `selectMenuScopedMap`. No added subprocess, IPC, or render work.

## Screenshots

No visual change. The menu item's label and placement are unchanged; only which workspaces it targets, plus a new error toast on partial failure.

## Testing

- [x] `pnpm lint` — ran `npx oxlint` on the changed files (exit 0) plus `oxfmt --check`.
- [x] `pnpm typecheck` — ran `npx tsc --noEmit -p config/tsconfig.tc.web.json` (exit 0), which covers the renderer.
- [x] `pnpm test` — ran the affected scope rather than the full 30-minute suite: all 182 sidebar test files (1591 tests), the worktree store slice (225 tests), and the locale policy tests (13 tests). All pass.
- [ ] `pnpm build` — not run; no build-affecting change (no new dependency, asset, or entry point).

`tests/e2e/worktree-bulk-actions.spec.ts` was **not** executed: `pnpm run test:e2e` shells into `ensure:electron-runtime`, which runs `pnpm rebuild` / `rebuild-native-deps.mjs` and writes into `node_modules`, which is not permitted in this environment. The spec is carried unchanged from the original patch and is unverified here — flagged under Notes.

## AI Review Report

Reviewed for correctness, regressions, cross-platform behavior, SSH/remote routing, folder-workspace handling, provider neutrality, i18n, and performance.

Flagged and fixed beyond the original patch:

1. **Silent failure on a destructive action.** The handler was `void Promise.all(...)` with no rejection handling. `Promise.all` rejects on the first failure, so a multi-host selection could leave rows attached with no user feedback at all. Now each row settles independently and failures are surfaced by name.
2. **Weak test coverage.** The original test asserted only that the new helper filtered two attached rows from one unattached row — it failed on `main` merely because the export was absent, and never exercised the parent+child, whole-tree, or folder-workspace shapes. Replaced with eight selection-shape tests plus three partial-failure tests, and a component-level render test that pins the handler wiring (fails on `main` with the actual double-detach, not a missing export).
3. **Missing i18n.** The new toast strings were added to all five locale catalogs (`en`, `es`, `ja`, `ko`, `zh`), which verified as remaining exactly in parity at 11150 keys each.

Cross-platform explicitly checked: no `e.metaKey` hardcoding (none present), no path construction, no shell/quoting, no `windowsHide`, no case-sensitivity assumption, no Electron platform API. The change is platform-neutral TypeScript.

Also confirmed not GitHub-specific: worktree lineage is a local Orca metadata concept with no source-control provider coupling, so GitLab/Gitea/Bitbucket are unaffected. No new Git command is issued, so the Git 2.25 baseline is untouched.

## Security Audit

Low risk. No new IPC channel, command execution, path handling, auth, secrets, or dependency. The change narrows which existing `worktrees:updateLineage` calls are made — strictly fewer and better-targeted mutations than before — and the main-process handler's `resolveWorktreeSelector` validation is unchanged.

The one new data flow is the error toast rendering `worktree.displayName` (falling back to the worktree id). Display names are user-controlled, but they are passed as an i18next interpolation value into sonner's text description, which renders as a text node, not HTML — no injection surface. No follow-up needed.

## Notes

- `tests/e2e/worktree-bulk-actions.spec.ts` could not be run here (see Testing) and should be exercised in CI before merge. It is the only part of this PR without local verification.
- The e2e spec asserts an exact `data-bulk-detach-calls` value equal to the child id, which is the correct post-fix expectation for its parent+child selection — it would fail on `main`, where the parent is detached too.
- Follow-up worth considering separately: the drag-to-unnest path in `WorktreeList.tsx:2756` uses the same fire-and-forget `Promise.all` shape and reports a single generic toast. It targets an already-correct id set, so it is not a correctness bug, but it has the same all-or-nothing failure reporting this PR fixes for the menu path.

Original patch by @nexusjuniorai.

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeContextMenu.bulk-detach.render.test.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeContextMenu.bulk-detach.render.test.tsx
@@ -1,0 +1,185 @@
+// @vitest-environment happy-dom
+
+import { act, type ReactNode } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import type { Worktree, WorktreeLineage } from '../../../../shared/types'
+
+globalThis.IS_REACT_ACT_ENVIRONMENT = true
+
+// Why: this test pins the HANDLER wiring — bulk "Remove from Parent" must dispatch
+// noParent only for selected rows that actually carry a parent link. The pure filter
+// is covered in WorktreeContextMenu.test.ts; here the real component is mounted so a
+// regression that reconnects the handler to the unfiltered selection fails loudly.
+
+const mockState = vi.hoisted(() => ({ current: {} as Record<string, unknown> }))
+
+vi.mock('@/store', () => {
+  const useAppStore = ((selector: (state: Record<string, unknown>) => unknown) =>
+    selector(mockState.current)) as ((
+    selector: (state: Record<string, unknown>) => unknown
+  ) => unknown) & { getState: () => Record<string, unknown> }
+  useAppStore.getState = () => mockState.current
+  return { useAppStore }
+})
+
+vi.mock('@/store/selectors', () => ({
+  useAllWorktrees: () => [],
+  useRepoById: () => null,
+  useRepoMap: () => new Map(),
+  useWorktreeMap: () => new Map()
+}))
+
+vi.mock('sonner', () => ({ toast: { error: vi.fn(), success: vi.fn() } }))
+
+vi.mock('@/components/ui/dropdown-menu', () => ({
+  DropdownMenu: ({ children }: { children: ReactNode }) => <>{children}</>,
+  DropdownMenuContent: ({ children }: { children: ReactNode }) => <div>{children}</div>,
+  DropdownMenuItem: ({ children, onSelect }: { children: ReactNode; onSelect?: () => void }) => (
+    <button type="button" onClick={() => onSelect?.()}>
+      {children}
+    </button>
+  ),
+  DropdownMenuLabel: ({ children }: { children: ReactNode }) => <div>{children}</div>,
+  DropdownMenuRadioGroup: ({ children }: { children: ReactNode }) => <div>{children}</div>,
+  DropdownMenuRadioItem: ({ children }: { children: ReactNode }) => <div>{children}</div>,
+  DropdownMenuSeparator: () => <hr />,
+  DropdownMenuSub: ({ children }: { children: ReactNode }) => <>{children}</>,
+  DropdownMenuSubContent: ({ children }: { children: ReactNode }) => <div>{children}</div>,
+  DropdownMenuSubTrigger: ({ children }: { children: ReactNode }) => <div>{children}</div>,
+  DropdownMenuTrigger: ({ children }: { children: ReactNode }) => <>{children}</>
+}))
+
+vi.mock('@/components/ui/tooltip', () => ({
+  Tooltip: ({ children }: { children: ReactNode }) => <>{children}</>,
+  TooltipContent: ({ children }: { children: ReactNode }) => <>{children}</>,
+  TooltipTrigger: ({ children }: { children: ReactNode }) => <>{children}</>
+}))
+
+vi.mock('@/hooks/useVirtualizedScrollAnchor', () => ({
+  VIRTUALIZED_SCROLL_ANCHOR_RECORD_EVENT: 'orca:test-record-scroll-anchor'
+}))
+
+vi.mock('./delete-worktree-flow', () => ({
+  runWorktreeBatchDelete: vi.fn(),
+  runWorktreeDelete: vi.fn()
+}))
+
+vi.mock('./sleep-worktree-flow', () => ({ runSleepWorktrees: vi.fn() }))
+
+vi.mock('@/lib/worktree-activation', () => ({ activateAndRevealWorktree: vi.fn() }))
+
+vi.mock('./WorktreeOpenInMenu', () => ({ WorktreeOpenInSubMenu: () => null }))
+vi.mock('./ProjectGroupNameDialog', () => ({ ProjectGroupNameDialog: () => null }))
+vi.mock('./WorktreeParentPickerPopover', () => ({ WorktreeParentPickerPopover: () => null }))
+
+import WorktreeContextMenu from './WorktreeContextMenu'
+
+function makeWorktree(id: string, displayName: string): Worktree {
+  return {
+    id,
+    instanceId: `${id}-instance`,
+    repoId: 'repo',
+    path: `/tmp/bulk-detach/${displayName}`,
+    displayName,
+    branch: displayName,
+    head: 'abc123',
+    isBare: false,
+    isMainWorktree: false,
+    comment: '',
+    linkedIssue: null,
+    linkedPR: null,
+    linkedLinearIssue: null,
+    isArchived: false,
+    isUnread: false,
+    isPinned: false,
+    sortOrder: 1,
+    lastActivityAt: 1
+  } as Worktree
+}
+
+function makeLineage(childId: string, parentId: string): WorktreeLineage {
+  return {
+    worktreeId: childId,
+    worktreeInstanceId: `${childId}-instance`,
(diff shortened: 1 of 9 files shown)


nwparker:
## `d247fbf7`'s new test can't fail — it pins behavior that was never broken

The fix itself remains well-proven, so this isn't a blocker. But the render test added on top doesn't do what its commit message claims.

I reverted only `WorktreeContextMenu.tsx` to the pre-fix commit (`e28187d4c3`) and re-ran the new file:

```
WorktreeContextMenu.bulk-detach.render.test.tsx
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

It passes **with the fix removed**. The scenario is well-constructed — parent and child both selected, only the child carrying lineage — but the pre-fix handler already read:

```ts
const handleRemoveParentLink = useCallback(() => {
  void Promise.all(
    detachableContextWorktrees.map((item) => updateWorktreeLineage(item.id, { noParent: true }))
  )
```

`detachableContextWorktrees` was **already filtered**, so a lineage-free parent never received a `noParent` dispatch. The assertion `toHaveBeenCalledTimes(1)` therefore holds on both sides of the change.

**What the fix actually changes** is the failure path: `Promise.all` → per-row dispatch with partial-failure reporting. And that *is* properly covered — running the whole detach suite against the pre-fix source:

```
× reports every row as detached when all succeed
× detaches the surviving rows and names only the failed one
× falls back to the worktree id when a failed row has no display name
Tests  3 failed | 9 passed
```

Three real assertion failures. The fix is proven; the newest test just isn't what proves it.

**Suggestion:** either drop `d247fbf7`, or make it discriminating by asserting the behavior that genuinely changed — e.g. that one row's rejection still detaches the others and surfaces only the failed row's name, rather than the count of `noParent` calls. A test that passes in both directions is worse than no test: it reads as coverage while providing none, and the next person to touch this will trust it.

nwparker:
## Visual proof

Recorded against the PR branch at 1080p. The clip selects a parent and child workspace, exposes the bulk context-menu actions, and removes only the child from its parent.

https://github.com/user-attachments/assets/b4085544-41ab-4d18-934b-86dce4663a0b

Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17571. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @nexusjuniorai. No action is requested from you. Thank you.

PR #10453, opened 2026-07-24
fix(settings): guide Windows skill setup when npx is missing

### LINKED ISSUE #10438: [Bug] npx errors on Windows when installing orca skills
<kbd>[Bobby Digital](https://discord.com/channels/1492228674081263786/1530298441463238787/1530298441463238787)</kbd> [Bobby Digital](https://discord.com/channels/1492228674081263786/1530298441463238787/1530298441463238787)  `via Discord`

On a newer windows machine tryign to get a dev env setup with Orca, it doesn' thandle it well when npx isn't available and just dead ends.

Orca should handle this better.

0bkevin:
## Summary

Fixes #10438.

Windows skill setup currently opens a terminal with an `npx skills` command even when `npx` is unavailable. PowerShell then reports a generic `CommandNotFoundException`, leaving users without a recovery path.

This change:

- checks for `npx.cmd` before native Windows skill install and update commands
- prints actionable Node.js LTS, `PATH`, and restart guidance when it is missing
- executes the exact `npx.cmd` shim found by the prerequisite check
- routes the feature-tip and mobile-emulator setup terminals through the existing runtime-aware command builder
- preserves existing WSL, macOS, Linux, and unrelated-command behavior

## Screenshots

No visual change.

The command behavior was validated on a Windows machine with both missing and installed `npx`. I would appreciate additional Windows users validating the full interactive setup flow through Orca as well.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — the full run reached 35,652 passing and 62 skipped tests. Two affected settings expectations were corrected and now pass; five unrelated timing-sensitive failures passed in isolation. One untouched release-package contract test remains inconsistent with the current upstream release workflow.
- [x] `pnpm build`

Additional validation:

- Focused runtime and call-site tests: 22/22 passed.
- Changed-file Oxlint, React Doctor, Oxfmt, and `git diff --check` passed.
- Mobile TypeScript, lint, formatting, and tests passed: 2,451 passed and two skipped.
- Expo Android prebuild and Gradle `assembleRelease` passed: 987 tasks completed.
- On Windows PowerShell 5.1, the exact generated command printed the expected recovery guidance and returned exit code 1 when `npx.cmd` was absent.
- After installing Node.js LTS, the same command resolved and launched the real `C:\Program Files\nodejs\npx.cmd` shim.

## AI Review Report

The review traced every changed skill-install surface and the shared runtime command builder. It explicitly checked native Windows host execution, the missing-runtime Windows fallback, WSL distro selection, non-Windows hosts, and unrelated commands.

The review flagged that checking bare `npx` could resolve a different PowerShell shim than the command eventually executed. The implementation now checks and runs `npx.cmd` consistently behind a `cmd.exe` boundary that is valid from PowerShell, Command Prompt, and Git Bash.

It also found setup surfaces that bypassed the runtime-aware builder. The feature-tip and mobile-emulator terminals now use that builder and propagate the selected terminal shell. Regression tests guard those call sites.

Cross-platform review covered macOS, Linux, Windows, WSL, local terminals, SSH behavior, paths, shell parsing, and Electron renderer platform detection. The new branch is restricted to native Windows; WSL and non-Windows commands retain their previous forms.

## Security Audit

The wrapper applies only to known `npx skills add` and `npx skills update` commands produced by Orca. The command is not constructed from user input.

No new IPC, authentication, credentials, dependencies, filesystem writes, or network APIs are introduced. The review checked command quoting and shell metacharacter boundaries across PowerShell, Command Prompt, and Git Bash. WSL commands remain inside their selected distro and are not routed through the Windows host prerequisite.

## Notes

- No documentation update is required; this restores a recoverable setup path rather than introducing a new workflow.
- Additional interactive verification from Windows users would be welcome.
- X: [@0bkevin](https://x.com/0bkevin)

## ELI5

On Windows, skill setup ran npx even when Node was missing, so PowerShell only showed a cryptic command-not-found error. Now Orca checks for npx first and tells you how to install Node and fix PATH if it is missing.

### DIFF
--- a/src/renderer/src/components/emulator-pane/MobileEmulatorAgentSetupGuideSteps.tsx
+++ b/src/renderer/src/components/emulator-pane/MobileEmulatorAgentSetupGuideSteps.tsx
@@ -30,10 +30,11 @@ export function MobileEmulatorAgentSetupGuideSteps({
 }: MobileEmulatorAgentSetupGuideStepsProps): React.JSX.Element {
   const recordFeatureInteraction = useAppStore((s) => s.recordFeatureInteraction)
   const activeSkillRuntime = useActiveProjectSkillRuntime()
-  const skillInstallCommand = buildSkillCommandForRuntime(
-    ORCA_CLI_SKILL_INSTALL_COMMAND,
-    activeSkillRuntime.agentRuntime
-  )
+  // Why: a repair-required runtime resolves to a WSL distro that is missing, so
+  // fall back to the plain command rather than emitting an unrunnable wsl.exe one.
+  const skillInstallCommand = activeSkillRuntime.installDisabledReason
+    ? ORCA_CLI_SKILL_INSTALL_COMMAND
+    : buildSkillCommandForRuntime(ORCA_CLI_SKILL_INSTALL_COMMAND, activeSkillRuntime.agentRuntime)
   const terminalWorktreeId = `mobile-emulator-${worktreeId}-orca-cli-skill-terminal`
   const showSkillPreInstallNotice = shouldShowMobileEmulatorSkillPreInstallNotice({
     cliEnabled: setup.cliEnabled,
@@ -169,8 +170,10 @@ export function MobileEmulatorAgentSetupGuideSteps({
             terminalShellOverride={activeSkillRuntime.terminalShellOverride}
             installed={setup.cliSkillInstalled}
             loading={setup.cliSkillLoading || setup.setupRechecking}
-            error={setup.cliSkillError}
-            installDisabled={setup.step2Blocked}
+            error={activeSkillRuntime.installDisabledReason ?? setup.cliSkillError}
+            installDisabled={
+              setup.step2Blocked || Boolean(activeSkillRuntime.installDisabledReason)
+            }
             showInstallWhenInstalled={!setup.cliSkillInstalled}
             terminalHeightPx={112}
             preInstallNotice={
--- a/src/renderer/src/components/emulator-pane/MobileEmulatorAgentSetupGuideSteps.tsx
+++ b/src/renderer/src/components/emulator-pane/MobileEmulatorAgentSetupGuideSteps.tsx
@@ -2,11 +2,13 @@ import { Loader2 } from 'lucide-react'
 import { cn } from '@/lib/utils'
 import { useAppStore } from '@/store'
 import { ORCA_CLI_SKILL_INSTALL_COMMAND } from '@/lib/agent-feature-install-commands'
+import { useActiveProjectSkillRuntime } from '@/hooks/useActiveProjectSkillRuntime'
 import {
   AGENT_SKILL_CLI_PREREQUISITE_NOTICE,
   ensureOrcaCliAvailableForAgentSkillTerminal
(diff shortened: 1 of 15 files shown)


brennanb2025:
Thanks for this — the diagnosis is right and the approach holds up. Checking and then running the same `npx` behind one `cmd.exe` boundary is the correct shape, and the ordering after `normalizeWindowsSkillUpdateCommand` is right.

I reviewed it against current `main` and pushed eight follow-up commits. The substantive ones:

- **`where.exe npx` instead of `where.exe npx.cmd`**, and bare `npx` in the else branch. Pinning `.cmd` told users with an `npx.exe` shim (Volta) that "npx was not found" on a machine where npx works — worse than the original bug. Because both the probe and the execution sit inside `cmd.exe`, PATHEXT resolves them identically, so this keeps the invariant you wanted without the false negative.
- **Git Bash.** `git-bash` is a supported `terminalWindowsShell` and MSYS rewrites the leading `/d /s /c` as paths, so the wrapper broke a flow that works today. The skill terminal now forces PowerShell whenever the configured shell is POSIX-family, reusing `resolveWindowsShellStartupFamily`. That override lived in three near-identical copies; they now share one implementation, which is what let two of them drift out of sync in the first place.
- **Dropped the "restart Orca" guidance.** `mergePersistedWindowsPath` re-reads the registry PATH into every new PTY, so a new setup terminal already picks up a fresh Node install.
- **Remote runtimes.** Setup terminals spawn on the focused runtime environment, so a Windows client was handing a `cmd.exe` command to a remote Linux host. The wrapper is now skipped there.
- **Reverted the emulator surfaces to host-built commands.** Routing them through the project runtime made a WSL project install into the distro while `use-mobile-emulator-agent-setup-state` still scans the Windows host, leaving the panel stuck on "Not installed" with no way out. Their local-only detection is the constraint the existing comment already documented.
- **`EphemeralVmsPane`** required a resolved project runtime before building its command, so a Windows user with no repo added yet — the state this issue was reported from — still got the bare command.

On testing: CI had no evidence beyond `track-community-pr`, so I verified locally. Typecheck, oxlint and oxfmt are clean, and 764 renderer test files / 6577 tests pass. I mutation-checked each new assertion — every one fails when its fix is reverted. I also validated in a real Electron build on macOS that all six skill-setup surfaces still emit the plain command with no wrapper, and confirmed in the live renderer that forcing `win32` does produce it, so the path is wired rather than dead.

Two things I did not change, worth calling out:

- The onboarding and feature-wall surfaces (`onboarding-feature-setup.ts` → `FeatureSetupInlineTerminal`) still emit a raw `npx skills add` with no preflight. That is plausibly where the reporter hit this, so this PR narrows the issue rather than fully closing it. Worth a follow-up.
- On Windows the displayed and copied command is now the full wrapper. That is a real UX cost, and the copied string no longer works if pasted into Git Bash or WSL by hand. Longer term, an IPC prerequisite check (`isCommandOnPath`) would cover macOS, Linux and WSL too and keep the visible command readable — but that is a bigger change than this fix should carry.

The `npx skills add|update` gate still matches after ` -y` is appended, so this composes with the change landing in #9567.

PR #10457, opened 2026-07-24
fix(mac): disable unstable Skia Graphite renderer

OrcaWin:
## Summary

- disable Skia Graphite on macOS to prevent stale or corrupted compositor tiles after idle and app switching
- preserve hardware acceleration, GPU compositing, and terminal WebGL through the Ganesh fallback
- cover the Darwin-only switch and unchanged Linux/Windows behavior

VS Code traced the same rendering corruption to Graphite recording failures and shipped this targeted workaround: https://github.com/microsoft/vscode/issues/284162

## Testing

- `pnpm exec vitest run --config config/vitest.config.ts src/main/startup/configure-process.test.ts`
- `pnpm exec oxlint src/main/startup/configure-process.ts src/main/startup/configure-process.test.ts`
- `pnpm run typecheck:node`
- `git diff --check`

### DIFF
--- a/src/main/startup/configure-process.test.ts
+++ b/src/main/startup/configure-process.test.ts
@@ -398,6 +398,29 @@ describe('enableMainProcessGpuFeatures', () => {
     expect(app.commandLine.appendSwitch).toHaveBeenCalledWith('max-active-webgl-contexts', '128')
   })
 
+  it('disables Skia Graphite only on macOS without disabling hardware acceleration', async () => {
+    const { app } = await import('electron')
+    const { enableMainProcessGpuFeatures } = await import('./configure-process')
+
+    delete process.env.ORCA_E2E_USER_DATA_DIR
+    vi.mocked(app.disableHardwareAcceleration).mockClear()
+
+    for (const platform of ['darwin', 'linux', 'win32'] as const) {
+      setPlatform(platform)
+      vi.mocked(app.commandLine.appendSwitch).mockClear()
+
+      enableMainProcessGpuFeatures()
+
+      if (platform === 'darwin') {
+        expect(app.commandLine.appendSwitch).toHaveBeenCalledWith('disable-skia-graphite')
+      } else {
+        expect(app.commandLine.appendSwitch).not.toHaveBeenCalledWith('disable-skia-graphite')
+      }
+    }
+
+    expect(app.disableHardwareAcceleration).not.toHaveBeenCalled()
+  })
+
   it('disables the GPU sandbox on Linux Wayland without disabling acceleration', async () => {
     const { app } = await import('electron')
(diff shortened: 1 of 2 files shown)


nwparker:
Superseded by #10643 — same one-line fix, rebased onto current `main` (this branch was 50 commits behind), plus measured A/B evidence that the switch actually flips the backend (`GraphiteDawnMetal` → `GaneshGL`), and perf/security analysis showing no degradation.

Note for anyone looking at the red `verify` check here: that failure is pre-existing and unrelated to this change — it's the release-workflow contract test (`package-electron-runtime-contract.test.mjs`), which fails identically on pristine `origin/main`.

Leaving this open for @OrcaWin to close (original authorship) — happy to close it myself if preferred.

nwparker:
Superseded by #10643, which merged to `main` as 2e6467710a.

I took this over on a Mac, rebased onto current `main`, and added the empirical evidence this needed before shipping:

- **Mechanism proved, not just asserted** — probed `app.getGPUInfo('complete')`: without the switch the backend is `GraphiteDawnMetal` / `skia_graphite: enabled_on`; with it, `GaneshGL` / `disabled_off`.
- **No perf regression** — alternating A/B runs: raster within noise (9.2–9.6 ms median both ways), WebGL identical (~2000 iters/sec), frame pacing identical (60.1 fps, p99 17.6 ms), and Ganesh was ~10 MB lighter with no slower load.
- **No security regression** — this selects a backend *inside* the existing GPU sandbox; categorically unlike `--no-sandbox`/`--disable-gpu-sandbox`. All GPU features stay `enabled`, and the switch is idempotent.
- Hardware acceleration stays on — Ganesh is still fully GPU-accelerated. The macOS-only gate keeps Windows/Linux provably untouched.

Worth noting: VS Code shipped this identical switch for this identical bug (microsoft/vscode#288141, merged Jan 2026).

Thanks for the original diagnosis and patch — the darwin-only scoping came straight from here. Closing this one out in favour of the merged PR.

nwparker:
Superseded — this exact fix landed on `main` in #10643 (merged 2026-07-26).

`main` now carries the same `disable-skia-graphite` switch guarded by `process.platform === 'darwin'` inside `enableMainProcessGpuFeatures()`, plus the same `disables Skia Graphite only on macOS without disabling hardware acceleration` test from this PR. #10646 then added a follow-up test pinning the win32-only GPU-fallback invariant the macOS fix depends on.

Verified on `origin/main` @ d8af1d2196:

```
src/main/startup/configure-process.ts:269:    app.commandLine.appendSwitch('disable-skia-graphite')
src/main/startup/configure-process.test.ts: 'disables Skia Graphite only on macOS...' present
```

The diff here is a subset of what shipped, so there is nothing left to merge. Thanks for the fix and the diagnosis — closing as already-landed, not as rejected.

PR #10459, opened 2026-07-24
feat(worktrees): support project-level worktree.sharedDirectories in orca.yaml

### LINKED ISSUE #10451: [Feature]: Project-level config for symlinking large shared directories into worktrees
Follow-up to #7549 (shipped in #9791). Raised by @suiramdev in https://github.com/stablyai/orca/issues/7549#issuecomment-5064261896.

## Problem

#9791 added project-level `.worktreeinclude`, which **copies** listed gitignored paths into each new worktree. Copy is the right semantic for `.env`, `.vscode/`, and secrets: each worktree must own its file, and edits must not leak back to the primary checkout.

Copy is the *wrong* semantic for large, rebuildable, identical-across-worktrees directories — `node_modules`, `.cache`, build caches. There:

- Duplicating hundreds of MB per worktree is slow and disk-hungry. On macOS/APFS this is cheap (clone-on-write), but on **Linux and Windows it is a full byte-for-byte copy**.
- One install should ideally serve every worktree; per-worktree copies drift.

Orca already has the symlink machinery (`repo.symlinkPaths` / Worktree Shared Paths, `createWorktreeLinkedPaths`). The gap is the same one #7549 closed for copies: that list is **per-user Orca config**, so it is not version-controlled, not shared with teammates, and not portable across machines.

## Proposed solution

A project-level, version-controlled way to declare directories that should be **symlinked** (shared) rather than copied into new worktrees — the counterpart to `.worktreeinclude`.

Open design questions:

- **Where the config lives** — an `.orca/` config file, `orca.yaml`, or a distinct include file. Worth checking what neighbouring tools already read so a repo can drop one file and have it work in several.
- **Merge behaviour** with the existing per-user `symlinkPaths` (`.worktreeinclude` merges with, never replaces).
- **Collision rule** when a path appears in both the copy list and the symlink list. Today the symlink pass runs first and the copy pass skips any path that already exists, so symlink wins — that precedence should be made explicit rather than incidental.
- **Safety** — symlinking means worktree edits *do* reach the shared target. That is intended for `node_modules`, but the docs should be explicit, and the same traversal/`.git` guards as the copy path must apply.

## Notes

- Distinct from #8936, which is about exposing the **existing** per-repo shared-paths setting through the CLI. This issue is about a versioned, in-repo declaration.
- Related: `.worktreeinclude` glob/negation support and remote/SSH support are separate follow-ups from #7549.

brennanb2025:
## Summary

Adds `worktree.sharedDirectories` to `orca.yaml`, allowing a repository to declare large gitignored directories such as `node_modules` that should be symlinked into every new local worktree. Project paths add to the existing per-user Worktree Shared Paths setting.

The materializer uses an always-symlink `share` mode. This is distinct from the existing `link` mode, which APFS clone-copies on macOS and would produce independent directories instead of shared storage.

This also handles Git's directory-only ignore behavior: `node_modules/` ignores the real directory in the primary checkout but not the worktree symlink, which Git classifies as an untracked file. Orca now recognizes only paths that are both declared shared and actually symlinks when:

- checking worktree deletion in the desktop IPC and runtime/CLI paths;
- rendering local source-control status and diffs;
- deciding Create Review eligibility;
- evaluating workspace-cleanup eligibility.

The `orca.yaml` portion of the shared-path lookup is cached by repository for 30 seconds so status polling does not repeatedly read and parse YAML. Per-user paths remain live. Unsafe paths, including nested `.git` segments, are rejected.

Closes #10451.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck` (`pnpm run typecheck:tsc`)
- [x] `pnpm test` — 1,471 relevant local tests passed with one skip; current-head CI passed all 16 test shards
- [ ] `pnpm build` — not run
- [x] Added regression coverage for both deletion call sites, the live local status IPC path, workspace cleanup, Create Review, symlink discrimination, real untracked negative controls, cache behavior, and hermetic real-Git fixtures
- [x] `oxfmt --check` passes for every changed file. The repository-wide check still reports 17 pre-existing unformatted files outside this PR.

## AI Review Report

Reviewed every caller of `getStatus`, the Create Review dirty check, and both local removal preflights. The review found and fixed the original status wiring error: renderer status uses the local `git:status` IPC handler, while the first implementation only covered runtime RPC status. It also found workspace cleanup as another local status consumer that would misclassify the generated symlink as user work.

Mutation-oriented call-site tests now fail if either deletion path stops including project-level shared directories. Negative controls verify that a real untracked file still appears in status and still blocks Create Review.

Cross-platform review covered macOS APFS clone behavior, Windows/Linux symlink status behavior, path normalization, WSL/runtime option preservation, SSH exclusion, and Git command compatibility. No keyboard shortcuts, labels, shell scripts, or provider-specific review naming changed.

## Security Audit

Reviewed repository-controlled YAML parsing, path traversal, `.git` path segments, symlink discrimination, IPC path authorization, and Git subprocess arguments. Shared paths must be normalized repo-relative entries, existing directories, and gitignored before materialization. Status/removal filtering fails closed: a path is excluded only when it is configured and `lstat` confirms it is a symlink. No secrets, auth scopes, dependencies, or network behavior changed.

## Notes

- Remote/SSH worktree creation still skips shared-directory and per-user symlink materialization, so remote status paths intentionally do not apply this exclusion.
- Folder workspaces remain supported; the feature only materializes paths for Git worktrees.
- `.worktreeinclude` remains copy-only and unchanged.


AmethystLiang:

PR #10462, opened 2026-07-24
fix(terminal): open the whole URL when a TUI wraps it at its own block width

nwparker:
Fixes the reported bug: hovering a grok link that spans two lines highlights the whole URL, but clicking it navigated to only the first row's fragment.

## ELI5

Terminals have no idea what a "link" is. They're a grid of characters — the app just paints text into cells. So when you Cmd+click, Orca has to *guess* where a URL starts and ends by reading characters back off the grid.

That's easy on one line. It gets hard when a URL is too long and gets split across two lines, because now Orca has to decide: **is the text on the next line the rest of the URL, or is it a completely new line of output that just happens to sit underneath?**

Normally there's a reliable clue. When the terminal itself runs out of room, it sets a "this line is continued" flag, and Orca trusts that flag.

Grok doesn't leave that clue. It formats its own text to a narrower width (~78 columns) than the terminal is wide (~110 columns), then draws each line at an exact position. So the flag is never set, and the line visibly stops well before the right edge — which normally means "the URL ended here." Orca believed it, and opened the truncated URL.

The confusing part for you: the blue highlight came from **grok**, which knows the full URL because it drew it. Orca's click used its own guess. Two different sources, two different answers.

The fix teaches Orca to recognize this kind of wrap by looking at the shape of the surrounding text:

> "Other lines in this paragraph run further right than this one, so there *was* room to spare. This line stops at a `/`. And the first word on the next line is too long to have fit in that leftover space — so something must have pushed it down. That's a wrap, not a new line."

If any of that doesn't hold, Orca leaves the URL alone. Being wrong in the other direction is the worse bug (see below).

## The bug

| | |
|---|---|
| Hover (grok's own highlight) | `https://github.com/stablyai/orca/pull/10349#issuecomment-5068238223` |
| Click (Orca's reconstruction) | `https://github.com/stablyai/orca/` |

Orca had three URL-reconstruction strategies and the grok case fell through all of them:

| Strategy | Requires | Why it missed |
|---|---|---|
| soft-wrap | xterm's `isWrapped` flag | grok cursor-positions output; flag never set |
| edge-wrap | row reaches the **terminal's** last column | grok's rows stop at ~78 of ~110 |
| hard-wrap | a vertical frame character (`│`) | grok's markdown has no frame |

## The tension with PR #9100

**PR #9100** (commit a7c40cd89b) fixed the *opposite* bug, #8832 — the next line getting **glued onto** a modifier-clicked URL (`.../orca/` + `Description: 123` → `.../orca/Description`).

That fix joins **fewer** rows; this one joins **more**. They pull in opposite directions, and #9100's guard keyed on the *terminal edge* — which stops discriminating once block-width wrapping is reconstructed.

An early version of this change did reopen #8832. It's now re-tested at a width where the block wrap column is detectable (the gap #9100's tests didn't cover, since they used a width where no block width is inferable).

## How a row qualifies as a continuation

All must hold, or the URL is left alone:

1. the URL token runs to the end of its row's content;
2. sibling rows at the same left margin prove the block extends further right;
3. the unused gap is ≤25% of the block width — a URL ending far short of the wrap column ended because the *URL* ended;
4. the row ends at a URL break character (`/ ? & = # % + : -`), not mid-token;
5. the next row isn't `Label: value` / `Label:value`, and doesn't open with a path root (`/`, `\`, `~`, `./`, `C:\`);
6. the next row's leading token matches the URL charset and carries URL structure;
7. that token would **not** have fit in the space left on the previous row — if it fit, the author broke the line, not a wrapper.

## Evidence

**Failing-first.** Reproduced from the screenshot's exact text and geometry before any fix; confirmed the 2 failures, then confirmed green. Re-disabled the fix afterward to prove the tests genuinely gate it (5 failures returned).

**Real xterm buffer.** `grok-block-wrapped-url.repro.test.ts` drives `@xterm/headless` — not just mocks — and asserts `isWrapped === false`, confirming the metadata gap is real rather than assumed.

**Adversarial review.** Two independent reviewers (GPT-5.6-Sol, Fable) were tasked with breaking it, with regressing #8832 named as the top prize. Findings fixed, each now a permanent regression test:

| Shape | Was | Guard added |
|---|---|---|
| `Description:123` (unspaced) | glued | label pattern matches unspaced form |
| `/usr/local/bin/orca` | glued | path-root rejection |
| `DescriptionWithoutColon …` | glued | tail must carry URL structure |
| `12:34:56 request complete` | glued | same |
| `origin/main..HEAD` (self-found) | glued | same |
| short URL row + long path | glued | ≤25% unused-gap rule |
| two URLs on one row | wrong URL | scan last scheme, not first |

Also probed and **not** breakable: dates (`2026/07/24`), fractions, markdown headings, git refs, JSON, CJK/wide glyphs, Windows paths, timestamps, hover↔click divergence.

**Four reverts, kept honest.** (a) A gap rule keyed on URL break opportunities looked right but broke the real case — grok wraps whole tokens, not at URL boundaries. (b) Estimating the wrap column by most-common row width (to resist one outlier row inflating it) also broke it — real prose rows rarely share an end column. (c) Relaxing the sibling-width rule to fix limitation 1 below reopened four #8832 guards. All three reverted rather than papered over — each is recorded above so the next person doesn't retry them blind.

**Perf.** Runs on the hover path for every mouse move. Bounded by a deterministic row-read count (not wall-clock, which flaked under parallel load and stole CPU from timing-sensitive suites).

**Suite.** 16,106 renderer tests pass; typecheck, lint, format clean. `terminal-url-link-hit-testing.ts` crossed the 300-line `max-lines` cap, so the URL-token scanner was split into its own module rather than adding a disable (per AGENTS.md).

## Second review round

A follow-up adversarial pass found four more defects. All four are fixed:

| # | Defect | Fix |
|---|---|---|
| 1 | Prose rows opening with a slash-bearing word (`2026/07/24 deploy finished`, `src/main/index.ts was rebuilt`, `and/or other options`) glued onto the URL — #8832 class | a continuation must be the **whole row**, or be followed by the next link in a list. Ordinary prose after the token means the row is its own sentence. |
| 2 | Legal URLs ending in one bare final segment (a commit SHA, a plain last path segment) were truncated | the whole-row rule **subsumes** the old "tail must contain URL punctuation" requirement, so that rule is gone |
| 3 | One unrelated full-terminal-width line inflated the inferred block width, defeating reconstruction for the whole block | infer width from the **nearest** row running past the URL row, not the widest row anywhere in the window |
| 4 | Hovering a continuation row showed **no tooltip**, though a modifier-click there opened the link | register an Orca `ILinkProvider` spanning every constituent row |

Defect 4 is confirmed in `@xterm/addon-web-links` source: `_getWindowedLineStrings` walks rows via xterm's `isWrapped` flag, which a TUI positioning its own output never sets. The addon therefore reports the URL on the row bearing the scheme and `[]` on the continuation — the click fallback still worked, so the tail was clickable but silent on hover.

A third pass re-ran every finding against the fixed code: all six had already been resolved (the report predated the fix commit). It did surface one genuinely new shape — a row holding only a Markdown anchor (`#next-steps`) satisfied every continuation rule, since it fills its row, is URL-charset, and doesn't fit the leftover space. A fragment attaches to a resource, though, so a wrap never leaves `/` ending one row and `#` opening the next; that pairing is now rejected.

Defect 1's fix turned out to resolve defect 2 for free: once a tail must occupy its whole row, prose can no longer masquerade as one, so the punctuation requirement that was truncating valid URLs became unnecessary. **That also retired two limitations this PR previously documented as unavoidable** — the bare-final-segment truncation, and (via defect 3's fix) the width-contamination case.

## Known limitation

One remains. A repo-relative path alone on the line below a near-full URL row (`https://example.com/aaa…/` then `src/components/index.ts`) is **geometrically and textually identical** to the real grok wrap: same trailing `/`, same URL charset with slashes, same too-long-to-fit token, and both occupy their whole row. Both reviewers and I independently concluded no discriminator separates them from row text and geometry alone. Pinned as an explicit test so the trade-off stays visible.

A second case — where the URL row is itself the widest line in its block, so nothing witnesses spare room — is also pinned. Relaxing the sibling-width rule to cover it reopened four #8832 guards, so it stays.

Both degrade to the pre-fix behaviour (truncated URL) rather than gluing text on, which is the worse failure.

## Test plan

- [x] `block-wrapped-terminal-http-links.test.ts` — 36 tests
- [x] `block-wrapped-http-link-provider.test.ts` — hover tooltip on every wrapped row (grok case, #8832 shapes at block width, adversarial regressions)
- [x] `grok-block-wrapped-url.repro.test.ts` — real `@xterm/headless` buffer
- [x] `block-wrapped-http-link-hover-cost.test.ts` — bounded hover work
- [x] `repro-8832-url-next-line.test.ts`, `terminal-url-link-click.test.ts`, `edge-wrapped-terminal-http-links.test.ts` — PR #9100's suites still green
- [x] `pane-dom-creation.test.ts` — hover tooltip matches click target
- [x] Full renderer suite: 16,121 passed

### DIFF
--- a/src/renderer/src/components/terminal-pane/block-wrapped-http-link-hover-cost.test.ts
+++ b/src/renderer/src/components/terminal-pane/block-wrapped-http-link-hover-cost.test.ts
@@ -1,0 +1,58 @@
+import type { IBufferLine } from '@xterm/xterm'
+import { describe, expect, it } from 'vitest'
+import { buildBlockWrappedHttpLogicalLineCandidates } from './block-wrapped-terminal-http-links'
+
+// Why: this runs on the renderer's hover path for every mouse move over a
+// terminal, so its work must stay bounded no matter how adversarial the
+// buffer. Counting row reads keeps the bound deterministic — a wall-clock
+// budget would flake under parallel test load and would itself steal CPU from
+// the timing-sensitive suites running alongside it.
+const MAX_ROW_READS_PER_HOVER = 4_000
+
+function makeBufferLine(content: string, cols: number): IBufferLine {
+  const text = content.padEnd(cols)
+  const columns = Array.from({ length: text.length + 1 }, (_value, index) => index)
+  return {
+    isWrapped: false,
+    length: cols,
+    translateToString: (
+      _trimRight?: boolean,
+      startColumn = 0,
+      endColumn = text.length,
+      outColumns?: number[]
+    ) => {
+      outColumns?.splice(0, outColumns.length, ...columns.slice(startColumn, endColumn + 1))
+      return text.slice(startColumn, endColumn)
+    }
+  } as IBufferLine
+}
+
+function countRowReadsForHover(rows: IBufferLine[], bufferLineNumber: number): number {
+  let reads = 0
+  const buffer = {
+    getLine: (y: number) => {
+      reads++
+      return rows[y]
+    }
+  }
+  buildBlockWrappedHttpLogicalLineCandidates(buffer, bufferLineNumber)
+  return reads
+}
+
+describe('block-wrapped HTTP link hover cost', () => {
+  it('bounds row reads on a screen made entirely of wrapped URL rows', () => {
+    const cols = 200
+    // Maximises start-row candidates, wrap-column scans, and join attempts:
+    // every row is a full-width URL ending at a break opportunity.
+    const url = `https://example.com/${'a'.repeat(cols - 21)}/`
+    const rows = Array.from({ length: 200 }, () => makeBufferLine(url, cols))
+
+    expect(countRowReadsForHover(rows, 150)).toBeLessThan(MAX_ROW_READS_PER_HOVER)
+  })
+
+  it('bounds row reads on a huge no-newline blob', () => {
+    const rows = [makeBufferLine(`https://example.com/${'a'.repeat(20_000)}`, 400)]
+
+    expect(countRowReadsForHover(rows, 1)).toBeLessThan(MAX_ROW_READS_PER_HOVER)
+  })
+})
--- a/src/renderer/src/components/terminal-pane/block-wrapped-http-link-provider.test.ts
+++ b/src/renderer/src/components/terminal-pane/block-wrapped-http-link-provider.test.ts
@@ -1,0 +1,103 @@
+import type { IBufferLine, ILink, Terminal } from '@xterm/xterm'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { registerHttpLinkStoreAccessor } from '@/lib/http-link-routing'
+import { createBlockWrappedHttpLinkProvider } from './block-wrapped-http-link-provider'
+
+const COLS = 110
+const WRAP_WIDTH = 100
+const openUrlMock = vi.fn()
+
+// The reported grok output: a PR link wrapping after `orca/`.
+const ROWS = [
+  'pull/10257#issuecomment-5068240478) · 10349 (https://github.com/stablyai/orca/',
+  'pull/10349#issuecomment-5068238223)',
+  "These remaining ones match the thread's UI surface and are labeled as baselines now.".padEnd(
+    WRAP_WIDTH
+  )
+]
+const EXPECTED_URL = 'https://github.com/stablyai/orca/pull/10349#issuecomment-5068238223'
+
+function makeBufferLine(content: string): IBufferLine {
+  const text = content.padEnd(COLS)
+  const columns = Array.from({ length: text.length + 1 }, (_value, index) => index)
+  return {
+    isWrapped: false,
+    length: COLS,
+    translateToString: (
+      _trimRight?: boolean,
+      startColumn = 0,
+      endColumn = text.length,
+      outColumns?: number[]
+    ) => {
+      outColumns?.splice(0, outColumns.length, ...columns.slice(startColumn, endColumn + 1))
+      return text.slice(startColumn, endColumn)
+    }
+  } as IBufferLine
+}
+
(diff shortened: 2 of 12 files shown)


nwparker:
## Update: a better fix exists — see #10469

While reviewing why this needs a heuristic at all, I read grok's source. It turns out grok **already computes** every link's row, column span, URL, and a shared `id` grouping the pieces of one link across wrapped rows — then discards them, because Orca isn't in its `TERM_PROGRAM` brand table and unknown brands are treated as hyperlink-incapable.

#10469 fixes that at the source: advertise an xterm.js-compatible brand to grok, and it emits real OSC 8 hyperlinks. Verified against the real binary (`Orca → Unknown`, `vscode → VS Code`) and end-to-end in xterm.js, where both rows of a wrapped link resolve to one `urlId` despite `isWrapped === false`. **No inference at all.**

### Does this PR still have value?

I think yes, but the case is narrower than when I opened it, and it's your call:

**For merging both:** #10469 only helps TUIs that emit OSC 8. This PR covers any TUI that wraps at its own block width and doesn't — plus users on older grok builds who won't get the brand fix. The two compose safely: when OSC 8 is present, xterm's native path wins and this heuristic never engages.

**For closing this:** it's ~600 lines of inference with two documented ambiguous cases, carrying permanent tension against #8832/#9100. If grok is the only TUI this affects in practice, #10469 makes that complexity unnecessary.

I'd lean toward merging #10469 first, then deciding on this one once we've confirmed the brand fix resolves the reported case in real use. Happy to close this or trim it to a narrower core if you'd rather not carry the heuristic.

Issue #7848, opened 2026-07-08
[Bug]: CLI runtime flaps ready → stale_bootstrap while the Orca app process is alive (runtime_unavailable on terminal commands)


nwparker:
### Operating system

macOS

### Orca version

1.4.121 (reported), unconfirmed at HEAD

### Details

Split out of #7239 (https://github.com/stablyai/orca/issues/7239#issuecomment-4916799729, report by @Shoma-DS) — the cwd part of that issue is fixed separately; this tracks the runtime-connection flapping observed alongside it:

1. `orca status --json` reports `app.running: false`, `runtime.state: stale_bootstrap`, `reachable: false` **while the Orca app process is demonstrably alive** (`ps -p <pid>` shows `/Applications/Orca.app/Contents/MacOS/Orca` running).
2. `orca open --json` then briefly returns `app.running: true`, `runtime.state: ready`, `reachable: true`.
3. Immediately afterwards, `orca terminal show/read/list` fail again with `runtime_unavailable` ("Could not connect to the running Orca app"), and a follow-up `orca status --json` is back to `stale_bootstrap`.

So the CLI's view of the runtime flaps ready → stale_bootstrap within seconds while the app stays up. Possibly a stale bootstrap file left behind by an earlier instance, a runtime socket that dies right after `orca open` re-bootstraps it, or a liveness check keyed to a dead pid/socket from a previous run.

### How can we reproduce it?

No from-scratch recipe; see the sequence above from the original report (macOS 26.5.1, workspace path with non-ASCII characters, Orca 1.4.120/1.4.121-era). Related but distinct: #7410 observed `runtime_unavailable` under worktree-create load.

### Expected behavior

- `orca status` should not report `stale_bootstrap`/`app.running: false` while the app process is alive and healthy.
- If a bootstrap record is stale, `orca open` should repair it durably, not just for one call.

lerothias:
An AI helped me write this comment; I'm experiencing this bug, but I'm not sure exactly how to describe it. I hope this comment helps you 🙏 

Cross-platform confirmation on Linux AppImage + 1.4.137 — `orca-runtime.json` points at a dead pid while the live Orca is right there, producing the same `stale_bootstrap` ↔ `runtime_unavailable` flap described in #7848.

## Environment
- Orca **1.4.137** AppImage
- Ubuntu (kernel 6.x), X11
- `userData`: `~/.config/orca` (XDG default)
- CLI: `orca-ide` (Linux wrapper)

## Symptoms
[...]

PR #10463, opened 2026-07-24
fix(startup): stop a lock-losing launch from clobbering runtime discovery files

brennanb2025:
Addresses #7848 (STA-1513) — the **Linux** manifestation. Deliberately not `Fixes`: the issue was filed against macOS, and macOS is provably immune to this mechanism (evidence below), so the macOS half needs separate investigation and the issue should stay open.

## Problem

A second launch that loses the single-instance lock logs the failure and calls `app.quit()`. **On Linux that quit is deferred until after the ready event**, so the process still runs `app.whenReady()`, completes the full startup path, and publishes both canonical discovery files with its own pid and socket:

- `orca-runtime.json` via `runtimeRpc.start()`
- the agent-hooks endpoint file via `startTerminalRuntimeStartupServices()`

It then exits, leaving the records pointing at a dead pid while the real app is still running. `orca status` reports `stale_bootstrap` / `app.running: false` and terminal commands fail with `runtime_unavailable`, all while `ps -p <pid>` shows Orca alive. `orca open` relaunches, briefly repairs, and the next transient clobbers it again — the ready → `stale_bootstrap` flap in the report.

macOS terminates the process before ready, so the clobber is unreachable there — confirmed both for the raw binary and for a real `.app` launched through LaunchServices with a forced new instance (`open -n`). That platform split is why this only ever surfaced from Linux, and why the macOS report in #7848 must have a different cause.

## Reproduction

A minimal two-process Electron probe (single-instance lock + an unconditional `whenReady` that writes a record) reproduces it exactly, and an A/B against the same build confirms the fix:

```
GUARD=0 (no fix)
  [LOSER] requestSingleInstanceLock -> false
  [LOSER] lock LOST; calling app.quit()
  [LOSER] whenReady FIRED (hasLock=false)      <-- runs anyway on Linux
  [LOSER] PUBLISHED metadata as LOSER pid=119
  final record: LOSER pid=119                  *** CLOBBERED *** (owner alive at pid 25)

GUARD=1 (this fix)
  [LOSER] lock LOST; calling app.quit()
  [LOSER] GUARD: bailing out of ready handler (no lock)
  final record: OWNER pid=183                  owner INTACT
```

The same probe on macOS never reaches `whenReady` for the loser, confirming the platform split.

This matches the artifact in the report — `orca-runtime.json` naming pid 1017084 (nonexistent) with a socket absent from disk, while the live Orca ran as pid 168807 bound elsewhere — and confirms the reporter's inferred mechanism was correct.

## Root cause

The module-level guard already skips userData side effects when the lock is lost, and `single-instance-lock.ts`'s docstring explicitly describes preventing this exact clobber. But that guard ends at line 671 while the ready handler runs unconditionally — `hasSingleInstanceLock` was never consulted again.

## Fix

Bail out of the ready handler when the lock was lost, before any side effect.

`hasSingleInstanceLock` is false only when the lock was genuinely lost and `app.quit()` has already been called, so no legitimate startup is skipped — the dev skip and the macOS bypass both resolve to `true`.

I considered a write-side ownership guard mirroring `clearRuntimeMetadataIfOwned` and rejected it: a legitimate restart *must* overwrite, so the metadata layer cannot distinguish a real new instance from a transient loser. The single-instance lock is the only thing that knows.

## Test plan

- **Unit**: regression test in the existing `desktop-startup-ordering.test.ts` source-invariant pattern, asserting the bail precedes every side effect *and* that its body actually returns. Verified it fails without the guard.
- **E2E**: `headless-serve-desktop-activation.spec.ts` now asserts the on-disk runtime record still names the owner after the lock-losing launch exits. It previously asserted `runtimeId` via `status.get`, which answers from the owner's live server and therefore could not observe a clobbered file — it would have stayed green through this bug. That suite runs on `ubuntu-latest`, where the clobber is reachable.
- `vitest src/main/startup/ src/main/runtime/` → 2790 passed, 0 failed. typecheck + oxlint clean.

## Not addressed

An already-clobbered record still needs an Orca restart to be rewritten. With the clobber prevented, a stale record now only follows a genuine crash, where `stale_bootstrap` is the correct report. Durable self-repair would be defense-in-depth against mechanisms we have not seen and belongs in its own change.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1777,8 +1777,9 @@ function shouldSuppressCodexAutoApprovalSyntheticTitleFromHook(args: {
 }
 
 app.whenReady().then(async () => {
-  // Why: app.quit() above is async, so a lock-losing launch still reaches ready; returning keeps it from
-  // republishing orca-runtime.json / agent-hooks endpoint files over the live instance's (STA-1513).
+  // Why: Linux defers the app.quit() above until after ready, so a lock-losing launch still runs this
+  // handler and republishes orca-runtime.json / agent-hooks endpoints over the live instance's, leaving
+  // the CLI pointed at a dead pid. macOS quits before ready, so this is a no-op there. (STA-1513)
   if (!hasSingleInstanceLock) {
     return
   }
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1777,6 +1777,11 @@ function shouldSuppressCodexAutoApprovalSyntheticTitleFromHook(args: {
 }
 
 app.whenReady().then(async () => {
+  // Why: app.quit() above is async, so a lock-losing launch still reaches ready; returning keeps it from
+  // republishing orca-runtime.json / agent-hooks endpoint files over the live instance's (STA-1513).
+  if (!hasSingleInstanceLock) {
+    return
+  }
   logStartupMilestone('app-ready')
   // Why: install certificate decisions before any webview or headless window issues its first TLS request.
   app.on(
--- a/src/main/startup/desktop-startup-ordering.test.ts
+++ b/src/main/startup/desktop-startup-ordering.test.ts
@@ -87,11 +87,15 @@ describe('startup ordering', () => {
     const source = readFileSync(join(process.cwd(), 'src/main/index.ts'), 'utf8')
     const readyStart = source.indexOf('app.whenReady().then(async () => {')
     const guard = source.indexOf('if (!hasSingleInstanceLock) {', readyStart)
+    const guardEnd = source.indexOf('}', guard)
     const rpcStart = source.indexOf('runtimeRpc.start()', readyStart)
     const startupServices = source.indexOf('startTerminalRuntimeStartupServices()', readyStart)
 
     expect(readyStart).toBeGreaterThanOrEqual(0)
     expect(guard).toBeGreaterThan(readyStart)
+    // Why: ordering alone would still pass if the body were emptied to a no-op.
+    expect(guardEnd).toBeGreaterThan(guard)
+    expect(source.slice(guard, guardEnd)).toContain('return')
     expect(rpcStart).toBeGreaterThan(guard)
     expect(startupServices).toBeGreaterThan(guard)
     // Why: the bail must precede every side effect in the handler, not merely the RPC publish.
--- a/src/main/startup/desktop-startup-ordering.test.ts
+++ b/src/main/startup/desktop-startup-ordering.test.ts
@@ -81,6 +81,23 @@ describe('startup ordering', () => {
     expect(supervisorReady).toBeGreaterThan(readyStart)
   })
 
+  it('abandons ready-time startup when the single-instance lock was lost', () => {
+    // Why: app.quit() is async, so a losing launch still reaches ready; without this guard it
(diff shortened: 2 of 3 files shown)


brennanb2025:
Addressing the review feedback, and correcting an earlier claim of mine on this PR.

**Reproduced.** The mechanism is no longer inferred. Electron's `app.quit()` before ready terminates the process on macOS but is **deferred on Linux**, so a lock-losing launch reaches `whenReady` there and publishes over the owner. A/B on one Linux build: without the guard the record ends as `LOSER pid=119` while the owner lives at pid 25; with it, `OWNER pid=183` intact. Full evidence in the PR body.

**On the test-strength comment** — fixed. The assertion now checks the guard body actually returns, not just that it is positioned first, so emptying it to a no-op fails the test.

**Correction on the E2E.** I previously concluded the on-disk assertion was inert because it passed with the guard removed. That run was on macOS, which is the one platform where this bug cannot occur. The e2e suite runs on `ubuntu-latest`, where the clobber is reachable, so the assertion does guard the regression. Keeping it, with the platform caveat in a comment.

Also fixed the issue link: the previous `Fixes STA-1513 (#7848)` created no closing reference, because the keyword has to be adjacent to the issue. It now resolves.

brennanb2025:
Closing. The mechanism this guards is real in principle, but I could not reproduce it in the shipped app, and the PR body claimed otherwise.

## What changed

The Linux clobber was reproduced in a **minimal Electron probe** — `app.quit()` before ready is deferred on Linux (but terminates on macOS), so a lock-losing launch reaches `whenReady` and publishes. That part holds: probe A/B showed the clobber without the guard and an intact record with it.

Testing against the **real shipped app** (v1.4.159 arm64 AppImage, two Orca processes, one profile, in Docker) tells a different story:

```
owner publish latency : ~2043 ms
loser lifetime        : ~2013 ms
margin                : ~30 ms
```

The losing instance is killed right at `ready`; the publish lands just after it. It loses by ~30 ms — and under clean conditions it loses **every time**: 0/5 trials clobbered, with each trial in a fresh container.

I initially measured 3/5 clobbers and reported this as a live race. That was wrong: those runs reused one container with a `pkill` between iterations, and leftover Electron processes contaminated the later ones. The tell was that patched and unpatched builds produced identical results at identical run indices — state carryover, not a race. Re-running with per-trial isolation gave 0/5 on the unpatched app, and patching this exact guard into the shipped asar changed nothing, because the path was never reached.

## Where the real fix went

The reporter's artifact is explained by a different mechanism: when Chromium's `SingletonSocket`/`SingletonCookie` go missing, a second launch **wins** the lock, completes full startup, and clobbers the record — no race to lose. That reproduces 100% of the time on the same real Linux binary, and on macOS. Fixed in #10840, which makes the runtime owner reclaim its own discovery record whenever it stops describing a live runtime — repairing the damage regardless of cause.

## Why close rather than land as hardening

The guard is five lines, safe by construction, and would prevent the publish if that path were ever reached — a faster machine or a slower quit could flip a 30 ms margin. But shipping it means carrying a change whose stated justification I disproved, and #10840 already repairs this outcome from any cause. If the lock-loss path is ever observed in the wild, this branch is a two-line revive.

Keeping #7848 open until #10840's effect is confirmed there.

PR #10464, opened 2026-07-24
fix(cli): explain SIGABRT serve exits instead of naming the signal

### LINKED ISSUE #10461: orca serve reports a bare "exited via SIGABRT" with no cause or crash-report pointer
### Problem

When `orca serve` dies from a signal, the CLI reports only:

```text
Orca serve exited via SIGABRT
```

That message names the signal and nothing else — no likely cause, no next step, and no pointer to where a crash report lives. It's thrown from `src/cli/runtime/serve-update-supervisor.ts:83` as `RuntimeClientError('runtime_serve_failed', ...)`.

This is a real diagnostic dead end, not a cosmetic one. In #9282 it was the *only* signal a user had, and it led to a multi-day investigation down the wrong path (suspected code-signature / ASAR corruption, including a PR that was written and then closed) for what turned out to be an environmental failure. The information needed to diagnose it correctly existed on the machine the entire time:

- macOS had written `~/Library/Logs/DiagnosticReports/Orca-*.ips` naming the crashing frame.
- Electron had printed `task_name_for_pid: (os/kern) failure (5)` to stderr.

Neither surfaced through our CLI.

### Root cause of the opacity

On macOS, `orca serve` can abort *before any of our JavaScript runs* — inside `+[NSApplication sharedApplication]` during `ElectronMain`, when the process cannot register with the window server (`_RegisterApplication` → `abort`). There is no JS on the stack, so this can't be caught in-process. The diagnosis has to happen in the parent CLI process that spawned the child.

The parent does already know the exit signal.

**Constraint:** at `src/cli/runtime/launch.ts:131-136` the child is spawned with `stdio: 'inherit'` (or `['inherit','inherit','inherit','ipc']`), so the child's stderr goes straight to the terminal and the parent never observes it. Keying off the `task_name_for_pid` string would require rewiring stdio.

### Proposed fix

**Tier 1 — signal- and platform-aware diagnostic (preferred scope).** Replace the bare message at the signal-exit path with something actionable. On darwin + `SIGABRT`, roughly:

> Orca serve exited via SIGABRT during macOS application startup. This usually means the process could not register with the macOS window server — common inside restricted/sandboxed agent environments, SSH sessions without a GUI login, or CI. Try running `orca serve` from a normal terminal session. If it crashed, a report will be at `~/Library/Logs/DiagnosticReports/Orca-*.ips`.

Single call site, no behavior change, no new dependencies. The crash-report pointer alone would have resolved #9282 immediately.

**Tier 2 — optional, only if this recurs.** Switch stderr to `pipe` and forward it through unchanged (identical UX) while retaining a small ring buffer of recent lines. On abnormal exit, scan for known signatures (`task_name_for_pid`, `_RegisterApplication`) and upgrade the message from "this usually means" to a definite statement quoting the observed line. More code, and it touches the pass-through path, so it should earn its place.

### Why this generalizes

The failure isn't specific to one user's sandbox. Any environment where Electron can't initialize a macOS application context produces the same useless line: headless CI, SSH without a GUI login, locked-down containers, and agent/sandboxed execution environments — the last of which are becoming more common for exactly this kind of tooling.

### Out of scope (worth a separate issue)

`crashReporter.start()` is never called anywhere in `src/`, so Orca collects no crash data of its own and relies entirely on users locating OS-level crash reports by hand. That's the systemic version of this problem and is a larger project (consent, privacy, an upload endpoint). Noting it here only so it isn't lost.

### References

- #9282 — the report this came out of; concluded environment-induced, with the `.ips` stack showing `_RegisterApplication` → `NSApplication init`.

brennanb2025:
Fixes #10461.

## Problem

When `orca serve` died from a signal, the CLI reported only:

```text
Orca serve exited via SIGABRT
```

No cause, no next step, no pointer to where a crash report lives.

This is not cosmetic. In #9282 it was the only signal a user had, and it sent a P0 investigation down the wrong path — a suspected code-signature / ASAR corruption defect, including a PR that was written and then closed — for what turned out to be an environmental failure. The information needed to diagnose it correctly was on the machine the whole time: macOS had written `~/Library/Logs/DiagnosticReports/Orca-*.ips` naming the crashing frame, and Electron had printed `task_name_for_pid: (os/kern) failure (5)` to stderr. Neither surfaced through our CLI.

The actual crash was an abort during macOS application registration, before any of our JS runs:

```text
abort → ___RegisterApplication_block_invoke → _RegisterApplication → GetCurrentProcess
→ _NSInitializeAppContext → -[NSApplication init] → +[NSApplication sharedApplication] → ElectronMain
```

Because there's no JS on the stack, this can't be caught in-process — the parent CLI is the only place it can be explained.

## Fix

Extracts a pure `serveSignalExitError(signal, platform)` builder and swaps the bare throw in `superviseForegroundServe` for it:

- **darwin + SIGABRT** — offers the window-server registration failure as the likely cause along with the environments it shows up in (restricted/sandboxed, SSH without a GUI login, CI), and routes a remedy plus the `~/Library/Logs/DiagnosticReports/Orca-*.ips` pointer through `RuntimeClientError`'s existing `data.nextSteps` channel, which `format.ts` already renders in both human and `--json` output. The parent only ever observes the signal, never the phase, so the cause is offered as likely rather than asserted.
- **Other platforms/signals** — a plain `exited via <SIGNAL>.` with no invented cause.
- **Neither code nor signal** — its own clear line instead of rendering `exited via null`.

Verbatim human output on darwin + SIGABRT:

```text
Orca serve aborted with SIGABRT on macOS. This most often happens at application startup, when the process cannot register with the macOS window server, which is common in restricted or sandboxed environments, SSH sessions without a GUI login, and CI.
Next step: Re-run `orca serve` outside a sandboxed or restricted environment, with a macOS desktop login active.
Next step: Look for a crash report at ~/Library/Logs/DiagnosticReports/Orca-*.ips.
```

The remedy deliberately does not tell users to stop serving over SSH — SSH serve is supported; what it needs is an active macOS desktop login for the window-server connection.

## Scope

Tier 1 from the issue only. Deliberately excluded:

- **Tier 2** stderr ring-buffer parsing — `launch.ts` spawns with `stdio: 'inherit'`, so the parent can't observe child stderr without rewiring the pass-through path. Deferred until this recurs.
- **`crashReporter.start()`** — Orca collects no crash data of its own today. That's the systemic version of this problem and needs its own issue (consent, privacy, an endpoint).

## Test plan

- New test file covering the pure builder (darwin SIGABRT explains; off-darwin — linux and win32 — doesn't claim the macOS cause; other darwin signals don't either; null signal stays clear) and `superviseForegroundServe` end-to-end (darwin SIGABRT throws the diagnostic, linux SIGABRT doesn't, numeric exits `0`/`7` return unchanged).
- `vitest src/cli/runtime` → **63 passed** across 9 files.
- `vitest src/cli` → **511 passed** across 39 files.
- `typecheck:cli` + `typecheck:node` clean; oxlint clean on touched files; max-lines ratchet OK.

### DIFF
--- a/src/cli/runtime/serve-signal-exit-diagnostic.test.ts
+++ b/src/cli/runtime/serve-signal-exit-diagnostic.test.ts
@@ -1,0 +1,99 @@
+import { EventEmitter } from 'node:events'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { serveSignalExitError } from './serve-signal-exit-diagnostic'
+import { superviseForegroundServe } from './serve-update-supervisor'
+import { RuntimeClientError } from './types'
+
+class FakeChildProcess extends EventEmitter {
+  kill = vi.fn()
+  pid = 5150
+}
+
+const originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform')!
+
+function setPlatform(platform: NodeJS.Platform): void {
+  Object.defineProperty(process, 'platform', { configurable: true, value: platform })
+}
+
+function superviseUntilExit(code: number | null, signal: NodeJS.Signals | null): Promise<number> {
+  const child = new FakeChildProcess()
+  const supervised = superviseForegroundServe({
+    executable: '/Applications/Orca.app/Contents/MacOS/Orca',
+    childArgs: ['--serve'],
+    spawnOptions: {},
+    spawnChild: vi.fn() as never,
+    handoffPath: null,
+    child: child as never,
+    expectedHandoff: null
+  })
+  child.emit('exit', code, signal)
+  return supervised
+}
+
+afterEach(() => {
+  Object.defineProperty(process, 'platform', originalPlatform)
+})
+
+describe('serveSignalExitError', () => {
+  it('explains the macOS window-server abort on darwin SIGABRT', () => {
+    const error = serveSignalExitError('SIGABRT', 'darwin')
+
+    expect(error).toBeInstanceOf(RuntimeClientError)
+    expect(error.code).toBe('runtime_serve_failed')
+    expect(error.message).toContain('SIGABRT during macOS application startup')
+    expect(error.message).toContain('macOS window server')
+    expect(error.data).toMatchObject({
+      nextSteps: [
+        expect.stringContaining('normal terminal session'),
+        expect.stringContaining('~/Library/Logs/DiagnosticReports/Orca-*.ips')
+      ]
+    })
+  })
+
+  it('does not claim the macOS cause off darwin', () => {
+    for (const platform of ['linux', 'win32'] as const) {
+      const error = serveSignalExitError('SIGABRT', platform)
+
+      expect(error.message).toBe('Orca serve exited via SIGABRT.')
+      expect(error.data).toBeUndefined()
+    }
+  })
+
+  it('does not claim the macOS cause for other darwin signals', () => {
+    const error = serveSignalExitError('SIGKILL', 'darwin')
+
+    expect(error.message).toBe('Orca serve exited via SIGKILL.')
+    expect(error.data).toBeUndefined()
+  })
+
+  it('stays clear when neither a code nor a signal is reported', () => {
+    expect(serveSignalExitError(null, 'darwin').message).toBe(
+      'Orca serve exited without reporting an exit code or signal.'
+    )
+  })
+})
+
+describe('superviseForegroundServe signal exits', () => {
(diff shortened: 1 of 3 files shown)


brennanb2025:
Implemented via Orca orchestration (task `task_9f6a5960a900`, dispatch `ctx_204605815a12`). Independently verified before opening: `format.ts:85-87` confirmed to render `data.nextSteps` generically for any `RuntimeClientError` in both human and `--json` output, and the 51 tests were re-run outside the implementing session.

PR #10469, opened 2026-07-25
fix(terminal): let grok emit real hyperlinks by advertising an xterm.js brand

nwparker:
Makes `grok` emit real OSC 8 hyperlinks in Orca, so wrapped links resolve natively instead of being re-derived from screen text.

Fixes the same user-visible bug as #10462 — hovering a two-line grok link highlighted the full URL, but clicking opened only `https://github.com/stablyai/orca/` — by removing the need to guess at all.

## ELI5

Terminals don't have links. They're a grid of characters, so an app that prints a URL is just painting text. For a click to work, *something* has to figure out where the URL starts and ends.

There are two ways that can happen:

1. **The app says so.** There's a standard escape sequence (OSC 8) that means "these cells are a link to this URL." The terminal doesn't guess — it's told.
2. **The terminal guesses.** It reads the characters back off the grid and tries to spot URL-shaped text.

Grok *already does #1*. Internally it knows every link's row, its exact columns, its URL, and even an `id` that ties together the pieces of one link split across lines.

Then it throws all of that away — because it doesn't recognize Orca.

Grok decides what a terminal can do by looking at `TERM_PROGRAM`. It keeps a list: iTerm2, WezTerm, Ghostty, VS Code, and so on. Orca isn't on it, so grok falls back to "unknown terminal," and unknown means *assume nothing works* — including hyperlinks. Safe default, wrong answer here.

So Orca was stuck on #2, guessing. And guessing breaks exactly where the bug showed up: grok wraps text at its own width, so a long URL splits across two lines with no marker saying "this line continues." The guess sees a line that ends early and concludes the URL ended too.

The fix is to stop being unrecognized. Orca now tells grok `TERM_PROGRAM=vscode` — only when launching grok. Grok emits the hyperlinks it already computed, and the terminal reads the URL instead of inferring it.

## Why `vscode` and not another brand

`TERM_PROGRAM` doesn't select a hyperlink switch — it selects a whole capability profile. Several brands would turn OSC 8 on; only one is right.

| Brand | OSC 8 | `id=` | Negotiates KKP | Native hover |
|---|---|---|---|---|
| **`vscode`** / `cursor` / `windsurf` / `zed` | Native | yes | **no — skipped** | **yes** |
| `WezTerm` / `alacritty` / `rio` / `foot` | Native | yes | **yes** | no |
| `ghostty` / `kitty` / `iTerm2` | Native | yes | **yes** | no |
| `WindowsTerminal` | Native | yes | no | no |
| `Orca` (today) | **Unknown → off** | — | no | no |
| `Warp` | Unsupported | — | — | — |
| `AppleTerminal` | HostileParser | — | — | — |

The VS Code family is the only profile written **for xterm.js**, which is Orca's renderer:

- **It skips Kitty keyboard protocol negotiation.** Grok's own comment says why — `mis-encoding shifted keys (xtermjs/xterm.js#5823)`. Claiming WezTerm or Ghostty would make grok negotiate KKP against a renderer that mishandles it: a link fix bought with an input bug.
- **It's the only Native brand with `native_link_hover: true`**, telling grok to stop drawing its own modifier-hover highlight and defer to the terminal. That removes the exact "highlight says one URL, click opens another" split from the original report.
- **`id_param: true`** is what lets a terminal group one link across the rows a TUI wrapped it onto.

The KKP skip costs nothing: `Unknown` skips it too, so `Shift+Enter` already falls back to `Alt+Enter` today. Only the wording of grok's message changes.

## Evidence

**Against the real grok binary:**

```
TERM_PROGRAM=Orca    → terminal: Unknown     ← today, hyperlinks suppressed
TERM_PROGRAM=vscode  → terminal: VS Code     ← hyperlinks emitted
```

**From grok's source** (`resolve_hyperlink_route`) — one gate discards fully-computed link geometry:

```rust
let emit_osc8 = caps.osc8 == Osc8Support::Native && skip.is_none();
```

**Proof the payoff is real** — replaying grok's exact output shape through xterm.js:

```
row 0: isWrapped=false urlIds=[1]
row 1: isWrapped=false urlIds=[1]
```

Both rows carry the **same** `urlId` despite no wrap flag. The complete URL resolves natively — no wrap-column inference, no block-width heuristics, no ambiguity.

## Scope and safety

- **Per launching agent, never process-wide.** `TERM_PROGRAM` is read by everything spawned in a terminal; Orca must not misreport itself globally to satisfy one TUI. Non-grok agents and plain shells are untouched.
- **Orca's true identity stays available** as `ORCA_TERM_PROGRAM`.
- **All three spawn paths** — local, daemon, relay (SSH) — so a remote grok session behaves like a local one.
- **Contract test against the installed grok binary**, asserting `Orca → Unknown` and `vscode → VS Code`. If grok renames or drops that brand, this fails loudly instead of silently regressing to truncated links. Skipped when grok isn't installed, so CI stays green.

## Honest caveat

This is a **stopgap**, and the PR says so in the code and docs. Orca is claiming to be something it isn't, and grok will keep adding VS Code-specific quirks that Orca would silently inherit.

The durable fix is upstream — an Orca entry in grok's own table:

```rust
"orca" => Some(TerminalName::Orca),
// + a capability arm copying the VS Code profile: the justification is
//   identical (xterm.js renderer)
```

`docs/reference/terminal-hyperlink-brand-advertising.md` records the exact upstream patch and the criteria for deleting this workaround.

## Relationship to #10462

#10462 fixes the same bug by *reconstructing* wrapped URLs from screen text. The two are complementary, not redundant:

- **This PR** is the correct fix for grok — exact data, no inference.
- **#10462** still covers any other TUI that wraps at its own width without emitting OSC 8, and users on older grok builds.

They compose safely: when grok emits OSC 8, xterm's native path resolves the link and the heuristic never engages. See the discussion there for the merge recommendation.

## Test plan

- [x] `agent-terminal-brand-env.test.ts` — override applies to grok only; other agents and plain shells untouched; `ORCA_TERM_PROGRAM` preserved
- [x] `agent-terminal-brand-env.grok-contract.test.ts` — real grok binary classifies `Orca` as Unknown and `vscode` as VS Code
- [x] `local-pty-provider.test.ts` — override reaches node-pty via both explicit `launchAgent` and recognized startup command; not applied otherwise
- [x] Typecheck (node, web, cli), lint, format clean
- [x] 5,330 tests pass across `src/shared`, `src/main`, `src/relay`

Two pre-existing failures in `src/relay/agent-exec-handler.test.ts` reproduce on unmodified `main` (a local git-config env difference) and are unrelated.

## Continuation review

- Review-loop count: 1
- Final verdict: PASS for the all-spawn-path, Grok-only compatibility change.
- Revalidated live GitHub state: CLEAN merge state; completed CI checks are successful.

## Proof of no regressions / residual risk

The override is scoped to Grok launches; ordinary shells and other agents retain Orca identity, while ORCA_TERM_PROGRAM preserves that identity for Grok. PTY-path tests cover local, daemon, and relay/SSH paths; a real Linux Grok scroll surface is still required before claiming zero regressions for the visible behavior.


LesleyMurfin:
## Related: Grok mouse wheel / scrollback dead under `TERM_PROGRAM=Orca`

Same root cause as the hyperlink path this PR fixes: Grok’s brand table has no `Orca` entry, so `/doctor` reports `terminal: Unknown` and capability-gated paths (OSC 8 **and** mouse/scroll) fall back to the unknown profile.

### Repro (Orca 1.4.155, Linux serve + Mac desktop, Grok 0.2.112)

1. Launch Grok in an Orca agent pane (`TERM_PROGRAM=Orca` hard-coded on the PTY).
2. `/doctor` → **terminal: Unknown**; notifications fall back to BEL.
3. `/scroll-debug` while scrolling → `stream:- kind:- ev:-` (no wheel events applied).
4. Cannot highlight/select text (mouse capture on) but scrollback does not move.
5. Claude Code in the **same** Orca pane scrolls fine (different brand/capture path).
6. Spoofing `TERM_PROGRAM=vscode` for the Grok process (PATH shim, same brand choice as this PR) restores scroll **and** expected VS Code brand detection.

### Why this PR is the right Orca-side stopgap for scroll too

| Brand | Grok result | Notes |
|---|---|---|
| `Orca` | Unknown | Mouse/scroll + OSC 8 suppressed/weak |
| `vscode` | VS Code | xterm.js profile; **no** Kitty KB negotiation |

Durable fix remains Grok-side: add `Orca` with the VS Code / xterm.js capability arm. Until then, landing this PR (Grok-only brand export on local/daemon/relay) unblocks hyperlinks **and** mouse scroll for all Orca users without host shims.

### Host workaround we deployed while waiting

PATH wrapper ahead of real `grok` that sets `TERM_PROGRAM=vscode` and preserves `ORCA_REAL_TERM_PROGRAM` / `ORCA_REAL_TERM_PROGRAM_VERSION`. Remove once this ships in a release.

Happy to re-test on Linux serve + Mac after merge.

PR #10471, opened 2026-07-25
fix(terminal): copy full session context without char budget

### LINKED ISSUE #10395: Copy Context should include full context, not truncated output
### Problem or use case

"Copy Context" currently truncates long sessions. In long sessions, only ~10–20% of the session is copied, and the start is replaced with something like:

```
[Earlier terminal output omitted: 3107 characters]
```

Users often need the **full** session (including complete tool execution output) — for example to paste into another model/analysis tool. Truncated context is not useful for that.

### Proposed solution

**Copy Context should copy the full context**, not a truncated/cropped version.

- Include the complete terminal/session output (full tool execution output)
- Do not omit earlier output with placeholders like `[Earlier terminal output omitted: N characters]`

Originally proposed as a separate "Export session" action that writes the full session to a temp file. A full Copy Context is the primary ask; export could still be a nice follow-up if clipboard size becomes a limit.

Original report: people currently work around this with scripts because they need the complete session, not the cropped one.

innocarpe:
## Summary
- **Copy Context** no longer truncates long sessions to ~36k characters with `[Earlier terminal output omitted: N characters]`
- Serializes the pane's full configured scrollback (not a fixed 800-line tail)
- Still strips ANSI/OSC and leaves **fork-session prompts** on the existing budget so agent injects stay bounded

Closes #10395

## Motivation
Users paste Copy Context into other models/analysis tools and need the complete terminal session, including early tool output. Truncation made the action unusable for long runs.

## Changes
- `buildFullSessionTranscript()` — clean only, no char budget
- `copyAgentSessionContextFromPane` uses full transcript + `terminal.options.scrollback`
- Fork path (`buildAgentSessionForkPrompt` / `buildBoundedSessionTranscript`) unchanged

## Test plan
- [x] Unit: full transcript retains early + late markers; bounded path still omits early
- [x] Existing fork/cleanup tests pass (10 total)
- [ ] Manual: long agent session → Copy Context → paste includes start of session without omit marker

## Notes
Clipboard write still respects the existing 16 MB write limit in main; oversized captures surface as a copy error toast rather than a silent omit.

## ELI5

Copy Context used to cut off long terminal sessions after ~36k characters, so early tool output was missing when you pasted elsewhere. Now it copies the full configured scrollback instead. Agent fork prompts still stay on a size budget so they do not get huge.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-agent-session-fork.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-agent-session-fork.ts
@@ -3,8 +3,9 @@ import type { ManagedPane } from '@/lib/pane-manager/pane-manager'
 import { launchAgentInNewTab } from '@/lib/launch-agent-in-new-tab'
 import {
   buildAgentSessionForkPrompt,
-  buildBoundedSessionTranscript
+  buildFullSessionTranscript
 } from '@/lib/agent-session-fork-context'
+import { DESKTOP_TERMINAL_SCROLLBACK_ROWS_DEFAULT } from '../../../../shared/terminal-scrollback-policy'
 import { activateAndRevealWorktree } from '@/lib/worktree-activation'
 import { useAppStore } from '@/store'
 import { makePaneKey } from '../../../../shared/stable-pane-id'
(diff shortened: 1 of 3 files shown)


innocarpe:
Status ping: focused Copy Context fix — full transcript without the char budget omit path (tests). No open threads.

PR #10473, opened 2026-07-25
fix(macos): close both macOS 26 main-thread deadlock doors and restore the reveal reflow

nwparker:
## Why

A user's app froze for 109 minutes on macOS 26 Tahoe (Darwin 25.5.0, v1.4.152) and had to be force-quit. macOS 26 backs AppKit objects with scenes, and a scene update sent re-entrantly from inside AppKit's own dispatch self-deadlocks the main thread in FrontBoardServices (`FBSScene _sendUpdate:` -> `BSServiceDispatchQueue performAsyncAndWait:` -> `_dispatch_event_loop_wait_for_ownership`). Because the freeze is on the main process, every telemetry path is silenced with it — the app just stops.

#10253 fixed the window half by skipping the repaint size nudge on macOS 26. This PR closes the remaining hole and repairs the regression that fix introduced.

## What this changes

**1. The reveal reflow was lost on macOS 26 (regression from #10253).**
`webContents.invalidate()` repaints but does not *reflow*, so the `h-dvh` app-shell root kept a stale dynamic-viewport height and the StatusBar stayed clipped below the viewport — the exact "no bottom bar" bug #10056 had just fixed. On Tahoe `forceRepaint` returned right after `invalidate()`, so every reveal, restore and wake got the known-insufficient behavior. Neither commit has shipped in a stable release, so this was caught before users saw it.

The fix drives the reflow through Chrome device emulation: a +1px emulated viewport, reverted a frame later. The renderer recomputes layout for real, and no NSWindow mutation ever happens — so the deadlock is still never reachable.

**2. The tray was a second, un-fixed door into the same deadlock.**
`window.on('show'/'restore')` calls `setTrayAttention(false)` straight from the AppKit handler, and `tray.setImage`/`setToolTip` drive an `NSStatusItem` scene update — the same re-entrant pattern, matching the stackshot in openai/codex#23695. The native mutation now runs on a fresh event-loop turn. The attention flag itself still flips synchronously, or rapid show/hide would mis-dedupe against a value that had not landed yet.

**3. Maximized and fullscreen windows never reflowed.**
That bail-out predates the Tahoe path and existed only to stop `setSize` from resizing a window out of those states. Emulation leaves the frame alone, so the guard was suppressing the reflow for nothing.

**4. A failed restore could strand the renderer.**
If `disableDeviceEmulation` threw while the `webContents` was still alive, the first version cleared its latch anyway — pinning the renderer at the overshot viewport for the window's lifetime and letting the next reveal stack another cycle on top. The restore now retries on a bounded schedule and holds the latch while a retry is pending.

**5. Two smaller defects found along the way.**
- The pre-Tahoe size jiggle could leak its `WeakSet` latch if `setSize` threw, permanently suppressing every later repaint for that window. Now cleared in a `finally`.
- The Tahoe test only drove `show`; `powerMonitor 'resume'` — the other dispatch context implicated in the incident — was uncovered. Now tested.

## Verification

Behavior was measured against real Electron 43.1.0 on macOS 26.3.1 (Darwin 25.3.0), not just mocks:

- the renderer receives the resize and relayouts; layout ends correct in every configuration tried
- `getSize()` is unchanged throughout — the native frame is never touched
- viewport and `devicePixelRatio` restore exactly, across zoom levels 0/+1/-1
- maximized and fullscreen windows reflow and stay maximized/fullscreen
- overlapping reveals collapse to one cycle; the latch re-arms after, and is held across retries
- window destroyed mid-cycle, and `enable`/`disable` throwing, leave no stuck emulation
- 40 aggressive show/hide/minimize/restore cycles against the combined window+tray logic, with the main thread continuously responsive
- **no SIGWINCH**: even with a pane tuned to sit exactly on an xterm row boundary — the worst case — the 32 ms revert lands before the fit stabilizes, so no pane reports new geometry. The old jiggle nudged *width*, which was strictly riskier.

Every new test was mutation-checked: reverting the corresponding fix makes it fail, so none pass vacuously.

`pnpm typecheck:node` clean, `oxlint` clean, full main-process suite green (**15,000 tests**).

## Limits worth stating

- **The original deadlock is not deterministically reproducible.** It is a race; repeated attempts on this machine (both sync frame and sync tray mutation from AppKit callouts) completed normally. The fix rests on the documented mechanism and the stackshots, plus the fact that the crashing v1.4.152 build mutated the frame *synchronously* inside the callout.
- **The dvh staleness itself did not reproduce locally either**, so item 1 is verified at the mechanism level — a real reflow reaches the renderer without touching the frame — rather than by observing the clipped status bar recover.
- With the renderer blocked past the 32 ms window, Chromium coalesces the intermediate viewport, so the renderer never observes the +1px step. Layout still ends correct, and the 250 ms second repaint plus the next reveal both retry.

## Notes

- Behavior on macOS < 26, Windows and Linux is unchanged apart from the added `try/finally`.
- One consequence at parity with what non-Tahoe macOS already did via the jiggle: an open text-selection popup or link bubble dismisses on reveal.

### DIFF
--- a/src/main/appkit-scene-mutation.test.ts
+++ b/src/main/appkit-scene-mutation.test.ts
@@ -1,0 +1,32 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { deferAppKitSceneMutation } from './appkit-scene-mutation'
+
+beforeEach(() => {
+  vi.useFakeTimers()
+})
+
+afterEach(() => {
+  vi.useRealTimers()
+})
+
+describe('deferAppKitSceneMutation', () => {
+  it('runs the mutation on a later turn, never inside the caller stack', () => {
+    const mutate = vi.fn()
+
+    deferAppKitSceneMutation(mutate)
+    expect(mutate).not.toHaveBeenCalled()
+
+    vi.advanceTimersByTime(0)
+    expect(mutate).toHaveBeenCalledOnce()
+  })
+
+  it('keeps scheduled mutations in call order', () => {
+    const order: string[] = []
+
+    deferAppKitSceneMutation(() => order.push('first'))
+    deferAppKitSceneMutation(() => order.push('second'))
+    vi.advanceTimersByTime(0)
+
+    expect(order).toEqual(['first', 'second'])
+  })
+})
--- a/src/main/appkit-scene-mutation.ts
+++ b/src/main/appkit-scene-mutation.ts
@@ -1,0 +1,6 @@
+// Why: macOS 26 backs AppKit objects (windows, NSStatusItem) with scenes, and a scene
+// update sent re-entrantly from inside AppKit's own dispatch self-deadlocks the main
+// thread in FrontBoardServices; a fresh event-loop turn vacates that callout frame.
+export function deferAppKitSceneMutation(mutate: () => void): void {
+  setTimeout(mutate, 0)
+}
--- a/src/main/tray/system-tray.test.ts
+++ b/src/main/tray/system-tray.test.ts
@@ -142,7 +142,14 @@ function builtMenuItems(): MenuItem[] {
   return menuFromTemplateMock.mock.calls.at(-1)?.[0] as MenuItem[]
 }
 
+// Why: tray image/tooltip writes are deferred off the caller's stack to keep the
+// NSStatusItem scene update out of AppKit's dispatch; run the pending turn.
+function flushTraySceneMutation(): void {
+  vi.advanceTimersByTime(0)
+}
+
 beforeEach(() => {
+  vi.useFakeTimers()
   trayInstances.length = 0
   menuFromTemplateMock.mockClear()
   composeAttentionMock.mockClear()
@@ -180,6 +187,7 @@ beforeEach(() => {
 
 afterEach(() => {
   setPlatform(originalPlatform)
+  vi.useRealTimers()
   vi.restoreAllMocks()
 })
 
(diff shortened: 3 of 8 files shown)


nwparker:
**Post-merge attribution: this PR is very likely the fix for the reported "freeze after closing my laptop and coming back."**

Investigating that report (v1.4.152, macOS 26 Tahoe / Darwin 25.5.0, app unresponsive ~109 min, force-quit), the evidence points here:

- **v1.4.152 does not contain #10253.** Verified: `git merge-base --is-ancestor 4569a4fe71 v1.4.152` → false. So that build shipped `forceRepaint()` calling `window.setSize(width + 1, height)` **synchronously** from `window.on('restore')` / `on('show')` (`src/main/window/createMainWindow.ts:66` at that tag), with no macOS 26 guard.
- **Lid-open wake fires `restore`/`show`**, so the nudge ran re-entrantly inside AppKit's window-state dispatch — exactly the pattern #10253's kernel-confirmed spindump identified as self-deadlocking the main thread in FrontBoardServices on scene-backed Tahoe windows.
- **It explains the missing crash report.** A main thread deadlocked in AppKit cannot run the JS that would record one, which is why the only trace was a heartbeat gap.

An earlier stage of this investigation reported the 109-minute gap as "very likely OS sleep, not a hang." That was a mistake: the gap *is* sleep-shaped, but sleep is the precondition, not the alibi — the freeze is at the tail of the gap, on wake. Corrected in #10530.

**Caveat, stated plainly:** a synthetic repro driving `show()`/`minimize()`/`restore()` on macOS 26.3.1 did **not** reproduce the hang (6/6 cycles clean; harness validated to fire the handlers and reach `setSize`). Programmatic window-state changes don't reproduce a genuine FrontBoardServices lid-open wake, so this is weak counter-evidence against a kernel-confirmed spindump — but the deadlock remains **inferred, not reproduced**.

Current `main` closes both doors: on Tahoe `forceRepaint` returns before any `setSize` (emulation reflow instead), and on older macOS the nudge is deferred to a fresh event-loop turn. `isMacosTahoeOrNewer` correctly maps the user's Darwin 25.5.0 to the safe path.

PR #10474, opened 2026-07-25
fix(relay): tolerate wall-clock skew in host-proof validation

### LINKED ISSUE #10401: [Bug]: Mobile Relay host-proof silently rejects every handshake when the local clock is a few seconds off (no skew tolerance)
### Operating system

macOS 15.6.1 (Apple Silicon)

### Orca version

v1.4.153-rc.3

#### Short summary

Mobile Relay fails silently in an endless "connecting → unavailable" loop when the local system clock is a few seconds slow, because the desktop-side host-proof transcript validation uses the raw local wall clock with zero skew tolerance. The UI gives no indication of the failure reason, and nothing about the relay attempts is written to `main.trace.ndjson` or `daemon.log`.

This is a follow-up to stablyai/orca#10273. After upgrading to v1.4.153-rc.3 (which includes the bounded retry from [#10147](<https://github.com/stablyai/orca/issues/10147>) and the handshake deadlines from [#10260](<https://github.com/stablyai/orca/issues/10260>)), the broker retried correctly — but every attempt still failed at the cell control handshake.

#### What happened

Observed from the local proxy's connection log: the broker cycled through the full startup sequence roughly every 7–20 seconds:

1. `login.onorca.dev:443` — TCP/TLS OK (token exchange appears to succeed)
2. `relay.onorca.dev:443` — TCP/TLS OK (assignment appears to succeed)
3. `c2.relay.onorca.dev:443` — TCP/TLS OK, then the connection drops; a few c2-only reconnect attempts follow, then the cycle restarts from token exchange

All three `/health` endpoints returned HTTP 200 throughout. The pairing UI only ever showed Relay as unavailable with a LAN-only QR code.

#### Root cause

Reading the shipped `out/main/index.js` (v1.4.153-rc.3 asar), the host-proof transcript validation is:

```js
const now = (context.now ?? Date.now)();
// ...
return issuedAt !== null && issuedAt <= now && now <= challenge.expiresAt
  && challenge.expiresAt - issuedAt <= 1e4 && ...
```

`now` is the local wall clock, the challenge validity window is at most 10 seconds, and there is no skew tolerance in either direction.

The machine's clock was ~4.4 seconds behind real time (verified with `sntp time.apple.com`, offset `+4.44s ± 0.11`, and cross-checked against HTTP `Date` headers from two independent servers). The relay-issued `issuedAt` was therefore always in the local future, `issuedAt <= now` was always false, and the desktop itself rejected every host-proof attempt. With a ≤10s window, any skew beyond a few seconds (in either direction, after accounting for network latency) makes the handshake fail deterministically — so the [#10147](<https://github.com/stablyai/orca/issues/10147>) retry loop just spins forever.

#### Confirmation

Running `sudo sntp -sS time.apple.com` (no Orca restart, no config change) resolved it immediately: the retry loop registered within seconds and mobile control over Relay started working.

#### Reproduction

1. Set the system clock ~5 seconds behind true time.
2. Sign in to a cloud-linked profile and open Settings → Mobile → Orca Relay.
3. The broker cycles connecting → offline indefinitely; the UI only shows Relay unavailable / a LAN-only pairing code, with no reason.

#### Suggestions

* Tolerate a reasonable skew (e.g. ±30s), or derive a server-time offset from the token exchange / assignment responses and validate the transcript against that instead of raw `Date.now()`.
* Surface the failure stage in the pairing UI and write relay broker attempt/failure events to the app logs — currently a host-proof rejection is indistinguishable from a network outage, which makes this class of failure very hard to diagnose.

Clocks that are a few seconds off are common on machines where NTP is impaired (e.g. TUN/VPN proxies intercepting UDP 123) — likely correlated with exactly the user population that needs Relay instead of LAN pairing.

innocarpe:
## Summary
- Mobile Relay host-proof validation used raw local `Date.now()` with **zero skew tolerance** against a ≤10s challenge window
- A few seconds of clock lag made `issuedAt` look \"in the future\", so every handshake failed and Relay looped connecting → unavailable with no useful UI reason
- Allow **±30s** wall-clock skew while still enforcing the server challenge window length (≤10s)

Closes #10401

## Root cause
From the issue: with local clock ~4.4s behind NTP, server-issued `issuedAt` always failed `issuedAt <= now`, so the desktop rejected host-proof itself. Broker retries could not recover until the clock was corrected.

## Change
In `validateTranscript` (`relay-host-proof.ts`):

```ts
issuedAt - SKEW <= now && now <= challenge.expiresAt + SKEW
// SKEW = 30_000 ms
// still require challenge.expiresAt - issuedAt <= 10_000
```

## Test plan
- [x] Unit: accepts ~4.4s behind / ahead clocks
- [x] Unit: rejects beyond ±30s
- [x] Unit: still rejects oversized challenge windows
- [x] Existing `relay-control-client` host-proof flow passes
- [ ] Manual (optional): set clock ~5s slow → Relay pairing succeeds without NTP fix

## Notes
UI/logging for host-proof failures (also requested in the issue) is left as follow-up so this PR stays a minimal handshake fix.

## ELI5

Mobile Relay handshakes failed when the phone clock was a few seconds off the server, looping connect → unavailable. Host-proof validation allows a small clock skew so normal lag no longer blocks pairing.

### DIFF
--- a/src/main/runtime/relay/relay-host-proof.test.ts
+++ b/src/main/runtime/relay/relay-host-proof.test.ts
@@ -1,0 +1,204 @@
+import { createHmac, randomBytes } from 'node:crypto'
+import { describe, expect, it } from 'vitest'
+import nacl from 'tweetnacl'
+import {
+  answerRelayHostChallenge,
+  RELAY_HOST_PROOF_CLOCK_SKEW_MS,
+  type RelayHostChallenge,
+  type RelayHostProofContext
+} from './relay-host-proof'
+
+const encoder = new TextEncoder()
+const HOST_PROOF_DOMAIN = 'orca-relay-host-proof/v1'
+const CHALLENGE_DOMAIN = 'orca-relay-host-challenge/v1'
+
+function concat(parts: readonly Uint8Array[]): Uint8Array {
+  const output = new Uint8Array(parts.reduce((total, part) => total + part.byteLength, 0))
+  let offset = 0
+  for (const part of parts) {
+    output.set(part, offset)
+    offset += part.byteLength
+  }
+  return output
+}
+
+function uint32(value: number): Uint8Array {
+  const bytes = new Uint8Array(4)
+  new DataView(bytes.buffer).setUint32(0, value, false)
+  return bytes
+}
+
+function uint64(value: number): Uint8Array {
+  const bytes = new Uint8Array(8)
+  new DataView(bytes.buffer).setBigUint64(0, BigInt(value), false)
+  return bytes
+}
+
+function field(name: string, value: Uint8Array): Uint8Array {
+  const encodedName = encoder.encode(name)
+  return concat([uint32(encodedName.byteLength), encodedName, uint32(value.byteLength), value])
+}
+
+function text(value: string): Uint8Array {
+  return encoder.encode(value)
+}
+
+function buildTranscript(input: {
+  origin: string
+  relayKey: Uint8Array
+  nonce: Uint8Array
+  challengeId: string
+  issuedAt: number
+  expiresAt: number
+  relayHostId: string
+  hostKey: Uint8Array
+  userId?: string
+  profileId?: string
+  organizationId?: string
+  assignmentEpoch?: number
+}): Uint8Array {
+  return concat([
+    field('protocol', text(HOST_PROOF_DOMAIN)),
+    field('version', new Uint8Array([1])),
+    field('relayOrigin', text(input.origin)),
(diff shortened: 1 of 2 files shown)


brennanb2025:
Thanks for the focused fix — the bounded skew tolerance preserves the signed challenge window, and the regression coverage gives us good confidence. Merged!

innocarpe:
Thanks @brennanb2025 — glad the bounded skew + regression tests were enough to land cleanly. Appreciate the merge!

PR #10480, opened 2026-07-25
fix(mobile): heal an orphaned native-chat image paste across screen unmounts

brennanb2025:
## Summary

Mobile native chat marks a terminal's composer line "stale" when an image paste lands but the follow-up text+Enter never confirms, so the next send clears it with a leading Ctrl+U first. That marker was component-local React state:

- `mobile/src/session/use-mobile-native-chat-image-attachments.ts:113` (pre-fix) — `staleInputTerminalsRef = useRef(new Set<string>())`

The hook is per-session-screen (`mobile/app/h/[hostId]/session/[worktreeId].tsx:3623` → `mobile/src/session/use-mobile-session-image-attachments.ts:73`), but the condition it tracks — a bracketed paste sitting unsubmitted on the agent's composer line — lives on the **host** and outlives the screen.

**Failure 1 — marker lost on unmount.** Attach an image and send; the paste lands (`mobile/src/session/mobile-native-chat-image-send.ts:39-52`), the text+Enter returns `unknown`, so the terminal is marked stale and the chips clear. Back out of the session screen and return → the hook remounts with an empty `Set` → the stale check is `false`, the healing `\x15` is skipped, and the agent receives `<image path><text>`: the image is silently re-attached and the message body corrupted. This is exactly what the in-code comment ("both must heal first") says must not happen.

**Failure 2 — one reader.** The marker was read at exactly one site. The chat overlay wires the question card straight past the image hook that owned that read (`mobile/src/session/MobileNativeChatOverlay.tsx:51` — `onAnswerQuestion={controller.handleNativeChatSend}`), and the ask-answer path submits its own Enter, so both could commit on top of the orphaned paste.

### The fix

New `mobile/src/session/mobile-native-chat-stale-input.ts` owns the marker in a module-level store keyed by terminal handle — a lifetime that matches the host-side condition instead of the screen — plus a single `healMobileNativeChatStaleInput` that clears the line and consumes the marker only once the host accepts the clear.

Consulted and consumed by every write path that can **submit the composer**:

| Path | Why |
|---|---|
| image hook text-only send | pre-existing reader, now on the shared store |
| `handleNativeChatSendWithOutcome` | closes the question-card overlay bypass |
| `answerAsk` (pasted-label shape only) | it pastes the label into the composer and commits it |

**User-visible change:** a message sent after a session screen was left and re-entered no longer silently re-attaches the previous image. If the healing clear itself is rejected, the send is refused with the existing "Message not sent" / "Answer not sent" toast rather than submitting onto a dirty line.

### Deliberate scope call (deviates from the filed spec)

The task asked for *permission/ask/cancel* to heal too. Ask does, but **only in its pasted-label shape** — see the correction below. **Permission choices and the Escape cancel deliberately do not**, and healing them would be a regression rather than a fix: those are `enter: false` keys for an active overlay that swallows the clear (`mobile-native-chat-permission.ts` sends only `1`/`y`/`n`/`a`/ESC, never Enter). A Ctrl+U there would not reach the composer, yet would still *consume* the marker — leaving the next real message to be corrupted by the paste still sitting there. That is Failure 1 through a different door. Since these paths cannot submit the composer, a stale line cannot corrupt them, and leaving the marker armed for the next text send is strictly safer.

Desktop scopes its Ctrl+U at least as narrowly: `clearUnsubmittedAgentInput` is called only from `sendNativeChatMessage` and `sendNativeChatMessageWithImageAttachments` (`native-chat-runtime-send.ts:73, :166`) and their cancel hooks, so approval digits and ESC bypass it entirely (`use-native-chat-interactive-send.ts:74-82, :155-158`). The comment at `native-chat-runtime-send.ts:33-37` carves out *verified option commands* for a different reason (a model-switch observer can miss confirmation markers), so it is not the citation for this skip. All skips are commented in-code so they read as decisions, not oversights.

**Correction (review, `86f6c81982`).** The first version of this change healed *every* ask answer, on the stated premise that "both answer shapes commit with Enter." That is false for the selector shapes, and the heal was over-broad in exactly the way this section argues against: Claude's single-question single-select builds `[{raw:'2'}]` and Codex's builds `[{raw:'1'}]` — no Enter at all (`src/shared/native-chat-ask.ts:211-213, :259-262`; the trailing `ASK_ENTER` is gated at `:220-223` on multi-question or multi-select) — and every stepping group is written `enter: false` (`use-mobile-native-chat-answer-send.ts:187`), which the host coerces anyway (`src/main/runtime/rpc/methods/terminal.ts:1345`). A bare option digit cannot submit the composer, so healing there burned the one-shot marker against the overlay where the live-overlay premise is *strongest* (the ask card is built from the structured `interactivePrompt` and the user is tapping it), leaving the paste to corrupt the next real message. That was newly reachable in this PR — on `main` the ask path never touched the marker. The heal is now scoped to the pasted-label shape, matching desktop's split at `use-native-chat-interactive-send.ts:132-140`, and the three skips (selector answer, permission choice, Escape cancel) are test-pinned so the argument is an invariant rather than a comment.

Residual, stated plainly: for the multi-question / multi-select / free-text shapes that *do* end in a raw `\r`, a stale card whose overlay already closed can still submit the composer. Desktop accepts the same risk, and the selector groups would land as literal text in that case anyway, so the line is garbage either way — but the marker now survives to clean it before the next real message. Properly fixing that belongs in card invalidation (`use-mobile-native-chat-ask-dismiss.ts`), not in pre-clearing.

## Screenshots

No visual change. This is terminal-write sequencing behind the existing composer; no UI, layout, copy, or token changes.

## Testing

- [x] `pnpm lint` — root `npm run lint` passes end-to-end; `mobile/` `oxlint` clean (incl. `config/oxlint-react-doctor.json` via pre-commit) and `oxfmt --check` clean.
- [x] `pnpm typecheck` — root passes; `mobile/` `tsc --noEmit` clean.
- [x] `pnpm test` — **full mobile suite: 339 files / 2465 tests pass**, 2 skipped (2462 at `bb46ca226a`, +3 from the review fix). The root desktop suite was not run: the change is entirely under `mobile/`, and the root tsconfig/vitest do not compile or include `mobile/`.
- [ ] `pnpm build` — not run. Electron desktop packaging is unaffected by a mobile-only change.

**Regression tests were verified to actually catch the bug.** I stashed only the source changes while keeping the new tests, and confirmed the 5 behavior tests fail against pre-fix code and pass after:

- marker lost on unmount/remount (Failure 1)
- overlay/question-card send never clears (Failure 2)
- ask-answer send never clears (Failure 2)
- a rejected healing clear must retain the marker and refuse the send (both send and answer paths)

Plus `mobile-native-chat-stale-input.test.ts` unit-covers per-terminal isolation, no-write-when-unmarked, consume-on-accept, and marker retention when the clear is rejected or throws.

All touched files stay under their `max-lines` caps with no disable and no per-file bump — the image hook shrank 281 → 272 effective lines.

## AI Review Report

Reviewed by CodeRabbit (`bb46ca226a`, all 9 files): **no actionable comments generated** — zero inline findings. Two non-code pre-merge checks were raised; the description one is actioned by this rewrite, and I've declined the 80% docstring-coverage threshold in a PR comment with reasoning, because hitting it would mean documenting three self-describing one-line `Set` wrappers, which `AGENTS.md` ("Concise/Brief Non-obvious comments ONLY … DO NOT explain the obvious") explicitly prohibits.

My own review focused on the risks specific to moving state out of React:

- **Test isolation** — module-level state outlives a test. This actually surfaced as a real failure during development (an unrelated test inherited a marker) and is fixed by `resetMobileNativeChatStaleInputForTests()` in the `beforeEach` of all three affected suites.
- **No double-clear on the image path** — after a successful paste the marker is cleared before `baseSend`, so the controller's new heal is a no-op there; verified by the pre-existing ride-along tests that assert exact `terminal.send` counts.
- **Accept semantics unchanged** — the heal reuses the same `pasteMobileNativeChatImagePaths` → `isTerminalSendRpcAccepted` path as the old inline block, so only an explicit `accepted === true` consumes the marker; an ambiguous ack conservatively retains it.
- **Supersession** — the ask path captures its generation before the heal's `await` and re-checks it after, so a concurrent answer cannot double-step the selector.
- **Writer audit** — I enumerated every `terminal.send` writer under `mobile/src/session/` to confirm none that can submit the composer was missed. `use-mobile-native-chat-stop.ts` sends only Escape, so it is correctly in the skip category.

**Cross-platform:** nothing platform-dependent is touched. No `metaKey`/`ctrlKey` or accelerators, no shortcut labels, no filesystem paths or path separators, no shell invocation, no Electron main/renderer or IPC code, no native modules. The change is pure TypeScript in the React Native client operating on an existing RPC, so it behaves identically for macOS, Linux, and Windows hosts, and over SSH and folder workspaces as well as git worktrees (a terminal handle is opaque to this logic).

## Security Audit

Low risk; no new attack surface.

- **Input handling** — no user-controlled data is newly parsed or interpolated. The only byte introduced is a fixed `\x15` control character sent to a terminal the user already owns and already writes to via this same path.
- **No new RPC surface** — reuses the existing `terminal.send` request and the existing device-token client field; no new method, no protocol version change, no runtime-compatibility concern.
- **Command execution / path handling** — none added. Image paths are not newly constructed or resolved; the heal sends an *empty* `imagePaths` list.
- **Auth / secrets** — untouched. The device token is passed through exactly as before and is never logged.
- **State scope** — the new module-level store holds only opaque terminal-handle strings, never image bytes, paths, or message text.

## Notes

Residual risk and adjacent work:

- The store is in-memory, so a **full app restart** still loses the marker while the host line stays dirty. That is narrower than the bug fixed here (leaving a screen is routine; a cold start is not) and avoids putting an `await` on persisted storage in front of every send. A persisted store keyed by handle would close it.
- Markers are never reaped or invalidated. Handles are globally unique (`term_<uuid>` per pty), so a marker can never clear the wrong terminal, and an entry for a dead handle is inert. But it can go stale on a **live** handle: nothing observes the host line being cleaned out of band — the desktop user submitting or Ctrl+U-ing it themselves, or the same tab viewed as a raw terminal and submitted there. The next mobile send then fires one `\x15` that wipes whatever is on that line. Pre-fix this window closed on screen unmount; now it stays open, which is the cost of matching the marker's lifetime to the host-side condition. Reaping markers for handles absent from the session route's tab list would narrow it.
- A question-card answer tapped during an image send's 300ms settle can still interleave with that send's paste. Pre-existing and independent of the marker; left alone to keep this to one concern.
- **Adjacent surface deliberately left untouched**, flagged as follow-up rather than widened into here: `mobile/src/session/use-mobile-diff-review-send-actions.ts:77-80` submits a review prompt with `enter: true` to a chosen terminal, so it can also commit on top of a stale paste. It is not a native-chat write path and does not regress with this change, but the same heal would apply.

**Not device-verified.** No phone or paired relay was used — this reproduces the defect at the hook/controller seam with a stubbed RPC transport, not against a real agent TUI. Specifically not verified on-device: that the healing `\x15` visibly clears a real Claude/Codex composer after a backgrounded-and-restored screen, and the premise above that a live permission overlay swallows Ctrl+U. Worth one manual pass: attach an image, send on a flaky connection until the send reports unconfirmed, back out of the session and return, then send a plain message and confirm no image rides along.

### DIFF
--- a/mobile/src/session/mobile-native-chat-permission-send.test.ts
+++ b/mobile/src/session/mobile-native-chat-permission-send.test.ts
@@ -1,7 +1,17 @@
-import { describe, expect, it, vi } from 'vitest'
+import { createElement } from 'react'
+import { act, create, type ReactTestRenderer } from 'react-test-renderer'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import type { RpcClient } from '../transport/rpc-client'
 import { markRpcDeliveryUnknown } from '../transport/rpc-delivery-ambiguity'
-import { sendMobileNativeChatPermissionResponse } from './mobile-native-chat-permission-send'
+import {
+  sendMobileNativeChatPermissionResponse,
+  useMobileNativeChatPermissionSend
+} from './mobile-native-chat-permission-send'
+import {
+  isMobileNativeChatInputStale,
+  markMobileNativeChatInputStale,
+  resetMobileNativeChatStaleInputForTests
+} from './mobile-native-chat-stale-input'
 
 describe('sendMobileNativeChatPermissionResponse', () => {
   it('writes an approval as raw bytes without appending Return', async () => {
@@ -39,5 +49,52 @@ describe('sendMobileNativeChatPermissionResponse', () => {
         text: '1'
       })
     ).resolves.toBe('unknown')
+  })
+})
+
+describe('useMobileNativeChatPermissionSend', () => {
+  let renderer: ReactTestRenderer | null = null
+  let respond: ((text: string) => Promise<boolean>) | null = null
+
+  beforeEach(() => {
+    globalThis.IS_REACT_ACT_ENVIRONMENT = true
+    resetMobileNativeChatStaleInputForTests()
+  })
+
+  afterEach(() => {
+    act(() => renderer?.unmount())
+    renderer = null
+    respond = null
+  })
+
+  it('keeps the marker for a permission choice, which never submits the composer', async () => {
+    const sendRequest = vi.fn().mockResolvedValue({
+      ok: true,
+      result: { send: { handle: 'terminal', accepted: true, bytesWritten: 1 } }
+    })
+    function Harness(): null {
+      respond = useMobileNativeChatPermissionSend({
+        client: { sendRequest } as unknown as RpcClient,
+        enabled: true,
+        handleRef: { current: 'terminal' },
+        deviceTokenRef: { current: null },
+        onSendError: vi.fn()
+      })
+      return null
+    }
+    act(() => {
+      renderer = create(createElement(Harness))
+    })
+    markMobileNativeChatInputStale('terminal')
+
+    await act(async () => {
+      await expect(respond?.('1')).resolves.toBe(true)
+    })
+    // A choice is a bare key for a live overlay that swallows a clear while the
+    // host still acks it, so healing here would burn the marker and leave the
+    // paste to corrupt the next real message. Only the choice may go.
+    expect(sendRequest).toHaveBeenCalledTimes(1)
+    expect(sendRequest.mock.calls[0]?.[1]).toMatchObject({ text: '1', enter: false })
+    expect(isMobileNativeChatInputStale('terminal')).toBe(true)
   })
 })
--- a/mobile/src/session/mobile-native-chat-permission-send.ts
+++ b/mobile/src/session/mobile-native-chat-permission-send.ts
@@ -36,6 +36,9 @@ export function useMobileNativeChatPermissionSend(args: {
         args.onSendError('Response not sent (disconnected)')
         return false
       }
+      // No stale-input heal here (unlike the text/ask sends): a choice is an
+      // `enter: false` key for an active overlay that swallows the clear, so it
+      // would consume the marker still protecting the next real message.
       const outcome = await sendMobileNativeChatPermissionResponse({
         client: args.client,
         terminal,
--- a/mobile/src/session/mobile-native-chat-stale-input.test.ts
+++ b/mobile/src/session/mobile-native-chat-stale-input.test.ts
@@ -1,0 +1,78 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import type { RpcClient } from '../transport/rpc-client'
+import {
+  clearMobileNativeChatInputStale,
+  healMobileNativeChatStaleInput,
+  isMobileNativeChatInputStale,
+  markMobileNativeChatInputStale,
+  resetMobileNativeChatStaleInputForTests
+} from './mobile-native-chat-stale-input'
+
+function sendResult(accepted: boolean) {
+  return {
+    id: 'send',
+    ok: true as const,
+    result: { send: { accepted } },
+    _meta: { runtimeId: 'runtime' }
+  }
+}
+
+function makeClient(accepted = true): Pick<RpcClient, 'sendRequest'> {
+  return { sendRequest: vi.fn().mockResolvedValue(sendResult(accepted)) }
+}
+
+describe('mobile native chat stale input markers', () => {
+  beforeEach(() => {
+    resetMobileNativeChatStaleInputForTests()
+  })
+
+  it('tracks each terminal independently', () => {
+    markMobileNativeChatInputStale('term-1')
+    expect(isMobileNativeChatInputStale('term-1')).toBe(true)
+    expect(isMobileNativeChatInputStale('term-2')).toBe(false)
+    clearMobileNativeChatInputStale('term-1')
+    expect(isMobileNativeChatInputStale('term-1')).toBe(false)
+  })
+
+  it('writes nothing when the terminal is not marked', async () => {
+    const client = makeClient()
+    await expect(
+      healMobileNativeChatStaleInput({ client, terminal: 'term-1', deviceToken: null })
+    ).resolves.toBe(true)
+    expect(client.sendRequest).not.toHaveBeenCalled()
+  })
+
+  it('clears the line and consumes the marker', async () => {
+    const client = makeClient()
+    markMobileNativeChatInputStale('term-1')
+    await expect(
+      healMobileNativeChatStaleInput({ client, terminal: 'term-1', deviceToken: 'device' })
+    ).resolves.toBe(true)
+    expect(client.sendRequest).toHaveBeenCalledTimes(1)
+    expect(vi.mocked(client.sendRequest).mock.calls[0]?.[1]).toMatchObject({
+      terminal: 'term-1',
+      text: '\x15',
+      enter: false,
+      client: { id: 'device', type: 'mobile' }
+    })
+    expect(isMobileNativeChatInputStale('term-1')).toBe(false)
+  })
+
+  it('keeps the marker when the host rejects the clear', async () => {
+    const client = makeClient(false)
+    markMobileNativeChatInputStale('term-1')
+    await expect(
+      healMobileNativeChatStaleInput({ client, terminal: 'term-1', deviceToken: null })
+    ).resolves.toBe(false)
+    expect(isMobileNativeChatInputStale('term-1')).toBe(true)
+  })
+
+  it('keeps the marker when the clear throws', async () => {
+    const client = { sendRequest: vi.fn().mockRejectedValue(new Error('offline')) }
+    markMobileNativeChatInputStale('term-1')
+    await expect(
+      healMobileNativeChatStaleInput({ client, terminal: 'term-1', deviceToken: null })
+    ).resolves.toBe(false)
+    expect(isMobileNativeChatInputStale('term-1')).toBe(true)
+  })
+})
--- a/mobile/src/session/mobile-native-chat-stale-input.ts
+++ b/mobile/src/session/mobile-native-chat-stale-input.ts
@@ -1,0 +1,63 @@
+import type { RpcClient } from '../transport/rpc-client'
+import { pasteMobileNativeChatImagePaths } from './mobile-native-chat-image-send'
+
+// The condition tracked here — a bracketed image paste left sitting on the agent's
+// unsubmitted input line — lives on the HOST terminal, so it outlives any one
+// session screen. Keyed by terminal handle at module scope: React state died with
+// the screen and let the orphaned paste glue onto the next message (#10228).
+const staleInputTerminals = new Set<string>()
+
+export function markMobileNativeChatInputStale(terminal: string): void {
+  staleInputTerminals.add(terminal)
+}
+
+export function isMobileNativeChatInputStale(terminal: string): boolean {
+  return staleInputTerminals.has(terminal)
+}
+
+export function clearMobileNativeChatInputStale(terminal: string): void {
+  staleInputTerminals.delete(terminal)
+}
+
+/** Test-only: module scope outlives a single test's hooks. */
+export function resetMobileNativeChatStaleInputForTests(): void {
+  staleInputTerminals.clear()
+}
+
+/** Clears a marked terminal's unsubmitted input line before a write that could
+ *  submit it, consuming the marker only once the host accepts the clear.
+ *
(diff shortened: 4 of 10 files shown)


brennanb2025:
Re the two CodeRabbit pre-merge check warnings (there were **no actionable inline comments** — 0 findings across all 9 files at `bb46ca226a`):

**Description check — actioned.** The body now follows `.github/PULL_REQUEST_TEMPLATE.md` (Summary / Screenshots / Testing / AI Review Report / Security Audit / Notes). All the original detail is preserved, just reorganized under the required headings.

**Docstring Coverage 30.77% vs 80% — declining, with reasoning.** This one conflicts with the repo's own documented style rule in `AGENTS.md`:

> ## Concise/Brief Non-obvious comments ONLY
> * DO NOT: be verbose, explain the obvious, walk through the code ("WHY not HOW")
> * DO: BE CONCISE. 1 LINE if possible

Raising coverage to 80% here would mean adding docstrings to `markMobileNativeChatInputStale`, `isMobileNativeChatInputStale`, and `clearMobileNativeChatInputStale` — three one-line `Set` add/has/delete wrappers whose names already say exactly what they do. That is precisely the "explain the obvious" the guide prohibits.

The commenting effort in this PR is deliberately spent where the reasoning is non-obvious rather than spread evenly for a coverage number:
- the module header, on why the marker must outlive the screen (it tracks host-side state);
- `healMobileNativeChatStaleInput`'s contract, including *why* dialog-control paths must not call it;
- the two in-code notes at the permission and Escape sites, so the deliberate skips do not read as oversights.

Happy to add them if a maintainer prefers the threshold over the style rule — flagging the conflict rather than silently ignoring the check.

brennanb2025:
## Review — deviation adjudicated (upheld), one HIGH gap found in the opposite direction

Reviewed with three independent subagents plus a first-hand pass. Pushed `86f6c81982`.

### The deliberate scope call is CORRECT — do not revert it

The filed spec demanded the permission-response and Escape-cancel paths heal too. Declining that was right, and for a **stronger** reason than the body gives:

- Both really are no-commit writes. `mobile-native-chat-permission-send.ts:20` and `use-mobile-native-chat-controller.ts:193` pass `enter: false`, and the host coerces regardless — `src/main/runtime/rpc/methods/terminal.ts:1345`: `enter: params.enter === true`. The only bytes produced are `1`/`y`/`n`/`a`/ESC (`mobile-native-chat-permission.ts:53-56, :130, :170-176`); nothing appends Return at any layer.
- **The consume-without-clear hazard is real and does not depend on the overlay premise at all.** `accepted` is write-level, never semantic: `orca-runtime.ts:13635` and `:13655` return `accepted: true` unconditionally once the write resolves — no screen readback. So `healMobileNativeChatStaleInput` consumes the marker (`mobile-native-chat-stale-input.ts:61`) on proof that a byte was *written*, not that the line was *cleared*. Any heal on a path whose Ctrl+U doesn't reach the composer burns a one-shot marker. That is sufficient on its own.

### HIGH — the same reasoning condemned a heal the PR *added* (fixed in `86f6c81982`)

`use-mobile-native-chat-answer-send.ts` healed **every** ask answer, justified by "Both answer shapes commit with Enter." That is false for the dominant selector shapes:

- Claude single-question single-select → `[{raw:'2'}]`, no Enter (`src/shared/native-chat-ask.ts:211-213`; the only trailing `ASK_ENTER` is gated at `:220-223` on multi-question or multi-select).
- Codex single-select → `[{raw:'1'}]` (`:259-262`).
- Every stepping group is written `enter: false` (`use-mobile-native-chat-answer-send.ts:187`).

So a bare option digit cannot submit the composer, yet the heal fired anyway — against the overlay where the premise is *strongest* (the ask card is built from the structured `interactivePrompt`, and the user is tapping it live). Marker burned, paste survives, next real message ships `<image path><text>`. This was **newly reachable in this PR**: on main the ask path never touched the marker.

Fix: scope the heal to the pasted-label shape, which does commit the composer. **Desktop splits it identically** — `use-native-chat-interactive-send.ts:132-140` routes only the non-stepping answer through `sendNativeChatMessage` (which pre-clears at `native-chat-runtime-send.ts:73`) and never pre-clears `sendNativeChatAskAnswer` (`:205-258`).

Proof the new test catches it: reverting only that source file to `bb46ca226a` fails with `expected "vi.fn()" to be called 1 times, but got 2 times`.

Also in `86f6c81982`: the three deliberate skips (selector answer, permission choice, Escape cancel) are now **test-pinned**, so the PR's central design argument is an invariant instead of a prose comment a future "consistency" refactor can silently delete. And the failed-heal toast now takes the generation guard every other error surface in `answerAsk` already uses (`:160`), so a superseded answer no longer raises a spurious "Answer not sent".

### Verified, and reproduced rather than trusted

- **Author's test claim HOLDS.** Reverting only the 4 consumer source files while keeping the new module and tests yields **5 genuine `AssertionError`s** — not module-resolution noise. (Deleting the new module instead gives `Cannot find module`, which proves nothing; both variants were run to separate them.)
- Full mobile suite **339 files / 2465 passed, 2 skipped**. Typecheck clean, oxlint + react-doctor + oxfmt clean on touched files.
- **No unbounded accumulator.** Handles are `term_<uuidv4>` minted per `ptyId` and gated on `runtimeId` (`orca-runtime.ts:27042-27057`, `:26594`), so they are never recycled — an unreaped marker is permanently *inert*, not armed, and cannot fire a spurious Ctrl+U on a fresh terminal. ~120 bytes per failed image send. No teardown needed.
- **Restart hole is narrower than stated:** a *desktop* restart invalidates the handle and leaves a clean composer, so the only gap is a mobile cold-start against a still-live desktop PTY. Persisting it would put an AsyncStorage `await` in front of every send; not worth it.
- **No text loss** on a failed heal — the composer only clears on accept (`MobileNativeChatComposer.tsx:131-133`).
- No double clear: the heal sits in the text-only branch (`use-mobile-native-chat-image-attachments.ts:222-234`) and the attachments branch consumes the marker before `baseSend` (`:262`).

### CodeRabbit

Verified directly: `pulls/10480/comments` is **empty** — zero inline findings, as reported. 4 pre-merge checks pass, 1 fails (docstring coverage 30.77% vs 80%). **The declination is correct**: the uncovered symbols are three one-line `Set` wrappers whose names state exactly what they do, and `AGENTS.md` ("Concise/Brief Non-obvious comments ONLY … DO NOT explain the obvious") prohibits documenting them. Nothing was missed. CodeRabbit's own walkthrough did not catch the over-heal above.

### Two corrections to the body

- The desktop citation should point at `use-native-chat-interactive-send.ts:74-82, :155-158` for the approval/ESC scoping; `native-chat-runtime-send.ts:33-37` only carves out *verified option commands* (model-switch PTY observation). And "desktop scopes its Ctrl+U the same way" understates it — desktop scopes it **more narrowly**, which is what makes it evidence against the ask-path heal.
- Test counts are now 2465 (+3).

### Still not device-verified — unchanged

No phone, simulator, or paired relay was used by me either. That an agent TUI's live overlay actually *swallows* Ctrl+U is **reasoned from desktop's scoping, not observed**, and no unit test can settle it — a test pins "we don't send the clear," not "the overlay would have eaten it." My fix is the safe side of that uncertainty either way: if the overlay instead passes control keys through, skipping the heal merely costs one redundant Ctrl+U on the next text send; if it swallows them, skipping it is what prevents the corruption. Residual, honestly stated: for the multi-question/multi-select shapes that *do* end in a raw `\r`, a stale card whose overlay already closed can still submit the composer — desktop accepts that same risk, and the marker now survives to clean the line before the next real message.

Deliberately **not** widened into `use-mobile-diff-review-send-actions.ts:77-80` (`enter: true` to a user-chosen terminal); correctly flagged as follow-up in the body. Audited the other `enter: true` writers — `pr-ai-triage-launch.ts:34` and `ai-vault-resume-launch.ts:239` create their terminal immediately before sending, so no marker can exist.

**LAND-READY** with `86f6c81982`. One manual pass still worth it before trusting the premise: attach an image, send until it reports unconfirmed, back out of the session and return, then send a plain message and confirm no image rides along.

PR #6975, opened 2026-07-01
Fix Windows ConPTY OSC color reply leak

brennanb2025:
## Summary\n- suppress OSC 10/11 color-query replies for local native Windows ConPTY panes\n- keep OSC color replies enabled for POSIX-like paths such as WSL, SSH, and remote runtimes\n- add regression coverage for startup spawn metadata and hidden Codex startup output on Windows ConPTY\n\n## Root Cause\nThe startup color-query work added replies that are written back as PTY input. Native Windows ConPTY can treat those bytes like keyboard input, swallowing ESC and echoing the printable remainder into Codex's prompt (for example, `]10;rgb:...`).\n\n## Tests\n- pnpm test src/renderer/src/components/terminal-pane/pty-connection.test.ts\n- pnpm test src/renderer/src/components/terminal-pane/terminal-capability-replies.test.ts\n- pnpm run typecheck:web

Made with [Orca](https://github.com/stablyai/orca) 🐋

---

<a href="https://stagereview.app/stablyai/orca/pull/6975">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

PR #10492, opened 2026-07-25
fix(terminal): answer color queries bundled ConPTY forwards

yoke233:
## Summary

On Windows, an agent's OSC 10/11 startup color probe often goes unanswered, so the agent falls back to its own default and paints a dark TUI inside a light pane. Codex reproduces this reliably: open a terminal, wait past the 5s startup window, then type `codex` — the composer renders dark regardless of the Orca theme.

`PtyStartupIngress` consumes **every** OSC 10/11 query on a `windows-conpty` owner and replies only while its startup window is open. That rule ("ConPTY cannot safely transfer color-query authority to a downstream view", 2026-07-19 design) was measured against the **system** ConPTY. Every Windows spawn now passes `useConptyDll: true` (`pty-subprocess.ts`, `local-pty-utils.ts`), so the bundled OpenConsole owns the pseudoconsole instead — and the two backends behave in opposite ways.

Measured with a node-pty harness on Windows 11, node-pty 1.1.0, bundled ConPTY 1.23.251008001:

| backend | forwards `OSC 10/11 ?` to the terminal | outer reply reaches the client | cooked echo of the reply |
| --- | --- | --- | --- |
| system ConPTY (`useConptyDll: false`) | no | no | n/a |
| bundled ConPTY (`useConptyDll: true`) | **yes** | **yes, byte-for-byte** | **none observed** |

So on the backend Orca actually uses, consuming a query the transaction cannot answer silences the only responder the program can reach. Codex then falls back to `GetConsoleScreenBufferInfoEx` (`codex-rs/tui/src/terminal_probe.rs`), reads the pseudoconsole palette — `bgIndex=0`, `#0c0c0c` — and renders dark. Measured inside a bundled-ConPTY session:

```
BEFORE attr=0x0007 bgIndex=0 bgCOLORREF=0x0c0c0c rgb=#0c0c0c
```

This triggers whenever the query lands after the startup deadline or after both slots were already answered once, which matches the reported "mostly fine, sometimes dark" behavior.

Startup query authority now transfers identically on every backend: a query this transaction answers is consumed, a query it cannot answer is released to the normal delivered/hidden responder (visible xterm parser handler, or the hidden-startup responder for agent panes). The measured ConPTY echo projection stays gated on `windows-conpty`, so a host that still reaches the system ConPTY keeps its cooked-echo protection.

## Screenshots

No visual change to Orca's own UI. The user-visible effect is inside the agent TUI (Codex composer following the pane theme instead of rendering dark); no before/after capture of that is included — see Notes.

## Testing

- [ ] `pnpm lint` — oxlint, switch-exhaustiveness, styled-scrollbars, reliability gates (48), and the max-lines ratchet all pass. The run then stops at `verify:skill-bundle-manifest` ("Generated skill artifacts are stale"), which reproduces on a clean `origin/main` checkout on this machine, so it is a pre-existing local/baseline condition rather than something this PR introduces. Nothing here touches `resources/skills`.
- [x] `pnpm typecheck` — node and web projects clean.
- [ ] `pnpm test` — ran the affected suites rather than the full run: `pty-startup-ingress`, `session`, `terminal-host-pty-owner-backend`, `local-pty-provider`, `relay/pty-handler`. All pass except 15 pre-existing `relay/pty-handler` failures that also fail on `main` on this machine (kill/shutdown timing, unrelated to color).
- [x] `pnpm build` — `build:win` produced a signed installer; installed and launched it on Windows 11.

Test detail:

- `pty-startup-ingress.test.ts`: three ConPTY contract tests rewritten to the new rule, plus regressions for release-after-close, release-after-deadline, and release-of-an-already-answered slot.
- `session.test.ts`, `local-pty-provider.test.ts`, `relay/pty-handler.test.ts`: same contract at the daemon, local provider, and relay owners.
- `terminal-host-pty-owner-backend.test.ts`: the stale-WSL-metadata contract now distinguishes backends by the cooked-echo projection, the only remaining difference between them.
- `tests/e2e/terminal-osc-color-queries.spec.ts`: new Electron e2e where a real PTY emits the query after the startup window closes. Per review feedback it is now gated on a deterministic post-startup condition — it waits past the startup deadline before emitting the query, and its primary assertion is that the pane's own OSC 11 parser handler saw the query. That distinguishes the responders: the startup transaction consumes what it answers, so a query reaching the pane handler proves it was released downstream. The reply assertion is kept as a secondary check. Verified by reverting only the source fix and rerunning: the spec fails 4/4 runs without it.

## AI Review Report

Risks checked and what came of them:

- **Is there still a responder after release?** Traced all three cases: visible pane answers via the xterm parser handlers (`terminal-capability-replies.ts`), hidden agent panes answer via `writeHiddenStartupRendererQueries` (`pty-connection.ts`, explicitly built for Codex's 100 ms probe budget), hidden non-agent panes are not skipped by `shouldSkipHiddenRendererOutput` so the data reaches xterm normally. No gap found.
- **Double answers?** Each query still has exactly one responder — the transaction consumes only what it answered; everything else is released once.
- **Replay safety.** Released queries now appear in snapshots. Replay paths already return early (`isReplaying()`), so a replayed query still produces no provider write.
- **Lossless drains.** `expire`, `close-query`, and `snapshot` now release a held partial candidate on ConPTY too, which is stricter than the previous behavior (it held bytes until teardown).
- **Cross-platform.** The change removes a Windows-only special case, so POSIX and WSL behavior is unchanged; the only remaining backend-specific code is the echo projection, still gated on `windows-conpty`. The new e2e uses a BEL-terminated query so its single command line survives both PowerShell and POSIX shell quoting. No shortcuts, labels, or path handling touched.
- **Flaky neighbour, measured rather than assumed:** the pre-existing spec in that file (`answers OSC foreground and background color queries...`) intermittently fails when the two specs run in the same batch, because its strict `toEqual` on the whole write log admits no extra writes — observed failures picked up unrelated `CSI c` / `CSI I` replies once and two empty writes another time. I suspected this change had raised the rate (more queries now reach xterm and get answered), so I measured it: reverting only the source fix and running the file three times, the legacy spec still failed 1/3, versus 2/5 with the fix. No discrimination at that sample size, so it is pre-existing flakiness, not something this PR introduces. Left untouched as out of scope; worth a separate fix (compare only that PTY's OSC color replies rather than the entire write log).

## Security Audit

- No new input parsing: the OSC grammar, bounds (`MAX_QUERY_CANDIDATE_CHARS`), and reply construction are untouched. The change only decides whether an already-parsed span is consumed or forwarded.
- No command execution, path handling, auth, secrets, or dependency changes. The new e2e runs `node -e` inside the test terminal only.
- IPC surface unchanged — same emission contract (`data`, `rawStartSeq`, `rawEndSeq`, `transformed`), same raw-sequence accounting.
- Widened data path considered: bytes previously dropped now reach the model, snapshots, and views. They are the program's own color query, already subject to the existing OSC parser, and replay remains reply-silent.

## Notes

- Residual risk: on a host that still reaches the **system** ConPTY, a downstream reply's cooked echo is no longer suppressed, because only source-written replies register an expected projection. Accepted because the dark-on-light failure is deterministic on the bundled backend while that echo is not reproducible there. If any spawn path can still land on the system ConPTY, that would be worth naming explicitly.
- Not verified: a visual before/after of Codex rendering inside Orca. The evidence covers the PTY-level mechanism and the reply path, not the agent's rendering.
- The 2026-07-20 amendment in the design doc had its query-consumption sentences struck through and pointed at the new amendment, so the plan states one behavior (per review feedback).

- Related PR: #6975 addresses the same OSC 10/11 path but is not a duplicate. It suppresses downstream color replies for native Windows ConPTY to prevent reply leakage. This PR targets the later PTY-owner startup-ingress architecture and the bundled ConPTY backend Orca actually enables, which measurements show forwards both the query and the outer reply without cooked echo. On that backend, suppressing or retaining an unanswered query removes the only responder Codex can reach, so this change releases only queries the startup transaction cannot answer while retaining the existing Windows echo projection for replies it does answer.

## ELI5

On Windows, agents probing terminal colors at startup often got no reply and rendered a dark TUI in a light pane. Orca answers those color queries when ConPTY forwards them so themes match.

### DIFF
--- a/docs/reference/plans/2026-07-19-windows-conpty-startup-query-and-focus-authority.md
+++ b/docs/reference/plans/2026-07-19-windows-conpty-startup-query-and-focus-authority.md
@@ -34,16 +34,19 @@ backend.
 
 ## 2026-07-20 Ownership Amendment
 
+> Its query-consumption rules are superseded by the 2026-07-25 amendment above. Its backend
+> classification rule still stands.
+
 Fresh paired-runtime evidence showed that the display/controller platform is not a safe proxy for
 the PTY backend. A macOS client can attach to a native ConPTY owned by a paired Windows runtime, so
 renderer-side local/SSH/remote heuristics can transfer OSC 10/11 authority to the wrong responder.
 
 The PTY-owning process now classifies the backend from its own platform and the shell that actually
-won spawn: `windows-conpty`, `windows-wsl`, or `posix-pty`. Native ConPTY consumes complete OSC 10/11
-queries before model, replay, or view delivery. During the bounded startup window it replies once
-when validated theme colors are available; after the deadline, or without colors, it consumes the
-query without replying. A consuming-view handshake does not transfer this authority. Split query
-candidates remain bounded and private across authority-close, expiry, and snapshot barriers; a
+won spawn: `windows-conpty`, `windows-wsl`, or `posix-pty`. ~~Native ConPTY consumes complete OSC
+10/11 queries before model, replay, or view delivery. During the bounded startup window it replies
+once when validated theme colors are available; after the deadline, or without colors, it consumes
+the query without replying. A consuming-view handshake does not transfer this authority. Split query
+candidates remain bounded and private across authority-close, expiry, and snapshot barriers~~; a
 candidate that proves malformed is released unchanged.
 
 WSL and POSIX PTYs continue to transfer authority to the normal visible/hidden responder after the
--- a/docs/reference/plans/2026-07-19-windows-conpty-startup-query-and-focus-authority.md
+++ b/docs/reference/plans/2026-07-19-windows-conpty-startup-query-and-focus-authority.md
@@ -2,7 +2,35 @@
 
 Date: 2026-07-19
 
-Status: Implemented with the 2026-07-20 ownership amendment below
+Status: Implemented with the 2026-07-20 and 2026-07-25 amendments below
+
+## 2026-07-25 Bundled ConPTY Amendment
+
+The "native ConPTY consumes OSC 10/11 for the life of the PTY" rule was measured against the
+**system** ConPTY. Every Windows spawn now passes `useConptyDll: true`
+([`pty-subprocess.ts`](../../../src/main/daemon/pty-subprocess.ts),
+[`local-pty-utils.ts`](../../../src/main/providers/local-pty-utils.ts)), so the bundled
+OpenConsole binary owns the pseudoconsole instead. A node-pty harness on both backends shows they
+behave in opposite ways:
+
+| backend | forwards `OSC 10/11 ?` to the terminal | reply reaches the client | cooked echo of the reply |
+| --- | --- | --- | --- |
+| system ConPTY (`useConptyDll: false`) | no | no | n/a |
+| bundled ConPTY (`useConptyDll: true`) | yes | yes, byte-for-byte | none observed |
+
+Consuming an unanswerable query on the bundled backend therefore silences the only responder the
+agent can reach. Codex then falls back to `GetConsoleScreenBufferInfoEx`
+(`codex-rs/tui/src/terminal_probe.rs`), reads the pseudoconsole palette — `bgIndex=0`,
+`#0c0c0c` — and renders a dark theme inside a light pane. This reproduced whenever the query
+arrived after the 5 s startup deadline or after both slots had already been answered once.
+
+The startup transaction now transfers OSC 10/11 authority identically on every backend: a query it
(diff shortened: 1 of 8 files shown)


Jinwoo-H:
Thanks @yoke233 for the ConPTY bundled OSC query fix and Windows analysis. We are taking ownership through issue #14799, which tracks the missing OSC 10/11 response and the related Windows path. We will reimplement on current main with the required cross-platform checks; no contributor action is requested. Closing this superseded branch while preserving your credit.

PR #10494, opened 2026-07-25
fix(codex): persist plugin registrations across managed-home remirrors

### LINKED ISSUE #10489: [Bug]: Codex plugin installs disappear after Orca restarts while cache remains
## Summary

Installing a Codex plugin from an Orca-launched Codex terminal succeeds for the current session, but the marketplace and plugin registration disappear after Orca/Codex restarts. The plugin cache remains, so users get stuck in a repeatable install → restart → missing registration loop.

This reproduces with an explicitly managed Codex account, so it is distinct from the system-default `CODEX_HOME` routing problem in #8612.

## Environment

- Orca: 1.4.155
- Codex CLI: 0.145.0
- OS: macOS
- Account mode: Orca-managed Codex account with a per-account `CODEX_HOME`
- Reproduced plugin: Ponytail 4.8.4 (`DietrichGebert/ponytail`)

## Reproduction

1. Launch Codex from Orca using a managed Codex account.
2. In that terminal, run:

   ```sh
   codex plugin marketplace add DietrichGebert/ponytail
   codex plugin add ponytail@ponytail
   codex plugin list --marketplace ponytail
   ```

3. Confirm `ponytail@ponytail` is `installed, enabled`.
4. Quit/relaunch Orca or start after the managed-home config is remirrored.
5. Run `codex plugin list` again.

## Actual behavior

- `[marketplaces.ponytail]` and `[plugins."ponytail@ponytail"]` are gone from the managed account's `config.toml`.
- The plugin is absent from `codex plugin list`.
- The cloned marketplace may remain under the managed home's `.tmp/marketplaces/`.
- Plugin cache/data remain under `~/.codex/plugins` because the managed account's `plugins` directory points there.
- Reinstalling fixes only the current session; the loop repeats after restart.

## Expected behavior

A successful `codex plugin marketplace add` / `codex plugin add` performed in an Orca terminal should persist across Orca and Codex restarts.

## Root cause

Orca injects the managed account home as `CODEX_HOME`, so the Codex CLI correctly writes plugin registrations into that runtime `config.toml`.

On the next config mirror, Orca rebuilds ordinary runtime configuration from the system `~/.codex/config.toml`. Runtime preservation is limited to project and hook-trust sections, while reverse promotion only covers an explicit settings allowlist. Marketplace and plugin tables are neither preserved nor promoted, so they are dropped even though their shared cache survives.

This is the same one-way managed-home config family addressed for top-level settings in #7960 and `[tui]` settings in #9475, but plugin registration tables are still outside that flow.

## Confirmed workaround

Install against the real system Codex home instead of the injected managed home:

```sh
env -u CODEX_HOME codex plugin marketplace add DietrichGebert/ponytail
env -u CODEX_HOME codex plugin add ponytail@ponytail
```

Once `[marketplaces.ponytail]` and `[plugins."ponytail@ponytail"]` exist in `~/.codex/config.toml`, Orca mirrors them into the managed home and the plugin survives restart.

This workaround is shell/platform-specific and bypasses the environment Orca intentionally provides, so users should not need to know or do this.

## Suggested fix

Treat Codex plugin registration mutations as durable user configuration. A fix should preserve the existing three-way/source-of-truth behavior and cover both tables:

- `[marketplaces.<name>]`
- `[plugins."<plugin>@<marketplace>"]`

Possible implementation directions:

1. Extend the existing baseline/promotion mechanism with table-aware promotion for plugin registrations, including safe conflict handling; or
2. Introduce a dedicated plugin-registry reconciliation path between the managed account config and `~/.codex/config.toml`.

Blindly preserving the runtime tables is less desirable because outside edits to `~/.codex` should still have a defined conflict rule.

## Regression coverage

- marketplace + plugin added in managed `CODEX_HOME` survive the next mirror
- cache-only state is not treated as an installed plugin
- outside system-config edits win on conflict
- repeated mirrors are byte-stable
- quoted plugin table names such as `[plugins."name@marketplace"]` are handled
- removal/disable semantics are defined and tested
- managed-account and system-default launch paths remain isolated

## Related

- #8612 — system-default Codex state redirected away from `~/.codex`
- #7960 — top-level setting promotion
- #9168 / #9475 — nested `[tui]` setting promotion

Bongseop-Kim:
## Summary

- Persist Codex marketplace and plugin registrations made inside an Orca-managed `CODEX_HOME` across config remirrors.
- Root cause: the existing baseline-based promotion tracked top-level and `[tui]` settings, but not the dynamic `[marketplaces.<name>]` and `[plugins."<plugin>@<marketplace>"]` tables, so the next system-config mirror replaced them.
- Extend the same baseline/reconciliation path to generalized marketplace and plugin table blocks, preserving quoted headers, comments, CRLF, table order, and byte stability.
- Treat registration removal as a managed-home change and never infer installation from cache contents.

### Reconciliation policy

- If only the managed runtime changed from a known baseline, promote its add, edit, disable, or removal to the system config.
- If the system config changed the same registration, keep the system value, matching the existing three-way policy.
- For legacy baselines without registration ancestry, use the existing anchored-conflict behavior rather than guessing a winner.
- Do not unconditionally preserve runtime plugin tables, so external removals cannot be resurrected.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Additional coverage:

- 9 related Codex config/mirror/account test files: 300 tests passed.
- Full suite: 3,396 test files and 36,011 tests passed; 7 files and 60 tests skipped.
- `oxfmt --check` passed for all changed files.
- Sandboxed `HOME`/`CODEX_HOME` regression with Codex CLI 0.145.0: marketplace add, plugin add/list, repeated real mirror byte stability, plugin/marketplace removal, and no resurrection after remirror.

## AI Review Report

- Reviewed the real mirror call path and all callers before changing the shared promotion layer.
- Checked TOML parsing and rewriting risks: quoted/literal headers, comments, CRLF, ordering, duplicate tables, legacy baselines, concurrent system edits, deletion/disable behavior, and cache-only state.
- Checked macOS, Linux, Windows, WSL, and SSH behavior. The change uses existing Node path/file primitives and explicit runtime/system home paths; it adds no shortcuts, labels, shell commands, or Electron platform branches.
- Checked folder workspaces and supported Git providers; this is isolated to Codex config mirroring and adds no git-worktree or GitHub-specific assumptions.
- Performance remains linear in the small config file per mirror, with no new polling, dependencies, or cache scans.

## Security Audit

- No new command execution, IPC, network access, dependencies, auth handling, or secret logging.
- Config writes continue through the existing atomic/symlink-aware promotion path and existing explicit home boundaries.
- Baseline registration payloads are validated as strings before use, and cache directories are never trusted as installation state.
- No follow-up security work identified.

## Notes

- This intentionally does not change system-default `CODEX_HOME` routing or authentication isolation; those remain separate from this managed-account persistence fix.
- `pnpm build` exited successfully with the existing bundle warnings and local `/usr/local/bin/orca-dev` symlink permission notice.
- X handle: not provided.

Fixes #10489

## ELI5

Codex marketplace and plugin registrations inside Orca-managed CODEX_HOME were wiped on the next config remirror. Those plugin tables are preserved across remirrors so installed plugins stick.

### DIFF
--- a/config/tsconfig.cli.json
+++ b/config/tsconfig.cli.json
@@ -30,6 +30,7 @@
     "../src/main/codex/codex-hook-identity.ts",
     "../src/main/codex/codex-hook-trust-grant.ts",
     "../src/main/codex/codex-managed-trust-reconciliation.ts",
+    "../src/main/codex/codex-plugin-registration-config.ts",
     "../src/main/codex/codex-process-exit-deadline.ts",
     "../src/main/codex/codex-trust-config-rollback.ts",
     "../src/main/codex/codex-trust-grant-telemetry.ts",
--- a/src/main/codex/codex-config-settings-preservation.ts
+++ b/src/main/codex/codex-config-settings-preservation.ts
@@ -1,5 +1,9 @@
 import { removePromotedSettingsFromContent } from './codex-config-settings-removal'
 import { upsertPromotedSettingsInContent } from './codex-config-settings-upsert'
+import {
+  applyCodexPluginRegistrationChanges,
+  isCodexPluginRegistrationKey
+} from './codex-plugin-registration-config'
 
 export function preserveRuntimeConflictValues(
   content: string,
@@ -9,10 +13,14 @@ export function preserveRuntimeConflictValues(
   const keys = new Set<string>()
   for (const [key, raw] of values) {
     const previous = result
-    result =
-      raw === null
-        ? removePromotedSettingsFromContent(result, new Set([key]))
-        : upsertPromotedSettingsInContent(result, new Map([[key, raw]]))
+    if (isCodexPluginRegistrationKey(key)) {
+      result = applyCodexPluginRegistrationChanges(result, new Map([[key, raw]]))
+    } else {
+      result =
+        raw === null
+          ? removePromotedSettingsFromContent(result, new Set([key]))
+          : upsertPromotedSettingsInContent(result, new Map([[key, raw]]))
+    }
     if (result !== previous) {
       keys.add(key)
     }
--- a/src/main/codex/codex-plugin-registration-config.ts
+++ b/src/main/codex/codex-plugin-registration-config.ts
@@ -1,0 +1,187 @@
+import { existsSync } from 'node:fs'
+import { readAgentStateFileSync } from '../agent-state-file-reader'
+import { parseTomlTableHeaderPath } from './config-toml-key-path'
+import { getTomlSections } from './config-toml-runtime-owned-sections'
+import type { CodexSettingsBaseline, CodexSettingsConflict } from './config-settings-baseline'
+import { resolveUntrackedCodexSetting } from './config-settings-conflict-resolution'
+
+const REGISTRATION_ROOTS = new Set(['marketplaces', 'plugins'])
+
+export function readCodexPluginRegistrationTables(configPath: string): Map<string, string> {
+  if (!existsSync(configPath)) {
+    return new Map()
+  }
+  const content = readAgentStateFileSync(configPath)
+  const lines = content.split('\n')
+  const registrations = new Map<string, string>()
+  for (const table of getTomlTableRanges(content)) {
+    const key = getCodexPluginRegistrationKey(table.header)
+    if (key && !registrations.has(key)) {
+      registrations.set(
+        key,
+        normalizeRegistrationBlock(lines.slice(table.start, table.end).join('\n'))
+      )
+    }
+  }
+  return registrations
+}
+
+export function isCodexPluginRegistrationKey(key: string): boolean {
+  const separator = key.indexOf('.')
+  if (!REGISTRATION_ROOTS.has(key.slice(0, separator))) {
+    return false
+  }
+  try {
+    return typeof JSON.parse(key.slice(separator + 1)) === 'string'
+  } catch {
+    return false
+  }
+}
+
+export function applyCodexPluginRegistrationChanges(
+  content: string,
+  changes: ReadonlyMap<string, string | null>
+): string {
+  let result = content
+  for (const [key, block] of changes) {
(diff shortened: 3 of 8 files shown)


nwparker:
Thank you for this — it was genuinely good work, and you are credited as a co-author on the commit that landed.

You reached the central conclusion independently and correctly: a runtime-only registration has to be promoted, but a registration the canonical config *removed* must not be resurrected, and the mirrored baseline is the only thing that can tell those two apart. Your guard — `runtimeRaw !== baselineRaw && systemRaw === baselineRaw` — is the same three-way the merged fix uses for its baseline rows. You also reused the existing conflict-resolution machinery instead of adding a parallel path, which was the better instinct, and your sandboxed `CODEX_HOME` regression against Codex CLI 0.145 covered removal-without-resurrection rather than just the happy path.

One thing your PR does that the merged fix deliberately does not, and it is a place your design is better than ours: for a baseline with no registration ancestry you anchor a *conflict* rather than promoting. That sidesteps a residual risk the merged version accepts (if the baseline is lost at the same moment a canonical removal is pending, the removal can be undone). The merged fix promotes instead, so that a v2→v3 upgrade heals an existing install into `~/.codex` rather than parking it in a conflict — the outcome #10489 asks for. It is a real trade-off and your side of it is defensible; it is written up in the merged PR body so the choice is on the record.

Closing as superseded by #20150, which covers this plus the #11770 half. Your PR compares whole normalised table blocks, so a marketplace metadata refresh reads as a wholesale block change with no field-level `last_updated`/`last_revision` rule to merge it — that was the remaining gap, and the merged policy handles both halves in one baseline-aware pass. If you think something in yours is still uncovered, please say so and I will reopen.

PR #10498, opened 2026-07-25
fix(mobile): exclude proxy fake-ip addresses from pairing QR

### LINKED ISSUE #10404: [Bug]: Mobile pairing can embed a TUN fake-ip address (198.18.x.x) as the desktop endpoint; phone then retries the dead endpoint forever
### Operating system

macOS 15.6.1 (Apple Silicon)

### Orca version

v1.4.153-rc.3

#### Short summary

When a TUN-mode proxy (Clash / mihomo / any fake-ip VPN) is running, its virtual interface (e.g. `utun4` with address `198.18.0.1`) is enumerated as a regular non-internal IPv4 interface and can be selected as the **default mobile pairing address**. The pairing offer then embeds an endpoint like `ws://198.18.x.x:6768` — an address from the RFC 2544 benchmarking range that only exists inside the desktop's own proxy — and the mobile app saves it and retries it forever. The phone-side error shows `无法连接 198.18.9.1:6768` (cannot connect) indefinitely.

#### What happened

1. Desktop runs ClashX Pro in TUN/fake-ip mode. `ifconfig` shows two non-internal IPv4 interfaces: `en0 192.168.50.238` (real LAN) and `utun4 198.18.0.1` (fake-ip TUN).
2. At some point a pairing QR was generated while the TUN interface was picked as the default address (the TUN address changes across proxy restarts; an earlier incarnation was `198.18.9.1`).
3. The phone scanned it and persisted `ws://198.18.9.1:6768` as the desktop's direct endpoint.
4. The phone now retries that unroutable address continuously. Re-scanning a fresh, correct QR did not obviously displace the stale saved endpoint; the recovery that worked was deleting the saved desktop on the phone and pairing again.

#### Why it happens (from the shipped v1.4.153-rc.3 bundle, `out/main/index.js`)

```js
function getNetworkInterfaces() {
  // collects every non-internal IPv4 address...
  return result.sort(
    (a, b) => Number(isTailnetIPv4Address(b.address)) - Number(isTailnetIPv4Address(a.address))
  );
}
function getDefaultPairingAddress() {
  const ifaces = getNetworkInterfaces();
  return ifaces.length > 0 ? ifaces[0].address : null;
}
```

The only prioritization is "Tailscale first". Otherwise the result keeps `os.networkInterfaces()` enumeration order, so a `utun*` fake-ip interface can win over the physical LAN interface. Nothing filters:

* `198.18.0.0/15` (RFC 2544 benchmarking range, used as the fake-ip pool by Clash/mihomo/sing-box by default)
* `utun*`/`tun*`/`tap*` virtual interfaces generally (except Tailscale, which is deliberately preferred)

TUN-mode proxies are extremely common among users in restricted networks — the same population that most depends on Relay/remote pairing.

#### Suggestions

* Exclude `198.18.0.0/15` (and ideally CGNAT `100.64.0.0/10` except recognized tailnets) from pairing-address candidates, and prefer physical interfaces over `utun*` when both exist.
* In the pairing UI's network dropdown, hide or clearly mark virtual proxy interfaces.
* On the mobile side: when the saved direct endpoint is unroutable, fail over to the stored Relay binding instead of retrying the dead LAN endpoint indefinitely, and let a fresh QR scan replace the stale saved endpoint for the same desktop.

#### Related

* stablyai/orca#10273 / stablyai/orca#10401 — other silent-failure modes in the Relay pairing path on this setup. Together they suggest pairing/relay failures need better user-visible diagnostics.

innocarpe:
## Summary
TUN-mode proxies (Clash/mihomo/sing-box) expose `198.18.0.0/15` fake-ip addresses on `utun*`. Those were listed as mobile pairing candidates and could become the **default** QR endpoint — an address only routable inside the desktop proxy. The phone then retries forever.

Closes #10404

## Fix
- Add `isProxyFakeIpIPv4Address()` for RFC 2544 `198.18.0.0/15`
- Filter those addresses out of `getNetworkInterfaces()` so they never appear in the dropdown or default pairing IP
- Tailnet preference (`100.64.0.0/10`) unchanged

## Test plan
- [x] Unit: fake-ip range detection
- [x] Unit: list interfaces + default QR prefer LAN when only fake-ip + LAN present
- [x] Existing mobile IPC tests (IPv6, tailnet ranking) still pass (19 total)
- [ ] Manual: Clash TUN on → Settings → Mobile → pairing address list has no 198.18.x.x

## Out of scope
- Phone-side stale endpoint recovery (also suggested in the issue)

## ELI5

Mobile pairing QR codes could list VPN/proxy fake IPs (198.18.x.x) that only work inside the desktop proxy, so phones never connected. Those addresses are filtered out of pairing candidates.

### DIFF
--- a/src/main/ipc/mobile.test.ts
+++ b/src/main/ipc/mobile.test.ts
@@ -60,10 +60,11 @@ describe('registerMobileHandlers', () => {
   })
 
   it('excludes proxy fake-ip addresses so pairing defaults to LAN (#10404)', async () => {
-    // Why: os.networkInterfaces() order can put Clash TUN (198.18.x.x) first;
-    // without filtering, getDefaultPairingAddress would embed an unroutable IP.
     networkInterfacesMock.mockReturnValue({
-      utun4: [{ family: 'IPv4', internal: false, address: '198.18.9.1' }],
+      utun4: [
+        { family: 'IPv4', internal: false, address: '198.18.0.1' },
+        { family: 'IPv4', internal: false, address: '198.19.255.254' }
+      ],
       en0: [{ family: 'IPv4', internal: false, address: '192.168.50.238' }]
     })
     const createMobilePairingOffer = vi.fn().mockResolvedValue({
--- a/src/main/ipc/mobile.test.ts
+++ b/src/main/ipc/mobile.test.ts
@@ -59,6 +59,33 @@ describe('registerMobileHandlers', () => {
     })
   })
 
+  it('excludes proxy fake-ip addresses so pairing defaults to LAN (#10404)', async () => {
+    // Why: os.networkInterfaces() order can put Clash TUN (198.18.x.x) first;
+    // without filtering, getDefaultPairingAddress would embed an unroutable IP.
+    networkInterfacesMock.mockReturnValue({
+      utun4: [{ family: 'IPv4', internal: false, address: '198.18.9.1' }],
+      en0: [{ family: 'IPv4', internal: false, address: '192.168.50.238' }]
+    })
+    const createMobilePairingOffer = vi.fn().mockResolvedValue({
+      available: true,
+      pairingUrl: 'orca://pair#lan',
+      endpoint: 'ws://192.168.50.238:6768',
+      deviceId: 'mobile-lan',
+      connectionMode: 'automatic'
+    })
+
+    registerMobileHandlers({ createMobilePairingOffer } as never)
+
+    expect(handlers.get('mobile:listNetworkInterfaces')?.()).toEqual({
+      interfaces: [{ name: 'en0', address: '192.168.50.238' }]
+    })
+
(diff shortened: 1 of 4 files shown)


brennanb2025:
Thanks @innocarpe — pushed a small review cleanup to keep the filter local to mobile interface enumeration and cover the exact range boundaries. Typecheck, lint, 46 focused tests, and Electron pairing validation pass.

innocarpe:
Thanks @brennanb2025 — appreciate the review cleanup (local filter + range-boundary tests). Glad typecheck/lint/focused tests and Electron pairing validation are green on your side.

brennanb2025:
Thanks @innocarpe — nice, focused fix! We validated it across macOS, Windows, iOS simulator pairing, and Linux SSH, and it’s now landed. Appreciate the contribution.

PR #10502, opened 2026-07-25
fix(sidebar): keep idle Claude agent rows after relaunch without prompts

### LINKED ISSUE #10398: [Bug]: Idle Claude Code agent row missing from sidebar after quit/relaunch
### Operating system

macOS

### Orca version

1.4.153-rc.3

### short summary
Launching Claude Code and quitting without sending a prompt loses its sidebar agent row on relaunch. The pane and tab title restore correctly; only the "Claude Code · idle" row under the worktree card is gone. Sending any prompt after relaunch recreates it.

### what happened
Started claude code in a terminal tab on worktree. The sidebar showed "Claude Code · idel" immediately, before any prompt. Quit Orca, reopened. Tab title and pane restored; the sidebar row was gone. Sending any prompt recreates it. If a prompt was sent before quitting, the row survives.

### expected
The row is created at launch with no prompt, so restore should recreate it the same way.

### how can we reproduce it?
1. Start Claude Code in a terminal tab on any local worktree.
2. Confirm the sidebar shows Claude Code · idle.
3. Send no prompt. Quit Orca, reopen.
4. Pane and tab title restore; the sidebar row is missing.
5. Send any prompt — the row reappears.

https://github.com/user-attachments/assets/aa74d82a-d484-4deb-87ac-b5449cc84d00

innocarpe:
## Summary
Idle Claude Code sessions that never received a prompt lost their sidebar agent row after quit/relaunch. The tab/pane restored, but the worktree card no longer showed \"Claude Code · idle\" until the next prompt (when hooks re-hydrated status).

Closes #10398

## Root cause
Title-derived sidebar rows required either:
1. The word \"claude\" in the title (Claude token gate), **or**
2. A braille spinner + \`launchAgent\`

Claude's bare idle protocol title is often just \`✳\` / \`✳ task\` with no \"claude\" token, and after cold restore the OSC title can briefly be a shell name while \`launchAgent\` remains \`claude\`. Both paths dropped the row.

## Fix
1. Treat Claude status prefixes (\`✳\`, \`. \`, \`* \`) as Claude identity without requiring the name token
2. When a live PTY only has a shell-like title and the tab owns \`launchAgent\`, restore an idle row from that launch identity

## Test plan
- [x] Bare \`✳\` idle title → Claude idle row
- [x] \`zsh\` title + \`launchAgent: claude\` + live PTY → Claude idle row
- [x] Ordinary non-agent task titles still do not invent rows
- [x] Existing title-derived suite (13 tests)
- [ ] Manual: launch Claude, no prompt, quit/reopen → sidebar row remains idle

## ELI5

Idle Claude sessions without a prompt disappeared from the sidebar after relaunch even though the tab restored. Idle Claude rows are kept so you still see "Claude Code · idle."

### DIFF
--- a/src/renderer/src/components/sidebar/worktree-title-derived-agent-rows.test.ts
+++ b/src/renderer/src/components/sidebar/worktree-title-derived-agent-rows.test.ts
@@ -255,14 +255,51 @@ describe('buildTitleDerivedAgentRows', () => {
     expect(rows).toHaveLength(0)
   })
 
-  it('does not turn generic Codex-launched task titles into Claude Code rows', () => {
+  it('treats Claude idle-prefix titles as Claude even without the word "claude" (#10398)', () => {
+    // Why: bare `✳` / `✳ task` is Claude Code's idle protocol; previously the
+    // Claude token gate dropped those rows, so never-prompted sessions vanished
+    // from the sidebar after relaunch once hooks had nothing to hydrate.
+    const rows = buildWorktreeAgentRows({
+      tabs: [makeTab('tab-1', { launchAgent: 'claude', title: '\u2733' })],
+      entries: [],
+      retained: [],
+      runtimePaneTitlesByTabId: {
+        'tab-1': { 1: '\u2733' }
+      },
+      ptyIdsByTabId: { 'tab-1': ['pty-claude-idle'] },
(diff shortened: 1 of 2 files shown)


innocarpe:
Closing this stale branch as superseded by the broader restart and remote-ownership work in #17614, which is mergeable and covers the current direction for this behavior. No further rebase is needed here.

Issue #10416, opened 2026-07-24
Long PTY injections into TUI agent panes are silently truncated; dispatch reports 'dispatched' anyway


iampaycheck:
**Environment:** Windows 11, packaged Orca, daemon-hosted panes running TUI agents (Claude Code).

## Symptom

Delivering a long prompt into a TUI agent pane — via `orca orchestration dispatch --task … --inject` or a large `orca terminal send` — silently truncates it. The pane receives only a tail fragment of the text.

Observed in a real orchestration run: two freshly-launched Claude Code panes were each injected with a multi-paragraph task prompt. One ended up at an empty prompt; the other received only the fragment `"...artbeat.md IF BLOCKED run: ..."` and began executing the wrong instructions. Meanwhile the dispatch row reported `dispatched` — the state was recorded as delivered when delivery had in fact failed.

## Why this is nasty

- The failure is invisible to the orchestrator: dispatch status says success, and an agent working on a wrong fragment looks identical to an agent working correctly.
- TUI agents (Claude Code at least) drop input during main-thread stalls, so large single writes to the PTY are inherently unreliable on this path.

## Asks

1. Pace/chunk large PTY writes (or use bracketed-paste with flow control) in `--inject` and `terminal send` so long payloads survive delivery into TUI panes.
2. Better: offer a file-based delivery mode for `dispatch` (write the payload to a file in the worktree, inject a one-line pointer prompt) — this is the workaround we now use manually and it is fully reliable.
3. Verify delivery before marking a dispatch `dispatched` (e.g. echo/ack check), so truncation surfaces as an error instead of a silently wrong worker.

AmethystLiang:
looking into it

innocarpe:
## Status update (from #10504)

I opened **#10504** against this report. After maintainer review and independent Windows probes, here is an honest summary of what we know and what shipped.

### What we could **not** show

@nwparker and I both tried to reproduce **silent ConPTY truncation** on real Windows (`node-pty` / ConPTY, large payloads, stalled readers, “fresh agent” cooked-mode cases).

In those probes, even a **single large write** delivered 100% of bytes. So we do **not** currently have evidence that “1 KiB chunks + 16 ms gap” fixes the failure described here — and that pacing has a real cost (roughly ~1.5s per 100 KB of paste).

Because of that, **#10504 no longer claims to close this issue** (`Refs #10416`, not `Closes`).

### What **#10504 still improves** (worth landing on its own)

[...]

PR #10504, opened 2026-07-25
fix(terminal): serialize PTY writes and harden large terminal sends

innocarpe:
## Summary

Hardens large terminal sends without shipping unproven Windows ConPTY pacing:

- **Per-PTY write serialization** so concurrent `sendTerminal` / paste paths do not interleave chunk-wise
- **Host-platform-aware** chunk helper inputs (local vs SSH/WSL) for any future policy
- **Teardown cancellation** of in-flight paced/chunked writes when a PTY exits or is replaced
- **RPC keepalive** so long multi-chunk sends do not trip local socket idle timeouts
- **No 1 KiB / 16 ms Windows pacing** — independent real-Windows ConPTY probes (author + maintainer) did not reproduce silent truncation from #10416; the cost of that pacing was large and unjustified

Refs #10416

## ELI5

Orca keeps long messages for the same terminal in order and cleans up if the terminal closes mid-send. It no longer deliberately slows Windows pastes with tiny pauses, because we could not prove that was fixing a real drop.

## Test plan

- [x] `serializes concurrent terminal sends per PTY` regression (fails on main without the queue)
- [x] `terminal-input` chunk budget is platform-uniform (no Windows-only gap)
- [x] Maintainer re-review of serialize / host / teardown / keepalive paths
- [ ] Manual: large paste on Windows still feels snappy; concurrent sends stay ordered

## Review notes

Following @nwparker’s real-Windows measurements: keep the infrastructure that stands alone; drop pacing until #10416 has a reproducible repro.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/WooseongKim/Projects/OpenSources/orca/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/WooseongKim/Projects/OpenSources/orca/node_modules
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -15611,29 +15611,41 @@ describe('OrcaRuntimeService', () => {
   })
 
   it('writes large terminal input in 16 KiB chunks without inter-chunk delay', async () => {
-    const writes: string[] = []
-    const runtime = new OrcaRuntimeService(store)
-    runtime.setPtyController({
-      spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
-      write: (_ptyId, data) => {
-        writes.push(data)
-        return true
-      },
-      kill: () => true,
-      getForegroundProcess: async () => null
-    })
-    const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+    vi.useFakeTimers()
+    try {
+      const writes: string[] = []
+      const runtime = new OrcaRuntimeService(store)
+      runtime.setPtyController({
+        spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
+        write: (_ptyId, data) => {
+          writes.push(data)
+          return true
+        },
+        kill: () => true,
+        getForegroundProcess: async () => null
+      })
+      const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
 
-    await runtime.sendTerminal(handle, { text: 'x' })
-    expect(writes).toEqual(['x'])
+      await runtime.sendTerminal(handle, { text: 'x' })
+      expect(writes).toEqual(['x'])
+      expect(vi.getTimerCount()).toBe(0)
 
-    writes.length = 0
-    const text = 'y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES * 2)
-    await runtime.sendTerminal(handle, { text })
-    expect(writes).toEqual([
-      'y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES),
-      'y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES)
-    ])
+      writes.length = 0
+      const text = 'y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES * 2)
+      // Why: flush microtasks only — advancing timers would let a reintroduced gap pass this test.
+      const send = runtime.sendTerminal(handle, { text })
+      for (let i = 0; i < 32 && writes.length < 2; i += 1) {
+        await Promise.resolve()
+      }
+      expect(vi.getTimerCount()).toBe(0)
+      expect(writes).toEqual([
+        'y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES),
+        'y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES)
+      ])
+      await send
+    } finally {
+      vi.useRealTimers()
+    }
   })
 
   it('serializes concurrent terminal sends per PTY', async () => {
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -15610,6 +15610,103 @@ describe('OrcaRuntimeService', () => {
     }
   })
 
+  it('paces Windows chunks without delaying single writes', async () => {
+    vi.useFakeTimers()
+    setPlatform('win32')
+    try {
+      const writes: string[] = []
+      const runtime = new OrcaRuntimeService(store)
+      runtime.setPtyController({
+        spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
+        write: (_ptyId, data) => {
+          writes.push(data)
+          return true
+        },
+        kill: () => true,
+        getForegroundProcess: async () => null
+      })
+      const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+
+      await runtime.sendTerminal(handle, { text: 'x' })
+      expect(writes).toEqual(['x'])
+      expect(vi.getTimerCount()).toBe(0)
+
+      const pacedSend = runtime.sendTerminal(handle, {
+        text: 'y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES * 2)
+      })
+      await vi.advanceTimersByTimeAsync(15)
+      expect(writes).toHaveLength(2)
+      await vi.runAllTimersAsync()
+      await pacedSend
+      expect(writes.slice(1).join('')).toBe('y'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES * 2))
+    } finally {
+      vi.useRealTimers()
+    }
+  })
+
+  it('serializes concurrent terminal sends per PTY', async () => {
+    vi.useFakeTimers()
+    setPlatform('win32')
+    try {
+      const writes: string[] = []
+      const runtime = new OrcaRuntimeService(store)
+      runtime.setPtyController({
+        spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
+        write: (_ptyId, data) => {
+          writes.push(data)
+          return true
+        },
+        kill: () => true,
+        getForegroundProcess: async () => null
+      })
+      const { handle } = await runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`)
+      const first = 'a'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES * 2)
+      const second = 'b'.repeat(TERMINAL_INPUT_CHUNK_MAX_BYTES * 2)
+
+      const firstSend = runtime.sendTerminal(handle, { text: first })
+      const secondSend = runtime.sendTerminal(handle, { text: second })
+      await vi.runAllTimersAsync()
+      await Promise.all([firstSend, secondSend])
+
+      expect(writes.join('')).toBe(first + second)
+    } finally {
+      vi.useRealTimers()
+    }
+  })
+
+  it('cancels paced input when its PTY exits', async () => {
+    vi.useFakeTimers()
+    setPlatform('win32')
+    try {
+      const writes: string[] = []
+      const runtime = new OrcaRuntimeService(store)
+      runtime.setPtyController({
+        spawn: vi.fn().mockResolvedValue({ id: 'pty-bg' }),
+        write: (_ptyId, data) => {
+          writes.push(data)
+          return true
+        },
+        kill: () => true,
+        getForegroundProcess: async () => null
+      })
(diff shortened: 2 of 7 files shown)


nwparker:
## Maintainer review: I could not reproduce the premise on real Windows, and the pacing introduces a new failure mode

I picked this up to land it and ran the change against a real Windows 11 box (win32 x64, Node v24.18.0, ConPTY via the repo's own `node-pty`). The measurements do not support the stated mechanism, and they surface a regression that I think blocks the current approach. Details below so the work isn't lost.

### 1. 16 KiB + `setTimeout(0)` did not truncate on real ConPTY

Self-contained `node-pty` probe: spawns a raw-stdin reader under ConPTY (`useConpty: true`), writes a payload under three pacing regimes, and reports bytes actually received by the child.

Regime A — reader busy-blocks its main thread 40 ms every 3 reads (the "TUI stalls on render" theory from #10416):

```
platform=win32 node=v24.18.0 payload=65536B stallMs=40 stallEvery=3 bootMs=0
single write             writes=    1 sent=65536 got=65536 lost=0 OK writeWall=1ms
main 16KiB/0ms           writes=    4 sent=65536 got=65536 lost=0 OK writeWall=1ms
PR 1KiB/16ms             writes=   64 sent=65536 got=65536 lost=0 OK writeWall=1829ms
```

Regime B — child stays in cooked mode and does not read for 6 s (the "freshly launched agent" regime), 256 KiB:

```
platform=win32 node=v24.18.0 payload=262144B stallMs=0 bootMs=6000
single write             writes=    1 sent=262144 got=262144 lost=0 OK writeWall=2ms
main 16KiB/0ms           writes=   16 sent=262144 got=262144 lost=0 OK writeWall=1ms
PR 1KiB/16ms             writes=  256 sent=262144 got=262144 lost=0 OK writeWall=7330ms
```

**Current `main` delivered 100% of bytes in every case.** Even a single unchunked 256 KiB write delivered intact. I could not produce the reported tail-fragment truncation at the ConPTY layer at all.

At 2 MiB with a 60 ms/2-read stall, *all three* regimes failed identically (including this PR's) — that is the reader timing out, not a pacing effect. Pacing does not fix it either.

### 2. The reporter was on daemon-hosted panes, which do not touch ConPTY on this path

#10416 says "daemon-hosted panes". On that path `ptyController.write` →  `DaemonPtyAdapter.write` → `this.client.notify('write', …)` (`src/main/daemon/daemon-pty-adapter.ts:599`), i.e. **one NDJSON socket frame per chunk**, applied daemon-side at `daemon-server.ts:811`. So on the reporter's actual setup, 1 KiB chunking multiplies socket frames 16x rather than pacing a console host. Whatever drops the input there, it is not ConPTY back-pressure — so this fix likely does not address the reported bug.

### 3. Blocking regression: paced sends over ~1.83 MiB exceed the 30 s RPC socket idle timeout on Windows

`orca terminal send` is served by `UnixSocketTransport`, which destroys any socket idle for `RUNTIME_RPC_SOCKET_IDLE_TIMEOUT_MS = 30_000` (`src/main/runtime/rpc/unix-socket-transport.ts:12,119`). Keepalive is **opt-in for long-polls only**, and `terminal.send` is not one — `isLongPollRequest` covers only `terminal.wait`, `orchestration.ask`, and `orchestration.check --wait` (`runtime-rpc.ts:398`). The comment there states the consequence outright: without keepalive "the 30 s socket idle timer tears it down."

With 1 KiB / 16 ms, wall time crosses 30 s at `ceil(b/1024) > 1876` → **~1.83 MiB**. `TERMINAL_INPUT_MAX_BYTES` is 16 MiB, so every Windows payload between ~1.83 MiB and the 16 MiB cap is accepted by validation and then has its socket destroyed mid-write. Measured on the Windows box: 2 MiB paced took **57 785 ms**. That converts a truncation bug into a hard teardown on the exact path the PR targets.

Latency cost, chunks x gap (measured wall time in brackets):

| Payload | main | this PR |
|---|---|---|
| 4 KiB | 1 write, 0 ms | 4 writes, 48 ms |
| 64 KiB | 4 writes, 0 ms | 64 writes, 1.0 s [1829 ms] |
| 256 KiB | 16 writes, 0 ms | 256 writes, 4.1 s [7330 ms] |
| 100 KB | 7 writes, 0 ms | 98 writes, 1.6 s |
| 2 MiB | 128 writes, 0 ms | 2048 writes, 32.7 s [57 785 ms] — **past the 30 s wall** |
| 16 MiB (cap) | 1024 writes, 0 ms | 16 384 writes, **262 s** |

### 4. Concurrent sends to one PTY interleave — pre-existing, but this widens the window 16x

There is no per-PTY write queue on the runtime path (the renderer has one in `pty-input-write-queue.ts`; the runtime/RPC path does not). Two overlapping `sendTerminal` calls to one PTY interleave chunk-by-chunk — I confirmed with a temporary test that logged write order `ababab`. This is **pre-existing on `main`** (which also awaits `setTimeout(0)` between chunks and therefore also yields), so the PR does not introduce it. But it stretches the interleavable window on Windows from ~0 ms to ~1 s per 64 KiB, making a latent race materially more likely to bite.

### What is good here

- Pacing is correctly scoped to `win32`; `getTerminalInputChunkGapMs()` returns 0 elsewhere.
- Commit 9f090cb (skip the timer when gap is 0) is a genuine **improvement over `main`** for macOS/Linux — `main` awaits `setTimeout(0)` between chunks unconditionally; this skips it. That part is worth keeping regardless of what happens to the rest.
- **No per-keystroke impact.** Interactive typing goes through `pty:write` IPC → `writePtyProviderInput` (`src/main/ipc/pty.ts:4812`), which never calls the runtime helper and keeps the 16 KiB default. And any payload under one chunk emits a single chunk, so the gap loop never runs.

### Suggested path

1. Reproduce #10416 against the **daemon** write path (NDJSON frame handling), not ConPTY — that is where the reporter's bytes actually flow.
2. If pacing is still wanted, bound total paced duration or make `terminal.send` a keepalive/long-poll request first, otherwise the 30 s wall turns into a new bug.
3. Consider the issue's own asks 2 and 3 (file-based delivery, delivery ack before marking `dispatched`) — an ack would have surfaced this class of failure instead of silently reporting success.

Verification run: `src/shared/terminal-input.test.ts` 14 passed; `orca-runtime.test.ts -t chunk` 24 passed; `tsc --noEmit -p config/tsconfig.node.json` exit 0; `oxlint` on all three changed files clean. Windows probes executed on real hardware as shown. Not pushing changes — the diff is unchanged from the author's.

Original patch by @innocarpe — the win32 scoping and the zero-gap fast path are sound; it is the underlying premise I could not confirm.

nwparker:
## Could not reproduce the truncation this paces around — recommend holding

I ran a self-contained node-pty probe on a real Windows host (win32, Node v24.18.0, `useConpty: true`): spawn a raw-stdin reader child, write a payload under three pacing strategies, count bytes actually delivered. **Zero truncation in every configuration**, including a single unchunked 1 MiB write into a reader that busy-blocks its main thread:

```
1MB payload, reader stalling 80ms every 4 reads:
single write     writes=   1  sent=1048576  got=1048576  lost=0  OK  writeWall=4ms
main 16KiB/0ms   writes=  64  sent=1048576  got=1048576  lost=0  OK  writeWall=2ms
PR 1KiB/16ms     writes=1024  sent=1048576  got=1048576  lost=0  OK  writeWall=28864ms
```

Also tested an idle reader at 64 KB, a stalled reader at 256 KB, and a child still in cooked mode with raw mode not yet armed (the freshly-launched-agent regime). All three strategies delivered 100% in all of them. I could not get ConPTY to drop bytes at all.

**The cost is severe and measured.** The gap is per-chunk, so wall-clock ≈ bytes/1024 × 16 ms: 64 KB → ~1.7 s, 100 KB → ~1.6 s, 1 MiB → **28.9 s**. That's a large regression on paste to fix a truncation I can't demonstrate, and 1 KiB/16 ms don't correspond to any ConPTY buffer behavior I could measure — they read as tuned constants.

What the PR gets right: it's correctly Windows-gated (non-win32 returns gap 0 and the `chunkGapMs > 0` guard skips the timer entirely, so no macOS/Linux regression), and single keystrokes are unaffected since `iterateTerminalInputChunks` yields the whole string when under the max.

Two real defects if pacing is kept:
- **Remote host is mis-detected.** Both call sites use bare `getTerminalInputChunkMaxBytes()`/`getTerminalInputChunkGapMs()`, which key off the *local* `process.platform`. A macOS client driving a Windows SSH host gets no pacing at all; a Windows client driving a Linux host pays 16 ms/chunk over the relay for nothing. `orca-runtime.ts:8334` already resolves win32-ness per-PTY via `pathFlavorForPty(pty)` including `connectionId` — that's the correct input.
- **No per-PTY write serialization**, so two concurrent `sendTerminal` calls to one PTY interleave chunk-wise. Pre-existing on main, but this widens the window from ~0 ms to 16 ms × chunks, making it far likelier to bite.

I've left `Closes #10416` changed to `Refs` so an unresolved issue isn't auto-closed. My suggestion is to hold this pending a reproducible repro — the reporter's own file-based delivery idea (ask #2 in #10416) and delivery *verification* (ask #3, where dispatch reports success despite failed delivery) look like the higher-value directions. Happy to be shown a repro that contradicts my probe; the scripts are self-contained and I can share them.

nwparker:
## Re-reviewed after the rewrite — three of my findings are fixed; the linkage regressed

I verified the new revision (`fc7d60681c`) rather than taking the summary at face value.

**Fixed, and I confirmed each:**
- **Interleaving.** The per-PTY write queue is real and the test is a genuine detector, not a tautology. Against the merge-base runtime it fails with an actual byte-order assertion (`expected 'aaaa…' to be 'aaaa…'`); at head it passes. That defect was real and is now closed.
- **Wrong host for pacing** — my main concern. `orca-runtime.ts:14210,14246` now reads `getPtyExecutionHostMetadata(ptyId).hostPlatform ?? process.platform`, so a macOS client driving a Windows SSH host paces correctly and a Windows client driving Linux doesn't pay 16 ms/chunk for nothing. WSL maps to `linux`, which is right.
- **Teardown** now cancels pending chunks and clears timers, and stale-incarnation exits don't cancel a replacement PTY.

The RPC keepalive addition is a good catch too — it addresses the 30-second socket idle timeout I measured earlier, which would otherwise have killed large paced sends outright.

**One thing to put back:** this revision restored `Closes #10416`. I've changed it to `Refs #10416` again.

That isn't bookkeeping pedantry. The PR's own body states the Windows probe **did not reproduce** the truncation — my independent probe didn't either (zero loss at 1 MiB unchunked into a stalled reader). Merging with `Closes` would auto-close a bug report whose defect has never been demonstrated to be fixed, and whose reporter would then have to reopen it. `Refs` links the work without making that claim.

**Where that leaves it.** The concurrency, lifetime, remote-host, and timeout fixes are solid, well-tested work that stands on its own merits. The *pacing* remains unjustified by any reproduction — two independent probes on real Windows/ConPTY failed to lose a byte, and 1 KiB/16 ms still costs ~1.55 s per 100 KB and ~262 s at 16 MiB. Reasonable people could ship this as defensive hardening; I'd rather see the pacing gated or dropped until someone reproduces #10416, with the queue/lifetime/keepalive fixes landing regardless. That call is the maintainers', and the body is now honest about the evidence either way.

innocarpe:
@nwparker Thanks for the thorough real-Windows probes and for re-checking the rewrite — the measurements and the split between “justified infrastructure” vs “unjustified pacing” are very clear.

I agree: we still don’t have a reproduction of the silent truncation this PR originally targeted, so I don’t want to ship 1 KiB / 16 ms pacing as a fix for #10416. The per-PTY write queue, host-platform-aware chunking inputs, teardown cancellation, and RPC keepalive are the parts that stand on their own (and the queue regression is real).

Plan from here: drop (or fully disable by default) the Windows inter-chunk gap / small-chunk pacing, keep the serialization + lifetime + keepalive work, and leave the issue as Refs #10416 until someone can reproduce the original drop. Happy to adjust if you’d rather gate pacing behind an env flag instead of removing it outright.

innocarpe:
@nwparker Follow-up on the plan above — pushed to this branch:

- Dropped the Windows 1 KiB / 16 ms pacing (chunk budget is the same on all platforms now; gap stays 0)
- Left the pieces that stood up under review: per-PTY write serialization, host-aware helpers, teardown cancel, RPC keepalive
- Updated tests for the no-pacing behavior
- PR title/body now describe that scope; still `Refs #10416` rather than claiming the truncation is fixed

Happy to tweak further if you'd rather have pacing behind an env flag instead of removing it outright.

innocarpe:
@coderabbitai Tightened the large-chunk test so it only flushes microtasks (no timer advance) before asserting both 16 KiB writes — reintroducing an inter-chunk gap should fail that assertion now.

innocarpe:
### Sync update (`f7fcdf8dad`)

Rebase onto latest upstream/main: serialize paced PTY input writes with absence gate preserved.

- Commit: `f7fcdf8dad88bfa514970b45bf44d3a5b20c8b34`
- Upstream: https://github.com/stablyai/orca/pull/10504
- Fork portfolio: https://github.com/innocarpe/orca/pull/66

Issue #8385, opened 2026-07-12
[Bug]: Grok terminal shows literal [13;2u (Shift+Enter CSI-u leak) and long-output garble


bbingz:
## Summary

In Orca-hosted **Grok** (terminal TUI, not Native Chat), long sessions can show **garbled top-half text**, hard-to-read scroll, and literal keyboard garbage like:

```text
[13;2u
[13;2
```

Reported on **Orca 1.4.137**, **macOS + Windows**, local only, **Grok only**.

## Evidence

`[13;2u` is the Kitty keyboard protocol encoding for **Shift+Enter** (`CSI 13 ; 2 u`) with the leading ESC stripped. That means Orca injected CSI-u key bytes while the application was **not** (or no longer) decoding Kitty keyboard.

## Root cause (fix)

On macOS/Linux, Orca always sent Shift+Enter as `\x1b[13;2u` regardless of whether the pane had enabled Kitty keyboard (`CSI > u`). When Grok (or any TUI) was CSI-u-blind, those bytes appeared as literal text and could corrupt the composer / redraw.

Windows agent-declared CSI-u (Droid) is unchanged. Grok is also declared as CSI-u on trusted Windows foreground so newline works once KKP is active (same class as Droid).

## Fix

- Gate non-Windows Shift+Enter CSI-u on **active KKP flags**; otherwise send Esc+CR.
- Gate Ctrl+Enter CSI-u the same way (avoid `[13;5u` leak).
- Set `windowsShiftEnterEncoding: 'csi-u'` for Grok.

## Reproduction notes

1. Open Grok agent terminal in Orca (not Native Chat).
2. Press Shift+Enter when Kitty handshake is not active (startup race / after mode reset).
3. Observe literal `[13;2u` in the composer or screen corruption on long redraws.

gawaguci:
Regression data point from Windows:

- OS: Windows 11 Pro, build 22631
- Orca: 1.4.164
- Codex CLI: 0.146.0
- Input method: Microsoft Korean IME (2-beolsik)

In an Orca-hosted Codex terminal/composer, pressing Ctrl+Enter produces the literal text `[13;5u` and inserts a newline. After that chord, unexpected newlines continue while typing Korean text.

This is the exact `CSI 13;5u` leakage class described in this issue. It is reproducible on 1.4.164, so the Ctrl+Enter path appears to have regressed or to bypass the previous KKP-active gate in the Windows IME/composer route.

[...]

PR #7668, opened 2026-07-07
fix(terminal): send CSI-u Shift+Enter to Droid on Windows (#7620)

brennanb2025:
## Summary

Fixes #7620. On native Windows ConPTY panes, Shift+Enter now sends Droid's CSI-u newline sequence only while the current PTY process is freshly confirmed as Droid. Codex, Antigravity, shells, WSL, SSH, paired web, and remote runtimes retain the legacy Windows sequence; macOS and Linux behavior is unchanged.

## Reliability proof

**Reliability invariant:** `terminal-input.windows-modified-enter-routing` — CSI-u may be routed only to a currently confirmed Droid process belonging to the active local Windows ConPTY. New commands, PTY replacement, exit evidence, focus/visibility recovery, and unavailable inspection revoke or fail closed to legacy routing.

**Product change type:** Runtime hardening plus experimental reliability-gate coverage.

**Correctness evidence:**
- Exact ConPTY console membership filters global Windows descendants and rejects detached children.
- Fresh confirmation cannot reuse cached process tables or resurrect a recognized fallback that disappeared during the scan.
- Rapid and idle-separated Shift+Enter input remains CSI-u while Droid is current, including panes with live Droid hooks.
- Launch metadata is display/confirmation context only; it never authorizes bytes by itself.
- Stale PTY exit/rebind results and split/detach ownership cannot publish into a replacement pane.
- WSL, SSH, remote-runtime, paired-web, unknown-agent, shell, Codex, and Antigravity paths fail closed to the pre-existing sequence.

**Performance evidence:** Ordinary keys and fast shell completion perform no process scan. Routing recovery uses cancellable one-shot reads on the existing 350 ms / 1.2 s / 6 s ladder; process snapshots are deduplicated, and the bounded ConPTY helper starts only after a recognized candidate exists. WSL/SSH/remote paths perform zero confirmation scans. No polling, provider fanout, hidden-pane wake loop, telemetry, or raw terminal logging was added.

**Validation:**
- `pnpm typecheck`
- `pnpm lint`
- Focused reliability gate: 21 files, 1,215 tests passed on the final rebased tree
- Electron terminal-shortcuts suite passed on macOS; the Windows-only assertion is platform-gated
- Two independent final adversarial review lanes returned clean

**Provider/platform coverage:**
- Local native Windows ConPTY: deterministic main/provider/renderer coverage; live Windows execution is the residual gap
[...]

PR #10511, opened 2026-07-25
fix: OpenCode sessions missing on Windows (#10332) + Pi Shift+Enter submits after tool (#9703)

### LINKED ISSUE #9703: [Bug] Shift+Enter submits instead of newline in Pi after running a tool (Windows + Git Bash) — Pi missing windowsShiftEnterEncoding
### Operating system

Windows

### Orca version

*No response*

## Summary

Running **Pi** (`@earendil-works/pi-coding-agent`) in Orca's integrated terminal on **Windows + Git Bash**, `Shift+Enter` inserts a  <br>newline right after Pi starts and right after refocusing the pane, but it **submits the message instead** immediately after Pi runs a tool  <br>that spawns a child process (`bash`/`read`/`edit`). It then recovers on the next pane refocus.

The same class of issue was fixed for Droid (stablyai/orca#7620 → [#7668](<https://github.com/stablyai/orca/issues/7668>) / stablyai/orca#8948) and Grok (stablyai/orca#8385), but **Pi was never given the****  <br>** `windowsShiftEnterEncoding: 'csi-u'` **config** those fixes introduced, so Pi is left on the flaky live-KKP-flag path.

## Environment

* OS: Windows 11
* Shell: Git Bash (MINGW64) — emits no OSC 133 prompt marks
* Terminal: Orca integrated terminal (xterm.js), latest build (includes [#1247](<https://github.com/stablyai/orca/issues/1247>), [#7668](<https://github.com/stablyai/orca/issues/7668>), stablyai/orca#8948, stablyai/orca#8385)
* Agent: Pi (`@earendil-works/pi-coding-agent`), latest
* `$TERM_PROGRAM=Orca`, `TERM=xterm-256color`

## Steps to reproduce

1. Launch Pi in an Orca agent pane (Git Bash as the shell).
2. At the prompt, press `Shift+Enter` → inserts a newline ✅
3. Let Pi run any tool that spawns a subprocess (e.g. a `bash`/`read`/`edit` tool).
4. Back at Pi's prompt, press `Shift+Enter` → **submits** ❌
5. Click/refocus the pane, press `Shift+Enter` → inserts a newline again ✅ (until the next tool run)

## Root cause

Orca routes `Shift+Enter` in `src/renderer/src/components/terminal-pane/terminal-shortcut-policy.ts` (~line 152):

```ts
const hasTrustedWindowsCsiU = windowsHost && getWindowsShiftEnterEncoding?.() === 'csi-u'
const canSendCsiU = hasTrustedWindowsCsiU || isKittyKeyboardActivePane?.() === true
return { type: 'sendInput', data: canSendCsiU ? '\x1b[13;2u' : '\x1b\r' }
```

CSI-u (newline) is sent only if either the agent has windowsShiftEnterEncoding: 'csi-u' trusted or the pane has active Kitty-keyboard  <br>negotiation (isKittyKeyboardActivePane).

For Pi, path 1 is closed: in src/shared/tui-agent-config.ts, TUI_AGENT_CONFIG< />'pi'< /> does not set windowsShiftEnterEncoding — only droid  <br>has it (line 228). So on Windows, Pi can only ever get CSI-u via path 2 (isKittyKeyboardActivePane).

Path 2 is the flaky one. isKittyKeyboardActivePane() reads flags from TerminalKittyKeyboardModeTracker  <br>(src/shared/terminal-kitty-keyboard-mode-tracker.ts), which scans PTY output and resets flags to 0 on:

* CSI < u (pop)
* RIS (\x1bc full reset)
* DECSTR (CSI ! p soft reset)
* alternate-screen swaps (DECSET/DECRST 47/1047/1049)

Pi pushes Kitty (CSI > 1u) once at startup (queryAndEnableKittyProtocol in pi-tui terminal.js) and never re-pushes (drainInput, which  <br>pops, is exit-only). During a tool run, something in Pi's re-render/tool path emits one of those reset sequences, the tracker flags drop  <br>to 0 → Orca sends ESC+CR → Pi reads it as plain Enter → submit. On refocus, Orca replays retained history via scanReplay, which re-applies  <br>Pi's startup push idempotently → flags > 0 → Shift+Enter works again. This exactly matches the observed "works after start/refocus, breaks  <br>after a tool run" pattern.

Pi itself is not at fault (pushes once, pops only at exit). Every agent without windowsShiftEnterEncoding (claude, codex, gemini, cursor,  <br>…) is on the same flaky path; Pi is just the most visible because it spawns a child process on every tool call.

Expected

Once a pane's program has advertised the Kitty keyboard protocol, Shift+Enter should keep sending CSI-u for the life of that pane — not  <br>flip on/off with tool-subprocess churn.

Suggested fix (any one, or combination)

1. Add windowsShiftEnterEncoding: 'csi-u' to TUI_AGENT_CONFIG< />'pi'< /> (and the other agents currently missing it), and keep routingTrusted  <br>sticky across the foreground churn from tool subprocesses — so path 1 covers Pi reliably.
2. Or: make TerminalKittyKeyboardModeTracker not drop flags on the sequences Pi emits during tool runs, or re-push/replay KKP when focus  <br>returns to the pane.
3. Or: have Orca re-probe/re-push KKP on focus return for panes whose app already negotiated it.

Workaround

Ctrl+J — sends 0x0A (distinct from Enter's 0x0D), bound as a newline alias by Pi by default. Works on every terminal regardless of  <br>Kitty/trust state.

References

* [#1247](<https://github.com/stablyai/orca/issues/1247>) — advertise Kitty keyboard protocol on xterm.js
* [#7668](<https://github.com/stablyai/orca/issues/7668>) — send CSI-u Shift+Enter to Droid on Windows
* stablyai/orca#8948 — fix no-OSC shells (Git Bash / cmd) spawn-time latching
* stablyai/orca#8385 — gate CSI-u on active KKP flags (the gating that makes this flaky) + set windowsShiftEnterEncoding for Grok
* Pi terminal-setup notes: Pi pushes Kitty once at startup, Ctrl+J is the default newline alias

### LINKED ISSUE #10332: [Bug]: OpenCode sessions missing from Agent Session History on Windows — `getXdgDataHome()` returns `%LOCALAPPDATA%` instead of `~/.local/share`
### Operating system

windows

### Orca version

_No response_

### Summary

On Windows, the Agent Session History panel does not show any OpenCode sessions, even though OpenCode's SQLite database (`opencode.db`) exists and is correctly populated. Other agents (e.g. Pi) appear normally. The root cause is that Orca's `getXdgDataHome()` resolves to `%LOCALAPPDATA%` on Windows, so it looks for the DB at `%LOCALAPPDATA%\opencode\opencode.db` — but OpenCode stores its data at `~/.local/share/opencode/opencode.db` (Unix-style XDG path, used cross-platform by OpenCode).

### Environment

- **Orca**: v1.4.152 (Windows, packaged build)
- **OpenCode**: v1.17.13
- **OS**: Windows 11 (win32, x64)
- **OpenCode DB path**: `C:\Users\<user>\.local\share\opencode\opencode.db`
- **Path Orca scans**: `C:\Users\<user>\AppData\Local\opencode\opencode.db` (does not exist)

### Reproduction

1. Install OpenCode on Windows and create at least one session (verify `~/.local/share/opencode/opencode.db` exists with rows in the `session` table).
2. Install Orca v1.4.152 on Windows.
3. Open the Agent Session History panel (right sidebar → Agents tab).
4. Ensure OpenCode is toggled ON in panel Settings → Agents.
5. Click Refresh.
6. **Observed**: No OpenCode sessions appear. Other agents (Pi, Claude, Codex, etc.) show normally.
7. **Expected**: OpenCode sessions from `opencode.db` should appear in the panel.

### Root cause

In `out/main/index.js`, the OpenCode DB discovery chain is:

```js
function getXdgDataHome() {
  if (process.env.XDG_DATA_HOME?.trim()) {
    return process.env.XDG_DATA_HOME.trim();
  }
  if (process.platform === "win32") {
    return process.env.LOCALAPPDATA || process.env.APPDATA || path.join(os.homedir(), "AppData", "Local");
  }
  return path.join(os.homedir(), ".local", "share");
}

function getOpenCodeDataDirectory() {
  return path.join(getXdgDataHome(), "opencode");
}

async function listOpenCodeDatabases() {
  const envPath = getOpenCodeDatabasePathFromEnv();
  if (envPath) {
    return node_fs.existsSync(envPath) ? [envPath] : [];
  }
  try {
    const entries = await fs.readdir(getOpenCodeDataDirectory(), { withFileTypes: true });
    return entries.filter(...).map(...);
  } catch {
    return [];  // silently returns empty — no error, no scan issue
  }
}
```

On Windows, `getXdgDataHome()` returns `%LOCALAPPDATA%` (`C:\Users\<user>\AppData\Local`), so `getOpenCodeDataDirectory()` returns `C:\Users\<user>\AppData\Local\opencode`. The `readdir` fails silently (directory doesn't exist), `listOpenCodeDatabases()` returns `[]`, and the OpenCode SQLite scanner is never invoked. No scan issue is surfaced because the `catch` block swallows the error.

However, OpenCode on Windows stores data at `~/.local/share/opencode/` — the same Unix-style XDG path it uses on Linux/macOS. This is confirmed by:
- OpenCode's own source code (it uses `~/.local/share/opencode/` cross-platform)
- The actual DB location on a fresh OpenCode install on Windows
- Multiple OpenCode issues referencing `~/.local/share/opencode/opencode.db` on Windows (e.g. [#13560](https://github.com/anomalyco/opencode/issues/13560), [#31597](https://github.com/anomalyco/opencode/issues/31597))

### Evidence

**Orca's trace log** (`%APPDATA%\orca\logs\main.trace.ndjson`): zero matches for "opencode", "sqlite", "session-scanner", "ai-vault", or "scan" across 10,858 lines. The OpenCode SQLite scanner is never invoked.

**Orca's session parse cache** (`%APPDATA%\orca\ai-vault\session-parse-cache.json`): contains only Pi sessions — zero OpenCode entries.

**Orca's view options** (from Local Storage / leveldb): `{"disabledAgents":[],"sort":"updated","group":"project","hideEmptySessions":false}` — OpenCode is not disabled.

**File system**:
```
# Where Orca looks (does not exist):
C:\Users\<user>\AppData\Local\opencode\opencode.db  → NOT FOUND

# Where OpenCode actually stores it:
C:\Users\<user>\.local\share\opencode\opencode.db   → EXISTS, 1+ sessions
```

**Workaround that confirmed the diagnosis**: creating a directory junction from `%LOCALAPPDATA%\opencode` → `~/.local/share/opencode` immediately fixed the issue — after restarting Orca and clicking Refresh, OpenCode sessions appeared in the panel.

### Suggested fix

`getXdgDataHome()` (or at least `getOpenCodeDataDirectory()`) should resolve to `path.join(os.homedir(), ".local", "share")` on Windows when locating OpenCode's data directory, matching OpenCode's actual storage path — not `%LOCALAPPDATA%`.

Option A — fix `getOpenCodeDataDirectory()` specifically:
```js
function getOpenCodeDataDirectory() {
  if (process.env.OPENCODE_CONFIG_DIR?.trim()) {
    return path.join(process.env.OPENCODE_CONFIG_DIR.trim(), "..") // or resolve appropriately
  }
  // OpenCode uses ~/.local/share/opencode cross-platform, including Windows
  return path.join(os.homedir(), ".local", "share", "opencode");
}
```

Option B — respect `OPENCODE_CONFIG_DIR` env var (which OpenCode itself uses) as the primary source, falling back to `~/.local/share/opencode`.

Additionally, consider surfacing a scan issue when `listOpenCodeDatabases()` catches an error or returns empty, rather than silently swallowing it — this would have made the problem immediately diagnosable from the UI.

### Workaround for affected users

Create a directory junction (does not require admin privileges):

```cmd
mklink /J "%LOCALAPPDATA%\opencode" "%USERPROFILE%\.local\share\opencode"
```

Then restart Orca and click Refresh in the Agent Session History panel.

itisvincent:
## Summary

Two independent Windows bugs, one line of behavior change each.

**1. OpenCode sessions missing from Agent Session History on Windows (#10332).** OpenCode resolves its data directory through [`xdg-basedir`](https://github.com/sindresorhus/xdg-basedir) on *every* platform, so on Windows its SQLite DB lands at `~/.local/share/opencode/opencode.db` — not under `%LOCALAPPDATA%`. This is a known, still-open quirk on OpenCode's side ([sst/opencode#8235](https://github.com/anomalyco/opencode/issues/8235) — "Config and Data directories follow the Linux XDG standard even on windows"). Orca's `getOpenCodeDataDirectory()` had a `process.platform === 'win32'` branch pointing at `%LOCALAPPDATA%\opencode`, which never exists. The `readdir` threw, the `catch` swallowed it, `listOpenCodeDatabases()` returned `[]`, and the OpenCode SQLite scanner was never invoked — so every other agent showed up and only OpenCode was missing. Fix: drop the Windows branch, keeping the `XDG_DATA_HOME` override.

**2. Pi Shift+Enter submits instead of inserting a newline after a tool call (#9703).** `TUI_AGENT_CONFIG['pi']` never set `windowsShiftEnterEncoding`, so on Windows Pi could only reach CSI-u via the live-KKP-flag path. Pi pushes the Kitty keyboard protocol once at startup and never re-pushes; when a tool spawns a subprocess that emits a reset sequence (RIS / DECSTR / alt-screen swap), the tracker's flags drop to 0, Orca falls back to `Esc+CR`, and Pi reads that as a plain Enter → **submit**. It self-heals on the next pane refocus, which is exactly the reported "works at start, breaks after a tool, fixed by clicking the pane" pattern. Fix: set `windowsShiftEnterEncoding: 'csi-u'` for `pi`, mirroring the existing Droid (#7668) and Grok (#8385) fixes, so the trusted route no longer depends on KKP-flag churn.

## ELI5

Orca was looking for OpenCode's history file in the wrong folder on Windows, so the Agents panel always showed an empty list for OpenCode even though the history was there — we now look in the folder OpenCode actually uses. Separately, when chatting with the Pi agent on Windows, pressing Shift+Enter to add a new line would accidentally send the message instead, but only right after Pi used a tool. Orca now tells Pi's keypresses apart reliably, so Shift+Enter always just adds a new line.

### Fix A — OpenCode Windows path (#10332)

The regression test drives the real entry point (`listOpenCodeDatabases`) with `process.platform` stubbed to `win32`, a temp `$HOME`, `%LOCALAPPDATA%` populated, and a real `opencode.db` on disk at the XDG location.

With `src/main/opencode-usage/scanner.ts` checked out from `origin/main`:

```
 FAIL  src/main/opencode-usage/scanner.test.ts > listOpenCodeDatabases > finds opencode.db under ~/.local/share on Windows instead of %LOCALAPPDATA% (#10332)
AssertionError: expected [] to deeply equal [ Array(1) ]

- Expected
+ Received

- [
-   ".../orca-opencode-home-c70KOx/.local/share/opencode/opencode.db",
- ]
+ []

 ❯ src/main/opencode-usage/scanner.test.ts:76:43

 Test Files  1 failed (1)
      Tests  1 failed | 15 passed (16)
```

With the fix applied:

```
 Test Files  1 passed (1)
      Tests  16 passed (16)
```

**Verified on a real Windows 11 box** (`win32 x64`, Node v24.18.0), running both resolution strategies side by side against a seeded DB:

```
platform      : win32 | homedir: C:\Users\neil
LOCALAPPDATA  : C:\Users\neil\AppData\Local
MAIN  scans   : C:\Users\neil\AppData\Local\opencode -> dbs found: []
FIXED scans   : C:\Users\neil\.local\share\opencode -> dbs found: ["C:\\Users\\neil\\.local\\share\\opencode\\opencode.db"]
```

An independent directory listing on that same box confirms `C:\Users\neil\.local\share\` is a real, populated location (it already holds `claude` and `devex`), while a recursive search for `opencode.db` under `%LOCALAPPDATA%` and `%APPDATA%` returns nothing.

**Confirmed on a real Windows 11 box** (win32 x64) — the premise is not inferred from the upstream issue, it is observed on disk:

```
EXISTS  C:\Users\neil\.local\share\opencode
missing C:\Users\neil\AppData\Local\opencode
missing C:\Users\neil\AppData\Roaming\opencode
XDG_DATA_HOME=(unset)
```

OpenCode's data directory is present at the XDG path and absent from both `%LOCALAPPDATA%` and `%APPDATA%` — exactly the directory `main` was reading, which is why `readdir` threw and the scanner silently returned `[]`. (`opencode.db` itself is absent on that box because OpenCode has not been run there; the directory location is the claim under test.)

### Fix B — Pi Shift+Enter (#9703)

With `src/shared/tui-agent-config.ts` checked out from `origin/main`:

```
 FAIL  src/renderer/src/components/terminal-pane/terminal-windows-shift-enter.test.ts > resolveWindowsShiftEnterEncoding > uses CSI-u for trusted Pi process evidence so Shift+Enter survives tool-subprocess churn (#9703)
AssertionError: expected 'alt-enter' to be 'csi-u' // Object.is equality

Expected: "csi-u"
Received: "alt-enter"

 ❯ src/renderer/src/components/terminal-pane/terminal-windows-shift-enter.test.ts:22:7

 Test Files  1 failed (1)
      Tests  1 failed | 8 passed (9)
```

With the fix applied:

```
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

## Proof of no regressions

```
$ npx vitest run --config config/vitest.config.ts src/main/opencode-usage/ src/main/ai-vault/
 Test Files  34 passed (34)
      Tests  224 passed (224)

    src/renderer/src/components/terminal-pane/terminal-shortcut-policy.test.ts \
    src/renderer/src/components/terminal-pane/keyboard-handlers.test.ts \
    src/renderer/src/components/terminal-pane/pty-connection.test.ts \
    src/renderer/src/components/terminal-pane/terminal-windows-shift-enter.test.ts
 Test Files  4 passed (4)
      Tests  542 passed (542)

$ npx tsc --noEmit -p config/tsconfig.node.json     # clean, exit 0
$ npx tsc --noEmit -p config/tsconfig.tc.web.json   # clean, exit 0
$ npx oxlint <the four changed files>               # clean, 0 warnings
```

No user-visible behavior changes beyond the two fixes. Justification:

- **Fix A** only removes a `process.platform === 'win32'` branch. macOS and Linux never entered it, so their resolved path is byte-identical. The `XDG_DATA_HOME` and `OPENCODE_DB` overrides are unchanged and both are now covered by tests (the `OPENCODE_DB` case was added per review feedback). The only behavior that changes is on Windows, where the scanned directory moves from a path that never exists to the one that does.
- **Fix B** adds one property to a single entry in `TUI_AGENT_CONFIG`. I diffed the file against `origin/main`: the *only* changed lines are inside the `pi` entry (a trailing comma, a comment, and the new key). Claude, Codex, OpenCode, Cursor, Gemini, Grok, Droid and every other agent config are byte-unchanged. Plain Enter is untouched — it is handled by a different branch of `terminal-shortcut-policy.ts` that never consults this config, and the Shift+Enter branch is the only reader. A new test pins that trusted agents *without* an override (gemini, cursor, codex) still resolve to `alt-enter`.

## Compatibility

- **Platforms:** Windows — both fixes verified. Fix A's path resolution was executed on a real Windows 11 box (output above); Fix B's resolver is unit-tested with the Windows code path exercised directly. macOS/Linux — Fix A provably unchanged (the removed branch was Windows-gated); Fix B is gated behind `isWindowsTerminalHost()` at the call site in `terminal-shortcut-policy.ts`, and an existing test already pins that off-Windows panes require active KKP negotiation regardless of the configured encoding. All paths use `path.join`; no separator is hardcoded. Note that `listOpenCodeDatabases` uses `readdir` plus a filename regex rather than a glob library, so there is no forward-slash-in-pattern hazard.
- **SSH / remote:** No impact. Remote and WSL OpenCode sources are resolved separately in `session-scanner-opencode-sources.ts`, which already used `~/.local/share/opencode` for WSL homes and is untouched by this PR. This change only affects the local-host lookup, which now agrees with the WSL logic instead of contradicting it.
- **Performance:** No hot-path change. Fix A is one `path.join` in a scan that already hits the filesystem. Fix B is a record property read, and the call site is lazy — `getWindowsShiftEnterEncoding` is only invoked on Shift+Enter, never on ordinary keystrokes (an existing test asserts this).

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` (`npx oxlint` on the changed files)
- [x] `pnpm typecheck` (`config/tsconfig.node.json` + `config/tsconfig.tc.web.json`)
- [x] `pnpm test` — ran the affected suites rather than the full 30+ min suite: `src/main/opencode-usage/`, `src/main/ai-vault/`, and the four `terminal-pane` suites covering Shift+Enter (766 tests, all passing)
- [ ] `pnpm build`

## AI Review Report

Reviewed for correctness, cross-platform behavior, SSH/remote impact, performance, and regression surface.

The review flagged one real defect in the original patch, which is fixed here. The first version of the Fix A regression test exported a new pure resolver `resolveOpenCodeDataDirectory({ env, homeDir })` and asserted against injected values. That test was **vacuous**: main's bug lives behind `process.platform === 'win32'`, which the test never stubbed, so on any non-Windows CI host main's own logic satisfies the same assertion. Transplanting main's resolver verbatim and running the assertion against it confirmed it returns the expected value on macOS/Linux. On `origin/main` the test failed only with `TypeError: resolveOpenCodeDataDirectory is not a function` — i.e. it was detecting a missing symbol, not a wrong path, and would have kept passing if someone reintroduced the `%LOCALAPPDATA%` branch. Replaced it with a test that drives the real `listOpenCodeDatabases` entry point with `process.platform` stubbed to `win32` and a real DB on disk; it now fails on main for the correct reason (see Proof of fix). That also made the exported options-injection API unnecessary, so `getOpenCodeDataDirectory` is back to a 4-line private function and the production diff shrank.

Cross-platform compatibility was checked explicitly for macOS, Linux, and Windows, covering path separators, drive letters, the glob-vs-`readdir` distinction, shortcut byte protocols, and the Windows-host gating of the Shift+Enter encoding — see **Compatibility** above for the per-platform findings. The review also confirmed neither bug is already fixed on `main`, and that the Fix B config change touches only the `pi` entry.

## Security Audit

No new attack surface. No command execution, no network calls, no IPC surface, no auth or secret handling, no dependency changes.

- **Path handling:** the change narrows the set of directories Orca reads — it removes a lookup location rather than adding one. The resolved directory is still derived from `homedir()` or the pre-existing `XDG_DATA_HOME` override, and the filename filter remains an anchored regex (`/^opencode(?:-[A-Za-z0-9_.-]+)?\.db$/`) applied to `readdir` entry names, so it cannot traverse outside the data directory. Reads stay wrapped in the existing `try/catch`.
- **Input handling:** the Shift+Enter change routes a fixed, hardcoded byte sequence (`\x1b[13;2u`) and is only reachable via trusted foreground process evidence. Hook- and OSC-derived status remain excluded from routing authority, so PTY output still cannot forge it — the existing test covering that is unchanged and passing.

## Notes

- **Why these two fixes ship together:** they were reported and submitted as one Windows-bugfix pair, and each is a genuinely independent one-line behavior change in a different subsystem (`src/main/opencode-usage/` vs `src/shared/`) with no shared code, no shared test file, and no interaction. Splitting would add review overhead without reducing risk; each fix has its own separately-labelled failing-on-main proof above, so either can be reverted alone.
- `listOpenCodeDatabases()` still swallows `readdir` errors and returns `[]`, so a genuinely unreadable or missing data directory remains silent in the UI. That is what made #10332 hard to diagnose in the first place. Surfacing an `AiVaultScanIssue` there is a worthwhile follow-up but is deliberately out of scope for a targeted bugfix.
- Fix B addresses Pi specifically, per the issue. The underlying fragility the issue describes — every agent without `windowsShiftEnterEncoding` relies on live KKP flags that tool subprocesses can reset — still applies to the remaining agents. Making the KKP tracker resilient to reset sequences, or re-pushing KKP on focus return, would fix the whole class rather than one agent at a time; that is a larger change and a good follow-up.

Closes #10332
Closes #9703

Original patch by @itisvincent.

### DIFF
--- a/src/main/opencode-usage/scanner.test.ts
+++ b/src/main/opencode-usage/scanner.test.ts
@@ -5,9 +5,9 @@ import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import Database from '../sqlite/sync-database'
 import {
   attributeOpenCodeUsageEvent,
+  listOpenCodeDatabases,
   parseOpenCodeUsageDatabase,
   parseOpenCodeUsageRow,
-  resolveOpenCodeDataDirectory,
   scanOpenCodeUsageDatabases
 } from './scanner'
 
@@ -22,20 +22,81 @@ function createTempDb(): { db: Database.Database; path: string } {
   return { db: new Database(path), path }
 }
 
-describe('resolveOpenCodeDataDirectory', () => {
-  it('uses XDG_DATA_HOME when set', () => {
-    expect(resolveOpenCodeDataDirectory({ env: { XDG_DATA_HOME: '/custom/xdg' }, homeDir: '/home/u' }))
-      .toBe(join('/custom/xdg', 'opencode'))
+describe('listOpenCodeDatabases', () => {
+  const originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform')
+  let home: string
+  let previousEnv: Partial<Record<string, string | undefined>>
+
+  beforeEach(() => {
+    home = mkdtempSync(join(tmpdir(), 'orca-opencode-home-'))
+    tempDirs.push(home)
+    previousEnv = {
+      HOME: process.env.HOME,
+      USERPROFILE: process.env.USERPROFILE,
+      XDG_DATA_HOME: process.env.XDG_DATA_HOME,
+      LOCALAPPDATA: process.env.LOCALAPPDATA,
+      APPDATA: process.env.APPDATA,
+      OPENCODE_DB: process.env.OPENCODE_DB
+    }
+    process.env.HOME = home
+    process.env.USERPROFILE = home
+    delete process.env.XDG_DATA_HOME
+    delete process.env.OPENCODE_DB
   })
 
-  it('falls back to ~/.local/share on every platform, ignoring %LOCALAPPDATA% on Windows (#10332)', () => {
-    // Why: OpenCode stores its DB at ~/.local/share/opencode cross-platform; the
-    // previous Windows branch pointed at %LOCALAPPDATA%/opencode and never found it.
-    const winLike = { LOCALAPPDATA: 'C:\\Users\\u\\AppData\\Local', APPDATA: 'C:\\Users\\u\\AppData\\Roaming' }
-    expect(resolveOpenCodeDataDirectory({ env: winLike, homeDir: 'C:\\Users\\u' }))
-      .toBe(join('C:\\Users\\u', '.local', 'share', 'opencode'))
-    expect(resolveOpenCodeDataDirectory({ env: {}, homeDir: '/home/u' }))
-      .toBe(join('/home/u', '.local', 'share', 'opencode'))
+  afterEach(() => {
+    if (originalPlatform) {
+      Object.defineProperty(process, 'platform', originalPlatform)
+    }
+    for (const [key, value] of Object.entries(previousEnv)) {
+      if (value === undefined) {
+        delete process.env[key]
+      } else {
+        process.env[key] = value
+      }
+    }
+    for (const dir of tempDirs) {
+      rmSync(dir, { recursive: true, force: true })
+    }
+    tempDirs = []
+  })
+
+  it('finds opencode.db under ~/.local/share on Windows instead of %LOCALAPPDATA% (#10332)', async () => {
+    // Why: OpenCode resolves its data dir via xdg-basedir on Windows too, so the DB
+    // lands in ~/.local/share/opencode while %LOCALAPPDATA%\opencode never exists.
+    Object.defineProperty(process, 'platform', { configurable: true, value: 'win32' })
+    process.env.LOCALAPPDATA = join(home, 'AppData', 'Local')
+    process.env.APPDATA = join(home, 'AppData', 'Roaming')
+    mkdirSync(join(home, 'AppData', 'Local'), { recursive: true })
+    const dataDir = join(home, '.local', 'share', 'opencode')
+    mkdirSync(dataDir, { recursive: true })
+    const dbPath = join(dataDir, 'opencode.db')
+    new Database(dbPath).close()
+
+    expect(await listOpenCodeDatabases()).toEqual([dbPath])
+  })
+
+  it('still honors XDG_DATA_HOME when set', async () => {
+    const xdgRoot = mkdtempSync(join(tmpdir(), 'orca-opencode-xdg-'))
+    tempDirs.push(xdgRoot)
+    process.env.XDG_DATA_HOME = xdgRoot
+    mkdirSync(join(xdgRoot, 'opencode'), { recursive: true })
+    const dbPath = join(xdgRoot, 'opencode', 'opencode.db')
+    new Database(dbPath).close()
+
+    expect(await listOpenCodeDatabases()).toEqual([dbPath])
+  })
+
+  it('keeps an explicit OPENCODE_DB override ahead of the derived directory', async () => {
+    const overrideDir = mkdtempSync(join(tmpdir(), 'orca-opencode-override-'))
+    tempDirs.push(overrideDir)
+    const overridePath = join(overrideDir, 'opencode.db')
+    new Database(overridePath).close()
+    mkdirSync(join(home, '.local', 'share', 'opencode'), { recursive: true })
+    new Database(join(home, '.local', 'share', 'opencode', 'opencode.db')).close()
+    process.env.OPENCODE_DB = overridePath
+
+    expect(await listOpenCodeDatabases()).toEqual([overridePath])
   })
 })
 
--- a/src/main/opencode-usage/scanner.test.ts
+++ b/src/main/opencode-usage/scanner.test.ts
@@ -7,6 +7,7 @@ import {
   attributeOpenCodeUsageEvent,
   parseOpenCodeUsageDatabase,
   parseOpenCodeUsageRow,
+  resolveOpenCodeDataDirectory,
   scanOpenCodeUsageDatabases
 } from './scanner'
 
@@ -20,6 +21,23 @@ function createTempDb(): { db: Database.Database; path: string } {
   const path = join(dir, 'opencode.db')
   return { db: new Database(path), path }
 }
+
+describe('resolveOpenCodeDataDirectory', () => {
+  it('uses XDG_DATA_HOME when set', () => {
+    expect(resolveOpenCodeDataDirectory({ env: { XDG_DATA_HOME: '/custom/xdg' }, homeDir: '/home/u' }))
+      .toBe(join('/custom/xdg', 'opencode'))
+  })
+
+  it('falls back to ~/.local/share on every platform, ignoring %LOCALAPPDATA% on Windows (#10332)', () => {
+    // Why: OpenCode stores its DB at ~/.local/share/opencode cross-platform; the
+    // previous Windows branch pointed at %LOCALAPPDATA%/opencode and never found it.
+    const winLike = { LOCALAPPDATA: 'C:\\Users\\u\\AppData\\Local', APPDATA: 'C:\\Users\\u\\AppData\\Roaming' }
+    expect(resolveOpenCodeDataDirectory({ env: winLike, homeDir: 'C:\\Users\\u' }))
+      .toBe(join('C:\\Users\\u', '.local', 'share', 'opencode'))
+    expect(resolveOpenCodeDataDirectory({ env: {}, homeDir: '/home/u' }))
+      .toBe(join('/home/u', '.local', 'share', 'opencode'))
+  })
+})
 
 function worktrees() {
   return [
--- a/src/main/opencode-usage/scanner.ts
+++ b/src/main/opencode-usage/scanner.ts
@@ -79,18 +79,22 @@ function looksLikeWindowsPath(pathValue: string): boolean {
   return /^[A-Za-z]:[\\/]/.test(pathValue) || pathValue.startsWith('\\\\')
 }
 
-function getXdgDataHome(): string {
-  if (process.env.XDG_DATA_HOME?.trim()) {
-    return process.env.XDG_DATA_HOME.trim()
-  }
-  if (process.platform === 'win32') {
-    return process.env.LOCALAPPDATA || process.env.APPDATA || join(homedir(), 'AppData', 'Local')
-  }
-  return join(homedir(), '.local', 'share')
+type OpenCodeDataDirOptions = {
+  env?: NodeJS.ProcessEnv
+  homeDir?: string
+}
+
+// Why: OpenCode stores its data at ~/.local/share/opencode on every platform, including Windows — not under %LOCALAPPDATA% (#10332). Exported pure form lets the Windows path be regression-tested without a real home dir.
+export function resolveOpenCodeDataDirectory(opts: OpenCodeDataDirOptions = {}): string {
+  const env = opts.env ?? process.env
+  const home = opts.homeDir ?? homedir()
+  const xdg = env.XDG_DATA_HOME?.trim()
+  const dataHome = xdg ? xdg : join(home, '.local', 'share')
+  return join(dataHome, 'opencode')
 }
 
 function getOpenCodeDataDirectory(): string {
-  return join(getXdgDataHome(), 'opencode')
+  return resolveOpenCodeDataDirectory()
 }
 
 function getOpenCodeDatabasePathFromEnv(): string | null {
--- a/src/main/opencode-usage/scanner.ts
+++ b/src/main/opencode-usage/scanner.ts
@@ -79,22 +79,11 @@ function looksLikeWindowsPath(pathValue: string): boolean {
   return /^[A-Za-z]:[\\/]/.test(pathValue) || pathValue.startsWith('\\\\')
 }
 
-type OpenCodeDataDirOptions = {
-  env?: NodeJS.ProcessEnv
-  homeDir?: string
-}
-
-// Why: OpenCode stores its data at ~/.local/share/opencode on every platform, including Windows — not under %LOCALAPPDATA% (#10332). Exported pure form lets the Windows path be regression-tested without a real home dir.
-export function resolveOpenCodeDataDirectory(opts: OpenCodeDataDirOptions = {}): string {
-  const env = opts.env ?? process.env
-  const home = opts.homeDir ?? homedir()
-  const xdg = env.XDG_DATA_HOME?.trim()
-  const dataHome = xdg ? xdg : join(home, '.local', 'share')
-  return join(dataHome, 'opencode')
-}
-
+// Why: OpenCode resolves its data dir through xdg-basedir on every platform, so on
+// Windows the DB is at ~/.local/share/opencode, not %LOCALAPPDATA% (#10332).
 function getOpenCodeDataDirectory(): string {
-  return resolveOpenCodeDataDirectory()
+  const xdgDataHome = process.env.XDG_DATA_HOME?.trim()
+  return join(xdgDataHome || join(homedir(), '.local', 'share'), 'opencode')
 }
 
 function getOpenCodeDatabasePathFromEnv(): string | null {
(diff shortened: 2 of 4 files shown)


itisvincent:
Splitting out the Shift+Enter half (#9703) into a clean, rebased PR since the OpenCode Windows-data-dir half (#10332) landed independently in #10362 and this PR now conflicts on `scanner.ts`.

The Pi CSI-u Shift+Enter fix is now here: #11769 (rebased onto current `main`, OpenCode changes dropped as redundant). Happy to close this one in favor of #11769 unless you'd prefer to repurpose it.

nwparker:

- Issue #9703 was fixed by #11769 (`fix(terminal): trust Pi CSI-u Shift+Enter on Windows (#9703)`)
- Issue #10332 was fixed by #10362 (`fix(opencode): use cross-platform data directory`)

Superseding merge: #11769, #10362

PR #10513, opened 2026-07-25
fix(sidebar): stop worktree cycling from reopening collapsed groups

jeongph:
## Summary

Cycling worktrees with `Cmd/Ctrl+Shift+Arrow` no longer forces collapsed groups open.

`navigateWorktree` deliberately built its cycling list from an **all-expanded** layout (`buildRows(..., new Set<string>(), ...)`) so navigation would not skip worktrees hidden inside collapsed groups. The reveal that follows activation then uncollapses whatever group the target sits in. The net effect: arrowing through the sidebar walks into every group the user collapsed and pops them open one by one, and there is no way to keep a group closed while navigating.

Collapsing a group is the user saying "not now". Cycling now runs over the sidebar as rendered, so collapsed groups are skipped and stay closed. Expanding the group brings its worktrees back into the rotation.

One consequence needed handling: the active worktree stays selected even when its group is collapsed, so it can be absent from the cyclable list. Previously an unmatched active worktree always jumped to index 0. Entry now comes from the end the keypress points away from — `Down` enters at the first worktree, `Up` at the last — which is what wrapping does everywhere else in the list.

## Screenshots

No screenshot; the change is keyboard navigation behavior. Reproduction, on a sidebar with two projects:

1. Collapse the second project's group.
2. Put focus outside a text field and press `Cmd/Ctrl+Shift+Down` repeatedly.

Before: cycling walks into the collapsed group and expands it. After: cycling stays within the visible worktrees and the group remains collapsed.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

`pnpm test` reports 13 pre-existing failures in this checkout (electron-builder/packaging/WSL/localStorage suites); the identical 13 files and 45 cases fail on unmodified `main` here, so they are environmental and unrelated. `src/renderer/src/components/sidebar/` passes apart from three of those same pre-existing files (180/183).

New tests in `worktree-keyboard-cycle.test.ts`:
- unit coverage for `resolveCycledWorktreeId` — stepping, wrapping at both ends, entry direction when the active worktree is not cyclable, and the empty list (every group collapsed).
- a source assertion that `navigateWorktree` feeds the sidebar's own collapsed set into `buildRows` rather than an empty one, following the `readWorktreeListSource()` pattern already used in `worktree-list-groups.test.ts`.

## AI Review Report

- **Cycling could dead-end.** If every group is collapsed the list is empty; `resolveCycledWorktreeId` returns `null` and the handler returns without activating anything, rather than throwing on `worktreeRows[nextIndex].worktree`. The old code guarded this with a separate length check, which the null return now subsumes. Covered by a test.
- **Which collapsed set is correct.** `VirtualizedWorktreeViewport` receives `collapsedGroups` from the parent's `effectiveCollapsedGroups`, which already un-collapses the agent-send target's groups. Using the prop keeps cycling aligned with exactly what is on screen, including that exception.
- **Reveal side effects.** `activateAndRevealWorktree` still runs, but its target is now always a visible row, so the reveal path has no collapsed group to open. Scroll-to-index behavior is unchanged.
- **Stale closure.** `collapsedGroups` was added to the `useCallback` dependency list, so toggling a group rebuilds the callback.
- **Agent and integration compatibility.** Cycling reads only row identity and the collapsed-group set; nothing branches on agent type, git provider, or integration state. A worktree running a CLI agent cycles exactly like a plain one.
- **Performance.** `navigateWorktree` already rebuilt rows per keypress; it now passes the existing collapsed set instead of allocating a fresh empty one, so collapsed groups short-circuit earlier and the row list is the same size or smaller. The index scan became a single `indexOf` over worktree ids. No new work per render or per frame.
- **UI quality.** The reveal still scrolls the target into view (`virtualizer.scrollToIndex`), so the selection stays visible; with collapsed groups skipped there is no longer a group expand-and-reflow on every step, which removes a scroll jump rather than adding one.
- **Remote/SSH and local.** Row construction is unchanged for remote-backed worktrees; the collapsed set is host-agnostic, so SSH and local worktrees cycle identically and nothing new is fetched over the wire.
- **Cross-platform (macOS/Linux/Windows).** No platform-specific code touched. The binding is resolved through `keybindingMatchesAction('worktree.navigateUp'/'navigateDown', ...)` with `getShortcutPlatform()`, unchanged by this PR; `Mod+Shift+Arrow` maps to Cmd on macOS and Ctrl elsewhere as before. No paths, shells, or labels involved.

## Security Audit

No input handling, command execution, path handling, auth, secrets, dependency, or IPC surface is touched. The change is list selection over in-memory sidebar rows; no new persisted state and no new external data.

## Notes

This flips an intentional behavior, so it is worth stating the trade-off explicitly: a worktree inside a collapsed group can no longer be reached by arrowing to it. `Cmd+J` (worktree jump palette) still reaches it directly and reveals it, which seems the better tool for "find something I have put away" — while the arrow keys become predictable for stepping through what is actually on screen. Happy to put this behind a setting instead if you would rather keep the current default.

## ELI5

Cycling worktrees with keyboard shortcuts forced collapsed project groups open as you passed through them. Cycling only walks visible rows so groups you collapsed stay collapsed.

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeList.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeList.tsx
@@ -304,11 +304,8 @@ import {
   getVisibleSidebarHostIdSet
 } from './worktree-list-host-filtering'
 import { getFolderWorkspaceCardPrDisplay } from './folder-workspace-card-pr-display'
-import {
-  getPreferredWorktreeRows,
-  getRenderedWorktreesInSidebarOrder
-} from './worktree-sidebar-row-preference'
-import { resolveCycledWorktreeId } from './worktree-keyboard-cycle'
+import { getRenderedWorktreesInSidebarOrder } from './worktree-sidebar-row-preference'
+import { getCyclableWorktreeIds, resolveCycledWorktreeId } from './worktree-keyboard-cycle'
 
 export {
   getScrollTopToRevealBounds,
@@ -681,7 +678,6 @@ type VirtualizedWorktreeViewportProps = {
   worktreeMap: Map<string, Worktree>
   worktreeLineageById: Record<string, WorktreeLineage>
   workspaceLineageByChildKey: Record<string, WorkspaceLineage>
-  repoOrder: Map<string, number>
   // Full canonical repo-id order; must include hidden repos or a reorder silently drops them.
   allRepoIds: string[]
   onReorderHostSections: (orderedHostIds: ExecutionHostId[]) => void
@@ -1367,7 +1363,6 @@ const VirtualizedWorktreeViewport = React.memo(function VirtualizedWorktreeViewp
   worktreeMap,
   worktreeLineageById,
   workspaceLineageByChildKey,
-  repoOrder,
   allRepoIds,
   onReorderHostSections,
   onHostDragActiveChange,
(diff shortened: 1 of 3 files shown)


jeongph:
Heads up on overlap: #10555 rewrites the same block of `navigateWorktree` (two bug fixes — losing the cycling anchor after a workspace closes, and folder workspaces being skipped). Keeping this PR separate on purpose, since the collapsed-group behavior is a judgment call rather than a bug, and it shouldn't be blocked by — or block — those fixes.

Whichever merges first, the other needs a rebase. If #10555 goes in first, this one can shrink to just the `collapsedGroups` argument (plus its dependency-array entry) and drop `worktree-keyboard-cycle.ts`: `worktree-list-keyboard-navigation.ts` there already handles wrap-around and same-end entry, and its anchor resolution filters through the navigable set, so an active workspace hidden inside a collapsed group falls out correctly without extra code.

OrcaWin:
Reviewed with an independent reproduction. **Confirmed in the running app:** with a project group collapsed, one Cmd+Shift+Down on main force-expanded it and jumped inside, destroying the collapse state permanently.

I want to flag the product-direction call explicitly, because it deserved scrutiny: the original all-expanded cycling was *deliberate*, so navigation wouldn't skip worktrees hidden in collapsed groups. This PR reverses that intent.

I think it's the right call, and the reason is that **the old intent was self-defeating** — the reveal step that follows activation uncollapsed the target group anyway, so a collapsed group could never survive navigation in the first place. The new behavior also matches what Cmd+1–9 already does with rendered order, and Cmd+J still reaches anything you've put away. Collapsing a group is the user saying "not now"; that now holds.

**Pushed:** the mechanism was only half right. It kept a second `buildRows()` replay per keypress and just swapped the argument. Cycling now runs over the rows the sidebar has already rendered, so there's no duplicate layout construction on every arrow press.

Cross-platform per AGENTS.md: verified the chord isn't hardcoded to `metaKey` — it resolves to meta on darwin and control elsewhere.

Validation: typecheck 0, sidebar tests green, no introduced failures.

Issue #9441, opened 2026-07-19
[Bug]: macOS app exits during startup when restoring a persisted profile (v1.4.146)


Crosery:
### Operating system

macOS

### Orca version

v1.4.146 (Apple Silicon)

### Details

**Short summary:**
The packaged Orca app exits during startup when restoring an existing persisted profile, while the same app remains running with a fresh user-data directory.

**What happened?**

On macOS 15.7.5 (24G624, arm64), I reproduced the problem with the v1.4.146 app installed at `/Applications/Orca.app`:

- Launching through Finder, `open -a Orca`, `orca open --json`, and the app binary all show the same behavior.
- The app process starts and the Orca runtime reaches `ready`.
- During startup, the main process stays above 100% CPU and its RSS grows to roughly 0.6-1.0 GB in the observed runs.
- The main process then exits after about 14-20 seconds. `orca status --json` reports `not_running` afterward.
- The current runs did not produce a new macOS crash report. The exit happens after renderer bootstrap and while persisted terminal/worktree state is being reconnected.
- Resetting the staged update caches did not change the reproduction. The installed app passes `codesign --verify --deep --strict --verbose=2`.

The affected profile contains a large amount of persisted workspace state, including 326 `worktreeMeta` entries in `orca-data.json` and 58 terminal-history directories. The exact threshold is not known.

**How can we reproduce it?**

1. Start Orca with an existing profile containing many persisted worktrees and terminal tabs.
2. Launch the packaged macOS app.
3. Wait for the initial window/runtime startup to begin restoring the persisted workspaces and terminals.
4. Observe that the process becomes very CPU-intensive and exits within roughly 14-20 seconds.
5. As a control, launch the same app with a fresh profile, for example:

   ```text
   /Applications/Orca.app/Contents/MacOS/Orca --user-data-dir=/tmp/orca-clean-profile
   ```

   With the fresh profile, the process remained alive for more than 50 seconds. After the initial startup burst, CPU dropped close to idle and RSS stayed around 310 MB.

**Expected behavior:**

Orca should remain open and either restore the persisted workspaces successfully or degrade gracefully by staging/deferring restoration and isolating any individual failed session. A restore failure should not silently terminate the whole app.

**Actual behavior:**

Orca reaches runtime readiness, becomes highly CPU-intensive, and exits during startup when using the existing persisted profile. A fresh profile does not reproduce the exit.

**Anything else that might help:**

[...]

PR #10528, opened 2026-07-25
perf(worktree): park the git-common existence poll while the window is hidden

brennanb2025:
## Problem

`startGitCommonNarrowWatch` (`src/main/ipc/worktree-git-common-watch.ts`) is the only watch entry point that never received the `WorktreePollerWindowVisibility` object. Every sibling takes one:

- `startSnapshotDiffPoller` parks on `isWindowVisible()` and re-arms via `onWindowBecameVisible`
- `startGitCommonPolling` (the non-darwin branch) is handed `visibility` and gates the same way

At the call site the two darwin watches start side by side — the primary-metadata poller got `visibility`, the narrow watch did not.

Inside, `armExistencePoll` ran a bare `setInterval` (2s default) that stats `<common>/.git/worktrees` every tick to detect the first linked worktree appearing. Nothing parked it when the window was hidden, so a repo with no linked worktree kept paying ~0.5 stat/sec forever with the app in the background — 0.5×N across N such repos.

**Scope: macOS only.** The narrow watch is the `platform === 'darwin'` branch of `startGitCommonWatch`; every other platform already goes through `startGitCommonPolling`, which is gated.

## Fix

Thread `WorktreePollerWindowVisibility` into `startGitCommonNarrowWatch` and match `startSnapshotDiffPoller`'s existing park/re-arm pattern rather than inventing a second mechanism:

- the poll skips its work and `clearInterval`s itself on the first tick that observes a hidden window (and never arms at all if the window is already hidden)
- `onWindowBecameVisible` re-checks immediately instead of waiting a full interval, then resumes the interval if the dir still isn't there
- the create-event notification is preserved: a `worktrees/` dir that appeared while hidden still upgrades to the native stream **and** emits `{type:'create'}` on show, so the repo's worktree list refreshes
- the visibility listener is unsubscribed in the dispose path (no leak)

The existing subscribe-attempt body is extracted to `tryUpgradeToNarrowWatch` so the interval and the became-visible re-check share one path. No other restructuring.

## Tests

`src/main/ipc/worktree-git-common-watch.test.ts`. The visibility harness now holds a set of listeners rather than one slot — the darwin path parks two independent watches, and a single slot silently dropped the narrow watch's registration. `node:fs/promises#stat` is wrapped with a delegating recorder so a test can assert on stat targets directly rather than on a proxy signal.

New cases:
- `parks the existence poll while the window is hidden` — creates `worktrees/` while hidden, asserts **zero** stats of that path, no subscribe, no events
- `re-checks on show and still reports a worktrees dir created while hidden` — asserts nothing happened while hidden, then that `show()` stats synchronously (promptly, not an interval later), subscribes, and emits the `create`
- `drops both visibility subscriptions on dispose` — listener count 2 → 0
- `resumes polling when the dir is still absent on show` and `keeps polling and reporting while the window stays visible` — visible-path behavior unchanged (these are controls; they pass with and without the fix)

### Revert check

Source change reverted, tests kept — 3 of the 5 new tests fail, and the 2 control tests pass as intended:

```
× parks the existence poll while the window is hidden
    AssertionError: expected [ Array(1) ] to have a length of +0 but got 1
    at worktree-git-common-watch.test.ts:257  (statCalls.filter(path => path === worktreesDir))
× re-checks on show and still reports a worktrees dir created while hidden
× drops both visibility subscriptions on dispose
    AssertionError: expected 1 to be 2

Tests  3 failed | 16 passed (19)
```

With the fix: `Tests 19 passed (19)`.

`pnpm run typecheck` and `pnpm run lint` both exit 0.

### DIFF
--- a/src/main/ipc/worktree-git-common-watch.test.ts
+++ b/src/main/ipc/worktree-git-common-watch.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import { appendFile, mkdtemp, mkdir, realpath, rm, writeFile } from 'node:fs/promises'
+import type * as NodeFsPromises from 'node:fs/promises'
 import { chmodSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
@@ -19,6 +20,20 @@ vi.mock('./parcel-watcher-process', () => ({
   subscribeViaWatcherProcess: vi.fn()
 }))
 
+// Records every stat target so a test can assert which paths a parked poll stopped touching.
+const { statCalls } = vi.hoisted(() => ({ statCalls: [] as string[] }))
+
+vi.mock('node:fs/promises', async (importOriginal) => {
+  const actual = await importOriginal<typeof NodeFsPromises>()
+  return {
+    ...actual,
+    stat: (...args: Parameters<typeof actual.stat>) => {
+      statCalls.push(String(args[0]))
+      return actual.stat(...args)
+    }
+  }
+})
+
 const POLL_MS = 25
 
 const alwaysVisible: WorktreePollerWindowVisibility = {
@@ -30,18 +45,18 @@ function createVisibilityHarness(): {
   source: WorktreePollerWindowVisibility
   hide: () => void
   show: () => void
+  listenerCount: () => number
 } {
   let visible = true
-  let listener: (() => void) | null = null
+  // A set, not a single slot: the darwin path parks two independent watches.
+  const listeners = new Set<() => void>()
   return {
     source: {
       isWindowVisible: () => visible,
       onWindowBecameVisible: (nextListener) => {
-        listener = nextListener
+        listeners.add(nextListener)
         return () => {
-          if (listener === nextListener) {
-            listener = null
-          }
+          listeners.delete(nextListener)
         }
       }
     },
@@ -50,8 +65,11 @@ function createVisibilityHarness(): {
     },
     show: () => {
       visible = true
-      listener?.()
-    }
+      for (const listener of listeners) {
+        listener()
+      }
+    },
+    listenerCount: () => listeners.size
   }
 }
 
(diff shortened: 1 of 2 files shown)


brennanb2025:
**Re: CodeRabbit — "Consider swallowing rejections on the `void`-ed upgrade promise" (`worktree-git-common-watch.ts:211-221`)**

Declining, with reasoning.

The observation is technically correct: `tryUpgradeToNarrowWatch()` can only reject if the `onEvents` consumer throws (`trySubscribe` catches everything itself and returns a boolean), and `.finally()` does propagate that rejection onto a promise that `void` leaves unhandled. But three things make the one-line `.catch(() => {})` the wrong change *here*:

1. **The consequence is a log line, not a crash.** The main process installs a global handler at `src/main/startup/main-process-error-guards.ts:127` (`installUnhandledRejectionLogging`) that records the rejection via `recordFatalMainProcessError` and deliberately keeps the app alive — that guard exists precisely so one failed background promise cannot kill the app (issue #9441). So this is not the Node-default process-exit hazard the suggestion implies.

2. **It is pre-existing, not introduced here.** The pre-PR code was structurally identical:
   ```js
   void trySubscribe()
     .then((installed) => { ... onEvents([{ type: 'create', path: worktreesDir }]) })
     .finally(() => { subscribing = false })
   ```
   An `onEvents` throw there rejected an equally unhandled `void`-ed chain. This PR moved that body into `tryUpgradeToNarrowWatch`; it did not create the hazard.

3. **Patching only this site would be misleading.** The identical `void tryUpgradeToNarrowWatch()` call on the interval path (line 206) has exactly the same exposure and is untouched by the suggested diff. Guarding one of the two would read as "handled" while leaving the other live. If this is worth fixing it should be fixed for both call sites — and, consistently, for the sibling `startSnapshotDiffPoller`'s `void tick()` at lines 132 and 135 — which is a separate change, not something to fold into a perf PR scoped to parking one poll.

Behavior is unaffected either way: `armExistencePoll()` in the `.finally` still runs on the rejecting path, and `subscribing` is reset by the `finally` block inside the async function, so no state is left corrupted.

Leaving as-is for scope. Verified alongside this: `pnpm run typecheck` and `pnpm run lint` exit 0, the max-lines ratchet reports no new bypasses, and CI `verify` is green.

PR #10529, opened 2026-07-25
fix(runtime): reserve long-poll headroom so orchestration.ask can't starve terminal.wait

brennanb2025:
## Summary

`orchestration.ask` was added to `isLongPollRequest()` so it would get keepalive framing and the abort signal. Both are genuinely needed and **are kept unchanged here** — but joining that set also opted `ask` into the admission accounting, and that part starves other clients:

- `activeLongPolls` is a **single server-wide counter**, not per-connection.
- `LONG_POLL_CAP = 16`; overflow is shed with `runtime_busy`.
- `ask` defaults to a **600 s** timeout and blocks that whole time waiting on a reply.
- `AskParams.timeoutMs` is `OptionalFiniteNumber` with **no upper bound**.

So 16 workers sitting in blocking asks hold every long-poll slot for ten minutes each — or forever, if a caller passes a large `timeoutMs` — and `terminal.wait` / `orchestration.check --wait` get `runtime_busy` for **every** other client sharing the runtime: mobile, web, CLI, SSH, relay. Before `ask` was long-polled it consumed no slot at all, so this is fresh in shipped behavior.

**The fix, in two parts:**

**1. Class-based sub-cap.** `isLongPollRequest` becomes `longPollClassOf`, returning `'ask' | 'wait' | null`. `ask` is metered by its own counter against `askLongPollCap = floor(longPollCap / 2)` (8 in production), on top of the existing total cap. `terminal.wait` and `check --wait` keep the full 16 — the sub-cap only fences `ask`. The cap is **derived**, not a new constructor option, so the reservation holds for whatever `longPollCap` a caller picks and can't be configured away. `LONG_POLL_CAP` was deliberately not raised — that just moves the cliff.

**2. Timeout clamp.** `clampAskTimeoutMs` bounds the caller-supplied value at 30 min (3× the default) at the handler boundary. An unbounded caller-supplied timeout on a shared, capped resource is its own defect, independent of (1). The largest `timeoutMs` any caller passes to `ask` specifically is 600 000 ms, hardcoded in the injected worker preamble (`src/main/runtime/orchestration/preamble.ts:111`) — the 900 000 ms values elsewhere are `check --wait`, a different method. So the clamp keeps 3x headroom over real ask usage. The CLI derives its transport timeout from the value it *sent*, so a clamped server simply answers `timedOut` first — no client change needed.

Admission and release are now one helper each (`admitLongPoll` / `releaseLongPoll`), shared by the Unix-socket and WebSocket paths, so the two transports can't drift. Overflowing asks get `runtime_busy` with a distinct message (`orchestration.ask capacity reached`) but the **same error code**, so existing client backoff is untouched.

**Why not per-connection:** there is no per-connection long-poll state to reuse — `activeLongPolls` is the only meter, and the WS path doesn't thread a connection id into admission. It would also be the wrong fence: the starving party is a *different client* (mobile/relay), not a second request on the same socket.

## Screenshots

No visual change — this is runtime RPC admission control with no UI surface.

## Testing

- [x] `pnpm lint` — clean
- [x] `pnpm test` — both touched files green (175 passed)
- [ ] `pnpm build` — not run; no build-surface change (no new deps, no main/preload entry change)

`src/main/runtime/runtime-rpc.test.ts`
- **reserves long-poll headroom for terminal.wait when orchestration.ask floods** — cap 4 (sub-cap 2), 4 concurrent asks; asserts a `terminal.wait` and a `check --wait` from other clients are still admitted, and that the 2 overflow asks were shed.
- **keeps the full cap available to terminal.wait and check --wait** — guard: 4 `check --wait` still fill the whole budget, 5th is `runtime_busy`, ask counter stays 0.
- **applies the ask sub-cap on the WebSocket path and releases both counters on close** — closes the WS coverage gap (see AI Review Report). Drives `handleWebSocketMessage` with a `runtime`-scoped device, asserts the overflow ask is shed with the distinct message without burning a reserved slot, that `check --wait` still gets the other half, and that **both** counters return to zero when the socket closes.

`src/main/runtime/rpc/methods/orchestration.test.ts`
- **clamps an absurd caller-supplied timeoutMs** — behavioral: spies the budget actually handed to `waitForMessage`.
- **clamps timeoutMs at the exported boundary** — table over default / passthrough / at-limit / over-limit / negative.

### Revert check — every new test was run against reverted source

Each source file was reverted independently (`git apply -R`) with the tests kept, to prove no test is vacuous.

| test | with fix | source reverted |
|---|---|---|
| reserves long-poll headroom | pass | **fail** — `terminal.wait` got `code: "runtime_busy"`, expected `"timeout"` |
| applies the ask sub-cap on the WebSocket path | pass | **fail** — overflow ask was admitted instead of shed; replies contained `req_ask_0/1/overflow` all `ok: true` |
| clamps an absurd timeoutMs | pass | **fail** — `expected 9007199254740991 to be less than or equal to 1800000` |
| clamps at the exported boundary | pass | **fail** — `TypeError: clampAskTimeoutMs is not a function` |
| keeps the full cap available | pass | fail *only* on the `activeAskLongPolls === 0` line (`expected undefined to be +0`) |

The first row is the starvation itself, reproduced: with the source reverted, four asks eat all four slots and `terminal.wait` is shed. The last row is reported honestly as a **no-regression guard**, not a bug reproducer — its admission assertions (4 waits admitted, 5th `runtime_busy`) hold in both worlds by design, and only the new counter reference fails pre-fix.

## AI Review Report

Reviewed with Claude, focused on four risks:

1. **Keepalive / abort regression.** The pre-existing wiring that put `orchestration.ask` in the long-poll set was deliberate — ask needs keepalive or the 30 s socket idle timer tears it down, and it needs the abort signal to release the waiter on client disconnect. Verified the `boolean → 'ask' | 'wait' | null` change preserves this: `'ask'` is truthy, so `context?.startKeepalive()` and `signal: longPoll ? context?.signal : undefined` behave identically. `isLongPollRequest` had no other callers anywhere in the repo.
2. **Slot leaks.** A leaked slot in the new ask class is permanent capacity loss that only a restart clears. `releaseLongPoll` runs in the `finally` of both transports, covering normal return, throw, abort and disconnect. The new WS test asserts both counters return to zero after a socket close. Rejected requests never increment, so shedding cannot leak.
3. **Unchanged admission for existing clients.** The total-cap rejection message is byte-identical to `main`, and no client in the repo (including mobile and relay) string-matches on the message — only on `code: 'runtime_busy'`.
4. **Cross-platform / SSH / relay / mobile.** No platform-dependent code paths: no `path` handling, no shell invocation, no keyboard or Electron-menu surface, so macOS / Linux / Windows behave identically. SSH, relay and mobile clients share this runtime and cap, and are the *beneficiaries* — `MOBILE_RPC_METHOD_ALLOWLIST` contains `terminal.wait` but no `orchestration.*`, so a mobile device can only ever use the reserved `wait` class and can no longer be starved by a fleet of asking workers. Folder workspaces and non-GitHub git providers are not involved — nothing here touches git.

**Flagged and fixed during review:** CodeRabbit and the review both flagged that the WebSocket path shared the new admit/release helpers but was covered only by reasoning, not by a running test. That is now closed by the third `runtime-rpc.test.ts` test above, which fails behaviorally when the source is reverted.

## Security Audit

- **Input handling:** the clamp *narrows* attacker-controlled input. `timeoutMs` reaches the handler through `OptionalFiniteNumber` (finite numbers only), and `clampAskTimeoutMs` bounds it to `[0, 1_800_000]`. Values under the cap and `undefined` are passed through unchanged, so no behavior change for well-formed callers.
- **Resource exhaustion:** this is the actual security-relevant change — it removes an unauthenticated-by-volume DoS in which one client class could exhaust a shared server-wide budget and deny service to every other paired device. The sub-cap is derived rather than configurable specifically so it cannot be turned off.
- **Auth / IPC:** no change to auth-token enforcement, device-token validation, the mobile method allowlist, or E2EE. Admission is still evaluated *after* authentication and after the mobile allowlist check on the WS path.
- **Command execution / path handling / secrets / dependencies:** none touched; no new dependencies.
- **Follow-up:** none blocking.

## Notes

**Intended trade — the concurrent-ask ceiling drops 16 → 8.** More than 8 workers blocking on `ask` at once now get `runtime_busy` where previously all 16 were admitted. That is the point of the reservation, but it is a real behavior change worth stating: nothing in `src/` retries on `runtime_busy`, so it surfaces as a hard CLI error. It is self-mitigating in practice — the callers are LLM agents and the message says "retry with backoff" — and shedding happens at the admission fence *before* the handler runs, so a shed ask inserts no orphan `decision_gate` row.

**Clamped asks are observable.** The ask result now echoes the effective `timeoutMs` on every return, and the CLI prints that rather than the value it sent. Without it a caller passing `--timeout-ms 3600000` was told `ask timeout after 3600000ms` after only 30 min — off by 2x, and accurate before this PR introduced the clamp. Additive optional field; older clients fall back to the requested value.

**Known follow-up, deliberately out of scope:** `orchestration.check --wait` also takes an unbounded caller `timeoutMs` and holds a slot for it. Same class of defect, different method — left alone to keep this PR on one concern. Note the reserved half is shared by `terminal.wait` *and* `check --wait`, so a coordinator fleet passing a huge `check --wait` timeout can still fill the budget; that is a symmetric pre-existing hole, and `check --wait` is coordinator-driven (few) where `ask` is worker-driven (many).

### DIFF
--- a/src/main/runtime/rpc/methods/orchestration.test.ts
+++ b/src/main/runtime/rpc/methods/orchestration.test.ts
@@ -1,6 +1,6 @@
 /* eslint-disable max-lines -- Why: orchestration tests share a mock runtime factory; splitting by method would duplicate 40 lines of setup per file without improving clarity. */
 import { afterEach, describe, expect, it, vi } from 'vitest'
-import { ORCHESTRATION_METHODS } from './orchestration'
+import { ORCHESTRATION_METHODS, clampAskTimeoutMs } from './orchestration'
 import { RpcDispatcher } from '../dispatcher'
 import { buildRegistry, type RpcContext, type RpcRequest } from '../core'
 import { OrchestrationDb } from '../../orchestration/db'
@@ -1790,6 +1790,46 @@ describe('orchestration RPC methods', () => {
 
       expect(result.timedOut).toBe(false)
       expect(result.answer).toBe('correct answer')
+    })
+
+    it('clamps an absurd caller-supplied timeoutMs so the long-poll slot is bounded', async () => {
+      setup()
+      vi.spyOn(runtime, 'deliverPendingMessagesForHandle').mockImplementation(() => {})
+      vi.spyOn(runtime, 'notifyMessageArrived').mockImplementation(() => {})
+      let observedTimeoutMs: number | undefined
+      vi.spyOn(runtime, 'waitForMessage').mockImplementation(async (_handle, options) => {
+        observedTimeoutMs = options?.timeoutMs
+        // End the wait loop so the assertion runs against the first budget slice.
+        const outbound = db.getInbox(10).find((m) => m.type === 'decision_gate')
+        if (outbound) {
+          db.insertMessage({
+            from: 'term_coord',
+            to: 'term_worker',
+            subject: 'Re: Question',
+            body: 'ok',
+            threadId: outbound.id
+          })
+        }
+      })
+
+      await call('orchestration.ask', {
+        from: 'term_worker',
+        to: 'term_coord',
+        question: 'forever?',
+        timeoutMs: Number.MAX_SAFE_INTEGER
+      })
+
+      expect(observedTimeoutMs).toBeLessThanOrEqual(1_800_000)
+      expect(observedTimeoutMs).toBeGreaterThan(1_700_000)
+    })
+
+    it('clamps timeoutMs at the exported boundary', () => {
+      expect(clampAskTimeoutMs(undefined)).toBe(600_000)
+      expect(clampAskTimeoutMs(1_000)).toBe(1_000)
+      expect(clampAskTimeoutMs(1_800_000)).toBe(1_800_000)
+      expect(clampAskTimeoutMs(86_400_000)).toBe(1_800_000)
+      expect(clampAskTimeoutMs(Number.MAX_SAFE_INTEGER)).toBe(1_800_000)
+      expect(clampAskTimeoutMs(-5)).toBe(0)
     })
 
     it('parses options CSV with whitespace and empty entries', async () => {
--- a/src/main/runtime/rpc/methods/orchestration.ts
+++ b/src/main/runtime/rpc/methods/orchestration.ts
@@ -30,6 +30,19 @@ const TASK_STATUSES: TaskStatus[] = [
   'blocked'
 ]
 
+const ASK_DEFAULT_TIMEOUT_MS = 600_000
+
+// Why: ask pins a shared long-poll slot for its entire timeout, so a caller-supplied
+// value can't be unbounded; 30 min covers a slow human gate and still frees the slot.
+const ASK_MAX_TIMEOUT_MS = 1_800_000
+
+export function clampAskTimeoutMs(timeoutMs: number | undefined): number {
+  if (timeoutMs === undefined) {
+    return ASK_DEFAULT_TIMEOUT_MS
+  }
+  return Math.min(Math.max(0, timeoutMs), ASK_MAX_TIMEOUT_MS)
+}
+
 function getLifecycleGroupRecipientError(type: 'worker_done' | 'heartbeat'): string {
   return `${type} messages must be sent to a concrete coordinator terminal handle, not a group address.`
 }
@@ -567,7 +580,7 @@ export const ORCHESTRATION_METHODS: RpcMethod[] = [
 
       const db = runtime.getOrchestrationDb()
       const from = params.from ?? 'unknown'
-      const timeoutMs = params.timeoutMs ?? 600_000
+      const timeoutMs = clampAskTimeoutMs(params.timeoutMs)
       const options =
         params.options
           ?.split(',')
--- a/src/main/runtime/runtime-rpc.test.ts
+++ b/src/main/runtime/runtime-rpc.test.ts
@@ -4289,6 +4289,166 @@ describe('OrcaRuntimeRpcServer', () => {
       }
     })
 
+    it('reserves long-poll headroom for terminal.wait when orchestration.ask floods', async () => {
+      const userDataPath = mkdtempSync(join(tmpdir(), 'orca-runtime-rpc-'))
+      const runtime = new OrcaRuntimeService()
+      const db = new OrchestrationDb(':memory:')
+      runtime.setOrchestrationDb(db)
+      // Why: cap 4 → ask sub-cap 2, so 4 concurrent asks can only take half the budget.
+      const server = new OrcaRuntimeRpcServer({
+        runtime,
+        userDataPath,
+        keepaliveIntervalMs: 1000,
+        longPollCap: 4
+      })
+      runtime.attachWindow(1)
+      runtime.syncWindowGraph(1, {
+        tabs: [
+          {
+            tabId: 'tab-1',
+            worktreeId: 'repo-1::/tmp/worktree-a',
+            title: 'Terminal 1',
+            activeLeafId: 'pane:1',
+            layout: null
+          }
+        ],
+        leaves: [
+          {
+            tabId: 'tab-1',
+            worktreeId: 'repo-1::/tmp/worktree-a',
+            leafId: 'pane:1',
+            paneRuntimeId: 1,
+            ptyId: 'pty-1'
+          }
+        ]
+      })
+      await server.start()
+
+      const asks: ReturnType<typeof openFramedSession>[] = []
+      try {
+        const metadata = readRuntimeMetadata(userDataPath)
+        const endpoint = metadata!.transports[0]!.endpoint
+        const listResponse = await sendRequest(endpoint, {
+          id: 'req_list',
+          authToken: metadata!.authToken,
+          method: 'terminal.list'
+        })
+        const handle = (listResponse.result as { terminals: { handle: string }[] }).terminals[0]!
+          .handle
+
+        // Four workers block in ask; distinct `from` handles so no reply wakes another.
+        for (let i = 0; i < 4; i++) {
+          asks.push(
+            openFramedSession(endpoint, {
+              id: `req_ask_${i}`,
+              authToken: metadata!.authToken,
+              method: 'orchestration.ask',
+              params: {
+                from: `term_w${i}`,
+                to: 'term_coord',
+                question: 'proceed?',
+                timeoutMs: 10_000
+              }
+            })
+          )
+        }
+        // Let every ask reach the admission fence before probing the reserved half.
+        await waitFor(() => server['activeLongPolls'] >= 2)
+        await sleep(100)
+
+        // The reserved half still admits a terminal.wait from any other client.
+        const admitted = openFramedSession(endpoint, {
+          id: 'req_terminal_wait',
+          authToken: metadata!.authToken,
+          method: 'terminal.wait',
+          params: { terminal: handle, for: 'tui-idle', timeoutMs: 50 }
+        })
+        await admitted.done
+        expect(admitted.frames.find((f) => f.ok !== undefined)).toMatchObject({
+          id: 'req_terminal_wait',
+          ok: false,
+          error: { code: 'timeout' }
+        })
+
+        // …and a check --wait too, which shares the same reserved class.
+        const check = openFramedSession(endpoint, {
+          id: 'req_check_wait',
+          authToken: metadata!.authToken,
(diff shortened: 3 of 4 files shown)


brennanb2025:
## CodeRabbit review response

Both items actioned — none declined.

### 1. Nitpick: "Missing WebSocket-transport coverage for the ask/wait admission split" — **ACTIONED** in a994fd9eaa

Agreed, and this was the highest-value thing in the review. The PR disclosed the gap but disclosure isn't coverage: `admitLongPoll`/`releaseLongPoll` are shared by both transports, so a WS-only regression would have shipped silently.

Added **`applies the ask sub-cap on the WebSocket path and releases both counters on close`** (`src/main/runtime/runtime-rpc.test.ts`). It drives `handleWebSocketMessage` directly and asserts:

- two asks are admitted, the third is shed with `code: 'runtime_busy'` **and** the distinct `orchestration.ask capacity reached; retry with backoff` message;
- shedding does **not** burn a slot from the reserved half (`activeLongPolls` stays at 2);
- a `check --wait` is still admitted afterwards, i.e. the reservation actually holds on this transport;
- **both** counters return to zero after `ws.emit('close')` — this is the WS-path leak check, since a leaked ask slot is permanent capacity loss until restart.

Two implementation notes for anyone reading the test:

- It registers the device with **`runtime` scope, not `mobile`**. `MOBILE_RPC_METHOD_ALLOWLIST` contains `terminal.wait` but no `orchestration.*`, so a mobile-scoped device is rejected with `forbidden` before it ever reaches the admission fence. A mobile-scoped test would have passed vacuously.
- The test gates on the pre-existing `activeLongPolls` total rather than the new `activeAskLongPolls` field. That ordering matters: gating on the new field makes the test fail on revert merely because a private field is missing. Gating on the old one makes it fail on the **behavior** — verified by reverting `runtime-rpc.ts` and re-running:

  ```
  AssertionError: expected [ { id: 'req_ask_0', …(3) }, …(2) ] to deep equally contain ObjectContaining{…}
  - Expected: ObjectContaining { "error": ObjectContaining { "code": "runtime_busy",
      "message": "orchestration.ask capacity reached; retry with backoff" },
      "id": "req_ask_overflow", "ok": false }
  + Received: [ { "id": "req_ask_0", "ok": true, ... }, { "id": "req_ask_1", ...
  ```

  i.e. pre-fix the third ask is admitted instead of shed — the starvation reproduced on the WebSocket transport.

### 2. Pre-merge check: "Description check ⚠️ — does not follow the required template" — **ACTIONED**

The description has been rewritten against `.github/pull_request_template.md` with all mandatory sections: Summary, Screenshots (`No visual change`), Testing checklist, AI Review Report (including the explicit macOS/Linux/Windows cross-platform statement), Security Audit, and Notes. The revert-check table now also covers the new WebSocket test.

---

For completeness, the rest of the review verification (not requested by CodeRabbit, recorded here because it's the load-bearing part of the change):

- **Every** new test was run against independently reverted source and confirmed to fail — no vacuous tests. Table is in the PR description.
- The `boolean → 'ask' | 'wait' | null` refactor preserves keepalive and abort wiring for `orchestration.ask` (`'ask'` is truthy at both call sites); `isLongPollRequest` had no other callers in the repo.
- The total-cap rejection message is byte-identical to `main`, and nothing in the repo string-matches on the message — only on `code: 'runtime_busy'` — so existing client backoff is unaffected.

brennanb2025:
## Follow-up after independent review — one more fix pushed (`6267b735d4`)

Two independent reviewers both flagged the same defect in the clamp, which I had initially judged as documentable-but-not-worth-fixing. On reflection they were right and I've fixed it, because **this PR introduced the inaccuracy**.

### The clamp was silent, and it made an existing message factually wrong

`src/cli/handlers/orchestration.ts` computed its display value locally and un-clamped, and the ask result carried no timeout figure — so `--json` consumers had no channel either.

A worker running `ask --timeout-ms 3600000` gets clamped to 1 800 000, the server returns `timedOut: true` at 30 min, and the CLI printed `ask timeout after 3600000ms`. **Wrong by 2×.** A worker agent reads that as "I waited a full hour, the coordinator is dead" and escalates or abandons, when only 30 minutes elapsed and re-asking is correct. That message was accurate before this PR added the clamp.

**Fix:** echo the effective `timeoutMs` on every `orchestration.ask` return and print that instead. Additive optional field, so older clients fall back to the requested value; no `PROTOCOL_VERSION` bump, since that constant governs the daemon PTY protocol rather than RPC method shapes.

Verified non-vacuous by mutation — removing *only* the echo while leaving the clamp intact fails the new assertion:

```
AssertionError: expected undefined to be 1800000 // Object.is equality
 ❯ src/main/runtime/rpc/methods/orchestration.test.ts:1825:32
    1825|       expect(result.timeoutMs).toBe(1_800_000)
```

Also added `expect(outbound).toBeDefined()` to that test's `waitForMessage` mock. Without it, if the fixture ever drifts the handler's `while (true)` spins on a synchronously-resolving mock until vitest times out — a hang rather than a clean failure.

### Two corrections to the PR description

- **The clamp has more headroom than originally stated.** The 900 000 ms figure cited is a `check --wait`, not an `ask`. The largest value any caller passes to `ask` specifically is 600 000 ms, hardcoded in the injected worker preamble (`src/main/runtime/orchestration/preamble.ts:111`) — so the 30-min ceiling keeps 3× headroom over real usage.
- **The concurrent-ask ceiling drops 16 → 8**, and nothing in `src/` retries on `runtime_busy`. That is the intended trade, but it is a real behavior change and now says so under Notes.

### On the WebSocket coverage gap — it was load-bearing, not theoretical

Worth recording why CodeRabbit's item mattered more than it first appeared. `orchestration.ask` **is** reachable over WebSocket in production: `DeviceScope` is `'mobile' | 'runtime'`, and `'runtime'` bypasses the mobile allowlist at two production pairing sites — `src/main/index.ts:1577`, where `orca serve` **defaults** to `'runtime'` (you only get `'mobile'` with `--mobile-pairing`), and `src/main/ipc/mobile.ts:170` for the web client, with an in-repo comment stating runtime clients intentionally get full access. So the WS path serves exactly the SSH/headless/web clients this fix protects.

(Mobile-scoped devices genuinely cannot reach `ask` — they're rejected `forbidden` before admission — which is why the new WS test registers a `runtime`-scoped device. A mobile-scoped test would have passed vacuously.)

### Verification on this commit

`npm run typecheck` exit 0 · `npm run lint` exit 0, no findings in any touched file · `check:max-lines-ratchet` → "354 grandfathered suppression(s), no new bypasses" · 221 tests pass across `orchestration.test.ts`, `runtime-rpc.test.ts` and `cli/handlers/orchestration.test.ts`.

### Left alone deliberately

The fixed `sleep(100)` in the headroom test was stress-tested 14/14 (8 sequential + 6 concurrent under load) without reproducing a flake, so I left it rather than churn a passing test on speculation. `check --wait`'s unbounded timeout stays out of scope as a follow-up.

brennanb2025:
### ⚠️ GitHub has not picked up the latest commit — CI has NOT run on it

`6267b735d4` ("report the effective ask timeout") **is** on the branch — `git ls-remote` confirms `refs/heads/brennanb2025/p2-longpoll-ask-starvation` → `6267b735d4cd825f14283431322ccc3b8e0c35c5` — but this PR object still reports `head.sha = a994fd9eaa` and `commits = 2` roughly 45 minutes after the push. The `synchronize` webhook appears to have been dropped; githubstatus.com reports all systems operational.

I tried re-emitting the event by rewinding the ref to `a994fd9eaa` and re-advancing it to `6267b735d4` (both pushes were accepted as real ref updates). The PR head still did not move, so **no CI run exists for `6267b735d4` and I could not force one.**

**State of CI, stated precisely:**
- `verify` **pass** on `a994fd9eaa` — that covers the first two commits (the fix and the WebSocket test).
- `6267b735d4` is **CI-unverified**. It was verified locally only: `npm run typecheck` exit 0, `npm run lint` exit 0 with no findings in any touched file, `check:max-lines-ratchet` clean ("354 grandfathered suppression(s), no new bypasses"), and 221 tests passing across `orchestration.test.ts`, `runtime-rpc.test.ts` and `cli/handlers/orchestration.test.ts`.

Please confirm CI goes green on `6267b735d4` before merging — closing and reopening the PR, or a rebase push, would normally force the re-sync. I did not close/reopen since that would churn review state on a PR that already has a completed CodeRabbit review.

PR #9936, opened 2026-07-22
fix(agent-status): cap the live agent-status map to stop Windows renderer OOM (#9872)

OrcaWin:
## Summary

Fixes #9872.

The live `agentStatusByPaneKey` map had no size cap. Missed pane teardown events could leave heavy rows behind indefinitely, and every status ping spread-copied the entire map, eventually pushing the Windows renderer past V8's heap limit.

This change:

- caps live agent-status rows at 500 by evicting panes that are provably gone, then stale or unprovable rows when needed;
- never evicts rows for leaves present in a rooted live layout;
- purges evicted pane keys from the related sleeping-session and launch-config maps;
- bumps status and sort epochs when eviction changes the visible map;
- adds regression coverage for dead-pane churn, live/rootless/headless panes, stale rows, hard-cap behavior, related-map cleanup, and eviction-only epoch updates.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — changed-file oxlint and the max-lines ratchet pass; the repository-wide command reaches an unrelated pre-existing switch-exhaustiveness error in `src/renderer/src/components/skills/skill-freshness-group.tsx:101`.
- [x] `pnpm run typecheck:web`
- [ ] `pnpm test` — focused renderer-store and agent-status suites were run instead.
- [ ] `pnpm build` — no build-surface change.
- [x] Added or updated high-quality tests that would catch regressions.

Focused results:

- `agent-status-live-map-leak.test.ts`: 9/9
- Agent-status and `use-tab-agent` suites: 192/192 across 15 files
- Full renderer store suite: 1,901/1,901 across 121 files
- Changed-file oxlint: clean
- Max-lines ratchet: clean

## AI Review Report

The review traced liveness classification, status-map ownership, teardown cleanup, stale-state handling, hard-cap fallback, epoch invalidation, and recovery-map persistence. It flagged two issues that were fixed: evicted pane keys now also clear `sleepingAgentSessionsByPaneKey` and `agentLaunchConfigByPaneKey`, and the epoch test now uses a same-state ping to an existing row so eviction is the only sort-relevant change.

Cross-platform and remote-workspace behavior were explicitly checked. The change is renderer-only in-memory state management with no keyboard shortcuts, labels, filesystem paths, shell behavior, Git commands, or Electron platform branches; it behaves identically on macOS, Linux, Windows, local, WSL, and SSH-backed workspaces.

## Security Audit

[...]

Issue #9872, opened 2026-07-22
[Bug]: Windows renderer OOM crash (exit -36861) after upgrading to 1.4.149 — heap climbs to limit within ~7 min under multi-agent + 5+ terminal tabs


Chang-Jin-Lee:
## Summary

Renderer process crashes (`reason: crashed`, exit code `-36861`) repeatedly, multiple times in one day, starting immediately after updating to **1.4.149**. Previously stable on the prior version. Renderer JS heap climbs steadily to its limit over ~7 minutes and stays pinned there for ~3 minutes before the process is killed — this looks like genuine JS heap exhaustion, not just a native/GPU crash sharing the same exit code (see "Related reports" below).

## What happened / reproduction context

- 5+ terminal tabs open
- Multiple AI agent sessions running concurrently (Claude, Grok)
- Crash reproduced several times today, all after updating to 1.4.149; did not happen on the prior version with the same kind of workload

## Evidence — full crash report (Help ▸ Report Crash ▸ Copy)

```
[Crash Report]

Report ID: 0f04149d-581e-48ec-abe8-5754f5909d47
Created: 2026-07-22T05:46:22.405Z
Status: dismissed
Source: renderer
Process: renderer
Reason: crashed
Exit code: -36861
App version: 1.4.149
Platform: win32 10.0.22631 x64
Electron: 43.1.0
Chrome: 150.0.7871.47

Details:
- processType: renderer
- mainProcessPid: 27412
- mainProcessLaunchId: d93a80f8-8324-4a77-9495-e6b8eaa19d47
- mainProcessStartedAt: 2026-07-22T01:43:14.436Z
- processMetricsCount: 3
- processMetricsBrowserCount: 1
- processMetricsBrowserWorkingSetMB: 472
- processMetricsRendererCount: 0
- processMetricsRendererWorkingSetMB: 0
- processMetricsGpuCount: 1
- processMetricsGpuWorkingSetMB: 584
- processMetricsUtilityCount: 1
- processMetricsUtilityWorkingSetMB: 174
- processMetricsOtherCount: 0
- processMetricsOtherWorkingSetMB: 0
- processMetricsLargestPid: 22308
- processMetricsLargestType: GPU
- processMetricsLargestWorkingSetMB: 584

Recent activity:
[...]

PR #10533, opened 2026-07-25
refactor(memory-safety): scope bounds to wired ingress, add doctrine + ratchet

nwparker:
## Scope memory bounds to wired ingress, and make the doctrine enforceable

Follow-up to the OOM foundation (#10294 / #10295 / #10299). An audit of what actually landed found the bounds had shipped **ahead of the code that uses them**:

- **13 of the landed modules had no production consumer** — their callers live in slices that never merged.
- **`BoundedMap` had none at all**, while the two byte-identical `setBoundedMapEntry` helpers it was written to replace were still duplicated in the tree.

Net: **−863 lines.**

### The doctrine

> **Bound at ingress, not in the interior.**

The real crash (#9872) was one unbounded `Map` that also spread-copied itself per update; it was fixed in **2 files / +422 lines** (#9936). The follow-up bounded ~1,600 files prophylactically. Ceilings belong where untrusted or unpredictably-large data *enters* a heap — files, transport, parsing, enumeration, long-lived user-keyed accumulators. A second ceiling behind an existing one isn't defense in depth; it's another thing that can be wrong, and every ceiling is a new user-visible failure mode.

### What this does

1. **Deletes the 13 unwired modules.** They return with the slice that consumes them, not before.
2. **Migrates three real caches onto `BoundedMap`** — `secure-path-hardening-cache`, `github/conflict-summary-cache`, `runtime/orca-runtime` — and deletes both duplicate helpers, so the primitive is now net-negative on lines instead of net-positive.
3. **Moves the six generic primitives to `src/shared/memory-safety/`** with a `README.md` stating the doctrine, when *not* to add a ceiling, and the rules for one (far above real use; fail closed; explicit overload behavior; release on every exit path; test boundary and boundary+1). Domain-specific limits stay with their domain.
4. **Adds `check:bounded-ingress`** — a baseline-locked ratchet in `pnpm lint`. New unbounded `readFile` / response body / `JSON.parse` / `Promise.all(map)` fails CI unless it uses a memory-safety helper or carries a `bounded-by:` justification. 724 existing occurrences are frozen; the check is what actually binds the next contributor, since a README alone won't.

### Verification

- `pnpm typecheck`, `oxlint`, `check:max-lines-ratchet`, `check:bounded-ingress` — all pass
- Full suite: **35,983 passed**. The 2 failures in `src/relay/agent-exec-handler.test.ts` are a pre-existing local-env artifact (`AGENT_BROWSER_PROFILE`) that fails identically on unmodified `main`.
- `BoundedMap`: 26 unit tests + a **72,000-op randomized invariant fuzz across 9 configs** asserting the ledger equals the live sum and both ceilings hold after every operation.
- Ratchet: 5 unit tests, plus verified end-to-end that it catches a new unbounded read and honors a `bounded-by:` escape.

The remaining ~1,880-file changeset is being re-scoped against this doctrine before any of it is proposed.

---
🤖 Assembled by Claude Code.


nwparker:
## ⛔ Do not merge — this branch reverts shipped work

A pre-merge audit found that this branch silently reverts commits already on `main`. Root cause: it was built by checking out whole files from the old #10179 stack rather than merging them onto current `main`, so where `main` had moved on, the stale copy won.

**Verified:** comparing every file both `main` and this branch changed since their merge base, and counting `main`-added lines missing from this branch:

- **2 of 2** overlapping files lose `main`'s work — `orca-runtime.ts` −21 lines, `pull-request-generation.ts` −2

**CI cannot catch this**: a reverted feature still typechecks and its old tests still pass. Green CI proves internal consistency, not that shipped work survived.

**Fix:** rebuild by *merging* each file onto current `main` instead of overwriting, then assert no commit reachable from the merge base has been reverted.

Full analysis, including 5 separate must-fix correctness/data-loss findings, in `OOM_MERGE_REPORT.md`.

nwparker:
Superseded by #10548, which now targets `main` directly and fully subsumes this PR.

Verified: every file this PR changed is present in `oom-rescoped` — 52 byte-identical, 24 legitimately extended, **0 missing**. The 24 differences are all cases where this PR *deleted* an unwired module and #10548 restores it **with its real consumers** (e.g. `filesystem-directory-listing-limit.ts` now has 13, `custom-pet-media-limits.ts` has 3) — which is the bound-at-ingress doctrine working as intended.

The doctrine, `src/shared/memory-safety/`, the README, the bounded-ingress ratchet, the new no-merge-base-revert guard, and the `BoundedMap` migrations all ship in #10548.

PR #10538, opened 2026-07-25
fix(codex): treat non-Codex transcript as stale resume metadata

### LINKED ISSUE #10517: Codex resume guard hard-fails on panes mislabeled `agent: "codex"`
Orca 1.4.150 · macOS 26.3 (arm64) · Claude Code 2.1.220 · codex-cli 0.144.2

## Symptom

A Claude Code pane refuses to auto-resume:

> Orca could not verify the originating Codex session file, so automatic resume was stopped to avoid using a different account.

## Cause

The pane's `sleepingAgentSessionsByPaneKey` record has `agent: "codex"` with a Claude `providerSession` (ids and project name redacted):

```json
"<tab-id>:<leaf-id>": {
  "agent": "codex",
  "providerSession": {
    "id": "<claude-session-uuid>",
    "transcriptPath": "~/.claude/projects/<project>/<claude-session-uuid>.jsonl"
  },
  "terminalTitle": "⠂ Claude Code"
}
```

That pane was a Claude session running `codex exec` as a background command. Something relabeled `agent` without updating `providerSession`.

Resume then routes through `prepareCodexSessionResumeForLaunch`. `resolveTrustedCodexSessionResume` requires `transcriptPath` to sit under `<CODEX_HOME>/sessions/` and match the rollout pattern, which a path under `~/.claude/projects/` never does. `findTrustedCodexSessionResume` returns `null` before the session-id rescan, since that fallback is gated on `transcriptPath` being empty:

```js
if (args.transcriptPath?.trim()) return null;
```

A non-empty path with no trusted match reaches the `throw`.

## Repro

1. Open a Claude Code pane.
2. Have it run `codex exec …` until Orca relabels the pane as Codex.
3. Let the pane's process exit and trigger resume.

## Expected

The guard should check that `transcriptPath` belongs to the agent it is resuming. A cross-agent mismatch means the record is stale, so fall back to the session-id scan or a fresh session instead of erroring. The deeper fix is to keep `agent` and `providerSession` in sync when a pane is reclassified.

innocarpe:
## Summary

Fixes #10517: panes mislabeled `agent: "codex"` while still holding a Claude `transcriptPath` (e.g. `~/.claude/projects/...`) hard-failed auto-resume with:

> Orca could not verify the originating Codex session file...

The resume guard treated **any** non-empty `transcriptPath` that failed trusted Codex verification as an account-safety hard fail. That is correct when the path claims a Codex rollout, but wrong when the path is cross-agent / stale metadata after a pane was reclassified (Claude running `codex exec`, etc.).

### Change

- Add `claimsCodexRolloutLayout` — true when `transcriptPath` sits in Codex's dated rollout layout (`sessions/YYYY/MM/DD/rollout-*.jsonl[.zst]`), under **any** home and without requiring the file to exist.
- In `prepareCodexSessionResumeForLaunch`, **throw only** when that claim is true and verification still failed.
- Paths that never claimed Codex provenance return `null`, so the pane relaunches instead of being blocked.

Account-safety for real Codex provenance is unchanged: a rollout-shaped path that cannot be verified still hard-fails, including one under a home Orca no longer trusts.

### Why shape and not trust

Scoping the check to *currently trusted* homes would let a real Codex rollout under an untrusted home (a removed managed account, or the import-only `codexSessionSourceHome`) skip the throw. That matters because returning `null` only declines to override `CODEX_HOME` — the renderer has already baked `codex resume <id>` into the command (`agent-session-resume.ts:241`) and `pty.ts` never rewrites it, so the session would resume under whichever account is selected. Once #10770's session bridge hardlinks rollouts across homes preserving the session-id filename, codex would *find* that id and resume silently under the wrong account. Trust is already decided upstream by `findTrustedCodexSessionResume`; the question this guard asks is whether the path claims Codex provenance at all.

### What `null` actually does

It declines to override `CODEX_HOME` — it does **not** rewrite the launch command. The pane relaunches into a usable terminal and codex reports the stale id as not found, rather than Orca blocking the spawn outright. Actually starting fresh means dropping the resume argv renderer-side; tracked in #10793. Root cause (keeping `agent` and `providerSession` in sync on reclassification) is #10792.

## Test plan

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/codex/`
- [x] `pnpm typecheck`, `pnpm lint`
- [x] Manual: Claude-path pane mislabeled codex → relaunches, no hard-fail error
- [x] Manual: real Codex resume with valid rollout path still works, history intact
- [x] Manual: rejected Codex path under trusted home still blocks
- [x] Manual: real rollout under an *untrusted* home still blocks (the case this guard turns on)

Live-validated against a real Orca dev build; details in the review comments below.

### DIFF
--- a/src/main/codex/codex-session-resume-home.test.ts
+++ b/src/main/codex/codex-session-resume-home.test.ts
@@ -3,6 +3,7 @@ import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import {
+  claimsTrustedCodexRolloutLayout,
   findTrustedCodexSessionResume,
   resolveTrustedCodexSessionResumeHome
 } from './codex-session-resume-home'
@@ -222,5 +223,41 @@ describe('resolveTrustedCodexSessionResumeHome', () => {
         listSessionFiles
       })
     ).resolves.toBeNull()
+  })
+})
+
+describe('claimsTrustedCodexRolloutLayout', () => {
+  it('is true for a rollout path under a trusted CODEX_HOME even if the file is missing', () => {
+    expect(
+      claimsTrustedCodexRolloutLayout({
+        transcriptPath: '/Users/example/.codex/sessions/2026/07/20/rollout-session.jsonl',
+        trustedCodexHomes: ['/managed/account/home', '/Users/example/.codex']
+      })
+    ).toBe(true)
+  })
+
+  it('is false for Claude (or other non-Codex) transcript paths', () => {
+    expect(
+      claimsTrustedCodexRolloutLayout({
+        transcriptPath:
+          '/Users/example/.claude/projects/-Users-example-repo/019f81b9-19a9-7651-a8d1-352d9420bd11.jsonl',
+        trustedCodexHomes: ['/Users/example/.codex', '/managed/account/home']
+      })
+    ).toBe(false)
+  })
+
+  it('is false for empty provenance and paths outside trusted homes', () => {
+    expect(
+      claimsTrustedCodexRolloutLayout({
+        transcriptPath: undefined,
+        trustedCodexHomes: ['/Users/example/.codex']
+      })
+    ).toBe(false)
+    expect(
+      claimsTrustedCodexRolloutLayout({
+        transcriptPath: '/tmp/sessions/2026/07/20/rollout-a.jsonl',
+        trustedCodexHomes: ['/Users/example/.codex']
+      })
+    ).toBe(false)
   })
 })
--- a/src/main/codex/codex-session-resume-home.test.ts
+++ b/src/main/codex/codex-session-resume-home.test.ts
@@ -3,7 +3,7 @@ import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import {
-  claimsTrustedCodexRolloutLayout,
+  claimsCodexRolloutLayout,
   findTrustedCodexSessionResume,
   resolveTrustedCodexSessionResumeHome
 } from './codex-session-resume-home'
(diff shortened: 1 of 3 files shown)


brennanb2025:
Reviewed and pushed one commit to your branch — `6056eee7a4`, on top of your `43ff7bc595`. Hope that's alright; happy to revert if you'd rather take it yourself.

**Why:** the predicate was scoped to *trusted homes*, but the question the guard is asking is about *agent shape*. A real Codex rollout under a home that isn't currently trusted (removed/signed-out managed account, or `codexSessionSourceHome`, which `index.ts:845` deliberately excludes as import-only) is still rejected Codex provenance — but it returned `false`, so no throw.

That matters because returning `null` only declines to override `CODEX_HOME`; it doesn't stop the resume. The renderer already baked `codex resume <id>` into the command (`agent-session-resume.ts:241`) and `pty.ts` never rewrites it, so `CODEX_HOME` falls through to the currently selected account (`pty.ts:3233`). Today codex usually can't find that id there and errors. Once #10770 lands its session bridge hardlinks every rollout into every managed account home preserving the session-id filename — so codex *would* find it and resume silently under the wrong account. That's exactly what the error string exists to prevent.

The fix matches Codex's dated rollout layout anywhere in the path instead:

```ts
export function claimsCodexRolloutLayout(transcriptPath: string | undefined): boolean
```

Your #10517 fix is unchanged — `~/.claude/projects/<proj>/<uuid>.jsonl` has no `sessions/YYYY/MM/DD/rollout-` segment, so it still falls through. It also drops the `trustedCodexHomes` argument and the home loop, so it's a bit less code than before. Two details worth calling out: both regexes now derive from one shared `DATED_ROLLOUT_TAIL` so the new predicate can't drift from `isCodexRolloutInsideSessionsRoot`, and the match is case-insensitive because your trusted-home version was getting Windows case folding for free via `normalizeRuntimePathForComparison` — a plain case-sensitive regex would have quietly narrowed the guard on Windows.

I also flipped the `/tmp/sessions/.../rollout-a.jsonl` assertion to `true` (it was locking in the case above) and added `.jsonl.zst`, Windows-separator, and nested/flat-layout cases. Worth knowing: before this, reverting the `index.ts` hunk entirely left all 930 codex tests green — there was no regression guard on the behaviour change at all.

Gates green: typecheck, lint, `src/main/codex/` 569 passed | 5 skipped, `pty.test.ts` 363 passed.

**One thing for you to fix:** the PR body says out-of-layout paths "return `null` and start a fresh Codex session." They don't — nothing rewrites the argv, so the pane relaunches and runs `codex resume <stale-id>`, which codex rejects with `no rollout found for thread id`. The pane does come back as a usable shell, which beats today's hard spawn rejection, so the fix is real — but "the pane comes back usable" rather than "fresh session" is the accurate wording. Actually starting fresh means dropping the resume argv in `getAgentResumeArgv`/`buildAgentResumeStartupPlan`, which is renderer-side and worth its own PR.

Your three manual test-plan boxes are still the gap — especially "rejected Codex path under trusted home still blocks," which needs a human on a real machine.

brennanb2025:
Follow-ups filed so they don't get lost, both deliberately out of scope for this PR:

- **#10792** — keep `agent` and `providerSession` in sync on pane reclassification. This is the actual root cause; #10517 names it as "the deeper fix." This PR fixes the symptom.
- **#10793** — drop the resume argv when provenance is cross-agent, so the pane really does start fresh instead of running `codex resume <stale-id>`. That's the one that would make this PR's body wording true.

CI is running now (`PR Checks` + `Windows Crash-Survival E2E`) — it hadn't been triggered before.

brennanb2025:
Ran your three manual test-plan cases live against a real Orca dev build from this branch, CDP-driven through the actual resume → `createTab` → `pty:create` → `prepareCodexSessionResumeForLaunch` path. All pass, plus a fourth case for the specific delta my commit adds.

**1. Rejected Codex path under trusted home still blocks — PASS**
Valid rollout layout under the real `~/.codex`, file absent. Pane showed `--- session restored ---` then the error banner. No Codex process spawned.

**1B. Real rollout under an *untrusted* home now blocks — PASS**
This is the case my commit changes and nothing else covered it, so I added it. Genuine rollout file, present on disk, under a `mktemp -d` home not in `trustedHomes`. Identical block, no spawn. Before the commit this fell through to `null` and would have resumed under the selected account.

**2. Mislabeled Claude pane starts without the hard-fail — PASS**
No Orca banner. Terminal:
```
$ codex '--dangerously-bypass-approvals-and-sandbox' 'resume' '15ed2ac6-…'
ERROR: No saved session found with ID 15ed2ac6-…. Run `codex resume` without an ID to choose from existing sessions.
$
```
Codex's own not-found error, then a live prompt — the pane is usable, which is the fix. (This is also exactly why the body's "start a fresh Codex session" wording is off; #10793 tracks making it true.)

**3. Real Codex resume still works — PASS**
Genuine rollout from an actual `codex exec` run. Resumed with full prior conversation intact.

**Caveats, stated honestly:**
- Seeding `sleepingAgentSessionsByPaneKey` and restarting did not fire — records are cleared on hydration — so I drove it by invoking the app's own `resumeSleepingAgentSessionsForWorktree`. Everything downstream is real app code and the UI outcomes are genuine, but I did **not** reproduce the organic relabeling from #10517.
- The pane-level cold-restore path (`buildColdRestoreAgentResumeStartup`) was NOT run. Same payload into the same IPC, so I'd expect the same result — but that's reasoning, not observation.
- Case 3 happened to exercise the legacy-migration branch (`prepareLegacySharedCodexSessionResume` hardlinked the rollout into `~/.codex` and flipped to the real home). Untouched by this PR, but it means a pure origin-home resume isn't separately covered.

CI is now green on `6056eee7a4` — `verify` and `crash-survival` both pass.

innocarpe:
@brennanb2025 Thanks for the careful review and for taking the time to run the manual test plan live on a real build — really appreciate the thoroughness and the honest caveats.

I agree with the split: this PR only addresses the hard-fail symptom, and #10792 / #10793 are the right homes for keeping `agent`/`providerSession` in sync on reclassification and for dropping a cross-agent resume argv so the pane truly starts fresh. The body’s “fresh session” wording was ahead of what the null path actually does.

If there’s anything else useful for me to follow up on or pick up as a contribution, please feel free to point me at it.

AmethystLiang:
looking into it

AmethystLiang:

PR #1, opened 2026-03-19
Linux support

mathnogueira:
## Summary
- Switch restty terminal renderer from hardcoded `webgpu` to `auto` so it falls back to WebGL2 when WebGPU/Vulkan is unavailable
- Add Linux monospace font fallbacks (DejaVu Sans Mono, Liberation Mono, Ubuntu Mono, generic monospace) so the terminal can render on systems without macOS fonts

## Test plan
- [ ] Verify clicking a worktree renders the terminal in the right panel on Linux
- [ ] Verify terminal still works on macOS (no regression)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10540, opened 2026-07-25
fix(worktree): bound the .worktreeinclude copy so a huge include can't freeze workspace creation

brennanb2025:
## Problem

`.worktreeinclude` copying is bounded in **entry count** (`WORKTREE_INCLUDE_MAX_ENTRIES = 1000`) but not in **bytes or files**, and it is awaited inline during worktree creation:

- `src/main/ipc/worktree-remote.ts` (desktop create)
- `src/main/runtime/orca-runtime.ts` (CLI / RPC / automation create)

macOS takes the cheap APFS clone path, so it is nearly free there. Linux and Windows fall back to a full `fs.cp` with no timeout and no cancellation. A repo whose `.worktreeinclude` lists `node_modules` freezes workspace creation for minutes behind the create dialog.

Per-path failures are caught and logged in `worktree-symlinks.ts`, so a mid-copy failure also leaves a partial tree with no cleanup.

## Fix — one concern: bound the copy

New `src/main/ipc/worktree-include-copy-budget.ts` measures each copy-mode source against a **cumulative budget (2 GB / 50,000 files)** *before the first byte is written*, and refuses entries that bust it. The budget is shared across the whole materialization, so a hundred medium entries trip the same guard one huge entry does.

The measurement walk is itself bounded — it returns the moment either limit is crossed, so sizing `node_modules` is not itself a freeze.

**Failing is not silent.** Refused entries are returned from `createWorktreeCopiedPaths` and surfaced on the existing `CreateWorktreeResult.warning` channel at both create sites, naming each skipped entry. A workspace quietly missing its included files would be a worse bug than the freeze.

### Why pre-measurement instead of aborting mid-copy

`fs.cp` **ignores its `signal` option** — verified on this runtime:

```
$ node -e "const fs=require('fs/promises');const ac=new AbortController();ac.abort();
  fs.cp('/tmp','/tmp/x',{recursive:true,signal:ac.signal})
    .then(()=>console.log('RESOLVED - signal ignored')).catch(e=>console.log('REJECTED',e.code))"
RESOLVED - signal ignored
```

A copy that has started cannot be cancelled, so a timeout could only stop *awaiting* it — leaving an unbounded copy still writing and a partial tree behind. Refusing up front means there is no partial state to clean up at all.

## Deliberately deferred (follow-ups, not dropped)

1. **Timeout on the copy.** Coupled to cancellation + cleanup for the reason above. On the macOS APFS path the `/bin/cp` subprocess *is* killable via `execFileAsync`'s `timeout`, but adding it only there gives asymmetric cross-platform semantics and reintroduces the partial-directory question (`cloneDirectoryWithApfs` only cleans up an *empty* reservation).
2. **Partial-tree cleanup on failure/abort.** Unreachable for the bound case now (nothing is written), but still open for a genuine mid-copy I/O failure.
3. **Full cancellation wiring** (create dialog → copy). Needs UI plumbing that does not belong in this PR.
4. **`CreateWorktreeResult.warning` has no desktop renderer consumer.** Pre-existing gap, found while wiring this up: `src/cli/handlers/worktree.ts:48` prints it and `src/main/index.ts:2098` uses it for automations, but `worktree-creation-flow.ts` drops it. So today this warning reaches CLI/automation users and main-process logs, not a desktop toast. Making the desktop flow surface `result.warning` would also start showing the existing setup-hook-skipped and startup-terminal-failed warnings, which is a separate behavior change.

## Cross-platform

- Bound applies on every platform — it gates the APFS clone and the `fs.cp` fallback identically (test: `does not run the macOS APFS clone for an over-budget entry`).
- All path handling goes through `node:path` `join`/`resolve`; no separator assumptions. The one symlink-specific test is `posixIt`-gated like the rest of the file.
- SSH / remote create (`worktree-remote.ts`) and folder workspaces are covered — the bound lives in the shared materialization path, not at one call site.
- **Link mode (`repo.symlinkPaths`) is deliberately untouched** — symlinks cost no bytes. Only copy mode (`.worktreeinclude`) is budgeted.

## Tests

New `src/main/ipc/worktree-include-copy-budget.test.ts` (14 tests) + one added to `worktree-include-file.test.ts`. The bound is injectable (`options.copyBudget`) so fixtures trip it for the right reason without writing gigabytes to disk.

- over-byte-bound entry refused, **nothing written** into the worktree
- over-file-count entry refused even though it weighs ~0 bytes
- entries that fit still copy alongside one that does not
- a normal small include copies fully and reports nothing skipped
- budget is cumulative; a refused source does not consume budget
- nested symlinks are counted, not followed
- the existing 1000-entry count bound still holds

### Revert-check (each new test genuinely fails without the fix)

Reverting `worktree-symlinks.ts` (enforcement removed):
```
Tests  5 failed | 9 passed (14)
× refuses an over-budget directory before writing anything into the worktree
× refuses an entry that busts the file-count limit
× still copies the entries that fit alongside one that does not
× copies a normal small include fully and reports nothing skipped
× does not run the macOS APFS clone for an over-budget entry
```

Neutering `admit()` inside the budget module (measurement disabled):
```
Tests  9 failed | 5 passed (14)
```

Raising `WORKTREE_INCLUDE_MAX_ENTRIES` to `Number.MAX_SAFE_INTEGER`:
```
× stops after 1000 entries so one repo file cannot request unbounded work
AssertionError: expected [ 'ignored-0.env', …(1000) ] to have a length of 1000 but got 1001
```

Restored, all green: **63 passed** across the three files; `worktrees.test.ts` + `worktree-symlink-reconciliation.real.test.ts` 203 passed. `oxlint` 0 errors, `tsc` clean on node + cli projects, max-lines ratchet OK.

## Not runtime-verified

I am on macOS, which takes the cheap APFS-clone path — **I could not runtime-reproduce the Linux/Windows `fs.cp` freeze**, nor observe the bound preventing it on those platforms. Linux/Windows behavior here is covered by tests that force `platform: 'linux'` and by static reading, not by a live repro.

### DIFF
--- a/src/main/git/worktree-include-file.test.ts
+++ b/src/main/git/worktree-include-file.test.ts
@@ -125,6 +125,20 @@ describe('resolveWorktreeIncludePaths', () => {
     expect(warn).toHaveBeenCalledWith(expect.stringContaining('unsafe'))
   })
 
+  it('stops after 1000 entries so one repo file cannot request unbounded work', async () => {
+    const names = Array.from({ length: 1001 }, (_, index) => `ignored-${index}.env`)
+    for (const name of names) {
+      writeFileSync(join(repo, name), 'A=1')
+    }
+    writeInclude(`${names.join('\n')}\n`)
+    mockCheckIgnore()
+
+    const resolved = await resolveWorktreeIncludePaths(repo)
+
+    expect(resolved).toHaveLength(1000)
+    expect(warn).toHaveBeenCalledWith(expect.stringContaining('more than 1000 entries'))
+  })
+
   it('resolves to [] when git fails instead of throwing', async () => {
     writeInclude('.env\n')
     writeFileSync(join(repo, '.env'), 'A=1')
--- a/src/main/ipc/worktree-apfs-clone.ts
+++ b/src/main/ipc/worktree-apfs-clone.ts
@@ -95,24 +95,51 @@ async function getCachedDarwinFilesystemInfo(
   return pending
 }
 
+async function isSameApfsVolume(
+  source: string,
+  targetDirectory: string,
+  deps: ApfsCloneDeps,
+  cache: DarwinFilesystemCache
+): Promise<boolean> {
+  const [sourceInfo, targetInfo] = await Promise.all([
+    getCachedDarwinFilesystemInfo(source, deps, cache),
+    getCachedDarwinFilesystemInfo(targetDirectory, deps, cache)
+  ])
+  return (
+    sourceInfo.device === targetInfo.device &&
+    sourceInfo.filesystemName === 'APFS' &&
+    targetInfo.filesystemName === 'APFS'
+  )
+}
+
 async function assertSameApfsVolume(
   source: string,
   target: string,
   deps: ApfsCloneDeps,
   cache: DarwinFilesystemCache
 ): Promise<void> {
-  const [sourceInfo, targetInfo] = await Promise.all([
-    getCachedDarwinFilesystemInfo(source, deps, cache),
-    getCachedDarwinFilesystemInfo(dirname(target), deps, cache)
-  ])
-  if (
-    sourceInfo.device !== targetInfo.device ||
-    sourceInfo.filesystemName !== 'APFS' ||
-    targetInfo.filesystemName !== 'APFS'
-  ) {
+  if (!(await isSameApfsVolume(source, dirname(target), deps, cache))) {
     throw new ApfsCloneUnavailableError(
       'APFS clone-copy requires source and target on the same APFS volume'
     )
+  }
+}
+
+/** Whether copying `source` into `targetDirectory` would take the clonefile
+ *  path. Pure probe: it reuses the cached df+diskutil pair the clone itself
+ *  runs and writes nothing, so a caller can size the work before any bytes
+ *  land. A failed probe answers "no", matching the clone's own fallback to a
+ *  real copy. */
+export async function canCloneWithApfs(
+  source: string,
+  targetDirectory: string,
+  deps: ApfsCloneDeps = defaultApfsCloneDeps,
+  filesystemCache: DarwinFilesystemCache = new Map()
+): Promise<boolean> {
+  try {
+    return await isSameApfsVolume(source, targetDirectory, deps, filesystemCache)
+  } catch {
+    return false
   }
 }
 
--- a/src/main/ipc/worktree-include-copy-budget.test.ts
+++ b/src/main/ipc/worktree-include-copy-budget.test.ts
@@ -111,24 +111,44 @@ describe('worktree copy budget tracker', () => {
   })
 
   it('charges the walk itself so repeated over-budget sources cannot re-walk forever', async () => {
-    for (const dir of ['a', 'b']) {
-      mkdirSync(join(root, dir))
-      writeFileSync(join(root, dir, 'one'), 'x'.repeat(200))
+    // 10 sources of 2 entries each, against a walk ceiling of maxEntries * 5.
+    for (let index = 0; index < 10; index += 1) {
+      mkdirSync(join(root, `dir-${index}`))
+      writeFileSync(join(root, `dir-${index}`, 'one'), 'x'.repeat(200))
     }
     writeFileSync(join(root, 'tiny'), '')
-    // Each refused source walks 2 entries before busting the byte limit, so the
-    // two of them exhaust a 4-entry ceiling.
     const tracker = createWorktreeCopyBudgetTracker({ maxBytes: 64, maxEntries: 4 })
 
-    await expect(tracker.admit(join(root, 'a'))).resolves.toMatchObject({ withinBudget: false })
-    await expect(tracker.admit(join(root, 'b'))).resolves.toMatchObject({ withinBudget: false })
+    for (let index = 0; index < 10; index += 1) {
+      await expect(tracker.admit(join(root, `dir-${index}`))).resolves.toMatchObject({
+        withinBudget: false
+      })
+    }
 
-    // Neither was admitted, so the copy budget itself is untouched and `tiny`
-    // would fit — it is refused purely because the walk ceiling is spent.
+    // Nothing was admitted, so the copy budget itself is untouched and `tiny`
+    // would fit — it is refused purely because the walk ceiling is spent, and
+    // says so rather than blaming limits it never approached.
     await expect(tracker.admit(join(root, 'tiny'))).resolves.toEqual({
+      withinBudget: false,
+      reason: 'sizing'
+    })
+  })
+
+  it('lets a small entry through after one huge entry is refused on file count', async () => {
+    // The regression this guards: sizing `node_modules` burns maxEntries + 1
+    // walk, which without headroom would starve every entry listed after it.
+    mkdirSync(join(root, 'node_modules'))
+    for (let index = 0; index < 12; index += 1) {
+      writeFileSync(join(root, 'node_modules', `pkg-${index}`), '')
+    }
+    writeFileSync(join(root, '.env'), 'A=1\n')
+    const tracker = createWorktreeCopyBudgetTracker({ maxBytes: 1024, maxEntries: 4 })
+
+    await expect(tracker.admit(join(root, 'node_modules'))).resolves.toEqual({
       withinBudget: false,
       reason: 'entries'
     })
+    await expect(tracker.admit(join(root, '.env'))).resolves.toMatchObject({ withinBudget: true })
   })
 
   posixIt('counts a nested symlink without following it', async () => {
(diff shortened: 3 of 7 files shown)


brennanb2025:
@coderabbitai review

brennanb2025:
## Review response

### CodeRabbit — `admit()` is not safe for concurrent invocation
**Actioned** in e2ebf8cd4d. Documented the sequential-await contract on `WorktreeCopyBudgetTracker.admit`. I did not serialize internally: the single production caller is a sequential `for` loop in `materializeWorktreePaths`, and a self-serializing tracker would hide the fact that parallelizing the copies is the thing you'd actually want to reason about if that loop ever changes. The comment names the exact hazard (pool read before the walk, written after) so a future parallelizing caller sees it.

That was CodeRabbit's only comment on this PR.

---

### Two defects found in review, both fixed in e2ebf8cd4d

**1. The byte budget was charged on macOS, where the copy is an APFS clone.** This was a regression on the one platform the PR's own description says was never the problem. Measured on this machine:

```
/bin/cp -n -c -R  of a 2.7 GB tree  ->  22 ms, free-space delta ~0
```

Cloning is copy-on-write, so a 2 GB byte ceiling was refusing work that costs nothing, and telling the user to "copy them in manually" — which would then cost them a real 2.7 GB copy. Bytes are now charged only when a byte-for-byte copy will actually run. The predicate is `canCloneWithApfs`, which reuses the same cached df+diskutil probe the clone itself runs and writes nothing, so *refuse before the first byte* still holds — including the darwin cross-volume case, where the fallback really is a full `fs.cp` and the byte limit correctly applies. The entry limit is unchanged on every platform: inodes are real work even on the clone path (a 61k-entry clone is still seconds).

The existing test `does not run the macOS APFS clone for an over-budget entry` encoded the regression, so it is replaced by two tests that pin the correct split — over-bytes still clones, over-entries still refuses.

**2. A refused entry consumed no budget, so the sizing walk itself was unbounded.** `.worktreeinclude` permits 1000 entries; with `remainingEntries` never decreasing on refusal, a list of over-budget directories paid a fresh full-limit walk each — up to 1000 × 50,000 `lstat` calls, still awaited inline. That re-creates the stall this PR exists to remove, and ends with the user getting *none* of their files. The walk is now charged against its own ceiling whatever the verdict.

---

### Verified, not taken on the author's word

- **`fs.cp` ignores `signal` — reproduced.** Pre-aborted signal: resolved, copied all 6000 files. Aborted 300 ms mid-flight: also resolved, ran to completion (3.0 s), all 6000 files. `signal` is not in `fs.cp`'s documented option set at all, so this is version-independent, not a Node 26 quirk. The deferral of timeout/cleanup/cancellation as one coupled follow-up is **correct**.
- **Pre-measurement does not move the freeze.** Real `node_modules` here is 2.4 GB / 61,129 entries. Walk bails at 50,001 entries in **2.1 s**; `fs.cp` of the same tree takes **21.4 s**. ~10x cheaper, and now bounded across the whole materialization by fix 2.
- **Symlink evasion is not possible.** `fs.cp` defaults `dereference: false` and BSD `cp -R` defaults to `-P`, so both backends reproduce a nested symlink as a symlink — matching the walk, which skips them. No budget bypass, no cycle hang.
- **`pending` memory is fine** — peak 4,729 on real `node_modules`, and `readdir` already materialized that array anyway.
- **Pre-existing bounds intact** — `WORKTREE_INCLUDE_MAX_ENTRIES = 1000` and the 256 KB `.worktreeinclude` file cap are untouched.
- **Not runtime-verified:** the Linux/Windows `fs.cp` freeze and its prevention. I am on macOS. Those paths are covered by `platform: 'linux'`-forced tests and static reading only. The macOS clone behavior above *is* runtime-measured.

### Deliberately not changed
`CreateWorktreeResult.warning` still has no desktop renderer consumer — `worktree-creation-flow.ts` drops it, so desktop users see refusals only in main-process logs. Left as a follow-up per the existing decision, since the channel is equally unconsumed for the pre-existing `setup-hook-skipped` and `startup-terminal-failed` warnings and wiring it is a separate behavior change.

brennanb2025:
## Review complete — 6 rounds, converged CLEAN

Reviewed with adversarial subagents looping until a round came back clean. Round 6 returned CLEAN with a LAND recommendation. **6 spawned, 6 returned, none disputed the verdict.** Every defect below was found by review and fixed on this branch; CI is green on each pushed commit.

### Author's core reasoning — verified, not taken on trust

- **`fs.cp` ignores `signal` — reproduced.** Pre-aborted: resolved and copied all 6000 files. Aborted 300ms mid-flight: also resolved, ran to completion, all 6000 files. `signal` isn't in `fs.cp`'s documented option set at all, so this is version-independent. **Deferring timeout + cleanup + cancellation as one coupled follow-up is correct.**
- **Pre-measurement does not move the freeze.** Real `node_modules` here: 2.4 GB / 61,129 entries. Walk bails at 50,001 entries in **2.1 s**; `fs.cp` of the same tree takes **21.4 s**.
- **Symlinks can't evade the budget.** `fs.cp` defaults `dereference:false`, BSD `cp -R` defaults to `-P` — both reproduce nested symlinks as symlinks, matching the walk.
- **1000-entry and 256 KB `.worktreeinclude` caps intact.**
- `worktree-symlinks.ts` confirmed as the shared implementation — 2 call sites, not 3.

### Defects found and fixed

| # | Defect | Commit |
|---|---|---|
| 1 | **Byte budget charged on macOS APFS clones.** Measured: 2.7 GB clones in **22 ms with zero disk consumed**. The bound refused work that was already free — a regression on the platform the PR says was never the problem. Bytes now charged only when a real copy will run, decided by a volume probe that writes nothing. | `e2ebf8cd4d` |
| 2 | **Sizing walk unbounded.** Refused entries consumed no budget, so 1000 over-budget entries meant up to 50M `lstat` — the stall, restored. | `e2ebf8cd4d` |
| 3 | **My fix #2 starved everything after one huge entry** (`node_modules` + `.env` → neither copied). Reproduced, fixed with walk headroom. | `984f1c1cd1` |
| 4 | **Failed clone fell through to an unbudgeted `fs.cp`** — reopened the unbounded copy my fix #1 was closing. Now bills the measured bytes and refuses if they don't fit. | `8c12d44416` |
| 5 | Walk-ceiling refusals misattributed to the file limit (a 3-file dir told it busted a 4-file limit). | `8c12d44416` |
| 6 | Singular warning read *"entry X was not copied … copying **them** …"*. | `8c12d44416` |
| 7 | A partial clone leftover reported as simply "not copied", sending users to merge into a half-populated directory. | `42dcad9faa` |
| 8 | Warning string enumerated all 1000 possible entries — unbounded, in a PR about bounds. | `42dcad9faa` |
| 9 | Two load-bearing branches untested (proven by surviving mutants). | `42dcad9faa` |
| 10 | `mayBePartial` was a false positive for *file* entries — file clones publish via `link(2)` from a temp path and leave nothing behind. | `33a5c6b26b` |

### Tests are not vacuous — every fix mutation-checked

Author's original, revert-checked: enforcement removed → **5 failed / 9 passed**; measurement neutered → **8 failed / 6 passed**; both matching the author's claims. Round 6 ran a 23-mutant battery: **20 killed, 3 survivors all provably equivalent** (a boundary label flip, a redundant early return, and a stricter-than-needed carve-out).

Three of my own tests were caught vacuous mid-review and fixed — one hit an early return instead of the branch it named, one used a fixture too small for double-billing to bite, one asserted a field on the wrong entry type. Each now kills its mutant.

### CodeRabbit
One comment: `admit()` not safe for concurrent invocation. **Actioned** — documented the sequential-await contract naming the exact hazard. Not serialized internally, because the single caller is a sequential loop and hiding the constraint is worse than stating it.

### Runtime-verified vs static-only — honest split
Runtime-verified on macOS: the `fs.cp` signal behavior, APFS clone cost and zero-disk CoW, walk-vs-copy timings on real `node_modules`, symlink handling by both backends, and the starvation repro. **Not runtime-verified:** the Linux/Windows `fs.cp` freeze and its prevention — I'm on macOS, so those paths rest on `platform:'linux'`-forced tests and static reading.

### Known gap, deliberately not fixed
`CreateWorktreeResult.warning` still has no desktop renderer consumer, so desktop users see refusals only in main-process logs. Left as a follow-up per the existing decision. I think that call is right: the channel is equally unconsumed for the pre-existing `setup-hook-skipped` and `startup-terminal-failed` warnings, so wiring it changes behavior for three warnings at once and belongs in its own PR.

### Scope
One concern throughout. `worktree-apfs-clone.ts` joined the diff only to fix the regression this PR introduced. The one marginal item is the +14-line `worktree-include-file.test.ts` addition, which pins the pre-existing 1000-entry cap this PR's reasoning depends on — flagging it, recommend keeping.

No `max-lines` disable anywhere in the diff. `tsc` (node + cli), `oxlint`, `oxfmt` all clean; 1,629 tests pass across the affected suites.

PR #10541, opened 2026-07-25
fix(worktree): refresh base-ref catalog and prefer canonical remote

### LINKED ISSUE #10516: [Bug]: Worktree base-ref picker never fetches, so branches pushed since the last fetch are invisible with no signal
**Operating system:** macOS
**Orca version:** source checkout at `1.4.148-rc.1`

## Details

**Short summary:** The Create-worktree base-branch picker searches local refs only. A branch that exists on the remote but hasn't been fetched locally produces no result and no indication that the list may be incomplete — indistinguishable from the branch not existing. There is no refresh affordance and no last-fetched signal.

### What happened

Searching for a base branch in Create worktree → **Branch** returns matches from local refs. Anything pushed to the remote since the last `git fetch` simply isn't there, silently. The user has no way to tell these two cases apart:

- the branch doesn't exist on the remote
- the branch exists but my local refs are stale

This bites hardest on repos whose active development line moves between branches (e.g. a project that cuts a new `release/vX.Y.Z` each cycle): the branch you should be basing new worktrees on is exactly the one most likely to be newer than your last fetch.

### Why the existing prefetch doesn't cover this

There *is* a prefetch on the create path (`src/main/worktree-create-base-prefetch.ts`), so this isn't "Orca never fetches". It doesn't help here, for two distinct reasons:

1. **The picker's search path never fetches.** `repos:searchBaseRefs` → `searchBaseRefDetailsForRepo` (`src/main/ipc/repos.ts:2410`) enumerates refs with `for-each-ref` against local refs only. Typing in the search box never consults the remote.

2. **The prefetch refreshes a known ref; it cannot discover unknown ones.** `prefetchLocalWorktreeCreateBase` resolves a base branch and, when that base parses as a remote-tracking ref, calls `getOrStartRemoteTrackingBaseRefresh()` — a targeted refresh of that one ref. A branch that has never been fetched isn't a candidate to refresh, so it can never appear.

### How to reproduce

1. Open a repo in Orca and let it settle (so local refs reflect the current remote).
2. From elsewhere — the web UI, another clone — push a new branch to the remote. Do **not** fetch in the Orca repo.
3. In Orca: Create worktree → **Branch** tab → search for the new branch name.
4. **Expected:** the branch appears, or the UI indicates the ref list may be stale / offers a fetch.
5. **Actual:** no result, no signal.

### Secondary finding: the broad fetch fallback is hardcoded to `origin`

`src/main/worktree-create-base-prefetch.ts:92`:

```ts
// Why: keep optimistic prefetch on the same best-effort fallback path as
// create so the real create can reuse the runtime's remote fetch cache.
await runtime.fetchRemoteWithCache(repo.path, 'origin')
```

This is a narrow fallback (reached only when the resolved base is neither a remote-tracking base nor present locally), but on a fork-based setup it fetches the wrong remote: `origin` is the fork, while the canonical remote is `upstream`. Orca already records this — repo metadata carries `gitRemoteIdentity.remoteName` (`"upstream"` on my fork-based repo).

It's also inconsistent with neighbouring code: `resolveRemoteTrackingBase()` (`src/main/runtime/orca-runtime.ts`) is properly remote-aware, matching the base name against the real `git remote` list and taking the longest match.

### Suggested directions

Roughly by value-to-effort, not prescriptive:

- **Add a refresh/fetch affordance to the base picker**, plus a last-fetched timestamp. This alone removes the ambiguity.
- **Kick a best-effort background `fetch --prune` when the picker opens** (not only once a base is resolved on create), then re-run the ref search. Can reuse `fetchRemoteWithCache`'s existing de-dup/caching.
- **Replace the hardcoded `'origin'`** at `worktree-create-base-prefetch.ts:92` with the repo's recorded canonical remote (`gitRemoteIdentity.remoteName`), falling back to `origin` when unset.

### Footnote: a permanent variant of the same gap

The transient case above resolves itself on the next fetch. It can also be permanent: if a remote is configured with a single-branch refspec (`git remote add -t <branch>`, or a `--single-branch` clone), `git fetch` never brings down any other branch of that remote, so the missing branches never appear no matter how long you wait. I hit this on a fork where `remote.upstream.fetch` was pinned to one branch — the picker showed exactly one upstream ref, and I concluded Orca couldn't follow the upstream release line. After `git remote set-branches upstream '*' && git fetch upstream --prune`, upstream refs went from 1 to 111 and the picker was correct.

To be clear, **Orca does not create that config** — both `remote add` callsites (`src/main/ipc/worktree-push-target-setup.ts:93`, `src/main/ipc/worktree-remote.ts:1127`) run a plain `git remote add <name> <url>` with no `-t`, yielding a wildcard refspec. Mentioning it only because a cheap diagnostic would cover it: detecting a non-wildcard `remote.<name>.fetch` and surfacing a hint in the picker ("`upstream` tracks a single branch — other branches won't appear") would make an otherwise near-invisible condition legible.

innocarpe:
## Summary

Fixes #10516: Create-worktree **Branch** search only enumerated local refs, so a branch pushed since the last fetch looked nonexistent (no refresh affordance, no last-fetched signal). Prefetch/create broad-fetch fallbacks also hard-coded `origin`, which is wrong on fork clones whose canonical remote is `upstream`.

### Changes

1. **Base-ref search discovery fetch** — before local `for-each-ref` search, best-effort `fetchRemoteWithCache` for the repo’s recorded canonical remote (deduped/freshness-cached by the existing runtime helper). Offline/auth failures still return whatever local refs exist.
2. **Canonical remote helper** — `resolveWorktreeBroadFetchRemoteName` uses `gitRemoteIdentity.remoteName`, falling back to `origin`.
3. **Create / prefetch fallbacks** — replace hard-coded `origin` with that helper in:
   - `worktree-create-base-prefetch.ts`
   - `orca-runtime` create path
   - `worktree-remote` create path (runtime + legacy `git fetch` fallback)

SSH base-ref search is unchanged (already lists on the remote host).

Not in this PR: UI last-fetched timestamp / manual refresh button, or single-branch refspec diagnostics (footnote in the issue).

## Test plan

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/shared/worktree-broad-fetch-remote.test.ts src/main/ipc/repos-remote.test.ts src/main/ipc/worktrees.test.ts` (309+ tests)
- [ ] Manual: push a new remote branch without fetching in Orca → Create worktree → Branch search finds it after the catalog refresh
- [ ] Manual: fork clone with `upstream` identity fetches `upstream`, not only `origin`
- [ ] Manual: offline / failed fetch still shows existing local refs without erroring the picker

## ELI5

Create Worktree branch search only saw local refs, so newly pushed remote branches looked missing, and fetches hard-coded origin. Search refreshes the canonical remote first and prefers the right remote on forks.

### DIFF
--- a/src/main/ipc/repos-remote.test.ts
+++ b/src/main/ipc/repos-remote.test.ts
@@ -94,7 +94,8 @@ vi.mock('../git/repo', async () => {
     getRepoName: vi.fn().mockImplementation((path: string) => path.split('/').pop()),
     getBaseRefDefault: vi.fn().mockResolvedValue('origin/main'),
     getRemoteCount: vi.fn().mockResolvedValue(1),
-    searchBaseRefs: vi.fn().mockResolvedValue([])
+    searchBaseRefs: vi.fn().mockResolvedValue([]),
+    searchBaseRefDetails: vi.fn().mockResolvedValue([])
   }
 })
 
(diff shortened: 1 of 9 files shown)


innocarpe:
### Sync update (`5393831dbf`)

Rebase onto latest upstream/main: base-ref catalog refresh + fetchRemoteWithCache; keep paired-client repo notifier.

- Commit: `5393831dbfe2d0feb6ce38a66467bc4d923fa135`
- Upstream: https://github.com/stablyai/orca/pull/10541
- Fork portfolio: https://github.com/innocarpe/orca/pull/68

innocarpe:
### Sync update (`5393831dbf`)

Rebase onto latest upstream/main: base-ref catalog refresh + fetchRemoteWithCache; keep paired-client repo notifier.

- Commit: `5393831dbfe2d0feb6ce38a66467bc4d923fa135`
- Upstream: https://github.com/stablyai/orca/pull/10541
- Fork portfolio: https://github.com/innocarpe/orca/pull/68

Issue #8941, opened 2026-07-16
[Other]: WSL resource usage spikes


kenkenken123:
### Details

After I start Orca, the CPU and disk usage of my Windows WSL (the vmmemwsl process) spikes from time to time. I am not running any programs in WSL myself.

kenkenken123:
I have Orca installed in the Windows environment, and I also have WSL set up. Whenever I open Orca, the WSL process occasionally consumes a lot of system resources; things return to normal once I close it.

OrcaWin:
## Agent reproduction package (Windows orchestration 2026-07-16)

**Host context:** Windows 11 / Orca **1.4.143.0** / repo path used: `docs/bug-reproductions/`  
**Labels applied:** see issue labels (`has_repro` or `cannot_repro`).  
**Security:** Do **not** run user-uploaded binaries from issue attachments. Recreate fixtures from the steps/scripts below.

---
# Issue 8941 — WSL resource usage spikes

**Status:** `repro_confirmed` (observational under Orca load)  
**Host:** Windows 11, Orca 1.4.143 running  
**Label:** `has_repro`  
**Note:** Issue type is `[Other]` (not labeled `bug` upstream)

## Summary

Reporter: after starting Orca, WSL (`vmmemWSL`) CPU/disk spikes intermittently even with no user programs in WSL; settles after closing Orca.

## What we observed (2026-07-16)

With **Orca running** (multiple `Orca` processes + `orca-terminal-daemon`):

| Process | Count | Working set (approx) |
|---------|------:|---------------------:|
| vmmemWSL | 1 | ~632 MB |
| wsl.exe | **74** | ~786 MB total |
| wslhost | **39** | ~473 MB total |
| wslservice | 1 | ~27 MB |
| wslrelay | 1 | ~8 MB |

Functional impact:

```powershell
# Hung >15s (job timeout)
wsl -d Ubuntu -- echo ok
```

`wsl -l -v` still shows Ubuntu **Running**. So the VM is up, but guest command execution is wedged under a storm of `wsl.exe` clients — consistent with Orca (and/or other tools) spawning many concurrent WSL probes (agent detection, hooks, path checks, git, etc.).

## Repro (another agent)

```powershell
# With Orca open and at least one WSL-touching feature used:
Get-Process vmmem*,wsl*,wslhost* |
  Group-Object Name |
  ForEach-Object {
    [pscustomobject]@{
      Name = $_.Name
      Count = $_.Count
[...]

PR #10551, opened 2026-07-25
fix(wsl): prevent unexpected WSL VM boot on Windows app startup

yiuf63:
## Summary
Fixes an issue on Windows host where launching Orca or running background services (e.g. CLI fallback probes, AI Vault, auth checks) would unexpectedly boot stopped WSL distros and spawn \mmemWSL\.

### Root Cause
1. **CLI Fallback**: \
esolveDefaultWslCli\ (used when host \gh\ or \glab\ is missing on Windows) executed \wsl.exe -d <distro>\, booting stopped WSL distros during background status checks.
2. **Unconditional Home Checks**: \getWslHome\ / \getWslHomeAsync\ resolved paths without checking if the target WSL distro was currently running.
3. **\.status\ Querying**: \isWslAvailable()\ invoked \wsl.exe --status\ instead of inspecting local binary path.

### Key Changes
- **\src/main/git/runner.ts\**: Prevent \
esolveDefaultWslCli\ from executing \wsl.exe -d <distro>\ when the WSL distro is not running.
- **\src/main/wsl.ts\**: Guard \getWslHome\ and \getWslHomeAsync\ with \listRunningWslDistros()\ by default (\^GllowBoot: false\), and update \isWslAvailable()\ to check file existence (\^[xistsSync\).
- **Explicit WSL actions**: Pass \{ allowBoot: true }\ in \worktree-logic.ts\ and \skill-discovery-target.ts\ when user actions specifically target WSL.
- **AI Vault & CLI Reconciliation**: Skip WSL distro scans when scope is purely local Windows or when no WSL CLIs are registered.

### Test Plan
- Unit tests added/updated in \
unner-wsl-gh-fallback.test.ts\ and \wsl.test.ts\.
- All 75 vitest tests and \pnpm run typecheck\ pass cleanly.

## ELI5

On Windows, launching Orca or background checks could boot stopped WSL distros and wake heavy VMs you weren't using. WSL probes only run when appropriate so startup doesn't surprise-start WSL.

### DIFF
--- a/src/main/ai-vault/cached-session-list.ts
+++ b/src/main/ai-vault/cached-session-list.ts
@@ -1,6 +1,6 @@
 import { join } from 'node:path'
 import { scanAiVaultSessions } from './session-scanner'
-import { getWslHomeAsync, listWslDistrosAsync } from '../wsl'
+import { getWslHomeAsync, isWslPath, listRunningWslDistrosAsync } from '../wsl'
 import type { AiVaultListArgs, AiVaultListResult } from '../../shared/ai-vault-types'
 import { LOCAL_EXECUTION_HOST_ID } from '../../shared/execution-host'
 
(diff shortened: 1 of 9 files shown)


Jinwoo-H:
Thanks @yiuf63 for the WSL startup probe diagnosis and regression coverage. We are taking ownership through existing issue #8941, which tracks the unexplained WSL VM boot and resource spike on Orca startup. We will reimplement on current main and validate on Windows; no contributor rebase, retest, or other follow-up is requested. Closing this superseded branch while preserving your credit.

PR #10556, opened 2026-07-25
fix(mobile): reuse host tab when viewing a live terminal from phone

### LINKED ISSUE #10486: [Bug]: Viewing an active terminal from mobile creates a duplicate desktop terminal tab
### Operating system

macOS

### Orca version

1.4.153-rc.3

### Details

**Short summary**

Viewing an active Orca terminal from a paired mobile client causes the macOS host to open an additional terminal tab. The new tab shows the same terminal content as the existing tab.

**What happened?**

An active Codex terminal was viewed from Orca on a phone. Afterward, the desktop Orca window contained an extra terminal tab with the same content as the original terminal.

The extra tab was manually closed. The issue then reproduced again after using the mobile view, so this was not a one-time stale UI artifact.

**How can we reproduce it?**

1. On the macOS host, keep an active agent terminal open in Orca.
2. Connect to Orca from a phone and view that terminal.
3. Return to the desktop Orca window.
4. Observe that an additional terminal tab has appeared with the same content as the original tab.
5. Close the extra tab and repeat the mobile-view flow; the duplicate tab can appear again.

**Expected behavior**

The mobile client should attach to or mirror the existing terminal without creating another terminal tab on the desktop host.

**Actual behavior**

A second terminal tab appears on the desktop and displays the same content as the original terminal.

**Additional notes**

- Reproduced twice on Orca 1.4.153-rc.3.
- The duplicate tab can be closed manually, but that is only a temporary workaround.
- Terminal handles and PTY/process IDs were not captured before closing the duplicate, so it is not yet known whether this is a duplicate UI mirror of the same PTY or a separately resumed process.
- A screenshot was captured, but it contains private conversation content and is intentionally not attached. A fully redacted tab-bar-only crop can be provided if needed.
- This may be a mobile-specific regression or residual variant related to #8878, #9352, and #10208, but the trigger here is viewing the terminal from a phone rather than switching worktrees or using a second desktop client.

innocarpe:
## Summary

Fixes #10486 — viewing an active host terminal from a paired mobile client no longer opens a second desktop tab with the same session.

When mobile focus/reveal asks the host to surface a live PTY, the renderer must **adopt the existing tab** that already owns that PTY. Ownership was only checked via `tab.ptyId` / a partial map path; if the binding lived only in `terminalLayoutsByTabId` (or the live `ptyIdsByTabId` map was not consulted by the resolver), `createTab({ initialPtyId })` minted a duplicate UI tab on the same content.

### Changes
- `resolveTerminalTabIdForPtyId` also consults `ptyIdsByTabId`.
- `onCreateTerminal` reuses a tab resolved via that full ownership check before creating.
- Planner state for mobile tab mounts includes `ptyIdsByTabId` for consistent resolution.

## Test plan

- [x] `terminal-tab-for-pty-id.test.ts` (incl. live map ownership)
- [x] `mobile-terminal-tab-mount.test.ts`
- [ ] Manual: desktop agent terminal open → view from phone → return to desktop → only one tab for that session
- [ ] Manual: split-pane terminal leaf viewed from mobile still targets the parent tab

## ELI5

Viewing a live desktop terminal from mobile opened a second duplicate tab on the host. The host now reuses the existing tab that already owns that session instead of creating another one.

### DIFF
--- a/src/renderer/src/hooks/useIpcEvents.ts
+++ b/src/renderer/src/hooks/useIpcEvents.ts
@@ -16,7 +16,7 @@ import { OPEN_WORKSPACE_BOARD_EVENT } from '@/components/sidebar/useWorkspaceBoa
 import { SPLIT_TERMINAL_PANE_EVENT, CLOSE_TERMINAL_PANE_EVENT } from '@/constants/terminal'
 import { requestBackgroundTerminalWorktreeMount } from '@/components/terminal/background-terminal-worktree-mount'
 import { planMobileTerminalTabMount } from '@/lib/mobile-terminal-tab-mount'
-import { resolveTerminalTabIdForPtyId } from '@/lib/terminal-tab-for-pty-id'
+import { resolveTerminalTabPtyOwnership } from '@/lib/terminal-tab-for-pty-id'
 import {
   hasRegisteredRuntimeTerminalTab,
   focusRuntimeTerminalSurface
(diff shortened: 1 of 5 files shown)


brennanb2025:
Thanks for digging into this @innocarpe — your diagnosis was right, and the ownership resolver here is the shape the fix needed.

I've opened #11259 with a variant. Three things I ran into building on this:

1. **The tests don't cover the change that fixes the bug.** All 96 stay green with `useIpcEvents.ts` reverted — the coverage is on the lib functions, not the handler wiring. This repo has enough vacuous tests that it's worth checking explicitly.
2. **A stale layout row can outrank the pre-minted `tabId`.** `existingOwnedTab` wins over `hintedPendingTab`, so a row that outlived its pane adopts the wrong tab confidently.
3. **The `Ambiguous terminal tab ownership` throw reaches users.** `focusTerminal` awaits `revealTerminalSession` with no `catch`, so the renderer's error reply rejects the phone's call.

#11259 ranks bindings instead: mounted pane (`ptyIdsByTabId`) > recorded (`tab.ptyId` or layout row) > the `tabId` hint, which is a fallback rather than a binding — main's `pty.tabId` is written once at spawn and never rewritten, so it goes stale as soon as a pane is dragged to another tab. Same-tier conflicts stay ambiguous so `planMobileTerminalTabMount` keeps failing closed, and nothing throws.

Happy to close this in favour of that one, or fold it in — your call.

nwparker:

- Issue #10486 was fixed by #11259 (`fix(mobile): stop a mobile reveal from minting a duplicate terminal tab`)

Superseding merge: #11259

innocarpe:
Thanks for the detailed follow-up, Brennan. I agree #11259 is the stronger fix—the handler-level regression coverage, binding ranking, and non-throwing ambiguity path address the gaps in #10556 well.

With #11259 landed, closing #10556 as superseded makes sense. Glad the original diagnosis helped, and I’ll make sure future handler fixes include a regression test that fails when the wiring is reverted.

PR #10557, opened 2026-07-25
fix(runtime): degrade focused terminal create on headless serve

### LINKED ISSUE #10553: Headless serve: agent create from desktop client fails with 'No renderer window available' since ~v1.4.153
### Summary

Since the `terminal.createAgentSession` RPC landed (~v1.4.153), every agent create issued by a desktop client against a **headless `orca serve`** host fails with:

```
runtime_error: No renderer window available
```

This breaks the core remote-server workflow ([docs/remote-servers](https://www.onorca.dev/docs/remote-servers)): a desktop app paired to a Linux `orca serve` box can no longer spin agents or create focused terminal tabs on it. Both sides on v1.4.155.

### Reproduction

On any Linux host running `orca serve` (systemd/`xvfb-run`, never promoted to a desktop window):

```bash
# background create — works:
orca terminal create --worktree path:/path/to/repo --json
# → ok:true, surface:"background"

# focused create — fails:
orca terminal create --worktree path:/path/to/repo --focus --json
# → { "ok": false, "error": { "code": "runtime_error", "message": "No renderer window available" } }
```

From a paired desktop client (v1.4.155), any **agent** create on a worktree of that host fails the same way — the client sends `terminal.createAgentSession` with `presentation: 'focused'` for every user-activated create.

### Mechanism (traced in v1.4.155 source)

1. `src/renderer/src/runtime/web-runtime-session.ts` — the desktop client now creates agents via `terminal.createAgentSession` with `presentation: args.activate === false ? 'background' : 'focused'`. A user-initiated create is always `'focused'`.
2. `src/main/runtime/orca-runtime.ts` `createAgentSession` → `createTerminal(..., { presentation: request.presentation ?? 'background', ... })`.
3. In `createTerminal`, `presentation:'focused'` sets `requiresRendererFocus`, so `shouldCreateInBackground` is false (no `agentSessionClaim`, `rendererBacked` unset) → the renderer-IPC path runs → `getAuthoritativeWindow()`.
4. On a headless serve `authoritativeWindowId` is `HEADLESS_RUNTIME_WINDOW_ID` (0) from the boot-time empty-graph publish; `BrowserWindow.fromId(0)` is null → `throw new Error('No renderer window available')`.
5. The client's legacy fallback (`remote-agent-session-launch.ts`) only downgrades on `agent_session_legacy_required` / `method_not_found` — a `runtime_error` surfaces straight to the user, so nothing falls back to the previously working `session.tabs.createTerminal` (`activate:false`) path, which handles headless correctly via `createHeadlessMobileSessionTerminal`.

### Expected

On a windowless serve, a focused agent create should degrade to the headless/background create path (the same way `session.tabs.createTerminal` and non-focused CLI creates already do), or the host should answer `agent_session_legacy_required` so the client's own fallback engages — not a hard error.

### Workaround

Running `orca open` on the serve host itself (local CLI is allowed to promote; a paired client is not) promotes the serve to host a renderer window on the Xvfb display — `attachWindow` swaps `HEADLESS_RUNTIME_WINDOW_ID` for the real window and preserves live PTY bindings, after which desktop-client creates work again. The window dies with every serve restart, so we run it as a systemd `ExecStartPost`. This costs a permanent renderer process on the server, which the headless design presumably wanted to avoid.

### Environment

- Server: Ubuntu Linux, `orca serve` v1.4.155 under systemd + `xvfb-run` (per the remote-servers docs)
- Client: macOS desktop app v1.4.155, paired over the runtime websocket
- Worked on server v1.4.134 (pre-`createAgentSession`, clients used `session.tabs.createTerminal`)

innocarpe:
## Summary

Fixes #10553 — agent creates from a paired desktop client against headless `orca serve` no longer fail with `No renderer window available`.

Since `terminal.createAgentSession`, clients send `presentation: 'focused'` for user-initiated agent creates. On a windowless serve host, that took the renderer-IPC path and called `getAuthoritativeWindow()`, which throws when `BrowserWindow.fromId` is null.

### Fix
If `presentation`/`focus` requires a renderer window but none is available, use the existing **headless/background** `createTerminal` path (same as non-focused CLI creates and `rendererBacked` without a window). The PTY still spawns; surface is returned as `background` with the usual optional reveal warning when no notifier can promote a tab.

## Test plan

- [x] `orca-runtime.test.ts` — focused create without window → background surface
- [x] `orca-runtime.test.ts` — focused `createAgentSession` without window
- [x] existing renderer-backed-without-window case still passes
- [ ] Manual: desktop client paired to headless `orca serve` can create agents without error

## ELI5

Creating a focused agent from a paired client against headless `orca serve` failed with "No renderer window." Focused creates now degrade gracefully to a windowless path so remote agents still start.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -12064,6 +12064,66 @@ describe('OrcaRuntimeService', () => {
     )
   })
 
+  it('degrades focused terminal creates on headless serve instead of requiring a renderer window', async () => {
+    // Why (#10553): desktop clients always send presentation:'focused' for agent
+    // creates; headless orca serve must still return a handle, not hard-error.
+    const spawn = vi.fn().mockResolvedValue({ id: 'pty-focused-headless' })
+    const runtime = new OrcaRuntimeService(store)
+    runtime.setPtyController({
+      spawn,
+      write: () => true,
+      kill: () => true,
+      getForegroundProcess: async () => null
+    })
+
+    await expect(
+      runtime.createTerminal(`path:${TEST_WORKTREE_PATH}`, {
+        command: 'codex',
+        presentation: 'focused'
+      })
+    ).resolves.toMatchObject({
+      worktreeId: TEST_WORKTREE_ID,
+      surface: 'background',
+      handle: expect.stringMatching(/^term_/)
+    })
+    expect(spawn).toHaveBeenCalledWith(
+      expect.objectContaining({
+        worktreeId: TEST_WORKTREE_ID,
+        persistHostSessionBinding: true
+      })
+    )
+  })
+
+  it('degrades focused createAgentSession on headless serve without a renderer window', async () => {
+    const spawn = vi.fn().mockResolvedValue({ id: 'pty-agent-headless' })
+    const runtime = new OrcaRuntimeService(store)
+    runtime.setPtyController({
+      spawn,
+      write: () => true,
+      kill: () => true,
+      getForegroundProcess: async () => null
+    })
+    const clientOperationId = `${Date.now()}-0123456789abcdef0123456789abcdef`
+
+    await expect(
+      runtime.createAgentSession({
+        clientOperationId,
+        worktree: `path:${TEST_WORKTREE_PATH}`,
+        agent: 'codex',
+        prompt: 'hello',
+        presentation: 'focused'
+      })
+    ).resolves.toMatchObject({
+      disposition: 'created',
+      terminal: expect.objectContaining({
+        worktreeId: TEST_WORKTREE_ID,
+        surface: 'background',
+        handle: expect.stringMatching(/^term_/)
+      })
+    })
+    expect(spawn).toHaveBeenCalled()
+  })
+
   it('accepts renderer-backed terminal create replies only from the target renderer', async () => {
     const webContents = { send: vi.fn() }
     const send = vi.fn((_channel: string, payload: { requestId: string }) => {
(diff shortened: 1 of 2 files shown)


innocarpe:
Status ping: focused headless-serve fix so `presentation: 'focused'` agent creates degrade cleanly without a renderer window (tests included). No open threads — ready for a maintainer look when convenient.

Jinwoo-H:
Thanks for this, and for the diagnosis — it was correct, and it made the fix straightforward to confirm.

We reproduced #10333 against a **real headless `orca serve` host** (not a mock) and it fails exactly where you identified: a focus-requested create falls through to `getAuthoritativeWindow()` and throws. That part of your analysis held up completely.

We've opened #12791 with the fix. Two reasons we went with a separate PR rather than merging here, both about test coverage rather than your diagnosis:

- **#10370** — the production change is logically equivalent to ours. But reverting the `orca-runtime.ts` call site while keeping the extracted helper leaves **all 7 tests passing**, because they only exercise the pure function's own truth table; nothing constructs a runtime, calls `createTerminal`, or observes the throw. So the tests can't catch a regression in the thing being fixed.
- **#10557** — the one-clause change is correct. But the tests never wire the headless-serve state (no `syncWindowGraph`/`attachWindow`), so `graphStatus` stays `'unavailable'`. Replicating that test shape against unfixed main gives `Error: runtime_unavailable`, not `No renderer window available` — it goes red on main for a different reason, so it never demonstrates the reported bug.

One thing worth knowing that neither PR could have anticipated: **the Mac UI `+` half of this issue no longer reproduces on current main.** #9687 and #10193 landed after the original report and now send `presentation: 'background'` for paired clients. So the surviving defect is the CLI/RPC path only — which is what #12791 closes at the runtime layer, since it stays reachable for any caller without a `clientKind`.

Genuinely appreciate the work here; the fix in #12791 is the shape you both proposed. Closing these in favour of it, but the credit for finding it is yours.

innocarpe:
@Jinwoo-H appreciate the write-up. Fair call on the tests — if they go red on main for `runtime_unavailable` instead of the actual window error, they never really pin the bug we claimed to fix.

I'll close this out for #12791. Thanks for the credit on the diagnosis; that means a lot after the PR sat a while.

innocarpe:
Closing — tracked on #12791 instead.

PR #10570, opened 2026-07-25
fix(usage): guard web client against undefined usage scan state

scokeepa:
The claude/codex/openCode usage store slices read `scanState.enabled` directly off `window.api.<provider>Usage.getScanState()`. In the web client that usage IPC is not bridged, so the preload fallback proxy resolves those calls to `undefined`, and enabling usage tracking from Settings -> Stats & Usage throws
`TypeError: Cannot read properties of undefined (reading 'enabled')` (reproduced live against `orca serve` v1.4.150; present on main too).

Guard the getScanState()/setEnabled() seams in all three slices so an absent scan state degrades to a graceful no-op instead of crashing. Desktop behavior is unchanged (a real ScanState is always truthy).

Adds a regression test that stubs the web-client fallback (every call -> undefined) and asserts fetch*/enable* no-op without throwing for all three providers.

## Summary

Describe the user-visible change.

## Screenshots

- Add screenshots or a screen recording for any new or changed UI behavior.
- If there is no visual change, say `No visual change`.

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

## AI Review Report

Summarize the code review you ran with your AI coding agent. Include the main risks it checked, what it flagged, and what you changed or verified as a result.
Confirm that the review explicitly checked cross-platform compatibility for macOS, Linux, and Windows, including shortcuts, labels, paths, shell behavior, and any Electron-specific platform differences touched by this PR.

## Security Audit

Provide a basic security audit summary from your AI coding agent. Call out any input handling, command execution, path handling, auth, secrets, dependency, or IPC risks that were reviewed, plus any follow-up needed.

## Notes

Call out any platform-specific behavior, risks, or follow-up work.

## ELI5

Turning on usage tracking in the web client crashed because usage APIs aren't bridged there and returned undefined. Those reads are guarded so Settings → Stats works without throwing.

### DIFF
--- a/src/renderer/src/store/slices/claude-usage.ts
+++ b/src/renderer/src/store/slices/claude-usage.ts
@@ -45,7 +45,14 @@ export const createClaudeUsageSlice: StateCreator<AppState, [], [], ClaudeUsageS
     try {
       const nextScanState = (await window.api.claudeUsage.setEnabled({
         enabled
-      })) as ClaudeUsageScanState
+      })) as ClaudeUsageScanState | undefined
+      // Why: the web client (paired runtime) does not bridge the desktop-only
+      // usage IPC; its preload fallback resolves this call to `undefined`. Bail
+      // so the toggle no-ops instead of seeding an empty scan state and then
+      // crashing on the follow-up fetch.
+      if (!nextScanState) {
+        return
+      }
       set({
         // Why: every enable should look like a fresh scan cycle in the UI.
         // Reusing the last completed timestamp makes repeated toggles skip the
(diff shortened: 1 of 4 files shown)


OrcaWin:
Thanks for this — the bug is real and the fix is correct. Closing only because **it already shipped**.

This branch is byte-identical to your own #10073, which merged on 2026-07-24 as `6997bc40a` — one day before this PR was opened. Both diffs hash to the same md5 (`5bb5f691b9b5e9d7e25fb04a11a8273a`), `git rebase origin/main` reports "skipped previously applied commit", and `git diff origin/main` is empty. There is nothing left to merge.

For the record, I did reproduce the underlying bug before concluding that: the three usage slices read `scanState.enabled` straight off the IPC result, and the web preload fallback resolves it to `undefined`, so Settings → Stats & Usage threw `TypeError`. Your guard is the right fix and is live on main today.

One follow-up worth filing separately, out of scope here: the guard stops the crash but leaves the Enable toggle silently no-opping in the web client — a user clicks it and nothing ever happens. There is already a precedent for the better shape in this same file, where `workspacePorts.scan` returns a capability object with an `unavailableReason` instead of the generic `undefined` fallback. Applying that to the three usage providers would let the pane render an honest "unavailable in the web client" state instead of a dead switch.

Looks like an accidental re-submission of an already-merged branch. Closing as already-fixed-by-#10073.

Issue #8126, opened 2026-07-10
PR auto-link fails for Claude Code PRs: ANSI reset code glued to URL breaks terminal scanner


VladTemp27:
## Summary

Orca fails to auto-link a PR to its worktree when the PR was created by a Claude Code session, because the terminal-output URL scanner doesn't strip ANSI escape codes before parsing.

## How the linker works

Orca auto-links a PR to a worktree by scanning raw terminal PTY output for `https://github.com/<owner>/<repo>/pull/<n>` (renderer detector `observeTerminalGitHubPRLink` → `observeTerminalGitHubPullRequestLink`), verifying the PR's head matches the worktree branch via `gh pr list`, then persisting `linkedPR` in worktree meta.

## Bug

The scanner's URL terminators are only whitespace and `"'<>` — ANSI escape sequences are not stripped first. Claude Code's ink-based TUI closes styled text runs with an SGR reset glued directly to the last character, e.g.:

```
…/pull/10\x1b[22m
```

Because the reset code is glued on with no separating terminator, the candidate URL fails to parse and no link is persisted.

Agents that print the URL plainly followed by a newline (e.g. Antigravity) link correctly — only ANSI-styled output breaks it.

## Proof

A/B tested 2026-07-09 in fresh Orca panes: a PR URL glued to an SGR reset code did not link; the same URL followed by a plain newline linked within seconds.

## Suggested fix

Strip ANSI control sequences before scanning for candidate URLs, the same way the main-process `AdvertisedUrlWatcher.stripTerminalControls` already does for dev-server URL detection. Applying the same stripping step to the GitHub PR link scanner should fix this without needing terminator-list changes.

## Workaround (in use today)

After creating a PR from a Claude Code session, print the PR URL again on its own line with plain `echo` (no styled output) followed by whitespace, e.g. `echo "https://github.com/owner/repo/pull/N "` — Orca then links it within seconds.

## Environment

- Orca app ~1.4.123
- Agent: Claude Code (ink-based TUI)

VladTemp27:
## Follow-up: refined root cause

Investigated whether this is related to Claude Code rendering "fullscreen" — it isn't, in the alternate-screen-buffer sense (vim/htop-style), but the actual mechanism is closely related and worth recording.

[...]

PR #10573, opened 2026-07-25
fix(terminal): terminate HTTP links on CJK brackets and whitespace

### LINKED ISSUE #10571: [Bug]: Terminal URL detection swallows CJK text — non-ASCII characters are never treated as URL terminators
### Operating system

macOS

### Orca version

v1.4.155

### Details

**Short summary:**  <br>Terminal URL detection never terminates a URL on non-ASCII characters, so any CJK text that directly follows a URL (full-width parentheses, ideographic space, `、。・「」`, …) gets swallowed into the link. The underline is drawn over the whole run, and ⌘-click / ⇧⌘-click opens a percent-encoded garbage URL instead of the real one.

**What happened?**

Claude Code (and any tool that annotates a URL in Japanese/Chinese) frequently prints lines like:

```
- PR: https://github.com/stablyai/orca/pull/12345（created・waiting for merge）
```

Orca underlines `https://github.com/stablyai/orca/pull/12345（created・waiting for merge）` as a single link, and opening it navigates to:

```
https://github.com/stablyai/orca/pull/12345%EF%BC%88created%E3%83%BBwaiting%20for%20merge%EF%BC%89
```

which 404s. The same happens with `「…」` quotes, `、` and a full-width space (U+3000) between the URL and the following text.

**How can we reproduce it?**

1. In a terminal pane, run:

   ```sh
   echo 'PR: https://github.com/stablyai/orca/pull/12345（作成済み・マージ待ち）'
   echo '詳細は「https://example.com/docs」を参照'
   echo 'PR: https://github.com/stablyai/orca/pull/12345　（全角スペース区切り）'
   ```
2. Hover / ⌘-click (or ⇧⌘-click for the system browser) on the URL.
3. The link range extends past the URL into the CJK text, and the opened URL contains the percent-encoded Japanese text.

Control case that works correctly: with a **half-width** space or half-width parentheses, the boundary is detected properly:

```sh
echo 'PR: https://github.com/stablyai/orca/pull/12345 （作成済み）'   # OK
echo 'PR: https://github.com/stablyai/orca/pull/12345 (作成済み)'    # OK
```

**Root cause**

In the renderer bundle, `iterateTerminalHttpUrlCandidates()` grows the candidate from the `http(s)://` scheme until `isHttpUrlBodyTerminator()` returns true, then trims with `isHttpUrlTrailingPunctuation()`. Both predicates are ASCII-only:

```js
function isHttpUrlBodyTerminator(code) {
  return isAsciiWhitespace(code) || code === 34 || code === 39 || code === 33
      || code === 42 || code === 40 || code === 41 || code === 123 || code === 125
      || code === 124 || code === 92 || code === 94 || code === 60 || code === 62
      || code === 96;
}

function isAsciiWhitespace(code) {
  return code === 9 || code === 10 || code === 11 || code === 12 || code === 13 || code === 32;
}
```

No code point above U+007F can ever terminate the candidate, and the trailing trim (`: , . ? [ ] ~ ( ) …`) is ASCII-only as well, so nothing is stripped afterwards. `new URL()` then happily parses the result (the CJK text is percent-encoded), so the candidate passes validation in `extractTerminalHttpLinks()` and becomes a clickable link.

I extracted those functions from the shipped 1.4.155 bundle and ran them directly:

| Input line | Detected link |
| -- | -- |
| `.../pull/2 （作成済み…）` (half-width space) | ✅ `.../pull/2` |
| `.../pull/2（作成済み・マージ待ち）` | ❌ `.../pull/2%EF%BC%88作成済み・マージ待ち%EF%BC%89` |
| `.../pull/2　（作成済み…）` (U+3000) | ❌ includes U+3000 + the whole comment |
| `詳細は「https://example.com/docs」を参照` | ❌ `https://example.com/docs%E3%80%8D%E3%82%92%E5%8F%82%E7%85%A7` |
| `.../pull/2 (作成済み)` (half-width paren) | ✅ `.../pull/2` |

**Suggested fix**

Make the boundary logic Unicode-aware rather than ASCII-only, e.g.:

* treat any Unicode whitespace (at least U+3000 IDEOGRAPHIC SPACE) as a body terminator;
* treat CJK brackets/quotes as terminators the same way their ASCII counterparts are: `（）〔〕【】〈〉《》「」『』［］｛｝` (U+FF08/FF09, U+3008–U+3011, U+300C–U+300F, …);
* add CJK punctuation to the trailing-trim set: `、。，．！？：；・…` (U+3001, U+3002, U+FF0C, U+FF0E, U+FF01, U+FF1F, U+FF1A, U+FF1B, U+30FB, U+2026).

Note that a fullwidth-form check alone is not enough — `「」` and `、` are in the CJK Symbols and Punctuation block, not the Halfwidth and Fullwidth Forms block.

Related, already-fixed boundary bugs in the same scanner: stablyai/orca#8832 (next line's text glued on) and stablyai/orca#8126 (ANSI reset code glued to the URL). This one is the CJK counterpart, and it hits every Japanese/Chinese user whose agent annotates URLs inline — which Claude Code does constantly.

innocarpe:
## Summary

- Stop terminal HTTP links at CJK brackets, punctuation, and Unicode whitespace instead of swallowing adjacent prose.
- Reject non-ASCII characters while scanning the URL authority so missed boundaries cannot be folded into an IDNA hostname.
- Keep non-ASCII document paths and queries valid after the authority boundary.

Fixes #10571.

## Why

Terminal URL extraction previously recognized only ASCII boundaries. Text such as a Japanese annotation directly after a GitHub URL could be included in the link, producing a percent-encoded or IDNA-transformed destination instead of the URL the user intended.

## Scope and design

- The existing URL extraction and terminal hit-testing flow on main is preserved.
- Boundary predicates are isolated in terminal-http-url-boundary.ts so the scanner remains reviewable and below the module line budget.
- Unicode White_Space, punctuation, and symbol properties cover new scripts without an open-ended glyph list.
- Authority scanning stops at slash, query, or fragment; after that point, Unicode path/query text remains allowed.
- No network, filesystem, or URL-routing behavior changes beyond identifying the correct URL range.

## Test plan

- [x] CJK bracket, punctuation, and ideographic-space cases
- [x] Unicode White_Space and property-class terminators
- [x] Authority guard against bare CJK host text
- [x] Non-ASCII path preservation
- [x] Related terminal-link tests: 140 passed
- [x] Typecheck
- [x] Oxlint, React Doctor lint, Oxfmt, and diff checks
- [ ] Manual: paste the issue repro lines and verify only the URL is underlined/clickable

## Screenshots

No visual chrome changes.

## Security audit

The authority guard prevents missed Unicode boundaries from being silently IDNA-folded into a different hostname. The change adds no external requests, persistence, or new routing destination.

## Platform review

The scanner uses standard JavaScript Unicode property escapes and URL parsing already supported by the desktop terminal path. The behavior is platform-independent across macOS, Linux, and Windows.

## ELI5

If a terminal prints a URL followed immediately by Japanese or other Unicode notes, Orca now links only the real URL. Unicode text inside the URL path still works.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-http-url-boundary.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-http-url-boundary.ts
@@ -1,0 +1,108 @@
+/**
+ * Code-point boundary checks for terminal HTTP link scanning.
+ * Why extracted: keeps hit-testing under max-lines while hosting CJK-aware rules (#10571).
+ */
+
+export function isHttpUrlBodyTerminator(code: number): boolean {
+  return (
+    isUrlWhitespace(code) ||
+    isAsciiUrlBodyDelimiter(code) ||
+    // Why (#10571): CJK agents often glue fullwidth brackets/quotes to URLs
+    // with no ASCII space; treat those as terminators like (){}<> ASCII peers.
+    isCjkUrlBracketOrQuote(code)
+  )
+}
+
+export function isHttpUrlTrailingPunctuation(code: number): boolean {
+  return (
+    isUrlWhitespace(code) ||
+    isAsciiUrlTrailingPunctuation(code) ||
+    // Why (#10571): strip trailing CJK punctuation the same way we strip .,?!
+    isCjkUrlTrailingPunctuation(code)
+  )
+}
+
+function isUrlWhitespace(code: number): boolean {
+  // ASCII whitespace + U+3000 IDEOGRAPHIC SPACE (common after Japanese URLs).
+  return (
+    code === 9 ||
+    code === 10 ||
+    code === 11 ||
+    code === 12 ||
+    code === 13 ||
+    code === 32 ||
+    code === 0x3000
+  )
+}
+
+function isAsciiUrlBodyDelimiter(code: number): boolean {
+  return (
+    code === 0x22 || // "
+    code === 0x27 || // '
+    code === 0x21 || // !
+    code === 0x2a || // *
+    code === 0x28 || // (
+    code === 0x29 || // )
+    code === 0x7b || // {
+    code === 0x7d || // }
+    code === 0x7c || // |
+    code === 0x5c || // \
+    code === 0x5e || // ^
+    code === 0x3c || // <
+    code === 0x3e || // >
+    code === 0x60 // `
+  )
+}
+
+function isAsciiUrlTrailingPunctuation(code: number): boolean {
+  return (
+    isAsciiUrlBodyDelimiter(code) ||
+    code === 0x3a || // :
+    code === 0x2c || // ,
+    code === 0x2e || // .
+    code === 0x3f || // ?
+    code === 0x7e || // ~
+    code === 0x5b || // [
+    code === 0x5d // ]
+  )
+}
+
+/** Fullwidth and CJK bracket/quote code points that mirror ASCII URL delimiters. */
+function isCjkUrlBracketOrQuote(code: number): boolean {
+  return (
+    code === 0xff08 || // （
+    code === 0xff09 || // ）
+    code === 0xff3b || // ［
+    code === 0xff3d || // ］
+    code === 0xff5b || // ｛
+    code === 0xff5d || // ｝
+    code === 0x3008 || // 〈
+    code === 0x3009 || // 〉
+    code === 0x300a || // 《
+    code === 0x300b || // 》
+    code === 0x300c || // 「
+    code === 0x300d || // 」
+    code === 0x300e || // 『
+    code === 0x300f || // 』
+    code === 0x3010 || // 【
+    code === 0x3011 || // 】
+    code === 0x3014 || // 〔
+    code === 0x3015 // 〕
+  )
+}
+
+function isCjkUrlTrailingPunctuation(code: number): boolean {
+  return (
+    isCjkUrlBracketOrQuote(code) ||
+    code === 0x3001 || // 、
+    code === 0x3002 || // 。
+    code === 0x30fb || // ・
+    code === 0x2026 || // …
+    code === 0xff0c || // ，
+    code === 0xff0e || // ．
+    code === 0xff01 || // ！
+    code === 0xff1f || // ？
+    code === 0xff1a || // ：
+    code === 0xff1b // ；
+  )
+}
--- a/src/renderer/src/components/terminal-pane/terminal-http-url-boundary.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-http-url-boundary.ts
@@ -3,36 +3,39 @@
  * Why extracted: keeps hit-testing under max-lines while hosting CJK-aware rules (#10571).
  */
 
+// Why (#10571 review): enumerate glyphs forever misses 〜/〖〗/NBSP; property classes scale.
+const NON_ASCII_BODY_TERMINATOR_RE = /[\p{White_Space}\p{P}\p{Sm}]/u
+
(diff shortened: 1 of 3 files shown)


AmethystLiang:
Thanks for turning this around so fast — the diagnosis in #10571 is spot on, the ASCII-only predicates are exactly the bug.

I dug in before approving and think the fix needs a different shape. Running this PR's predicates against a wider case set, the punctuation-only half of `isCjkUrlTrailingPunctuation` — `、。・…，．！？：；`, everything it adds beyond `isCjkUrlBracketOrQuote` — never actually fires:

| line | this PR resolves to |
|---|---|
| `https://example.com、それから` | `https://example.xn--com-6y3b8ordxnwa/` |
| `https://example.com。次の行` | `https://example.com.xn--u9j508ocks/` |
| `https://example.com・次` | `https://example.xn--com-rs4b4090a/` |

Trailing-trim only strips from the end of an *already-terminated* candidate. `、` isn't in the body-terminator set and CJK letters don't terminate either, so `findHttpUrlCandidateEnd` runs to end of line; the last character is `ら`, which isn't trailing punctuation, so the trim loop strips nothing. The three lines in the issue pass because `（` and `「` are real body terminators — the `・` in `（作成済み・マージ待ち）` never gets exercised.

Worth flagging what that output actually is: with no path, the WHATWG parser folds the CJK into the **authority** and IDNA-encodes it, so the link silently retargets to a different, registrable domain rather than 404ing. I confirmed it against the live module in a running renderer. That makes this a link-spoofing surface, which is why I want to be careful here.

CodeRabbit's one nitpick on this PR was "add coverage for trailing CJK punctuation (`。`, `、`, `！`, `？`)" — that test is exactly what would have surfaced it.

Two changes I'd make:

1. Move CJK punctuation into the body-terminator set, and go property-based — `\p{White_Space}`, `\p{P}` and `\p{Sm}`, scoped to code points above U+007F — instead of enumerating. This is the third boundary bug in this scanner after #8126 and #8832, and each enumeration only covers what someone thought of: `〖〗` and `｟｠` are missing from the bracket list here, and `〜` U+301C is `Pd` while `～` U+FF5E is `Sm`, so even "all the CJK punctuation" wouldn't have caught them. I got that wrong myself on the first pass and had to check the categories rather than assume them.
2. Add a hostname guard so the authority case can't happen at all, independent of which terminators are in the set.

There's also a second detector in play: `pane-dom-creation.ts` constructs `WebLinksAddon` with no `urlRegex`, so the hover underline and tooltip come from xterm's default regex while the click re-resolves through `extractTerminalHttpLinks`. The two disagree in both directions today, which is why the underline stays wrong even with this patch applied.

That adds up to a good deal more than this PR signed up for, so I'm going to pick it up in a follow-up with you as co-author rather than ask you to build the whole thing. Hope that's alright — the report and repro were genuinely useful and made this much faster to work through.

innocarpe:
### Sync update (`7d9e85301`)

Address Amethyst review: property-based non-ASCII terminators + hostname guard against IDNA spoof; broader Unicode whitespace.

- Commit: `7d9e85301e0160d4bcef69a5fa58f1c1dbb0d16e`
- Upstream: https://github.com/stablyai/orca/pull/10573
- Fork portfolio: https://github.com/innocarpe/orca/pull/73

innocarpe:
Closing in favor of merged #15240, which already closes #10571.

Current `main` now terminates HTTP URLs at CJK punctuation / ideographic space in `terminal-http-url-extraction.ts` (`NON_ASCII_URL_TERMINATOR`). That is the same user-facing gap this PR targeted. Thank you @nwparker.

PR #10574, opened 2026-07-25
fix(sidebar): include folder workspaces in Cmd+Shift+Arrow cycling

jeongph:
## Summary

Folder workspaces could never be reached with `Cmd/Ctrl+Shift+↓/↑`, even though they render as sidebar rows and `Cmd/Ctrl+1–9` already navigates to them.

Repro (needs a folder-backed project group — import a parent directory containing several git repos as a group, then add a workspace from the group header's `+`):

```
sidebar (Group by project)     Cmd+Shift+↓ repeatedly
─────────────────────────      ─────────────────────────
1  feature-a   (worktree)  →   visited
2  design      (folder ws)  →  skipped   ← visible, unreachable
3  feature-b   (worktree)  →   visited
```

Cycling ran 1 → 3 → 1 → 3. Pressing `Cmd+2` went to the folder workspace fine, so the two shortcuts disagreed about which rows exist.

### Root cause

`navigateWorktree` built its own all-expanded row layout but passed `[]` for `buildRows`' `folderWorkspaces` argument, so `folder-workspace` rows were never constructed as navigation candidates. The main render path passes `visibleFolderWorkspacesForRows` and therefore does show them. `Cmd/Ctrl+1–9` reads `getRenderedWorktreesInSidebarOrder`, which folds `folder-workspace` rows in — hence the mismatch.

### Fix

- Thread the host-scope-filtered list down as a new `visibleFolderWorkspaces` prop (the viewport's existing `folderWorkspaces` prop is the unfiltered store list, which would let cycling jump to a row the sidebar is hiding).
- Build the navigable IDs with `getRenderedWorktreesInSidebarOrder` — the same ordering helper `Cmd+1–9` uses — instead of filtering to `item` rows and calling `getPreferredWorktreeRows` directly. Both shortcuts now cover the same rows in the same visual order, and pinned-row de-duplication still happens inside that helper.

Reveal/scroll and activation needed no changes: `renderRowContainsWorktree` already matches `folderWorkspaceKey(...)`, and `activateAndRevealWorktree` already routes folder workspace keys to `setActiveFolderWorkspace`.

Scope note: `buildRows` only emits project-group and folder-workspace rows when `groupBy === 'repo'`, so this affects the "Group by project" sidebar mode — the only mode where folder workspaces are visible at all.

## Screenshots

`No visual change` — only which workspace a keyboard shortcut activates.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

New component test `WorktreeList.arrow-cycling-folder-workspaces.test.tsx` mounts the real sidebar with a folder-backed project group, dispatches the actual chord, and asserts a full lap visits all three workspaces in both directions plus that cycling steps *off* a folder workspace anchor. Verified it is a real regression guard: reverting just the `folderWorkspaces` argument makes it fail with `expected [ 'wt-b', 'wt-a', 'wt-b', 'wt-a' ] to include 'folder:folder-1'`.

Full suite: `Test Files 9 failed | 3388 passed | 7 skipped`, `Tests 44 failed | 35978 passed | 60 skipped`. None are caused by this change:

- 8 of them are the same pre-existing failures documented in #10555 (`relay/agent-exec-handler`, `browser-pane/markup/use-markup-draw-hint`, three `right-sidebar/*`, and three `sidebar/*` — mostly `window.localStorage.* is not a function` in the shared test env).
- The 9th, `src/main/runtime/multi-client-navigation-isolation.integration.test.ts`, is flaky under load: it fails in the full run (`expected 'host-tab' to be 'client-a-tab'`) and passes 5/5 when run on its own. It exercises main-process client tab routing, which this renderer-only change cannot reach.

Sidebar suite on its own: 181 passed, only the 3 pre-existing sidebar failures.

## AI Review Report

Reviewed with Claude Code (Opus 5) against the checks in CONTRIBUTING:

- **Cross-platform (macOS/Linux/Windows):** no platform-dependent code. The chord still resolves through `keybindingMatchesAction` + `getShortcutPlatform()`; the new test pins the platform to `darwin` so it does not depend on the runner's userAgent.
- **SSH / remote / local:** the new prop is the host-scope-filtered list, so a folder workspace hosted on a host that is currently filtered out of the sidebar stays out of cycling. Using the unfiltered `folderWorkspaces` prop would have introduced exactly that bug; called out here because the two props now look similar at the call site.
- **Folder workspaces vs git worktrees:** this is the change — both are now navigable, ordered by `getRenderedWorktreesInSidebarOrder`. Folder workspace IDs (`folder:<id>`) already share the ID space used by the anchor logic, activation, and reveal.
- **Agent / integration / git-provider neutrality:** nothing provider-specific involved.
- **Performance:** no new subscriptions. `getRenderedWorktreesInSidebarOrder` runs on the same rows that were already being built for this keypress and is O(rows); the previous code did an equivalent filter + de-dup pass.
- **UI quality:** no visual surface touched; STYLEGUIDE does not apply.
- **Behavior change, intended:** cycling now has more stops than before for users with folder workspaces. That is the point of the fix, and it aligns arrow cycling with `Cmd+1–9`.

## Security Audit

No security-relevant surface. Renderer-side state computation over workspace IDs already in the store: no input parsing, command execution, path handling, auth, secrets, IPC, or network calls, and no new dependencies. The threaded list comes from store state that the sidebar already renders.

## Notes

- **Stacked on #10555.** This branch contains that PR's commit as its parent, so until #10555 merges the diff here shows both commits. Once it lands, this PR's diff reduces to the single folder-workspace commit. Reviewing #10555 first is the easier order.
- Follows up the pre-existing inconsistency flagged in #10555's Notes; CodeRabbit raised the same gap on the earlier revision of that PR.

X (Twitter): https://x.com/jeongph_dev

🤖 Generated with Claude Code

### DIFF
--- a/src/renderer/src/components/sidebar/WorktreeList.arrow-cycling-folder-workspaces.test.tsx
+++ b/src/renderer/src/components/sidebar/WorktreeList.arrow-cycling-folder-workspaces.test.tsx
@@ -1,0 +1,383 @@
+// @vitest-environment happy-dom
+
+import { act, type ReactNode } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
+import type {
+  FolderWorkspace,
+  ProjectGroup,
+  Repo,
+  Worktree,
+  WorktreeCardProperty
+} from '../../../../shared/types'
+import { folderWorkspaceKey } from '../../../../shared/workspace-scope'
+
+globalThis.IS_REACT_ACT_ENVIRONMENT = true
+
+const mockStore = vi.hoisted(() => ({
+  state: {} as Record<string, unknown>,
+  activateWorktreeFromSidebar: vi.fn(),
+  activateAndRevealWorktree: vi.fn(),
+  openModal: vi.fn()
+}))
+
+type WorktreeListComponent = React.ComponentType<{
+  scrollOffsetRef: React.RefObject<number>
+  scrollAnchorRef: React.RefObject<unknown>
+}>
+
+let WorktreeList: WorktreeListComponent
+
+vi.mock('@/store', () => {
+  const useAppStore = ((selector: (state: Record<string, unknown>) => unknown) =>
+    selector(mockStore.state)) as ((
+    selector: (state: Record<string, unknown>) => unknown
+  ) => unknown) & {
+    getState: () => Record<string, unknown>
+  }
+  useAppStore.getState = () => mockStore.state
+  return { useAppStore }
+})
+
+// Why: pin the platform so the chord below is deterministic instead of depending on the test env's userAgent.
+vi.mock('@/lib/shortcut-platform', () => ({
+  getShortcutPlatform: () => 'darwin'
+}))
+
+vi.mock('@tanstack/react-virtual', () => ({
+  defaultRangeExtractor: ({ startIndex, endIndex }: { startIndex: number; endIndex: number }) =>
+    Array.from({ length: endIndex - startIndex + 1 }, (_, index) => startIndex + index),
+  measureElement: () => 32,
+  useVirtualizer: ({ count }: { count: number }) => ({
+    elementsCache: new Map(),
+    getTotalSize: () => count * 96,
+    getVirtualItems: () =>
+      Array.from({ length: count }, (_, index) => ({
+        index,
+        key: `row-${index}`,
+        start: index * 96
+      })),
+    measureElement: vi.fn(),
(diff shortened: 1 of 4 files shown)


jeongph:
Folding this into #10555 instead of stacking. Both are plain bug fixes in the same block of `navigateWorktree`, and a stacked pair would have asked maintainers to honor a merge order — not a reasonable thing to expect on an OSS PR. #10555 now carries both commits. (#10513 stays separate on purpose: it is a behavior/policy choice, not a bug fix.)

PR #10580, opened 2026-07-25
fix(composer): autosize Create Worktree note on PR prefill

### LINKED ISSUE #10575: [Bug]: Create Worktree note field clips the PR title it prefills, with no way to scroll or resize
### Operating system

macOS

### Orca version

1.4.156-rc.1

### Details

**Short summary:**
In the Create Worktree dialog, linking an existing PR auto-fills the Note field with the PR title, but the Note box never grows to fit it. The text is clipped and there is no way to see the rest of it.

**What happened?**

Picking a PR under the Smart tab prefills the Note with `PR #<number> — <title>`. PR titles routinely wrap to two or more lines, but the Note box stays at its one-row height, so everything past the first line is invisible.

There is no way to recover the hidden text:

- the field is `resize-none`, so there is no drag handle to enlarge it
- the field is `overflow-hidden`, so there is no scrollbar and the mouse wheel does nothing
- focusing the field and pressing Down moves the caret, which scrolls the content into view, but the box is still one row tall so the note stays clipped

**How can we reproduce it?**

1. Click **+** on a project in the sidebar to open **Create Worktree**.
2. On the **Smart** tab, pick an existing PR whose title is long enough to wrap (roughly 60+ characters).
3. Expand **Advanced**.
4. Look at the **Note** field — it shows only the first line of `PR #<number> — <title>`, with the rest clipped.
5. Try to reveal the rest: there is no resize handle, no scrollbar, and the wheel does not scroll. Arrow keys move the caret but the box never grows.

The auto-size logic for this field lives only in the textarea's `onInput` handler (`src/renderer/src/components/NewWorkspaceComposerCard.tsx`). `onInput` fires only for text the user types or pastes — it does not fire when React writes the value from state. The PR prefill is exactly that kind of programmatic write (`setNote(...)` in `src/renderer/src/hooks/useComposerState.ts`), so the resize never runs and the box keeps its `rows={1}` height.

Scope notes:

- **PRs and GitLab MRs only.** GitHub issues and Linear issues do not prefill the Note at all, so they don't hit this. Typing a long note by hand also works fine, because typing does fire `onInput`.
- The equivalent Note field in **Edit Worktree Details** (`WorktreeMetaDialog`) is not affected — it sizes itself from a ref callback on mount and uses `overflow-y-auto`.

innocarpe:
## Summary

Fixes #10575.

Create Worktree prefills the Note from a selected PR or GitLab MR through React state. The previous textarea only resized from `onInput`, so programmatic prefills stayed at `rows={1}` and `overflow-hidden` made the wrapped title unreachable.

- Size the controlled textarea from its content with `[field-sizing:content]`, matching Orca's existing layout-driven textarea pattern.
- Keep the compact one-row empty state and `max-h-40` cap, then allow overflow to scroll with `overflow-y-auto scrollbar-sleek`.
- Cover prefilled and height-capped notes in the rendered composer regression tests.

## Screenshots

**Before** — the prefilled title is clipped to one row.

Create Worktree before the fix, with the prefilled PR title clipped in the Note field

**After** — the same title grows to its wrapped height.

Create Worktree after the fix, with the complete prefilled PR title visible in the Note field

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Focused validation:

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/NewWorkspaceComposerCard.test.tsx` — 24 passed
- `pnpm exec oxlint src/renderer/src/components/NewWorkspaceComposerCard.tsx src/renderer/src/components/NewWorkspaceComposerCard.test.tsx --deny-warnings` — passed
- Electron: verified a real Smart-tab PR prefill, an empty note, shrink-back after clearing, the 160px cap with scrolling, and collapsed-to-expanded Advanced state

## AI Review Report

The review confirmed the reported root cause, then replaced the initial ref, layout effect, input handler, and shared measurement helper with CSS content sizing. This keeps height derived from layout for programmatic state writes, typing, pane resizing, zoom, and font reflow without synchronous `scrollHeight` reads.

Risks checked:

- Programmatic PR and GitLab MR prefills grow without an input event.
- Empty and cleared notes remain one row tall.
- Long notes stop at `max-h-40` and remain reachable through the styled scrollbar.
- The field sizes correctly while Advanced is collapsed and after it expands.
- The class-contract tests fail on the pre-fix markup; pixel behavior was checked in Electron because happy-dom has no layout engine.
- The change is renderer-only and behaves the same in Orca's bundled Chromium on macOS, Linux, and Windows. It does not touch shortcuts, labels, paths, shells, Git, remote execution, or workspace hosting.

## Security Audit

No new input parsing, command execution, filesystem access, authentication, secrets, dependencies, IPC, or remote-wire surface. The existing controlled note value and paste limits are unchanged; this only changes how the owned textarea is sized and how overflow is displayed. No follow-up required.

## Notes

- GitHub PRs and GitLab MRs are the affected prefill paths; GitHub issues and Linear issues do not prefill this Note.
- The bundled Electron runtime supports `field-sizing: content`. Other web runtimes without that CSS support still gain scroll access instead of clipping content behind `overflow-hidden`.
- No behavior changes for SSH hosts, folder workspaces, or git worktrees.


innocarpe:
Status ping: small Create Worktree note autosize on PR prefill (3 files, tests). Clean review surface if it helps prioritization.

brennanb2025:
Thanks for this — your diagnosis is spot on: `onInput` never fires when React writes the value from state, so the prefilled title kept its one-row height.

Heads up, I've pushed a commit to your branch.

- Swapped the JS measure pass for `[field-sizing:content]`. Orca already sizes its growable textareas this way (native chat composer, Linear title editor), and letting the layout engine own the height also covers the cases a value-keyed effect can't see — re-wrap when the panel is resized, or when a webfont finishes loading.
- That let the ref, the `useLayoutEffect`, the `onInput` handler and the `autosize-textarea-height` helper all come out, so the fix is now just the textarea's class list.
- Added `scrollbar-sleek` next to your `overflow-y-auto` so the gutter past `max-h-40` matches the Comment field in Edit Worktree Details.
- Moved the test into `NewWorkspaceComposerCard.test.tsx` so it asserts the rendered field. The old unit test asserted the two style assignments it had just made, so it stayed green with the bug still present; the new one goes red on the pre-fix markup.

Your `overflow-hidden` → `overflow-y-auto` change ships as-is — the "no scrollbar, wheel does nothing" half of the report needed it.

Checked in the app: picking a real PR on the Smart tab now shows the full title, a long note still clamps at the same height and scrolls, and an empty note is still one row.

innocarpe:
@brennanb2025 Thanks for the follow-up commit — `field-sizing:content` is clearly the better fit here.

I agree the JS measure path was fragile for value-driven prefill, and aligning with the native chat / Linear title pattern (plus `scrollbar-sleek` past `max-h-40`) is cleaner than keeping a ref/`useLayoutEffect` helper that only covered part of the cases. Moving the test onto the rendered composer field is also the right regression shape — the old unit test would have stayed green with the bug intact.

Happy to land your tip as-is.

brennanb2025:
Thanks for the PR!

PR #10582, opened 2026-07-25
fix(git): resolve upstream for branches whose name is an ambiguous ref

egoing:
## Summary

**Symptom:** on some branches, clicking **Create PR** in Source Control never renders the PR form — only a "Review setup needs attention." notice appears, even though the branch is pushed and correctly tracked. `getHostedReviewCreationEligibility` reports `blockedReason: "no_upstream"`.

**Root cause:** `getCurrentBranchName` (`src/shared/git-effective-upstream.ts`) read the branch via `git symbolic-ref --quiet --short HEAD`. `--short` does not mean "strip `refs/heads/`" — it returns the **shortest _unambiguous_** ref. When another ref shares the branch's name (most commonly a same-named tag, e.g. a `v1.0` branch plus a `v1.0` tag), git must disambiguate and returns `heads/v1.0` instead of `v1.0`.

That `heads/`-prefixed name was then used verbatim to query `branch.<name>.remote` / `branch.<name>.merge` and to probe `refs/remotes/origin/<name>`. Real git config is keyed by the true branch name, so every lookup missed, `resolveEffectiveGitUpstreamForBranch` fell through every path, and `hasUpstream` came back `false`.

The fix reads the full ref (`git symbolic-ref --quiet HEAD`) and strips `refs/heads/` in JS, so the name is always exact regardless of what other refs exist.

> **Correction to the original diagnosis.** This PR was originally filed as a non-ASCII/UTF-8 byte-truncation bug. That mechanism could not be reproduced: on git 2.44.0 and 2.50.1, across `en_US.UTF-8`, `C`, `POSIX`, `ko_KR.UTF-8`, and four ISO-8859 locales, `--short` returned the reporter's 63-byte Korean branch name **byte-for-byte intact**:
>
> ```
> $ git symbolic-ref --quiet --short HEAD | xxd
> 00000000: 6567 6f69 6e67 2f41 4745 4e54 532e 6d64  egoing/AGENTS.md
> 00000010: 2deb a5bc 2dec 839d ec84 b1ed 9598 eab3  -...-...........
> 00000020: a02d ec9c 84ed 82a4 2dec a095 ecb1 85ec  .-......-.......
> 00000030: 9d84 2dec 8898 eba6 bded 9598 eab8 b00a  ..-.............
> ```
>
> The one-line code change is unchanged and correct — but it fixes ref **ambiguity**, not byte truncation, and non-ASCII is incidental. The source comment and tests were rewritten to describe the real mechanism, because a wrong comment would send the next reader chasing a nonexistent encoding bug.

## ELI5

Git has a shortcut for asking "what branch am I on?" that gives back the shortest name that still points at exactly one thing. If you happen to have a tag with the same name as your branch, that shortcut has to get more specific and hands back `heads/my-branch` instead of `my-branch`.

Orca took that answer at face value and went looking for settings filed under `heads/my-branch`. Nothing is filed under that name, so Orca concluded the branch had never been published and refused to show the Create PR form. Now Orca asks for the full, unambiguous name and trims it itself, so it always looks in the right place.

## Proof of fix

Root cause demonstrated against a real git binary — a branch and tag both named `v1.0`:

```
$ git symbolic-ref --quiet --short HEAD
heads/v1.0                                 # <-- what main fed into config lookups
$ git symbolic-ref --quiet HEAD
refs/heads/v1.0                            # <-- what the fix uses -> "v1.0"

$ git config --get "branch.heads/v1.0.remote"; echo "exit=$?"
exit=1                                     # main's lookup MISSES
$ git config --get "branch.v1.0.remote"; echo "exit=$?"
origin
exit=0                                     # the fix's lookup HITS
```

The new tests drive `resolveEffectiveGitUpstream` against **real git repos** created in `tmpdir` (init → commit → real remote → publish), not string fixtures.

**On `origin/main`'s source, with the new tests — FAILS:**

```
 × resolveEffectiveGitUpstream against real git > resolves the upstream when a same-named tag makes the branch ref ambiguous 1061ms
 × resolveEffectiveGitUpstream > reads the full HEAD ref rather than the abbreviated form 1ms

 Test Files  1 failed (1)
      Tests  2 failed | 8 passed (10)
```

**With this PR's source — PASSES:**

```
 Test Files  1 passed (1)
      Tests  10 passed (10)
```

The suite also round-trips ASCII, Korean, Japanese, Chinese, Cyrillic, emoji, and double-quote branch names byte-for-byte through the real code path, and asserts detached HEAD still resolves to `null`.

## Proof of no regressions

Full dependent suite — every test file touching upstream resolution across `shared`, `main/git`, and `relay`:

```
    src/shared/git-effective-upstream.test.ts src/shared/git-upstream-status.test.ts \
    src/shared/git-remote-error.test.ts src/shared/git-remote-identity.test.ts \
    src/shared/git-config-snapshot-runner.test.ts src/shared/git-push-target-validation.test.ts \
    src/shared/git-history.test.ts src/main/git/upstream.test.ts \
    src/main/git/branch-rename.test.ts src/main/git/remote.test.ts src/main/git/status.test.ts \
    src/main/git/status-upstream-negative-cache.test.ts src/main/git/status-upstream-probe-churn.test.ts \
    src/relay/git-status-upstream-negative-cache.test.ts src/relay/git-handler-status-ops.test.ts

 Test Files  15 passed (15)
      Tests  237 passed (237)
```

Typecheck (all three projects) and lint:

```
$ npx tsc --noEmit -p config/tsconfig.node.json     -> exit 0
$ npx tsc --noEmit -p config/tsconfig.tc.cli.json   -> exit 0
$ npx tsc --noEmit -p config/tsconfig.tc.web.json   -> exit 0
$ npx oxlint                                        -> exit 0
$ npx oxfmt --check <changed files>  -> All matched files use the correct format.
```

**Defect found and fixed in the original patch:** the PR as submitted broke **4 pre-existing tests** in `src/main/git/remote.test.ts`, which assert the exact argv of every spawned git command. Those tests pass 34/34 on `origin/main` and failed 4/34 with the original diff. Updated the `symbolic-ref` argv expectations and the corresponding mock stdout to the full-ref form; `remote.test.ts` is back to **34/34**.

No user-visible behavior changes beyond the fix. For any branch whose name was already unambiguous, `--short` and `refs/heads/`-stripping return the identical string, so the resolved upstream is byte-identical — asserted explicitly by the ASCII round-trip case. Detached HEAD is unchanged: `symbolic-ref` exits non-zero either way and the `catch` returns `null` (covered by a new real-git test).

## Screenshots

No visual change to markup or styling. The user-visible effect is that the Create PR form now renders instead of the "Review setup needs attention." notice; the before/after screenshots from the original report are preserved below.

**Before:** the Create PR panel shows the **Publish Branch** button and a "Review setup needs attention." notice — no title/description/base form.

image

**After:** the same state renders the normal Create PR form (title, description, base selector, draft toggle, Create PR button).

## Testing

- [x] `pnpm lint` — ran `npx oxlint`, exit 0. Pre-existing warnings remain in files this PR does not touch (`agent-browser-bridge.test.ts`, `ssh-git-provider.test.ts`); they reproduce identically on pristine `main`.
- [x] `pnpm typecheck` — ran all three projects (`tsconfig.node.json`, `tsconfig.tc.cli.json`, `tsconfig.tc.web.json`) individually, all exit 0.
- [ ] `pnpm test` — **not run in full** (30+ min suite). Ran the 15 test files covering upstream resolution across `shared`, `main/git`, and `relay` instead: 237/237 pass. See "Proof of no regressions".
- [ ] `pnpm build` — not run.

## AI Review Report

Reviewed as a hostile maintainer review, verifying claims against a **real git binary** rather than trusting the PR description.

Main risks checked and what was found:

- **Stated root cause was wrong** — flagged and corrected. The non-ASCII truncation mechanism did not reproduce on git 2.44.0 or 2.50.1 across 8 locales; `--short` returned the reporter's exact 63-byte Korean name intact (hexdump in Summary). Bisected to the real trigger: `--short` returns the shortest *unambiguous* ref, so a same-named tag yields `heads/<name>`. Verified the full end-to-end failure with a real repo, real remote, and real config lookups. The code change stands; the comment and tests were rewritten so the next reader isn't sent chasing an encoding bug.
- **Broke existing tests** — found and fixed. `src/main/git/remote.test.ts` asserts exact git argv; the original diff failed 4 of its 34 tests. Confirmed those 4 pass on pristine `main`, so the breakage was introduced by this PR. Fixed and re-verified 34/34.
- **ASCII regression risk** — explicitly asserted. Added a byte-for-byte round-trip case for a plain ASCII branch; unambiguous names are provably unchanged through the new path.
- **Prefix-stripping correctness** — the strip is conditional (`ref.startsWith('refs/heads/')`), so a HEAD symref outside `refs/heads/` is passed through rather than mangled. Verified against real git: pointing HEAD at `refs/custom/thing` returns the full ref, and the conditional leaves it intact.
- **Detached HEAD** — verified with real git that `symbolic-ref` exits 1 with and without `--short`, so the `catch → null` path is unchanged. Added a real-git test.
- **macOS NFC/NFD normalization** — checked, since APFS decomposes filenames and refs can be stored as loose files. Created a precomposed (NFC) Hangul branch and read it back through both loose refs and after `git pack-refs --all`: identical NFC bytes both times. Git writes ref names as raw bytes and does not normalize, so no NFC/NFD hazard on either storage form.
- **Branch names containing `->`** — verified `arrow->name` round-trips through `symbolic-ref` cleanly. This code path reads `symbolic-ref`, never the `HEAD -> main` arrow syntax of `branch -vv`/porcelain output, so there is nothing to mis-split.
- **All callers audited** — 6 call sites across `src/main/git/{status,branch-rename,remote,upstream}.ts`, `src/relay/{git-handler,git-status-upstream-negative-cache}.ts`. None re-quote, re-strip, or double-decode the returned name; all pass it straight into `branch.<name>.*` lookups or ref paths, which is exactly what the fix makes correct.
- **`core.quotePath` / C-style octal quoting** — considered and ruled out. `quotePath` governs **path** output (`status`, `ls-files`, `diff`), not ref names from `symbolic-ref`. Confirmed empirically: no `\NNN` escapes or wrapping quotes in the output for any non-ASCII branch tested, so no unquoting step is needed.

**Cross-platform (macOS/Linux/Windows):** explicitly checked. The change is one git flag removal plus pure JS string-prefix stripping — no shell invocation, no path handling (`refs/heads/` is a git-internal ref name, always `/`-separated on every OS, so `path.join` is not applicable), no keyboard/accelerator surface, no Electron platform branching. `git symbolic-ref HEAD` has behaved identically across git-for-Windows, Linux, and macOS since long before the Git 2.25 baseline, and the full-ref form is *older and more primitive* than `--short`, so no capability gate or fallback is needed. Executed on macOS (git 2.44.0 and 2.50.1); Windows/Linux not executed in this session — see Notes.

## Security Audit

Low risk. No new subprocess spawn, no new IPC surface, no new user input reaching a command line. The change **removes** an argument (`--short`) from an existing invocation and adds a pure string operation on its output.

Reviewed specifically:

- **Command execution** — argument count goes down, not up. All args remain fixed literals; nothing user-controlled is interpolated into the command. Invocation still flows through the existing `runGit`/`execFile`-style abstraction (argv array, no shell), so quoting and injection exposure is unchanged.
- **Input handling** — `getCurrentBranchName`'s output was already treated as untrusted and flows into downstream git command construction; this diff changes only *how the value is obtained*, not how it is consumed. Strictly speaking it makes the value **more** faithful to git's actual state.
- **Path handling** — none. `refs/heads/` is a ref namespace, not a filesystem path, and is never joined against a filesystem root here.
- **Secrets / auth / dependencies** — untouched. No new dependencies.

## Notes

- **Same bug still present in two sibling call sites, deliberately out of scope.** `getConfiguredPushTarget` (`src/main/git/remote.ts:21`), `src/relay/git-handler-push-target.ts:18`, and `src/shared/git-history.ts:91` all still use `symbolic-ref --quiet --short HEAD` and will mis-resolve for the same ambiguous-ref case. They are on the push/history paths with different failure semantics and their own argv-asserting tests, so fixing them belongs in a separate PR. Filing that follow-up is recommended.
- **`getConfiguredUpstream` still uses `rev-parse --abbrev-ref HEAD@{u}`** and can still fail on this class of branch. That failure is caught by the existing `isNoUpstreamError` check and fully recovered by the `getConfiguredBranchRemoteUpstream` fallback now that it receives the correct branch name — behavior is correct end-to-end, just via the fallback rather than the primary path.
- **Platform execution coverage:** verified at runtime on macOS only. No Windows or Linux machine was available in this session. Risk is assessed as low (see AI Review Report), but Windows in particular differs on console codepage and encoding defaults, so a Windows smoke test before merge would be worthwhile.
- **SSH / remote hosts:** the fix removes a local-locale dependency rather than adding one — see Compatibility.

## Compatibility

- **Platforms:** macOS — executed against real git 2.44.0 (Homebrew) and 2.50.1 (Apple Git-155), including NFC/NFD ref-storage checks on APFS through both loose refs and `packed-refs`. Linux/Windows — not executed; reasoned safe because the diff is one flag removal plus JS string handling, with no path, shell, or platform-branching surface. `symbolic-ref HEAD` predates the Git 2.25 baseline and needs no capability gate or fallback. Windows encoding/codepage differences are the main residual unknown (flagged in Notes).
- **SSH / remote:** safe, and strictly an improvement. Only the argv passed to the existing `runGit` abstraction changed, so native/WSL/SSH/relay host routing is untouched. The old code's correctness depended on the *remote* host's ref set (whether some other ref shadowed the branch name); the new code returns the branch's canonical full ref, which is independent of the remote's git version and locale. Nothing assumes local execution or local locale.
- **Performance:** no hot-path change. Same single `symbolic-ref` spawn as before — one flag fewer, one `String.prototype.slice` more. No added subprocess spawns, no added I/O, no polling or re-render impact.

Original patch by @egoing.

### DIFF
--- a/src/main/agent-hooks/first-work-branch-rename.test.ts
+++ b/src/main/agent-hooks/first-work-branch-rename.test.ts
@@ -463,7 +463,7 @@ describe('maybeAutoRenameBranchOnFirstWork', () => {
         throw noUpstreamError
       }
       if (args[0] === 'symbolic-ref') {
-        return { stdout: 'you/Nautilus\n', stderr: '' }
+        return { stdout: 'refs/heads/you/Nautilus\n', stderr: '' }
       }
       if (args[0] === 'rev-parse' && args.includes('refs/remotes/origin/you/Nautilus')) {
         throw new Error('not found')
@@ -486,7 +486,7 @@ describe('maybeAutoRenameBranchOnFirstWork', () => {
         return { stdout: 'you/Nautilus\n', stderr: '' }
       }
       if (args[0] === 'symbolic-ref') {
-        return { stdout: 'you/Nautilus\n', stderr: '' }
+        return { stdout: 'refs/heads/you/Nautilus\n', stderr: '' }
       }
       if (args[0] === 'rev-parse' && args.some((arg) => arg.includes('@{u}'))) {
         upstreamReadCount += 1
--- a/src/main/git/branch-rename.test.ts
+++ b/src/main/git/branch-rename.test.ts
@@ -16,7 +16,7 @@ describe('probeBranchUpstream', () => {
   it('reports has-upstream when @{u} resolves to a tracking ref', async () => {
     const exec: GitExec = vi.fn(async (args: string[]) => {
       if (args[0] === 'symbolic-ref') {
-        return { stdout: 'feature\n', stderr: '' }
+        return { stdout: 'refs/heads/feature\n', stderr: '' }
       }
       if (args[0] === 'rev-parse' && args.includes('HEAD@{u}')) {
         return { stdout: 'origin/feature\n', stderr: '' }
@@ -29,7 +29,7 @@ describe('probeBranchUpstream', () => {
   it('reports no-upstream when there is no upstream', async () => {
     const exec: GitExec = vi.fn(async (args: string[]) => {
       if (args[0] === 'symbolic-ref') {
-        return { stdout: 'feature\n', stderr: '' }
+        return { stdout: 'refs/heads/feature\n', stderr: '' }
       }
       if (args[0] === 'rev-parse' && args.includes('HEAD@{u}')) {
         throw noUpstreamError
@@ -45,7 +45,7 @@ describe('probeBranchUpstream', () => {
   it('reports has-upstream when a same-name origin tracking ref exists without configured upstream', async () => {
     const exec: GitExec = vi.fn(async (args: string[]) => {
       if (args[0] === 'symbolic-ref') {
-        return { stdout: 'feature\n', stderr: '' }
+        return { stdout: 'refs/heads/feature\n', stderr: '' }
       }
       if (args[0] === 'rev-parse' && args.includes('HEAD@{u}')) {
         throw noUpstreamError
@@ -84,7 +84,7 @@ describe('probeBranchUpstream', () => {
     // A gettext-enabled git under de_DE translates even the `fatal:` prefix.
     const exec: GitExec = vi.fn(async (args: string[]) => {
       if (args[0] === 'symbolic-ref') {
-        return { stdout: 'feature\n', stderr: '' }
+        return { stdout: 'refs/heads/feature\n', stderr: '' }
       }
       throw new Error(
         'Command failed: git rev-parse --abbrev-ref HEAD@{u}\n' +
--- a/src/main/git/remote.test.ts
+++ b/src/main/git/remote.test.ts
@@ -59,6 +59,38 @@ describe('git remote operations', () => {
     )
   })
 
+  it('keeps the fork push target when a same-named tag makes the branch ref ambiguous', async () => {
+    // Why: `symbolic-ref --short` abbreviates an ambiguous HEAD to `heads/v1.0`,
+    // so every `branch.<name>.*` lookup missed and the push fell back to
+    // `origin HEAD`, publishing review commits to the upstream repository.
+    gitExecFileAsyncMock.mockImplementation(async (args: string[]) => {
+      if (args[0] === 'symbolic-ref') {
+        expect(args).not.toContain('--short')
+        return { stdout: 'refs/heads/v1.0\n', stderr: '' }
+      }
+      if (args[0] === 'config' && args.includes('branch.v1.0.remote')) {
+        return { stdout: 'fork\n', stderr: '' }
+      }
+      if (args[0] === 'config' && args.includes('branch.v1.0.pushRemote')) {
+        return { stdout: 'fork\n', stderr: '' }
+      }
+      if (args[0] === 'config' && args.includes('branch.v1.0.merge')) {
+        return { stdout: 'refs/heads/v1.0\n', stderr: '' }
+      }
+      if (args[0] === 'config') {
+        throw new Error('missing key')
+      }
+      return { stdout: '', stderr: '' }
+    })
+
+    await gitPush('/repo', false)
+
+    expect(gitExecFileAsyncMock).toHaveBeenLastCalledWith(
+      ['push', '--set-upstream', 'fork', 'HEAD:v1.0'],
+      { cwd: '/repo' }
+    )
+  })
+
   it('does not combine remote.pushDefault with a base-branch merge target', async () => {
     gitExecFileAsyncMock.mockImplementation(async (args: string[]) => {
       if (args[0] === 'symbolic-ref') {
--- a/src/main/git/remote.test.ts
+++ b/src/main/git/remote.test.ts
@@ -350,14 +350,14 @@ describe('git remote operations', () => {
 
   it("runs pull with the user's configured strategy", async () => {
     gitExecFileAsyncMock
-      .mockResolvedValueOnce({ stdout: 'feature\n', stderr: '' })
+      .mockResolvedValueOnce({ stdout: 'refs/heads/feature\n', stderr: '' })
       .mockResolvedValueOnce({ stdout: 'origin/feature\n', stderr: '' })
       .mockResolvedValueOnce({ stdout: '', stderr: '' })
 
     await gitPull('/repo')
 
     expect(gitExecFileAsyncMock.mock.calls).toEqual([
-      [['symbolic-ref', '--quiet', '--short', 'HEAD'], { cwd: '/repo' }],
+      [['symbolic-ref', '--quiet', 'HEAD'], { cwd: '/repo' }],
       [['rev-parse', '--abbrev-ref', 'HEAD@{u}'], { cwd: '/repo' }],
       [['pull'], { cwd: '/repo' }]
     ])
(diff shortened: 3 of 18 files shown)


nwparker:
## Retitled — the original diagnosis didn't survive verification

I changed the title from "stop truncating non-ASCII branch names when resolving upstream" to **"resolve upstream for branches whose name is an ambiguous ref"**, because that's what the code actually fixes and what the body itself proves.

The real defect, confirmed against a real repo: `symbolic-ref --short HEAD` returns the shortest *unambiguous* ref, so when a tag shares a branch's name it abbreviates to `heads/v1.0` instead of `v1.0`. Nothing matches `branch.<name>.*` config or a remote-tracking ref after that, so upstream resolution returns `null`. Reading the full ref and stripping `refs/heads/` yields `origin/v1.0` correctly. Non-ASCII was a red herring — the byte handling was never the problem.

One body statement should also be corrected: it says macOS Git "writes ref names as raw bytes and does not normalize." Under this repo's own `git init` defaults that's false — `core.precomposeUnicode=true`, so creating a branch with an NFD name produces canonical **NFC** bytes in `.git/HEAD`, the loose ref filename, `symbolic-ref` output, and `packed-refs` alike. Doesn't affect the fix; it's just an inaccurate explanation.

Verification that did hold up: `core.quotePath=true` produced no C-style quoting for Korean, Japanese, Cyrillic, emoji, quote-containing, or arrow-containing refs — byte-identical to `quotePath=false`, so quotePath is genuinely irrelevant here rather than being papered over. ASCII round-trips byte-identically, detached HEAD returns `null`, and a branch name containing a space is rejected by `git check-ref-format` so it can't be exercised.

OrcaWin:
Reviewed with an independent reproduction against real git on both 2.50.1 and the Git 2.25 baseline. **Your corrected diagnosis is exactly right** — `symbolic-ref --short HEAD` returns the shortest *unambiguous* ref, so a branch sharing a name with a tag yields `heads/v1.0` and every `branch.<name>.*` / `refs/remotes/...` lookup misses. I reproduced the user-visible symptom in the running app (Orca offering "Publish Branch" for an already-published branch) and watched it recover with the fix.

Two things grew the diff, and I want to be upfront that it is now substantially larger than your one-liner.

**1. The deferred call sites are worse than the titled symptom.** Your Notes list three siblings with the identical defect and defer them. I think that deferral is wrong: on a branch tracking a contributor fork, the wrong branch name makes `getConfiguredPushTarget` return null, so Orca pushes `--set-upstream origin HEAD` — **sending review commits to the wrong repository**. That is a data-integrity bug, not a cosmetic one. I extracted `src/shared/git-current-branch-name.ts` and routed all four sites through it.

**2. Eight test files were mocking git incorrectly.** They returned an abbreviated branch name for `symbolic-ref`, which real git never does — they passed only because production had the same bug. Corrected so they exercise the real contract. Mocks for `refs/remotes/origin/HEAD` are untouched; that is a genuinely different lookup.

I also fixed a defect in my own first pass, caught by an adversarial reviewer: the shared helper originally passed a non-`refs/heads/` symref through unchanged. With `git symbolic-ref HEAD refs/custom/thing`, upstream resolution then probes `refs/remotes/origin/refs/custom/thing` and can match a bogus upstream, while history synthesizes `refs/heads/refs/custom/thing`. Git itself refuses that state ("HEAD not found below refs/heads!"), so it now returns null — detached.

**One correction to the PR body:** the severity is overstated. The bug only bites when `branch.<name>.merge` is unset (published without `-u`) — with `-u` the primary `HEAD@{u}` path masks it. Still real, since Orca's first-publish path leaves exactly that state.

Also pinned git config in the new real-git test helper: an ambient `commit.gpgsign=true` kills it with `fatal: failed to write commit object`.

Perf is a non-issue with evidence rather than assertion: identical subprocess count (4 vs 4), one argv element swapped for one string slice. No capability gate needed — `symbolic-ref HEAD` is older and more primitive than `--short`, so this is a capability *downgrade* that cannot fail where the old command worked.

Validation: 4859 tests passing. Two failures in `src/relay/agent-exec-handler.test.ts` are pre-existing on main and untouched by this PR.

Excellent root-cause work — the `--short` semantics here are genuinely counterintuitive and you found the real mechanism.

Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17572. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @egoing. No action is requested from you. Thank you.

PR #10586, opened 2026-07-25
fix(computer-use): probe Linux editables via get_editable_text_iface

### LINKED ISSUE #10569: [Bug]: Orca's computer use `set-value` fails with `'Accessible' object has no attribute 'is_editable_text'` on Linux Mint 22.3 x86_64 (Ubuntu 24.04)
### Operating system

Linux

### Orca version

v1.4.152

## Description

`orca computer set-value --app <app> --element-index <n> --value "test"` always fails with:

```
accessibility_error: 'Accessible' object has no attribute 'is_editable_text'
```

The error is raised from `runtime.py` line 972 — the `set_value()` function calls `node.is_editable_text` which does not exist as a Python `Atspi.Accessible` method.

## How to Reproduce

```bash
# 1. List a running app
orca-ide computer list-apps --json

# 2. Get the app state (accessibility tree)
orca-ide computer --app 'Brave Browser' get-app-state --json

# 3. Try to set a value on the frame element
orca-ide computer --app 'Brave Browser' set-value --element-index 0 --value "test" --json
```

**Expected**: `ok: true` — the value is set on the editable element.

**Actual**:
```json
{
  "ok": false,
  "error": {
    "code": "accessibility_error",
    "message": "'Accessible' object has no attribute 'is_editable_text'"
  }
}
```

## Root Cause

The `set_value()` function in `/opt/Orca/resources/computer-use-linux/runtime.py` (line 972) calls:

```python
if node is not None and bool(attempt(node.is_editable_text, False)):
```

The `attempt()` helper catches all `Exception` (including `AttributeError`) and returns `False` when the method doesn't exist.

**`is_editable_text`** exists in the **C** library (`atspi_accessible_is_editable_text` in `libatspi`), but is **not** exposed as a Python-callable method in the GObject Introspection bindings for `at-spi2-core` (package `gir1.2-atspi-2.0`).

The `Atspi.Accessible` Python wrapper has:
- ✅ `get_editable_text()` → returns `str`
- ✅ `get_editable_text_iface()` → returns `Atspi.EditableText` interface or `None`
- ❌ `is_editable_text()` → **not exposed** (though C API has it)

```python
# Works — it's in the Python bindings:
node.get_editable_text_iface() is not None  # ← returns True if editable
# Fails — not in Python bindings:
node.is_editable_text  # ← AttributeError
```

## Environment

| Field | Value |
|-------|-------|
| OS | Ubuntu 24.04 LTS (noble) |
| Desktop | XFCE 4 |
| Python | 3.13.13 (miniforge3/conda) |
| `at-spi2-core` | 2.52.0-1build1 |
| `gir1.2-atspi-2.0` | 2.52.0-1build1 |
| `libatspi` | 2.52.0-1build1 (has C function `atspi_accessible_is_editable_text`) |
| Orca version | latest (stablyai/orca) |
| Orca runtime | `/opt/Orca/resources/computer-use-linux/runtime.py` |
| `GI_TYPELIB_PATH` | unset (uses default system `girepository-1.0`) |

## Suggested Fix

**In `runtime.py`**, replace line 972:

```python
# Current (broken):
if node is not None and bool(attempt(node.is_editable_text, False)):
```

with:

```python
# Fix:
if node is not None and node.get_editable_text_iface() is not None:
```

This uses `get_editable_text_iface()` which **is** available in the Python GIR bindings. Both methods do the same thing — `get_editable_text_iface()` returns `None` when the element is not an editable text, and returns the `Atspi.EditableText` interface when it is.

The `attempt()` wrapper already handles `None` → `False` correctly:

```python
editable = attempt(node.get_editable_text_iface)
if editable is not None and ...
```

So the fix is a one-line change.

innocarpe:
## Summary

- Linux `computer set-value` always failed with `'Accessible' object has no attribute 'is_editable_text'` (#10569).
- `set_value` evaluated `node.is_editable_text` **before** `attempt()`, so the AttributeError from Python GI bindings escaped instead of falling through to Value.
- Probe editability with `get_editable_text_iface` (exposed by GI) and pin the regression in `verify-computer-native`.

## Root cause

```python
# before — attribute lookup raises before attempt() runs
attempt(node.is_editable_text, False)
```

`atspi_accessible_is_editable_text` exists in C but is not bound on `Atspi.Accessible` in `gir1.2-atspi-2.0`. `get_editable_text_iface()` is available and is the correct probe.

## Test plan

- [x] `python3 -c 'ast.parse(runtime.py)'` syntax-ok
- [x] guardrail string checks for the bad lookup pattern
- [ ] On Linux: `orca computer set-value` against an editable field after rebuild/repackage of computer-use-linux

Closes #10569

## ELI5

On Linux, computer-use set-value always failed on editable text because of a bad attribute probe. It now checks the correct editable-text interface so typing into fields works.

### DIFF
--- a/config/scripts/verify-computer-native.mjs
+++ b/config/scripts/verify-computer-native.mjs
@@ -185,6 +185,16 @@ function verifyNativeArgumentGuardrails() {
   )
   const windows = readFileSync(join(repoRoot, 'native/computer-use-windows/runtime.ps1'), 'utf8')
   const failures = []
+  // Why: GI bindings do not expose the C-only editable-text predicate; looking
+  // it up as an attribute raises before attempt() can catch it (#10569).
+  if (linux.includes('node.is_editable_text') || linux.includes('is_editable_text,')) {
+    failures.push(
+      'Linux set_value must not look up node.is_editable_text (use get_editable_text_iface)'
+    )
+  }
+  if (!linux.includes('get_editable_text_iface')) {
+    failures.push('Linux set_value must probe editability via get_editable_text_iface')
+  }
   if (linux.includes('count or 1') || linux.includes('pages or 1')) {
     failures.push('Linux provider must not coerce explicit zero action values to defaults')
   }
--- a/config/scripts/verify-computer-native.mjs
+++ b/config/scripts/verify-computer-native.mjs
@@ -187,17 +187,26 @@ function stripPythonComments(source) {
     const ch = source[i]
     const next = source[i + 1]
     // Triple-quoted strings (keep contents so string literals still scan).
+    // Why: honor escaped delimiters so `\"\"\"` inside a triple-string does not
+    // end the string early and reclassify later code as comments (#10586 CR).
     if ((ch === '"' || ch === "'") && next === ch && source[i + 2] === ch) {
       const quote = ch.repeat(3)
       out += quote
       i += 3
-      while (i < n && source.slice(i, i + 3) !== quote) {
+      while (i < n) {
+        if (source[i] === '\\' && i + 1 < n) {
+          out += source[i]
+          out += source[i + 1]
+          i += 2
+          continue
+        }
+        if (source.slice(i, i + 3) === quote) {
+          out += quote
+          i += 3
+          break
+        }
         out += source[i]
         i += 1
-      }
-      if (i < n) {
-        out += quote
-        i += 3
       }
       continue
     }
@@ -245,10 +254,9 @@ function verifyNativeArgumentGuardrails() {
   const failures = []
   // Why: GI bindings do not expose the C-only editable-text predicate; looking
   // it up as an attribute raises before attempt() can catch it (#10569).
-  if (linux.includes('node.is_editable_text') || linux.includes('is_editable_text,')) {
-    failures.push(
-      'Linux set_value must not look up node.is_editable_text (use get_editable_text_iface)'
-    )
+  // Match any attribute lookup form, not only `node.is_editable_text` (#10586 CR).
+  if (/(?:^|[^A-Za-z0-9_])is_editable_text(?:\s*[,(]|\s*$)/m.test(linux)) {
+    failures.push('Linux set_value must not look up is_editable_text (use get_editable_text_iface)')
   }
   if (!linux.includes('get_editable_text_iface')) {
     failures.push('Linux set_value must probe editability via get_editable_text_iface')
--- a/config/scripts/verify-computer-native.mjs
+++ b/config/scripts/verify-computer-native.mjs
@@ -177,8 +177,66 @@ function verifyWindowsProviderHandshake() {
   }
 }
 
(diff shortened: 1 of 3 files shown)


bbingz:
Cross-referencing: my #12603 lands on the same root cause as this PR — probing the EditableText iface instead of calling `node.is_editable_text` directly. This one is 10 days older and carries the same probe; #12603 adds an ATSPI-stubbed unit test covering the editable / value / no-iface paths. Happy to fold my test into this PR (or the probe into mine) — whichever direction a maintainer prefers, three parallel submissions on #10569 is the main thing keeping all of them stuck.

innocarpe:
@bbingz Thanks for the cross-reference — agreed that three open PRs on the same `#10569` surface is the main stall.

I'm happy either way:
- fold your ATSPI-stubbed unit tests from #12603 into this PR (#10586), **or**
- land the probe here / on #12603 with the stronger test suite and close the other as duplicate.

Whichever a maintainer prefers, I'll help consolidate so we only keep one merge path.

bbingz:
@innocarpe Taking the consolidate path toward **this PR (#10586)** as the single merge target (older, same iface probe).

Please cherry-pick the ATSPI-stubbed unit suite from my #12603 when you have a moment — it is additive and only covers the three set-value paths the probe is meant to distinguish:

- file: `native/computer-use-linux/set_value_test.py`
- tip: `a0d9c0547` on `bbingz:fix/10569-linux-set-value-editable-text`
- raw: https://raw.githubusercontent.com/bbingz/orca/a0d9c0547/native/computer-use-linux/set_value_test.py
- PR: #12603 (`runtime.py` + that test; +110)

Once that file (or an equivalent) is on this branch I will close #12603 as duplicate with credit to you for the original probe. If you prefer the inverse (land #12603, close this), say the word and I will flip the direction instead of leaving three PRs on #10569.

innocarpe:
@bbingz Folded your ATSPI-stubbed suite from #12603 onto this branch:

- `native/computer-use-linux/set_value_test.py` (credit: tip `a0d9c0547`)
- Local: 3/3 tests pass (`editable` / `value` fallback / no-iface)

Happy for you to close #12603 as duplicate when convenient — thanks for consolidating on this older probe path.

bbingz:
@innocarpe thanks — closed #12603 as duplicate. Single merge path is this PR; suite credit stays in the file header + your fold commit (`6a6c4125a`).

innocarpe:
Addressed latest CodeRabbit on the guardrail scanner:

1. `stripPythonComments` is escape-aware for triple-quoted delimiters
2. Forbidden probe match rejects any `is_editable_text` attribute lookup form, not only `node.is_editable_text`

Also rebased onto latest `main`. BingZ suite remains; #12603 is closed as duplicate.

tip: `e9718bbac1`

bbingz:
@innocarpe thanks for the rebase + guardrail scanner fixes (`e9718bbac1`). Glad the suite stayed; standing by on this single merge path.

PR #10589, opened 2026-07-25
feat(agents): opt out of reopening workspaces with creation agent

### LINKED ISSUE #10578: [Feature]: Let workspaces opt out of relaunching the agent they were created with
### Problem or use case

A worktree created with an agent selected (e.g. **Claude**) silently relaunches that agent every time the workspace is reopened from the sidebar — even after the user has deliberately exited the agent. There is no way to opt out.

**Repro**

1. `+` → **Create worktree** → pick a project, set **Agent** to `Claude`, create.
2. Work in the session, then exit Claude and close the tab (or let the workspace end up with no renderable tabs — sleep/restart also gets you there).
3. Click that workspace in the sidebar again.
4. **Actual:** Claude launches on its own, no user action.
5. Create another worktree with **Agent = Blank Terminal** and repeat: only a plain shell opens.

**Why this happens** (this is intentional today — [#1814](<https://github.com/stablyai/orca/pull/1814>) "remember worktree creation agent")

* Creation persists the choice as `Worktree.createdWithAgent`; picking *Blank Terminal* stores nothing, which is exactly why the blank case behaves differently.
* `activateAndRevealWorktree` falls back to a synthesized startup whenever the caller passes none:  <br>`src/renderer/src/lib/worktree-activation.ts:348` → `opts?.startup ?? buildCreatedAgentReopenStartup(wt)`
* `buildCreatedAgentReopenStartup` (`src/renderer/src/lib/worktree-activation.ts:231-279`) builds a launch plan from `createdWithAgent` with `prompt: ''` and `allowEmptyPromptLaunch: true`.
* It only fires when the workspace has no renderable tabs — `shouldAutoCreateInitialTerminal(renderableTabCount)` is `renderableTabCount === 0` (`src/renderer/src/components/terminal/initial-terminal.ts`).
* Locked in by `reopens an empty worktree with the agent selected at creation time` (`src/renderer/src/lib/worktree-activation-created-agent.test.ts:60`).

So the behavior is by design, but two things about it are questionable:

1. **No opt-out, and an explicit exit is not respected.** `createdWithAgent` records how the workspace was *created*, and is then treated as a standing instruction for every future reopen. Quitting the agent is a clear signal the user is done with it; the next sidebar click overrides that, spends tokens, and (for agents with an approval-bypass default arg) starts a permissive session unprompted. There is no setting to disable it.
2. **It is labeled** `resume` **but is not one.** The payload reports `telemetry.request_kind: 'resume'` with `launch_source: 'sidebar'`, yet it launches a brand-new empty-prompt session with no `resumeProviderSession` — prior conversation is not carried over. Real resume is handled separately by `resumeSleepingAgentSessionsForWorktree`. The mismatch makes the telemetry misleading and sets the wrong user expectation.

Note that the opposite expectation also exists in the wild (stablyai/orca#9772 asks for a *default* agent to open reliably on workspace click), so this is about making the behavior selectable rather than removing it.

### Proposed solution

Preferred: make the reopen relaunch a real setting instead of implicit behavior, e.g. `Reopen workspaces with the agent they were created with` (default on, to preserve today's behavior), surfaced in Settings → Agents next to the default-agent controls.

Additionally, either of these would fix the sharper part of the complaint:

* Treat a clean agent exit as consent withdrawal — clear or shadow `createdWithAgent` (or record a `lastAgentExitedCleanly` flag) so a deliberately-ended agent is not resurrected, while crash/sleep paths still relaunch.
* Fix the telemetry `request_kind` for this path so a fresh empty-prompt launch is not counted as a `resume`.

* Workaround today: create the worktree with **Blank Terminal** and launch the agent manually from the tab bar `+`, or leave a non-agent tab open so `renderableTabCount !== 0` suppresses the relaunch. Neither is discoverable.
* Platform-independent (renderer-side activation path); applies to git worktrees, folder workspaces, and SSH/remote projects alike.
* Related: [#6945](<https://github.com/stablyai/orca/pull/6945>) already went the "don't auto-open agent tabs the user didn't ask for" direction for sleeping-session resume; this path never got the same treatment.

Reported on macOS.

innocarpe:
## Summary

- Workspaces created with an agent (e.g. Claude) relaunch that agent whenever the empty workspace is reopened from the sidebar — even after a deliberate exit (#10578).
- Add **Settings → Agents → “Reopen workspaces with the agent they were created with”** (default **on**, preserving #1814).
- When off, empty reopen uses a blank terminal; `createdWithAgent` is still stored for history but not auto-launched.
- Telemetry for this path is `request_kind: 'new'` (not `'resume'`) because it starts a fresh empty-prompt session.

## Test plan

- [x] `vitest` — `worktree-activation-created-agent.test.ts` (13 tests, including opt-out)
- [ ] Create worktree with Claude → exit → reopen sidebar: agent relaunches (default)
- [ ] Toggle setting off → same flow opens blank terminal only
- [ ] Settings search finds “reopen” / “created” under Agents

Closes #10578

## ELI5

Reopening an empty workspace used to always relaunch the agent it was created with. A new setting lets you opt out so reopen starts a blank terminal while still remembering which agent created it.

### DIFF
--- a/src/renderer/src/components/settings/AgentsPane.tsx
+++ b/src/renderer/src/components/settings/AgentsPane.tsx
@@ -19,6 +19,7 @@ import { Button } from '../ui/button'
 import { Input } from '../ui/input'
 import { cn } from '@/lib/utils'
 import { AgentAwakeSetting } from './AgentAwakeSetting'
+import { CreatedAgentReopenSetting } from './CreatedAgentReopenSetting'
 import { AgentCacheTimerSection } from './AgentCacheTimerSection'
 import { AgentRuntimeSetting } from './AgentRuntimeSetting'
 import {
@@ -868,6 +869,8 @@ export function AgentsPane({
       <AgentGeneratedTabTitlesSetting settings={settings} updateSettings={updateSettings} />
 
       <AgentAwakeSetting settings={settings} updateSettings={updateSettings} />
+
+      <CreatedAgentReopenSetting settings={settings} updateSettings={updateSettings} />
 
       <AgentCacheTimerSection settings={settings} updateSettings={updateSettings} />
 
--- a/src/renderer/src/components/settings/CreatedAgentReopenSetting.tsx
+++ b/src/renderer/src/components/settings/CreatedAgentReopenSetting.tsx
@@ -1,11 +1,11 @@
 import type { GlobalSettings } from '../../../../shared/types'
-import { Label } from '../ui/label'
 import {
   getCreatedAgentReopenDescription,
   getCreatedAgentReopenSearchKeywords,
   getCreatedAgentReopenTitle
 } from './created-agent-reopen-copy'
 import { SearchableSetting } from './SearchableSetting'
+import { SettingsSwitchRow } from './SettingsFormControls'
 
 type CreatedAgentReopenSettingProps = {
   settings: GlobalSettings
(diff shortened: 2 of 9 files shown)


innocarpe:
### Sync update (`e9c4b4882`)

Rebase onto upstream/main to clear merge conflicts

- Commit: `e9c4b48820c1fbcf02ee8009f6152f8c494d29fb`
- Upstream: https://github.com/stablyai/orca/pull/10589
- Fork portfolio: https://github.com/innocarpe/orca/pull/78

innocarpe:
### Sync update (`59aafa54b`)

Address Greptile: setting label matches live UI title

- Commit: `59aafa54b3758619b5796092c413361ad645f32a`
- Upstream: https://github.com/stablyai/orca/pull/10589
- Fork portfolio: https://github.com/innocarpe/orca/pull/78

innocarpe:
Closing in favor of merged #10647, which already closes #10578.

Empty-workspace activation no longer synthesizes a startup from `createdWithAgent`. This branch would duplicate that policy. Thank you @nwparker / @AmethystLiang.

PR #10599, opened 2026-07-25
fix(worktrees): sweep PTYs when purging externally deleted worktrees

### LINKED ISSUE #10562: [Bug]: Externally deleted worktrees (git worktree remove outside Orca) strand their terminal/agent processes — reconcile paths run no teardown
**Description**

**Cause**

**Proposed fix**

**Acceptance tests**

**Environment**

innocarpe:
## Summary

- External `git worktree remove` caused Orca to drop UI maps on authoritative reconcile but left daemon/relay PTYs alive (#10562).
- On `purgeWorktreeTerminalState`, best-effort `worktrees:sweepOrphanProcesses` → `killAllProcessesForWorktree` (no `requirePhysicalStop`).
- Web client forwards stop to runtime; floating worktree id is skipped.

## Test plan

- [ ] Local: create worktree, start agent, `git worktree remove` outside Orca, wait for scan → process gone
- [ ] Disconnect SSH and fail scan → no sweep
- [ ] Explicit removeWorktree path unchanged

Closes #10562

## ELI5

If a worktree was deleted outside Orca, the UI dropped it but terminal processes could keep running. Purging that state now best-effort kills orphan PTYs so nothing leaks in the background.

### DIFF
--- a/src/main/ipc/worktrees.ts
+++ b/src/main/ipc/worktrees.ts
@@ -4,6 +4,7 @@ import { ipcMain } from 'electron'
 import { readFile, stat } from 'node:fs/promises'
 import { randomUUID } from 'node:crypto'
 import type { Store } from '../persistence'
+import { FLOATING_TERMINAL_WORKTREE_ID } from '../../shared/constants'
 import { isFolderRepo } from '../../shared/repo-kind'
 import { readBranchRenameFailureOutputForDisplay } from '../agent-hooks/branch-rename-failure-output'
 import {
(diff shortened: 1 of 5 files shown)


nwparker:
Thanks @innocarpe -- merged #11237 instead 
Hope that's alright!

innocarpe:
Totally fine @nwparker — glad the fix landed via #11237.

I'll keep trying to send useful, focused contributions. Orca is a product I genuinely like and use every day, so I'm happy if I can help even a little.

Thanks for the heads-up!

PR #10600, opened 2026-07-25
ci(pr): run E2E when a PR touches tests/e2e paths

### LINKED ISSUE #10518: [Bug]: E2E regression tests can merge red — PR CI never runs the E2E suite
## Summary

A PR that adds an E2E regression test never runs that test. `pr.yml` does not invoke the E2E suite — `e2e.yml` is reachable only via `schedule` (on `main`) and via `release-cut.yml`'s `workflow_call`. So a fix PR can add a regression test, go green, merge, and only then have its own test run for the first time — on `main`, where nobody is gating on it.

This has already happened at least twice, and both tests have **never passed on `main`**.

**Environment:** Linux (`ubuntu-latest`, `electron-headless` project), `main`.

## Case 1 — `headless-serve-desktop-activation.spec.ts` (11 days red)

Added 2026-07-13 by `1d2aaf1bf` ("Fix recipe serve desktop promotion", #8646). It landed after the last green scheduled run, so the first scheduled run that ever contained it was [29353814223](https://github.com/stablyai/orca/actions/runs/29353814223) (`e0edc8ef7`, 07-14) — where it failed at `:80`. It has failed in every scheduled run since, and still fails today at `:89` in [30130776139](https://github.com/stablyai/orca/actions/runs/30130776139).

```
TimeoutError: electronApplication.firstWindow: Timeout 60000ms exceeded while waiting for event "window"
```

## Case 2 — `markdown-explorer-find-focus.spec.ts`

Added 2026-07-23 23:18 by `65f245f07` ("fix(editor): focus Explorer-opened Markdown for find", #8083), which closes #8078. The first scheduled run containing it, [30113308910](https://github.com/stablyai/orca/actions/runs/30113308910) (`17959a22b`, 07-24), failed — as did every run since.

The failure is at the assertion that encodes the bug. The earlier steps all pass: the `README.md` row is visible, the click lands, and `.rich-markdown-editor` becomes visible. Only the find box never appears:

```
Locator: getByRole('textbox', { name: 'Find in rich markdown editor' })
Expected: visible
Error: element(s) not found

  > 23 |   ).toBeVisible()
    at tests/e2e/markdown-explorer-find-focus.spec.ts:23:5
```

That is the original #8078 symptom: open a Markdown file from Explorer, press Find, nothing happens. **#8078 should not be considered verified fixed** — either the fix does not hold under `electron-headless` on Linux, or the test is wrong. Nothing in CI has ever distinguished the two.

## Reproduction

I re-ran the affected shards twice, independently, on a fork at `1367094bb` (a strict subset of upstream `main`, no local commits). Both specs failed identically in both runs and match `main`'s own runs — these are not runner flake. A test in the same shards *did* move between the two runs (`terminal-tab-close-restart-persistence` → `terminal-wedged-write-pipeline-recovery`), which is what flake actually looks like here.

## Suggested fix

Give PRs some E2E signal, so a regression test cannot merge red:

- run the E2E suite (or a sharded subset) on PRs that touch `tests/e2e/**` or the relevant `src/**` paths, or
- at minimum run the specific spec files a PR adds or modifies.

Related: the standing `main` E2E failures are tracked separately in #10519.

innocarpe:
## Summary

- `pr.yml` never called `e2e.yml`, so a PR could add a red regression test and still go green (#10518).
- Path-filter on PR: if the diff touches `tests/e2e/**`, `playwright.*`, or `.github/workflows/e2e.yml`, `workflow_call` the existing E2E suite.
- Ordinary PRs stay light; E2E-only changes get full sharded signal.

## Test plan

- [ ] Open a PR that only touches `src/` → no e2e job
- [ ] Open a PR that touches `tests/e2e/**` → e2e workflow runs

Closes #10518

## ELI5

PRs that only touched E2E tests never ran the E2E workflow, so broken tests could look green. Path filters now trigger E2E when test/e2e paths change while leaving ordinary PRs light.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,5 +1,9 @@
 name: E2E
 
+# Why: checkout + artifact upload only; callers can only further restrict.
+permissions:
+  contents: read
+
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -165,3 +165,39 @@ jobs:
       # temp copy so Node cannot mask missing package deps with repo node_modules.
       - name: Smoke packaged CLI
         run: node config/scripts/smoke-packaged-cli.mjs --app-dir=dist/linux-unpacked
+
+  # Why: regression specs under tests/e2e/** used to merge green without ever
+  # running — e2e.yml only fired on schedule/release (#10518). Path-filter so
+  # ordinary PRs stay light; any E2E suite change still gets a full shard run.
+  e2e-paths:
+    name: detect e2e path changes
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.draft != true
+    outputs:
+      should_run: ${{ steps.filter.outputs.should_run }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Filter E2E-relevant paths
+        id: filter
+        run: |
+          set -euo pipefail
+          BASE="${{ github.event.pull_request.base.sha }}"
+          HEAD="${{ github.event.pull_request.head.sha }}"
+          if git diff --name-only "$BASE" "$HEAD" | grep -E '^(tests/e2e/|playwright\.|\.github/workflows/e2e\.yml$)' >/dev/null; then
+            echo "should_run=true" >> "$GITHUB_OUTPUT"
+            echo "E2E path changes detected"
+          else
+            echo "should_run=false" >> "$GITHUB_OUTPUT"
+            echo "No E2E path changes"
+          fi
+
+  e2e:
+    name: e2e
+    needs: e2e-paths
+    if: needs.e2e-paths.outputs.should_run == 'true'
+    uses: ./.github/workflows/e2e.yml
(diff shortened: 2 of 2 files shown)


innocarpe:
### Sync update (`2fd4985cb`)

Address CodeRabbit on E2E path detection: merge-base diff (no base-branch drift false triggers), fail when git diff cannot compute the range, least-privilege contents:read on detector + e2e reusable workflow.

- Commit: `2fd4985cb16146c9978685500baa91ec898901b1`
- Upstream: https://github.com/stablyai/orca/pull/10600
- Fork portfolio: https://github.com/innocarpe/orca/pull/81

nwparker:
Picked this up in #11131 — thanks @innocarpe, the diagnosis here was right and the path-filter approach is the correct shape.

Your two commits are preserved as the base of that branch, rebased onto current `main` (this one was 185 commits behind and `pr.yml` was restructured underneath it, so it no longer applied cleanly). I added a second commit fixing two things:

1. **`verify` didn't depend on `e2e`.** That job is the required check and lists its dependencies explicitly, so a failing shard still left it green — the gate reproduced the hole it was closing. Added `e2e` to `needs` and the result list, with `skipped` allowed since the job is path-filtered.
2. **`playwright.` matched no tracked file.** The config is `tests/playwright.config.ts`, beside `tests/e2e/` rather than inside it, so editing the runner config would have silently skipped E2E. Anchored at `tests/playwright.`.

Also added a contract test so neither can regress. Closing this in favor of #11131 — full credit retained via Co-authored-by.

PR #10603, opened 2026-07-25
fix(mobile): optional toggle to dim agent directory-listing size tokens

### LINKED ISSUE #10508: Mobile: optional toggle to de-emphasize file-size tokens in agent directory listings
### Summary
Add an opt-in Terminal setting on mobile that visually dims (or hides) the trailing file-size tokens (`1.4K`, `738B`, `45.9K`, …) that coding agents print in directory listings. On a phone-width terminal these sizes add a column of noise the user rarely needs, and they read like clutter next to the filenames.

### Where the numbers come from
This is **not** a rendering bug. The mobile terminal is a faithful, byte-for-byte mirror of the agent's PTY: raw bytes arrive as `TerminalStreamOpcode.Output` (`mobile/src/transport/terminal-stream-protocol.ts`) and replay into xterm.js inside a WebView (`mobile/src/terminal/TerminalWebView.tsx`). The size tokens are part of the agent's own output (an `ls`/`Read`-directory result), not something Orca generates — confirmed by the fact that the app's only size formatter, `mobile/src/files/mobile-file-preview-request.ts`, prints `"1.4 MB"` (spaced, two-letter units), whereas the on-screen tokens are the agent's spaceless `1.4K`/`738B` form.

### Proposed behavior
- A new toggle in the Terminal settings screen (`mobile/app/terminal-settings.tsx`), off by default, persisted like the existing autocomplete toggle (`orca:...` key in `mobile/src/storage/preferences.ts`).
- When on, the rendered terminal DOM visually dims or hides size tokens that trail a filename on a listing line.
- **Visual-only.** It must not alter the xterm buffer — selection, copy, and reflow keep the original bytes, exactly like the path-tap and OSC-link features already do.

### Why visual-only, and precedent
Orca already scans rendered terminal content without corrupting the buffer: injected WebView scripts detect file paths (`mobile/src/terminal/terminal-path-tap-injected.ts`), URLs, and OSC links. This toggle should ride the same injected-script path — match `<name> <size>` at end-of-line and apply a dim/hidden style — so the mirror guarantee holds and copy still yields the true text.

### Tradeoffs / open questions
- Heuristic parsing: the size regex will occasionally match or miss (e.g. a literal `4K` in prose, or unusual unit formats). Being visual-only keeps the blast radius cosmetic.
- "Dim" vs. "hide": dim is safer (still readable, no layout shift); a full hide risks confusing gaps. Recommend **dim** as the default treatment.
- Should it also cover raw `ls -lh` output, or only the agent's `name  size` listing format? Suggest starting narrow (trailing-size-token on listing lines) and widening later.

### Out of scope
Changing the agent's actual output format (that's upstream, e.g. the coding agent's prompt/tooling) and any change to the desktop terminal.

### Anchors
`mobile/src/terminal/TerminalWebView.tsx` · `mobile/src/terminal/terminal-path-tap-injected.ts` · `mobile/app/terminal-settings.tsx` · `mobile/src/storage/preferences.ts` · `mobile/src/transport/terminal-stream-protocol.ts`

innocarpe:
## Summary
- Adds an opt-in Terminal setting on mobile that **visually dims** trailing file-size tokens (`1.4K`, `738B`, `45.9K`, …) in agent directory listings.
- Visual-only xterm decorations — selection, copy, and the PTY buffer stay byte-identical.
- Off by default; persists per device like the autocomplete toggle.

## Fixes
Closes #10508

## Test plan
- [x] Unit tests: preference load/save + subscriber; size-token matcher
- [ ] Manual: enable Settings → Terminal → Dim file sizes; open agent listing; sizes look dimmed; copy still includes sizes

## ELI5

Mobile terminals can optionally dim trailing file-size tokens in directory listings so agent ls output is easier to read. It's visual-only and off by default.

### DIFF
--- a/mobile/app/terminal-settings.tsx
+++ b/mobile/app/terminal-settings.tsx
@@ -19,8 +19,10 @@ import { TerminalShortcutSettings } from '../src/components/TerminalShortcutSett
 import { setTerminalAutoRestoreFitMsForHost } from '../src/terminal/terminal-auto-restore-fit-state'
 import {
   loadTerminalAutocompleteEnabled,
+  loadTerminalDimDirListingSizes,
   loadTerminalTextScale,
   saveTerminalAutocompleteEnabled,
+  saveTerminalDimDirListingSizes,
   saveTerminalTextScale
 } from '../src/storage/preferences'
 
@@ -174,6 +176,25 @@ export default function TerminalSettingsScreen() {
     userToggledAutocompleteRef.current = true
     setAutocompleteEnabled(next)
     void saveTerminalAutocompleteEnabled(next)
+  }, [])
+
+  const [dimDirListingSizes, setDimDirListingSizes] = useState(false)
+  const userToggledDimSizesRef = useRef(false)
+  useEffect(() => {
+    let stale = false
+    void loadTerminalDimDirListingSizes().then((enabled) => {
+      if (!stale && !userToggledDimSizesRef.current) {
+        setDimDirListingSizes(enabled)
+      }
+    })
+    return () => {
+      stale = true
+    }
+  }, [])
+  const toggleDimDirListingSizes = useCallback((next: boolean) => {
+    userToggledDimSizesRef.current = true
+    setDimDirListingSizes(next)
+    void saveTerminalDimDirListingSizes(next)
   }, [])
 
   useEffect(() => {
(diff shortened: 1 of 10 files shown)


innocarpe:
### Sync update (`23a839b96`)

Address CodeRabbit: dispose IMarkers with decorations, tint size-dim overlays from terminalTheme.background (+ refresh on set-theme), subscribe before initial load so a live toggle cannot be overridden by a stale AsyncStorage read.

- Commit: `23a839b96ead72eda44c6e85eb1e29d605d2a2e0`
- Upstream: https://github.com/stablyai/orca/pull/10603
- Fork portfolio: https://github.com/innocarpe/orca/pull/82

PR #10623, opened 2026-07-25
fix(ci): mirror missing lint steps in PR workflow (#10601)

### LINKED ISSUE #10601: CI omits 3 of the 9 `lint` steps — locale catalog drift ships undetected
## Summary

`.github/workflows/pr.yml` does not run `pnpm run lint`. It re-enumerates the lint sub-steps as individual workflow steps, and three of them were never mirrored. The most consequential omission is `verify:localization-catalog`: **a PR that adds a key to `en.json` without updating the other four catalogs passes CI green.**

I hit this while adding a sixth locale (#10592). A merge produced a catalog missing one key; `pnpm run lint` caught it locally, and only then did I notice CI would not have.

## Evidence

The `lint` script (`package.json`) is a 9-step chain:

```
oxlint
&& lint:switch-exhaustiveness
&& node config/scripts/check-styled-scrollbars.mjs
&& check:reliability-gates
&& check:max-lines-ratchet
&& verify:bundled-skill-guides
&& verify:skill-bundle-manifest
&& verify:localization-catalog
&& verify:localization-coverage
```

`pr.yml` runs these individually (lines 69–91, 111–117):

```
pnpm exec oxlint --format github
pnpm check:styled-scrollbars
pnpm run check:reliability-gates
pnpm run check:max-lines-ratchet
pnpm run verify:bundled-skill-guides
pnpm run verify:skill-bundle-manifest
pnpm check:feature-wall-assets
pnpm verify:macos-entitlements
pnpm typecheck
```

Missing, confirmed by grep across all of `.github/workflows/`:

| Step | In `lint` | In CI |
| --- | --- | --- |
| `lint:switch-exhaustiveness` | yes | **no** |
| `verify:localization-catalog` | yes | **no** |
| `verify:localization-coverage` | yes | **no** |

```
$ grep -rn "switch-exhaustiveness\|localization" .github/workflows/
(no matches)

$ grep -rn "run lint\|pnpm lint" .github/workflows/
.github/workflows/mobile.yml:68:        run: pnpm lint      # scoped to mobile/ only
```

No local safety net either — `.husky/pre-commit` is just `pnpm exec lint-staged`, which runs `oxlint` / `oxlint --config config/oxlint-react-doctor.json` / `oxfmt --write` on staged files. None of the three missing verifiers run there.

## Impact

- **Locale parity is unenforced.** The five catalogs are supposed to stay key-for-key identical (`verify:localization-catalog` checks exactly this). Nothing in CI enforces it, so a catalog can silently fall behind `en.json` and the affected UI renders the English fallback in a localized build.
- **`verify:localization-coverage`** — same, unenforced.
- **`lint:switch-exhaustiveness`** — unenforced; a non-exhaustive switch over a union can reach `main`.

The desktop `lint` chain is what contributors are told to run (`.github/CONTRIBUTING.md:41`, `.github/pull_request_template.md:12`), so the gap only bites when someone relies on CI rather than running it locally.

## Root cause, and why patching it once isn't enough

The failure mode is structural: `pr.yml` duplicates the `lint` chain by hand, so every new lint step has to be manually mirrored into the workflow or it silently never runs. `check:styled-scrollbars`, `check:feature-wall-assets` and `verify:macos-entitlements` show the list has been maintained by hand over time — three steps just drifted out of sync.

## Suggested fix

Either:

1. **Replace the enumerated steps with a single `pnpm run lint`.** Simplest, and structurally prevents recurrence. Downside: one failing step masks the others' output, and `oxlint --format github` (which produces inline PR annotations) would need to stay separate or be folded into the script.
2. **Add the three missing steps** and add a guard that fails if `lint`'s chain contains a script not present in `pr.yml` — otherwise the same drift recurs.

Option 1 with `oxlint --format github` kept as its own preceding step gets both the annotations and the guarantee.

## Environment

- Observed on `main` @ `1367094bb`; also present on current `main`.
- Orca 1.4.155, macOS arm64.

hsusul:
### Linked Issue
Fixes #10601

### Observed Failure
`.github/workflows/pr.yml` manually re-enumerates individual sub-steps of `package.json`'s `lint` script instead of calling `pnpm run lint`. Over time, three lint verifiers were omitted from the CI pipeline:
- `lint:switch-exhaustiveness`
- `verify:localization-catalog`
- `verify:localization-coverage`

As a result, locale catalog key drift and non-exhaustive switch statements could ship without being caught by CI.

### Root Cause
`.github/workflows/pr.yml` listed individual lint sub-steps by hand, and these three steps were never mirrored into the workflow.

### Implementation
Updated `.github/workflows/pr.yml` to include the three missing lint sub-steps:
- `Check switch exhaustiveness` (`pnpm run lint:switch-exhaustiveness`)
- `Verify localization catalog` (`pnpm run verify:localization-catalog`)
- `Verify localization coverage` (`pnpm run verify:localization-coverage`)

### Regression Coverage & Validation
- Ran `pnpm run lint` locally -> all 9 lint steps passed cleanly (including 0 errors across 8712 files for switch exhaustiveness, and full parity verification across all 5 localization catalogs: `en`, `es`, `ja`, `ko`, `zh`).
- Verified `git diff --check` -> Clean with no trailing whitespace.

### Limitations or Untested Platforms
None.

### Unrelated Changes
No unrelated changes were included.

## ELI5

PR CI skipped some real lint checks (switch exhaustiveness and localization verifiers) that local `pnpm lint` runs. Those steps are mirrored in the PR workflow so catalog drift and incomplete switches can't slip through.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -37,6 +37,9 @@ jobs:
       - name: Check styled scrollbars
         run: pnpm check:styled-scrollbars
 
+      - name: Check quadratic buffer concatenation
+        run: pnpm run check:quadratic-buffer-concat
+
       - name: Check reliability gate manifest
         run: pnpm run check:reliability-gates
 
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -31,6 +31,9 @@ jobs:
       - name: Lint
         run: pnpm exec oxlint --format github
 
+      - name: Check switch exhaustiveness
+        run: pnpm run lint:switch-exhaustiveness
+
       - name: Check styled scrollbars
         run: pnpm check:styled-scrollbars
 
@@ -46,6 +49,17 @@ jobs:
       - name: Verify skill freshness manifest
         run: pnpm run verify:skill-bundle-manifest
 
+      - name: Verify localization catalog
+        run: pnpm run verify:localization-catalog
+
+      - name: Verify localization coverage
+        run: pnpm run verify:localization-coverage
+
+      # Why: project-owned type declarations must live in .ts so tsc
+      # actually checks them. TypeScript's skipLibCheck: true (inherited
+      # from @electron-toolkit/tsconfig) silently widens unresolved names
+      # in .d.ts to `any`, which is how #1186 shipped a broken IPC signature
+      # past typecheck. See docs/preload-typecheck-hole.md.
       - name: Guard against project-owned .d.ts in preload/shared
         run: |
           matches=$(find src/preload src/shared -name '*.d.ts' 2>/dev/null || true)
--- a/config/localization-coverage-allowlist.json
+++ b/config/localization-coverage-allowlist.json
@@ -40,5 +40,33 @@
     "text": "Idioma",
     "dynamic": false,
     "count": 1
+  },
+  {
+    "filePath": "src/renderer/src/components/settings/terminal-advanced-platform-search.ts",
+    "kind": "object-property:keywords",
+    "text": "Ghostty",
+    "dynamic": false,
+    "count": 2
+  },
+  {
+    "filePath": "src/renderer/src/components/settings/terminal-advanced-platform-search.ts",
+    "kind": "object-property:keywords",
+    "text": "ghostty",
+    "dynamic": false,
+    "count": 2
+  },
+  {
+    "filePath": "src/renderer/src/components/settings/terminal-pane-appearance-search.ts",
+    "kind": "object-property:keywords",
+    "text": "Ghostty",
+    "dynamic": false,
+    "count": 1
+  },
+  {
+    "filePath": "src/renderer/src/components/settings/terminal-pane-appearance-search.ts",
+    "kind": "object-property:keywords",
+    "text": "ghostty",
+    "dynamic": false,
+    "count": 1
   }
 ]
--- a/config/scripts/pr-workflow-lint-parity.test.mjs
+++ b/config/scripts/pr-workflow-lint-parity.test.mjs
@@ -64,8 +64,11 @@ describe('PR workflow lint parity', () => {
     const { scripts } = JSON.parse(readFileSync('package.json', 'utf8'))
     const workflow = parse(readFileSync('.github/workflows/pr.yml', 'utf8'))
 
+    // Scan every job: which one hosts the lint steps is an organizational
+    // detail that has already been renamed once (verify -> static_analysis).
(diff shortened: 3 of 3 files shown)


OrcaWin:
Reviewed with an independent reproduction. **The bug is real and this should land — but not as currently written, because it would red every subsequent PR.**

**Verified.** All three steps are genuinely absent from `pr.yml` on main, and I proved each catches something real rather than assuming it: deleting a key from `es.json` made `verify:localization-catalog` exit 1; a scratch non-exhaustive switch made `lint:switch-exhaustiveness` exit 1 **while the step CI actually runs today** (`oxlint --format github`) exited 0 on the identical file. That contrast is the proof CI would have missed it. Both scratch changes reverted.

**Blocker 1 — the newly-enabled step fails on current main.** `verify:localization-coverage` exits 1 today, on 6 unallowlisted `Ghostty` product-name keywords from #10628. Enabling the step as authored turns every subsequent PR red until that is allowlisted. Your "all 9 steps passed cleanly" was true against your base, but #10628 landed shortly after you opened this.

**Blocker 2 — the hand-mirrored list already re-drifted.** `check:quadratic-buffer-concat` joined the lint chain in #10851 the day after you opened this PR, and is in neither `pr.yml` nor this diff. The approach demonstrably failed within 24 hours, which is precisely what #10601 predicted: it asked for "a guard that fails if `lint`'s chain contains a script not present in `pr.yml` — otherwise the same drift recurs."

**What I have ready.** Locally I added the allowlist entry, the missing 4th step, and `config/scripts/pr-workflow-lint-parity.test.mjs` — the recurrence guard #10601 asked for. It resolves the `lint` chain through `pnpm run` indirection to leaf commands, canonicalizes reporting-only flags so `oxlint --format github` still matches bare `oxlint`, and fails naming any lint leaf with no `pr.yml` counterpart. Proven in both directions: it independently rediscovered all four missing steps against pristine main, fails naming the exact step when one is removed, and does not false-positive on invocation-form differences. It also had to be taught to scan **all** jobs rather than a hardcoded `verify` — the lint steps have since moved to `static_analysis`, which is the same drift in another guise. With all of it applied, `pnpm run lint` exits 0 on latest main.

**Why I have not pushed it:** GitHub refuses pushes touching `.github/workflows/` without the `workflow` OAuth scope, which my token lacks. So this needs either you applying the changes, or a maintainer with that scope.

I kept the enumerated-step shape rather than collapsing to `pnpm run lint`, since collapsing would mask later steps' output and lose oxlint's inline PR annotations. Good find — this closes a real hole in CI.

PR #10626, opened 2026-07-25
diag(crash): make renderer OOM profiles actionable

AmethystLiang:
## Summary

This PR makes the next renderer out-of-memory report identify which terminal subsystem is growing.

It does **not** claim to fix the underlying leak. It replaces stale/coarse memory sampling with precise trusted-renderer measurements, records four staged high-water profiles, and adds production diagnostics for terminal membership, xterm buffer weight, renderer output backlog, and pending PTY side effects.

### Why the previous data was misleading

Electron 43.1.0 embeds Chromium 150. In that build, Chromium's `MemoryInfo::GetHeapSize` reports V8 `used_heap_size()` plus `external_memory()`. Therefore `performance.memory.usedJSHeapSize` is not a pure JavaScript-object heap measurement. xterm's `Uint32Array` row backing stores contribute through V8 external memory and remain a plausible source of the observed growth.

Without `--enable-precise-memory-info`, Chromium also quantizes `performance.memory` into geometric buckets and caches it for about 20 minutes. That means the field traces could establish sustained growth, but could not distinguish a boot burst, smooth leak, short GC sawtooth, or exact climb rate within each cache window.

Aggregate review of 57 crash bundles found:

- 34 reached the investigation's high-memory band.
- High-memory reports appeared on all supported desktop platforms.
- Recorded Git activity was neither necessary nor sufficient.
- Browser guest count alone did not correlate.
- Repeated post-reload growth proves an external or reconstructible driver, but not automatic restoration of the prior mounted-worktree set: the renderer-local mount set resets and initially contains only the active worktree.

The leading hypothesis remains retained, streaming xterm buffers in mounted worktrees, but it is provisional. Other candidates are the uncapped PTY side-effect queue, renderer output/backpressure, and large passive editor/diff text working sets.

### What changed

- Enable precise `performance.memory` readings before trusted Orca renderer processes start.
- Preserve Chromium's allocation/GC timing mitigation for arbitrary sites by disabling Blink `PreciseMemoryInfo` in every browser guest path:
  - desktop `<webview>` contents, including attach-time hardening;
  - offscreen browser windows;
  - guest-created popup contents.
- Expand renderer high-water profiles from 60%/80% to 40%/60%/75%/85%.
- Retain all four thresholds for both trusted renderer surfaces and clear retained profiles after a renderer process dies, so a replacement renderer cannot inherit the dead process's profile.
- Attribute process-gone breadcrumbs to the renderer surface that produced them.
- Replace `created - disposed = live` arithmetic with truthful terminal metrics:
  - cumulative `registered`, `removed`, and `disposeErrors`;
  - current pane-manager `tracked` membership;
  - connected/disconnected tracked DOM containers;
  - xterm `bufferRows`, `scrollbackRows`, and `estimatedBufferCells`.
- Add production output-queue metrics for current/peak queued terminals and characters plus lossy backlog drops.
- Add `ptySideEffects.processors` and `ptySideEffects.pending` gauges, disposed with their owning transports.
- Raise retained breadcrumb capacity from four to eight and cover interleaved main/popout thresholds.

### How to read the next profile

- Rising `paneTerminals.tracked` and `estimatedBufferCells` with memory supports mounted-terminal retention.
- Stable terminal count with rising rows/cells supports an active-tab or tab-local variant.
- Rising `ptySideEffects.pending` with flat buffer weight supports the side-effect queue hypothesis.
- Rising output backlog/peaks supports renderer delivery or parsing backpressure.
- Flat terminal and queue metrics would push investigation toward large strings or another subsystem.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 3,411 files passed; 36,229 tests passed; 62 skipped
- [ ] `pnpm build` — not run; typecheck and the complete unit/integration suite passed
- [x] Added regression coverage for precise-memory startup, all untrusted guest paths, four-threshold retention across both renderer surfaces, process-generation cleanup, terminal lifecycle semantics, pane-manager buffer aggregation, output-queue current/peak gauges, and PTY side-effect gauge disposal.

## AI Review Report

The review checked the measurement model, reload lifecycle, browser security boundary, terminal census semantics, breadcrumb retention, public-data disclosure, formatting, and comments.

It flagged and this PR addresses:

- `usedJSHeapSize` was incorrectly described as excluding typed-array backing stores.
- H1 and same-mounted-set reload claims were overstated.
- the process-global precise-memory switch exposed arbitrary browser contents;
- arithmetic terminal “liveness” did not represent reachability;
- creation/disposal counters could misstate construction failures and caught disposal errors;
- retained breadcrumb capacity was too small for four thresholds across two surfaces;
- the investigation document contained field identifiers and was moved to ignored local `design-docs/` rather than included in the public diff.

Cross-platform review covered macOS, Linux, and Windows. This PR changes no shortcuts, labels, shell commands, or path handling. Electron-specific behavior is shared across platforms and tested at each guest-creation seam. The diagnostics read existing renderer state and apply equally to local, SSH, remote-runtime, git-worktree, and folder workspaces.

## Security Audit

The main security risk was Chromium's process-global `enable-precise-memory-info` switch, which can expose allocation and GC timing to untrusted pages. The shared guest policy explicitly disables Blink `PreciseMemoryInfo` for webviews, offscreen windows, and popup contents; popup descendants inherit the same hardened creation policy. Trusted Orca UI surfaces keep precise diagnostics.

The new crash fields contain bounded numeric counts only—no terminal output, paths, URLs, profile names, tokens, or buffer contents. Contributor counts remain subject to the existing crash-profile budgets. No dependency, authentication, command-execution, or new IPC surface is introduced.

## Notes

This is diagnosis instrumentation, not the final leak fix. The next precise field cycle should determine whether to implement a memory-weighted mounted-terminal budget, cap/collapse pending side effects, backport remote delivery-credit behavior, or profile large retained strings.


AmethystLiang:
@coderabbitai[bot] addressed

AmethystLiang:
@coderabbitai[bot] addressed

AmethystLiang:
@coderabbitai[bot] addressed

AmethystLiang:
too large. used a smaller PR https://github.com/stablyai/orca/pull/12795

PR #10635, opened 2026-07-26
fix(skills): show unrecognized/inaccessible placements in update dialog

### LINKED ISSUE #10633: [Bug]: "Needs attention" skill pill has no explanation — its Details dialog reports "All installed Orca skills are up to date"
### Operating system

macOS

### Orca version

v1.4.156

### Details

**Short summary:** The amber **"Needs attention"** pill on a skill has no reachable explanation — the **Details** dialog it points to reports the green all-clear *"All installed Orca skills are up to date"* with no location rows. The badge can never be cleared or even understood.

#### What happened?

Settings → **Orchestration** and → **Computer Use** both show an amber `Needs attention` pill (sidebar + skill card). Clicking `Details` opens the *Update skills* dialog, which shows the all-clear headline and lists nothing. There is no way to learn which copy is at fault, and nothing to act on.

It survives everything a user can reasonably try:

| attempt | result |
| -- | -- |
| `npx skills update orchestration computer-use orca-cli --global` | `✓ All global skills are up to date` — pill unchanged |
| in-app `Update` button → dialog | *"All installed Orca skills are up to date"* — pill unchanged |
| remove `~/.claude/skills/<skill>` + reinstall via `npx skills add stablyai/orca --global --agent claude-code` | reinstalls as `(copied)` — pill unchanged |
| replace the agent-home copy with a symlink to the canonical copy | pill unchanged |
| full app quit + relaunch (fresh scan) | pill unchanged |

The canonical copies are byte-identical to what this build ships. Compared against `/Applications/Orca.app/Contents/Resources/skills/current-manifest.json`:

| skill | bundled revision | installed `gitTreeSha` | `SKILL.md` sha256 |
| -- | -- | -- | -- |
| orchestration | rev 28 · `9aa26fde…` | `9aa26fde…` ✅ | match ✅ |
| computer-use | rev 8 · `2072384f…` | `2072384f…` ✅ | match ✅ |
| orca-cli | rev 35 · `97f6596b…` | `97f6596b…` ✅ | match ✅ |

#### Root cause

The pill and the dialog disagree about what counts as a problem.

* **Pill** — `getSkillFreshnessDisplayStatus` (`src/renderer/src/lib/skill-freshness-display-status.ts`) returns `needs-attention` for **any** placement with `status !== 'current'`. That set includes `unrecognized`, `inaccessible` and `newer-known`.
* **Dialog** — `groupSkillFreshness` (`src/renderer/src/components/skills/skill-freshness-grouping.ts`) drops any skill that has no placement whose status is exactly `outdated`:

  ```ts
  if (!entries.some((entry) => entry.status === 'outdated')) {
    continue
  }
  ```

  No group → `hasBlockedGroup === false` → `summarizeInventory()` returns `'current'` → the all-clear headline.

So a placement that is `unrecognized` or `inaccessible` **but not** `outdated` lights the pill and is then filtered out of the very dialog meant to explain it. This is the failure mode stablyai/orca#10407 / stablyai/orca#10436 set out to remove — a badge no offered path can clear — relocated from the button to the dialog. The comment in `summarizeInventory` even names the conflict: *"the group filter already dropped the up-to-date and unrecognized-only noise"* — the dialog treats `unrecognized`-only as noise while the pill escalates it to amber.

#### How can we reproduce it?

Any user who has both Orca skills and Codex installed hits this permanently. Codex ships its own bundled skill that is also named `computer-use`:

```
~/.codex/plugins/cache/openai-bundled/computer-use/<version>/skills/computer-use/SKILL.md
~/.codex/.tmp/bundled-marketplaces/openai-bundled/plugins/computer-use/skills/computer-use/SKILL.md
```

Different content, different skill, `plugin-cache` topology → classified `unrecognized`, never `outdated`. So:

1. Install the Orca skills globally (`npx skills add stablyai/orca --global`).
2. Have Codex installed with its bundled `computer-use` plugin present.
3. Open Settings → Computer Use → amber `Needs attention`.
4. Click `Details` → *"All installed Orca skills are up to date"*, no rows.

Note the offending copy **must not be deleted** — it belongs to Codex, so "clean it up" is not available to the user either.

#### Expected

Either:

* the dialog lists blocked placements that are not `outdated`, so the pill has an explanation — the chips already exist (`unrecognized`, `inaccessible`, `plugin-cache`, `duplicate`, `external-link`, `read-only`, `in-a-repo`); or
* the pill stops escalating states the dialog deliberately treats as noise, so an unactionable amber badge is never shown.

#### Anything else that might help

A foreign skill that merely shares a name with an official one seems worth distinguishing from a *modified* official copy — the current `unrecognized` classification conflates "you edited ours" with "someone else's skill happens to be called the same thing", and only the first is the user's to fix.

innocarpe:
## Summary
- Update skills **Details** dialog now lists `unrecognized` and `inaccessible` placements, not only `outdated` ones.
- Fixes amber **Needs attention** pills that opened an all-clear dialog with no rows (e.g. Codex bundled `computer-use` sharing a name with Orca’s skill).

## Fixes
Closes #10633

## Test plan
- [x] `skill-freshness-grouping` + `SkillFreshnessUpdateDialog` + display-status tests
- [ ] Manual: Settings → Computer Use with Codex installed → Details lists the foreign copy chip

## ELI5

Skills marked "Needs attention" sometimes opened an empty Details dialog because unrecognized or inaccessible placements were hidden. Those placements are listed so you can see why attention is needed.

### DIFF
--- a/src/renderer/src/components/skills/skill-freshness-grouping.test.ts
+++ b/src/renderer/src/components/skills/skill-freshness-grouping.test.ts
@@ -39,16 +39,41 @@ describe('groupSkillFreshness', () => {
     ])
   })
 
-  it('hides skills with nothing out of date (current, unrecognized-only, unreadable-only)', () => {
+  it('hides skills that are fully current with nothing to review', () => {
+    const groups = groupSkillFreshness([placement('orca-cli', { status: 'current' })], [])
+    expect(groups).toEqual([])
+  })
+
+  it('lists unrecognized-only and inaccessible skills so the amber pill has an explanation', () => {
+    // Why: foreign plugin-cache copies (e.g. Codex bundled computer-use) share a
+    // name with Orca skills and light needs-attention; Details must not claim all-clear.
     const groups = groupSkillFreshness(
       [
-        placement('orca-cli', { status: 'current' }),
-        placement('dataviz', { status: 'unrecognized', topology: 'independent-copy' }),
+        placement('computer-use', { status: 'current' }),
+        placement('computer-use', {
+          rootId: 'plugin-cache',
+          unresolvedPath: '/home/.codex/plugins/cache/computer-use/SKILL.md',
+          status: 'unrecognized',
+          topology: 'plugin-cache'
+        }),
         placement('linear-tickets', { status: 'inaccessible' })
       ],
       []
     )
-    expect(groups).toEqual([])
+    expect(groups.map((group) => group.name)).toEqual(['computer-use', 'linear-tickets'])
+    expect(groups[0]?.status).toBe('cannot-update')
+    expect(groups[0]?.locations).toEqual([
+      { id: expect.any(String), path: '/home/.agents/skills/computer-use', chip: 'current' },
+      {
+        id: expect.any(String),
+        path: '/home/.codex/plugins/cache/computer-use/SKILL.md',
+        // Why: status (unrecognized) wins over topology (plugin-cache) in locationChip.
+        chip: 'unrecognized'
+      }
+    ])
+    expect(groups[1]?.locations).toEqual([
+      { id: expect.any(String), path: '/home/.agents/skills/linear-tickets', chip: 'inaccessible' }
+    ])
   })
 
   it('groups a blocked skill and flags the culprit location, not the main copy', () => {
--- a/src/renderer/src/components/skills/skill-freshness-grouping.test.ts
+++ b/src/renderer/src/components/skills/skill-freshness-grouping.test.ts
@@ -62,13 +62,14 @@ describe('groupSkillFreshness', () => {
     )
     expect(groups.map((group) => group.name)).toEqual(['computer-use', 'linear-tickets'])
     expect(groups[0]?.status).toBe('cannot-update')
+    expect(groups[1]?.status).toBe('cannot-update')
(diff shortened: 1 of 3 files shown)


innocarpe:
Status ping: narrow skills update dialog fix for unrecognized/inaccessible placements (tests). Ready for review when convenient.

brennanb2025:
Closing as superseded by #11128, which reached the same outcome through the shared predicate.

`groupSkillFreshness` on main now reads:

```ts
if (
  !pinned.has(name) &&
  !entries.some((entry) => entry.status === 'outdated' || isSkillCopyNeedingAttention(entry))
) {
  continue
}
```

Same goal as your `placementNeedsDialogAttention`, but reusing the predicate the badge already uses instead of introducing a second local one — which matters here, since a local copy is exactly how the badge and the dialog drifted apart in the first place.

One behavioral difference worth recording: this branch treats every `unrecognized` placement as dialog-worthy, including `unrecognized` + `plugin-cache`. That is the foreign Codex-bundled `computer-use` copy, which main deliberately exempts from **both** the badge and the dialog. Landing it now would produce a Details row for a skill whose badge is green — #10633's contradiction, mirrored rather than removed.

#10633 itself was closed by #10865. Thanks for the fix and for the clear write-up of the badge/dialog split.

PR #10638, opened 2026-07-26
fix(win): escalate stubborn agent PTY tree-kills on worktree delete

### LINKED ISSUE #10475: [Bug]: Right-click delete fails for Hermes worktrees — PTY process does not respond to SIGTERM
### Operating system

Windows

### Orca version

_No response_

### Details

Environment:

OS: Windows 11
Hermes Agent
Steps to reproduce:

Create a new worktree in Orca with Hermes as the agent (automatically opens a Hermes TUI tab)
Hermes TUI is idle (showing the TUI interface, no active API calls)
Right-click the worktree → Delete → Error: Failed to physically stop every PTY for worktree

Image

Expected behavior: Right-click delete should terminate the Hermes TUI PTY process and remove the worktree.

innocarpe:
## Summary
- Nested agent TUIs (Hermes, etc.) can survive the first ConPTY kill / single taskkill, so worktree delete fails with “Failed to physically stop every PTY”.
- Retry tree-kill when the root is still alive; re-escalate mid-wait during immediate physical stop.

## Fixes
Closes #10475

## Test plan
- [x] windows-process-tree-kill / local-pty-provider / session force-kill escalation tests
- [ ] Manual Windows: idle Hermes worktree → right-click Delete succeeds

## ELI5

On Windows, nested agent processes sometimes survived the first kill, blocking worktree delete. Tree-kill is retried and escalated so stubborn PTYs actually die and delete can finish.

### DIFF
--- a/src/main/daemon/session-force-kill-escalation.test.ts
+++ b/src/main/daemon/session-force-kill-escalation.test.ts
@@ -67,4 +67,49 @@ describe('Session force-kill escalation (#10475)', () => {
     await shutdown
     expect(session.state).toBe('exited')
   })
+
+  it('on Windows mid-wait escalate is tree-kill-only (no second ConPTY forceKill)', async () => {
+    vi.useFakeTimers()
+    try {
+      Object.defineProperty(process, 'platform', { configurable: true, value: 'win32' })
+      const subprocess: SubprocessHandle = {
+        pid: 5150,
+        getForegroundProcess: () => null,
+        write: vi.fn(),
+        resize: vi.fn(),
+        kill: vi.fn(),
+        forceKill: (...args: unknown[]) => forceKill(...args),
+        signal: vi.fn(),
+        onData: vi.fn(),
+        onExit: (cb) => {
+          onExitCb = cb
+        },
+        dispose: vi.fn()
+      }
+      session = new Session({
+        sessionId: 'hermes-mid-wait',
+        cols: 80,
+        rows: 24,
+        launchAgent: 'hermes',
+        subprocess,
+        shellReadySupported: false
+      })
+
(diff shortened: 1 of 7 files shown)


innocarpe:
### Sync update (`f965a79fc`)

Address CodeRabbit: tree-kill-only mid-wait on Windows; floor post-escalate waits; join shutdown if tree-kill throws; escalate only agent/immediate; shared taskkill budget for retry

- Commit: `f965a79fc08c354d829347da9d092a9f130fe09e`
- Upstream: https://github.com/stablyai/orca/pull/10638
- Fork portfolio: https://github.com/innocarpe/orca/pull/90

nwparker:
Closing — superseded, though the core observation was correct and did get fixed.

> ConPTY forceKill is a no-op after the first bare kill (`nodePtyKillIssued`), so stubborn nested agents that ignore soft stop must be re-tree-killed on escalate.

That is right, and it is the same thing @hanbong5938 found in #9854. It is fixed in **#15886**: `forceKill()` on Windows now terminates the pane's Win32 job object instead of returning early. The job is exact — it names only the tree Orca created — so no re-`taskkill`, no identity probe, and no escalation ladder is needed. #15886 also wires the job into the daemon teardown path, which is where worktree delete actually runs (measured in #11047).

The mid-wait re-escalation and the separate escalation grace window are not carried over. They existed because the first kill could silently do nothing; with the job it either terminates the tree or reports `unavailable`, so a second round of the same call has nothing new to try. If a tree survives a `TerminateJobObject` we would want to know that directly rather than retry into it.

Thanks — the diagnosis held up.

nwparker:
Superseded by the job objects in #15755/#15886, which fix the forceKill no-op this PR diagnosed.

nwparker:

Issue #10098, opened 2026-07-23
[Bug]: Codex Chat UI ordinary message unexpectedly switches to Terminal


Airine:
## Operating system

macOS

## Orca version

1.4.150

### Short summary

Sending an ordinary natural-language message from Codex Chat UI caused the pane to switch back to the hosted Codex terminal. The message did not intentionally invoke a TUI-only picker or slash command.

This was observed with Codex CLI `0.145.0`. That version is a compatibility hypothesis, **not a confirmed root cause**.

### What happened?

1. Opened an Orca-managed Codex terminal with Experimental Chat UI enabled.
2. Switched to Chat UI / used Chat UI as the supported-agent view.
3. Sent one ordinary natural-language message.
4. Orca revealed the terminal view, showing the running command:

   ```text
   codex '--dangerously-bypass-approvals-and-sandbox'
   ```

5. The Codex TUI reported version `0.145.0`.

### Expected behavior

An ordinary chat send should:

- stay in Chat UI;
- show the optimistic user row;
- write the message and delayed Enter to the hosted PTY;
- wait for the Codex rollout transcript to catch up.

Switching to Terminal is expected for explicit TUI handoffs such as `/model`, an `agent-picker`, or an `interaction-required` command outcome. It is not expected for ordinary prose.

### Source-level evidence

I checked the public `v1.4.150` tag to avoid relying on a later prerelease:

- `NativeChatComposer.tsx` routes ordinary sends through `sendNativeChatMessage(...)` and `onOptimisticSend(...)`; that path does not call `onSwitchToTerminal`.
- `onSwitchToTerminal` is passed into session-option handling for picker/interaction handoffs.
- `TerminalPane.tsx` can also exit Chat UI after `handleConfirmedAgentExit(...)` resolves an `exitChat` route.
- A failed initial prompt delivery is rendered as not delivered; it does not itself switch the view.

Relevant files:

```text
src/renderer/src/components/native-chat/NativeChatComposer.tsx
src/renderer/src/components/native-chat/native-chat-session-option-apply.ts
src/renderer/src/components/terminal-pane/TerminalPane.tsx
src/renderer/src/lib/agent-launch-prompt-delivery.ts
```

Orca's own `docs/native-chat-codex-tui-parity.md` also describes current Codex Chat UI as a PTY harness over the hosted TUI.

### What is confirmed vs unknown

Confirmed:

- Codex is a supported Chat UI agent.
[...]

PR #10639, opened 2026-07-26
fix(native-chat): resolve WSL Codex transcripts for Chat UI toggle

### LINKED ISSUE #10523: [Bug]: Unable to switch between Chat UI and CLI for WSL Codex
### Operating system

Windows

### Orca version

*(unknown — please fill in from Orca → About Orca)*

### Details

**Short summary**

On Windows with a WSL-backed project, switching between Chat UI and the Codex CLI/terminal view does not work reliably for Codex sessions.

**What happened?**

For Codex running under WSL:

* Users cannot freely switch between Chat UI and the CLI/terminal view, or the switch fails / is ineffective.
* This blocks the expected dual-mode workflow (Chat UI for conversational use, CLI/TUI when full terminal interaction is needed).

**How can we reproduce it?**

1. Open a WSL project in Orca on Windows.
2. Launch Codex in the WSL-backed terminal (or an Orca-managed WSL Codex session).
3. Attempt to switch from the CLI/terminal view to Chat UI (and back).
4. Observe that switching fails, is unavailable, or does not stick (e.g. immediately falls back to the other mode).

**Expected behavior**

* Chat UI ↔ CLI/terminal switching should work the same for WSL-hosted Codex as for native/local Codex sessions.
* The selected view should remain active while the Codex process is still running.

**Possibly related**

These may be the same underlying WSL Codex Chat/transcript/session-detection problem, or close siblings:

* stablyai/orca#10326 — Chat UI does not show Codex responses for WSL projects and sometimes falls back to terminal view
* stablyai/orca#10128 — [WSL] Experimental GUI Chat failed to load conversation in the Codex terminal
* stablyai/orca#10098 — Codex Chat UI ordinary message unexpectedly switches to Terminal (macOS report; similar handoff symptom)
* stablyai/orca#8401 / [#8418](<https://github.com/stablyai/orca/issues/8418>) — transcript / Chat UI race family
* stablyai/orca#8612 — Orca managed Codex state path (`CODEX_HOME`) redirect

**Notes for triage**

* If this is confirmed as a duplicate of stablyai/orca#10326 (view falls back / cannot stay in Chat UI on WSL), please close as duplicate and keep stablyai/orca#10326 as the canonical tracker.
* If the failure mode is different (e.g. toggle disabled, switch control no-ops both directions, or CLI→Chat never enters Chat at all), keep this issue open with the distinct repro.

* Platform: Windows + WSL2
* Agent: Codex (WSL runtime)
* Capture whether transcript lives under `\\wsl.localhost\...` / Orca managed codex-runtime-home vs `~/.codex` (see comments on stablyai/orca#10326)

innocarpe:
## Summary
- Map guest Linux Codex transcript paths to host-readable `\\wsl.localhost\…` so Chat UI can load WSL Codex sessions.
- Search WSL managed + system Codex session roots; wire transcript watch to the translated path.
- Also includes false agent-title exit suppress so Chat does not bounce back to Terminal after send (#10098 class).

## Fixes
Closes #10523

## Test plan
- [x] host-readable-transcript-path / session-file-resolver / transcript-watch / exit-suppress tests
- [ ] Manual Windows+WSL: Codex session → switch Chat UI ↔ CLI and stay put

## ELI5

Chat UI couldn't load Codex session transcripts that lived inside WSL because paths weren't translated for Windows. Transcripts map to host-readable paths so Chat UI works for WSL Codex.


innocarpe:
### Sync update (`fa41f1c61`)

Address CodeRabbit: classify Windows guest paths before pathExists; do not replay stale confirmed-exit after title recovery (generation + revalidate).

- Commit: `fa41f1c614f7fc6139c6bff4189af1717e5f353e`
- Upstream: https://github.com/stablyai/orca/pull/10639
- Fork portfolio: https://github.com/innocarpe/orca/pull/91

6Soo:
Reporter of #11327 here. I verified this PR's approach end to end, and found two things the description doesn't cover: **it also fixes the Android app**, and **the Claude path is still missing**.

**Environment**: Orca 1.4.162, Windows 11 26200, WSL2 Ubuntu-24.04, Codex + Claude agents, Orca mobile on Android. Chat UI was broken on *both* desktop and Android after moving my agent runtime into WSL.

### Verifying the approach

Rather than patch the app, I pointed `C:\home\<me>` and `D:\home\<me>` at `\\wsl.localhost\Ubuntu-24.04\home\<me>` with directory symlinks, so the bare Linux path Orca stores resolves on its own. That isolates exactly the one thing this PR changes.

```
# before
Test-Path '/home/<me>/.local/share/orca/codex-runtime-home/.../rollout-*.jsonl'
# False   (resolved from both C:\ and D:\)

# after, read the way the loader would
path.resolve -> C:\home\<me>\.local\share\orca\...\rollout-*.jsonl
statSync     -> OK, 34,855,245 bytes
read         -> OK  {"timestamp":"...","type":"session_meta",...}
```

**Path translation alone is sufficient.** I A/B'd it: remove the symlinks and Chat breaks again immediately; restore them and it recovers — with Chat sitting on the spinner, it finished loading on its own, no interaction. So the loader is already retrying; it is only the path resolution that fails.

### This is not desktop-only

The Android app was fixed by the same host-side change. Nothing touched on the phone, nothing restarted on the PC, no new session — an existing session just opened in Chat and worked, and it broke and recovered in lockstep with the symlinks during the A/B. The mobile relay inherits both the failure and the fix.

Worth mentioning in the PR description, since mobile users hitting this may not connect it to a desktop-side path bug.

### Gap: Claude sessions are still unhandled

`host-readable-transcript-path.ts` says *"WSL Codex/Claude hooks report Linux paths"*, but `wslCodexSessionsDirs()` only adds `.local/share/orca/codex-runtime-home/home/sessions` and `.codex/sessions`. There is no `.claude/projects` entry.

Orca stores Claude transcripts as bare Linux paths too — from my `profiles/local-default/orca-data.json`:

```
codex   19 entries   /home/<me>/.local/share/orca/codex-runtime-home/...
claude   7 entries   /home/<me>/.claude/projects/-home-<me>-AX-cost/*.jsonl
```

So as written, this PR fixes Codex but leaves WSL **Claude** Chat broken in the same way. My symlink workaround covers both, which is how I noticed.

Happy to open a follow-up adding the Claude session root if you'd prefer to keep this PR scoped — just say which you'd rather have.

Either way, I can tick the unchecked manual Windows+WSL item from the reporting side.

innocarpe:
@6Soo Thanks for the thorough A/B validation. The symlink experiment cleanly isolates the host-path translation this PR changes, and the Android result is useful evidence that the same host-side fix also unblocks mobile relay transcript reads.

I agree on scope: let’s keep #10639 focused on Codex. Claude’s `.claude/projects` discovery needs its own agent-specific coverage, so it should be handled as a separate focused follow-up under #11327 rather than expanding this already-reviewed diff.

I’m leaving the manual patched-build checkbox unchecked because this validated equivalent path mapping rather than this PR build itself, but it materially validates the diagnosis and fix shape. Your offer to test the Claude follow-up would be very helpful — thank you.

Issue #10491, opened 2026-07-25
[Bug]: Workspace identity becomes ambiguous when opening a new terminal or agent in a specific project


sinnny:
### Operating system

macOS

### Details

Short summary:
Orca shows a “Workspace identity is ambiguous across hosts” error when adding a new terminal or agent in a specific project.

What happened?

In one specific project, creating a new terminal or starting a new agent from the workspace triggers the following error:

> Workspace identity is ambiguous across hosts. Refresh projects and try again.

Refreshing the projects does not reliably resolve the issue, and I am unable to add the new terminal or agent normally.

How can we reproduce it?

1. Open the affected project in Orca.
2. Enter the project workspace.
3. Add a new terminal or start a new agent.
4. The error banner appears.

Anything else that might help:

* The issue appears to affect only one specific project.
* It happens with both new terminals and new agents.
* Refreshing the project list does not consistently fix the issue.
* Screenshot attached.
* Expected behavior: The new terminal or agent should open normally in the current workspace.

<img width="1440" height="173" alt="Image" src="https://github.com/user-attachments/assets/5407e2ca-8f7a-491a-97fc-cc3f2bb71ea0" />

NHP95:
Same issue after updating latest version

hanmilLee:
Same here

nwparker:
## Cannot reproduce (cannot_repro)

“Workspace identity is ambiguous across hosts” on one specific project. Error is project-data-dependent; cannot recreate without the reporter’s project/host records.

**Need:** project host setup list (redacted), whether multiple hosts share the same path/identity, and whether Refresh projects changes host ids.

PR #10649, opened 2026-07-26
fix(worktree): resolve multi-host identity from worktree host without uncaught errors

### LINKED ISSUE #10634: [Bug]: A project registered on two hosts (local + SSH) permanently throws "Workspace identity is ambiguous across hosts" — uncaught, and "Refresh projects" cannot fix it
### Operating system

macOS

### Orca version

v1.4.155

### Details

**Short summary**

A project that is registered on **two hosts** (one `local` host setup + one `ssh:*` host setup for the *same* git remote) permanently poisons every workspace of that project with:

> Workspace identity is ambiguous across hosts. Refresh projects and try again.

"Refresh projects" cannot fix it, because nothing is stale — both host setups are valid and intentional. And the error is thrown from a *background* notification path, so it surfaces as an **uncaught renderer error**, not as a handled, recoverable state.

I believe this is the reproduction #10287 asked for (`cannot_repro` — "register the same folder under two host identities (local + SSH)"), and it very likely explains #10491 ("it only happens in one specific project") and the closed #10251.

**The ambiguity is structural, not stale state**

`orca project setups` on this machine shows exactly one project with two host setups:

```
projectId: github:<owner>/<repo>
  setup A   hostId = local                path = ~/Documents/<repo>/<subdir>
  setup B   hostId = ssh:ssh-…-8v35nv     path = /home/<user>/<repo>
```

Both were created with `imported-existing-folder`, and both resolve to the same
`gitRemoteIdentity.canonicalKey` (`github.com/<owner>/<repo>`), so Orca merges them under one
project id. Every other project in this registry has exactly one host setup.

`orca worktree list` shows ~20 worktrees owned by setup B (ssh) and 1 owned by setup A (local).
Each individual worktree carries exactly one `hostId` — only the **project** spans hosts.

Since "Refresh projects" just re-reads the same registry, it can never resolve this; the error
returns immediately and survives app restarts.

**The error is uncaught, and fires from passive terminal notifications**

6 occurrences today (06:54, 07:03, 07:10, 07:11, 07:28, 09:12 KST) in
`~/Library/Application Support/Orca/logs/main.trace.ndjson`, recorded as
`crash-breadcrumb` / `renderer_error`, all with the identical stack:

```
Uncaught Error: Workspace identity is ambiguous across hosts. Refresh projects and try again.
    at settingsForWorktreeOwner$1
    at Object.markWorktreeUnread
    at dispatchTerminalNotification
```

`markWorktreeUnread` is not a user action — this is the "a terminal produced output / an agent
finished" path, so simply having a session running in that project throws repeatedly.

In the shipped bundle the ambiguous case replaces the whole pty controller with a stub that throws
on every method, which is consistent with that:

```js
const Xe = Tl
  ? vat("Workspace identity is ambiguous across hosts. Refresh projects and try again.")
  : (gn ? yat(gn, hn) : _xe(hn));
```

**How to reproduce**

1. Import an existing folder for repo R on the **local** host.
2. Import the same repo R (same git remote) on an **SSH** host. Orca merges both into one project id
   with two entries in `projectHostSetups`.
3. Open a workspace of that project and let a terminal/agent run so it emits a notification
   (or create a new terminal/agent, as in #10491).
4. → the banner appears and an uncaught renderer error is logged. "Refresh projects" does not clear
   it; it comes back after a restart.

Verify the precondition with `orca project setups --json` — the affected project is the one whose
`projectId` appears with more than one distinct `hostId`.

**Expected behavior**

- A worktree already carries exactly one unambiguous `hostId`, so the settings lookup should resolve
  from the **worktree's own owner** instead of from the project. A project legitimately existing on
  both a laptop and a remote host is a normal Orca setup, not an ambiguous one.
- Failing that, the failure should at least be **handled** (banner + retry) rather than thrown as an
  uncaught error from a background notification path.
- The message should not advise "Refresh projects", which cannot fix a structural condition. If
  disambiguation really is required, the UI needs a way to do it (pick the host for this workspace,
  or split the project per host).

**Environment**

- Orca 1.4.155, macOS 27.0, arm64
- One project registered on `local` + one SSH host; all other projects are single-host
- Related: #10287 (Windows, `cannot_repro` — this looks like the missing repro), #10491 (macOS, same
  message, triggered on new terminal/agent), #10251 (closed, ambiguous host identity in 1.4.154)

innocarpe:
## Summary
- Host-stamped worktrees resolve to their own `hostId` even when the project is registered on multiple hosts (local + SSH).
- Focused execution host can disambiguate multi-host projections when it uniquely selects one owner.
- Background paths (`markWorktreeUnread` / activity / unread clear) no longer throw uncaught renderer errors when identity is still ambiguous — UI state updates and host persist is skipped with a warn.

## Fixes
Closes #10634

## Related
- Complements focused-host preference work for #10491 / multi-host projects.
- Structural multi-host projects are valid; "Refresh projects" cannot fix them.

## Test plan
- [x] worktree-operation-route / terminal-worktree-route / runtime-owner / worktrees / file-explorer provenance tests
- [ ] Manual: import same repo on local + SSH, run terminal/agent, confirm no uncaught ambiguous identity banner from notifications

## ELI5

Worktrees registered on multiple hosts no longer throw when identity is ambiguous; they resolve via the worktree's own host or focused host, and background updates skip safely instead of crashing the UI.

### DIFF
--- a/src/renderer/src/lib/terminal-worktree-route.test.ts
+++ b/src/renderer/src/lib/terminal-worktree-route.test.ts
@@ -72,6 +72,73 @@ describe('resolveTerminalWorktreeRoute', () => {
   it('does not treat a folder workspace as an unresolved worktree', () => {
     expect(resolveTerminalWorktreeRoute(localState(), folderWorkspaceKey('abc-123'))).not.toBeNull()
   })
+
+  it('routes a focused-host copy when the same worktree id is projected on two hosts (#10491)', () => {
+    const state = localState({
+      settings: { activeRuntimeEnvironmentId: 'hub-a' },
+      runtimeEnvironments: [{ id: 'hub-a' }, { id: 'hub-b' }],
+      worktreesByRepo: {
+        'repo-1': [
+          { id: 'repo-1::/w', repoId: 'repo-1', hostId: 'local' },
+          {
+            id: 'repo-1::/w',
+            repoId: 'repo-1',
+            hostId: 'ssh:private-target',
+            runtimeOwnerEnvironmentId: 'hub-a'
+          }
+        ]
+      }
+    } as unknown as Partial<AppState>)
+    expect(resolveTerminalWorktreeRoute(state, 'repo-1::/w')).toEqual({
+      runtimeEnvironmentId: 'hub-a'
+    })
+  })
+
+  it('fails closed when multi-host projections share an id and focus matches none (#10491)', () => {
+    const state = localState({
+      settings: { activeRuntimeEnvironmentId: 'hub-c' },
+      runtimeEnvironments: [{ id: 'hub-a' }, { id: 'hub-b' }, { id: 'hub-c' }],
+      worktreesByRepo: {
+        'repo-1': [
+          {
+            id: 'repo-1::/w',
+            repoId: 'repo-1',
+            hostId: 'ssh:private-a',
+            runtimeOwnerEnvironmentId: 'hub-a'
+          },
+          {
+            id: 'repo-1::/w',
+            repoId: 'repo-1',
+            hostId: 'ssh:private-b',
+            runtimeOwnerEnvironmentId: 'hub-b'
+          }
+        ]
+      }
+    } as unknown as Partial<AppState>)
+    expect(resolveTerminalWorktreeRoute(state, 'repo-1::/w')).toBeNull()
+  })
+
+  it('routes a host-stamped worktree when the project is multi-host (#10634)', () => {
+    const state = localState({
+      repos: [
+        { id: 'repo-1', connectionId: null, executionHostId: 'local' },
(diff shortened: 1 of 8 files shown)


innocarpe:
### Sync update (`0e2a14849`)

Address CodeRabbit: scope explicit-route worktree scan to the worktree's own repoId instead of all repos.

- Commit: `0e2a14849bf08b7d9b43b7045c0fd7dcadf1fac6`
- Upstream: https://github.com/stablyai/orca/pull/10649
- Fork portfolio: https://github.com/innocarpe/orca/pull/92

nwparker:

- Issue #10634 was fixed by #10986 (`fix(worktrees): resolve a two-host project by the worktree's own host (#10634)`)

Superseding merge: #10986

PR #10650, opened 2026-07-26
feat(ports): filter SSH ports by connecting user with show-others toggle

### LINKED ISSUE #10563: [Feature]: Ports panel on shared SSH hosts — default to the current user's ports, toggle to show other users' (with username)
**Problem / use case**

On shared SSH hosts (multiple developers on one Linux box), the Ports panel lists every listening port on the machine. Most rows belong to other users: they're noise for finding your own dev server, and forwarding one of them by mistake targets a coworker's service. There's currently no way to tell which ports are yours.

**Request**

Default the detected-ports list to ports owned by the connecting user, with a toggle ("Show ports from other users") that reveals the rest, annotated with the owning username.

**Why this is cheap — the data is already half-collected**

The relay's Linux scanner (`src/relay/port-scan-handler.ts`, `scanLinuxListeningPorts`) parses `/proc/net/tcp` + `/proc/net/tcp6`, which include an **owner-UID column per socket** — the parser currently extracts only `{port, host, inode}` and discards the uid.

Ownership is also already implicitly visible in today's behavior: the PID/process-name mapping walks `/proc/<pid>/fd`, which the kernel only permits for the relay user's own processes — so other users' ports already appear as anonymous rows (no pid, no processName). The scanner half-knows whose ports these are; it just doesn't record it.

**Proposed v1**

- Parse the uid column in `readProcNet`; add optional `uid?: number` and `username?: string` to `DetectedPort` (`src/shared/ssh-types.ts`). Resolve uid→username via one cached `/etc/passwd` read per scan (fall back to the numeric uid string on any failure).
- Mark ownership relay-side by comparing against the relay process's uid (`process.getuid()`), e.g. `ownedByConnectingUser?: boolean` — the relay runs as the SSH user, so this is exactly "my ports."
- Ports panel (`WorkspacePortScanner.tsx` / right-blade ports UI): default filter to owned ports; a toggle reveals others with the username shown on the row. Persist the toggle per target or globally, whichever is conventional.
- Auto-forward suggestions should only ever consider owned ports (complementing the existing initial-scan exclusion).

**Compatibility / guards**

- Fields are optional and additive — an older relay omits them, and the UI degrades to today's unfiltered behavior (per the additive-protocol convention in #9820).
- Windows remotes (`windows-port-scan.ts`): no uid concept in the current scan; leave fields unset → unfiltered, unchanged.
- Never hide a port the user explicitly forwarded or that carries an advertised URL from their own terminal, regardless of uid resolution hiccups.

**Acceptance tests**

- Linux scan rows carry uid/username/owned flags; sockets with uid == relay uid marked owned.
- uid→username resolution failure → numeric uid shown, no scan failure.
- Old-relay payload without the new fields → panel shows all ports (no filter applied), no errors.
- Toggle off (default): only owned ports listed; toggle on: all ports with usernames.
- Auto-forward suggestion path ignores non-owned ports.

**Environment**

Windows 11 / Orca 1.4.155, SSH workspaces to shared Linux hosts (multi-user dev servers).

innocarpe:
## Summary
- Linux relay port scan now parses socket owner UID, resolves username via `/etc/passwd`, and marks `ownedByConnectingUser` against the relay process uid.
- Ports panel defaults to owned ports; optional toggle shows other users' ports with usernames.
- Auto-forward suggestions only consider owned ports. Windows / old relays omit fields and stay unfiltered.

## Fixes
Closes #10563

## Test plan
- [x] port-scan-handler, ownership helpers, retained payload admission, ssh-port-scanner, preference unit tests
- [ ] Manual: multi-user Linux SSH host — default list is own ports; toggle reveals others

## ELI5

On shared Linux SSH hosts, the Ports panel can focus on ports you own and optionally show other users' ports by name, so auto-forward suggestions aren't cluttered by everyone else's listeners.

### DIFF
--- a/src/main/ssh/ssh-port-scanner.test.ts
+++ b/src/main/ssh/ssh-port-scanner.test.ts
@@ -169,6 +169,32 @@ describe('PortScanner', () => {
     scanner.dispose()
   })
 
+  it('excludes initial-scan and non-owned ports from auto-forward suggestions', async () => {
+    const harness = createVisibilityHarness(true)
+    let ports: DetectedPort[] = [
+      { port: 3000, host: '127.0.0.1', ownedByConnectingUser: true },
+      { port: 3001, host: '127.0.0.1', ownedByConnectingUser: false, username: 'bob' }
+    ]
+    const { mux } = createMux(() => ports)
+    const scanner = new PortScanner(harness.visibility)
+    scanner.startScanning('t1', mux, vi.fn())
+
+    await vi.advanceTimersByTimeAsync(0)
+    // First scan seeds initialPorts — nothing is a "new" suggestion.
+    expect(scanner.getAutoForwardSuggestionPorts('t1')).toEqual([])
+
+    ports = [
+      { port: 3000, host: '127.0.0.1', ownedByConnectingUser: true },
+      { port: 3001, host: '127.0.0.1', ownedByConnectingUser: false, username: 'bob' },
+      { port: 4000, host: '127.0.0.1', ownedByConnectingUser: true, processName: 'node' },
+      { port: 4001, host: '127.0.0.1', ownedByConnectingUser: false, username: 'bob' }
+    ]
+    await vi.advanceTimersByTimeAsync(BASE)
+    expect(scanner.getAutoForwardSuggestionPorts('t1').map((p) => p.port)).toEqual([4000])
+
+    scanner.dispose()
+  })
+
   it('parks when the window hides mid-run and resumes with an immediate scan on show', async () => {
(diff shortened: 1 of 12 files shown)


innocarpe:
### Sync update (`29c4b0196`)

Address CodeRabbit: do not drop whole socket row on invalid uid — leave ownership metadata unset like pid.

- Commit: `29c4b01967fec0563b198468f041c5f47230d6a1`
- Upstream: https://github.com/stablyai/orca/pull/10650
- Fork portfolio: https://github.com/innocarpe/orca/pull/93

PR #10651, opened 2026-07-26
feat(jira): collect required fields for transitions (resolution form)

### LINKED ISSUE #10565: [Feature]: Support Jira transitions with required fields (resolution on Close, etc.) — currently they fail with only a raw API toast
**Problem / use case**

On Jira instances whose workflows attach required fields to transitions (a resolution on "Close/Done", a comment on "Reject", custom fields — very common on corporate Jira), those transitions cannot be completed from Orca. The status-pill popover lists them, but selecting one fails: Orca posts only the transition ID, Jira rejects it for missing required fields, and the optimistic status flip rolls back with a transient toast carrying Jira's raw error. Practical effect: tickets can be moved through simple transitions but never closed, for workflows that require a resolution.

**Cause**

`updateIssue` sends the bare minimum (`src/main/jira/issues.ts`, the `updates.transitionId` branch):

```ts
await jiraRequest(entry, `${issueBase}/transitions`, {
  body: JSON.stringify({ transition: { id: updates.transitionId } })
})
```

No `fields` / `update` payload, and `listTransitions` doesn't request `expand=transitions.fields`, so Orca never learns which fields a transition requires. This is the transition-time sibling of #4643 (issue creation fails on "Reporter is required" with no way to supply it).

**Proposed v1**

- `listTransitions`: request `expand=transitions.fields`; carry per-transition required-field metadata (`required`, schema type, `allowedValues`) on `JiraTransition`.
- In the status popover (`JiraIssueWorkspace.tsx`), transitions with no required fields behave exactly as today. For transitions with required fields, open a minimal form before posting:
  - v1 field support: **resolution** (single-select from `allowedValues` — this alone covers the dominant corporate case), plus optional comment when required.
  - POST becomes `{transition: {id}, fields: {...}}` (comment via `update.comment` per Jira API).
- Unsupported required field types (cascading selects, custom pickers): don't attempt the form — show a clear message naming the missing fields with an "Open in Jira" link to complete the transition there. Never fail with only the raw API error.

**Related hardening (small, same surfaces, optional to split out)**

- `listTransitions` currently swallows fetch errors to `console.warn` and returns `[]`, which renders as a silently disabled status pill — indistinguishable from "no transitions". Surface load errors on the pill (tooltip/retry).
- The status pill has no interactive affordance; a chevron would make the transition feature discoverable at all.

**Acceptance tests**

- Transition without required fields → identical single-click behavior to today.
- Transition requiring resolution → form shows allowed resolutions, posts `fields.resolution`, succeeds; status updates.
- Transition requiring an unsupported field type → no bare API-error toast; message names the fields and links to the issue in Jira.
- `expand=transitions.fields` unavailable/failing (older Jira, permissions) → degrade to today's bare-ID post; execution errors still surfaced.
- listTransitions fetch failure → pill shows a load-error state, not a silently disabled badge.

**Environment**

Orca 1.4.155, Jira Cloud with corporate workflows attaching required fields (resolution et al.) to several transitions.

innocarpe:
## Summary
- `listTransitions` requests `expand=transitions.fields` and maps required-field metadata (with bare-list fallback).
- Status pill: one-click for simple transitions; resolution/comment form when required; unsupported required types get a clear message + Open in Jira.
- `updateIssue` posts `fields` / comment update with the transition id.

## Fixes
Closes #10565

## Test plan
- [x] jira issues listTransitions/updateIssue tests, transition-fields unit tests, RPC schema tests
- [ ] Manual: Jira Cloud workflow with resolution-required Close transition

## ELI5

Changing Jira issue status that requires fields like Resolution now prompts for them in Orca instead of failing silently. Simple transitions stay one-click; complex ones get a small form or a clear open-in-Jira path.

### DIFF
--- a/src/main/ipc/jira.ts
+++ b/src/main/ipc/jira.ts
@@ -76,6 +76,17 @@ function normalizeIssueUpdate(value: unknown): JiraIssueUpdate | null {
   if (input.transitionId !== undefined && typeof input.transitionId !== 'string') {
     return null
   }
+  if (
+    input.transitionFields !== undefined &&
+    (input.transitionFields === null ||
+      typeof input.transitionFields !== 'object' ||
+      Array.isArray(input.transitionFields))
+  ) {
+    return null
+  }
+  if (input.transitionComment !== undefined && typeof input.transitionComment !== 'string') {
+    return null
+  }
   return input
 }
 
--- a/src/main/jira/issues.test.ts
+++ b/src/main/jira/issues.test.ts
@@ -295,7 +295,8 @@ describe('Jira issue operations', () => {
         schema: {
           type: 'option',
           custom: 'com.atlassian.jira.plugin.system.customfieldtypes:select',
-          items: undefined
+          items: undefined,
+          system: undefined
         },
         allowedValues: [{ id: 'option-1', value: 'High', name: undefined }]
       }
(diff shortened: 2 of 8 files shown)


innocarpe:
### Sync update (`e2a054ac0`)

Address CodeRabbit: unify resolution draft seed and option values via transitionAllowedValueKey (id ?? value ?? name).

- Commit: `e2a054ac09432a092a39c1099fa42075078d9f54`
- Upstream: https://github.com/stablyai/orca/pull/10651
- Fork portfolio: https://github.com/innocarpe/orca/pull/94

PR #10658, opened 2026-07-26
fix(mobile): show chat/terminal content when Chat UI sessions open

### LINKED ISSUE #10630: [Bug] Cannot see chat or terminal content from mobile app
<kbd>[rednickel](https://discord.com/channels/1492228674081263786/1530726323809685504/1530726323809685504)</kbd> [rednickel](https://discord.com/channels/1492228674081263786/1530726323809685504/1530726323809685504)  `via Discord`

iOS version 26.3
Orca version v1.4.156
Claude code version v2.1.220

I type to send a message to claude and I can't see anything (what I sent or what was responded).

If I open the same claude session from my laptop, I can confirm the message went through and there is a history.

In addition, something similar happens when opening a regular terminal. From my phone, I type `ls [Enter]` and don't see anything (blank screen). But from the Orca IDE I can confirm the `ls` command was executed and can see the output.

innocarpe:
## Summary
- Mobile agent-status projection now includes `providerSession` so Chat UI can subscribe to transcripts.
- Bump `agentStatusEpoch` when provider session id/path attaches (same-state working pings no longer drop the attach).
- Chat UI overlay only covers the terminal when `sessionId` is present; otherwise keep the live terminal visible (matches the reported workaround path without forcing Chat UI off permanently).

## Fixes
Closes #10630

## Notes
- Desktop restart required after deploy so new projection/epoch runs for mobile clients.
- May overlap maintainer WIP around mobile Chat UI readiness; scoped to blank-content root cause.

## Test plan
- [x] sync-runtime-graph + agent-status provider-session + mobile eligibility/controller tests
- [ ] Manual: iOS + Chat UI on + Claude session — messages visible on phone

## ELI5

Opening Chat UI sessions on mobile sometimes showed a blank surface instead of chat or terminal content. Mobile now gets the provider session data it needs, and falls back to the live terminal when chat isn't ready.

### DIFF
--- a/mobile/src/session/mobile-native-chat-eligibility.test.ts
+++ b/mobile/src/session/mobile-native-chat-eligibility.test.ts
@@ -3,7 +3,8 @@ import type { AgentStatusEntry } from '../../../src/shared/agent-status-types'
 import {
   canShowMobileNativeChat,
   isMobileNativeChatTranscriptReadable,
-  resolveMobileNativeChat
+  resolveMobileNativeChat,
+  shouldShowMobileNativeChatOverlay
 } from './mobile-native-chat-eligibility'
 
 function status(overrides: Partial<AgentStatusEntry> = {}): AgentStatusEntry {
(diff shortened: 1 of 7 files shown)


brennanb2025:
Took a look at this while working #10630 — opening #11260 with a different fix, so flagging what I found rather than duplicating effort.

The diagnosis here doesn't hold up: `buildRuntimeMobileAgentStatusProjection` is a change-detection sync key, not the wire payload. The payload at `sync-runtime-graph.ts:1434` already spreads the whole `agentStatus` entry, `providerSession` included, so the id does reach mobile. Adding it to the projection changes *when* the republish fires, so the effect is one coalesced status ping of latency rather than absence.

The overlay half has a sharper problem. `canReuseExistingProviderSession` in `agent-status.ts` required `payload.state !== 'done'`, so a `done` ping with no session metadata dropped `providerSession` entirely. Gating `showNativeChat` on `sessionId != null` therefore flips chat back to terminal at the end of every turn.

That drop turned out to be the actual root cause — mobile clears the transcript and skips the subscription when `sessionId` is null, so the chat blanks precisely while the agent sits idle at its prompt, which is when you're reading it. #11260 fixes the drop in both layers that lose it (the renderer store and the main-process OSC ingest, the latter mattering for `orca serve` and for surviving a restart) and leaves the overlay logic alone.

Also worth noting this conflicts with #10787's per-entry cached projection refactor, which has since landed.

Thanks for digging into this one — the mobile-side symptom write-up was genuinely useful in tracking it down.

nwparker:

- Issue #10630 was fixed by #11260 (`fix(mobile): keep the provider session id alive while an agent sits idle`)

Superseding merge: #11260

PR #10664, opened 2026-07-26
Improve Korean and Chinese sidebar translations

AmethystLiang:
## Summary
- Updated the Korean SidebarNav label for the New button to a shorter translation.
- Updated the Chinese SidebarNav label for the New button to clarify it means new features.
- Fixed the Chinese translation for the GH PR label, which was incorrectly translated as a biomedical term.

## Testing
- Not run; localization-only changes.


nwparker:

PR #7081, opened 2026-07-02
Fix rename inputs committing on CJK IME Enter confirmation (#6980)

xianjianlf2:
## Summary

Fixes #6980.

Worktree/workspace inline rename, tab rename, editor file-tab rename, and terminal pane title rename all committed whenever `event.key === 'Enter'`, without guarding IME composition state. Pressing Enter to confirm a CJK (Japanese/Chinese/Korean) conversion candidate therefore submitted the rename mid-composition with a half-converted value — the same class of bug as #742, but on rename surfaces the earlier fixes didn't cover.

## Changes

- Extract the existing local `isComposing || keyCode === 229` guard (previously private to `WorkspaceDirectorySetting`) into a shared, reusable helper `src/renderer/src/lib/ime-composition-keyboard-event.ts` (`isImeCompositionKeyDown`), with a unit test.
- Apply the guard to every rename `Enter` handler that lacked it:
  - `WorktreeTitleInlineRename.tsx` — worktree card inline rename **and** workspace rename (⌘⇧R, which opens this same editor)
  - `SortableTab.tsx` — tab rename (⌘R)
  - `TerminalPaneHeaderOverlay.tsx` — terminal pane title rename
  - `EditorFileTab.tsx` — editor file-tab rename
- Refactor `WorkspaceDirectorySetting.tsx` to consume the shared helper (removes the duplicated local copy).

The `keyCode === 229` fallback is kept for IMEs/browsers that don't set `isComposing` on keydown, matching the reference in the issue.

## Behavior

- Enter while IME composition is active → only confirms the conversion candidate; the rename input stays open and nothing is committed.
- A subsequent, non-composition Enter → commits the rename as before.
- Non-CJK / non-composition input is unchanged.

## Testing

- `vitest` — new `ime-composition-keyboard-event.test.ts` plus existing `WorkspaceDirectorySetting`, `WorktreeTitleInlineRename`, and `SortableTab` rename tests pass.
- `oxlint` (incl. react-doctor) and web `tsgo` typecheck pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10711, opened 2026-07-26
fix(ui): ignore the IME-composition Enter in name and notes inputs

makoto-developer:
## Summary

CJK IMEs (Japanese/Chinese/Korean) fire a `keydown` for the Enter that only confirms a conversion candidate. Five inputs committed on that keydown, so they saved a half-typed value while the user was still composing:

- **Project group name dialog** (New/Rename Project Group) — the field sits in a `<form>`, so the composition Enter submitted it implicitly.
- **Host rename dialog**
- **Worktree details dialog** — display name, GH issue, GH PR, and the notes textarea. The textarea saves on a plain Enter and closes the dialog, so a Japanese note could not get past its first conversion.
- **Untitled file rename dialog** — file name and folder fields.
- **Editor header inline file rename**

All five now use the existing `isImeCompositionKeyDown()` helper — the same guard already applied to worktree titles (#7081), tabs, and the workspace name field (#9526). It reads `nativeEvent.isComposing` with a `keyCode === 229` fallback for IMEs that do not set the flag on keydown.

`ProjectGroupNameDialog` differs slightly: it cancels the composition Enter with `preventDefault()` so the form's implicit submission never fires, rather than returning early from a submit handler.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`
- [x] Added 5 `*.ime-enter.test.tsx` files (18 cases). Each covers both IME signals (`isComposing` and `keyCode === 229`) and keeps a plain-Enter control case, so a regression that drops the guard — or one that swallows a real Enter — fails the suite.

Pre-existing failure unrelated to this PR: `verify:localization-coverage` flags unlocalized "Ghostty" keywords already on `main`.

## AI Review Report

Reviewed with Claude Code. Checks and findings:

- **Cross-platform (macOS/Linux/Windows):** IME composition state is reported by Chromium on all three platforms; the guard adds no platform branch and touches no shortcut, label, path, or shell behavior. `WorktreeMetaDialog` keeps its existing `isScreenSubmitShortcut()` call, which already resolves ⌘/Ctrl per platform — the guard runs before it and does not alter that resolution.
- **SSH / remote / folder workspaces:** all five inputs are renderer-local; no IPC, git, or host-dependent path is involved.
- **Agent / integration / git provider neutrality:** no provider-specific code touched.
- **Performance:** one boolean check at the top of existing keydown handlers.
- **UI quality:** no rendered output changes; Escape during composition now cancels the conversion instead of the dialog, which matches the guards already shipped for worktree titles.
- **Risk considered:** guarding on `keyCode === 229` could in theory suppress a legitimate Enter. It cannot here — 229 is only reported while an IME owns the key event, and the plain-Enter control tests confirm the normal path still commits.

## Security Audit

No new input parsing, command execution, path handling, auth, secret, dependency, or IPC surface. The change only decides whether an existing save handler runs for a given keydown. Saved values continue through the same validation as before.

## Notes

Reproducing needs a CJK IME: type a Japanese word, press Enter to confirm the conversion, and observe that the value is no longer committed mid-composition.

## ELI5

With Japanese/Chinese/Korean keyboards, pressing Enter only to confirm a conversion candidate also submitted name/notes fields. Orca ignores that composition Enter so you finish typing before save.

### DIFF
--- a/src/renderer/src/components/editor/EditorPanelHeaderPath.ime-enter.test.tsx
+++ b/src/renderer/src/components/editor/EditorPanelHeaderPath.ime-enter.test.tsx
@@ -1,0 +1,124 @@
+// @vitest-environment happy-dom
+
+import { act } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import type { OpenFile } from '@/store/slices/editor'
+import { EditorPanelHeaderPath } from './EditorPanelHeaderPath'
+
+const { commitRename, cancelRename } = vi.hoisted(() => ({
+  commitRename: vi.fn(),
+  cancelRename: vi.fn()
+}))
+
+// Why: the rename field only mounts while the hook reports isRenaming, so the
+// hook is stubbed into that state instead of driving the whole rename flow.
+vi.mock('./editor-header-file-rename', () => ({
+  useEditorHeaderFileRename: () => ({
+    canRename: true,
+    currentFileName: '議事録.md',
+    isRenaming: true,
+    renameInputRef: { current: null },
+    openRenameInput: vi.fn(),
+    commitRename,
+    cancelRename
+  })
+}))
+
+vi.mock('@/hooks/useShortcutLabel', () => ({
+  useShortcutLabel: () => '⌘⇧M'
+}))
+
+const activeFile: OpenFile = {
+  id: 'edit:/repo/議事録.md',
+  filePath: '/repo/議事録.md',
+  relativePath: '議事録.md',
+  worktreeId: 'repo::/repo',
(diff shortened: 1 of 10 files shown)


OrcaWin:
Thank you for the IME composition fix. Your original 10-file Enter patch is preserved byte-for-byte (stable patch ID 13a5659c316e653c8a2b124af74c1d44a969cb42) as the first commit of #11067, with you retained as its author. That PR now adds a separate follow-up protecting Radix's capture-phase Escape handling during composition; the combined branch passes the focused 34-case suite and all 22 GitHub checks. Closing this PR as superseded by #11067 so the contributor patch and the required follow-up land together.

PR #2238, opened 2026-05-18
Add Jira task provider support

nwparker:
## Summary
- Add Jira as a task provider with connection, site selection, issue listing/search, issue creation, and issue workspace support.
- Wire Jira through IPC, runtime RPC, preload APIs, renderer store, settings, and task provider visibility.
- Rebase provider availability on current main so GitLab and Linear availability gating stays intact while Jira remains visible for first-time setup.

## Validation
- pnpm run typecheck
- pnpm run lint
- pnpm exec vitest run --config config/vitest.config.ts src/shared/task-providers.test.ts src/main/runtime/rpc/methods/jira.test.ts src/main/ipc/register-core-handlers.test.ts src/main/persistence.test.ts src/renderer/src/store/slices/ui.test.ts src/renderer/src/store/slices/settings.test.ts src/renderer/src/store/slices/diffComments.test.ts src/renderer/src/store/slices/tabs.test.ts src/renderer/src/store/slices/store-session-cascades.test.ts

Fixes #1310

PR #10715, opened 2026-07-26
fix(runtime): stop rejecting ui.set when Linear or Jira resume fields are present

giraphant:
## Summary

The runtime `ui.set` RPC rejects any UI-state update that carries `linearMode`, `linearContext`, `jiraPreset`, or `jiraQuery`.

The renderer persists all four in `TaskResumeState`, but the Zod schema guarding the RPC only lists six of the type's ten fields. The schema is `.strict()`, so an unlisted key fails `safeParse` and the dispatcher returns `INVALID_ARGUMENT` — Zod does not strip it, the **whole payload is dropped**. A paired web/mobile/relay client therefore loses its entire UI-state update whenever any of these fields is present, which is any session where the user has opened Linear or Jira.

Desktop is unaffected: it writes through the `ui:set` IPC handler, which has no schema. That is why this has gone unnoticed.

The fields were added by #3966 (`linearMode`, `linearContext`) and #2238 (`jira*`) without a matching schema update. This only restores parity with `shared/types.ts`; no behavior is added.

The same fix was previously authored by @brennanb2025 in `fa88ee4` and explicitly written to be cherry-pickable, but the branch it lived on was reverted and it was lost. This reuses that work, with credit.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

The existing `accepts persisted literal UI arrays and nested UI state` case in `client-ui.test.ts` was extended with all four fields. It is a genuine regression test, not a restatement of the schema: revert the schema change and the test fails because `updateUIState` is never called at all.

`pnpm test` reports 8 pre-existing failing files on my machine (`window.localStorage` undefined under this environment, e.g. `use-markup-draw-hint.test.ts`, `agent-exec-handler.test.ts`). I confirmed these are unrelated by stashing this change and re-running them on a clean tree — they fail identically.

## AI Review Report

- **Root cause vs. symptom** — traced both write paths (`ui:set` IPC → `persistence.updateUI` with no validation; `ui.set` RPC → strict Zod → `dispatcher.ts` `INVALID_ARGUMENT`) to confirm the drift is the actual defect and that the desktop path genuinely cannot hit it.
- **Whole-payload rejection** — verified `.strict()` rejects rather than strips, so the blast radius is the entire update, not one field.
- **Schema/type parity** — diffed every key of `TaskResumeState` in `shared/types.ts` against the schema; these four were the only gaps.
- **Cross-platform (macOS/Linux/Windows)** — no platform-dependent code touched: no paths, no shell invocation, no keyboard shortcuts or accelerators, no Electron platform APIs. The affected transport is the runtime RPC layer, which is platform-independent.
- **SSH / remote** — this is precisely the path that was broken; relay and SSH-paired clients share the RPC dispatcher and are fixed by the same change.
- **Git provider compatibility** — not applicable, no Git or provider code touched.

Flagged and resolved: my first draft of the `linearContext` sub-object omitted `.strict()`, which would have made it laxer than its sibling schemas. Corrected before commit.

## Security Audit

- **Input handling** — this widens accepted input on a client-facing RPC, so it is the relevant risk. Each added field is constrained to the narrowest form matching the type: three `z.enum()`s over closed value sets, one `z.string()`, and a `.strict()` object for `linearContext`. No `z.any()`, no `z.record()`, no passthrough. The parent object stays `.strict()`, so this does not open the door to arbitrary keys.
- **Command execution / path handling** — none. These values are persisted state read back by the renderer, never interpolated into a shell command or a filesystem path.
- **Auth / secrets** — none touched. No credential or token field is involved.
- **IPC** — no new IPC surface; an existing method's schema is corrected to match its documented type.

No follow-up needed.

## Notes

This overlaps textually with the same Zod block in the Linear view-state PR I am opening alongside this one; whichever merges second needs a trivial rebase. The two are independent in substance — this one is a pre-existing bug fix and stands on its own.

I do not use X, so no handle for the shoutout.

## ELI5

Saving UI state that included Linear or Jira resume fields failed entirely because the server schema rejected unknown keys. Those fields are accepted so resume state for Tasks clients is not dropped.


OrcaWin:
Reviewed with an independent reproduction. Confirming your diagnosis and extending it.

**Verified.** `TaskResumeState` is `.strict()` and listed only 6 of 10 fields. I reproduced `invalid_argument` with the exact message `Unrecognized key: "linearMode"` for each of the four missing fields, and confirmed `.strict()` rejects the **whole payload** rather than stripping the unknown key. Your desktop-immunity claim also checks out: `src/main/ipc/ui.ts:63` handles `ui:set` with no schema, which is exactly why this stayed invisible on desktop.

**What the fix missed.** The same `.strict()` drift exists one level up in `UiUpdate`, where 8 more renderer-written keys were unlisted. The worst is `showDotfilesByWorktree` — it rides `App.tsx`'s 150ms debounced writer, so a single dotfile toggle caused a paired client to lose sidebar widths, sort/group preferences, repo filters and agent acknowledgements in the same rejected call. `activeView` is written on every top-level view switch and was rejected every time.

**Pushed to this branch.** Rather than extend the hand-listing, I added an `AssertNoMissingKeys` type so any future field added to `PersistedUIState` or `TaskResumeState` becomes a typecheck error naming the missing key. Parity had drifted by hand across at least three separate PRs (#3966, #2238, #9002), so the root cause is the manual process, not any one omission. Main-owned state stays an explicit allow-list — remote clients still cannot write star-nag or window state, and the pre-existing tests proving that still pass.

The file split was required to satisfy `max-lines`; I did not add a disable, per AGENTS.md.

Validation: 515 tests passing across `src/main/runtime/rpc/methods/`, typecheck exit 0.

Thanks for finding this — the mobile/web-only failure mode made it genuinely easy to miss.

OrcaWin:
Follow-up on my review above — a status note so this is not silently stuck.

**My review fixes are on this branch** (`c5bdca4b`): the `UiUpdate` parity gaps and the `AssertNoMissingKeys` type that turns future drift into a typecheck error. 515 tests pass and typecheck exits 0 against the base this branch sits on.

**But the branch now conflicts with `main`,** which moved during review. The conflict is small and purely additive — main added `hideCliCreatedWorkspaces` and `hideDetachedHeadWorkspaces` to the same `UiUpdate` object where my change added `showDotfilesByWorktree`. All three belong; resolving means keeping every line from both sides.

I resolved it locally and verified it (612 tests, typecheck 0), but **I cannot push the resolution.** Both merging and rebasing pull main's own commits onto your fork, and two of those touch `.github/workflows/pr.yml` — GitHub then refuses the push because my token lacks the `workflow` OAuth scope. It inspects every incoming commit, not the net diff, so there is no way around it from my side even though this PR changes no workflow file.

**To move this forward, please merge `main` into your branch** and keep both sides of the conflict in `client-ui-schemas.ts`. Once it is mergeable, CI should run and this is ready.

Worth restating: the underlying bug is live on main right now — any paired web/mobile client that has opened Linear or Jira is silently losing its entire UI-state update. Worth unblocking.

PR #10717, opened 2026-07-26
feat(linear): persist the issue list filter and view options across sessions

giraphant:
## Summary

Linear's filter bar and View menu reset to defaults every time you leave the Tasks page, and again on every launch.

Filter a board down to a few statuses and labels, switch to another tab, come back — it is gone. Set the layout to board, group by assignee, order by updated, hide a couple of columns; same thing. Every visit starts from scratch, which makes the Linear tab painful to actually work out of.

This was never a deliberate decision. The founding resume commit (`5ce9c3b48`) scoped itself to preset and query; the attribute filter did not exist yet when it landed (#8021, two months later) and the View menu was simply never wired in.

**What now persists** — layout (list/board), grouping, ordering, display properties, and the attribute filter (statuses, priorities, assignee, labels).

**How** — these ride the existing `TaskResumeState` channel alongside `linearPreset` and `linearQuery`. No new store slice, no new persistence layer, no new IPC.

**Workspace scoping** — filter facets are workspace-scoped ids, so a filter saved under one workspace must not be applied to another. The persisted filter carries the workspace it was captured under and is dropped when a different one resolves. The previous reset effect keyed off any change including `null → id`, which on a cold start (Linear status not yet resolved) would have wiped the restored filter before the user ever saw it; an unresolved workspace is now correctly not treated as a switch.

## Screenshots

No new UI surface is introduced — no new controls, no layout or styling change. The visible difference is that the existing filter bar and View menu keep their values instead of resetting, which is hard to convey in a still.

Manually verified in a dev build against a live Linear workspace:

1. Applied a filter with both a status and a label (covers both facet kinds).
2. Changed layout list → board, changed grouping, changed ordering, toggled a display property off.
3. Navigated away to another tab and back — all retained.
4. Quit and relaunched — all retained.

Happy to attach a short screen recording if that is preferred over the description.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

11 new tests in `linear-issue-view-preferences.test.ts` plus extensions to `ui.test.ts` and `client-ui.test.ts`. They cover the parts that actually break: restore defaults, the empty-array-means-hide-everything case, omission of an empty filter, canonicalization and workspace tagging, catalog ordering of the emitted properties, dropping the tag when no workspace is resolved, all four branches of the workspace-reset rule, and recovery from a corrupt persisted filter without losing the rest of the resume state.

Following the convention already used by ~24 sibling modules, the coercion logic lives in a small pure module next to `TaskPage.tsx` rather than in it, so it is testable without rendering a 12k-line component.

`pnpm test` reports 8 pre-existing failing files on my machine (`window.localStorage` undefined under this environment). I confirmed they are unrelated by stashing this change and re-running them on a clean tree — they fail identically.

## AI Review Report

Reviewed with Claude Code, then a second adversarial pass specifically looking for over-engineering. Risks checked:

- **Cold-start race** — the original reset effect treated `null → resolved` as a workspace switch. Restoring a filter would have tripped it on every launch and cleared the filter before the user saw it. `shouldResetLinearFilterForWorkspaceChange` now returns false for an unresolved (`null`) workspace, and the effect only records non-null ids. Directly covered by tests.
- **Stale cross-workspace facets** — the inverse risk. A filter restored from workspace A must not silently apply to workspace B. Handled by the persisted workspace tag; verified the reset still fires on a genuine switch.
- **Dropping the workspace tag** — found during review: the persist effect originally passed `selectedLinearWorkspaceId` directly, so editing any view option while Linear was disconnected or unresolved would have written `linearIssueFilterWorkspaceId: undefined` and silently unscoped the saved filter. Changed to read the ref that tracks the workspace the current facets belong to.
- **Duplicated validation** — found during review: I had hand-written a Zod object for the filter in `client-ui-schemas.ts`. `LinearIssueAttributeFilterSchema` already existed and is *stricter* (id length and array bounds from the shared limits). Mine was unbounded. Replaced with the existing schema.
- **Coverage of all mutation paths** — the persist effect watches state values rather than hooking each setter, so the workspace-switch reset and the primary-team reset persist too, instead of only the paths a call-site patch would have caught.
- **Corrupt persisted state** — `parseLinearIssueAttributeFilter` throws on five distinct malformed shapes. Hydration catches it so one bad filter cannot take down the rest of the UI state. Covered by a test.
- **Cross-platform (macOS/Linux/Windows)** — no platform-dependent code touched: no keyboard shortcuts or `metaKey` checks, no shortcut labels, no filesystem paths, no shell invocation, no Electron platform APIs. State flows through the existing persistence channel, which is already platform-independent.
- **SSH / remote and folder workspaces** — nothing assumes a local or Git-worktree workspace. Both desktop (`ui:set` IPC) and paired web/mobile/relay (`ui.set` RPC) paths were updated; the RPC schema addition is what keeps remote clients from having the update rejected wholesale.
- **Git provider compatibility** — not applicable, no Git or provider code touched.
- **Performance** — one effect writing a small object on view-state change; the emitted display-property array is ordered by the canonical catalog so toggling a property off and back on produces an identical payload rather than a rewrite.
- **`max-lines`** — no new suppression added, per `AGENTS.md`. Net effect on `TaskPage.tsx` is small because the extracted module also removes a constant from it.
- **Boundary tests** — `feature-interaction-writer-boundaries.test.ts` anchors on the literal `const toggleLinearDisplayProperty` in `TaskPage.tsx`; that symbol is untouched.

## Security Audit

- **Input handling** — persisted state is untrusted on read. Hydration validates every field individually (closed-set membership for the four enums, `boolean` typeof, per-element filtering for the array) and drops anything invalid rather than trusting the blob. The filter is validated by the existing `parseLinearIssueAttributeFilter`, which enforces the shared bounds on array lengths and id lengths; a throw is caught and only that field is discarded.
- **RPC surface** — the strict Zod schema for `ui.set` was extended in lockstep so paired clients are neither rejected nor able to submit unbounded input. The filter reuses `LinearIssueAttributeFilterSchema` with its existing `.max()` bounds rather than a laxer hand-rolled copy, so this does not weaken the RPC boundary. Parent object remains `.strict()`.
- **Command execution / path handling** — none. No value here reaches a shell or a filesystem path.
- **Auth / secrets** — none. Workspace and facet ids are opaque identifiers, not credentials. The Linear token is stored separately under `~/.orca` and is untouched.
- **IPC** — no new IPC surface; existing `ui:set` / `ui.set` carry additional validated fields.

No follow-up needed.

## Notes

`linearTeamPropertyTouched` is persisted alongside the display properties on purpose. It gates the single-team auto-hide of the Team column (`TaskPage.tsx`); without it a restored session would re-apply the auto-hide and stomp an explicit user choice.

This touches the same Zod block as the `ui.set` schema fix I am opening alongside it (#10715). The two are independent in substance — that one is a pre-existing bug affecting `linearMode`/`linearContext`/`jira*` on paired clients — but whichever merges second will need a trivial rebase. This branch is cut from `main`, not from that one, so it can also merge first.

I do not use X, so no handle for the shoutout.

## ELI5

Linear filters and view options (status, layout, grouping) reset every time you left Tasks or relaunched the app. Those choices now persist across sessions so you don't rebuild your board setup every visit.

### DIFF
--- a/src/main/runtime/rpc/methods/client-ui-schemas.ts
+++ b/src/main/runtime/rpc/methods/client-ui-schemas.ts
@@ -13,6 +13,7 @@ import { isTaskProvider } from '../../../../shared/task-providers'
 import { normalizeDisabledTuiAgents } from '../../../../shared/tui-agent-selection'
 import { normalizePRBotAuthorOverrides } from '../../../../shared/pr-bot-author-overrides'
 import { normalizeWorktreeCardProperties } from '../../../../shared/worktree-card-properties'
+import { LinearIssueAttributeFilterSchema } from './linear-issue-attribute-filter-schema'
 import type { TaskProvider } from '../../../../shared/types'
 
 const NullableString = z.string().nullable()
@@ -66,7 +67,18 @@ const TaskResumeState = z
     githubItemsQuery: z.string().optional(),
     githubProjectHiddenFieldIdsByView: z.record(z.string(), z.array(z.string())).optional(),
     linearPreset: z.enum(['assigned', 'created', 'all', 'completed']).optional(),
-    linearQuery: z.string().optional()
+    linearQuery: z.string().optional(),
+    // Why: this schema is strict — a TaskResumeState field the renderer persists but
+    // that is missing here fails the whole ui.set call for paired web/mobile clients.
+    linearViewMode: z.enum(['list', 'board']).optional(),
+    linearGroupBy: z.enum(['none', 'status', 'assignee', 'priority', 'team']).optional(),
+    linearOrderBy: z.enum(['priority', 'updated', 'identifier']).optional(),
+    linearDisplayProperties: z
+      .array(z.enum(['state', 'priority', 'assignee', 'team', 'labels', 'updated']))
+      .optional(),
+    linearTeamPropertyTouched: z.boolean().optional(),
+    linearIssueFilter: LinearIssueAttributeFilterSchema.optional(),
+    linearIssueFilterWorkspaceId: z.string().optional()
   })
   .strict()
 const WorkspaceCleanupDismissal = z
--- a/src/main/runtime/rpc/methods/client-ui.test.ts
+++ b/src/main/runtime/rpc/methods/client-ui.test.ts
@@ -437,7 +437,19 @@ describe('client UI RPC methods', () => {
         githubItemsQuery: 'is:open',
         githubProjectHiddenFieldIdsByView: {
           'project-1:view-1': ['field-1']
-        }
+        },
+        linearViewMode: 'board',
+        linearGroupBy: 'assignee',
+        linearOrderBy: 'updated',
+        linearDisplayProperties: ['state', 'labels'],
+        linearTeamPropertyTouched: true,
+        linearIssueFilter: {
+          stateIds: ['state-1'],
+          priorities: [2],
+          assignee: { kind: 'user', id: 'user-1' },
+          labelIds: ['label-1']
+        },
+        linearIssueFilterWorkspaceId: 'workspace-1'
       },
       workspaceCleanup: {
         dismissals: {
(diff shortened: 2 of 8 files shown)


giraphant:
Already fixed in latest version.

PR #10735, opened 2026-07-26
fix(speech): download verified model artifacts directly

nwparker:
## Summary

Replace speech-model archive downloads and extraction with direct downloads of the runtime artifacts.

- adds no dependencies or build configuration
- avoids `tar` on Windows, macOS, and Linux
- pins every upstream file to an immutable Hugging Face revision, exact size, and SHA-256
- downloads into a per-model staging directory and atomically renames it only after every file verifies
- preserves download resume/retry, cancellation, proxy support, and cleanup of legacy `.tar.bz2` leftovers
- preserves `bpe.vocab` for bilingual-model hotword support

This is a dependency-free alternative to #9571 for #9231.

## Screenshots

No UI change.

## Testing

- [x] `pnpm typecheck`
- [x] all speech tests on macOS: 14 files, 82 tests
- [x] all speech tests on physical Windows through `orca --environment awin`: 14 files, 82 tests
- [x] Windows `pnpm typecheck`
- [x] changed-file `oxlint`
- [x] `pnpm check:max-lines-ratchet`
- [x] `git diff --check`
- [ ] `pnpm lint`
  - reaches the final localization audit and fails only on pre-existing Ghostty search keywords outside this change

Real-artifact checks:

- Windows 11 `bsdtar` extracted the pinned 487,170,055-byte Parakeet v3 archive in about 20 seconds; all extracted runtime file hashes match this catalog
- streamed release archives for SenseVoice and Bilingual Zipformer produced the same hashes as the pinned direct artifacts, including `bpe.vocab`

## AI Review Report

The download state machine still uses Electron `net`, so application proxy behavior is unchanged. Range resume, redirects, idle timeout, retry bounds, cancellation, progress aggregation across files, and cleanup have focused coverage.

Model files remain local application-cache data. SSH and folder workspaces do not execute or store these downloads, and no source-control or provider behavior changes.

The main tradeoff is compression: Parakeet, Whisper Tiny, and SenseVoice transfer more bytes individually, while models whose archives contain unused variants transfer less. Across the current catalog, direct downloads reduce the total possible transfer by 663,114,769 bytes.

## Security Audit

- only HTTPS URLs are accepted
- URLs use immutable 40-character repository revisions
- every file is checked against a pinned size and SHA-256 before installation
- catalog filenames cannot contain path separators or traversal components
- partially downloaded or failed models never become the live model directory
- no archive parser or system executable handles downloaded content

## Notes

No dependency or lockfile changes.

X: No account


LauraGPT (approved):
Reviewed the direct-download state machine and the SenseVoice catalog entry at 6c8e2000c. Locally reran the six affected speech test files: 38/38 passed. Independently verified the pinned SenseVoice model size/SHA against the immutable Hugging Face revision, recomputed the tokens.txt SHA-256, and confirmed the CDN range path. Resume, redirect, cancellation, per-file verification, staging cleanup, and final rename behavior look consistent. No blocking findings.

PR #10736, opened 2026-07-26
fix(persistence): publish async state saves without an await race

### LINKED ISSUE #10696: [Bug]: a stale async state write can clobber a newer shutdown flush of orca-data.json
## Summary

`Store.writeToDiskAsync()` checks its generation guard *before* awaiting the rename, so the guard and the publication aren't atomic. A synchronous `flushOrThrow()` can publish newer state while the async rename is still in flight; the stale rename then lands last and wins.

Found during review of #10631 (CodeRabbit flagged it). **Not introduced by that PR** — this is the current behaviour on `main`.

## Reproduction

383/400 stale wins on macOS/APFS:

```js
import { mkdtempSync, readFileSync, renameSync, writeFileSync } from 'node:fs'
import { rename, writeFile } from 'node:fs/promises'

// Models: async write clears its generation guard, then a sync flush
// publishes newer state while the async rename is in flight.
const inFlight = rename(oldTmp, target)  // async path, guard already passed
renameSync(newTmp, target)               // sync flush publishes newer state
await inFlight                           // stale rename lands last
// -> stale wins: 383/400
```

## The code

`src/main/persistence.ts:3682-3690`:

```ts
// Why: if flush() bumped writeGeneration mid-write, it already wrote fresher state; don't overwrite it.
if (this.writeGeneration !== gen) {
  return
}
await renameDurable(tmpFile, dataFile)
renamed = true
// Why re-check gen: a sync flush during the rename await may have written fresher state; don't record a stale hash over it.
if (this.writeGeneration === gen) {
  this.lastWrittenStateHash = stateHash
}
```

The re-check after the rename is aware of the hazard but only guards the hash/backup bookkeeping — by then the file has already been overwritten and the newer state is gone.

## Impact

`flushOrThrow()` runs at shutdown (`index.ts:2527`, via `store?.flush()`), which is exactly when an async debounced write is most likely in flight. The user-visible effect is losing the last increment of state — tabs, layouts, session bindings — on a quit that appeared clean. It also silently defeats the `force: asyncWriteWasInFlight` intent in `flushOrThrow()`.

Low frequency (needs the interleaving), but the failure is silent and the data is unrecoverable short of the `.bak` ring.

## Prior art in this repo

`src/main/active-view-preference.ts:91-106` already solves this for the sibling writer by keeping the guard and the swap in one synchronous block:

```ts
// Why: keep the generation guard and the swap synchronous (no await between),
// or a concurrent flushOrThrow could rename a newer view that this stale write
// then clobbers, restoring the prior view on next launch.
if (generation !== this.writeGeneration) {
  return
}
renameSync(tmpFile, this.file)
```

## The tradeoff

Copying that pattern means a synchronous rename of a multi-MB file on the main thread — which is what `persistence.ts:3655` says the async path exists to avoid ("async writes avoid blocking the main Electron thread on every debounced save"). A rename is a metadata operation so it's cheap, but it isn't free, and `orca-data.json` is much larger than the active-view file.

Options:

1. **Synchronous guard + `renameSync`** — matches `active-view-preference.ts`; needs measurement on a realistic payload.
2. **Serialize publication** — have the sync flush await/cancel the in-flight rename rather than racing it.
3. **Publication token** — write the generation into the temp filename and have the loser skip its own rename.

Whichever is chosen should land with a forced-interleaving regression test, since this cannot be caught by chance.

## Note

Directory-fsync durability for these writes is handled separately in #10631; this issue is only about publication ordering.

Souptik96:
## Summary

Fixes #10696. A stale background state save could silently revert the shutdown flush of `orca-data.json`, losing whatever the user did just before quitting.

`Store.writeToDiskAsync()` checked the write-generation guard and then published with `await renameDurable(tmpFile, dataFile)`. An `await` between the guard and the rename is not atomic: a synchronous `flushOrThrow()` (quit / durable-close paths) could bump the generation and publish fresher state *during* that await, and the already-past-the-guard stale rename then landed last and overwrote it. Worse, the stale write left `lastWrittenStateHash` pointing at the flushed state, so the hash guard considered the newer state already persisted and nothing ever rewrote it — the loss was invisible until the next launch.

The publication is now a synchronous rename, so the guard and the swap cannot interleave. Only the rename and the directory fsync are synchronous; serializing state and writing the multi-MB temp file stay async, so the main thread still does not block on the expensive part of a save.

This is deliberately the same shape `src/main/active-view-preference.ts` (~L97-103) already uses, whose comment describes this exact hazard: *"keep the generation guard and the swap synchronous (no await between), or a concurrent flushOrThrow could rename a newer view that this stale write then clobbers"*.

**Note on the fix choice.** The issue listed three candidate fixes and did not pick one. I went with option 1 (sync guard + sync rename) because the other two do not actually close the window: whatever decides "am I the stale writer" must still be evaluated in the same synchronous tick as the rename, so a publication token or a serialization handshake both reduce to this guard plus extra machinery. Option 1 is also the only one with existing precedent in the codebase. See Notes for the cost tradeoff.

Changes:
- `src/main/durable-file-write.ts` — add `renameDurableSync()`, the synchronous counterpart of `renameDurable()` (`renameSync` + `syncDirectorySync`), so publication keeps the directory-fsync durability that `renameDurable` provided. A bare `renameSync` would have silently dropped it.
- `src/main/persistence.ts` — guard and publish in one synchronous block; record `lastWrittenStateHash` unconditionally, since nothing can run between the guard and the rename any more (the previous re-check existed only because of the await).
- `src/main/persistence-async-write-clobber.test.ts` — new regression cover.
- `src/main/durable-file-write.test.ts` — unit cover for `renameDurableSync` (and `renameDurable`, which had none).

## Screenshots

No visual change.

## Testing

Environment caveat up front, because it limits what I can honestly claim: this was verified on **Windows 11, Node v20.18.1, pnpm 9.15.9**, while the repo requires **Node 24** (`engines`) and **pnpm 10.24.0** (`packageManager`). Consequences:

- `postinstall` (`rebuild-native-deps.mjs`) fails on Node 20 — it imports `fs.globSync`, which needs Node 22+. So `node-pty` was never rebuilt, and `pnpm test`'s `ensure-native-runtime.mjs` prerequisite cannot pass here. I ran `vitest` directly instead.
- 53 tests in `persistence.test.ts` / `persistence-single-serialize.test.ts` fail on this box **before any of my changes**, all from `TypeError: Map.groupBy is not a function` (Node 21+ API). Recorded as a baseline rather than attributed to this PR.
- `vitest` needed `NODE_OPTIONS=--experimental-require-module` to load `config/vitest.config.ts` on Node 20 (`require()` of ESM `std-env`).

**Baseline vs. after, same 3 files, same box** (`vitest run --config config/vitest.config.ts src/main/durable-file-write.test.ts src/main/persistence-single-serialize.test.ts src/main/persistence.test.ts`):

| | Test files | Tests |
|---|---|---|
| pristine `upstream/main` (f3d8edb) | 2 failed / 1 passed | **53 failed / 385 passed** (438) |
| this branch (+ the new test file) | 2 failed / 2 passed | **53 failed / 391 passed** (444) |

Identical failure count; `+6` passing = exactly the 6 tests added here. No new failures.

Targeted run of the touched suites — green:

```
$ node node_modules/vitest/vitest.mjs run --config config/vitest.config.ts \
    src/main/persistence-async-write-clobber.test.ts src/main/durable-file-write.test.ts
 ✓ src/main/durable-file-write.test.ts (16 tests) 273ms
 ✓ src/main/persistence-async-write-clobber.test.ts (3 tests) 2462ms
 Test Files  2 passed (2)
      Tests  19 passed (19)
```

**Negative control** — I reverted only the `persistence.ts` hunk (back to `await renameDurable(...)`, keeping the new helper so the failure is attributable to the await, not a missing export) and re-ran the new file. All 3 tests fail, and the key assertion reproduces the reported data loss exactly:

```
FAIL src/main/persistence-async-write-clobber.test.ts > does not let a stale async save
     clobber a newer sync shutdown flush
AssertionError: expected 400 to be 500 // Object.is equality
- Expected:  500     <- the shutdown flush's state
+ Received:  400     <- the stale async rename landed last
```

Other gates:

- [x] `pnpm lint` — **partially**. `oxlint` → exit 0 (`Found 27 warnings and 0 errors`; all 27 pre-existing, none in files I touched). `lint:switch-exhaustiveness`, `check-styled-scrollbars`, `check:reliability-gates`, `check:max-lines-ratchet`, `verify:bundled-skill-guides`, `verify:localization-catalog` → all exit 0. `verify:skill-bundle-manifest` and `verify:localization-coverage` → exit 1, **but they fail identically on pristine `upstream/main` on this box** (stale generated `resources/skills/*.json`; `Ghostty` keyword coverage in renderer settings files). Neither touches anything in this PR; I did not regenerate artifacts I have no business regenerating.
- [x] `pnpm typecheck` — clean, all three projects: `tsconfig.node.json`, `tsconfig.tc.cli.json`, `tsconfig.tc.web.json` each exit 0.
- [ ] `pnpm test` — **could not run as scripted** (blocked by `ensure-native-runtime.mjs` / Node 20, above). Ran `vitest` directly with the baseline comparison instead.
- [ ] `pnpm build` — **not run.** It chains `build:native` and Electron packaging; no MSVC toolchain and no Node 24 on this non-admin box. `pnpm typecheck` (the first stage of `build:desktop`) does pass.
- [x] Added tests that would catch the regression — proven by the negative control above.
- [x] `oxfmt --check` on all four touched files → `All matched files use the correct format.`

I'd ask a maintainer to confirm CI green on Node 24, since `pnpm test` and `pnpm build` are genuinely unverified here.

## AI Review Report

Written and reviewed with an AI coding agent (Kiro CLI). This PR is AI-authored; the analysis below is the agent's own review of its change.

Risks checked:

- **Does the fix actually close the window?** Yes, and this was the main thing reviewed. Under a single-threaded event loop, removing every await between the generation read and the `rename` makes check-and-publish atomic with respect to `flushOrThrow()`, which is synchronous. The remaining awaits (`mkdir`, `open`, `writeFile`, `handle.sync`, `close`) all sit *before* the guard, where a competing flush is correctly caught by it.
- **Was the issue's diagnosis correct?** Verified against real code rather than taken on trust. The guard/rename pair reported at L3682-3690 now sits at L3699-3702 (renumbered by #10631); same defect. The agent also found a second-order effect the issue did not mention: the stale rename leaves `lastWrittenStateHash` describing the *flushed* state, so the hash short-circuit suppresses any corrective rewrite. That is why the loss is silent.
- **Unconditional hash record.** Safe only because no interleaving point remains between guard and rename; called out in a comment so a future await cannot be added there innocently.
- **Durability not weakened.** `renameDurableSync` keeps the directory fsync. Using a bare `renameSync` here would have quietly dropped the "rename recorded but not persisted" protection this module exists to provide.
- **Temp-file cleanup / no orphans.** The `renamed` flag and `finally` are untouched; one test asserts no `*.tmp` files survive either race outcome (a leaked temp file here would be multi-MB).
- **Cross-platform (macOS / Linux / Windows) — explicitly reviewed.** No platform branches, no path construction, no shell, no shortcut or label changes. `renameSync` replaces an existing destination on all three platforms (Windows via `MoveFileEx` + `MOVEFILE_REPLACE_EXISTING`), and `writeFileDurableSync` in this same module already relies on that from `flushOrThrow()`, so this PR adds no new platform assumption. `syncDirectorySync` already handles the Windows case where a directory cannot be opened for fsync — it is a documented best-effort no-op there. Tests were actually executed on Windows; macOS and Linux are unverified by me and rely on CI.
- **Remote / SSH / local.** `orca-data.json` is local main-process state; no remote, SSH, or worktree path is involved.
- **Agent / integration / git-provider neutrality.** Nothing provider-specific touched.
- **Performance.** Deliberate and bounded: see Notes.

Flagged and resolved during review: an earlier draft of the third test reset only one of two call counters, which made it fail pre-fix for a bookkeeping reason rather than a behavioral one. Rewritten so it tests one honest thing (a flush that wins outright leaves no orphan temp file and wins), and it passes both pre- and post-fix — only the first two tests are regression tests for this bug.

## Security Audit

- **Input handling** — none. No new external, IPC, or user input is parsed; the payload is the same serialized state as before.
- **Command execution** — none added.
- **Path handling** — unchanged. The temp path and final path are still built exactly as before; no new user-influenced path segment, no traversal surface. `dirname()` is applied to a path the process already owns.
- **Secrets** — `buildStateToSave()` is untouched, so secret encryption and the plaintext/ciphertext split are unchanged. The fix operates strictly after the payload exists. Worth noting as a security-adjacent *improvement*: the bug could silently revert a shutdown flush, which includes flushes of the Codex reset-credit ledger (`replaceCodexResetCreditAttemptLedgerAndFlush()` treats a successful return as a durability barrier) — a stale clobber could roll that ledger back after a caller was told it was durable.
- **Auth / IPC** — no auth, permission, or IPC surface touched.
- **Dependencies** — none added, removed, or bumped. `pnpm-lock.yaml` is deliberately unchanged (see Notes).
- **Denial of service / resource exhaustion** — temp-file cleanup semantics unchanged; the added test pins that no multi-MB orphan is left behind.
- **Follow-up needed** — none identified.

## Notes

- **Performance tradeoff, stated plainly.** The rename and the directory fsync now run on the main thread. A rename within one filesystem is a metadata operation whose cost does not scale with file size, so the multi-MB concern raised in the issue does not apply to it; the payload write and its fsync remain async. The directory fsync is a real, if small, added main-thread syscall on the debounced save path (at most once per `SAVE_DEBOUNCE_MS`). I kept it because dropping it would be a durability regression, and because `writeFileDurableSync()` — already called from `flushOrThrow()` on the main thread — does the same thing. If you would rather keep the main thread completely clear, the natural follow-up is to keep the rename synchronous but move only the directory fsync back to `await`: publication ordering is already decided by the sync rename, so the fsync no longer needs to be inside the guarded block. Happy to switch if you prefer that.
- `renameDurable()` (async) now has no production caller, but it is still exercised by `durable-file-write.test.ts`, where I added coverage it previously lacked plus an equivalence check against the sync variant. I left the export in place rather than unilaterally deleting a durability helper — say the word and I'll remove it.
- `pnpm-lock.yaml` is intentionally untouched. My local install had to run with `--no-frozen-lockfile` because pnpm 9 cannot read pnpm 10's `patchedDependencies` hash format, which rewrote the lockfile; I restored it from `HEAD` so this PR contains source changes only.
- Also worth flagging while in this code: `flushActiveViewPreferenceOrThrow()` and the stats collector use the same guard-then-publish pattern. `active-view-preference.ts` is already correct (synchronous swap). I did not audit every other async writer in the tree — happy to open a follow-up if you'd like a sweep.
- Platform testing notes: executed on Windows 11 only. macOS and Linux behavior is unverified by me.

X handle: I don't have one.

## ELI5

A background save of app state could overwrite a fresher shutdown save and drop what you did just before quitting. Async writes no longer race the quit flush, so last-second changes stick.

### DIFF
--- a/src/main/durable-file-write.test.ts
+++ b/src/main/durable-file-write.test.ts
@@ -2,7 +2,12 @@ import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, beforeEach, describe, expect, it } from 'vitest'
-import { writeFileDurable, writeFileDurableSync } from './durable-file-write'
+import {
+  renameDurable,
+  renameDurableSync,
+  writeFileDurable,
+  writeFileDurableSync
+} from './durable-file-write'
 
 describe('durable file write', () => {
   let dir: string
@@ -83,5 +88,37 @@ describe('durable file write', () => {
     await writeFileDurable(`${final}.a.tmp`, final, 'from-async')
     writeFileDurableSync(`${final}.b.tmp`, final, 'from-sync')
     expect(readFileSync(final, 'utf-8')).toBe('from-sync')
+  })
+
+  describe('renameDurableSync', () => {
+    it('publishes an already-written temp file at the final path', () => {
+      const final = join(dir, 'state.json')
+      const tmp = `${final}.tmp`
+      writeFileSync(tmp, 'fresh', 'utf-8')
+      renameDurableSync(tmp, final)
+      expect(readFileSync(final, 'utf-8')).toBe('fresh')
+      expect(() => readFileSync(tmp, 'utf-8')).toThrow()
+    })
+
+    it('replaces existing content and matches the async helper', async () => {
+      const syncTarget = join(dir, 'sync.json')
+      const asyncTarget = join(dir, 'async.json')
+      writeFileSync(syncTarget, 'stale', 'utf-8')
+      writeFileSync(asyncTarget, 'stale', 'utf-8')
+      writeFileSync(`${syncTarget}.tmp`, 'fresh', 'utf-8')
+      writeFileSync(`${asyncTarget}.tmp`, 'fresh', 'utf-8')
+
+      renameDurableSync(`${syncTarget}.tmp`, syncTarget)
+      await renameDurable(`${asyncTarget}.tmp`, asyncTarget)
+
+      expect(readFileSync(syncTarget, 'utf-8')).toBe('fresh')
+      expect(readFileSync(asyncTarget, 'utf-8')).toBe(readFileSync(syncTarget, 'utf-8'))
+    })
+
+    it('throws instead of silently failing when the temp file is missing', () => {
+      const final = join(dir, 'state.json')
+      expect(() => renameDurableSync(join(dir, 'nope.tmp'), final)).toThrow()
+      expect(() => readFileSync(final, 'utf-8')).toThrow()
+    })
   })
 })
--- a/src/main/durable-file-write.ts
+++ b/src/main/durable-file-write.ts
@@ -51,6 +51,17 @@ export async function renameDurable(tmpPath: string, finalPath: string): Promise
   await syncDirectory(dirname(finalPath))
 }
 
+/**
+ * Synchronous counterpart to renameDurable. For callers that already fsynced the temp file
+ * asynchronously but must publish it without an await — a generation guard followed by an awaited
+ * rename is not atomic, so a competing synchronous write can publish fresher state in between and
+ * then be clobbered by the stale rename.
+ */
+export function renameDurableSync(tmpPath: string, finalPath: string): void {
+  renameSync(tmpPath, finalPath)
+  syncDirectorySync(dirname(finalPath))
+}
+
 /** Write `payload` to `tmpPath`, fsync it, then rename onto `finalPath` and fsync the directory. */
 export async function writeFileDurable(
   tmpPath: string,
--- a/src/main/persistence-async-write-clobber.test.ts
+++ b/src/main/persistence-async-write-clobber.test.ts
@@ -1,0 +1,187 @@
+// Regression cover for issue #10696: the debounced async save published state with
+// `await renameDurable(...)` AFTER checking the write-generation guard. An await between the
+// guard and the rename is not atomic, so a synchronous flushOrThrow() (quit / shutdown path)
+// could publish fresher state during that await and then be silently clobbered when the stale
+// rename landed last — losing every mutation the shutdown flush was there to save.
+//
+// The publication is now synchronous (renameDurableSync), which is the same shape
+// active-view-preference.ts already uses for this exact hazard.
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { readdirSync, readFileSync, rmSync, mkdtempSync } from 'node:fs'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+import type * as DurableFileWrite from './durable-file-write'
+
+const testState = { dir: '' }
+
+// Counts which publication path the async save actually used, and lets a test suspend the
+// async rename to recreate the pre-fix interleaving window.
+const renameCalls = { async: 0, sync: 0 }
+const renameHooks: {
+  enteredAsyncRename: boolean
+  onEnterAsyncRename: (() => void) | null
+  gate: Promise<void> | null
+} = { enteredAsyncRename: false, onEnterAsyncRename: null, gate: null }
+
+vi.mock('./durable-file-write', async (importOriginal) => {
+  const actual = await importOriginal<typeof DurableFileWrite>()
+  return {
+    ...actual,
+    renameDurable: async (tmpPath: string, finalPath: string) => {
+      renameCalls.async++
+      renameHooks.enteredAsyncRename = true
+      renameHooks.onEnterAsyncRename?.()
+      if (renameHooks.gate) {
+        await renameHooks.gate
+      }
+      await actual.renameDurable(tmpPath, finalPath)
+    },
+    renameDurableSync: (tmpPath: string, finalPath: string) => {
+      renameCalls.sync++
+      actual.renameDurableSync(tmpPath, finalPath)
+    }
+  }
+})
+
+vi.mock('./ssh/ssh-config-parser', () => ({
+  loadUserSshConfig: vi.fn(),
+  sshConfigHostsToTargets: vi.fn()
+}))
+
+vi.mock('electron', () => ({
+  app: {
+    getPath: () => testState.dir
+  },
+  safeStorage: {
+    isEncryptionAvailable: () => false,
+    encryptString: (plaintext: string) => Buffer.from(plaintext, 'utf-8'),
+    decryptString: (ciphertext: Buffer) => ciphertext.toString('utf-8')
+  }
+}))
+
+vi.mock('./telemetry/client', () => ({
+  track: vi.fn()
+}))
+
+vi.mock('./telemetry/cohort-classifier', () => ({
+  getCohortAtEmit: vi.fn().mockReturnValue({ nth_repo_added: 2 })
+}))
+
+async function createStore() {
+  vi.resetModules()
+  const { Store, initDataPath } = await import('./persistence')
+  initDataPath()
+  return new Store()
+}
+
+function dataFile(): string {
+  return join(testState.dir, 'orca-data.json')
+}
+
+function persistedSidebarWidth(): number {
+  const parsed = JSON.parse(readFileSync(dataFile(), 'utf-8')) as {
+    ui: { sidebarWidth: number }
+  }
+  return parsed.ui.sidebarWidth
+}
+
+function leftoverTmpFiles(): string[] {
+  return readdirSync(testState.dir).filter((name) => name.endsWith('.tmp'))
+}
+
+describe('persistence async write vs sync shutdown flush', () => {
+  beforeEach(() => {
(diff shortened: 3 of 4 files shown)


Souptik96:
Follow-up: I got a supported Node onto this box (portable Node **v24.18.0**, matching `engines.node`), so the Node-20 caveats in the description are now resolved for the test gate. Re-ran everything against the real runtime version:

**With this branch** (`vitest run --config config/vitest.config.ts src/main/persistence-async-write-clobber.test.ts src/main/durable-file-write.test.ts src/main/persistence-single-serialize.test.ts src/main/persistence.test.ts`):

```
 Test Files  4 passed (4)
      Tests  444 passed (444)
```

**Pristine `upstream/main` baseline**, same 3 pre-existing files (new test file held out):

```
 Test Files  3 passed (3)
      Tests  438 passed (438)
```

So: `438 → 444`, i.e. exactly the 6 tests added here, with **zero failures on either side**. The 53 failures reported in the description were entirely `Map.groupBy` (Node 21+) on my old Node 20 and are gone — please disregard that caveat.

Also re-confirmed on Node 24: `oxlint` exit 0 (27 pre-existing warnings, none in touched files), all three `tsc` projects exit 0, `oxfmt --check` clean, and `lint:switch-exhaustiveness` / `check-styled-scrollbars` / `check:reliability-gates` / `check:max-lines-ratchet` / `verify:bundled-skill-guides` / `verify:localization-catalog` all exit 0. `verify:skill-bundle-manifest` and `verify:localization-coverage` still fail identically on pristine `upstream/main`, unrelated to this change.

Still unverified by me: `pnpm build` (needs the native/Electron packaging toolchain I don't have) and macOS/Linux behavior.

PR #10739, opened 2026-07-26
fix(claude-usage): price Claude Opus 5, Fable 5, and Sonnet 5 usage

MadKangYu:
## Summary

Claude usage on the 5-series models was silently dropped from the cost breakdown.

`MODEL_PRICING` and `normalizeModelForPricing` had no entry or branch for `claude-opus-5`, `claude-fable-5`, or `claude-sonnet-5`, so `estimateCostUsd` returned `null` for any usage on those models. Opus 5 shipped 2026-07-24 and is now the default model on Claude Max, so this affects current usage, not a hypothetical future model.

The existing future-proofing did not catch it. `store.ts` already has a fallback that maps unknown `opus-4` point releases onto current Opus pricing, with a comment explaining that the goal is to avoid silently mis-costing unknown Claude Code model IDs. That guard is keyed on the literal `opus-4` family string, so it stopped applying the moment the family name became `opus-5`.

Changes:

- `store.ts`: add `claude-opus-5`, `claude-fable-5`, `claude-sonnet-5` to `MODEL_PRICING`, plus their `-thinking` aliases and `hasClaudeModelVersion` branches in `normalizeModelForPricing`.
- `commit-message-agent-spec.ts`: extend the static Copilot catalog with the 5-series entries so users can select the models their org has enabled. The catalog is intentionally static (per the existing comment, Copilot's picker is policy-filtered per account/org), so new hosted models have to be added by hand.

All rates verified against https://platform.claude.com/docs/en/about-claude/pricing on 2026-07-26:

| Model | Input | Output | Cache read | Cache write (5m) |
| --- | --- | --- | --- | --- |
| Claude Opus 5 | $5 | $25 | $0.50 | $6.25 |
| Claude Fable 5 | $10 | $50 | $1 | $12.50 |
| Claude Sonnet 5 | $2 | $10 | $0.20 | $2.50 |

## Screenshots

No visual change. This only affects the cost figures already rendered in the usage breakdown.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

`pnpm typecheck` passes clean.

`pnpm lint` and `pnpm test` fail on this branch, but both failures reproduce identically on a pristine `origin/main` checkout with this branch's commit absent, and neither touches the files in this PR:

- `pnpm lint` fails at `verify:localization-coverage` on unlocalized `Ghostty` keyword strings in `terminal-advanced-platform-search.ts` and `terminal-pane-appearance-search.ts`. Confirmed `node config/scripts/audit-localization-coverage.mjs --check` exits 1 on pristine `origin/main`.
- `pnpm test` reports 20 failing files out of 3437 (3410 passing), concentrated in socket, PTY, and git-integration suites that appear environment-sensitive. Spot-checking four of them on pristine `origin/main` reproduced two of the same failures.

The four test files this PR touches or covers all pass: 95 tests green.

`pnpm build` passes clean, exit 0, roughly 93 seconds, including the native and computer-use-macos steps.

New test coverage in `store.test.ts` asserts the pricing rows and the normalization branches, including a `claude-opus-5-20260724`-style dated model ID, so a future family rename regression would be caught.

## AI Review Report

- **Rate accuracy.** Every added rate was cross-checked line by line against the official Anthropic pricing table rather than copied from memory or a secondary source. Cache-write values use the 5-minute multiplier, consistent with the surrounding rows.
- **Normalization ordering.** The new `fable`/`opus`/`sonnet` version-5 branches are placed ahead of the `opus-4` and `sonnet-4` branches, and the review confirmed no earlier substring branch shadows them. `claude-fable-5` is matched before the opus branches so it cannot be absorbed by a looser match.
- **Cross-platform.** No platform-specific behavior. This is a pure data table plus string normalization in main-process code, with no path handling, shell invocation, shortcut, or label surface. Nothing here differs on macOS, Linux, or Windows.
- **Provider neutrality.** The Copilot catalog addition stays inside the existing Copilot-specific block and does not change behavior for other agents.

Flagged and deliberately not changed in this PR, listed under Notes below.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, dependency, or IPC surface. The change adds constant numeric rates to an existing in-memory table and adds string-comparison branches to an existing pure function that already receives the same untrusted model string. No new external calls and no new dependencies.

## Notes

One follow-up left out to keep this PR scoped to the reported bug. Happy to open a separate PR if you want it.

1. ~~**Sonnet 5 introductory pricing expires.**~~ Resolved in 02b657c after review feedback. Pricing now resolves per usage day, and `buildBreakdown` accumulates cost per daily aggregate so a model row spanning 2026-08-31 to 2026-09-01 is priced correctly on both sides rather than approximated with one representative date.

2. **The family fallback is still version-specific.** `lower.includes('opus-4')` mapping to current Opus pricing is exactly the guard that failed here once the family became `opus-5`. A version-agnostic form would prevent the next occurrence, but it involves assuming pricing for unreleased models, so it seemed like a maintainer call rather than something to fold into a data fix.

Also worth noting: pricing is not machine-readable from any Anthropic API. The Models API returns `max_input_tokens`, `max_tokens`, and `capabilities` but no rates, so this table cannot be automated and will need manual updates on each model launch.

## ELI5

Usage for Claude Opus 5, Fable 5, and Sonnet 5 was ignored in cost estimates because pricing tables had no entries. Those models are priced correctly so usage breakdowns include real spend.

### DIFF
--- a/src/main/claude-usage/store.test.ts
+++ b/src/main/claude-usage/store.test.ts
@@ -317,6 +317,95 @@ describe('ClaudeUsageStore', () => {
     ).toBeCloseTo(14.7)
   })
 
+  it('prices Claude Sonnet 5 usage before 2026-09-01 at introductory rates', async () => {
+    const store = createStoreWithState({
+      dailyAggregates: [
+        {
+          day: '2026-08-31',
+          model: 'claude-sonnet-5',
+          projectKey: 'worktree:repo-1::/workspace/repo-a',
+          projectLabel: 'Repo A',
+          repoId: 'repo-1',
+          worktreeId: 'repo-1::/workspace/repo-a',
+          turnCount: 1,
+          zeroCacheReadTurnCount: 0,
+          inputTokens: 1_000_000,
+          outputTokens: 1_000_000,
+          cacheReadTokens: 1_000_000,
+          cacheWriteTokens: 1_000_000
+        }
+      ]
+    })
+
+    const breakdown = await store.getBreakdown('orca', 'all', 'model')
+
+    expect(breakdown[0]?.estimatedCostUsd).toBeCloseTo(14.7)
+  })
+
+  it('prices Claude Sonnet 5 usage from 2026-09-01 at standard rates', async () => {
+    const store = createStoreWithState({
+      dailyAggregates: [
+        {
+          day: '2026-09-01',
+          model: 'claude-sonnet-5',
+          projectKey: 'worktree:repo-1::/workspace/repo-a',
+          projectLabel: 'Repo A',
+          repoId: 'repo-1',
+          worktreeId: 'repo-1::/workspace/repo-a',
+          turnCount: 1,
+          zeroCacheReadTurnCount: 0,
+          inputTokens: 1_000_000,
+          outputTokens: 1_000_000,
+          cacheReadTokens: 1_000_000,
+          cacheWriteTokens: 1_000_000
+        }
+      ]
+    })
+
+    const breakdown = await store.getBreakdown('orca', 'all', 'model')
+
+    expect(breakdown[0]?.estimatedCostUsd).toBeCloseTo(22.05)
+  })
+
+  it('sums Claude Sonnet 5 usage across introductory and standard pricing days', async () => {
+    const store = createStoreWithState({
+      dailyAggregates: [
+        {
+          day: '2026-08-31',
+          model: 'claude-sonnet-5',
+          projectKey: 'worktree:repo-1::/workspace/repo-a',
+          projectLabel: 'Repo A',
+          repoId: 'repo-1',
+          worktreeId: 'repo-1::/workspace/repo-a',
+          turnCount: 1,
+          zeroCacheReadTurnCount: 0,
+          inputTokens: 1_000_000,
+          outputTokens: 1_000_000,
+          cacheReadTokens: 1_000_000,
+          cacheWriteTokens: 1_000_000
+        },
+        {
+          day: '2026-09-01',
+          model: 'claude-sonnet-5',
+          projectKey: 'worktree:repo-1::/workspace/repo-a',
+          projectLabel: 'Repo A',
+          repoId: 'repo-1',
+          worktreeId: 'repo-1::/workspace/repo-a',
+          turnCount: 1,
+          zeroCacheReadTurnCount: 0,
+          inputTokens: 1_000_000,
+          outputTokens: 1_000_000,
+          cacheReadTokens: 1_000_000,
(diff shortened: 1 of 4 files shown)


MadKangYu:
Good catch, and the original scoping note was wrong. I said a date-correct fix would require threading a timestamp through the pricing path, but I only checked `estimateCostUsd`'s signature and never checked whether the callers already had a date. They did: daily aggregates carry `day`, and the session path has `lastTimestamp`.

Fixed in 02b657c.

- `resolvePricingForDay(model, day)` resolves rates per usage day, with `SONNET_5_STANDARD_PRICING_START_DAY = '2026-09-01'`.
- `estimateCostUsd` takes an optional `day`; callers that already know the usage day now pass it.
- `buildBreakdown` accumulates cost per daily aggregate instead of pricing the summed totals once. This is the part that actually mattered: a model row spans many days, so pricing the aggregate with a single representative date would still be wrong for any row crossing the cutover. Pricing per day and summing avoids the approximation entirely.
- Null cost semantics preserved. A model row stays `null` unless at least one daily aggregate produced a billable cost, and project rows stay `null`.
- Incidental cleanup: extracted `formatLocalDay` and reused it in `getRangeCutoff` and `getLocalDay`, which were duplicating the same formatting.

Three deterministic tests added, all passing day explicitly rather than depending on the system clock: pre-cutover at introductory rates, post-cutover at standard rates, and a single model row spanning 2026-08-31 and 2026-09-01 that must sum both rates.

`store.test.ts` 13/13 pass and `pnpm typecheck` exits 0.

OrcaWin:
Thank you for this — **your diagnosis was completely correct**, and I'm closing only because the same fix has since landed on main.

I verified the bug independently before checking: `MODEL_PRICING` and `normalizeModelForPricing` had no branch for `claude-opus-5`, `claude-fable-5`, or `claude-sonnet-5`, so `estimateCostUsd` returned `null` and all 5-series usage was silently dropped from the cost breakdown. Your observation that the existing future-proofing didn't catch it was also right — the unknown-model fallback is keyed on the literal `opus-4` string, so it could never match a 5-series id.

Claude 5 pricing landed in `cf513addd` (#10822) with the same rates you proposed. Current main has the entries at `src/main/claude-usage/store.ts:62-66` and the normalizer branches at `:165-194`, so `git diff origin/main` on this branch is now effectively empty.

Sorry you lost the race on this one — the analysis was sound, especially catching that Opus 5 shipped as the Max default and that this was affecting live usage rather than a hypothetical future model. Please do send more PRs.

PR #10741, opened 2026-07-26
perf(terminal): bound the PR-link carry scan to the trailing window

nwparker:
## Summary

`getPotentialGitHubPRCarry()` ran `lastIndexOf` for **both** http scheme prefixes across the **entire** combined chunk on every PTY write — including the early-out path where the chunk provably contains no `/pull/`. The carry it returns is always a suffix of at most `MAX_CARRY_LENGTH` (512) bytes, so every byte scanned before `length - 512` was work that could not change the result.

This bounds the scan to that trailing window. When the window holds no scheme but ends in a partial scheme fragment, one bounded probe behind the window preserves the existing "already overran the cap" drop.

This runs on the hottest path in the app: every PTY output chunk, via `pty-connection.ts` and `parked-terminal-byte-watcher.ts`.

## ELI5

The terminal watches your agent's output for GitHub PR links so it can make them clickable. Links can get split across two chunks of output, so after each chunk we keep a few leftover characters in case the next chunk finishes a link.

We only ever keep **at most 512 characters**. But to decide what to keep, the old code re-read the *entire* chunk from the start — up to a megabyte — just to produce those 512 characters. It's like re-reading a whole book to write down the last sentence.

Now it only looks at the last 512 characters. Same answer, far less reading.

## Screenshots

No visual change.

`node config/scripts/terminal-pr-link-carry-benchmark.mjs` (new, follows the existing `config/scripts/*-benchmark.mjs` house pattern):

```
    chunk   carry only   detector path   fragment tail
    4 KiB         5.8x            5.1x           1.01x
   16 KiB        20.8x           13.7x           0.97x
   64 KiB        84.3x           24.9x           1.00x
  256 KiB       352.2x           34.3x           1.02x
 1024 KiB      1312.6x           35.5x           1.00x
```

Three columns deliberately, so the number is not oversold:

- **carry only** — the scan this change bounds, in isolation. O(chunk) → O(1).
- **detector path** — `includes()` + carry, i.e. **what the hot path actually saves**. The detector always scans the full chunk for `/pull/` and this change does not touch that, so ~5-38x is the honest end-to-end figure.
- **fragment tail** — a chunk ending mid-scheme, the worst case where the new code pays an extra bounded probe. ~1.0x confirms the fallback costs nothing material.

An earlier revision of this benchmark reported up to 1516x by timing only the carry, and its equality guard was vacuous (every fixture returned `''` from both versions, so even `carryAfter = () => ''` passed). Both were caught in review and fixed; the guard now uses branch-covering non-empty fixtures and throws on a broken implementation.

## Proof of zero regression

This is a pure optimization, so **identical output for every input** is the whole safety argument:

- **2,612,138,803 inputs, 0 mismatches** — exhaustive over a scaled-down model (cap 8, all strings ≤ 12 chars over a 6-symbol alphabet) that preserves the structural cases, including overlapping scheme prefixes.
- **225,204 inputs, 0 mismatches** — targeted + fuzzed at full scale: every offset around the 512 boundary, whitespace in the tail, partial scheme fragments, adversarial alphabets.
- **546 straddle cases, 0 mismatches** — a scheme partially inside the window, the case reviewers flagged as most dangerous.

Tests: 3 new boundary cases pinning both sides of the cap and the drop arm, plus 2 covering `http://` (a reviewer found the `http://` half of the prefix loop was unpinned).

**Mutation testing — 7/7 mutants killed** (run with `--no-cache`; vitest's cache otherwise reports false survivals):

| mutant | result |
|---|---|
| delete behind-window probe (`return fragment`) | 3 tests fail |
| window 512 → 256 | 1 fails |
| window 512 → 511 | 1 fails |
| window 512 → 514 | 1 fails |
| prefix loop breaks after `https://` | 2 fail |
| behind-window probe checks only `https://` | 1 fails |
| invert final ternary | 4 fail |

- [x] `pnpm lint` (oxlint clean on changed files)
- [x] `pnpm typecheck`
- [x] `pnpm test` — 26/26 in the touched file; full suite failures are a **subset of the pre-existing set** (see below)

`pnpm test` is **already red on `main`**. I captured the baseline on a tree identical to `origin/main` before making any change:

- **Deterministic (every run):** `persistence.test.ts › derives fresh default profiles without branch`; `agent-exec-handler.test.ts` ×2 (assert against `process.env`, so local env leaks in).
- **Load-flaky (vary run to run, all pass in isolation):** `remote-runtime-shared-control-connection`, `terminal-history-incremental-restore` ×2, `cli/runtime-client` timeout override.

Two consecutive full-suite runs on **unmodified `main`** produced *different* failure sets (4, then 3), which is what establishes the pool as pre-existing. None of these files import the changed code.

**2 rounds, 3 independent adversarial reviewers each (equivalence / integration / house-style).**

- **Round 1 — needs-changes** (2 of 3). Vacuous benchmark guard; overstated speedup; drop arm unpinned by tests; `const window` shadowing the DOM global; comments too verbose for AGENTS.md.
- **Round 2 — all round-1 items confirmed fixed.** Found 2 further real gaps: both the prefix loop and the behind-window probe were pinned only for `https://`, so an `http://` GHES PR link split across chunks could regress undetected. Both now covered and mutant-killed.
- **Final verdict: clean.**

## AI Review Report

Reviewers independently re-derived every claim rather than trusting the summary — re-running mutants, re-measuring the benchmark, and re-fuzzing equivalence (one ran a 2.32M-input exhaustive sweep and a 3,000-session whole-detector differential against `origin/main`; 0 mismatches).

Cross-platform: no path, shell, or keyboard behavior touched; the change is pure string arithmetic on PTY bytes. Behavior is byte-identical for local, SSH, and remote-runtime PTYs, which matters because the file header calls out that the renderer scan path and the main-process authority path must not diverge.

Reviewed and dismissed with evidence:

- **Unicode/surrogates** — `slice()` at the window start can split a surrogate pair, but a lone surrogate matches no prefix of `https://`/`http://`, and the returned carry always comes from `value.slice(schemeIndex)` over the full value. Verified with surrogate-heavy fuzz and code-unit-at-a-time delivery.
- **Sliced-string retention** — V8 sliced strings pin their parent (~200 MB retained for 6 KB of carry). Measured old vs new: **150.0 MB vs 150.0 MB**. Identical, because both return `value.slice(schemeIndex)`. Pre-existing, not introduced here; flagged as a possible separate follow-up rather than silently expanding this PR's scope.
- **Asymptotics** — never worse than before: old is unconditionally O(n); new is O(512) plus an O(n) probe reachable only in the fragment arm.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. The change strictly *reduces* how much attacker-influenced PTY output is scanned per chunk. The 512-byte cap and the existing `MAX_TERMINAL_GITHUB_PR_URL_LENGTH` bound are both preserved, so a malicious agent still cannot grow the carry buffer without bound. No dependency changes.

## Notes

The `fragment tail` benchmark column exists specifically so the one branch where this change adds work stays visible rather than hidden behind a favorable average.

### DIFF
--- a/config/scripts/terminal-pr-link-carry-benchmark.mjs
+++ b/config/scripts/terminal-pr-link-carry-benchmark.mjs
@@ -1,0 +1,221 @@
+#!/usr/bin/env node
+// Benchmark: per-chunk cost of the GitHub PR-link carry scan on the PTY output path.
+//
+// createTerminalGitHubPRLinkDetector() runs on every PTY chunk (renderer
+// pty-connection + parked-terminal-byte-watcher). Before the fix,
+// getPotentialGitHubPRCarry() ran `lastIndexOf` for BOTH http scheme prefixes
+// across the entire combined chunk — even on the early-out path where the chunk
+// provably has no `/pull/`. The carry it returns is always a suffix of at most
+// MAX_CARRY_LENGTH (512) bytes, so every byte scanned before
+// `length - 512` was guaranteed-wasted work.
+//
+// The fix bounds the scan to that trailing window. This script measures the
+// scan itself across chunk sizes so the saved work is quantified.
+//
+// carryBefore/carryAfter are mirrors: node cannot import the .ts source, which is
+// why the sibling benchmarks in this directory inline their subject too. The
+// constants below are re-read from the real module at startup so a drifted cap or
+// scheme list fails loudly here instead of quietly benchmarking dead code.
+import { readFileSync } from 'node:fs'
+import { performance } from 'node:perf_hooks'
+import { fileURLToPath } from 'node:url'
+
+const DETECTOR_SOURCE = readFileSync(
+  fileURLToPath(new URL('../../src/shared/terminal-github-pr-link-detector.ts', import.meta.url)),
+  'utf8'
+)
+
+function readMirroredConstants(source) {
+  const cap = source.match(/const MAX_CARRY_LENGTH = (\d+)/)
+  const prefixes = source.match(/const HTTP_SCHEME_PREFIXES = \[([^\]]+)\]/)
+  if (!cap || !prefixes) {
+    throw new Error(
+      'terminal-github-pr-link-detector.ts no longer exposes MAX_CARRY_LENGTH / HTTP_SCHEME_PREFIXES in the expected shape; re-sync this benchmark with the implementation.'
+    )
+  }
+  return {
+    maxCarryLength: Number(cap[1]),
+    httpSchemePrefixes: prefixes[1]
+      .split(',')
+      .map((entry) => entry.trim().replace(/^['"]|['"]$/g, ''))
+      .filter(Boolean)
+  }
+}
+
+const { maxCarryLength: MAX_CARRY_LENGTH, httpSchemePrefixes: HTTP_SCHEME_PREFIXES } =
+  readMirroredConstants(DETECTOR_SOURCE)
+const ITERATIONS = Number.parseInt(process.env.ORCA_PR_CARRY_BENCH_ITERATIONS ?? '2000', 10)
+const WARMUP = Number.parseInt(process.env.ORCA_PR_CARRY_BENCH_WARMUP ?? '200', 10)
+
+for (const [name, value] of [
+  ['ORCA_PR_CARRY_BENCH_ITERATIONS', ITERATIONS],
+  ['ORCA_PR_CARRY_BENCH_WARMUP', WARMUP]
+]) {
+  if (!Number.isInteger(value) || value <= 0) {
+    throw new Error(`${name} must be a positive integer, received ${value}`)
+  }
+}
+
+function hasTerminalUrlWhitespace(value, start, end) {
+  for (let index = start; index < end; index += 1) {
+    if (/\s/.test(value.charAt(index))) {
+      return true
+    }
+  }
+  return false
+}
+
+function endsWithHttpSchemePrefixFragment(value) {
+  for (const prefix of HTTP_SCHEME_PREFIXES) {
+    for (let length = Math.min(prefix.length - 1, value.length); length > 0; length--) {
+      if (value.endsWith(prefix.slice(0, length))) {
+        return value.slice(value.length - length)
+      }
+    }
+  }
+  return ''
+}
+
+// Pre-fix implementation, kept verbatim for comparison.
+function carryBefore(value) {
+  const schemeIndex = Math.max(...HTTP_SCHEME_PREFIXES.map((prefix) => value.lastIndexOf(prefix)))
+  if (schemeIndex !== -1) {
+    const tailLength = value.length - schemeIndex
+    if (tailLength > MAX_CARRY_LENGTH) {
+      return ''
+    }
+    return hasTerminalUrlWhitespace(value, schemeIndex, value.length)
+      ? ''
+      : value.slice(schemeIndex)
+  }
+  return endsWithHttpSchemePrefixFragment(value)
+}
+
+// Post-fix implementation, mirroring src/shared/terminal-github-pr-link-detector.ts.
+function lastIndexOfHttpScheme(value, fromIndex) {
+  let lastIndex = -1
+  for (const prefix of HTTP_SCHEME_PREFIXES) {
+    const candidate =
+      fromIndex === undefined ? value.lastIndexOf(prefix) : value.lastIndexOf(prefix, fromIndex)
+    if (candidate > lastIndex) {
+      lastIndex = candidate
+    }
+  }
+  return lastIndex
+}
+
+function carryAfter(value) {
+  const windowStart = value.length > MAX_CARRY_LENGTH ? value.length - MAX_CARRY_LENGTH : 0
+  const window = windowStart === 0 ? value : value.slice(windowStart)
+  const schemeIndexInWindow = lastIndexOfHttpScheme(window)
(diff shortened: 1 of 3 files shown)


nwparker:
## Follow-up: would a deque / ring buffer help here?

Asked whether we can do better than the bounded-window fix — e.g. a deque for the carry. **Answer: no, and the profile says why.** Posting the numbers since the conclusion is "stop here," which is worth having on the record.

### Where the time actually goes now

Per-chunk breakdown on the early-out path (no `/pull/` present — the overwhelmingly common case), after this PR:

| chunk | `carry + data` | `includes()` | carry scan | **includes share** |
|---|---|---|---|---|
| 4 KiB | 0.00002 ms | 0.00010 ms | 0.00057 ms | 14.9% |
| 64 KiB | 0.00001 ms | 0.00109 ms | 0.00047 ms | 69.3% |
| 256 KiB | 0.00000 ms | 0.00428 ms | 0.00046 ms | 90.1% |
| 1 MiB | 0.00000 ms | 0.01709 ms | 0.00043 ms | **97.5%** |

Two things fall out:

1. **The concat a deque would eliminate is already free.** It measures 0.00000 ms at 256 KiB and above — V8 ropes make `carry + data` O(1) amortized, and the rope is only flattened by the search that follows, which we'd pay regardless. A ring buffer would replace a zero-cost operation with bookkeeping.
2. **`includes()` now dominates** — up to 97.5% — and this PR doesn't touch it. The carry scan this PR fixed is now ~2.5% of the chunk cost at 1 MiB. Further carry work is optimizing the wrong term.

### So could we beat the `includes()` instead?

That's the only lever left, so I tried it:

| approach | 64 KiB | 1 MiB |
|---|---|---|
| native `includes()` | 1.0× | 1.0× |
| `indexOf('/')` chain + `startsWith` | 1.0× | 1.1× **slower** |
| `/\/pull\//` regex | 1.0× | 1.0× |
| manual `charCodeAt` scan | **85× slower** | **90× slower** |

V8's `includes()` is SIMD-optimized native code. Nothing written in JS gets near it, and the hand-rolled scan is catastrophically worse.

### Recommendation

**No further work on this path.** The remaining cost is an unavoidable full-chunk substring search that has to happen for correctness — any chunk really can contain a PR link. A deque would add a stateful data structure and its own correctness surface (the carry semantics are subtle enough that this PR needed 26 tests and 7 mutants to pin) in exchange for optimizing a term that already measures zero.

If the terminal output path needs more headroom later, the productive direction is upstream — fewer/larger chunks reaching the detector, or skipping it entirely for panes with no PR-link consumers — not a faster carry.

nwparker:
## Deque follow-up: it was built, measured, and rejected

Following up on the earlier profile — a second independent investigation **actually implemented** the deque rather than reasoning about it, and reached the same conclusion with harder evidence. Worth recording, since "we tried it and it doesn't help" is the useful artifact here.

### The deque was built and verified correct

- Passes **26/26** existing tests (`vitest --no-cache`)
- **17,429 differential fuzz observations**, 0 mismatches vs the current implementation

So this is not "we couldn't make it work." It works; it just isn't faster:

| chunk size | deque vs current |
|---|---|
| 4 KB / 48 KB | 1.00-1.02x (noise) |
| **1 byte (SSH)** | **0.878x - a net loss** |

### Why: an ablation on the real code path

Running the real detector, then re-running with one stage removed, over **captured terminal output from this repo** (`git log -n 3000`, `find src`, `npm ls --all`) at the real PTY chunk cap (`TERMINAL_STREAM_CHUNK_BYTES` = 48 KiB):

| corpus (48 KB chunk) | full | `includes()` | carry | concat |
|---|---|---|---|---|
| git log | 3,000 ns | 2,250 ns | 625 ns | **0 ns** |
| find src | 32,459 ns | 31,084 ns | 792 ns | **41 ns** |
| npm ls | 23,250 ns | 20,792 ns | 626 ns | **41 ns** |

- `includes('/pull/')` is **75-97%** of the detector.
- `getPotentialGitHubPRCarry` is **flat at ~390-960 ns from 1 byte to 1 MB** — this PR working exactly as intended; there is nothing left in it.
- The concat a deque would remove costs **0-83 ns**. And carry is non-empty on **<2% of real chunks**, so it rarely even runs.

### A harness warning worth passing on

The investigator's first two harnesses reported `includes()` on a 1 MB string at **29 TB/s**, and rope-`includes` measuring *faster* than flat-`includes`. Both were V8 hoisting the loop-invariant pure call out of the timing loop. Every number above is first-touch: one `hrtime.bigint()` bracket per never-before-searched string, 150-200 distinct strings, median — sanity-checked at a stable 1.25-2.2 bytes/ns, i.e. genuinely memory-bandwidth-bound.

This is the same class of error that produced a false regression earlier in this series. **If you benchmark a pure function in a loop in this repo, assume LICM until you have proven otherwise.**

### Also measured and rejected

| idea | result |
|---|---|
| Rope-flatten avoidance | Real (+5.6-9.8% at 16-48 KB) but rides on the <2% carry chunks, so **<0.2% end-to-end** |
| Cheaper needle (`ull/`, `pull/`, `ll/`) | Corpus-dependent sign flip, 0.34x-3.53x; killed by `git log` at **0.34x** |
| `indexOf('/')` memchr prefilter | Genuinely fast (1560 bytes/ns) but hits on ~100% of real chunks, so it never short-circuits — pure added cost |

**Conclusion unchanged: no further work here.** The remaining cost is a full-chunk substring search that has to happen for correctness, already running as native SIMD code.

PR #10742, opened 2026-07-26
perf(agent-status): scan Command Code transcripts backward from EOF

nwparker:
## Summary

`readLastCommandCodeUserPromptEntryFromTranscript()` read up to `TRANSCRIPT_MAX_SCAN_BYTES` (4 MB) synchronously, decoded it all to a JS string, and `JSON.parse`d **every line to the end of the buffer** to find the *last* user entry — on **every** Command Code hook event.

`PreToolUse`/`PostToolUse` fire once per tool call, so an active agent turn paid that scan many times a second, and the cost grew with the transcript as the session ran.

This scans backward from EOF in 64 KiB blocks and stops at the first user line — the shape the sibling `readLastTextFromTranscriptOnce()` in the same file already used. The current turn's prompt sits near EOF, so the scan now reads one or two blocks. Byte offsets stay absolute so the prompt `interactionKey` is unchanged.

## ELI5

To show what you asked the agent, Orca finds the most recent thing you typed in the session log.

The log is append-only and the newest entry is at the **end**. The old code started at the beginning and read the *whole* file — parsing every line as JSON — keeping whichever match it saw last. Every single time a tool ran.

It's like re-reading a diary from page one to find yesterday's entry. Now it opens at the back and reads until it hits the first entry it wants, which is usually the last page or two. Longer sessions no longer cost more.

## Screenshots

No visual change.

`node config/scripts/command-code-transcript-scan-benchmark.mjs` (new):

```
     size   before ms   after ms   speedup
  0.18 MB       0.277      0.068        4x
  0.64 MB       0.899      0.067       13x
  1.86 MB       2.402      0.068       35x
  3.70 MB       4.861      0.066       74x
```

The multiple is not the point — **the new cost is flat**. Old was O(transcript) and grew all session; new is O(tail). The benchmark asserts both implementations resolve the same prompt before timing, and re-reads `TRANSCRIPT_CHUNK_BYTES`/`TRANSCRIPT_MAX_SCAN_BYTES` from the real module so a drifted constant fails loudly instead of benchmarking dead code.

The fixture models a real session: many completed turns, then this turn's prompt, then the tool output produced since.

## Proof of zero regression

Pure optimization — same prompt **and** same `interactionKey` for every transcript. The key embeds the prompt line's **absolute byte offset**, so the backward scan has to reproduce offsets exactly; that was the hard part.

**13-shape differential** against a verbatim copy of the original implementation, matching on both text and `interactionKey`: empty, only-assistant, user-last, user-first, no-trailing-newline, blank lines, malformed JSON, unicode, multi-user, CRLF, 3,000-line multi-chunk, >4 MB over-cap, and over-cap-with-no-user-line-inside. I verified the differential itself is not vacuous by confirming it catches a deliberate `offset + 1` bug and a wrong-direction scan.

**260 byte-alignments** of multi-byte UTF-8 swept across the 64 KiB boundary — the highest-risk case, since the old code decoded one buffer and the new one decodes per region. All correct. The invariant: a region always starts immediately after `0x0a`, and `\n` can never be part of a multi-byte sequence, so a region start is always a character boundary; `carryBytes` is never decoded alone.

**Mutation testing — 6/6 killed** (`--no-cache` required):

| mutant | result |
|---|---|
| drop `regionPosition` from the offset | 1 fails |
| scan last line of region only | 4 fail |
| drop the inter-chunk carry | 1 fails |
| keep the leading partial line | 1 fails |
| offset skips the newline | 1 fails |
| offset + 1 | 1 fails |

Two of my tests initially *failed* to kill mutants and were rewritten: an offset test that compared two different files (vacuous — the key hashes the path, so they differ regardless), and a "straddling" fixture whose prompt I computed to actually sit *inside* a chunk rather than across the boundary.

3 new tests: chunk-boundary straddle (with the boundary arithmetic computed in-test, not hardcoded), last-prompt-wins, and absolute-offset keying.

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 113/113 in the touched file; **676 tests pass across all command-code + agent-hooks suites**; full-suite failures are a subset of the pre-existing set

`pnpm test` is already red on `main`; baseline captured on a tree identical to `origin/main`. Deterministic: `persistence.test.ts` ×1, `agent-exec-handler.test.ts` ×2 (assert against `process.env`). Load-flaky, all passing in isolation: `remote-runtime-shared-control-connection`, `terminal-history-incremental-restore` ×2, `cli/runtime-client` timeout. Two consecutive runs on unmodified `main` gave different failure sets (4, then 3).

**1 review round, 4 independent reviewers. Final verdict: ship.**

**1 round, 4 independent adversarial reviewers** (equivalence / mutation / integration / house-style), following 2 rounds on the sibling PR that established the methodology (vacuous-guard and unpinned-branch classes of finding).

A design proposal was **rejected before implementation**: gating the scan on `isNewTurnEvent('command-code', …)`. That returns **hardcoded `false`** for this source, so the gate would have skipped every scan and silently broken prompt detection — the transcript read *is* Command Code's only prompt-discovery mechanism, as it has no `UserPromptSubmit` event. A `(path, size, mtime)` cache was also considered and rejected: staleness risk for no algorithmic gain once the scan is O(tail).

## AI Review Report

Reviewed for cross-platform correctness: CRLF transcripts (Windows) are covered in the differential; no path, shell, or platform-specific behavior is touched. The reader is shared by local, SSH, and remote-runtime hosts and its contract is unchanged.

Specifically probed: the 4 MB cap boundary and its leading-partial-line discard; lines longer than one 64 KiB chunk; files with no trailing newline; concurrent appends producing a torn last line at EOF; and whether returning on the first backward match can ever differ from the old "last match in window" (it cannot — they are the same line by definition, which the differential confirms empirically).

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. Transcript paths still arrive from the same hook payload field and are still guarded by the same `typeof`/length check and `try`/`catch`. The change **reduces** how much untrusted transcript content is parsed per event, and the 4 MB read cap is preserved exactly — including its discard-partial-first-line behavior, so a crafted transcript cannot make the reader parse beyond the cap. No dependency changes.

## Notes

The now-unused `iterateTranscriptLinesWithByteOffsets` generator is deleted; its only caller was the loop this replaces (verified by grep, and typecheck fails loudly on an unused private function).

### DIFF
--- a/config/scripts/command-code-transcript-scan-benchmark.mjs
+++ b/config/scripts/command-code-transcript-scan-benchmark.mjs
@@ -64,9 +64,62 @@ for (const [name, value] of [
   }
 }
 
+// Mirror of parseAgentHookJson: the real reader scans a line's structure before
+// parsing it, on BOTH sides of this comparison. Omitting it made the pre-fix
+// column ~9x too fast and invented a regression that does not exist.
+const HOOK_STRUCTURAL_TOKENS = 128 * 1024
+const HOOK_NESTING_DEPTH = 64
+
+function assertJsonStructure(content) {
+  let structuralTokens = 0
+  let depth = 0
+  let inString = false
+  let escaped = false
+  for (let index = 0; index < content.length; index += 1) {
+    const character = content[index]
+    if (inString) {
+      if (escaped) {
+        escaped = false
+      } else if (character === '\\') {
+        escaped = true
+      } else if (character === '"') {
+        inString = false
+      }
+      continue
+    }
+    if (character === '"') {
+      inString = true
+      continue
+    }
+    if (
+      character !== '{' &&
+      character !== '}' &&
+      character !== '[' &&
+      character !== ']' &&
+      character !== ',' &&
+      character !== ':'
+    ) {
+      continue
+    }
+    structuralTokens += 1
+    if (structuralTokens > HOOK_STRUCTURAL_TOKENS) {
+      throw new Error('structuralTokens')
+    }
+    if (character === '{' || character === '[') {
+      depth += 1
+      if (depth > HOOK_NESTING_DEPTH) {
+        throw new Error('nestingDepth')
+      }
+    } else if (character === '}' || character === ']') {
+      depth = Math.max(0, depth - 1)
+    }
+  }
+}
+
 function extractUserPrompt(line) {
   let entry
   try {
+    assertJsonStructure(line)
     entry = JSON.parse(line)
   } catch {
     return undefined
@@ -367,7 +420,7 @@ try {
     )
   }
   console.log(
-    '\nThe single-line row is the one shape that is slower: with no newline to stop\non, the scan reads the whole line in blocks and joins once, where the old code\nissued one flat read. Cost stays linear in the line (the carry is a chunk list,\nnot a re-joined buffer), and a transcript of ordinary JSONL lines never hits it.'
+    '\nThe win shrinks toward parity as output accumulates after the prompt, since\nthe backward scan has to read past all of it. The single-line row is the floor:\nno newline to stop on, so the scan reads the line in blocks and joins once where\nthe old code issued one flat read. Both sides pay the same per-line structure\nscan, and the carry is a chunk list, so cost stays linear either way.'
   )
 } finally {
   rmSync(dir, { recursive: true, force: true })
--- a/config/scripts/command-code-transcript-scan-benchmark.mjs
+++ b/config/scripts/command-code-transcript-scan-benchmark.mjs
@@ -51,6 +51,7 @@ function readMirroredConstant(name) {
 
 const TRANSCRIPT_CHUNK_BYTES = readMirroredConstant('TRANSCRIPT_CHUNK_BYTES')
 const TRANSCRIPT_MAX_SCAN_BYTES = readMirroredConstant('TRANSCRIPT_MAX_SCAN_BYTES')
+const EMPTY_REGION = Buffer.alloc(0)
 const ITERATIONS = Number.parseInt(process.env.ORCA_CC_SCAN_BENCH_ITERATIONS ?? '150', 10)
 const WARMUP = Number.parseInt(process.env.ORCA_CC_SCAN_BENCH_WARMUP ?? '20', 10)
 
@@ -145,7 +146,8 @@ function findLastPromptInRegion(region) {
   return undefined
 }
 
-// Post-fix: walk backward from EOF, return on the first user line.
+// Post-fix: walk backward from EOF, return on the first user line. The carry is
+// a chunk list, not a re-joined buffer, so one oversized line stays linear.
 function readBackward(path) {
   const size = statSync(path).size
   if (size <= 0) {
@@ -153,15 +155,16 @@ function readBackward(path) {
   }
   const fd = openSync(path, 'r')
   try {
-    let carryBytes = Buffer.alloc(0)
+    let carryChunks = []
     let bytesRead = 0
-    while (bytesRead < size && bytesRead < TRANSCRIPT_MAX_SCAN_BYTES) {
+    let scanEnd = size
+    while (scanEnd > 0 && bytesRead < TRANSCRIPT_MAX_SCAN_BYTES) {
       const chunkSize = Math.min(
-        size - bytesRead,
+        scanEnd,
         TRANSCRIPT_CHUNK_BYTES,
         TRANSCRIPT_MAX_SCAN_BYTES - bytesRead
       )
-      const position = size - bytesRead - chunkSize
+      const position = scanEnd - chunkSize
       const buffer = Buffer.alloc(chunkSize)
       let filled = 0
       while (filled < chunkSize) {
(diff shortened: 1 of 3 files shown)


nwparker:
## Review round 1 — a real regression found and fixed

An adversarial reviewer found that the backward scan could be **slower than the code it replaced**, and it was right. Reproduced independently before fixing.

### The bug

The carry buffer was re-joined with `Buffer.concat` on every block that contained no newline. A transcript whose tail is one large tool-result line (a file read, a long command output) has no newline to stop on, so the carry was re-copied per block — **O(line²)**.

Measured on the original commit, trailing region = one huge line:

| line size | before (flat read) | after (backward scan) | ratio |
|---|---|---|---|
| 2 MB | 1.10 ms | 4.24 ms | **0.26x** |
| 3.9 MB | 2.11 ms | 10.30 ms | **0.20x** |

That is a genuine regression on a realistic shape, not a synthetic one — agent turns embed file reads and command output as single JSONL lines.

### The fix

Keep the carry as a **chunk list** and join once, instead of re-joining per block. At 3.9 MB the same case goes 10.30 ms → 4.68 ms, and the over-cap case now beats the old code outright.

Also hardened per the same review: a **short read now bails out** rather than deriving an offset from a size the file no longer has. Without it, a file truncated between `stat()` and `read()` desynchronizes the absolute offset that `interactionKey` depends on. I confirmed the short read is reachable (requested 65536 bytes, got 0 after truncation).

### Guarding it

The obvious wall-clock test was the wrong instrument — 32 MB of copying is still fast enough in absolute terms that a timing threshold is flaky. The regression test instead **counts bytes copied** through `Buffer.concat`, which is deterministic:

- linear (fixed): well under the 4× budget
- quadratic (mutant): **36,634,737 bytes vs 8,388,608 allowed** — fails decisively

Mutation battery re-run: **5/6 killed.** The sixth is the short-read guard, which needs a genuine mid-scan truncation race; I removed a test that appeared to cover it but did not, rather than leave a misleading green.

### Benchmark now reports where this loses

The reviewer's fair criticism was that the fixture only modelled the *start* of a turn. The table now sweeps bytes-after-prompt and includes rows **below 1x**:

```
Worst cases (win decays as a turn progresses; <1x means slower):
                  case   before ms   after ms     ratio
  0.03 MB after prompt       1.283      0.070    18.22x
  0.25 MB after prompt       1.388      0.502     2.77x
  1.00 MB after prompt       2.032      1.810     1.12x
  3.00 MB after prompt       3.768      5.660     0.67x
      3 MB single line       1.507      4.311     0.35x
```

**The honest trade:** this wins big early in a turn (when hook events are most frequent) and loses when several megabytes of output have accumulated since the prompt. The single-line row is the one shape that is structurally slower — no newline to stop on, so the scan reads the line in blocks and joins once where the old code issued one flat read. Cost stays linear either way.

I've left those rows in the committed benchmark rather than tuning the fixture to look better.

Gates re-run: oxlint clean, typecheck clean, 114/114 in the touched file, **582 tests across all command-code + agent-hooks suites**.

nwparker:
### Full-suite verification after the fix

Cleanest run of the series — **2 failures, 36,703 passed**:

```
Test Files  1 failed | 3441 passed | 8 skipped (3450)
     Tests  2 failed | 36703 passed | 61 skipped (36766)
```

Both are `agent-exec-handler.test.ts`, which asserts against the machine's real `process.env` — a strict **subset** of the pre-existing baseline, and neither imports the changed code.

One correction to my own earlier characterisation: I had listed `persistence.test.ts › derives fresh default profiles without branch` as *deterministically* failing. It **passed** in this run. So that test is load-sensitive too, not deterministic — the pre-existing failure pool is entirely load-dependent apart from the two env-dependent relay tests.

Observed failure sets across five full-suite runs:

| run | tree | failures |
|---|---|---|
| A | clean `main` | 4 |
| B | clean `main` (rerun) | 3 |
| C | #10741 | 5 |
| D | #10742 (original) | 4 |
| E | #10742 (fixed) | **2** |

Two consecutive runs on *unmodified* `main` producing different sets is what establishes the pool as pre-existing rather than change-induced.

nwparker:
## Review round 2 — I have to retract the regression claim

Round 2 found that **my own benchmark was wrong**, and the correction goes against my earlier comment. Retracting it explicitly.

### The benchmark's "before" column was measuring the wrong function

The mirror used a bare `JSON.parse` per line. The real pre-change reader called `parseAgentHookJson`, which runs `assertJsonTextStructureWithinLimits` **before** parsing — a per-character scan of every line. So the "before" column was timing something ~9x faster than the code actually being replaced.

With the guard mirrored on both sides:

| case | old (reported) | old (corrected) |
|---|---|---|
| 3 MB single line | 0.35x | **0.93x** |
| 3.00 MB after prompt | 0.67x | **1.34x** |

**So the regression I reported in my previous comment does not exist against the real pre-change implementation.** The single-line floor is ~parity, not a 3x slowdown, and the 3 MB-trailing case is a win.

To be clear about what stands and what doesn't:

- ✅ **The O(line²) carry was a real bug** and the fix is real — that was measured against the *new* code's own earlier revision, where both sides shared the same parse cost, so that comparison was sound.
- ❌ **"Measurably slower than the flat read it replaced (0.20x)"** — wrong. That number came from the unfaithful mirror. The commit message on `52e0e83` carries this error; the correction is in `5b1cba9`.

Benchmark prose updated to match: the win shrinks toward parity as output accumulates after the prompt, and the single-line row is the floor rather than a regression.

### Two genuinely untested branches, now covered

Mutation review also found the carry and cap branches unpinned. Both mutants previously survived all 113 tests:

- `carryChunks.unshift(buffer)` → `carryChunks = []` — silently loses any prompt on a line spanning a whole 64 KiB block (a large tool result, or a pasted prompt). **Now fails 1 test.**
- Removing the 4 MB cap from the loop condition — restores the unbounded read this PR exists to remove.

Added: a >64 KiB prompt line spanning multiple blocks, and a >4 MB transcript whose only user line sits beyond the cap.

### One reported mutant I'm calling equivalent, not a gap

`atStart = position === 0 || bytesRead >= TRANSCRIPT_MAX_SCAN_BYTES` (keep the partial line at the cap instead of discarding it) survives — but it is **unobservable**: a truncated suffix of a JSON line is not valid JSON, so it fails to parse whether kept or discarded. Reported here rather than papering over it with a test that couldn't distinguish the two.

Gates: typecheck clean, oxlint clean, **116/116** in the touched file, full suite **2 failures / 36,762** — both the pre-existing `agent-exec-handler` env-dependent cases.

PR #10743, opened 2026-07-26
perf(renderer): give owner-routed settings a stable identity

nwparker:
## Summary

`getSettingsForRepoRuntimeOwner()` spread the whole `GlobalSettings` object (**185 fields**) on every call. It runs inside `useShallow` selectors at **~43 call sites** across `PullRequestPage` and `TaskPage`.

Zustand re-runs every subscribed selector on every store write, so each *unrelated* write allocated one settings-sized object per row and then shallow-compared all 185 fields to conclude nothing had changed.

This caches by repo id and reuses the reference while the settings object, repo list, and resolved owner are unchanged, so `useShallow` settles on the `Object.is` fast path.

## ELI5

Each row on the PR and task pages needs the app's settings, but with one field swapped to point at the machine that owns that row's repo.

The old code rebuilt a fresh 185-field copy for every row, every time *anything* in the app changed — even something unrelated like a terminal printing a character. Then React compared all 185 fields, field by field, to discover nothing had changed and nothing needed redrawing.

Now, if the settings and the repo's owner haven't changed, each row gets back the *exact same object* it had before. React sees it's literally the same thing and skips the comparison entirely.

## Screenshots

No visual change — this only affects object identity, never the values a component reads.

`node config/scripts/repo-owner-settings-selector-benchmark.mjs` (new):

```
GlobalSettings fields=185 writes=2000 warmup=200 (median of 5 rounds)
  rows   before ms   after ms   speedup
     1      0.0251     0.0003       79x
    10      0.2411     0.0014      178x
    43      1.0206     0.0069      149x
   100      2.4084     0.0309       78x
```

At the realistic ~43 on-screen call sites this saves **~1 ms of main-thread work per store write** — and that cost was paid on writes that touch nothing these selectors read, which is exactly what happens while an agent streams terminal output.

The benchmark simulates the real loop (selector + `useShallow` compare per row), verifies both versions produce equal values before timing, and reads the field count from the real `GlobalSettings` type so a drifted shape fails loudly rather than flattering the result with a stale, smaller object. I confirmed the equivalence guard is non-vacuous by breaking the cached selector and watching it throw.

## Proof of zero regression

The returned **value** is unchanged in every case; only its **identity** is now reused when nothing it derives from has changed. That makes cache invalidation the entire risk surface, so each invalidation source is tested in isolation.

5 new tests: stable identity across repeated calls; per-repo-id identities that survive interleaving; new value when the settings object changes; new value when the repo list changes *with the owner held constant*; new value when the owner changes *with the repo list identity held constant*; and cache bounding.

**Mutation testing — 6/6 killed** (`--no-cache`):

| mutant | result |
|---|---|
| ignore settings-source change | 1 fails |
| ignore repos-source change | 1 fails |
| ignore resolved-owner change | 1 fails |
| collapse to a single cache key | 2 fail |
| remove the size bound | 1 fails |
| break identity (`return { ...cached.value }`) | 2 fail |

The repos-source and resolved-owner mutants **initially survived**: my first fixture changed the repo array *and* the resolved owner together, so dropping either check alone was still caught by the other. The two isolating tests above were written specifically to close that.

The cache is bounded at 256 keys with insertion-order eviction, matching the existing idiom in `local-preflight-context-cache.ts`, because repo ids come from remote listings and the key space is not closed. `releaseRepoRuntimeOwnerSettingsCache()` is exported for test isolation.

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 24/24 in the touched file; full-suite failures are a subset of the pre-existing set

`pnpm test` is already red on `main`; baseline captured on a tree identical to `origin/main`. Deterministic: `persistence.test.ts` ×1, `agent-exec-handler.test.ts` ×2 (assert against `process.env`). Load-flaky, all passing in isolation: `remote-runtime-shared-control-connection`, `terminal-history-incremental-restore` ×2, `cli/runtime-client` timeout. Two consecutive runs on unmodified `main` gave different failure sets (4, then 3).

**2 review loops: adversarial review plus mutation-driven revision. Final verdict: ship.**

**1 round**, plus the mutation-testing loop above which drove 2 test revisions. Methodology carried over from the sibling PRs in this series, where 2 review rounds established the vacuous-guard and unpinned-branch failure classes that I then checked for here proactively.

## AI Review Report

Main risks reviewed:

- **Stale settings after a mutation.** The guard compares source *identities* (`state.settings`, `state.repos`) plus the resolved owner. Zustand state is immutable-by-convention, so any real settings change produces a new object and misses the cache. The in-place-mutation test covers the case where a caller mutates a repo row without replacing the array — the resolved-owner check catches it.
- **Cross-tenant leakage between repos.** Keyed per repo id, with an interleaving test proving one repo's entry cannot be served for another.
- **Unbounded growth.** Bounded at 256 with a test that proves eviction actually happens.
- **Module-level state in the renderer.** Matches the existing house pattern (`worktree-agent-orchestration-index.ts`, `local-preflight-context-cache.ts`), including an exported release function.

Cross-platform: no path, shell, keyboard, or platform-dependent behavior touched. Runtime-owner routing (local / SSH / runtime hosts) is resolved by the unchanged `getRuntimeEnvironmentIdForRepo`; this PR only affects the identity of the object wrapping it, so SSH and folder-workspace routing are unaffected.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. One consideration reviewed explicitly: the cache holds references to settings objects, which include AI/provider configuration. It lives in renderer memory that already holds the same objects via the store, adds no persistence and no new exposure, and is bounded so it cannot grow without limit. Cache keys are repo ids, never secrets. No dependency changes.

## Notes

`getRepoOwnerRoutedSettings()` — the sibling that rebinds settings for a single repo — is deliberately left uncached: it takes a settings object rather than store state, so it has no stable identity to key on, and its callers are event handlers rather than per-render selectors.

### DIFF
--- a/config/scripts/repo-owner-settings-selector-benchmark.mjs
+++ b/config/scripts/repo-owner-settings-selector-benchmark.mjs
@@ -1,0 +1,207 @@
+#!/usr/bin/env node
+// Benchmark: cost of the owner-routed settings selector per store write.
+//
+// getSettingsForRepoRuntimeOwner() is called from useShallow selectors at ~43
+// sites across PullRequestPage and TaskPage. Zustand re-runs every subscribed
+// selector on every store write, so before the fix each unrelated write
+// allocated one settings-sized object per row and then shallow-compared every
+// field to conclude nothing had changed.
+//
+// The fix caches by repo id and reuses the reference while the settings object,
+// repo list, and resolved owner are unchanged, so useShallow's equality check
+// short-circuits on Object.is.
+//
+// The selector body is mirrored here (node cannot import the .ts source, matching
+// the sibling benchmarks). The settings field count is read from the real
+// GlobalSettings type so a drifted shape fails loudly instead of flattering the
+// result with a stale, smaller object.
+import { readFileSync } from 'node:fs'
+import { performance } from 'node:perf_hooks'
+import { fileURLToPath } from 'node:url'
+
+const TYPES_SOURCE = readFileSync(
+  fileURLToPath(new URL('../../src/shared/types.ts', import.meta.url)),
+  'utf8'
+)
+
+function countGlobalSettingsFields(source) {
+  const block = source.match(/export type GlobalSettings = \{([\s\S]*?)\n\}/)
+  if (!block) {
+    throw new Error('types.ts no longer declares GlobalSettings in the expected shape')
+  }
+  const fields = block[1].match(/^\s{2}\w+\??:/gm) ?? []
+  if (fields.length < 50) {
+    throw new Error(`GlobalSettings parsed as only ${fields.length} fields; re-sync this benchmark`)
+  }
+  return fields.length
+}
+
+const SETTINGS_FIELDS = countGlobalSettingsFields(TYPES_SOURCE)
+const ROWS = Number.parseInt(process.env.ORCA_OWNER_SETTINGS_BENCH_ROWS ?? '43', 10)
+const WRITES = Number.parseInt(process.env.ORCA_OWNER_SETTINGS_BENCH_WRITES ?? '2000', 10)
+const WARMUP = Number.parseInt(process.env.ORCA_OWNER_SETTINGS_BENCH_WARMUP ?? '200', 10)
+
+for (const [name, value] of [
+  ['ORCA_OWNER_SETTINGS_BENCH_ROWS', ROWS],
+  ['ORCA_OWNER_SETTINGS_BENCH_WRITES', WRITES],
+  ['ORCA_OWNER_SETTINGS_BENCH_WARMUP', WARMUP]
+]) {
+  if (!Number.isInteger(value) || value <= 0) {
+    throw new Error(`${name} must be a positive integer, received ${value}`)
+  }
+}
+
+function makeSettings() {
+  const settings = { activeRuntimeEnvironmentId: 'focused-runtime' }
+  for (let index = 0; index < SETTINGS_FIELDS - 1; index += 1) {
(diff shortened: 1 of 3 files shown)


nwparker:
## Note on the sibling PRs in this series

This is one of three related perf PRs (#10741, #10742, #10743). Worth flagging for reviewers of this one: **#10742's first commit shipped a real regression** — an O(line²) carry copy that made it up to 5× slower than the code it replaced on a realistic input. It was caught by adversarial review, reproduced independently, and fixed before merge.

That is the reason this PR leans hard on mutation testing rather than on a green suite. Two of my own tests here initially **failed to kill their mutants** and were rewritten:

- One fixture changed the repo list *and* the resolved owner together, so dropping either invalidation check alone still passed. Split into two tests that hold the other variable constant.
- Verified with `--no-cache` — vitest's cache otherwise reports false survivals, which cost me a wrong conclusion earlier in the series.

Current state: **6/6 mutants killed**, including a `return { ...cached.value }` mutant that breaks identity while keeping values equal — the exact failure this optimization could have.

The one residual risk I'd want a human eye on: the cache assumes Zustand state is immutable-by-convention, so a caller that mutates a settings object in place without replacing the reference would serve a stale value. The resolved-owner check catches the repo-row case (there's a test for it), but a same-identity mutation of `settings` itself would not be detected. I believe that violates the store's existing contract everywhere else in the codebase, so I did not defend against it — flagging it explicitly rather than leaving it implicit.

PR #10744, opened 2026-07-26
fix(terminal): durable lost-worker archive with atomic retire and mixed-version SSH revive

eisen0419:
## Motivation

When a recognized worker's PTY is lost (daemon crash, SSH relay upgrade, dead remote process) or a user closes a tab, terminal history could be silently dropped or the tab left in a broken state: the all-leaf capture invariant (`terminal-archive-store.ts`) fails closed for multi-pane tabs whose sibling leaves have no snapshot source, SSH archives never retired their tab, and several exit/receipt races could close a replacement PTY or lose the archived-exit receipt.

## What this PR does

- **Durable archive store for tab close and lost workers**: every user tab close and lost-worker path routes through a single durable archive transition; archive metadata, tab retirement, and SSH `termination-pending` leases commit in **one atomic flush** with rollback on failure.
- **Typed revive protocol + v2 persistence envelope**: staged pre-teardown serialization gives every entry a `replayTail` bound to its `sourceIncarnationId`; the archive snapshot source resolves sibling leaves by paneKey with same-generation validation. Whole-envelope validation rejects malformed v2 payloads (fail-closed, no per-entry salvage).
- **Fail-closed cold probe** (`probeColdRestore` discriminated union): log read/decode/generation/replay failures are reported as completeness-lost instead of being disguised as untruncated snapshots; `captured-empty` requires provably empty (`data.length===0 && truncated===false`), unified across all four capture entry points via a shared constructor.
- **Exit/receipt correctness under PTY id reuse**: archived-exit receipts are bound to their incarnation (a replacement cannot consume an older generation's receipt; a rejected old-generation exit no longer suppresses the synthetic archived exit), reconnect-attempt abort is fenced up to (and re-checked at) archive commit, and shutdown cleanup/synthetic exits only act on the incarnation the shutdown captured — a plain kill can no longer close a replacement spawned during the await window.
- **Renderer-first completion**: the renderer's first/only exit for an archived PTY carries the archive receipt (pre-registered before shutdown to survive synchronous provider exits).

## Validation scope

Mixed-version SSH archive success is guaranteed only when **both ends support the v2 staged serializer**. A legacy (pre-v2) envelope is expected to fail closed with distinct diagnostics (`staged-state-legacy` / `staged-state-missing`) — no second-hand data is fabricated.

## Testing

- Unit/integration: capture fence (per-leaf + pre-commit), Store-level flush failure injection asserting three-way rollback, receipt incarnation binding (both race directions), plain kill/stop id-reuse (three caller paths), envelope validation, cold-probe failure taxonomy.
- E2E (Docker SSH relay, mixed-version): 4 scenarios including a true mixed outcome (one recognized worker lost + one ordinary sibling revived) and remote-process termination assertions — 4 passed.
- Full suite: no new failures vs. base (3 pre-existing environment-sensitive failures unrelated to this change).

## ELI5

When a terminal worker dies (crash, SSH upgrade, closed tab), history could vanish or tabs stuck broken. This archives lost workers more safely, retires tabs cleanly, and handles mixed-version SSH revive so you keep history and a consistent UI.

### DIFF
--- a/src/cli/terminal-format.ts
+++ b/src/cli/terminal-format.ts
@@ -169,7 +169,7 @@ export function formatTerminalFocus(result: { focus: RuntimeTerminalFocus }): st
 
 export function formatTerminalClose(result: { close: RuntimeTerminalClose }): string {
   if (result.close.closeMode === 'tab') {
-    return `Closed terminal tab ${result.close.tabId} (${result.close.handle}).`
+    return `Closed terminal tab ${result.close.tabId} (${result.close.handle}).${result.close.archiveId ? ` Archive ${result.close.archiveId}.` : ''}`
   }
   const ptyNote = result.close.ptyKilled ? ' PTY killed.' : ''
   return `Closed terminal ${result.close.handle}.${ptyNote}`
--- a/src/main/daemon/daemon-admission-limits.ts
+++ b/src/main/daemon/daemon-admission-limits.ts
@@ -197,6 +197,10 @@ function createOrAttachPayloadError(payload: unknown): string | null {
         DAEMON_PTY_HISTORY_SEED_MAX_BYTES
       ),
     () =>
+      payload.attachOnly === undefined || typeof payload.attachOnly === 'boolean'
+        ? null
+        : 'createOrAttach payload.attachOnly must be a boolean',
+    () =>
       optionalBoundedStringError(
         payload.shellOverride,
         'createOrAttach payload.shellOverride',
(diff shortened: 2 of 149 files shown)


Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17588. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @eisen0419. No action is requested from you. Thank you.

PR #10748, opened 2026-07-26
fix(shortcuts): gate Cmd/Ctrl+N folder-workspace jumps on path status

### LINKED ISSUE #10716: Cmd/Ctrl+N on a folder workspace skips the path-status gate that clicking its card enforces
Split out of #9497 / #10693 so it doesn't get lost. Credit to @Hanxcellent, who had this in the (self-closed) PR #9560.

**What happens:** `onJumpToWorktreeIndex` (`src/renderer/src/hooks/useIpcEvents.ts`) calls `activateAndRevealWorktree(visibleIds[index])` unconditionally. Since #10693, and in fact whenever the sidebar is open, that id can be a folder-workspace key (`folder:<id>`) — `getRenderedWorktreesInSidebarOrder` emits folder-workspace rows via `folderWorkspaceToWorktree`.

**This is not a crash.** I verified it: `activateAndRevealWorktree('folder:fw-1')` resolves fine (`findKnownWorktreeById`, `store/slices/worktrees.ts:687-702`, has a folder branch), sets `activeWorktreeId`/`activeWorkspaceKey`, nulls `activeRepoId`, and creates the initial terminal. Cmd+N on a folder workspace works.

**The inconsistency:** `activateAndRevealFolderWorkspace` (`worktree-activation.ts:194-206`) first runs `getFreshFolderWorkspacePathStatus` and bails with an error toast if `folderWorkspaceActivationBlocked(pathStatus)`. The plain path skips that gate entirely. So for a folder workspace whose path is **missing, unmounted, or on a disconnected host**:

- Clicking its sidebar card → error toast, activation blocked
- Pressing its Cmd/Ctrl+N number → activates silently, into a folder that isn't there

**Suggested fix** (Hanxcellent's, essentially): a small `activateWorkspaceNumberShortcut(workspaceId)` that branches on `parseWorkspaceKey` and routes `type === 'folder'` to `activateAndRevealFolderWorkspace`, mirroring `activateTabNumberShortcut`.

**Severity:** low — needs a folder workspace *and* a broken path *and* the number shortcut. Filing rather than fixing inline to keep #10693 to a single reviewable concern.

*Found during a bug-bash triage sweep; not a maintainer verdict.*

vansh2408:
## Summary

Pressing Cmd/Ctrl+1–9 on a folder-workspace sidebar slot now enforces the same path-status gate as clicking its card. Previously the number shortcut activated the workspace unconditionally, so a folder workspace whose path is missing,
unmounted, or on a disconnected SSH host activated silently into a folder that isn't there; clicking the same card correctly blocked with an error toast.

The `onJumpToWorktreeIndex` handler now routes through a new `activateWorkspaceNumberShortcut(workspaceId)` that branches on `parseWorkspaceKey` — `folder:` keys go through the guarded `activateAndRevealFolderWorkspace`, everything else through `activateAndRevealWorktree` — mirroring `activateWorktreeFromSidebar` and the nav-history activator.

Fixes #10716. Credit to @Hanxcellent, whose self-closed #9560 contained this fix.

## Screenshots

No visual change - the blocked case reuses the existing path-status error toast
already shown by the sidebar click path.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

## AI Review Report

- **Key-parsing asymmetry:** the folder path takes the stripped `folderWorkspaceId` while the worktree path takes the raw key — verified against `activateAndRevealFolderWorkspace`'s lookup of `state.folderWorkspaces` by raw id and the two existing dispatch sites, which use the identical shape.
- **Behavior parity:** the shortcut keeps default sidebar-reveal behavior (unlike the click path's `revealInSidebar: false`) because the shortcut can fire with the sidebar closed; `parseWorkspaceKey` returning `null` for plain
  worktree ids falls through to the generic path, preserving prior behavior for all non-folder targets.
- **Cross-platform (macOS/Linux/Windows):** explicitly checked. The change is downstream of the `onJumpToWorktreeIndex` IPC event — shortcut registration (`CmdOrCtrl` accelerators), labels, and key handling are untouched, so Cmd+N on macOS and Ctrl+N on Linux/Windows route identically. No paths, no shell invocation, no Electron platform branches. The SSH case is the point of the fix: the gate blocks folder workspaces on disconnected SSH hosts. Folder workspaces and git worktrees are both handled per AGENTS.md.

## Security Audit

- **Input handling:** `workspaceId` comes from the internally computed rendered sidebar order, not user-typed input; `parseWorkspaceKey` is plain prefix parsing with no regex or injection surface.
- **No command execution, filesystem, or path handling** added; no auth, secrets, or dependency changes.
- **IPC:** no new channels or handlers — the existing `onJumpToWorktreeIndex` listener's behavior is narrowed, not widened.
- **Net effect is fail-closed:** the change adds a gate, removing the ability to activate a workspace whose backing path is unavailable. No follow-up needed.

## Notes

Call out any platform-specific behavior, risks, or follow-up work.


vansh2408:
**Re: the Docstring Coverage warning** - this is CodeRabbit's default advisory threshold reacting to a small diff (2 eligible functions, 1 documented). The documented one is the new exported `activateWorkspaceNumberShortcut`, which carries the non-obvious constraint (mixed worktree-id / folder:-key input). The undocumented one is an anonymous test mock / inline callback; adding a docstring there would just restate the obvious, which `AGENTS.md` ("concise, non-obvious comments only") asks us to avoid. Leaving it as is - happy to add one if reviewers prefer.

OrcaWin:
Reviewed with an independent reproduction — including a live-app check of Orca opening a terminal into a directory I had just deleted. The bug is real and the gate is the right fix. Two things needed rework, both pushed to this branch.

**1. The tests did not exercise the fix.** `workspace-number-shortcut.test.ts` mocks `@/lib/worktree-activation` — the module whose folder branch *is* the fix — so it only asserted that a new wrapper forwards to a mock. I reverted the `useIpcEvents` wiring (restoring the exact original bug) with your tests still in place, and they passed 2/2. Replaced with a suite that drives the real activator against a real store, adds the missing-path and disconnected-SSH cases, adds darwin/win32/linux chord assertions, and pins the handler's call site so the wiring cannot silently regress.

**2. The fix duplicated code it already imported.** `worktree-activation.ts` already contained this exact worktree-vs-folder dispatch inline as the nav-history activator; the new module was a byte-identical copy, making it the 4th instance of the pattern. I exported the existing one as `activateAndRevealWorkspace` and pointed both the nav activator and the number shortcut at it. Net result is 2 files changed instead of a new module — the diff got smaller.

**Cross-platform (AGENTS.md):** clean. No `metaKey` is hardcoded; the binding is `Mod+1`, resolving to meta on darwin and control elsewhere, now asserted with a negative case for win32. Both AGENTS.md-flagged scenarios — folder workspace and disconnected SSH — verified blocking in the running app.

**Perf:** the guard is a synchronous in-memory cache read, not IPC or a `stat`, so no blocking work is added to the keystroke path.

Validation: typecheck exit 0, oxlint exit 0, 3516 tests passing across the touched trees.

Good catch on the underlying inconsistency — the click path enforcing a gate the shortcut skipped is exactly the kind of asymmetry that's easy to miss.

PR #10773, opened 2026-07-26
fix: prevent listen EINVAL on long sun_path in SSH relay

### LINKED ISSUE #10726: [Bug]: SSH relay fails to start (`listen EINVAL`) when the remote $HOME makes the socket path exceed the 108-byte `sun_path` limit
### Operating system

Linux

### Orca version

1.4.155.0

### Details

Connecting to an SSH target fails during relay startup:

**Root cause:** that path is 111 bytes. A UNIX domain socket address is capped at
`sizeof(sun_path)` = 108 bytes (107 usable) on Linux, and libuv returns `UV_EINVAL`
when the pipe name is too long — hence `listen EINVAL` rather than a clearer error.

The remote `$HOME` is 45 characters (a managed-hosting container where the home
directory is `/var/www/<uuid>`). The relay's own suffix consumes 66 characters: 
/.orca-remote/relay-<version>+<build>/relay-<id>.sock

So any host whose `$HOME` is longer than ~41 characters cannot start the relay.

```console
$ echo -n '/var/www/611f7cf9-f715-49e6-91d9-0ffac1d7c4c0/.orca-remote/relay-0.1.0+8d4e15ad63eb/relay-3f49cb06c8af7d25.sock' | wc -c
111

### PR DESCRIPTION
**Pull request by stroland02**
Resolves #10726.

### Summary

Prevents SSH relay startup from passing overlength Unix-domain socket paths to listen.

- Uses the UTF-8 byte limits of 107 bytes on Linux and 103 bytes on macOS.
- Shortens overlength endpoints with a deterministic full-path hash while retaining the target-specific socket suffix and relay-version isolation.
- Falls back to a short /tmp endpoint when the remote home path itself is too long.
- Keeps Reset Relay able to find and stop shortened endpoints across relay versions.
- Leaves Windows named-pipe behavior unchanged.

### Testing

- [x] Focused endpoint and reset tests: 11 passed.
- [x] Focused SSH deploy, live-connect, relay-loss, and system-transport suite: 44 passed, 1 platform skip.
- [x] Node typecheck.
- [x] Changed-code quality gate and full lint/reliability/localization suite.
- [x] Relay build for linux-x64, linux-arm64, darwin-x64, darwin-arm64, win32-x64, and win32-arm64.
- [x] Production Electron bundle build.
- [x] Linux openclaw runtime proof: a 165-byte raw socket path failed with EINVAL; the actual endpoint helper produced a 50-byte path; a real installed Orca relay bound it and returned ORCA-RELAY v0.1.0 READY; cleanup left no socket residue.

### AI Review

Reviewed SSH execution ownership, disconnect/reconnect behavior, stale socket cleanup, target-scoped reset behavior, crash/cast/retry/growth risk, security and supply chain, performance/resources, backward compatibility, folder workspaces, mixed-version wire behavior, Git provider neutrality, and macOS/Linux/Windows behavior. Two proven issues found during review (Darwin byte limit and reset cleanup for shortened sockets) were fixed and covered by tests. No remaining P0, P1, or P2 finding.

### Security Audit

No dependency, lockfile, binary, credential, shell-evaluation, or network-sink changes. Relay sockets remain created atomically with mode 0600 and endpoint credentials remain separate; shortened names are deterministic identifiers, not authentication.

### Visual Proof

N/A — this is main-process SSH relay endpoint handling with no rendered UI or visible layout change.

### Compatibility

No persisted state, mobile-facing surface, RPC field, stream opcode, or published remote-wire content changed. Folder workspaces and GitHub/GitLab review flows are unaffected because the endpoint is scoped to the SSH target rather than a repository workspace or provider.

### DIFF
--- a/src/main/ssh/ssh-relay-endpoints.test.ts
+++ b/src/main/ssh/ssh-relay-endpoints.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from 'vitest'
+import { relayEndpointForHost } from './ssh-relay-endpoints'
+import { getRemoteHostPlatform } from './ssh-remote-platform'
+
+const SOCK_NAME = 'relay-1234567890abcdef.sock'
+
+function asciiEndpointWithLength(byteLength: number): { remoteDir: string; sockName: string } {
+  return {
+    remoteDir: `/${'a'.repeat(byteLength - Buffer.byteLength(SOCK_NAME, 'utf8') - 2)}`,
+    sockName: SOCK_NAME
+  }
+}
+
+describe('ssh-relay-endpoints', () => {
+  describe('relayEndpointForHost', () => {
+    it('keeps Linux paths at the 107-byte limit', () => {
+      const hostPlatform = getRemoteHostPlatform('linux-x64')
+      const { remoteDir, sockName } = asciiEndpointWithLength(107)
+      const result = relayEndpointForHost(hostPlatform, remoteDir, sockName)
+      expect(result).toBe(`${remoteDir}/${sockName}`)
+      expect(Buffer.byteLength(result, 'utf8')).toBe(107)
+    })
+
+    it('shortens Linux paths above the 107-byte limit', () => {
+      const hostPlatform = getRemoteHostPlatform('linux-x64')
+      const { remoteDir, sockName } = asciiEndpointWithLength(108)
+      const result = relayEndpointForHost(hostPlatform, remoteDir, sockName)
+      expect(result).not.toBe(`${remoteDir}/${sockName}`)
+      expect(result).toMatch(/relay-1234567890abcdef\.sock$/)
+      expect(Buffer.byteLength(result, 'utf8')).toBeLessThanOrEqual(107)
+    })
+
+    it('keeps Darwin paths at the 103-byte limit and shortens 104 bytes', () => {
+      const hostPlatform = getRemoteHostPlatform('darwin-arm64')
+      const atLimit = asciiEndpointWithLength(103)
+      const overLimit = asciiEndpointWithLength(104)
+
+      expect(relayEndpointForHost(hostPlatform, atLimit.remoteDir, atLimit.sockName)).toBe(
+        `${atLimit.remoteDir}/${atLimit.sockName}`
+      )
+      const shortened = relayEndpointForHost(hostPlatform, overLimit.remoteDir, overLimit.sockName)
+      expect(shortened).not.toBe(`${overLimit.remoteDir}/${overLimit.sockName}`)
+      expect(Buffer.byteLength(shortened, 'utf8')).toBeLessThanOrEqual(103)
+    })
+
+    it('measures non-ASCII paths in UTF-8 bytes', () => {
+      const hostPlatform = getRemoteHostPlatform('darwin-x64')
+      const remoteDir = `/${'é'.repeat(38)}`
+      const fullPath = `${remoteDir}/${SOCK_NAME}`
+
+      expect(fullPath.length).toBeLessThanOrEqual(103)
+      expect(Buffer.byteLength(fullPath, 'utf8')).toBeGreaterThan(103)
+      const shortened = relayEndpointForHost(hostPlatform, remoteDir, SOCK_NAME)
+      expect(shortened).not.toBe(fullPath)
+      expect(Buffer.byteLength(shortened, 'utf8')).toBeLessThanOrEqual(103)
+    })
+
+    it('keeps version isolation when shortening into the relay parent directory', () => {
+      const hostPlatform = getRemoteHostPlatform('linux-x64')
+      const remoteDir = `/home/user/.orca-remote/relay-${'v'.repeat(100)}`
+      const result = relayEndpointForHost(hostPlatform, remoteDir, SOCK_NAME)
+
+      expect(result).toMatch(
+        /^\/home\/user\/\.orca-remote\/[a-f0-9]{12}-relay-1234567890abcdef\.sock$/
+      )
+    })
+
+    it('falls back to a target-identifiable /tmp path if the parent path is too long', () => {
+      const hostPlatform = getRemoteHostPlatform('linux-x64')
+      const remoteDir = `/${'a'.repeat(100)}/.orca-remote/relay-v1`
+      const result = relayEndpointForHost(hostPlatform, remoteDir, SOCK_NAME)
+      expect(result).toMatch(/^\/tmp\/orca-[a-f0-9]{12}-relay-1234567890abcdef\.sock$/)
+      expect(Buffer.byteLength(result, 'utf8')).toBeLessThanOrEqual(107)
+    })
+
+    it('returns the complete Windows named pipe format', () => {
+      const hostPlatform = getRemoteHostPlatform('win32-x64')
+      const remoteDir = 'C:/Users/user/.orca-remote/relay'
+      const sockName = 'relay.sock'
+      const result = relayEndpointForHost(hostPlatform, remoteDir, sockName)
+      expect(result).toMatch(/^\\\\\.\\pipe\\orca-relay-[a-f0-9]{20}$/)
+    })
+  })
+})
--- a/src/main/ssh/ssh-relay-endpoints.ts
+++ b/src/main/ssh/ssh-relay-endpoints.ts
@@ -8,13 +8,34 @@ import {
 
 export const WINDOWS_ACTIVE_PIPE_MARKER_PREFIX = '.windows-active-pipe-'
 
+const LINUX_UNIX_SOCKET_MAX_PATH_BYTES = 107
+const DARWIN_UNIX_SOCKET_MAX_PATH_BYTES = 103
+
 export function relayEndpointForHost(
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by nwparker** [comment]
## Readiness review — PR #10773 (`fix: prevent listen EINVAL on long sun_path in SSH relay`)

**Verdict: READY TO MERGE.** No P0 or P1 findings. One P2 residual risk noted below (follow-up, not blocking).

Reviewed at head `aa37252e9be`. Scope is small and well-contained: `ssh-relay-endpoints.ts` (+23/-1) and `ssh-relay-reset.ts` (+1/-1), plus 106 lines of new tests.

---

### Evidence

**Root cause reproduced with the real code path.** Feeding `relayEndpointForHost` the exact host shape from #10726 (`$HOME=/var/www/<uuid>`, 45 bytes) reproduces the reported failure and shows the fix:

```
BEFORE 111 bytes: /var/www/611f7cf9-.../.orca-remote/relay-0.1.0+8d4e15ad63eb/relay-3f49cb06c8af7d25.sock
AFTER   99 bytes: /var/www/611f7cf9-.../.orca-remote/49e44f997327-relay-3f49cb06c8af7d25.sock
```

**`sun_path` boundary measured on real Linux** (Ubuntu, `7.0.0-28-generic x86_64`, via SSH), binding real `AF_UNIX` sockets rather than trusting the header constant:

| path bytes | `bind()` result |
|---|---|
| 106 | OK |
| 107 | OK |
| 108 | OK |
| **109** | **EINVAL** |
| 111 | EINVAL (the reported failure) |

Same measurement on macOS (darwin arm64): 103 OK, 104 OK, **105 EINVAL**.

So the real usable limits are 108 (Linux) and 104 (Darwin); the PR thresholds of 107/103 are each one byte conservative. That is the safe direction — it can only shorten marginally earlier than strictly required, never too late — so the constants are correct as written.

**Reset-script globs verified against real bound sockets on Linux.** The widened loop in `ssh-relay-reset.ts` was exercised against all four layouts with actual `AF_UNIX` sockets bound at each location:

```
MATCH /home/neil/.orca-remote/relay-0.1.0+abc/relay-3f49cb06c8af7d25.sock   (legacy versioned dir)
MATCH /home/neil/.orca-remote/relay-3f49cb06c8af7d25.sock                    (legacy flat)
MATCH /home/neil/.orca-remote/49e44f997327-relay-3f49cb06c8af7d25.sock       (new: shortened parent)
MATCH /tmp/orca-49e44f997327-relay-3f49cb06c8af7d25.sock                     (new: /tmp last resort)
```

Both new socket locations are reachable by force-stop; no orphaned-relay regression.

**Local gates (macOS, this worktree at `aa37252e9be`):**

- `vitest src/main/ssh src/relay` — **3083 passed, 16 skipped, 0 failed** (252 files)
- Focused suites (`ssh-relay-endpoints`, `ssh-relay-reset`, `agent-hook-server`) — **27/27 passed**
- `pnpm run typecheck` — clean
- `oxlint` + `oxfmt --check` on all four changed files — clean

**CI:** all 49 checks pass, 1 conditional matrix job skipping, 0 failures. The previously-failing `e2e / ssh docker watcher isolation` job passed on rerun (4m14s). `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

---

### Checklist dimensions

**01 SSH / remote / network — pass.** This is the core dimension. Verified on real Linux above. The secondary path was traced and is not affected: on POSIX the relay derives its hook endpoint via `endpointDirForRelaySocket()` and writes a *regular file* holding a loopback TCP port + token (`agent-hook-endpoint-coordinates.ts:30`, `agent-hook-server.ts:88`). It is not an `AF_UNIX` bind, so it is bounded by `PATH_MAX`, not `sun_path` — the longer basename introduced here cannot trigger a second EINVAL. `--endpoint-dir` is passed only on the Windows launch path (`ssh-relay-deploy.ts:1817`); the value computed at `:1383` is unused on POSIX, so the main-side/relay-side derivations cannot diverge in practice.

**02 Crash / cast / retry / unbounded growth — pass.** Pure function, no casts, no `!`, no `JSON.parse`, no retry or state machine, no accumulating structure. One `sha256` computed only on the over-limit branch.

**03 Security — pass, with one P2 below.** The socket path stays shell-escaped at both consumers (`shellEscape(sockFile)`, `ssh-relay-deploy.ts:1468`); `sockName` is shell-escaped in the reset script. No new network sink, secret, or install hook.

**04 Performance — pass.** One SHA-256 over a <200-byte string, only when the path exceeds the limit. Not a hot path.

**05 Functional correctness — pass.** The `parentDir` heuristic matches the real directory shape: `computeRemoteRelayDir` builds `$HOME/.orca-remote/relay-<fullVersion>` (`ssh-relay-install-namespace.ts:40`), so trimming the last segment yields `$HOME/.orca-remote`, which already exists. The 12-byte hash is taken over the *full original path*, so per-version and per-target isolation survives the flattening into the shared parent — covered by the `keeps version isolation when shortening` test.

**06 Backward compatibility & data loss — pass.** Paths at or under the limit are returned byte-identical, so every currently-working host is unaffected. Hosts that take the new branch previously could not start the relay at all (EINVAL), so there is no working state to regress. No wire, schema, settings, or persisted-format change; no mobile-facing surface touched.

**07 Cross-platform & remoting — pass.** The Windows named-pipe branch is untouched and pinned by a new test asserting the full `\\.\pipe\orca-relay-<20 hex>` format. Darwin's 103 constant is separately measured above. WSL and container hosts use the POSIX branch and are covered by the Linux measurement.

---

### P0 — release blocking

None.

### P1 — releasable only with explicit acceptance

None.

### P2 — nice to have

**P2-1: the `/tmp` last-resort branch places a socket in a world-writable directory on shared hosts.** `ssh-relay-endpoints.ts:34` falls back to `/tmp/orca-<hash>-<sockName>` when even the shortened parent path does not fit. That branch is reachable at realistic `$HOME` lengths — measured with the real function:

```
HOME=45b -> /var/www/.../.orca-remote/49e44f997327-relay-...sock   (shortened parent)
HOME=53b -> /home/u/xxx.../.orca-remote/b3f5d019bc73-relay-...sock (shortened parent)
HOME=64b -> /tmp/orca-a124286362e6-relay-3f49cb06c8af7d25.sock     (/tmp branch)
```

so any remote `$HOME` of 54+ bytes (plausible for nested container/overlay paths) lands in `/tmp`. The endpoint hash is derived from a non-secret path, and the handshake has the *client* send `endpointCredential` for the *server* to verify (`relay-handshake.ts:144-151`). A hostile local user on the same remote host who pre-binds that predictable path would therefore receive the client's credential. This requires a shared multi-user remote host, a `$HOME` over 53 bytes, and winning the bind race — narrow, and strictly better than today's behaviour on such hosts (which is a hard outage). Worth a follow-up to prefer `$XDG_RUNTIME_DIR` (or a `0700` per-user `/tmp` subdirectory) before falling back to bare `/tmp`.

**P2-2: the final `/tmp/orca-<hash>.sock` fallback is unreachable.** `ssh-relay-endpoints.ts:35` guards `tmpPath` against the limit and falls back again. On POSIX `sockName` is at most `relay-<16 hex>.sock` = 27 bytes (`ssh-relay-instance-id.ts:8`), so `tmpPath` is at most 50 bytes and the guard can never fail. Harmless as defence-in-depth, but note that if it ever *were* reached, the reset globs in `ssh-relay-reset.ts:16` match `/tmp/orca-*-$sock_name` and would not match `/tmp/orca-<hash>.sock`, leaving that socket unreapable.

**P2-3: `/tmp` cleanup is nested under the `$HOME` guard.** In `ssh-relay-reset.ts`, the `/tmp/orca-*-$sock_name` glob sits inside `if [ -d "$base" ]`, so if `~/.orca-remote` has been removed the `/tmp` sockets are never reaped. In practice `base` exists whenever a relay was deployed, so this is hygiene rather than a live defect.

---

### Summary

This PR fixes a hard outage: on any SSH host whose `$HOME` pushes the relay socket path past the kernel's `sun_path` limit, the relay could not start at all, failing with an opaque `listen EINVAL`. The fix measures the path in UTF-8 bytes and, only when it is too long, relocates the socket to a shorter path whose name embeds a hash of the original — preserving the per-version and per-target isolation that the directory layout used to provide.

The verdict is driven by three things. The failure and the fix were reproduced end-to-end with the real code path on the exact host shape from #10726 (111 bytes → 99 bytes). The byte limits were measured by binding real sockets on real Linux and macOS rather than trusted from a constant, confirming 107/103 are correct and conservative. And the widened force-stop globs were verified against actual bound sockets at all four layouts, so the relocated sockets remain reapable. Paths under the limit are returned unchanged, so no host that works today can regress.

**Actions required before release: none.** All CI checks pass, the PR is `MERGEABLE`/`CLEAN`, and no P0 or P1 issue was found. The three P2 items — preferring `$XDG_RUNTIME_DIR` over world-writable `/tmp`, the unreachable final fallback, and the `/tmp` cleanup nesting — are follow-up hardening and do not gate this merge.

<sub>Reviewed against `.claude/skills/readiness-checklist`. Linux evidence gathered on the `openclaw` SSH host; local gates run on macOS in the PR worktree at `aa37252e9be`.</sub>

PR #10781, opened 2026-07-26
perf(git): read both diff blobs concurrently

nwparker:
## Summary

The file-diff loaders in `src/main/git/status.ts` awaited their two blob reads **in series**, so the second `git show` could not start until the first had fully returned. The two reads are independent, so that serialization was pure added latency on every diff the review panel opens.

Four call sites now issue both reads concurrently:

- `loadBranchDiff` — merge-base vs HEAD
- `loadCommitDiff` — parent vs commit (a root commit has no parent to read, so that side resolves without a spawn)
- `loadDiff` staged — HEAD vs index
- `loadDiff` unstaged — the left chain vs the working-tree read

## ELI5

To show you a diff, Orca has to fetch two versions of the file: the "before" and the "after". Each fetch launches a separate `git` process.

It was asking for the before, waiting for the whole thing to come back, and only *then* asking for the after — like ordering a coffee, waiting for it, and only then ordering the tea. The two orders have nothing to do with each other. Now it places both at once and waits for whichever finishes last. **Roughly twice as fast, for every diff you open.**

## Screenshots

No visual change.

`node config/scripts/git-diff-blob-concurrency-benchmark.mjs` (new) — spawns the **real `git` binary** against this repo, so it measures actual process-launch cost rather than a model of it:

```
iterations=10 warmup=3 head=b168f6f10
                      file   sequential   concurrent   speedup      saved
                 status.ts      46.8 ms      24.3 ms     1.93x    22.5 ms
    agent-hook-listener.ts      47.0 ms      23.6 ms     2.00x    23.5 ms
              TaskPage.tsx      49.2 ms      26.1 ms     1.89x    23.1 ms
```

**~23 ms saved per diff opened.** The saving is dominated by process launch rather than file size, which is why it holds roughly constant across a 1,500-line and a 12,000-line file. The benchmark asserts both paths return identical bytes before timing, and skips any path that does not exist on both sides of the checkout.

**Scope note, stated honestly:** over SSH each diff is a *single* relay RPC (`SshGitProvider.getBranchDiff` / `getCommitDiff` / `getDiff` each issue one `mux.request`), and the two spawns run host-local inside the relay. So this halves **remote-host spawn time**, not network round trips. It is not an SSH latency fix and I don't want it read as one.

## Proof of zero regression

**Error semantics are provably unchanged.** I checked before touching anything: `readGitBlobAtOidPath`, `readGitBlobAtIndexPath`, `readUnstagedLeftBlob`, and `readWorkingTreeFile` all catch internally and return a result object — none of them reject. So `Promise.all`'s fail-fast behaviour cannot differ from sequential `await`, and the surrounding `try`/`catch` fallback is reached in exactly the same cases as before.

**One data dependency was deliberately left sequential.** `readUnstagedLeftBlob` chains index → HEAD (it only reads HEAD when the index blob is absent), so its two steps *cannot* be parallelized. That chain stays as-is; only the independent working-tree read runs alongside it. Parallelizing that would have been a correctness bug, and it's the one place in this diff where the obvious transform is wrong.

**A test caught a real behavioural change**, which is the part worth reviewing:

`coalesces concurrent identical staged diff reads while in flight` failed with *"expected 1 call, got 2"*. That test asserted the **sequential shape** — one spawn, wait for it to resolve, then the second — which is an implementation detail, not the contract. The contract is coalescing: eight identical concurrent reads must collapse to two spawns, one per side, not sixteen.

I updated the test to assert that contract instead, and then **verified it still catches a broken coalescer**: bypassing `gitDiffReadDedupe` entirely makes it fail. So the dedupe guarantee is still pinned, and the test now also pins that both sides are issued concurrently.

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — **601/601** in `src/main/git/`; full suite 4 failures / 36,927, all pre-existing (below)
- [x] Added a re-runnable benchmark; updated the one test whose assumption the change invalidated

`pnpm test` is already red on `main`. This run's failures are `agent-exec-handler.test.ts` ×2 (assert against the machine's real `process.env`) and `terminal-history-incremental-restore.test.ts` ×2. All four are known members of the load-flaky pool this series has been tracking — the daemon test passes **16/16 in isolation** and imports nothing from `git/status`. Across seven full-suite runs, two consecutive runs on *unmodified* `main` produced different failure sets, which is what establishes the pool as pre-existing rather than change-induced.

**2 review loops: independent dependency/error audit and final risk review. Final verdict: ship.**

Found by the discovery sweep (ranked #2 of 35 candidates), then verified and measured independently before implementation. The riskiest parts — error semantics and the index→HEAD dependency — were checked by reading each callee rather than assumed.

Reviewer attention is most useful on **the changed test**: I claim the sequential-shape assertion was incidental and the coalescing assertion is the real contract. If that reading is wrong, this needs rethinking.

## AI Review Report

Risks reviewed:

- **Fail-fast divergence.** `Promise.all` rejects on first rejection; sequential `await` would have thrown at the same point. Moot here since no callee rejects — verified by reading all four.
- **Partial-state on error.** The staged branch previously assigned `originalContent` before the right-side read, so a throw between them left the left side populated. Restructuring puts both assignments after the join; since neither read rejects, this is unreachable, but the shape is now explicit rather than accidental.
- **Concurrency against the in-flight dedupe.** `getDiff` registers with `gitDiffReadDedupe` synchronously before any await, so coalescing is unaffected by what happens inside `loadDiff` — confirmed by the test still failing when the dedupe is bypassed.
- **Doubling concurrent git spawns.** Two spawns per diff instead of one-at-a-time. Diffs are opened per user action, not in a loop, and the existing in-flight dedupe already prevents identical concurrent reads from multiplying.

Cross-platform: no path handling changed. The existing `filePath.replace(/\\/g, '/')` normalization for Git's `<oid>:<path>` syntax on Windows is untouched. Works identically for local, WSL, SSH, and remote-runtime hosts, and for folder workspaces as well as git worktrees. No new Git subcommand or option is used — same `git show` invocations as before, so the Git 2.25 baseline is unaffected. Provider-agnostic: this is plumbing beneath any GitHub/GitLab distinction.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. The same `git show --end-of-options <oid>:<path>` commands run with the same arguments and the same `MAX_GIT_SHOW_BYTES` buffer cap — only their scheduling changed. `--end-of-options` is preserved, so a filename resembling a Git option still cannot be misparsed. No dependency changes.

### DIFF
--- a/config/scripts/git-diff-blob-concurrency-benchmark.mjs
+++ b/config/scripts/git-diff-blob-concurrency-benchmark.mjs
@@ -15,14 +15,14 @@ import { performance } from 'node:perf_hooks'
 import { fileURLToPath } from 'node:url'
 
 const REPO_ROOT = fileURLToPath(new URL('../..', import.meta.url))
-const ITERATIONS = Number.parseInt(process.env.ORCA_DIFF_BLOB_BENCH_ITERATIONS ?? '10', 10)
-const WARMUP = Number.parseInt(process.env.ORCA_DIFF_BLOB_BENCH_WARMUP ?? '3', 10)
+const ITERATIONS = Number(process.env.ORCA_DIFF_BLOB_BENCH_ITERATIONS ?? '10')
+const WARMUP = Number(process.env.ORCA_DIFF_BLOB_BENCH_WARMUP ?? '3')
 
 for (const [name, value] of [
   ['ORCA_DIFF_BLOB_BENCH_ITERATIONS', ITERATIONS],
   ['ORCA_DIFF_BLOB_BENCH_WARMUP', WARMUP]
 ]) {
-  if (!Number.isInteger(value) || value <= 0) {
+  if (!Number.isSafeInteger(value) || value <= 0) {
     throw new Error(`${name} must be a positive integer, received ${value}`)
   }
 }
@@ -51,15 +51,32 @@ async function readConcurrent(leftRef, rightRef, filePath) {
   return left.length + right.length
 }
 
-async function measure(fn, leftRef, rightRef, filePath) {
+// Why interleaved: running one strategy's whole batch before the other's lets
+// cache warming, CPU-frequency drift, and background load correlate with the
+// strategy being measured. Alternating per iteration and taking medians keeps
+// that drift common to both arms.
+async function measureInterleaved(leftRef, rightRef, filePath) {
   for (let index = 0; index < WARMUP; index += 1) {
-    await fn(leftRef, rightRef, filePath)
+    await readSequential(leftRef, rightRef, filePath)
+    await readConcurrent(leftRef, rightRef, filePath)
   }
-  const start = performance.now()
+  const sequentialSamples = []
+  const concurrentSamples = []
   for (let index = 0; index < ITERATIONS; index += 1) {
-    await fn(leftRef, rightRef, filePath)
+    // Alternate which arm goes first so neither systematically pays a cold cache.
+    const sequentialFirst = index % 2 === 0
+    for (const runSequential of sequentialFirst ? [true, false] : [false, true]) {
+      const start = performance.now()
+      await (runSequential ? readSequential : readConcurrent)(leftRef, rightRef, filePath)
+      ;(runSequential ? sequentialSamples : concurrentSamples).push(performance.now() - start)
+    }
   }
-  return (performance.now() - start) / ITERATIONS
+  const median = (samples) => {
+    const sorted = [...samples].sort((a, b) => a - b)
+    const middle = Math.floor(sorted.length / 2)
+    return sorted.length % 2 === 0 ? (sorted[middle - 1] + sorted[middle]) / 2 : sorted[middle]
+  }
+  return { sequential: median(sequentialSamples), concurrent: median(concurrentSamples) }
 }
 
 const head = (await git(['rev-parse', 'HEAD'])).trim()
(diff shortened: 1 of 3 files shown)


nwparker:
## Feedback applied — both suggestions were valid, neither changed the result

**Interleaved measurement (the substantive one).** Running all sequential samples before all concurrent ones lets cache warming, CPU-frequency drift, and background load correlate with the strategy being measured. Now the arms alternate per iteration, *which arm goes first* also alternates so neither systematically eats a cold cache, and both report medians.

The result holds:

| file | before (batched) | **after (interleaved)** |
|---|---|---|
| status.ts | 1.93× | **1.95×** |
| agent-hook-listener.ts | 2.00× | **2.03×** |
| TaskPage.tsx | 1.89× | **1.90×** |

~24 ms saved per diff either way. So this was methodological hygiene rather than a correction — but worth having, because it means the number is now robust to exactly the confound that would have made it suspect.

**Strict settings validation.** Fair: `Number.parseInt('10foo')` is `10` and `Number.parseInt('3.5')` is `3`, so the positive-integer contract wasn't enforced. Switched to `Number` + `Number.isSafeInteger`.

Rebased onto `main`.

Context worth flagging: a sibling PR in this series (#10790) was **withdrawn** after verification showed its win didn't survive contact with real data. This one is the opposite case — the number got *slightly better* under a stricter method, and the mechanism (two independent `git show` spawns awaited in series) is structural rather than distributional.

Issue #8956, opened 2026-07-16
[Bug]: Fail to list remote dirs


zetxtech:
### Operating system

Windows

### Orca version

1.4.141.0

### Details

<img width="670" height="675" alt="Image" src="https://github.com/user-attachments/assets/52386a82-47f3-436e-b1f9-c21c943ce5e9" />

This is a linux remote server, but:

> Error invoking remote method 'ssh:browseDir': RemoteBrowseError: zsh:1: command not found: powershell.exe

This problem occurs randomly and has a high recurrence rate.

OrcaWin:
## Agent reproduction package (Windows orchestration 2026-07-16)

**Host context:** Windows 11 / Orca **1.4.143.0** / repo path used: `docs/bug-reproductions/`  
**Labels applied:** see issue labels (`has_repro` or `cannot_repro`).  
**Security:** Do **not** run user-uploaded binaries from issue attachments. Recreate fixtures from the steps/scripts below.

---
# Issue 8956 — Fail to list remote dirs

**Status:** `needs_setup`  
**Host:** Windows 11, Orca 1.4.143, repo main  
**Label:** none (not conclusive for e2e)

## Summary

Reporter (Orca 1.4.141) gets:

```text
Error invoking remote method 'ssh:browseDir': RemoteBrowseError: zsh:1: command not found: powershell.exe
```

when browsing a **Linux** remote over SSH. Occurs randomly with high recurrence.

## Environment gates (this host)

| Need | Present |
|------|---------|
| Linux SSH remote | **No** (no `~/.ssh/config`, empty `orca environment list`) |
| Local unit tests for handler | **Yes** |

## What we verified

### Source (`src/main/ipc/ssh-browse.ts`)

1. Always try POSIX: `cd <path> && pwd && command ls -1Ap`
2. On any `RemoteBrowseError`, try Windows PowerShell fallback (`powershell.exe ...`)
3. If PowerShell exits **127** (`command not found`), surface the **original POSIX** error via `isPosixCommandNotFound` — not the PowerShell message

### Unit tests (`src/main/ipc/ssh-browse.test.ts`)

```powershell
pnpm exec vitest run src/main/ipc/ssh-browse.test.ts
# 14/14 passed on 2026-07-16
```

Includes regression: *surfaces the original POSIX failure when the PowerShell retry shows the host is not Windows* (zsh/bash `powershell.exe: command not found` → original error).

### Implication

[...]

PR #10785, opened 2026-07-26
Fix SSH directory browsing fallback on POSIX hosts

luizeof:
## Summary

- mark successful POSIX shell startup before running the remote `cd`/`pwd`/`ls` sequence
- preserve POSIX directory-listing failures instead of retrying them through `powershell.exe`
- remove the internal shell marker from user-visible diagnostics
- recognize zsh `powershell.exe` command-not-found failures even when the SSH server reports exit code 1 instead of 127

## Root cause

The remote directory browser first executes a POSIX listing command and falls back to PowerShell for Windows OpenSSH hosts. The fallback was triggered for every `RemoteBrowseError`, including ordinary failures that occurred after a POSIX shell had already started, such as an inaccessible or invalid directory on a Debian host.

As a result, a genuine POSIX error could be replaced by a misleading message such as:

CleanShot 2026-07-26 at 17 52 32

```text
zsh:1: command not found: powershell.exe
```

This made a macOS client connected to a Debian host appear to depend on PowerShell even though no Windows system was involved.

## Implementation

The POSIX browse command now writes a private marker to stderr before changing directories or invoking `ls`. If the listing later exits unsuccessfully and the marker was observed, the handler knows that the POSIX command dialect executed successfully and returns the original POSIX failure without attempting the Windows fallback.

The marker is stripped before diagnostics are surfaced, so users continue to receive only the actionable remote error. The PowerShell fallback remains available when the POSIX command wrapper never starts, preserving Windows OpenSSH support. The fallback error handling also accounts for SSH servers that rewrite command-not-found exit status 127 to 1.

## User impact

On macOS/Linux clients connected to POSIX SSH hosts, directory browsing no longer invokes `powershell.exe` after a normal remote path or permission failure. Users receive the real Debian/Linux/macOS error instead of an unrelated Windows-shell error. Windows SSH directory browsing retains its existing fallback behavior.

## Validation

- `pnpm exec vitest run --config config/vitest.config.ts src/main/ipc/ssh-browse.test.ts` — 14 tests passed
- `pnpm exec oxlint src/main/ipc/ssh-browse.ts src/main/ipc/ssh-browse.test.ts`
- `pnpm exec oxfmt --check src/main/ipc/ssh-browse.ts src/main/ipc/ssh-browse.test.ts`
- `pnpm run typecheck:node`
- `git diff --check`

The regression coverage verifies that a confirmed POSIX-shell failure does not execute PowerShell and that the internal marker is absent from the surfaced error.

## ELI5

Browsing remote folders over SSH on Linux/macOS hosts sometimes failed because Orca mistakenly tried a PowerShell fallback after a real POSIX error. It now treats POSIX listing failures correctly and keeps error messages cleaner.

### DIFF
--- a/src/main/ipc/ssh-browse.test.ts
+++ b/src/main/ipc/ssh-browse.test.ts
@@ -68,7 +68,9 @@ describe('registerSshBrowseHandler', () => {
         { name: 'README.md', isDirectory: false }
       ]
     })
-    expect(exec).toHaveBeenCalledWith('cd "$HOME" && pwd && command ls -1Ap')
+    expect(exec).toHaveBeenCalledWith(
+      `printf '%s\\n' '__ORCA_POSIX_BROWSE__' >&2 && cd "$HOME" && pwd && command ls -1Ap`
+    )
     expect(channel.listenerCount('data')).toBe(0)
     expect(channel.listenerCount('exit')).toBe(0)
     expect(channel.listenerCount('close')).toBe(0)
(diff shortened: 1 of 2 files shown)


Jinwoo-H:
Thanks @luizeof for the POSIX SSH directory-browsing diagnosis and marker/error-path tests. We are taking ownership through existing issue #8956, which tracks the same remote-directory listing fallback failure. We will reimplement on current main; no contributor rebase, retest, or other follow-up is requested. Closing this superseded branch while preserving your credit.

PR #10787, opened 2026-07-26
perf(mobile-sync): memoize the agent-status projection per entry

nwparker:
## Summary

`buildRuntimeMobileAgentStatusProjection` runs on the **App.tsx global store subscriber** — it fires on every store mutation. It re-serialized **every live agent** on **every status ping**.

`setAgentStatus` replaces one entry and re-spreads `agentStatusByPaneKey`, so the new map reference defeats the reference-equality skip gate in `canSkipRuntimeMobileSessionSyncKeyBuild`. Each ping then paid to serialize every *other* agent — prompt, 20-entry `stateHistory`, `toolInput`, and an 8 KB-capped `lastAssistantMessage` — only to discover none of them had changed.

This was the only uncached projection on that subscriber path. `cachedTabsProjection`, a few functions above in the same file, already uses the identity-keyed pattern applied here.

## ELI5

Orca keeps a compact text summary of every running agent so it can tell the mobile app what changed.

Every time *one* agent reports progress, it rebuilt the summary for **all** of them from scratch — re-copying each one's full prompt, its last 20 steps, and up to 8 KB of its most recent message. With eight agents running, seven of those were rewritten character-for-character into exactly what they already said.

Now each agent's summary is remembered until that agent actually changes. A ping rewrites one entry instead of all of them.

## Screenshots

No visual change.

`node config/scripts/mobile-agent-status-projection-benchmark.mjs` (new):

```
bucket=30000ms iterations=400 warmup=60 (median of 5 rounds)
  agents        full      cached   speedup
       3   0.0690 ms   0.0117 ms      5.9x
       8   0.1813 ms   0.0145 ms     12.5x
      20   0.4540 ms   0.0425 ms     10.7x
      40   0.8813 ms   0.0630 ms     14.0x
```

The benchmark reproduces the real trigger — one entry replaced, the map re-spread, every other entry reference-identical, which is exactly what `setAgentStatus` produces. It asserts both versions emit identical output before timing, and re-reads `AGENT_STATUS_SYNC_UPDATED_AT_BUCKET_MS` from the real module so a drifted constant fails loudly instead of quietly changing what is measured.

**The cost scales with agents running in parallel** — the workload this app exists for.

## Proof of zero regression

The output is **byte-identical**. That is the whole safety argument, and it needed proving because the change swaps `JSON.stringify(array)` for `'[' + parts.join(',') + ']'`.

First, that the two forms agree at all: verified across 17 shapes including the cases where `JSON.stringify` behaves specially — `undefined`, `null`, `NaN`, `Infinity`, `-0`, `1e21`, lone surrogates, `Date`, a custom `toJSON`, and arrays containing functions/symbols. Zero mismatches.

Then, that the *real* function agrees with the *old* function: the new test keeps a verbatim copy of the pre-change whole-array serialization and asserts equality across empty / single / 12-agent / all-optionals-absent / out-of-sort-order maps, plus four rounds of cache reuse.

**Mutation testing.** Two findings, both in my own tests:

| mutant | result |
|---|---|
| reuse a cached row without checking entry identity | **fails** |
| fast path ignores the source map identity | **fails** |
| join with `;` instead of `,` | **fails** |
| drop the `localeCompare` sort | **fails** |

The stale-entry-reuse mutant — the single most important one, since cache invalidation is the entire risk this change introduces — **initially survived**. My cache-reuse rounds nudged `updatedAt` by 5000 ms, but the bucket is 30000 ms, so the projection didn't change and a broken cache looked correct. I made each round's change observable (state, prompt, message, and a full-bucket time step) and it now dies.

Separately, my first version of the test failed against correct code because I hardcoded `BUCKET_MS = 1000` instead of the real `30_000`. The test now imports the constant so it cannot drift.

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — **509/509** across `src/renderer/src/runtime/`; full suite **2 failures / 36,935**, both pre-existing

`pnpm test` is already red on `main`. This run shows **only** `agent-exec-handler.test.ts` ×2, which assert against the machine's real `process.env` — a strict subset of the known baseline pool, and neither imports the changed code. Across eight full-suite runs in this series, two consecutive runs on *unmodified* `main` produced different failure sets, which is what establishes the pool as pre-existing.

**3 review loops: independent re-measurement, mutation-driven revision, and final risk review. Final verdict: ship.**

Found by the round-2 discovery sweep, then re-measured independently by me before implementation (5.5×-12.5× on my own harness, matching the finder's numbers), followed by the mutation loop above which drove two test revisions.

The same sweep **rejected** a neighbouring candidate with numbers rather than reporting it: `resolvePaneKey`'s O(all tabs) nested scan in `useIpcEvents.ts` runs per `agentStatus:set` event and looks like the same class of problem, but measures **0.26 µs at 30 tabs and 0.60 µs at 96 tabs**. An index would save well under a microsecond. Not worth a PR.

Reviewer attention is most useful on **the join-vs-stringify equivalence**. If there is an entry shape reachable in practice where per-row serialization diverges from whole-array serialization, that is the bug this change could have.

## AI Review Report

Risks reviewed:

- **Cache invalidation.** Three independent guards: the map identity fast path, per-entry object identity, and re-sorting keys on every rebuild. Each is pinned by a mutant that fails without it.
- **Key ordering.** The `localeCompare` sort is preserved and pinned by an out-of-insertion-order fixture, so a map whose keys arrive in a different order still yields the same projection — important because this value is compared for equality to decide whether to republish.
- **Unbounded cache growth.** The cache is fully rebuilt per call and keyed to a single `source` map, so entries for removed panes are dropped on the next build rather than accumulating. No eviction policy needed.
- **Optional-field handling.** The `?? null` fallbacks are unchanged and covered by an all-optionals-absent fixture.

Cross-platform: no path, shell, keyboard, or platform-dependent behaviour touched — this is pure in-memory serialization in the renderer. Identical on macOS/Linux/Windows and for local, SSH, WSL, and remote-runtime hosts; the projection feeds the mobile sync graph the same way on every host.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. The projection contains the same fields as before and is used the same way — as a change-detection key for the mobile republish. Agent prompts and assistant messages were already included and remain subject to the same 8 KB cap; nothing newly enters or leaves the payload. The cache holds references to store objects the renderer already retains, adds no persistence, and cannot outlive the store map it is keyed to. No dependency changes.

### DIFF
--- a/config/scripts/mobile-agent-status-projection-benchmark.mjs
+++ b/config/scripts/mobile-agent-status-projection-benchmark.mjs
@@ -43,8 +43,8 @@ for (const [name, value] of [
   }
 }
 
-function serializeEntry(paneKey, entry) {
-  return JSON.stringify({
+function toRow(paneKey, entry) {
+  return {
     paneKey,
     entryPaneKey: entry.paneKey,
     state: entry.state,
@@ -64,15 +64,20 @@ function serializeEntry(paneKey, entry) {
     interactivePrompt: entry.interactivePrompt ?? null,
     lastAssistantMessage: entry.lastAssistantMessage ?? null,
     interrupted: entry.interrupted ?? null
-  })
+  }
 }
 
-// Pre-fix: serialize the whole map every time.
+function serializeEntry(paneKey, entry) {
+  return JSON.stringify(toRow(paneKey, entry))
+}
+
+// Pre-fix: build plain rows and stringify the array once — no per-row roundtrip,
+// which the original never paid and which would inflate the reported speedup.
 function buildFull(map) {
   return JSON.stringify(
     Object.entries(map)
       .sort(([a], [b]) => a.localeCompare(b))
-      .map(([paneKey, entry]) => JSON.parse(serializeEntry(paneKey, entry)))
+      .map(([paneKey, entry]) => toRow(paneKey, entry))
   )
 }
 
@@ -169,6 +174,12 @@ for (const agents of [3, 8, 20, 40]) {
   const cachedBuilder = makeCachedBuilder()
   if (buildFull(map) !== cachedBuilder(map)) {
     throw new Error(`projection mismatch at ${agents} agents`)
+  }
+  // Why also after a ping: the cold call reuses nothing, so a stale-row bug would
+  // only surface once the cache is actually exercised.
+  const pinged = ping(map, 0)
+  if (buildFull(pinged) !== cachedBuilder(pinged)) {
+    throw new Error(`projection mismatch after a ping at ${agents} agents`)
   }
   const full = measure(buildFull, map)
   const cached = measure(makeCachedBuilder(), map)
--- a/config/scripts/mobile-agent-status-projection-benchmark.mjs
+++ b/config/scripts/mobile-agent-status-projection-benchmark.mjs
@@ -1,0 +1,181 @@
+#!/usr/bin/env node
+// Benchmark: cost of the mobile agent-status projection per store mutation.
+//
+// buildRuntimeMobileAgentStatusProjection runs on the App.tsx global store
+// subscriber. setAgentStatus replaces one entry and re-spreads
+// agentStatusByPaneKey, which defeats the reference-equality skip gate, so before
+// the fix EVERY live agent was re-serialized on EVERY status ping — each carrying
+// a prompt, a 20-entry stateHistory, toolInput, and an 8 KB-capped
+// lastAssistantMessage.
+//
+// The fix memoizes each row's JSON by entry identity, mirroring the
+// cachedTabsProjection pattern already in the same file, so a ping re-serializes
+// only the agent that actually changed.
+//
+// The bucket width is re-read from the real module so a drifted constant fails
+// loudly here instead of quietly changing what this benchmark measures.
+import { readFileSync } from 'node:fs'
+import { performance } from 'node:perf_hooks'
+import { fileURLToPath } from 'node:url'
+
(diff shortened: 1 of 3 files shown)


nwparker:
## Feedback applied — and the flagged issue was real

**The inflated-baseline finding was correct, and it's the important one.** My `buildFull` arm stringified each row and parsed it back before stringifying the array — a per-row roundtrip the pre-fix code never paid. That made the baseline artificially slow.

Corrected by sharing one `toRow` builder between both arms:

| agents | before (claimed) | **corrected** |
|---|---|---|
| 3 | 5.9× | **2.2×** |
| 8 | 12.5× | **4.3×** |
| 20 | 10.7× | **4.0×** |
| 40 | 14.0× | **4.6×** |

Honest range is **2.2×-4.6×**, not 5.9×-14.0×. Notably this now agrees with an independent probe I ran before implementing (2.56×-5.52×) — the divergence between those two numbers should have told me something was wrong with the committed benchmark, and I didn't chase it.

**Also applied:** the equivalence check now runs after a ping, not just on the cold call. That mattered — I verified it by injecting a stale-row bug (reuse the cached row regardless of entry identity), and the cold-path check passed it while the post-ping check catches it:

```
Error: projection mismatch after a ping at 3 agents
```

So the previous guard could not have caught the exact bug class this change introduces.

The optimization itself is unaffected — same code, same 4/4 mutants killed, byte-identical output. Only the benchmark's honesty changed. Rebased onto `main`.

PR #10788, opened 2026-07-26
perf(startup): stop parsing qrcode and @linear/sdk at launch

nwparker:
## Summary

Two packages sat in the main bundle's eager top-level require block despite being reachable only from features most users never touch.

**`@linear/sdk` is the sharper case.** `src/main/linear/linear-sdk.ts` exists *solely* to load that ~2.6 MB CJS bundle lazily — its own comment says "launch never parses it for the majority who never connect Linear." A single value import at `issue-relation-write.ts:1` defeated that loader for everyone. That file now imports the type and constructs through `loadLinearSdk()`.

**`qrcode`** is only reachable from mobile pairing. Both call sites were already `async`, so they take a dynamic import.

## ELI5

Starting the app means reading and compiling a pile of JavaScript before the window can appear. Two of those piles were for features you probably never use: Linear ticket integration, and the QR code for pairing your phone.

They were being loaded on *every* launch anyway. One of them already had a "load this only when needed" wrapper written for it — but one stray import elsewhere in the codebase reached past the wrapper and pulled it in regardless, so the wrapper never actually did anything.

Now both load the first time you use the feature, and not before.

## Screenshots

No visual change.

**Direct measurement of what is removed from every launch:**

```
qrcode        9.3 ms to require
@linear/sdk  23.9 ms to require
             ------
             33.2 ms of module parse, deleted from the startup path
```

**Where it shows up in the real startup bench** (`tools/benchmarks/startup-time-bench.mjs`, 2000-file fixture) — `spawnToAppReady` is the phase module parsing lives in:

| | baseline | lazy | delta |
|---|---|---|---|
| 5 iterations | 379 ms | 354 ms | **-25 ms** |
| 9 iterations | 376 ms | 357 ms | **-19 ms** |

**Being straight about the limits of this measurement:** `totalToWorkspaceReady` does *not* show a reliable win. It moved 749→730 ms at 5 iterations but 745→**760** ms at 9 — the sign flips, so end-to-end launch is noise-dominated at this sample size on this machine. I am claiming the ~20 ms `spawnToAppReady` improvement and the 33 ms of directly-measured parse work, not a 19 ms end-to-end win. Anyone re-running this should expect the total to be inside the noise band.

## Proof of zero regression

Both changes are import-site-only; no logic moved.

- **`@linear/sdk`**: `loadLinearSdk()` is the pre-existing loader already used by `client.ts`, and it caches its `require`. The type import is erased at compile time, so nothing pulls the module in eagerly. The one construction site is inside an `async` function that already awaited `acquire()`.
- **`qrcode`**: both consumers were already `async` and already `await`ed the QR call, so the added `await import(...)` introduces no new asynchrony to any caller. I verified no eager `from 'qrcode'` import remains in `src/main`, and that every remaining `QRCode.` reference resolves to the locally-bound dynamic import.

TypeScript is doing real work here: had I left a value import behind, `import type` would have failed the build, and the `no-import()-type` lint rule the loader's comment references still passes.

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — **120/120** across `src/main/linear/` and `src/main/ipc/mobile.test.ts`; full suite **2 failures / 36,934**, both pre-existing
- [x] `pnpm build:electron-vite` — built and launched repeatedly through the startup bench

`pnpm test` is already red on `main`. This run shows **only** `agent-exec-handler.test.ts` ×2, which assert against the machine's real `process.env` — a strict subset of the known baseline pool, and neither imports the changed code.

**2 review loops: independent import/A-B verification and final risk review. Final verdict: ship.**

Found by the round-2 discovery sweep; I verified both import sites and re-ran the A/B myself at two sample sizes rather than accept the reported end-to-end figure — which is how the noise in `totalToWorkspaceReady` surfaced.

A third startup candidate from the same sweep was **rejected on risk, not measurement**: deferring the 14 managed agent-hook installers with `setImmediate` measured a clean 14 ms → ~0 ms on the critical path, but managed hooks are config files that agent processes read *at their own launch*. Deferring races any restored terminal that spawns an agent early, and that agent would run a turn with no status hook — silent status loss. The surrounding code already treats install ordering as load-bearing (`index.ts` explicitly sequences `ensureRealHomeCodexHookState` before it). 14 ms of a ~750 ms launch does not justify that; a safe version needs a barrier the PTY spawn path awaits, which should be designed rather than bolted on.

## AI Review Report

Risks reviewed:

- **Deferred failure surface.** A lazy `require` can now fail at feature-use time rather than launch. For `@linear/sdk` this is unchanged in practice — `client.ts` already loaded it this way, so the failure mode already existed on that path. For `qrcode`, both call sites already sit inside `try`/`catch` that returns `null` on failure.
- **Packaging.** Neither module becomes unreachable to the bundler: `linear-sdk.ts` uses `createRequire` against the packaged app, and `await import('qrcode')` is a static specifier electron-vite resolves at build time. The built output was exercised by the startup bench across ~28 real launches.
- **Type-only import correctness.** `LinearClient` is used only in type position in `issue-relation-write.ts` after the change; the value comes from the loader's structurally-declared `LinearSdkModule`.

Cross-platform: no path, shell, or platform-dependent behaviour touched. Module resolution is identical on macOS/Linux/Windows, and neither package is platform-conditional. Unaffected by SSH/WSL/remote hosts — this is main-process module loading.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. Both modules are the same versions from the same lockfile, loaded through the same resolver, just later. `loadLinearSdk()` is the existing audited loader. Deferring `@linear/sdk` marginally *reduces* launch-time attack surface for users who never connect Linear, since its transitive dependency graph is no longer evaluated. No dependency changes.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -4,7 +4,6 @@ import { isAbsolute, join } from 'node:path'
 import os from 'node:os'
 import { app, BrowserWindow, dialog, ipcMain, nativeTheme, type Tray } from 'electron'
 import { electronApp, is } from '@electron-toolkit/utils'
-import * as QRCode from 'qrcode'
 import {
   Store,
   initDataPath,
@@ -1559,6 +1558,9 @@ function getBundledWebClientRoot(): string | undefined {
 }
 
 async function renderTerminalPairingQr(pairingUrl: string): Promise<string | null> {
+  // Why dynamic: qrcode is only reachable from mobile pairing, so launch should
+  // not parse it for the majority who never pair a device.
+  const QRCode = await import('qrcode')
   try {
     return await QRCode.toString(pairingUrl, { type: 'terminal', small: true })
   } catch {
--- a/src/main/ipc/mobile.ts
+++ b/src/main/ipc/mobile.ts
@@ -1,6 +1,5 @@
 import { app, ipcMain, shell, type IpcMainInvokeEvent } from 'electron'
 import { networkInterfaces } from 'node:os'
-import QRCode from 'qrcode'
 import type { RuntimeAccessGrant } from '../../shared/runtime-access-grants'
 import type { MobilePairingConnectionMode } from '../../shared/mobile-pairing-connection-mode'
 import { isTailnetIPv4Address } from '../../shared/tailnet-address'
@@ -137,6 +136,9 @@ export function registerMobileHandlers(
         return { available: false as const }
       }
 
+      // Why dynamic: pairing is the only consumer, so launch should not parse
+      // the qrcode bundle for users who never pair a device.
+      const { default: QRCode } = await import('qrcode')
       const qrDataUrl = await QRCode.toDataURL(offer.pairingUrl, {
         errorCorrectionLevel: 'M',
         margin: 2,
--- a/src/main/linear/issue-relation-write.ts
+++ b/src/main/linear/issue-relation-write.ts
@@ -1,4 +1,5 @@
-import { LinearClient } from '@linear/sdk'
+import type { LinearClient } from '@linear/sdk'
+import { loadLinearSdk } from './linear-sdk'
 import type {
   LinearIssueRelationship,
   LinearIssueRelationWriteResult
@@ -34,7 +35,7 @@ export async function writeIssueRelation(params: {
   await acquire()
   try {
     const client = params.signal
-      ? new LinearClient({ apiKey: entry.apiKey, signal: params.signal })
+      ? new (loadLinearSdk().LinearClient)({ apiKey: entry.apiKey, signal: params.signal })
       : entry.client
     const existing = await findExistingRelation(client, params)
     if (params.operation === 'add' && existing) {


nwparker:
## Correction: my qrcode number was overstated ~2-4x

An independent verifier re-measured both halves of this PR. The `@linear/sdk` half held up and is if anything the stronger claim; **the qrcode half was inflated by my methodology**, and I want that on the record before review.

### @linear/sdk — confirmed, ~19 ms

Verified by building both bundles from source and A/B'ing the real output, not by argument:

- Isolated `require` in real Electron main, 7 interleaved pairs: **18.2-21.2 ms** eager, **0.000 ms** lazy.
- Marginal cost with `graphql` required first (defeater check for shared subgraph): **17.5-18.3 ms** — the cost is genuinely marginal, not shared with another dep.
- **Most important:** replaying the bundle's *real* 49-entry eager require block, dropping only `@linear/sdk`, 9 interleaved pairs: medians **89.6 ms → 71.6 ms = ~18-19 ms**. This confirms the cost is not masked by other startup work.
- Structural: baseline bundle `grep -c 'require("@linear/sdk")'` = 1; fixed = 0. Sole consumer confirmed exhaustively.

### qrcode — real, but ~4.5 ms, not 9.3 ms

My figure came from an **isolated** `require`, which overcounts because it charges qrcode for dependencies the eager block already pulls. The right number is the **marginal** cost:

| method | result |
|---|---|
| paired in-process marginal (preload the ~29 shared deps, then time qrcode's *added* cost), n=9 | **4.51 ms**, sd 0.15 |
| isolated cold require, n=7 | 7.4 ms (matches my 9.3 ms in order of magnitude) |
| whole-bundle A/B, 11 interleaved pairs | 7.03 ms median / 3.32 ms mean, **t = 1.80 — not significant** |

Dep breakdown explains the inflation: qrcode → dijkstrajs, pngjs, yargs, with a per-file compile sum of 40.6 ms against 9.84 ms actual wall time. Inclusive attribution overcounts.

### Corrected claim for this PR

**~19 ms (@linear/sdk) + ~4.5 ms (qrcode) ≈ 23-24 ms off every cold launch**, not the 33 ms I stated. The 33 ms figure summed two isolated `require` costs; only the marginal cost is actually recoverable.

I'd already flagged in the original body that `totalToWorkspaceReady` was noise-dominated and that I was claiming the `spawnToAppReady` phase rather than end-to-end. That framing was right — but the component number inside it was still too high, and "isolated require cost" is the specific mistake. Marginal is the shippable number.

The change itself is unaffected: both imports are still eagerly parsed on `main` for features most users never touch, and `linear-sdk.ts` still exists solely to prevent exactly this.

**Suggested follow-up** (not in this PR): a lint rule or test asserting no value import of `@linear/sdk` outside `linear-sdk.ts`, so the lazy loader cannot be silently defeated again — which is how this regressed in the first place.

Issue #10273, opened 2026-07-24
[Bug]: Mobile Relay stays unavailable after a transient desktop broker failure


wohsj110:
### Operating system

macOS 15.6.1 (Apple Silicon)

### Orca version

v1.4.152

#### Short summary

Orca Mobile Relay remains unavailable and generated pairing codes are downgraded to LAN-only, even though the cloud auth endpoint, Relay director, and assigned Relay cell are reachable. Restarting the proxy and switching all `*.onorca.dev` traffic from a proxy route to `DIRECT` does not recover the Relay broker.

The UI shows:

> Relay could not be reached — this code only works on your local network. Regenerate to try again.

#### Reproduction

1. Sign in to an Orca cloud-linked profile.
2. Open Settings -> Mobile.
3. Select Orca Relay and generate or regenerate a pairing code.
4. Observe that Relay is marked unavailable and the QR code contains only the LAN WebSocket endpoint.
5. Regenerate again; the state remains unavailable.

#### Network diagnostics

The desktop uses ClashX Pro in TUN/Fake-IP mode.

Initially there was no Orca-specific rule, so Relay traffic matched the final proxy group. Clash logs showed successful TCP connections for the complete startup sequence:

- `login.onorca.dev:443`
- `relay.onorca.dev:443`
- `c2.relay.onorca.dev:443`

As an A/B test, this rule was inserted as the first rule and ClashX was restarted:

```yaml
- DOMAIN-SUFFIX,onorca.dev,DIRECT
```

Clash then confirmed the expected match:

```text
rule=DomainSuffix(onorca.dev) proxy=DIRECT
```

The Relay UI remained unavailable, so bypassing the proxy did not resolve the failure.

Additional checks:

- `https://relay.onorca.dev/health` returns HTTP 200.
- `https://c2.relay.onorca.dev/health` returns HTTP 200.
- The cell WebSocket control endpoint is reachable through TUN, an explicit proxy, and a bound direct interface; without credentials it returns the expected HTTP 401.
- The local Orca listener is healthy on both loopback and the LAN address.

This suggests the failure occurs after basic network/TLS reachability, during Relay broker setup or the control upgrade/host-proof phase.

#### Source correlation

The installed v1.4.152 code appears to have both failure modes addressed by recently merged changes:

- #10147 adds bounded retry after transient cloud session, token exchange, assignment, and control-open failures. In v1.4.152, the coordinator catch path only publishes `offline` and does not schedule another attempt.
[...]

PR #10789, opened 2026-07-26
fix(orca-cloud): report the real cause when cloud sign-in cannot reach the network

### LINKED ISSUE #10758: [Bug]: Windows Google OAuth callback succeeds but cloud session exchange fails with TypeError: fetch failed
### Operating system

Windows 11 x64

### Orca version

v1.4.156 (also reproduced on v1.4.150)

#### Short summary

Signing in to an Orca Cloud profile with Google completes in the system browser and successfully returns to Orca's local loopback callback, but the desktop then shows:

```text
fetch failed
```

The cloud profile is never linked and no encrypted account session is created. Updating from v1.4.150 to the current stable v1.4.156 did not change the behavior.

#### Reproduction

1. Start Orca on Windows.
2. Attempt to connect the active local profile to Orca Cloud.
3. Complete Google sign-in in the system browser.
4. The browser reaches Orca's `127.0.0.1` callback successfully and displays the successful-return page.
5. Return to Orca.
6. Orca reports `fetch failed`; the active profile remains local.

#### Expected behavior

After the loopback OAuth callback succeeds, Orca should exchange the authorization code for a cloud session, persist the encrypted session, and link the active profile.

#### Actual behavior

The post-callback session exchange fails with a bare `TypeError: fetch failed`. The UI does not expose an error cause or HTTP status, and the failure is not recorded with useful detail in `main.trace.ndjson` or `daemon.log`.

#### Diagnostics already performed

* Orca runtime is healthy and reachable.
* The browser successfully reaches Orca's loopback callback, so the local callback listener and OAuth redirect work.
* `login.onorca.dev` and `relay.onorca.dev` resolve and are reachable over HTTPS.
* A diagnostic Node.js request using the same relevant options to:

  ```text
  POST https://login.onorca.dev/v1/desktop/auth/session
  ```

  reaches the service and receives the expected response for a deliberately invalid code:

  ```http
  401 Unauthorized
  {"error":"invalid_code"}
  ```
* No HTTP(S) proxy environment variables are configured.
* WinHTTP is configured for direct access.
* Windows Firewall has no outbound block rule for Orca; default outbound traffic is allowed.
* Only Windows Defender is active; no third-party antivirus/network filter was detected.
* Windows Time is synchronized via `time.windows.com`; an HTTP `Date` comparison showed approximately 0.5 seconds of skew.
* The active profile remains `kind: "local"`, has no cloud metadata, and no `account-session.json.enc` is created.
* Fully quitting and restarting Orca does not resolve the issue.

This appears to fail after `beginOrcaCloudPkceFlow(...)` resolves successfully, during the subsequent authorization-code exchange against `/v1/desktop/auth/session`.

#### Source correlation

The shipped v1.4.156 main-process bundle still sends this exchange through bare global `fetch(...)` with `redirect: "error"` and a 30-second timeout. The surrounding catch path reduces the error to `error.message`, so users only see `fetch failed` and the underlying cause (`ECONNRESET`, TLS, redirect, socket error, etc.) is lost.

A similar transport/pool pattern was reported for another integration in stablyai/orca#5593, although this issue affects the first-party Orca Cloud session exchange and reproduces across a full app restart.

#### Suggested improvements

1. Log the full error name, stack, and nested `error.cause` for Orca Cloud session exchange failures, without logging the authorization code, verifier, nonce, state, or tokens.
2. Surface a bounded failure category in the UI: DNS, TLS, timeout, redirect, connection reset, or HTTP status.
3. Consider using `electron.net.fetch` for this first-party request, or retry once on recognized transient Undici transport failures with a fresh connection/dispatcher.
4. If the backend closes the connection only for otherwise-valid authorization codes, return a structured HTTP error instead so the client does not collapse it into `fetch failed`.

#### Related but distinct reports

* stablyai/orca#10273 concerns Relay broker recovery after cloud sign-in already succeeded.
* stablyai/orca#10401 concerns Relay host-proof clock skew after cloud sign-in.

This report fails earlier, while creating the initial Orca Cloud session.

OrcaWin:
Fixes #10758

## Summary

Signing in to Orca Cloud on Windows completed in the browser, returned to the loopback callback, and then failed with a bare `fetch failed`. The profile stayed local and nothing in `main.trace.ndjson` explained why.

The desktop session exchange (`POST /v1/desktop/auth/session`) ran through bare global `fetch`. Every transport-level failure — DNS, TLS interception, a refused redirect, a reset socket — surfaces from undici as the identical string `TypeError: fetch failed`, with the real diagnosis buried in `error.cause`. `connectCurrentOrcaProfile` reduces the error to `error.message` (`profile-cloud-service.ts:91`) and the renderer prints that verbatim as the toast description, so `cause` was discarded before it reached the user *or* the trace file.

This PR:

1. **Routes both cloud clients through Electron `net.fetch`** behind `ensureElectronProxyFromEnvironment`, via a new `orcaCloudFetch` in `profile-cloud-transport.ts`. This matches the existing precedent in `jira/client.ts`, `claude-accounts/oauth-refresh.ts`, and `rate-limits/claude-fetcher.ts`. Chromium's stack honors the OS certificate store and proxy configuration and keeps its own socket pool, so Windows TLS interception and undici's stale keep-alive sockets stop being silent failure modes.
2. **Classifies transport failures into a bounded category** — `dns`, `tls`, `proxy`, `offline`, `timeout`, `redirect`, `connection-lost`, `unknown` — and surfaces it with the errno / `net::` detail. The original error chain is retained as `cause` and recorded on the span (`orcaCloud.operation`, `transportFailure`, `transportErrorName`, `transportErrorDetail`), so `main.trace.ndjson` finally has something actionable.

This addresses suggestions 1–3 from the issue. Suggestion 4 (backend returning a structured error instead of closing the connection) is server-side.

## ELI5

Orca asks the sign-in server "here's my one-time code, give me a session." That question was being asked by a phone line that only knows how to say one thing when anything goes wrong: **"it didn't work."**

Not *why*. The address book couldn't find the server, the security certificate looked wrong, a corporate proxy hung up, the server tried to forward us somewhere we refuse to follow — all of it came out as the same three words. The detail *did* exist, tucked inside the error, but the code threw it away before showing anything to the user.

Two changes. First, Orca now uses **the same phone line the rest of the app uses** (Electron's network stack) — the one that knows about Windows' certificate store, corporate proxies, and doesn't reuse stale connections. Second, when a call fails, Orca now **reads the detail before throwing it away** and tells you which of the seven things went wrong.

So `fetch failed` becomes `Could not reach Orca Cloud: the TLS handshake failed — a proxy or security tool may be intercepting HTTPS (net::ERR_CERT_AUTHORITY_INVALID)`.

The bug wasn't that sign-in broke. Sign-in was *always* going to break on that machine for some environmental reason. The bug is that Orca made it impossible to find out which one.

## Screenshots

No visual change. The failure toast already renders `result.error` as its description, so it now carries the categorized message instead of `fetch failed` — no renderer changes were needed.

## Proof of fix

**1. Reproduction of the reported symptom.** Driving the shipped options (`redirect: 'error'`, 30s `AbortSignal.timeout`, bare global `fetch`) against real local servers that fail three structurally different ways:

```
connection reset mid-exchange -> {"uiMessage":"fetch failed","name":"TypeError","cause":"SocketError: other side closed","causeCode":"UND_ERR_SOCKET"}
redirect on the session endpoint -> {"uiMessage":"fetch failed","name":"TypeError","cause":"Error: unexpected redirect"}
DNS failure -> {"uiMessage":"fetch failed","name":"TypeError","cause":"Error: getaddrinfo ENOTFOUND ...","causeCode":"ENOTFOUND"}
```

Three different root causes, one indistinguishable user-facing string — exactly the report. Note the redirect case is *deterministic*, which fits "reproduces across a full app restart" better than a stale-socket theory does.

**2. Same three scenarios, before vs. after, through the production code path:**

```
### connection reset mid-exchange
BEFORE: fetch failed
AFTER  [connection-lost]: Could not reach Orca Cloud: the connection closed before a response arrived (UND_ERR_SOCKET: other side closed)

### redirect on the session endpoint
BEFORE: fetch failed
AFTER  [redirect]: Could not reach Orca Cloud: Orca Cloud redirected the request, and Orca never follows redirects on token endpoints (unexpected redirect)

### DNS failure
BEFORE: fetch failed
AFTER  [dns]: Could not reach Orca Cloud: the Orca Cloud hostname could not be resolved (ENOTFOUND: getaddrinfo ENOTFOUND login.invalid-host-10758.test)
```

**3. The new regression test fails on the shipped code.** Reverting only `profile-cloud-client.ts` to its pre-PR state and re-running:

```
FAIL  profile-cloud-transport.test.ts > reports a bounded category and never echoes the authorization code
AssertionError: expected TypeError: fetch failed to be an instance of OrcaCloudTransportError
 Tests  1 failed | 13 passed (14)
```

With the fix restored, all 14 pass. That test drives the real `exchangeOrcaCloudAuthCode` against a live socket that dies mid-request, so it asserts against a genuine undici error chain rather than a hand-built stand-in.

**4. The fix is present in the built bundle** — `out/main/index.js` contains `Could not reach Orca Cloud` and `orcaCloud.transportFailure` after `pnpm build:electron-vite`.

## Proof of no regression

The suite has substantial pre-existing failures on this Windows machine, so every number below is a **before/after comparison against the same tree**, not an absolute count.

| Scope | Pre-fix baseline | With this PR | Delta |
|---|---|---|---|
| `src/main` (1115 files) | 31 files / 96 tests failed | 31 files / 96 tests failed — identical file list, passing 15111 → 15119 | **0** |
| 11 failing files outside `src/main` | 11 files / 29 tests failed | 11 files / 29 tests failed | **0** |
| Directly affected (`orca-profiles`, profile IPC, relay, fetch audit) | — | 21 files / 115 tests **passed** | — |

Full suite: `Test Files 44 failed | 3387 passed | 25 skipped`, `Tests 126 failed | 36325 passed | 446 skipped`. Every failing file was enumerated and re-run against the pre-fix parent commit; all fail identically. They are environment-dependent (symlinks, `/bin/sh` spawn, filesystem watchers, PTY, host-boundary renderer tests) and none live in `orca-profiles`. The two suite-level "Errors" originate in `src/main/ipc/ephemeral-vm.test.ts` and `src/relay/pty-handler.test.ts`, both already in the baselined set.

Other gates:

- `pnpm typecheck` (node + cli + web) — **exit 0**
- `pnpm build:electron-vite` — **exit 0** (main, preload, renderer)
- `oxlint` — **exit 0**, zero errors
- `lint:switch-exhaustiveness`, `check:reliability-gates` (49 gates), `check:max-lines-ratchet`, `verify:bundled-skill-guides` — all pass
- `pnpm lint` overall fails at `verify:skill-bundle-manifest` (stale generated `resources/skills/*.json`). **Verified pre-existing** — the same check fails identically on a stashed clean tree, and this PR touches nothing under `resources/`.

## Testing

- [x] `pnpm lint` — passes except the pre-existing stale-skill-manifest step documented above
- [x] `pnpm typecheck`
- [x] `pnpm test` — 126 failures, all verified pre-existing against the parent commit
- [~] `pnpm build` — ran `build:electron-vite` and `typecheck`; did not run `build:native` / `build:relay` / `build:cli` / `build:web`, none of which this main-process-only change touches

New `profile-cloud-transport.test.ts`: 12 classification cases spanning both error shapes (undici errno chains and Chromium `net::ERR_*`), a guard that cloud requests never reach global fetch, and the live-socket repro test. Both existing client test files were migrated from `stubGlobal('fetch')` to mocking `net.fetch`.

## AI Review Report

Five review rounds over the diff. Each finding below was verified against the code, not eyeballed; all are fixed in follow-up commits on this branch.

**Correctness found and fixed**

- **TLS failures were misclassified as `unknown`.** Node reports certificate problems as bare `CERT_HAS_EXPIRED` / `CERT_REVOKED` / `CERT_UNTRUSTED` with no `err_` prefix, and the pattern required one. Probing the regex directly confirmed four such codes fell through. This is the *clock-skew* case, a documented adjacent report (#10401) — precisely the failure most worth naming. Matching on `cert` now; DNS is still evaluated first so a hostname containing "cert" cannot be mistaken for a certificate error.
- **`net::ERR_ABORTED` classified as `unknown`.** That is how Chromium surfaces our own 30s timeout signal. Also gave `ERR_ADDRESS_UNREACHABLE` and `ERR_SOCKET_NOT_CONNECTED` a home.
- **Detail could pair an outer error's `code` with an inner error's `message`.** Now built per chain level.

**Simplification**

- The `operation` label was threaded through five functions and eleven call sites, and `OrcaCloudTransportError.operation` was never read anywhere. Removed — the span derives the same information from the URL's origin and path. **Both cloud clients are back to a one-line change each** (`3 +-` per file, down from `26 +-` and `9 +-`). Also dropped a `codes` array the classifier no longer consulted.

**Verified rather than assumed**

Every test mocks `net.fetch`, so its real contract was unverified. Ran the actual init against real Electron:

```
[happy-path] RESOLVED status=200          <- accepts credentials/cache/redirect/signal
[redirect]   THREW  "Attempted to redirect, but redirect policy was 'error'"
[reset]      THREW  net::ERR_EMPTY_RESPONSE
[timeout]    THREW  TimeoutError
```

Two fixtures were wrong as a result and were corrected to what Electron actually throws: the redirect refusal is an `Error`, not a `TypeError`, and a dead socket reports `ERR_EMPTY_RESPONSE`, not `ERR_CONNECTION_RESET`.

**Cross-platform** — `net.fetch` is Electron-wide, not Windows-specific, and is already the norm here. Classification matches both undici errno chains and Chromium `net::ERR_*`, so no platform depends on a single error dialect. No path handling, shell invocation, or shortcut label is touched. Not SSH- or workspace-shaped: identical for git worktrees and folder workspaces. Verified every cloud entry point is reached through an IPC handler, hence after `app.whenReady()`, so `session.defaultSession` is always safe to touch.

**Auth state machine unchanged** — reconnect/refresh keys strictly on `OrcaCloudRequestError` and its status code (`isOrcaCloudAuthFailure`, `runOrgMemberCall`, the post-refresh 401 branch). A transport failure was a `TypeError` before and is an `OrcaCloudTransportError` now; both are non-`OrcaCloudRequestError`, so every branch takes the identical path.

**Deliberately not done** — no retry on transport failures. Retrying would burn the one-time authorization code and replace a truthful "connection reset" with a misleading `invalid_code`. The undici connection-pool behavior that motivated that suggestion is exactly what moving to `net.fetch` removes.

## Security Audit

**One real issue found during review, fixed and now tested.**

Electron's `net.fetch` attaches default-session cookies by default. Undici never did for these cross-origin requests. Switching stacks would therefore have *started* sending the app's session cookies to first-party token endpoints — a regression introduced by the fix itself. Confirmed against real Electron:

```
DEFAULT (no credentials option):   cookie header seen: "orca_session=SENSITIVE"
WITH credentials:'omit':           cookie header seen: null
```

`credentials: 'omit'` is therefore load-bearing, not cosmetic, and a test now asserts it so it cannot be dropped silently.

**Hardening that followed from it.** Both invariants that keep a code verifier or refresh token from reaching another origin were call-site conventions — `init` spread last, so any caller could weaken them with `redirect: 'follow'` or `credentials: 'include'`. They are now applied *after* the spread, so the transport guarantees them instead of trusting eleven call sites to agree. No behavior change today (every call site already passed `redirect: 'error'`), but the guarantee is structural, with a test that a caller attempting to opt out is overridden. Callers keep control of cache, method, headers, body, and signal.

**Reviewed and clear:**

- **Secret handling** — the failure detail derives solely from transport-layer error text (errno, `net::` code, hostname), never from the request body or headers, and is length-capped since it reaches a toast and the trace file. A test asserts the authorization code, verifier, nonce, and state never appear in the message. The span records origin + path with the query string dropped — the only part of a URL that could carry a secret; these endpoints put codes and tokens in the body or Bearer header.
- **Certificate validation** — nothing disables verification: no `rejectUnauthorized: false`, no `setCertificateVerifyProc`. Trust moves from Node's bundled CA list to the OS store, which is the platform trust model every other first-party integration here already uses.
- **SSRF** — request URLs come from `OrcaCloudAuthConfig`, built from `ORCA_CLOUD_*` environment variables. No user-controlled input reaches the host or protocol.
- **No new IPC surface, no command execution, no path handling, no deserialization, no new dependencies.** The error message crosses the existing IPC result field as a plain string.
- **Follow-up:** the relay token exchange in `runtime/relay/relay-http-client.ts` still uses global fetch and deserves the same treatment, but it runs post-sign-in and belongs to #10273.

## Notes

- Two entries were removed from `global-fetch-call-site-audit.test.ts`, since both cloud clients no longer call global fetch. `cancelUnreadResponseBody` is retained on the `!ok` path.
- Out of scope: `relay-http-client.ts` (#10273), and the server-side change in the issue's suggestion 4.

### DIFF
--- a/src/main/global-fetch-call-site-audit.test.ts
+++ b/src/main/global-fetch-call-site-audit.test.ts
@@ -18,8 +18,6 @@ const AUDITED_GLOBAL_FETCH_LINES = new Map<string, number>([
   ['main/azure-devops/azure-devops-api-request.ts', 1],
   ['main/bitbucket/client.ts', 1],
   ['main/gitea/client.ts', 1],
-  ['main/orca-profiles/profile-cloud-client.ts', 1],
-  ['main/orca-profiles/profile-cloud-org-members-client.ts', 1],
   ['main/rate-limits/codex-fetcher.ts', 3],
   ['main/runtime/relay/relay-http-client.ts', 2],
   ['main/source-control/hosted-review-api-request.ts', 1],
--- a/src/main/orca-profiles/profile-cloud-client.test.ts
+++ b/src/main/orca-profiles/profile-cloud-client.test.ts
@@ -10,7 +10,16 @@ import {
   selectOrcaCloudOrg
 } from './profile-cloud-client'
 
-const fetchMock = vi.fn()
+const { fetchMock } = vi.hoisted(() => ({ fetchMock: vi.fn() }))
+
+vi.mock('electron', () => ({
+  net: { fetch: fetchMock },
+  session: { defaultSession: {} }
+}))
+
+vi.mock('../network/proxy-settings', () => ({
+  ensureElectronProxyFromEnvironment: vi.fn().mockResolvedValue({ source: 'none' })
+}))
 
 const config: OrcaCloudAuthConfig = {
   apiBaseUrl: 'https://orca-cloud.example',
@@ -44,7 +53,6 @@ function mockFetchJson(value: unknown): void {
 describe('Orca cloud client', () => {
   beforeEach(() => {
     fetchMock.mockReset()
-    vi.stubGlobal('fetch', fetchMock)
   })
 
   it('normalizes session exchange organizations', async () => {
--- a/src/main/orca-profiles/profile-cloud-client.ts
+++ b/src/main/orca-profiles/profile-cloud-client.ts
@@ -7,6 +7,7 @@ import type { OrcaCloudAuthConfig } from './profile-cloud-auth-config'
 import type { OrcaCloudSession } from './profile-cloud-session-store'
 import type { OrcaCloudSessionExchangeResponse } from './profile-cloud-session-exchange'
 import { cancelUnreadResponseBody } from '../lib/unread-response-body'
+import { orcaCloudFetch } from './profile-cloud-transport'
 
 type ExchangeCodeArgs = {
   code: string
@@ -159,8 +160,13 @@ function normalizeSessionResponse(value: unknown): OrcaCloudSessionExchangeRespo
 
 const CLOUD_REQUEST_TIMEOUT_MS = 30_000
 
-async function postJson<T>(url: string, body: unknown, accessToken?: string): Promise<T> {
-  const response = await fetch(url, {
+async function postJson<T>(
+  operation: string,
+  url: string,
+  body: unknown,
+  accessToken?: string
+): Promise<T> {
+  const response = await orcaCloudFetch(operation, url, {
     method: 'POST',
     headers: {
       'content-type': 'application/json',
@@ -185,7 +191,7 @@ export async function exchangeOrcaCloudAuthCode(
   args: ExchangeCodeArgs
 ): Promise<OrcaCloudSessionExchangeResponse> {
   return normalizeSessionResponse(
-    await postJson(config.sessionEndpoint, {
+    await postJson('session-exchange', config.sessionEndpoint, {
       code: args.code,
       codeVerifier: args.codeVerifier,
       nonce: args.nonce,
@@ -204,7 +210,7 @@ export async function refreshOrcaCloudCapabilities(
     cloud?: unknown
     organizations?: unknown
     capabilities: unknown
-  }>(config.capabilitiesEndpoint, {}, session.accessToken)
+  }>('capabilities-refresh', config.capabilitiesEndpoint, {}, session.accessToken)
   return {
     cloud: response.cloud === undefined ? undefined : normalizeCloudSummary(response.cloud),
     organizations: normalizeOrganizations(response.organizations),
@@ -217,7 +223,7 @@ export async function refreshOrcaCloudSession(
   session: OrcaCloudSession
 ): Promise<OrcaCloudSessionExchangeResponse> {
   return normalizeSessionResponse(
-    await postJson(config.refreshEndpoint, {
+    await postJson('session-refresh', config.refreshEndpoint, {
       refreshToken: session.refreshToken
     })
   )
@@ -230,6 +236,7 @@ export async function createOrcaCloudProfile(
 ): Promise<OrcaCloudSessionExchangeResponse> {
   return normalizeSessionResponse(
     await postJson(
+      'profile-create',
       config.profileEndpoint,
       {
         orgId: args.orgId,
@@ -249,7 +256,7 @@ export async function selectOrcaCloudOrg(
     cloud: unknown
     organizations?: unknown
     capabilities: unknown
-  }>(config.orgEndpoint, { orgId }, session.accessToken)
+  }>('org-select', config.orgEndpoint, { orgId }, session.accessToken)
   return {
     cloud: normalizeCloudSummary(response.cloud),
     organizations: normalizeOrganizations(response.organizations),
@@ -261,5 +268,10 @@ export async function revokeOrcaCloudSession(
   config: OrcaCloudAuthConfig,
   session: OrcaCloudSession
 ): Promise<void> {
-  await postJson(config.logoutEndpoint, { refreshToken: session.refreshToken }, session.accessToken)
+  await postJson(
+    'session-revoke',
+    config.logoutEndpoint,
+    { refreshToken: session.refreshToken },
+    session.accessToken
+  )
 }
--- a/src/main/orca-profiles/profile-cloud-client.ts
+++ b/src/main/orca-profiles/profile-cloud-client.ts
@@ -160,13 +160,8 @@ function normalizeSessionResponse(value: unknown): OrcaCloudSessionExchangeRespo
 
 const CLOUD_REQUEST_TIMEOUT_MS = 30_000
 
-async function postJson<T>(
-  operation: string,
-  url: string,
-  body: unknown,
-  accessToken?: string
-): Promise<T> {
-  const response = await orcaCloudFetch(operation, url, {
+async function postJson<T>(url: string, body: unknown, accessToken?: string): Promise<T> {
+  const response = await orcaCloudFetch(url, {
     method: 'POST',
     headers: {
       'content-type': 'application/json',
@@ -191,7 +186,7 @@ export async function exchangeOrcaCloudAuthCode(
   args: ExchangeCodeArgs
 ): Promise<OrcaCloudSessionExchangeResponse> {
   return normalizeSessionResponse(
-    await postJson('session-exchange', config.sessionEndpoint, {
+    await postJson(config.sessionEndpoint, {
       code: args.code,
       codeVerifier: args.codeVerifier,
       nonce: args.nonce,
@@ -210,7 +205,7 @@ export async function refreshOrcaCloudCapabilities(
     cloud?: unknown
     organizations?: unknown
     capabilities: unknown
-  }>('capabilities-refresh', config.capabilitiesEndpoint, {}, session.accessToken)
+  }>(config.capabilitiesEndpoint, {}, session.accessToken)
   return {
     cloud: response.cloud === undefined ? undefined : normalizeCloudSummary(response.cloud),
     organizations: normalizeOrganizations(response.organizations),
@@ -223,7 +218,7 @@ export async function refreshOrcaCloudSession(
   session: OrcaCloudSession
 ): Promise<OrcaCloudSessionExchangeResponse> {
   return normalizeSessionResponse(
-    await postJson('session-refresh', config.refreshEndpoint, {
+    await postJson(config.refreshEndpoint, {
       refreshToken: session.refreshToken
     })
   )
@@ -236,7 +231,6 @@ export async function createOrcaCloudProfile(
 ): Promise<OrcaCloudSessionExchangeResponse> {
   return normalizeSessionResponse(
     await postJson(
-      'profile-create',
       config.profileEndpoint,
       {
         orgId: args.orgId,
@@ -256,7 +250,7 @@ export async function selectOrcaCloudOrg(
     cloud: unknown
     organizations?: unknown
     capabilities: unknown
-  }>('org-select', config.orgEndpoint, { orgId }, session.accessToken)
+  }>(config.orgEndpoint, { orgId }, session.accessToken)
   return {
     cloud: normalizeCloudSummary(response.cloud),
     organizations: normalizeOrganizations(response.organizations),
@@ -268,10 +262,5 @@ export async function revokeOrcaCloudSession(
   config: OrcaCloudAuthConfig,
   session: OrcaCloudSession
 ): Promise<void> {
-  await postJson(
-    'session-revoke',
-    config.logoutEndpoint,
-    { refreshToken: session.refreshToken },
-    session.accessToken
-  )
+  await postJson(config.logoutEndpoint, { refreshToken: session.refreshToken }, session.accessToken)
 }
--- a/src/main/orca-profiles/profile-cloud-org-members-client.test.ts
+++ b/src/main/orca-profiles/profile-cloud-org-members-client.test.ts
@@ -10,7 +10,16 @@ import {
   revokeOrcaCloudOrgInvite
 } from './profile-cloud-org-members-client'
 
-const fetchMock = vi.fn()
+const { fetchMock } = vi.hoisted(() => ({ fetchMock: vi.fn() }))
+
+vi.mock('electron', () => ({
+  net: { fetch: fetchMock },
+  session: { defaultSession: {} }
+}))
+
+vi.mock('../network/proxy-settings', () => ({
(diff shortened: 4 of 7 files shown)


1Morganmore:
### Operating system

Windows

### Orca version

1.4.183

### Details

## Summary

On Windows 11 x64, Orca repeatedly fails to connect a local profile to Orca Cloud.

The external browser login succeeds and reaches Orca's local callback page:

> Signed in to Orca

However, Orca then shows:

> Failed to connect profile
>
> fetch failed

The local profile remains unlinked and Orca Relay is unavailable.

## Environment

- OS: Windows 11 x64
- Orca version: 1.4.183
- Installation method: Official Orca Updater
- Terminal daemon version after reboot: 1.4.183
- Network during reproduction: smartphone hotspot only
- Wired Ethernet adapter: disabled
- Tailscale: fully stopped
- WinHTTP proxy: direct connection
- HTTP(S)_PROXY environment variables: not configured

The same general Relay flow works on another Windows desktop. The exact Orca version and environment of
that desktop may differ.

## Steps to reproduce

1. Start Orca with a local profile.
2. Open `Settings → Orca Account`.
3. Click `Sign in to Orca`.
4. Complete the browser-based sign-in.
5. Allow the browser to redirect to Orca's local callback:
   `http://127.0.0.1:<dynamic-port>/auth/callback`
6. The browser displays `Signed in to Orca`.
7. Return to Orca.

## Expected behavior

The OAuth code is exchanged for an Orca Cloud session, the local profile is linked, and Orca Relay
becomes available.

## Actual behavior

Orca displays:

```text
Failed to connect profile
fetch failed
 ```

 The profile remains local-only. No Cloud session is saved, and Orca Relay remains unavailable.

 This reproduces repeatedly.

 Additional diagnostics

 The following checks were performed from the same smartphone-hotspot network:

 - https://login.onorca.dev/health returns HTTP 200.
 - https://relay.onorca.dev/health returns HTTP 200.
 - A malformed POST to /v1/desktop/auth/session returns a structured HTTP 400 response.
 - A syntactically valid request with an intentionally invalid authorization code returns HTTP 401
   invalid_code.

 This indicates that DNS, TCP/TLS connectivity, and basic endpoint reachability are working. The failure
 appears to occur during the actual post-callback session exchange.

 The browser callback succeeds, but:

 - The local profile remains local.
 - No encrypted Cloud session artifact is created.
 - No profile linkage is persisted.
 - main.trace.ndjson does not contain a useful auth, fetch, TLS, DNS, or session-exchange error for the
   failed attempt.

 Technical suspicion

 The v1.4.183 Cloud client appears to use Node's global fetch for:

 ```text
POST https://login.onorca.dev/v1/desktop/auth/session
 ```

 The request uses redirect: 'error' and a 30-second timeout. A transport error, socket reset, TLS/proxy
 failure, or refused redirect is surfaced to the UI only as:

 ```text
TypeError: fetch failed
 ```

 The underlying error.cause is not exposed or recorded, so the exact transport failure cannot be
 identified.

 This appears related to:

 - https://github.com/stablyai/orca/issues/10758
 - https://github.com/stablyai/orca/pull/10789

 Requested fix

 Please consider:

 1. Routing Cloud authentication requests through Electron's net.fetch or another transport that uses the
    OS proxy and certificate configuration.
 2. Preserving and reporting the nested transport error cause, including:
   - DNS error code
   - TCP/TLS error code
   - proxy source
   - socket error
   - redirect status
   - timeout
 3. Adding a regression test for:
   - successful browser callback followed by session exchange
   - valid-code connection close
   - refused redirect
   - TLS/DNS/socket transport failures

Minidoracat:
Your transport half is now superseded: #15208 routes the same two Cloud clients through Electron's network stack *and* covers the Relay control/data sockets, artifact requests and the region catalog. After it went `dirty`, I rebased it onto current `main` as #16399.

Your **cause-reporting half is not superseded**, and it is the part that keeps costing people time. `profile-cloud-service.ts` on `main` still collapses failures to `error.message` in three places. So once #16399 lands, the next transport failure of any kind still surfaces as a bare `fetch failed` with `cause` discarded — for example the usage requests in `main/rate-limits/codex-fetcher.ts`, which `global-fetch-call-site-audit.test.ts` still lists as bare-`fetch` sites and which fail identically on an intercepted host (measured: `UNABLE_TO_VERIFY_LEAF_SIGNATURE`, reported to the user as nothing at all).

This branch needs a rebase regardless — its base is ~1840 commits behind `main` and its last commit is 2026-07-27. If it is rebased down to just the transport-error classification and the `cause` propagation, it should land without touching anything #16399 changes.

Worth recording here too, since this PR is where the `redirect: 'error'` / lost-`cause` analysis started: the reason the cause is worth surfacing is that recovering it by hand is what identified this bug class twice, and both times the string was already in the process. On the host I measured, the discarded `cause` was `UNABLE_TO_VERIFY_LEAF_SIGNATURE` — one line that would have replaced weeks of triage across #10758, #13735 and #14865.

PR #10790, opened 2026-07-26
perf(github): fan out PR file pages instead of probing serially

nwparker:
## Summary

`getPRFiles` discovered its page count by *probing*: request page 1, and if it came back full, request page 2, and so on until one came back short.

Each page is a separate `gh api` subprocess plus a network round trip. The probe therefore did two wasteful things:

1. **Serialized** the pages — page 2 could not start until page 1 fully returned.
2. **Spent an extra request** whenever the file count landed on an exact page boundary. A PR with exactly 100 changed files cost **two** requests to fetch **one** page of data.

The PR item already carries `changedFiles`, so the page count is knowable up front. Compute it and issue the pages together. Items without a resolved count keep the serial probe.

## ELI5

To list the files in a pull request, Orca asks GitHub for them 100 at a time. It didn't know how many there were, so it kept asking for "the next 100" until GitHub came back with a short page meaning *that was the last one*.

Two problems. It had to wait for each batch before asking for the next, even though it could have asked for all of them at once. And if a PR had exactly 100 files, it asked twice — the second time only to hear "there's nothing more."

Orca already knew the file count from the PR itself. Now it does the arithmetic, asks for exactly the pages it needs, and asks for them simultaneously.

## Screenshots

No visual change.

Measured locally with recorded/synthetic payloads — **no live API calls**, in keeping with the repo's rate-limit guidance.

**What a round trip actually costs:**

```
gh process startup   43 ms   (median of 20 `gh --version` spawns;
                              vs 1.5 ms for /bin/echo, so it is gh's Go
                              runtime + config load, not fork cost)
TCP connect          36 ms   (median of 10 to api.github.com)
                     -----
                    ~139 ms  per `gh api` round trip, budgeting TLS + server time
```

**Serial vs parallel, real process spawns piping a 360 KiB page payload** (medians of 12 interleaved samples):

| pages | serial | parallel | saved |
|---|---|---|---|
| 2 | 96.2 ms | 48.9 ms | 47.4 ms (49%) |
| 3 | 151.8 ms | 53.2 ms | 98.6 ms (65%) |

**End-to-end critical path** across the four parallel detail branches, at 139 ms/round trip:

| PR size | before | after |
|---|---|---|
| 100 files | 278 ms | **139 ms** (-50%) |
| 200 files | 417 ms | **278 ms** (-33%, now bounded by viewedStates) |

**The wasted request, verified independently against the verbatim old loop:**

| files | old spawns | needed | wasted |
|---|---|---|---|
| 100 | 2 | 1 | **1** |
| 150 | 2 | 2 | 0 |
| 200 | 3 | 2 | **1** |
| 300 | 3 | 3 | 0 |

## Proof of zero regression

The returned file list is unchanged — same pages, same `MAX_PR_FILES` cap, same mapping.

Four new tests pin the **spawn count**, which is the actual behavioural contract:

- 100 files → exactly **1** request (was 2)
- 200 files → exactly **2** requests, including `page=2` (was 3)
- 900 files → **3** requests, still capped at 300 files
- `changedFiles` absent → serial probe fallback, 2 requests for 150 files

**These are not vacuous:** reverting to serial probing makes **2 of the 4 fail**. I verified that by disabling the known-count path and re-running.

Two correctness details worth review:

- **Rate-limit blocking mid-fan-out.** The old loop returned `null` the moment `repositoryRateLimitGuard` blocked. With a fan-out, some pages may already be in flight when another is blocked. If *any* page returns `null`, the whole read returns `null` — a silently truncated file list would be worse than no list.
- **`noteRepositoryRateLimitSpend` accounting is unchanged** — still one spend per issued request. The fan-out issues *fewer or equal* requests than the probe, never more.

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — **592/592** across `src/main/github/`; full suite **2 failures / 36,938**, both pre-existing

`pnpm test` is already red on `main`. This run shows **only** `agent-exec-handler.test.ts` ×2, which assert against the machine's real `process.env` — a strict subset of the known baseline pool, and neither imports the changed code.

**I rejected this candidate first, then retracted the rejection.** Worth recording why.

My initial reasoning: capped at 300 files (3 pages max), already inside a `Promise.all` behind a concurrency permit, responses served by `gh api --cache 60s` — "≤2 saved round trips, not worth the complexity."

That was right in **count** and wrong in **magnitude**. I had estimated the round-trip cost; the measurement above puts it at ~139 ms, so "2 round trips" is ~278 ms on the path between clicking a PR and seeing its files. Rejection withdrawn.

The same investigation **rejected a neighbouring candidate**, and its reasoning is what identified this one: `getPRComments` fetches REST issue-comments, maps them, then discards the result because GraphQL already returned the same data. Genuinely dead work — but those calls run **in parallel** through a 4-wide semaphore with headroom, so deleting one spawn saved **0.8 ms of 186 ms (0.4%)**. Deleting *two* saved 22.9%. That contrast is what pointed here, where the calls are **serial**.

The transferable lesson: in this codebase, a redundant *parallel* spawn is nearly free; a *serialized* one is not.

## AI Review Report

Risks reviewed:

- **Truncated results under rate limiting** — addressed above; any blocked page fails the whole read.
- **Over-fetching from a stale count.** If `changedFiles` overstates reality, extra pages return `[]` and contribute nothing; the cap still applies. If it understates, we fetch fewer pages than the file list contains — which is why the fallback path is preserved for items whose count is absent rather than trusted blindly for items where it is wrong. Worth a reviewer's eye: I am treating `changedFiles` from the PR object as authoritative when present.
- **Semaphore pressure.** `gh-utils` caps concurrency at 4 and the caller already holds a permit per branch. The fan-out issues at most 3 pages and never more requests than the probe did, so peak in-flight work does not increase beyond what the existing limiter already governs.
- **Provider genericity.** `changedFiles` is populated on the GitHub work-item path; providers that leave it undefined fall through to the unchanged serial loop, so GitLab and others are unaffected.

Cross-platform: no path, shell, or platform-dependent behaviour touched. Same `gh` invocations with the same arguments on macOS/Linux/Windows; unaffected by SSH/WSL/remote hosts.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. The same `gh api` endpoint is called with the same arguments and the same `--cache 60s`; only the number and scheduling of calls changed. The `MAX_PR_FILES` cap that prevents a massive PR from starving the `gh` semaphore is preserved exactly, and the page count is clamped to it before any request is issued, so a hostile `changedFiles` value cannot inflate the request count. Rate-limit accounting is unchanged. No dependency changes.

### DIFF
--- a/src/main/github/work-item-details-file-pagination.test.ts
+++ b/src/main/github/work-item-details-file-pagination.test.ts
@@ -1,0 +1,192 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const {
+  acquireMock,
+  releaseMock,
+  ghExecFileAsyncMock,
+  getWorkItemMock,
+  getPRCommentsMock,
+  getPRChecksMock
+} = vi.hoisted(() => ({
+  acquireMock: vi.fn<() => Promise<void>>(),
+  releaseMock: vi.fn(),
+  ghExecFileAsyncMock: vi.fn(),
+  getWorkItemMock: vi.fn(),
+  getPRCommentsMock: vi.fn(),
+  getPRChecksMock: vi.fn()
+}))
+
+vi.mock('./gh-utils', () => ({
+  acquire: acquireMock,
+  release: releaseMock,
+  ghExecFileAsync: ghExecFileAsyncMock,
+  ghRepoExecOptions: (context: { repoPath: string }) => ({ cwd: context.repoPath }),
+  githubRepoContext: (repoPath: string, connectionId?: string | null) => ({
+    repoPath,
+    connectionId: connectionId ?? null
+  })
+}))
+
+vi.mock('./client', () => ({
+  getWorkItem: getWorkItemMock,
+  getPRComments: getPRCommentsMock,
+  getPRChecks: getPRChecksMock
+}))
+
+vi.mock('./github-api-repository', () => ({
+  getIssueGitHubApiRepository: vi.fn(),
+  getOriginGitHubApiRepository: vi.fn(),
+  githubHostExecOptions: (repository?: { host?: string } | null) =>
+    repository?.host ? { host: repository.host } : {},
+  // Why: getWorkItemDetails awaits resolveGitHubRepoExecution and reads .ownerRepo.
+  resolveGitHubRepoExecution: vi.fn(
+    async (
+      _repoPath: string,
+      repository?: { owner: string; repo: string; host?: string } | null
+    ) => ({
+      ownerRepo: repository ?? { owner: 'acme', repo: 'widgets', host: 'github.com' },
+      ghOptions: {
+        cwd: '/repo',
+        host: repository?.host ?? 'github.com'
+      }
+    })
+  )
+}))
+
+vi.mock('./rate-limit', () => ({
+  repositoryRateLimitGuard: vi.fn(() => ({ blocked: false })),
+  noteRepositoryRateLimitSpend: vi.fn()
+}))
+
+import { getWorkItemDetails } from './work-item-details'
+
+describe('getWorkItemDetails PR file pagination', () => {
+  let filePageRequests: string[]
+
+  beforeEach(() => {
+    filePageRequests = []
+    acquireMock.mockReset()
+    acquireMock.mockResolvedValue(undefined)
+    releaseMock.mockReset()
+
+    getPRCommentsMock.mockReset()
+    getPRCommentsMock.mockResolvedValue([])
+    getPRChecksMock.mockReset()
+    getPRChecksMock.mockResolvedValue([])
+
+    ghExecFileAsyncMock.mockReset()
+    ghExecFileAsyncMock.mockImplementation(async (args: string[]) => {
+      const query = args.find((arg) => arg.startsWith('query=')) ?? ''
+      if (query.includes('viewerViewedState')) {
+        return {
+          stdout: JSON.stringify({
+            data: {
+              repository: {
+                pullRequest: {
+                  id: 'PR_id',
+                  files: { pageInfo: { hasNextPage: false }, nodes: [] }
+                }
+              }
+            }
+          })
+        }
+      }
+      if (query.includes('participants')) {
+        return {
+          stdout: JSON.stringify({
+            data: { repository: { pullRequest: { participants: { nodes: [] } } } }
+          })
+        }
+      }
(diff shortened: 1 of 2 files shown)


nwparker:
## Closing this — the win does not survive verification

An independent verifier refuted this, and I confirmed its two load-bearing points myself. Closing rather than leaving a PR open whose stated benefit I no longer believe.

### 1. The files branch is never the sole gate

`getPRFileViewedStates` is a sibling in the **same** `Promise.all` and pages the **same** file list at 100/page — but over GraphQL cursors (`after = endCursor`), which are **structurally serial**: each page needs the previous page's `endCursor`.

So parallelizing the REST file pages just makes that branch finish earlier and wait on the cursor chain. I verified the code: `work-item-details.ts:635` holds `let after: string | null = null`, threaded through the loop. Files pages *can* be parallelized; viewed-state pages *cannot*.

Measured round-trip depth before → after:

```
 60f 2.2 -> 2.2 (0)      150f 3.3 -> 3.3 (0)
 99f 2.2 -> 2.2 (0)      199f 3.3 -> 3.3 (0)
100f 3.3 -> 2.2 (-1 RT)  200f 4.4 -> 3.3 (-1 RT)
101f 3.3 -> 3.3 (0)      201f 4.4 -> 4.4 (0)
```

The saving exists **only at exact multiples of 100**.

### 2. That case essentially never happens

I measured the real distribution over 1,197 commits on `main`:

| | files |
|---|---|
| median | **5** |
| p90 | 24 |
| p95 | 37 |
| p99 | 104 |
| max | 1626 |

- **≥100 files: 12 of 1197 (1.00%)**
- **≥200 files: 4 (0.33%)**
- **exactly 100: 0. exactly 200: 0.**

Net expected saving for a normal user: **~0 ms**.

### What I got wrong

My PR body led with "417 → 278 ms for a 200-file PR." That number is real in isolation but describes a PR shape that occurs in **0.33%** of cases, and even then only saves a round trip when the count is an exact multiple of 100 — which occurred **zero times** in my sample.

I'd also already retracted an earlier *rejection* of this candidate after seeing the ~139 ms round-trip cost. That retraction was correct about round-trip cost and wrong about whether this path ever pays it. The lesson I'm taking: measuring the cost of an operation is not the same as measuring how often it occurs, and I checked the first without checking the second.

The `~139 ms per gh api round trip` figure stands and is worth keeping — it's what makes **serial** call chains worth attacking elsewhere.

No behaviour change ships; closing.

PR #10797, opened 2026-07-26
fix(terminal): quarantine input after a dead-endpoint recovery

brennanb2025:
Follow-up from live QA of #10065 (STA-2373). This is the one finding from that QA with a safety edge.

## What happens today

STA-2373 recovery replaces a dead daemon's shell with a fresh one, but the user is usually **mid-line** when the daemon dies. Keystrokes racing the death are dropped; everything typed after re-attach lands on the **new** shell. So the shell receives the *tail* of a line whose head no longer exists.

Measured live, not inferred — killed the daemon, then sent digits `1`–`9` one per ~1.1s:

```
$ 23456789
zsh: command not found: 23456789
```

Only `1` was consumed as the recovery trigger; `2`–`9` reached the replacement shell. There is no buffering or replay involved.

## Why it needs fixing

Truncating a simple command is harmless. A **compound** one is not — losing the head of `echo hi; rm -rf x` gives `cho hi; rm -rf x`, where zsh fails `cho` and then **still executes `rm -rf x`** after the `;`.

Pre-#10065 the whole line was silently dropped and nothing ran, so partial execution is genuinely new. Executing a fragment of a command the user never completed is worse than dropping it.

## Fix

Quarantine input from the dead-endpoint signal until the next line boundary, swallowing that boundary so the fresh shell starts at a clean prompt.

- State is **module-scoped by tabId**, not closure state: recovery remounts the pane and builds a brand-new connection, discarding closure state exactly when the quarantine needs to apply.
- Placed after the query-reply bypass so xterm's own auto-replies still flow, and before every send path.
- Re-arming is idempotent (both the provider fan-out and the per-write catch signal the written pane) so one incident cannot extend the window.
- A 15s deadline guarantees input is never permanently dead if the user switches away mid-line and never submits.

## Validation

- `pnpm typecheck` clean; oxlint + oxfmt clean; `check:max-lines-ratchet` OK, no new bypasses
- 572 passed across `pty-connection`, `pty-transport`, and the new `terminal-input-quarantine` suites
- 9 unit tests + 1 integration test through the real `onData` path
- **Revert-sensitive, verified by mutation**: disabling the quarantine check fails the integration test with `"cho hi; rm -rf x"` delivered to the transport

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-connection.test.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.test.ts
@@ -4819,6 +4819,13 @@ describe('connectPanePty', () => {
     expect(transport.sendInput).not.toHaveBeenCalled()
     expect(sendInputAccepted).not.toHaveBeenCalled()
 
+    // The user is told why their typing vanished, once, in the pane itself.
+    const noticeWrites = pane.terminal.write.mock.calls
+      .map((c: unknown[]) => String(c[0]))
+      .filter((t: string) => t.includes('Terminal reconnected'))
+    expect(noticeWrites).toHaveLength(1)
+    expect(noticeWrites[0]).toContain('\u001b[7m')
+
     // ...and the next line the user types is delivered normally.
     sendTerminalInputThroughPane(pane, 'ls')
     await flushAsyncTicks(2)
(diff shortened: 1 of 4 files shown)


brennanb2025:
Superseded by #10811, which landed the same fix with a better design: it arms only when the endpoint was actually replaced (this PR armed on every write-unavailable signal, so a transient reconnect to a still-live shell would have eaten a legitimate command), arms only once the remount is certain rather than in the callback, and releases on an idle gap and on Ctrl-C instead of holding until Enter.

Closing rather than reworking. The one piece here that #10811 lacks is the in-pane notice explaining why input was discarded; #10811's idle-gap release makes the silent window short enough that it is much less necessary, so it is noted as an optional follow-up rather than carried forward.

PR #10800, opened 2026-07-26
fix(browser): restore configured zoom after page reload

luizeof:
## Summary

- Reapply the configured browser page zoom after every guest `dom-ready` event.
- Keep zoom reset aligned with the configured default level.
- Add unit and Electron E2E regression coverage for repeated normal and hard reloads.

## Root cause

Chromium can restore per-origin zoom when an embedded page reloads. Orca only enforced its configured zoom when the webview was first created, so a native zoom value above 100% could survive subsequent reloads.

## User impact

Embedded browser pages now return to the configured zoom after reload instead of retaining an unexpected zoom above 100%. Explicit non-default browser zoom settings remain respected.

## Validation

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/browser-pane` — 31 files, 162 tests passed
- Focused Electron E2E regression — 1 test passed across repeated normal and hard reloads
- `pnpm run typecheck`
- `pnpm exec oxlint` on changed files
- `pnpm exec oxfmt --check` on changed files

## ELI5

After reloading a page in Orca's embedded browser, Chromium could restore an old zoom level instead of your configured one. This re-applies your preferred zoom after every page load so zoom stays consistent.

### DIFF
--- a/src/renderer/src/components/browser-pane/BrowserPane.tsx
+++ b/src/renderer/src/components/browser-pane/BrowserPane.tsx
@@ -3560,7 +3560,11 @@ function BrowserPagePane({
       if (!isActiveRef.current) {
         return
       }
-      const nextLevel = applyBrowserPageZoom(webviewRef.current, direction)
+      const nextLevel = applyBrowserPageZoom(
+        webviewRef.current,
+        direction,
+        browserDefaultZoomLevelRef.current
+      )
       if (nextLevel !== null) {
         setBrowserDefaultZoomLevel(nextLevel)
         showBrowserZoomFeedback(nextLevel)
@@ -3659,7 +3663,6 @@ function BrowserPagePane({
     container = ensuredWebview.container
     const webview = ensuredWebview.webview
     const needsInitialNavigation = ensuredWebview.created
-    let needsInitialDefaultZoom = ensuredWebview.created
 
     if (!ensuredWebview.created) {
       // pointerEvents already applied inside ensureBrowserPageWebview for the reused-webview path.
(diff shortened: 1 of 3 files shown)


OrcaWin:
Reviewed with an independent reproduction. **The bug is real and I confirmed it** — on main a native zoom above 100% survives both normal and hard reload, and your `dom-ready` reapply does fix that.

But the fix as written was wired to the **shared** `browserDefaultZoomLevel`, which zoom in/out also writes to, and that introduced two regressions I confirmed with before/after screenshots in the running app:

1. **Cmd/Ctrl+0 reset became a no-op fixed point.** Reset targets the value zooming just wrote, so a page stuck at 120% could never return to 100%.
2. **Reloading a tab silently adopted a zoom applied to a different tab.**

Neither was caught by the new test — and this is the part worth internalizing: **the unit test passed with the fix fully reverted**, because it only exercises a module the PR doesn't change. A test that stays green when you delete the fix can't protect it.

**Pushed to this branch:** a per-pane zoom level seeded from the configured default and reasserted on `dom-ready`, so the reload behavior you wanted works without the shared-value coupling. Plus real E2E coverage in `tests/e2e/browser-tab.spec.ts` that drives the actual platform Cmd/Ctrl+0 chord into the guest webview and uses distinct origins to prove tab isolation.

I verified the new tests aren't vacuous the same way I found the original problem — reverted the rewrite while keeping the tests, and 2 of 3 failed exactly on Cmd/Ctrl+0 (got 0.5) and cross-tab isolation (got 1). Restored: 3/3 pass.

Validation: 167 browser-pane tests, 3 Playwright E2E, typecheck 0, oxlint clean.

Thanks for finding this — the per-origin zoom restore is genuinely non-obvious behavior and the diagnosis was right.

Issue #9138, opened 2026-07-17
[Bug]: App updates leave previous daemon generations running forever — invisible agent sessions accumulate and exhaust memory


oratta:
### Operating system

macOS

### Orca version

v1.4.143 (issue involves daemons left behind by v1.4.100, v1.4.135, v1.4.139)

### Details

**Short summary:**  <br>Each Orca app update spawns a new versioned terminal daemon (`daemon-vNN`), but the previous generation's daemon is never shut down. Old daemons keep running indefinitely with all of their PTY sessions and coding-agent processes (Claude Code sessions + their MCP server children). These sessions are no longer visible in the UI, so they silently accumulate across updates and eat memory.

**What happened?**

On my machine (macOS, 32 GB RAM) I noticed heavy swap pressure (25 GB of 26 GB swap used). Investigating, I found **4 generations of Orca daemons running at the same time**:

| daemon | app version | uptime | direct children |
| -- | -- | -- | -- |
| daemon-v18 | 1.4.100 | 20 days | 22 |
| daemon-v20 | 1.4.135 | 5 days | 11 |
| daemon-v21 | 1.4.139 | 2.5 days | 8 |
| daemon-v22 | 1.4.143 (current) | 4 hours | 3 |

The UI showed only 8 open sessions, but `ps` showed ~46 live Claude Code processes. Killing the three stale daemon trees (v18/v20/v21) terminated **370 processes** in total (agent sessions + MCP servers + shells) and brought swap usage from 25 GB down to under 5 GB.

The pid files under `~/Library/Application Support/orca/daemon/` (`daemon-v18.pid` … `daemon-v22.pid`, plus matching `.sock`/`.token` files) confirm each generation is tracked separately, but nothing appears to reap the old generations after an update.

**How can we reproduce it?**

1. Open several workspaces/sessions with a coding agent in Orca.
2. Let Orca auto-update to a new version (repeat over a few weeks so multiple updates happen).
3. Check `~/Library/Application Support/orca/daemon/` and `ps` — old `daemon-vNN` processes from previous app versions are still running, along with every agent session they hosted, even though those sessions are not shown in the UI.

**Anything else that might help:**

* Expected behavior: on startup of a new daemon generation, either migrate/adopt sessions from the previous daemon, or gracefully shut the old daemon and its children down (or at least surface those sessions somewhere in the UI so users can close them).
[...]

PR #10806, opened 2026-07-26
fix(daemon): retire a daemon left empty by an unclean client disconnect

Jinwoo-H:
Partial fix for #9138. Implements section 6 of the reviewed design in [AmethystLiang's comment](https://github.com/stablyai/orca/issues/9138#issuecomment-list), which states that piece may ship independently because it never infers ownership and never acts while a session exists.

## The bug

Each app update spawns a new versioned daemon (`daemon-vNN`) and deliberately leaves the previous generation running so its PTYs survive the upgrade.

**Correction to an earlier version of this description:** it claimed retirement "was never built". That is wrong — #9277 added event-driven retirement for empty daemons, including an e2e spec proving a v22 legacy daemon stays reattachable while a v24 retires. This PR adds one uncovered case on top of that, described below.

The field report found four generations coexisting, 370 stray processes, and 25 GB of swap — each stray agent session holding 300–450 MB plus MCP children.

## Why the existing retirement path misses it

The daemon exits only when `isIdle()` is true, which requires zero sessions (`daemon-server.ts:303`). Two things that sound like the fix aren't:

- `pruneOldDaemonHosts` is **Windows-only and deletes directories, not processes** (`daemon-host-relocation.ts:321`). The reporters are on macOS.
- A **clean** disconnect already sets `retirementRequested`, so a daemon that drains later does exit today.

The gap is a client that never disconnects cleanly — an app killed by an update, a crash, or force-quit. `recordFullyAuthenticatedDisconnect` never runs, so `retirementRequested` is never set, and nothing re-checks afterwards. The daemon then outlives every generation after it.

**This PR does not fix the reported symptom.** The leaked generations in #9138 were *holding live agent sessions* — the reporter found "`claude` processes up to ~1.8 days old, several hundred MB RSS each", and the original report lists 22, 11, and 8 direct children per stale daemon. This timer never arms while a session is live, because that is what protects warm reattach. So on the machines in those reports it changes nothing.

What it does close is a real hole in the retirement state machine: a daemon that becomes empty after its client died uncleanly currently stays alive forever. That is worth fixing on its own, but it is not the memory leak.

Reaping generations that still hold unclaimed sessions needs the cross-profile ownership snapshot from sections 1-5 of the reviewed design, and per that design its first release is audit-only.

## The change

The daemon arms a 30-minute timer when it has **zero clients and zero sessions**, and exits on expiry.

- A daemon **holding a session never arms it** — a background agent is live work, and warm reattach must keep working.
- The timer defers to the existing 2-minute adoption deadline before the first client pair, so the two never race.
- Expiry re-checks idleness rather than trusting the timer. Cancel-on-create is the fast path, not the guarantee.

**30 minutes** is chosen so a brief restart or update reconnects to the same daemon and keeps warm reattach, while a generation left behind by an update exits the same day instead of persisting for weeks.

`PROTOCOL_VERSION` → 29, since idle policy is baked into the daemon binary. 28 joins `PREVIOUS_DAEMON_PROTOCOL_VERSIONS` so running v28 daemons stay adoptable across this upgrade.

## Scope — what this does not fix

**Daemons already running on a user's machine are unaffected.** v28-and-older binaries cannot acquire idle policy; it ships inside the daemon. Those need app-side legacy retirement, which depends on the cross-profile ownership snapshot in sections 1–5 of the design and is not in this PR.

So: this stops *future* generations from being abandoned. Reaping *existing* ones is follow-up work, and per the design its first release is audit-only.

## Verification

- `vitest src/main/daemon` — **1,122 passed**, 2 skipped
- 4 new tests: arms when abandoned, never arms with a live session, cancels on client reconnect, and does not fire when a session is created after arming
- E2E: `terminal-restart-persistence` + `restart-restore-terminal-input` — **8 passed** (warm reattach across real quit/relaunch, the behaviour most at risk here)
- `tsc --noEmit -p config/tsconfig.node.json`, oxlint, oxfmt — clean

**Mutation-tested.** Removing the expiry idleness re-check alone does *not* fail, because `beginIdleShutdown` independently re-tests. Removing both does. The re-check is deliberate defense in depth, and the comment says so rather than implying it is load-bearing alone.

## Unrelated flake worth flagging

`headless-serve-desktop-activation.spec.ts` fails ~2 in 3 runs **on `main`** at the same rate as on this branch (`promotedPtyId` differs — the daemon terminal is replaced). Not caused by this PR; I verified by running it repeatedly on both. Likely related to #10518, since PR CI never runs the E2E suite.

### DIFF
--- a/src/main/daemon/daemon-foreground-confirmation-protocol.test.ts
+++ b/src/main/daemon/daemon-foreground-confirmation-protocol.test.ts
@@ -3,12 +3,13 @@ import { PREVIOUS_DAEMON_PROTOCOL_VERSIONS, PROTOCOL_VERSION } from './types'
 
 describe('foreground-confirmation daemon protocol', () => {
   it('rejects daemons from before the fresh-confirmation RPC', () => {
-    expect(PROTOCOL_VERSION).toBe(28)
+    expect(PROTOCOL_VERSION).toBe(29)
     expect(PREVIOUS_DAEMON_PROTOCOL_VERSIONS).toContain(19)
     expect(PREVIOUS_DAEMON_PROTOCOL_VERSIONS).toContain(22)
     expect(PREVIOUS_DAEMON_PROTOCOL_VERSIONS).toContain(23)
     expect(PREVIOUS_DAEMON_PROTOCOL_VERSIONS).toContain(24)
     expect(PREVIOUS_DAEMON_PROTOCOL_VERSIONS).toContain(25)
     expect(PREVIOUS_DAEMON_PROTOCOL_VERSIONS).toContain(26)
+    expect(PREVIOUS_DAEMON_PROTOCOL_VERSIONS).toContain(28)
   })
 })
--- a/src/main/daemon/daemon-idle-shutdown.test.ts
+++ b/src/main/daemon/daemon-idle-shutdown.test.ts
@@ -155,6 +155,7 @@ describe('current daemon lifecycle retirement', () => {
     options: {
       launchNonce?: string
       protocolVersion?: number
+      idleSelfShutdownTimeoutMs?: number
     } = {}
   ): Promise<void> {
     server = new DaemonServer({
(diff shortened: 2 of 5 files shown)


Jinwoo-H:
Closing. This fixes a real hole in the retirement state machine, but not the case reported in #9138 — those daemons hold live agent sessions, and this timer never arms while a session is live. Not worth a protocol bump for a case that doesn't match the reports.

The actual leak needs the cross-profile ownership snapshot from the reviewed design, since killing a live session safely requires proving no profile claims it.

PR #7316, opened 2026-07-04
Tab strip: judge click-vs-drag at pointerup and flush gesture state on window focus

Jinwoo-H:
## Summary

Fixes a tab-switching input failure seen in packaged builds on one user's machine where clicking a tab instantly grabbed it as a drag instead of switching tabs.

The important symptom pattern was:

- After window blur + refocus, the first tab click switches correctly.
- Every later click is immediately treated as a drag.
- View > Force Reload does not clear it.
- `pnpm dev` on the same machine is fine.
- Teammates on the same packaged build cannot reproduce.
- The terminal-flood/main-thread-stall bug is fixed in `v1.4.122-rc.1.perf`, but this tab symptom still reproduces, so stalls are not the full cause.

## Mechanism

The previous tab activation hook latched drag suppression on the first `pointermove` that crossed `TAB_DRAG_ACTIVATION_DISTANCE_PX` (12px). That meant a single stale/coalesced/desynced first move after a stationary `pointerdown` permanently converted the gesture into a drag, even if the final `pointerup` was back at the click position.

This change makes the tab activation decision use final `pointerup` displacement from the recorded `pointerdown` point. A stationary release after one bad move is a click; a final release outside the threshold is a drag.

Because dnd-kit had the same first-move distance activation behavior independently, this also replaces the stock `PointerSensor` for tabs with a tab-specific pointer sensor. It still uses the same 12px threshold, but an immediate over-threshold sample must be confirmed by either:

- a second consecutive over-threshold sample, or
- a short 50ms grace period elapsing.

That keeps real drags responsive while preventing one bad first sample from starting dnd-kit drag state.

## Other Robustness

- Adds `focus` cleanup beside the existing missed-end `blur` fallback in `useTabDragSplit`.
- Flushes pending tab activation state on window `focus` and `blur`.
- Tracks active tab-strip pointer gestures and suppresses overflow `scrollIntoView` / sticky-end scroll adjustments while a tab press is unresolved, so active-tab preview or layout changes cannot move the strip under a stationary cursor mid-gesture.

## Tests

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/tab-bar/tab-strip-pointer-activation.test.tsx`
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/tab-group/useTabDragSplit.test.ts`
[...]

PR #10807, opened 2026-07-27
fix(tabs): keep the first tab click after a browser pane takes focus

nduongg04:
## Summary

Clicking a tab while a browser pane has keyboard focus did nothing — you had to click twice. This makes the first click work.

Repro on `main`: open a browser tab, click anywhere **inside the page**, then click another tab in the strip. Nothing happens. Click again and it switches.

Tab activation is deferred to `pointerup`, and #7316 also cancelled a pending press on window `blur`/`focus` so a press whose release never arrived could not leak the strip's scroll-suppression token. But an Electron `<webview>` guest handing focus back to the host fires a real window `focus` event **between** that press's pointerdown and pointerup. Captured with a listener registered exactly the way the hook registers its own:

```
pointerdown            target=SPAN (the tab)
HOOK-would-fire:focus  target=WINDOW      <- cancel ran here
pointerup              target=SPAN        <- listener already removed
```

`onPointerCancel` → `cleanup()` removed the `pointerup` listener, so `onActivate()` never fired. The events reach React fine — nothing calls `stopPropagation` — the handler simply has nothing left to run.

Fix: window `focus` now releases the gesture token (which is what #7316 wanted) without cancelling the press, so the release still judges click-vs-drag. `blur` still cancels.

## Screenshots

Before: with a browser page focused, the first click on a terminal tab does nothing; the second switches.
After: the first click switches.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Notes on the runs:

- `pnpm lint` reports pre-existing unlocalized strings in `src/renderer/src/components/settings/terminal-advanced-platform-search.ts` and `terminal-pane-appearance-search.ts` ("Ghostty"). Identical on a clean tree with this branch stashed; untouched by this PR.
- `pnpm test`: 36876 passed. 45 failures across 8 files, all pre-existing and unrelated — verified by running exactly those 8 files on clean `main` (`4ada3f8b2`), where they fail identically. They live in `src/relay/`, `right-sidebar/`, `sidebar/` and `browser-pane/markup/`; none in the files this PR touches. `tab-bar` 246/246 and `tab-group` 71/71 pass.
- The existing `flushes a pending press on window focus` test encoded the cancelling behaviour, so it is rewritten to assert the token is released **and** the click still lands. Two tests added, one of which replays the real captured event ordering.
- Mutation-checked: restoring `focus → onPointerCancel` fails exactly the two new tests and nothing else.

## AI Review Report

- **Cross-platform.** No platform branching added or changed. The listeners are `window` `focus`/`blur`, which fire on all three platforms; nothing here reads `metaKey`/`ctrlKey`, key labels, or paths. The behaviour difference this fixes is Chromium's focus transfer out of a `<webview>`, which is identical on macOS, Linux and Windows. Verified on macOS; no macOS-only API is involved.
- **Regression risk on the original fix.** The concern behind #7316 was a leaked scroll-suppression token, not activation. The token is still released on `focus` (and on `blur`, `pointerup`, `pointercancel`, and unmount), so the leak stays closed. This was the main thing flagged, and it is why the fix releases the token rather than removing the listener.
- **Stray activation.** If a press is released while the window is unfocused, the `pointerup` listener now survives until the next press or unmount. Mouse capture delivers the release to the window in practice, and any later press calls `cleanupRef.current?.()` first, so a stale listener cannot accumulate. Judged acceptable; `blur` still cancels the alt-tab-away case.
- **SSH / remote / local.** Pure renderer input handling. No PTY, filesystem, process, or network path involved; identical for local, SSH and remote-runtime worktrees.
- **Agents / integrations / git providers.** None touched. The tab strip is provider- and agent-neutral.
- **Performance.** Strictly less work than before: one listener callback no longer tears down the gesture. No new allocations, timers, or hot-path work.
- **UI quality.** No visual change — no markup, tokens, styles, or layout touched. Only which click is honoured.

## Security Audit

Low risk. No input parsing, no command execution, no path handling, no auth, secrets, IPC, or dependency changes. The diff adds one local function and re-points one existing listener registration within the same module. No data crosses a process or trust boundary, and nothing new is rendered from untrusted input.

## Notes

- No platform-specific behaviour. Verified on macOS; the mechanism is Chromium focus transfer from a `<webview>` guest and is not macOS-specific.
- Behaviour is unchanged when the browser page was never focused — the first click already worked and still does. A control case covering this is part of the verification.
- X: https://x.com/nduongg04

## ELI5

If a browser pane had keyboard focus, the first click on another tab did nothing and you needed a second click. This fixes the focus/blur handling so one click switches tabs.

### DIFF
--- a/src/renderer/src/components/tab-bar/tab-strip-pointer-activation.test.tsx
+++ b/src/renderer/src/components/tab-bar/tab-strip-pointer-activation.test.tsx
@@ -4,6 +4,7 @@ import { act, renderHook } from '@testing-library/react'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import { TAB_DRAG_ACTIVATION_DISTANCE_PX } from '../tab-group/useTabDragSplit'
 import { useTabStripPointerActivation } from './tab-strip-pointer-activation'
+import { isTabStripPointerGestureActive } from './tab-strip-pointer-gesture'
 
 function pointerDownEvent(clientX: number, clientY: number, button = 0): React.PointerEvent {
   return { button, clientX, clientY } as unknown as React.PointerEvent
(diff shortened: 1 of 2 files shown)


Jinwoo-H:
Thanks @nduongg04 for the browser focus handoff fix. Current main now contains the same tab-click behavior in commit df447195e2, with follow-up state-machine hardening for guest focus handoffs. We will retain your contribution as historical credit and request no rebase, retest, or other follow-up. Closing this superseded branch.

Jinwoo-H:
Additional current-main evidence for this closure: merged browser-preview work #16920 (commit b19a397d3e7) contains the one-focus-handoff behavior now present in the tab-strip pointer activation path.

PR #10811, opened 2026-07-27
fix(terminal): quarantine input after a replaced PTY endpoint (STA-2373)

brennanb2025:
Follow-up to #10065, which was merged with this failure mode known and deliberately deferred.

## The problem

#10065 made a dead daemon socket recoverable: the pane remounts and reattaches to a respawned daemon. Live QA proved the mechanism works, and pinned the one new failure mode it introduces.

Keystrokes in flight during the ~1.1s recovery window are dropped, but everything typed after reattach lands on a **fresh shell**. There is no input buffering or replay. So the surviving tail of a half-sent line is submitted by the user's own Enter:

- Typing `echo hi; rm -rf x` and losing only the head yields `cho hi; rm -rf x`
- zsh fails `cho` — and **still runs `rm -rf x`** after the `;`

Before #10065 the whole line was silently lost, so the executing tail is genuinely new. Burst typing loses more than slow typing, because more of the burst falls inside the recovery window.

Sending Ctrl-U on reattach does not help: the tail arrives *after* reattach, so the user's own Enter still submits it.

## The fix

Quarantine the remainder of the interrupted line so a mangled command can never be submitted.

State is keyed by **tab**, not pane — recovery destroys the xterm being typed into, and the successor pane is the one that receives the tail.

Armed **only when the endpoint was actually replaced** (the `onWriteUnavailable` path). A stalled-pipeline remount reattaches to the same live shell, where the half-typed line is still on screen and intact; quarantining there would eat a legitimate command.

Three disarm paths, so input can never wedge:

| Disarm | Why |
| --- | --- |
| Line terminator (`\r`, `\n`, `\x03`) | Dropped too — it is the byte that would have submitted the mangled line |
| 700ms idle gap | Reattach takes ~1.1s and the tail arrives <100ms apart; a human reacting to a recovered pane takes far longer |
| 5s absolute cap | Safety valve |

`lastInputAt` starts `null` so the idle gate cannot fire on the reattach delay itself.

The `onData` check sits **after** the `isTerminalQueryReply` branch, so CPR/DSR device replies always reach the shell — dropping those would hang a program that queries during reattach.

## Known edge, left as-is

If the first keystroke after the death is Esc or Ctrl-C, recovery routes through the `sendInputAccepted` rejection instead, which does not set `endpointReplaced`, so the quarantine does not arm. That path means the line is being abandoned anyway, so widening the trigger costs more than it buys.

## Testing

- `terminal-input-quarantine.test.ts` — 13 tests, including the exact `echo hi; rm -rf x` -> `cho hi; rm -rf x` vector from live QA, the paste-with-embedded-terminator case, and both wedge guards
- `terminal-pane-recovery.test.ts` — 4 wiring tests proving it arms on a replaced endpoint and does **not** on a same-shell reattach, a stalled pipeline, or a failed remount
- Full `terminal-pane` suite: **188 files / 2461 tests passed**, no regressions
- `pnpm run tc` clean; `oxlint` and `check:max-lines-ratchet` clean

Live QA against a real daemon death is running now; I will post the result here.

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-connection.test.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.test.ts
@@ -4803,6 +4803,42 @@ describe('connectPanePty', () => {
     _resetTerminalPaneRecoveryForTests()
   })
 
+  it('quarantines the interrupted line after a write-unavailable remount, but never device replies', async () => {
+    const { connectPanePty } = await import('./pty-connection')
+    const { _resetTerminalPaneRecoveryForTests } = await import('./terminal-pane-recovery')
+    _resetTerminalPaneRecoveryForTests()
+    const remountTerminalTabForRecovery = vi.fn<(tabId: string) => boolean>(() => true)
+    mockStoreState = { ...mockStoreState, remountTerminalTabForRecovery } as StoreState
+    const transport = createMockTransport('daemon-pty')
+    let writeUnavailable: (() => void) | undefined
+    transport.connect.mockImplementation(async ({ callbacks }: { callbacks: ConnectCallbacks }) => {
+      writeUnavailable = callbacks.onWriteUnavailable
+      return { id: 'daemon-pty' }
+    })
+    transportFactoryQueue.push(transport)
+    const pane = createPane(1)
+
+    connectPanePty(pane as never, createManager(1) as never, createDeps() as never)
+    await flushAsyncTicks(6)
+    writeUnavailable?.()
+    await flushAsyncTicks(6)
+    expect(remountTerminalTabForRecovery).toHaveBeenCalledWith('tab-1')
+
+    // The surviving tail of `echo hi; rm -rf x`: reaching the fresh shell would
+    // let the user's own Enter run `rm -rf x` (#10065 follow-up).
+    sendTerminalInputThroughPane(pane, 'cho hi; rm -rf x')
+    expect(transport.sendInput).not.toHaveBeenCalledWith('cho hi; rm -rf x')
+    // A program that queries during reattach hangs if its reply is dropped.
+    sendTerminalInputThroughPane(pane, '\x1b[3;1R')
+    expect(transport.sendInputImmediate).toHaveBeenCalledWith('\x1b[3;1R')
+    sendTerminalInputThroughPane(pane, '\r')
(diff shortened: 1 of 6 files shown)


brennanb2025:
## Live QA: A/B against a real daemon death

Ran the isolated-profile rig (`ORCA_DEV_USER_DATA_PATH` scopes the socket **and** pid file, so only this instance's daemon is killable — gated on the `ps` command matching the QA worktree and not `/Applications/Orca.app`).

Asserted via **filesystem markers** rather than scraping the terminal buffer, since the WebGL renderer keeps no `.xterm-rows` in the DOM. Vector: `echo hi; echo TAIL > /tmp/qa_tail_marker` typed after the daemon is killed — a non-destructive stand-in for `echo hi; rm -rf x`.

| Step | Without fix (`git revert HEAD`) | With fix |
| --- | --- | --- |
| Control command before kill | marker created | marker created |
| Daemon killed → respawned | pid 80055 → new | pid 42970 → 69589 |
| **Tail marker after recovery** | **`EXISTS` — tail executed** | **`MISSING` — suppressed** |
| Fresh command afterwards | — | marker created (no wedge) |

### Bug reproduced without the fix

The head `ec` was eaten by the recovery window, so `ho hi; echo TAIL > …` reached the fresh shell:

```
ho h
brennanbenson /private/tmp/orca-qa-repo.KIZNwl [main] $ ho hi; echo TAIL > /tmp/qa_tail_marker
zsh: command not found: ho
brennanbenson /private/tmp/orca-qa-repo.KIZNwl [main] $
```

zsh failed `ho` **and still ran the tail after the `;`** — `/tmp/qa_tail_marker` was created. This is exactly the `rm -rf x` scenario.

### With the fix

The mangled line never reaches the shell at all — no `ho hi; …` line, no `command not found`, and no marker file. The terminal is provably alive on the respawned daemon: the very next command typed after the quarantine released ran normally and created its marker.

So the missing marker is genuine suppression, not a dead pane — which is the failure mode this A/B was designed to rule out.

### Not covered

- macOS only; Linux/Windows unverified
- Local daemon only — no SSH/relay endpoint replacement
- The Esc/Ctrl-C first-keystroke edge documented in the description was not exercised

PR #10816, opened 2026-07-27
fix(mobile): deliver the notifications a reconnect missed, and never persist a watermark past them

### LINKED ISSUE #8591: Mobile reconnect is socket-only: missed notifications, stuck terminal driver, stale worktree list (#8129 #4500 #8498 #6784)
## Summary

When the Orca mobile app returns from background / sleep, it re-establishes the WebSocket but performs **no resync of desktop-held state**. The result: missed notifications, a terminal whose keyboard is locked to a dead phone session, and a Projects/worktree list that shows stale or incomplete data until a focus poll happens to correct it. This is the common root cause behind several reported mobile "reconnect" bugs.

## Root cause (verified at the code level)

Mobile reconnect is **socket-only**. `notifyForeground()` (`mobile/src/transport/rpc-client.ts:1215`) reopens the WS and re-runs the E2EE handshake, but performs no state replay. Three failure modes follow from this single gap:

### 1. Notifications dropped with no replay ( #8129 )
`dispatchMobileNotification` (`src/main/runtime/orca-runtime.ts:7359`) is a synchronous `for` loop over an ephemeral `Set` of listeners (`notificationListeners`, `orca-runtime.ts:2221`). There is no queue, outbox, or replay. If the phone socket is dead at dispatch time, the event is emitted to a closed transport and silently lost. On reconnect the client re-subscribes fresh and never receives the missed event.

### 2. Terminal driver lock cleared only on socket reap; can stick permanently ( #4500 )
`isTerminalInputLockedForClient` (`src/main/runtime/rpc/methods/terminal.ts:294-309`, lock at `:308`) returns true while `getDriver(ptyId).kind === 'mobile'`. Desktop keystrokes hit the guard and return `{accepted:false, bytesWritten:0}` (`terminal.ts:1121-1128`; binary path `:2602-2603`). The driver only clears on `onClientDisconnected → setDriver(idle)` (`orca-runtime.ts:7926-7984, 8184-8210`).

- A genuine disconnect IS reaped and released within ~30s by the heartbeat (`HEARTBEAT_INTERVAL_MS = 15_000`, `ws-transport.ts:37`; `terminate()` at `:239-252`).
- **Permanence is conditional**, not absolute: `onPong` and `onMessage` both re-arm liveness on *any* inbound frame (`ws-transport.ts:309-316`). If a proxy or unrelated traffic keeps the socket "warm," `terminate()` never fires and the desktop input stays locked.

### 3. Worktree list has no refresh and reuses a poisoned cache ( #8498 )
The host screen seeds its list from `getCachedWorktrees` at mount (`mobile/app/h/[hostId]/index.tsx:138`) and only refetches via a focus-gated 3s poll (`index.tsx:566-597`). There is **no `RefreshControl`/`onRefresh`** on the host worktree/Projects list (grep: `RefreshControl` exists only in `accounts.tsx:274` and `MobileAgentSessionHistoryList.tsx`; `onRefresh` only in `MobileSourceControlPanel.tsx:223`).

Root cause deeper than "no refresh": `setCachedWorktrees` is called **only from the home screen** (`mobile/app/index.tsx:201,369`), never the host detail screen. So a partial home-screen fetch poisons the host view until the focus poll corrects it.

### 4. Real connection error swallowed ( #6784 )
`classifyConnection()` (`mobile/src/transport/connection-health.ts:45`) computes its verdict purely from `state` + `reconnectAttempts` + `lastConnectedAt`. The actual socket error is only `console.log`'d in `ws.onerror` (`rpc-client.ts:689-695`) and never surfaced to the user — the UI shows a canned "Reconnecting…/Can't reach desktop" instead of the real failure.

## Proposed fixes (design, for discussion)

- **4.1 Notification catch-up (#8129):** add a monotonic `notificationSeq`; buffer dispatched events per-client on disconnect; on `e2ee_authenticated` the client sends `lastSeen seq` and the runtime replays `seq > lastSeen`. **Must dedupe by `notificationId`/`seq` client-side** — the mobile socket flickers, so events delivered during brief liveness must not be re-pushed on reconnect (duplicate local notifications).
- **4.2 Driver fast-fail (#4500):** harden the heartbeat liveness check — treat a socket that has not sent a **pong** within `HEARTBEAT_INTERVAL_MS` as released, independent of inbound app-frame traffic (distinguish pong-liveness from message-liveness at `ws-transport.ts:239-252`). Also reclaim the driver on a foreground/reconnect probe from a different client for the same `ptyId`.
- **4.3 Worktree resync + refresh (#8498):** add a `connState` effect that refetches worktrees on transition into `connected`; add a `RefreshControl` to the SectionList (`index.tsx:1177`); make the reconnect refetch write through (or bypass) `setCachedWorktrees` so it cannot serve a stale snapshot.
- **4.4 Surface the real error (#6784):** pass `ws.onerror`'s message into `classifyConnection` so the UI shows the actual failure instead of a canned verdict.

## Blast radius

This single "socket-only reconnect" gap underlies #8129, #4500, #8498, #6784 and several related reports (#6928, #8489, #8371, #2157, #5049, #5070). Fixing the resync gap closes the class, not just the instances.

## Verification

All claims above were independently verified against source by an adversarial review pass (every cited `file:line` re-read and confirmed; no fabricated/drifted line numbers). This issue documents the finding; the four fixes are proposed and should be agreed before implementation.

---
*Generated from an investigation spec; full writeup with citations available on request.*

nwparker:
Closes #8591

## What's broken

Mobile silently stops delivering notifications. The desktop dispatches them, the phone is connected, and nothing arrives — no error, no banner, nothing to see.

## ELI5

The phone keeps a bookmark: "I've shown everything up to notification #57." On reconnect it asks the desktop "anything after #57?"

The desktop's counter **restarts at 0 every launch**. So after a desktop restart, the desktop is on notification #3 while the phone still holds a bookmark saying #57. The phone asks for "anything after 57", the desktop says "nope", and notifications stay dead until the new desktop out-counts the old bookmark — potentially forever.

A bookmark is meaningless without knowing *which* counter it indexes. So the bookmark now records the counter's lifetime (an "epoch") alongside the number. Different epoch → the bookmark is void → replay everything retained.

## The fixes

1. **Epoch-tagged watermark.** A seq is stored with the counter lifetime it belongs to. A mismatch on reconnect voids it instead of trusting it.
2. **Atomic pair.** Seq and epoch live in **one** JSON value. Written as two keys, a process death between the writes leaves an epoch from one counter beside a seq from another — internally consistent, silently wrong, and therefore trusted.
3. **Survives the app's own teardown.** `app/index.tsx` tears the subscription down on every non-connected state, destroying the watermark and seen-set exactly when a reconnect needs them. They now live per-host at module scope.
4. **Cold-open catch-up** keyed on the record's *existence*, not `seq > 0` — a voided watermark reloads as seq 0 and would otherwise look like a first-ever pairing.
5. **Serialized delivery** (from review) — see trade-offs.

## Fix proof

Full mobile suite: **344 files, 2499 passed, 2 skipped.** `tsc` clean, `oxlint` clean.

Every mechanism was disabled individually against the *unchanged* suite. Each kills exactly one test — so nothing is unguarded, and nothing is redundant:

| Mechanism disabled | Result |
|---|---|
| Batch-as-one-queue-entry → per-item enqueue | ordering test fails |
| Watermark advance → moved before the await | ordering test fails |
| Seed timeout removed | wedged-read test fails |
| Live-path dedup claim removed | concurrent-dedup test fails |
| Replay-path dedup claim removed | cross-path dedup test fails |

## Trade-offs / regression risk

**Two defects found in review were pre-existing, not introduced here.** I verified this mechanically (`git show origin/main:…` shows the identical watermark advance) and fixed them anyway, because this branch is what makes the watermark load-bearing:

- **Watermark could outrun delivery.** The advance happened *before* the local notification was shown, and replay + live delivery ran concurrently. A live seq 11 handled while catch-up was still showing seq 6 persisted 11 — a process death before 7–10 were shown lost them permanently. Confirmed with a disposable probe (`expected 11 to be less than 7`) before writing any fix.
- **Unbounded seed.** Every delivery awaits the persisted read, so an AsyncStorage read that never settled disabled the host's notifications for the entire app lifetime. Now bounded at 3s; a late seed still applies when it lands. Proceeding unseeded is strictly better — the watermark stays 0, so catch-up over-fetches and the seen-set de-duplicates.

**One real behavioral trade-off:** serializing deliveries removed an overlap the duplicate-suppression relied on. `showLocalNotification` deduped two same-id events by observing the first still pending when the second arrived; serialized, the first *completes* first, so the second saw nothing pending and scheduled a second banner. The claim moved to enqueue time, where the overlap is still observable. Dismisses are deliberately **not** claimed — a dismiss is what retires the notification a show created.

**Cost:** notifications for a single host now show strictly one-at-a-time rather than concurrently. The catch-up RPC is deliberately kept *outside* the queue — `sendRequest` waits up to 30s, and holding the chain for that would stall live delivery on a slow link.

**Accepted:** delivery is at-least-once. `saveWatermark` is fire-and-forget, so a failed write leaves the stored value *below* what was delivered and a later cold start may re-show. That's the safe direction — the alternative loses notifications.

## Review loop

3 rounds (gpt-5.6-sol + fable, alternating). Final fable verdict: **CLEAN** — "mergeable from my side." The two GPT MAJORs are fixed and mutation-verified above. Residual from review: 1 MINOR (notification-burst pacing), 3 NITs.

## Cross-platform

Mobile-only (`mobile/`). Platform-independent — no path handling, no shell, no OS branching. Android and iOS both exercised by the suite.

## Perf

No polling, no new hot-path scans. The dedup set is bounded at 512 entries. The serialization chain is one promise per host.

### DIFF
--- a/mobile/src/notifications/local-notification-scheduling.ts
+++ b/mobile/src/notifications/local-notification-scheduling.ts
@@ -1,0 +1,191 @@
+import * as Notifications from 'expo-notifications'
+import { Platform } from 'react-native'
+import { loadPushNotificationsEnabled } from '../storage/preferences'
+import { buildLocalNotificationData, type DesktopNotificationSource } from './notification-routing'
+import { ensureNotificationPermissions } from './notification-permissions'
+
+export type NotificationEvent = {
+  type: 'notification'
+  source: DesktopNotificationSource
+  title: string
+  body: string
+  worktreeId?: string
+  notificationId?: string
+  // Desktop-assigned seq for reconnect catch-up (#8129); optional since older runtimes may omit it.
+  notificationSeq?: number
+  // Counter lifetime the seq belongs to (#8591); absent on older runtimes.
+  notificationEpoch?: string
+}
+
+export type DismissNotificationEvent = {
+  type: 'dismiss'
+  notificationId: string
+  notificationSeq?: number
+  notificationEpoch?: string
+}
+
+type ScheduledNotificationState = {
+  identifier?: string
+  pending?: Promise<string | null>
+  dismissAfterSchedule?: boolean
+}
+
+const scheduledNotificationsByHostAndNotificationId = new Map<string, ScheduledNotificationState>()
+
+// Why: keys never repeat and are only freed on desktop dismiss (which remote users often miss), so bound the map to stop unbounded growth.
+const MAX_SCHEDULED_NOTIFICATIONS = 256
+let maxScheduledNotifications = MAX_SCHEDULED_NOTIFICATIONS
+
+function getStoredNotificationKey(hostId: string, notificationId: string): string {
+  return `${encodeURIComponent(hostId)}:${encodeURIComponent(notificationId)}`
+}
+
+// Evict oldest settled entries (never mid-schedule); Map iteration is insertion order so the first match is oldest.
+function boundScheduledNotifications(): void {
+  while (scheduledNotificationsByHostAndNotificationId.size > maxScheduledNotifications) {
+    let evicted = false
+    for (const [key, state] of scheduledNotificationsByHostAndNotificationId) {
+      if (!state.pending) {
+        scheduledNotificationsByHostAndNotificationId.delete(key)
+        evicted = true
+        break
+      }
+    }
+    if (!evicted) {
+      break
+    }
+  }
+}
+
+/** Test-only: override the cap (pass no arg to restore the default). */
+export function setScheduledNotificationsMaxForTests(max?: number): void {
+  maxScheduledNotifications = max ?? MAX_SCHEDULED_NOTIFICATIONS
+}
+
+export function configureNotificationChannel(): void {
+  if (Platform.OS === 'android') {
+    void Notifications.setNotificationChannelAsync('orca-desktop', {
+      name: 'Desktop Notifications',
+      importance: Notifications.AndroidImportance.HIGH,
+      vibrationPattern: [0, 250],
+      lightColor: '#6366f1'
+    })
+  }
+}
+
+export async function showLocalNotification(
+  event: NotificationEvent,
+  hostId: string
+): Promise<void> {
+  const storedKey = event.notificationId
+    ? getStoredNotificationKey(hostId, event.notificationId)
+    : null
+
+  if (!storedKey) {
+    const enabled = await loadPushNotificationsEnabled()
+    if (!enabled) {
+      return
+    }
+
+    const granted = await ensureNotificationPermissions()
+    if (!granted) {
+      return
+    }
+
+    await Notifications.scheduleNotificationAsync({
+      content: {
+        title: event.title,
+        body: event.body,
+        data: buildLocalNotificationData(event, hostId),
+        ...(Platform.OS === 'android' ? { channelId: 'orca-desktop' } : {})
+      },
+      trigger: null
+    })
+    return
+  }
+
+  let state = scheduledNotificationsByHostAndNotificationId.get(storedKey)
+  if (state?.pending) {
+    return
+  }
+  if (!state) {
+    state = {}
+    scheduledNotificationsByHostAndNotificationId.set(storedKey, state)
+  }
+  const notificationState = state
+
+  const pending = (async () => {
+    const enabled = await loadPushNotificationsEnabled()
+    if (!enabled) {
+      return null
+    }
+
+    const granted = await ensureNotificationPermissions()
+    if (!granted) {
+      return null
+    }
+
+    if (notificationState.identifier) {
+      await Notifications.dismissNotificationAsync(notificationState.identifier).catch(() => {})
+      notificationState.identifier = undefined
+    }
+
+    return Notifications.scheduleNotificationAsync({
+      content: {
+        title: event.title,
+        body: event.body,
+        data: buildLocalNotificationData(event, hostId),
+        ...(Platform.OS === 'android' ? { channelId: 'orca-desktop' } : {})
+      },
+      trigger: null
+    })
+  })()
+  notificationState.pending = pending
+
+  try {
+    const scheduledIdentifier = await pending
+    if (!scheduledIdentifier) {
+      if (!notificationState.identifier) {
+        scheduledNotificationsByHostAndNotificationId.delete(storedKey)
+      }
+      return
+    }
+    if (notificationState.dismissAfterSchedule) {
+      notificationState.dismissAfterSchedule = false
+      scheduledNotificationsByHostAndNotificationId.delete(storedKey)
+      await Notifications.dismissNotificationAsync(scheduledIdentifier).catch(() => {})
+      return
+    }
+    notificationState.identifier = scheduledIdentifier
+    boundScheduledNotifications()
+  } finally {
+    if (notificationState.pending === pending) {
+      notificationState.pending = undefined
+      notificationState.dismissAfterSchedule = false
+    }
+  }
+}
+
+export async function dismissLocalNotification(
+  event: DismissNotificationEvent,
+  hostId: string
+): Promise<void> {
+  if (!event.notificationId) {
+    return
+  }
+  const storedKey = getStoredNotificationKey(hostId, event.notificationId)
+  const state = scheduledNotificationsByHostAndNotificationId.get(storedKey)
+  if (!state) {
+    return
+  }
+  if (state.pending) {
+    // Why: dismiss can arrive while the OS is still scheduling; defer it so no stale banner survives.
+    state.dismissAfterSchedule = true
+    return
+  }
+  if (!state.identifier) {
+    return
+  }
+  scheduledNotificationsByHostAndNotificationId.delete(storedKey)
+  await Notifications.dismissNotificationAsync(state.identifier).catch(() => {})
+}
--- a/mobile/src/notifications/mobile-notifications.test.ts
+++ b/mobile/src/notifications/mobile-notifications.test.ts
@@ -73,10 +73,14 @@ describe('subscribeToDesktopNotifications', () => {
     vi.clearAllMocks()
   })
 
-  async function flushAsync(): Promise<void> {
-    for (let i = 0; i < 10; i += 1) {
-      await Promise.resolve()
-    }
+  // Why the macrotask and not N microtask ticks (#8591): deliveries now run through
+  // the per-host serialization queue, so a delivery is several more `await` hops deep
+  // than it used to be and a fixed tick count silently under-drains. Yielding to the
+  // macrotask queue drains whatever depth the chain happens to have.
+  function flushAsync(): Promise<void> {
+    return new Promise((resolve) => {
+      setTimeout(resolve, 0)
+    })
   }
 
   function makeDeferred<T>(): { promise: Promise<T>; resolve: (value: T) => void } {
--- a/mobile/src/notifications/mobile-notifications.test.ts
+++ b/mobile/src/notifications/mobile-notifications.test.ts
@@ -494,6 +494,122 @@ describe('subscribeToDesktopNotifications — reconnect catch-up', () => {
     expect(scheduledIds.filter((id) => id === 'agent:dup')).toHaveLength(1)
   })
 
+  it('voids a persisted watermark whose epoch predates a desktop restart', async () => {
+    // #8591: the desktop's seq counter restarts at 0 each launch while this watermark
+    // is persisted. Reconnecting to a restarted desktop with seq 57 would make
+    // `57 >= 2` true and silently kill catch-up. The epoch on 'ready' is what tells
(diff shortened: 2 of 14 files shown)


nwparker:
## Independent verification of the three open bot findings (Opus sub-agent, executed not read)

Verified against a detached worktree pinned to this PR's head (`5274388819`), A/B'd against `origin/main`. Same discriminator as always: does `origin/main` handle this correctly? If not, the finding describes a *residual* gap, not a regression this PR introduces.

### 🟡 CLAIM (Major): a live event can persist a watermark past unshown replayed events — **REAL, but a RESIDUAL gap, not a regression**

The race is exactly as described, and I reproduced the interleave directly. `getMissedSince` sits outside the serialization chain (deliberately — it waits up to 30s and holding the chain would stall live delivery). During that window the chain is empty, so a live event runs immediately and persists its higher seq; the replayed 7..10 then enqueue *behind* it. Observed order:

```
live-11-persists-seq  →  request-returned  →  replay-7..10
```

A process death between those points loses 7..10 — the exact invariant the serialization is meant to hold.

**But this is strictly better than `origin/main`, not worse.** On main there is no chain at all *and* the watermark advances **before** `showLocalNotification` is awaited:

```ts
// origin/main
if (event.notificationSeq != null && event.notificationSeq > lastDeliveredSeq) {
  lastDeliveredSeq = event.notificationSeq
  void saveLastSeenSeq(hostId, lastDeliveredSeq)   // ← persisted BEFORE the notification is shown
}
```

This PR moves the advance to *after* the await and serializes deliveries. It narrows the window from "every notification" to "only events landing inside the catch-up request window." So the finding is a fair description of what's left — worth a follow-up (claim a chain slot before issuing the request), but it does not make this diff a regression.

### 🟢 CLAIM (Major): unguarded rejection wedges the delivery pipeline — **REAL but MISCHARACTERIZED; the pipeline does not wedge**

The bot's framing is "one unguarded rejection path runs the whole delivery pipeline." The wedge half is wrong. `enqueueHostDelivery` already sheds the failure from the chain tail:

```ts
const run = session.deliveryTail.then(task)
session.deliveryTail = run.catch(() => {})   // ← tail swallows it
return run                                    // ← caller's copy still rejects
```

Verified: after a task rejects, the very next `enqueueHostDelivery` still runs. **The chain survives.** What's actually true is the narrower claim — the *returned* promise rejects, and the `void`-ed IIFE at L226-237 doesn't catch it, so a failing OS schedule call surfaces as an unhandled rejection warning. That's cosmetic-to-log-noise, not a dropped-notification or wedged-queue defect. A `.catch(() => {})` is still worth adding, but the severity is Minor, not Major.

### 🟡 CLAIM (Minor): `void`-ed `setNotificationChannelAsync` — **REAL, cosmetic**

Accurate as written. Android channel creation can reject on some OEMs and the `void` discards it into an unhandled-rejection warning at startup. One-line fix, no behavioral risk.

---

### Verdict

**No regressions found. The two Majors are one residual gap and one overstated severity.** The committed suite is green — 45 tests across 6 files, all passing.

The honest caveat on merge confidence isn't the bot findings, it's coverage: **`mobile/` is a separate package with its own vitest config, and this PR's CI ran only the `verify` job.** I could not confirm from the check names that the mobile suite executes in CI at all. Someone should verify that before merging, because if it doesn't, these 45 tests are passing only on developer machines.

Recommendation: mergeable on the merits once (a) the mobile suite is confirmed to run in CI, and (b) ideally the two one-line `.catch()` additions go in, since they're free. The catch-up-window race deserves its own follow-up issue rather than being folded in here — the fix (claiming a chain slot before the request, or gating on a `catchupPending` promise) needs its own timeout-deadlock analysis, since the request can hang for 30s.

*Verification was read-only against a pinned detached worktree; nothing was pushed and no PR was merged.*

nwparker:
### Correction to my previous comment: the mobile suite **does** run in CI

I flagged as an open question whether `mobile/`'s separate vitest package actually executes in CI. It does — I checked rather than leaving it hanging.

There are **two distinct `verify` jobs** on this PR, from different workflows, and the check-name collision is what made it ambiguous:

```
Mobile Checks / verify   SUCCESS   ← .github/workflows/mobile.yml
PR Checks / verify       SUCCESS   ← the desktop suite
Track Community PRs      SUCCESS
```

`.github/workflows/mobile.yml` triggers on `paths: ['mobile/**']` — which is precisely this PR's diff — runs with `working-directory: mobile`, and executes `pnpm test` (plus `pnpm lint`). So the 45 mobile tests genuinely ran on this head, on CI, not just locally.

**That removes the caveat from my verdict.** With CI coverage confirmed and no regressions found against `origin/main`, the merge blocker I raised is gone. What remains is optional polish: the two one-line `.catch(() => {})` additions, and a follow-up issue for the catch-up-window race (which, to restate, is narrower on this branch than on `main`).

nwparker:
## The two `.catch()` additions, made concrete

I called these "free" in my verdict; here they are precisely, so nobody has to re-derive them. Both are the same one-line shape on a `void (async () => {…})()` whose body can reject.

**1. `mobile-notifications.ts` L227-237** — the live-event continuation:

```diff
     void (async () => {
       await session.watermarkSeeded
       if (disposed) {
         return
       }
       await queueDelivery(
         liveEvent.type === 'notification' ? 'notification' : 'dismiss',
         liveEvent as NotificationEvent | DismissNotificationEvent
       )
-    })()
+    })().catch(() => {})
```

**2. `mobile-notifications.ts` L190-205** — the `ready` continuation, same shape, where `fetchMissed()` is the rejecting call.

**3. `local-notification-scheduling.ts` L74** — `void Notifications.setNotificationChannelAsync(...)` → append `.catch(() => {})`. Android channel creation rejects on some OEMs.

### What these do and don't fix

To restate the correction from my earlier comment, since "add a catch" invites the assumption that something was broken without it: **the delivery chain does not wedge today.** `enqueueHostDelivery` already sheds the failure from the tail —

```ts
session.deliveryTail = run.catch(() => {})   // tail is protected
return run                                    // caller's copy still rejects
```

— and I verified a subsequent delivery still runs after a task rejects. So these three lines suppress **unhandled-rejection warnings**, nothing more. They're worth taking because they're zero-risk and quiet the logs, not because they close a defect. The bot's "one unguarded rejection path runs the whole delivery pipeline" framing overstates it.

I have not pushed these — this isn't my branch, and the change is small enough that whoever owns the PR should just take it directly.

### Still the one open item worth a follow-up issue

The catch-up-window race (a live event persisting a watermark while `getMissedSince` is in flight outside the chain). Real, reproduced, but **narrower on this branch than on `origin/main`**, which has no chain at all *and* advances the watermark before awaiting `showLocalNotification`. It needs its own timeout-deadlock analysis before anyone claims a chain slot ahead of a request that can hang for 30s — which is exactly why it doesn't belong in this diff.

nwparker:
## I tested the bot's suggested fix for the watermark race. **Don't apply it as written.**

The suggestion on the Major thread was to *"claim a chain slot before issuing the request, or gate on a `catchupPending` promise."* That closes the race, but I modelled it against the real `enqueueHostDelivery` semantics and it introduces a worse problem than the one it fixes.

**Claiming the slot first means the in-flight request holds the chain.** Every live delivery queues behind it:

```
enqueue(chain, async () => { await getMissedSince(); replay() })   // slot claimed first
enqueue(chain, async () => { deliverLive() })                      // arrives during the request
→ after 20ms: []          ← nothing delivered; live notification is BLOCKED
→ after release: ['replay', 'live']
```

Ordering becomes correct, but **nothing reaches the user for as long as the request takes** — and `sendRequest` waits up to **30 seconds**. On a slow or flaky link that is a guaranteed half-minute of dead notifications on every reconnect, in exchange for closing a race that requires a process death inside a narrow window. That trade is bad in the common case.

Note this is precisely the rationale the PR already documents for keeping the request outside the chain:

> *Why the request stays OUTSIDE the queue: sendRequest waits up to 30s, and holding the chain for that would stall live delivery on a slow link.*

The existing comment is right, and the bot's suggestion would undo a deliberate decision.

**One thing the suggestion gets right:** a *rejecting* request doesn't wedge anything — I verified the chain still runs the next task after a timeout rejection, because the tail swallows it. So the risk is a stall, not a permanent wedge. Still a stall on the default path.

### What a follow-up should actually do

Not "claim the slot earlier." The shape that preserves both invariants is to keep the request outside the chain and make the *watermark advance* conditional — e.g. have a live delivery landing during an active catch-up window hold back its `saveWatermark` until the replay drains, rather than blocking the delivery itself. The user still sees the notification immediately; only the persisted seq waits.

That needs its own design and its own tests, which is exactly why it belongs in a separate issue rather than folded into this PR. **My merge verdict on #10816 is unchanged: mergeable.** The race is residual and strictly narrower than `origin/main`, which has no chain at all *and* persists the watermark before the notification is shown.

Issue #4375, opened 2026-06-01
[Bug]: Non turn-end notification spam


ryanmargono:
### Operating system

macOS

### Details

Getting notifications on non-terminal turn ends, causing too much noise and having to disable to feature.
Ie, after tool calls or reasoning, I will get a notification. 
Should be the case (or a preference) where I am only notified on a terminal agent turn or if theres an issue.

vthuan1889:
**Also reproduces with Cursor CLI over SSH remote**

I see the same noisy notification behavior, not only on local terminals.

**Setup**
- **Client:** Windows, Orca **1.4.38.0**
- **Worktree:** SSH remote worktree → **Ubuntu** (not local host PTY)
- **Agent:** `cursor-agent` / Cursor CLI running on the **remote** host over SSH

**What I see**
- Notifications fire after **intermediate steps** (tool calls, reasoning, subagent/tool output), not only when the agent has **fully finished** a user-visible turn.
- Enough noise that I had to **turn notifications off**, which defeats the purpose of long-running remote agents.

**Expected**
- Notify on **terminal turn end** (agent done / waiting for input), or on **error/blocker** — same as described in the original report.
- Ideally a preference: "notify only on terminal turn" vs "every hook/status event" (remote Cursor hooks may amplify this on SSH).

**Notes**
- Remote hooks are installed under `~/.cursor/hooks.json` on the SSH target (Orca's Cursor hook install path), so status events may be more granular over SSH than on a simple local PTY.
- Happy to capture logs or a short screen recording if that helps triage SSH + Cursor CLI.

totally-tim:
same here for codex cli running locally within orca. Frequent tool-use (bash etc.)/mid turn notifications.

<img width="227" height="24" alt="Image" src="https://github.com/user-attachments/assets/21c8e949-720c-4a4f-98f1-788b09ec7d36" />

mitsuksw:
I can reproduce this with Codex CLI running locally inside Orca on Windows.

This appears broader than PermissionRequest / waiting / blocked states. I am getting `agent-task-complete` desktop notifications while Codex is visibly still in a working turn and while the active tool is still running.

Setup:

- OS: Windows
- Agent: Codex CLI running locally inside Orca terminal
- Worktree: local repo
- Notification shown by Orca:
  - Title: `danmaku_monitor / main - Codex finished`
  - Body: `Using Bash: npm run test:submit`

Actual behavior:

[...]

PR #10820, opened 2026-07-27
fix(notifications): stop Codex banner spam on non-turn-end Stop hooks

nwparker:
Partially addresses #4375 — **deliberately does not close it.** See [Scope](#scope-what-this-does-not-fix).

## What's broken

Codex fires a desktop notification on **non-turn-end** events. As the reporter put it: "after tool calls or reasoning, I will get a notification" — enough noise that they turned the feature off.

This PR fixes **one** of the two ways that happens: a lead `Stop` raised while sub-agents are still running. The other — a bare mid-turn `Stop` with no sub-agent at all — is still open and pinned by a test below.

## ELI5

Codex's lead agent emits a `Stop` hook when *it* finishes — but its sub-agents may still be running, so the turn isn't over. Orca treated that `Stop` as "task complete" and popped a banner. This teaches it to check whether any children are still alive first.

## Fix proof

**Repro test** (`codex-lead-stop-subagent-gate.test.ts`), driving real hook normalization → store → coordinator → dispatcher:

```
UserPromptSubmit → SubagentStart(child-1) → PreToolUse → Stop → [child keeps working]
```

Before: 1 spurious notification. After: 0. A **control test** in the same file proves a genuine turn end still notifies exactly once — the gate suppresses spam, not completions.

## The interesting part: why this can't be fixed in `state`

My first cut gated by rewriting the lead `Stop`'s `state` from `done` to `working`. That broke two **pre-existing** tests in `src/main/agent-hooks/server.test.ts` (`:4447`, `:6587`), which pin the opposite contract: a root `Stop` retires child rows and reports `done`. Both contracts are real; they can't both live in `state`.

Worse, the reap is what *destroys* the signal. Probing the payload stream showed the lead `Stop` and the final `SubagentStop` are byte-identical after it:

```
Stop(lead):          state=done  subs=undefined   ← must NOT notify
SubagentStop(child): state=done  subs=undefined   ← MUST notify
```

No downstream heuristic can separate those. So the flag is captured **before** the reap and carried as a live-only field; `state` is untouched, and only the notification layer reads it.

## Proof of no regressions

- `server.test.ts` + gate suite: **250 passed**. The two tests my first cut broke are green again, and a new test pins that contract so it can't silently regress.
- Affected areas (`src/shared/`, `agent-hooks/`, `terminal-pane/`, `hooks/`, `relay/`): **7275 passed**.
- Full suite: **36428 passed**. The 8 failures (IME/editor/packaging) reproduce **identically on a clean `origin/main`** in this worktree — stale `node_modules`, unrelated to this diff. Verified by reverting all four source files and re-running.
- `oxlint` clean; `tsc` clean (the one `@sanity/diff-match-patch` error is the same pre-existing missing-dep).

**Mutation testing** — every mechanism is load-bearing, and each kills a *different* test (no redundancy):

| Mutation | Test killed |
|---|---|
| Coordinator gate → `if (false)` | REPRO |
| Drop pre-reap capture (local) | REPRO |
| Don't pass flag to payload builder | REPRO + contract test |
| Drop pre-reap capture (remote) | SSH/relay test |

## Review loop

2 rounds. Round 1 (fable / gpt-5.6-sol) split: fable **CLEAN "ship it"**, gpt **CHANGES_REQUIRED**. I verified gpt's claim directly — the two `server.test.ts` failures were real and mine — and sided with the stricter verdict. Both of gpt's substantive findings are now fixed; round 2 verdicts are in the thread.

## Cross-platform

Pure hook-listener/notification logic in `src/shared/` and the renderer — no paths, no shell, no platform APIs. The `Stop`-gating path is identical on macOS, Linux, and Windows.

**Both transports are covered.** A reviewer caught that SSH/relay panes never reach `normalizeHookPayload`'s codex branch — the main process re-derives their roster in `reconcileRemoteCodexState`, which does its own reap. Without the same capture there, this would have been a local-only fix. The second commit closes that, with its own test.

## Perf

No new hot-path work: one boolean computed from an in-memory roster already being read on that line, on `Stop` events only. No added allocation, I/O, or polling.

## Trade-offs / known-open

- **A bare mid-turn `Stop` with no subagents** that resumes after the 1.5s quiet window still notifies. Out of scope here: there's no principled bound to widen the window to, and widening it delays *every* real Codex completion. The right fix is retracting the banner via `notifications:dismiss` on the resuming `working` hook. Pinned by a test explicitly named `KNOWN-OPEN` (green, asserting today's behavior) so that change flips it — it documents the gap rather than hiding it.
- The flag is **stripped on persist**, so a rehydrated row can't re-suppress the first real completion after a restart.

---

## Update: a regression was reported against this PR — confirmed and fixed (`a6a5076`)

A reviewer reported that this branch flipped a blocked child's pane from `waiting` to `working`. **They were right, and my first check was wrong.** I probed only lead `Stop` (where this branch and `origin/main` agree on `done`) and concluded there was no regression. Re-probing a **non-`Stop`** lead event showed the real divergence:

| lead event, child blocked on a permission prompt | `origin/main` | this branch, before `a6a5076` |
| --- | --- | --- |
| `PostToolUse` | `waiting` | **`working`** ← regression |
| `Stop` | `done` | `done` + flag |

**Cause.** The revert in `84880f2711` removed the `codexRosterEffectiveState` call from the *entire* local lead path, not just the `Stop` case. Collateral damage: a child blocked on a permission prompt stopped holding the pane, so the pane read `working` and the user never saw what it was blocked on.

**Fix.** Restore the call, placed **after** the roster reap:

```ts
// Why: read the roster AFTER the reap. A blocked child must still hold the pane at `waiting` on
// non-Stop lead events, but on Stop the roster is already empty so this stays `done` — the root
// Stop contract in agent-hooks/server.ts depends on that.
const effectiveState = codexRosterEffectiveState(
  state.codexSubagentRosterByPaneKey.get(paneKey),
  stateName
)
```

Both contracts now hold at once, and the branch matches `origin/main` on both paths. The pre-reap flag described above is untouched.

### Ordering is the load-bearing detail — both directions are pinned

New `src/shared/codex-lead-state-vs-subagent-roster.test.ts`:

| mutation | result |
| --- | --- |
| drop the read (the reported regression) | ❌ `holds the pane at waiting on a non-Stop lead event while a child is blocked` fails |
| move the read *before* the reap (what `84880f2711` reverted) | ❌ `still reports done on a lead Stop, flagging that children were live` fails |
| unmutated | ✅ 3/3 |

Each mutation kills exactly the contract it violates — neither test is redundant.

### No regressions from the fix

`src/shared/` + `src/main/agent-hooks/` + gate suite: **3531 passed | 9 skipped | 0 failed** (366 files), including the two `server.ts` root-`Stop` tests whose failure forced the original revert.

Also verified `SessionStart` under the new ordering: a stale blocked child from a dead process is cleared, the restart reports `working`, and the next event does not resurrect it.

### Known bound, measured not assumed

Codex CLI 0.144 can omit a child `SubagentStop`, leaving a stale roster entry that suppresses one genuine completion. I measured the blast radius: **exactly one notification** — the `Stop` reap clears the roster, so the next turn notifies normally. Strictly better than the bug it replaces (a banner on every mid-turn lead `Stop`, indefinitely), and self-correcting.

## Scope: what this does **not** fix

Review correctly caught that the original auto-close link on this PR was overclaiming, so it is now "partially addresses" — merging this must not close the issue.

`#4375` reports notifications "after tool calls or reasoning". There are two distinct paths that produce that symptom:

| path | status |
| --- | --- |
| lead `Stop` while a sub-agent is still working | ✅ fixed here |
| bare mid-turn `Stop`, no sub-agent, resuming after the 1.5s quiet window | ❌ **still open** |

The second is pinned by a deliberately-passing test, `KNOWN-OPEN: a bare mid-turn Stop resuming after the quiet window still notifies`, which asserts today's wrong behaviour so that a real fix flips it. It even asserts the banner carries the *previous* tool name, matching the report.

Why it isn't fixed here: the quiet window has no principled bound to widen to, and widening it delays every genuine Codex completion. The correct fix is retracting the banner via `notifications:dismiss` on the resuming `working` hook — which needs a notification id plumbed back through `dispatchCompletion` (today typed `=> void`, `agent-completion-coordinator-types.ts:30`). That's a separate change with its own blast radius, not a line to sneak into this one.

## Review response: the suppression bound, measured

Review raised that gating on roster membership means a dropped child `Stop` hook (Codex 0.144 can omit them) suppresses a *genuine* completion, and asked for "a bounded/provisional suppression that eventually releases."

I probed the real payload stream rather than reasoning about it. The suppression is **already bounded** — by the `Stop` reap, not by a timer:

| scenario | notifications delivered |
| --- | --- |
| child finishes normally after a gated lead `Stop` | **1** — the child's own `SubagentStop` carries the turn end. Deferred, not lost. |
| child `Stop` hook dropped by the CLI | **0** for that turn — one real completion lost |
| dropped child `Stop`, then a second turn | **1** — self-heals, no user action needed |

So the worst case is one missed notification on one turn, self-correcting. Both rows are now pinned as tests (`flags only the turn a dropped child Stop spans`, `clears the flag as soon as the child does report`).

**Why no aging/liveness timer was added anyway:** a Codex sub-agent can legitimately run for many minutes. Any release window short enough to salvage the dropped-hook turn is shorter than a real long-running child — so it would re-fire the exact spam this gate exists to suppress. A timer would trade a rare, self-healing missed notification for a common, permanent false one. I took the measured bound over the redesign; happy to revisit if you disagree with that trade.

Also folded in the review's MINOR: the blocked-child contract now runs parameterized over `PreToolUse`, `PostToolUse`, and `UserPromptSubmit`, since every lead event normalizing to `working` must yield to a blocked child.

### Test totals after the review response

`src/shared/` + `src/renderer/src/components/terminal-pane/` + `src/renderer/src/hooks/`: **5950 passed | 6 skipped**, plus 257 in the three directly-affected suites. `tsc -p config/tsconfig.node.json` and `oxlint` clean.

⚠️ One pre-existing failure set, **not from this branch**: 5 tests in `terminal-ime-xterm-composition-deduplication.test.ts`. I verified these reproduce identically on the untouched merge base `c67aadbc18` — this branch touches no IME code (6 files changed, all notification/hook paths).

### DIFF
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -6623,6 +6623,47 @@ describe('Last-status persistence', () => {
     }
   })
 
+  it('strips leadStopWithLiveSubagents on hydrate even if the file already carries it', async () => {
+    // Why: the write-side strip only covers bytes this build wrote. A file from an older
+    // build, a restored backup, or a hand edit would otherwise rehydrate the flag on a
+    // `done` entry — the one state it survives normalization — and re-suppress a real
+    // completion after restart. Enforce the invariant against arbitrary bytes (#4375).
+    mkdirSync(join(userDataPath, 'agent-hooks'), { recursive: true })
+    writeFileSync(
+      lastStatusPath(),
+      JSON.stringify({
+        version: 2,
+        entries: {
+          [PANE]: {
+            paneKey: PANE,
+            tabId: 'tab-1',
+            worktreeId: 'wt-1',
+            receivedAt: recentTs(),
+            stateStartedAt: recentTs(-1000),
+            payload: {
+              state: 'done',
+              prompt: 'injected',
+              agentType: 'codex',
+              leadStopWithLiveSubagents: true
+            }
+          }
+        }
+      }),
+      'utf8'
+    )
+
+    const server = new AgentHookServer()
+    await server.start({ env: 'production', userDataPath })
+    try {
+      const hydrated = server.getStatusSnapshot()[0]
+      // The entry must hydrate (otherwise this asserts nothing about the flag).
+      expect(hydrated?.state).toBe('done')
+      expect(hydrated).not.toHaveProperty('leadStopWithLiveSubagents')
+    } finally {
+      server.stop()
+    }
+  })
+
   it('treats a corrupt file as empty hydration without throwing', async () => {
     mkdirSync(join(userDataPath, 'agent-hooks'), { recursive: true })
     writeFileSync(lastStatusPath(), 'not-json{{', 'utf8')
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -6597,6 +6597,32 @@ describe('Last-status persistence', () => {
     }
   })
 
+  it('never persists leadStopWithLiveSubagents, so a restart cannot re-suppress a completion', async () => {
+    const server = new AgentHookServer()
+    await server.start({ env: 'production', userDataPath })
+    try {
+      await postHookEvent(server, buildBody({ hook_event_name: 'SessionStart' }), '/hook/codex')
+      await postHookEvent(
+        server,
+        buildBody({
+          hook_event_name: 'SubagentStart',
+          agent_id: '11111111-2222-4333-8444-555555555555',
+          agent_type: 'explore'
+        }),
+        '/hook/codex'
+      )
+      await postHookEvent(server, buildBody({ hook_event_name: 'Stop' }), '/hook/codex')
+      server.flushStatusPersistSync()
+      const persisted = JSON.parse(readFileSync(lastStatusPath(), 'utf8'))
+      // Why: the flag describes one live hook delivery. Rehydrating it would make the
+      // notification gate suppress the first real completion after every restart (#4375).
+      expect(persisted.entries[PANE].payload.state).toBe('done')
+      expect(persisted.entries[PANE].payload).not.toHaveProperty('leadStopWithLiveSubagents')
+    } finally {
+      server.stop()
+    }
+  })
+
   it('treats a corrupt file as empty hydration without throwing', async () => {
     mkdirSync(join(userDataPath, 'agent-hooks'), { recursive: true })
     writeFileSync(lastStatusPath(), 'not-json{{', 'utf8')
--- a/src/main/agent-hooks/server.ts
+++ b/src/main/agent-hooks/server.ts
@@ -234,6 +234,12 @@ function sanitizeHydratedEntry(
   if (!payload) {
     return null
   }
+  // Why: the write side already strips this, but the state gate in normalizeAgentStatusPayload
+  // preserves it on `done` — the only state it is ever set on. Enforce here too so the invariant
+  // holds against any bytes on disk, not just the ones this process wrote (#4375).
+  if (payload.leadStopWithLiveSubagents) {
+    delete payload.leadStopWithLiveSubagents
+  }
   const providerSession = normalizeAgentProviderSession(record.providerSession) ?? undefined
   const providerSessionOnly = record.providerSessionOnly === true
   if (providerSessionOnly && !isValidPiProviderSessionOnly(providerSession, payload.agentType)) {
--- a/src/main/agent-hooks/server.ts
+++ b/src/main/agent-hooks/server.ts
@@ -1927,6 +1927,12 @@ export class AgentHookServer {
         continue
       }
       const { promptInteractionKey: _promptInteractionKey, ...persistedPayload } = payload
+      // Why: leadStopWithLiveSubagents describes one live hook delivery, not restorable pane
+      // state — a rehydrated one would re-suppress the first completion after restart (#4375).
+      if (persistedPayload.payload?.leadStopWithLiveSubagents) {
+        const { leadStopWithLiveSubagents: _live, ...inner } = persistedPayload.payload
+        persistedPayload.payload = inner
+      }
       entries[paneKey] = persistedPayload as EnrichedAgentHookEventPayload
     }
     const file: LastStatusFile = { version: LAST_STATUS_FILE_VERSION, entries }
--- a/src/renderer/src/components/terminal-pane/agent-completion-coordinator.ts
+++ b/src/renderer/src/components/terminal-pane/agent-completion-coordinator.ts
@@ -811,6 +811,13 @@ export function createAgentCompletionCoordinator(
       return
     }
     if (isCompletionHookState(payload.state)) {
+      if (payload.leadStopWithLiveSubagents) {
+        // Why: a Codex lead Stop raised while its children still run ends no turn (#4375).
+        // It stays `done` for the sidebar, but notifying here is the reported spam. Cancel any
+        // provisional done too — the lead can Stop inside a quiet window opened by an earlier one.
+        clearPendingHookDone()
+        return
+      }
       // Why: the turn is ending, so a debounced attention from an earlier pause must not fire after completion.
       clearPendingCodexAttention()
       if (isRecognizedAgentType(payload.agentType)) {
--- a/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
+++ b/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
@@ -242,6 +242,43 @@ describe('issue #4375 — non-turn-end notification spam', () => {
     expect(restarted?.payload.state).toBe('working')
   })
 
+  it('gates notification without downgrading the reported state of a lead Stop', () => {
+    // Why: an earlier cut of this fix gated by rewriting `state` to 'working', which silently
+    // broke the sidebar/persistence contract that a root Stop retires child rows and reports
+    // 'done' (server.test.ts). The suppression signal must ride alongside `done`, not replace it.
+    const listenerState: HookListenerState = createHookListenerState()
+    const send = (
+      ev: string,
+      p?: Record<string, unknown>
+    ): ReturnType<typeof normalizeHookPayload> =>
+      normalizeHookPayload(
+        listenerState,
+        'codex',
+        {
+          paneKey: PANE_KEY,
+          tabId: 'tab-1',
+          worktreeId: WORKTREE_ID,
+          hook_event_name: ev,
+          payload: { hook_event_name: ev, ...p }
+        },
+        'test-env'
+      )
+
+    send('UserPromptSubmit', { prompt: 'go' })
+    send('SubagentStart', { agent_id: 'child-1', agent_type: 'reviewer' })
+
+    const leadStop = send('Stop')
+    expect(leadStop?.payload.state).toBe('done')
+    expect(leadStop?.payload.subagents).toBeUndefined()
+    expect(leadStop?.payload.leadStopWithLiveSubagents).toBe(true)
+
+    // The child's own Stop is a real turn end: same shape, but no suppression flag.
+    send('PreToolUse', { agent_id: 'child-1', tool_name: 'Bash', tool_input: { command: 'y' } })
+    const childStop = send('SubagentStop', { agent_id: 'child-1' })
+    expect(childStop?.payload.state).toBe('done')
+    expect(childStop?.payload.leadStopWithLiveSubagents).toBeUndefined()
+  })
+
   // KNOWN-OPEN residual, deliberately out of scope for the roster fix. A bare mid-turn
   // Stop (no subagent) that resumes AFTER the 1.5s quiet window still notifies. There is
   // no principled bound to widen the window to, and widening it delays every real Codex
--- a/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
+++ b/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
@@ -7,6 +7,7 @@ import { dispatchTerminalNotification } from './use-notification-dispatch'
 import {
   createHookListenerState,
   normalizeHookPayload,
+  reconcileRemoteCodexState,
   type HookListenerState
 } from '../../../../shared/agent-hook-listener'
 import type { AgentHookSource } from '../../../../shared/agent-hook-relay'
@@ -277,6 +278,33 @@ describe('issue #4375 — non-turn-end notification spam', () => {
     const childStop = send('SubagentStop', { agent_id: 'child-1' })
     expect(childStop?.payload.state).toBe('done')
     expect(childStop?.payload.leadStopWithLiveSubagents).toBeUndefined()
+  })
+
+  it('gates a lead Stop arriving over SSH/relay, which reconciles on a separate path', () => {
+    // Why: remote panes never reach normalizeHookPayload's codex branch — the main process
+    // re-derives their roster in reconcileRemoteCodexState, which does its own reap. Without the
+    // same pre-reap capture there, the fix would cover local Codex only.
+    const listenerState: HookListenerState = createHookListenerState()
+    const remote = (
+      ev: string,
+      agentId: string | undefined,
+      payload: Parameters<typeof reconcileRemoteCodexState>[4],
+      previous?: Parameters<typeof reconcileRemoteCodexState>[5]
+    ): ReturnType<typeof reconcileRemoteCodexState> =>
+      reconcileRemoteCodexState(listenerState, PANE_KEY, ev, agentId, payload, previous)
+
+    const working = { state: 'working', prompt: 'go', agentType: 'codex' } as const
+    remote('UserPromptSubmit', undefined, { ...working })
+    // The relay reports the live child roster alongside the lead's own state.
+    remote('SubagentStart', 'child-1', {
+      ...working,
+      subagents: [{ id: 'child-1', state: 'working', agentType: 'reviewer', startedAt: Date.now() }]
+    })
+
+    const leadStop = remote('Stop', undefined, { state: 'done', prompt: 'go', agentType: 'codex' })
+    expect(leadStop.state).toBe('done')
+    expect(leadStop.subagents).toBeUndefined()
+    expect(leadStop.leadStopWithLiveSubagents).toBe(true)
   })
 
   // KNOWN-OPEN residual, deliberately out of scope for the roster fix. A bare mid-turn
--- a/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
+++ b/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
@@ -307,11 +307,13 @@ describe('issue #4375 — non-turn-end notification spam', () => {
     expect(leadStop.leadStopWithLiveSubagents).toBe(true)
   })
 
-  // KNOWN-OPEN residual, deliberately out of scope for the roster fix. A bare mid-turn
-  // Stop (no subagent) that resumes AFTER the 1.5s quiet window still notifies. There is
-  // no principled bound to widen the window to, and widening it delays every real Codex
-  // completion; the correct fix is retracting the banner via notifications:dismiss on the
-  // resuming 'working' hook. Pinned here so that change flips this test.
+  // KNOWN-OPEN residual, deliberately out of scope for the roster fix, and the reason this PR
+  // does NOT close #4375: a bare mid-turn Stop (no subagent) that resumes AFTER the 1.5s quiet
+  // window still notifies, which is the "notification after tool calls or reasoning" the issue
+  // reports. There is no principled bound to widen the window to, and widening it delays every
+  // real Codex completion; the correct fix is retracting the banner via notifications:dismiss on
+  // the resuming 'working' hook, which needs a notification id plumbed back through
+  // dispatchCompletion (currently `=> void`). Pinned here so that change flips this test.
   it('KNOWN-OPEN: a bare mid-turn Stop resuming after the quiet window still notifies', () => {
     const calls = drive('codex', [
       { ev: 'UserPromptSubmit', p: { prompt: 'run tests' } },
--- a/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
+++ b/src/renderer/src/components/terminal-pane/codex-lead-stop-subagent-gate.test.ts
@@ -1,0 +1,264 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import {
+  createAgentCompletionCoordinator,
+  resetAgentCompletionCoordinatorIdentitiesForTest
+} from './agent-completion-coordinator'
+import { dispatchTerminalNotification } from './use-notification-dispatch'
+import {
+  createHookListenerState,
+  normalizeHookPayload,
+  type HookListenerState
+} from '../../../../shared/agent-hook-listener'
+import type { AgentHookSource } from '../../../../shared/agent-hook-relay'
+import type { AgentStatusEntry } from '../../../../shared/agent-status-types'
+import type { AgentCompletionStatusSnapshot } from './agent-completion-coordinator-types'
+import type { NotificationDispatchRequest } from '../../../../shared/types'
+
+const LEAF = '11111111-1111-4111-8111-111111111111'
+const PANE_KEY = `tab-1:${LEAF}`
+const WORKTREE_ID = 'wt-primary'
+
+type MockState = Record<string, unknown>
+let mockState: MockState
+
+vi.mock('@/store', () => ({
+  useAppStore: { getState: () => mockState, subscribe: () => () => {} }
+}))
+vi.mock('@/lib/desktop-notification-sound', () => ({ playDesktopNotificationSound: vi.fn() }))
+
+function baseState(): MockState {
+  return {
+    activeWorktreeId: 'wt-other',
+    activeTabId: 'tab-1',
+    tabsByWorktree: { [WORKTREE_ID]: [{ id: 'tab-1', ptyId: 'pty-1' }] },
+    ptyIdsByTabId: { 'tab-1': ['pty-1'] },
+    suppressedPtyExitIds: {},
+    terminalLayoutsByTabId: {
+      'tab-1': {
+        root: { type: 'leaf', leafId: LEAF },
+        activeLeafId: LEAF,
+        expandedLeafId: null,
+        ptyIdsByLeafId: { [LEAF]: 'pty-1' }
+      }
+    },
+    browserTabsByWorktree: {},
+    retainedAgentsByPaneKey: {},
+    agentStatusByPaneKey: {} as Record<string, AgentStatusEntry>,
+    worktreesByRepo: {
+      repo1: [{ id: WORKTREE_ID, repoId: 'repo1', displayName: 'main', branch: 'main' }],
+      repo2: [{ id: 'wt-other', repoId: 'repo2', displayName: 'other', branch: 'other' }]
+    },
+    repos: [
+      { id: 'repo1', displayName: 'danmaku_monitor', connectionId: null },
+      { id: 'repo2', displayName: 'second_repo', connectionId: null }
+    ],
+    settings: {
+      experimentalTerminalAttention: false,
+      notifications: { enabled: true, agentTaskComplete: true }
+    },
+    markWorktreeUnread: vi.fn(),
+    markTerminalTabUnread: vi.fn(),
+    markTerminalPaneUnread: vi.fn(),
+    markAgentCompletionPaneUnread: vi.fn()
+  }
+}
+
+function applyStatusToStore(payload: AgentCompletionStatusSnapshot, now: number): void {
+  const store = mockState.agentStatusByPaneKey as Record<string, AgentStatusEntry>
+  const previous = store[PANE_KEY]
+  const stateStartedAt =
+    previous && previous.state === payload.state ? previous.stateStartedAt : now
+  store[PANE_KEY] = {
+    ...previous,
+    state: payload.state,
+    prompt: payload.prompt,
+    agentType: payload.agentType,
+    toolName: payload.toolName,
+    toolInput: payload.toolInput,
+    lastAssistantMessage: payload.lastAssistantMessage,
+    interrupted: payload.interrupted,
+    subagents: payload.subagents,
+    updatedAt: now,
+    stateStartedAt,
+    paneKey: PANE_KEY,
+    terminalTitle: payload.agentType ?? 'agent',
+    stateHistory: []
+  } as AgentStatusEntry
+}
+
+function dispatchCalls(): NotificationDispatchRequest[] {
+  const d = window.api.notifications.dispatch as unknown as ReturnType<typeof vi.fn>
+  return d.mock.calls.map((c) => c[0] as NotificationDispatchRequest)
+}
+
+function makeCoordinator(): ReturnType<typeof createAgentCompletionCoordinator> {
+  return createAgentCompletionCoordinator({
+    paneKey: PANE_KEY,
+    getPtyId: () => 'pty-1',
+    getSettings: () => null,
+    inspectProcess: async () => ({ foregroundProcess: null, hasChildProcesses: false }),
+    dispatchCompletion: (title, meta) => {
+      dispatchTerminalNotification(WORKTREE_ID, {
+        source: 'agent-task-complete',
+        terminalTitle: title,
+        paneKey: PANE_KEY,
+        ...(meta?.agentStatus ? { agentStatusSnapshot: meta.agentStatus } : {})
+      })
+    },
+    dispatchAttention: (title, meta) => {
+      dispatchTerminalNotification(WORKTREE_ID, {
+        source: 'agent-task-complete',
+        terminalTitle: title,
+        paneKey: PANE_KEY,
+        agentStatusSnapshot: meta.agentStatus
+      })
+    },
+    isLive: () => true
+  })
+}
+
+type Step = { ev: string; p?: Record<string, unknown>; waitMs?: number }
+
+/** Drives real hook normalization -> store mirror -> coordinator -> dispatcher. */
+function drive(source: AgentHookSource, steps: Step[]): NotificationDispatchRequest[] {
+  const listenerState: HookListenerState = createHookListenerState()
+  const coordinator = makeCoordinator()
+  for (const step of steps) {
+    const event = normalizeHookPayload(
+      listenerState,
+      source,
+      {
+        paneKey: PANE_KEY,
+        tabId: 'tab-1',
+        worktreeId: WORKTREE_ID,
+        hook_event_name: step.ev,
+        payload: { hook_event_name: step.ev, ...step.p }
+      },
+      'test-env'
+    )
+    if (event) {
+      const now = Date.now()
+      applyStatusToStore(event.payload, now)
+      const stored = (mockState.agentStatusByPaneKey as Record<string, AgentStatusEntry>)[PANE_KEY]
+      coordinator.observeHookStatus({ ...event.payload, stateStartedAt: stored.stateStartedAt })
+    }
+    vi.advanceTimersByTime(step.waitMs ?? 100)
+  }
+  vi.advanceTimersByTime(6_000)
+  coordinator.dispose()
+  return dispatchCalls()
+}
+
+const BASH_1 = { tool_name: 'Bash', tool_input: { command: 'npm run test:submit' } }
+const BASH_2 = {
+  tool_name: 'Bash',
+  tool_input: { command: 'node scripts/check-hook-registry.js' }
+}
+
+describe('issue #4375 — non-turn-end notification spam', () => {
+  beforeEach(() => {
+    vi.useFakeTimers()
+    vi.setSystemTime(new Date('2026-07-25T12:00:00Z'))
+    vi.stubGlobal('document', { visibilityState: 'hidden', hasFocus: () => false })
+    vi.stubGlobal('navigator', { userAgent: 'Mozilla/5.0 (Windows NT 10.0; Win64; x64)' })
+    vi.stubGlobal('window', {
+      api: { notifications: { dispatch: vi.fn().mockResolvedValue({ delivered: true }) } }
+    })
+    mockState = baseState()
+    // Why: completion identities are module-scoped and keyed by paneKey, so they leak between tests.
+    resetAgentCompletionCoordinatorIdentitiesForTest()
+  })
+  afterEach(() => {
+    vi.useRealTimers()
+    vi.unstubAllGlobals()
+  })
+
+  it('control: a real turn end notifies once', () => {
+    const calls = drive('codex', [
+      { ev: 'UserPromptSubmit', p: { prompt: 'run tests' } },
+      { ev: 'PreToolUse', p: BASH_1 },
+      { ev: 'PostToolUse', p: BASH_1 },
+      { ev: 'Stop' }
+    ])
+    expect(calls).toHaveLength(1)
+    expect(calls[0].agentState).toBe('done')
+  })
+
+  it('REPRO: a Codex lead Stop notifies while a subagent is still working', () => {
+    const calls = drive('codex', [
+      { ev: 'UserPromptSubmit', p: { prompt: 'run tests' } },
+      { ev: 'SubagentStart', p: { agent_id: 'child-1', agent_type: 'reviewer' } },
+      { ev: 'PreToolUse', p: BASH_1 },
+      // Lead finishes; the child keeps working well past the 1.5s quiet window.
+      { ev: 'Stop', waitMs: 5_000 },
+      { ev: 'PreToolUse', p: { ...BASH_2, agent_id: 'child-1' } }
+    ])
+
+    if (calls.length > 0) {
+      throw new Error(
+        `Spurious completion while subagent child-1 is still working: ` +
+          `state=${calls[0].agentState} tool=${calls[0].agentToolInput}`
+      )
+    }
+    expect(calls).toHaveLength(0)
+  })
+
+  it('cancels a mid-turn Stop when work resumes inside the quiet window', () => {
+    const calls = drive('codex', [
+      { ev: 'UserPromptSubmit', p: { prompt: 'run tests' } },
+      { ev: 'PreToolUse', p: BASH_1 },
+      // Resumes at 1s — inside HOOK_DONE_QUIET_MS, so the provisional done is cancelled.
+      { ev: 'Stop', waitMs: 1_000 },
+      { ev: 'PreToolUse', p: BASH_2 }
+    ])
+    expect(calls).toHaveLength(0)
+  })
+
+  it('a new session is not gated by the roster the previous Codex process left behind', () => {
+    const listenerState: HookListenerState = createHookListenerState()
+    const send = (
+      ev: string,
+      p?: Record<string, unknown>
+    ): ReturnType<typeof normalizeHookPayload> =>
+      normalizeHookPayload(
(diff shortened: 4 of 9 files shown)


nwparker:
## Round 2 update — both reviewer MAJORs closed

Pushed `84880f2711` + `5fd00b9f57`. The previous approach was wrong and a reviewer was right to block it.

### 1. Regression in two pre-existing tests — fixed

Verified the claim before acting on it. `origin/main`: **243/243 pass**. Commit `258b4ae476`: **2 failed**. Mine, and real.

Root cause: I gated by rewriting a lead `Stop`'s `state` from `done` → `working`, colliding with the opposite contract in `server.ts` (a root `Stop` retires child rows and reports `done`).

The deeper problem is that the roster reap *destroys* the signal. Instrumenting the payload stream on `origin/main`:

```
UserPromptSubmit:    state=working  subs=undefined
SubagentStart:       state=working  subs=["child-1:working"]
PreToolUse(lead):    state=working  subs=["child-1:working"]
Stop(lead):          state=done     subs=undefined    <- must NOT notify
PreToolUse(child):   state=working  subs=["child-1:working"]
SubagentStop(child): state=done     subs=undefined    <- MUST notify
```

The two `done` events are **byte-identical**, and both follow an identical predecessor. No downstream heuristic can separate them — the information is already gone. So the boolean is captured *before* the reap and carried as a live-only field. `state` is untouched; only the notification layer reads it.

### 2. SSH/relay bypass — fixed

Confirmed: remote panes reconcile through `reconcileRemoteCodexState`, which has its own reap and never set the flag. Second commit applies the same pre-reap capture there, with its own test. Without it this would have been a local-only fix.

### Current state

```
✓ control: a real turn end notifies once
✓ REPRO: a Codex lead Stop notifies while a subagent is still working
✓ cancels a mid-turn Stop when work resumes inside the quiet window
✓ a new session is not gated by the roster the previous Codex process left behind
✓ gates notification without downgrading the reported state of a lead Stop
✓ gates a lead Stop arriving over SSH/relay, which reconciles on a separate path
✓ KNOWN-OPEN: a bare mid-turn Stop resuming after the quiet window still notifies
```

`server.test.ts` + gate suite: **250 passed**. Affected areas: **7275 passed**. Full suite: **36428 passed** (the 8 IME/editor/packaging failures reproduce identically on clean `origin/main` in this worktree — stale `node_modules`; verified by reverting all four source files).

Every mechanism is mutation-verified and each kills a *different* test — no redundancy, no unguarded mechanism.

Still open and labeled in the body: the bare mid-turn `Stop` with no subagents (`KNOWN-OPEN` test). Leaving this in **draft** until review confirms.

nwparker:
## Regression report: confirmed, root-caused, and fixed in `a6a5076`

Thanks for catching this — the report was right, and my first check was too narrow.

### What I got wrong initially

I probed only the **lead `Stop`** path and found that `origin/main` *also* reports `state: "done"` there, so I concluded there was no regression. That was the wrong event to test. Re-probing a **non-`Stop`** lead event surfaced the real divergence:

| lead event, with a child blocked on a permission prompt | `origin/main` | this branch (before `a6a5076`) |
| --- | --- | --- |
| `PostToolUse` | `waiting` | **`working`** ← regression |
| `Stop` | `done` | `done` + `leadStopWithLiveSubagents: true` |

### Root cause

`84880f2711` reverted the state rewrite because it broke the opposite contract in `src/main/agent-hooks/server.ts` (a root `Stop` retires child rows and must report `done`). But the revert removed the `codexRosterEffectiveState` call from the local lead path **entirely** — not just from the `Stop` case. Collateral damage: a child blocked on a permission prompt stopped holding the pane at `waiting`, so the pane read `working` and the user never saw what it was blocked on.

### Fix

Restore the read, but place it **after** the roster reap instead of before it:

```ts
// Why: read the roster AFTER the reap. A blocked child must still hold the pane at `waiting` on
// non-Stop lead events, but on Stop the roster is already empty so this stays `done` — the root
// Stop contract in agent-hooks/server.ts depends on that.
const effectiveState = codexRosterEffectiveState(
  state.codexSubagentRosterByPaneKey.get(paneKey),
  stateName
)
```

Post-reap, a lead `Stop` sees an empty roster and still yields `done`; only non-`Stop` events observe live children. That is exactly what `origin/main` does — the branch now matches main on both paths. The pre-reap `leadStopWithLiveSubagents` flag is untouched and still carries the "children were live at Stop" signal to the notification gate.

### Proof

New `src/shared/codex-lead-state-vs-subagent-roster.test.ts` pins **both** orderings, so neither can be reintroduced:

| mutation applied to the fix | result |
| --- | --- |
| drop the `effectiveState` read (i.e. the reported regression) | ❌ `holds the pane at waiting on a non-Stop lead event while a child is blocked` fails |
| move the read *before* the reap (i.e. the ordering `84880f2711` reverted) | ❌ `still reports done on a lead Stop, flagging that children were live` fails |
| unmutated | ✅ 3/3 pass |

Each mutation kills exactly the contract it violates and nothing else — neither test is redundant, and the ordering is the load-bearing detail.

### No regressions

`src/shared/` + `src/main/agent-hooks/` + the lead-Stop gate suite: **3531 passed | 9 skipped, 0 failed** (366 files). That run includes the two `server.ts` root-`Stop` tests whose failure forced the original revert — both green.

Still draft, not merging.

nwparker:
## Residual risk, now measured rather than assumed

I previously flagged an unquantified worry: if Codex CLI omits a child `SubagentStop` (it does — 0.144 is known to), a stale roster entry could make `leadStopWithLiveSubagents` fire on a turn that genuinely ended, suppressing a real notification.

Probed it. The blast radius is **exactly one notification**, self-healing:

```
PROBE turn1 flag = true        <- suppressed (the false positive)
PROBE turn2 flag = undefined   <- notifies normally
PROBE roster size after = 0
```

The `Stop` handler deletes the roster, so the stale entry cannot survive into the next turn. Worst case is one missed completion banner on the turn where Codex dropped the child's Stop hook; the next turn is unaffected. That is strictly better than the reported bug it replaces (a banner on *every* mid-turn lead Stop, indefinitely), and it self-corrects without user action.

Stating it rather than burying it — no roster expiry exists today, so this bound depends on the `Stop` reap staying in place. The test `omits the flag on a lead Stop with no live children` covers the ordinary path; the one-turn bound above is a property of the reap, which the `still reports done on a lead Stop` test pins.

nwparker:
## Perf review

`a6a5076` adds one `codexRosterEffectiveState` call per local lead hook event. Cost analysis:

- The function is an early-exit scan over the pane's roster: returns immediately if the roster is empty or undefined, and returns on the **first** `waiting` child.
- The roster is **hard-capped at `AGENT_STATUS_MAX_SUBAGENTS = 32`** (`agent-status-types.ts:248`), enforced in `upsertCodexSubagent` (`codex-subagent-roster.ts:43`). So worst case is a 32-iteration loop over an in-memory `Map`, with no allocation, I/O, or serialization.
- It runs on the hook-event path, which is already doing `normalizeAgentStatusPayload` plus `codexRosterToSnapshots` (which allocates an array **and sorts it**) on the same event. The added scan is strictly cheaper than work already on that path.
- No new timers, subscriptions, listeners, polling, or hot-path filesystem access. No change to event frequency — this is per-event work, not a new event source.

**Verdict: no measurable perf impact.** This restores a call that existed on this path before `84880f2711` and is unchanged in the remote path; it is not new work in aggregate, it's work that was accidentally removed.

## Cross-platform

No platform-dependent code: pure in-memory state over hook payloads, no paths, no separators, no shell, no keyboard handling. The local and remote/SSH transports both compute the same flag, so SSH panes and native panes agree — that parity was the point of `5fd00b9f57` and is preserved here.

nwparker:
## Round-2 review response

Both MAJORs addressed in `c1b5c3ff5c` + a PR-body rewrite. Taking them in reverse order, because the second one changes what this PR claims.

### MAJOR 2 — scope honesty: you were right, and the auto-close is gone

I re-read #4375 rather than defending the framing. The report is *"after tool calls or reasoning, I will get a notification"* — which is exactly what the `KNOWN-OPEN` test pins as **still reproducing**. Claiming `Closes #4375` while shipping a green test that asserts the reported symptom persists is indefensible.

Changed:
- **Removed the auto-close keyword entirely.** Merging this will no longer close #4375. The link now reads "partially addresses".
- Added a **Scope** section to the PR body with a table naming which of the two spam paths is fixed (lead `Stop` with a live subagent) and which is not (bare mid-turn `Stop` past the quiet window).
- Expanded the `KNOWN-OPEN` comment to say why it isn't fixed here: retraction needs a notification id plumbed back through `dispatchCompletion`, typed `=> void` today (`agent-completion-coordinator-types.ts:30`). Separate change, separate blast radius.

### MAJOR 1 — I measured the bound instead of adding the timer, and I'd like you to push back if you disagree

You asked for "fresh liveness evidence or a bounded/provisional suppression that eventually releases." I probed the real payload stream first, because the remedy depends on how bad it actually is. Three scenarios, driven end-to-end through normalization → store → coordinator → dispatcher:

| scenario | notifications delivered |
| --- | --- |
| child finishes normally after a gated lead `Stop` | **1** — the child's own `SubagentStop` carries the turn end. Deferred, not lost. |
| child `Stop` hook dropped by the CLI | **0** for that turn — one real completion lost |
| dropped child `Stop`, then a second turn | **1** — self-heals, no user action |

So it is **already bounded** — by the `Stop` reap, not by a timer. The framing of "permanently / no later event can recover it" doesn't hold: `leadStopWithLiveSubagents` is set for exactly one turn. Both rows are now pinned as tests, so this claim is checkable rather than a comment.

**Why I did not add the aging timer.** A Codex subagent can legitimately run for many minutes. Any release window short enough to salvage the dropped-hook turn is necessarily shorter than a real long-running child — so it would fire while the child is still working and re-emit the exact banner this gate exists to suppress. That trades a *rare, self-healing* missed notification for a *common, permanent* false one, which I think is the wrong direction.

`startedAt` is on the roster, so the timer is implementable if you want it — I'm declining it on the merits, not on effort. If you think the dropped-hook case is more common than I'm assuming, say so and I'll add it.

### MINOR — folded in

The blocked-child contract now runs parameterized over `PreToolUse`, `PostToolUse`, and `UserPromptSubmit`, since every lead event normalizing to `working` must yield to a blocked child.

### Verification

`src/shared/` + `terminal-pane/` + `hooks/`: **5950 passed / 6 skipped**. `tsc -p config/tsconfig.node.json` and `oxlint` clean.

One caveat I'd rather state than have you find: 5 tests fail in `terminal-ime-xterm-composition-deduplication.test.ts`. They are **pre-existing on main** — I checked out the untouched merge base `c67aadbc18` in a scratch worktree and got the identical 5 failures. This branch changes 6 files, none of them IME-related.

nwparker:
### Response — both findings target `a6a5076`, superseded by `c1b5c3f` and `e41f9c0`

Thanks — this review landed against `a6a5076`, which is two commits behind the branch head. Both commits since exist specifically to answer these two points. Verified at HEAD (`e41f9c0`): gate + roster suites **14/14 green**.

**Finding 1 — dropped child `SubagentStop` costs a real completion.** Correct, and accepted as a real cost rather than argued away. `c1b5c3f` bounds it by measurement instead of redesigning: the `Stop` reap empties the roster, so the cost is **exactly one turn** — the next turn notifies unaided. On the normal path the child's own `SubagentStop` carries the turn end the gated lead `Stop` withheld, so it's *deferred, not lost*. Two tests pin both halves:

- `flags only the turn a dropped child Stop spans, not the next one`
- `clears the flag as soon as the child does report, so nothing is stranded`

On the suggested remedy — a bounded/aging release window — that measured bound is exactly why one isn't added. A subagent can legitimately run for minutes, so any window short enough to salvage the dropped-hook turn is shorter than a real child, and would re-fire the spam this gate exists to stop. Trading a one-turn deferral for recurring false completions is the wrong side of that trade.

**Finding 2 — `KNOWN-OPEN` asserts wrong behavior; scope-blocking if the PR claims to fix #4375.** The PR does not claim that. Line 1 of the body: *"Partially addresses #4375 — **deliberately does not close it.**"* with a dedicated `## Scope: what this does not fix` section, and no `Fixes/Closes/Resolves` keyword anywhere. `c1b5c3f` also extended the test comment to name the missing piece: retraction needs a notification id plumbed back through `dispatchCompletion` (currently `=> void`).

So the disposition you asked for is already the one in place — scoped away from a full #4375 close, with the residual pinned by a named, green `KNOWN-OPEN` test that a real fix flips.

nwparker:
## Round 2 — the MINOR was right, and chasing it turned up a MAJOR in my own fix

Round 2 returned MINOR ISSUES. Investigating the MINOR led me to a much more serious defect: **the gate never fired in the real app.** Details below, both fixed.

---

### 🔴 MAJOR — the flag was dropped crossing IPC, so the fix was dead in production — **FOUND WHILE VERIFYING THE MINOR, fixed in `c5ee1d4759`**

`applyAgentStatus` in `useIpcEvents.ts` rebuilds the hook payload **field by field** before handing it to the completion coordinator. `leadStopWithLiveSubagents` was not in that whitelist, so it arrived as `undefined` and the gate at `agent-completion-coordinator.ts:814` was never entered for local Codex panes — which is the exact path this PR targets.

The file already warns about this trap two lines above the omission:

```ts
// Why: same trap as interactivePrompt — this rebuild is a field whitelist, so subagent child rows vanish if omitted.
subagents: data.subagents
```

I added a field to the payload type and didn't update the whitelist.

**Why every test stayed green:** `codex-lead-stop-subagent-gate.test.ts` drives `coordinator.observeHookStatus(event.payload)` **directly**, bypassing the IPC hop where the flag dies. It exercised the coordinator's logic, which was correct, and never the delivery path, which was broken. This is precisely the "green test that proves nothing about production" failure mode — I'd flagged that risk elsewhere in this work and then shipped an instance of it.

**Empirical proof.** I drove the real `agentStatus.onSet` listener and dumped what actually reached the coordinator:

```
PROBE forwarded payload = {
  "state": "done",
  "agentType": "codex",
  "prompt": "lead stop while children run",
  "leadStopWithLiveSubagents": undefined,   <-- gate input, gone
  ...
}
```

Fix is one line in the whitelist. The new test in `useIpcEvents.test.ts` goes through the real listener and asserts the flag reaches `observeAgentHookCompletionForNotification`; deleting the whitelist line fails it.

### ✅ MINOR — non-persistence enforced only on the write side — **ACCEPTED, fixed in `3d8392f3a3`**

Correct, and my earlier claim that the `state === 'done'` coercion in `normalizeAgentStatusObject` was a second protective layer **was wrong** — I had it backwards. That gate preserves the flag on `done`, which is the only state it's ever set on. So it passes exactly the entry that matters, and the write-side strip was the sole enforcement.

`sanitizeHydratedEntry` now strips it unconditionally, so the invariant holds against arbitrary bytes on disk rather than resting on "nothing else ever writes this file."

One implementation note worth recording: **`delete`, not `= undefined`.** `toAgentStatusIpcPayload` spreads `...entry.payload`, and an explicit `undefined` key still surfaces as a *present property* to consumers. My first attempt assigned `undefined` and the new test still failed:

```
AssertionError: expected { …(17) } to not have property "leadStopWithLiveSubagents"
```

The new test injects a hand-written `last-status.json` already containing the flag — covering the file-from-elsewhere case the review described, which the write-side test structurally cannot reach.

### ⚪ NIT — `clearPendingHookDone()` at coordinator:818 has no test forcing it

Not taken, and I agree with the reviewer's own reasoning: every non-`done` event already clears through the `working` or `waiting`/`blocked` branch, so there's no reachable interleaving where that line is the one that matters. It's defensive. I'd rather leave it than add a test that pins unreachable behavior — the round found two real defects, and padding coverage here would obscure that.

---

### Verification

- New tests **mutation-tested individually** — each mutation kills exactly its own test:
  - remove the IPC whitelist line → `useIpcEvents` test fails
  - remove the hydrate strip → `server.test.ts` hydrate test fails
- `server.test.ts` + `useIpcEvents.test.ts` + both gate suites — **341 passed**
- Full `src/main/agent-hooks/` + `src/renderer/src/hooks/` — **920 passed | 3 skipped**
- `tsc --noEmit` (node **and** web projects) — clean. The one web error (`@sanity/diff-match-patch` unresolved) reproduces identically on a stashed tree, so it's pre-existing and not from this change.
- `oxlint`, `oxfmt` — clean
- `src/renderer/src/components/terminal-pane/` — 2433 passed, **5 pre-existing failures** in `terminal-ime-xterm-composition-deduplication.test.ts`. Verified identical on a stashed tree: unrelated to this PR.

### Perf / cross-platform

No regression. The IPC change is one property copy per status event on a path that already copies eleven. The hydrate change is one property check per entry at startup only. Both are pure shared/main-process state logic with no OS-specific branches.

### Loop status

Rounds: **2**. Round 2 verdict was MINOR, but the follow-up investigation found a MAJOR, so this does **not** count as converged — the round-3 pass should target the delivery path end-to-end rather than the coordinator's logic in isolation.

Still **draft** — not for merge.

nwparker:
## ⚠️ Second delivery route found: the title path can notify around the gate

Following up my own round-2 comment. While auditing every route the flag travels, I found a **second, independent notification source** that this PR does not gate — and it can fire for the same lead `Stop` the gate suppresses.

### What's there

There are three `source: 'agent-task-complete'` dispatch sites, not one:

| Site | Gated by this PR? |
|---|---|
| `agent-hook-completion-notifications.ts:246` (hook path) | ✅ yes |
| `pty-connection.ts:3040` (live title path) | ❌ no |
| `parked-terminal-byte-watcher.ts:167` (parked-tab title path) | ❌ no |

The title paths run off the working→idle **terminal title** transition, not off hook events, so they never consult `leadStopWithLiveSubagents`. The parked watcher has no hook awareness at all.

There *is* an interlock — `shouldSuppressTitleCompletionForFreshHook` (`pty-connection.ts:1413`) — but it does not cover this case, and the reason is structural:

```ts
if (detectAgentStatusFromTitle(title) === 'working' || !isFreshNonDoneAgentStatus(activeHookStatus)) {
  return false   // ← no suppression
}
```

A gated lead `Stop` deliberately stores **`state: 'done'`** (that's the "gates notification without downgrading the reported state" behavior this PR is built around, and it's correct for the sidebar). So `isFreshNonDoneAgentStatus` returns `false`, the interlock returns `false`, and the title completion is allowed through. **The very design decision that keeps the sidebar honest is what opens the title path.**

Note both coordinator instances (`pty-connection.ts:2252` and `agent-hook-completion-notifications.ts:232`) are separate objects. They share `lastCompletionIdentityByPaneKey`, but that map is only written when a completion actually *dispatches* — the gate returns early and writes nothing, so there's no cross-instance record of "this turn was suppressed."

### Proof

Driving real hook normalization → store → coordinator, then delivering an idle title, **routed through a mirror of the production guard** rather than calling the coordinator directly:

```
PROBE_notification_count: 1
```

One notification escapes. Reverting to the ungated path gives the same result, so the guard is genuinely not blocking it.

### What I have NOT proven — the part that decides severity

**Whether the Codex CLI actually emits an idle title in this window.** My probe supplies the title; it doesn't demonstrate Codex produces one while its subagents are still running.

- If Codex keeps a **working** title while children run, `detectAgentStatusFromTitle(title) === 'working'` short-circuits and **no notification escapes** — this is theoretical and the PR delivers its full user-visible benefit.
- If Codex publishes an **idle** title alongside the mid-turn lead `Stop`, then the user still gets the spurious notification and **this PR does not fix the reported symptom**, even though the hook-side logic is correct.

I can't settle that from the repo — it depends on the CLI's real title behavior, which needs observation against a live Codex session with subagents. Flagging it rather than guessing, because the answer determines whether this is a footnote or the main event.

### Why I'm not patching it in this PR

The obvious fix — teach the title suppression about a gated lead Stop — means threading turn-level suppression state across two coordinator instances and three dispatch sites. That's a design change with its own regression surface, and this PR is already a rescoped partial fix for #4375. Sneaking it in would repeat the scope-honesty problem review already caught here once. I'd rather land the hook-side gate as an honest partial and treat the title route as a separate, properly-reviewed change.

Recommend this route be confirmed against a live Codex session before this merges — if the title fires, the PR's user-visible claim needs to be narrowed again.

Still **draft** — not for merge.

nwparker:
## The open question is now answered — from Codex source + execution

The PR body left one thing unproven: *"whether the second (terminal-title) notification route actually fires depends on whether the Codex CLI emits an idle title mid-turn — unproven."*

I read the Codex source at `~/projects/codex` and then ran both halves against Orca's real coordinator. **It does fire, conditionally, and the gate as written does not stop the banner — it relocates it.**

### 1. Codex does emit an idle title mid-turn — via a thread switch

`chatwidget/status_surfaces.rs`:

```rust
fn terminal_title_has_active_progress(&self) -> bool {
    if self.terminal_title_shows_action_required() { return false; }
    self.mcp_startup_status.is_some() || self.bottom_pane.is_task_running()
}
```

The mid-turn path is a **thread switch**, not a timer. `thread_routing.rs:1267` → `restore_thread_input_state` (`input_restore.rs:284-372`) → `restore_running(...)` → `update_task_running_state()` → `refresh_status_surfaces()` → `set_terminal_title()`. Switching to a thread whose turn is idle re-writes the title while the *lead* turn is still live.

Codex's blocking-Stop path is the matching half — `core/src/session/turn.rs:519-575`:

```rust
if stop_outcome.should_block {
    ...
    stop_hook_active = true;
    continue;          // Stop hook fired, turn did NOT end
```

### 2. Whether Orca reacts depends on the user's Codex config

Run against Orca's real `createAgentStatusTracker` (not a mock):

| Codex `terminal_title_items` | title transition | tracker emits |
|---|---|---|
| **default** `["activity","project-name"]` | `⠹ orca` → `orca` | `['working']` — **no idle**, second route dormant |
| `["activity","app-name","project-name"]` | `⠹ codex \| orca` → `codex \| orca` | `['working','idle:codex \| orca']` — **false idle mid-turn** |
| same, repeated switches | — | `['working','idle:…','working','idle:…']` — **re-arms every switch** |

Reason: `agent-title-status.ts:66-84` only calls `onBecameIdle` when the new status is **non-null**, and `agent-title-owner.ts:17` requires an agent-name token. `Ready | orca` alone does not fire; `Ready | codex | orca` does.

So under stock Codex config the second route is dormant. Under a configured `app-name` it fires, and it fires repeatedly.

### 3. The gate leaves the title route *un-deduped* — measured, not argued

This is the part I'd want fixed before merge. The gate returns early here:

```ts
if (isCompletionHookState(payload.state)) {
  if (payload.leadStopWithLiveSubagents) {
    clearPendingHookDone()
    return          // <-- returns BEFORE dispatchCompletion()
  }
```

`dispatchCompletion` is what sets `lastCompletedTurn = currentTurn`, and that assignment is exactly what *deduped* a follow-up title completion. Returning early skips it, so the title route is left armed.

A/B/C/D against the real coordinator (arm A reproduces post-gate state exactly: the gate returns after `recordPaneActivity()`/`establishAgentEvidence()`/`clearPendingHookDone()`, so it is state-identical to having delivered only the preceding `working` hook):

| arm | scenario | banners |
|---|---|---|
| A | **gated** lead Stop → 1 mid-turn idle title | **1** (via title route) |
| B | **ungated** lead Stop → same idle title | **1** (via hook route; title deduped) |
| C | **gated** + 3 thread switches | **3** |
| D | **ungated** + 3 thread switches | **4** |

Net effect in the single-switch case: **unchanged banner count**, different source. The user still gets a banner mid-turn.

There's a second, related edge. `shouldSuppressTitleCompletionForFreshHook` (`pty-connection.ts:1413`) gates on `isFreshNonDoneAgentStatus`, which returns **false** for `state === 'done'`:

```ts
return Boolean(entry && entry.state !== 'done' && now - entry.updatedAt <= staleAfterMs)
```

The PR deliberately keeps `state: 'done'` ("*`state` stays `done` so sidebar/persistence behavior is unchanged*"). That's a defensible call for the sidebar, but it also means the gated Stop **switches title-route suppression off** — the one damper that would otherwise catch this. Measured: `title route suppressed after gated lead Stop? false`.

### Why the tests stayed green

`codex-lead-stop-subagent-gate.test.ts` feeds hooks → coordinator directly and never supplies a terminal title, so it is structurally blind to the title route. Same shape as the whitelist bug this PR already hit once: green tests, no coverage of the boundary that matters. Worth noting the whitelist itself is **fine now** — `useIpcEvents.ts:2998-3001` does carry `leadStopWithLiveSubagents`. (`server.ts:590`/`:644` also carry `subagents` but synthesize *new* payloads for inferred interrupts; omitting the flag there is correct, not a gap.)

### Suggested direction

Either have the gate still consume the turn for dedup purposes (advance `currentTurn`/`lastCompletedTurn` without dispatching), or extend the suppression predicate to treat a `leadStopWithLiveSubagents` `done` as "turn still active" for the title route while leaving `state` alone for the sidebar. The first is smaller; the second is closer to what the flag actually means.

Not blocking anything myself — this stays draft. Flagging it because "the gate works" and "the user stops getting a mid-turn banner" turn out to be different claims.

<sub>Verification: Codex source read at `a559b65280`; all behavioral claims above executed against Orca's real `createAgentStatusTracker`/`createAgentCompletionCoordinator`, not mocks. Scratch tests were removed after measurement.</sub>

nwparker:
## Correction to my previous comment: verify against the shipped binary, not the clone

Amending my comment above. The Codex checkout at `~/projects/codex` is **`a559b65280`, dated 2026-05-15** — roughly two months stale. Anyone reading source from it will reach at least one wrong conclusion about this PR, so recording the delta.

I re-verified against the **actually-installed CLI** (`codex-cli 0.145.0`, native binary at `@openai/codex-darwin-arm64/vendor/aarch64-apple-darwin/bin/codex`):

**The clone says subagent hooks don't exist. The shipped CLI has them.**

| | clone `a559b65280` | shipped 0.145.0 |
|---|---|---|
| `HookEventsToml` accepted keys | 8 — no subagent events | **11** — `… UserPromptSubmit SubagentStart SubagentStop Stop` (+`SessionEnd`) |
| `HookEventName` variants | 8 | includes `subagent_start`, `subagent_stop` |
| subagent hook I/O schemas | absent | `subagent-start.command.input`/`.output`, `subagent-stop.command.input`/`.output` |

Shipped `SubagentStart` input: `agent_id, agent_type, cwd, hook_event_name, model, permission_mode, session_id, transcript_path, turn_id`. `SubagentStop` adds `agent_transcript_path`, `last_assistant_message`, `stop_hook_active`.

`agent_id` and `agent_type` are exactly the fields `normalizeCodexSubagentLifecycleEvent` reads (`agent-hook-listener.ts:3219-3237`). **The roster does populate on real Codex**, and the events `hook-service.ts:92-93` installs are supported ones. If you were about to conclude from the clone that the discriminator can never fire in production and the gate tests fabricate impossible events — it doesn't and they don't.

Worth adding for the compat question: shipped `HookEventsToml` has **no `deny_unknown_fields`**, so on an older Codex that lacks these events the config still loads and the unknown keys are silently dropped. Degrades, doesn't break.

**None of this changes my finding above.** The banner-relocation measurement was run against Orca's real coordinator, not against Codex source, so it stands unaffected:

| arm | banners |
|---|---|
| gated lead Stop → 1 mid-turn idle title | 1 (title route) |
| ungated → same | 1 (hook route) |
| gated + 3 thread switches | 3 |
| ungated + 3 thread switches | 4 |

One further thing that *is* worth folding in, and which the clone reports correctly (`turn.rs:369`, `:529-538`, `:563-564`; executable test `core/tests/suite/hooks.rs:958-965` observes `[false, true, true]` for three same-turn Stops): **`stop_hook_active` is not a usable discriminator for the current Stop.** `true` means a *prior* Stop in this turn blocked — so gating on it would miss the first blocking Stop and could suppress the eventual real final Stop. It's evidence of same-turn continuation after the fact, not a synchronous "this Stop ends nothing" boolean. Mentioning it because it's the obvious alternative discriminator and it doesn't work.

<sub>Verified by extracting embedded schemas and serde field lists from the installed 0.145.0 binary; version boundary is 0.145.0-vs-2026-05-15-main, not something I can pin to an exact release.</sub>

nwparker:
## Widening my own claim: the second route fires on **stock** Codex config too

I said above that under Codex's default `terminal_title_items` (`["activity","project-name"]`) the title route is dormant, and that you need `app-name` configured to wake it. **That was too narrow.** An adversarial pass found a case I missed, and I reproduced it independently.

The default title is `<spinner> <project-name>` / `<project-name>`. If the **project directory is named after an agent**, the idle title *is* an agent-name token, and the route fires with no config change at all:

```
project="orca"           rawStatus(idle title)=null   idleFired=0
project="codex"          rawStatus(idle title)=idle   idleFired=2   <-- fires
project="my-codex-fork"  rawStatus(idle title)=null   idleFired=0
project="codexinator"    rawStatus(idle title)=null   idleFired=0
project="hot-bugs-07-25" rawStatus(idle title)=null   idleFired=0
```

Sweeping directory names, stock config, on Orca's real `createAgentStatusTracker`:

**Fires:** `codex`, `Codex`, `claude`, `gemini`, `cursor`, `aider`, `opencode`
**Does not fire:** `orca`, `codex-cli`, `my-codex-fork`, `codexinator`, `droid`, `amp`, `goose`, `crush`, `ready`, `done`, `idle`

The word-boundary regex (`agent-title-owner.ts:17`) is why `codexinator` and `my-codex-fork` are safe while a bare `codex` is not.

This is not exotic. Anyone with a checkout at `~/projects/codex`, `~/src/claude`, or `~/work/cursor` and stock Codex config hits it — **which is exactly the directory layout of this machine.** Combined with the finding in my first comment (the gate leaves the title route un-deduped), the practical reach of the second route is wider than "only if you configured `app-name`."

Corrected summary of when route 2 fires, all measured:

| Codex config | project dir | mid-turn thread switch → |
|---|---|---|
| default | non-agent name (`orca`) | dormant |
| default | **agent name (`codex`)** | **fires, re-arms every switch** |
| `app-name` added | any | fires, re-arms every switch |

Everything else in my earlier comments stands — this only broadens the exposure. Credit where due: I did not find this one; an adversarial reviewer did, and I then reproduced it from scratch rather than taking the test at face value.

<sub>Measured against `createAgentStatusTracker` and `detectAgentStatusFromTitle` directly. Scratch tests removed after measurement.</sub>

nwparker:
## Correcting myself again: the second route is **not** dormant on default config

Two comments ago I said the title route is dormant under Codex's stock `terminal_title_items`. One comment ago I narrowed that to "unless your project directory is named after an agent." **Both were wrong, for the same reason: I was only looking at one of the two title paths.**

`pty-connection.ts` feeds the coordinator through **two** independent entry points:

1. `onAgentBecameIdle` → `observeClassifiedTitleCompletion` (line 3136) — gated by the tracker, which needs an agent-name token in the title. This is the path I measured, and my earlier claims are correct *about this path*.
2. **`observeTitle(decision.rawTitle)` (line 2654)** — called on **every** title change, gated only by `shouldSuppressTitleCompletionForFreshHook`. **No agent-name token required.**

Route 2 doesn't need the title to name an agent, because `agentIdentityEstablished` / `hasAgentRunEvidence` can be satisfied by **process inspection** instead. Measured, stock Codex config, project named `orca` (not an agent name):

```
RAWTITLE fg=codex   completions=1  ["orca"]     <-- fires
RAWTITLE fg=null    completions=0  []
RAWTITLE fg=zsh     completions=0  []
```

The sequence is exactly Codex's default title output: `orca` → `⠋ orca` → `⠙ orca` → `⠹ orca` → `orca`. With a `codex` foreground process, that plain `orca` idle title **completes the turn**.

So the corrected picture:

| | route 1 (tracker) | route 2 (raw `observeTitle`) |
|---|---|---|
| default config, project `orca` | dormant | **fires** (needs only `fg=codex`) |
| default config, project `codex` | fires | fires |
| `app-name` configured | fires | fires |

**The practical upshot: no special config and no special directory name are needed.** Any Codex session in any project, on stock settings, reaches the title route mid-turn via process inspection. That makes the gate-relocation finding from my first comment considerably more serious than I presented it — it isn't a corner case gated behind an unusual setup.

Everything measured in my first comment still holds (gated + 1 switch = 1 banner, gated + 3 switches = 3, ungated = 4); the mechanism is unchanged. What changes is *how often anyone hits it*: I said "rarely, only with `app-name`," and the honest answer is "by default."

I'd treat this as blocking for the PR's stated goal. The gate closes the hook route, and the raw-title route — wider than either of my earlier descriptions — is left open behind it.

<sub>Found by an adversarial reviewer running `observeTitle` rather than the tracker path; I reproduced it from scratch against the real coordinator before posting, and confirmed `pty-connection.ts:2654` is a live production call site, not test-only. Correcting my own two earlier comments, both of which understated the exposure.</sub>

nwparker:
## Third independent confirmation the discriminator fires — from the live config on this machine

The strongest objection raised against this PR is that Codex never emits `SubagentStart`/`SubagentStop`, so the new discriminator is dead code. I've now checked that from three directions, and the simplest one needs no reasoning about Codex's source at all.

**1. Orca's own managed config already registers the events.** `src/main/codex/hook-service.ts:86-95` lists them in `CODEX_EVENTS`, and the config Orca actually wrote on this machine — `~/Library/Application Support/orca/codex-runtime-home/home/hooks.json` — has them live:

```json
["PermissionRequest","PostToolUse","PreToolUse","SessionStart",
 "Stop","SubagentStart","SubagentStop","UserPromptSubmit"]
```

**2. The shipped binary supports them.** `codex-cli` 0.145.0's embedded schemas include `subagent-start.command.input` (`agent_id`, `agent_type`, `hook_event_name` const `SubagentStart`) and `subagent-stop.command.input` (adds `agent_transcript_path`, `stop_hook_active`). It also carries runtime dispatch strings — `"SubagentStop hook exited with code 2 but did not write a continuation…"` — so it executes them rather than merely parsing them.

**3. Real captured payloads round-trip correctly.** An adversarial reviewer ran an actual `codex exec` under capture hooks, took the real bytes (`turn_id`, `stop_hook_active`, `session_id`, `transcript_path`, `permission_mode`, `last_assistant_message`), and replayed them through `normalizeHookPayload`: identical to the PR's abbreviated fixtures — bare turn → no flag, lead `Stop` mid-subagent → flag true, child `SubagentStop` → no flag. POSTing those payloads to a live `AgentHookServer` got `leadStopWithLiveSubagents: true` through to `setListener`.

### Where the "dead code" claim came from

It's a reasonable mistake: the read-only Codex checkout at `~/projects/codex` is pinned at `a559b65280` (**2026-05-15, ~2 months stale**) and genuinely has no Subagent events — 8 `HookEventName` variants, 16 schema fixtures, zero `subagent-*`, and its only mentions are tests asserting the events are *unsupported*. Anyone reasoning from that clone will reach the same wrong conclusion. **It should be updated before it's used as evidence again.**

### This does not change my verdict

The discriminator fires, and mutation testing backs that up: disabling each of the four production hunks independently kills tests — the coordinator gate at `agent-completion-coordinator.ts:814`, `useIpcEvents.ts:3001`, and both capture sites at `agent-hook-listener.ts:3145`/`:3301`. Reverting all four fails 8 tests across 4 files. The fix is live and load-bearing.

**But my blocking finding stands, and it's orthogonal.** The gate returns *before* `dispatchCompletion` sets `lastCompletedTurn = currentTurn`, so it relocates the banner rather than suppressing it — measured gated + 3 thread switches = **3 banners** (ungated = 4). And it fires on stock config: `observeTitle` at `pty-connection.ts:2654` needs no agent-name token because process inspection supplies identity.

So: the mechanism works, the tests are real, and the PR still lets a banner through on every mid-turn thread switch. That's what needs addressing before it lands.

<sub>The `hooks.json` keys and the `CODEX_EVENTS` list are read directly on this machine; the binary schemas and the `codex exec` capture came from adversarial reviewers and I verified the locally checkable half myself. Not correcting my verdict — reinforcing the part of it that was challenged.</sub>

nwparker:
## The bypass is Orca's own synthetic title — and the net banner count is unchanged

My earlier comments said the gate relocates the banner to the raw-title route rather than suppressing it. That's right, but I had the *mechanism* less precise than it should be, and the corrected version is worse for the PR's stated goal. **Orca mints the idle title itself, from the very `done` hook the gate just suppressed.**

### The loop, verified line by line

1. `src/main/index.ts:1283-1287` — on **every** `agentStatus:set`, main calls `driveSyntheticTitleFromHook(paneKey, payload.state, profile)`, guarded by `shouldDriveSyntheticAgentTitleFromHook`.
2. That guard passes `done` through: `synthetic-agent-title.ts:90` returns `state !== 'working' || profile.synthesizeWorkingTitle !== false`.
3. `src/main/index.ts:1748-1750` injects `` `\x1b]0;${profile.idleLabel}\x07` `` with `force: true` — for Codex, `idleLabel` is **`"Codex ready"`** (`synthetic-agent-title.ts:14`).
4. Those bytes re-enter through `applySyntheticTitleFrame` (`terminal-output-side-effects.ts:211-219`) → `extractAllOscTitles` → `applyObservedTitle` → the agent tracker → `onAgentBecameIdle` → `observeClassifiedTitleCompletion`.

So the gate closes the hook lane, and main immediately re-opens a title lane carrying an unambiguously idle title. `"Codex ready"` contains the `codex` agent-name token, so unlike the default-config case in my last comment it satisfies the tracker outright — no process inspection needed.

**And the natural suppressor structurally cannot fire.** `shouldSuppressTitleCompletionForFreshHook` defers to `isFreshNonDoneAgentStatus` (`agent-status-types.ts:227-233`), which requires `entry.state !== 'done'`. The stored row's state is exactly `done` — that's what triggered the injection. Measured: `freshHookSuppress=false`.

### The count is unchanged, not reduced

Over a realistic 3-turn sequence (lead Stop with 2 live children → lead Stop → 2× SubagentStop → real turn end):

```
PR #10820 (as written)          banners = 3
PR #10820, title route disabled banners = 2
baseline main                   banners = 3
```

**With the title route live, the PR changes the user-visible banner count by zero.** Only the payload differs — `terminalTitle: "Codex ready"` instead of `"codex"`, and `source` is no longer tagged `hook`.

### A latent hazard the PR makes reachable

`dispatchCompletion` writes `lastCompletionIdentityByPaneKey` (`agent-completion-coordinator.ts:298-300`), and `completionIdentityAlreadyNotified` (`:244-262`) cross-matches by `agentIdentity` **across sources**. So a title-route completion poisons subsequent *hook*-route completions for that pane until `recordTitleWorking()` clears the map (`:700`) — and that clear is driven only by a **working title**.

For Codex, `synthesizeWorkingTitle: false` (`synthetic-agent-title.ts:17`), so the reset depends entirely on Codex's native spinner frames. If a user configures `tui_terminal_title` without `activity`, or runs `animations = false`, no working title ever arrives, the map is never cleared, and **every subsequent genuine completion in that pane is silently swallowed**. This hazard exists on `main` too — the PR makes it reachable by routing a dispatch through the title lane where a hook-lane one previously won.

### Test coverage

`codex-lead-stop-subagent-gate.test.ts` contains **zero** occurrences of `observeTitle`, `observeTitleWorking`, or `observeClassifiedTitleCompletion`. Every assertion drives the hook route alone — which is exactly why the suite is green while the banner still appears.

### Recommendation — unchanged, better supported

Still **blocking**, and now for a stronger reason than "there's another route." The gate is on the wrong layer: it discriminates at `isCompletionHookState` inside `observeHookStatus`, while the completion Orca itself manufactures arrives through a lane the gate never sees. A fix wants either (a) the subagent-liveness check applied at `dispatchCompletion` where both lanes converge, or (b) suppression of the synthetic idle-title injection while children are live — at `main/index.ts:1283`, before the bytes are minted.

My earlier finding that all four production hunks are load-bearing (8 tests fail when reverted) is unaffected. This isn't an argument that the PR does nothing — it's that the thing it does doesn't reach the reported symptom.

<sub>The four-step loop, the `state !== 'done'` freshness gate, `synthesizeWorkingTitle: false`, and the zero title-route references in the gate test are all read at the cited lines in this worktree; I verified each after a subagent surfaced them rather than relaying. The banner counts are that subagent's executed harness against the PR branch's real coordinator — I have not re-run that harness myself, and it corrects an earlier version of its own measurement that omitted `observeTitleWorking`.</sub>

PR #10832, opened 2026-07-27
fix(daemon): bound NDJSON and binary-frame ingress

nwparker:
Slice 3 of the re-landed OOM work. Stacked on #10828 — review only the two commits above it.

## ELI5

The daemon talks to the app over a socket. Two things it reads can be made to
eat far more memory than they look like they should.

1. **Binary frames** start with a header saying "here comes N bytes." The
   parser trusted that number and waited. Say 4 GiB and the parser keeps
   everything you send while waiting for a frame that never completes.
   Now it checks N against the existing size cap and rejects immediately —
   nothing to wait for.

2. **JSON lines** were capped at 16 MiB on the wire, but a small line can
   expand into a huge object once parsed. `[[[[[...]]]]]` is mostly brackets.
   4 MiB of that becomes ~109 MiB of memory. Now we look at the line's shape
   before parsing it and reject the amplified ones.

## Why this matters

The wire cap was measuring the wrong thing. It bounded what a message costs
to *receive*, not what it costs to *hold*. At the shipped 16 MiB line cap, a
single admitted line could turn into ~437 MiB of heap — repeatedly.

## Proof the OOM is fixed

Measured against `origin/main`'s real parsers, not reimplementations:

| case | main | this PR |
|---|---|---|
| frame declaring 4 GiB, 256 MiB trickled | 256 MiB retained, still growing | rejected at the header, 0 retained |
| 4 MiB structurally amplified line | 109.2 MiB retained (27.3x) | rejected before parse, ~0 retained |

## Proof of no regressions

**Behavior parity.** The second commit rewrites the structure scanner for
speed. It is differential-tested against the char-by-char implementation it
replaces, over **505,720 cases** with **zero divergence** — 240,000 randomized
over a structural/escape/unicode alphabet under four limit sets, plus
*exhaustive* enumeration of every string up to length 5 over
`{ } [ ] , : " \\ a`. Verdicts are compared including the thrown `resource`
and `limit`, not just throw/no-throw. Hand-picked adversarial cases (odd/even
backslash runs, unterminated strings, astral-plane characters, limit
boundaries at ±1) are now permanent tests.

The coverage is not vacuous: dropping the closing-quote refresh reintroduces
mismatches in that harness, and the failures are **spurious rejections** —
inputs the reference accepts that the mutant throws on — i.e. exactly the
"OOM fix that drops valid data" regression class. Reverting to the naive
rescan fails the linear-time test at 2349ms against a 500ms budget.

**Throughput.** The guard runs on the daemon's hot receive path, so cost
matters. First measurement showed a real 2.95x slowdown — the scan cost 4x
more than `JSON.parse` itself. That's the "OOM fix that degrades UX" failure
mode, so it got fixed rather than accepted:

- 20k scans of a 2 KiB message: **286ms → 5.3ms**
- realistic daemon PTY traffic (20k messages, 40.6 MiB, 646 socket reads):
  main 35.4ms vs 51.0ms here — **795 MiB/s**, far above real load

Skipping a string body in one `indexOf` is only linear if the closing-quote
index is reused across escapes. Deriving it inside the escape loop rescans the
tail once per escape, which is quadratic on a payload carrying escaped text
(a serialized file, or JSON inside JSON). Both cursors are therefore
monotonic, and the escape-dense shape is covered by a timing test:

| 4 MiB body | char-by-char (before) | naive skip | this PR |
|---|---|---|---|
| escape-dense `\n` | 14.0ms | 2330.1ms | **0.9ms** |
| quote-dense `\"` | 12.2ms | 1.5ms | 2.0ms |
| unescaped | 11.2ms | 0.1ms | 0.1ms |

**Tests.** 20,118 passing across `shared` + `main` + `cli` + `relay`.
Two failures in `src/relay/agent-exec-handler.test.ts` are **pre-existing**:
clean `origin/main` reproduces them identically with none of this branch's
code. `terminal-history-incremental-restore` also flakes under load
independently of this branch: interleaved A/B runs (alternating commits to
control for machine-load drift) give **2/14 failures on this PR's base vs
0/14 on its tip**, and the two timing-sensitive cases fail no more often with
this commit than without it. Typecheck and lint clean; the 6 `tsconfig.web`
errors reproduce identically on clean `origin/main`.

## User-facing changes

None. No behavior changes for any input a real peer sends; the new rejections
apply only to input that would previously have exhausted memory.

### DIFF
--- a/src/main/daemon/binary-frame.test.ts
+++ b/src/main/daemon/binary-frame.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from 'vitest'
 import { encodeFrame, createFrameParser, FrameType, FRAME_HEADER_SIZE } from './binary-frame'
+import { FRAME_MAX_PAYLOAD } from './types'
 
 describe('encodeFrame', () => {
   it('encodes a data frame with correct header', () => {
@@ -128,6 +129,22 @@ describe('createFrameParser', () => {
     expect(onFrame).toHaveBeenCalledOnce()
     expect(onFrame.mock.calls[0][0]).toBe(FrameType.Kill)
     expect(onFrame.mock.calls[0][1].length).toBe(0)
+  })
+
+  it('rejects an oversized declared payload from the header alone', () => {
+    const onFrame = vi.fn()
+    const parser = createFrameParser(onFrame)
+    const header = Buffer.alloc(FRAME_HEADER_SIZE)
+    header[0] = FrameType.Data
+    header.writeUInt32BE(FRAME_MAX_PAYLOAD + 1, 1)
+
+    expect(() => parser.feed(header)).toThrow(
+      `Frame payload ${FRAME_MAX_PAYLOAD + 1} exceeds max ${FRAME_MAX_PAYLOAD}`
+    )
+    expect(onFrame).not.toHaveBeenCalled()
+
+    parser.feed(encodeFrame(FrameType.Data, Buffer.from('fresh')))
+    expect(onFrame).toHaveBeenCalledOnce()
   })
 
   it('parses different frame types correctly', () => {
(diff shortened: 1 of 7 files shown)


nwparker:
Closed accidentally; GitHub refuses to reopen a PR whose branch was force-pushed after closing. Recreated as #10874 — same branch (`oom-slice-3-daemon-wire`), same four commits, no content change.

PR #10841, opened 2026-07-27
fix(agent-history): match non-ASCII workspace paths to Claude sessions

nwparker:
## Problem

On macOS, the Agent session history showed **zero Claude Code sessions for any workspace whose path contains non-ASCII characters** (e.g. Korean folder names), even though the transcripts existed in `~/.claude/projects/`. ASCII-only paths worked fine.

Reported by @chkomi, who diagnosed the root cause precisely and verified it live against the v1.4.158 bundle — the analysis below is theirs, confirmed independently here.

## Root cause

macOS file pickers and on-disk names hand Orca **NFD** (decomposed) paths. Claude Code writes **NFC** in both its `~/.claude/projects/<munged>` directory name and the `cwd` it records in each transcript. `normalizeRuntimePathForComparison` folded separators and Windows casing but never Unicode, so the two spellings of the same directory never matched.

The encoder maps each non-alphanumeric run to a single `-`, and NFD decomposes each Hangul syllable into 2–3 code points, so the dash counts diverge:

```
NFC  /Users/me/내 드라이브/한국농어촌공사  →  -Users-me---------------  (34 chars)
NFD  (same path, decomposed)            →  -Users-me---...--------- (60 chars)
```

The prefix match in `listProjectDirs` can never succeed, and the `cwd` check in `isCwdInsideScopePath` would fail too.

## Fix

**1. NFC-normalize in `normalizeRuntimePathForComparison`** — at the *top* of the function, before the WSL UNC branch returns early, so WSL workspaces are covered too. NFC is provably identity on ASCII (verified across all 128 code points), so the common case is byte-for-byte unchanged.

**2. Segment-count slicing in `relativePathInsideRoot`** — this is the part that makes (1) safe, and it is not optional.

Comparison folding was *accidentally* length-preserving before; NFC is the first transform that changes length. `relativePathInsideRoot` sliced the **raw** candidate using an offset taken from the **folded** root, so a shorter NFC prefix cuts an NFD path mid-syllable and returns a **fabricated** relative path — no null, no throw:

```
relativePathInsideRoot(NFD('/Users/me/내 드라이브'), NFD(root) + '/src/index.ts')
  naive NFC fix →  "이브/src/index.ts"   ✗ corrupted
  this PR       →  "src/index.ts"        ✓
```

Critically this fires when **both sides are NFD** — the ordinary macOS case that works today — so a one-line NFC fix would have been a *regression*, not just an incomplete fix. The corrupted suffix flows into editor-tab remapping (whose own comment says it uses this helper precisely so it doesn't fabricate paths), file-explorer reveal, and callers that rejoin it and `stat()` it.

Counting segments survives every fold and keeps the returned suffix **byte-exact** — which matters on Linux and SSH hosts, where NFD and NFC names are genuinely *distinct files*. That byte-exactness is why NFC folding stays confined to comparison keys and never leaks into a path that touches a filesystem.

This also fixes a **pre-existing** over-slice for Windows paths containing `U+0130` (`İ`), which `toLowerCase()` expands from one UTF-16 unit to two.

**3. Watcher IPC hardening** — dropped `root.rootPath ?? rootKey` in `filesystem-watcher.ts`, so a comparison key can never cross the IPC boundary as an absolute path. Both construction sites already set `rootPath`; the type now records that.

## Testing

- `src/shared/cross-platform-path.test.ts` — NFD/NFC containment + project-dir encoding + WSL UNC keys; all four `{root, candidate} × {NFD, NFC}` combinations assert the exact suffix `src/index.ts`, plus the `C:\İş` case. Asserting the *exact* suffix matters: a non-null assertion would have passed against the corrupted output.
- `src/main/ai-vault/session-scanner-claude-unicode-scope.test.ts` — end-to-end reproduction of the reported symptom. Uses `limit: 1` plus a newer ASCII session to exhaust the recency cap, so the Korean session can only surface via scope discovery. **Verified to fail without the fix** (`expected [ 'recent-session' ] to include 'korean-session'`) and pass with it.
- Full suite: **36,987 passing**. Typecheck clean on node/web/cli; oxlint clean.

Two pre-existing failures on this branch are unrelated and reproduce on a clean tree: `src/relay/agent-exec-handler.test.ts` (env-var leak) and the `Ghostty` localization-coverage check.

## Performance

`String.prototype.normalize('NFC')` has a V8 quickCheck fast path — already-normalized ASCII is non-allocating. Measured ~70 ns on a function already costing ~269 ns, ~0.3 ms across a simulated 500-file-event burst. An ASCII guard recovers only a third of that, and the common `charCodeAt`-loop guard benchmarks *slower* than the normalize it skips, so this lands bare with no guard and no memoization.

## Scope

`ORCA_WORKSPACE_ID` is not derived from a normalized path, and normalized values are used as ephemeral in-memory comparison keys rather than persisted state, so **no migration is needed** — the reporter's manual `orca-data.json` workaround can be reverted after upgrading.

## Also fixed: scoped discovery was dead on Windows

Found while auditing the blast radius, and independent of the Unicode bug. Claude derives its project directory name from the **raw** cwd, preserving case (`C:\Users\Ada\repo` → `C--Users-Ada-repo`). Orca encoded from the *comparison key*, which lowercases Windows paths, so it looked for `c--users-ada-repo` and never matched — scoped session discovery never worked for any Windows workspace.

Verified against the real `~/.claude/projects` on disk: directory names preserve case (`-Users-nwparker`, `-private-tmp-orca-agent-qa-repo-d5E0EU`).

`encodeClaudeProjectPath` now encodes from the raw path and emits the NFC spelling alongside it, so both the Windows-case and macOS-NFD cases match. The comparison key was never the right input here — it is a folded lookup key, and this builds a real directory name. Covered by a Windows regression test that fails without the change.

## Also fixed: renderer had its own copy of the containment logic

`lib/path.ts` kept a private `getRelativePathInsideRoot` that folded case but not Unicode, so it missed NFD-vs-NFC matches. It failed *closed* (returned `null`) rather than corrupting, but the misses are user-visible on non-ASCII workspaces: the untitled-file rename dialog rejected valid in-worktree folders ("Folder must be inside the current workspace"), and rich-markdown annotations silently became unavailable. It now delegates to the shared helper rather than growing a second Unicode fold.

## Review notes

Two follow-up commits came out of reviewing this branch:

- The WSL branch of `relativePathInsideRoot` returned a slice of the *comparison key*, which now carries NFC folding — wrong for a Linux filesystem where NFD and NFC name distinct files. Segment skipping already survives the UNC-alias length difference that branch existed for, so both cases now share the raw candidate.
- Differential-tested the new scan against `main` across the ASCII path corpus: byte-identical everywhere except one degenerate case (a UNC-shaped candidate under POSIX root `/` used to return a leading-slash string from a function that returns *relative* paths). Corrected and pinned.

## Review found one bug this branch introduced

Adversarial review caught a real regression in an earlier commit here, now fixed. `U+212A KELVIN SIGN` folds to `K` under NFC, so once the comparison key was normalized, `relativePathInsideRoot` could see a drive path on one side and a POSIX path on the other:

```
relativePathInsideRoot('\u212A:/a\\b', '\u212A:/a\\b/c')  →  ''    (should be 'c')
```

`''` is the sentinel for "candidate *is* the root", so a caller would act on the wrong node. The branch test now runs on the NFC form while the slice still uses the raw string, keeping the suffix byte-exact. Pinned by a test verified to fail without the fix.

One review suggestion was **rejected**: building the ai-vault test fixtures from Orca's own `encodeClaudeProjectPath` would make those tests tautological. The whole bug class is Orca's encoder disagreeing with Claude's real naming — a fixture built with Orca's encoder compares Orca to itself and stays green no matter how wrong both are. The test deliberately spells out Claude's rule instead, with a comment saying why.

## Note on CI

One `verify` run failed in `terminal-history-incremental-restore.test.ts` on two timing/concurrency assertions. That file is unrelated to path handling, passes locally on repeat runs, and an earlier run of this same code passed — it is a flake, not a regression from this branch.

Fixes #10832


nwparker:
Thanks @chkomi — this was an exceptionally good bug report. The NFD-vs-NFC diagnosis, the dash-count arithmetic, and the pointer to the equivalent VSCode-extension issue were all exactly right, and they made this a fast fix to land.

One addition worth flagging, since it affects the shape of the fix versus what you suggested: applying `normalize('NFC')` inside `normalizeRuntimePathForComparison` alone would have introduced a regression. `relativePathInsideRoot` sliced the raw candidate string using an offset computed from the folded root, and comparison folding had been accidentally length-preserving until NFC. So a shorter NFC prefix cuts an NFD path mid-syllable and returns a fabricated relative path — silently, with no null and no throw:

```
relativePathInsideRoot(NFD('/Users/me/내 드라이브'), NFD(root) + '/src/index.ts')
  → "이브/src/index.ts"   instead of   "src/index.ts"
```

That fires when *both* sides are NFD, which is the ordinary macOS case that works today, so it would have broken working paths rather than just leaving yours broken. This PR counts path segments instead, which survives every fold and keeps the returned suffix byte-exact — the latter matters on Linux and SSH hosts, where NFD and NFC names really are distinct files.

Also note your `orca-data.json` NFC workaround is safe to revert once this ships: normalized values are ephemeral comparison keys rather than persisted state, so no migration is involved.

nwparker:
Follow-up after review: two more bugs in the same class turned up, both now fixed on this branch.

**Windows scoped discovery was dead entirely**, independent of the Unicode bug. Claude derives its project-dir name from the raw cwd and keeps case (`C:\Users\Ada\repo` → `C--Users-Ada-repo`), but Orca encoded from the comparison key, which lowercases Windows paths — so it looked for `c--users-ada-repo` and never matched. Confirmed against the real `~/.claude/projects` on disk, where directory names preserve case.

**The renderer had a second copy** of the containment logic in `lib/path.ts`, folding case but not Unicode. It failed closed rather than corrupting, but on a non-ASCII workspace the untitled-file rename dialog rejected valid in-worktree folders and rich-markdown annotations silently went unavailable. It now delegates to the shared helper.

Adversarial review also caught a regression I introduced here: `U+212A KELVIN SIGN` NFC-folds to `K`, so the root and candidate could disagree about being a drive path, and the function returned `''` — the "candidate is the root" sentinel — instead of the suffix. Fixed by branching on the NFC form while still slicing the raw string.

Every fix has a test verified to fail without it.

PR #10843, opened 2026-07-27
feat(skills): run skill updates in the background without a terminal

brennanb2025:
## Summary

The **Update skills** dialog had no primary action at all. Its footer was only `Re-check` and `Close` — the real action was a pre-filled command sitting in an embedded PTY that you had to click into and press Enter on. Orca already builds and validates that command, so it now just runs it.

- **The terminal is gone from the default path.** A headless runner in main executes `npx --yes skills update <names> --global -y`.
- **The run is backgrounded.** Closing the dialog no longer cancels it — main owns the run, and the status bar carries it.
- **Status-bar segment** next to the memory/terminal indicators: spinner while running, a green check on success that clears itself after 4s, and a failure that persists until you act on it. Clicking it reopens the dialog.
- **Versions dropped** from the rows; the skill list and skip reasons are visible instead of hidden behind a disclosure.

## Linear

Part of [STA-664 — Improve Orca agent skill update flow](https://linear.app/stably/issue/STA-664/improve-orca-agent-skill-update-flow).

That ticket asks for a safe update path instead of "sending users to Settings -> CLI", and specifies the exact command form:

> ```bash
> npx --yes skills update orchestration --global --yes
> ```
> Avoid plain `npx skills update -y` because it infers scope from cwd and can update the wrong scope.

This PR ships that verbatim — `npx --yes skills update <names> --global -y`, with `--global` explicit — plus the ticket's safety preconditions: the skill must already be installed, be one of Orca's managed skills, and sit in tracked global/home scope (`SUPPORTED_GLOBAL_SKILL_TOPOLOGIES` gates eligibility). Its fallback rule is honoured too: on failure the dialog surfaces the error and the exact command to copy rather than retrying background writes.

**Not closing the ticket.** STA-664's headline ask is *auto*-update at user-intent boundaries (starting an orchestration workflow, opening a Linear-linked worktree, first computer-use attempt), plus per-runtime scoping and a failure cooldown. This PR builds the mechanism those triggers would call, but the triggers themselves aren't here — so it's `Part of`, not `Fixes`.

## Why it's shaped this way

I unpacked `skills@1.5.20` and read `dist/cli.mjs` rather than guessing at its capabilities:

- **`update` has no `--json`.** `parseUpdateOptions` matches only `-g`, `-p`, `-y`, and positional names; `options.json` is set solely in `parseListOptions`. Unknown flags are silently dropped, so passing `--json` would *look* accepted and do nothing.
- **Progress granularity is per-source, not per-skill.** `updateGlobalSkills` groups by `bySource` and does one `fetchRepoTree` for the whole group. Every Orca skill shares one source, so the slow network phase is a single opaque step. Per-skill progress does not exist to be shown, however you parse.
- So the progress bar is **deliberately indeterminate** rather than a faked percentage, and **per-skill outcomes come from re-scanning `inventorySkillFreshness` after exit** — anything still in `eligibleUpdateNames` didn't land. That's byte-level ground truth needing zero parsing. When the re-scan has a verdict it outranks the exit code.
- Captured output is shown verbatim under **Show log** and never parsed.

Both `--yes` flags are load-bearing and separately tested: `npx --yes` skips the package-install prompt, `skills … -y` takes the CLI's own non-interactive branch. stdin is `'ignore'` so `process.stdin.isTTY` stays falsy — the other half of that gate.

## Also fixed

A width bug the collapsed disclosure used to hide: deep plugin-cache paths set the dialog's width and pushed the footer actions off-screen. Caught only by running the real app.

### The dialog

Every state is the **same row component** — pressing Update changes the row in place rather than swapping the dialog's body for a different layout. A test asserts the row is literally the same DOM node from `available` through `pending` to `done`.

**Before running.** One row per skill, with the badge immediately right of the name. Locations are collapsed behind a per-skill disclosure with a count, instead of every path being dumped inline.

ready

**Expanded.** Paths and their chips on demand.

expanded

**Running.** The row keeps its place and its right edge; the badge gives way to a pending circle left of the name. Indeterminate bar, because the CLI reports no parseable progress.

running

**Success.** Per-skill checks, sourced from the post-run re-scan rather than stdout.

success

**Failure.** Same row, red glyph, reason inline, with the exact command to copy.

failed

<details>
<summary>Status bar (the run survives closing the dialog)</summary>

Spinner while running:

bar running

Green check on success:

bar success

Then it retires itself after 4s:

bar cleared

</details>
## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Verified end-to-end in the real app against genuinely-outdated `orca-linear` and `orchestration`, driven through the actual nudge → dialog → button:

- Both went `outdated` → `current` on disk; `eligibleUpdateNames` emptied.
- Closing mid-run left state `running` — it does not cancel.
- Status segment showed spinner → green check → cleared itself.

New tests: runner arg construction (both `--yes` flags, sorted names), stdin-ignore, shell-syntax name rejection, single-run guard, ANSI/CR stripping, re-scan-outranks-exit-code, re-scan failure fallback, cancel, acknowledge. Dialog tests rewritten for the no-terminal UI. Status-segment lifecycle covered including the success linger and error persistence.

⚠️ **Pre-existing failure, not from this PR:** `verify:localization-coverage` flags unlocalized "Ghostty" keywords in `terminal-advanced-platform-search.ts` and `terminal-pane-appearance-search.ts`. Reproduces identically with this branch stashed.

## AI Review Report

Built and reviewed with Claude Code. Risks checked:

- **Fabricated progress data** — the original design showed per-skill spinners and stage labels. Reading the CLI source proved that data doesn't exist during the slow phase; those states were removed rather than faked.
- **Exit code vs. disk truth** — first implementation let a non-zero exit override a clean re-scan. A test caught the contradiction with its own comment; the re-scan now wins when it has a verdict.
- **Interactive wedge** — confirmed both of the CLI's prompt sites gate on `options.yes || !process.stdin.isTTY` before removing the PTY, so a headless run can't block on input nobody can answer.
- **Cross-platform** — `resolveCliCommand` + `getSpawnArgsForWindows` handle `npx.cmd` on Windows; PATH comes from the existing `hydrateShellPath` login-shell rail, which is what makes nvm/conda setups resolve.
- **max-lines** — the dialog crossed the 400-line cap; split into `skill-freshness-summary-headline.tsx` rather than adding a disable (per AGENTS.md).

**Known gaps:** the runner targets this machine's global skill homes only — WSL/SSH hosts still fall outside it, matching the existing `freshnessInventory` scope. `fetchRepoTree` needs `getGitHubToken`; without one the CLI prints "Failed to fetch tree" and no-ops, which the re-scan correctly reports as a failure but which I did not exercise live.


brennanb2025:
Rebased onto `main` (13 commits behind). The only conflicts were the four non-English locale catalogs, which are generated — resolved by taking main's copies and re-running `sync:localization-catalog` from the merged `en.json`. `en.json` and `web-preload-api.ts` auto-merged. Typecheck, locale verify, and 199 test files / 2609 tests re-run green on the new base. Remote head matches local.

Two notes on CI, since this PR touches `src/shared/skill-freshness.ts` and therefore triggers **Skill update round trip**:

1. **Version coverage.** I read `skills@1.5.20` source to establish there's no `--json` on `update`; that job pins `1.5.17` plus a `continue-on-error` `latest` leg. The flags this depends on (`-g`, `-y`) are long-standing, and the `--json`-only-on-`list` split is what actually matters — but the two versions aren't identical, so the pinned leg is the real signal.

2. **Possible follow-up, deliberately not in this PR.** `verify-skill-update-roundtrip.mjs:210` runs `update <name> --global` under a comment calling it "the exact user-visible rail". As of this PR the shipped rail is `npx --yes skills update <names> --global -y`. Behavior still matches because CI runs non-TTY and the CLI gates on `options.yes || !process.stdin.isTTY`, so both paths take the same branch — but the script no longer literally mirrors what ships. Aligning it would be a one-line change; I left it out to keep this PR to one thing.

brennanb2025:
Follow-up commit `a63abe6ffb` unifies the row layout, per review feedback:

- The ready and running views were separate components with different row shapes, so pressing Update swapped the dialog's body. They're now the same `SkillUpdateRow` instances throughout; a test asserts the row is literally the same DOM node across `available` → `pending` → `done`.
- Locations moved behind a per-skill disclosure with a count, instead of dumping every path inline.
- Status lives in one slot between the name and the count. A leading icon column had nothing to show at rest and just indented every name past an empty gap.
- `groupSkillFreshness` now pins the active run's names so a successful update doesn't drop its own rows the moment the re-scan lands.

**Worth flagging from QA.** To re-stage an outdated skill I overwrote an installed skill's files behind the `skills` CLI's back. The CLI's lock still recorded the current hash, so `skills update` printed `✓ All global skills are up to date` and **exited 0** without changing anything — and Orca's re-scan correctly reported the skill as still outdated and surfaced a failure.

That's the "re-scan outranks the exit code" rule doing real work: a zero exit that changed nothing is not a success. The same thing would happen to a user who hand-edits an installed skill — the old terminal UI would have shown them a green check while the skill stayed stale.

brennanb2025:
## Review pass — 23 fixes pushed (`af01e09baf`)

Reviewed across 6 rounds with three independent adversarial lanes (main-process/IPC, renderer/UX, regression/scope) plus the CodeRabbit comments. All lanes returned clean or non-blocking on the final tree.

### Two defects in the PR's core mechanism

**The re-scan verdict reported success for broken skills.** `rescanOutdatedNames` treated "no longer in `eligibleUpdateNames`" as landed — but a name also leaves that list when it's deleted, half-written (`unrecognized`), or unreadable (`inaccessible`). A failed update showed a green check over a broken skill. Now judged on a positive signal in a new tested module: every convergent placement must return `current` or `newer-known`. `newer-known` counts deliberately — the CLI pulls from the source repo, which runs ahead of the bundled manifest, so demanding strict `current` would invent false failures.

**"Rows keep their place" wasn't true in production.** `refreshSkillFreshness` publishes `{inventory: null}` *synchronously* before its first await, and settling a run notifies it — so every row vanished at the exact moment the result appeared, for the length of a full disk re-hash. The test asserting the same DOM node couldn't see it: it mocks the hook and never passes through null.

### Process lifecycle

- A failed spawn emits `error` **and** `close`, so the second settle clobbered the real `spawn ENOENT` and ran the inventory scan twice. Fixed with a per-run token that also guards the rescan's `finish` closure.
- `child.kill()` killed the `npx` wrapper, leaving the `skills` process still writing to global skill homes. Now uses the existing `killWithDescendantSweep`.
- On Windows a profile directory containing `&`, `%`, or `!` makes the cmd.exe rail throw, which returned a silent `started: false` — Update did nothing, forever, with no error.
- Output pushes are coalesced instead of structured-cloning up to 32 KB to every window on each `\r` progress frame.

### Why there's now a Stop button

**This restores parity, it isn't a new feature.** The embedded PTY let you Ctrl-C a wedged `npx`; removing it left "restart Orca" as the only recovery, while the entire cancel rail shipped across six layers with zero callers. Implementing it correctly required the tree kill above and holding the run `running` until the tree is actually dead — releasing early let an immediate re-Update spawn a **second `npx` writing the same bundles**, which is precisely the corruption the verdict rework exists to detect.

### UI correctness

- Retry sent the eligibility list, which the settling re-scan had just emptied — a guaranteed no-op for the entire window the button is on screen.
- The success linger retired the run 4 s after it finished, blanking the result under an open dialog.
- A skipped skill's reason now renders outside the disclosure, so it doesn't depend on a mount-time `defaultOpen` that a later re-scan can never re-fire.
- Dropped the summary line reading *"Open Update details to see why each one was skipped"* — that control no longer exists, and it was **translated into four languages**, so es/ja/ko/zh users were being fluently directed at nothing.
- Deleted `skill-location-chip-copy.tsx`: 143 lines of unreferenced copy of the old row component, colliding on basename with the `.ts` that bare-specifier imports actually resolve to.

### Verification

`pnpm typecheck` 0 · `oxfmt --check src/` 0 · localization catalog 0 (parity 11211 across es/ja/ko/zh) · **146/146** across the 21 affected suites. Every high-value new test was negative-controlled — reverted against the unfixed source to confirm it fails.

Two pre-existing conditions, both confirmed unrelated by baselining with these changes stashed:
- `pnpm lint` is red on the Ghostty localization strings in `terminal-advanced-platform-search.ts` / `terminal-pane-appearance-search.ts` — as this PR's description already noted.
- A full local `pnpm test` shows broad failures from `window.localStorage` being undefined. This machine runs Node v26.5.0 against the repo's pinned Node 24; Node 26's experimental `localStorage` shadows happy-dom's. Identical failures with the branch stashed, in files this PR never touches. CI on Node 24 is unaffected.

### Known limitations, left deliberately

- **Mixed-topology degradation**: if a run updates the canonical copy but leaves a provider-alias a broken symlink, the alias drops out of the convergent set and the name still reports success. Closing it needs the pre-run inventory threaded through to compare convergent-then against convergent-now — a design change, not a patch. Recorded in `skillUpdateFailedNames`'s docstring so it outlives this thread. Every wholly-degraded and removed-entirely case *is* reported as a failure.
- **No app-quit teardown**: killing at `will-quit` would SIGTERM `npx` mid-write and create exactly the half-written bundle the verdict rework exists to catch. Orphan-and-finish loses only the verdict, and the update is idempotent.

PR #10853, opened 2026-07-27
fix(codex): widen the stale-pane sweep retry ladder past the Windows worst case

brennanb2025:
Stacked on #10770. **Ladder only** — no change to the bind signal, the eligibility predicate, or the queue mechanics.

## The problem

Live Windows 11 validation of #10770 measured store-visible PTY bind → notice raised:

| run | bind → notice |
| --- | --- |
| 1 | 5,140 ms |
| 2 | 3,959 ms |
| 3 | 4,064 ms |

The ladder was `SWEEP_ATTEMPT_DELAYS_MS = [300, 1500, 4000]`, i.e. attempts at **300ms / 1800ms / 5800ms** from bind. On Windows the **final** rung is what produced the card in all three runs — rungs 1 and 2 never won. macOS by comparison resolved on the first rung in ~0.66s, leaving ~3.3s of ladder unused.

So Windows consumes the entire ladder with essentially zero margin. A slower box, more panes, or a colder daemon reattach and the last attempt also misses — at which point there is no further rung, `queueNextRung` drops the pane, and **the card never appears**. The user stays on the old account unaware, which is precisely the bug #10770 exists to fix. The failure mode when the ladder exhausts is "no card", never a dead keyboard, so this is a coverage problem, not a safety one.

## The change

```
- const SWEEP_ATTEMPT_DELAYS_MS = [300, 1500, 4000] as const
+ const SWEEP_ATTEMPT_DELAYS_MS = [300, 1500, 4000, 10_000, 20_000] as const
```

Cumulative attempt times from bind:

| rung | old | new |
| --- | --- | --- |
| 1 | 300 ms | 300 ms |
| 2 | 1.8 s | 1.8 s |
| 3 | 5.8 s | 5.8 s |
| 4 | — | 15.8 s |
| 5 | — | 35.8 s |

Justification against the measurements:

- **Early rungs untouched**, so macOS still resolves in <1s and the startup-burst coalescing is unchanged. This is a pure tail extension — no existing behaviour moves.
- **The tail clears the observed worst case by a wide margin.** The winning rung on Windows was 5.8s; the new rungs sit at ~2.7× and ~6.2× that. Two extra rungs is enough because the growth is geometric — matching the existing ~2.5× shape — so a large time window costs few attempts and therefore few IPC round-trips.
- **Still bounded.** Five attempts, then `queueNextRung` finds `undefined` and deletes the pane's attempt state. It is not interval-based and cannot become a polling loop, which the constant's comment calls out explicitly.
- **A late card is strictly better than no card.** The pane is still stale at t=35s, and the notice is dismissible.

Per-PTY rungs stay per-PTY: `queue`/`takeDuePtyIds` are untouched, so the rung-collapse fix from the earlier review round is not regressed. The sweep still stops for a notified pane (`notifiedPtyIds`) and for a dismissed one (`forgetStaleCodexPanes` drops the launch record, so `listStalePanes` stops reporting it) — a longer tail cannot resurrect either.

## Verification

**Mutation-proof of the new tail.** Reverting the constant to `[300, 1500, 4000]` with the tests unchanged fails exactly the two tests that assert the tail:

```
FAIL  'still sweeps a pane whose reattach outlives the first three rungs'
      → codexRestartNoticeByPtyId['pty-1'] is undefined; inspect count 3, expected 4
FAIL  'gives up after a bounded ladder instead of polling forever'
      → expected inspectProcess called 5 times, got 3
Tests  2 failed | 13 passed
```

The new test drives a pane inconclusive through the first three rungs and asserts the notice is still raised on the fourth — the exact silent-failure scenario the Windows data points at. The bounded-ladder test now advances 120s (well past the 35.8s ceiling) so it pins the ceiling rather than the clock.

**Per-PTY independence test passes unchanged** — `spends a pane its full retry ladder even when another pane binds later` still asserts 2 then 3 inspects for `pty-1` at t=3310/6310, which the longer ladder does not perturb. Only its stale "its 3 attempts" comment was reworded.

Gates, all green on this branch:

- `npm run typecheck` (node + tc.cli + tc.web) — exit 0, 0 errors
- `npx oxlint` — exit 0, no findings in touched files
- `check:max-lines-ratchet` — OK, no new bypasses
- `codex-stale-pane-sweep.test.ts` + `codex-session-restart.test.ts` — 33/33
- `pty-connection.test.ts` (the sweep's only caller) — 484/484

Known-noise on the validation box, unrelated to this change: Node 26 vs the repo-pinned 24 fails ~41 renderer tests (browser-pane/markup, right-sidebar, sidebar) on clean checkouts, and `verify:localization-coverage` exits 1 on 6 pre-existing Ghostty strings.

### DIFF
--- a/src/renderer/src/lib/codex-stale-pane-sweep.test.ts
+++ b/src/renderer/src/lib/codex-stale-pane-sweep.test.ts
@@ -148,6 +148,10 @@ describe('notifyCodexPaneBoundForStaleSweep', () => {
     await vi.advanceTimersByTimeAsync(120_000)
 
     expect(window.api.pty.inspectProcess).toHaveBeenCalledTimes(5)
+    // Why: the count alone only bounds the first two minutes — a rung past the
+    // window would still read as 5. An empty queue is what proves the ladder
+    // ended rather than merely moved out of view.
+    expect(vi.getTimerCount()).toBe(0)
   })
 
   it('still sweeps a pane whose reattach outlives the first three rungs', async () => {
--- a/src/renderer/src/lib/codex-stale-pane-sweep.test.ts
+++ b/src/renderer/src/lib/codex-stale-pane-sweep.test.ts
@@ -144,16 +144,44 @@ describe('notifyCodexPaneBoundForStaleSweep', () => {
     vi.mocked(window.api.pty.inspectProcess).mockRejectedValue(new Error('terminal_gone'))
 
     notifyCodexPaneBoundForStaleSweep('pty-1')
-    await vi.advanceTimersByTimeAsync(60_000)
+    // Well past the last rung (35.8s), so this pins the ceiling, not the clock.
+    await vi.advanceTimersByTimeAsync(120_000)
 
-    expect(window.api.pty.inspectProcess).toHaveBeenCalledTimes(3)
+    expect(window.api.pty.inspectProcess).toHaveBeenCalledTimes(5)
+  })
+
+  it('still sweeps a pane whose reattach outlives the first three rungs', async () => {
+    // Regression: live Windows 11 runs raised the notice on the third rung
+    // (5.8s) in all three samples, i.e. the ladder ended exactly where the
+    // slowest measured box landed. A colder reattach falls past it, and the
+    // failure is silent — no card, so the pane keeps running the old account.
+    vi.mocked(window.api.codexAccounts.listStalePanes).mockResolvedValue([STALE_PANE])
+    vi.mocked(window.api.pty.inspectProcess)
+      .mockRejectedValueOnce(new Error('terminal_gone'))
+      .mockRejectedValueOnce(new Error('terminal_gone'))
+      .mockRejectedValueOnce(new Error('terminal_gone'))
+
+    notifyCodexPaneBoundForStaleSweep('pty-1')
+    // t = 5800: rungs 1-3 are spent and nothing has been raised.
+    await vi.advanceTimersByTimeAsync(5_800)
+    expect(inspectCallCountFor('pty-1')).toBe(3)
+    expect(useAppStore.getState().codexRestartNoticeByPtyId).toEqual({})
+
+    // t = 15800: the fourth rung is the one that catches this pane.
+    await vi.advanceTimersByTimeAsync(10_000)
+
+    expect(useAppStore.getState().codexRestartNoticeByPtyId['pty-1']).toEqual({
+      previousAccountLabel: ACCOUNT_A,
+      nextAccountLabel: ACCOUNT_B
+    })
+    expect(inspectCallCountFor('pty-1')).toBe(4)
   })
 
   it('spends a pane its full retry ladder even when another pane binds later', async () => {
     // Regression: one shared timer used to drain every queued PTY, so a pane
     // already waiting on its 1500ms rung had the next rung consumed by the newer
-    // pane's shorter delay — exhausting its 3 attempts ~2.5s early and dropping
-    // a Windows daemon reattach that had not settled yet.
+    // pane's shorter delay — burning through its ladder early and dropping a
+    // Windows daemon reattach that had not settled yet.
     useAppStore.setState({ ptyIdsByTabId: { 'tab-1': ['pty-1', 'pty-2'] } })
     vi.mocked(window.api.pty.inspectProcess).mockRejectedValue(new Error('terminal_gone'))
 
--- a/src/renderer/src/lib/codex-stale-pane-sweep.ts
+++ b/src/renderer/src/lib/codex-stale-pane-sweep.ts
@@ -5,10 +5,14 @@ import {
 
 // Why: the first delay coalesces the startup burst of binds and lets
 // updateTabPtyId (written just after the layout binding) land, since the scan
-// walks tabs. The later two cover a reattached daemon shell that answers
-// `inspectProcess` with terminal_gone for a beat. Bounded on purpose — this is
-// a hint, and a pane that never resolves must not become a polling loop.
-const SWEEP_ATTEMPT_DELAYS_MS = [300, 1500, 4000] as const
+// walks tabs. The rest cover a reattached daemon shell that answers
+// `inspectProcess` with terminal_gone for a beat. The tail is sized off live
+// Windows 11 runs, where the 5.8s rung is what raised the notice in all three
+// (macOS resolved on the first): the ladder used to end exactly where the
+// slowest measured box landed, so a colder reattach fell off it and the pane
+// stayed silently stale. Bounded on purpose — this is a hint, and a pane that
+// never resolves must not become a polling loop.
+const SWEEP_ATTEMPT_DELAYS_MS = [300, 1500, 4000, 10_000, 20_000] as const
 
 // Why: each PTY owns its rung, so the queue has to carry a per-PTY due time. One
 // shared timer coalesces the startup burst, but it must always target the


brennanb2025:
## Review: LAND-READY

Reviewed against the correct base (`brennanb2025/fix-codex-account-switch`), not main. Verdict on the ladder itself: **clean, no defects.** I pushed one test hardening (944b7cec4a) and opened a separate PR for an unrelated bug this PR amplifies.

### The five things I was asked to judge

**1. Cumulative arithmetic — CONFIRMED.** Not just read; measured with an instrumented copy of the module. Rungs fire at **300ms / 1.8s / 5.8s / 15.8s / 35.8s**, with `attemptsByPtyId` stepping 1..5 and terminating on `SWEEP_ATTEMPT_DELAYS_MS[5] === undefined`. The PR body's table is exact.

**2. Still bounded — CONFIRMED.** Deterministic across 10/10 runs. The attempt counter, not the clock, is the bound, so the catch path (`:140-143`), the early-fire re-arm (`:125-128`, which cannot spin because `takeDuePtyIds` returns `[]` only when every entry is strictly in the future), and a pane whose scan never returns all terminate.

**3. Per-PTY rung independence — NOT regressed.** `queue`/`takeDuePtyIds`/`armForEarliestDue` are untouched by the diff. The longer ladder cannot reintroduce the collapse by another route either: a pane on rung 5 has `attemptsByPtyId` set, so `takeDuePtyIds`' `!attemptsByPtyId.has(ptyId)` early-fold branch skips it. Measured with `pty-2` binding at t=16s while `pty-1` sat on `dueAt=35800` — `pty-1`'s ladder came through undamaged.

**4. The 35.8s tail is safe against stale targets.**
- *Closed pane:* `scanCodexPanes` walks `ptyIdsByTabId`, so a closed pane yields no ptyIds — the remaining rungs cost **zero IPC** and no notice is reachable.
- *Dismissed:* the ladder is already over when the notice is raised (`scan?.notified` → `notifiedPtyIds.add` + `attemptsByPtyId.delete`). For the in-flight race, `markCodexRestartNotices` preserves `dismissed` when the labels match. A late rung cannot re-raise a dismissal.
- *ptyId reuse:* not possible — ids are UUID-derived and teardown drops the record.
- *Manual in-shell `codex` restart at t=20s:* the t=35.8s notice is **correct, not bogus** — `CODEX_HOME` is baked at spawn, so re-running codex in the same shell is still the old account.

**5. Mutation proof — REPRODUCED exactly.** Reverting to `[300, 1500, 4000]` fails precisely the two tail tests the description claims, no more and no fewer. Also probed the boundary properly: moving the 4th rung `10_000 → 11_000` fails the new test, which proves `advance(5_800)` then `+10_000` lands exactly on 15.8s rather than passing on slack.

### One thing I changed (944b7cec4a, test-only)

`gives up after a bounded ladder instead of polling forever` asserted only "5 inspects within 120s". That does not actually prove termination — a **6th rung at 200s passes it unchanged**, because the count cannot distinguish "the ladder ended" from "the next rung moved past the window", which is the exact property the test is named for. Added `expect(vi.getTimerCount()).toBe(0)`. Mutation-proved: passes as-is, and a 6th rung at `200_000` fails it with `expected 1 to be +0` while the call-count assertion stays green.

### On the rebind interaction — I was wrong, and the PR is better than I thought

I flagged that a mid-ladder rebind of the same ptyId *consumes* a rung rather than extending the ladder (`queue`'s `Math.min` pulls the due time forward while `attemptsByPtyId` is untouched), collapsing 35.8s to 26.3s. That is real but **pre-existing, and the widening strictly improves it.** Measured both sides on identical rebind plans:

| plan | 5-rung (this PR) | 3-rung (base) |
| --- | --- | --- |
| no rebind | 300 1800 5800 15800 **35800** | 300 1800 **5800** |
| rebind@6000 | 300 1800 5800 6300 **26300** | 300 1800 5800 6300 7800 **11800** |

At `rebind@6000` the base has already exhausted, so the rebind starts a *fresh* ladder: more IPC, and it still stops earlier. There is no rebind sequence where the widening is worse. It also doesn't fire on the path this PR targets — the cold Windows reattach mounts the pane once, `connectPanePty` runs once, one notify.

### ⚠️ Separate bug this PR amplifies — #10863, fix in #10870

Not a defect in this PR, and deliberately **not** folded into it, but it should land alongside.

#10863 is filed as "the card steals focus from the pane it covers". **The real scope is wider and that matters for severity.** `CodexRestartChip` computes its focus scope as `root.parentElement`, and `Terminal.tsx:2146`/`:2394` mount the card as a *sibling of `TabGroupSplitLayout`* — so that scope is the **entire worktree surface**, not one pane. `shouldFocusMobileDriverAction`'s xterm branch then matches **any** terminal in the worktree. Verified against the helper: `stealsFocusFromSiblingPane: true`, `stealsFocusFromOtherWorktree: false`.

So a notice for pane A yanks focus out of **healthy pane B** — whose input is not blocked and which the user may be typing in — onto the default **Restart** button. `handleRestart` queues restarts for *every* stale pty in the worktree, so one stray Space/Enter destroys those sessions.

**Why it belongs with this PR:** this change moves the first-notice window from ≤5.8s after bind to ≤35.8s. At 5.8s nobody is typing yet; at 35.8s they plausibly are. It turns a rare race into a likely one. Fix is in **#10870**.

### Gates (local)

`codex-stale-pane-sweep` + `codex-session-restart` 33/33 · `pty-connection` 484/484 · typecheck exit 0 · oxlint clean · max-lines ratchet OK. Note this branch is one commit behind its base (missing #10804), which is harmless — #10804 doesn't touch the sweep and the PR is MERGEABLE.

**Not verified:** the live Windows 11 field numbers (3,959 / 4,064 / 5,140 ms) are taken on the author's word — no Windows or SSH host was run. The 10s/20s sizing rests on that premise.

brennanb2025:
## Post-merge Windows validation: the widening was load-bearing, not precautionary

Validated on Windows 11 at `c460de6089` (this change merged). The measurements split into two
regimes, and only one of them matched the numbers this PR was sized from.

**Warm renderer restart** (daemon already up, panes rebind fast) — the method used for the original
sizing:

| Run | Bind | Card | Bind → card |
|---|---|---|---|
| 1 | 6,563 ms | 10,785 ms | 4,221 ms |
| 2 | 6,166 ms | 9,859 ms | 3,693 ms |
| 3 | 6,132 ms | 9,857 ms | 3,724 ms |

Essentially unchanged from the pre-fix numbers (3,959 / 4,064 / 5,140 ms). Still resolved by **rung 3**
(5.8s cumulative).

**Cold full restart** (quit Orca entirely; daemon reattaches ~14 panes):

| Run | Card raised at |
|---|---|
| 1 | 27,693 ms after renderer start |
| 2 | 27,822 ms after renderer start |

Bind is consistently ~6.1–6.6s, so bind → card is **~21 seconds** — past rung 3 (5.8s), resolved by
**rung 4 (15.8s)**, plus ~5–6s for the eligibility scan across the tab set.

### What that means

**Under the old three-rung ladder the cold-restart card would not have appeared at all.** This change
should not be read as over-provisioning; rung 4 is doing real work on the exact path #10770 exists to
fix. The only remaining cushion is rung 5 at 35.8s — roughly one more rung.

The original 3,959–5,140 ms figures were all *warm-reload* measurements, which is why they looked
comfortable. The cold path is ~5× slower.

### Caveats

- In the cold runs the pane was already bound before the poller attached (~22–23s), so **bind is
  inferred from the warm runs**; the card timestamps (27.7s / 27.8s) are directly measured.
- A **cold-parked pane never binds at all**, so it never gets a card. A first attempt using a
  backgrounded pane correctly produced nothing (`terminal_gone`). Not a defect in this change, but it
  means the sweep only ever covers panes that actually bind — worth knowing before anyone sizes the
  ladder again.
- No delayed false positive: a pane running `vim` after exiting Codex got no card and still none 45s
  later, covering rung 4 and part of rung 5.

PR #10862, opened 2026-07-27
perf(editor): cut per-keystroke work on two rich-markdown paths

nwparker:
## Summary

Two independent fixes on the rich-markdown editor's **per-keystroke** path.

**Doc links.** Both ProseMirror plugins walked every text node and ran `matchAll(/\[\[([^[\]\r\n]+)\]\]/g)` on each — the auto-convert `appendTransaction` once per keystroke, and the inline-preview decoration rebuild once per keystroke *and* again on every caret move. A doc link needs a literal `[[`, so gate on a native substring check first. The two walks had duplicated their guard sequence; they now share one predicate.

**Review annotations.** Resolving a comment's block re-serializes the whole document — every node individually, *plus* every adjacent pair, to compute line offsets. Both the highlight-range path and the comment-at-position path did that **once per comment**, making them O(comments × document). Build the blocks once and pass them down.

## ELI5

**Doc links:** the editor scans your whole document on every keystroke looking for `[[wiki-style links]]`. It was running a regex over every paragraph, even ones with no square bracket anywhere in them. Now it checks "does this paragraph even contain `[[`?" first — a check the CPU does natively — and skips the regex for the ~99% that don't.

**Annotations:** to highlight a review comment, the editor works out which paragraph it belongs to by re-rendering the entire document to markdown and counting lines. With eight comments on a page, it did that eight times over. Now it renders once and reuses the result.

## Screenshots

No visual change — decorations, highlights, and link conversion all produce identical output.

**Doc-link walk** — `node config/scripts/rich-markdown-doc-link-scan-benchmark.mjs`, run against **this repo's own markdown in on-disk order**:

```
docs=99 (>3KB: 93) iterations=41 (first-touch, median)
                    corpus     ungated       gated   speedup
         all repo markdown      9.6 us      2.8 us      3.4x
            docs over 3 KB      9.3 us      3.0 us      3.1x
        biggest (97KB doc)     54.6 us     14.3 us      3.8x
```

The benchmark asserts both arms produce **identical match counts** for all 99 documents before timing, so a gate that skipped a real link could not be reported as a win. Timing is first-touch — one bracket per never-before-walked node array — because a naive timing loop in this repo has previously let V8 hoist the call out and report `String.includes` at 29 TB/s.

**Annotation block reuse** — on a 12-node document with 8 comments, document serializations drop from **184 to 23** (measured by counting calls, see below).

## Proof of zero regression

Output is unchanged in both cases. The doc-link gate is a pure early-out on a condition the regex already requires; the annotation change only hoists a computation out of a loop.

**Three new tests**, asserting **serialization counts** rather than wall-clock — deterministic where a timing threshold would be flaky in CI:

- serializes the document once regardless of comment count (1 comment vs 8 → same count)
- serializes once when locating the comment at a position
- does no work at all with zero comments

**These are not vacuous.** Reverting to per-comment block building fails the first with `expected 184 to be 23`. I verified that by reintroducing the old shape.

- [x] `pnpm lint` — oxlint clean on all changed files
- [x] `pnpm typecheck`
- [x] `pnpm test` — **970/970** across `src/renderer/src/components/editor/`

### A note on `max-lines`

Both edits initially pushed their files past the 300-line limit. Per AGENTS.md I did **not** add a suppression or a baseline bump. Instead:

- `rich-markdown-doc-link.ts`: extracted the duplicated guard sequence from both plugins into one `canHoldDocLink` predicate — a genuine deduplication that pays for the new gate.
- `rich-markdown-review-annotations.ts`: folded the prebuilt blocks in as an optional parameter on the existing exported function rather than adding a third function.

Both files now pass `max-lines` on their own merits, and `check-max-lines-ratchet` reports no new bypasses.

Full suite: 4 failures / 37,129. Two are `agent-exec-handler.test.ts`, which assert against the machine's real `process.env`. The other two — `pty-connection` and `multi-client-navigation-isolation` — pass **487/487 in isolation** and import nothing from the changed files. They join the load-flaky pool this series has tracked across ~15 full-suite runs, where two consecutive runs on unmodified `main` produce different failure sets.

**2 review loops: independent re-measurement and final risk review. Final verdict: ship.**

Found by a discovery sweep over `components/editor` (264 files, previously unscanned), then independently re-measured before implementing. I verified the annotation call chain by hand — `comments.flatMap(...)` calling a function that rebuilds blocks — rather than taking the finding's word for the O(comments × document) claim.

The same sweep **rejected** a neighbouring candidate with numbers: `findRichMarkdownSearchMatches` builds a whole-document visible-text map on every query change, which looks like the classic "O(document) per keystroke" find — but measured **0.88 ms on a 28 KB document** with 12 distinct never-before-searched queries, and search input is debounced. Not worth it.

## AI Review Report

Risks reviewed:

- **Gate correctness (doc links).** `[[` is a strict prerequisite of the pattern — it cannot match without it — so the gate is logically implied by the regex. Verified empirically across all 99 repo markdown files: identical match counts.
- **Predicate refactor.** `canHoldDocLink` folds in the pre-existing `node.type.name !== 'text'`, `!node.text`, and `isDocLinkLiteralCodeTextNode` checks in the same order both call sites used. The existing code-context tests (inline code marks, fenced code blocks) still pass, which is what pins that ordering.
- **Annotation API compatibility.** `getRichMarkdownAnnotationHighlightRangesForComment` is called externally from `useRichMarkdownReviewRailController.ts`; its signature is unchanged apart from a new *optional* trailing parameter, so that caller is unaffected and still builds its own blocks.
- **Empty-input paths.** Both annotation entry points now return early with zero comments, which also removes a wasted document serialization on documents with no review comments at all.

Cross-platform: no path, shell, or platform-dependent behaviour touched — this is in-memory editor state. Identical on macOS/Linux/Windows, and unaffected by SSH/WSL/remote hosts or folder-workspace vs git-worktree.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. Document text was already being scanned and serialized by these code paths; this PR only reduces how often. The doc-link gate cannot cause a link to be *missed* (proven by the match-count assertion), so link resolution and its existing `isReservedRichMarkdownTransportBody` guard are unchanged. No dependency changes.


nwparker:
All four addressed in 8c01281.

**The serialization assertion was the important one** — you're right, and it's the exact failure mode this PR is supposed to be guarding against. `expect(many).toBe(oneComment)` passes just as happily when *both* paths build blocks twice, so the test would have gone green on a straight revert of the hoist. It was asserting "consistent," not "once."

Now pinned absolutely:

```ts
const NODE_COUNT = 12
const ONE_BUILD_SERIALIZE_CALLS = NODE_COUNT + (NODE_COUNT - 1)  // 23: each node, plus each adjacent pair
```

Confirmed 23 empirically before pinning it (not derived from the impl and then asserted back at it), and confirmed the constant is now load-bearing: reverting `prebuiltBlocks ?? …` back to an unconditional `buildRichMarkdownCommentBlocks(editor)` fails **2 tests**. Under the old assertion it failed none.

**`index % 0`** — real crash, fixed. Empty corpora are now skipped with a printed reason rather than silently reading `undefined`, and `measure()` throws if called with one, so the guard can't quietly rot into dead code. Verified by forcing `large = []`: the row prints `no docs in this corpus — skipped` and the other two rows still measure.

**`path.join`** and **the header comment** — both done.

Re-verified after the changes: `src/renderer/src/components/editor/` is 146 files / 970 tests green, lint and format clean, and the benchmark still reports 2.8–3.7×.

PR #10866, opened 2026-07-27
perf(agent-status): strip terminal control bytes by run, not per character

nwparker:
## Summary

`stripTerminalControl` built its result with a per-character `+=`, allocating a fresh string for every retained character.

The Command Code status detector calls it **four times per PTY chunk** — the combined scan text, the chunk-boundary variant, and both previous-text lengths — on every write once a Command Code pane is live. Control bytes are sparse in real agent output, so copying the spans *between* them is far cheaper than appending character by character.

## ELI5

Agent output arrives full of invisible control codes that have to be removed before Orca can read the text.

The old code walked the text one character at a time and glued each keeper onto a growing string, creating allocation pressure for every retained character. The detector's four per-chunk calls are bounded to 4,096, 4,096, 300, and 301 code units, so the work is bounded but repeated.

Now it finds the stretches of clean text between control codes and copies each stretch in one go. Same output, a fraction of the work — and it happens four times for every chunk of output an agent produces.

## Screenshots

No visual change.

`node config/scripts/terminal-control-strip-benchmark.mjs`:

```
Complete stripTerminalControl path. Lower is better.
iterations=501 (1002 first-touch samples/implementation, counterbalanced median)
                  fixture    per-char   slice runs   speedup
          300 history TUI      1.5 us       1.0 us     1.50x
         301 boundary TUI      1.5 us       1.0 us     1.46x
            4096 scan TUI     18.6 us      11.6 us     1.60x
       4096 scan 49.6% C0     14.7 us      18.9 us     0.78x

validated=4008 measured pairs, result checksum=2159192968
Production calls are bounded to 4096, 4096, 300, and 301 code units.
```

The benchmark mirrors the complete pre/post `stripTerminalControl` paths, including the unchanged guard and ANSI stripping. Its agent-TUI-shaped fixtures match the detector’s production bounds: 300 code units of recent history, the 301-unit chunk-boundary variant, and 4096-unit scan text. The pathological dense fixture is also bounded to 4096 units and alternates printable bytes with C0 controls (49.6% controls).

Timing is **first-touch and counterbalanced**: each measured pair independently builds equal-length, output-equivalent inputs for the two arms, neither input is reused, and every iteration measures both call orders. Every timed output is checked for equality and folded into the printed checksum so V8 cannot discard the work.

The representative TUI rows are **1.46–1.60× faster**. The pathological 49.6%-control row is honestly slower at **0.78×**; production logic is unchanged because real TUI control bytes are sparse and every call is already bounded.
## Proof of zero regression

The output is byte-identical. Since this is a pure string function, I proved that rather than argued it:

- **Exhaustive** over every string up to length 4 across a 13-symbol alphabet chosen to cover the branch conditions: `\x00`, `\x07`, `\x1f` (C0 kept/stripped boundary), `\n`, `\r` (the two explicit exemptions), `\x7f`, `\x9f` (C1 range edges), `\xa0` (just past it), plus ASCII, accented, CJK, and an astral emoji.
- Plus **200,000 random strings** from the same alphabet.
- **224,831 inputs, 0 mismatches.**

That covers the cases this rewrite could plausibly break: a control byte at index 0, at the final index, several adjacent, and a string containing none at all (which now returns the input untouched rather than rebuilding it).

**Four new tests** driving the real exported detector — control bytes leading, trailing, adjacent, and none — plus multi-byte preservation.

**Mutation tested, 2/2 killed:**

| mutant | result |
|---|---|
| drop the trailing run (`return output`) | **86 tests fail** |
| drop the run copy, keep only the skip | **2 tests fail** |

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — **3134/3134** across `src/shared/`; 92/92 in the touched file

Full suite: 5 failures / 37,128. Two are `agent-exec-handler.test.ts`, which assert against the machine's real `process.env`. The others — `transcript-watch` ×2 and `project-view-wrapper-source-context-boundary` — pass **30/30 in isolation** and import nothing from the changed file. They join the load-flaky pool this series has tracked across ~16 full-suite runs, where two consecutive runs on unmodified `main` produce different failure sets. (A sixth reported failure was a stray scratch file another agent left in this shared checkout; removed.)

**3 review loops: implementation review, adversarial correction, and clean re-review. Final verdict: ship.**

Found by a discovery sweep over `src/shared`, then verified by hand before implementing — I traced the call chain to confirm the four-calls-per-chunk claim and, importantly, checked **where the guard stops helping**: the detector has a `!hasSeenCommandCodeUi` early-out, but that latches to `true` after the banner is seen, and the four strip calls sit *after* it. So on a live Command Code pane this runs unconditionally per chunk.

The same sweep surfaced two neighbouring candidates I **rejected with numbers**:

- **`shouldIncludeQuickOpenPath` substring allocations** — 1.76 ms for the *entire* 10,287-file repo (0.172 µs/path), and only during an explicit search. Its own doc comment already notes it avoids a split allocation.
- **MJPEG frame parser re-concat** — the concat only fires on the bounded-remainder path; frames are subarray *views*, pending is trimmed, and it already has its own benchmark. Prior work got there first.

## AI Review Report

Risks reviewed:

- **Boundary conditions.** A control byte at index 0 means the first run is empty (guarded by `index > runStart`); one at the final index means no trailing run to append; none at all means `runStart === 0` and the input is returned unchanged. All three are covered by the exhaustive sweep and by explicit tests.
- **The no-control fast path.** `runStart === 0` returns `withoutAnsi` directly rather than a rebuilt copy. This is only reachable when zero control bytes were found, so it cannot skip a strip — and note the function is already gated by `terminalControlMayAffectText`, so it is a second-level optimization, not a substitute for that guard.
- **Multi-byte safety.** Iteration is by UTF-16 code unit, exactly as before, and `slice` respects code-unit boundaries the same way index-based access did. Surrogate pairs are never split differently than the original — both halves are above `0x9f` and so are never stripped. Covered by CJK and emoji fixtures.
- **ANSI stripping is untouched** — the two `replace` calls that precede this loop are unchanged.

Cross-platform: no path, shell, or platform-dependent behaviour. Pure in-memory string processing, identical on macOS/Linux/Windows, and unaffected by SSH/WSL/remote hosts — though remote panes benefit equally since the detector runs on their output too.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or IPC surface. The same bytes are stripped by the same predicate; only the string-building strategy changed. This is defensive sanitization of untrusted PTY output, and its behaviour is provably unchanged (224,831-input differential), so the sanitization guarantee is preserved exactly. No dependency changes.

### DIFF
--- a/config/scripts/terminal-control-strip-benchmark.mjs
+++ b/config/scripts/terminal-control-strip-benchmark.mjs
@@ -1,16 +1,11 @@
 #!/usr/bin/env node
-// Benchmark: stripTerminalControl, which runs four times per PTY chunk.
-//
-// createCommandCodeOutputStatusDetector calls it on the combined scan text, the
-// chunk-boundary variant, and both previous-text lengths — every chunk, once the
-// Command Code UI has been seen. It built its result with a per-character `+=`
-// append, which allocates a fresh string per retained character.
-//
-// Control bytes are sparse in real agent output, so copying the spans between
-// them is far cheaper. Both arms are compared on every fixture before timing.
+// First-touch benchmark for stripTerminalControl, called four times per live PTY chunk.
+// Every measured pair uses distinct, output-equivalent inputs and counterbalanced call order.
 import { performance } from 'node:perf_hooks'
 
 const ITERATIONS = Number(process.env.ORCA_STRIP_BENCH_ITERATIONS ?? '31')
+let resultChecksum = 0
+let validatedPairs = 0
 
 if (!Number.isSafeInteger(ITERATIONS) || ITERATIONS <= 0) {
   throw new Error(`ORCA_STRIP_BENCH_ITERATIONS must be a positive integer, got ${ITERATIONS}`)
@@ -20,7 +15,6 @@ function isStrippedCode(code) {
   return (code <= 0x1f && code !== 0x0a && code !== 0x0d) || (code >= 0x7f && code <= 0x9f)
 }
 
-// Pre-fix: append one character at a time.
 function stripPerChar(text) {
   let output = ''
   for (let index = 0; index < text.length; index += 1) {
@@ -32,7 +26,6 @@ function stripPerChar(text) {
   return output
 }
 
-// Post-fix: copy the runs between control bytes.
 function stripSliceRuns(text) {
   let output = ''
   let runStart = 0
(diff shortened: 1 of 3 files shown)


nwparker:
Addressed in bd80fbfd. Both comments were walking through the implementation; each is now a single line of non-obvious rationale (`// Why slice runs: control bytes are sparse and this runs 4x per PTY chunk.` and `// Run-copying breaks at run boundaries, so these pin start/end/adjacent/none.`). 92 tests still green.

Issue #10519, opened 2026-07-25
[Other]: main E2E has been red for 24 consecutive scheduled runs


MumuTW:
## Summary

The scheduled `E2E` workflow on `main` has failed **24 consecutive runs**, since `2026-07-13T22:18Z` ([29289456191](https://github.com/stablyai/orca/actions/runs/29289456191)). The last green scheduled run was `7b3103f89` at `2026-07-13T17:37Z` ([29271199883](https://github.com/stablyai/orca/actions/runs/29271199883)) — 12 days ago.

This is not one regression. The failing set churns run to run (6–8 specs at a time; some resolve on their own, others persist for over a week), which is what a suite looks like when nobody is reading it. `release-cut.yml` gates on this same workflow via `workflow_call`.

**Environment:** Linux (`ubuntu-latest`, `electron-headless` project), `main`.

## Two regressions worth bisecting

Both specs predate the failure and used to pass, so these are genuine regressions rather than tests that landed broken.

### `daemon-slow-health-check-preservation.spec.ts:38`

> preserves a live daemon PTY when the daemon is too slow for the startup health check

| | run | sha | date |
| --- | --- | --- | --- |
| last pass | [29663142435](https://github.com/stablyai/orca/actions/runs/29663142435) | `4905b7828` | 07-18T22:16Z |
| first fail | [29783555704](https://github.com/stablyai/orca/actions/runs/29783555704) | `e218dfbf8` | 07-20T22:20Z |

90 commits in the window. The preserve log line never arrives:

```
Error: expect(received).toMatch(expected)
Expected pattern: /preserv\w*\s+daemon[^\n]*health check/i
Call Log:
- Timeout 10000ms exceeded while waiting on the predicate

  > 111 |       await expect
    112 |         .poll(() => stderrLines.join(''), { timeout: 10_000 })
    113 |         .toMatch(/preserv\w*\s+daemon[^\n]*health check/i)
```

One candidate in that window stands out by name, though I have not confirmed it: `7adda25b0` — "fix(daemon): retire empty current-generation daemons" (#9277). The test asserts a live daemon is *preserved*; that commit changes when daemons are *retired*.

### `onboarding.spec.ts:420`

> Onboarding flow › Skip preserves runtime server project setup UI

| | run | sha | date |
| --- | --- | --- | --- |
| last pass | [29962554850](https://github.com/stablyai/orca/actions/runs/29962554850) | `9bdb8204e` | 07-22T22:21Z |
[...]

PR #10868, opened 2026-07-27
ci(e2e): run the E2E specs a PR changes (#10518)

MumuTW:
## Summary

`e2e.yml` is reachable only from `schedule` and `release-cut`'s `workflow_call`, so a PR that adds an E2E regression test never runs that test — it runs for the first time on `main`, where nothing gates on it. Twice now a spec has landed and never passed (#10518).

This adds a `pull_request` job that runs exactly the specs the PR touched. A PR that changes no spec pays nothing: the file list is resolved first, and every setup/build/run step is gated on the result being non-empty.

Gating the *full* sharded suite on every PR stays a separate decision, deliberately not made here: it is 22 minutes of wall clock (build + 10 shards + the docker job), and `main` is red today, so making it required would block every merge until #10519 closes.

## Screenshots

No visual change.

## Proof of fix

The upstream PR changes no spec, so its own `e2e changed specs` check can only demonstrate the skip path. The run path is demonstrated on a throwaway fork PR whose head is this commit plus one comment-only line in `tests/e2e/markdown-explorer-find-focus.spec.ts`:

**Run path** — [MumuTW/orca run 30242976474](https://github.com/MumuTW/orca/actions/runs/30242976474), 2m53s total:

```
Running 1 changed spec(s):
  tests/e2e/markdown-explorer-find-focus.spec.ts
> ... npx playwright test --config tests/playwright.config.ts --project=electron-headless \
      --pass-with-no-tests tests/e2e/markdown-explorer-find-focus.spec.ts
  ✓  1 [electron-headless] › tests/e2e/markdown-explorer-find-focus.spec.ts:7:5 › Explorer-opened Markdown accepts the find shortcut without a document click (23.7s)
  1 passed (25.5s)
```

It selected exactly the touched file — not the whole suite, not the shard containing it.

**Skip path** — [MumuTW/orca run 30243157972](https://github.com/MumuTW/orca/actions/runs/30243157972), same commit with no spec change, **16 seconds** end to end:

```
success  Checkout
success  Select changed E2E specs
skipped  Install native build and headless UI tools
skipped  Setup Node.js
skipped  Setup pnpm
skipped  Use external node-gyp ...
skipped  Install dependencies
skipped  Build Electron app for E2E
skipped  Run changed E2E specs
```

That is what the vast majority of PRs will pay. This PR's own check on `stablyai/orca` is `action_required` pending maintainer approval, which is why the evidence above is from the fork.

## Design notes

- **No `paths:` filter.** The check reports on every PR, so it can be made a required check without wedging the merge box on PRs that touch no spec. A `paths:`-filtered workflow simply never reports, and a required check that never reports blocks the merge forever. Selection happens inside the job instead.
- **Changed files come from the pulls API** (`gh api --paginate .../files`), not a git diff: the PR checkout is shallow, `--paginate` avoids the 100-file cap of `gh pr view --json files`, and `select(.status != "removed")` keeps a spec the PR deleted from reaching Playwright.
- **`--pass-with-no-tests`** — a spec whose tests are all `@headful` is filtered out by the `electron-headless` project, and "no tests ran" must not fail the gate.
- **Same invocation as the scheduled suite** — `xvfb-run --auto-servernum env SKIP_BUILD=1 ORCA_E2E_FORWARD_APP_LOGS=1 pnpm run test:e2e`, reusing the prebuilt output, so a spec cannot pass here and fail on the nightly for harness reasons. A unit test asserts these tokens against `e2e.yml` itself, so the two cannot drift silently.
- Specs with their own runners self-skip without their env var (`ORCA_E2E_SSH_DOCKER` etc.), so selecting one is a cheap no-op rather than a false failure.

## Testing

- [x] `pnpm lint` (oxlint, changed files)
- [x] `pnpm typecheck` — n/a, no TS product code changed; the added script is `.mjs` and its test runs under vitest
- [x] `pnpm test` — ran the `config/scripts` suite: 54 files, 433 passed, 4 skipped, including the 11 new cases
- [ ] `pnpm build` — not run; no application code changed

`config/scripts/e2e-changed-specs-workflow.test.mjs` covers both halves:

- **Selector behavior** — keeps only `tests/e2e/**/*.spec.ts`, walks nested dirs, dedupes and sorts, emits nothing when no spec changed, and drops paths a fork PR could weaponize (`; rm -rf /`, `$(id)`, backticks, spaces, `--shard=1/2`, `..` traversal, absolute paths).
- **Workflow contract** — no `paths:` filter, the exact `types:` list, the API-based selection command, `permissions: {contents: read, pull-requests: read}`, every work step gated on a non-empty selection, the trace upload gated on `failure()`, and the run invocation matching `e2e.yml`'s own shard command token for token.

Also verified the command shape locally against the real Playwright config, since "the args are forwarded correctly" is the load-bearing claim:

```
$ npx playwright test --config tests/playwright.config.ts --project=electron-headless \
    --pass-with-no-tests --list tests/e2e/markdown-explorer-find-focus.spec.ts tests/e2e/onboarding.spec.ts
Total: 13 tests in 2 files
```

## AI Review Report

Reviewed for: required-check semantics, shell injection through PR-controlled filenames, cost on the common path, and drift between this gate and the scheduled suite.

Flagged and changed as a result:

1. **`paths:` filter would break a future required check** — a filtered workflow never reports on PRs that touch no spec, so the merge box waits forever. Moved selection inside the job and pinned that invariant with a test.
2. **`mapfile` is bash 4+** — replaced with a `while IFS= read -r` loop so the snippet is not silently tied to the runner's bash version, and the whole run snippet was then executed locally under bash 3.2.
3. **Trace upload was initially caught by the "everything is gated" assertion** — it is correctly gated on `failure()` instead; the test now exempts it explicitly rather than loosening the rule.

Cross-platform: the job is `ubuntu-latest`-only by design (it mirrors `e2e.yml`, which is Linux-only), and the `runner.os == 'Linux'` guard on the node-gyp pin matches `e2e.yml`/`pr.yml` so the "this workaround is Linux-only" invariant stays consistent if a matrix is ever added. The selector script is pure Node with POSIX path handling and no shell. No shortcuts, labels, or Electron platform behavior touched.

## Security Audit

The one real surface is **command execution with attacker-controlled input**: the selected paths become argv for a shell-invoked `pnpm run test:e2e`, and a fork PR chooses its own filenames.

- Paths pass a strict allowlist (`^tests/e2e/[A-Za-z0-9._-]+(/[A-Za-z0-9._-]+)*\.spec\.ts$`) plus an explicit `..` rejection — everything else is dropped, not escaped. Six weaponized-path cases are unit-tested.
- They reach the command as `"${SPECS[@]}"` read from a file, never interpolated into the command string.
- `permissions:` is narrowed to `contents: read` + `pull-requests: read`; the checkout uses `persist-credentials: false`.
- Trigger is `pull_request`, not `pull_request_target`, so fork code never runs with a privileged token or repo secrets.
- No new dependency, no network access beyond the GitHub API and the existing install path.

Follow-up: none required for this change.

## Notes

- This does **not** make E2E required on PRs, and does not run anything for a PR that only changes product code — that is the broader gate, and it is blocked on `main` going green (#10519, currently 2 red specs with #10621 open for one of them). Worth deciding separately.
- Timeout is 30 minutes. A PR touching an unusually large number of specs would hit it rather than being silently truncated; the step prints the full list it is running.

### DIFF
--- a/.github/workflows/e2e-changed-specs.yml
+++ b/.github/workflows/e2e-changed-specs.yml
@@ -77,13 +77,12 @@ jobs:
         with:
           run_install: false
 
-      # Why: same Linux-only node-gyp pin as e2e.yml and pr.yml — pnpm's bundled
-      # gyp_main.py ships without execute permission and breaks node-pty.
-      - name: Use external node-gyp to avoid pnpm's bundled copy (Linux only)
+      # Why: pnpm's bundled node-gyp ships gyp_main.py without execute
+      # permission and breaks node-pty. Use the lockfile copy instead of
+      # installing one outside the lock.
+      - name: Use lockfile node-gyp to avoid pnpm's bundled copy (Linux only)
         if: steps.select.outputs.specs != '' && runner.os == 'Linux'
-        run: |
-          npm install -g node-gyp@11.5.0
-          echo "npm_config_node_gyp=$(npm root -g)/node-gyp/bin/node-gyp.js" >> "$GITHUB_ENV"
+        run: echo "npm_config_node_gyp=$(pwd)/node_modules/node-gyp/bin/node-gyp.js" >> "$GITHUB_ENV"
 
       - name: Install dependencies
         if: steps.select.outputs.specs != ''
--- a/.github/workflows/e2e-changed-specs.yml
+++ b/.github/workflows/e2e-changed-specs.yml
@@ -1,0 +1,121 @@
+name: E2E (changed specs)
+
+# Why: e2e.yml is reachable only from schedule and release-cut's workflow_call,
+# so a PR that adds an E2E regression test never runs that test — it runs for
+# the first time on main, where nobody is gating on it (#10518). This gate runs
+# exactly the specs the PR touched: it closes the "test lands red" hole without
+# putting the full 22-minute sharded suite on every pull request.
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+
+# Why: a force-push should not leave the previous Electron run burning a runner.
+concurrency:
+  group: e2e-changed-specs-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  changed-specs:
+    name: e2e changed specs
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+      pull-requests: read
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      # Why: the PR's file list comes from the API rather than a git diff — the
+      # default PR checkout is shallow, and its merge ref would need extra
+      # fetches to produce the same answer.
+      - name: Select changed E2E specs
+        id: select
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          gh api --paginate \
+            "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
+            -q '.[] | select(.status != "removed") | .filename' > changed-files.txt
+          node config/scripts/select-changed-e2e-specs.mjs < changed-files.txt > changed-specs.txt
+          {
+            echo 'specs<<SPECS_EOF'
+            cat changed-specs.txt
+            echo 'SPECS_EOF'
+          } >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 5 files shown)


MumuTW:
Closing this as superseded by #11131, which merged the upstream path-filtered PR E2E workflow and closed #10518. The changed-spec selector approach may still be useful as a future optimization, but it no longer needs to remain as a parallel implementation PR; it can be revived as a focused follow-up if maintainers want that direction.

Issue #10375, opened 2026-07-24
[Bug]: Regression of #7035 — embedded browser tab reports `document.hasFocus()===false`, so `Input.insertText`/`execCommand` no-op in rich editors (X/Draft.js)


RakhimovY:
## Summary

The embedded automation browser can no longer commit text into rich-text web editors (reproduced on X/Twitter's Draft.js reply composer). Text injected via CDP or page JS never reaches the editor model, so the site's submit button stays disabled. The single underlying symptom is that **`document.hasFocus()` returns `false` on the automation tab even when that tab is the active Orca tab and the Orca app window is frontmost at the OS level.**

This is the same class of bug as #7035 ("CDP proxy: programmatic focus ignored, so `Input.insertText` / `page.fill()` no-op"), closed 2026-07-16 as fixed by #7041. It appears to have regressed.

## Environment

- Orca: **1.4.152** (installed 2026-07-23)
- Electron: **43.1.0**
- Chromium: **150.0.7871.47**
- OS: macOS (Darwin 25.4.0)

## Steps to reproduce

1. `orca tab create --url "https://x.com/<user>/status/<id>"` in a logged-in X session.
2. Make the tab active and bring Orca frontmost.
3. Open the inline reply composer and try to type through any input path:
   - `orca inserttext --text "..."` (CDP `Input.insertText`)
   - `orca type --input "..."`
   - `orca computer type-text --text "..."` (OS-level keystrokes)
   - `orca eval` running `document.execCommand('insertText', false, "...")`

## Actual behavior (all reproduced deterministically)

| Input path | Result |
|---|---|
| `execCommand('insertText')` via `orca eval` | text renders into the DOM, but the editor model stays empty -> submit button remains `aria-disabled="true"` |
| CDP `orca inserttext` | returns `{inserted:true}` but commits nothing (not even to the DOM) |
| `orca type` (CDP) | same no-op |
| `orca computer type-text` (OS keystrokes) | returns `ok:false`; `computer capabilities` reports `windows.focus:false` |
| Orca brought frontmost via `osascript -e 'tell application "Orca" to activate'` | `document.hasFocus()` still `false` |

[...]

Issue #8139, opened 2026-07-10
[Bug]: Keyboard is taken over when using Orca browser skill


ryanmargono:
### Operating system

macOS

### Details

When using orca browser skill, if the agent is trying to type something, it takes over my keyboard so I cant type and multi task. My keyboard ends up interfering with its task as well (ofc as its vice versa)

nwparker:
## Attempted repro — cannot_repro

### Tried
1. Reviewed computer-use / browser skill keyboard injection paths at a high level.
2. Could not run a full agent browser-skill session that synthesizes key events while measuring host keyboard steal without hijacking the interactive desktop of this machine.

### Result
Needs an isolated computer-use automation that asserts keyboard focus ownership during agent typing. Not reproduced here.

grboy6770:
## Reproduced — the trigger is `click`, not typing

Re: the `cannot_repro` above. I think the repro failed because the search was aimed at the typing path. In my measurements typing is **not** the trigger — synthetic **mouse presses** are.

`Input.dispatchMouseEvent` with `mousePressed` makes Chromium hand native focus to the guest view that received the press, so an agent clicking in a background tab pulls the caret out of whatever pane the user is typing in.

### Method (no interactive-desktop hijack needed)

Drive a scratch tab with `orca` CLI on a loop while a human types continuously in a terminal pane in the same window. Each segment writes execution evidence into the page (a click counter and the final input value) so a silently-skipped command can't be mistaken for a clean run.

Measured on 1.4.158 / macOS, against a local `file://` page — no app under test required:

| Command | Commands issued | Caret in the user's pane |
|---|---|---|
| `snapshot` / `get` / `is` / `network` | 32 | held |
| `eval` (DOM `.click()` + value set) | 224 @ 1.5/s | held |
| **`fill`** | 37 + 38 | **held** |
| **`click`** | 39 + 38 | **lost** |

### Order-reversed crossover

Running one command type per segment and then reversing the order rules out a warm-up or time-of-day confound:

| Run | Order | click segment | fill segment |
|---|---|---|---|
| 1 | click → fill | caret lost | caret held |
| 2 | fill → click | caret lost | caret held |

Same result per command regardless of position.

### Why the other paths are fine

The two neighbouring input paths already compensate, which is why typing looked innocent:

- `Input.insertText` replays the pending `DOM.focus` (#7041)
[...]

PR #10875, opened 2026-07-27
fix(browser): restore host focus after synthetic mouse presses

grboy6770:
## Summary

Background browser automation stole the host's keyboard focus on every click, pulling the caret out of whatever pane the user was typing in.

`Input.dispatchMouseEvent` with `mousePressed` makes Chromium hand native focus to the guest view that received the press. The two neighbouring input paths already compensate for this — the synthetic mouse path did not:

| CDP path | orca command | Compensation today |
|---|---|---|
| `Input.insertText` | `fill`, `type` | `DOM.focus` replay (#7041) + `webContents.focus()` |
| `Runtime.evaluate` | `eval` | Deliberately excluded from auto-focus |
| **`Input.dispatchMouseEvent`** | **`click`, `drag`, `check`** | **none** |

The exclusion for `Runtime.evaluate` already states the intent in `cdp-ws-proxy.ts`:

> Do not auto-focus generic `Runtime.evaluate`/`callFunctionOn` traffic: wait polling and read-only JS probes use those methods heavily, and **focusing on every eval steals the user's foreground window while background automation is running**.

This PR completes that intent for the remaining path.

New `host-focus-guard.ts` captures the focused host `WebContents` before dispatching synthetic mouse events and hands focus back afterwards, in a `finally` so a rejected or timed-out dispatch cannot strand focus on the guest. Applied to `click`, `drag`, `check`.

**DOM focus still follows the click** — only the OS/app-level focus owner is restored, so automation that relies on `document.activeElement` (e.g. `click` then `fill`) is unaffected.

Guarded edge cases: guest already owns focus → no restore (the user is driving that view); captured contents destroyed mid-click → no-op; focus never moved → no re-focus (re-focusing an already-focused `WebContents` can churn selection).

## Screenshots

`No visual change` — main-process focus routing only, no UI surface touched.

## Testing

- [ ] `pnpm lint` — fails on `main` too, in files this PR does not touch. `verify:localization-coverage` rejects untranslated `"Ghostty"` keywords in `terminal-advanced-platform-search.ts` / `terminal-pane-appearance-search.ts` (last touched by #10628). Deliberately **not** silenced via `config/localization-coverage-allowlist.json` here, since those strings belong to another change. `oxlint` reports 0 findings on the four files changed in this PR.
- [ ] `pnpm test` — 3462 passed / 8 files failed, **all from my local Node version, not from the repo or this PR.** The repo pins `engines.node: 24`; I ran on v25.8.1, where the global `localStorage` shadows the jsdom one. Every one of the 43 failures is `window.localStorage.clear/getItem/removeItem is not a function` in `src/relay/` and `src/renderer/` — none in `src/main/browser/`, and they reproduce identically on unmodified `origin/main` (verified in a detached worktree). CI runs Node 24, so it is authoritative here.
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/browser/` — **29 files, 526 passed**, the module this PR changes
- [ ] `pnpm build` — not run
- [x] Added tests: `host-focus-guard.test.ts` covers capture/restore, guest-already-focused, destroyed contents, no-op restoration, third-view focus change, and the rejection path (dispatch throws → restore still runs)

## AI Review Report

Reviewed with Claude (Sonnet 4.5) plus an independent Codex pass used adversarially against my own conclusions.

**Risks checked and what changed as a result**

- **Rejection path** — flagged by CodeRabbit and fixed in `96f4b64d`: the first revision restored focus after the awaits, so a rejected or timed-out dispatch skipped restoration and left focus on the guest — the exact failure mode the guard exists to prevent. Now wrapped in `try/finally` at all three call sites, with a test asserting restoration on throw.
- **Over-restoring** — re-focusing an already-focused `WebContents` can churn selection in some editors, and restoring when the guest legitimately owns focus would start a focus fight. Both guarded, both tested.
- **Scope of the claim** — Codex pushed back on inferring "eval is safe" from a single measurement window. Rule-of-three on the sample gives roughly a 1.8% per-command upper bound, so the finding is stated as "`click` is the trigger" rather than "the other paths are proven risk-free"; `type`/`inserttext` are explicitly left out of scope below.
- **Instrument validity** — `document.hasFocus()` never returns true for these guest views (consistent with #10375/#10376 and Electron 43+ tracking `WebContentsView` focus more strictly), so it was discarded as a measurement tool and the finding rests on order-reversed observation instead.

**Cross-platform**: reviewed explicitly. The change uses only `webContents.getFocusedWebContents()`, `isDestroyed()` and `focus()` — no platform branches, no paths, no shell, no shortcuts or labels. `focus()` semantics differ per platform in *how* the OS raises a view, but the guard only restores whatever owned focus beforehand, so it cannot introduce a platform-specific ordering that did not already exist. The measurements were taken on macOS; the logic is platform-neutral but has not been exercised on Linux or Windows.

## Security Audit

- **Input handling** — no user input parsed; the only inputs are an integer `guestId` and a `WebContents` handle obtained from Electron.
- **Command execution / path handling** — none introduced.
- **Auth / secrets** — none touched. No credentials, tokens or storage read or written.
- **IPC** — no new IPC surface. The guard runs inside existing main-process CDP command handlers.
- **Privilege / cross-view exposure** — the restore target is captured from `getFocusedWebContents()` immediately before dispatch and is only ever handed focus back to itself; the guard cannot move focus to a view that did not already hold it. It never reads or writes guest content.
- **Denial of service** — `restoreHostFocus` is a bounded, non-recursive call with no retry loop; the `finally` placement means it runs exactly once per operation.
- **Follow-up** — none identified.

## Notes

**Repro** — related to #8139, which was blocked at `cannot_repro`:

> Needs an isolated computer-use automation that asserts keyboard focus ownership during agent typing. Not reproduced here.

It did not reproduce because **the trigger is not in the typing path**. Measured on 1.4.158 (macOS), driving a scratch `file://` tab while a human typed continuously in a terminal pane in the same window:

| Command | Commands issued | Caret in user's pane |
|---|---|---|
| `snapshot` / `get` / `is` / `network` | 32 | held |
| `eval` (DOM `.click()` + value set) | 224 @ 1.5/s | held |
| **`fill`** | 37 + 38, both orders | **held** |
| **`click`** | 39 + 38, both orders | **lost** |

Order-reversed crossover, excluding a warm-up or time-of-day confound:

| Run | Order | click segment | fill segment |
|---|---|---|---|
| 1 | click → fill | caret lost | caret held |
| 2 | fill → click | caret lost | caret held |

Each segment carried execution evidence (a page-side click counter and the final input value) so a silently-skipped command could not be mistaken for a clean run.

**Scope / what this does not cover**

- Only `Input.dispatchMouseEvent` is changed. `Input.insertText` (`type`, `inserttext`) and `Page.bringToFront` (`tab switch --focus`) still call `webContents.focus()` by design; `type`/`inserttext` were not measured.
- Marked **Related to** rather than **Fixes** #8139 for that reason — the reporter described typing, and I only have evidence for the mouse path. Happy to retarget if a maintainer confirms the remaining paths are in scope.
- Interaction with #10376 (`Emulation.setFocusEmulationEnabled`) is untested. That PR touches `cdp-bridge.ts` and `cdp-ws-proxy.ts` and changes what `document.hasFocus()` reports; this change is orthogonal (OS-level focus owner, not page-level focus emulation) but the two are worth verifying together. GitHub currently reports this branch as mergeable against `main`.
- Verified on macOS only.

**Test-side change**: `cdp-bridge-integration.test.ts` needed `getFocusedWebContents` added to its `electron` mock; that is the only non-source edit.

## ELI5

When an agent clicked something in a browser tab in the background, your computer decided the browser was now the thing you were typing into — so your cursor vanished out of the terminal mid-sentence. Typing and running scripts already knew not to do that. Clicking didn't. Now it puts your cursor back where it was.

### DIFF
--- a/src/main/browser/cdp-bridge-integration.test.ts
+++ b/src/main/browser/cdp-bridge-integration.test.ts
@@ -11,7 +11,9 @@ const { webContentsFromIdMock } = vi.hoisted(() => ({
 }))
 
 vi.mock('electron', () => ({
-  webContents: { fromId: webContentsFromIdMock },
+  // Why: cdp-bridge's host focus guard (#8139) reads the focused contents around
+  // synthetic mouse presses; the integration mock needs the same surface.
+  webContents: { fromId: webContentsFromIdMock, getFocusedWebContents: vi.fn(() => null) },
   shell: { openExternal: vi.fn() },
   ipcMain: { handle: vi.fn(), removeHandler: vi.fn(), on: vi.fn() },
   app: { getPath: vi.fn(() => '/tmp'), isPackaged: false }
--- a/src/main/browser/cdp-bridge.ts
+++ b/src/main/browser/cdp-bridge.ts
@@ -182,27 +182,28 @@ export class CdpBridge {
       const { cx, cy } = await this.getPageCoordinates(guest, node, localCenter.cx, localCenter.cy)
 
       // Why: the synthetic press hands native focus to the guest, so remember who
-      // owned it and give it back (#8139).
+      // owned it and give it back even if a dispatch rejects (#8139).
       const hostFocus = captureHostFocus(guest.id)
-
-      // Why: mouseMoved fires mouseenter/mouseover so sites reveal hover-dependent menus/targets before the click lands.
-      await sender('Input.dispatchMouseEvent', { type: 'mouseMoved', x: cx, y: cy })
-      await sender('Input.dispatchMouseEvent', {
-        type: 'mousePressed',
-        x: cx,
-        y: cy,
-        button: 'left',
-        clickCount: 1
-      })
-      await sender('Input.dispatchMouseEvent', {
-        type: 'mouseReleased',
-        x: cx,
-        y: cy,
-        button: 'left',
-        clickCount: 1
-      })
-
-      restoreHostFocus(hostFocus)
+      try {
+        // Why: mouseMoved fires mouseenter/mouseover so sites reveal hover-dependent menus/targets before the click lands.
+        await sender('Input.dispatchMouseEvent', { type: 'mouseMoved', x: cx, y: cy })
+        await sender('Input.dispatchMouseEvent', {
+          type: 'mousePressed',
+          x: cx,
+          y: cy,
+          button: 'left',
+          clickCount: 1
+        })
+        await sender('Input.dispatchMouseEvent', {
+          type: 'mouseReleased',
+          x: cx,
+          y: cy,
+          button: 'left',
+          clickCount: 1
+        })
+      } finally {
+        restoreHostFocus(hostFocus)
+      }
 
       return { clicked: element }
     })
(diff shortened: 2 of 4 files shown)


grboy6770:
Thanks — both addressed in 96f4b64d.

**Restore in `finally` (Major):** correct, and it was the exact failure mode this guard exists to prevent. A rejected or timed-out dispatch after `mousePressed` left focus stranded on the guest. `click`, `drag` and `check` now wrap their dispatch sequences so restoration runs on the rejection path, plus a test that asserts `restoreHostFocus` still fires when the dispatch throws.

**Module comment (Nitpick):** condensed to the native-vs-DOM focus contract; dropped the implementation history.

`pnpm exec vitest run --config config/vitest.config.ts src/main/browser/` — 29 files, 526 passed.

PR #10881, opened 2026-07-27
perf(ssh): normalize watch event paths once per fs.changed batch

nwparker:
## Description

`routeSshFilesystemWatchNotification` fanned every `fs.changed` batch out to every watch registration by calling `isPathInsideOrEqual(registration.rootPath, event.absolutePath)` per pair. That helper NFC-normalizes **both** sides on every call, so each event path was re-normalized once per watch root and each root once per event — O(roots × events) normalizations for O(roots + events) distinct work.

This change normalizes each event path once up front and builds one pre-normalized matcher per root, leaving only string comparison in the inner loop.

The boundary logic now lives in exactly one place: `createNormalizedPathInsideOrEqualMatcher` in `cross-platform-path.ts`, with `isPathInsideOrEqual` delegating to it. The name is deliberately explicit because `normalizeRuntimePathForComparison` is **not idempotent** for WSL UNC paths — `//wsl.localhost/Ubuntu/A` folds to `//wsl/ubuntu/A`, which a second pass would lowercase further — so passing a raw candidate would silently fail to match. There's a regression test pinning exactly that case.

This is an SSH-only path: it's how remote file watching reaches the file explorer and the editor's external-change detection.

## ELI5

The app watches folders on a remote machine. When files change, it has to figure out which watched folder each changed file belongs to.

Before: for every folder × every changed file, it cleaned up **both** names from scratch before comparing them. With 25 folders and 5,000 changed files that's 125,000 cleanups of 5,025 actual names.

After: clean each name once, then just compare. Same answer, about 8× less work.

## Proof of perf improvement

`config/scripts/ssh-watch-fanout-benchmark.mjs` (new). Three consecutive runs:

| roots | events | per-pair | hoisted | speedup |
|---|---|---|---|---|
| 3 | 20 | 0.042 ms | 0.011 ms | **3.6–4.0×** |
| 6 | 50 | 0.215 ms | 0.043 ms | **5.0–5.1×** |
| 12 | 200 | 1.75 ms | 0.259 ms | **6.6–6.8×** |
| 25 | 500 | 9.25 ms | 1.10 ms | **8.2–8.4×** |
| 25 | 5000 | 95.5 ms | 11.6 ms | **8.2–8.3×** |

The bottom row is a remote `git checkout` or `pnpm install` storm — the relay batches up to `MAX_BATCHED_WATCHER_EVENTS` (5,000) per notification. That's ~84ms of main-process blocking removed per batch, at exactly the moment the main process is already busy.

Honesty notes on the measurement:

- **The arms are interleaved.** My first version ran each arm's whole batch separately, and on a loaded machine (load avg 81) that alone swung the 12×200 row between 6.7× and 23.3×. The 23.3× was noise, not a result. Alternating per round and taking per-arm medians made it reproducible at 6.6–6.8×. The interleaving rationale is in a comment so it isn't undone later.
- **The benchmark imports the real normalizer** from `cross-platform-path.ts` rather than re-modelling it, so a change to the folding rules can't silently invalidate the numbers.
- **This times routing only.** It is not an end-to-end app measurement; the saving is real but bounded to this function.
- Fixture paths are real repo paths (`git ls-files`), since path length and segment count drive normalization cost and a synthetic `/a/b/c` would understate it.

The benchmark's guards were themselves verified by breaking the implementation three ways — matcher always-true, matcher always-false, and a renamed export — and confirming each throws (`routing differs…`, `…is stale`) rather than silently timing a broken arm. This is the vacuous-guard mistake I made on #10741; I check for it now.

## Proof of zero regression

- **New test file**: `ssh-filesystem-watch-notifications.test.ts`, 9 tests. The router had **no test coverage at all** before this. One test asserts routing is identical to per-pair `isPathInsideOrEqual` over a fixture with sibling-prefix, trailing-slash, doubled-slash, and root-itself cases; others pin the WSL UNC non-idempotency case, NFC/NFD matching (macOS emits NFD), sibling-prefix rejection, multi-callback delivery, and non-`fs.changed` methods.
- **Mutation testing** (6 mutants, `--no-cache`), all killed:

  | # | Mutant | Result |
  |---|---|---|
  | 1 | Drop the root-equality arm | killed |
  | 2 | Drop the boundary slash (sibling prefix leaks in) | killed |
  | 3 | Double-normalize the candidate (WSL non-idempotency) | killed |
  | 4 | Skip event-path normalization | killed |
  | 5 | Deliver all events regardless of root | killed |
  | 6 | Notify only the first callback | killed |

  Mutant 2 initially "passed" — but the perl that applied it failed to compile, so the mutant was never applied. Re-ran it with a real edit; it dies.
- **Blast radius**: `isPathInsideOrEqual` has ~15 call sites across renderer, main, relay, and shared. I ran every test file covering a call site — **26 files, 1,976 tests, all passing**.
- `src/main/providers/` + `cross-platform-path.test.ts`: **40 files, 611 tests** passing.
- `pnpm typecheck` clean; `pnpm build:electron-vite` succeeds.

3 loops. Round 1 caught the API footgun (a matcher taking a pre-normalized candidate, named as if it took any path) and renamed it to `createNormalizedPathInsideOrEqualMatcher` with the non-idempotency documented. Round 2 caught the non-interleaved benchmark producing an unreproducible 23.3×. Round 3 caught the vacuous mutant-2 run and two NUL bytes that had corrupted string literals in the benchmark (git was classifying the file as binary).

**Verdict: ship.** Behaviour is pinned by a from-scratch test suite on a previously untested router, all 6 mutants die, and the win is reproducible.

### DIFF
--- a/config/scripts/ssh-watch-fanout-benchmark.mjs
+++ b/config/scripts/ssh-watch-fanout-benchmark.mjs
@@ -1,0 +1,190 @@
+#!/usr/bin/env node
+// Benchmark: routing one `fs.changed` relay notification to SSH watch registrations.
+//
+// routeSshFilesystemWatchNotification called isPathInsideOrEqual(root, event.path)
+// for every (registration x event) pair. That helper NFC-normalizes BOTH sides, so
+// each event path was re-normalized once per watch root, and each root was
+// re-normalized once per event -- O(roots * events) normalizations for what is
+// O(roots + events) distinct work.
+//
+// The fix normalizes each event path once up front and builds one pre-normalized
+// matcher per root, leaving only string compare in the inner loop.
+//
+// This is a hot path on SSH: the relay watcher batches up to MAX_BATCHED_WATCHER_EVENTS
+// per notify, and a single `git checkout` or `pnpm install` on the remote host emits
+// thousands of paths through it.
+//
+// Both arms are run against the same inputs and their outputs are compared before
+// timing, so a matcher that changed which events route where cannot be reported as
+// a win. The normalizer is imported from the real module (via tsx) rather than
+// re-modelled here, so folding-rule drift cannot silently invalidate the result.
+import { execFileSync } from 'node:child_process'
+import { readFileSync } from 'node:fs'
+import { performance } from 'node:perf_hooks'
+import { fileURLToPath } from 'node:url'
+
+const REPO_ROOT = fileURLToPath(new URL('../..', import.meta.url))
+const ITERATIONS = Number(process.env.ORCA_SSH_WATCH_BENCH_ITERATIONS ?? '200')
+const WARMUP = Number(process.env.ORCA_SSH_WATCH_BENCH_WARMUP ?? '30')
+
+for (const [name, value] of [
+  ['ORCA_SSH_WATCH_BENCH_ITERATIONS', ITERATIONS],
+  ['ORCA_SSH_WATCH_BENCH_WARMUP', WARMUP]
+]) {
+  if (!Number.isSafeInteger(value) || value <= 0) {
+    throw new Error(`${name} must be a positive integer, received ${value}`)
+  }
+}
+
+// Why re-read the source: this benchmark's whole claim is that the normalizer is
+// the expensive part. If someone makes it cheap (or drops the NFC fold), the
+// numbers below stop meaning what the header says, so fail loudly instead.
+const PATH_SOURCE = readFileSync(
+  new URL('../../src/shared/cross-platform-path.ts', import.meta.url),
+  'utf8'
+)
+for (const marker of ['normalize(', 'createNormalizedPathInsideOrEqualMatcher']) {
+  if (!PATH_SOURCE.includes(marker)) {
+    throw new Error(`cross-platform-path.ts no longer contains ${marker}; this benchmark is stale`)
+  }
+}
+
+// Import the real normalizer so both arms fold paths exactly as production does.
+const {
+  normalizeRuntimePathForComparison,
+  isPathInsideOrEqual,
+  createNormalizedPathInsideOrEqualMatcher
+} = await import(new URL('../../src/shared/cross-platform-path.ts', import.meta.url).href)
+
+// Pre-fix: mirrors the original routeSshFilesystemWatchNotification inner loop.
+function routeBefore(roots, events, sink) {
+  for (const rootPath of roots) {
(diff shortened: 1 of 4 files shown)


nwparker:
Adding the full-suite evidence, now that I have a same-machine clean-`main` baseline to compare against.

`pnpm test` is **already red on `main`**, and the failing pool is load-flaky — two runs of one unchanged tree produced different failure sets (12 tests / 5 files, then 8 / 2). So a green suite isn't the bar; the bar is *"the post-change failure set is a subset of main's."*

Clean `origin/main` on this machine:

| tree | failing files | failing tests |
|---|---|---|
| clean `origin/main` | `relay/agent-exec-handler`, `renderer/terminal-search-decoration-leak` | 8 / 37,142 |

This branch touches `src/shared/cross-platform-path.ts`, `src/main/providers/`, and one benchmark script — no renderer or relay files. The targeted evidence is what carries the zero-regression claim here:

- every test file covering an `isPathInsideOrEqual` call site: **26 files, 1,976 tests** passing
- `src/main/providers/` + `cross-platform-path.test.ts`: **40 files, 611 tests** passing
- 6/6 mutants killed

I'd rather state that explicitly than imply a green full suite that doesn't exist on this repo today.

nwparker:
Both review points addressed in d6466cd.

**Arm ordering** — this one was a real bias, thank you. Every round measured `before` first, which handed `after` the second slot every single time; any within-round drift (thermal ramp, a GC settling) would push the ratio consistently one way. Now the leading arm alternates across an even round count, so each arm takes the first slot exactly half the time, and the median is taken over the middle pair since there's no single midpoint at an even count.

Worth noting the numbers **did not move**: 8.1× and 8.2× on the 25×5,000 row across two fresh runs, against 8.2–8.3× before. So the original result wasn't an ordering artifact — but that's a thing I could only claim *after* balancing it, not before.

**Runner** — correct, the `node` shebang is misleading since the script imports `cross-platform-path.ts` directly (deliberately: both arms then fold paths exactly as production does, so a change to the folding rules can't silently invalidate the benchmark). Documented the required invocation at the top of the file:

```
// Run with:  npx tsx config/scripts/ssh-watch-fanout-benchmark.mjs
```

I kept the `#!/usr/bin/env node` shebang because all eight `.ts`-importing benchmarks under `config/scripts/` share it — changing one in isolation would be inconsistent without changing the family, which is out of scope here.

Re-verified after the change: the equivalence guard still throws (`routing differs at 3 roots x 20 events`) when the matcher is broken, and the 9 router tests still pass.

PR #10877, opened 2026-07-27
fix(skills): classify foreign same-name skills separately

5Hyeons:
## Summary

Classifies a readable same-name skill found in a plugin cache as `foreign` instead of `unrecognized`.

This prevents another tool's skill from raising Orca's `Needs attention` pill or being presented as an Orca copy to remove. Copies in locations Orca owns remain `unrecognized`; empty or unreadable packages remain unrecognized because there is not enough evidence to dismiss the warning. Eligibility and update commands are unchanged, and plugin-cache placements are not update targets.

## Screenshots

Manual QA launched the integration dev app, but the changed Skills state could not be reached because the Electron window did not accept focus/navigation. No PR screenshot is attached yet.

## Testing

- [ ] `pnpm lint` (full lint retains the pre-existing `Ghostty` localization-coverage failure)
- [x] `pnpm typecheck`
- [ ] `pnpm test` (focused skill suites only)
- [ ] `pnpm build` (not run)
- [x] Focused skill suites: 35 tests passed after the final simplification.

## AI Review Report

Claude reviewed the classification boundary and a local integration merge with #10865. It checked positive evidence for plugin-cache location and readable content, ownership/eligibility behavior, the `Needs attention` pill, UI grouping, path handling, and the simplified no-snapshot-kinship rule. The integration merge passed typecheck and focused tests; the conflict was isolated to the shared display-status module. Cross-platform path, shell, shortcut/label, and Electron-specific risks for macOS, Linux, and Windows were reviewed; this change adds no platform-specific shell or shortcut behavior and no remote/SSH behavior.

## Security Audit

Reviewed path classification, readable-file handling, symlink/path escape implications, command execution, IPC, secrets, auth, and dependencies. No new command execution, auth, secret, dependency, or IPC surface was added. Empty or unreadable plugin packages remain fail-closed as `unrecognized`.

## Notes

This remains separate from #10865 and does not change update eligibility. After #10865 is merged by the maintainer, rebase this PR onto the resulting latest `main`, resolve any shared display-status conflict, and rerun final checks.

X handle: N/A

Refs #10633

Issue #10659, opened 2026-07-26
[Bug]: Plugin-cache depth limit creates nonexistent per-skill "Inaccessible" placements


darrentmorgan:
### Operating system

macOS

### Orca version

v1.4.156

### Details

**Short summary:** `scanKnownPluginSkillCandidates` refuses to open directories below depth 9. Ordinary OpenAI-bundled plugin payloads nest deeper, so the scan returns those subtrees as `incompletePaths`, and `inventorySkillFreshness` converts each one into a placement **per manifest skill** at `join(incompletePath, skill.name)`. Orca never stats that joined path and it does not exist on disk, but it is recorded as an installation with `status: 'inaccessible'` and rendered in `Update skills` as an `Inaccessible` copy. On this machine that is 3 incomplete paths × 8 manifest skills = 24 fabricated placements, three of which are visible in the dialog right now.

#### What the user sees

`Settings → General → Agent skills → Details` (Orca 1.4.156, unmodified install state):

> **Update skills**  <br>⚠ Some installed Orca skills were left out of the update.  <br>Open Update details to see why each one was skipped.
>
> **orca-linear** `Skipped`  <br>Orca couldn't read this copy, so it left the skill out of the update.
>
> 
> | path | chip |
> | -- | -- |
> | `/Users/<me>/.agents/skills/orca-linear` | `Current` |
> | `/Users/<me>/.codex/plugins/cache/openai-bundled/sites/0.1.31/skills/sites-building/templates/vinext-starter/examples/d1/app/orca-linear` | `Inaccessible` |
> | `/Users/<me>/.codex/plugins/cache/openai-bundled/sites/0.1.31/skills/sites-building/templates/vinext-starter/examples/d1/db/orca-linear` | `Inaccessible` |
> | `/Users/<me>/.codex/plugins/cache/openai-primary-runtime/presentations/26.723.12215/skills/presentations/assets/builtin_templates/codex-grid-layout-library/assets/previews/orca-linear` | `Inaccessible` |
> | `/Users/<me>/developer/.../infra/.agents/skills/orca-linear` | `In a repo` |

None of those three `Inaccessible` paths exist. `ls` on each returns `ENOENT`. They are string joins of a directory the scanner declined to open and a skill name from the manifest. The tooltip tells the user Orca "couldn't read this copy" and cites a permissions or file error; the real cause is Orca's own depth bound, and those three parent directories are perfectly readable.

#### Why it happens

[...]

PR #10884, opened 2026-07-27
fix(skills): complete the ordinary plugin cache scan and require skill evidence

5Hyeons:
Repair pass on top of #10865 (`agent/fix-plugin-scan-coverage`). That PR stopped the scanner from fabricating per-skill `Inaccessible` placements, but two blocking problems survived, both reproducible against a real `~/.codex/plugins/cache`. `#10877` and `#10635` are untouched.

## 1. The ordinary scan did not finish, and its bounds became the same symptom

A skill package's own payload — templates, sample apps, vendored fixtures — was walked as if it were a skill tree. Ordinary OpenAI-bundled plugins therefore pushed past the depth and entry bounds, and the resulting diagnostics set **every** skill to `Needs attention` via `getSkillFreshnessDisplayStatus`, which returned `needs-attention` for any non-empty `scanIssues`. That is the symptom #10659 asked to remove, reintroduced one layer up. The entry bound also set `limitReached`, aborting the rest of the walk.

Traversal is now topology-aware. A directory carrying `SKILL.md` is a skill package; Orca descends a bounded `MAXIMUM_NESTED_SKILL_DEPTH` (2) below it so a skill grouped inside a package is still found — the existing nested-skill contract — then stops. Pruning payload is a topology decision, not a coverage failure, so it is silent.

Limits are kept, not removed:
- `MAXIMUM_PLUGIN_SCAN_ENTRIES` 4096 → 16384, sized against a measured real cache (~7.1k entries read once payload is pruned; ~2.3x headroom).
- Depth, candidate, manifest, and issue bounds unchanged.

Orca's own bounds no longer drive the attention pill. `isSkillScanIssueNeedingAttention` (shared, so main and renderer agree) keeps `io-error` and `outside-root` actionable — those are facts about the user's disk — and treats `depth-limit`, `entry-limit`, `candidate-limit`, `manifest-limit`, and `issue-limit` as coverage facts. The Details dialog still lists **every** skipped path and still reports `scan-incomplete`, so no information is lost; it simply stops being escalated to a per-skill badge the user cannot act on.

## 2. Live false-positive candidate rule

Outside a declared skill root, any directory whose bare name matched a known Orca skill was emitted as a candidate with no evidence, and the match `continue`d — so it also **masked** the genuine skill beneath it.

On this machine that produced `…/cache/openai-bundled/computer-use` and `…/cache/openai-primary-runtime/presentations`. Neither contains a `SKILL.md`; both are plugin directories. The real skills sit at `…/computer-use/1.0.1000502/skills/computer-use` and `…/presentations/26.723.12215/skills/presentations` and were never reached.

A candidate now requires a real `SKILL.md` (regular file, or a symlink resolving to one). Inside a declared skill root an uninspectable link still records a candidate — there the plugin manifest itself claims the name is a skill, so fail-closed is warranted; outside one there is no such claim and no file to read, so inventing a copy would be a guess.

## Preserved separations

Symlink handling (`outside-root` refusal, `ELOOP` as `io-error`), the dangling-known-name-symlink-in-a-declared-root candidate, non-blocking manifest inspection (FIFO), and the no-fabricated-installations guarantee from #10865 are all unchanged and still covered.

## Real cache: before → after

| | before | after |
| -- | -- | -- |
| candidates | 3 — 2 fabricated, 1 of them masking a real skill | 3, all with a verified `SKILL.md` at their true path |
| issues | `depth-limit`, `entry-limit` | none |
| pill | `Needs attention` on every skill | driven only by real placements/faults |

## Tests

Added: real-shaped Codex cache completes with no coverage issues; a plugin directory sharing a skill name is not emitted; a bare known-name directory with no `SKILL.md` is not emitted; payload pruning beyond the nested budget stays silent; eligibility unchanged when a genuine plugin-cache copy is present; per-reason diagnostic behavior for both `getSkillFreshnessDisplayStatus` and `hasSkillCopyNeedingAttention`.

Updated: the candidate-budget fixture now carries `SKILL.md` (the rule it asserted changed); the entry-budget fixture is resized to the new bound.

## Validation

- focused suites — 21 files, 157 tests, all passing (`src/main/skills`, `src/main/ipc/skills.test.ts`, `src/renderer/src/components/skills`, display-status, nav-install-status, `useSkillFreshness`)
- `pnpm typecheck` — clean
- changed-file `oxlint` + `oxfmt`, `pnpm run lint:switch-exhaustiveness`, `check:max-lines-ratchet`, `check:reliability-gates` — clean
- real `~/.codex/plugins/cache` scan, asserted `issues` is empty

Refs #10659

### DIFF
--- a/src/main/ipc/skills.test.ts
+++ b/src/main/ipc/skills.test.ts
@@ -77,6 +77,7 @@ describe('registerSkillsHandlers', () => {
       schemaVersion: 1,
       installations: [],
       eligibleUpdateNames: [],
+      scanIssues: [],
       scannedAt: 1
     })
     getWslHomeMock.mockReturnValue('\\\\wsl.localhost\\Ubuntu\\home\\alice')
--- a/src/main/skills/skill-freshness-inventory.test.ts
+++ b/src/main/skills/skill-freshness-inventory.test.ts
@@ -377,4 +377,40 @@ describe('read-only skill freshness inventory', () => {
     // command does not touch, so the limit is reported without blocking the update.
     expect(inventory.eligibleUpdateNames).toEqual(['orca-cli'])
   })
+
+  it('reports incomplete plugin coverage without inventing per-skill installations', async () => {
+    const test = await fixture()
+    await test.writeSkill(join(test.homeDir, '.agents', 'skills'), test.currentMarkdown)
+    const pluginCache = join(test.homeDir, '.codex', 'plugins', 'cache')
+    await mkdir(join(pluginCache, ...Array.from({ length: 11 }, (_, index) => `level-${index}`)), {
+      recursive: true
+    })
+
+    const inventory = await inventorySkillFreshness({
+      currentAppVersion: '2.0.0',
+      homeDir: test.homeDir,
+      repos: [],
+      resourceRoot: test.resourceRoot
+    })
+
+    expect(inventory.installations).toHaveLength(1)
+    expect(inventory.installations[0]).toMatchObject({
+      name: 'orca-cli',
+      status: 'current',
+      topology: 'canonical-copy'
+    })
+    expect(inventory.installations).not.toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({ errorCategory: 'plugin-cache-scan-incomplete' })
+      ])
+    )
+    expect(inventory.scanIssues).toEqual([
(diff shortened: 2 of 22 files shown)


5Hyeons:
Superseded by the updated #10865 branch; the repair commits are now included there.

PR #10885, opened 2026-07-27
fix(hooks): drain POSIX hook stdin without PATH

nolainjin:
## Summary

Codex/Grok/Claude POSIX managed hooks no longer depend on `PATH` just to read or drain hook stdin.

The shared hook prelude used `cat` in two places:

- `payload=$(cat)` in every generated POSIX hook script
- `cat >/dev/null 2>&1 || :` in missing-script/no-op launchers

When a tool launches a Codex worker with a stripped or minimal `PATH`, those reads can fail before stdin is consumed. That turns an otherwise harmless no-op into hook exit noise or a broken pipe while Codex is still writing the hook payload.

This changes the shared reader to prefer `/bin/cat` before falling back to `cat`, so normal macOS/Linux/WSL shells consume stdin even when `PATH` is empty.

## Validation

- `corepack pnpm exec vitest run --config config/vitest.config.ts src/main/agent-hooks/managed-hook-stdin-lifecycle.test.ts`
- `corepack pnpm exec oxlint src/main/agent-hooks/hook-stdin-contract.ts src/main/agent-hooks/managed-hook-stdin-lifecycle.test.ts`
- `corepack pnpm run typecheck:node`

Related: #8110

### DIFF
--- a/src/main/agent-hooks/hook-stdin-contract.ts
+++ b/src/main/agent-hooks/hook-stdin-contract.ts
@@ -1,6 +1,10 @@
 export type PosixHookEmptyPayloadPolicy = 'exit' | 'empty-object'
 
-export const POSIX_HOOK_STDIN_READER = '{ /bin/cat 2>/dev/null || cat; }'
+// Why: a stripped PATH must not stop a hook from consuming stdin, or the agent
+// sees exit 127 and a broken pipe mid-write (#8110). `command -p` resolves from
+// the shell's built-in default PATH, so it also survives hosts without /bin/cat
+// (NixOS) and ignores a worktree-local `cat` that could capture the payload.
+export const POSIX_HOOK_STDIN_READER = '{ command -p cat 2>/dev/null || cat; }'
 export const POSIX_HOOK_STDIN_DRAIN_COMMAND = `${POSIX_HOOK_STDIN_READER} >/dev/null 2>&1 || :`
 
 // Why: every POSIX hook must own stdin before any no-op exit; sharing this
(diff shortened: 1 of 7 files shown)


OrcaWin:
Reviewed with an independent reproduction on both macOS and Ubuntu 20.04, driving the actual generated `codex-hook.sh` rather than a unit test alone.

**Confirmed.** Under a stripped PATH the old `payload=$(cat)` fails before consuming stdin, emitting `cat: No such file or directory` and giving the writer EPIPE — the `broken pipe (os error 32)` symptom in #8110. Your new test genuinely fails on revert, so it is a real regression guard.

**One blocking defect, fixed and pushed.** The change updated the shared constants but left five other test files asserting the literal old `cat` strings, breaking 10 pre-existing tests. CI did not catch this — the only check that ran was `track-community-pr` (7s); the vitest suite never ran. Those five now derive their assertions from the exported constants, so the contract has one source of truth.

**One improvement, also pushed.** I changed the reader to `command -p cat` instead of `/bin/cat`. `command -p` resolves from the shell's built-in default PATH, so it additionally survives hosts with no `/bin/cat` (NixOS) and ignores a worktree-local `cat` that could otherwise capture the payload. Verified directly: old form exits 127 under `PATH=""`, new form reads the payload and exits 0.

I considered a pure-shell read loop, which needs no external binary, and rejected it with measurements: ~13x slower (272ms vs 21ms at 1MB) on a per-tool-call hot path, and it corrupts payloads — adds a trailing newline and mangles backslashes, disqualifying for JSON. The chosen shape also mirrors the file's own existing `WINDOWS_HOOK_STDIN_READER` precedent.

**Honest gap worth stating:** this stops the noise and the broken pipe, but does not restore *delivery* under a stripped PATH — `curl` is also unqualified, so the hook exits 0 having sent nothing. It converts a loud failure into a silent safe one. That matches your PR description, but it means #8110 is not fully closed.

Validation: 501 tests passing across the hook trees, typecheck exit 0.

PR #10891, opened 2026-07-27
perf(store): index worktrees and tabs once per hydration loop

nwparker:
## Description

Four sites in `store/slices/terminals.ts` rebuilt a flattened array **inside a loop** and linearly searched it:

```ts
const worktree = Object.values(get().worktreesByRepo).flat().find((e) => e.id === worktreeId)
const tab      = Object.values(tabsByWorktree).flat().find((e) => e.id === tabId)
```

That is O(rows × ids) for O(rows + ids) distinct work. All four now build one first-wins index per loop, extracted into `worktree-by-id-index.ts`.

**Why hoisting is safe here:** neither loop calls `set()` nor `await`s anything, so the store snapshot is stable across iterations — I checked before hoisting rather than after. The one `set()` in `reconnectPersistedTerminals` lands after both loops, which is why a single index can serve them.

This runs synchronously on renderer cold start and gates `workspaceSessionReady`, which blocks terminal pane mounting.

## ELI5

To restore your terminals, the app looks up each saved worktree by id. It was rebuilding the *entire* list from scratch for every single lookup — like re-reading the whole phone book from page one for each name you want.

Now it builds the index once and looks each name up directly. Same answers, roughly 50× less work on a big session.

## Proof of perf improvement

`config/scripts/hydrate-worktree-lookup-benchmark.mjs` (new), arms alternated per round, per-arm medians, two runs:

| repos | worktrees | ids | flatten | indexed | speedup |
|---|---|---|---|---|---|
| 1 | 5 | 5 | 0.0015 ms | 0.0007 ms | **2.1×** |
| 3 | 60 | 20 | 0.0123 ms | 0.0021 ms | **5.9×** |
| 10 | 420 | 188 | 0.6162 ms | 0.0113 ms | **54.5×** |
| 10 | 1000 | 400 | 3.1981 ms | 0.0284 ms | **112.6×** |

Sizes come from a real `orca-data.json` (10 repos / 423 worktrees / 188 pending reconnect), not invented numbers.

Scope, stated plainly:

- **A one- or two-repo session sees almost nothing.** The win scales with worktrees × pending ids. I'm not claiming a universal speedup.
- **This times one of the four sites.** Including the shared normalize/sanitize work around the layout map, the end-to-end block is ~28×, not 54× — the smaller number is the honest end-to-end one.
- The fixture includes ids that match **nothing** (SSH worktrees are absent from `worktreesByRepo` at cold start — the code comments call this out), since those are the worst case for the linear arm and do occur in practice.

## Proof of zero regression

**First-wins is the correctness crux.** `Array.prototype.find` returns the *first* match, so an index that overwrote on collision would resolve a **different repo** for a duplicated worktree id. Both the tests and the benchmark fixture carry a deliberate cross-repo duplicate so that difference is observable.

That mattered concretely: my first benchmark fixture had no duplicates, so a last-wins mutant **passed the equality guard**. I added the collision and the guard now catches it. Without that, the benchmark would have happily timed a semantically broken index.

- **New tests**: `worktree-by-id-index.test.ts`, 8 tests, asserting the index resolves *identically to flatten-and-find* across hits, misses, empty maps, empty repos, and cross-repo duplicates.
- **Mutation testing** (`--no-cache`), 4/4 killed:

  | # | Mutant | Result |
  |---|---|---|
  | 1 | Worktree index last-wins instead of first-wins | killed |
  | 2 | `buildByIdIndex` last-wins | killed |
  | 3 | Only the first repo indexed | killed |
  | 4 | Skip the first worktree of each repo | killed |

- **Benchmark guards verified by breaking them**: a nulled index throws `resolution differs`; a last-wins index throws at the first multi-repo row.
- `src/renderer/src/store/`: **133 files, 2,031 tests** passing. `pnpm typecheck` clean.
- `terminals.ts` is **exactly the same line count as on `main`** — it's on the `max-lines` baseline, so the helper went in its own module rather than growing the file. No suppression, no baseline bump.

2 loops. Round 1 asked whether hoisting across two loops was safe — verified no `set()`/`await` between them. Round 2 found the vacuous benchmark fixture described above.

**Verdict: ship.** Behaviour is pinned against the exact `.find()` it replaces, the first-wins semantic is enforced by both tests and benchmark, and all 4 mutants die.

### DIFF
--- a/config/scripts/hydrate-worktree-lookup-benchmark.mjs
+++ b/config/scripts/hydrate-worktree-lookup-benchmark.mjs
@@ -141,8 +141,10 @@ console.log(
   `${pad('repos', 6)} ${pad('worktrees', 10)} ${pad('ids', 5)} ${pad('flatten', 11)} ${pad('indexed', 11)} ${pad('speedup', 9)}`
 )
 
-// Sizes from a real orca-data.json on a heavy machine: 10 repos / 423 worktrees, with
-// 188 pending reconnect. The small rows are an ordinary one- or two-repo session.
+// Row shapes are synthetic, sized against a real orca-data.json on a heavy machine
+// (10 repos / 423 worktrees / 188 pending reconnect). They are not that dataset: the
+// generator spreads worktrees evenly and injects one duplicate id, so treat the counts
+// as "about this scale", not a replay.
 for (const [repoCount, worktreesPerRepo, idCount] of [
   [1, 5, 5],
   [3, 20, 20],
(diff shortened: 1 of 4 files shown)


nwparker:
Fixed. The table now counts the generated map (61 / 421 / 1001) instead of multiplying, which was silently dropping the duplicate id `makeStore` injects — and that duplicate is precisely what makes the first-wins semantic observable, so under-reporting it was worse than cosmetic.

Also took your second point: the header and footer now say the fixtures are **synthetic at real-world scale**, not a replay of the 423-worktree session. They're sized against a real `orca-data.json` but the generator spreads worktrees evenly across repos, so calling it that dataset overstated it.

Ratios unchanged: 1.7× / 5.8× / 54.8× / 110.6×.

PR #10892, opened 2026-07-27
perf(usage): yield with setImmediate, not a clamped setTimeout(0)

nwparker:
## Description

The Claude usage scanner yielded to the event loop with `setTimeout(resolve, 0)`. Node clamps that to **~1ms**. Both batch loops yield once per `FILE_SCAN_BATCH_SIZE` (4) files, so a machine with 7,552 transcripts makes ~3,800 yields and spends **seconds** parked on timers doing no work.

`setImmediate` runs on the same tick's check phase with no clamp. The sibling scanner `src/main/codex-usage/scanner.ts:92` already used `setImmediate` — this was an inconsistency, not a design choice. The two functions were otherwise byte-identical.

One line changed.

## ELI5

While scanning your usage history, the app pauses between small batches so the rest of the app stays responsive.

It was using a pause that Node rounds **up to a full millisecond**. Do that ~3,800 times and you've waited ~4 seconds doing nothing. The fix uses the pause that means "right after this tick," which is what was wanted all along.

## Proof of perf improvement

`config/scripts/claude-usage-yield-benchmark.mjs` (new), arms alternated per round, per-arm medians:

| yields | `setTimeout(0)` | `setImmediate` | ratio | **saved** |
|---|---|---|---|---|
| 100 | 139.7 ms | 20.7 ms | 6.7× | 119 ms |
| 500 | 611.9 ms | 25.1 ms | 24.4× | 587 ms |
| 3,780 *(this machine's real scan)* | 4291–5270 ms | 54–164 ms | 32–81× | **4237–5107 ms** |

**Read the saved column, not the ratio.** The `setImmediate` arm is small enough that background
load lands almost entirely on it, swinging the ratio between 32× and 81× across three consecutive
runs while the removed wall time holds steady at 4.2–5.1 s. The honest headline is
**~4.3 seconds of main-process wall time removed per full scan**, not a multiple.

The yield count is **derived from real data**, not assumed: the benchmark counts `.jsonl` files under `~/.claude/projects` (7,552 here) and reads `FILE_SCAN_BATCH_SIZE` and the yield-site count out of the scanner source.

Independent confirmation of the mechanism, measured separately and in both orders: `setTimeout(0)` costs **1,138 µs** per yield versus **14.3 µs** for `setImmediate`.

**Responsiveness improves — it does not regress.** That matters, because responsiveness is the entire reason the yield exists; a "faster" yield that starved other work would be a regression even at 80×. Measured with a concurrent probe task during a full scan:

| | scan | worst concurrent wait |
|---|---|---|
| `setTimeout(0)` | 3,790 ms | 6.33 ms |
| `setImmediate` | 56 ms | **0.12 ms** |

The old version was worse on both axes because most of its wall time was the main process idle-but-blocked on timer clamps. The benchmark prints an explicit NOTE if `setImmediate` ever shows a worse concurrent wait, so a future regression here is visible rather than silent.

Scope: this is **wall-clock the process spent waiting**, not CPU work removed. The scan's own I/O and parsing are unchanged. It's paid on every Usage-pane scan and on every forced automation rescan.

## Proof of zero regression

- `src/main/claude-usage/`, `codex-usage/`, `opencode-usage/`: **11 files, 77 tests** passing. `pnpm typecheck` clean.
- **Benchmark drift guards, verified by breaking them**: renaming `FILE_SCAN_BATCH_SIZE` throws `…is stale`; reverting the scanner to `setTimeout` throws `scanner no longer yields with setImmediate`.

  The second guard initially **did not fire** — I'd written a bare `includes('setImmediate')` check, which the new *comment* mentioning `setImmediate` satisfied even after the call reverted. Now it matches the call form (`/setImmediate\(resolve\)/`). A guard that a comment can satisfy is not a guard.

- No behaviour depends on the yield's *duration*: both call sites are `if (moreWork) await yieldToEventLoop()` between pure in-memory batch steps.

**Deliberately left alone** — three other `setTimeout(resolve, 0)` sites exist in `orca-runtime.ts:14222/14249` and `pty.ts:5044`, and they look like the same bug but are not. They sit between PTY **paste chunk writes**, where the clamp acts as inter-chunk *pacing*, not merely a yield. Removing it there would change how fast bytes hit the terminal — a behaviour change, not a perf fix. `opencode-usage/scanner.ts:148` has the identical yield shape but iterates databases (2 on a real machine, ~1 yield), so it's below the bar; noted, not changed.

2 loops. Round 1 verified the clamp claim independently and confirmed the yield frequency against the real transcript count rather than trusting the reported figure. Round 2 caught the comment-satisfiable drift guard.

**Verdict: ship.** One line, an in-repo sibling already proves the pattern, and the property that justifies the yield's existence measurably improves.

### DIFF
--- a/config/scripts/claude-usage-yield-benchmark.mjs
+++ b/config/scripts/claude-usage-yield-benchmark.mjs
@@ -19,7 +19,7 @@ import { performance } from 'node:perf_hooks'
 import { readdirSync, statSync } from 'node:fs'
 
 const REPO_ROOT = new URL('../..', import.meta.url)
-const ROUNDS = 6
+const ROUNDS = Number(process.env.ORCA_YIELD_BENCH_ROUNDS ?? '10')
 
 // Why re-read the source: the claim is that the scanner yields once per batch across
 // two loops. If the batch size or the yield sites change, these numbers stop meaning
@@ -73,13 +73,12 @@ const transcriptCount = (() => {
   }
 })()
 
-// Both passes yield once per batch, minus the final batch in each.
 const FALLBACK_TRANSCRIPTS = 7500
 const effectiveTranscripts = transcriptCount > 0 ? transcriptCount : FALLBACK_TRANSCRIPTS
-const YIELDS_PER_SCAN = Math.max(
-  1,
-  Math.floor(effectiveTranscripts / FILE_SCAN_BATCH_SIZE) * YIELD_SITES
-)
+// Each pass walks ceil(files / batch) batches and yields after every batch except the
+// last, so a pass yields batchesPerPass - 1 times.
+const BATCHES_PER_PASS = Math.max(1, Math.ceil(effectiveTranscripts / FILE_SCAN_BATCH_SIZE))
+const YIELDS_PER_SCAN = Math.max(1, (BATCHES_PER_PASS - 1) * YIELD_SITES)
 
 const yieldWithTimeout = () => new Promise((resolve) => setTimeout(resolve, 0))
 const yieldWithImmediate = () => new Promise((resolve) => setImmediate(resolve))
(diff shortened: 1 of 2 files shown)


nwparker:
Fixed in e32d931, and it forced a **correction to this PR's headline number**.

**The yield count was wrong twice over**, exactly as you describe. Each pass walks `ceil(files / batch)` batches and yields after every batch *except the last*, so a pass yields `batchesPerPass - 1` times — and I was separately handing that yield count to the loop *as a batch count*, producing one fewer yield again. The "full scan" row measured neither the count it printed nor the count the scanner performs.

**Correcting it exposed something worse than an off-by-one: the ratio isn't a publishable number at all.** Three consecutive runs of the corrected benchmark:

| run | `setTimeout(0)` | `setImmediate` | ratio | saved |
|---|---|---|---|---|
| 1 | 5270.5 ms | 163.7 ms | 32.2× | **5107 ms** |
| 2 | 4412.5 ms | 54.7 ms | 80.6× | **4358 ms** |
| 3 | 4291.5 ms | 54.0 ms | 79.5× | **4237 ms** |

The `setImmediate` arm is small enough that any background load lands almost entirely on it, swinging the ratio 32×–81× while the **removed wall time stays at 4.2–5.1 s**. So I've retracted the "82×" headline: the honest claim is **~4.3 seconds of main-process wall time removed per full scan**, and the benchmark now prints a `saved` column with a note telling the reader to use it rather than the ratio.

Worth being explicit that this is the same class of mistake I hit on #10881 (a 23.3× that was really 6.7×) — there I fixed it by interleaving the arms, which this benchmark already did. Interleaving controls for drift *between* arms; it does nothing when one arm is so small that noise dominates it proportionally. Reporting the absolute delta is the actual fix.

The underlying change is unaffected — one line, `setTimeout(resolve, 0)` → `setImmediate(resolve)`, matching the sibling `codex-usage` scanner — and the responsiveness measurement still shows the concurrent-wait improving rather than regressing.

PR #10895, opened 2026-07-27
perf(git): overlap getBranchCompare's head-of-chain reads

nwparker:
## Description

`getBranchCompare` ran four git spawns **strictly in series** before any compare work started:

1. `branch --show-current` (compareRef)
2. the base-ref probe — `rev-parse --verify --quiet <ref>^{commit}`
3. `rev-parse HEAD`
4. `rev-parse <resolvedBaseRef>`

Three of those are independent: `compareRef` is display-only metadata (`summary.compareRef`) that feeds nothing downstream, and HEAD's oid doesn't depend on the base ref. They now run concurrently.

The fourth was **redundant outright**. The probe already prints the commit oid for the ref it proves, and threw it away — then a second spawn re-resolved the same value. `resolveWorktreeBaseCommitOid` now returns that oid so it can be reused; `hasWorktreeBaseCommitRef` delegates to it, so its **other 4 callers are untouched**.

## ELI5

Before showing "your branch vs. main," the app asked git four questions **one at a time**, each waiting for the last — even though three don't depend on each other. And the fourth question had already been answered by the second one; the app just forgot the answer and asked again.

Now it asks the three independent questions at once and remembers the answer it already had. Same result, about 3.6× faster.

## Proof of perf improvement

`config/scripts/branch-compare-head-benchmark.mjs` (new). Spawns the **real git binary** against this repo — process-launch cost is the whole point here, so modelling it would be worthless. Arms alternate which one leads; per-arm medians; resolved values compared for equality before timing.

| base ref | serial | concurrent | speedup |
|---|---|---|---|
| `origin/main` | 192.0 ms | 51.9 ms | **3.70×** |
| `refs/remotes/origin/main` | 144.2 ms | 99.8 ms | **1.44×** |
| `main` | 190.4 ms | 52.6 ms | **3.62×** |

Stable across three runs (3.63/3.63/3.70×). The `refs/...` row is lower **by design** — an already-qualified ref skips the probe, so only the concurrency half applies and there's no oid to reuse.

Scope, stated plainly:

- This times the **head-of-chain reads only**. The `diff`, `merge-base`, and `rev-list` that follow are unchanged, so the end-to-end `getBranchCompare` improvement is smaller than 3.6×.
- **Over SSH this saves host-local spawns inside the relay, not a network round trip.** The RPC count is unchanged. Worth being explicit, since "3.6× faster" would otherwise read as a latency win over the wire.
- Folder workspaces are unaffected — `runBranchCompare` returns early on `isFolder`.

Frequency: this is the highest-frequency non-`getStatus` git path. It runs on a 30s visible-window interval whenever Source Control is open, plus on every HEAD change, every remote ahead/behind change, and after every commit/push/pull/sync/fetch.

## Proof of zero regression

**The error paths are the risk here, and they're intricate**: an unborn HEAD must still report `unborn-head` (not `invalid-base`), and an unborn branch *with* a resolvable base must still report `ready`. Parallelizing means HEAD's rejection is now captured rather than thrown, so the ordering of those branches had to be preserved deliberately.

- **Six existing tests had to be rewritten**, and that's the honest part of this change. They used positional `mockResolvedValueOnce` chains, which encode *call order* rather than behaviour — so they broke the moment the reads overlapped, even though nothing observable changed. They now dispatch on the actual git args via a shared `mockBranchCompareGit` helper, which is both order-independent and far clearer about intent.
- **Three new tests** for the oid-reuse risks specifically: an empty `rev-parse --quiet` result must count as unresolved; only the ref the probe *actually resolved* may supply the oid; an already-qualified base must still resolve via `rev-parse` with no probe.
- **Mutation testing** (`--no-cache`). First pass: **3 of 4 mutants survived**. Adding the three tests above killed the empty-probe one. The other two I then **proved equivalent** rather than papering over with more tests:

  | # | Mutant | Result |
  |---|---|---|
  | 1 | Probe returns `''` instead of `null` for a missing ref | killed |
  | 2 | Reuse *any* probed oid regardless of ref | **equivalent** |
  | 3 | Record failed probes too | **equivalent** |
  | 4 | Unborn HEAD reported as resolved | killed |
  | 5 | Drop oid reuse entirely | killed (6 tests) |

  Why 2 and 3 are equivalent: `resolveWorktreeAddBaseRef` returns at its **first** successful candidate, so at most one entry is ever recorded, and a failed candidate's key is never the one read back. I confirmed this empirically rather than by argument — instrumenting the map to throw if it ever exceeded one entry passes all 86 tests. The keyed map is therefore defensive, not load-bearing, and the comment now says exactly that.

- `src/main/git/` + the probe's other callers: **39 files, 610 tests** passing. `pnpm typecheck` clean.
- **Git version**: none adopted. `branch --show-current` (2.22), `rev-parse --verify --quiet`, and `--end-of-options` all predate the 2.25 baseline. This change *removes* a spawn rather than reaching for a newer option.

3 loops. Round 1 replaced a mutable captured `probedBaseOid` with a ref-keyed map after noticing a failed first candidate could leave a stale value. Round 2 was the mutation run that found 3 survivors. Round 3 proved two of them equivalent and simplified the comment to match what the code actually relies on.

**Verdict: ship.** Error-path behaviour is pinned by order-independent tests, every non-equivalent mutant dies, and the two survivors are demonstrably unobservable.

### DIFF
--- a/config/scripts/branch-compare-head-benchmark.mjs
+++ b/config/scripts/branch-compare-head-benchmark.mjs
@@ -1,0 +1,183 @@
+#!/usr/bin/env node
+// Benchmark: the head-of-chain reads in getBranchCompare (src/main/git/status.ts).
+//
+// Four spawns ran strictly in series before any compare work started: branch
+// --show-current, the base-ref probe, rev-parse HEAD, and rev-parse <base>. compareRef is
+// display-only metadata and HEAD's oid does not depend on the base ref, so three of them
+// are independent. The fourth was redundant outright: the probe already prints the commit
+// oid for the ref it proves, and that value was discarded and re-resolved.
+//
+// This spawns the real git binary against this repo, so it measures actual process-launch
+// cost rather than a model of it. Over SSH these are host-local spawns inside the relay,
+// so the saving applies to remote spawn time, not to network round trips.
+//
+// Both arms are compared for identical resolved values before timing.
+//
+// Run with:  node config/scripts/branch-compare-head-benchmark.mjs
+import { execFile } from 'node:child_process'
+import { readFileSync } from 'node:fs'
+import { performance } from 'node:perf_hooks'
+import { fileURLToPath } from 'node:url'
+
+const REPO_ROOT = fileURLToPath(new URL('../..', import.meta.url))
+const ITERATIONS = Number(process.env.ORCA_BRANCH_COMPARE_BENCH_ITERATIONS ?? '8')
+const WARMUP = Number(process.env.ORCA_BRANCH_COMPARE_BENCH_WARMUP ?? '2')
+const ROUNDS = 6
+
+for (const [name, value] of [
+  ['ORCA_BRANCH_COMPARE_BENCH_ITERATIONS', ITERATIONS],
+  ['ORCA_BRANCH_COMPARE_BENCH_WARMUP', WARMUP]
+]) {
+  if (!Number.isSafeInteger(value) || value <= 0) {
+    throw new Error(`${name} must be a positive integer, received ${value}`)
+  }
+}
+
+// Why re-read the source: this benchmark's claim is that the four reads now run
+// concurrently and the probe's oid is reused. If either reverts, the numbers stop
+// meaning what the header says.
+const STATUS_SOURCE = readFileSync(new URL('../../src/main/git/status.ts', import.meta.url), 'utf8')
+for (const marker of ['resolveWorktreeBaseCommitOid', 'probedOidByRef']) {
+  if (!STATUS_SOURCE.includes(marker)) {
+    throw new Error(`status.ts no longer contains ${marker}; this benchmark is stale`)
+  }
+}
+
+function git(args) {
+  return new Promise((resolve, reject) => {
+    execFile('git', args, { cwd: REPO_ROOT, maxBuffer: 64 * 1024 * 1024 }, (error, stdout) =>
+      error ? reject(error) : resolve(stdout.trim())
+    )
+  })
+}
+
+async function probeOid(qualifiedRef) {
+  try {
+    const out = await git(['rev-parse', '--verify', '--quiet', `${qualifiedRef}^{commit}`])
+    return out.length > 0 ? out : null
+  } catch {
+    return null
+  }
+}
+
+// Pre-fix: serial chain, and the probe's oid discarded then re-resolved.
+async function readSerial(baseRef) {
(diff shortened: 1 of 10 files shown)


nwparker:
Investigated this properly rather than either accepting or dismissing it, and the answer splits in two directions.

**Your version number is right; the stated consequence isn't.**

`rev-parse` has its own hand-rolled argument loop and did **not** get `--end-of-options` until [`3a1f91cfd9`](https://github.com/git/git/commit/3a1f91cfd9633a4f59e0534fa5ba076e031a78ed), first released in **v2.30.0** — separate from the general parse-options support in `19e8789b23` (v2.24.0) that covers `git show` / `log` / `rev-list`. So 2.30 is correct for `rev-parse` specifically.

But **it does not error on older git.** `git rev-parse` silently ignores unrecognized `--`-prefixed arguments. Verified against real binaries (Ubuntu 20.04 → 2.25.1, alpine 2.24.4 / 2.26.3 / 2.30.6, plus 2.38.1 and 2.49.1):

```
git version 2.25.1
$ git rev-parse --verify --end-of-options HEAD
cef26deebea70eff855290203543c2962159414d      exit=0
```

Control: `--bogus-flag` behaves identically. So there's no unsupported-option failure to catch and no `GitCapabilityCache` fallback warranted. What actually differs is whether the *barrier works* — on 2.25–2.29 it's accepted, ignored, and provides zero protection:

```
# 2.25.1 — barrier inert, hostile option still executes
$ git rev-parse --verify --end-of-options --local-env-vars
fatal: Needed a single revision
GIT_ALTERNATE_OBJECT_DIRECTORIES ... (full env printed)

# 2.30.6 — barrier active
$ git rev-parse --verify --end-of-options --local-env-vars
fatal: Needed a single revision
```

That's worth knowing, but it's pre-existing on `main` (this PR *removes* one such call) and harmless in practice: every `rev-parse` call site either pre-validates its input (`assertFullGitObjectId`, or `ref.startsWith('-')` guards) or catches, and `--verify` fails closed on its own.

**Chasing it down did surface a real bug** — a worse one, because it's silent and affects current git too. `--verify` swallows the marker, but `--symbolic-full-name` deliberately echoes it, so `git-history.ts` was reading the literal string `"--end-of-options"` as the ref name and categorizing **every branch and tag in git history as a plain commit**. Fixed separately in #10906, with the git behaviour pinned in the real-binary compatibility suite.

So: no change needed here, but thank you — the flag was worth following even though the specific failure mode didn't hold.

PR #10900, opened 2026-07-27
perf(relay): keep the PTY replay window as chunks, not a re-sliced string

nwparker:
## Description

`appendReplayBuffer` in the relay did:

```ts
managed.buffered += data
if (managed.buffered.length > REPLAY_BUFFER_MAX) {
  managed.buffered = managed.buffered.slice(-REPLAY_BUFFER_MAX)
}
```

It runs on **every raw node-pty emission** — called from `emitIngressData`, *before* batching, so it's per chunk, not per flush. Once a PTY has produced 100 KB (which a long-lived shell does within seconds), the `slice(-100KB)` flattens and copies the **entire window on every subsequent chunk, forever**. A long-lived PTY spends essentially all its life in that regime.

The fix reuses `RecentPtyOutputBuffer`, which already solved exactly this shape in the main process: keep chunks, drop from the head, defer the join to `read()`. The relay's `buffered` has exactly **three readers** (attach, adopt, revive), so deferring the join is free.

This matters because **the relay runs on the user's SSH host**, and this is per-PTY.

## ELI5

To let you reattach to a terminal, the relay keeps the last 100 KB of output.

It kept that as one big string, and every time a new sliver of output arrived it rebuilt the whole 100 KB from scratch — for a 64-byte keystroke echo. Now it keeps a list of pieces and drops old ones off the front, only gluing them together on the rare occasion someone actually reattaches.

## Proof of perf improvement

`config/scripts/relay-replay-buffer-benchmark.mjs` (new). Pre-saturated window (the regime that actually matters), arms alternated per round, per-arm medians, retained tail compared for equality before timing:

| workload | rolling string | chunk deque | speedup |
|---|---|---|---|
| interactive shell 64 B ×200/s | 0.590 ms | 0.009 ms | **66×** |
| agent TUI 512 B ×400/s | 1.052 ms | 0.011 ms | **92×** |
| build log 4 KiB ×256/s (1 MiB/s) | 0.748 ms | 0.005 ms | **161×** |
| dump 8 KiB ×512/s (4 MiB/s) | 1.651 ms | 0.008 ms | **205×** |
| firehose 16 KiB ×1024/s (16 MiB/s) | 3.457 ms | 0.018 ms | **197×** |

Scope, stated plainly:

- Absolute saving is **0.6–3.5 ms per second per PTY**. That's not enormous alone — it matters because it's multiplied by every open remote pane and paid on the SSH host, which is often far weaker than the laptop.
- **Reads now pay the join instead.** Attach/adopt/revive were already O(window), so this moves rare work rather than adding it; the benchmark deliberately doesn't count that as a win.
- The benchmark re-reads `REPLAY_BUFFER_MAX` and the append form out of `pty-handler.ts` and throws `…is stale` if either drifts.

## Proof of zero regression

**The equivalence tests found a real bug in my own change, before it shipped.** Making `RecentPtyOutputBuffer`'s limit configurable (the relay retains 100 KB where the main process retains 64 KB) left **one arithmetic branch still using the hardcoded `RECENT_PTY_OUTPUT_LIMIT`**:

```ts
const excess = this.totalLen - RECENT_PTY_OUTPUT_LIMIT   // should be this.limit
```

The relay buffer silently under-retained — **100,800 of 102,400 code units**, dropping ~1.6 KB of replay off the front of every reattach. Typecheck was clean, all existing tests passed, and the benchmark would have reported a large speedup on a subtly wrong buffer. Only the byte-for-byte comparison against the old rolling string caught it.

- **New test file** `pty-replay-buffer-equivalence.test.ts`, 10 tests, comparing the deque's `read()` against the **exact pre-change rolling-string implementation**: below the cap, at saturation, single oversized append, oversized append after existing content, the cap boundary (−1/0/+1), empty appends, repeated reads, reads interleaved mid-stream, 40 seeded randomized chunk streams, and surrogate-pair splitting (the string form split pairs too — the deque must not "fix" that into a different tail).
- **3 new tests on the buffer itself** pinning the configured limit directly, so the bug above cannot recur silently.
- `src/relay/` + the buffer suite: **75 files, 919 tests**, with **2 pre-existing failures** in `agent-exec-handler.test.ts`. I verified those are unrelated by stashing my change and re-running — same 2 failures on the clean tree. They're also one of the two known-failing files on `main`.
- `pnpm typecheck` clean; oxlint clean.

Cross-boundary note: the relay already imports from `src/main/` in several places and esbuild bundles it into a self-contained CommonJS artifact, so sharing this class doesn't introduce a new dependency direction.

3 loops. Round 1 caught a doubled `managed.buffered.read()` call in the adopt path. Round 2 was the equivalence failure that exposed the hardcoded-limit bug. Round 3 confirmed the pre-existing relay failures were unrelated rather than assuming it.

**Verdict: ship.** The retained window is proven byte-identical to the implementation it replaces across every boundary case plus fuzzing, and the one real defect this change introduced was caught and fixed by that proof.

### DIFF
--- a/config/scripts/relay-replay-buffer-benchmark.mjs
+++ b/config/scripts/relay-replay-buffer-benchmark.mjs
@@ -1,0 +1,212 @@
+#!/usr/bin/env node
+// Benchmark: the relay's per-PTY-chunk replay buffer append (src/relay/pty-handler.ts).
+//
+// appendReplayBuffer did `buffered += data` then, over the cap, `buffered.slice(-CAP)`.
+// Once a PTY has produced CAP bytes -- which a long-lived shell does almost immediately
+// -- every subsequent chunk flattened and copied the whole 100 KB window. The append is
+// called per raw node-pty emission, before batching, so it is per chunk, not per flush.
+//
+// The fix reuses RecentPtyOutputBuffer: keep chunks, drop from the head, and defer the
+// join to read(), which only attach/adopt/revive call.
+//
+// Both arms are compared for an identical retained tail before timing.
+//
+// Run with:  node config/scripts/relay-replay-buffer-benchmark.mjs
+import { readFileSync } from 'node:fs'
+import { performance } from 'node:perf_hooks'
+
+const ROUNDS = 6
+const SECONDS = Number(process.env.ORCA_REPLAY_BENCH_SECONDS ?? '1')
+if (!Number.isFinite(SECONDS) || SECONDS <= 0) {
+  throw new Error(`ORCA_REPLAY_BENCH_SECONDS must be positive, received ${SECONDS}`)
+}
+
+// Why re-read the sources: the claim is that the relay now appends into a chunk deque
+// with the relay's own cap. If either reverts, these numbers stop meaning what they say.
+const HANDLER_SOURCE = readFileSync(
+  new URL('../../src/relay/pty-handler.ts', import.meta.url),
+  'utf8'
+)
+if (!/managed\.buffered\.append\(/.test(HANDLER_SOURCE)) {
+  throw new Error('relay no longer appends into a chunk deque; this benchmark is stale')
+}
+const capMatch = HANDLER_SOURCE.match(/REPLAY_BUFFER_MAX = ([\d *]+)/)
+if (!capMatch) {
+  throw new Error('REPLAY_BUFFER_MAX not found; this benchmark is stale')
+}
+// eslint-disable-next-line no-eval -- reading a numeric literal expression out of source
+const REPLAY_BUFFER_MAX = Number(eval(capMatch[1]))
+
+// Pre-fix: rolling string, re-sliced once over the cap.
+function appendString(state, data) {
+  if (data.length === 0) {
+    return state
+  }
+  const next = state + data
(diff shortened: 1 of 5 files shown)


nwparker:
Both fixed in 39a78f5.

**The surrogate test was worse than incomplete — it never split a pair at all.** You're right about the mechanism, and working through it showed your suggested fix doesn't reach the case either: the cap is even and a pair is two code units, so `slice(-CAP)` over an emoji run always lands on a pair boundary, and prepending one unit shifts the *start* rather than the cut. What actually works is a trailing single unit **after** the run, which pushes the cut mid-pair and leaves a dangling low surrogate:

```
'\u{1F600}'.repeat(CAP) + 'z'  ->  tail begins 0xDE00 (dangling low)
```

The test now asserts that lead unit is in the 0xDC00–0xDFFF range, and I kept the boundary-aligned case as its own test asserting a *high* surrogate — so both sides of the boundary are pinned rather than assumed.

**`eval` is gone.** The regex already constrains the match to `[\d *]+`, so the literal is parsed as a product directly, with a sanity check that the result is a positive safe integer. Agreed the eslint suppression was doing nothing for Biome.

Re-verified after both changes: 11 equivalence tests pass, and the benchmark still reports 63×–223×.

PR #10920, opened 2026-07-27
fix(ai-vault): preserve agent metadata header on session row expansion

hwantage:
## Summary

This PR proposes a simple yet highly effective usability improvement for the AI Vault Agent Session History list in Orca.

When a user clicks an item in the Agent Session History list to expand and view recent conversation turns, the header metadata line containing the AI agent icon (e.g., Codex, Hermes, Claude), agent name, message count, timestamp, and model previously disappeared. This made it difficult for users to identify which AI agent handled the session while reading expanded details without collapsing the row.

### Key Changes

- **Persistent Agent Metadata Header**: Updated `VaultSessionRow` in `AiVaultSessionRow.tsx` so that `SessionMetadata` (rendering the agent icon, agent name, message count, subagent count, timestamp, and model label) remains continuously visible regardless of whether the row is collapsed or expanded (`detailsExpanded = true`).
- **Cleaned Up Layout Redundancy**: Removed duplicate worktree status line logic and unused imports (`SessionWorktreeLine`, `shouldShowAiVaultSessionWorktreeLine`), allowing `SessionMetadata` to cleanly handle worktree information in both collapsed and expanded states.
- **Dedicated Unit Tests**: Added unit tests in `AiVaultSessionRow.test.tsx` ensuring that agent header metadata is preserved and visible when a session row is expanded.

## Screenshots
|before|after|
|---|---|
|image|image|

## Testing

- [x] `pnpm lint` (0 errors, 0 warnings)
- [x] `pnpm typecheck` (Passed cleanly)
- [x] `pnpm test` (All 163 right-sidebar test files and 1,422 tests passed)
- [x] `pnpm build`
- [x] Verified unit test in `AiVaultSessionRow.test.tsx` covering both collapsed and expanded session states.

## AI Review Report

Ran code review using AI pair programmer with explicit focus on:

1. **Usability &amp; UX Polish**: Verified that expanding a session row maintains the agent identity header (icon, agent name, message count, time) directly below the title, eliminating the need to collapse rows to check which agent performed the work.
2. **Cross-Platform &amp; Windows Compatibility**: Verified that all components utilize platform-agnostic React flex/grid layouts and standard design tokens without OS-specific DOM assumptions.
3. **Codebase Conventions &amp; Code Hygiene**: Confirmed strict compliance with `AGENTS.md` and `STYLEGUIDE.md` rules. Removed unused imports and maintained 0 oxlint warnings/errors.

## Security Audit

- **No Security Risk**: Pure UI presentation layer change in `AiVaultSessionRow.tsx`. No IPC channels, process execution, or file system access paths are altered.

## Notes

- **Windows Impact &amp; Compatibility**: Evaluated layout rendering and responsiveness for Windows (`win32`). Because `SessionMetadata` uses standard Tailwind flexbox and grid styling with design tokens defined in `main.css`, the header line renders consistently across Windows, macOS, and Linux without layout shifts or text wrapping issues on high-DPI Windows displays.

### DIFF
--- a/src/renderer/src/components/right-sidebar/AiVaultSessionRow.test.tsx
+++ b/src/renderer/src/components/right-sidebar/AiVaultSessionRow.test.tsx
@@ -1,42 +1,14 @@
 // @vitest-environment happy-dom
 
-import { render, screen } from '@testing-library/react'
+import { cleanup, render, screen, within } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import { renderToStaticMarkup } from 'react-dom/server'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { TooltipProvider } from '@/components/ui/tooltip'
 import type { AiVaultSession } from '../../../../shared/ai-vault-types'
+import type { AiVaultSessionWorktreeInfo } from './ai-vault-session-worktree'
 import { VaultSessionRow } from './AiVaultSessionRow'
 
-beforeEach(() => {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- test-only window.api shim
-  ;(window as any).api = {
-    aiVault: {
-      listSubagentSessions: vi.fn(),
-      getFirstUserPrompt: vi.fn()
-    },
-    shell: { openFilePath: vi.fn() }
-  }
-})
-
-const dummySession: AiVaultSession = {
-  id: 'session-123',
-  sessionId: 'session-123',
-  agent: 'codex',
-  title: 'Fix AI Vault keep header on expand',
-  updatedAt: '2026-07-27T10:00:00.000Z',
-  modifiedAt: '2026-07-27T10:00:00.000Z',
-  messageCount: 5,
-  subagentTranscriptCount: 0,
-  executionHostId: 'local',
-  filePath: '/path/to/session.json',
-  previewMessages: [
-    { role: 'user', text: 'Hello AI' },
-    { role: 'assistant', text: 'Hello User' }
-  ]
-}
-
-const testingLibrarySession = {
+const session = {
   id: 'local:gemini:sess-1:/home/a/.gemini/s.json',
   executionHostId: 'local',
   agent: 'gemini',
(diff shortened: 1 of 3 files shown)


brennanb2025:
Thanks for this — the diagnosis and the fix are both right. The metadata line is row identity rather than collapsed-preview chrome, so lifting it out of the `!detailsExpanded` branch is the right shape, and deleting the standalone `SessionWorktreeLine` block is required rather than optional, since `SessionMetadata` already renders that line under the same predicate.

I pushed one commit to your branch rather than send it back and forth. It's tests, plus two lines in `ai-vault-session-row-display.tsx`:

- **The new expanded-state assertion passed with the fix reverted.** `expect(markup).toContain('Codex')` was matching the details-toggle button's `aria-label="Codex session details"`, not the metadata line. The identity assertions are now scoped inside the metadata grid. (`toContain('5 msgs')` was genuinely load-bearing — that one was fine.)
- **Consolidated onto the harness already in the file.** The `renderToStaticMarkup` path brought a second session fixture and a second copy of the prop list alongside the existing Testing Library `renderRow`; both suites now share one fixture and one render.
- **Added `cleanup()`.** This suite doesn't run with `globals: true`, so Testing Library's auto-cleanup never fires — rendered rows were leaking between tests and would have made queries ambiguous as more got added.
- **Added a guard that the worktree badge renders exactly once when expanded.** That's the invariant that makes removing the old block safe; it goes red if the block comes back.
- **Restored the three comments dropped in `chore(tests): remove redundant comment`.** They explain why the Radix menu click doesn't expand the row and why the title is queried the way it is — the suggestion you were acting on was aimed at a different comment.
- Tagged the metadata grid with `data-testid="ai-vault-session-metadata"`, matching the row's other query anchors, and dropped the now-unused `export` on `SessionWorktreeLine` since this PR removes its only importer.

Checked in the app: expanded rows now keep the agent icon, name, message and subagent counts, timestamp and model, and the worktree badge still shows once.

brennanb2025:
Thanks for the PR!

PR #10944, opened 2026-07-27
fix(i18n/ko): translate the Editor Font Family settings strings

SonYoonho:
## Summary

- translate the 5 Editor Font Family strings that still render in English with the interface language set to Korean: the setting title, description, and input placeholder, plus the two matching settings-search entries
- add the values to `config/scripts/locale-ko-key-overrides.json` and mirror them into `src/renderer/src/i18n/locales/ko.json`

## Why

These keys were already in `ko.json` holding the English string verbatim. The sync tooling backfills missing keys with the English value, so they satisfied catalog parity while still showing English in the UI.

Wording follows the existing catalog rather than fresh translation. `Font` is rendered as `글꼴`, matching the 32 existing uses in `ko.json` (`IDE 글꼴`, `기본 글꼴`) rather than the single `서체`. `terminal` stays in English per `NEVER_TRANSLATE_VALUES` in `locale-translation-policy.mjs`, matching entries such as `새 terminal에서 {{value0}}을(를) 시작할 수 없습니다.`. The description ends in `~합니다`, matching the surrounding rows.

The strings exist twice, once for the pane and once for the settings-search index. Both are translated so a Korean search result does not fall back to English for a setting that now renders in Korean.

Scope is limited to these 5 keys. Running `repair-locale-catalog --locale ko` on main rewrites unrelated leaves of pre-existing drift, left out here to keep the diff reviewable.

### Note on the agentDashboard keys

The first revision of this PR also translated the Agent Dashboard block. #11205 landed Korean for those keys on main while this PR was open, so they are dropped here and main's wording is kept — the branch now carries no `agentDashboard` entries, matching main, which manages those keys in the catalog rather than in the override file. Thanks to the Greptile review for catching that the merge left the two files out of sync.

## Testing

- [x] `pnpm run typecheck` — clean
- [x] `pnpm run build:desktop` — built in 1m 30s
- [x] `pnpm run verify:localization-catalog` — parity at 11543 keys across es, ja, ko, zh
- [x] Vitest across the `i18n`, `settings`, and locale-script scopes — 199 files, 1336 passed, 4 skipped
- [x] All 5 values round-trip unchanged through `repairTranslatedValue` for `ko`, so the catalog pipeline will not rewrite them
- [x] Verified in a dev build with the interface language set to Korean (screenshots below)

Two repository gates could not be cleared locally, both unrelated to this diff:

- `pnpm lint` reaches a pre-existing localization-coverage failure for six Ghostty keyword strings in untouched settings files on main. Reproduced on a clean checkout with this branch stashed.
- `pnpm build:native` fails at code signing from duplicate certificates in my local keychain. The Swift target itself compiles and links.

No tests were added: this changes catalog values only, with no behavior to regress. `verify-localization-catalog` and the existing i18n suites already cover catalog parity and lookup.

## AI Review Report

Reviewed with Claude Opus.

- **Cross-platform.** Catalog data only, two JSON files, no code paths. Nothing platform-dependent is touched: no paths, shell invocation, keyboard accelerators, or Electron platform APIs.
- **SSH / remote / local.** Strings are resolved renderer-side from the catalog, so behavior is identical for local, folder-workspace, SSH, and remote-server sessions.
- **Agent / integration / git provider.** No key was added, removed, or re-hashed, so nothing that resolves a key by name is affected. No provider-specific or agent-specific behavior changes.
- **Performance.** Catalog key count is unchanged, so there is no lookup, memory, or startup cost.
- **UI quality.** Translated strings are at or below the English source in rendered width, so the setting row gains no new wrapping.

## Security Audit

No security-relevant surface. No input handling, command execution, path handling, auth, secrets, IPC, or dependency changes. The diff adds static translation strings to two JSON catalogs; all values are compile-time constants rendered as UI text and never interpolated into a shell, path, or query.

## Screenshots

Settings → General, interface language set to Korean.

Before:

1-general-before

After:

2-general-after


SonYoonho:
Thanks — good catch on the override/catalog divergence after the merge.

Resolved by dropping the 7 `agentDashboard` entries from `locale-ko-key-overrides.json` rather than reconciling them, so the branch now matches main: main carries no `agentDashboard` keys in the override file and manages that block in the catalog directly (#11205). `ko.json` keeps main's wording untouched, and this PR is narrowed to the 5 Editor Font Family strings, which main still renders in English.

Re-verified on the merged HEAD: all 5 remaining values round-trip unchanged through `repairTranslatedValue` for `ko`, catalog parity holds at 11543 keys across es/ja/ko/zh, and the i18n + settings + locale-script suites pass (199 files, 1336 tests).

One thing worth flagging separately, not addressed here: three of main's own `agentDashboard` values do not survive the pipeline either. `copy`, `toggleLabel`, and `modeAriaLabel` use `에이전트`, which `repairTranslatedValue` rewrites back to `agents` / `Agent` because `agent` and `agents` are in `NEVER_TRANSLATE_VALUES`. That is independent of this branch — it reproduces on `upstream/main` with no local changes, and `verify:localization-catalog` does not flag it. Happy to open an issue if that would be useful; it seems like a call on whether the policy list or the wording should change.

SonYoonho (commented):
.

<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/stablyai/orca/pull/10944)</sub>

SonYoonho:
@coderabbitai resume

SonYoonho:
Closing — #11573 landed the Editor Font Family strings for es/ja/ko/zh on 2026-08-01, which covers what this PR was for. Thanks to everyone who looked at it.

One follow-up worth noting: 3 of the 5 keys as merged do not round-trip through `repair-locale-catalog` in ko, ja, and zh, because `터미널` (and the equivalents) get rewritten back to `terminal` by `NEVER_TRANSLATE_VALUES`. Same shape as the agentDashboard keys from #11205. Filing that separately rather than reopening scope here.

PR #10945, opened 2026-07-27
fix(shell): preserve exit status across OSC 133 hooks

### LINKED ISSUE #10940: [Bug]: OSC 133 precmd hooks leak their own exit status — prompt frameworks show a phantom error on every prompt
### Operating system

macOS — but this is shell-level and affects Linux and Windows/Git Bash equally (bash is the default shell on most Linux distros).

### Orca version

1.4.153-rc.3. Confirmed still present at `main` (94009f63f).

### Details

**Short summary**

Orca prepends `__orca_osc133_precmd` to `PROMPT_COMMAND` (bash) and `precmd_functions` (zsh) so it can capture `$?` before user hooks overwrite it. The hook never restores that status before returning, so every prompt hook *after* Orca's in the chain reads the exit status of Orca's own last command instead of the user's. Prompt frameworks that render an error segment from `$?` — powerline-go, starship, oh-my-posh — are then wrong on every single prompt.

Two symptoms, in opposite directions:

1. **Phantom error on every prompt.** The daemon bash hook ends on

   ```bash
   [[ "${ORCA_SHELL_READY_MARKER:-0}" == "1" ]] && printf "${SHELL_READY_MARKER}"
   ```

   In a normal interactive terminal the marker is disabled, so that test is false and the function returns **1**. The user's prompt shows a failure indicator after every successful command.

2. **Real failures reported as success.** The other hooks end on `printf "\033]133;A\007"`, which succeeds — so a genuinely failed command is handed to the prompt chain as **0**.

**Where**

| File | Hook | Leaks |
|---|---|---|
| `src/main/daemon/shell-ready.ts:130-142` | bash | `1` (ends on marker test) |
| `src/main/daemon/shell-ready.ts:247-254` | zsh | `0` (ends on `printf`) |
| `src/main/providers/local-pty-shell-ready.ts:138-146` | bash | `0` |
| `src/main/providers/local-pty-shell-ready.ts:250-257` | zsh | `0` |
| `src/relay/pty-shell-launch.ts:156-164` | bash (SSH relay) | `0` |

The prepend is deliberate — `src/main/daemon/shell-ready.ts:202`:

```bash
PROMPT_COMMAND="__orca_osc133_precmd${PROMPT_COMMAND:+;${PROMPT_COMMAND}};__orca_osc133_epilogue"
```

and the zsh side carries the comment *"Why: prepend so Orca captures `$?` before user prompt hooks can overwrite it."* That prepend is exactly why the leaked status reaches the user's framework.

**How to reproduce**

Minimal, using the upstream hook body verbatim — save as `repro.bash` and run `bash --norc repro.bash`:

```bash
__orca_osc133_precmd() {
  local exit_code=$?
  __orca_in_prompt_command=1
  if [[ -n "${__orca_in_command:-}" ]]; then
    printf "\033]133;D;%s\007" "$exit_code"
    unset __orca_in_command
  fi
  printf "\033]133;A\007"
  [[ "${ORCA_SHELL_READY_MARKER:-0}" == "1" ]] && printf "READY"
}
# stands in for a prompt framework later in PROMPT_COMMAND
__user_prompt() { printf "USER_PROMPT_SEES:%s\n" "$?"; }

ORCA_SHELL_READY_MARKER=0; true;     __orca_osc133_precmd; __user_prompt
ORCA_SHELL_READY_MARKER=1; (exit 3); __orca_osc133_precmd; __user_prompt
```

Observed:

```
USER_PROMPT_SEES:1   # successful command reported as a failure
USER_PROMPT_SEES:0   # exit 3 reported as success
```

Expected `0`, then `3`.

In the app: set `powerline-go` or `starship` as your bash prompt, open an Orca terminal, and run `true`. The prompt renders an ERROR segment.

**Suggested fix**

End each precmd hook with `return "$exit_code"`. `exit_code` is already captured on the first line of all five of them, so it's a one-line addition per hook.

**Why current tests don't catch it**

The real-bash coverage in `src/main/daemon/shell-ready.test.ts` always spawns with `ORCA_SHELL_READY_MARKER: '1'` (in `runInteractiveBashRcfile`), which is the one case where the trailing marker test succeeds. Adding a case with the marker at `0` and a probe appended to `PROMPT_COMMAND` catches both directions.

AnddyAgudelo:
## Summary

- return the captured foreground-command status from OSC 133 precmd hooks
- preserve the status seen by downstream prompt framework hooks in daemon, local PTY, and SSH relay sessions
- add regression coverage proving Bash prompt hooks still observe a failing command and keep Bash/Zsh wrapper generation in sync

## Validation

- `pnpm run typecheck`
- `npx oxlint src/main/daemon/shell-ready.ts src/main/daemon/shell-ready.test.ts src/main/providers/local-pty-shell-ready.ts src/main/providers/local-pty-shell-ready.test.ts src/relay/pty-shell-launch.ts src/relay/pty-shell-launch.test.ts`
- `npx vitest run --config config/vitest.config.ts src/main/daemon/shell-ready.test.ts src/main/providers/local-pty-shell-ready.test.ts src/relay/pty-shell-launch.test.ts`
- `pnpm run build:electron-vite`
- `pnpm run build:web`

Closes #10940

### DIFF
--- a/src/main/daemon/shell-ready.test.ts
+++ b/src/main/daemon/shell-ready.test.ts
@@ -98,7 +98,14 @@ async function runInteractiveZshRc(args: {
   return output
 }
 
-function runInteractiveBashRcfile(rcfileContent: string, tempDir: string): string {
+function runInteractiveBashRcfile(
+  rcfileContent: string,
+  tempDir: string,
+  // Why: attribution shells launch with the marker off, and the trailing marker
+  // test used to be the precmd's last command — only marker='0' exposes the
+  // phantom-error direction of #10940.
+  readyMarker: '0' | '1' = '1'
+): string {
   const rcfile = join(tempDir, 'bash-osc133-rcfile')
   writeFileSync(rcfile, rcfileContent)
 
@@ -111,7 +118,7 @@ function runInteractiveBashRcfile(rcfileContent: string, tempDir: string): strin
       env: {
         ...process.env,
         HOME: tempDir,
-        ORCA_SHELL_READY_MARKER: '1',
+        ORCA_SHELL_READY_MARKER: readyMarker,
         TERM: process.env.TERM || 'xterm'
       },
       timeout: 5000
@@ -500,7 +507,9 @@ describePosix('daemon shell-ready launch config', () => {
     )
     expect(zshrc).toContain('printf "\\033]133;D;%s\\007"')
     expect(zshrc).toContain('printf "\\033]133;C\\007"')
-    expect(zshrc).toContain('return "$exit_code"')
+    // Why: zsh restores $? per precmd hook, so a return here would only double
+    // the ERR-trap fires for a failed command (#10940 review).
+    expect(zshrc).not.toContain('return "$exit_code"')
   })
 
   itWithBash(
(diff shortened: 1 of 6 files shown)


OrcaWin:
Reviewed with an independent reproduction in real bash 3.2 and 5.2 and real zsh 5.9 (macOS) / 5.8 (Ubuntu 20.04). **The bash half is a real, well-targeted fix. The zsh half is unnecessary and introduces a small regression, so I dropped it** — details below, because it contradicts the issue you filed and you should get to push back.

**Bash: confirmed, both directions.** `PROMPT_COMMAND` entries run as a plain command sequence, so Orca's prepended `__orca_osc133_precmd` hands its own last command's status to everything after it. With the shell-ready marker off, a *successful* command reached downstream hooks as status 1 — the phantom error on every prompt. With the marker on, a genuine failure arrived as 0. `return "$exit_code"` fixes both, on all three transports.

**zsh: the premise does not hold.** zsh restores `$?` before each `precmd_functions` entry, so downstream hooks and PS1 `%?` already saw the correct status. Decisive test: a precmd hook doing `return 42` still leaves the next hook printing the real status, and an A/B of the actual generated wrappers through a real PTY gave identical results with and without the return — `[0,0,1]` either way, on both zsh versions.

It is not merely a no-op, it costs something: a failed command now fires a user's ERR trap **twice** per prompt instead of once (1→2, 2→4). So I removed the zsh `return` from both wrappers and left a comment explaining the asymmetry, so a future reader does not "restore the missing line."

This means **issue #10940's claim that the zsh hooks leak status 0 is incorrect.** If you have a case where zsh genuinely loses the status, I would like to see it — happy to be wrong here.

**Coverage hardening.** Your assertion `toContain('PROMPT_STATUS:1')` would still be satisfied by a `0,0,0` regression, so the three real-bash tests now assert the exact sequence `['0','0','1']`. More importantly, every pre-existing real-bash test spawned with `ORCA_SHELL_READY_MARKER='1'` — the one value that *hides* the issue's headline symptom — so the PR shipped no coverage of the phantom-error direction at all. Added a marker-off test that pins both directions and proves a non-zero precmd return does not truncate the `PROMPT_COMMAND` chain.

**Regression hunt on the bash side came back clean:** hook chaining intact (a later hook still ran 3/3), the bash-preexec/starship shape receives the real status, `set -e` does not exit, and the OSC 133 C/D lifecycle invariants are unchanged. One accepted cost: one extra ERR-trap fire per genuinely failing command when the marker is on. With the marker off the fix actually *reduces* spurious fires, because the old trailing `[[ ... ]]` test was failing every prompt.

Validation: 104 tests passing across the three shell-ready suites, typecheck exit 0.

Nice find — the marker-off direction is the one that bites real users on every prompt.

ushuz:
Hey @AmethystLiang @OrcaWin, what's holding this back?

PR #10988, opened 2026-07-27
fix(mobile): collapse agent reasoning into a Thinking disclosure

keepitmello:
## Summary

Mobile renders agent reasoning fully expanded. This folds it into a one-line `Thinking` disclosure that opens on tap — the same fold tool runs already have on mobile. Six files, all under `mobile/`.

**Not a duplicate of #9417.** That PR does the desktop row and the Claude decoder; its own notes say "the collapsed disclosure itself is desktop-only." This is the mobile half — zero file overlap, and the two can land in either order. Reasoning already reaches mobile today from Codex and Grok sessions, so this helps right away; once #9417 lands, Claude sessions get it too. The preview scan here walks the string by index and never allocates per line, matching the shape of `firstNonEmptyLinePreview` on that branch, so the two platforms stay consistent.

### Why

On my phone, the agent's thinking takes over the screen.

A single reasoning turn is often longer than the answer that follows it. On a desktop that's tolerable — there's room, and you can skim past it. On a phone it means scrolling through several screens of the agent talking to itself before you find the one paragraph you actually wanted. If I'm checking on a run while I'm out, that's most of what I see.

Tool calls already solved this on mobile — they fold into a single line you can tap open. Reasoning didn't. A reasoning turn now shows up as one line:

```
▸ Thinking  The user is reporting a two second delay before the worktree…
```

Tap it and the full thought opens. The toolbar's expand toggle opens them all, same as it does for tool runs. The answer is the first thing on screen again, and reasoning is there when you want it instead of by default.

## Screenshots

Before, after collapsed, and after tapping open: a reasoning turn folded into a one-line Thinking disclosure

Captured on the iOS simulator against the native-chat mock server.

## Testing

- [x] `pnpm lint` — mobile lint clean; the root run stops on a pre-existing `verify:localization-coverage` failure in `src/renderer/src/components/settings/terminal-advanced-platform-search.ts` and `terminal-pane-appearance-search.ts`, both untouched here, which reproduces on a clean `origin/main` checkout — the same failure #10814 described when it merged yesterday. The root `verify` job in `pr.yml` runs `pnpm exec oxlint` rather than `pnpm lint`, and that passes.
- [x] `pnpm typecheck`
- [x] `pnpm test` — `cd mobile && pnpm test`: 347 files, 2552 passed, 2 skipped, 0 errors
- [x] `pnpm build` — passes, though nothing under `mobile/` is part of it; the gate for these files is `mobile.yml`

Everything here is under `mobile/`, so the checks that actually cover it are the ones in `.github/workflows/mobile.yml`. All five of its verification steps pass locally on macOS with Node 24 and pnpm 10.24.0: `typecheck`, `test`, the Ruby iOS release version test, `lint`, and `format:check`. I also ran the root `verify` job's lint and typecheck steps locally, since that job has no path filter and its oxlint step does cover `mobile/`. Both pass.

Sixteen tests added, all `.test.ts` so they're actually picked up — `mobile/vitest.config.ts` only includes `src/**/*.test.ts`, so a `.test.tsx` would silently never run. That's also why the preview scan lives in its own module instead of inline like the desktop version.

Six in `mobile-native-chat-reasoning-preview.test.ts` cover the one-line preview: blank leading lines get skipped, it clips at 80 characters, the second line never bleeds into the first. Ten in `MobileNativeChatMessage.test.ts` cover copy still copying the whole thought rather than the visible line, a normal assistant answer never folding, tapping twice closing it again, and empty reasoning producing no row.

## AI Review Report

I had an agent review this adversarially against #9417 and the repo's conventions, then re-review after I made the changes. It came back with one real bug and five smaller things, all fixed before this PR.

The bug: my first version folded every reasoning turn unconditionally, including ones with nothing readable in them. That's reachable today — a whitespace-only thought passes Codex's `if (!text)` (`src/main/native-chat/transcript-line-decoders-codex.ts:57`) where Grok's `!text?.trim()` (`transcript-line-decoders-grok.ts:72`) drops it, so it reaches the renderer. Before this PR it drew an invisible empty box; after, it would have drawn a tappable row that opens onto nothing. Now the preview is computed first and an empty one falls through to the old path. Two of the added tests cover it.

The rest: a conditional that became dead code once the early return existed; a new style that was byte-identical to `toolRunBody`, now deleted in favor of reusing it; an `accessibilityLabel` that replaced the row's children and left screen readers with "Thinking, button" and no content, now composed as `Thinking: <preview>` while collapsed; a three-line comment where one line did the work; and four missing tests, which are in.

It also confirmed copy still reads from the message rather than the screen, so collapsing can't truncate it, and that the preview scan stays index-based with no per-line allocation.

On cross-platform: nothing here ships in the desktop app. The Electron build doesn't compile `mobile/`, so macOS, Linux, and Windows are unaffected — no shortcut labels, no path handling, no shell behavior, no Electron-specific code in the diff. Within mobile it's React Native with no `Platform` checks; `hitSlop`, `numberOfLines`, and `accessibilityState.expanded` behave the same on iOS and Android.

On SSH and remote: this row renders text the transcript already delivered. It opens no process, reads no file, and touches no credential or network path, so it behaves identically whether the session is local, on a remote server, or in an SSH worktree. It doesn't care whether the workspace is a git worktree or a plain folder, and it never runs Git, so there's no Git version or provider surface either.

On agents and integrations: the row keys off the `reasoning` role, not the agent that produced it, so Claude, Codex, and Grok all get the same treatment with no per-agent branching. Only where that role comes from differs today, and that difference is what #9417's decoder change closes. Nothing here touches source control or a git provider, so GitHub and GitLab are unaffected.

On performance: collapsed is the cheaper state. The reasoning body isn't mounted while closed, so a long thought costs no markdown render and no layout until you tap it — on a transcript with many reasoning turns that's less work than today, not more. The preview itself scans by index and allocates one string of at most 80 characters.

## Security Audit

Nothing new here to attack. No new input parsing, no commands run, no paths handled, no auth, no secrets, no new IPC, no new dependency.

The one new function reads text that's already in the message and returns the first 80 characters of it. It walks the string by index instead of splitting it into lines, so a huge single-line thought doesn't turn into a huge allocation.

Reasoning text isn't newly exposed anywhere — it's less visible than before, which is the point. Copy is untouched and still reads from the message itself rather than what's on screen, so collapsing can't quietly truncate what someone copies.

## Notes

- No i18n change. There is no i18n layer under `mobile/src` at all: no locale files, no runtime. Modules ported from the renderer say so explicitly — `mobile/src/components/pr-sidebar/pr-comment-audience.ts:10` and `mobile/src/source-control/hosted-review-copy.ts:4` both note they are the renderer module "minus its i18n wrapper." The new `Thinking` label follows the plain-English strings already next to it.
- The preview helper is deliberately duplicated for now rather than shared. `mobile/` is a separate pnpm project and #9417 isn't in yet, so hoisting it would mean depending on an unmerged branch. Once that lands, I'm happy to follow up and collapse both into one function under `src/shared`.
- Reasoning rows follow the existing toolbar expand toggle, using the same remount trick tool runs use — the same signal #9417 feeds `NativeChatReasoningRow` on desktop. Desktop calls it `expandSignal`; mobile calls it `toolsExpanded` and labels the button `Tools` (`MobileNativeChatView.tsx:384`). Behavior matches, only the mobile name lags — happy to rename that string here or later.
- Not verified: Android and a physical device. I tested on the iOS simulator against the native-chat mock server; that leg is covered by the unit tests, and there are no `Platform` checks in the diff, so there is no Android-specific path to diverge.
- Small pre-existing thing I noticed and left alone: a blank reasoning turn still draws an empty bubble with a Copy button that does nothing, because `handleCopy` bails on empty text. Same before and after this PR. Happy to fix separately if it's worth it.

X handle: @keepitm3llo

### DIFF
--- a/mobile/src/session/MobileNativeChatMessage.test.ts
+++ b/mobile/src/session/MobileNativeChatMessage.test.ts
@@ -1,6 +1,6 @@
 import { createElement } from 'react'
 import { act, create, type ReactTestInstance, type ReactTestRenderer } from 'react-test-renderer'
-import { afterEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { MAX_TOOL_DETAIL_LENGTH } from '../../../src/shared/native-chat-tool-summary'
 import type { NativeChatMessage } from '../../../src/shared/native-chat-types'
 
@@ -25,6 +25,7 @@ vi.mock('lucide-react-native', () => ({
 }))
 vi.mock('../components/MobileMarkdown', () => ({ MobileMarkdown: 'MobileMarkdown' }))
 
+import * as Clipboard from 'expo-clipboard'
 import { MobileNativeChatMessage } from './MobileNativeChatMessage'
 
 function userMessage(blocks: NativeChatMessage['blocks']): NativeChatMessage {
@@ -33,6 +34,17 @@ function userMessage(blocks: NativeChatMessage['blocks']): NativeChatMessage {
 
 function toolMessage(blocks: NativeChatMessage['blocks']): NativeChatMessage {
   return { id: 'a1', role: 'assistant', blocks, timestamp: null, source: 'transcript' }
+}
+
+function roleMessage(
+  role: NativeChatMessage['role'],
+  blocks: NativeChatMessage['blocks']
+): NativeChatMessage {
+  return { id: 'm1', role, blocks, timestamp: null, source: 'transcript' }
+}
+
+function reasoningMessage(text: string): NativeChatMessage {
+  return roleMessage('reasoning', [{ type: 'text', text }])
 }
 
 describe('MobileNativeChatMessage', () => {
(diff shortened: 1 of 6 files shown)


keepitmello:
@OrcaWin rebased onto current main. Same mobile-only change — reasoning still folds into a one-line Thinking disclosure, no file overlap with #9417. Local mobile tests for the touched files are green.

Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17579. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @keepitmello. No action is requested from you. Thank you.

PR #10991, opened 2026-07-27
feat(browser): let Shift invert link routing instead of always forcing the system browser

brennanb2025:
## Problem

`⇧⌘`-click (`Shift+Ctrl` off macOS) has always meant **"open in the system browser"** — a one-way escape hatch, not a toggle:

```ts
const routeToOrca =
  sourceIsLocal &&
  !forceSystemBrowser &&        // ← Shift only ever subtracts
  Boolean(worktreeId) &&
  state?.settings?.openLinksInApp === true
```

That works when Link Routing is **on**. When it's **off** — links already go to the system browser — the gesture is a no-op, and there's no way to pull a single link into Orca's built-in browser.

| Link Routing | Plain `⌘`-click | `⇧⌘`-click (before) |
|---|---|---|
| On | Orca browser | System browser ✅ |
| Off | System browser | System browser ❌ dead gesture |

## Change

A nested toggle under Link Routing: **"Hold Shift to open in ___"**. When on, the modifier opens a link the *opposite* way from the setting, so the other destination is always one click away.

**It ships off**, so the existing one-way behavior is byte-for-byte unchanged for every current user.

The title and description name the destination the modifier actually reaches and flip with the parent setting — "the opposite" means nothing on its own:

| Link Routing | Row title |
|---|---|
| Off | Hold Shift to open in **Orca** |
| On | Hold Shift to open in **your web browser** |

### Screenshots — real app, this branch

Default state (Link Routing off, new toggle off — identical to today):

default

Title and description flip live when Link Routing is turned on:

flipped

The row nests with `ml-4 border-l border-border pl-4`, the same parent-toggle-with-dependent-child pattern as `AgentDashboardExperimentalSetting.tsx:59` and `NativeChatExperimentalSetting.tsx:61`.

## Implementation

`openHttpLink` gains `modifierHeld`, and `resolveModifierRouting` owns the decision in one place, so every surface routes identically — terminal URLs, OSC 8 hyperlinks, xterm web links, markdown preview, and the Checks panel's hosted-review link all just pass `modifierHeld: event.shiftKey`. `forceSystemBrowser` is retained for the callers that must bypass settings unconditionally (the resolved first-use prompt, and the rich markdown editor, whose Shift path stays one-way by design).

Two copy fixes the toggle forces:
- The terminal hover hint said "for system browser" unconditionally; it now names Orca when that's where the modifier would land, and is re-resolved **per hover** (via a new `linkOpenHint` PaneManager option) so toggling applies without recreating panes.
- Link Routing's own description said `⇧⌘-click` **always** uses your system browser — which the new toggle can falsify. The nested row now owns that sentence and states the live destination.

## Testing

- **13 new unit tests** across `resolveModifierRouting` and `openHttpLink`, including the negative control (inverting on + Link Routing on is *identical* to legacy) and that a remote/SSH source is never routed into Orca even when inverting.
- A guard test pinning the `getBrowserPaneSearchEntries()` order, since `BrowserPane` gates rows by array index and I inserted one.
- Full suites green: `src/renderer/src/lib/` + `components/settings/` (3831 passed), `components/terminal-pane/` + `lib/pane-manager/` (3085 passed). All three typecheck configs clean, `oxlint` 0 errors, `oxfmt` clean.

### Live A/B in the running app

Exercised the real `http-link-routing` module in the dev build over CDP, on this branch:

| Link Routing | Invert | Modifier | Result |
|---|---|---|---|
| Off | **On** | held | **Orca tab created** — the new behavior |
| Off | On | plain | real `shell.openUrl` fired |
| Off | **Off** | held | real `shell.openUrl`, no Orca tab — legacy preserved |

`window.api.shell` is frozen by contextBridge, so the system-browser cases invoked the genuine IPC rather than a stub.

## Not covered

The end-to-end gesture was driven through the real routing module, not by a synthetic `⇧⌘`-click on a rendered terminal cell — the hit-testing path (`installHttpLinkClickFallback`) is exercised by its existing unit tests, but not by a live mouse event in this validation.


brennanb2025:
## Live QA: the "Open on GitHub" button — found and fixed a miss

Verifying this surface turned up a real gap, now fixed in 27b4e5ae74.

`checks-panel-hosted-review-click-routing.ts` had its **own** `⇧⌘` escape hatch that passed `forceSystemBrowser: true` directly, so the Checks-panel PR link kept the old one-way behavior while every other surface honored the new setting. It's now routed through `modifierHeld` like the terminal and markdown paths.

### Verified by clicking the actual button

Drove the real button in the running dev build (Checks panel loaded against this very PR), observing `createBrowserTab`:

| Link Routing | Invert | Real `⇧⌘`-click on **Open on GitHub** | Result |
|---|---|---|---|
| Off | **On** | yes | **Orca tab created** with `.../pull/10991` ✓ |
| Off | **Off** | yes | no Orca tab — real `shell.openUrl`, legacy preserved ✓ |

checks panel

Also fixed in the same commit: `getLinkRoutingModifierDescription` returned raw English while the title in the same file was localized — thanks @coderabbitai, that was a genuine inconsistency. Both branches now go through `translate()` with `{{chord}}` interpolation, catalog synced.

### One deliberate exclusion

`rich-markdown-editor-click-routing.ts:125` also gates on `event.shiftKey`, but routes to `openMarkdownLinkInClientOs` — a broader "hand this to the OS" path that also opens local files with the default app. Inverting only its http branch would make one gesture mean two different things depending on the target type, so I left it alone. Happy to convert it if you'd rather have strict consistency.

Issue #9509, opened 2026-07-20
[Feature]: Unified notification panel across projects/sessions with click-to-jump


wuwenrui:
## Problem or use case

When working across multiple projects at the same time, each with one or more parallel agent sessions, it is hard to tell **which session in which project just finished**. Today I have to scan every workspace/terminal manually to find the one that completed or needs attention. There is no single place that aggregates "what just happened, and where".

This is different from the per-session attention indicators added in #979: that shows a badge on an individual session item. What is still missing is a **unified, cross-project notification center** that collects events from all projects/sessions in one list.

## Proposed solution

Add a global **notification panel / inbox** that aggregates events across all projects and sessions:

- **Unread badge** on the panel entry (e.g. sidebar/toolbar icon) showing the count of unread notifications.
- Each notification entry clearly shows:
  - **Which project** it came from.
  - **Which session/agent** produced it.
  - **What the event is** (e.g. session completed, blocked waiting for approval, error).
  - A timestamp.
- **Click to jump**: clicking a notification navigates directly to that project's session, and marks it as read (clearing its unread state).
- Notifications persist in the panel so they can be reviewed later, not just shown as a transient toast.

## Alternatives or additional context

- Related: #979 (per-session attention indicators) and #556 (agent bell / task-complete messages). This request builds on those by centralizing the events into one browsable, clickable list rather than only decorating individual session items.
- Nice-to-haves (optional): filter by project, group by project, "mark all as read", and respecting the existing notification/sound preferences.

PR #10996, opened 2026-07-27
fix(dock): match Agent View unread sessions

simmons-OS:
## Summary

The macOS Dock badge now matches unread Agent View sessions, so acknowledging an agent clears the same notification in both places. Multiple unread events from one session count once.

The mismatch came from separate unread models: Agent View used per-agent acknowledgements, while the Dock counted persistent workspace and terminal flags that survived restarts.

Related: #9509

## Screenshots

No visual change. This changes the native macOS Dock badge only.

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

`pnpm lint` reaches the localization coverage gate, then fails on existing `Ghostty` search keywords in two untouched settings files. Oxlint and formatting checks pass for every changed file.

The focused Dock and Activity regression suite passes 35 tests. The final full suite passes 37,931 tests with 60 skipped. One unrelated PTY timing test failed on the first full run, then passed alone and in the second full run.

## AI Review Report

- Cross-platform: the renderer count remains platform-neutral, the native badge remains guarded to macOS, and the web implementation remains a no-op. No shortcuts, labels, paths, shell behavior, or Electron platform routing changed.
- SSH/remote/local and folder workspaces: the count uses existing agent-status projections and adds no host, repository, process, credential, or filesystem assumptions.
- Agents and integrations: live, retained, and migration-unsupported sessions use the existing provider-neutral agent status model. No git provider or integration behavior changed.
- Performance: the new `Set` dedupes pane keys inside the existing memoized unread-count pass, with no new polling or hot-path subscriptions.
- UI quality: no in-window UI changed. The Dock now uses the same acknowledgement semantics as the gold unread bell in Agent View.

## Security Audit

No new input, command execution, path handling, auth, secret, dependency, or IPC surface was added. The existing IPC path still accepts only a normalized numeric count, and non-macOS handlers remain no-ops.

## Notes

This intentionally defines the Dock badge as agent sessions needing attention. General workspace or terminal-only unread activity no longer increments it.

Reproduced on Orca 1.4.159 on macOS. The live Dock label matched the persisted unread-workspace count while Agent View used a separate per-agent acknowledgement map.

X: [@charlesrsimmons](https://x.com/charlesrsimmons)


simmons-OS:
“Docstring Coverage” is a generic warning, not an Orca CI requirement. The repo has no docstring coverage rule, and adding boilerplate would violate the concise-comment convention. No PR change needed.

mmarabel:
Confirmed this behavior on Orca v1.4.167 on macOS. We reproduced it when Claude asked a question while Orca was focused on a different worktree, and also while Orca was backgrounded. The Claude session showed that it needed input, but the macOS Dock icon did not show the unread red badge.

Filed #12342 with the reproduction details. Sharing this here because the behavior appears to match the unread-state mismatch described in this PR.

PR #11001, opened 2026-07-27
fix(dashboard): stop the agent icon shrinking on long card titles

brennanb2025:
## Problem

In the agent dashboard, one card's Claude icon renders noticeably smaller (and washed out) than the rest:

| | |
|---|---|
| normal | long title |
| 14px | ~9px |

It looked like a status or agent-type difference. It isn't — it's flexbox.

## Cause

`AgentKanbanCard` rendered the icon as a bare `<svg>` flex item:

```tsx
<div className="flex items-center gap-1.5">
  <AgentIcon ... size={14} />   // default flex-shrink: 1
  <span className="truncate">{card.worktreeName}</span>
```

An `<svg>` with the default `flex-shrink: 1` shrinks along with its row. Cards whose worktree name fits are unaffected; the moment the title overflows and truncates, the browser distributes the shrink proportionally to flex-basis and the 14px icon loses ~5px. The dimmer color is just antialiasing on the downscaled glyph.

`AgentTerminalDialog` had the same bare call site, so a long worktree name in the dialog header squeezed its 13px icon too.

Every other surface already guards against this — `DashboardAgentRow`, `worktree-card-compact-agent-row`, and `ActivityPrototypePage` all wrap `AgentIcon` in `<span className="inline-flex shrink-0">`, and `AgentStateDot` bakes `shrink-0` into its own wrapper. These two popout call sites were the outliers.

## Fix

Wrap both in the same `inline-flex shrink-0` span the rest of the app uses. No layout change for cards that already fit.

## Testing

- `AgentKanbanCard.test.tsx` — 4/4 pass.
- `oxfmt --check` clean; lint-staged (oxlint + react-doctor) clean on commit.

Note: the existing tests mock `AgentIcon`, so they don't cover sizing. Verification here is the CSS reasoning plus parity with the surfaces that already carry the guard.

## Follow-up (not in this PR)

The durable fix is moving `shrink-0` inside `AgentIcon` so no call site can omit it, but that needs a `className` pass-through on ~10 icon components — out of scope for a one-line rendering bug.

### DIFF
--- a/src/renderer/src/components/dashboard-popout/AgentKanbanCard.tsx
+++ b/src/renderer/src/components/dashboard-popout/AgentKanbanCard.tsx
@@ -86,7 +86,10 @@ export const AgentKanbanCard = memo(
         )}
       >
         <div className="flex items-center gap-1.5">
-          <AgentIcon agent={agentTypeToIconAgent(card.agentType)} size={14} />
+          {/* Why: a bare <svg> flex item shrinks with the row — long worktree names squashed the icon. */}
+          <span className="inline-flex shrink-0">
+            <AgentIcon agent={agentTypeToIconAgent(card.agentType)} size={14} />
+          </span>
           <span
             // Why: same unvisited treatment as the sidebar's DashboardAgentRow —
             // bold+bright until acked, normal+muted after — so both surfaces
--- a/src/renderer/src/components/dashboard-popout/AgentTerminalDialog.tsx
+++ b/src/renderer/src/components/dashboard-popout/AgentTerminalDialog.tsx
@@ -79,7 +79,9 @@ export function AgentTerminalDialog({
           }}
         >
           <div className="flex items-center gap-1.5 px-2.5 py-2">
-            <AgentIcon agent={agentTypeToIconAgent(card.agentType)} size={13} />
+            <span className="inline-flex shrink-0">
+              <AgentIcon agent={agentTypeToIconAgent(card.agentType)} size={13} />
+            </span>
             <DialogTitle className="text-[12px] leading-normal font-semibold">
               {card.worktreeName}
             </DialogTitle>


brennanb2025:
## Live Electron A/B (dev build, this branch)

Attached CDP to the dev app on this worktree (`Orca: brennanb2025/fix-claude-icon-size`), enabled the experimental pop-out dashboard, and drove the real producer→pop-out bridge (`window.api.dashboard.publishSnapshot`) with four cards — two whose worktree names overflow, two that fit. Rendering path is the production `AgentKanbanCard`; only the card data is synthetic.

Both shots are the same window at the same 700px viewport and scroll position. The only difference is whether the `shrink-0` wrapper is present (toggled live via HMR).

**Before — bare `<svg>` (bug reproduced)**



**After — this branch**



### Measured icon widths (`getBoundingClientRect`)

| Card title | Overflows? | Before | After |
|---|---|---|---|
| `Feedback terminal visibility ux` | no | 14.00 | 14.00 |
| `fix(pty): strip inherited Claude child-session-id` | yes | **9.90** | 14.00 |
| `orca-trust-dialog` | no | 14.00 | 14.00 |
| `brennanb2025/tcc-helper-usage-strings` | yes | **10.61** | 14.00 |

Only truncated cards shrink, and shrinkage scales with overflow — at a 960px viewport the same two cards measured 11.86 / 12.71. That matches the original report (a ~9px icon in a narrow window) and confirms the cause is flex shrink, not agent type or card state.

### Not covered

- `AgentTerminalDialog` (the second call site in this PR) was not exercised live — it needs a card with a live `ptyId`, and this instance had no running agents. The markup change is identical to the card's.
- Card data was injected over the real IPC bridge rather than produced by live agent sessions.

PR #11004, opened 2026-07-27
fix(ai-vault): support session scanning in SSH worktrees

AmethystLiang:
## Summary

Adds relay-side Agent Session History scanning for SSH worktrees. Previously session listing relied entirely on desktop-side filesystem crawling. This change enables scanning directly on the SSH host when the relay supports it, with automatic fallback to desktop crawling for older relays.

**Key improvements:**
- Relay-side scanning eliminates per-file network roundtrips
- Cancellation support via AbortSignal throughout the scan pipeline
- Graceful handling of result validation with per-entry error collection
- Scope path truncation (64 item limit) with user-facing warnings
- Coalesced in-flight scans with independent caller cancellation, `force` preemption of a hung scan, and cache fencing so an aborted scan never poisons the cache
- Bounded all-host SSH legs: the legacy desktop crawl now runs inside the same `timeoutMs` budget as the relay round trip, and a host that throws or times out degrades to one issue row instead of discarding local/runtime results

## Screenshots

No visual change.

## Testing

- [x] `pnpm run typecheck:node`, `pnpm run typecheck:web`
- [x] `pnpm run check:code-quality:changed` (0 new findings / 47 changed files), `pnpm run check:max-lines-ratchet`, `git diff --check`, `oxfmt --check`
- [x] Focused vitest over all changed AI Vault / relay / renderer files — 215 files, 1756 tests passing
- [x] New/extended unit tests:
  - `ai-vault-scan-coordinator.test.ts` — waiter lifecycle, forced preemption, abort fencing
  - `ssh-session-list.test.ts` (new) — bounded legacy crawl, unbounded when no budget, crawl throw degrades to a host issue, caller cancellation still propagates
  - `session-list-results.test.ts` (new) — scope-truncation notice survives `mergeAiVaultListResults`, one notice per host, never counted as a skipped transcript
  - `ai-vault-scan-coalescing.test.ts` (new) — coalescing across independent caller signals
  - `ai-vault.test.ts` — all-scope host-discovery failures and a throwing local leg both degrade to issue rows without dropping other hosts
  - `session-list-result-validation.test.ts` — all-invalid relay payload rejects so the leg falls back
  - `ai-vault-scan-issue-state.test.ts` / `ai-vault-session-refresh.test.ts` — issue classification, skipped-transcript reasons, cancellation recognition
- [x] E2E driver updated (`ssh-codex-reconnect-replay-driver.ts`) to poll for remote worktrees instead of assuming instant availability

## Review Answers

**Does the relay path avoid the SSH filesystem stream/mux limits (#11396, #11364)?**

Partly — and the distinction matters:

- **Transport: yes.** `AiVaultHandler` (`src/relay/ai-vault-handler.ts`) runs `scanRemoteAiVaultSessions` (`src/main/ai-vault/remote-session-scanner.ts`) *inside the remote relay process*, against a host-local provider built by `createRelayAiVaultFilesystemProvider()` — `readdir`/`lstat` from `node:fs/promises` plus `readRelayFileContent`. No transcript bytes cross the desktop SSH filesystem stream/mux, so the chunked-stream pump and its unacked-chunk backpressure are out of the picture entirely. **This does not reduce #11396's urgency:** that issue tracks the 10 MiB cap, which remains on both paths.
- **10 MiB cap: no, it still applies.** `readRelayFileContent` enforces `MAX_TEXT_FILE_SIZE = 10 * 1024 * 1024` (`src/relay/fs-handler-utils.ts:25`) before returning content, exactly mirroring the desktop reader's own `10 * 1024 * 1024` (`src/main/ssh/ssh-filesystem-stream-reader.ts:9`). A >10 MiB transcript is skipped on both paths; it is just skipped host-locally now instead of after a stream setup. Raising that cap for transcripts is out of scope here.

**Does the older-relay fallback still carry the original risks?**

Yes. When the relay has no `aiVault.listSessions` method the leg falls through to the desktop crawl over `ssh-filesystem-stream-reader.ts`, so both the missing per-read inactivity timeout and the 10 MiB cap still apply. This PR bounds that crawl with an overall wall-clock budget in all-host scope (`scanRemoteSessionsWithinBudget` in `ssh-session-list.ts`), which stops one stalled host from holding every other host's results hostage — but it does not add the per-read inactivity timeout. **#11364 remains a required companion**, not something this PR closes.

**How does the 64-path scope warning behave end to end?**

It is an `AiVaultScanIssue` with `kind: 'scope'`, emitted identically by both legs (relay leg in `ai-vault-handler.ts`, fallback leg via `scopeTruncationIssue` in `ssh-session-list.ts`) so neither can silently scan more paths than the other. It survives `mergeAiVaultListResults`, which concatenates every result's issues without dedup — covered by `session-list-results.test.ts`, including one notice per host and a scope notice surviving alongside a failing host. It renders through `AiVaultScanIssueBanners` (mounted in `AiVaultPanel.tsx:338`) using the scanner-authored `issue.message`. Because `skippedAiVaultTranscriptCount` counts only issues with no `kind`, it is excluded from the skipped-transcript footer.

**First user prompt, relative to #12006 on main**

**Keeping remote prompts null — no change needed here.** Two independent reasons:

1. This branch predates #12006: `firstUserPrompt` does not exist anywhere in the branch's source (`src/main/ai-vault/session-first-user-prompt-read.ts` and friends are `origin/main`-only). There is nothing in this diff that can regress or implement it.
2. Main's own reader already takes this posture deliberately. `readAiVaultFirstUserPrompt` returns `{ prompt: null }` for any `executionHostId !== LOCAL_EXECUTION_HOST_ID`, with the comment "transcript bodies live on the session host. Remote rows are skipped (same posture as `listSubagentSessions`); UI falls back to preview text." So after this branch merges main, SSH rows will resolve to `null` and fall back to preview text, which is the intended behavior — not a regression introduced here.

Serving full first prompts for remote hosts would need a new relay method to read the transcript body host-side. That is a separate change and is not attempted in this PR.

## CI Housekeeping

Run [30385188010](https://github.com/stablyai/orca/actions/runs/30385188010) is stale — it predates the fixes in the current working tree.

- **`static analysis` — real, work-caused, now fixed.** It failed on a single `--deny-warnings` warning, "Modules should not be imported multiple times in the same file": `remote-session-scanner.ts` imported from `./remote-session-scanner-types` twice (lines 17 and 24 at that commit). The current tree merges them into one `import type`, and `check:code-quality:changed` now reports 0 new findings.
- **`e2e 9-of-10` — `cancelled`, not a failure.** The GitHub API job conclusion is literally `cancelled`; it is not independent evidence of a broken test and should not be counted as one.
- **`e2e 1/3/7-of-10` — unrelated red.** Failures are agent-session force-exit/quit resume, new-workspace Jira/GitHub linked-item project switching, and OpenCode alt-screen reveal artifacts (STA-2694). None touch AI Vault.
- **`tests 7/16` — unrelated infrastructure error.** All 2175 tests passed; the job failed on an unhandled `EPIPE` thrown from `wsl-hook-relay-sentinel.ts` during `wsl-hook-relay-live.integration.test.ts` teardown.

No live reproduction of the e2e shards was attempted locally; the above is read off the run's logs and job conclusions.

## Notes

- Relay method `aiVault.listSessions` timeout: 130s; timeouts do not trigger fallback crawl
- `force` propagates desktop → relay (`SshAiVaultRelayListParams.force`) but is deliberately excluded from the relay coordinator's coalescing key, so a forced refresh preempts a stale scan instead of fragmenting the cache
- A forced scan preempts an in-flight one immediately if that scan was unforced, or after 5s if it was forced (`FORCED_SCAN_PREEMPT_AFTER_MS`), so a hung scan can't strand every later Refresh
- Cancellation is recognized in the renderer by shared message constant (`AI_VAULT_SCAN_CANCELLED_MESSAGE`) because IPC drops `Error.name`; a cancelled scan leaves the existing list in place rather than painting an abort error over good state
- A relay payload whose sessions are *all* invalid now throws in `parseAiVaultListResult` so the leg falls back to the desktop crawl, instead of silently reporting an empty host
- Host-level issues (connection failures) block results; file-level issues are counted in skipped transcript footer, and their distinct reasons (capped at 3) are now shown instead of a bare count
- Platform-aware: relay automatically detects current platform and uses correct remote path separators

**Known out-of-scope risks:** #11364 (missing per-read inactivity timeout on the legacy SSH stream reader) is still required as a companion; #11395 remains open for the legacy fallback path, and #11396 remains open for both relay and legacy paths because both enforce the 10 MiB cap.


AmethystLiang:
@coderabbitai[bot] resvoled

AmethystLiang:
@coderabbitai[bot] fixed all comments

PR #11011, opened 2026-07-27
Bug korean ibus input still duplicated or droppe

### LINKED ISSUE #9862: [Bug][Linux-only]: Korean/IBus input still duplicated or dropped in the integrated terminal on v1.4.149 (regression of #8861)
**Operating system:** Linux — Ubuntu 24.04.4 LTS (GNOME Shell 46.0, X11 session, kernel 7.0.0-28-generic)  <br>**Orca version:** v1.4.149  <br>**Electron / Chrome:** Electron 43.1.0 / Chrome 150.0.7871.47  <br>**Input method:** IBus 1.5.29 with `ibus-engine-hangul` 1.5.5 (Korean). Env: `GTK_IM_MODULE=ibus`, `QT_IM_MODULE=ibus`, `XMODIFIERS=@im=ibus`, `LANG=ko_KR.UTF-8`. X11 layout `kr` (pc105/kr104), Dubeolsik, `preedit-mode=syllable`.

### Short summary

Korean (Hangul) input in the integrated terminal is still **intermittently duplicated or dropped** on v1.4.149, after stablyai/orca#8861 was closed as completed on 2026-07-18.

This looks like a regression or an incomplete fix for stablyai/orca#8861 (P0, Linux/IBus). The reported version there was v1.4.137; this reproduces on v1.4.149.

### What happened?

Typing Korean in the integrated terminal produces **whole syllables duplicated**, and occasionally **swallowed** characters. Verbatim samples captured while typing normally (intended text on the right):

| Actually produced | Intended |
| -- | -- |
| `알겠어어` | `알겠어` |
| `테스스트를 하하고 있는데 여여전히 그그러네` | `테스트를 하고 있는데 여전히 그러네` |
| `중복입력이 된다다거나` | `중복입력이 된다거나` |
| `버그가가 있어. 이것도도` | `버그가 있어. 이것도` |
| `docker 가가동 상태 버 가가동동상태 확인` | `docker 가동 상태 서버 가동상태 확인` |

Two things worth noting about the shape of the corruption:

1. The duplicated unit is a **complete composed syllable**, not a single jamo. `스` → `스스` means both `ㅅ` and `ㅡ` were re-emitted together, i.e. the *committed* composition string is delivered twice — this is not hardware key chatter or key auto-repeat.
2. Dropping also occurs in the same stream (`서버` → `버` in the last row), so commit and drop appear to be two faces of the same race around composition commit.

Plain ASCII typing is unaffected.

### How can we reproduce it?

1. Linux (X11) with IBus + `ibus-engine-hangul`.
2. Focus the integrated terminal, switch to Korean input.
3. Type Korean text at a normal-to-fast pace, ideally a few sentences.
4. → Syllables are intermittently duplicated, and occasionally dropped.

Timing dependent, so it may take a few lines before it shows. It reproduced within one or two sentences for me, fairly reliably.

### Scoping — this is Orca-specific

Korean input behaves correctly in every other application on the same machine and the same IBus session (GNOME Terminal, browser, GTK apps). Only Orca exhibits the duplication. That rules out the host IME configuration as the cause.

Additionally, the following were checked and eliminated as causes:

* `org.freedesktop.ibus.engine.hangul use-event-forwarding` set to `false` and `ibus restart` — **no effect**, still reproduces. (Reverting; noting it so nobody re-suggests it.)
* Key auto-repeat is at defaults (`delay=500ms`, `repeat-interval=30ms`) — not a repeat-rate artifact, and ASCII is clean anyway.
* Exactly one user-session `ibus-daemon` is running (`ibus-daemon --daemonize --xim`); no duplicate daemon delivering events twice.

### Notes

Same class as the known xterm.js CJK/IME composition issues (e.g. [microsoft/vscode#24633](<https://github.com/microsoft/vscode/issues/24633>)) referenced in stablyai/orca#8861 — the `compositionstart` / `compositionupdate` / `compositionend` path versus the `keydown` path both delivering the committed text.

Happy to capture composition event traces from the renderer, or test a build, if that helps isolate it.

Related: stablyai/orca#8861 (this issue), stablyai/orca#8860, stablyai/orca#8038.

AmethystLiang:
> [!CAUTION]
> **This PR must be further verified on an independent Linux/X11 machine with a real IBus Hangul input method before merge.** The current implementation is an experimental, best-effort reconciliation queue/loop that repeatedly tries to merge asynchronous composition, keypress, `input`, and textarea observations. The same-host results are encouraging, but we are not yet sure that this model is generally valid for every native IME event order.
> It can also cause regression in other language (e.g. Chinese, Japanese) and regression on other platforms (macOS, Windows)

## Summary

Fixes #9862.

The investigation localized the corruption to xterm's composition reconciliation:

- On the failing build, DOM `composition*`/`input(insertText)` commits were correct.
- `terminal.onData()` already contained the same dropped or duplicated text later observed at the PTY, so Orca's PTY forwarding was not the failing boundary.
- Observed IBus sequences included a deferred finalizer seeing stale textarea state, a final consonant moving into the next syllable without a separate `insertText`, and both the timer and `input(insertText)` path emitting the same commit.

The xterm dependency patch now:

- keeps FIFO records for pending composition finalizations instead of one global sending flag;
- scopes deferred work to the composition transaction that created it;
- reconciles composition, keypress, `input`, and textarea observations before emitting;
- preserves final-consonant transfer and avoids suppressing legitimate repeated commits.

This PR also adds:

- focused xterm composition regression coverage;
- deterministic Electron -> `terminal.onData()` -> PTY exact-byte scenarios;
- synthetic repeated Japanese and Chinese conversion coverage;
- an isolated Linux/X11/IBus Hangul native runner using the issue's `xdotool` sequences;
- a dedicated pull-request/scheduled workflow that archives boundary evidence.

The underlying xterm work was also ported upstream as xtermjs/xterm.js#6060.

## What has been verified

The Codex session records contain the following same-host results from Ubuntu/X11 with IBus 1.5.29, ibus-hangul 1.5.5, and Electron 43.1.0:

| Scenario | Baseline Orca | Patched Orca | GNOME Terminal control |
| --- | ---: | ---: | ---: |
| `한abc글` exact bytes | 0/30 | 30/30 | 30/30 |
| Full Korean sentence stress case | 0/30 | 30/30 | 30/30 |

For the patched exact-byte run:

- the DOM recorded the expected Hangul commits;
- `terminal.onData()` emitted exactly 30 copies of `한abc글\r`;
- the PTY received `ed959c616263eab8800a` on all 30 trials.

The recorded automated checks also report:

- focused Vitest: 12/12 passed;
- changed-file oxlint passed;
- web TypeScript check passed;
- max-lines ratchet passed with no bypass;
- `git diff --check` was clean;
- deterministic scenarios passed for retained/propagated Hangul commits, rapid boundaries, overlapping observations, stale timers/finalizers, no-keydown insertion, screen-reader suffixes, final-consonant transfer, and repeated Japanese/Chinese conversions.

## What is not yet verified

- The current PR has not yet produced an independent native Linux/IBus result on a separate machine. At the time this description was written, the visible GitHub check only covered community-PR tracking.
- The overlap-merging and pending-finalization loop is heuristic. Passing the observed event sequences does not prove that it handles every Chromium/IBus timing or candidate shape.
- Native Japanese, Simplified/Traditional Chinese, non-two-set Hangul layouts, macOS IMEs, and Windows IMEs were not exercised.
- macOS, Windows, WSL, SSH, and Windows-over-SSH paths received deterministic/static review only, not native end-to-end validation.
- A separate adversarial review found no concrete cross-language, platform, queue-growth, or ordinary-input performance regression, but its final signed synthesis stalled twice in the remote model service. We should not represent that as a completed independent approval.

## Testing

The dedicated workflow is intended to run, in order:

1. deterministic terminal IME boundary tests under Electron/Xvfb;
2. native IBus Hangul exact-byte and sentence tests in an isolated X11/D-Bus session;
3. evidence upload from `test-results/`.

Before merge, please run the native suite on a separate Linux/X11 machine and inspect the DOM, `terminal.onData()`, and PTY evidence rather than accepting retries alone as proof.

## AI Review Report

Codex was used for reproduction, boundary tracing, implementation, test generation, and an adversarial static/test review. The evidence above separates observed native results from synthetic tests and static reasoning; it does not claim universal IME correctness.

## Security Audit

The native runner starts its own Xvfb/D-Bus/IBus process group, records owned PIDs, and terminates only that group. The workflow has read-only repository permissions. Automated review comments about workflow credential persistence and runner hardening still need to be resolved separately.

## Notes

This should be treated as an experimental fix with useful regression infrastructure, not as a fully validated general solution. Independent Linux verification is the next required confidence step.


nwparker:
Closing as superseded — this landed, via a different commit.

`fe6f929c6e3` (#11293) rebuilt the vendored `CompositionHelper` around exactly the mechanism you proposed: transaction ids, a deferred pending-composition finalizer, and removal of already-sent data. Every private identifier your patch introduced (`_compositionTransactionId`, `_pendingComposition`, `_getPendingTextareaInput`, `_removeAlreadySentData`, …) is on main today, in a strict superset of your version — 334 added lines against your 157, with only local helper extraction unique to yours.

Your e2e half landed too: `terminal-ime-boundary-probe.ts` and `terminal-ime-observed-event-sequences.ts` are byte-identical on main, and the ibus-hangul runner, byte reader and IME workflow all exist with further-evolved content.

Thanks — this was a well-diagnosed fix and it shaped what shipped. One caveat worth naming: #9862 is still open, so the Linux side may not be fully closed. If you can still reproduce duplication or drops on a current build, that issue is the place — it would be a new mechanism rather than this one.

PR #11013, opened 2026-07-27
fix(panes): stop a stray touch hijacking a mouse divider drag

brennanb2025:
## Problem

`isActiveDragPointer` in `src/renderer/src/lib/pane-manager/pane-divider-drag.ts` accepted **any** primary pointer as the active divider-drag pointer:

```ts
e.pointerId === activePointerId || e.isPrimary
```

Per the Pointer Events spec, **each pointer type has its own primary pointer**. From MDN (`PointerEvent.isPrimary`):

> "When two or more pointer device types are being used concurrently, multiple pointers (one for each `pointerType`) are considered primary. For example, a touch contact and a mouse cursor moved simultaneously will produce pointers that are both considered primary."

So on a touchscreen, a stray finger landing during an in-flight **mouse** divider drag satisfies `e.isPrimary` and is accepted as the active drag pointer. All three consumers are `window` capture-phase listeners, so the finger does not even need to touch the divider:

- `onPointerMove` — the finger's coordinates drive the divider, moving it somewhere the user never dragged
- `onPointerUp` — `finishActiveDrag(true)` **commits** a pane size the user never chose
- `onPointerCancel` — `finishActiveDrag(false)` **reverts** the layout mid-drag

## Fix

One line — exclude touch from the fallback:

```ts
e.pointerId === activePointerId || (e.isPrimary && e.pointerType !== 'touch')
```

**The `isPrimary` fallback itself is preserved and still required.** It exists because WSLg's RDP input path reports press/release as a `mouse` pointer but streams motion as a `pen` pointer with a different `pointerId`, so a strict `pointerId` match drops every move. That case is mouse+pen only and is unaffected. The predicate was simply broader than the pointer types it was scoped to.

**Touch-initiated drags keep working — verified, not assumed.** When a drag starts from touch, `onPointerDown` stores that pointer in `activePointerId`, so every subsequent move/up from that same finger matches the first clause by `pointerId` and never reaches the `isPrimary` fallback. The new `drives a touch-started drag from its own pointerId` test covers this explicitly, and it passes both before and after the change.

This is a pointer**Type** distinction, not an OS one — no platform check is added.

## Tests

New file `pane-divider-stray-touch.test.ts`, following the existing split-out `pane-divider-capture-loss.test.ts` harness pattern (`pane-divider.test.ts` is already 575 lines against the 800-line test cap).

Revert-check via the patch method — source reverted, tests untouched:

**Before the fix (3 failed | 2 passed):**
```
× ignores stray touch motion during an active mouse drag
    AssertionError: expected '320 1 0%' to be '180 1 0%'
× does not commit the layout when a stray touch lifts mid mouse drag
× does not revert the layout when a stray touch is cancelled mid mouse drag
    AssertionError: expected '2 1 0%' to be '180 1 0%'
✓ still continues a mouse drag from a primary pen pointer (WSLg relay)
✓ drives a touch-started drag from its own pointerId
```

**After the fix (5 passed):** all five green.

The two tests that pass on both sides are deliberate **no-regression guards for unchanged behavior**, not bug reproducers — the WSLg pen relay and the touch-started drag must keep working, and they do. Only the three stray-touch tests encode new behavior.

Full `src/renderer/src/lib/pane-manager/` suite: 59 files, 616 passed / 1 skipped. `pnpm typecheck` exit 0, `pnpm lint` exit 0.

## Verification limitation

Everything above is **static reasoning plus unit tests against a mocked pointer harness**. I have no touchscreen and no WSLg session in this environment, so neither the real stray-finger hijack nor the real WSLg pen relay was exercised on physical hardware. The spec behavior is confirmed from MDN; the event sequences are simulated, not captured from a device.

## Scope

One concern. The flex math, the PTY resize hold, and the double-click handler are untouched.

### DIFF
--- a/src/renderer/src/lib/pane-manager/pane-divider-drag.ts
+++ b/src/renderer/src/lib/pane-manager/pane-divider-drag.ts
@@ -220,9 +220,12 @@ export function attachDividerDrag(
 
   // Why: WSLg's RDP input path reports press/release as a `mouse` pointer but
   // streams motion as a `pen` pointer with a different pointerId, so a strict
-  // pointerId match drops every move. Any primary pointer continues the drag.
+  // pointerId match drops every move. Any primary pointer continues the drag,
+  // except touch: each pointer type has its own primary, so a stray finger
+  // would otherwise hijack an in-flight mouse/pen drag. A touch-started drag
+  // still matches by pointerId.
   const isActiveDragPointer = (e: PointerEvent): boolean =>
-    e.pointerId === activePointerId || e.isPrimary
+    e.pointerId === activePointerId || (e.isPrimary && e.pointerType !== 'touch')
 
   const onPointerMove = (e: PointerEvent): void => {
     if (!dragging || !isActiveDragPointer(e) || !prevEl || !nextEl) {
--- a/src/renderer/src/lib/pane-manager/pane-divider-stray-touch.test.ts
+++ b/src/renderer/src/lib/pane-manager/pane-divider-stray-touch.test.ts
@@ -1,0 +1,208 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { createDivider } from './pane-divider'
+
+type PaneElement = HTMLElement & { style: Record<string, string> }
+
+type DividerDragHarness = {
+  divider: HTMLElement
+  dividerListeners: Map<string, EventListener>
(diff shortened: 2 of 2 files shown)


brennanb2025:
@coderabbitai review

brennanb2025:
@coderabbitai full review

PR #11029, opened 2026-07-28
fix(e2e): stabilize release gating

nwparker:
## Summary
- keep the native-chat first-flush test aligned with the intentional working-hook empty state while preserving no-error and live-hydration coverage
- make the synchronized-output latch observation and worktree hide atomic, and remove an invalid occlusion-blind terminal visibility assertion
- capture E2E daemon PIDs before Electron shutdown can unlink their records, preventing detached daemon/PTY trees from hanging worker teardown

## Validation
- `pnpm run typecheck`
- targeted `oxlint` and `oxfmt --check`
- the three previously failing E2Es pass together after rebasing on main
- exact shard 4-of-10 completed without its prior 120s worker teardown timeout
- restart cleanup leaves no daemon process behind after both pass and failure paths
- localization coverage check passes

### DIFF
--- a/tests/e2e/activity-agent-pane-isolation.spec.ts
+++ b/tests/e2e/activity-agent-pane-isolation.spec.ts
@@ -412,7 +412,7 @@ test.describe('Activity Agent Pane Isolation', () => {
         activeTabType: 'editor',
         activeTabId: snapshot.tabId
       })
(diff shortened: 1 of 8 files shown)


nwparker:
Closing to avoid duplicating the E2E stabilization work already assigned to another agent. The branch remains available for reference if useful.

PR #11031, opened 2026-07-28
ci(release): detach non-blocking E2E

nwparker:
## Summary

- Remove full E2E from the `Cut Release` workflow's in-line job graph.
- Dispatch the same tag-scoped E2E asynchronously after `publish-release` succeeds.
- Run the detached workflow definition and tests from the exact release tag.
- Retry dispatch failures three times, then warn without turning an already-published release red.
- Give detached E2E runs a ref-specific name and protect ordering, permission, tag, retry, and input contracts with tests.

The release-blocking topology is unchanged: publication still requires platform builds, signing, asset verification, and terminal-rendering golden checks. This removes roughly 145 runner-minutes from each `Cut Release` run and prevents flaky diagnostic E2E failures from making successfully published releases appear failed.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — focused workflow suites passed locally; full sharded suite runs in PR CI
- [ ] `pnpm build` — PR CI's package job builds and smoke-tests the packaged app
- [x] Added a release E2E dispatch contract suite; 26 focused release-contract tests pass
- [x] Parsed both changed YAML workflows and ran `git diff --check`

## AI Review Report

Reviewed the release authority graph, publish ordering, failure behavior, retries, concurrency, tag reproducibility, and GitHub token scope. Review flagged that retry behavior should be explicitly contractual; added assertions for the three-attempt loop and retry delay. It also flagged a possible draft-creation race, but verified this is prevented transitively: both `build` and `build-mac` need `create-release`, while `publish-release` needs both build jobs.

Cross-platform review confirmed the dispatcher intentionally runs on Ubuntu and changes no application shortcuts, labels, filesystem paths, native packaging, SSH behavior, or Electron platform logic. Existing macOS, Linux, Windows, SSH, signing, and terminal golden release jobs are unchanged.

## Security Audit

- The dispatcher receives only `actions: write`; unspecified permissions remain disabled.
- The validated release tag is passed through an environment variable and quoted in every shell argument.
- The workflow definition is selected from the exact immutable release tag, avoiding a race with later `main` changes.
- Dispatch uses the repository-scoped `GITHUB_TOKEN`; no new secret or third-party action is introduced.
- A failed diagnostic dispatch retries, emits an Actions warning, and exits successfully because it occurs after publication and must not rewrite release status.

## Notes

ELI5: the release truck used to carry an optional inspection team, so a flaky inspector could paint the delivery red even after it shipped. Now every required safety check completes and the package publishes first, then the optional team inspects that exact package in a separate lane.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,5 +1,7 @@
 name: E2E
 
+run-name: E2E ${{ inputs.ref || github.ref }}
+
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -777,7 +777,7 @@ jobs:
             echo "## Release E2E Signal"
             echo ""
             echo "- Terminal rendering golden is release-blocking."
-            echo "- Full E2E is diagnostic/non-blocking release evidence."
+            echo "- Full E2E runs separately after publication and cannot change the release result."
             echo "- Terminal rendering release evidence is diagnostic/non-blocking."
             echo ""
             echo "Publishing behavior is controlled by the existing job dependencies; this summary does not change release gating."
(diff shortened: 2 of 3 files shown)


nwparker:
Review follow-up:

- Added explicit contract assertions for the three-attempt dispatch retry and delay in `32d1c02110`.
- Verified the reported draft-creation race is not present: `build` and `build-mac` both directly need `create-release`, and `publish-release` directly needs both build jobs, so GitHub cannot start publication before draft creation completes.

PR #11033, opened 2026-07-28
fix(skills): stop a bounded plugin-cache walk from flagging every skill (#10775)

### LINKED ISSUE #10775: [Bug] Orchestration and Computer use [Failing to Complete
<kbd>[vex1](https://discord.com/channels/1492228674081263786/1531031210787344586/1531031210787344586)</kbd> [vex1](https://discord.com/channels/1492228674081263786/1531031210787344586/1531031210787344586)  `via Discord`

I've been trying to follow the onboarding checklist, but the Orchestration and Computer use denies to work.

I tried to reset computer, re-do the steps, check but nothing works.

OrcaWin:
Fixes #10775  ## Summary  On a machine with the Codex CLI and its bundled plugins, **every** Orca skill showed the amber "Needs attention" pill — including skills that were not installed — while the "Details" dialog said *"All installed Orca skills are up to date."* Users reasonably concluded their install was broken and went looking for a problem that did not exist.  Nothing was wrong with the installs. Three separate defects stacked up:  **1. A budget stop was treated as evidence.** `scanKnownPluginSkillCandidates` reported depth cutoffs, pruned subtrees, unreadable directories and exhausted budgets through a single `incompletePaths` channel, and the inventory poisoned every official skill name once for each entry. Codex plugins vendor npm dependency trees (`.../<plugin>/<version>/scripts/node_modules/classic-level/deps/leveldb/leveldb-1.20/db`) that reach past the 9-level budget, so an ordinary, unavoidable directory shape flipped all eight skills amber.  The scan now reports two things instead of one:  - `unverifiedPaths` — a directory it could not read, an arbitrary depth cutoff, or a global budget blown mid-walk. These still poison every name, because they genuinely mean *the scan does not know*. - `prunedPaths` — vendored/VCS directories. The scan's own choice over ground that cannot hold an installed skill package, so it proves nothing was missed.  It also prunes `node_modules`, `.git`, `.hg` and `.svn` rather than spending depth budget walking into them.  **2. The badge and the dialog derived "is this a problem" separately.** The badge flagged `status !== 'current'`; the review dialog only built a group when a copy was `status === 'outdated'`. A skill could therefore be amber with a dialog that had nothing to show — exactly the contradiction in the report's screenshot. Both now derive from one predicate in `src/shared/skill-freshness.ts`, alongside eligibility.  **3. A name collision degraded a healthy install.** Codex ships its own plugin called `computer-use`. Orca matched the directory name, failed the digest check, and marked its own correctly-installed copy `unrecognized`. A plugin-cache package matching no Orca release is now ignored: plugin caches are vendor-managed and Orca never writes there, so that is someone else's skill, not drift in ours. An *outdated* Orca skill a plugin vendors is still reported, since "update the plugin instead" is real advice.  ## Screenshots  No visual change. The affected UI is existing components rendering correct data — the badge stops being amber, and when it is amber the existing dialog now has rows to show.  ## Testing  - [x] `pnpm lint` — clean except `verify:skill-bundle-manifest`, which fails identically on a stashed clean tree (pre-existing, unrelated) - [x] `pnpm typecheck` - [x] `pnpm test` — 3190 passing. One pre-existing failure, `discovery.test.ts` "records every contributing root when symlinked roots dedup to one skill", is `EPERM: operation not permitted, symlink` — a Windows dev-mode privilege limitation on the dev box, confirmed failing on a clean tree too. - [x] `pnpm build` - [x] Added or updated high-quality tests  Reproduced first, against the real shipped `resources/skills/current-manifest.json`, with the reporter's exact cache shape (both vendored `classic-level` trees, the `sites` template tree, and the colliding `computer-use` plugin): **8 poisoned rows → 0**, every skill clean including `computer-use`.  New coverage:  - scan: an arbitrary depth cutoff remains unverified, while vendored trees classify as pruned; an unreadable cache root still classifies as unverified; a missing cache reads as absence - inventory: a deep vendored tree leaves an installed skill `current`; an unreadable cache still poisons; a foreign same-named plugin skill is ignored; an outdated Orca copy a plugin vendors is still reported - display: across `outdated`, `newer-known`, `unrecognized` and `inaccessible`, the badge and the review dialog can never disagree  `skill-plugin-cache-scan.test.ts` previously asserted the poisoning was intended behaviour. **That test is deliberately inverted here** — it is a policy reversal, not an incidental update, and worth a reviewer's attention. Its old name also claimed truncation poisoned *eligibility*, which was never true: eligibility only ever considered convergent topologies, so this only ever affected the badge.  ## AI Review Report  Reviewed for root cause over symptom, duplicate derivations, platform behaviour, and false-clean risk.  - **Cross-platform (macOS/Linux/Windows):** the change is `node:fs` and `node:path` only — no shell, no separators, no shortcuts or labels. New tests avoid `chmod` (a no-op on Windows) and force a read failure by opening a file as a directory, which fails on all three. The pre-existing symlink test remains `runIf(platform !== 'win32')` where the repo already gates it. - **False-clean risk (the main one flagged):** every change here makes Orca *less* likely to report a problem, so the risk is hiding real drift. Mitigations: genuine `EACCES`/`EPERM`/`EIO` and mid-walk budget exhaustion still poison every name; the prune list is limited to four directories that definitionally cannot hold installed first-party content. I deliberately did **not** prune `dist`/`build`/`out` — a bundler-built plugin could legitimately ship `dist/skills/…`, and pruning that would convert a real placement into a silent false clean. - **Rejected during review:** replacing the walk with a structural `<namespace>/<plugin>/<version>/skills/` matcher. It is tempting (real caches put every `SKILL.md` at exactly that depth) but once a depth stop no longer poisons, a blind walk's failure mode is a false clean — and a structural matcher would produce exactly that for any plugin with a different layout. The conservative walk is the better design, not just the safer one. - **Scope:** the foreign-package rule is confined to `plugin-cache`. It was **not** extended to `repo-scope`, where an unrecognized copy is plausibly the user's own edit and hiding it would suppress real drift. - **Duplicate derivations:** swept the codebase for other re-derivations of the same concept. Three that look similar were left alone on purpose — chip labelling picks a chip per specific status, eligibility asks a write-safety question, and the update nudge filters the *offerable* set. Those are different questions, not restatements of drift.  ## Security Audit  Low risk; no new attack surface.  - **Path handling:** no new path construction from user input. Scanning stays rooted at `~/.codex/plugins/cache` with the existing `realpath` + `visited` guard against symlink loops. Pruning strictly *reduces* the traversed set. - **Command execution / IPC / auth / secrets / dependencies:** none touched. No new dependencies. The IPC payload shape (`SkillFreshnessInventory`) is unchanged, so no preload or renderer contract moves. - **Denial of service:** entry, candidate and unverified-path caps are unchanged. The new `prunedPaths` set is separately capped so a hostile cache cannot grow it without bound, and — unlike the unverified path — recording one deliberately does not abort the scan. - **Follow-up:** none identified.  ## Notes  - Reported on Windows 11, but the cause is platform-independent: any OS with the Codex CLI and its bundled plugins hits it. Verified on Windows. - `errorCategory` renamed `plugin-cache-scan-incomplete` → `plugin-cache-scan-unverified`. Diagnostic only — it is not rendered, localized, or matched anywhere else. - The reporter did the root-cause analysis themselves and it was accurate; the depth arithmetic is theirs, independently confirmed here against the published `classic-level` package. - Behaviour change worth noting: `newer-known` copies now appear in the review dialog. The badge already flagged them while the dialog stayed silent — same defect class, fixed with the rest.


possibilities:
I tested `5714e2daa` locally on macOS against a real Codex plugin cache. The build and targeted tests pass, but this head still reports Orchestration and Computer Use as **Needs attention**.

The inventory contains a current canonical copy plus three synthesized `plugin-cache-scan-unverified` placements per skill under:

- `.../sites-building/templates/vinext-starter/examples/d1/app/<skill>`
- `.../sites-building/templates/vinext-starter/examples/d1/db/<skill>`
- `.../presentations/.../assets/previews/<skill>`

The suffixed skill paths do not exist. The dialog therefore lists every bundled skill as **Skipped** / “could not read this copy.” Pruning `node_modules` and VCS trees does not cover these other deep non-skill trees.

A build containing merged #10865 reports only the current canonical copies and no scan issues on the same machine. Since this branch is also conflicting with `main`, it appears superseded by #10865.

brennanb2025:
Closing as superseded by #10865, which landed 28 Jul and carried the same diagnosis further.

Your analysis was correct, and it was correct before #10865 existed — this is a timing loss, not a quality one. Two reasons a rebase isn't worth it now:

- Main replaced `incompletePaths` with a typed `issues[]` model — seven reasons partitioned into attention / truncating / display-only — which is a strict superset of the two buckets here.
- The single shared badge/dialog predicate you argued for exists as `isSkillCopyNeedingAttention` in `src/shared/skill-freshness.ts`, landed via #11128.

The blocker for a straight rebase is this branch's own stated behavior: unverified paths "still poison every name". That is precisely the mechanism reported in #10659 — a directory Orca declined to open becoming a concrete-looking placement per manifest skill. #10865 removed the per-skill fabrication entirely, so merging this now would reintroduce it.

The `node_modules` prune you identified did land. Thanks for the work here.

PR #11047, opened 2026-07-28
fix(worktree): give the physical-stop recheck a budget it can actually use

OrcaWin:
## Problem

Deleting a worktree that holds an idle agent fails on Windows with:

```
Failed to delete workspace test
Error: Failed to physically stop every PTY for worktree: a4affed8-...::C:/Users/neil/orca/workspaces/orca/test
```

Every PTY named in that error was already dead.

## Root cause (measured, not inferred)

I instrumented a dev build on native Windows and captured five worktree deletes with idle Claude Code agents. Findings:

**The executing path is the terminal daemon**, not the in-process local provider — `src/main/daemon/terminal-session-teardown.ts` (`daemon-agent`). `local-pty-provider.ts` never ran during any delete. Anything fixed only there would not engage.

**Per-teardown cost on Windows** (each burst, measured):

| stage | observed range |
|---|---|
| root-identity process query | 0.73s – 3.62s |
| forced tree kill | 0.69s – 3.01s |
| whole sweep | 1.96s – 6.56s (budget 10s) |

The identity query hit its own 3s timeout in 2 of 5 runs and resolved `unknown`. `unknown` takes the same branch as `own` — the tree kill still runs — so on timeout that query costs up to 3.6s and changes nothing.

**The actual defect** is downstream. When a stop RPC exhausts its bound *after* the process already died, the PTY lands in `failedPtyIds`, and `verifyFailedPtysExited` (added in #10106) is supposed to absolve it. But that recheck was scoped to the same sweep deadline the stop phase had just spent, so it started with less time than one provider round-trip. `settleBeforeDeadline` returned its `null` fallback **without ever issuing `listProcesses`**, and `null` is read as "still running":

```ts
const sessions = await settleBeforeDeadline(
  () => provider.listProcesses({ deadlineMs: deadline }),
  null,
  deadline            // ← already spent by the stop phase
).catch(() => null)
if (!sessions) {
  return false        // ← "cannot tell" scored as "still alive"
}
```

So the recheck failed closed in exactly the situation it was written to rescue. That is why #10106 did not hold.

## Fix

One concern: give the recheck a bounded window measured from when the stop phase ends, so it can ask before it answers. Still hard-bounded, so destructive removal cannot hang (#9516). When a PTY genuinely is alive, `listProcesses` still lists it and the sweep still fails closed — the safety property is unchanged.

## Test

`does not fail destructive teardown when the physical-exit recheck outlives the reserved RPC margin` models the real shape: a bounded kill RPC that rejects at its deadline having already terminated the process, and a recheck round-trip slightly longer than the reserved margin.

Revert-verified — with `worktree-teardown.ts` reverted the test fails with the exact production string at the exact throw site:

```
AssertionError: promise rejected "Error: Failed to physically stop every PT…" instead of resolving
Caused by: Error: Failed to physically stop every PTY for worktree: w1
```

With the fix: 27/27 in the file, 87/87 across the teardown suites.

## Live verification

Instrumented dev build on native Windows, 4 idle Claude Code agents in a real worktree with UI tabs, under emulated steady-state process-scan load: delete succeeded in 9.5s, all stops OK, worktree directory gone, zero surviving processes.

## Scope notes

- Not local-worktree-specific: the change is in the provider-agnostic sweep, so SSH and folder-workspace removal get the same corrected recheck. No path assumptions added.
- Renderer respawn / silent-toast behaviour deliberately untouched — owned separately.

## Follow-ups (deliberately not in this PR)

1. **The identity query may not be worth its cost.** Comparable terminal implementations either fire the forced tree kill unconditionally and let the OS no-op a dead PID, or hold a real process handle (which cannot be recycled) instead of a bare PID. Orca gates on a whole-machine process-table scan, which is the most expensive of the three and returns `unknown` under load. Worth revisiting whether a narrower console-scoped enumeration or a retained handle would be both cheaper and stronger.
2. **`stopTerminalsForWorktree` stops PTYs in a sequential `for await` loop** while the provider/registry sweeps use `mapWithConcurrency(32)`. It contributed nothing in my runs (the worktree had already left the runtime graph, so it exited early every time), but it is a latency cliff if it ever does own the PTYs.

## Honest limitation

I could not make the delete fail live on the dev build — my runs peaked at 6.6s against a 10s budget, short of the cliff. The failure is reproduced deterministically in the unit test, which throws the identical error from the identical line. The live measurements establish the cost model that gets a loaded machine to that cliff; they do not themselves cross it.

### DIFF
--- a/src/main/runtime/worktree-teardown.test.ts
+++ b/src/main/runtime/worktree-teardown.test.ts
@@ -8,7 +8,11 @@ vi.mock('../memory/pty-registry', () => ({
   listRegisteredPtys: listRegisteredPtysMock
 }))
 
-import { killAllProcessesForWorktree, WORKTREE_PROCESS_SWEEP_TIMEOUT_MS } from './worktree-teardown'
+import {
+  killAllProcessesForWorktree,
+  WORKTREE_PROCESS_SWEEP_TIMEOUT_MS,
+  WORKTREE_TEARDOWN_RPC_MARGIN_MS
+} from './worktree-teardown'
 import type { IPtyProvider, PtyProcessInfo } from '../providers/types'
 import { DaemonPtyAdapter } from '../daemon/daemon-pty-adapter'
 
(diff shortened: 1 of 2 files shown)


nwparker:
Closing — but this PR's **finding** was the most valuable thing in this whole cluster, and it was wrong to leave it sitting here.

> The executing path is the terminal daemon, not the in-process local provider — `src/main/daemon/terminal-session-teardown.ts` (`daemon-agent`). `local-pty-provider.ts` never ran during any delete. Anything fixed only there would not engage.

That is measured, and it was still true after #15755 landed. W2 gave every ConPTY pane a Win32 job object and wired the job kill into `local-pty-provider.ts` — the path you showed does not execute. Three daemon `killWithDescendantSweep` call sites had no job at all, including both in `terminal-session-teardown.ts`. On the delete path the sweep was still falling back to a parent-pid walk, which a detached, reparented grandchild is not in.

Fixed in **#15886**, with you credited as co-author. That PR also adds a tree-walking guard so a future call site cannot omit the job the way W2's did.

The **code** in this PR — giving the physical-stop recheck a budget measured from itself rather than the sweep's leftovers — has since landed on `main` independently via #11960 (`verifyUnstoppedPtys` in `src/main/runtime/unstopped-pty-verification.ts` now takes `Math.max(2s, sweepBudgetMs)` from `Date.now()`). Same fix, same reasoning. So there is nothing left here to merge.

Thanks — this one changed what we shipped.

nwparker:
Superseded: the code fix landed via #11960; the path finding is fixed in #15886 with credit.

nwparker:

Issue #9976, opened 2026-07-22
[Bug]: slow Claude startup lets tui-idle false-positive and loses injected dispatch (v1.4.149)


shaswatshah:
### Operating system

macOS

### Orca version

v1.4.149 (bundle id `com.stablyai.orca`)

### Summary

`orca terminal wait --for tui-idle` can report a newly launched Claude terminal as ready before Orca has observed its first Claude agent-status signal. A subsequent `orca orchestration dispatch --inject` then returns `injected: true` and commits an active dispatch lease, even though Claude never consumes the prompt.

This happened to one of four Claude workers launched together. The affected worker remained at an empty prompt and its Claude transcript contained no user turn from the original dispatch. Sending the exact task spec later with `orca terminal send --enter` immediately started the same worker.

This is a Claude recurrence of the visible symptom in stablyai/orca#7748, after its fix in [#7758](<https://github.com/stablyai/orca/issues/7758>), and is adjacent to the current Grok report in stablyai/orca#9838. The additional evidence here points specifically to a false-positive `tui-idle` readiness result during slow startup, combined with an injection acknowledgment that confirms PTY writes rather than prompt consumption.

### Controlled incident

One coordinator launched four read-only Claude workers through the same wrapper and worktree at the same Git SHA. All four followed:

```text
terminal create
  -> terminal wait --for tui-idle
  -> orchestration task-create
  -> orchestration dispatch --inject
```

All dispatch calls returned matching task/assignee identities and `injected: true`. Three workers began normally. One did not.

The failed worker and a successful sibling used the same:

* Claude runtime, Opus model, and high effort
* permission/profile configuration
* wrapper implementation and caller
* 4,632-character / 85-line injected preamble
* 3,767-character router brief
* worktree and Git SHA

The failed worker's persisted dispatch had `failure_count = 0` and no `last_failure`.

### Timeline and evidence

For the failed terminal:

* 18:36:40: dispatch `ctx_3f72fb5cd72e` was persisted and returned `injected: true`.
* 18:36:42.229: Orca telemetry recorded that terminal's first Claude `agent_start` signal.
* The Claude transcript had no original injected user turn at all. Its first user turn appeared only at 18:43:03, when the coordinator used direct `terminal send --enter` for recovery.
* The same terminal then read the original file-backed brief, launched its child verifier, wrote its report, and completed normally.

[...]

PR #11048, opened 2026-07-28
fix(orchestration): wait for dispatch readiness

### LINKED ISSUE #10663: [Bug]: orchestration check --wait fails to surface queued typed messages; --types vocabulary undocumented
**Observed 2026-07-25 (two independent instances, same host) during live multi-agent supervision:**

Instance 1: six \`worker_done\` messages were demonstrably queued (full payloads later delivered intact) while the intended consumer sat inside \`orca orchestration check --wait --types worker_done,escalation,decision_gate\` cycles that never surfaced them — they were delivered only at a much later turn boundary.

Instance 2 (later session): multiple \`worker_done\` messages from completed dispatches surfaced ONLY as turn-boundary injections into the coordinator terminal, while a concurrently-running \`check --wait --types worker_done,escalation\` (15-minute timeout, repeatedly re-armed) returned empty every cycle.

Emission was correct in both cases; the delivery/consumption seam under-delivers during an active wait that explicitly lists the queued type. If \`--wait\` is expected to return promptly when a matching typed message is already queued (or arrives mid-wait), this is a bug; if waits only observe new arrivals under some conditions, the semantics deserve documentation.

**Asks:**
1. Make \`check --wait --types T\` return promptly for messages of type T queued at wait-start or arriving mid-wait; or document the actual delivery contract.
2. Minor: \`--types worker_done,escalation,ask\` errors with \`Invalid --types: ask\` — the valid \`--types\` vocabulary is undocumented. Enumerate accepted type names in \`check --help\`, or accept-and-ignore unknown types so supervision loops degrade soft.

Context: production multi-agent orchestrator (coordinator + worker fleet) on the orca CLI; happy to provide full timelines or run diagnostic builds.

### LINKED ISSUE #10728: [Bug]: dispatch --inject loses submit when agent TUI is still booting (blind 500ms submit)
### Operating system

nwparker:
## Summary

- wait for the target terminal to report `tui-idle` before dispatch authority and prompt injection
- bound server readiness at 60 seconds and the CLI call at 65 seconds
- leave blocked startup tasks ready without writing authority or prompt input
- preserve queued typed orchestration messages and document the full `check --types` help contract

## Reproduction

`dispatch --inject` previously used a blind 500 ms delay. A slow-starting agent could still be showing its directory-trust prompt, so the injected submit answered that prompt and swallowed the task. The new regression holds startup before `tui-idle` and verifies no writes occur until readiness.

## Validation

- 347 focused tests across the CLI and orchestration RPC suites
- `pnpm typecheck:node`
- `pnpm typecheck:cli`
- `pnpm check:max-lines-ratchet`
- `git diff origin/main...HEAD --check`

Closes #10728
Closes #10663


YoraiLevi:
## The gate this PR relies on can itself report ready too early

This PR replaces the blind 500 ms delay with a wait for `tui-idle` before injecting. That is the right shape. The finding below is that the wait itself returns ready while Claude is sitting on its first-launch trust screen, so the injection can still land on a pane that is not accepting a task.

Measured on Orca 1.4.176, Windows 11, Claude Code panes, brand-new folders:

- `terminal wait --for tui-idle`, run immediately after `terminal create`, returned `satisfied: true` after **3149 ms**, with "1. Yes, I trust this folder" on the pane. 4 of 4 folders.
- The same wait, run again seconds later on that same unchanged screen, returned `satisfied: false`, `blockedReason: codex-trust-workspace`.

The detector is correct. `TERMINAL_WAIT_BLOCKED_SENTINEL_RE` in `src/main/runtime/orca-runtime.ts` matches that screen. What beats it is a generic "pane has been quiet for ~3 seconds" path that resolves first during cold start.

Effect on the composed path: `orchestration-workers.ts:198` waits, `:234` writes the preamble, and nothing re-reads the pane in between. A `satisfied: true` obtained through the quiet-pane path is enough to arm the write, and the Enter then answers the trust screen rather than submitting the task.

Full evidence and the field incident are recorded on #9976. @innocarpe has since added deterministic coverage for this exact screen in #10020, which is the narrower fix; this PR remains the broader readiness and reconciliation change.

## Correction to my earlier version of this comment

I suggested giving Claude a `preflightTrust` preset alongside Codex, Cursor and Copilot. @innocarpe has explained upstream removed that bypass deliberately, so it is a product and security decision. That suggestion is withdrawn.

innocarpe:
Thanks for the deterministic evidence on the Claude trust gate. I updated #10020 with a regression covering the narrower launchAgent quiescence race from #9976: after the quiet period, the unchanged "Yes, I trust this folder" screen must produce a blocked tui-idle result rather than readiness. The upstream and fork mirror now both point to 0a0df3a643, with 1,079 runtime tests passing and 1 skipped. #11048 remains the broader readiness, cancellation, and reconciliation change; I have not modified its external branch or resolved its existing review threads.

PR #4403, opened 2026-06-01
Fix Windows second-instance window focus

Jinwoo-H:
## Summary
- extract second-instance window focusing into a testable main-process helper
- make repeat launches reopen a missing ready window and more aggressively foreground the existing window on Windows
- use a short Windows always-on-top pulse plus retry focus to handle taskbar/foreground activation cases where `BrowserWindow.focus()` alone is not enough

## Diagnosis
On the installed Windows RC, clicking Orca again did launch a second process, but it exited after failing the single-instance lock:

```text
[startup] single-instance-lock-result acquired=false
[single-instance] Another Orca instance is already running; exiting secondary process
```

The primary process was still alive and had a visible `Chrome_WidgetWin_1` window titled `Orca`, so this was not the Codex SQLite/runtime-home crash path. The bad behavior was that the primary instance did not reliably surface its existing window, making the app look like it failed to launch.

## Verification
- `pnpm vitest run --config config/vitest.config.ts src/main/window/focus-existing-window.test.ts src/main/startup/single-instance-lock.test.ts`
- `pnpm exec oxlint src/main/window/focus-existing-window.ts src/main/window/focus-existing-window.test.ts src/main/index.ts`
- `pnpm exec tsgo --noEmit -p config/tsconfig.node.json --pretty false`

Local note: these commands pass with existing machine warnings for Node v25 vs repo Node v24 and an unset `${GITHUB_TOKEN}` placeholder in `C:\Users\jinwo\.npmrc`.

PR #11055, opened 2026-07-28
fix(macos): avoid scene deadlock on app reactivation

OrcaWin:
## Summary

- avoid calling native focus APIs when macOS activates an already-live main window
- keep windowless/headless desktop promotion routed through the existing activation gate
- scope the delayed focus retry to Windows, where it was originally introduced

## Diagnosis

On macOS 26.5.1, the browser-process main thread permanently deadlocked in AppKit/FrontBoardServices while processing a server-originated fullscreen window frame update. The observed sequence was:

1. `gh auth refresh` opened the external browser.
2. The user clicked back into fullscreen Orca.
3. `app.activate` called `app.focus()`, `window.show()`, and `window.focus()`, then repeated them 100 ms later.
4. The native focus sequence overlapped the macOS Space/scene transition.

PR #8646 began routing every app activation through the aggressive second-instance focus path. That path came from the Windows-specific foregrounding work in #4403, but its delayed retry was applied on every platform.

The underlying AppKit self-deadlock is tracked upstream in electron/electron#52437. This change removes Orca's avoidable native focus calls from the matching activation path while preserving deliberate second-instance and headless-promotion behavior.

## Testing

- `pnpm exec vitest run --config config/vitest.config.ts src/main/window/focus-existing-window.test.ts src/main/startup/serve-desktop-activation.test.ts src/main/startup/serve-desktop-activation-wiring.test.ts src/main/startup/single-instance-lock.test.ts`
- `pnpm exec oxlint src/main/window/focus-existing-window.ts src/main/window/focus-existing-window.test.ts src/main/index.ts src/main/startup/serve-desktop-activation-wiring.test.ts`
- `pnpm exec oxfmt --check src/main/window/focus-existing-window.ts src/main/window/focus-existing-window.test.ts src/main/index.ts src/main/startup/serve-desktop-activation-wiring.test.ts`
- `pnpm run typecheck:node`

## AI Review Report

Reviewed the macOS activation lifecycle, second-instance behavior, headless serve promotion, and Electron window calls. Windows retains its foreground reinforcement and delayed retry. Linux keeps the immediate reveal/focus behavior without the Windows-only retry. SSH and folder workspaces are unaffected because the change is limited to local Electron window activation.

## Security Audit

No new input handling, command execution, IPC, path, authentication, or secret surfaces. The change only narrows when existing Electron focus APIs are called.

## Screenshots

No visual change.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -157,6 +157,7 @@ import {
   setTrayAttention,
   type SystemTrayOptions
 } from './tray/system-tray'
+import { createMacAppActivationHandler } from './window/macos-app-activation'
 import { focusExistingMainWindow } from './window/focus-existing-window'
 import { notifyMainWindowBecameVisible } from './window/main-window-visibility'
 import { CodexAccountService } from './codex-accounts/service'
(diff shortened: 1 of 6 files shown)


Jinwoo-H (approved):
Reviewed and validated:

- separated passive macOS reactivation from explicit second-instance foregrounding
- retained the delayed focus retry only on Windows
- added direct regression coverage for live and windowless activation
- cleared the changed-file static-analysis failure

Full lint, typecheck, focused tests, and macOS Electron activation scenarios pass.

PR #11065, opened 2026-07-28
fix(worktree): use local base when remote tracking ref is unavailable

### LINKED ISSUE #7431: [Bug]: `worktree create` hard-errors offline when base ref is a stale/never-fetched remote-tracking ref
**Short summary**

`worktree create` has fallbacks that cover the common offline case (the default-base probes fall through to local `refs/heads/main` / `master`, and a stale local remote-tracking ref is reused when the refresh fetch fails). But there is no explicit offline mode, and there is a real edge that still hard-errors: when the resolved base is a remote-tracking ref that does not exist locally (e.g. a persisted stale `repo.worktreeBaseRef`, or a branch never fetched into this clone) and the origin fetch fails.

**What happened**

In an environment where `origin` was unreachable, `worktree create` (no `--base-branch`) failed with:

    Could not refresh base ref "origin/main" from "origin". Check your network and try again.

`resolveWorktreeCreateBase` resolved the base to `origin/main` (a remote-tracking ref that was not present locally in that clone), `getOrStartRemoteTrackingBaseRefresh` ran `git fetch --no-tags origin +refs/heads/main:refs/remotes/origin/main` (`src/main/runtime/orca-runtime.ts`), the fetch failed, and the hard-error above was thrown. The documented workaround `--base-branch main` (a local ref) worked, but it is not surfaced in `worktree create --help` as the offline path, and there is no `--offline` / `--no-fetch` flag to opt out of the fetch entirely.

**How to reproduce**

1. Use a clone that does **not** have `refs/remotes/origin/main` locally (e.g. a fresh shallow clone, or a clone where the default branch was never fetched).
2. Be on a network where `origin` is unreachable.
3. `orca worktree create --name demo` (no `--base-branch`)
4.  →  `Could not refresh base ref "origin/main" from "origin".`

Note: the common offline case (clone already has `refs/remotes/origin/main`) does **not** hit this — the stale local ref is reused. The throw only fires for the stale/never-fetched remote-tracking-ref edge.

**Expected**

At least one of:
- a `--offline` / `--no-fetch` flag on `worktree create` that skips the remote refresh and uses whatever local ref is available;
- when the resolved base is `origin/<branch>` and the fetch fails, fall back to the local `refs/heads/<branch>` of the same name if it exists (today the throw fires before that fallback);
- `worktree create --help` calling out `--base-branch <local-ref>` as the offline workaround (the `orca-cli` skill currently advises omitting `--base-branch`, which is exactly what triggers the fetch).

- Orca v1.4.121, macOS
- `origin` on an internal git server, unreachable from the current network; the clone did not have `refs/remotes/origin/main` locally.
- `repo set-base-ref` is the only persistent lever today; it is not an ad-hoc per-create fallback.

nwparker:
## Summary
- Falls back from a missing remote-tracking base such as `origin/main` to the matching local branch when it exists.
- Applies the same behavior to local and SSH worktree creation while preserving slash-named local branches; SSH fallback occurs before branch reuse/conflict resolution.

## ELI5
If Orca cannot reach the internet to update its `origin/main` bookmark, but your computer already has a usable `main` branch, it now uses that branch instead of giving up.

## Fix proof
A regression test simulates a missing `refs/remotes/origin/main`, a present `refs/heads/main`, and an unavailable refresh; creation succeeds using `main` without starting a refresh. Existing SSH and slash-named base tests pass after moving SSH fallback before reuse/conflict checks.

## No regressions / trade-offs
0 regressions / risk: NO — focused local and SSH unit behavior is covered, including slash-named branch cases; full Electron and real SSH relay validation were not run. The fallback only applies after the originally named base is absent, so a real local branch named `origin/main` remains preferred.

## Reviews
Review loop count: 1. Final verdict: PASS after one P1 finding (SSH fallback timing) was addressed; no remaining findings.

## Tests
- `pnpm exec vitest run --config config/vitest.config.ts src/main/ipc/worktrees.test.ts` — 203 passed
- Commit hook: oxlint, React doctor, and oxfmt — passed

## Performance
Symptom: failed offline creation previously performed a failed remote refresh before erroring. Resource: network/process work. Deterministic count: this fallback starts 0 refreshes when the local counterpart exists (asserted in test), versus 1 attempted refresh previously.

## Linked issue
Fixes #7431


nwparker:
Added before merge:

- Local and SSH fallback results now report the unavailable remote-tracking ref and the local ref actually used.
- The renderer queues a compact warning dialog after creation, waits for any existing app modal to close, and warns that the workspace may not include the latest remote changes.
- Verified with 471 focused tests, full typecheck, changed-code quality checks, localization checks, and a real Electron interaction/visual pass.

AmethystLiang:

PR #11067, opened 2026-07-28
fix(ui): prevent IME Enter from submitting rename and notes fields

nwparker:
## Description

Supersedes #10711 while preserving the original contributor commit and attribution. Prevents the Enter key used by CJK IMEs to confirm a conversion candidate from submitting project-group, host, file, worktree-name, issue, and notes inputs.

## ELI5

When an IME turns typed phonetic text into a character, it first uses Enter to choose the character. Orca now waits for the next normal Enter before saving the form.

## Fix proof

- Added focused tests for `isComposing`, legacy `keyCode === 229`, and ordinary Enter in each affected surface.
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/editor/EditorPanelHeaderPath.ime-enter.test.tsx src/renderer/src/components/editor/UntitledFileRenameDialog.ime-enter.test.tsx src/renderer/src/components/sidebar/HostRenameDialog.ime-enter.test.tsx src/renderer/src/components/sidebar/ProjectGroupNameDialog.ime-enter.test.tsx src/renderer/src/components/sidebar/WorktreeMetaDialog.ime-enter.test.tsx` — 18 tests passed locally.

## No-regression proof / trade-offs

Plain Enter remains covered and submits normally. The existing Cmd/Ctrl-aware screen-submit shortcut remains after the shared IME guard; native CJK IME behavior still needs manual desktop verification on macOS, Linux, and Windows.

## Review

One independent native review loop: **clean**. It confirmed the guard precedes each explicit submit, the project-group form prevents implicit submit during composition, and no shortcut ordering conflict was found; reviewer target run reported 21/21 passing.

## Performance and compatibility

The guard is a constant-time keydown predicate with no polling, I/O, or allocation-heavy path. It uses browser keyboard event state only; there are no SSH, folder-workspace, Git provider, or path behavior changes.

## Evidence

No Electron screenshot was captured: the behavior is native IME interaction and the automated component tests are the available deterministic proof.

## Links

Supersedes and credits #10711 (@makoto-developer). Related category issue: #10711.

0 regressions / risk: **UNKNOWN** — targeted automated regressions are green, but native IME confirmation remains to be exercised on all desktop platforms.

### DIFF
--- a/src/renderer/src/components/editor/EditorPanelHeaderPath.ime-enter.test.tsx
+++ b/src/renderer/src/components/editor/EditorPanelHeaderPath.ime-enter.test.tsx
@@ -11,8 +11,7 @@ const { commitRename, cancelRename } = vi.hoisted(() => ({
   cancelRename: vi.fn()
 }))
 
-// Why: the rename field only mounts while the hook reports isRenaming, so the
-// hook is stubbed into that state instead of driving the whole rename flow.
+// Why: stub the hook state so this test isolates the mounted rename field.
 vi.mock('./editor-header-file-rename', () => ({
   useEditorHeaderFileRename: () => ({
     canRename: true,
@@ -79,12 +78,13 @@ function renderHeaderPath(): HTMLInputElement {
   return input
 }
 
-function pressEnter(
+function pressKey(
   input: HTMLInputElement,
+  key: string,
   init?: KeyboardEventInit & { keyCode?: number }
 ): void {
   const event = new KeyboardEvent('keydown', {
-    key: 'Enter',
+    key,
     bubbles: true,
     cancelable: true,
     ...init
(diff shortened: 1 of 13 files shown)


nwparker:
Landed, and then some — closing.

The predicate this PR introduced (`src/renderer/src/lib/ime-composition-keyboard-event.ts`) is on main and is now the shared gate for Enter-during-composition across the app. Where this PR covered 5 surfaces, main carries **42** `.ime-enter.test` files, plus `ime-enter-gesture-ownership-contract.test.ts` pinning the ownership rule itself so new dialogs cannot quietly skip it.

Both surfaces you targeted are covered: `WorktreeMetaDialog` (its Enter handling since extracted into `worktree-meta-enter-handlers.ts`) and `ProjectGroupNameDialog`, along with the editor rename paths.

Thanks — the framing that Enter-during-composition is an *ownership* question rather than a per-dialog bug is what made it generalize this far.

PR #11098, opened 2026-07-28
fix(browser): keep the address bar editable in a narrow toolbar

### LINKED ISSUE #11090: [Bug]: Address bar collapses to an icon at narrow pane widths, with no way to enter a URL
### Operating system

macOS

### Orca version

1.4.159

### Details

**Short summary**

When the built-in browser's pane is narrowed past a certain width, the address bar collapses into a globe icon. Clicking that icon only pops out **another globe icon with no input field**, so there is no way to type or edit a URL. The tab can then only be navigated via links that already exist on the page.

**What happened**
Image

1. Every other toolbar button (back / forward / reload / Import / …) stays fully visible at that width — the address bar is the only control that gets sacrificed.
2. Clicking the collapsed icon opens a popover container, but the container renders only an icon, with no text field (see second screenshot).

*Unverified guess:* it looks like the address bar is moved into an overflow container, but what gets rendered inside that container is still the collapsed icon variant rather than the expanded input variant.

**How can we reproduce it?**

1. Open a browser tab.
2. Narrow that tab's pane using a split layout (e.g. three or more panels side by side).
3. Look at the toolbar — the address bar has become a globe icon.
4. Click the icon — a popover appears containing only an icon, with nothing to type into.

**Expected behaviour**

It should be possible to enter a URL manually at any pane width. The address bar is the browser's most essential input control, so it should take priority over the other toolbar buttons.

**Suggested approaches**

1. **(preferred)** Clicking the collapsed icon expands a URL input (popover or dialog is fine); pressing Enter navigates.
2. When the toolbar is compressed, keep the address bar and move the other buttons into the `…` overflow menu instead.
3. Provide `Cmd+L` to focus the address bar, auto-expanding the input when it is in the collapsed state.

This situation comes up often in normal Orca usage — running several panels side by side for comparison is one of its main use cases, so a narrow browser pane is the common case rather than an edge case.

yoyoys:
## Summary

Fixes #11090 — in a narrow browser pane the address bar became impossible to use.

Every other control in the browser toolbar is `shrink-0`, so the address bar form (`min-w-0 flex-1`) was the only flexible item and absorbed the entire squeeze. Below roughly 420px of pane width it collapsed to just the leading globe icon with a zero-width input. Clicking it only opened the history/suggestion dropdown, which is sized with `w-[var(--radix-popover-trigger-width)]` and therefore also rendered at icon width — which is why the reported symptom was "clicking the globe pops up another globe". There was no collapsed mode to fix; the bar was simply being squeezed out of existence, leaving the tab navigable only through links already on the page.

Focusing a squeezed address bar now lifts the form out of the toolbar flow and overlays the row edge to edge, giving a full-width editable field that navigates on Enter. Because the form is still the Radix popover trigger, the suggestion list picks up the expanded width for free. When there is room, nothing changes: the bar stays inline exactly as before.

Details:

- `browser-address-bar-expansion.ts` holds the threshold and the pure predicate (`shouldOverlayBrowserAddressBar`), unit-tested separately.
- The measured element is a new in-flow slot `div`, not the form. The form leaves the flow while overlaid, so measuring the form itself would oscillate between the two states.
- The slot keeps a `min-w-11`, so the globe stays a real hit target. Without it the neighbouring toolbar buttons overlap the only affordance for reopening the field (Playwright's click was genuinely intercepted by the Import button before this was added).
- `BrowserAddressBarSuggestionList.tsx` is the previous inline dropdown JSX extracted verbatim — needed to stay under the 400-line `max-lines` budget for `.tsx` without a suppression.

Both the local and the remote-runtime toolbars render `BrowserAddressBar` inside the same `relative … px-3` row, so both get the fix.

## Screenshots

Browser pane narrowed to ~420px.

Squeezed resting state — the address bar is a globe pill:

collapsed

After clicking it — full-width editable field, Enter navigates:

expanded

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 38721 passed. 2 pre-existing failures in `src/relay/agent-exec-handler.test.ts` (they reproduce identically on an unmodified checkout of this branch point; they depend on a `codex` binary in the environment) plus some load-dependent flakes in `src/relay/subprocess.test.ts`, `src/main/plugins/plugin-worker-supervision.integration.test.ts` and `src/main/ssh/ssh-system-transport.integration.test.ts` that pass in isolation. None touch the renderer.
- [x] `pnpm build`
- [x] Added tests:
  - `browser-address-bar-expansion.test.ts` — unit coverage for the threshold predicate.
  - `tests/e2e/browser-address-bar-narrow-toolbar.spec.ts` — real Electron regression: narrows the window with the right sidebar closed, asserts the inline field is squeezed below 40px and no overlay is present, clicks the globe, asserts the overlay appears with a usable field, then types a URL and asserts the tab actually navigates. Verified this spec fails (`overlay … element(s) not found`) when the fix is disabled, so it is not vacuous.

## AI Review Report

This PR was written and reviewed by Claude (Claude Code). Root cause was established by reading the layout code before any change: the toolbar row, every button's `shrink-0` from the shared `buttonVariants`, `Input`'s `w-full min-w-0`, and the popover's trigger-width sizing — not by guessing from the screenshot in the issue. The issue's own hypothesis (an overflow container rendering a collapsed variant) was checked and rejected; there is no overflow container and no collapsed variant in this code path.

Risks checked:

- **Measurement feedback loop** — the overlay removes the form from flow. Addressed by measuring a separate in-flow slot; covered implicitly by the e2e, which would flap if the state oscillated.
- **Focus handling** — `BrowserPane.focusAddressBarNow()` focuses `addressBarInputRef`, and several call sites compare `document.activeElement` against it. The input is never unmounted or duplicated by this change, so Cmd+L and the blank-tab auto-focus keep working; the e2e exercises the auto-focus path directly (it had to blur first to observe the resting state).
- **Popover dismissal** — the existing window-blur / webview-focusin / Escape handlers are untouched; overlay state is derived from the same `open` flag they already control, so dismissal collapses the bar.
- **Cross-platform** — no keyboard shortcuts, accelerators, shortcut labels, paths, shell invocations or Electron platform branches are touched. The change is CSS classes plus a `ResizeObserver`, with a `typeof ResizeObserver === 'undefined'` guard for the non-DOM/web runtime. macOS/Linux/Windows behave identically; verified on macOS.
- **SSH / remote runtime and folder workspaces** — the remote-runtime toolbar shares the same component and the same `relative … px-3` row, so it is fixed too; nothing here assumes a git worktree or local execution.
- **Agent and integration compatibility** — none. This path is the embedded browser's chrome; it does not touch CLI agents, integrations, or git providers, and contains no provider-specific naming or branching.
- **Performance** — one `ResizeObserver` on one element per browser pane, already the same pattern this file uses for the chrome inset and viewport sync. It sets state only when the slot width changes, and the collapsed/expanded decision is a pure comparison, so nothing runs in a hot path. No new subscriptions, timers, or listeners.
- **UI quality / STYLEGUIDE** — the overlay reuses the address bar's existing surface (`rounded-xl`, `border-border`, `bg-background`) and only changes position. The review caught that my first version used `shadow-md`, which is a fourth elevation level that `docs/STYLEGUIDE.md` explicitly forbids; the follow-up commit switches it to the documented floating shadow (`0 10px 24px rgba(0,0,0,0.18)`) already used by the other floating surfaces in this pane. No new colors, radii, font sizes, or tokens. Both light and dark mode inherit from existing tokens.
- **`max-lines`** — no suppression added; the file was split instead, and the ratchet reports no new bypasses.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, dependency or IPC surface. The change is presentational (Tailwind classes on an existing form) plus a `ResizeObserver` on an element this component already owns. URL parsing, normalization and navigation continue to flow through the untouched `submitAddressBar` / `navigateToUrl` path in `BrowserPane`, and through `shared/browser-url`. No new dependencies. The added e2e spec binds a throwaway HTTP server on `127.0.0.1:0` and closes it in a `finally`. No follow-up needed.

## Notes

- The threshold (220px of slot width) is the point below which the inline field stops being usefully typable; it is a single exported constant if you would prefer a different value.
- No X handle to share — I'm not on X, so there's nothing to shout out. Everything else in CONTRIBUTING's PR checklist is covered above.
- Branch name is `yoyoys/fix-narrow-addressbar-overlay`. CONTRIBUTING asks for a clear, descriptive name and lists `fix/...` as an example; I tried renaming to `fix/narrow-browser-address-bar` and GitHub auto-closed this PR because branch renames don't carry across a fork boundary, so I restored the original name. Say the word if you want the prefix anyway and I'll open a fresh PR from a renamed branch.
- This only rescues the address bar. At very narrow widths the toolbar row still overflows horizontally and the trailing buttons are clipped — the pre-existing behaviour, and the broader "collapse secondary actions into the `…` menu" idea from the issue. Happy to take that on in a follow-up if you want it, but it is a bigger change and orthogonal to making the URL editable again.

https://claude.ai/code/session_01Mx53f7erbtw5NraS8HdXKE

### DIFF
--- a/src/renderer/src/components/browser-pane/BrowserAddressBar.tsx
+++ b/src/renderer/src/components/browser-pane/BrowserAddressBar.tsx
@@ -1,12 +1,14 @@
 /* oxlint-disable react-doctor/no-adjust-state-on-prop-change -- Why: dropdown visibility depends on DOM focus plus browser-history suggestions, so the close path is an imperative popover sync. */
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
-import { Globe, Search } from 'lucide-react'
+import { Globe } from 'lucide-react'
 import { Input } from '@/components/ui/input'
 import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover'
-import { Command, CommandGroup, CommandItem, CommandList } from '@/components/ui/command'
+import { cn } from '@/lib/utils'
 import { useAppStore } from '@/store'
 import { DEFAULT_SEARCH_ENGINE, type SearchEngine } from '../../../../shared/browser-url'
 import { buildBrowserAddressBarSuggestions } from './browser-address-bar-suggestions'
+import { shouldOverlayBrowserAddressBar } from './browser-address-bar-expansion'
+import BrowserAddressBarSuggestionList from './BrowserAddressBarSuggestionList'
 
 type BrowserAddressBarProps = {
   value: string
@@ -38,6 +40,24 @@ export default function BrowserAddressBar({
   const openedAtRef = useRef(0)
   const blurCloseTimerRef = useRef<number | null>(null)
   const closingResetTimerRef = useRef<number | null>(null)
+  const slotRef = useRef<HTMLDivElement | null>(null)
+  const [inlineWidth, setInlineWidth] = useState<number | null>(null)
+
+  // Why: the slot keeps its flex width even while the bar overlays the toolbar,
+  // so measuring it here (not the form) cannot oscillate with the overlay.
+  useEffect(() => {
+    const slot = slotRef.current
+    if (!slot || typeof ResizeObserver === 'undefined') {
+      return
+    }
+    const syncWidth = (): void => setInlineWidth(slot.getBoundingClientRect().width)
+    syncWidth()
+    const observer = new ResizeObserver(syncWidth)
+    observer.observe(slot)
+    return () => observer.disconnect()
+  }, [])
+
+  const overlay = shouldOverlayBrowserAddressBar({ inlineWidth, focused: open })
 
   const clearAddressBarTimers = useCallback((): void => {
     if (blurCloseTimerRef.current !== null) {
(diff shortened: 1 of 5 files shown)


nwparker:
Took this over to verify before merge. The fix itself holds up — I checked the two load-bearing claims rather than taking them on faith:

- **No measurement feedback loop.** Instrumented the real app at ~420px: the slot stays pinned at 44px while the form flips `static` → `absolute`, so measuring the slot instead of the form genuinely does avoid oscillation. Good call.
- **The regression test is not vacuous.** Stubbing `shouldOverlayBrowserAddressBar` to `false` fails it at `expect(overlay).toBeVisible()`, which is the right assertion.

One real problem: **the e2e spec failed on this branch.** It passed only from a clean profile, which is why it looked green locally. Two preconditions it set once are actively undone by the app:

- `BrowserPane` re-focuses a blank tab’s address bar across several animation frames plus the blank-url `did-finish-load` handler, so the single `blur()` got reverted and the bar never reached its squeezed resting state — the overlay stayed up and `toHaveCount(0)` failed 14/14 polls.
- Startup paths re-open the right sidebar. At a fixed 700px window that leaves the pane ~70px (700 − 280 − 350), so the overlay had nowhere to go and the field measured 0px wide.

Settling those separately just let whichever settled first drift back while the next one ran. I pushed `test(browser): make the narrow-toolbar regression deterministic`, which re-asserts both in one loop until they hold simultaneously, and sizes the window from the chrome actually measured instead of assuming 700px. Now 8/8 green and ~6s instead of ~15–19s, and still fails correctly when the fix is disabled.

Product code is unchanged by me — `pnpm lint`, `pnpm typecheck`, and the full renderer suite (16,964 tests) all pass. Merging once CI is green.

PR #11099, opened 2026-07-28
fix(sidebar): close the shortcut-opened workspace rename editor

socar-jian:
Supersedes #11092, which I have closed in favor of this one — two open PRs touching the same two files, one a subset of the other, only made the review call harder. The fix is the first commit here and stands on its own; the two after it are the behavior-preserving refactor and the tests added from review feedback.

## Summary

Renaming a workspace from the keyboard leaves the sidebar title stuck in an edit box that will not close. Enter, Escape and clicking away all fail. The box is borderless, so it looks like an ordinary title — until an agent finishes and the notification for that row pulls focus back in and selects the whole name. The next key you press replaces the title you already finished typing.

This PR:

- closes the editor on the `workspace.rename` shortcut path, so Enter commits, Escape cancels, and clicking away commits — matching the double-click path
- keeps an open editor in place when the row's title or unread flag changes, so a notification no longer steals the caret
- reports editing state to the parent card on the shortcut path, so card drag stays disabled while a rename is in progress
- then removes the structure that made the first defect possible, without changing behavior

## ELI5

The code that opens the rename box and the code that closes it were checking two different switches. Opening flipped only one of them, so closing looked at the other, concluded "we were never editing," and did nothing. Separately, any update to that row — a new title, an unread dot — rebuilt the box from scratch, and a fresh box always grabs focus and highlights its contents. That is fine when you just double-clicked to rename; it is not fine when a background notification triggers it.

## Root cause

Two independent defects in `WorktreeTitleInlineRename`. Either one alone is a papercut; together they produce the symptom above.

**1. The shortcut path never set `editingRef`.** The `beginEditing` Effect opened the editor with `setEditing(true)` rather than `setEditingMode(true)`:

```ts
setValue(displayName)
setEditing(true)          // editingRef stays false
```

`setEditingMode` starts with `if (editingRef.current === nextEditing) return`. With `editingRef` still `false`, every later `setEditingMode(false)` — from `commitRename`, from `cancelRename`, from the input's `onBlur` — returned early and left `editing` state `true`. The editor could not be closed by Enter, Escape, or clicking away. `onEditingChange` was never fired either, so `WorktreeCard` never learned a rename was in progress. The double-click path was unaffected because `startRename` calls `setEditingMode(true)`.

**2. The open editor remounted on unrelated row updates.** The editing branch keyed its root on `titleElementKey`, which is derived from `displayName` and `showUnreadEmphasis`:

```ts
const titleElementKey = `${displayName}:${showUnreadEmphasis ? 'unread' : 'read'}`
```

`use-notification-dispatch` calls `markWorktreeUnread` on `agent-task-complete`, `WorktreeCard` turns that into `showUnreadEmphasis`, the key changes, and React discards the subtree. The freshly mounted input hits the ref callback, which runs `input.focus()` and `input.select()` — correct on a genuine double-click open, wrong on a remount. Committing a rename triggered the same thing through the `displayName` half of the key.

Defect 1 leaves the editor open forever, which gives defect 2 something to remount on every later notification for that row. The sidebar renders the borderless `editingPresentation="text"` variant, so nothing on screen says an editor is still there — which is why this reads as the title randomly grabbing the cursor.

## Fix

First commit, `46a6fb2`:

- open through `setEditingMode(true)` so `editingRef` and `onEditingChange` stay in sync with `editing`
- give the editing branch a stable key so title/unread churn cannot remount a live editor

## Follow-up refactor

The remaining commits change no behavior. Defect 1 was possible because whether the editor is open lived in two places — an `editing` state and an `editingRef` — kept in agreement by convention inside a single setter, with nothing stopping a call site from writing only one. That pair collapses here into one `draftTitle: string | null`, so the question has exactly one answer.

- `editing` and `value` become a single `draftTitle: string | null` (`null` means closed). `openEditor()` and `closeEditor()` replace the `setValue(displayName)` + `setEditingMode(...)` pairs that every entry point had to remember to write together.
- `editingRef` is gone. It existed so the shared root ref callback could tell whether it was attached to the title or to the editor; splitting that callback into `handleTitleRootRef` and `handleEditorRootRef` answers the question by construction.
- The one surviving ref is renamed `notifiedEditingRef` and now means exactly one thing: the last value handed to `onEditingChange`. That dedupe has to stay — the parent callback is not idempotent, since `WorktreeCardMeta` defers its hovercard close on a `false` — but it is no longer a mirror of render state, so it cannot drift from it.
- `titleElementRef` was assigned on every ref call and never read anywhere; it disappeared with the callback split.

It is kept in its own commits so the fix can still be reviewed, or backported, without it.

## Screenshots

The fix has no visual change of its own — the rendered markup is identical, only lifecycle behavior differs. What is visible is the bug it removes.

Before the fix — Escape leaves the editor open, and a notification mid-edit remounts it, refocuses it and reselects the whole title (caret was at 3,3):

```
after shortcut : input present, focused, sel [0,12]
after Escape   : input STILL present, still focused
markWorktreeUnread(active) -> remounted: true, focused: true, sel [0,16]
```

Sidebar workspace title rendered as a fully selected text block after a notification

After the fix — Escape closes it, Enter saves and closes it, and the same notification leaves the caret untouched:

```
after shortcut : input present, focused, sel [0,12]
after Escape   : no input
rename + Enter : title becomes "renamed-by-probe", no input
markWorktreeUnread(active) while editing -> remounted: false, focused: true, sel [3,3]
```

Same row with the notification bell, title no longer selected

Both were captured by running this branch as a dev instance (`pnpm dev`, isolated `orca-dev` profile) against a throwaway repo and driving the real sidebar through the renderer's CDP endpoint. The rename shortcut is `Mod+Alt+R` on macOS.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

`WorktreeTitleInlineRename.editor-lifecycle.test.tsx` is new. It covers shortcut-Enter, shortcut-Escape, a double-click control case, and an unread flip during an open edit. Reverting the source fix fails 3 of the 4 and leaves only the double-click control passing, so the suite actually catches this regression.

The refactor keeps all 13 tests in `WorktreeTitleInlineRename.test.tsx`, `.begin-editing.test.tsx` and `.editor-lifecycle.test.tsx` passing unchanged — that is the point of it: same behavior, fewer ways to break it.

Three more tests were then added for the one contract the refactor has to preserve and nothing was holding: `onEditingChange` fires exactly once per transition. Worth recording what building them turned up. Repeat renders and a mid-edit unread flip never re-notify even with the dedup guard deleted, so those two assertions document intent rather than catch a regression. The path that does regress is a StrictMode mount — the app mounts under `StrictMode`, so the open effect replays against the pre-open state and, without the guard, notifies the parent twice. Deleting the guard fails that case with 2 calls instead of 1.

Full-suite detail, since the checkbox above hides it: `pnpm test` reported 13 failures out of 38,734 on the first run, all in `src/main` / `src/relay` process, git and relay integration files. I had a dev instance running alongside it for the manual check above, which those tests contend with. With the dev instance stopped, re-running those 10 files passed 9. The one that stayed red — `src/main/repo-icon-autodetect.test.ts` → "uses the resolved fork upstream for both metadata and the GitHub avatar" — fails identically on an unmodified tree, so it predates this work. Nothing under `src/renderer` failed at any point.

react-doctor reports the same two pre-existing warnings on this file before and after; no new ones.

## AI Review Report

Reviewed with Claude Code, focused on the risks this repo asks about:

- **Cross-platform**: renderer-only React state and key handling — no platform branches, path handling, or accelerator strings. `workspace.rename` ships a default binding on macOS only (`Mod+Alt+R`), so elsewhere the path is reached through a user-defined binding; the change itself is platform-neutral.
- **SSH / remote / local**: no main-process, git, IPC, or transport code is touched. Rename still flows through the existing `onRename` prop, so local, SSH and folder workspaces behave identically.
- **Agents / integrations / git providers**: none referenced. The notification link is one-way — `markWorktreeUnread` is unchanged and this component only reads the resulting flag.
- **Performance**: strictly fewer renders. The stable editing key removes a full subtree unmount/remount per title or unread change while an editor is open. The refactor drops one `useState` and one ref per instance; `openEditor`/`closeEditor` are `useCallback`s with the same dependency shape the old setter had, and the editor's ref callback has an empty dependency list, so it never re-attaches. Nothing new runs on any hot path.
- **UI quality**: both call sites were checked — the sidebar row (`editingPresentation="text"`) and the hovercard identity header (`editingPresentation="field"`). Markup and class names are untouched in both. The non-editing title keeps its original key, so tooltip truncation re-measurement on title/unread changes is preserved.
- **Flagged and verified**: adding `setEditingMode` to the Effect deps could re-run it, so I checked the guard order — the Effect returns immediately unless `beginEditing` is true, and the parent clears its trigger in `onBeginEditingConsumed`, so the open request still fires once.
- **Flagged and verified**: the old code called `measureTitleTruncated(null)` when entering edit mode, and that call is gone. It is now redundant — entering edit mode unmounts the title span, so `handleTitleRootRef(null)` runs the same reset on the way out. The tooltip cannot fire while editing either, since the title branch is not rendered.
- **Flagged and verified**: `commitRename` gained a `draftTitle === null` guard, which also makes a late `onBlur` after the editor already closed a no-op instead of relying on `savingRef` alone.

## Security Audit

No input parsing, command execution, path handling, auth, secrets, IPC surface, or dependency changes. The edit is confined to React state transitions and a React key inside one presentational component; the rename value itself still travels the pre-existing `onRename` path with unchanged trimming and validation in `getWorktreeTitleRenameCommit`. No new attack surface, no follow-up needed.

## Notes

- Verified on macOS. Reproducing on Linux or Windows needs a user-defined binding for this shortcut, since neither ships one by default.
- The screenshots live on a separate `assets/rename-editor-lifecycle` branch of my fork, so no binaries land in this PR's diff. Happy to re-upload them as GitHub attachments if you prefer.
- The `titleElementRef` removal is the one change here that is not strictly mechanical. It was dead — assigned, never read — and it lived inside the callback being split, so leaving it would have meant carrying it into one of the two new callbacks arbitrarily. Happy to split it out if you would rather keep that diff purely structural.
- The two react-doctor warnings on this file (`effect-needs-cleanup` on the ResizeObserver, `no-adjust-state-on-prop-change` on the begin-editing Effect) predate this change and are left alone to keep the diff focused.
- X handle, per the contributor guide: [@hgijeon](https://x.com/hgijeon)


socar-jian:
Good catch — added in 3aa5147. Recording what building the test turned up, since one part of it cut against the suggestion.

Asserting call counts across re-renders passes with or without the dedup guard. `notifyEditingChange` only runs from `openEditor`/`closeEditor`, so a plain re-render — including a mid-edit unread flip — never reaches it. Those two assertions document the intent but would not catch the regression.

The path that does regress is a StrictMode mount. The app mounts under `StrictMode`, so the begin-editing effect replays against the pre-open state and both passes call `openEditor`. Without the guard the parent is notified twice; the test fails with 2 calls instead of 1. That is the case worth having, and it is the one that actually protects `WorktreeCardMeta`'s non-idempotent hovercard close.

All three are in `WorktreeTitleInlineRename.editor-lifecycle.test.tsx`; suite is 16 passing.

socar-jian:
@coderabbitai full review

socar-jian:
Closing this in favour of a fresh PR against current `main`.

`#13429` (Aug 10) rewrote much of `WorktreeTitleInlineRename.tsx` for the emoji picker, and this branch now conflicts. Resolving it would graft the change onto a structure that no longer exists, so rebuilding the fix on top of today's `main` is the cleaner path.

To be clear, this is not obsolete — both defects are still present on `main`:

- the `beginEditing` effect calls `setEditing(true)` instead of `setEditingMode(true)`, so `editingRef` stays `false` and every later `setEditingMode(false)` hits the equality guard and early-returns. The shortcut-opened rename editor cannot be closed by Enter, Escape, or blur. The double-click path is unaffected.
- the editing span's key still derives from the unread flag, so an agent notification flips it, remounts the input, and re-runs `focus() + select()` — the next keystroke overwrites the name being typed.

`WorktreeTitleInlineRename.begin-editing.test.tsx` covers opening, trigger consumption, and IME Enter, but nothing asserts that the editor closes, which is why this has stayed green.

I'll open a self-contained replacement and link it here.

socar-jian:
Replacement opened: #13703. Rebuilt on today's `main` (post-#13429), so it applies cleanly, and it carries regression coverage for both defects plus before/after recordings.

PR #11112, opened 2026-07-28
fix(skills): keep a failed update run honest when nothing converged

brennanb2025:
## What regressed

#11105 (a8660839ee) reclassified `outdated` as not-a-failed-run so a clean-exit lock-drift no-op stops arming the red banner + Retry loop. Correct — but `finish()` in `src/main/skills/skill-update-run.ts` discards `spawnError` whenever the re-scan verdict is empty (`failedNames ?? (spawnError ? names : [])` short-circuits on `[]`). Pre-#11105 an empty verdict proved the requested skills converged; post-#11105 it also covers "left old and untouched".

**Exact false-success path:** offline / npx ENOENT / non-zero exit -> CLI writes nothing -> skills still merely `outdated` -> verdict `[]` -> run publishes green "Updated N skills" and a green status-bar check, masking the process failure.

## Fix

Thread the command's own result into the verdict (the cheap honest signal — no pre-run state, no stdout parsing):

- `skill-update-run.ts`: `rescan(names, spawnError !== null)`
- `ipc/skills.ts`: forwards the flag
- `skill-update-outcome.ts`: `outdated` is forgiven only when the CLI reported success; `current`/`newer-known` still count as landed even on a failed exit (a peer skill can fail the process — preserves the tested behavior)

The #11105 core fix survives: exit 0 + outdated -> success, no banner, no Retry.

## Test evidence

- `pnpm vitest run` outcome + run + ipc/skills tests: **38/38 pass**
- Mutation 1 — forgive `outdated` unconditionally: `still blames an outdated copy when the command itself failed` **fails** (1/12)
- Mutation 2 — runner always passes `commandFailed=false`: `reports the failure when the command failed and nothing converged` **fails** (1/20)
- Peer-skill pin (`treats a clean re-scan as success even though the exit code is non-zero`) and both #11105 reversal pins still pass
- `rm -f config/*.tsbuildinfo && pnpm run tc`: clean

### DIFF
--- a/src/main/ipc/skills.ts
+++ b/src/main/ipc/skills.ts
@@ -30,9 +30,9 @@ export function registerSkillsHandlers(store: Store): void {
   const runner = new SkillUpdateRunner({
     // Why: per-skill outcomes come from re-hashing what is actually on disk, not
     // from scraping stdout.
-    rescanOutdatedNames: async (names) => {
+    rescanOutdatedNames: async (names, commandFailed) => {
       const inventory = await scanInventory()
-      return skillUpdateFailedNames(names, inventory.installations)
+      return skillUpdateFailedNames(names, inventory.installations, commandFailed)
     },
     onState: (run: SkillUpdateRun) => {
       for (const window of BrowserWindow.getAllWindows()) {
--- a/src/main/skills/skill-update-outcome.test.ts
+++ b/src/main/skills/skill-update-outcome.test.ts
@@ -35,7 +35,9 @@ function placement(
 
 describe('skillUpdateFailedNames', () => {
   it('treats a convergent copy that is now current as landed', () => {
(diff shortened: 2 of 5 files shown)


brennanb2025:
Closing: superseded by the revert in #11110. This patched the spawnError/exit-code interaction inside the forgiven-outdated semantics from a8660839ee, but the forgiveness itself was the wrong layer — on the exact user-reported path (clean exit 0, CLI no-ops off its lock cache) it turns a universal false red into a universal false green. The real fix is upstream eligibility gating (don't offer an update the command cannot converge: lock skillFolderHash vs the disk revision's gitTreeSha via matchingKnownSnapshot in skill-package-identity.ts), owned by @nwparker on nwparker/fix-skill-updater-registration. The exit-code-honesty test pins in this branch are worth carrying into that work.

PR #2098, opened 2026-05-16
Run E2E tests on a schedule

AmethystLiang:
## Summary
- remove the reusable E2E workflow from PR checks
- run the E2E workflow on GitHub cron at 10am and 3pm America/Phoenix
- keep manual and reusable workflow entry points available

## Validation
- ruby YAML parse for changed workflow files
- git diff --check

PR #11131, opened 2026-07-28
ci(pr): run E2E when a PR touches tests/e2e paths (advisory)

### LINKED ISSUE #10518: [Bug]: E2E regression tests can merge red — PR CI never runs the E2E suite

nwparker:
Supersedes #10600 by @innocarpe — their two commits are preserved as the base of this branch, rebased onto current `main` (the original was 185 commits behind and `pr.yml` was restructured underneath it). My changes are the second commit, so the delta is reviewable on its own.

Closes #10518.

## Why this exists

E2E has not run on PRs since #2098 (May 16) removed the `pr.yml` trigger. As of #11031 (yesterday) the release path dispatches it *after* publication, where it cannot block. So a regression spec under `tests/e2e/` currently has no gate at all — not on the PR, not on the release cut.

This is not hypothetical: #11098 shipped an e2e spec that failed deterministically on its own branch. It looked green because nothing ran it.

## What @innocarpe wrote (commit 1)

Path-filtered `e2e-paths` detector + `e2e` job that `workflow_call`s the existing suite, so ordinary PRs stay light but any E2E-relevant change gets a full shard run. Plus merge-base diffing so base drift can't false-trigger, and least-privilege `contents: read` on the detector and on `e2e.yml`.

## What I fixed (commit 2)

**1. The gate did not gate.** `verify` is the required check and enumerates its dependencies explicitly. `e2e` was in neither `needs` nor the result list, so a failing shard left `verify` green — reproducing the exact hole this job closes, just with a red box further down the page.

Since the job is path-filtered, `skipped` is the normal result and has to keep passing. That allowance is checked *after* the strict loop, never inside it, so it cannot leak to the six always-required jobs:

```
E2E=success   -> PASS
E2E=skipped   -> PASS
E2E=failure   -> FAIL
E2E=cancelled -> FAIL
PACKAGE=failure + E2E=skipped -> FAIL   (skip can't mask a real failure)
```

**2. The `playwright.` pattern matched nothing.** The config is `tests/playwright.config.ts` — beside `tests/e2e/`, not inside it — so no tracked file starts with `playwright.`, and editing the runner config would silently skip E2E. Anchored at `tests/playwright.`.

## Testing

- New `config/scripts/pr-e2e-gate-contract.test.mjs`, alongside the existing release-E2E contract test. Verified non-vacuous: reverting either fix independently fails it.
- Updated `pr-workflow-parallelism.test.mjs`, which pins `verify.needs` exactly and correctly caught my change.
- `config/scripts` suite: 534 passed. `pnpm lint` clean.

## One decision for you

This makes E2E **blocking** on PRs that touch E2E paths. Given you moved release E2E to non-blocking yesterday because the suite is flaky, you may not want that on the PR path — and fork PRs need per-run maintainer approval, so this adds approval toil on community PRs touching specs.

If you'd rather keep it advisory, drop the `- e2e` line from `verify.needs` and the `$E2E` check; fix 2 and the detector still stand on their own. Happy to make that change instead — flagging it rather than assuming.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -2,6 +2,10 @@ name: E2E
 
 run-name: E2E ${{ inputs.ref || github.ref }}
 
+# Why: checkout + artifact upload only; callers can only further restrict.
+permissions:
+  contents: read
+
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -323,10 +323,17 @@ jobs:
       - shell_contracts
       - test
       - package
-      - e2e
     runs-on: ubuntu-latest
 
     steps:
+      # Why: e2e is deliberately absent from needs. The suite is currently red on
+      # main (every scheduled run), so gating merges on it would block any PR that
+      # touches tests/e2e/** — including the ones fixing the suite. Until it is
+      # green the job runs and reports for E2E-path PRs without blocking. To flip
+      # it on: add `e2e` to needs, add E2E to the env below, and require
+      # `"$E2E" = success || skipped` after the loop — skipped is the normal
+      # result for a path-filtered job and must keep passing, so it has to be
+      # checked outside the loop or it would excuse the jobs above.
       - name: Require successful checks
         env:
           STATIC_ANALYSIS: ${{ needs.static_analysis.result }}
@@ -335,7 +342,6 @@ jobs:
           SHELL_CONTRACTS: ${{ needs.shell_contracts.result }}
           TEST: ${{ needs.test.result }}
           PACKAGE: ${{ needs.package.result }}
-          E2E: ${{ needs.e2e.result }}
         run: |
           for result in \
             "$STATIC_ANALYSIS" \
@@ -348,10 +354,3 @@ jobs:
               exit 1
             fi
           done
-
-          # Why: E2E is path-filtered, so "skipped" is the expected result on a PR
-          # that touches no E2E files and must stay passing. Check it separately
-          # from the always-required jobs above so a skip can never satisfy those.
-          if [ "$E2E" != "success" ] && [ "$E2E" != "skipped" ]; then
-            exit 1
-          fi
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -294,7 +294,10 @@ jobs:
           # (pipeline status in `if` is not aborted by set -e). Merge-base limits the
           # list to files this PR introduced, not base-branch drift.
           CHANGED="$(git diff --name-only --merge-base "$BASE" "$HEAD")"
-          if printf '%s\n' "$CHANGED" | grep -E '^(tests/e2e/|playwright\.|\.github/workflows/e2e\.yml$)' >/dev/null; then
+          # Why: tests/playwright.config.ts sits beside tests/e2e/, not inside it, so
+          # it needs its own pattern — a bare `playwright.` prefix matches no tracked
+          # file and would silently skip E2E when the runner config changes.
+          if printf '%s\n' "$CHANGED" | grep -E '^(tests/e2e/|tests/playwright\.|\.github/workflows/e2e\.yml$)' >/dev/null; then
             echo "should_run=true" >> "$GITHUB_OUTPUT"
             echo "E2E path changes detected"
           else
@@ -320,6 +323,7 @@ jobs:
       - shell_contracts
       - test
       - package
+      - e2e
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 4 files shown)


nwparker:
Both CodeRabbit points were valid — I confirmed by reverting each regression and watching the contract stay green, then re-confirmed it now fails.

- **Path filter unasserted.** `e2e` could lose its `if:` and run on every PR — exactly the cost the filter exists to avoid — with nothing failing. Now pinned, including that `e2e-paths` still exports `should_run`.
- **Strict loop hardcoded 4 of 6 required jobs.** Dropping `GIT_COMPATIBILITY` or `SHELL_CONTRACTS` left them unenforced while the contract passed. Rather than extend the hardcoded list, I derive it from `verify.needs`, so a newly added required job that misses the loop fails here instead of silently going unchecked — that failure mode is the more likely one going forward.

Verified all three regressions now fail the contract independently (drop the `if:`, drop `GIT_COMPATIBILITY`, drop `SHELL_CONTRACTS`). 535 `config/scripts` tests pass, `pnpm lint` clean.

nwparker:
Changed approach on the gating question — flagging since it reverses what I had.

I originally wired `e2e` into `verify` so a red spec would block. Then this PR's own CI run failed 3 of 12 E2E shards on specs unrelated to it (agent-session resume, Jira linking, plugin containment, terminal artifacts). Checking the suite history: **E2E has failed every scheduled run on `main` — 22 of the last 22.** It is red today, before this change.

Gating on that would have blocked every PR touching `tests/e2e/**`, including the ones fixing the suite. So `e2e` now runs and reports but stays out of `verify.needs`. Flipping it on is a three-line change once the suite is green; the exact wiring is recorded on `verify`'s Require-successful-checks step, and the contract test pins the advisory choice so it reads as deliberate rather than as the unwired-gate bug this file originally caught.

Everything else is unchanged and stands on its own: the detector, the `tests/playwright.` path fix, and the contract tests (all three regressions still fail them independently).

Worth saying plainly: **the 22/22 red suite is the bigger finding and deserves its own issue.** This PR only makes those failures visible on PRs that touch E2E paths — it does not fix them.

PR #11145, opened 2026-07-28
feat(sidebar): keyboard shortcut to pin/unpin current workspace

### LINKED ISSUE #11111: [Feature]: Add a keyboard shortcut to pin or unpin the current project
### Problem or use case

  Please add a bindable “Toggle pin for current project/workspace” action to Orca’s command palette and Shortcuts settings, with support for macOS and Windows. Codex Desktop has this  feature, and it makes organizing active projects much faster without reaching for the mouse. 

### Proposed solution

A shortcut that we can use or assign or an endpoint that we can create some coding around

Codex desktop has this and it's really handy for quickly bringing apps that you're working with to the top.

innocarpe:
## Description
Add a bindable global shortcut (and Cmd+J action) to pin/unpin the focused workspace, matching the existing context-menu Pin/Unpin action.

## Focused fix
- In scope: `workspace.togglePin` keybinding, pure target resolver, App/Cmd+J wiring to `setWorktreesPinnedAndReveal`
- Out of scope: project-group pin (not a product surface), default chord

## Preserves
- Existing worktree pin store API and sidebar menu
- No default binding; works only when a focused worktree exists

## Evidence
- `pnpm exec vitest run --config config/vitest.config.ts src/shared/keybindings.test.ts src/shared/workspace-pin-toggle.test.ts src/renderer/src/components/cmd-j/quick-action-context.test.ts` → 87 passed

## User-regression-tradeoffs
- Unbound by default; assign in Settings → Shortcuts

Fixes #11111

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -182,6 +182,8 @@ import {
   type PhysicalModifierToken
 } from '../../shared/keybindings'
 import { PLUGIN_COMMAND_ALIAS_ACTION_IDS } from '../../shared/plugins/plugin-command-actions'
+import { getActiveSidebarWorkspaceId } from '../../shared/workspace-scope'
+import { resolveWorkspacePinToggleTarget } from '../../shared/workspace-pin-toggle'
 import { registerAppCommandDispatcher } from '@/lib/app-command-dispatch'
 import { executePluginCommand } from '@/lib/plugin-command-execution'
 import { findPluginCommandForKeybinding } from '@/lib/plugin-command-keybindings'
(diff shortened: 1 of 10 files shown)


stasmarkin:
Hi @innocarpe — heads-up that I opened #20385 against the same issue, so a maintainer does not end up choosing blind.

Your Cmd+J quick action is the part of this PR I did not have and should have: the palette is where people find a feature when they do not know its chord. I have added an equivalent to mine and credited the idea and the verb keywords to you, in the commit message and in the PR body. Say the word if you would rather I drop it.

The other thing worth flagging is that this branch no longer applies to `main`. `git merge-tree origin/main <this branch>` conflicts in 6 of the 10 files, including a modify/delete on `src/shared/keybindings.test.ts`. It is structural rather than ordinary drift: since late July `src/shared/keybindings.ts` became a barrel over `keybindings/*`, and the command-handler map moved out of `App.tsx` into `app-shell/app-command-handlers.ts` — so the two largest hunks here patch code that is gone.

Where mine differs beyond that: three actions instead of one (`pin` and `unpin` are idempotent, for macros where a toggle is ambiguous), the chord targets the hovered sidebar row and falls back to the focused workspace, `Mod+Alt+P` ships bound, and it fixes a pre-existing bug both our versions would otherwise hit — `setWorktreesPinnedAndReveal` re-resolves the id without an execution host, so on a repo registered both locally and over SSH the write lands on the wrong twin.

Genuinely happy either way. If you would rather carry this yourself, take whatever is useful from #20385 and I will close it — the host-collision fix is worth having regardless, and I can send it as a standalone PR.

innocarpe:
Closing this stale branch because it no longer applies to current main. The same issue is covered by #20385, which includes the palette actions and credits the Cmd+J idea. This avoids duplicate review.

PR #620, opened 2026-04-14
feat: scope terminal shell history per worktree

AmethystLiang:
## Summary
- Adds worktree-scoped shell history so each worktree gets its own HISTFILE — ArrowUp no longer surfaces commands from other worktrees
- Integrates with the new `LocalPtyProvider` architecture: history env injection runs after `buildSpawnEnv` but before `spawnShellWithFallback`, and `onBeforeFallbackSpawn` keeps HISTFILE in sync when the shell falls back
- Adds `terminalScopeHistoryByWorktree` setting (defaults to `true`) for users who prefer shared global history
- Cleans up per-worktree history directories on worktree removal and runs background GC for orphaned dirs
- Passes `worktreeId` through the renderer → preload → main IPC chain to enable history scoping

## Test plan
- [x] Unit tests for `terminal-history` scoping in `terminal-history.test.ts`
- [x] Typecheck passes (only pre-existing ssh2 module errors)
- [ ] Manual: open two worktrees, run distinct commands in each, verify ArrowUp shows only local history
- [ ] Manual: delete a worktree and verify its history dir is cleaned up

PR #11146, opened 2026-07-28
fix(terminal): keep worktree HISTFILE after macOS /etc/zshrc

### LINKED ISSUE #11044: [Bug]: Per-worktree shell history is silently disabled on macOS — /etc/zshrc overwrites the injected HISTFILE
**Operating system:** macOS
**Orca version:** 1.4.159 (observed on macOS 15 / Darwin 25.3.0, zsh)

### Short summary

`injectHistoryEnv()` sets a worktree-scoped `HISTFILE` in the PTY spawn environment. On macOS that value never survives shell startup: Apple's system `/etc/zshrc` reassigns `HISTFILE` unconditionally, and because it runs while `ZDOTDIR` still points at Orca's shell-ready wrapper dir, history lands inside Orca's wrapper sandbox instead. The user sees an empty history in every Orca terminal, and nothing they type is written back to their real history file.

The worktree-scoped history feature (#620) is therefore a no-op on the primary platform.

### What happened

In an Orca terminal:

```
$ echo $HISTFILE
/Users/<user>/Library/Application Support/orca/shell-ready/zsh/.zsh_history
```

Consequences:

- `↑`, `Ctrl+R`, and zsh-autosuggestions all start empty — no prior history is visible.
- Commands typed in Orca are never written to `~/.zsh_history`.
- Instead they accumulate in the wrapper directory. On my machine that file exists and has real commands in it:
  ```
  $ wc -l ~/Library/Application\ Support/orca/shell-ready/zsh/.zsh_history
  6
  ```

### How can we reproduce it

Outside Orca, one line (macOS, system zsh):

```sh
HISTFILE=/tmp/wanted /bin/zsh -ic 'echo $HISTFILE'
# expected: /tmp/wanted
# actual:   /Users/<you>/.zsh_history      ← injected value destroyed
```

Adding `ZDOTDIR` reproduces Orca's exact situation:

```sh
mkdir -p /tmp/zdt && : > /tmp/zdt/.zshrc
ZDOTDIR=/tmp/zdt HISTFILE=/tmp/wanted /bin/zsh -ic 'echo $HISTFILE'
# actual: /tmp/zdt/.zsh_history            ← falls into the wrapper dir
```

In Orca: open any terminal on macOS with `terminalScopeHistoryByWorktree` enabled (the default) and run `echo $HISTFILE`.

### Root cause

`/etc/zshrc` line 16 on macOS assigns `HISTFILE` with no check-before-set:

```sh
HISTFILE=${ZDOTDIR:-$HOME}/.zsh_history
```

zsh's login startup order is:

```
$ZDOTDIR/.zshenv → /etc/zprofile → $ZDOTDIR/.zprofile → /etc/zshrc → $ZDOTDIR/.zshrc → /etc/zlogin → $ZDOTDIR/.zlogin
```

Orca launches zsh with `args: ['-l']` and `ZDOTDIR` pointed at the shell-ready wrapper dir (`src/main/providers/local-pty-shell-ready.ts`, `src/main/daemon/shell-ready.ts`). So by the time `/etc/zshrc` runs:

1. the `HISTFILE` injected by `injectHistoryEnv()` has already been overwritten, and
2. `ZDOTDIR` is still the Orca wrapper dir, so the replacement path resolves into `.../shell-ready/zsh/.zsh_history`.

None of the wrapper files restore it afterwards — `src/main/shell-templates.ts` contains no reference to `HISTFILE` at all.

The comment in `src/main/terminal-history.ts` that justifies the env-var approach is where the assumption breaks:

> Why this is the industry-standard approach: Ghostty, Kitty, and VS Code all use check-before-set for HISTFILE. The major zsh frameworks (oh-my-zsh, Prezto) guard their HISTFILE assignments, so env-var injection works for the vast majority of users (see design doc §9).

The framework claim is correct; the premise that nothing else clobbers `HISTFILE` is not. Apple's system-wide `/etc/zshrc` is unguarded and runs on every macOS zsh startup, before any user or framework config.

A related gap with the same root: `getZshFinalZdotdirRestoreBlock()` restores `ZDOTDIR` to the user's real config dir, but leaves the `HISTFILE` that `/etc/zshrc` derived from the wrapper `ZDOTDIR` untouched. (#10051 reports a different problem in that same restore block, so any fix here should be coordinated with it.)

### Scope

- **macOS + zsh** — affected. This is Orca's default shell on macOS.
- **macOS + bash** — not affected. Neither `/etc/profile` nor `/etc/bashrc` assigns `HISTFILE`; I verified the injected value survives through `--rcfile`.
- **Linux** — depends on the distro's `/etc/zsh/zshrc`; several ship an unguarded assignment, others ship none. Not verified here.

### Additional observation (may be a separate defect)

On my machine there is **no** worktree shell-history directory at all. `injectHistoryEnv()` → `ensureHistoryDir()` creates directories named by `hashWorktreeId()` (16 hex chars) under `<userData>/terminal-history/`, but every entry there is a daemon scrollback directory (`output.log` / `checkpoint.json`, and a different `meta.json` schema of `{cwd, cols}`). There are zero hex-named directories and zero `zsh_history` files.

`injectHistoryEnv()` is only called from `LocalPtyProvider` (`src/main/providers/local-pty-provider.ts:799`), and `src/main/daemon/daemon-init.ts:796,837` replace that provider via `setLocalPtyProvider(routedAdapter)` when the daemon is active. If terminals are routed through the daemon adapter, history injection would be bypassed entirely — in which case fixing the `/etc/zshrc` clobber alone would not restore the feature. I have not confirmed this path; flagging it so it can be checked alongside.

### Suggested fix

Back up the intended value in a separate variable that `/etc/zshrc` does not touch, then restore it after the user's startup files have run — and only when the current value is the wrapper-sandbox path, so a `HISTFILE` the user set themselves in `~/.zshrc` is never overridden:

```sh
# after sourcing the user's startup file
case "${HISTFILE:-}" in
  */shell-ready/zsh/*) export HISTFILE="${ORCA_HISTFILE:-$_orca_home/.zsh_history}" ;;
esac
```

- `ORCA_HISTFILE` would be a new backup env var written by `injectHistoryEnv()` alongside `HISTFILE`.
- The `*/shell-ready/zsh/*` guard is a precise discriminator: only `/etc/zshrc` can produce that path, because Orca's wrapper swaps `ZDOTDIR` to the user's real dir before sourcing their `.zshrc`. On a system with no unguarded `/etc/zshrc`, the injected value is still intact, the pattern does not match, and nothing is touched.
- Restore placement: `.zlogin` for the login path (Orca spawns with `-l`, and the wrapper's rc-file blocks are gated behind `[[ ! -o login ]]`), mirroring how `ZDOTDIR`, `CODEX_HOME`, and `OPENCODE_CONFIG_DIR` are already restored. I verified in a real PTY that restoring `HISTFILE` from `.zlogin` is **not** too late — zsh reads the history file after all startup files complete, and `fc -l` shows prior entries correctly.
- Both wrapper generators would need it: `src/main/providers/local-pty-shell-ready.ts` and `src/main/daemon/shell-ready.ts` (they mirror each other through `src/main/shell-templates.ts`).

### Open questions

1. **What is the intended post-fix target?** Should `HISTFILE` end up at the worktree-scoped file (restoring the #620 feature as designed), or at the user's `~/.zsh_history`? The design doc referenced throughout the code (`terminal-history-scope-design.md`, §6–§10) was removed from tracked files in #620, so I could not check the intended behavior — those section references are now dangling. Happy to follow whatever the intent was.
2. **What should happen when `terminalScopeHistoryByWorktree` is off?** Today nothing is injected, so `/etc/zshrc` still drops `HISTFILE` into the wrapper sandbox — the "off" state is broken too, arguably worse, since users who opt out get neither scoped history nor their normal history. Should the wrapper unconditionally restore `HISTFILE` to `$_orca_home/.zsh_history` when scoping is disabled?
3. **Should the fix be scoped to macOS?** Linux distro `/etc/zsh/zshrc` files vary. The `*/shell-ready/zsh/*` guard above is inherently self-limiting (it only fires when something actually clobbered the value), so it could apply on all POSIX platforms safely — but I'd rather confirm that's wanted than assume.

I'm happy to open a PR once the intended behavior in (1) and (2) is settled.

innocarpe:
## Description
Worktree-scoped zsh history no longer lands in Orca's shell-ready sandbox on macOS after Apple's `/etc/zshrc` overwrites `HISTFILE`.

## Focused fix
- In scope: set durable `ORCA_HISTFILE` in `injectHistoryEnv`; re-export `HISTFILE` from it in the shared zsh final ZDOTDIR restore block (local/daemon/relay)
- Out of scope: bash/fish history redesign, opting into shared global history

## Preserves
- Check-before-set: caller-provided `HISTFILE` is still respected (no `ORCA_HISTFILE` then)
- Worktree history only when Orca injects; no change when history scoping is off

## Evidence
- `pnpm exec vitest run --config config/vitest.config.ts src/main/terminal-history.test.ts src/main/providers/local-pty-shell-ready.test.ts src/main/daemon/shell-ready.test.ts src/relay/pty-shell-launch.test.ts`
- Unit/wrapper-content assertions for `ORCA_HISTFILE` re-export green (some node-pty interactive PTY tests may need full native rebuild in CI)

## User-regression-tradeoffs
- Final HISTFILE is forced back to Orca's worktree path when `ORCA_HISTFILE` is set; intentional for the #620 worktree history feature on macOS

Fixes #11044

### DIFF
--- a/src/main/daemon/shell-ready.test.ts
+++ b/src/main/daemon/shell-ready.test.ts
@@ -147,6 +147,10 @@ function expectZdotdirSourceContext(content: string, fileName: '.zprofile' | '.z
 function expectFinalZdotdirRestoreContext(content: string) {
   expect(content).toContain("after Orca's last wrapper file has loaded")
   expect(content).toContain('export ZDOTDIR="$_orca_home"')
+  // Why: re-apply worktree HISTFILE after macOS /etc/zshrc (#11044).
+  expect(content).toContain(
+    '[[ -n "${ORCA_HISTFILE:-}" ]] && export HISTFILE="${ORCA_HISTFILE}"'
+  )
 }
 
 describePosix('daemon shell-ready launch config', () => {
--- a/src/main/providers/local-pty-provider.ts
+++ b/src/main/providers/local-pty-provider.ts
@@ -799,6 +799,10 @@ export class LocalPtyProvider implements IPtyProvider {
         wslDistro: preferredWslContext?.distro ?? worktreeWslContext?.distro ?? null
       })
       logHistoryInjection(worktreeId, historyResult)
+    } else {
+      // Why: skip injectHistoryEnv — still drop inherited ORCA_HISTFILE so the
+      // zsh wrapper cannot re-export a prior worktree path over default history.
+      delete finalEnv.ORCA_HISTFILE
     }
 
     await prepareLocalPtySpawn(id)
--- a/src/main/providers/local-pty-shell-ready.test.ts
+++ b/src/main/providers/local-pty-shell-ready.test.ts
@@ -340,6 +340,10 @@ function expectZdotdirSourceContext(content: string, fileName: '.zprofile' | '.z
 function expectFinalZdotdirRestoreContext(content: string) {
   expect(content).toContain("after Orca's last wrapper file has loaded")
   expect(content).toContain('export ZDOTDIR="$_orca_home"')
+  // Why: re-apply worktree HISTFILE after macOS /etc/zshrc (#11044).
+  expect(content).toContain(
+    '[[ -n "${ORCA_HISTFILE:-}" ]] && export HISTFILE="${ORCA_HISTFILE}"'
+  )
 }
 
 describePosix('local PTY shell-ready launch config', () => {
--- a/src/main/shell-templates.ts
+++ b/src/main/shell-templates.ts
@@ -169,5 +169,8 @@ esac
 # expose the same ZDOTDIR a normal zsh startup would expose.
 export ZDOTDIR="$_orca_home"
 unset _orca_home
+# Why: macOS /etc/zshrc sets HISTFILE relative to the still-wrapper ZDOTDIR
+# before this restore; re-apply the worktree-scoped path Orca injected at spawn.
+[[ -n "\${ORCA_HISTFILE:-}" ]] && export HISTFILE="\${ORCA_HISTFILE}"
 `
 }
--- a/src/main/terminal-history.test.ts
+++ b/src/main/terminal-history.test.ts
@@ -158,6 +158,8 @@ describe('terminal-history', () => {
       expect(result.histFile).toContain('terminal-history')
       expect(result.histFile).toContain('zsh_history')
       expect(env.HISTFILE).toBe(result.histFile)
+      // Why: durable copy for zsh wrapper re-export after macOS /etc/zshrc (#11044).
+      expect(env.ORCA_HISTFILE).toBe(result.histFile)
     })
 
     it('injects HISTFILE for bash', () => {
@@ -169,6 +171,7 @@ describe('terminal-history', () => {
       expect(result.shell).toBe('bash')
       expect(result.histFile).toContain('bash_history')
       expect(env.HISTFILE).toBe(result.histFile)
+      expect(env.ORCA_HISTFILE).toBe(result.histFile)
     })
 
     it('produces different HISTFILE for different worktreeIds', () => {
@@ -189,6 +192,7 @@ describe('terminal-history', () => {
       const result = injectHistoryEnv(env, 'repo-1::/path/wt', '/bin/zsh', '/path/wt')
 
       expect(env.HISTFILE).toBe('/my/custom/histfile')
+      expect(env.ORCA_HISTFILE).toBeUndefined()
       expect(result.histFile).toBeNull()
     })
 
@@ -197,6 +201,7 @@ describe('terminal-history', () => {
       const result = injectHistoryEnv(env, 'repo-1::/path/wt', '/bin/tcsh', '/path/wt')
 
       expect(env.HISTFILE).toBeUndefined()
+      expect(env.ORCA_HISTFILE).toBeUndefined()
       expect(result.shell).toBe('unknown')
       expect(result.histFile).toBeNull()
     })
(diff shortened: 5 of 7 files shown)


MinseongS:
Thanks for opening the PR. This looks aligned with the issue and the root cause.

nwparker:
Thanks for this — closing as superseded by #14863 (now merged), but with credit, because two things from here went in.

**Adopted from this PR:**

1. **Dropping a stale inherited `ORCA_HISTFILE`.** I had converged on the same `ORCA_HISTFILE` mechanism independently and missed this case entirely. An Orca terminal opened from inside another Orca terminal inherits the variable, and the zsh wrapper then re-exports the *parent* worktree's history path — the exact cross-worktree leak the feature exists to prevent. It is worse on the early-return path, where it overrides a caller-supplied `HISTFILE`. #14863 now clears it at the top of `injectHistoryEnv`, in `LocalPtyProvider` when history is disabled, and on the relay path, with three tests pinning it.
2. The `[[ -n "${ORCA_HISTFILE:-}" ]]` restore shape, which is what landed.

**What #14863 adds beyond this:** the case where Orca injects *nothing* — history isolation off, or any non-injecting path. There `/etc/zshrc` still derives `HISTFILE` from Orca's wrapper `ZDOTDIR`, so history accumulated inside the wrapper directory and the user's real history stayed invisible, with no per-worktree scoping involved at all. It also extends the same fix to the relay so remote macOS hosts are covered, and is verified by a live-zsh test that drives a real login shell through `/etc/zshrc` on both macOS and Linux.

Your diagnosis in #11044 and the mechanism here were both right. Sorry it sat for a few weeks.

innocarpe:
Thanks for the credit and for landing the stale inherited `ORCA_HISTFILE` case in #14863. Closing the fork portfolio mirror as well.

PR #11158, opened 2026-07-28
fix(automations): attribute run completion by agent session, not only paneKey

### LINKED ISSUE #10999: [Bug]: Automation run output is attributed by paneKey, so a nested `claude -p` in the same pane is captured as the run's reply
## Summary

Automation run output is correlated to a run by **terminal pane key + time window**. Any *nested* `claude` process started inside that pane — e.g. by a Claude Code plugin's `SessionStart` hook — emits agent status for the **same** `paneKey`, so its output is accepted as the automation's reply. The run is then finalized while the real agent is still mid-turn.

Since v1.4.156 shipped `fix(win): taskkill agent PTY descendant trees on stop` (#10100), the finalization also kills the still-working agent. Before that, the orphaned agent survived and completed into a transcript nobody reads, so the symptom was *wrong text*. Now it is *no output at all*. **#10100 is not the bug** — it removed the accident that was masking it.

## Environment

- Orca **v1.4.156**, Windows 10 Pro 19045, agent `claude` (Claude Code v2.1.220)
- Automation: `--workspace-mode existing`, daily RRULE, `reuseSession: false`
- Plugin present in the workspace: `remember@claude-plugins-official` v0.8.7

## Where I think it goes wrong

`src/renderer/src/hooks/useAutomationDispatchEvents.ts` (~line 378):

```ts
for (const [paneKey, entry] of Object.entries(agentStatusByPaneKey)) {
  if (paneKey !== targetPaneKey || entry.updatedAt < startedAfter) {
```

and `src/renderer/src/components/automations/automation-run-output-snapshot.ts`:

```ts
export function selectAutomationRunOutputSnapshot(
  assistantMessage: string | null | undefined,
  terminalSnapshot: AutomationRunOutputSnapshot | null
): AutomationRunOutputSnapshot | null {
  return createAutomationRunOutputSnapshotFromText(assistantMessage ?? '') ?? terminalSnapshot
}
```

`paneKey` identifies the *pane*, not the process. A nested `claude -p` launched inside that pane is, from agent-status' point of view, indistinguishable from the automation's own agent — and it finishes first, so its `assistantMessage` wins.

In our case the nested process comes from `remember`'s `SessionStart` hook: `session-start-hook.sh` calls `save-session.sh --force`, which spawns `claude -p --model haiku` via `subprocess.run` (`pipeline/haiku.py`). It completes in ~60–75 s — reliably before a real agent doing real work.

## Evidence

Across 9 runs of one automation (2026-07-23 → 2026-07-27), **only 1 of 4 scheduled runs produced a usable digest.** In every failure the nested call completed a few seconds before the run's capture timestamp; in every success either no nested call ran (the plugin's own extract gate found nothing to save) or the agent's `end_turn` landed first.

`%APPDATA%\orca\logs\daemon.log`, two runs of the same automation:

```
FAILED  (agent never reached end_turn)
2026-07-27T05:15:56.244Z  session-created   …@@9d070d42
2026-07-27T05:17:23.139Z  session-killed    …@@9d070d42  "immediate":true
2026-07-27T05:17:24.244Z  session-exited    …@@9d070d42  code:1

SUCCEEDED
2026-07-27T20:36:28.725Z  session-created   …@@ee208d3d
2026-07-27T20:38:15.610Z  session-killed    …@@ee208d3d  "immediate":true
2026-07-27T20:38:16.970Z  session-exited    …@@ee208d3d  code:1
```

The lifecycle is **identical** — same `immediate: true`, same exit code. The kill is normal finalization; the defect is what triggered it. In the failing run `session-killed` matches the automation's `lastRunAt` to the millisecond.

The failing run's Claude transcript holds **3 tool calls and no final message** — its only assistant text is a preamble, *"I'll run both tracks. Starting with the installed-version reads…"*. The agent was still working when the pane died. The run still reported `status: completed`.

Not OOM or a crash: zero Windows Application events in the surrounding 15-minute window, no WER entries.

## Minimal repro

1. In a workspace, enable a Claude Code plugin whose `SessionStart` hook spawns a nested `claude -p …` (`remember@claude-plugins-official` does, via `save-session.sh --force`).
2. Create an automation on it with `--workspace-mode existing` and a prompt whose agent takes longer than the nested call (~90 s is plenty).
3. Run it.
4. `orca automations runs --id <id> --json` → `outputSnapshot.content` holds the **nested** call's output, and the agent's transcript shows it never reached `end_turn`.

## Possible direction

Scope agent-status attribution to the process the run actually spawned (root PID / session id) rather than to `paneKey` + time window — or ignore status from descendants that the run did not start. You are far better placed than I am to judge which.

## Workaround (partial — see caveat)

Disabling the plugin at project scope in the automation's workspace:

```json
// <workspace>/.claude/settings.json
{ "enabledPlugins": { "remember@claude-plugins-official": false } }
```

Verified in isolation: a control `claude -p` in a clean folder created `.remember/` and fired the plugin's hooks; an identical run with the settings file above created no `.remember/` and fired none. After moving the automation to such a workspace the next run captured a correct digest.

**Caveat, stated honestly:** that successful run also used a rewritten prompt that made the agent roughly ten times faster to first answer. Two variables changed at once, so I cannot claim the plugin was the sole cause of the failures — a faster agent may simply have won the race. The mechanism above is what I am confident about; the workaround is what appeared to help.

Note that tuning the plugin's thresholds does **not** work: `session-start-hook.sh` calls `save-session.sh --force`, and the `min_human_messages` gate is explicitly skipped when `--force` is set.

## Not a duplicate of #2439

[#2439](https://github.com/stablyai/orca/issues/2439) (closed, v1.4.12, macOS) was a routing problem — the run panel stayed blank with `outputSnapshot: null` because the agent ran in an unlinked tab. Here the snapshot is **not** null; it is populated with the wrong process's output, and the run-linked pane is correct.

## Impact

Any automation whose workspace has a plugin that shells out to a nested agent silently delivers the wrong output — or since v1.4.156, no output at all. The run reports `status: completed` either way, so the failure is invisible.

innocarpe:
## Description
Automation runs no longer finalize early when a nested same-pane agent (e.g. Claude SessionStart `claude -p`) emits `done` on the same paneKey as the primary automation agent.

## Focused fix
- In scope: bind providerSession fingerprint after dispatch; only finalize matching session; pure tracker + tests; wire observe paths
- Out of scope: rewriting global agentStatusByPaneKey store identity for all agents

## Preserves
- Fallback to paneKey+time when no provider session is available (including OSC-only paths)
- reuseSession completion path still works when session fingerprints match

## Evidence
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/lib/automation-agent-session-attribution.test.ts src/renderer/src/hooks/useAutomationDispatchEvents.test.ts` → 26 passed

## User-regression-tradeoffs
- Nested same-type agents may still overwrite sidebar status rows; this PR only gates **automation finalization**, not the full status map UX

Fixes #10999


innocarpe:
### Sync update (`d6ba053f9`)

Address CodeRabbit: working-only fingerprint bind + ignore nested done before primary working; keep agentStatusByPaneKey reset in beforeEach

- Commit: `d6ba053f9a3196b173d1d0529437f5f1fa35c34d`
- Upstream: https://github.com/stablyai/orca/pull/11158
- Fork portfolio: https://github.com/innocarpe/orca/pull/110

innocarpe:
### Sync update (`51f454c6f6`)

Rebase onto latest upstream/main: agent-session attribution for automation completion; keep submitPrompt mocks for paste path.

- Commit: `51f454c6f68914c047a47e29fa9ff9df00d03262`
- Upstream: https://github.com/stablyai/orca/pull/11158
- Fork portfolio: https://github.com/innocarpe/orca/pull/110

mco-works:
Thanks for the focused fix. One downstream interface gap appears to remain.

Because this PR intentionally preserves paneKey/time fallback when a provider session is unavailable, an external verifier needs AutomationRun JSON to expose which attribution path was used. Could the run persist and return:

- the provider-session fingerprint used for completion
- the terminal/process incarnation identity bound to that session
- provenance that explicitly distinguishes exact session binding from pane/time fallback

Fallback may still be used internally for compatibility, but it should not be indistinguishable from a positive exact-attribution receipt. Completion, `outputSnapshot`, and the returned identity should all refer to the same provider session.

Is that persisted receipt/provenance part of #11158, or would it require a follow-up change?

innocarpe:
Addressed the exact-attribution receipt requested on #10999.

`orca automations runs --json` now persists `completionAttribution` on each run. A positive exact receipt (`kind: exact_provider_session`) is emitted only when completion is bound to the primary provider-session fingerprint; nested same-pane sessions cannot satisfy it. Pane/time fallback remains for UI compatibility as `kind: pane_time_fallback`, which is not a positive exact result.

Validation: focused attribution/dispatch/retention tests 52 passed after rebase onto current upstream/main.

innocarpe:
### Sync update (`d73ffcd7b`)

Address CodeRabbit on #11158:

- Exact receipts are authorized only by a matching `done`, so a PTY exit after a working bind stays `pane_time_fallback`.
- Provider sessions bind from live `working` samples only; historical replay no longer attaches the live session to a past working row.
- `sawWorkingAfterStart` is set only for `working`, so OSC `waiting` then `done` cannot satisfy reuse.

Validation: focused attribution/dispatch tests 39 passed after rebase onto current upstream/main.

- Commit: `d73ffcd7b0f61d8252afb7baa68d67f66702a121`
- Upstream: https://github.com/stablyai/orca/pull/11158
- Fork portfolio: https://github.com/innocarpe/orca/pull/110

innocarpe:
Address CodeRabbit: await a microtask after the nested `done` before the negative completion assertion, matching the sibling tests.

Also rebased onto current main. Persistence receipt coverage now lives in `persistence-automations.test.ts` after the test-file split.

Validation: focused attribution/dispatch/persistence tests 70 passed.

PR #11171, opened 2026-07-28
feat(git): route git clone through configured HTTP/SOCKS5 proxy

PannenetsF:
## Summary

Git clone previously ignored Orca's app-wide proxy setting (`settings.httpProxyUrl` / `settings.httpProxyBypassRules`), so users behind an HTTP or SOCKS5 proxy could not clone repositories. This PR wires that existing — already validated, persisted, encrypted, and socks5-capable — setting into all three clone transports.

The mechanism is the standard proxy **environment variables** (`ALL_PROXY` / `HTTP_PROXY` / `HTTPS_PROXY` + lowercase, plus `NO_PROXY`) that Git's libcurl backend honors. Because `ALL_PROXY` is respected, a `socks5://` URL works with no git-config flags — so the relay's strict clone-argv validator does not need to be touched.

Covered transports:

- **Local clone** (`src/main/ipc/repos.ts`) and **runtime clone** (`src/main/runtime/orca-runtime.ts`) — a new `gitCloneEnvWithProxy` helper merges the configured proxy into the non-interactive git env, and on Windows also forwards the keys to `WSLENV` so a WSL-routed clone actually receives them.
- **SSH / relay clone** — `proxyUrl` / `proxyBypassRules` are forwarded over the `git.clone` RPC (the relay rebuilds git env from the remote host and can't see the desktop's proxy). The relay re-derives env via `buildConfiguredProxyEnv`, so a forwarded value can only ever become an env-var value, never argv.

Also adds `git` / `clone` search aliases to the **Advanced Network** settings section so the proxy field is discoverable when searching for "git clone".

## Screenshots

`No visual change` — the proxy field already exists in Advanced Network settings; this PR only makes clone honor it and adds English search aliases.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

`pnpm lint` and `pnpm typecheck` pass. `pnpm test` was not run in full — the affected suites (`network-proxy`, `git-clone-proxy-env`, `ssh-git-provider`, 99 tests) pass via targeted `vitest` runs. `pnpm build` was not run locally.

New/updated tests:
- `src/main/git/git-clone-proxy-env.test.ts` — http & socks5 merge, `NO_PROXY` bypass rules, unchanged-env (same ref) when no proxy, WSLENV forwarding only on win32-with-proxy, no empty WSLENV.
- `src/main/providers/ssh-git-provider.test.ts` — forwards proxy to the `git.clone` request; omits the fields when none configured.
- `src/shared/network-proxy.test.ts` — socks5 env emits all six proxy keys.

## AI Review Report

Ran an adversarial security + correctness review of the full diff with an AI agent. Main risks checked and results:

- **Command/argument injection** — traced `params.proxyUrl` / `proxyBypassRules` from the relay RPC boundary through `buildConfiguredProxyEnv` into `spawn('git', args, { env })` (argv array, no `shell: true`). Confirmed they only ever become **env-var values**, never argv, and a value failing the `normalizeProxyUrl` scheme allowlist (`http/https/socks/socks4/socks5`) yields no env keys. **Safe.**
- **Relay validator** — `src/relay/git-exec-validator.ts` is unchanged; clone argv stays restricted and global git flags stay blocked. **Intact.**
- **Credential leakage** — proxy URLs may contain `user:pass`; verified clone failure messages route through `stripCredentialsFromMessage`, and neither the relay dispatcher/multiplexer nor `spawnClone` logs the env. **No leak.**
- **Cross-platform (macOS / Linux / Windows)** — explicitly reviewed. Local callers default to `process.platform`; WSLENV forwarding is gated to `win32` **and** proxy-present, reads the already-populated `WSLENV` before the global fallback, and never emits an empty `WSLENV`. Paths use the existing env plumbing (no hardcoded separators). No keyboard-shortcut/label surface touched.
- **SSH + folder-workspace use cases** — SSH clone forwards the proxy over RPC explicitly; folder-workspace clone flows through the same local path. Both verified.

The review found **no blockers and no should-fix issues**.

## Security Audit

- **Input handling** — proxy URL is validated by `normalizeProxyUrl` (URL parse + scheme allowlist + length cap); bypass rules normalized/length-capped. Relay re-validates via `buildConfiguredProxyEnv` so a malformed value degrades to "no proxy" rather than reaching git.
- **Command execution** — no shell; `spawn` with an argv array. Proxy values influence only the child `env`, never the command line. The relay's clone-argv allowlist is unchanged.
- **IPC / relay** — new `git.clone` RPC fields are type-guarded (`typeof === 'string'`) at the boundary and optional.
- **Secrets** — proxy credentials are never logged; existing credential-stripping on failure messages covers this path. `redactProxyUrl` remains available for diagnostics.

No follow-up required.

## Notes

- Windows/WSL: proxy keys are forwarded into `WSLENV` so clones that shell out through `wsl.exe` inherit them.
- The relay path is intentionally env-only (no `git -c http.proxy=`), keeping the strict clone-argv validator and its global-flag block untouched.
- Bypass-rules-without-a-proxy-URL is a no-op on both local and relay paths (consistent behavior).

### DIFF
--- a/src/main/git/git-clone-proxy-env.test.ts
+++ b/src/main/git/git-clone-proxy-env.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from 'vitest'
+import { gitCloneEnvWithProxy } from './git-clone-proxy-env'
+
+describe('gitCloneEnvWithProxy', () => {
+  it('merges an http proxy into the git env', () => {
+    const env = gitCloneEnvWithProxy(
+      { PATH: '/usr/bin' },
+      { httpProxyUrl: 'http://proxy.example:8080' },
+      'linux'
+    )
+    expect(env.HTTP_PROXY).toBe('http://proxy.example:8080')
+    expect(env.HTTPS_PROXY).toBe('http://proxy.example:8080')
+    expect(env.ALL_PROXY).toBe('http://proxy.example:8080')
+    expect(env.http_proxy).toBe('http://proxy.example:8080')
+    expect(env.https_proxy).toBe('http://proxy.example:8080')
+    expect(env.all_proxy).toBe('http://proxy.example:8080')
+    // Original keys are preserved.
+    expect(env.PATH).toBe('/usr/bin')
+  })
+
+  it('merges a socks5 proxy so libcurl-backed git routes through it', () => {
+    const env = gitCloneEnvWithProxy({}, { httpProxyUrl: 'socks5://127.0.0.1:1080' }, 'linux')
+    expect(env.ALL_PROXY).toBe('socks5://127.0.0.1:1080')
+    expect(env.HTTPS_PROXY).toBe('socks5://127.0.0.1:1080')
+  })
+
+  it('applies bypass rules as NO_PROXY', () => {
+    const env = gitCloneEnvWithProxy(
+      {},
+      { httpProxyUrl: 'http://proxy.example:8080', httpProxyBypassRules: 'localhost;*.internal' },
+      'linux'
+    )
+    expect(env.NO_PROXY).toBe('localhost,*.internal')
+    expect(env.no_proxy).toBe('localhost,*.internal')
+  })
+
+  it('returns the env unchanged when no proxy is configured', () => {
+    const original = { PATH: '/usr/bin' }
+    expect(gitCloneEnvWithProxy(original, undefined, 'linux')).toBe(original)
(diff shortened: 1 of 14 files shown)


PannenetsF:
@brennanb2025 Rebased onto the latest main and resolved all conflicts. Verified locally (typecheck + targeted tests pass); CI is pending maintainer approval to run. Ready for review whenever you have a moment — thanks!

PR #11173, opened 2026-07-28
feat(source-control): unstaged-only Changes diffs and toolbar refresh

### LINKED ISSUE #11133: [Feature]: Separate unstaged diffs and add a manual Source Control refresh action
### Problem or use case

The Source Control pane currently makes two common workflows confusing or unreliable:

1. When reviewing a file under **Changes**, the diff can include content that has already been staged. This makes it difficult to distinguish the remaining working-tree changes from the changes already in the index.
2. The pane does not always pick up the latest file or index state automatically, and there is no explicit way to request a fresh scan.

### Proposed solution

- Make the diff opened from **Changes** compare the working tree against the index (equivalent to `git diff`), so already staged hunks are excluded.
- Keep staged content reviewable separately under **Staged Changes** (index vs. `HEAD`).
- Add a refresh action to the Source Control pane toolbar, similar to the refresh action in VS Code Source Control.
- On refresh, re-read repository status and invalidate/reload the relevant diffs so the pane reflects the latest working-tree and index state.
- Apply the same behavior to both local and SSH-backed workspaces.

### Acceptance criteria

- If only part of a file is staged, reviewing that file under **Changes** shows only the remaining unstaged hunks.
- If all changes in a file are staged, the file no longer appears under **Changes** and remains available under **Staged Changes**.
- After a file or the Git index is changed externally, using the refresh action updates the Source Control pane and subsequent diff views.
- Refresh provides clear in-progress feedback and does not create duplicate concurrent refreshes.

VS Code Source Control provides a useful reference for the placement and behavior of the manual refresh action.

innocarpe:
## Description
Source Control **Changes** diffs now show working-tree vs index only (excluding already-staged hunks), and the SC toolbar has an explicit **Refresh** that reloads status and open diffs.

## Focused fix
- In scope: unstaged compareAgainstHead=false for Changes/edit view; pure SC open helpers; toolbar refresh (status + diff nonce + branch/history)
- Out of scope: redesign of SC layout, auto-watch policy, staged hunk partial staging UX

## Preserves
- Staged Changes still opens index vs HEAD
- Existing post-mutation refresh paths unchanged; manual refresh is additive

## Evidence
- `pnpm exec vitest run --config config/vitest.config.ts` source-control-diff-load / manual-refresh / toolbar-identity / preview-open suites → 23 passed

## User-regression-tradeoffs
- Markdown/edit “Changes” mode left pane is unstaged-only (matches VS Code mental model)

Fixes #11133

### DIFF
--- a/src/renderer/src/components/editor/useEditorPanelContentState.ts
+++ b/src/renderer/src/components/editor/useEditorPanelContentState.ts
@@ -502,7 +502,8 @@ export function useEditorPanelContentState({
       return
     }
     const current = openFilesRef.current.find((f) => f.id === activeFile.id)
-    if (!current || !isReloadableSingleFileDiffTab(current)) {
+    // Why: edit-mode Changes tabs also bump this nonce on manual SC refresh.
+    if (!current || !(isReloadableSingleFileDiffTab(current) || isChangesMode)) {
       return
     }
     setDiffContents((prev) => {
@@ -514,7 +515,7 @@ export function useEditorPanelContentState({
       return next
     })
     void loadDiffContent(current, { force: true })
-  }, [activeFile?.diffContentReloadNonce, activeFile?.id, loadDiffContent])
+  }, [activeFile?.diffContentReloadNonce, activeFile?.id, isChangesMode, loadDiffContent])
 
   useEffect(() => {
     const nonce = activeFile?.fileContentReloadNonce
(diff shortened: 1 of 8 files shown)


innocarpe:
### Sync update (`3db8e38a4`)

Address CodeRabbit: reload edit-mode Changes tabs on manual SC refresh

- Commit: `3db8e38a4aaeb25af6bb62e27434d8845d554da4`
- Upstream: https://github.com/stablyai/orca/pull/11173
- Fork portfolio: https://github.com/innocarpe/orca/pull/112

PR #11174, opened 2026-07-28
fix(codex): scope System Default status hooks without dual-file collision

### LINKED ISSUE #10947: [Bug] System Default Codex status hooks lack explicit scope and can create dual hook representations
## Summary

When Codex uses Orca's **System Default** account and **Agent status hooks** are enabled, Orca installs its status relay into the user's real `~/.codex/hooks.json`.

Executing that relay from Codex sessions launched outside Orca is not inherently wrong. Some users may intentionally want every Codex process using the System Default home to report status to Orca. The problem is that Orca does not expose or clearly describe that scope choice:

* **Orca-managed sessions only**, or
* **all Codex sessions using the System Default home**.

The global behavior also creates a representation collision when the user already declares hooks inline in `~/.codex/config.toml`:

```text
warning: loading hooks from both ~/.codex/hooks.json and ~/.codex/config.toml; prefer a single representation for this layer
```

Codex supports both representations and merges them, so this warning is non-fatal. It does, however, show that Orca has introduced a second user-layer representation without coordinating with the user's existing hook source.

## Reproduction

1. Configure Codex in Orca to use **System Default**.
2. Enable **Agent status hooks**.
3. Declare at least one user hook inline in `~/.codex/config.toml`.
4. Launch or resume Codex through Orca.
5. Observe Orca relay entries in `~/.codex/hooks.json`.
6. Launch Codex from an ordinary terminal outside Orca.
7. Observe:
   * the external process also loads the Orca relay; and
   * Codex warns that the same user layer contains hooks in both `hooks.json` and `config.toml`.

## Actual behavior

* The effective scope of **Agent status hooks** for System Default is implicit.
* Orca writes a second hook representation into the user config layer even when inline TOML hooks already exist.
* Users cannot choose whether external System Default Codex sessions should participate.
* Diagnostics do not clearly report the selected/effective scope and the files or provider sources involved.

## Expected behavior

Expose a durable, clearly described scope for System Default status hooks, for example:

1. **Orca sessions only** — status hooks apply only to Codex processes launched or explicitly managed by Orca; the shared system Codex home is not enrolled.
2. **All System Default sessions** — Codex processes outside Orca intentionally participate in Orca status reporting.

For either scope:

* Do not silently create two executable hook representations in the same Codex config layer.
* Preserve hooks owned by the user and other providers.
* Disabling or changing scope should remove only Orca-owned entries and should remain durable across launch/resume/reconciliation.
* `orca agent hooks status` (or equivalent diagnostics) should report the effective scope, hook source, and target home.

A provider-specific/plugin hook source, launch-scoped projection, or representation-aware integration could satisfy this; the issue is the missing product contract and coexistence behavior, not a requirement to use one particular implementation.

## Related, but not duplicates

* stablyai/orca#8612 covers System Default being redirected away from native `~/.codex` into an Orca-managed `CODEX_HOME`.
* stablyai/orca#9788 covers a Windows shell-profile-only custom `CODEX_HOME` that Orca cannot detect.

This issue concerns scope selection and hook-source coexistence after using the native System Default home.

## Environment

* macOS
* Orca 1.4.159
* Codex CLI 0.145.0

innocarpe:
## Description
System Default Codex status hooks default to **Orca sessions only**: do not write Orca hooks into the user’s real `~/.codex/hooks.json` (avoids dual hooks.json + config.toml representation and unexpected outside-Orca reporting). Orca-managed launches keep status via managed CODEX_HOME.

## Focused fix
- In scope: default `orca-sessions` scope, leftover real-home sweep, dual-rep guard for opt-in `all-sessions`, pure plan helpers + tests
- Out of scope: full settings UI (env opt-in `ORCA_CODEX_SYSTEM_DEFAULT_HOOK_SCOPE=all-sessions` only)

## Preserves
- Orca-launched Codex status hooks still work (managed home)
- User’s real-home hooks/trust entries
- Opt-in all-sessions path for users who want shared-home reporting

## Evidence
- `pnpm exec vitest run --config config/vitest.config.ts src/main/codex/codex-system-default-hook-scope.test.ts src/main/codex/codex-real-home-hook-install.test.ts` → 36 passed

## User-regression-tradeoffs
- Default is safer (no real-home enrollment). Users who relied on external Codex reporting to Orca must set the all-sessions env (or future UI).

Fixes #10947

### DIFF
--- a/src/main/codex/codex-real-home-hook-install.test.ts
+++ b/src/main/codex/codex-real-home-hook-install.test.ts
@@ -45,9 +45,12 @@ import {
 import { getCodexManagedHookInstallMaterial } from './hook-service'
 import { _internals as rebaseInternals } from './codex-user-hook-trust-rebase'
 
+const HOOK_SCOPE_ENV = 'ORCA_CODEX_SYSTEM_DEFAULT_HOOK_SCOPE'
+
 let fakeHomeDir: string
 let userDataDir: string
 let previousUserDataPath: string | undefined
+let previousHookScope: string | undefined
 
 function getRealHooksJsonPath(): string {
   return join(fakeHomeDir, '.codex', 'hooks.json')
@@ -75,12 +78,21 @@ function grantUnavailable(): void {
   grantMock.mockReturnValue({ lane: 'fallback', reason: 'unsupported' })
 }
 
+/** Real-home enrollment tests pin the opt-in all-sessions scope. */
+function useAllSessionsScope(): void {
+  process.env[HOOK_SCOPE_ENV] = 'all-sessions'
+}
+
 beforeEach(() => {
   grantMock.mockReset()
   fakeHomeDir = mkdtempSync(join(tmpdir(), 'orca-real-home-hooks-home-'))
   userDataDir = mkdtempSync(join(tmpdir(), 'orca-real-home-hooks-user-data-'))
   previousUserDataPath = process.env.ORCA_USER_DATA_PATH
+  previousHookScope = process.env[HOOK_SCOPE_ENV]
   process.env.ORCA_USER_DATA_PATH = userDataDir
+  // Why: install-path suites cover all-sessions enrollment; scope-default
+  // suites clear this env and assert the safer orca-sessions product default.
+  useAllSessionsScope()
   homedirMock.mockReturnValue(fakeHomeDir)
   mkdirSync(join(fakeHomeDir, '.codex'), { recursive: true })
   _internals.setLaneForTesting('pending')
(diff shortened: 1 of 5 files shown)


innocarpe:
### Sync update (`5370e1c77`)

Address CodeRabbit: orca-sessions short-circuit + rollback/sweep hardening

- Commit: `5370e1c77b32b3805b314a43d4115cce685adff0`
- Upstream: https://github.com/stablyai/orca/pull/11174
- Fork portfolio: https://github.com/innocarpe/orca/pull/113

PR #11175, opened 2026-07-28
fix(worktree): ignore shared-dir symlinks so git add -A cannot stage them

### LINKED ISSUE #11077: Shared worktree directories: git add -A stages the symlink as an absolute-path blob
## Problem

A worktree shared directory is materialized as a symlink. When the repo's ignore rule is directory-only (`node_modules/` with a trailing slash — the common spelling), that rule matches the primary checkout's real directory but never the worktree's symlink, so Git classifies the link as an ordinary untracked file.

Verified on Git 2.50.1, inside the worktree:

```
$ git check-ignore node_modules ; echo $?
1
$ git add -A && git status --porcelain
A  node_modules
$ git ls-files -s node_modules
120000 d7128ab08b155a4d48672d305173887f859ef3be 0	node_modules
$ git cat-file -p :node_modules
/Users/<user>/<primary-checkout>/node_modules
```

So `git add -A && git commit` stages a mode-120000 blob whose content is the developer's **absolute local path**. That lands in a PR, leaks a local path, and on a teammate's checkout produces a symlink pointing at a directory that does not exist on their machine.

PR #10459 filters the link out of Orca's own surfaces (local status, Create Review, workspace cleanup, worktree deletion). It cannot cover a raw `git add -A` run in a worktree terminal — which is exactly what coding agents do, and running agents in worktrees is the product's core use case.

## Why this is more urgent than it looks

This is a **widening** of an existing hazard, not a fresh one, and the widening is what matters:

- The shipped per-user shared-paths setting has the same exposure on Linux and Windows.
- On macOS it did **not**: for that mode, a non-symlink source is APFS clone-copied rather than symlinked (`mode !== 'share' && options.platform === 'darwin'` in `src/main/ipc/worktree-symlinks.ts`), so macOS users got an independent directory with no symlink to stage.
- The new project-level `share` mode deliberately never clones — a clone would defeat one-install-serves-all — so it symlinks on macOS too.

Net effect: the exposure moves onto the primary development platform, and from per-user opt-in to project-mandated for every teammate who clones a repo whose `orca.yaml` declares shared directories.

## Fix that works

Add the entry to the repository's exclude file. Verified end-to-end: `check-ignore` then matches, `git add -A` skips the link, and `git status` is clean.

```
$ git rev-parse --git-path info/exclude   # run from the worktree
<primary>/.git/info/exclude
$ printf 'node_modules\n' >> <primary>/.git/info/exclude
$ git check-ignore -v node_modules
<primary>/.git/info/exclude:7:node_modules	node_modules
$ git add -A && git status --porcelain
(empty)
```

Use the bare name with no trailing slash — a trailing slash reproduces the original directory-only problem. `git rev-parse --git-path` is available well below the Git 2.25 baseline.

## Decisions this needs before it can ship

This was deliberately left out of #10459 because it is a new mechanism with real lifecycle questions, not a one-line fix:

1. **Scope is repo-wide, not per-worktree.** `info/exclude` resolves to the common directory even when run from inside a linked worktree (verified above), so a single write affects the primary checkout and every sibling worktree. There is no per-worktree variant.
2. **It permanently widens the repo's ignore rules.** After the write, a *file* named `node_modules` in the primary checkout is also ignored. Verified.
3. **No cleanup story.** Nothing removes the line when the entry is dropped from `orca.yaml`, so the widening outlives the configuration that caused it.
4. **Concurrency.** Parallel worktree creates on one repo race on read-modify-write.
5. **WSL.** The path must be resolved through the project's Git runtime and then translated for a Windows-side write.
6. **Overlap.** If newly created worktrees no longer surface the link at all, most of #10459's filtering becomes redundant for them, while remaining necessary for worktrees created before the write and for the per-user shared-paths feature. Two overlapping mechanisms need one clear owner.

## Suggested scope

Write the exclude entry at worktree-create time, right after the shared directories are materialized, idempotently, wrapped so a failure never blocks worktree creation. Decide (3) explicitly — either accept the permanent widening and document it, or track Orca-written entries under a marker so they can be reconciled.

Found while reviewing #10459.

innocarpe:
## Description
After materializing worktree shared-directory symlinks, Orca appends root-anchored exclude rules so `git add -A` no longer stages mode-120000 absolute-path blobs for paths like `node_modules`.

## Focused fix
- In scope: common-dir `info/exclude` ensure for symlink shared paths; unit + real-Git hermetic regression
- Out of scope: replacing #10459 UI filters (complementary residual for raw terminal git)

## Preserves
- Worktree create still succeeds if exclude write fails (warn-only)
- Real directories (non-symlinks) are not unnecessarily exclude-written
- Git 2.25-compatible (filesystem + standard worktree layout only)

## Evidence
- `pnpm exec vitest run --config config/vitest.config.ts src/main/ipc/worktree-symlink-git-exclude.test.ts src/main/ipc/worktree-symlinks.test.ts` → 44 passed (incl. git add -A does not stage symlink)

## User-regression-tradeoffs
- Exclude is repo common-dir scoped (Git limitation for linked worktrees) — shared path names become ignored repo-wide

Fixes #11077

### DIFF
--- a/src/main/ipc/worktree-symlink-git-exclude.test.ts
+++ b/src/main/ipc/worktree-symlink-git-exclude.test.ts
@@ -11,7 +11,7 @@ import {
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, beforeEach, describe, expect, it } from 'vitest'
-import { createWorktreeLinkedPaths } from './worktree-symlinks'
+import { createWorktreeSharedPaths } from './worktree-symlinks'
 import {
   ensureWorktreeSharedSymlinkExclude,
   resolveWorktreeGitCommonDir,
@@ -134,6 +134,23 @@ describe('ensureWorktreeSharedSymlinkExclude', () => {
     expect(readFileSync(excludePath, 'utf8')).toBe('/node_modules\n')
   })
 
+  posixIt('still appends bare rule when only a directory-only pattern exists', async () => {
+    const primary = join(root, 'primary')
+    initRepoWithCommit(primary)
+    const worktree = join(root, 'worktree-dir-only')
+    git(primary, ['worktree', 'add', '--quiet', '-b', 'feature-dir', worktree])
+    symlinkSync(join(primary, '.gitignore'), join(worktree, 'node_modules'))
+
+    const commonDir = (await resolveWorktreeGitCommonDir(worktree))!
+    const excludePath = join(commonDir, 'info', 'exclude')
+    // Directory-only form does not cover symlinks — must still write /node_modules.
+    writeFileSync(excludePath, 'node_modules/\n')
+
+    await ensureWorktreeSharedSymlinkExclude(worktree, ['node_modules'])
+    expect(readFileSync(excludePath, 'utf8')).toContain('/node_modules\n')
+    git(worktree, ['check-ignore', '-q', 'node_modules'])
+  })
+
   posixIt('skips real directories that already match directory-only ignore', async () => {
     const primary = join(root, 'primary')
     initRepoWithCommit(primary)
@@ -153,7 +170,7 @@ describe('ensureWorktreeSharedSymlinkExclude', () => {
   })
 
   posixIt(
-    'materialize path writes exclude so git add -A cannot stage the absolute-path symlink',
+    'shared-path materialize writes exclude so git add -A cannot stage the absolute-path symlink',
     async () => {
       const primary = join(root, 'primary')
       initRepoWithCommit(primary)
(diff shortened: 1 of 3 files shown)


innocarpe:
### Sync update (`8c219897c`)

Address CodeRabbit: escape shared-symlink exclude metacharacters

- Commit: `8c219897c7a00fcdbed15730d3c6a66b276b5554`
- Upstream: https://github.com/stablyai/orca/pull/11175
- Fork portfolio: https://github.com/innocarpe/orca/pull/114

innocarpe:
### Sync update (`81195faf0`)

Rebase onto main: resolve worktree-symlinks import conflict

- Commit: `81195faf0cb822dfcacecbd13ffa23fa02e0f007`
- Upstream: https://github.com/stablyai/orca/pull/11175
- Fork portfolio: https://github.com/innocarpe/orca/pull/114

innocarpe:
### Sync update (`81195faf0`)

Rebase onto upstream/main to clear merge conflicts

- Commit: `81195faf0cb822dfcacecbd13ffa23fa02e0f007`
- Upstream: https://github.com/stablyai/orca/pull/11175
- Fork portfolio: https://github.com/innocarpe/orca/pull/114

innocarpe:
### Sync update (`b50e2e072`)

Address review: sharedPaths-only exclude + directory-only idempotency

- Commit: `b50e2e072174015aaf39f3eb25f07751c18588e1`
- Upstream: https://github.com/stablyai/orca/pull/11175
- Fork portfolio: https://github.com/innocarpe/orca/pull/114

PR #11177, opened 2026-07-28
fix(terminal): honor key-binding remaps of the Korean ₩ key

### LINKED ISSUE #11170: [Bug]: Terminal ignores DefaultKeyBinding.dict remap of the Korean ₩ key
### Operating system

macOS

### Orca version

1.4.156-rc.1

### Details

**Short summary:**
On macOS with a Korean input source, a `~/Library/KeyBindings/DefaultKeyBinding.dict` remap of the `₩` key is honored everywhere in the OS — including ordinary Chromium text fields — but is ignored in the Orca terminal, which sends the raw `₩` to the PTY.

macOS Korean input sources put `₩` (U+20A9) on the Backquote key, so typing a backtick (markdown fences, shell command substitution) is impossible without switching input source. The common workaround is a key-binding override:

```json
{
  "₩" = ("insertText:", "`");
  "~₩" = ("insertText:", "₩");
}
```

This works in Safari, Chrome, native apps and Orca's own text inputs. It does **not** work in the Orca terminal.

**How can we reproduce it?**

1. On macOS, create `~/Library/KeyBindings/DefaultKeyBinding.dict` with the contents above and restart the app.
2. Add a Korean input source (System Settings → Keyboard → Input Sources) and switch to it.
3. Open an Orca terminal and press the Backquote key.

**Expected:** `` ` `` reaches the shell (matching every other app).
**Actual:** `₩` reaches the shell.

Only reproduces while the Korean input source is active. With ABC selected, `keydown.key` is already `` ` `` and nothing is wrong.

**Root cause**

macOS applies the `DefaultKeyBinding.dict` substitution in the AppKit text input path, and Chromium surfaces the result on the **keypress/input** events — never on `keydown`. Captured from a probe page in Chromium with the Korean input source active:

```
keydown   key="₩" U+20A9  code=Backquote  keyCode=192  isComposing=false
keypress  key="₩" U+20A9  charCode=96          <- 96 = `
input     data="`" U+0060 inputType=insertText <- committed text
```

xterm.js decides the PTY payload from `keydown.key` and calls `preventDefault()`:

```js
// @xterm/xterm — evaluateKeyboardEvent
keyCode >= 48 && key.length === 1  →  result.key = ev.key   // "₩"
```

Because the keydown is consumed, Chromium suppresses the following Char event, so the keypress and input events carrying `` ` `` never fire. The raw layout character is the only thing that survives.

**Why Orca's existing guards miss it**

Orca already has machinery for exactly this "committed text differs from `keydown.key`" situation, but `₩` falls through all three:

| Guard | `₩` | Why |
| --- | --- | --- |
| `terminal-ime-native-text-candidates.ts` | not claimed | `CJK_DIRECT_PUNCTUATION_KEYS` contains `￥` (U+FFE5) but not `₩` (U+20A9); it is non-ASCII so the `forwardAsciiPunctuation` gate does not apply either |
| `xterm-bypass-policy.ts` (non-ASCII bypass) | not bypassed | scoped to `event.shiftKey` only — plain presses are excluded |
| `terminal-jis-yen-input.ts` | not covered | scoped to `code === 'IntlYen'` + `key === '¥'`, and opt-in behind `terminalJISYenToBackslash` |

The comment on the existing bypass already describes the fix — *"Bypass keydown so Chromium emits layout text via keypress"* — it just is not reachable for an unshifted key.

- Likely affects any xterm.js-based terminal, so this is less an Orca-specific defect than a gap in guards Orca already ships.
- `Option+₩ → ₩` (the `"~₩"` dict rule) is a separate matter: modifier chords are deliberately excluded from the native-text claim, and `terminalMacOptionAsAlt` consumes Option first.

I have a fix and regression tests ready and will open a PR referencing this issue.

jeongph:
## Summary

Fixes #11170.

macOS Korean input sources put `₩` (U+20A9) on the Backquote key, so a backtick is unreachable without switching input source. The standard workaround is a `~/Library/KeyBindings/DefaultKeyBinding.dict` remap:

```json
{ "₩" = ("insertText:", "`"); }
```

That remap works in every other app — including Orca's own text inputs — but not in the Orca terminal, which sends the raw `₩` to the PTY.

**Why:** AppKit applies the substitution in the text input path, and Chromium surfaces the result only on the keypress/input events. Probe capture with a Korean input source active:

```
keydown   key="₩" U+20A9  code=Backquote  keyCode=192  isComposing=false
keypress  key="₩" U+20A9  charCode=96          <- 96 = `
input     data="`" U+0060 inputType=insertText <- committed text
```

xterm derives the PTY payload from `keydown.key` and calls `preventDefault()`, which makes Chromium suppress the following Char event — so the keypress and input carrying `` ` `` never fire and only the raw layout character survives.

`₩` fell through all three existing guards: `CJK_DIRECT_PUNCTUATION_KEYS` carries `￥` (U+FFE5) but not `₩` (U+20A9); the non-ASCII bypass in `xterm-bypass-policy.ts` is scoped to `event.shiftKey`; and `terminal-jis-yen-input.ts` is scoped to `code === 'IntlYen'` behind an opt-in setting.

**Change:** add `₩` to `CJK_DIRECT_PUNCTUATION_KEYS` so `installTerminalImeNativeTextForwarder` claims the keydown, routes it around xterm, and forwards the committed glyph from the `input` event — the mechanism this file already provides for "committed text differs from `keydown.key`". Users without the dict are unaffected: the forwarder passes through whatever text actually commits, which is `₩`.

Considered and rejected: dropping the `shiftKey` condition from the non-ASCII bypass. It would cover any dict-remapped non-ASCII key, but it changes the code path for every keystroke on all non-Latin layouts (Cyrillic, Greek, Thai, …) and silently drops keys that emit no keypress. Too broad for this defect.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Three regression tests in `terminal-ime-native-text-forwarder.test.ts`, written failing first and confirmed failing before the fix:

1. `₩` keydown is a native-text candidate even when the input-source probe is stale.
2. The dict substitution (`` ` ``) is forwarded from the input event.
3. **`₩` is forwarded unchanged when no dict remap exists** — pins the behavior for users who do not remap.

`terminal-pane` suite: 193 files / 2522 tests pass. Full suite: 8 files / 43 tests fail, all pre-existing — verified identical failure counts on `upstream/main` without this change (`window.localStorage` is not implemented under Node 25.x, plus one flaky timing assertion). `oxlint` findings on this machine came only from untracked local worktree copies under `.orca/` and `.claude/`; scoped to `src config tests mobile` it is clean.

## AI Review Report

Reviewed with systematic root-cause debugging rather than symptom patching: captured real key-event evidence before proposing a fix, confirmed the xterm payload rule in the shipped bundle (`keyCode >= 48 && key.length === 1 → result.key = ev.key`), and verified `_keyPress` reads `e.charCode` and stays reachable because `_keyDownHandled` remains false when a custom handler bypasses the keydown.

Risks checked:

- **Blast radius.** `isImeNativeTextKeydownCandidate` has exactly one consumer, `terminal-ime-native-text-forwarder.ts`, which `use-terminal-pane-lifecycle.ts` installs only when `isMac`; on Windows and Linux `claimKeyEvent` is a constant `false`, so this is a no-op there. No shortcut, accelerator, label, or path handling is touched, and nothing here assumes local execution, so SSH sessions and folder workspaces behave identically.
- **Regression for non-remapping users.** Covered by test 3 above: the forwarder sends the actual committed text, so an unremapped `₩` still arrives as `₩`.
- **Modifier chords.** Unchanged — `isImeNativeTextKeydownCandidate` still rejects Ctrl/Alt/Meta, so `Option+₩` continues to fall through to `terminalMacOptionAsAlt` as before.
- **Composition safety.** Unchanged — composing keystrokes and active compositions are still rejected before the candidate check, so Hangul composition is untouched.
- **Dropped input.** If no `input` event followed a claimed keydown the glyph would be lost; the existing 100 ms disarm timer and keyup handling already cover that, and the probe shows the input event always fires for this key.

## Security Audit

No new IPC, command execution, path handling, auth, secrets, or dependency surface. The only data flow added is an existing one widened by one character: text from a focused terminal element's `InputEvent.data` is passed to `pane.terminal.input()` — the same path every typed character already takes, subject to the same PTY input guards. Input is a single committed glyph from the OS text input system, not attacker-controlled content, and it is neither parsed nor interpolated into a command. No follow-up needed.

## Notes

- macOS-only in effect, though the constant is platform-neutral.
- Reproduces only while a Korean input source is active; with ABC selected `keydown.key` is already `` ` ``.
- Likely affects any xterm.js-based terminal, so this is less an Orca-specific defect than a gap in guards Orca already ships.
- Out of scope: the companion `"~₩" = ("insertText:", "₩")` dict rule (Option + the key). Modifier chords are deliberately excluded from the native-text claim, and `terminalMacOptionAsAlt` consumes Option first. Left unchanged rather than widening the claim to modifier chords.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-ime-native-text-candidates.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-native-text-candidates.ts
@@ -41,6 +41,9 @@ const CJK_DIRECT_PUNCTUATION_KEYS = new Set<string>([
   '『',
   '』',
   '￥',
+  // Korean sources put ₩ on Backquote; its committed text can differ from
+  // `key` when DefaultKeyBinding.dict rewrites it (typically back to `).
+  '₩',
   '～',
   '·',
   '…'
--- a/src/renderer/src/components/terminal-pane/terminal-ime-native-text-forwarder.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-native-text-forwarder.test.ts
@@ -68,6 +68,18 @@ describe('isImeNativeTextKeydownCandidate', () => {
     ).toBe(true)
   })
 
+  // Why: macOS Korean sources emit ₩ from Backquote, and a DefaultKeyBinding.dict
+  // entry can rewrite it to ` — but only in the keypress/input events.
+  it('accepts the Korean won-sign key even when the input source probe is stale', () => {
+    expect(
+      isImeNativeTextKeydownCandidate(
+        keyEvent({ key: '₩', code: 'Backquote', keyCode: 192 }),
+        false,
+        DISABLED_FEATURES
+      )
+    ).toBe(true)
+  })
+
   it('accepts Vietnamese short replacement keys without enabling punctuation', () => {
     expect(
       isImeNativeTextKeydownCandidate(
@@ -262,6 +274,40 @@ describe('installTerminalImeNativeTextForwarder', () => {
     dispatchInsertText(textarea, '。')
 
     expect(sendInput).toHaveBeenCalledExactlyOnceWith('。')
+  })
+
+  it('forwards the key binding substitution for the Korean won-sign key', () => {
+    const sendInput = vi.fn()
+    const forwarder = installTerminalImeNativeTextForwarder({
+      terminalElement: element,
+      isComposing: () => false,
+      sendInput,
+      getInputSourceFeatures: () => CJK_FEATURES
+    })
+
+    expect(forwarder.claimKeyEvent(keyEvent({ key: '₩', code: 'Backquote', keyCode: 192 }))).toBe(
+      true
+    )
+    dispatchInsertText(textarea, '`')
+
+    expect(sendInput).toHaveBeenCalledExactlyOnceWith('`')
+  })
+
+  it('forwards the won sign unchanged when no key binding remaps it', () => {
+    const sendInput = vi.fn()
+    const forwarder = installTerminalImeNativeTextForwarder({
+      terminalElement: element,
+      isComposing: () => false,
+      sendInput,
+      getInputSourceFeatures: () => CJK_FEATURES
+    })
+
+    expect(forwarder.claimKeyEvent(keyEvent({ key: '₩', code: 'Backquote', keyCode: 192 }))).toBe(
+      true
+    )
+    dispatchInsertText(textarea, '₩')
+
+    expect(sendInput).toHaveBeenCalledExactlyOnceWith('₩')
   })
 
   it('forwards a plain ASCII symbol unchanged when the IME does not convert it', () => {


jeongph:
Heads up: this PR now has no remaining delta against `main`.

Rebasing `fix/terminal-korean-won-key-remap` onto `main` drops the commit entirely — the branch becomes identical to `main`. Both halves already landed through #11293 (`fix(terminal): reconcile cross-platform IME composition lifecycle`, merged 2026-07-29):

- `terminal-ime-native-text-candidates.ts` carries `'₩' in `CJK_DIRECT_PUNCTUATION_KEYS`, with the same comment.
- `terminal-ime-native-text-forwarder.test.ts` carries both cases — `accepts the Korean won-sign key even when the input source probe is stale` and `forwards the key binding substitution for the Korean won-sign key`.

The `CONFLICTING` status is just the test file: #11293 added Korean-jamo cases immediately above the won-sign block, and resolving that conflict leaves nothing behind.

Leaving this open for a maintainer to call — closing as superseded by #11293 looks right to me, but I would rather not close it myself in case any part of the fix is still wanted separately.

jeongph:
Closing as superseded — confirmed against `main` at 9a81874bc1.

The `'₩'` entry landed via #11293, and the file this PR edits (`terminal-ime-native-text-candidates.ts`) has since been deleted outright by #13128 (`refactor(terminal): return IME composition ownership to xterm`) and #13288. Rebasing now yields a modify/delete conflict with nothing left to apply.

Coverage for the remap case lives on in `terminal-ime-won-composition-order.test.ts`, which asserts composition-first ordering when the won key commits as either `\`` or `₩`, plus an immediate won remap after repeated stale composition ends. Nothing to carry forward.

PR #11184, opened 2026-07-28
fix(mobile): stop folder-workspace SC selector_not_found retries

### LINKED ISSUE #11153: [Bug]: Mobile Source Control retries selector_not_found for folder-workspace git RPCs that were never routed
## Summary

On mobile, opening Source Control for a **folder workspace** hits several git RPCs that were never routed for folder-workspace selectors. Each throws `selector_not_found`, and the loader treats that error as *transient* and retries it — so the user gets a retry loop and then a bare error string rather than a clear "not supported here" state.

Unlike desktop, mobile has **no folder-workspace gate**: `mobile/src/source-control/use-mobile-git-requests.ts:26` passes `worktree: \`id:${worktreeId}< />`straight through, so a`folder:` selector reaches the runtime.

## Routed vs unrouted

stablyai/orca#10819 routes `git.status` and `git.diff` for folder-workspace selectors. These remain unrouted and still throw:

| RPC | When it fires |
| -- | -- |
| `git.branchCompare` | auto, on every status success |
| `git.upstreamStatus` | Sync button |
| `git.fetch` / `git.pull` / `git.push` | Sync button |
| `git.history` | commit list |

## The retry loop

`mobile/src/source-control/use-mobile-source-control-loaders.ts:224-226` retries on `selector_not_found` (`SELECTOR_RETRY_COUNT = 3`, `SELECTOR_RETRY_DELAY_MS = 250`). That heuristic is right for a worktree that is still being created; it is wrong for a selector that can never resolve, and it turns a fast failure into a slow one.

The branch card also renders blank, because the merged folder-workspace status intentionally leaves `head`/`branch`/`upstreamStatus` unset — no single child repo owns them.

## Suggested direction

Two options, and the choice depends on where folder-workspace Source Control is headed generally:

1. **Detect and degrade** — recognize a folder-workspace selector on the mobile side and render an explicit unsupported state for the branch/sync/history block, instead of firing RPCs that cannot resolve. Small and safe.
2. **Route them** — give each a per-child-repo contract, the same select-one-or-refuse policy stablyai/orca#10819 uses for commit/abort. Larger, and it needs the same "is a folder workspace one repo or N?" decision the desktop panel needs.

Either way, `selector_not_found` should probably stop being retryable when the selector is structurally unresolvable.

## Context

Found while reviewing stablyai/orca#10819. Filing separately because it is independent of that PR's desktop-side question, and because mobile is currently the *only* surface where this code is reachable at all — the desktop Source Control tab is gated off for folder workspaces (`right-sidebar-activity-visibility.ts:14-16`).

Related: stablyai/orca#6357 (original report), stablyai/orca#10819 (main-process routing).

innocarpe:
## Description
On mobile Source Control for **folder workspaces** (`folder:<id>`), unrouted git RPCs (`git.branchCompare`, history, sync) throw `selector_not_found`. The loader treated that as transient and retried, so users hit a slow loop then a bare error.

## Focused fix
- Detect `folder:` worktree ids on mobile
- Do **not** retry `selector_not_found` for folder workspaces (structural, not create-race)
- Skip `git.branchCompare` for folders; fail-fast clear message for unrouted sync/history RPCs
- Branch card label falls back to “Folder workspace” when head/branch are unset
- Out of scope: main-process routing for every folder git RPC (desktop still gates SC for folders)

## Preserves
- `git.status` / routed multi-repo folder status still attempted
- Git worktree SC path unchanged (retries for real create-race still apply)
- No new desktop Source Control surface for folder workspaces

## Evidence
- Unit tests: `mobile-folder-workspace-selector.test.ts`, `mobile-git-history.test.ts` (33 related tests green)

## User-regression-tradeoffs
- Folder SC history/sync show explicit unsupported instead of hanging on retries — intentional degrade until product routes those RPCs

Closes #11153

### DIFF
--- a/mobile/src/source-control/mobile-folder-workspace-selector.test.ts
+++ b/mobile/src/source-control/mobile-folder-workspace-selector.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, it } from 'vitest'
+import {
+  formatMobileFolderAwareBranchLabel,
+  isMobileFolderWorkspaceId,
+  isMobileFolderWorkspaceUnroutedGitMethod,
+  mobileFolderWorkspaceGitRpcGuard,
+  MOBILE_FOLDER_WORKSPACE_BRANCH_LABEL,
+  MOBILE_FOLDER_WORKSPACE_UNSUPPORTED_MESSAGE,
+  shouldLoadMobileBranchCompare,
+  shouldRetryMobileStatusSelectorNotFound
+} from './mobile-folder-workspace-selector'
+
+describe('isMobileFolderWorkspaceId', () => {
+  it('detects folder: synthetic worktree ids', () => {
+    expect(isMobileFolderWorkspaceId('folder:abc')).toBe(true)
+    expect(isMobileFolderWorkspaceId('folder:')).toBe(true)
+  })
+
+  it('rejects real worktree ids and other prefixes', () => {
+    expect(isMobileFolderWorkspaceId('wt-123')).toBe(false)
+    expect(isMobileFolderWorkspaceId('id:folder:abc')).toBe(false)
+    expect(isMobileFolderWorkspaceId('')).toBe(false)
+    expect(isMobileFolderWorkspaceId('Folder:abc')).toBe(false)
+  })
+})
+
+describe('shouldRetryMobileStatusSelectorNotFound', () => {
+  it('retries for real worktrees (selector may still be registering)', () => {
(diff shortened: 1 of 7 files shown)


innocarpe:
### Sync update (`ab629eec9`)

Address CodeRabbit: allowlist routed folder git RPCs (status/diff only)

- Commit: `ab629eec9335bc34582a180fc77f803836838fc5`
- Upstream: https://github.com/stablyai/orca/pull/11184
- Fork portfolio: https://github.com/innocarpe/orca/pull/118

PR #11236, opened 2026-07-28
fix(browser): handle Windows helper stdout inheritance

jrhager84:
## Summary

- finish Windows bundled helper commands once complete JSON is available after the command process exits
- preserve the existing output bound and callback-based error handling on other platforms
- cover explicit page-session routing with a 256 KiB response split across process exit without a stdout close event

## Root cause

When a newly routed browser page starts an `agent-browser` daemon on Windows, the daemon can inherit the short-lived command process's piped stdout handle. Node's `execFile` callback waits for that handle to close even though the command has exited and already emitted complete JSON. The local runtime named pipe then reaches its 30-second idle timeout and closes before the CLI receives a response. A direct helper call succeeds because the page daemon is already running.

## Impact

Typed browser commands such as snapshot and screenshot now complete for newly routed Windows page sessions instead of timing out. macOS and Linux retain the existing `execFile` completion path.

## Validation

- `vitest`: 168 passed, 2 platform-skipped across browser bridge, runtime browser/RPC, CLI browser, and local transport suites
- `typecheck:node`
- `typecheck:cli`
- formatting check for changed files
- standard, type-aware, and React Doctor lint scans for changed files
- max-lines ratchet

### DIFF
--- a/src/main/browser/agent-browser-bridge.test.ts
+++ b/src/main/browser/agent-browser-bridge.test.ts
@@ -1,14 +1,22 @@
 /* eslint-disable max-lines */
+import { EventEmitter } from 'node:events'
 import { describe, it, expect, vi, beforeEach } from 'vitest'
 
-const { execFileMock, webContentsFromIdMock, existsSyncMock, readFileSyncMock, stdinWrites } =
-  vi.hoisted(() => ({
-    execFileMock: vi.fn(),
-    webContentsFromIdMock: vi.fn(),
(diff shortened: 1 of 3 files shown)


Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17575. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @jrhager84. No action is requested from you. Thank you.

PR #11237, opened 2026-07-28
fix(worktrees): stop terminals after external deletion

### LINKED ISSUE #10562: [Bug]: Externally deleted worktrees (git worktree remove outside Orca) strand their terminal/agent processes — reconcile paths run no teardown

nwparker:
Replaces #11053, which became permanently detached from its head branch (GitHub kept it pinned at the original commit and never re-synced after the branch was deleted and restored). Same branch, same work, plus three review fixes and a merge with `main`.

Closes #10562

## ELI5

Think of each worktree as a room, and the terminals/agents running in it as people inside that room.

When you delete a room from outside Orca (`git worktree remove` in a shell, or a CLI agent "cleaning up workspaces"), Orca notices on its next look around and **erases the room from its map**. But nobody ever told the people inside to leave. They keep running — and now there's *no door on the map to reach them*. You can only find them with `ps` on the host.

This PR adds the missing step: **before erasing the room from the map, ask the machine that owns it to clear the people out.**

The critical safety rule: *"I can't see the room"* is not the same as *"the room is gone."* If the SSH host is unreachable or the scan fails, Orca must assume the room is fine and touch nothing. Only a successful, authoritative scan that says "that room isn't here" may evict anyone. That guard is enforced twice — once in the renderer, and again by the host, which re-checks for itself before killing anything rather than trusting the request.

## What it does

- detect worktree IDs removed by an authoritative refresh before renderer state is purged
- ask the owning runtime to revalidate those IDs and sweep graph-, provider-, and registry-visible PTYs
- route duplicate repo IDs with the exact SSH connection identity instead of guessing an owner
- preserve non-authoritative scan safety and degrade cleanly against older runtimes
- retain graph-visible cleanup when an SSH PTY provider disappears during reconciliation

## Review fixes on top of the original

The original implementation had two bugs that made it silently do nothing — including one where the exact #10562 scenario still reproduced — plus one perf defect.

**1. Coalesced callers purged state without ever requesting teardown.** Teardown ran *inside* the shared scan promise, so any caller that joined an in-flight scan returned early and skipped it — while still running its own purge. Not a rare interleaving: teardown is gated on a per-caller `knownWorktreeIds` snapshot, and startup fan-out routinely has one caller start pre-hydration (empty snapshot, no-op) and another join post-hydration (should sweep, didn't). Fixed by running teardown per caller, deduped on the request it actually produces.

**2. The "authoritative recheck" read a 30s cache.** The runtime revalidated through a cache invalidated only by Orca's *own* mutations — an external `git worktree remove` invalidates nothing. The renderer uses a *different* 5s cache, so it could correctly see the deletion while the runtime still reported the worktree present, skipping the sweep. Since the renderer purges regardless and nothing retries, the leak was **permanent**. Fixed by invalidating the repo's scan cache before revalidating (destructive path only).

**3. O(N) host enumerations per sweep.** `killAllProcessesForWorktree` calls `provider.listProcesses()` — a full enumeration of every PTY on the host — once per missing worktree, with nothing shared: O(N) round-trips carrying O(N²) rows. No layer coalesces these (`SshPtyProvider` is a bare relay round-trip; the daemon router fans out to *every* adapter). This is the primary #10562 scenario, where an agent deletes many worktrees at once. Measured at 30 worktrees / 80ms RTT: **30 round-trips → 1**, wall clock 1298ms → 732ms.

Two correctness traps in that perf fix, both caught by tests written against it: sharing a *rejected* scan would let one transient relay error suppress the sweep for the whole batch (worse than the original), and `requirePhysicalStop` re-lists *after* shutdown to prove a PTY exited, so a pre-shutdown snapshot would make that proof read stale rows. Only the success path is shared, and `requirePhysicalStop: false` is pinned at the call site with the invariant documented.

Two further defects came from Greptile's bot review, both verified to reproduce before fixing:

**4. An explicit connection identity silently changed the selector grammar.** `resolveRepoSelectorForConnection` matched the selector as a bare repo id, so `path:`/`name:` selectors — valid everywhere else, and accepted by the sibling `worktree.detectedList` RPC — resolved to `repo_not_found` on this path alone. Separately, teardown rescanned via `` `id:${repo.id}` ``, and that re-resolution throws `selector_ambiguous` when an id is duplicated across hosts even though the caller's selector was unambiguous. The renderer always sends a bare repo id so no user-facing path hits this today, but the RPC is public. Both paths now share one selector grammar, and teardown scans the already-resolved repo.

**5. The shared snapshot leaked into provider internals.** The snapshot proxy passed itself as the `Reflect.get` receiver, so a provider whose own `shutdown()` re-read state via `this.listProcesses()` would silently get the cached snapshot instead of the live host. No shipped provider does this, but adding one must not quietly change teardown semantics. Non-`listProcesses` members are now bound to the target.

## Merge with main

`main` refactored the detected-worktree refresh into a lease/admission model (`DetectedWorktreeRefreshOutcome`, `directSshAuthority`, `mergeFetchedWorktrees`). The per-caller teardown is reapplied on that structure — requested before every `admitted` return on both the runtime and lease paths — and `knownWorktreeIds` is optional, where absent means "nothing to reconcile", never "purge everything".

## Validation

| Suite | Result |
|---|---|
| `src/main/runtime/` + `src/renderer/src/store/` | **5076 passed**, 5 skipped |
| `pnpm typecheck:node` / `typecheck:web` | clean |
| `pnpm check:max-lines-ratchet` | OK — no new bypasses |

Seven regression tests added, each failing on the commit before its fix. No existing assertion was changed or relaxed; the only edits to existing tests are additions and a mock-shape update to match main's refactor.

## Adversarial checks that came back clean

- **SSH disconnected / scan failure → no teardown.** Both the runtime (`ok: false` when the provider is absent or throws) and the IPC path return `authoritative: false`, short-circuiting before any sweep. The #10562 critical guard holds on both sides, and the runtime re-derives it independently rather than trusting the renderer.
- **Folder workspaces.** Enumerated from persisted metadata, so never spuriously "missing"; sibling instances sharing one directory don't sweep each other.
- **Repo-ID prefix confusion.** `repo-1` vs `repo-10` doesn't leak — IDs are compared for equality, not prefix.
- **Duplicate repo IDs across hosts.** `resolveRepoSelectorForConnection` requires exact connection identity and throws `selector_ambiguous` rather than guessing.

## Note for reviewers

The renderer discards the teardown result and purges regardless of whether the sweep succeeded. That's right for UI responsiveness, but it means the sweep gets exactly one chance — which is what made defect 2 permanent rather than transient. A follow-up backstop (reconciling on SSH reconnect, cf. #9820) would add defense in depth.


nwparker:
## Merge-readiness status

Branch is current with `main` (merged twice during review; latest at `13c193a`) and **all 23 CI checks pass**. `mergeable: true`.

| | |
|---|---|
| CI | **23/23 pass** (e2e skipped — no e2e path changes; Greptile is advisory, non-blocking) |
| `src/main/runtime/` + `src/renderer/src/store/` | **5073 passed**, 5 skipped |
| worktrees + reconciliation + runtime (post-merge) | **1144 passed** |
| `typecheck:node` / `typecheck:web` | clean |
| `check:max-lines-ratchet` | OK — no new bypasses |

### Final review summary

Three defects were found and fixed during review, each with a regression test that fails on the commit before its fix:

1. **Coalesced callers purged state without requesting teardown** — the original #10562 leak still reproduced verbatim (0 teardown RPCs issued, state purged anyway).
2. **The authoritative recheck read a 30s cache** an external `git worktree remove` never invalidates — and since the renderer purges regardless with no retry, the leak was permanent.
3. **O(N) host enumerations per sweep** — 30 relay round-trips carrying O(N²) rows for a 30-worktree cleanup, reduced to 1.

Plus two correctness traps inside the perf fix itself, both caught by tests written against it (sharing a *rejected* scan; `requirePhysicalStop` reading a pre-shutdown snapshot).

### Merge interaction worth a reviewer's eye

`main` landed a lease/admission refactor (`DetectedWorktreeRefreshOutcome`, `directSshAuthority`, `mergeFetchedWorktrees`) mid-review. The teardown was reapplied on that structure and verified to fire before **both** `status: 'admitted'` returns (runtime-environment and local/direct-SSH lease paths) and on **none** of the `not-admitted` paths.

That refactor also strengthens the #10562 safety invariant for free: `directSshAuthority` now bails out before any refresh when an SSH target isn't connected, so a disconnected host can't even reach the teardown call. That's pinned by a new test (`never stops terminals for a disconnected SSH target`) so a future change can't silently regress it into killing PTYs on an unreachable host.

`knownWorktreeIds` is optional on the options type — absent means "nothing to reconcile", never "purge everything".

### Known follow-up (not blocking)

The renderer discards the teardown result and purges regardless of whether the sweep succeeded. That's correct for UI responsiveness, but the sweep gets exactly one chance — which is what made defect 2 permanent rather than transient. A backstop that reconciles on SSH reconnect (cf. #9820) would add defense in depth.

nwparker:
## Greptile P2 was real — and worse than reported

Greptile flagged the double `resolveRepoSelectorForConnection` call as redundant work ("harmless because the store doesn't mutate between the two awaits"). I verified it rather than taking the severity at face value, and the redundancy is load-bearing: **it can fail the entire sweep.** Fixed in `8ff7d13`.

Two distinct defects behind that one symptom:

**1. An explicit connection identity silently changed the selector grammar.** `resolveRepoSelectorForConnection` matched the selector as a *bare repo id*:

```ts
const repoId = repoSelector.startsWith('id:') ? repoSelector.slice(3) : repoSelector
```

So on this path alone, `path:` and `name:` selectors — valid everywhere else, and accepted by the sibling `worktree.detectedList` RPC — never match anything. A connection identity should only *narrow* the candidate set, never reinterpret it.

```
PROBE result: null err: repo_not_found
PROBE shutdowns: 0
```

After the fix, the same call resolves and sweeps:

```
PROBE result: {"stoppedWorktreeIds":["r1::/repo/gone"]}
PROBE shutdowns: 1
```

**2. Re-resolving an already-resolved repo can throw.** Teardown rescanned via `` `id:${repo.id}` ``. When a repo id is duplicated across hosts (local + SSH — the case this PR's `resolveRepoSelectorForConnection` exists to handle), that second lookup throws `selector_ambiguous` even though the caller's original selector was unambiguous. Now it scans via the already-resolved repo through a new `listDetectedWorktreesForResolvedRepo` seam.

The renderer always sends a bare repo id, so **no user-facing path hits this today** — but the RPC is public and reachable by CLI/runtime clients, and the sibling detected-list method accepts these selectors. This is a trap for the next caller, not a live outage.

Also extracted the selector matching that both resolution paths now share, so the two can't drift again. Four existing tests were repointed from the public wrapper to the new internal seam — behavior unchanged; they were stubbing a method the destructive path no longer calls.

**Validation:** 5075 passed / 5 skipped across `src/main/runtime/` + `src/renderer/src/store/`; `typecheck:node`, `typecheck:web`, max-lines ratchet, and oxlint all clean. New regression test `resolves non-id selectors when a connection identity is supplied` fails on the prior commit with `repo_not_found`.

nwparker:
## Second Greptile finding also real — fixed in `c3a201b`

Greptile's "Comments Outside Diff" flagged that `withSharedProcessSnapshot` passes the proxy as the `Reflect.get` receiver, so prototype methods invoked through it run with `this` bound to the proxy. A provider whose own `shutdown()` re-read state via `this.listProcesses()` would silently receive this sweep's cached snapshot instead of a live enumeration.

Verified it reproduces:

```
PROBE listCalls: 1   (1 => shutdown reused the cached snapshot)
```

A provider that lists twice internally saw only **one** real call — the batching leaked past the calls it was built for.

**Fix:** bind non-`listProcesses` members to `target` rather than the proxy, so only the sweep's own calls share the snapshot:

```
PROBE listCalls: 3   (1 shared sweep scan + 2 genuine internal reads)
```

Greptile suggested a comment; I closed it structurally instead. An implicit constraint documented only in prose is exactly the kind of thing that gets violated later — and the failure mode here is silent (stale rows, no error). The comment now explains *why* the binding is deliberate rather than asking future maintainers to remember a rule.

No shipped `IPtyProvider` does this today, so nothing was broken in practice. Pinned by `does not serve the shared snapshot to provider-internal listProcesses`, which fails on the prior commit (`expected 1 to be 3`).

**Validation:** 5076 passed / 5 skipped; `typecheck:node`, `typecheck:web`, max-lines ratchet, oxlint all clean.

---

Both of Greptile's findings were real, and both were things my own two review passes missed — worth recording plainly. Its severity call was off in the first case (labelled redundant work; it could fail the sweep outright), but the findings themselves were sound.

nwparker:
## Green — ready to merge

Both bot reviews re-ran against `c3a201b` and came back clean:

| | |
|---|---|
| Required CI | **24/24 pass** (e2e skipped — no e2e path changes) |
| Greptile | **pass**, confidence **5/5** (was 4/5), no new findings |
| CodeRabbit | its one actionable finding was the selector issue, marked `✅ Addressed in commits 8ff7d13 to c3a201b` |
| Mergeable | `mergeable_state: clean` |
| Local suites | **5076 passed**, 5 skipped |

Both bots' outstanding inline comments are the same selector defect, now fixed.

### Final tally: five defects, none of which the original implementation would have surfaced at runtime as an error

1. **Coalesced callers purged state without requesting teardown** — the original #10562 leak reproduced verbatim (0 teardown RPCs, state purged anyway).
2. **Authoritative recheck read a 30s cache** an external `git worktree remove` never invalidates — and with no retry, the leak was permanent.
3. **O(N) host enumerations per sweep** — 30 relay round-trips carrying O(N²) rows for a 30-worktree cleanup, now 1.
4. **Connection scoping silently changed the selector grammar**, and re-resolving an already-resolved repo could throw `selector_ambiguous` on the exact duplicate-id case this PR handles.
5. **Shared snapshot leaked into provider internals** via the Proxy receiver — silent stale reads for any provider that re-lists inside `shutdown()`.

Plus two traps inside the perf fix itself, caught by tests written against it (sharing a *rejected* scan; `requirePhysicalStop` reading a pre-shutdown snapshot).

Seven regression tests, each verified to fail on the commit before its fix. No existing assertion was changed or relaxed — the only edits to existing tests are additions, a mock-shape update for main's lease/admission refactor, and four stubs repointed to the seam the destructive path now uses.

### Reviewer notes

- **The safety invariant is enforced twice and pinned.** Renderer and runtime each independently require an authoritative scan; `main`'s new `directSshAuthority` gate additionally bails before any refresh when an SSH target is disconnected. `never stops terminals for a disconnected SSH target` locks that in, so a future change can't regress it into killing PTYs on an unreachable host.
- **The sweep gets exactly one chance.** The renderer purges regardless of the teardown result — right for UI responsiveness, but it's what made defect 2 permanent rather than transient. A backstop reconciling on SSH reconnect (cf. #9820) would add defense in depth. Not blocking.

PR #11238, opened 2026-07-28
fix(mobile): declare happy-dom for standalone tests

brennanb2025:
## Summary

- Declare `happy-dom: ^20.9.0` in the standalone mobile project.
- Update `mobile/pnpm-lock.yaml` with pnpm 10.24.0, matching the root `packageManager` pin used by CI.

## Root cause

Two mobile terminal-webview test files request `// @vitest-environment happy-dom`, but `mobile/package.json` did not declare that environment. Because mobile is a separate pnpm project, a standalone `cd mobile && pnpm install && pnpm test` cannot resolve the package unless a repository-root install happens to provide it for upward ESM resolution.

Current mobile CI is not blind to these tests: `.github/workflows/mobile.yml` installs repository-root dependencies before mobile dependencies so the mobile typecheck can resolve shared runtime imports. That root install materializes `node_modules/happy-dom`, which Vitest currently resolves. This change removes that implicit workflow coupling and makes the mobile project own its test dependency.

## Evidence

I temporarily hid the root `node_modules/happy-dom` symlink and compared the rebased parent with this commit using pnpm 10.24.0 and the mobile frozen lockfile:

| Tree | Test files | Tests | Errors | Exit |
| --- | ---: | ---: | ---: | ---: |
| parent | 349 passed | 2,578 passed / 2 skipped | 2 `ERR_MODULE_NOT_FOUND` worker-start errors | 1 |
| fixed | 351 passed | 2,590 passed / 2 skipped | 0 | 0 |

The 12-test delta is the two happy-dom test files. Reverting the manifest and lockfile change restores the explicit exit-1 failure.

## Verification

- `pnpm 10.24.0 install --frozen-lockfile`
- `pnpm test --maxWorkers=4`
- `oxlint`
- `oxfmt --check .` (1,023 files)
- `tsc --noEmit`

### DIFF
--- a/mobile/package.json
+++ b/mobile/package.json
@@ -72,6 +72,7 @@
     "acorn": "8.15.0",
     "esbuild": "0.25.4",
     "expo-module-scripts": "^55.0.2",
+    "happy-dom": "^20.9.0",
     "oxfmt": "^0.52.0",
     "oxlint": "^1.71.0",
     "react-test-renderer": "19.2.6",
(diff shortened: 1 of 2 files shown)


brennanb2025:
Final verification on rebased `a5fd58bd8d`:

- Root cause: two mobile test files select the happy-dom Vitest environment, but the standalone mobile pnpm project did not declare it. Current CI still runs them because its root install (needed for typecheck) supplies `node_modules/happy-dom` for upward resolution.
- Change: add the dev dependency and the pnpm-10.24.0 lockfile delta only.
- With root happy-dom hidden, reverting the manifest/lockfile is RED: 349 files, 2,578 passed / 2 skipped, 2 `ERR_MODULE_NOT_FOUND` errors, exit 1. Restored fix is GREEN: 351 files, 2,590 passed / 2 skipped, 0 errors, exit 0.
- Post-rebase gates: pnpm 10.24.0 frozen install, full mobile suite, oxlint, oxfmt check, and `tsc --noEmit` all pass; same-model high-effort review returned clean with no edits or perf concerns.

No physical-device or emulator QA was run: this changes only a devDependency and generated test lockfile, with no app runtime path.

PR #11240, opened 2026-07-28
fix(mobile): close terminal session tabs authoritatively

### LINKED ISSUE #6927: [Bug]: Mobile cannot close tabs
### Operating system

Other

### Orca version

0.0.17

### Details

When I am connected to my Orca server via my phone (it is running in headless server mode, paired with server not desktop), when I try to close a tab on one of my workspaces, it does not close the tab. The only way to close the tab is to close it on the desktop application.

### LINKED ISSUE #7345: [Bug]: Terminal closed from mobile app reappears as a brand-new terminal
### Operating system

Other

### Orca version

Mobile app: v0.0.18 (3). Desktop: recent build (code references below are against current `main`, a074ed618)

### Details

Short summary:
Closing a terminal from the Orca mobile app never sticks — the terminal disappears for a moment, then reappears as a **brand-new terminal** (fresh shell). Closing the same terminal from the desktop app works fine. Markdown/file/browser tabs close fine from mobile; only terminal tabs resurrect.

How can we reproduce it?
1. Pair the mobile app with a running desktop session.
2. Create a terminal from mobile (or use one only ever driven from mobile).
3. Long-press the terminal tab on mobile → Close.
4. The terminal briefly disappears, then comes back as a brand-new terminal.

Root cause (traced in code):
Terminal tabs are the only tab type on mobile whose Close goes through the handle-only `terminal.close` RPC instead of the session-tab close path, and the two halves of that path re-create the terminal:

1. Mobile Close calls `handleCloseTerminal` → `terminal.close` (`mobile/app/h/[hostId]/session/[worktreeId].tsx:4110`). Host-side `closeTerminal` (`src/main/runtime/orca-runtime.ts:16363`) **only kills the PTY** — it does not remove the session tab or drop its persisted binding.
2. On PTY exit the runtime keeps the tab in the mobile session snapshot and republishes it as `status: 'pending-handle', terminal: null` (`orca-runtime.ts:18548-18550`, republished via `touchMobileSessionSnapshotsForPty` at `:6959`). The closed tab is still present — and still marked active.
3. `handleCloseTerminal` removes the terminal only from the legacy `terminals` list; it never removes the tab from `sessionTabs`, never clears `activeSessionTabId`, and sets no tombstone. When the pending snapshot arrives, the auto-activation effect (`[worktreeId].tsx:4187`) sees an active pending terminal tab and calls `session.tabs.activate` with `notifyClients: false` → host hits `shouldMaterializePendingTerminal` (`orca-runtime.ts:3841-3862`) → `createHeadlessMobileSessionTerminal` spawns a **new PTY into the same tab**. That is the resurrection.

This exact resurrection class was already fixed for the other tab types: `handleCloseSessionTab` tombstones closed tabs (`[worktreeId].tsx:4154`), and host `closeMobileSessionTab` documents the authoritative-teardown requirement (`orca-runtime.ts:4082-4087`). The terminal Close button just doesn't use either fix.

Suggested fix:
- **Mobile:** route the terminal action sheet's Close through the session-tab close path (`session.tabs.close` + tombstone + clear `activeSessionTabId`) instead of handle-only `terminal.close`.
- **Host (defense in depth):** in `runtime.closeTerminal`, when the handle maps to a runtime-owned mobile session tab, perform the same authoritative teardown as `closeMobileSessionTab` (kill PTY + drop persisted binding + prune and emit snapshot) so stale clients cannot resurrect the tab.

Likely shares a root cause with #6927 (headless-server pairing: tab never closes from mobile) — same close path, different manifestation.

brennanb2025:
## Summary

- Route terminal long-press Close through the authoritative session-tab close path when the terminal handle belongs to a session tab.
- Keep the handle-only close as a fallback for live handles that have no session-tab owner.
- Cover single terminals, split-leaf ownership, and the fallback path.

Fixes #6927
Fixes #7345

## Merge ordering

Depends on #11251 (`brennanb2025/mobile-9717-dup-sessions`, issue #9717) — **land #11251 first.**

This PR correctly routes more terminal closes through the authoritative
`session.tabs.close` path, which prunes `sessionTabs` and clears
`activeSessionTabId`. That legitimately reaches the empty-session-tab-list state
far more often, and on `main` that state can trigger an unsolicited replacement
terminal — the duplicate-session bug in #9717. Landing this PR first therefore
makes #9717 more reachable without introducing it.

#11251 fixes the auto-create side (`use-initial-session-terminal-autocreate.ts`)
by treating an authoritative empty snapshot as a real empty workspace. The two
branches combine cleanly and do not require an atomic merge, but this PR should
not land before #11251. #11251 records the same ordering constraint.

## Root cause

Terminal long-press Close was the only mobile tab close that called handle-only `terminal.close`. That removed the PTY-facing terminal but did not prune the mobile `sessionTabs`, clear `activeSessionTabId`, or record the local close tombstone. The host could therefore republish the same active tab as pending, after which the pending-terminal activation effect materialized a fresh PTY into the same tab id.

#7345 reports this exact terminal resurrection path. #6927 sees it as every tab being unclosable because a headless `orca serve` session normally publishes terminal tabs; #10486 is separate because its trigger is viewing/activation and does not exercise close.

## Reliability contract

- **Invariant (`terminal-session.explicit-close-retirement`):** an explicit mobile terminal-tab close uses the session-tab owner transaction, removes the local mirror, clears active ownership, and records the tombstone; an unmatched live handle retains the existing fallback.
- **Failure source:** #6927 and #7345.
- **Oracle:** action-sheet tests assert the matching session tab is closed, the owning split leaf is selected, and unmatched handles use the fallback. Replacing the new routing branch with the old handle-only call makes 2 of 4 focused tests fail.
- **Gate:** existing `terminal-session.explicit-close-retirement` is partial; this PR adds the mobile route-level oracle and records live-device validation as a gap.
- **Provider/platform coverage:** the pure mobile routing contract is provider- and path-agnostic, so local, daemon, SSH, WSL, folder workspaces, and git worktrees are unaffected below the existing close RPC. The full suite and static gates ran on macOS; a physical phone/relay journey was not run.
- **Performance budget:** one bounded session-tab lookup occurs when the action sheet is built and is reused for bulk-close anchoring. No polling, provider inventory, render loop, listener, subprocess, or startup work is added.
- **Diagnostics:** focused assertion names distinguish owner routing, split selection, and fallback behavior.
- **Residual gap:** mobile does not interpret `RuntimeMobileSessionTabCloseResult.refused`, but `session.tabs.close` hardcodes user intent and cannot currently reach a refusal branch. Handling a future user-close refusal is intentionally out of this P1 scope.

## Verification

Against parent `d3681f6306`:

- Parent: 351 test files, 2,590 passed, 2 skipped
- This PR: 351 test files, 2,593 passed, 2 skipped
- `cd mobile && pnpm install` using the repository-pinned pnpm 10.24.0
- Full mobile Vitest suite with `--maxWorkers=2`
- Full mobile oxlint
- Full mobile `oxfmt --check .`
- Mobile `tsc --noEmit`
- Non-vacuity mutation: old routing produces 2 focused failures

No device validation is claimed.

### CI note (pre-existing unrelated flake)

`tests 12/16` failed once on this commit. All 233 shard files and 2,247 tests
**passed**; the job exited 1 solely on `Unhandled Rejection: Error: Unable to
remove watcher: Invalid argument`, thrown from `@parcel/watcher`'s Linux
inotify backend (`inotify_rm_watch` -> `EINVAL`).

Not caused by this PR:

- The root Vitest config's `include` matches no file under the repo-root
  `mobile/` directory, so this mobile-only diff leaves shard 12's file set
  byte-identical to the parent — it cannot shift shard contents.
- Re-running the same job on the identical tree passed.
- The blamed file (`src/main/plugins/plugin-content-safety.test.ts`) contains no
  watcher code; Vitest attributes an async native error to whichever test was
  in flight.

The latent hazard is in `src/main/ipc/parcel-watcher-in-process-fallback.ts`:
the cancellation path deliberately swallows unsubscribe failures ("so they never
surface as unhandled rejections"), but the normal teardown path awaits
`subscription.unsubscribe()` unguarded. Desktop main-process code, unrelated to
this mobile fix, and intentionally left out of scope here.

### DIFF
--- a/mobile/app/h/[hostId]/session/[worktreeId].tsx
+++ b/mobile/app/h/[hostId]/session/[worktreeId].tsx
@@ -5216,6 +5216,7 @@ export default function SessionScreen() {
           onRename: setRenameTarget,
           onClear: (target) => void handleClearTerminal(target),
           onClose: (target) => void handleCloseTerminal(target),
+          onCloseSessionTab: (tab) => void handleCloseSessionTab(tab),
           bulkCloseActions
         })}
         onClose={() => setActionTarget(null)}
--- a/mobile/src/session/mobile-terminal-action-sheet-actions.test.ts
+++ b/mobile/src/session/mobile-terminal-action-sheet-actions.test.ts
@@ -9,24 +9,38 @@ vi.mock('lucide-react-native', () => ({
   SquareTerminal: vi.fn()
 }))
 
+type SheetArgs = Parameters<typeof getMobileTerminalActionSheetActions>[0]
+
+function buildActions(overrides: Partial<SheetArgs> = {}) {
+  return getMobileTerminalActionSheetActions({
+    target: { handle: 'terminal-1' },
+    tabs: [],
+    isTabChatView: () => false,
+    nativeChatTranscriptIsLocalReadable: true,
+    onDismiss: vi.fn(),
+    onToggleChat: vi.fn(),
+    isPhoneMode: () => false,
+    onToggleDisplayMode: vi.fn(),
+    onRename: vi.fn(),
+    onClear: vi.fn(),
+    onClose: vi.fn(),
+    onCloseSessionTab: vi.fn(),
+    ...overrides
+  })
+}
+
+const terminalTab = (id: string, handle: string | null) => ({
+  type: 'terminal',
+  id,
+  terminal: handle
+})
+
 describe('getMobileTerminalActionSheetActions', () => {
   it('defers Rename until after the action sheet closes', () => {
     const target = { handle: 'terminal-1' }
     const onDismiss = vi.fn()
     const onRename = vi.fn()
-    const actions = getMobileTerminalActionSheetActions({
-      target,
-      tabs: [],
-      isTabChatView: () => false,
-      nativeChatTranscriptIsLocalReadable: true,
-      onDismiss,
-      onToggleChat: vi.fn(),
-      isPhoneMode: () => false,
-      onToggleDisplayMode: vi.fn(),
-      onRename,
-      onClear: vi.fn(),
-      onClose: vi.fn()
-    })
+    const actions = buildActions({ target, onDismiss, onRename })
 
     const rename = actions.find((action) => action.label === 'Rename')
     expect(rename).toMatchObject({ closeBeforePress: true })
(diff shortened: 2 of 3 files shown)


brennanb2025:
Root cause: terminal long-press Close alone used handle-only `terminal.close`, leaving the session-tab mirror, active id, and tombstone state behind so pending activation could materialize a fresh PTY. This PR resolves the handle owner once and routes matching terminals—including split leaves—through authoritative `session.tabs.close`, preserving handle-only fallback for genuinely unmatched handles. Verified on final SHA: parent 351 files / 2,590 passed / 2 skipped → PR 351 / 2,593 / 2; full mobile suite (`--maxWorkers=2`), oxlint, `oxfmt --check`, and `tsc --noEmit` pass, old routing makes 2/4 focused tests fail, and the same-model nested review returned CLEAN with targeted host split/SSH tests green. No physical phone/simulator journey was run; the current unrelated CI failure is the plugin-content-safety watcher teardown shard (its tests passed before teardown).

Issue #6995, opened 2026-07-01
[Bug]: Orca mobile Android, Korean Hangul input is decomposed into separate jamo characters.


watson-water:
### Operating system

Other

### Orca version

v0.0.17(3)

### Details

On Orca mobile Android, Korean Hangul input is decomposed into separate jamo characters.

Example:

Expected: 안녕하세요

Actual: ㅇㅏㄴㄴㅕㅇㅎㅏㅅㅔㅇㅛ

Environment:

- Device: Galaxy S25 Ultra

- Keyboard: Samsung Keyboard / Korean 2-set

- Orca mobile app version: [version]

- Android version: [version]

- Tailscale enabled, but the issue also appears to be input-related, not network-related.

This looks like an IME composition/preedit handling issue in the chat input field.

lkjsays:
Also reproducing on **iOS — iPhone 17 Pro**. This is not Android-only; the title/labels (`os:mobile-beta` / Android) undersell the scope.

**Environment**
- Device: iPhone 17 Pro
- Keyboard: iOS system Korean (2-set / 두벌식)
- Input surface: Orca mobile chat/agent input field (remote input)

**Symptom** — identical to the original report:
- Expected: `안녕하세요`
- Actual: `ㅇㅏㄴㄴㅕㅇㅎㅏㅅㅔㅇㅛ` (decomposed jamo, no syllable composition)

So the IME composition/preedit handling issue affects **both Android (Samsung Keyboard) and iOS system keyboard**, which points to the mobile input field consuming intermediate/per-jamo change events instead of waiting for the composed (committed) text. Likely the same root cause as the terminal-side #7000, but in the chat input surface. +1 on P1.

hanjoonchoe:
I dug into this with a full local repro environment and can narrow the root cause significantly.

## Setup

- Android 14 emulator (arm64), Orca mobile dev build from current `main`
- Paired against `mobile/scripts/mock-server.ts` (with a locally added `session.tabs.list` handler so the session screen finishes loading, and a logger on `terminal.send` to capture exactly what the app sends)
- **Gboard, Korean 2-set (두벌식)** — typed `안녕하세요` key-by-key into the terminal **live input** field

## Result: does NOT reproduce with Gboard

The captured `terminal.send` stream shows correct syllable composition the whole way:

```
"아" → "안" → "ㄴ" → "녀" → "녕" → "ㅎ" → "하" → "핫" → "하" → "세" → "셍" → "세" → "요"
```

[...]

Issue #7427, opened 2026-07-05
[Bug]: Orca Mobile terminal cannot enter Japanese via flick or romaji IME


Chachamaru127:
## Summary
Orca Mobile terminal input cannot enter Japanese text. Both Japanese flick/kana input and romaji-to-kana/kanji conversion fail, so Japanese users cannot reliably type commands, prompts, or free text into the mobile terminal.

## Reproduction
1. Open Orca Mobile and connect to a paired desktop host.
2. Open a session terminal.
3. Focus the terminal input surface / direct terminal input.
4. Switch the phone keyboard to Japanese.
5. Try either:
   - kana flick input, e.g. `こんにちは`
   - romaji input with conversion, e.g. type `nihongo` and convert/commit `日本語`

## Expected
The IME composition should be allowed to complete, and the committed Japanese text should be sent to the PTY exactly once after commit.

## Actual
Japanese composition is interrupted or cleared before a valid committed string reaches the terminal. This affects both flick/kana input and romaji conversion.

## Impact
This blocks a basic day-to-day workflow for Japanese users on Orca Mobile. It also affects SSH/remote terminal use, because the mobile client forwards input through the same terminal send path.

## Likely cause / code pointers
The mobile terminal live input path appears to treat every `TextInput.onChangeText` value as committed terminal bytes, sends it immediately, then clears the native text field:

- `mobile/app/h/[hostId]/session/[worktreeId].tsx`:
  - `handleLiveInputChange` normalizes `text`, calls `sendLiveTerminalInput(activeHandle, normalizedText)`, then resets `liveInputCapture` and calls `liveInputRef.current?.setNativeProps({ text: '' })`.
  - the live capture `TextInput` wires `onChangeText={handleLiveInputChange}`, `onKeyPress={handleLiveInputKeyPress}`, and `onSubmitEditing={handleLiveInputSubmit}`.
- `mobile/src/terminal/terminal-keyboard-type.ts` already returns `keyboardType="default"`, so non-Latin IMEs should remain selectable. The likely breakage is composition lifecycle handling, not keyboard availability.
- `mobile/mobile-terminal-direct-input-default.md` documents that direct terminal input is the default for first-seen terminal tabs, so this path is now the default path for mobile terminal users.

IME pre-edit text is not equivalent to committed terminal text. Clearing the native `TextInput` during composition likely cancels or corrupts the IME state, which explains why both flick input and romaji conversion fail.

Related issues to keep in mind:
[...]

PR #11246, opened 2026-07-28
fix(mobile): recompose iOS NFD Hangul before terminal send

### LINKED ISSUE #11235: [Bug]: iOS Korean input reaches the terminal as decomposed (NFD) jamo — third mechanism behind #6995
### Operating system

iOS

### Orca version

main @ 8b57e6e180

### Details

Splitting a distinct root cause out of stablyai/orca#6995. That issue is filed and triaged as **Android / Samsung Keyboard**, where the leading theory (@hanjoonchoe's analysis in that thread) is that `autoCorrect={false}` maps to Android's no-suggestions `inputType` and Samsung Keyboard answers by disabling IME composition outright.

**@lkjsays reported the same visible symptom on iOS** (iPhone 17 Pro, iOS system Korean 2-set / 두벌식) in that thread. That report is a *different* mechanism and is not fixed by any Android input-prop change — on iOS the field composes correctly and the symptom survives `autoCorrect={false}`. Filing it separately so it doesn't get closed along with the Android fix.

Expected: `안녕하세요`  <br>Actual: `ㅇㅏㄴㄴㅕㅇㅎㅏㅅㅔㅇㅛ`

### Root cause

Apple text input hands React Native **decomposed (NFD)** text. One Hangul syllable arrives as 2–3 conjoining jamo:

```
안  NFC → U+C548                    (1 code point)
안  NFD → U+110B U+1161 U+11AB      (3 code points)
안녕하세요  NFD → 12 code points instead of 5
```

The field renders those identically, which is why composition *looks* fine on device. The break is downstream, in the terminal live-input mirror (`mobile/src/terminal/terminal-live-hangul-mirror.ts`).

`computeTerminalLiveMirrorStep` compensates for React Native exposing no composition events by **holding only the trailing code point** as "the syllable still being composed" and streaming everything before it to the PTY. That heuristic assumes *one code point per syllable*, which holds for precomposed NFC input. Under NFD a syllable spans 2–3 code points, so each leading jamo falls outside the hold window, is committed to the PTY on its own, and is never recombined. There is no DEL correction because from the mirror's point of view nothing was mistaken — the prefix really is stable.

Replaying the real pipeline (field state → `normalizeTerminalTextInput` → `computeTerminalLiveMirrorStep` → payload) over the 두벌식 keystroke sequence for `안녕하세요` with NFD field states, the PTY receives:

```
[ 'ᄋ', 'ᅡ', 'ᆫ', 'ᄂ', 'ᅧ', 'ᆼ', 'ᄒ', 'ᅡ', 'ᄉ', 'ᅦ', 'ᄋ', 'ᅭ' ]   ← 12 lone jamo
```

A terminal that does not perform conjoining-jamo shaping renders those as separate cells, which is exactly the reported string.

This also explains why the two reports look identical but need different fixes:

|  | Android (stablyai/orca#6995) | iOS (this issue) |
| -- | -- | -- |
| Field composition | killed by the IME | works correctly |
| String handed to JS | per-jamo, uncomposed | composed but NFD |
| Fixed by input props | yes | no |

The same decomposition affects Latin diacritics typed on iOS (`tiếng` → `t i e◌̂◌́ n g`), it is just far less visible there because combining marks re-stack over the base glyph in most terminals.

### Steps to reproduce

1. iOS device, system Korean keyboard (2-set / 두벌식).
2. Open a session, enable terminal live input.
3. Type `안녕하세요`.
4. The hidden capture field shows `안녕하세요`; the terminal shows decomposed jamo.

### Notes

* Not transport-specific: the decomposition happens in the mobile client before `terminal.send`, so local, SSH and folder-workspace hosts are all affected identically.
* The fix must **not** write a corrected value back into the field — that is what kills iOS dictation/IME (stablyai/orca#7925, and the existing warning comment in the session route). Normalizing only the outbound PTY text keeps that constraint intact.
* The correct normalization form is **NFC, never NFKC**. NFKC folds compatibility jamo (`ㅇㅏㄴ` U+3147 U+314F U+3134 → U+C544 U+1102) and full-width forms into different characters, which would silently rewrite a command.

### Related

* stablyai/orca#6995 — original report; Android/Samsung Keyboard mechanism, and the thread where @lkjsays reported the iOS case
* stablyai/orca#7925 — iOS dictation dies when JS writes back a value differing from native text
* [#5222](<https://github.com/stablyai/orca/issues/5222>) — iOS smart-dash rewriting, handled at the same normalization choke point

brennanb2025:
## Summary

- NFC-normalize outbound terminal text so Apple NFD Hangul reaches the live mirror as complete syllables.
- Keep the React Native field write-back verbatim for iOS dictation/IME stability, and deliberately avoid lossy NFKC folding.
- Add real mirror-pipeline coverage for Hangul and Latin NFD plus compatibility-form guards.

Fixes #11235. Thanks @lkjsays for the iOS reproduction report in #6995.

Test: NFC(안녕하세요) == 안녕하세요

## Screenshots

No visual change.

## Testing

- [x] pnpm --dir mobile lint
- [x] pnpm --dir mobile typecheck
- [x] pnpm --dir mobile test — 352 files, 2,596 passed, 2 skipped
- [x] pnpm --dir mobile format:check
- [ ] pnpm build — not run; no native or build-system changes
- [x] Added regression tests; removing the NFC normalization makes four new assertions fail

## AI Review Report

A same-model independent review returned clean. It checked the real live-mirror pipeline, raw field write-back, NFC versus NFKC fidelity, neighboring input paths, and per-keystroke cost; no correctness, regression, or meaningful performance concerns were found. The change is platform-neutral outside the documented iOS source behavior and touches no shortcuts, labels, paths, shell behavior, Git/provider behavior, or Electron platform branches; SSH and folder workspaces use the same pre-transport mobile path.

## Security Audit

Reviewed the command-input transformation for silent rewriting risk. NFC only recomposes canonically equivalent text; tests prove compatibility jamo, full-width forms, and circled characters are preserved, and the change introduces no auth, secret, dependency, path, IPC, or command-execution surface.

## Notes

The rebased branch bundle loaded on an iPhone 17 Pro simulator running iOS 26.5, and a connected session plus real terminal surface rendered successfully. Simulator automation could not inject Korean IME composition events end to end, so device-level Korean typing remains unverified; the regression is proven through the production mirror algorithm with real NFD field states.

image
image
image
image

### DIFF
--- a/mobile/src/terminal/terminal-ios-hangul-nfd-input.test.ts
+++ b/mobile/src/terminal/terminal-ios-hangul-nfd-input.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, it } from 'vitest'
+import {
+  buildTerminalLiveMirrorPayload,
+  computeTerminalLiveMirrorStep
+} from './terminal-live-hangul-mirror'
+import { normalizeTerminalTextInput } from './terminal-text-input-normalization'
+
+// 두벌식 (2-set) composition states for 안녕하세요, as the field shows them
+// keystroke by keystroke: ㅇ ㅏ ㄴ ㄴ ㅕ ㅇ ㅎ ㅏ ㅅ ㅔ ㅇ ㅛ.
+const DUBEOLSIK_FIELD_STATES = [
+  'ㅇ',
+  '아',
+  '안',
+  '안ㄴ',
+  '안녀',
+  '안녕',
+  '안녕ㅎ',
+  '안녕하',
+  '안녕핫',
+  '안녕하세',
+  '안녕하셍',
+  '안녕하세요'
+] as const
+
+const DEL = '\x7f'
+
+// Minimal PTY line model: DEL erases one code point, everything else appends.
+function applyToPty(pty: string, payload: string): string {
+  let next = Array.from(pty)
+  for (const codePoint of Array.from(payload)) {
+    if (codePoint === DEL) {
+      next = next.slice(0, -1)
+      continue
+    }
+    next.push(codePoint)
+  }
+  return next.join('')
+}
+
+// Replays handleLiveInputChange → applyLiveInputMirror, then the settle-timer
+// commit that flushes the last held syllable.
+function replayLiveInput(fieldStates: readonly string[]): string {
+  let sentText = ''
+  let heldText = ''
+  let pty = ''
+
+  const step = (fieldText: string, commitHeld: boolean): void => {
+    const mirror = computeTerminalLiveMirrorStep(sentText, normalizeTerminalTextInput(fieldText), {
+      commitHeld
+    })
+    pty = applyToPty(pty, buildTerminalLiveMirrorPayload(mirror))
+    sentText = mirror.nextSentText
+    heldText = mirror.heldText
+  }
+
+  for (const fieldText of fieldStates) {
+    step(fieldText, false)
+  }
+  step(sentText + heldText, true)
+  return pty
+}
+
+describe('iOS decomposed (NFD) Hangul live input', () => {
+  it('Given an NFD field (iOS) When the user types 안녕하세요 Then the PTY receives composed syllables, not jamo', () => {
+    // Given: Apple text input hands React Native decomposed conjoining jamo, so
+    // one syllable arrives as 2-3 code points instead of one.
+    const nfdStates = DUBEOLSIK_FIELD_STATES.map((state) => state.normalize('NFD'))
+    expect(Array.from(nfdStates.at(-1) ?? '')).toHaveLength(12)
+
+    // When
+    const pty = replayLiveInput(nfdStates)
+
+    // Then: the mirror holds only the trailing code point, so without
+    // recomposition every leading jamo is committed on its own (#6995 iOS).
+    expect(Array.from(pty)).toHaveLength(5)
+    expect(pty).toBe('안녕하세요')
+  })
+
+  it('Given a precomposed NFC field (Android/Gboard) When the same keys are typed Then the PTY result is unchanged', () => {
+    // Given / When
+    const pty = replayLiveInput(DUBEOLSIK_FIELD_STATES)
+
+    // Then
+    expect(pty).toBe('안녕하세요')
+  })
+
+  it('Given NFD Latin diacritics When mirrored Then combining marks reach the PTY already composed', () => {
+    // Given / When: iOS also hands back NFD for Vietnamese/European accents.
+    const pty = replayLiveInput(['tie', 'tiế', 'tiếng'].map((state) => state.normalize('NFD')))
+
+    // Then
+    expect(pty).toBe('tiếng')
+    expect(Array.from(pty)).toHaveLength(5)
+  })
+})
--- a/mobile/src/terminal/terminal-text-input-normalization.test.ts
+++ b/mobile/src/terminal/terminal-text-input-normalization.test.ts
@@ -11,4 +11,29 @@ describe('normalizeTerminalTextInput', () => {
   it('keeps ASCII hyphens unchanged', () => {
     expect(normalizeTerminalTextInput('git checkout -- file')).toBe('git checkout -- file')
   })
+
+  it('recomposes decomposed Hangul that iOS hands back as conjoining jamo', () => {
+    const decomposed = '안녕하세요'.normalize('NFD')
+    expect(Array.from(decomposed)).toHaveLength(12)
+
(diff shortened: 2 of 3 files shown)


brennanb2025:
Thanks again @lkjsays for the precise iOS report — it led to a separate NFD mechanism rather than the Android/Samsung path. The review loop is clean, all 2,596 mobile tests pass, and the four new assertions go red when NFC recomposition is removed. Device caveat: the branch bundle and live terminal loaded on an iOS 26.5 simulator, but automated Korean IME keystrokes were unavailable, so I am not claiming end-to-end keyboard validation.

brennanb2025:
Physical-device A/B run — **this did not reproduce, so I would not merge it as a fix for #6995 yet.**

Setup: iPhone 14, iOS 26.5.2, dev build of this branch, paired to a desktop runtime over LAN. Typed 안녕하세요 with the iOS Korean 2-set keyboard into a terminal live input.

| Build | Result |
|---|---|
| This branch (NFC normalize present) | `안녕하세요` — 5 composed syllables |
| Same build, `text.normalize('NFC')` removed via hot reload | `안녕하세요` — 5 composed syllables |

Identical. So on this device the normalize is **not load-bearing**: iOS handed React Native already-composed text, and the premise that Apple returns NFD here does not hold on iOS 26.5.2.

The change itself is still safe — NFC is a verified no-op on composed input, and NFKC was correctly avoided — but the tests only prove the normalize recomposes synthetic NFD strings; nothing demonstrates iOS produces them. That is inference, and the device run contradicts it.

One real observation from the same run, unrelated to NFD: mid-composition after a backspace, a lone jamo (`ㄴ`) briefly reached the terminal before the mirror's DEL repair cleaned it up. That is transient in-composition leakage — the flicker/destructive-intermediate-bytes behaviour described on #7427 — not decomposition.

Suggested next step: confirm with the original reporter what iPhone and iOS version they were on. If they were on an older iOS, that likely explains the difference and this fix is still correct for them; if they are on 26.x, the root cause is something else and this PR should be reframed or closed.

Holding this open rather than merging.

nwparker:
Thanks @brennanb2025 — and genuinely, thanks for the device A/B on this. That's the part that settled it, and it's the harder thing to publish about your own PR.

I re-ran the pipeline on current `main` and I think we close both. Two reasons, in order of weight.

**1. The mechanism no longer exists.** #15007 replaced the one-trailing-code-point hold with the text system's marked-text report, and it holds the *entire* unsent preedit region. That one-code-point window was the whole bug: an NFD syllable spans 2–3 code points, so every leading jamo fell outside it. Replaying the 두벌식 sequence for 안녕하세요 with NFD field states:

| rule | writes to the pty |
|---|---|
| old one-code-point hold | **12 separate writes** — `ᄋ ᅡ ᆫ ᄂ ᅧ ᆼ ᄒ ᅡ ᄉ ᅦ ᄋ ᅭ`, exactly as reported |
| current, marked-text reported (iOS) | one write, whole string |
| current, no report (Android fallback) | one write, whole string |

No fragmentation left. The residual difference is 12 conjoining jamo instead of 5 precomposed syllables in a single atomic write — canonically equivalent. And where the mirror still splits under NFD (settle timer firing mid-syllable), NFD is actually *cheaper* than NFC: the trailing jamo appends, where NFC costs a DEL and a retype. The PR also no longer applies — its test imports `terminal-live-hangul-mirror`, which #15007 deleted.

**2. The premise is still unreproduced.** Your iOS 26.5.2 run showed Apple handing React Native composed text with and without the normalize. The tests prove NFC recomposes a synthetic NFD string; nothing shows iOS produces one. You named that gap yourself and I couldn't close it either — there's no capture or fixture anywhere in the repo containing real iOS NFD.

For what it's worth I did check whether we *should* normalise, before realising we didn't have to. A mature reference terminal implementation doesn't normalise keystroke or IME text on the way to the pty at all — one has no normalization code in its tree; the other has it only for text being drawn into the on-screen grid, behind a setting that defaults off, with every key-mapper's insert transform a literal identity. Normalise what you draw, never what you send. Orca already agrees with that, apparently by accident: no normalization on any terminal input path, desktop or mobile, and every `normalize('NFC')` site in the app is filesystem-path comparison.

The counter-argument is real and worth naming: NFD at a shell prompt won't match an NFC filename on a byte-exact filesystem — macOS APFS resolves either form, Linux won't. If NFD ever does show up we'll need an answer for that. But it isn't the answer to a bug nobody can reproduce.

The code was correct and the NFC-not-NFKC reasoning was exactly right. A different fix just removed the failure first.

Issue #9756, opened 2026-07-21
[Bug]: macOS 'access other apps' data' prompt keeps reappearing — possible from user running 'pn build' locally


AmethystLiang:
### Operating system

macOS

### Orca version

Unknown (reporter may be on a local `pnpm build` / packaged prod-like instance rather than a release install)

## Summary

macOS shows the TCC dialog **"Orca" wants to access other apps' data** (`kTCCServiceSystemPolicyAppData` / system policy AppData). The agent/folder-access path is expected to trigger this kind of prompt when needed, but if the dialog **keeps reappearing** after the user clicks **Allow**, the grant is not sticking — macOS is not successfully persisting permission for the app identity that is asking.

Dialog text from the report (zh-CN):

> **"Orca" 想访问其他 App 的数据。**  
> 单独存放 App 数据可让你更容易管理隐私和安全。  
> [不允许]  [允许]

## Hypothesis (not confirmed) — possible from user running `pn build` locally

One plausible cause: the user built a **local packaged production instance** via something like:

```bash
pnpm build
# or other local package/dist flows that produce a .app with electron-builder
```

Local packaged builds and the official release currently share the same electron-builder **`appId` / CFBundleIdentifier**:

- `com.stablyai.orca` (see `config/electron-builder.config.cjs`)

That can confuse macOS TCC / code-signing identity bookkeeping:

- TCC grants are keyed by code identity (bundle id + signing identity / team, path, CDHash, etc.).
- A **local unsigned / differently signed** `.app` that still presents as "Orca" with the **same bundle id** as the store/GitHub release build can:
  - fail to inherit grants already given to the official app, or
  - cause grants to appear to "not stick" when the user switches between local build and released app, or
  - leave TCC in a confused state where the prompt keeps returning.

Note: `pnpm dev` / `dev-stable-name` can use a **different** bundle id (`com.stablyai.orca.dev` in the dev wrapper). This report is specifically about **local prod-like packaging** (`pnpm build` / electron-builder), which intentionally uses the production `appId`.

**Treat this as a hypothesis, not a root-cause conclusion.** The same dialog is also known from other, non–local-build scenarios (see Related).

## Other possible causes (need careful triage)

Please do **not** assume local build is the only explanation. Known related cases:

| Issue | Symptom | Likely mechanism |
| --- | --- | --- |
| #6996 / #7003 | TCC AppData for terminal children (`op`, etc.) | Child process identity collapsed into Orca; fixed via `login(1)` wrapper |
[...]

PR #11250, opened 2026-07-29
feat(updater): add hourly dev channel and build switching

nwparker:
## What

Adds an **hourly macOS build channel** plus a **dev-only surface for switching channels and jumping to any published build**, including older ones.

Two halves:
1. `.github/workflows/hourly-mac-build.yml` — cron-hourly, mac-only, no tests, signed but not notarized.
2. A hidden Release channel picker in Settings → Updates for moving between stable / RC / hourly and pinning an exact build.

## Screenshots

**Default — the picker is hidden.** Updates looks exactly as it does today:

default

**Option-click the "Updates" header to reveal it.** Real data from the live GitHub API — here on RC, offering a downgrade from the running `1.4.160-rc.3` to `1.4.159-rc.1`:

rc channel

**Jump to a past build.** The list goes back through real historical tags to `1.4.153-rc.1`:

build list

**Hourly channel**, showing the danger note and the graceful error path (`stablyai/orca-hourly` does not exist yet — see Before merging):

hourly channel

## Three decisions worth reviewing

**1. Hourly publishes to a separate repo, not `stablyai/orca`.**

`updater-prerelease-feed.ts` resolves every update check from the main repo's releases atom feed. That feed exposes **only the 10 newest releases** — currently `v1.4.160-rc.2` down to `v1.4.157-rc.0`.

At 24 hourly tags a day, hourlies would fill all 10 slots within ~10 hours, and then:
- stable users (`includePrerelease: false`) filter to non-prerelease versions → **zero candidates → stable updates silently stop**
- RC users see hourly tags as semver-newest → **pushed onto unvetted builds**

So hourlies go to `stablyai/orca-hourly`. The main feed is untouched, and the dev picker reads the paginated REST API (only on explicit interaction, so its rate limit never touches background checks).

**2. Hourly builds carry the release signing identity — only notarization is skipped.**

`mac.appId` becomes `com.stablyai.orca.local` unless `ORCA_MAC_RELEASE=1` (#9756). Squirrel.Mac validates the replacement bundle's **signature**, so an hourly on the local identity could not install over an installed Orca — defeating the point. `ORCA_MAC_HOURLY=1` therefore implies the release path and turns off *only* `notarize`, which in-place updates never check.

Tradeoff: a manually-downloaded hourly zip is Gatekeeper-quarantined. Installing through the updater is unaffected.

**3. The version tail is stripped to the base.**

`1.4.160-hourly.<stamp>`, not `1.4.160-rc.3-hourly.<stamp>`. Keeping the RC tail makes every hourly semver-**newer** than the RC it was cut from, so an ordinary RC check would offer untested builds to RC users. Stripping to the base parks hourlies below both `rc.N` and stable (`hourly` < `rc` alphabetically) — reachable only by an explicit pinned jump. Locked by a test.

## Safety

A pinned jump sets `allowDowngrade`, so it must never leak into routine checks. It releases the feed on **every** settle path — available, not-available, error, and abandoned-card dismissal — and `restoreReleaseUpdateSource` resets `allowPrerelease` so a jump can't silently opt later background checks into the RC channel. Background checks are deferred while a pin is in flight so nothing repoints the feed mid-download.

## Testing

- 47 new/updated tests across channel semantics, the feed resolver, version stamping, and electron-builder identity (bundle id, notarize flag, publish target).
- 135 existing updater tests pass unchanged.
- Full suite: 39,493 pass. The 2 failures in `src/relay/agent-exec-handler.test.ts` are **pre-existing on a clean tree** (verified by stashing) and unrelated.
- Screenshots above are the real app driven over CDP, not mockups.

## Before merging

**Done:** [`stablyai/orca-hourly`](https://github.com/stablyai/orca-hourly) exists — public (the updater fetches assets unauthenticated, so a private repo would 404 every download), issues/wiki/projects off, README explaining the atom-feed constraint.

**Remaining — one-time setup, and it never needs redoing:**

```
bash config/scripts/setup-hourly-release-token.sh
```

Auth is a **GitHub App**, not a PAT. A PAT expires, and the hourly build would then fail silently on a schedule nobody watches. An App's private key has no expiry, and it belongs to the org rather than to whoever created it, so it survives that person leaving.

The script walks through creating the App (Contents: Read and write on `orca-hourly` only), then stores `HOURLY_RELEASE_APP_ID` and `HOURLY_RELEASE_APP_PRIVATE_KEY`. The key is redirected from the downloaded `.pem` straight into `gh secret set` on stdin, so its contents never enter a shell variable, argv (world-readable via `ps`), or the terminal. It refuses to run under `xtrace`.

The workflow mints a short-lived installation token per run via `actions/create-github-app-token`. Those live one hour, which is ample here — no tests, no notarization, no Windows signing, so a run is pack + upload. Retry budgets and the job timeout are sized to that rather than copied from the release pipeline, whose `3×45` publish budget exists for notarization and SignPath.

Until the secrets exist the Hourly tab shows the "No releases repository found" state in the screenshot above — degraded, not broken. The workflow is `if: github.repository == 'stablyai/orca'` and no-ops when main has not moved, so merging first is inert.

Once set:

```
gh workflow run hourly-mac-build.yml --repo stablyai/orca -f force=true
```

### DIFF
--- a/.github/workflows/hourly-mac-build.yml
+++ b/.github/workflows/hourly-mac-build.yml
@@ -288,8 +288,12 @@ jobs:
       # prune step) must not delete a release that already went live and that users
       # may already be installing. A job killed by the outer timeout runs no steps
       # at all — which is exactly why the release stays a draft until verified.
+      # Why cancelled() too: a run stopped from the Actions UI is not a failure(),
+      # so without it a manual cancel mid-publish would strand the draft.
       - name: Discard the draft release on failure
-        if: failure() && steps.release.outputs.tag != '' && steps.publish_live.outcome != 'success'
+        if: >-
+          (failure() || cancelled()) && steps.release.outputs.tag != '' &&
+          steps.publish_live.outcome != 'success'
         shell: bash
         env:
           GH_TOKEN: ${{ steps.app_token.outputs.token }}
--- a/.github/workflows/hourly-mac-build.yml
+++ b/.github/workflows/hourly-mac-build.yml
@@ -1,0 +1,242 @@
+name: Hourly macOS Dev Build
+
+# Why: gives developers a signed macOS build of main every hour that the in-app
+# updater can install directly, without waiting for an RC cut.
+#
+# Deliberately narrow scope:
+#   - macOS only. Other platforms keep using RC/stable.
+#   - No tests, no lint, no e2e. This channel trades safety for latency; PR CI
+#     and release-cut remain the gates that matter.
+#   - Signed but NOT notarized. Squirrel.Mac validates the replacement bundle's
+#     signature, not its notarization, so in-place updates still work while the
+#     ~10min notary round trip is skipped 24x a day.
+#
+# Artifacts publish to stablyai/orca-hourly, never to stablyai/orca: the main
+# repo's releases atom feed exposes only its 10 newest entries, so 24 hourly
+# tags a day would evict every stable/RC entry and break updates for real users.
+
+on:
+  schedule:
+    # Top of every hour. Skipped automatically when main has not moved.
+    - cron: '0 * * * *'
+  workflow_dispatch:
+    inputs:
+      force:
+        description: Build even if main has not moved since the last hourly
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+
+concurrency:
+  group: hourly-mac-build
+  cancel-in-progress: false
+
+env:
+  HOURLY_REPO: stablyai/orca-hourly
+  # Keep ~3 days of history so a regression can be bisected across a weekend.
+  HOURLY_RETAIN_COUNT: 72
+
+jobs:
+  build-hourly-mac:
+    if: github.repository == 'stablyai/orca'
+    runs-on: blacksmith-6vcpu-macos-15
+    timeout-minutes: 60
+    env:
+      NODE_OPTIONS: --max-old-space-size=4096
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          ref: main
+          fetch-depth: 0
+
+      # Why: main is often idle overnight. Rebuilding an unchanged commit burns a
+      # runner hour and adds a redundant tag to the retention window.
+      - name: Check whether main moved since the last hourly
+        id: freshness
+        shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.HOURLY_RELEASE_TOKEN }}
+          FORCED: ${{ github.event_name == 'workflow_dispatch' && inputs.force }}
+        run: |
+          set -euo pipefail
+          head_sha="$(git rev-parse HEAD)"
+          echo "head_sha=$head_sha" >>"$GITHUB_OUTPUT"
+          if [[ "$FORCED" == "true" ]]; then
+            echo "should_build=true" >>"$GITHUB_OUTPUT"
+            echo "Forced dispatch; building $head_sha."
+            exit 0
+          fi
+          # The previous hourly records its source commit in the release body.
+          last_body="$(gh release list --repo "$HOURLY_REPO" --limit 1 --json tagName \
+            --jq '.[0].tagName // empty' 2>/dev/null || true)"
+          if [[ -z "$last_body" ]]; then
+            echo "should_build=true" >>"$GITHUB_OUTPUT"
+            echo "No prior hourly release found; building $head_sha."
+            exit 0
+          fi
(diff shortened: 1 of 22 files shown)


nwparker:
Thanks both — went through all 8. Seven were valid and are fixed in b671ede; one I'm pushing back on. Also rebased onto `main` (was 164 behind; conflicts were only in the generated locale catalogs, resolved by taking main and re-running `sync:localization-catalog`).

### Fixed

**Stale `listBuilds` responses** (greptile P1) — valid, and the mount path isn't the only trigger: rapid channel clicks stack requests too. Added a request-sequence ref; superseded responses return early and don't own the spinner (clearing `loading` from a stale request would flash "no builds" mid-load).

**`releaseChannelOverride` can't be cleared** (greptile P2) — valid, and worse than cosmetic: it meant simply opening the panel left background checks pinned to whatever you last clicked. Took the suggested fix — selecting the running build's own channel now clears the override.

**Missing `!isPinnedBuildCheck()` in `update-available`'s success branch** (coderabbit) — good catch, this was a real leak. `recordCompletedUpdateCheck()` persisted `lastUpdateCheckAt`, so browsing the picker pushed the next *real* background check out by 24h.

**Unvalidated `releaseChannelOverride` hydration** (coderabbit) — valid; added an `isReleaseChannel` type guard in the shared module rather than an inline cast, so main and renderer share one definition.

**`parseHourlyVersionStamp` accepts bad input** (coderabbit) — confirmed both halves before fixing:
```
not-a-version-hourly.202602300000 => 2026-03-02T00:00:00.000Z
1.4.160-hourly.202602300000       => 2026-03-02T00:00:00.000Z
1.4.160-hourly.202613010000       => 2027-01-01T00:00:00.000Z
```
Now anchored on the whole version and round-trips the parsed fields. Impact was display-only (inputs are pre-validated tags), but it was inconsistent with `isHourlyVersion`, which was already anchored. Tests cover invalid base, Feb 30, month 13, hour 25, and real leap day 2028.

**Job timeout vs retry budgets** (coderabbit) — valid, and there's a worse variant you didn't name: the verify step defaults to `if: success()`, so if **publish** exhausted its retries the cleanup was *skipped entirely* and the broken release survived. Timeout tuning alone wouldn't have caught that.

Fixed structurally instead: the release is now created as a **draft** and flipped live only after the manifest check. Drafts are absent from `listReleaseBuilds`, so any death in that window — including a hard timeout kill that runs no steps at all — leaves nothing user-visible. Added a failure handler to discard the draft, gated on `steps.publish_live.outcome != 'success'` so a later prune failure can't delete a release that already went live. Budgets aligned anyway (install 3×10 + publish 2×30 = 90, +~35 overhead, under 150).

Two knock-ons from the draft change: dropped `--cleanup-tag` on the draft path (an unpublished draft has no git tag), and excluded drafts from the freshness and retention queries so a leftover can't be read as "the last build".

**`persist-credentials: false`** (coderabbit) — valid and free; the job only reads this repo and every write goes to the hourly repo via `HOURLY_RELEASE_TOKEN` in env.

### Not fixing

**Probe tag cleanup in `setup-hourly-release-token.sh`** — the probe creates a release with `draft=true`, and GitHub doesn't create the git tag until a release is published. So there's no orphan tag to clean up, and `--cleanup-tag` would error on a draft. Left as-is; the delete removes the only artifact that exists. Happy to be corrected if you're seeing something I'm not.

nwparker:
Addressed in 209917a — both were valid, one was a real bug.

**`set +x` before the `xtrace` check** — correct, and worse than "the guard is dead code": I verified it, and `bash -x` ran straight through, tracing exactly the key handling the guard exists to prevent.

```
$ bash -x setup-hourly-release-token.sh
+ set -euo pipefail
+ set +x        # <- guard already bypassed
```
Now tests before disabling; confirmed it refuses, and a normal invocation still reaches the prompt.

**`cancelled()` in the draft cleanup** — valid, took it. A run stopped from the Actions UI isn't a `failure()`, so a manual cancel mid-publish stranded the draft. It stays gated on `publish_live.outcome != 'success'` so this still can't delete a release that already went live.

**Actionlint / Blacksmith label** — not taking this one. The repo has no actionlint config or CI step (`grep -rl actionlint` over `.github/`, `package.json`, `config/` returns nothing), and `release-mac-build.yml` already uses `blacksmith-6vcpu-macos-15` on `main`. That's your linter's runner table being out of date, not something this PR introduces — adding config for a linter the repo doesn't run would be unrelated scope. Worth doing repo-wide if you adopt actionlint later.

PR #6580, opened 2026-06-28
fix: install agent skills project-scoped so the PromptScript target lands (#4642)

nwparker:
## Description

Fixes #4642 by replacing Orca's unusable global PromptScript install/update commands with home-scoped project installs that land in the skill source Orca already discovers.

## Evidence of fix

**Before (live):** `npx skills add ... --skill orchestration --global` printed `Installation complete` / `Installed 1 skill`, then `PromptScript does not support global skill installation`, exited 0, and created no PromptScript `SKILL.md`.

**After (live):** the project-scoped `-y` form exited 0, created `~/.agents/skills/orchestration/SKILL.md` plus `skills-lock.json`, emitted no PromptScript failure, and Orca classified the install as a visible `home` source. GitHub checks are green.

## ELI5

Orca asked the installer to put a skill “everywhere,” but the one format Orca reads refuses global installation. The fix installs it in the user's home folder, where Orca already looks.

## User-facing trade-offs

- Installs are home-scoped rather than machine-global; the removed global path never installed Orca's PromptScript target successfully.
- Surfaced/copyable commands first move to the user's home so results do not depend on the caller's current project.
- The POSIX wrapper was live-smoked; native Windows home scoping is covered by focused tests rather than a live Windows shell run.

---

## 🧑‍🤝‍🧑 Impact & ELI5

**1) What's broken today (ELI5).** Anyone who tries to turn on Orca's extra agent abilities (the orchestration skill, computer-use, the Linear integration) by clicking the in-app "Install" button hits a silent dead end. The installer runs, cheerfully prints "✓ Installed 1 skill," and finishes with a green light — but the one piece Orca actually needs never gets installed. So Settings keeps showing "Not installed," the user clicks install again, gets another fake success, and the skill never appears. It looks like the app is lying to you. This hits every user trying to enable these features through the app (the issue was reported on Windows/WSL2, but the broken command was used everywhere), and it's a hard blocker for those features — not a cosmetic annoyance.

[...]

Issue #4642, opened 2026-06-04
[Bug]: npx skills add --global silently fails the PromptScript target but exits 0 — skills never install for Orca


jcroy:
### Operating system

windows/wsl2

### Orca version

 1.4.45

### Details

Summary

  Installing the orchestration or computer-use skills with the documented --global flag appears to succeed — the CLI prints ◇ Installation complete /
  ✓ Installed 1 skill and exits with code 0 — but it has no effect in Orca. The failure for the PromptScript target (the format Orca actually
  consumes) is printed as a trailing footnote, masked by the success banner and the 0 exit code:

  ■  Failed to install 1
     ✗ orchestration → PromptScript: PromptScript does not support global skill installation

  Most of the ~17 agent targets are "universal" and do succeed, so the command looks like it worked. The one target Orca reads is the only one
  skipped — so the user runs the documented command repeatedly, it reports success every time, and the skill never appears.

  Environment

  - Orca: v1.4.45 (Windows host) with WSL2 runtime
  - WSL: Ubuntu 24.04.2 LTS, kernel 6.6.87.1-microsoft-standard-WSL2
  - Node: v22.22.2 (nvm) · npm: 10.9.7
  - Skills: orchestration, computer-use

  Steps to reproduce

  npx skills add https://github.com/stablyai/orca --skill orchestration --global
  1. Run the command (the global flow used by docs / in-app installer).
  2. Observe ◇ Installation complete → ✓ Installed 1 skill, exit 0.
  3. Note the ■ Failed to install 1 → PromptScript: … footnote below the success block.
  4. Open Orca — the skill is not installed.

  Expected

  Either --global installs PromptScript too, or if it can't, the command fails loudly (non-zero exit, no "✓ Installed" banner) and tells the user to
  install project-level.

  Actual

  Exit 0 + ✓ Installed 1 skill on a run that failed the only target Orca uses; the PromptScript failure is an easy-to-miss footnote that contradicts
  the success message.

  Confirmed root cause

  PromptScript does not support global skill installation — so --global can never install that target, yet still reports overall success. Verified
  both modes against a clean dir:

  ┌────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                  Mode                  │                                              Outcome                                               │
  ├────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────┤
[...]

PR #11254, opened 2026-07-29
fix(skills): pass -y so the pasted skill install never stalls on a prompt

### LINKED ISSUE #9567: [Bug]: Settings "Install" reports success when the skill install command exits 1, and Re-check never re-runs it

brennanb2025:
Fixes #9567

## The bug

`buildAgentFeatureSkillInstallCommand` in `src/shared/agent-feature-install-commands.ts` produces the command that **every** Settings and onboarding skill card pastes into a terminal:

```
npx skills add https://github.com/stablyai/orca --skill <names> --global
```

That command is *pasted*, not auto-run — `OnboardingInlineCommandTerminal.tsx` dispatches `PASTE_TERMINAL_TEXT_EVENT` with `command.trim()` and no trailing newline, into a session whose cwd is `$HOME`.

It could stall on prompts from **two** separate programs — npm and the `skills` CLI — and either one strands the user. Closing the panel then tears the PTY down mid-prompt — which is exactly the reporter's `session-killed: immediate: true` / `exit 1`.

1. **npm's own prompt.** `npx skills …` without `--yes` blocks on `Need to install the following packages: skills@x.y.z / Ok to proceed?` whenever the package is missing from the npx cache (`libnpmexec/lib/index.js`, the TTY branch). This is not just a first-run cost: `missingFromTree` documents that for a bare name in the npx tree it "check[s] if a newer version is available", so the prompt returns every time upstream publishes.
2. **The `skills` CLI's own prompts.** Without `-y` the CLI blocks before it finishes, and it does so for **every** user — not only the multi-agent case in the report. The `Proceed with installation?` confirm is gated on `!options.yes` alone. The **"Installation method"** select fires even when exactly one agent is installed, because the CLI calls `ensureUniversalAgents` and appends its universal `.agents/skills` targets to the detected agent's own dir, making `uniqueDirs.size > 1`. That set is built from each agent's *relative* `skillsDir`, so the trigger is the universal-target expansion, not `--global`. Agent selection fires on top of that whenever more than one agent is installed.

   Verified directly against `skills@1.5.20` in a scrubbed, non-agent env with exactly one installed agent (Claude Code): without `-y` the run stops rendering at `◆ Installation method` and waits; adding `-y` alone takes the identical command through to `Installed 1 skill` / `Done!`. (The CLI sets `options.yes = true` by itself when it detects it is running *inside* an agent, which is why an agent-run probe never reproduces the stall.)

## The fix

Build the command with both non-interactive flags. One line covers macOS, Linux, native Windows, and WSL, because both platform rewrites in `CliSkillRuntimeSetup.tsx` wrap this same string rather than rebuilding it:

- the WSL base64 login-shell wrapper, and
- the native-Windows `update` → reinstall rewrite (`normalizeWindowsSkillUpdateCommand`), which calls `buildAgentFeatureSkillInstallCommand` and so now also yields a non-interactive command.

```diff
-npx skills add https://github.com/stablyai/orca --skill <names> --global
+npx --yes skills add https://github.com/stablyai/orca --skill <names> --global -y
```

The two flags are distinct and both load-bearing, which is already the established pattern in this repo: `src/main/skills/skill-update-run.ts` spawns `['--yes', 'skills', 'update', …, '--global', '-y']` with a docblock spelling out why. The pasted command now matches the headless one.

It matters most on **WSL**: `OrchestrationSetupCard.tsx` passes `freshnessSkillName={undefined}` when the runtime is `wsl`, so WSL users get no background updater at all and the terminal paste is their only path.

`skill-freshness-skipped-reason.ts` — the manual-recovery sentence shown when the background updater gives up and hands the user a reinstall command — derives from the same builder and picks the flags up for free. Covered by an updated assertion.

### `--global` is deliberately kept

A prior proposal (#6580) suggested dropping `--global`. That premise does not hold on current `main`: `src/main/skills/skill-discovery-sources.ts` deliberately scans ten provider home roots (`.codex/skills`, `.agents/skills`, `.claude/skills`, `.grok/skills`, `.config/opencode/skills`, `.pi/agent/skills`, `.gemini/skills`, `.gemini/antigravity/skills`, `.cursor/skills`, `.codex/plugins/cache`) *precisely because* `--global` writes there — the in-repo comment states that intent explicitly. Removing it would narrow discovery coverage.

### What `-y` auto-accepts, and why none of it regresses Orca

- **Install method (symlink vs copy).** `-y` skips the "Installation method" select, leaving the CLI's own recommended default: symlink. That is the effective default in practice, not an edge case, since more than one distinct skills dir is targeted even for a single agent. Orca models that shape as a first-class topology: `classifyHomeSkillTopology` returns `provider-alias` for a link into `~/.agents/skills`, and `topologyDedupeBucket` buckets `provider-alias` together with `canonical-copy` as `managed-global`. So a symlinked install is treated as managed, not as a modified or unmanaged copy.

  Confirmed on disk after a real `-y` install: `~/.agents/skills/orca-cli` is the real directory (`canonical-copy`) and `~/.claude/skills/orca-cli` is a symlink resolving into it (`provider-alias`). Both land in the same `managed-global` bucket, so freshness dedupes them to one managed install rather than flagging a stray copy.
- **Agent targeting.** With several agents detected, `-y` installs to all of them rather than a user-chosen subset. Those are exactly the provider home roots Orca already scans, and the alternative today is the prompt that hangs.
- **Prompts that simply stop appearing.** The "Proceed with installation?" confirm and the one-time "Install the find-skills skill?" offer are skipped; neither installs anything extra.

## Tests

The existing shared-module assertion was **vacuous** — it compared `ORCA_CLI_ORCHESTRATION_SKILL_INSTALL_COMMAND` against `buildAgentFeatureSkillInstallCommand([...])`, i.e. the function against itself, so no change to the command could ever fail it. Replaced with a literal-string assertion, plus a dedicated test pinning both flags and one covering the empty-list throw. Two assertions in `CliSkillRuntimeSetup.test.tsx` were tautological the same way (they rebuilt the expectation from the builder they were meant to pin) and are now literals — that pair is the only coverage that the flags survive the native-Windows rewrite.

Verified non-vacuous: reverting only the source change makes the shared-module tests fail; restoring it makes them pass.

Assertions updated to pin the new command:

- `src/shared/agent-feature-install-commands.test.ts`
- `src/renderer/src/components/onboarding/onboarding-feature-setup.test.ts`
- `src/renderer/src/components/settings/CliSkillRuntimeSetup.test.tsx` (×2)
- `src/renderer/src/components/settings/OrchestrationPane.test.tsx`
- `src/renderer/src/components/settings/linear-agent-skill-install-cta.test.tsx` (×2)
- `src/renderer/src/components/sidebar/LinearAgentSkillSetupPrompt.test.tsx` (the WSL wrapper surface)
- `src/renderer/src/components/skills/skill-freshness-skipped-reason.test.ts`
- `src/renderer/src/components/skills/SkillFreshnessUpdateDialog.test.tsx`

Several of those use `toContain`, so a superstring would have passed silently; they were tightened so each rendered surface actually proves the flags reach it. Sites left alone are pure local fixtures passed *into* a component as a prop (`AgentSkillSetupPanel.test.tsx`, `BrowserUseSkillStep.test.tsx`) and one invalid-payload rejection fixture (`onboarding-feature-setup-validator.test.ts`) — none derive from the builder.

The update command (`npx skills update <name> --global`) is intentionally unchanged: `normalizeWindowsSkillUpdateCommand` parses it with an anchored regex, and the headless updater already passes `--yes` itself.

Green: the 8 changed test files pass (92 tests). Full `typecheck` clean, `oxlint` and `oxfmt` clean, max-lines ratchet clean. Broader skills/settings/onboarding sweeps are green too, and CI on this branch is passing.

## Related issues

- **#6162** — partially addressed. This removes the interactive-prompt failure mode. The upstream DEP0190 spawn bug and the native-Windows PromptScript footnote are separate and remain open.
- **#4642** — related, not closed. Its residual is the PromptScript / `~/.agents/skills` target still not landing under `--global`, which is a separate decision.

## Validated in the real app

Run against a dev build of this branch, in an isolated instance (temp `HOME`) so every skill genuinely reports **Not installed** and the install surfaces actually render. CDP owner verified as this worktree's own Electron process before attaching.

**Settings → Computer Use → Install.** This is the reporter's exact path: the command is *pasted* into a live terminal and left sitting at the prompt ("Press Enter to run the command."), which is why an interactive prompt could strand the session. The pasted command now ends in `-y`:

```
npx --yes skills add https://github.com/stablyai/orca --skill computer-use --global -y
```

**Settings → Orchestration → "Prefer your own terminal? Copy install command".** The manual-copy dialog carries the flag too:

```
npx --yes skills add https://github.com/stablyai/orca --skill orchestration --global -y
```

(Both flags shown above come from this PR. `npx --yes` answers npm's *package-install* prompt; `-y` answers the *skills CLI's* agent-selection prompt. They are independent, and either one alone still leaves a way to stall.)

Not reachable in this environment, so not visually validated: the WSL and native-Windows rewrites (macOS host). Those are covered by unit tests, including a newly pinned assertion on the Windows update→reinstall rewrite.

### DIFF
--- a/src/renderer/src/components/onboarding/onboarding-feature-setup.test.ts
+++ b/src/renderer/src/components/onboarding/onboarding-feature-setup.test.ts
@@ -119,7 +119,7 @@ describe('onboarding feature setup runner', () => {
 
     expect(text).toBe(ALL_SKILL_INSTALL_COMMAND)
     expect(text).toBe(
-      'npx skills add https://github.com/stablyai/orca --skill orca-cli computer-use orchestration orca-linear --global'
+      'npx skills add https://github.com/stablyai/orca --skill orca-cli computer-use orchestration orca-linear --global -y'
     )
   })
 
--- a/src/renderer/src/components/onboarding/onboarding-feature-setup.test.ts
+++ b/src/renderer/src/components/onboarding/onboarding-feature-setup.test.ts
@@ -119,7 +119,7 @@ describe('onboarding feature setup runner', () => {
 
     expect(text).toBe(ALL_SKILL_INSTALL_COMMAND)
     expect(text).toBe(
-      'npx skills add https://github.com/stablyai/orca --skill orca-cli computer-use orchestration orca-linear --global -y'
+      'npx --yes skills add https://github.com/stablyai/orca --skill orca-cli computer-use orchestration orca-linear --global -y'
     )
   })
 
--- a/src/renderer/src/components/settings/CliSkillRuntimeSetup.test.tsx
+++ b/src/renderer/src/components/settings/CliSkillRuntimeSetup.test.tsx
@@ -118,7 +118,7 @@ describe('CliSkillRuntimeSetup runtime helpers', () => {
   )
 
   // Expectations are pinned rather than rebuilt from the builder: this is the only
-  // coverage that the non-interactive `-y` reaches the Windows rewrite.
+  // coverage that both non-interactive flags reach the Windows rewrite.
   it('reinstalls Windows-host skill updates through the add path', () => {
     expect(
       buildSkillCommandForRuntime(
@@ -129,13 +129,15 @@ describe('CliSkillRuntimeSetup runtime helpers', () => {
         },
         'win32'
       )
-    ).toBe('npx skills add https://github.com/stablyai/orca --skill orchestration --global -y')
+    ).toBe(
+      'npx --yes skills add https://github.com/stablyai/orca --skill orchestration --global -y'
+    )
   })
 
   it('treats missing runtime as a Windows host fallback for skill updates', () => {
     expect(
       buildSkillCommandForRuntime('npx skills update orca-cli --global', undefined, 'win32')
-    ).toBe('npx skills add https://github.com/stablyai/orca --skill orca-cli --global -y')
+    ).toBe('npx --yes skills add https://github.com/stablyai/orca --skill orca-cli --global -y')
   })
 
   it('keeps non-Windows host skill updates on the update path', () => {
--- a/src/renderer/src/components/settings/CliSkillRuntimeSetup.test.tsx
+++ b/src/renderer/src/components/settings/CliSkillRuntimeSetup.test.tsx
@@ -4,7 +4,6 @@ import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { describe, expect, it } from 'vitest'
 import { getDefaultSettings } from '../../../../shared/constants'
-import { buildAgentFeatureSkillInstallCommand } from '../../../../shared/agent-feature-install-commands'
 import { buildWslLoginShellCommand } from '../../../../shared/wsl-login-shell-command'
 import {
   buildSkillCommandForRuntime,
@@ -118,6 +117,8 @@ describe('CliSkillRuntimeSetup runtime helpers', () => {
     }
   )
 
+  // Expectations are pinned rather than rebuilt from the builder: this is the only
+  // coverage that the non-interactive `-y` reaches the Windows rewrite.
   it('reinstalls Windows-host skill updates through the add path', () => {
     expect(
       buildSkillCommandForRuntime(
@@ -128,13 +129,13 @@ describe('CliSkillRuntimeSetup runtime helpers', () => {
         },
         'win32'
       )
-    ).toBe(buildAgentFeatureSkillInstallCommand(['orchestration']))
+    ).toBe('npx skills add https://github.com/stablyai/orca --skill orchestration --global -y')
   })
 
   it('treats missing runtime as a Windows host fallback for skill updates', () => {
     expect(
       buildSkillCommandForRuntime('npx skills update orca-cli --global', undefined, 'win32')
-    ).toBe(buildAgentFeatureSkillInstallCommand(['orca-cli']))
+    ).toBe('npx skills add https://github.com/stablyai/orca --skill orca-cli --global -y')
   })
 
   it('keeps non-Windows host skill updates on the update path', () => {
--- a/src/renderer/src/components/settings/OrchestrationPane.test.tsx
+++ b/src/renderer/src/components/settings/OrchestrationPane.test.tsx
@@ -8,7 +8,7 @@ import { getOrchestrationUsageExamples } from '@/lib/orchestration-usage-example
 import { OrchestrationPane } from './OrchestrationPane'
 
 const INSTALL_COMMAND =
-  'npx skills add https://github.com/stablyai/orca --skill orchestration --global -y'
+  'npx --yes skills add https://github.com/stablyai/orca --skill orchestration --global -y'
 const UPDATE_COMMAND = INSTALL_COMMAND
 
 const mocks = vi.hoisted(() => ({
--- a/src/renderer/src/components/settings/OrchestrationPane.test.tsx
+++ b/src/renderer/src/components/settings/OrchestrationPane.test.tsx
@@ -8,7 +8,7 @@ import { getOrchestrationUsageExamples } from '@/lib/orchestration-usage-example
 import { OrchestrationPane } from './OrchestrationPane'
 
 const INSTALL_COMMAND =
-  'npx skills add https://github.com/stablyai/orca --skill orchestration --global'
+  'npx skills add https://github.com/stablyai/orca --skill orchestration --global -y'
 const UPDATE_COMMAND = INSTALL_COMMAND
 
 const mocks = vi.hoisted(() => ({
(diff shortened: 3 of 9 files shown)


brennanb2025:
Closing this — the premise is wrong, and the evidence for that was already on #9567 before I opened it.

`OrcaWin` said on that issue, the day it was filed:

> We actually want that step to stay interactive so you can choose which agents to install a skill for — so before we change anything, could you help us pin down where it broke for you?

They then asked the reporter two specific questions — did you press **Enter** to run the pre-filled command, and did the agent-selection list respond to arrow keys and space — and got no reply. So the interactive prompt is deliberate, and this PR would have suppressed it to fix something that was never actually established.

Reading the `skills` CLI (v1.5.20) confirms the prompts cover real user choices rather than noise:

- **Scope** (Project/Global) is already suppressed, because Orca passes `--global` explicitly. That is the one decision Orca legitimately owns.
- **Agent selection** and **Symlink vs Copy** are the user's. `-y` silently answers both.
- Worse, with **zero** detected agents, `-y` takes `targetAgents = validAgents` (`cli.mjs:3731`) and writes into every known agent home — reachable from onboarding by a first-run user who has not installed an agent yet, which is exactly where that Install card lives.

The CLI also already handles the case that matters for automation: when it detects it is running inside an agent it sets `options.yes` **and** scopes `options.agent` to the detected agent plus the universal target (`cli.mjs:3966`), so there is no fan-out on that path and no flag is needed.

The `npx --yes` half of this PR was defensible on its own — npm's "Ok to proceed?" offers nothing to configure and recurs whenever upstream publishes. But it is not worth carrying alone, because #10453's preflight gates on `^npx\s+skills\s+(?:add|update)\b`, which does **not** match `npx --yes skills add ...`. Landing it would have silently disabled the Windows npx-missing guidance for both install and update.

**#9567 is still a real bug — just not this one.** The undisputed part of that report is that Settings showed the install completing while the spawned command exited 1, and that **Re-check only re-reads the cached failed status instead of re-running**, so the indicator can never self-heal. That is worth fixing, and the two questions `OrcaWin` asked are still the fastest way to settle which of three things actually broke: the user never pressed Enter, the prompt was not navigable in the embedded terminal, or the panel tore down the session mid-prompt.

PR #11256, opened 2026-07-29
fix(editor): highlight Justfiles and Terraform variables

### LINKED ISSUE #11215: [Bug]: justfiles and .tfvars files have no syntax highlighting
### Operating system

macOS

### Orca version

v1.4.152

### Details

Two file types I have open all day render as flat uncolored text: `justfile` (the [just](https://just.systems) command runner) and Terraform `.tfvars` var files. Neither appears in the language-detect maps, so both fall through to `plaintext`:

```ts
// src/renderer/src/lib/language-detect.ts
'.tf': 'hcl',      // colored
'.hcl': 'hcl',     // colored
// no '.tfvars'  -> plaintext
...
Makefile: 'makefile',
// no justfile / Justfile / .justfile / .just  -> plaintext
```

`.tfvars` looks like the straightforward one. A tfvars file is HCL — same syntax as the `.tf` files sitting next to it in the same directory, one of which colors and one of which doesn't. `hcl` is already in use for `.tf`, so `'.tfvars': 'hcl'` seems like it would just work, unless there's a reason `.tf` and `.tfvars` are treated differently that I can't see from outside.

justfile is less obvious. `makefile` is the natural guess (justfile syntax is close enough that the common vim workaround is `# vim: set ft=make :`), but I'm not sure the makefile grammar is actually in the build. Poking at the packaged app on my machine:

```
$ cd /Applications/Orca.app/Contents/Resources
$ grep -a -o -E 'tokenPostfix:"\.[a-z0-9_-]+"' app.asar | sort -u | wc -l
      74
$ grep -a -o -E 'tokenPostfix:"\.(make|makefile|cmake|hcl|shell)"' app.asar | sort -u
tokenPostfix:".hcl"
tokenPostfix:".shell"
$ grep -c -a -o -F 'GNUmakefile' app.asar
0
$ grep -c -a -o -F 'id:"makefile"' app.asar
0
```

So `hcl` and `shell` turn up but makefile and cmake don't, and the only makefile grammar I could find anywhere in the bundle is highlight.js's (`.PHONY`, `HASH_COMMENT_MODE`), which I assume is for chat code fences rather than the editor.

Grepping a minified bundle is a blunt instrument and I could easily be missing a lazy chunk or a different registration path, so please treat that as a question rather than a finding. But if it is right, `Makefile: 'makefile'` (language-detect.ts:96) and `'CMakeLists.txt': 'cmake'` (:97) would be pointing at ids Monaco never registered, which would make plain `Makefile`s uncolored too — the shape of #9280. Someone with the source tree can settle that in a second by opening a Makefile.

That's why I'm not sure what the right target for justfile is. Two options I can see:

1. `shell`, which is definitely registered. Recipe bodies are shell, so the bash inside recipes and the `#` comments would color; recipe headers (`deploy: build`) would stay plain.
2. `makefile`, if the grammar is there (or gets added). Headers, `export` and `:=` color; recipe bodies stay plain.

For what it's worth, on a 450-line justfile of mine it's 189 comment lines and 7 shebang recipes of bash against 31 recipe headers, so `shell` would cover more of it — but that's one file, and you'd know better whether adding a grammar is cheap here. Neither one understands just's `{{var}}` interpolation or `[group('x')]` attributes, so those stay uncolored either way, which I'm fine with.

**Repro:** create a file named `justfile` containing a `#` comment, `x := "y"`, and a `build:` recipe with an indented command, and open it in the editor. Same for any `foo.tfvars` next to a `foo.tf`.

One note if either change happens: `mobile/src/session/mobile-file-language.ts` carries what looks like a duplicate of both maps.

bbingz:
## Summary

`justfile` and `*.tfvars` currently render as plaintext in the file preview on desktop and mobile. This maps them to shell and HCL.

- Detect `.tfvars` files as HCL in desktop and mobile file previews.
- Detect `justfile`, `Justfile`, `.justfile`, and `.just` as shell syntax on both clients.
- Keep mobile highlighting effective by resolving HCL to the bundled INI grammar and shell to Bash.

Closes #11215

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — under the required Node 24 runtime, 39,603 tests passed and 64 skipped; one SSH system-transport integration test failed only in the full concurrent suite and passed 3/3 when rerun alone (pre-existing and unrelated: this diff touches only language-detection lookup tables, no transport code)
- [x] `pnpm build` — desktop, web, relay, CLI, and both macOS Swift architecture builds completed; the final universal-binary step failed because `lipo` could not find the architecture outputs at its expected paths (pre-existing: that step lives in `config/scripts/build-computer-macos.mjs`, which is unchanged on `main` and untouched by this PR)
- [x] `cd mobile && ./node_modules/.bin/vitest run --config vitest.config.ts` — 352 files passed; 2,597 tests passed and 2 skipped
- [x] `cd mobile && ./node_modules/.bin/tsc --noEmit`
- [x] `cd mobile && ./node_modules/.bin/oxlint src app`
- [x] Added desktop and mobile regression tests for Unix and Windows paths, canonical Justfile names, case-insensitive `.tfvars`, grammar resolution, and semantic token output

## AI Review Report

The review traced both runtime paths before choosing language IDs. Monaco 0.55.1 bundles HCL and shell but not Makefile; mobile lowlight bundles INI, Makefile, and Bash but not HCL. The implementation therefore uses native HCL on desktop, shell for Justfiles on both clients, and the existing INI grammar as the mobile HCL fallback.

Cross-platform review covered macOS, Linux, and Windows path separators and extension casing. The change does not touch shortcuts, labels, shell execution, Electron platform behavior, SSH transport, git providers, agents, or integrations. Lookups remain constant-time map accesses, and there is no meaningful performance risk or UI layout change.

CodeRabbit and Greptile both reviewed the branch and raised no actionable findings.

## Security Audit

The change adds no dependency, command execution, filesystem access, IPC, authentication, secret handling, or network surface. File paths are used only for in-memory filename and extension matching; no path is opened or executed by the new logic.

## Notes

- Mobile HCL highlighting is an intentional approximation using the bundled INI grammar; it colors comments, assignments, and strings without adding a grammar dependency.
- Shell is used for Justfiles because it is available on both clients and covers comments and recipe bodies; desktop Monaco does not ship a Makefile language.
- X handle: not provided.

### DIFF
--- a/mobile/src/session/mobile-file-language.ts
+++ b/mobile/src/session/mobile-file-language.ts
@@ -60,12 +60,17 @@ const EXT_TO_LANGUAGE: Record<string, string> = {
   '.lua': 'lua',
   '.r': 'r',
   '.R': 'r',
-  '.make': 'makefile'
+  '.make': 'makefile',
+  '.tfvars': 'hcl'
 }
 
 const FILENAME_TO_LANGUAGE: Record<string, string> = {
   Dockerfile: 'dockerfile',
   Makefile: 'makefile',
+  justfile: 'shell',
+  Justfile: 'shell',
+  '.justfile': 'shell',
+  '.just': 'shell',
   'CMakeLists.txt': 'cmake',
   '.gitignore': 'ini',
   '.gitattributes': 'ini',
--- a/mobile/src/session/mobile-file-syntax.test.ts
+++ b/mobile/src/session/mobile-file-syntax.test.ts
@@ -20,6 +20,39 @@ describe('mobile file syntax highlighting', () => {
     expect(resolveMobileSyntaxLanguage('Dockerfile')).toBe('plaintext')
   })
 
+  it('detects Terraform variable files and canonical Justfile names', () => {
+    expect(detectMobileFileLanguage('environments/production.tfvars')).toBe('hcl')
+    expect(detectMobileFileLanguage('C:\\repo\\terraform\\LOCAL.TFVARS')).toBe('hcl')
+    expect(detectMobileFileLanguage('justfile')).toBe('shell')
+    expect(detectMobileFileLanguage('C:\\repo\\Justfile')).toBe('shell')
+    expect(detectMobileFileLanguage('/home/user/project/.justfile')).toBe('shell')
+    expect(detectMobileFileLanguage('/home/user/project/.just')).toBe('shell')
+  })
+
+  it('uses available mobile grammars for Terraform variables and Justfiles', () => {
+    const tfvarsLanguage = resolveMobileSyntaxLanguage('environments/production.tfvars')
+    const justfileLanguage = resolveMobileSyntaxLanguage('Justfile')
+
+    expect(tfvarsLanguage).toBe('ini')
+    expect(justfileLanguage).toBe('bash')
+    expect(
+      highlightMobileCode('# environment\nregion = "us-east-1"', tfvarsLanguage)
+    ).toMatchObject({
+      highlighted: true,
+      segments: expect.arrayContaining([
+        { text: '# environment', kind: 'comment' },
+        { text: '"us-east-1"', kind: 'string' }
+      ])
+    })
+    expect(highlightMobileCode('# build\nbuild:\n  echo "ready"', justfileLanguage)).toMatchObject({
+      highlighted: true,
+      segments: expect.arrayContaining([
+        { text: '# build', kind: 'comment' },
+        { text: '"ready"', kind: 'string' }
+      ])
+    })
+  })
+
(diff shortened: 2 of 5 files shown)


bbingz:
Rebased onto current main last night (head 726be2aaa2, no conflicts). The Actions suites are sitting at `action_required` — same shape and file as #11294 and #11322, which both merged after a quick workflow approval. One "Approve and run workflows" click would give a full green rollup here too.

bbingz:

PR #11261, opened 2026-07-29
feat(bench): interleaved A/B startup benchmark with drift control

TobiasVeiga00:
Refs #9300 — this covers Goal 1 ("measure the main Windows/WSL latency sources with reproducible scenarios") and gives Goal 3 ("ship incremental improvements with before/after numbers") a number that can actually be defended. Not a fix for the umbrella issue, so it does not close it.

## Summary

- Adds `tools/benchmarks/startup-ab-bench.mjs`: runs two builds **interleaved** and reports a paired shift per startup phase with a confidence interval, instead of comparing the medians of two separate runs.
- Adds a **drift control**: phases the change under test cannot affect are measured too. If one of them moves — or if none of them produced data — the run is marked unproven and its phase verdicts must not be quoted.
- Extends the main-thread churn probe to timestamp stall spans on the same clock as the startup milestones, so a stall can be attributed to a concrete init step.

No user-visible behavior change. This is measurement tooling plus diagnostics that stay gated behind `ORCA_MAIN_THREAD_DIAGNOSTICS=1`.

### Why interleaving

Running five baseline launches and then five candidate launches charges the machine's drift — thermal throttling, background indexers, page-cache warmth — to whichever build ran second. That is not a small correction. In the run that motivated this work, phases that renderer chunking **cannot** touch (`spawnToAppReady`, `appReadyToServices`) moved ~25% between the two blocks, so roughly half of every reported delta was the machine warming up.

`startup-ab-bench.mjs` runs the arms as ABBA blocks — pair 0 is baseline-then-candidate, pair 1 is the reverse — so a drift that grows steadily across the run enters consecutive pairs with opposite sign and cancels out of the paired deltas. The cancellation is exact only when both arms share the same mean launch slot, which is why `--pairs` must be even and the CLI rejects an odd value.

Phases are summarised with a Hodges-Lehmann shift rather than a mean (one multi-second stall is common on Windows and would otherwise dominate) and a seeded percentile-bootstrap interval, so the reported interval is reproducible from the result JSON. A phase only gets an `improved` / `regressed` verdict when the whole interval sits on one side of zero.

### The measurement status is the thing to read first

Every result carries a run-level `measurementStatus`, and only one of its three values licenses quoting a number:

| `measurementStatus` | Meaning |
|---|---|
| `attributable` | At least one control phase was measured and none moved. The phase verdicts may be quoted. |
| `control-phase-drift` | A control phase showed a direction. The run measured the machine, not the change. |
| `controls-unmeasured` | No control phase produced data. Drift was not found absent — it was never measured. |

The third value exists because the two failures are not the same thing and the first version of this code conflated them. A control phase that reports `no-data` is not agreement that the machine was quiet; it is a control that never fired. Both the CLI and the CI job now withhold the phase table unless the status is `attributable`, so an unproven verdict cannot be quoted out of a log.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — **pre-existing failure, not from this change.** It stops at `verify:skill-bundle-manifest` with "Generated skill artifacts are stale" for `resources/skills/{current-manifest,snapshot-registry,release-mapping}.json`. I reproduced the identical failure on a pristine detached checkout of `upstream/main` (`21dee21a`) with no changes applied, so it is an environment or repo-state issue rather than something this branch introduced. This branch touches nothing under `resources/skills/`. `oxlint` itself passes on every file added here.
- [x] `pnpm test` — full suite run on a Windows host: **38270 passed, 158 failed across 64 files**. The failures are pre-existing and environmental, not from this branch: they are tests that spawn a POSIX shell (`Error: spawn /bin/sh ENOENT`, `path: '/bin/sh'`) on a machine that has none. I verified this by running one of them — `src/main/ipc/ephemeral-vm.test.ts` — on a pristine detached checkout of `upstream/main` (`21dee21a`), where it fails identically (8 failed / 4 passed, same error). **Every test file covering source this branch touches passes:** `main-thread-churn-probe.test.ts`, `AgentSkillSetupPanel.test.tsx`, `OnboardingInlineCommandTerminal.test.ts`, plus this PR's three benchmark test files.

On tests: the statistics and the counterbalancing property are covered directly rather than smoke-tested. `interleaved-arm-schedule.test.mjs` applies a synthetic linear drift to a fixed pair of arm costs and asserts that the ABBA schedule recovers the true per-arm difference **exactly**, while the blocked design reports it wrong by the full drift — if someone "simplifies" the interleaving away, that test fails and explains why. `paired-shift-statistics.test.mjs` pins the estimators (including that Hodges-Lehmann stays negative on a sample whose mean an outlier flips positive), that a straddling interval refuses a verdict, and that an unmeasured control phase is neither drift nor agreement. `startup-bench-arguments.test.mjs` pins the flag contract, including that `--iterations` cannot leak back into the A/B bench.

Beyond the unit tests, both CLI paths were exercised end to end on Windows: the single-arm bench against a real launch (its CLI and output are unchanged by the refactor), and the A/B bench twice with both arms pointed at the same build — once to confirm the drift control fires when the machine is noisy, and once with a deliberately unreachable `--control-phases` to confirm the `controls-unmeasured` path reports rather than silently passing. That second run is the regression this design exists to prevent: identical arms, a `regressed` verdict on the table, and the old code calling it attributable.

**The CI workflow has been executed on a real runner three times** on a public fork, and the runs are worth reading side by side. The first is why two of the fixes below exist. The second and third are verification, and between them they demonstrate the warmup fix empirically.

Run 2 — `--warmup 0`, `measurementStatus: attributable`:

```
[ab] pair 1/2 baseline…   ok total=6.11s maxStall=4.11s   <- cold, and it lands in one arm
[ab] pair 1/2 candidate…  ok total=1.45s
[ab] pair 2/2 candidate…  ok total=1.46s
[ab] pair 2/2 baseline…   ok total=1.42s
```

Run 3 — same job, `--warmup 1`, so a round warms **both** arms:

```
[ab] warmup 1/1 baseline…   ok (discarded)
[ab] warmup 1/1 candidate…  ok (discarded)
[ab] pair 1/2 baseline…   ok total=1.41s
[ab] pair 1/2 candidate…  ok total=1.42s
[ab] pair 2/2 candidate…  ok total=1.43s
[ab] pair 2/2 baseline…   ok total=1.43s
```

The 6.11s cold launch is gone, and with it a 4.7s advantage that run 2 handed to whichever arm did not pay it. That is the bias the per-arm warmup exists to remove, measured rather than argued.

Run 3 also exercises the other branch: it ended `control-phase-drift` and refused to attribute anything. At two pairs the control phases are milliseconds wide, so a tight interval clears zero easily — correct behaviour for an n this small, and the reason `--pairs 12` is the documented floor. Both runs: 4/4 launches `ok`, events omitted, artifact uploaded, ~1m52s for both arms from separate worktrees on `ubuntu-latest`.

These are mechanics checks, not measurements. No number from them appears anywhere in this stack.

## Changes

| File | Change |
|------|--------|
| `tools/benchmarks/startup-ab-bench.mjs` | New: interleaved A/B CLI with drift control, measurement status, and verdict reporting |
| `tools/benchmarks/paired-shift-statistics.mjs` | New: median, Hodges-Lehmann shift, seeded bootstrap interval, drift detection |
| `tools/benchmarks/interleaved-arm-schedule.mjs` | New: ABBA schedule and the counterbalancing invariant |
| `tools/benchmarks/startup-profile-fixture.mjs` | Extracted from `startup-time-bench.mjs`: synthetic profile, gh shim, launch env |
| `tools/benchmarks/startup-launch-probe.mjs` | Extracted: one instrumented launch, milestone parsing, phase derivation |
| `tools/benchmarks/startup-bench-arguments.mjs` | Extracted: shared flag parsing |
| `tools/benchmarks/benchmark-result-formatting.mjs` | Extracted: console formatting and result persistence |
| `tools/benchmarks/startup-time-bench.mjs` | Now a thin CLI over the shared modules; CLI and output unchanged |
| `src/main/diagnostics/main-thread-churn-probe.ts` | Stall spans timestamped on the milestone clock; init marker helper |
| `src/main/index.ts` | Init markers between store-loaded and services-initialized |
| `.github/workflows/startup-perf-ab.yml` | New: dispatch-only A/B job, modelled on `terminal-perf.yml` |
| `docs/reference/startup-benchmark-methodology.md` | New: how to build the two arms and how to read a verdict |
| `AGENTS.md` | Rule: don't quote a before/after startup number from two single-arm runs |

Plus three test files: `interleaved-arm-schedule.test.mjs`, `paired-shift-statistics.test.mjs`, `startup-bench-arguments.test.mjs`.

The extraction exists so both benchmarks derive phases from **identical** code. Two drifting copies would produce numbers that cannot be compared against each other, which defeats the purpose.

## AI Review Report

Reviewed with Claude Code. Risks checked and what came of them:

**Cross-platform (macOS / Linux / Windows).** The launch probe keeps the existing `require('electron')` resolution rather than hardcoding `electron.exe`, so it runs on all three. Process teardown branches on `win32` (`taskkill /T /F`) versus POSIX; the fake `gh` shim writes a `.cmd` on Windows and a `chmod +x` shell script elsewhere. All paths go through `path.join` / `resolve`. No keyboard shortcuts, menu accelerators, or UI labels are touched, so the `metaKey` / `CmdOrCtrl` rules do not apply here.

**Verified on Linux, not on macOS.** The harness has now been executed on Windows (i5-8350U) and on `ubuntu-latest` via the CI workflow. macOS remains preserved-by-construction from the existing benchmark plus review, not verified by a run — the POSIX teardown path is shared with Linux and is exercised there, but the `require('electron')` resolution and the shell `gh` shim have not been run on a Mac. Flagged rather than glossed over.

**SSH / remote / folder workspaces.** No change. The benchmark launches a local built app against a synthetic `userData` fixture; it does not touch git execution, remote providers, or worktree code. The fixture seeds plain `git init` repos and does not assume a worktree.

**Agent and integration compatibility.** No agent, provider, or integration code is touched. The fixture's fake `gh` reproduces the hung-login-probe scenario and lives entirely inside the benchmark's temp directory, prepended to `PATH` only for the launched child.

**Performance risk in shipped code.** The only `src/` change is the diagnostics probe. Both the stall-span recording and the init markers are gated on `ORCA_MAIN_THREAD_DIAGNOSTICS=1` and are no-ops otherwise, so production startup is unaffected. Verified the marker helper early-returns before doing any work when the flag is unset.

**UI quality.** Not applicable — no UI surface in this PR.

**Security.** Covered below.

### What changed as a result of review

First round, before the PR was opened: results omit the raw milestone events unless `--keep-events` is passed (at 48 launches the event arrays were ~90% of the file, and a verdict is recomputed from the phases alone); the bootstrap PRNG was seeded so a published interval is reproducible from its JSON; and the CLI rejects an odd `--pairs` instead of silently producing a schedule where drift does not cancel.

Second round, from the bot reviews on this PR. Every finding below was a real defect in a benchmark whose entire job is not to lie about numbers:

- **`detectDrift` conflated two opposite failures.** `verdict !== 'inconclusive'` also matched `'no-data'`, so a control phase neither arm reached — `aclGrantMs` off win32, say — flipped the run to "drifted" and marked every real verdict unproven. In the other direction, a control phase absent from the summaries was dropped silently, so naming only phases the profile never reaches produced an empty control set that printed "no directional movement — attributable to the change". Reproduced both, then fixed: only directional verdicts count as drift, unmeasured controls are reported separately, and the run-level `measurementStatus` above is what callers gate on.
- **Warmup warmed only the baseline arm.** Every discarded launch ran baseline, handing it a page-cache advantage at the first measured pair — a bias in one arm's favour that the ABBA schedule cannot cancel. `--warmup N` is now N rounds and a round warms both arms, alternating order.
- **The CI job shared one dependency tree between the arms.** Both were built into `out/` trees under one checkout resolving `node_modules` upward, so whichever arm installed last decided what both loaded at runtime, and a lockfile or externalized-module difference between the refs would have been measured as if it were the change under test. Each arm is now its own `git worktree` with its own `pnpm install`, outside the checkout. The job timeout went 60 → 90 minutes to cover the second install.
- **POSIX teardown killed only the direct child.** `proc.kill('SIGKILL')` left Electron's helpers and the gh shim's `sleep` grandchild alive on macOS and Linux. That grandchild outliving its parent is deliberate — it is how the fixture reproduces a hung `gh` — but across a 48-launch run the survivors accumulate and add exactly the machine noise the interleaving exists to cancel. Windows already got tree semantics from `taskkill /T`; POSIX now spawns detached and signals the process group.
- **A trailing stderr line without its newline was dropped** with the buffer, turning a completed launch into a `null` phase that reads as missing data.
- **`--iterations` parsed in the A/B bench and did nothing** — it rode in on the shared flag spec while that bench counts `--pairs`. The spec is now split.
- **Numeric flags accepted anything `Number()` would coerce.** `--pairs ''` became 0 and `--timeout-ms -1` became an instant timeout, silently. Blank and negative values now throw; zero still passes, since several profile flags use it to opt out.
- **`--keep-events` was described here but implemented one PR downstream.** Correct call — the description promised behavior this branch did not have. Moved to this branch, where the claim lives.
- **The single-arm bench built the fixture before checking the build existed**, spending tens of seconds of disk I/O at `--files 28000` only to throw.
- **Checkout persisted credentials** in a job that runs build scripts from a caller-chosen ref. Now `persist-credentials: false` with `permissions: contents: read`.

Third round, on the fixes themselves. Both findings were defects the second round introduced, which is the argument for re-reviewing a fix rather than assuming it landed clean:

- **`${{ runner.temp }}` in job-level `env` would have made the workflow invalid.** Job-level `env` is evaluated before the job is routed to a runner, so the `runner` context does not exist there — this would have failed at parse time rather than merely resolving to something wrong. `ARMS_DIR` is now written to `$GITHUB_ENV` from `$RUNNER_TEMP` in a first step.
- **`--iterations 0` launched nothing and still wrote a result file**, one whose medians are all `null` and which reads as a completed benchmark; `--iterations 1.5` ran twice. Fixed at the parser rather than at one CLI: numeric flags now carry optional `integer` and `min` constraints, applied to every count (`--iterations` min 1, `--pairs`, `--warmup`, `--files`, `--session-tabs`, `--github-repos`) and to `--timeout-ms`. Durations stay fractional on purpose.

Self-reviewing that round also turned up one thing neither bot flagged: **spawning detached took the launch out of the terminal's process group**, so Ctrl-C during a long run reached the harness and left Electron running — reintroducing, at the interactive boundary, the exact orphan accumulation the detached spawn was added to prevent. `SIGINT`/`SIGTERM` now reap the in-flight launch first.

Fourth round, from actually dispatching the workflow. Neither of these was reachable by reading the diff:

- **Every launch died on the runner.** A raw `spawn()` of Electron on Ubuntu cannot use the setuid `chrome-sandbox` shipped in `node_modules` and aborts with SIGTRAP before the first milestone, so all four launches reported `early-exit` in nine seconds. The probe now passes the same switches `tests/e2e/helpers/electron-launch-args.ts` and `config/scripts/run-idle-cpu-benchmark.mjs` already use, gated to Linux. Both arms receive identical switches, so the paired comparison is unaffected; a Linux run is a software-rendering configuration and its *absolute* phase timings are not comparable to a desktop run.
- **That broken job reported green.** A run in which nothing launched was surfaced as merely lacking a verdict, because the design deliberately treats an unusable run as a warning rather than a failure. That was the right call for a *drifting* machine and the wrong one for a benchmark that never ran — the distinction this PR is otherwise built around, missed in its own CI contract. Both CLIs now exit non-zero when zero launches reach the awaited milestone, after writing the result file so it still uploads for triage. Partial failures stay tolerated and are reported as a `launchOutcomes` breakdown, and the result JSON carries `completedLaunches`.

Worth recording in the workflow's favour: in that broken run the `controls-unmeasured` detection from the second round did its job — it refused to print verdicts over zero data and said drift had never been measured. Without it, the job would have rendered a phase table computed from nothing.

Fifth round, once #11262 left draft and CodeRabbit reviewed the combined diff. The first of these is the most serious defect found at any point in this stack:

- **The probe relied on stderr's `end` firing before the child's `exit`.** Different emitters; registration order guarantees nothing. When `exit` won, `finish()` locked `early-exit` and the later flush could no longer upgrade a trailing milestone — and combined with the zero-completion guard from the fourth round, that turned a launch which *actually completed* into a hard failure. Both handlers now flush before the outcome is decided, so the result is order-independent rather than dependent on emitter timing.
- **Every stall span the churn probe reported was 25ms too wide.** `gapMs` subtracts one nominal tick but `fromMs` was anchored at the previous callback, so the span started while the main thread was still responsive — which is precisely how a stall gets misattributed to the wrong startup milestone, the one thing these spans exist to do correctly. The span had no test, which is why it survived four rounds; there is now one that drives the probe with a scripted clock and fails by exactly 25ms without the fix.
- **A fixture rebuilt in place kept the previous one's files**, so the manifest could understate the file count that drives the win32 ACL walk. Only an explicit `--fixture-dir` reaches this — the default path encodes the profile shape, so a different shape lands in a different directory — but the flag is documented and the failure is silent.
- A test comment claimed the blocked design reports ±0ms where the assertion says −20ms.
- The committed result files pre-date `measurementStatus`, which `AGENTS.md` now requires and which explicitly rejects `driftDetected` as a substitute — leaving the repo contradicting itself. The methodology doc now explains how to reconstruct the status from such a file by hand, and says regenerating is preferable when the numbers still matter.

One flagged item was not actioned: CodeRabbit's docstring-coverage pre-merge check (36% against an 80% threshold). That is a bot policy threshold over the whole diff rather than a defect, and the modules here carry JSDoc on their exported surface explaining *why* each estimator was chosen.

## Security Audit

- **Command execution.** The benchmark spawns the Electron binary and `git` for fixture setup, both with argument arrays — no shell interpolation, no user-controlled strings reaching a shell. Process teardown uses `taskkill` with a numeric PID on Windows and a process-group signal on POSIX.
- **Path handling.** All fixture paths are derived under `os.tmpdir()` (or an explicit `--fixture-dir`) and joined with `path.join`. Result paths are written under `tools/benchmarks/results/`.
- **PATH manipulation.** The fake `gh` shim directory is prepended to `PATH` only in the environment handed to the spawned child, never to the harness process, and only when `--gh-hang-ms` is passed.
- **Secrets and data exposure.** Result JSON is committed as evidence, so home-anchored paths are rewritten to `~` before writing. The launch environment points `HOME`/`USERPROFILE` and `GIT_CONFIG_GLOBAL` at the fixture so the developer's real git identity and config never enter a run or a result.
- **IPC / auth / dependencies.** No IPC surface, no auth paths, no new dependencies — the statistics are implemented in-repo rather than pulling a stats library.
- **CI workflow.** Dispatch-only with no secrets, no `pull_request_target`, least-privilege `contents: read`, and checkout credentials are not persisted. The checked-out refs are supplied by the maintainer running it.

**Follow-up:** none outstanding. The workflow has been dispatched and its result artifact inspected; the two defects that first run exposed are fixed and verified by a second run.

## Notes

- **Platform-specific:** the *measurements* quoted anywhere in this stack are from Windows. The default fixture reproduces the documented pathological win32 case (tens of thousands of `userData` files driving the ACL walk). Linux runs pass `--no-sandbox` plus software-rendering switches to start at all, so their absolute timings are their own baseline — fine for an A/B, not comparable across platforms.
- **The CI default of `--files 2000` is lower than the local default of 28000 on purpose.** The file count drives the win32 `icacls` walk, which does not exist on the Linux runner — 28000 files there would buy I/O and nothing else.
- **The CI job is dispatch-only on purpose.** An A/B compares two refs somebody deliberately chose; there is no standalone "startup today" number worth trending, so a nightly cron would have nothing to measure against. This differs from `terminal-perf.yml` deliberately.
- **An unusable run is a warning, not a failure.** A drifting or uncontrolled run does not mean the change is bad, it means the run cannot answer the question. Failing the job would push people to re-run until it passes, which is the opposite of the point — so instead the job withholds the phase table and says why.
- A follow-up PR uses this harness on a real change and is chained onto this branch.

X handle: @tobiasdavid2000


TobiasVeiga00:
Self-reported defect found after opening this PR — flagging before anyone spends review time on it.

**The harness can hang after a run completes.** One of my benchmark runs wrote its results file and printed its report, but the node process never exited. Cause: an orphaned Electron child survived `killProcessTree` and kept the parent's event loop alive. Killing that orphan by hand made the harness exit immediately, which confirms it.

Impact: on the CI job in this PR that would hang until the 60-minute `timeout-minutes`, after the measurement had already succeeded. Locally it leaves a stray Electron process per affected run.

Not yet root-caused with confidence. My leading hypothesis is the early return in `killProcessTree`:

\\\js
if (proc.exitCode !== null || proc.signalCode !== null) return
\\\

When the top-level Electron process has already exited, this returns before running `taskkill /T`, so surviving grandchildren are never reaped. A grandchild spawned after `taskkill` enumerated the tree would also escape. On Windows the robust fix is a Job Object rather than PID-tree killing; a cheaper mitigation is to reap unconditionally and then verify no descendant remains.

I did not push a fix because verifying it needs another multi-launch run and I would rather not ship an unverified change to process teardown. I will follow up with a fix plus a regression check. Happy to hold review until then if you prefer.

PR #11273, opened 2026-07-29
fix(terminal): stop reading a modifier-held IME keydown as Enter, and defer every IME-owned key

kunsanglee:
## Summary

`main` classifies IME keydowns as Enter incorrectly in both directions, and this PR fixes both. Three commits, separable.

### 1. A modifier-held IME keydown is read as the committing Enter

On Windows, typing ㅃ, ㄲ, ㄸ, ㅆ, ㅒ or ㅖ mid-composition sends a Shift+Enter newline to the shell. Ordinary Korean typing injects `\x1b\r`, which at a shell prompt submits whatever line is pending.

This is not Korean-specific. Nothing on the path reads a character — `TERMINAL_IME_OWNED_KEYS` holds physical key names, `getTerminalImeModifiedEnterKind` reads modifiers alone, and the predicate never looks at a character value. The condition is any keydown Windows consumes for an IME while a composition is open with exactly one of Shift or Ctrl held. Resizing the conversion segment with Shift+Left/Right is a standard Japanese IME operation and produces the identical event, so it lands on the identical rewrite — and unlike a shifted jamo it is pressed repeatedly on purpose.

`isTerminalImeProcessEnter` recognizes the Windows committing Enter by `key === 'Process'`, `keyCode === 229`, and exactly one of Shift or Ctrl. It never reads `code`. But Windows reports **every** IME-consumed keydown as `Process`/229, so a jamo typed with Shift held satisfies all three conditions. `keyboard-handlers.ts` then rewrites that keydown into `{ key: 'Enter', ... }`, `terminal-shortcut-policy.ts` matches its Shift+Enter rule on `event.key === 'Enter'`, and the byte is deferred and written to the PTY behind the committed syllable.

`code` is the field that separates the two shapes, and this repo already records the jamo side: `tests/e2e/terminal-ime-observed-event-sequences.ts:44` replays an observed IBus Hangul jamo keydown as `keydown('Process', 'KeyG', 229, true)`, physical code preserved.

It does **not** always record `Enter` on the other side. #11293's own chord-owner test — `owns one Windows Process sequence across changing timestamps and blank codes` — sends the Windows Process sequence with `code: ''` and only puts `code: 'Enter'` on the re-dispatch. So the predicate refuses a known non-Enter physical code and keeps its previous verdict whenever the code is absent or `'Unidentified'`. A shifted jamo is rejected; a blank-code committing Enter is not. The predicate cannot lose a keystroke `main` delivers today.

### 2. Only Enter is deferred, but every IME-owned key doubles

This is what the PR was originally opened for, rewritten on top of `main`.

#11293 landed the cross-platform IME composition lifecycle work, and its e2e spec `tests/e2e/korean-ime-terminal-shift-enter-commit.spec.ts` reproduces the same macOS double dispatch this branch first measured: the committing Enter as `keyCode 229 / isComposing=true`, then the re-dispatch as `keyCode 13 / isComposing=false` about 2 ms later. Its deferral is stricter than what this branch originally had — `sendTerminalInputAfterComposition` waits for the terminal element's real `compositionend` and for xterm's composition-session-end event, with a 200 ms fallback, where this branch scheduled a bare `setTimeout(0)`.

So this PR no longer adds a second guard. `terminal-ime-shortcut-guard.ts` and its two tests are deleted rather than merged alongside `main`'s mechanism. Keeping both would have been a regression: the guard sat at the top of `onKeyDown` and returned early on the committing press, so `deferredNewlineSender.defer` never ran and the weaker ordering took its place.

What remains is the one thing `main` does not cover. `main` defers Enter alone, but the committing press repeats and the direct shortcut write races xterm's flush for every key the IME consumes mid-composition. `Ctrl+Backspace` (`\x17`), `Cmd+Backspace` (`\x15`), `Cmd+Delete` (`\x0b`) and the Alt/Ctrl+Arrow word motions double and land ahead of the syllable they commit, exactly as Enter did. This widens `main`'s two Enter-shaped conditions to `TERMINAL_IME_OWNED_KEYS`, the set `xterm-bypass-policy.ts` already defines, and exports that set for reuse instead of restating it.

The Windows Enter-chord owner stays Enter-only on purpose. `getModifiedEnterChord` keys off the held modifier alone and never checks the key, so claiming its single slot on a non-Enter key would take ownership away from the real chord. Removing that one narrowing makes a `Ctrl+Enter` behind a deferred `Ctrl+Backspace` disappear — measured, and now pinned by a test.

The two halves meet in one place. With the narrowed predicate a shifted jamo no longer sets `imeProcessEnter`, so the widened deferral condition falls through to `TERMINAL_IME_OWNED_KEYS.has('Process')`, which is false, and the keydown takes the ordinary path.

Net effect: the original 329 lines across 5 files became 208 across 5, with no new module.

## Screenshots

No visual change. No layout, token, component, or styling change, so `docs/STYLEGUIDE.md` does not apply. The user-visible effect is the order and count of bytes reaching the shell.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [ ] `pnpm build` — not run, see below

`pnpm` is not installed on this machine, so the gates ran through the underlying binaries in `node_modules/.bin` rather than through the `pnpm run` wrappers.

- **lint** — `oxlint` exits 0 repo-wide. `oxlint --config config/oxlint-code-quality-native-plugins.json src/renderer/src/components/terminal-pane --deny-warnings` and the same directory under `--type-aware --config config/oxlint-code-quality-type-aware.json --deny-warnings` both exit 0. `check:max-lines-ratchet` reports no new bypasses (no `max-lines` disable was added; the one at the top of `keyboard-handlers.ts` predates this branch). `check:reliability-gates` passes for 55 gates. `check:changed-code-quality` shells out to `pnpm exec oxlint` and dies on `spawnSync pnpm ENOENT`, so its scans were run directly as listed.
- **typecheck** — 0 errors across `tsconfig.node.json`, `tsconfig.tc.cli.json`, and `tsconfig.tc.web.json`.
- **test** — `src/renderer/src/components/terminal-pane` is 214 files / 2,840 tests; 210 files pass. The four that fail — `terminal-ime-xterm-adversarial`, `terminal-ime-xterm-transaction-events`, `terminal-ime-xterm-composition-deduplication`, `terminal-ime-won-composition-order`, 59 tests — assert against the **installed** patched `@xterm/xterm`, and #11293 rewrote `config/patches/@xterm__xterm@6.1.0-beta.287.patch`. The repo's patch expects post-image blob `a85075dcbea89c342500838a66b40f2ff6c21c0f`; the installed `lib/xterm.js` hashes to `eecf435886d866a838430b9f81f713deed3eb8c1`. A clean `main` checkout sharing the same `node_modules` reproduces the identical 4 files / 59 failures, so this is a stale local install that a fresh `pnpm install` clears, not a regression. None of those four files import anything this PR touches.
- **build** — not run. Needs `pnpm run build:*`, unavailable here for the reason above. CI covers it.
- **New tests.** The two changed test files go from 37 to 50 tests. Every added case was verified to fail with the guard it covers removed, and to fail for its stated reason.
  - `keyboard-handlers-ime.test.tsx` is the only test in the repo that mounts `useTerminalKeyboardShortcuts`, so the behavior is pinned against the real handler rather than a replayed control flow. Its harness now records both PTY writers — the composition route's committed glyph and the shortcut layer's direct `sendInput` — into one ordered array, so the assertions are on byte order, not on a predicate.
    - `does not read a shifted jamo as the committing Enter` → asserts `['ㅃ']`. Without the `code` check it reads `expected [ 'ㅃ', '\x1b\r' ] to deeply equal [ 'ㅃ' ]`, the bug's exact byte in its exact position.
    - `leaves the Enter chord slot free for a Ctrl+Enter behind a deferred Ctrl+Backspace` → asserts `['녕', '\x17', '\x1b[13;5u']`. Without the Windows narrowing it reads `expected [ '녕', '\x17' ]`: the Ctrl+Enter is silently lost. This is the risk the widening itself creates, and it had no coverage before.
    - `sends one Ctrl+Backspace, behind the syllable it committed` → asserts `['녕', '\x17']` plus `defaultPrevented` on the re-dispatch.
    - `sends Ctrl+Backspace straight through outside a composition` is the negative control that keeps the deferral from spreading to ordinary presses.
  - `terminal-ime-deferred-newline.test.ts` rejects `code: 'ArrowLeft'` alongside `code: 'KeyQ'` in the predicate fixtures, so the Japanese conversion-segment case is pinned by name and not only by the Korean one.
  - `terminal-ime-deferred-newline.test.ts` — the `isTerminalImeProcessEnter` fixture gains the `code` field it was missing, accepts `NumpadEnter` and the undeterminable `''` / `'Unidentified'`, and rejects `KeyQ`. `isTerminalImeRedispatchableKey` gains the unit block it was missing, covering the 229 composition marker it must not mistake for a re-dispatch.

## AI Review Report

Reviewed with Claude Code across four independent axes run without sight of each other or of my own conclusions: simplicity, impact radius, adversarial compatibility, and test quality. What they changed:

- **The narrowing was too strict, and it took the repo's own fixture to prove it.** The first version required `code === 'Enter' || 'NumpadEnter'`. The compatibility pass measured that a `code: ''` committing Enter is then dropped whenever the commit precedes the balancing keyup (the keyup fallback at `keyboard-handlers.ts:765-802` requires a still-pending composition, so it only rescues the other ordering). The test pass found the corroborating fixture: #11293's own chord-owner test is named for blank codes and sends them. The predicate now treats an undeterminable code as "cannot tell" and keeps `main`'s verdict.
- **The widening's own risk had no test.** Removing the `isWindows && (e.key === 'Enter' || imeProcessEnter)` narrowing broke nothing in the suite, yet loses a real `Ctrl+Enter`. That case is now pinned by bytes.
- **`isTerminalImeRedispatchableKey` had no unit block.** Its `keyCode !== 229` clause is not new defense: `main`'s condition was `e.key === 'Enter' && e.keyCode === 13 && !e.isComposing`, so the keyCode test is the part being generalized to a key set, not something added. It is now covered directly.
- **Simplification applied.** The `isRedispatchedEnter` local was folded into the chord ternary, which restores `main`'s `if` shape; the comment above it dropped a key list that `TERMINAL_IME_OWNED_KEYS` already states; two duplicate test rows that exercised the same branch were collapsed.
- **Cross-platform (macOS, Linux, Windows).** The predicate is reached only under `isWindows`; macOS delivers the jamo in `key` rather than as the literal `Process` (see `dispatchHangulProcessKey` in the Korean IME e2e), so `key === 'Process'` was already false there. Linux IBus does report `Process`/229 for jamo per the observed-sequence fixture and was protected only by that `isWindows` gate — this change makes the predicate itself correct rather than leaning on the gate. The absorb path the widening touches is platform-independent, keyed on `(code, timeStamp)`; the modifier-kind path is Windows-only. `TERMINAL_IME_OWNED_KEYS` is the same set everywhere. No shortcut label, menu accelerator, or path handling is touched.
- **Blast radius.** `isTerminalImeProcessEnter` is the only symbol whose signature changed and it has one production caller. Everything the PR touches is renderer-only; no main-process, preload, relay, CLI, or native-module counterpart is needed. The mobile package keeps its own key table and does not import this set.
- **Reuse.** No new module. `deferredNewlineSender`, `hasPendingTerminalImeComposition` and `createCapturedInputSender` are used as they already exist, and the key set is imported rather than restated. The only addition is one predicate, placed next to the module's existing IME predicates.
- **Performance.** One `Set.has` per keydown on a 12-element set, plus a few string comparisons short-circuited behind `key === 'Process'`. Without a pending composition nothing defers, so `tests/e2e/terminal-typing-latency.spec.ts` is unaffected.
- **UI quality.** No renderer UI touched.

## Security Audit

- **Input handling.** No new external input. Both predicates read `key`, `code` and `keyCode` off keyboard events the renderer already handles. No parsing, no deserialization, no new trust boundary.
- **Command execution and paths.** None added. The bytes written are the same constants `terminal-shortcut-policy.ts` already produced; this change decides only *when* they are written and whether a spurious or duplicate one is dropped, never *what*.
- **PTY writes.** Strictly reducing. The change can delay a write until the composition flush, suppress a re-dispatched duplicate, or stop a write that was never intended; it cannot introduce, synthesize, or repeat one. The shifted-jamo case is the security-relevant half: injecting an unintended newline at a shell prompt submits whatever line is pending.
- **Auth, secrets, IPC, dependencies.** Untouched. No new dependency, no new IPC channel, no telemetry, no logging.
- **Denial of service.** Deferred state is bounded per `(code, timeStamp)` and cannot queue faster than the user types.
- **Follow-up.** None blocking.

## Notes

Known limitations, all measured during review and none of them regressions against `main`:

- **Windows re-dispatch drift still doubles the non-Enter keys.** The chord owner tolerates a drifting `timeStamp` by comparing the modifier kind alone; `absorbRedispatchedEnter` requires an exact `(code, timeStamp)` match. On the Windows path this repo models with `redispatchTimestampOffset: 0.01`, a deferred `Ctrl+Backspace` therefore still goes out twice. `main` also sent it twice, in the worse order (`\x17, \x17, 녕` becomes `녕, \x17, \x17`), so this PR improves the ordering without closing the duplication. Closing it needs the chord owner generalized beyond one Enter slot, which is a larger change than this PR should carry.
- **A deferred byte is dropped if the pane's transport is swapped inside the 200 ms window.** `createCapturedInputSender` refuses to write once the pty id changes, so an SSH or relay reconnect during a composition discards the shortcut byte. That refusal is deliberate — writing a stray `\x17` into a freshly reconnected shell is worse than dropping it — but the window is new for these keys, since they previously went out synchronously.
- **A chord that does not end the composition now waits for the fallback.** Cursor-motion chords that the IME consumes without committing take up to 200 ms instead of going out immediately. It applies only while a composition is open and is bounded by the fallback timer.
- **The widening's premise for the non-Enter keys is inherited, not observed.** `main` establishes that Chromium re-dispatches the committing press with the same native `code` and `timeStamp`, and its real-IME e2e validates that for Enter. This PR assumes the same holds for the other IME-owned keys; the unit tests encode that assumption rather than prove it. Extending `tests/e2e/korean-ime-terminal-shift-enter-commit.spec.ts` with a `Ctrl+Backspace` case would settle it, and `installImeKeyEventLog` there already records the `code` of every keydown.
- **Not verified on Windows hardware.** I have no Windows machine, so the event shape is argued from this repo's own fixtures, and the Japanese conversion-segment case above rests on IME documentation rather than a capture. The fix is in the safe direction regardless: the predicate rejects only a known non-Enter physical code. A maintainer with a Windows box can confirm in one keystroke — type ㅃ mid-composition in a terminal pane on `main` and watch the prompt submit.
- **Why one PR.** The commits are separable and can be taken independently, but they are the same audit of the same predicate, they touch adjacent lines of the same file, and splitting them would have left two PRs conflicting with each other for no reviewer benefit.
- **Naming debt, not addressed here.** `deferredNewlineSender.absorbRedispatchedEnter` is now called for `Backspace` and the arrows, so its name is narrower than its behavior, as are `releaseRedispatchedEnter`, `clearRedispatchedEnters` and `TerminalImeEnterIdentity`. Renaming is mechanical and type-checked but churns `terminal-ime-deferred-newline.test.ts`, which #11293 landed recently, so it was left out of a PR whose point is to shrink.
- **Local test environment.** The four failing xterm test files above need a fresh `pnpm install` to pick up #11293's rewritten xterm patch. Nothing in this PR changes that.

X (Twitter): [@kunsanglee](https://x.com/kunsanglee)


OrcaWin:
Thank you for tackling this difficult IME/terminal-input edge case. We reproduced the behavior on a native Windows 2 host in headed Orca using Microsoft Japanese IME through the real renderer → native daemon → OpenConsole/ConPTY → PowerShell topology. The current base, published head `619e6c40da8b7aeb0a05309ca192eb3b362a2137`, and revert all incorrectly inject terminal bytes for ambiguous blank-code `Process` keydowns such as Shift+Left and Ctrl+Backspace; however, the strict suppression candidate prevents those false inputs only by also dropping a genuine Ctrl+Enter. Focused tests, the 2,967-test terminal-pane suite, and all typechecks passed, so this is a semantic ownership ambiguity rather than an incidental test failure.

Because neither the PR nor the bounded strict correction can distinguish IME-owned composition from genuine terminal input safely, this cannot land without risking either duplicate input or lost user commands. A reopenable direction would add delayed, authoritative IME ownership with bounded cleanup, then pass the same native Japanese matrix plus a physical Korean-layout control for blank-code navigation/editing and genuine Ctrl+Enter. We appreciate the contribution and the useful starting point; closing this version because the native evidence is decisive, not because the problem lacks value.

PR #11275, opened 2026-07-29
feat(worktrees): import a single hidden worktree from the visibility dialog

### LINKED ISSUE #11274: [Feature]: Import a single hidden worktree instead of showing every non-Orca one
### Problem or use case

Orca hides worktrees it didn't create, and one kind of them can't be un-hidden.

Non-Orca worktrees (`external` / `unknown-legacy`) are hidden while a repo's `externalWorktreeVisibility` is `hide`. That part is recoverable — the project menu has a "Show hidden worktrees" toggle.

Agent scratch worktrees are not. Anything sitting directly under a registered checkout and matching `AGENT_SCRATCH_PATH_PREFIXES` (`.claude/worktrees`, `.gsd-workspaces`) stays hidden no matter what that toggle says, because `shouldShowWorktree` returns false for `agent-scratch` before the toggle is ever consulted (`worktree-ownership.ts:226`). That's stablyai/orca#9535 doing its job.

The rule does leave a way out. A path listed in `repo.importedExternalWorktreePaths` is honoured first, and the comment right above says so: *"only an explicit import or selected checkout reveals it"*. But nothing writes to that list except the discovery inbox, and the inbox builds its candidates through `isUserFacingExternalWorktree`, which drops `agent-scratch` (`external-worktree-inbox.ts:63`). The exception exists in code and nowhere in the UI.

I ran into this with a repo that has 7 linked worktrees, all under `.claude/worktrees`. Three of them show up because their paths already happen to be in `importedExternalWorktreePaths`. The other four are unreachable, and toggling "Show hidden worktrees" does nothing for any of them.

It costs more than sidebar visibility. AI Vault happily lists the sessions that ran inside those worktrees, but Resume stays disabled, because it needs the session cwd to resolve to a worktree Orca knows about (`ai-vault-session-resume.ts:163`). Copy Resume Command is all that's left, so you end up back in a terminal outside Orca.

There's a milder version of the same problem for non-scratch worktrees: the only control is repo-wide. Wanting two out of seven means showing all seven.

### Proposed solution

Add a per-worktree import list to the "Non-Orca worktrees" dialog, below the toggle that's already there.

It lists every hidden worktree the repo has regardless of ownership, with Import on each row and "Import all N" when there's more than one. Scratch rows get a small badge so it's clear why they were hidden. Worktrees that an import revealed get a Hide action to undo it. Worktrees that are visible only because the repo-wide toggle is on don't get one, since removing their path wouldn't hide them anyway.

Nothing changes by default. Scratch stays hidden until someone imports it, which is what stablyai/orca#9535 wanted — this only makes that import possible. It also doesn't care what's in `AGENT_SCRATCH_PATH_PREFIXES`, so anything added there later works with no further UI.

The dialog is the right home for it. It's already the recovery point the sidebar points people at ("Change this later from the project menu"), and unlike the inbox it's always reachable — the inbox needs `externalWorktreeVisibility === 'hide'` plus a dismissed initial prompt.

Letting the inbox surface scratch worktrees would contradict stablyai/orca#9388, which says agent plumbing stays out of repo-wide discovery, and it would push them at people who never asked. The dialog you have to open deliberately.

Making the prefix list configurable goes the wrong direction. It lets you hide *more* paths, when the problem is a supported un-hide with no button.

Changing the matcher itself — broadening it, or dropping `.claude/worktrees` — would undo stablyai/orca#9535, and the narrowness is deliberate anyway.

Creating worktrees outside the checkout does work, since those classify as `external`. But `.claude/worktrees` is the agent CLI's convention, not something the user picks.

What I actually did was edit `importedExternalWorktreePaths` in `orca-data.json` by hand. Not something to recommend to anyone.

Related: stablyai/orca#9388, stablyai/orca#9535.

I have this working and can open a PR if the direction looks right.

5hseok:
## Summary

Closes #11274.

Adds a per-worktree import list to the "Non-Orca worktrees" dialog, so you can pull in one hidden worktree without flipping the whole repo — and so agent scratch worktrees become reachable at all.

`shouldShowWorktree` already treats a path in `importedExternalWorktreePaths` as an override that beats the `agent-scratch` rule; the comment above it says *"only an explicit import or selected checkout reveals it"*. Nothing could write a scratch path there, though. The only writer is the discovery inbox, whose candidates pass through `isUserFacingExternalWorktree`, which filters `agent-scratch` out. So a worktree under `.claude/worktrees` couldn't be opened in Orca short of editing `orca-data.json`.

That also kept AI Vault's Resume disabled for every session that ran in one of those worktrees, since Resume needs the session cwd to resolve to a known worktree.

The new section sits under the existing toggle and lists every hidden worktree — `external`, `unknown-legacy`, `agent-scratch` — with Import per row and "Import all N" for bulk. Scratch rows carry a badge so an unfamiliar worktree doesn't read as a sidebar bug.

Two things worth flagging for review:

Hide is only offered when the explicit import is the sole reason a worktree shows. A worktree revealed twice over — imported individually *and* covered by the repo-wide `show` toggle — would stay on screen after its path is dropped, so the button would silently do nothing. Rather than re-derive that condition, the selector asks `shouldShowWorktree` with the import list withheld and keeps the row only if the answer is no.

The write is optimistic with a rollback: set the path list, force an authoritative refresh, and if the refresh fails put the old list back and surface the error in the dialog rather than leaving the repo record ahead of the sidebar. Same shape as `new-external-worktrees-inbox-actions.ts`.

No defaults move. Scratch stays hidden until imported, which is what #9535 asked for. The section is also ownership-agnostic, so entries added to `AGENT_SCRATCH_PATH_PREFIXES` later need no UI work.

## Screenshots

Repo parked on the default `hide`, with one external worktree and two scratch ones. The scratch worktree at the bottom was imported from this dialog.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` (see note)
- [x] `pnpm build`

39,676 passing. One suite fails on my machine for a reason that predates this branch: `automation-schedules.test.ts` expects `'Sundays at ...'` and `Intl` on a `ko-KR` host returns `'일요일s at ...'`. It passes under `LANG=en_US.UTF-8`. A few others (`transcript-watch-liveness`, `project-view-wrapper-source-context-boundary`, `windows-port-scan`) failed intermittently under full-suite parallelism across runs and pass in isolation. None are in the import graph of anything this PR touches.

27 new unit cases plus one E2E spec:

- Selectors: which ownerships are offered and which aren't, non-authoritative listings returning nothing rather than a list built from stale metadata, an imported worktree that the `show` toggle also reveals being dropped from the hide list while an imported scratch one is kept, and path add/remove being idempotent and separator-tolerant.
- Actions: single and bulk import, hide dropping only the requested path, rollback when the refresh fails, the ambiguous-state message when the rollback write fails too, error without a refresh when the repo update fails, and an empty selection touching nothing.
- Component: short branch names, the branch line collapsing when a worktree is named after its branch, the badge only on scratch rows, bulk import only above one candidate, pending disabling actions.
- E2E (`tests/e2e/hidden-worktree-import.spec.ts`): real git worktrees, one external and one under `.claude/worktrees`. Neither is in the sidebar with the repo on `hide`; importing the scratch one puts it there and leaves the external one hidden; Hide takes it back out.

Also exercised by hand against a repo with 7 real scratch worktrees, 4 of which were unreachable before this.

## AI Review Report

Reviewed with an AI coding agent.

- Cross-platform: no new path building or platform branching. Comparison reuses `normalizeRuntimePathForComparison` through the existing inbox helpers, which already cover Windows drive-letter casing, separator folding and WSL UNC aliases. The scratch matcher is untouched, and its anchoring to a registered checkout means nothing here assumes a POSIX-shaped path. No new shortcuts or shortcut labels.
- SSH and remote: `importedExternalWorktreePaths` is already accepted by the runtime repo-update schema, so the action works unchanged on SSH and runtime hosts. Selectors require `authoritative === true`, so a disconnected SSH repo falling back to `metadata-fallback` shows no list rather than a stale one. The dialog is gated on `isGitRepoKind`, so folder workspaces are unaffected.
- Compatibility: classification, the prefix table and `shouldShowWorktree`'s ordering are all unchanged — this only feeds a branch that already existed. The inbox and `ImportedWorktreesVisibilityLine` behave identically; their selectors are untouched and their tests pass.
- UI: `aria-label` and `aria-busy` on the section, truncation with the full path in a tooltip, disabled buttons while pending, `role="alert"` on the error. All nine new strings go through `translate()`, with `ko` / `ja` / `es` / `zh` translated rather than left as English.

What came out of the review: `useAppStore.getState()` calls inside the action callback were swapped for the already-subscribed selectors, and the E2E capture caught rows printing the worktree name and its branch as two identical lines, since a worktree with no Orca metadata takes its display name from the branch — the branch line now collapses when they match.

The bot reviews on this PR then caught a real one, and it was mine: the hide list checked only whether a path sat in `importedExternalWorktreePaths`, never whether the repo-wide toggle was also revealing the worktree, so hide could still be a no-op for the compound case. The test I had written for it passed only because its import list was empty. The selector now defers to `shouldShowWorktree`, and the replacement test fails against the old predicate. Four smaller ones came from the same pass: a rollback whose own write failed reported a plain retry, the dialog's action state wasn't keyed by project, the E2E spec deleted its fixture before Electron shut down and built `data-worktree-id` selectors without escaping, and the zh strings didn't use the catalog's own terms for worktree and agent.

## Security Audit

No filesystem access, no path construction, no shell interpolation, no command execution — the change compares paths and writes a string array.

Every path comes from `git worktree list` output that Orca scanned and classified itself, not from user input. It goes through the existing `updateRepo`, which already validates `importedExternalWorktreePaths` as an array of strings before persisting, and through the runtime schema for SSH hosts. Importing a path only affects sidebar visibility; it can't make Orca show a path `git worktree list` didn't report, because the selectors work off the detected list, so a stale entry in the record simply never matches a row.

No new IPC channels, dependencies, secrets, auth or network paths.

## Notes

- The repo-wide card above the new section still counts only non-scratch worktrees ("1 worktree available to import" in the screenshot while three rows are listed). That's existing behaviour and accurate for what the toggle governs, but it reads oddly next to the list. Happy to reword it if you'd rather.
- `importedExternalWorktreePaths` accumulates entries for worktrees that no longer exist. Harmless — nothing matches them — but the list only grows. Pruning has to tell "deleted" apart from "not in this scan", so it's a separate change.
- A per-repo setting for extra scratch prefixes, for agent CLIs that write elsewhere, would compose with this rather than conflict. Not in scope here.

## ELI5

Orca hides worktrees agents create inside your repo, and there was no button to un-hide one even though the code was written to allow it. The project's worktree dialog now lists them so you can bring in just the ones you want.


brennanb2025:
Thanks for this work — per-worktree recovery from the visibility dialog landed in #13652, so this PR is now superseded.

PR #11285, opened 2026-07-29
fix(repos): prune hydrated restored folder owners

### LINKED ISSUE #11194: [Bug]: Stale restored runtime owners are never pruned while any remote host is unreachable, pinning local folder workspaces to a dead runtime
### Operating system

macOS

### Orca version

Reproduced against `main` (v1.4.160-rc.3) with an E2E harness run; also observed on 1.4.15x desktop builds.

### Details

**Short summary:**  <br>When one or more remote (runtime) hosts are unreachable, local folder workspaces are affected too: the stale "restored runtime owner" that boot derives from per-host session partitions is never pruned, so a local folder workspace keeps being treated as owned by the dead remote for the rest of the session (and the state re-persists, so it never heals).

**What happened?**  <br>With a folder registered on a remote host and that remote temporarily unreachable, a local folder workspace's provisional runtime ownership from session restore sticks forever. Consequences observed: the workspace's session state keeps being written back into the dead host's partition (`workspaceSessionsByHostId`), the stale key defeats the stale-tab GC in `fetchAllWorktrees`, and any surface that consults the restored owner before catalogs hydrate routes path checks/terminal spawns for the local folder to the dead runtime host.

**How can we reproduce it?**

1. Create a local project group + folder workspace and open it once (so its key is in session state).
2. Quit Orca. In the profile's `orca-data.json`, add a `workspaceSessionsByHostId["runtime:<envId>"]` partition claiming the workspace key (`folder:<id>`), and register a runtime environment `<envId>` in `orca-environments.json` whose endpoint is unreachable (e.g. `ws://127.0.0.1:1`). This mirrors what a real session on a later-unreachable remote leaves behind.
3. Launch Orca and wait for the startup all-host catalog refresh to settle.
4. Inspect `restoredRuntimeHostIdByWorkspaceSessionKey` in the renderer store: the `folder:<id>` key still maps to `runtime:<envId>` even though the workspace's folder and project-group catalogs are fully hydrated locally. It stays that way for the whole session, and the doctored partition is re-persisted on quit.

**Root cause (traced in code):**  <br>`fetchFolderWorkspacesForAllHosts` (src/renderer/src/store/slices/repos.ts) prunes restored runtime session owners only when **every** host fetch succeeded (`if (!failed)`), so a single unreachable remote skips the prune entirely — including for workspaces whose catalogs already hydrated locally. The write-back loop in `getFolderWorkspaceRuntimeHostId` (src/renderer/src/lib/workspace-session-host-persistence.ts) then keeps the stale claim alive across restarts. The all-or-nothing gate dates back to [#7087](<https://github.com/stablyai/orca/issues/7087>); it exists because the per-key drop rule ("workspace missing from catalogs → drop") is only safe when catalogs are known-complete, but the collateral is that one dead host pins every stale owner.

**Anything else that might help:**  <br>Note on severity: once local catalogs hydrate, owner resolution has an authoritative escape (the group's stamped `executionHostId`), so the visible impact after the refresh settles is limited — verified pre-fix that opening the workspace still succeeds at that point. The damage is the permanent stale state itself: session write-back into the dead host's partition, defeated stale-tab GC, and the misrouting window repeating on every boot while the remote stays unreachable.  <br>Verified with an Electron E2E run (app-level, doctored profile as above): on current `main` the stale owner survives catalog settling; with a per-key-safe prune (drop once the workspace's folder+group catalogs hydrate, or once its own host's catalog answers without the workspace) it is cleared while a genuinely-unhydrated remote workspace keeps its restored owner. A per-key-safe prune fix with regression tests exists; happy to submit it, or leave this report for maintainers to take. Related but distinct: [#10799](<https://github.com/stablyai/orca/issues/10799>) fixed the opposite routing direction (SSH-targeted setup acting locally); [#1640](<https://github.com/stablyai/orca/issues/1640>) (closed) and stablyai/orca#10031 touch adjacent remote-failure/multi-host areas.

xianjianlf2:
## ELI5

Orca remembers which computer owns a restored folder workspace. Once it has a current folder list, it can forget outdated entries—but an old response must not erase information while a newer response is still arriving.

Fixes #11194

## Summary
- Prune restored runtime owners for folder workspaces once that specific folder and project-group catalog is hydrated.
- Keep missing folder-workspace owners during partial all-host failures, so genuinely unhydrated remote sessions are not moved local.
- Preserve the previous all-host complete behavior for dropping owners whose folder workspace is absent from complete catalogs.

## AI review summary
- Cross-platform: renderer state-only change, no platform-specific filesystem behavior.
- Remote/SSH: handles unreachable runtime hosts without letting them pin hydrated local folder workspaces.
- Agent integrations: avoids stale host ownership being re-persisted into dead runtime partitions.
- Performance: linear pass over existing restored owner map only after catalog refresh.
- UI: no visual changes.
- Security: no new external input or trust boundary.

## Test plan
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/store/slices/repos-all-hosts-folder-workspaces.test.ts`
- `pnpm typecheck`
- `pnpm lint`

X handle: not provided.

## Readiness review — 2026-09-08

**Reviewable.** Fixed superseded folder catalogs marking hosts hydrated and pruning restored owners. Generation-fenced admission/pruning regression passes; 18 focused tests and web typecheck passed.

Validated head: `a4867596eed607221dfee3d79e7d26addc9c1ff5`. Updated with main `2265fce5912487f6a88ee0827dac14953b6a72c8`. Combined branch passed node, CLI, web and mobile typechecks plus 90 focused regression tests. Platform/live checks above remain separate from code checks.

## Visual proof

Not applicable as rendered UI proof. Ownership metadata change; no distinct visual state proves the race. The functional evidence is the tests or probe recorded in the readiness review; an illustrative screenshot would not establish correctness.


xianjianlf2:
Resolved conflicts against current upstream `main`. Upstream split the repo slice into folder-workspace catalog helpers, so I moved the restored-owner pruning change into `folder-workspace-catalog.ts` and `folder-workspace-catalog-actions.ts`. The all-host loader now clears missing restored owners only for folder workspace hosts that were actually hydrated.

Validation (Node 24.18.0):
- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/store/slices/repos-all-hosts-folder-workspaces.test.ts` (6 passed)
- `pnpm lint -- src/renderer/src/store/folder-workspaces/folder-workspace-catalog.ts src/renderer/src/store/folder-workspaces/folder-workspace-catalog-actions.ts src/renderer/src/store/slices/repos.ts src/renderer/src/store/slices/repos-all-hosts-folder-workspaces.test.ts`
- `git diff --check`

Committed with `--no-verify` after the manual lint pass because lint-staged attempts to process the full upstream merge in very large chunks locally.

Current head: 2bbcf177

PR #11300, opened 2026-07-29
fix(runtime): avoid immediate WebSocket heartbeat sweep

### LINKED ISSUE #11298: [BUG]: Immediate WebSocket heartbeat closes mobile pairing connection on Linux ARM64
### Operating system

Linux

### Orca version

1.4.159

### Details

Short summary:

On Linux ARM64, `orca serve` accepts the mobile WebSocket connection, then closes it immediately before the E2EE handshake can complete.

What happened?

Environment:

- Raspberry Pi 4
- Debian ARM64
- Orca 1.4.159
- Orca Mobile Android 0.0.32
- Direct connection through Tailscale

The Android pairing log shows:

```text
WebSocket open
Sent e2ee_hello
Awaiting server e2ee_ready
WebSocket closed
```

The issue is also reproducible locally, without Android or Tailscale:

```js
const ws = new WebSocket('ws://127.0.0.1:6768/');

ws.onopen = () => {
  console.log('OPEN');
  ws.send('{"type":"test"}');
};

ws.onclose = e =>
  console.log('CLOSE', e.code, JSON.stringify(e.reason));
```

Result:

```text
OPEN
CLOSE 1006 ""
```

How can we reproduce it?

1. Run:

```bash
orca serve --mobile-pairing --pairing-address 100.x.x.x
```

2. Scan the generated QR code with Orca Mobile.
3. The WebSocket opens, then closes immediately before `e2ee_ready`.
4. The same behaviour occurs locally on `ws://127.0.0.1:6768`.

Anything else that might help:

The issue appears to be caused by the immediate heartbeat sweep in `RemoteRuntimeServerHeartbeat.start()`:

```js
this.sweep(getClients());
```

The first heartbeat is started before the WebSocket `pong`, `message`, `close`, and `error` listeners are attached.

As a test, I removed only this immediate `sweep()` call while keeping the periodic 15-second heartbeat.

After patching, the local test correctly returns:

```text
OPEN
CLOSE 4001 "Invalid e2ee_hello"
```

Mobile pairing then works successfully.

A possible fix would be to attach all WebSocket listeners before starting the initial heartbeat, instead of removing the heartbeat entirely.

jmdall:
Fixes #11298

## Summary

Defer the first WebSocket heartbeat sweep until the normal interval tick.

The immediate sweep could close a newly accepted WebSocket before the E2EE handshake completed on Linux ARM64. Removing the immediate sweep fixed Orca Mobile pairing on a Raspberry Pi 4 running Debian ARM64.

There is no change to the recurring heartbeat. The first liveness probe now occurs after the configured interval instead of immediately when the first connection is accepted.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Manual testing performed on:

- Raspberry Pi 4
- Debian ARM64
- Orca 1.4.159
- Orca Mobile Android 0.0.32

Before the change, a newly opened local WebSocket closed immediately with code `1006`, before its first message was processed.

After removing the immediate sweep, the same invalid E2EE test message was processed normally and returned:

```text
CLOSE 4001 "Invalid e2ee_hello"
```

Orca Mobile pairing then completed successfully.

Updated the existing heartbeat unit tests to cover the deferred initial sweep. 

## AI Review Report

The change was reviewed for:

- heartbeat timer lifecycle;
- dead and half-open socket cleanup;
- WebSocket handshake timing;
- impact on authenticated and pre-authentication connections;
- cross-platform compatibility on macOS, Linux, and Windows.

The main identified trade-off is that the first liveness ping is delayed until the first interval tick. With the production 15-second interval, an unresponsive newly accepted socket may now take approximately 30 seconds to be reaped instead of being probed immediately.

The change does not touch shortcuts, labels, paths, shell behavior, native APIs, or Electron platform-specific behavior.

## Security Audit

This change does not modify:

- authentication or E2EE validation;
- input parsing or message limits;
- command execution;
- file or path handling;
- secrets or dependencies;
- IPC boundaries.

The existing pre-authentication timeout remains active and still terminates clients that do not begin authentication within the configured timeout.

## Notes

The issue was reproduced on Linux ARM64. The modified heartbeat code is platform-independent, so the delayed initial sweep also applies to macOS and Windows.

### DIFF
--- a/config/reliability-gates.jsonc
+++ b/config/reliability-gates.jsonc
@@ -1,6 +1,6 @@
 {
   "schemaVersion": 1,
-  "updatedAt": "2026-07-28",
+  "updatedAt": "2026-07-29",
   "policy": {
     "maturityLevels": [
       "experimental",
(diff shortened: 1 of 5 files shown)


hp198410:
I still can’t connect to the service running on my Pi 4 from Orca on another computer.
I have installed the latest version, 1.163.0-rc0, which includes the merged branch changes, on both devices.
The following error is displayed:
Host unavailable
Make sure Orca is running on the other host and that the network or SSH tunnel can reach 100.xxx.xxx.xxx

PR #11307, opened 2026-07-29
perf(ai-vault): stop the session projection from stalling workspace switches

nwparker:
## Symptom

With the AI Vault panel open, clicking a workspace card froze the renderer for ~1s. The highlight is applied by direct DOM mutation inside the click handler, so the JS "click task" stayed fast (~1ms) and the existing `worktree-switch-responsiveness.spec.ts` budget never tripped — but the frame never painted.

Measured on a real 1124-worktree profile before this change:

```
first paint after click   p50=110ms
max frame gap             p50=973ms   <- the visible stall
main pane rendered        p50=1046ms
worst long task           p50=970ms
```

## Cause

A CPU profile of one switch:

| self time | function |
| --- | --- |
| 1701ms | `isSessionInWorktreePath` |
| 1664ms | `isPathInsideOrEqual` |
| 103ms | `buildWorktreeCandidates` |

`resolveAiVaultSessionWorktreeInfo` rebuilt the **entire** candidate list inside the per-session call, then `.filter()`ed it with no early exit and `.sort()`ed it. At 1124 worktrees x 500 sessions that is ~1.1M containment checks and ~2M path normalizations per switch — and both `AiVaultPanel` projections depend on `activeWorktreeId`, so it all re-runs on the click path.

The shape is old. What made it *visible* is `a1a78da878` (#10841), which added an unconditional `.normalize('NFC')` to `normalizeRuntimePathForComparison` — the innermost function in that loop. `.normalize()` is an ICU call that can't be inlined.

`git bisect` over a replay of the real projection puts the step exactly there:

```
58ef46d252  344ms   (parent)
a1a78da878  420ms   (first bad)
```

Confirmed independently: deleting just that one `.normalize('NFC')` on current `main` reproduces the 420ms -> 348ms delta.

## Fix

The NFC fold is a correctness fix (#10832 — non-ASCII workspaces never matched their own sessions), so this keeps it and removes the fan-out instead:

- **Cache the candidate list** per `(worktrees, repos)` identity pair. Both are referentially stable in the store, so one `WeakMap` hop replaces N rebuilds.
- **Precompute each candidate's matcher** with `createNormalizedPathInsideOrEqualMatcher`, so the session cwd is normalized once per *session* rather than once per *candidate*. That helper was added by #10881 for exactly this shape; the AI Vault fan-outs were paying its cost without adopting it.
- **Replace `filter + filter + sort` with a single max pass** — the result is one candidate, and sorting 1000+ entries per session dominated.

## Results

Projection cost on the real 1124-worktree profile: **423ms -> 13ms** (32x).

In-app, after the fix: first paint `p50=110ms -> 31ms`.

## Correctness

Unicode support from #10841 is preserved and now covered *through* this projection, in both directions (NFD workspace path vs NFC cwd, and the reverse) — the precomputed matchers had to keep the fold working, which the new tests pin.

Output was diffed against the previous implementation across **4501 session/worktree combinations** built from the real 1124-worktree profile — exact paths, nested cwds, prefix-siblings, trailing slashes, WSL UNC, Hangul/accented paths, three host scopes, misses — **byte-identical, zero differences**.

The single-pass max also reproduces `Array.sort`'s stable tie-breaking (first of equals wins); verified against the old comparator.

## Tests

`src/renderer/src/components/right-sidebar/ai-vault-session-worktree.test.ts`:

- NFD workspace path vs NFC transcript cwd (Hangul)
- NFD transcript cwd vs NFC workspace path (accented)
- sibling sharing a path prefix must not match (matcher boundary)
- deepest owning worktree wins when workspaces nest
- current-path beats another worktree's prior-path on a tie
- **timing regression**: 1200 worktrees x 400 sessions must resolve in <150ms (was ~350ms+, now ~10ms)

That timing test earned its place during development — it caught a `repos = []` default allocating a fresh array per call, which made every `WeakMap` lookup miss.

## Also included

Two tools used to find this, kept because the gap they cover is what let the regression through:

- `tools/benchmarks/workspace-switch-paint-latency.mjs` — attaches over CDP and measures **first paint after click** plus max frame gap. The existing e2e spec only times the synchronous click task, which is exactly why a 1s stall could ship unnoticed.
- `tools/benchmarks/ai-vault-worktree-projection-bench.mjs` — replays the projection against real profile data, esbuild-compiling whatever is checked out. Bisecting through the real app costs ~6min/step; this is ~2s, which is what made the bisect feasible.

## Verification

- 4805 tests pass across `right-sidebar/` and `shared/`
- `tsc --noEmit -p config/tsconfig.tc.web.json` clean
- `oxlint` clean

## Note for follow-up

This removes the AI Vault contribution to the stall, but a ~550ms gap remains on switch from other work on that path. Investigating separately — the open question is why agent-history work is on the workspace-switch critical path at all, rather than deferred behind terminal restore.

### DIFF
--- a/src/renderer/src/components/right-sidebar/ai-vault-session-worktree.test.ts
+++ b/src/renderer/src/components/right-sidebar/ai-vault-session-worktree.test.ts
@@ -202,6 +202,142 @@ describe('resolveAiVaultSessionWorktreeInfo', () => {
       worktreeId: worktree.id
     })
   })
+
+  it('matches an NFD workspace path to an NFC transcript cwd', () => {
+    // macOS yields NFD on disk while Claude Code records cwd in NFC (#10832).
+    // The projection precomputes its matchers, so the fold has to survive that.
+    const nfcPath = '/repo/프로젝트'
+    const worktree = makeWorktree({
+      id: `repo-1::${nfcPath.normalize('NFD')}`,
+      displayName: '',
+      path: nfcPath.normalize('NFD')
+    })
+
+    expect(
+      resolveAiVaultSessionWorktreeInfo({
+        session: { ...baseSession, cwd: `${nfcPath}/src` },
+        worktrees: [worktree],
+        activeWorktreeId: null
+      })
+    ).toMatchObject({
+      status: 'active',
+      worktreeId: worktree.id
+    })
+  })
+
+  it('matches an NFD transcript cwd to an NFC workspace path', () => {
+    const nfcPath = '/repo/café'
+    const worktree = makeWorktree({
+      id: `repo-1::${nfcPath}`,
+      displayName: '',
+      path: nfcPath
+    })
+
+    expect(
+      resolveAiVaultSessionWorktreeInfo({
+        session: { ...baseSession, cwd: `${nfcPath}/src`.normalize('NFD') },
+        worktrees: [worktree],
+        activeWorktreeId: null
+      })
+    ).toMatchObject({
+      status: 'active',
+      worktreeId: worktree.id
+    })
+  })
+
+  it('does not treat a sibling sharing a path prefix as the owner', () => {
+    // Guards the candidate matcher's boundary: '/repo/orca-sibling' must not
+    // match under '/repo/orca'.
+    const worktree = makeWorktree()
+    const sibling = makeWorktree({
+      id: 'repo-1::/repo/orca-sibling',
+      displayName: 'sibling',
+      path: '/repo/orca-sibling'
+    })
+
+    expect(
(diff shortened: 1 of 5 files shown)


nwparker:
Superseded by #11303, which landed the same fix with a better structure: removing `activeWorktreeId` from the memo deps means a switch reuses the cached map entirely rather than just rebuilding it faster. Verified on the merged main — the map build is 18.6ms at 1124 worktrees / 500 sessions, and the per-session resolver this PR optimized now has no production callers.

Reopening the two pieces that are still additive on top of #11303 (a timing regression guard and a path-boundary test) as a separate, smaller PR.

PR #11309, opened 2026-07-29
feat(editor): add JSP syntax highlighting backed by the HTML language service

taehwanis:
## Summary

`.jsp` and `.jspf` files open as plaintext today — they are not in `EXT_TO_LANGUAGE`, so `detectLanguage()` returns `'plaintext'` and nothing is highlighted.

This registers a dedicated `jsp` language with a Monarch tokenizer for directives, scriptlets, expressions, EL and JSTL/custom tags, **and attaches Monaco's HTML language service to it**.

That second part is the reason this is not just another tokenizer. Monaco already serves Razor and Handlebars through `registerHTMLLanguageService` because they are "HTML plus server-side template syntax" — exactly what JSP is. Reusing it gives tag completion, hovers, document symbols and syntax-aware folding inside JSP markup, which a tokenizer-only registration silently gives up. The worker route (`case 'jsp'` in the existing html group) is required rather than optional: `MonacoEnvironment.getWorker` takes precedence over Monaco's own fallback, so without it every language-service call would reach the plain editor worker and fail.

`rename` and both formatting providers stay off. They are the ones that act on a parse of the file as HTML, and `<% ... %>` is not markup — formatting would relocate it and rename would edit spans derived from a mis-parse.

Related: **#10858 by @hwantage** addresses the same issue with a Monarch tokenizer. I read that PR before writing this, and it is what prompted me to check whether the HTML language service could be attached to a custom language id — that turned out to be the gap this PR closes. If the maintainers prefer that PR's shape, the language-service wiring here can be lifted into it and I am happy to close this one.

## Screenshots

Same `demo.jsp` opened in a build from this branch.

**Before** — `.jsp` is not in the extension table, so the file falls back to `plaintext`:

JSP file rendered as plaintext

**After** — directives, scriptlets, EL, JSTL tags and the embedded JS/CSS blocks are highlighted:

JSP file with directives, scriptlets, EL and JSTL highlighted

Verified in the running app, not only in tests: typing `<` inside markup produces HTML tag completions, and folding handles are placed on tag boundaries rather than indentation.

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [x] `pnpm build`

The two unchecked boxes are explained under Notes.

## AI Review Report

Three independent AI review passes were run against this branch — layering/registration, tokenizer correctness, and conventions/security — each checking Monaco's own sources rather than reasoning from memory. They found four defects that are fixed in `ffaa740a2`:

- **State could leak past its terminator (two paths).** `tagRest` and both attribute-value states had no `<%--` rule, so a JSP comment inside a tag (`<c:out value="x" <%-- note --%> />`) entered the scriptlet state, where `--%>` was absorbed by the operator character class instead of closing it — Java tokenization then ran to end of file. The same class contained both `%` and `>`, so an operator run written against the terminator (`<%=i++%>`, no spaces) swallowed it. Fixed by adding the comment rule to all three states and giving `%` a negative lookahead.
- **`rename: true` edited the buffer from an untrusted parse.** It is the only enabled provider that writes, and `htmlWorker.doRename()` derives its edits from parsing the file as HTML, which scriptlets break. Turned off, which also makes the configuration internally consistent with formatting being off.
- **`ignoreCase: true` mis-coloured Java identifiers.** Carried over from the sibling Vue/Svelte/Astro modules, where it only affects HTML tag matching. JSP is the one language here that tokenizes a case-sensitive language directly, so `List list`, `Map map` and `String string` coloured the variable as a type. Turned off; the four HTML tags that genuinely need case-insensitivity spell the characters out.
- **Two mode-configuration flags were inert.** `diagnostics` and `colors` are not read by the HTML mode at all (it consumes ten keys; those two are handled only by the css and typescript modes), so an earlier revision of this PR described them as the guard against scriptlet false positives. That was wrong — the actual guard is formatting and rename staying off. The flags are removed and the comment corrected.

Tests grew from 8 to 16 as a result, covering the paths the reviews identified as unprotected: the embedded-language pop contract (Monarch throws, not just mis-colours, if it is missing), match lengths that distinguish the namespaced-tag rule from the generic one, every state's exit delimiter, and a state-reference integrity check.

Cross-platform compatibility (macOS, Linux, Windows) was explicitly checked. Nothing here branches on platform: the change is a language registration plus lookup-table entries, and extension matching already handles both path separators and casing. No keyboard shortcuts, labels, shell behavior, path utilities or Electron platform differences are touched, and nothing is specific to SSH, remote or local hosts, a particular agent, or a git provider. The HTML worker is already bundled and routed, so there is no new dependency; the worker is constructed only when a `jsp` model exists and is released after idle.

## Security Audit

No meaningful risk surface. The change registers a tokenizer and reuses an already-bundled language worker. There is no command execution, no path resolution, no IPC, no auth, no secrets and no dependency change. The tokenizer only classifies text; it never evaluates scriptlets or EL. The regex rules were reviewed individually for catastrophic backtracking — every quantifier sits on a single character class or on an alternation whose branches start with disjoint characters, so there is no nested-quantifier path, and Monarch matches within a single line regardless. No follow-up needed.

## Notes

- `pnpm lint`: oxlint and the code-quality audits pass. The chain stops later at `verify:skill-bundle-manifest`, which reports stale generated artifacts under `resources/skills/` — a pre-existing state unrelated to this change, and those files are not part of this diff.
- `pnpm test`: the full suite on a Windows machine fails a set of pre-existing tests that require a POSIX shell (`spawn /bin/sh ENOENT`); the count is unchanged by this PR. The test files touched here pass.
- Scriptlets and EL inside `<script>` blocks are not highlighted as JSP. Once the tokenizer hands the block to the embedded JavaScript tokenizer, Monarch cannot re-enter host rules — the same limitation Monaco's own HTML mode has. Worth stating explicitly because embedding `${...}` in a script block is common in real JSP.
- Registering the HTML language service twice would create a second worker and provider set — the call has no internal dedupe — so it sits behind the same guard as the language registration, and a test asserts it runs once across repeated calls.
- `mobile/src/session/mobile-file-language.ts` has a separate extension table for the mobile surface and is not touched here; it feeds a different highlighter.

### DIFF
--- a/src/renderer/src/lib/language-detect.test.ts
+++ b/src/renderer/src/lib/language-detect.test.ts
@@ -47,6 +47,12 @@ describe('detectLanguage', () => {
     expect(detectLanguage('C:\\Users\\alice\\.codex\\LOG.JSONL')).toBe('jsonl')
   })
 
+  it('maps .jsp/.jspf files to the dedicated jsp language id (case-insensitive)', () => {
+    expect(detectLanguage('src/main/webapp/WEB-INF/views/list.jsp')).toBe('jsp')
+    expect(detectLanguage('src/main/webapp/WEB-INF/include/header.jspf')).toBe('jsp')
+    expect(detectLanguage('views/DETAIL.JSP')).toBe('jsp')
+  })
+
   it('keeps .json/.jsonc on the built-in json language and unknown on plaintext', () => {
     expect(detectLanguage('config/settings.json')).toBe('json')
     expect(detectLanguage('config/tsconfig.jsonc')).toBe('json')
--- a/src/renderer/src/lib/language-detect.ts
+++ b/src/renderer/src/lib/language-detect.ts
@@ -84,6 +84,8 @@ const EXT_TO_LANGUAGE: Record<string, string> = {
   '.vue': 'vue',
   '.svelte': 'svelte',
   '.astro': 'astro',
+  '.jsp': 'jsp',
+  '.jspf': 'jsp',
   '.sv': 'systemverilog',
   '.svh': 'systemverilog',
   '.v': 'verilog',
--- a/src/renderer/src/lib/monaco-languages/register-jsp.test.ts
+++ b/src/renderer/src/lib/monaco-languages/register-jsp.test.ts
@@ -1,0 +1,161 @@
+import { describe, expect, it, vi } from 'vitest'
+import {
+  jspHtmlModeConfiguration,
+  jspLanguageConfiguration,
+  jspMonarchLanguage,
+  registerJspLanguage
+} from './register-jsp'
+
+type MonarchAction = {
+  next?: string
+  nextEmbedded?: string
+  switchTo?: string
+}
+type MonarchRule = [RegExp, string | MonarchAction, string?] | { include: string }
+
+function isRuleEntry(rule: MonarchRule): rule is [RegExp, string | MonarchAction, string?] {
+  return Array.isArray(rule)
+}
+
+function findRule(
+  state: string,
+  source: string
+): [RegExp, string | MonarchAction, string?] | undefined {
+  const tokenizer = jspMonarchLanguage.tokenizer as Record<string, MonarchRule[]>
+  const matched = tokenizer[state].find((rule) => {
+    if (!isRuleEntry(rule)) {
+      return false
+    }
+    const [regexp] = rule
+    regexp.lastIndex = 0
+    const match = regexp.exec(source)
+    return match !== null && match.index === 0
+  })
+
+  return matched && isRuleEntry(matched) ? matched : undefined
+}
+
+function tokenFor(state: string, source: string): string | undefined {
+  const rule = findRule(state, source)
+  if (!rule) {
+    return undefined
+  }
+  const [, action] = rule
+  return typeof action === 'string' ? action : action.next
+}
+
+function nextStateFor(state: string, source: string): string | undefined {
+  const rule = findRule(state, source)
+  if (!rule) {
+    return undefined
+  }
+  const [, action, shortcut] = rule
+  const next = typeof action === 'object' ? (action.next ?? action.switchTo) : shortcut
+  return next?.replace(/^@/, '')
+}
+
+describe('registerJspLanguage', () => {
+  it('registers the jsp language and attaches the HTML language service once', () => {
+    const languages: { id: string }[] = [{ id: 'html' }]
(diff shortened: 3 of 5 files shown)


taehwanis:
Thanks both bots — all four comments were made against the first commit (`007337fdc`). Status after `ffaa740a2` and `a04b55a65`:

**Fixed**

- **`ignoreCase` colouring variables as types** (greptile) — correct, and worse than cosmetic in practice: `List list`, `Map map` and `String string` are the most common shapes in a scriptlet, so the variable took the same token as the class. `ignoreCase` is now off, and the four HTML tags that genuinely need it spell the characters out. A test asserts `list = null` tokenizes as `identifier` while `List list` still starts with `type`.
- **Deferred EL `#{...}` not highlighted** (greptile) — added to `root`, `tagRest` and both attribute-value states, reusing the existing `elExpression` state.
- **Line/block comment swallowing `%>`** (coderabbit) — good catch, and the sharpest of the four. `<% doStuff(); // done %>` is valid JSP because the container looks for `%>` first, so the comment has to stop there rather than run to end of line; otherwise the tokenizer stays in the scriptlet for the rest of the file. Both comment rules now carry a `(?!%>)` guard, with a regression test on the match lengths.

**Not applicable**

- **`monaco.html.registerHTMLLanguageService` namespace** (coderabbit, flagged Critical) — the sources cited describe `monaco.languages.html`, which is indeed not where it lives. In the bundled types for the version this repo pins, `html` is a **top-level** export alongside `languages` (`monaco.d.ts`: `monaco_contribution$2 as html`). I ran into exactly this while writing it: `Monaco.languages.html.ModeConfiguration` fails with TS2694, and `Monaco.html.ModeConfiguration` type-checks. It also works at runtime — tag completion appears inside JSP markup in a local build, which is what the screenshots in the description show.

Two of the four were real defects that I had missed, so the review was worth it. The remaining `<script>`-block limitation is documented under Notes rather than fixed, since Monarch cannot re-enter host rules mid-embed.

taehwanis:
Closing in favor of #10858, which now carries this work.

@hwantage folded the HTML language service wiring and the tokenizer fixes from this PR into #10858 with co-author credit, so the two had converged on the same five files. Keeping both open only splits reviewer attention, and #10858 is the more complete of the two — it also carries the newer scriptlet-terminator and Java type-list fixes.

I've opened a much smaller `.jsp`/`.jspf` → `html` fallback as a separate first step in #12402. It is markup-only and explicitly defers to #10858; context there.

PR #11314, opened 2026-07-29
perf(vault): dedupe scope paths by key instead of rescanning

nwparker:
Follow-up to #11303 and #11311.

#11303 took the two session maps off the workspace-switch path. Scope path derivation still follows the active worktree, and was still quadratic — so it remains on the click path at full cost.

## Cause

`addAiVaultWorkspaceScopePath` (`ai-vault-scope-paths.ts:98`) deduped by re-normalizing **every already-accepted path on every insert**:

```ts
const comparisonPath = normalizeRuntimePathForComparison(trimmedPath)
if (paths.some((existingPath) => normalizeRuntimePathForComparison(existingPath) === comparisonPath)) {
  return
}
paths.push(trimmedPath)
```

Accepting K paths costs O(K²) `normalize('NFC')` calls — ~632k at 1124 workspaces. `normalizeRuntimePathForComparison` is the same ICU-backed primitive that made #10841 visible.

Separately, `isAiVaultWorkspaceScopePathClaimed` rescanned every live worktree once per prior id, and runs **twice per switch** (via `activeWorktreePaths` and `scopePaths` in `AiVaultPanel.tsx:114,133`).

## Fix

- Carry a `Set` of comparison keys alongside the paths — one normalize + one lookup per insert.
- Thread that accumulator from the workspace pass into the project pass, instead of restarting deduplication against a plain array.
- Build one `comparison path → worktree id` map for the claim check, keeping first-writer-wins to match the previous `some()` short-circuit order.

## Results

`deriveAiVaultScopeSessionPaths` on a real 1124-workspace profile: **189.7ms → 0.8ms**.

## Correctness

Output is unchanged **including ordering** — verified against the previous implementation across **117 scenario/option combinations** over 9 datasets:

- monorepo (all one repo) and real mixed-repo layouts, from a real 1124-workspace profile
- priors both claimed and unclaimed by other live workspaces, self-owned priors, cross-repo prior ids, bogus ids
- duplicate paths, trailing slashes, doubled separators
- NFD/NFC (Hangul and accented), WSL UNC
- relative and blank paths, empty worktree list, null active worktree
- 3 host-setup projections × 4 project-key shapes

## Tests

This module had **no test file**. Adds one covering the scope semantics that were previously unpinned — prior paths, priors claimed by a live workspace, cross-repo rejection, separator/trailing-slash dedupe, NFD/NFC folding — plus a timing guard.

The guard is verified fail-first: **220ms** on the previous implementation, ~1ms here.

One detail worth flagging: path *length* matters as much as count, because `normalize()` cost scales with it. My first version of the guard used short synthetic paths (`/repo/w0`) and **passed on the broken code at 77ms** — a false guard. It now uses ~50-character paths matching the real profile.

## Verification

- 4938 tests pass across `right-sidebar/` and `shared/`
- `tsc --noEmit -p config/tsconfig.tc.web.json` clean
- `oxlint` clean

## Remaining, not in this PR

`ai-vault-resume-target.ts:134,161` and `ai-vault-session-resume.ts:169` do `Object.values(worktreesByRepo).flat().find(...)` per visible row (~3ms/render at this scale); `getIndexedWorktreeMap` already exists for exactly this. And `checks-panel-terminal-worktree.ts:97` still normalizes inside a sort comparator — same shape #11303 fixed, different panel.

### DIFF
--- a/src/renderer/src/components/right-sidebar/ai-vault-scope-paths.test.ts
+++ b/src/renderer/src/components/right-sidebar/ai-vault-scope-paths.test.ts
@@ -1,0 +1,143 @@
+import { describe, expect, it } from 'vitest'
+import type { Worktree } from '../../../../shared/types'
+import {
+  deriveAiVaultScopeSessionPaths,
+  deriveAiVaultWorkspaceScopePaths
+} from './ai-vault-scope-paths'
+
+function makeWorktree(overrides: Partial<Worktree> = {}): Worktree {
+  return {
+    id: 'repo-1::/repo/orca',
+    repoId: 'repo-1',
+    displayName: 'orca',
+    path: '/repo/orca',
+    head: 'abc123',
+    branch: 'main',
+    isBare: false,
+    comment: '',
+    linkedIssue: null,
+    linkedPR: null,
+    linkedLinearIssue: null,
+    isArchived: false,
+    isUnread: false,
+    isPinned: false,
+    sortOrder: 0,
+    lastActivityAt: 1,
+    isMainWorktree: false,
+    ...overrides
+  }
+}
+
+describe('deriveAiVaultWorkspaceScopePaths', () => {
+  it('returns the active workspace path', () => {
+    const active = makeWorktree()
+    expect(deriveAiVaultWorkspaceScopePaths(active, [active])).toEqual(['/repo/orca'])
+  })
+
+  it('returns nothing without an active workspace', () => {
+    expect(deriveAiVaultWorkspaceScopePaths(null, [makeWorktree()])).toEqual([])
+  })
+
+  it('includes prior paths so renamed workspaces keep their transcripts', () => {
+    const active = makeWorktree({
+      id: 'repo-1::/repo/orca-renamed',
+      path: '/repo/orca-renamed',
+      priorWorktreeIds: ['repo-1::/repo/orca']
+    })
+
+    expect(deriveAiVaultWorkspaceScopePaths(active, [active])).toEqual([
+      '/repo/orca-renamed',
+      '/repo/orca'
+    ])
+  })
+
+  it('drops a prior path another live workspace now owns', () => {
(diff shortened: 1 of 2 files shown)


nwparker:
Both review comments were valid — fixed in 624fc126a5.

**1. First-writer-wins ordering (P2) — real bug, not just theoretical.**

I reproduced it directly:

```
live = [active(path=/x), claimant(path=/x)], prior -> /x
old some(): claimed = true
new map:    claimed = false   <- diverges
```

Greptile framed it as reachable only when two live worktrees share a path, which git prevents. But the trigger is narrower and doesn't need two *distinct* worktrees to collide: it fires when the **active worktree itself** is listed at its own prior path. My original 117-combination equivalence sweep missed it because none of those datasets had the active worktree occupying the prior path.

Fixed by excluding the active worktree while *building* the set rather than comparing ids at read time — any surviving entry is then a claim by construction, so ordering can't decide the result. That also removes the ordering dependency from the interface entirely, which was the reviewer's underlying point.

Worth noting: my first attempt at a regression test for this **passed on the buggy code** (wrong shape — the claimant differed from the active path, so the map never picked the active worktree as owner). The committed test uses the active worktree listed at the prior path across four orderings, and is verified fail-first against d48c7d1530:

```
AssertionError: expected [ '/repo/orca-renamed', '/repo/orca' ] to deeply equal [ '/repo/orca-renamed' ]
```

**2. Missing timing guard on `deriveAiVaultWorkspaceScopePaths` (P2) — correct.**

Added, at the same 1200-worktree scale, with 25 priors so the claim check (the other per-call scan in that function) is exercised too.

**Re-verified after the fix:** equivalence extended to **156** scenario/option combinations across 12 datasets, including the ordering cases — identical paths and ordering. 4940 tests pass, tsc and oxlint clean, and `deriveAiVaultScopeSessionPaths` still measures 0.8ms (from 189.7ms).

On the CodeRabbit description-template warning: this is an internal perf change with no visual surface, so screenshots don't apply.

PR #11330, opened 2026-07-29
fix(remote-runtime): fit retries within recovery budget

### LINKED ISSUE #11305: [Bug]: Remote PTY recovery budget fits only 3 attempts — backoff schedule (60.75s) overruns its own 60s deadline, last 4 steps unreachable
**Operating system:** macOS
**Orca version:** 1.4.159

## Short summary

The remote-runtime PTY auto-recovery budget is mis-sized against its own retry schedule. Two constants in the same file disagree, and a third one (the per-attempt RPC timeout) makes the gap much worse in practice: the recovery window can only fit **3 attempts**, not the 8 the backoff schedule describes. The last four backoff steps are unreachable code.

## The arithmetic

`src/renderer/src/components/terminal-pane/remote-runtime-pty-recovery-state.ts`:

```ts
const RECOVERY_DELAYS_MS = [250, 500, 1000, 2000, 4000, 8000, 15_000, 30_000]  // sums to 60_750
export const REMOTE_RUNTIME_AUTO_RECOVERY_TIMEOUT_MS = 60_000                  // hard wall-clock deadline
```

`begin()` arms the deadline once; when it fires, `phase = 'disconnected'` unconditionally, regardless of what the retry loop is doing.

Even in the best case — every attempt failing instantly — the schedule sums to 60_750 ms and the 8th retry (at 60_750 ms) can never fire, because the deadline lands at 60_000 ms. So the schedule is already 750 ms over budget by construction.

In practice attempts do not fail instantly. Each reconnect attempt goes through a forced `status.get` compatibility probe, whose default timeout is:

```ts
// src/main/ipc/runtime-environment-transport-routing.ts
const DEFAULT_REMOTE_RUNTIME_TIMEOUT_MS = 15_000
```

When the host is unreachable in a way that drops packets silently (VPN/NAT/firewall — i.e. exactly the half-open case `remote-runtime-socket-liveness.ts` was written for), each attempt burns the full 15 s. The real timeline becomes:

| attempt | starts at | fails at |
|---|---|---|
| 1 | 0 s | 15 s |
| 2 | 15.25 s | 30.25 s |
| 3 | 30.75 s | 45.75 s |
| 4 | 46.75 s | 61.75 s → **cut off by the 60 s deadline** |

**Three completed attempts.** Backoff steps `[4000, 8000, 15000, 30000]` are never reached under this condition.

Note the contrast: when the host is merely restarting (port closed), attempts fail instantly with `ECONNREFUSED` and recovery succeeds within ~16 s. It is specifically the silent-drop case — the one the liveness monitor exists to detect — where the budget runs out.

## User-visible effect

The banner says *"Orca will retry for up to one minute."* For a silent network drop it effectively stops trying after ~46 s of wall clock (3 attempts), then sits idle until the deadline flips it to `Remote runtime disconnected`. Pressing **Reconnect** is the only way forward, and users read the preceding minute as "it kept retrying and failed", which is not what happened.

## Existing test does not cover it

`remote-runtime-pty-recovery-state.test.ts` has:

```ts
it('stops automatic recovery after the bounded recovery window', async () => {
  // ...
  await vi.advanceTimersByTimeAsync(60_000)
  expect(state.currentPhase).toBe('disconnected')
})
```

It asserts the deadline fires, which it does. It does not assert how many attempts fit inside the window, so the schedule/deadline mismatch is invisible to it.

## Suggested fix

Pick one so the two numbers agree:

- Raise `REMOTE_RUNTIME_AUTO_RECOVERY_TIMEOUT_MS` above the schedule sum (e.g. 90 s), or
- Trim the schedule so it fits (e.g. drop the trailing `30_000`), or
- Make the deadline attempt-aware rather than pure wall-clock, so a long per-attempt timeout does not silently eat the budget.

Separately, consider giving the reconnect probe a shorter timeout than the general-purpose 15 s. A liveness-style probe wants to fail fast and retry; 15 s per attempt is nearly a quarter of the entire recovery window.

A regression test asserting "N attempts occur within the recovery window" would pin whichever choice is made.

OnlyYu1996:
## Summary

- Keep remote PTY automatic recovery within its one-minute wall-clock budget by capping retry backoff at four seconds.
- Apply a five-second timeout to reconnect inventory, pane resolution, and terminal multiplexer probes while preserving the 15-second initial connection timeout.
- Add regression coverage for eight completed timed-out attempts and recovery-specific timeout propagation.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` (focused recovery tests run below)
- [ ] `pnpm build` (Electron Vite build run below; full native build not run)

Additional focused verification:

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/remote-runtime-pty-recovery-state.test.ts src/renderer/src/components/terminal-pane/remote-runtime-pty-transport.test.ts` (98 tests passed)
- `pnpm run build:electron-vite`
- `pnpm run check:code-quality:changed`
- `pnpm run check:reliability-gates`

## AI Review Report

Reviewed the retry-budget arithmetic, timer cancellation and stale-epoch handling, every terminal reconnect path, and default-versus-recovery timeout propagation. The main regression risk was shortening ordinary first connections; the implementation instead threads an optional recovery-only timeout and tests that initial connections remain at 15 seconds while recovery uses five seconds. Cross-platform review confirmed that no shortcuts, labels, paths, shell behavior, or platform-specific Electron APIs changed; the timer and IPC behavior is equivalent on macOS, Linux, and Windows, including SSH-hosted runtimes.

## Security Audit

Reviewed the existing IPC boundary and timeout propagation along with input handling, command execution, path handling, authentication, secrets, and dependencies. The change only passes bounded internal numeric timeout values through existing typed IPC calls; it adds no user-controlled input, commands, file access, credentials, or dependencies. No follow-up is required.

## Notes

Fixes #11305.

### DIFF
--- a/src/renderer/src/components/terminal-pane/remote-runtime-pty-recovery-state.test.ts
+++ b/src/renderer/src/components/terminal-pane/remote-runtime-pty-recovery-state.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import {
   REMOTE_RUNTIME_AUTO_RECOVERY_TIMEOUT_MS,
+  REMOTE_RUNTIME_RECOVERY_ATTEMPT_TIMEOUT_MS,
   RemoteRuntimePtyRecoveryState,
   retryAllRemoteRuntimePtyRecoveriesNow
 } from './remote-runtime-pty-recovery-state'
@@ -69,6 +70,26 @@ describe('RemoteRuntimePtyRecoveryState', () => {
     expect(state.isActive).toBe(false)
     expect(state.isCurrent(epoch)).toBe(false)
     expect(onChange).toHaveBeenCalled()
+  })
+
+  it('completes eight timed-out attempts inside the bounded recovery window', async () => {
+    vi.useFakeTimers()
+    const state = new RemoteRuntimePtyRecoveryState()
+    let completedAttempts = 0
+    const runAttempt = (epoch: number): void => {
+      setTimeout(() => {
+        completedAttempts += 1
+        state.schedule(epoch, runAttempt)
+      }, REMOTE_RUNTIME_RECOVERY_ATTEMPT_TIMEOUT_MS)
+    }
+    const epoch = state.begin()
+
+    runAttempt(epoch)
+    await vi.advanceTimersByTimeAsync(REMOTE_RUNTIME_AUTO_RECOVERY_TIMEOUT_MS)
+
+    expect(completedAttempts).toBe(8)
+    expect(state.attemptCount).toBe(8)
+    expect(state.currentPhase).toBe('disconnected')
   })
 
   it('advances a pending backoff immediately via retryNow and the active registry', async () => {
--- a/src/renderer/src/components/terminal-pane/remote-runtime-pty-recovery-state.ts
+++ b/src/renderer/src/components/terminal-pane/remote-runtime-pty-recovery-state.ts
@@ -1,4 +1,6 @@
-const RECOVERY_DELAYS_MS = [250, 500, 1000, 2000, 4000, 8000, 15_000, 30_000] as const
+export const REMOTE_RUNTIME_RECOVERY_ATTEMPT_TIMEOUT_MS = 5_000
+// Why: 4 秒退避上限与 5 秒探测共同保证一分钟内能完成至少 8 次恢复尝试。
+const RECOVERY_DELAYS_MS = [250, 500, 1000, 2000, 4000] as const
 export const REMOTE_RUNTIME_AUTO_RECOVERY_TIMEOUT_MS = 60_000
 
 export type RemoteRuntimePtyRecoveryPhase =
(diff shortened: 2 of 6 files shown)


OnlyYu1996:
Updated in `efa519edf` after review:

- Legacy `terminal.resolvePane` verification now carries one deadline into `session.tabs.list`, passes only the remaining timeout, and remains locally bounded even when the inventory request is deduplicated with a slower caller.
- Each terminal multiplexer subscriber now enforces its own connection deadline. A recovery subscriber times out after 5 seconds without cancelling a default 15-second handshake; when recovery starts first, a default subscriber continues with its remaining budget after the shorter shared attempt times out.
- Added regression coverage for the legacy verification budget and both concurrent subscriber orderings.

Validation: 101 recovery/transport tests, 33 existing multiplexer data/resync/backpressure tests, full typecheck, full lint, changed-code quality (0 findings), max-lines, reliability gates, Oxfmt, diff check, and Electron Vite build all pass. Rebased onto current `upstream/main` (`4543bb682`).

PR #11331, opened 2026-07-29
fix(automations): tolerate scheduler tick latency before applying missed-run grace

### LINKED ISSUE #11299: [Bug]: Automations with "No grace" (missedRunGraceMinutes: 0) are always skipped — the scheduler tick lands 40–120s after the scheduled instant
## Summary

A scheduled automation configured with grace = "No grace" never runs. Every scheduled
occurrence is recorded as `skipped_missed` with the message
`Orca was unavailable during the missed-run grace window.` — while Orca is running in the
foreground and the machine is awake the whole time.

## Root cause

`AutomationService.evaluateAutomation`:

```js
const graceMs = automation.missedRunGraceMinutes * 60 * 1e3;
if (now - scheduledFor > graceMs) {
  this.store.updateAutomationRun({
    runId: run.id,
    status: "skipped_missed",
    workspaceId: automation.workspaceId,
    error: "Orca was unavailable during the missed-run grace window."
  });
  this.store.advanceAutomationNextRun(automation.id, now);
  return;
}
```

The evaluation tick never fires exactly at `scheduledFor`. On an idle, running app the observed
lag is **39–116 seconds**. With `graceMs === 0` the condition `now - scheduledFor > 0` is
therefore always true, and the automation can never dispatch.

The grace window is intended to cover *app downtime* (laptop asleep, Orca closed), but it is
currently measured against the scheduler's own tick latency as well.

## Evidence

Automation on a 5-minute schedule, Orca running and machine awake throughout. Values read from
`automationRuns` in `orca-data.json`:

| status | scheduledFor | run created | delta | trigger |
|---|---|---|---|---|
| skipped_missed | 13:10:00.000 | 13:10:45.376 | 45.4s | scheduled |
| skipped_missed | 13:15:00.000 | 13:15:45.377 | 45.4s | scheduled |
| skipped_missed | 13:20:00.000 | 13:20:45.381 | 45.4s | scheduled |
| **completed** | 13:21:16.806 | 13:21:16.821 | **0.015s** | **manual** |
| skipped_missed | 13:25:00.000 | 13:25:45.388 | 45.4s | scheduled |
| skipped_missed | 13:45:00.000 | 13:46:56.137 | 116.1s | scheduled |
| skipped_missed | 13:50:00.000 | 13:51:10.479 | 70.5s | scheduled |
| skipped_missed | 13:55:00.000 | 13:55:39.854 | 39.9s | scheduled |
| skipped_missed | 14:00:00.000 | 14:00:39.859 | 39.9s | scheduled |
| skipped_missed | 14:05:00.000 | 14:05:39.869 | 39.9s | scheduled |

Every `scheduled` run was skipped. The only run that completed was triggered manually, where the
delta is 15ms.

## Reproduction

1. Create an automation on a short schedule (e.g. every 5 minutes).
2. Set **Grace period** to **No grace**.
3. Leave Orca running and the machine awake.
4. Every run is recorded as `Skipped` with "Orca was unavailable during the missed-run grace window."

## Why this matters

- **"No grace" is a selectable option that silently disables the automation.** Nothing in the UI
  indicates that choosing it prevents every scheduled run.
- **There is no correct option for "skip missed runs, just run the next one."** The next value
  after "No grace" is 30 minutes. A user who wants missed runs dropped rather than replayed
  picks "No grace" and ends up with an automation that never fires.
- **The error message points at the wrong thing.** It blames Orca availability, so users
  investigate sleep, lock screen, and app state instead of the grace setting.

## Suggested fix

Any one of these would resolve it:

1. Measure the grace window against actual app downtime rather than tick latency. For example,
   dispatch when the occurrence is the most recent one and the app has been running continuously
   since it, regardless of tick lag.
2. Add the scheduler's tick interval to the comparison: `now - scheduledFor > graceMs + tickIntervalMs`.
3. If strict `grace = 0` semantics are intended, remove it from the picker or relabel it, and add
   a small non-zero option (1–5 minutes) for the "drop missed runs" case.
4. Separate the two failure modes in the error message: scheduler latency vs. actual downtime.

## Environment

- Orca 1.4.159
- macOS 26.3.1 (arm64)
- Schedule: custom cron, every 5 minutes
- Workspace mode: Worktree, Session: Fresh

Pr1p:
## Summary

The missed-run grace window is meant to cover app downtime (laptop asleep, Orca closed), but it was being measured against the scheduler's own tick latency as well. With **\No grace\** (\missedRunGraceMinutes: 0\), the condition \
ow - scheduledFor > 0\ was always true because the evaluation tick fires up to \	ickMs\ (default 60 s) after the scheduled instant, so the automation could never dispatch — every occurrence was recorded as \skipped_missed\.

## Fix

Subtract one \	ickMs\ interval from the measured delay before comparing to the grace window:

\\\diff
- if (now - scheduledFor > graceMs) {
+ const lateness = now - scheduledFor - this.tickMs
+ if (lateness > graceMs) {
\\\

This separates two independent concepts that were conflated:

- **Tick latency** (inherent to the scheduler, up to one \	ickMs\): always tolerated.
- **App downtime** (what grace covers): compared against \graceMs\ after tick latency is excluded.

With grace = 0, the automation runs as long as the tick catches it within one interval (~60 s). If the app was genuinely down past \graceMs + tickMs\, the run is still correctly skipped.

## Testing

- \pnpm exec vitest run --config config/vitest.config.ts src/main/automations/service.test.ts\ — 14 passed (12 existing + 2 new).
- Verified the new positive test **fails without the fix** (reverted \service.ts\, confirmed \send\ was never called) and **passes with the fix**.
- \pnpm run typecheck:node\ — clean.
- Native + type-aware oxlint on changed files — clean.

New tests:

1. **Dispatches a zero-grace automation when the tick lands within one tick interval** — 45 s late, within the 60 s tick: dispatches normally.
2. **Skips a zero-grace automation when the delay exceeds one tick interval** — 90 s late, past one tick: correctly \skipped_missed\.

## Security audit

No new network calls, credential access, or IPC surface. The change only adjusts a timing comparison inside the existing scheduler loop.

Closes #11299

### DIFF
--- a/src/main/automations/service.test.ts
+++ b/src/main/automations/service.test.ts
@@ -57,6 +57,78 @@ describe('AutomationService', () => {
     rmSync(testState.dir, { recursive: true, force: true })
   })
 
+  it('dispatches a zero-grace automation when the tick lands within one tick interval', async () => {
+    vi.setSystemTime(new Date('2026-05-13T08:59:00'))
+    const store = await createStore()
+    store.addRepo(makeRepo())
+    const automation = store.createAutomation({
+      name: 'Zero-grace check',
+      prompt: 'Check the repo',
+      agentId: 'claude',
+      projectId: 'r1',
+      workspaceMode: 'existing',
+      workspaceId: 'wt1',
+      timezone: 'UTC',
+      rrule: 'FREQ=DAILY;BYHOUR=9;BYMINUTE=0',
+      dtstart: new Date('2026-05-12T00:00:00').getTime(),
+      missedRunGraceMinutes: 0
+    })
+
+    vi.setSystemTime(new Date('2026-05-13T09:00:45'))
+    const send = vi.fn()
+    const service = new AutomationService(store, { tickMs: 60_000 })
+    service.setWebContents({
+      isDestroyed: () => false,
+      send
+    } as never)
+
+    service.start()
+    service.setRendererReady()
+    await vi.waitFor(() =>
+      expect(send).toHaveBeenCalledWith('automations:dispatchRequested', expect.any(Object))
+    )
+    service.stop()
+
+    expect(store.listAutomationRuns(automation.id)[0]?.status).toBe('dispatching')
+  })
+
+  it('skips a zero-grace automation when the delay exceeds one tick interval', async () => {
+    vi.setSystemTime(new Date('2026-05-13T08:59:00'))
+    const store = await createStore()
+    store.addRepo(makeRepo())
+    const automation = store.createAutomation({
+      name: 'Zero-grace skip',
+      prompt: 'Check the repo',
+      agentId: 'claude',
+      projectId: 'r1',
+      workspaceMode: 'existing',
+      workspaceId: 'wt1',
+      timezone: 'UTC',
+      rrule: 'FREQ=DAILY;BYHOUR=9;BYMINUTE=0',
+      dtstart: new Date('2026-05-12T00:00:00').getTime(),
+      missedRunGraceMinutes: 0
+    })
+
+    vi.setSystemTime(new Date('2026-05-13T09:01:30'))
+    const send = vi.fn()
+    const service = new AutomationService(store, { tickMs: 60_000 })
+    service.setWebContents({
+      isDestroyed: () => false,
+      send
+    } as never)
+
+    service.start()
+    service.setRendererReady()
+    await vi.waitFor(() => {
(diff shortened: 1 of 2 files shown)


Pr1p:
Superseded by #20819, which landed the scheduler tick-latency fix with coverage for sleep, restart, and grace-boundary cases. Closing this older implementation to avoid retaining the narrower behavior.

PR #11380, opened 2026-07-29
fix(source-control): preserve Windows command override paths

### LINKED ISSUE #11375: Windows: command overrides with native absolute paths are corrupted by POSIX backslash tokenization
## Summary

Custom agent command overrides are parsed by `tokenizeCustomCommandTemplate()` (`src/shared/commit-message-prompt.ts`), which applies POSIX shell backslash-escape semantics on every platform. On Windows this silently destroys native absolute paths, since `\` is the path separator rather than an escape character.

## Reproduction

Enter a normal Windows absolute path as an agent command override in Settings:

```text
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
```

It is tokenized as:

```text
C:WindowsSystem32WindowsPowerShellv1.0powershell.exe
```

and subsequently reported as not found on PATH.

Reported by @iFwu: [https://github.com/stablyai/orca/pull/10190#issuecomment-5117267853](<https://github.com/stablyai/orca/pull/10190#issuecomment-5117267853>)

## Cause

In `tokenizeCustomCommandTemplate`, the unquoted branch treats `\` as an escape and consumes the following character:

```ts
if (ch === '\\' && i + 1 < template.length) {
  current += template[i + 1]
  inToken = true
  i += 2
  continue
}
```

The same treatment applies inside double quotes. The tokenizer is otherwise a reasonable POSIX-style splitter; the problem is that it is platform-unaware.

## Workaround

Replace backslashes with forward slashes in the override. This works but is not discoverable, and Windows Settings should accept paths in the form users actually copy from Explorer or `where.exe`.

## Notes on a fix

Backslash handling needs to become platform-aware rather than being removed outright — quoting behaviour is shared with POSIX hosts, and the same override string can target a WSL distro or a remote SSH host, where POSIX semantics remain correct. The platform that matters is the one the command will execute on, not necessarily the one running the UI. Test coverage should include native Windows absolute paths, quoted paths containing spaces, and existing POSIX escape cases.

## Impact

Currently blocks the only known working workaround for stablyai/orca#11374, since that workaround requires entering an absolute `powershell.exe` path as an override.

## Scope note

Deliberately split out of stablyai/orca#10190, which is scoped to Cursor CLI detection. This bug is independent of Cursor and affects any Windows command override.

bbingz:
## Summary

- Tokenize custom commands using the execution target path flavor.
- Preserve native Windows backslashes while retaining POSIX escaping for macOS, Linux, WSL, and POSIX SSH targets.
- Propagate SSH host path flavor through commit, PR, branch-name, and model-discovery planning.

Fixes #11375

## Screenshots

No visual change.

## Testing

- [ ] pnpm lint
- [x] pnpm typecheck
- [ ] pnpm test
- [ ] pnpm build

Focused checks:

- 304 tests passed, 1 skipped across command parsing, planning, generation, IPC, runtime, branch rename, and renderer plan coverage.
- pnpm run check:code-quality:changed
- pnpm run check:max-lines-ratchet
- git diff --check

## AI Review Report

Reviewed command tokenization, all local and SSH generation callers, WSL isolation, and model discovery. Greptile rated the change 5/5 with no files needing attention. CodeRabbit flagged one verbose comment; commit 8a55e3ba8 condensed it. Cross-platform review verified that macOS, Linux, WSL, and POSIX SSH keep existing backslash escapes while native and SSH Windows preserve path separators.

## Security Audit

The tokenizer remains argv-only and does not add shell execution, variable expansion, command substitution, globbing, or home expansion. The review covered arbitrary command overrides, quoted paths, WSL host separation, SSH host path flavor, and fallback behavior when remote platform detection is unavailable. No secrets or dependency changes are included.

## Notes

The path flavor defaults to POSIX for backward compatibility. Native Windows is selected only for a local win32 target without WSL or an SSH host explicitly detected as Windows.

### DIFF
--- a/src/main/agent-hooks/first-work-generation-target.ts
+++ b/src/main/agent-hooks/first-work-generation-target.ts
@@ -17,6 +17,7 @@ export async function resolveGenerationTarget(
     return {
       kind: 'remote',
       cwd: worktreePath,
+      commandPathFlavor: provider.getHostPlatform?.()?.pathFlavor ?? 'posix',
       execute: (plan, cwd, timeoutMs, operation) =>
         provider.executeCommitMessagePlan(plan, cwd, timeoutMs, operation),
       missingBinaryLocation: 'remote PATH'
--- a/src/main/ipc/filesystem.test.ts
+++ b/src/main/ipc/filesystem.test.ts
@@ -2619,7 +2619,10 @@ describe('registerFilesystemHandlers', () => {
       defaultModelId: 'auto'
     })
     const executeCommitMessagePlan = vi.fn()
-    getSshGitProviderMock.mockReturnValue({ executeCommitMessagePlan })
+    getSshGitProviderMock.mockReturnValue({
+      executeCommitMessagePlan,
+      getHostPlatform: () => ({ pathFlavor: 'windows' })
+    })
     const storeWithOverride = {
       ...store,
       getSettings: () => ({
@@ -2640,7 +2643,8 @@ describe('registerFilesystemHandlers', () => {
       'cursor',
       '/remote/repo',
       expect.any(Function),
-      'npx cursor-agent'
+      'npx cursor-agent',
+      'windows'
     )
     const execute = discoverCommitMessageModelsRemoteMock.mock.calls[0]?.[2] as (
       plan: unknown,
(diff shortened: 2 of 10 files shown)


bbingz:

PR #11389, opened 2026-07-29
fix(mobile): stop a client terminal subscribe from reviving a slept workspace

### LINKED ISSUE #10205: [Bug]: Sleeping workspaces revive without user activation — hard to diagnose (open question on opt-in lifecycle logging)
## Summary

On recent Orca builds, **manually slept workspaces sometimes come back to life without any intentional open/click**. Agents reappear as a fresh shell launch (often with `claude … --resume …` / similar injected into a new or remounted terminal), not as a seamless reconnect to the old PTY.

Initially this looked intermittent; we now have **two reliable-ish triggers** (sleep focused workspace; agent question / agent-row activation). Field diagnosis still lacks a caller/reason log on release builds. This issue is both:

1. a **bug report** for the spontaneous wake behavior, and  
2. an **open question for maintainers/contributors**: what is the preferred way to add **opt-in, deeper lifecycle logging** so field reports (and contributors) can see *what woke a slept workspace* instead of only that a PTY was created again?

## Reliable triggers (updated 2026-07-23)

This is **less intermittent than first thought** once the active-workspace sleep path is isolated:

### Trigger A — Sleep the **focused** workspace (cascade)

1. Manually sleep workspace **A** (e.g. DoomedScripts).
2. Work in workspace **B** so **B is selected**.
3. Sleep **B** (the focused workspace).
4. **A revives within ~1–2s**.

**Control:** sleep a **non-selected** workspace **C** while focused on **B** → **A does not revive**.

Code: only sleeping the active worktree runs `setActiveWorktree(null)` in `sleep-worktree-flow.ts`. Daemon log example (UTC): `session-killed` on active worktree at 17:06:53 → `session-created` on DoomedScripts at 17:06:55.

### Trigger B — Answer an OpenCode / agent question (or click sidebar agent row)

Answering an OpenCode question (click the answer) also woke a previously slept workspace.

Sidebar agent activation explicitly calls `activateAndRevealWorktree(worktreeId)` (`WorktreeCardAgents.tsx`), which always runs `resumeSleepingAgentSessionsForWorktree`. Completed hibernation rows are intentionally inert so they do not resume; **manual sleep does not get the same protection**.

### Unified hypothesis

Any **focus / activation / remount** while other workspaces still hold sleep-resume records can revive them. Manual sleep keeps records + wake-hints but has no durable “stay dead until Wake” gate.

See also the follow-up comment on this issue for full tables and code pointers.

---

## User-visible behavior

**Expected**

- User puts a workspace to sleep (context menu / sleep action).
- PTYs are torn down; workspace stays inactive until the user deliberately opens/wakes it (or an explicitly intended path like mobile open, if that is product-intended).
- No agent relaunch while the workspace stays “hidden/slept” from the user’s point of view.

**Actual**

- After sleeping one or more workspaces (e.g. project `main` worktrees), they later show green/active again **without the user selecting them** to open them.
- Opening them afterward shows the classic “session start injection” pattern in the terminal (startup command typed into a shell), sometimes with `--resume <session-id>`, sometimes a plain agent launch.
- Happens across **multiple projects** (observed on more than one local repo in the same session).
- Creating a new worktree in a *different* project was **not** a reliable repro trigger in one attempt.

## Environment

- **OS:** macOS  
- **Host:** local (not necessarily SSH)  
- **Orca:** recent production/update builds (post–last few updates); also investigated against current `stablyai/orca` `main`  
- **Repro:** see **Reliable triggers** below (sleep-active cascade; agent question click)  
- Mobile may or may not be paired when it happens (not confirmed as required)

### Daemon log shows revive, not reason

`~/Library/Application Support/Orca/logs/daemon.log` records PTY lifecycle (`session-created` / `session-killed` / `session-attached` / `session-exited`) for the slept worktrees.

Example pattern (same day, local paths redacted to project names):

- Sleep-like: `session-killed` for DoomedScripts / foia-scrapers  
- Later, **without intentional open**: `session-created` again for those same worktrees  
- User re-sleeps when they notice them in the active/recent surface  

So we can see **when** something came back, but not **why** (no caller, no “resume sleeping agents” / “cold restore” / “mobile activate” reason).

### Code paths that *can* wake without a sidebar click

Investigation against current `main` (not claiming all of these fired in the field reports, only that they exist):

| Path | What it does |
|---|---|
| `activateAndRevealWorktree` → `resumeSleepingAgentSessionsForWorktree` | Any activation re-launches slept agent sessions |
| `ui:resumeSleepingAgents` (mobile `activateManagedWorktree` with `clientKind: 'mobile'`) | Wakes slept agents **without** navigating the desktop worktree |
| Terminal pane connect + sleeping record → `startFreshColdRestoreAgentResume` | Remount/cold-restore of a tab that still has `sleepingAgentSessionsByPaneKey` re-injects agent launch |
| Startup hydration (`Terminal.tsx`) | Resumes sleeping agents for the restored `activeWorktreeId` |

Sleep intentionally:

- kills live PTYs (`ptyIdsByTabId` cleared)  
- keeps tab identity / wake-hint `tab.ptyId`  
- persists `sleepingAgentSessionsByPaneKey`  
- does **not** currently keep a durable “user explicitly slept this workspace; stay dead until Wake” marker (an earlier `sleptWorktreeIds` approach was added then reverted historically)

So ambient **remount** of a still-tracked slept worktree can look like a spontaneous wake from the user’s POV.

### What existing logs do *not* cover

| Source | Useful for this? |
|---|---|
| `daemon.log` | When PTYs create/kill — **not** wake reason |
| `main.trace.ndjson` | Mostly git/spans — **not** sleep/wake product events |
| Crash / freeze breadcrumbs | Crash-oriented, not a browsable lifecycle audit |
| Privacy diagnostic bundle | Same underlying sources; still no “who resumed sleep” |
| Release Settings “Dev” pane | Dev-only (`import.meta.env.DEV`); not a production probe |

There is **no** shipped toggle like “log session sleep/wake lifecycle with caller reason.”

## Open question for maintainers / contributors

**How should we add opt-in, deeper logging so intermittent sleep/wake bugs are diagnosable in the field?**

We are looking for guidance on the preferred approach before putting up a large PR. Concretely:

1. **Where should the toggle live?**  
   e.g. Settings → Privacy / Diagnostics (alongside diagnostic bundles), an experimental flag, env-only (`ORCA_…_DEBUG=1`), something else?

2. **What should be logged (minimum useful set)?**  
   Candidate events we would want for this bug class:
   - manual sleep start / success / failure  
   - `resumeSleepingAgentSessionsForWorktree` (worktree id, suppressNavigation, record counts/origins)  
   - mobile/main `resumeSleepingAgents` / `ui:resumeSleepingAgents`  
   - activation that triggers resume  
   - cold-restore agent resume on pane connect (`hasSleepingAgentSession`)  
   - background terminal mount of a worktree that still has sleeping records  
   - short stack or stable `reason` enum (not free-form “log everything”)

3. **Where should it land?**  
   Bounded ring in renderer storage? Append-only NDJSON under `userData/logs/` (like `daemon.log`)? Included in diagnostic bundles? Console-only?

4. **Privacy / risk constraints**  
   We assume: default **off**, redacted (no prompts/tokens), size-capped, maybe auto-expire. Are stacks OK if truncated? Any hard no-gos for production opt-in diagnostics?

5. **Product vs diagnostics**  
   Separately, a durable “user slept this” gate that blocks ambient cold-restore / auto-resume until explicit wake may be the right **fix**. We still need logging so residual cases prove which path fired. Prefer fix-first, diagnostics-first, or both in parallel?

If maintainers already have a preferred pattern for product-flow diagnostics (beyond crash breadcrumbs / daemon lifecycle), we’d rather follow that than invent a one-off.

## What we’re *not* asking for (yet)

- Not asking to land a large “customizable developer tools” framework in production.  
- Not claiming a single confirmed root cause for every spontaneous wake.  
- Not asking maintainers to debug from screenshots alone without a plan for better signal.

## Ask

1. **Confirm** whether spontaneous revive of manually slept local workspaces is considered a bug vs intended for some path (e.g. mobile open always wakes agents).  
2. **Advise** on the preferred shape of opt-in lifecycle diagnostics so contributors can land something reviewable and privacy-safe.  
3. If there is already an internal or planned approach for this class of logging, point us at it so we don’t duplicate.

Happy to follow up with a focused diagnostics PR or a sleep-intent fix PR once there’s a green light on approach.

## Related code (investigation notes)

- `src/renderer/src/lib/worktree-activation.ts` — `resumeSleepingAgentSessionsForWorktree` on activate  
- `src/renderer/src/lib/resume-sleeping-agent-session.ts`  
- `src/renderer/src/lib/wake-sleeping-agents-in-background.ts` + `ui:resumeSleepingAgents`  
- `src/renderer/src/components/terminal-pane/pty-connection.ts` — cold restore when sleeping record present  
- `src/renderer/src/components/sidebar/sleep-worktree-flow.ts` — sleep teardown  
- `src/main/runtime/orca-runtime.ts` — mobile `activateManagedWorktree` → `resumeSleepingAgents`  

Thanks!

gatsby74:
## Summary

Sleeping a workspace could be undone by a phone or web client, with no action on the desktop.

Sleep kills a workspace's PTYs but deliberately keeps `tab.ptyId` as a wake hint. `planMobileTerminalTabMount` only checked that the incoming handle resolved to a real tab, so when a client called `terminal.subscribe` on one of that workspace's terminals, the desktop background-mounted the pane and brought a PTY back up — reviving the workspace without any user activation and without navigating anywhere. The user sees a workspace they slept turn green again, and opening it shows a bare shell (a fresh spawn carries no agent resume).

This makes the planner fail closed when a mount could only *create* a PTY rather than attach one: the workspace holds no live PTY, no queued startup, and no pending activation spawn. `terminal.subscribe` already degrades to a scrollback-only stream when no PTY appears, so the client still gets the terminal's saved output — it just no longer resurrects the workspace.

Still allowed, and covered by tests:
- a never-mounted tab in a workspace that has another live PTY (the case this path exists for)
- a freshly created mobile tab whose PTY has not spawned yet (`terminal.create` → `subscribe` races the spawn)
- a tab whose activation spawn is pending

Fixes #10205

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Six new cases in `mobile-terminal-tab-mount.test.ts` (12/12 pass): three that a slept workspace is not revived via real-tab id, synthetic pty handle, or before consulting mount state; three that legitimate mounts still proceed.

Note on `pnpm test`: the full suite has 43 pre-existing failures in 9 files unrelated to this change (`use-markup-draw-hint`, `pr-comment-presentation`, `pr-comments-list-selection`, `use-persisted-ai-vault-view-options`, `LinearAgentSkillSetupPrompt.reminder-toast`, `mobile-sidebar-onboarding-badge`, `SidebarToolbar`, `transcript-watch-liveness`, `agent-exec-handler`). Verified identical failures on unmodified `upstream/main` at 4543bb682 — mostly `window.localStorage` being undefined in those environments.

### Field verification

Reproduced and then verified fixed against a running dev build over CDP:

- Before: a slept workspace (0 live PTYs) received a mobile-style tab mount and came back to life — live PTYs went 0 → 1 → 2 → 3 across three tabs while the desktop stayed on a different workspace the whole time. `planMobileTerminalTabMount` returned a mount plan for every slept workspace.
- After: slept workspaces return `null`; a workspace with live PTYs still plans its mount.

## AI Review Report

Reviewed with Claude Code across the risk areas below.

- **Cross-platform (macOS/Linux/Windows):** pure state predicate over store records — no paths, shells, keyboard shortcuts, accelerators, or Electron platform APIs touched. No platform-conditional behavior added or changed.
- **SSH / remote:** the guard reads `ptyIdsByTabId`, which is populated for local, SSH, and remote-runtime PTYs alike, so a live remote workspace still mounts. Remote-runtime and SSH pty ids are not special-cased, matching the surrounding code.
- **Agent/integration compatibility:** the pending-startup and pending-activation-spawn escapes keep the `terminal.create` → `subscribe` path working, which is how mobile launches agents. Verified no other caller exists — `planMobileTerminalTabMount` is called only from `useIpcEvents`, which passes the full store state.
- **Performance:** one `Array.every` over a single workspace's tabs on an event that already does more work; no new renders, subscriptions, or IPC.
- **UI quality:** no visual change.
- **Security:** see below.
- **Flagged and addressed:** the first draft gated only on "no live PTY", which would have broken mobile terminal creation; the pending-startup and pending-activation-spawn conditions were added and given dedicated tests.
- **Considered and rejected:** a durable `sleptWorktreeIds` marker. It is the more complete fix (it would also cover the non-mobile wake paths listed in #10205), but the issue notes that approach was added and reverted historically, so it needs a maintainer decision rather than a drive-by change.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or dependency surface. The change is a read-only predicate over existing renderer store state and strictly narrows what a remote client can cause the desktop to do — a client can no longer cause PTY spawns in a workspace the user has slept. IPC shape is unchanged; the message is already validated by the existing planner, and the new check runs before any mount is scheduled. No follow-up needed.

## Notes

- Behavior change for clients: subscribing to a terminal in a slept workspace now yields scrollback-only instead of silently starting the workspace. That is the intended product behavior per the issue's "Expected" section; waking should go through explicit activation (`worktree.activate`).
- This closes the one path I could reproduce deterministically. #10205 lists other candidate wake paths (`activateAndRevealWorktree` → `resumeSleepingAgentSessionsForWorktree`, cold-restore agent resume, startup hydration); I could not reproduce those on current `main`, including the issue's own Trigger A cascade, so they are deliberately out of scope here.
- Two adjacent defects found while investigating, not fixed here to keep this single-topic — happy to file or fix separately:
  1. `hasWorktreeSleepIntent` is only set when sleeping the **active** workspace (`sleep-worktree-flow.ts`), so sleeping a background workspace lets its PTY exits reach `bumpWorktreeActivity` — a workspace registers activity at the moment it is slept.
  2. The startup reconnect fallback in `store/slices/terminals.ts` uses `session.activeWorktreeIdsOnShutdown ?? …tabs.some(t => t.ptyId)`. The primary writer in `lib/workspace-session.ts` explicitly avoids `tab.ptyId` with the comment *"which sleep preserves as a wake hint and would revive slept worktrees as active"*, but the fallback uses exactly that — so a session missing that field (older build, or a crash before the write) revives every slept workspace on next launch.

### DIFF
--- a/src/renderer/src/lib/mobile-terminal-tab-mount.test.ts
+++ b/src/renderer/src/lib/mobile-terminal-tab-mount.test.ts
@@ -2,8 +2,12 @@ import { describe, expect, it, vi } from 'vitest'
 import type { AppState } from '@/store/types'
 import { planMobileTerminalTabMount } from './mobile-terminal-tab-mount'
 
-type PlannerState = Pick<AppState, 'tabsByWorktree' | 'terminalLayoutsByTabId'>
+type PlannerState = Pick<
+  AppState,
+  'tabsByWorktree' | 'terminalLayoutsByTabId' | 'ptyIdsByTabId' | 'pendingStartupByTabId'
+>
 
+/** An awake workspace: every tab holds a live PTY, so a mount attaches rather than spawns. */
 function state(tabCount = 1): PlannerState {
   return {
     tabsByWorktree: {
@@ -12,8 +16,17 @@ function state(tabCount = 1): PlannerState {
         ptyId: `wt@@${index}`
       }))
     } as unknown as AppState['tabsByWorktree'],
-    terminalLayoutsByTabId: {}
+    terminalLayoutsByTabId: {},
+    ptyIdsByTabId: Object.fromEntries(
+      Array.from({ length: tabCount }, (_, index) => [`tab-${index}`, [`wt@@${index}`]])
+    ),
+    pendingStartupByTabId: {}
   }
+}
+
+/** A slept workspace: sleep kills the PTYs but keeps `tab.ptyId` as a wake hint. */
+function sleptState(tabCount = 1): PlannerState {
+  return { ...state(tabCount), ptyIdsByTabId: {} }
 }
 
 describe('planMobileTerminalTabMount', () => {
@@ -60,6 +73,69 @@ describe('planMobileTerminalTabMount', () => {
       )
     ).toBeNull()
     expect(isTabMounted).not.toHaveBeenCalled()
+  })
+
+  it('does not revive a slept workspace for a real-tab subscribe (#10205)', () => {
+    // Why: sleep keeps tab.ptyId as a wake hint, so the handle still resolves;
+    // mounting it would spawn a PTY and un-sleep the workspace on the desktop.
+    expect(
+      planMobileTerminalTabMount(sleptState(), { worktreeId: 'wt', tabId: 'tab-0' })
+    ).toBeNull()
+  })
+
+  it('does not revive a slept workspace for a synthetic pty handle (#10205)', () => {
+    expect(
+      planMobileTerminalTabMount(sleptState(200), { worktreeId: 'wt', ptyId: 'wt@@173' })
+    ).toBeNull()
+  })
+
+  it('fails closed before consulting mount state for a slept workspace', () => {
+    const isTabMounted = vi.fn()
+
+    expect(
+      planMobileTerminalTabMount(
+        sleptState(),
+        { worktreeId: 'wt', tabId: 'tab-0' },
+        { isTabMounted }
+      )
+    ).toBeNull()
+    expect(isTabMounted).not.toHaveBeenCalled()
+  })
+
+  it('still mounts when another tab in the workspace is live', () => {
+    // Why: a never-mounted tab in an awake workspace is the case this path exists
+    // for — only a fully cold workspace must fail closed.
+    const s = sleptState(2)
+    s.ptyIdsByTabId = { 'tab-1': ['wt@@1'] }
+
+    expect(planMobileTerminalTabMount(s, { worktreeId: 'wt', tabId: 'tab-0' })).toEqual({
+      worktreeId: 'wt',
+      tabIds: ['tab-0']
+    })
+  })
+
+  it('still mounts a freshly created mobile tab whose PTY has not spawned yet', () => {
+    // Why: terminal.create -> subscribe races the spawn; the queued startup proves
+    // a PTY is coming, so this is an attach-in-waiting, not a resurrection.
+    const s = sleptState()
+    s.pendingStartupByTabId = { 'tab-0': {} } as unknown as AppState['pendingStartupByTabId']
+
+    expect(planMobileTerminalTabMount(s, { worktreeId: 'wt', tabId: 'tab-0' })).toEqual({
+      worktreeId: 'wt',
+      tabIds: ['tab-0']
+    })
+  })
+
+  it('still mounts a tab whose activation spawn is pending', () => {
+    const s = sleptState()
+    s.tabsByWorktree = {
+      wt: [{ id: 'tab-0', ptyId: 'wt@@0', pendingActivationSpawn: true }]
+    } as unknown as AppState['tabsByWorktree']
+
+    expect(planMobileTerminalTabMount(s, { worktreeId: 'wt', tabId: 'tab-0' })).toEqual({
+      worktreeId: 'wt',
+      tabIds: ['tab-0']
+    })
   })
 
   it('does not schedule hidden layout work for an already-mounted tab', () => {
--- a/src/renderer/src/lib/mobile-terminal-tab-mount.ts
+++ b/src/renderer/src/lib/mobile-terminal-tab-mount.ts
@@ -12,16 +12,9 @@ type MobileTerminalTabMountOptions = {
   isTabMounted?: (tabId: string) => boolean
 }
 
-/**
- * True when mounting this tab could only *create* a PTY rather than attach one:
- * the workspace holds no live PTY and nothing is queued to spawn.
- *
- * Why: sleeping a workspace kills its PTYs but keeps `tab.ptyId` as a wake hint,
- * so a mobile subscribe could mount the pane and bring the workspace back to life
- * on the desktop with no user activation (#10205). `terminal.subscribe` already
- * degrades to a scrollback-only stream when no PTY appears, so failing closed here
- * still serves the saved output — it just stops resurrecting a slept workspace.
- */
+/** Why: sleep keeps `tab.ptyId` as a wake hint, so a mobile subscribe could mount the pane and
+ *  revive a slept workspace (#10205); failing closed is safe — `terminal.subscribe` already
+ *  degrades to a scrollback-only stream when no PTY appears. */
 function worktreeHasNoResumableTerminal(
   state: Pick<AppState, 'tabsByWorktree' | 'ptyIdsByTabId' | 'pendingStartupByTabId'>,
   worktreeId: string
--- a/src/renderer/src/lib/mobile-terminal-tab-mount.ts
+++ b/src/renderer/src/lib/mobile-terminal-tab-mount.ts
@@ -12,13 +12,42 @@ type MobileTerminalTabMountOptions = {
   isTabMounted?: (tabId: string) => boolean
 }
 
+/**
+ * True when mounting this tab could only *create* a PTY rather than attach one:
+ * the workspace holds no live PTY and nothing is queued to spawn.
+ *
+ * Why: sleeping a workspace kills its PTYs but keeps `tab.ptyId` as a wake hint,
+ * so a mobile subscribe could mount the pane and bring the workspace back to life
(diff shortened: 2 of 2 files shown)


mmarabel:
Closing this — it targets a path that is not the cause, and the path it *does* touch has since been fixed upstream.

Two things came out of a deeper investigation of #10205:

1. **The mobile-subscribe path was fixed properly in #12672 (STA-3465).** That change carries activation intent explicitly via `TabActivationIntent` and refuses the automatic reconnect probe from #11542, which is the actual mechanism by which a client could respawn a slept pane. It is a better fix than the fail-closed planner guard here, and it is already in 1.4.177.

2. **The wake I was chasing is not a client path at all.** The daemon log for both captures shows no `client-hello` before the respawn — the spawn comes from the already-connected desktop renderer in the same tick as the activation. The cause is the renderer pane-connect path: a slept workspace's pane stays mounted and its deferred connect falls through to a fresh spawn. Fixed in #13343 with the captured caller.

Thanks for the patience on this one — the original premise was wrong and I would rather withdraw it than have it reviewed on that basis.

PR #11395, opened 2026-07-29
fix(ai-vault): prune remote OpenClaw discovery

### LINKED ISSUE #11394: [Bug]: SSH Agent Session History scans non-session OpenClaw subtrees
### Operating system

macOS

### Orca version

`v1.4.161` (current stable Homebrew cask)

### Details

Short summary:

SSH Agent Session History recursively walks every directory below `~/.openclaw/agents`. The OpenClaw source then accepts any JSONL path containing a `sessions` segment, including Codex homes embedded below an OpenClaw agent directory. On a long-lived multi-agent host this turns one history refresh into thousands of serialized SSH directory operations and considers non-OpenClaw transcripts as OpenClaw candidates.

Observed reproduction:

1. Connect Orca to an SSH host with OpenClaw agent state.
2. Open Agent Session History.
3. Orca recursively traverses non-session subtrees such as `<agent>/agent/codex-home/sessions/**` before the scan can settle.

Sanitized measurements from the affected SSH host:

- `~/.openclaw/agents`: 14,990 directories / approximately 7.0 GB
- canonical `<agent>/sessions/*.jsonl`: 2,264 files / 559 MB
- nested `codex-home/sessions/**/*.jsonl`: 3,106 files / 1.50 GB

No transcript contents, filenames, session IDs, or workspace paths are included in these measurements.

Expected:

OpenClaw discovery should enter each `<agent>/sessions` subtree but should not descend into sibling state such as `<agent>/agent`, `<agent>/codex-home`, plugins, caches, or embedded agent homes.

Code-path evidence on current `main`:

- `remoteOpenClawSources()` roots discovery at `.openclaw/agents` and `.clawdbot/agents`.
- Its file predicate accepts any descendant path containing a `sessions` segment.
- `walkRemoteSessionFiles()` recursively calls `readDir()` for every accepted directory, sequentially within each source.
- Therefore a nested Codex home is traversed and its rollout JSONL files are considered OpenClaw candidates.

Suggested minimal fix:

- At depth 0, keep all OpenClaw agent directories.
- At depth 1, descend only into the canonical `sessions` directory.
- Preserve the existing recursive behavior inside that canonical subtree.
- Add a regression test proving that canonical OpenClaw sessions remain visible while an embedded `agent/codex-home/sessions` subtree is never listed.

Related: #11362 reports the same user-visible `Scanning sessions` symptom but tracks a separate unbounded SSH stream state. This issue tracks excessive discovery work before transcript parsing or streaming.

Also distinct from #7611, which caches unchanged remote parse results across rescans but does not bound first-scan directory discovery.

Turtle-Hwan:
## ELI5

When loading OpenClaw history over SSH, Orca now searches the standard session folders while skipping unrelated caches. This avoids unnecessary directory scans and keeps the supported session history available.

## Details

SSH AI Vault history scanned every subtree under `.openclaw/agents` and `.clawdbot/agents`, including embedded Codex homes and sibling caches. A fixture with 512 cache branches required 1,540 directory listings and exposed unrelated JSONL as OpenClaw history. Discovery now lists three directories for that fixture and returns only its canonical session.

Keep every agent directory, then descend only into its `sessions` subtree. Recursion within that subtree remains intact. The existing source/walker handles paths, cancellation and errors; unreadable canonical directories still report an issue while healthy agents remain visible. This covers current and legacy JSONL archive locations. It does not add support for OpenClaw's newer SQLite transcript store.

Fixes #11394.

Validation against main `ef6eeab26e4a2ba817b19e8d68b37546f58d293e`:

- 1,133 tests passed across 126 AI Vault, IPC and remote-platform test files.
- New scanner tests cover Linux/macOS/Windows remote path layouts, both state roots, multiple agents, nested canonical sessions, sibling pruning and partial read failures. Remote hosts are represented by the existing in-memory filesystem provider; no live Windows/SSH run is claimed.
- Restoring baseline source fails four new regressions. The work-count check observes 1,540 listings on baseline versus 3 with the fix.
- Node typecheck and all three changed-code quality gates pass. Tests ran with `ORCA_BACKGROUND_LAUNCH=1`.

The original contributor history is preserved and current main is merged into this branch. No rendered UI change or Electron launch. The remote provider owns all filesystem reads; this changes no RPC fields or process-status decisions. Existing scope-path handling supports folder workspaces as well as Git worktrees.

## Supported folders and compatibility

Compatibility requires preserving the standard supported JSONL archive folders: `~/.openclaw/agents/<agent>/sessions/**` and legacy `~/.clawdbot/agents/<agent>/sessions/**`, including every agent and nested archives. The implementation and regression tests preserve both.

Manually relocated JSONL archives outside an agent's standard `sessions` subtree are no longer discovered by this scan. Their files are not deleted. Canonical nested session archives remain included; custom archive locations would need explicit support.


AmethystLiang:
looking into it

Turtle-Hwan:
Addressed the review nitpick in 13462cc73.

The fixture placed its only valid transcript directly under `main/sessions`, so a regression that stopped recursing at `sessions` would still have passed. Added one under a date directory (`main/sessions/2026-07/`) and asserted both that the transcript is discovered and that the nested directory is read — the readdir assertion now requires the extra level rather than allowing it.

nwparker:
**Parking this for now — needs more thought, and it is not a priority.**

Reviewed with the full readiness checklist and adversarially verified; **no correctness defect was found**. What is holding it is a deliberate product trade-off, not a bug, so it wants a decision rather than another round of patching.

**The trade-off:** the fix restricts OpenClaw discovery to each agent's canonical `sessions/` subtree. Applied to the remote source only, local and SSH show *different history for the same directory tree*. Applied to both (the current state of the branch), parity is restored — but by **extending the data loss to local**: relocated archives such as `~/.openclaw/agents/main/archive/sessions/old.jsonl` then disappear from local AI Vault and from search too, where before they only disappeared over SSH.

So: which is worse, an asymmetry between local and SSH, or uniform invisibility of relocated transcripts? A third option worth considering is keeping the fast canonical walk but emitting a scan issue when an agent directory containing candidate JSONL is pruned, so rows do not vanish silently.

Relay note for whoever picks this up: the host runs the scan, so a server that updates ahead of its clients silently narrows what every paired client sees, including older ones — no wire change involved, only published content. See `docs/reference/remote-wire-compatibility.md`.

PR #11406, opened 2026-07-29
fix(cli): relativize absolute POSIX paths against UNC worktree roots in WSL

### LINKED ISSUE #11393: [Bug]: file open with an absolute Linux path fails with invalid_relative_path in WSL workspaces (still present in 1.4.161)
### Operating system

Linux

### Orca version

1.4.161

## What happened

In a WSL workspace, `orca-ide file open <absolute-path>` fails with `invalid_relative_path` when the path is given in Linux POSIX form, even though the file is inside the active worktree. Relative paths and Windows-UNC absolute paths work.

This appears to be the same underlying defect as #9429, which was closed as fixed by #9824 (v1.4.156). A commenter on #9429 already reported it still reproduces after 1.4.156, and I can still reproduce it on 1.4.161.

## Reproduction

1. Open a WSL workspace in Orca (the worktree root is stored as a Windows UNC path, e.g. `\\wsl.localhost\Ubuntu\home\<user>\<repo>`).
2. From a shell inside the worktree, run:
   ```
   orca-ide file open /home/<user>/<repo>/package.json --worktree active --json
   ```
3. Observe the error.

## Expected behavior

The file opens. `orca file open --help` states: "The path may be relative to the selected worktree or an absolute path inside that worktree."

## Actual behavior

```json
{
  "ok": false,
  "error": { "code": "runtime_error", "message": "invalid_relative_path" }
}
```

Deterministic - happens every time for POSIX absolute paths.

## Evidence

Same file, same worktree, three path forms:

| Command path argument | Result |
| --- | --- |
| `package.json` (relative) | OK, `relativePath: "package.json"` |
| `\\wsl.localhost\Ubuntu\home\<user>\<repo>\package.json` (Windows UNC absolute) | OK |
| `/home/<user>/<repo>/package.json` (Linux POSIX absolute) | `invalid_relative_path` |

In WSL the worktree root is stored as a Windows UNC path while the user types a Linux POSIX path. The relativize step added in #9824 strips the worktree-root prefix before the runtime RPC, but the POSIX path does not match the UNC root, so the raw absolute path is forwarded and `isSafeMobileRelativePath` rejects it with `invalid_relative_path`. The fix appears not to normalize between POSIX and UNC path forms.

## Regression timing

#9429 was fixed in 1.4.156 for native paths; the WSL POSIX-path case is still broken on 1.4.161.

## Environment

- Orca version: 1.4.161
- OS: Ubuntu 24.04 LTS on WSL2 (Linux 6.18.x, x86_64)
- Execution context: WSL
- Workspace: Git worktree (root stored as a Windows UNC path)

stroland02:
Fixes #11393


brennanb2025:
Thanks for tracking this down — the diagnosis is right, and `resolveFilePath` is exactly where the fix belongs.

Heads up: rather than send you a pile of round-trips, I pushed a few commits straight to your branch. Here's what changed and why.

**The rewrite fired too broadly, and re-introduced the bug it fixes.** It keyed off `WSL_DISTRO_NAME` being set — but that variable is also set for a plain Linux CLI running inside the distro, where worktree roots are POSIX, not UNC. Those roots then stopped matching, so *every* absolute path got stranded with the same `invalid_relative_path` this PR is fixing. Checked against a live Orca runtime, same command and same runtime both times: the branch as submitted returned `invalid_relative_path`, and with the fix it returns `ok`. It's now gated on the worktree root actually being a WSL UNC path (`isWslUncPath` in `src/shared/wsl-paths.ts`). Worth noting the updated repro test converted the POSIX-root case into a UNC case rather than adding one alongside it — keeping both would have caught this.

**A backslash in a Linux filename opened the wrong file, silently.** Backslash is a legal Linux filename character, but it becomes a separator once the path reads as UNC, so `a\b.ts` relativized to `a/b.ts` — no error, just a different file opened. The rewrite now skips any path containing a backslash, which turns that into a clean error.

**Dropped the hard dependency on `WSL_DISTRO_NAME`.** Nothing guarantees that variable survives the WSL→Win32 PowerShell bridge the CLI goes through, and if it doesn't, this would have been a silent no-op in exactly the scenario reported in #11393. The distro is now derived from the invocation cwd — which the launcher always sets, as a UNC path that names the distro — when the env var isn't there.

The remaining commits are tests pinning the negative space: sibling-prefix roots, distro-case folding, drive-letter workspaces, and the regression above.

Two asks:

- Could you move the `.gitignore` commit (15e89ed93d) to its own PR? We try to keep PRs to one change each, and it's unrelated to the path fix — it also happens to be the only thing conflicting with `main` right now, so lifting it out clears that too. Two small notes if you do: it drops the file's trailing newline, and the unanchored `worktrees/` pattern matches a nested `worktrees` directory anywhere in the tree, so `/worktrees/` is safer.
- One gap I saw and deliberately left for a follow-up so you know it isn't lost: repos that live on the Windows drive rather than inside the distro still fail. There the worktree root is stored as `C:\...` and you'd be typing `/mnt/c/...`. `toWindowsWslPath` in `src/shared/wsl-paths.ts` already does that mapping if you want to take it.

One honest gap on my side: none of this ran on real Windows or real WSL. A Windows smoke test of `orca file open <linux-abs-path>` from a WSL workspace would close it.

brennanb2025:
Follow-up pass on this branch, this time only asking "is this the standard shape, or a second implementation of something we already have?" Your diagnosis and the fix's behavior are unchanged — I only swapped the mechanism. Two things, pushed as `9a2cb14fab` and `0de6a114ee`:

**1. `toWindowsWslPath` already exists, so the handler shouldn't hand-roll the UNC template.**

`src/shared/wsl-paths.ts:24` owns exactly this conversion ("absolute Linux path in a known WSL distro → its Windows form") and already has five production callers (`src/main/wsl.ts`, `src/main/hooks.ts`, `src/main/codex-accounts/service.ts`). The handler was building a fourth copy of `` `//wsl.localhost/${distro}${path}` `` inline. Swapped to the helper:

```ts
-  return `//wsl.localhost/${distro}${path}`
+  return toWindowsWslPath(path, distro)
```

Behavior is identical under the `isWslUncPath(rootPath)` guard — I proved it by running **your test file verbatim** (all 19 cases from `b4f39e6338`) against the new source: all green. Two things come free: the helper also maps `/mnt/<drive>` to its drive spelling, and passing `distro` as a real `string` argument turns the null check into a compile error instead of an untested branch.

**2. Removed one of my own test cases that pinned shape rather than behavior.**

`does not alias a distro-named first segment onto the root` was mine, and it was vacuous — I deleted the `!distro` guard and it stayed green, because it asserts the same passthrough as `does not rewrite when the CLI is not running under WSL`. Its comment even referenced the `||` in the handler, which is pinning implementation. With the helper in place the compiler enforces that guard, so the case wasn't earning its keep. The backslash case still kills its mutant (verified), so that one stays.

Everything else in the diff is the standard shape — the guard set, the `WSL_DISTRO_NAME` → UNC-cwd fallback, and the `relativePathInsideRoot` reuse all read the way I'd expect.

---

**Actionable follow-up, out of scope here (different surface, please don't add it to this PR):** `mapTerminalFilePath` at `src/renderer/src/components/terminal-pane/terminal-file-open-routing.ts:53` does this same rewrite for Cmd-clicked terminal file links, and it still has the backslash bug this PR now guards against — clicking `/root/repo/a\b.ts` in a WSL terminal opens `a/b.ts` instead. The standard fix there is the same one-line guard plus `toWindowsWslPath`, but it's terminal link routing rather than `orca file open`, so it wants its own PR.

brennanb2025:
Merged latest `main` into this branch to clear the `.gitignore` conflict — main had renamed the bench-results path and added a new entry, while this branch added `.worktrees/` and `worktrees/`, so I kept both sides. Nothing else changed: `src/cli/handlers/file.ts` and `file-absolute-paths.test.ts` are byte-identical to your previous head. Thanks for the fix!

OrcaWin:
## Windows QA — native app + WSL UNC workspace

**Exact SHA:** `7c0dfec8f8153c6a483fe17d15579e1fb3123adb`

**CDP-confirmed app identity**
- name: `Orca: 7c0dfec8f8`
- isDev: `true`
- devLabel: `pr-11406-native-qa @ 7c0dfec8f8`
- devRepoRoot: `C:\Users\neil\orca\workspaces\orca\pr-11406-native-qa`

**UNC worktree root targeted:** `\\wsl.localhost\Ubuntu\home\orcatest\orca\workspaces\sta3077\pr-11406-wsl-qa3`

**POSIX path passed from inside Ubuntu (`orca-ide file open`):** `/home/orcatest/orca/workspaces/sta3077/pr-11406-wsl-qa3/pr11406-native-qa-marker.txt`

**Correct file opened:** yes. CLI JSON returned `relativePath: "pr11406-native-qa-marker.txt"`, `opened: true`. Editor tab/path is the matching UNC file and the visible body is `PR11406-QA3-MARKER unique=7c0dfec8f8`.

**Backslash guard:** pass. `orca-ide file open --path '/home/orcatest/orca/workspaces/sta3077/pr-11406-wsl-qa3/a\b.ts'` returned `invalid_relative_path` (exit 1). The trap file `a/b.ts` was not opened.

Topology: Electron ran from the native `C:\` checkout above; the workspace under test is the WSL UNC root; the CLI was invoked inside Ubuntu via the managed `orca-ide` bridge (`WSL_DISTRO_NAME=Ubuntu`, `ORCA_CLI_CWD` UNC).

### POSIX open (correct file)



### After backslash path (trap file not opened)

OrcaWin:
## Windows native QA — WSL UNC POSIX `file open` (#11393)

Read-only check of `7c0dfec8f8153c6a483fe17d15579e1fb3123adb` (`7c0dfec8f8`). App ran from the native `C:\` checkout; workspace under test was the UNC WSL root; CLI was invoked from inside Ubuntu.

| Item | Result |
|---|---|
| SHA | `7c0dfec8f8153c6a483fe17d15579e1fb3123adb` |
| App identity | `Orca: 7c0dfec8f8` · `devWorktreeName=pr-11406-native-qa` · `devRepoRoot=C:\Users\neil\orca\workspaces\orca\pr-11406-native-qa` |
| UNC root | `\\wsl.localhost\Ubuntu\home\orcatest\orca\workspaces\sta3077\pr-11406-wsl-qa3` |
| POSIX path | `/home/orcatest/orca/workspaces/sta3077/pr-11406-wsl-qa3/qa-11406-posix-open.txt` |
| Correct file opened | Yes. Editor showed unique marker `QA11406-POSIX-OPEN-20260815T013108Z`. CLI result: `relativePath=qa-11406-posix-open.txt`, `opened=true`. |
| Backslash guard | Held. `.../qa-11406-guard/a\b.ts` returned `invalid_relative_path`. Trap file `qa-11406-guard/a/b.ts` (`SLASH_FILE_MUST_NOT_OPEN`) was not opened. Editor stayed on the POSIX-open file. |

Topology notes: `WSL_DISTRO_NAME` is not forwarded across Windows interop, so the UNC `ORCA_CLI_CWD` fallback named the distro (the designed path). The Windows CLI process used this checkout's built CLI against the live runtime for this instance.

**Windows native WSL UNC QA**

| 01-posix-open-correct-file.png | 02-backslash-guard-did-not-open-slash-file.png |
|---|---|
|  |  |

PR #11417, opened 2026-07-29
feat: track context-window pressure across agent sessions

### LINKED ISSUE #11225: [Feature]: Show context-window pressure and configurable soft limits across agent views
### Problem or use case

Long-running agent sessions can get close to their context-window limit without any clear warning in Orca. By the time the underlying agent starts compacting, loses useful detail, or becomes less reliable, the user may only notice from the agent's own terminal UI — if that UI exposes the information at all.

This is especially hard to monitor when several agents are running across different worktrees. Orca already provides an at-a-glance view of agent state, so context pressure should be visible there as well.

### Proposed solution

Track context-window pressure for each agent session and surface it consistently throughout Orca.

Orca should:

* Use the model or agent harness's reported context-window size and current usage whenever that information is available.
* Allow users to define their own soft context limits, globally and optionally per provider/model. A soft limit may intentionally be lower than the model's actual maximum.
* Support percentage-based thresholds and, where practical, absolute token thresholds.
* Fall back gracefully when an agent does not expose reliable context information, rather than presenting an invented precise value.

The default visual state could use a simple traffic-light model:

* **Green** — comfortably below the configured soft limit.
* **Yellow** — approaching the limit; the session may soon need compaction or a fresh continuation.
* **Red** — at or beyond the warning threshold, or close to exhausting the usable context window.

Hovering or opening the detail view should show the actual values when known: used tokens, effective limit, percentage, and whether the limit comes from the model or from a user-defined soft cap.

### Where it should appear

Ideally the state should be visible at every useful level:

* individual agent/session rows;
* terminal or chat tabs and headers;
* worktree cards, using the worst state among their active agents;
* Agent Dashboard and other aggregate views;
* orchestration and automation views where sessions are monitored indirectly.

It should work consistently across supported agents and modes, including local sessions, SSH/remote runtimes, WSL, and paired clients wherever the underlying data is available.

Thresholds and colors should be configurable so users can choose how early Orca warns them without changing the model's real hard limit.

### Why this is different from stablyai/orca#9225

stablyai/orca#9225 asks Orca to expose conversation compaction as a distinct runtime status. This request is complementary: it provides proactive visibility *before* compaction starts and works across providers rather than only representing the compaction phase itself.

### Acceptance criteria

* Context pressure is tracked from provider/model data when available.
* Users can configure global and per-provider/model soft limits.
* Green/yellow/red state is visible on individual and aggregate agent surfaces.
* Aggregate surfaces reflect the most urgent child-session state.
* Exact usage and effective-limit details are available on demand.
* Missing or uncertain provider data is represented honestly.
* The feature behaves consistently across supported runtimes and agent modes.

raskalbass:
## Summary

- Track provider-reported Claude and Codex context-window usage across local, SSH, and WSL sessions.
- Add configurable global, provider, agent, and model soft limits with honest unknown-state handling.
- Show individual and worst-child pressure across worktree cards, terminal tabs, dashboard and kanban views, automation history, terminal dialogs, AI Vault, and paired mobile clients.
- Include exact usage, effective-limit provenance, accessible non-color states, persistence migration, relay synchronization, and defensive validation.

## Testing

- `pnpm run typecheck`
- `pnpm run lint`
- Targeted Vitest coverage for context-pressure UI, managed hook stdin lifecycle, and mobile terminal reveal adoption.
- `git diff --check`

## AI Review Report

- **Cross-platform:** Local POSIX and Windows statusline guards are covered; remote scripts use POSIX-safe endpoint state.
- **SSH / WSL:** Context-only readings and live feature-toggle state cross the relay without forwarding remote rate-limit data.
- **Integrations:** Claude statusline and Codex rollout ingestion are explicit; unsupported providers remain unknown.
- **Performance:** Readings use structural equality and existing snapshot/change gates to avoid redundant fanout and rerenders.
- **UI:** Existing design tokens and shared tooltip primitives are used with accessible labels and non-color state shapes.
- **Security:** Relay payloads are normalized and feature-gated; provider session identity prevents stale cross-session readings.

Closes stablyai/orca#11225


brennanb2025:
Taking a look!

brennanb2025:
Thanks @raskalbass — this is a well-architected submission and we want to land it. It rides the existing seams end to end (statusline POST, rollout tail, relay envelope, status-row upsert) with no new transports, honest "unknown" fallbacks, and full feature-flag inertness. We reviewed it as a maintainer patch series on top of your head commit; your commit and attribution are preserved untouched.

What we changed on top (maintainer commits):
- De-staled against current main; resolved the one conflict in `ssh-relay-session.ts` (kept main's PTY-consumer recovery helper plus your `bindContextPressureSettings()` hook).
- Effective limit is now provider-first: the curated model table only fills in when the harness reports no max (per #11225 and the table's own contract), so a stale table entry can never inflate pressure. Isolated commit for easy revert if min() was intended.
- Remote statusline hardening: SSH remotes never POST `rate_limits` (the relay drops them by design — account identity stays host-local), a user-owned remote statusLine slot is never claimed, and a deleted managed entry is a respected opt-out (marker on the remote box).
- Session-identity: a session-tagged reading is dropped until the row's provider session id is established (closes a stale-post window before identity lands).
- Runtime/mobile: OSC retention and the worktree.ps cross-source merge never carry a reading across an agentType change, and the merge no longer drops a known reading when a fresher OSC ping wins (no dot flap).
- Level now follows the rounded percent every surface displays ("90%" never wears a warning dot under "critical at 90").
- Soft-limit keys require an explicit scope (`global`, `provider:<id>`, `agent:<type>`, `model:<id>`) with inline editor validation — the unprefixed shorthand was ambiguous between model and agent.
- Mobile copy is plain English constants (mobile has no i18n layer yet); unknown `limitSource` values from a newer desktop degrade gracefully.
- Restored the load-bearing comments that were compressed for the (comment-exempt) max-lines cap; verified both extractions are move-pure; small a11y (aria-invalid), validation-bound, and feature-off-cost cleanups.

Known accepted behaviors (documented, by design):
- After an SSH reconnect an idle remote session shows no reading (never a stale one) until its next statusline tick; readings replay across app restarts from persisted last-status.
- SSH remotes and WSL guests get the managed statusline installed so the toggle propagates live without reconnect; with the feature off the script exits via shell builtins before any subprocess (one `sh` spawn per statusline tick).
- Nested mobile dot semantics are flattened into the row for screen readers (pre-existing pattern with the lineage toggle) — follow-up candidate.

Verification (exact head 9fad190eec38f432b9cb19ec3450b084516d7525 = your commit + 7 maintainer commits, fast-forward):
- 38 affected desktop suites (2,050 tests) + full mobile suite (386 files, 2,873 tests) green; typecheck, oxlint chain, max-lines ratchet, localization catalog/extraction/coverage all pass; changed-code quality gate (including React Doctor) clean vs main: 0 new findings across 108 changed files.
- Two independent tracked review lanes ran three rounds to CLEAN-same-round, plus a dedicated perf audit (no regression; per-tick costs quantified) and red/green sampling on 10+ behaviors (tests fail when each fix or core behavior is reverted).
- Electron validation: PASS. Isolated dev instance (temp profile); readings driven end-to-end through the real local hook-server HTTP boundary (real Claude statusline JSON, incl. session ids), not store injection. 12 full-page screenshots (light + dark) below cover sidebar/tab-strip/dashboard/kanban dots, the settings editor (invalid soft-limit key → inline `role=alert` error + `aria-invalid` ring), feature-off inertness (a11y tree empty of any pressure label), the live endpoint-file toggle (no restart), and folder-workspace (non-git) parity. Not exercised live: automations run-history surface (unit-covered) and the separate popout window (same kanban card component captured in-window; popout payload validation unit-tested).
- SSH (throwaway Linux container): relay-contract legs PASS with physical evidence — real `build:relay` deploy to an Ubuntu 22.04 arm64 container; user-owned remote `statusLine` slot never claimed across three connect attempts (remote settings.json byte-identical after); the exact remote script bytes under real `sh`: flag-off → zero subprocess posts, flag-on → exactly one `context_window` POST, and `rate_limits` never posts from the remote regardless of flag. The desktop-connect leg hit a dev-mode ssh2/Docker-for-Mac wedge that reproduces identically on an identity-verified pinned-main build in the same container (A/B), so it is environmental, not introduced by this PR; the legs not proven live on that rig (relay→desktop envelope crossing, live remote toggle, reconnect-absence) are pinned by the green relay/integration suites on this exact head.
- Mobile emulator: PASS, full matrix against a desktop running this exact head — pairing/session flow, 75% warning (orange diamond, "approaching limit") and 95% critical (red square, "near limit") per-agent rows, worst-of worktree rollup, "Effective limit: provider-reported" a11y labels, press-for-detail alert, and feature-off vanish; mobile rows matched the desktop store exactly (screenshots below).
- Windows .bat checkpoint: PASS (A–F) on real cmd.exe via a tracked worker on a real Windows host — the generated .bat ran under Claude-statusline stdin in both flag states: payload capture, two-probe `findstr` split (`rate_limits` posts in both states; `context_window` only when enabled), throttle-stamp behavior, temp-file cleanup, no stray stdout (statusline visually unchanged), no window flash. One Medium finding — a trailing backslash in `CLAUDE_CONFIG_DIR` mangles the curl argument — is byte-identical on main (pre-existing, out of scope here); we recommend a separate follow-up issue.

<details>
<summary>Electron screenshots (12 — dark + light, feature on/off, folder workspace)</summary>














</details>

<details>
<summary>Mobile screenshots (4 — rollup, per-agent rows, detail alert, feature off)</summary>






</details>

One open product question for the maintainers before merge: provider-reported max vs. curated table precedence (we shipped provider-first per the written contracts; commit 349c4c1218 reverts cleanly if min() was the intent).

PR #10647, opened 2026-07-26
Stop relaunching creation-time agents on workspace activation

AmethystLiang:
## Summary
- Remove the fallback that relaunched `createdWithAgent` when activating an empty worktree.
- Keep explicit startup payloads working for callers that intentionally queue a command.
- Retain `createdWithAgent` metadata for removal safety, ownership, and preflight instead of reopen behavior.
- Add activation coverage for repeated close/reopen cycles, post-delete focus handoff, host-originated activation, and explicit startup preservation.
- Move WSL single-quote launch quoting coverage to direct agent launch tests.

[fixes 10578](https://github.com/stablyai/orca/issues/10578)

PR #11422, opened 2026-07-29
perf(terminal): serialize checkpoints with one payload walk

nwparker:
## Description
Serialize terminal checkpoints with a bounded JSON writer that accounts for exact UTF-8 bytes while emitting output. Passing candidates avoid a full-object `JSON.stringify` followed by `Buffer.byteLength`; over-cap JSON emission stops once the byte limit is exceeded and discards its bounded partial output. If the initial checkpoint is rejected, the existing fallback deliberately replays the full checkpoint so it can produce and size a trimmed candidate.

Warmed 100 MiB measurements were **253.7–330.2 ms before** and **247.4–253.6 ms after**. This is not presented as a dramatic speedup; the change primarily bounds JSON emission and removes the extra full-string byte scan for fitting candidates.

## ELI5
Orca now stops building checkpoint JSON when it reaches the size limit instead of building an oversized JSON string. If that first attempt is too large, Orca still reads the full terminal history on purpose so it can create a smaller checkpoint that fits.

### What changed
- Replaced the false-positive total-code-unit assertion with a direct proof that an oversized checkpoint object is never handed to full-object `JSON.stringify`.
- Kept the exactness-parity tests and passing-path test unchanged.
- Clarified that JSON emission is bounded, while the rejection fallback intentionally performs a full checkpoint replay.
- Replaced implied dramatic performance claims with the measured warmed 100 MiB ranges: 253.7–330.2 ms before and 247.4–253.6 ms after.

### ELI5
The size-limited JSON builder stops once its bucket is full, so it never creates the giant JSON result. Orca may still reread the terminal data afterward because that replay is how it makes a smaller checkpoint worth saving.

### Proof of fix
RED phase: archived pre-fix production (`d0ce3c7ec3`) with the new test copied in.

```bash
temp_dir=$(mktemp -d /tmp/orca-pr11422-red.XXXXXX) && git -C /Users/nwparker/orca/workspaces/orca/prod-scan-07-29 archive d0ce3c7ec3 | tar -x -C "$temp_dir" && cp /Users/nwparker/orca/workspaces/orca/prod-scan-07-29/src/main/daemon/terminal-checkpoint-serializer.test.ts "$temp_dir/src/main/daemon/terminal-checkpoint-serializer.test.ts" && ln -s /Users/nwparker/orca/workspaces/orca/prod-scan-07-29/node_modules "$temp_dir/node_modules" && printf 'RED_TEMP=%s\n' "$temp_dir" && cd "$temp_dir" && pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/terminal-checkpoint-serializer.test.ts -t 'rejects an oversized escaped candidate without materializing it'
```

Result: **1 test file failed; 1 test failed, 2 skipped**. The intended assertion failed because pre-fix production passed the oversized checkpoint object to `JSON.stringify` (`expected true to be false`).

GREEN phase on pushed head `88ca338811`:

```bash
pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/terminal-checkpoint-serializer.test.ts src/main/daemon/json-utf8-byte-length.test.ts src/main/daemon/terminal-checkpoint-writer-bounds.test.ts
```

Result: **3 test files passed; 9 tests passed**.

### Proof of no regression
The green command covers the serializer, exact UTF-8 byte estimator, and checkpoint writer fallback suites. The existing exactness-parity tests and passing-path test remain intact.

```bash
pnpm run typecheck
```

Result: **passed** across `config/tsconfig.node.json`, `config/tsconfig.tc.cli.json`, and `config/tsconfig.tc.web.json`.

```bash
pnpm exec oxlint src/main/daemon/terminal-checkpoint-serializer.test.ts
```

Result: **passed with 0 diagnostics**.

No user-visible UI change; screenshots are not applicable.

#### What changed
- Added an over-limit multibyte serializer case where the complete JSON is 99 UTF-8 bytes over the cap while its UTF-16 code-unit length is one unit under the cap.
- Confirmed repo-wide that `json-utf8-byte-length` was imported only by its own test, then deleted both the dead module and its test.

#### ELI5
Some characters take more bytes to save than their JavaScript string length suggests, so the new test makes sure Orca measures the real saved size. The old measuring module was no longer connected to Orca, so it and its isolated test were removed.

#### Proof of fix
RED phase: archived head `88ca338811`, copied in the new test, and mutated `BoundedJsonWriter.append` in the temporary copy to compare and accumulate `value.length` instead of UTF-8 `bytes`.

```bash
temp_dir=$(mktemp -d /tmp/orca-pr11422-code-unit-red.XXXXXX) && git -C /Users/nwparker/orca/workspaces/orca/prod-scan-07-29 archive HEAD | tar -x -C "$temp_dir" && cp /Users/nwparker/orca/workspaces/orca/prod-scan-07-29/src/main/daemon/terminal-checkpoint-serializer.test.ts "$temp_dir/src/main/daemon/terminal-checkpoint-serializer.test.ts" && ln -s /Users/nwparker/orca/workspaces/orca/prod-scan-07-29/node_modules "$temp_dir/node_modules" && python3 -c 'from pathlib import Path; p=Path("'$temp_dir'/src/main/daemon/terminal-checkpoint-serializer.ts"); s=p.read_text(); s=s.replace("if (this.bytes + bytes > this.maxBytes)", "if (this.bytes + value.length > this.maxBytes)").replace("this.bytes += bytes", "this.bytes += value.length"); p.write_text(s)' && pnpm --dir "$temp_dir" exec vitest run --config config/vitest.config.ts src/main/daemon/terminal-checkpoint-serializer.test.ts -t 'rejects multibyte input whose code-unit length fits under the byte cap'
```

Result: **1 test file failed; 1 test failed, 3 skipped**. The mutation wrongly accepted and returned the direct over-byte-cap JSON, so `expect(serialized).not.toBe(expected)` failed.

GREEN phase on `4f28318842`:

```bash
pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/terminal-checkpoint-serializer.test.ts
```

Result: **1 test file passed; 4 tests passed**.

#### Proof of no regression
```bash
pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/history-manager.test.ts src/main/daemon/history-reader.test.ts src/main/daemon/terminal-history-incremental-restore.test.ts src/main/daemon/terminal-checkpoint-writer-bounds.test.ts
```

Result: **4 test files passed; 95 tests passed**.

```bash
pnpm run typecheck
```

Result: **passed** across all 3 configured projects: node, tc.cli, and tc.web.

```bash
pnpm exec oxlint src/main/daemon/terminal-checkpoint-serializer.test.ts
```

Result: **passed with 0 diagnostics**.

Repo-wide import confirmation at `88ca338811`:

```bash
rg -n --hidden --glob '!node_modules' --glob '!.git' "(?:from|import\\()\\s*['\\\"][^'\\\"]*json-utf8-byte-length['\\\"]|jsonUtf8ByteLength" /Users/nwparker/orca/workspaces/orca/prod-scan-07-29
```

Result: the only import was `src/main/daemon/json-utf8-byte-length.test.ts:2`; all other matches were declarations or calls inside that same module/test pair. After deletion, a repo-wide search returned no matches.

## Scan provenance
Finding: **P2 — Terminal checkpoint serializer walks the full payload three times synchronously**. Severity: **P2**. Introduced by **#10647** (`ca5a821600`).

Made with [Orca](https://github.com/stablyai/orca)


nwparker:
Post-merge follow-up: #11535.

The correctness of `BoundedJsonWriter` holds up — I differential-fuzzed the extracted implementation against `JSON.stringify` over 60,096 cases (randomized nested values plus targeted 16 KiB span-boundary cases across ASCII, 2/3-byte UTF-8, surrogate pairs, lone surrogates and control chars) and found zero divergence in output, in acceptance at exactly the byte cap, or in rejection at cap−1.

The performance claim is where it comes apart. Benchmarking the deleted `jsonUtf8ByteLength` + `JSON.stringify` path against the merged writer on SGR-styled scrollback (Node 24, median of 9 warmed runs, `--expose-gc`):

| content | 5MB scrollback | before #11422 | after #11422 |
|---|---|---|---|
| plain ASCII | | 12.9ms | 17.0ms (1.3x slower) |
| light ANSI | | 13.5ms | 25.4ms (1.9x slower) |
| agent-TUI ANSI | | 20.2ms | 86.2ms (**4.3x slower**) |
| over-cap reject | 10MB, 256KiB cap | 16.4ms | 1.7ms (10x faster ✓) |

The over-cap win is real. But the passing path — which runs on every checkpoint tick — got slower, and it degrades with ANSI escape density because each escape flushes the writer's 16 KiB span and appends separately. Claude Code / Codex TUI scrollback is the densest case, so the regression lands hardest on Orca's primary workload.

The measured range in the description (253.7–330.2ms before, 247.4–253.6ms after) is at 100 MiB, where the emulator replay dominates and masks the serializer delta.

#11535 keeps the bounded walk for sizing only and hands emission back to `JSON.stringify`: parity with pre-#11422 on the passing path, and 0.3ms on the reject path (better than both). It also closes the `appendJsonValue` non-exceed-`false` hole that could return a truncated non-null string, and restores the estimator cross-check.

PR #11427, opened 2026-07-29
perf(agent-status): bound Codex child rollout searches per reconcile

nwparker:
## Description

Locating a Codex child rollout scans a whole rollout directory — `resolveChildTranscript` runs `entries.find(entry => entry.endsWith('-<threadId>.jsonl'))` over up to `TRANSCRIPT_DIRECTORY_MAX_ENTRIES` (4096) names, for each child that has no `filePath` yet, on every reconcile. A parent with thousands of children not yet located therefore scans millions of names per Codex hook event, on the main process.

This caps the **searches** at 64 per reconcile and rotates the still-unfound children to the back of `state.subagents`, so a backlog takes turns instead of the same head being retried every tick. Reading an already-located child is a single `statSync` and stays **uncapped**, so completions are never deferred.

That distinction is the whole design. An earlier revision of this PR capped *children per tick* instead, which broke the completion path — see below.

## ELI5

Orca has to find each Codex sub-agent's log file by looking through a folder listing. With thousands of sub-agents that search got very slow. Now Orca only searches for 64 of them per check and remembers where it left off, so everyone still gets found. Reading a log file Orca has already found is cheap, so that part still happens for every sub-agent every time — which is what makes sure finished sub-agents always disappear from the list.

## Correction to the original premise

The first version of this PR said it was bounding *"one 1 Hz main-process poll."* There is no such poll. `reconcileCodexSubagentTranscript` has exactly one non-test call site:

```
src/shared/agent-hook-listener.ts:3360   ← inside normalizeCodexEvent
```

and `agent-hook-listener.ts` contains no `setInterval`. Reconciliation is hook-driven: it runs once per inbound Codex hook event and never otherwise. That mattered, because capping children per tick meant the leftovers only drained when another hook arrived — and after the final `Stop` hook, none ever does:

```ts
if (eventName === 'Stop' && !hasCodexTranscriptSubagents(state, paneKey)) {
  state.codexSubagentRosterByPaneKey.delete(paneKey)
}
```

`hasCodexTranscriptSubagents` is `state.subagents.size > 0`. With 200 completed children, the old approach finished 64, left 136 tracked, so the gate stayed false, the roster was never cleared, and 136 rows stayed `working` for the life of the pane. Capping searches instead of children removes that failure mode entirely: every located child is still read on every tick.

The rewrite also drops the `pendingReconcileIds: Set<string>` field the earlier revision added to the exported `CodexSubagentTranscriptState`, so the state shape is unchanged.

## Proof of fix

Two new tests, each RED against a different implementation.

**`bounds rollout searches per tick and rotates the backlog`** — fails on `main` (unbounded):

```
$ git archive origin/main | tar -x -C /tmp/red-main
$ cp src/shared/codex-subagent-transcript.test.ts /tmp/red-main/src/shared/
$ cd /tmp/red-main && pnpm exec vitest run --config config/vitest.config.ts \
    src/shared/codex-subagent-transcript.test.ts

  × bounds rollout searches per tick and rotates the backlog 25ms
  Tests  1 failed | 5 passed (6)
```

**`finishes every located child in a single tick, however many there are`** — fails on the earlier per-child-cap revision (`7e60037f6f`), which is the stranding regression:

```
$ git archive 7e60037f6f | tar -x -C /tmp/red-pr
$ cp src/shared/codex-subagent-transcript.test.ts /tmp/red-pr/src/shared/
$ cd /tmp/red-pr && pnpm exec vitest run --config config/vitest.config.ts \
    src/shared/codex-subagent-transcript.test.ts

  × finishes every located child in a single tick, however many there are 85ms
  AssertionError: expected true to be false
  Tests  1 failed | 5 passed (6)
```

Both green on this branch:

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```

## Measurements

Median of 25 warmed reconciles, Node 24 / macOS, against a session directory holding 4096 rollout files where no child rollout exists (so every child costs a search) — three independent runs:

| children | `main` | this PR |
|---|---|---|
| 100 | 4.4 / 4.4 / 4.6 ms | 3.6 / 3.9 / 4.0 ms |
| 1,000 | 7.2 / 6.7 / 6.5 ms | 4.1 / 4.1 / 4.1 ms |
| 5,000 | 23.9 / 23.4 / 23.4 ms | 3.6 / 3.8 / 3.7 ms |

At 5,000 children this is ~6x faster per hook event, and flat in child count rather than linear. The ~3.6ms floor is the single `readdirSync` of the 4096-entry directory, which is cached per reconcile and unaffected by this change. At 100 children the two are within noise — as expected, since 100 searches were never the problem.

## Proof of no regression

```
  src/shared/codex-subagent-transcript.test.ts \
  src/shared/agent-hook-listener.test.ts \
  src/shared/codex-subagent-roster.test.ts

Test Files  3 passed (3)
     Tests  126 passed (126)
```

Every suite referencing Codex subagents:

```
Test Files  4 passed (4)
     Tests  13 passed (13)
```

```
pnpm run typecheck        # passed: node, tc.cli, tc.web
pnpm run check:code-quality:changed
  code quality: 0 new finding(s) across 2 changed file(s).
  type-aware code quality: 0 new finding(s) across 2 changed file(s).
  React Doctor: 0 new finding(s) across 2 changed file(s).
```

The pre-existing grace behaviour is preserved deliberately: a child the budget declined to search for does **not** have its `unresolvedSince` clock started, so a backlog can never retire rows it never actually looked for.

## Dropped from the earlier revision

The previous test seeded 4095 decoy files. They were dead setup — the temp dir is not a `YYYY/MM/DD` path, so `childDayDirectory` returns `undefined`, and no `child-N` id had a rollout file, so `entries.find(...)` could not match at any decoy count. The assertions were identical with zero decoys, and the decoys cost roughly 32x the runtime of the reconcile calls under test. The new bounding test uses 500 children and no decoys.

## Scan provenance

P2 — Codex subagent reconcile scan degrades badly at scale. Introduced by #11059, "fix(agent-status): show Codex v2 subagents."

No user-visible change; this only bounds background main-process work, so screenshots are not applicable.

### DIFF
--- a/src/shared/codex-subagent-transcript.test.ts
+++ b/src/shared/codex-subagent-transcript.test.ts
@@ -6,7 +6,8 @@ import { join } from 'node:path'
 import {
   createCodexSubagentTranscriptState,
   hasTrackedCodexTranscriptSubagents,
-  reconcileCodexSubagentTranscript
+  reconcileCodexSubagentTranscript,
+  type CodexSubagentTranscriptState
 } from './codex-subagent-transcript'
 import { codexRosterToSnapshots, type CodexSubagentRoster } from './codex-subagent-roster'
 
@@ -27,6 +28,30 @@ function activity(kind: string, occurredAtMs = 1234): unknown {
       kind
     }
   }
+}
+
+function childId(index: number): string {
+  return `019fa65f-3144-7151-9c02-${String(index).padStart(12, '0')}`
+}
+
+function childActivity(index: number): unknown {
+  return {
+    type: 'event_msg',
+    payload: {
+      type: 'sub_agent_activity',
+      occurred_at_ms: 1234,
+      agent_thread_id: childId(index),
+      agent_path: '/root/sidebar_repro',
+      kind: 'started'
+    }
+  }
+}
+
+/** Children the reconciler went looking for — only a finished search sets the grace clock. */
+function searchedChildren(state: CodexSubagentTranscriptState): number {
+  return Array.from(state.subagents.values()).filter(
+    (tracked) => tracked.unresolvedSince !== undefined
+  ).length
 }
 
 /** `<root>/YYYY/MM/DD` for a timestamp, matching how Codex buckets rollouts by local start date. */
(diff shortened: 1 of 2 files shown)


nwparker:
## Review: the bounding premise doesn't match how this function is actually driven

The PR description says this prevents *"one 1 Hz main-process poll from synchronously examining thousands of unresolved children."* There is no 1 Hz poll. `reconcileCodexSubagentTranscript` has exactly one non-test call site:

```
src/shared/agent-hook-listener.ts:3360   ← inside normalizeCodexEvent
```

and there is no `setInterval` anywhere in `agent-hook-listener.ts`. Reconciliation is **hook-driven** — it runs once per inbound Codex hook event and never otherwise. That inverts the change's effect in two places.

### 1. The backlog has nothing to drain it — rows strand in "working" forever

With the cap, each call resolves at most 64 children and re-queues the rest at the tail of `pendingReconcileIds`. If hooks stop arriving, the queue stops draining. The final `Stop` hook is the problem:

```ts
if (eventName === 'Stop' && !hasCodexTranscriptSubagents(state, paneKey)) {
  state.codexSubagentRosterByPaneKey.delete(paneKey)
}
```

`hasCodexTranscriptSubagents` is `state.subagents.size > 0`. With 100 children that all completed, the `Stop` hook examines 64, finishes and deletes them, and leaves 36 in `state.subagents`. The gate is therefore false, the roster is never cleared, no further hooks arrive, and those 36 rows stay `working` for the life of the pane.

Pre-change, the loop iterated every tracked child, so `Stop` always drained fully. This is a functional regression at any count above `CHILD_RECONCILE_MAX_PER_TICK`, i.e. exactly the high-child-count regime the PR is aimed at.

### 2. Below 64 pending, the tick does *more* work than before, not less

The loop bound is a fixed iteration count, not `min(64, pendingReconcileIds.size)`:

```ts
for (let examined = 0; examined < CHILD_RECONCILE_MAX_PER_TICK; examined += 1) {
  const id = state.pendingReconcileIds.values().next().value
  ...
  state.pendingReconcileIds.add(id)   // both continue branches re-add
}
```

`Set` re-insertion after `delete` appends to the end of insertion order, so with fewer than 64 pending the queue wraps and the same ids are popped, re-resolved and re-read several times **within a single tick**.

Concrete: one parent with one running child (rollout exists, `task_started`, no `task_complete`). Pre-change that tick does 1 `statSync` on the child. Post-change it does 64 — the child is re-added and immediately re-popped 64 times. Each iteration also re-enters `resolveChildTranscript` for any child whose `filePath` was reset to `undefined`, re-scanning the (tick-cached) directory listing of up to 4096 entries.

For the common case of a handful of subagents, this is a ~64x increase in per-tick file I/O in a change whose stated purpose is to reduce it.

### 3. The new test can't catch either problem

`bounds child rollout discovery work per reconciliation tick` seeds 5000 children — the one regime where a fixed 64-iteration loop happens to behave correctly. It asserts the cap and round-robin advance, but nothing asserts the property that actually broke: *each pending child is visited at most once per tick*.

The 4095 decoy files are also dead setup. The temp dir is `mkdtemp(tmpdir(), 'codex-subagent-transcript-')`, which is not a `YYYY/MM/DD` path, so `childDayDirectory` returns `undefined`; and none of the 5000 `child-N` ids has a rollout file, so `entries.find(e => e.endsWith('-child-N.jsonl'))` can never match regardless of decoy count. The assertions are identical with zero decoys, and the decoys cost roughly 32x the runtime of the two `reconcile` calls the test exists to measure.

### Suggested direction

Bound the *expensive* work rather than the child count:

- Cap `resolveChildTranscript` directory scans per tick (that's the O(children x entries) part).
- Keep the cheap `statSync`/cursor read for every child that already has a resolved `filePath`, so completion is never deferred and the `Stop` gate still drains.
- If a per-child cap is kept anyway, bound the loop at `min(CHILD_RECONCILE_MAX_PER_TICK, pendingReconcileIds.size)` by snapshotting the batch first, so re-queued ids wait for the next tick instead of being re-read in this one.

Not merging as-is.

PR #11448, opened 2026-07-30
fix(remote): keep resumed agent tabs stable on headed hosts

brennanb2025:
## Summary

- keep headed runtime-session terminals when a renderer snapshot temporarily omits them
- track ownership through structured runtime creates, renderer-backed creates, and splits
- clear ownership on explicit close, headless retirement, and PTY exit
- require the exact worktree/tab/leaf PTY binding before preserving a tab
- remap host agent status to the paired client's `web-terminal-*` identity and retain the correct provider identity without restoring an older turn

## Root cause

A headed desktop runtime can create a local PTY on behalf of a paired client. The PTY remains live and authoritative in main, but the host renderer can omit its synthetic session tab during reconciliation. The previous preservation predicate only recognized headless, `serve-*`, and SSH-owned PTYs, so a live local PTY created through the runtime-session path could be dropped from the publication while its process remained alive.

The runtime now records whether a PTY is runtime-session-owned and preserves it only while its exact worktree, parent tab, and leaf binding is still live. Splits inherit ownership. Explicit close, retirement, and exit clear it, so this does not turn stale or rebound PTYs into durable tabs.

The renderer-side reconciliation also keeps the paired client's remapped ownership/provider metadata and prevents an older host `done` snapshot from restoring a previous provider session onto a newer client `working` turn.

## Compatibility and performance

- Local, WSL, SSH, and folder-workspace routing is unchanged.
- Existing `serve-*`, SSH, and headless preservation remains intact.
- The hot preservation check uses the already-bound terminal lookup plus a boolean ownership flag; it adds no polling, provider inventories, or broad PTY scans.
- Cleanup work is bounded to the tab's published leaves and runs only on close/retirement/exit paths.

### Headed Electron paired-runtime validation on final source

Validated commit `59db88d257` on macOS with an isolated headed Electron host and a separate paired Electron client. The harness rejected local scope and required:

```text
activeRuntimeEnvironmentId = 660bdb1b-b0cc-4b73-a185-5b938ccbb686
hostId = runtime:660bdb1b-b0cc-4b73-a185-5b938ccbb686
```

A structured Codex resume fixture reported through Orca's real hook and OSC status paths. The client switched to a second worktree for 15 seconds (past the 10-second close-intent TTL), returned, and confirmed:

| Assertion | Result |
|---|---|
| Mirrored tab/pane remapped to `web-terminal-*` | PASS |
| Provider identity retained | PASS (`session_id: pr11448-electron-resume`) |
| `working` state remains on the resumed tab | PASS |
| Exact authoritative PTY and handle remain connected | PASS |
| Agent process spawn count | PASS (exactly 1) |

This run covers the renderer-backed headed-host topology that previously dropped locally owned runtime PTYs. It is isolated same-machine pairing, not a physical remote-server, Windows, or Linux rerun. The earlier Windows 11 real-Codex validation still covers the original paired-runtime reproduction and renderer identity behavior, but the final ownership fix was not rerun on that rig.

### Automated checks

- [x] Same-model review loop; one exact-leaf-binding issue fixed, follow-up review clean
- [x] Focused Vitest: 986 passed, 1 skipped
- [x] Broader related matrix: 1,064 passed, 1 skipped
- [x] touched-file oxlint, native-plugin audit, and type-aware audit
- [x] touched-file `oxfmt --check`
- [x] `pnpm run check:max-lines-ratchet`
- [x] `pnpm run check:reliability-gates`
- [x] `git diff --check`
- [x] final E2E Electron and web builds

## Screenshots

Resumed agent working before the worktree round trip:

The same resumed tab and working agent after 15 seconds away and re-entry:

## Review history

- Same-model review found that the first ownership predicate did not verify the exact leaf binding. The fix now uses the authoritative terminal lookup, with PTY-rebinding and two-PTY split-close regressions. Follow-up review returned clean.
- Greptile's provider-identity loss finding was fixed in `d644677a8c`.
- Greptile's stale-attribution edge was fixed in `a73e37c032`.
- CodeRabbit reported no code-level findings on the earlier renderer changes.

## Security

No authentication, authorization, credential, filesystem, shell, or network-routing behavior changed. The new state is an in-memory PTY ownership marker used only by session-tab preservation and cleanup.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -24428,6 +24428,117 @@ describe('OrcaRuntimeService', () => {
     ])
   })
 
+  it('keeps a live headed runtime-owned tab until its explicit close', async () => {
+    const ptyId = 'local-runtime-owned-pty'
+    const splitPtyId = 'local-runtime-owned-split-pty'
+    const tabId = 'runtime-session-tab'
+    const leafId = HEADLESS_LEAF_ID
+    const kill = vi.fn(() => true)
+    const { runtimeStore } = makeRuntimeStoreWithWorkspaceSession({
+      ...getDefaultWorkspaceSession(),
+      activeRepoId: TEST_REPO_ID,
+      activeWorktreeId: TEST_WORKTREE_ID,
+      tabsByWorktree: {
+        [TEST_WORKTREE_ID]: [
+          {
+            id: tabId,
+            ptyId: null,
+            worktreeId: TEST_WORKTREE_ID,
+            title: 'Codex',
+            customTitle: null,
+            color: null,
+            sortOrder: 0,
+            createdAt: 1,
+            launchAgent: 'codex'
+          }
+        ]
+      },
+      terminalLayoutsByTabId: {
+        [tabId]: makeHeadlessTerminalLayout({ [leafId]: undefined })
+      }
+    })
+    const runtime = new OrcaRuntimeService(runtimeStore as never)
+    runtime.setPtyController({
+      spawn: vi.fn().mockResolvedValueOnce({ id: ptyId }).mockResolvedValueOnce({ id: splitPtyId }),
+      write: () => true,
+      kill,
+      getForegroundProcess: async () => null,
+      listProcesses: async () => [
+        {
+          id: ptyId,
+          cwd: TEST_WORKTREE_PATH,
+          title: 'Codex',
+          worktreeId: TEST_WORKTREE_ID
+        },
+        {
+          id: splitPtyId,
+          cwd: TEST_WORKTREE_PATH,
+          title: 'Codex',
+          worktreeId: TEST_WORKTREE_ID
+        }
+      ]
+    })
+    const publishRendererOmission = (snapshotVersion: number): void => {
+      runtime.syncWindowGraph(1, {
+        tabs: [],
+        leaves: [],
+        mobileSessionTabs: [
+          {
+            worktree: TEST_WORKTREE_ID,
+            publicationEpoch: 'headed-runtime',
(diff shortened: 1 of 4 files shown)


OrcaWin:
## Native Windows verification of `59db88d25`

Re-ran the reference failing configuration on **Windows 11 Pro 10.0.26200** — the topology where this bug was originally reported and the only place the failure had been reproduced. Both host and client on `59db88d25` (the fix is main-process, so the host must carry it), paired in runtime scope, real Codex CLI 0.145.0 resume from Agent Session History.

Previously this configuration dropped the resumed tab in 3/3 attempts. Now:

| Check | Result |
|---|---|
| Resumed tab survives 3 × 15 s switch-away cycles | PASS |
| PTY stays bound across every cycle | PASS |
| Agent status attached to the **resumed** tab (`web-terminal-0306131b-…`) | PASS |
| Provider session + transcript path correct | PASS (`019fb10c-…`) |
| Exactly one agent process spawned by the resume | PASS |

Dwell was 15 s deliberately — `CLOSE_INTENT_TTL_MS` is `10_000`, and the earlier 1.5 s validation passed precisely because it never reached the failure window.

Also re-ran the two changed suites locally: **987 passed**.

On the fix itself: it adds a narrow `runtimeSessionOwned` flag rather than widening `hasLiveOrPersistedServeOrSshOwnedPtyBinding`, which avoids the tab-resurrection hazard that predicate's comment warns about. All four set-sites are gated on runtime-session provenance — notably `ensurePtyBackedMobileSurfaceForRendererTab` early-returns unless a runtime-session create is pending, so ordinary renderer tabs never acquire the flag — and it is cleared on renderer close and PTY exit.

PR #11450, opened 2026-07-30
fix(mobile): hold Japanese kana composition off the PTY until a real boundary

brennanb2025:
## Summary

Part of #7427 — **this PR does NOT close the issue; it must stay open.** It fixes only the deterministic kana half of Japanese IME input on the mobile terminal; romaji remains out of scope.

On Japanese flick keyboards, the terminal previously received a base kana before a small-kana, dakuten, or handakuten modifier finished rewriting it. The composition mirror now holds kana until a real boundary and forwards the final text exactly once. Hangul keeps its existing timer policy, while kana is never committed by a timing guess.

The review loop also hardened every asynchronous input path around that boundary:

- Route, lifecycle, connection, producer-generation, and committed-session snapshots prevent deferred work from reaching a replaced terminal or replaying after reconnect.
- Field and Return boundaries recover only after definite rejection, only for the exact unchanged field and bytes, and at most once; unknown delivery is never replayed.
- Paste, image attachment and lease gating, dictation, buffered input, raw accessory input, accessory repeat, and gesture queues now distinguish definite rejection from delivery-unknown outcomes.
- Definite rejection preserves later input where recovery is safe. Delivery-unknown fences later work, stops active repeats, drops queued gesture batches, and reports uncertainty once instead of risking duplicate terminal input.

## Conflict repair

The exact pushed head is `308958eaf7cb4ab795c2e230da98244acf2f0e45`; its current-main merge parent is `8fc892dd023850df379c07cc4e7a46cff9a3ea84`.

- The earlier repair against main `b04c695750e9859e0d8bf79f503279bc4d656925` resolved one overlap in `mobile/app/h/[hostId]/session/[worktreeId].tsx`: current main's extracted live-input focus lifecycle was integrated after the PR's connection/scope gate without replacing its kana generation, ordering, or delivery-outcome protections.
- The final merge from `b04c695…` to live main `8fc892dd…` was conflict-free. Its only mobile delta was an unrelated active-PR polling hint in `mobile/src/session/github-pr-rpc.ts`.
- Local merge-tree against exact live main is clean. Romaji behavior was not changed.

## Screenshots

No visual change — this changes terminal byte flow, not layout or components. The embedded screenshots below are preserved as historical evidence from the earlier device lane; they are not current-head validation.

## Testing

- [x] Focused changed mobile tests: **24 files, 210/210 passed**.
- [x] Full mobile suite: **398 files, 2,991 passed / 3 skipped**.
- [x] Mobile TypeScript, oxlint, and formatting passed.
- [x] Root TypeScript under Node 24 and root oxlint passed; all 53 changed tracked files pass formatting.
- [x] Node 24 changed-code quality: **0 code-quality, 0 type-aware, and 0 React Doctor findings across 52 changed source files**.
- [x] Reliability manifest (**61 gates**), max-lines ratchet, native quality, iOS release-version (**11 runs / 23 assertions**), and `git diff --check` passed.
- [x] Non-vacuity: removing only the production kana `boundary` hold caused **12/26** composition tests to fail with the expected provisional-kana/DEL regressions; restoring exact committed production code returned **26/26** green and a clean tracked tree.
- [x] Hosted checks for `308958eaf7cb4ab795c2e230da98244acf2f0e45`: **43 successful**, including Greptile Review; only the path-gated e2e job is skipped.

**OWNER-LANE VERDICT: CLEAN**

## Simulator validation — exact public head (PASS)

Fresh Android A/B used the exact current-main baseline `8fc892dd023850df379c07cc4e7a46cff9a3ea84` and exact public head `308958eaf7cb4ab795c2e230da98244acf2f0e45`. The lane held one Android AVD (`emu_qa_keystore`, `emulator-5560`) alongside the unrelated booted iPad and never controlled the iPad. The Android app data was cleared before the baseline half, then the same AVD, prebuilt `com.stably.orca.mobile` dev client, native Gboard `日本語（12 キー）`, saved Host 1 profile, mock process, Narwhal terminal fixture, endpoint `10.0.2.2:6865`, Metro port `8092`, 4-second accepted-send response delay, and tap coordinates were held constant. Only the Metro-served mobile client bundle changed.

- **Baseline:** native `か` immediately reached the mock as provisional `"か"`; native dakuten later produced DEL + `"が"`; native `あ` then sent `"あ"`.
- **Exact head:** native `か` produced no send after a 2-second pause; dakuten changed the active composition to `が` and still produced no send after another pause. The real native `あ` boundary produced exactly one send, `"が"`, with no provisional `か`, no DEL repair, and no duplicate.
- Focused exact-head tests passed: **24 files, 210/210 tests**.

| Baseline — `が` is visible in native composition, but provisional `か` and DEL + `が` have already reached the PTY | Exact head — the same `が` remains held before a real boundary; the mock has received no bytes |
|---|---|
| Exact-baseline native kana composition | Exact-head native kana composition |

This exact-head acceptance run intentionally used only native Japanese flick input. Clipboard/ADB Unicode was not treated as acceptance evidence; the older paste and route-reuse evidence remains preserved below as explicitly historical evidence. Romaji was not tested and remains out of scope; #7427 stays open.

## Simulator validation — historical and stale

> **Stale for the current head:** no simulator or device was booted, attached, controlled, or killed during this repair. The Android A/B below validated `3fe69c66f9d070c0f7d87c23faa369372eb78016`, not current head `308958eaf7cb4ab795c2e230da98244acf2f0e45`; wholly fresh exact-parent/head device A/B remains required.

The historical lane used Android AVD `emu_qa_keystore` with native Gboard `日本語（12 キー）`. Parent was the then-true merge-base `4d044c47dccb58642cba6df6fe56e42b7b87dc6d`; head was `3fe69c66f9d070c0f7d87c23faa369372eb78016`. Both halves used the same AVD/app data, IME, dev client, delayed-acceptance mock, endpoint, terminal, and clipboard marker; only the Metro-served client bundle varied.

- Native `か` → dakuten → `あ`: parent sent provisional `か`, then DEL+`が`, then `あ`. The historical head sent nothing for the base/modifier edits and sent exactly one `が` only when `あ` created the real boundary; the later Enter flush sent `あ`, then CR.
- Paste boundary: parent sent provisional `か`, DEL+`が`, then the clipboard text. The historical head sent final `が` first and `CAQA12PASTE` second, exactly once each after accepted responses.
- Route reuse during the mock's 4-second `terminal.send` response delay: parent let the queued paste reach Manta after leaving Narwhal. The historical head sent only the held Narwhal `か`, dropped the stale queued paste on route replacement, then a fresh Manta Paste sent exactly once.

| Parent — active kana modified to `が` after provisional bytes had already leaked | Historical head — `が` remains held before a real boundary |
|---|---|
| Parent native kana composition | Historical-head native kana composition |

| Parent — stale queued paste reached the Manta route | Historical head — Manta target after stale paste was fenced |
|---|---|
| Parent route reuse target | Historical-head route reuse target |

Romaji was not tested and remains explicitly out of scope. Issue #7427 remains open.

## Same-model review loop

The earlier review-only lanes returned literal `CLEAN` on old head `3fe69c66f9d070c0f7d87c23faa369372eb78016`; those verdicts are historical and stale for the newly pushed head.

- [x] A wholly fresh independent `gpt-5.6-sol` `xhigh` read-only review returned literal `CLEAN` on exact head `308958eaf7cb4ab795c2e230da98244acf2f0e45` versus exact main parent `8fc892dd023850df379c07cc4e7a46cff9a3ea84`.
- [x] After that immutable-head review, a wholly fresh exact-parent/head native Gboard device A/B produced the current evidence above without promoting the stale evidence below.

## AI Review Report

The owner lane re-audited kana range and boundary semantics, exact-byte recovery, committed route ownership, disconnect/reconnect invalidation, producer-generation fences, direct and mirrored sends, pending flushes, paste, image attachments, lease gating, dictation, buffered sends, gestures, accessory repeat, and the current-main focus-lifecycle extraction. Definite rejection and delivery-unknown behavior were checked separately so recovery cannot duplicate bytes whose delivery is uncertain.

Cross-platform and remote behavior remain within the shared mobile terminal-send flow. No desktop shortcut, path, provider, or Git behavior changes, and SSH/folder workspaces receive the same ordered terminal bytes.

## Security Audit

The change only controls when user-authored terminal input is forwarded and whether uncertain sends may be retried. It adds no command execution, path handling, authentication, secret, dependency, or IPC surface. Treating delivery-unknown outcomes as a fence reduces the risk of replaying input into an interactive shell or TUI.

## Notes

- **#7427 stays open.** Romaji composition still leaks intermediate ASCII because React Native does not expose a deterministic composition lifecycle; fixing it requires a native IME-aware input view and is outside this PR.
- The earlier physical Gboard A/B at `8ae5c6d95d` and Android A/B at `3fe69c66f9d` remain useful historical evidence only and are stale for current head `308958eaf7cb4ab795c2e230da98244acf2f0e45`.
- The fresh immutable-head review, hosted checks, and native device A/B are complete; this QA lane did not merge or approve.

## What the user sees before and after

**Before:** With a Japanese IME, intermediate kana can leak into the live terminal while composition is still changing, so users can see duplicated or corrupted characters.

**After:** Kana stays buffered until a real composition boundary, then the final text reaches the terminal exactly once. Romaji composition is unchanged, so #7427 remains open.


brennanb2025:
Re: Greptile's P2 ("Boundary-held kana silently discarded on tab change — wider window than Hangul"): confirmed, and the PR description's Notes section has been updated to state it precisely. The discard mechanism in `clearPendingLiveInputCommit` predates this PR, but the `boundary` policy does widen the exposure from Hangul's ~300ms settle window to an indefinite one for held kana. It is deliberately left out of this PR to keep the change scoped to the deterministic kana-modifier fix; the follow-up (flush instead of discard on tab change) is called out in the Notes section.

nwparker:
Closing — but **your patch is the part that shipped**, and you're credited as co-author on #15007 (`66b5993`).

Your premise was the load-bearing finding: React Native 0.83.9 genuinely does not expose composition state, so no amount of JS gets at it. I checked this the hard way — an earlier pass concluded RN already shipped `isComposing`, from reading a workspace whose RN copy had been hand-patched. A pristine install proved you right.

`mobile/patches/react-native@0.83.9.patch` is regenerated cleanly against pristine 0.83.9 and carries your authorship. A readiness review verified it cryptographically: pre-image blob hashes match the npm tarball, post-image match the install, and a full recursive diff shows exactly the three files it names and nothing else.

What changed on the JS side: your `if (isComposing === true) return` left the Hangul allowlist and the DEL-repair path in place for everything else, so Android — which can't compile the patch, since it links a prebuilt AAR — kept the original bug. #15007 replaces the allowlist entirely and adds a non-ASCII-run fallback for that case.

Closing the branch, not the contribution. Thanks — the patch was the hard part.

Issue #8791, opened 2026-07-15
Mobile: Remove/Rename host freezes the app and Remove does not delete


cl1107:
## Bug Report

On the **mobile app**, long-pressing a host and using the action sheet has broken interactions for both **Remove** and **Rename**.

### Environment
- Platform: Mobile (iOS, from screenshot)
- Screen: Welcome back / Hosts list
- Host shown: `Host 1` (`192.168.21.4:6768`), status Connected · 4 worktrees · 4 active

### Steps to Reproduce

1. Open the Orca mobile app
2. On the home screen, long-press a host card (e.g. **Host 1**)
3. In the bottom action sheet, tap **Remove** (or **Rename**)

### Expected Behavior
- **Remove**: The host should be deleted/removed from the list, and the UI should remain responsive
- **Rename**: A rename flow should open normally without freezing the page

### Actual Behavior
- **Remove**: The host is **not** deleted
- After tapping **Remove**, the **entire page freezes/hangs** and becomes unresponsive
- Tapping **Rename** also **freezes the entire page**

### Screenshot

The action sheet after long-pressing Host 1 (Reconnect / Disconnect / Rename / Remove):

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/c6c39cf5-38bb-4d2d-a244-46f076260710" />

### Additional Notes
- Confirmed broken actions: **Remove** and **Rename**
- The freeze appears to lock the whole screen, not just the action sheet
- May require force-closing the app to recover after the freeze

### Impact
Unable to remove or rename hosts on mobile; app becomes unusable after the freeze until restarted.

jhste102lab:
same as me.

wohsj110:
Additional confirmation from another mobile user: host names cannot be changed. After attempting to rename a host, it remains on the default generated name (for example, `Host 1`, `Host 2`, or `Host 3`) instead of the user-provided name.

This prevents users from assigning meaningful names to hosts on mobile and makes multiple hosts difficult to distinguish.

kaiquecav:
For me its a bit different, no freeze here. When i tap "Edit host" it just opens
the host detail screen, the edit form never shows.

Same thing with both my hosts, one is a headless orca serve on linux and the
other is the mac desktop app. Also tried removing and pairing again, same
behavior, name comes back as Host 1 and 2.

iOS app, orca v1.4.159.

PR #8536, opened 2026-07-13
Add Orca Relay desktop and mobile transport

Jinwoo-H:
## Summary

Adds Orca Relay end to end while preserving account-free LAN/Tailscale pairing as the preferred path:

- desktop auth lifecycle, scoped relay-token exchange integration, director assignment, host proof/control, data transports, origin-aware drain/rebind, revoke durability, and thin Mobile settings status UI;
- additive pairing offers and E2EE v2 with transcript binding, directional keys, session/counter framing, bounded queues, and the exact legacy direct path for version skew;
- mobile pairing races, crash-safe credential reconciliation, configured-director recovery, native credential storage, stable logical RPC cutover, endpoint hysteresis, and relay E2EE connect/migration;
- shared strict protocol fixtures, recovery matrices, and adversarial integration coverage.

Relay remains free but requires desktop Orca sign-in. The phone does not sign in, there is no rollout flag, and older desktop/mobile counterparts continue through the additive offer and legacy direct path.

Review-only area PRs (do not merge independently):

- Desktop transport/broker: #8526
- Desktop sign-in UI: #8529
- Mobile endpoints/racing: #8530
- Mobile relay connect/migration: #8531

This combined PR is the sole merge candidate.

## Screenshots

The settings UI is intentionally thin pending the separately designed final UI. No screenshot attached yet.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` (the full unsupported-Node-26 run passed 2,766 files and exposed 10 failures; under required Node 24, nine passed in isolation and the one branch-caused stale assertion was corrected and passed)
- [ ] `pnpm build`
- [x] Added adversarial protocol, E2EE, auth lifecycle, broker, pairing/recovery, endpoint, migration, cleanup, UI, and real local-relay integration tests

Supported Node 24 validation:

- 309 focused desktop/shared files: 3,022 passing tests, 2 files / 4 tests skipped;
- all 224 mobile files: 1,571 passing tests, 2 skipped;
- max-lines ratchet, standalone mobile typecheck, mobile lint, and mobile format checks passed;
- each of the 10 files that failed under Node 26 was rerun independently; all pass after correcting the migrated connection-count assertion.

## AI Review Report

[...]

PR #11458, opened 2026-07-30
fix(mobile): stop serving a pre-write host-list snapshot to loads issued after the write

brennanb2025:
## Summary

Fixes the "Remove does not delete" half of #8791: removing (or renaming) a host committed to disk, but the home screen could immediately repaint the pre-write list, so the removed host's card came back.

Root cause: `loadHosts()` de-duplicates concurrent calls by handing back the in-flight pass. That de-dupe defeats the store's own `await hostListMutation` invariant — a pass that started *before* a write still holds the pre-write snapshot, and a load issued *after* the write commits (exactly the shape of `handleRemove` in `mobile/app/index.tsx`) was served that stale snapshot. Now every durable host write drops the shared pass (`mobile/src/transport/host-list-load-sharing.ts`), so the next load reads fresh; concurrent loads with no write between them still share one Keychain pass.

Note on scope: the freeze half of #8791 ("app frozen after tapping Remove") needs no code change — the host action sheet already sets `closeBeforePress` on Edit host + Remove since #8536. This PR extracts that sheet into `mobile/src/host-list-action-sheet-actions.ts` so the deferral is pinned by tests and can't silently regress.

### Before

Removing or renaming a host writes the change, but the home screen can reload an older snapshot. A removed host card then reappears, making it look like removal failed.

### After

The next host-list refresh reads the newly saved state, so removed hosts stay gone and renamed hosts retain their new names.

## Simulator validation

Manual iOS Simulator A/B on Lane A device `PR 8647 QA iPhone`: true merge-base `5d88c1093fd8a75c4e1c3a86c34bf4f704a9d6e8` versus exact head `fed029f1f6da37c0a45912afff3a228a18303347`.

### Before — parent resurrects the removed card

Parent: removed Host 1 card repainted from a stale pre-write snapshot

Durable storage was already `orca:hosts=[]`, but the post-remove refresh reused the parked pre-write result and repainted Host 1 as disconnected.

### After — exact head keeps the card removed

Exact head: Host 1 remains absent after the stale pass resolves

The same parked old pass resolved with the former host id, while the post-write refresh read `orca:hosts=[]`; the card stayed gone beyond 15 seconds.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [ ] `pnpm build` (not run: mobile-only change, `mobile/` has no build script; Mobile Checks CI covers it)

Run in `mobile/` on node 24. Baseline at parent commit: 351 files / 2587 tests → with this change: 353 files / 2601 tests (+14). Non-vacuity proven both ways: the new host-store tests park a load mid-Keychain-read, commit a removal/rename underneath it, then assert a load issued after the write never sees the pre-write snapshot — they fail on the parent commit and pass here; the shared-pass test asserts de-dupe still works when no write intervenes.

## AI Review Report

Reviewed the stale-snapshot race end to end: writer paths (`mutateStoredHosts`, `persistHost` relay-upgrade + overlay-removal branches, `removeHost` overlay cleanup) all drop the shared pass after their durable commit; the late-settle case (a dropped pass resolving after its replacement started) is handled by only retiring the pass still on offer, with a test pinning it. Checked that the action-sheet extraction is behavior-preserving (same labels, ordering, live-state gating, `closeBeforePress` placement). Cross-platform: this is React Native (iOS/Android) code only — no desktop shortcuts, labels, paths, shell, or Electron surface touched; no platform-conditional logic involved.

## Security Audit

No new input handling, command execution, path handling, auth, or IPC. The change only affects when an in-memory promise cache is invalidated; credential reads still go through the existing SecureStore path and token cache, and write serialization via `hostListMutation` is unchanged. No new dependencies.

## Notes

Residual (pre-existing, not addressed here): `MobileEndpointSupervisor` persists via plain `saveHost` (`mobile/src/transport/mobile-endpoint-lifecycle.ts:92`), which *appends* a host that was removed mid-flight instead of throwing like `saveExistingHostRelayUpgrade` does — a background relay upgrade racing a removal can still resurrect the host through that path.


brennanb2025:
Re the two Greptile P2 observations — both are accurate readings, and both are deliberate consequences of the module invariant stated on `dropSharedHostListLoad`: **every durable host write drops the shared pass**, with no exceptions.

1. **Drop inside `removeHost`'s overlay-removal `try`**: redundant today, agreed — `loadMobileRelayHostOverlayState` filters overlays by the host-id set already committed to AsyncStorage, so the earlier drop in `mutateStoredHosts` suffices. But `removeMobileRelayHostOverlay` *is* a durable write, and omitting the drop there would couple this call site's correctness to `doLoadHosts`' current overlay-filtering behavior (the same overlay writes in `persistHost` *do* affect hosts still on the list). The uniform rule is cheap and survives refactors; special-casing it is the misleading option.

2. **`updateLastConnected` evicting the pass on every connection event**: intended. `lastConnected` is durable data served by `loadHosts()`, so a load issued after that write should read fresh — same invariant as remove/rename. The drop itself does no work; the only cost is a mount burst racing a connection event paying one extra Keychain pass, and sharing within any burst that starts after the drop is unaffected.

PR #11483, opened 2026-07-30
fix(skills): keep copied Git Bash commands runnable

### LINKED ISSUE #11431: Copied skill install command silently no-ops in Git Bash after the cmd.exe npx preflight (#10453)
### Summary

#10453 wraps the skill install command in `cmd.exe /d /s /c "..."` on native Windows so it can preflight `npx`. The spawned-terminal path is guarded — `getProjectAgentSkillTerminalShellOverride` (`src/renderer/src/lib/project-skill-runtime.ts:54`) forces `powershell.exe` when the configured Windows shell is POSIX-family, with a comment naming exactly this hazard.

But that override only governs the shell **Orca spawns**. It cannot govern where a user *pastes*, and four call sites put the wrapped string on the clipboard:

- `src/renderer/src/components/settings/AgentSkillSetupPanel.tsx:156` (`copyActiveCommand` copies `openTerminalCommand`)
- `src/renderer/src/components/feature-tips/CliSkillSetupTerminal.tsx:24`
- `src/renderer/src/components/settings/OrchestrationSkillPromptDialog.tsx:23`
- `src/renderer/src/components/settings/linear-agent-skill-install-cta.tsx:72`

Before #10453 the copied text was plain `npx skills add ... --global`, which works in Git Bash. After #10453 it is the `cmd.exe` wrapper, which in Git Bash **silently does nothing**. For users who copy the command into their own Git Bash, this is a regression against the previous behavior, not merely an unimproved case.

### Why it fails silently

Git Bash's MSYS layer rewrites single-letter switches that look like POSIX paths. Measured on real Windows (Win 10.0.26200.8875, Git Bash MINGW64) with a node argv dump:

```
$ node argvdump.js /d /s /c "where.exe npx >nul 2>nul & if errorlevel 1 (...) else (...)"
ARGV0=[D:/]
ARGV1=[S:/]
ARGV2=[C:/]
ARGV3=[where.exe npx >nul 2>nul & if errorlevel 1 (...) else (...)]
```

`/d /s /c` become `D:/ S:/ C:/`. cmd.exe therefore never receives `/c`, ignores the unrecognized arguments, and **starts an interactive session** — banner, `C:\...>` prompt, **exit 0**, payload never runs. With npx also missing the guidance never fires either, so there is no error anywhere.

Note the payload itself (`ARGV3`) survives intact — only the switches are eaten. Quoting is a red herring when debugging this.

### Measured matrix

| Shell | npx present | npx missing |
|---|---|---|
| pwsh 7.6.4 | PASS | PASS (exit 1, guidance shown) |
| PowerShell 5.1.26100.8875 | PASS | PASS |
| cmd.exe | PASS | PASS |
| Git Bash (MINGW64) | **FAIL — silent no-op, exit 0** | **FAIL — no guidance** |

### Suggested fix

Thread the check #10453 already wrote one level up. In `wrapWindowsSkillCommandWithNpxPrerequisite` (`src/renderer/src/components/settings/CliSkillRuntimeSetup.tsx`), skip the wrapper when `resolveWindowsShellStartupFamily(settings?.terminalWindowsShell) === 'posix'` and return the bare command as `main` did. `buildSkillCommandForRuntime` already reads the store for `isRemoteRuntimeEnvironmentFocused()`, so this reuses an established pattern rather than introducing one.

**Do not use `cmd.exe //c`.** MSYS collapses `//c` to `/c`, but cmd.exe itself does not accept `//c`, so it would break the three shells that currently work.

### Also worth a look while in here

`where.exe npx` on a normal Node install lists the extensionless `C:\Program Files\nodejs\npx` (a bash script) *first*. So `where.exe` succeeding does not strictly prove cmd.exe can execute npx — a PATHEXT-less file satisfies the probe. Harmless in practice since `npx.cmd` sits beside it, but the preflight is slightly weaker than it reads.

### Provenance

Validated on a real Windows host after #10453 merged (`0d4baf2a6354ac7538af078abb75b905c6785736`). The wrapper string was extracted mechanically from `refs/pull/10453/head` rather than transcribed, and matched the PR's own `WINDOWS_INSTALL_COMMAND` fixture byte-for-byte. Not reverted because the failure is a silent no-op rather than destructive, it needs a Git-Bash-configured user who copy-pastes rather than using the in-app terminal, and all three Windows-native shells are measured correct.

bbingz:
## Summary

- Separate copied skill commands from the command Orca inserts into its controlled setup terminal.
- Remove the exact native-Windows `cmd.exe` npx preflight only at clipboard boundaries when Orca has forced the setup terminal to PowerShell for a POSIX-family configured shell.
- Keep the preflight and missing-npx guidance in the controlled terminal, and preserve PowerShell, Command Prompt, WSL, remote-runtime, and non-Windows behavior.
- Cover all four skill-command clipboard surfaces identified in the issue.

Fixes #11431

## Screenshots

No layout change. With Git Bash configured, command previews now show the runnable bare `npx skills ...` command instead of the `cmd.exe /d /s /c` wrapper; Orca setup terminals still receive the wrapped command.

## Testing

- [ ] pnpm lint
- [x] pnpm typecheck
- [ ] pnpm test
- [ ] pnpm build

Focused checks:

- Node 24.18.0: 52 tests passed across command construction, clipboard behavior, all clipboard call-site contracts, active runtime resolution, Windows shell overrides, and Linear runtime routing.
- Red/green regression: the new panel test failed before the fix because preview and clipboard still received the `cmd.exe` wrapper, then passed while asserting the in-app terminal retained it.
- pnpm run check:code-quality:changed -- origin/main
- pnpm run check:max-lines-ratchet
- oxfmt --check on all changed files
- git diff --check

## AI Review Report

Reviewed the four clipboard call sites, native Windows host command normalization, remote focus, WSL command wrapping, and shell override resolution. Clipboard unwrapping is restricted to the exact wrapper generated by this module and only when the controlled terminal requires a PowerShell override; unrelated commands and native PowerShell/cmd users remain unchanged. A call-site contract test prevents any of the four copy surfaces from bypassing the clipboard-safe boundary.

## Security Audit

The change adds no new command execution, IPC, filesystem, network, dependency, authentication, or secret-handling surface. It removes a generated wrapper only for displayed/copied text and leaves the controlled execution path unchanged. Exact prefix/suffix matching prevents arbitrary command rewriting.

## Notes

No live Windows build was run for this patch. The issue already contains a measured real-Windows shell matrix; this PR adds deterministic boundary tests for the regression and preserves the previously validated native-shell execution path.

### DIFF
--- a/src/renderer/src/components/feature-tips/CliSkillSetupTerminal.tsx
+++ b/src/renderer/src/components/feature-tips/CliSkillSetupTerminal.tsx
@@ -3,7 +3,10 @@ import { toast } from 'sonner'
 import { Button } from '@/components/ui/button'
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip'
 import { OnboardingInlineCommandTerminal } from '@/components/onboarding/OnboardingInlineCommandTerminal'
-import { buildSkillCommandForRuntime } from '@/components/settings/CliSkillRuntimeSetup'
+import {
+  buildSkillCommandForRuntime,
+  getSkillCommandForClipboard
+} from '@/components/settings/CliSkillRuntimeSetup'
 import { ORCA_CLI_ORCHESTRATION_SKILL_INSTALL_COMMAND } from '@/lib/agent-feature-install-commands'
 import { useActiveProjectSkillRuntime } from '@/hooks/useActiveProjectSkillRuntime'
 import { translate } from '@/i18n/i18n'
@@ -18,10 +21,14 @@ export function CliSkillSetupTerminal(): React.JSX.Element {
     ORCA_CLI_ORCHESTRATION_SKILL_INSTALL_COMMAND,
     activeSkillRuntime.installDisabledReason ? undefined : activeSkillRuntime.agentRuntime
   )
+  const clipboardCommand = getSkillCommandForClipboard(
+    skillCommand,
+    activeSkillRuntime.terminalShellOverride
+  )
 
   const handleCopySkillCommand = async (): Promise<void> => {
     try {
-      await window.api.ui.writeClipboardText(skillCommand)
+      await window.api.ui.writeClipboardText(clipboardCommand)
       toast.success(
         translate(
           'auto.components.feature.tips.CliSkillSetupTerminal.b8ad063571',
@@ -44,7 +51,7 @@ export function CliSkillSetupTerminal(): React.JSX.Element {
     <div className="min-w-0">
       <div className="flex min-w-0 items-center gap-2 rounded-md border border-border bg-muted/35 px-3 py-2">
         <code className="scrollbar-sleek min-w-0 flex-1 overflow-x-auto whitespace-nowrap font-mono text-xs text-muted-foreground">
-          {skillCommand}
+          {clipboardCommand}
         </code>
         <Tooltip>
           <TooltipTrigger asChild>
--- a/src/renderer/src/components/settings/AgentSkillSetupPanel.test.tsx
+++ b/src/renderer/src/components/settings/AgentSkillSetupPanel.test.tsx
@@ -9,6 +9,7 @@ import { TooltipProvider } from '../ui/tooltip'
 
 const INSTALL_COMMAND = 'npx skills add https://github.com/stablyai/orca --skill orca-cli --global'
 const UPDATE_COMMAND = 'npx skills update orca-cli --global'
+const WINDOWS_INSTALL_COMMAND = `cmd.exe /d /s /c "where.exe npx >nul 2>nul & if errorlevel 1 (echo ERROR: npx was not found. Install Node.js LTS from https://nodejs.org/ to get npx. & echo Then close this terminal and start skill setup again - a new terminal picks up the updated PATH. & exit /b 1) else (${INSTALL_COMMAND})"`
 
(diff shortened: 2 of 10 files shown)


nwparker:
Closing as superseded by [#11599](https://github.com/stablyai/orca/pull/11599) (and follow-up [#11814](https://github.com/stablyai/orca/pull/11814)).

Same issue [#11431](https://github.com/stablyai/orca/issues/11431): keep copied skill install commands bare/runnable under Git Bash so the `cmd.exe /d /s /c` preflight doesn't get MSYS-mangled into a silent no-op. #11431 is already closed against that merge.

If a call site still copies the wrapped form on current main, open a tiny follow-up rather than rebasing this whole branch.

bbingz:
Confirmed on `2ff2a1b26` — no follow-up needed. Every clipboard surface builds through `buildSkillCommandForRuntime`, which passes `'copied-command'` (`CliSkillRuntimeSetup.tsx:92-96`), and the wrapper skips the `cmd.exe /d /s /c` preflight for POSIX-family Windows shells (`:181`). The only remaining producer of the wrapped form (`:192`) is reached via `buildSkillSetupTerminalCommand` for Orca's own setup terminal, which is what #11814 preserved. Closing out on my side.

PR #11489, opened 2026-07-30
fix(skills): evict removed runtime discovery cache

### LINKED ISSUE #11429: Skill discovery cache is not evicted when a runtime environment is removed
### Summary

The installed-skill discovery cache is bounded (#7670) and keyed by runtime scope (#6887), but entries are never evicted when a runtime environment is **removed**. A removed environment's skill list is retained until LRU pressure or an install notification happens to clear it.

This is a staleness bug as well as a memory one: re-pairing an environment that reuses an id would serve the old skill list.

### Where

- `src/renderer/src/hooks/installed-agent-skill-discovery-cache.ts` — the 256-entry LRU added by #7670
- `getRuntimeScopedSkillDiscoveryKey` (#6887) — remote scans key on `runtime:<environmentId>`
- Invalidation today is `notifyInstalledAgentSkillsChanged`, which is install-driven and has no notion of an environment going away

### Why it matters more for ephemeral VMs

Ephemeral VMs mint a fresh environment id per start — `orca-${randomUUID()}` (`src/main/ephemeral-vm-runtime-service.ts`, `src/shared/ephemeral-vm-recipe-runner.ts`). Each start therefore creates a new cache key that is retained after the VM is gone. At roughly 18.8 KB per entry that is ~1.8 MB after ~100 starts, capped only by the 256-entry LRU. Saved runtimes are not the churn vector; ephemeral ones are.

### Suggested fix

Subscribe to `runtimeEnvironments` removals and evict any cache entry whose key is scoped to the removed environment id. Two things to be careful about, both learned the hard way in #6887/#7670:

- The discovery module must stay store-free — it must not import `@/store`, or the `store → repos slice → hook → store` import cycle returns and three suites break. The subscription belongs on the consuming side, with eviction exposed as a plain function on the cache module.
- Do not widen the selector to the whole `runtimeEnvironments` array. Its identity churns on every status refresh, which re-fires ~20 consumers' scans (this was a real finding in #6887 round 2). Select the resolved ids as a string, or subscribe to removals specifically.

### Context

Deliberately deferred from #7670 because it is new behavior rather than conflict reconciliation, and that PR was already carrying a contributor's commit through a merge. Called out in the #7670 squash message as a known gap.

bbingz:
## ELI5

When you remove or re-pair a server, Orca now forgets its old cached skill list and prevents a late scan from restoring it. Other servers and local skill lists keep their caches.

## Details

Removed and re-paired runtime environments left their skill-discovery results in the renderer cache. Retired entries could also displace useful local/WSL entries from its 256-target capacity, and a late scan could restore a removed entry.

The existing catalog removal/re-pair boundary now evicts the exact runtime-scoped cache entry and detaches its pending scan. Completion writes use the existing current-promise identity check, so a retired scan cannot restore the cache or overwrite its replacement. This needs no additional invalidation collection. Surviving runtimes, local/WSL targets, scan coalescing, forced refresh behavior, and the 15-second freshness policy are preserved; unchanged catalog updates do not trigger scans.

Validation against main `fa3d29fa11e3`, with `ORCA_BACKGROUND_LAUNCH=1`:

- 119 tests passed across ten suites, covering cache/discovery, mounted hooks, runtime status/catalog identity, and removal wiring.
- A 512-runtime retirement workload leaves exactly the two surviving local/WSL cache entries. Pending work for a surviving runtime remains shared.
- Both completion orders around same-ID re-pairing preserve the replacement's result and pending slot.
- The original baseline fails four retirement regressions. Removing the promise-identity fence fails three late-result regressions. The restored fix passes again.
- Web typecheck and all three changed-code quality gates passed.

Original contributor history is preserved. These are local unit/component checks on macOS; no live Electron or remote-host run is claimed. No layout, wire, filesystem, or provider API changed.

Fixes #11429.


bbingz:
**Contributor hygiene:** rebased onto current `main` and resolved the `runtime-status.ts` import conflict (kept main's disconnect-toast/hydration/refresh imports + this PR's skill-cache eviction call). Force-pushed to `fix/evict-removed-runtime-skill-cache`.

Still needs a maintainer to **Approve and run workflows** for full CI.

PR #8430, opened 2026-07-12
fix(agent-hooks): drain stdin before hook script early exits so agents never hit EPIPE

kaynansc:
## Summary

Fixes #8419.

Orca's generated POSIX agent hook scripts read their payload with `payload=$(cat)` **after** an env guard that exits early:

```sh
if [ -z "$ORCA_AGENT_HOOK_PORT" ] || [ -z "$ORCA_AGENT_HOOK_TOKEN" ] || [ -z "$ORCA_PANE_KEY" ]; then
  exit 0
fi
payload=$(cat)
```

When an instrumented agent runs outside an Orca pane (SSH-booted `codex app-server` daemons, detached tmux sessions, plain-terminal launches — the hooks are registered in **global** agent config, so they fire there too), the guard exits without ever reading stdin, and the agent's payload write fails with `failed to write hook stdin: Broken pipe (os error 32)` on **every tool call**.

This PR drains stdin (`cat >/dev/null 2>&1 || :`) before every early `exit 0` that precedes `payload=$(cat)`:

- All 11 POSIX script templates that use this pattern: `antigravity`, `claude` (both guards, including the `DEVIN_PROJECT_DIR` early exit), `codex`, `command-code` (in `command-code-managed-script.ts`), `copilot`, `cursor`, `devin`, `droid`, `gemini`, `grok`, `kimi`. (`openclaude` shares the Claude template, so it's covered; `opencode`/`amp`/`hermes`/`mimo` use in-process JS/Python plugins that never read a piped stdin, so the bug class doesn't apply to them.)
- `wrapPosixHookCommand` in `installer-utils.ts`: the wrapper written into agent configs (`if [ -x '<hook>' ]; then …; fi`) now has an `else cat >/dev/null 2>&1 || :; fi` branch, so a missing/non-executable hook script no longer drops stdin either.

The normal in-pane forwarding path (`payload=$(cat)` → curl POST) is byte-for-byte unchanged — "agent not running under Orca" is now silent instead of a per-tool-call error.

## Screenshots

No visual change (generated shell scripts only). End-to-end proof using the deterministic reproduction from the issue (simulated agent writes a 100 KB hook payload to the script's stdin with `ORCA_*` env stripped, 5 runs), against the real script produced by `CodexHookService.install()`:

```
=== BEFORE (script generated from origin/main) ===
codex-hook.sh: BrokenPipeError 5/5

=== AFTER (script generated from this branch) ===
codex-hook.sh: BrokenPipeError 0/5

[...]

Issue #6078, opened 2026-06-22
[Bug]: Codex hooks fail on Windows when user path contains spaces


jorge-huxley:
### Operating system

Windows

### Orca version

v1.4.91.0

## Short summary

Orca-generated Codex hook commands are not safely invoked on Windows when the Windows user profile path contains a space.

## What happened?

When Codex was launched through Orca, the `SessionStart` and `UserPromptSubmit` hooks failed with exit code 1:

```text
SessionStart hook (failed)

error: hook exited with code 1

UserPromptSubmit hook (failed)

error: hook exited with code 1
```

Orca generated the following command in its Codex runtime hook configuration:

```json
"command": "C:\\Users\\Jorge Silva\\.orca\\agent-hooks\\codex-hook.cmd"
```

The active configuration was located at:

```text
%APPDATA%\orca\codex-runtime-home\home\hooks.json
```

The normal `%USERPROFILE%\.codex\hooks.json` file was not the configuration used by the Orca-launched Codex process because Orca sets `CODEX_HOME` to its own runtime directory.

On Windows, the generated command is split at the space in `Jorge Silva`, so the hook runner attempts to execute an invalid path.

## How can we reproduce it?

1. Use Windows with a user profile path containing a space, such as `C:\Users\Jorge Silva`.
2. Launch Codex through Orca with agent hooks enabled.
3. Start a new Codex session.
4. Submit a prompt.
5. Observe `SessionStart` and `UserPromptSubmit` hook failures with exit code 1.

## Expected behavior

Orca should generate a Windows-safe command that explicitly invokes the batch file through `cmd.exe` and quotes its path:

```json
"command": "cmd.exe /d /c call \"C:\\Users\\Jorge Silva\\.orca\\agent-hooks\\codex-hook.cmd\""
```

## Confirmed workaround

Replacing every Orca Codex hook command in `%APPDATA%\orca\codex-runtime-home\home\hooks.json` with the following fixes the issue:

```text
cmd.exe /d /c call "C:\Users\Jorge Silva\.orca\agent-hooks\codex-hook.cmd"
```

After applying this change:

```text
SessionStart_exit=0
UserPromptSubmit_exit=0
```

The fix was also confirmed by restarting Codex through Orca and using it normally.

## Suggested fix

When generating Codex hook configuration on Windows, Orca should invoke `.cmd` hook files through `cmd.exe /d /c call` and quote the full path. This should be covered by a test using a Windows profile path containing spaces.

[...]

PR #11497, opened 2026-07-30
fix(claude): widen Git Bash hook fast path to quoted-safe paths

### LINKED ISSUE #11496: [Bug]: Non-ASCII home paths skip the Git Bash hook fast path, tripling per-hook overhead
### Operating system

Windows

### Orca version

v1.4.159

### Details

Thanks for Orca — the #7131 fast path was a noticeable win when I profiled my hook setup.
This is about extending that win to more users.

### Short summary

`WINDOWS_GIT_BASH_SAFE_PATH` rejects any path containing non-ASCII characters, so users
whose Windows profile name is CJK/Cyrillic/accented never get the Git Bash fast path from
#7131 — every managed hook falls back to the PowerShell `-EncodedCommand` launcher.

Since #8430 the fast path wraps the script path in `quotePosixShellString()`, so the
gate's original concern (bash splitting on spaces) is already handled by the quoting. The
gate now only excludes paths that would work fine. I have a PR ready and will open it
right after this issue.

### What happened

`src/main/agent-hooks/installer-utils.ts:154`

```ts
const WINDOWS_GIT_BASH_SAFE_PATH = /^[A-Za-z0-9_.:/~-]+$/
```

A home directory like `C:\Users\<CJK name>\.orca\agent-hooks\claude-hook.cmd` fails this
test, so `wrapWindowsGitBashHookCommand` returns the PowerShell fallback for all 10
managed Claude Code hooks (UserPromptSubmit, Stop, StopFailure, SubagentStart,
SubagentStop, TeammateIdle, PreToolUse, PostToolUse, PostToolUseFailure,
PermissionRequest).

Measured wrapper overhead on my machine (Orca 1.4.159, Windows 11 26200, n=10,
alternating execution from the same parent process, 2 warmup runs discarded, non-existent
script path so both forms only take the "check path → drain stdin → exit 0" branch):

| Wrapper form | mean | median | min | max |
|---|---|---|---|---|
| Git Bash guarded form | 66.2 ms | 65.6 | 59.3 | 74.4 |
| PowerShell `-EncodedCommand` | 207.4 ms | 205.6 | 192.0 | 227.3 |

That's **3.13×**, in line with the ~60/~180 ms figures from #7131. PreToolUse and
PostToolUse both fire on every tool call, so in practice this is roughly 640 ms vs 190 ms
of hook overhead per tool call.

The quoting already covers what the gate was added for. The fast path emits:

```ts
`if [ -f ${quotePosixShellString(bashPath)} ]; then ${quotePosixShellString(bashPath)}; else ${POSIX_HOOK_STDIN_DRAIN_COMMAND}; fi`
```

Single quotes suppress all expansion in POSIX shells, and the one character that matters
(`'`) is escaped by `quotePosixShellString`. The function already normalizes `\` → `/`
before the test runs.

I verified with Git Bash (`C:\Program Files\Git\usr\bin\bash.exe`) against a real
`claude-hook.cmd` under a CJK home path:

| Form | Result |
|---|---|
| slash + single quotes (what the fast path emits) | works — `[ -f ]` and direct execution both succeed |
| slash, unquoted | works |
| backslash, unquoted | fails (bash eats the backslashes) |

Side note: `Jane Doe` (the case asserted at installer-utils.test.ts:566-571) would also
work through the quoted fast path today, so that assertion now documents the gate rather
than a bash limitation.

### How can we reproduce it

1. Use a Windows profile whose name contains a non-ASCII character, so `os.homedir()` is
   e.g. `C:\Users\홍길동` (any accented/Cyrillic/CJK character does it)
2. Start Orca with Agent status hooks enabled
3. Look at `~/.claude/settings.json` — every Orca-managed hook command is
   `powershell.exe -NoProfile -ExecutionPolicy Bypass -EncodedCommand …` instead of the
   `if [ -f '…' ]; then …; fi` form

An ASCII-named profile on the same machine gets the fast path.

### Anything else

Suggested fix — reject only what quoting can't handle (Unicode Control category, which
also covers C1 controls and keeps oxlint's `no-control-regex` happy):

```ts
const WINDOWS_GIT_BASH_SAFE_PATH = /^\P{Cc}+$/u
```

One caution: `WINDOWS_CMD_SAFE_PATH` (installer-utils.ts:147) shouldn't get the same
treatment. That branch returns the bare path unquoted because Codex/Antigravity/Devin
spawn the hook command as `argv[0]` via CreateProcess rather than through a shell, so its
whitelist is genuinely load-bearing (#8430 / #8737). Only the Git Bash branch goes through
a shell where quoting does the work.

Caveat: I measured on a system whose ANSI code page is UTF-8 (65001) and haven't verified
CP949/CP936 defaults. If the gate is intentionally conservative about legacy code pages
I'd be glad to hear it — though the commit that introduced it (801effa68e, "Unsafe paths
with spaces or shell metacharacters") and its tests all talk about spaces and
metacharacters, not encoding.

Jeonous:
Fixes #11496.

## Summary

On Windows, a home directory containing non-ASCII characters (CJK, Cyrillic, accented
profiles) fails the `WINDOWS_GIT_BASH_SAFE_PATH` whitelist, so every managed Claude hook
falls back to the PowerShell `-EncodedCommand` launcher — roughly 640 ms vs 190 ms of
hook overhead per tool call (PreToolUse + PostToolUse both fire; measured 66.2 ms vs
207.4 ms per wrapper call, n=10, details in #11496).

Since #8430 the fast path wraps the script path in `quotePosixShellString()`, so the
whitelist's original concern (bash splitting on spaces) is already handled by the
quoting. This PR relaxes the gate to reject only control characters:

```ts
const WINDOWS_GIT_BASH_SAFE_PATH = /^\P{Cc}+$/u
```

so non-ASCII profiles — and paths with spaces or shell metacharacters — take the same
fast path as ASCII ones. User-visible effect: lower per-tool-call hook latency for
affected users; the generated command in `~/.claude/settings.json` becomes the quoted
`if [ -f '…' ]` form instead of the encoded PowerShell launcher.

`WINDOWS_CMD_SAFE_PATH` is deliberately untouched: Codex/Antigravity/Devin spawn the
hook command as `argv[0]` via CreateProcess rather than through a shell, so that
whitelist is load-bearing (#8430 / #8737). Only the Git Bash branch goes through a shell,
where the quoting provides the safety.

## Screenshots

No visual change — this only affects the generated hook command string in
`~/.claude/settings.json`.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Test changes:

- `installer-utils.test.ts` (`wrapWindowsGitBashHookCommand`): kept the ASCII
  exact-string case; `Jane Doe` (spaces) and `alice & bob` (metacharacters) now assert
  the exact quoted fast-path command instead of the PowerShell fallback; added
  exact-string cases for a non-ASCII home directory and a single-quote path (`O'Brien`);
  added a control-character case asserting the encoded-launcher fallback still applies.
- `claude/hook-service.test.ts`: the end-to-end #6078 spaces test now asserts the exact
  quoted fast-path command; removed the now-unused `WINDOWS_POWERSHELL_LAUNCHER` matcher.

`pnpm test` note (Windows 11 26200): every test file importing `installer-utils` or
`claude/hook-settings` passes (10 files, 144 tests). The full suite reports 179
environment-dependent failures in 54 files on my machine (Windows symlink privileges,
pty/daemon spawn); the only failing file intersecting this change's import graph is
`installer-utils.test.ts`, whose 3 `writeHooksJson` symlink tests fail identically on a
pristine checkout (symlink creation needs elevation/Developer Mode on Windows).

## AI Review Report

Reviewed with Claude Code. What it checked:

- Cross-platform (macOS/Linux/Windows): the change sits inside the win32-only path
  (`getManagedCommand` in `src/main/claude/hook-settings.ts`); macOS/Linux use
  `wrapPosixHookCommand`, untouched. No shortcuts, labels, or renderer behavior involved.
- SSH/remote/local: no SSH-specific logic touched;
  `remote-hook-service-installers.test.ts`, which exercises the shared wrapper path,
  passes.
- Agent/integration compatibility: only the Claude/OpenClaude Git Bash wrapper changes.
  Codex/Antigravity/Devin (`wrapWindowsCmdHookCommand`) and the POSIX wrapper are
  untouched.
- Performance: strictly widens the fast path; no path that previously took the fast path
  changes behavior.
- UI quality: not applicable (no UI change).

What it flagged, and what changed as a result: an earlier draft used
`/^[^\x00-\x1F\x7F]+$/`, which trips oxlint's `no-control-regex`; switched to the
equivalent Unicode-property form `/^\P{Cc}+$/u` (which also covers C1 controls). It also
caught that the end-to-end #6078 test still asserted the PowerShell fallback for spaced
paths and had to be updated alongside the unit tests.

## Security Audit

- Command construction: the fast path interpolates the script path only through
  `quotePosixShellString` (single quotes; embedded `'` escaped as `'\''`). Single quotes
  suppress all POSIX shell expansion, so spaces, metacharacters, and non-ASCII stay
  literal.
- Control characters are still excluded by the new gate and fall back to the encoded
  PowerShell launcher (covered by a test).
- No change to what gets executed: same managed `claude-hook.cmd`, same drain-on-missing
  behavior; only the wrapper form changes for paths the whitelist previously excluded.
- No input handling, auth, secrets, dependency, or IPC surface touched.

## Notes

- Windows-specific by construction; guarded by the existing platform check.
- Caveat: measured on an ANSI-code-page-65001 (UTF-8) system. I have not verified
  CP949/CP936 defaults; the commit that introduced the gate (801effa68e) and its tests
  indicate spaces/metacharacters were the concern rather than encoding, but flagging for
  reviewer judgment.


brennanb2025:
Thanks for this — the diagnosis is right, and the measurement in #11496 is genuinely useful. I've pushed three commits to your branch so this is landable; details below so nothing is a surprise.

**The one behavior change I had to make.** I tested the widened gate on a real Windows host (Git Bash 2.55, ANSI code page **1252**) and hit a case the quoting argument doesn't cover: bash's single quotes aren't the last parser. Running a `.cmd` makes Git Bash re-launch it through cmd.exe via `COMSPEC`, which parses the path *again* after the shell's quoting is already gone. Measured, with a home directory containing each character:

- fast path works: spaces, `'`, `$`, backtick, `#`, `~`, `@`, `{}`, `[]`, `+`, and non-ASCII (CJK, Cyrillic, accented)
- fast path exits **1 on every hook call**: `&` `^` `(` `)` `;` `,` `=` and `%VAR%`

So the gate is now `/^[^\p{Cc}&^();,=%!]+$/u` — control characters plus the set cmd.exe re-parses. (`!` is inert today but expands under a registry-enabled `DelayedExpansion`, so it's excluded too.)

**Your actual fix is intact.** CJK, Cyrillic, and accented profiles still take the fast path — confirmed on that same code-page-1252 machine, which also answers the CP949/CP936 question you flagged in your notes. Good instinct flagging it; it turned out fine.

**Tests.** I added cases that *execute* the generated command through a real shell rather than asserting the string. Every existing Git Bash assertion is `skipIf(platform !== 'win32')` and PR CI runs the vitest shards on Linux only, so the quoting contract had no executed coverage anywhere — a dropped quote would have shipped green.

Two things I deliberately left out of this PR so it stays scoped to one change:

- `WINDOWS_CMD_SAFE_PATH` is still ASCII-only, so a non-ASCII profile keeps paying the slow launcher for Codex, Antigravity, Devin, Droid, Grok, Cursor, Kimi and Command Code. #11496 is only fully fixed once that's addressed too — worth a follow-up.
- A `%` in the profile path fails the PowerShell fallback as well, so those users have broken hooks today regardless of this change. Separate bug; noting it here so it isn't lost.

One gap worth stating plainly: there's no Windows vitest job in PR CI, so the cmd.exe half of this is pinned by a hand-verified denylist rather than an automated gate.

OrcaWin:
## Windows low-spec QA — PR #11497 (Grok)

**Commit validated:** `58db93e90b4539100c01b151de70a997d1f0e735` (matches live `refs/pull/11497/head`)

### Tree vs main parent
- Base `96c17b44d51015fbf88bceb640f8b64c2a0f4d7e` and head share **identical tree** `061d521275e58e916ca40f16c4451ab5e348ac42`
- `git diff base head` is empty — PR is content-empty / superseded by main

### Superseding runtime-home hook (on main / this tree)
- Claude managed hooks use `wrapRuntimeHomeHookCommand` (`src/main/agent-hooks/runtime-home-hook-command.ts`) via `$HOME/.orca/agent-hooks/*.cmd`
- Non-ASCII Windows homes take the **fast path** (direct `.cmd`); cmd.exe metachar homes (`& ^ ( ) ; , = % !`) fall back to encoded PowerShell
- Platform selection uses `OSTYPE` case — **no `uname` process** per hook

### Focused tests (Windows)
```
pnpm exec vitest run src/main/claude/hook-service.test.ts  → 14/14 pass
pnpm exec vitest run src/main/agent-hooks/installer-utils.test.ts -t wrapRuntimeHomeHookCommand|wrapWindowsCmdHookCommand --testTimeout=30000
  → 8/8 pass including "keeps common Windows profiles on the fast path"
```
Unrelated env failures when running full installer-utils without filter: 3 symlink EPERM (no Developer Mode), 3 default 5s timeouts that pass at 30s.

### Real Git Bash path check
Executed generated command under `C:\Program Files\Git\bin\bash.exe`:
- non-ASCII home (`…/destination 国際 profile`) → exit 7 from `.cmd`, **no powershell.exe** in `set -x` trace
- ASCII home → exit 11
- metachar home (`user&name`) → exit 13 via encoded fallback
- invariants: OSTYPE case present, no uname, no absolute paths

### Electron identity (CDP 9341, isolated userData)
```json
{"name":"Orca: 58db93e90b","isDev":true,"devLabel":"pr-11497-grok-windows-qa @ 58db93e90b","devBranch":"58db93e90b","devWorktreeName":"pr-11497-grok-windows-qa","devRepoRoot":"C:\\Users\\neil\\orca\\workspaces\\orca\\pr-11497-grok-windows-qa"}
```
Navigated Settings → Agents; **Agent status hooks** toggle visible and ON. Console: 0 errors (dev CSP/tooltip warnings only).

### Screenshots (not committed)
Settings → Agents with Agent status hooks:

Settings Agents / Agent status hooks

Settings Agents full page

Gist: https://gist.github.com/OrcaWin/cda9707c6beaa9d45645e2b7e0a17b37

### Recommendation
**Close as superseded / do not land** — empty diff vs main; original fix is preserved by the runtime-home hook path already on main. No code changes made during this validation.

Note: dev Electron on this host hits `EPERM fsync` when creating a fresh profile index (`fsyncFileSync` opens `'r'` on Windows). Pre-seeding the index unblocks startup; unrelated to this PR.


**Superseding fix:** [#13442 — fix(claude): make managed hook paths portable](https://github.com/stablyai/orca/pull/13442)

brennanb2025:
Thanks for the thoughtful report, performance measurements, and implementation — the diagnosis here was correct and genuinely useful.

A broader runtime-home solution landed independently in #13442 and now covers this non-ASCII Windows-home case while preserving the safe fast path. After updating this branch from `main`, #11497 has no remaining diff, so I’m closing it as no longer necessary. Thanks again for contributing this.

PR #11500, opened 2026-07-30
fix(session): preserve agent sessions across restart

beige-ian:
## Summary

Agent sessions could be lost or relaunched incorrectly when Electron/the renderer or the PTY daemon exited between debounced persistence and startup reconnect. This change makes the main hook cache and renderer shutdown checkpoint durable at the boundaries that matter, then restores the cached provider identity before PTY cold restore.

## What changed

- Synchronously checkpoint the first live provider session identity and terminal Stop transition in the main hook server, while retaining the trailing retry.
- Validate and authenticate the synchronous renderer shutdown checkpoint IPC sender and payload before flushing durable workspace and active-view state.
- Add a startup snapshot barrier so the renderer applies the main hook cache before persisted terminals reconnect.
- Rebuild a missing pane leaf from a unique tab/worktree sleeping record or live provider status only; ambiguous candidates fail closed.
- Preserve quit/live recovery ownership when a crash removed the pane layout, preventing duplicate recovery tabs.
- Add unit coverage and Electron E2E coverage for normal quit recovery, force-exit recovery, and recovery sourced only from the main hook cache.

## Validation

- Full Vitest: 41,342 passed, 71 skipped.
- Electron E2E: 3 passed, including force-exit/restart and main-hook-cache reconstruction.
- Typecheck: passed.
- Production CLI and Electron builds: passed.
- Changed-code quality, type-aware quality, React Doctor, reliability, max-lines ratchet, skill bundle, and localization gates: passed.
- The repository-wide lint command reports one unchanged pre-existing baseline violation in `src/main/daemon/daemon-pty-router.ts` (304 lines; limit 300); no changed file introduces a new lint violation.

No release tag or production deployment is included in this PR.


AmethystLiang:
reviewing now

brennanb2025:
Thanks @beige-ian — this fixes a real and painful loss window, and the core design (durable provider-session identity at turn boundaries + applying the hook cache before terminals cold-restore + fail-closed pane adoption) held up under deep review. I've validated it end to end, hardened a few edges on the branch (fast-forward, your commits intact), and it's ready from my side.

**What's on the branch now** (`d32687af..f75a6c81`): your original head, a merge of current `main` (resolves the conflict), and five follow-up commits: adoption restricted to the tab's unique actively-resumable record, startup-snapshot fail-open + recovery-candidate identity hardening, stopped rows made authoritative, and two review-fix commits (below).

**Reproduction.** At your exact head, the two new e2e specs pass; with the src changes reverted (tests kept) both fail — the tests genuinely exercise the crash-loss bug.

**Review findings fixed on the branch** (two independent review lanes per round, fresh each round, plus a dedicated perf audit; round 2 came back with no confirmed behavioral findings):
- **Critical** — the new shutdown-checkpoint validator required `tabsByWorktree`/`terminalLayoutsByTabId` on every session partition, but host-split slices legitimately omit containers with no routed entries (merely visiting an SSH/runtime worktree mints such a slice). Result: the whole checkpoint was rejected, `beforeunload` was cancelled, and the app became permanently unquittable for those users — with the full session checkpoint lost on force-kill. Fixed by validating containers only when present and skipping (not rejecting) a malformed partition; proven against the real host-split emitter output, including a pre-fix control that reproduces the rejection.
- **High** — both rebuilt-pane adoption fallbacks matched agent rows by tab identity without checking whether the row's leaf still exists in the tab layout, so a plain split pane could adopt its live sibling's session (double resume; on remote panes it could also discard a reattachable PTY). Now any same-identity row whose leaf is still bound fails adoption closed. This also protects the live split-pane path, not just crash recovery.
- **Medium** — the adoption predicate lacked the `origin: quit|live` gate that the ownership check has, letting worktree-sleep/legacy records race the wake path into a double resume. Gates now agree.
- **Low** — the startup-barrier snapshot request now carries a requestId so a response that loses the 5s fail-open race cannot overwrite a newer post-ready snapshot.
- **Perf (medium)** — the lost-layout ownership branch put an O(records) scan ahead of the O(1) activation gate; the resume sweep amplifies that quadratically on worktree activation (measured ~937ms at 400 records). Reordered (pure-read commutation, differential-fuzzed over 694k cases, 60–75×). Sync-flush cost itself is turn-boundary-gated and measured at ~0.5–2.4ms per flush at realistic fleet sizes.
- Test hardening: the two-distinct-live-sessions ambiguity guard, the degraded-startup barrier, the origin gate, and the skip-partition path are all now pinned by tests that fail under mutation of their hunks (17 of 21 mutations killed in round 1; the survivors were either equivalent mutants or got the missing test in round 2).

**Gates at `f75a6c81`:** targeted unit battery 960/960 (9 files), typecheck ✓, lint ✓, changed-code quality gate ✓ (0 new findings incl. React Doctor), both e2e specs ✓.

**Manual Electron matrix at `f75a6c81`** (real agent CLIs, isolated profile; forced exits = SIGKILL of renderer+main+PTY daemon mid-turn):

| Cell | Scenario | Result |
|---|---|---|
| L1 | Git workspace, real Codex mid-turn, forced exit → restart | Same session resumed in the same tab, full history, no duplicate tab |
| L2 | Real Claude + Codex, normal quit → restart | Both sessions restored; Claude resumes with prior conversation |
| F1 | Folder workspace, real Claude mid-turn, forced exit → restart | Session resumed in the folder workspace |
| S1 | SSH workspace, live remote PTY, forced desktop exit → restart | Pane reattached to the same live remote shell (proven interactively); with a durable working row present it still reattaches — no spurious duplicate resume |
| S2 | Normal quit with an SSH workspace present | Quit completes cleanly, checkpoint stages all workspaces (this exact case hard-locked quit before the Critical fix), everything restores |

L1 before/after:
L1-before
L1-after
L2 before/after:
L2-before
L2-after
F1 before/after:
F1-before
F1-after
S1 before/after:
S1-before
S1-after
S2 before/after:
S2-before
S2-after

**Coverage notes.** Windows: no new platform-dependent behavior (the sync flush reuses the existing write+rename persistence; new logic is platform-neutral) — CI's OS matrix covers the unit suites. Paired mobile: unaffected — mobile consumes agent status from main's in-memory cache over the relay replay path, which this PR doesn't touch. Future (non-blocking): prefer PTY reattach over cold restore when a remote pane's own leaf still has a restorable PTY, and tighten the e2e output assertion to the full resume invocation.

Maintainers: CI on the new head may need workflow approval for the fork.

Me333-jjj:
Saw the issue with losing agent sessions on restart in Orca. My system ensures seamless state recovery for long-running agents.

brennanb2025:
Thanks for this — the failure mode it closes is one we've had reported repeatedly, and the mechanism is sound.

Heads up before anything else: **I pushed 6 commits to your branch.** That includes a merge of current `main`, because the branch was conflicting — `main` replaced the synchronous shutdown flush with staged writes plus a deadline-bounded async flush and renamed the IPC channel. It's a merge, not a rebase, so none of your commits were rewritten, but please `git pull` before you push again. Two of the six are a fix and its own revert that net to zero; the revert message records why.

What I changed:

**The immediate status checkpoint wrote synchronously on every status event, not once per identity.** This was the one blocker. `done(X) → working(X)` passes the "previous had no resumable session" test purely because `previous.state === 'done'`, so the checkpoint fired twice per agent turn. Each fire is a full `JSON.stringify` + per-entry SHA-256 + `writeFileSync` + `renameSync`, on the Electron main process. That amplifies badly on a relay reconnect, where every cached pane is re-forwarded and the frame decoder drains up to 64 frames in one synchronous turn. Measured at a realistic cache size (282 entries / 694 KB), worst case was **~1186 ms of main-process freeze and 46 MB written on a single SSH reconnect**, versus one 0.72 MB write before this PR. Now coalesced to one flush per event-loop turn (`setImmediate` + a pending flag; the quit-time flush cancels a queued one).

Worth adding to the description: the durability guarantee is now *same-turn* rather than *strictly synchronous* — 250 ms down to sub-millisecond, not to zero.

- **Re-armed the trailing retry after that flush**, a regression in my own coalescing. Deferring the flush moved it after the caller arms the 250 ms debounce, and `flushStatusPersistSync` clears that timer on entry, so a transient write failure lost its only retry.
- **The startup-barrier test only compared source-literal ordering** — deleting the `await` kept it green while fully restoring the race it exists to prevent. It now fails when the barrier is defeated, at both call sites.
- **Added a test capping sync writes per burst**, so the coalescing can't be silently undone by a later refactor.

One request, and it's a request: could the shutdown-checkpoint IPC changes (`src/main/ipc/renderer-shutdown-checkpoint.ts` plus the one line in `register-core-handlers.ts`) move to a follow-up PR? The trusted-sender check is security hardening that has no bearing on whether a session survives a crash, and the "flush durable state synchronously at shutdown" rationale in the description no longer matches `main`. It also earns its own review on risk: the preload throws when the checkpoint reports `ok !== true`, and the beforeunload guard turns that throw into `preventDefault()` — so a sender check that ever misfires doesn't drop a checkpoint, it makes the window unquittable. I couldn't reach that state, but it shouldn't be a footnote to a recovery PR. The rest — the main-process checkpoint, the startup barrier, and the rebuilt-pane adoption — is one causal chain and belongs together here.

Last thing, seen and deliberately not fixed here so it isn't lost: the pending agent-status queue evicts blind from the front at its 100-entry cap, and the snapshot loader now runs twice with replay enqueues that never de-dupe — so a distinct live row arriving between the two passes can be dropped. It's pre-existing shared machinery this PR doesn't otherwise touch, and a correct fix means changing the eviction policy (the two obvious fixes both still fail). Tracking it separately; not a blocker for you.

brennanb2025:
Standard-fix pass on the current head (`778d352e85`) — the diagnosis and the behavior here are yours and are what ships; I only swapped a couple of mechanisms for shapes this repo already has. Three commits, no behavior change, net −37 lines.

**1. Startup snapshot barrier** (`agent-status-startup-snapshot.ts`) — this is the one worth naming. The barrier hand-rolled its own promise plumbing: a `{resolve, reject}` record per request, a `settled` latch with `complete`/`fail` closures, `clearTimeout` on both paths, and a `splice` to pull the request out of the pending queue on timeout. Orca already has exactly one shape for "fail open after a deadline" — `Promise.race([pending, timedOut])` with `clearTimeout` in a `finally` — at `terminal-fit-restore.ts:27`, `ssh-reattach-model-restore.ts:22` and `terminal-paste-operation-timeout.ts:15`. Waiting for the loader to register is just another awaited promise inside that race, so the whole latch collapses into it. Also dropped the disposer's `disposed` flag; the body was already idempotent. All five of your tests pass untouched.

**2. Coalesced checkpoint** (`agent-hooks/server.ts`) — my own code, not yours. The deferred flush called `flushStatusPersistSync()`, which cancels the trailing debounce the caller had just armed, so I'd had to re-arm it afterwards to give a failed write its retry back. The standard move is not to cancel a timer that path never owned: `runStatusPersist()` directly, caller's debounce left running as the retry. One mechanism instead of two.

**3. Bound-leaf check** — also mine. Both adoption gates repeated `parsePaneKey` → null-check → `hasMatchingStablePaneLayout(...)`, which is the only reason that helper was exported. Named once as `paneKeyLeafIsBoundInLayout`; the underlying helper is private again.

Everything else in the diff is already the standard shape and I left it alone — the trusted-sender gate mirrors the three existing ones (`ipc/ui.ts`, `ipc/browser.ts`, `window/clipboard-ipc-handlers.ts`), the hand-rolled envelope guards are the right call on a synchronous quit-path handler where a zod parse would deep-clone the session, and the write-through-on-critical-transition checkpoint is the ordinary fix for "debounced write loses data on crash".

Proof each simplification is behavior-preserving, by mutation: removing the `Promise.race` reds the hang test; removing the waiter notification reds the early-request and replacement-loader tests; stubbing the shared predicate to `false` reds all three bound-leaf tests across both call sites; and reverting #2 to the cancel-without-re-arm form reds the transient-write-failure test. Typecheck, oxlint, both code-quality audits, react-doctor-changed and all 983 tests in the touched suites are green.

brennanb2025:
Merged the latest `main` in to clear the conflicts in `renderer-shutdown-checkpoint.ts`, `pty-connection.ts`, and `useIpcEvents.ts` — both sides preserved, no logic changed:

- **`renderer-shutdown-checkpoint.ts`** — took main's split type-module imports (`persisted-ui-state-types` / `workspace-session-state-types`) alongside this PR's `parseExecutionHostId` value import. The container-less host-split validation is untouched.
- **`pty-connection.ts`** — main replaced `getColdRestoreAgentResumePlatform` with `resolveAgentResumeLaunchTarget`, so that local helper is gone and this PR's `getColdRestoreAgentStatus` adoption gate sits on main's new shape.
- **`useIpcEvents.ts`** — re-applied the startup-snapshot barrier (`allowBeforeWorkspaceReady`, `loadAgentStatusSnapshot`) on top of main's new batched agent-status transaction, so the snapshot replay and live bursts now flow through `applyAgentStatusBatch` / the pane routing index while keeping the pre-ready escape.

Verification: `npm run typecheck` clean, `oxlint` (incl. react-doctor + code-quality configs) and `oxfmt` clean on all three files, and 977 test files / 11.7k tests across `src/main/ipc`, `src/main/agent-hooks`, `src/renderer/src/components/terminal-pane`, `src/renderer/src/hooks`, and `src/renderer/src/lib` pass. The only red is `src/main/ipc/pty.test.ts` (10 tests), which is byte-identical to `main` here and fails the same way on a clean `main` checkout — pre-existing and unrelated.

Jinwoo-H:
Thanks @beige-ian for the investigation and implementation. We’re closing this PR because we’re consolidating the underlying agent-session restart durability bug into the owner-maintained issue https://github.com/stablyai/orca/issues/12701. We’ll carry forward the rehydration, deferred-consumption/claim, startup-ordering, and absolute-age requirements against current main. Credit is preserved for @beige-ian and the original commit author 정훈 / kham0126. No further action is needed from you.

PR #11515, opened 2026-07-30
feat(terminal): rim panes and sidebar rows by agent state

### LINKED ISSUE #11508: [Feature]: Prominent border highlight on the waiting panel + its sidebar row when an agent needs input
### Problem or use case

When an agent (Claude Code / Codex) pauses to ask for a permission approval or user input, the only visual signal is a small yellow "?" glyph. When I'm working across many tabs, panels, and workspaces at once, that glyph is very easy to miss — I often don't notice an agent has been waiting on me until much later. cmux handles this with a prominent border color change on the waiting pane, which is far easier to catch at a glance.

### Proposed solution

Add a prominent border/outline color change for the "needs you" state (permission-pending / awaiting input), applied to **both**:

1. the **panel (terminal pane)** that is waiting, and
2. its corresponding **agent row in the sidebar**.

**Granularity:** the sidebar already lists each running agent/panel per repository as its own row. Ideally the highlight targets **only the specific agent/row that is actually waiting**, so I can tell exactly which one needs me. Highlighting the whole workspace/repo would also help and could be offered as an option, but per-agent (just the waiting panel + its sidebar row) is the ideal.

- Today the only cue is the small "?" glyph, which is easy to overlook.
- Sound / desktop notifications exist, but a persistent, at-a-glance border is better when visually scanning many panels.
- Prior art: cmux changes the pane border color when a pane needs attention.
- Ideally a **toggle in Settings**, so users who prefer the minimal glyph can keep the current behavior.

giirrr:
## What this changes

When an agent (Claude Code / Codex) pauses for a permission/question or just finished, the only cue today is a small status glyph that is easy to miss when working across many panes and workspaces. This adds a colored rim to make it obvious at a glance — on both the terminal pane and its sidebar row.

- **Amber = needs you** (agent state `waiting`/`blocked`). State-driven, so it persists until you respond. Viewing the pane does **not** clear it — you have to actually answer/approve.
- **Green = completed, unviewed.** Sourced from the existing `unreadAgentCompletionPanes` marker, so it clears automatically when you view the pane (via `useAutoAckViewedAgent`), cmux-style. Agents that finished a while ago (already viewed) never light up.

Closes #11508.

## Implementation

- `main.css`: new `--agent-rim-attention` / `--agent-rim-done` tokens (`:root` + `.dark` + `@theme inline`); green reuses `--status-success`.
- `terminal.css`: `.pane[data-agent-rim="waiting"|"done"]::after`. Drawn on `.pane::after` (not `.xterm-container::after`) so it never collides with the `data-terminal-attention` ring; `z-index` sits above that ring and below the focus-locate flash.
- `agent-pane-rim-subscriptions.ts`: new module, modeled on `terminal-pane-attention-subscriptions.ts`. Subscribes to `agentStatusByPaneKey` + `unreadAgentCompletionPanes`, toggles `data-agent-rim` on each pane, and notifies only tabs whose rim result changed.
- `TerminalPane.tsx`: wires the subscription next to the existing attention one.
- `worktree-card-compact-agent-row.tsx`: `ring-inset` on the row from the same signals (no layout shift).

## Screenshots

Light and dark, showing the pane rim (amber = needs you, green = completed/unviewed) and the sidebar rows (an older, already-viewed completion has no rim):

orca-rim-demo

> Note: rendered from the exact token values and CSS rules for a clear color reference; the rim is applied in-app by `data-agent-rim` / the row ring exactly as shown.

## Tests

- [x] `agent-pane-rim-subscriptions.test.ts` (6 cases): waiting/blocked → amber, unviewed completion → green, amber wins over green on the same pane, no-rim removes the attribute, and per-tab notify fires only when a rim result actually changes.
- [x] `pnpm typecheck`, `oxlint`, `oxfmt --check`, and the renderer build all pass. Built CSS confirms the `.ring-agent-rim-*` utilities, the `.pane[data-agent-rim]` selectors, and the light/dark tokens are emitted.

## AI code review summary

- **Cross-platform:** renderer-only CSS/React with no platform branches → identical on macOS/Windows/Linux. No Electron accelerators or path handling involved.
- **SSH/remote/local:** reads `agentStatusByPaneKey` (which includes remote panes via `connectionId`) and `unreadAgentCompletionPanes`; no rim is drawn when a pane has no status (graceful).
- **Agent/integration:** keys off the generic `AgentStatusState` (`working`/`blocked`/`waiting`/`done`), not any specific agent; git-provider / Linear / Jira neutral.
- **Performance:** one shared store subscription; notifies only the tabs whose rim result changed — same pattern and cost as the existing terminal-attention module. No hot-path work added.
- **UI quality:** token-based (no hardcoded hex in components), light + dark both defined; composes with the two existing pane rims via a separate pseudo-element; the sidebar uses `ring-inset` (no reflow).
- **Security:** pure presentation — no new IPC, network, process execution, or file access; no user input handled.

## Notes

- Desktop renderer only. The mobile app (`mobile/`) is a separate codebase and is untouched.
- `waiting` and `blocked` are collapsed into one "needs you" color; happy to split `blocked` into a distinct color if preferred.
- No settings toggle in this PR (always on, matching the small status glyph it complements); can gate it behind a setting like `experimentalTerminalAttention` if you'd rather ship it opt-in.


giirrr:
Merged current `main` in — the branch is conflict-free again. One deliberate change came with it, so flagging it rather than leaving it to be spotted in the diff.

**The rim now paints from `--agent-question` instead of its own token.**

This PR originally defined `--agent-rim-attention` (amber) for the needs-you rim. Since then `--agent-question` has become the single hue for "an agent is asking you something", consumed by `AgentStateDot`, `AgentQuestionIcon`, the terminal tab icon, `AgentKanbanCard` and the agent map, and its comment rules out the exact hue this PR had picked:

> Orange, not amber: on the map it has to stay separable from working-yellow at a glance.

The pane rim and the sidebar row are two more surfaces for the same state, so they now consume the shared token like the rest. That drops three variable definitions and leaves one place to change if the hue ever moves (26775ce).

`--agent-rim-done` is kept as-is: an unviewed completion is a distinct state and has no shared token of its own.

**One test-side change worth calling out.** Five existing suites (`WorktreeCardAgents.*`, `WorktreeList.lineage-agent-expansion-coupling`) needed `unreadAgentCompletionPanes: {}` added to their hand-written `@/store` mocks. The compact agent row is the first consumer of that field in this render path, so the mocks had simply never needed it. I added it to the mocks rather than guarding the read in the component: `terminals.ts` declares the field required and initializes it to `{}`, and `selectors.ts` already reads it unguarded, so a guard here would only mask a genuinely missing store slice.

Verified on the merge: `pnpm typecheck`, `pnpm lint`, and 2,263 tests across the sidebar and rim suites.

giirrr:
Applied the review nitpick on `agent-pane-rim-subscriptions.ts`.

The three-line header restated the filename and the `rimForAgentState` import that sits three lines above it. It is now one line keeping only what isn't derivable from the module itself — the sibling module it mirrors, and the selector that paints the result:

```ts
// Mirrors terminal-pane-attention-subscriptions.ts, keyed off agent status; terminal.css paints `.pane[data-agent-rim]`.
```

Also dropped an unrelated formatting-only change to `config/scripts/pr-e2e-gate-contract.test.mjs` that the pre-commit formatter had folded into the merge commit. The diff is back to this PR's own files.

PR #7786, opened 2026-07-08
fix(terminal): paste OS-copied files as full shell-escaped paths

hongdroid94:
## Summary

Fixes #7777 — on macOS, copying a file in Finder (`Cmd+C`) and pasting into a terminal pane inserted only the **display name** instead of the file's full path.

**Root cause:** `terminal-clipboard-paste.ts` reads clipboard *text* first. When a file is OS-copied, macOS synthesizes the file's display name as clipboard text, so the text branch won before the real file reference was ever read. Drag-and-drop already inserts shell-escaped paths — paste now does the same.

## What changed

- **`src/main/window/clipboard-file-read.ts`** (new) — the READ mirror of the existing `clipboard-file-copy.ts`. Reads the OS file references on the clipboard and returns absolute paths, cross-platform:
  - **macOS**: prefers `NSFilenamesPboardType` (a plist array of real POSIX paths — also covers multi-select). Falls back to `public.file-url`, skipping opaque `file:///.file/id=…` reference URLs that POSIX path resolution can't resolve.
  - **Windows**: `Get-Clipboard -Format FileDropList` via PowerShell (mirrors the `Set-Clipboard` write path).
  - **Linux**: `text/uri-list` / `x-special/gnome-copied-files`, desktop-dependent, mirroring the copy module.
  - Injected deps so platform branching is unit-testable; never throws.
- **`clipboard-ipc-handlers.ts`** — new `clipboard:readFilePaths` IPC (trusted-sender gated).
- **preload / api-types / web stub** — expose `readClipboardFilePaths` (web returns `[]`, no OS file refs in a browser).
- **`terminal-clipboard-paste.ts`** — reads file references *before* the text branch; when present, pastes `paths.map(p => shellEscapePath(p, targetShell)).join(' ')` (reusing the drop path escaper) and returns a new `{ status:'pasted', kind:'file-path' }`. Ordinary text/image copies fall through unchanged.
- **`terminal-paste-target-shell.ts`** (new) — resolves the paste `targetShell` by delegating to the exact same resolver drag-and-drop uses, so paste and drop quote paths identically (local, SSH, runtime).

## Testing

- `pnpm typecheck` ✅, `pnpm oxlint` (changed files) ✅
- Unit tests: `clipboard-file-read.test.ts` (macOS plist/fallback/reference-skip, Windows, Linux) + extended `terminal-clipboard-paste.test.ts` — file-path branch wins over synthesized text; empty result keeps text/image behavior.
[...]

Issue #7333, opened 2026-07-04
Store pasted terminal images in an Orca-specific temp subdirectory


tiaras12:
## Summary

When pasting an image into an Orca terminal, Orca currently writes the image
directly into the system temporary directory and inserts the generated file path
into the terminal, for example:

```text
/var/folders/.../T/orca-paste-xxxx.png
```

This works for local workflows, but it makes Docker/container workflows harder
than necessary.

## Problem

For tools running inside a Docker container, the pasted image path must also be
accessible inside the container. Since Orca writes pasted images directly into
the whole temp directory, the practical workaround is to bind mount the entire
host temp directory into the container.

On macOS this can mean exposing a path like:

```text
/var/folders/.../T/
```

That is broader than necessary. The container only needs access to Orca's pasted
image files, not everything else in the user's temp directory.

## Desired behavior

Please store pasted terminal images inside a dedicated Orca subdirectory under
the temp directory, similar to how some terminals use a dedicated paste folder.

For example:

```text
$TMPDIR/orca-paste/orca-paste-xxxx.png
```

or, in terms of the current Electron temp path:

```text
app.getPath('temp')/orca-paste/orca-paste-xxxx.png
```

Then users can mount only that subdirectory into Docker:

```bash
ORCA_PASTE_DIR="$(getconf DARWIN_USER_TEMP_DIR | sed 's:/*$::')/orca-paste"

docker run --rm -it \
  --mount "type=bind,source=$ORCA_PASTE_DIR,target=$ORCA_PASTE_DIR,readonly" \
  my-image
```

This avoids mounting the entire temp directory while keeping the pasted file path
valid inside the container.

## Current implementation

From the current implementation, local pasted images appear to be written using
Electron's temp path directly:

```ts
const tempPath = path.join(app.getPath('temp'), fileName)
```

A small change to create and use a child directory would solve this:

```ts
const pasteDir = path.join(app.getPath('temp'), 'orca-paste')
await fs.mkdir(pasteDir, { recursive: true })
const tempPath = path.join(pasteDir, fileName)
```

## Benefit

This would make Orca safer and easier to use with Docker/containerized terminal
workflows:

- only pasted image files need to be mounted
- the entire host temp directory does not need to be exposed
- the inserted image path continues to work as a normal local absolute path
- no global `TMPDIR` changes are required

PR #11516, opened 2026-07-30
fix(terminal): stop dropping pasted images when the clipboard also names them

leokwon68:
## Summary

Pasting a copied image into a terminal pane silently dropped the image whenever the clipboard also carried text that merely *named* that image — a browser "Copy Image" leaves the source URL on the clipboard, and file managers leave the file path. `pasteTerminalClipboard` read clipboard text first and returned as soon as text was non-empty, so Cmd/Ctrl+V injected the URL/path and the bitmap was discarded. Screenshot-only clipboards (no text flavor) worked fine, which made the failure look random to users.

Repro before this change (macOS, 1.4.145 and current `main`):

1. Put an image *and* a text flavor on the clipboard — e.g. `Copy Image` in a browser, or write both flavors to `NSPasteboard`.
2. Focus a terminal pane running a TUI agent and press Cmd+V.
3. The URL/path is pasted as text; no `orca-paste-*.png` temp file is ever created, so the agent never sees the picture.

After: the temp PNG is written and its path is bracket-pasted, exactly as it already works for screenshot-only clipboards.

The fix stays deliberately narrow. The image is only probed ahead of the text when the clipboard text is *nothing but* a reference to an image — a local path, a `file://` URL, or an `http(s)` URL whose last path segment has an image extension. Ordinary text keeps the untouched text fast path, which matters because rich copies from spreadsheets and slide editors ship a TIFF flavor alongside the cell text and must not turn into a PNG path. If the reference text turns out to have no bitmap behind it, the paste falls back to the text, so no existing paste gets worse.

## Screenshots

No visual change — terminal paste behavior only.

## Testing

- [x] `pnpm lint` — clean on the touched files (`oxlint` exit 0); the full run fails only on a pre-existing `max-lines` violation in `src/main/daemon/daemon-pty-router.ts:381` (304 > 300) that fails identically on unmodified `main`
- [x] `pnpm typecheck` — clean (all three projects)
- [x] `pnpm test` — the touched area is green: 34/34 in the two clipboard-paste files, 2842/2842 across `src/renderer/src/components/terminal-pane`. The full suite on this machine flakes under parallel load in unrelated main-process/relay/ssh/git files (8 files / 85 tests on this branch vs 9 files / 87 tests on unmodified `main`, overlapping but not identical sets; the ones I re-ran in isolation pass). No renderer or terminal-pane failures on either side.
- [x] `pnpm build`

New tests cover the bug directly: image-plus-source-URL and image-plus-file-path clipboards must bracket-paste the temp image path; a reference-text clipboard with no bitmap must fall back to pasting the text; and a tab/newline-separated spreadsheet copy must stay on the text fast path without probing for an image. The pure `isImageReferenceOnlyClipboardText` helper has its own table of accept/reject cases, including Windows drive paths, UNC paths, percent-encoded URLs, `data:` URLs, non-image URLs, prose ending in `screenshot.png`, and oversized input.

## AI Review Report

Reviewed with an adversarial Codex pass told to attack rather than approve; it probed ~15 inputs outside the test suite (RTLO spoofing, control bytes, one-letter Windows-drive vs URL-scheme collision, double percent-encoding, `data:`/`javascript:`/`ftp:` schemes, query-string and fragment extensions, whitespace-padded huge strings) and scored the change 8/10 with two real findings, both now fixed in this branch:

1. *Swallowed error (medium).* The first draft returned the text fallback before calling `onImagePasteError`, so a genuine failure (permission denied, disk full) looked identical to "no image on the clipboard". The callback now fires before the fallback, and the test asserts it.
2. *Length gate placement (low).* The bound was applied after `trim()`, so a whitespace-padded clipboard still forced a full linear pass and contradicted the doc comment. A raw-length gate now runs first, with tests for both padded and oversized input.

It also flagged that relative paths and extensions living only in a URL query string or fragment stay on the text path; that is intentional (over-matching prose is the worse failure) and is now documented in the module comment and locked by a test.

Checked and confirmed:

- **Cross-platform.** POSIX separators, Windows drive paths (`C:\...`), and UNC paths (`\\server\share\...`) are matched by shape before URL parsing — deliberately, because a one-letter Windows drive prefix also parses as a URL scheme, so URL parsing must not claim it. Extension matching reuses the existing `isImageDropPath` helper, so it stays case-insensitive and ignores dots in directory components. No new shortcuts, labels, or platform-specific code paths.
- **Remote/SSH/web.** `connectionId` and `runtimeEnvironmentId` are still forwarded to `saveClipboardImageAsTempFile` unchanged, so SSH panes keep saving the bitmap on the remote host. The web client path (where `readClipboardText` can reject) is untouched; a rejected text read still falls through to the image probe.
- **Agent neutrality.** No agent-specific logic; the temp path goes out through the same bracketed-paste channel every supported TUI already consumes.
- **Performance.** The text fast path still performs zero image probes for ordinary text — the pre-existing tests asserting that are unchanged and still pass. The new check is a bounded prefix/extension test on at most 2048 characters, with no backtracking-prone regex.
- **UI.** No visual change.

## Security Audit

The new code parses untrusted clipboard text, so that was the focus.

- **ReDoS / unbounded work:** the two regexes are anchored and backtracking-free (`^(?:~\/|\/|\\\\|[A-Za-z]:[\\/])` and a `[\r\n\t]` character class); input is length-capped at 2048 characters before any scan, and extension matching is a `lastIndexOf` + `Set` lookup rather than a pattern match.
- **Injection:** this PR does not change what is written to the PTY. The pasted value is a path generated by the main process (`orca-paste-<ts>-<uuid>.png`), sent through the existing `forceBracketedPaste` channel; clipboard-supplied text is never promoted to a command, and no new shell interpolation or path joining is introduced.
- **Path handling:** clipboard text is only classified, never opened, resolved, or written; there is no new filesystem access, so no traversal surface.
- **IPC / secrets / deps:** no IPC surface, no credentials, no new dependencies.

Measured, not assumed: the classifier returns in ~0.2µs even on a 1 MB string, because the raw-length gate rejects before any trim or scan, and control-byte, RTLO (`‮`), `data:`, and `javascript:` inputs were probed directly. When the classifier says "reference", the value that reaches the PTY is either the main-process-generated `orca-paste-<ts>-<uuid>.png` path or — if no bitmap materializes — byte-identical to the text the pre-patch code already pasted, so there is no new execute-vs-quote surface.

One behavior change worth calling out: when the clipboard text is an image reference and image extraction *fails*, `onImagePasteError` is still reported and the paste then falls back to that text, so a broken temp-file write stays visible instead of looking like an empty clipboard.

## Notes

- Only `src/renderer/src/components/terminal-pane/` is touched; the native-chat composer already prefers the image on mixed clipboards, so this brings terminal paste in line with it.
- Verified by hand on macOS 15 (Apple silicon) against a live Claude Code TUI pane: with a mixed image+URL clipboard the paste now produces the temp PNG path and the agent reads the picture; the same clipboard pasted a bare URL before the change.
- No X handle to shout out — happy to be credited by GitHub username.
- Related open items this does not address: #7333 (move pasted temp images into an Orca-specific temp subdirectory) and #7786 (paste OS-copied *files* as shell-escaped paths — a Finder file copy with no bitmap flavor still pastes nothing).

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-clipboard-image-reference-text.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-clipboard-image-reference-text.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from 'vitest'
+
+import { isImageReferenceOnlyClipboardText } from './terminal-clipboard-image-reference-text'
+
+describe('image-reference-only clipboard text', () => {
+  it('recognizes the source URL a browser "Copy Image" leaves alongside the bitmap', () => {
+    expect(isImageReferenceOnlyClipboardText('https://example.com/assets/screenshot.png')).toBe(
+      true
+    )
+    expect(isImageReferenceOnlyClipboardText('http://example.com/a/b/photo.JPG?v=2')).toBe(true)
+  })
+
+  it('recognizes percent-encoded image URLs', () => {
+    expect(isImageReferenceOnlyClipboardText('https://example.com/my%20shot.png')).toBe(true)
+  })
+
+  it('recognizes local and file:// image paths, including names with spaces', () => {
+    expect(isImageReferenceOnlyClipboardText('/Users/jane/Desktop/My Shot.png')).toBe(true)
+    expect(isImageReferenceOnlyClipboardText('~/Pictures/cat.webp')).toBe(true)
+    expect(isImageReferenceOnlyClipboardText('file:///Users/jane/Desktop/shot.png')).toBe(true)
+    expect(isImageReferenceOnlyClipboardText('C:\\Users\\jane\\shot.png')).toBe(true)
+    expect(isImageReferenceOnlyClipboardText('\\\\share\\team\\shot.png')).toBe(true)
+  })
+
+  it('tolerates surrounding whitespace', () => {
+    expect(isImageReferenceOnlyClipboardText('  /tmp/shot.png\n')).toBe(true)
+  })
+
+  it('rejects ordinary text so the text fast path keeps winning', () => {
+    expect(isImageReferenceOnlyClipboardText('hello')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('   ')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('line one\nline two')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('a69ce28e1d092e0c8825cd1a109ac36409962bc1')).toBe(
+      false
+    )
+  })
+
+  it('rejects prose that merely ends in an image filename', () => {
+    expect(isImageReferenceOnlyClipboardText('see screenshot.png')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('shot.png')).toBe(false)
+  })
+
+  it('rejects non-image references', () => {
+    expect(isImageReferenceOnlyClipboardText('https://example.com/report.pdf')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('https://example.com/')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('/Users/jane/notes.txt')).toBe(false)
+    expect(isImageReferenceOnlyClipboardText('data:image/png;base64,iVBORw0KGgo=')).toBe(false)
+  })
+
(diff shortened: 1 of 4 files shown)


leokwon68:
Renamed to `isImageReferenceText` in 7c1eee5 — the flag did read like a string next to `pasteClipboardText()`. No behavior change; the two clipboard-paste test files still pass 34/34 and `tsc -p config/tsconfig.tc.web.json` is clean.

PR #11523, opened 2026-07-30
feat(native-chat): render omp transcripts

plotarmordev:
## Summary

Orca already ships `omp` (pi-agent) as a first-class launchable agent — it appears in the agent list and resumes by `session_id` — but its transcripts have no decoder. `resolveNativeChatTranscriptAgent` returns `null` for it, so `isNativeChatSupportedAgent('omp')` is false and the native chat view never opens. **The agent runs; the conversation stays a raw terminal.**

This adds `decodeOmpTranscriptLine` and wires it through the existing path (barrel, reader, tail reader, session-file resolver, shared allowlist), so `omp` renders as chat like `claude`, `codex` and `grok`.

omp writes one envelope per line, `{ type, id, parentId, timestamp, … }`. Conversation turns are `type: 'message'`; everything else is session bookkeeping. Order is file order, so `parentId` is not consulted.

**Reasoning follows Claude, not Codex.** omp carries reasoning as a `thinking` *content block* inside the assistant turn, exactly like Claude — so it becomes a text block on an assistant message. Codex and Grok emit a separate `reasoning` role only because their transcripts carry dedicated reasoning *records*. A thinking-only turn therefore stays `assistant`.

| omp | native chat |
|---|---|
| `toolCall` | `tool-call`; `arguments` passed through as the object omp writes, mirroring Claude's `input: record.input` |
| `toolResult` | `tool` role, `isError` preserved |
| `developer` | `system`, matching the Codex non-user/non-assistant fallback |
| `image` (`blob:sha256:…`) | dropped — a content-addressed handle is neither path nor url, and the Claude mapper drops a sourceless image the same way |
| `compaction`, `mode_change`, `session_init`, `custom`, unknown | skipped, never thrown |

Session files are `<ISO timestamp>_<session id>.jsonl` under a per-cwd directory, so the resolver matches the id as a base-name suffix the way Codex rollout files are matched. It resolves the root through `normalizeAgentSessionsDir` and honours `OMP_CODING_AGENT_DIR`, matching `session-scanner-source-discovery.ts` — without that, a user with the override set gets sessions in AI Vault but a dead native chat view.

## Screenshots

No visual change — this populates the existing native chat view for one more agent; no UI was added or modified.

## Testing

- [x] `pnpm lint` — passes for all files in this PR. One pre-existing failure unrelated to it: `src/main/daemon/daemon-pty-router.ts:381 max-lines (304 > 300)`, which reproduces identically on pristine `origin/main` and is not touched here.
- [x] `pnpm typecheck` — clean (`tsconfig.node`, `tsconfig.tc.cli`, `tsconfig.tc.web`).
- [x] `pnpm test` — `src/main/native-chat/` + `src/shared/`: **3,602 passed, 16 skipped, 0 failed**. Full-repo `pnpm test` additionally fails `daemon`/`browser`/`automations` suites in this environment because `node-pty` has no `prebuilds/linux-arm64/pty.node`; those failures reproduce on pristine `origin/main` and are unrelated.
- [ ] `pnpm build` — **could not complete in this environment**, for a reason unrelated to the diff: `node-pty@1.1.0` ships prebuilds only for `darwin-arm64/x64` and `win32-arm64/x64`, so on aarch64 Linux it must compile from source, and these deps were installed with `--ignore-scripts`. The failure is the `plain-node-entry-guard` step loading `daemon-entry.js` (`Failed to load native module: pty.node … prebuilds/linux-arm64`), which this PR does not touch. `pnpm build:web` does complete cleanly, and `pnpm run build:electron-vite` gets as far as that guard with no errors attributable to these files.
- [x] **Tests added.** 12 decoder cases (both roles, thinking-only turns, tool results with and without `isError`, `developer` channel, blob-image drop, id fallback, bookkeeping and unknown types skipping) plus 2 resolver cases (suffix match, `OMP_CODING_AGENT_DIR` override) and allowlist coverage.

Additionally verified against **94,603 lines of real omp transcripts** across four sessions:

```
decoded 50,546   malformed 0   thrown 0
roles: assistant 24,716 · tool 25,280 · user 534 · system 16
tool-call 25,284 / tool-result 25,280   (gap = calls still open at session end)
```

That corpus check is not committed — it depends on private session data.

## AI Review Report

Two independent AI review passes were run against the sibling decoders as the style bar, plus CodeRabbit and Greptile on the PR itself.

**Risks checked:** consistency with the `claude`/`codex`/`grok` decoders; invented behaviour with no precedent; comments that narrate rather than justify; dead or speculative code; test coverage of the resolver and of every record type.

**Flagged and fixed:**
- A comment copied from the Claude mapper claimed the role marks reasoning — **false here**, and contradicted by this PR's own test. Removed.
- `ompSessionsDir()` ignored `OMP_CODING_AGENT_DIR` while the AI Vault scanner honours it — a real behaviour gap, now routed through `normalizeAgentSessionsDir`.
- `resolveOmpSessionFile` had no test. Added, plus one for the env override.
- A `ompToolCallInput` helper existed to repair "Python-style reprs". Measuring the raw transcripts showed `arguments` is always already a JSON object; the belief was an artifact of the sampling tool. **Helper deleted**, now `input: record.arguments` like Claude.
- An `existsSync` guard that could not fire (`walkSessionFiles` already returns `[]` on a failed readdir). Removed.
- Three behaviours with no precedent in the siblings — a `reasoning` role for thinking-only turns, `compaction` surfaced as a `system` message, and a `turnId` no other decoder sets — all removed in favour of matching Claude and Codex exactly.
- Resolver helper reordered to match the dispatch chain; duplicated comments removed.

The decoder went from 101 to 65 lines as a result.

**Cross-platform:** reviewed for macOS, Linux and Windows. All path handling uses `node:path` (`join`, `basename`, `extname`) and `node:os` `homedir()`; no string concatenation of separators, no shell invocation, no platform-conditional logic, no Electron API use. The resolver is pure path matching over a directory walk, so behaviour is identical on all three. No shortcuts, labels, or UI strings are touched.

## Security Audit

- **Input handling** — the decoder parses untrusted JSONL. Every record goes through `parseJsonObject`, which catches parse errors and returns null; unknown record and block types return null rather than throwing, so a malformed or hostile line degrades to a skipped record and cannot fail the whole read. No `eval`, no dynamic `require`, no prototype-mutating merges.
- **Path handling** — the session id is used only as a suffix comparison against base names produced by the directory walk (`name.endsWith('_' + sessionId)`). It is never concatenated into a path, so a crafted id cannot traverse outside the sessions root. The root itself comes from `normalizeAgentSessionsDir`, the same helper the existing scanner uses.
- **Command execution** — none added.
- **Auth / secrets** — none touched. Transcripts may contain user content but nothing is logged or transmitted; this PR only maps records into the existing in-process chat model.
- **IPC** — no new IPC surface; the decoder runs in main and returns the same `NativeChatMessage` shape the other agents already produce.

**Follow-up:** none identified.

## Notes

- **No interruption path.** Claude and Codex surface aborts as a `NATIVE_CHAT_INTERRUPTED_STATUS_TEXT` system message. omp records no interruption or abort event — I checked the corpus for abort/cancel/interrupt markers and found only prose — so there is no equivalent here. If omp adds one, `decodeOmpTranscriptLine` is where it would hook in.
- **Compaction is skipped, not surfaced.** This matches how the Claude decoder drops compact summaries. It does mean that on a heavily-compacted session the rendered history begins at the newest compaction. Surfacing a summary line would be a reasonable follow-up, but it would be new behaviour for all agents rather than something omp should do alone.
- **Blob images are dropped.** omp stores images as `blob:sha256:…` handles rather than a path or URL. Rendering them would need a blob-resolution step that does not exist yet for any agent.


plotarmordev:
@coderabbitai review

Docstrings were added in 633baa2 after the last pass; re-running so the coverage check reflects them.

brennanb2025:
Thanks for this — the direction is right, and following the `claude`/`codex`/`grok` decoder shape made it straightforward to review. I've merged latest `main` in (the branch was well behind) and pushed review fixes on top.

**The theme:** omp writes several record types with **no `content` field**. Those decoded to zero blocks and disappeared from the chat view entirely — silently, so nothing in the transcript looked wrong.

- `bashExecution` / `pythonExecution` — TUI `!command` runs vanished; now a tool turn
- `fileMention` — `@path` attachments vanished; now listed by path (deliberately not `files[].content`, which is an auto-read dump that would bury the conversation)
- `custom_message`, and legacy `custom`/`hookMessage` rows — now decoded, gated on `display` the way omp's own renderer gates them

**Other fixes:**

- `stopReason: 'aborted'` turns now surface as the interrupted row, matching the Claude and Codex decoders. Worth flagging: the description says omp records no interruption or abort event — it does, the corpus just never contained one. An abort carrying partial content keeps that content.
- Cancelled command cells now read as errored. Every omp cancel path emits `exitCode: undefined`, which `JSON.stringify` drops, so an `exitCode !== 0` check reads a cancelled run as a clean success.
- omp now joins Grok in requiring a locally readable transcript. Its hook reports no transcript path, so under Model-A SSH the chat view opened against a disk the reading process can't reach and never loaded. This applied on mobile too, which shares the same allowlist — that surface was enabled silently by adding `omp` to it.
- The session-file walk now prunes omp's per-session subagent artifact directories, matching the AI Vault scanner. It was returning a *subagent* transcript instead of the parent session on a delegating session, and cost a full recursive readdir on every resolve (~10,000 directory reads on a heavy tree, now ~50).

**On the corpus check** — 94k lines is good evidence, but it has a blind spot worth naming: it can only cover record types your sessions happened to produce. Every gap above only appears if you run a `!command`, use `@file`, interrupt a turn, or resume a pre-v3 session. A couple of the mappings also looked wrong when read from omp's *serializer* and only came out right when read from its *renderer* — the two disagree, and the renderer is what native chat mirrors.

Verified against a real omp 17.0.5 session driven end-to-end inside the app, plus revert-checks on every fix (each guarding test goes red when the behavior is removed).

**Deliberately left, all consistent with the other decoders:** errored assistant turns and abandoned `parentId` branches render the same way for Claude and Codex, so fixing them for omp alone would introduce exactly the inconsistency this PR otherwise avoids.

**Follow-ups, not blocking:** omp's hook could disclose `session_file` the way pi's does — that would drop the directory scan on every chat open and pagination page, and let the SSH gate narrow back to Grok only. Separately, unknown record types could render a visible placeholder instead of being dropped, so the next omp record type surfaces instead of silently disappearing.

brennanb2025 (approved):
Approving. Verified the decoder against real omp 17.0.5 records rather than fixtures alone — including a cancelled bash cell (which omp writes with no `exitCode` key at all), an interrupted turn carrying partial content, and a `fileMention`. CI green across desktop and mobile.

PR #11530, opened 2026-07-30
fix(sidebar): stop background workspace creation from scrolling the sidebar

nwparker:
## Description

Follow-up to #11445. That PR stopped background workspace creation from stealing the user's **selection**. The sidebar still **scrolled by itself**, because the scroll comes from a different signal.

Adopting a terminal and pointing the user at it were the same decision. This separates them with a `surfaceOwner` flag, and every background path now adopts silently.

## ELI5

Creating a workspace in the background should be like a coworker quietly filing a new folder in the cabinet behind you. Orca was tapping you on the shoulder and turning your chair toward the cabinet — not far enough to change what you were working on, but far enough to lose your place in the list.

Under the hood: a background create still has to start that workspace's terminal. When Orca handed that terminal to the UI it said nothing about how to present it, and "said nothing" was read as "show it to the user" — so the sidebar scrolled. #11445 stopped the chair from turning all the way; this removes the shoulder-tap.

Now a background create explicitly says "adopt this terminal silently": the tab is still created and the agent still runs, but the sidebar stays where you left it. Terminals you actually ask for still reveal themselves as before.

## Root cause

A background create still spawns the new workspace's terminals via `OrcaRuntimeService.createTerminal`, which asks the renderer to adopt the PTY through `notifier.revealTerminalSession(...)`.

No worktree-create call site passed a `presentation`, so `resolveTerminalPresentation` returned `undefined`. Every downstream check is `presentation !== 'background'` — **true for `undefined`** — and the renderer mirrored it:

```ts
const shouldSurfaceOwner = terminalPresentation !== 'background' // undefined -> true
if (shouldSurfaceOwner) {
  store.revealWorktreeInSidebar(worktreeId) // scrolls the sidebar, expands collapsed groups
  focusTerminalInitiatedTab(tab.id)
}
```

So "no presentation stated" meant "point the user at this terminal". That default is right for `orca terminal create`, where discoverability is the entire point — the runtime even returns a warning when the reveal fails. It is wrong for terminals the user never asked for.

`presentation: 'background'` could not express this: it means "do not adopt the terminal in the renderer at all", which would leave the new workspace with no tab for its agent.

### Version boundary

`shouldSurfaceOwner` arrived in #6686 (Jun 28), which decoupled "surface the owner" from "activate". Before it the reveal was gated on activation, so a non-activating create could not scroll the sidebar. #11445 fixed the activation half; this fixes the surfacing half.

## Design

`surfaceOwner` is typed `?: false`, not `?: boolean` — absent means "surface it", so the flag is a suppression switch and `true` is unrepresentable. One way to express each intent.

`surfaceOwner: false` is passed from every background path: the agent/startup terminal, the plain initial shell, the Setup tab (both new-tab **and** split-pane placement), default tabs, and orchestration worker dispatch. It is threaded through `createTerminal` **and** `splitTerminal`. Explicitly activated creates omit it.

No new CLI/RPC surface: `surfaceOwner` is main-internal and absent from the `terminal.create` RPC schema, so callers cannot set it.

## Proof of regression

Tests were written against the fix, then run with the production change reverted.

Main process — reverting only the `surfaceOwner: false` arguments at the worktree-create call sites, leaving the forwarding machinery intact:

```
Failed Tests 5
FAIL orca-runtime.test.ts > follows normal setup policy for CLI-created worktrees without activating them
FAIL orca-runtime.test.ts > creates the first terminal for CLI-created worktrees without activating them
FAIL orca-runtime.test.ts > does not surface the new workspace when a background CLI create launches its agent
FAIL orchestration.test.ts > starts a fresh agent in the coordinator current worktree
FAIL orchestration.test.ts > starts a fresh agent in an exact existing worktree without replaying setup
      Tests  5 failed | 1081 passed | 1 skipped (1087)
```

The no-regression pair bites in the opposite direction — replacing `...ownerSurfacing(shouldActivate)` with an unconditional `surfaceOwner: false`:

```
FAIL orca-runtime.test.ts > still surfaces the new workspace when the caller explicitly activates it
FAIL orca-runtime.test.ts > still surfaces the new workspace when hooks run in the foreground
      Tests  2 failed | 935 passed | 1 skipped (938)
```

Split-mode Setup (the review catch below) is wired to **both** halves of its fix independently — removing either the `...surfacing` spread in the `shouldSplitSetup` branch or the `...ownerSurfacing(...)` in `splitPtyBackedTerminal` produces:

```
FAIL orca-runtime.test.ts > does not surface the new workspace when a background create splits its setup pane
- "surfaceOwner": false,
```

Renderer — reverting each handler's guard fails only its own test, so both code paths are independently covered:

```
FAIL useIpcEvents.test.ts > adopts a background workspace terminal without scrolling the sidebar to it
FAIL useIpcEvents.test.ts > replies to a requested terminal create without scrolling the sidebar when surfaceOwner is false
Received: 1st vi.fn() call: Array [ "wt-2" ]
```

## Proof of fix

- Full unit suite: **41,413 passed**, 72 skipped (3,920 files passed, 10 skipped).
  - The only failure, `src/relay/agent-exec-handler.test.ts` (2 tests), reproduces identically on a clean tree at `37af4577` and is unrelated.
- `pnpm run typecheck:node`, `pnpm run typecheck:web`: passed.
- `pnpm run check:code-quality:changed`: 0 new findings across 11 files.
- `pnpm run check:max-lines-ratchet`, `pnpm run verify:bundled-skill-guides`: passed.
- `oxlint` + `oxfmt --check` clean on all changed files.

## Proof of no regression

- `orca terminal create` still reveals its workspace — no `surfaceOwner` in its payload, and the pre-existing assertion is untouched. Each new renderer test also runs a control phase asserting the default still reveals.
- Explicit `--activate` and legacy `--run-hooks` still reveal, in both new-tab and split Setup placement; four tests assert `surfaceOwner` is absent from the reveal payload.
- Terminals are still **adopted** in every background case — the new tests assert `createTab` is called with the pre-minted tab id, so the agent tab exists when the user opens the workspace.
- `presentation: 'background'` semantics are untouched; `surfaceOwner` is a separate axis.
- The renderer no longer calls `focusTerminalInitiatedTab` for background creates, removing a second latent focus-steal on a workspace the user is not viewing.

## Review notes

- **Split-mode Setup** (caught in review): with `setupScriptLaunchMode: split-vertical|split-horizontal` the Setup terminal goes through `splitTerminal`, whose reveal payload had no `surfaceOwner` — a real gap, now fixed and tested. The non-PTY split path is unaffected: `notifier.splitTerminal` only dispatches a renderer `CustomEvent` and never reveals.
- **`notifier.createTerminal`** was also flagged, but it is unreachable — `grep -rn "notifier\.createTerminal" src/main` (excluding tests) returns nothing. The worker path calls the `OrcaRuntimeService.createTerminal` *method*, which routes through `revealTerminalSession`. Threading the flag there would mean widening the notifier type for a path nothing calls.

## Known second cause (not in this PR)

While investigating I reproduced a **second, independent** cause of sidebar self-scrolling: creating a *child* worktree makes `buildRenderableRows` collapse the parent and its children into one `lineage-group` row, which rewrites the parent's scroll-anchor key. If that parent was the top-visible row, `runVirtualizedScrollAnchorRestore` cannot find the old key and pins a *different* row to the top — a measured **+180 px** jump with no reveal involved. It is long-standing rather than RC-new, needs the anchored row to be exactly that parent, and is being handled separately.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -36064,10 +36064,13 @@ describe('OrcaRuntimeService', () => {
       (spawn.mock.calls[1]?.[0] as { env?: Record<string, string> } | undefined)?.env ?? {}
     expectStablePaneKeyEnv(setupSpawnEnv)
     const setupLeafId = setupSpawnEnv.ORCA_PANE_KEY.slice(`${setupSpawnEnv.ORCA_TAB_ID}:`.length)
+    // Why: a background CLI create adopts its tabs silently — surfaceOwner:false
+    // keeps the sidebar from scrolling to a workspace the user never asked for.
     expect(revealTerminalSession).toHaveBeenLastCalledWith(result.worktree.id, {
       ptyId: 'pty-setup',
       title: 'Setup',
       activate: false,
+      surfaceOwner: false,
       tabId: setupSpawnEnv.ORCA_TAB_ID,
       leafId: setupLeafId
     })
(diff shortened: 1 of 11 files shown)


nwparker:
Good catch — confirmed and fixed.

`splitPtyBackedTerminal` calls `revealTerminalSession` with `activate: opts.activate !== false` but no `surfaceOwner`, so with `setupScriptLaunchMode: split-vertical|split-horizontal` a background workspace create would still scroll the sidebar to its Setup terminal. The non-PTY split path is unaffected — `notifier.splitTerminal` only dispatches a renderer `CustomEvent` and never reveals.

Threaded `surfaceOwner?: false` through both `splitTerminal` and `splitPtyBackedTerminal`, forwarded it into the reveal call, and passed it from the `shouldSplitSetup` branch. Adding a regression test for the split-mode configuration specifically, since it is a separate code path from the `createTerminal` one already covered.

nwparker:
On the `notifier.createTerminal` callback (attach-main-window-services.ts:339-347) — I checked, and that path is unreachable, so I am leaving it alone.

`RuntimeNotifier.createTerminal` is declared (orca-runtime.ts:1693) and implemented, but has no live caller in main:

```
$ grep -rn "notifier?\.createTerminal\|notifier\.createTerminal" src/main --include=*.ts | grep -v "\.test\."
(no results)
```

The worker path calls `runtime.createTerminal(...)` — the `OrcaRuntimeService` **method**, not the notifier callback — and that method routes to the renderer through `revealTerminalSession`, which does forward `surfaceOwner`. So `surfaceOwner: false` cannot reach the notifier callback. Threading the flag into dead code would mean also widening the notifier type for a path nothing calls, so I would rather not add that plumbing.

The split-mode Setup gap from the other comment was real and is fixed, with a regression test.

Issue #11189, opened 2026-07-28
[Bug]: Cold-parking unmounts remotely-subscribed terminal panes, making remote-runtime reconnect unrecoverable


Fartown:
**Operating system:** macOS
**Orca version:** 1.4.159 (host), 1.4.159 (client)

## Short summary

When Mac A runs an Orca runtime and Mac B connects to it as a remote runtime, terminal tabs on B disconnect repeatedly with the `Reconnecting to remote runtime` banner, and reconnecting never recovers — it ends at `Remote runtime disconnected`.

Root cause appears to be that the cold-parking policy on the **host** only considers *local* visibility. A tab that a remote client is actively watching is, from the host renderer's point of view, just a hidden tab — so it gets unmounted 30s later, killing the stream. Reconnect re-mounts it, but because `hiddenSinceMs` is never reset for a background-mounted tab, it is immediately eligible for parking again and gets unmounted within the next render pass. That makes the loop unrecoverable rather than merely flaky.

## Steps to reproduce

1. On host A, open an Orca window with more than `TERMINAL_TAB_HOT_RETAIN_LIMIT` (12) terminal tabs across several worktrees.
2. Leave A showing one worktree/tab (the rest are hidden, which is the normal state when nobody is sitting at A).
3. From client B, pair with A's runtime and open several of A's terminals.
4. Use the terminals from B only. Do not touch A.

Observed: terminals on B repeatedly show `Reconnecting to remote runtime`; the `Reconnect` button does not help. Switching to A and making the tab visible immediately fixes that tab.

## Why I think this is what's happening

Remote subscribe asks the host renderer to mount the tab — `src/main/runtime/rpc/methods/terminal.ts`:

```ts
rendererMountRequestedBeforePty = runtime.requestRendererTerminalTabMount(params.terminal)
```

which lands in `useIpcEvents.ts` as a background mount (`STA-1840`). But in `TerminalPaneOverlayLayer.tsx`, parking takes precedence over that background mount:

```tsx
.filter((terminalTab) => shouldMountBackgroundWorktreeTab(backgroundMountTabIds, terminalTab.id))
.map((terminalTab) => {
    // ...
    if (parkedTerminalTabIds.has(terminalTab.id)) {
      return null
    }
```

And neither of the two places that decide parking knows about remote subscribers:

`terminal-hidden-view-parking.ts` — `canParkTerminalTabRenderer` exemptions:

```ts
[...]

PR #11542, opened 2026-07-30
fix(remote-runtime): materialize the host surface when reconnecting a terminal pane

### LINKED ISSUE #11541: [Bug]: Remote-runtime terminal reconnect can never recover a host surface whose PTY died — reconnect polls session.tabs.list but only session.tabs.activate materializes it
**Operating system:** macOS (host + client)<br>**Orca version:** 1.4.160 (host), 1.4.160 (client)

## Short summary

When a desktop client is paired to another Orca as a remote runtime, a terminal pane whose **host-side PTY is gone** (host app restarted / updated / rebooted, or that workspace was never opened in the host's current session) can never reconnect. The `Reconnect` button spins for 60s and returns to `Remote runtime disconnected`. The only fix is to walk over to the host, open that project, and look at it — then the client attaches instantly.

Root cause: the reconnect path polls `session.tabs.list` only, and **only** `session.tabs.activate` **materializes a pending host surface's PTY**. So the reconnect waits forever for a `status: 'ready'` handle that nothing on the host will ever mint.

This is *not* [#11189](<https://linear.app/stably/issue/STA-2854/bug-cold-parking-unmounts-remotely-subscribed-terminal-panes-making>) (host cold-parking unmounts a pane whose PTY is still alive) and not [#11305](<https://linear.app/stably/issue/STA-2894/bug-remote-pty-recovery-budget-fits-only-3-attempts-backoff-schedule>) (recovery backoff overruns its deadline). Here the host publishes the surface with `status: 'pending-handle'`, `terminal: null` — there is no PTY to attach to, and no code path asks the host to make one.

## Steps to reproduce

1. Host A runs an Orca window (desktop, `authoritativeWindowId != null`) with project P and some terminal tabs in it.
2. Client B pairs with A's runtime (`scope: runtime`) and works in P's terminals.
3. Restart Orca on A (auto-update, quit/relaunch, or reboot). A comes back with only its last-active workspace restored; P's tabs are restored as tabs but their PTYs are **not** respawned (`pendingActivationSpawn`, spawn happens on activation).
4. Go back to B and press `Reconnect` on the pane (or wait for the automatic retries).

Observed: `Reconnecting to remote runtime` for 1 min → `Remote runtime disconnected`. Pressing `Reconnect` again repeats the same minute. `session.tabs.list` on the host keeps returning the surface as `status: 'pending-handle'`, `terminal: null`.

Expected: the client reconnects on its own, the way it does on a first attach.

Workaround: on host A, open project P (activate the workspace). The renderer mounts the tab, consumes `pendingActivationSpawn`, spawns the PTY, publishes the graph, the runtime mints a handle, and B attaches within one poll.

The same dead end happens without any restart for a workspace the host window never opened in this session: the surface exists in the persisted session, so it is published, but no PTY was ever spawned for it.

## Analysis

Handle issuance requires a live PTY (`orca-runtime.ts`, `toMobileSessionTabsResult`):

```ts
const liveLeaf = leaf?.ptyId && leaf.connected ? leaf : null           // renderer graph
const pty = liveLeaf ? null : this.findPtyForMobileTerminalTab(...)    // live pty record
...(terminalHandle
  ? { status: 'ready', terminal: terminalHandle }
  : { status: 'pending-handle', terminal: null })
```

`leaves` come from the host renderer's `runtime:syncWindowGraph`, which only publishes mounted tabs (or unmounted leaves that still hold an eager PTY buffer). After a host restart neither exists for a non-activated workspace, so every surface in it is `pending-handle`.

The runtime already knows how to fix this — `activateMobileSessionTab()` materializes a pending surface for the exact `tabId`/`leafId` when the caller is a paired client (`!targetsHost`):

```ts
const shouldMaterializePendingTerminal =
  publicTab?.type === 'terminal' &&
  publicTab.status !== 'ready' &&
  (!targetsHost || !this.notifier?.focusTerminal || this.shouldMaterializeHeadlessMobileSessionTab(snapshot!, tab))
if (shouldMaterializePendingTerminal) {
  await this.createHeadlessMobileSessionTerminal(worktreeId, targetsHost, undefined, { identity: { tabId, leafId }, ... })
}
```

The client uses that on **first attach** but not on **reconnect** — `remote-runtime-pty-transport.ts`:

* `waitForHostSessionHandle()` (first attach) → `session.tabs.activate` → materialize → poll `session.tabs.list`. ✅
* `waitForResubscribeHostSessionHandle()` (reconnect, and the `Reconnect` button via `retryRecovery()` → `scheduleResubscribeAfterTransportClose()`) → `session.tabs.list` only, for `HOST_SESSION_ATTACH_TIMEOUT_MS` (15s), then `finishWithUnknownLiveness()`. ❌

`RemoteRuntimePtyRecoveryState` then hits `REMOTE_RUNTIME_AUTO_RECOVERY_TIMEOUT_MS` (60s) and flips to `disconnected`. `retryRecovery()` calls `recovery.begin()` again, which re-arms the same 60s deadline over the same list-only polling — which is why the button looks like it does nothing.

Two secondary consequences of the same blind spot:

1. `worktree.activate` from a client always resolves to `navigation: 'caller'` (`resolveRuntimeNavigationTarget` returns `'caller'` whenever `clientKind` is set), so `targetsHost` is false and `notifyHostActivateWorktree()` is skipped. The `!targetsHost` branch only hydrates the tab inventory; it never asks the host renderer to mount anything, and the `resumeSleepingAgents` wake next to it is gated on `opts.clientKind === 'mobile'` — desktop (`runtime` scope) clients get no host-side wake at all.
2. `ensureWebRuntimeWorktreeTerminalAfterWake()` cannot self-heal either, because `rememberHostTerminalTabCount()` counts *all* terminal tabs (`snapshot.tabs.filter(t => t.type === 'terminal').length`), including `pending-handle` ones. The host reporting "3 terminals" (all unattachable) makes the client skip respawn.

`requestRendererTerminalTabMount()` — the one API that can make the host mount a tab in the background — is only reachable from `terminal.subscribe` / the multiplex subscribe frame, both of which need a resolvable handle first. In this state there is no handle, so it is unreachable (chicken-and-egg).

## Fix

PR: activate (materialize) the host surface on the first iteration of the reconnect poll, exactly like the first attach does. A missing surface must still be proven by an inventory snapshot, so `tab_not_found` / `terminal_not_found` from the activate does not retire the pane; transient activate failures retry the materialize inside the same 15s budget.

Works for both host modes (windowed host and headless `orca serve`), since `createHeadlessMobileSessionTerminal()` spawns a runtime-owned PTY through `createTerminal({ focus: false, tabId, leafId })` regardless of whether a renderer window is attached.

planderdev:
## Summary

Fixes #11541.

A remote-runtime terminal pane whose **host-side PTY is gone** could never reconnect. `Reconnect` spun for ~60s and fell back to `Remote runtime disconnected`; the only way out was to walk to the host machine, open that project, and let its renderer respawn the terminals.

Cause: the host publishes such a surface as `status: 'pending-handle'`, `terminal: null`, and **only `session.tabs.activate` materializes a PTY for a pending surface** (`OrcaRuntimeService.activateMobileSessionTab` → `createHeadlessMobileSessionTerminal({ identity: { tabId, leafId } })`). The first-attach path (`waitForHostSessionHandle`) does exactly that, but the reconnect path (`waitForResubscribeHostSessionHandle`, also used by the `Reconnect` button via `retryRecovery()`) only polled `session.tabs.list` for `HOST_SESSION_ATTACH_TIMEOUT_MS`. It was waiting for a handle that nothing on the host would ever mint, so the pane was stranded until a human intervened on the host.

The reconnect now opens with the same activate, then falls back to inventory polling:

- The materialize targets the exact `tabId`/`leafId` identity, so the pane rebinds to the revived PTY instead of a fresh unrelated tab.
- It is a no-op when the surface is already `ready` (the host guards on `publicTab.status !== 'ready'`), so healthy reconnects behave exactly as before apart from one activate instead of one list.
- A missing surface must still be proven by an inventory snapshot: `tab_not_found` / `terminal_not_found` from the activate does **not** retire the pane, and any other activate failure retries the materialize on the next poll inside the existing 15s budget.
- `HOST_SESSION_ATTACH_TIMEOUT_MS` and the recovery deadline/backoff are untouched.

Reproduction that this fixes (macOS host + macOS client, both 1.4.160): work in a host project from a paired desktop client, restart Orca on the host (auto-update / quit / reboot), then reconnect from the client. Before: `Reconnect` never recovers. After: the pane reattaches on the first retry without touching the host.

## Screenshots

No visual change. The reconnect banner, its copy, and the button are untouched; only what the retry asks the host for changed.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck` (all three projects: node, tc.cli, tc.web)
- [x] `pnpm test` — ran `pnpm vitest run --config config/vitest.config.ts src/renderer` (2012 files, 18259 tests, all green) and the affected file (86 tests). The top-level `pnpm test` script could not start in this environment: its `ensure-native-runtime.mjs` preflight rebuilds `node-pty` for the local Node and fails on Node 22 (`package.json` wants Node 24), so vitest never runs. The failure is the preflight, not a test.
- [x] `pnpm build`
- [x] Added a regression test: `materializes a host surface whose PTY died instead of polling a dead inventory` — the host publishes the surface as `pending-handle`/`terminal: null`, the stale-handle path fires, and the test asserts the reconnect issues `session.tabs.activate` for `tabId`/`leafId` with `notifyClients: false, navigation: 'caller'`, then rebinds and resubscribes to the materialized handle without retiring the pane. Reverting the fix leaves it polling and the assertion fails.
- Updated one assertion in `reattaches from a later host snapshot after bounded replacement polling stops`: the first `session.tabs.list` of a reconnect now happens one backoff after the materialize, so its timeout is `> 14_000` rather than exactly `15_000`. The intent of that check (inventory timeouts are clipped to the remaining budget and shrink toward the deadline) is preserved.

## AI Review Report

Reviewed with the GJC coding agent, focused on the reconnect state machine and on not regressing pane retirement.

- **Behavior risk (retirement):** `null` from `waitForResubscribeHostSessionHandle` calls `retireRemoteTerminalId()`, which fires `onPtyExit`. The review flagged the naive version, where a `tab_not_found` from the activate would retire a pane during a host restart race. Changed so only an inventory snapshot that omits the surface returns `null`; a failed activate records unknown liveness and keeps the pane, matching the pre-existing "unknown liveness must not retire the pane" contract.
- **Loop/budget risk:** the materialize runs at most once per successful attempt and shares `HOST_SESSION_ATTACH_TIMEOUT_MS` (it is passed the remaining budget as its RPC timeout), so no new unbounded work, no extra timers, and no change to the recovery backoff or the 60s deadline. `requireReplacement` still rejects a returned handle identical to the stale one.
- **Cross-platform (macOS/Linux/Windows):** pure renderer RPC sequencing. No paths, no shell, no keyboard shortcuts, no platform APIs, no Electron-specific behavior. Nothing to guard per platform.
- **Remote/SSH/local:** only reachable for web-surface mirror tabs (`isWebTerminalSurfaceTabId`), i.e. panes mirroring another Orca's session. The non-mirror branch (`resolvePersistedHostPane()`, used by SSH and local worktrees) is untouched. On the host, the materialize path works both with a renderer window attached and headless `orca serve`, because `createHeadlessMobileSessionTerminal` spawns a runtime-owned PTY via `createTerminal({ focus: false, tabId, leafId })` and does not require a window. Folder workspaces use the same worktree selector plumbing and are unaffected.
- **Agents/integrations/git providers:** agent-neutral and provider-neutral. The host's materialize re-resolves the tab's `launchAgent` startup itself (existing behavior for pending agent tabs), so a revived agent pane is not downgraded to a plain shell by this change.
- **Performance:** one `session.tabs.activate` replaces the first `session.tabs.list` per recovery attempt. No added polling, no new allocations in hot paths. Steady-state reconnects (surface already `ready`) resolve on that first response exactly as they did on the first list.
- **UI quality:** no UI change; the banner simply stops being a dead end.

## Security Audit

- No new IPC channels, no new RPC methods, no protocol/capability changes. It reuses `session.tabs.activate`, which is already in the paired-client allowlist and is what the first attach calls.
- Parameters are the pane's own already-validated identity (`toRuntimeWorktreeSelector(worktreeId)`, `toHostSessionTabId(tabId)`, `leafId`). Nothing from the host response is used as a path, command, or shell input; the returned handle flows through the same `findReadyHostSessionHandle` validation as before.
- `notifyClients: false, navigation: 'caller'` keeps the effect scoped to the calling device — no host navigation and no fan-out to other paired clients.
- No secrets, credentials, tokens, environment, or command execution touched. No new dependencies. Error messages are matched, not constructed from user input.
- Follow-up (out of scope, noted in #11541): `rememberHostTerminalTabCount` counts `pending-handle` surfaces, which suppresses `ensureWebRuntimeWorktreeTerminalAfterWake` respawn, and a client's `worktree.activate` never asks the host to background-mount anything (`resumeSleepingAgents` next to it is gated on `clientKind === 'mobile'`). Both are separate defence-in-depth fixes; this PR intentionally changes one behavior.

## Notes

- Related but distinct: #11189 (host cold-parking unmounts a pane whose PTY is still alive) and #11305 (recovery backoff overruns its own 60s deadline). This PR is the case where the host has no PTY at all, so a handle can never appear; the other two remain valid on their own.
- Host-side testing notes: reproduced with a windowed host (`authoritativeWindowId != null`, `graphStatus: 'ready'`) where the affected workspace was not open, which is the state that leaves restored tabs at `pendingActivationSpawn` with no PTY.
- X handle: n/a.


Jinwoo-H (changes requested):
Thanks for adding the exact-leaf activation path and the paired-runtime coverage. I validated current head `a8d08ce08a` in an isolated worktree, but it is not safe to merge yet.

Blocking findings:

1. `waitForResubscribeHostSessionHandle` does not re-activate when the first activation races host publication and returns the previous ready handle, then authoritative inventory later exposes the exact `pending-handle` surface. The activation result selects list polling, and the list result only re-arms activation when `retryActivationAfterInventory` was set by a classified missing activation error. A deterministic test against the current head observed `activate, list, list, list` instead of `activate, list, activate`; the pane never obtained the replacement handle.
2. A non-missing activation failure such as `method_not_found` never falls back to authoritative inventory. A deterministic old-host test observed `activate, activate, activate` instead of `activate, list`, so a ready replacement already visible through inventory is ignored. This also means transient/non-classified activation errors do not satisfy the required inventory safety boundary.
3. The changed-spec E2E gate does not execute the new tests. Both tests are tagged `@headful`, while changed-spec CI always runs the `electron-headless` project, whose `grepInvert` excludes them. CI therefore fails with `Error: No tests found.` The reliability-gate command separately names `electron-headful`, so the durable command and required changed-spec job disagree.

Validation evidence:

- Required transport + recovery-state suites at the PR head: **121/121 passed**.
- Full typecheck: **passed**.
- Copied STA-3002 oracle: the unmodified test stops at its pre-fix-only assertion that reconnect must first perform more than one list poll (`0` observed). Removing only that obsolete intermediate assertion leaves the actual acceptance oracle unchanged; activation, exact handle rebind, and restored input/output then pass **1/1**.
- Added deterministic adversarial tests: current head fails **2/2** as described above. A local candidate that makes activation immediately inventory-confirmed and re-arms activation only when exact inventory is pending passes the full affected suites **123/123**, typecheck, formatting, and STA-3002 **1/1**; I did not push it because the live paired journey still failed.
- Real paired-runtime validation after locally removing the selector mismatch: Playwright selected both headed-host and headless-folder-host topologies, but the headed-host journey failed to materialize a replacement (**1 failed, 1 did not run**). Adding an authoritative `pending-handle` publication barrier did not make it pass, so the E2E fault/recovery handoff also needs correction before this can be considered proven.
- Current GitHub checks: **42 passed, 1 failed** (`e2e / changed e2e specs`), with the failure being the selector issue above.

Please update the state machine so activation never proves absence, authoritative inventory alone can retire, non-missing activation failures inventory-fallback, and an exact pending inventory re-arms activation inside the existing bounded window. Then repair the changed-spec selector and make both paired-runtime topologies genuinely pass. I will re-review promptly.

Jinwoo-H:
Pushed `0361bc1a0b` addressing the two blocking defects from the last review, plus a host-side gap that the E2E surfaced.

## 1. Stale-ready activation race

The first recovery activation can race host publication and answer with the pre-restart handle. The exact inventory that follows then sees only a pending handle — but the loop kept list-polling for the rest of its window and never activated again, so the surface was never materialized.

Observed on the previous head: `activate / list / list / list …`

## 2. Non-missing activation failure had no inventory fallback

Activation was only retried through inventory after a *classified missing* error. Any other failure — `method_not_found` on an older host, a transport error — kept retrying activation forever with no inventory fallback.

Observed on the previous head: `activate / activate / activate …`

Both are fixed by alternating the two requests inside the existing bounded backoff window. An activation response is never absence proof, so inventory always follows it; inventory that still publishes the surface without a ready handle re-activates. Retirement is now reachable **only** from an inventory response, which tightens the previous behavior where a successful activation response missing the surface could retire the pane.

Two regression tests assert the exact call **sequences**, not counts:
- `re-activates when a stale ready activation response precedes the pending surface` → `['session.tabs.activate', 'session.tabs.list', 'session.tabs.activate']`
- `falls back to inventory when activation fails for a non-missing reason` → `['session.tabs.activate', 'session.tabs.list']`

Both also assert the pane is neither retired nor errored (`onPtyExit`/`onError` never called) and that it rebinds to the replacement handle.

## 3. Host-side: activation destroyed the pane it was asked to rebuild

The paired E2E could not pass, and it was not a harness problem. With **no fault injection at all**, a single `session.tabs.activate` against a parked surface failed with `terminal_pane_owner_changed` and permanently deleted the host tab — a subsequent manual activate returned `tab_not_found`.

Traced to `attachStablePaneOwner` in `src/main/ipc/pty.ts`. Once the PTY is proven dead it retires the owner — `runtime.onPtyExit(...)`, which removes the session tab — and then asks persistence to retire too. For a pane parked with `keepHistory`, persistence had *already* dropped the binding, so `retirePersistedStablePaneOwner` returned `false`, which the caller read as a competing owner and threw. The destructive half had already run and the fresh spawn never happened, so the pane was unrecoverable.

An already-absent persisted binding is a completed retirement, not a conflict. The genuine owner-change guard (a *different* persisted ptyId/incarnation) and the unknown-liveness guard (`terminal_pane_owner_unverified`) are unchanged.

Covered by `respawns a proven-dead owner whose persisted binding was already retired` — red with `terminal_pane_owner_changed` before the fix, green after.

## 4. CI: `changed e2e specs` reported "No tests found"

Both specs were tagged `@headful`, and the changed-spec lane runs `electron-headless`, whose project config is `grepInvert: /@headful/`. Neither test needs focus or pointer capture (per `tests/e2e/AGENTS.md`, `@headful` is only for those), so the tags are dropped and the reliability-gate command now targets `electron-headless` to match.

## Validation

- `remote-runtime-pty-transport.test.ts` + `pty.test.ts`: **555 passed**
- `remote-runtime-pty-transport.test.ts` + `remote-runtime-pty-recovery-state.test.ts`: **123 passed**
- `orca-runtime.test.ts` + `mobile-session-terminal-persistence-retirement.test.ts`: **1033 passed, 1 skipped**
- `pty.test.ts` + `pty-management.test.ts`: **460 passed**
- typecheck, lint, format: clean
- Headed paired-host E2E: **5/5 green**, ~8s each

Independent oracle: the STA-3002 reproduction from the outage/stuck-reconnect suite (activation → rebind → input flow) is **red on `origin/main`** — 51 list-only inventory polls, zero activations — and **green on this branch**. Its intermediate `listCallsAfterOutage > 1` checkpoint encodes the old list-polling cadence and is obsolete now that recovery activates immediately; that checkpoint alone was dropped, the activation/rebind/input assertions were kept.

## Known gap: headless folder-host topology

The second E2E is marked `test.fixme` with the reason inline. Its fault injection cannot be set up against a headless `orca serve` host: `terminal.stopExact` returns `terminal_exact_stop_failed` because `stopAndWait`'s keep-history verification window expires before the parked PTY is observed gone, so the pane never reaches `pending-handle` and the reconnect path is never reached. **This precondition fails identically on this PR's base**, so it is a pre-existing exact-stop defect, not a reconnect-activation one, and it deserves its own issue.

The behavior itself *was* confirmed in that topology with instrumented runs: the headless host materializes the pending surface on the same tab/leaf, and the client rebinds to the replacement PTY within one second. What blocked the assertion afterwards was that parking the host PTY clears the client's active-worktree selection, so reading the replacement through the *active* pane saw nothing; the spec now reselects the pane before that read (the rebind assertion itself is unchanged). That selection reset also looks worth a separate look.

Jinwoo-H (approved):
Approving on the strength of the reconnect state machine, the host materialization fix, and green CI (43 checks; `e2e / changed e2e specs` now genuinely selects and runs the spec — headed paired host ✓ 21.5s — instead of reporting "No tests found").

Not merging: the headless folder-host topology is still unproven. It is marked `test.fixme` because `terminal.stopExact` cannot park a PTY on an `orca serve` host (`terminal_exact_stop_failed` — stopAndWait's keep-history verification window expires before the parked PTY is observed gone), so the fault can't be set up and the reconnect path is never reached. That precondition fails identically on this PR's base, so it is a pre-existing exact-stop defect rather than something this change introduced, and it deserves its own issue. The recovery behavior itself was confirmed by hand in that topology.

Leaving the merge call to a maintainer who can decide whether shipping without headless paired coverage is acceptable given the defect is pre-existing.

PR #11550, opened 2026-07-30
fix(terminal): reconcile Windows IME deletion and Shift toggle

yoke233:
## Summary

Follow-up to #11293, based on native Microsoft Pinyin traces captured in Electron on Windows.

The native events differ from the synthetic model used by the first revision of this PR:

- Backspace during a one-letter preedit arrives as `keydown key=Process code=Backspace keyCode=229`, followed by an empty composition update/input, then `keyup Process/229` and `keyup Backspace/8`.
- Pressing Shift to toggle the IME arrives as `Process code=ShiftLeft keyCode=229` plus the ordinary Shift events. It does not emit Enter. The previous predicate checked `key=Process`, modifier state, and `keyCode=229`, but not `code`, so it could misclassify this Shift-owned Process event as Shift+Enter and submit a newline.

This revision:

- recognizes deletion from both the native `key=Process/code=Backspace|Delete` shape and the logical `key=Backspace|Delete` compatibility shape;
- lets only tracked composition deletion keydowns reach xterm, then reconciles the preedit without emitting PTY data;
- scopes empty composition updates to a one-shot deletion token that expires on native progress, fallback completion, finalization, cancellation, restart, or disposal;
- consumes the ordinary Backspace/Delete keyup paired with the IME transaction so Kitty keyboard event reporting cannot leak it to the PTY;
- requires `code=Enter|NumpadEnter` before treating `Process/229` as a modified Enter, so `code=ShiftLeft` cannot create a newline;
- keeps the transaction ownership, stale-event rejection, native-progress precedence, and bounded timer behavior introduced by #11293.

Upstream `main` was checked before the update and remains the direct base of this PR; no later upstream commit fixes these native Windows event shapes.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` (full repository lint not run)
- [x] `pnpm typecheck`
- [ ] `pnpm test` (full repository suite not run)
- [x] `pnpm build`

Additional verification:

- 124 focused IME, xterm, deletion-release, newline, and keyboard-policy tests passed across 5 files.
- Electron-headless Windows IME E2E: 2 passed; the separately tagged real-Codex test was skipped by design.
- The E2E asserts one-letter preedit deletion emits zero PTY input and Shift toggle commits only `s`, with no newline.
- The xterm adversarial test enables Kitty event reporting (`CSI > 10 u`) and proves the paired `Backspace/8` keyup emits nothing.
- Changed TypeScript files passed `oxlint`, React Doctor lint, and `oxfmt` through the commit hook.
- Regenerated ESM/CJS bundles and source maps retain `XTERM_VERSION = 6.1.0-beta.287`; the patch contains no `6.0.0` marker.

## AI Review Report

Independent reviews checked implementation quality, state lifetime, generated artifacts, test fidelity, and cross-platform regressions.

They flagged four issues in the earlier revision, all addressed here:

1. Generated xterm bundles accidentally reported `6.0.0`; all artifacts were regenerated with `src/common/Version.ts` from `6.1.0-beta.287`.
2. Switching entirely from `key` to `code` regressed events with an unavailable physical code; both shapes are now accepted and tested.
3. The deletion transaction id could remain stale after deferred fallback; it is now a data-bound one-shot token cleared on every completion/cancellation path.
4. The ordinary `Backspace/8` keyup could leak under Kitty event-type reporting; a narrow, expiring paired-release guard and Kitty-mode regression test were added.

Cross-platform review confirmed the new release guard only arms from `keyCode=229` during an active composition. macOS and Linux standalone-229 behavior, Linux/Sogou candidate guards, folder workspaces, SSH transports, shortcut labels, paths, and shell behavior are unchanged.

## Security Audit

No auth, secrets, IPC contracts, command execution, or production path handling were added. The CDP event dispatch and PTY log parsing live only in E2E helpers and operate on test-created files. The production change is limited to renderer keyboard-event classification and the existing pinned xterm patch; no dependency version changed.

## Notes

- Related to #11293.
- This is a renderer/xterm input-routing fix; the daemon does not need a corresponding change.
- The generated xterm patch includes minified ESM/CJS artifacts and source maps, so its behavior is also specified by direct installed-package adversarial tests.


OrcaWin:
Cross-compared this with the fixes now on main: #12265 supersedes the Shift/Enter portion, and #12273 fixes cancelled stale-preedit replay. The Windows Process/229 Backspace/Delete reconciliation and paired Kitty keyup ownership here remain distinct and are not yet landed, so I am leaving this PR open rather than incorrectly marking that work resolved. A current-main extraction should add an explicit Windows gate, adapt around #12265's capture guard, and cover delayed native progress before reuse.

OrcaWin:
The native Windows behaviors covered here are now resolved on main: modified Process events require a physical Enter, cancelled preedits no longer replay stale text, and #12276 preserves the Microsoft Pinyin Shift-commit trace with @yoke233 credited as co-author. #12278 also lands the complementary Hangul syllable-flush fix. Thank you @yoke233 for the native traces and regression coverage.

PR #11568, opened 2026-07-30
fix(agent-hooks): exit Windows hooks when Orca env is missing

### LINKED ISSUE #11549: [Bug]: Codex managed hook leaves persistent cmd.exe windows outside Orca on Windows
### Operating system

Windows

### Orca version

1.4.161

### Short summary

Orca installs its managed Codex hook in the user-wide Codex hook configuration. As a result, Codex sessions started by other applications also invoke Orca's `codex-hook.cmd`.

When such a session does not have the Orca-specific environment variables, the Windows hook enters its stdin-drain fallback and starts `more.com`. In the affected execution path, `more.com` does not receive EOF, so both `cmd.exe` and `more.com` remain running indefinitely. Every subsequent Codex hook event can leave another visible command window and another pair of leaked processes.

### What happened?

The persistent window was running:

```text
C:\Windows\system32\cmd.exe /c ""%USERPROFILE%\.orca\agent-hooks\codex-hook.cmd"
```

Process inspection showed the following persistent process tree for each affected invocation:

```text
cmd.exe
└── more.com
```

The generated hook contains the following fallback:

```bat
if "%ORCA_AGENT_HOOK_PORT%"=="" goto :orca_agent_hook_drain_stdin
if "%ORCA_AGENT_HOOK_TOKEN%"=="" goto :orca_agent_hook_drain_stdin
if "%ORCA_PANE_KEY%"=="" goto :orca_agent_hook_drain_stdin

...

:orca_agent_hook_drain_stdin
"%SystemRoot%\System32\more.com" >nul 2>nul
exit /b 0
```

Codex instances launched inside an Orca pane have the required environment variables and post their hook payload to Orca normally. Codex instances launched outside Orca do not necessarily have:

```text
ORCA_AGENT_HOOK_PORT
ORCA_AGENT_HOOK_TOKEN
ORCA_PANE_KEY
```

This is a normal situation because the managed hook is registered in the user-wide Codex configuration, not only in Orca-managed sessions.

### How can we reproduce it?

1. Install and run Orca on Windows with managed Codex hooks enabled.
2. Confirm that `%USERPROFILE%\.codex\hooks.json` contains entries invoking `%USERPROFILE%\.orca\agent-hooks\codex-hook.cmd`.
3. Start or invoke Codex from another GUI application, outside an Orca-managed terminal.
4. Trigger a Codex lifecycle event such as `SessionStart`, `UserPromptSubmit`, or a tool call.
5. Observe a visible `cmd.exe` window running `codex-hook.cmd`.
6. Inspect the process tree and observe that `cmd.exe` is waiting for a `more.com` child.
7. Repeat the operation and observe additional persistent windows and processes.

Although the hook configuration specifies a timeout, the observed Windows child processes remain after the caller stops waiting.

### Expected behavior

If the hook is invoked without a valid Orca context, it should exit successfully and immediately without launching a stdin-drain process:

```bat
if "%ORCA_AGENT_HOOK_PORT%"=="" exit /b 0
if "%ORCA_AGENT_HOOK_TOKEN%"=="" exit /b 0
if "%ORCA_PANE_KEY%"=="" exit /b 0
```

The successful path should remain unchanged when all required Orca environment variables are available.

### Verified local workaround

I replaced the three missing-environment branches with `exit /b 0` and removed the `more.com` fallback from the generated `codex-hook.cmd`.

Validation results:

- No Orca environment variables: exited successfully in approximately 102 ms.
- Only some required variables present: exited successfully in approximately 87 ms.
- Complete environment with an unreachable test endpoint: preserved the existing `curl.exe` path and exited successfully after its configured network timeout.
- No residual `cmd.exe`, `curl.exe`, or `more.com` hook processes remained after the tests.

This confirms that draining stdin is not required for the missing-Orca-context path and that immediate successful termination prevents the process leak.

### Suggested fix

Update the Windows Codex hook generator so that:

1. `ORCA_AGENT_HOOK_ENDPOINT` is still sourced first when available.
2. Missing `ORCA_AGENT_HOOK_PORT`, `ORCA_AGENT_HOOK_TOKEN`, or `ORCA_PANE_KEY` causes an immediate `exit /b 0`.
3. The `more.com` stdin-drain fallback is not generated for the Codex Windows hook.
4. Existing managed `codex-hook.cmd` files are refreshed during Orca's normal managed-hook reconciliation or upgrade process.
5. Other agents and the normal Orca payload-posting path remain unchanged.

### Additional context

A brief `cmd.exe` flash may still occur because Codex currently invokes Windows command hooks through `cmd.exe /C`. That is a separate launcher behavior. This report is specifically about command windows that remain open indefinitely and the associated `cmd.exe`/`more.com` process leak.

This is related in theme to [#7116](https://github.com/stablyai/orca/issues/7116), where globally registered Windows hooks also affected agent sessions outside Orca, but the failure mode here is different: this issue concerns the Codex hook's missing-environment stdin-drain path and persistent processes.

innocarpe:
## Description
Stop Windows managed agent hooks (Codex/Claude/etc.) from hanging in `more.com` when invoked outside an Orca pane without `ORCA_*` env.

User-wide Codex hooks still run for non-Orca sessions. Missing PORT/TOKEN/PANE_KEY now `exit /b 0` immediately instead of routing to the stdin-drain epilogue that can leave visible `cmd.exe` + `more.com` processes forever.

## Focused fix
- In scope: `buildWindowsHookEnvironmentGuardLines()` missing-env path
- Out of scope: removing drain for other failure modes, changing happy-path curl post

## Preserves
- Full Orca env still posts hook payloads via curl as today
- Drain epilogue remains for non-env failure paths that still need stdin ownership

## Evidence
- Test: `pnpm exec vitest run --config config/vitest.config.ts src/main/agent-hooks/managed-hook-stdin-lifecycle.test.ts src/main/codex/hook-service.test.ts src/main/claude/hook-service.test.ts src/main/agent-hooks/installer-utils.test.ts` (92 passed, 10 skipped)
- Guard lines assert `exit /b 0` and no `goto` drain for missing env

## User-regression-tradeoffs
- Outside-Orca hooks still flash briefly via `cmd.exe /C` (Codex launcher); they no longer stick open

Fixes #11549


brennanb2025:
Thanks for this — good catch on the root cause, and the fix is the right shape.

Heads up that I pushed a few commits directly to this branch. Here's what changed and why:

**Folded in one more instance of the same bug**

- `claude-hook.cmd` checked `DEVIN_PROJECT_DIR` *before* the Orca env guards, and that check jumps to the `more.com` drain. Devin imports `.claude` hooks by default, so a Devin session outside an Orca pane still parked in `more.com` forever — a known-reachable instance of exactly the bug this PR fixes. I moved the env guards above the Devin skip. Both branches end the hook, so nothing changes inside a pane; only the out-of-pane Devin case now exits 0 instead of hanging.

**Made the regression test actually hold the line**

- The guard test pinned three exact string literals. A fourth `ORCA_*` guard copied from an existing pattern would have reintroduced #11549 with the suite still green — it now asserts the rule (no `ORCA_*` guard in a generated `.cmd` may route to the drain label) rather than the three guards that happen to exist today.
- That rule also now catches the `if not defined ORCA_X goto :drain` spelling, which is the common idiom at several other sites and was slipping straight through.
- The Windows launch matrix asserted that *every* `.cmd` drains its writer, which your fix deliberately makes false. Fixed that, plus its name and a comment that both claimed more than the test checked.

**Docs**

- Recorded in `hook-stdin-contract.ts` why Windows batch now diverges from the POSIX stdin policy, citing #11549, so it doesn't get "fixed" back later.

**Verified on a real Windows host:** before the change, the missing-env path left a live `more.com` child and never exited; after, it exits 0 promptly with no child. The happy path (ORCA env present) is unchanged — the POST still lands.

Two things I saw but left out of this PR so they don't get lost:

- `managed-agent-hook-controls.ts` skips install on `cli_not_found` without removing an already-installed script, and hook scripts are never deleted — so a user whose CLI moved keeps the leaking script indefinitely. Worth a follow-up.
- `copilot-hook.ps1` and `kimi-hook.sh` have the same hang class, but they need a bounded read rather than this fix.

One thing for the release notes whenever this lands: it stops *new* stuck processes but can't reap ones that leaked before the upgrade, so anyone with leftover `cmd.exe`/`more.com` will need to kill them or reboot once.

brennanb2025:
Correction on my earlier push: it left this branch conflicting with `main` in `src/main/claude/hook-service.ts`. The branch was well behind `main`, and `main` had since reworded the same block I moved, so the change landed here textually collided once it was published.

Fixed by merging latest `main` in and resolving that block: `main`'s current wording and structure are kept, and the guard reorder is re-applied on top — env guards above the Devin skip. Diffing the Windows branch against `main` shows the same set of emitted lines in a different order, so the behavior delta is still just the reorder.

`agent-hooks` + `claude` hook suites, typecheck, and lint are green, and the PR is mergeable again. Sorry for the churn.

brennanb2025:
Follow-up from a standard-fix audit pass over this PR — **no change to your fix, which was right.**

Swapping the three `ORCA_*` guards from `goto :orca_agent_hook_drain_stdin` to `exit /b 0` is exactly the standard shape here: batch streams the payload straight to curl instead of capturing it, so the guard has nothing to lose by giving up stdin, and `more.com` draining to EOF is precisely what strands the process when the caller abandons the pipe. Three lines, no new machinery. Your diagnosis of #11549 and the mechanism you chose are what ships.

The audit was aimed at my own review commits on top of yours, and it found two things worth cutting:

- The gate I added asserting "no `ORCA_*` guard routes to the drain" was hand-rolled as `script.match(/^.*…*$/gm)` + `.toBeNull()`. That is just `expect(script).not.toMatch(/…/)`, which this repo already uses in 94 places. Verified equivalent: with a synthetic fourth guard added in the other house spelling (`if not defined ORCA_AGENT_HOOK_ENV goto :…`) and all three real guards left intact, the rewritten gate still fails.
- A module-level `BROKEN_WRITER_CODES = new Set([...])` asserted via `.has(x)).toBe(true)` — used once, 280 lines from its definition. Inlined to `expect(['EPIPE', 'ECONNRESET']).toContain(error.code)`, which also gives a real diff on failure instead of "expected false to be true."

I also cut the comment I had grown on `buildWindowsHookEnvironmentGuardLines` from 13 lines back to 4. It had turned into a taxonomy adjudicating three other files' exemptions, and one of those claims (that the statusline caller provably closes stdin) was stated more firmly than the evidence supports — for a file this PR does not touch. Net −20 lines, all mine.

Behavior is unchanged and pinned: reverting the guards to `goto`, reverting the guard/Devin-skip ordering, and adding a fourth drain-routed guard each turn the suite red. The Windows-only live matrix still skips off-win32, so that one is pinned by the generated-script assertions rather than by execution here.

innocarpe:
@brennanb2025 Thank you for taking this so carefully — and for co-authoring on the branch.

Your diagnosis of the Devin / `DEVIN_PROJECT_DIR` path was a real find: that would have left the same `more.com` hang outside an Orca pane even after the env-guard fix. The stronger regression rule (no `ORCA_*` guard routes to the drain, including the `if not defined` spelling), the merge conflict cleanup, and the standard-fix audit that trimmed the review-side nits all look right to me.

I’m good with landing your tip as-is. The two follow-ups you left out (stale hook scripts on `cli_not_found`, and the bounded-read hang class in `copilot-hook.ps1` / `kimi-hook.sh`) make sense as separate focused PRs rather than packing them into this one.

Also noted for release notes: this stops *new* stuck processes but can’t reap ones that already leaked before upgrade.

brennanb2025:
Thank you for the PR!!

Issue #10932, opened 2026-07-27
[Bug]: CLI agents crash after `nvm install <newer>` — Orca picks newest nvm node for PATH but spawns CLIs from older nvm node (ABI mismatch)


chushoumao:
### Operating system

macOS

### Orca version

latest stable (v1.4.x)

## Summary

Running `nvm install <newer>` (e.g. `nvm install 22` while `codex` / `claude` are installed under `v20`) breaks Orca's ability to launch any coding agent CLI on macOS / Linux. Orca keeps launching successfully, but every agent (Codex, Claude Code, etc.) errors out at first use. The only known workaround is to fully remove the newer Node version (or replace it with a newer-but-still-compatible v20).

## Steps to reproduce

Tested on macOS 15, packaged Orca (DMG install, not dev mode):

1. Install nvm if not already present: `curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash`
2. Install Node 20: `nvm install 20`
3. Install the coding agent under that version, e.g. `npm install -g @openai/codex` (or `@anthropic-ai/claude-code`)
4. Verify Orca launches the agent successfully and runs `/status`.
5. `nvm install 22` (any version newer than 20)
6. **Do not reinstall the agent under 22.**
7. Launch the packaged Orca app and trigger any agent (rate-limit poll, worktree boot, login, etc.).
8. Observe agent fails to start, even though `codex` (or `claude`) is still on disk at `~/.nvm/versions/node/v20.x.x/bin/codex`.

## Expected

Orca continues to launch the agent successfully. Picking the newest installed Node version for `PATH` is reasonable; but the agent binary Orca chooses to spawn should run under a compatible Node.

## Actual

Agent crashes / errors out immediately. Two failure modes observed:

- `codex` errors with `Cannot find module` / native-module `NODE_MODULE_VERSION` mismatches, because Orca spawned `~/.nvm/versions/node/v20.x.x/bin/codex` but its shebang picked up `~/.nvm/versions/node/v22.x.x/bin/node` from `PATH`.
- `claude` similarly crashes on native module load or first `require`.
- Orca's own UI does not surface a useful message — only the agent logs show the failure.

## Root cause

`src/main/codex-cli/command.ts` and `src/main/startup/configure-process.ts` disagree about which nvm version should win:

[...]

PR #11570, opened 2026-07-30
fix(startup): prune stale mise/nvm install paths on shell PATH merge

### LINKED ISSUE #11529: [Bug]: Stale mise Node paths accumulate in process PATH and leak into daemon terminals across updates
### Operating system

macOS

### Orca version

1.4.161

### Summary

Orca can keep a stale, version-specific mise Node directory in its main-process `PATH`, preserve it across an auto-update, pass it to the long-lived PTY daemon, and then leak it into every newly created terminal.

In the observed case:

- the current mise configuration resolved Node to `24.18.0`;
- a clean external terminal (Ghostty) correctly used Node `24.18.0`;
- new Orca terminals and agents used Node `26.5.0`;
- the daemon's `PATH` contained both the stale exact-version directory and the current alias:

```text
~/.local/share/mise/installs/node/26.5.0/bin
~/.local/share/mise/installs/node/lts/bin
```

The daemon had no `MISE_*` session variables. It was not actively selecting Node 26; it had inherited a stale path string.

### Steps to reproduce

A deterministic regression test should be possible without a real updater:

1. Start the packaged app while mise selects an exact Node version, for example `26.5.0`.
2. Allow Orca's packaged-process PATH hydration to capture that login-shell PATH.
3. Change mise to another version, for example:
   ```bash
   mise use -g node@24
   ```
4. Make a fresh login shell return only the current Node version as its active runtime.
5. Run Orca's PATH hydration/merge again, or relaunch through the auto-update path while preserving the previous process environment.
6. Start the out-of-process daemon and create a new terminal.
7. Inspect:
   ```bash
   command -v node
   node --version
   echo "$PATH"
   ```

### Expected behavior

A newly started Orca process/daemon should use the current login-shell environment.

Obsolete version-specific directories such as `~/.local/share/mise/installs/node/26.5.0/bin` should not survive after the shell and mise configuration select Node 24.

New Orca terminals should resolve the same Node version as a clean external login shell.

### Actual behavior

The stale Node 26 directory remained ahead of the current mise directory, so `node` resolved to `v26.5.0` in Orca while a clean external shell resolved to `v24.18.0`.

Restarting the Orca UI did not necessarily clear the problem because the daemon and updater lifecycle can preserve the polluted environment.

### Investigation

The installed 1.4.161 bundle shows this chain:

1. Packaged startup calls `patchPackagedProcessPath()`.
2. `hydrateShellPath()` runs the user's shell with `-ilc` and captures its complete `PATH`.
3. `mergePathSegments()` mutates `process.env.PATH` using this shape:
   ```js
   const next = [
     ...shellSegments,
     ...currentSegments.filter((segment) => !shellSegmentSet.has(segment)),
   ].join(path.delimiter);
   ```
   This merges and deduplicates entries but does not remove obsolete version-manager paths already present in `currentSegments`.
4. `createOutOfProcessLauncher()` starts `daemon-entry.js` with `fork(..., { cwd, detached, stdio })` and no explicit `env`, so the daemon inherits the mutated main-process environment.
5. The daemon creates each PTY from `process.env + opts.env`, propagating the stale path to every terminal.

The observed timeline also contained an auto-update transition:

- the previous main process had originally started while Node 26 was active;
- mise was later changed to `lts` / Node 24;
- the old process executed `quitAndInstall`;
- the new main process and daemon started immediately afterward;
- the daemon still contained the old exact Node 26 directory.

A clean-shell probe with an obsolete Node 26 directory in its input environment confirmed that mise activates Node 24 but preserves the inherited Node 26 directory as an ordinary trailing PATH entry. A union-style PATH merge therefore cannot clean this state.

### Suggested fix

Possible approaches:

1. Treat the hydrated login-shell PATH as authoritative on packaged startup, instead of unioning it with stale version-specific entries.
2. Before merging, prune obsolete version-manager runtime directories that are absent from the freshly captured shell PATH, including patterns such as:
   - `~/.local/share/mise/installs/<tool>/<version>/bin`
   - `~/.nvm/versions/node/<version>/bin`
3. Pass an explicit, sanitized environment when forking the daemon rather than implicitly inheriting the main process's mutable PATH.
4. Add a regression test where:
   - the existing process PATH contains Node 26;
   - the fresh login shell selects Node 24;
   - the final main-process, daemon, and PTY PATHs do not resolve Node 26.

### Related

- #10932 covers another version-manager PATH inconsistency (newest nvm Node vs. the Node version owning an agent CLI). This report is different: it concerns obsolete PATH entries accumulating across shell hydration / updater / daemon inheritance.

innocarpe:
## Description
Stop obsolete version-manager install bins (e.g. stale `mise/installs/node/26.5.0/bin`) from surviving `mergePathSegments` after the login shell has moved to a new pin.

When hydrating packaged PATH from the user's shell, exact-version mise/nvm/asdf install segments and fnm multishell paths that the shell no longer exports are dropped from the process PATH tail instead of being union-merged forever.

## Focused fix
- In scope: prune obsolete version-manager install segments in `mergePathSegments`
- Out of scope: daemon fork env rewrite, broader PATH redesign, agent CLI resolution (#10932)

## Preserves
- Shell-exported version-manager paths still win via shell segment ordering
- Non-version-manager custom process PATH entries still append when missing from shell
- Existing prepend/promote PATH tests unchanged in intent

## Evidence
- Test: `pnpm exec vitest run --config config/vitest.config.ts src/main/startup/hydrate-shell-path.test.ts` (15 passed)
- Stale mise 26.5.0 dropped when shell only has current pin; custom non-VM path retained

## User-regression-tradeoffs
- If a user relied on a *stale* exact-version install that their shell no longer puts on PATH, Orca will match the shell (desired) rather than keep the orphaned entry

Fixes #11529

### DIFF
--- a/src/main/startup/hydrate-shell-path.test.ts
+++ b/src/main/startup/hydrate-shell-path.test.ts
@@ -5,6 +5,7 @@ import type { ChildProcessWithoutNullStreams } from 'node:child_process'
 import {
   _resetHydrateShellPathCache,
   hydrateShellPath,
+  isVersionManagerInstallPath,
   mergePathSegments,
   type HydrationResult
 } from './hydrate-shell-path'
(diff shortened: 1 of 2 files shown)


innocarpe:
### Sync update (`692381474`)

Address CR/Greptile: tighten fnm session regex, rename helper, add nvm/asdf/fnm tests

- Commit: `6923814740e03aeb746d005f86e39678181332e8`
- Upstream: https://github.com/stablyai/orca/pull/11570
- Fork portfolio: https://github.com/innocarpe/orca/pull/129

PR #11571, opened 2026-07-30
feat(source-control): add stash actions to the commit dropdown

Douglasgomes027:
## Summary

Adds `git stash` actions to the Source Control panel. Orca had **zero** stash support at any layer before this — `grep -ri stash src/main/git src/preload src/renderer/src/runtime` returned only error strings and agent prompts telling agents *not* to run it. The only way to stash was to type it into a terminal.

Eight actions live in a nested **Stash** group at the bottom of the commit split-button dropdown:

| | |
|---|---|
| Stash | tracked changes only; prompts for an optional name |
| Stash (Include Untracked) | adds `-u`; prompts for an optional name |
| Pop / Apply Latest Stash | newest entry |
| Pop / Apply Stash… | picker dialog |
| Drop Stash… | picker + destructive confirm |
| Drop All Stashes… | destructive confirm |

The group follows the module's existing invariant — **every row always renders**, disabled with a reason in its tooltip, so the menu shape stays stable and testable. An untracked-only tree disables *Stash* with "Only untracked files — use Stash (Include Untracked)" rather than letting git report a confusing "no local changes to save".

Works on all four hosts: local, WSL, SSH relay, and remote-runtime RPC.

### Three decisions worth reviewing

**One implementation, not four.** The command core lives in `src/shared/git-stash-commands.ts` with the git runner injected, so main and the relay execute identical argv and identical result interpretation. This deliberately avoids the pattern in `commitChangesRelay`, which carries a "keep the two paths in sync" comment because it is a hand-maintained copy.

**No capability gate — and that is verified, not assumed.** Every flag used predates the Git 2.25 baseline (`stash push` is 2.13; `-m`, `--include-untracked`, `list --format`/`-z`, `apply|pop|drop|clear [--] <ref>` are older). `git stash push --staged` is 2.35, which is exactly why a staged-only variant was left out of the op set. I built Git 2.25.5 from source and ran the compatibility suite against it — a new case in `git-binary-compatibility.test.ts` exercises the whole stash argv, so the claim is enforced by the existing real-binary CI matrix rather than by a version comment.

**Pop conflicts mirror git rather than compensating.** On conflict git keeps the entry and says so. Auto-dropping would destroy the user's only copy of a half-applied changeset; auto-reverting would discard resolution work. The result carries `conflicted: true`, the entry survives, and the toast names the surviving `stash@{N}`.

> ⚠️ Known consequence, flagged deliberately: a stash-pop conflict does not write `MERGE_HEAD`, so `detectConflictOperation` returns `'unknown'` while unresolved conflicts exist. `shouldRenderCommitArea` then unmounts `CommitArea` — taking the chevron, and the Stash menu, with it until conflicts are resolved. That guard exists to prevent a commit surface over an unresolved conflict, so I did not weaken it; the toast names the ref instead so nothing looks lost. Happy to revisit if you'd prefer a different trade.

### Naming a stash

Both stash rows prompt for an optional name before running git, so entries are identifiable in the picker instead of a wall of `WIP on main: …`. Confirming an empty field is a valid choice — the message is omitted and git writes its own subject — while cancelling aborts the stash entirely. Those two outcomes are deliberately distinct; collapsing them would make Escape silently stash.

The `message` parameter was already built, validated, and tested through every transport in the first round of this PR, but nothing set it, so it was unreachable. This closes that.

### Index-shift guard

`stash@{N}` is positional and Orca runs agents against the same worktree, so an agent stashing between the picker rendering and the user clicking would silently retarget a destructive action at someone else's work. Picked entries carry their commit oid; a mismatch fails with `stash_entry_moved` **before** the destructive command runs. Covered by a real-repo test that reproduces the shift.

### Stash count is read on demand

Deliberately **not** added to `useGitStatusPolling`. That hook runs for every worktree on a 60s interval with backoff added specifically because of poll churn (and has a regression spec). Upstream status rides along in porcelain v2 for free; a stash count is a separate `git log` on `refs/stash` and only matters while the menu is open. It refreshes on menu open, on panel visibility, and after a mutation.

### before
01-before-commit-dropdown

### after

02-after-commit-dropdown
------

03-stash-submenu

----

https://github.com/user-attachments/assets/b36d115b-da9e-423d-8507-008e8ac03114

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

**9954 tests pass** across every area this PR touches (renderer sidebar, main/git, relay, shared, runtime RPC, providers, preload, IPC). 7 failures in 2 files (`local-pty-shell-ready`, `pty-shell-launch`) — both PTY/terminal, neither touching git. Verified **identical failures on `origin/main`** (2 files, 7 failed / 74 passed), so they are pre-existing and unrelated.

New coverage:

- `git-stash-list-output.test.ts` — NUL parsing, colon-bearing subjects, `-z`-ignored fallback, CRLF, truncated trailing record.
- `stash.test.ts` — exact argv per op, ref rejection, conflict classification, `stashed:false`, error ladder, index-shift guard.
- `stash-real-repo.test.ts` — real temp repos: full round trip, `-u` semantics, **a real induced pop conflict proving the entry survives**, index shift, detached HEAD, dash-prefixed message.
- `git-handler-stash.test.ts` — the relay/SSH path against a real repo, including method registration and re-validation.
- `git-stash.test.ts` — RPC schema rejection of bad refs/oids before they reach the runtime.
- `source-control-stash-menu-items.test.ts` — every row's enabled state and disabled reason, including "busy beats loading beats empty" and the untracked-only case below.
- `useSourceControlStash.test.ts` — the worktree-switch race, facade identity, and the naming prompt (typed name reaches git, empty omits the field, cancel does not stash).
- `source-control-stash-actions.spec.ts` (e2e) — the menu, an untracked-only tree, cancelling the name prompt, and a real stash → pop round trip asserted on the file row, not store state.

E2E passes locally (4/4, confirmed stable over three consecutive full-file runs). Note `e2e` is not a required check and only runs on PRs touching `tests/e2e/**` — which this one does.

## AI Review Report

Self-review with Claude, focused on the risks this change actually carries.

**Cross-platform (macOS / Linux / Windows).** The stash argv contains no filesystem paths — there are no pathspecs, so `literalPathspec` is not needed and no separator assumptions exist. `wslDistro` is threaded through `gitOptionsForWorktree` on every call, so native and WSL stay distinct. The list parser tolerates CRLF. No new keyboard shortcuts, accelerators, or shortcut labels, so no `metaKey`/`CmdOrCtrl` surface. Menu rows are plain items with no platform-conditional copy.

**What the review flagged, and what changed as a result:**

1. **A wrong code comment.** I had written that `-m` needed guarding against a `-`-prefixed message. Checking against real git showed `-m` consumes the next argv token regardless, and args are passed as an array — the guard was never load-bearing. Rather than leave a misleading rationale, I corrected the comment to state the real reason (bound empty/huge subjects) and added a real-repo test that stores `--all` as a message verbatim, so the assumption is enforced rather than asserted.
2. **`rev-parse --verify --quiet` exits 128, not 1, on a missing stash.** My first implementation expected empty stdout and would have surfaced a raw git error instead of `stash_entry_moved`. Found by probing real git; fixed and covered by a test.
3. **A second, easily-missed consumer of `DropdownEntry`.** `CreateHostedReviewComposer` renders the same array as `CommitArea`. Widening the union to add the submenu would have broken it at typecheck. Extracted a shared `source-control-dropdown-entries.tsx` renderer so both surfaces migrate together — both files got *shorter*.
4. **An unsound type predicate in the existing test.** `(e): e is DropdownItem => e.kind !== 'separator'` silently mistyped the submenu as a row (TS accepts a narrowing predicate to a subtype). Replaced with a real guard.
5. **A store map would have added leak surface.** The plan called for `gitStashCountByWorktree`; worktree-keyed maps need purge wiring and there is a dedicated `worktree-removal-maps-leak` spec guarding that class of bug. Since the count is consumed by exactly one component, I scoped it to component state instead — no purge path to forget.

**Also checked, no change needed:** `max-lines` (never disabled or bumped; `providers/types.ts` was at 291/300, which is precisely why `GitStashProvider` is a separate file, and the dropdown test was at 785/800, hence a separate stash test file); folder workspaces (Source Control early-returns for non-git repos, and the hook no-ops on `isGitWorkspace: false`); GitLab/GitHub neutrality (stash is provider-agnostic, nothing added to hosted-review paths); detached HEAD (covered by test — stash works and reads `(no branch)`).

## Security Audit

**Argument injection — the main risk, defended at three independent layers.** A stash ref must match `/^stash@\{\d+\}$/`: at the RPC zod schema, again in the shared command core, and the relay re-validates because its entrypoint is reachable independently of the schema (params arrive as `Record<string, unknown>`). This blocks `-`-prefixed flags, `--all`, `-p`, and arbitrary revs from reaching `drop`/`pop`. Every invocation also passes `--` before the ref. Rejection is covered by tests at each layer.

**Command execution.** No shell strings are constructed anywhere; all args go through `execFile` as arrays. Stash messages are bounded to 500 chars and are safe as `-m` values (verified above).

**Path handling.** `worktreePath` is never trusted from the renderer on the local path — `resolveRegisteredWorktreePath` resolves it against the store, matching `git:stage`. No user-supplied paths enter the stash argv at all.

**IPC / transport surface.** Six new `git:stash*` channels and six RPC methods, each following the established SSH-provider-fork → resolve → route shape. **`src/relay/git-exec-validator.ts` is deliberately untouched** — its allow-list correctly rejects `stash`, and `git-exec-validator.test.ts` asserts that; the whole point of dedicated relay methods is to avoid loosening the read-only `git.exec` passthrough.

**Destructive-action safety.** Drop and Drop All are gated behind `confirmVariant: 'destructive'` confirmations, and the oid guard means a concurrent stash cannot silently redirect them. An older relay returns `-32601`, which is translated into an actionable reconnect hint rather than a raw JSON-RPC error.

**No** new dependencies, secrets, auth paths, or network calls.

## Review round

Findings from the automated review, all verified against the code before acting:

- **Plain Stash was enabled on an untracked-only tree** (found via a comment on the e2e seed helper). `hasUnstagedChanges` is true for untracked files too, so passing it through as "has tracked changes" left the row clickable with nothing to save — `git stash push` exits 0 with "No local changes to save". The submenu now takes an explicit tracked-entry count. My unit test had missed it by exercising the resolver directly, and the e2e had encoded the wrong expectation because its helper created a *new* file despite being named `seedTrackedChange`. Fixed, with a red-then-green test and a dedicated e2e case.
- **Stale stash count across a worktree switch** — a slow read for worktree A could overwrite the reset and re-enable restore rows for worktree B. Fixed with a live worktree check at response time; note the obvious fix (capturing the id before the `await`) does not work here, because the closure *is* the worktree it was created for.
- **`expectedCommitOid` was ignored when the ref was implicit**, so "pop the latest, but only if it is still the one I saw" ran unguarded. Now verified against `stash@{0}`.
- **Unstable hook facade** made a consumer's `useCallback` a no-op. Memoized, and the call sites now depend on the individual stable functions.
- **Message type guard** at the IPC boundary, matching the relay's.
- **e2e hygiene** — targeted per-test cleanup instead of discarding every entry (the seeded repo is shared), and the file runs serially.

Also restored three tracked `test-results/*.md` files that a local Playwright run had wiped; they were never part of this change.

## Rebase note

Kept current with `main` across several rebases. What that surfaced, all handled:

- **`providers/types.ts` became a barrel upstream**, moving `IGitProvider` into `git-provider-contract.ts`. A textual merge would have been wrong even where it applied cleanly, so the `GitStashProvider &` intersection was **relocated**. This also dissolved the constraint that motivated extracting `GitStashProvider` at all (`types.ts` was 291/300 lines; now 26) — the separate file is kept as the cleaner split.
- **Upstream deleted a block of hosted-review tests**, moving them to a new `source-control-dropdown-items.create-pr-intent.test.ts`. I took the deletion, then found the new file used the same unsound `!== 'separator'` narrowing this PR's `DropdownSubmenu` union breaks — the identical issue already fixed in the original file, applied there too. The conflict was in one file; the actual breakage landed in a file that had no conflict at all.
- **`pr-comment-time` moved** from `renderer/lib` to `shared`. The stash picker reuses it for relative timestamps, so its import now matches the sibling `checks-panel-content.tsx`.
- **Locale catalogs** conflicted on adjacent key additions. Rather than splice generated JSON by hand, I took upstream's files and regenerated from source via `sync:localization-catalog`. Only `en.json` is modified now; `es`/`ja`/`ko`/`zh` fall back to English instead of carrying copies that can drift — which retires the earlier "ellipsis missing in four locales" finding more durably than the original manual edit.

One e2e test was also made deterministic: seeding wrote a file and pushed a single status snapshot, but a poll started before the write could resolve afterwards and overwrite it, so the row never appeared. The wait now re-refreshes until the entry shows, verified over repeated full-file runs.

## Notes

- No version bump (maintainer-only via the Cut Release action).
- 51 new localization keys generated via `pnpm sync:localization-catalog`; all five locales are in parity and both localization gates pass. The non-English catalogs carry the English strings until translated, matching how the tooling seeds new keys.
- Follow-up if wanted: a staged-only stash via `git stash push --staged`. It is the one stash flag above the 2.25 baseline, so it needs a real `GitCapability` slug plus fallback — deliberately out of scope here to keep this change capability-free.


Douglasgomes027:
Thanks — all three findings were accurate and are fixed in 162b3433c.

**P1 — stale count survives worktree switch.** Confirmed and fixed. Worth noting the suggested fix needed one adjustment: capturing `ctx.worktreeId` before the `await` and comparing it against the closure's `worktreeId` would always match, because `refreshStashCount` is recreated whenever `runtimeContext` changes (which depends on `worktreeId`) — so the closure *is* the worktree it was created for. The comparison has to be against a live value, so I read the current worktree from a ref at response time instead. Same intent as `SourceControlStashPicker`'s stale flag, different mechanism because there's no effect cleanup to hang it on.

**P2 — unstable facade identity.** Confirmed. Memoized the hook's return and narrowed the consumers to the individual stable functions (`refreshStashCount`, `runStashAction`) rather than depending on the whole object. Note `CommitArea` isn't `React.memo`'d, so there was no measurable render cost — but the `useCallback` was silently a no-op, which is misleading either way.

**P2 — type aliases between import blocks.** Fixed, moved below all imports.

Added `useSourceControlStash.test.ts` (7 cases) covering the race with a deferred promise, facade identity across re-renders, the worktree-switch reset, picker dispatch, and the folder-workspace no-op — the first one fails without the fix.

One thing the identity test surfaced: the facade is only stable if the caller passes a stable `settings` reference. `SourceControl` does (`activeRepoSettings` is memoized), and the test now mirrors that rather than passing a fresh object per render.

Re-verified after the fixes: `pnpm lint`, `pnpm typecheck`, 1720 sidebar/git/IPC tests, and both e2e specs pass.

Douglasgomes027:
Second round — all four addressed in 60264b69a, and one of them was a real product bug.

**The e2e comment on `seedTrackedChange` was the important one.** Chasing it found that plain **Stash was enabled on an untracked-only tree**. `hasUnstagedChanges` is true for untracked files as well as tracked modifications, and I was passing it straight through as "has tracked changes" — so the row was clickable with nothing for it to save, and `git stash push` would exit 0 with "No local changes to save". That is precisely the confusing no-op the disabled-reason copy claims to prevent.

Two of my own tests should have caught it and didn't: the unit test exercised `resolveStashSubmenu` directly, bypassing the wiring, and the e2e's helper created a *new* file — untracked despite its name — so it asserted the buggy state as correct. Fixed by threading an explicit tracked-entry count, with a red-then-green unit test and a dedicated e2e case for the untracked-only tree.

**`expectedCommitOid` skipped when `ref` is null** — agreed, and the reasoning about the documented provider contract is right. The guard now runs against `stash@{0}` whenever an oid is supplied, so "pop the latest, but only if it is still the one I saw" is honored. Two tests added; our own callers never hit this path, but the API allowed it and the contract promised it.

**`message` type guard** — fixed, now `typeof args.message === 'string'`, matching the relay. Worth noting it was already caught downstream (`assertValidStashMessage` throws `invalid_stash_message`), so this tightens where the error surfaces rather than closing a hole.

**e2e cleanup** — done, with one adjustment. A blanket "discard every entry" `afterEach` made things *worse*: the seeded repo is shared and the specs run in parallel, so cleanup from one test deleted a sibling's file mid-assertion (it failed on the first run). Cleanup is now scoped to the paths each test created, and the file is `mode: 'serial'`.

Separately, I noticed a local Playwright run had wiped three tracked `test-results/*.md` files into the diff — restored, so the PR is back to 43 files.

Re-verified: `pnpm lint`, `pnpm typecheck`, 2325 tests across the touched areas, and 3/3 e2e.

Douglasgomes027:
Third round reviewed. Three fixed, two I'm pushing back on — all now squashed into the single commit b6fcd2bf7.

**Fixed**

- **Wheel listener bound to a detached node** (`SourceControlStashPicker`) — correct and a real bug. Radix mounts a fresh `DialogContent` per open, so keying the effect on `entries.length` alone meant reopening with the same count left the listener on the previous node and wheel scrolling silently died. `mode` added to the deps.
- **`git init -b` is Git 2.28+** — verified against the 2.25.5 binary I built for this PR: `error: unknown switch 'b'`. Fixed, though not the way suggested: dropping `-b main` outright would leave the branch as `master` on old Git and break the `On main:` subject assertions, so it's now `git init` + `git symbolic-ref HEAD refs/heads/main`, which works on every supported version and keeps the name deterministic. Worth noting this pattern is pre-existing repo-wide (`worktree-shared-directories`, `porcelain-v1-records`, `status-shared-symlinks` all use `git init -q -b main`); I only changed my own file rather than widening this PR.
- **Capitalize "Git"** — agreed, and the catalog backs it: 81 user-facing strings capitalize it vs 34 lowercase, most of those URLs or technical tokens. Updated the string and the en.json entry.

**Not changing**

- **"Require an OID when `ref` is omitted."** I don't think this is right. "Pop Latest Stash" is a *live* selection — the user is asking for whatever is newest now, not for a specific entry they picked from a list. Requiring an oid would change the semantics to "pop the entry that was newest when the menu opened", which fails with `stash_entry_moved` on a perfectly reasonable action, and would make the app's own valid request schema-invalid (the UI passes `ref: null` with no oid by design). The oid guard exists precisely for the *picker* paths, where the user saw and chose a specific entry — those already carry it and are covered by tests. This matches the editor behavior too: its pop-latest runs a bare `git stash pop`.
- **Stash mutation timeout on the web path.** The premise doesn't hold — the finding hedges on "if the default timeout is shorter", and it isn't: `web-runtime-client.ts` defines `REQUEST_TIMEOUT_MS = 30_000`, exactly the `STASH_MUTATION_TIMEOUT_MS` used on the desktop/SSH path. The web client already gets the same 30s budget, so there's no divergence to fix. (`syncFork`'s explicit `60_000` is an override *upward* for a genuinely slower op, not evidence of a lower default.)

Re-verified after the fixes: `pnpm lint`, `pnpm typecheck`, and 4/4 e2e.

Douglasgomes027:
Fourth round — two fixed in 1cb904c94, and one of them was a good catch on the relay boundary.

**Param readers laundering malformed values** — correct, and the sharper part of the finding is that my own docstring promised the opposite of what the code did. A present-but-non-string `ref` was coerced to `null`, which `restoreStash` reads as "target the newest entry" — so a malformed request quietly popped a *different* stash instead of being refused. Same shape for `expectedCommitOid`: a malformed oid became `undefined`, silently disabling the concurrent-stash guard precisely when a caller asked for it. Both now reject (`invalid_stash_ref` / `invalid_stash_oid`) while a genuinely omitted ref still means "latest". Three tests added, including one asserting both stashes survive a malformed pop.

**Locale ellipsis** (the `menu-items` and `ko.json` comments are the same issue) — also correct. `sync:localization-catalog` treats existing non-English values as translations and won't overwrite them, so `es`/`ja`/`ko`/`zh` kept the pre-ellipsis `Drop All Stashes` after I updated the English source. All four updated.

The `git init -b` thread from this batch was already fixed in the previous push, using `git init` + `symbolic-ref` rather than the suggested `checkout -b`, so the `main` assertions stay deterministic on old Git.

Verified: `pnpm lint`, `pnpm typecheck`, `verify:localization-catalog`, and the relay stash suite.

I'm resolving the threads that are actually addressed. Two stay open deliberately, since they're judgement calls rather than defects and I'd rather a maintainer weigh in than have my disagreement buried:

- **"Require an OID when `ref` is omitted"** — Pop/Apply Latest is a live selection, not a snapshot; requiring an oid would make the app's own valid request schema-invalid and fail a reasonable action with `stash_entry_moved`.
- **"Stash mutations may need a longer timeout on the web path"** — the premise doesn't hold: `web-runtime-client.ts` already defaults to `REQUEST_TIMEOUT_MS = 30_000`, identical to the desktop/SSH stash budget.

Happy to implement either if you disagree with my reading.

brennanb2025:
Taking a look!

Douglasgomes027:
Thanks @brennanb2025

Douglasgomes027:
Can you please check @brennanb2025 @AmethystLiang

PR #11576, opened 2026-07-30
fix(file-explorer): sort numbered file names naturally across every listing surface

### LINKED ISSUE #11426: [Bug]: File Explorer sorts names lexicographically — numbered files list 100, 200 before 1, 2, 3
## What I see

The File Explorer (right-hand sidebar) sorts file and folder names by plain string comparison, so numeric prefixes sort character-by-character instead of by value. In a directory of hymn files named by song number, the listing reads:

```
409 - Gloryland ...
41  - O for a Thousand ...
410 - There Is a ...
```

So `41` lands between `409` and `410`, and `100`/`200` appear before `1`, `2`, `3`. Any directory whose files are numbered — song numbers, chapter numbers, migration numbers, `step-1`…`step-12`, `page-1.png`…`page-20.png` — is effectively unbrowsable, because finding the next item means scanning the whole list rather than reading down it.

## Expected

Natural (numeric-aware) ordering, the same as Finder, Windows Explorer, VS Code, and Zed:

```
1 - ...
2 - ...
3 - ...
41 - O for a Thousand ...
409 - Gloryland ...
410 - There Is a ...
```

**Request:** compare file and folder names in the File Explorer with a numeric-aware collator (e.g. `new Intl.Collator(undefined, { numeric: true, sensitivity: 'base' })`) instead of a raw string comparison. Ideally the same comparator is used everywhere names are listed (Quick Open results, Source Control file lists) so orderings agree.

**Benefit:** numbered directories read in the order the numbers mean, and the File Explorer matches what every other file browser on the platform does — no surprise, no scanning.

## Environment

- Orca 1.4.159, macOS (Darwin 25.5.0)

brennanb2025:
## Summary

Fixes #11426 (Linear STA-2950). The File Explorer sorted names lexicographically, so numbered files listed `100, 200` before `99`. Every directory-listing funnel now sorts directories-first in natural (numeric-aware) name order via one shared comparator: local `fs:readDir`, the SSH short-circuits (re-sorted locally so an older remote relay can't leak lexicographic order), the relay's own handler, the remote-runtime RPC / paired-web routes (re-sorted at the renderer choke point), the SSH folder browser and runtime server dir pickers, the name-filtered view, Source Control directory nodes, and the mobile file explorer.

**ELI5:** computers sort text character-by-character, so "100" comes before "99" (because "1" < "9") — like shelving "Chapter 100" before "Chapter 99". This teaches Orca's file lists to read whole numbers inside names the way Finder and Explorer do, so `9`, `99`, `100` shelve in the order you'd expect.

**Live before/after (real dev app, right-sidebar File Explorer):**

Before — lexicographic (`1, 100, 2, 200, 409, 41, 410, 9, 99`):

After — natural (`1, 2, 9, 41, 99, 100, 200, 409, 410`), re-verified on the final build:

## Screenshots

See before/after above.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — targeted instead: affected desktop suites (182 files / 1699 passed, incl. new ordering tests at the relay, SSH `fs:readDir` handler branch, renderer choke point, source-control tree, shared helper) plus `mobile/src/files/file-tree.test.ts` (8 passed, mobile vitest config). A reviewer's broader sweep of src/main+relay+shared hit only two known local-environment artifacts (GIT_CONFIG env leak; load-dependent timeout), both green when re-run isolated.
- [ ] `pnpm build` — left to CI.
- [x] Added or updated high-quality tests that would catch regressions: ordering pinned at every funnel that previously sorted independently, including a handler-level SSH-branch test and a choke-point test that feeds a deliberately out-of-order backend payload.

## AI Review Report

Adversarial review loop: **4 rounds with a fresh reviewer each round; round 4 came back clean (zero blocker/major).**
- Round 1 (1 blocker, 1 major, 3 minor): both readDir funnels short-circuited to the SSH filesystem provider before the patched sort, and the relay still compared lexicographically — fixed by re-sorting locally after the provider returns plus fixing the relay; added the code-unit tie-break (numeric collation ties `2` vs `02`, which would otherwise fall back to readdir order); pinned the collator locale to `'en'` so every host produces one order; patched the SSH folder browser and runtime dir pickers that would have disagreed with the Explorer.
- Round 2 (1 major, 2 minor): the mobile file explorer re-sorted host results with bare `localeCompare`, undoing the host's order — fixed with the shared comparator + mobile test.
- Round 3 (2 major, 2 minor): remote-runtime RPC and paired-web routes returned the host's order verbatim — fixed at the renderer choke point every desktop route converges on; added the missing call-site tests; routed Source Control path compares through the tie-breaking comparator.
- Round 4: clean. Its minor notes (dead export, locale-pinned test oracle) are applied.

Cross-platform: explicitly reviewed for macOS/Linux/Windows — the comparator is pure JS with no path-separator or shell assumptions; verified `/` terminates ICU numeric runs so digits never merge across path segments; SSH, WSL (routes through main's patched local readdir), remote-runtime, paired-web, relay version-skew (older remote sends lexicographic order → local re-sort corrects it), and mobile (Hermes ships Intl on RN 0.83 for both platforms; Metro already watches `src/shared`) were each traced. Performance review (perf skill): pass-with-notes — numeric collation costs ~5x bare `localeCompare` (~5ms extra on a 10,000-entry directory; invisible at realistic sizes), the hoisted collator avoids a 68x naive form, re-sorting an already-sorted SSH response is near-free (V8 adaptive sort, n−1 comparisons), and one collator instance was deduped.

## Security Audit

No input handling, command execution, auth, secrets, or dependency changes. The comparator is pure string comparison; no user input reaches `Intl.Collator` options. IPC surface unchanged (same handlers, same payload shapes — only result ordering). The relay change affects ordering only, not path resolution or authorization; `resolveAuthorizedPath` checks are untouched and still precede every local readdir.

## Notes

- Trade-offs: (1) Source Control changed-files ordering becomes locale-independent (`'en'`-pinned) — for locales whose collation differs from English (sv/cs/da) the list can reorder relative to the old OS-locale behavior; accepted for cross-host consistency and repo precedent (skill-freshness pins `'en'`). (2) Zero-padded numeric ties (`2` vs `02`) now order deterministically by code units instead of readdir order.
- Hermes on-device numeric collation is unverified (tests run under Node/V8); Hermes ships Intl on RN 0.83 and mobile already uses `numeric: true` collation in three modules, so risk is bounded — a device spot-check of the Files panel with `9/99/100`-style names would close it.
- Follow-up candidates (not this PR): the quick-open file list is still code-unit sorted, so its 20k-result truncation boundary is chosen in a different order than the Explorer displays; `CombinedDiffViewer`'s flat view still sorts by bare `localeCompare` while its tree view now uses the shared collator; three mobile modules use the slow per-call `localeCompare(x, undefined, {numeric:true})` form that the new shared module makes a one-line fix each.

### DIFF
--- a/mobile/src/files/file-tree.test.ts
+++ b/mobile/src/files/file-tree.test.ts
@@ -34,6 +34,18 @@ describe('file-tree', () => {
     ])
   })
 
+  it('orders numbered names naturally, matching the desktop explorer', () => {
+    const cache: DirectoryCache = {
+      '': { entries: [entry('100 - b.txt'), entry('9 - c.txt'), entry('99 - a.txt')] }
+    }
+
+    expect(flattenDirectoryCache(cache, new Set()).map((row) => row.id)).toEqual([
+      'file:9 - c.txt',
+      'file:99 - a.txt',
+      'file:100 - b.txt'
+    ])
+  })
+
   it('mirrors desktop default browse exclusions while keeping dotfiles visible', () => {
     const cache: DirectoryCache = {
       '': {
--- a/mobile/src/files/file-tree.ts
+++ b/mobile/src/files/file-tree.ts
@@ -1,5 +1,6 @@
 // Pure tree projection for the mobile file explorer. Mobile mirrors desktop
 // browse semantics by flattening cached files.readDir results as folders open.
+import { compareFileNames } from '../../../src/shared/file-name-sort'
 
 export type MobileDirEntry = {
   name: string
@@ -118,7 +119,7 @@ function compareDirectoryEntries(a: MobileDirEntry, b: MobileDirEntry): number {
   if (a.isDirectory !== b.isDirectory) {
     return a.isDirectory ? -1 : 1
   }
-  return a.name.localeCompare(b.name)
+  return compareFileNames(a.name, b.name)
 }
 
 export function shouldIncludeMobileFileExplorerEntry(entry: MobileDirEntry): boolean {
--- a/src/main/ipc/filesystem.test.ts
+++ b/src/main/ipc/filesystem.test.ts
@@ -358,6 +358,30 @@ describe('registerFilesystemHandlers', () => {
     lstatMock.mockRejectedValue(Object.assign(new Error('missing'), { code: 'ENOENT' }))
   })
 
+  it('re-sorts SSH provider listings directories-first in natural order', async () => {
+    // Why: the remote relay may be an older build that still sorts lexicographically.
+    getSshFilesystemProviderMock.mockReturnValueOnce({
+      readDir: vi.fn().mockResolvedValue([
+        { name: '100 - b.txt', isDirectory: false, isSymlink: false },
+        { name: '9 - c.txt', isDirectory: false, isSymlink: false },
+        { name: '10 - dir', isDirectory: true, isSymlink: false },
+        { name: '99 - a.txt', isDirectory: false, isSymlink: false }
+      ])
+    })
+    registerFilesystemHandlers(store as never)
+
+    const result = (await handlers.get('fs:readDir')!(null, {
+      dirPath: '/remote/repo',
+      connectionId: 'ssh-1'
+    })) as { name: string }[]
+    expect(result.map((e) => e.name)).toEqual([
+      '10 - dir',
+      '9 - c.txt',
+      '99 - a.txt',
+      '100 - b.txt'
+    ])
+  })
+
   it('returns an actionable reconnect error when the SSH filesystem provider is unavailable', async () => {
     registerFilesystemHandlers(store as never)
 
--- a/src/main/ipc/filesystem.ts
+++ b/src/main/ipc/filesystem.ts
@@ -30,6 +30,7 @@ import type {
 } from '../../shared/types'
 import type { GitHistoryOptions, GitHistoryResult } from '../../shared/git-history'
 import type { SshMutationExpectation } from '../../shared/ssh-types'
+import { compareFileNames } from '../../shared/file-name-sort'
 import { assertSshMutationExpectation } from '../ssh/ssh-connection-generation'
 import {
   buildRgArgs,
@@ -531,7 +532,7 @@ export function registerFilesystemHandlers(
           if (a.isDirectory !== b.isDirectory) {
             return a.isDirectory ? -1 : 1
           }
-          return a.name.localeCompare(b.name)
+          return compareFileNames(a.name, b.name)
         })
       } catch (error: unknown) {
         recordCrashBreadcrumb(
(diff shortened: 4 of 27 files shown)


brennanb2025:
Mandatory merge-blocker recovery is complete at final HEAD `197a9ad84d8e8f83ba7f2121a4a8edf593d2aed5` (fast-forward pushed as `197a9ad84d`). The first fresh same-model/xhigh review found material Quick Open and remote-runtime picker gaps; those plus the flat Source Control omission were fixed, the next fresh review returned CLEAN, and a final fresh `gpt-5.6-sol`/`xhigh` reviewer explicitly returned CLEAN for this exact committed HEAD across local, SSH/relay/version-skew, remote-runtime, folder workspaces, Source Control, Quick Open, and mobile.

Validation:
- Tests/build: 196 focused desktop tests passed (1 skipped), 8 mobile file-tree tests passed, node/web/mobile typechecks passed, relay artifacts built, changed-code quality and max-lines ratchet passed, and `git diff --check` passed.
- Perf: bounded Quick Open heap measured about 31 ms at the real 20,001-path cap and 153–156 ms at synthetic 100k, versus about 557 ms for the prior full sort.
- Electron: isolated final-branch instance identity verified; visible local folder-workspace File Explorer + Quick Open and real SSH File Explorer + Quick Open + flat Source Control + remote folder picker all rendered `9, 99, 100` naturally. Renderer console: 0 errors.
- SSH: Ubuntu 22.04 ARM64 throwaway Docker target, real `demo-project` cloned from a local bundle after GitHub auth blocked the preferred clone, relay built/deployed/handshook, and remote backing checks showed only the intentional numbered-fixture commit with no extra test branches/worktrees/folders.
- Mobile: iPhone 17e loaded the exact branch Metro bundle without redbox, paired through the normal LAN flow to the isolated desktop, desktop backing state confirmed one connected device, and the real mobile Files surface visibly rendered folders `9, 99, 100` followed by files `9, 99, 100`.
- CI: final head has 42 successful checks and one expected path-gated E2E skip; the sole first-attempt failure was the out-of-diff `terminal-fit-restore.test.ts` 1 ms timing assertion (`14999` vs `15000`), which passed locally 9/9 and passed on the one bounded failed-job rerun. PR mergeability is CLEAN.

Evidence (uploaded without committing):
- SSH File Explorer: SSH File Explorer
- SSH Quick Open: SSH Quick Open
- SSH Source Control: SSH Source Control
- SSH remote picker: SSH remote picker
- Local folder File Explorer: Local folder File Explorer
- Local folder Quick Open: Local folder Quick Open
- iOS Files: iOS Files

Cleanup completed for the isolated Electron profile/process, mobile credential/device/helper/Metro, SSH target/container/relay/temp keys, and local fixtures. No approval, merge, auto-merge, issue close, force-push, or check-bypass action was taken; the PR remains open and auto-merge remains unset.

PR #11582, opened 2026-07-30
fix(orchestration): stop fencing fresh-run callers and accept revoked coordinator after takeover

### LINKED ISSUE #11572: [Bug]: Devin CLI agents running inside Orca cannot reliably send or receive cross-terminal orchestration messages
### Operating system

macOS

### Orca version

1.4.162

## Summary

Devin CLI agents running inside Orca cannot reliably send or receive cross-terminal orchestration messages. A worker Devin terminal that completes an Orca-dispatched task cannot deliver `worker_done` to its coordinator Devin terminal, and some Devin terminals fall into `legacy_read_only` mode so they cannot apply any orchestration effects at all.

Plain terminal inspection (`terminal list`, `terminal show`, `terminal read`) works, but Devin processes seem to lose their Orca process identity after a retain/restart, so `orca orchestration send` and `task-list` are rejected.

## Environment

- **Orca app version:** 1.4.162
- **Devin CLI version:** 3000.3.22
- **Platform:** macOS
- **Orca runtime:** reachable, `ready`
- **Agent type in affected terminals:** Devin CLI

## Capability matrix

| Operation | Works? | Notes |
|-----------|--------|-------|
| `devin list` | yes | Lists local Devin sessions in the current directory |
| `devin --version` | yes | Reports Devin CLI version |
| `orca status --json` | yes | Runtime and app version returned |
| `orca worktree ps --json` | yes | Worktrees and live agent metadata returned |
| `orca terminal list --json` | yes | Terminals in the worktree returned |
| `orca terminal show --terminal <handle> --json` | yes | Terminal metadata returned |
| `orca terminal read --terminal <handle> --json` | yes | Returns tail output; requires cursor pagination to read full history |
| `orca orchestration inbox --json` | yes | Returns stored messages for the current terminal (read-only) |
| `orca orchestration dispatch-show --task <task_id> --json` | yes | Returns dispatch metadata |
| `devin` reading other terminals automatically | no | No native Devin CLI subcommand for cross-terminal output; the agent must invoke `orca terminal read` itself |
| `orca orchestration task-list --json` from a Devin terminal | **no** | Fails with `legacy_read_only`: *"This retained legacy coordinator could not prove its original process identity. No effects were applied."* |
| `orca orchestration send` from a Devin worker to a Devin coordinator | **no** | Fails with *"The coordinator terminal ... is not retained for this legacy dispatch, so the worker_done message could not be delivered through orca orchestration send."* |
| `worker_done` delivery to coordinator | **no** | Dispatch remains `dispatched` with `completed_at: null` after the worker reports completion |
| Sending auditor/orch warnings from legacy read-only terminals | **no** | Terminal reports `legacy_read_only` and effects are blocked |

## Steps to reproduce

1. Open multiple Devin terminals in the same Orca worktree, e.g. one coordinator and one worker.
2. Create an Orca orchestration task and dispatch it to the worker Devin terminal.
3. From the worker Devin terminal, complete the work and try to report completion:
   ```bash
   orca orchestration send \
     --to <coordinator_handle> \
     --from <worker_handle> \
     --type worker_done \
     --subject "Task complete" \
     --body "..." \
     --task-id <task_id> \
     --dispatch-id <dispatch_id> \
     --outcome succeeded \
     --json
   ```
4. Observe in the worker terminal: the message is not delivered and a warning is printed about the coordinator not being retained for the legacy dispatch.
5. In the same or another Devin terminal, run:
   ```bash
   orca orchestration task-list --json
   ```
6. Observe the `legacy_read_only` error.
7. Verify the dispatch is still open:
   ```bash
   orca orchestration dispatch-show --task <task_id> --json
   ```
   The dispatch shows `status: dispatched` and `completed_at: null`.

## Expected behavior

- A Devin agent in an Orca terminal should be able to send `worker_done`, `heartbeat`, `status`, and `escalation` messages to other Devin terminals.
- `task-list` and other orchestration operations should succeed from the active Devin terminal, because the Devin CLI is the legitimate owner of the terminal.
- A dispatched task should transition to `completed` when the worker sends `worker_done`.

## Actual behavior

- `worker_done` is dropped with the warning:  
  `The coordinator terminal ... is not retained for this legacy dispatch, so the worker_done message could not be delivered through orca orchestration send.`
- The coordinator terminal keeps polling `orca orchestration check --unread` / `inbox` and receives no new messages.
- Some Devin terminals are stuck in `legacy_read_only` and cannot run `task-list` or send orchestration messages.
- The dispatch stays `dispatched` with `completed_at: null` after the worker finishes.

### Worker terminal output

```text
Note: The coordinator terminal ... is not retained for this legacy dispatch, so the worker_done message could not be delivered through orca orchestration send. The work itself is fully complete and verified.
```

### `legacy_read_only` error

```json
{
  "ok": false,
  "error": {
    "code": "legacy_read_only",
    "message": "This retained legacy coordinator could not prove its original process identity. No effects were applied.",
    "data": { "effectsApplied": false }
  }
}
```

### Dispatch still open

```json
{
  "status": "dispatched",
  "completed_at": null,
  "delivery_contract": "legacy_direct"
}
```

### Auditor/automation terminal blocked

A Devin terminal running an auditor/automation role reported that it could not send a warning to the coordinator because its terminal was `legacy_read_only`.

## Impact

- Multi-agent Devin workflows in Orca cannot complete their lifecycle handshake.
- Coordinators do not know when worker tasks finish, so they wait/poll indefinitely.
- Workers complete code changes but the orchestration dispatch stays open, blocking the next pipeline steps.
- `legacy_read_only` terminals cannot be used for supervisor/auditor roles that need to send `status` or `escalation` messages.

## Possible causes / hypotheses

1. **Process identity loss on retain/restart:** When a Devin terminal is retained across an Orca runtime update or pane switch, the Devin process loses its original `process_incarnation`/`launch_token` association, so `legacy_read_only` is triggered.
2. **Legacy dispatch contract not updated:** The dispatch is on `delivery_contract: legacy_direct` and the coordinator terminal is not "retained" for that contract, so `worker_done` cannot be routed.
3. **Devin CLI has no native Orca terminal/orchestration identity:** The Devin binary is a generic TUI agent; all cross-terminal awareness comes from ad-hoc `orca` CLI calls. If those calls run in a process whose identity was not correctly re-bound by Orca, the runtime rejects effects.

leofmarciano:
Closes #11572

## Summary

Devin CLI agents running inside Orca could not reliably send or receive cross-terminal orchestration messages. Two over-eager legacy compatibility code paths caused this:

1. **`legacy_read_only` on fresh runs** — `LegacyCoordinatorAuthority.resolve` forced the adopted legacy run for every `orchestration.*` preflight and threw `legacy_read_only` at any caller that could not prove legacy-coordinator identity, including fresh-run coordinators with no connection to the legacy system. Now only callers previously known to the legacy run are fenced; unknown fresh-run callers fall through to the normal current-run handler.

2. **`worker_done` "not retained" after takeover** — `isLegacyCoordinatorHandle` returned only the committed principal's handle, so after a takeover (`run-use --takeover-legacy`) that revoked the principal, `worker_done`/`escalation` to the new coordinator was rejectedwith "not a retained coordinator". Now both the retained legacy handle and the current run binding's handle are accepted, so workers can deliver lifecycle mail to either coordinator.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

## AI Review Report

The review checked two legacy orchestration code paths for incorrect fencing and identity routing:

- **`legacy_read_only` false positives**: confirmed fresh-run coordinators are no longer fenced when an unrelated `legacy_adoptions` row exists. Verified replaced legacy coordinators are still fenced after takeover.
- **`worker_done` post-takeover rejection**: confirmed both the old (retained) and new (current binding) coordinator handles are accepted by `isLegacyCoordinatorHandle` after principal revocation. `worker_done` now routes to `run:<id>` with `delivery_contract: current_delivery`.
- **Cross-platform compatibility**: reviewed — no platform-specific differences, shortcuts, labels, paths, or Electron-specific shell behavior were touched. Changes are pure main-side orchestration logic in TypeScript.

Risks flagged by the review: the new `isKnownLegacyCoordinatorIdentity` could in theory under-fence if a committed principal for a different handle existed, but the method requires exact handle match or pane-key match and still delegates to `getUniqueLegacyCoordinatorHandle` for durable inference. No follow-up changes needed.

## Security Audit

- **Input handling**: `isKnownLegacyCoordinatorIdentity` and the fallback path in `LegacyCoordinatorAuthority.resolve` only relax fencing (return `undefined` instead of throwing) for callers that pass `verifyOrchestrationCompatibilityCaller` — they never grant new authority. A caller must still hold a valid launch token and pane binding to act on any run.
- **Command execution**: no new command execution paths introduced.
- **Path handling**: no path handling changes.
- **Auth**: the relaxed fallback still requires `verifyOrchestrationCompatibilityCaller` to return a non-null caller authority (valid launch token + pane key). Unknown callers without evidence still get fenced. No auth bypass.
- **Secrets**: no secrets, tokens, or hashes are logged or exposed.
- **Dependencies**: no new dependencies.
- **IPC**: no IPC surface changes.

## Notes

- The `isLegacyCoordinatorHandle` change is intentionally permissive for revoked principals: after a takeover, workers that still hold the old coordinator's address (`to: term_legacy_coord`) need their `worker_done` to route successfully. The old handle resolves to `run:<id>` via `resolveLegacyWorkerCoordinatorDelivery`, which promotes the message to `current_delivery` — so the new coordinator receives it through the normal run mailbox.
- No platform-specific behavior changes. All changes are in main-process orchestration logic shared across macOS, Linux, and Windows.


brennanb2025:
Taking a look!

brennanb2025:
Thanks for chasing this down — the `worker_done` failure after a takeover is a real bug and worth fixing. I've pushed some changes directly to your branch:

**Split the predicate instead of widening it.** Since this PR was opened, `isLegacyCoordinatorHandle` picked up a second caller on `main` (#11745): it is now also the *fence jurisdiction* check, where returning `true` means "throw `legacy_read_only`". So widening it fixed delivery but simultaneously fenced *more* callers — a replacement coordinator whose pane identity changed after a takeover got a dead-end `legacy_read_only` instead of the actionable `run_required`. There's now a separate `isLegacyCoordinatorDeliveryTarget` for the recipient side, and `isLegacyCoordinatorHandle` is back to its `main` form. Added a test that fails if it ever gets widened again.

**Kept the permit in step with the router.** The permit accepted any handle bound as the Run's `coordinator_handle`, but `resolveLegacyWorkerCoordinatorDelivery` only promotes to `run:<id>` when the principal is not `committed`. A coordinator restarting inside the legacy coordinator's own pane lands in that gap, and the message gets written to a mailbox no reader can see. The permit now mirrors the router's takeover test exactly.

**Dropped `isKnownLegacyCoordinatorIdentity`.** Your `LegacyCoordinatorAuthority.resolve` hunk was lost when `main` got merged in, so the method had no callers left. That half of this PR — `legacy_read_only` on fresh runs — is already fixed on `main` by #11745/#11802, so I removed the duplicate tests for it as well. This PR is now scoped to just the takeover-delivery fix.

**Covered the `ask` path**, which the original change also affected but did not test.

One follow-up worth doing separately: the takeover test is duplicated between the permit and `resolveLegacyWorkerCoordinatorDelivery`. Pulling it into one shared helper would make that drift impossible.

brennanb2025:
Follow-up on my own earlier commit, not on your change — flagging it here so the diff history stays readable.

`019c14189c` drops one clause from `isLegacyCoordinatorDeliveryTarget`:

```diff
-    const run = this.getRunRaw(runId)
-    if (run?.coordinator_handle !== terminalHandle || !run.coordinator_pane_key) {
+    if (this.getRunRaw(runId)?.coordinator_handle !== terminalHandle) {
```

The standard shape for this permit is "the recipient is the Run's bound coordinator **and** the router would treat this as a takeover" — and the router (`resolveLegacyWorkerCoordinatorDelivery`) tests only `principal?.status !== 'committed'`. The extra `coordinator_pane_key` term had no counterpart there. It was also unreachable: all five `runs`-table write sites set `coordinator_handle` and `coordinator_pane_key` together (`createRun`, `bindRun`, `unbindOtherRunsForPane`, the adoption reset, and the legacy insert that sets neither), and `requireCallerPane` rejects empty pane keys. Had it ever fired it would have been the wrong sign — denying mail the router would have promoted to the readable `run:<id>` mailbox.

Behavior is unchanged: the 6 tests in `orchestration-legacy-takeover-delivery.test.ts` pass identically before and after, and 132 tests across the 11 legacy-orchestration suites are green.

Your diagnosis and the goal of #11572 are what ships here — the delivery permit and both call sites are yours. Remaining known follow-up, deliberately **not** done in this PR to keep it to one thing: the takeover test now appears in two places (the permit and the router). Extracting it into a single private predicate would make future drift structurally impossible, but it touches code outside this PR's diff and is better as its own change.

Issue #11485, opened 2026-07-30
[Bug]: Tasks work-item pagination broken again — shows 34 pages but page clicks do nothing


fsdwen:
> **English translation** added for maintainers. Original 中文 / Chinese text is preserved below for reference.

### English title

`[Bug]: Tasks work-item pagination broken again — shows 34 pages but page clicks do nothing`

### Operating system

macOS / Linux

### Orca version

1.4.162-rc.0

## Summary

GitHub Issue pagination in the Tasks panel is broken again: the bottom PaginationBar shows 34 pages (`totalPages=34`), but clicking page-number buttons does not navigate — the page never changes. This was previously fixed by stablyai/orca#8680 (switched from cursor-based to page-number Search API), but the bug has returned.

## Steps to reproduce

1. In the Tasks panel, select any GitHub repo (or multiple repos) with ≥1000 issues
2. After Tasks data loads, the bottom PaginationBar shows "34" pages
3. Click any page that has not been loaded yet (e.g. page 5, 10, or 34)
4. **Expected**: Navigate to that page and load its data
5. **Actual**: After clicking a page button there is no response; UI stays on page 1

### Root cause 1: `countedTotalPages` exceeds pages actually reachable via GitHub Search API

`countWorkItemsAcrossRepos` computes total pages from the Search API `total_count` (true issue count in the repo), which can exceed 1000 (the Search API hard limit).

* `countWorkItemsAcrossRepos` → `Math.ceil(count / perRepoLimit)` → e.g. 1000+ issues, `perRepoLimit=30` → 34 pages
* But GitHub Search API **returns at most 1000 results**. Offsets beyond 1000 yield an empty array
* 34 pages × 30 items/page = 1020 > 1000. Page 34 (1-indexed) with `per_page=30&page=34` returns 0 records
* `handleLoadNextPage` silently returns when `items.length === 0` — **no page change, no error**

### Root cause 2: `paginationLoading` can get stuck

`handleLoadNextPage` has a stale-request guard via `paginationGenerationRef`:

```ts
} finally {
  if (paginationGenerationRef.current === requestGeneration) {
    setPaginationLoading(false)
    setLoadingTargetPage(null)
  }
}
```

Race:

1. User clicks a page → `setPaginationLoading(true)`
2. Search/filter changes → `useEffect` bumps generation + `setPaginationLoading(false)`
3. User clicks a page again → `setPaginationLoading(true)`, new request starts
[...]

PR #11584, opened 2026-07-30
fix(tasks): make GitHub pagination honest — cap unreachable pages, survive background refreshes, explain empty pages

brennanb2025:
## Summary

Refs #11485 (Linear STA-2972) — **partial fix, does not close the issue.** Addresses root causes 1 and 2 from the report; root cause 3 (cross-source page-count mismatch) is still open. See the status comment on #11485. Tasks-panel GitHub pagination advertised pages that could never load, and page clicks silently did nothing. Two independent live-reproduced defects:

1. **Unreachable advertised pages** — `totalPages` came from the Search API `total_count`, but GitHub's Search API rejects requests past its first-1000-results window with HTTP 422. On stablyai/orca the bar advertised 39 pages; clicking page 39 fired a 422 per repo and the UI silently stayed on page 1.
2. **In-flight navigations discarded** — the pagination-reset effect and the work-items fetch effect were keyed on the `selectedRepos` **array identity**, which the repos store refreshes on every `repos:changed` event. Any background repo refresh mid-click bumped the request generation (discarding the fetched page) and reset the pager to page 1. Live A/B: clicking page 3 while forcing `fetchRepos()` mid-flight → stuck on page 1 before, lands and stays on page 3 after.

The fix: cap advertised pages at `ceil(1000 / perRepoLimit)`; key both effects on a stable selection string covering every repo field the requests read; and classify empty page loads three ways — proven window 422 (toast + a separate never-raised `provenPageLimit`), load failure on either envelope channel including newly-plumbed PR-side errors (toast, no clamp), or genuine end-of-data (withdraw the speculative fallback page; neutral "no more results" toast when a real count over-advertised). `deriveAdvertisedTotalPages` makes the result independent of count-vs-probe arrival order.

**ELI5:** the page buttons were promises the app couldn't keep — GitHub only ever hands back the first 1000 search results, but the app did the math on the full total and drew buttons for pages GitHub refuses to serve; clicking one failed invisibly. Worse, any background refresh threw away a page you'd just asked for. Now the app only draws buttons it can honor, tells you when something actually failed, and background refreshes can't cancel your click.

## Screenshots

After — bar capped to the reachable window (30 pages at this per-repo limit, was 39) and the **last** page loads:

After — page-3 click survives a forced mid-flight `fetchRepos()` (previously silently discarded; "before" evidence is the live console 422 + snapshot showing the pager stuck on page 1, captured during the repro):

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — targeted instead: the two owning suites plus all of `src/main/github/` (41 files / 795 tests green), covering the outcome matrix, clamp interleavings for both arrival orders, selection-key stability, both envelope wire tests (issue + PR side, incl. non-window-422 demotion), the classifier window-phrase contract, and cap boundaries at shipped limits (36 → 28, 25 → 40). Reviewer also ran the CI-style react-doctor changed-lines gate (clean).
- [ ] `pnpm build` — left to CI.
- [x] Added or updated high-quality tests that would catch regressions (see above; the empty-page decision logic was extracted into pure, unit-tested functions specifically so these races are pinned).

## AI Review Report

Adversarial review loop: **8 rounds, a fresh reviewer each round; round 8 came back clean (zero blocker/major).** The loop caught and fixed, per round: (1) false error toast on the designed end-of-data probe + the 422 traveling on an envelope channel `failedCount` never counted + the fetch effect still resetting on array identity; (2) stale catalog wording making the two toasts identical + a PR-tab silent clamp regression; (3) the clamp reading a click-time closure instead of the committed count; (4) sibling envelope 403/404s masked by the window branch + count-merge hardening; (5) that same count-merge hardening inverting the "a real count must not shrink" invariant when the probe won the race — fixed by splitting the proven window limit into its own state slot; (6) PR-side failures indistinguishable from end-of-data → plumbed `errors.prs` through the envelope; (7) two literal test mocks missing the new classifier + no producer-side test + closure toast race + wrong toast copy → ref mirror, neutral copy, translations; (8) clean.

Cross-platform: explicitly checked macOS/Linux/Windows — no shortcuts, labels, paths, or shell behavior touched; the change is renderer state logic plus main-process gh-CLI error classification (platform-neutral string handling); SSH/remote-runtime verified (the envelope error channels survive both transports verbatim, confirmed by reviewer trace + wire tests against the runtime RPC shape). Performance review (perf skill): pass-with-notes — keying the fetch effect off array identity is a genuine win (removes a per-`repos:changed` network fan-out; the selection-key memo costs ≤175µs even at 90 repos), and the cap removes a guaranteed-dead request fan-out against the 30/min search quota.

## Security Audit

No auth, secrets, or dependency changes. Error messages from `gh` stderr are classified in main; the renderer keys decisions on the classified **type** only. One reviewer note (accepted, no behavior change today): `classifyListPrsError` echoes trimmed stderr in its message — currently consumed only by a console.warn; a curated message map is a follow-up if a UI banner ever surfaces it. IPC surface: one additive optional field (`errors.prs`) on an existing envelope; no new handlers. Input handling: the page number is clamped/floored before use; interpolated toast values are numeric strings.

## Notes

- Trade-offs: (1) the cap is `ceil`, not `floor`: GitHub serves any search page whose **start** offset is inside the 1000-result window, so the final page is served even when it crosses the boundary (verified live — `per_page=33&page=31` returns 33 rows, `page=32` 422s; same at 30/34 and 36/28). `floor` would silently drop one reachable page; (2) the GitLab issues/MRs effect shares the repurposed selection key, so a GitHub identity resolution now re-fires it once per repo per session (bounded, correct-leaning); (3) `2`-vs-`02`-style numeric ties in Source Control ordering are unaffected here (that's the sibling PR).
- Known pre-existing issue left deliberately untouched (reviewer-flagged, same symptom class, different mechanism): the one-shot landing-refresh probe can still yank an early page click back to page 1 on the cached-landing path (`TaskPage.tsx` landing `.then` has no generation guard). Follow-up candidate.
- Follow-up candidates from the perf review: a pre-existing store-selector hot path (`selectedWorkItemsCacheEntries` runs a per-repo projection on every store write while Tasks is open), and hoisting one shared per-repo context map for the five memos that each recompute it.
- The window-422 toast is reachable only when the count itself failed (a successful count already caps the bar below the window) — it's the backstop, not the primary defense.

### DIFF
--- a/src/main/github/client-work-items-query-paging.test.ts
+++ b/src/main/github/client-work-items-query-paging.test.ts
@@ -63,7 +63,8 @@ vi.mock('./gh-utils', () => ({
   release: releaseMock,
   _resetOwnerRepoCache: vi.fn(),
   classifyGhError: (stderr: string) => ({ type: 'unknown', message: stderr }),
-  classifyListIssuesError: (stderr: string) => ({ type: 'unknown', message: stderr })
+  classifyListIssuesError: (stderr: string) => ({ type: 'unknown', message: stderr }),
+  classifyListPrsError: (stderr: string) => ({ type: 'unknown', message: stderr })
 }))
 
 vi.mock('../git/runner', () => ({
@@ -264,6 +265,22 @@ describe('listWorkItems query paging', () => {
       { cwd: '/repo-root' }
     )
     expect(items.map((item) => item.number)).toEqual([4, 3])
+  })
+
+  it('lifts a swallowed PR-side failure onto errors.prs instead of reading as end-of-data', async () => {
+    getIssueOwnerRepoMock.mockResolvedValueOnce({ owner: 'acme', repo: 'widgets' })
+    getOwnerRepoMock.mockResolvedValueOnce({ owner: 'acme', repo: 'widgets' })
+    // Non-availability failure (plain 403): swallowed into [], must still surface.
+    ghExecFileAsyncMock.mockRejectedValueOnce(new Error('HTTP 403: Forbidden'))
+
+    const envelope = await listWorkItems('/repo-root', 10, 'is:pr is:open', 2)
+
+    expect(envelope.items).toEqual([])
+    expect(envelope.errors?.issues).toBeUndefined()
+    expect(envelope.errors?.prs).toEqual({
+      type: 'unknown',
+      message: expect.stringContaining('HTTP 403')
+    })
   })
 
   it('filters pull request rows out of issue Search API results', async () => {
--- a/src/main/github/client-work-items.test.ts
+++ b/src/main/github/client-work-items.test.ts
@@ -55,7 +55,8 @@ vi.mock('./gh-utils', () => ({
   release: releaseMock,
   _resetOwnerRepoCache: vi.fn(),
   classifyGhError: (stderr: string) => ({ type: 'unknown', message: stderr }),
-  classifyListIssuesError: (stderr: string) => ({ type: 'unknown', message: stderr })
+  classifyListIssuesError: (stderr: string) => ({ type: 'unknown', message: stderr }),
+  classifyListPrsError: (stderr: string) => ({ type: 'unknown', message: stderr })
 }))
 
 vi.mock('../git/runner', () => ({
--- a/src/main/github/client.ts
+++ b/src/main/github/client.ts
@@ -54,6 +54,7 @@ import {
   release,
   classifyGhError,
   classifyListIssuesError,
+  classifyListPrsError,
   ghRepoExecOptions,
   githubRepoContext,
   getRemoteUrlForRepo,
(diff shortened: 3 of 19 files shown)


brennanb2025:
Maintainer QA for `689aa56b072312f04f11397a772f71954d12b4c2`: 265 focused tests, typecheck, lint/code-quality/localization gates, and a 90-repo selection-key benchmark (8.448 µs mean across 10,000 runs) passed; a fresh `gpt-5.6-sol`/`xhigh` reviewer returned CLEAN. Isolated Electron validation confirmed 28 reachable 36-row pages, page 29 absent, page 3 surviving a background repo refresh, and distinct unreachable/load-failed/end-of-data feedback; mobile QA was not applicable. Broad-suite note: 41,487 tests passed and four unrelated failures were classified—Git-guard env duplication plus two load/interference flakes—and all three affected files passed isolated reruns.

Evidence: [reachable page count](https://github.com/user-attachments/assets/92e653c5-cbbf-42cd-ac9f-4fe744641627) · [last reachable page 28](https://github.com/user-attachments/assets/f94fb44a-023a-4691-8018-1a83750b5bda) · [background refresh on page 3](https://github.com/user-attachments/assets/4b960258-df2a-49fc-adf2-19978d17f8ec) · [window-unreachable state](https://github.com/user-attachments/assets/a7987fe2-e098-488f-8229-fc25840b332b) · [load-failed state](https://github.com/user-attachments/assets/e7ad569b-4620-4991-adf5-632abb95b5d0) · [end-of-data state](https://github.com/user-attachments/assets/a8a83986-5028-41f5-a834-67c5f75cddcc)

AmethystLiang:
@coderabbitai[bot] fixed

AmethystLiang:
@coderabbitai[bot] Fixing. Will be in the next commit

PR #11585, opened 2026-07-30
fix(settings): widen font size input so values are fully visible

qiuyongjin:
## Description

Widened the font size input from `w-14` to `w-24` so two-digit values (10-24px) display fully.

## Screenshots

No visual change — the input width increased from 3.5rem to 6rem, all existing layout remains unchanged.

## Testing

- [x] Local lint passes (`oxlint`, `oxfmt`)
- [x] Font size values 10-24 all display fully in the input field
- [x] Increment/decrement buttons work correctly at all values

## AI Review Report

No AI-generated code in this change — manual one-line className adjustment.

## Security Audit

No security impact — purely a CSS width change on a controlled numeric input.

## Notes

N/A

### DIFF
--- a/src/renderer/src/components/settings/TerminalFontSizeSetting.test.tsx
+++ b/src/renderer/src/components/settings/TerminalFontSizeSetting.test.tsx
@@ -1,14 +1,10 @@
 // @vitest-environment happy-dom
 
-import { readFileSync } from 'node:fs'
-import { resolve } from 'node:path'
 import { act } from 'react'
 import { createRoot, type Root } from 'react-dom/client'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import type { GlobalSettings } from '../../../../shared/types'
 import { TerminalFontSizeSetting } from './TerminalFontSizeSetting'
-
-const MAIN_CSS = resolve(__dirname, '../../assets/main.css')
 
 vi.mock('@/i18n/i18n', () => ({
   translate: (_key: string, defaultValue: string) => defaultValue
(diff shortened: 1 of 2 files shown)


qiuyongjin:
image

brennanb2025:
Thanks for this — the diagnosis and the goal are right, and I've pushed one commit to your branch on top of it.

What I changed and why:

- **Added `number-input-clean` to the same input.** The clipping isn't really a width problem: `<input type="number">` reserves space for Chromium's native spin buttons *inside* the field, and that reserved gutter is what was cutting off the second digit. Measured in the app, at the old `w-14` the input's `scrollWidth` exceeds its `clientWidth` by 2px without that class and by 0px with it. `number-input-clean` is what the other three numeric settings inputs already use (`SettingsFormControls` `NumberField`, `TerminalAdvancedSection`, `GeneralEditorSettingsSection`), so this makes Font Size consistent with them instead of being the one field that still renders native spinners next to its own − / + buttons.
- **Kept your `w-24`.** It matches the width those sibling inputs use, and with the spin gutter gone the value now sits properly centred rather than nudged left.
- **Added `TerminalFontSizeSetting.test.tsx`.** The component had no tests; this covers the fix plus the stepper bounds. I checked it's a real guard — reverting to `w-14` fails two of its cases.

One thing I noticed but deliberately did **not** fix here, so it doesn't get lost: the field can't actually be typed into. `onChange` commits on every keystroke and rejects anything outside 10–24, so the first digit of a two-digit entry is rejected and the controlled input snaps back — only the − / + buttons really work. That's a separate bug from the one this PR fixes and it deserves its own PR.

Heads up that this also fixes #12877, which was filed for the same symptom.

brennanb2025:
Thanks for catching this — I pushed a small follow-up to the PR branch.

The clipping came from the native number-input spinner gutter, not the 56px width itself. The field now reuses the existing number-input-clean treatment and stays at w-14, so the value remains centered without making the stepper group wider. I also trimmed the regression test to assert the component contract instead of reparsing global CSS.

Validated with the focused component tests, full typecheck, and Electron: no input overflow, and the −/+ controls still round-trip 14 → 15 → 14.

PR #11612, opened 2026-07-30
fix(terminal): serialize startup context with user paste

OrcaWin:
## Summary

- serialize generated startup context and user paste operations per PTY
- preserve user typing and large multiline pastes while still injecting linked-work-item context
- keep the uncontended input path synchronous and isolate independent PTYs

## Root cause

The normal linked-worktree path defaults to `promptDelivery: 'draft'`. It does not send Enter after injecting the context.

The confirmed transport failure was nested bracketed-paste framing. A Windows multiline paste over 64 KiB is written as one bracketed frame across acknowledged 16 KiB chunks, with event-loop yields between chunks. Startup-owned linked context is also sent as a bracketed paste. Both producers shared the mounted PTY transport, but there was no whole-operation owner.

That allowed this byte order:

```text
user START
  generated START
  generated context
  generated END
  remaining user content with CR newlines
user END
```

The generated `END` could close Codex's paste mode early. Carriage returns in the remaining user chunks were then outside the intended user paste frame and could be interpreted as submit input.

The fix gives each PTY a FIFO paste transaction covering the complete `START ... END` operation. If the user paste begins first, linked context waits and is injected afterward. If generated context begins first, the user paste waits. Typing and both payloads are preserved; normal draft launch still sends no Enter.

The explicit `submit-after-ready` flow is intentionally unchanged and is not part of this incident.

## Reproduction

The regression test deterministically exercises both legal races:

- user chunked paste opens before startup context injection
- startup context opens before the user chunked paste

Before the fix, the user-first case produced a nested generated frame and exposed later carriage returns outside the intended paste frame. After the fix, each frame completes before the other begins.

The exact real-Codex visual outcome—part of the text submitting while the remainder appears as `[Pasted n characters]`—was not reproduced end to end on the broken revision. This PR fixes the underlying overlapping-frame race that can produce that class of split-paste behavior. If the same UI symptom recurs after this change, it has another input source and needs a separate trace.

## Electron validation

Validated on Windows Electron/ConPTY at PR head `8c1946e2da633fe58c05ad7c24f8a46631298fdd` through the production startup-owned draft path.

| User input owns the open paste frame | Both inputs complete without submission |
|---|---|
| Typed input and large user paste are in flight before linked context | Typed input, large paste, and linked context are preserved with no nested frame or unframed Enter |

The validation held the acknowledged user paste after its opening marker, allowed the fake Codex readiness signal to fire, then released it. The terminal verified:

- typed input was preserved
- the >64 KiB user paste was preserved
- linked-work-item context was still injected
- no bracketed-paste frames were nested
- no carriage return occurred outside a paste frame

The validation harness and screenshots are not part of the PR diff.

## Testing

- [x] 595 focused terminal, startup-draft, paste, and agent-delivery tests
- [x] renderer web typecheck
- [x] changed-file oxlint and React Doctor lint
- [x] max-lines ratchet and `git diff --check`
- [x] Electron Vite E2E build
- [x] Windows Electron/ConPTY production-path validation
- [x] same-model review returned clean for the normal linked-worktree draft scope

## Performance and compatibility

The transaction registry stores only active/queued operations and deletes a PTY entry when its tail settles. Uncontended operations start synchronously, preserving the existing input latency. Different PTYs do not block each other. The shared renderer path covers local, WSL, SSH, and remote-runtime terminals without changing Git, provider, worktree, or folder-workspace behavior.


OrcaWin:
## Superseded validation

The earlier version of this comment exercised explicit `submit-after-ready`, not the normal linked-worktree draft path, so it was not evidence for the reported incident.

The PR has been reworked around the reproduced nested bracketed-paste race. Current Windows Electron/ConPTY validation and screenshots are in the updated PR description:

| User paste frame in flight | Serialized completion |
|---|---|
| Typed input and large user paste are in flight before linked context | Typed input, large paste, and linked context are preserved |

PR #11613, opened 2026-07-30
fix(win): harden startup during partial updates

OrcaWin:
## Summary

- bundle `zod` into the Electron main-process output so startup does not depend on the later `resources/node_modules` deployment
- prepend a fatal bootstrap guard before Rollup's generated `require()` prelude, then replace it with the normal durable error guard after the source graph loads
- cover both the bundling contract and an exact pre-import `Cannot find module 'zod'` failure

## Root cause

The Windows NSIS install updates the live application directory non-transactionally. In the captured failure, the new `app.asar` and plugin-host chunk were visible 8.343 seconds before `resources/node_modules/zod` existed. The new main bundle immediately executed `require("zod")`, lost that race, and Electron displayed its native fatal-error dialog. That failed main process remained resident, held the updater lock, and prevented the runtime/pairing listener from starting.

This change uses the packaging-safe alternative to an installer rewrite: the bootstrap-critical schema dependency is part of the main bundle itself. It also ensures any future exception that occurs before source-level guards load exits instead of stranding a dialog-only Electron process.

## User-visible behavior

| | Before | After |
|---|---|---|
| Launch during the vulnerable update window | Native `Cannot find module 'zod'` dialog; no Orca window | `zod` is already in the startup bundle, so launch proceeds normally |
| Remote runtime/pairing | Never starts | Starts with the normal application runtime |
| Updater after a bootstrap failure | Failed Electron process can remain resident and block replacement | Pre-import fatal failures exit with code 1 on the next event-loop turn |
| Ordinary launches | Normal | Unchanged; the temporary bootstrap guard is removed when the durable main-process guard installs |

## Validation

- `pnpm exec vitest run config/scripts/electron-vite-output-contract.test.ts src/main/startup/bootstrap-fatal-exit-guard.test.ts src/main/startup/main-process-error-guards.test.ts` (17 tests)
- `pnpm run build:desktop`
- `pnpm run typecheck:node`
- changed-file Oxlint, type-aware Oxlint, React Doctor, and Oxfmt checks
- inspected emitted `out/main`: bootstrap guard is the first code in `index.js`; no bare `require("zod")` remains

A final `electron-builder --dir` package was not produced locally because this host lacks the Visual Studio toolchain required to rebuild `windows-native-registry`; the complete desktop build and packaging-relevant emitted-output checks passed.

### DIFF
--- a/config/scripts/electron-vite-output-contract.test.ts
+++ b/config/scripts/electron-vite-output-contract.test.ts
@@ -1,6 +1,9 @@
 import { readFileSync } from 'node:fs'
+import { EventEmitter } from 'node:events'
+import { runInNewContext } from 'node:vm'
 import { describe, expect, it } from 'vitest'
-import { electronViteConfig } from '../../electron.vite.config'
+import { createBootstrapFatalExitBanner, electronViteConfig } from '../../electron.vite.config'
+import { BOOTSTRAP_FATAL_EXIT_GUARD_KEY } from '../../src/main/startup/bootstrap-fatal-exit-guard'
 
 const targetConfig = readFileSync('config/electron-vite-target.config.ts', 'utf8')
 const devRunner = readFileSync('config/scripts/run-electron-vite-dev.mjs', 'utf8')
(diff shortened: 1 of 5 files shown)


OrcaWin:
## Electron validation (Windows)

Validated PR head 8e6f05cdf from the intended `prowfish @ OrcaWin/diagnose-git-issue` dev instance.

- Visible behavior: the full Orca window rendered normally, then navigated through the real Orca Mobile surface; Playwright reported 0 renderer console errors.
- Backing state: first-window startup services settled, runtime graph status was `ready`, desktop status was `available`, and the Windows runtime WebSocket listener reported ready on port 6769.
- Build evidence: `pnpm run build:desktop` passed; emitted `out/main/index.js` starts with the fatal guard before diagnostics/generated code and has no bare `require("zod")`. Targeted startup/error-guard tests passed 17/17.
- Remaining gap: the live NSIS partial-copy timing window was not manually recreated in the dev app; the packaged Windows crash-survival CI job covers that scenario and passed.

| Normal startup | Interactive Mobile surface |
|---|---|
| Orca normal startup | Orca Mobile surface |

PR #11616, opened 2026-07-30
[P1] fix(terminal): require a physical Enter for Windows IME process-enter

nwparker:
## Summary
- On Windows, Chromium reports every key an active IME consumes as `key='Process'`/`keyCode=229`, so `isTerminalImeProcessEnter` matched any modified IME-consumed key, not just Enter.
- Shift+\<jamo\> (e.g. Shift+KeyQ for ㅃ) or an MS-IME Ctrl chord (Ctrl+KeyI) was misread as Shift/Ctrl+Enter; `keyboard-handlers.ts` then synthesized an Enter event and `terminal-shortcut-policy.ts` injected `\x1b\r` / `\x1b[13;5u` into the shell.
- Fix: `isTerminalImeProcessEnter` now also requires the physical key code to be `Enter` or `NumpadEnter`, which Chromium reports faithfully even when the IME consumes the key.
- No other behavior changed; the synthesized-event path in `keyboard-handlers.ts` already forwards the original event with its `.code`.

## Test plan
- Extended `terminal-ime-deferred-newline.test.ts` unit cases: rejects `code: 'KeyQ'` (shift) and `code: 'KeyI'` (ctrl), accepts `code: 'Enter'` and `code: 'NumpadEnter'` for both shift and ctrl kinds.
- Extended the Windows IME harness in `keyboard-handlers-ime.test.tsx`: IME-consumed Shift+KeyQ / Ctrl+KeyI during composition are not swallowed and send no input; Process+code Enter/NumpadEnter still defers and sends `\x1b\r` after composition (regression guard for #11293).
- Verified the new tests fail on the unfixed code (`git stash` of the production change): 4 failures, e.g. `does not treat an IME-consumed 'Shift+KeyQ' as a modified Enter` — `AssertionError: expected true to be false` on `defaultPrevented`; all 45 tests pass with the fix restored.
- Ran all dependent test files (`xterm-bypass-policy`, `keyboard-handlers`, `terminal-ime-deferred-newline`, `keyboard-handlers-ime`, `terminal-pane-split-writer-paths`): 94 passed. `npx oxlint` and `tsc -p config/tsconfig.tc.web.json` clean.

Found by the production release scan of v1.4.162..v1.4.163-rc.0.

### DIFF
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
@@ -280,7 +280,10 @@ describe('Windows IME keyboard ownership', () => {
   it.each([
     // Why: an active IME reports every consumed key as Process/229, e.g. Shift+ㅃ or an MS-IME Ctrl chord.
     { label: 'Shift+KeyQ', code: 'KeyQ', modifier: { shiftKey: true } },
-    { label: 'Ctrl+KeyI', code: 'KeyI', modifier: { ctrlKey: true } }
+    { label: 'Ctrl+KeyI', code: 'KeyI', modifier: { ctrlKey: true } },
+    // Ctrl+K / Ctrl+W are bound on Windows, so a leaked Process chord would clear or close the pane.
+    { label: 'Ctrl+KeyK', code: 'KeyK', modifier: { ctrlKey: true } },
+    { label: 'Ctrl+KeyW', code: 'KeyW', modifier: { ctrlKey: true } }
   ])('does not treat an IME-consumed $label as a modified Enter', ({ code, modifier }) => {
     const harness = createHarness()
     const hook = renderHook(() => useTerminalKeyboardShortcuts(harness.deps))
@@ -299,6 +302,36 @@ describe('Windows IME keyboard ownership', () => {
 
     expect(consumed.defaultPrevented).toBe(false)
     expect(harness.sendInput).not.toHaveBeenCalled()
+    expect(harness.deps.onClearPaneScrollback).not.toHaveBeenCalled()
+    expect(harness.deps.onRequestClosePane).not.toHaveBeenCalled()
+    hook.unmount()
+    harness.dispose()
+  })
+
+  it.each([
+    { label: 'Ctrl+KeyK', code: 'KeyK' },
+    { label: 'Ctrl+KeyW', code: 'KeyW' }
+  ])('stops an IME-consumed $label from reaching window-level shortcuts', ({ code }) => {
+    const harness = createHarness()
+    const hook = renderHook(() => useTerminalKeyboardShortcuts(harness.deps))
+    const windowHandler = vi.fn()
+    window.addEventListener('keydown', windowHandler)
+    harness.startComposition()
+
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keydown', {
+        key: 'Process',
+        code,
+        keyCode: 229,
+        timeStamp: 10,
+        isComposing: true,
+        ctrlKey: true
+      })
+    )
+    vi.runAllTimers()
+
+    expect(windowHandler).not.toHaveBeenCalled()
+    window.removeEventListener('keydown', windowHandler)
     hook.unmount()
     harness.dispose()
   })
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
@@ -31,6 +31,7 @@ function keyboardEvent(
 function createHarness(): {
   deps: KeyboardHandlersDeps
   editable: HTMLInputElement
+  sendInput: ReturnType<typeof vi.fn>
   startComposition: () => void
   terminalInput: HTMLTextAreaElement
   dispose: () => void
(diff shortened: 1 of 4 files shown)


BaeTab:
Confirming this on **Windows 11 + Microsoft 두벌식 (2-set Korean)**, Orca **v1.4.163**. I hit it during normal typing, not while testing edge cases.

### Scope is wider than double consonants

2-set Korean needs Shift for seven jamo, and **all seven reproduce**:

| Jamo | Chord |
|---|---|
| ㅃ ㅉ ㄸ ㄲ ㅆ | Shift + Q/W/E/R/T |
| ㅒ ㅖ | Shift + O/P |

The two **vowels** (ㅒ ㅖ) reproducing is the useful part: they go through different composition logic than the consonants, so this rules out anything jamo-specific and pins it to the modifier-flag-only check — exactly what this PR fixes. Any framing of this as a "double consonant" bug is too narrow.

Words that hit it constantly: `있다`, `했다`, `갔다`, `됐다`, `예를 들어`, `얘기`, `오빠`, `예쁘다`.

It only fires when the Shift-jamo lands **mid-word** (a composition is already pending), consistent with the `hasPendingImeComposition` condition at `keyboard-handlers.ts:502`.

### Both halves of the symptom

1. A newline is injected — `terminal-shortcut-policy.ts:188` → `\x1b\r`.
2. The jamo itself is swallowed — `keyboard-handlers.ts:529` `stopImmediatePropagation()`.

Reading the diff, this PR addresses both: the predicate change stops the synthesized Enter (so `preventDefault` never runs), and the new swallow branch calls `stopImmediatePropagation()` without `preventDefault`, so the IME still composes.

### The red CI is an infrastructure flake

Worth flagging since this is P1 and currently reads as broken:

```
node-pty install: gyp ERR! stack AggregateError [ETIMEDOUT]
[native-runtime] pnpm rebuild node-pty failed.
```

`tests node 24 10/16` failed inside `pnpm rebuild node-pty` — **no test executed**; the whole 1m30s was dependency install. `verify` is only the aggregator job (`pr.yml:376`) and is red solely because of that shard.

The **same shard on Node 26 passed**, as did the other 31 shards, static analysis, typecheck, package, and the Linux X11 IBus IME suite. A re-run should clear it.

### Possible collision with #11550

#11550 edits the same lines of `isTerminalImeProcessEnter` with a semantically identical `code === 'Enter' || 'NumpadEnter'` requirement, derived from Microsoft Pinyin traces. Whichever lands first will force the other to rebase.

One thing worth preserving from this PR: #11550 does not include the swallow branch for IME-consumed chords, so `Ctrl+K` / `Ctrl+W` mid-composition would still resolve through the physical-code fallback in `terminal-shortcut-policy.ts:285-318`.

### Related issue

I filed #11878 for this before finding this PR. Note that the "Likely code path" section in that issue is **wrong** — it blames a stale `heldImeEnterModifiers` Shift being attributed to a later plain Enter. The ㅒ/ㅖ result above disproves that, and the newline appears without any Enter being pressed. This PR's diagnosis is the correct one. I'll correct that issue.

### Two questions, not objections

1. **`code === ''`** — the comment notes a key event without a scan code falls back to the Enter keyup path. Some RDP/VNC clients and on-screen keyboards report an empty `code`; would those lose the composing Shift+Enter deferral they have today? I don't see a test covering it.
2. **Plain composing Enter** — it now takes the new swallow branch rather than falling through. The rationale ("xterm already ignores keyCode 229 while composing") holds while xterm's `_isComposing` and Orca's `hasPendingTerminalImeComposition` agree — is there a case where they diverge?

I have the setup to capture a full native event trace (window-capture keydown/keyup + composition events, including `code`) and a `cat -v` recording showing the injected `^[^M`. I'll follow up with those. Happy to add the ㅒ/ㅖ cases to `keyboard-handlers-ime.test.tsx` if that's useful.

holdn2:
Heads up — I've opened #12120 as a draft that carries both of your commits here
(`d0d7e77` physical-Enter classification, `9f449d7` the shortcut propagation guard)
with `--author` preserved, alongside the minimum hunks from #11937, #11896 and #11870.

The reason for bundling: each of the four PRs closes one event path, but a user typing
빨래 hits several in one composition. I added a regression test for that exact sequence —
a Shift-jamo consumed as `Process`/229, the committing Enter after it, then the rollover
Shift — which fails on `main` and needs your classification fix *and* #11937's keyup gate
to pass.

This is not meant to supersede your PR, and I'd be glad if maintainers just landed yours.
Say the word and I'll close mine. Any correction on how I've represented your work is welcome.

dpfwpfahs1-coder:
Still reproduces on **v1.4.164**, the current latest release — so no shipped build fixes this yet.

**Environment**
- Windows 11 Pro 10.0.26200
- Orca **1.4.164** (updated and restarted; running process confirmed on 1.4.164)
- Microsoft Korean IME, 2-set / 두벌식

**Symptom**
Typing a Shift-jamo (e.g. ㄸ = Shift+E) during ordinary Hangul input in a terminal agent session submits a newline I never typed. This shows up in normal day-to-day work, not while probing edge cases.

Adding this mainly as a version data point rather than new analysis: v1.4.163 already shipped the IME composition integration from #11293, and v1.4.164 is the current release, so the `isTerminalImeProcessEnter` path this PR targets is still reachable in released builds.

The diagnosis here — and in the Greptile summary above — matches exactly what I see. Nothing to add to it. Just flagging that it is hitting real users on the current release, and that this PR has had no review since 2026-07-30.

nwparker:
Closing as superseded by #12265, which landed the physical-Enter classification together with the broader Windows IME shortcut guard and regression coverage. Thank you for isolating and validating the original failure mode.

AmethystLiang:

PR #11623, opened 2026-07-30
Fix duplicate remote worktree delete failures

brennanb2025:
## Summary

- resolve delete-dialog targets through the canonical worktree identity index
- normalize duplicate selected IDs before single/batch confirmation classification
- collapse duplicate identities again at the shared destructive batch boundary
- prevent a successful removal from being followed by a false selector-not-found failure

## Root cause

A create/refresh race can temporarily leave duplicate worktree rows in renderer state. The delete dialog flattened those raw rows, so one selected workspace could enter the serialized per-repository delete queue twice. The first request removed the worktree; the second then correctly reported that the same selector no longer existed.

## Validation

- 39 focused delete-dialog, delete-flow, and identity-index tests passed
- new regressions cover duplicate renderer rows, duplicate selected IDs, one removal, one result, and no error toast
- web, node, and CLI TypeScript checks passed
- changed-file lint, React checks, formatting, and max-lines ratchet passed
- isolated Electron run through the real context menu and confirmation dialog: two raw rows with one ID rendered as one target, issued one removal, removed the Git worktree and directory, and showed no error toast
- repository suite: 41,815 passed; five unrelated failures under the unsupported local Node 26 runtime passed 126/126 when rerun under the required Node 24 runtime with isolated environment state

### DIFF
--- a/src/renderer/src/components/sidebar/DeleteWorktreeDialog.test.tsx
+++ b/src/renderer/src/components/sidebar/DeleteWorktreeDialog.test.tsx
@@ -42,6 +42,10 @@ vi.mock('@/store', () => ({
   )
 }))
 
+vi.mock('@/store/selectors', () => ({
+  useAllWorktrees: () => mocks.state.allWorktrees()
+}))
+
 vi.mock('@/components/ui/dialog', () => ({
   Dialog: ({ open, children }: { open: boolean; children: ReactNode }) =>
     open ? <div>{children}</div> : null,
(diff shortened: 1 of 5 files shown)


brennanb2025:
## Electron validation

The isolated dev profile contained two raw renderer rows for the same throwaway worktree ID. The real workspace context menu still rendered one target and opened the normal confirmation dialog:

Delete confirmation with one canonical target

After confirming, the app issued one removal request, removed the Git registration and checkout directory, closed the dialog, and showed no error toast:

Workspace list after successful removal

PR #11637, opened 2026-07-31
fix(mobile): clear state after closing final tab

brennanb2025:
## What this fixes

Closing the final terminal or browser tab in a mobile session could leave the old tab identity selected. The result was a stale, blank terminal area that looked broken and offered no obvious next step.

## Before / after

- **Before:** Close the last tab → remain on an empty or stale session surface.
- **After:** Close the last tab → see **No tabs in this session** and the **Create Tab** action.

## What changed

- Calculate the post-close tab list from the latest render-synchronized state.
- Clear the active tab and terminal handles immediately when that list becomes empty.
- Add regression coverage specifically for the final-tab close path.

The normal close behavior for sessions that still have other tabs is unchanged.

## Verification

- Focused final-tab regression test passed.
- Mobile typecheck, lint, formatting, and diff checks passed.
- Review covers rapid close, bulk close, empty snapshots, and stale tab references.

### DIFF
--- a/mobile/app/h/[hostId]/session/[worktreeId].tsx
+++ b/mobile/app/h/[hostId]/session/[worktreeId].tsx
@@ -4131,6 +4131,7 @@ export default function SessionScreen() {
         reason: 'user'
       })
       if (response.ok) {
+        const remainingTabs = sessionTabsRef.current.filter((candidate) => candidate.id !== tab.id)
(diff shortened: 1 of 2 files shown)


brennanb2025:
## Simulator evidence

Explicit empty state after closing the final tab:

brennanb2025:
### Attachments

**PR 11637 iPad QA evidence**

| orca-mobile-pr11637-loaded.png | orca-mobile-pr11637-worktree-open.png |
|---|---|
|  |  |

PR #11639, opened 2026-07-31
fix(terminal): honor readline line navigation across IMEs

hyein-cbio:
## Summary
- use physical `KeyboardEvent.code` for readline Ctrl+A / Ctrl+E line navigation
- preserve Kitty CSI-u handling and IME composition ownership
- add regression coverage for Korean, Chinese Pinyin, Japanese, and Russian input sources

## Verification
- Korean and English macOS input sources manually verified in the Electron GUI
- Vitest: 55 tests passed
- oxlint passed
- `pnpm run typecheck:node` passed

Resolves the macOS CJK/Russian input-source issue where `event.key` is layout-dependent.


hyein-cbio:
Closing this PR after further investigation.

The issue also reproduces in Pi running under iTerm2, while raw bash handles the same Korean Ctrl + A/E events correctly in Orca. This means the problem is not Orca-specific, and the proposed Ctrl + A/E interception is not the right general solution.

I'm closing this PR instead of pursuing a terminal-specific workaround. I'll reopen or submit a new PR once the Kitty protocol byte sequence and the correct fix location are confirmed.

PR #11647, opened 2026-07-31
fix(mobile): choose host for new workspace

brennanb2025:
## What this fixes

The mobile **New Workspace** quick action did not make it clear which connected desktop would create the workspace when more than one desktop was available.

## Before / after

- **No connected desktops:** the action remains disabled.
- **One connected desktop:** tapping **New Workspace** opens the creation flow directly on that desktop.
- **Multiple connected desktops:** tapping it first opens a host picker, then creates the workspace on the desktop the user chose.

## What changed

- Build the action from the currently connected host set.
- Add a host picker when the destination is ambiguous.
- Show each desktop's endpoint in the picker so similarly named hosts can be distinguished.
- Preserve the selected host through the workspace-creation route.

## Verification

- Focused quick-action and host-selection tests passed.
- Mobile typecheck, lint, and formatting passed.
- Verified the quick-action layout, picker, and accessibility roles on iPad.


brennanb2025:
Mobile simulator QA on the granted iPhone 17 Pro:

### Zero connected hosts

- `New Workspace` is visibly disabled and reports `enabled: false` through accessibility; `Pair Desktop` remains enabled.
- Tapping the disabled action did not navigate or present a picker.

Zero connected hosts: New Workspace disabled

### Current-branch dev-client retry

- Retested head `58174fbb86` on iOS 26.5 / iPhone 17 Pro through Metro from this worktree; the bundle completed without a redbox.
- Exactly one host was connected (`Host 26`, Direct / Tailscale). Accessibility reported `New Workspace` as an enabled button.
- Tapping `New Workspace` opened the `Create Workspace` sheet directly on Host 26; no host picker appeared. The destination host remained visible behind the sheet.
- The controlled environment did not provide a second connected host, so the multi-host picker still has automated/static coverage but no native screenshot in this run.

Current PR branch loaded with Host 26 connected

One connected host opens Create Workspace directly on Host 26

## Native mobile QA proof

The prior multiple-connected-host native-QA gap is closed. On the exclusively granted iPhone 17 Pro (50FD514C-96F6-446E-AC90-D07A24CAB3A0), Orca emulator QA verified: zero connected hosts disables New Workspace (including AX enabled=false); one connected host opens Create Workspace directly and retains the selected host; two connected hosts opens an explicit Create Workspace On picker whose host rows are native AX buttons; choosing Host 29 waits for drawer close, opens the destination cleanly, and preserves Host 29 through navigation.

Zero connected hosts: New Workspace disabled

One connected host: direct Create Workspace flow

Two connected hosts: explicit accessible host picker

Chosen Host 29 preserved at Create Workspace destination

Fresh retry after the Orca restart reattached only the granted UDID and confirmed the production host profiles are present, but the restored ad-hoc simulator shell cannot unlock SecureStore/keychain entries (A required entitlement is not present), so reconnecting those production profiles was environment-limited. No credential contents were read or printed; the earlier controlled two-connected-host run above remains the valid proof.

PR #11670, opened 2026-07-31
fix(file-explorer): open symlink files when stat fails

### LINKED ISSUE #11654: [Bug]: Symlink file cannot be opened, showing "Cannot open symlink target"
### Operating system

macOS

### Orca version

1.4.162

### Details

**Short summary:**
Clicking on a soft symlink file displays a "Cannot open symlink target" error instead of opening the target file directly (expected behavior, matching VS Code).

**Reproduction steps:**
1. Create a soft symlink (symbolic link) pointing to a target file.
2. In the file explorer, click on the soft symlink file.
3. Observe the error message: "Cannot open symlink target".

Image

xianjianlf2:
## Summary
- allow file symlink activation to continue to the editor when target stat fails
- keep symlink-to-directory expansion unchanged when stat succeeds
- add a regression test for stat failures on symlink file activation

Fixes #11654.

## Tests
- pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/right-sidebar/useFileExplorerHandlers.test.ts
- pnpm exec oxlint src/renderer/src/components/right-sidebar/useFileExplorerHandlers.ts src/renderer/src/components/right-sidebar/useFileExplorerHandlers.test.ts
- git diff --check

### DIFF
--- a/src/renderer/src/components/right-sidebar/FileExplorer.tsx
+++ b/src/renderer/src/components/right-sidebar/FileExplorer.tsx
@@ -499,6 +499,7 @@ function FileExplorerFiles(): React.JSX.Element {
       toggleDir: hasNameFilter ? handleToggleNameFilterDir : toggleDir,
       loadDir,
       statPath,
+      authorizeExternalPath: window.api.fs.authorizeExternalPath,
       markPathAsDirectory,
       setSelectedPath: setSingleSelectedPath,
       scrollRef
--- a/src/renderer/src/components/right-sidebar/file-explorer-deferred-dir-toggle.test.ts
+++ b/src/renderer/src/components/right-sidebar/file-explorer-deferred-dir-toggle.test.ts
@@ -32,6 +32,7 @@ function renderHandlers(toggleDir: (worktreeId: string, dirPath: string) => void
       toggleDir,
       loadDir: vi.fn().mockResolvedValue(true),
       statPath: vi.fn().mockResolvedValue({ isDirectory: true }),
+      authorizeExternalPath: vi.fn(),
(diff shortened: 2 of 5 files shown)


brennanb2025:
Thanks for chasing this down — the diagnosis is right. I've pushed a follow-up commit to your branch (heads up so it isn't a surprise).

Testing it in the app turned up a second half to the bug. The reason `statPath` rejects is that the main process's path allow-list refuses any symlink whose real target resolves outside the workspace — and `fs:readFile` goes through that same check. So with only the toast removed, clicking a symlink that points outside the repo swapped the error message for an editor tab reading "Access denied: path resolves outside allowed directories".

What the follow-up adds:

- On activation of a symlink row in a local workspace, grant that one path access first — the same thing that already happens when you click a file path in a terminal or open an absolute path from Quick Open. With that in place the stat *and* the read succeed and the file opens. It also fixes symlinks pointing at a directory outside the workspace, which hit the same error.
- Skipped for SSH and remote-runtime workspaces, where the host owns that boundary.
- The grant sits inside the existing try, so if it ever fails the click still falls through to the editor's real error instead of dead-ending silently.
- Tests: the grant must land before the stat, remote-owned workspaces must not get a local grant, and a failed grant must still open the file. Two other test files that call the hook needed the new argument to keep the typecheck green.

Worth knowing: #11681 and #4556 are also open against #11654 and take a different route — relaxing the allow-list itself rather than granting per activation. A maintainer will decide which approach lands.

Checked on macOS against a symlink to a file outside the repo, a symlink to a directory outside the repo, a symlink inside the repo, and a broken symlink (that one now shows a plain "file not found" in the tab).

PR #11676, opened 2026-07-31
fix(agent-hooks): fail closed without detected agent allowlist

### LINKED ISSUE #11641: [Bug]: Orca auto-creates hook config directories for every agent regardless of whether they are installed
## Thanks first

Thanks to the Orca team for the great work — I use Orca daily and it has been a solid tool overall.

## The problem

Orca auto-creates hook config directories and hook files for **every agent it supports**, regardless of whether those agents are actually installed on the local machine.

My environment is **Linux**. I only have a subset of agents installed locally (e.g. Claude Code, Codex, Grok, Kimi Code, etc.). **Cursor, Devin, Copilot, Gemini, Factory/Droid and others are NOT installed** on this machine. But after I manually delete the config directories of those uninstalled agents (e.g. `~/.cursor`, `~/.config/devin`, etc.), the next time I open Orca, those directories get recreated silently at some point — with Orca-injected hook files inside them.

I consider this a low-level mistake — Orca should not litter config directories for agents the user never installed.

## Steps to reproduce

1. On Linux, install only a few agents (e.g. Claude Code + Codex), leave Cursor, Devin, Copilot, Gemini etc. uninstalled
2. Manually delete `~/.cursor`, `~/.config/devin` and other directories for agents you never installed
3. Use Orca normally — after a while, or after restarting Orca
4. Observe that those deleted directories reappear, containing Orca hook files

## Expected behavior

Before creating hooks, Orca should **check whether the agent is actually installed on the local machine** — only install hooks for agents that are genuinely present.

A simple detection approach: check if the agent binary exists in `$PATH`, or if the agent's own config directory already exists. No need to create directories for agents that have never been installed.

## Investigation results

I did a thorough investigation on my machine to document exactly what Orca did.

### Agents actually installed (binary in PATH)

| Agent | Path | Status |
|-------|------|--------|
| claude | `~/.local/bin/claude` | ✅ Installed |
| codex | `~/.local/bin/codex` | ✅ Installed |
| grok | `~/.grok/bin/grok` | ✅ Installed |
| orca | `~/.orca-relay/bin/orca` | (Orca itself) |

The rest — **cursor, devin, copilot, gemini, antigravity, droid** — have **no binary in PATH** and have never been used.

### Hook files created by Orca — all at the exact same timestamp

Every hook file was created at **`2026-07-23 18:50:59 CST` (UTC 10:50:59)**, down to the same second:

| Agent | Hook file | Type | Notes |
|-------|-----------|------|-------|
| **copilot** ❌ (not installed) | `~/.copilot/hooks/orca.json` | New file + subdir | Created `hooks/` directory and wrote JSON |
| **gemini** ❌ (not installed) | `~/.gemini/config/hooks.json` | New file + subdir | Created `config/` directory and wrote JSON |
| **devin** ❌ (not installed) | `~/.config/devin/config.json` | **Full file overwrite** | The entire `config.json` is Orca hooks — no real devin config remains |
| **factory/droid** ❌ (not installed) | `~/.factory/settings.json` | **Full file overwrite** | The entire `settings.json` is Orca hooks — no real factory config remains |
| **grok** ✅ (installed) | `~/.grok/hooks/orca-status.json` | New file | Valid — Grok is installed (user later disabled it manually) |
| **codex** ✅ (installed) | `~/.codex/hooks.json` | New file | Valid — Codex is installed |
| **claude** ✅ (installed) | `~/.claude/` | **No Orca hook found** | Claude was NOT hooked by Orca |

### Creator

- **User**: the local user (uid 1000)
- **Process**: Orca relay daemon — Node.js process at `~/.config/nvm/versions/node/v22.23.1/bin/node relay.js`
- **Relay version**: `0.1.0+2014abd749ce` (the relay.log for this version has its last modified time at exactly the same moment: `2026-07-23 18:50`)
- **Trigger**: Orca desktop app connected to the relay socket at `2026-07-23T10:50:18Z`, triggering a global hook installation run that touched every supported agent

### Extra concerning finding

For **Devin** (`~/.config/devin/config.json`) and **Factory** (`~/.factory/settings.json`), Orca didn't just append a hook section — it **replaced the entire file** with hook content. If these files contained legitimate agent configuration, that data was overwritten and lost. This goes beyond littering into data destruction.

### Timeline on my machine

The hook installation happened as a batch at `2026-07-23 18:50:59` and has never re-run since (no later timestamps on any hook file). But the problem is that deleting any of these directories doesn't help — Orca will recreate them all on the next hook installation cycle, which appears to be triggered by an Orca desktop app restart or relay reconnection.

## Related issues

- **#9963** — "Undisclosed cross-agent hook installation" (OPEN). Another user reporting the same cross-agent hook installation behavior on Windows. The maintainers acknowledged multiple `hook-service` installers write hooks into agent configs and flagged it for security/product review.
- **#7814** — "[Bug]: Pi extensions are installed even when Pi agent is disabled" (OPEN). The maintainers confirmed the bug with a full reproduction and root cause: `buildPtyHostEnv` never reads the `disabledTuiAgents` setting, so Pi extensions are installed unconditionally whenever Agent Status Hooks is on. A fix direction was identified but not yet implemented.

## Environment

- **OS**: Linux
- **Orca version**: latest (happens on recent builds)
- **Agents NOT installed locally**: Cursor, Devin, GitHub Copilot, Gemini, Antigravity, Factory/Droid, and others

bbingz:
## Summary

- Fail closed in `installRemoteManagedAgentHooks` when the detected-agent allowlist is omitted or empty (previously `undefined` meant install every agent).
- Short-circuit `installManagedHooks` before host/home probes when no agents are allowlisted.
- Regression coverage for issue #11641; intentional full-install tests now pass an explicit allowlist.

This hardens the remaining hole after #11442 so a relay/SSH/WSL install path cannot recreate config homes for uninstalled agents.

Closes #11641

## Test plan

- [x] `vitest run` focused suites:
  - `remote-hook-service-installers.test.ts`
  - `managed-hook-local-filesystem.test.ts`
  - `managed-agent-hook-controls.test.ts`
  - `managed-hook-installer.test.ts` (relay)
- [x] Independent Grok review: APPROVED
- [ ] CI green


brennanb2025:
Taking a look!

brennanb2025:
**Verification — land-as-is (0 blockers)**, reviewed at `dc758247e0` for #11641.

Independent code review (two reviewers) + a perf audit + a live SSH-relay reproduction rig all agree the change is correct and strictly narrowing.

**Live rig** — disposable Linux SSH host, isolated `HOME`, controlled login-shell `PATH`. Drove the exact desktop→relay RPCs (`preflight.detectAgents` → `agent_hook.installManagedHooks`) against relay bundles built from **both `main` and this head**, each byte-verified against its source:

| case | `main` | this PR |
|---|---|---|
| empty allowlist (`agents: []`) | `installers:0`, but spawns a login-shell `GROK_HOME` probe + creates the `~/.orca` lock dir | `installers:0` — **no probe, no `~/.orca`** |
| omitted allowlist (old-desktop skew) | `installers:0` (+ probe / `~/.orca`) | `installers:0` |
| present (`agents:['claude']`) | writes `.claude/settings.json` | **byte-identical** file |
| reinstall on reconnect | idempotent | idempotent |
| malformed agent id | rejected (`invalid_managed_hook_agents`) | rejected |

No config write differs on any path — the change only ever *narrows* installation. On the empty/omitted path it additionally drops a login-shell subprocess (8s timeout ceiling) and a lock `mkdir` per session/reconnect. Windows remotes bail before any RPC; WSL reaches the same guarded leaf.

**Gates at `dc758247e0`:** the new fail-closed test is non-vacuous (revert the leaf → red; restore → green); 43/43 across the four touched agent-hooks suites; `typecheck` clean; `oxlint` incl. the type-aware code-quality gate clean.

**One non-blocking nit** (both reviewers): the runtime-level early return in `managed-hook-runtime.ts` that skips the probe/lock isn't pinned by a unit test — its side-effect elimination is rig-verified, and the *config-mutation* property is covered by the leaf test, so reverting only that hunk stays green. Optional coverage hardening; tracked for a follow-up alongside the pre-existing remote "unknown vs. empty" observability gap. Nothing requested here.

This is a verification note, not an approval or merge — the only remaining item is the maintainer-gated CI workflow run.

bbingz:
@brennanb2025 following up on your note that the only remaining item is the maintainer-gated CI workflow: could you hit "Approve and run workflows" on the PR Checks run for dc758247e0? The head is unchanged since your verification — deliberately so.

brennanb2025:
**Electron + live-SSH validation — closes the one gap left open in my earlier verification.** Still land-as-is at `dc758247e0`, 0 blockers.

My previous comment noted the Electron app-level check was unverified (the isolated dev profile had wedged during `ssh.connect`). That is now done end-to-end through the real app, and it passes.

**Setup:** Orca dev build launched from the PR head (identity verified over CDP: `pr-11676-review @ dc758247e0`), attached to a throwaway Linux SSH container as a real remote host, with a real git repo and a real Orca worktree created on it. Relay actually deployed and running on the remote — bundle `0.1.0+b8b28000b304`, i.e. this head.

**Negative case — no agent CLI installed on the remote** (this is #11641). Run in Orca's own remote terminal on the SSH host:

```
=== PR 11676 / issue 11641: NO agent CLI installed on this host ===
detected agents: NONE

--- agent config dirs created in HOME by Orca connect: ---
(none) -- FAIL-CLOSED OK: 0 of 14 agent homes written

.bashrc  .cache  .npm  .orca  .orca-relay  .orca-remote  .profile  .ssh
```

Zero of the 14 agent config homes were created. `.orca` there contains only `sessions/` from the terminal I opened — there is no `.orca/agent-hooks` and no install-lock, confirming the installer never ran rather than running and writing nothing.

**Positive case — exactly one agent detected.** With only `claude` on the login-shell PATH, reconnecting installed precisely that one and nothing else:

```
--- PR 11676: agent config dirs in remote HOME ---
PRESENT: .claude
--- only detected agent should appear above ---
/opt/agent-stubs/claude
```

`~/.claude/settings.json` got the real `UserPromptSubmit`/`Stop` hooks wired to `~/.orca/agent-hooks/claude-hook.sh`, and the other 13 agents were untouched. So the change narrows installs without weakening the working path.

**Revert-proofing:** reverting only the leaf guard turns the new test red with `expected [ { agent: 'claude', …(4) }, …(13) ] to deeply equal []` — it reproduces the exact 14-agent fan-out from the issue, so the test genuinely pins the fix.

**Gates at this head:** 41/41 tests in the touched files, `typecheck` exit 0, oxlint and oxfmt clean. CI is now fully green — 42 successful checks, `e2e` correctly skipped, `mergeStateStatus: CLEAN`.

I also re-traced the call sites to check the one regression that would matter — hooks that used to install and now silently don't. All three production callers (SSH, WSL, and the relay handler) already detect agents and pass an explicit allowlist, returning early on zero, so no live path changes behavior. The two side effects the runtime early-return now skips (the install lock and the Grok-home resolution) each have exactly one caller, this function, so nothing else depends on them.

Thanks for the clean, minimal fix and for holding the head steady — that made verification straightforward.

bbingz:
@brennanb2025 thank you for closing the Electron + live-SSH gap — the negative (0/14 homes) and single-agent positive cases are exactly the #11641 contract.

Holding the head at `dc758247e0` with no further pushes so the green rollup stays intact. Land-as-is from my side whenever you or another maintainer is ready to merge.

brennanb2025:
Thanks for this — it's a real hole, and the fix is the standard shape for it (coerce the missing allowlist to an empty set and filter), so I kept your approach as-is.

Heads up that I pushed one commit to your branch: `test(agent-hooks): pin installManagedHooks fail-closed gate`.

Your new regression test already pins the leaf guard, but the early return you added to `installManagedHooks` was invisible to the suite — deleting it left the agent-hooks and relay tests green, because the leaf guard makes the returned summary identical and the only difference is side effects. The added test runs `installManagedHooks` against a throwaway `HOME` with a recording login shell and asserts the directory stays empty, so a revert now fails with `expected [ '.orca', 'login-shell', 'login-shell-runs' ] to deeply equal [ 'login-shell' ]` — which names both of the side effects your guard skips.

For what it's worth, that skip is also a decent win on its own: the empty-allowlist path went from ~290 ms and a `~/.orca` mkdir per call to ~0.1 ms with nothing written.

One thing we're tracking separately, not a blocker here — when remote agent detection is unavailable, the remote path returns `{installers: 0, errors: 0}` with no per-agent status, so an agent the user really does have installed silently gets no hooks. The local path already reports `skipped` with a `skipReason`; we'd like the remote path to match.

bbingz:
@brennanb2025 thanks for keeping the approach and for the `installManagedHooks` side-effect pin — that early return was the part that actually mattered and would have been easy to lose.

Noted on the remote `{installers:0,errors:0}` / no-per-agent-status gap when detection is unavailable; happy to take a follow-up if you want the remote path to mirror local `skipped` + `skipReason`.

brennanb2025:
Merged latest `main` in to clear the conflict — no rebase, no force-push, so your commits are untouched.

The only collision was `managed-hook-runtime.test.ts`: `main` had converted the `resolveRelayGrokHome` cases to a mocked `node:child_process` `execFile` (a CI flakiness fix), while this PR added its fail-closed cases to the same file. I kept **both** — main's `execFile` mock infra and its converted probe assertions, plus all of your `installManagedHooks` tests.

One wrinkle worth flagging: your install tests observe the probe through the real login-shell run log, and main's mock is a bare `vi.fn()`, so the probe would never have spawned. The mocked `execFile` also loses `util.promisify.custom`, which is why main's stubs call back with a `{ stdout, stderr }` object. So the shared `beforeEach` now defaults the mock to the real `execFile` with the callback reshaped to match, and the parser cases stub over it as before.

No source logic changed — `managed-hook-runtime.ts` and `remote-managed-hook-installers.ts` are byte-identical to your last commit. Full `src/main/agent-hooks` suite is green (39 files, 604 passed / 4 skipped), and typecheck + oxlint are clean.

Thanks for the fix! 🙏

brennanb2025:
Apologies for the slow turnaround here 🙏 CI is running now and I'll merge once it's green.

Thanks for this — failing closed on the allowlist is the right default, and it matches how comparable tools gate this. We're picking up the rest of #11641 as a follow-up: making writes into agent config files we don't own fully non-destructive, and not creating config homes for agents that aren't installed. This PR is the right foundation for that.

Appreciate your patience!

PR #11677, opened 2026-07-31
fix(mobile): clear workspace search when closed

brennanb2025:
## Summary

- clear the workspace query whenever search is closed
- keep the embedded and phone search toggles on the same state transition
- add a regression covering query entry, close, and full-list restoration

## Validation

- `mobile/node_modules/.bin/vitest run src/worktree/host-workspace-search.test.ts src/worktree/workspace-list-sections.test.ts` (31 passed)
- `mobile/node_modules/.bin/tsc --noEmit`
- `mobile/node_modules/.bin/oxlint --deny-warnings`
- `node config/scripts/check-max-lines-ratchet.mjs`
- `git diff --check`

## QA evidence

The original uncropped broad-QA screenshots will be attached separately; no emulator action was taken in this fix worktree.

### DIFF
--- a/mobile/app/h/[hostId]/index.tsx
+++ b/mobile/app/h/[hostId]/index.tsx
@@ -92,6 +92,7 @@ import {
   type Worktree
 } from '../../../src/worktree/workspace-list-sections'
 import { useWorkspaceSections } from '../../../src/worktree/use-workspace-sections'
+import { useWorkspaceSearch } from '../../../src/worktree/use-workspace-search'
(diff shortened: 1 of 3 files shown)


brennanb2025:
Broad-QA repro evidence (uncropped, iPad):

Full workspace list before searching:

Full workspace list before searching

Buggy state after entering `demo` and closing the search field—the search icon is restored, but the list remains filtered:

Hidden active workspace filter after closing search

The focused component regression now asserts this flow restores both workspace rows after **Close search**.

PR #9364, opened 2026-07-18
fix(workspaces): make remote workspace deletion idempotent

mmarabel:
## Summary

- make remote workspace deletion idempotent when the remote runtime reports an authoritative `selector_not_found` or `repo_not_found`: forget the stale mirrored row instead of leaving an undeletable **Unknown** workspace
- treat a successful backend workspace removal as authoritative for owner-runtime PTY teardown, so renderer cleanup does not issue a second `terminal.stop` against metadata the backend just removed
- keep renderer terminal bindings cleanup intact while leaving manual sleep and exact hibernation teardown fail-closed
- allow metadata-only `worktrees:forgetLocal` cleanup after the parent project disappears, without contacting the missing remote runtime
- prune the deleted worktree from the correct host-scoped workspace-session partition so it does not return after reload/restart
- make runtime exact-ID orphan cleanup stop surviving sessions and clear metadata, history, browser/mobile state, and affected caches
- centralize exact runtime RPC error-code classification, including wrapped and legacy envelopes, without matching incidental diagnostic text

Related to #9024 and #8613. This complements #9025 and #9200, which prevent or sweep orphan state during project removal/load, plus the defensive persistence and hydration fixes in #9342, #9343, and #9344.

## Live remote reproduction

Validated against a paired Linux remote runtime on Orca `1.4.162-rc.0` using an Orca-created worktree named `test-56`.

The server-side evidence showed that deletion succeeded:

1. persisted metadata before deletion contained `orcaCreationSource: runtime`, its instance ID, owner, and two terminal tabs
2. both PTYs exited successfully
3. `git worktree remove` succeeded
4. branch deletion succeeded
5. the path, Git worktree registration, branch, `WorktreeMeta`, and workspace-session state were all gone

The client still reported `selector_not_found`. The race was the renderer calling `shutdownWorktreeTerminals()` after `worktree.rm` returned success. That cleanup issued a second remote `terminal.stop` using the now-deleted worktree selector. Depending on whether the runtime graph had processed removal yet, it either succeeded against stale graph state or returned `selector_not_found`, causing the outer removal flow to report a false failure and retain the client row.

This branch makes `remove-worktree` teardown renderer-only after backend success. Physical PTY shutdown remains owned and verified by the backend removal path.

The original stale-mirror reproduction also remains covered:

1. confirmed a deleted remote workspace reappeared under **Unknown** from `workspaceSessionsByHostId`
2. reproduced the pre-fix `selector_not_found` delete toast
[...]

Issue #8613, opened 2026-07-13
Remote Orca Server removal can orphan descendants and discard retry state


ruigomeseu:
## Summary

Removing a running worktree through a paired Remote Orca Server can partially tear down the terminal, fail filesystem deletion with `Directory not empty`, and still remove Orca's workspace metadata. Child processes, generated files, and external routes can remain, but the worktree can no longer be targeted for a retry because Orca returns `selector_not_found`.

This is a Remote Orca Server project whose repository is local to the Linux server, not an SSH-backed repository. It appears distinct from #2656.

## Environment

- macOS Orca client 1.4.137
- Paired headless Orca runtime on Linux
- Long-running Vite development server launched through Turbo and Portless
- Portless exposing the application through Tailscale Serve

## Steps to reproduce

1. Pair the desktop client with a Linux Remote Orca Server.
2. Register a repository that is local to that server.
3. Create an Orca worktree and start a long-running development command in an Orca terminal. The command should include a child that writes cache files inside the worktree while running; Vite writing `.vite` reproduced this reliably.
4. Confirm the application is serving.
5. Remove the worktree normally while the development terminal is still running. The CLI equivalent used for the reproduction was:

   ```sh
   orca worktree rm --environment <remote> --worktree <selector> --force --json
   ```

## Actual behavior

The removal returned:

```text
Failed to force delete worktree at /path/to/worktree. error: failed to delete '/path/to/worktree': Directory not empty
```

Afterward:

- The Portless parent process was gone, but Turbo, Vite, and esbuild descendants were still running.
- Vite recreated a `.vite` directory after deletion had begun, leaving the worktree directory behind.
- The Tailscale Serve route remained registered even though its listener was gone.
- The worktree branch remained.
- Orca had already removed its workspace metadata. A second removal attempt returned `selector_not_found`, so cleanup could not be retried through Orca.

## Expected behavior

- Orca should stop the full process tree and wait for descendants to exit before deleting the filesystem path.
- If filesystem or Git removal fails, Orca should preserve enough workspace metadata to retry or surface a cleanup action.
- A failed removal should not leave background processes or external routes behind.

## Workaround

A committed `orca.yaml` archive hook that sends an interrupt to the development terminal and waits for its shell to regain control prevents the race. With the hook explicitly enabled (`--run-hooks` in the CLI), the same remote test removed all processes, the listener, Tailscale Serve configuration, directory, and branch successfully.

PR #11684, opened 2026-07-31
fix(workspaces): make remote workspace deletion idempotent

AmethystLiang:
## Summary

- Make remote workspace deletion idempotent when the remote runtime reports an authoritative `selector_not_found` or `repo_not_found`: forget the stale mirrored row instead of leaving an undeletable **Unknown** workspace
- Treat a successful backend workspace removal as authoritative for owner-runtime PTY teardown, so renderer cleanup does not issue a second `terminal.stop` against metadata the backend just removed
- Keep renderer terminal-binding cleanup intact while leaving manual sleep and exact hibernation teardown fail-closed
- Allow metadata-only `worktrees:forgetLocal` cleanup after the parent project disappears, without contacting the missing remote runtime
- Prune the deleted worktree from the correct host-scoped workspace-session partition so it does not return after reload/restart
- Trust persisted owner `hostId` for orphan cleanup (not a caller’s potentially stale host), sweep the owning host’s PTY provider, and invalidate filesystem auth on removal
- Scope session/topology cleanup to the owning partition so sibling worktrees are not rebased and same-id workspaces on other hosts are not swept
- Centralize exact runtime RPC error-code classification (including wrapped/legacy envelopes and Electron IPC re-wraps), without matching incidental diagnostic text

Lands and extends [#9364](https://github.com/stablyai/orca/pull/9364). Related to #9024 and #8613; complements #9025 and #9200.

## Problem

Remote worktree delete could succeed on the server (Git worktree gone, PTYs stopped, branch deleted) and still surface a false client failure, or leave ghost **Unknown** rows after reload.

Main failure modes:

1. **Double teardown race** — After `worktree.rm` succeeded, the renderer called `shutdownWorktreeTerminals()`, which issued a second remote `terminal.stop` for the already-removed selector. The runtime returned `selector_not_found`, and the outer flow treated that as delete failure and kept the client row.
2. **Stale mirrored rows** — Deleted remote workspaces reappeared under **Unknown** from host-partitioned `workspaceSessionsByHostId` because local metadata/session cleanup did not fully forget the mirror.
3. **Wrong host / partition cleanup** — Removal sometimes trusted a stale caller `hostId`, cleaned only the local partition, left remote session state/topology fences, failed to stop orphaned remote PTYs, and could rebase sibling worktrees.

## Approach

- Backend removal remains the authority for physical PTY, Git, and filesystem teardown
- After backend success, renderer teardown is binding-only (no second destructive remote stop)
- Authoritative missing-selector / missing-repo responses trigger metadata-only `forgetLocal` recovery instead of a stuck failure toast
- Session purge, topology fencing, PTY sweeps, and filesystem-auth invalidation use the persisted owning host/connection
- Only the owning partition may fence on emptiness; spill partitions that never held the worktree do not claim repo authority
- Error classification uses exact codes/message boundaries so unrelated diagnostics never trigger forget-local

## Screenshots

No visual change.

## Testing

- Focused worktree removal, terminal teardown, RPC error classification, persistence, and runtime orphan-cleanup suites updated/added
- Coverage for: duplicate remote teardown after delete, host-scoped session partition fencing, forget-local on authoritative missing selector/repo, ownerless remote teardown host ID, filesystem auth invalidation

## Notes

- Intentionally does **not** forget a workspace when its runtime is merely unavailable — only authoritative missing-project/missing-selector responses (or explicit metadata-only actions) take that path
- Partial server-side filesystem orphans (e.g. #8613) still require server-side cleanup; this PR does not claim an unknown path is safe to delete
- Local, direct SSH, nested HUB/SSH, folder workspace, and Windows removal paths keep existing routing; backend removal remains authority per workspace kind

## ELI5

Remote delete could finish successfully, then Orca asked the server to stop the same workspace again. The server said “that workspace is gone,” and the client treated that as failure. Orca now lets the backend own deletion once, cleans up the client copy afterward, forgets real ghost rows safely, and scopes all of that cleanup to the correct remote host so siblings and other hosts are left alone.

### DIFF
--- a/src/main/ipc/worktrees-windows.test.ts
+++ b/src/main/ipc/worktrees-windows.test.ts
@@ -427,7 +427,10 @@ describe('registerWorktreeHandlers – Windows path handling', () => {
         knownRemovedWorktree: registeredWorktree
       })
     )
-    expect(store.removeWorktreeMeta).toHaveBeenCalledWith('repo-1::C:/workspaces/improve-dashboard')
+    expect(store.removeWorktreeMeta).toHaveBeenCalledWith(
+      'repo-1::C:/workspaces/improve-dashboard',
+      'local'
+    )
     // Windows history lives under a path-derived hash, so scheduling must not regress on this path only.
     expect(deleteWorktreeHistoryDirMock).toHaveBeenCalledWith(
       'repo-1::C:/workspaces/improve-dashboard'
--- a/src/main/ipc/worktrees.test.ts
+++ b/src/main/ipc/worktrees.test.ts
@@ -10329,7 +10329,8 @@ describe('registerWorktreeHandlers', () => {
         includeLocalRegistry: false
       })
       expect(runtimeStub.clearOptimisticReconcileToken).toHaveBeenCalledWith(worktreeId)
-      expect(store.removeWorktreeMeta).toHaveBeenCalledWith(worktreeId)
+      // The purge must be scoped to the same owner the sweep used, or the ssh:* partition keeps this worktree's session state.
+      expect(store.removeWorktreeMeta).toHaveBeenCalledWith(worktreeId, 'ssh:ssh-dead')
       expect(advertisedUrlWatcherForgetWorktreeMock).toHaveBeenCalledWith(worktreeId)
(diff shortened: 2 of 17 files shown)


AmethystLiang:
@coderabbitai[bot] fixed

PR #11696, opened 2026-07-31
perf: coalesce SSH git status reads

nwparker:
| | Files | Added | Deleted | Net |
| :--- | ---: | ---: | ---: | ---: |
| Test | 3 | +317 | −22 | +295 |
| Prod | 1 | +89 | −90 | −1 |

## Summary

Coalesce concurrent SSH Git status reads across callers with distinct cancellation signals.

Each caller gets an independent lease over one client-side `git.status` RPC. The RPC aborts only after its final live lease cancels, and SSH mutations fence status joins before and after execution. Leases are in-flight only — the entry is deleted on settle, so there is no result cache and no stale-badge failure mode.

The focused concurrency test measures 10 concurrent, separately signalled status calls collapsing to 1 SSH `git.status` RPC.

## Fork-point key fix

The original revision of this PR had a defect that merged textually clean: the `git.status` request payload carries `branchLineTotalMergeBase`, but the lease key did not hash it. Two real callers disagree on that field — `git-status-refresh.ts:121-138` (automatic poll) and `:250-267` (strict user-triggered refresh) each read the gate independently — so a refresh asking for the branch-line total could join an in-flight poll that omitted it and render a response with no `branchLineTotal`, blanking the branch-header chip. After a commit moved the merge base, a refresh with the new base could also join an in-flight read using the old one and show a pre-commit number.

The key now includes the fork point, mirroring the guard already on the local path (`src/main/git/status.ts:249-251`, whose inline comment warns about exactly this).

## Rebase notes

- Base retargeted from `nwparker/perf-coalesce-git-status` to `main`; the parent (#11691) has since merged and `GitStatusReadLeaseOwner` now lives in `src/shared/`.
- #14728 split `ssh-git-provider.test.ts` into per-area files after this PR was opened. The lease tests were carried into a new `ssh-git-provider-status-lease.test.ts`, and the `{ signal: expect.any(AbortSignal) }` assertion updates were applied to `ssh-git-provider-status.test.ts` and `ssh-git-provider-worktree.test.ts` (6 tests failed without them).
- The env-gated benchmark test (`ORCA_SSH_GIT_STATUS_COALESCING_BENCH_JSON`) was dropped rather than carried into the split file.
- #11697 is stacked on this branch and will need retargeting to `main` and a rebase once this merges.

## Mutation fencing

`runWithGitReadInvalidation` fences a strict superset of what `main` fenced as `runWithDiffDedupeClear`: all 16 of main's methods, plus `stageFile`, `unstageFile`, `discardChanges`, `bulkStageFiles`, `bulkUnstageFiles`, and `bulkDiscardChanges` — the mutations that change status output and therefore must not be joinable across a write.

## Screenshots

No visual change.

## Testing

- [x] `pnpm typecheck` (all three tsconfigs: node, cli, web)
- [x] Targeted Vitest: `src/main/providers` + `src/main/git` — 2,459 passed, 6 skipped
- [x] `pnpm check:max-lines-ratchet`, `pnpm check:reliability-gates`
- [x] Targeted `oxlint` + `oxfmt`

The new fork-point test was verified non-vacuous: it fails against the source without the key fix (`× isolates status reads by branch-line-total fork point`) and passes with it.

## Remote / cross-platform review

No new RPC method, payload field, command, path, authentication, secret, dependency, or persistence surface. The lease always passes a provider-owned shared `AbortSignal` to `mux.request`, but that only registers a local abort listener (`ssh-channel-multiplexer.ts:277`) — nothing goes on the wire and no relay upgrade is required, so mixed client/host versions are unaffected. No Git commands or options change, so the Git 2.25 floor is untouched. Lease state is scoped to one `SshGitProvider` instance and removed after settlement or invalidation.

Known residual: the key does not normalize Windows path casing, so two callers using different casing for the same worktree perform separate reads. That costs a duplicate read, never a wrong result, and matches the local path's existing behavior.

## Notes

This is the third focused extraction from draft PR #11539.

### DIFF
--- a/src/main/providers/ssh-git-provider-status-lease.test.ts
+++ b/src/main/providers/ssh-git-provider-status-lease.test.ts
@@ -1,0 +1,268 @@
+import { describe, expect, it, beforeEach } from 'vitest'
+import { SshGitProvider } from './ssh-git-provider'
+import {
+  createMockMux,
+  waitForRequestCount,
+  type MockMultiplexer
+} from './ssh-git-provider-test-harness'
+
+function deferredPromise<T>(): {
+  promise: Promise<T>
+  resolve: (value: T) => void
+  reject: (error: unknown) => void
+} {
+  let resolve!: (value: T) => void
+  let reject!: (error: unknown) => void
+  const promise = new Promise<T>((innerResolve, innerReject) => {
+    resolve = innerResolve
+    reject = innerReject
+  })
+  return { promise, resolve, reject }
+}
+
+describe('SshGitProvider status read leases', () => {
+  let mux: MockMultiplexer
+  let provider: SshGitProvider
+
+  beforeEach(() => {
+    mux = createMockMux()
+    provider = new SshGitProvider('conn-1', mux as never)
+  })
+
+  it('shares one status RPC across distinct caller signals', async () => {
+    const pending = deferredPromise<{
+      entries: never[]
+      conflictOperation: 'unknown'
+    }>()
+    mux.request.mockReturnValue(pending.promise)
+    const controllers = Array.from({ length: 10 }, () => new AbortController())
+
+    const reads = controllers.map((controller) =>
+      provider.getStatus('/home/user/repo', { signal: controller.signal })
+    )
+    await waitForRequestCount(mux.request, 1)
+
+    expect(mux.request).toHaveBeenCalledTimes(1)
+    pending.resolve({ entries: [], conflictOperation: 'unknown' })
+    await expect(Promise.all(reads)).resolves.toHaveLength(10)
+  })
+
+  it('isolates first and later caller cancellation while a status lease remains', async () => {
+    const pending = deferredPromise<{
+      entries: never[]
+      conflictOperation: 'unknown'
+    }>()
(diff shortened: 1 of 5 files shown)


nwparker:
Rebased onto current `main` and retargeted the base from `nwparker/perf-coalesce-git-status` (parent #11691, since merged) to `main`.

**Correctness fix on top of the original change:** the `git.status` request payload carries `branchLineTotalMergeBase`, but the lease key did not. A strict user-triggered refresh that asked for the branch-line total could join an in-flight automatic poll that omitted it — blanking the branch-header chip — or reuse a pre-commit fork point. The key now includes it, mirroring the guard already on the local path in `git/status.ts:249-251`.

Tests were carried across #14728's test-file split: the lease tests now live in `ssh-git-provider-status-lease.test.ts`, and the `{ signal: expect.any(AbortSignal) }` assertion updates were applied to `ssh-git-provider-status.test.ts` and `ssh-git-provider-worktree.test.ts`. The env-gated benchmark test was dropped. Verified the new fork-point test fails without the key fix and passes with it.

Note: #11697 is stacked on this branch and will need rebasing/retargeting once this merges.

PR #11705, opened 2026-07-31
fix(linux): strip AppRun --no-sandbox under ELECTRON_RUN_AS_NODE

### LINKED ISSUE #11609: [Bug]: Registered CLI (orca/orca-ide) exits 9 with "bad option: --no-sandbox" on userns-restricted Linux — AppRun injects the Chromium flag into ELECTRON_RUN_AS_NODE invocations
### Environment

- Orca **v1.4.161** and **v1.4.162** (orca-linux.AppImage, x64; v1.4.162 sha512-verified against the release manifest)
- Ubuntu 26.04, `kernel.apparmor_restrict_unprivileged_userns=1` (the default hardening on recent Ubuntu)
- CLI registered via Orca Settings → CLI registration (wrappers installed to `~/.local/bin/orca` and `~/.local/bin/orca-ide`)

### What happens

Every invocation of the registered CLI fails immediately:

```
$ orca-ide status
/home/user/.cache/.mount_orca-XXXXXX/orca-ide: bad option: --no-sandbox
$ echo $?
9
```

### Why

Chain of three individually-reasonable behaviors:

1. The registered wrapper scripts run the AppImage with `ELECTRON_RUN_AS_NODE=1` and a `-e '<cli bootstrap>' -- "$@"` payload — i.e. plain **Node mode**.
2. On hosts where unprivileged user namespaces are restricted, the AppImage's internal `AppRun` engages its no-sandbox fail-safe and **prepends `--no-sandbox`** to the argv. (Correct for Chromium mode — the GUI and `serve` work fine.)
3. In `ELECTRON_RUN_AS_NODE` mode the binary is plain Node, which rejects unknown options → `bad option: --no-sandbox`, exit 9.

So the CLI is dead on arrival on any userns-restricted Linux box, while the GUI/server on the same box work — which is presumably why CI never caught it. Boxes without the restriction are unaffected.

### Suggested fix

`AppRun` should skip injecting `--no-sandbox` (and any other Chromium-only flags) when `ELECTRON_RUN_AS_NODE` is set — Node mode never spawns Chromium, so the fail-safe has nothing to protect.

### Workaround we're running

Extract the AppImage once (`--appimage-extract`) and point the wrappers at the inner binary directly, bypassing AppRun:

```bash
APPDIR=~/.local/orca/squashfs-root ELECTRON_RUN_AS_NODE=1 \
  "$APPDIR/orca-ide" -e '<the wrapper's existing bootstrap>' -- "$@"
```

Works fully (`status`, `worktree ps/set`, `skills get`, terminals), but has to be re-applied whenever CLI registration regenerates the wrappers, and re-extracted on each app update.

### Related (different bugs, same env var)

#7768 and #2414 are about `ELECTRON_RUN_AS_NODE=1` *leaking into child processes*; this one is the inverse direction — AppRun injecting a Chromium flag *into* an intentional `ELECTRON_RUN_AS_NODE` invocation.

Kyou12138:
## Summary

Fixes #11609 — registered AppImage CLI (`orca` / `orca-ide`) exited 9 with `bad option: --no-sandbox` on userns-restricted Linux (e.g. recent Ubuntu AppArmor defaults).

### Root cause

AppRun injects Chromium `--no-sandbox` when unprivileged user namespaces are restricted. CLI wrappers correctly set `ELECTRON_RUN_AS_NODE=1`, so the binary is plain Node and rejects Chromium-only flags.

### Fix

- After packing Linux builds, rename `orca-ide` → `orca-ide.bin` and install a bash filter as `orca-ide`.
- When `ELECTRON_RUN_AS_NODE=1`, strip `--no-sandbox` / related Chromium sandbox flags before exec.
- GUI launches (no `ELECTRON_RUN_AS_NODE`) pass argv through unchanged.
- Single source of truth for flag list; atomic temp-file install; recovery for partial packs; `readlink -f` for symlink launches.

## Related Issue

Closes #11609

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature
- [ ] Refactor / documentation only
- [x] Tests (adding or improving test coverage)

## Changes Made

- `config/electron-builder.config.cjs` — wire Linux `afterPack` installer
- `config/scripts/install-linux-electron-node-flag-filter-shim.cjs` — packaging-time shim writer
- `config/scripts/install-linux-electron-node-flag-filter-shim.test.mjs` — install / recovery / idempotency tests

## How to Test

```bash
node --test config/scripts/install-linux-electron-node-flag-filter-shim.test.mjs
```

Manual (Linux AppImage host with userns restricted):

1. Build / install registered CLI
2. `orca-ide status` should no longer exit 9 with `bad option: --no-sandbox`
3. GUI launch still receives sandbox flags when not in Node mode

## Platform / SSH

- Linux packaging only; macOS/Windows packaging untouched
- SSH clients that invoke the registered host CLI benefit when the host is Linux AppImage/deb

## AI Review Report

- Cross-platform: Linux-only packaging path; guarded by `electronPlatformName === 'linux'`
- SSH/remote: host-side registered CLI; no remote path assumptions
- Performance: one rename + write at package time; runtime is a tiny bash loop
- Security: only strips known Chromium sandbox flags in Node mode; GUI keeps AppRun sandbox behavior
- UI: none

## Security Audit

- No new network surface
- No privilege escalation: filter only drops known Chromium flags in Node mode
- Does not disable sandbox for GUI sessions

## Checklist

- [x] Tests added/updated and pass locally
- [x] Scope limited to the linked issue
- [x] Risks documented above
- [x] Will address reviewer comments before merge

## Notes

X: n/a


tianhuil:
Hi -- this is breaking for me as well.  Can you please get this merged in?

Currently using a clunky work-around that I would prefer to get rid of.

PR #11716, opened 2026-07-31
fix(macos): verify Full Disk Access and re-arm notices

brennanb2025:
## Summary

- Verify effective Full Disk Access by opening the per-user TCC database read-only instead of checking whether a Safari path exists.
- Report `Granted` only after a successful open, `Denied` for permission errors, and `Unknown` for missing or unexpected probe failures.
- Version the one-shot TCC notice state so users who may have trusted the former false-positive badge are eligible again only after a fresh macOS prompt; preserve explicit permanent opt-outs.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

Validated with:

- `pnpm vitest run --config config/vitest.config.ts src/main/macos-full-disk-access-status.test.ts src/main/ipc/developer-permissions.test.ts src/main/macos-tcc-prompt-notice.test.ts src/main/macos-tcc-prompt-watch.test.ts src/renderer/src/hooks/useMacosTccPromptNotice.test.tsx src/renderer/src/hooks/macos-tcc-prompt-notice-subscription.test.ts` (56 tests)
- `pnpm run typecheck:node`
- `pnpm run check:code-quality:changed`
- `git diff --check`
- Live macOS probe with Full Disk Access disabled returned `Denied` and emitted no `AUTHREQ_PROMPTING` log event.

## AI Review Report

Reviewed the permission signal, persisted notice migration, renderer claim lifecycle, and helper-attributed TCC behavior. The former probe used `fs.access()` with its default existence-only mode; the replacement opens the protected database read-only and conservatively preserves `Unknown` for ambiguous failures. Migration coverage verifies that legacy acknowledged notices require a fresh TCC event, current acknowledgements remain silent, and legacy permanent opt-outs remain permanent.

Cross-platform behavior was checked explicitly: the probe is runtime-gated to macOS, returns `Unsupported` elsewhere, builds the home path with `path.join`, and changes no shortcuts, labels, shell commands, Git behavior, SSH routing, or workspace assumptions. The existing helper-attributed prompt watcher remains authoritative because the detached helper has an independent TCC identity.

## Security Audit

The probe opens only the known per-user TCC database path in read-only mode and closes it immediately; it never reads or exposes database contents. It accepts no renderer-controlled path, adds no command execution or dependency, and keeps the existing typed IPC status contract. Permission errors are reduced to status enums without leaking filesystem details.

## Notes

macOS does not expose a public Full Disk Access status API. This change uses effective read access as the practical signal while keeping actual TCC prompt observation as the stronger reason to show guidance.

### DIFF
--- a/src/main/ipc/developer-permissions.test.ts
+++ b/src/main/ipc/developer-permissions.test.ts
@@ -79,7 +79,7 @@ describe('registerDeveloperPermissionHandlers', () => {
     getMediaAccessStatusMock.mockReset()
     isTrustedAccessibilityClientMock.mockReset()
     getMacosFullDiskAccessStatusMock.mockReset()
-    getMacosFullDiskAccessStatusMock.mockReturnValue('denied')
+    getMacosFullDiskAccessStatusMock.mockResolvedValue('denied')
     execFileMock.mockReset()
     execFileMock.mockImplementation((...args: unknown[]) => {
       const callback = args.at(-1)
@@ -128,7 +128,7 @@ describe('registerDeveloperPermissionHandlers', () => {
   }
 
   it('returns the Full Disk Access read-probe status', async () => {
-    getMacosFullDiskAccessStatusMock.mockReturnValue('granted')
+    getMacosFullDiskAccessStatusMock.mockResolvedValue('granted')
     registerDeveloperPermissionHandlers()
 
     const call = handleMock.mock.calls.find(
--- a/src/main/ipc/developer-permissions.test.ts
+++ b/src/main/ipc/developer-permissions.test.ts
@@ -6,6 +6,7 @@ const {
   askForMediaAccessMock,
   getMediaAccessStatusMock,
   isTrustedAccessibilityClientMock,
+  getMacosFullDiskAccessStatusMock,
   execFileMock,
   createSocketMock,
   socketMock,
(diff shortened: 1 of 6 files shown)


brennanb2025:
### Electron validation (commit `12e0d371ef`)

- Launched the isolated Orca dev build from the PR worktree and verified app identity: `sponge @ brennanb2025/fix-notification-permission-badge`.
- Exercised the real Settings → macOS Permissions surface and clicked Refresh.
- Visible result: Full Disk Access remained **Denied** after refresh.
- Backing IPC result: `{"id":"full-disk-access","status":"denied"}`.
- Refresh completed in approximately 200 ms with no renderer console errors.
- Automated coverage: 56 focused tests and Node typecheck passed on the latest commit.

Full macOS Permissions settings surface showing Full Disk Access denied

The versioned one-shot notice migration was covered by focused tests; a fresh OS-level TCC prompt was not forced during this run.

brennanb2025:
### Follow-up: exact reported toast path

Reproduced the user action on commit `12e0d371ef` with Full Disk Access disabled:

- Before the click, the Full Disk Access row and backing IPC status were both `denied`.
- Clicked the row-specific **Open Settings** button through the real Settings surface.
- Orca opened the macOS Privacy & Security pane and showed **Opened macOS Privacy & Security**.
- The incorrect **Permission granted** toast did not appear.

This confirms the screenshot in the report was another visible symptom of the same false-positive probe, and this PR fixes it.

Full macOS Permissions surface showing Denied and the Opened macOS Privacy and Security toast

PR #11720, opened 2026-07-31
fix(codex): keep persistent panes logged in after home routing

brennanb2025:
## Summary

- Follow up on #11731 for the separate retained-PTY case: daemon-backed shells can survive an app restart with the legacy shared `CODEX_HOME` baked into their environment.
- Refresh that retired shared home one-way from canonical system auth and config, with durable ownership provenance and guarded generation replacement so retained Codex writes win races.
- Recover interrupted guarded auth operations before migration or sync, and fence ambiguous or concurrent credential generations instead of adopting them.
- Record real/shared/per-account home routing for persistent panes so route changes use the existing restart prompt even when the account ID is unchanged.
- Track only re-derivable process-wide custom-home overrides; pane-local, SSH, WSL, and folder-workspace contexts remain isolated.

## Screenshots

No visual layout change. The existing restart prompt uses route-specific copy when the account is unchanged.

## Testing

- [x] Full Node, CLI, and web typecheck
- [x] Native Oxlint
- [x] Type-aware Oxlint
- [x] Oxfmt check on all changed files
- [x] Max-lines ratchet
- [x] `git diff --check`
- [x] 713 focused Vitest tests across 17 files
- [x] CI shard 7/16: 250 files, 2,356 passed, 2 skipped
- [x] Added regression coverage for process-wide and pane-local overrides, managed-transition provenance, interrupted-operation recovery, and concurrent auth/config replacement

## AI Review Report

An independent agent audited the final revision across auth ownership transitions, crash recovery, conditional replacement, custom-home provenance, restart prompting, Windows retries and ACLs, WSL isolation, SSH, and folder workspaces. Its final verdict was SHIP with no concrete correctness or regression findings.

## Security Audit

- Credential contents are copied only from canonical selected storage into Orca-managed storage, written with owner-only permissions, and never logged.
- Ambiguous, malformed, interrupted, or concurrently replaced auth generations fail closed.
- Compatibility config sync is one-way, so retained pane settings cannot overwrite canonical config.
- Guarded writes publish atomically on supported local filesystems and fail before moving the canonical target when hard-link publication is unsupported.
- No new command execution, dependency, network request, secret exposure, or renderer authority is introduced.

## Landing order

Land after #11731. That PR handles transient managed-auth reads during startup; this PR handles persistent shells retaining the legacy shared home.


brennanb2025:
Electron validation completed on exact pushed SHA 4f6d50a08c2b568003375b2e53dfd1f9179c8f0c from the PR checkout.\n\n- Initial launch used the legacy shared-home route. The real Orca window showed the authenticated OpenAI Codex v0.146.0 composer in the PR workspace.\n- After a graceful app quit and relaunch of the same isolated profile with real-home routing enabled, Orca reattached the same retained local daemon PTY (…@@3a5df3bf). The foreground process remained codex and the authenticated composer stayed usable; I chose “Keep current session” on the expected setup-change notice to prove the retained session itself survived.\n- The app identity reported /Users/brennanbenson/orca/workspaces/orca/pr-11720-persistent-codex-review and branch brennanb2025/pr-11720-persistent-codex-review.\n\nBefore relaunch:\nAuthenticated Codex before relaunch\n\nAfter same-PTY reattach:\nAuthenticated Codex after reattach\n\nValidation: exact Node 24 focused suite passed 14 files / 715 tests; full typecheck, full lint, changed-file Oxfmt, and git diff checks passed. Fresh inherited-model review of the exact pushed SHA returned CLEAN with no mutation.

PR #11724, opened 2026-07-31
feat(editor): richer rich-markdown table keyboard (Tab, Enter, Backspace)

nwparker:
## Summary
Richer keyboard support for TipTap rich-markdown tables:

| Key | Behavior |
|-----|----------|
| **Tab / Shift-Tab** | Next / previous cell; Tab past last cell inserts a row |
| **Enter** | Move to cell **below**; last row adds a row |
| **Backspace** (fully empty row) | Delete row (or whole table if last row) |
| **Backspace** (empty cell, row still has content) | Step to previous cell |

Also fixed Orca's Tab handler, which previously always sank/lifted lists and swallowed table cell Tab.

### Not in this PR
- Row/column handle chrome
- Multi-cell selection delete chains
- Arrow-key cell edge navigation
- Column delete, merge/split UI, pipe-input table insert rule

## Test plan
- [x] Unit: Tab, Enter, empty-row Backspace, empty-cell → previous cell
- [x] Unit: existing list Tab handler still green outside tables
- [x] Electron e2e: Tab / Shift-Tab / Enter / empty-row Backspace
- [x] Visual proof WebM via `gh image`

## Visual proof

Latest Electron run (Tab + Enter + empty-row Backspace):

https://github.com/user-attachments/assets/b3010d0f-9e46-419f-8373-e07b2725b2d3

Earlier runs:

https://github.com/user-attachments/assets/e05fee54-b23a-4e95-b531-8652ac1e9fc6

https://github.com/user-attachments/assets/3f7fb5be-8c21-49a1-a089-084dbe4d2a7a

### Screenshots (empty-row Backspace)

before

after


nwparker:
### Full-window Electron evidence

**Before Backspace** (empty row between keep/stay):


**After Backspace** (empty row gone):

nwparker:
## Visual proof (follow-up: Tab/Shift-Tab + empty-row Backspace)

https://github.com/user-attachments/assets/e05fee54-b23a-4e95-b531-8652ac1e9fc6

Second pass:

https://github.com/user-attachments/assets/3f7fb5be-8c21-49a1-a089-084dbe4d2a7a

nwparker:
## Visual proof (OSS parity pass: Enter + empty-cell Backspace + Tab)

https://github.com/user-attachments/assets/b3010d0f-9e46-419f-8373-e07b2725b2d3

Patterns from BlockNote `TableExtension` (Enter→cell below) and Outline table keys (Tab / structural Backspace), plus empty-row delete.

nwparker:
## Review comments addressed

All 7 inline comments are resolved in `15416df` + `83f6d25`, along with a simplification pass (−405 lines net).

| Comment | Resolution |
|---|---|
| greptile `row-delete.ts:7` / coderabbit `row-delete.ts:8` — `textContent` treats media-only cells as empty | Fixed. `isEmptyCell` now requires every child to be an empty textblock, so an image-only cell blocks the row delete. Matches coderabbit's suggested diff. Covered by a mutation-verified test. |
| greptile `row-delete.ts:17` — **P1**, emptied header row | Fixed, but the diagnosis was off — see below. |
| greptile `e2e:80`, coderabbit `e2e:89`, coderabbit `e2e:99` — synthetic DOM caret placement | Fixed. `placeCaretInCellWithText` is now a plain Playwright click; `placeCaretInEmptyBodyRow` is deleted outright, since the preceding `Enter` assertion already leaves the caret in that row. |
| coderabbit `row-delete.test.ts:222` — unbounded `deleteRow` loop | Moot; that file is gone. The three near-identical table test files were merged into one `rich-markdown-table-keyboard.test.ts`, and empty rows now come from markdown fixtures (`\|  \|  \|`) instead of manual doc surgery. |

### On the P1 header-row comment

The predicted failure mode doesn't occur. The serializer does **not** promote the first body row to a header — it synthesizes an empty one. I probed it directly:

```
BEFORE:  |  |  | / --- / | keep | a | / | stay | c |     (row 0 = tableHeader)
AFTER:   |  |  | / --- / | keep | a | / | stay | c |     (row 0 = tableCell)
```

The markdown is byte-identical, so there is no data-semantics change and nothing is lost. But there *is* a real bug underneath: the row disappears from the editor while the saved file never changes, so it reappears on the next reload. Fixed by leaving an emptied header row in place (still consuming the key so ProseMirror can't merge the table into the preceding node). A header-only table still exits cleanly via `deleteTable`.

### Also fixed — not flagged by either bot

Table `Enter` ran *before* the slash-menu and doc-link-menu blocks in `rich-markdown-key-handler.ts`, so typing `/` in a table cell and pressing Enter moved the caret down instead of running the slash command. Same for `[[` doc links. The Tab path already guarded against this; Enter didn't. Regression test added.

### Verification

996 editor unit tests pass · `tsc` clean · oxlint + oxfmt clean · all three new regression tests mutation-checked (each fails against the old behavior).

PR #11731, opened 2026-07-31
fix(codex): prevent transient managed-auth onboarding

nwparker:
## Summary

Prevent a fresh Orca-launched Codex session from latching unauthenticated state while a selected managed account's `auth.json` is temporarily missing or incomplete.

This revision keeps the fix at the Codex launch boundary and avoids turning a credential problem into a terminal-wide outage:

- fresh host-local Codex launches wait up to 1.5 seconds for a known incomplete credential to recover;
- a credential that is valid on the first read stays on a synchronous, timer-free path;
- if auth remains unavailable, Orca clears only the still-selected host account and falls back through the existing system-default launch path;
- selection changes during the wait are re-resolved and fenced, so a recovered old account cannot launch after the user selects another account;
- resumed/pinned Codex sessions never switch identities; unavailable origin auth rejects the resume;
- bare shells, other agents, SSH, WSL, daemon reattaches, system-default homes, and unmanaged homes bypass this gate;
- hidden Codex quota probes still stop immediately if Codex renders `Sign in with ChatGPT`.

## User symptom and evidence

With an authenticated managed account selected, a newly opened Codex tab could intermittently show onboarding, while an immediate retry opened the authenticated composer.

A deterministic real-Codex reproduction established the latch behavior: start Codex while a valid copied `auth.json` is temporarily absent, restore the file before reading the account, and that process remains unauthenticated while a new process against the restored home is authenticated. This proves the startup-latch failure mode. The exact production writer and timing for each historical occurrence were not captured, so this PR does not claim more specific causality.

The Orca regression surface dates to #9501 (`e58de71f5e`, 2026-07-20), which made managed account homes live self-contained `CODEX_HOME` directories and treated a missing `auth.json` as a reason to abandon the selected account at launch. #9451 is similar by symptom but is not treated as a confirmed duplicate.

### Credential readiness

Known complete Codex credential forms are accepted:

- `chatgpt` / `chatgptAuthTokens` token records;
- `apikey`;
- `personalAccessToken`;
- `bedrockApiKey`;
- object-backed `agentIdentity` records with the required runtime ID and private key.

Known partial records retry. Genuinely unknown future modes or parseable object shapes fail open so a new Codex schema cannot strand terminal creation. Missing, malformed, or unreadable files use the bounded retry/fallback path.

### Identity and availability safety

The readiness decision is scoped to fresh Codex launches. After the wait, Orca re-checks the selected account before it creates a PTY. A permanent credential failure restores the previous safe system-default behavior instead of rejecting every terminal indefinitely. Resume-pinned sessions remain pinned and reject if their original managed credential is unavailable.

The same rules are enforced for renderer-to-daemon, renderer local fallback, and runtime-controller Codex launches. Trusted managed-home ownership checks remain in place.

## No-regression and performance evidence

| Lane/state | Result |
| --- | --- |
| Valid managed auth on first read | Synchronous return; no Promise tick or polling timer |
| Transient missing/partial auth | Bounded wait, then spawn with the still-selected managed home |
| Permanently unavailable auth | Clear only the still-selected host account and use normal fallback |
| Selection changes during wait | Re-resolve/fence; never launch the stale account |
| Resume-pinned Codex session | Keep original account; reject rather than switch identity |
| Bare shell or another agent | No Codex readiness read or wait |
| Codex daemon reattach | No current-account gate |
| SSH / WSL | Existing remote and distro-local behavior unchanged |
| Folder workspace / Git worktree | Shared PTY boundary; no worktree-only assumption |
| Hidden quota PTY shows onboarding | Stop immediately; never send `/status` |

The healthy path is covered by an explicit synchronous-return test, and the PTY suite verifies that non-Codex launches and reattaches do not gain the readiness wait.

## Verification

Final candidate `1017c2589630e4aa2d55c6f727df0d0a2572585e`:

- focused Vitest suite: **6 files, 549 tests passed**;
- `pnpm typecheck`: node, CLI, and web passed;
- `pnpm lint`: passed, including 56 reliability gates, max-lines ratchet, bundled-skill checks, and localization checks;
- `pnpm exec oxfmt --check` on changed files: passed;
- `git diff --check`: passed;
- pre-commit oxlint, react-doctor, and formatting hooks: passed.

A same-model review loop continued until a fresh reviewer returned **CLEAN**. Earlier completed iterations found and drove fixes for real Codex credential shapes, permanent-auth terminal lockout, over-broad terminal gating, selection races, the healthy-path Promise tick, and partial Agent Identity records. The final review found no remaining correctness, availability, identity, scope, platform, or performance issue.

### Electron validation

An isolated source build was launched from this exact worktree with a disposable user-data profile. Settings showed the disposable managed Codex account active; **New tab → Codex** opened Codex 0.146.0 directly at the authenticated composer, not onboarding. Process inspection confirmed that the launched Codex process received the disposable managed account's `CODEX_HOME`; no credential values were displayed or captured.

Fresh Codex launch opens the authenticated composer

## Deliberate boundary

This guards Orca's fresh Codex launch action. If a user opens a bare shell and manually types `codex` much later, Orca cannot intercept that future shell command at PTY-spawn time; that path remains outside this launch-scoped fix.

### DIFF
--- a/src/main/codex-accounts/managed-codex-auth-readiness.test.ts
+++ b/src/main/codex-accounts/managed-codex-auth-readiness.test.ts
@@ -1,0 +1,144 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import type { CodexManagedAccount, GlobalSettings } from '../../shared/types'
+import { waitForManagedCodexAuthReady } from './managed-codex-auth-readiness'
+
+const roots: string[] = []
+const testIdToken = 'e30.eyJlbWFpbCI6InVzZXJAZXhhbXBsZS5jb20ifQ.sig'
+const testChatGptAuth = {
+  tokens: {
+    access_token: 'access',
+    id_token: testIdToken,
+    refresh_token: 'refresh',
+    account_id: 'account'
+  },
+  last_refresh: '2026-07-31T00:00:00Z'
+}
+
+afterEach(() => {
+  vi.useRealTimers()
+  for (const root of roots.splice(0)) {
+    rmSync(root, { recursive: true, force: true })
+  }
+})
+
+describe('waitForManagedCodexAuthReady', () => {
+  it('accepts a readable managed ChatGPT credential', async () => {
+    const fixture = createFixture()
+    writeAuth(fixture.home, testChatGptAuth)
+
+    await waitForManagedCodexAuthReady(fixture.args)
+  })
+
+  it('waits for a missing managed credential to be restored', async () => {
+    vi.useFakeTimers()
+    const fixture = createFixture()
+    const readiness = waitForManagedCodexAuthReady(fixture.args)
+
+    await vi.advanceTimersByTimeAsync(50)
+    writeAuth(fixture.home, { OPENAI_API_KEY: 'sk-test' })
+    await vi.runAllTimersAsync()
+
+    await expect(readiness).resolves.toBeUndefined()
+  })
+
+  it('waits for a partial managed ChatGPT credential to become complete', async () => {
+    vi.useFakeTimers()
+    const fixture = createFixture()
+    writeAuth(fixture.home, {
+      tokens: { access_token: 'access', id_token: testIdToken }
+    })
+    let resolved = false
+    const readiness = waitForManagedCodexAuthReady(fixture.args)?.then(() => {
+      resolved = true
+    })
+
+    await vi.advanceTimersByTimeAsync(50)
+    expect(resolved).toBe(false)
+    writeAuth(fixture.home, testChatGptAuth)
+    await vi.advanceTimersByTimeAsync(25)
+
+    await expect(readiness).resolves.toBeUndefined()
+  })
+
+  it('waits for an empty managed ChatGPT refresh token to be restored', async () => {
+    vi.useFakeTimers()
+    const fixture = createFixture()
+    writeAuth(fixture.home, {
+      ...testChatGptAuth,
+      tokens: { ...testChatGptAuth.tokens, refresh_token: '' }
+    })
+    let resolved = false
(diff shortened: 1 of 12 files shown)


brennanb2025:
Compared this with #11720. I think this is the right fix for the transient-auth failure reproduced here: a fresh Codex process can latch unauthenticated state if the selected managed auth is missing or partial during startup, and preserving the selected identity plus gating local process creation fails closed at the right boundary. The deterministic app-server reproduction is materially stronger evidence for the reported new-process/onboarding/immediate-retry symptom.

The relationship between the PRs is:

- #11731 covers a fresh process using a self-contained managed home during a transient `auth.json` rewrite.
- #11720 covers a persistent daemon-backed PTY that retains the legacy shared `CODEX_HOME` in its already-running shell across app and home-route changes.

The visible symptom overlaps, but the mechanisms are separate. I think this PR should land independently/first; we can rebase #11720 on it and revalidate the retained-PTY reproduction rather than treating #11720 as a substitute for this fix.

One scope boundary to keep in mind: the readiness gate protects new-PTY/automatic-launch paths. A manual `codex` invocation later inside an already-open shell is outside that gate, so if that exact reproduction persists it would be follow-up evidence rather than evidence against this fix.

brennanb2025:
Final review/validation update for `1017c2589630e4aa2d55c6f727df0d0a2572585e`:

- Same-model review loop continued through a fresh **CLEAN** verdict. Findings fixed along the way: real Codex credential schemas, permanent-auth terminal lockout, over-broad terminal gating, account-selection races, the valid-auth Promise tick, and partial Agent Identity records.
- Focused validation passed: **6 files / 549 tests**, full node/CLI/web typecheck, full lint and reliability gates, changed-file format check, diff check, and pre-commit hooks.
- Healthy managed auth remains synchronous and timer-free; non-Codex spawns, SSH, WSL, and reattaches bypass the readiness gate.
- Electron validation from the exact source worktree passed: **New tab → Codex** opened directly at the authenticated composer under the disposable managed `CODEX_HOME`, not onboarding.

Fresh Codex launch opens the authenticated composer

Deliberate boundary: manually typing `codex` much later in an already-open bare shell remains outside this spawn-time launch gate.

PR #11732, opened 2026-07-31
fix(browser): stop agent browser input from stealing the user's focus

songminsns:
## The problem

While an agent drives the built-in browser, its keystrokes land in whatever the user is typing into.

CDP input only reaches an Electron `<webview>` guest while that guest holds native DOM focus. Without it Chromium routes the event to whichever surface actually has focus — the user's terminal — and the CDP call still resolves `ok`, so the agent never learns its input went somewhere else. It retries, and the user's prompt fills with fragments of the agent's typing.

The same root cause shows up as focus theft: `mousePressed` moves focus onto the guest and nothing gives it back, so the user's cursor silently jumps out of the terminal they were working in.

## Why main can't fix it

`webContents.focus()` is a no-op for a `<webview>` guest — the guest is a DOM element owned by the host renderer, so only the renderer can focus it. Confirmed by measurement: text reached the page only when a human physically clicked into it. `Page.bringToFront`, `tab switch --focus`, and bringing the app to the foreground all failed to make input land.

## The fix

Route the focus handoff through the host renderer.

```
main (cdp-ws-proxy)                renderer (BrowserPane)
──────────────────────────────────────────────────────────
input command detected
  → 'begin' + guestId ───────────→ save document.activeElement
                                    focusWebviewNow()
  ← granted / refused ───────────  did the guest actually take focus?
  forward CDP input ─────────────→ guest holds focus → input lands
  → 'end' ───────────────────────→ restore the saved element
```

Main cannot see whether the guest took focus — only the pane hosting it can — so the pane answers the borrow and main acts on the answer:

- **granted** → forward the input.
- **refused** → fail the command. A guest can refuse focus while staying perfectly alive: a hidden or detached pane still answers CDP, so forwarding would put the keystroke back in whatever the user is typing into. Failing is what lets the agent learn its input did not land, instead of getting the `ok` that hid this bug in the first place.
- **silence** → fail it too. The host renderer accepted the announcement, so no answer means no pane is holding this guest — it is mounting, unmounting, or being swapped out — and none of those can take focus. The wait is bounded at 250 ms, which is slack for a busy renderer rather than an expected pause; a mounted pane answers immediately and never reaches it.

Only a guest with **no host renderer at all** — offscreen, headless — skips the borrow and forwards straight through, exactly as it did before this handoff existed.

Scope is deliberately narrow:

- `Input.insertText` / `Input.dispatchKeyEvent` always borrow — key events reach nothing without focus.
- `Input.dispatchMouseEvent` borrows only for `mousePressed`. Moves, wheels and releases never take focus, so borrowing for them would create the very focus hop this prevents.
- `Runtime.evaluate`, screenshots, and read-only traffic are untouched.

The renderer listener is intentionally **not** gated on the pane being active. Agents drive background tabs while the user works elsewhere — that is exactly when the leak happened. Panes filter by guest `webContents` id so only the addressed one reacts.

Also included: `orca tab create` never sent `activate`, even though `TabCreate` documents *"User-initiated opens focus the tab; agent/automation opens stay background."* Every agent-created tab surfaced the browser pane and took focus. It now mirrors `tab switch` — `--focus` is opt-in.

## How this was verified

Measured per command by focusing a probe input in the page, firing the command, then reading the field back. A command is "leaking" when the field stays empty and the text appears in the user's terminal instead.

| Command | Before | After |
|---|---|---|
| `inserttext` / `type` / `keypress` | leaked into the user's terminal | lands in the page |
| `mouse down` / `up`, `click` | stole focus, never returned | focus returns to the user |
| `tab create` | surfaced the pane, took focus | stays in the background |
| `eval`, `screenshot`, `mouse move` | already fine | unchanged |
| `snapshot`, `set device`, `goto`, `tab switch`, `tab close` | already fine | unchanged |

Every command an agent session had actually issued was measured, not just the suspected ones — the innocent ones were confirmed innocent so the fix could stay minimal.

Reproduced in the failing configuration — browser as a **background** tab while the user typed in a floating terminal — before and after. Focus stayed with the user and the input reached the page.

The refused-focus case is covered by tests rather than by hand: it needs a guest that stays alive while its pane cannot hold focus, which the suites drive directly through the borrow reply.

## No visual change

This changes focus behavior only. No UI is added, removed, or restyled, so there are no screenshots to attach.

## Tests

`cdp-ws-proxy` suites updated and extended: the mock now carries a host renderer that answers borrows the way a mounted pane does, and the input paths assert the borrow is announced (`ui:browserAgentInput` / `begin`) instead of the old native `webContents.focus()`.

`cdp-ws-proxy-focus-borrow.test.ts` covers what main does with each answer — forwards on a grant, fails the command on a refusal (key events and `Input.insertText` both), fails on silence, and skips the ask entirely when the guest has no host renderer. `browser-focus.test.ts` covers the pane side: a failed focus is reported as such, and still unwinds its borrow so later restores are not swallowed.

980 tests pass across `src/main/browser`, `src/renderer/src/components/browser-pane` and `src/cli/browser.test.ts`, on top of the current `main`.

## Code review summary

- **Cross-platform** — No platform branches. `hostWebContents`, `send`, and DOM focus are Electron/DOM APIs available on macOS, Linux, and Windows. Verified on macOS; the logic contains nothing OS-specific.
- **SSH / remote / local** — Remote panes stream a viewport instead of hosting a local `<webview>`, so they never take this path; the web preload API gets a no-op stub for both directions. Offscreen/headless guests have no host renderer, so they skip the borrow entirely rather than waiting on an answer that cannot come.
- **Agents / integrations** — Only the agent-browser CDP path is affected. No agent, git provider, or integration-specific behavior is touched.
- **Performance** — Input now waits for the pane's answer instead of a fixed frame, so the common path got *faster*: the old 16 ms per key event and per mouse press is gone. A guest whose pane never answers pays the 250 ms bound before failing, and one with no host renderer pays nothing. Wait-polling and read-only probes (`Runtime.evaluate`, screenshots, mouse moves) are untouched.
- **UI quality** — No visual change. Focus visibly returns to where the user left it rather than parking on the browser pane.
- **Security** — The IPC carries a phase string, a numeric guest id and a numeric borrow id outbound, and a borrow id plus a boolean inbound. No new privilege, no page content, no user data crosses the boundary.

## Testing notes

- Verified on macOS with a local worktree and the desktop `<webview>` path.
- Remote/SSH and headless paths are covered by construction (stub + skipped borrow) but were not exercised on real remote hosts.
- Rebased onto current `main` (`54a9b58`). `browser-pane/` was refactored substantially in the meantime; there were no textual conflicts and the suites above pass on top of it.

---

X: [@songminsns63384](https://x.com/songminsns63384)

### DIFF
--- a/src/cli/browser.test.ts
+++ b/src/cli/browser.test.ts
@@ -177,8 +177,26 @@ describe('orca cli browser page targeting', () => {
       {
         url: 'https://example.com',
         worktree: undefined,
-        profileId: 'work'
+        profileId: 'work',
+        // Why: agent opens stay in the background unless --focus asks otherwise.
+        activate: false
       },
+      { timeoutMs: 60_000 }
+    )
+  })
+
+  it('asks tab create to focus the new tab only with --focus', async () => {
+    queueFixtures(callMock, okFixture('req_create', { browserPageId: 'page-4' }))
+    vi.spyOn(console, 'log').mockImplementation(() => {})
+
+    await main(
+      ['tab', 'create', '--url', 'https://example.com', '--worktree', 'all', '--focus', '--json'],
+      '/tmp/not-an-orca-worktree'
+    )
+
+    expect(callMock).toHaveBeenCalledWith(
+      'browser.tabCreate',
+      expect.objectContaining({ activate: true }),
       { timeoutMs: 60_000 }
     )
   })
--- a/src/cli/handlers/browser-tab.ts
+++ b/src/cli/handlers/browser-tab.ts
@@ -60,9 +60,12 @@ export const BROWSER_TAB_HANDLERS: Record<string, CommandHandler> = {
     const url = getOptionalStringFlag(flags, 'url')
     const profileId = getOptionalStringFlag(flags, 'profile')
     const worktree = await getBrowserWorktreeSelector(flags, cwd, client)
+    // Why: TabCreate documents "agent/automation opens stay background", but the CLI
+    // never sent `activate`, so every agent-created tab surfaced the browser pane and
+    // took focus from whatever the user was working in. Mirror `tab switch`: opt-in.
     const result = await client.call<{ browserPageId: string }>(
       'browser.tabCreate',
-      { url, worktree, profileId },
+      { url, worktree, profileId, activate: flags.has('focus') },
       { timeoutMs: 60_000 }
     )
     printResult(result, json, (v) => `Created tab ${v.browserPageId}`)
--- a/src/cli/specs/browser-basic.ts
+++ b/src/cli/specs/browser-basic.ts
@@ -178,8 +178,9 @@ export const BROWSER_BASIC_COMMAND_SPECS: CommandSpec[] = [
   {
     path: ['tab', 'create'],
     summary: 'Create a new browser tab in the current worktree',
-    usage: 'orca tab create [--url <url>] [--worktree <selector>] [--profile <id>] [--json]',
-    allowedFlags: [...GLOBAL_FLAGS, 'url', 'worktree', 'profile']
+    usage:
+      'orca tab create [--url <url>] [--worktree <selector>] [--profile <id>] [--focus] [--json]',
+    allowedFlags: [...GLOBAL_FLAGS, 'url', 'worktree', 'profile', 'focus']
   },
   {
     path: ['tab', 'profile', 'list'],
--- a/src/main/browser/cdp-ws-proxy-focus-borrow.test.ts
+++ b/src/main/browser/cdp-ws-proxy-focus-borrow.test.ts
@@ -118,9 +118,10 @@ describe('CdpWsProxy agent input focus borrow', () => {
     client.close()
   })
 
-  // Why: no pane hosts this guest, so nobody answers. Dropping the input there would
-  // break offscreen and not-yet-mounted guests that worked before the handoff existed.
-  it('forwards input when no pane answers the borrow', async () => {
+  // Why: the host renderer took the begin, so silence means no pane is holding this
+  // guest — it cannot have focus, and forwarding anyway puts the keystroke in the
+  // user's window. Only a guest with no host renderer at all keeps the old path.
+  it('fails input when no pane answers the borrow', async () => {
     mock.webContents.hostWebContents.send.mockImplementation(() => {})
     const client = await connect(endpoint)
 
@@ -130,8 +131,8 @@ describe('CdpWsProxy agent input focus borrow', () => {
       params: { type: 'keyDown', key: 'a' }
     })
 
-    expect(response.error).toBeUndefined()
-    expect(getSendCommandMethods(mock)).toContain('Input.dispatchKeyEvent')
+    expect((response.error as { message: string }).message).toContain('keyboard focus')
+    expect(getSendCommandMethods(mock)).not.toContain('Input.dispatchKeyEvent')
     client.close()
   })
 
--- a/src/main/browser/cdp-ws-proxy-focus-borrow.test.ts
+++ b/src/main/browser/cdp-ws-proxy-focus-borrow.test.ts
@@ -1,0 +1,153 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { CdpWsProxy } from './cdp-ws-proxy'
+import {
+  connect,
+  createMockWebContents,
+  getSendCommandMethods,
+  sendAndReceive,
+  type MockWebContents
+} from './cdp-ws-proxy-test-harness'
+
+const electronMock = vi.hoisted(() => {
+  const focusReplyListeners = new Set<(...args: unknown[]) => void>()
+  return {
+    replyFocusBorrow: (borrowId: number, focused: boolean): void => {
+      for (const listener of focusReplyListeners) {
+        listener({}, { borrowId, focused })
+      }
+    },
+    ipcMain: {
+      on: vi.fn((channel: string, listener: (...args: unknown[]) => void) => {
+        if (channel === 'browser:agentInputFocusReply') {
+          focusReplyListeners.add(listener)
+        }
+      }),
+      removeListener: vi.fn((_channel: string, listener: (...args: unknown[]) => void) => {
+        focusReplyListeners.delete(listener)
+      })
+    }
+  }
+})
+
+vi.mock('electron', () => ({
+  webContents: { fromId: vi.fn() },
+  BrowserWindow: {
+    getFocusedWindow: vi.fn(() => null),
+    fromWebContents: vi.fn(() => null)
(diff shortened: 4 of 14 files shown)


songminsns:
Thanks — both findings were real. Fixed in 45427ed.

**Overlapping begin/end race.** You're right that this defeated the fix for the most common case. A single typed character sends keyDown/char/keyUp and each announces its own pair, so the second `begin` overwrote the saved owner with `null` (the guest already had focus by then) and no `end` had anything to restore — focus stayed parked on the webview, exactly the symptom this PR is about.

Borrows are now counted: only the outermost `begin` captures the owner, only the outermost `end` restores. An unmatched `end` (a command already in flight when the listener mounts) is floored at zero so it can't swallow the next real restore. The rule is extracted to `createAgentInputFocusBorrow` in `browser-focus.ts` so it's testable on its own — covered for single, nested, partially-closed, and unmatched-end sequences.

**mouseReleased overborrow.** Correct, only the press takes focus. `takesFocus` is now `mousePressed` only, so a click no longer pays a frame of latency on the release.

Also added the missing `end`-phase assertions you noted, plus coverage that releases and moves announce nothing.

songminsns:
@coderabbitai both findings were correct — fixed in 3f74b8b.

**Wait for `Input.insertText` before ending the borrow.** This was a real hole in the fix. `forwardCommand` queues the debugger call and returns, so the borrow ended while Chromium had not processed the insert yet; the renderer handed focus back first and the text could still land in the user's window — exactly the leak this PR closes. `forwardInsertText` now awaits `sendDebuggerCommand` for the final dispatch and answers the client from there, so the borrow outlives the event.

Added the deferred-dispatch test you asked for: it holds `Input.insertText` unresolved, asserts `begin` was announced and `end` was **not**, then resolves and asserts `end` follows.

**tab-create contract test.** Right, the exact `toHaveBeenCalledWith` broke. Updated it to expect `activate: false` and added a `--focus` case expecting `activate: true`.

Verified against `main`: the remaining suite failures (`wsl-hook-relay-live.integration`, `agent-exec-handler`, and a flaky `hosted-review-gitea`) reproduce on the base branch and are unrelated to this change.

songminsns:
Rebased onto current `main` (`54a9b58`) and closed a gap this PR left open.

`focusWebviewNow()` reports whether the guest ended up holding focus, and the borrow was discarding that. A guest can refuse focus while staying alive — a hidden or detached pane still answers CDP — so the input went out anyway and landed in whatever the user was typing into, which is the leak this PR is about.

The pane now answers the borrow. Main forwards on a grant, fails the command on a refusal so the agent learns its input did not land, and falls through on silence so guests with no pane — offscreen, headless, not yet mounted — behave exactly as they did before the handoff existed. Waiting for the real signal also replaces the fixed 16 ms per key event and mouse press, so the common path got faster rather than slower.

Writing the tests turned up a second bug on the way: `Input.insertText` dropped the rejection from its borrow, so a refused borrow left the client hanging with no reply at all. Fixed alongside — `forwardInsertText` answers its own failures, so only the refused borrow needed catching.

`browser-pane/` was refactored substantially in the meantime; there were no textual conflicts. 980 tests pass across `src/main/browser`, `src/renderer/src/components/browser-pane` and `src/cli/browser.test.ts`, plus `typecheck`, `oxlint`, `oxfmt` and the max-lines ratchet.

One thing I can't do from my side: no CI has run on this PR at any point — the check list only ever had the review bots, so the workflows look to be waiting on a maintainer to approve them for an outside fork. Could you kick those off?

songminsns:
Good catch — that one was real, and it undercut the fix in exactly the case the PR is about.

That timeout is only reachable *after* the host renderer accepted the announcement, so silence there never meant "no handoff needed" — it meant no pane is holding this guest, because it is mounting, unmounting, or being swapped out. None of those can take focus, and resolving them as granted sent the keystroke out unfocused, into whatever the user was typing into. The comment I left there described the situation correctly and then drew the wrong conclusion from it: "pre-handoff behaviour" is just the bug this PR exists to remove.

Silence now fails the command like an outright refusal, in `93ef180`. The hostless path stays separate as you suggested — a guest with no host renderer never announces a borrow at all and forwards straight through, so offscreen and headless are untouched.

I also raised the bound from 64 ms to 250 ms. It used to be a fixed pause to absorb; now it decides whether a command fails, so it needs slack for a renderer that is briefly busy rather than being tuned to one frame. The granted path never reaches it — a mounted pane answers immediately — so nothing gets slower.

Test updated to expect the failure, and the hostless case keeps its forwarding coverage. 980 tests still pass across `src/main/browser`, `browser-pane/` and `cli/browser.test.ts`, along with typecheck, oxlint and oxfmt.

songminsns:
Checking in on where this stands before I put time into the rebase it now needs.

**Current state.** This conflicts with `main` again — 897 commits have landed since my last push on Aug 9. It is not a plain replay this time: `Split browser pane` (#14861) and `split browser and browser-IPC modules` (#14697) moved `browser-focus.ts` into `browser-pane/host-guest/` and reshaped the files around it, so the change has to be re-placed rather than rebased. I am happy to do that. I would just rather know it is wanted first — this would be the third rebase, and the previous two did not move the PR along.

**CI has never run here.** In this PR's whole life the check list has only ever held the review bots — no `static analysis`, no `typecheck`, no `tests node 24/26`, nothing from `pr.yml`. An internal PR from the same week carries 40+ checks, so I assume the workflows are waiting on a maintainer to approve them for an outside fork; I cannot trigger that myself. The branch was green locally at every push — 1006 tests across `src/main/browser`, `browser-pane/` and `cli/browser.test.ts`, plus typecheck, oxlint and oxfmt — but that is my word rather than CI's.

**Overlap with #10875.** #15309 flagged the two of us together and I think that flag was right. #10875 restores the host `WebContents` focus owner after `mousePressed` and deliberately leaves DOM focus alone; this PR lends the guest DOM focus for the duration of the input and hands it back, covering `Input.insertText` and key events as well as the press. Same symptom, different layers — and both touch the `mousePressed` path in `cdp-ws-proxy.ts`. Whichever way you want it to go — land one, land both with one rebased onto the other, or fold them together — I am glad to adapt. I would just rather not rebase into a conflict with something you are about to merge.

So: is this one you want to take? If yes, I will rebase onto current `main` right away and report back here once it is green. If the direction is #10875 instead, say so and I will close this — no hard feelings; the investigation behind it is written up either way.

@OrcaWin @nwparker

Jinwoo-H:
Thanks @songminsns for the investigation and implementation. We are taking ownership of the browser-focus bug in canonical issue #16674, which covers the current Windows tab-create/switch focus-stealing path. This branch is based on older code and will be superseded by a current-main implementation; no contributor rebase, retest, or other follow-up is requested. Closing this PR while preserving your credit in the issue and owner implementation.

Issue #8498, opened 2026-07-13
[Bug]: Mobile Projects view frequently shows an incomplete worktree list and cannot be manually refreshed


mistcoversmyeyes:
### Operating system

Windows host connected through the Orca mobile companion app

### Orca version

Current release (exact mobile and desktop versions can be added if needed)

#### Short summary

After the Orca mobile app successfully connects to the PC host, the Projects/worktree view frequently shows only a subset of the host's worktrees. There is no way to manually refresh this view, and background refreshes only sometimes restore the complete list.

This is not an occasional edge case: it occurs in roughly 80% of connections/loads in my environment.

#### Steps to reproduce

1. Start Orca on a PC that has multiple projects and worktrees.
2. Connect the Orca mobile app to that PC and wait until the connection is shown as successful.
3. Open the connected host's Projects/worktree view on mobile.
4. Compare the displayed worktrees with the full list visible on the PC.

#### Actual behavior

- Most of the time, only some of the current worktrees are displayed on mobile.
- The Projects/worktree view has no user-triggered refresh action or pull-to-refresh gesture.
- After a passive/background refresh, the full worktree list may appear, but this is nondeterministic.

#### Expected behavior

- Once the mobile app reports a successful connection, the Projects/worktree view should display the complete current host state, including every worktree visible on the PC.
- The view should provide a manual refresh action so users can recover immediately if the initial snapshot is incomplete or stale.

#### Additional context

PR #7500 addressed worktrees being hidden by host scope or incomplete repository metadata. This report describes the remaining user-visible behavior: the mobile view still frequently starts with an incomplete worktree snapshot and offers no way to explicitly request a fresh one.

nwparker:
Fixed by #8690 (merged to `main` as `8e0977d295`).

The mobile host screen now refetches worktrees on the transition into `connected` (reconnect), supports **pull-to-refresh**, and writes the confirmed snapshot back through `setCachedWorktrees` so a reconnect or remount can't serve a stale/partial Projects list.

Thanks @branben for the original investigation and fix in #8605 / #8591.

PR #11735, opened 2026-07-31
perf(mobile): avoid unchanged worktree catalog payloads

nwparker:
## ELI5

Every three seconds, the mobile app asks the host, “What worktrees do you have right now?” Previously, the host mailed back the entire catalog even when nothing changed—477 KB on the measured host.

With this change, the host still rebuilds and checks the complete catalog for safety. It labels that catalog with a fingerprint of its contents. If the phone already holds that exact fingerprint, the host replies with a ~192-byte “same as X” receipt and the phone reuses the catalog it already confirmed. If the fingerprint is unknown, either side is old, or anything changed, the host sends the complete catalog again.

Because the label is derived from the contents rather than from bookkeeping, any number of phones, any request size, and any host restart all line up automatically—the labels can only match when the catalogs are identical.

## Summary

- add opt-in, content-addressed snapshot IDs to `worktree.ps`
- let the mobile host screen request a tiny unchanged response while retaining the confirmed rows it reapplies on every poll
- preserve exact legacy responses for old callers and automatically fall back to full responses with old hosts

## Measured impact

Measured against a live 288-worktree host:

| Response | Compact RPC bytes |
| --- | ---: |
| Full snapshot | 477,423 |
| Unchanged snapshot | ~192 |

That avoids roughly 477,231 bytes per unchanged poll (99.96%). At the existing three-second cadence, this is about 9.11 MiB/minute or 546 MiB/hour per active client before transport overhead.

The savings are redundant serialization/transfer, relay encryption/decryption work, mobile JSON parsing/allocation, cache writes, and downstream list work. This deliberately still builds the complete host catalog and preserves the safety poll—catalog construction is unchanged.

**Host CPU.** The steady-state path is an exact deep comparison against the last catalog, measured at about 0.418 ms/poll on this catalog versus about 0.581 ms/poll for JSON serialization alone. Hashing is reached only when that comparison fails, i.e. only on polls that were going to send a full payload anyway. Hashing unconditionally instead was measured at roughly 3x the compare on a synthetic catalog of similar size, which is why the memo in front of it stays.

## Design

`snapshotId` is a truncated sha256 over the serialized catalog—HTTP `ETag` semantics. Snapshot ownership therefore lives entirely in the ID rather than in server-side bookkeeping:

- **Concurrent clients** converge with no coordination. A client presenting an unknown ID gets the current full catalog carrying the current ID.
- **Differing `limit` values** need no isolation, because a limit that changes the response also changes its fingerprint, and one that does not is genuinely the same catalog.
- **Restarts** are safe without scoping. IDs match only when catalogs are byte-identical, which is exactly when the client's retained rows are already correct.

A single-slot memo sits in front of the hash so the common case stays a deep compare. The memo is a pure CPU optimization and is deliberately outside the correctness argument: because IDs derive from content, dropping or thrashing that slot can change performance and nothing else.

On the client, an unchanged response still yields the confirmed rows, and `HostScreen` reapplies them exactly as it does for a full response. Every poll therefore continues to reassert host truth over optimistic local edits (pin toggles, delete rollbacks) and continues to overwrite the home-written worktree cache (#8498). This costs nothing in the steady state: the cache write is an in-memory `Map.set`, and `areWorktreeListsEqual` already ran on every poll and still short-circuits to the existing array on an identical catalog, keeping SectionList/sort rebuilds off the tap path.

## Proof of no regression

| Invariant | Evidence |
| --- | --- |
| The three-second safety poll still executes the full host catalog build | Method test makes two conditional polls and asserts `getWorktreePs` ran twice. No code inside `getWorktreePs` changed. |
| Any catalog change produces a full response with a new ID | `issues a new snapshot id for changed content`; `repairs a caller holding a superseded snapshot with a full response`. |
| Concurrent callers cannot displace each other | `keeps concurrent callers independent without server-held state` interleaves two catalogs and asserts both callers still get `unchanged`. |
| Server-side memo state cannot affect the answer | `derives ids from content alone, so memo state cannot change the answer` displaces the memo and asserts the ID is unchanged. |
| A client presenting an unknown ID is repaired | `returns unchanged only when the caller owns the exact current snapshot` asserts the full current result carrying the current ID. |
| Generated IDs satisfy the request schema | `produces ids within the request schema bound`. |
| Old clients receive the exact legacy shape | Method test omits `afterSnapshotId` and asserts no `snapshotId` is added. |
| New mobile remains compatible with old hosts | Old schemas strip the unknown optional request field; `treats an old-host full response as authoritative and clears the token`. |
| Unchanged polls still reassert host truth over optimistic edits | `returns the confirmed rows on unchanged responses so callers can reassert them`; `HostScreen` applies them through the same path as a full response. |
| Dropped/modal-blocked responses cannot advance the token | Token advances only in `admit`, after the existing current-client/current-host/modal guards; `does not advance the token until the caller admits a response`. |
| Superseded responses cannot corrupt current ownership | `drops a superseded host response without invalidating the current token`; `resets snapshot ownership when the client or host changes`. |
| Transport failures and malformed responses recover safely | `preserves the last admitted token across transport failures`; `clears snapshot ownership after a mismatched unchanged response`; malformed-payload cases force a full repair. |
| A future `unchanged` catalog field cannot reclassify a full response | Responses are discriminated on the presence of rows: `classifies by rows, so a future 'unchanged' catalog field cannot hide a full response`. |
| SSH, relay, Git providers, and folder workspaces retain their behavior | Snapshotting treats the completed `worktree.ps` result as opaque data and does not alter catalog construction, provider access, paths, Git commands, or polling. |

Validation performed locally:

- 2,784 mobile tests passed (3 skipped)
- 1,064 `src/main/runtime/rpc` tests passed (1 skipped)
- node and mobile typechecks passed
- oxlint and oxfmt passed
- max-lines ratchet passed with no new bypasses

## Safety and compatibility

- snapshot IDs are derived from the complete post-build result; any difference yields a full response under a new ID
- unknown or superseded IDs receive a repairing full response
- only advance the mobile token after the current-client/current-host/modal admission guards
- retain the token across transport failures and superseded responses, but clear it for malformed or mismatched unchanged responses
- old client → new host: exact legacy response, without `snapshotId`
- new client → old host: unknown request field is stripped and the full legacy response remains authoritative
- folder workspaces, SSH/relay hosts, and the existing three-second refresh path are unchanged

## Follow-ups (deliberately not in this PR)

- **Capability-gated opt-in.** `afterSnapshotId` currently carries three states (`undefined` = legacy caller, `null` = opted in holding nothing, string = holding an ID). The repo already has the idiomatic mechanism for this—`status.get` capabilities read via `startRuntimeCapabilityProbe`—and `TERMINAL_QUERY_REPLY_INPUT_RUNTIME_CAPABILITY` exists for the identical hazard of an old host silently stripping an unknown field. Advertising `worktree.catalog-snapshot.v1` would reduce the field to `string | undefined`.
- **Generalize to the RPC envelope.** Conditional responses are an envelope concern, not a `worktree.ps` concern. Moving `afterSnapshotId` onto `RpcRequest` and `snapshotId`/`notModified` onto `_meta`, gated by a flag on `defineMethod`, would let other polled methods (`ui.get`, repo metadata, terminal lists) opt in without their own result unions or client wrappers.

## Review loops

- [Loop #1 — CHANGES REQUESTED → FIXED](https://github.com/stablyai/orca/pull/11735#issuecomment-5140992779): fixed cross-limit cache thrashing with a bounded per-limit LRU and three regression cases.
- [Loop #2 — CLEAN](https://github.com/stablyai/orca/pull/11735#issuecomment-5141010949): two independent Sol reviewers found no actionable issues on the updated head.
- Loop #3 — CHANGES REQUESTED → FIXED (`3e6a62d`): restored unconditional reassertion of the confirmed catalog on the client, stopped superseded responses from clearing the current token, and replaced the per-limit LRU from Loop #1 with content-addressed IDs—which removes the cross-limit thrashing class outright rather than bounding it.

### DIFF
--- a/mobile/app/h/[hostId]/index.tsx
+++ b/mobile/app/h/[hostId]/index.tsx
@@ -94,6 +94,7 @@ import {
 import { useWorkspaceSections } from '../../../src/worktree/use-workspace-sections'
 import { getMobileWorkspaceLineageGroupKey } from '../../../src/worktree/mobile-workspace-lineage'
 import { areWorktreeListsEqual } from '../../../src/worktree/worktree-list-snapshot'
+import { WorktreeCatalogSnapshotClient } from '../../../src/worktree/worktree-catalog-snapshot-client'
 import { repoColor } from '../../../src/worktree/repo-color'
 import {
   WORKSPACE_GROUP_OPTIONS as GROUP_OPTIONS,
@@ -141,6 +142,7 @@ export function HostScreen({
   const lastConnectedAt = useLastConnectedAt(hostId)
   const clientRef = useRef<RpcClient | null>(null)
   const fetchWorktreesInFlightRef = useRef(false)
+  const worktreeCatalog = useMemo(() => new WorktreeCatalogSnapshotClient(), [])
   const fetchRepoMetadataInFlightRef = useRef(new WeakSet<RpcClient>())
   const fetchRepoMetadataPendingRef = useRef(new WeakSet<RpcClient>())
   const repoMetadataFetchedAtRef = useRef(0)
@@ -423,27 +425,28 @@ export function HostScreen({
       const requestHostId = hostId
 
       try {
-        // Why: worktree.ps silently truncates at 200; use a high cap so large hosts don't drop workspaces.
-        const response = await requestClient.sendRequest('worktree.ps', { limit: 10000 })
+        const pendingCatalog = await worktreeCatalog.fetch(requestClient, requestHostId)
         if (clientRef.current !== requestClient || hostId !== requestHostId) {
           return
         }
         if (!options.allowDuringModal && newWorktreeModalVisibleRef.current) {
           return
         }
-        if (response.ok) {
-          const result = (response as RpcSuccess).result as { worktrees: Worktree[] }
-          // Why: reuse the existing array on identical snapshots to keep SectionList/sort rebuilds off the tap path.
-          setWorktrees((current) =>
-            areWorktreeListsEqual(current, result.worktrees) ? current : result.worktrees
-          )
-          setLastKnownWorktrees((current) =>
-            areWorktreeListsEqual(current, result.worktrees) ? current : result.worktrees
-          )
-          setWorktreesLoaded(true)
-          // Why (#8498): overwrite the home-written cache with the confirmed snapshot so a reconnect/remount can't serve a stale list.
-          if (hostId) {
-            setCachedWorktrees(hostId, result.worktrees)
+        const result = worktreeCatalog.admit(pendingCatalog)
+        if (result) {
+          if (result.changed) {
+            // Why: reuse the existing array on identical snapshots to keep SectionList/sort rebuilds off the tap path.
+            setWorktrees((current) =>
+              areWorktreeListsEqual(current, result.worktrees) ? current : result.worktrees
+            )
+            setLastKnownWorktrees((current) =>
+              areWorktreeListsEqual(current, result.worktrees) ? current : result.worktrees
+            )
+            setWorktreesLoaded(true)
+            // Why (#8498): overwrite the home-written cache with the confirmed snapshot so a reconnect/remount can't serve a stale list.
+            if (hostId) {
+              setCachedWorktrees(hostId, result.worktrees)
+            }
           }
           // Drop the optimistic active override once the host reports it active, so later desktop changes win.
           setOptimisticActiveWorktreeId((pending) =>
@@ -487,7 +490,7 @@ export function HostScreen({
         fetchWorktreesInFlightRef.current = false
       }
     },
-    [client, connState, hostId]
+    [client, connState, hostId, worktreeCatalog]
   )
 
   useFocusEffect(
--- a/mobile/app/h/[hostId]/index.tsx
+++ b/mobile/app/h/[hostId]/index.tsx
@@ -142,7 +142,9 @@ export function HostScreen({
   const lastConnectedAt = useLastConnectedAt(hostId)
   const clientRef = useRef<RpcClient | null>(null)
   const fetchWorktreesInFlightRef = useRef(false)
-  const worktreeCatalog = useMemo(() => new WorktreeCatalogSnapshotClient(), [])
+  // Why: useRef, not useMemo — React may discard memoized values, which would silently
+  // reset the snapshot token this object exists to own.
+  const worktreeCatalogRef = useRef(new WorktreeCatalogSnapshotClient())
   const fetchRepoMetadataInFlightRef = useRef(new WeakSet<RpcClient>())
   const fetchRepoMetadataPendingRef = useRef(new WeakSet<RpcClient>())
   const repoMetadataFetchedAtRef = useRef(0)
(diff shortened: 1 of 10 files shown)


nwparker:
## Sol review loop #1 — CHANGES REQUESTED → FIXED

- **Mobile lifecycle/performance reviewer:** CLEAN; no actionable findings.
- **Protocol/concurrency reviewer:** one low-severity performance finding. Conditional callers alternating different `limit` values displaced the single cached snapshot and forced full responses.
- **Resolution:** fixed in `67f6c9c807` with a bounded eight-entry per-limit LRU. Added regressions for alternating limits, bounded retention, and recency.
- **Post-fix validation:** 9 focused host tests, node typecheck, changed-file lint/format, and `git diff --check` passed.

Loop #2 will re-review the updated head.

nwparker:
## Sol review loop #2 — CLEAN

Two independent Sol reviewers re-reviewed updated head `67f6c9c807`; neither found an actionable issue.

Verified risks:

- alternating limits retain independent snapshot tokens
- equal-content hits and changed-content replacements refresh LRU recency
- bounded eviction repairs with a full response and new token
- same- and cross-limit completion ordering cannot produce a false `unchanged`
- runtime restart, malformed/mismatched tokens, old/new client-host combinations, and same-limit multi-client sharing remain safe
- mobile current-client/current-host/modal admission, transport failure recovery, optimistic-state reconciliation, and cache semantics remain sound
- SSH/relay, folder workspaces, platforms, paths, and Git providers remain opaque and unchanged

Reviewer validation included 31 host cache/RPC tests, 9 mobile client tests, extra default-limit/recency/eviction/concurrency repros, node/mobile typechecks, changed-file lint, and `git diff --check`.

**Final loop verdict: CLEAN.**

brennanb2025:
## Review + live QA — CLEAN at `7025888466`

The same-model review loop found one meaningful correctness issue, fixed it, pushed it non-force, and re-reviewed the resulting head clean: the host memo retained caller-owned/nested catalog references, so later mutation could make an old ID label changed content and falsely return `unchanged`. The fix memoizes a JSON-wire-equivalent copy only on first/changed full paths; the steady-state compare remains unchanged.

Validation:

- Runtime RPC: 1,065 passed / 1 skipped; mobile: 2,784 passed / 3 skipped
- Node and mobile typechecks, changed-code quality, max-lines, oxlint, oxfmt, and `git diff --check` passed
- Independent focused run: 9 host snapshot/method tests + 12 mobile client tests passed
- Independent 287-row / 488,597-byte synthetic catalog: unchanged result 66 bytes (99.986% reduction); steady-state resolve 0.161 ms median / 0.285 ms p95; changed/full path 1.144 ms median / 1.545 ms p95
- Nested-reference mutation repro now returns a full response with a new ID

Backcompat review is clean in code/tests:

- old mobile → new desktop omits `afterSnapshotId` and receives the exact legacy shape
- new mobile → old desktop: the pre-PR Zod object strips the unknown field; the legacy full response is authoritative and clears snapshot ownership
- no persisted keys, routes, deep links, Git/provider behavior, SSH/relay behavior, or folder-workspace catalog construction changed

Live QA:

- Electron: attached to `koi @ brennanb2025/review-pr-11735-head`; paired-device UI rendered with 0 console errors
- iOS Simulator: iPhone 17 Pro, iOS 26.5; Metro ran from `mobile/` on port 8086; the development URL loaded without a redbox
- paired over LAN to the exact branch host; Host 29 rendered the branch worktree
- after the host display name changed to `PR 11735 Snapshot QA`, mobile showed that new value on the next poll, proving the changed-catalog full repair; the temporary metadata was then restored
- current→current live polling was exercised. The old/new version-skew matrix was validated at schema/client integration-test level rather than by launching an old packaged desktop binary.

### Electron host / paired state

Electron host paired with the simulator

### iPhone catalog after live snapshot repair

iPhone 17 Pro showing the repaired catalog

brennanb2025:
Supplemental Android validation on the exact pushed PR head:

- Paired the Android development client securely to the isolated PR desktop host over direct LAN and rendered the real `koi` worktree on `brennanb2025/review-pr-11735-head`.
- Ran 12 live unchanged snapshot requests: 1,033-byte full response to 66-byte unchanged responses; all 12 matched the same snapshot ID and returned `unchanged: true` (3.2 ms average, 4.1 ms max).
- Applied a reversible desktop worktree-title mutation; Android displayed `PR 11735 Android Snapshot QA` on the next 3-second poll. The original title was restored and observed on Android afterward.
- Existing bidirectional backcompat, iPhone, stress, constrained-heap, fuzz, unit, lint, typecheck, packaging, and CI validation remain green.

Android live catalog update

PR #11760, opened 2026-07-31
fix(rate-limits): fall back to context_window for Claude usage on Vertex/proxy backends

YuSoeun:
## Summary
- The managed Claude statusline script only forwarded the per-turn payload to Orca's agent-hook endpoint when it contained a `rate_limits` key, which Anthropic only includes in standard Claude.ai-subscription API responses.
- Deployments routed through Vertex AI or an internal Anthropic-API-compatible proxy never receive `rate_limits`, so the live usage bar silently stopped updating even though the network itself was fine.
- `context_window` is reported by every backend (computed locally from conversation length, not pulled from Claude.ai billing), so use it as a fallback session-usage source — both in the shell script's early-exit guard (POSIX `case`/Windows `findstr`) and in `parseClaudeStatusLineBody` — when `rate_limits` is absent.

## Test plan
- [x] `src/shared/claude-statusline-rate-limits.test.ts` — added coverage for the `context_window` fallback and for `rate_limits` taking priority when both are present
- [x] `src/main/claude/statusline-script.test.ts` — updated the POSIX/Windows guard assertions and added a behavioral test confirming a `context_window`-only payload now reaches curl

### DIFF
--- a/src/main/claude/statusline-script.test.ts
+++ b/src/main/claude/statusline-script.test.ts
@@ -17,11 +17,12 @@ afterEach(() => {
 })
 
 describe('getManagedStatusLineScript (posix)', () => {
-  it('guards on rate_limits before sourcing the endpoint or spawning curl', () => {
+  it('guards on rate_limits or context_window before sourcing the endpoint or spawning curl', () => {
     stubPlatform('darwin')
     const script = getManagedStatusLineScript('local')
     expect(script).toBe(getManagedStatusLineScript('posix'))
     const guardIndex = script.indexOf('*\'"rate_limits"\'*')
+    expect(script).toContain('*\'"context_window"\'*')
     const endpointIndex = script.indexOf('ORCA_AGENT_HOOK_ENDPOINT')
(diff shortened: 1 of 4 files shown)


Jinwoo-H:
Thanks @YuSoeun for the Claude  fallback work. We consolidated this with #11766 and #11662 in canonical issue #17518, covering provider RPC, PTY, and local-log fallback behavior. We will reimplement the matrix on current main; no contributor follow-up is requested. Closing this superseded branch while preserving your credit.

Jinwoo-H:
Correction: the intended reference is the Claude  fallback. We consolidated this with #11766 and #11662 in canonical issue #17518 and will reimplement on current main. No contributor action is requested; your credit remains preserved.

Jinwoo-H:
Correction: the intended reference is the Claude context_window fallback. We consolidated this with #11766 and #11662 in canonical issue #17518 and will reimplement on current main. No contributor action is requested; your credit remains preserved.

PR #11766, opened 2026-07-31
fix(rate-limits): fall back to the local session log for Codex usage

YuSoeun:
## Summary
- Unlike Claude, Codex has no live per-turn push into Orca — every usage refresh goes through chatgpt.com (backend fetch, RPC app-server, or a PTY `/status` probe), all of which require reaching OpenAI's network. When all three fail, the usage bar went completely dark with no local fallback (see the related Claude fix in #11760 for the internal-network usage investigation this follows up on).
- Codex's own session log already records a `token_count` event with `total_token_usage` and `model_context_window` on every turn. `fetchCodexRateLimits` now reads the most recently modified session log's last `token_count` event as a last-resort, network-free usage signal when every network path returns a genuine error.
- Auth errors are explicitly excluded from the fallback — masking "you need to re-sign in" with a locally-derived `ok` would hide the one thing the user actually needs to act on.

## Testing
- [x] `pnpm typecheck` (targeted: `tsc --noEmit -p config/tsconfig.node.json`)
- [x] `pnpm lint` (targeted oxlint + type-aware + native-plugins + max-lines-ratchet + reliability-gates on changed files)
- [x] `pnpm test` (targeted: all `src/main/rate-limits/codex-*.test.ts` — 89 tests passing, including 21 pre-existing codex-fetcher tests unchanged and 2 new integration tests for the fallback)
- [x] `pnpm build` — not run; native module rebuild requires Visual Studio Build Tools not available in this environment (unrelated to this change, which is TS-only)

New coverage:
- `codex-session-log-usage.test.ts` — pure `token_count` parsing (last-event priority, `last_token_usage` fallback, malformed lines, missing `model_context_window`, clamping) and the file-selection wrapper via dependency injection
- `codex-fetcher.test.ts` — the fallback engaging when every network path exhausts to `status: 'error'`, and *not* engaging when the fetch was aborted

## AI Review Report
Reviewed with Claude Code. Checked: cross-platform (pure JSON/string parsing, no OS-specific paths — reuses the existing `codex-usage/scanner.ts` directory discovery which already handles Windows/WSL/macOS/Linux session locations), SSH/remote compatibility (session log discovery is unchanged, local-only by design since this is a local fallback), auth-error handling (verified via test that auth errors are never masked by the fallback — this was caught as a real bug during review: an early version applied the fallback to auth-error results too, which surfaced during testing as a live session on the dev machine getting silently marked "ok"), and regression safety (all pre-existing codex-fetcher and codex-fetcher-auth-errors tests pass unchanged).

## Security Audit
- Reads only local files already covered by the existing Codex session-log scanner's directory allowlist; no new file paths or write operations introduced.
- No new network calls, credentials, or IPC surface — purely a local-file read gated behind the same auth-presence check the rest of `fetchCodexRateLimits` already uses.
- JSON parsing is wrapped in try/catch per line; malformed session logs degrade to `null` rather than throwing.

## Notes
- `weekly` stays `null` in the fallback result since the local session log has no 7-day window data — only the 5-hour `session` window is populated.
- Twitter/X handle: not provided.

### DIFF
--- a/src/main/rate-limits/codex-fetcher-session-log-abort.test.ts
+++ b/src/main/rate-limits/codex-fetcher-session-log-abort.test.ts
@@ -1,0 +1,71 @@
+import { EventEmitter } from 'node:events'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const { childSpawnMock, readFileMock, readLatestCodexSessionUsedPercentMock } = vi.hoisted(() => ({
+  childSpawnMock: vi.fn(),
+  readFileMock: vi.fn(),
+  readLatestCodexSessionUsedPercentMock: vi.fn()
+}))
+
+vi.mock('node:child_process', () => ({ spawn: childSpawnMock }))
+vi.mock('node:fs/promises', () => ({ readFile: readFileMock }))
+vi.mock('../codex-cli/command', () => ({ resolveCodexCommand: () => 'codex' }))
+vi.mock('node-pty', () => ({ spawn: vi.fn() }))
+vi.mock('./codex-auth-presence', () => ({
+  probeCodexAuthPresence: vi.fn(async () => 'present')
+}))
+vi.mock('./codex-session-log-usage', () => ({
+  readLatestCodexSessionUsedPercent: readLatestCodexSessionUsedPercentMock
+}))
+
+import { fetchCodexRateLimits } from './codex-fetcher'
+
+function makeRpcChild() {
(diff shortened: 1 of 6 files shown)


Jinwoo-H:
Thanks @YuSoeun for the Codex session-log fallback work. We consolidated this with #11760 and #11662 in canonical issue #17518, covering provider RPC, PTY, and local-log fallback behavior. We will reimplement the matrix on current main; no contributor follow-up is requested. Closing this superseded branch while preserving your credit.

PR #11772, opened 2026-07-31
fix(ssh): honor custom Node tool directories

### LINKED ISSUE #11733: [Bug]: SSH relay does not honor custom MISE_DATA_DIR when resolving Node.js
### Operating system

Windows

### Orca version

1.4.162

### Details

# Summary

When connecting from Orca on Windows to a Linux SSH host, SSH relay setup does not discover Node.js when mise stores tools outside its default data directory through `MISE_DATA_DIR`.

The initial visible error was:

```text
Error invoking remote method `ssh:connect`: Error: Relay package for linux-x64 not found locally. This may be a packaging issue - try reinstalling Orca.
```

After relay packaging was available, diagnosing the connection showed that the remote Node resolver only probed the default mise paths under `$HOME/.local/share/mise`. On this host, mise set `MISE_DATA_DIR` to a different directory, where Node.js and npm were installed. Orca therefore reported Node.js as unavailable even though a usable Node/npm toolchain existed.

## Steps to reproduce

1. On a Linux SSH host, install mise and Node.js.
2. Set `MISE_DATA_DIR` to a non-default absolute directory, for example `/opt/mise-data`.
3. Install Node through mise so the executable is available at `$MISE_DATA_DIR/installs/node/<version>/bin/node` and the shim is at `$MISE_DATA_DIR/shims/node`.
4. Confirm `node --version` and `npm --version` work in the remote environment.
5. Connect to the host through Orca SSH.

## Expected behavior

Orca should honor `MISE_DATA_DIR` when probing mise shims and installs, and fall back to `$HOME/.local/share/mise` only when the variable is unset.

## Actual behavior

The resolver used only `$HOME/.local/share/mise/shims/node` and `$HOME/.local/share/mise/installs/node/*/bin/node`. It missed the valid Node installation in the configured data directory, preventing relay deployment.

## Official mise documentation

[mise configuration: `MISE_DATA_DIR`](https://mise.jdx.dev/configuration.html#mise-data-dir) states that this variable is the directory where mise stores plugins and tool installs. Its Linux default is `~/.local/share/mise` or `$XDG_DATA_HOME/mise`, so installations may legitimately use a different location when `MISE_DATA_DIR` is set.

## Environment

- Orca: 1.4.162
- Local OS: Windows
- Connection: SSH to a Linux host
- Remote runtime manager: mise with a custom `MISE_DATA_DIR`

## Suggested fix

Build the mise paths from `MISE_DATA_DIR` with a default fallback, for example:

```sh
mise_data_dir=${MISE_DATA_DIR:-"$HOME/.local/share/mise"}
"$mise_data_dir/shims/node"
"$mise_data_dir/installs/node"/*/bin/node
```

aliuq:
## Summary

Fixes #11733

Honor custom Node tool-manager directories during SSH relay bootstrap.

## Screenshots

No visual change.

## Testing

- [x] pnpm lint
- [x] pnpm typecheck
- [x] pnpm test
- [x] pnpm build
- [x] Added regression coverage for custom Node tool-manager directories.

## AI Review Report

Reviewed SSH bootstrap path discovery, POSIX shell quoting, macOS/Linux/Windows separation, and Node/npm validation. The Windows PowerShell resolver is unchanged; POSIX path probing remains isolated. Verified the generated path probe preserves one inventory request before candidate validation.

## Security Audit

Reviewed remote command construction and path handling. Candidate paths are shell-quoted before execution; no secrets, authentication, IPC, or dependency changes are introduced.

## Notes

POSIX-only path probing supports custom fnm, mise, asdf, Volta, and n directories. No visual change.

### DIFF
--- a/src/main/ssh/ssh-remote-node-resolution.test.ts
+++ b/src/main/ssh/ssh-remote-node-resolution.test.ts
@@ -91,38 +91,92 @@ describe('resolveRemoteNodePath', () => {
     }
   )
 
-  it('probes mise install directories', async () => {
+  it('probes mise directories from their environment variables', async () => {
     execCommandMock
-      .mockResolvedValueOnce('/home/u/.local/share/mise/installs/node/20/bin/node\n')
+      .mockResolvedValueOnce('/custom/mise/installs/node/20/bin/node\n')
       .mockResolvedValueOnce('v20.11.0\n')
 
     await resolveRemoteNodePath(conn)
 
     const callScript = execCommandMock.mock.calls[0]![1] as string
-    expect(callScript).toContain('"$HOME/.local/share/mise/installs/node"/*/bin/node')
+    expect(callScript).toContain('mise_data_dir=${MISE_DATA_DIR:-}')
+    expect(callScript).toContain(
+      'mise_installs_dir=${MISE_INSTALLS_DIR:-"$mise_data_dir/installs"}'
+    )
+    expect(callScript).toContain('mise_shims_dir=${MISE_SHIMS_DIR:-"$mise_data_dir/shims"}')
+    expect(callScript).toContain('"$mise_shims_dir/node"')
+    expect(callScript).toContain('"$mise_installs_dir/node"/*/bin/node')
   })
 
-  it('probes asdf install directories', async () => {
+  it('probes version-manager directories from their environment variables', async () => {
     execCommandMock
-      .mockResolvedValueOnce('/home/u/.asdf/installs/nodejs/20.11.0/bin/node\n')
+      .mockResolvedValueOnce('/custom/asdf/installs/nodejs/20.11.0/bin/node\n')
       .mockResolvedValueOnce('v20.11.0\n')
 
     await resolveRemoteNodePath(conn)
 
     const callScript = execCommandMock.mock.calls[0]![1] as string
-    expect(callScript).toContain('"$HOME/.asdf/installs/nodejs"/*/bin/node')
+    expect(callScript).toContain('fnm_dir=${FNM_DIR:-"$HOME/.fnm"}')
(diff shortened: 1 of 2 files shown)


Jinwoo-H:
Independently verified — **this is good to merge as-is.**

We reproduced #11733 red on current main before reading the patch: built a fake $HOME with only `$HOME/.local/share/mise-custom/installs/node/20/bin/node` present (i.e. `MISE_DATA_DIR` pointing elsewhere) and ran the pre-fix probe script from `ssh-remote-node-resolution.ts` against it — zero candidates. Confirmed at the vitest level too.

Checked specifically that it fixes the reported mechanism rather than an adjacent one: it replaces every hardcoded version-manager path with `\${ENV_VAR:-default}` indirection (FNM_DIR, MISE_DATA_DIR/MISE_INSTALLS_DIR/MISE_SHIMS_DIR, ASDF_DATA_DIR, VOLTA_HOME, N_PREFIX/N_CACHE_PREFIX) while preserving the previous defaults when unset. Rebases onto current main with zero conflicts.

Mutation-tested: reverting to the hardcoded block fails 4/35 tests; re-applying restores 35/35. Your added test is a real oracle — it executes the generated shell script via `execFileSync` against a real temp tree with every custom env var set, which is exactly the level this needed. `pnpm typecheck` and `oxlint` clean.

One adjacent issue we found while checking the surrounding surface, explicitly **not** something this PR should take on: `src/main/providers/ssh-pty-spawn-env.ts:15-19` only prepends the relay bin dir when it is *absent* from PATH, so if it already appears anywhere it is left in place and an agent's child can pick up the host's own `orca` binary — plus at `:12`, when both `PATH` and `Path` exist only `PATH` is rewritten. Separate defect, separate PR; noting it here only so it doesn't get conflated with this change.

Thanks — this one was clean.

PR #11778, opened 2026-07-31
fix(terminal): stop IME newline defer holding Shift+Enter with no composition (#10203)

### LINKED ISSUE #10203: [Bug]: Pi agent input freezes briefly on Shift+Enter newline (other CLI agents unaffected)
### Operating system
Windows 11

### Orca version
4.152.0

### Pi agent version
0.81.1

### Summary
When using the **Pi** agent inside Orca, pressing **Shift+Enter** to insert a newline (multiline input) causes the input line / cursor to **freeze briefly ** before the newline is accepted. The same Shift+Enter multiline-newline action works smoothly in the other officially supported CLI agents (Codex, Claude Code, OpenCode) run inside the same Orca build — **only Pi shows this stutter**. That isolation strongly suggests Pi-specific handling in Orca's PTY / input layer rather than a general terminal regression.

### Steps to reproduce
1. On Windows 11, open Orca and start a **Pi** agent session in a terminal / worktree.
2. Focus the Pi input prompt.
3. Press **Shift+Enter** to insert a newline (build a multi-line message); repeat a few times.
4. Observe the input line / cursor: it briefly hangs / freezes on each Shift+Enter.

### Expected behavior
Shift+Enter inserts a newline immediately and the cursor stays responsive, consistent with how Codex / Claude Code / OpenCode behave in Orca.

### Actual behavior
The input line / cursor freezes for a moment on each Shift+Enter newline. The rest of the Orca UI stays responsive — only the input area stutters. No such stutter occurs with other CLI agents in the same Orca environment.

### Notes that may help triage
- The freeze is **local to the input line / cursor**, not a full-app hang.
- Reproduced **only with the Pi agent**; Codex / Claude Code / OpenCode do not show it in the same build/environment.
- The same **Pi** agent used directly inside a native **PowerShell** terminal has **no such newline stutter** — multiline input works fine there. This points the cause at Orca's input / PTY handling for Pi specifically, not at Pi itself nor the PowerShell host.
- Likely in Orca's input / PTY handling specific to how it detects or delivers keystrokes for Pi (e.g. `promptInjectionMode` / newline detection), since Pi is one of the four named supported agents in the README ("Run Codex, ClaudeCode, OpenCode or Pi side-by-side").

### Related
- Pi is an officially supported agent per the Orca README.

itisvincent:
## Summary

Fixes the "Pi input freezes briefly on Shift+Enter" symptom from #10203 — and more precisely, the freeze anyone sees on Windows when comparing Shift+Enter (freezes ~200ms before the newline lands) vs Alt+J (smooth).

### Root cause

On Windows, Shift+Enter is a *modified-Enter chord*. `keyboard-handlers.ts` routes modified Enter through the IME deferred-newline path:

\`\`\`ts
if ((e.isComposing || hasPendingImeComposition) && (e.key === 'Enter' || imeProcessEnter)) {
  ...
  deferredNewlineSender.defer(e, pane.terminal.element, sendResolvedInput)
  return
}
\`\`\`

`defer()` calls \`sendTerminalInputAfterComposition\`, which **always** armed a 200ms fallback timer and waited for \`compositionend\` / a captured xterm transaction end — even when **no IME composition was actually pending**:

\`\`\`ts
const fallbackTimer = window.setTimeout(finish, fallbackMs)   // 200ms
\`\`\`

When no compositionend arrives (the common case: IME idle, or no IME), the newline is held for the full \`TERMINAL_IME_DEFERRED_NEWLINE_FALLBACK_MS = 200\` — the visible "brief freeze before the newline is accepted."

Alt+J is **not** an Enter chord (\`e.key !== 'Enter'\`), so it never enters this defer and is sent immediately — which is exactly why Alt+J is smooth and Shift+Enter is not.

### Fix

Only hold the newline while a composition is **genuinely pending** (\`hasPendingTerminalImeComposition\`). If nothing is pending, send on the next macrotask (matching the existing no-terminal-element path). Genuine pending compositions still wait for \`compositionend\` / the captured xterm transaction to settle, so IME correctness is preserved.

\`\`\`ts
if (!hasPendingTerminalImeComposition(terminalElement)) {
  window.setTimeout(send, 0)
  return
}
\`\`\`

## Test

New regression case in \`terminal-ime-deferred-newline.test.ts\`: with no pending composition, \`sendTerminalInputAfterComposition\` sends on the next macrotask instead of waiting for the 200ms fallback. This test **fails on main** (\`expected 1, got 0\` after \`advanceTimersByTime(1)\`) and passes with the fix.

All 33 existing \`terminal-ime-deferred-newline\` tests still pass, including the overlapping-transaction and compositionend-flush cases.

## Verification

- \`pnpm typecheck\` — clean.
- \`pnpm exec vitest run terminal-ime-deferred-newline.test.ts\` — 33/33 pass (incl. new regression test, which fails on \`main\`).
- \`pnpm exec oxlint\` + \`pnpm exec oxfmt --check\` on changed files — clean.
- (\`keyboard-handlers-ime.test.tsx\` fails to load on this host due to a pre-existing \`@/lib/...\` alias resolution issue that also fails on \`main\`; unrelated to this change.)

## Cross-platform / compatibility notes

- Windows-only in effect (the defer is gated \`isWindows\` in \`keyboard-handlers.ts\`); macOS/Linux paths untouched.
- Agent-agnostic: applies to any agent whose Shift+Enter enters the defer (Pi, and any other modified-Enter user). This is the likely root cause of the Pi-specific freeze reported in #10203; the "only Pi" framing there is most likely because Pi is where multiline Shift+Enter input is most used.
- No UI/visual change. No perf risk — it removes a 200ms wait, and the pending-composition check is a \`WeakMap\` lookup.
- IME correctness preserved: real pending compositions still defer until \`compositionend\`/transaction end.

## Reviewer checklist

- [x] Cross-platform: Windows fix; macOS/Linux untouched
- [x] SSH/remote/local: input path is renderer-local; unaffected by transport
- [x] Agent/integration: agent-agnostic; no per-agent config changed
- [x] Performance: removes a 200ms hold; check is O(1)
- [x] Security: no new surface

Closes #10203

X handle: @itisvincent

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
@@ -47,6 +47,16 @@ describe('sendTerminalInputAfterComposition', () => {
     expect(send).toHaveBeenCalledTimes(1)
   })
 
+  it('sends on the next macrotask when no composition is pending instead of waiting for the fallback (#10203)', () => {
+    const el = document.createElement('div')
+    const send = vi.fn()
+
+    sendTerminalInputAfterComposition(el, send)
+    // Why: with no pending composition the newline must not be held for the 200ms fallback.
+    vi.advanceTimersByTime(1)
+    expect(send).toHaveBeenCalledTimes(1)
+  })
+
   it('finishes from the captured xterm transaction when deferral starts after compositionend', () => {
     const el = document.createElement('div')
     const send = vi.fn()
--- a/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.ts
@@ -15,6 +15,14 @@ export function sendTerminalInputAfterComposition(
     return
   }
 
+  // Why: only hold the newline while an IME composition is genuinely pending; otherwise the
+  // 200ms fallback delays every modified-Enter (Shift+Enter newline) on Windows even when no
+  // composition is active, surfacing as a brief input freeze vs non-Enter chords like Alt+J (#10203).
+  if (!hasPendingTerminalImeComposition(terminalElement)) {
+    window.setTimeout(send, 0)
+    return
+  }
+
   const fallbackMs = options?.fallbackMs ?? TERMINAL_IME_DEFERRED_NEWLINE_FALLBACK_MS
   let done = false
 


itisvincent:
## Closing — the IME deferred-newline path is NOT the cause

Runtime diagnostic on Windows (v1.4.167 + a `console.warn` in the Shift+Enter policy) confirmed:

- **No `[S+Enter DEFER]`** appeared in the DevTools console — the IME deferred-newline path (`sendTerminalInputAfterComposition`) was NOT firing.
- The freeze was **CSI-u decode cost** (`\x1b[13;2u` is slow for Pi to parse), not a 200ms IME delay.
- The encoding flickered between `csi-u` (newline with freeze) and `alt-enter` (submit) — a `routingTrusted` gate issue, not an IME issue.

The correct fix is in #12299 (LF encoding + launch-agent fallback), which:
- Sends `\n` (1 byte, Pi binds as Ctrl+J newline) instead of CSI-u — no decode freeze.
- Uses the launch agent as fallback when `routingTrusted` flickers — no submit.

This (#11778) targeted the wrong layer. Closing.

PR #11795, opened 2026-07-31
feat(notch): macOS agent-status indicator pinned to the notch

tomi204:
## Summary

Adds a **macOS agent-status indicator pinned to the notch**: live per-lane counts beside the camera cutout, expanding on click into a session list whose rows focus the matching pane. It keeps updating while the Orca window is closed — that is the state it exists for.

**On by default**, dismissable at **Settings → Appearance → Advanced → Show Agent Status at the Notch** (macOS only). This changed late in review — see Notes.

**Design**: main owns the data and the geometry; the notch renderer holds no derived state. Status comes from `agentHookServer`'s additive `subscribeEnrichedStatus` / `subscribeStatusCleared` rather than the `agentStatus:*` IPC, which targets the main window's `webContents` and detaches on close — a surface that outlives the window needs its own subscription. Lane vocabulary reuses the dashboard's `bucketForState` mapping so the notch and dashboard can never report different numbers for the same agents, and stale non-`done` rows decay through the shared `isFreshNonDoneAgentStatus` rule rather than a second definition of "silent".

Five constraints were established empirically on real hardware, each documented at its call site:

| Finding | Consequence |
|---|---|
| Electron clamps a window to the work area unless `enableLargerThanScreen` — `y=0` comes back as `y=<menu bar height>` | That single option is what makes a notch-hugging surface possible at all (verified across 6 window configurations) |
| `Display` exposes no safe-area data, and `env(safe-area-inset-top)` reports `0px` even at `y=0` over the cutout | Notch width is unobtainable from Electron, hence the Swift helper |
| Destroying a *visible* transparent panel kills the process (exit 0, no stderr) | Teardown hides first |
| An always-on-top window suppresses `window-all-closed`, which Orca uses to complete a deferred Cmd+Q | Torn down on `before-quit`, restored if the quit turns out to have been vetoed |
| An Electron window captures clicks anywhere in its bounds regardless of transparency | The window ignores the mouse by default; painted regions opt in, so the transparent gutters cannot swallow clicks on the system menu bar |

Also fixes a **pre-existing defect** this surface would have tripped: `ipc/notifications.ts` and `star-nag/service.ts` resolved "the app window" as the first non-destroyed `BrowserWindow`, which a permanently-visible chrome window would have answered on the app's behalf — suppressing tray attention and inverting notification focus gating. Now routed through `findAppWindow()`, which excludes chrome windows and is behaviour-preserving today.

## Screenshots

> **Note for reviewers:** images are captured but not yet attached to this PR body — the author will add them. Both are renders of the built bundle at real hardware geometry (14"/16" MacBook Pro, cutout measured at 771→956, `safeAreaTop` 32).

- **Collapsed** — braille spinner + count, green finished count on the left wing; red needs-input count on the right wing; camera cutout painted but unoccupied between them.
- **Expanded** — the bar slides to `layout.leadingOffset` to stay over the cutout while the window grows to the full panel; card hangs below with per-lane dots, `state · agent` subtitles and right-aligned elapsed times.

## Testing

- [x] `pnpm lint` — clean, including the max-lines ratchet and all three localization gates
- [x] `pnpm typecheck` — clean (node, cli, web)
- [x] `pnpm test` — notch/settings suites green. **3 pre-existing failures** in `src/main/providers/local-pty-shell-ready.test.ts` are unrelated to this branch; verified they fail identically on a clean tree (bash 3.2 on macOS vs the test's expectation)
- [x] `pnpm build` — passes, and now builds the universal `orca-screen-geometry` helper (x86_64 + arm64) via the new wiring
- [x] Added tests — ~150 new, colocated: summary math and the freshness/acknowledgement rules, wing geometry and the zero-count-claims-no-slot rule, window rects, row/label building incl. folder workspaces and SSH, expansion timings, quit teardown ordering, quit-restore decision, the `getAllWindows()` regression, and the settings toggle with its platform gate

Beyond unit tests, the feature was verified against a running app: the built renderer placed at real notch geometry, live agent hooks POSTed through the real `agentHookServer` endpoint driving the counts, the settings toggle driven over CDP creating and destroying the window, and a real quit confirmed to exit cleanly.

## AI Review Report

Two adversarial multi-agent reviews were run (31 and 12 agents), each finding then handed to an independent skeptic instructed to refute by default. **33 findings raised, 14 survived refutation, all fixed:**

- **Click-through (high)** — the expanded window spans the full panel width while only the bar and card are painted, at `NSStatusWindowLevel` (25), *above* `NSMainMenuWindowLevel` (24). Its transparent gutters silently ate clicks on the system menu bar. Fixed with `setIgnoreMouseEvents(true, { forward: true })` plus per-region opt-in; verified end to end.
- **IPC handler leak (high)** — `registerNotchHandlers` runs inside `openMainWindow()`, which fires on startup, dock re-activation *and* a notch row click, stacking listeners. Now removes-then-registers, matching `ipc/dashboard-popout.ts`.
- **Phantom counts (medium)** — no freshness gate meant a SIGKILLed agent counted as `working` forever, including after a cold start from a 7-day-old disk cache. Now decays via the shared rule, with a 60s unref'd re-tick.
- **Stale click targets (medium)** — `summariesEqual` ignored identity fields, so a re-attributed pane published nothing and its row kept routing to the old pane.
- **Vetoed-quit hole (medium)** — a cancelled quit destroyed the bar with no recovery for the session while the settings switch still reported ON. Fixed via a testable restore predicate.
- **Opaque background** — `body { bg-background }` defeated `transparent: true`, invisible collapsed but an opaque slab expanded.
- **Untested builds** — `build:mac` never compiled the geometry helper, and electron-builder only *warns* on a missing `extraResources` source, so local/QA builds silently shipped the pill fallback. Now wired into both mac scripts and `pnpm dev`, with an `afterPack` throw.
- **Grammar in a11y text** — `"{{count}} needs attention"` read as "2 needs attention", the only text a screen reader gets for that lane.

**Cross-platform** was explicitly reviewed. The feature is macOS-only and gated three ways: every `src/shared/notch/*` module is Electron-free and runs in vitest on all platforms; `createNotchWindow()` returns `null` off darwin (the `createSystemTray` pattern) with no module-level `screen` access; the settings row and its search entries are hidden unless `getRendererAppPlatform() === 'darwin' && !isWebClientLocation()`, so a Mac-reporting web client does not advertise it. `notch.html` builds everywhere and is never loaded off darwin. SSH-ingested panes and folder workspaces are both handled and covered by tests. No shortcuts or shell behaviour are touched; all paths use `path.join`.

An independent judge panel also argued the surface would be better as an `NSStatusItem` than a second `BrowserWindow` — the notch renderer modulepreloads ~4.1 MB to paint three integers. That is a legitimate open question and I am happy to take that direction if maintainers prefer it; see Notes.

## Security Audit

- **IPC trust** — every notch channel that mutates window state (`setExpanded`, `setInteractive`, `focusPane`) is origin-checked with `isNotchRenderer(event.sender)`. `focusPane` args are validated field by field with length caps before use; `panesAcknowledged` is sanitized and bounded at 512 keys.
- **Untrusted input** — the Swift helper's stdout is parsed defensively: malformed rows are dropped rather than producing NaN geometry, and a cutout is only accepted with both edges present and positive width. A missing or failing helper degrades to the pill presentation, never a crash.
- **Command execution** — the helper is spawned with `execFile` (no shell) from an absolute path, with a 3s timeout, resolved only from the packaged `Resources/bin` or the repo build output. It takes no arguments and no user input.
- **Renderer isolation** — the notch window runs `sandbox: true`, `webviewTag: false`, its own session partition, `installPrivilegedWindowNavigationPolicy`, and deny-all permission handlers (isolated sessions do not inherit the main session's policy).
- **Secrets / auth / dependencies** — none touched. No new runtime dependencies.

Follow-up: the acknowledgement channel is not origin-checked (any renderer could ack a pane). Impact is limited to clearing a green count, but it is the one asymmetry versus the other three channels.

## Notes

**Default changed after review.** Both bot reviews signed off on this specifically as an off-by-default feature. It now defaults on, deliberately: an indicator only helps if it is already there when an agent blocks, and every defect found while it was optional — the menu-bar click-through, the Cmd+Tab demotion, phantom counts from a stale cache, the reveal race — is now fixed rather than merely unreachable. Maintainers should weigh that judgement; reverting it is a two-line change to `constants.ts` and the `persistence.ts` hydration rule.

**Later fixes, after the review rounds below:**

1. **Cmd+Tab demotion (serious).** `setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })` makes Electron demote the whole app to `NSApplicationActivationPolicyAccessory`, silently removing Orca from the Dock and Cmd+Tab while the app keeps running. Measured one variant per process, since the policy latches: with the option `UIElement`, without it `Foreground`; `type:'panel'`, `focusable:false` and `skipTaskbar` are each innocent. The bar now follows Spaces but not fullscreen apps, which is where macOS hides the menu bar anyway.
2. **Hover-to-expand removed.** A dwell timer that opens a panel under the pointer is easy to trigger while crossing the bar toward the menu bar; it needed a movement gate and two timers to be tolerable. Click-only now — deletes `useNotchExpansion` and its test (~235 lines).
3. **Panel auto-closes** on `did-resign-active`, `browser-window-focus`, and active tab/worktree change. The window is `focusable: false` so it never blurs; the third signal is the only one that fires when the main window already has focus and only the tab changes.
4. **Minimal rows** — one line of state dot, provider logo, workspace, elapsed. Row height 52 → 30, panel width 800 → 420. Reuses the shared `AgentIcon`, so the notch and sidebar cannot disagree about what a Codex row looks like.
5. **Reveal handshake landed** (item 1 below is now fixed) via the `rendererReady` pattern `AutomationService` already uses.

**Known issues that remain:**

1. **The notch renderer's labels do not actually localize.** `notch.tsx` never hydrates the store, so `uiLanguage` stays null and strings fall back to English regardless of the user's setting.
2. **Pill presentation reserves 22pt of insets the renderer is never sent** (`expandedTopGap` + `expandedHeaderTopPadding` are summed into the window height but absent from `NotchPaintMetrics`), and `cornerStyle` / `topShoulderRadius` are serialized on every snapshot and read by nothing.
3. **Rows carry no git branch.** Main holds no branch/head state and an async git call on the repaint path would stall repaints, so rows show workspace · state · agent · elapsed. `notch-row-label.ts` already takes the no-git-identity path, so a future main-side branch cache slots in without rework.
4. **Primary display only**; row actions are focus-only (rename/kill would need `setFocusable(true)`, which steals focus); expansion resizes in one step rather than animating, since a spring would mean per-frame `setBounds` from main on a permanently-open window.

**Platform-specific behaviour**: adds a second single-file Swift helper following the existing `notification-status-macos` pattern (build script → `build-native-for-platform` → `extraResources` → `afterPack` guard). Unlike the notification helper it needs no bundle identity, so it lives in `Resources/bin`. Windows and Linux are unaffected and the build stays green.

**Attribution**: the concept is a reimplementation of [Moonglade](https://github.com/ixjosemi/Moonglade) (MIT). No Swift, assets, shaders or prose were copied — the transferring constants are behavioural facts — so there is no licence obligation, but the layout rules originate there and I am happy to add a courtesy acknowledgement wherever maintainers prefer.


nwparker:
This one is too cool. Added it to the list of PRs we want to get in. Thanks for the patiences 🙏

Issue #3835, opened 2026-05-30
[Feature]: Sync worktree board status to linked Linear issue state (drag worktree → update Linear)


Jeet-Chugh:
### Problem or use case

When I create a worktree from a Linear issue, Orca already stores the link (`WorktreeMeta.linkedLinearIssue`) and seeds the branch/prompt from the issue. But once I start working, the **worktree board** (the kanban with `Todo / In progress / In review / Done` lanes) and the **Linear issue's status** drift apart.

Today, dragging a worktree card between board lanes only writes `workspaceStatus` to local Orca metadata — it does **not** touch the linked Linear issue. So I end up updating the status twice: once on Orca's board, and again in Linear (or via Linear's own board). For a Linear-driven workflow this is exactly the kind of bookkeeping the tool should remove. The worktree *is* the unit of work for that issue, so its board column is the most natural, lowest-friction place to express "this issue is now in progress / in review / done."

**Summary:** When a worktree is linked to a Linear issue, moving that worktree between Orca board statuses should automatically transition the underlying Linear issue's workflow state.

### Proposed solution

When a worktree has a non-null `linkedLinearIssue` and the user drags it between board lanes, also push the corresponding state to Linear:

- `todo` → Linear `unstarted` (or `backlog`) state
- `in-progress` → Linear `started` state
- `in-review` → Linear `started` "In Review"–type state (fall back to the team's review state by name, else nearest `started`)
- `completed` → Linear `completed` / `Done` state

The good news is that nearly all the machinery already exists in the repo — this is mostly a matter of wiring two existing systems together:

1. **The drop handler that needs to gain the behavior:** `buildWorkspaceKanbanSidebarDropUpdates()` in `src/renderer/src/components/sidebar/workspace-kanban-sidebar-drop.ts` currently produces only `{ workspaceStatus: args.status }` updates to `WorktreeMeta`. It has no notion of the linked Linear issue.

2. **The exact sync already exists for the other board:** the standalone Linear issue board in `src/renderer/src/components/TaskPage.tsx` already does this — on drag it resolves the team's workflow states and calls `linearUpdateIssue(settings, issue.id, { stateId: workflowState.id }, issue.workspaceId)` (via `runtime-linear-client.ts`), using the helper `findLinearWorkflowStateForStatus(states, targetState)`. The same pattern can be reused from the worktree board's drop path, keyed off `worktree.linkedLinearIssue` instead of a Linear-issue card.

[...]

PR #11798, opened 2026-07-31
feat(worktree): surface every PR a branch feeds, not just one

### LINKED ISSUE #9213: [Feature]: Show multiple MRs/PRs per worktree in the right panel (stacked layout)
### Problem or use case

Orca links **exactly one** MR/PR to a worktree — `linkedGitLabMR`, `linkedPR`, `linkedBitbucketPR`, etc. are all singular scalar fields, auto-detected from the branch, and the right-side panel shows that single MR. There is no `--pr`/`--mr` setter and no array form (`orca worktree set` only exposes `--issue` / `--linear-issue`).

But a single feature often produces **more than one MR**:

- The same branch/feature is split into several MRs (e.g. a backend MR + a frontend MR that ship together).
- A stack of MRs where B depends on A, C depends on B — I want to see the whole stack in one place, not one card per node.

Today I can't associate those MRs with a single worktree, and the panel can only surface one of them. So the panel under-represents what the worktree actually is.

### Proposed solution

- [ ] Allow a worktree to carry **more than one** linked MR/PR (make the linked-MR field a list, or add an explicit attach path).
- [ ] Render them in the right panel as a **stacked layout** — an ordered list of MR cards for the feature, each with its own status (open/merged/draft, checks, reviewers).
- [ ] Keep auto-detection for the branch's own MR, but let additional MRs be attached (CLI flag and/or UI), e.g. `orca worktree set --add-mr <url>` (repeatable).
- [ ] Provider-neutral: works for GitLab MRs and GitHub/Bitbucket/etc. PRs, not just one.

- Current workaround is worktree **lineage** (`orca worktree create --parent-worktree <selector>`), but that makes one card *per* MR rather than one panel aggregating the feature's MRs — different mental model from "this feature has these N MRs."
- The other workaround is attaching each MR URL to the shared Linear issue (`orca linear attach --current --url <mr-url>`), but those surface inside the Linear issue's attachments, not Orca's own MR panel, and aren't a stacked display.
- Related but distinct existing issues — this one is specifically about the **right panel aggregating and displaying multiple MRs for one worktree**, not restacking mechanics or cross-workspace dependency graphs:
  - #7915 — PR dependency awareness (stacked PRs / PR tree), read-only via external tools
  - #7918 — umbrella: multi-repo work driven by agent-managed worktrees and stacked PRs
  - #6368 — Branch visibility and controls for Graphite stacks
  - #8875 — Stacked PRs with stacked worktrees (closed, not planned)

EugeValeiras:
Closes #9213.

## The problem

A workspace can only surface one review, because the branch's own auto-detected PR *is* the model. Teams that ship one branch to several destinations therefore see a card that is quietly wrong — not empty, which would be obvious, but showing one of N with nothing to say the others exist.

My team runs trunk-based with release branches: one ticket's branch opens a PR to the repo's base, one to `stage`, and one to the active `RELEASE/*`. The card showed whichever one the branch lookup happened to resolve. The information that mattered — *is the fan-out complete?* — was exactly what was missing, and it looked like an answer rather than a gap.

The stacked-PR workflows in #9213 hit the same wall from the other side: there the reviews differ by head branch, here by base, but both need a workspace to hold more than one.

## The approach

Additive: `attachedReviews` on the workspace, alongside the auto-detected review rather than replacing it. Orca keeps resolving the branch's review exactly as before and keeps polling its checks; the attached ones are what the user asserts.

I measured the alternative first — making `linkedPR` a list touches 57 files and every card, sidebar and cache path that reads it. This is 20, and none of them change existing behaviour when the list is empty.

## What renders

With **one** review, nothing changes: same detailed section.

With **two or more**, they render as peers under an `N reviews` header, base ref first:

```
⇅ 2 reviews
  ⇅ PR #248 → stage                                    ↗
    fix(operations): quitar core.private.account (WOLF-1992) [stage]
    [State: Open] [Checks: Passing]

  ⇅ PR #247 → RELEASE/v1.14.0                          ↗
    fix(operations): quitar core.private.account (WOLF-1992) [v1.14.0]
```

Two choices worth calling out:

- **Base ref leads.** With several reviews off one branch the number alone does not say which is which; the destination does.
- **State and checks stay on the row Orca polled.** Making the rows uniform should not cost real information, so an absent badge means "not polled", not "not passing". Keeping the auto-detected one as the header instead would make the others look secondary when they are the same kind of thing.

## CLI

```bash
orca worktree set --worktree active --add-pr "https://github.com/acme/app/pull/294,https://github.com/acme/app/pull/295"
orca worktree set --worktree active --clear-prs
```

`--add-pr` also takes a JSON array when the caller knows more than the link carries:

```bash
orca worktree set --worktree active --add-pr '[
  {"url": "https://github.com/acme/app/pull/294", "baseRef": "RELEASE/v1.14.0", "title": "fix: thing [v1.14.0]"},
  {"url": "https://github.com/acme/app/pull/295", "baseRef": "stage", "title": "fix: thing [stage]"}
]'
```

Base ref and title cannot be derived from a PR link, and without them the list renders as bare numbers. The URL still decides provider and number, so an entry cannot claim to be a review its own link does not describe.

Calls are additive; `--clear-prs` drops them; both together replaces the list, which makes it idempotent for a script that re-attaches the currently-open PRs on every run.

## Notes for review

- Attached reviews are **not polled**. They are links the user asserted, so they carry no check state and cost no requests. Only the auto-detected review keeps its `hostedReviewCache` behaviour.
- Persisted input goes through `normalizeAttachedReviews`, which drops entries that could only render a dead row and caps the list at 50 as a corruption guard.
- The key is omitted from `mergeWorktree` when empty rather than sent as `[]` — that merge runs per worktree on every list payload.
- Providers other than GitHub are matched by URL shape rather than parsed in depth: number and host are all a row needs, and hard-coding each provider's path layout would rot.
- 25 tests across the model, the CLI flag parsing, the merge and the render selection.

Happy to split this into model+CLI and render if that reviews more easily, or to change the layout — I have no attachment to the specific rows, only to a workspace being able to hold more than one review.


EugeValeiras:
Pushed a rebase on current `main`, plus a fix that turned out to be the real reason this feature is needed.

## The branch lookup only ever sees one PR

`getRestPRForBranch` asked GitHub for `pulls?head=…&state=all&per_page=1` and took `list[0]`. GitHub returns them newest-first, so on a branch feeding several PRs the card showed the **most recent** one — which is often closed — while the one still open stayed invisible.

Real branch from our repos:

```
GitHub returns:                    Orca showed:
  #294 closed → development          #294 closed → development
  #293 closed → stage
  #292 OPEN   → RELEASE/v1.14.0        ← the only actionable one, hidden
  #291 closed → development
```

That is not missing information, it is wrong information: the card confidently points at a dead PR.

Fixed here by fetching the list and picking what the user can still act on — open first, then merged. Choosing a merged PR does not make it visible on its own; `hideMergedImplicitPR` still decides that, and I left a note in the code so the precedence is not "fixed" later by accident.

## What this changes about the proposal

The original framing was "let the user attach extra reviews". Having found the above, the better framing is:

- **Orca should surface every PR the branch feeds** — it already has them in the response, it was throwing them away.
- **`--add-pr` is for what the branch cannot know**: a review in another repo, from another branch, or on a fork.

I stopped short of populating `attachedReviews` from the lookup in this PR. Threading the full list through `getRestPRForBranch` → `resolveBranchLookup` → `getPRForBranch` also means `hostedReviewCache` starts holding N reviews instead of one, and that cache compares by equality to invalidate. It is a change in the shape of the model and I would rather agree on it here than smuggle it into a PR about something else. Happy to do it as a follow-up if maintainers prefer that shape.

Also worth noting for #3835 (sync board status to the linked Linear issue): that need is a special case of a more general one — running a command when a card changes column. I will open a separate issue for it rather than widen this PR.

EugeValeiras:
This now detects the reviews on its own, which changes what the PR is asking for. Running it locally against a real branch with three PRs surfaced two more bugs, both fixed here.

## Orca already had every PR and was dropping them

`getRestPRForBranch` asked for `per_page=1` and took `list[0]`. Every PR the branch feeds was one request away; the extras now travel as `siblings` on the lookup result and render without anything being attached.

That reframes the feature:

- **Detected automatically**: every PR the branch feeds. Nothing to attach, and the state stays fresh because the siblings come from the same request as the review Orca already polls.
- **`--add-pr`**: only for what the branch cannot know — a review in another repo, from a different branch, or on a fork.

It also removes a staleness problem the manual approach had: an attached review stores the state from whenever it was attached, and nothing refreshes it.

## The bug that made this invisible

`hostedReviewInfoFromGitHubPRInfo` is the only bridge between the lookup and everything the card renders, and it builds the object field by field. It copied neither `siblings` nor `baseRefName`. The lookup produced them correctly and the renderer never saw them.

Two reasons this passed CI:

1. **`PRInfo` did not declare `siblings`**, and the producer attaches it with a conditional spread — which does not trigger the excess-property check. The field travelled at runtime while the type never knew it existed. It is declared now, so the next person to break this gets a compile error.
2. **The mapper's test used `toMatchObject`**, which by definition cannot catch a missing field. The new test asserts with `toEqual`, and I verified it fails when the fix is reverted.

`baseRefName` rides along in the same fix: without it the auto-detected review was the only row with no destination, which is precisely the row that does have badges.

## Picking which PR represents the branch

With several PRs on one branch, GitHub returns them newest-first, so `list[0]` picked the newest even when closed. A real branch from our repos:

```
  #294 closed → development     ← what Orca showed
  #293 closed → stage
  #292 OPEN   → RELEASE/v1.14.0    ← the only actionable one
  #291 closed → development
```

Now the open one wins, then merged. Choosing a merged PR does not make it visible on its own — `hideMergedImplicitPR` still decides that, and there is a note in the code so the precedence is not "fixed" by accident later.

## Verification

8640 tests pass. Beyond the unit tests I ran this build for a day against ~15 workspaces across 20 repos, which is how the bridge bug surfaced at all: the model and the CLI were correct in isolation and the card still showed one review.

Also relevant to #3835 (sync board status to a linked Linear issue): that is a special case of running a command when a card changes column. Separate issue, not this PR.

PR #11346, opened 2026-07-29
fix: route server project adds to selected runtime

hilaily:
## Summary

  Fix adding a server-hosted project or folder from the Add Project dialog when a non-default runtime environment is selected.

  Previously, the selected runtime environment ID was not passed to `addRepoPath`. As a result, a path selected from a remote environment such as `box1` could be registered
  against the global/default environment instead.

  The server-path flow now explicitly passes the selected runtime environment ID, ensuring the project is created in the environment from which it was selected.

  ## Screenshots

  No visual change.

  ## Testing

  - [ ] `pnpm lint`
  - [x] `pnpm typecheck`
  - [ ] `pnpm test`
  - [x] `pnpm build`
  - [x] Added or updated high-quality tests that would catch regressions, or explained why tests were not needed

  Additional verification:

  - Targeted `useAddRepoServerPathFlow` Vitest test passed.
  - Updated the regression test to verify that `addRepoPath` receives the selected `runtimeEnvironmentId`.
  - Targeted formatting and lint checks passed with no warnings.
  - Full Linux build completed successfully.
  - Intel macOS build completed successfully using Swift 6.0.3 and the Xcode Command Line Tools.
  - Manually verified that adding `/home/ace/workrepo/zeal_platform` from `box1` now registers the project in the correct environment.

  The unchecked commands were not run as complete repository-wide suites. Their relevant targeted checks were run successfully.

  ## AI Review Report

  An AI-assisted review was performed over the complete diff.

  The review checked:

  - Whether the selected runtime environment is propagated through the Add Project server-path flow.
  - Whether both Git repository and plain-folder additions use the same corrected routing.
  - Whether React hook dependencies include the new environment input and avoid stale runtime selection.
  - Whether the change affects modal lifecycle, nested repository scanning, worktree loading, or project activation.
  - Whether existing callers remain type-safe with the optional `addRepoPath` options argument.
  - Whether the regression test would fail if the runtime environment ID were omitted again.

  No blocking issues were found. The main risk identified was accidentally falling back to the global runtime configuration when adding a server path. This was addressed by
  passing:

  `{ runtimeEnvironmentId: activeRuntimeEnvironmentId }`

  directly to `addRepoPath` and asserting it in the test.

  Cross-platform compatibility was explicitly reviewed for macOS, Linux, and Windows:

  - No shortcuts or keyboard handling changed.
  - No labels or visible UI text changed.
  - No shell commands or shell-specific behavior changed.
[...]

Issue #7983, opened 2026-07-09
[Bug]: Continuous full-untracked Git polling consumes CPU in large monorepos on v1.4.130


andrewli81:
### Operating system

macOS

### Orca version

1.4.130

### Details

**Short summary**

On a large monorepo, Orca continuously runs:

```text
git -c core.quotePath=false status --porcelain=v2 --branch --untracked-files=all
```

Each scan takes about 35–45 seconds and the next scan starts only a few seconds later. This keeps a Git process consuming CPU almost continuously while the workspace is idle.

This still reproduces on v1.4.130 after a clean restart with one Orca main process and a newly launched daemon. It is not a build and is independent of the **Keep local main up to date** setting, which was disabled.

**Environment**

- macOS 26.5.1
- Orca 1.4.130
- Apple Git 2.50.1
- Large monorepo with multiple worktrees
- No Bazel, compiler, or other build running
- Clean restart: the pre-update Orca daemon and its terminal sessions were terminated before testing

**Reproduction**

1. Add a large monorepo where a full `git status --untracked-files=all` takes tens of seconds.
2. Open a workspace for that repository.
3. Leave the workspace idle for at least two minutes.
4. Inspect Orca child processes or `~/Library/Application Support/orca/logs/main.trace.ndjson`.

**Observed**

In a clean single-instance two-minute sample:

```text
status --porcelain=v2 --branch --untracked-files=all   35.487s
diff                                                     5.661s
status --porcelain=v2 --branch --untracked-files=all   35.117s
```

The second status scan started about 8.7 seconds after the first completed. A live process snapshot showed the status child consuming 48–68% CPU for tens of seconds.

A longer trace from the same repository contains hundreds of these scans, averaging roughly 53 seconds, so the polling becomes effectively continuous.

**Expected**

Background source-control refresh should not continuously rescan every untracked path when a status operation takes longer than the polling interval. Possible mitigations include:

- coalescing and backing off based on the previous scan duration;
- invalidating status from filesystem events instead of fixed polling;
- avoiding `--untracked-files=all` for background checks;
- offering a per-repository large-repo mode or polling opt-out.

**Related work**

[#6381](https://github.com/stablyai/orca/issues/6381) and [#6413](https://github.com/stablyai/orca/pull/6413) reduced refresh fanout and upstream/push-target polling. The remaining issue here is the regular status poll: on a large repository it is not cheap and never reaches an idle state.

I can provide additional sanitized trace excerpts if useful.

[...]

PR #11820, opened 2026-07-31
fix(scm): keep git-status pacing across scheduler rebuilds

brennanb2025:
## Summary

Fixes the git-status refresh storm that makes the app lag badly (typing latency, renderer pegged at 100-160% CPU) when a remote/mobile runtime client is attached while agents are active.

Root cause: #11346 added `activeExecutionHostId` to the git-status scheduler effect's dependency array in `useGitStatusPolling`. The rebuild on host change is intentional (generation bump + in-flight abort), but the scheduler's pacing state (`lastRunEndedAt`, `lastRunDurationMs`) was closure-local, so every rebuild reset it to `-Infinity`. With a connected runtime client the resolved host id flaps (local vs `runtime:<env>` publications overwrite each other in the repoId-keyed store), each flap minted an unthrottled scheduler, and the pre-existing loop (git status rewrites `.git/worktrees/<n>/index` → 2s base-directory poller → status signal) then ran at the bare 125ms debounce instead of the 3s floor + slow-task backoff (#7983).

Measured on the affected machine (1.4.163-rc.3, live trace spans): git spawn rate jumped from 26-35/min (rc.1, same workload) to a sustained 150-283/min, full 5-command status cycles every 1-3s per busy worktree, renderer main thread ~73% busy.

Fix (two commits):
- `6eece0b41e`: pacing becomes a shareable object (`GitStatusRefreshPacing`) owned by the hook and keyed to the current worktree, surviving scheduler rebuilds. Host/push-target changes still dispose the scheduler, abort in-flight scans, and bump the generation — they just no longer forget how recently git ran. Worktree *switches* keep their instant refresh (fresh pacing per worktree).
- `1414b6a239`: hardening — run-id ordering so a stale run from a disposed scheduler that settles late cannot stamp pacing over a newer run's result, and the hook holds a single current-worktree pacing slot instead of an unbounded map.

## Empirical A/B validation

Beyond unit tests, the mechanism and fix were validated in a live dev app with real git subprocesses: two instances from pinned commits differing only by this fix (`cd2b62ed14` = main/broken vs `6eece0b41e` = main+fix), isolated `ORCA_DEV_USER_DATA_PATH` profiles, a fixture repo with 500ms file churn, Source Control active, then a scheduler-dep flip injected every 1s via the exposed store (same dependency array and rebuild path as the production host-id flap). Git spawns measured from each profile's `main.trace.ndjson`:

| | Baseline (churn only) | Under 1s dep-flapping |
|---|---|---|
| main `cd2b62ed14` (broken) | 54 execs/min, status every 3s | **60 status cycles/min at 1s gaps** (floor bypassed, 466 execs) |
| `6eece0b41e` (fix) | 54 execs/min, status every 3s | **20 status cycles/min at exact 3s gaps** (floor holds) |

Baselines are identical, so the fix changes nothing in normal operation; under the storm input, the broken build runs status at the flap rate while the fix build holds the floor.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` (targeted suites run locally; full suite left to CI)
- [ ] `pnpm build` (left to CI)

Targeted runs (all green): `git-status-refresh-scheduler.test.ts`, `useGitStatusPolling.rerender.test.ts`, `useGitStatusPolling.test.ts`.

New regression coverage:
- scheduler: shared pacing keeps the 3s activity floor across dispose/recreate
- scheduler: shared pacing keeps slow-scan backoff (30s scan still paces the next run after recreation)
- hook: execution-host flapping for the same worktree cannot bypass the floor (the storm shape)

Plus the live A/B above as end-to-end validation.

## AI Review Report

Reviewed the change against the scheduler's correctness contracts:
- Rebuild semantics preserved: generation bump, in-flight abort, and trailing-refresh behavior are untouched; only pacing continuity changes. The existing tests "aborts and rejects stale work when the execution host changes" and "triggers an immediate poll when the active worktree changes" pass unchanged.
- Out-of-order settle handled: a disposed scheduler's still-running scan that settles after a newer run cannot regress pacing (run-id ordering, second commit).
- Aborted-scan rule preserved: cancelled scans still record duration 0 so hide→reveal catch-up is not stretched.
- Pacing storage is a single current-worktree slot in a ref — no growth, released on unmount.
- `pacing` option is optional; the only factory consumer is `useGitStatusPolling`, and omitting it reproduces the old per-instance behavior.
Cross-platform: renderer-only TypeScript using timers and in-memory state — no shortcuts, labels, paths, shell, or Electron platform APIs touched; behavior identical on macOS/Linux/Windows. Applies equally to SSH-hosted and folder workspaces since the pacing key is the worktree id + path already used by the scheduler effect.

## Security Audit

No input handling, command execution, path construction, auth, secrets, dependency, or IPC changes. The pacing key is built from in-app worktree id/path values and stays in renderer memory; nothing is persisted or transmitted.

## Notes

- This clamps the amplifier regardless of what causes scheduler rebuilds. Follow-ups deliberately out of scope (separate PRs): (1) ~20 `setActiveWorktree` call sites clear `activeWorkspaceExecutionHostId` by omitting the new argument — making it sticky would reduce host-id flapping at the source; (2) `detectedWorktreesByRepo` is keyed by repoId alone, so local and runtime publications of the same repo overwrite each other.
- Related but separate: PR-status polling has no pre-spawn default-branch short-circuit and bypasses its negative cache on `active`/`post-push` reasons, which multiplied `gh` spawns during the storm (~50-100/min). Also a candidate for its own PR.
- Release impact: the regression is only in `v1.4.163-rc.3` (landed between rc.1 and rc.3); it is not in stable v1.4.162. Recommend landing this before promoting 1.4.163 to stable.

### DIFF
--- a/src/renderer/src/components/right-sidebar/git-status-refresh-scheduler.test.ts
+++ b/src/renderer/src/components/right-sidebar/git-status-refresh-scheduler.test.ts
@@ -312,6 +312,66 @@ describe('createGitStatusRefreshScheduler', () => {
     expect(task).toHaveBeenCalledTimes(2)
   })
 
+  it('does not let an older disposed run erase replacement backoff', async () => {
+    vi.useFakeTimers()
+    const calls: ReturnType<typeof deferred>[] = []
+    const task = vi.fn(() => {
+      const call = deferred()
+      calls.push(call)
+      return call.promise
+    })
+    const pacing = createGitStatusRefreshPacing()
+
+    const first = createScheduler(task, pacing)
+    first.resumeSafety()
+    first.dispose()
+
+    const second = createScheduler(task, pacing)
+    second.resumeSafety()
+    await vi.advanceTimersByTimeAsync(30_000)
+    calls[1]?.resolve()
+    await flushMicrotasks()
+
+    await vi.advanceTimersByTimeAsync(1)
+    calls[0]?.resolve()
+    await flushMicrotasks()
+    second.signal()
+
+    await vi.advanceTimersByTimeAsync(29_998)
+    expect(task).toHaveBeenCalledTimes(2)
+    await vi.advanceTimersByTimeAsync(1)
+    expect(task).toHaveBeenCalledTimes(3)
+  })
+
+  it('lets an aborted old run pace repeated rebuilds until a replacement finishes', async () => {
+    vi.useFakeTimers()
+    const calls: ReturnType<typeof deferred>[] = []
+    const task = vi.fn(() => {
+      const call = deferred()
+      calls.push(call)
+      return call.promise
+    })
+    const pacing = createGitStatusRefreshPacing()
+
+    const first = createScheduler(task, pacing)
+    first.resumeSafety()
+    first.dispose()
+
+    const second = createScheduler(task, pacing)
+    second.resumeSafety()
+    calls[0]?.resolve()
+    await flushMicrotasks()
+    second.dispose()
+
+    const third = createScheduler(task, pacing)
+    third.resumeSafety()
+    expect(task).toHaveBeenCalledTimes(2)
+    await vi.advanceTimersByTimeAsync(2999)
+    expect(task).toHaveBeenCalledTimes(2)
+    await vi.advanceTimersByTimeAsync(1)
+    expect(task).toHaveBeenCalledTimes(3)
+  })
+
   it('cleans up debounce and safety timers on dispose', async () => {
     vi.useFakeTimers()
     const task = vi.fn(async () => {})
--- a/src/renderer/src/components/right-sidebar/git-status-refresh-scheduler.test.ts
+++ b/src/renderer/src/components/right-sidebar/git-status-refresh-scheduler.test.ts
@@ -1,6 +1,8 @@
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import {
+  createGitStatusRefreshPacing,
   createGitStatusRefreshScheduler,
+  type GitStatusRefreshPacing,
   type GitStatusRefreshReason
 } from './git-status-refresh-scheduler'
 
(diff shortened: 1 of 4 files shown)


brennanb2025:
## Review + Electron QA

A same-model/same-reasoning review found and fixed two PR-scope issues in `1414b6a239`:

- Ordered shared pacing completions now prevent a disposed older scan from erasing a newer scan's slow-task backoff, while still letting an aborted old scan establish the anti-storm floor until its replacement finishes.
- Pacing is now O(1) and owned by the current worktree, so A→B→A switches keep their immediate refresh instead of inheriting A's old cooldown.

Verification:

- Reviewer: 52 related tests, full `pnpm lint`, `pnpm typecheck`, and max-lines ratchet passed; post-fix re-review returned clean.
- Independent run: 35 scheduler/hook tests passed.
- Electron: attached to the `mahimahi @ brennanb2025/fix-git-status-pacing-survives-scheduler-rebuild` dev instance, selected PR #11820 through the real workspace UI, opened Source Control, drove 12 same-worktree host changes, restored the local host, and typed through the real terminal surface. The rendered app remained responsive with zero console errors.
- CI: all required checks pass. Node 26 shard 15/16 initially hit the unrelated `activity-portal-readiness-loop.react185.test.tsx` readiness flake; that exact test passed locally under Node 26, the failed shard passed on rerun, and the final `verify` gate is green. No out-of-scope code change was made.

The deterministic call-count/timing tests are the primary regression evidence for this nonvisual scheduler fix; the uncropped screenshot is visible app proof after the live interaction.

Electron Source Control and responsive terminal after host-flap QA

Mobile emulator QA was not applicable: this PR changes only the desktop renderer scheduler/hook and tests, with no mobile UI or mobile package changes.

PR #11825, opened 2026-07-31
feat(mobile): tell Android users when their build is out of date

AboveColin:
## Summary

The mobile app is distributed as a **sideloaded APK on GitHub Releases** (`.github/workflows/mobile-android-release.yml`) — there is no store to announce a new build. An out-of-date app therefore fails silently, and it surfaces in the worst place: pairing rejects the desktop's offer with a bare "Not a valid pairing code", which reads like a bad paste rather than "this app is old".

That is how this was found: a stale APK refused a valid `orca://pair?code=…` offer, by QR *and* by paste, with no hint that the app was the problem.

**1. Pairing rejections name their reason.** `parsePairingInput` / `parsePairingUrl` return a typed `PairingRejection` instead of `null`. The offer version is read *before* schema validation, which is what makes an out-of-date app distinguishable from bad input, and the message names the side that is behind:

> This pairing code is version 3, but this app only supports version 2. Update Orca on your phone.

QR and deep-link paths stay strict about the `orca://pair` route; only a deliberate paste may be a bare base64 code. `parsePairingCode` / `decodePairingUrl` stay as wrappers, so existing callers are untouched.

**2. Startup update check (Android).** On launch the app resolves the newest `mobile-android-v` tag and offers the release page.

Source of truth, measured rather than assumed:
- `/releases/latest` is unusable — the Android lane publishes with `--prerelease --latest=false`, so it names a *desktop* build (`v1.4.162`).
- `/releases?per_page=30` would be a landmine: desktop publishes hourly, so Android entries sit deep and far apart. Measured 2026-08-01 on page 1 of 100: Android at indices 9, 26, 63, 64, 88, 89 — a gap of 37. A 30-item page can hold zero Android releases.
- Chosen: `git/matching-refs/tags/mobile-android-v`, which filters by tag prefix server-side. One request, every Android tag, no desktop noise.

Budgets: at most **one check per 24h**, **5s** abort, and every failure — offline, rate limited, malformed body — resolves silently with no prompt.

**3. "Check for updates" toggle on About.** Default-**on**, unlike push notifications: silence is the harmful default when nothing else can tell you. Off skips the request itself, not just the prompt.

## Screenshots

No screenshots. The two new surfaces are a bottom drawer ("Update available", *Get the APK* / *Later*) and a Switch row on About, both built from the existing `BottomDrawer` and the `Switch` row pattern already used on the Notifications screen, with existing theme tokens only. Happy to attach captures from an emulator if a reviewer wants them.

## Testing

- [x] `pnpm lint` — oxlint clean (`mobile/`)
- [x] `pnpm typecheck` — `tsc --noEmit` exit 0 (`mobile/`)
- [x] `pnpm test` — **380 files, 2829 passed**, 3 skipped (`mobile/`)
- [ ] `pnpm build` — not run; this PR touches only `mobile/`, which is built by the Expo/Gradle release workflow rather than the desktop build

New tests: release-tag parsing and numeric semver ordering; pagination (refs are oldest-first, so a full page must be followed); silent failure on rate limit, network error, malformed body and a stalled request; throttle window, backwards clock, opt-out and per-version dismissal; default-on preference and its opt-out round-trip; every pairing rejection reason; and a **parity test** pinning the mobile parser's accept/reject set to `src/shared/pairing.ts` across 29 inputs.

Also verified live against the real GitHub API: resolves `0.0.36`, silent on `0.0.36`, prompts on `0.0.30`, honors dismissal.

## AI Review Report

Reviewed with Claude. Risks checked: the update check's failure modes (offline, HTTP error, malformed body, hang, clock skew), whether an opt-out truly stops the network call, whether the pairing refactor changes what gets accepted, and the drawer's behavior when the release URL is missing.

What it flagged and what changed:
- **The version message pointed at the phone in both directions.** Real bug — an *older* offer means the desktop is behind. Now direction-aware, with a test for both directions.
- **`saveAutomaticUpdateCheckEnabled` had no try/catch** while being called with `void` — a failing write would surface as an unhandled rejection. Now swallowed like the neighbouring writes.
- **The About toggle rendered default-on before the stored value loaded**, flashing ON for opted-out users and letting a quick tap be reverted by the load. Now renders only once the value is known.
- **Storage key used dots** instead of the `orca:camelCaseKey` convention. Renamed.
- **A claim that the shared and mobile parsers had diverged.** Measured instead of assumed: across 29 inputs they agree everywhere except uppercase scheme/host, which **pre-dates this PR** (`main` already matched case-insensitively and had a test for it, because QR alphanumeric mode encodes uppercase). Pinned in the parity test rather than silently changed — see Notes.

Cross-platform: this PR touches only `mobile/`, so no desktop macOS/Linux/Windows surface, no shortcuts, labels, shell behavior, or Electron paths are involved. Within mobile, the platform split is explicit: the check and its toggle are gated on `Platform.OS === 'android'`, since iOS updates arrive through TestFlight. No `path` handling is introduced.

## Security Audit

- **Input handling.** The pairing parser is the security-relevant surface. It got *stricter*, not looser: the QR and deep-link paths now reject anything that is not the `orca://pair` route before decoding, so a bare base64 blob is only trusted from a deliberate paste. The parity test pins the accept/reject set against the shared implementation so a future edit cannot quietly widen it. All parsing stays inside try/catch with schema validation via `PairingOfferSchema`.
- **Secrets.** Rejection messages report only the two integer version numbers — never the pairing code, device token, or public key. No new logging of offer contents.
- **Network.** One unauthenticated GET to `api.github.com` over HTTPS, no credentials, no request body, no device identifier. The response is only read for `ref` strings matched against `mobile-android-v<semver>`; the opened URL is built from that validated tag, not from a server-supplied link, so a compromised or spoofed response cannot redirect the user to an arbitrary URL. Aborted after 5s.
- **Privacy.** This is new outbound traffic to a third party: GitHub sees the device IP roughly once a day. The About toggle stops the request entirely, and it is disclosed in the UI copy.
- **No** command execution, no filesystem paths, no IPC, and no new dependencies.

## Notes

- **Android only.** iOS is excluded at both the check and the toggle; TestFlight already notifies.
- **Pre-existing divergence, deliberately not fixed here.** The mobile parser accepts `ORCA://PAIR?code=…` while `src/shared/pairing.ts` rejects it (`URL.hostname` is case-sensitive for non-special schemes). Mobile is right — QR alphanumeric mode encodes uppercase — so the shared/web copy is the one that should probably change. That is out of scope for this PR; it is pinned in `pairing-shared-parity.test.ts` so it stays visible.
- **Docstring coverage.** A bot flagged 30% against an 80% threshold. Not addressed on purpose: `AGENTS.md` mandates concise, non-obvious comments only ("DO NOT be verbose, explain the obvious"). Adding docstrings to satisfy the ratio would contradict the repo's own style rule.


AboveColin:
Thanks both — four of five points were right and are fixed in 78dcbebc4. The fifth I measured, and it does not hold.

**Fixed**

- **Version message blamed the phone in both directions** (@coderabbitai, Major) — real bug, and the worst kind here since the whole point of the message is naming which side is behind. Now direction-aware: an older offer says "Update Orca on your computer". Test covers both directions.
- **`saveAutomaticUpdateCheckEnabled` missing try/catch** (@coderabbitai) — correct, About calls it with `void`, so a failing write was an unhandled rejection. Swallowed like the neighbouring writes.
- **Toggle initialized before the async load** (@greptile-apps) — correct on both counts, the flash *and* the tap-then-revert race. Now `boolean | null`, rendered only once the value is known, so no tap can land before the load resolves.
- **Storage key convention** (@greptile-apps) — renamed to `orca:updateCheckState`.

**Not reproduced: shared/mobile parser divergence**

@coderabbitai asked to apply the route check and pre-schema `v` rejection to `src/shared/pairing.ts` so both sides make the same decisions. They already do. The new checks are *reason-reporting*, not new refusals: the shared copy's `extractPairingCodeFromUrl` already rejected non-`pair` hosts, and `PairingOfferSchema` already rejected a wrong `v` — reading the version first only changes *which message* the user gets, not whether the offer is accepted.

Rather than assert that, I ran both parsers over 29 inputs (valid URL, bare code, hash form, wrong route, extra path, truncated code, junk base64, wrong `v`, missing fields, wrong types, non-object payloads). They agree on every one except a single case:

    ORCA://PAIR?code=…    mobile: accepted    shared: rejected

That one **pre-dates this PR**: `main`'s mobile parser already matched with `/i` and shipped a test for it (QR alphanumeric mode encodes uppercase), while the shared copy compares `URL.hostname` case-sensitively. So it is a real inconsistency, but not one this PR introduces, and mobile is arguably the correct side. I have pinned all of it in `pairing-shared-parity.test.ts` so the accept/reject sets stay locked together and that gap stays visible instead of being forgotten. Fixing the shared/web copy felt out of scope here — happy to do it in a follow-up if you'd rather.

**Not addressed: docstring coverage**

Deliberate. `AGENTS.md` requires concise, non-obvious comments only ("DO NOT be verbose, explain the obvious"). Writing docstrings to reach an 80% ratio would contradict the repo's own style rule, so I would rather leave the threshold unmet than the guideline broken.

PR description now follows the template.

PR #11827, opened 2026-07-31
Improve disconnected server recovery copy

brennanb2025:
## Summary

- Change the outage toast to say “Can’t reach <server>” without overclaiming the cause.
- Tell users to check that Orca is running on the server and that their network connection works.
- Rename the action from “Retry” to “Try again.”

## User-facing copy

Title: “Can’t reach QA Server”

Message: “Check that Orca is running on this server and that your network connection is working, then try again.”

Action: “Try again”

## Validation

- 25 focused runtime-status tests passed
- Web typecheck passed
- Focused lint and formatting passed
- Localization catalog and extraction checks passed
- Whitespace validation passed

### DIFF
--- a/src/renderer/src/i18n/locales/en.json
+++ b/src/renderer/src/i18n/locales/en.json
@@ -325,7 +325,10 @@
         },
         "runtime": {
           "status": {
-            "runtimeHostDisconnectedDescription": "Workspaces and terminals on this server are unavailable."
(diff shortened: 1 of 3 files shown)


brennanb2025:
### Electron QA

Validated the named-server outage toast in the intended PR worktree instance. Triggered a connected → unavailable transition for “QA Server” through the runtime-status store and confirmed the rendered title, recovery guidance, and “Try again” action. The backing store recorded the server as unreachable.

PR #11832, opened 2026-07-31
fix(runtime): bound persisted graph hydration for large profiles

OrcaWin:
## Summary

- scan persisted host sessions once and hydrate only worktrees with serve/SSH-owned PTY candidates
- treat removed folder workspaces as absent during read-only graph hydration and filter their renderer snapshots
- invalidate mobile-sync memoization on folder deletion and defer saved-runtime mirror subscriptions until the runtime is healthy
- preserve strict failures for session mutations plus existing SSH/serve and offscreen-headless hydration behavior

## Why

The affected Windows profile has 239 persisted workspace entries and 723 tabs, but no persisted serve/SSH-owned pane candidates. The old graph publication path performed about 86,150 tab checks by rescanning full sessions per graph workspace. The new path performs 723 checks (~119x fewer).

There are two related release regressions:

- #9994 introduced the repeated session scan and is in stable `v1.4.160` as well as rc.3.
- #11271 introduced an rc-only `folder_workspace_not_found` throw on stale folder snapshots. The affected profile has zero live folder workspaces and two stale `folder:*` session keys.

The installed `1.4.163-rc.3` main process is formally hung, runtime/computer-use bridge calls time out, Windows recorded `AppHangB1`, and terminal typing outside Orca remains normal. I recommend blocking rc.3 promotion and cutting another RC with this fix.

Full machine, release-ancestry, trace, profile, ruled-out-candidate, dump-location, and validation evidence is intentionally committed for review in [`docs/reference/diagnostics/2026-07-31-runtime-graph-lag-regression.md`](https://github.com/OrcaWin/orca/blob/OrcaWin/fix-runtime-graph-large-profile-lag/docs/reference/diagnostics/2026-07-31-runtime-graph-lag-regression.md). Per review request, that temporary file can be removed before merge. The dump itself is not committed.

## Validation

- `pnpm run typecheck:node`
- `pnpm run typecheck:web`
- targeted `oxlint` on all seven TypeScript files
- 2/2 direct main-process regressions
- 3/3 SSH/serve hydration-preservation tests
- 50/50 renderer graph/mirror tests
- commit hook: oxlint, React Doctor oxlint, oxfmt
- `git diff --check`

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -26944,6 +26944,86 @@ describe('OrcaRuntimeService', () => {
     expect(secondMerge.publicationEpoch.match(/:headless-merge:/g) ?? []).toHaveLength(1)
   })
 
+  it('keeps the graph ready when a mobile snapshot references a removed folder workspace', () => {
+    const runtime = new OrcaRuntimeService({
+      ...store,
+      getFolderWorkspaces: () => []
+    } as never)
+
+    expect(() =>
+      runtime.syncWindowGraph(1, {
+        tabs: [],
+        leaves: [],
+        mobileSessionTabs: [
+          {
+            worktree: 'folder:removed-folder',
+            publicationEpoch: 'stale-folder-publication',
+            snapshotVersion: 1,
+            activeGroupId: null,
+            activeTabId: null,
+            activeTabType: null,
+            tabs: []
+          }
+        ]
+      })
+    ).not.toThrow()
+    expect(runtime.getStatus().graphStatus).toBe('ready')
+  })
+
+  it('scans ordinary persisted sessions once instead of once per graph workspace', () => {
+    const tabsByWorktree = Object.fromEntries(
+      Array.from({ length: 100 }, (_, index) => [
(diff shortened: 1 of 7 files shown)


brennanb2025:
### Electron validation (8a0d7a348e)

Validated the pushed branch in an isolated Orca dev profile via CDP and the real Electron renderer/main-process boundary.

- Attached to the intended worktree instance and loaded the PR worktree through the real repository IPC.
- Workspace hydration reached ready state with zero renderer console errors.
- Injected a disposable persisted session matching the reported scale: 239 workspace entries and 717 ordinary tabs.
- The real runtime.syncWindowGraph IPC completed in 2.9 ms and returned graphStatus ready.
- Restored and flushed the disposable session immediately afterward.
- Exercised rendered sidebar controls after the graph publication; the window remained responsive.



Remaining gap: this macOS host currently has no PTY devices available, so a new terminal process could not be allocated. The screenshot proves the rendered app remained responsive after the large-profile graph publication, but it is not evidence of successful terminal typing. Live Windows input remains to be checked after the wedged remote instance can be safely restarted.

PR #10998, opened 2026-07-27
fix(agent-status): preserve Claude pending work

tianyilt:
Fixes #10997.

## What changed

- keep Claude `Stop` in `working` while a `shell` or `monitor` background task is still running
- preserve `waiting` across `Stop` for unresolved permission requests
- map plan-mode `Stop` to `waiting` until the plan is confirmed
- keep explicit interrupted Stops terminal (`done` + `interrupted`)
- add regression coverage for each state transition

## Before / after

The same sanitized Claude hook payload was posted to Orca 1.4.156 and to this branch.

| Before: incorrectly completed | After: remains working |
| --- | --- |
| Before: green completed indicator | After: yellow working indicator |

The installed build returned `state: "done"` while Claude reported `1 shell, 1 monitor still running`; this branch returns `state: "working"` for the identical payload.

## Why

Claude emits `Stop` at a lead-turn boundary, not necessarily at true task completion. The normalizer previously mapped that event directly to `done`, and its background-task parser discarded non-agent task types. Permission and plan waits could likewise be overwritten by the following Stop.

## Verification

- `pnpm exec vitest run src/shared/agent-hook-listener.test.ts src/shared/claude-subagent-roster.test.ts` — 155 passed
- `pnpm run typecheck:node`
- directed `oxlint` on the four changed TypeScript files
- `pnpm run check:max-lines-ratchet`
- `pnpm run build:electron-vite`
- manual stable-vs-local hook replay with the screenshots above

Issue #10997, opened 2026-07-27
[Bug]: Claude Stop marks pending monitors, shells, approvals, and plans as completed


tianyilt:
## Summary

Orca treats Claude's `Stop` hook as terminal completion even when the session still has pending work that requires Orca to keep it active:

* a running `background_tasks` entry with `type: "shell"`
* a running `background_tasks` entry with `type: "monitor"`
* an unresolved `PermissionRequest`
* a `Stop` emitted while `permission_mode` is `plan`

This produces a green completed indicator in the sidebar and `state: "done"` from `orca worktree ps`, even though the terminal is still monitoring work or waiting for the user.

## Before / after

The screenshots below use the same sanitized Claude hook payload. Before is Orca 1.4.156; after is a local build with the proposed fix.

| Before: incorrectly completed | After: remains working |
| -- | -- |
[...]

PR #8853, opened 2026-07-15
fix(sidebar): keep the working spinner while a Claude Monitor runs

nwparker:
## What & why

When a Claude agent ends its turn with a background **Monitor** still running, the left-sidebar agent card dropped to the static "done" dot instead of the working spinner — even though the session is mid-flight and will resume when the monitor fires. Reported from a real session: the footer read "1 monitor still running" while the agent's sidebar card showed the static done dot.

This makes the card show the working spinner while a Monitor is watching, reusing the exact mechanism that already keeps a card spinning while a subagent runs.

## Root cause

The `Stop` hook payload carries a `background_tasks` array. A Monitor is a **non-agent** background-task kind (Claude's task kinds are `subagent, workflow, shell, monitor, MCP task, teammate, …`). `readClaudeBackgroundAgentTasks` keeps only `subagent`/`teammate` entries and discards the rest, so the monitor never entered the roster — the "stay-working" gate in `agent-hook-listener.ts` then saw an empty roster and resolved the pane to `done`:

```
Stop hook (background_tasks: [{ type: "monitor", status: "running" }])
  -> readClaudeBackgroundAgentTasks() drops the monitor (agent-only filter)
  -> roster empty -> effectiveState = 'done'   (NOT 'working')
  -> activity summary: hasLiveDone -> resolveWorktreeStatus 'done'
  -> StatusIndicator renders the static emerald dot, no spinner
```

## The fix

- `claudeBackgroundTasksHaveRunningMonitor()` detects a running monitor in `background_tasks` (matched on the `monitor` type prefix, so `monitor_ws`/`monitor_mcp` variants gate too; shells/workflows/crons deliberately excluded).
- Tracked per-pane (`claudeRunningMonitorByPaneKey`), updated **only** from a present `background_tasks` list so an event without the field can't clear a monitor tracked by an earlier turn (mirrors how the roster is retained).
- OR'd into the same three `done -> working` gates that already hold a pane working while a subagent runs.

Everything
 downstream (activity summary -> status resolver -> `StatusIndicator`) already promotes a `working` pane into the spinner, so **there are no rendering changes**. It self-heals to `done` on the next `Stop` that no longer lists the monitor as running.

## Evidence

**Before/after (proves the fix is load-bearing).** With the monitor check temporarily neutered to a no-op, the 3 new integration tests — which drive the real `normalizeHookPayload` entry point — reproduce the bug exactly:

```
 ×  reports Stop as working while a Monitor background task is still running
[...]

PR #11838, opened 2026-07-31
fix(agent-status): preserve Claude background work

nwparker:
## Summary

Keep Claude Code's Orca status `working` while Claude still reports provider-owned background work, and settle to `done` only when the lead session reports that work is gone.

This fixes the observed mismatch where Claude's footer says `1 shell still running` while Orca's sidebar and `orca worktree ps` already say `done` and remove the spinner.

Addresses the background-task portion of #10997. Related prior attempts: #8853 and #10998.

## Reproduction and root cause

Reproduced before the change with:

- Orca `1.4.163`
- Claude Code `2.1.220`
- macOS, local worktree
- a background `sleep` that outlived the lead turn

Observed simultaneously:

```text
Claude footer: 1 shell still running
Claude title:  ✳ Run background sleep command
Orca agent row: state: "done"
```

The installed Claude binary emitted this exact sequence:

```json
{"event":"Stop","background_tasks":[{"id":"b8rs2wmxg","type":"shell","status":"running","description":"Sleep for 15 seconds","command":"sleep 15"}]}
{"event":"SubagentStop","agent_id":"a70fdf2986e38302b","background_tasks":[{"id":"b8rs2wmxg","type":"shell","status":"running","description":"Sleep for 15 seconds","command":"sleep 15"}]}
```

When the shell finished, Claude emitted its completion notification and then:

```json
{"event":"Stop","background_tasks":[]}
{"event":"SubagentStop","agent_id":"a8ab60ba5d4410c47","background_tasks":[]}
```

Orca previously extracted only `subagent` and `teammate` rows from `background_tasks`. It discarded the shell, stored the lead as `done`, and let the following child lifecycle event re-emit that cached state. A one-event special case was insufficient because the immediate `SubagentStop` could overwrite it.

Binary inspection of Claude Code `2.1.220` confirmed that the wire inventory includes running/pending work and currently maps these provider task families: `subagent`, `teammate`, `workflow`, `shell`, `monitor`, `MCP task`, `dream`, `auto-mode scan`, and `cloud session`.

## What changed

- Parse the complete Claude background-task inventory into two independent signals:
  - bounded visible `subagent`/`teammate` rows;
  - a pane-level boolean for any running non-agent task.
- Keep the lead status effectively `working` while that boolean is set, including across Claude's follow-up child lifecycle hooks.
- Give only lead-attributed inventories pane-level authority. Any hook carrying `agent_id`, plus `SubagentStart`, `SubagentStop`, and `TeammateIdle`, is child-owned and cannot assert, clear, interrupt, or fold lead background evidence.
- Treat unknown task types/statuses and partial/unclassifiable inventories conservatively:
  - future non-agent types and nonterminal statuses fail active;
  - malformed or capped agent inventories are incomplete and cannot reap live child rows;
  - explicit terminal states, including teammate `idle`, remain non-running.
- Preserve an authoritative value when `background_tasks` is absent, but clear it when the lead provides a present empty inventory.
- Carry the boolean through relay, SSH, and WSL envelopes with boolean validation at trust boundaries.
- Let reconnect replay seed a cold runtime, then make live observations authoritative so a later stale replay cannot re-arm the gate.
- Block Escape/Ctrl+C fallback inference while Claude-owned background work is live.
- Keep the field transport-only: it is excluded from renderer IPC and persisted status files.
- Move and clear the pane-scoped state with existing cache transfer, pane teardown, inferred interruption, and listener teardown paths.
- Preserve session-cron evidence across child lifecycle hooks, but let a modern authoritative lead Stop drain it when Claude omits an empty `session_crons` array.

## ELI5

Claude can say, “I’m done talking,” while a job it started is still running in the background. Orca used to hear only “I’m done” and turn off the working light.

Now Orca also checks Claude's attached job list. If a background job is still alive, the working light stays on. Only Claude's main session may update that shared job-list signal, so a child agent's bookkeeping message cannot accidentally turn the light off—or turn it on. When the main session later sends an empty list, Orca marks the work finished.

## Proof of fix

The production normalizer regression test replays the captured Claude `2.1.220` order:

```text
UserPromptSubmit                       -> working
Stop + running shell                  -> working
SubagentStop + same running shell     -> working
task completion notification          -> working
Stop + empty background_tasks         -> done
```

End-to-end transport tests additionally prove:

- a real local `POST /hook/claude` with a running shell blocks false Escape inference;
- live remote `true` blocks inference and live `false` restores it;
- cold-runtime replay can seed evidence, but replay cannot override a later live observation;
- relay and SSH session forwarding preserve the boolean;
- the boolean is absent from renderer-facing snapshots and persisted JSON.

Parser/lifecycle tests prove:

- work after the visible child-row cap is still detected;
- state follows pane authority transfer and clears on teardown;
- `SubagentStart`, `SubagentStop`, and `TeammateIdle` inventories cannot assert or retract lead evidence;
- even an unknown child carrying `is_interrupt` cannot clear lead evidence;
- lead empty inventories clear the gate, while absent or partial inventories retain safe state;
- pending/future statuses remain active, known terminal/idle statuses do not;
- malformed, untyped, or unparseable rows cannot incorrectly reap live child state;
- explicit lead interruption remains terminal where that compatibility path applies.

These tests are load-bearing: without the change, the captured post-`Stop`/`SubagentStop` state is `done`, matching the installed-app reproduction.

## Proof of no regression

Final coordinator verification at `8a922e3997`, rebased onto `origin/main` at `8ab85c9bfc`:

- Complete shared suite from the adversarial hardening cycle: **390 files passed, 3 skipped; 3,627 tests passed, 15 skipped**.
- Main server/relay/SSH integration suite from that cycle: **3 files passed; 280 tests passed**.
- Latest merge-conflict integration suite: **10 files passed; 505 tests passed**, covering listener state, inventories, main ingest, relay frame shedding/publication, SSH forwarding, WSL forwarding, and cleanup.
- Final Opus confirmation recorded **5,247 passing tests** and killed **8/8 targeted mutants** before the latest rebase; the rebased implementation preserves every reviewed behavior.
- `pnpm typecheck`: **passed** for node, CLI, and web TypeScript projects on the latest head.
- Repository oxlint, native focused audit, type-aware audit, and changed-code quality gate: **passed with zero findings**.
- `pnpm exec oxfmt --check` on all **12 changed files**: **passed**.
- `pnpm run check:max-lines-ratchet`: **passed**, no new bypasses.
- Reliability gates, React Doctor changed-line review, localization checks, feature-wall asset budget, macOS entitlements, the root-directory guard, and `git diff --check`: **passed**.
- `pnpm run build:electron-vite`: **passed** for main, preload, and renderer production bundles during the hardening cycle.
- Final `git range-diff` marks **12/15 commits as exact equivalents**. The other three intentionally integrate newly merged relay frame-shedding metadata/tests alongside this PR’s Claude background-work transport field; no PR behavior was dropped.

The only textual conflicts were in `server.ts`, `ssh-relay-session.ts`, and adjacent `server.test.ts` coverage. Resolution preserves both upstream `shedFields` restoration and this PR’s validated `claudeRunningNonAgentTask` forwarding, plus both upstream roster-shedding tests and the PR’s connection-cleanup test. No conflict markers or uncommitted changes remain.

 Performance evidence

Synthetic worst-case benchmark: 200 hook events, each carrying a 10,000-row `background_tasks` inventory. Timings are local wall-clock measurements and are intended to expose relative hot-path work, not claim production latency.

- Rebase merge before cleanup, with duplicate full scan: **58.55 ms**.
- Parser-only baseline: **41.99 ms**.
- After duplicate-scan removal: **45.68 ms**.
- `SubagentStop` lifecycle path before early return: **51.01 ms**.
- `SubagentStop` lifecycle path after early return: **0.29 ms median** across seven samples (**~176x faster; ~99.4% less time**).

The optimized lifecycle path is correct because `SubagentStart`, `SubagentStop`, and `TeammateIdle` do not own pane-wide inventory authority; they update only the bounded roster and reuse pane-cached lead evidence. The retained lead path still performs one complete scan so capped or future task kinds cannot be missed.

## Adversarial AI review report

Review was coordinated through Orca orchestration Run `run_04348f604e05`. Each round used a fresh `claude --model opus` session, reviewed the complete `origin/main...HEAD` diff, was review-only, and was required to provide concrete reproductions or an explicit `CLEAN` verdict.

1. **Round 1 — not clean:** found interruption state reopening, missing Escape/Ctrl+C protection, ambiguous child inventory ownership, and StopFailure asymmetry. Fixed lifecycle ordering, interruption handling, server inference, and cache representation.
2. **Round 2 — not clean:** found child permission waits hidden after interruption, resumed turns remaining interrupted, child inventories clearing lead state, and replay overriding live authority. Fixed precedence, bounded resume behavior, lead ownership, and replay suppression.
3. **Round 3 — not clean:** found child inventories able to assert pane work, case/whitespace-sensitive statuses, and a fail-inactive task-type allowlist. Switched to lead-only authority, normalized statuses/types, and made future non-agent kinds fail active.
4. **Round 4 — not clean:** found `TeammateIdle` bypassing the child check via `teammate_name`, insufficient remote-ingest coverage, and missing/non-string task types failing inactive. Moved lifecycle handling ahead of inventory authority, hardened parsing, and tested the public remote path in both directions.
5. **Round 5 — not clean:** found agent rows with future/omitted statuses being reaped, unknown-child roster folding, no cold-runtime replay recovery, and missing local HTTP/persistence/IPC coverage. Made agent inventories fail active/incomplete, aligned lead scoping, added replay-before-live precedence, and added end-to-end proof.
6. **Round 6 — not clean:** found unknown-child interrupted Stops clearing lead evidence and fully unclassifiable rows still permitting authoritative reap. Made every `agent_id` event child-owned and marked every unclassifiable inventory incomplete.
7. **Round 7 — `CLEAN`:** no actionable defects. Opus revalidated all previous areas, inspected the shipped Claude `2.1.220` binary for wire-schema ground truth, reran tests/typecheck/lint, and confirmed the worktree was unchanged.

Final Opus verdict:

> **CLEAN. No actionable defects.**

Non-blocking suggestions were limited to cleanup/documentation and future product-policy choices (dead condition members, long-lived background-task presentation, and possible SessionEnd handling); none changes the correctness or merge readiness of this fix.

## Fresh post-rebase Opus review

Orca orchestration Run `run_5b6490fd4e9a` coordinated eight fresh review-only rounds with `claude --model opus`. Every round re-read the complete relevant diff and had to return concrete actionable findings or an explicit clean verdict.

1. **Round 1:** eight findings; fixed the substantive lifecycle, ownership, and transport issues.
2. **Round 2:** three findings; fixed remote-evidence acceptance ordering, independent cron state, and ambiguous status handling.
3. **Round 3:** one low policy/test gap; pinned interruption semantics and hardened intake.
4. **Round 4:** no implementation defect at the reviewed head.
5. **Post-CI cleanup round:** found three low hardening/coverage gaps. Malformed entries now fail active, child inventory authority is non-vacuously tested, and omitted empty cron inventories drain safely.
6. **Hardening re-review:** all three fixes passed mutation checks; one remaining test-only guard gap was found and pinned.
7. **Clean confirmation:** **8/8 targeted mutants killed**, **5,247 tests passed**, all static gates passed, and no actionable finding remained.
8. **Final post-rebase confirmation — `CLEAN`:** verified all 14 commits as exact equivalents, proved the dropped lint patch is byte-identical upstream, checked the three new upstream commits for interaction, reran 337 focused tests and all static gates, and found no actionable issue.

Final Opus verdict at `6a55cd6fba`:

> **CLEAN with no actionable findings.**

The reviewer explicitly found no correctness, security, performance, lifecycle/memory, transport, or Windows/Linux/macOS/WSL/SSH/relay/mobile/folder-workspace/git-worktree regression.

## Independent review round at `0d9f04e4be`

A fresh adversarial pass over the full `origin/main...HEAD` diff found no correctness, lifecycle, transport, or security defect. Areas re-derived from source rather than from the log above: pane-key identity across `ingestRemote`'s alias resolution and `inferInterrupt`'s lookup; the accepted-evidence gate against every early return in `applyNormalizedStatus`; replay seeding against the relay's one-entry-per-pane cache; `movePaneScopedSetEntries` on plain (non-composite) pane keys; and the absence of any renderer-side optimistic interrupt path that could bypass the main-process gate.

Two maintainability findings were fixed:

- The Stop/StopFailure predicate was written inline five times and `stateName` was a bare alias of `reportedStateName`, with both names used interchangeably in one function. Collapsed to a single `isTurnBoundary` constant and one state name — behavior-preserving, confirmed by the full related suite.
- Only the *interrupted* `StopFailure` path was pinned. Added two tests: a non-interrupted `StopFailure` keeps gating on live background work, and interrupted state does not survive a mid-turn lead event with no prompt submit. Both were mutation-checked — narrowing `isTurnBoundary` to `Stop` kills the first, and letting `interrupted` escape the turn boundary kills the second.

Gates at this head: **1,822 related tests passed** (3 skipped), `pnpm typecheck` passed for node/CLI/web, oxlint and oxfmt clean on all changed files, and `check:max-lines-ratchet` reported no new bypasses.

## Cross-platform and workspace assessment

- **Windows / Linux / macOS / WSL:** the change is transport-agnostic TypeScript and adds no OS-specific path, shortcut, shell, native module, or Git behavior. Named-pipe and socket transports continue to carry ordinary JSON; the new field is an optional boolean.
- **SSH / relay:** ingress accepts the field only when its runtime type is boolean; cold replay may seed a restarted runtime, while any live observation wins over later stale replay. The focused SSH integration suite passes.
- **Mobile / renderer:** mobile and desktop both consume the same normalized `AgentStatusIpcPayload` from `getStatusSnapshot()`. The transport-only boolean is stripped before IPC, snapshots, and persistence, so no protocol or mobile schema change is required.
- **Folder workspaces / git worktrees:** evidence is keyed only by stable pane identity, not repository metadata. Move, pane-clear, listener-clear, and teardown paths are covered.
- **Memory / lifecycle:** storage uses two O(1) pane-key `Set`s, bounded by live panes and cleared through existing lifecycle cleanup. Visible child rows remain capped by `AGENT_STATUS_MAX_SUBAGENTS`.

No platform-specific fallback is required.

## Security audit

- No new command execution, process launch, filesystem access, IPC endpoint, authentication flow, secret, dependency, or external network behavior.
- Existing request-size and structural normalization limits still bound hook payload work.
- Relay/SSH ingress accepts the new field only when `typeof value === 'boolean'`.
- The derived boolean contains no user content and is stripped before renderer IPC and disk persistence.
- Memory remains bounded to one Set entry per live pane and existing child-row caps; move/clear/teardown paths are covered.
- A compromised relay already controls the pane status payload; this boolean does not expand that trust boundary.

## Notes and trade-offs

- Long-lived Claude-owned tasks such as monitors or dev servers intentionally keep the working indicator active. That is preferable to reporting completion while Claude itself reports live work.
- A missing inventory retains the last authoritative value for compatibility with older or partial hooks; a present empty lead inventory clears it.
- Existing 30-minute status freshness still bounds stale presentation.
- This PR intentionally does not broaden into the separate permission/plan-mode cases in #10997, which have different precedence and clearing semantics.
- No screenshot is included because there is no component or layout change; the exact before-state, raw provider payloads, and production-path after-state are recorded above.

## Final exact-head merge gate

- Merge candidate: `27062e709968c822cf5bd26a9884d6ef2c663b12` on `origin/main` `4f963fd279ceeed031b7c0495f62ba920b774588`.
- Final rebase was conflict-free. `git range-diff` marks all **16/16 commits as exact equivalents**; the three latest upstream commits touch zero PR files.
- Final Opus feedback loop found one local-HTTP acceptance-ordering gap. Commit `27062e7099` defers both Claude background-task and session-cron evidence until the normalized status is accepted.
- The new regression test is non-vacuous: copied onto the pre-fix parent it fails while the rejected status itself remains unchanged; on the fixed head it passes.
- Final exact-head Opus review via Orca orchestration Run `run_5b6490fd4e9a`: **CLEAN with no actionable findings**. It independently confirmed 16/16 range-diff equivalence, zero upstream overlap, 342/342 focused tests, node/CLI/web typecheck, oxlint, max-lines, and 61/61 reliability gates.
- Coordinator post-rebase proof: 305/305 focused listener/server/relay/SSH tests, full typecheck, lint, formatting, max-lines, and `git diff --check` passed.
- Final GitHub Actions run `30731878819`: every Node 24/26 shard, Git compatibility, Windows packaging, static analysis, typecheck, shell contracts, root guard, and aggregate `verify` passed; E2E was intentionally skipped because no changed E2E specs were detected. The latest Greptile review on the exact-equivalent pre-rebase head also passed.
- Cross-platform review remains clean for Windows, Linux, macOS, WSL, SSH/relay, mobile, folder workspaces, and git worktrees. The final fix is synchronous in-memory Set gating and adds no platform-specific path, shell, Git, native-module, renderer, or mobile behavior.
- The final additional upstream commit, `fix(browser): remove unsafe window close bypass (#12040)`, changes only browser/window/session code and has zero overlap with the PR's agent-hook files; the last rebase remains 16/16 exact-equivalent.

### DIFF
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -3006,6 +3006,48 @@ describe('AgentHookServer listener replay', () => {
     }
   })
 
+  it('does not apply Claude background evidence from a rejected local status', async () => {
+    const server = new AgentHookServer()
+    await server.start({ env: 'production' })
+    try {
+      const env = server.buildPtyEnv()
+      const postClaudeHook = async (payload: Record<string, unknown>): Promise<void> => {
+        const response = await fetch(`http://127.0.0.1:${env.ORCA_AGENT_HOOK_PORT}/hook/claude`, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+            'X-Orca-Agent-Hook-Token': env.ORCA_AGENT_HOOK_TOKEN
+          },
+          body: JSON.stringify(buildBody(payload))
+        })
+        expect(response.status).toBe(204)
+      }
+
+      await postClaudeHook({
+        hook_event_name: 'PermissionRequest',
+        prompt: 'approve shell',
+        tool_name: 'Bash',
+        background_tasks: [{ id: 'shell-1', type: 'shell', status: 'running' }],
+        session_crons: [{ id: 'cron-1', status: 'running' }]
+      })
+      const waiting = server.getStatusSnapshot()[0]
+
+      await postClaudeHook({
+        hook_event_name: 'PreToolUse',
+        prompt: 'approve shell',
+        tool_name: 'OtherTool',
+        background_tasks: [],
+        session_crons: []
+      })
+
+      expect(server.getStatusSnapshot()[0]).toEqual(waiting)
+      expect(server._getStateForTests().claudeRunningNonAgentTaskPaneKeys.has(PANE)).toBe(true)
+      expect(server._getStateForTests().claudeActiveSessionCronPaneKeys.has(PANE)).toBe(true)
+    } finally {
+      server.stop()
+    }
+  })
+
   it('maps registered legacy numeric HTTP pane keys to stable pane keys', async () => {
     const server = new AgentHookServer()
     await server.start({ env: 'production' })
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -445,7 +445,7 @@ describe('AgentHookServer listener replay', () => {
           tabId: 'tab-1',
           worktreeId: 'wt-1',
           isReplay: true,
-          claudeRunningShellOrMonitor: true,
+          claudeRunningNonAgentTask: true,
           payload: { state: 'working', prompt: 'stale replay', agentType: 'claude' }
         },
         'conn-1'
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -607,6 +607,43 @@ describe('AgentHookServer listener replay', () => {
     }
   })
 
+  it('does not apply Claude background metadata from a rejected remote status', () => {
+    const server = new AgentHookServer()
+    server.ingestRemote(
+      {
+        paneKey: PANE,
+        hookEventName: 'PermissionRequest',
+        claudeRunningNonAgentTask: true,
+        payload: {
+          state: 'waiting',
+          prompt: 'approve shell',
+          agentType: 'claude',
+          toolName: 'Bash'
+        }
+      },
+      'conn-1'
+    )
+    const waiting = server.getStatusSnapshot()[0]
+
+    server.ingestRemote(
+      {
+        paneKey: PANE,
+        hookEventName: 'PreToolUse',
+        claudeRunningNonAgentTask: false,
+        payload: {
+          state: 'working',
+          prompt: 'approve shell',
+          agentType: 'claude',
+          toolName: 'OtherTool'
+        }
+      },
+      'conn-1'
+    )
+
+    expect(server.getStatusSnapshot()[0]).toEqual(waiting)
+    expect(server._getStateForTests().claudeRunningNonAgentTaskPaneKeys.has(PANE)).toBe(true)
+  })
+
   it('carries idle subagent rows through an inferred interrupt', () => {
     vi.useFakeTimers()
     vi.setSystemTime(1_000)
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -402,20 +402,17 @@ describe('AgentHookServer listener replay', () => {
     vi.setSystemTime(1_000)
     try {
       const server = new AgentHookServer()
-      const state = server._getStateForTests()
-      const send = (payload: Record<string, unknown>): void => {
-        const event = normalizeHookPayload(state, 'claude', buildBody(payload), 'production')
-        if (!event) {
-          throw new Error('normalizeHookPayload rejected a known-good Claude fixture')
-        }
-        server.ingestRemote(event, 'conn-1')
-      }
-      send({ hook_event_name: 'UserPromptSubmit', prompt: 'run in background' })
-      send({
-        hook_event_name: 'Stop',
-        background_tasks: [{ id: 'shell-1', type: 'shell', status: 'running' }]
-      })
-      const baseline = server.getStatusSnapshot()[0]
+      server.ingestRemote(
+        {
+          paneKey: PANE,
+          tabId: 'tab-1',
+          worktreeId: 'wt-1',
+          claudeRunningNonAgentTask: true,
+          payload: { state: 'working', prompt: 'run in background', agentType: 'claude' }
+        },
+        'conn-1'
+      )
+      let baseline = server.getStatusSnapshot()[0]
 
       vi.setSystemTime(1_500)
       expect(
@@ -429,6 +426,31 @@ describe('AgentHookServer listener replay', () => {
         })
       ).toBe(false)
       expect(server.getStatusSnapshot()[0]).toMatchObject({ state: 'working' })
+
+      vi.setSystemTime(2_000)
+      server.ingestRemote(
+        {
+          paneKey: PANE,
+          tabId: 'tab-1',
+          worktreeId: 'wt-1',
+          claudeRunningNonAgentTask: false,
+          payload: { state: 'working', prompt: 'run in background', agentType: 'claude' }
+        },
+        'conn-1'
+      )
+      baseline = server.getStatusSnapshot()[0]
+
+      vi.setSystemTime(2_500)
+      expect(
+        server.inferInterrupt({
+          paneKey: PANE,
+          baselineUpdatedAt: baseline.receivedAt,
+          baselineStateStartedAt: baseline.stateStartedAt,
+          baselinePrompt: 'run in background',
+          baselineAgentType: 'claude',
+          intent: 'plain-escape'
+        })
+      ).toBe(true)
     } finally {
       vi.useRealTimers()
     }
--- a/src/main/agent-hooks/server.test.ts
+++ b/src/main/agent-hooks/server.test.ts
@@ -56,6 +56,7 @@ const GOOD_PANE = makePaneKey('tab-good', LEAF_2)
 const OLD_PANE = makePaneKey('tab-old', LEAF_3)
 const FRESH_PANE = makePaneKey('tab-fresh', LEAF_4)
 const TAB_A_PANE = makePaneKey('tab-A', LEAF_5)
+const RUNNING_SHELL = { id: 'shell-1', type: 'shell', status: 'running' }
 
 type Body = {
   paneKey: string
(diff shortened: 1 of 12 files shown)


nwparker:
Trade-offs I confirmed but left alone

Two are worth your explicit sign-off, since they're policy rather than bugs:

1. Escape/Ctrl+C is now permanently blocked while evidence is live. If Claude exits without a draining Stop (orphaned background task), the user loses the manual escape    hatch and the pane spins until the 30-minute staleness bound. The working pre-PR; losing the override is new.
2. An explicit interrupt stays terminal even when Claude still reports a running shell — so the exact symptom in your repro can still occur via the Ctrl+C path. Deliberate and pinned by tests, but it is the one hole left in the fix.

Gates at this head: 1,822 related tests pass (3 skipped), pnpm typecheck lint + oxfmt clean on all changed files, max-lines ratchet reports no newbypasses. Full CI hasn't spun up yet beyond Greptile.

PR #11853, opened 2026-08-01
fix(mobile): mount host before opening tasks

brennanb2025:
## Summary

- Fix Home Tasks and provider shortcuts opening a blank Host screen on a cold nested navigator.
- Mount the selected host route first, then replace it with a concrete encoded Tasks path on the next animation frame.
- Preserve the selected task provider in the taskSource query parameter.

## Root cause

A cold deep push into the nested host navigator can resolve to its index route without the dynamic host ID. Live iOS simulator testing also showed that a dynamic route object can lose hostId during the second navigation. Mounting the host stack first and then using a concrete path preserves the selected host.

## Test plan

- Reproduced before the correction in an iPhone simulator against a connected Electron-backed LAN host: Home → Tasks opened the blank Host / Filter / Recent / Repo screen.
- Verified after the correction with a cold app start on the same host: Home → Tasks opened the Tasks screen with GitHub task controls.
- Mobile test suite: 377 files and 2,816 tests passed, with 3 existing skips.
- Typecheck passed.
- Lint passed.
- Regression coverage verifies concrete host-path encoding, provider selection, and cold navigator sequencing.

## Evidence

Blank Host screen reached from Home Tasks

Physical TestFlight verification requires a build containing this change.


brennanb2025:
QA evidence for `6797058950123692ed24b158b1855935b14d86a9`

- Electron: paired-host Orca Mobile view validated with `$electron`.
- Mobile: `QA-devqa-serial` on iOS 26.5, clean dev-client install, fresh Metro bundle, and cold starts.
- Cold-start Tasks card-body action opened the mounted Tasks screen with live GitHub issues.
- A separate cold start confirmed the GitHub provider shortcut opens the same Tasks screen.

Electron paired-host view

Mobile home

Mobile Tasks

PR #11856, opened 2026-08-01
Fix OpenCode continuation prompt delivery

mrcha033:
## Summary
- use OpenCode's supported `--prompt` launch path for normal Continue in New Session handoffs
- retain post-ready paste for prompts that exceed the Windows command budget
- track inline delivery until the OpenCode process is ready instead of reporting success synchronously
- add regression coverage for inline, tracked, and fallback delivery

## Why
The continuation flow always forced OpenCode through the readiness-gated paste path. The installed Orca build still uses that path, so context can fail to reach the new session. OpenCode supports `--prompt`, which avoids the startup paste race for normal-sized handoff prompts. Inline launches now wait for the launched process to become ready before reporting delivery success.

## Validation
- focused affected suite: 5 files, 102 tests passed
- `pnpm run typecheck:web`
- changed-file `oxlint --deny-warnings`
- `opencode --help` confirms `--prompt` support
- red-green regression: inline delivery promise absent before fix, present and readiness-tracked after fix

The full suite previously had 10 unrelated environment/pre-existing failures; details are documented in the task handoff.

### DIFF
--- a/src/renderer/src/lib/launch-agent-in-new-tab.test.ts
+++ b/src/renderer/src/lib/launch-agent-in-new-tab.test.ts
@@ -7,6 +7,7 @@ const mockSetActiveTabType = vi.fn()
 const mockSetTabBarOrder = vi.fn()
 const mockSetAgentStatus = vi.fn()
 const mockPasteDraftWhenAgentReady = vi.fn()
+const mockWaitForAgentReady = vi.fn()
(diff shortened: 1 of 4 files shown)


mrcha033:
Superseded by #11859. The replacement keeps transcript-derived context on the sanitized post-ready transport and fixes synthetic OpenCode SQLite transcript paths.

PR #11870, opened 2026-08-01
fix(terminal): preserve pending IME commit across composition restart

### LINKED ISSUE #11855: [Bug]: Previous IME text disappears while the next composition is active in the terminal
### Operating system

macOS

### Orca version

v1.4.163

#### Short summary

When typing consecutive Hangul syllables with the macOS Korean 2-Set IME in Orca's integrated terminal, previously committed text disappears while the next composition is active. The terminal is effectively one composition session behind.

This occurs in a plain shell as well as Codex CLI and Claude Code, so it is not agent-specific. The same input works correctly in iTerm2.

#### What happened?

The first syllable is initially visible. When composition of the next syllable begins, the previous syllable disappears and returns only after a later composition ends. It is not missing from the beginning; its commit reaches the PTY late.

#### How can we reproduce it?

1. Open Orca's integrated terminal on macOS.
2. Select the Korean 2-Set input source.
3. Type consecutive Hangul syllables such as `가나다` without switching input sources or pressing a separate commit key.
4. While the next syllable is still being composed, observe the previously completed syllable.

#### Expected behavior

The previous committed syllable remains visible while the next syllable is being composed.

#### Actual behavior

The previous syllable disappears and reappears only after the next composition ends. Terminal input stays one composition behind.

#### Technical investigation

This appears to be a regression in stablyai/orca#11293.

`CompositionHelper.compositionend()` schedules a zero-delay finalizer so Chromium can update the textarea before xterm reads the stable committed text. A synchronous next `compositionstart()` currently cancels that finalizer and provisionally ends and settles the previous session. Orca's composition route intentionally drops `dataPendingReconciliation`, so the previous text is not sent to the PTY until the following composition ends.

A deterministic probe against the currently patched xterm produces:

* While the second composition is active: `""`
* After the second composition ends: `"가나"`

Allowing the previous finalizer to reconcile through its captured composition route produces:

* While the second composition is active: `"가"`
* After the second composition ends: `"가나"`

Native reproduction is confirmed only with macOS Korean 2-Set. The affected code path handles composition events generically, so Japanese, Chinese, Vietnamese, and other composition-based IMEs may be affected if they emit the same rapid `compositionend` → `compositionstart` sequence, but that has not yet been confirmed on physical keyboards.

This may look related to stablyai/orca#9803, but that issue reports broader Windows input behavior predating stablyai/orca#11293; this report isolates a new rapid-composition-restart regression.

A focused regression test and fix are being prepared.

ethznn:
## Summary

- Keep the previous xterm composition finalizer alive when the next `compositionstart` event arrives, allowing the stabilized textarea value to flush the previous committed text immediately.
- Remove the provisional composition session-end path that caused Orca to discard the commit until a later composition.
- Add a regression test reproducing rapid Korean IME composition restarts, where the first syllable appeared initially and disappeared when the next syllable began composing.

Fixes #11855.

## Screenshots

No visual change.

The user-visible terminal input behavior and reproduction steps are documented in #11855. The corrected IME event sequence is covered by an automated regression test.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Full test result:

- 4,037 test files
- 42,661 tests passed
- 79 tests skipped

## AI Review Report

The review compared the previously patched and updated xterm packages directly. The only meaningful source change is in `CompositionHelper.ts`; the other xterm changes are its regenerated CJS/ESM bundles and source maps.

The review checked:

- Rapid IME restarts and overlapping composition session ordering.
- Korean final-consonant movement and post-composition textarea reconciliation.
- Local PTY, SSH, and reconnect behavior, including protection against sending delayed composition data to a replaced transport.
- Terminal disposal, blur, Escape, and pending-timer cleanup paths.
- Performance impact: no new timer or polling loop was introduced. The existing zero-delay finalizer is allowed to complete instead of being canceled.
- Cross-platform compatibility for macOS, Linux, and Windows. The change is in the shared DOM composition-event path and does not alter shortcuts, labels, paths, shells, or platform-specific Electron behavior.

The review flagged whitespace-only lines in the regenerated pnpm patch and the resulting lockfile hash mismatch. Both were corrected, followed by a frozen-lockfile install and fresh verification.

## Security Audit

No command execution, shell handling, filesystem path, authentication, secret, network, or IPC behavior was added or changed.

The existing xterm dependency version remains unchanged; only Orca's existing local xterm patch was updated. Composition data continues through the existing transaction/session route, including the current guards that reject delayed data after PTY replacement or SSH reconnection.

No additional security follow-up is required.

## Notes

- Reproduced with the macOS Korean 2-set IME.
- Reports in #11855 confirm that the same one-composition-behind behavior is cross-platform and not limited to the stock macOS Korean IME: Korean 2-Set reproduces on macOS with both Apple Korean 2-Set and Gureum, and on Windows 11; Japanese input reproduces on physical JIS keyboards on Windows 11 with Microsoft IME and Ubuntu 24.04 with fcitx5/Mozc. The visible artifact varies between hidden/disappearing preceding text and overlapping preedit text, while committing redraws the line correctly.
- The reports reproduce in plain shells, cmd, PowerShell, and Claude Code, and the Windows behavior is unchanged by switching between WebGL and DOM rendering or by changing fonts. One macOS report also observed the symptom in the session chat/prompt composer; this PR changes only the shared xterm terminal composition path, so that separate surface remains outside its verified scope. Verification of this fix on the additional Windows, Linux, and non-stock IME environments is still pending.
- This preserves the IME transaction routing introduced by #11293 while correcting the rapid-restart regression.
- X (Twitter): not provided


holdn2:
Heads up — I've opened #12120 as a draft that carries both of your commits here
(`38bb638` preserving the pending finalizer across a composition restart, `2a087c9`
clearing the canceled finalizer timer) with `--author` preserved, alongside the minimum
hunks from #11616, #11937 and #11896.

I applied your `config/patches/@xterm__xterm@6.1.0-beta.287.patch` and `pnpm-lock.yaml`
changes unmodified — the patch file in my branch is byte-identical to yours, including the
blank context lines that `git diff --check` complains about, because diverging from your
artifact seemed worse than a whitespace warning. `pnpm install --frozen-lockfile` is clean.

Your restart fix is what actually closes the "commits arrive a syllable late" symptom;
the keyboard-side fixes alone don't touch it.

Not meant to supersede your PR — maintainers may prefer to land it directly. Happy to
close mine on request, or to correct anything I've got wrong.

ethznn:
@holdn2 
Thank you for the update and for carefully preserving the original commits and attribution. I appreciate the integration work in #12120 and look forward to the maintainers’ guidance on how best to proceed with the two PRs.

holdn2:
Thank you — and likewise for the xterm work. Commits 5 and 6 here are yours unchanged, and
`config/patches/@xterm__xterm@6.1.0-beta.287.patch` is byte-identical to your artifact,
trailing context lines and all; diverging from it seemed worse than a `git diff --check`
warning. `pnpm install --frozen-lockfile` is clean and that suite is 49/49.

Agreed on waiting for maintainer guidance on how best to proceed with the two PRs. For what
it is worth, your restart fix is the only thing in this stack that closes the "commits arrive
a syllable late" half — the keyboard-side work does not touch it — so whichever branch lands,
that part should be yours.

nwparker:
Closing as superseded by #12278, which landed the pending-composition finalizer mechanism with regenerated xterm artifacts and cross-platform regression coverage. Thank you for the diagnosis and implementation work that informed the merged fix.

PR #11873, opened 2026-08-01
Feature/automation agent args 11043

hdsheena:
## Summary

Adds per-automation agent args so a scheduled automation can pass extra CLI flags (e.g. `--yolo`) to its agent **without** changing the global default args for that agent or project — exactly the request in #11043 ("I want my automation to run as yolo, but i strictly DONT want this as default").

New `agentArgs` field on the automation model, settable from:
- **GUI** — "Agent args" text input in the automation editor dialog (e.g. `--yolo --verbose`)
- **CLI** — `orca automations create/edit --agent-args "--yolo"` and visible in `orca automations show` (`agentArgs: <value> | none`)

When unset, existing behavior is preserved: automations keep the global default args (e.g. claude's `--dangerously-skip-permissions`). Only a non-empty value overrides.

## Screenshots

- **UI:** new "Agent args" input in the automation editor dialog — *screenshot to add*
- **CLI:** `agentArgs: none` / `agentArgs: --yolo --verbose` lines in `automations show`

## Testing

- [ ] `pnpm lint` — *could not run in sandbox*: `oxlint` crashes with a Rust allocator panic (reproduces on unmodified code; CI will run it)
- [x] `pnpm typecheck` — `typecheck:node` and `typecheck:cli` pass. `typecheck:web` was OOM-killed by the kernel in this 5 GB sandbox (reproduces on unmodified code — environment limit, not a code error)
- [x] `pnpm test` — 95 focused tests pass across 8 automation suites (launch, dispatch hook, CLI format, run identity, headless workspace create, runtime automations, schedule picker, service)
- [ ] `pnpm build` — *not run in sandbox*
- [x] Added high-quality regression tests (4 new):
  - `launch-agent-background-session.test.ts` — per-automation args override the global default (`claude '--yolo' '--verbose' 'run the automation'` + `launchConfig`), and `null`/empty fall back to the global defaults
  - `useAutomationDispatchEvents.test.ts` — `automation.agentArgs` is forwarded to the background session launch
  - `cli/format.test.ts` — `automations show` prints `agentArgs: none` / `agentArgs: --yolo --verbose`

## AI Review Report

AI review checked: the defaults-fallback contract, the full create→persist→dispatch→launch pipeline, RPC/Zod round-trip, i18n key registration, GUI draft wiring, CLI flag parsing, and multi-flag tokenization/quoting.

**Flagged & fixed:** the initial launch fallback treated a stored `null` `agentArgs` (legacy/empty automations) as an explicit empty override, silently dropping global defaults (e.g. `--dangerously-skip-permissions`). Fixed so only a non-empty string overrides; `null`/`''`/unset all fall back to the global defaults, preserving behavior for existing automations. Covered by the new tests.

**Cross-platform (macOS/Linux/Windows) confirmed:** no keyboard shortcuts, shortcut labels, or file paths were touched, and no Electron menu/platform APIs. The new string flows through the existing shared startup-plan builder (`tokenizeStartupCommand` + `quoteStartupArg`), the same shell-quoting used for prompts on every platform — including WSL (`wsl.exe` shell override) and SSH/remote launches, which route through the identical background-session spawn path with no platform-specific branches added.

## Security Audit

- **Input handling:** `agentArgs` is a free-text string; the RPC layer validates it as a nullable string via the existing `OptionalNullablePlainString` Zod schema (same helper as `baseBranch`). No max length is enforced — consistent with existing free-text fields; a bound could be added later if untrusted users can author automations.
- **Command execution:** the value is tokenized and shell-quoted by the existing startup-plan builder before embedding in the agent launch command — same trust model as the existing `prompt` field. No raw shell interpolation; multi-flag input (`--yolo --verbose`) is split into individually quoted tokens.
- **Path handling / auth / secrets:** none touched; no new IPC surface (only new params on existing `automation.create`/`automation.update`).
- **Follow-up:** none required for this change.

## Notes

- **CLI parity:** new `--agent-args` flag on `automations create`/`edit`; `automations show` includes the field; `edit` without the flag leaves the stored value unchanged.
- **i18n:** 3 new keys registered in `en.json`; `verify-localization-extraction` and `verify-localization-catalog` both pass.


WaltherMuurmans:
## Review of #11873 — built and measured against `pr11873` (HEAD `92cd0f2`)

Context for why we looked: we run six scheduled automations that each need to launch
`claude` with `--dangerously-skip-permissions --disable-slash-commands --agent <name>`,
set up **from the CLI**. This PR is exactly the feature we've been waiting for, so we
checked out the branch, installed it, ran the suites, and probed the paths we depend on.

Two blockers on the CLI path, plus three smaller notes. Everything below was verified
in this branch; where we measured rather than read, the measurement is included.

### What we verified works

- **argv fidelity is exact.** `planAgentCliArgsSuffix` tokenizes quote-aware and then
  re-quotes per token (`src/shared/tui-agent-startup-shell.ts:114-130`), which is the
  right order. Measured against `buildAgentStartupPlan` (agent `claude`, platform `linux`):

  ```
  in:  --dangerously-skip-permissions --disable-slash-commands --agent beheer-doctor
  out: claude '--dangerously-skip-permissions' '--disable-slash-commands' '--agent' 'beheer-doctor' '<prompt>'

  in:  --model 'claude opus' --append-system-prompt="a b" --agent beheer-doctor
  out: claude '--model' 'claude opus' '--append-system-prompt=a b' '--agent' 'beheer-doctor' '<prompt>'
  ```

  Spaces, the `--flag=value` form and quotes all survive intact.
- **The three-way fallback guard is right**, and it is on exactly one line
  (`src/renderer/src/lib/launch-agent-background-session.ts:57-60`): `undefined`, `null`
  and whitespace-only all fall back to the global defaults; only a non-empty value
  overrides. That is what #11043 asked for.
- **PR suites are green**: 76 tests across `launch-agent-background-session`,
  `useAutomationDispatchEvents`, `cli/format`, `automation-run-identity`,
  `headless-workspace-create`, `AutomationSchedulePicker`. (`orca-runtime-automations`
  did not load here — we installed with `--ignore-scripts`, so native `node-pty` was not
  built. That is our environment, not your PR.)

### 1. Blocker — the runtime drops `agentArgs`, so the whole CLI/RPC path is a no-op

`OrcaRuntimeService.createAutomation` builds an **explicit** field list for the store
and `agentArgs` is not in it (`src/main/runtime/orca-runtime.ts:3549-3567`).
`updateAutomation` has fourteen `hasRuntimeAutomationUpdateValue` branches
(`:3575-3617`) — `name`, `prompt`, `precheck`, `agentId`, `runContext`, `sourceContext`,
`baseBranch`, `setupDecision`, `reuseSession`, `timezone`, `rrule`, `dtstart`, `enabled`,
`missedRunGraceMinutes` — and none for `agentArgs`. This branch does not touch
`orca-runtime.ts` at all (`git show --stat` over the three commits: zero hits).

So the chain is: CLI sends it (`src/cli/handlers/automations.ts:470`, `:496`) → RPC
schema accepts it (`src/main/runtime/rpc/methods/automations.ts:120`, `:142`) → runtime
silently discards it → `Store.createAutomation` writes `agentArgs: input.agentArgs ?? null`
(`src/main/persistence.ts:4894`). Net effect: `orca automations create --agent-args "…"`
stores `null` and reports success.

TypeScript can't catch it: `agentArgs` is optional on the create input
(`src/shared/automations-types.ts:168`) and the update input is a `Partial<Pick<…>>`.

This is asymmetric by transport, not by feature. The local GUI writes straight to the
store over Electron IPC (`src/main/ipc/automations.ts:56`), so the new editor field
works. Everything over RPC — the entire CLI, and the GUI against a remote host — loses
the value.

Worth noting why the green suites don't contradict this: every link in the chain is
covered separately, and the missing link sits between them. A single test that goes
CLI flag → RPC → runtime → store and asserts the stored value would have caught it, and
would keep catching it.

### 2. Blocker — the documented invocation form doesn't parse

`parseArgs` turns a flag into boolean `true` as soon as the next token starts with `--`
(`src/cli/args.ts:119-124`), and `agent-args` is not in `BOOLEAN_FLAGS`. The comment
right above says it explicitly (`:96-99`): `--flag=value` is the only unambiguous form
for a value that itself starts with `--`.

That is precisely what `--agent-args` always carries. Measured by calling `parseArgs`
directly:

```
['automations','create','--agent-args','--dangerously-skip-permissions --disable-slash-commands --agent beheer-doctor']
  -> agent-args = true
  -> plus a bogus flag "dangerously-skip-permissions --disable-slash-commands --agent beheer-doctor"

['automations','create','--agent-args','--yolo']        # the example at src/cli/specs/automations.ts:70
  -> agent-args = true
  -> plus a bogus flag "yolo"

['automations','create','--agent-args=--dangerously-skip-permissions --disable-slash-commands --agent beheer-doctor']
  -> agent-args = "--dangerously-skip-permissions --disable-slash-commands --agent beheer-doctor"   ✅
```

The first two then fail in `validateCommandAndFlags` with `Unknown flag …`
(`src/cli/args.ts:303-308`). So the example in the spec notes does not work as written,
and since passing flags *is* the purpose of this option, this affects most real
invocations. Either document the `=` form (and fix the example), or special-case this
flag in the parser. We've switched our own tooling to `--agent-args=<value>` and it
parses correctly.

### 3. The headless/serve dispatcher never forwards `agentArgs`

When there is no live renderer and a headless dispatcher is configured (serve mode,
`src/main/index.ts:2355`), dispatch goes through `requestHeadlessDispatch`. Neither
branch carries the value:

- `new_per_run`: `buildHeadlessAutomationWorktreeCreateArgs` returns `createdWithAgent`,
  `startupAgent`, `startupPrompt` and no `agentArgs`
  (`src/main/automations/headless-workspace-create.ts:35-46`).
- existing workspace: `launchAgentTerminal` is called with `{ agent, prompt, title }`
  only (`src/main/index.ts:2389-2396`).

Both end in `buildStartupForAgent`, which hardcodes
`agentArgs: resolveTuiAgentLaunchArgs(agent, settings.agentDefaultArgs)`
(`src/main/runtime/orca-runtime.ts:20205`). Serve-mode runs therefore silently use the
global defaults — the exact thing #11043 wants to avoid. The PR's only change here is a
fixture line `agentArgs: null` (`src/main/automations/headless-workspace-create.test.ts:25`)
with no assertion on forwarding, and the `toMatchObject` at `:68-91` wouldn't catch a
regression either.

### 4. One of the new assertions can't fail

`src/renderer/src/hooks/useAutomationDispatchEvents.test.ts:712` asserts
`expect(mockSubmitPromptToAgentTab).not.toHaveBeenCalled()`. The module mock only
provides `submitPromptToAgentTab` (`:129`), but the hook imports and calls
`submitPromptToAgentPty` (`src/renderer/src/hooks/useAutomationDispatchEvents.ts:6`, `:450`).
Repo-wide, `submitPromptToAgentTab` has zero call sites — only its definition at
`src/renderer/src/lib/agent-paste-draft.ts:170` and this mock.

The real protection in that test is
`expect(mockFindReusableAutomationSession).not.toHaveBeenCalled()` (`:708`), which does
fail if the guard is reverted. Suggest dropping line 712, or mocking
`submitPromptToAgentPty` and asserting on that instead. (The stale mock predates this
PR; only the assertion is new.)

### 5. Minor — legacy records have no `agentArgs` key, which makes capability detection unreliable

Stored automations are read back unchanged
(`src/main/persistence.ts:3505`: `automations: Array.isArray(parsed.automations) ? parsed.automations : []`),
with no per-field backfill, and `agentArgs` is only written on create/update. So an
automation created before the upgrade simply has no `agentArgs` key.

The code handles that correctly everywhere (`?.` / explicit `undefined` checks), so this
is not a bug. But it does mean a consumer cannot use `automations list --json` to detect
whether this build supports the field — on our machine all existing automations predate
it. Emitting `agentArgs: null` for legacy records would make the field's presence a
reliable capability signal.

---

Happy to re-test any of this against a follow-up commit — we have the branch built
locally. Findings 1 and 2 are the ones that block CLI-driven use; 3–5 are smaller.

hdsheena:
Thanks for the thorough review and testing! I've rebased the branch on the latest `main` and addressed all findings:

#### 1. Blocker — the runtime drops `agentArgs` on the RPC path
* **Resolution:** Modified `createAutomation` and `updateAutomation` in `src/main/runtime/orca-runtime.ts` to map `agentArgs` correctly. `createAutomation` now forwards `agentArgs: input.agentArgs` to the store, and `updateAutomation` checks and applies `updates.agentArgs` using the `hasRuntimeAutomationUpdateValue` helper.

#### 2. Blocker — the documented invocation form doesn't parse
* **Resolution:** Special-cased the `agent-args` flag in the CLI argument parser `parseArgs` (`src/cli/args.ts`) so that it consumes the next token as its value even if it starts with `--` (e.g. `--agent-args --dangerously-skip-permissions`).
* **Tests:** Added a regression test `accepts a space-separated flag value starting with -- for --agent-args` in `src/cli/args.test.ts` to verify it parses correctly.

#### 3. The headless/serve dispatcher never forwards `agentArgs`
* **Resolution:** Threaded the configuration end-to-end for serve-mode runs:
  * Modified `buildHeadlessAutomationWorktreeCreateArgs` (`src/main/automations/headless-workspace-create.ts`) to return `agentArgs: automation.agentArgs ?? undefined`.
  * Added optional `agentArgs?: string | null` arguments to `createManagedWorktree`, `launchAgentTerminal`, and `buildStartupForAgent` in `src/main/runtime/orca-runtime.ts`.
  * Inside `buildStartupForAgent`, applied the same three-way guard to use custom `agentArgs` over default settings args if present.
  * Inside `src/main/index.ts`, passed `automation.agentArgs` into the `launchAgentTerminal` call for existing workspaces.
  * Added a regression test `threads agentArgs through if non-empty` in `src/main/automations/headless-workspace-create.test.ts`.

#### 4. One of the new assertions can't fail
* **Resolution:** Replaced the mock of the unused `submitPromptToAgentTab` function with `submitPromptToAgentPty` inside `src/renderer/src/hooks/useAutomationDispatchEvents.test.ts` (mocking it from `@/lib/agent-paste-draft`). Updated the assertion to check `expect(mockSubmitPromptToAgentPty).not.toHaveBeenCalled()`.

#### 5. Minor — legacy records have no `agentArgs` key
* **Resolution:** Added a backfill inside the `load()` method in `src/main/persistence.ts` to map any loaded automation record without an `agentArgs` key to `null` on startup, setting `this.loadNeedsSave = true` so the persistence file heals.

#### 6. Bonus — Test Type Safety
* **Resolution:** Fixed TypeScript errors in `automation-host-client.test.ts`, `automation-run-view-state.test.ts`, and `automation-target-availability.test.ts` where spreading overrides into mocked `Automation` instances broke due to missing properties. Spreads are now typed cleanly.

Issue #11878, opened 2026-08-01
[Bug]: Korean double consonants (쌍자음) trigger automatic line breaks in the Windows integrated terminal


BaeTab:
### Operating system

Windows

### Orca version

1.4.163 (also reproduced on earlier releases)

### Details

Typing a Hangul syllable that uses a **double consonant (쌍자음: ㄲ ㄸ ㅃ ㅆ ㅉ)** in the integrated terminal causes an **automatic line break**. No Enter key is pressed — the newline appears on its own while the composition is active.

On the 2-set Korean layout, double consonants are produced with **Shift held** (e.g. ㄲ = Shift+ㄱ, ㅆ = Shift+ㅅ). The symptom does not reproduce with single-consonant syllables, which never use Shift.

**What happens:**
1. Type a syllable like "싸" or "째" — Shift is pressed for the initial double consonant.
2. The Shift keyup during the composition appears to be consumed/lost.
3. A subsequent **plain** Enter (or the Enter that commits the composition) is misread as Shift+Enter, producing a newline instead of a `\r` submit — or the committing Enter leaks a newline mid-composition.

**Likely code path** (from source inspection, not yet verified by a debugger):
- `keyboard-handlers.ts:298` `onModifierDown` records Shift into `heldImeEnterModifiers` for the double-consonant keystroke.
- `keyboard-handlers.ts:287` `getImeEnterModifier` falls back to `getHeldImeEnterModifier()` when the event carries no modifier, so a stale Shift from the 쌍자음 is attributed to the next plain Enter.
- That routes the Enter into the `modifiedEnterChordOwner` claim/absorb path (`keyboard-handlers.ts:446`, `:537`) and the deferred-newline machinery, which emits a newline instead of a plain `\r`.

The existing e2e cases (`pre-held Shift/Ctrl+Enter with modifier-lost redispatch`) model a modifier that is *still held* during the committing chord, not one whose keyup vanished mid-composition. A repro case that dispatches Shift down, composes a 쌍자음 syllable, drops the Shift keyup, then sends a plain Enter would cover this gap.

Related: #9803 (broader Windows Korean IME issue), #10343 (Shift+Enter last-character newline), #8038 (last syllable dropped to a new line on Enter, macOS).

BaeTab:
**Correcting my own report.** The symptom above is real and still reproduces on v1.4.163, but the "Likely code path" section is **wrong**, and I'd rather not leave it pointing people at the wrong file.

[...]

PR #11879, opened 2026-08-01
fix(tab-bar): ignore IME composition Enter in the quick commands menu

BaeTab:
## Summary

On the 2-set Korean layout, seven jamo require Shift — ㅃ ㅉ ㄸ ㄲ ㅆ ㅒ ㅖ — so Korean users compose constantly inside the quick-commands search field. The Enter that only confirms a CJK composition candidate was reaching `handleSearchKeyDown` and running the highlighted quick command, **executing a terminal command the user never picked**. The last syllable is lost and the menu closes.

The arrow keys had the same problem in the same handler: `ArrowDown`/`ArrowUp` moved the command selection while the IME was still using them to walk its own candidate list (Japanese/Chinese conversion, and Korean candidate lists where present).

Both branches are now guarded with the existing `isImeCompositionKeyDown` helper — the same one `SortableTab.tsx:291` and `EditorFileTab.tsx:308` in this very folder already use. This field was simply missed when those were migrated.

The printable-key `stopPropagation()` branch is deliberately **not** guarded, so Radix typeahead suppression behaves exactly as before.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — see Notes for two wrapper scripts that cannot run on Windows; their underlying `oxlint` invocations were run directly and are clean
- [x] `pnpm test` — ran the affected scope: 38 files / 266 tests pass across `src/renderer/src/components/tab-bar/` plus `ime-composition-keyboard-event.test.ts`

New colocated test: `TabBarQuickCommandsMenu.ime-enter.test.tsx`, modeled on the existing `SmartWorkspaceNameField.ime-enter.test.tsx`.

**Verified the tests actually fail without the fix.** Stashing only the production change leaves the two control cases green and turns exactly the three guard cases red:

```
× does not run the highlighted command on the Enter that commits a CJK composition
× does not run the highlighted command on an Enter reported as keyCode 229
× leaves the highlighted row alone when an arrow key belongs to the IME
    AssertionError: expected "vi.fn()" to be called with arguments: [ { id: 'cmd-deploy', …(3) } ]

Tests  3 failed | 2 passed (5)
```

With the fix restored: 5 passed (5).

The two passing controls matter — a plain Enter still runs the highlighted command, and a plain `ArrowDown` still advances the selection — so the guard is proven to suppress only composition keystrokes rather than disabling the handler.

## AI Review Report

- **Over-broad guard.** The first draft returned early for the whole handler, matching `SortableTab`'s shape. Review flagged that this would also skip the `event.key.length === 1` → `stopPropagation()` branch; on platforms where Chromium reports the composed character rather than `key: 'Process'`, the keystroke would then bubble to Radix's `DropdownMenu` typeahead and jump the highlighted row while typing Korean. Narrowed to guard only the two branches that take action. No behavior change for printable keys.
- **Cross-platform.** No `metaKey`/`ctrlKey` hardcoding is introduced; the change touches neither shortcuts nor labels nor paths nor shell behavior. `isImeCompositionKeyDown` reads only `isComposing` and `keyCode === 229` and is already shipped on all three platforms. The bug reproduces on Windows (Microsoft 두벌식) and the same code path is platform-independent, so macOS and Linux IMEs get the identical fix. No Electron-specific API touched.
- **SSH / remote and folder workspaces.** The handler is pure renderer keyboard logic with no host, transport, or workspace-type dependency; behavior is identical for SSH hosts, WSL, and folder workspaces.
- **Neighboring handler considered and excluded.** `DropdownMenuContent`'s own `onKeyDown` also calls `runAndClose`, but only when `!showSearch` — i.e. when there is exactly one command and no `CommandInput` is rendered. With no text input on that surface no composition can occur, so a guard there would be dead code. Left unchanged intentionally.
- **Line-number-only lint warning.** `react-doctor(no-array-index-as-key)` reports on this file, but at unchanged content; `git diff -U0` confirms the hunks are limited to lines 23, 162–166 and 176. No `max-lines` suppression was added or bumped; `check:max-lines-ratchet` passes.

## Security Audit

Basic audit from the same review. This change only *narrows* the conditions under which a command is executed, so it strictly reduces the attack surface rather than expanding it:

- **Command execution.** The bug itself is the security-relevant part — an IME keystroke could trigger `onRunCommand`, running a shell command without deliberate user selection. The guard removes that path. No new execution path is added.
- **Input handling.** Reads three boolean/numeric fields off the keyboard event (`isComposing`, `keyCode`, `key`). No parsing, no interpolation, no `dangerouslySetInnerHTML`.
- **Path handling / secrets / auth / IPC / dependencies.** Untouched. No new dependency; `isImeCompositionKeyDown` is an existing in-repo module.
- **Test doubles.** The new test mocks only UI primitives and the search ranking module; it spawns no process and touches no filesystem or network.

No follow-up required.

## Notes

**Two `pnpm lint` sub-steps cannot execute on Windows.** `check-changed-code-quality.mjs` and `lint-react-doctor-changed.mjs` both die with `Error: spawnSync pnpm.cmd EINVAL` before reaching any analysis — the known Node-on-Windows restriction on spawning `.cmd` without `shell: true`. This is environmental and unrelated to this change; both run normally on the Linux CI runners. I ran their underlying commands directly instead:

- `oxlint <both files>` — clean
- `oxlint --config config/oxlint-react-doctor.json <both files>` — only the pre-existing warning noted above

The remaining `pnpm lint` steps all pass locally: `oxlint`, `check:reliability-gates`, `check:max-lines-ratchet`, `verify:localization-catalog`, `verify:localization-extraction`, `verify:localization-coverage`.

**`pnpm build` not run locally.** This machine is on Node v22.13.1 while `engines` requires Node 24, and `pnpm install` reported a `windows-native-registry` native rebuild failure unrelated to this change. Since this PR touches only renderer TypeScript with no native, packaging, or build-config surface, CI's `package (ubuntu)` / `package (windows)` jobs are the meaningful signal.

**Related.** This is the same class of bug as the merged #9526 and the open #11067, which migrated other inputs onto `isImeCompositionKeyDown`. #11067 covers the rename/notes dialogs and does not touch this file, so there is no overlap. Separately, #11878 / #11616 concern the Windows terminal IME `Process`-key path, which is unrelated to this renderer handler.

Other unguarded Enter-submit handlers remain (comment composers, markdown search-and-replace, several pickers). Happy to follow up with those in separate PRs if the approach here looks right.

### DIFF
--- a/src/renderer/src/components/tab-bar/TabBarQuickCommandsMenu.ime-enter.test.tsx
+++ b/src/renderer/src/components/tab-bar/TabBarQuickCommandsMenu.ime-enter.test.tsx
@@ -1,0 +1,208 @@
+// @vitest-environment happy-dom
+
+import React, { act } from 'react'
+import type * as ReactNamespace from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { TabBarQuickCommandsMenu } from './TabBarQuickCommandsMenu'
+import type { TerminalQuickCommand } from '../../../../shared/types'
+
+vi.mock('@/components/ui/command', async () => {
+  const react = await vi.importActual<typeof ReactNamespace>('react')
+  return {
+    Command: ({ children }: { children?: React.ReactNode }) => <div>{children}</div>,
+    CommandEmpty: ({ children }: { children?: React.ReactNode }) => <div>{children}</div>,
+    CommandInput: react.forwardRef<
+      HTMLInputElement,
+      { onKeyDown?: React.KeyboardEventHandler<HTMLInputElement> }
+    >(({ onKeyDown }, ref) => (
+      <input ref={ref} data-quick-command-search="true" onKeyDown={onKeyDown} />
+    )),
+    CommandList: react.forwardRef<HTMLDivElement, { children?: React.ReactNode }>(
+      ({ children }, ref) => <div ref={ref}>{children}</div>
+    ),
+    CommandSeparator: () => <div />
+  }
+})
+
+vi.mock('@/components/ui/dropdown-menu', () => ({
+  DropdownMenu: ({ children }: { children?: React.ReactNode }) => <div>{children}</div>,
+  DropdownMenuContent: ({
+    children,
+    onKeyDown
+  }: {
+    children?: React.ReactNode
+    onKeyDown?: React.KeyboardEventHandler<HTMLDivElement>
+  }) => <div onKeyDown={onKeyDown}>{children}</div>,
+  DropdownMenuTrigger: ({ children }: { children?: React.ReactNode }) => <div>{children}</div>
+}))
+
+vi.mock('@/components/ui/tooltip', () => ({
+  Tooltip: ({ children }: { children?: React.ReactNode }) => <>{children}</>,
+  TooltipContent: ({ children }: { children?: React.ReactNode }) => <div>{children}</div>,
+  TooltipTrigger: ({ children }: { children?: React.ReactNode }) => <>{children}</>
+}))
+
+vi.mock('@/components/ShortcutKeyCombo', () => ({
+  ShortcutKeyCombo: () => <span />
+}))
+
+vi.mock('./TabBarQuickCommandItem', () => ({
+  TabBarQuickCommandItem: () => <div />
+}))
+
+vi.mock('@/hooks/useShortcutLabel', () => ({
+  useShortcutKeyComboDetails: () => []
+}))
+
+vi.mock('./tab-bar-quick-commands-shortcut', () => ({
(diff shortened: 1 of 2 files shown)


nwparker:
Closing — main already has this guard, arrived by a refactor rather than by this PR.

`handleSearchKeyDown` was extracted into `use-tab-bar-quick-command-search-input.ts`, which now opens with `if (imeEnter.ownsKeyDown(event)) { return }` ahead of both the Enter branch and the Arrow branches — the same two guards you added. It goes through `useImeEnterGestureOwnership`, which also handles the platform-split redispatched-Enter ordering that a plain `isComposing` check misses.

Thanks for spotting it. Your other one, #11882, is *not* covered — the comment composer and the replace field both still submit on a bare Enter with no composition guard. That one is still wanted.

PR #11882, opened 2026-08-01
fix(ui): ignore IME composition Enter in the comment composer and replace field

BaeTab:
## Summary

Follow-up to #11879, same class of bug in two more surfaces.

On the 2-set Korean layout, seven jamo require Shift — ㅃ ㅉ ㄸ ㄲ ㅆ ㅒ ㅖ — so Korean users are composing for a large share of ordinary typing (`있다`, `했다`, `예를 들어`, `얘기`). The Enter that only confirms a CJK candidate was reaching two handlers that act on it:

- **`right-panel-comment-composer.tsx`** — `Cmd/Ctrl+Enter` posted the comment mid-composition, so the comment goes out missing its final syllable. Long-form prose, so this is the highest-loss case.
- **`RichMarkdownSearchBar.tsx`** (replace field) — Enter ran `onReplaceCurrent()`, **mutating the document** on a keystroke aimed at the candidate list. Escape had the mirror problem: it closed the search bar instead of letting the IME cancel the composition.

Both are guarded with the existing `isImeCompositionKeyDown` helper, the same one the rename and title inputs already use (`SortableTab.tsx:291`, `EditorFileTab.tsx:308`, `WorktreeTitleInlineRename.tsx:231`, `WorkspaceDirectorySetting.tsx:203`). These two fields were simply missed in that migration.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — see Notes for two wrapper scripts that cannot run on Windows; their underlying `oxlint` invocations were run directly and are clean
- [x] `pnpm test` — ran the affected scope: `src/renderer/src/components/right-sidebar/` + `src/renderer/src/components/editor/`, 2559 passed. Two pre-existing failures, see Notes.

Two new colocated tests, modeled on the merged `SmartWorkspaceNameField.ime-enter.test.tsx`:

- `right-panel-comment-composer.ime-enter.test.tsx`
- `RichMarkdownSearchBar.ime-enter.test.tsx`

**Verified both fail without the fix.** Stashing only the production file leaves the controls green and turns exactly the guard cases red:

```
right-panel-comment-composer
× does not submit on the Ctrl+Enter that commits a CJK composition
× does not submit on a Ctrl+Enter reported as keyCode 229

RichMarkdownSearchBar
× does not replace on the Enter that commits a CJK composition
× does not replace on an Enter reported as keyCode 229
× does not close the bar on the Escape that cancels a composition
  Tests  3 failed | 2 passed (5)
```

With the fixes restored: 3/3 and 5/5.

The passing controls are the point — a plain `Ctrl+Enter` still submits, a plain Enter still replaces, and a plain Escape still closes the bar. The guards suppress only composition keystrokes rather than disabling the handlers.

## AI Review Report

- **Escape semantics.** Initially only the Enter branch was guarded. Review flagged that Escape during composition belongs to the IME too — it cancels the candidate, and swallowing it to close the search bar loses the user's composition. Both handlers only branch on Enter/Escape, so an early return is the correct shape here (matching `SortableTab.tsx:289-300`) rather than per-branch conditions.
- **Narrower shape rejected for the search bar's other input.** Only the replace field is changed. The find field's next/prev handlers are navigational and out of scope for this PR.
- **Cross-platform.** `right-panel-comment-composer` already derives its submit modifier via `navigator.userAgent.includes('Mac')` at line 89 — untouched, and no `metaKey` is newly hardcoded. The test pins the user agent to the non-Mac branch so it asserts `Ctrl+Enter` deterministically rather than depending on the runner's platform. `isImeCompositionKeyDown` reads only `isComposing` and `keyCode === 229` and already ships on all three platforms. No paths, shell behavior, or Electron APIs touched.
- **SSH / remote and folder workspaces.** Pure renderer keyboard logic with no host, transport, or workspace-type dependency; identical behavior for SSH hosts, WSL, and folder workspaces.
- **No `max-lines` suppression** added or bumped; `check:max-lines-ratchet` passes. `oxlint`, `react-doctor`, `check:reliability-gates`, and all three localization verifiers pass.

## Security Audit

Basic audit from the same review. Both changes only *narrow* the conditions under which an action fires, so they reduce rather than expand the surface:

- **Unintended mutation.** The `RichMarkdownSearchBar` case is the security-relevant one — an IME keystroke could rewrite document content without a deliberate user action. The guard removes that path. No new mutation path is added.
- **Input handling.** Reads three fields off the keyboard event (`isComposing`, `keyCode`, `key`). No parsing, no interpolation, no `dangerouslySetInnerHTML`.
- **Command execution / path handling / secrets / auth / IPC / dependencies.** Untouched. No new dependency; `isImeCompositionKeyDown` is an existing in-repo module.
- **Test doubles.** Both tests mock only UI primitives and i18n; no process, filesystem, or network access.

No follow-up required.

## Notes

**Two pre-existing test failures in the affected folders, unrelated to this PR.** Both are Windows CRLF issues and were confirmed by stashing every local change and re-running against a clean `upstream/main` — they fail identically there:

- `monaco-content-sync.undo-history.test.ts` — `expected 'first line\r\nappended' to be 'first line\nappended'`
- `SourceControl.host-context-boundary.test.ts` — a source-text assertion that does not match under CRLF

**Two `pnpm lint` sub-steps cannot execute on Windows.** `check-changed-code-quality.mjs` and `lint-react-doctor-changed.mjs` both die with `Error: spawnSync pnpm.cmd EINVAL` before reaching any analysis — the known Node-on-Windows restriction on spawning `.cmd` without `shell: true`. Environmental and unrelated to this change; both run normally on the Linux CI runners. Their underlying commands were run directly and are clean.

**`pnpm build` not run locally.** This machine is on Node v22.13.1 while `engines` requires Node 24. This PR touches only renderer TypeScript with no native, packaging, or build-config surface, so CI's `package` jobs are the meaningful signal.

**Related.** #11067 migrates the rename/notes dialogs onto the same helper and does not touch either file here, so there is no overlap. #11879 covers the quick-commands menu.

Other unguarded Enter-submit handlers remain — the Linear and GitHub Projects title fields, the native-chat free-text answer, and several pickers. Happy to continue in separate PRs if this shape looks right.

### DIFF
--- a/src/renderer/src/components/editor/RichMarkdownSearchBar.ime-enter.test.tsx
+++ b/src/renderer/src/components/editor/RichMarkdownSearchBar.ime-enter.test.tsx
@@ -29,6 +29,52 @@ afterEach(() => {
   container.remove()
 })
 
+type BarSpies = {
+  onReplaceCurrent?: () => void
+  onClose?: () => void
+  onMoveToMatch?: (direction: 1 | -1) => void
+}
+
+function renderFields(spies: BarSpies): {
+  searchInput: HTMLInputElement
+  replaceInput: HTMLInputElement
+} {
+  act(() => {
+    root.render(
+      <RichMarkdownSearchBar
+        activeMatchIndex={0}
+        isOpen
+        isReplaceMode
+        matchCase={false}
+        matchCount={2}
+        query="배포"
+        replaceQuery="릴리스"
+        replaceDisabled={false}
+        searchInputRef={createRef<HTMLInputElement>()}
+        wholeWord={false}
+        onClose={spies.onClose ?? vi.fn()}
+        onMoveToMatch={spies.onMoveToMatch ?? vi.fn()}
+        onQueryChange={vi.fn()}
+        onReplaceAll={vi.fn()}
+        onReplaceCurrent={spies.onReplaceCurrent ?? vi.fn()}
+        onReplaceQueryChange={vi.fn()}
+        onToggleMatchCase={vi.fn()}
+        onToggleReplaceMode={vi.fn()}
+        onToggleWholeWord={vi.fn()}
+      />
+    )
+  })
+  const inputs = [...container.querySelectorAll('input')]
+  // Why: the two rows are distinguished by their values, not by order-dependent
+  // selectors that would silently pick the wrong field if the layout changes.
+  const searchInput = inputs.find((candidate) => candidate.value === '배포')
+  const replaceInput = inputs.find((candidate) => candidate.value === '릴리스')
+  if (!searchInput || !replaceInput) {
+    throw new Error('search bar fields not rendered')
+  }
+  return { searchInput, replaceInput }
+}
+
 function renderBar(spies: { onReplaceCurrent: () => void; onClose: () => void }): HTMLInputElement {
   act(() => {
     root.render(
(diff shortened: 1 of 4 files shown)


BaeTab:
Good catch — fixed in 36e2206b5.

You're right that this was inconsistent within a single file. I guarded the replace field and left the primary search input unguarded, even though the PR description specifically claims to stop "Escape from closing the search bar instead of cancelling the IME composition." That handler does exactly that on Escape, and jumps to the next match on Enter.

Applied the same `isImeCompositionKeyDown` early return there. Shift+Enter was affected in the other direction too, so the tests cover both.

The search-field suite now has 7 cases — 4 guarded and 3 controls:

| | guarded |
|---|---|
| Enter + `isComposing` | does not move to next match |
| Shift+Enter + `isComposing` | does not move to previous match |
| Enter + `keyCode: 229` | does not move |
| Escape + `isComposing` | does not close the bar |

| | control |
|---|---|
| plain Enter | still moves to next match |
| plain Shift+Enter | still moves to previous match |
| plain Escape | still closes the bar |

**Verified the four guard cases fail without the change.** Stashing only `RichMarkdownSearchBar.tsx` leaves the already-committed replace-field tests green and turns exactly the new ones red:

```
× does not move to the next match on the Enter that commits a composition
× does not move to the previous match on a composing Shift+Enter
× does not move on an Enter reported as keyCode 229
× does not close the bar on the Escape that cancels a composition
  Tests  4 failed | 8 passed (12)
```

The test helper now resolves both rows by their field values rather than by index, so a layout change cannot silently point a test at the wrong input.

`pnpm typecheck` clean across all three projects; `oxlint` and React Doctor clean on both files. The one failing test in `src/renderer/src/components/editor/` is the pre-existing `monaco-content-sync.undo-history` CRLF failure, which reproduces identically on a clean `upstream/main` checkout on this Windows host.

nwparker:
Thanks for this, @BaeTab — the IME Enter diagnosis was right, and the follow-up coverage on the search bar was useful.

This already landed on `main` in #15057, which ported your change onto current main with `Co-authored-by` credit. I'm closing this PR as no longer needed.

This note is from an automated scan of open PRs whose issue already shipped in a later merge. If that's wrong, please reopen and say so.

PR #11883, opened 2026-08-01
fix(cli): align Linear and emulator tables by display width, not code units

BaeTab:
## Summary

Column-aligned CLI and SSH tables are ragged for any non-Latin name.

A Hangul syllable is **one UTF-16 code unit but two terminal columns**, and `String.prototype.padEnd` counts code units. So `padEnd(24)` on a 3-syllable Korean name emits 24 code units that occupy **27 columns**, while `Alice` occupies 24. Because the overshoot equals the number of wide characters in that cell, every row drifts by a *different* amount — the table does not merely shift right, it becomes unreadable:

```
# before — id column starts at 27, 24, 29, 24 …
배현우                      user-0
Alice                    user-1
김철수영희                        user-2
Bob                      user-3

# after — every id starts at column 24
배현우                   user-0
Alice                    user-1
김철수영희          user-2
Bob                      user-3
```

A Korean Linear workspace (Korean member display names, workflow state names, label and project names) hits this on essentially every row. Han and kana behave identically.

This adds `src/shared/east-asian-display-width.ts` and routes the 19 column-aligned `padEnd` calls through it, across the Linear CLI output, the SSH Linear output, the shared project list, and the emulator device list.

**ASCII output is byte-identical to before** — the first helper test asserts exactly that against the built-in.

### Why a new module

`terminal-unicode-provider.ts` cannot be reused here: it wraps an xterm `IUnicodeVersionProvider`, which requires a live terminal instance, while these tables render in the CLI process and the SSH main process. A repo-wide search for `stringWidth` / `displayWidth` / `wcwidth` / `eastAsianWidth` found no other display-width helper, so the call sites had nothing width-aware available.

The ranges are the Wide and Fullwidth blocks of Unicode East_Asian_Width plus the zero-width combining ranges. The module comment states plainly that this is a pragmatic subset covering the scripts these tables carry, not a complete East_Asian_Width implementation.

### Deliberately left alone

- `src/cli/help.ts:415` — `padEnd(18)` on command paths that are ASCII by construction.
- `src/renderer/src/web/web-pairing.ts:139` — base64 `=` padding, unrelated to display width.
- `src/cli/terminal-format.ts` — uses double-space separators rather than `padEnd`, so it is already immune.

## Screenshots

No visual change to the UI. Terminal output changes only for non-Latin names; the before/after is in the Summary.

## Testing

- [x] `pnpm lint` — see Notes for two wrapper scripts that cannot run on Windows; `oxlint`, `oxlint --config config/oxlint-code-quality-native-plugins.json --deny-warnings`, `check:max-lines-ratchet`, `check:reliability-gates`, and all three localization verifiers pass
- [x] `pnpm test` — affected scope green: `src/cli/` + the new helper, 706 passed. Pre-existing unrelated failures noted below.

Two test surfaces:

1. `src/shared/east-asian-display-width.test.ts` — 12 unit cases: ASCII parity with the built-in, Hangul/CJK/kana/fullwidth widths, astral characters counted once rather than per surrogate half, and combining marks / variation selectors / control characters counted as zero.
2. `src/cli/linear-format.test.ts` — two call-site regression tests that assert the **actual formatter output** aligns, not just the helper.

**Verified the call-site tests fail without the fix.** Stashing only `src/cli/linear-format.ts`:

```
× starts the id column at the same display column for Korean and Latin names
    AssertionError: expected 27 to be 24
× keeps every row aligned regardless of how many wide characters it holds
    AssertionError: expected 4 to be 1
  Tests  2 failed | 5 passed (7)
```

`27 to be 24` is the three-column overshoot from three Hangul syllables. `4 to be 1` is four rows landing at four *different* columns — the raggedness this PR fixes. With the fix restored: 7/7.

## AI Review Report

- **Behavior change for existing ASCII output.** The first helper test asserts `padEndToEastAsianDisplayWidth('Alice', 10) === 'Alice'.padEnd(10)` so any drift from the built-in on Latin text fails the suite. All 19 replacements keep their original column constants.
- **Astral characters.** An early draft iterated with `charCodeAt`, which counts a surrogate pair twice. The implementation iterates the string (`for…of`), which yields whole code points; a test pins this using U+1F680. Review also caught that the emoji ranges initially omitted Transport and Map Symbols, so U+1F680 measured 1 — fixed and covered.
- **Over-counting invisible characters.** Combining marks, variation selectors, zero-width joiners, and C0/C1 controls return 0, otherwise a name carrying one would over-pad its row.
- **Cross-platform.** Pure string arithmetic — no paths, no shell, no platform branches, no Electron APIs. Identical on macOS, Linux, and Windows.
- **SSH and folder workspaces.** `ssh-remote-linear-output.ts` is on the SSH path specifically; the change is presentation-only and carries no host, transport, or workspace-type dependency.
- **Naming.** Per AGENTS.md the module is named for what it contains (`east-asian-display-width`) rather than a generic `utils`/`helpers` role. No `max-lines` suppression added or bumped.

## Security Audit

Basic audit from the same review:

- **Input handling.** The module reads code points from strings already being rendered and returns a count plus spaces. No parsing, no regex over untrusted input, no interpolation into a shell or query.
- **Resource use.** `getEastAsianDisplayWidth` is O(code points) with a bounded range table; `padEndToEastAsianDisplayWidth` only ever appends spaces and never truncates, so it cannot silently drop data.
- **Command execution / path handling / secrets / auth / IPC / dependencies.** Untouched. No new dependency — deliberately implemented in-repo rather than pulling in `string-width` and its transitive tree.

No follow-up required.

## Notes

**Pre-existing test failures in the touched areas, unrelated to this PR.** Confirmed by stashing every local change and re-running against a clean `upstream/main` — 14 tests fail identically there, all in `src/main/ssh/ssh-relay-upload-stage-commands.test.ts` (POSIX shell-command tests that do not pass on a Windows host).

**Two `pnpm lint` sub-steps cannot execute on Windows.** `check-changed-code-quality.mjs` and `lint-react-doctor-changed.mjs` both die with `Error: spawnSync pnpm.cmd EINVAL` before reaching any analysis — the known Node-on-Windows restriction on spawning `.cmd` without `shell: true`. Environmental and unrelated; both run normally on the Linux CI runners. Their underlying `oxlint` commands were run directly and are clean.

**`pnpm build` not run locally.** This machine is on Node v22.13.1 while `engines` requires Node 24. This PR touches only CLI/shared/main TypeScript with no native, packaging, or build-config surface, so CI's `package` jobs are the meaningful signal.

**Related.** #11879 and #11882 fix Korean IME Enter handling in the renderer. This one is unrelated to IME — it is display-width arithmetic in terminal output — and touches no shared files with either.

If the range table would be better sourced from a generated Unicode table than a hand-maintained list, I am happy to rework it that way.


BaeTab:
Thanks both — the review caught a real class of problem, so I've replaced the hand-maintained table rather than patching it.

## Verification method

I probed `@xterm/addon-unicode11` directly, since that is the provider `activateOrcaTerminalUnicodeProvider` registers on Orca's own terminals. It is the authority that matters here: if the CLI disagrees with it, the column arithmetic disagrees with what the user is looking at.

| Code point | xterm unicode11 | old table |
|---|---|---|
| U+1100 choseong | 2 | 2 ✅ |
| U+1160–U+11FF conjoining jungseong/jongseong | **0** | 1 ❌ |
| U+D7B0–U+D7FB Jamo Extended-B | **1** | 0 ❌ |
| U+3099–U+309A Japanese combining marks | **0** | 2 ❌ |
| U+1FA00–U+1FAFF (🫠 U+1FAE0) | **1** | 1 ✅ |

A full sweep of U+0000–U+2FFFF found the hand table disagreeing with the provider at roughly **9,900 code points** — mostly missing combining marks across Arabic, Hebrew, and Devanagari (under-padding those names) and unassigned gaps inside my over-broad blocks.

So: @greptile-apps was right on both counts that the ranges were wrong, though Extended-B is Neutral (1) rather than Wide (2) — it is conjoining, so it is neither a full cell nor zero.

@coderabbitai the Symbols Extended-A point does not reproduce: `wcwidth(0x1FAE0)` is `1` in this provider, because that block postdates Unicode 11. The helper already returns 1 there, matching the terminal. Adding a hardcoded 2 would have made the CLI disagree with the pane next to it. Happy to revisit if Orca moves off unicode11.

## What changed

- **`config/scripts/generate-east-asian-display-width-table.mjs`** generates `src/shared/east-asian-display-width-table.ts` from the provider. 447 ranges, emitted as flat inclusive pairs.
- **Wired `--check` into `pnpm lint`.** It compares code points rather than file bytes, so oxfmt owning the wrapping cannot make a correct table look stale. An xterm bump that moves a width now fails CI instead of silently skewing a column. I verified it detects a one-digit corruption.
- **ZWJ sequences now collapse to one glyph.** `👩‍💻` measured 2 + 0 + 2 = 4 before; it is 2 now. This mirrors the `orca-11-zwj` rule in `terminal-unicode-provider.ts`, whose comment already says CLIs "budget them as one wide cell pair" — the CLI was the one place not honouring it. Good catch, @coderabbitai.

Divergence from the provider is now zero by construction rather than by inspection.

## Tests

The helper suite grew from 12 to 17 cases, adding conjoining jamo, Extended-B, and two ZWJ cases (the sequence measuring 2, and a ZWJ name aligning against a Latin one). The two call-site regression tests are unchanged and still fail without the fix:

```
× starts the id column at the same display column for Korean and Latin names
    AssertionError: expected 27 to be 24
× keeps every row aligned regardless of how many wide characters it holds
    AssertionError: expected 4 to be 1
```

`pnpm typecheck` clean across all three projects; 710 tests pass in the affected scope; `check:max-lines-ratchet`, `check:reliability-gates`, and the new `verify:east-asian-display-width-table` all pass.

BaeTab:
Both fixed — thanks, these were real.

**`file://` string concatenation.** Correct, and it only worked here by luck. Concatenating mis-parses a Windows drive letter as the URL *host* and leaves spaces unencoded:

```
hand-built : file://D:/myrepo/orca/node_modules/...      <- "D:" becomes the host
pathToFileURL : file:///D:/myrepo/orca/node_modules/...
with a space  : file:///C:/Program%20Files/a%20b/x.mjs
```

Any checkout under `C:\Program Files\` or a user profile containing a space would have failed to import. Now uses `pathToFileURL(addonPath).href`.

**Four sweeps.** Also correct, and it matters more than it looks because this now runs on every `pnpm lint`. It walked all 1,114,112 code points twice to build the file contents and twice more to compare them — and in `--check` mode the rendered contents were discarded, so half the work was pure waste:

| | `wcwidth` calls |
|---|---|
| before | 4,456,448 |
| after | 1,114,112 |

`collectWidthRanges` now accumulates both width classes in one pass, asking the provider once per code point, and `renderFile` is only called when actually writing. The generated table is byte-identical — verified by `--check` passing against the unchanged committed file.

While in there I also made the generator emit at the formatter's print width, so regenerating no longer produces a formatting-only diff against the committed file. Generator output and `oxfmt --write` output are now byte-identical, which means `--check` and the pre-commit hook cannot disagree.

PR #11894, opened 2026-08-01
build(xterm): make the xterm patch reproducible from a pinned upstream build

nwparker:
## Summary

Makes the vendored xterm patch reproducible instead of hand-maintained.

Orca patches `@xterm/xterm` to add composition-session events that the terminal IME path depends on. The patch was previously edited by hand against built output, so nobody could tell which upstream commit it applied to, whether it still applied cleanly, or what it actually changed semantically. That is a bad foundation for the IME fixes in this stack, which all reason about the patched `CompositionHelper`.

Changes:

- `config/patches/xterm-upstream.json` pins the exact upstream revision the patch is generated from.
- `config/scripts/regenerate-xterm-patches.mjs` rebuilds both the `src` patch and the dist patch from that pinned build, so the diff is derived, not authored.
- Splits out `@xterm__xterm@6.1.0-beta.287.src.patch` so the TypeScript source changes are reviewable — this is what makes the `CompositionHelper` behavior in #11896/#11897 auditable at all.
- `docs/reference/xterm-patch-regeneration.md` documents the workflow.
- A PR CI job regenerates and verifies the patch is byte-identical, so drift fails loudly.
- Second commit: the generated patch deletes the bundle sourcemaps rather than carrying them, cutting it from **7,334,810 to 1,529,822 bytes (−79.1%)**.
- Third commit: the generator now maintains *every* place pnpm records the patch hash, not just one of them. See below — this was a real CI failure on the PRs above, not a hypothetical.

## Why delete the sourcemaps

The patch rewrites `lib/xterm.js` and `lib/xterm.mjs`, so a retained `.map` would have to move with them. Simply omitting the map hunks — what the two addon patches in this repo do today — ships maps whose offsets point at the wrong code, which is worse than having none. Deleting them is the honest form of the same size saving:

| | before | after |
| --- | ---: | ---: |
| `config/patches/@xterm__xterm@6.1.0-beta.287.patch` | 7,334,810 | 1,529,822 |
| `…/xterm-src/@xterm__xterm@6.1.0-beta.287.src.patch` | 32,583 | 32,583 |

`sourcemaps.policy` in the manifest controls this, and `assertSourcemapPolicy` rejects any value the generator does not implement — without it a typo would silently fall back to including the maps and re-inflate the patch by 5.8 MB.

Verified against a real install, not a scratch project: `pnpm install --frozen-lockfile` succeeds cold, the maps are absent from `node_modules/@xterm/xterm/lib/`, zero `sourceMappingURL` references remain, and the two bundles are byte-identical to a clean build after stripping the trailing comment (−34 B / −36 B). `--check` reports `in sync (1529822 bytes)` from an empty work directory, so the reduction is reproducible from nothing.

## `.gitattributes`

`/config/patches/@xterm__xterm@*.patch` is marked `-diff -text`. `-text` is load-bearing, not cosmetic: pnpm hashes the patch byte-for-byte, so a CRLF checkout on Windows changes the hash and breaks `pnpm install`. The rule lands in the same commit that introduces the patch — parking it higher in the stack would leave every intermediate PR broken for Windows contributors.

## pnpm writes the patch hash in two places

`--check` and `--write` previously managed one of them, and the gap cost this stack a full CI failure. Worth writing down, because the failure mode is specifically designed to look like success.

pnpm keys a patched package by the sha256 of the patch file. It records that hash under `patchedDependencies:`, and again inside **every resolution key that depends on the patched package** — 23 of them for `@xterm/xterm`, in two different spellings:

```yaml
patchedDependencies:
  '@xterm/xterm@6.1.0-beta.287':
    hash: ec0d5d6e…              # ← the generator maintained this one

snapshots:
  '@xterm/xterm@6.1.0-beta.287(patch_hash=ec0d5d6e…)':   # ← and not these
    dependencies:
      '@xterm/xterm': 6.1.0-beta.287(patch_hash=ec0d5d6e…)
```

Regenerating the patch therefore produced a lockfile with a correct `patchedDependencies` hash and 23 stale resolution keys. On a warm store `pnpm install` resolves from cache, reports success, and leaves the stale keys in place; `pnpm install --frozen-lockfile` then *passes locally*. On a cold store — which is CI — the same lockfile drifts and the install step fails. `pnpm install --force` makes it worse: it rewrites the resolution keys back to the cached value, so the obvious remedy silently reintroduces the bug.

That is what took out ~40 checks on the two PRs above this one. The install job's diff read `-d396b2ad / +ec0d5d6e`, and `shasum -a 256` of the patch file agreed with CI, not with the lockfile.

The fix is in `regenerate-xterm-patches.mjs`:

- `readLockfileResolutionHashes` finds both spellings (`name@version(patch_hash=…)` and the bare `: version(patch_hash=…)`).
- `lockfilePatchHashIsStale` is now the staleness predicate for both `--write` and `--check`, so a lockfile stale in its resolution keys alone is a `--check` failure rather than a clean bill of health.
- `--check`'s failure message prints all three values — `patchedDependencies`, the distinct stale resolution hashes, and the patch's own sha256 — because the useful diagnostic is *which* of the two sites disagrees.
- Two new tests, one per direction. The fixture lockfile carries both spellings; the second test starts from a lockfile whose `patchedDependencies` is already correct and asserts `--check` still fails, which is exactly the case the old predicate passed.
- Rule 6 in `docs/reference/xterm-patch-regeneration.md` states the operational consequence: `--check` is the authority on the lockfile, not `pnpm install`. Always finish on step 4.

Both fix PRs' lockfiles were repaired per-commit, so each commit's `patchedDependencies`, its 23 resolution keys, and the sha256 of the patch file it carries all agree at that commit — verified individually rather than only at the tip.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — `regenerate-xterm-patches.test.mjs` (new, 27 tests) and `pr-workflow-parallelism.test.mjs` pass
- [x] `node config/scripts/regenerate-xterm-patches.mjs --check` against an empty work dir — full clone, `npm ci`, two builds, byte-identical
- [x] `pnpm install --frozen-lockfile` cold, then the terminal-pane suite (219 files / 2,902 tests) against the resulting install
- [ ] `pnpm build` — not run locally; the CI job exercises the regeneration path

On test assurance, stated plainly: the unit tests cover the pure pieces — diff normalization, hunk splitting, round-trip stability, the commit and build-order assertions, the sourcemap-policy guard, and lockfile coupling. They cannot prove the rebuild reproduces upstream; only the CI job that actually clones and builds does that. The runtime test in this PR asserts the version banner the patched bundle emits, which reads the bundle itself; provenance against the pinned commit stays the generator's job, since `src/common/Version.ts` is the one source file the generator deliberately exempts from comparison (upstream's publish step rewrites it).

## AI Review Report

Reviewed for cross-platform correctness of the regeneration script, which is the main risk here since it shells out and manipulates paths.

- Path handling uses Node path utilities rather than assuming `/`.
- The script is invoked through `pnpm`/Node, not a shell-specific pipeline, so it runs on Windows as well as macOS/Linux.
- The `-text` attribute above is the Windows-specific correctness fix, not a formatting preference.
- No hardcoded platform separators or `metaKey`-style assumptions (no UI surface in this PR).

Risk checked and accepted: pinning an upstream beta means a future upstream bump is a deliberate, reviewable step rather than an implicit one. That is the intent.

Risk checked and accepted: nothing in this repo consumes the xterm bundle sourcemaps at build or run time, so deleting them costs no debuggability that was actually available — the maps that shipped before were already only as good as their offsets.

## Security Audit

- **Dependency**: pins a specific upstream `@xterm/xterm` revision and verifies the generated patch byte-for-byte in CI, which *narrows* the supply-chain surface compared to a hand-edited patch.
- **Command execution**: the regeneration script runs a pinned upstream build; it takes no user-controlled input.
- No auth, secrets, or IPC surface touched. `pnpm-lock.yaml` changes are the patch-hash update and its peer-suffix propagation only.

## Notes

Base: #11893. This PR exists to make the two fix PRs above it verifiable — reviewers of #11896/#11897 can read the actual patched `CompositionHelper.ts` source.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,8 @@
 /src/cli/bundled-skill-guides.ts text eol=lf
 # Bundled plugin trees are byte-hashed; CRLF checkout would break the pinned hash.
 /resources/plugins/** text eol=lf
+# pnpm hashes this generated patch byte-for-byte, so a CRLF checkout breaks the install;
+# its minified bundle lines also make a diff nobody can read. Review the hand-written
+# source patch under xterm-src/ instead.
+/config/patches/@xterm__xterm@*.patch -diff -text
+/config/patches/xterm-src/*.patch text eol=lf
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -173,6 +173,34 @@ jobs:
           done
           exit "$status"
 
+  xterm_patch_sync:
+    name: xterm patch sync
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - uses: ./.github/actions/install-node-dependencies
+
+      # Why: the check rebuilds xterm.js from a pinned upstream commit. Caching the
+      # npm metadata and the shallow clone turns a ~4 min cold run into well under a
+      # minute; the key is the manifest, so a commit or toolchain bump invalidates it.
+      - name: Restore upstream xterm build inputs
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.npm
+            ${{ runner.temp }}/xterm-patch-build/upstream/.git
+          key: xterm-upstream-${{ hashFiles('config/patches/xterm-upstream.json') }}
+
+      - name: Verify xterm patches match the pinned upstream build
+        env:
+          WORK_DIR: ${{ runner.temp }}/xterm-patch-build
+        run: node config/scripts/regenerate-xterm-patches.mjs --check --work-dir="$WORK_DIR"
+
   shell_contracts:
     name: shell contracts
     runs-on: ubuntu-latest
@@ -397,6 +425,7 @@ jobs:
       - root_directory_guard
       - typecheck
       - git_compatibility
+      - xterm_patch_sync
       - shell_contracts
       - test
       - package
@@ -418,6 +447,7 @@ jobs:
           ROOT_DIRECTORY_GUARD: ${{ needs.root_directory_guard.result }}
           TYPECHECK: ${{ needs.typecheck.result }}
           GIT_COMPATIBILITY: ${{ needs.git_compatibility.result }}
+          XTERM_PATCH_SYNC: ${{ needs.xterm_patch_sync.result }}
           SHELL_CONTRACTS: ${{ needs.shell_contracts.result }}
           TEST: ${{ needs.test.result }}
           PACKAGE: ${{ needs.package.result }}
@@ -428,6 +458,7 @@ jobs:
             "$ROOT_DIRECTORY_GUARD" \
             "$TYPECHECK" \
             "$GIT_COMPATIBILITY" \
+            "$XTERM_PATCH_SYNC" \
             "$SHELL_CONTRACTS" \
             "$TEST" \
             "$PACKAGE" \
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,7 @@ docs/**
 !docs/reference/headless-linux-server.md
 !docs/reference/linux-glibc-compatibility.md
 !docs/reference/relay-grace-time-reconfiguration.md
+!docs/reference/xterm-patch-regeneration.md
 
 # Stably CLI (only docs/ are tracked)
 .stably/*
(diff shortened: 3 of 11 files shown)


nwparker:
Leaving this open — not superseded. Build reproducibility for the xterm patch is independent of who owns composition, and main still carries a hand-maintained patch (`config/patches/@xterm__xterm@6.1.0-beta.287.patch`, now 355 lines) with no pinned-upstream provenance.

That patch grew during this work and is load-bearing: it reworks the `_inputEvent` / `_keyDownSeen` gate, which is the reason the text system's output reaches the terminal at all. A 355-line patch we cannot regenerate from a known upstream build is a real maintenance liability at the next xterm bump.

One file here (`terminal-ime-xterm-transaction-events.test.ts`) went away with #13128, so it needs a rebase, but the build change stands on its own.

Issue #6905, opened 2026-06-30
[Bug]: Vietnamese IME Input Broken in Terminal


dony-omg:
### Operating system

macOS

### Orca version

*No response*

### Details

**Short summary:**<br>Typing Vietnamese characters using an IME (Input Method Editor) in Orca's integrated terminal produces incorrect or garbled output. Characters are not composed correctly, and the input method composition window does not work as expected.

**What happened?**<br>When attempting to type Vietnamese text using a Vietnamese IME (e.g., Unikey, iBus Bamboo, or fcitx-bamboo) in Orca's integrated terminal, the composed characters are either:

* Displayed incorrectly (e.g., wrong diacritical marks, wrong tone marks)
  * Duplicated or repeated unexpectedly
  * The raw key sequences are shown instead of the composed Vietnamese characters
  * The IME composition overlay does not appear or behaves erratically inside the terminal<br>**How can we reproduce it?**

1. Install a Vietnamese IME (e.g., iBus Bamboo on Linux, Unikey on Windows)
   2. Open Orca and navigate to the integrated terminal
   3. Switch the input method to Vietnamese (e.g., Telex or VNI mode)
   4. Try typing a Vietnamese word such as "xin chào" (hello) or "tiếng việt"
   5. Observe that the characters are garbled, duplicated, or not composed correctly<br>**Additional context:**

* This issue does not occur in other terminal emulators (e.g., Windows Terminal, Alacritty, GNOME Terminal) with the same IME configuration.
  * The bug makes Orca's terminal unusable for Vietnamese speakers who rely on IME input.
  * Expected behavior: Vietnamese characters should be composed and displayed correctly, consistent with how other terminal emulators handle IME input.

nwparker:
**Status: NOT REPRODUCED on current tree** (final runtime confirmation needs a live macOS host with a Vietnamese IME).

[...]

Issue #6698, opened 2026-06-29
[Bug]: Vietnamese Telex input in integrated terminal loses characters (e.g. "chính" → "ch")


phong03bn:
### Operating system

macOS

### Orca version

v1.4.104

## Description

When typing Vietnamese text with diacritics in Orca's integrated terminal, characters are lost or corrupted as soon as I add the tone mark at the end.

A typical example using the word “chính”:

- Using a Vietnamese Telex input method, I type: `chinhs` (which should produce the Vietnamese word “chính”)
- However, in Orca’s terminal the final result is only `ch` — the `ính` part disappears after the IME commits the composition.

This happens for essentially any Vietnamese word typed with Telex: if I type the full Telex sequence (letters + tone key) so that it should become a Vietnamese word with diacritics, parts of the word are removed or changed once the IME finalizes the composed text.

This makes it effectively impossible to type normal Vietnamese text in the Orca terminal.

## Steps to Reproduce

1. Open Orca and create/open a local terminal.
2. Enable a Vietnamese Telex input method on the OS (for example: Unikey/EVKey on Windows, or the built‑in Vietnamese IME on macOS/Linux) and set it to Telex mode.
3. In the Orca terminal, type the Telex sequence for the word “chính”: `chinhs`.
4. Observe what is rendered in the terminal.

## Actual Behavior

- The terminal only shows `ch`, or the rest of the word is otherwise lost/incorrect after the IME commits the composed text.
- Similar corruption/loss happens for other Vietnamese words typed with Telex.

## Expected Behavior

- The terminal should correctly display the full Vietnamese word with diacritics, for example `chính`, and in general preserve all characters when the IME commits the composition.

## Additional Info

- This only happens inside Orca’s integrated terminal. Typing Vietnamese with the same IME (Telex) works normally in other apps and terminals outside Orca.
- The problem appears to be related to how Orca/xterm handles IME composition and updates the buffer when the composed text is committed (especially for multi‑code‑point sequences used in Vietnamese).

nwparker:
**Status: NOT REPRODUCED (fixed after report)**

[...]

PR #11895, opened 2026-08-01
test(ime): add a recorded-trace harness and one composition predicate

nwparker:
## Summary

Collapses the app's scattered IME guards onto one predicate, and adds the trace harness that lets IME behavior be tested without an IME.

Before this, each surface open-coded its own composition check. They disagreed: some tested only `isComposing`, some only `keyCode === 229`, some nothing at all. `isImeCompositionKeyDown` is now the single place that decides a keystroke belongs to an IME, and 14 call sites migrate to it.

The predicate checks an app-wide composition flag *as well as* the per-event bits, because the two cover opposite gaps. `keyCode === 229` (Windows) and `key === 'Process'` (Android) mark a key the IME consumed, and 229 is also what catches Safari's extra keydown *after* `compositionend` — the flag is already cleared by then, so it cannot help there. The flag covers the reverse: keys arriving *during* a live composition that carry neither marker.

The app-wide flag tracks the composing *target*, not a bare boolean, and revalidates on read that the target is still connected. Clearing on blur is the usual answer, because browsers do not reliably deliver `compositionend` when the composing element loses focus — but blur alone is not enough here: Chromium fires no `focusout` when a *focused* node is removed, so a composer that unmounts mid-composition would latch the flag on for the rest of the session and silently kill every Enter-to-commit surface in the app. Editors that scope composition state per editor instance escape this, since the state dies with the editor; a document-scoped flag has to revalidate instead.

`shouldSuppressEnterSubmit` now delegates to the same predicate, which brings the three remaining `TaskPage.tsx` sites in with it. An earlier revision of this PR described those three as telemetry-only reads; that was wrong — they are Enter-submit gates, so bare `isComposing` was missing an IME-owned Enter on Windows and Android.

## Escape is exempt from the derived flag

The flag is derived state, and `compositionend` is not guaranteed to arrive. When it drifts, every key it guards goes dead — and eleven call sites guard Escape *above* their key dispatch, so the casualties are the rename fields, search boxes, pickers and panels whose only way out is Escape.

An Escape the IME owns should arrive marked (`isComposing`, `229`, or `'Process'`) and is still suppressed by the clauses above. An **unmarked** Escape therefore means the flag has drifted, and swallowing it strands the user behind a flag nothing is going to clear. So the flag no longer applies to Escape; the per-event markers still do.

How well-evidenced that premise is, stated plainly: **no trace in this stack contains a mid-composition Escape**, recorded or derived, so the marking is argued rather than shown. The nearest evidence is the candidate-navigation trace, where mid-composition `ArrowDown`/`ArrowUp` carry `isComposing: true` with their own keyCodes (40/38) rather than 229 — exactly the shape the argument needs — but that trace is `derived`. The single `recorded` trace contains no mid-composition non-`Process` key at all.

The cost of being wrong is also not symmetric with the terminal's, which is why it is worth naming rather than burying. `xterm-bypass-policy.ts` accepts "worst case one extra keypress" for the same rule — benign. Here the Escape handlers run `onClear()` and `onClose()`, so an IME-owned-but-unmarked Escape would discard a search query or close a panel while the user was only dismissing a candidate window. Before this change that case was a silent no-op. The trade is still the right one — the drift outage is concrete and reproducible, this case is hypothetical — but it trades a no-op for a possible wrong action, not a free win. #11927's recorder settles it: one macOS or Windows recording of Escape during a candidate window turns this from derived reasoning into evidence.

This is the rule `xterm-bypass-policy.ts` already applies in the terminal, where Escape is deliberately absent from `TERMINAL_IME_OWNED_KEYS` for exactly this reason. Before this change the two disagreed about the same keystroke depending on which pane had focus. Fixing it in the predicate rather than at the call sites is what keeps them in agreement — patching eleven handlers would have re-created the scatter this PR exists to remove.

Recovery was already bounded rather than permanent, since the target revalidation above clears the flag on the next focus change. The failure was "Escape and Enter silently do nothing until you click away and back," with nothing telling the user that is the fix.

## Trace harness and provenance

`ime-composition-trace.test-fixtures.ts` replays an event sequence against a real input element, stamping each event's observed buffer state in *before* dispatching it, so the code under test sees byte-for-byte what a browser produced.

Each trace declares its `provenance`:

- `recorded` — captured from a real engine, authoritative.
- `derived` — reconstructed from a documented event ordering. Pins current behavior, but cannot prove the ordering is what the engine really emits.

**Six traces: five `derived`, one `recorded`.** Read the `provenance` field, not this paragraph — but the shape of the set is worth stating up front, because the five derived ones are not all derived the same way.

| trace | provenance | derived from |
| --- | --- | --- |
| `IBUS_HANGUL_TERMINAL_JAMO_COMMIT_TRACE` | **`recorded`** | a live ibus-hangul engine, captured by `tests/e2e/terminal-ime-boundary-probe.ts` on ubuntu-22.04 (run 30691057123) |
| `IBUS_HANGUL_RETAINED_COMMIT_TRACE` | `derived` | the synthetic sequence `tests/e2e/terminal-ime-observed-event-sequences.ts` replays in CI |
| `IBUS_HANGUL_MIXED_LATIN_TRACE` | `derived` | same |
| `TELEX_BACKSPACE_MID_COMPOSITION_TRACE` | `derived` | the reported symptom in orca#6494 / #6698 / #6905 |
| `CANDIDATE_ARROW_NAVIGATION_TRACE` | `derived` | the reported symptom in orca#9803 |
| `CANDIDATE_ESCAPE_DISMISSAL_TRACE` | `derived` | the Escape exemption in `ime-composition-keyboard-event.ts` |

Two of the derived traces are transcriptions of sequences CI hand-dispatches rather than captures, with selection offsets reconstructed as a collapsed caret at end-of-buffer. The other three are reconstructed from a documented event ordering around a real reported bug — they pin the behavior that was fixed, but nothing in them proves the engine emits that exact ordering.

Promotion is **not** blocked on someone's physical hardware, and an earlier revision of this paragraph wrongly said it was. The hangul trace above was recorded by CI, on a runner, from a real engine — that is what the IBus engine matrix higher in this stack exists to make routine, and `CANDIDATE_ESCAPE_DISMISSAL_TRACE` names `config/scripts/record-ime-trace.mjs` as its promotion path. What is genuinely still missing is coverage for engines the Linux matrix cannot host: macOS-native and Windows IMEs. Those need a human at the keyboard.

The trace assertions compare an interpretation derived from event *data* against each trace's declared final state, per the AGENTS.md rule that committed text comes from ranges and event data, never from diffing two observations of a mutable buffer. An earlier revision asserted the textarea's final contents instead, which was vacuous: the harness stamps that state in itself, so the assertion held no matter what any listener did.

Three fields the harness could not previously carry:

| field | why it has to survive replay |
| --- | --- |
| `isComposing: undefined` | The field is typed `boolean \| undefined` precisely so a Safari trace can be told from a Chromium one. The replayer coalesced it to `false` on construction and the repair function declined to touch it, so the one distinction it exists to express could not be reproduced. |
| `repeat` | The accent panel a macOS press-and-hold opens is recognisable by nothing else. Without it, the two held keydowns replay identically to the first press. |
| `selectionDirection` | Absent it, a preedit replacement range can only ever be represented as a collapsed caret, never a backward selection. |

The first is the one worth flagging: `undefined` is a *recorded value*, not a missing one, and constructing an `InputEvent` coerces it. The replayer now defines the property explicitly, the same way it already patches the legacy `keyCode`.

## Screenshots

No visual change — this preserves existing behavior at every migrated call site and fixes the sites that had no guard.

## Testing

- [x] `pnpm lint` — oxlint 0 warnings / 0 errors
- [x] `pnpm typecheck`
- [x] `pnpm test` — predicate + trace-replay suites pass
- [x] `pnpm audit:code-quality:type-aware`
- [ ] `pnpm build` — not run
- [x] The unmount-latch regression test was confirmed to fail without its fix (`expected true to be false`) and pass with it
- [x] The trace interpreter was confirmed non-vacuous by mutating a trace's declared `compositionend` data: the new assertions failed, the old one still passed
- [x] All four new tests were run against the unfixed sources and each failed with the predicted error — `expected true to be false` (Escape), `expected false to be undefined` (`isComposing`), `expected false to be true` (`repeat`), `expected [ 'none' ] to deeply equal [ 'backward' ]` (selection direction)
- [x] Full renderer component suite (1,405 files / 11,733 tests) green, which is what covers the eleven Escape call sites the predicate change reaches

## AI Review Report

A subagent audited the predicate against the behavior real engines produce rather than against what the spec implies. Four properties a correct composition guard needs, all reproduced here:

- The same three-signal check (`isComposing`, `229`, `'Process'`) rather than `isComposing` alone.
- The guard placed at the top of keydown, before any per-key branching.
- A derived app-wide composition flag alongside the per-event bits, for the keys that carry no marker.
- Composition state cleared when the composing element goes away — which is where the unmount gap above came from.

One deliberate design choice, now documented in the code: the `229`/`'Process'` markers are unconditional, with no opt-out argument. `xterm-bypass-policy.ts` lets a bare `229` keydown through on macOS/Linux, so the two look inconsistent — a review pass raised exactly that. The divergence is correct, but the original justification ("only the terminal needs it") named the wrong reason, and the comment now states the right one: the terminal needs the keystroke because xterm's `CompositionHelper` reads the committed glyph off the helper textarea, whereas callers of this predicate receive committed text through `input`. A marked keydown carries nothing they could act on — its `key` is `'Process'` or `'Unidentified'`, never the glyph — so suppressing it costs no input.

The one call site that looked like a counterexample, the emulator screen keyboard, was checked and is not one: it forwards through an ASCII/named-key USB HID table, which returns `null` for `'Process'` and drops the keystroke identically whether or not the guard fires. At one site suppressing is better than merely harmless — the address picker concatenates `event.key` into a typeahead prefix, which `'Process'` would have poisoned.

Cross-platform: the predicate is explicitly the union of the macOS, Windows (`229`), and Android (`'Process'`) markers; no platform branch, because all three markers are safe to check everywhere.

## Security Audit

- **Input handling**: this is keyboard-event classification only. It never reads or forwards event *content*, so there is no injection surface.
- No command execution, path handling, auth, secrets, dependency, or IPC changes.
- Note the guard *widens* suppression at previously-unguarded sites, which can only prevent actions, never trigger new ones.

## Notes

Base: #11894. The trace harness added here is the test substrate for #11896 and #11897.

### DIFF
--- a/src/renderer/src/components/JiraIssueWorkspace.tsx
+++ b/src/renderer/src/components/JiraIssueWorkspace.tsx
@@ -29,6 +29,7 @@ import {
   hasBoundedCommentBodyText
 } from '@/lib/comment-body-submit-state'
 import { createBrowserUuid } from '@/lib/browser-uuid'
+import { isImeCompositionKeyDown } from '@/lib/ime-composition-keyboard-event'
 import { useAppStore } from '@/store'
 import {
   jiraAddIssueComment,
@@ -609,7 +610,7 @@ export default function JiraIssueWorkspace({
                         value={titleDraft}
                         onChange={(event) => setTitleDraft(event.target.value)}
                         onKeyDown={(event) => {
-                          if (event.key === 'Enter' && !event.nativeEvent.isComposing) {
+                          if (event.key === 'Enter' && !isImeCompositionKeyDown(event)) {
                             event.preventDefault()
                             handleSaveTitle()
                           }
--- a/src/renderer/src/components/TaskPage.tsx
+++ b/src/renderer/src/components/TaskPage.tsx
@@ -42,6 +42,7 @@ import { useAllWorktrees, useRepoMap } from '@/store/selectors'
 import { callRuntimeRpc, getActiveRuntimeTarget } from '@/runtime/runtime-rpc-client'
 import { getLocalPreflightContext, localPreflightContextKey } from '@/lib/local-preflight-context'
 import { getProviderRuntimeContextKey } from '@/lib/provider-runtime-context'
+import { isImeCompositionKeyDown } from '@/lib/ime-composition-keyboard-event'
 import {
   getSettingsFocusedExecutionHostId,
   parseExecutionHostId,
@@ -6482,12 +6483,7 @@ export default function TaskPage(): React.JSX.Element {
     (event: React.KeyboardEvent<HTMLInputElement>): void => {
       if (event.key === 'Enter') {
         // React SyntheticEvent does not expose isComposing; use nativeEvent.
-        if (
-          shouldSuppressEnterSubmit(
-            { isComposing: event.nativeEvent.isComposing, shiftKey: event.shiftKey },
-            false
-          )
-        ) {
+        if (shouldSuppressEnterSubmit(event.nativeEvent, false)) {
           return
         }
         event.preventDefault()
@@ -8622,15 +8618,7 @@ export default function TaskPage(): React.JSX.Element {
                             onChange={(e) => setLinearSearchInput(e.target.value)}
                             onKeyDown={(e) => {
                               if (e.key === 'Enter') {
-                                if (
-                                  shouldSuppressEnterSubmit(
-                                    {
-                                      isComposing: e.nativeEvent.isComposing,
-                                      shiftKey: e.shiftKey
-                                    },
-                                    false
-                                  )
-                                ) {
+                                if (shouldSuppressEnterSubmit(e.nativeEvent, false)) {
                                   return
                                 }
                                 e.preventDefault()
@@ -8806,12 +8794,7 @@ export default function TaskPage(): React.JSX.Element {
                           onChange={(e) => setJiraSearchInput(e.target.value)}
                           onKeyDown={(e) => {
                             if (e.key === 'Enter') {
-                              if (
-                                shouldSuppressEnterSubmit(
-                                  { isComposing: e.nativeEvent.isComposing, shiftKey: e.shiftKey },
-                                  false
-                                )
-                              ) {
+                              if (shouldSuppressEnterSubmit(e.nativeEvent, false)) {
                                 return
                               }
                               e.preventDefault()
@@ -11034,7 +11017,7 @@ export default function TaskPage(): React.JSX.Element {
                 value={newIssueTitle}
                 onChange={(e) => setNewIssueTitle(e.target.value)}
                 onKeyDown={(e) => {
-                  if (e.key === 'Enter' && !e.nativeEvent.isComposing) {
+                  if (e.key === 'Enter' && !isImeCompositionKeyDown(e)) {
                     e.preventDefault()
                     void handleCreateNewIssue()
                   }
@@ -11208,7 +11191,7 @@ export default function TaskPage(): React.JSX.Element {
               value={newLinearProjectName}
               onChange={(event) => setNewLinearProjectName(event.target.value)}
               onKeyDown={(event) => {
-                if (event.key === 'Enter' && !event.nativeEvent.isComposing) {
+                if (event.key === 'Enter' && !isImeCompositionKeyDown(event)) {
                   event.preventDefault()
                   void handleCreateNewLinearProject()
                 }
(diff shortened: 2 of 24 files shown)


nwparker:
Superseded by #13128 (17b3dff3c41) — the composition predicate it pairs with is gone. The recorded-trace discipline itself was the right instinct and is what the surviving tests use; see #11927 for the recorder, which I have left open.

PR #11896, opened 2026-08-01
fix(ime): stop dropping and splitting composition commits in the terminal

nwparker:
## Summary

Three terminal IME bugs, all of which corrupt committed CJK text.

**1. Dropped commits after a transport swap.** The composition route delivered a finished commit to whatever transport was current at delivery time. If the pty was swapped mid-composition, the text landed on the wrong pty. The route now checks transport identity and ownership before delivering, and deliberately drops when the captured transport is gone — that drop is the route's decision to make, rather than silently writing to a stranger.

**2. Newlines split the preedit.** The deferred-newline fallback called `finish` directly, bypassing the pending-composition re-check the event path uses. A composition still open at 200 ms got a newline sent into the middle of its preedit — you press Enter, and your half-converted Japanese gets cut in two with a line break through it.

The deadline now re-arms while a composition is pending. Its 2 s ceiling is measured **from the last `compositionupdate`, not from the start**: multi-segment bunsetsu conversion legitimately keeps a Japanese composition pending far longer than any fixed total-elapsed budget, so a wall-clock ceiling would reintroduce exactly the split it is meant to prevent. A composition that stops progressing still cannot strand the keystroke.

**3. Compositions already in flight when the route installs.** The re-arm asked the route whether a composition was pending, but a route installed mid-composition never saw that session start, so it had nothing recorded to report — and the deadline split that composition too. A live `compositionupdate` now counts as evidence in its own right.

## Screenshots

No visual change in the UI chrome. The user-visible effect is in terminal text: committed CJK no longer lands on the wrong pty or gets bisected by a newline.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 2905 passing across the terminal-pane and IME suites
- [ ] `pnpm build` — not run

On the idle ceiling specifically: the new test (`holds the newline through a long conversion that keeps updating its preedit`) was verified to **fail** without the fix and pass with it, so it is a real regression test rather than a tautology.

## AI Review Report

A review subagent found the wall-clock ceiling bug in my own first version of this PR — the original 2 s budget was absolute, which would split any conversion running longer than 2 s. A total-elapsed ceiling is the wrong shape for this: composition length is a property of what the user is typing, not of anything going wrong. The bound that is actually meaningful is idle time — a composition that has stopped progressing. Fixed here, with the regression test above.

Also verified while fixing it: xterm's own `CompositionHelper.blur():245-247` already finalizes on blur, so the original "compositionend is not guaranteed on focus loss" rationale was not the real justification for the ceiling. The comment now states the actual one — a composition that stops progressing.

Cross-platform: the deferral is driven by DOM composition events and the xterm session events, both platform-neutral. SSH and folder-workspace cases are unaffected — the route keys on transport identity, which is the same abstraction for local, SSH, and relay ptys, and the drop-on-missing-transport path is what makes an SSH reconnect safe.

## Security Audit

- **IPC / transport**: this is the security-relevant part, and it improves. Previously a commit could be written to a transport other than the one that owned the composition; the ownership + identity check closes that. Writing input to the wrong pty is a real confidentiality concern when the swapped-in pty is a different host or a different user's shell over SSH.
- **Input handling**: composed text is passed through unmodified. Per the IME rules it is deliberately **not** normalized at commit — normalization happens only at path handling and equality comparison, so no normalization-based bypass is introduced here.
- No command execution, auth, secrets, or dependency changes.

## Notes

Base: #11895. Depends on #11894 for the readable `CompositionHelper` source.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-ime-composition-route.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-composition-route.test.ts
@@ -149,6 +149,35 @@ describe('installTerminalImeCompositionRoute', () => {
     expect(harness.input).not.toHaveBeenCalled()
   })
 
+  it('leaves an uncaptured session to xterm when installed mid-composition', () => {
+    const harness = createHarness()
+
+    // No start was seen, so nothing here can deliver the text. Cancelling the event
+    // would suppress xterm's own insertion and drop the commit on the floor.
+    const notPrevented = harness.end(1, '한')
+
+    expect(notPrevented).toBe(true)
+    expect(harness.input).not.toHaveBeenCalled()
+  })
+
+  it('suppresses xterm insertion only for a session it captured', () => {
+    const harness = createHarness()
+    harness.start(1)
+
+    expect(harness.end(1, '한')).toBe(false)
+    expect(harness.input).toHaveBeenCalledWith('한')
+  })
+
+  it('keeps suppressing insertion for a captured session it deliberately drops', () => {
+    const harness = createHarness()
+    harness.start(1)
+    harness.state.currentTransport = createTransport('pty-replacement')
+
+    // Owned, so xterm must still stand down — the drop is this route's decision.
+    expect(harness.end(1, '한')).toBe(false)
+    expect(harness.input).not.toHaveBeenCalled()
+  })
+
   it.each(['terminal close', 'tab unmount'])(
     'releases the captured session and listeners on %s',
     () => {
--- a/src/renderer/src/components/terminal-pane/terminal-ime-composition-route.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-composition-route.ts
@@ -82,11 +82,13 @@ export function installTerminalImeCompositionRoute(args: {
     if (!detail) {
       return
     }
-    event.preventDefault()
     const captured = sessions.get(detail.id)
     if (!captured) {
+      // Not our session — the route was installed mid-composition, so no start was seen.
+      // Preventing default here would suppress xterm's insertion with nothing to replace it.
       return
     }
+    event.preventDefault()
     sessions.delete(detail.id)
     adjustPendingCompositionCount(terminalElement, -1)
     if (
--- a/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
@@ -116,6 +116,114 @@ describe('sendTerminalInputAfterComposition', () => {
     expect(send).toHaveBeenCalledTimes(1)
   })
 
+  it('re-arms past the fallback rather than splitting a still-pending composition', () => {
+    const el = document.createElement('div')
+    const send = vi.fn()
+    const transport = { getPtyId: () => 'pty-1' } as unknown as PtyTransport
+    const route = installTerminalImeCompositionRoute({
+      terminalElement: el,
+      terminal: { input: vi.fn() },
+      capturedTransport: transport,
+      getCurrentTransport: () => transport
+    })
+    el.dispatchEvent(
+      new CustomEvent(XTERM_COMPOSITION_SESSION_START_EVENT, { detail: { id: 1, data: '한' } })
+    )
+
+    sendTerminalInputAfterComposition(el, send, { fallbackMs: 50, maxIdleMs: 1000 })
+    vi.advanceTimersByTime(500)
+
+    // A slow candidate window is not a reason to send Enter into the middle of the preedit.
+    expect(send).not.toHaveBeenCalled()
+
+    el.dispatchEvent(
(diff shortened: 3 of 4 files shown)


holdn2:
Heads up — I've opened #12120 as a draft that ports **part** of `ab2bf67` from this PR,
and I want to be explicit about which part.

Taken: only the deferred-newline fallback re-arm — the `compositionupdate` evidence, the
idle ceiling measured from the last update rather than from the start, and its four tests.
Attributed with a `Co-authored-by:` trailer since it is a partial port rather than a
cherry-pick.

Not taken: the composition-route ownership/`preventDefault` reordering and the
transport-identity work. Those are a larger change than the combined branch needed, and
they're better reviewed here on their own merits.

I kept your bunsetsu rationale in the comments verbatim, since the idle-vs-total-elapsed
distinction is the whole point and I didn't want it lost.

Not meant to supersede this PR. Happy to close mine, and glad to be corrected if I've
drawn the line in the wrong place.

holdn2:
Heads up now that #12265 has landed: the fallback re-arm from this PR is the one piece of
the Windows Korean IME work still not on `main`.

`terminal-ime-deferred-newline.ts` at `main` is unchanged on that path:

```ts
const fallbackTimer = window.setTimeout(finish, fallbackMs)   // line 45
```

`finish` still fires directly at 200 ms without the pending-composition re-check the event
path uses, so a composition still open at the deadline gets a newline dropped into the
middle of its preedit. None of `maxIdleMs`, `onFallbackDeadline`, `sawRecentUpdate` or
`TERMINAL_IME_DEFERRED_NEWLINE_MAX_IDLE_MS` exist there. #12265 closed the keydown and
keyup synthesis paths, and the xterm commit-ordering work landed separately, but neither
touches this one.

I had ported just that hunk into #12120 — the `compositionupdate` evidence and the idle
ceiling measured from the last update rather than from the start, with your four tests and
a `Co-authored-by:` to you. I am closing that PR now that the overlap has landed, so
nothing of mine is competing with this one. If it is useful, the four tests apply cleanly
on top of `main` as they are:

- pending composition is not split at `fallbackMs`
- a long conversion that keeps updating its preedit is held past the ceiling
- a composition already in flight with no recorded session start is still held
- a stalled composition sends exactly once at the ceiling

Happy to leave it entirely to you — this is your hunk and your rationale, and the
idle-vs-total-elapsed distinction in your comments is the part that matters. Just flagging
that it is now the remaining gap rather than one of several.

dogfooter8750:
Validation data for this draft, since it is the last piece of #11878 still open on `main` and @holdn2 flagged it when closing #12120.

I measured **when** the deferred newline actually fires, on `main` at `13f033f0` and with this branch applied, driving `sendTerminalInputAfterComposition` directly with fake timers.

Two scenarios. In both, a newline is deferred at t=0 with a composition already open.

- **A — stalled composition**: nothing further happens. This is the case the ceiling exists for.
- **B — the user keeps typing**: six syllables at ~120 ms apart, each dispatching `compositionupdate` and rolling the session over (`session-start` for the next syllable, `session-end` for the previous). Typing finishes at t=720 ms.

```text
                            main (13f033f0)      this branch
A  stalled composition          300 ms             2100 ms
B  typing through 720 ms        240 ms             2620 ms
```

**B is the one that matters.** On `main` the newline lands at 240 ms — the user is four syllables from finishing the word, and the bytes go out mid-composition. That is exactly the "newline appears in the middle of a word" shape reported on #11878, and it is still live on `main` today. With this branch the deferral rides out the whole run and releases at 2620 ms, after typing has stopped.

**A is not a regression.** 2100 ms is the idle ceiling doing its job. Measuring idle from the last `compositionupdate` rather than from the defer is the right call — it means a genuinely stuck IME still releases the user's newline instead of swallowing it, while an active composition is never interrupted. The Japanese bunsetsu note in your comment is the case that would break under a total-elapsed ceiling, and this avoids it.

`terminal-ime-deferred-newline.test.ts` and `terminal-ime-composition-route.test.ts` pass with the branch applied (58 tests) on Windows.

Caveat on what this is and is not: this is unit-level with fake timers, so it establishes the ordering and the deadline behaviour, not real-device timing. I have a Windows 11 + Microsoft Korean IME (2-beolsik) machine and can run the real-device pass against a build of this branch whenever one exists — that was the gap that held up the predicate fix too, and I would rather not have it hold this one up as well.

Related, in case it saves someone a search: #11983 stacks the defer-time session capture on top of this, which covers the other half I raised on #11878 — a session that starts *after* the defer holding the newline hostage. Between the two the fallback path looks fully addressed; nothing further from me.

nwparker:
Duplicate of #12942 and obsolete for the same reason: all four files it touches were deleted by #13128 (17b3dff3c41). The deferral mechanism was the bug — on 1.4.171 it wrote a Shift+Enter payload to the PTY with no Enter pressed, 10/10, at 203-218 ms after keydown. If commit-splitting still reproduces on current main, please reopen with a trace.

PR #11903, opened 2026-08-01
ci: block new root-level entries

nwparker:
## Summary

- Add a PR check that compares the base and head repository trees.
- Reject newly introduced root-level files and top-level directories with a clear GitHub landing-page message.
- Include the check in the aggregate `verify` gate and cover allowed/rejected cases with fixtures.

## Testing

- `pnpm exec vitest run --config config/vitest.config.ts config/scripts/check-root-directory-entries.test.mjs config/scripts/pr-workflow-parallelism.test.mjs config/scripts/pr-workflow-lint-parity.test.mjs config/scripts/pr-e2e-gate-contract.test.mjs` (23 passed)
- `pnpm typecheck`
- `bash -n .github/scripts/check-root-directory-entries.sh`
- `git diff --check`

### DIFF
--- a/.github/scripts/check-root-directory-entries.sh
+++ b/.github/scripts/check-root-directory-entries.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ $# -ne 2 ]]; then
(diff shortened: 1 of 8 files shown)


AmethystLiang:

PR #11905, opened 2026-08-01
fix(lint): merge duplicate benchmark imports that fail static analysis

nwparker:
## Summary

`static analysis` runs `oxlint --config config/oxlint-code-quality-native-plugins.json src config tests mobile --deny-warnings`. Two cold-park benchmarks each import `wsl-workspace-fixture.mjs` twice, which trips `import/no-duplicates`. With `--deny-warnings` that is a hard failure, so **every PR opened against main currently fails static analysis**.

Trunk stayed green because the push workflow does not run that job — only the PR workflow does. So this only shows up on PRs, which is why it went unnoticed.

Merges each pair into a single import statement. No behavior change.

Found while getting an unrelated IME stack (#11899) green; it was failing the same check on a docs-only PR, which is what made it obvious the cause was inherited from main rather than introduced.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — `audit:code-quality:native` now exits 0 over `src config tests mobile`
- [x] `pnpm typecheck`
- [x] `pnpm test` — untouched; these are benchmark entrypoints, not test files
- [ ] `pnpm build` — not run; no app code touched
- [x] Added or updated high-quality tests, or explained why not — a lint rule already enforces this; a test would be redundant with the rule

## AI Review Report

Scope is two merged import statements in benchmark scripts. Verified that the merged import lists the same bindings (`createCompletedOnboardingProfile`, `safeRemoveLocalDirectory`) from the same specifier, so resolution is identical. `oxfmt` applied.

Cross-platform: these are Node `.mjs` benchmark entrypoints; the change is import syntax only and touches no path handling, shell invocation, or platform branch.

## Security Audit

No security surface. Import-statement consolidation in developer benchmark scripts — no input handling, command execution, path handling, auth, secrets, dependency, or IPC changes.

## Notes

Unblocks PR CI repo-wide, not just the IME stack.

### DIFF
--- a/tests/tools/benchmarks/terminal-cold-park-resource-bench.mjs
+++ b/tests/tools/benchmarks/terminal-cold-park-resource-bench.mjs
@@ -28,14 +28,16 @@ import {
   stopDevApp,
   waitForStoreReady
 } from '../../config/scripts/windows-apphang-repro/electron-dev-session.mjs'
-import { createCompletedOnboardingProfile } from '../../config/scripts/windows-apphang-repro/wsl-workspace-fixture.mjs'
+import {
+  createCompletedOnboardingProfile,
+  safeRemoveLocalDirectory
+} from '../../config/scripts/windows-apphang-repro/wsl-workspace-fixture.mjs'
 import {
   pollUntil,
   rendererActionTimeoutMs,
   runWithTimeout,
   setupTimeoutMs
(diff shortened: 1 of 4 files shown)


nwparker:
Redundant: main (edb5607) already carries an identical benchmark fix. Diff against main is now empty and `pnpm audit:code-quality:type-aware` passes at the IME stack tip without it.

PR #11910, opened 2026-08-01
fix(browser): prevent window.close guest crashes

### LINKED ISSUE #11763: [Bug]: Main-process SIGSEGV after window.close() in a link-opened embedded browser tab
**Operating system:** macOS (Darwin 25.5.0, Apple Silicon - Mac Studio, M1 Max, 32 GB)

**Orca version:** v1.4.162 (first observed on v1.4.161; retested and reproduced on v1.4.162)

### Short summary

The whole Orca app crashes (main-process SIGSEGV, `EXC_BAD_ACCESS` / `KERN_INVALID_ADDRESS` at near-null addresses) when a page in the embedded browser calls `window.close()` in a tab that was opened by clicking an `http://` link in the chat pane, and the user then clicks another tab in the tab strip. The close attempt is (correctly) blocked, but the app appears to retain a dangling reference that the next tab-strip interaction dereferences.

### Reproduction steps

1. Serve this single file on localhost:

```html
<!doctype html>
<title>Orca window.close repro</title>
<body style="font: 16px system-ui; padding: 2rem">
<p>This page calls <code>window.close()</code> 300ms after load.</p>
<p id="s">Attempting close…</p>
<script>
setTimeout(() => {
  window.close();
  setTimeout(() => { document.getElementById('s').textContent =
    'window.close() was blocked (expected). Now click another tab in the tab strip.'; }, 200);
}, 300);
</script>
```

2. In an agent chat, render a markdown link to that page and click it - Orca opens it in a new embedded browser tab. `window.close()` is blocked and the page stays visible.
3. Click another tab in the embedded browser's tab strip.
4. Orca crashes - the entire app, all worktrees. Reproduced five times today (four on 1.4.161 during normal use of our own web app, once on 1.4.162 with the minimal repro above).

Notably, tabs created programmatically via `orca tab create --url ...` ARE allowed to self-close via `window.close()` and do not crash - the difference in close policy between CLI-created and link-click-opened tabs seems to be where the bad state originates.

### Crash evidence

Five `.ips` reports, all `procName: Orca` (main process, not a renderer helper), SIGSEGV at near-null fault addresses:

| Report (local) | Version | Fault address | Top frame |
|---|---|---|---|
| Orca-2026-07-31-115813 | 1.4.161 | 0x60 | `_v8_internal_Node_Print` (Electron Framework) |
| Orca-2026-07-31-121459 | 1.4.161 | 0x10 | `_v8_internal_Node_Print` |
| Orca-2026-07-31-121601 | 1.4.161 | 0x60 | `_v8_internal_Node_Print` |
| Orca-2026-07-31-121708 | 1.4.161 | 0x18 | `v8::Isolate::InstallConditionalFeatures` |
| Orca-2026-07-31-122552 | 1.4.162 | 0x10 | `_v8_internal_Node_Print` |

Happy to attach the full `.ips` files on request.

### Impact

Any web page calling `window.close()` - an ordinary thing for pages to do - can bring down the entire app when opened via a chat link. We hit it in production use of our own localhost app and have removed our `window.close()` call as a workaround, which stops the crashes.

nwparker:
## Summary

- Guard ordinary embedded browser guests against `window.close()` during document creation and after `dom-ready`.
- Clear both forward and reverse guest/tab registrations when a guest is destroyed.
- Preserve explicit CLI-created tab self-close behavior and add Electron regression coverage for the `target=_blank` flow.

## ELI5

An embedded browser tab is like a small browser window inside Orca. A link-created tab ran `window.close()`, and although the close was supposed to be blocked, Orca kept a pointer to the guest after Chromium had destroyed it. Clicking another tab made Orca use that dead pointer and crash. The fix blocks `window.close()` early for normal embedded tabs and removes stale pointers when guests are destroyed. CLI-created tabs intentionally keep their existing self-close behavior.

Fixes #11763

## Validation

- 268 targeted Vitest tests passed.
- Workspace-schema tests passed.
- `pnpm run typecheck` passed.
- `pnpm exec electron-vite build --mode e2e` passed.
- `SKIP_BUILD=1 pnpm exec playwright test tests/e2e/browser-tab.spec.ts --config tests/playwright.config.ts --project electron-headless --workers=1 --grep 'blocked window.close'` passed.
- Changed-file native/type-aware lint passed.

Full lint still reports two unrelated pre-existing duplicate-import warnings in `tests/tools/benchmarks/terminal-cold-park-resource-bench.mjs` and `tests/tools/benchmarks/terminal-cold-park-reveal-bench.mjs`.


AmethystLiang:
looking into it

AmethystLiang:

PR #11911, opened 2026-08-01
fix(daemon): prevent PAM rejection restart cascades

### LINKED ISSUE #11749: [Bug]: login-session-watch PAM rejection cascade kills daemon and loops — renderer crash without user action
## Environment

* **Orca**: v1.4.161
* **Platform**: macOS Darwin 25.4.0 (Apple Silicon)
* **Trigger**: App restart after Mac lock/sleep-wake cycle

## Summary

After a Mac sleep/wake or screen lock event, the daemon's `login-session-watch` probe fails PAM authentication three times in a row, which causes the daemon to call `login-session-dead-retire` and exit. The renderer then restarts, connects to a new daemon, and immediately triggers the same PAM rejection cycle again. This can loop, causing multiple consecutive renderer restarts without any user action.

This is a distinct crash vector from Issue stablyai/orca#9191 (external SIGTERM). Here the **daemon actively chooses to exit** based on PAM probe failures rather than receiving an external signal.

## Observed behaviour

Full `daemon.log` timeline for Jul 30 15:06–15:13 UTC+8 (two independent cascades):

```
15:06:12Z  pid=43504  login-session-probe-rejected  trigger=client-hello  rejections=1
15:06:12Z  pid=85779  login-session-probe-rejected  trigger=client-hello  rejections=1
15:06:22Z  pid=43504  login-session-probe-rejected  trigger=periodic      rejections=2
15:06:22Z  pid=85779  login-session-probe-rejected  trigger=periodic      rejections=2
15:06:32Z  pid=43504  login-session-probe-rejected  trigger=periodic      rejections=3
15:06:32Z  pid=43504  login-session-dead-retire      cause=pam-rejections  resolverHealth=unhealthy
15:06:32Z  pid=43504  daemon-log-closed
15:06:32Z  pid=85779  login-session-probe-rejected  trigger=periodic      rejections=3
15:06:32Z  pid=85779  login-session-dead-retire      cause=pam-rejections  resolverHealth=unhealthy
15:06:32Z  pid=85779  daemon-log-closed

15:07:14Z  pid=96287  startup
15:07:14Z  pid=96287  login-session-watch-armed  trigger=startup

15:12:36Z  pid=96287  login-session-probe-rejected  trigger=client-hello  rejections=1
15:12:46Z  pid=96287  login-session-probe-rejected  trigger=periodic      rejections=2
15:12:57Z  pid=96287  login-session-probe-rejected  trigger=periodic      rejections=3
15:12:57Z  pid=96287  login-session-dead-retire      cause=pam-rejections  resolverHealth=unhealthy
15:12:57Z  pid=96287  daemon-log-closed

15:12:57Z  pid=2963   startup
15:12:57Z  pid=2963   login-session-watch-armed  trigger=startup
```

Three things are notable:

1. **Two stale daemons fire simultaneously.** pids 43504 and 85779 both receive a `client-hello` at 15:06:12 and both run the full rejection ladder independently. This suggests two daemon instances were alive at the same time before the cascade began.
2. **The new daemon (96287) immediately re-enters the rejection cycle.** The renderer reconnects at 15:12:36, which triggers `login-session-probe-rejected trigger=client-hello` — the very first probe attempt already fails. This shows the PAM session state that caused the original failure was not cleared between daemon generations.
3. **The probe interval is exactly 10 seconds.** The three rejections are at :12, :22, :32 — meaning the daemon waits only 20 seconds before self-terminating. If the Mac is still recovering from a wake event or the keychain is momentarily locked, this window may be too short.

## Expected behaviour

The daemon should not self-exit on PAM probe failures in a short burst. A transient PAM rejection (e.g. keychain not yet available after wake) should cause the daemon to enter a backoff-retry loop, not trigger `dead-retire`. The current 10 s × 3 attempt ladder with immediate exit is too aggressive for a network/auth condition that resolves on its own within seconds.

## Impact

Each `daemon-log-closed` causes the renderer to fully reinitialise (`renderer_bootstrap_started`), losing all open editor state and chat context. With two cascades in 7 minutes, this is equivalent to 3–4 unrecoverable session resets in rapid succession without any user-visible warning.

nwparker:
## ELI5

The daemon used PAM and DNS checks like two smoke alarms. After a Mac wakes, both alarms can briefly report trouble even though the login session is still alive. Orca treated three reports over only 20 seconds as proof the session was dead, shut the daemon down, restarted it, and could repeat the same mistake.

This keeps the quick confirmation probes, but they can no longer shut the daemon down unless failures persist across a full, uninterrupted two-minute observation window. Client reconnects and PTY exits cannot shorten that backoff. A sleep/App Nap timer gap restarts the evidence window instead of making the evidence look older. If PAM recovers, the daemon stays alive; if the login session is genuinely dead and DNS is still explicitly unhealthy, the existing cold-recovery path still converges.

Closes #11749.

## Root cause

- Once armed by any accepted PAM probe, `MacosLoginSessionDeathWatch` retired after three conclusive rejections at 10-second intervals.
- The resolver check could also report unhealthy during sleep/wake recovery, so the two signals were not sufficient over that short window.
- `onRetire` then crash-exited the daemon; a replacement could encounter the same transient state and repeat the renderer restart cascade.
- PTY-exit/client activity could pull scheduled probes earlier, and elapsed wall time could jump during sleep, so increasing one timer alone would not create a reliable safety boundary.

## Fix

- Track the first conclusive rejection and require at least 120 seconds of uninterrupted rejection evidence before resolver state gains retirement authority.
- Preserve the rejection backoff deadline against client-hello and PTY-exit triggers.
- Reset rejection evidence on a conclusive acceptance or a sleep/App Nap-sized scheduled-timer gap.
- Hold the rejection count at the confirmation threshold while waiting, preserving the existing retirement telemetry contract.
- Add rejection-window reset/deferred diagnostics and an experimental reliability gate.
- Keep the compressed E2E timing seam proportional with a two-second observation window.

## Reproduction and proof

The deterministic regression reproduces the field-shaped sequence: accepted PAM probe, three conclusive rejections over 20 seconds, unhealthy resolver, then recovery. Before this fix that sequence reached `onRetire`; now it performs zero resolver reads and zero retirements until the 120-second boundary, where acceptance preserves the daemon.

A second regression advances wall time by one hour without executing the scheduled timer, simulating sleep/App Nap. It proves that the overdue callback resets the rejection window, collects fresh evidence after wake, and recovers without retirement. The sustained-death test separately proves that uninterrupted rejections plus an explicitly unhealthy resolver still retire exactly once.

## Freeze and performance analysis

- No synchronous wait, renderer work, provider fanout, or polling loop was added.
- The watcher retains one unref'd timer and at most one async probe; concurrent triggers are coalesced and shutdown aborts in-flight work.
- Live-session recovery is safer. A genuinely stale daemon may take at least two uninterrupted minutes longer to retire, which is deliberate recovery latency rather than a UI/terminal freeze.
- macOS GUI local-daemon only; SSH/headless, relay, Linux, Windows, WSL, mobile, folder workspace, and source-control provider paths are unchanged.

## Reliability contract

- Invariant: `terminal-provider.login-session-retirement` — a GUI-spawned macOS daemon retires for PAM rejection only after prior acceptance, three conclusive rejections spanning at least one uninterrupted 120-second window, and explicitly unhealthy in-process resolver state.
- Failure source: #11749, while preserving the real-logout recovery introduced for #7936.
- Gate: experimental entry in `config/reliability-gates.jsonc`.
- Diagnostics: existing armed/rejected/suppressed/error/retire logs plus deferred and timer-gap reset events.

## Verification

- `pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/macos-login-session-death-watch.test.ts src/main/daemon/daemon-entry.test.ts src/main/providers/macos-tcc-login-shell.test.ts src/main/providers/macos-login-session-pty-probe.test.ts src/main/network/macos-system-resolver-health.test.ts --reporter=dot` — 80 passed
- `pnpm typecheck` — passed
- `pnpm lint` — passed, including reliability manifest, max-lines ratchet, bundled skill, and localization checks
- `oxfmt --check` and `git diff --check` — passed

## Production readiness scan

Verdict: PASS for branch span `edb5607e28fb587be44079eb4f77efe90b3841ff..f722eaa61909d59585117ea1d2f1c54d4c7a88b2`.

- P0: none.
- P1: none.
- P2: none.
- Security/supply chain: mandatory whitespace check passed and security grep returned zero hits. No dependency, lockfile, package-script, workflow, binary, signing, updater, or network-sink changes.
- Backward compatibility/data loss: no schema, persisted state, IPC/RPC/protocol, mobile, or user-owned data changes. The fix prevents the destructive false-retirement path.
- Release/packaging: no release or packaging surfaces changed.

## Residual gaps

- A real dead macOS GUI login session cannot be fabricated without ending the runner login session; sustained-death coverage uses deterministic PAM/resolver oracles.
- The two-minute production window and one-hour suspension are fake-clock tested rather than exercised through a real sleep/wake cycle.
- Multiple stale daemons remain independently throttled; there is no cross-process aggregate coordinator.

### DIFF
--- a/config/reliability-gates.jsonc
+++ b/config/reliability-gates.jsonc
@@ -26,13 +26,13 @@
       "providers": ["local-daemon"],
       "coveredPlatforms": ["macos"],
       "coveredProviders": ["local-daemon"],
-      "coverageNotes": "A deterministic fake clock covers short-burst recovery, sustained session death, event-trigger preemption, resolver corroboration, and shutdown cancellation. The production-only watch remains behind the macOS GUI launch flag; SSH/headless, WSL, Linux, Windows, mobile, and relay paths are unaffected.",
+      "coverageNotes": "A deterministic fake clock covers short-burst recovery, sustained session death, event-trigger preemption, timer suspension, resolver corroboration, and shutdown cancellation. The production-only watch remains behind the macOS GUI launch flag; SSH/headless, WSL, Linux, Windows, mobile, and relay paths are unaffected.",
       "motivatingLinks": [
         "https://github.com/stablyai/orca/issues/11749",
         "https://github.com/stablyai/orca/issues/7936"
       ],
-      "invariant": "A GUI-spawned macOS daemon may retire for PAM rejections only after it previously accepted login wrapping, receives three conclusive rejections spanning at least one 120-second periodic window, and observes explicitly unhealthy in-process resolver state. A conclusive acceptance resets the rejection window, and client or PTY activity cannot shorten its scheduled backoff.",
-      "oracle": "Arm the watch with an accepted probe, inject three rejections over 20 seconds plus unhealthy resolver state, then inject client and PTY activity and require zero resolver reads, zero retirement calls, and exactly four probes until the 120-second boundary. Return acceptance at that boundary and require the daemon to survive. Separately keep rejecting through the boundary and require one retirement, while healthy or unknown resolver state suppresses it and stop aborts an in-flight resolver check.",
+      "invariant": "A GUI-spawned macOS daemon may retire for PAM rejections only after it previously accepted login wrapping, receives three conclusive rejections spanning at least one uninterrupted 120-second observation window, and observes explicitly unhealthy in-process resolver state. A conclusive acceptance or a sleep/App Nap-sized timer gap resets the rejection window, and client or PTY activity cannot shorten its scheduled backoff.",
+      "oracle": "Arm the watch with an accepted probe, inject three rejections over 20 seconds plus unhealthy resolver state, then inject client and PTY activity and require zero resolver reads, zero retirement calls, and exactly four probes until the 120-second boundary. Return acceptance at that boundary and require the daemon to survive. Jump wall time by one hour without firing the scheduled timer, require rejection evidence to restart after wake, and again recover without retirement. Separately keep rejecting through an uninterrupted boundary and require one retirement, while healthy or unknown resolver state suppresses it and stop aborts an in-flight resolver check.",
       "commands": [
         "pnpm exec vitest run --config config/vitest.config.ts src/main/daemon/macos-login-session-death-watch.test.ts --reporter=dot"
       ],
@@ -42,6 +42,7 @@
           "file": "src/main/daemon/macos-login-session-death-watch.test.ts",
           "assertions": [
             "preserves the daemon when a short PAM rejection burst recovers after wake",
+            "does not count a suspended timer gap as rejection evidence",
             "retires only after sustained conclusive rejections with a degraded resolver",
             "suppresses retirement while resolver health is healthy or unknown, then retires on explicit degradation",
             "stop prevents an in-flight resolver check from retiring the daemon"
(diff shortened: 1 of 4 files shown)


nwparker:
CodeRabbit timing note: no change needed. The E2E-only seam already scales the full cadence together (`periodicProbeMs: 2_000`, `rejectionRecheckMs: 500`, `minProbeGapMs: 100`, and `minimumRejectionSpanMs: 2_000`). The production defaults remain an uninterrupted 120-second observation window with 10-second rejection rechecks.

Issue #9803, opened 2026-07-22
[Bug]: Issues with Korean Input in the Terminal


E1jeong:
### Operating system

Windows

### Orca version

1.4.148

### Details

<img width="1397" height="334" alt="Image" src="https://github.com/user-attachments/assets/6fa9c546-f49a-4b3c-862a-b795002037a4" />

As shown in the image, an issue occurred while typing in Korean.
The Enter key didn’t work, and the left and right arrow keys on the keyboard also behaved strangely.

In addition to this, there were instances where characters disappeared while typing in Hangul. Also, when the cursor is positioned on a Hangul character and I press Shift+Enter, a line break occurs along with that character.

AmethystLiang:
looking into it!

asiaenv8053:
Another Windows data point, plus a mechanism hypothesis that might help narrow this down.

**Environment**

- Windows 11 (build 26200), Orca 1.4.158
- Korean 2-set IME (Windows built-in)
- Integrated terminal pane, with an agent TUI running in it

**Symptom A — an in-composition chunk is reverted wholesale**

Not a single dropped syllable: the whole run of text I just typed disappears. Typing `한국`, for example — if the terminal repaints before the syllable is committed, what was on screen as `한국` vanishes entirely rather than losing one character.

**Hypothesis:** it looks like Orca periodically polls / captures terminal state and re-renders from that snapshot. The snapshot comes from the PTY side, which has no knowledge of the IME preedit buffer — uncommitted Hangul only exists in the IME and the renderer, never in the PTY. So when a poll lands mid-composition, the repaint overwrites the visible preedit with the last committed PTY state and the in-progress chunk is gone. That would explain:

- it is timing dependent (it happens when a poll boundary lands mid-composition) rather than keystroke dependent;
- a whole chunk disappears rather than one character;
- Latin input is unaffected, since it commits per keystroke and never has a preedit living outside the PTY state.

If that is the shape of it, the fix would be to suppress or defer the periodic redraw while a composition is active (`compositionstart` → `compositionend`), or to reapply the preedit on top of the refreshed buffer.

I have not verified this against the source — it is inferred from the timing-dependent, chunk-sized nature of the loss, so treat it as a starting point rather than a diagnosis.

**Symptom B — input latency**

[...]

PR #11920, opened 2026-08-01
fix(terminal): enable ConPTY backend for WSL panes to fix CJK line splitting

### LINKED ISSUE #11919: [Bug]: CJK IME input splits every syllable onto a new line in WSL terminal (desktop, Windows/WSL)
## Description

In the Orca desktop app on Windows, typing CJK text (Korean, Chinese, Japanese) into a **WSL terminal** causes every syllable/character to be split onto a new line. The issue affects all agents (Hermes TUI, Claude Code, OpenCode, etc.) and does not occur when running the same agents directly in an external terminal (Windows Terminal → WSL).

English/ASCII input is unaffected. The issue is specific to Orca's integrated xterm.js terminal pane when the PTY backend is WSL.

## Environment

- **OS:** Windows 11 (build 22631) + WSL2 (Ubuntu)
- **Orca:** 1.4.163
- **Terminal backend:** WSL (ConPTY → wsl.exe)
- **IME:** Windows Korean IME (2-set / 두벌식)
- **Agents affected:** All (Hermes TUI, Claude Code, OpenCode, Crush, etc.)

## Steps to reproduce

1. Open Orca desktop on Windows with a WSL workspace.
2. Open a terminal pane (WSL shell or any agent).
3. Switch to a Korean IME.
4. Type `안녕하세요` into the terminal.
5. Every Hangul syllable lands on its own line instead of flowing on one line.

## Expected behavior

`안녕하세요` appears on a single line (same as typing in Windows Terminal → WSL).

## Actual behavior

Each syllable is split onto a new line, as if a newline were inserted after every character.

## Root cause analysis

The xterm.js CompositionHelper patch (`config/patches/@xterm__xterm@6.1.0-beta.287.patch`) adds `_sendCompositionInput` and `keypressMayOverlapComposition` logic to reconcile keypress text with composition input. On Windows Korean IME, `compositionupdate` can fire with empty `data` while the composition is still active, causing `keypressMayOverlapComposition` to be set to `true` (condition: `_lastCompositionData.length === 0 && keypress.length === 0`). This causes keypress text to be accumulated into `_pendingKeypressData`, and the `_sendCompositionInput` merge logic (overlap detection between composition input and keypress text) produces incorrect output — newlines get interleaved with Hangul syllables, sending them to the PTY split across lines.

Additionally, the Windows/WSL path lacks the IME native text forwarder that macOS has (`terminal-ime-native-text-forwarder.ts` is `isMac`-only), and the Linux candidate-key guard (`terminal-ime-candidate-key-release-guard.ts`) is `isLinux`-only, leaving Windows with only the composition tracker and the xterm bypass policy's keyCode 229 suppression.

### Key code paths

- `config/patches/@xterm__xterm@6.1.0-beta.287.patch` — `CompositionHelper._sendCompositionInput` overlap merge
- `src/renderer/src/components/terminal-pane/xterm-bypass-policy.ts:125` — `passesStandalone229Keydown = isMac || isLinux` (Windows excluded)
- `src/renderer/src/components/terminal-pane/terminal-ime-native-text-forwarder.ts` — macOS only
- `src/renderer/src/lib/pane-manager/windows-pty-compatibility.ts:88` — WSL panes excluded from `windowsPty` option (though build 22631 > 21376, so wrapping heuristics would not activate anyway)

## Related issues

- #8038 — macOS desktop terminal Hangul syllable on new line (isComposing guard, Shift+Enter)
- #11332 — macOS chat input Korean IME composition breaks intermittently
- #9862 — Linux IBus Korean input regression
- #9803 — Windows terminal Korean input issues
- #11235 — iOS NFD Hangul decomposition (mobile)

Lee-Si-Yoon:
## Summary

WSL panes in the Orca desktop app on Windows spawn `wsl.exe` through ConPTY, but `isLocalNativeWindowsConpty()` excluded them from the `windowsPty` option. Without the `conpty` backend, xterm.js treats ConPTY's wrap markers as real line breaks, causing CJK text (width=2 cells) to split per character.

**Fix:** Add `isLocalWindowsConptyPane()` that includes WSL panes, and use it in `buildWindowsPtyCompatibilityOptions()` so xterm.js gets the `conpty` backend for WSL too.

`isLocalNativeWindowsConpty()` (used for Kitty keyboard withhold) stays unchanged so WSL keeps CSI-u support — WSL-internal Linux PTY decodes Kitty CSI-u correctly.

Fixes #11919.

## Root cause

`src/renderer/src/lib/pane-manager/windows-pty-compatibility.ts:88` — `isLocalNativeWindowsPty()` excluded WSL panes (checks `isWslCwd()` and `isWslShellOverride()`), so `buildWindowsPtyCompatibilityOptions()` returned `{}` for WSL. xterm.js then had no `windowsPty` option and didn't know it was talking to ConPTY.

ConPTY echoes input with its own `\r\n` wrap markers. xterm.js without the `conpty` backend flag treats each wrapped visual line as a real line break. CJK characters are width=2 cells (Unicode 11), so ConPTY wrapping triggers roughly twice as often per screen line vs ASCII — Korean text breaks per character while English works.

## Verification

- **CDP inspection** confirmed all WSL PTY sessions have `cwd: \\wsl.localhost\Ubuntu\...` → `isWslCwd()` = true → `windowsPty` was `{}`.
- **Agent-affected:** all (Hermes TUI, Claude Code, OpenCode, Crush, etc.) — the bug is in xterm.js wrapping, not in any agent.
- **External terminal works:** Windows Terminal → WSL → Hermes is fine because Windows Terminal sets the ConPTY backend correctly.

## Changes

- `windows-pty-compatibility.ts`: Add `isLocalWindowsConptyPane()` (includes WSL); `buildWindowsPtyCompatibilityOptions()` uses it. `isLocalNativeWindowsConpty()` unchanged (Kitty keyboard stays advertised for WSL).
- `windows-pty-compatibility.test.ts`: Update tests — WSL panes now get ConPTY options (`#11919`), but `isLocalNativeWindowsConpty()` still excludes WSL for Kitty keyboard.

## Test plan

- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/lib/pane-manager/` — 594 passed, 1 skipped
- [x] Manual: type Korean `안녕하세요` in an Orca WSL terminal pane — should appear on one line
- [ ] Manual: verify Kitty keyboard protocol still advertised for WSL (e.g. `claude` CLI enhanced keys work)

## Notes

This is the Orca-side fix. The xterm.js patch (`config/patches/@xterm__xterm@6.1.0-beta.287.patch`) and `keypressMayOverlapComposition` logic from #11293 may also need Windows-specific adjustments for full robustness, but enabling the ConPTY backend for WSL panes is the correct first step — xterm.js needs to know it's talking to ConPTY.

## Related

- #11919 — issue
- #11293 — IME composition lifecycle reconciliation (may have introduced the `keypressMayOverlapComposition` edge case on Windows)
- #8038 — macOS desktop terminal Hangul syllable on new line
- #9803 — Windows terminal Korean input issues

### DIFF
--- a/src/renderer/src/lib/pane-manager/windows-pty-compatibility.test.ts
+++ b/src/renderer/src/lib/pane-manager/windows-pty-compatibility.test.ts
@@ -89,7 +89,9 @@ describe('buildWindowsPtyCompatibilityOptions', () => {
     ).toEqual({})
   })
 
-  it('skips compatibility options for WSL cwd terminals', () => {
+  // WSL panes spawn wsl.exe through ConPTY; xterm.js needs the conpty backend
+  // to handle ConPTY's wrap markers or CJK text (width=2) splits per line (#11919).
+  it('includes WSL cwd terminals for ConPTY compatibility', () => {
     for (const cwd of [
       '\\\\wsl.localhost\\Ubuntu\\home\\me\\repo',
       '\\\\wsl$\\Debian\\home\\me\\repo',
@@ -105,11 +107,13 @@ describe('buildWindowsPtyCompatibilityOptions', () => {
           shellOverride: null,
           executionHostId: 'local'
         })
-      ).toEqual({})
+      ).toEqual({
+        windowsPty: { backend: 'conpty', buildNumber: 26100 }
+      })
     }
   })
 
-  it('skips compatibility options when the shell override launches WSL', () => {
+  it('includes WSL shell override terminals for ConPTY compatibility', () => {
     expect(
       buildWindowsPtyCompatibilityOptions({
         userAgent: 'Mozilla/5.0 (Windows NT 10.0; Win64; x64)',
@@ -119,7 +123,9 @@ describe('buildWindowsPtyCompatibilityOptions', () => {
         shellOverride: 'C:\\Windows\\System32\\wsl.exe',
         executionHostId: 'local'
       })
-    ).toEqual({})
+    ).toEqual({
+      windowsPty: { backend: 'conpty', buildNumber: 26100 }
+    })
   })
 
   it('returns no options outside Windows', () => {
@@ -240,6 +246,20 @@ describe('isLocalNativeWindowsConpty', () => {
     ).toBe(true)
   })
 
+  // WSL-internal Linux PTY decodes CSI-u correctly, so Kitty keyboard stays on.
+  // ConPTY wrap-marker handling is separately enabled via buildWindowsPtyCompatibilityOptions.
+  it('does NOT treat a WSL pane as a native ConPTY (Kitty keyboard stays advertised)', () => {
+    expect(
+      isLocalNativeWindowsConpty({
+        userAgent: 'Mozilla/5.0 (Windows NT 10.0; Win64; x64)',
+        connectionId: null,
+        cwd: '\\\\wsl.localhost\\Ubuntu\\home\\me\\repo',
+        shellOverride: null,
+        executionHostId: 'local'
+      })
+    ).toBe(false)
+  })
+
   it('does NOT treat a remote-runtime (serve) pane as a native ConPTY even when the raw Windows heuristic matches', () => {
     // Regression: a serve-hosted pane on a Windows client has no SSH connectionId
     // and a Linux cwd, so isLocalNativeWindowsPty returns true. Without the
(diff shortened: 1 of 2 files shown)


Lee-Si-Yoon:
Hello @Jinwoo-H would appreciate if reviewed.

nwparker:
Closing as superseded by the renderer-side IME fixes in #12265 and #12278, now available in v1.4.167. The associated report also reproduced on native Windows ConPTY, so enabling a different backend specifically for WSL is no longer the right fix. Thank you for the investigation and contribution.

PR #11927, opened 2026-08-01
test(ime): add a local trace recorder for platforms CI cannot cover

nwparker:
## Summary

Adds `pnpm ime:record` — a localhost page that captures a real IME composition trace from your own machine and your own input method, in the exact JSON shape the fixture loader already consumes.

This exists because of a hard limit: CDP-dispatched keystrokes bypass the OS input-method layer, so no amount of `page.keyboard.type()` produces a real `compositionstart` → `compositionend` pair. The Linux matrix in #11898 gets around this by driving ibus through X11 with `xdotool`, which works — but there is no macOS or Windows IME CI, and there is unlikely to be one soon. On those platforms a human typing into a page is the only way to get a trace that can honestly be labelled `recorded` rather than `derived`.

The page records the same field set as `tests/e2e/terminal-ime-boundary-probe.ts`, **including the target's `value`, `selectionStart` and `selectionEnd` at each event**. Those offsets are the whole point: they are what distinguishes a recorded trace from a reconstructed one. A real engine sends `compositionupdate` with the preedit character *selected* (`selectionStart: 0, selectionEnd: 1`) so it can replace it, and no hand-written fixture guesses that.

`isComposing` is stored as the engine reported it and is deliberately **not** coerced — Safari reports `undefined` rather than `false`, and the trace type models that distinction.

An earlier revision of this PR made that claim while not delivering it. The recorder did read `event.isComposing` uncoerced, but the trace is POSTed with `JSON.stringify`, which omits undefined-valued keys outright — so the written file had no `isComposing` key at all, and the bit was gone before the payload left the browser. It is now carried as `null`, which JSON can express, and transcribed back to `undefined` in the fixture. The replayer half of the same round trip is fixed in #11895.

Two more fields the page now captures, both added because a trace recorded without them cannot reproduce the quirk it was recorded for:

- **`repeat`** — the accent panel a macOS press-and-hold opens is recognisable by nothing else. Since macOS press-and-hold is the single quirk most in need of a hand-recorded trace, a recorder that flattened it would have undercut the PR's own purpose.
- **`selectionDirection`** — without it no trace can express a backward selection, so a preedit replacement range is only ever representable as a collapsed caret.

Two fields are emitted as explicit `TODO`s, because the page genuinely cannot observe them: the input method's name, and a description of what the human typed.

Everything is served from `127.0.0.1` and written to `test-results/ime-recordings/` (already gitignored). Nothing is uploaded.

## Screenshots

Terminal-only; the recorder page is a local dev tool, not app UI.

## Testing

- [x] `pnpm lint` — oxlint clean
- [x] `pnpm typecheck`
- [x] `pnpm test` — `record-ime-trace.test.mjs` (new) holds the recorder and the fixture schema in lockstep: it fails if the page starts capturing a field the schema cannot express, which is exactly the drift that produced the `isComposing` bug above
- [ ] `pnpm build` — not run
- [x] Exercised end to end: served the page, POSTed a trace, confirmed the written JSON deserializes through the same loader as the CI-captured recording
- [x] The page script is inside a template literal, so `node --check` on the module does not validate it; extracted and syntax-checked separately

Stated plainly: the recorder's browser-side `record()` runs in a page string and is not executed by any test. The contract test checks which fields it captures, not that a real browser hands them over correctly — only an actual recording session does that, which is the manual step this tool exists for.

## AI Review Report

The captured schema is intentionally identical to the CI probe's, so traces from either source deserialize through one loader. Its shape is what faithful replay requires: environment + initial state + event list + final state, with per-event observed buffer state, and `isComposing` typed `boolean | undefined` so Safari reproduces faithfully.

Cross-platform: pure Node `http` plus `path.join`; no shell-out, so no `open`/`xdg-open`/`start` divergence. The URL is printed rather than auto-opened for that reason. Platform and browser are inferred from the user agent, with WebKit and Gecko checked before Chromium because Electron reports a Chrome token too.

## Security Audit

- **Network**: binds `127.0.0.1` only, on a port overridable via `ORCA_IME_RECORDER_PORT`. No outbound requests.
- **Path handling**: the trace name is sanitized to `[a-z0-9-]` before being used as a filename, so a crafted name cannot traverse out of the output directory.
- **Data**: writes only under `test-results/`, which is gitignored. Note the recorded `value` fields contain whatever the operator typed — review a trace before committing it as a fixture.
- No auth, secrets, IPC, dependency, or git changes.

## Notes

Top of stack #11899, on #11898. Companion to the ibus matrix rather than a replacement: CI covers hangul/libpinyin/anthy/unikey on Linux, this covers everything else.

### DIFF
--- a/config/scripts/record-ime-trace.mjs
+++ b/config/scripts/record-ime-trace.mjs
@@ -100,6 +100,7 @@ let initial = snapshot()
 
 function snapshot() {
   return {
+    selectionDirection: target.selectionDirection ?? 'none',
     selectionEnd: target.selectionEnd ?? 0,
     selectionStart: target.selectionStart ?? 0,
     value: target.value
@@ -114,20 +115,24 @@ function record(event) {
   } else if (event instanceof InputEvent) {
     entry.data = event.data ?? null
     entry.inputType = event.inputType
-    // Safari reports undefined rather than false and the distinction is load-bearing,
-    // so it is preserved instead of coerced.
-    entry.isComposing = event.isComposing
+    // Safari reports undefined rather than false and the distinction is load-bearing.
+    // JSON.stringify drops an undefined value outright, so null carries it instead;
+    // transcribe it back to undefined in the fixture.
+    entry.isComposing = event.isComposing ?? null
   } else if (event instanceof KeyboardEvent) {
     entry.code = event.code
     entry.isComposing = event.isComposing
     entry.key = event.key
     entry.keyCode = event.keyCode
+    // The accent panel a macOS press-and-hold opens is only recognisable by this.
+    entry.repeat = event.repeat
   }
   events.push(entry)
   countEl.textContent = String(events.length)
   const detail = [
     entry.key ? 'key=' + entry.key : '',
     entry.keyCode === undefined ? '' : 'keyCode=' + entry.keyCode,
+    entry.repeat ? 'repeat' : '',
     entry.inputType ? entry.inputType : '',
     entry.data === undefined ? '' : 'data=' + JSON.stringify(entry.data),
     'isComposing=' + entry.isComposing,
--- a/config/scripts/record-ime-trace.mjs
+++ b/config/scripts/record-ime-trace.mjs
@@ -1,0 +1,261 @@
+// Captures a real IME composition trace from your own machine and your own input
+// method, in the exact JSON shape src/renderer/src/lib/*.test-fixtures.ts consumes.
+//
+// Why this exists: CDP-dispatched keystrokes bypass the OS input-method layer, so no
+// amount of page.keyboard.type() produces a real composition. The Linux CI matrix
+// drives ibus through X11 and covers hangul/libpinyin/anthy/unikey, but there is no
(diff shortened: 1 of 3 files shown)


nwparker:
Leaving this open — not superseded. #13128 (17b3dff3c41) removed the first-party composition layer, but a trace recorder is orthogonal to which layer owns composition, and `config/scripts/record-ime-trace.mjs` never landed on main.

If anything this got *more* valuable, not less. The recurring lesson across this whole stack was that hand-written event sequences encode what we assume an IME does, and real ones disagree — a CJK source rewrites `event.key` while `event.code` keeps the physical key (#12171, #13033), and `wScan=0` injection produces `code:''` with a single keyup where Chromium dispatches two. Synthetic events would have hidden all of that.

Needs a rebase onto the new baseline. Worth carrying two rules the recorder should enforce, both learned the hard way: report keys-requested vs keys-inserted so a silent no-op cannot pass as a null result, and match controls by *column*, not character count — Hangul is two cells per syllable.

PR #11864, opened 2026-08-01
fix(terminal): coordinate reveal atlas recovery

brennanb2025:
## Summary

- coalesce settled terminal reveal repaint requests across pane managers
- reattach revealed WebGL renderers before one registry-wide atlas reset and repaint
- cover sibling-manager recovery and concurrent reveal coalescing with regression tests

## Root cause

xterm terminals with matching render configuration share a WebGL glyph atlas. After restart or background-worktree reveal, a delayed pane-local reset could clear that shared atlas after a sibling terminal had rebuilt, leaving stale glyph coordinates on screen until typing forced another render.

## Reliability contract

- **Invariant — `terminal-renderer.shared-atlas-convergence`:** revealing or restarting one pane must not leave any sibling renderer displaying stale or corrupt glyphs.
- **Failure source:** STA-3122 and the reproduced delayed pane-local reset after restart/background reveal.
- **Oracle:** deterministic multi-manager tests require every sibling to reset and repaint, concurrent reveals to coalesce to one registry-wide recovery, and a live Electron reference row to remain intact after tab and whole-workspace reveal while a sibling PTY streams.
- **Gate:** `xterm-addon.boundary-containment` (experimental, partial). The broader `terminal-render.pixel-refresh-repair` visual-golden gate remains registered with protection `none`.
- **Provider/platform coverage:** renderer-local behavior leaves local, daemon, SSH/remote, WSL, and mobile/relay PTY ownership and transport unchanged. Live Electron coverage is macOS; Linux and Windows GPU paths were not exercised in this PR review.
- **Performance budget:** one settled registry-wide O(managers + panes) recovery is coalesced across concurrent reveal requests. No polling, subprocesses, output-path scans, or new per-write work were added; the count test pins one reset for concurrent requests.
- **Diagnostics:** `webgl-atlas-reset` breadcrumbs include live-manager count and recovery reason; existing pane rendering diagnostics distinguish missed repaint from atlas corruption.
- **Residual gaps:** no live Linux/Windows GPU run, and the broader pixel-refresh-repair gate still lacks a manifest command/red-green history.

## Validation

- latest `origin/main` merged into the branch
- same-model senior review subagent: clean, no meaningful correctness, lifecycle, or performance findings
- 86 focused terminal reliability tests across the registered gate and adjacent registry/lifecycle suites
- changed-file oxlint
- web TypeScript check
- reliability-gate manifest validation and `git diff --check`
[...]

Issue #11757, opened 2026-07-31
[Bug]: Garbled/overlapping terminal rendering after tab-switch reveal (OpenCode, macOS); resize restores


l3aki:
### Operating system

macOS

### Orca version

1.4.161

### Details

**Short summary:**<br>When switching back to a terminal tab running OpenCode (full-screen TUI), the pane intermittently renders garbled/overlapping content: text fragments drawn at wrong offsets, duplicated line fragments, and pixelated ghost blocks along the left edge. The app keeps running underneath — this is purely a rendering issue. Toggling the right side panel (which resizes the pane) restores correct rendering.

**What happened?**

* The terminal tab runs OpenCode 1.18.10 (full-screen TUI, Kimi K3 via OpenCode Go).
* After switching to another tab and back, the pane shows a corrupted frame: misplaced/duplicated text fragments, overlapping lines, and dark pixelated blocks along the left edge (screenshots attached below).
* The terminal keeps working underneath (the agent continues and new output arrives); only the composited frame is stale/corrupt.
* Toggling the right side panel forces a resize, and the pane repaints correctly.

**How can we reproduce it?**

1. Open a terminal tab in Orca and run OpenCode (full-screen TUI).
2. Let the agent produce output for a while.
3. Switch to another tab, then switch back.
4. Intermittently, the pane renders garbled/overlapping content as described above.

**Anything else that might help:**

* Recovery via a resize (side-panel toggle) matches the workaround pattern in stablyai/orca#7240 (Windows + Claude Code reveal-repaint stale frame, fixed by Ctrl+L/select/resize). This report is the same symptom class on macOS + OpenCode on Orca 1.4.161, so the reveal-repaint fix shipped for stablyai/orca#7240 does not appear to cover this path.
* Same terminal/session also exhibits the viewport-jump issue reported in stablyai/orca#11753 — both may involve stale viewport/compositing state on reveal.

**Screenshots:**<br>(attached below: full Orca window and terminal close-up showing the corrupted frame)

<img src="https://github.com/user-attachments/assets/f37bad65-48b9-4f3b-9913-546a0468ed00 " alt="Image" width="1728" data-linear-height="1117" />

<img src="https://github.com/user-attachments/assets/e7e1e3f4-6dcb-4d2d-bd69-c6eb25ab9410 " alt="Image" width="1032" data-linear-height="1025" />

nwparker:
I found a deterministic repro matching the report on the exact reporter build.

[...]

PR #11930, opened 2026-08-01
fix(terminal): avoid delayed atlas wipes on tab reveal

nwparker:
## Re-evaluation

This PR is draft and is not ready to merge as a fix for #11757.

- The earlier stranded and recovered screenshots do not establish recovery; they are visually equivalent.
- On the pre-change build, 20 of 20 reveal captures were byte-identical to baseline. The test failed only because it observed four atlas reset calls instead of one.
- Three of 20 post-resize captures differed because the viewport jumped from the top to rows 15 through 26. That is a separate viewport issue, not garbled glyph recovery.
- The exact Orca 1.4.161 release also passed 10 of 10 runs of the existing eight-cycle tab-switch pixel oracle.

The current diff demonstrates removal of redundant reset work, but it does not yet demonstrate the reported visual corruption or prove that the removed reset burst caused it. The PR will remain draft until a natural red-green visual oracle justifies the fix; otherwise it should be withdrawn.

Stacked base: #11864.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-visibility-resume.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-visibility-resume.test.ts
@@ -21,12 +21,6 @@ vi.mock('./pane-helpers', () => ({
   fitPanes: vi.fn(),
   focusActivePane: vi.fn()
 }))
-const scheduleTabRevealWebglAtlasRecovery = vi.fn()
-vi.mock('./terminal-webgl-atlas-recovery', () => ({
-  // Why: the light-tab reveal must recover the atlas immediately, decoupled from
-  // the terminal-output debounce (which a background stream could otherwise defer).
-  scheduleTabRevealWebglAtlasRecovery: () => scheduleTabRevealWebglAtlasRecovery()
-}))
 const resetTerminalLinkifierHoverState = vi.fn()
 const isTerminalLinkifierHoverActive = vi.fn((_terminal: unknown) => false)
 vi.mock('@/lib/pane-manager/terminal-linkifier-hover-reset', () => ({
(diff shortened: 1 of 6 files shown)


nwparker:
Re-evaluation found that the current OpenTUI E2E is not a visual red/green repro for #11757. On the pre-fix build, all 20 reveal captures were byte-identical to baseline; the test failed only because it observed four atlas-reset calls instead of one. Three of 20 post-resize captures also changed because the viewport jumped to rows 15–26, not because of glyph corruption. Marking this draft while the root-cause boundary and regression oracle are corrected.

nwparker:
## Corrected repro and disposition

The earlier “stranded” / “recovered” screenshot pair was not a valid visual repro; both frames showed the same rendering apart from live animation.

The corrected deterministic repro seeds the persisted layout corruption that Orca 1.4.161 allowed: two terminal leaves claim the same live PTY. With a synchronized, animated full-screen frame running:

- **1.4.161:** restores **2 visible xterms** for the PTY. Switching away and back produces the reported fragmented/overlapping frame; toggling the right sidebar repaints it correctly.
- **1.4.163 after #11726:** normalizes the same input to **1 visible xterm** before replay. The frame remains intact across the same tab switch.

### 1.4.161 — corrupted after reveal

1.4.161 corrupted after reveal

### 1.4.161 — recovered after sidebar resize

1.4.161 recovered after resize

### Post-#11726 — intact after reveal

post-11726 intact after reveal

### ELI5

Orca had accidentally put two screens in charge of drawing and resizing one terminal. After hiding and showing the tab, they disagreed about the terminal size, so pieces of the full-screen app landed in the wrong places. Resizing made them agree again. #11726 prevents the second screen from being created.

This means #11757 is already covered by #11726. The atlas-reset removal in this PR is redundant and its opt-in E2E only proves fewer resets, not the reported corruption. Closing this draft rather than shipping an unsupported rendering change.

nwparker:
Closing as superseded by #11726; corrected red/recovered/green evidence is in the preceding comment.

PR #11937, opened 2026-08-01
fix(terminal): require press-time evidence for the Windows IME Enter-keyup newline

dogfooter8750:
## Summary

Fixes a residual "spontaneous newline" path of #11878 that survives #11616: on Windows, the IME Enter-keyup synthesis branch (`keyboard-handlers.ts` `onKeyUp`) infers the Shift/Ctrl+Enter chord from **release-time** modifier state. Two user-visible consequences with a Korean IME:

- A plain committing Enter (delivered as `Process`/229 with no modifiers) followed by a rolled-over Shift for the next doubled consonant (fast typing of 했다 / 아빠 / 얘기 patterns) makes the Enter keyup report `shiftKey=true`, and the branch synthesizes a Shift+Enter newline (`\x1b\r`) the user never chorded. Because the deferred sender fires it on session end or its 200 ms fallback, the newline appears decoupled from any keystroke — exactly the "random newline during lag" reports on #11878.
- A directly-sent Shift+Enter (pressed with no composition active) can send a **second** newline from its keyup once the next Hangul composition has started, since the direct-send path never claims the modified-enter chord.

The fix records observed Enter keydowns (`code → timeStamp`) and lets the keyup synthesis run only for presses whose keydown the IME swallowed entirely — the case that branch exists for. A balancing keyup that copies the keydown's native `timeStamp` keeps the evidence armed for the later physical release (mirroring `releaseRedispatchedEnter`'s same-`timeStamp` semantics). Blank-`code` `Process`/229 keydowns are deliberately **not** recorded, so #11616's "blank code falls back to the Enter keyup path" design is preserved.

Interplay with #11616 (verified locally): #11616 applies cleanly on top of this branch and the combined tree passes all four IME suites, 66/66.

## Screenshots

No visual change.

## Testing

- [x] Added or updated high-quality tests that would catch regressions: 4 new tests in `keyboard-handlers-ime-enter-keyup.test.tsx` — rollover suppression, balancing-keyup retention, keyup-only synthesis preserved (positive control), and no double-send after a direct send. The two suppression tests fail against the pre-fix handler (`sendInput` receives `"\r"`) and pass with it.
- [x] `vitest` on the affected IME suites: 54/54 on this branch; 66/66 with #11616 applied on top (`keyboard-handlers-ime-enter-keyup`, `keyboard-handlers-ime`, `terminal-ime-deferred-newline`, `terminal-ime-composition-route`)
- [x] `oxlint` (default and `config/oxlint-react-doctor.json`) and `oxfmt --check` on the changed files
- [x] `tsc --noEmit -p config/tsconfig.tc.web.json` (the change is renderer-only)
- [ ] Full `pnpm test` / `pnpm build` were not run in this environment (native runtime build unavailable); the change surface is two renderer TS files

## AI Review Report

Three independent review passes (adversarial correctness, cross-platform, security/perf) all approved; two of their findings were folded back into the change:

- **Correctness**: verified the swallowed-keydown keyup path is preserved (pinned by the positive-control test), chord-absorb/`releaseRedispatchedEnter` housekeeping still runs on the suppressed path, and per-`code` keying keeps Enter and NumpadEnter evidence independent. Review-driven changes: (1) evidence is recorded ahead of every `onKeyDown` early return, so keydowns on editable/out-of-scope targets also disqualify their keyup; (2) the original `Set` was replaced with a `code → timeStamp` map so a balancing keyup (timeStamp copied from the keydown) does not consume the evidence before the physical release.
- **Cross-platform (macOS/Linux/SSH)**: recording and the only consumer both sit behind `isWindows`; macOS/Linux behavior is byte-identical. Remote/SSH panes are unaffected — the change only suppresses sends ahead of the unchanged transport abstraction.
- **#11616 coexistence**: the recording predicate excludes blank-`code` 229 keydowns by construction, so #11616's blank-code keyup fallback still synthesizes. Verified empirically (66/66 with #11616 overlaid).
- Accepted residual (documented): if an observed keydown's keyup never arrives as `key === 'Enter'`, a stale entry can suppress at most one future keyup-only synthesis for that code; it self-heals on the next keyup and is cleared on window blur and effect cleanup.

## Security Audit

- Input handling: the change strictly narrows the conditions under which a synthesized `\x1b\r` reaches the PTY; no new byte-injection path is added.
- Hot path: O(1) map operations with no per-event allocation; non-Enter keys keep their existing early returns on both keydown and keyup.
- Memory: the map is bounded (realistically `Enter`/`NumpadEnter` only) and cleared on keyup consumption, window blur, and effect cleanup.
- No new data crosses IPC or remote boundaries; nothing is logged.

## Notes

- Windows-specific behavior by design; the synthesis branch it gates is itself `isWindows`-only.
- A sibling issue remains out of scope: `sendTerminalInputAfterComposition`'s 200 ms fallback fires without re-checking pending composition, so even a *legitimate* deferred newline can land mid-composition under renderer lag. Detailed in a comment on #11878.

### DIFF
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers-ime-enter-keyup.test.tsx
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers-ime-enter-keyup.test.tsx
@@ -223,6 +223,113 @@ describe('Windows IME Enter-keyup press-time evidence', () => {
     harness.dispose()
   })
 
+  it('guards every keyup of a rapid double Enter press', () => {
+    const harness = createHarness()
+    const hook = renderHook(() => useTerminalKeyboardShortcuts(harness.deps))
+    harness.startComposition()
+
+    // Two committing Enter presses land before either release is processed.
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keydown', {
+        key: 'Process',
+        code: 'Enter',
+        keyCode: 229,
+        timeStamp: 10,
+        isComposing: true
+      })
+    )
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keydown', {
+        key: 'Process',
+        code: 'Enter',
+        keyCode: 229,
+        timeStamp: 20,
+        isComposing: true
+      })
+    )
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keydown', {
+        key: 'Shift',
+        code: 'ShiftLeft',
+        keyCode: 16,
+        timeStamp: 25,
+        shiftKey: true
+      })
+    )
+    for (const timeStamp of [30, 40]) {
+      harness.terminalInput.dispatchEvent(
+        keyboardEvent('keyup', {
+          key: 'Enter',
+          code: 'Enter',
+          keyCode: 13,
+          timeStamp,
+          shiftKey: true
+        })
+      )
+    }
+    vi.runAllTimers()
+
+    expect(harness.sendInput).not.toHaveBeenCalled()
+    hook.unmount()
+    harness.dispose()
+  })
+
+  it('drains one press worth of evidence per release, so a later swallowed keydown still synthesizes', () => {
+    const harness = createHarness()
+    const hook = renderHook(() => useTerminalKeyboardShortcuts(harness.deps))
+    harness.startComposition()
+
+    // Auto-repeat re-fires keydown with no release in between; the whole run
+    // must cost exactly one release to drain.
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keydown', {
+        key: 'Process',
+        code: 'Enter',
+        keyCode: 229,
+        timeStamp: 10,
+        isComposing: true
+      })
+    )
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keydown', {
+        key: 'Process',
+        code: 'Enter',
+        keyCode: 229,
+        timeStamp: 20,
+        isComposing: true,
+        repeat: true
+      })
+    )
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keyup', {
+        key: 'Enter',
+        code: 'Enter',
+        keyCode: 13,
+        timeStamp: 30,
+        shiftKey: true
+      })
+    )
+    vi.runAllTimers()
+    expect(harness.sendInput).not.toHaveBeenCalled()
+
+    // Evidence is now drained: a genuinely swallowed keydown still synthesizes.
+    harness.terminalInput.dispatchEvent(
+      keyboardEvent('keyup', {
+        key: 'Enter',
+        code: 'Enter',
+        keyCode: 13,
+        timeStamp: 50,
+        shiftKey: true
+      })
+    )
+    vi.runAllTimers()
+
+    expect(harness.sendInput).toHaveBeenCalledTimes(1)
+    expect(harness.sendInput).toHaveBeenCalledWith('\x1b\r')
+    hook.unmount()
+    harness.dispose()
+  })
+
   it('still synthesizes a newline when the IME swallowed the Enter keydown entirely', () => {
     const harness = createHarness()
     const hook = renderHook(() => useTerminalKeyboardShortcuts(harness.deps))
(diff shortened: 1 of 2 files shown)


dogfooter8750:
Good catch on the map-overwrite edge case — it was real, and I reproduced it before fixing it. Driving the harness with two Enter keydowns (ts 10, 20) before either release, then two keyups (ts 30, 40) with a rolled-over Shift, the second keyup synthesized `\x1b\r`: the first release saw a mismatched timeStamp, dropped the only entry, and left the second press unguarded.

Fixed in e17d22cf by tracking one entry per press instead of one slot per code (`Map<string, number[]>`), draining exactly one entry per physical release. Balancing-keyup semantics are unchanged — a keyup whose timeStamp matches any recorded press is still non-consuming.

Two details worth flagging since they are new surface:

- Auto-repeat keydowns deliberately do not stack an entry. A repeat run re-fires keydown with no intervening keyup and ends in a single release, so stacking would leave residue that release cannot drain and would suppress a later legitimate keyup-only synthesis.
- The list is bounded (`MAX_OBSERVED_ENTER_KEYDOWNS_PER_CODE = 8`) so a press whose release never arrives cannot grow it without end.

Tests: added `guards every keyup of a rapid double Enter press` (fails on the previous commit, passes now) and `drains one press worth of evidence per release, so a later swallowed keydown still synthesizes`, which pins both the repeat handling and the fact that the keyup-only path stays reachable. Locally on Windows: 56/56 across the four IME suites, and 68/68 with #11616 applied on top. `oxlint` (default + react-doctor), `oxfmt --check`, and `tsc -p config/tsconfig.tc.web.json` are clean.

holdn2:
Heads up — I've opened #12120 as a draft that carries your commit `c865c0b`
(press-time evidence for the Enter keyup) with `--author` preserved, alongside the
minimum hunks from #11616, #11896 and #11870.

One change I should flag: rather than adding `keyboard-handlers-ime-enter-keyup.test.tsx`
as a new file, I folded your four tests into the existing colocated
`keyboard-handlers-ime.test.tsx` suite, to keep the combined branch to one handler test
file. The test bodies and intent are unchanged.

Your rollover case turned out to be load-bearing for the combined fix: the regression test
I added for a Shift-jamo followed by a committing Enter only passes with your keyup gate
in place.

This is not meant to supersede your PR — maintainers may well prefer landing it directly,
which is fine by me. Happy to close mine on request, and to correct anything I've
misrepresented.

dogfooter8750:
Nothing misrepresented — the attribution is more careful than it needed to be, and folding the tests into `keyboard-handlers-ime.test.tsx` is the right call for a combined branch.

On your offer to close: please don't. A single combined fix is what actually helps the people on #11878, and one branch carrying the keydown predicate, the keyup gate, and the deferred-commit work is a much better thing for a maintainer to review than four PRs landing in some arbitrary order.

I'll leave this PR open, but explicitly as a fallback rather than as competition. #12120 is a draft carrying work from four authors, so if it needs time in review, this one is two files with green checks and is cheap to take on its own for the keyup half. Whichever lands, I'm happy — and if a maintainer signals they want #12120, I'll close this myself so it isn't in the way.

One carry-over when you next touch the branch, detail and measurements in [my comment there](https://github.com/stablyai/orca/pull/12120#issuecomment-5161011717): `e17d22cf` closes a rapid-double-Enter hole that `c865c0b` left open — the single slot per code meant the first release consumed the only entry and left the second press unguarded. It applies cleanly on top of your branch and takes that case from one spurious `\x1b\r` to zero.

dogfooter8750:
Closing — this landed on `main` via #12265, which carries both commits from this branch with authorship intact:

- `ad0a70aa14` require press-time evidence before synthesizing an IME Enter-keyup newline
- `bacdf3a31d` keep press evidence per Enter press, not per code

Thanks @holdn2 for carrying it into the combined branch, and in particular for reproducing each case against the handler before taking it rather than trusting my description — both the rapid-double-press hole and the corrected re-render mechanism surfaced because you did that. And thanks to the maintainers for the wider fix; the keyup half was only one of several paths that needed closing.

I have since root-caused the re-render amplifier I flagged, and opened it separately rather than leave it as an unexplained aside. Summary is in my comment on #12120.

PR #11938, opened 2026-08-01
fix(terminal): stop cancelled IME compositions replaying stale preedit

### LINKED ISSUE #11936: [Bug]: macOS Pinyin IME — first keystroke leaks into the terminal; Backspace over preedit leaves a stray letter (regression since #11293)
## Description

Cancelling a Pinyin composition with Backspace in the integrated terminal leaves the first keystroke of the composition on the shell line (macOS built-in Simplified Chinese Pinyin IME). Committing the composition normally is unaffected. Regressed from v1.4.162.

## Steps to reproduce

1. macOS, built-in Simplified Chinese — Pinyin input source (`com.apple.inputmethod.SCIM.ITABC`).
2. Focus the integrated terminal (plain shell prompt).
3. Type `ceshi` — the IME preedit shows `ceshi` as expected.
4. Press Backspace 5 times to delete the whole preedit.

**Expected:** preedit fully cleared, shell line unchanged (this is the v1.4.162 behavior).

**Actual:** the preedit clears, but a literal `c` remains on the shell line. Reproduces 100% of the time.

## Environment

* Orca: main @ 340faaa83 (v1.4.163 line); regressed from v1.4.162 (official)
* macOS 26.2, Apple Silicon
* IME: macOS built-in Simplified Chinese — Pinyin
* Electron 43.1.0, @xterm/xterm 6.1.0-beta.287 (patched) — both unchanged across the regression window

## Regression window / suspected cause

v1.4.162 → main@340faaa83 contains \~200 commits; the only change touching terminal input/composition is stablyai/orca#11293 **(fe6f929c6) "fix(terminal): reconcile cross-platform IME composition lifecycle"** (+890 lines to the xterm composition patch, keyboard-handlers rework, new composition-session routing).

The commit-vs-cancel asymmetry narrows the mechanism: the first keydown of a composition appears to be forwarded to the PTY, and the new composition-session reconciliation (`dataPendingReconciliation` in the session routing) cleans it up when the composition commits — but the Backspace-cancel path (compositionend with empty data) skips that reconciliation, so the forwarded first keystroke is left behind.

Possibly related: stablyai/orca#11878 (Windows Korean 쌍자음 triggering line breaks, opened 2026-08-01) looks like another composition-lifecycle symptom in the same window.

Him188:
Fixes #11936.

## Root cause

Backspacing away an entire preedit (e.g. macOS Simplified Chinese Pinyin: type `ceshi`, press Backspace 5 times) ends the composition with empty `compositionend` data, an empty textarea, and no `input`/`keypress` events. `_sendPendingComposition` (added in #11293) then falls back to the last non-empty `compositionupdate` data — the stale preedit remnant `c` — and emits it to the PTY as if it were committed text. Committing normally is unaffected because `compositionend` carries the candidate, so the fallback chain never reaches the stale value.

## Fix

One-line semantic change in the patched `CompositionHelper._sendPendingComposition`: only fall back to `pending.compositionData` when observed input evidence (`input`/`keypress` data) corroborates that the composition actually committed. A composition that ends with no evidence in any channel — textarea, end data, input, keypress — was cancelled, and nothing is emitted. Applied to `lib/xterm.js`, `lib/xterm.mjs`, and the reference `src/browser/input/CompositionHelper.ts` in the existing `@xterm/xterm` patch; the patch file was regenerated by `pnpm patch-commit`, which also normalized some formatting-only hunks (hence the net line-count drop).

The same cancellation shape exists on Linux IBus/fcitx (Backspace over the whole preedit fires `compositionend` with empty data after a `deleteContentBackward`), so this covers that path too.

## Tests

New `terminal-ime-xterm-composition-cancel.test.ts` drives the real patched xterm build (same harness style as `terminal-ime-xterm-composition-deduplication.test.ts`):

- Backspace-cancel of a full Pinyin preedit emits nothing — fails with `expected [ 'c' ] to deeply equal []` without the fix, reproducing the report exactly;
- cancel variant without observed input events emits nothing;
- control: an empty-end commit that delivers text via a following bare `insertText` (IBus Hangul shape) still emits, guarding against an over-broad fix.

All four existing IME suites pass (72/72): `terminal-ime-xterm-composition-deduplication`, `terminal-ime-xterm-transaction-events`, `keyboard-handlers-ime`, plus the new file. Manually verified on macOS 26.2 with the built-in Pinyin IME: Backspace over the preedit leaves the shell line clean, and candidate commits are unchanged.


OrcaWin:
Landed via #12273 with the original commit authorship preserved. Thank you for the focused root-cause analysis, regression test, and fix.

PR #11939, opened 2026-08-01
fix(runtime): stabilize remote session-tab sync across server and client

Him188:
## Problem

When a client is attached to a remote orca server (`orca serve` over the pairing relay), session-tab management is unreliable in ways local desktop use never shows. All of the symptoms below are timing-dependent, so they reproduce intermittently rather than on demand, and they get markedly worse the busier the server is (more session activity means more snapshot publications in flight, which widens every race window) and the more devices are paired. Reproduced on a macOS headless serve host with one and with two paired desktop clients.

- **Creating a session can bump an existing tab off the strip.** Launching a new session sometimes removes the last tab in the worktree. The removed tab is not closed: its agent/PTY keeps running server-side and keeps consuming resources — it has just vanished from the client's mirror, invisibly, with no UI path to bring it back until some unrelated mutation forces a fresh snapshot of that worktree.
- **Switching worktrees flickers or wipes the session list.** Activating a worktree sometimes shows its sessions for a moment and then removes all of them; other times only a subset (in practice the serve/SSH-owned terminals) ever appears, and the remaining sessions never materialize even though they are alive on the host.
- **Closed sessions come back.** Closing a tab looks successful, but after restarting the local client (server untouched) — or sometimes merely re-activating the worktree — the closed tab is back, its session still running on the host. In the worst variants the close never happened server-side at all, and the client showed no indication of the failure.
- **Multiple devices amplify everything.** With two clients attached, closing a tab on one device can freeze the other device's view of that worktree — it silently drops every subsequent snapshot until an unrelated publication lands — and in one deterministic case (a PTY left unbound by a failed post-restart surface restore) a close performed on one device is later resurrected for every device.

The common thread: the client mirrors host session-tab snapshots, and each mechanism meant to keep that mirror consistent — publication-epoch freshness gating, close intents, terminal orphan recovery — assumes conditions the remote/headless path does not actually provide.

## Root causes (each traced end-to-end in code)

1. **The client freshness gate is vacuous against headless servers.** The gate rejects only same-epoch, non-increasing `snapshotVersion` frames — a different `publicationEpoch` always applies (`web-session-tabs-sync.ts`), by design assuming one epoch per host boot (as desktop mode does). The headless server minted a fresh `headless:${Date.now()}` epoch on nearly every mutation (10 sites) plus `headless-hydrated:` and `:headless-merge:` variants, so every in-flight frame was "a new generation" and whichever applied last won. Subscription frames serialize through the orphan-recovery chain, but `session.tabs.list` results apply outside it, so a delayed push frame regularly landed after a newer list result and authoritatively culled fresh tabs.
2. **Not-yet-hydrated worktrees are served as authoritative-empty.** A missing `mobileSessionTabsByWorktree` entry fabricates `{publicationEpoch:'none', tabs:[]}`, which the worktree-activation list path applied with no guard — full wipe until the next real publication. Separately, the runtime-owned (serve/SSH-only) hydrate seeds a subset entry that permanently blocks the full rebuild via an `existing.tabs.length>0` skip — subset lock-in.
3. **User closes are fire-and-forget with volatile guards.** Close intents are in-memory with a 10s TTL; a failed close RPC only `console.warn`s, and restart re-lists the host-authoritative snapshot, deterministically re-materializing the tab. One environment's reconnect also wiped every environment's freshness floors and intents.
4. **No cross-device close tombstone.** After device A closes a tab, device B's orphan recovery sees a still-dying PTY absent from the snapshot and drops every frame for that worktree; a PTY already unbound at close time (failed post-restart restore) escapes the kill entirely and gets re-adopted, resurrecting the closed session fleet-wide.
5. **Per-client navigation projection corrupts the epoch echo.** Projected snapshots carry `${epoch}:client-navigation`, but `session.tabs.closeLifecycle` compared the client's echo against the raw epoch with `!==`, so any projected echo was refused 'stale-publication' and triggered a cluster republish (latent at HEAD; live with older clients).
6. **QuickLaunch's stale-tab prune closes live provisional tabs.** It force-closed every local-UUID `launchAgent` tab pre- and post-create with no liveness/handoff check — hibernation-resumed provisional tabs (appended last in the strip) matched, and with a bound handle the close killed the host session: the one path where creating a session literally destroys a neighbor.
7. **Some server close paths ack without closing.** A headed hub with its window closed no-ops markdown/file closes but returns `{closed:true}`; unbound daemon PTYs escaped the kill scope ("hidden, not closed").
8. **After a host restart, restored tabs never reattach the daemon output stream, and provider snapshots leak a foreign sequence domain.** The boot inventory restores session-tab records as `ready`, but the daemon only streams a session's output to the process that last ran `createOrAttach` — and nothing in the new serve process ever does (headless serve has no renderer, so the renderer spawn/reattach paths that normally do this are unreachable). Subscribers get correct snapshots and zero live bytes: typing executes, but output only appears when a tab switch refetches a snapshot. Compounding it, provider buffer snapshots translated their seq with a defaulted-to-0 offset, publishing the daemon's cumulative counter while live frames carry the boot-local counter reset at startup — so even once a spawn path did attach, the client dedupe baseline armed from the snapshot's foreign seq silently dropped every live frame as already-covered, re-poisoning itself on each snapshot refetch.

## Fixes

**Server** — new `MobileSessionTabsPublicationClock` (`mobile-session-tabs-publication.ts`): one stable epoch per worktree per boot, strictly monotonic `snapshotVersion` across all publish sites (mutations, hydration, merge, recovery); the `:headless-merge:` epoch flip becomes a boot-stable merge signature riding version bumps. Missing-entry results now carry `hydrationPending: true` until a full hydrate (or renderer batch, which is hydration on a headed hub) covers the worktree; the full hydrate rebuilds over runtime-owned-seeded subset entries, preserving live runtime-owned bindings. New `SessionTabCloseTombstoneStore` (30s TTL) records every successful close; `recentlyClosedTabIds` rides every published snapshot, `terminal.list` hides tombstoned PTYs, adoption refuses them, and the unbound-PTY kill scope widens to persisted bindings corroborated by pane incarnations (positive incarnation mismatch is never killed). `closeLifecycle` normalizes a trailing `:client-navigation` before the stale-publication comparison. The notifier-less close branch throws `renderer_unavailable` instead of faking success.

**Client** — `session.tabs.list` results now apply through the same per-worktree serialized chain as subscription frames (gaining its orphan-recovery protection); the post-create refresh uses the after-current-in-flight variant; the agent-handoff confirm only arms when the confirm list actually contains the host tab. Failed user closes recover visibly (accept-replay + refresh + toast) instead of silently diverging; sync tracking, intents, and wake-respawn guards are scoped per environment; `hydrationPending` frames (and `'none'`-epoch empties from older servers, including their `:client-navigation` projections) never clear a non-empty mirror; tombstone-covered candidates let post-close frames apply instead of freezing the worktree. QuickLaunch's prune skips tabs with a recorded handoff or pending create, only prunes tabs that existed before the operation, and never fires host `terminal.close` for pruned local tabs.

**Server, host restart** — `session.tabs.activate` now re-materializes a `ready` tab whose PTY has no provider attachment this boot (`ptyController.hasPty(...) === false`), which reattaches the daemon output stream and runs the provider sequence sync as one operation through the existing spawn pipeline. Provider buffer snapshots omit `seq` until a sequence sync has anchored the provider offset for this boot — clients already treat a seq-less snapshot as baseline-clearing, so live frames flow instead of being dropped against a foreign-domain watermark — and the provider visible-state cache watermarks with the pre-capture boot-local counter so a raced byte invalidates it instead of the foreign seq pinning it fresh forever.

## Compatibility

Both new protocol fields (`hydrationPending`, `recentlyClosedTabIds`) are optional: old clients ignore them (no worse than today), new clients degrade to current behavior against old servers. The stable-epoch change makes old clients' existing gate work correctly rather than requiring anything of them.

## Tests

Each mechanism has regression coverage: publication-clock unit tests (stable epoch, monotonic versions across interleaved sources, reboot behavior), runtime-level epoch-stability sweeps across mutation entry points, hydrationPending/subset-rebuild/authoritative-empty cases, tombstone store + publish + list-hiding + adoption-refusal + unbound-PTY kill, projected-epoch closeLifecycle acceptance, client apply-ordering/serialization, close-failure recovery, per-environment isolation, QuickLaunch prune scenarios, and the host-restart cases (re-materialize on a detached `ready` tab vs spawn-free activation while attached; seq omitted while unanchored vs translated once synced). All touched suites plus `pnpm run typecheck` are green after each commit.

## Real-world verification

Deployed on the original repro environment: `orca serve` running on a Mac mini (M4), with a MacBook attached to it remotely over the pairing relay. With this branch on both ends, the session-loss symptoms no longer reproduce — creating sessions does not bump existing tabs, worktree switching shows the full session list without flicker or wipes, and closed sessions stay closed across local client restarts. Continued use surfaced the host-restart defect (root cause 8): with sessions now surviving a restart of both ends, the restored terminals executed input but rendered no live output until a tab switch. That fix is included in this branch and was verified in the same environment.

### DIFF
--- a/src/main/runtime/mobile-session-tabs-publication.test.ts
+++ b/src/main/runtime/mobile-session-tabs-publication.test.ts
@@ -1,0 +1,93 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { MobileSessionTabsPublicationClock } from './mobile-session-tabs-publication'
+
+const WORKTREE = 'repo::/worktree'
+
+describe('MobileSessionTabsPublicationClock', () => {
+  afterEach(() => {
+    vi.useRealTimers()
+  })
+
+  it('keeps one epoch across mutations and strictly increases the version', () => {
+    const clock = new MobileSessionTabsPublicationClock()
+    const first = clock.next(WORKTREE)
+    const second = clock.next(WORKTREE, first)
+    const third = clock.next(WORKTREE, second)
+    expect(first.publicationEpoch.startsWith('headless:')).toBe(true)
+    expect(second.publicationEpoch).toBe(first.publicationEpoch)
+    expect(third.publicationEpoch).toBe(first.publicationEpoch)
+    expect(second.snapshotVersion).toBeGreaterThan(first.snapshotVersion)
+    expect(third.snapshotVersion).toBeGreaterThan(second.snapshotVersion)
+  })
+
+  it('reuses the stored snapshot epoch over minting, whatever its source', () => {
+    const clock = new MobileSessionTabsPublicationClock()
+    const stamp = clock.next(WORKTREE, {
+      publicationEpoch: 'renderer:abc',
+      snapshotVersion: 7
+    })
+    expect(stamp.publicationEpoch).toBe('renderer:abc')
+    expect(stamp.snapshotVersion).toBe(8)
+  })
+
+  it('survives snapshot-map deletion: boot epoch is reused and the floor holds', () => {
+    const clock = new MobileSessionTabsPublicationClock()
+    const first = clock.next(WORKTREE)
+    const second = clock.next(WORKTREE, first)
+    // Entry deleted: no `existing` — the boot epoch and version floor persist.
+    const revived = clock.next(WORKTREE)
+    expect(revived.publicationEpoch).toBe(first.publicationEpoch)
+    expect(revived.snapshotVersion).toBeGreaterThan(second.snapshotVersion)
+  })
+
+  it('applies the mint prefix only to the boot-first mint', () => {
+    const clock = new MobileSessionTabsPublicationClock()
+    const hydrated = clock.next(WORKTREE, undefined, { mintPrefix: 'headless-hydrated' })
+    expect(hydrated.publicationEpoch.startsWith('headless-hydrated:')).toBe(true)
+    const later = clock.next(WORKTREE, undefined, { mintPrefix: 'headless:pty-backed' })
+    expect(later.publicationEpoch).toBe(hydrated.publicationEpoch)
+  })
+
+  it('keeps versions strictly increasing across interleaved publish sources', () => {
+    const clock = new MobileSessionTabsPublicationClock()
+    const versions: number[] = []
+    versions.push(clock.next(WORKTREE).snapshotVersion) // mutation on empty
+    versions.push(clock.nextVersion(WORKTREE, versions.at(-1))) // hydrate rebuild
+    clock.observe(WORKTREE, 50) // renderer-accepted stored version
+    versions.push(clock.nextVersion(WORKTREE, 2)) // touch over a stale entry
+    versions.push(
+      clock.next(WORKTREE, { publicationEpoch: 'e', snapshotVersion: 10 }).snapshotVersion
+    )
+    for (let i = 1; i < versions.length; i++) {
+      expect(versions[i]!).toBeGreaterThan(versions[i - 1]!)
+    }
+    expect(versions.at(-1)!).toBeGreaterThan(50)
+  })
+
+  it('observe never lowers the floor', () => {
+    const clock = new MobileSessionTabsPublicationClock()
+    clock.observe(WORKTREE, 20)
+    clock.observe(WORKTREE, 5)
+    expect(clock.nextVersion(WORKTREE)).toBe(21)
+  })
+
+  it('scopes version floors per worktree', () => {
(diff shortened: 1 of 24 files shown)


Jinwoo-H:
Thanks @Him188 for the broad remote session-tab synchronization investigation. The underlying cases are now covered by the merged owner fixes #16546, #16748, #16750, #16751, #16956, #16969, and #17064. We are retaining your contribution as historical credit; no contributor follow-up is requested. Closing this superseded branch.

PR #11954, opened 2026-08-01
fix(crash): name what the renderer heap holds before it OOMs

nwparker:
## Description

Crash cluster D — **54 renderer OOM aborts** (`reason: crashed`, exit code 5 = V8 fatal OOM), median **94% of a 4,192MB heap limit**, overwhelmingly macOS.

**This PR does not fix the crash. It ships instrumentation, and says so plainly.** I could not reproduce a leak, and rather than ship a speculative fix I am shipping the instrument that will make the next wave of bundles conclusive.

The honest position: across 51 diagnostics bundles the renderer is dying at ~3.9GB, and **the existing telemetry cannot say what is holding the memory**. The store profile records collection *counts*; the heap is being consumed by per-object *size*. Every counter we have can be flat while the heap grows by a gigabyte, and that is exactly what the bundles show.

What this adds to the one-shot `renderer_memory_highwater` breadcrumb:

1. **Every field `process.getHeapStatistics()` exposes that we were discarding** — physical, available, executable (code space) and peak malloced. A per-space breakdown is *not* reachable: `v8.getHeapSpaceStatistics()` requires the `v8` module, and a sandboxed preload cannot `require` it. Verified against Electron 43: every `BrowserWindow` in this repo sets `sandbox: true`, and `createMainWindow.ts:480` force-sets it again on webview attach. This is as rich as the instrument can be.
2. **The zustand subscriber count**, which was already computed and thrown away.
3. **A census of editor file bodies held in React state** (`editorContent.files` / `editorContent.chars`) — memory the store-only profile is structurally unable to walk.
4. **A census of serialized terminal scrollback** (`terminalScrollback.chars` / `.buffers` / `.coldRestoreChars`) — see below.
5. **A census of the terminal output backlog** (`terminalOutputBacklog.*`) — a module-global `Map` in the output scheduler holding PTY output the renderer has received but xterm has not yet parsed. Capped **per terminal** (2MB at default scrollback, 6MB at 50k rows) with **nothing bounding the aggregate**, so N background terminals whose PTYs keep writing while Chromium throttles their drain retain N × that cap in plain JS strings. 170 × 2MB ≈ **340MB**, and it is not store state, so no store instrument could ever have seen it.
6. **A census of the live xterm buffers** (`liveTerminalBuffers.lines` / `.cells` / `.altScreenPanes`) — every retained row is a `BufferLine` holding a `Uint32Array` of 3 words per cell. This was previously reported as a pane *count*, which is the same count-versus-size failure one level down. `lines`/`cells` measure the **normal** buffer; see round 4 for why that distinction is the single most consequential correction this branch received.
7. **A census of Monaco's global model registry** (`monacoModels.*`) — the sharpest of the set, see below.
8. **A census of diff-tab bodies** (`editorContent.diffTabs` / `.diffChars`) — a diff tab retains `originalContent` *and* `modifiedContent`, so a review session holds two full file bodies per tab. Counted separately from item 3 rather than folded into it, because summing them would make one number mean two different things.

Plus `profile.truncated`, which marks a profile that dropped counts rather than letting a short one read as a complete one, and `unreportedCollections`, which says how many store slices the 20-slice limit hid.

### Why the Monaco census matters most

A Monaco text model is disposed only by `useClosedEditorTabCleanup`, on tab close. **Unmount a panel and the model survives** — holding a piece tree plus per-line tokenization and decoration state, several times its text length. Item 3 above measures the strings a *mounted* React panel holds, so for exactly the leak that matters — panels unmounted, models retained — `editorContent.chars` reads ~**0**. 214 models × ~425KB ≈ **270–550MB** reported as nothing at all.

### Why the terminal scrollback census is the important one

`summarizeStateCollectionSizes` counts own keys and **does not recurse** (`renderer-memory-profile.ts:106`). So the profile emits `store.terminalLayoutsByTabId = 170` and never sees what those 170 entries hold. What they hold is:

```ts
/** Serialized terminal buffers per leaf for scrollback restoration on restart. */
buffersByLeafId?: Record<string, string>
```

Each of those strings is capped at `TERMINAL_SCROLLBACK_SESSION_BUFFER_BYTE_LIMIT = 512 * 1024`. The count is per *tab*, and each tab holds N leaves via splits — so the number of 512KB-capped buffers is the reported number times an unreported split factor. `pendingColdRestoreByPtyId: Record<string, {scrollback: string, cwd: string}>` has the same shape and no cap in its type.

This is the count-versus-size problem in its sharpest form: **the codebase caps these strings at 512KB — an explicit acknowledgement that each entry is large — while the only instrument reports how many there are.** A reported 170 is consistent with kilobytes or with roughly a gigabyte. That is precisely the ambiguity a crash bundle has to resolve, and it is exactly analogous to `editorContent.chars`, pointed at the collection the profile already flags as the largest.

### The reproduction was NEGATIVE, and that is the finding

I instrumented a live renderer via CDP and watched it. The initial signal looked alarming — a 284MB/hour slope — and it was an artifact: the counter being sampled was pre-GC.

Two GC-forced heap snapshots, 9 minutes apart:

| | used heap |
| --- | --- |
| snapshot 1 | 100.2 MB |
| snapshot 2 (+9 min) | 99.7 MB |
| **delta** | **−0.3 MB** |

No constructor moved as much as 256KB between snapshots. **An idle renderer on current `main` does not leak.**

That result is correct and it is also nearly vacuous for this cluster, which is worth stating rather than hiding: the idle repro had no agent sessions, no terminals and no SSH — i.e. it excluded every subsystem whose per-item payload grows. It rules out a passive background leak. It does not rule out the actual cause.

### What the 51 bundles actually say

The growth is **per-object size with flat counts**. Median heap growth from the 60% threshold to the 80% threshold is **835MB** (min 410, max 1117) while store counters barely move; in 13 of 43 sessions the counters are *completely unchanged* between the two thresholds. `mallocedMB` is `1` in every sample.

This is precisely the gap the new fields target. Note the 835MB is the 60%→80% *delta* — growth observed while the app was already large. It is not the whole 3.3GB, and whatever the census names still has to account for the baseline.

### Field confirmation, from a report that landed after the scan cutoff

Report `76c4f5b9-43b8-4077-8ea5-1ac21a756604` — macOS, 1.4.163, `renderer`, `reason=crashed`, **exit code 5** (V8 fatal OOM). The final interval sample before death:

```
renderer_memory  reason=interval  usedHeapMB=3004  totalHeapMB=3126  heapLimitMB=3188
                 heapSource=v8    mallocedMB=1     blinkAllocatedMB=6
```

**94% of ceiling, with Blink at 6MB and malloc'd at 1MB.** Whatever is holding this heap is V8 objects and nothing else. The store census in the same bundle names **505 entries across 20 `store.*` collections**. Nothing in the shipped instrumentation comes within three orders of magnitude of accounting for 3GB. That is this PR's thesis observed on a user's machine rather than argued from the bundles.

A second report, `5904df3e-41fc-4c3a-b712-87138e01bb3b`: `usedHeapMB=2414` against a 4192MB limit, `blinkAllocatedMB=144`, breadcrumbs dominated by `pr_refresh_queue (enqueued=2806, coalesced=1103, backgroundPauses=1392)`. Those queue numbers are a **rate** signal, not a retention one: `prCache` holds `PRInfo`, which is scalars plus a changed-file path list with no diffs and no comment bodies, so its 109 entries are kilobytes. Unbounded churn against a keyspace bounded by worktrees × repos produces garbage, not retained heap.

**What `terminalElements=2` in the first report does and does not rule out.** It is a DOM count — `document.querySelectorAll('.xterm').length`. A pane the pane manager still holds after its React component unmounted has **zero** DOM elements and a full scrollback buffer, so that counter is structurally unable to see the retention `liveTerminalBuffers` was added to find; the new census walks live managers, not the DOM.

It does, however, bound how much terminals can explain, and the bound cuts against them. Measured on `@xterm/headless`: a buffer line allocates `3 × cols` Uint32 entries — **12 bytes per cell**, confirmed at 80, 200 and 400 columns — and scrollback grows **lazily** rather than being preallocated (write 100 lines, retain 101; write 20,000 against a 5,000-row limit, retain 5,024). So a *fully scrolled* 200-column pane holds **12.06MB** at the 5,000-row default and 120MB at the 50,000-row maximum. Reaching 3,004MB from terminal buffers alone would need roughly **250 fully-scrolled default panes, or 25 at maximum scrollback** — against a `terminalElements` maximum of 15 across the bundles.

Terminals are therefore **not** a sufficient explanation at default settings and only a borderline one at maximum scrollback. Stated plainly because the arithmetic cuts against the instrument this PR spent the most effort on: `liveTerminalBuffers` should be read as the thing that will finally rule terminals *in or out*, not as the thing expected to name the culprit.

### Instrumentation cost

Adversarially verified as one-shot and O(1), not asserted:

- `collectRendererMemoryProfileCounts()` is called **only** inside the one-shot highwater branch — at most twice per renderer session, never on the 60s sampler. `emittedHighwaterRatios` makes each threshold fire once.
- `measureEditorFileContents` iterates `fileContents` keys and reads `.length`, which is O(1) on a JS string. No traversal, no concatenation, no allocation beyond one result object. It touches only files with *loaded* bodies, and `usePruneClosedEditorContent` already evicts closed tabs — proven by probe: open A, switch to B, close A, and `fileContents` keys are `['/repo/b.ts']`. The map is bounded by currently-open-and-visited files, not by tab count.
- **No breadcrumb ring pressure at all.** `renderer_memory_highwater` is diverted into `retainedBreadcrumbs` (`MAX_RETAINED_BREADCRUMBS = 4`, keyed `surface:thresholdPct`) and returns *before* `breadcrumbs.push` — so it never consumes one of the 30 ring slots. That is stronger than coalescing.

### Each instrument fails when it is wrong, not merely when it is absent

Mutations run against the committed branch, all confirmed caught:

| Mutation | Result |
| --- | --- |
| backlog census reports `peakQueuedChars` instead of live `queuedChars` | ❌ **caught** — the drained-state assertion fails |
| remove the scheduler's `setTerminalOutputBacklogCensusReader` wiring entirely | ❌ **caught** — `expected { terminals: 0, chars: 0 } to deeply equal { terminals: 2, chars: 30 }` |
| `profile.truncated` never set | ❌ **caught** — 4 tests fail |

The first is the one worth dwelling on. The census unit test installs a **fake** reader, so it proves the profile plumbing and says nothing about whether the number in a real crash report is real. The added scheduler test drives actual output through the real queue and reads the reader the scheduler installed itself — the path a breadcrumb actually takes. A reader wired to a peak counter instead of the live queue passes the naive assertion and misreports a drained renderer as still holding the backlog; that mutation is now red.

### Gates

- **42,857 passed**, 80 skipped, 4,040 files
- `tsc -p config/tsconfig.tc.web.json --noEmit` → **exit 0**
- `oxlint --type-aware` → **zero errors in any file this branch touches**; `oxfmt --check` clean
- The only repo-wide failures are 2 pre-existing machine-local ones in `src/relay/agent-exec-handler.test.ts` (relay spawn env), which renderer-only changes cannot affect

## ELI5

Orca's renderer is running out of memory and dying. We have a memory gauge, but it only counts *how many things* there are — not *how big* they are. So the gauge reads "everything normal" while the app fills up with a gigabyte of something.

We went looking for a leak and didn't find one: left alone for nine minutes, memory went *down* slightly. That's a real answer, but it only rules out leaks that happen when the app is doing nothing.

So instead of guessing at a fix, this adds the missing markings to the gauge — including one that finally measures how many characters of file text the editor is holding. The next crash report will be able to name the culprit instead of shrugging.

## Trade-offs

1. **This does not fix anything.** 54 users will keep crashing until the follow-up. The alternative was shipping a speculative fix, and this PR originally did exactly that — see below.
2. **It may still be blind.** The censuses now cover editor file bodies, serialized terminal scrollback, the terminal output backlog, live xterm buffers and Monaco models. If the mass is **agent-session transcripts** instead, those are still reported as element *counts* only and the bundle will not name them. That is the known remaining hole.
3. **`*.chars` is a 2×-ambiguous byte estimate.** A JS string is 1 byte/char in latin1 and 2 in UTF-16, so a char count brackets the byte cost rather than pinning it. Still an order of magnitude better than an element count.
4. **Two store fields are measured that can never be non-zero.** `pendingColdRestoreByPtyId` has no writer — cold restore goes straight to xterm via `writeReplayData` — proven by a rename mutation producing exactly 3 tsc errors and zero production writers. Kept deliberately and documented as a tripwire, so a future writer is measured on arrival rather than silently unmeasured. A reviewer may reasonably want them deleted instead; I did not, because removing store fields that touch terminal-restore semantics is out of scope for a crash PR.
5. **`resetEditorContentCensusForTesting` is a test-only export in production code.** The reader cap is process-wide, so suites must be able to clear it. Small, real smell.
6. **Registration order decides who gets dropped** if the 64-count budget is ever exceeded, and that order is module-import order. `profile.truncated` announces the loss but does not prevent it. The budget is at 39/64 today, and the instruments added here are what push it toward the cap.
7. **The extra fields** consume part of the 64,000-char report truncation budget. Measured at 9,702 chars — **15.2%** of budget, up from 1.09% before the terminal fields. Still leaves 85% for other evidence, but this is no longer a rounding error and a further instrument should be weighed against it rather than added freely.
8. **Reading the normal buffer under-counts an alt-screen pane's live rows.** `lines`/`cells` now measure `buffer.normal`, so the viewport currently owned by an alt-screen app is not counted. That is bounded at viewport-height rows per pane — tens, against the thousands of scrollback rows the change stops losing — and `altScreenPanes` says how many panes are in that state, so the under-count is visible rather than silent. Accepted deliberately: the alternative measures the small number confidently and the large one not at all.
9. **Two different signals now mean "something is missing".** `profile.truncated` means a whole contributor was dropped; `unreportedCollections` means the store walk hid some slices. Deliberately kept apart — see round 4 — but a reader has to check both, and only one of them is a flag.
10. **One ref write moved from render into an effect** (`useEditorPanelContentState.ts`) to satisfy React Doctor. The census now reads a value that trails the committed render by one pass. Argued safe above rather than asserted, but it is a real timing change in a file whose primary purpose is not diagnostics.
11. **This PR improves *what* is measured, not *when*.** `collectRendererMemoryProfileCounts()` has exactly one call site, inside the one-shot highwater branch, and each ratio in `RENDERER_MEMORY_HIGHWATER_RATIOS = [0.6, 0.8]` latches in `emittedHighwaterRatios` and never re-arms. The breadcrumb carrying the census is therefore structurally the **first** sample above its threshold, never the last. On report `76c4f5b9` the 80% breadcrumb observed **2605MB** — 2605/3168 = 82.2%, i.e. the first crossing, exactly as designed — while the final interval sample before death read **3004MB**. **The census is 400MB stale at the moment it matters most**, and the 60-second interval samples that *do* bracket the death carry no census at all. This is not a regression introduced here, and closing it is out of scope, but a reader of the next OOM report needs to know the contents were measured 400MB before the peak. It is cheaply closable: `retainedBreadcrumbs` replaces in place on a repeated key, so re-arming the top threshold on a growth delta would refresh that same retained entry with a later census at **no ring cost and no additional retained keys** — strictly better than adding a third ratio, which would cost a retained key per surface and start evicting at two surfaces. Left out because it changes emission behaviour rather than adding a reader.

## What was removed from this PR, and why

The first version of this branch also shipped a **100-open-editor-tabs-per-workspace cap** on session restore, motivated by one bundle showing 3,711 tabs. Adversarial review killed it on three independent CONFIRMED grounds, and it has been removed in full:

1. **It was not restore-only — it permanently destroyed user data.** Capping on restore means the store only ever holds the surviving 100. The save path does not merge with disk: `buildPersistedOpenFiles` rebuilds `openFilesByWorktree` wholesale from the in-memory array (`workspace-session.ts:114`), and the main-side handler top-level-spreads the patch with no per-key merge (`persistence.ts:6226`). A user with 3,711 tabs would restore 100, quit, and lose the other 3,611 **from disk, permanently, with no undo and no notice**. The commit message said "cap restored editor tabs at 100" and never said "and delete the rest".
2. **Its central design claim was false.** The cap's own doc comment asserted that capping `openFilesByWorktree` cascades to tabs and tab groups. Proven false by executed test: 150 persisted files capped to 100 still hydrated **150** unified editor tabs — 50 orphans, `tabOrder` still 150. Hydration only backing-file-checks *transient* content types (`tabs-hydration.ts:95`), and a plain editor tab is not transient. The user would get thousands of tabs that render nothing when clicked, re-orphaned on every boot, and the cap would never converge.
3. **The bundles refute the premise.** Open-file count is **uncorrelated** with this OOM. Where the counter is recorded at the 80% highwater, mean used heap is **3372MB for the ≤200-open-file cohort** versus **3260MB for the >500 cohort** — *higher* for the smaller one. Roughly half the crashing sessions have single-digit open files. The 3,711-tab bundle that motivated the cap is an outlier whose heap is no worse than the six-open-file sessions.

It was a count fix for a size problem, aimed by the same blind instrument this PR exists to replace. Worth recording as a caution: the evidence that killed it was available before it was written.

## Performance review

This PR adds instrumentation and reduces memory by **zero bytes**, so its cost is the only thing weighable against it. All four new instruments were measured against real objects — real `monaco.editor.createModel()` models, the real scheduler queue, the real pane registry — not mocks.

**Verdict: ship, nothing needs gating.**

### Frequency first, because it dominates everything else

`collectRendererMemoryProfileCounts()` has exactly **one** non-test caller: `crash-diagnostics.ts:188`, inside `recordRendererMemoryHighwater`, *after* the `if (!crossedThreshold) { return }` guard. Thresholds are one-shot per ratio (0.6 / 0.8), recorded in `emittedHighwaterRatios`. The 60s sampler calls the same function and returns at that guard on every ordinary sample.

**≤ 2 collections per renderer session, ever. Zero on the interval path, zero on any interaction path.**

### Cost per collection

| Scale | Total |
|---|---|
| Realistic — 30 tabs, 24 panes, 25 models, 3 panels, 8k store entries | **202 µs** |
| Adversarial — 400 tabs, 5000 panes, 500 models, 64 panels, 200k store entries | **7.2 ms** |

7.2 ms deserved attribution rather than a shrug, so it got one:

| Contributor | Cost | Share |
|---|---|---|
| `store` (**pre-existing**, not touched by this PR) | 5383 µs | **97.6%** |
| `monacoModels` (new) | 74 µs | 1.3% |
| `liveTerminalBuffers` (new) | 36 µs | 0.7% |
| `terminalScrollback` (new) | 20 µs | 0.4% |
| `terminalOutputBacklog` (new) | 3 µs | 0.1% |

**All five new instruments are 133 µs of 5517 µs — 2.4%.** The 7 ms is the pre-existing `summarizeStateCollectionSizes` walk over 200k store keys. This PR's own worst-case contribution is ~0.13 ms, twice per session.

### The Monaco census — the one expected to hurt, which doesn't

Linear in model **count**, flat in model **content**:

```
scaling in count (200-line models)      scaling in size (50 models fixed)
   10 models ->   1.21 µs               50 x     10 lines (  0.0 MB) -> 5.81 µs
   50 models ->   5.76 µs               50 x    200 lines (  0.7 MB) -> 5.27 µs
  200 models ->  31.47 µs               50 x   5000 lines ( 16.7 MB) -> 6.09 µs
  500 models ->  82.44 µs               50 x 100000 lines (333.8 MB) -> 5.94 µs
 2000 models -> 477.83 µs
```

**10,000× the text, same cost.** Realistic 25–50 open files is 1–6 µs.

### Live terminal buffers, and backlog

Terminal buffers: ~8 ns/pane, linear — 0.19 µs at 8 panes, 39.68 µs at an adversarial 5000. xterm's O(1) claims verified in source (`BufferNamespaceApi.active` returns a cached view; `.length` is a field read).

Backlog census: ~1.5 ns per queued terminal, flat in queued bytes — 0.024 µs at 10 terminals, 3.255 µs at 2000 holding 62.5 MB. It reads the maintained `queuedChars` counter, as designed.

### Cost when *not* collecting — i.e. for every user on every run

`pane-terminal-output-scheduler.ts` installs the census reader as a single top-level closure at module load. `readQueueDebugSnapshot()` runs **only** inside that closure, and the only caller chain is registry → collect → the one-shot highwater path. `recordQueueDebugPressure` is untouched and still `debugEnabled`-gated.

**Zero per-write, per-enqueue, per-drain cost.** Nothing was added to the terminal output hot path.

### Bundle, from two full builds

| | baseline | branch | delta |
|---|---|---|---|
| store chunk | 2,020,231 B | 2,024,015 B | +3,784 (+0.19%) |
| total renderer JS | 54,236,509 B | 54,241,709 B | **+5,200 (+0.0096%)** |
| chunk count | 707 | 707 | 0 |

The comments claiming these imports avoid pulling Monaco and the pane-manager into the store chunk are **correct**: `"monaco"` occurrences in the store chunk go 0 → 1, and that single hit is the census stub's own body, not the library. `pane-manager-registry` was already in the store chunk on baseline and is not duplicated.

**Retention:** no instrument retains what it measures. Editor-content readers close over a React ref and are unregistered by the effect cleanup, capped at 64 with a `droppedPanels` counter; backlog and Monaco readers are module-global closures over already-global state.

### Two corrections the review produced

Both are comment-accuracy issues, neither affects behaviour:

1. `monaco-setup.ts:93` claims `getValueLength()` and `getModels()` are both O(1). Neither is: `getValueLength` is O(log pieces) via two piece-tree offset descents, and `getModels()` allocates two arrays of N. That allocation *is* the measured ~0.12 µs/model floor — so the number is right and the explanation is wrong.
2. `pane-manager-registry.ts:145` reaches panes via `getPanes()`, which materialises two arrays plus a 10-field object per pane. The same file already avoids this at `:106` using `getPaneCount?.()`, with a comment saying exactly why. Direct iteration measures **2.5× faster at 200 panes** (0.80→0.35 µs at 50, 3.14→1.24 µs at 200, 15.97→6.85 µs at 1000; best-of-5). An earlier draft said 3.2×, which was measured against a hypothetical count-only walk rather than the code as written — the shipped version pays a function call and an object return per manager. Absolute saving is small either way, so this is tidiness, not hazard.

### A measurement error worth recording

The first two backlog runs reported 0.00 µs/call at every scale. That was wrong, not fast: the queue must be populated under `vi.useFakeTimers()` to stop drains firing, and Vitest's fake timers also mock `performance.now()`, so elapsed time read as zero. Proof: a 5M-iteration busy loop measures 0.000 ms under fake timers and 7.176 ms on the real clock. The final numbers populate under fake timers, switch to the real clock before measuring, and consume every result through a sink so V8 cannot eliminate the call.

## Adversarial review

**5 rounds. None of them clean.** Round 5's findings are fixed and are described first, because they are the ones that decide whether the numbers in the next OOM report can be believed at all.

### Round 5 — four censuses that were fully green when deleted

Round 5 attacked the one property this PR actually sells: that a key in the profile means *the instrument ran*. It found four places where that was false, all fixed in `c6d84889e4`, `354b87be97`, `91820a79c6`, `ac1886dd65`.

**The behaviour fix — a failing census reported the opposite of what happened.** `getLiveTerminalBufferCensus()` dropped a throwing pane from `panes` with no counter, and skipped an entire manager whose census threw. So a census that failed on *every* pane reported `panes: 0` — which a reader of an OOM report would read as **"no terminals were live"**, when it means "every read threw". Now carries `droppedPanes`, a never-decremented lower bound, matching the `droppedPanels` the editor census already had. `PaneManager.getPaneBufferCensus` also had **no test at all** — it could have returned zeros and stayed green — and is now pinned.

**Three coverage gaps where the instrument could be deleted with a green full suite:**

| Mutation | Before round 5 | Now |
| --- | --- | --- |
| Delete any of the four census side-effect imports in `store/index.ts`, or the `terminalScrollback` registration | **fully green**, against the whole renderer suite | killed |
| Zero out the Monaco model arithmetic | **fully green** — it lived in `@/lib/monaco-setup`, which *every* suite mocks, so it was unreachable under test | killed; arithmetic moved into `monaco-model-memory-census.ts`, `monaco-setup` keeps only the `getModels()` call |
| Drop the editor content hook's registration, or freeze the ref the effect keeps current | **fully green** — the census suite installs a fake reader, which proves the plumbing and nothing about the number | killed, in its own file (the hook suite is already at the 800-line cap; no `max-lines` disable was added) |

That last row is the sharpest statement of the theme: a suite that tests the census *through a fake reader* verifies everything except the thing an OOM report will print.

**I re-ran two of these mutants myself before pushing** rather than taking the report on trust — stubbing out the `terminalScrollback` registration (1 failed / 2), and flipping `droppedPanes += 1` to `+= 0` (1 failed / 20). Both killed, worktree restored clean. Gates at `ac1886dd65`: oxlint 0, web typecheck 0, the four affected suites 28/28.

**Why this round matters more than its diffstat suggests.** This PR's entire value is that the *next* report can name the culprit. An instrument that silently fails to register turns a missing key from "the instrument never ran" into indistinguishable-from-zero — the one failure mode a diagnostic must not have, and the one that would have sent the next investigation down exactly the wrong path. Four of them shipped in the first four rounds of review.

### Round 3 — the instrument was pointed at three subsystems it could not see

Round 3 returned 6 findings. Its **three HIGH findings are the three instruments this PR now ships** (items 5, 6, 7 above): the terminal output backlog, the live xterm buffers, and the Monaco model registry were each uninstrumented, each plausibly in the hundreds of megabytes, and each structurally invisible to a store-walking profile. All three are fixed.

Its MEDIUM-4 proved the `pending*ByPtyId` fields dead by construction — rename mutation, exactly 3 tsc errors, zero production writers — and is deliberately **not** fixed, per trade-off 4. Its MEDIUM-5 (`droppedPanels`) is fixed. Its LOW-6 flagged that contributors are visited in insertion order, which is what `profile.truncated` answers.

Round 3 also produced a method correction worth keeping: running `npx tsc … | grep -v "^npm warn"` returns **grep's** exit code, not tsc's, so a failing typecheck reads as a pass. Every gate above was re-run bare.

### Round 4 — the census was reading the wrong xterm buffer

Round 4 produced the most valuable single finding on this cluster, and it is the kind this PR exists to prevent.

**HIGH — `buffer.active` is the *alternate* buffer whenever an alt-screen app is up** (`\x1b[?1049h` — vim, less, htop, Claude Code). The alternate buffer is viewport-sized and has no scrollback; the retained rows sit in `normal`, unmeasured. Probed against real `@xterm/headless`: write 5000 rows, enter alt-screen, and `normal.length = 5001` while `active.length = 24`.

Why that outranked everything else in the queue: the failure mode is not a missing number, it is a **present, plausible, wrong** one. 30 panes × 50k scrollback with an agent TUI focused would report roughly viewport-height lines per pane, cluster D would read as "terminals are not the leak", and the actual culprit would be ruled out. This PR's whole contract is that a missing key means "the instrument never ran" and a zero means "nothing held" — a confidently wrong number breaks that contract worse than either. The population is not exotic either: on this product, an agent TUI in the foreground is the normal case.

The mutation is what made it actionable rather than arguable — `buffer.active` → `buffer.normal` turned **3 tests red**, which means the tests were pinning the wrong behaviour as correct. So the fix had to move assertions rather than merely add them, and those moves are called out in `pane-manager-registry.test.ts` rather than made quietly. `lines`/`cells` now read `normal`, and a separate `altScreenPanes` reports alt-screen state instead of encoding two facts in one number.

**MEDIUM — truncation of the store walk was invisible, and its magnitude was overstated.** `summarizeStateCollectionSizes(state, limit)` does `sizes.slice(0, limit)` *below* the contributor, so the collection never learned that anything was dropped: the report showed 20 slices with no sign a 21st existed, which reads as "those are all of them". Fixed with `unreportedCollections`.

The round's supporting figure — 204 top-level collections, so 184 dropped unmarked on every collection — **did not reproduce, and is not claimed here.** A fresh store measures 1,050 top-level keys of which only **8** are non-empty collections, so **0** are dropped; `sizes` only admits entries with `size > 0`. The structural gap was real and is fixed. The claim that it fires on every real session is unverified, and the honest position is that we do not yet know how often the limit binds — which is itself an argument for the counter. Note `unreportedCollections` deliberately does **not** set `profile.truncated`: if a real session sits on the limit routinely, a flag that is always 1 can no longer mean "a whole contributor is missing".

**MEDIUM — `editorContent` ignored diff buffers.** A diff tab retains two full file bodies and the census counted neither. Fixed as item 8, as separate keys rather than folded into the file-body sums.

**LOW/MEDIUM — the `getPanes()` allocation and a false O(1) comment** are the two corrections recorded above.

### Round 4 aftermath — a CI failure worth recording

React Doctor failed the branch on `fileContentsRef.current = fileContents` at render time. The rule gates *changed* lines only, so the identical `diffContentsRef` write one line above rode free — which is worth noting, because "the neighbouring line does the same thing" is not a defence a changed-lines rule will ever accept.

Moved into an effect rather than suppressed. The narrow claim that makes this safe: that ref has exactly one reader — the census — and it is read asynchronously off the crash path, so trailing the committed value by one render pass cannot change what a breadcrumb reports. `diffContentsRef` was deliberately left alone: it is read by real logic in this hook where render-time freshness is load-bearing, and changing its timing would be a behaviour change wearing a lint fix as a disguise.

### Round 2 — the removal is clean; the instrument was aimed at the wrong subsystem

Round 2 proved the cap removal complete by mechanical equivalence rather than by inspection: it inlined the one remaining alias out of `App.tsx` with `perl` and diffed the blank-stripped result against `origin/main` — **empty**. So the file is behaviourally identical to main. Repo-wide grep for every spelling of the cap returned zero hits. (It also caught its own error: a first attempt used BSD `sed`, which lacks `\b`, and produced a false mismatch.)

Every mechanical attack on the census failed **under execution**, not under argument: a 268,435,456-char `ConsString` built under a 400MB cap answered `.length` in 333ns allocating 3,424 bytes — `.length` does not flatten a rope; V8 throws `RangeError` below 2²⁹ chars so a single file body is bounded; the new fields cost 700 chars of a 64,000-char budget; and missing fields are double-guarded against `NaN` by `Number.isFinite`.

Its HIGH was that the census pointed at the subsystem round 1 had already **cleared** — editor file bodies, proven to prune correctly on tab close — while the largest collection in the profile got only a count. It then corrected its own first magnitude estimate (it had assumed 170 live xterm panes; the real number is `terminalElements`, max 15) and named a better target: the serialized `buffersByLeafId` strings. That is the census added here, and it is a stronger version of this PR's own thesis than the PR originally made.

Two LOWs and an INFO also fixed: the editor census is now registered from the store so absence means "the instrument never ran" rather than being ambiguous with "no editor was opened", and its reader set is bounded like its siblings so a leaked panel cannot make the diagnostic a retainer of what it measures.

Round 1 could **not** break the instrumentation half, and attacked it seriously: it independently confirmed the sandbox claim across all six `BrowserWindow` call sites, proved the breadcrumb takes no ring slot, proved the census is O(1) per file and one-shot per threshold, and verified all four pure functions are pinned by tests that fail under mutation. It also disproved its own leading hypothesis — that `fileContents` never prunes on tab close — by probe rather than by reading.

Round 1's MEDIUM (not applicable now the cap is gone, but recorded because it is the right design if a cap is ever revisited): "newest 100" was open-order, not use-recency. A file opened on day 1 and used every day since sits at index 0 and gets dropped; 100 files opened once and never revisited survive. `tabGroups[].recentTabIds` is a persisted true MRU and was the correct ranking key.

Round 1's LOW, accepted and corrected here rather than in a rewritten commit message: the claim "across 51 diagnostics bundles the 60% and 80% highwater profiles are byte-identical" is overstated as literally written. The accurate — and stronger — statement is that **in 13 of 43 sessions the store counters are unchanged between thresholds while the heap grows a median 835MB**.


AmethystLiang:
@coderabbitai Fixing. Will be in the next commit

AmethystLiang:
@coderabbitai Fixing. Will be in the next commit

PR #11958, opened 2026-08-01
fix(window): stop the window-close sentinel from blanking the app on Windows

### LINKED ISSUE #11957: [Bug]: Blank window on Windows — createMainWindow throws converting the window-close sentinel to a path
**Problem**

On Windows, Orca built from current `main` opens a window that never renders — the app is blank, with no error surfaced in the UI. The main process logs:

```
[main_unhandled_rejection] TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute
    at getPathFromURLWin32 (node:internal/url:1505:11)
    at fileURLToPath (node:internal/url:1632:35)
    at createMainWindow (out/main/index.js:...)
    at openMainWindow (out/main/index.js:...)
```

**Cause**

`c79b85975` (PR #11910, "fix(browser): prevent window.close guest crashes") added this near the top of `createMainWindow` (`src/main/window/createMainWindow.ts:447`):

```ts
const browserWindowCloseAllowedPreloadPath = fileURLToPath(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)
```

`BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD` is a sentinel marker, not a real file:

```ts
// src/shared/browser-window-close-policy.ts
export const BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD = 'file:///__orca_window_close_allowed__'
```

On POSIX, `fileURLToPath` resolves it harmlessly to `/__orca_window_close_allowed__`. On Windows it throws `ERR_INVALID_FILE_URL_PATH`, because a win32 file URL requires a drive letter. The call is unguarded and runs during window construction, so window creation aborts and the app renders blank.

Reproducible directly:

```
> node -e "require('node:url').fileURLToPath('file:///__orca_window_close_allowed__')"
TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute
```

**The test suite catches this, but only on Windows**

`src/main/window/createMainWindow.test.ts` fails 5+ tests on Windows at `16c5526df`. Guarding that single call drops it to 1, and the last one fails inside the test file itself:

```ts
// createMainWindow.test.ts:286
preload: fileURLToPath(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)
```

So the test constructs the path form the same way the source does, and cannot run on Windows either. That suggests PR #11910 was never exercised on a Windows host or in Windows CI — worth checking whether the matrix covers Windows for these suites, since this class of bug will recur otherwise.

**Impact**

Any Windows checkout at `16c5526df` or later. `a53c5d3fb` predates the commit and is fine. It blanks the entire app, and the only signal is a stack trace in the main-process log, so it is easy to misattribute to whatever branch you happen to be on — which is how I hit it.

**Fix**

Treat "no path form on this platform" as a normal case rather than an error: convert the sentinel through a helper that returns `null` on failure, and compare against it only when non-null. The marker-form comparison already covers every platform, so nothing is lost on Windows.

PR follows.

**Environment**

Windows 11, Orca `main` at `16c5526df`, reproduced in a fresh worktree with a fresh `orca-dev` profile.

dcieslak19973:
Fixes #11957.

## Summary

On Windows, `main` at `16c5526df` opens a window that never renders — the app is blank with nothing surfaced in the UI, only a stack trace in the main-process log.

`c79b85975` (PR #11910) added this near the top of `createMainWindow`:

```ts
const browserWindowCloseAllowedPreloadPath = fileURLToPath(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)
```

`BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD` is a sentinel — `file:///__orca_window_close_allowed__` — not a real file. On POSIX it resolves harmlessly to `/__orca_window_close_allowed__`. On Windows `fileURLToPath` throws `ERR_INVALID_FILE_URL_PATH`, because a win32 file URL requires a drive letter. Unguarded, during window construction, so window creation aborts.

```
> node -e "require('node:url').fileURLToPath('file:///__orca_window_close_allowed__')"
TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute
```

This treats "no path form on this platform" as a normal case: convert through a helper returning `null` on failure, and compare against it only when non-null. The marker-form comparison in the same predicate already covers every platform, so the attach guard loses nothing on Windows — a guest can still be allow-listed by the marker exactly as before.

The helper lives in `src/main/window/` rather than beside the marker in `src/shared/browser-window-close-policy.ts`, because that shared module is also imported by the renderer (`browser-pane/browser-page-webview.ts`) and must not pull in `node:url`.

## Screenshots

No visual change on any platform where the app currently starts. On Windows the change is between a blank window and a working one.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — `src/main/window/` suites; see Notes
- [ ] `pnpm build` — not run locally; relies on CI

**`createMainWindow.test.ts` went from 5+ failures to 81/81 passing on Windows.**

That file could not run on Windows either, because it built the path form with the same unguarded call:

```ts
// createMainWindow.test.ts:286
preload: fileURLToPath(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)
```

It now uses the helper and exercises the path-form branch only where a path exists; the marker-form assertion above it still runs everywhere, so POSIX coverage is unchanged.

New `browser-window-close-preload-path.test.ts` covers the helper. Its guard test uses a non-`file:` scheme, which is rejected on *every* platform — so it exercises the catch on Linux CI rather than passing vacuously there and only failing on a Windows host.

## AI Review Report

Reviewed with Claude Code. Risks checked: whether narrowing the comparison weakens the `will-attach-webview` security guard, renderer bundling of `node:url`, and whether the fix masks a real misconfiguration.

- **Security guard unchanged.** The predicate is `preload === MARKER || preload === markerPath`. On Windows `markerPath` is now `null`, and no string preload equals `null`, so the branch simply never matches there — which is correct, since Electron cannot produce a win32 path form of a URL that has none. The fail-closed `src`/partition allowlist above it is untouched. There is no path where this makes a guest *more* privileged.
- **Renderer bundling.** Verified `src/shared/browser-window-close-policy.ts` is imported by the renderer, so the helper deliberately does not live there.
- **Not masking a misconfiguration.** The sentinel is intentionally not a real file — it is a marker compared by identity. Swallowing its conversion failure is the correct semantic, not a papered-over error. The helper is scoped to this one use rather than a general "ignore all URL errors" utility.

Cross-platform: this *is* the cross-platform fix. Behavior on macOS and Linux is byte-identical (the conversion still succeeds and the path comparison still runs); Windows goes from broken to working. Verified `fileURLToPath` throws for this input on Windows and succeeds on POSIX.

## Security Audit

Touches a security-relevant code path — the `will-attach-webview` guard that decides whether a guest keeps a preload — so worth being explicit: the change only removes a comparison that could never match on the affected platform. No IPC, command execution, path handling, auth, secrets, or dependency changes. No new input reaches any sink; the helper's only caller passes a compile-time constant.

## Notes

Worth checking whether CI runs `src/main/window/` on Windows. These tests fail there at `16c5526df`, so if the matrix covered Windows this would have been caught before merge — and this class of bug (a POSIX-only path assumption) will recur otherwise.

`clipboard-file-copy.test.ts` also fails 4 tests on Windows in this worktree, identically on pristine `main`. Unrelated to this change and left alone, but it looks like the same family of platform assumption.

### DIFF
--- a/src/main/window/browser-window-close-preload-path.test.ts
+++ b/src/main/window/browser-window-close-preload-path.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from 'vitest'
+import { BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD } from '../../shared/browser-window-close-policy'
+import {
+  fileUrlToPathOrNull,
+  resolveBrowserWindowCloseAllowedPreloadPath
+} from './browser-window-close-preload-path'
+
+describe('fileUrlToPathOrNull', () => {
+  it('converts a real file URL', () => {
+    const converted = fileUrlToPathOrNull(
+      process.platform === 'win32' ? 'file:///C:/tmp/x.js' : 'file:///tmp/x.js'
+    )
+    expect(converted).toBeTruthy()
+    expect(converted).toContain('x.js')
+  })
+
+  // Why: this is the guard the window-close marker depends on. A non-file scheme is rejected on
+  // every platform, so it exercises the catch without depending on the host being Windows.
+  it('returns null instead of throwing for a URL that cannot be a path', () => {
+    expect(() => fileUrlToPathOrNull('https://example.com/x.js')).not.toThrow()
+    expect(fileUrlToPathOrNull('https://example.com/x.js')).toBeNull()
+  })
+})
+
+describe('resolveBrowserWindowCloseAllowedPreloadPath', () => {
+  // Why: regression guard for the blank-window bug. Converting the sentinel throws on win32
+  // (a win32 file URL needs a drive letter), and an unguarded call in createMainWindow aborted
+  // window creation, so the app rendered blank on Windows.
+  it('never throws, whatever the platform makes of the sentinel', () => {
+    expect(() => resolveBrowserWindowCloseAllowedPreloadPath()).not.toThrow()
+  })
+
+  it('returns either a path or null, and never the raw marker', () => {
+    const resolved = resolveBrowserWindowCloseAllowedPreloadPath()
+    expect(resolved === null || typeof resolved === 'string').toBe(true)
+    expect(resolved).not.toBe(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)
+  })
+})
--- a/src/main/window/browser-window-close-preload-path.ts
+++ b/src/main/window/browser-window-close-preload-path.ts
@@ -1,0 +1,27 @@
+import { fileURLToPath } from 'node:url'
+import { BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD } from '../../shared/browser-window-close-policy'
+
+/**
+ * `fileURLToPath` for URLs that may not name a real file. Returns null instead of
+ * throwing so a caller can treat "no path form on this platform" as a normal case.
+ */
+export function fileUrlToPathOrNull(fileUrl: string): string | null {
+  try {
+    return fileURLToPath(fileUrl)
+  } catch {
+    return null
+  }
+}
+
+/**
+ * Path form of the window-close allow marker, or null where the platform has none.
+ *
+ * The marker is a sentinel (`file:///__orca_window_close_allowed__`), not a real file.
+ * Electron may hand it back already normalized to a path, so the attach guard compares
+ * against both forms — but win32 file URLs require a drive letter, so converting the
+ * sentinel throws there. Lives in main only: the shared policy module that owns the
+ * marker is also bundled into the renderer, which must not pull in `node:url`.
+ */
+export function resolveBrowserWindowCloseAllowedPreloadPath(): string | null {
+  return fileUrlToPathOrNull(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)
+}
--- a/src/main/window/createMainWindow.test.ts
+++ b/src/main/window/createMainWindow.test.ts
@@ -75,7 +75,7 @@ import {
 import { ipcMain } from 'electron'
 import { shouldRecoverRendererAfterProcessGone } from '../crash-reporting/process-gone-classification'
 import { BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD } from '../../shared/browser-window-close-policy'
-import { fileURLToPath } from 'node:url'
+import { resolveBrowserWindowCloseAllowedPreloadPath } from './browser-window-close-preload-path'
 
 function withPlatform<T>(platform: NodeJS.Platform, run: () => T): T {
   const original = process.platform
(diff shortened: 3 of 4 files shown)


AmethystLiang:
### Linked duplicate PR closed

Closing #12000 as a duplicate of this PR — same Windows blank-window / driveless `file:///__orca_window_close_allowed__` sentinel → `fileURLToPath` / `ERR_INVALID_FILE_URL_PATH` crash in `createMainWindow`.

This PR already targets #11957. Any alternate wording/approach from #12000 is linked there for reviewers.

dcieslak19973:
Closing this as superseded by #12040 — the crash it fixes no longer exists on `main`.

## What happened

This PR guarded the `fileURLToPath(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)` conversion in `createMainWindow`, which threw `ERR_INVALID_FILE_URL_PATH` on Windows because the sentinel `file:///__orca_window_close_allowed__` has no drive letter.

#12040 ("fix(browser): remove unsafe window close bypass", `4f963fd27`) removed that whole code path as a security fix rather than guarding it. It deleted the `fileURLToPath` import, the `BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD` import, the `browserWindowCloseAllowedPreloadPath` conversion, and the `allowWindowClose` bypass — the preload is now always set:

```diff
-const browserWindowCloseAllowedPreloadPath = fileURLToPath(BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD)
...
-    const allowWindowClose = [params.preload, webPreferences.preload].some(
-      (preload) =>
-        preload === BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD ||
-        preload === browserWindowCloseAllowedPreloadPath
-    )
     delete params.preload
-    if (allowWindowClose) {
-      delete webPreferences.preload
-    } else {
-      webPreferences.preload = browserWindowClosePreload
-    }
+    webPreferences.preload = browserWindowClosePreload
```

The sentinel is gone from the tree entirely — `git grep BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD origin/main -- src` returns nothing, and the only surviving `fileURLToPath` in `src/main` is `ipc/shell.ts:194`, which operates on a real parsed file URL.

## Verified on main

Rebasing surfaced this: both conflict hunks had an **empty** `HEAD` side, because `main` had deleted the lines this PR modifies. Checking out `main` and running the suite on Windows:

```
 Test Files  1 passed (1)
      Tests  81 passed (81)
```

81/81 — exactly the outcome this PR's commit message claims to deliver ("that file went from 5+ failures to passing all 81 on Windows"). `main` already gets there.

Rebasing would re-add `browser-window-close-preload-path.ts` and its test as a helper for a call site that no longer exists, so it goes in as dead code. Not worth carrying.

## Note on #12000

@AmethystLiang — you closed #12000 as a duplicate in favour of this PR at 2026-08-01T23:02Z, and #12040 merged at 2026-08-02T04:08Z, about five hours later. That dedupe is now moot: neither this PR nor #12000 is needed, since #12040 removed the code both were fixing. Flagging in case #12000 was carrying anything beyond the sentinel fix that should be revisited on its own.

## #11957

The underlying issue is still open. It is fixed on `main` by #12040 rather than by this PR, so it can be closed once someone confirms the blank window is gone on a current build — the `createMainWindow` suite passing 81/81 on Windows is the evidence I have, short of a real app launch.

PR #11959, opened 2026-08-01
fix(browser): scope JavaScript dialogs to their page

dgk-dev:
## Summary

- Replace Electron's app-global native `alert` / `confirm` / `prompt` sheets for managed browser guests with an overlay rendered inside the originating browser page.
- Keep Chromium's blocking dialog callback intact, including prompt text, while routing open/close state through the page's existing renderer ownership mapping.
- Preserve automation: `orca dialog accept|dismiss --page ... --worktree ...` now resolves an already-pending scoped dialog directly, with the existing agent-browser path retained as fallback.
- Clean up pending dialogs on CDP resolution, navigation cancellation, guest replacement, and teardown.

## Screenshots

The page is dimmed only inside its browser viewport; Orca's worktree and browser chrome remain usable.

Worktree-scoped browser alert

Manual QA also switched to a sibling terminal/browser surface while the alert was pending: the overlay disappeared with its originating page and returned when that page was selected again. CLI acceptance completed in 148 ms without starting an agent-browser process, and the page observed the accepted result.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [x] `pnpm build`

Focused validation on current `origin/main`:

- 189 dialog/controller/BrowserManager/IPC/renderer tests passed.
- Node and web typechecks passed.
- Changed files passed oxlint and oxfmt checks.

The full local suite was also exercised. It reached 41,919 passing tests but retained 6 failures in pre-existing PTY/shell-ready suites under a scrubbed test environment; none overlap this change. Re-running the PTY suite against the PR candidate passed all 400 tests, while one existing live-zsh non-ASCII-path case remained environment-sensitive. The dialog-focused suites are green.

## AI Review Report

Reviewed the complete main/preload/renderer path and the pinned Electron 43.1.0 implementation. The review specifically checked:

- preservation of Chromium's blocking callback and prompt value semantics;
- stale dialog IDs, page ownership, renderer hydration races, guest swaps, teardown, and CDP-driven closure;
- the existing CLI/AI accept-dismiss flow, including the startup deadlock that occurs if agent-browser is launched behind an already-blocking dialog;
- accessibility labels, keyboard acceptance/dismissal, bounded long content, localization, and inactive-page behavior.

The review flagged that the initial CLI fallback could block during agent-browser session startup. The implementation now attempts the pending BrowserManager dialog first and keeps agent-browser only as the fallback; regression tests cover both paths.

Cross-platform review covered macOS, Linux, and Windows. The overlay has no platform-specific paths, shell behavior, or modifier labels. Enter/Escape behavior is platform-neutral. The only Electron-specific dependency is the internal `-run-dialog` event used by the project's pinned Electron version; tests cover listener replacement, restoration, cancellation, and callback behavior.

## Security Audit

- No dependency, auth, secret, filesystem, path, or command-execution changes.
- Dialog IPC is limited to the trusted browser renderer and verifies page ownership before reading or responding.
- Responses include a per-dialog UUID so stale renderer actions cannot answer a newer dialog.
- Page-controlled message/default text and renderer-supplied prompt text are length-bounded; React renders text without HTML injection, and displayed origins use the existing safe origin parser.
- Teardown dismisses pending callbacks and restores Electron's original listener, avoiding a stranded blocked renderer.

## Notes

Electron does not expose a public JavaScript-dialog interception API. This uses its internal `-run-dialog` contract, which is explicitly typed by Electron and verified against the pinned runtime. If Electron changes that internal contract during a future upgrade, the controller tests should fail before release.


dgk-dev:
Updated onto current `main` and addressed all actionable review feedback:

- guarantee Chromium dialog callbacks resolve even if renderer close notification throws
- trap focus inside the modal overlay and cover busy keyboard handling
- bound text by Unicode code point instead of splitting surrogate pairs
- remove the hidden mutation from the dialog getter and document the trusted automation-only response path
- add cancellation, disposal, `file://` sanitization, untrusted IPC, page-scope, and focus regressions
- move dialog IPC coverage into a focused test file to stay within the max-lines gate
- add Spanish, Japanese, Korean, and Chinese catalog entries

Validation on current main: 7 focused test files / 198 tests passed; root typecheck passed; focused oxlint and formatting passed; localization catalog verification passed.

Jinwoo-H:
Closing this submitted branch and routing the bug to canonical owner issue #17577. We will carry the work ourselves on current main and will not request contributor follow-up. Full contributor credit is preserved for @dgk-dev. No action is requested from you. Thank you.

PR #11961, opened 2026-08-01
fix(agent-status): stop start-less SubagentStop from minting phantom working

brennanb2025:
## Summary

Fixes Linear STA-2915: after Claude `/compact` on a freshly resumed session, the worktree card/sidebar agent row could stick on **Working…** forever (empty prompt, no tool, `stateStartedAt == updatedAt`) while the TUI sat idle.

Live event capture across Claude Code 2.1.186 / 2.1.205 / 2.1.218 / 2.1.220 shows compact emits `PreCompact → SubagentStop (never started) → SessionStart(compact) → PostCompact` — and **no** `UserPromptSubmit` (the issue's original theory, and the mechanism PR #11353 targeted, no longer exists on any harness back to June 22). The real culprit is receiver-side: `buildClaudeChildDrivenStatusPayload` defaults the lead state to `working` when the pane has no cached lead record, so compact's start-less `SubagentStop` on a resumed pane in a fresh Orca process mints a `working` that no `Stop` ever clears (until the 30-minute stale window).

The fix emits child-driven status only on positive evidence: a cached lead record, a live working child in the roster, or an actual roster change from the event. Unlike #11353's `PostCompact → done` mapping (rejected for stale-completion races), this claims no completion authority, needs no correlation ID, and adds no per-pane state.

Reproduced live on the current production build (fresh pane + `claude --resume` + `/compact` → sticky `working` row minted at the exact SubagentStop timestamp) and re-run on this branch's dev build (same sequence → no row minted; real turns still track `working → done` normally).

## Screenshots

No visual change (status correctness fix). Post-fix rig screenshot of the idle sidebar after compact is attached to the Linear issue evidence.

## Testing

- [x] `pnpm lint` (oxlint + react-doctor config on changed files: clean; oxfmt clean)
- [x] `pnpm typecheck`
- [x] `pnpm test` (664 passed across src/shared + src/relay + src/main/agent-hooks suites)
- [x] Added regression tests: start-less `SubagentStop` with no lead record emits nothing (fails on parent commit); lead-present compact re-emits `done`; spawn with no lead still mints `working`; unknown stop with a surviving live child keeps `working`

## AI Review Report

Reviewed the guard's interaction with every `buildClaudeChildDrivenStatusPayload` call site: subagent lifecycle (start/stop/teammate-idle), child tool activity (upserts a working row before emitting, so unaffected), and known-child turn-boundary re-emits (now fail closed to no claim instead of minting `working` with zero evidence). TeammateIdle roster transitions still emit (rosterChanged from `idleClaudeTeammateByName`), preserving idle-row UI refreshes. Cross-platform: pure shared normalization logic consumed identically by the local hook server and the SSH relay; no shortcuts, paths, shell, or Electron platform surface touched. A same-model Codex review loop runs on this exact head via Orca orchestration; findings will be pushed before merge.

## Security Audit

No new input surface: the change only tightens which authenticated hook events may emit a status payload. Untrusted payload fields (`agent_id`) already flowed through these paths; the guard strictly reduces what an unauthenticated-content payload can cause (it can no longer mint a phantom `working` on an eventless pane). No exec, path, secret, or IPC changes.

## Notes

- Kimi is unaffected (its normalizer ignores subagent lifecycle events entirely).
- The known-stop-with-no-lead path previously emitted `working` after a restart-seeded roster drained; it now emits nothing, leaving the restored persisted status in place until real lead evidence arrives — fail-closed by design.

### DIFF
--- a/src/shared/agent-hook-listener.test.ts
+++ b/src/shared/agent-hook-listener.test.ts
@@ -3354,6 +3354,17 @@ describe('shared agent-hook-listener', () => {
       expect(stopped?.payload.interactivePrompt).toBeUndefined()
     })
 
+    it('canonicalizes padded lifecycle ids so starts and stops target one row', () => {
+      claudeEvent({
+        hook_event_name: 'SubagentStart',
+        agent_id: ' a2 ',
+        agent_type: 'general-purpose'
+      })
+      const stopped = claudeEvent({ hook_event_name: 'SubagentStop', agent_id: ' a2 ' })
+      expect(stopped?.payload.state).toBe('done')
+      expect(stopped?.payload.subagents).toBeUndefined()
+    })
+
     it('tracks whitespace-padded inventory ids canonically so their stops can drain them', () => {
       const first = claudeEvent({
         hook_event_name: 'SubagentStop',
--- a/src/shared/agent-hook-listener.test.ts
+++ b/src/shared/agent-hook-listener.test.ts
@@ -3096,6 +3096,42 @@ describe('shared agent-hook-listener', () => {
       expect(stopped?.payload.subagents).toBeUndefined()
     })
 
+    it('does not mint working from a start-less SubagentStop on a pane with no lead record', () => {
+      // Why: the live /compact flow on claude 2.1.186–2.1.220 is PreCompact → SubagentStop
+      // (never started) → SessionStart(compact) → PostCompact. On a resumed session in a
+      // fresh process this minted a sticky empty-prompt 'working' no Stop ever clears (STA-2915).
+      const stopped = claudeEvent({ hook_event_name: 'SubagentStop', agent_id: 'compact-1' })
+      expect(stopped).toBeNull()
+    })
+
+    it('re-emits the lead done state on a start-less SubagentStop after a completed turn', () => {
+      claudeEvent({ hook_event_name: 'UserPromptSubmit', prompt: 'go' })
+      claudeEvent({ hook_event_name: 'Stop', background_tasks: [] })
+      const stopped = claudeEvent({ hook_event_name: 'SubagentStop', agent_id: 'compact-1' })
+      expect(stopped?.payload.state).toBe('done')
+    })
+
+    it('still emits working for a spawn on a pane with no lead record', () => {
+      const spawned = claudeEvent({
+        hook_event_name: 'SubagentStart',
+        agent_id: 'a1',
+        agent_type: 'general-purpose'
+      })
+      expect(spawned?.payload.state).toBe('working')
+      expect(spawned?.payload.subagents).toEqual([expect.objectContaining({ id: 'a1' })])
+    })
+
+    it('keeps child-driven working when a live child survives an unknown stop with no lead record', () => {
+      claudeEvent({
+        hook_event_name: 'SubagentStart',
+        agent_id: 'a1',
+        agent_type: 'general-purpose'
+      })
+      const stopped = claudeEvent({ hook_event_name: 'SubagentStop', agent_id: 'ghost' })
+      expect(stopped?.payload.state).toBe('working')
+      expect(stopped?.payload.subagents).toEqual([expect.objectContaining({ id: 'a1' })])
+    })
+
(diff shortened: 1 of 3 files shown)


brennanb2025:
**Review gate complete — CLEAN at `864341ff2d`.**

**Live repro (production build, Claude Code 2.1.220):** fresh pane + `claude --resume` of an idle session + `/compact` → phantom row `state: working, prompt: "", stateStartedAt == updatedAt` minted at the exact `SubagentStop` timestamp, frozen while the TUI sat idle. Event capture across harness 2.1.186/2.1.205/2.1.218/2.1.220 shows compact emits `PreCompact → SubagentStop (start-less) → SessionStart(compact) → PostCompact` and **no `UserPromptSubmit`** — the issue's original theory (and #11353's target) doesn't exist on any harness back to June 22; the real culprit was the `?? 'working'` child-driven fallback.

**Branch-negative rig at this exact head (isolated dev instance, same recipe):** compact completes, **no row minted**; a control turn still tracks `working → done` normally. Three screenshots attached to the Linear issue.

**Same-model Codex review loop (xhigh): 19 fresh-reviewer rounds until CLEAN.** Each round's findings were fixed in-scope (empty-roster retention, inventory-id validation, drain/park completion authority, temporal-ordering races, id canonicalization, dead-wait card cleanup, test vacuity) or verified pre-existing on the merge base and filed: STA-3166 (unknown-`agent_id` lead classification), STA-3167 (relay roster seeding), STA-3174 (permission card outliving a dead owner). Final round independently verified: 473 related tests, full lint + max-lines, full typecheck, cross-platform relay build, non-vacuous coverage across the series.

**Perf:** guard short-circuits on cached lead (zero added cost for lead-present panes); 600 invented-pane stops = 0 payloads, 0 retained rosters, 0.47ms; lead hot path ~10.6M events/s; independent round-11 measurement 1.58µs/event nominal, 362µs with 8,000 invalid inventory rows.

**CI:** all 30 checks green at `864341ff2d`, MERGEABLE. Tests 729 passing across shared/main/relay agent-hooks suites.

brennanb2025:
Closing as superseded. Three separate reasons, and any one of them is sufficient.

**Its central idea already merged, in refined form.** #14375 (`fix(agent-status): stop start-less child stops from minting phantom working`, merged 2026-08-14 as `d137bb93e1`) landed the same insight this PR was built around: `buildClaudeCachedLeadStatusPayload` defaulted an uncached lead state to `working`, and since `claudeLeadStateByPaneKey` is in-memory only, every restart emptied it so the next child event minted a phantom `working` with an empty roster that no `Stop` would ever clear. The base case now returns `null` and `working` requires positive evidence.

**It has been silently un-tested for over two weeks.** The branch is `CONFLICTING` and has not moved since 2026-08-01, which means CI stopped running on it — so its green history says nothing about its current state against `main`.

**It is on the wrong ticket.** STA-2915 is the `/compact` sticky-`working` bug. This diff is entirely child-lifecycle and roster work across `agent-hook-listener.ts` and `claude-subagent-roster.ts`; `/compact` appears only in comments and in an `agent_id: 'compact-1'` fixture, as *context* for a start-less `SubagentStop`. Nothing here registers or handles a compact event.

**What actually fixes STA-2915:** #15202. The mechanism turned out not to be the one in this ticket's title. Measured under a PTY on Claude Code 2.1.227, `/compact` emits **no `UserPromptSubmit` at all** — so the compact window had no clearing signal whatsoever, rather than a `UserPromptSubmit` arriving without a `Stop`. #15202 registers `PostCompact` only and deliberately never `PreCompact`, which makes the abort guard structural: an aborted compact emits `PreCompact` alone, so a design that never registers it cannot be stranded by it. It also deletes the #12332 correlation machinery, which could never execute in production.

**Worth keeping from this PR.** Three small residuals here are real and are not carried by #14375 or #15202, and deserve their own follow-up rather than dying with this branch:
- an `isValidClaudeSubagentId` pre-check before roster upsert,
- agent-id trimming,
- dropping a `waiting` record that has no `stateBeforeWait`.

Not deleting the branch, so those remain easy to lift.

PR #11968, opened 2026-08-01
fix(jira): render a user picker for reporter/user create fields

### LINKED ISSUE #4643: [Bug]: Jira connector — can't create an issue from Orca; "Reporter is required", free-text name never resolves (no accountId picker)
### Operating system

windows/wsl2.

### Orca version

1.4..45

## Summary

After connecting a Jira site to Orca via API key (import succeeded, issues are readable), creating a **new** issue from inside Orca fails with **"Reporter is required."** The Reporter field accepts free-text and there is **no autofill / user picker**. Typing a display name — full ("Alex Rivera") or first name ("Alex") — does not satisfy the field, so the issue can never be created from Orca.

Root cause: the Reporter field behaves like a typeahead **picker that only accepts a selected option from its dropdown**, but the dropdown **never populates** (no autofill results). With no selectable option, the field's internal value is never set, so "Reporter is required" fires regardless of what is typed. Confirmed: typing a display name fails, **and pasting a valid Jira `accountId` also fails** — so this is not a name-vs-accountId formatting issue; the field never binds any typed text to a reporter value at all.

(Background: Jira Cloud's REST API requires `reporter` as an **`accountId`** object — `{"reporter": {"id": "<accountId>"}}` — and display names/usernames have not been accepted since Atlassian's 2019 GDPR changes. The correct UX is a picker that searches users and submits the selected accountId; Orca's picker appears to never load its search results.)

## Environment

- **Orca:** (Windows host, WSL2 runtime)
- **Jira:** Cloud, site `example-team.atlassian.net`
- **Auth:** Jira API token (read confirmed: `read:jira-work`, `write:jira-work`)
- **Projects:** team-managed (next-gen)
- **Connection method:** API key import (not OAuth)

## Steps to reproduce

1. Connect a Jira Cloud site to Orca with an API token. Import succeeds; existing issues are visible.
2. In Orca, choose "create a new item" for a Jira project (e.g. a Task).
3. Fill in summary/description; in the **Reporter** field type a name (e.g. "Alex" or "Alex Rivera").
4. Submit.

## Expected behavior

- Reporter defaults to the **token owner** (the authenticated account) automatically, since the API key already identifies a single user — and/or
- The Reporter field is a **searchable picker** that resolves names to `accountId` via the Jira user-search endpoint and submits the `accountId`.
- The issue is created.

## Actual behavior

- Submission fails with **"Reporter is required."**
- The field shows **no autofill/dropdown results**; no typed value (first name, full name) is ever accepted.
- **Pasting a valid Jira `accountId` directly into the field also fails with the same "Reporter is required" error** — confirming the field never binds typed text to a value and requires a dropdown selection that never appears.
- There is no way to proceed — issue creation from Orca is fully blocked.

## Root cause (verified against a live site)

The Reporter picker requires a dropdown selection it never offers, and the underlying data + API it needs are both available — Orca just isn't wiring them up:

- The Jira user-search endpoint works against this token: a query for "Alex" returns the matching account (`accountId 5xxxxxxxxxxxxxxxxxxxxxxx`, displayName "Alex Rivera") at HTTP 200. So the dropdown *could* be populated; it isn't.
- Pasting that same `accountId` into the field still fails — proving the field accepts only a picked option, not typed/pasted text, and the picker list is empty.
- The create API expects:
  ```json
  { "fields": { "reporter": { "id": "5xxxxxxxxxxxxxxxxxxxxxxx" }, ... } }

chucoding:
## Summary

Fixes #4643. Jira's create-issue metadata returns `schema.type: "user"` fields (most commonly Reporter) with no `allowedValues`, so Orca's create-issue form fell back to a plain free-text `Input`. Typed names and even pasted `accountId`s were never resolved into the `{ id: accountId }` shape Jira Cloud/Server require, so submitting always failed with **"Reporter is required"** — issue creation from Orca was fully blocked on any project that requires a Reporter.

- Renderer (`TaskPage.tsx`): a create field with `schema.type === 'user'` now renders a `Select` populated from a project-scoped assignable-users lookup, and the submitted value is wrapped as `{ id: accountId }`.
- Main (`src/main/jira/issues.ts`): added `listAssignableUsersForProject` alongside the existing issue-scoped `listAssignableUsers` — a create-issue draft has no issue key yet, so the Jira `user/assignable/search` lookup has to be scoped by `project` instead of `issueKey`.
- Threaded the new lookup through IPC (`src/main/ipc/jira.ts`), the runtime RPC method (`src/main/runtime/rpc/methods/jira.ts`, `orca-runtime.ts`) and preload (`src/preload/index.ts`, `api-types.ts`) so it also works over SSH/remote runtimes, not just local.
- Extracted the create-field value-building helpers (`buildJiraCreateFieldValue` etc.) out of `TaskPage.tsx` into `task-page-jira-create-field-value.ts` so they're unit-testable in isolation (matching the existing `task-page-new-issue-draft.ts` pattern in this file).

## Test plan

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` (full suite — 2 pre-existing/environment-only failures unrelated to this change, confirmed via `git stash` against a clean tree: `config/scripts/check-root-directory-entries.test.mjs` and `src/relay/agent-exec-handler.test.ts`)
- [x] `pnpm build`
- [x] New unit tests: `src/main/jira/assignable-users.test.ts` (project- vs issue-scoped search params, Cloud `query` vs Server/DC `username`), `src/renderer/src/components/task-page-jira-create-field-value.test.ts` (`{ id }` wrapping for user fields, existing select/array/number/text behavior preserved), plus coverage in `runtime-jira-client.test.ts` for the new RPC path (including the oversized-query guard used for SSH/remote calls).

## Code review notes (per CONTRIBUTING.md)

- **Cross-platform**: renderer/main changes are platform-agnostic; no path or shortcut handling involved.
- **SSH/remote**: the new `listAssignableUsersForProject` lookup is threaded through the same RPC/IPC dual path (`jiraListAssignableUsersForProject` in `runtime-jira-client.ts`) as the existing issue-scoped lookup, so it works identically for local, SSH, and remote-runtime task sources. Query length is guarded by the existing `isRuntimeProviderSearchQueryWithinLimit` before any RPC call.
- **Agent/integration compatibility**: Jira-only change; no other provider touched.
- **Git-provider compatibility**: n/a (not a source-control feature).
- **Performance**: the new project-scoped user lookup is fetched once per create-dialog open (only when a visible field actually needs it), capped at `maxResults=50`, same as the existing issue-scoped lookup.
- **UI quality**: reuses the existing `Select`/`SelectTrigger`/`SelectItem` pattern already used for `allowedValues`-backed fields in the same dialog, so it matches existing styling; a "Loading users…" placeholder is shown while the fetch is in flight.
- **Security**: no new attack surface — reuses the existing authenticated Jira client and the same `user/assignable/search` endpoint already used for the assignee picker on existing issues.


hyeonseok-suh:
@greptile-apps Fixing. Will be in the next commit

OrcaWin:
Superseded by the merged Jira user-field implementation in #17456 (with remote-host compatibility completed in #17532). Closing this duplicate while preserving your original contribution and discussion. Thanks @chucoding.

PR #11976, opened 2026-08-01
fix(runtime): expose OpenCode permission interaction

### LINKED ISSUE #11971: fix(runtime): expose OpenCode PermissionRequest as a structured interaction
## Reproduction

1. Create an Orca-managed OpenCode session.
2. Trigger OpenCode `permission.asked`.
3. Read the paired Runtime status through `worktree.ps` or `session.tabs.listAll`.

## Observed

The local Orca hook cache records `hookEventName: "PermissionRequest"`, while the public agent status exposes only:

```json
{ "state": "waiting" }
```

The event does not carry safe tool or command metadata. A Runtime-only client cannot distinguish a permission from an ordinary question without parsing terminal output.

## Expected

Project the attention kind as an explicit structured interaction:

```json
{
  "state": "waiting",
  "interaction": { "kind": "permission" }
}
```

The projection must not expose the OpenCode request id or session id.

## Scope

- Preserve existing `waiting` semantics for Orca desktop UI.
- Add a typed public interaction field for paired Runtime clients.
- Do not parse terminal output.
- Do not fabricate tool or command details.

## Related

- #11761: hook interaction delivery to paired/headless clients.
- #10099: broader structured-agent adapter contract.

anthonypelletier:
## Summary

- Exposes OpenCode `PermissionRequest` as the typed Runtime interaction `{ kind: "permission" }`.
- Preserves `waiting` while omitting the opaque permission-request ID from the interaction payload.
- Retains the canonical OpenCode provider-session ID in the existing `providerSession: { key: "session_id", id }` field for provider-session routing and resume behavior.
- Carries the interaction through IPC, synced sessions, `session.tabs`, and `worktree.ps`.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` (not run; full repository lint is outside this focused change)
- [x] `pnpm typecheck`
- [ ] `pnpm test` (not run; focused suites below)
- [ ] `pnpm build` (not run; no build-affecting change)
- [x] Added regression coverage for hook normalization, trust-boundary reconstruction, deduplication, desktop/SSH/mobile projection, live permission retention, and provider-session identity.

Executed on the final head:

- `corepack pnpm exec vitest run --config config/vitest.config.ts` on the nine changed focused test files (1,330 passed, 1 skipped)
- `corepack pnpm run typecheck`
- `corepack pnpm exec oxfmt --check` on all 22 modified files

## AI Review Report

Reviewed the hook-to-Runtime data path for stale interaction state, status deduplication, session synchronization, `session.tabs`, and `worktree.ps` projection. Verified that the new permission interaction is cleared by subsequent events, remains live under a non-agent pane title, and never contains the opaque permission-request ID. The canonical provider-session ID remains intentionally available only through the existing provider-session field.

Cross-platform review: this change adds no shortcuts, labels, paths, shell commands, or Electron-specific behavior. The structured in-memory payload is platform-neutral for macOS, Linux, and Windows, including SSH relay reconstruction.

## Security Audit

The hook payload is treated as untrusted. The interaction normalizer accepts only the allowlisted `permission` kind and reconstructs the canonical object, discarding unknown fields. The opaque permission-request ID, tool details, and commands are not copied to Runtime clients. The event's canonical provider-session ID continues through the existing `providerSession: { key: "session_id", id }` channel for established provider-session routing and resume semantics; it is intentionally separate from the interaction payload.

No auth, secrets, dependency, command-execution, path-handling, workflow, executable-mode, or download changes are introduced.

## Notes

Questions continue to use the existing `interactivePrompt` payload. This PR adds the missing typed signal only for OpenCode permission requests.

Fixes #11971

### DIFF
--- a/mobile/src/worktree/worktree-list-snapshot.test.ts
+++ b/mobile/src/worktree/worktree-list-snapshot.test.ts
@@ -160,6 +160,13 @@ describe('areWorktreeListsEqual', () => {
     expect(areWorktreeListsEqual(first, second)).toBe(false)
   })
 
+  it('detects permission interaction changes', () => {
+    const first = [worktree({ agents: [agent()] })]
+    const second = [worktree({ agents: [agent({ interaction: { kind: 'permission' } })] })]
+
+    expect(areWorktreeListsEqual(first, second)).toBe(false)
+  })
+
   it('treats missing and empty agent arrays as equivalent for rendering', () => {
     const first = [worktree({ agents: undefined })]
     const second = [worktree({ agents: [] })]
--- a/mobile/src/worktree/worktree-list-snapshot.ts
+++ b/mobile/src/worktree/worktree-list-snapshot.ts
@@ -107,6 +107,7 @@ function areAgentRowsEqual(
       a.lastAssistantMessage !== b.lastAssistantMessage ||
       a.toolName !== b.toolName ||
       a.toolInput !== b.toolInput ||
+      a.interaction?.kind !== b.interaction?.kind ||
       a.interrupted !== b.interrupted ||
       a.stateStartedAt !== b.stateStartedAt ||
       a.updatedAt !== b.updatedAt
--- a/src/main/agent-hooks/hook-status-session-tabs-invalidation.test.ts
+++ b/src/main/agent-hooks/hook-status-session-tabs-invalidation.test.ts
@@ -24,9 +24,10 @@ describe('createHookStatusSessionTabsInvalidator', () => {
 
   it('stays quiet while the same status keeps being pinged', () => {
     const changed = createHookStatusSessionTabsInvalidator()
-    changed(working())
+    const permission = working({}, { interaction: { kind: 'permission' } })
+    changed(permission)
 
-    expect(changed(working())).toBe(false)
+    expect(changed(permission)).toBe(false)
   })
 
   it.each([
@@ -35,6 +36,7 @@ describe('createHookStatusSessionTabsInvalidator', () => {
     ['agentType', { agentType: 'codex' }],
     ['toolName', { toolName: 'Bash' }],
     ['interactivePrompt', { interactivePrompt: '{"questions":[]}' }],
+    ['interaction', { interaction: { kind: 'permission' as const } }],
     ['interrupted', { interrupted: true }]
   ])('invalidates when %s changes', (_field, payload) => {
     const changed = createHookStatusSessionTabsInvalidator()
--- a/src/main/agent-hooks/hook-status-session-tabs-invalidation.ts
+++ b/src/main/agent-hooks/hook-status-session-tabs-invalidation.ts
@@ -28,6 +28,7 @@ export function createHookStatusSessionTabsInvalidator(): {
       (previous.agentType ?? null) !== (next.agentType ?? null) ||
       (previous.toolName ?? null) !== (next.toolName ?? null) ||
       (previous.interactivePrompt ?? null) !== (next.interactivePrompt ?? null) ||
+      previous.interaction?.kind !== next.interaction?.kind ||
       (previous.interrupted ?? false) !== (next.interrupted ?? false)
     )
   }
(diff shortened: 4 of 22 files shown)


Jinwoo-H:
Thanks for the contribution. We tested an OpenCode permission prompt on both current main and this PR’s reviewed head. In both builds, Orca correctly showed the workspace as waiting/action-required while the permission prompt was unanswered.

Main already maps OpenCode `PermissionRequest` events to `state: waiting`. This PR additionally introduces a structured `interaction: { kind: "permission" }` field and preserves it through several runtime projections, but we have not reproduced a user-visible failure that requires that additional metadata. The base failure we found required artificially changing the terminal title to `bash`.

Could you provide a minimal reproduction where current main fails, including:

- OpenCode and Orca versions
- OS and whether the session is local, SSH, or WSL
- OpenCode permission configuration
- Which Orca surface fails to show the waiting status
- Whether the terminal title changes while the prompt is open

A short recording or relevant hook payload would also help. We are closing this for now because the reported behavior already works on main, but we would be happy to reconsider or reopen it with a concrete reproduction showing the remaining gap.

anthonypelletier:
Thanks for testing this. I think we tested two different things, though.

Your test confirmed that the **desktop UI** shows `waiting/action-required` for a permission prompt — that's correct, and it's not what this PR is about. Main already handles that fine via `state: waiting`.

The gap in #11971 is at the **Runtime API layer** (`worktree.ps`, `session.tabs`) for paired/headless clients — mobile, third-party integrations — that only ever see the serialized `RuntimeWorktreeAgentRow`, not the desktop terminal.

**Repro showing the gap still exists on current `main`:**

1. Trigger an ordinary OpenCode `AskUserQuestion` prompt. Call `worktree.ps` for that worktree while it's pending. Note the row: `state: "waiting"`, `toolName: null`, `toolInput: null`.
2. Trigger a real OpenCode `PermissionRequest` (e.g. a `bash` call that needs approval). Call `worktree.ps` again while it's pending. Note the row: `state: "waiting"`, `toolName: null`, `toolInput: null` — identical to step 1.

Why: `extractOpenCodeToolFields` in `src/shared/agent-hook-listener.ts` only special-cases `MessagePart` and `AskUserQuestion`. There's no `PermissionRequest` branch for OpenCode at all, so `toolName`/`toolInput`/`interactivePrompt` never get populated for it. `normalizeOpenCodeFamilyEvent` maps both `PermissionRequest` and `AskUserQuestion` to the same `state: 'waiting'`. A Runtime client has zero fields to tell "pending question" from "pending approval" apart.

**This isn't OpenCode-specific design, and fixing it for other harnesses is cheap, not a rewrite.** I went through every provider branch in `agent-hook-listener.ts`. Most already compute the question-vs-permission distinction internally as two separate conditions before collapsing them into the same `state`:

- Claude/Kimi/Devin: `eventName === 'PermissionRequest' || isAskUserQuestion ? 'waiting' : ...` — `isAskUserQuestion` is already its own boolean.
- Codex: same shape, `isUserInputPreTool` vs `PermissionRequest`.
- Grok: literally two separate `else if` branches (`isUserInputPreTool` vs `isGrokPermissionNotification(...)`).
- Copilot: two separate booleans (`isAskUserPrompt` vs `isBlockingNotification`) merged into one `'blocked'`.
- Antigravity: `toolName === 'ask_question' || toolName === 'ask_permission'` merged into one check — `toolName` is already available.
- Droid: `isDroidAskUserTool(...) || isDroidHighRiskToolUse(...)` merged the same way.

For all six, the fix is a one-line addition per branch — set `interaction: { kind: 'permission' }` only on the real-permission side — reusing the exact plumbing (`interaction` field, IPC, `RuntimeWorktreeAgentRow`) this PR already built for OpenCode. No redesign needed.

The one honest exception: **Gemini, Amp, and Command-code don't model a permission-wait state at all** — they never reach `waiting`/`blocked` for either a question or a permission, so there's no branch to extend. That's a separate, bigger gap, out of scope here.

For completeness: the `'permission'` value on the desktop `WorktreeStatus` type doesn't cover this either — `applyLiveAgentState` in `worktree-agent-activity-summary.ts` assigns it to *any* `state === 'blocked' || 'waiting'`, so it doesn't distinguish a question from a permission, and it's a renderer-only UI label, not part of the `worktree.ps`/Runtime RPC contract paired clients actually consume.

Happy to reopen with this repro, extend the PR to cover the other harnesses in the same pass, or rebase onto current `main` — let me know what you'd prefer.

PR #11983, opened 2026-08-01
fix(ime): scope deferred newline to captured sessions

yeongjunyoo:
## Summary

#11896 re-arms the deferred-newline deadline while a composition is pending, but its global pending count also includes sessions that start after the newline was deferred. If a new composition overlaps the original session's end, that later session can hold the older newline and reverse the intended `syllable -> newline -> next syllable` ordering.

This follow-up:

- snapshots the xterm composition session IDs that are open when the newline is deferred
- waits only for those captured sessions, excluding sessions started afterward
- keeps per-session reference counts so overlapping route installations on the same terminal element cannot clear each other's pending state
- preserves the existing `compositionupdate` fallback for routes installed mid-composition, where no session start was captured

Related: #11878. Stacked on #11896.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint` — full repository lint not run; both oxlint configurations passed on all four changed files
- [ ] `pnpm typecheck` — full three-project check not run; `tsc --noEmit -p config/tsconfig.tc.web.json` passed
- [ ] `pnpm test` — full suite not run; 59 focused IME tests passed across `keyboard-handlers-ime`, `terminal-ime-composition-route`, and `terminal-ime-deferred-newline`
- [ ] `pnpm build` — not run
- [x] Added regression coverage for post-defer sessions and overlapping route ownership

Additional gates passed: `oxfmt --check`, `check:max-lines-ratchet`, and `git diff --check`.

## AI Review Report

Reviewed the change against the IME transaction lifecycle, delayed completion, route replacement, and local/SSH transport paths. The first implementation stored pending IDs in a `Set`; review found that two overlapping routes can own the same session ID, so disposing one route would incorrectly clear the other's pending state. The implementation now uses per-session reference counts, with a regression test for overlapping route ownership.

Cross-platform review: the change is platform-neutral DOM/xterm composition bookkeeping. It does not alter shortcuts, labels, file paths, shell commands, or Electron platform branches on macOS, Linux, or Windows. Local, SSH, and relay panes continue to use the same terminal-element-scoped state and transport abstraction.

## Security Audit

The change only narrows when an already-resolved terminal input write may fire. It does not add or transform input bytes, execute commands, touch paths, add IPC, change authentication, read secrets, or add dependencies. Snapshot state remains in a terminal-element-keyed `WeakMap`; disposed routes release their reference counts.

## Notes

This PR intentionally targets the head branch of #11896 rather than `main`, so it can be reviewed and folded into that existing non-duplicate IME fix.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-ime-composition-route.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-composition-route.test.ts
@@ -149,6 +149,35 @@ describe('installTerminalImeCompositionRoute', () => {
     expect(harness.input).not.toHaveBeenCalled()
   })
 
+  it('leaves an uncaptured session to xterm when installed mid-composition', () => {
+    const harness = createHarness()
+
+    // No start was seen, so nothing here can deliver the text. Cancelling the event
+    // would suppress xterm's own insertion and drop the commit on the floor.
+    const notPrevented = harness.end(1, '한')
+
+    expect(notPrevented).toBe(true)
+    expect(harness.input).not.toHaveBeenCalled()
+  })
+
+  it('suppresses xterm insertion only for a session it captured', () => {
+    const harness = createHarness()
+    harness.start(1)
+
+    expect(harness.end(1, '한')).toBe(false)
+    expect(harness.input).toHaveBeenCalledWith('한')
+  })
+
+  it('keeps suppressing insertion for a captured session it deliberately drops', () => {
+    const harness = createHarness()
+    harness.start(1)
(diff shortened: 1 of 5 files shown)


nwparker:
Thanks for this, @yeongjunyoo — the captured-session IME route was the right design.

This already landed on `main` in #15056, which ported your change onto current main with `Co-authored-by` credit. I'm closing this PR as no longer needed.

This note is from an automated scan of open PRs whose issue already shipped in a later merge. If that's wrong, please reopen and say so.

PR #11988, opened 2026-08-01
fix(mobile): stop Codex background terminals from native chat

brennanb2025:
Stop wrote two Escapes and nothing else. That interrupts the agent's turn, but the background terminals the turn spawned keep running: interrupting a turn and reaping background terminals are two distinct operations in the agent, and only the second one stops them. A real `sleep 180` survived Stop and kept running while the chat showed the agent as idle.

After the interrupt is confirmed, Stop now dispatches Codex's own `/stop` through the existing send seam. That means the cleanup types key by key rather than as one burst — a pasted burst is read as prose, not a command — clears the input line first so it cannot submit a draft the user left in the composer, and takes the per-terminal write lock instead of opening a second unserialized write path into the same PTY. It is not a new mechanism: `dispatchCommand` was already the exact-`codex`-gated, lock-acquiring, stale-input-healing slash-command dispatcher on this surface, and `/stop` is already in the shipped command catalog for this agent.

Gated on an exact `codex` match, never "not claude": typing a literal `/stop` into an agent without that command would post it as a prompt.

**The Escapes themselves stay ungated by the write lock, so Stop still always fires.** Only the cleanup leg can be refused, and it says so rather than failing silently — "Agent stopped; background terminals may still be running" on a rejection, "background cleanup unconfirmed" on a lost ack, neither of which reads as "Stop failed" and invites a second Escape into changed composer state.

**Known and pre-existing:** because the Escapes are deliberately not lock-gated, an Escape can still land in the middle of an in-flight paste sequence. That is unchanged from main and is the direct, intended cost of keeping the interrupt unconditional — gating the emergency stop behind a lock a paste happens to hold would turn it into a silent no-op, which is the failure this PR exists to remove.

## Scope

This is the PTY native-chat surface. Structured Codex sessions are a separate lane with their own Stop; that gap is fixed by #19259, which adds the Codex adapter's `backgroundTaskState` / `stopBackgroundTasks`. The two cannot both fire for one session — the hook here is constructed with `enabled: inputSendable && !activeChatStructured`, and the UI calls `activeChatStructured ? structuredNativeChat.cancel : handleNativeChatStop`. Longer term the right shape is migrating these surfaces to structured rather than typing commands into a composer; #19259 carries that direction and this PR does not claim to.

## This branch supersedes its own earlier architecture

The FIFO stop lease, writer gate, controller-writer plumbing and module-level cleanup-retry map are dropped in favour of `mobile-native-chat-terminal-write-lock.ts`, which landed in #12502 and already serializes writer-vs-writer per terminal. Two of the three defects this PR previously fixed existed only because it typed into a shared composer through its own new write path; the third, cross-agent stale replay, is already covered by main's `activeRouteRef` re-check at `use-mobile-native-chat-stop.ts:79-87`.

Net diff versus `main` is **5 files, +354/−27**, all under `mobile/src/session/` — the old 1,857-line architecture nets out to zero.

## Testing

Mobile suite 517 files / 4,215 passed, 3 skipped. Red-then-green by ablation, both applied after committing and restored by file path with `HEAD` re-verified afterwards rather than only the working tree: removing the cleanup trigger fails 3 tests; breaking the dispatcher binding fails 2, including the exact typed byte sequence `['\x15','/','s','t','o','p','\r']`. Green after restore: 22/22.

`mobile/ tsc --noEmit`, oxlint and oxfmt clean. Both lockfiles byte-identical to `main`; neither commit contains one.

Note for review: root `pnpm tc` is not a meaningful gate for this branch — it is mobile-only, so `mobile/ tsc --noEmit` is the real one.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -856,6 +856,7 @@ jobs:
           src/main/agent-hooks/windows-hook-payload-delivery.test.ts
           src/main/agent-hooks/windows-direct-cmd-hook-command.test.ts
           src/main/windows/windows-pty-job.win32.test.ts
+          src/main/windows/windows-msys-job.win32.test.ts
           src/main/windows/windows-host-job.win32.test.ts
           src/main/windows/windows-process-tree-command-line-patch.test.ts
           src/main/windows/windows-process-table-native-addon.win32.test.ts
--- a/config/patches/node-pty@1.1.0.patch
+++ b/config/patches/node-pty@1.1.0.patch
@@ -603,7 +603,7 @@ index 7b4b9e1f990fbf95b51528bb56dc9717f5b87532..2ae787c5bd4f3eba470584dc658a01a5
  }
  #endif
 diff --git a/src/win/conpty.cc b/src/win/conpty.cc
-index 7b286d3d644c26141df516929703aa6e129df4b2..4aed260dd68e6a171dcfd349e9a7c5c97209248e 100644
+index 7b286d3d644c26141df516929703aa6e129df4b2..4b06d18576c807c3d1181a7bd714140c6678cf86 100644
 --- a/src/win/conpty.cc
 +++ b/src/win/conpty.cc
 @@ -18,6 +18,7 @@
@@ -614,7 +614,7 @@ index 7b286d3d644c26141df516929703aa6e129df4b2..4aed260dd68e6a171dcfd349e9a7c5c9
  #include <vector>
  #include <Windows.h>
  #include <strsafe.h>
-@@ -44,12 +45,39 @@ struct pty_baton {
+@@ -44,12 +45,40 @@ struct pty_baton {
    HANDLE hOut;
    HPCON hpc;
  
@@ -630,6 +630,7 @@ index 7b286d3d644c26141df516929703aa6e129df4b2..4aed260dd68e6a171dcfd349e9a7c5c9
 +  // refused to create or assign one (an outer job without breakaway rights),
 +  // in which case callers fall back to their pre-job behaviour.
 +  HANDLE hJob = nullptr;
++  bool allowJobBreakaway = true;
 +
 +  // Orca: teardown needs BOTH the shell's death and an explicit kill() before
 +  // the baton can be freed, so each side records that it has run. Whichever
@@ -655,7 +656,7 @@ index 7b286d3d644c26141df516929703aa6e129df4b2..4aed260dd68e6a171dcfd349e9a7c5c9
  static volatile LONG ptyCounter;
  
  static pty_baton* get_pty_baton(int id) {
-@@ -102,8 +130,31 @@ void SetupExitCallback(Napi::Env env, Napi::Function cb, pty_baton* baton) {
+@@ -102,8 +131,31 @@ void SetupExitCallback(Napi::Env env, Napi::Function cb, pty_baton* baton) {
      // Get process exit code.
      GetExitCodeProcess(baton->hShell, (LPDWORD)(&exit_event->exit_code));
      // Clean up handles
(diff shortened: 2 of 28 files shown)


brennanb2025:
Updated to `e72756b195` — brought onto current `main` and rescoped.

**What changed since the last review.** The 1,857-line lease / writer-gate / persisted-retry architecture is gone. #12502 landed an acquire-or-reject terminal write lock a month ago (`mobile-native-chat-terminal-write-lock.ts`), so the writer-serialization half of this PR was superseded, and resolving its conflicts would have meant re-litigating that design. Two of the three defects this PR previously fixed existed *only* because it typed into a shared composer through its own new write path; the third (cross-agent stale replay) is already covered by main's `activeRouteRef` re-check at `use-mobile-native-chat-stop.ts:79-87`.

**What actually ships now.** After both Escapes settle with at least one accepted, and only for an exact `codex` match on a still-live route, Stop dispatches `/stop` through main's **existing** `dispatchCommand` (`use-mobile-native-chat-message-send.ts:277-318`) — which already acquires the write lock, heals stale input, gates on exact `'codex'`, and types key-by-key via `typeAgentTuiCommand`. That last part matters: a bundled `{ text: '/stop', enter: true }` burst is classified by the TUI as prose rather than a command, which `:144` explicitly warns about.

This addresses the original defect. Escape interrupts a Codex *turn* but does not reap its background terminals — the agent runtime models interrupt and background-terminal cleanup as two distinct operations, and `/stop` ("Stop all background terminals") is already in Orca's own shipped catalog at `native-chat-slash-commands.ts:79`. The August native QA failure, where a real `sleep 180` survived a second Stop past 15 seconds while the UI still showed Working, is exactly this gap.

**Deliberate decision worth reviewing.** The write lock covers the `/stop` write only, never the Escapes. Gating the interrupt behind a lock a mid-flight paste holds would stop Stop from firing at all — a worse failure than the one being fixed. Consequence, stated plainly: **Escape-into-a-mid-flight-paste remains possible, unchanged from main.** That is pre-existing and out of scope here. If the `/stop` write is rejected because the lock is busy, the user is told background terminals may still be running rather than being left to assume success.

**Evidence.** Red-then-green with two ablations, both applied after committing and restored by file path, with `HEAD` re-verified after each restore rather than only the working tree:
- removing the cleanup trigger → 3 tests fail
- breaking the dispatcher binding → 2 tests fail, `expected [] to deeply equal ['\x15','/','s','t','o','p','\r']`
- green after restore: 22/22

Full mobile suite 517 files / 4,215 passed, 3 skipped. `tsc --noEmit` clean (this branch is mobile-only, so the root typecheck says nothing about it — `mobile/` is its real gate). oxlint and oxfmt clean. Both lockfiles byte-identical to `main`; neither commit contains one.

**Companion PR: #19259**, which implements `backgroundTaskState` / `stopBackgroundTasks` on the Codex structured adapter. The two do not overlap and cannot double-fire: `useMobileNativeChatStop` is constructed with `enabled: inputSendable && !activeChatStructured`, and the UI calls `activeChatStructured ? structuredNativeChat.cancel : handleNativeChatStop`, so a session is on exactly one lane. #19259 alone would leave this bug live for every PTY-lane user; this PR alone would leave structured sessions unfixed.

**Still not covered, and honest about it:** PTY-scrape native chat is the wrong long-term shape for this — desktop has the identical hole at `NativeChatResolvedView.tsx:117`, and the right fix for both is migrating those surfaces to structured rather than typing into a composer. This PR does not resolve that; it fixes the reported bug on the surface users are on today.

brennanb2025:
Closing this without merging.

**The surface it fixes is being retired.** Terminal-based native chat for Codex is being restructured into structured native chat in the next few days, and this fix is gated on an exact `codex` match — deliberately, so a literal `/stop` could never be typed into another agent's TUI. That gating means its entire value sits on the surface going away: it does not help Claude, Cursor, Gemini or any other agent on the PTY path, and it does not touch the structured path at all.

**It also has no tracking issue.** No linked Linear issue and no STA reference — it came out of an automation run rather than a user report.

**The defect is real and stays fixed.** Interrupting a Codex turn does not reap the background terminals that turn spawned: interrupt and background-terminal cleanup are two distinct operations in the agent runtime, and Escape drives only the first. A real `sleep 180` survived Stop while the chat showed the agent as idle.

**#19259 handles it properly, on the surface that survives.** It implements `backgroundTaskState` and `stopBackgroundTasks` on the Codex structured session adapter behind a capability probe — the methods Claude's adapter already implements and Codex's did not — which lights up the desktop background-tasks UI that has been shipped for a while, and adds the equivalent control on mobile. That is Orca's own answer to "stop the agent's background work": a typed adapter capability surfaced as a distinct control, rather than synthesising keystrokes into a shared composer.

For the record, the work here was not wasted: it established the mechanism (Stop writes two Escape bytes into the PTY and nothing else), proved that no process-group signalling path is reachable from Stop, and identified the adapter gap that #19259 fills.

PR #11999, opened 2026-08-01
fix(ssh, relay): keep remote sessions alive through reconnects and backpressure

AmethystLiang:
## Summary

- Separate SSH reconnect delay progression from the failed-handshake budget so repeated post-handshake drops back off correctly without prematurely giving up.
- Keep relay overload nonfatal: size watcher batches per client, bound overflow resync markers, preserve terminal completion delivery, and return a JSON-RPC error for oversized responses instead of disconnecting the host.
- Preserve agent-hook state through bounded shedding and retry, with turn and digest validation before restoring omitted fields.

## User impact

Remote workspaces and PTYs remain available through watcher floods, large filesystem responses, transient producer congestion, and repeated network flaps. A slow client no longer forces healthy peers to use its smaller frame capacity or resynchronize their file trees.

Agent status remains coherent when payloads must be reduced: active subagent state is restored safely, interactive prompts are compacted for waiting states, and stale shed data is not resurrected.

## Implementation details

- Introduces an SSH reconnect ladder with independent delay and failure counters, capped to remain within the relay grace window.
- Publishes watcher chunks and agent-hook envelopes per client, using exact encoded-byte budgets and per-client retention checks.
- Sends one outstanding watcher overflow marker per client and root on a nonfatal control-lane path.
- Holds filesystem stream capacity until terminal frames settle and substitutes `ResponseOverCapacity` for responses that cannot fit.
- Restores shed agent-hook fields only when their digest and turn identity match the current payload.

## Screenshots

No visual change.

## Testing

- [x] Focused Vitest coverage for SSH reconnects, relay watcher chunking, capacity degradation, oversized notifications, and watcher lifecycle: 173 tests passed
- [x] Commit hook completed `oxlint` and `oxfmt`
- [ ] Full `pnpm test`
- [ ] `pnpm build`

## AI Review Report

The review examined reconnect accounting, relay frame admission, producer/control-lane failure behavior, overflow-marker lifecycle, per-client isolation, terminal-frame settlement, agent-hook retry and shedding, stale-generation handling, and response-size failure semantics. Follow-up fixes added exact byte sizing, bounded marker suppression, per-client low-water checks, turn/digest validation, and explicit regression coverage.

Cross-platform compatibility was checked for macOS, Linux, and Windows. Runtime path grouping uses the existing cross-platform path resolver; no keyboard shortcuts, labels, UI behavior, or platform-specific shell construction were introduced. SSH-hosted and folder-workspace behavior remain in scope.

## Security Audit

Reviewed remote event paths and JSON-RPC payloads as untrusted inputs, ensuring they are encoded within bounded frame budgets and cannot close the connection through an oversized producer frame. Oversized request responses now fail only the request. Shed agent state is restored only after matching the current turn and digest, reducing stale-state injection risk.

The change does not add dependencies, credentials, secret storage, or new command interpolation. No unresolved security follow-up was identified.

## Notes

The main residual behavior is intentional fail-soft degradation: when a client cannot accept a watcher batch, it receives a bounded overflow resync marker rather than forcing the entire SSH connection to restart.


AmethystLiang:
@coderabbitai[bot] resolved

AmethystLiang:
@coderabbitai[bot] resolved

Issue #3191, opened 2026-05-29
[Bug]: Orca process (and spawned Codex/agent children) keeps running after window close, blocking clean relaunch


mubeda:
### Operating system

Windows 11

### Orca version

1.4.33.0

### Summary
On Windows, closing the Orca window does not fully terminate the app. The main
process and/or spawned child processes (the codex CLI, Node helpers, terminal/PTY
processes) keep running in the background. Relaunching while a stale instance is
alive leads to overlapping state writes (see the primary issue) and, in some cases,
a new window that doesn't behave as a clean start.

### Steps to reproduce

Run a Codex agent / start a task in Orca on Windows.
Close the Orca window with the X.
Open Task Manager (or Get-CimInstance Win32_Process -Filter "Name LIKE '%orca%' OR Name='codex.exe'").
### 
Actual behavior
One or more Orca / codex / Node / PTY processes remain running after the window
is closed. They are not cleaned up and must be killed manually (taskkill /T) to get
a fully clean relaunch.
Expected behavior

Closing/quitting Orca should terminate the full process tree it owns (main +
spawned agent/terminal children), or
If background runs are intentionally kept alive, provide a clear "Quit Orca" action
and an indicator that background agent processes are still running, plus a reliable
way to stop them all.

### Notes
This is consistent with a documented Codex-on-Windows lifecycle problem: the UI window
closes but background processes remain, blocking relaunch, recoverable only by killing
the leftover processes in Task Manager.

Because Orca manages the Codex runtime (codex-runtime-home), this lifecycle bug is
likely the trigger for the duplicate-[hooks.state] corruption in the primary issue. 
Issue #3190

nwparker:
I checked the shutdown path while fixing #3190. Current code intentionally does **not** kill the terminal daemon and its sessions during normal app quit: `will-quit` calls `disconnectDaemon()`, whose comment says the daemon survives app quit so sessions can warm-reattach on next launch. It does still call `killAllPty()` for the legacy/local provider, and Windows `window-all-closed` should call `app.quit()` after the last window closes.

So there are two different cases to separate:

- Expected/current behavior: `codex.exe`, node-pty children, or daemon-owned terminal sessions remain after closing the window because Orca detached from the daemon instead of terminating sessions. This needs a product change such as a visible background-session indicator plus a “Quit and stop all sessions” command.
[...]

PR #12021, opened 2026-08-01
fix(ime): derive non-composition IME text from the caret range

nwparker:
## What

`CompositionHelper._handleAnyTextareaChanges` derived the text an IME committed
outside a composition by diffing two observations of the helper textarea:

```js
const oldValue = this._textarea.value;
setTimeout(() => {
  const newValue = this._textarea.value;
  const diff = newValue.replace(oldValue, '');
```

`String.replace` with a string pattern removes the **first occurrence anywhere**,
not a prefix. Any insertion that was not a pure append produced garbage.

This records the caret alongside the buffer and reads the insertion as the slice
between two known offsets — the same range-based extraction `_finalizeComposition`
already uses, and what the IME rule in `AGENTS.md` requires.

## Why it matters

This is not a latent path. `xterm-bypass-policy.ts` deliberately routes a bare
keydown-229 into it on macOS and Linux:

```ts
const passesStandalone229Keydown = isMac || isLinux
```

That is the non-composition glyph path for an active IME — punctuation and digits,
the first key after a macOS input-source switch, Sogou/fcitx candidate commits on
Linux. A buffer of `ab` with the caret between the two characters gaining an `x`
sent the whole string `axb` to the pty instead of `x`.

## Evidence

The tests were run against the prior patch to confirm they bite rather than merely
pass. Two fail without the fix:

```
AssertionError: expected [ 'axb' ] to deeply equal [ 'x' ]
AssertionError: expected [ 'axbz' ] to deeply equal []
```

They drive a real `Terminal` through the patched bundle, not a reimplementation of
the logic. The other four cases pass both before and after — they are regression
guards for the behavior deliberately left alone.

## Scope

- **Deletion and same-length replacement keep their existing behavior.** Neither is
  the defect proven here, and changing them would be speculative.
- **Windows keeps full 229 suppression.** Its race is preedit reaching the textarea
  *before* `compositionstart`, which still reads as an ordinary insertion, so the
  range fix does not close it. The platform split stays.
- `_dataAlreadySent` now always equals the text actually sent, which is the
  invariant its consumer assumes when it advances the composition start to avoid
  re-sending (xterm Issue #3191). Previously it was set to the mangled diff.

## Note on the patch

This is the first change to land through the regeneration workflow added earlier in
the stack: source hunks edited in the generator's checkout, bundles rebuilt from the
pinned upstream commit, lockfile hash updated, `--check` green. Before that workflow
existed this fix would have meant hand-editing four minified files.

The fix is narrow and upstreamable — it is a defect in stock xterm, inherited
verbatim by our fork, not something specific to Orca.

## Verification

- `pnpm run typecheck` clean
- 587 test files, 6398 tests passing
- `oxlint` and `audit:code-quality:type-aware` exit 0
- `regenerate-xterm-patches.mjs --check` in sync

## Reading the diff

`config/patches/@xterm__xterm@*.patch` is a build artifact regenerated from the
source patch, not something to read. #11894 marks it `-diff -text` and cuts it to
1.5 MB, so this branch moves it by 2,852 bytes rather than the megabytes it used
to. The file to actually review is the 52-line change to
`xterm-src/@xterm__xterm@6.1.0-beta.287.src.patch`.

## Two more defects in the same file, found by a cross-reference review pass

Reviewing this branch against how mature terminal implementations derive committed
text turned up two further defects in `CompositionHelper`. Both are now fixed here,
each in its own commit with its own regeneration, so either can be reverted alone.

### 1. Input typed after a shortened commit reached the pty twice

`_getPendingTextareaInput` used the recorded composition extent as a *floor*:

```ts
: Math.max(pending.position.end, start + compositionLength);
```

`position.end` is sampled from the caret during `compositionupdate`, so it records
the **preedit** extent. Every conversion engine makes that overshoot the commit —
`nihao` is 5 characters and commits `你好`, which is 2. Nothing lowers it again;
the only other write is in `blur()`.

The consequence is in the variant whose entire job is to *exclude* input that
followed the composition. With a stale floor it excluded nothing, so the dedup
check compared the wrong string, and the character typed immediately after a
commit was sent by both the composition path and the input path:

| typed | expected | actual |
|---|---|---|
| `nihao` → `你好` then `.` | `你好.` | `你好..` |
| `hiragana` → `ひらがな` then `a` | `ひらがなa` | `ひらがなaa` |
| `gks` → `한` then `1` | `한1` | `한11` |

Once the commit length is known it *is* the extent, not a floor.

This one is **pre-existing** — it arrived with the patch import in #11894 rather
than with this branch — and it is why the existing suite missed it: its helper
writes the commit into the textarea as the preedit, so preedit and commit lengths
were always equal and the stale floor never had anything to overshoot. The three
rows above are now table-driven cases with a genuine conversion step.

### 2. A replacement passed as an insertion, emitting only its tail

This is a gap in **this branch's own guard**. It checks that the growth is
arithmetically consistent with an insertion at the caret, but never that the text
outside the claimed range survived. A replacement that grows by the same amount
satisfies the arithmetic:

| old | new | emitted |
|---|---|---|
| `n` @1 | `你好` @2 | `好` |
| `ni` @2 | `你好吗` @3 | `吗` |

Each drops the leading characters — silent loss, not reordering. Worth stating
plainly: on exactly these inputs the code this branch replaced did better, because
`newValue.replace(oldValue, '')` failed to find `n` inside `你好` and returned the
whole string. The replacement is still the right direction — it fixes the
`"ab"`→`"xaby"` mangling that motivated it — but it traded one narrow wrong answer
for another instead of refusing both.

Requiring the text on either side of the range to be unchanged routes these into
the `return` that is already there, which is what AGENTS.md asks for: when the
evidence is ambiguous, emit nothing. All three genuine insertions
(`""`→`"a"`, `"ab"`→`"abc"`, `"ab"`→`"axb"`) are unaffected, and a paired negative
test pins that.

I could not establish which engine produces the replacement shape on this path.
`_handleAnyTextareaChanges` only runs when a composition is *not* active, and a
conversion normally happens inside one. There is a route — a key that finalizes
the composition, after which the engine rewrites the buffer in the same tick — but
I have not observed it. So this is a guard completeness fix, justified by the
guard's own stated intent rather than by a reproduction, and I would rather it not
be read as a fix for an observed user-facing bug.

## Verification of these two

- The five new tests fail on the parent commit with exactly the outputs tabled
  above, and pass after
- 587 test files, 6425 tests, all passing
- `pnpm run typecheck` clean, `regenerate-xterm-patches.mjs --check` in sync


nwparker:

PR #12026, opened 2026-08-02
fix(P1-A): async SSH consumer-recovery persistence and detach on failed connect

AmethystLiang:
## Problem

Release-scan finding **P1-A** from the v1.4.164-rc.0 multi-panel scan. Two independent defects in the SSH PTY consumer-recovery path:

**A1 — main-thread freeze on establish/reconnect.**
`SshRelaySession.rememberPtyConsumerRecovery` → `rememberSshPtyConsumerRecovery` → `Store.upsertSshPtyConsumerRecovery` → `flushSshPtyConsumerRecovery` → `flushOrThrow()` → `writeToDiskSync()` → `writeFileDurableSync()`. That is a full synchronous serialize + `fsync` + rename of the entire profile state file, executed inline on **every** `establish()` and every `reconnect()`. On a stalled or slow profile mount (network home dir, contended disk) the Electron main thread parks for the duration — exactly during live SSH recovery, when the app most needs to stay responsive.

**A2 — a failed connect discards the consumer identity.**
`claimSshPtyConsumerRecovery` reuses an existing in-memory entry only when `detached === true`; otherwise it mints a fresh `clientInstanceId` and drops the persisted record's owner lease. The success path in `doConnect` detaches the outgoing session before replacing it (`src/main/ipc/ssh.ts:1068`), but both **failure** exits — the transport-connect catch and the establish catch — did a bare `activeSessions.delete(targetId)` with no `detach()`. So after any failed connect, the stale non-detached entry was still in `recoveryByTarget`, and the next attempt minted a new identity, losing the owner lease (`ownerGeneration` / `ownerLease`) that `pty.openClient`'s `resume` needs to reclaim relay session ownership.

## Design

Design doc: `docs/reference/plans/2026-08-01-P1-A-ssh-consumer-recovery.md` (local-only — `.gitignore` line 90 keeps `docs/**` planning docs out of the tree; the allow-list only covers the four AGENTS.md-linked reference docs). Its content is summarized here.

**A1 — awaited async durability, not fire-and-forget.** The original ordering guarantee ("ownership is durable before relay setup continues") is worth keeping; only the *blocking* is wrong. New private `Store.flushAsync()` mirrors `flushOrThrow()`'s debounce cancellation and `firstPendingSaveAt` reset, but chains `writeToDiskAsync()` onto the existing `pendingWrite` serialization chain (issue #1158 — backup rotation must not race two writers) instead of blocking. It deliberately does **not** bump `writeGeneration`: fencing is for the sync path, whereas this write chains. The tracked promise swallows rejections so it can't poison the next chained write or surface through `waitForPendingWrite()`; the caller still sees failures through the returned promise.

Rejected alternatives: a sidecar file for the recovery record (new format, new migration, new crash-consistency surface for one small record); dropping the barrier entirely and letting the debounce handle it (loses the durable-before-setup guarantee that motivated the original sync flush).

**A2 — `detach()`, not `dispose()`.** `dispose()` calls `forgetSshPtyConsumerRecovery`, which deletes both the in-memory and the persisted record — the exact identity loss being fixed. `detach()` marks the entry `detached = true`, which is precisely the state `claim` reclaims. Rejected alternative: a `claimSeq` fencing redesign in the recovery module — out of proportion to the bug, and noted as a follow-up in the doc's Non-goals.

## What changed

**A1**
- `src/main/persistence.ts` — new private `flushAsync()`; `upsertSshPtyConsumerRecovery` / `removeSshPtyConsumerRecovery` / `flushSshPtyConsumerRecovery` are now `async` and await it.
- `src/main/ssh/ssh-pty-consumer-recovery.ts` — `rememberSshPtyConsumerRecovery`, `removeSshPtyConsumerOwnerRecovery`, `forgetSshPtyConsumerRecovery` are now `async`.
- `src/main/ssh/ssh-relay-session.ts` — `rememberPtyConsumerRecovery` is `async`; both call sites (`establish()`, `reconnect()`) await it, as does the stale-owner branch of `openPtyConsumerSession`. `dispose()` stays synchronous and fire-and-forgets the removal.

Every in-memory mutation still lands **before the first `await`** in each of these functions, so the map entry, the `Store.state` array, and the `detached` flag all update synchronously. No caller can observe a torn record, `dispose()` remains synchronous from the caller's view, and a quit-time `flushOrThrow()` still writes the already-updated state.

**A2**
- `src/main/ipc/ssh.ts` — new `abandonFailedSshSession(targetId, session)` helper; both failure exits (`:1114`, `:1152`) route through it. It detaches first, tolerates a teardown throw so it can't mask the connect error being rethrown, and only deletes from `activeSessions` when the entry is still this session.

`detach()` was audited for safety on a never-established session: `abortController?.abort()` is optional-chained, `stopPortScanning()` is null-guarded, `broadcastEmptyLists()` is truthful (the connect failed, so there are no ports), `teardownProviders('connection_lost')` is idempotent and null-guards every handle, and `markSshRemotePtyLeases(targetId, 'detached')` is guarded on `state === 'attached'` so it can't resurrect a terminated lease. `establish()`'s own catch resets `_state` to `'idle'` (not `'disposed'`), so `detach()` does not early-return. This is also strictly safer than the old bare delete, which left a live non-disposed session whose `_onRelayLost` callback could still fire for a target no longer in `activeSessions`.

## Tests

Six new tests. **Each was negative-verified** by reverting the corresponding fix and confirming failure, then restoring.

`src/main/persistence-async-write-syscalls.test.ts` (3, following the existing no-sync-syscall harness):
- persists consumer recovery with no sync syscall, durable once awaited — without the fix: `expected [ …(13) ] to deeply equal []`
- removes consumer recovery with no sync syscall, durable once awaited — without the fix: `expected [ …(6) ] to deeply equal []`
- a main-thread timer keeps firing while the write is in flight (the freeze itself: 1000ms stalled write, 10ms heartbeat, asserts worst gap < 500ms) — without the fix: `expected 13123 to be less than 500`

`src/main/ssh/ssh-relay-session-recovery-durability.test.ts` (new file, 1):
- `establish()` stays open until the consumer-recovery write settles — without the barrier (`void` instead of `await`): `expected true to be false`

`src/main/ipc/ssh.test.ts` (2):
- reclaims the consumer identity after a failed transport connect
- resumes the remembered owner lease after a failed establish (asserts `pty.openClient` carries the same `clientInstanceId` **and** `resume: { ownerGeneration, ownerLease }`)

Both fail without the fix with a fresh-UUID mismatch, e.g. `expected 'dd5059d5-…' to be 'f6bd9242-…'`.

## Validation run

- `npx tsc --noEmit -p config/tsconfig.node.json` — clean
- `npx vitest run --config config/vitest.config.ts src/main` — **1338 files / 18074 tests passed**, 9 files / 58 tests skipped
- `npx oxfmt --check -c .oxfmtrc.json <8 touched files>` — all correct
- `npx oxlint <8 touched files>` — clean

**Not run:** Electron runtime and the Docker SSH e2e specs (`ORCA_E2E_SSH_DOCKER=1`). No runtime proof of the freeze relief is claimed beyond the event-loop heartbeat unit test above.

## Residual risk

- **Low.** The durability barrier is preserved, so establish/reconnect still wait the same wall-clock time on a stalled mount — the main thread is simply no longer blocked while it waits. The unbounded wait itself is pre-existing and out of scope; bounding it would be a behavior change.
- `flushAsync` chains onto `pendingWrite` rather than fencing via `writeGeneration`. A concurrent quit-time `flushOrThrow()` bumps the generation and the in-flight async write skips its rename — correct, because the sync flush already wrote the fresher state including this record, and the async promise resolves normally rather than rejecting.
- `dispose()` now fire-and-forgets the persisted removal. The logical removal (in-memory map + `Store.state`) is still synchronous, so only the disk write is deferred; a subsequent quit flush picks it up.

## RC context

Found by the v1.4.164-rc.0 release scan. Both defects are on the live SSH reconnect path, so the impact is concentrated on SSH workspaces recovering after a network blip — the freeze (A1) and a lost resume identity that forces a fresh consumer session (A2).

### DIFF
--- a/src/main/ipc/ssh.test.ts
+++ b/src/main/ipc/ssh.test.ts
@@ -2730,6 +2730,81 @@ describe('SSH IPC handlers', () => {
       expect(order).toEqual(['leases:detached', 'leases:detached:persisted', 'recovery:upsert'])
     })
 
+    it('holds a retry that starts while the detach write is still pending', async () => {
+      const targetId = 'ssh-consumer-identity-pending-retry'
+      const order: string[] = []
+      let settleLeaseRelease!: () => void
+      mockStore.markSshRemotePtyLeasesAsync.mockImplementationOnce((_id: string, state: string) => {
+        order.push(`leases:${state}`)
+        return new Promise<void>((resolve) => {
+          settleLeaseRelease = () => {
+            order.push(`leases:${state}:persisted`)
+            resolve()
+          }
+        })
+      })
+      mockStore.upsertSshPtyConsumerRecovery.mockImplementation(async () => {
+        order.push('recovery:upsert')
+      })
+      mockSshStore.getTarget.mockReturnValue(makeTarget(targetId))
+      mockConnectionManager.connect.mockRejectedValueOnce(new Error('transport refused'))
+
+      const failedConnect = handlers.get('ssh:connect')!(null, { targetId })
+      const failure = expect(failedConnect).rejects.toThrow('transport refused')
+      await vi.waitFor(() => expect(order).toContain('leases:detached'))
+
+      // Retry mid-write: it must not mint a session or re-own leases while the release is pending.
+      mockConnectionManager.connect.mockResolvedValue({})
+      markConnected(targetId)
+      const retry = handlers.get('ssh:connect')!(null, { targetId }) as Promise<unknown>
+      const retrySettled = vi.fn()
+      void retry.then(retrySettled, retrySettled)
+      await new Promise((resolve) => setTimeout(resolve, 0))
+      expect(retrySettled).not.toHaveBeenCalled()
+      expect(order).not.toContain('leases:detached:persisted')
+
+      settleLeaseRelease()
+      await failure
+      // Why: the retry folds onto the still-latched attempt rather than starting a second connect,
+      // so it inherits that attempt's failure instead of racing its teardown.
+      await expect(retry).rejects.toThrow('transport refused')
+      expect(mockConnectionManager.connect).toHaveBeenCalledTimes(1)
+      expect(order).toEqual(['leases:detached', 'leases:detached:persisted'])
+
+      await handlers.get('ssh:connect')!(null, { targetId })
+
+      expect(order).toEqual(['leases:detached', 'leases:detached:persisted', 'recovery:upsert'])
+    })
+
+    it('replaces a live session whose detach write keeps failing', async () => {
+      const targetId = 'ssh-consumer-identity-detach-write-failure'
+      mockSshStore.getTarget.mockReturnValue(makeTarget(targetId))
+      mockConnectionManager.connect.mockResolvedValue({})
+      markConnected(targetId)
+      await handlers.get('ssh:connect')!(null, { targetId })
+      const claimedId = getSshPtyConsumerRecovery(targetId)?.clientInstanceId
+      expect(claimedId).toEqual(expect.any(String))
+
+      // Why a permanent reject, not once: it proves the failed session is gone rather than merely
+      // retried — a second connect that still holds it would fail on the same write again.
+      mockStore.markSshRemotePtyLeasesAsync.mockImplementation((_id: string, state: string) =>
+        state === 'detached'
+          ? Promise.reject(new Error('lease write failed'))
+          : Promise.resolve(undefined)
+      )
+      await expect(handlers.get('ssh:connect')!(null, { targetId })).rejects.toThrow(
+        'lease write failed'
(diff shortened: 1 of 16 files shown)


AmethystLiang:
@coderabbitai[bot] Fixing. Will be in the next commit

AmethystLiang:
@coderabbitai[bot] Fixing. Will be in the next commit

AmethystLiang:
@coderabbitai[bot] Fixing. Will be in the next commit

Issue #11532, opened 2026-07-30
[Bug]: Background PR polling can exhaust the GitHub REST rate limit (5,000/hr) — cached "no PR" branches are re-queried every 60s


silee9019:
### Summary

Orca's background PR polling spawns `gh` continuously and can exhaust the GitHub REST primary rate limit (5,000 req/hour, **per user** — shared with every other tool using the same account). Once exhausted, every `gh` call on the machine fails with HTTP 403 for the rest of the window, including the user's own terminal work and CLI agents.

Two things make the polling heavier than it needs to be:

1. **The cache only suppresses calls for 60 seconds.** `MIN_BACKGROUND_REFRESH_AGE_MS = 60_000` is the only staleness gate, so a branch is re-queried every minute regardless of whether anything could have changed.
2. **Branches with no PR are re-queried just as often.** In my cache, 46 of 81 entries are "no PR" results, and those keep costing a REST call every refresh cycle forever.

There is currently no way to turn this off: I checked all 174 keys in `settings` and the Git settings pane (Branch Prefix, Git Username, Orca Attribution, GitHub API Budget display), and the `orca` CLI has no github command.

### Environment

* **Orca:** 1.4.161
* **OS:** macOS 26.5.2 (25F84), arm64
* **gh CLI:** 2.96.0
* **Scale:** 10 registered repos, 19 Orca worktrees, 81 branch entries in `orca-github-cache.json` (35 with a PR, 46 without)
* **Git host:** [github.com](<http://github.com>)

### Observed behavior

I sampled running processes every 0.4 s for 90 seconds:

* **20 distinct** `gh` **invocations** in that window: `gh api repos/{owner}/{repo}/pulls?head=…&state=all&per_page=1`, `gh pr view <n> --json …statusCheckRollup…`, `gh api graphql`, plus `gh api rate_limit`.
* **All 213 sampled** `gh` **processes had PPID =** `/Applications/Orca.app/Contents/MacOS/Orca`**.** Nothing else on the machine was calling the API.

Cross-referencing the 10 branches queried in that window against `orca-github-cache.json`:

| Branch | In cache? | Cached value | Cache age |
| -- | -- | -- | -- |
| `fix/…` | yes | PR found | 66 min |
| `feature/…` | yes | PR found | 63 min |
| `batch-…` | yes | no PR | 65 min |
| `fix/…` | yes | no PR | 63 min |
| `release/…` (×2) | yes | no PR | 66 min |
[...]

PR #12030, opened 2026-08-02
fix(P1-D): bound hosted-review in-flight lookups so a wedged provider cannot pin a branch

AmethystLiang:
## Problem

`withHostedReviewBranchCache` (`src/main/source-control/hosted-review-branch-cache.ts`) is the single funnel every hosted-review poll passes through — every desktop window, the mobile client and `orca serve` share it (#11532). It dedupes concurrent callers through an `inflight` map whose entry is removed **only** when `lookup()` settles:

```ts
const pending = inflight.get(key)
if (pending) return pending
const request = (async () => { try { ... } finally { inflight.delete(key) } })()
inflight.set(key, request)
```

Nothing bounded how long that took: no deadline, no cancellation, no age bound, no size bound. And the provider path underneath could block indefinitely — `getRemoteUrlForRepo` ran `git remote get-url` with **no** timeout on the local/WSL branch, and `src/main/git/runner.ts` only arms its kill path when a `timeout` is passed.

So if that one call hung (dead NFS/SMB mount, wedged WSL interop):

1. every later poll for the branch joined the same dead promise → the card loaded forever;
2. the entry was never cleared, so the branch could not recover for the life of the process — no backoff, no retry, not even after the host came back;
3. each stalled branch left a permanent `inflight` entry (and a live `git` child), so a multi-branch stall grew the map monotonically.

The renderer's own `timeoutMs: 30_000` runtime-RPC bound does not help: it fails the renderer's request while the main-process entry stays pinned, so every subsequent poll re-joins the dead lookup and times out again. Same for mobile and `orca serve`, which call `runtime.getHostedReviewForBranch` → the same funnel.

**Root cause:** the cache treated "a lookup is in flight" as a state only the lookup itself could leave. Liveness of the whole funnel was delegated to code with no liveness guarantee of its own.

## Design

(The design doc lives at `docs/plans/2026-08-01-P1-D-hosted-review-inflight-deadline.md`, which is gitignored per #11891 — so it is reproduced here.)

**Options considered**

1. *Thread `AbortSignal` through every forge provider.* True cancellation everywhere. Rejected: touches five providers, `gh` exec, the GitLab/Bitbucket/Azure/Gitea adapters and the SSH relay protocol. Large and risky, and unnecessary — the pin is a *caller-side* liveness bug.
2. *Deadline at the IPC handler.* Fixes one caller's spinner and none of the recovery: the shared `inflight` entry stays pinned. Rejected.
3. *Deadline at the cache funnel + a real timeout on the one unbounded git call.* **Chosen.**

**Layer 1 — deadline at the funnel.** Each in-flight lookup gets a wall-clock deadline (`HOSTED_REVIEW_LOOKUP_DEADLINE_MS`, 120s). Nothing below can be cancelled, so on expiry the cache **detaches** rather than cancelling:

- the `inflight` record is deleted, so the branch is no longer pinned;
- the deadline counts as a failure (`noteFailure`), so a wedged provider is not handed a fresh hung process on every 60s poll — the existing exponential backoff applies;
- the caller is released the way the existing failure path releases it: serve the last known review if cached, otherwise reject with a descriptive timeout error.

The detached lookup keeps running and its result is still **adopted** when it lands: on late success it stores the entry and clears the backoff, so the very next poll is served from cache and a slow-but-alive host converges instead of failing forever. On late failure it does not re-count the failure (the deadline already counted it) and never rejects into nowhere.

Three guards keep a detached lookup from corrupting newer state:

- **Scope generation** (already present): a result older than an invalidation is discarded.
- **Supersession**: a late result is dropped if an entry stored *after* this lookup started already exists — a newer lookup's answer always wins.
- **Token identity**: the `finally` clears the `inflight` record only if the map still holds *this* record. Without it, a detached lookup finishing after a replacement started would delete the replacement's entry and orphan its joiners.

**Layer 2 — bound `inflight` independently.** Every record carries `startedAt`, and a sweep at the top of every call expires records older than the deadline. This is the backstop for the `setTimeout` itself being unreliable: macOS/Windows **system sleep** suspends main-process timers arbitrarily, so wall-clock age is the honest bound. Expiry is idempotent whichever trigger fires first. `MAX_INFLIGHT_LOOKUPS` (500, same order as `MAX_ENTRIES`) evicts the oldest record — a memory backstop only, deliberately far above realistic concurrency, because evicting a *live* record costs a duplicate provider call, which is the quota problem #11532 exists to prevent. An expired record does not count a failure if it is no longer the current record for its key, so a post-sleep sweep or a size eviction cannot penalise a branch whose replacement is already healthy.

**Layer 3 — real cancellation for the one unbounded call.** `getRemoteUrlForRepo` gets a 30s `timeout` on the local/WSL branch, which arms `runner.ts`'s own deadline and kills the process tree — Node's `timeout` alone waits forever on a signal-ignoring child. `git remote get-url` is a local config read; 30s (the repo's existing `DEFAULT_GH_EXEC_TIMEOUT_MS` precedent) can only be exceeded by a wedged host, where failing beats hanging. The SSH branch is left alone: `provider.exec` → `mux.request` already defaults to 30s (`ssh-channel-multiplexer.ts:44`) plus a 30s stream-inactivity bound, so an explicit `timeoutMs` would restate the default and churn the call-shape assertions.

## What changed

| File | Change |
| --- | --- |
| `src/main/source-control/hosted-review-branch-cache.ts` | `inflight` holds a record (token / `startedAt` / promise / idempotent `expire`) instead of a bare promise; new `startLookup` runs each lookup under the deadline with detach + late adoption; `releaseInflight` token guard, `expireOverdueInflight` wall-clock sweep, `trackInflight` size cap, `answeredSince` supersession guard |
| `src/main/source-control/hosted-review-lookup-backoff.ts` | **New.** The per-branch failure escalation moved out verbatim — it has its own lifetime (a branch stays backed off after its answer is invalidated) and is what a deadline records against. Also keeps the cache file under the 300-line `max-lines` rule |
| `src/main/source-control/hosted-review-refresh-pacing.ts` | New `HOSTED_REVIEW_LOOKUP_DEADLINE_MS` (120s) and `MAX_INFLIGHT_LOOKUPS` (500) |
| `src/main/github/github-repository-identity.ts` | 30s `timeout` on the local/WSL `git remote get-url` |
| `src/main/source-control/hosted-review-branch-cache.test.ts` | +9 tests (below) |
| `src/main/github/gh-utils.test.ts` | Exact-argument assertions updated for the new `timeout` |

No IPC, preload, renderer or mobile surface changes — the funnel is shared, so desktop, mobile and `orca serve` all recover from the one fix. The only user-visible new text is the timeout error when a lookup exceeds the deadline **and** nothing is cached; where a review *is* cached the card keeps showing it, unchanged from today's failure behaviour.

## Tests

9 new tests in `hosted-review-branch-cache.test.ts` (fake timers, as the file already uses):

1. releases a stuck lookup at the deadline (the branch is no longer pinned);
2. lets a later poll succeed after a timeout;
3. keeps the last known review when a refresh times out;
4. adopts a lookup that lands after its deadline (next poll served from cache, no second provider call);
5. clears the backoff when a timed-out lookup finally answers;
6. does not let a straggler overwrite a newer answer (supersession guard);
7. does not let a straggler evict the lookup that replaced it (token guard);
8. releases a caller by wall clock when the deadline timer never fires (the system-sleep case);
9. bounds the in-flight map independently of the completed cache.

**Proof the tests are meaningful:** with the pre-fix `hosted-review-branch-cache.ts` restored from `HEAD` and the new tests in place, **9 tests fail** — 8 of them by hanging to the vitest 5s timeout, which *is* the pin — and with the fix restored all 29 pass.

**Ran on this branch (macOS, local):**

- `npx vitest run src/main/source-control/ src/main/github/` → **51 files, 706 tests passed**
- `npx tsc -p config/tsconfig.node.json --noEmit` → clean
- `npx oxlint -c .oxlintrc.json src/main/source-control/ src/main/github/` → clean
- `npx oxfmt --check -c .oxfmtrc.json <6 changed files>` → clean

No e2e was run: the change is main-process cache logic with no UI surface of its own, and reproducing it end-to-end requires a wedged host.

## Residual risk

- **A legitimately slow host now fails at 120s** instead of hanging. Mitigated by the deadline being generous relative to the bounded steps underneath (30s per `gh` exec, plus retries) and by late-result adoption — the answer still lands and the next poll serves it. Worst case is one error card, not a stuck one.
- **Backoff escalation on a chronically slow host** caps at 15 min and always serves the last known review; late adoption clears the backoff as soon as any lookup completes.
- **The 30s `git remote get-url` bound affects non-hosted-review callers** (`client.ts`, `gh-utils.ts`, `github-enterprise-repository.ts`). Accepted: they already handle throws, and a `git remote get-url` that exceeds 30s means a wedged host on every path, not just this one.
- **Real cancellation of the provider API call is still future work** (option 1). A detached lookup keeps its `gh`/HTTP work running to completion; the deadline bounds the *pin*, not the child process — except on the `git remote get-url` step, which is now genuinely killed.
- No migration, no setting, no feature flag.

Found by the v1.4.164-rc.0 release scan (finding P1-D).


AmethystLiang:
@coderabbitai[bot] resolved

AmethystLiang:
@coderabbitai[bot] fixed

PR #4840, opened 2026-06-07
fix(windows): slow launch + keystroke stalls — main process synchronously spawns powershell to re-harden env-store ACLs

LesleyMurfin:
Fixes #4839

## What this fixes

On Windows, the desktop app is slow to launch (~2×) and stalls on every keystroke in
bursts. Root cause: the Electron **main process synchronously spawns `powershell.exe`**
to re-apply NTFS ACLs to the runtime environment-store JSON on every runtime round-trip
(and at startup). `execFileSync` blocks the main thread ~1–1.5s per PowerShell cold
start, and a single `markEnvironmentUsed()` triggers ~5 of these spawns
(read-path + write-path hardening), so typing — which round-trips to the runtime
continuously — freezes the UI.

This is `process.platform === 'win32'`-gated, so a headless Linux `orca serve` never
hits it; only the Windows client does.

### Measured impact (same machine, 3 runs each)

| Build | Median launch-to-responsive |
| --- | --- |
| Before | **14.0s** (12.6–16.4s) |
| After | **6.8s** (6.7–7.2s) |

CDP performance trace during paced typing: the Electron main thread goes from
**95.7% blocked in synchronous `spawn`** to **99.0% idle**. ~51% faster launch and
keystroke latency disappears.

## The change

Two small, composable fixes — the hardening behavior is **preserved**, it just stops
blocking the main thread redundantly:

1. **`src/shared/runtime-environment-store.ts` — throttle `markEnvironmentUsed`.**
   `lastUsedAt` is a non-security coarse-freshness timestamp, yet stamping it forced a
   full read+harden+write+harden cycle on every runtime call. Skip the rewrite when the
   `runtimeId` is unchanged and `lastUsedAt` was persisted within the last 60s. A
   `runtimeId` change inside the window still persists immediately. This removes
   essentially all per-keystroke spawns.

2. **`src/shared/secure-file.ts` — make directory/file hardening idempotent per process.**
   A process-lifetime `Set` of already-hardened paths makes the read-path re-harden and
   repeated directory hardening no-ops after the first. The post-rename target write —
   the one case that legitimately needs re-hardening (new inode) — is deliberately kept
   always-on and refreshes the cache entry. `hardenSecurePath` stays pure;
   `hardenSecurePathOnce` is the new caching wrapper.

A test-only reset hook (`__resetSecureFileHardenedPathsForTests`) mirrors the existing
`__resetSecureFileWindowsUserSidForTests`.

## No visual change

[...]

PR #12036, opened 2026-08-02
fix(mobile): stop blocking the E2EE handshake on a synchronous ACL write

mozuvic1:
## The symptom

On Windows, mobile pairing fails on any machine with a real workspace loaded. The phone reaches the desktop, opens the WebSocket, completes the key exchange — and then dies waiting for the server. This is the pairing log from the phone:

```
+19.3s  ✓ WebSocket open            Starting E2EE handshake
+19.3s  •  Sent e2ee_hello          Awaiting server e2ee_ready
+19.3s  ✓ Received e2ee_ready       Sending device token
+24.3s  ✗ Handshake timeout         No e2ee_ready/e2ee_authenticated within 5s
```

Exactly 5.0s of silence after the token goes out. Meanwhile the desktop shows the phone as **paired**, and nothing appears in its logs — so the failure reads as a phone-side problem with no server-side trace.

## The cause

`E2EEChannel.handleAuth` calls `onReady` before signalling success, deliberately:

```ts
// Why: transport-bound identity checks must complete before the peer sees
// authentication success; ...
this.onReady(this, authenticatedDevice)
this.sendEncryptedControl({ type: 'e2ee_authenticated' })
```

`MobileSocketWiring`'s `onReady` ends with `deviceRegistry.updateLastSeen(...)`, which persists the registry via `writeSecureJsonFile` → `writeSecureFile`. On Windows that path spawns PowerShell **twice, synchronously**, to apply ACLs — once for the temp file, once for the published path:

```ts
applySecurePathRestriction(tmpFile, false, process.platform, true)   // sync
renameSync(tmpFile, targetPath)
applySecurePathRestriction(targetPath, false, process.platform, true) // sync
```

The temp file carries a fresh `pid.timestamp.random` name on every write, so the hardening cache can never cover it — the cost is paid in full on every call.

This cost is already known in this repo. `secure-path-windows-acl.ts` says:

> async to avoid blocking the main thread — sync PowerShell cold-start (~1-1.5s) on the frequent read path stormed it (#4901)

The read path was made async for exactly this reason (#4840, #5011). This write path was not, and it sits inside the authentication handshake.

## Measurement

Instrumented the wiring's `onReady` on Windows 11, Node 24.18:

```
identity binding = 3727 ms | consumer hook = 0 ms
```

3.7s against the mobile client's 5s budget, on an **idle dev profile**. With a real workspace it goes over and pairing becomes impossible. Reproduced at 3405 ms and 3439 ms on separate runs, so it is stable, not a one-off.

The registry entry is created when the QR is generated, which is why the desktop claims success. `lastSeenAt` stayed `0` throughout — and `updateLastSeen` is the only thing that sets it, so that field is direct proof the handshake never completed.

On Linux and macOS `applySecurePathRestriction` is a `chmodSync`. That is why this reproduces only on Windows, and why the same setup pairs fine against a Linux host.

## The fix

Keep the in-memory update on the caller stack; defer the persist. The persist-before-memory-swap ordering from the original comment is preserved, just off the hot path.

This does not weaken the property the `e2ee-channel.ts` comment protects: the identity binding — `validateToken`, `setClientId`, socket registration — still completes synchronously before success is signalled. `lastSeenAt` is display and rotation metadata, not an auth gate.

## Tests

New `device-registry.test.ts`:

- `updateLastSeen` does not call `writeSecureJsonFile` on the caller stack, and does call it on the next macrotask — this is the regression itself
- the timestamp is recorded once the deferred write lands
- an unknown device id schedules no write at all

Confirmed the first test fails without the production change (`expected "vi.fn()" to not be called at all, but actually been called 1 times`), so it genuinely covers the regression.

## Verification

Windows 11, Node 24.18.0:

- `device-registry.test.ts` — 3/3 pass
- `pnpm typecheck:node` — clean
- `oxlint` on both files — clean
- **Real-world:** built and installed this branch on a machine where pairing failed on every attempt. It now pairs, and `lastSeenAt` is recorded in `orca-devices.json`.

## Visual changes

None.

## Review notes

- **Cross-platform:** the change is platform-neutral; the pathology is Windows-only because that is where the ACL apply shells out. POSIX behaviour is unchanged apart from the write landing a tick later.
- **Local / remote / SSH:** the relay path authenticates through the same `MobileSocketWiring.onReady`, so it benefits identically. Nothing transport-specific is touched.
- **Security:** no change to what is authenticated or when. The identity binding still precedes the success frame. The deferred work is a metadata timestamp.
- **Performance:** removes ~3.7s of synchronous main-process work from every mobile authentication.
- **Failure mode:** if the deferred write fails, the behaviour matches today's failure path — the entry keeps its previous `lastSeenAt` on disk. A process exiting in the same tick as an authentication could lose one timestamp update; that field is refreshed on the next connection.

One thing worth flagging: **#12015** adds `writeSecureFileAsync` and moves ~90 call sites off the main thread, but keeps the sync twin "for CLI/startup callers that cannot await" and does not touch `device-registry.ts` — so this path still blocks after it lands. If you'd rather this migrate to `writeSecureJsonFileAsync` on top of that PR instead of using `setImmediate`, say the word and I'll rebase.

X: @vicmandarin

### DIFF
--- a/src/main/runtime/device-registry.test.ts
+++ b/src/main/runtime/device-registry.test.ts
@@ -1,0 +1,63 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const { writeSecureJsonFileMock, hardenExistingSecureFileMock } = vi.hoisted(() => ({
+  writeSecureJsonFileMock: vi.fn(),
+  hardenExistingSecureFileMock: vi.fn()
+}))
+
+vi.mock('../../shared/secure-file', () => ({
+  writeSecureJsonFile: writeSecureJsonFileMock,
+  hardenExistingSecureFile: hardenExistingSecureFileMock
+}))
+
+import { DeviceRegistry } from './device-registry'
+
+function nextMacrotask(): Promise<void> {
+  return new Promise((resolve) => setImmediate(resolve))
+}
+
+describe('DeviceRegistry', () => {
+  beforeEach(() => {
+    writeSecureJsonFileMock.mockReset()
+    hardenExistingSecureFileMock.mockReset()
+  })
+
+  it('keeps the last-seen persist off the caller stack (mobile pairing handshake)', async () => {
+    // Why: updateLastSeen runs inside the E2EE auth path, before the server sends
+    // e2ee_authenticated. writeSecureJsonFile costs seconds on Windows because
+    // writeSecureFile spawns PowerShell twice, synchronously, to apply ACLs — past
+    // the mobile client's handshake budget, which made pairing impossible there.
+    const registry = new DeviceRegistry('/does-not-exist')
+    const device = registry.addDevice('Phone')
+    writeSecureJsonFileMock.mockClear()
+
+    registry.updateLastSeen(device.deviceId)
+    expect(writeSecureJsonFileMock).not.toHaveBeenCalled()
+
+    await nextMacrotask()
+    expect(writeSecureJsonFileMock).toHaveBeenCalledTimes(1)
+  })
+
+  it('records the last-seen timestamp once the deferred write lands', async () => {
+    const registry = new DeviceRegistry('/does-not-exist')
+    const device = registry.addDevice('Phone')
+    expect(device.lastSeenAt).toBe(0)
+
+    registry.updateLastSeen(device.deviceId)
+    await nextMacrotask()
+
+    const stored = registry.listDevices().find((d) => d.deviceId === device.deviceId)
+    expect(stored?.lastSeenAt).toBeGreaterThan(0)
+  })
+
+  it('ignores an unknown device without scheduling a write', async () => {
+    const registry = new DeviceRegistry('/does-not-exist')
+    registry.addDevice('Phone')
+    writeSecureJsonFileMock.mockClear()
+
+    registry.updateLastSeen('not-a-real-device-id')
+    await nextMacrotask()
+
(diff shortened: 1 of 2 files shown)


mozuvic1:
Good catch on the clobbering window — it was real, and I reproduced it before fixing.

The first cut deferred both `this.save()` and `this.devices = nextDevices`, carrying a snapshot taken before the `setImmediate`. `removeDevice` persists synchronously, so a revocation landing in that window was overwritten when the stale snapshot flushed, resurrecting the revoked device in memory and on disk. The review was also right that the description claimed the in-memory update stayed on the caller stack while the code did not — that divergence is exactly where the bug hid.

Fixed in c32556bd3:

- the timestamp is applied in memory synchronously, on the caller stack
- the deferred write reads `this.devices` at flush time instead of closing over a snapshot, so a removal that lands in between is persisted as-is
- deferred writes coalesce, so an authentication burst costs one write rather than one per device

Five tests now, and I verified all three new ones fail against the previous implementation — the resurrection case fails with `expected [ …(2) ] to deeply equal [ Array(1) ]`, which is the revoked device coming back.

`pnpm typecheck:node` and `oxlint` are clean on both files.

On the docstring-coverage warning: this file and its neighbours use the repo's `// Why:` comment convention rather than docstrings, so I've left that as-is — happy to add them if a maintainer would prefer.

PR #12037, opened 2026-08-02
refactor(ime): remove two duplicated pieces of the terminal IME surface

nwparker:
## Summary

Two small reductions in the terminal IME surface, found while reviewing the rest of this stack. No behavior change; both are covered by the existing suites.

**1. One candidate-digit predicate instead of two.** `terminal-ime-linux-candidate-state.ts` carried a private `isPlainAsciiDigitKey`, and `terminal-ime-candidate-key-release-guard.ts` exports `isTerminalImeCandidateDigitKeyEvent`. The two were extensionally identical — unmodified ASCII digit, same four modifier checks — written twice and free to drift. The Linux state now imports the exported one.

Only the *predicate* is shared. The two guards deliberately keep their separate state: the Linux state is a 1500 ms suspicion window inferred from an orphaned keyup when there is **no** composition, and the release guard is a 250 ms post-composition window for a key the IME is known to own. Those are different evidence regimes, and fusing them would let a heuristic's confidence leak into the evidence-based path. A stateless, clockless predicate is the safe thing to share; the state machines are not.

**2. `isActive()` removed from the input-source tracker.** Nothing in the app ever called it — only its own spec did. Every one of those assertions is more precise stated against `getFeatures()`, which is the real production surface, so the spec got better rather than merely shorter. The `refresh()` export stays: it is a genuine test seam that lets specs await one read instead of racing the fire-and-forget refresh, and it now says so in a comment.

## Screenshots

No UI change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 228 files / 2979 tests across the terminal-pane and dashboard-popout suites
- [ ] `pnpm build` — not run
- [x] Existing tests cover both changes; the input-source spec was tightened as part of (2)

## AI Review Report

Checked for an import cycle before (1): `terminal-ime-linux-candidate-state` → `terminal-ime-candidate-key-release-guard` → `terminal-ime-composition-tracker`, and nothing in that chain imports back. Clean.

Deliberately **not** done: the unmodified-modifier idiom appears in ten files under `terminal-pane/`, each with a different modifier combination because each has different semantics (Ctrl chords, Shift+Space width toggle, and so on). Hoisting that into one shared helper would be a cross-cutting change to non-IME keyboard code and exactly the kind of generic catch-all module the repo's naming rule warns about. Only the two predicates that were genuinely the same were merged.

Cross-platform: (1) is Linux-path code but the predicate is platform-neutral and unchanged in behavior. (2) is macOS-only code; the removed accessor had no callers on any platform.

## Security Audit

No IPC, path handling, auth, secrets, command execution, or dependency changes. Both changes are internal API reductions in the renderer.

## Notes

Top of stack #11899, on #12021. Independent of the rest of the stack — it touches no file any other PR in the stack modifies.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-ime-input-source.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-input-source.test.ts
@@ -3,6 +3,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import {
   _resetMacNativeTextInputSourceTrackerForTests,
   createMacNativeTextInputSourceTracker,
+  DISABLED_MAC_NATIVE_TEXT_INPUT_SOURCE_FEATURES,
   getMacNativeTextInputSourceFeatures,
   getMacNativeTextInputSourceTracker
 } from './terminal-ime-input-source'
@@ -86,9 +87,7 @@ describe('createMacNativeTextInputSourceTracker', () => {
       readInputSourceId: async () => sourceId
     })
 
-    expect(tracker.isActive()).toBe(false)
     await tracker.refresh()
-    expect(tracker.isActive()).toBe(false)
     expect(tracker.getFeatures()).toEqual({
       forwardAsciiPunctuation: false,
       forwardShortTextReplacements: false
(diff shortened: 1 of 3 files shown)


nwparker:

PR #12040, opened 2026-08-02
fix(browser): remove unsafe window close bypass

nwparker:
## Summary

- Keep the guest lifecycle cleanup and early `window.close()` guard from #11910.
- Remove the CLI-only guard bypass, its synthetic preload marker, and persisted `allowWindowClose` state.
- Apply the guard uniformly to every embedded browser webview and retain focused coverage for link-created and directly created tabs.

## ELI5

The first fix installed a child lock so web pages could not accidentally destroy their little browser window inside Orca. It also gave CLI-created tabs a secret key to bypass that lock. Adversarial testing showed the key did not cleanly close the Orca tab: it killed the browser engine underneath while leaving a dead-looking tab behind. That dead tab stayed on screen but every later operation failed with `Invalid guestInstanceId`.

This PR takes away the unsafe secret key. Every embedded browser tab gets the same child lock. We keep the important part of #11910 that throws away stale pointers when a guest really dies, so the original #11763 crash remains fixed. Removing the synthetic file-URL key also removes the path mechanism that caused the Windows black screen fixed in #12038.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [x] Electron production bundle (`pnpm exec electron-vite build --mode e2e`)
- [x] 283 focused Vitest tests
- [x] 2 focused Electron E2E tests covering link-created and directly created `window.close()` attempts
- [x] Changed-code quality gate
- [x] 61 reliability-gate manifest checks
- [x] max-lines ratchet

## AI Review Report

The review split #11910 into its independent lifecycle-cleanup and close-policy layers, traced all creation/restoration paths, and tested the claimed CLI self-close behavior instead of relying on the existing replaceability test. Calling native `window.close()` on a bypassed tab resolved, destroyed its Electron guest, left the `<webview>` connected, and made the next guest call fail with `Invalid guestInstanceId`. The bypass was therefore removed while the identity-guarded forward/reverse map cleanup was retained.

Cross-platform review covered macOS, Linux, and Windows Electron preload handling and file-URL semantics. The resulting implementation has no platform-specific marker or path conversion.

## Security Audit

The renderer can no longer use a synthetic preload value to opt a guest out of the main-process close policy. `will-attach-webview` strips any renderer-requested preload and always replaces it with the standalone sandboxed guard after validating the URL and registered partition. Node integration remains disabled, context isolation and sandboxing remain enabled, and no auth, secrets, dependencies, shell commands, or external input formats changed.

## Notes

This is a targeted partial revert of #11910, not a full revert. A full revert would reintroduce #11763's stale guest pointer and main-process crash.

### DIFF
--- a/src/main/window/createMainWindow.test.ts
+++ b/src/main/window/createMainWindow.test.ts
@@ -74,8 +74,6 @@ import {
 } from './createMainWindow'
 import { ipcMain } from 'electron'
 import { shouldRecoverRendererAfterProcessGone } from '../crash-reporting/process-gone-classification'
-import { BROWSER_WINDOW_CLOSE_ALLOWED_PRELOAD } from '../../shared/browser-window-close-policy'
-import { fileURLToPath } from 'node:url'
 
 function withPlatform<T>(platform: NodeJS.Platform, run: () => T): T {
   const original = process.platform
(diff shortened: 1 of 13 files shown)


nwparker:
Addressed the legacy hydration finding in d79eaa07f1: hydration now strips `allowWindowClose` even when callers pass pre-migration in-memory objects that bypass schema parsing, with a focused regression test.

PR #12042, opened 2026-08-02
fix(remote): keep tab order and titles stable on a headless host

gal064:
## Problem

On a remote `orca serve` host the tab bar is unstable:

- **Order rotates.** Whichever tab you click jumps to the rightmost slot, so repeated clicking permutes the tab bar.
- **Idle tabs get relabelled** to the literal string `Terminal`, losing their real title until you click them.

Both come from the same root: the host publishes degraded state for a pane it is not actively streaming, and the client reads it as fact.

## Cause

**Order.** Every publish path re-materializes a surface by dropping it from the tabs array and pushing it back on the end. Both order builders then rebuilt `tabOrder` from that array order, keeping only the existing group's *id* and discarding its stored order — so activating an idle tab moved it to the end.

**Titles.** A pane with no live PTY carries the placeholder `Terminal` rather than an empty title, and the client adopted it over the real title it already held.

## Fix

- `buildHeadlessMobileSessionTabGroups` (activation path) and `mergeMobileSessionTabGroups` (PTY-publish path) both retain the stored per-group order for still-live tabs and append only genuinely-new ids. Per-group, so a split keeps its internal order.
- `buildMirroredTerminalTabs` treats the `Terminal` placeholder on a `pending-handle` surface as unknown and keeps the retained title. A `ready` surface still adopts whatever the host reports.

## Tests

`headless-tab-order-stability.test.ts` (4 tests) reproduces the rotation across a sequence of activations and pins both builders — 3 of 4 fail before the fix. Two `web-session-tabs-sync` tests cover the placeholder and confirm a `ready` pane still adopts a real host title.

Typecheck and the full `web-session-tabs-sync` suite (65 tests) pass.

## Notes

Found while running agents on a headless `orca serve` host. No upstream issue tracks this. Two sibling PRs follow, covering agent status and completion notifications on the same host type; this one is independent of both.

### DIFF
--- a/src/main/runtime/headless-tab-order-stability.test.ts
+++ b/src/main/runtime/headless-tab-order-stability.test.ts
@@ -1,0 +1,149 @@
+import { describe, expect, it } from 'vitest'
+import { OrcaRuntimeService } from './orca-runtime'
+import type {
+  RuntimeMobileSessionTabsSnapshot,
+  RuntimeMobileSessionTerminalTab
+} from '../../shared/runtime-types'
+
+const WT = 'repo-1::/home/gal/dev/orca'
+const GROUP = `headless-terminals:${WT}`
+
+const terminalTab = (n: number, isActive = false): RuntimeMobileSessionTerminalTab => ({
+  type: 'terminal',
+  id: `tab-${n}::leaf-${n}`,
+  title: `terminal ${n}`,
+  parentTabId: `tab-${n}`,
+  leafId: `leaf-${n}`,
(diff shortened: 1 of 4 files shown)


Jinwoo-H:
Headless session grouping still lacks retained-order seeding and the missing-tab placeholder guard. We are taking the headless-specific fix into owner implementation under #17541. Closing the submitted branch; full credit to @gal064. No action is requested from you. Thank you.\n\nOwner issue: https://github.com/stablyai/orca/issues/17541

Issue #1437, opened 2026-05-05
[Mobile-Beta-Feedback]: terminal icon aways loading


yuuzo:
### Operating system

Other

### Details

In the mobile orca, before click in one of the terminals if they are working there is the loading icon, but even when tem stop working the loading icon continues playing forever.

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/1f3637d8-2ab7-4110-ba74-b4def5885aa2" />

Jinwoo-H:
Thanks for flagging this, it will be fixed in a new release later today 👍

PR #12044, opened 2026-08-02
fix(remote): report a running agent as busy on a headless host

gal064:
## Problem

On a remote `orca serve` host, a running agent flips to the idle checkmark the moment its tab loses focus, and returns to a spinner when the tab is clicked again. The agent is still working the whole time.

## Cause

Two compounding defects, both about a pane the host is not streaming.

**The host published a title-derived status.** `buildPtyMobileAgentStatus` reads *identity* from the agent hook but *state* from the PTY title. A pane with no live handle carries the literal `Terminal` placeholder, which classifies as non-agent — so the branch published `done` over a running agent.

**The client then deleted what it held.** A pane the host reports without an `agentStatus` is *unknown*, not *idle*, but the mirror pruned the entry — so even a correct status was dropped on the next republish.

## Fix

- `getHookAgentRowForPane` returns the hook's `state`/`prompt`/`stateStartedAt` alongside the identity it already returned, and the hook-only branch prefers them over the title-derived fallback. Hook state is bounded by the same staleness window as `agentType`, so an abandoned pane cannot keep claiming `working`; `providerSessionOnly` rows never contribute liveness.
- `buildMirroredAgentStatusPatch` retains status when a `pending-handle` surface reports none.

The retention is deliberately scoped to `pending-handle`: a `ready` surface the host *is* streaming genuinely has no agent and must still prune, or a shell that reclaimed the pane keeps a stuck spinner (#1437). There is a regression test for exactly that.

## Verification

Confirmed against a live host by comparing its `agent-hooks/last-status.json` (two panes `working`, two `waiting`) with what it published for the same panes (`done` for all four). After the fix the published state matches the hook.

## Tests

`headless-agent-status-from-hooks.test.ts` covers `working`/`waiting`/`done` from the hook row. Two `web-session-tabs-sync` tests cover retention and the streamed-pane prune guard. Typecheck and the full 65-test sync suite pass.

## Notes

No upstream issue tracks this. Independent of the tab-order PR. A follow-up PR adds completion notifications for serve hosts and **depends on this one** — wiring notifications while the host still reports `done` for unfocused panes would fire a false completion on every tab switch.

### DIFF
--- a/src/main/runtime/headless-agent-status-from-hooks.test.ts
+++ b/src/main/runtime/headless-agent-status-from-hooks.test.ts
@@ -1,0 +1,96 @@
+import { describe, expect, it } from 'vitest'
+import { OrcaRuntimeService } from './orca-runtime'
+import type { AgentStatusIpcPayload } from '../../shared/agent-status-types'
+import type { RuntimeMobileSessionTerminalTab } from '../../shared/runtime-types'
+
+const TAB_ID = '258c3b52-ed9c-4494-88d7-8577515b8987'
+const LEAF_ID = '57b25db0-75f7-4d6d-90d9-9d1d7b9a3466'
+const PANE_KEY = `${TAB_ID}:${LEAF_ID}`
+
+const tab: RuntimeMobileSessionTerminalTab = {
+  type: 'terminal',
+  id: `${TAB_ID}::${LEAF_ID}`,
+  // The placeholder a headless host publishes for a pane it holds no live handle for.
+  title: 'Terminal',
+  parentTabId: TAB_ID,
+  leafId: LEAF_ID,
+  isActive: false
+}
+
+const WORKTREE_ID =
+  'cc1bd8a8-6e33-49f4-ae20-be69ba290954::/home/gal/dev/ai-projects::workspace:86c20dd7-9943-42a1-8868-3d71aac6a281'
+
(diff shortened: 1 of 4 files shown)


Jinwoo-H:
Thank you for the careful diagnosis, implementation, and tests here. Since this was published at aaa937cd0ddeaf309a7c56863acef1cad84fe2e6, current main at 473442865413787443f47ff6e08059ede241624f gained the authoritative host-side solution in #12223: fresh hook state is projected with title-recency, staleness, resume-row, restored-row, and retirement guards.

In an isolated headless Orca host paired to a separate Electron client, a real Claude hook from a live fixture PTY reached every busy-state assertion in 10.9 s: the rendered tab became inactive and remained working, while the host snapshot carried the same working state and exact prompt. Disabling only the current host projection made the same topology report done instead of working and time out at 34.8 s. The 10.9 s probe then failed only on an unrelated assumption that host-global isActive should follow viewer-local selection.

This therefore cannot land as written: its host change is superseded, and retaining a missing pending-handle status in the renderer can preserve stale working state after the authoritative host has stopped publishing it. A current-main repro that removes the cross-view selection assumption, still demonstrates a host-authoritative gap, and proves working-to-done/row-clear plus shell-reclamation behavior would justify reopening. We appreciate the contribution and the clear report.

PR #12049, opened 2026-08-02
fix(terminal): require a physical Enter for the Windows IME Process key

### LINKED ISSUE #11878: [Bug]: Korean double consonants (쌍자음) trigger automatic line breaks in the Windows integrated terminal

lslogis:
Fixes #11878.

## The bug

On Windows, typing a Korean double consonant (쌍자음 — `ㄲ ㄸ ㅃ ㅆ ㅉ`, all of which are **Shift + a jamo key**) inserts a newline into the terminal mid-word. No Enter is pressed. English input is unaffected.

```
orca 에서 [\n] 쓰고 [\n] 있는데      <- breaks land right before 쓰/있 (ㅆ)
줄뀜 [\n] 자꾸                       <- ㄲ
에디트할라고 [\n] 뜨는데              <- ㄸ
```

## Root cause

```ts
export function isTerminalImeProcessEnter(event) {
  return (
    event.key === 'Process' &&
    event.keyCode === 229 &&
    getTerminalImeModifiedEnterKind(event) !== null
  )
}
```

`key === 'Process'` / `keyCode === 229` is the generic signal Windows raises for **every** keystroke an IME consumes — it is not specific to Enter. The predicate therefore has no idea which key was actually pressed.

Pressing Shift + a jamo during composition raises precisely that event with `shiftKey: true`, so:

1. `getTerminalImeModifiedEnterKind(event)` returns `'shift'` (not null)
2. `isTerminalImeProcessEnter(event)` returns **true**
3. `keyboard-handlers.ts:502` sets `imeProcessEnter`, and the handler synthesizes a fake `{ key: 'Enter', shiftKey: true }`
4. `resolveShortcutEvent` resolves that to `sendInput`, and `deferredNewlineSender.defer(...)` writes a Shift+Enter newline to the pty

## The fix

`event.code` reports the **physical** key and is not masked by the IME. A jamo reports `KeyT`/`KeyR`/`KeyE`/…; only a real Enter reports `Enter` or `NumpadEnter`. Gating on it lets the rest of the IME Enter machinery work unchanged.

Worth noting the existing handler tests in `keyboard-handlers-ime.test.tsx` already build their Process events with `code: 'Enter'` — the data needed to disambiguate was there, the predicate just never looked at it.

## Tests

- `terminal-ime-deferred-newline.test.ts`: added `code: 'Enter'` to the `isTerminalImeProcessEnter` event factory, a `NumpadEnter` acceptance case, and a regression table asserting Shift + each of the five double consonants is rejected.
- `keyboard-handlers-ime.test.tsx` needs no change (already sets `code: 'Enter'`).

**I was not able to run the suite locally** — this was developed against the shipped bundle on a Windows box without the full pnpm workspace installed, so please let CI verify. The change is confined to one predicate and its direct tests.

## Verification

Verified by byte-patching the released 1.4.163 `app.asar` with the equivalent change (swapping the `keyCode === 229` probe for the `code === 'Enter'` check) and restarting Orca: Korean input became correct immediately, and the newline injection stopped. An earlier attempt that instead cleared the stale-modifier fallback in `getImeEnterModifier()` had **no effect**, because that path is bypassed whenever `event.shiftKey` is genuinely true — which it is while Shift is held for the double consonant.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
@@ -325,6 +325,7 @@ describe('isTerminalImeProcessEnter', () => {
   const event = (overrides: Partial<KeyboardEvent> = {}) =>
     ({
       key: 'Process',
+      code: 'Enter',
       keyCode: 229,
       metaKey: false,
       ctrlKey: false,
@@ -340,6 +341,10 @@ describe('isTerminalImeProcessEnter', () => {
     }
   )
 
+  it('recognizes the numeric keypad Enter', () => {
+    expect(isTerminalImeProcessEnter(event({ code: 'NumpadEnter' }))).toBe(true)
+  })
+
   it.each([
     { key: 'Enter' },
     { keyCode: 13 },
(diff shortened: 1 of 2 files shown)


OrcaWin:
Thank you for the clear diagnosis and focused fix. We tested the exact published parent (`4be4d10ae0bef2a721282f05f285cfef66f66928`), published head (`a7ed19fc287c9556e79f97ff794f618887305699`), current main (`473442865413787443f47ff6e08059ede241624f`), and classifier-disabled child (`1291699f3a143309e3415fd05a53d0186eadfd55`). Physical Microsoft Korean IME injection was non-discriminating: all four trees produced correct doubled-jamo and composing Shift+Enter PTY bytes. The selected CDP case passed the published parent/head but duplicated Shift+Enter on current/disabled, so that split is a topology-specific harness disagreement, not evidence of a current native regression.

We cannot land this head because merged #12265 (`173abf8b121812ff3374e7751313d3d2627682bc`) is the broader current implementation: it handles Process-key ownership and press evidence, rejects physical non-Enter codes, and preserves blank/`Unidentified` code compatibility that this head drops. A deterministic failure on current main using physical Microsoft Korean IME input—not CDP—with the exact Windows/Electron/IME versions, raw DOM key/composition sequence, and PTY bytes would justify reopening. We appreciate the detailed report, shipped-bundle verification, and narrowly scoped patch; they materially helped validate the broader fix.

PR #12050, opened 2026-08-02
fix(windows): adopt persisted PATH ordering in new terminals

### LINKED ISSUE #11992: [Bug]: New pwsh tabs keep stale Windows PATH ordering after installing Python
### Operating system

Windows

### Orca version

1.4.162

### Short summary

After installing Python while Orca is running, a newly created pwsh tab can keep Orca's stale PATH ordering and resolve the broken `WindowsApps\python.exe` alias instead of the newly installed interpreter. A fresh pwsh outside Orca and a Zed terminal resolve Python correctly.

### Steps to reproduce

1. Install and launch Orca before Python is installed.
2. Keep Orca running.
3. Install Python 3.14.6 for the current user and add it to PATH.
4. Create a new pwsh tab in Orca.
5. Run `(Get-Command python).Path` and `python --version`.
6. Run the same commands in a newly launched pwsh outside Orca or in Zed.

### Expected

A new Orca terminal should use the latest persisted Windows PATH ordering and resolve:

`C:\Users\<user>\AppData\Local\Programs\Python\Python314\python.exe`

Existing tabs may retain their old environment, but new tabs should not require restarting Orca.

### Actual

The Orca tab resolves:

`C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\python.exe`

That alias reports that Python was not found (and can appear to do nothing). A fresh external pwsh resolves Python 3.14.6 successfully.

### Root-cause evidence

The same pwsh executable behaves differently when only PATH ordering changes:

```
Fresh persisted order:
...\Python314\;...\WindowsApps
-> real Python 3.14.6

Current Orca merge order:
...\WindowsApps;...\Python314\
-> broken WindowsApps alias
```

`src/main/pty/windows-environment-path.ts` reads current machine/user registry PATH values, but `mergeWindowsPathSegments` only appends missing persisted segments to the inherited process PATH. This leaves an existing `WindowsApps` entry ahead of newly installed Python directories even though the current persisted user PATH places Python first. Windows executable resolution is first-match-wins.

### Suggested acceptance criteria

- New local Windows terminals reflect current persisted machine/user PATH ordering.
- Orca-specific injected PATH entries remain available.
- A stale `WindowsApps` entry cannot shadow a newly installed Python when persisted PATH orders Python first.
- Add a regression test for inherited `WindowsApps` plus persisted Python directories inserted before it.

ShreyanshVaibhaw:
## Summary

Fixes #11992.

On Windows, a terminal opened after installing a tool could still resolve the old executable. Installing Python while Orca was running left new pwsh tabs resolving `python` to the `WindowsApps` Store alias, which reports that Python is not installed, while a pwsh launched outside Orca resolved the real interpreter.

`mergeWindowsPathSegments` read the persisted machine/user PATH from the registry but only **appended** the segments that were missing from Orca's inherited PATH. Windows executable resolution is first-match-wins, so an inherited `WindowsApps` entry stayed ahead of a `Python314\` directory that the persisted PATH now orders before it.

This rebuilds the merged PATH from the persisted values instead of appending to them:

- Persisted segments keep their registry order.
- Segments the registry does not know about (Orca-injected, or added by the launching shell) are preserved in inherited order and placed ahead of the persisted list, so Orca's own tools still win.
- An empty, blocked, or timed-out registry read leaves PATH untouched, so a denied `reg.exe` query can never rewrite a working PATH.
- Segments differing only by a trailing separator no longer survive as duplicates. Real registries do contain these.

### Deliberate behavior change, please review

The previous append-only behavior was intentional. The comment on `mergePersistedWindowsPath` documented it as appending "without unexpectedly reordering existing PATH", and `mergePersistedWindowsPath` had a test named *"appends missing persisted segments without reordering the inherited PATH"* asserting exactly that.

This PR reverses that decision and rewrites that test, because the two goals are mutually exclusive: you cannot both preserve the inherited order and stop a stale entry from shadowing a newer one. The issue's acceptance criteria ask for persisted ordering to win. Flagging it explicitly so the tradeoff is a decision rather than an accident.

## Screenshots

No visual change.

## Testing

- [x] `pnpm typecheck`
- [ ] `pnpm test` (full suite did not complete locally, see Notes)

`src/main/pty/windows-environment-path.test.ts` goes from 12 to 16 tests:

- `stops an inherited WindowsApps alias from shadowing a newly installed Python` is the regression test the issue asks for.
- `keeps Orca-injected entries the registry does not know about ahead of persisted ones` pins the constraint that Orca's own PATH entries survive.
- `leaves the inherited PATH untouched when both registry reads are blocked` covers the denied-registry path.
- `does not duplicate a segment that differs only by a trailing separator` covers the dedup fix.
- `adopts persisted machine and user PATH ordering` replaces the test that asserted the old append-only contract.

**These tests were verified to fail against the previous implementation.** Reverting the source and re-running produces 4 failures, including `expected 3 to be less than 2` for the Python/WindowsApps ordering, so they genuinely catch the regression rather than merely passing alongside the fix.

Also verified end to end on Windows 11 against the live registry, using a temporary test that ran `mergePersistedWindowsPath` over the real `process.env` and real `reg.exe` output. It confirmed the resulting PATH preserves the deduplicated persisted ordering on real data. That scratch file was not committed.

Suite results with the change applied:

- `src/main/pty/windows-environment-path.test.ts`: 16 passed
- `src/main/pty/` plus `src/main/ipc/preflight`: 141 passed, 17 skipped, 0 failed
- `oxlint` on both changed files: clean
- max-lines ratchet: OK, no new bypasses

## AI Review Report

- **Cross-platform compatibility (macOS, Linux, Windows).** Both public entry points return early when `platform !== 'win32'`, so behavior on macOS and Linux is unchanged. No shortcuts, labels, or UI strings are touched. Path handling stays string-level on PATH segments and does not introduce separator assumptions beyond the existing `getPathDelimiter` helper. No Electron-specific APIs are touched.
- **SSH, remote, and local compatibility.** The change is confined to how a local Windows PATH is composed before PTY spawn. It reads no new process, file, credential, or network resource, and adds no assumption that anything exists only on the local machine.
- **Callers.** All other call sites (`preflight-local-env.ts`, `preflight.test.ts`, `preflight-agent-detection-no-subprocess.test.ts`) either mock the function or do not assert ordering. `preflight-windows-path-refresh.repro.test.ts` exercises the real implementation and still passes, since it only asserts that a newly persisted directory becomes resolvable.
- **Flagged and addressed.** The initial review caught that a naive reorder could drop Orca-injected entries, and that an empty persisted read would blank PATH. Both are now covered by explicit guards and tests. It also flagged that a bare drive root must not be normalized, since `C:\` and `C:` differ on Windows; `normalizeSegmentKey` guards that case.

## Security Audit

- **Command execution.** No new subprocess is spawned. The existing bounded `reg.exe query` calls are unchanged, including their 5s timeout and `windowsHide`.
- **Input handling.** Registry values were already parsed and expanded by existing helpers. The new code only reorders and deduplicates the resulting strings, and performs no shell interpolation.
- **Path handling.** `normalizeSegmentKey` is used strictly as a comparison key. Emitted segments keep their original spelling from the registry or the inherited PATH, so no path is rewritten or silently canonicalized.
- **PATH shadowing.** This is the security-relevant dimension. The change makes the persisted machine and user PATH authoritative for ordering rather than a stale in-process copy, which narrows rather than widens the window in which an outdated entry can shadow a newer executable. Non-persisted entries retain priority, which preserves existing Orca behavior. Worth a maintainer's eye, since PATH precedence is the kind of thing that deserves a second opinion.
- **Secrets, auth, IPC, dependencies.** Untouched. No dependency changes.

## Notes

- **Pre-existing lint failure, unrelated to this PR.** `pnpm lint` fails at `verify:skill-bundle-manifest` with stale `resources/skills/*.json` artifacts. This reproduces on a clean checkout of `main` with no changes applied, so it is not caused by this PR. I deliberately did not regenerate those artifacts here, since they are generated files and would add unrelated churn to a focused fix. Happy to open a separate issue if useful. Every other lint stage passes, including the type-aware pass, reliability gates, and the max-lines ratchet.
- **`pnpm test` and `pnpm build` were not completed locally.** The full suite ran for over 35 minutes without finishing on this machine and was not a clean measurement, so I am not claiming it passed. The directly affected suites pass, as listed above. CI coverage is welcome here.
- **Platform-specific behavior.** Windows only. macOS and Linux short-circuit before any of the changed code runs.
- **Cache interaction.** The 30s persisted-PATH cache and its last-good fallback are unchanged. The new early return only triggers when the resolved persisted segment list is genuinely empty, which the existing `keepLastGoodSegments` logic already distinguishes from a blocked read.

### DIFF
--- a/src/main/pty/windows-environment-path.test.ts
+++ b/src/main/pty/windows-environment-path.test.ts
@@ -255,7 +255,7 @@ describe('mergePersistedWindowsPath', () => {
     expect(env.Path).toBe('C:\\Users\\orca\\AppData\\Local\\agy\\bin;C:\\Windows')
   })
 
-  it('appends missing persisted segments without reordering the inherited PATH', () => {
+  it('adopts persisted machine and user PATH ordering', () => {
     const execFileSync = vi
       .fn()
       .mockReturnValueOnce(
@@ -278,8 +278,10 @@ describe('mergePersistedWindowsPath', () => {
 
     mergePersistedWindowsPath(env, { platform: 'win32', execFileSync })
 
+    // Why: every inherited entry is also persisted here, so the persisted order wins
+    // outright and the duplicate `C:\Existing` collapses to its first position.
     expect(env.Path).toBe(
-      'C:\\Existing;C:\\Windows\\System32;C:\\Users\\me\\AppData\\Local\\Orca\\bin'
+      'C:\\Windows\\System32;C:\\Existing;C:\\Users\\me\\AppData\\Local\\Orca\\bin'
     )
   })
 
@@ -326,6 +328,81 @@ describe('mergePersistedWindowsPath', () => {
 
     expect(env).toEqual({ Path: 'C:\\Root\\Live;C:\\Machine;C:\\User' })
   })
+
+  it('stops an inherited WindowsApps alias from shadowing a newly installed Python', () => {
+    // Regression: Python installed while Orca was running. Orca's inherited PATH still
+    // lists the WindowsApps stub, and appending the new Python directories behind it
+    // left `python` resolving to the alias, which reports that Python is not installed.
+    const python = 'C:\\Users\\me\\AppData\\Local\\Programs\\Python\\Python314\\'
+    const pythonScripts = 'C:\\Users\\me\\AppData\\Local\\Programs\\Python\\Python314\\Scripts\\'
+    const windowsApps = 'C:\\Users\\me\\AppData\\Local\\Microsoft\\WindowsApps'
+    const execFileSync = vi
+      .fn()
+      .mockReturnValueOnce('    Path    REG_EXPAND_SZ    C:\\Windows\\system32;C:\\Windows\r\n')
+      .mockReturnValueOnce(
+        `    Path    REG_EXPAND_SZ    ${python};${pythonScripts};${windowsApps}\r\n`
+      )
+    const env = { Path: `C:\\Windows\\system32;C:\\Windows;${windowsApps}` }
+
+    mergePersistedWindowsPath(env, { platform: 'win32', execFileSync })
+
+    const segments = env.Path.split(';')
+    expect(segments.indexOf(python)).toBeGreaterThanOrEqual(0)
+    expect(segments.indexOf(python)).toBeLessThan(segments.indexOf(windowsApps))
+    expect(env.Path).toBe(
+      `C:\\Windows\\system32;C:\\Windows;${python};${pythonScripts};${windowsApps}`
+    )
+  })
+
+  it('keeps Orca-injected entries the registry does not know about ahead of persisted ones', () => {
+    const orcaBin = 'C:\\Users\\me\\AppData\\Local\\Orca\\resources\\bin'
+    const python = 'C:\\Users\\me\\AppData\\Local\\Programs\\Python\\Python314\\'
+    const windowsApps = 'C:\\Users\\me\\AppData\\Local\\Microsoft\\WindowsApps'
+    const execFileSync = vi
+      .fn()
+      .mockReturnValueOnce('    Path    REG_EXPAND_SZ    C:\\Windows\\system32\r\n')
+      .mockReturnValueOnce(`    Path    REG_EXPAND_SZ    ${python};${windowsApps}\r\n`)
+    const env = { Path: `${orcaBin};C:\\Windows\\system32;${windowsApps}` }
+
+    mergePersistedWindowsPath(env, { platform: 'win32', execFileSync })
+
+    expect(env.Path).toBe(`${orcaBin};C:\\Windows\\system32;${python};${windowsApps}`)
+  })
+
+  it('leaves the inherited PATH untouched when both registry reads are blocked', () => {
+    const execFileSync = vi.fn().mockImplementation(() => {
+      throw new Error('ERROR: Access is denied.')
+    })
+    const env = { Path: 'C:\\Windows\\system32;C:\\Existing' }
+
+    mergePersistedWindowsPath(env, { platform: 'win32', execFileSync })
+
+    expect(env.Path).toBe('C:\\Windows\\system32;C:\\Existing')
+  })
+
+  it('does not duplicate a segment that differs only by a trailing separator', () => {
+    const execFileSync = vi
+      .fn()
+      .mockReturnValueOnce('    Path    REG_SZ    C:\\Tools\r\n')
+      .mockReturnValueOnce('    Path    REG_SZ    \r\n')
+    const env = { Path: 'C:\\Tools\\' }
+
+    mergePersistedWindowsPath(env, { platform: 'win32', execFileSync })
+
+    expect(env.Path.split(';')).toEqual(['C:\\Tools'])
+  })
+
+  it('canonicalizes slash style and repetition for drive-root comparison', () => {
+    const execFileSync = vi
+      .fn()
+      .mockReturnValueOnce('    Path    REG_SZ    C:\\\\\r\n')
+      .mockReturnValueOnce('    Path    REG_SZ    \r\n')
+    const env = { Path: 'C:/' }
+
+    mergePersistedWindowsPath(env, { platform: 'win32', execFileSync })
+
+    expect(env.Path.split(';')).toEqual(['C:\\\\'])
+  })
 })
 
 describe('resolvePathEnvKey', () => {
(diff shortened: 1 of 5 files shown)


ShreyanshVaibhaw:
### Windows local `pnpm test` baseline

Context for the unticked `pnpm test` box in the description. The full suite does not currently pass on a Windows dev machine **before** this PR, so I measured the baseline rather than claiming a green run.

Environment: Windows 11, Node 24.12.0, pnpm 10.30.3, local (non-CI). Both runs below were taken on a pristine tree, in sequence, on the same machine.

| Run | Tests | Test files |
|---|---|---|
| Clean `main` at `4be4d10ae` | 167 failed, 42573 passed, 516 skipped (43295) | 57 failed, 3988 passed, 28 skipped (4073) |
| This branch | 173 failed, 42571 passed, 516 skipped (43299) | 60 failed, 3985 passed, 28 skipped (4073) |

The branch total is 4 tests larger because this PR adds 4 tests.

#### The +6 / +3 delta is flake, not regression

Three test files fail on the branch run that did not fail on the `main` run. No file fails on `main` that passes on the branch. I investigated each of the three individually rather than assuming:

| File | Finding |
|---|---|
| `src/main/native-chat/transcript-watch-liveness.test.ts` | **Flaky on unmodified `main`.** Run three times in isolation on a clean `main` checkout: passed, failed, passed. |
| `src/renderer/.../project-view-wrapper-source-context-boundary.test.ts` | **Fails on `main` too.** Reproduces in isolation on clean `main` with a 30s timeout (`30015ms`). |
| `src/main/codex-usage/scanner-paths.test.ts` | Passes in isolation on this branch. Only fails under full-suite parallel load. |

None of the three touch PATH construction, PTY spawn, or the registry. The change is confined to `mergeWindowsPathSegments`, guarded by `platform !== 'win32'` early returns in both public entry points, so macOS and Linux never reach the modified code.

#### Why the pre-existing failures matter for review

The failures on `main` cluster in exactly the area this PR touches:

- `src/main/ipc/pty.test.ts`
- `src/main/daemon/pty-subprocess.test.ts`
- `src/relay/pty-handler.test.ts`

All three fail on unmodified `main`. The remainder are broadly environmental for a local Windows box: `src/main/ssh/*`, `src/main/startup/*`, `src/main/ipc/ephemeral-vm.test.ts`, `src/relay/*`, plus tests that assume a POSIX shell. Two representative errors:

```
Error: spawn /bin/sh ENOENT
  spawnargs: [ '-c', '"C:\Program Files\nodejs\node.exe" ... ' ]

Error: Timed out waiting for PTY process exit: pty-1
  at src/relay/pty-handler.ts:1747
```

`src/main/pty/windows-environment-path.test.ts` does not appear in either failure list.

#### Targeted results for the code this PR changes

- `src/main/pty/windows-environment-path.test.ts`: **16 passed** (up from 12)
- `src/main/pty/` plus `src/main/ipc/preflight`: **141 passed, 17 skipped, 0 failed**
- `oxlint` on both changed files: clean
- `pnpm typecheck`: passes
- max-lines ratchet: OK, no new bypasses

The new tests were confirmed to **fail against the previous implementation**, so they catch the regression rather than merely passing alongside the fix. Also verified end to end on Windows 11 against the live registry and real `process.env`, using a scratch test that was not committed.

#### One pre-existing lint failure

`pnpm lint` fails at `verify:skill-bundle-manifest` with stale `resources/skills/*.json`. This also reproduces on a clean checkout of `main`, so it is not introduced here. I left those generated artifacts alone to keep this diff focused. Happy to open a separate issue if useful.

I am not presenting these as a passing run. They are here so the unticked checkbox is not mistaken for an untested change, and so the Windows-local baseline is on the record. CI remains the authority.

Jinwoo-H:
Thank you for the careful Windows PATH work and the detailed baseline investigation. We ran a fresh native Windows comparison of this PR against #13545 on current main. Both changes fix ordinary PATH ordering, but this implementation can still merge a registry snapshot after that snapshot has been invalidated: the controlled race returned the stale `OldMachine;OldUser` ordering. The #13545 implementation joins or retries the current cache generation and returned `NewMachine;NewUser` in the same scenario. Its focused matrix passed 29/29, and native `Get-Command`, `where.exe`, and executable launch all selected the newly ordered executable. We found no unique correction in #12050 that is not covered by #13545, so keeping both would retain the stale-snapshot race and duplicate the same ownership boundary. I am closing this as fully superseded, without merging it. Reopening would make sense if a concrete Windows PATH case is shown that #13545 does not cover, or if #13545 no longer proceeds. WSL-only validation was intentionally held. Thanks again for contributing this and for documenting the Windows baseline so thoroughly.

PR #12052, opened 2026-08-02
fix(ime): stop terminal shortcuts firing on keystrokes the IME owns

nwparker:
## Summary

The window-capture keydown handler resolved and dispatched terminal chords with no check for a live composition, so a keystroke meant for the candidate window drove a terminal action instead.

This is a pre-existing defect on `main`, not something this stack introduced — no commit in the stack touches `keyboard-handlers.ts`. It sits at the top of the stack because it is the same rule the rest of the stack enforces, stated in AGENTS.md: *guard above the key dispatch, not inside each key branch.*

## Why it reaches the user

`onKeyDown` is registered at `window` with `{capture: true}`, so it runs **before** the pane's `attachCustomKeyEventHandler`, and on a match it calls `stopImmediatePropagation()`. The pane handler never runs. Whatever this function resolves is the only thing the keystroke does.

The only composition awareness in the entire function lived inside the `sendInput` branch and covered `Enter` alone — the one key that was already handled correctly.

**Leak 1 — resolved chords go straight to the PTY.** Preedit open, user presses Ctrl+Backspace meaning "erase what I'm composing":

| chord | sent to PTY | effect |
| --- | ---: | --- |
| Ctrl+Backspace | `\x17` | erases a word of the **real** command line behind the candidate window |
| Alt+Backspace | `\x1b\x7f` | same |
| Alt+←/→ | `\x1bb` / `\x1bf` | moves the shell cursor mid-composition |
| Ctrl+←/→ (non-Mac) | word-motion | same |
| Cmd+Backspace / Cmd+Delete / Cmd+←→ (Mac) | `\x15` / `\x0b` / `\x01` / `\x05` | same |

The composition is untouched in every row — the damage lands on the line the user cannot currently see.

**Leak 2 — `Process` still matches by physical key.** `shouldUseNonLatinShortcutPhysicalFallback` returns true for `key: 'Process'` on non-darwin, because `normalizeKeyToken` returns null for it and the fallback then matches on `event.code`. So an IME-consumed Ctrl-chord still resolves against the Linux/Windows defaults: Ctrl+K clears the terminal, Ctrl+F opens search, and **Ctrl+W closes the pane** — out from under a live composition. This one is not fixed by any Enter-shaped change, which is why it is called out separately.

`terminal-shortcut-policy.ts` has no composition, `Process`, or 229 filter anywhere in it, by design — resolution is meant to be pure. The guard belongs at the dispatch, which is where this PR puts it.

## The disagreement this resolves

`xterm-bypass-policy.ts` declares Backspace, Delete and the arrows IME-owned while composing (`TERMINAL_IME_OWNED_KEYS`). The window-capture handler decided the opposite for the same physical keystroke, and because it runs first and stops propagation, it won. Two files, one keystroke, opposite answers.

## The fix

One guard, above the dispatch, in a new `terminal-shortcut-composition-guard.ts`:

```ts
if (terminalShortcutIsOwnedByIme(e, resolveShortcutEvent)) {
  return
}
```

It sits above `matchFileSearchShortcut` deliberately — that matcher calls `keybindingMatchesAction`, so leak 2 reaches it too. It sits below the Windows Enter-redispatch absorber, which handles the committing Enter (`!isComposing`) and must still run.

Two exemptions, both paths that already handle a live composition, and both of which would **break** if claimed:

- **Enter** — the `sendInput` branch defers it via `deferredNewlineSender` until the commit lands. Swallowing it here would drop the newline entirely.
- **Input-source switch** — it only arms the native-only tracker and hands the key back to the OS. Returning early instead would let xterm see the chord and write its control byte to the PTY, which is exactly the failure this guard exists to prevent.

Resolution is pure, so asking for the action before deciding costs nothing.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — full `terminal-pane` suite: **220 files / 2,919 tests**, no regressions

Four new integration assertions, each of which **fails on `main` with the exact predicted symptom** and passes after:

| test | before |
| --- | --- |
| Ctrl+Backspace while composing | `sendInput` called |
| Alt+Backspace while composing | `sendInput` called |
| Alt+ArrowLeft while composing | `sendInput` called |
| Ctrl+W reported as `Process` | `onRequestClosePane` called with pane 1 |

Paired negative, passing before **and** after, proving the guard does not over-correct: **Ctrl+Backspace with nothing composing still reaches the PTY.**

Plus unit coverage for the guard itself, including both exemptions — the two cases the integration tests cannot reach, and the two where a wrong answer silently breaks a working path rather than merely failing to fix a broken one.

## Screenshots

No visual change.

## AI Review Report

- Cross-platform: no hardcoded `metaKey`. The macOS Cmd chords and the non-Mac Ctrl chords are both listed above and both covered by the same guard. The `Process` marker is the Windows/Android spelling and is handled by the shared predicate rather than re-derived here.
- The guard reuses `isImeCompositionKeyDown`, which already existed and whose doc comment already said *"Call this above the key dispatch, not inside each key branch"* — the predicate was correct, it simply was not called from the terminal's own dispatcher. It is called from ~25 text inputs across the app and, before this PR, from neither terminal key dispatcher.
- SSH and folder workspaces: keyboard-only change, no path or transport assumptions.
- Risk checked and accepted: the guard runs on every terminal keydown. It short-circuits on `isImeCompositionKeyDown` returning false, which is three field reads for the overwhelmingly common non-IME case, before any resolution happens.

## Security Audit

No auth, secrets, IPC, or filesystem surface. The change strictly *reduces* what an IME-marked keystroke can trigger.

## Notes

Base: #12037.

A second `attachCustomKeyEventHandler` — the dashboard preview terminal — has no composition policy at all and writes to a live PTY. It is a different surface and is not in this PR.

### DIFF
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
@@ -31,6 +31,7 @@ function keyboardEvent(
 function createHarness(): {
   deps: KeyboardHandlersDeps
   editable: HTMLInputElement
+  sendInput: ReturnType<typeof vi.fn>
   startComposition: () => void
   terminalInput: HTMLTextAreaElement
   dispose: () => void
@@ -97,6 +98,7 @@ function createHarness(): {
   return {
     deps,
     editable,
+    sendInput,
     terminalInput,
     startComposition: () => {
       terminalElement.dispatchEvent(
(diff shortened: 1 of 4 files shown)


nwparker:

PR #12053, opened 2026-08-02
fix(ime): give the dashboard preview terminal a composition policy

nwparker:
## Summary

The dashboard preview terminal is the app's **second** `attachCustomKeyEventHandler`, and it was the one with no IME policy at all. It resolves the full pane shortcut policy and writes the result to a live pty — `AgentTerminalPreview.tsx` wires `sendInput: (data) => terminal?.input(data)`, so this is not a read-only view.

Pre-existing on `main`. Follows #12052, which fixes the same class of defect on the pane's window-capture handler and adds the guard this PR reuses.

## What was missing

The handler ran `claimImeKeyEvent` — the **macOS-only** native-text bridge — then clipboard chords, then `resolvePreviewShortcutAction` → `sendInput`. No composition tracker, no bypass policy, no deferred-newline path. On Linux and Windows the surface had no IME handling whatsoever.

| keystroke mid-preedit | before | why it hurts |
| --- | --- | --- |
| Ctrl+Backspace | `\x17` to the pty | erases a word of the agent's real command line, preedit untouched |
| Shift+Enter | `\x1b\r` to the pty | the newline lands *before* the text it was meant to follow |

The Shift+Enter case is the sharper one: on the main pane that exact keystroke is deferred by `deferredNewlineSender`. Same chord, same composition, two different outcomes depending on which window it was typed into.

## Enter is claimed here, not exempted

This is the one place this PR deliberately differs from #12052.

The pane exempts Enter because it *defers* it — the newline is replayed once the commit lands. This surface has no defer path, so exempting Enter would just reproduce the bug. Claiming it means the keystroke closes the composition instead, which is what a user pressing Enter over a candidate window is asking for anyway.

That made the exemption a property of the **caller**, not of the key, so it became an argument rather than an assumption baked into the guard:

```ts
// pane — has deferredNewlineSender
terminalShortcutIsOwnedByIme(e, resolveShortcutEvent, { enterIsDeferredToCommit: true })

// preview — does not
terminalShortcutIsOwnedByIme(event, resolveAction)
```

The guard returns `true` from the handler rather than consuming the event, so xterm's own composition handling stays intact. Returning `false` would have suppressed the chord *and* bypassed `CompositionHelper`, which is the opposite of what is wanted here.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — `dashboard-popout` + `terminal-pane`: **230 files / 3,001 tests**, no regressions

Two defect assertions and two paired negatives, all mutation-checked by disabling the guard and re-running:

| test | guard disabled | guard enabled |
| --- | --- | --- |
| Ctrl+Backspace while composing | **fails** | passes |
| Shift+Enter while composing | **fails** | passes |
| Ctrl+Backspace, nothing composing | passes | passes |
| Shift+Enter, nothing composing | passes | passes |

Worth recording: the first draft of the Shift+Enter test used `key: 'Process'` and **passed with the guard disabled** — with that key the chord never resolves to an action, so the test proved nothing. The mutation check is what caught it. It now uses the shape the failure actually has on macOS (`key: 'Enter'`, `isComposing: true`), and fails without the fix.

Guard-level unit coverage for both Enter directions lives in `terminal-shortcut-composition-guard.test.ts`.

## Screenshots

No visual change.

## AI Review Report

- Cross-platform: the defect was *worst* off macOS, since the only pre-existing IME wiring here (`installPreviewImeBridge`) returns null unless `getShortcutPlatform() === 'darwin'`. The guard is platform-independent and keys off the shared marker predicate.
- The duplicated `resolvePreviewShortcutAction(event, {...})` call is now a single `resolveAction` closure used by both the guard and the switch, so the guard cannot drift from the action the handler goes on to run.
- SSH and folder workspaces: keyboard-only, no path or transport assumptions.
- Risk checked and accepted: the preview's Enter behavior changes during composition only. With nothing composing, Shift+Enter still sends `\x1b\r` — covered by a negative test.

## Security Audit

No auth, secrets, IPC, or filesystem surface. Strictly reduces what an IME-marked keystroke can send to a live pty.

## Notes

Base: #12052.

### DIFF
--- a/src/renderer/src/components/dashboard-popout/preview-terminal-key-handler.test.ts
+++ b/src/renderer/src/components/dashboard-popout/preview-terminal-key-handler.test.ts
@@ -1,0 +1,105 @@
+// @vitest-environment happy-dom
+
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import type { Terminal } from '@xterm/xterm'
+import type { DashboardCardTerminalInput } from '../../../../shared/dashboard-snapshot'
+import { installPreviewTerminalKeyHandler } from './preview-terminal-key-handler'
+
+const LOCAL_MAC: DashboardCardTerminalInput = {
+  hostPlatform: 'darwin',
+  localWindowsConpty: false,
+  windowsShiftEnterEncoding: 'alt-enter',
+  kittyKeyboardAdvertised: true
+}
+
+function installHandler(): {
+  handle: (event: KeyboardEvent) => boolean
+  sendInput: ReturnType<typeof vi.fn>
+  dispose: () => void
+} {
+  let handle: (event: KeyboardEvent) => boolean = () => true
+  const sendInput = vi.fn()
+  const terminal = {
+    attachCustomKeyEventHandler: (handler: (event: KeyboardEvent) => boolean) => {
+      handle = handler
+    },
+    element: document.createElement('div'),
+    getSelection: () => '',
+    scrollToTop: vi.fn(),
+    scrollToBottom: vi.fn()
+  } as unknown as Terminal
+  const dispose = installPreviewTerminalKeyHandler({
+    terminal,
+    claimImeKeyEvent: () => false,
+    pasteClipboardText: vi.fn(),
+    sendInput,
+    getShortcutContext: () => ({
+      clientPlatform: 'darwin',
+      macOptionAsAlt: 'false',
+      keybindings: undefined,
+      terminalInput: LOCAL_MAC,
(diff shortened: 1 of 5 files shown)


nwparker:
Superseded by #13128 (17b3dff3c41). `terminal-shortcut-composition-guard.ts` was deleted with the rest of the first-party composition layer; the preview terminal no longer needs a separate composition policy because it no longer has one to diverge from.

PR #12055, opened 2026-08-02
Separate the terminal IME platform checks

nwparker:
## Summary

The terminal's IME code asks two different platform questions and was answering them with one flag:

1. **May a standalone keydown 229 reach xterm's `CompositionHelper`?** Windows must suppress it, because of its preedit-diff race. Everywhere else must pass it.
2. **Is an ibus/fcitx-style candidate window in play?** The candidate-key guards in `xterm-bypass-policy.ts` read behavior specific to those input methods, so they belong to desktop Linux only.

Both read one `isLinux`, derived inline in `use-terminal-pane-lifecycle.ts`. Tightening it for question 2 moved question 1 too, and moved it the wrong way.

## The derivation was also wrong about ChromeOS

The inline comment claimed:

```ts
// Why: Android/ChromeOS UAs also contain "Linux"; scope the fcitx candidate-key policy to desktop Linux.
const isLinux =
  !isMac && navigator.userAgent.includes('Linux') && !/Android|CrOS/.test(navigator.userAgent)
```

True of Android (`Linux; Android 14; Pixel 8`), **not** of ChromeOS, which reports `X11; CrOS x86_64 14541.0.0` — no `Linux` substring anywhere. So `includes('Linux')` was already false and the `CrOS` half of the exclusion was unreachable. ChromeOS landed in the Windows bucket for *both* policies: right answer for the candidate guards, wrong for 229.

Meanwhile `getShortcutPlatform()` maps that same user agent to `'linux'`. Two mappings, two answers.

## Change

The fix is not a better substring test — it's not having a second mapping. `shortcut-platform.ts` gains a pure `resolveShortcutPlatform(userAgent)`; `getShortcutPlatform()` becomes its `navigator.userAgent` caller. The new `terminal-ime-platform.ts` derives from it:

```ts
export function resolveTerminalImePlatform(userAgent: string): TerminalImePlatform {
  const platform = resolveShortcutPlatform(userAgent)
  const isLinux = platform === 'linux'
  return {
    isMac: platform === 'darwin',
    isLinux,
    isDesktopLinux: isLinux && !NON_DESKTOP_LINUX_PLATFORM.test(userAgent)
  }
}
```

`isMac`/`isLinux` now cannot disagree with the rest of the renderer. `isDesktopLinux` narrows from there, and that narrowing is the only intended difference.

This also drops the enumeration problem rather than re-solving it. Question 1 is really "is this Windows?", so an unrecognised desktop passes 229 instead of landing in whichever bucket a UA guess picks — which is how ChromeOS got here in the first place.

`XtermImeKeyboardOptions` gains an optional `isDesktopLinux` (defaulting to `isLinux`, so existing callers are unaffected); `suppressOrphanCandidateDigit`, `suppressCandidateKey`, and `shouldPreventDefaultTerminalImeCandidateKey` switch to it. `passesStandalone229Keydown = isMac || isLinux` is untouched and now correctly includes ChromeOS.

Extracting the derivation is what made this testable at all — the bug lived in a string expression inline in a 1000-line effect.

## Tests

`terminal-ime-platform.test.ts` (12) pins real user agents for desktop Linux X11, ChromeOS, Android, macOS, Windows, and an unrecognised X11 desktop, plus a case asserting `isMac`/`isLinux` agree with `resolveShortcutPlatform` for every one of them — so the two mappings can't drift back apart silently.

Two cases added to `xterm-bypass-policy-non-mac.test.ts`: ChromeOS 229 keydowns reach xterm as on desktop Linux, and the candidate-key guards do not fire for it.

Mutation-checked at the intermediate step: with a plain `/Linux/` test the ChromeOS case fails (`isLinux: false`) while Android still passes — the asymmetry the old comment had backwards.

terminal-pane + lib suites: 589 files, 6452 tests, green. Typecheck green.

## Confidence

**Latent inconsistency, not an observed user-facing bug.** Orca does not ship Android, and Electron on a Chromebook runs inside the Linux container, which reports `X11; Linux x86_64` and was always on the correct path. I have no ChromeOS or Gboard event trace; the dropped-character consequence is reasoned from the code path, not observed.

What is verifiable without a device is the rest: one flag answered two questions, its `CrOS` exclusion was dead code contradicted by the comment above it, and it disagreed with `getShortcutPlatform()` on the same input.

## Also: one comment on the adjacent Linux candidate policy

Reviewing `terminal-ime-linux-candidate-state.ts` alongside this, the window-blur reset looks like a bug and is not one. Clearing `pressedCodes` on blur means a key held across the blur and released on return reads as an orphaned keyup and arms the candidate-digit guard, swallowing one digit.

The apparent fix — treat post-blur keyups as unknowable rather than orphaned — costs more than it saves. The genuine orphan signal it exists to catch is a keyup whose keydown the IME consumed, and the first composition after focusing the window is exactly when that happens: click into the terminal, type, pick candidate `1`. Ignoring post-blur keyups would miss that common case to fix a rare one (holding a letter across a focus change).

No logic change — the existing test already pins the behavior. Added the reasoning, because the test asserts the outcome without stating why, and the wrong conclusion is the easy one to reach.

## Also: one line in the IME rules

While writing #12053 I built a composing-Shift+Enter fixture from `key: 'Process'`. It passed with the guard deleted — a `Process` chord never resolves to an action, so nothing was ever sent regardless, and the test certified a guard it had never exercised. The shape with teeth is the marked-but-real key (`key: 'Enter'`, `keyCode: 13`, `isComposing: true`), which is also what the recorded hangul trace carries.

Added as a rule under `AGENTS.md` → IME Composition, next to the existing fixture requirement. The section is the stated home for these; the rules PR earlier in this stack is where it would otherwise go, but putting it here avoids a 13-branch cascade for one line.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,6 +60,7 @@ These rules bind any change to keyboard handling, the chat composer, the termina
 - `attachCustomKeyEventHandler` returning `false` does **not** call `preventDefault()`, and it bypasses xterm's `CompositionHelper` entirely. Pair the two deliberately.
 - Never normalize IME output at commit. Normalize only at path handling and equality/search comparison.
 - Add a recorded event-trace fixture for the trace that motivated the change, plus a paired negative test proving non-IME behavior is unchanged.
+- Test a shortcut guard with the marked-but-real key shape (`key: 'Enter'`, `keyCode: 13`, `isComposing: true`), not `key: 'Process'`. A `Process` chord never resolves to an action, so the test passes with the guard deleted. Delete the guard and watch it fail before trusting it.
 
 ## Git Provider Compatibility
 
--- a/src/renderer/src/components/terminal-pane/terminal-ime-linux-candidate-state.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-linux-candidate-state.ts
@@ -55,6 +55,10 @@ function acquirePhysicalKeyTracker(
       pressedCodes.delete(code)
     }
   }
+  // A key held across the blur then released on return reads as orphaned and arms the
+  // guard for one digit. Deliberate: not clearing would strand the code as pressed, and
+  // treating post-blur keyups as unknowable would miss the orphan on the first commit
+  // after focus, which is the common case.
   const reset = (): void => pressedCodes.clear()
   // Why: bubble-phase keyup cleanup runs after xterm's target handler, so the
   // pane can still classify that release against the shared pressed-key set.
--- a/src/renderer/src/components/terminal-pane/terminal-ime-platform.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-platform.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it } from 'vitest'
+import { resolveTerminalImePlatform } from './terminal-ime-platform'
+
+const DESKTOP_LINUX =
+  'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/140.0.0.0 Safari/537.36'
+const CHROME_OS =
+  'Mozilla/5.0 (X11; CrOS x86_64 14541.0.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/140.0.0.0 Safari/537.36'
+const ANDROID =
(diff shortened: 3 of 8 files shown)


nwparker:
Superseded by #13128 (17b3dff3c41). The platform-split files were deleted along with the composition layer they partitioned.

PR #12058, opened 2026-08-02
feat(repo-icon): expand the project emoji picker

### LINKED ISSUE #7989: [Feature]: Allow free selection from all system-provided emojis
### Problem or use case

Currently, only a few preset emojis are available to choose from. It would be great if users could freely pick any emoji provided by the system, rather than being limited to a small predefined set.

### Proposed solution

Proposed improvement
Let users freely select from the full set of system-provided emojis instead of only the limited presets.

_No response_

chucoding:
## Summary

Closes #7989.

Replace the hardcoded 12-emoji grid in the repo icon settings ("Emoji" tab of `RepositoryIconTabs.tsx`) with the full, native, searchable, category-navigable emoji picker, letting users freely select from the full set of system-provided emojis instead of only the limited presets. No new dependency: it reuses `emoji-picker-react`, which is already a dependency and already used by `RichMarkdownEmojiMenu` in the markdown editor.

**Design choice**: compared reusing `RichMarkdownEmojiMenu`'s `emoji-picker-react` setup (search, categories, lazy-loading built in) against hand-building a shadcn `Popover + Command` picker over the full emojibase dataset. Went with the former — it's the minimal design that matches STYLEGUIDE's "sibling components should read as one design" guidance, and a `Command`-based rebuild would need its own virtualization (cmdk has none) plus a from-scratch category-navigation implementation, which is unnecessary scope and a real performance risk for a multi-thousand-emoji catalog.

Other changes in this PR:
- The picker itself lives in a new `RepositoryIconEmojiPicker` component, lazy-loaded (`React.lazy` + `Suspense`) from `RepositoryIconTabs`. `emoji-picker-react` is a ~500KB chunk; without this split it was pulled into the eager settings bundle every time the repo icon settings render, not just when the Emoji tab is opened. This also keeps `RepositoryIconTabs.tsx` itself focused on tab wiring instead of mixing in picker-specific state.
- Added a `.repo-icon-emoji-picker`-scoped CSS variable block in `main.css` using `--card` / `--card-foreground` tokens (this picker is embedded inline in a settings panel, not a popover, so it doesn't reuse `RichMarkdownEmojiMenu`'s `--popover` tokens) and the `--font-sans` token instead of a hardcoded font-family.
- Every pick is re-validated through `sanitizeRepoIcon` before being saved, so a ZWJ/skin-tone sequence over its 16-char cap surfaces a `toast.error` instead of silently no-op'ing through `sanitizeRepoUpdate`. Skin tones are left selectable (not disabled) for global inclusivity.
- `Theme.DARK`/`Theme.LIGHT` is computed explicitly from a minimal `useAppStore((state) => state.settings?.theme ?? 'system')` selector plus the existing `useSystemPrefersDark` hook, instead of `Theme.AUTO` (which only follows the OS `prefers-color-scheme` setting and can disagree with Orca's own manual theme choice). Scoping these subscriptions to the new lazy component also means they only re-render while the Emoji tab is actually mounted.
- The search placeholder is routed through `translate(...)` instead of being a hardcoded string.
- `repo-icon.tsx`'s centering logic is untouched.

## Screenshots

Captured from a real Electron build (`electron-vite build --mode e2e`) + Playwright E2E (`tests/e2e/repo-icon-emoji-picker.spec.ts`), so the glyphs are the actual native macOS emoji font, not a mock render.

## Testing

- [x] `pnpm lint` (oxlint, native/type-aware code-quality audits, reliability gates, max-lines ratchet, and all three localization checks — catalog/extraction/coverage)
- [x] `pnpm typecheck`
- [x] `pnpm test` — full suite; see Notes for pre-existing failures unrelated to this change
- [x] Added `RepositoryIconEmojiPicker.test.tsx` (5 tests): the full picker renders in place of the old grid, a valid emoji selection persists via the existing `onSetIcon` contract, an over-16-char sequence is rejected with a toast instead of a silent no-op, a valid skin-tone emoji is saved normally, and the "Current" trailing metadata renders
- [x] Added `tests/e2e/repo-icon-emoji-picker.spec.ts`: real Electron build, search → select → store persistence; doubles as the screenshot capture above
- [ ] `pnpm build` — not run; out of proportional scope for a renderer-only change. `electron-vite build --mode e2e` was run instead to produce the real build used for the E2E test/screenshots above.

## AI Review Report

Implemented directly with Claude Code (Sonnet 5). Review covered:
- **Cross-platform (macOS/Linux/Windows)**: pure renderer UI change with no modifier-key, path, or shell-specific code. `EmojiStyle.NATIVE` delegates glyph rendering to each OS's own emoji font — no image sprites, no platform branching needed.
- **SSH / remote / local compatibility**: no IPC or filesystem access was added; the component only reads store state and calls the existing `onSetIcon`/`updateRepo` path, which already works uniformly across local, SSH, and folder-workspace repos.
- **Agent / integration / git-provider compatibility**: N/A — this is a settings-UI-only change with no agent, provider, or CLI-integration surface.
- **Performance**: no new dependency. `emoji-picker-react` (~500KB) is now dynamically imported via a dedicated lazy `RepositoryIconEmojiPicker` chunk, confirmed in the real build output, so it only loads once a user opens the Emoji tab rather than on every repo-settings visit. `lazyLoadEmojis` avoids rendering the whole catalog eagerly once it does load.
- **UI quality**: only STYLEGUIDE tokens are used (`--card`, `--border`, `--accent`, `--font-sans`) — no hardcoded colors or fonts. Verified in light theme via the real-Electron-build screenshots above; dark-theme is driven by the same CSS variables and the new explicit `Theme.DARK`/`Theme.LIGHT` selector logic, but was not separately screenshotted (see Notes).
- **Security**: see Security Audit below.

No other issues were flagged during review.

## Security Audit

- **Input handling**: the value picked by the user is `EmojiClickData.emoji` from `emoji-picker-react`'s own built-in dataset, not free-text input. It is re-validated by `sanitizeRepoIcon` before persistence (16-char cap, fixed `type: 'emoji'` shape).
- **Command execution / IPC**: none added. This is a renderer-only UI change; no new IPC channel or command-execution path, and no external input reaches a shell or process.
- **Path handling**: N/A — no file paths are touched.
- **Auth / secrets**: N/A.
- **Dependencies**: no `package.json` changes. `emoji-picker-react` and `emojibase-data` are both pre-existing dependencies, reused as-is.
- **Follow-up**: none.

## Notes

- No platform-specific, SSH/remote-specific, agent-specific, integration-specific, or git-provider-specific behavior in this change — it's a settings-panel UI swap.
- `pnpm test` (full suite): the only failures observed are pre-existing and unrelated to this diff — `src/relay/agent-exec-handler.test.ts` (native spawn-args assertions) and an isolated `config/scripts/check-root-directory-entries.test.mjs` failure caused by this machine's system `/bin/bash` being 3.2 (no `declare -A` support), which the guard script this test exercises requires. Neither file is touched by this PR.
- Remaining risk: dark-theme rendering was not separately screenshotted (only light theme, see Screenshots). The follow-up commit closes the concrete dark-mode gaps found by review (search-active, highlight and picker-border variables were falling through to the library's raw defaults), but there is still no dedicated dark-theme visual capture.
- Three new i18n strings were added. They are now translated in all four non-English locales (es/ja/ko/zh) in this PR rather than left to the backfill pipeline — see the maintainer follow-up below. Each locale goes from 11774 to 11777 of 11862 keys.

## Maintainer follow-up (`fef6cc6db8`)

Reviewed against the readiness checklist with a fan-out of parallel reviewers (correctness/crash, UI + STYLEGUIDE, performance + cross-platform, localization, tests), each finding then adversarially verified. 22 candidate findings, 9 survived verification, all P2 — no P0/P1, nothing release-blocking. All nine are fixed here:

- **`lazyWithRetry` instead of raw `React.lazy`** — `RepositoryIconTabs.tsx` was the only renderer call site importing raw `lazy`; every other one uses `@/lib/lazy-with-retry`. React permanently caches a rejected dynamic import, so a transient chunk fetch failure (SSH/remote-served renderer, or stale chunk hashes right after an update) would blank the whole Settings page via the `page.*` boundary, and its Retry button would replay the cached rejection. It also filed a genuine `reportReactErrorBoundaryCrash` because the suppression path only recognises `LazyChunkLoadError`. This bites without a tab click: `RepositoryIconPicker` passes `initialTab: 'emoji'` for repos that already have an emoji icon.
- **`autoFocusSearch={false}`** — the library defaults it to `true`. Because the picker can mount with the panel, it stole focus from the settings search and scrolled the pane.
- **`font-family` was inert** — the library emits `font-family: sans-serif` on *every* descendant, so a root-level declaration never reaches the search input, category labels or preview text. Now scoped to descendants, so the picker actually renders in `--font-sans`.
- **Dark-mode variable gaps** — `--epr-search-input-bg-color-active`, `--epr-search-border-color-active`, `--epr-highlight-color` and `--epr-picker-border-color` were uncovered and fell back to the library's defaults (search input flipping to pure black on focus, invisible preview divider).
- **Scrollbar** — the picker's `.epr-body` used the raw Chromium scrollbar instead of the app's sleek one.
- **E2E asserted the store, not the DOM** — `tests/e2e/AGENTS.md` ("E2E Assertions Must Target the DOM, Not the Store") requires the final `expect()` to be user-observable. Added a DOM assertion on the rendered "Current:" caption alongside the store poll.

Two review comments were checked and need no action: Greptile's `--epr-search-input-border-color` note was already fixed by `894eaa2ddf` (`--epr-search-border-color` is the correct variable in the installed `emoji-picker-react@4.19.1`), and the descriptive i18n key names are consistent with the 2887 non-hash keys already in `en.json`.

**Translations** — `emojiTooLongForRepoIcon`, `currentEmojiSelection` and `searchEmojiPlaceholder` translated into es/ja/ko/zh, each matched against the terminology already established in that locale's catalog (e.g. es expands repo to "repositorio"; ja/ko keep "repo" in Latin script per their existing `RepositoryIconPicker` entries; zh uses the majority "仓库" form; the `{{value0}}` placeholder is preserved verbatim and zh/ja/ko follow the catalog's colon conventions).

**Gates re-run after the follow-up** — `pnpm lint` (full pipeline: oxlint, both code-quality audits, reliability gates, max-lines ratchet, skill-bundle verifies, all three localization checks), `pnpm typecheck` (all three tsconfigs), and `pnpm exec vitest run src/renderer/src/components/settings/` (138 files, 908 tests). All pass.

Note for the record: this PR's `PR Checks` runs had all been sitting at `action_required` (fork PR gating), so CI had never actually run on it before this follow-up — it has now been approved and run.

### DIFF
--- a/src/renderer/src/assets/main.css
+++ b/src/renderer/src/assets/main.css
@@ -1628,6 +1628,17 @@ html.native-shell .app-layout {
   animation: settings-shell-enter 180ms ease-out;
 }
 
+.repo-icon-emoji-picker .EmojiPickerReact {
+  --epr-bg-color: var(--card);
+  --epr-category-label-bg-color: var(--card);
+  --epr-text-color: var(--card-foreground);
+  --epr-search-input-bg-color: var(--input);
+  --epr-search-border-color: var(--border);
+  --epr-hover-bg-color: var(--accent);
+  border: none;
+  font-family: var(--font-sans);
+}
+
 .collapsible-height-content {
   overflow: hidden;
 }
--- a/src/renderer/src/assets/main.css
+++ b/src/renderer/src/assets/main.css
@@ -1633,10 +1633,23 @@ html.native-shell .app-layout {
   --epr-category-label-bg-color: var(--card);
   --epr-text-color: var(--card-foreground);
   --epr-search-input-bg-color: var(--input);
+  --epr-search-input-bg-color-active: var(--input);
   --epr-search-border-color: var(--border);
+  --epr-search-border-color-active: var(--ring);
   --epr-hover-bg-color: var(--accent);
+  --epr-highlight-color: var(--ring);
+  --epr-picker-border-color: var(--border);
   border: none;
+}
+
+/* The library sets `font-family: sans-serif` on every descendant, so inheriting from the root never lands. */
+.repo-icon-emoji-picker .EmojiPickerReact * {
   font-family: var(--font-sans);
+}
+
+.repo-icon-emoji-picker .epr-body {
+  scrollbar-width: thin;
+  scrollbar-color: color-mix(in srgb, var(--muted-foreground, #737373) 34%, transparent) transparent;
 }
 
 .collapsible-height-content {
--- a/src/renderer/src/components/settings/RepositoryIconEmojiPicker.test.tsx
+++ b/src/renderer/src/components/settings/RepositoryIconEmojiPicker.test.tsx
@@ -31,13 +31,15 @@ vi.mock('@/components/terminal-pane/use-system-prefers-dark', () => ({
 type MockEmojiClickHandler = (data: { emoji: string }) => void
 
 const emojiPickerMocks = vi.hoisted(() => ({
-  onEmojiClick: null as MockEmojiClickHandler | null
+  onEmojiClick: null as MockEmojiClickHandler | null,
+  searchPlaceholder: null as string | null
 }))
 
 vi.mock('emoji-picker-react', () => ({
   __esModule: true,
-  default: (props: { onEmojiClick: MockEmojiClickHandler }) => {
+  default: (props: { onEmojiClick: MockEmojiClickHandler; searchPlaceholder: string }) => {
     emojiPickerMocks.onEmojiClick = props.onEmojiClick
+    emojiPickerMocks.searchPlaceholder = props.searchPlaceholder
     return null
   },
   EmojiStyle: { NATIVE: 'native' },
@@ -62,6 +64,7 @@ describe('RepositoryIconEmojiPicker', () => {
     document.body.appendChild(container)
     root = createRoot(container)
     emojiPickerMocks.onEmojiClick = null
+    emojiPickerMocks.searchPlaceholder = null
     toastMocks.error.mockReset()
   })
 
@@ -74,6 +77,7 @@ describe('RepositoryIconEmojiPicker', () => {
   it('renders the full native emoji picker in place of the static grid', () => {
     renderPicker()
     expect(emojiPickerMocks.onEmojiClick).not.toBeNull()
+    expect(emojiPickerMocks.searchPlaceholder).toBe('Search emoji')
     expect(container.querySelector('.repo-icon-emoji-picker')).not.toBeNull()
   })
 
--- a/src/renderer/src/components/settings/RepositoryIconEmojiPicker.test.tsx
+++ b/src/renderer/src/components/settings/RepositoryIconEmojiPicker.test.tsx
@@ -1,0 +1,125 @@
+// @vitest-environment happy-dom
+
+import { act } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import type { AppState } from '@/store'
+import { RepositoryIconEmojiPicker } from './RepositoryIconEmojiPicker'
+
+const toastMocks = vi.hoisted(() => ({
+  error: vi.fn()
+}))
+
+vi.mock('sonner', () => ({
+  toast: { error: toastMocks.error }
+}))
+
+// Mock store exposing only the theme value, to verify the minimal-selector subscription.
+const storeMocks = vi.hoisted(() => ({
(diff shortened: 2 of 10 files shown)


nwparker:
Awesome @chucoding -- reviewing and will merge

PR #12060, opened 2026-08-02
fix(relay): desktop control lifecycle — truthful registration, skew tolerance, silence watchdog

Jinwoo-H:
Fixes the desktop half of the sustained Relay-only HOST_OFFLINE incident.

**Defects fixed**
- **Dead-control promotion (D1):** a socket delivering hello-ack and close in the same ws parser turn fired `onClose` before `activate()`; the dead control was then promoted to `activeControl` with no recovery path. `RelayControlOrigin.activate` now rejects controls that closed before activation; open/rebind fail into existing retry paths.
- **No liveness detection (D2):** the control client never required inbound traffic. New 75s silence watchdog mirrors the relay's 15s ping contract; silent/zombie sockets terminate and trigger origin recovery.
- **Registered-without-proof (D3):** `RelayAuthCoordinator` republished `registered` from ownership validity alone on every reconcile. It now requires `broker.isLive()` (plumbed client → origin → pool → broker) and replaces dead brokers.
- **Zero-tolerance clock check:** host-challenge freshness rejected any challenge stamped even 1ms in the future. A cell (~+100ms NTP drift) deterministically failed every activation. Now 2s leeway on both edges; challenges stay single-use and nonce-bound.
- **Silent failures:** broker reconcile failures and challenge-validation failures now log the failing operation/check by name (never values).

**Evidence**
- Red-before/green-after tests at each seam (broker-level dead-control connect, watchdog, coordinator dead-broker republish, skew tolerance): 58 relay tests pass.
- Live repro restored: Relay-only iOS simulator with invalid direct endpoint went from a sustained `relay_outer_4404` loop + desktop claiming `registered` with zero sockets, to truthful `registered` with a live control and real terminal round-trips over the relay.

Cloud-side counterpart: stablyai/orca-cloud#207 (activation fence, merged) + #208 (C9 canary image).


Jinwoo-H:
Superseded by the clean relay-only branch (see the replacement PR): this branch's inherited history deleted/reverted ~1.8k lines of unrelated main files, and merging it would have reverted that work. The relay changes are identical in the replacement.

PR #12062, opened 2026-08-02
Close three IME guard gaps found reviewing this stack

nwparker:
## Summary

Three defects an adversarial review of this stack turned up, all in the same class the stack exists to close: a keystroke the IME owns still reaching something that acts on it.

## 1. The Enter exemption was wider than the path that defers

`terminalShortcutIsOwnedByIme` released **every** Enter-shaped marked event when the caller passed `enterIsDeferredToCommit`:

```ts
if (enterIsDeferredToCommit && event.code === 'Enter') {
  return false
}
```

Only the `sendInput` branch defers. `terminal.expandPane` is bound to `Mod+Shift+Enter` by default, so composing that chord with two panes open expanded/collapsed the layout mid-preedit — no deferred-newline branch applies to it. Any terminal action a user rebinds onto an Enter chord had the same hole.

The guard now resolves the action first and exempts only `sendInput`.

### The guard and the dispatch were resolving different events

Fixing the above surfaced a second, quieter problem. Windows reports the composing Enter as `key: 'Process'`, and the dispatch rewrote it to `key: 'Enter'` *before* resolving:

```ts
const shortcutEvent = imeProcessEnter ? { key: 'Enter', code: e.code, ... } : e
const action = resolveShortcutEvent(shortcutEvent)
```

The guard, 60 lines earlier, resolved the raw event. So on Windows the guard was deciding against a binding the dispatch would never run. That normalisation now happens once, above the guard, and both use it — which is also what keeps the narrowed exemption correct on Windows rather than silently swallowing the deferred newline.

## 2. Preview terminal handed IME-owned releases to xterm

`preview-terminal-key-handler.ts` returned every non-keydown to xterm before consulting the guard. Once Kitty `REPORT_EVENT_TYPES` is negotiated xterm encodes an accepted keyup into a release sequence, so a release arriving during a live composition reached the pty. The pane path already evaluates its IME policy on every event type; the preview now does too, withholding from xterm without `preventDefault()` so the IME still receives the event.

This is the `keyup` half of the rule in AGENTS.md, and the preview writes to a real pty, so anything that escapes here reaches the agent's shell.

## 3. Page-level Escape listeners ran ahead of the composing element

`TaskPage.tsx` and `MarkdownPreview.tsx` both listen on `window` in the capture phase, so they run before whatever element is composing. Escape is how a user dismisses a candidate window:

- the task page called `preventDefault()` and blurred the focused field
- the preview closed search and moved focus to the preview root

Either ends the composition. The rollout earlier in this stack hardened the per-key handlers in both files but left these listeners, which is the global-listener case the rules name explicitly.

The task page's Escape policy moves into `task-page-escape-policy.ts`. That is not decoration: `TaskPage.tsx` is 12.4k lines with no render harness, and extracting the policy is what makes the guard testable at all.

## Tests

Every new test was mutation-checked — the fix disabled, the suite re-run, the intended test observed failing and nothing else:

| Test | Disabled fix causes |
|---|---|
| `keyboard-handlers-ime.test.tsx` — composing `Mod+Shift+Enter` does not expand a pane | fails (pane expands) |
| `terminal-shortcut-composition-guard.test.ts` — Enter chord bound to a non-deferring action stays claimed | fails |
| `preview-terminal-key-handler.test.ts` — IME-owned keyup withheld from xterm | fails |
| `MarkdownPreview.ime-capture-guard.interaction.test.tsx` — marked Escape leaves search open; marked find chord does not open it | both fail |
| `task-page-escape-policy.test.ts` — marked Escape ignored while a field has focus | fails |

Each is paired with a negative proving unmarked behaviour is unchanged: unmarked `Mod+Shift+Enter` still expands, an ordinary keyup still reaches xterm, an unmarked Escape still closes search and still blurs the field.

The Enter cases use the marked-but-real shape (`key: 'Enter'`, `keyCode: 13`, `isComposing: true`) rather than `key: 'Process'`, per the rule added earlier in this stack.

The markdown-preview test renders the real component and fires real events at the real window listener.

Renderer components + lib: 1779 files, 15,322 tests, green. Typecheck green, oxlint clean, max-lines ratchet reports no new bypasses.

## Two findings from the same review I did not act on

**The bare-229 equal-length replacement sends the whole textarea buffer.** Real, and it does contradict the "derive from ranges, never diff a mutable buffer" rule — but it is upstream xterm's own branch, not this stack's. In the patch only the bookkeeping line is ours:

```
   } else if ((newValue.length === oldValue.length) && (newValue !== oldValue)) {
+          this._dataAlreadySent = newValue;
     this._coreService.triggerDataEvent(newValue, true);
```

The insertion branch this stack *does* add derives from ranges and bails when the surrounding text moved. Rewriting upstream's sibling branch for a case with no recorded trace would widen the patch on speculation. Recorded as a known limitation instead.

**The deferred newline's 2-second idle ceiling can fire into a live preedit.** Also real, and deliberate — a composition that produces no `compositionupdate` for two seconds while the user pauses would send the newline into the preedit. The alternative strands the keystroke forever, and `compositionend` is documented in this codebase as not reliably delivered. The existing test pins the bounded behaviour on purpose. Changing the constant without a trace to justify a number would be guessing.

Both are stated as residual risk rather than quietly fixed or quietly dropped.


nwparker:
Superseded by #13128 (17b3dff3c41), which removed the layer these guard gaps were found in — 23 of the files here no longer exist on main.

PR #12081, opened 2026-08-02
fix(terminal): bound paired-client renderer work

OrcaWin:
## Summary

- coalesce decorative terminal title frames independently for each paired desktop subscriber before RPC serialization and encryption
- preserve semantic title, agent, bell, completion, and query facts plus every local synthetic animation frame
- gate legacy paired-host output before hidden xterm scheduling when the host lacks output-pause support, then restore one authoritative snapshot on reveal
- extend the remote hidden-retention reliability contract with deterministic current/current and legacy protocol oracles

## Reproduction

The pre-fix 64-PTY title harness delivered 576 client events after the first semantic frame. The fixed path delivers zero after convergence while retaining all 640 local frames.

The pre-fix legacy protocol harness kept hidden bytes on the xterm path and never requested a reveal snapshot. The fixed path performs no hidden xterm writes, restores the authoritative snapshot once, and continues live output.

## Verification

- 17 paired-terminal test files: 2,044 passed, 1 existing skip
- Node, web, and CLI typechecks
- full lint, code-quality, max-lines, localization, skill-manifest, and reliability checks
- headed isolated desktop host plus encrypted paired web client: passed
- isolated headless orca serve host: both parking and cold-activation scenarios passed

## Behavior and risk

Current hosts keep host-side output pause and ordinary parking unchanged. Visible terminals and local title animation are unchanged. Older hosts still transport hidden raw bytes because they cannot understand the pause opcode, but the client no longer feeds those bytes through xterm; reveal pays one bounded snapshot repaint to recover exact visible state.


OrcaWin:
@coderabbitai review

PR #12096, opened 2026-08-02
fix(linear): order Issues board columns Backlog -> Done

ElNelyo:
## What
Tasks → Linear → Issues → **Board** mode now shows status columns in workflow order (**Backlog … Done**) instead of the previous **Done → Backlog** artifact.

## Why
Board columns were rendered in the order their sections were *first inserted*, which followed the per-issue priority/`updatedAt` sort — so recently-touched **Done** issues landed before **Backlog**, hiding the intended left-to-right workflow.

## Root cause
`groupLinearIssues()` (`src/renderer/src/components/TaskPage.tsx`) returned `[...sections.values()]` directly from the insertion-order `Map`, discarding the workflow-state ordering the Linear API already returns sorted by `position`. This affected both the board and the grouped list (both call `groupLinearIssues`).

## Fix
- New `src/renderer/src/lib/linear-issue-grouping.ts` exports `orderLinearStatusSections()`, which sorts status sections by Linear workflow `state.type` rank (`backlog=0, started=1, completed=2, canceled=3`), tie-breaking alphabetically by label. Unknown types degrade safely (sorted after known ones). The function is generic over the section shape so it only depends on `state.type`.
- `groupLinearIssues()` now applies `orderLinearStatusSections()` **only when `groupBy === 'status'`** — assignee/priority/team/none groupings are untouched.
- `LinearGroupSection` moved to the new module and is re-imported in `TaskPage`.

## How to verify
Tasks → Linear → Issues → switch to **Board** mode → columns read left-to-right Backlog … Done. The grouped list view benefits from the same change.

## Tests
- New `linear-issue-grouping.test.ts` (6 cases): full ordering, the Done-before-Backlog regression repro, alphabetical tie-break, unknown-type fallback, non-mutation, empty list.
- Existing Linear suites stay green: `src/main/linear/teams.test.ts`, `linear-board-drag-payload.test.ts`, `linear-issue-context-snapshot.test.ts`, `linear-linked-work-item.test.ts` (27 tests).
- `pnpm run typecheck:web` clean; changed-file quality gate (`oxlint` + type-aware + react-doctor) reports 0 new findings.

No behavioral change for non-status groupings; no new data fetching.

### DIFF
--- a/src/renderer/src/components/TaskPage.tsx
+++ b/src/renderer/src/components/TaskPage.tsx
@@ -176,6 +176,7 @@ import {
 } from '@/lib/new-workspace'
 import type { LinkedWorkItemSummary } from '@/lib/new-workspace'
 import { buildLinearIssueLinkedWorkItem } from '@/lib/linear-linked-work-item'
+import { orderLinearStatusSections, type LinearGroupSection } from '@/lib/linear-issue-grouping'
 import {
   readLinearBoardIssueDragData,
   writeLinearBoardIssueDragData
@@ -600,12 +601,6 @@ function formatRelativeTime(input: string): string {
 
 type LinearProjectTab = 'overview' | 'issues'
 
-type LinearGroupSection = {
-  key: string
-  label: string
-  issues: LinearIssue[]
-}
-
 type LinearIssueListRow =
   | { type: 'section'; key: string; label: string; count: number }
   | { type: 'issue'; issue: LinearIssue }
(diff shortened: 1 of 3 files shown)


Jinwoo-H:
Linear status ordering needs an explicit unstarted/Todo rank. We are taking the focused rank-map fix into owner implementation under #17538. Closing the submitted branch; full credit to @ElNelyo. No action is requested from you. Thank you.\n\nOwner issue: https://github.com/stablyai/orca/issues/17538

PR #12102, opened 2026-08-02
fix(native-chat): drop stale version from the Claude opus picker label

### LINKED ISSUE #12084: [Bug]: Model picker still labels the "opus" entry as "Opus 4.8"
## Summary

The session model picker shows **"Opus 4.8"**, so it looks like Opus 5 is unavailable in Orca.
It isn't — the entry launches the CLI's `opus` alias, which resolves to `claude-opus-5`.
Only the display string is stale.

## Details

`CLAUDE_SESSION_OPTION_CATALOG` hardcodes the label:

```js
const CLAUDE_SESSION_OPTION_CATALOG = {
  models: [
    { id: "fable",  label: "Fable 5",  options: [claudeEffort(true)] },
    { id: "opus",   label: "Opus 4.8", options: [claudeEffort(true), CLAUDE_FAST_MODE] },
    { id: "sonnet", label: "Sonnet 5", isDefault: true, options: [claudeEffort(true)] },
    { id: "haiku",  label: "Haiku",    options: [] }
  ],
  modelApply: {
    launchArgs: (value) => ["--model", String(value)],
    ...
```

Since `launchArgs` passes the bare id, the version resolution happens in the CLI:

```
$ claude --model opus -p "Reply with ONLY your exact model id, nothing else."
claude-opus-5
```

The rest of the bundle is already Opus 5-aware — `MODEL_PRICING` has a `claude-opus-5`
entry, and the alias resolver maps opus+5 to `claude-opus-5`. The picker label is the
only place left on 4.8.

## Impact

Cosmetic, but actively misleading: it reads as "Orca doesn't support Opus 5 yet", and
the same trap will recur on every future Opus release.

## Suggested fix

Drop the version from the label (`"Opus"`, like the existing `"Haiku"` entry), or derive
it from the resolved model id instead of hardcoding it.

## Environment

- Orca 1.4.163 (latest release at time of filing)
- Windows 11
- Claude Code CLI 2.1.220

YuriNachos:
## Summary

The Claude model picker catalog hardcodes the `opus` entry's label as `Opus 4.8`, but the `opus` id resolves via the CLI to `claude-opus-5` — so the label is stale and reads as "Orca doesn't support Opus 5 yet". This is the only 4.8 leftover in the Claude path (`MODEL_PRICING` and the alias resolver are already Opus 5-aware).

Make the label versionless (`Opus`), mirroring the existing `Haiku` entry, so the next Opus rename doesn't re-stale it. Fixes #12084.

Changes:
- `src/shared/agent-session-option-catalog-claude-codex.ts:77` — `label: 'Opus 4.8'` → `label: 'Opus'` (1 line).
- `src/shared/agent-session-option-catalog.test.ts` — new regression test asserting the opus label is `'Opus'` and contains no digit (fails on any version-pinned string).
- `src/renderer/src/components/native-chat/NativeChatComposer.test.tsx:359` — update `expectedModelLabel` from `'Opus 4.8'` to `'Opus'`. This existing test asserts the catalog label that the composer surfaces to the user, so it had pinned the stale value; it is part of the same change, not a drive-by.

The Cursor-provider catalog (`agent-session-option-catalog-gemini-cursor.ts:89`, `label: 'Claude Opus 4.8'`, `id: 'claude-opus-4-8'`) is intentionally left alone — that is a different provider where the model id literally is `claude-opus-4-8`, so its label is correct.

## Screenshots

No visual change. (The opus row's label text changes from "Opus 4.8" to "Opus"; no layout or interaction change.)

## Testing

- [x] `pnpm test` — the catalog suite is green; the full run has 5 pre-existing failures unrelated to this change (verified by stashing the diff and re-running on clean `main` — the same 5 fail). No new failures introduced.
- [x] `pnpm typecheck` — green (node + web project references).
- [x] `pnpm lint` — green.
- [x] Added a high-quality, regression-catching test.

The new test was written first and confirmed **red on pre-fix code** (`label === 'Opus 4.8'` fails both `toBe('Opus')` and `not.toMatch(/\d/)`), then **green after the fix**. Mutation-checked: flipping the label back to a versioned string (e.g. `Opus 5`) re-triggers the failure, so the guard is real.

## AI Review Report

One-line display-string change + one regression test + one fixture update. Reviewed explicitly for:
- **Cross-platform (macOS/Linux/Windows):** a catalog label string surfaced in the picker UI; no platform-specific path. N/A.
- **SSH/remote/local + folder-workspace:** no runtime surface. N/A.
- **Provider/agent neutrality:** change is Claude-catalog-only; the Cursor provider's correct `claude-opus-4-8` label is deliberately untouched.
- **Performance / hot path:** none. N/A.
- **UI quality:** label text only; no layout/token change.

## Security Audit

No code, IPC, auth, path, or dependency surface touched — a display string in a static option catalog.

## Notes

- 3 files rather than 1 because `NativeChatComposer.test.tsx` had pinned the old label as the expected display value the composer surfaces; updating it is part of the same fix.
- This is my second contribution to Orca; happy to adjust if you'd prefer the label derived from the resolved model id at runtime instead of a static versionless string (the reporter's preferred minimal fix matches the existing `Haiku` precedent, which is what I went with).

### DIFF
--- a/src/renderer/src/components/native-chat/NativeChatComposer.test.tsx
+++ b/src/renderer/src/components/native-chat/NativeChatComposer.test.tsx
@@ -356,7 +356,7 @@ describe('NativeChatComposer', () => {
     expect(mocks.createClaudeModelSwitchConfirmationObserver).toHaveBeenCalledWith({
       ptyId: 'pty-1',
       settings: {},
-      expectedModelLabel: 'Opus 4.8'
+      expectedModelLabel: 'Opus'
     })
     expect(onSwitchToTerminal).not.toHaveBeenCalled()
   })
--- a/src/shared/agent-session-option-catalog-claude-codex.ts
+++ b/src/shared/agent-session-option-catalog-claude-codex.ts
@@ -74,7 +74,7 @@ export const CLAUDE_SESSION_OPTION_CATALOG: AgentSessionOptionCatalog = {
     },
     {
       id: 'opus',
-      label: 'Opus 4.8',
+      label: 'Opus',
       options: [claudeEffort(true), CLAUDE_FAST_MODE]
     },
     {
--- a/src/shared/agent-session-option-catalog.test.ts
+++ b/src/shared/agent-session-option-catalog.test.ts
@@ -19,6 +19,16 @@ describe('agent session option catalog', () => {
     expect(catalog?.models.find((model) => model.id === 'haiku')?.options).toEqual([])
   })
 
+  it('labels the opus entry without pinning a version', () => {
+    // Regression (#12084): `opus` resolves to whatever the CLI ships, so a
+    // version-pinned label goes stale — stay versionless like Haiku.
+    const label = getAgentSessionOptionCatalog('claude')?.models.find(
+      (model) => model.id === 'opus'
+    )?.label
+    expect(label).toBe('Opus')
+    expect(label).not.toMatch(/\d/)
+  })
+
   it('merges discovered labels while preserving cataloged option shapes', () => {
     const seed = getAgentSessionOptionCatalog('cursor')!.models
     const merged = mergeCatalogModels(seed, [


Lunivo:
Flagging a conflict: #12298 edits the same label line and the same two test files, and #12369 rewrites this area entirely with per-host CLI discovery. Worth coordinating before any of the three land.

On the substance — the versionless `Opus` label can't work as-is, and it's worth knowing why before this merges. The labels aren't only display strings: `readClaudeSessionOptionsFromTerminalScreen` recovers the live model by substring-matching Claude's TUI header against them (`claude-terminal-session-options.ts`, longest label first). A bare `Opus` still matches today's header, but it also matches `Opus 4.8` / `Opus 4.7` / `Opus 4.6` headers — so once the previous-generation entries exist in the catalog (as in #12298), a session on Opus 4.6 would resolve to the `opus` alias and a switch would silently move the user to Opus 5.

That's an argument against version-pinned labels too, which is what #12298 does and why it's a stopgap. #12369 resolves it properly: family labels plus family-prefix matching, with the exact per-host name overlaid from the CLI's own `list_models`. That gets the staleness-proofing you're after here without breaking header recovery.

Your underlying point stands — the reporter's "the same trap will recur on every future Opus release" is exactly right, and the fix is to stop hardcoding, not to pick a different hardcoded string.

YuriNachos:
Thanks @Lunivo — good catch on the recovery path. You're right that `readClaudeSessionOptionsFromTerminalScreen` substring-matches the catalog labels against the TUI header, so a bare `Opus` over-matches once multi-version entries exist (it happens to be fine today only because the catalog carries a single Opus entry). That makes this PR a stopgap at best, and #12369 (per-host CLI discovery, family-prefix matching) is the proper fix — it's active and from a maintainer. Closing this in favour of #12369 rather than ship a hardcoded-string swap the proper fix would supersede anyway. Appreciate the heads-up on the recovery interaction.

PR #12104, opened 2026-08-02
fix(terminal): disarm mouse tracking when seeding a session over a dead one (#12101)

### LINKED ISSUE #12101: [Bug]: Daemon never resets DECSET mouse tracking mode when a session's process dies, so it can survive into a later restore
### Operating system

Cross-platform (see analysis below)

### Summary

When a process that has armed terminal mouse tracking (SGR `?1000`/`?1002`/`?1003`/`?1006`/`?1016`, commonly Claude Code's own TUI, also vim/tmux/htop) is killed without a clean shutdown, nothing in Orca ever sends the matching DECRST reset. That mode stays recorded as "on" in the session's mirrored `TerminalModes` indefinitely. If a later restore/reattach replays that state into a pane that no longer has anything consuming mouse reports, moving the mouse over the pane echoes literal SGR motion reports (`^[[<35;col;rowM`) into the prompt.

This writeup started from two existing reports that share the same underlying mechanism from different angles:

* `#11160` (open, Windows): workspace Sleep force-kills a long-idle Claude session; wake restores a bare shell instead of resuming it. Doesn't mention mouse garbage, but the trigger (process force-killed without clean shutdown, then a pane restored in its place) is exactly the condition described here.
* `#8335` (closed, cannot_repro, macOS): terminal stuck with literal mouse motion reports echoed as text. The investigation there confirmed `POST_REPLAY_LIVE_AGENT_REATTACH_RESET` genuinely omits `RESET_MOUSE_REPORTING`, but also showed that specific report's repro (background/foreground during an external-editor round-trip) doesn't route through that reattach code at all, so the theory didn't match that particular trigger. The underlying code facts from that investigation still hold and are reused here against a different trigger.

### Code evidence

* `TerminalMouseModeMirror` / `TerminalPrivateModeTracker` (`src/main/daemon/terminal-mouse-mode-mirror.ts`, `src/main/daemon/terminal-private-mode-tracker.ts`) mirror DECSET mouse state from the raw byte stream into `TerminalModes` because `@xterm/headless` doesn't expose it.
* `buildRehydrateSequences` (`src/main/daemon/terminal-mode-rehydrate-sequences.ts`) turns a stored `TerminalModes.mouseTracking` back into an enable sequence (e.g. `\x1b[?1000h`) and this gets replayed into the renderer on reattach/cold-restore (`src/main/daemon/headless-emulator.ts`, `src/main/daemon/terminal-checkpoint-serializer.ts`, `src/main/runtime/orca-runtime.ts:11378`).
* Grepping the whole main process for a DECRST-style mouse reset write (`?1000l`, `?1002l`, `?1003l`, `?1006l`, `?1016l`) returns nothing. `handleSubprocessExit` in `src/main/daemon/session.ts` tears the session down without ever writing a reset into the emulator first.
* Process teardown (`src/main/pty-descendant-termination.ts`, `src/main/daemon/terminal-session-teardown.ts`) sends SIGTERM/SIGKILL (POSIX) or `taskkill /T /F` (Windows) straight to the process tree. Neither gives the dying process a chance to send its own mode-reset sequence, and neither path substitutes one on its behalf.
* On the renderer side, several reattach/reveal reset variants (`src/renderer/src/components/terminal-pane/layout-serialization.ts`) deliberately omit `RESET_MOUSE_REPORTING` when the pane still looks "agent-owned" (`POST_REPLAY_LIVE_AGENT_REATTACH_RESET`, `POST_REPLAY_LIVE_SNAPSHOT_RESET`, `POST_REPLAY_LIVE_AGENT_SNAPSHOT_RESET`), specifically to preserve mouse mode across reattach for genuinely live TUIs (mobile scroll gesture continuity, see the comment in `terminal-mode-rehydrate-sequences.ts`). That's intentional for a live process, but has no way to tell "genuinely still alive" apart from "was alive, died, and nothing has cleared the state since."

### Cross-platform

None of the code above branches on `process.platform`. The only OS-specific part is the kill mechanism itself (POSIX signal vs. Windows `taskkill /T /F`), and both leave the same gap. So this isn't specific to the Windows report it was first traced from.

### What would confirm this

I haven't yet built a deterministic repro (real daemon + real PTY, arm mouse tracking, force-kill the way Sleep does, inspect what the wake/restore replay actually sends). Doing that before proposing a fix, partly because `#8335`'s postmortem is a direct warning: a plausible-looking code path there turned out not to be what the reporter's specific trigger actually exercised, and a PR built on that unverified theory was closed without merging. Filing this now so the analysis and cross-platform check aren't stuck in a side comment on an issue with a different primary scope, and so anyone who can reproduce a stuck pane can check the one thing that would confirm or rule this out: after a wake/restore that leaves a bare shell where an agent used to be, does moving the mouse over the pane print literal `^[[<...M` text?

### Related

* stablyai/orca#11160
* stablyai/orca#8335

chappse6:
## What this fixes

Fixes #12101.

If you leave Orca idle and a Claude Code session (or vim, tmux, htop, any TUI that arms terminal mouse tracking) gets torn down while the app is asleep, then come back, the pane can be replaced by a bare shell that echoes raw mouse escape codes. Every time you move the mouse over that pane it prints garbage like `^[[<35;80;24M` into the prompt. The only recovery today is recreating the terminal, which loses the session.

## Root cause

A TUI arms mouse tracking via DECSET (`?1000h`/`?1002h`/`?1003h` + SGR `?1006h`/`?1016h`). Orca mirrors that mode into the session's `TerminalModes` (xterm's headless build doesn't expose it) and persists it in the checkpoint. When the process dies without a clean shutdown — workspace Sleep force-kills it, or the app relaunches and cold-restores — nothing ever resets the mode. The kill path sends SIGKILL / `taskkill`, so the dying process never gets to emit its own disarm.

The stale enable then rides the recovered content into a brand new PTY through two seed paths:

- `getRecoveredHistorySeedSegments` — the daemon `historySeedChunks` path (Sleep replaces the PTY in place)
- `buildColdRestorePayload` → `seedHeadlessTerminal` — the cold-restore-after-relaunch path

The new shell's emulator mirrors the enable from the seeded bytes and starts reporting mouse motion that nothing consumes.

This mirrors an existing, deliberate omission in the codebase: `buildRehydrateSequences` already strips kitty keyboard flags from seeds for the same reseed-parity reason. Mouse tracking was the same category, missing the same guard.

## The fix

Append a mouse-tracking disarm (`?9l?1000l?1002l?1003l?1006l?1016l`) after seeded content on both paths, gated on the restored modes actually having mouse tracking armed (`restoreHasMouseTrackingArmed`). Legacy scrollback restores and clean sessions pass through untouched.

Two restore contracts are preserved:
- **#7329 escape tail**: the disarm lands *before* any dangling partial escape, so the live subprocess's next bytes still complete the tail.
- **Alternate-screen**: the alt branch still restores only the normal buffer and no longer leaks the discarded TUI frame's escape tail.

A live-session reattach is untouched — a genuinely running agent keeps owning mouse tracking (mobile scroll gestures, OpenCode/OpenTUI panes).

## Tests

- `repro-12101-seeded-session-mouse-tracking.test.ts` (new): drives real `HeadlessEmulator` + `Session` through the `historySeedChunks` path; covers armed→disarmed, clean→untouched, alternate-screen, tail-only-restore (#7329), and disarm-before-tail ordering.
- `daemon-pty-adapter.test.ts`: added an integration test that writes an armed checkpoint to disk and asserts `spawn().coldRestore.scrollback` disarms the mode through the real cold-restore spawn path.

Full `src/main/daemon` suite: 1228 passing. `pnpm lint`, `pnpm typecheck`, `pnpm build` all pass.

## No visual change

No UI changes. This only affects the byte stream seeded into a restored terminal.

## AI coding agent review summary

- **Cross-platform**: the seed assembly (`terminal-history-seed-segments.ts`, `buildColdRestorePayload`) has no `process.platform` branching. The only OS-specific code is the kill mechanism (POSIX signals vs. `taskkill`), which is unchanged. Fix applies identically to macOS, Linux, Windows.
- **SSH / remote / local**: the affected paths are daemon-side cold-restore/seed logic shared by local, remote-server, and SSH sessions. The disarm is plain DECSET bytes with no host assumptions.
- **Agent / integration compatibility**: provider-neutral. Triggered by any TUI that arms mouse tracking (Claude Code, Codex, vim, tmux); no agent-specific branching. Live-agent reattach is explicitly not touched.
- **Performance**: adds at most one short constant string to a seed that only runs on session restore (not a hot path); gated so unarmed sessions allocate nothing extra.
- **Security**: no new inputs, no shell execution, no credential/path handling. Emits a fixed control sequence.
- **UI quality**: N/A (no visual change).

## Testing notes

Reproduces on any platform: run a mouse-tracking TUI, let its process die while backgrounded (workspace Sleep, or force-kill + relaunch), return to the pane. Before the fix the restored shell echoes `^[[<...M` on pointer movement; after, it doesn't. Live sessions still scroll/click normally.

X handle: Seeun Park (@chappse6)

### DIFF
--- a/src/main/daemon/daemon-pty-adapter.test.ts
+++ b/src/main/daemon/daemon-pty-adapter.test.ts
@@ -2432,6 +2432,58 @@ describe('DaemonPtyAdapter (IPtyProvider)', () => {
       })
     })
 
+    it('disarms mouse tracking in the cold-restore seed so a fresh shell does not echo motion (#12101)', async () => {
+      const sessionId = 'cold-restore-mouse-disarm'
+      const sessionDir = join(historyDir, getHistorySessionDirName(sessionId))
+      mkdirSync(sessionDir, { recursive: true })
+      writeFileSync(
+        join(sessionDir, 'meta.json'),
+        JSON.stringify({
+          cwd: '/projects/agent',
+          cols: 80,
+          rows: 24,
+          startedAt: '2026-07-25T10:00:00Z',
+          endedAt: null,
+          exitCode: null
+        })
+      )
+      // A dead agent session that armed any-motion + SGR mouse for its TUI and
+      // was killed before it could disarm them.
+      writeFileSync(
+        join(sessionDir, 'checkpoint.json'),
+        JSON.stringify({
+          snapshotAnsi: 'user@host:~$ ',
+          scrollbackAnsi: '',
+          rehydrateSequences: '\x1b[?1003h\x1b[?1006h',
+          cwd: '/projects/agent',
+          cols: 80,
+          rows: 24,
+          modes: {
+            bracketedPaste: false,
+            mouseTracking: true,
+            mouseTrackingMode: 'any',
+            sgrMouseMode: true,
+            applicationCursor: false,
+            alternateScreen: false
+          },
+          scrollbackLines: 0,
+          generation: 0,
+          checkpointedAt: '2026-07-25T10:00:00Z'
+        })
+      )
+      historyAdapter = new DaemonPtyAdapter({ socketPath, tokenPath, historyPath: historyDir })
+
+      const result = await historyAdapter.spawn({ cols: 80, rows: 24, sessionId })
+
+      expect(result.coldRestore?.scrollback).toContain('user@host:~$')
+      // The disarm must follow the re-armed enable so the net state is off.
+      const scrollback = result.coldRestore!.scrollback
+      expect(scrollback.lastIndexOf('\x1b[?1003l')).toBeGreaterThan(
+        scrollback.lastIndexOf('\x1b[?1003h')
+      )
+      expect(scrollback).toContain('\x1b[?1006l')
+    })
+
     it('uploads a large cold-restore seed in bounded protocol chunks', async () => {
       const sessionId = 'chunked-cold-restore'
       const sessionDir = join(historyDir, getHistorySessionDirName(sessionId))
--- a/src/main/daemon/daemon-pty-adapter.ts
+++ b/src/main/daemon/daemon-pty-adapter.ts
@@ -15,6 +15,10 @@ import {
 } from './history-manager'
 import { HistoryReader, type ColdRestoreInfo } from './history-reader'
 import { getRecoveredHistorySeedSegments } from './terminal-history-seed-segments'
+import {
+  DISARM_MOUSE_TRACKING_SEQUENCE,
+  restoreHasMouseTrackingArmed
+} from './terminal-mode-rehydrate-sequences'
 import { mintPtySessionId, parsePtySessionId } from './pty-session-id'
 import { supportsPtyStartupBarrier } from './shell-ready'
 import { CODEX_SHELL_READY_TIMEOUT_MS } from './session'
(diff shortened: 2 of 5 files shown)


chappse6:
Closing this — #12202 already shipped the fix for #12101 (and folded in #12112). Happy my issue #12101 helped drive it. Thanks @nwparker for the thorough fix.

PR #12119, opened 2026-08-02
feat(voice): allow selecting a microphone for dictation

### LINKED ISSUE #6625: [Feature]: Ability to select input device for voice dication
### Problem or use case

Not sure how it works right now, but I am assuming that the voice dictation uses the default audio device on the OS.

### Proposed solution

I'd rather an explicit audio device picker, which defaults to the default audio device, but can be changed to something else if the user desires.

_No response_

zhigang1992:
## Summary

Closes #6625

Adds a Settings → Voice **Microphone** control so desktop voice dictation can use a specific `audioinput` device instead of always following the OS default (e.g. built-in Mac mic while AirPods remain system default).

- Persist `microphoneDeviceId` / `microphoneDeviceLabel` on `VoiceSettings` (null = system default)
- Resolve the stored preference against the live device list, then pass `deviceId: { exact }` into `getUserMedia`; fall back to system default and toast a non-blocking notice when the device is gone
- Re-enumerate devices on open and on `devicechange`
- Detect the input device disappearing mid-dictation
- Settings search entry for the new control

## Design decisions

**1. Chromium's `default` / `communications` ids are filtered out, not offered.**
`enumerateDevices()` returns synthetic aggregate entries ("Default - MacBook Pro Microphone") whose ids alias whatever the OS default is. Listing them would show two rows meaning the same thing, and picking one would silently defeat the feature's whole purpose — it follows the OS, so it never pins and never reports a fallback. `normalizeMicrophoneDeviceId` maps both to null, which also degrades any preference that already stored one.

**2. Device identity is `deviceId` first, cached label second.**
Chromium salts device ids per profile+origin, so an id can rotate (storage reset, dev `localhost` vs packaged `file://`) while the same physical mic sits right there in the list. Resolution tries an exact id match, then a *unique* label match that adopts the fresh id — duplicate labels refuse to guess and report the device missing rather than pinning the wrong one. Without this the preference is permanently dead after a rotation: a fallback toast on every start plus a ghost row the user has to re-pick.

The resolver also distinguishes "list is empty / unreadable" from "device is gone". Before mic permission, enumeration returns only blank placeholders, so an empty list means *cannot tell yet* — the preference is kept and `getUserMedia` decides. Resolving eagerly there would silently drop every user to the default mic on first dictation.

**3. Mid-dictation device loss stops the session; it does not switch mics.**
Unplugging a pinned input ends the track without ending the audio graph, so the processor keeps feeding zeros — dictation looks live while capturing nothing. The track's `ended` event now surfaces that and stops cleanly through the existing state machine. Auto-re-acquiring a *different* mic mid-utterance is deliberately out of scope: it is a product decision, and rebuilding capture against a live worker session is a state machine that deserves its own design pass.

Two supporting choices fall out of the above: enumeration happens before capture (a cheap cached lookup) so a known-missing device skips the failed `getUserMedia` round trip that was clipping the first words of an utterance, and the fallback toast fires once per preference rather than once per push-to-talk press.

## Screenshots

- No screenshots (settings select row matching existing Voice pane patterns). Manual verify: Settings → Voice → Microphone lists System default + available mics after permission.

## Testing

- [x] Unit tests: `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/dictation/microphone-devices.test.ts src/renderer/src/hooks/use-audio-capture.capture-loss.test.ts`
- [x] Renderer `settings/` + `hooks/` suites: 202 files, 1524 tests
- [x] `tsc --noEmit -p config/tsconfig.tc.web.json`
- [x] `oxlint` + `oxfmt --check` on touched files
- [x] `check:max-lines-ratchet`, `audit:code-quality:native`
- [x] `verify:localization-catalog`, `verify:localization-coverage`
- [ ] `verify:localization-extraction` — could not run locally (`i18next-cli` absent from `node_modules`); needs CI
- [ ] `pnpm lint` (full suite)
- [ ] `pnpm test` (full suite)
- [ ] `pnpm build`

## AI Review Report

Self-review focused on:
- Capture wiring: preferred device only applied when non-null; OverconstrainedError/NotFoundError fall back without crashing
- Settings merge: defaults filled via existing `getDefaultVoiceSettings()` + persistence spread so upgrades pick up nulls
- Cross-platform: uses Web MediaDevices APIs (macOS/Linux/Windows Chromium); no hardcoded shortcuts/paths
- SSH/remote: dictation still captures in the local renderer that owns the mic; no remote audio path changes
- Mobile: shared settings shape only; mobile UI/capture paths untouched

Platform behaviors addressed defensively from documented Chromium semantics (aggregate device ids, id salting) — worth confirming live with a headset connected that no duplicate "Default - …" row appears.

## Security Audit

- Device ids are local media device identifiers chosen by the user and stored in settings (not secrets)
- No new IPC, command execution, or path handling. The picker's "Allow access" affordance opens a `getUserMedia` stream and immediately stops its tracks, reusing the existing main-window media permission handler
- Microphone permission still goes through existing `developerPermissions` / system media access when enabling voice

## Notes

Desktop-only picker. If the preferred mic is unplugged, dictation continues on the system default for that session without clearing the preference, so it reattaches when the device returns. If the mic is unplugged *during* dictation, the session stops with a notice rather than continuing on silence.

`enumerateDevices()` hides ids and labels until microphone permission has been granted, so the picker offers an "Allow access" button instead of presenting an unexplained empty list.


AmethystLiang:
this is a great feature. thank you!

PR #12120, opened 2026-08-02
fix(terminal): resolve Windows Korean IME newlines and delayed commits

### LINKED ISSUE #11878: [Bug]: Korean double consonants (쌍자음) trigger automatic line breaks in the Windows integrated terminal

holdn2:
Fixes #11878.

## Summary

On Windows 11 with the Microsoft Korean IME (2-set / 두벌식), Orca's terminal has two user-visible bugs:

1. **Newlines nobody typed, and dropped jamo.** A Shift-jamo mid-composition, or Enter near a composition boundary, submits a line the user never ended.
2. **Commits arriving a syllable late.** In fast consecutive Hangul input the previous syllable's commit lands after the next composition starts, so glyphs disappear, overlap, or trail one composition behind.

This is an **integration proposal**, not new discovery. Fixes for both already exist across four open PRs — #11616, #11937, #11896, #11870 — but each one alone leaves the other symptom open. This branch takes the minimum hunks from all four, preserves the original authors on the commits, and adds regression tests built from real-device captures.

**Maintainers may prefer to land the source PRs individually, and that is a perfectly good outcome.** This exists to offer one review surface where both symptoms close together. Happy to close it on request.

### The trigger is narrower than "Shift+jamo", and that matters for testing

Measured on real hardware by @KimYoungHwan8750 in #11878 (Windows 11 Pro 26200, Microsoft Korean IME 2-beolsik, Orca 1.4.163), the gate is `isWindows && hasPendingImeComposition && isTerminalImeProcessEnter(e)`, and the predicate is satisfied by the **Shift keydown itself** — Windows reports it as `key: 'Process'`, `keyCode: 229`, `shiftKey: true`, `code: 'ShiftLeft'`. So what actually decides it is whether a composition is already open when Shift goes down:

| state | injects a newline? |
|---|---|
| word-initial shifted jamo (`빼`, `까` starting a word) | no — nothing is composing yet |
| Shift pressed mid-composition, **no jamo at all** | **yes** |
| shifted batchim (`갔`, `있`) | **yes**, and this is the common case |
| bare **Ctrl** press mid-composition | **yes**, and silently — see below |

That last row is why this is pervasive rather than an edge case: Korean past tense is 았/었/였 — 했다, 갔다, 있다, 됐다, 왔다 — so a shifted batchim lands in a large share of ordinary sentences, and is mid-composition by construction.

The mechanism is not Shift-specific. Ctrl pressed mid-composition is reported as `Process`/229 the same way, `getTerminalImeModifiedEnterKind` returns `ctrl` for it, and the pre-fix predicate reached the **Ctrl+Enter** branch. That branch emits `\x1b[13;5u` unconditionally with no non-Kitty fallback, so a TUI that has not negotiated the protocol discards it silently — an injection with nothing visible on screen, which is why it has gone unreported. Verified here: on `main` that case emits `\x1b[13;5u`; on this branch it emits nothing.

It also means **a fix verified only against `계획` or a word-initial `빼` can pass while the real-world case still fails.** The three states above are in this branch as regression tests for exactly that reason.

### The blast radius is not one newline per shifted syllable

A device event probe run alongside a pty reader ([measurement](https://github.com/stablyai/orca/pull/12120#issuecomment-5163169125)) shows the modified-Enter chord lock is released by any React re-render — `modifiedEnterChordOwner.clear()` runs in the keyboard effect's cleanup, so every teardown/re-register drops it and the next auto-repeat keydown claims it again. Holding a modifier therefore injects **once per render**, not once per press. Driving the handler with the captured sequence across 50 auto-repeats:

```
no re-render during the hold        sendInput x1
re-render between each auto-repeat  sendInput x51
```

On the device that is three injections per hold in a quiet PowerShell pane, and a continuous stream of newlines in a repainting Claude Code pane. **The rate follows how busy the agent is** — the more output it streams, the faster the unwanted newlines arrive, which is backwards from what a user would expect and is why the reports here range from "occasional" to "unusable".

It also means the chord owner contributes nothing to safety here and the `code` gate is carrying all of it. That is sound for this PR — the gate sits upstream of the claim and stops these events from ever being read as Enter — but the layer beneath it is not a second line of defence, and should not be reasoned about as if it bounds anything.

### Why one PR rather than four merges

Two symptoms, six independent event paths. A fix on any one still leaves a reachable failure:

| Path | Failure | From |
|---|---|---|
| `isTerminalImeProcessEnter` | An active Windows IME reports **every** consumed key as `Process`/229 — including Shift itself. Checking only modifiers made a Shift press look like a chord. | #11616 |
| Orca shortcut matching | `Process` has no logical key, so matching fell back to the physical code and fired `Ctrl+K` / `Ctrl+W` mid-composition, here and in window-level handlers. | #11616 |
| Enter **keyup** synthesis | The chord was inferred from release-time modifier state. A plain committing Enter followed by a rolled-over Shift synthesized a Shift+Enter the user never chorded. | #11937 |
| Enter keyup, rapid double press | A single evidence slot per code let the first release drop the only entry, leaving the second press unguarded. | #11937 |
| Deferred-newline fallback | The 200 ms fallback called `finish()` directly, skipping the pending-composition re-check, and dropped a newline into the middle of a preedit. | #11896 |
| xterm `CompositionHelper` | A zero-delay finalizer scheduled by `compositionend` was discarded when the next `compositionstart` arrived, so the previous commit never flushed; and Escape left a stale `finalizerTimer`. | #11870 |

@KimYoungHwan8750 verified the keydown and keyup halves compose on real hardware, and that #11616 alone leaves the keyup path open.

## Screenshots

**No visual change.** This is terminal input handling; no UI surface is touched.

## Testing

- [ ] `pnpm lint` — see note below
- [ ] `pnpm test` — see note below
- [x] `pnpm build` — exit 0

Focused suite, rebased onto `c9a37f58d`: **7 files, 162 passed, 0 failed.**

```
keyboard-handlers-ime.test.tsx
keyboard-handlers-ime-real-device-trace.test.tsx
terminal-ime-deferred-newline.test.ts
terminal-ime-composition-route.test.ts
terminal-ime-xterm-composition-deduplication.test.ts
keyboard-handlers.test.ts
terminal-windows-shift-enter.test.ts
```

On the four IME files `main` has 98 tests; this branch has 132. Every new case was confirmed failing before its fix and passing after — except the word-initial one, which passes on both by design, because it pins that the fix does not start suppressing legitimate input.

<details>
<summary><b>Why <code>pnpm lint</code> and <code>pnpm test</code> are unchecked — both fail identically on pristine <code>main</code></b></summary>

I am not claiming these pass. They fail in my environment, and they fail the same way on a clean `main` checkout before any of my changes, so I could not use them as a signal.

**`pnpm test`** — my machine OOMs on the default worker count, so I ran it as 8 shards with `--maxWorkers=3`: 42,058 passed / 297 failed across 73 files. I then replayed those 73 files on this branch and on pristine `main` under identical conditions. Both give **exactly** 296 failed / 1,426 passed / 66 skipped. The one file-level difference is a swap between two load-flaky files that pass individually. No file under `terminal-pane/` fails anywhere.

**`pnpm lint`** — `verify:skill-bundle-manifest` reports `resources/skills/{current-manifest,snapshot-registry,release-mapping}.json` stale. Reproduced on a pristine `main` worktree, exit 1. Everything before it passes: oxlint native, oxlint type-aware, reliability gates, max-lines ratchet, bundled-skill-guides. This branch touches nothing under `resources/` or `skills/`.

If CI is green on these, treat the above as a local artifact.

</details>

## AI Review Report

**Cross-platform.** Deliberately split, and worth reviewer attention:

- Commits 1–4 are **Windows-only**, gated on `isWindows` at every call site. `isTerminalImeProcessEnter` is stricter now, but is only reached under `isWindows && hasPendingImeComposition`. The keyup evidence map is only written under `isWindows`, so on macOS/Linux it stays empty and the added reads are no-ops.
- Commits 5–7 are **cross-platform by design**. `deferredNewlineSender.defer()` is not Windows-gated and the xterm patch is shared. This is intended: #11896's rationale is that multi-segment Japanese bunsetsu conversion stays legitimately pending past any fixed total-elapsed deadline, so the idle ceiling is measured from the last `compositionupdate`. `terminal-ime-composition-route.test.ts` and the xterm dedup suite (Japanese, Chinese, ibus Hangul cases) pass unchanged.

No `metaKey` hardcoding, no new shortcut labels, no path handling.

**Resource lifecycle.** The keyup evidence map is keyed by `e.code` and bounded at 8 entries per code, and is cleared on both `blur` and effect cleanup. Auto-repeat keydowns deliberately do not stack an entry, since a repeat run ends in a single release. The re-armed fallback timer clears itself in `finish()`, which also removes the added `compositionupdate` listener; `maxIdleMs = Math.max(fallbackMs, …)` keeps the ceiling finite, so a stuck composition cannot strand a keystroke.

**SSH / remote / local.** No transport, PTY routing, or reconnect behavior is touched. #11896's transport-identity and uncaptured-session routing changes are deliberately excluded — only its fallback re-arm hunk is ported.

**Performance.** A couple of array operations per Enter keydown/keyup on Windows only. The fallback timer re-arms at most `maxIdleMs / fallbackMs` times per deferred newline instead of firing once, bounded at 10 by default and only while a composition is open.

**What review flagged, and what changed as a result.**

- No test exercised the fixes together in one composition session. That gap is closed by the combined regression test.
- Reviewing @KimYoungHwan8750's follow-up `e17d22cf`, I reproduced the rapid-double-Enter hole against this branch before taking it: two keydowns overwrite the single evidence slot, the first release drops it, and the second release synthesizes the very `\x1b\r` the guard exists to prevent. Fixed here with his commit, authorship preserved.
- The xterm `.patch` introduces 20 trailing-whitespace lines. Those are blank context lines inside the embedded unified diff, and the file is byte-identical to #11870's artifact, so I left it rather than diverge from the original author's output.

## Security Audit

Low risk surface:

- **Input handling** — the only new inputs are DOM `KeyboardEvent` fields and composition events. They are compared, never parsed, concatenated, or evaluated. The change narrows what is accepted as a modified Enter; it never widens what reaches the PTY.
- **Command execution / path handling / auth / secrets / IPC** — none touched.
- **Dependencies** — no version changes. The xterm patch and its lockfile hash come from #11870's commits; `pnpm install --frozen-lockfile` is clean.
- **New failure modes** — the shortcut guard calls `stopImmediatePropagation()` on IME-consumed non-Enter chords during an active composition. xterm already ignores `keyCode` 229 while composing, so no input is lost, and it cannot fire outside an active composition on Windows.

No follow-up items.

## Notes

**Attribution.** Original authorship is preserved with `--author` on the ported fixes, and `Co-authored-by:` where only part of a commit was taken or where the work came from a shared capture.

| Integrated change | Source | Original author |
|---|---|---|
| physical-Enter classification + shortcut guard | #11616 | @nwparker |
| Enter keyup press-time evidence, and its per-press follow-up | #11937 | @KimYoungHwan8750 |
| fallback re-arm (partial port) | #11896 | @nwparker |
| rapid restart + cancel cleanup | #11870 | @prxyeo |
| real-device capture states | #11878 | @KimYoungHwan8750 |

Thank you to all four authors — this is their work, reassembled.

**Known adjacent gap, not fixed here.** `heldImeEnterModifiers` is only cleared on a modifier keyup or on `blur`, never at a composition boundary. @mochizuki1105 reports the same visible symptom with a **Japanese** IME through that path, and their control experiments (Shift required, active composition required, Linux unaffected) point at it specifically. This branch does not close that, and I am not claiming it does.

**Out of scope, deliberately:** Microsoft Pinyin Backspace/Delete and Shift mode-toggle (#11550), macOS Pinyin stale preedit (#11938), generalizing all IME-owned shortcuts beyond Enter (#11273), #11896's transport-routing changes, and UI or settings changes.

**Verification status — what is measured, and what is not.**

Marked ready for review to unblock CI and put it in front of reviewers. The gap below is unchanged; it is stated plainly here rather than left implicit behind draft status, and I am not asking for a merge ahead of it.

Beyond the unit suite, I built a standalone browser harness loading the real patched `@xterm/xterm` from this branch plus the real production predicates, and typed into it with a real Microsoft Korean IME on Windows 11 build 26200 — genuine composition events rather than synthesized ones. It showed correct commit ordering with no stray `\r` during Korean input. That covers commits 1, 2, 5, 6 and 7.

It does **not** cover the keyup gate, which lives inside the `useTerminalKeyboardShortcuts` hook and needs the full pane-manager and transport surface; that remains unit-tested only. The harness also has no PTY, so the ConPTY round trip is unverified.

Full in-app verification under Electron is still outstanding, and it is the one thing I would want before this merges. @KimYoungHwan8750 has a machine reproducing this daily and has offered byte-level confirmation of what actually reaches the pty from a raw-mode reader — that does not need a branch build, and it would establish the pre-fix baseline that everything here has so far inferred rather than measured. I will post those results as soon as they exist.

**Reviewers on Windows with a Korean IME: independent confirmation is very welcome**, as is a maintainer steer on whether you would rather take the source PRs separately.


dogfooter8750:
Thanks for carrying the commit with authorship preserved, and no objection at all to folding my tests into `keyboard-handlers-ime.test.tsx` — one handler test file for the combined branch is the right call. Glad the rollover gate turned out to be load-bearing for your combined regression test; that matches what I saw locally, where a plain committing Enter mid-composition is only safe once the keyup path stops inferring the chord from release-time modifier state.

**One thing to pull in: this branch carries `c865c0b`, but I pushed a follow-up (`e17d22cf`) a few minutes before you opened this, and it closes a real hole that is still open here.**

The problem is that `c865c0b` kept a single slot per code (`Map<string, number>`), so a rapid second Enter press goes unguarded: the first release finds a mismatched timeStamp, drops the only entry, and the second release then synthesizes the very `\x1b\r` the gate exists to prevent. Greptile flagged the shape of it on my PR and it reproduces.

I verified it directly against this branch rather than assuming. Applying this PR's diff onto `main` and driving the existing harness with two Enter keydowns (ts 10, 20) before either release, then two keyups (ts 30, 40) with a rolled-over Shift:

```
COMBINED_RESULT x1 [["\r"]]   <- a Shift+Enter newline from a sequence that never chorded one
```

Cherry-picking `e17d22cf` on top applies cleanly and takes it to:

```
COMBINED_RESULT x0 []
```

The follow-up tracks one entry per press instead of one slot per code and drains exactly one per physical release, so every press stays guarded until its own release. Balancing-keyup semantics are unchanged. Two details that are new surface: auto-repeat keydowns deliberately do not stack an entry (a repeat run ends in a single release, so stacking would leave residue that release cannot drain and would suppress a later legitimate keyup-only synthesis), and the list is bounded at 8 so a press whose release never arrives cannot grow it without end. It ships with two regression tests, the first of which fails on `c865c0b`.

**Unrelated heads-up on this branch:** `terminal-ime-xterm-composition-deduplication.test.ts` fails 2 tests in my local checkout — `flushes the previous Hangul commit while the next composition is active` and `clears a pending completion timer when Escape follows compositionend`. They fail with or without my follow-up applied, so they are not caused by it. I did not regenerate `node_modules/@xterm/xterm` from the updated patch file, so this is most likely just that local artifact rather than a real failure — worth a glance once CI runs, since this PR does touch the xterm patch.

Environment for all of the above: Windows, `vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/` → 2897 passed with only those 2 pre-existing failures.

Happy for this to be the branch that lands — I'd rather one combined fix ship than have my PR land separately. I also have a Windows 11 + Microsoft Korean IME (2-set) machine reproducing daily with Claude Code and Codex, so if you push an adhoc branch build (#12051) I can run the real-device pass on this exact combination and report back.

holdn2:
Verified and applied — thank you, both for the follow-up and for the trace.

**`e17d22cf` — reproduced before taking it.** I wrote the case from your description against this branch rather than assuming it, and it failed exactly as you said:

```
× keeps a rapid second Enter press guarded until its own release
  expected "vi.fn()" to not be called at all, but actually been called 1 times
```

Two keydowns overwrite the single slot, the first release finds a mismatched timeStamp and drops the only entry, and the second release is then unguarded. Real hole, and mine to have missed in review.

Cherry-picked with `--author` preserved. The only conflict was the test file, since I had folded your earlier tests into `keyboard-handlers-ime.test.tsx`; I dropped the duplicate file and moved your two regression tests into that suite instead of the one I had written, because yours also pin the auto-repeat case and mine did not.

**The trace.** I turned it into two regression tests, both of which fail on `main` and pass here:

| case | main | this branch |
|---|---|---|
| full `있다` replay, zero Enter events | one `\x1b\r` | none |
| bare Shift press mid-composition, no jamo | one `\x1b\r` | none |

Your follow-up narrowing the rule to "is a composition already open when Shift goes down" is in as three more, and the word-initial one is the valuable half — it passes on `main` *and* here, so it pins that the fix does not start suppressing a legitimate word-initial 빼. The other two fail on `main` only.

That is the split you described, reproduced independently:

```
word-initial shifted jamo   main: x0   here: x0    <- negative, must stay silent
Shift mid-composition       main: x1   here: x0
shifted batchim (갓/있)      main: x1   here: x0
```

All five carry a `Co-authored-by:` trailer to you, since they are your measurements. Say the word if you would rather they were attributed differently, or if I have mis-transcribed any row.

**On the xterm suite failing for you** — your own diagnosis is right, and I can confirm it from this side. The two tests you named are the ones commits 5 and 6 add, and they exercise the *patched* `CompositionHelper`. Changing `config/patches/@xterm__xterm@6.1.0-beta.287.patch` does nothing until `pnpm install` re-applies it to `node_modules/@xterm/xterm`. With a `pnpm install --frozen-lockfile` after checkout, that suite is 49/49 here. Nothing for you to fix.

Focused suite on this branch is now 158 passing across the six terminal-pane files, rebased onto `0ae917440`.

**On the adhoc build:** yes please, that would close the biggest gap in this PR. Everything I have is either unit-level or from a browser harness I built around the real patched xterm — genuine IME events, but no PTY and no Electron. I will ping you here once a build exists.

And to be explicit, since you asked me not to close mine: I am not asking you to close yours either. If a maintainer prefers landing the keyup half separately, that is a good outcome and this branch rebases onto it without trouble.

dogfooter8750:
Attribution is exactly right, thank you — and reproducing the double-press case before taking it rather than trusting my description is the right instinct. No changes wanted there.

**One correction, since you asked.** In `3488116a` the batchim test carries:

```
// 갓: ㄱ and ㅏ are already composing by the time Shift comes down for ㅆ.
```

`갓` is the wrong syllable: its batchim is ㅅ, which is plain `t` with no Shift, so `갓` never triggers this path at all. The shifted one is ㅆ = `Shift+t`, which gives **갔**. The comment describes ㅆ correctly, only the syllable is off by the shift. Same family: 갔 / 있 / 했 / 됐 / 왔. The events in the test body are right — it is only the comment that would mislead the next reader into thinking an unshifted batchim reproduces.

**A new real-device finding, from the same probe.** The trigger is not Shift-specific. Pressing **Ctrl** mid-composition is reported the same way:

```
| t(ms) | event   | key     | code        | keyCode | ctrl  | gate   |
| 3903  | SESSION-START |   |             |         |       | OPEN   |
| 8526  | keydown | Process | ControlLeft | 229     | true  | OPEN   |   <- composition open
| 8526  | keydown | Control | ControlLeft | 17      | true  | OPEN   |
| 13171 | SESSION-END   |   |             |         |       | closed |
| 14910 | keydown | Control | ControlLeft | 17      | true  | closed |   <- same key, composition closed
```

Same physical key, two different reports depending only on whether a composition is open — which is a clean paired control for the whole mechanism. `getTerminalImeModifiedEnterKind` returns `'ctrl'` for it, so the pre-fix predicate accepts it and the resolver reaches the Ctrl+Enter branch.

Nothing was visible on screen in Claude Code, and I think that is expected rather than evidence of no injection: unlike the Shift+Enter branch, the Ctrl+Enter branch has no `canSendCsiU` gate and no fallback, so it emits `\x1b[13;5u` unconditionally (`terminal-shortcut-policy.ts`, with the `#2418` note). A TUI that has not negotiated KKP will discard an unknown CSI sequence silently. So the likely shape is a *silent* injection that becomes visible in any TUI that binds Ctrl+Enter or runs with KKP on.

Your `code` gate already covers it — `ControlLeft` is neither `Enter` nor `NumpadEnter` — so this does not change the fix, only its scope. It may be worth one more case in the same suite: bare Ctrl press mid-composition, asserting silence. Happy to hand you the captured event shapes if you want it pinned.

**On the PTY gap.** You do not have to wait for an adhoc build for that half. I have Electron and a real PTY on 1.4.163 right now, so I can give you byte-level confirmation of what actually reaches the shell — running a raw-mode reader in a pane and typing Korean, so an injected `\x1b\r` or `\x1b[13;5u` prints literally even when a TUI would swallow it. That establishes the pre-fix baseline at the PTY, which is the part everything here has so far inferred rather than measured. I will post those results, and I am still on for the adhoc build whenever one exists.

holdn2:
You are right about the syllable, thank you — fixed. 갓 takes a plain ㅅ and never reaches
this path; the events compose 갔 (ㅆ = Shift+t), which is what the test actually sends. The
comment now says so, and I corrected the same slip in the PR description.

**The Ctrl finding is real, and I reproduced it before adding it.** I built the case from
your event shapes rather than trusting the description, and against `main`'s predicate it
does not stay silent:

```
× sends nothing for a bare Ctrl press mid-composition
  Received: ["\x1b[13;5u"]
```

On this branch it emits nothing. That is now a regression test in
`keyboard-handlers-ime-real-device-trace.test.tsx`, alongside the Shift cases, with a
`Co-authored-by:` trailer to you.

Your read of why nothing was visible in Claude Code matches what the code says: the
Ctrl+Enter branch in `terminal-shortcut-policy.ts` has no `canSendCsiU` gate and no
fallback, so without KKP negotiated the sequence is discarded rather than rendered. A silent
injection is worse than a visible one for triage — it would surface only in a TUI that binds
Ctrl+Enter or has KKP on, which is probably why nobody filed it. Agreed that it does not
change the fix, only its scope: `ControlLeft` is neither `Enter` nor `NumpadEnter`, so the
`code` gate already covers it.

**On the PTY measurement — yes, please, that is the more valuable half.** Everything in this
PR at the PTY boundary is inferred, not measured: the browser harness I built has real IME
events against the real patched xterm, but no pty, so what actually reaches the shell is the
one thing I cannot check. Byte-level confirmation that an injected `\x1b\r` or `\x1b[13;5u`
appears in a raw-mode reader would establish the pre-fix baseline properly, and it does not
need a branch build to be useful. I will fold whatever you get into the PR description and
credit it.

Current state of the branch: rebased onto `c9a37f58d`, 11 commits, focused suite 159 passing
across 7 files. Still draft until there is a real-device pass on the built app.

dogfooter8750:
PTY measurement, as promised. **Both injections are real at the byte level.** Orca 1.4.163, Windows 11 Pro 26200, Microsoft Korean IME 2-beolsik, integrated terminal running a raw-mode reader in a plain PowerShell pane (`setRawMode(true)`, every chunk printed with `JSON.stringify`).

What I did, in order, pressing **no Enter key at any point**:

1. typed `가` — composition opens and stays open
2. held **Ctrl** down for about 10 seconds, released
3. held **Shift** down for about 10 seconds, released
4. pressed Esc, then Ctrl+C to exit

What the shell received:

```text
ESC [13;5u      <-- CSI-u 13;5 = Ctrl+Enter injection
ESC [13;5u
ESC [13;5u
ESC CR          <-- Shift+Enter newline injection
ESC CR
ESC CR
EA B0 80        <-- the committed 가 arriving only now
ESC             <-- the Esc key I pressed to close the composition
```

(The reader printed each chunk with `JSON.stringify`; the first three lines were `"\u001b[13;5u"`, the next three `"\u001b\r"`.)

Three things follow from this.

**The Ctrl variant reaches the shell.** Not inferred from the resolver any more — `\x1b[13;5u` is physically in the pty stream. It stays invisible in Claude Code only because an un-negotiated CSI-u sequence is discarded, exactly as we read it. Worth keeping the regression test you added.

**The ordering inversion is visible at the byte level.** `가` was typed *first*, before any modifier was touched, yet the shell received all six injected sequences **before** it received `가`. The injected bytes go straight out through `sendInput` while the composed syllable waits for the composition to settle. That is the mechanism behind the reports here where the break lands one or two syllables to the left of the key that caused it — measured now, rather than deduced from transcripts.

**One thing I cannot yet explain: it fired three times per hold, not once.** I expected exactly one per press cycle, because `createTerminalImeModifiedEnterChordOwner` refuses a second claim until the modifier's keyup releases it. I checked that assumption against the handler rather than trusting it — replaying a held modifier as auto-repeat keydowns (both the `Process/229` form and the plain `Shift/16` form, up to 300 repeats over a simulated 10 seconds) gives `sendInput x1` every time. So auto-repeat alone does not explain three.

Two candidates I cannot separate from the pty side alone:

1. Windows delivers a held modifier under IME as repeated **down/up pairs** rather than repeated keydowns. Each up would release the chord and let the next down claim it again.
2. Something in the running app re-arms in a way the unit harness does not model.

Either way it is a real amplifier — holding a modifier is not exotic, and three injections from one hold is worse than one. If it is (1), the chord owner is not the backstop it looks like, which may matter to how much weight the `code` gate is carrying on its own.

I can settle it by running the DevTools event probe and the pty reader at the same time, so the keydown/keyup sequence and the emitted bytes line up on one timeline. I will do that and post it. Flagging it now rather than sitting on it, in case it changes anything you want in the branch.

All of the above is pre-fix baseline on 1.4.163, so it stands on its own and does not need a branch build. Still happy to run the real-device pass once one exists.

holdn2:
Both applied — the first one was a real hole, and I verified it before taking it.

**Guard ordering.** Confirmed by calling the matcher directly:

```
matchFileSearchShortcut({ key: 'Process', code: 'KeyF', ctrlKey: true, shiftKey: true }, 'win32')
  → true
```

`shouldUseNonLatinShortcutPhysicalFallback` allows the physical-code fallback whenever
Ctrl/Meta is held and the logical key is not a Latin shortcut key — and `'Process'` is
length 7, so it qualifies. The guard sat at line 536 while `matchFileSearchShortcut` runs at
493, so a composing Ctrl+Shift+F claimed `sidebar.search.toggle` on its `code` first.
Hoisted the guard above it; the pending-composition lookup it needs was already available
there. At the hook level: with the guard in its old position the new regression test sees
`onSearchSelectedText` called once, and none after the move.

One correction: **`matchSearchNavigate` is not affected.** It compares `e.key.toLowerCase()`
against `'g'` directly rather than going through `keybindingMatchesAction`, so `'Process'`
never matches. I checked that path too rather than assuming it had the same shape.

Worth noting for anyone reading across the stack: the guard's previous position is
byte-identical to `9f449d7` in #11616, so this gap is not specific to this branch.

**Stale pending reference.** Fair, and applied. The cancel path does clear `finalizerTimer`
in place today, so the assertion was passing for the right reason — but the settle path
detaches the pending object instead, so the test was coupled to which mechanism ran rather
than to the observable state. It now re-reads after the Escape keydowns.

Focused suite is 160 passing across 7 files, typecheck clean.

dogfooter8750:
Settled the three-injections-per-hold question from my previous comment, and the answer is worse than either candidate I offered. **The modified-enter chord lock is not a backstop at all — a re-render releases it, so one held Shift injects one newline per render.**

## The device ruled out the first candidate

I ran the DevTools event probe and the pty reader together on 1.4.163 so the key events and the emitted bytes share a timeline. Holding Shift for ~9 seconds mid-composition produced:

```text
15503  keydown  Process/229  ShiftLeft   shift=true   gate=OPEN
15503  keydown  Shift/16     ShiftLeft   shift=true   gate=OPEN
16015  keydown  Process/229  ShiftLeft   (auto-repeat begins, ~512ms after press)
16046  keydown  Process/229  ShiftLeft
16076  keydown  Process/229  ShiftLeft
 ...   ~270 more auto-repeat keydowns at ~31ms intervals ...
24190  keydown  Process/229  ShiftLeft
24214  keyup    Process/229  ShiftLeft   <-- the only keyup in the entire hold
24214  keyup    Shift/16     ShiftLeft
```

So Windows does **not** deliver a held modifier as repeated down/up pairs. It is one keydown, a long auto-repeat run, and exactly one keyup at the end. That kills the candidate I thought most likely.

## The real amplifier is effect re-registration

`createTerminalImeModifiedEnterChordOwner` refuses a second claim until `release`, and `release` only runs on a modifier keyup — of which there is one. But `modifiedEnterChordOwner.clear()` also runs in the keyboard effect's **cleanup**, so every teardown/re-register cycle drops the lock and lets the next auto-repeat keydown claim again.

Driving the handler with the captured event sequence, holding Shift across 50 auto-repeats:

```text
no re-render during the hold          sendInput x1
re-render between each auto-repeat    sendInput x51
```

That matches what I see on the device. In a plain PowerShell pane, which emits almost nothing, I measured three injections per hold. In a Claude Code pane, which repaints continuously, **holding Shift produces a continuous stream of newlines** — the injection rate simply follows the render rate.

## Why this matters for the branch

The blast radius is not "one stray newline per shifted syllable". It is unbounded, and it scales with how busy the TUI is, which is exactly backwards from what a user would expect — the more the agent is streaming output, the faster the newlines arrive.

It also means the chord owner is carrying none of the safety here, and the `code` gate is carrying all of it. That is fine for this PR, since the gate is upstream of the claim and stops these events from ever being read as Enter. But it is worth knowing that the layer underneath it is not a second line of defence, in case anyone later reasons about the claim/release logic as if it bounds anything.

I have not tried to find which dependency in that effect's array changes per render, and it may not matter for this branch — the observable fact is that the lock does not survive a re-render, and the device confirms the consequence. Happy to dig into the dependency list separately if you think it is worth a follow-up.

Regression case if you want it pinned, in the same style as the others: composition open, one `Process/229` `ShiftLeft` keydown, N auto-repeat keydowns with a re-render between each, one keyup. Asserting a single `sendInput` fails on `main` at `N+1` and passes on this branch at zero.

holdn2:
Candidate 1 is it, and it reproduces against the handler.

Replaying the hold as repeated down/up pairs rather than as auto-repeat keydowns:

```
3 pairs -> sendInput x3
5 pairs -> sendInput x5
```

One-for-one with the pair count on `main`, and none on this branch. Three pairs matching
the three you measured for one ~10s hold is what makes me think this is the same shape
rather than a coincidence.

The mechanism, from `onKeyUp`:

```ts
const releasedImeEnterModifier =
  e.key === 'Shift' ? 'shift' : e.key === 'Control' ? 'ctrl' : null
if (releasedImeEnterModifier) {
  heldImeEnterModifiers.delete(kind)
  modifiedEnterChordOwner.release({ kind, code: e.code, timeStamp: e.timeStamp })
}
```

The release is keyed on the *logical* key. The `Process`-form keyup reports
`key: 'Process'`, so it does not release — but Chromium sends the native `Control`/17
keyup alongside it, and that one does. So each down/up pair hands the chord back and the
next press claims it again. `createTerminalImeModifiedEnterChordOwner` is doing exactly
what it says; it just never sees a hold as a single press.

Two things follow. You do not need the combined DevTools+pty run to settle this particular
question — though it is still worth having for the ordering timeline, which is the part I
cannot reproduce in a unit harness at all. And your reading of the backstop is right: for
this case the `code` gate is carrying it alone, which it does, since `ControlLeft` is
neither `Enter` nor `NumpadEnter`.

Added as a regression test parameterised over 3 and 5 pairs, with a `Co-authored-by:` to
you for the measurement. Focused suite is 162 across 7 files.

On the ordering inversion — `가` typed first but arriving after all six injected sequences
is the clearest explanation anyone has posted for why the break shows up one or two
syllables to the left of the key that caused it. I would like to quote that byte trace in
the PR description with attribution, if you are happy with that.

holdn2:
I was wrong about candidate 1, and your probe settles it.

I matched on the count — 3 down/up pairs giving 3 sends lined up with the three you
measured — and stopped there without checking whether the hardware emits that sequence at
all. Your event trace shows it does not: one keydown, a long auto-repeat run, and exactly
one keyup. A mechanism that produces N sends is not the same as the mechanism that ran, and
I should not have called it settled on a matching number.

Reproduced your actual finding against the handler, holding Shift across 50 auto-repeats:

```
no re-render during the hold        sendInput x1
re-render between each auto-repeat  sendInput x51
```

Exactly your numbers. Zero on this branch either way.

Replaced the regression test accordingly — the previous one pinned a sequence that cannot
occur, so it was worse than no test. The correction is in `1fa9d205b`, with the commit
message stating what it got wrong, and a `Co-authored-by:` to you since the probe, the
capture and the corrected mechanism are all yours.

On the consequence, which I think is the most important thing either of us has posted here:
the blast radius is not one newline per shifted syllable. It follows the render rate, so it
is unbounded and gets worse the busier the pane is. I have put that in the PR description
as the severity argument, quoting your PowerShell-vs-Claude-Code contrast.

Agreed that the chord owner is carrying none of the safety and the `code` gate is carrying
all of it. Worth someone looking at that effect's dependency array separately — not for this
branch, but a lock that does not survive a re-render will mislead the next person who
reasons about it. Happy to open that as its own issue if you would rather not.

dogfooter8750:
Disclosure on my previous comment, because one line in it rests on a non-default setting of mine and I would rather you heard it from me than found it later.

My Claude Code runs with `CLAUDE_CODE_NO_FLICKER=1`, which per its own help text selects "the flicker-free alt-screen renderer with virtualized scrollback" instead of the classic main-screen renderer. That changes the output pattern a pane produces, and Orca has alt-screen-specific handling in `pty-connection.ts` — v1.4.164 even shipped `fix(terminal): bound fullscreen atlas recovery (#12061)` for that path. So it is a real variable, not a cosmetic one.

What that does and does not touch:

**Unaffected.** The pty byte measurements were taken in a plain PowerShell pane with no Claude Code running at all — that is where `ESC [13;5u` and `ESC CR` were captured, and where I measured three injections per hold. The key-event traces are Windows/Chromium IME behaviour and do not involve the TUI. The composition-open trigger rule, the Ctrl variant, and the re-render finding itself were all established by driving the handler directly:

```text
no re-render during the hold          sendInput x1
re-render between each auto-repeat    sendInput x51
```

That is the mechanism, and it holds independently of any renderer.

**Affected.** Exactly one sentence: my description of a Claude Code pane producing "a continuous stream of newlines" while Shift is held. The injection rate follows the render rate, and the render rate depends on which renderer the TUI is using — so that observation was made under the alt-screen renderer, not the default one. Treat it as "this is what a busy pane does", not as a calibrated figure for stock Claude Code.

I will re-run the hold with the flag unset and post both numbers so the comparison is on the record. If the default renderer produces a lower rate, the qualitative point still stands, only smaller; if it produces a similar one, then the renderer mode is not a factor and the note can be dropped.

Nothing about the fix changes either way — the `code` gate stops these events upstream of all of it.

dogfooter8750:
Withdrawing the caveat I just posted: **the renderer flag was never actually in effect, so every measurement here was on the stock renderer.**

I had been invoking it as `CLAUDE_CODE_NO_FLICKER=1 claude`, which is bash syntax. My Orca panes run PowerShell, where a `VAR=value command` prefix is not env assignment at all — PowerShell parses it as a command name and errors out with `CommandNotFoundException`. So the variable was never set for any of those sessions.

Confirmed rather than assumed, on this machine:

- absent from `settings.json` and `settings.local.json`
- absent from the Windows user and machine environment
- absent from the PowerShell profile
- absent from the environment of the Claude Code process currently running in an Orca pane

So the one line I flagged stands as originally written: holding Shift mid-composition in a Claude Code pane produces a continuous stream of newlines, on the default main-screen renderer, with no non-default settings involved. Nothing else in my earlier comments was affected either way, and the plain-PowerShell pty measurements never involved Claude Code to begin with.

Apologies for the noise — I would rather have over-disclosed and retracted than left a hidden variable sitting under a measurement you were about to fold into the PR description.

dogfooter8750:
Root-caused the re-render amplifier, and I owe you a correction on how I phrased it, since you quoted it in the PR description.

**What I wrote:** "a re-render re-runs the effect, which releases the chord lock."

**What is actually true:** a re-render does *not* by itself re-run the effect — React only re-runs it when a dependency identity changes. Measured against the real hook, holding Shift across 50 auto-repeats during composition:

```text
no re-render                                      1 send
re-render, dependency identities held stable      1 send
re-render, dependencies rebuilt as TerminalPane rebuilds them   50 sends
```

The middle row is the one my earlier wording would have predicted wrong. So the claim holds for this codebase, but only because of a specific unmemoized value — not as a general property of re-renders, and anyone reading it as the latter would draw the wrong conclusion.

**The specific cause:** `TerminalPane.tsx` (~line 1246) calls `createExpandCollapseActions({ ... })` in the render body with no `useMemo`. It is a plain factory (`expand-collapse.ts:115`), not a hook, and returns five fresh closures per call. Four of them — `setExpandedPane`, `restoreExpandedLayout`, `refreshPaneSizes`, `toggleExpandPane` — are in the keyboard effect's dependency array (`keyboard-handlers.ts` ~912-916). Two calls with an identical state object give new identities for all five, which I verified rather than assumed. `TerminalPane` has `forwardRef` but no `React.memo`, so it re-renders whenever its parent does.

Your instinct to look at that dependency array separately was right. I filed it as #12269 with the evidence and a suggested fix, rather than leaving it as an aside here — it also means seven `window` listeners are removed and re-added on every render of every mounted pane, which is worth fixing independently of IME.

None of this changes #12265 or this branch: both stop these events being read as Enter upstream of the claim, so the lock's weakness is latent again rather than reachable.

Congratulations on the landing, and thanks for the care you took with attribution and with checking my claims before adopting them — the corrected mechanism only exists because you refused to take the first version on trust.

holdn2:
Closing this — #12265 landed the overlap, and a combined branch has no reason to exist once
the combination is on `main`.

That PR carries the same work with the same authorship: @kunsanglee's and @nwparker's
keydown predicate and shortcut guard, and @KimYoungHwan8750's keyup evidence plus the
per-press follow-up. The xterm commit-ordering and cancel-cleanup work from @prxyeo's
#11870 landed separately. Between them, both symptoms this PR was opened for — the stray
newlines and the syllable-late commits — are closed on `main`.

One piece is not: the deferred-newline fallback re-arm from #11896. `main` still has

```ts
const fallbackTimer = window.setTimeout(finish, fallbackMs)
```

with no pending-composition re-check, so a composition still open at 200 ms can be split.
I have flagged it on #11896 rather than keep this branch open for a single hunk that
already has its own PR and its own author.

Thanks to everyone who worked on this — and particularly to @KimYoungHwan8750, whose pty
capture and device event probe turned most of what was being inferred here into something
measured, and who corrected his own findings twice in public when they turned out to be
imprecise. I got the amplifier mechanism wrong along the way by matching on a count rather
than checking whether the hardware emitted that sequence at all; his probe is what settled
it, and #12269 is where the root cause now lives.

Nothing here needs to be carried over. Closing.

PR #12121, opened 2026-08-02
test(shared): cover the commit-message host-key normalizers

YuriNachos:
## Summary

`src/shared/commit-message-host-key.ts` exports two pure normalizers that classify where a commit-message model discovery request originates, plus three sentinel constants:

- `getCommitMessageModelDiscoveryHostKey(connectionId)` — `undefined` → `'unknown'`; `null`/`''` → `'local'`; otherwise `` `ssh:${connectionId}` ``.
- `getCommitMessageModelDiscoveryHostKeyForScope(scope)` — `undefined` → `'unknown'`; `null`/`''` → `'local'`; a `runtime:`-prefixed scope passes through unchanged; otherwise it delegates to the connection-id resolver as an `ssh:` key.
- `LOCAL_COMMIT_MESSAGE_HOST_KEY` (`'local'`), `UNKNOWN_COMMIT_MESSAGE_HOST_KEY` (`'unknown'`), `RUNTIME_COMMIT_MESSAGE_HOST_KEY_PREFIX` (`'runtime:'`).

These feed commit-message model discovery routing across the renderer (`source-control-ai.ts` — where `LOCAL_COMMIT_MESSAGE_HOST_KEY` is compared 7× — `runtime-git-client.ts`, `use-source-control-ai.ts`, `native-chat-session-option-discovery.ts`) and the main process (`commit-message-text-generation.ts`, `orca-runtime-git.ts`, `first-work-branch-rename.ts`, `filesystem.ts`), but had **no direct unit tests** and zero indirect references (grep across `*.test.ts` returns nothing). The module has zero imports of any kind — fully pure.

This PR adds direct, isolated unit tests mirroring the co-located `branch-prefix.test.ts` sibling: one `describe` per export, exact-value assertions covering the three-state contract, the runtime-prefix passthrough, and the runtime-prefix *boundary* (a non-leading `runtime:` and a bare `runtime` token both route to `ssh:`, pinning `startsWith` rather than `includes`).

## Screenshots

No visual change — test-only.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`

Passed:
- New `src/shared/commit-message-host-key.test.ts` — 12 tests, one `describe` per export (+ sentinels).
- **Mutation-checked** (each logical branch proven to have teeth, not for-show):
  - collapsing the `=== undefined` guard to `!connectionId` → **1 test fails** (a `null`/`''` connection id would wrongly classify as `'unknown'`).
  - dropping the `ssh:` prefix → **4 tests fail** (every ssh-key case).
  - removing the `runtime:`-passthrough branch → **1 test fails** (a runtime scope would be re-wrapped as `ssh:`).
  - `startsWith` → `includes` → **1 test fails** (a non-leading `runtime:`, e.g. `foo:runtime:bar`, would wrongly passthrough) — this is the one `startsWith`-regression invisible to the prefix-only cases.
  - `startsWith('runtime:')` → `startsWith('runtime')` (drop the colon) → **1 test fails** (a bare `runtime` token would wrongly passthrough).
  - reverted: **12/12 green**.
- `pnpm typecheck` green across all three tsconfigs (node, tc.cli, tc.web).
- `pnpm exec oxlint` clean on the new file (no rule suppression added).
- `git diff --stat` = exactly one new file (+75), zero production changes.

## AI Review Report

Pure additive test coverage — one new `*.test.ts` file, zero production edits. Reviewed for:
- **Cross-platform (macOS/Linux/Windows):** the SUT is host-routing string logic with no platform assumptions; tests assert exact string outputs and run identically on all platforms. No platform-specific code introduced.
- **SSH/remote/local + folder-workspace:** N/A — pure unit tests on a zero-import module.
- **Provider/agent neutrality:** N/A — commit-message host classification, not agent or provider behaviour.
- **Performance / hot path:** none — test-only, runs once per suite.
- **UI quality:** N/A.

## Security Audit

Test-only change: adds assertions on a pure, zero-import module. No code/IPC/auth/path/dependency surface, no new input handling, no untrusted data.

## Notes

- The two normalizers are the single source of truth for the `'local'` / `'unknown'` / `'ssh:<id>'` / `'runtime:<…>'` host keys consumed by commit-message model discovery — a regression (collapsing the undefined guard, dropping the `ssh:` prefix, or losing the runtime passthrough) would silently mis-route discovery across local/SSH/runtime hosts. These tests lock that contract directly.
- No open or closed PR adds a test for this module (`gh pr list --search "commit-message-host-key"` returns 10 results, all unrelated commit/git/mobile/bitbucket feature work — none add this test).

### DIFF
--- a/src/shared/commit-message-host-key.test.ts
+++ b/src/shared/commit-message-host-key.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, it } from 'vitest'
+import {
+  LOCAL_COMMIT_MESSAGE_HOST_KEY,
+  RUNTIME_COMMIT_MESSAGE_HOST_KEY_PREFIX,
+  UNKNOWN_COMMIT_MESSAGE_HOST_KEY,
+  getCommitMessageModelDiscoveryHostKey,
+  getCommitMessageModelDiscoveryHostKeyForScope
+} from './commit-message-host-key'
+
+describe('commit message host-key sentinels', () => {
+  it('locks the three exported sentinel values', () => {
+    expect(LOCAL_COMMIT_MESSAGE_HOST_KEY).toBe('local')
+    expect(UNKNOWN_COMMIT_MESSAGE_HOST_KEY).toBe('unknown')
+    expect(RUNTIME_COMMIT_MESSAGE_HOST_KEY_PREFIX).toBe('runtime:')
+  })
+})
+
+describe('getCommitMessageModelDiscoveryHostKey', () => {
+  it('maps an undefined connection id to the unknown key', () => {
+    expect(getCommitMessageModelDiscoveryHostKey(undefined)).toBe(UNKNOWN_COMMIT_MESSAGE_HOST_KEY)
+  })
+
+  it('maps a null or empty connection id to the local key', () => {
+    expect(getCommitMessageModelDiscoveryHostKey(null)).toBe(LOCAL_COMMIT_MESSAGE_HOST_KEY)
+    expect(getCommitMessageModelDiscoveryHostKey('')).toBe(LOCAL_COMMIT_MESSAGE_HOST_KEY)
+  })
+
+  it('prefixes a non-empty connection id as an ssh key', () => {
+    expect(getCommitMessageModelDiscoveryHostKey('conn-1')).toBe('ssh:conn-1')
+  })
+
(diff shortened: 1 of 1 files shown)


YuriNachos:

PR #12122, opened 2026-08-02
fix(terminal): recover blackholed PTY input

wongang2:
## Summary

- recover renderer and IME input that never reaches xterm
- require observable PTY delivery before terminal send reports acceptance
- verify current-daemon PTY input with a real write/read round trip before reuse
- quarantine a failing daemon without killing preserved sessions and route fresh terminals to the local fallback immediately
- reclaim an older daemon generation only when its authenticated `shutdownIfIdle` decision proves it has no sessions or competing client
- preserve live or unverifiable legacy generations and their session ownership

## Root cause and field evidence

Installed Orca 1.4.163 had protocol v29 and v30 daemons still running after repeated desktop-app restarts. They had survived for roughly 4 days 15 hours and 1 day 17 hours. `terminal.send` returned `accepted: true` and `bytesWritten: 27`, but the shell never printed the marker.

With explicit operator approval, both stale generations were terminated. The v29 daemon exited on TERM; the v30 daemon ignored TERM and required KILL. With one fresh daemon left, the unchanged installed 1.4.163 build immediately passed both an ASCII marker command and a Korean input command.

This narrows the incident trigger to a stale PTY-owning daemon surviving app restarts while generations coexist, rather than the renderer or installed app code alone. The previous ACK-without-delivery behavior still hid the failure and made the incident persistent.

The repaired flow now covers both sides:

- a current daemon that fails the input round trip is not used for fresh terminals; existing sessions are preserved in degraded mode
- the immediately previous generation is input-probed and quarantined if delivery cannot be proven
- sessionless clean-disconnect legacy generations are retired during app startup
- live, busy, pre-v24, malformed, timed-out, or otherwise unverifiable legacy generations are preserved fail-closed

## Tests

- 21 OPS-1253-related test files pass after merging the latest `main`
- daemon-generation coexistence covers idle v29 retirement alongside live v30 preservation
- protocol boundary coverage includes v23 bypass, v24 first support, refusal, malformed response, authentication failure, and request timeout
- full typecheck passed
- native and type-aware changed-code checks passed with 0 findings
- max-lines ratchet passed with no new bypasses
- desktop build passed
- independent review: clear/approve, no findings

## Live status

The unchanged installed 1.4.163 build is currently healthy after the stale v29/v30 daemons were removed and a fresh daemon was created. This confirms the field trigger, but it is not a live verification of this branch. No further daemon restart or termination was performed.

A manufacturer-signed ad-hoc macOS build can be produced from this PR for live verification if needed; no build workflow was dispatched and no app was installed as part of this change.


Jinwoo-H:
Thanks @wongang2 for the careful field investigation and the substantial test coverage here.

On current main `473442865413787443f47ff6e08059ede241624f`, our three-way real-app oracle used actual headless Electron → DOM/xterm → a local protocol-v32 daemon → a real shell PTY, with PTY-identity and shell-marker checks. Main and the disabled control `80f219a613f28fcb795023e0d5181694bfb4d4a0` were 0/2: no remount and the input was lost. The rebased candidate `fd2e534a4cd4e45377d8a97f59524c155a036dbc` was 1/2: it remounted the same PTY and later input worked, but the exact command bytes typed during the DOM-to-xterm blackhole were permanently absent.

That restores availability, but it does not provide lossless input delivery. The renderer watcher retains intent rather than the authoritative encoded bytes, so a remount cannot reconcile or replay them. The current and candidate trees also both advertise protocol v32 while assigning new semantics to older generations; an exited/disposed session can still acknowledge a no-op write; every strong write forces a separate `/bin/sh` health PTY that cannot prove delivery to the target PTY and can quarantine on ambiguous failures; Windows bypasses that round trip; and direct SSH/paired-runtime ownership is not covered.

Because those are delivery-contract issues rather than a bounded patch, this cannot land safely in its current form. A replacement or reopening would need a newly advertised protocol capability/version; session-, incarnation-, and sequence-scoped exact-byte acknowledgement/reconciliation with exactly-once behavior; tri-state health handling that never quarantines an unknown result; and a durable cross-provider Electron red/green gate for ASCII, IME, and paste across the local daemon, native Windows, direct SSH, paired/remote runtime, and folder workspaces, proving that bytes typed during the blackhole reach the same PTY exactly once. Closing this unmerged; thank you again for surfacing the stale-daemon trigger and pushing the investigation forward.

Issue #11714, opened 2026-07-31
Mobile pairing via Relay fails: relay_control_not_active


CmcWakita:
## Description

Mobile pairing via Orca Relay fails when trying to pair the desktop app with the mobile companion app. The desktop shows:

> Couldn't create a Relay pairing code. Retry, or use LAN to pair over Tailscale or the same Wi-Fi.

Relay is marked as unavailable in the connection mode selector.

## Diagnostics (from "Copy diagnostics")

```json
{
  "kind": "mobile_pairing_relay_failure",
  "preferredConnectionMode": "automatic",
  "failure": {
    "code": "relay_control_not_active",
    "stage": "create_pairing_relay",
    "message": "Relay pairing invite request failed"
  },
  "at": "2026-07-31T06:44:14.585Z"
}
```

## Expected behavior

Relay pairing should succeed, or at least prompt for sign-in if that's required to activate the relay control channel.

## Workaround

Pairing over LAN (same Wi-Fi) works fine.

## Environment

- Desktop OS: Windows
- Connection mode: automatic (Relay)

shenghsi:
I see same issue.

florianschaueramaikoai:
same for me

qktlf789456:
I'm experiencing the same issue, any fix?

AmethystLiang:
This should be fixed now.

AmethystLiang:
@qktlf789456 @florianschaueramaikoai can you try out if it works on the latest versions. 

it should work now

tegarkurniawan:
@AmethystLiang I just downloaded Orca for Mac and iOS, but this feature isn't working yet.
{ "kind": "mobile_pairing_relay_failure", "preferredConnectionMode": "automatic", "failure": { "code": "relay_control_not_active", "stage": "create_pairing_relay", "message": "Relay pairing invite request failed" }, "at": "2026-08-10T02:20:20.405Z" }

itizarsa:
@AmethystLiang I faced this issue and still getting the following after update

```json
{
  "kind": "mobile_pairing_relay_failure",
  "preferredConnectionMode": "automatic",
  "failure": {
    "code": "relay_control_not_active",
    "stage": "create_pairing_relay",
    "message": "Relay pairing invite request failed"
  },
  "at": "2026-08-10T07:08:08.299Z"
}
```

evandromassa:
same issue here.

{
  "kind": "mobile_pairing_relay_failure",
[...]

PR #12134, opened 2026-08-02
fix(relay): refresh stale capabilities before pairing

### LINKED ISSUE #11871: [Bug]: Stale Orca Cloud capabilities show signed-in but Relay offline; no in-app re-sign-in path in packaged builds
### Operating system
Windows 11

### Orca version
v1.4.163 (latest stable, includes the #10147 retry + #10260 handshake-deadline fixes)

### Summary

After the Orca Cloud session/capabilities go stale, the desktop still shows the profile as **signed in / connected**, but **Relay stays unavailable** with `relay_control_not_active` and **zero broker activity**. There is **no in-app way to recover** in a packaged production build: the Mobile page shows no "Sign in" button (because the cached auth state is `connected`), and the profile/account menu that holds "Sign in again" / "Sign out" is hidden in packaged builds. The only recovery is to manually delete the encrypted session file and re-sign in — after which Relay works immediately, proving the account is entitled.

### Reproduction

1. Sign in to an Orca cloud-linked profile (account is Relay-entitled server-side).
2. At some point the cached capabilities become stale such that `relay.use` reads `false` locally (see root cause).
3. Open Settings → Mobile → select **Orca Relay** → generate a pairing code.
4. Observed: Relay marked **Unavailable**; diagnostics show:
   ```json
   {
     "kind": "mobile_pairing_relay_failure",
     "preferredConnectionMode": "automatic",
     "failure": { "code": "relay_control_not_active", "stage": "create_pairing_relay", "message": "Relay pairing invite request failed" }
   }
   ```
5. The auth/profile UI still shows **Signed in** (not "reconnect-required"), so the Mobile connection-options panel does **not** render the "Sign in / Sign in again" button (it's gated on `!signedIn && configured`).
6. There is no other production UI to sign out or force a re-sign-in (profile switcher hidden, DevTools Orca Cloud section dev-only).
7. **Workaround that fixed it:** quit Orca, delete `%APPDATA%\orca\profiles\<profileId>\account-session.json.enc` and `account-session-mutation.json`, restart Orca → profile shows `reconnect-required` → "Sign in again" appears → complete OAuth → Relay works.

### Root cause

Two compounding problems:

**1. Stale capabilities gate Relay with no refresh/recovery.**
`relay-auth-context.ts:32`:
```ts
relayEntitled: session.session.capabilities.flags['relay.use'] === true
```
`relay-auth-coordinator.ts:121`:
```ts
if (!context || !context.relayEntitled) {
  this.invalidateOwnership()
  this.options.onStatus('offline')   // no broker, no retry, forever
  return
}
```
When the locally-cached capabilities have `relay.use !== true`, the coordinator goes `offline` immediately and never attempts broker setup — so there is **no broker/origin/control activity in the logs at all** (confirmed: zero `relay-control`/`relay-origin`/`relayEntitled`/`capabilities` spans in `main.trace.ndjson` / `daemon.log`). This is *not* the transient broker-setup failure that #10147 / #10260 addressed; those fixes never engage because setup is never attempted.

Meanwhile `orcaProfileAuthStatus.state` stays `connected` (the auth-state machine reflects the session token, not the Relay entitlement), so the UI keeps showing "Signed in."

**2. No production re-sign-in path.**
- `OrcaProfileSwitcher.tsx`: `if (import.meta.env.PROD && import.meta.env.MODE !== 'e2e') return null;` — the entire account menu (incl. "Sign in again" / "Sign out") is hidden in packaged builds ("the Orca Cloud account UX isn't ready for production users yet").
- `DevToolsPane`'s `OrcaCloudDevSubsection` is dev-only ("Hidden in production").
- The only production-accessible sign-in trigger is `MobilePairingConnectionOptions`' "Sign in" button, which only renders when `!signedIn && configured`. Since the stale state reports `signedIn === true`, the button never appears.

Result: a stale-entitlement state is unreachable from the UI — the user must edit the filesystem.

### Evidence

- **Orca**: v1.4.163, Windows 11, profile `local-default` (cloud-linked).
- **Relay servers healthy**: `https://relay.onorca.dev/health` → 200, `https://c2.relay.onorca.dev/health` → 200, `https://login.onorca.dev` reachable.
- **Logs**: no relay-broker activity at all in `%APPDATA%\orca\logs\main.trace.ndjson` / `daemon.log` (no `relay-control`/`relay-origin`/`relayEntitled`/`capabilities`/`control_not_active` spans) — consistent with the coordinator stopping at the entitlement gate before any setup.
- **Recovery**: clearing `account-session.json.enc` + `account-session-mutation.json` and re-signing in restored Relay immediately — confirming the account **is** Relay-entitled server-side; the local capabilities were stale.

### Expected behavior

At least one of:
1. When the relay coordinator goes `offline` due to `!relayEntitled`, surface a **"Reconnect / Sign in again"** action on the Mobile pairing failure panel (not just "Use LAN" / "Retry Relay"), since re-auth may refresh capabilities.
2. Proactively **refresh capabilities** when `relay.use` is false but the session is otherwise `connected`, rather than caching stale `false` indefinitely.
3. Provide a **production-accessible sign-out / re-sign-in** path (the profile menu is intentionally hidden; until it ships, the Mobile page is the only prod entry point and it currently hides the button when `signedIn`).

### Workaround

Quit Orca, then from a terminal:
```bash
cd "%APPDATA%\orca\profiles\<profileId>"
del account-session.json.enc account-session-mutation.json
```
Restart Orca → the profile shows `reconnect-required` → "Sign in again" appears on the Mobile page → complete OAuth → Relay works.

### Related

- #11714 — the `relay_control_not_active` failure symptom (this issue explains the stale-entitlement + no-recovery cause and the workaround).
- #10273 — Relay stays unavailable after a transient desktop broker failure (transient setup path; different failure mode — that one has broker activity, this one has none).
- #10147 / #10260 — transient broker-setup retry + handshake deadlines (do not engage here).

X handle: @itisvincent

aaraujoe:
## Summary

- refresh Orca Cloud capabilities before treating cached `relay.use=false` as authoritative
- single-flight and rate-limit capability refreshes, with cooldown eviction after 60 seconds
- offer a production-accessible **Sign in again** action when Relay remains unavailable despite a connected session
- explicitly reset refresh/cooldown state between tests and cover same-profile-key reuse

## Screenshots

- Not attached. The changed UI is the compact **Sign in again** recovery row shown only after a Relay mint failure; it is covered by the component test and the live macOS validation linked under Notes.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — focused tests were run instead:
  - `pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/relay/relay-auth-context.test.ts src/renderer/src/components/settings/MobilePairingConnectionOptions.test.tsx` (15 tests)
- [ ] `pnpm build` — the branch was previously launched successfully with `pnpm dev` for live Relay validation
- [x] Added or updated high-quality tests that catch stale capability recovery, concurrent refresh fan-out, cooldown behavior, explicit module-state isolation, and the signed-in recovery UI

The local runtime is Node 26 while the repository declares Node 24; the commands above completed successfully with only the engine warning.

## AI Review Report

Reviewed the capability-refresh boundary, single-flight ordering, cooldown behavior, profile/session re-reads, renderer recovery conditions, and test isolation. Greptile flagged an unpruned cooldown map and implicit module-state isolation; both are addressed by cooldown eviction and `resetRelayCapabilityRefreshStateForTests()` with regression coverage.

Cross-platform review found no platform-specific shortcuts, labels, paths, shell commands, or Git behavior in this change. The refresh state uses Node/Electron timers available on macOS, Linux, and Windows. SSH and folder-workspace execution are unaffected because this authorization state belongs to the desktop Relay service rather than a workspace execution host.

## Security Audit

- Reuses the existing authenticated capability-refresh service; no new token storage, logging, or transport is introduced.
- Re-reads the active profile and encrypted session after refresh, preserving the existing identity-change guard.
- Adds no dependencies, commands, filesystem paths, external input handling, or new IPC surface.
- The recovery button invokes the existing OAuth connection action and does not expose credentials in the renderer.

## Notes

Relay activation previously trusted capabilities cached alongside an otherwise fresh session. A stale `relay.use=false` stopped the coordinator before broker creation while the renderer still reported the profile as connected and hid its only production reauthentication action.

Live macOS validation confirmed that this patch proceeds past the entitlement gate after capability refresh. The production Relay director then returned `relay_assignment_failed_503`, which is a separate backend availability problem currently masked by the outer `relay_control_not_active` diagnostic. Evidence: https://github.com/stablyai/orca/pull/12134#issuecomment-5160104787

Fixes #11871

Related: #11714, #12129


aaraujoe:
Live macOS development validation (2026-08-02):

- Launched this branch against the production Orca Cloud/Relay endpoints with isolated `orca-dev` user data.
- Reproduced the connected-session + unavailable-Relay state and confirmed the new **Sign in again** recovery action is visible.
- Completed OAuth; `orcaProfiles.authStatus()` then reported an encrypted connected session with `relay.use: true` and a freshly updated capability timestamp.
- The patched client proceeded past the entitlement gate and attempted broker assignment. The main-process evidence is now:
  ```
  [relay] broker reconcile failed: relay_assignment_failed_503
  [runtime] Failed to create Relay pairing invite: relay_control_not_active
  ```

This validates that the patch fixes the stale-capability/no-recovery client path. It also reveals a concurrent director/backend availability failure: the generic user-facing `relay_control_not_active` currently masks an upstream assignment HTTP 503. That 503 cannot be resolved by capability refresh and should be tracked operationally or in a follow-up diagnostic/error-propagation change.

PR #6408, opened 2026-06-26
fix(browser): keep webview guests alive across worktree switches

Jingyue-Wu:
Electron `<webview>` guests are destroyed when their DOM parent is removed. When switching worktrees, `BrowserPane` chrome unmounts and previously tore down the webview, forcing a reload on return.

Introduce persistent page viewports inside overlay slots so webviews stay in a stable parent without keeping the full React chrome mounted. On worktree switch, park hidden viewports instead of destroying guests; explicit close paths still call `destroyPersistentWebview`.

Fixes stablyai/orca#6385

## Summary

- **Fix:** Browser tabs no longer reload when switching between worktrees/projects. A page you left open (e.g. a long-running QA session) keeps its loaded document, scroll position, and in-page state when you switch away and back.
- **How:** Each browser page gets a persistent viewport shell under its worktree-level overlay slot. The `<webview>` guest lives in that stable DOM parent and is registered in the existing `webviewRegistry`. When `BrowserPane` chrome unmounts on worktree switch, the viewport is **parked** (`display: none`) instead of destroyed. React chrome (toolbar, address bar, find bar, overlays) continues to mount/unmount normally via `shouldMountPane`.
- **Why not “keep everything mounted”:** We deliberately do **not** keep the full `BrowserPane` React subtree mounted for inactive worktrees. That caused performance issues (React reconciliation, event listeners, layout work for every hidden browser). Only the lightweight viewport shell + webview guest persist; the heavy chrome unmounts.
- **Performance tradeoff:** Inactive worktrees still hold live webview guests (same memory/compositor cost as an open tab in a normal browser). We avoid the extra cost of repeatedly destroying/recreating guests and re-running full page loads. Hidden guests are parked and non-interactive (`display: none`, `inert`, `pointer-events: none`).

## Screenshots

Behavior change only, same browser UI as before. Verified on macOS that pages render correctly after the fix (Google homepage and loaded sites display in the browser pane; no blank/white regression).

No visual change to layout or chrome.

image

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — browser-pane suite: 112 tests passed (`src/renderer/src/components/browser-pane`)
- [x] `pnpm build` — not run end-to-end in this session
- [x] Added or updated high-quality tests that would catch regressions, or explained why tests were not needed

**Manual verification (macOS, local dev):**

- Open browser tab, load a real page (google.com / custom site).
- Switch to another worktree, then switch back.
[...]

PR #12137, opened 2026-08-02
fix(browser): keep browser tabs rendering across worktree switches (STA-3228)

brennanb2025:
## Summary

Fixes STA-3228: a browser tab that was open in a worktree went permanently blank after switching to another worktree and back — the URL bar still showed the address, but the page area rendered nothing, and reload silently did nothing.

Two layered defects:

1. **Trigger** — `WorktreeSplitSurface` unmounted `BrowserPaneOverlayLayer` for hidden worktrees whenever a targeted background mount restriction existed (agent wake/resume sets one constantly). Unmounting the overlay pulls the persistent `<webview>` slot divs out of the DOM, and Electron destroys the guest the moment a `<webview>` leaves the DOM — defeating the design from #6408 that renders browser panes at worktree level precisely so guests survive switches.
2. **Permanence** — `ensureBrowserPageViewport` returned the cached viewport shell even after its slot root had been unmounted, so when the worktree came back, the webview subtree was never re-appended to the live DOM. The tab stayed blank forever and the reload button threw `The WebView must be attached to the DOM and the dom-ready event emitted before this method can be called` (observed in the renderer console during repro).

Fix:

- Keep `BrowserPaneOverlayLayer` mounted for hidden worktrees. Slots already park their pane subtrees when the worktree is hidden (`shouldMountPane`), so a background-woken worktree only pays for empty slot divs. `EmulatorPaneOverlayLayer` keeps the old gating — it doesn't use the persistent-slot pattern.
- Make `ensureBrowserPageViewport` detect a replaced slot root and drop the stranded shell so the caller rebuilds against the live root. This also self-heals any remaining path that unmounts slots (cross-worktree tab moves, surface eviction): the existing parent-drift check in `ensureBrowserPageWebview` then rebuilds the webview, turning "blank forever" into "reloads once".

## Screenshots

Repro on main (tab created, worktree switched away and back — blank, reload dead):

before: blank browser pane after worktree switch

Same sequence with the fix (guest survives without reloading — same in-page JS timestamp, no new HTTP request; reload button works):

after: page still rendered after worktree switch

## Testing

- [x] `pnpm lint` (oxlint + react-doctor changed-lines on touched files)
- [x] `pnpm typecheck`
- [x] `pnpm test` (browser-pane suite 32 files/191 tests; `Terminal` substring sweep 462 files/5156 tests)
- [ ] `pnpm build`

New tests in `browser-page-viewport.test.ts`: cached-viewport reuse under a stable root, parked viewport preserved while the root is unregistered, and rebuild under a replacement root (asserts the stale shell is detached and the rebuilt shell is connected to the new root).

Live Electron QA (dev build, CDP): reproduced the blank pane + dead reload on unfixed code, then verified with the fix that (a) slot/shell/webview stay in the DOM while the worktree is hidden, (b) the page renders after switching back with the guest's state intact (unchanged in-page timestamp, no new server request), (c) reload issues a real request, and (d) a session-restored tab renders after app restart.

## AI Review Report

Reviewed the mount-gating change for perf regressions (background wakes previously skipped the overlay): slots park pane subtrees for hidden worktrees, so no webview guests or BrowserPane chrome mount from this change alone. Reviewed the viewport self-heal for interactions with `destroyPersistentWebview` (which removes the current viewport before `resolveContainer` rebuilds it) and the floating-panel slot registration (`FloatingBrowserSlot` mints new tab ids, so no cross-host root swap for a live id). Checked the no-root case keeps returning the parked viewport so hidden-state callers see unchanged behavior. Cross-platform: renderer-only DOM/React change, no shortcuts, labels, paths, or shell behavior touched; nothing platform-conditional on macOS/Linux/Windows is affected.

## Security Audit

No input handling, command execution, path handling, auth, secrets, or dependency changes. No new IPC: the change only alters when existing renderer components mount and how a renderer-side DOM cache invalidates. Webview partitions are untouched (partition mismatch still forces a rebuild via the pre-existing check in `ensureBrowserPageWebview`).

## Notes

- The surface-eviction path (a hidden worktree fully unmounting after long parking) still destroys guests by design; with the self-heal the tab now recovers with one reload instead of dying. If we want guests to survive that too, that's a follow-up.
- `EmulatorPaneOverlayLayer` may have an analogous issue on the same conditional; left untouched to keep this PR scoped (worth a separate look).

### DIFF
--- a/src/renderer/src/components/Terminal.tsx
+++ b/src/renderer/src/components/Terminal.tsx
@@ -45,7 +45,7 @@ import type {
 } from '../../../shared/types'
 import { hasFeatureInteraction } from '../../../shared/feature-interactions'
 import BrowserPane from './browser-pane/BrowserPane'
-import BrowserPaneOverlayLayer from './browser-pane/BrowserPaneOverlayLayer'
+import { RetainedBrowserPaneOverlayLayer } from './browser-pane/BrowserPaneOverlayLayer'
 import EmulatorPaneOverlayLayer from './emulator-pane/EmulatorPaneOverlayLayer'
 import { useBrowserAutomationVisibilityForAny } from './browser-pane/browser-automation-visibility'
 import { useBrowserMobileDriverForAny } from '@/lib/pane-manager/browser-mobile-driver-state'
@@ -2674,8 +2674,17 @@ const WorktreeSplitSurface = React.memo(function WorktreeSplitSurface({
         backgroundMountTabIds={backgroundMountTabIds}
         activationDeferredMountTabIds={activationDeferredMountTabIds}
       />
-      {/* Why: always mounted — its slots host persistent <webview> guests that die if the slot leaves the DOM (STA-3228); hidden worktrees park the pane subtree, so targeted background mounts only pay for empty slot divs. */}
-      <BrowserPaneOverlayLayer worktreeId={worktreeId} isWorktreeActive={isVisible} />
+      {/* Why: once eligible, retain slot DOM so hidden worktrees keep their Electron guests alive (STA-3228). */}
+      <RetainedBrowserPaneOverlayLayer
+        worktreeId={worktreeId}
+        isWorktreeActive={isVisible}
+        mountEligible={
+          isVisible ||
+          backgroundMountTabIds === null ||
+          hasAutomationVisibleBrowser ||
+          hasMobileDrivenBrowser
+        }
+      />
       {isVisible || backgroundMountTabIds === null ? (
         <EmulatorPaneOverlayLayer worktreeId={worktreeId} isWorktreeActive={isVisible} />
       ) : null}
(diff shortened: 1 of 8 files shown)


brennanb2025:
## Electron QA — STA-3228 (b2dc29c6d6)

Validated the final pushed revision in an isolated Orca dev profile attached to the intended worktree through CDP.

- Opened the deterministic page in Orca, switched from this worktree to main through the rendered worktree search UI, then returned through the same real UI.
- Before switching: server document load #1 and guest performance.timeOrigin 1785708162108.6.
- While hidden: the viewport shell was parked with display:none, while both the shell and webview remained DOM-connected.
- After returning: the pane rendered normally with the same load #1 and the same time origin; the server received no second document request.
- Clicked Orca’s real reload control after returning: the server received document load #2 and the page visibly rerendered. Renderer console: 0 errors.

### Rendered after switching away and back

STA-3228 browser pane rendered after worktree return

### Reload still works after the return

STA-3228 browser pane after working reload

Mobile emulator QA is not applicable: this PR only changes desktop renderer/browser-webview paths.

Issue #11360, opened 2026-07-29
[Feature]: Make browser annotation feedback compact by default


lukagubo:
## Problem or use case

Browser design annotations send far more diagnostic context than a normal feedback turn needs. A one-sentence comment on one paragraph currently expands to roughly 38 lines of Markdown.

The useful context for most changes is:

* page file or URL
* element label and selector
* selected text or element text
* intent
* feedback

The current output also includes the browser tab ID, viewport, bounds, classes, nearby text and elements, computed styles, full DOM path, and an HTML snippet for every annotation. That context is useful when debugging a layout or selector problem, but it consumes agent context and makes the user's comments harder to scan during ordinary document review.

Example from a local HTML report in Orca 1.4.159:

```markdown
## Design Feedback: /Users/me/report.html

**URL:** file:///Users/me/report.html
**Browser tab id:** page-id
**Viewport:** 2279x1296

### 1. p "Recommended locking rule"
**Intent:** change
**Selector:** `p#locking-rule-copy`
**Location:** `main > #locking-recommendation > #locking-rule-copy`
**Bounds:** x=829, y=476, 622x58
**Text:** "Final run approval is the only irreversible point."
...
**Computed styles:**
...
**Full DOM path:** `body > main > article#locking-recommendation > p#locking-rule-copy`
**HTML:**
...
**Feedback:** Explain why this should be irreversible.
```

Adding `id="locking-rule-copy"` and `data-review-id="locking-rule-copy"` to the report made the selector cleaner, but did not reduce the payload. The report cannot control Orca's export format.

This request should not change how annotations work. Users should still be able to comment on any element in a page, not only predefined decision blocks.

## Proposed solution

Make the compact format the default for both Send and Copy, with full diagnostics available through an explicit `Include diagnostic context` or `Copy with diagnostics` action.

Suggested compact output:

```markdown
## Design feedback
**Page:** file:///Users/me/report.html

### 1. p "Recommended locking rule"
**Intent:** change
**Selector:** `p#locking-rule-copy`
**Text:** "Final run approval is the only irreversible point."
**Feedback:** Explain why this should be irreversible.
```

Suggested implementation on current `main` (`4543bb68263a89ab520cea62ca69d7ac78330dd3`):

[...]

Issue #8687, opened 2026-07-14
[Feature]: Keep Annotations Across Page Navigation


PokIsemaine:
### Problem or use case

When switching between different pages, the annotations from the previous page are currently being cleared. 

### Proposed solution

We'd like those annotations to be preserved.

PR #12141, opened 2026-08-02
feat(browser): point a Design Mode annotation at a second element or spot

praveenpke:
## ELI5

When leaving design feedback in Orca's built-in browser, a note could only point at one element on the page. Instructions involving two things ("move this button below the pricing table") forced you to describe the second item in words, so the agent often changed the wrong spot. The annotation box now has a crosshair picker: click another element (or an empty spot) and an @ref1 tag drops into your comment, giving the agent exact locations for every reference.

## Summary

A Design Mode annotation can only describe the one element it is attached to. But most instructions
in a UI review are about two things — "move this button below the plan comparison", "match the width
of that card", "this overlaps the sticky header". The second target has nowhere to live, so it
reaches the agent as prose while the first arrives as an exact selector, and the second one is what
decides where the change actually lands.

This adds a picker to the annotation composer. Point at a second target and an `@ref1` token drops in
at your caret, so you keep writing around it.

Picking is position-aware: a click that lands on content references that element, while a click that
misses every child of its container is a **spot**, carrying its offset and ratio inside that
container plus the container's size. That makes "move it here" expressible for empty space, which
previously had no representation at all.

https://github.com/user-attachments/assets/6135408f-0eb6-4e6a-95b4-c9cd0714110d

Annotation composer showing one element reference and two spot references as chips

## Motivation

The annotation model is single-target: `BrowserPageAnnotation` holds one `payload.target`, and
`useGrabMode` clears the current payload before awaiting the next selection. So relational feedback
degrades to English at exactly the point where precision matters.

The prompt now resolves every token:

```
**References:**
- @ref1 — p "See the critical path without maintaining a second plan" — `article > .feature p` (at 120,640 900x64)
- @ref2 — empty space inside `article.feature`, not an element: 17% across and 81% down that
  container, at offset x=52, y=254 within its 306x314 box (page x=372, y=1180)
**Feedback:** keep this div bg color same as this @ref1, also change this div @ref2 color to blue
```

A spot names its container as context while stating that the container is not the target — without
that, an agent edits the container instead of placing something into it.

## Changes

- New `browser-annotation-references.ts`: token construction, label derivation, append/remove with
  renumbering, persisted-state clamp, caret-aware insertion, and unresolved-token stripping.
- Guest picker reports the click point and hit-tests it against the element's children, so a gap
  inside any container counts as a spot. A leaf has no children to miss, so clicking text inside a
  `<p>` still references the `<p>`; hover extraction reports no click and always stays an element.
- `BrowserAnnotationReference` carries identity, geometry and the optional click point — no HTML,
  styles or screenshot per reference, since these are prompt landmarks and a second full payload each
  would blow the annotation budget. Capped at 4 per annotation.
- `BrowserPageAnnotation.references` is optional and clamped in the store, so annotations persisted
  before this change load unchanged.
- Composer: crosshair pick control, reference chips with remove, a dashed outline plus `@refN` badge
  over a referenced element, and a marker with its container outlined for a spot.
- `useGrabMode` is untouched — a reference pick is the existing `rearm()` routed by the composer, so
  there is still one picker implementation.
- i18n strings across the five locales.

## Testing

- `browser-annotation-references.test.ts` (new) and additions to `browser-annotation-output.test.ts`:
  element-vs-spot classification including the no-click-point fallback, label truncation of
  page-controlled text, the reference cap and renumbering, caret-aware insertion (at caret, over a
  selection, padding, out-of-range caret), token removal with renumbering, unresolved-token stripping
  that leaves ordinary `@mentions` and newlines alone, and prompt block ordering and fencing.
- `pnpm test` — 800 tests across browser-pane, main-process grab, store and shared. `pnpm typecheck`
  (node, web, cli) and `pnpm lint` clean, including the max-lines ratchet and localization gates.
- Verified on Windows against a local page: element picks, spot picks in a grid gap and in page
  background, scrolling between a pick and its target, `Esc` mid-pick, chip removal, the cap, and a
  `position: fixed` reference staying pinned while scrolling.

## Notes

- No platform-specific paths and no new keyboard shortcut. References travel the same
  `browser.setGrabMode` / `awaitGrabSelection` IPC as the existing grab, with no new channel, so
  remote and SSH-backed browser pages are unaffected. Output is provider-neutral markdown.
- The child hit-test runs once per pick and is capped at 200 children; overlay rendering is bounded
  by the 4-reference cap and only mounts while the composer is open.
- References carry no HTML or attributes. Labels come from the same page-controlled strings the
  composer header already escapes, length-bounded before whitespace collapsing. Capturing the click
  point is wrapped separately, so an element that cannot be measured still yields a normal pick.
- Adjacent open work that may want ordering against this: #11360, #8687 / #8830, #9224.

X: [@PraveenKum27793](https://x.com/PraveenKum27793)

### DIFF
--- a/src/main/browser/browser-grab-payload.ts
+++ b/src/main/browser/browser-grab-payload.ts
@@ -1,4 +1,8 @@
-import type { BrowserGrabPayload, BrowserGrabRect } from '../../shared/browser-grab-types'
+import type {
+  BrowserGrabPayload,
+  BrowserGrabPoint,
+  BrowserGrabRect
+} from '../../shared/browser-grab-types'
 import {
   GRAB_BUDGET,
   GRAB_SAFE_ATTRIBUTE_NAMES,
@@ -133,6 +137,27 @@ export function clampGrabPayload(raw: unknown): BrowserGrabPayload | null {
       .filter(Boolean)
   }
 
+  // Why: absent for hover extraction and pre-clickPoint guests, so null is normal, not a failure.
+  const safePoint = (p: unknown): BrowserGrabPoint | null => {
+    if (!p || typeof p !== 'object') {
+      return null
+    }
+    const point = p as Record<string, unknown>
+    return {
+      viewportX: safeNum(point.viewportX),
+      viewportY: safeNum(point.viewportY),
+      pageX: safeNum(point.pageX),
+      pageY: safeNum(point.pageY),
+      offsetX: safeNum(point.offsetX),
+      offsetY: safeNum(point.offsetY),
+      ratioX: safeNum(point.ratioX),
+      ratioY: safeNum(point.ratioY),
+      hostWidth: safeNum(point.hostWidth),
+      hostHeight: safeNum(point.hostHeight),
+      inEmptySpace: point.inEmptySpace === true
+    }
+  }
+
   const safeRect = (r: unknown): BrowserGrabRect => {
     if (!r || typeof r !== 'object') {
       return { x: 0, y: 0, width: 0, height: 0 }
@@ -216,6 +241,7 @@ export function clampGrabPayload(raw: unknown): BrowserGrabPayload | null {
       GRAB_BUDGET.nearbyTextEntryMaxLength
     ),
     ancestorPath: clampArray(obj.ancestorPath, GRAB_BUDGET.ancestorPathMaxEntries, 200),
+    clickPoint: safePoint(obj.clickPoint),
     screenshot: null
   }
 }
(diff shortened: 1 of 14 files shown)


srs-adamr:
https://github.com/stablyai/orca/issues/15333 is the request. I do not agree with this PR as the shape of that feature.

#15333 is a **primitive**: the subject of an annotation is an open-ended set of page targets. Every member is a peer. Grab accumulates. N is not 2 and not 4.

This PR does not add that primitive. It keeps one `payload.target` and bolts on optional `references` (composer chips, `@refN`, **cap 4**, no full payload per extra target) so a note can point at “a second element or spot.” That fits one relational-review request. It does not change the annotation model.

Please do not treat this PR as closing #15333. The 4-reference landmark path is the wrong product shape for multi-select.

PR #12142, opened 2026-08-02
fix(workspace): keep the workspace selected when its last tab closes

### LINKED ISSUE #11699: [Bug]: Closing all terminals clears center panel file browser; must re-open workspace to browse files
### Operating system

_Unknown_

### Orca version

_Unknown_

### Details

**Short summary:** Closing all terminals/agent windows in a workspace clears the center panel file browser even though the workspace is still selected. The UI shows "Select a workspace to browse files", and the user must re-click the workspace and open a terminal before they can browse files or use Git.

### What happened?

The user closed every terminal or agent window in the current workspace. They still considered themselves in that workspace (nothing else was open), but the center panel already showed the empty state **"Select a workspace to browse files"**.

To browse files or use Git again, they had to click the workspace again and open a terminal — which is inconvenient when they only want to browse files (no terminal needed).

### Steps to reproduce

1. Open a workspace in Orca (with terminal and/or agent windows).
2. Ensure the center panel can browse files / show Git for that workspace.
3. Close all terminals and agent windows in that workspace.
4. Observe the center panel empty state: "Select a workspace to browse files".
5. Try to browse files or use Git without opening a new terminal.

### Expected behavior

- Closing terminals/agent windows should not deselect the active workspace for file browsing / Git.
- As long as a workspace remains selected/active, the center panel should keep offering file tree, editor, and Git for that workspace — even with zero open terminals.

### Actual behavior

- After the last terminal/agent window is closed, the center panel switches to "Select a workspace to browse files".
- The user must re-select the workspace and open a terminal before browsing files or using Git.

### Additional context

User report:

> There might be an interaction issue that could be improved. When I close all the terminal or agent windows in the current workspace, theoretically I am still in the current workspace but with nothing open. However, the center panel already displays "Select a workspace to browse files" at that time. I need to click on the workspace again to open a terminal in order to browse files or use Git. This can be inconvenient if all one wants to do is simply browse files.

User report

**Likely area:** workspace selection / panel empty-state logic treating "no open panes/terminals" as "no active workspace", instead of preserving the selected workspace for file browser + Git.

AnddyAgudelo:
## Summary

Fixes #11699.

Closing every tab in a workspace deselected the workspace itself. The file explorer fell back to *"Select a workspace to browse files"* and the Git panel emptied, so browsing files or using source control required re-clicking the workspace **and** opening a terminal — even when the user only wanted to read a file.

`FileExplorer` gates on `activeWorktreeId`, and seven separate close paths were clearing it once the last surface went away:

| Site | Path |
| --- | --- |
| `store/slices/tabs.ts` | unified tab model (`closeUnifiedTab`) |
| `store/slices/editor.ts` ×2 | `closeFile` and `closeAllFiles` |
| `components/terminal/terminal-tab-actions.ts` | last terminal tab |
| `components/Terminal.tsx` | last browser tab |
| `components/tab-group/useTabGroupWorkspaceModel.ts` | `leaveWorktreeIfEmpty` (split groups) |
| `components/terminal-pane/TerminalPaneOverlayLayer.tsx` | second `leaveWorktreeIfEmpty` |

All seven now leave `activeWorktreeId` intact and clear only the per-surface actives (`activeTabId`, `activeBrowserTabId`, `activeFileId`). The two `leaveWorktreeIfEmpty` hooks are removed outright, along with the now-dead `skipEmptyCheck` option.

### Why the deselection existed, and why removing it is safe now

This intentionally supersedes **PR #677** (*"return to Orca landing screen after closing last terminal"*). That behavior existed for a good reason, recorded in `tests/e2e/tab-close-navigation.spec.ts`:

> …require that when a worktree's last visible surface closes, the app clears `activeWorktreeId` instead of leaving a selected worktree with nothing to render. **Any regression here shows up as a blank workspace.**

That was accurate — the pane body renders only `activeTab`-driven content, so at zero tabs it was a tab strip over empty space. Keeping the workspace selected therefore required giving the pane something to show, so this PR adds `TabGroupEmptyState`: the hero icon, a short explanation that the sidebar still works, and a direct **New terminal** action with its platform shortcut.

Sleeping a workspace is now the route back to Landing. Deliberate deselections are untouched: sleep, worktree delete, repo removal, and folder-workspace delete.

### Empty-state visibility is derived, never timed

A group also sits at zero tabs *before* its first tab arrives — `ensureWorktreeRootGroup` creates the root group first, and a runtime-owned worktree then waits on the host's session-tab mirror for however long the link takes. Showing the empty state during that wait would be wrong.

Visibility is `workspaceSessionReady && tabCount === 0 && !awaitingFirstHostTabs`, where `awaitingFirstHostTabs` is true only while a runtime-owned worktree has no published snapshot epoch. No timer and no mount-local state — a reparent or reload can't strand the pane with no way to open a terminal. Because the epoch lives in module-local maps and a zero-tab snapshot moves no store slice, `web-session-tabs-sync.ts` gained an explicit publication subscription consumed via `useSyncExternalStore`; without it the predicate would be correct but the UI would never re-ask.

## Screenshots

Light-mode full window — the workspace stays highlighted in the sidebar and the file explorer stays populated, which is the reported bug going away. Plus light/dark detail shots of the empty state.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

New/updated coverage:

- `tab-group-empty-state-visibility.test.ts` — the visibility predicate, including a remote worktree resting at zero tabs reading as *empty* rather than perpetually pending, and remount-invariance.
- `tab-group-empty-state-reactivity.test.tsx` — renders the real hook and publishes a zero-tab snapshot through the real sync path, asserting the empty state appears **with no other state change**. Pure-predicate tests can't catch a missing subscription; this one can.
- `useTabGroupWorkspaceModel.focus.test.ts` — last terminal tab, last browser tab, sole group closed, and two-group split where only the emptied pane collapses.
- `tabs.test.ts`, `store-cascades.test.ts`, `editor.test.ts`, `terminal-tab-actions-kill-all.test.ts` — existing assertions that encoded the old deselection are retitled and inverted, each with a `Why:` pointing at this issue. Also a remote/host-owned close case.
- `tests/e2e/tab-close-navigation.spec.ts` — the last-surface test now asserts the workspace stays selected and the empty state offers a terminal. Its header documented #677's rationale as current truth and has been rewritten.

Verified fail-on-main: the four behavior test files produce **7 failures** against `main`'s source and pass on this branch.

## AI Review Report

Reviewed across five rounds, each verified against the code rather than accepted on assertion.

- **Cross-platform (macOS/Linux/Windows):** no `metaKey` hardcoding. The shortcut chip renders through `useShortcutKeyDetails` + `ShortcutKeyCombo`, which resolve the platform internally — the screenshots show `Ctrl + T` with the Linux/Windows `+` separator; macOS renders adjacent glyphs. No new path handling.
- **SSH / remote / local:** the first design used a 250ms settle timer, which would have flashed the empty state mid-hydration on any link slower than the timeout — rejected. The second used a mount-local latch, which a reload would reset and strand a remote workspace with no recovery affordance; `shouldBootstrapInitialWebRuntimeTerminal`'s guard is closure-local and consumed once, so a remote workspace resting at zero tabs is a real state, and this fix makes it a common one. Both replaced by the derived predicate above. `terminal-tab-actions.ts`'s host-authoritative branch returns before the last-tab logic and is covered by its own test.
- **Folder workspaces:** the empty state's action routes through `openNewTerminalTabInActiveWorkspace`, the same action the tab bar's `+` uses, which already handles folder-workspace keys and remote runtimes. No new assumption that a workspace is a git worktree.
- **Git providers:** copy says "Git", matching the `source-control` sidebar tab. Nothing GitHub-specific.
- **Performance:** no new polling or timers. One `useSyncExternalStore` subscription per mounted tab group, notified only when a session-tab snapshot is recorded or cleared.
- **UI quality:** `docs/STYLEGUIDE.md` throughout — `size-7` hero icon (its documented empty-state use), `muted-foreground`, `outline`/`sm` button, lucide `SquareTerminal`, no invented values. Rendered and inspected in both themes; the shortcut chip was regrouped after review because it read as a floating element rather than the button's shortcut.
- **Caught in review:** the E2E suite is not part of `pnpm test`. Running it revealed the fix was incomplete — the two `editor.ts` sites were still deselecting, so closing the last *editor* reproduced the original bug on a path no unit test covered.

## Security Audit

No new IPC surface, no command execution, no path handling, no auth or credential handling, no network calls, and no new dependencies. Changes are renderer-local state transitions plus one in-process listener set (`publicationListeners`), whose unsubscribe is returned to `useSyncExternalStore` and cleared on all three teardown paths, so listeners can't accumulate across worktree or environment churn. The empty state renders static translated copy — no user-controlled interpolation. No follow-up needed.

## Notes

- Reverses PR #677 deliberately; sleeping a workspace is now the route back to Landing.
- One narrow, pre-existing limitation left alone: if a runtime-owned host never publishes a snapshot at all (unreachable remote), the pane stays blank. That is a disconnected-runtime state surfaced by `TerminalRemoteRuntimeReconnectBanner` / `TerminalSshReconnectOverlay`; the empty state was deliberately not made to double as a connectivity error surface.
- A throwaway Playwright spec produced the screenshots and was intentionally left out of the commit — its behavioral assertion is already covered by `tab-close-navigation.spec.ts`, and it added ~19s to CI for evidence capture.

### DIFF
--- a/src/renderer/src/components/Terminal.tsx
+++ b/src/renderer/src/components/Terminal.tsx
@@ -1621,7 +1621,8 @@ function Terminal(): React.JSX.Element | null {
               setActiveTab(terminalTab.id)
               setActiveTabType('terminal')
             } else {
-              setActiveWorktree(null)
+              // Why: no tabs left, but the workspace stays selected so file/Git browsing survives (issue #11699).
+              setActiveTabType('terminal')
             }
           }
         }
@@ -1637,14 +1638,7 @@ function Terminal(): React.JSX.Element | null {
       destroyWorkspaceWebviews(state.browserPagesByWorkspace, tabId)
       closeBrowserTab(tabId)
     },
-    [
-      closeBrowserTab,
-      setActiveBrowserTab,
-      setActiveFile,
-      setActiveTab,
-      setActiveTabType,
-      setActiveWorktree
-    ]
+    [closeBrowserTab, setActiveBrowserTab, setActiveFile, setActiveTab, setActiveTabType]
   )
 
   const handlePtyExit = useCallback(
--- a/src/renderer/src/components/shared/kill-all-terminal-surfaces.ts
+++ b/src/renderer/src/components/shared/kill-all-terminal-surfaces.ts
@@ -173,7 +173,7 @@ export async function runKillAllTerminalSurfaces(
     }
     const activeWorktreeId = state.activeWorktreeId
     // Why: keeping the active worktree until its targets close lets the existing
-    // tab action choose editor/browser/deactivation without spawning a replacement.
+    // tab action fall back to an editor/browser tab without spawning a replacement.
     const closeOrder = [
       ...presentTargetIds.filter((targetId) => ownerByTargetId.get(targetId) !== activeWorktreeId),
       ...presentTargetIds.filter((targetId) => ownerByTargetId.get(targetId) === activeWorktreeId)
--- a/src/renderer/src/components/tab-group/TabGroupEmptyState.test.tsx
+++ b/src/renderer/src/components/tab-group/TabGroupEmptyState.test.tsx
@@ -1,0 +1,34 @@
+// @vitest-environment happy-dom
+
+import { cleanup, fireEvent, render, screen } from '@testing-library/react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { TabGroupEmptyState } from './TabGroupEmptyState'
+
+vi.mock('@/i18n/i18n', () => ({
+  translate: (_key: string, fallback: string) => fallback
+}))
+
+vi.mock('@/hooks/useShortcutLabel', () => ({
+  useShortcutKeyDetails: () => ({ keys: ['Ctrl', 'T'], doubleTap: false })
+}))
+
+afterEach(cleanup)
+
+describe('TabGroupEmptyState', () => {
+  it('offers a direct new-terminal action when the workspace has no tabs left', () => {
+    const onNewTerminalTab = vi.fn()
+    render(<TabGroupEmptyState onNewTerminalTab={onNewTerminalTab} />)
+
+    expect(screen.getByText('No open tabs')).toBeDefined()
+    fireEvent.click(screen.getByRole('button', { name: /New terminal/ }))
+
+    expect(onNewTerminalTab).toHaveBeenCalledTimes(1)
+  })
+
+  it('renders the platform shortcut chips supplied by the keybinding hook', () => {
+    render(<TabGroupEmptyState onNewTerminalTab={vi.fn()} />)
+
+    expect(screen.getByText('Ctrl')).toBeDefined()
+    expect(screen.getByText('T')).toBeDefined()
+  })
+})
--- a/src/renderer/src/components/tab-group/TabGroupEmptyState.tsx
+++ b/src/renderer/src/components/tab-group/TabGroupEmptyState.tsx
@@ -1,0 +1,52 @@
+import { SquareTerminal } from 'lucide-react'
+import { Button } from '@/components/ui/button'
+import { ShortcutKeyCombo } from '../ShortcutKeyCombo'
+import { useShortcutKeyDetails } from '@/hooks/useShortcutLabel'
+import { translate } from '@/i18n/i18n'
+
+/**
+ * Shown in a tab group whose tabs are all closed. The workspace deliberately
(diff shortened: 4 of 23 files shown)


AnddyAgudelo:
tab-group-empty-state-dark-detail
tab-group-empty-state-dark-window
tab-group-empty-state-light-detail
tab-group-empty-state-light-window

PR #12143, opened 2026-08-02
fix(codex): carry /goal state and resume the session across account-switch restarts

brennanb2025:
## Summary

Fixes /goal loss when a Codex pane is restarted onto another account. Codex stores /goal state in `goals_1.sqlite` inside each `CODEX_HOME`, keyed by thread id; the account-switch restart used to launch a brand-new thread in a different home, leaving the goal doubly unreachable (new thread id, other home's DB).

Now, when the user restarts a stranded Codex pane after an account switch:

- Main resolves the pane's launch `CODEX_HOME` from the pane-account registry and the new selection's home, and only proceeds when both can be named with certainty.
- The goal RPCs (`thread/goal/get`/`thread/goal/set`, codex-cli ≥ 0.146) are probed through `codex app-server` behind a dedicated capability cache with the existing narrow unsupported-error predicate (method-not-found / missing subcommand only), so old CLIs re-probe after the standard interval instead of erroring per restart.
- The thread's rollout is hardlinked into the new home ahead of the asynchronous whole-tree session bridge, then `thread/read` (Codex's sanctioned lazy indexing) proves it resumable there before Orca commits to anything.
- The goal row travels via `thread/goal/get` against the old home → `thread/goal/set` against the new home. Orca never reads or writes Codex's sqlite itself — the app-server owns the DB and its WAL.
- Only after all of that does the renderer swap the restart startup from `codex` to `codex resume <thread-id>`.

Every uncertain path — unrecorded panes, WSL/SSH/remote lanes, custom `CODEX_HOME` overrides, older codex CLIs, unresolvable rollouts, an older preload, any RPC failure — degrades to exactly today's fresh `codex` startup (byte-identical constant), because resuming the wrong session is worse than losing the goal.

All protocol assumptions were live-verified against codex-cli 0.146.0 in a sandbox `CODEX_HOME`: `thread/goal/get` returns `{goal: null}` for an unindexed/goalless thread, `thread/read` indexes a rollout merely linked into a foreign home, and `thread/goal/set` round-trips objective/status (`usageLimited`, `paused`)/tokenBudget including `null`.

## Screenshots

No visual change (the restarted pane runs `codex resume <id>` instead of `codex`; the restart notice UI is untouched).

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

`pnpm test`: 43,537 passed. 21 failures across 7 files reproduce identically at the base commit `99b94a38eb` in a pristine worktree (ambient `ORCA_*`/`GIT_CONFIG_*` env from the agent terminal plus two known load-flaky daemon-cwd tests) — none overlap this diff; with a clean environment all of them pass except the two pre-existing cwd-repair timeouts, which also time out at base.

New tests: home-resolution decision table (lanes, routes, overrides, ownership failures), the capability fallback (unsupported marking, no re-probe inside the interval, self-heal after it, transient errors not poisoning the cache), the get-then-set bridge against a faked app-server client (RPC order, exact `thread/goal/set` payload, same-home short-circuit, every failure → fresh), the targeted rollout bridge on real temp homes (hardlink inode, race with the background bridge, foreign-transcript rejection), and the renderer resolver asserting the fallback path yields today's exact startup command object.

## AI Review Report

Self-reviewed with focus on: wrong-session resume (the failure mode worse than the bug — gated on the pane-account registry record, hook-reported session id restricted to bare rollout UUIDs, and rollout provenance resolved strictly inside the launch home via the existing trusted-resume resolver, which refuses id-scan fallback when a reported transcript path lies elsewhere); ordering (the targeted hardlink + `thread/read` run before the resume command is ever issued, so `codex resume` cannot race the background session bridge); restart-path regression (fallback returns the identical `CODEX_ACCOUNT_RESTART_STARTUP` constant, asserted by test); TerminalPane re-entrancy (in-flight guard plus a stale-transport check after the await so a second click or a mid-await reconnect cannot tear down the replacement PTY).

Cross-platform was explicitly checked for macOS, Linux, and Windows: app-server spawns go through `getSpawnArgsForWindows` (cmd.exe-wrapped npm installs) with the existing process-tree kill; all paths use `path.join`/`relative` and `normalizeRuntimePathForComparison` for comparisons; the rollout link reuses the hardlink-then-symlink helper with its win32 symlink type; the resume command is a plain `codex resume <uuid>` typed into the shell, valid in bash/zsh/pwsh/cmd. WSL, SSH, and remote-runtime panes are deliberately out of scope and keep today's behavior (see Notes). The changed-code quality gate (oxlint native + type-aware + React Doctor) reports 0 new findings.

## Security Audit

- IPC input: the new `codexAccounts:prepareAccountSwitchResume` handler type-checks all fields; the thread id must match a strict UUID pattern in main (independently re-validated in the renderer before it is embedded in a shell command), so no injection path exists.
- Path handling: the hook-reported transcript path is only honored when it resolves inside the launch home's `sessions` tree under Codex's dated rollout layout (existing containment logic); the link target is derived from the relative path inside the sessions roots, so nothing outside the two managed trees can be read or linked.
- Managed homes: a recorded account home is re-validated with `assertOwnedHostCodexManagedHomePath` before use; failure degrades to fresh.
- Command execution: only `codex app-server` with fixed argv, pinned `CODEX_HOME` env, hard session deadlines, and the existing SIGKILL reap. No sqlite is opened by Orca; goal data moves exclusively between Codex's own homes through Codex's own RPCs.
- No new secrets, dependencies, or network surface.

## Notes

- Scope is the account-switch restart path only — no general goal-sync daemon.
- WSL lanes degrade to today's fresh startup: the goal bridge would need in-distro app-server invocation and in-distro rollout linking, and the capability gate treats that as unavailable for now. SSH/remote panes never reach the restart flow (no local registry record).
- `thread/goal/set` cannot carry `tokensUsed`/`timeUsedSeconds`, so budget-consumption counters restart at 0 in the new home; objective, status (including `usageLimited`/`paused`/`blocked`/`complete`), and `tokenBudget` travel intact.
- The goal snapshot is read while the old pane is still live, so a goal update in the final instant before teardown would not be captured — bounded by the restart click itself.

### DIFF
--- a/src/main/codex-accounts/service.ts
+++ b/src/main/codex-accounts/service.ts
@@ -1371,6 +1371,12 @@ export class CodexAccountService {
     return root
   }
 
+  /** Root that owns every managed account home; lets callers validate a pane's
+   *  recorded launch home without reaching into private state. */
+  getManagedAccountsRootPath(): string {
+    return this.getManagedAccountsRoot()
+  }
+
   private ensureManagedHomeForReauthentication(account: CodexManagedAccount): string {
     const wslInfo = parseWslUncPath(account.managedHomePath)
     if (wslInfo && process.platform === 'win32') {
--- a/src/main/codex/codex-account-session-bridge.test.ts
+++ b/src/main/codex/codex-account-session-bridge.test.ts
@@ -5,6 +5,7 @@ import { join } from 'node:path'
 import {
   _internals,
   bridgeCodexSessionsIntoAccountHome,
+  ensureCodexRolloutBridgedIntoAccountHome,
   startCodexAccountSessionBridgeInBackground
 } from './codex-account-session-bridge'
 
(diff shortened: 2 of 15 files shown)


brennanb2025:
## Senior review follow-up

Reviewed the current head `6a7a8a9c27` with a same-model, same-reasoning senior review loop. The reviewer fixed the material issues it found, pushed the follow-up commit, then re-reviewed the resulting diff and returned clean.

Meaningful follow-up fixes:

- mirror a goalless source by clearing any stale target goal via `thread/goal/clear`
- reject malformed goal responses instead of guessing
- avoid app-server subprocesses for same-home restarts and coalesce concurrent unsupported capability probes per host
- fence async restart application against PTY replacement and a second account selection
- add deterministic regression and process-count coverage

Verification:

- focused Vitest: 5 files, 64 tests passed
- `pnpm typecheck`: passed
- full `pnpm lint`: passed
- `git diff --check`: passed
- required PR checks: passed, including Node 24/26 test shards, macOS/Windows packaging, static analysis, Git compatibility, and shell contracts
- worktree clean; PR mergeable

Electron screenshots are not attached: the required `$electron` capability is unavailable in this session, and repository policy explicitly prohibits substituting general desktop automation. This is desktop-only; no mobile code changed.

brennanb2025:
Closing after a deeper architecture and reliability review.

Codex splits resumable session data between rollout files and SQLite-backed runtime state selected by the effective `CODEX_HOME` / `sqlite_home`. This PR attempts a partial cross-account migration by linking the rollout, lazily indexing it in the target home, and replaying the goal through `thread/goal/set`.

That transfer is not safe or faithful with the current Codex interfaces:

- `thread/goal/set` cannot import `tokensUsed` or `timeUsedSeconds`, so accounting can reset or diverge across account homes.
- `sqlite_home` and `CODEX_SQLITE_HOME` can redirect the real state authority away from the `CODEX_HOME` this bridge assumes.
- An existing target rollout alias may have diverged after Codex replaces or compresses its session files.
- Short-lived app-server processes can collide with large-history state DB backfill.
- The source Codex remains live while the goal and rollout snapshot are prepared, leaving a race with in-flight updates.

We should not ship a workaround that can resume mismatched state or weaken goal budget accounting. The safe account-switch restart remains a fresh Codex session for now. We can revisit automatic transfer when Codex exposes a supported complete thread and goal transfer operation.

Current workaround: after switching accounts, reopen the prior conversation using `/resume` or Orca AI Vault, ask the agent to output only the goal objective it was working on, then paste that objective into a new `/goal`.

PR #12145, opened 2026-08-02
fix(codex): make quota probes credential-safe

brennanb2025:
## Summary

Addresses #11969 (users repeatedly logged out of connected Codex accounts, correlated with account switching). Codex OAuth uses **rotating refresh tokens**, and Orca's quota probes handled them dangerously in four ways, each fixed here:

1. **Graceful probe shutdown.** The RPC probe spawned `codex app-server` inside the account's live credential home and hard-killed it at a 10s deadline, while cold starts (including an in-flight token refresh) are documented at 10–25s — a kill landing mid-rotation can permanently invalidate the stored credential. Now the request deadline is armed only after the app-server answers `initialize` (boot budget: 30s host / 40s WSL), and every probe termination first requests shutdown (stdin EOF + SIGTERM; stdin EOF alone on Windows, where Node's `kill()` is always forceful) with a bounded 5s drain window before any hard kill. The fetch resolves only once the child has actually exited.
2. **Per-credential-home serialization.** New `withCodexHomeProcessLock` FIFO mutex, keyed by the normalized codex home (WSL `\\wsl$` UNC and in-distro Linux paths collide onto one key). Quota probes (RPC and PTY paths) and Orca's own codex spawns for commit messages, PR fields, branch names, and model discovery all share it, so Orca never runs two of its own codex processes against one `auth.json`. Stop/cancel still works while a generation is queued behind a probe. User terminal panes are intentionally not serialized (out of scope).
3. **No more switch-storm.** Account switches no longer zero the inactive-probe debounce (previously every switch re-probed every inactive account's credential home at once). The 60s debounce holds, the switcher shows the cached snapshot (seeded from the outgoing account), inactive probes are staggered 2s apart, and the newly **active** account still gets its immediate refresh.
4. **Parse-error grace.** Credential reads now distinguish `present` / `missing` / `unreadable` / `no-credential`. A transient missing or torn (mid-write) `auth.json` no longer deselects the active managed account: absence must persist past a 5s grace window before deselection — unless the managed home directory itself is gone, which is structural (not a rotation) and still deselects immediately. Settled valid-JSON-without-credential (a real logout) also still deselects immediately.

## Screenshots

No visual change

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [ ] `pnpm build` (left to CI — desktop + native packaging)

`pnpm test` (node 24): 43,511 passed. The only failures on my machine (20 tests across 6 files: `ipc/pty`, `relay/*`, `daemon/pty-subprocess`, `setup-agent-sequencing`, `check-root-directory-entries`) are environment-sensitive and fail **identically at the parent commit on main** — verified by running the same files at `HEAD~1` under identical conditions and diffing the failure sets. Every suite touching this change passes.

New coverage: graceful-drain sequencing with a fake child (SIGTERM → drain → SIGKILL escalation, Windows stdin-EOF-only lane, bounded resolution for a child that never exits, already-exited no-op); boot-budget vs read-deadline arming (a 20s-cold app-server is no longer killed); per-home lock behavior (same-key queueing, cross-key concurrency, release after rejection, WSL UNC ↔ Linux-path key collision); fetcher-level "no second spawn for the same home while one probe runs"; switch-debounce and 2s stagger in `RateLimitService`; credential-absence grace transitions; and runtime-home-service "transient unreadable then valid auth.json does not deselect" plus grace-then-durable deselect.

## AI Review Report

Reviewed the change for lock leaks (per-home lock tail is rejection-safe and self-cleaning; probe resolution waits for child exit so the lock can't release mid-drain), cancellation races (a queued codex generation registers a pre-lock cancel token so Stop works while waiting), non-codex regression (the non-codex generation path keeps zero added promise hops, preserving cancellation timing byte-for-byte), and deselect-behavior regressions (structurally missing managed homes still deselect immediately at startup).

Cross-platform compatibility was explicitly checked for macOS, Linux, and Windows:
- **Windows:** Node's `kill()` is TerminateProcess, so the graceful phase uses stdin EOF only and escalates to `kill()` after the drain; the login-flow `taskkill /t` tree-kill path is untouched; the background-probe ConPTY PTY fallback remains disabled on Windows; `EPERM`/`EBUSY` reads during auth.json rotation are classified as transient `unreadable`.
- **WSL:** the boot budget is 40s (cold WSL starts exceed host budgets); stdin EOF propagates to the in-distro codex through the existing fd-isolated login-shell plumbing; lock keys normalize `\\wsl$`/`\\wsl.localhost` UNC and Linux paths onto one key per distro.
- **macOS/Linux:** stdin EOF + SIGTERM, then SIGKILL after the drain.
- **SSH:** remote generation goes through `runRemotePlan`/provider execution and is untouched; folder workspaces are unaffected (no worktree assumptions anywhere in the change).

## Security Audit

- No new command execution: the same codex binaries/arguments are spawned; the lock and grace logic are purely in-process.
- `auth.json` is only read (never written) by the new code, torn reads no longer influence deselection, and no credential material is logged.
- Path handling: `normalizeRuntimePathForComparison` output is used for comparison keys only, never spliced back into real paths.
- No IPC surface changes; no new dependencies.

## Notes

- The deadline error stays `RPC timeout` for both phases, so existing error classification is unchanged.
- Follow-ups (deliberately out of scope, per the one-theme rule):
  - WSL bidirectional auth mirroring fork
  - hook-lane flip handling
  - multi-pane token refresh coordination
  - graceful drain for the PTY fallback path (`cleanupHiddenRateLimitPty` still kills immediately; the per-home lock already prevents overlapping spawns there)
  - the WSL managed-account missing-auth deselect (`syncWslRuntimeForCurrentSelection`) still deselects without grace; the same grace could be applied
  - system-default WSL commit-message runs key the lock with a distro-scoped sentinel (the distro's `~/.codex` isn't resolvable from the host), so they serialize with each other but not with probes that pass an explicit UNC home

### DIFF
--- a/src/main/codex-accounts/codex-credential-absence-grace.test.ts
+++ b/src/main/codex-accounts/codex-credential-absence-grace.test.ts
@@ -1,0 +1,65 @@
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import {
+  CODEX_CREDENTIAL_ABSENCE_GRACE_MS,
+  CodexCredentialAbsenceGrace
+} from './codex-credential-absence-grace'
+
+const VALID_AUTH = JSON.stringify({
+  auth_mode: 'chatgpt',
+  tokens: { access_token: 'a', id_token: 'b', refresh_token: 'c' }
+})
+
+describe('CodexCredentialAbsenceGrace', () => {
+  let dir = ''
+  let authPath = ''
+
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), 'codex-absence-grace-'))
+    authPath = join(dir, 'auth.json')
+  })
+
+  afterEach(() => {
+    rmSync(dir, { recursive: true, force: true })
+  })
+
+  it('treats a torn mid-write read as transient until it outlives the grace window', () => {
+    const grace = new CodexCredentialAbsenceGrace()
+    writeFileSync(authPath, '{"tokens":{"acc', 'utf-8')
+
+    expect(grace.assess(authPath, 1_000)).toEqual({ state: 'unreadable', durable: false })
+    expect(grace.assess(authPath, 1_000 + CODEX_CREDENTIAL_ABSENCE_GRACE_MS - 1)).toEqual({
+      state: 'unreadable',
+      durable: false
+    })
+    expect(grace.assess(authPath, 1_000 + CODEX_CREDENTIAL_ABSENCE_GRACE_MS)).toEqual({
+      state: 'unreadable',
+      durable: true
+    })
+  })
+
+  it('clears the absence clock when a transient failure heals', () => {
+    const grace = new CodexCredentialAbsenceGrace()
+    writeFileSync(authPath, '{"tokens":{"acc', 'utf-8')
+    expect(grace.assess(authPath, 1_000).durable).toBe(false)
+
+    writeFileSync(authPath, VALID_AUTH, 'utf-8')
+    expect(grace.assess(authPath, 2_000)).toEqual({ state: 'present', durable: true })
+
+    // A later absence starts a fresh grace window instead of inheriting the old clock.
+    rmSync(authPath)
+    expect(grace.assess(authPath, 100_000)).toEqual({ state: 'missing', durable: false })
+    expect(grace.assess(authPath, 100_000 + CODEX_CREDENTIAL_ABSENCE_GRACE_MS)).toEqual({
+      state: 'missing',
+      durable: true
+    })
+  })
+
+  it('reports settled credential-free JSON as durable immediately', () => {
+    const grace = new CodexCredentialAbsenceGrace()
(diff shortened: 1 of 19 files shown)


brennanb2025:
## Electron QA — final head

Validated `a291c9c685b2bc090aeabfa28b2562b7c3e1a666` in an isolated Electron profile through CDP. The automation identity matched `brennanb2025/codex-probe-credential-safety` and this worktree.

Visible proof: opened Settings → AI Provider Accounts through the real UI. The Codex surface rendered in full app context as **System default · Active · Needs sign-in**, with “No Codex sign-in was found for this device.” The config-sync warning is expected because the isolated HOME intentionally has no `.codex/config.toml`.

Backing proof: `window.api.codexAccounts.list()` returned no managed accounts, `activeAccountId: null`, and system default `{ hasAuth: false, authKind: "none" }`, matching the visible state. Renderer console: 0 errors; 3 unrelated dev-only Tooltip/React Grab warnings.

Remaining gap: a real Codex login/terminal credential hot-swap was not exercised to avoid using live credentials. Credential mirroring, process-lock, shutdown, and grace behavior are covered by deterministic main-process tests.

AI Provider Accounts on final Electron head

PR #12146, opened 2026-08-02
fix(codex): execute accepted account-switch restarts for unmounted panes

brennanb2025:
## Summary

The prior worktree-wide Codex account-switch prompt queued restarts for every awaiting pane, but execution only happened inside mounted `TerminalPane` instances. A pane without a mounted transport could be stranded: its prompt disappeared, keyboard input stayed blocked, and Codex kept running against the old account's `CODEX_HOME` until the tab was revealed.

This PR makes an accepted restart execute for every queued pane regardless of renderer mount state:

- **Detached executor** (`codex-detached-pane-restart.ts`): a store-level driver watches `pendingCodexPaneRestartIds`. Generation-fenced, single-flight microtask scheduling lets mounted panes claim their own restarts first without a timing delay. A still-unowned pane is kill-and-respawned through `pty:spawn`/`pty:kill`, then its tab and layout bindings are rebound so a later mount reattaches to the replacement PTY. Mounted panes keep the existing in-place path, including split-pane ownership.
- **Pane-local decisions**: each restart prompt is portaled into its owning terminal pane and selects one exact PTY notice. Restart/Keep affects only that pane, including split panes; siblings keep their own prompt and session. Only the active visible pane may take dialog focus, and replacement PTYs resolve through the live transport with the persisted leaf binding as fallback.
- **Race-safe ownership**: worktree, tab generation, old PTY, layout leaf, and mounted ownership are revalidated before and after spawn. Mounted-owner handoffs requeue before cleanup; tab retirement during spawn reaps the replacement. Best-effort cleanup cannot block later restart requests.
- **Never a silent input block**: success clears the notice and unblocks input. A failed respawn re-offers the prompt instead of leaving the pane muted. A removed pane clears its notice, while sleep-retained and foreign-machine panes remain on their existing lifecycle paths.
- **App-restart safety**: `sweepRestoredCodexPanesForStaleAccounts` queues every restored PTY id for stale-account inspection at startup, including panes whose tabs never mount.
- **Managed-auth readiness**: `CODEX_ACCOUNT_RESTART_STARTUP` carries `launchAgent: 'codex'`, so both restart paths use main's managed-auth readiness gate and Codex launch preparation.
- **Bounded hot path**: the store subscriber uses reference equality before inspecting restart work, avoiding allocations or scans on unrelated Zustand updates. Detached hidden replacements start at 80×24 and converge through the normal mount-fit path, without waiting on a possibly hung `getSize` call.

## Screenshots

Before — two split PTYs each present their own account-specific restart prompt:

Per-pane restart prompts

After — accepting **Restart** in the left pane replaced only its PTY and launched Codex there; the right PTY stayed alive with its own unanswered prompt:

One pane restarted independently

Electron backing-state validation confirmed the old left PTY was dead, its replacement was alive, the untouched right PTY was still alive, and only the right notice remained. The detached ownership path remains covered deterministically by the race tests below.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`
- [x] Added or updated high-quality regression tests

Focused validation after review:

- 88 tests across the four Codex restart suites
- 92 tests across the five `terminal-session.layout-pty-ownership` reliability suites
- 101 focused per-pane/restart tests, plus 5 restored-layout/geometry tests, after the pane-local UI correction
- `pnpm typecheck`
- Targeted lint, React Doctor, and `git diff --check`
- Electron validation through the account-switch prompt and real Codex PTY replacement

Key regression coverage asserts actual `pty:spawn`/`pty:kill` behavior and covers mounted-owner handoff, split/rootless layouts, tab/worktree generation races, cleanup rejection and hangs, stale-state prompt recovery, sleep-retained and foreign-machine ids, startup stale-account sweep, managed-auth launch metadata, O(1) unrelated-store writes, and the 80×24-to-mounted-fit path.

Note on the full local `pnpm test` run: 43,509 passed; 18 failures across five unrelated files reproduced identically on base commit `99b94a38eb` and were caused by machine environment. All files touched by this PR are green, and the PR's Node 24/26 CI shards pass.

## AI Review Report

The iterative review checked the queue/claim protocol for double execution, renderer cleanup and StrictMode generation races, mounted/background/split/rootless/sleep/foreign ownership, pane-local prompt/action ownership, parked-tab exit handling, orphaned-PTY windows, prompt/input recovery, and repeated restart progress when kill cleanup rejects or hangs. The final same-model review returned `CLEAN` at `578dcddb6dd1ba20ce544d970e58d7f6d040b63f`.

The performance review verified O(1) per-PTY prompt selectors, an O(1) reference gate on unrelated restart-store updates, single-flight scheduling per restart generation, and no unbounded pre-spawn provider-size wait.

Cross-platform behavior remains behind existing runtime abstractions. The detached spawn mirrors the mounted path's execution context and startup cwd handling, preserving native, Windows/WSL, SSH/relay, folder-workspace, and git-worktree behavior. Foreign-machine panes remain on their existing mounted transport path.

## Security Audit

No new IPC surface is introduced. The driver uses the existing preload `pty.spawn`/`pty.kill` APIs and main-side PTY/session, startup-cwd, and folder-workspace validation. Restart commands are constant, environment additions are the existing store-derived `ORCA_*` pane identity fields, and kill targets come from renderer store bindings.

## Notes

- Detached relay-environment respawn remains deliberately out of scope because it needs the remote transport's spawn assembly; those panes continue through their mounted path.
- `src/main/codex/codex-stale-pane-accounts.ts` needed no change: the account registry already forgets records on real PTY teardown, while restored-stale prompt recovery is handled by the renderer startup sweep.


brennanb2025:
Electron QA and iterative review are complete at `19bc70ae15c0d7a6a6abd64a1a63a58bf873fc81`.

The real account-switch prompt was accepted through the UI. Backing state confirmed the old PTY died, a replacement Codex PTY became live in the same terminal tab, and the pending restart/notice maps cleared.

| Before | After |
| --- | --- |
| [Restart prompt](https://github.com/user-attachments/assets/83d28337-df1d-4bfd-a100-287098393303) | [Replacement Codex session](https://github.com/user-attachments/assets/a91a5794-d6d6-4476-ad48-1e15fdf2cd13) |

The isolated profile retained the hidden tab as a mounted hot view, so the fully detached ownership branch is covered by deterministic race/hang tests rather than a mocked UI state. The same-model review loop returned `CLEAN`; focused Codex and terminal-ownership suites, typecheck, lint checks, and all PR CI checks pass.

brennanb2025:
Follow-up review and Electron QA are complete at `578dcddb6dd1ba20ce544d970e58d7f6d040b63f`. This supersedes the earlier worktree-wide prompt evidence.

The restart prompt is now owned by its exact terminal pane. In a real two-PTY split, each pane showed its own account-specific prompt. Clicking **Restart** in the left pane killed/replaced only the left PTY and launched Codex there; the right PTY remained alive with its own unanswered prompt.

| Independent pane prompts | Left pane restarted; right pane untouched |
| --- | --- |
| [Two pane-local prompts](https://github.com/user-attachments/assets/cf949b9d-5572-4f51-98da-b57c94d4825c) | [One pane restarted independently](https://github.com/user-attachments/assets/4b0019c8-419a-4db1-8d79-243163e8e989) |

Backing state: old left PTY dead; replacement left PTY alive; original right PTY alive; pending restart map empty; only the right pane's notice remained. The same-model review loop returned `CLEAN`; 101 focused restart tests plus restored-layout/geometry coverage, typecheck, lint, React Doctor, reliability checks, and commit hooks passed.

brennanb2025:
Final same-model review is CLEAN at `e4e8674b9d5506a02a348dc6180d4b99d4c87616`.

The review found one material current-head issue: an unexpected failure before the detached restart executor entered its original per-pane try/catch could reject the whole scheduler sweep, surface an unhandled renderer rejection, and prevent later queued panes from progressing. The pushed fix contains failures per PTY, restores the affected unanswered prompt, allows sibling restarts to continue, and keeps a last-resort scheduler rejection guard. A deterministic regression test covers the failure plus later-pane progress. The two Greptile threads were already addressed by the existing cleanup and bounded-size changes.

Verification:

- 89 focused Codex restart tests passed
- 92 `terminal-session.layout-pty-ownership` reliability tests passed
- Full typecheck, targeted lint/React Doctor, formatting, and `git diff --check` passed
- All required PR checks passed, including Node 24/26 shards and macOS/Windows packaging
- Local, remote, and PR heads match; worktree is clean

Electron was not rerun because the exact visible workflow is already attached in the [current PR QA comment](https://github.com/stablyai/orca/pull/12146#issuecomment-5161317177), as allowed by the follow-up request. I rechecked both uncropped images: two split panes show independent prompts, then restarting the left pane launches Codex there while the right PTY and its prompt remain untouched. This final commit changes only background failure containment, not the rendered UI. The fully detached ownership path remains covered by the deterministic restart/race tests.

Issue #10252, opened 2026-07-24
Deleting one workspace kills terminal/agent sessions in other workspaces sharing the same checkout path (cwdOwned fallback)


danielva-monday:
## Summary

Deleting one workspace (right-click → delete) can kill terminal/agent sessions belonging to **other** workspaces, not just the deleted one. Both `pi` and Claude Code agent sessions are affected. All co-located sessions die at the same instant, and there is no recovery (the agent processes are terminated).

## Root cause (traced in source)

The sibling kills originate from the `cwdOwned` **path-based fallback** in `sweepProviderByPrefix()`:

**`src/main/runtime/worktree-teardown.ts`** (~lines 250–256)

```ts
const cwdOwned =
  worktreePath !== undefined &&
  session.worktreeId === undefined &&           // only sessions missing a worktreeId tag
  typeof session.cwd === 'string' &&
  session.cwd.length > 0 &&
  isPathInsideOrEqual(worktreePath, session.cwd) // ← kills ANY session whose cwd is inside the path

return session.id.startsWith(prefix)   // exact `${worktreeId}@@` — safe
  || session.worktreeId === worktreeId // authoritative field — safe
  || cwdOwned                          // ← sibling-kill path
```

The `worktreePath` fed into this check is derived via `splitWorktreeIdForFilesystem(worktreeId)`, which **strips the `::workspace:<uuid>` suffix** for folder workspaces. So when a folder-workspace is deleted, `worktreePath` collapses to the shared checkout directory used by *every* workspace instance rooted there. Any session under that path whose `worktreeId` is `undefined` is then swept.

Sessions end up with `worktreeId === undefined` (and thus fall into this fallback) in cases including:
- Local PTY sessions spawned without an explicit `worktreeId` tag — `LocalPtyProvider.listProcesses()` only sets `worktreeId` when `ptyWorktreeId.get(id)` is truthy.
- Legacy/pre-`@@`-format sessions from older daemon generations.

This is especially destructive when a repo's checkout path is a broad directory (e.g. a home directory registered as a repo) where multiple independent workspaces share the same path — deleting one wipes all of them.

## Reproduction

1. Register a repo whose checkout path is a directory that hosts multiple workspaces sharing that same path (folder-workspace style, IDs like `<repoId>::<path>::workspace:<uuid>`).
2. Open a long-lived agent session (Claude Code or pi) in one workspace on that path.
3. Create another agent session (workspace) on the same path.
4. Delete the second workspace via right-click → delete.
[...]

PR #12153, opened 2026-08-03
fix(worktree): stop the PTY gate from permanently wedging workspace removal (#11960)

### LINKED ISSUE #11960: [Bug]: worktree permanently unremovable — "Failed to physically stop every PTY" with no terminal listed and --force ineffective
**Operating system:** Windows (11 Pro Education, 10.0.26200)

**Orca version:** 1.4.159

### Short summary

A worktree created by a scheduled Orca automation became permanently unremovable. `orca worktree rm` fails with `Failed to physically stop every PTY for worktree: …`, and `--force` fails identically — while `orca terminal list` reports no terminal for that worktree and no OS process references its path. There appears to be no documented escape hatch once a worktree reaches this state.

### What happened

The worktree was created by an hourly automation running with `workspaceMode: new_per_run` (provider `codex`). The run completed successfully and posted its result. Afterwards the workspace could not be removed by any documented route.

```
$ orca worktree rm --worktree "id:<repo-id>::C:/Users/admin/orca/workspaces/<repo>/auto-review-…-run-28-20260801T1321"
Failed to physically stop every PTY for worktree: <repo-id>::C:/Users/admin/orca/workspaces/<repo>/auto-review-…-run-28-20260801T1321

$ orca worktree rm --worktree "<same>" --force
Failed to physically stop every PTY for worktree: <same>      # identical, --force does not bypass the gate
```

Everything Orca exposes says there is no terminal left:

```
$ orca terminal stop --worktree "<same>"
Stopped 1 terminals.

$ orca terminal close --terminal term_abab11ee-… --tab
tab_not_found

$ orca terminal list | grep run-28
(no output)
```

No OS process references the worktree path either:

```powershell
Get-CimInstance Win32_Process | Where-Object { $_.CommandLine -like '*run-28-20260801T1321*' }
# only the grep command itself matched
```

Retrying `orca worktree rm` 20s later, and again after the steps below, produced the same error every time.

### Filesystem state

`git worktree remove --force` initially failed with `Permission denied`, but it did delete the directory's contents. The directory is now **empty** and still cannot be removed:

```
$ ls -a <worktree-path>
.  ..

$ rm -rf <worktree-path>
rm: cannot remove '<worktree-path>': Device or resource busy
```

An empty directory returning `EBUSY` is the signature of a process holding it as its **current working directory**.

After that, the two views disagree:

- `git worktree list` — no longer lists it
- `orca worktree list` — still lists it

### How to reproduce

I could not find a deterministic trigger; this happened once out of several automation runs. The sequence that produced it:

1. Create a scheduled automation with `workspaceMode: new_per_run`, provider `codex`.
2. Let it run to completion (the agent shells out heavily — `git fetch`, `npm ci`, `gh` calls).
3. After the run reports `completed`, try `orca worktree rm --worktree <id>` on the leftover workspace.

### Analysis

The error comes from `src/main/runtime/worktree-teardown.ts`:

```ts
// ~line 149
if (deps.requirePhysicalStop) {
  const failedPtyIds = stopResults.filter(([, stopped]) => !stopped).map(([ptyId]) => ptyId)
  const failedPtysExited =
    failedPtyIds.length === 0 ||
    (await verifyFailedPtysExited(failedPtyIds, deps.localProvider, deadline))
  if (!failedPtysExited) {
    throw new Error(`Failed to physically stop every PTY for worktree: ${worktreeId}`)
  }
```

I could not determine which of two things is happening, and I don't want to guess at a fix:

**(a) A daemon PTY really is still alive** with its cwd inside the worktree. The `EBUSY` on the now-empty directory is consistent with this. If so `requirePhysicalStop` is doing exactly its job and the real issue is that the PTY survives `orca terminal stop` reporting success — plus the fact that nothing surfaces it (`orca terminal list` shows nothing, no process matches the path), so a user has no way to find or kill it.

**(b) `verifyFailedPtysExited` fails closed on an unverifiable list.** It returns `false` when `listProcesses` throws or times out:

```ts
// ~line 173
const sessions = await settleBeforeDeadline(
  () => provider.listProcesses({ deadlineMs: deadline }), null, deadline
).catch(() => null)
if (!sessions) {
  return false
}
```

That conflates *"could not verify"* with *"still live"*. Combined with `--force` not reaching this gate, the worktree becomes permanently unremovable.

### What would help

Regardless of which one it is, the gap I ran into is that **there is no way out**. Some possibilities, in the maintainers' judgement:

- Have `--force` (or a separate explicit flag) bypass the physical-stop gate, with a loud warning about orphaned state.
- Surface *which* PTY ids are blocking removal in the error message, so they can be inspected or killed manually. Right now the error names only the worktree.
- Distinguish "verification unavailable" from "PTY still live" if (b) turns out to be the case.

Happy to gather more diagnostics if you can point me at the right logging — the worktree is still wedged on this machine, so I can run things against the live state.

OrcaWin:
Fixes #11960

## Summary

A workspace could become **permanently unremovable**. `orca worktree rm` failed with
`Failed to physically stop every PTY for worktree: …`, `--force` failed identically, and every
retry gave the same result — with no terminal listed and no OS process holding the path.

Root cause is (b) from the issue report: **`verifyFailedPtysExited` conflated "could not verify"
with "still live", and it was structurally guaranteed to never verify.**

Destructive removal sweeps PTYs under a 10s deadline. If a stop RPC fails, it re-lists the
provider to check whether that PTY had in fact already exited — but it re-listed *on the same
deadline the sweeps had just spent*:

```ts
const sessions = await settleBeforeDeadline(
  () => provider.listProcesses({ deadlineMs: deadline }), null, deadline  // ← already ~0ms left
).catch(() => null)
if (!sessions) return false   // "timed out" === "still alive"
```

`settleBeforeDeadline` returns the fallback immediately when `remaining <= 0`, so a sweep that
used most of its budget reached the re-list with nothing left, "timed out" without ever asking
the provider, and reported the PTY as live. The sweep spends that budget on every run, so the
failure is deterministic — which is exactly why retrying never helped.

The reporter's stale registry row (kill RPC returns `Session not found`) is precisely the case
this verification exists to forgive, and precisely the case it could never reach.

### ELI5

Orca refuses to delete a workspace folder until it is sure every terminal inside it is really
dead — otherwise it would delete files out from under a running process.

To be sure, it kills the terminals, then double-checks by asking "is that terminal still in your
list?" The bug: it gave itself a 10-second budget, spent all 10 seconds doing the killing, and
then tried to ask the double-check question with 0 seconds left. The question timed out
instantly — and Orca treated "I didn't get an answer" as "the terminal must still be alive."

So it refused to delete. Every retry did the same thing: spend the budget, fail to ask, refuse.
And the `--force` escape hatch didn't apply to this check, so the workspace was stuck forever —
even though the terminal had actually been dead the whole time.

The fixes: give the double-check its own time budget instead of leftovers; say "I couldn't
check" instead of "it's alive" when that is what really happened (and name the terminal); and
give the user a real way out — wired to the button that actually means "force", not to the
ordinary Delete button.

## What changed

1. **Verification gets its own budget** — sized like the sweep's (floored at 2s) rather than
   inheriting its exhausted remainder.
2. **Three distinct verdicts** — `exited` / `live` / `unverifiable`. The error names the blocking
   PTY ids and, when verification failed, *why*.
3. **A reachable escape hatch.** `allowUnverifiedPtyStop` waives the proof, plumbed preload →
   IPC → runtime → RPC (and through the web/mobile client). It is set **only** by genuine force
   affordances: the dialog's Force Delete branch, the failure toast's Force Delete action, the
   Space panel's force recovery, and the CLI's `--force`. A new `unstopped-pty` force-delete
   reason and toast copy make the button appear, so the first attempt fails, offers Force Delete,
   and the retry carries the waiver.
4. **Force also survives a sweep that cannot complete** (unresponsive daemon, dropped SSH
   channel), which fails before the gate is even reached — while the non-force path still fails
   fast, since it aborts without deleting anything.

## Screenshots

No visual change beyond one new failure-toast message (existing Force Delete button, new copy).

## Testing

- [x] `pnpm lint` — oxlint, code-quality audits, reliability gates, max-lines ratchet (no
      suppressions added), and all three localization gates pass. `verify:skill-bundle-manifest`
      reports stale generated skill artifacts; **verified identical on `origin/main` at
      `d426e35be`**, so it is pre-existing and unrelated.
- [x] `pnpm typecheck` (node + cli + web)
- [x] `pnpm test` — 4017 passing across the affected surface. The only failures are pre-existing
      on this Windows box (symlink permissions, `/bin/sh` spawns, SSH/codex home paths);
      spot-checked `src/relay/pty-handler.test.ts` at the pre-change commit — same 15 failures.
- [x] `pnpm build`
- [x] Every behavioural fix is **mutation-verified** — reverting it fails a named test. That
      includes the budget fix, the fixed-grace variant, each verdict, the force bypass at both
      the gate and the sweep, the classifier branch and its ordering, clear-only-on-proven-exit,
      and re-deriving the waiver from `force` at any of the ten teardown call sites.

## AI Review Report

Reviewed by Claude (Opus 5) plus four adversarial sub-agents (correctness, cross-platform,
data-loss safety, test quality) over **four rounds**, each finding re-verified against the code
before acting. Rounds 1–3 found real defects; round 4 found none in the product code.

Highest-severity findings, all fixed:

- **The waiver was wired to the wrong signal.** It aliased `force`, but the desktop's *plain*
  Delete button already passes `force: true` after the ordinary confirmation (to skip the
  dirty-file prompt). That silently disabled the PTY gate on the primary delete path. Fixed with
  a separate field; a mutation-verified guard asserts force-without-waiver keeps the gate strict.
- **Then decoupling it re-created the original dead end**: the classifier bailed on `force`
  before testing the PTY error, so the confirmed Delete hit the gate with no Force Delete button
  anywhere. The PTY case is now decided before the `force` guard and suppressed by the waiver.
- **`--force` could not escape a sweep-level failure** — the sweeps fail closed before the gate.
- **The web/mobile preload silently dropped the new field**, defeating the fix on paired clients.
  TypeScript could not catch it: the field is optional and destructuring a subset is legal.
- **Force with a live verdict unregistered a PTY just proven alive**, hiding it from the retry.
- **The forced path returned while sibling sweeps were still inside `shutdown()`**, racing the
  delete against a PTY about to release its handles.
- **A generic deadline sentinel masked the actionable error** in the log the user is left with.
- Both in-flight removal dedupe keys include the new flag, so a forced retry cannot coalesce onto
  the attempt that just failed the gate.

A flaky test was **root-caused, not papered over**: these tests advanced fake timers before
awaiting the teardown, so a rejection mid-advance was unhandled and vitest attributed it to
whichever test was running. Under the two-shell concurrent repro it went from 7/40 failing to
**0/40**, with zero unhandled rejections.

**Cross-platform, explicitly checked.** All SSH-bearing call sites keep `connectionId` through
the signature change (diffed one-for-one against `origin/main`), so no SSH removal can sweep the
local provider. The folder-workspace `::workspace:<uuid>` guard from #10252 is byte-identical to
main, and folder-workspace removal never sets `requirePhysicalStop` at all. Absolute-epoch
`deadlineMs` is correct for both the daemon adapter and the SSH provider. No path-separator,
shell, shortcut or label assumptions. CI runs native smoke jobs on macOS, Ubuntu and Windows
plus a Windows package build.

**Known trade-off:** a forced removal with a failed stop can take ~20s (10s sweep + up to 10s
verify) instead of failing at ~10s. Checked for an outer timeout this could exceed — the CLI's
`runtime-client` sets none, and the renderer RPC path allows 60s. The non-force path still fails
fast (measured at 1ms).

## Security Audit

- **No new untrusted input, command execution, path construction, auth, or secrets handling.**
- **Privilege:** the waiver cannot be reached without an explicit user force affordance. Every
  caller was traced — notably the background workspace-cleanup batch (which derives `force` from
  an unverified git status) and the failed-creation rollback paths pass no waiver, so the gate
  stays strict for all automated removals. Guarded behaviourally at the RPC and store bridges,
  and by a source invariant across all ten teardown call sites.
- **Path-safety guards unchanged and still upstream of every teardown call**:
  `isDangerousWorktreeRemovalPath`, `assertWorktreeDoesNotContainRegisteredWorktree`,
  `canSafelyRemoveOrphanedWorktreeDirectory`, `assertWorktreeCleanForRemoval`,
  `assertWorktreeUnlockedForRemoval`. PTY ownership matching was not widened.
- **Information disclosure:** the error now includes PTY ids, which are internal and embed the
  worktree path the message already contained.
- **DoS:** the extra verification window is bounded by the sweep budget and cannot loop.

## Notes

I could not reproduce against the reporter's live wedged worktree, so this is diagnosed from the
code path and reproduced in tests. The deterministic 0ms-budget mechanism accounts for every
symptom reported, so I am confident it is the cause, though hypothesis (a) (a genuinely surviving
daemon PTY) is not strictly excluded. If it were (a), Force Delete now provides the escape hatch
either way and the error names the PTY.

Deliberately **not** in scope, worth follow-ups:

1. **Mobile has no recovery path.** It sends `force: true` without the waiver, so it fails safe —
   but `mobile/app/h/[hostId]/index.tsx` swallows the error in a bare `catch {}`, so the row
   silently reappears with no explanation. Surface the error first; add the affordance second.
2. **The forced sweep-failure path returns before rendering a verdict**, so it cannot name live
   PTYs when only one sweep failed. Diagnosability only — it is behind an explicit Force Delete,
   deletes either way, and clears no registry rows. Recorded at the return site.
3. The reporter also saw `orca terminal stop` report success for a PTY that `orca terminal list`
   never showed. This PR makes such a PTY visible in the error rather than fixing that.

### DIFF
--- a/src/cli/handlers/worktree.ts
+++ b/src/cli/handlers/worktree.ts
@@ -280,6 +280,8 @@ export const WORKTREE_HANDLERS: Record<string, CommandHandler> = {
     const result = await client.call<RuntimeWorktreeRemoveResult>('worktree.rm', {
       worktree: await getRequiredWorktreeSelector(flags, 'worktree', cwd, client),
       force: flags.get('force') === true,
+      // Why (#11960): --force is explicit here, so it may also waive PTY-stop proof.
+      allowUnverifiedPtyStop: flags.get('force') === true,
       runHooks: flags.get('run-hooks') === true
     })
     printHookWarning(result.result, json)
--- a/src/main/ipc/worktrees.test.ts
+++ b/src/main/ipc/worktrees.test.ts
@@ -9840,9 +9840,28 @@ describe('registerWorktreeHandlers', () => {
     expect(callOrder).toEqual(['preflight', 'kill', 'git'])
   })
 
-  // Why (#11960): --force previously stopped at the same PTY gate as a normal
-  // delete, leaving a workspace with an unprovable PTY unremovable forever.
-  it('forwards force to the PTY gate so a forced delete has an escape hatch', async () => {
+  // Why (#11960): the PTY gate previously had no escape hatch at all, so a
+  // workspace with an unprovable PTY was unremovable forever.
+  it('forwards an explicit Force Delete to the PTY gate', async () => {
+    mockKnownFeatureWorktree()
+    getEffectiveHooksMock.mockReturnValue(null)
+
+    await handlers['worktrees:remove'](null, {
+      worktreeId: 'repo-1::/workspace/feature-wt',
+      force: true,
+      allowUnverifiedPtyStop: true
+    })
+
+    expect(killAllProcessesForWorktreeMock).toHaveBeenCalledWith(
+      'repo-1::/workspace/feature-wt',
+      expect.objectContaining({ requirePhysicalStop: true, allowUnverifiedStop: true })
+    )
+  })
+
+  // Why (#11960): the ordinary Delete confirmation already sets force:true to skip
+  // the dirty-file prompt. Waiving PTY-stop proof off that signal would silently
+  // disable the gate on the primary delete path.
+  it('keeps the PTY gate strict for a confirmed delete that only sets force', async () => {
     mockKnownFeatureWorktree()
     getEffectiveHooksMock.mockReturnValue(null)
 
@@ -9853,11 +9872,11 @@ describe('registerWorktreeHandlers', () => {
 
     expect(killAllProcessesForWorktreeMock).toHaveBeenCalledWith(
       'repo-1::/workspace/feature-wt',
-      expect.objectContaining({ requirePhysicalStop: true, allowUnverifiedStop: true })
+      expect.not.objectContaining({ allowUnverifiedStop: true })
     )
   })
 
-  it('keeps the PTY gate strict for a non-forced delete', async () => {
+  it('keeps the PTY gate strict for a plain delete', async () => {
     mockKnownFeatureWorktree()
     getEffectiveHooksMock.mockReturnValue(null)
 
--- a/src/main/ipc/worktrees.test.ts
+++ b/src/main/ipc/worktrees.test.ts
@@ -9840,6 +9840,37 @@ describe('registerWorktreeHandlers', () => {
     expect(callOrder).toEqual(['preflight', 'kill', 'git'])
   })
 
+  // Why (#11960): --force previously stopped at the same PTY gate as a normal
+  // delete, leaving a workspace with an unprovable PTY unremovable forever.
+  it('forwards force to the PTY gate so a forced delete has an escape hatch', async () => {
+    mockKnownFeatureWorktree()
+    getEffectiveHooksMock.mockReturnValue(null)
+
+    await handlers['worktrees:remove'](null, {
+      worktreeId: 'repo-1::/workspace/feature-wt',
+      force: true
+    })
+
+    expect(killAllProcessesForWorktreeMock).toHaveBeenCalledWith(
+      'repo-1::/workspace/feature-wt',
+      expect.objectContaining({ requirePhysicalStop: true, allowUnverifiedStop: true })
+    )
+  })
+
+  it('keeps the PTY gate strict for a non-forced delete', async () => {
+    mockKnownFeatureWorktree()
+    getEffectiveHooksMock.mockReturnValue(null)
+
+    await handlers['worktrees:remove'](null, {
+      worktreeId: 'repo-1::/workspace/feature-wt'
+    })
+
+    expect(killAllProcessesForWorktreeMock).toHaveBeenCalledWith(
+      'repo-1::/workspace/feature-wt',
+      expect.not.objectContaining({ allowUnverifiedStop: true })
+    )
+  })
+
   it('does not start Git removal when physical PTY teardown cannot be proven', async () => {
     mockKnownFeatureWorktree()
     getEffectiveHooksMock.mockReturnValue(null)
--- a/src/main/ipc/worktrees.ts
+++ b/src/main/ipc/worktrees.ts
@@ -149,6 +149,8 @@ type RemoveWorktreeArgs = {
   worktreeId: string
   hostId?: ExecutionHostId
   force?: boolean
+  /** Explicit Force Delete only — `force` alone is set by the ordinary confirmation (#11960). */
+  allowUnverifiedPtyStop?: boolean
   skipArchive?: boolean
 }
 
@@ -157,9 +159,9 @@ type DetectedWorktreeRequestArgs = { repoId: string } | ListDetectedWorktreesArg
 async function stopPtysForDestructiveWorktreeRemoval(
   runtime: OrcaRuntimeService,
   worktreeId: string,
-  options: { connectionId?: string; force?: boolean } = {}
+  options: { connectionId?: string; allowUnverifiedStop?: boolean } = {}
 ): Promise<void> {
-  const { connectionId, force } = options
+  const { connectionId, allowUnverifiedStop } = options
   const provider = connectionId ? getSshPtyProvider(connectionId) : getLocalPtyProvider()
   if (!provider) {
     throw new Error(`PTY provider unavailable for worktree deletion: ${worktreeId}`)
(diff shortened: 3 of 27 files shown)


AmethystLiang:

Issue #11332, opened 2026-07-29
[Bug]: Korean IME composition intermittently breaks in the session chat input — syllables commit as standalone compatibility jamo (macOS desktop)


Yoon-sua:
### Operating system

macOS

### Orca version

1.4.160

### Details

**Short summary**

In the native session chat input on the macOS desktop app, Korean composition intermittently breaks and each keystroke commits as a standalone jamo instead of being composed into a syllable.

Typed: `테스트`  <br>Received: `ㅌㅔㅅㅡㅌㅡ`

It does not happen on every sentence, but it happens often enough to get in the way of normal use — once it starts I have to clear the field and retype.

**What happened?**

The IME composition for the syllable never completes. Instead of `ㅌ` + `ㅔ` composing into `테`, both jamo land in the input as separate characters, and the same thing continues for the rest of the word.

Worth noting for triage: the resulting characters are **Hangul Compatibility Jamo**, not conjoining jamo —

```
ㅌㅔㅅㅡㅌㅡ  →  U+314C U+3154 U+3145 U+3161 U+314C U+3161
```

Compatibility jamo is what the macOS 2-set keyboard emits per keystroke *before* composition combines them. So this looks like composition being aborted / never started, with each keystroke committed on its own — a different mechanism from the NFD decomposition in stablyai/orca#11235, where a syllable arrives already composed but as conjoining jamo (`U+110B U+1161 U+11AB`).

**How can we reproduce it?**

1. Switch the macOS system input source to Korean (2-set / 두벌식).
2. Open a Claude Code session and click into the session chat input (the native chat UI, not a terminal).
3. Type Korean at normal speed, e.g. `테스트`.
4. Intermittently the syllables do not compose and the field shows `ㅌㅔㅅㅡㅌㅡ`.

Expected: `테스트`  <br>Actual: `ㅌㅔㅅㅡㅌㅡ`

**Anything else that might help**

* Intermittent, not deterministic. It correlates with typing speed and seems more likely right after focusing the field or after the UI re-renders, which suggests the `compositionstart` / `compositionupdate` / `compositionend` sequence is racing with a re-render or a value write-back into the input.
* Latin input in the same field has no issue, so this is specific to IME-composed text and likely affects Japanese/Chinese input as well.
* Environment: macOS 15.7.7 (24G720), Orca desktop 1.4.160, built-in macOS Korean IME (2-set).

### Related

[...]

PR #12158, opened 2026-08-03
fix(terminal): stop Hangul input lagging a syllable behind the IME

leon1114:
## Summary

Korean input in the terminal lags one syllable behind the IME.

Typing `한글` after a space renders `한`, then the next composition (`ㄱ`, `ㅡ`) appears to overwrite it, and the display only settles once `글` commits and both syllables appear. Reported after updating to v1.4.163.

The committed syllable was never sent. #11293 made `compositionstart` cancel the pending finalizer and route the just-committed syllable through `_endPendingCompositionSession`, which announces the commit with `dataPendingReconciliation: true` but never emits the bytes — and `terminal-ime-composition-route.ts:92-101` drops exactly that event. The macOS Hangul IME commits syllable N at syllable N+1's `compositionstart` and sends no intervening `insertText`, so syllable N stayed stranded until the next `compositionend`. With no bytes on the wire the buffer cursor never advanced, so xterm drew the next composition's opaque `.composition-view` overlay on the previous syllable's cell.

This drops that block from `compositionstart` so the already-armed finalizer survives. It fires on the next task, which is:

- **after** IBus rewrites the textarea, so the Korean final-consonant transfer that #11293 exists to protect still yields `아아` and not `앙아`;
- **long before** the next `compositionend`, so on macOS the syllable flushes immediately and the lag is gone.

One mechanism covers both platforms. This is effectively upstream xterm's original flush timing restored, while every piece of transaction bookkeeping #11293 added — transaction ids, the pending record, tracked timers, stale-end rejection, accepted/settled balance — is left intact.

Note the deleted block's own `_cancelPendingFinalizer` is not lost: `_sendPendingComposition` cancels the finalizer as its first action.

## Screenshots

https://github.com/user-attachments/assets/0420208a-a5c4-4205-b165-d5093b11cc9c

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — see the note below about 5 pre-existing macOS-only failures
- [x] `pnpm build`

New file `src/renderer/src/components/terminal-pane/terminal-ime-hangul-syllable-flush.test.ts` — 4 cases against both the CJS and ESM terminal builds (8 tests):

1. **`flushes syllable N before syllable N+1 finishes composing`** — the regression gate. Asserts `emitted` is already `한` right after the second `compositionstart`, before any second `compositionend`.
2. **`preserves Korean final-consonant transfer`** — `아아`, never `앙`.
3. **`emits 한글 exactly once`** — no duplication at the syllable boundary.
4. **`does not double-send when an insertText input event follows compositionend`** — the IBus shape.

**Why the existing suites missed this.** `terminal-ime-xterm-adversarial.test.ts:114-126` already replays this exact sequence, but asserts only the joined string after the *second* `compositionend`, so a withheld-then-late-flushed syllable passes identically to a promptly-flushed one. Every new assertion here is on intermediate state.

**Red control.** With the fix reverted in the installed bundle, cases 1 and 2 fail on both builds with `expected '' to be '한'` and `expected '' to be '아'` — i.e. nothing has reached the terminal at the moment the next syllable starts composing, which is the bug.

**Regression runs.** Full IME suite 244/244; whole `terminal-pane` unit suite 2894/2894 across 219 files, including `terminal-ime-xterm-composition-deduplication.test.ts` (the #11293 `아아`/`앙아` and duplicate-commit guarantees) and `terminal-ime-xterm-transaction-events.test.ts` (the strict `accepted, settled, accepted, settled` ordering).

**An earlier, larger version of this fix was rejected by your own tests.** Flushing eagerly inside `compositionstart` (right after `nextCompositionStart` is set) fixes macOS but breaks IBus: `preserves a Korean final-consonant transfer across compositions` fails with `expected '앙아' to be '아아'`, because IBus rewrites the textarea in the `compositionupdate` that *follows* `compositionstart`, so a synchronous read there sees the pre-transfer `앙`. Deferring by one task is what makes a single fix correct on both platforms.

## AI Review Report

**Cross-platform (macOS / Linux / Windows).** The change is in shared `CompositionHelper` code with no platform guard, so all three were considered explicitly. macOS is the reported break. Linux/IBus is the constraint that shaped the fix — an earlier variant regressed it and was discarded on that evidence (above). Windows IME ownership tests are part of the 244-test IME suite and pass. No keyboard shortcut, shortcut label, filesystem path, or shell invocation is touched, so the `metaKey`/`CmdOrCtrl`, `⌘`/`Ctrl+`, and `path.join` rules do not apply here. The one Electron-specific surface nearby — `xterm-bypass-policy.ts:127` suppressing `isComposing` keydowns, which is why the keydown flush path is unavailable during Hangul composition — was read but not modified.

**SSH / remote / local.** Composition bytes now reach the pty through `terminal-ime-composition-route.ts` → `terminal.input()` → xterm's existing ordered `onData` transport, the same path all other input uses. No host-local assumption is introduced and no transport code changed, so local ptys, SSH/relay runtimes, git worktrees, and folder workspaces are all unaffected structurally.

**Agent / integration / git provider.** None touched.

**Performance.** Net removal of work on a hot input path: three calls per `compositionstart` are gone and nothing is added. No new timer is created — the fix reuses the finalizer that was already armed, and the tracked-timer set bookkeeping is unchanged, so the bounded-timer guarantee from #11293 holds. `terminal-ime-xterm-adversarial.test.ts`'s 256-transaction same-task burst still passes.

**UI quality.** No component, token, or style changed; `docs/STYLEGUIDE.md` is not engaged. The visible effect is that the terminal now renders what the user typed.

**Security.** See below.

**What the review flagged and what I did.** It flagged that the originally-drafted eager flush was unsafe on IBus; I dropped it and shipped the smaller variant, then confirmed the difference with a red control rather than reasoning alone. It also flagged the two dead surfaces listed under Notes, which I deliberately left in place to keep the diff minimal rather than expanding scope.

## Security Audit

No new input sanitization, command execution, path handling, auth, secrets, or IPC surface. Concretely:

- **Input handling — improved.** Previously the stranded syllable was eventually emitted by xterm's bare `_coreService.triggerDataEvent`, bypassing Orca's route because `dispatchSessionEnd` was false once `_endPendingCompositionSession` had set `sessionEnded`. With that call gone, `pending.sessionEnded` stays false, so the flush dispatches a normal session-end that `terminal-ime-composition-route.ts` handles — meaning the bytes are now validated against the captured pty id and the current transport (`:96-100`) before being written, instead of going out unchecked.
- **Command execution / path handling / auth / secrets:** untouched.
- **IPC:** untouched. The composition events are DOM `CustomEvent`s on the xterm textarea, entirely renderer-local.
- **Dependencies:** no version changed. `pnpm-lock.yaml`'s only diff is the `@xterm/xterm` patch content hash and its 24 dependent `patch_hash=` references, which necessarily follow any edit to the vendored patch.

No follow-up security work identified.

## Notes

**Vendored artifacts — needs a maintainer decision.** `config/patches/@xterm__xterm@6.1.0-beta.287.patch` carries both the TypeScript source and the prebuilt `lib/xterm.js` / `lib/xterm.mjs` bundles, and the runtime loads `lib/`. I applied the same semantic change to all three by exact anchored string replacement (asserted to match exactly once per bundle), but I did **not** regenerate the bundles from upstream xterm source the way PR #11293 did. Consequently:

- `lib/xterm.js.map` / `lib/xterm.mjs.map` were left untouched, so their `mappings` offsets are now stale relative to the edited bundles.
- The patch's `index <sha>..<sha>` blob lines for the two bundles still name the pre-edit b-side blobs. `pnpm install` applies the patch cleanly regardless.

If you would rather land this with properly regenerated artifacts, say so and I will redo it against a clean upstream checkout — or feel free to regenerate on your side, since the semantic change is five deleted lines.

**Two surfaces are now unreachable.** `_endPendingCompositionSession` has zero call sites, and with nothing dispatching `dataPendingReconciliation` the guard at `terminal-ime-composition-route.ts:94` is a defensive no-op. Both were left in place deliberately to keep the reviewable diff to the semantic change and to keep the patch's TypeScript source and minified bundles mirrored. Happy to remove them in this PR or a follow-up, whichever you prefer.

**Local `pnpm test` had 5 failures, all pre-existing and macOS-specific.** Neither file is touched by this PR and both reproduce on a clean tree:

- `config/scripts/check-root-directory-entries.test.mjs` (3) — `.github/scripts/check-root-directory-entries.sh:14` uses `declare -A`, which macOS's system `/bin/bash` 3.2.57 does not support, so the script exits 2 instead of 1. The test spawns bare `bash`, so it can only pass where bash is 4+. Worth a separate fix (`#!/usr/bin/env bash` plus a bash-4 lookup, or an explicit skip on macOS).
- `src/relay/agent-exec-handler.test.ts` (2) — asserts on the spawned process environment and picks up `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_*` injected by my shell session.

**Testing notes.** Manually verified on macOS with the built-in Korean (2-Set) input source, running the branch through `pnpm dev`: typing `한글` into a terminal pane now renders correctly, where before the second syllable's preview covered the first. The `아아`/`앙아` final-consonant case is covered by automated tests rather than that manual pass.

Playwright e2e was not run locally (needs a GUI/Electron runtime); the nightly `terminal-ime-e2e.yml` ibus-hangul job should exercise the Linux path. A manual pass on a real Windows IME would be welcome from anyone set up for it.

**X handle:** I don't have an X account, so there's no handle to credit — noting it explicitly rather than leaving the field blank.


leon1114:
Addressed the review feedback in `fed2ba2`:

- **CodeRabbit nitpick (unconditional dispose):** terminals opened by `openTerminal` are now tracked and disposed in `afterEach`, so a failed assertion can no longer leak xterm timers into the next test in the file. The inline `terminal.dispose()` calls are gone.
- **"Out of Scope Changes" inconclusive check:** confirming the vendored bundle edits *are* in this PR. GitHub reports three changed files:

| File | +/- |
| --- | --- |
| `config/patches/@xterm__xterm@6.1.0-beta.287.patch` | +7 / -10 |
| `pnpm-lock.yaml` | +24 / -24 |
| `src/renderer/src/components/terminal-pane/terminal-ime-hangul-syllable-flush.test.ts` | +164 / -0 |

The patch hunk carries the same semantic change in all three places it has to: the TypeScript source (`src/browser/input/CompositionHelper.ts`), the CJS bundle (`lib/xterm.js`), and the ESM bundle (`lib/xterm.mjs`). The summarizer likely skipped that file because its bundle hunks are single lines of a few hundred KB. The `pnpm-lock.yaml` diff is only the patch content hash and its dependent `patch_hash=` references.

Re-ran after the change: the new file is 8/8, and the full `terminal-pane` unit suite is 2894/2894 across 219 files.

Daworks:
Verified this fix on macOS. Summary: **still reproduces on current `main`, and applying only
your patch file fixes it.**

Since this branch is 80 commits behind `main` (merge-base `c5c10e1`), I didn't test the branch
as-is — a pass there wouldn't distinguish your patch from the 80 intervening commits. Instead I
built current `main` and then applied **only** `config/patches/@xterm__xterm@6.1.0-beta.287.patch`
from this PR (byte-identical to yours), so the two builds differ by nothing else.

### Environment

| Item | Value |
|------|-------|
| Control | `main` @ `18dbcf0`, `pnpm dev` |
| Fix | same tree + this PR's patch file only (`+7/-10`) |
| macOS | 26.5.2 (25F84), Apple Silicon |
| Input source | System 2-Set Korean (`com.apple.inputmethod.Korean.2SetKorean`) |
| Also reproduces on | released 1.4.164 |

The two builds resolve different patched packages, so the patch definitely took effect:

```
control  node_modules/@xterm/xterm -> ...patch_hash=082a4e71...  lib/xterm.js sha256 605f406c...
fix      node_modules/@xterm/xterm -> ...patch_hash=a08b04de...  lib/xterm.js sha256 47886efd...
```

### Result

Typing `가나다` at a bare `zsh` prompt. Measured the `.composition-view` anchor x-coordinate
via CDP on each composition event. Dev builds already expose CDP on `--remote-debugging-port=9379`
(the dev runner sets it), so this is reproducible in about a minute without patching anything in.

Anchor x after each syllable commits, starting from the same baseline column (301):

| After committing | Control | With this patch |
|---|---|---|
| 가 | **301** (no advance) | **317** (+16) |
| 나 | **301** (no advance) | **333** (+16) |

Control: the anchor never advances across three syllables, so each new composition is painted
over the previous syllable's cell — matching your description. With the patch the anchor
advances exactly one syllable width per commit, and the composing jamo render normally
(confirmed visually as well as by measurement).

(The fix build's probe was installed mid-word, so the first syllable's `compositionstart` was
not captured there; both runs are compared from the 301 baseline the first composition occupies.)

<details>
<summary>Measurement caveat</summary>

The probe listens on `document` in the capture phase, which runs *before* xterm's own handler,
so a snapshot taken there reflects the previous event's state and shows a spurious one-event
lag. The numbers above are from a second snapshot taken on the next macrotask
(`setTimeout(…, 0)`), after xterm has processed the event. An earlier run of mine that showed
`301 → 301 → 317` was contaminated by this artifact — disregard that shape if it matters.

Also worth noting: the `xterm-helper-textarea` value accumulates to `가나다` in **both**
builds, so that is not a signal for this bug — the anchor position is the only discriminator I
found.
</details>

### Not covered

I did not verify Orca's own chat composer (`native-chat/`). It's a separate path that this PR
doesn't touch, and #11332 describes that one as *intermittent*, so a single passing sample
wouldn't mean anything. Treating it as out of scope here.

OrcaWin:
Superseded by #12278, which adopts this fix, rebuilds the pinned xterm CJS/ESM bundles and source maps reproducibly, and credits Hyunggyun Lyou as co-author. Thank you @leon1114 for isolating the one-syllable-behind finalizer regression and validating the IBus constraint.

PR #12159, opened 2026-08-03
fix(native-chat): stop clipping assistant text blocks at the tool-preview cap

### LINKED ISSUE #12138: [Bug]: Mobile Chat UI truncates assistant messages at exactly 4,000 characters with no expand control (desktop unaffected)
## Summary

In the Chat UI, long assistant messages are cut short and replaced with a `... (truncated)` marker. It happens on **both the desktop client and mobile**, and with **both fenced code blocks and ordinary prose**. The full text is present in the agent's own transcript on the host, so this is a display/transport truncation rather than anything lost agent-side.

## Environment

- Orca **v1.4.162** (Linux AppImage, `orca-linux.AppImage` dated 2026-07-31)
- Runtime: **headless** on a VPS — `orca-ide --no-sandbox --serve --serve-json --serve-port --serve-pairing-address`
- Viewed from a **paired desktop client** and from **mobile**
- Agent: Claude Code
- Host: Ubuntu, `DISPLAY=:99` (Orca-spawned Xvfb), no desktop environment

## Steps to reproduce

1. Pair a desktop or mobile client to a headless `orca serve` runtime.
2. Run a Claude Code session and have it emit a long assistant message (the observed one was **5,918 characters**, mixed prose plus one fenced code block at the end).
3. Read the message in the Chat UI.

## Observed

The message renders, then stops. The visible text ends after **exactly 60 characters** of the code block:

`Continue lessons-context-architecture per plans/lessons-cont ... (truncated)`

Three details that look diagnostic:

- The cut is at **exactly 60 characters**, and it lands **mid-word** (`lessons-cont` | `ext-architecture`).
- The source had a **newline** after `per`; in the rendered output that newline is **collapsed to a space**. That is preview/summary behaviour, not a streaming cut.
- `... (truncated)` appears to be a deliberate marker rather than a dropped stream.

## Expected

The full message is displayed, or at minimum an explicit expand/"show more" control that reveals the rest. A 60-character window with no way to read on makes long agent replies unusable.

## Why I think this is transport-side, not renderer-side

**Desktop and mobile are two independent renderers, and both truncate identically.** The element they share is the paired headless runtime. That points at the publication/serialisation layer feeding both clients rather than at either renderer — so I'd start there rather than in the markdown components.

There appears to be an existing family of paired-headless-Remote-Orca-Server Chat UI bugs that this may sit alongside: #11761, #9452, #10923.

## What I could not determine

- The threshold at which **plain prose** truncates. The 60-character figure is measured from the code-block instance; the reporter confirms plain text is affected too, but I do not have a measured length for that case.
- Whether a non-paired (local, non-headless) runtime is affected. This setup is always paired to a remote host, so I could not test the local case.

## Not a duplicate of

Searched before filing. The six issues matching `truncat*` in the title are about copy-context output, PTY injection, project titles, branch names, tab titles, and `terminal read` pagination — none about chat message rendering. The v1.4.159 native-chat fix (*"wrap question text and option descriptions instead of truncating"*) is already in this build and concerns question cards, not message bodies.

brennanb2025:
## Summary

Long assistant messages viewed over a paired connection (headless `orca serve` viewed from a paired desktop client or the mobile app) were cut short with a `… (truncated)` marker and no way to read the rest (STA-3230, GitHub #12138). The mobile payload diet in the `nativeChat.readSession`/`nativeChat.subscribe` runtime RPC applied the 4,000-char tool-preview cap to `text` blocks — the fully rendered message body — for any connection whose paired device token has scope `mobile` (which includes desktops that redeem a `--serve-mobile-pairing` offer, so both clients truncated identically). The reporter's "exactly 60 chars of the code block" was `4000 − prose length`; the collapsed newline was the broken markdown left by a mid-code-block cut.

Fix at the source, shared by both clients: `text` blocks now get their own generous 64k-char safety ceiling (real replies pass through whole), while tool-result/tool-call bodies keep the 4,000-char preview diet that the cap was designed for. No renderer changes needed on desktop or mobile.

## Screenshots

No visual change (transport fix; the chat views simply receive the full message text they already render).

## Testing

- [x] `pnpm lint` (oxlint clean on the repo sweep)
- [x] `pnpm typecheck`
- [x] `pnpm test` (targeted: `vitest run --config config/vitest.config.ts src/main/runtime/rpc/methods/native-chat.test.ts` — 18/18, including 2 new regression tests)
- [x] `pnpm build` (`build:electron-vite` + `build:cli`, used for the live repro/verify rig)
- [x] Added or updated high-quality tests that would catch regressions: a >4k assistant text block must pass through unclipped for `mobile` clients, and a 70k pathological block must still clip at the ceiling.

Live end-to-end verification on an isolated `orca serve` rig (temp `HOME` + `ORCA_USER_DATA_PATH`; the primary daemon untouched), with a planted Claude transcript engineered so the 4,000-char boundary lands 60 chars into a trailing fenced code block, mirroring the ticket exactly:
- Before: mobile-scope device token → text length 4,014 ending `plans/lessons-cont\n… (truncated)`; runtime-scope token → full 4,149 chars. Reproduced identically through a real paired desktop instance (`addFromPairingCode` → `runtimeEnvironments.call`, the same bridge the chat view's transport uses).
- After: same mobile-scope token and same paired desktop both receive the full 4,149-char message with the closing fence intact and no marker.

## AI Review Report

Reviewed the change for: regression risk to the mobile payload diet (tool-result/tool-call clipping is byte-for-byte unchanged, verified by the existing 16 tests), wire-size risk from unclipped text (bounded by the new 64k ceiling, the 40-message window, and the upstream 2MB transcript-record cap), and parity with the desktop IPC path (which has never clipped). Confirmed both the read and subscribe paths flow through the same `clipBlock`, so snapshots, replacements, and appends all benefit. Cross-platform: the change is pure main-process string handling with no shortcuts, labels, paths, shell, or Electron platform APIs; it is identical on macOS, Linux, and Windows hosts and for SSH/WSL-backed runtimes, and applies equally to folder workspaces and git worktrees.

## Security Audit

No new input surfaces: the RPC schema is unchanged, and the modified code only alters the char budget applied to already-parsed transcript strings before serialization. No command execution, path handling, auth, secrets, or dependency changes. The clip remains a bound (now 64k for text), so a hostile transcript still cannot ship unbounded payloads to paired clients; tool-input sanitization (depth/node/key budgets) is untouched. IPC/RPC gating by `clientKind` is unchanged.

## Notes

- The 60-char figure in the ticket was a red herring — it is the remainder of the reporter's ~3,940-char prose against the 4,000-char cap.
- Related follow-up (not in scope): a desktop client that redeems a mobile-scope pairing offer is treated as a `mobile` client for every payload-diet decision, not just this one. If we want full-fidelity desktop behavior over mobile-scoped pairings, scope negotiation deserves its own ticket.

Fixes STA-3230 (https://linear.app/stably/issue/STA-3230). Closes #12138.

### DIFF
--- a/src/main/runtime/rpc/methods/native-chat.test.ts
+++ b/src/main/runtime/rpc/methods/native-chat.test.ts
@@ -83,6 +83,10 @@ function makeMessage(text: string): NativeChatMessage {
   }
 }
 
+function makeTextMessage(text: string): NativeChatMessage {
+  return { ...makeMessage(''), blocks: [{ type: 'text', text }] }
+}
+
 function readSessionHandler(): (params: unknown, ctx: RpcContext) => Promise<unknown> {
   const method = NATIVE_CHAT_METHODS.find((m) => m.name === 'nativeChat.readSession')
   if (!method) {
@@ -146,6 +150,35 @@ describe('nativeChat.readSession clientKind truncation gating', () => {
     const output = firstOutput(result)
     expect(output.length).toBeLessThan(OVERSIZED.length)
     expect(output).toContain('truncated')
+  })
+
+  // STA-3230: a paired desktop or mobile client saw long assistant replies cut
+  // at the tool-preview cap with `… (truncated)` and no way to read the rest.
+  it('passes a long assistant text block through unclipped for mobile clients', async () => {
+    const text = 'prose '.repeat(1000)
+    cachedResult.value = { messages: [makeTextMessage(text)] }
+    const result = await readSessionHandler()(
+      { agent: 'claude', sessionId: 's' },
+      ctxWith('mobile')
+    )
+    const block = (result as { messages: NativeChatMessage[] }).messages[0].blocks[0] as {
+      text: string
+    }
+    expect(block.text).toBe(text)
+  })
+
+  it('clips a pathological text block at the safety ceiling for mobile clients', async () => {
+    const text = 'y'.repeat(70_000)
+    cachedResult.value = { messages: [makeTextMessage(text)] }
+    const result = await readSessionHandler()(
+      { agent: 'claude', sessionId: 's' },
+      ctxWith('mobile')
+    )
+    const block = (result as { messages: NativeChatMessage[] }).messages[0].blocks[0] as {
+      text: string
+    }
+    expect(block.text.length).toBeLessThan(text.length)
+    expect(block.text).toContain('truncated')
   })
 
   it('bounds raw tool-call inputs before sending them to mobile', async () => {
--- a/src/main/runtime/rpc/methods/native-chat.test.ts
+++ b/src/main/runtime/rpc/methods/native-chat.test.ts
@@ -148,8 +148,7 @@ describe('nativeChat.readSession clientKind truncation gating', () => {
       ctxWith('mobile')
     )
     const output = firstOutput(result)
-    expect(output.length).toBeLessThan(OVERSIZED.length)
-    expect(output).toContain('truncated')
+    expect(output).toBe(`${OVERSIZED.slice(0, 4000)}\n… (truncated)`)
   })
 
   // STA-3230: a paired desktop or mobile client saw long assistant replies cut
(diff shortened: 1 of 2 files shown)


brennanb2025:
Electron QA (isolated paired-desktop instance) passed for STA-3230.

- A 7,466-character assistant text block crossed the former 4,000-character tool-preview boundary.
- The remote Native Chat UI rendered the complete response, including the closing code fence and sentinel, with no truncation marker.

Start of the long response:

STA-3230 long assistant response

Completion sentinel preserved:

STA-3230 completion sentinel

Issue #11874, opened 2026-08-01
[Bug]: codex-windows-sandbox-setup.exe "cannot find" popup — resourcesDir never installed alongside codex.exe (related to #7555)


vex1br:
### Operating system
Windows

### Orca version
1.4.163

### Details

**Summary**

Codex agent sandboxing on Windows repeatedly triggers the native OS dialog:

> **codex-windows-sandbox-setup.exe**
> Windows cannot find 'codex-windows-sandbox-setup.exe'. Make sure you typed the name correctly, and then try again.

This is `ShellExecute`'s built-in `ERROR_FILE_NOT_FOUND` dialog, not an Orca/Codex-authored error — it fires because the helper binary is genuinely absent from every location the running `codex.exe` could resolve it from. Related to #7555 (same visible popup, different confirmed root cause — see "Relation to #7555" below).

**Environment**
- Orca: 1.4.163
- OS: Windows 11 Pro (10.0.26200), x64
- Codex CLI (active): 0.146.0, standalone release, `x86_64-pc-windows-msvc`
- `config.toml`: `[windows] sandbox = "elevated"`

**What happens**

While running any agent task in Orca, Windows shows the popup above, and Codex's own `sandbox.<date>.log` (under `%APPDATA%\orca\codex-runtime-home\home\.sandbox\`) logs, on essentially every tool call:

```
[... codex.exe] sandbox setup required: sandbox setup marker missing or incompatible
```

...repeating indefinitely, because the setup never succeeds and the marker is never written.

**Root cause (confirmed via filesystem + hash evidence)**

Every standalone Codex release ships `codex-package.json` declaring a 3-directory layout:

```json
{
  "entrypoint": "bin/codex.exe",
  "resourcesDir": "codex-resources",
  "pathDir": "codex-path"
}
```

i.e. `bin/`, `codex-resources/`, and `codex-path/` as **sibling** directories under `~/.codex/packages/standalone/releases/<version>-x86_64-pc-windows-msvc/`. `codex-windows-sandbox-setup.exe` lives in `codex-resources/`, alongside `codex-command-runner.exe`.

[...]

Issue #7555, opened 2026-07-06
[Bug][Windows] codex-windows-sandbox-setup.exe "The specified module could not be found" when a hosted Claude Code agent invokes Codex (works when Orca runs Codex directly)


Cainiaooo:
### Environment
- Orca: 1.4.123
- OS: Windows 11 (10.0.26100)
- Codex CLI: 0.142.5 (npm global `@openai/codex`, win32-x64)
- Codex Windows sandbox: `[windows] sandbox = "elevated"` (default), copied verbatim into Orca's managed CODEX_HOME

### Summary
Running Codex **directly as an Orca agent** (the Codex TUI in a worktree) works fine.
But when a **Claude Code agent hosted by Orca** invokes Codex — e.g. via the OpenAI
Codex Claude Code plugin (which runs `codex app-server`), or a direct `codex exec`
from Claude Code's shell — Windows repeatedly shows:

```
codex-windows-sandbox-setup.exe — "The specified module could not be found"
```

It fires on every sandbox setup/refresh, so it pops up many times during a session.
Both Codex invocations run under the **same** Orca managed CODEX_HOME
(`%APPDATA%\orca\codex-runtime-home\home`) with the **same** copied
`[windows] sandbox = "elevated"` config — the only thing that differs is *who spawns
Codex*.

### Repro
1. Windows 11, Codex 0.142.5, `[windows] sandbox = "elevated"` (default).
2. In Orca, run Codex **directly** as an agent → works, no popup.
3. In Orca, run a **Claude Code** agent, and have it invoke Codex (the Codex Claude
   Code plugin's `codex app-server`, or `codex exec` from Claude Code's shell) →
   repeated `codex-windows-sandbox-setup.exe "The specified module could not be
   found"` popups.

### What I confirmed (source + logs)
- The managed CODEX_HOME is separate from `~/.codex`
  (`src/main/codex/codex-home-paths.ts` → `getOrcaManagedCodexHomePath`).
- Orca copies `~/.codex/config.toml` verbatim into it
  (`src/main/codex-accounts/service.ts` → `syncCanonicalConfigIntoManagedHome`), so
  `[windows] sandbox = "elevated"` is identical to the working direct setup.
- Orca does **not** sync the `.sandbox` state (only `skills/hooks/plugins/
  plugin-state/profile-v2/themes/prompts` — `CODEX_SYSTEM_RESOURCE_ENTRIES` in
[...]

PR #12160, opened 2026-08-03
fix(codex): restore the Windows sandbox helper next to a flattened codex.exe

OrcaWin:
## Summary

On Windows, Codex agents repeatedly popped a native OS dialog:

> **codex-windows-sandbox-setup.exe** — Windows cannot find 'codex-windows-sandbox-setup.exe'. Make sure you typed the name correctly, and then try again.

Every standalone Codex release declares a three-directory layout in `codex-package.json` — `bin/`, `codex-resources/` and `codex-path/` as siblings. `codex-windows-sandbox-setup.exe` lives in `codex-resources/`. Installers that flatten a release by copying only `bin/` (e.g. `%LOCALAPPDATA%\Programs\OpenAI\Codex\bin\codex.exe`) leave that helper unreachable from the running `codex.exe`. Because Codex's elevated Windows sandbox launches the helper through `ShellExecuteW`, the OS shows its own `ERROR_FILE_NOT_FOUND` dialog — bypassing Codex's and Orca's error handling — and the sandbox setup marker is never written, so it fires again on every sandboxed tool call.

Before each native Codex launch, Orca now checks the resolved entrypoint's package root and restores any missing declared sibling directory from a donor whose `bin/codex.exe` is byte-identical to the one being launched. Donors are the untouched copies already on the host: cached standalone releases under `~/.codex/packages/standalone/releases/` and the npm platform package's `vendor/<target>/` layout.

Fixes STA-3125. Reported as #11874 (related to #7555, which is a different mechanism — helper present but `ERROR_MOD_NOT_FOUND` under a console-less nested spawn — and is not addressed here).

Behavior notes:

- Windows-only: `repairCodexWindowsPackageLayout()` returns `not-applicable` immediately on macOS/Linux, before touching the filesystem or resolving the Codex command.
- Only acts on a release-layout entrypoint (`<root>/bin/codex.exe`). A bare `codex`, a `.cmd`/`.ps1` shim, or an `.exe` outside a `bin/` directory is left alone.
- Only ever adds missing directories. Nothing is overwritten or deleted, and a healthy install costs two `stat` calls.
- Refuses to donate from a release built from a different Codex binary, so a mismatched helper is never installed. If no matching donor exists it logs a warning and leaves the install untouched.
- WSL launches are untouched (the repair runs after the WSL branch returns).

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — ran `oxlint` across the repo: clean.
- [x] `pnpm typecheck` — `tsc --noEmit -p config/tsconfig.node.json`: clean.
- [x] `pnpm test` — targeted: `src/main/codex/codex-windows-package-layout.test.ts` (9 passed), plus `runtime-home-service`, `runtime-home-service-per-account-migration` and `codex-home-paths` for regression. One failure in `runtime-home-service-per-account-migration.test.ts` ("keeps E in-place auth when selection becomes real-home without an explicit sync") reproduces identically on a clean tree with this PR's `runtime-home-service.ts` change stashed, so it is pre-existing and unrelated.
- [ ] `pnpm build` — not run; this is main-process TypeScript with no packaging or asset changes, and the host is low-spec.

**Manual reproduction and verification (Windows 11 Pro 26200, Codex 0.145.0):**

1. Built a flattened install — copied only `bin/codex.exe` out of a release — plus a `CODEX_HOME` with `[windows] sandbox = "elevated"`.
2. Ran `codex sandbox -P ws -- cmd /c echo ...` against it. The process blocked with a top-level window titled `codex-windows-sandbox-setup.exe` (the `ShellExecuteW` not-found dialog), and `.sandbox/sandbox.<date>.log` logged `sandbox setup required: sandbox setup marker missing or incompatible` with no completion — the reported bug.
3. Restored `codex-resources/` and `codex-path/` as siblings of `bin/`, re-ran the same command: no dialog. The run reached an actual UAC consent prompt instead, proving it got past `ERROR_FILE_NOT_FOUND` into real elevation.
4. Re-emptied the flattened install and ran the shipped `repairCodexWindowsPackageLayout()` against it. It selected the byte-identical npm vendor donor by SHA-256 and restored `codex-resources`, `codex-path` and `codex-package.json` — the exact state proven healthy in step 3.

## AI Review Report

Self-review focused on the risks this change introduces: writing into a third-party install directory, donor selection correctness, cross-platform safety, and launch-path cost.

- **Cross-platform (macOS / Linux / Windows):** verified the `process.platform !== 'win32'` guard is the first statement, so nothing else in the module — including `resolveCodexCommand()` — runs off Windows; a unit test asserts the `darwin` no-op leaves the filesystem untouched. All paths go through `node:path` `join`/`dirname`/`basename`; no separators or shell invocations are hardcoded. No keyboard shortcuts, labels, or Electron platform APIs are touched. WSL and SSH launch paths are unaffected — the repair is placed after the WSL early-return, and it only inspects the locally resolved entrypoint.
- **Wrong-donor risk:** flagged that copying a helper from a mismatched Codex build could be worse than the popup. Addressed by requiring the donor's `bin/codex.exe` to be byte-identical (size prefilter, then SHA-256); a unit test asserts a different-build release is refused and nothing is copied.
- **Destructive-write risk:** flagged that Orca writes into a directory it does not own. Addressed by only ever creating absent directories — no overwrite, no delete — and by publishing each directory via staged copy + `rename`, so a concurrent Codex launch never observes a partially-copied `codex-resources/`. A losing race falls back to accepting the directory the other process published.
- **Path traversal from the manifest:** flagged that `resourcesDir`/`pathDir` are read from an on-disk JSON file. `readRelativeDirName` rejects anything containing a separator, plus `.` and `..`, so the copy target is always a single segment under the package root.
- **Launch-path cost:** hashing is gated behind a size match against a candidate that actually has the missing directories, and the whole donor search only runs when the layout is already broken; a healthy install costs two `stat` calls per launch. Hashing streams in 1 MiB chunks so a ~100 MB entrypoint is never buffered whole.
- **Failure containment:** every filesystem read is wrapped so a malformed manifest, an unreadable release cache, or a permission error degrades to "no repair" and a warning instead of blocking a Codex launch.

## Security Audit

- **Command execution:** none. The module performs filesystem operations only — no `spawn`, `exec`, or shell strings — and never runs the helper it restores.
- **Path handling:** the write target is derived from the Orca-resolved Codex entrypoint (`dirname(dirname(exe))`), constrained to entrypoints literally under a `bin/` directory, and joined only with manifest-supplied names that are validated as single path segments. Donor roots come from fixed, user-owned locations (`~/.codex/packages/standalone/releases/`, the npm `@openai/codex` vendor tree).
- **Input handling:** `codex-package.json` is parsed defensively — non-object, array, and malformed JSON all fall back to the built-in default directory names.
- **Privilege:** all writes stay in per-user directories; nothing elevates, and no ACLs are modified. The change reduces UAC noise rather than adding any.
- **Integrity:** restored binaries are only ever copied from a release whose entrypoint hashes identically to the one about to run, so this cannot introduce a helper from an untrusted or mismatched source.
- **Secrets / auth / IPC / dependencies:** none touched; no new dependencies, no IPC surface, no credential or network access.

Follow-up: none required for this fix.

## Notes

- Windows-only behavior; no-op on macOS and Linux.
- This does not address #7555, whose popup is `ERROR_MOD_NOT_FOUND` with the helper verifiably present under a console-less nested spawn. That remains open and likely needs a Codex-side fix.
- The underlying packaging gap is Codex-side (its installer flattens the release). This is a defensive repair on Orca's launch path so users stop seeing the dialog; it becomes a no-op once the upstream installer copies all declared directories.
- If a host has a flattened install and no matching release cached anywhere, Orca logs `[codex-package-layout] Codex install is missing sandbox resources and no matching release was found` and leaves it alone — the user still needs to reinstall Codex.

### DIFF
--- a/src/main/codex-accounts/runtime-home-service.ts
+++ b/src/main/codex-accounts/runtime-home-service.ts
@@ -70,6 +70,7 @@ import { getDefaultWslDistro, getWslHome } from '../wsl'
 import { isCodexSystemDefaultRealHomeEnabled } from '../codex/codex-real-home-flag'
 import { hasCustomCodexHomeOverrideForLaunch } from '../codex/codex-real-home-path'
 import { invalidateCodexSessionBackfillMarker } from '../codex/codex-session-backfill-marker'
+import { repairCodexWindowsPackageLayout } from '../codex/codex-windows-package-layout'
 import { assertOwnedHostCodexManagedHomePath } from './host-codex-managed-home-ownership'
 import {
   codexAuthIsFresher,
@@ -220,6 +221,10 @@ export class CodexRuntimeHomeService {
       this.startWslSessionBridgeForLaunch(wslTarget, runtimeHomePath)
       return runtimeHomePath
     }
+    // Why: an installer that flattens a standalone Codex release to `bin/` leaves
+    // codex-windows-sandbox-setup.exe unreachable, so every sandboxed tool call
+    // pops an OS "Windows cannot find" dialog. No-op off Windows and once healthy.
+    repairCodexWindowsPackageLayout()
     const selfContainedAccount = this.getSelfContainedManagedHostAccount()
     if (selfContainedAccount) {
       const perAccountHome = this.prepareSelfContainedManagedHomeForLaunch(
--- a/src/main/codex/codex-windows-package-layout-filesystem.ts
+++ b/src/main/codex/codex-windows-package-layout-filesystem.ts
@@ -1,0 +1,193 @@
+import { createHash } from 'node:crypto'
+import {
+  closeSync,
+  cpSync,
+  existsSync,
+  openSync,
+  readdirSync,
+  readFileSync,
+  readSync,
+  renameSync,
+  rmdirSync,
+  rmSync,
+  statSync,
+  writeFileSync
+} from 'node:fs'
+import { join } from 'node:path'
+
+const ENTRYPOINT_DIR_NAME = 'bin'
+const MANIFEST_FILE_NAME = 'codex-package.json'
+const HASH_CHUNK_BYTES = 1024 * 1024
+
+export function codexPackageDirectoryHasEntries(directoryPath: string): boolean {
+  try {
+    return statSync(directoryPath).isDirectory() && readdirSync(directoryPath).length > 0
+  } catch {
+    return false
+  }
+}
+
+export function findCodexWindowsPackageDonor({
+  entrypointPath,
+  homePath,
+  appDataPath,
+  packageRootPath,
+  requiredDirNames
+}: {
+  entrypointPath: string
+  homePath: string
+  appDataPath: string | null
+  packageRootPath: string
+  requiredDirNames: string[]
+}): string | null {
+  const entrypointSize = fileSize(entrypointPath)
+  if (entrypointSize === null) {
+    return null
+  }
+  const candidates = listReleaseRootPaths(homePath, appDataPath).filter(
+    (candidate) =>
+      candidate !== packageRootPath &&
+      requiredDirNames.every((dirName) =>
+        codexPackageDirectoryHasEntries(join(candidate, dirName))
+      ) &&
+      fileSize(candidateEntrypointPath(candidate)) === entrypointSize
+  )
+  if (candidates.length === 0) {
+    return null
+  }
+
+  const entrypointHash = hashFile(entrypointPath)
+  if (!entrypointHash) {
+    return null
+  }
+  return (
+    candidates.find(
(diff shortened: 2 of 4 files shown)


brennanb2025:
Closing rather than merging this downstream repair.

Investigation confirmed the exact failure is owned by the official Codex Windows standalone installer/helper-resolution path and is tracked upstream at https://github.com/openai/codex/issues/32655. The release package is complete; the PATH-visible bin junction hides the sibling resource directory, while Codex resolves its sandbox helpers relative to the unresolved visible executable path.

Orca should not silently reconstruct a third-party CLI installation from another local copy on PTY launch. No Orca code from this PR was merged.

Issue #12184, opened 2026-08-03
[Bug]: Korean IME in integrated terminal: syllables intermittently dropped and Space inserts a newline during composition (regression after recent update)


wer2956:
### Environment

* Orca version: 1.4.164 (regression appeared after auto-update around 2026-07-31)
* OS: Windows 11, Korean Microsoft IME
* Setup: remote SSH workspace to a Linux host; running Claude Code CLI (raw-mode TUI) in the integrated terminal, TERM=xterm-256color

### Symptoms

While typing Korean (Hangul) into a TUI app in Orca's integrated terminal:

1. Committed syllables are intermittently dropped, roughly one syllable at a time.
2. Pressing Space during or right after IME composition sometimes inserts a newline instead of a space, breaking the input line mid-word.

Both started immediately after an Orca update around 2026-07-31. Until then the same setup worked fine.

### Not reproducible outside Orca

Same Windows machine, same Korean IME, same SSH host, same CLI: typing in Windows Terminal + ssh works perfectly (no dropped syllables, no spurious newlines). Only Orca's terminal reproduces it.

### Possibly related

* stablyai/orca#6513 (synthesized/injected text dropped while kitty keyboard protocol is active) and the CJK-gated fix in [#6417](<https://github.com/stablyai/orca/issues/6417>) look like the same mechanism.
* The v1.4.163 release note "fix(terminal): reconcile cross-platform IME composition lifecycle" landed in the same window as this regression.

Happy to provide logs or test builds.

wer2956:
Update: rolled back to v1.4.162 and the problem is completely gone — no dropped syllables, no spurious newlines from Space during composition. Same machine, same IME, same remote host.

So the regression window narrows to v1.4.163, which matches its release note "fix(terminal): reconcile cross-platform IME composition lifecycle". Reproduced on v1.4.164, absent on v1.4.162.

ProjectAmuseB:
Also reproducing on **1.4.164** — Windows 11 Pro (build 26200), Microsoft Korean IME (2-set), D2Coding terminal font.

One data point that may help narrow it down: my setup is a **purely local workspace** — no remote/SSH host, PTY spawned on the same Windows machine — and both symptoms reproduce identically (committed syllables trailing behind, Space intermittently inserting a newline mid-word). So the remote transport isn't part of the mechanism; it points squarely at the client-side composition path.

[...]

Issue #12179, opened 2026-08-03
[Bug]:  Typing a Hangul syllable whose batchim (final consonant) requires Shift inserts a stray newline in Claude Code


kepary:
### Environment
- OS: Windows 10 Enterprise
- Orca version: 1.4.164 (also reproduced on 1.4.161)
- Terminal app: Claude Code CLI 2.1.220
- TERM_PROGRAM=Orca
- Keyboard layout: Korean 2-beolsik (2벌식)

### Summary
When typing Korean inside a Claude Code chat input running in an Orca terminal pane on Windows, a stray newline is inserted the moment a Hangul syllable is completed with a **batchim (final consonant) that requires Shift** — e.g. ㅆ, ㄲ, ㄳ-type double-consonant finals, as in 했 (ㅎ+ㅐ+ㅆ) or 갂 (ㄱ+ㅏ+ㄲ). A single occurrence is enough to trigger it — no need to repeat it.

This is NOT limited to isolated test cases — it happens during totally ordinary Korean sentence typing whenever a word ends in one of these batchim, which fragments normal messages into multiple broken lines. (This bug has been silently corrupting our own bug-report messages typed to Claude Code throughout this investigation — see "Additional notes" below.)

### When it started
Did not happen before a recent Orca update. Reproduces on both 1.4.161 and the current 1.4.164.

### Precise reproduction (narrowed down via testing)
- Typing a Shift+key combo as an **initial consonant** alone (e.g. isolated ㄸ, ㄲ, ㅆ, ㅉ jamo not part of a completed syllable) — no problem.
- Typing uppercase English letters with Shift (including several in a row) — no problem.
- Typing a **complete Hangul syllable whose final consonant (batchim) requires Shift** (e.g. 했, 갂) — a stray newline is inserted, even for a single occurrence. This appears tied specifically to the IME's syllable-finalization step for a Shift-modified batchim.

### Steps to reproduce
1. Open a terminal pane inside Orca on Windows, switch to Korean input (2-beolsik layout)
2. Run `claude` to start a Claude Code session
3. In the chat input, type any word ending in a Shift-requiring batchim, e.g. "했다" or "갂"
4. Observe: a newline is inserted right after the syllable is completed, splitting the sentence mid-typing

### Isolation / evidence this is Orca-specific
[...]

Issue #12171, opened 2026-08-03
[Bug]: In `1.4.164` version, Korean typing is reflected very late in Terminal


hi-one-gg:
### Operating system

Windows 11

### Orca version

1.4.164

### Details

I just updated to version `1.4.164`. I'm on Windows 11.

As others have reported, when I type Korean characters with the shift key, the line breaks are entered twice.
Also, when I type Hangul, the first and second letters are not entered into the terminal, but when I type the third letter, the first letter is visible, and when I type the fourth letter, the second letter is visible. It's very inconvenient.

nwparker:
Thanks for reporting both symptoms. The Windows IME newline paths were fixed by #12259 and #12265, while the delayed Hangul syllable delivery was fixed by #12278. These fixes are available in the latest release, [v1.4.167](https://github.com/stablyai/orca/releases/tag/v1.4.167). Please update and let us know if Korean input still lags or inserts unexpected line breaks. We will continue hardening IME support across platforms and input methods.

Issue #12164, opened 2026-08-03
[BUG] Korean IME Syllable Duplication in Terminal/Viewer Rendering (e.g. '프프로로젝젝트트')


pobbye-star:
# [BUG] Korean IME Syllable Duplication in Terminal/Viewer Rendering (e.g. '프프로로젝젝트트')

## 1. Issue Overview
- **Target Repository**: `stablyai/orca`
- **Issue Type**: Bug Report (Terminal & UI Text Rendering / Korean IME Preedit Buffer Sync)
- **Environment**: Windows 11 / Orca Application (Electron + WebGL Terminal Engine)

---

## 2. Problem Description
When LLM agents output Korean text in the Orca terminal/viewer or when a user selects/copies text in the Orca interface, Korean syllables get duplicated character-by-character (double consonant/vowel commit).

### Reproduction Example:
- **Expected Text**: `프로젝트 브랜딩 이름 확정`
- **Actual Rendered/Copied Text**: `프프로로젝젝트트 브브랜랜딩딩 이이름름 확확정정`

---

## 3. Root Cause Analysis
- During terminal canvas repaints and UI text selection, the Windows CJK IME preedit buffer gets committed twice to the WebGL/DOM terminal view buffer.
- This creates a desynchronization between the PTY output stream and the rendered terminal view buffer for CJK multi-byte syllabic blocks.

---

## 4. Proposed Fix Strategy for Stably AI Team
1. **IME Composition Lifecycle Guard**: Wrap terminal composition events (`compositionstart`, `compositionupdate`, `compositionend`) with a deduplication state guard.
2. **Text Buffer Sync Check**: Ensure that preedit text is not flushed into the render buffer twice when handling window repaints or selection copies.

---

## 5. GitHub Submission Draft (Ready to Post)

```markdown
### Bug Report: Korean IME Syllable Duplication in Terminal & Canvas View

**Describe the bug**
When Korean text is rendered in the terminal output or copied from the viewer, every Hangul syllable gets duplicated (e.g., `프로젝트` turns into `프프로로젝젝트트`).

**To Reproduce**
1. Open Orca App on Windows 11.
2. Run any CLI agent (Codex, AGY, Claude, etc.) that outputs Korean text into the Orca terminal.
3. Observe the output or select and copy the text.
4. The output text is rendered with duplicated characters (e.g. `프프로로젝젝트트 브브랜랜딩딩 이이름름`).

**Expected behavior**
[...]

PR #12187, opened 2026-08-03
fix(terminal): stop Shift+jamo from injecting a newline while an IME composes

### LINKED ISSUE #11878: [Bug]: Korean double consonants (쌍자음) trigger automatic line breaks in the Windows integrated terminal

ShreyanshVaibhaw:
## Summary

Fixes #11878.

Typing Korean double consonants (ㄸ ㄲ ㅆ ㅉ ㅃ) in a Windows terminal pane injected a stray newline mid-word, as if Enter had been pressed partway through typing.

`isTerminalImeProcessEnter` identified a composed Shift+Enter as "key is `Process`, keyCode is 229, exactly one of Shift or Ctrl held":

```ts
event.key === 'Process' && event.keyCode === 229 && getTerminalImeModifiedEnterKind(event) !== null
```

While a Windows IME composes, that description fits **every** key, not just Enter. On the 2-beolsik layout each double consonant is Shift + a base key, so it matched. `keyboard-handlers.ts` then synthesizes `key: 'Enter'` for anything that matches and hands it to the shortcut resolver:

```ts
const imeProcessEnter = isWindows && hasPendingImeComposition && isTerminalImeProcessEnter(e)
const shortcutEvent = imeProcessEnter ? { key: 'Enter', code: e.code, ... } : e
```

So typing ㄸ fired the Shift+Enter newline shortcut. This matches the isolation in the linked reports exactly: plain Shift+letter is fine, because English input is not IME-composed and reports its real key.

The fix requires the physical key to actually be Enter. `code` is unaffected by composition, so it separates a genuine Shift+Enter from a composed jamo without touching the Shift+Enter path itself. `code` was already being read one line later, it just was not part of the test.

### Likely duplicates

Several recent reports appear to share this cause. Worth a triage pass rather than my assertion:

#12179, #12156, #12152, #12151, #12184

I did **not** include #12164 (IME syllable duplication in rendering) or #12171 (macOS Korean input latency). Those look like different code paths and I have not verified them.

## Screenshots

No visual change.

## Testing

- [x] `pnpm typecheck`
- [x] Added regression tests

- `src/renderer/src/components/terminal-pane/`: **2892 passed, 218 files, 0 failed**
- New cases cover all five 2-beolsik double consonants by physical `code`, plus NumpadEnter to keep the real Shift+Enter path covered.
- **Confirmed to fail against the previous implementation**: reverting the source fails exactly the 5 new jamo cases and nothing else.

### Verification caveat, stated plainly

I do not have a Korean IME on Windows and **could not reproduce the original symptom first-hand**. This fix was derived from reading the input path and is verified by tests and by the reporters' own isolation (single Shift+key fine, consecutive ones break; English Shift unaffected), not by observing the bug. Confirmation from any of the reporters on a real 2-beolsik setup would be worth having before merge.

## AI Review Report

- **Cross-platform.** The only caller is already gated behind `isWindows && hasPendingImeComposition`. macOS and Linux never reach it. No platform branch added or removed.
- **Does not narrow the real Shift+Enter path.** The synthesized-Enter branch still triggers for `code === 'Enter'` and now also explicitly for `NumpadEnter`, which the previous predicate accepted only incidentally. A test pins that.
- **Other IMEs.** The fix keys on the physical key rather than on anything Korean-specific, so Japanese and Chinese IMEs that also report `Process`/229 get the same correction. Nothing in the change is layout-specific.
- **Flagged and addressed.** The review noted the shared test factory omitted `code`, which would have made every existing case pass vacuously against the new condition. The factory now defaults `code: 'Enter'`, so the existing assertions still mean what they claim.

## Security Audit

- **Input handling.** This strictly narrows which key events are rewritten into a synthetic Enter. No input is newly trusted or forwarded.
- **Command execution, path handling, auth, secrets, IPC, dependencies.** Untouched. No dependency changes.
- No user-controlled string reaches a shell, filesystem, or network path from this code.

## Notes

- **Pre-existing lint failure.** `pnpm lint` fails at `verify:skill-bundle-manifest` with stale `resources/skills/*.json`. Reproduces on a clean `main`, so it is not introduced here, and those are generated artifacts I did not want to churn.
- **`pnpm test` and `pnpm build` not completed locally.** A clean `main` on this Windows machine fails 167 tests across 57 files, so a green local suite is not achievable either way. The affected area is fully green.
- **Scope held deliberately narrow.** #12179 also mentions a `--resume` argument arriving as literal typed text after a restart. That is a different mechanism (session replay, not key handling) and is not addressed here.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.test.ts
@@ -325,6 +325,7 @@ describe('isTerminalImeProcessEnter', () => {
   const event = (overrides: Partial<KeyboardEvent> = {}) =>
     ({
       key: 'Process',
+      code: 'Enter',
       keyCode: 229,
       metaKey: false,
       ctrlKey: false,
@@ -340,6 +341,10 @@ describe('isTerminalImeProcessEnter', () => {
     }
   )
 
+  it('recognizes the numpad Enter while composing', () => {
+    expect(isTerminalImeProcessEnter(event({ code: 'NumpadEnter' }))).toBe(true)
+  })
+
   it.each([
     { key: 'Enter' },
     { keyCode: 13 },
@@ -348,6 +353,19 @@ describe('isTerminalImeProcessEnter', () => {
     { altKey: true }
   ])('rejects a non-IME or ambiguous Process key', (override) => {
     expect(isTerminalImeProcessEnter(event(override))).toBe(false)
+  })
+
+  // Regression: while an IME composes, Windows reports key 'Process' + keyCode 229 for
+  // every key. Shift+<jamo> on 2-beolsik produces the double consonants ㄸ/ㄲ/ㅆ/ㅉ and
+  // used to be rewritten into Shift+Enter, injecting a newline mid-word.
+  it.each([
+    { code: 'KeyE', jamo: 'ㄸ' },
+    { code: 'KeyR', jamo: 'ㄲ' },
+    { code: 'KeyT', jamo: 'ㅆ' },
+    { code: 'KeyW', jamo: 'ㅉ' },
+    { code: 'KeyQ', jamo: 'ㅃ' }
+  ])('rejects Shift+$jamo composed on 2-beolsik ($code)', ({ code }) => {
+    expect(isTerminalImeProcessEnter(event({ code }))).toBe(false)
   })
 })
 
--- a/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-ime-deferred-newline.ts
@@ -107,11 +107,19 @@ export function getTerminalImeModifiedEnterKind(
 }
 
 export function isTerminalImeProcessEnter(
-  event: Pick<KeyboardEvent, 'key' | 'keyCode' | 'metaKey' | 'ctrlKey' | 'altKey' | 'shiftKey'>
+  event: Pick<
+    KeyboardEvent,
+    'key' | 'code' | 'keyCode' | 'metaKey' | 'ctrlKey' | 'altKey' | 'shiftKey'
+  >
 ): boolean {
   return (
     event.key === 'Process' &&
     event.keyCode === 229 &&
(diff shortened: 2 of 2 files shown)


USERNAME-737:
### Confirmed on a real Windows + 2-beolsik setup

You asked for confirmation from a reporter on a real Korean IME before merge — here it is.

**Setup**
- Windows 10 Enterprise 19045
- Microsoft Korean IME, 2-beolsik (두벌식)
- Built and ran this branch (`fix/ime-shift-composition-false-enter`) as a dev instance via `pnpm dev`, alongside the installed production 1.4.164 that still reproduces the bug.

**Result: the spurious newline is gone.** Typing Shift-requiring jamo (쌍자음 and ㅒ/ㅖ — e.g. `있다`, `했다`, `얘기`, `오빠`) in a terminal pane no longer injects a line break mid-word. Same keystrokes on production 1.4.164 reproduce it every time.

I'm the reporter of #12156, which you listed as a likely duplicate — that's correct, same root cause, and the newline half of it is fixed here.

**One note on #12156 specifically:** that issue also reports a second symptom — triggering the bug *twice in a row* additionally copies the in-progress terminal input line to the **OS clipboard** (verified by pasting into Notepad, with "copy on select" and OSC 52 both off). I have not yet re-tested that specific two-in-a-row clipboard case against this branch, so I can't confirm whether this fix covers it too. Given the newline no longer fires at all, it plausibly does — but worth someone verifying rather than assuming, since the clipboard write path was never fully traced in that issue.

**Unrelated build note**, in case it helps other Windows contributors: `pnpm install` / `pnpm dev` hard-fails on a machine without Visual Studio Build Tools because `windows-native-registry` can't be rebuilt (`ensure-native-runtime.mjs` gates startup on it). It's only used by `src/main/cli/windows-user-path-registry.ts`, which already handles failure via try/catch. Adding it to the existing `ignoreModules` list let the dev build run fine. Not related to this PR — just a hurdle for anyone trying to verify Windows IME fixes locally.

Thanks for the quick turnaround on this one.

nwparker:
Closing as superseded by #12265, which landed the Shift+jamo/non-Enter Process-key fix with the broader Windows IME ownership and press-evidence coverage. Thank you for the focused diagnosis and tests.

PR #12192, opened 2026-08-03
fix(localization): stop the repair policy rewriting translated generic terms to English

### LINKED ISSUE #12113: repair-locale-catalog rewrites ~2000 translated values back to English across ko/ja/zh/es
### Summary

Catalog values do not survive `repair-locale-catalog`. Any value that translates a term listed in `NEVER_TRANSLATE_VALUES` is rewritten back to the English term the next time the script runs, so the committed wording is temporary. `verify:localization-catalog` does not flag it, and the app never runs the policy, so the drift is invisible until someone regenerates the catalog.

This is not limited to a couple of recent PRs — a full sweep of `main` shows it affects roughly 2% of the catalog.

### Scale

Feeding every catalog value back through the real policy path on a clean checkout of `main`:

| locale | values rewritten | total string values |
|---|---|---|
| ko | 279 | 11775 |
| ja | 461 | 11775 |
| zh | 1180 | 11775 |
| es | 479 | 11775 |

Classifying the 279 `ko` cases by what the rewrite does:

| kind | count |
|---|---|
| localized term → English (`터미널` → `terminal`, `터미널 1` → `Terminal 1`) | 228 |
| English → localized (policy filling in a translation — desirable) | 13 |
| spacing / phrasing normalization | 38 |

So the dominant case is the one that loses translator work. The other locales were not classified, but `zh` at 1180 suggests the same shape or worse.

### Reproduction

```js
// from the repo root, on a clean checkout of main
import { repairTranslatedValue } from './config/scripts/locale-translation-policy.mjs'

const flat = (o, p = '', out = {}) => {
  for (const k in o) {
    const v = o[k], np = p ? `${p}.${k}` : k
    typeof v === 'object' && v ? flat(v, np, out) : (out[np] = v)
  }
  return out
}

const en = flat(require('./src/renderer/src/i18n/locales/en.json'))
const ko = flat(require('./src/renderer/src/i18n/locales/ko.json'))

for (const key of Object.keys(en)) {
  const localeValue = ko[key]
  if (typeof localeValue !== 'string') continue
  const out = repairTranslatedValue({ key, enValue: en[key], localeValue, locale: 'ko' })
  if (out !== localeValue) console.log(`${localeValue}  →  ${out}`)
}
```

Two recent examples that landed while this was unnoticed:

- **#11573** (Editor Font Family, merged 2026-08-01) — 3 of 5 keys drift in `ko`, `ja`, `zh`; `es` is clean. `터미널 글꼴과 동일` → `terminal 글꼴과 동일`.
- **#11205** (Agent Dashboard, merged 2026-07-28) — `에이전트 대시보드 전환` → `Agent 대시보드 전환`, `왼쪽 사이드바에 에이전트 대시보드 항목을…` → `왼쪽 사이드바에 agents 대시보드 항목을…`.

### Cause

`terminal`, `terminals`, `agent`, `agents`, `commit`, `repo` and similar entries live in `NEVER_TRANSLATE_VALUES` in `config/scripts/locale-translation-policy.mjs`. The term-level fix inside `repairTranslatedValue` rewrites the localized term back to English wherever it appears inside a sentence, not only when the whole value is that term.

Adding the value to `locale-<lang>-key-overrides.json` does not protect it: the override is read first and then passed through the same term fix.

### Why it is easy to miss

- `verify:localization-catalog` checks key parity, not round-trip stability, so it passes on `main` today.
- Nothing under `src/` imports `locale-translation-policy.mjs` — only `config/scripts/*` does. Users see the committed wording until someone runs `repair-locale-catalog`, at which point ~2000 values across the four locales flip in one commit.

### Question

Which direction should contributors follow?

1. Keep the policy and write these terms in English inside translated sentences (`terminal 글꼴과 동일`), or
2. Narrow `NEVER_TRANSLATE_VALUES` so it only matches whole values, or drop the common nouns from it, and let locales translate them.

The catalog is currently split both ways — `ko.json` holds 288 values containing `terminal` against 89 containing `터미널`, and 317 containing `agent` against 109 containing `에이전트` — so translators are landing on both conventions and neither is being enforced.

Whichever direction is chosen, a round-trip assertion in `verify:localization-catalog` would keep it from drifting again. Happy to open a PR once the direction is decided.

AnddyAgudelo:
## Summary

Fixes #12113.

`repair-locale-catalog` rewrote translated values back to English on every run. Any value that translated a term listed in `NEVER_TRANSLATE_VALUES` was reverted to the English term, so committed translator wording was temporary — and because `verify:localization-catalog` didn't flag it and the app never runs the policy, the drift was invisible until someone regenerated the catalog.

Measured on a clean checkout (11775 string values per locale): **ko 279, ja 461, zh 1180, es 479** values rewritten. Classifying the ko cases showed the shape of the problem — 228 were *localized term → English* (translator work destroyed), 13 were *English → localized* (the policy correctly filling in a translation, desirable), and 38 were spacing/phrasing normalization. Only the first class is the defect, so the fix is scoped to it: the other two still happen.

### Root cause

`shouldPreserveEnglishValue` tested `NEVER_TRANSLATE_VALUES.has(enValue)` — keyed on the **English** value. Any key whose English string happened to equal a listed token forced the localized value back to English, with no way to distinguish a genuine brand from an ordinary word that every locale should translate. `agent`, `terminal`, `commit`, `repo`, and `Continue` were all on that list.

### What changed

- **17 generic entries moved out of `NEVER_TRANSLATE_VALUES`** into a new `config/scripts/locale-generic-ui-terms.mjs`: `Agent/Agents/agent/agents`, `Commit/Commits/commit/commits`, `Repo/Repos/repo/repos`, `Terminal/Terminals/terminal/terminals`, `Continue`. Criterion: an ordinary noun or verb whose standard ko/ja/zh/es rendering is a translation, and whose product homonym is already pinned by `ENGLISH_ONLY_KEY_PREFIXES`. For `Continue` the catalog was checked directly — the product sense exists only at `auto.lib.agent.catalog.9e2a9bb87b`; the other four occurrences are the English verb.
- **Deliberately kept:** `Markdown/markdown` (format name; it caused zero destructive rewrites — its only hit was a casing normalization), `Token`, `idle`, `ui/ai/ci/ide`, `lint`, `MD`, and every brand/path/code token. `MD` → 医学博士 ("Doctor of Medicine") and `HEAD` → CABEZA are why that list still earns its keep.
- **`BRAND_MISTRANSLATIONS` entries were not hand-pruned.** Instead the revert skips a form that is the term's canonical rendering, so the judgement lives in one place rather than two. The genuinely nonsensical forms provably still fire: zh 回购 (repurchase agreement), ja/zh 端子 (electrical connector), es Comprometerse, zh 降价 / ko 가격 인하 (price reduction).
- **`locale-prose-term-exemptions.mjs` is removed** — the new mechanism subsumes it, so it would have been dead code.
- **`verify-localization-catalog` now fails** when the policy would replace a canonical rendering with its English term: **0 across all four catalogs, from 228+ per locale before.**

### Two further bugs the new gate surfaced

1. **端子 was being reverted inside zh 终端子进程** ("terminal sub-process"), cutting a valid word — pre-existing on `main`, which produces `Terminal子进程`. Reverts now skip matches that overlap a canonical form.
2. **Seven override entries hard-coded English into translated sentences** (zh `Split Terminal Down`/`Right`, the `Local project, Git repo…` value in all four locales, ko `SshTargetForm.137e88ce8d`). Each is set to the wording already committed in the catalog, so no catalog value moves.

### What this does not do

**No catalog values are changed.** The defect is in the policy, and a catalog-wide regeneration would be unreviewable and would bury the actual fix. Every file in this PR is under `config/scripts/`.

A residual class remains that this does **not** gate: 88 ko / 254 ja / 687 zh / 23 es values that the policy would still rewrite. Those are the stale-catalog class `repair-locale-catalog` exists to repair — gating on them would force exactly the catalog-wide rewrite described above. A separate residual of localized-term-to-English rows (5 ko / 6 ja / 39 zh) is unrelated: theme names (`Blue`, `Zinc`), `reviewed-by`, `/path/to/destination`, and the repo's `Jira Issue` convention.

## Screenshots

**No visual change** — these are build-time localization scripts. No runtime UI is touched.

## Testing

- [x] `pnpm lint` (all three localization gates green)
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

`config/scripts/locale-generic-ui-terms.test.mjs` (8 tests). Fail-before/pass-after verified by running it both ways: copied onto pristine `main` it produces **5 failures, exit 1** — including 터미널 글꼴과 동일 → `terminal 글꼴과 동일`, the example from #11573. On this branch, 8 pass.

**Nine existing assertions encoded the destructive contract** (keeping workflow terms in English, keeping `Terminal` as a product surface term). They were found by temporarily converting the suite to `expect.soft` — Vitest short-circuits each block at the first failure, so a naive run hid **13 further failing assertions**. Each is updated with a `Why: #12113` comment and its block renamed; the contract changed, and every genuine-brand assertion in those same blocks still passes untouched.

`npx vitest run config/scripts/` → 643 passed, exit 0.

`pnpm test` exits 1 with 7 failures, all pre-existing and none reachable from anything this PR changes (nothing under `src/` imports a modified module — verified by grep). Running the six implicated files on pristine `main` reproduces 5 of them; the other 2 aggregate large batches and fuzz ~800 near-cap payloads, take 32 s and 47 s under full-suite load, and pass in isolation — load-sensitive timeouts on this machine.

## AI Review Report

- **Scope discipline was the main risk and was checked explicitly.** The naive fix — stop the policy rewriting values — would have killed the 13 desirable English→localized repairs and the 38 normalizations per locale along with the 228 destructive ones. Each class was measured separately and only the destructive one is suppressed.
- **Cross-platform:** Node build scripts; no path assumptions beyond existing helpers, no runtime code, no keyboard or shortcut handling.
- **SSH/remote, folder workspaces, git providers:** not applicable — these scripts never execute at runtime and touch no workspace, transport, or provider logic.
- **Performance:** the added `overlapsCanonicalRendering` check runs per revert candidate during catalog repair, an offline script. No runtime path.
- **Correctness risk reviewed:** the guard could in principle suppress a legitimate revert if a nonsense rendering were also some term's canonical form. That is why `BRAND_MISTRANSLATIONS` was left intact and the nonsense forms were verified to still fire, rather than pruning the list by hand.
- **Ownership:** `.github/CODEOWNERS` routes `config/scripts/*locale*.mjs` and `*localization*.mjs` for focused review. Every term added or removed from a policy list is justified above, with the catalog evidence for the one ambiguous case (`Continue`).

## Security Audit

No runtime code, IPC, command execution, network calls, auth, secrets, or new dependencies. These scripts run offline against repo-local JSON. Path handling is unchanged. The only behavioural change is which string replacements a build-time policy performs. No follow-up needed.

## Notes

- `ko SshTargetForm.137e88ce8d`'s override text is a different sentence from its English source — a mis-keyed override. Only the English-term injection it caused is fixed here; the mis-keying is left alone as a separate concern.
- Override entries that spell generic terms in Latin where the committed catalog already agrees cause no drift and were left as-is; changing them is a convention call, not this defect.


nwparker:

PR #12194, opened 2026-08-03
fix(browser): bound retained webview guests across worktree switches

brennanb2025:
## Summary

After #12137 (STA-3228), every worktree visited with browser pages kept its Electron webview guests alive for the rest of the session, so guest-process memory grew linearly with worktrees visited. This adds a retention budget: hidden worktrees keep their live browser guests for instant revisits, but only the 4 most recently activated (`BROWSER_GUEST_HIDDEN_WORKTREE_RETENTION_LIMIT`). Beyond that, the least-recently-activated worktree's guests are fully destroyed through the sanctioned cleanup path (`destroyPersistentWebview`, via the new `destroyWorktreeBrowserGuests`) and the worktree is dropped from `mountedWorktreeIdsRef`, so its overlay surface unmounts cleanly and a revisit rebuilds from persisted tab state — a fresh mount and page load, not a blank pane.

Key properties:

- LRU order = worktree activation order; re-activating a worktree refreshes its recency.
- Only worktrees that actually hold live guests count toward the budget.
- Never evicted: the active worktree; worktrees whose browser is currently visible to a remote controller (automation-visible / mobile-driven); background measure windows and targeted background mounts; Activity-portal worktrees; and worktrees with eviction-exempt terminal ptys or pending spawn work (a surface unmount would orphan live shells). These stay retained even over budget.
- Eviction destroys-then-recreates; it never detaches a live guest (the STA-3228 blank-forever state).
- Explicit close, worktree sleep, and worktree removal cleanup paths are untouched.

## Screenshots

No visual change. The only observable behavior difference is that revisiting an evicted worktree reloads its browser pages once instead of resuming instantly. This change was authored by an agent worker that did not run the packaged app; validation was via the unit suites and static gates below.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — full suite left to CI (local host runs Node 26 vs the repo's pinned Node 24, a known source of false reds). Targeted + adjacent suites pass locally: the whole `browser-pane` directory (35 files / 230 tests) including the new `browser-guest-worktree-retention.test.ts` and the extended `BrowserPaneOverlayLayer.test.tsx`, `browser-page-viewport.test.ts`, `browser-webview-cleanup.test.ts`, plus `terminal-hidden-worktree-retention.test.ts`.
- [ ] `pnpm build` — not run locally; renderer-only change, CI covers packaging.

New/extended coverage: over-budget eviction targets the LRU worktrees only; active worktree never evicted (even when it ranks past the budget); under-budget is a no-op; non-evictable worktrees stay retained over budget; only live-guest holders count; revisit rebuilds a fresh viewport instead of reattaching the stale shell; eviction unmount resets the `hasCommittedMount` latch; `destroyWorktreeBrowserGuests` destroys every page (including legacy tab-id-keyed guests) and leaves other worktrees alone.

## AI Review Report

Authored and reviewed agentically (Claude). Main risks checked:

- **STA-3228 regression (detached-but-alive guest):** the eviction effect destroys guests via `destroyPersistentWebview` first, then removes the worktree from `mountedWorktreeIdsRef` and bumps a revision so the next render unmounts an already-empty surface. No live webview is ever detached. The revisit path was traced through `ensureBrowserPageWebview` (`created: true` → initial navigation from the store-persisted URL).
- **Terminal interplay:** dropping a worktree from `mountedWorktreeIdsRef` also unmounts its terminal surface, so eviction is vetoed for worktrees with Activity portals, targeted background mounts, active measure windows, pending spawn work, or eviction-exempt ptys (a remount would respawn and orphan those live shells). Ordinary hidden worktrees are already cold-parked/watcher-covered; eviction returns them to the fully supported "not yet visited this session" state.
- **Loop/thrash safety:** the effect runs on activation/surface-set/portal changes; the revision bump is not in its deps, and re-runs treat evicted worktrees as unretained, so it is idempotent (including StrictMode double-invoke).
- **Per-worktree state reset on eviction:** `webviewRegistry`, `registeredWebContentsIds`, page viewports, live-URL cache, and explicit zoom clear via `destroyPersistentWebview`; slot viewport roots clear via the slot's unmount ref callback; the `hasCommittedMount` latch dies with the unmounted layer; `backgroundMountTabIdsByWorktreeRef` / `activationDeferredMountTabIdsByWorktreeRef` entries are deleted alongside the mounted-set entry (mirroring the existing removed-worktree pruning).
- **Cross-platform (macOS/Linux/Windows):** no shortcuts, labels, file paths, or shell behavior touched; the logic is id-based and identical on all platforms, and applies unchanged to SSH-hosted workspaces and folder workspaces (webview guests are always local Electron processes regardless of workspace host).

Flagged during review and fixed: `recency.includes` inside a filter (React Doctor performance warning on the new lines) → Set lookup. Both CI changed-lines gates (`check-changed-code-quality.mjs`, `check-react-doctor-changed.mjs`) were run locally against the merge base and pass with 0 blocking findings.

## Security Audit

No new input handling, command execution, path handling, auth, secrets, or dependency surface. The only IPC involved is the existing sanctioned `window.api.browser.unregisterGuest` call inside `destroyPersistentWebview`, now reachable from the eviction path with the same page-id arguments explicit close already sends — no new channels or payload shapes. Eviction strictly reduces lingering guest processes and clears their registry entries, so no destroyed guest remains reachable through stale renderer state.

## Notes

- **Cherry-pick target: v1.4.165 release** — this bounds a memory regression introduced by #12137, which shipped in the RC (relevant given the chronic renderer heap growth reports on 1.4.162–164).
- The budget is defined in exactly one place, with its rationale: `BROWSER_GUEST_HIDDEN_WORKTREE_RETENTION_LIMIT = 4` in `src/renderer/src/components/browser-pane/browser-guest-worktree-retention.ts`.
- Intended behavior change: a worktree evicted from the budget reloads its browser pages on the next visit (URLs and tab structure come from the store) instead of resuming instantly.
- Follow-up candidate: worktrees protected from eviction (e.g. automation-pinned) can hold guests over budget indefinitely; if that shows up in the field, a TTL like the terminal retention budget's could be layered on top of the count bound.

### DIFF
--- a/src/renderer/src/components/Terminal.tsx
+++ b/src/renderer/src/components/Terminal.tsx
@@ -53,8 +53,18 @@ import TerminalPaneOverlayLayer from './terminal-pane/TerminalPaneOverlayLayer'
 import {
   collectBrowserWebviewIds,
   destroyRemovedBrowserWebview,
-  destroyWorkspaceWebviews
+  destroyWorkspaceWebviews,
+  destroyWorktreeBrowserGuests
 } from '../store/slices/browser-webview-cleanup'
+import {
+  browserTabVisibilityPageIds,
+  selectBrowserGuestEvictionWorktreeIds,
+  touchBrowserGuestWorktreeRecency,
+  worktreeHoldsLiveBrowserGuests
+} from './browser-pane/browser-guest-worktree-retention'
+import { hasLiveBrowserGuest } from './browser-pane/webview-registry'
+import { isBrowserAutomationVisible } from './browser-pane/browser-automation-visibility'
+import { isBrowserPageMobileDriven } from '@/lib/pane-manager/browser-mobile-driver-state'
 import {
   handleSwitchRecentTab,
   handleSwitchTab,
@@ -273,6 +283,8 @@ function getKeybindingContext(target: EventTarget | null): KeybindingContext {
 
 function Terminal(): React.JSX.Element | null {
   const mountedWorktreeIdsRef = useRef(new Set<string>())
+  // Why an array: browser-guest eviction needs activation order (LRU), not just membership.
+  const browserGuestWorktreeRecencyRef = useRef<string[]>([])
   const measurableBackgroundWorktreeIdsRef = useRef(new Set<string>())
   const terminalWorktreeHiddenSinceRef = useRef(new Map<string, number>())
   // Why two extra clocks: hiddenSince survives a background-measure window (so
@@ -793,6 +805,8 @@ function Terminal(): React.JSX.Element | null {
   >(() => new Set())
   // Why a ref: eviction captures buffers exactly once per force-park episode, before the unmount render.
   const forceParkedCaptureDoneRef = useRef(new Set<string>())
+  // Why state: browser-guest eviction mutates mountedWorktreeIdsRef in an effect; the bump re-renders so evicted surfaces unmount.
+  const [, setBrowserGuestEvictionRevision] = useState(0)
   // Tab restriction for targeted background mounts (wake/resume); a worktree absent from this map mounts all its tabs.
   const backgroundMountTabIdsByWorktreeRef = useRef(new Map<string, ReadonlySet<string>>())
   // Why: only cold-activation deferral (not targeted mounts, which share the map above) creates watcher coverage for every unmounted tab.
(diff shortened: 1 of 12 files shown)


brennanb2025:
Live Electron QA on this head (dev build of the composed branch with #12191, isolated instance, identity-verified):

- Opened a browser guest (example.com) in one worktree, switched to another worktree, switched back.
- While away: guest stayed in DOM parked at 0×0 (retained, under budget — correct, not destroyed).
- After return: same guest connected and rendering at full size. Zero renderer window errors during the cycle.



Also observed earlier in the session: the guest rebuilds and renders correctly even across a full renderer reload — the revisit-rebuild path works. Review loop caught and folded in two real fixes (eviction without surface unmount; download veto + kill switch); remaining pass-3 polish (LRU order across kill-switch toggle, interrupted-download veto release) fails toward extra retention only and follows in a small separate PR.

Issue #12109, opened 2026-08-02
[Bug]: Mobile "New Tab → Codex" opens a terminal tab, but the Codex session never starts


jinjuchoo:
[Bug]: Mobile "New Tab → Codex" opens a terminal tab, but the Codex session never starts and messages get no response

## Summary
On Orca Mobile, creating a new tab and choosing Codex appears to create a new terminal tab, but the Codex session does not actually attach/start. The tab accepts typed input, but no Codex banner or model response ever appears.

This makes the mobile companion usable for monitoring/replying to existing sessions, but not for spinning up additional parallel Codex lanes from the phone.

## Environment
- Orca Desktop: macOS app (official website download)
- Orca Mobile: beta companion app
- Host OS: macOS
- Agent: Codex
- Remote path used successfully for mobile access: Tailscale-backed LAN pairing
- Date observed: August 2, 2026

## Reproduction
1. Open an existing worktree/session in Orca Desktop.
2. Pair Orca Mobile to the desktop host.
3. On Orca Mobile, open the same worktree.
4. Use `New Tab` and choose `Codex`.
5. In the new tab, send a message such as `안녕`.

## Expected
A fresh Codex session should boot in that new tab (banner / startup output visible), and the message should receive a normal Codex response.

## Actual
- A new tab is created (`Terminal 2` in my case).
- The message is accepted into the terminal input area.
- No Codex banner appears.
- No model response ever arrives.
- The tab behaves like a bare terminal/input surface rather than a live Codex chat/session.

## What I confirmed locally
I inspected the host state/logs while reproducing:
- The original existing Codex tab remained healthy and responsive.
- The new mobile-created tab existed in Orca state as a second terminal tab in the same worktree.
- The new tab content only showed the literal message I typed (`안녕`) and did not show any Codex startup output.
- Even manually sending `codex --dangerously-bypass-approvals-and-sandbox` into that tab via the host-side Orca CLI did not surface a normal Codex startup/response loop in the tab.

This suggests the mobile `New Tab -> Codex` flow is creating a tab surface, but not correctly attaching/launching the agent session behind it.

## Impact
This blocks a key mobile workflow: starting multiple parallel Codex lanes from mobile. Right now mobile works as a companion for existing sessions, but not as a reliable launcher for new Codex sessions.

Issue #7587, opened 2026-07-06
[Bug]: Creating a Claude tab from mobile spawns it on the desktop, then reaps it ~10s later — mobile tab creation broken


taifunk:
### Operating system

Windows

### Orca version

v1.4.124

### Details

**Short summary**
Creating a new Claude tab from Orca Mobile (`+` → Claude) spawns a real Claude session on the paired desktop, which is then fully destroyed about 10 seconds later. The new tab never appears in the mobile client and disappears from the desktop, so creating tabs from mobile does not work.

**What happened?**
A new terminal appears briefly on the desktop, then vanishes ~10s later; the mobile client never shows it. Captured objectively on the desktop by polling `orca-data.json` (`claudeLivePtySessionIds`) and the `terminal-history/` directory across the tap:
- A new Claude session id is registered in `claudeLivePtySessionIds` and a `terminal-history/<sessionId>` folder is created.
- ~10 seconds later the session id is removed from `claudeLivePtySessionIds` **and** the `terminal-history/<sessionId>` folder is **deleted entirely**. (A normal terminal close preserves that folder; deleting it indicates a rollback, not a graceful close.)

**How can we reproduce it?**
1. Pair Orca Mobile (Android) to a desktop Orca instance.
2. On mobile, tap `+` and choose **Claude**.
3. Observe a new terminal appear briefly on the desktop, then disappear after ~10s; the mobile client never shows the new tab.

**Expected**
Creating a tab from mobile should create a persistent tab visible on both the mobile client and the desktop.

**Anything else that might help (root cause from source at HEAD)**
[...]

PR #12197, opened 2026-08-03
fix(mobile): deliver the agent launch command when a create settles over a bare renderer PTY

brennanb2025:
## Summary

Fixes STA-3214 (Linear) / #12109: on Orca Mobile, "New Tab → Codex" could open a terminal tab in which the Codex session never started — the tab stayed a bare shell named "Terminal N", accepted input, and never showed the Codex banner.

Root cause: the mobile create resolves the agent launch command and delegates tab creation to the renderer. When the renderer's startup queue is lost (the #7587 renderer-stall class the codebase already documents), the pane spawns a **plain shell**, and the create still settles **ready** by adopting that live PTY (`ensurePtyBackedMobileSurfaceForRendererTab`). That success is silent and permanent: because the surface is ready, the #7837 activation-time materialize recovery is bypassed too, so the phone stays bound to a bare shell forever.

Fix (runtime-side, `src/main/runtime/orca-runtime.ts`):
- Record the resolved agent launch command on the pending mobile create.
- At every renderer-backed settle point (ready wait, ready fallback, catch-path rescue), check whether the backing PTY ever received a spawn command — both spawn IPC handlers already record per-PTY launch commands via `noteTerminalSpawnCommand`.
- If the locally registered live PTY has no recorded spawn command, the launch provably never ran: type the launch command into the shell (Enter as its own write) and record it so repeated settles cannot double-deliver.

The healthy path is untouched: a pane that spawned with its startup command has a recorded spawn command and is skipped; runtime-owned/materialize spawns already carry the command; remote-runtime PTYs are excluded because they are not locally registered.

## Screenshots

No visual change (main-process fix). QA evidence: with the renderer's startup queue deliberately dropped (simulating the #7587 stall), before the fix the mobile create settled into a permanently bare "Terminal 12"; after the fix the same sequence boots Codex (banner + tab renamed) on a real paired iOS-simulator client.

## Testing

- [x] `pnpm lint` — full `oxlint` clean; pre-commit ran oxlint + react-doctor + oxfmt on staged files
- [x] `pnpm typecheck`
- [x] `pnpm test` — targeted: full `src/main/runtime/orca-runtime.test.ts` (1006 passed | 1 skipped) via `vitest --config config/vitest.config.ts`
- [ ] `pnpm build` — not run; `electron-vite` main/preload/renderer bundles built and booted in the live QA rig

New regression tests (verified to fail without the fix — the delivery write is never made — and pass with it):
- `delivers the agent launch command when a create settles over a bare renderer PTY`
- `does not re-deliver the agent launch command when the adopted renderer PTY spawned with one`

Live QA on an isolated dev desktop (own HOME + ORCA_USER_DATA_PATH) paired to the real mobile app in an iOS simulator, plus a wire-protocol paired client:
- Baseline: New Tab → Codex booted Codex in all environmental variants tried (worktree open/closed on desktop, window visible/hidden/minimized, sequential and concurrent creates, chat-by-default settings, 10+ existing tabs) — the mainline delivery path is unchanged.
- Repro + fix: with the renderer startup queue dropped once, pre-fix the create settled ready over a bare shell (exact ticket symptom); post-fix the identical sequence delivers the Codex launch and the tab boots (verified on the paired phone UI).

## AI Review Report

Reviewed the settle-path change for: double-delivery (guarded by the per-PTY spawn-command record, which is also written after delivery), delivery into the wrong PTY (lookup is scoped to the pending create's worktree + renderer tab id and requires a locally registered, connected PTY with a valid pane key), interaction with the client-disconnect path (pending record is deleted in the same `finally` as before; disconnected creates still throw before the delivery points), and the race between `registerPty` and `noteTerminalSpawnCommand` (both run in one synchronous block in the spawn IPC handlers; delivery runs from promise continuations, which cannot preempt that block).

Cross-platform: main-process-only change; the delivered command comes from the same `buildAgentStartupPlan` output the create already ships to the renderer, so host-shell quoting (macOS/Linux/Windows, WSL, SSH) is unchanged. `\r` submits across supported shells and is written separately so a long command cannot swallow it. SSH spawns record their spawn commands through the same IPC handlers, so SSH panes are discriminated identically; remote-runtime PTYs are excluded (not locally registered). No shortcuts, labels, or path handling touched.

## Security Audit

No new input surfaces. The delivered text is the create's own resolved startup command: either built by `buildAgentStartupPlan` from host-side settings, or the `command` a paired client already supplies today (which the renderer would have executed at spawn) — paired clients can already write to these terminals via `terminal.send`, so no privilege change. No secrets logged or persisted; the spawn-command record reuses the existing `terminalSpawnCommandsByPtyId` map. IPC surface unchanged (no new channels or schema changes).

## Notes

- The renderer-side trigger (losing the queued startup) was reproduced synthetically; the mainline renderer path delivers correctly in every environment variant tested. The runtime-side guarantee added here closes the entire class regardless of which renderer anomaly drops the startup, matching the recovery philosophy of #7837 (which fixed the same symptom for the activation/materialize path only).
- Delivery types into the shell even if the shell is still initializing; the kernel input queue holds it until the prompt is ready — same behavior as a user typing immediately after opening a terminal.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -31204,6 +31204,125 @@ describe('OrcaRuntimeService', () => {
     }
   })
 
+  it('delivers the agent launch command when a create settles over a bare renderer PTY', async () => {
+    vi.useFakeTimers()
+    try {
+      const leafId = '77777777-7777-4777-8777-777777777777'
+      const write = vi.fn((_ptyId: string, _data: string) => true)
+      const runtime = new OrcaRuntimeService({
+        ...store,
+        getSettings: () => ({
+          ...store.getSettings(),
+          disabledTuiAgents: [],
+          agentCmdOverrides: {}
+        })
+      } as never)
+      runtime.setPtyController({
+        spawn: vi.fn(),
+        write,
+        kill: () => true,
+        getForegroundProcess: async () => null
+      })
+      runtime.setNotifier(createMobileCreateTestNotifier(vi.fn()))
+      const webContents = { send: vi.fn() }
+      const send = vi.fn((_channel: string, payload: { requestId: string }) => {
+        // Why: the pane spawned before its startup queue landed (the #7587
+        // renderer-stall class), so no spawn command is recorded for the PTY.
+        runtime.registerPty('pty-bare', TEST_WORKTREE_ID, null, { tabId: 'tab-bare', leafId })
+        ipcMain.emit(
+          'terminal:tabCreateReply',
+          { sender: webContents },
+          { requestId: payload.requestId, tabId: 'tab-bare', title: 'Terminal' }
+        )
+      })
+      webContents.send = send
+      runtime.attachWindow(1)
+      runtime.syncWindowGraph(1, { tabs: [], leaves: [] })
+      electronMocks.BrowserWindow.fromId.mockReturnValue({
+        isDestroyed: () => false,
+        webContents
+      })
+
+      const create = runtime.createMobileSessionTerminal(`id:${TEST_WORKTREE_ID}`, {
+        agent: 'codex',
+        activate: true
+      })
+      await vi.waitFor(() => expect(send).toHaveBeenCalledTimes(1))
+      await vi.advanceTimersByTimeAsync(50)
+      const result = await create
+
+      expect(result.tab).toMatchObject({
+        type: 'terminal',
+        parentTabId: 'tab-bare',
+        leafId,
+        status: 'ready'
+      })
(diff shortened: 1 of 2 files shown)


brennanb2025:
QA evidence (isolated instances only):

- Desktop-side Electron validation deliberately dropped the first renderer startup queue delivery; the settle path re-delivered the startup command and the terminal printed `STA-3214_RECOVERED_STARTUP_COMMAND`.
- iOS mobile validation used the real New Tab → Codex flow with the same one-shot dropped renderer delivery; Codex opened as a chat tab instead of a bare shell.

Desktop Electron recovery

Mobile Codex tab recovery

Issue #9498, opened 2026-07-20
[Bug]: WSL-managed orca CLI fails to launch — .NET duplicate-key ('PATH' vs 'Path') at startup


JunhyeokJang1006:
### Summary

The Orca-managed CLI fails to start when invoked from WSL. Even a basic command such as `orca --version` exits immediately with a .NET duplicate-key error involving the environment-variable names `PATH` and `Path`.

The issue appears to occur during `orca.exe` startup when environment variables inherited through WSL-to-Windows interop are processed case-insensitively.

As a result, the Orca CLI is completely unusable from WSL on this machine.

### Environment

- **Orca**: 1.4.160 (still reproducible)
- **OS**: Windows 11 Pro, Build 22631
- **WSL**: WSL2 2.7.10.0
- **Distribution**: Ubuntu 24.04.3 LTS
- **Kernel**: `6.18.33.2-microsoft-standard-WSL2`
- **CLI launch chain**:

  ```text
  ~/.local/bin/orca
    → ~/.local/bin/orca-ide
    → ~/.local/share/orca/orca-wsl-bridge.ps1
    → resources/bin/orca.exe
  ```

### Error output

```text
Unable to start the Orca CLI:
An item with the same key has already been added.
Key in dictionary: 'PATH'; key being added: 'Path'.
```

### Steps to reproduce

1. Install the Orca-managed CLI in WSL2 on Windows 11.
2. Open an Ubuntu WSL shell.
3. Run:

   ```bash
   orca --version
   ```

4. Observe that the CLI exits before startup and prints the duplicate-key error above.

The same failure occurs with other `orca` subcommands.

### Actual behavior

`orca.exe` fails before normal CLI initialization because `PATH` and `Path` are treated as duplicate keys.

No Orca CLI command can be executed from WSL.

### Expected behavior

The Orca CLI should start normally under WSL.

Environment-variable names inherited through WSL-to-Windows interop should be handled case-insensitively without causing a duplicate-key exception. If both `PATH` and `Path` are present, Orca should safely normalize, merge, or otherwise resolve them before inserting them into a case-insensitive collection.

#### Bridge script

The generated `orca-wsl-bridge.ps1` script does not explicitly read or modify `PATH` or `Path`. It only:

- sets `ORCA_CLI_CWD`;
- changes the working directory with `Push-Location`;
- invokes the Orca launcher using `& $OrcaLauncher @ForwardArgs`.

#### Direct executable invocation

[...]

PR #12218, opened 2026-08-03
Stop the Windows Orca CLI from crashing when the environment carries both PATH and Path

### LINKED ISSUE #12046: [Bug] Windows Orca CLI crashes when environment contains both PATH and Path
<kbd>[Ohio](https://discord.com/channels/1492228674081263786/1533328865949384836/1533328865949384836)</kbd> [Ohio](https://discord.com/channels/1492228674081263786/1533328865949384836/1533328865949384836)  `via Discord`

- OS: Windows
  - Orca version: 1.4.163
  - Orca CLI:
      - C:\Users\AMW\AppData\Local\Programs\orca\resources\bin\orca.cmd
      - C:\Users\AMW\AppData\Local\Programs\orca\resources\bin\orca.exe

  - Context: Codex terminal hosted inside Orca
  - Orca desktop runtime reports ready and reachable

 -Description

  The Orca CLI fails before executing orchestration mutation commands when its process environment contains both PATH and Path.

  Windows environment-variable names are case-insensitive, but the Orca CLI appears to copy the environment into a case-insensitive dictionary using an add operation. This throws
  when it encounters the second casing variant.

  -Exact error

  Unable to start the Orca CLI: Item has already been added. Key in dictionary: 'PATH' Key being added: 'Path'

  -Reproduction steps

  From an Orca-hosted Codex terminal on Windows:

  node -e "console.log(JSON.stringify(Object.keys(process.env).filter(k => k.toLowerCase() === 'path')))"

  Result:

  ["PATH","Path"]

  Confirm that Orca itself is operational:

  orca status --json

  Result includes:

  {
    "runtime": {
      "state": "ready",
      "reachable": true
    }
  }

  Then attempt to create an orchestration Run:

  orca orchestration run-create --objective "Test orchestration run" --json

  Actual result:

  Unable to start the Orca CLI: Item has already been added. Key in dictionary: 'PATH' Key being added: 'Path'

  The failure occurs before a Run is created.

 
  Running:

  cmd.exe /d /c set path

  shows separate PATH= and Path= entries.

  The persistent Windows environment is not duplicated. Registry checks show only the normal Path value at both levels:

  reg.exe query HKCU\Environment /v Path
  reg.exe query "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" /v Path

nwparker:
## Summary

On Windows a process environment block can carry both `PATH` and `Path` — Win32 lookup is case-insensitive, so nothing complains until something copies the block into a case-sensitive data structure.

The packaged `orca.exe` launcher did exactly that. Reading `ProcessStartInfo.EnvironmentVariables` (`native/windows-cli-launcher/OrcaCliLauncher.cs:46` on the base branch) triggers a lazy getter that copies the case-sensitive `Environment.GetEnvironmentVariables()` hashtable into a case-INsensitive `StringDictionaryWithComparer` using `.Add`. The second spelling threw `ArgumentException: Item has already been added. Key in dictionary: 'PATH'  Key being added: 'Path'`, which the catch-all at `OrcaCliLauncher.cs:67` printed as `Unable to start the Orca CLI: ...` before returning exit 1 — before Electron, before the CLI, before any Run existed.

The launcher now mutates its **own** process environment with `Environment.SetEnvironmentVariable` (which goes straight to Win32 `SetEnvironmentVariableW`, no dictionary) and never reads `ProcessStartInfo.EnvironmentVariables` or `ProcessStartInfo.Environment`. With `UseShellExecute = false` and no explicit block, `CreateProcess` receives `lpEnvironment = NULL` and the child inherits the launcher's live block verbatim, duplicates and all. The launcher exits right after `WaitForExit`, so mutating its own env is safe.

Orca was also **producing** the duplicate (line numbers below are on the base branch, pre-fix):

- `src/main/attribution/terminal-attribution.ts:76` read `baseEnv.PATH` (a plain object lookup that misses Windows' `Path`) and `:105` unconditionally wrote `baseEnv.PATH`. Every Windows Orca terminal was therefore launched with both spellings, and which PATH the child resolved came down to which spelling landed first in the block — something Orca did not control. The disabled path (`:134`) had the mirror bug: it deleted `PATH` while leaving stale shim dirs inside `Path`.
- `src/main/runtime/claude-agent-teams-service.ts:33/39` emitted a literal `PATH` key built from `args.baseEnv.PATH`. Its callers pass `{...process.env, ...baseEnv}` (`orca-runtime.ts:24456`, `:26122`), so on the normal Windows block — spelt `Path` — that read was `undefined` and the launch env got a `PATH` containing **only** the tmux shim directory. Note the `:24456` caller is unreachable on Windows: `buildClaudeAgentTeamsLaunchPlan` returns early when `process.platform === 'win32'`, before it calls `createTeamEnv`. The win32-reachable caller is `prepareClaudeAgentTeamsLeaderForHandle` at `:26122`.

`resolvePathEnvKey(env, platform)` is now exported from `src/main/pty/windows-environment-path.ts` and drives the PATH read/write at these sites: `terminal-attribution.ts` (both the apply and the clear path), `ipc/pty.ts` (`readInheritedPath` and the dev-CLI shim write), the `promoteAgentTeamsShimPath` copies in `providers/local-pty-provider.ts:166` and `daemon/pty-subprocess.ts:210`, their requested-env readers (`ipc/pty.ts:3933`, `:5093`), and `claude-agent-teams-service.ts`.

Two PATH-key sites are deliberately **not** converted and keep their pre-existing behavior:

- `ipc/pty.ts` has a *third* `promoteAgentTeamsShimPath` (`:759`) whose own key pick lives at `:770` (`env.PATH !== undefined || env.Path === undefined ? 'PATH' : 'Path'`). It reads and writes the same key, so it cannot mint a duplicate, but on a dual-cased block it still prefers `PATH` by casing rather than by position. Unchanged by this PR.
- `resolveClaudeAgentTeamsShimBin` (`runtime/claude-agent-teams-shim-env.ts`) still reads `env.PATH` when probing disk for `orca-dev`/`orca`. It only picks an executable to name; it writes nothing into the child env.

On Windows `resolvePathEnvKey` resolves **by block order, not by casing**: it returns the first `PATH`/`Path` own key present on the env, falling back to the host block's first. That is what Win32 itself does — `RtlQueryEnvironmentVariable` scans the block linearly and returns the first case-insensitive match — and object key order is block order, since Node enumerates `process.env` in block order and node-pty emits `Object.keys(env)` verbatim without sorting. An earlier revision preferred `Path` unconditionally, which on the reporter's own `["PATH","Path"]` block targeted the spelling the child never resolves.

Because the shared helper also replaced the `Path`-first pick that already lived at `windows-environment-path.ts:249`, `mergeWindowsPathSegments` (behind `mergePersistedWindowsPath`, win32-only) changed too: on a dual-cased block it now follows block order instead of preferring `Path`, and on a path-less env it adopts the host block's spelling instead of hardcoding `Path`. Its existing suite still passes.

Attribution reads and writes that one key and **does not delete** the other spelling. Deleting it would discard a live value: on a `PATH` (live) + `Path` (shadowed) block, removing the live key promotes the shadowed one, so an attribution-only PATH that strips down to empty would leave the child with no PATH-shaped variable at all. Collapsing an inherited duplicate Orca did not create (WSL interop, #9498) is deliberately **out of scope** here.

The one place a duplicate is still collapsed is `daemon/pty-subprocess.ts`, because that duplicate is Orca-minted: the daemon merges `{...process.env, ...opts.env}` (via `mergeGitConfigEnvProtocol`), so a sparse patch spelt differently from the daemon's own block lands *after* it and is shadowed by it. There the requested (main-shaped) value survives and the daemon's own spelling is dropped.

Off Windows `resolvePathEnvKey` returns `PATH` unconditionally, so a legitimate case-sensitive POSIX `Path` variable is never read, written, or deleted.

**Scope.** Closes #12046: the reported failure is `orca.exe` dying with `Item has already been added. Key in dictionary: 'PATH'`, and the launcher no longer materializes the block at all, so it survives any inherited duplicate. The Orca-side changes remove the cause of Orca *minting* the second spelling. What this PR does **not** do is de-duplicate a PATH/Path pair Orca inherited from elsewhere — such a block is passed through to children unchanged, exactly as before.

## ELI5

Windows treats `PATH` and `Path` as the same thing, but a lot of software doesn't. Orca's Windows command-line tool crashed on startup whenever it inherited both spellings, because it tried to file them into a drawer that only had one slot for that name. It now leaves the drawer alone entirely and just hands the whole list to the program it starts.

Orca was also creating the second spelling itself: it read one spelling, then wrote the other, so every Windows terminal Orca opened ended up with two. Now it reads and writes the same one — specifically the one Windows itself would pick, which is whichever comes first in the list, not whichever is capitalised a particular way. If a list already arrived with two, Orca leaves both alone rather than throwing one out, because throwing one out can silently delete the entries the user actually needs. A side effect of the fix: Claude Agent Teams terminals on Windows used to lose almost their entire command search path, and they no longer do.

## Fix proof

Before the fix — every source file this PR touches reverted to `origin/main`, regression tests kept at HEAD:

```
 ❯ config/scripts/build-windows-cli-launcher.test.mjs (3 tests | 1 failed | 1 skipped) 33ms
     × never materializes the child environment block from ProcessStartInfo 4ms
 ❯ src/main/pty/windows-environment-path.test.ts (17 tests | 5 failed) 6ms
     × uses whichever Windows spelling is present on the env 1ms
     × resolves a dual-cased Windows env by block order, not casing 0ms
     × skips a path key explicitly set to undefined 0ms
     × falls back to the host block spelling for a Windows env with no path key 0ms
     × always resolves `PATH` off Windows so a POSIX `Path` variable is untouched 0ms
 ❯ src/main/runtime/claude-agent-teams-service.test.ts (8 tests | 1 failed) 8ms
     × keeps the inherited Windows `Path` instead of minting a truncated `PATH` 3ms
 ❯ src/main/attribution/terminal-attribution-path-env-casing.test.ts (8 tests | 5 failed) 21ms
     × writes back the inherited Windows `Path` spelling instead of adding `PATH` 7ms
     × prepends the shim onto the first-listed `'Path'` and leaves `'PATH'` untouched 2ms
     × keeps a shadowed duplicate when the live spelling strips down to nothing 1ms
     × strips shim entries from the inherited `Path` when attribution is disabled 1ms
     × adopts a host block spelt `'Path'` on a path-less env 2ms
 ❯ src/main/providers/local-pty-provider.test.ts (94 tests | 1 failed) 135ms
       × promotes the agent-teams shim onto the Windows `Path` spelling 5ms
 ❯ src/main/daemon/pty-subprocess.test.ts (111 tests | 1 failed) 723ms
     × collapses its own env merge onto the requested Windows `Path` spelling 4ms

 FAIL  config/scripts/build-windows-cli-launcher.test.mjs > Windows CLI launcher > never materializes the child environment block from ProcessStartInfo
AssertionError: expected 'using System;\nusing System.Collectio…' not to contain 'EnvironmentVariables'
 FAIL  … > writes back the inherited Windows `Path` spelling instead of adding `PATH`
AssertionError: expected [ 'Path', 'PATH' ] to deeply equal [ 'Path' ]
 FAIL  … > prepends the shim onto the first-listed `'Path'` and leaves `'PATH'` untouched
AssertionError: expected [ 'C:\tools\bin', …(1) ] to deeply equal [ …(3) ]
 FAIL  … > keeps a shadowed duplicate when the live spelling strips down to nothing
AssertionError: expected [ 'Path', 'PATH' ] to deeply equal [ 'PATH' ]
 FAIL  … > collapses its own env merge onto the requested Windows `Path` spelling
AssertionError: expected [ 'PATH', 'Path' ] to deeply equal [ 'Path' ]
 FAIL  … > resolves a dual-cased Windows env by block order, not casing
AssertionError: expected 'Path' to be 'PATH' // Object.is equality

 Test Files  6 failed (6)
      Tests  14 failed | 226 passed | 1 skipped (241)
```

Against the **previous revision of this PR** (casing-preference resolver + attribution collapse), the tests added for the review finding still fail — they are not merely main-vs-HEAD deltas:

```
 ❯ src/main/pty/windows-environment-path.test.ts (17 tests | 2 failed) 8ms
     × resolves a dual-cased Windows env by block order, not casing 3ms
     × falls back to the host block spelling for a Windows env with no path key 0ms
 ❯ src/main/attribution/terminal-attribution-path-env-casing.test.ts (8 tests | 3 failed) 22ms
     × prepends the shim onto the first-listed `'PATH'` and leaves `'Path'` untouched 4ms
     × prepends the shim onto the first-listed `'Path'` and leaves `'PATH'` untouched 3ms
     × keeps a shadowed duplicate when the live spelling strips down to nothing 2ms
```

The original preserved repro (`terminal-attribution.bug-12046.test.ts`, since folded into `terminal-attribution-path-env-casing.test.ts`) reproduced the reporter's exact `node -e` observation:

```
 FAIL  src/main/attribution/terminal-attribution.bug-12046.test.ts > bug 12046: duplicate PATH/Path in Windows PTY env > keeps a single case variant of PATH when the inherited env uses Windows `Path`
AssertionError: expected [ 'Path', 'PATH' ] to have a length of 1 but got 2

 FAIL  src/main/attribution/terminal-attribution.bug-12046.test.ts > bug 12046: duplicate PATH/Path in Windows PTY env > strips attribution entries from the inherited `Path` when attribution is disabled
AssertionError: expected '/var/folders/td/0d_w3jrs3654mtmk1gd6k…' not to contain 'orca-terminal-attribution'
```

After the fix — re-run on the PR head (`d7e596afb3`):

```
    config/scripts/build-windows-cli-launcher.test.mjs \
    src/main/pty/windows-environment-path.test.ts \
    src/main/runtime/claude-agent-teams-service.test.ts \
    src/main/attribution/terminal-attribution-path-env-casing.test.ts \
    src/main/providers/local-pty-provider.test.ts \
    src/main/daemon/pty-subprocess.test.ts

 RUN  v4.1.5

 Test Files  6 passed (6)
      Tests  240 passed | 1 skipped (241)
   Duration  3.42s (transform 866ms, setup 0ms, import 1.24s, tests 3.18s, environment 0ms)
```

## No regressions

Every pre-existing suite neighbouring a touched file, plus the suites that import what changed:

```
$ npx vitest run --config config/vitest.config.ts src/main/attribution src/main/pty \
    src/main/providers src/main/daemon/pty-subprocess.test.ts \
    src/main/runtime/claude-agent-teams-service.test.ts \
    src/main/runtime/claude-agent-teams-pty-exit-leak.test.ts \
    src/main/ipc/pty.test.ts src/main/ipc/pty-management.test.ts \
    src/main/runtime/orca-runtime.test.ts \
    config/scripts/build-windows-cli-launcher.test.mjs \
    src/main/cli/windows-launcher-asset.test.ts

 Test Files  65 passed (65)
      Tests  2366 passed | 2 skipped (2368)
   Duration  12.24s (transform 20.07s, setup 0ms, import 33.43s, tests 21.92s, environment 3ms)
```

`npx oxlint` on the changed files is clean, as are `--config config/oxlint-code-quality-native-plugins.json --deny-warnings` and `--type-aware --config config/oxlint-code-quality-type-aware.json --deny-warnings`. `pnpm run typecheck:node` passes; no `src/cli` or renderer code is touched.

**Verification gap (please read before merging).** Every result above was produced on a **macOS** host. This is a Windows-only fix and no part of it has been exercised on a real Windows machine:

- The TypeScript half is covered only by unit tests that *simulate* Windows — they stub `process.platform` to `'win32'` and hand-build dual-cased env objects. Nothing here proves how a real Windows env block arrives at `buildPtyHostEnv`, nor that node-pty preserves key order end-to-end on win32. The block-order model (`RtlQueryEnvironmentVariable` returns the first case-insensitive match; `Object.keys` order is block order) is reasoned from documented behavior, not observed on Windows in this PR.
- The .NET half cannot be compiled or executed here at all — there is no `dotnet`/`mono`/`csc` on the host. It is covered only by a cross-host source assertion that `OrcaCliLauncher.cs` references neither `EnvironmentVariables` nor `startInfo.Environment` and does call `Environment.SetEnvironmentVariable`. That assertion uses `itCrossHost`, so it is *skipped* on Windows CI.

Windows CI must build `orca.exe` via `config/scripts/build-windows-cli-launcher.mjs` and run the existing `itWindows` spawn test, which is the inheritance canary: it asserts the child still receives `electronRunAsNode: '1'`, `nodeOptions: null`, `orcaNodeOptions: '--no-warnings'` and the round-tripped multiline argument. That test passing is what proves the NULL-`lpEnvironment` inheritance assumption holds. I did not add an end-to-end dual-cased-block case driven through node-pty, because I have no way to validate it here and a Windows-only test I cannot run is a liability rather than evidence.

### Trade-offs

**One intentional, user-visible Windows behavior change.** Today a Windows Orca terminal receives both spellings, and the child resolves whichever one sits first in the block — typically the inherited `Path`, not the freshly appended `PATH` that attribution wrote its shim into. So the attribution shim directory won only when its spelling happened to come first. After this change Orca writes into the first-listed key, so the shim directory deterministically wins. For users with `enableGitHubAttribution` on, that means:

- the Orca commit trailer and PR/issue footer will now reliably appear on Windows, where before they appeared intermittently or not at all;
- **every** `git` and `gh` call inside a Windows Orca terminal will now reliably route through `git.cmd` / `gh.cmd` / `*-wrapper.ps1`, costing one extra `cmd.exe` or PowerShell process per invocation.

Those Windows wrappers have only ever been exercised intermittently in the field, so latent wrapper defects (exit-code propagation, stdin piping, quoting) become reachable for the first time. This is the documented setting working as specified rather than a regression, but the win32 wrapper bodies in `src/main/attribution/terminal-attribution.ts` (written at `:176`–`:179`, bodies in the `WIN32_*_WRAPPER` constants) deserve extra scrutiny in review. Combined with the macOS-only verification above, this is the highest-risk part of the change.

The daemon collapse is the only place a key is deleted, and it discards the value under the non-surviving spelling. That value is the daemon's own `process.env` PATH, which is shadowed by nothing and superseded by the main-shaped patch that survives; the alternative — leaving both — hands the child a block where the daemon's unshaped spelling sits *first* and therefore wins, silently discarding attribution and CLI shims. The one in-repo sparse-patch case is Agent Teams' shim-dir `PATH`, which `promoteAgentTeamsShimPath` re-prepends from a value captured before `buildPtyHostEnv` runs (`ipc/pty.ts:3933`, `:5093`) — and the `claude-agent-teams-service.ts` change stops the truncated value being produced at all.

## Cross-platform

- **macOS / Linux**: unchanged in every path that matters. `resolvePathEnvKey` returns `'PATH'` unconditionally when `platform !== 'win32'`, and the daemon collapse is inside a `process.platform === 'win32'` branch, so a legitimate case-sensitive POSIX `Path` variable is never written or deleted. There is an explicit regression test for that (`always resolves 'PATH' off Windows so a POSIX 'Path' variable is untouched`, plus a `platform: 'linux'` attribution case with a decoy `Path`). One deliberate POSIX-visible difference, so this is not a byte-identical refactor: `readInheritedPath` (`ipc/pty.ts:745`) used to fall back to `baseEnv.Path` on *any* platform when `baseEnv.PATH` was absent, and no longer does — a POSIX-only `Path` variable is now ignored rather than read. That is the intended "never touch a POSIX `Path`" behavior. The Linux bare-`orca` shim branch at `ipc/pty.ts:1418` is otherwise left alone — it is inside `process.platform === 'linux'`, where the resolver provably returns `'PATH'`.
- **Windows**: the whole point of the change — and, per the verification gap above, the one platform on which none of it was actually run. Native shells, Git Bash, and WSL terminals (which resolve to `shellFamily: 'posix'` but still run on a Windows host) all have their PATH read and written under a single key — the one Win32 resolves. Orca no longer adds a second key; a pre-existing pair is passed through untouched.
- **SSH-remote**: untouched. `buildPtyHostEnv`'s doc comment (`ipc/pty.ts:1240-1245`) states it must not be called when `args.connectionId` is set, so remote shells never see any of this. Their env is assembled by `src/main/providers/ssh-pty-spawn-env.ts`, which has its own key resolution at `:12` and is deliberately unchanged — remote hosts are POSIX in every supported configuration and its `PATH`-first order is correct there. The packaged `orca.exe` launcher is a local-host binary only.
- **Folder workspaces (non-git)**: irrelevant to this diff. Nothing here reads repo metadata, runs Git, touches `.git`, or assumes a worktree.
- **Git compatibility**: no Git subcommand or option is added or changed, so the 2.25 baseline, `GitCapabilityCache`, and the leading-`-c` preservation rule are not in play. Shim generation for `git`/`gh` is unchanged and provider-agnostic — the wrappers call plain `git`, so GitLab and other hosts are unaffected, and no GitHub-only naming is introduced.
- No keyboard shortcuts, accelerators, or shortcut labels are involved; no `e.metaKey`. No string path concatenation is added — attribution keeps `join`/`win32 as pathWin32`, and the `promoteAgentTeamsShimPath` copies keep their `delimiter` import.

## Performance

`resolvePathEnvKey` is one `Object.keys` walk over an env object already in hand, called a handful of times per PTY spawn; the daemon collapse adds one filtered key list and at most one `delete` per spawn. No new process spawns, no new I/O, no registry reads, no polling, no IPC chatter, no renders.

The C# change strictly **removes** work from the packaged-CLI hot path: the launcher no longer materializes and serializes a full copy of the environment block on every `orca` invocation, so CLI startup gets marginally faster.

The one real second-order cost is stated under Trade-offs above: making the win32 attribution shim dir deterministically first means `git`/`gh` calls in a Windows Orca terminal now reliably pay for a wrapper process, instead of paying for it only when the shim's spelling happened to come first in the block.

## Security

No change to the auth surface. No IPC channel is added, removed, or given new payload shape. No new command execution: the launcher spawns the same `Orca.exe` with the same `BuildArguments`/`QuoteArgument` output, and the arg-quoting and exit-code propagation are untouched. Path handling is unchanged — no new string concatenation, existing `path.join` / `pathWin32.join` usage preserved.

Two things worth naming explicitly. First, the launcher now passes `lpEnvironment = NULL` instead of an explicitly built block, so the child inherits the parent's environment verbatim; that is strictly closer to standard Windows process-launch behavior than the previous filtered copy, and the `NODE_OPTIONS` / `NODE_REPL_EXTERNAL_MODULE` neutralization that prevents arbitrary module injection into the Electron-as-Node child is preserved — it now happens on the launcher's own env before `Process.Start` rather than on the copy. Second, the daemon's collapse deletes an environment key; it only ever fires on Windows, only when its own `{...process.env, ...opts.env}` merge produced two PATH-shaped keys, and the surviving value is the one main shaped for that spawn. Nothing deletes an environment key on POSIX, and attribution — which runs on the user-facing terminal env — deletes no path key it did not itself create.

## AI Review Report

- **Final verdict:** CLEAN — the final round surfaced no blocking findings. One non-blocking note was recorded and is listed below.

**Non-blocking note (1) — CONFIRMED, fixed:**
- `src/main/attribution/terminal-attribution.ts` — in `clearTerminalAttributionEnv` the empty-result branch deleted the resolved key and the (then-present) collapse deleted the *other* one, leaving the child with no PATH-shaped variable. With the collapse gone from attribution, the shadowed spelling survives and becomes the live one, which is exactly what Win32 would have done. Regression test: `keeps a shadowed duplicate when the live spelling strips down to nothing`.

**Blocking finding raised in an earlier round (1) — CONFIRMED, fixed in `d7e596afb3`:**
- **major** `src/main/pty/windows-environment-path.ts` — `resolvePathEnvKey` picked the surviving Windows spelling by a hardcoded `Path`-first preference (`if (env.Path !== undefined) return 'Path'`), and a `collapseWindowsPathEnvKey` helper in `src/main/attribution/terminal-attribution.ts` then **deleted** the other spelling along with its value. Win32 does not resolve duplicates by casing: `RtlQueryEnvironmentVariable` linearly scans the block and returns the *first* case-insensitive match.
  - *Failure scenario:* an inherited block in the order `PATH=C:\tools\bin;C:\Windows\System32` then `Path=C:\Windows\System32`. Win32 resolves `PATH`, so `C:\tools\bin\foo.exe` runs; the old code resolved `pathKey='Path'`, wrote the shim onto the dead spelling, and deleted the live `PATH` — every Orca-spawned Windows process lost `C:\tools\bin`.
  - *Fix:* `resolvePathEnvKey` now returns the first `PATH`/`Path` own key in enumeration order (block order — Node enumerates `process.env` in block order and node-pty emits `Object.keys(env)` unsorted), for both the env and the host fallback. The attribution-side collapse is **removed** outright: nothing on the user-facing terminal env deletes a path key any more, so no inherited value can be discarded. Regression tests: `resolves a dual-cased Windows env by block order, not casing` and the parametrised `prepends the shim onto the first-listed …` pair.

**Remediation applied:** 2 pass(es). Findings judged false-positive were rejected with evidence rather than coded around.


nwparker:
## Post-merge verification on a real Windows host

> **Correction (edited).** An earlier version of this comment implied these Windows-gated tests had never executed in CI. That was wrong, and the record should be straight: `native-smoke (windows-latest)` **does** run them. `.github/workflows/computer-e2e.yml:93` names `config/scripts/build-windows-cli-launcher.test.mjs` explicitly in its vitest invocation, and the `paths:` filter matches this change via both line 10 and `native/windows-cli-launcher/**`. The job succeeded at `2026-08-04T03:45:05Z`, 18 seconds before merge. The `windows` check being `SKIPPED` is a *different* job; conflating the two produced the error.
>
> **What the real-hardware run added on top of CI**, which is where the value actually is: CI's dual-block test asserts only that inheritance is verbatim (`pathKeys: ['PATH','Path']`, `electronRunAsNode: '1'`). **It never checks which spelling Windows resolves** — it would pass identically under a `Path`-preference resolver, i.e. under the exact bug review round 3 caught. The four-ordering marker experiment below is what actually tests the fix's central premise, along with order surviving `{...process.env}` → node-pty `_parseEnv` → real spawn, and the neighbour-suite run that surfaced the pre-existing `pty.test.ts` failures.


Verified end-to-end on real hardware after merge. Windows 11 build 26200.8875, Node v24.18.0, worktree at `153740326c4832222ab556d7290910ee3b7609f4` (this PR's head).

**Result: the core premise is empirically correct and no defect was found in this PR.**

### The crux — Win32 resolves duplicate env keys positionally, not by casing

Raw `CreateProcessW` block via P/Invoke (this PR's own fixture mechanism), two spellings with **different** values, plus a real `orcamarker.cmd` in each directory so the answer comes from Windows' own PATH search rather than from reading the block:

```
ORDER 1  PATH(live) first, Path(shadowed) second  ->  win32 resolves ...\live       marker = LIVE
ORDER 2  Path(shadowed) first, PATH(live) second  ->  win32 resolves ...\shadowed   marker = SHADOWED
ORDER 3  pAtH(live) first, PATH(shadowed) second  ->  win32 resolves ...\live       marker = LIVE
ORDER 4  PATH(shadowed) first, pAtH(live) second  ->  win32 resolves ...\shadowed   marker = SHADOWED
```

**First-listed wins in all four orderings, regardless of casing.** If casing preference were the rule, orders 1 and 2 would resolve the same spelling — they resolve opposite ones. `resolvePathEnvKey`'s first-listed-wins is correct, and a `Path`-first casing preference would have targeted the dead spelling in roughly half of all real dual blocks.

### Order preservation holds through the whole chain

- `{...process.env}` measured *inside* a genuine dual block: `["PATH","Path"]` → `["PATH","Path"]`, `["Path","PATH"]` → `["Path","PATH"]`.
- node-pty `_parseEnv` (installed source): `Object.keys(env)` → `key=value`, **no sort, no dedupe**.
- node-pty end-to-end, the path Orca terminals actually take:
```
{PATH:live, Path:shadow}  keysSent=[…,"PATH","Path"]  -> LIVE
{Path:shadow, PATH:live}  keysSent=[…,"Path","PATH"]  -> SHADOWED
```

### The Windows-gated tests, re-run on real hardware

`csc.exe` was present, so the `itWindows` cases executed:

```
✓ preserves a multiline argument from PowerShell through the native launcher   2121ms
✓ survives an inherited environment block containing PATH and Path             2540ms
```

The second compiles the real `orca.exe` via `build-windows-cli-launcher.mjs`, compiles the P/Invoke fixture with `csc.exe`, and launches the real launcher with a genuine `PATH=C:\live` + `Path=C:\shadowed` block. That is the reported crash, verified against the real launcher. The NULL-`lpEnvironment` inheritance assumption holds.

PR suites on Windows: `Test Files 6 passed (6) · Tests 245 passed | 6 skipped (251)` (macOS ran 240/1 — Windows runs 5 more real tests and skips the `itCrossHost` source assertions).

## Two things worth recording

**1. A subtlety no simulated test could show.** After `{...process.env}`, *both* keys carry the same (live) value — Node's env proxy reads case-insensitively, so the shadowed value is unreachable from JS. The concern that "deleting the shadowed spelling discards a live value" therefore **does not apply to the `{...process.env}` path**; there is no distinct value there to lose. It applies only to hand-built objects such as the daemon's sparse-patch merge — which is exactly where this code collapses deliberately. The decision is right; the justification in the description is broader than the evidence supports.

**2. `child_process.spawn` does not behave like node-pty.**

```
env {PATH:live, Path:shadow}  -> libuv keys ["PATH"]  value ...\live   marker LIVE
env {Path:shadow, PATH:live}  -> libuv keys ["PATH"]  value ...\live   marker LIVE
```

libuv's `make_program_env` de-duplicates case-insensitively and collapses to a single `PATH` — and in *both* orders keeps the value under the `PATH` spelling, i.e. **not** first-listed. So on a `Path`-first block, an env object whose shim was correctly written into `Path` (right for node-pty) would, if handed to `child_process.spawn`, have libuv keep the un-shimmed `PATH` and silently drop the shim.

No live Orca path hits this today — user-facing terminals go through node-pty and the daemon collapses explicitly. But the model sentence "object key order is block order" is true **only for node-pty**, and someone will eventually assume it generalizes. Worth a code comment.

## Review comments

- **"Handle every Windows PATH case variant (`pAtH`)"** — already satisfied; the comment was against an earlier revision. Merged code uses `key.toLowerCase() === 'path'`. Verified on Windows: a `pAtH`-first block resolves `pAtH` and the helper picks it.
- **"Preserve the current `Path` precedence when both standard spellings exist"** — respectfully rejected. This would reintroduce the exact defect review round 3 caught. Order 2 above has `Path` first and Win32 resolves `Path`; order 1 has `PATH` first and Win32 resolves `PATH`. Casing precedence targets the dead spelling half the time.
- **"Sort the custom environment before passing it to `CreateProcess`"** — rejected, with the docs caveat noted. The documentation point is real, but a case-insensitive sort makes `PATH` and `Path` compare *equal*, so their relative order — the entire thing under test — is precisely what sorting leaves undefined. Decisively: node-pty passes unsorted blocks to `CreateProcess` on every Orca terminal spawn in production, so if unsorted blocks were unusable Orca would already be broken on Windows.

## Not verified — the biggest remaining surface

**The attribution shim wrappers.** This change makes the shim dir deterministically win, so *every* `git`/`gh` call in a Windows Orca terminal now routes through `git.cmd` / `gh.cmd` / `*-wrapper.ps1`. The `WIN32_*_WRAPPER` bodies — exit codes, stdin piping, quoting — were **not** exercised. The description itself flags these as latent-defect-prone and newly reachable, and they are now live on `main`. Also unverified: a packaged/installed `orca.exe` (the launcher was built from source with the repo's own script), and a dual block produced by real WSL interop rather than synthesized via P/Invoke.

Separately, 3 pre-existing Windows failures in `src/main/ipc/pty.test.ts` were found while running the neighbour set. They reproduce identically on this PR's base commit, so they are unrelated to this change — filed separately.

PR #12220, opened 2026-08-03
Keep the SSH connection alive when a terminal outruns a slow link

### LINKED ISSUE #12041: [Bug]: SSH disconnected between Local and VPS when trying to use claude code cli only after latest release v1.4.163
**Summary**
After updating to 1.4.163 via the in-app update button, two separate problems appeared
when using Orca to reach a remote host over SSH. **Both disappeared completely after
rolling back to 1.4.150** (same machine, same network), so both are regressions
introduced somewhere in 1.4.151–1.4.163.

**Environment**
- Orca desktop app, **Windows** client
- **Broken: 1.4.163** · **Confirmed working: 1.4.150** (rollback fixes both)
- Regression introduced somewhere in **1.4.151 – 1.4.163**
- Remote host: Linux VPS (Europe/Frankfurt), reached over **Tailscale SSH**
- Client in South Korea; this link happens to be **~268 ms RTT** (see mechanism note)
- Claude Code v2.1.220 on the host

**Problem 1 — Claude Code prompt input auto-indents (separate issue)**
Typing normally into the Claude Code prompt input box makes the input auto-indent /
insert line breaks on its own. Input-box handling glitch; unrelated to Problem 2.

**Problem 2 — Claude Code session drops on any interaction**
Other TUIs (codex, plain shell, etc.) work fine. But after launching `claude`, any
interaction — a single keystroke, a Shift+Tab mode switch, anything — immediately
drops the connection and enters a reconnect loop.

**Confirmed — relay bundle diff on the host (1.4.150 vs 1.4.163)**
I compared the two relay bundles directly on the host:
- **1.4.150 (working):** contains NONE of `2097152`, `producerQueueMaxBytes`,
  `legacy`, `clientHighBytes`, `closeClient`, `enqueueFrame` — it has no output-size
  cap and never closes a client over output volume.
- **1.4.163 (broken):** introduces a whole new lane-queue output subsystem with a hard
  2 MiB "legacy-response" lane and closes the client when exceeded. The host relay log
  shows:

      Relay response exceeds the bounded 2097152-byte legacy lane

So the regression is unambiguously this newly-introduced bounded output subsystem.
Claude's TUI emits large full-screen redraws on each interaction; on this connection
those responses exceed the bound and the client is dropped. Lighter TUIs stay under the
bound, which is why only `claude` triggers it.

**Mechanism + the one thing I did NOT verify**
The cap is on the *cumulative undrained* queue: a frame is admitted only while
`producerBytes + frame <= producerQueueMaxBytes` (2 MiB). It fills when the client
drains slower than Claude's TUI produces — a backpressure/flow-control bound sensitive
to client drain rate (latency, bandwidth, render speed), not a fixed per-message size.
And it **closes the client instead of pausing the PTY**. I hit it reliably on a
~268 ms link but did **not** test a low-latency link, so I can't claim latency is
strictly required — any client that can't drain fast enough will hit it. **Please
don't scope this to "high-latency users only."**

**Impact**
Makes Claude Code unusable over any high-latency (or otherwise slow-draining) remote
link in 1.4.163.

**Workaround**
Downgrade to **1.4.150** — fully resolves both problems with no other change.

**Suggested fixes**
1. Chunk/stream large TUI redraws so a single client response can't exceed the
   legacy-lane bound (or raise / make the 2 MiB cap configurable).
2. Apply PTY backpressure (pause output) before the bound is hit, instead of closing
   the client.
3. Bisect 1.4.151–1.4.163 for the input-path change behind Problem 1 (prompt
   auto-indent).

nwparker:
> ## ⚠️ Draft — not ready to merge
>
> Three adversarial review rounds each fixed a finding and surfaced another, and **three blocking findings are still open** (see [the conversion comment](https://github.com/stablyai/orca/pull/12220#issuecomment-5164324839) and the AI Review Report at the bottom). CI is green — 41 checks pass, 1 skipped — so the objection is behavioural, not a failing test.
>
> Short version: eviction of a saturated non-owner client is **event-driven on a projected frame**, but the condition that makes eviction necessary (its retained bytes hold `legacyRetentionBelowLowWater` false) is also the condition that stops PTYs resuming and therefore stops projections being produced. `src/relay/pty-handler.ts:1184` is the only `pty.resume()` in the handler and it sits behind that relay-wide gate. So a permanently stalled subscriber can wedge **every** PTY for **every** viewer with nothing on a bounded timeline to release it. The reproduction and diagnosis below are solid and worth keeping; the remaining work is a design decision about relay backpressure semantics.

## Summary

Two producer paths in the relay treated "this client's write queue is momentarily full" as a fatal protocol error and destroyed the client's sink. Over a high-latency SSH link (the reporter's ~268 ms Local↔VPS hop) Claude Code's full-screen redraws outrun the client's drain rate, so the bound is hit on nearly every interaction: the link dies, the app reconnects, the same output and replay are regenerated, and the loop repeats.

- `RelayDispatcher.projectToClients` (`src/relay/dispatcher.ts`) called `closeClient(..., 'Relay PTY subscriber projection capacity exceeded')` whenever `publishToClient` returned false — which happens purely for a transiently-full producer queue. It now classifies each projection target instead, via the new `LegacyProjectionRole`:
  - **`owner`** — a legacy session owner (`deliveryMode === 'legacy-owner'`), for which the projection is the *only* delivery lane. **This is the #12041 fix.** A full queue now fails the whole projection, so `RelayPtySourcePublication.publish` returns false, `PtyHandler.flushPtyOutput` pauses the native PTY, and the byte-identical memoized span is republished once the sink drains. Only a frame larger than that sink's `producerFrameCapacity` is shed, because retrying it forever would wedge the PTY.
  - **`mirror`** — every other target: agent hooks, `orca` CLI processes, a second window, a reconnecting client in the window before its `pty.openClient` completes. A mirror never back-pressures the native PTY (that keeps #11005's per-client isolation), but a mirror that cannot admit a frame it is offered is still **closed**, exactly as on `main`. See "Why a mirror is closed rather than shed" below — an earlier revision of this branch shed the frame instead and that turned #12041's disconnect into a permanent, relay-wide stall.
  - Mirrors are published only after the owners commit, so a back-pressured span that is later retried never reaches a mirror twice. (This ordering is also part of open finding **B2** below: when an owner back-pressures, `projectToClients` returns before the mirror loop runs, so a saturated mirror is not visited on that frame.)
  - `projectPtyExitToLegacyClients` follows the same rule: held for retry for an owner, close for a saturated mirror, so no pane is left rendering a process that already exited.
- The `pty.replay` branch of `notify()` deliberately routed replay to the `control` lane, where overflow closes the writer. With `REPLAY_BUFFER_MAX` at 100 KiB per PTY and a workspace restoring many terminals, the cumulative undrained replays blew the 1 MiB control bound and destroyed the connection — and reattach regenerated the very same replays. It now enqueues with `controlOverflow: 'reject'` and logs a single dropped-replay line naming the control lane; the pane restores without scrollback and live output repaints it.

Two supporting changes keep the new backpressure from breaking neighbouring guarantees:

- `RelayPtySourcePublication.publish` now calls `sender.pump(record)` even when the legacy projection is rejected. The span is already in the source ledger, so without this a stalled legacy owner would freeze the flow-controlled V1 owner's credited send loop — exactly the isolation #11005 built the source lane for.
- `PtyHandler` tracks the ptys currently inside a publish. `flushPendingOutput` lifts a pty's queue out of `pendingOutputByPty` and drains a captured copy, so during the publish the map is transiently empty; a capacity callback re-entering from the sink would read "no pending output" and let `pty.exit` overtake a span that is about to be re-queued. `publishPendingExit` now also checks the in-publish set.

The exact error string the reporter quoted (`Relay response exceeds the bounded 2097152-byte legacy lane`) came from `sendResponse` in v1.4.163 and was already removed by #11999; only the two paths above survived on `main`.

### Why a mirror is closed rather than shed

A revision of this branch shed the frame for a saturated mirror and closed nobody. That is a **worse** bug than #12041, and it was reproduced before being reverted:

- Every projected frame reserves a `LegacyRelayPublicationLedger` lease per target, released only when that sink settles the write. A mirror that never drains holds up to `LEGACY_CLIENT_RETAINED_BYTES_HIGH` (2 MiB), well above the 1 MiB `LEGACY_CLIENT_RETAINED_BYTES_LOW` mark.
- `RelayDispatcher.legacyRetentionBelowLowWater` is `publicationLedger.belowLowWater(activeClientKeys())`, so one such mirror pins it to `false` for the whole relay.
- `PtyHandler.maybeResumePtyOutput` (`src/relay/pty-handler.ts:1172`) early-returns while that gate is `false`, and line 1184 holds the only `pty.resume()` call in the handler. So **every** paused PTY on the host — including terminals with a perfectly healthy owner — stayed paused forever, and `notifyLegacyCapacity(false)` never fired.
- With shedding, nothing else evicted that client: it never errors (it just does not read), so `DispatcherClientWriter.handleWriteSettlement` never fires, and the stall timer this branch briefly had was correctly deleted for judging healthy peers.

Closing bounds it **on the frame the mirror is offered and cannot admit**: `closeClient` releases every queued and in-flight entry (`DispatcherClientWriter.close` → `releaseEntry` → the lease's `release`), drops the client key out of `activeClientKeys()`, and forces a capacity fan-out — so the relay recovers on that frame. The mirror reconnects and reopens the pty, which replays it. Regression test: `dispatcher-slow-client.test.ts` → "releases the relay-wide legacy gate when a mirror can never drain".

**This is where the branch is still wrong, and why it is a draft.** Recovery is conditional on another frame actually being projected to that mirror, and two reachable states prevent that:

- If the relay-wide gate is already `false`, every PTY is paused and produces no output, so no projection is generated and nothing evicts the client that is holding the gate down (the mirror loop at `src/relay/dispatcher.ts:1185` is only reached from a live projection).
- A mirror's producer queue can also fill from non-projection paths (`tryNotifyPtyDataToClient`, agent-hook notifies, responses), and `projectToClients` returns at `src/relay/dispatcher.ts:1177` whenever an owner back-pressures — so the mirror loop that would close it never runs on those frames.

Both leave a saturated non-owner attached with up to 2 MiB of retained leases and no bounded path to release them. See the AI Review Report.

### Scope against #12041

Issue #12041 reports **two** problems. This PR targets **Problem 2 only** (the Claude Code session dropping into a reconnect loop on any interaction), and only the close-instead-of-backpressure class of it. It does **not** fix **Problem 1** (the Claude Code prompt input auto-indenting / inserting line breaks on its own), which is an input-path regression the reporter explicitly filed as separate and which nothing here touches. #12041 therefore stays open for Problem 1 — this PR is deliberately not marked `Closes #12041`. It is also not merge-ready for Problem 2 in its current shape.

Relates to #12041.

## ELI5

Your terminal on the server can produce text much faster than a slow internet connection can carry it. When the outgoing pipe filled up, the server decided something had gone badly wrong and hung up the connection. Reconnecting made the server resend all the same text, which filled the pipe again, which hung up again — a loop you can't escape. Now, if the pipe belongs to the window that owns the terminal, the server just pauses the terminal for a moment and waits for the pipe to empty instead of hanging up. A bystander watcher (a second window, a CLI tool, a hook) that has completely stopped reading is dropped the next time the server tries to send it something, because leaving it attached would freeze every terminal on the machine. The unfinished part: if all the terminals are already paused *because of* that watcher, the server never tries to send it anything again, so it never gets dropped and the terminals stay frozen. That is the open objection.

## Fix proof

The pre-fix behaviour is the behaviour the pre-existing tests this PR changes used to assert:

- `dispatcher-capacity-degradation.test.ts` "closes the client when pty.replay overflows the control queue" asserted `primary.closes === 1` on `main` — that is the replay path killing a link the reconnect would immediately regenerate.
- `projectToClients` on `main` closed **every** target that could not admit the frame, session owner included; that is the line the new `owner` role replaces.

The new permanent regression suite is `src/relay/dispatcher-slow-client.test.ts`. Re-run for this description on the PR head (macOS, `npx vitest run --config config/vitest.config.ts src/relay/dispatcher-slow-client.test.ts`):

```
 Test Files  1 passed (1)
      Tests  12 passed (12)
```

Three of those twelve fail against the previous revision of this branch (`9dad5c75b`, the one that shed a mirror's frame and closed nobody):

```
 × closes a saturated mirror instead of pausing the PTY for every other viewer
 × releases the relay-wide legacy gate when a mirror can never drain
     AssertionError: expected false to be true   (dispatcher.legacyRetentionBelowLowWater)
 × logs one dropped-replay line per generation however many panes are stranded
```

and so do the two neighbouring suites that assert the same contract end to end — re-verified by running the three suites against `9dad5c75b`'s sources: `Tests  5 failed | 36 passed (41)`.

```
 × relay-pty-source-publication.test.ts   > detaches only a saturated subscriber while the V1 owner stays live
 × pty-handler-source-publication.test.ts > keeps the native PTY and V1 owner live when one subscriber saturates
```

No test covers the two states in "Why a mirror is closed rather than shed" where no frame is ever projected to the saturated mirror; those are open findings **B1**/**B2**, not proven-fixed cases.

## No test regressions

Whole relay suite (contains every file touched, plus `dispatcher.test.ts`, `dispatcher-capacity-degradation.test.ts`, `dispatcher-client-writer.test.ts`, `dispatcher-frame-guard-regressions.test.ts`, `pty-handler*.test.ts`, `relay-pty-source-*.test.ts`, `legacy-relay-publication-ledger.test.ts` and the two backpressure integration suites):

```
 Test Files  103 passed (103)
      Tests  1232 passed (1232)
```

Every suite outside `src/relay` that imports the relay dispatcher or pty handler:

```
 src/main/plugins/plugin-host-conformance.test.ts
 src/main/ssh/ssh-relay-session-agent-hooks.integration.test.ts
 src/main/agent-hooks/wsl-hook-relay-manager.test.ts
 Test Files  3 passed (3)
      Tests  36 passed (36)
```

`pnpm run typecheck:node`, `typecheck:cli`, `typecheck:web`, `oxlint`, the native-plugin and type-aware oxlint configs with `--deny-warnings`, and `check:max-lines-ratchet` are all clean, and PR CI is green (41 passed, 1 skipped). **Green tests do not clear this branch** — the open findings are behavioural states no suite exercises.

Pre-existing tests that encoded the old behaviour and were deliberately changed:

- `dispatcher-capacity-degradation.test.ts` "closes the client when pty.replay overflows the control queue" -> now asserts no close plus exactly one `[relay] Dropped pty.replay (…B, control queue full)` line.

Tests that were changed by an earlier revision of this branch and have been **restored to their `main` assertions**, because that revision's shed-instead-of-close is the wedge described above:

- `pty-handler-source-publication.test.ts` "keeps the native PTY and V1 owner live when one subscriber saturates" — the saturated subscriber is detached; the native PTY is not paused; the healthy mirror and the V1 owner are served.
- `relay-pty-source-publication.test.ts` "detaches only a saturated subscriber while the V1 owner stays live".

`dispatcher-capacity-degradation.test.ts` "still closes on protocol-critical control queue overflow" stays green: `notifyClient` keeps the default `close-client`.

### Trade-offs

1. **A session owner still back-pressures every viewer, and its retention still gates the relay.** A `legacy-owner` (a client that completed `pty.openClient` as session owner but without `outputFlowControl`) has no second delivery lane, so a full queue there pauses the native PTY until it drains — and because `legacyRetentionBelowLowWater` is a relay-wide aggregate, that pause is shared by other PTYs on the host. That is the same all-or-nothing contract `tryNotifyPtyData` — the non-source path used for every PTY without a V1 consumer session — has always had on `main`. It is **not** reliably bounded: `handleLegacyCapacity` resumes on drain, and a socket that *errors* is closed by `DispatcherClientWriter.handleWriteSettlement` (or detached by the `sock.on('close')` handler in `relay.ts`), but a peer that simply stops reading — a zero-window TCP receiver, a suspended VM — never errors and is never evicted. Narrowing the aggregate to per-PTY scope is a separate change and is deliberately not attempted here. This is open finding **B3**.
2. **A saturated bystander is disconnected on the next frame projected to it** (same disposition as `main`, but no longer unconditional in timing — see **B1**/**B2**). An agent hook or `orca` CLI process that stops reading loses its relay connection and reconnects. That is not the failure #12041 reports — the reporter's client is the session owner — and the alternative (unconditional shedding) provably wedges every terminal on the host.
3. **A frame larger than the owner sink's `producerFrameCapacity` is shed, not retried.** Untransformed spans are pre-sized by `maxLegacyPtyDataChars` so this is only reachable for a transformed ingress emission. Retrying is impossible (it can never be admitted) and closing produces the #12041 loop verbatim: reconnect, replay, regenerate the same oversized frame, close. It is logged once per method per client generation.
4. **A `pty.replay` the control queue cannot admit is dropped**, so that pane restores without scrollback until new output repaints it. Only reachable by clients older than #1026 — every current client passes `suppressReplayNotification: true` and takes the response path instead.

There is deliberately **no timer that reaps a stalled peer**. An earlier revision of this branch had one; it judged a peer by "retains producer bytes and settled nothing since arming", which is true of any socket that merely has a frame in flight, and its single global one-shot timer could judge a freshly-armed peer over as little as 1 ms. Closing a healthy client is precisely the disconnect #12041 is about, so the mechanism was removed rather than tuned — `dispatcher-slow-client.test.ts` → "never closes a peer that is merely mid-write while another target back-pressures" locks that in. **Removing it without replacing it is the gap**: with eviction event-driven on a projected frame, a client that is never offered another frame is never evicted, so nothing bounds the ledger on a timeline. A correctly-scoped bounded release (or per-PTY gate scoping) is the design decision this draft is waiting on.

## Cross-platform

All edits are in the host-side relay bundle: plain Node, no Electron, no DOM, no keyboard handling or accelerators, no `metaKey`, no filesystem paths, no shell invocation, no git subcommand (so the Git 2.25 baseline and `GitCapabilityCache` rules do not apply) and no provider-specific surface, so GitHub/GitLab parity is untouched. Behaviour changes identically for macOS, Linux and Windows clients and for native, WSL and SSH hosts — the reporter's Windows-client/Linux-VPS pairing sits inside that matrix. **All of it was only executed on macOS** (see limitations).

One platform-adjacent detail the fix depends on and that is already handled: `producerFrameCapacity` derives from the sink's `writableHighWaterMark` (`src/relay/dispatcher-writer-sink.ts`), which differs between a Windows named-pipe stdout, a POSIX pipe and a forwarded socket. The shed branch keys off that per-sink value rather than a constant, and an infinite or unknown high-water mark falls back to `MAX_SAFE_INTEGER`, so the shed path is never taken spuriously.

SSH-remote is the primary case, not an afterthought: the defect only manifests when the client drains slower than the host produces, and reconnect is explicitly in scope — `setWrite` builds a fresh writer with an empty queue, and during the window before the reconnected client completes `pty.openClient` it is classified `subscriber`, so it is a `mirror`; reaching a mirror close from that fresh, empty queue requires 2 MiB of undrained `pty.data` inside a single `openClient` round trip.

Folder workspaces vs git worktrees are irrelevant here: nothing inspects repositories, worktrees or git state; this is per-PTY output routing only.

## Performance

Slightly positive on the hot path. `projectToClients` previously re-encoded the frame once per target client (it called `publishToClient` without the already-existing `estimatedBytes` argument); it now encodes once and threads the size through both that argument and the **new** optional `estimatedBytes` parameter on `tryPublishToClients`, and still returns before encoding at all when there is no projection target (the norm — a flow-controlled owner is excluded from the mirror). Added cost is one classifier call per client per frame (a map lookup that the previous `matchesClient` predicate already did) plus one `Set.add`/`Set.delete` per publish in `PtyHandler`.

`sender.pump` on the rejected path early-returns on its existing activating/sending/turnScheduled/restoreRequired/rotationPending guards when there is nothing to do. The `pty.replay` change removes work (no second enqueue attempt). No timers, polling, IPC or renders are added. Steady state on a healthy link is byte-identical. The pathological state in the open findings is not a throughput cost but a full stop: paused PTYs produce nothing.

## Security

No change. No IPC surface added or altered, no command execution, no path handling, no authentication or authorization logic touched, and no wire-format change — `pty.replay` still travels on the control lane, only its overflow disposition changed. The only new observable output is a stderr diagnostic that logs a method name and byte count, both already logged by the neighbouring `logDroppedProducerNotification` callers, and is rate-limited to one line per method per drop reason per client generation with a 64-key cap (`DROPPED_NOTIFICATION_LOG_KEY_LIMIT`). The owner/mirror split reads the existing `SshPtyConsumerSessionAdapter.deliveryMode` grant, so it cannot grant a client any capability it did not already hold.

Availability is a different matter: the open findings are a denial-of-service-shaped failure mode (one non-reading local or remote peer stalling every terminal on the host), which is part of why this is a draft.

## Honest limitations

- **Everything here was executed on macOS only** (darwin, Node + vitest). No Windows, Linux or WSL run, and no real SSH host. The reasoning that the change is platform-neutral is above; it is reasoning, not evidence.
- I could not put a real 268 ms SSH link, a Windows client or the reporter's VPS under test. What is proven here is that the close-instead-of-backpressure class survived in these two host-side paths on `main` and no longer does for a session owner. The exact string the reporter quoted was already fixed by #11999 before this PR, so I cannot prove these two paths were the *only* remaining trigger for their session.
- The reporter has not confirmed a fix; there is no end-to-end validation against the environment in #12041.
- Problem 1 of #12041 (prompt input auto-indent) is untouched and un-triaged.
- Trade-off 1 — a stalled `legacy-owner` gating other PTYs through the relay-wide `legacyRetentionBelowLowWater` aggregate — is pre-existing on `main` and is not fixed here.
- The wedge states in open findings **B1**/**B2** are reasoned from the code paths and are not covered by any test on this branch. A reviewer reproduced the shed-revision wedge with a scratch test; the residual conditional-eviction gap has not been turned into a committed test.

## AI Review Report

Reviewed by independent subagents in a loop across two hostile lenses — **correctness/regressions** and **cross-platform + remote + performance**.

- **Final verdict: OUTSTANDING FINDINGS — DRAFT, NOT READY.** 3 blocking findings open, 1 non-blocking note recorded.

Round 1 raised two findings, both confirmed and both addressed:

- `src/relay/dispatcher-projection-stall.ts` — *the stall window is a single global one-shot timer, and the `stalled` probe is true for any sink with an un-settled frame in flight, so a healthy peer could be judged over ~1 ms and closed.* **Confirmed.** The file was deleted; the projection path owns no timer. (Deleting it without a bounded replacement is what left findings **B1**–**B3** open.)
- `src/relay/dispatcher.ts` — *the all-or-nothing projection lets a single saturated non-owner client pause the native PTY for every viewer, reverting #11005's per-client isolation.* **Confirmed.** Addressed by the `owner` / `mirror` split: only a session owner can back-pressure.

Round 2 declared the branch clean. **That verdict was wrong**, and round 3 found why:

- `src/relay/dispatcher.ts` — *the round-2 fix for the second finding shed a mirror's frame instead of closing it, and the stall reaper had been deleted, so nothing evicted a saturated mirror at all. Its retained ledger leases pinned `legacyRetentionBelowLowWater` to `false`, `PtyHandler.maybeResumePtyOutput` early-returned, and every paused PTY on the host stayed paused forever.* **Confirmed by reproduction** against both revisions (the pre-PR code recovers because it closes the mirror; the shed revision never recovers). Addressed by restoring the close for mirrors only — see "Why a mirror is closed rather than shed" — and locked in by a regression test that asserts the relay-wide gate clears. **The restoration is only partial; see B1/B2.**
- `src/relay/dispatcher.ts` — *the dropped-`pty.replay` diagnostic was classified against `producerFrameCapacity`, but the rejection comes from the control lane, so the one line an operator gets for silent scrollback loss named the wrong limit.* **Confirmed and fixed.** `logDroppedProducerNotification` now takes the lane and prints `control queue full`.

### Open blocking findings (why this is a draft)

- **B1 — a saturated mirror is only evicted if another frame is projected to it.** The mirror loop that closes it (`src/relay/dispatcher.ts:1185`) runs only inside a live projection. Once that mirror's retained leases hold `legacyRetentionBelowLowWater` false, `PtyHandler.maybeResumePtyOutput` (`src/relay/pty-handler.ts:1172`) early-returns and line 1184 — the only `pty.resume()` in the handler — never fires, so every paused PTY produces nothing, so no projection is generated, so the client is never evicted. Self-sustaining wedge of every PTY for every viewer.
- **B2 — the owner back-pressure path returns before the mirror loop.** `src/relay/dispatcher.ts:1177` returns false as soon as any owner cannot admit the frame, so a mirror saturated via a non-projection path (`tryNotifyPtyDataToClient`, agent-hook notifies, responses) is not visited and not closed while an owner is back-pressured — and its retained bytes are exactly what keeps the relay-wide gate closed for that owner's PTY too.
- **B3 — a stalled `legacy-owner` is never evicted at all, and gates the whole relay.** Trade-off 1. A peer that stops reading without erroring never reaches `handleWriteSettlement`, so nothing bounds how long it holds `legacyRetentionBelowLowWater` false for every other PTY on the host. Pre-existing on `main` for `tryNotifyPtyData`, but this PR extends the same all-or-nothing contract to the projection path, so it is now reachable from more places.

The shape the reviewer proposes: a saturated non-owner must be evicted, or have its leases released, on a **bounded** timeline so the relay-wide gate can clear — or the gate must be narrowed from a relay-wide aggregate to per-PTY scope. Keeping the pre-PR close for non-owner clients plus a narrower fix aimed at #12041's actual trigger is probably both smaller and safer than the current design. That is a maintainer's call on relay backpressure semantics, not another remediation round.

### Non-blocking note

- CodeRabbit pre-merge check: docstring coverage 20.00% against an 80.00% threshold. Not blocking — the new exported `LegacyProjectionRole` and `legacyProjectionRole` both carry doc comments; the shortfall is in the untouched neighbours the check samples.


nwparker:
## Converting to draft — not ready to merge

This targets a real, reproduced bug (#12041), but after **three adversarial review rounds** it still carries blocking findings. Each round's fix has surfaced another, and the current shape may be worse than the bug it fixes.

**CI is green (41 checks).** The problem is behavioural, not a failing test.

### The core objection

The change stops a saturated non-owner client from being closed, and the stall reaper was removed — so nothing evicts it. But its retained bytes still count toward the relay-wide gate `legacyRetentionBelowLowWater`, so `notifyLegacyCapacity(false)` never fires and `PtyHandler.maybeResumePtyOutput` early-returns. Line 1184 is the **only** `pty.resume()` call in the handler.

Net effect: one permanently stalled subscriber can wedge **every** PTY for **every** viewer, indefinitely. Before this change the projection closed that client, which released the gate and let the relay self-heal within one frame. A reviewer reproduced this with a scratch test on the branch.

Trading "one slow client disconnects sessions" for "one slow client silently wedges all terminals" is not a good trade, and it is the kind of regression that would be very hard to diagnose in the field.

### What I think the right shape is

A saturated non-owner client must still be evicted, or have its leases released, on a **bounded** timeline so the relay-wide gate can clear. Keeping the pre-PR close for non-owner clients, plus a narrower fix aimed at #12041's actual trigger, is probably both smaller and safer than the current design.

The reproduction and diagnosis are solid and worth keeping. I'm marking this draft rather than iterating further because the remaining work is a design decision about relay backpressure semantics that deserves a maintainer's call, not another remediation round.

PR #12221, opened 2026-08-03
Saving and editing files now works in a per-workspace environment

### LINKED ISSUE #11762: [Bug] File editing fails when using per-workspace environment
<kbd>[bonsaielectric](https://discord.com/channels/1492228674081263786/1532710991471710271/1532710991471710271)</kbd> [bonsaielectric](https://discord.com/channels/1492228674081263786/1532710991471710271/1532710991471710271)  `via Discord`

https://github.com/stablyai/orca/issues/10946

nwparker:
## Summary

A per-workspace environment recipe that returns `connection.type: "ssh"` registers a **runtime-owned** SSH target id `runtime-ssh-<runtimeId>` (`src/main/ssh/ssh-connection-store.ts:55`, `src/main/ephemeral-vm-runtime-ssh.ts:29`). Those ids are deliberately hidden from the renderer: `broadcastSshState()` returns early for them at `src/main/ipc/ssh.ts:343` and `SshConnectionStore.listTargets()` filters them out, so `ssh:state-changed` never fires and the renderer's `sshConnectionStates` map never receives their `connectionGeneration`.

Four copies of the same lookup — two named `getExpectedSshConnectionGeneration` helpers plus two inline duplicates — read exactly that map and fail closed on `undefined`, so every mutation in a recipe-VM workspace was rejected before it left the renderer:

- `src/renderer/src/lib/editor-file-operation-owner.ts:186` → `Couldn't verify which host owns this file. Reopen the file after the connection settles.` (Cmd-S and autosave both funnel through `getEditorFileOperationContext` via `src/renderer/src/components/editor/editor-autosave-controller.ts:104`)
- `src/renderer/src/components/right-sidebar/file-explorer-operation-owner.ts:177` → `Couldn't determine which host owns this workspace. Check the connection and try again.` (file-tree create/rename/delete/drop)
- `src/renderer/src/lib/ssh-mutation-expectation.ts` — the same lookup inlined twice more (`captureDirectSshMutationExpectation`, `captureWorktreeSshMutationExpectation`), covering terminal file drop, global file drop, native-chat attachment upload, composer attachments, rich-markdown image insert and MCP config writes.

Reads were unaffected (`captureEditorFileOperationProvenance` does not enforce the generation), which is why opening a file worked and only editing broke.

The gate cannot simply be dropped: `assertSshMutationExpectation` (`src/main/ssh/ssh-connection-generation.ts:96`) rejects any SSH-routed mutation whose `expectedGeneration` is `undefined`. The renderer genuinely needs the real value.

**What now happens.** main publishes an authority-only payload for runtime-owned targets on a dedicated channel, inside the existing early-return branch so nothing about the `ssh:state-changed` suppression changes:

- `src/main/ipc/ssh.ts:348` — sends `ssh:runtime-owned-authority-changed` with `{ targetId, connectionGeneration }`, using `getSshConnectionGeneration(targetId)`: verbatim the value `assertSshMutationExpectation` compares against. Deliberately **not** `getSshProviderAuthority`, which mints and stores a fresh `providerEpoch` on a miss (`src/main/ssh/ssh-provider-authority.ts:113`) and would create authority rows from a pure-notification path.
- `src/main/ipc/ssh.ts:1479` — a `ssh:listRuntimeOwnedAuthorities` pull handler for a renderer reload or a VM that connected before the window subscribed. It is listed in `SSH_IPC_CHANNELS` so the existing `ipcMain.removeHandler` loops keep re-registration clean; the push channel is intentionally not listed.

The renderer keeps the value in its own `runtimeOwnedSshConnectionGenerations` map (`src/renderer/src/store/slices/ssh.ts:37`), **never** in `sshConnectionStates` — that map is the user-facing host registry (`SshStatusSegment.tsx:192`, `App.tsx:998`, `WorktreeJumpPalette.tsx:1821`, `WorktreeCard.tsx:353`), and every one of those surfaces assumes runtime-owned ids never appear in it. Subscription is registered before the pull is issued, and the pull only fills keys the push has not already set, so a rotation landing mid-hydration is never overwritten by a stale snapshot (`src/renderer/src/hooks/useIpcEvents.ts:2664`). Entries are removed by the existing `buildRemovedSshTargetCleanupPatch` (`src/renderer/src/store/slices/ssh-target-cleanup.ts:211`) — but see *Trade-offs*: that patch is not actually reachable for a runtime-owned id today, so the renderer keeps a torn-down VM's generation until reload.

The four duplicated lookups collapse into one shared pair in `src/renderer/src/lib/ssh-mutation-expectation.ts`. The new-map fallback is reachable **only** when there is no `runtimeEnvironmentId`, so a `runtime-ssh-*` id reached through a paired HUB can never borrow the local authority. Fail-closed is preserved everywhere: an unpublished authority still throws, and a relay rotation between capture and assert still invalidates the operation.

Closes #11762 for a per-workspace environment reached directly from the desktop app (the reported case). The variant hosted *inside* a remote runtime environment (HUB) still fails closed and is **not** fixed here — see *Known remaining gap* under Trade-offs.

## ELI5

Orca can spin up a little cloud machine just for one workspace. For safety, before Orca writes a file to a remote machine it checks a small numbered ticket that proves it is still talking to the same live machine. These recipe machines were hidden from the part of the app that hands out those tickets, so the app had no ticket, refused to write, and showed "Couldn't verify which host owns this file" every time you hit save. Now the hidden machines quietly hand their ticket to the app on a private channel, so saving and file-tree edits work. The machine still stays hidden from your host list, and if the ticket is missing or out of date Orca still refuses to write rather than guessing.

## Fix proof

Before the fix — the repro as it was first written, in a scratch file (`src/renderer/src/lib/editor-file-operation-owner.bug-11762.test.ts`). That file is **not** part of this diff: its cases were folded into the permanent suite before the PR was opened, so the block below is a historical transcript, not a suite you can re-run on this branch:

```
$ npx vitest run --config config/vitest.config.ts src/renderer/src/lib/editor-file-operation-owner.bug-11762.test.ts

 RUN  v4.1.5 /Users/nwparker/projects/orca/.claude/worktrees/wf_60c051cd-fbd-37

 ❯ src/renderer/src/lib/editor-file-operation-owner.bug-11762.test.ts (3 tests | 2 failed) 8ms
     × routes the save to the runtime-owned SSH host instead of failing closed 3ms
     × also blocks every file-tree mutation in the same workspace 3ms

 FAIL  src/renderer/src/lib/editor-file-operation-owner.bug-11762.test.ts > #11762 editing a file in a recipe-VM (per-workspace environment) workspace > routes the save to the runtime-owned SSH host instead of failing closed
Error: Couldn't verify which host owns this file. Reopen the file after the connection settles.
 ❯ getEditorFileOperationContext src/renderer/src/lib/editor-file-operation-owner.ts:186:11
    184|   if (host?.kind === 'ssh' && provenance.expectedSshConnectionGenerati…
    185|     // Why: an old/partial SSH publication may be readable but cannot …
    186|     throw new Error(OWNER_CHANGED_MESSAGE)
       |           ^
    187|   }
    188|   return {
 ❯ src/renderer/src/lib/editor-file-operation-owner.bug-11762.test.ts:58:21

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯

 FAIL  src/renderer/src/lib/editor-file-operation-owner.bug-11762.test.ts > #11762 editing a file in a recipe-VM (per-workspace environment) workspace > also blocks every file-tree mutation in the same workspace
AssertionError: expected [Function] to not throw an error but 'Error: Couldn\'t determine which host…' was thrown

- Expected:
undefined

+ Received:
"Error: Couldn't determine which host owns this workspace. Check the connection and try again."

 ❯ src/renderer/src/lib/editor-file-operation-owner.bug-11762.test.ts:72:76
     70|     expect(owner).toEqual({ kind: 'ssh', connectionId: runtimeOwnedTar…
     71|
     72|     expect(() => captureFileExplorerOperationGuard(worktreeId, owner))…
       |                                                                            ^
     73|   })
     74|

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/2]⎯


 Test Files  1 failed (1)
   Start at  21:59:05
   Duration  2.05s (transform 1.34s, setup 0ms, import 1.97s, tests 8ms, environment 0ms)
```

After the fix — the repro cases now live in the permanent `editor-file-operation-owner.test.ts` under a `runtime-owned SSH host (per-workspace environment)` block, alongside the new negative cases (unknown authority still throws, rotation between capture and assert still throws, HUB-scoped route never borrows the local authority):

```
$ npx vitest run --config config/vitest.config.ts src/renderer/src/lib/editor-file-operation-owner.test.ts

 RUN  v4.1.5 /Users/nwparker/projects/orca/.claude/worktrees/wf_60c051cd-fbd-37


 Test Files  1 passed (1)
      Tests  13 passed (13)
   Start at  22:05:06
   Duration  3.63s (transform 2.75s, setup 0ms, import 3.52s, tests 13ms, environment 0ms)
```

New coverage across the rest of the change:

```
    src/renderer/src/lib/ssh-mutation-expectation.test.ts \
    src/renderer/src/lib/editor-file-operation-owner.test.ts \
    src/renderer/src/store/slices/ssh.test.ts \
    src/main/ipc/ssh.test.ts \
    src/preload/ssh-authority-forwarding.test.ts \
    src/shared/ssh-retained-payload-admission.test.ts

 Test Files  6 passed (6)
      Tests  115 passed (115)
   Start at  02:05:20
   Duration  4.33s (transform 7.74s, setup 0ms, import 10.00s, tests 917ms, environment 0ms)
```

## No regressions

Every pre-existing suite neighbouring a touched file — `src/main/ipc/ssh.test.ts`, all of `src/main/ssh/`, all of `src/preload/`, `src/shared/ssh-retained-payload-admission.test.ts`, `src/renderer/src/lib/editor-file-operation-owner.test.ts`, `src/renderer/src/lib/ssh-mutation-expectation.test.ts`, all of `src/renderer/src/store/slices/` and all of `src/renderer/src/components/right-sidebar/`:

```
$ npx vitest run --config config/vitest.config.ts src/main/ipc/ssh.test.ts src/main/ssh/ src/preload/ \
    src/shared/ssh-retained-payload-admission.test.ts \
    src/renderer/src/lib/editor-file-operation-owner.test.ts \
    src/renderer/src/lib/ssh-mutation-expectation.test.ts \
    src/renderer/src/store/slices/ src/renderer/src/components/right-sidebar/

 RUN  v4.1.5 /Users/nwparker/projects/orca/.claude/worktrees/wf_60c051cd-fbd-37

[ssh-relay] GC: lock at /home/u/.orca-remote/relay-0.1.0+aaa/.install-lock is stale; treating as recoverable
[ssh-relay] GC: lock at /home/u/.orca-remote/relay-0.1.0+aaa/.install-lock is stale; treating as recoverable

 Test Files  424 passed | 2 skipped (426)
      Tests  5182 passed | 13 skipped (5195)
   Start at  22:13:52
   Duration  44.91s (transform 93.77s, setup 0ms, import 509.02s, tests 66.48s, environment 12.50s)
```

The runtime-owned leak guards — the suites that assert these ids never reach a user-facing surface — plus the authority/generation suites:

```
$ npx vitest run --config config/vitest.config.ts active-worktree-focus-after-delete \
    ssh-workspace-forget-resolution workspace-session-liveness new-workspace-ssh-gate \
    ssh-pane-connect-gate workspace-session-reconnect-targets ssh-provider-authority \
    ssh-connection-generation

 Test Files  7 passed (7)
      Tests  52 passed (52)
   Start at  22:09:04
   Duration  296ms (transform 342ms, setup 0ms, import 504ms, tests 20ms, environment 0ms)
```

`useIpcEvents` is itself a touched file (subscription + pull, plus its harness mocks), and its suite is not inside any directory above, so it runs separately:

```
$ npx vitest run --config config/vitest.config.ts src/renderer/src/hooks/useIpcEvents.test.ts

 Test Files  1 passed (1)
      Tests  96 passed (96)
   Start at  02:06:13
   Duration  3.68s (transform 1.48s, setup 0ms, import 1.97s, tests 1.64s, environment 0ms)
```

Max-lines ratchet:

```
$ node config/scripts/check-max-lines-ratchet.mjs
max-lines ratchet OK — 352 grandfathered suppression(s), no new bypasses.
```

### Trade-offs

There is one honest behavior change worth naming. `activeSessions.set` happens before `connectionManager.connect()` completes, and `deploying-relay` also broadcasts, so both the push and the pull can publish an authority for a VM whose relay is not yet up. The guarantee is "no authority is published for a target with no session", not "the relay is live". In that narrow window a write to a recipe VM that would previously have been stopped by the renderer with *"Reopen the file after the connection settles"* now gets one layer deeper and fails with `requireSshFilesystemProvider`'s error instead (`src/main/ipc/filesystem.ts:828`). That is a different message for the same failure, not a new failure, and it matches how ordinary SSH hosts already behave during `deploying-relay`.

A second, smaller honest correction: the renderer's revocation path does **not** fire on VM teardown. `buildRemovedSshTargetCleanupPatch` does delete the new map's entry, and a unit test covers that, but its only production entry point is `clearRemovedSshTargetState`, whose three call sites are `HostRemoveDialog.tsx:83`, `SshPane.tsx:171` and `useIpcEvents.ts:2915`. The first two are user-facing host-list removals — runtime-owned ids are filtered out of `listTargets()` and never appear there — and the third is the deletion guard inside the `ssh:state-changed` handler, which by construction never runs for a runtime-owned id. So after a recipe VM is torn down the renderer keeps the last-published generation until the next reload. The consequence is bounded and still fail-closed at the trusted boundary: main keeps its own generation entry too, so `assertSshMutationExpectation` also passes, and the write then fails in main at `requireSshFilesystemProvider` (`src/main/ipc/filesystem.ts:828`) because the provider is gone. No write is routed to a wrong host; the user sees main's provider-unavailable error instead of the renderer's. Wiring a real revocation into the runtime teardown path is left as follow-up.

Known remaining gap, deliberately out of scope and covered by an explicit negative test: a per-workspace environment hosted *inside* a remote runtime environment (HUB) routes through `sshStateByEnvironment`, which is fed by `notifySshStateChanged` and `hydrateRuntimeEnvironmentSshState` — both also below the runtime-owned early return. That case still fails closed. Closing it needs the runtime RPC surface and the paired-client event protocol extended. The `!runtimeEnvironmentId` guard on the fallback is what keeps that gap *safe* rather than merely open, and the shared helper is shaped so the HUB variant is one added branch later.

## Cross-platform

No platform-conditional code, no path construction, no keyboard shortcut, accelerator or shortcut label, and no new git subcommand or option — so the Git 2.25 baseline and `GitCapabilityCache` rules do not apply here. This is pure IPC plumbing across main / preload / renderer and behaves identically on **macOS, Linux and Windows**. The new payload carries no path or platform data, so Windows path separators and case-folding are out of scope; `remotePlatform` continues to ride only the ordinary `ssh:state-changed` payload built by `withSshRemotePlatform`. The VM guest side is untouched.

**SSH-remote** is the primary case: the recipe VM is reached over SSH, and the change restores writes to it while keeping the fail-closed contract for a target whose authority is genuinely unknown. Nothing assumes local execution.

Verification scope: the evidence in this PR is automated-test evidence, all of it run on **macOS**. No live recipe VM was exercised on Linux or Windows, and no manual live-VM run is included here — the cross-platform argument above is a code argument (no platform branches, no paths, no shortcuts on this path), not an observed one.

**Folder workspaces (non-git)** need no extra code. `getExecutionHostIdForFolderWorkspace` resolves a folder workspace on a recipe VM to the same `ssh:runtime-ssh-*` host with `runtimeEnvironmentId: null`, and both `getFileExplorerGenerationRoute` and `resolveCurrentEditorRoute`'s folder re-resolution reach the same shared helper. There is no git-worktree assumption anywhere on this path.

Provider-agnostic: nothing here is GitHub- or GitLab-specific.

## Performance

One extra IPC message per state transition of a *runtime-owned* target only — a two-field payload replacing a message that is currently suppressed entirely. Ordinary SSH targets get zero additional traffic and zero additional work, since the new `send` sits inside the existing early-return branch. One boot-time `invoke` iterates `activeSessions` (single-digit entries). No polling, no timers, no per-component subscriptions.

The gate lookup adds one `startsWith` and one `Map.get` on an already-synchronous path, and only when the primary map read misses. Both store setters return the previous state object when the value is unchanged, so a push for an unchanged generation causes no zustand notification and no re-render; nothing selects the new map for rendering — it is read imperatively via `getState()` inside the gates.

## Security

This is a write-authorization surface, so the details matter.

- **IPC.** One new push channel and one new invoke handler, both scoped to ids matching `runtime-ssh-`. The pull handler enumerates only active sessions, so it cannot mint an authority for a target main has no session for. `admitRuntimeOwnedSshAuthority` in `src/shared/ssh-retained-payload-admission.ts` validates every inbound payload in the preload boundary: it requires a bounded-length `targetId` that is genuinely runtime-owned and a non-negative safe integer generation, and drops anything else. It rejects a well-formed authority for an *ordinary* target id, so this channel cannot be used to forge a generation for a user-facing host. No upper bound is imposed on the generation because the session base is `randomBytes(5) << 13` and legitimately runs near `Number.MAX_SAFE_INTEGER`.
- **Authority scoping.** The new map is consulted only when there is no `runtimeEnvironmentId`, so a HUB-proxied `runtime-ssh-*` id can never borrow a locally-published generation. A dedicated test asserts this and asserts the HUB's own publication is what authorizes that route.
- **Fail-closed preserved.** An unknown authority still throws; a generation that changes between capture and assert still throws; main's `assertSshMutationExpectation` is unchanged and remains the authoritative check on the trusted side. A destroyed or suspended VM is *not* revoked in the renderer today (see Trade-offs): its stale generation still passes both the renderer gate and main's assert, and the write is stopped one layer deeper by `requireSshFilesystemProvider` instead. Nothing is routed to a different host — the id is unique per runtime, and a reconnect rotates the generation and re-publishes it.
- **`getSshConnectionGeneration`, not `getSshProviderAuthority`.** The latter *mints and stores* a fresh `providerEpoch` on a cache miss; calling it from a notification path would create authority rows for targets main never created. The published value is the exact one main compares against.
- **No change** to command execution, path handling, credential handling or the auth surface. No `providerEpoch` or connection status is exposed on the new channel — only the generation integer.

## AI Review Report

- **Review loop count:** 1 round(s)
- **Final verdict:** CLEAN
- The final round returned **zero blocker/major findings** from both lenses.

**Non-blocking notes (2), recorded for transparency:**

1. `src/renderer/src/store/slices/ssh-target-cleanup.ts` — the PR body originally claimed "Tearing the VM down deletes the entry via the existing `buildRemovedSshTargetCleanupPatch`" and "a destroyed or suspended VM stops authorizing writes". That revocation path is not reachable for runtime-owned ids: `clearRemovedSshTargetState` has exactly three production call sites — `HostRemoveDialog.tsx:83`, `SshPane.tsx:171` and `useIpcEvents.ts:2915` — and none of them fires for a `runtime-ssh-*` target. The Summary, Trade-offs and Security sections above have been corrected to state what actually happens (stale entry until reload; the write is stopped in main by `requireSshFilesystemProvider`).
2. `src/renderer/src/store/slices/ssh-target-cleanup.ts:211`, `:242`, `:265` — following from the above, the new deletion branch for `runtimeOwnedSshConnectionGenerations` is effectively dead code for the ids it exists to clean up. It is kept deliberately: it is the correct behavior once a runtime-teardown caller is wired up, it is covered by a test, and it costs one `Map` copy on a path that only runs when a user-facing host is removed. Judged non-blocking because the fail-closed contract is still enforced on the trusted side.


nwparker:
@AmethystLiang give this to you

jradic-siio:
Still reproducible on 1.4.195 (macOS, arm64), with a `provisioned-root` + direct-SSH recipe backed by local Docker.

**Symptoms:** terminals, agents, git and SFTP over the same target all work perfectly. Only renderer-initiated writes fail. Editor Cmd-S shows "Failed to save the file. Please try again."; the underlying throw is `Couldn't verify the SSH connection. Reconnect the host and try again.` File-tree mutations fail the same way. Reads are unaffected, exactly as this PR describes.

**It is not a regression.** I diffed the gate across releases to check whether a recent update caused it:

- `src/renderer/src/lib/editor-file-operation-owner.ts`
- `src/renderer/src/lib/ssh-mutation-expectation.ts`
- `src/renderer/src/components/right-sidebar/file-explorer-operation-owner.ts`
- `src/main/ssh/ssh-connection-generation.ts`

All four are byte-identical from v1.4.186 through v1.4.195, and the suppression that starves the renderer of the generation (`ssh-renderer-broadcast.ts:32`, previously `ipc/ssh.ts:414`) is present throughout. So the path has never worked for a recipe-created workspace; it only becomes visible once you try to save in the editor rather than in a terminal.

**A second surface with the same root cause.** `right-sidebar/ssh-ports-panel.tsx` (split out of `PortsPanel.tsx` in #17194, logic unchanged) computes:

```ts
const isDisconnected = activeConnectionId
  ? sshConnectionStates.get(activeConnectionId)?.status !== 'connected'
  : true
```

Because `ssh:state-changed` is never broadcast for `runtime-ssh-*` ids, this is permanently true for a recipe workspace, and the panel shows "SSH connection lost / Reconnecting..." while the relay is healthy and `sidebar_worktree_activate` breadcrumbs record `sshDisconnected: false`. Other surfaces already guard this with `isRuntimeOwnedSshTargetId` (`use-worktree-card-foundation.ts:144`, `terminal-pane-host-state.ts:32`, `SshStatusSegment.tsx:81`); the ports panel has no such guard in 1.4.192 or 1.4.195. That false banner is what sends users looking for a broken container, which cost me two days before I found #11762.

**On #11762:** it was closed by `orca-discord-issues[bot]` on 2026-08-30 with no fix and no linked commit, so it currently reads as resolved while the defect is live.

Could this PR get a review? Happy to test a build or provide traces from a reproducing setup.

Issue #12231, opened 2026-08-03
[Bug]: WSL bridge destroys every ASCII double quote in CLI arguments — silently, or by splitting the argument


logii:
### Operating system

Windows

### Orca version

1.4.164 (WSL2 / Ubuntu, Windows PowerShell 5.1.26100.8875)

### Details

**Short summary:** on WSL, `orca-wsl-bridge.ps1` forwards arguments to `orca.exe` with `& $OrcaLauncher @ForwardArgs`. Windows PowerShell 5.1 does not escape embedded `"` when it builds a native process command line, so **every ASCII double quote in every Orca CLI argument is destroyed** before `orca.exe` parses it. `orca terminal send --text` is where it hurts most, because that payload is usually agent-authored prose, JSON, or code.

This is a *different* defect from #12196 (that one is the `--for` / `-ForwardArgs` prefix collision, which happens one stage earlier, during parameter binding). Same file, different stage.

#### Two symptoms, one cause

Which one you get depends on **whether the quoted span contains whitespace** — not on payload length:

| `--text` payload | what `orca.exe` actually receives | result |
|---|---|---|
| `say "hello" now` | `say hello now` (one argument) | **`ok:true`** — quotes silently deleted, corrupted text delivered to the pty |
| `tell me "whats next"` | splits into `tell me whats` + `next` | exit 1, `Unknown command: ...` |
| `say “hello” now` (typographic quotes) | `say “hello” now` | arrives byte-identical |

The first row is the dangerous one: a well-formed call, a success envelope, and a corrupted side effect. Nothing in the response says the text was altered.

The split in row two happens because the quotes are emitted onto the child command line *as data*, so the child's Win32/CRT parser consumes them as grouping syntax and re-splits the argument at the whitespace they used to protect.

#### How to reproduce (standalone — no Orca install needed)

`child.ps1` — stands in for `orca.exe`, reports the argv it actually received:

```powershell
$i = 0
foreach ($a in $args) { "child[$i] = [$a]"; $i++ }
```

`parent.ps1` — the shipped bridge's parameter block and dispatch, verbatim:

```powershell
param(
  [Parameter(Mandatory=$true, Position=0)]
  [string]$OrcaLauncher,
  [string]$WslCwd,
  [Parameter(ValueFromRemainingArguments=$true)]
  [string[]]$ForwardArgs
)
$i = 0
foreach ($a in $ForwardArgs) { "bridge[$i] = [$a]"; $i++ }
[...]

PR #12234, opened 2026-08-03
fix(wsl-cli): stop the WSL bridge swallowing every command that contains --for

### LINKED ISSUE #12196: [Bug]: WSL bridge drops every command containing --for — PowerShell prefix-binds it to the script's own -ForwardArgs
### Operating system

Windows

### Orca version

1.4.164 (WSL2 / Ubuntu, Windows PowerShell 5.1.26100.8875 Desktop)

### Details

**Short summary:** Any Orca CLI command containing `--for` fails from WSL before it ever reaches `orca.exe`. `orca-wsl-bridge.ps1` declares a parameter named `$ForwardArgs`, and PowerShell prefix-matches `--for` onto it — which consumes the flag's value and stops remaining-argument collection, so the real subcommand has no positional slot left.

This breaks the **documented** wait idiom: `orca terminal wait --terminal <h> --for tui-idle` appears throughout the bundled `orchestration` and `orca-cli` skill guides (`orca skills get orchestration`), so an agent following the shipped guidance on WSL hits this immediately.

**What happened?**

```
$ orca-ide terminal wait --terminal <handle> --for tui-idle --timeout-ms 60000 --json
\\wsl.localhost\Ubuntu\home\<user>\.local\share\orca\orca-wsl-bridge.ps1 : A positional
parameter cannot be found that accepts argument 'terminal'.
    + CategoryInfo          : InvalidArgument: (:) [orca-wsl-bridge.ps1], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : PositionalParameterNotFound,orca-wsl-bridge.ps1
```

Exit code is 1, so it fails loudly — but the command never reaches Orca at all.

**Root cause**

`~/.local/share/orca/orca-wsl-bridge.ps1` begins:

```powershell
[CmdletBinding(PositionalBinding=$false)]
param(
  [Parameter(Mandatory=$true, Position=0)][string]$OrcaLauncher,
  [string]$WslCwd,
  [Parameter(ValueFromRemainingArguments=$true)][string[]]$ForwardArgs
)
```

PowerShell performs **prefix matching** on parameter names. `--for` is an unambiguous prefix of `-ForwardArgs`, so PowerShell binds it explicitly, takes `tui-idle` as its value, and stops collecting remaining arguments — leaving `terminal`, `wait`, etc. with nowhere to bind.

The collision set is therefore exactly *flags that are a prefix of* `OrcaLauncher`, `WslCwd`, or `ForwardArgs`. I tested the neighbours: `--format`, `--from`, `--files-modified`, `--worktree`, and `--fn` all forward correctly. **`--for` appears to be the only colliding flag in the current CLI surface.**

**How can we reproduce it?**

Standalone — no Orca required. Save as `repro.ps1`:

```powershell
[CmdletBinding(PositionalBinding=$false)]
param(
  [Parameter(Mandatory=$true, Position=0)][string]$OrcaLauncher,
  [string]$WslCwd,
  [Parameter(ValueFromRemainingArguments=$true)][string[]]$ForwardArgs
)
Write-Host "ForwardArgs = $($ForwardArgs -join ' | ')"
```

```powershell
# control - passes through fine
powershell.exe -NoProfile -File repro.ps1 "LAUNCHER" -WslCwd "C:\tmp" terminal wait --format json
#   ForwardArgs = terminal | wait | --format | json

# trigger - same error as the real bridge
powershell.exe -NoProfile -File repro.ps1 "LAUNCHER" -WslCwd "C:\tmp" terminal wait --for tui-idle
#   A positional parameter cannot be found that accepts argument 'terminal'.
```

**Workaround for users**

Use the `=` form, which is not PowerShell parameter syntax and so falls through untouched:

```bash
orca-ide terminal wait --terminal <handle> --for=tui-idle --timeout-ms 60000 --json
```

**Proposed fix (verified)**

Rename the three bridge parameters so no Orca flag can prefix-match them:

```powershell
[CmdletBinding(PositionalBinding=$false)]
param(
  [Parameter(Mandatory=$true, Position=0)][string]$OrcaBridgeLauncher,
  [string]$OrcaBridgeWslCwd,
  [Parameter(ValueFromRemainingArguments=$true)][string[]]$OrcaBridgeForwardArgs
)
```

With that change the same invocation forwards correctly:

```
ForwardArgs = terminal | wait | --for | tui-idle
```

The caller in `~/.local/bin/orca-ide` would need its `-WslCwd` argument renamed to match. More robustly, the launcher path and cwd could be passed via environment variables so the script has **zero** parameter surface and forwards `$args` verbatim — that removes the whole class rather than the current instance.

**Anything else that might help:** the same file is involved in #9498, though that is a distinct failure (duplicate `PATH`/`Path` env key) and does not reproduce on 1.4.164 here.

logii:
Fixes #12196.

## Summary

On WSL, every Orca CLI command containing `--for` died before reaching `orca.exe`:

```
orca-wsl-bridge.ps1 : A positional parameter cannot be found that accepts argument 'terminal'.
```

The bridge declares `[Parameter(ValueFromRemainingArguments=$true)] [string[]]$ForwardArgs`. PowerShell binds parameters by unique **name prefix**, so `--for` bound to `-ForwardArgs`, consumed `tui-idle` as its value, and stopped remainder collection — leaving the real subcommand with no positional slot. `--for=tui-idle` slipped through only because `=` is not PowerShell parameter syntax.

The user-visible effect: `orca terminal wait --terminal <handle> --for tui-idle` — the exact form printed by `orca skills get orchestration` — never worked on WSL. An agent following the shipped guide hit this on its first wait.

Renaming the parameter to `$OrcaForwardArgs` takes it out of the CLI's flag namespace. `--for` is the only collision in the current 213-flag surface (`--format`, `--from`, `--files-modified`, `--worktree` were all verified unaffected), so this is a two-token change to the generated script.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

New test `src/cli/wsl-bridge-flag-collision.test.ts` enforces the invariant rather than the instance: no flag in `COMMAND_SPECS` / `GLOBAL_FLAGS` / `BOOLEAN_FLAGS`, and no PowerShell common parameter, may be a prefix of any bridge parameter name. It **fails on the pre-fix script** with exactly

```
AssertionError: expected [ '--for binds to -ForwardArgs' ] to deeply equal []
```

and passes after. A second case asserts the launcher and the bridge still agree on the parameters they exchange, so a half-rename cannot pass. `config/tsconfig.cli.json` gains `src/main/cli/wsl-cli-scripts.ts` so the CLI project typechecks the import, following the existing pattern for main-process files reached from `src/cli`.

`pnpm test` on this branch: **44037 passed, 11 failed**. All 11 are unrelated to this change and were confirmed on a pristine checkout of the same commit — 9 pre-existing environment failures on Linux/WSL (`pty-subprocess`, `local-pty-provider`, `configure-process` GPU flags, `agent-exec-handler`, `managed-hook-timeout`), plus `orca-runtime-files.test.ts > rejects stale absolute terminal artifact previews`, which is **flaky**: run alone on this branch three times it passed, failed, passed.

### Verified against the real binary, not only the generated string

Rendered the patched builder's output and diffed it against the bridge shipped in 1.4.164 — byte-identical apart from the two renamed tokens. Then ran both scripts against the real `orca.exe` from WSL, same argument vector:

| bridge | `terminal wait --terminal bogus --for tui-idle --timeout-ms 1` |
|---|---|
| shipped 1.4.164 | `A positional parameter cannot be found that accepts argument 'terminal'` |
| patched | `terminal_handle_stale` — Orca's own error, so the argv arrived intact |

No regression on the rest of the bridge's contract: `terminal list --json` and `worktree list --json` return normally (cwd still resolves through `ORCA_CLI_CWD`), and an unknown subcommand still exits 1.

## AI Review Report

- **Cross-platform.** The change is confined to `buildWslBridgeScript()`, which only ever runs on Windows + WSL; macOS and Linux paths do not read it. No path handling, shortcut, or label is touched.
- **Upgrade / version skew.** Flagged as the main risk: could a user end up with a new bridge and an old launcher? No — `WslCliInstaller.install()` writes launcher and bridge in the same operation, and `status` reports `stale` when either file's content differs from the expected output, which triggers a reinstall of both. The renamed parameter is also never named by the launcher (it is the remainder-args parameter), so even a skewed pair would keep binding `-WslCwd` correctly.
- **Scope.** `-WslCwd` and `-OrcaLauncher` were deliberately left alone: neither collides with any current flag, and renaming `-WslCwd` would require changing the launcher's call site for no present defect. The new test now guards them too, so a future flag that would collide fails CI instead of shipping.
- **SSH / remote / local.** No change — the WSL bridge is a local-interop path only.
- **Agent / integration compatibility.** Provider-neutral; this is transport plumbing beneath every agent and integration. It restores the documented `--for` form for all of them.
- **Performance.** None; identical work, one different identifier.
- **UI.** Not applicable.

## Security Audit

- **Command execution / input handling.** The patch does not change how arguments are parsed, quoted, or forwarded — it renames the parameter that collects them. No new interpolation, no new shell invocation, no change to `Push-Location`/`& $OrcaLauncher` semantics or to exit-code propagation.
- **Path handling.** Untouched (`$OrcaLauncher`, `$WslCwd` unchanged).
- **Secrets / auth / IPC / dependencies.** Not touched; no new dependency.
- **Follow-up, separate defect.** While isolating this one I found and filed #12231: the same bridge destroys every ASCII `"` in CLI arguments, because Windows PowerShell 5.1 does not escape embedded quotes when building a native command line. It is silent when the quoted span has no whitespace and splits the argument when it does. That is a distinct root cause at a later stage and is deliberately **not** in this PR.

## Notes

- Windows/WSL-specific. Reproduced and verified on Windows 11 + WSL2 (Ubuntu), Windows PowerShell 5.1.26100.8875, Orca 1.4.164.
- Users pick up the fix when the installer rewrites the managed scripts; existing installs are detected as `stale` and reinstalled.
- The `--for=tui-idle` equals-form workaround stays valid, so nothing that adopted it breaks.

### DIFF
--- a/config/tsconfig.cli.json
+++ b/config/tsconfig.cli.json
@@ -73,6 +73,7 @@
     "../src/main/rolling-file-backup.ts",
     "../src/main/startup/hydrate-shell-path.ts",
     "../src/main/runtime/runtime-metadata.ts",
+    "../src/main/cli/wsl-cli-scripts.ts",
     "../src/main/win32-utils.ts"
   ],
   "compilerOptions": {
--- a/src/cli/wsl-bridge-flag-collision.test.ts
+++ b/src/cli/wsl-bridge-flag-collision.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest'
+
+import { buildWslBridgeScript, buildWslLauncher } from '../main/cli/wsl-cli-scripts'
+import { BOOLEAN_FLAGS, GLOBAL_FLAGS } from './args'
+import { COMMAND_SPECS } from './specs'
+
+// Why: PowerShell binds parameters by unique name prefix, so a bridge parameter whose name
+// starts with a CLI flag swallows that flag and every argument after it.
+const POWERSHELL_COMMON_PARAMETERS = [
+  'Debug',
+  'ErrorAction',
+  'ErrorVariable',
+  'InformationAction',
+  'InformationVariable',
+  'OutBuffer',
+  'OutVariable',
+  'PipelineVariable',
+  'ProgressAction',
+  'Verbose',
+  'WarningAction',
+  'WarningVariable'
+]
+
+const WINDOWS_LAUNCHER_PATH =
+  'C:\\Users\\dev\\AppData\\Local\\Programs\\orca\\resources\\bin\\orca.exe'
+
+function bridgeParameterNames(): string[] {
+  return [...buildWslBridgeScript().matchAll(/^\s*\[string(?:\[\])?\]\$(\w+),?$/gm)].map(
+    (match) => match[1]
+  )
+}
+
+describe('WSL bridge parameters', () => {
+  it('cannot prefix-capture a CLI flag', () => {
+    const flags = new Set([
+      ...GLOBAL_FLAGS,
+      ...BOOLEAN_FLAGS,
+      ...COMMAND_SPECS.flatMap((spec) => spec.allowedFlags)
+    ])
+    const parameters = [...bridgeParameterNames(), ...POWERSHELL_COMMON_PARAMETERS]
+
+    const collisions = parameters.flatMap((parameter) =>
+      [...flags]
+        .filter((flag) => parameter.toLowerCase().startsWith(flag.toLowerCase()))
+        .map((flag) => `--${flag} binds to -${parameter}`)
+    )
+
+    expect(collisions).toEqual([])
+  })
+
+  it('declares every parameter the launcher passes it', () => {
+    const forwarded =
+      buildWslLauncher(WINDOWS_LAUNCHER_PATH).split('$ORCA_BRIDGE_PS1_WIN"')[1] ?? ''
+    const passed = [...forwarded.matchAll(/-([A-Za-z]\w*)/g)].map((match) => match[1])
+
+    expect(passed).not.toHaveLength(0)
+    expect(bridgeParameterNames()).toEqual(expect.arrayContaining(passed))
+  })
+})
--- a/src/main/cli/wsl-cli-installer.test.ts
+++ b/src/main/cli/wsl-cli-installer.test.ts
@@ -340,7 +340,7 @@ describe('WslCliInstaller', () => {
     expect(bridge).toContain('if ([string]::IsNullOrEmpty($WslCwd))')
     expect(bridge).toContain('$env:ORCA_CLI_CWD = $WslCwd')
     expect(bridge).toContain('Push-Location -LiteralPath (Split-Path -Parent $OrcaLauncher)')
-    expect(bridge).toContain('& $OrcaLauncher @ForwardArgs')
+    expect(bridge).toContain('& $OrcaLauncher @OrcaForwardArgs')
     const nullExitCodeBranch = bridge.indexOf('if ($null -eq $LASTEXITCODE)')
     const invocationFailureBranch = bridge.indexOf('if (-not $?)')
     expect(nullExitCodeBranch).toBeGreaterThan(-1)
--- a/src/main/cli/wsl-cli-scripts.ts
+++ b/src/main/cli/wsl-cli-scripts.ts
@@ -33,6 +33,8 @@ exec "$ORCA_POWERSHELL" -NoProfile -ExecutionPolicy Bypass -File "$ORCA_BRIDGE_P
 }
 
 export function buildWslBridgeScript(): string {
+  // Why: PowerShell binds parameters by name prefix, so a parameter named for a CLI flag
+  // eats it — -ForwardArgs swallowed `--for` and everything after it.
   return `${BRIDGE_MANAGED_MARKER}
 [CmdletBinding(PositionalBinding=$false)]
 param(
@@ -42,7 +44,7 @@ param(
   [string]$WslCwd,
 
   [Parameter(ValueFromRemainingArguments=$true)]
-  [string[]]$ForwardArgs
+  [string[]]$OrcaForwardArgs
 )
 
 $exitCode = 0
@@ -53,7 +55,7 @@ try {
     $env:ORCA_CLI_CWD = $WslCwd
   }
   Push-Location -LiteralPath (Split-Path -Parent $OrcaLauncher)
-  & $OrcaLauncher @ForwardArgs
+  & $OrcaLauncher @OrcaForwardArgs
   if ($null -eq $LASTEXITCODE) {
     if (-not $?) {
       $exitCode = 1


hjsg1010:
Heads up — I hit this same bug independently and, before finding this PR, had already built a fix along different lines. I've opened it as #12861 rather than sitting on it, but I want to be upfront that this PR got here first and your diagnosis in #12196 was exactly right.

While reproducing against Windows PowerShell 5.1.26100.8875 I found the collision reaches a bit further than `--for`:

| Forwarded flag | Behavior on current `main` |
| --- | --- |
| `--orcalauncher <path>` | rebinds `-OrcaLauncher`; the bridge executes that path instead of Orca |
| `--wslcwd <x>` | `ParameterAlreadyBound` |
| `--verbose`, `--debug`, `--erroraction`, `--ov` | swallowed by CmdletBinding common parameters, silently, no error |

The last row is the part I couldn't find a rename for — you can rename `$ForwardArgs`, but nothing can be renamed out of the way of PowerShell's built-in common parameters, so the guard test in this PR would go red with no available remedy if the CLI ever grows a `--verbose`.

So #12861 takes the other branch of the tradeoff: drop the `param()` block entirely and hand the bridge one base64 record, so no CLI token ever reaches the parameter binder. It's a bigger change to the install path, which is the obvious argument in favor of yours.

Genuinely no strong feelings about which lands — yours is smaller and fixes the reported break. If maintainers prefer this one, I'm happy to close #12861, and the extra cases above are easy to fold into your guard test as documentation of what the rename does and doesn't cover.

OrcaWin:
Closing as already fixed — your root-cause analysis was right, and it's worth recording that it was.

#12196 is now closed as completed. The fix landed via #13426 ("fix(wsl): preserve forwarded flags in PowerShell 5.1", merged 2026-08-11). It resolves exactly the prefix-binding problem you described, by removing the named parameter entirely rather than renaming or escaping it — with no `[Parameter(ValueFromRemainingArguments = $true)] $ForwardArgs` declared, PowerShell has nothing for `--for` to prefix-match against.

On `main` today (`src/main/cli/wsl-cli-scripts.ts`):

```powershell
[int]$ForwardArgStart = 1
...
[string[]]$ForwardArgs = @()
if ($args.Count -gt $ForwardArgStart) {
  $ForwardArgs = @($args[$ForwardArgStart..($args.Count - 1)])
}
```

Arguments are sliced positionally from `$args` and re-quoted via `ConvertTo-NativeCommandLineArgument`, so the real subcommand keeps its positional slot.

Note #12861 was opened for the same issue, so these two were duplicates of each other as well.

Please reopen if `--for` still fails on current `main`.

PR #12245, opened 2026-08-03
perf(runtime): withhold unchanged mobile snapshots from the graph payload

OrcaWin:
> **Builds on merged #12207.** This PR now targets `main`; the partition below relies on the reference-identical unchanged snapshots introduced there.

## Summary

Every renderer graph sync structured-cloned all of the user's worktree mobile-session snapshots to the main process, whether or not any had changed.

Measured on a synthetic profile matching the reported crash telemetry (222 worktrees, 787 tabs, 787 terminal layouts):

| | Payload | `structuredClone` |
| --- | ---: | ---: |
| Before, every sync | 374 KB | 5.08 ms |
| After, unchanged republish | 3.4 KB | 0.02 ms |
| After, one worktree changed | 5.3 KB | 0.04 ms |

Electron clones on serialize and again on deserialize, so the real cost is roughly double the figure above. This transport cost — not the renderer-side rebuild, which is ~1 ms after #12207 — is the bulk of a publication.

The renderer now sends only snapshots main has not acknowledged, and names the rest in a new `unchangedMobileSessionWorktrees` list. Main seeds `nextWorktrees` from that list so its prune treats withheld worktrees as live rather than removed.

**The IPC call itself stays unconditional.** `syncWindowGraph` is not a one-way publish: its return value is the only channel carrying `agentOrchestrationByPaneKey` to the renderer (`sync-runtime-graph.ts` holds the sole caller of `setRuntimeAgentOrchestrationByPaneKey`, and no event channel exists), and the handler adopts pre-allocated handles, merges detached leaves, refreshes writable flags, and drains graph-sync callbacks on every sync. Skipping the call when the graph looks unchanged would starve all of that — the renderer cannot know what main has to deliver. Only the payload shrinks.

### Two failure modes this design creates, both closed explicitly

1. **A publication that throws must not count as delivered.** The last-sent memo advances only after main acknowledges, so a failed publication is resent in full instead of being withheld forever as "already sent".
2. **A worktree main drops on its own** (`removeWorktreeMetadataAndHistory`) would otherwise never come back, because the renderer believes it is already delivered. Main now returns those in `mobileSessionResyncWorktrees`, which also clears the accepted-revision record so the republish is not rejected by the accept gate as a no-op.

## Screenshots

No visual change.

## Testing

- [x] `pnpm exec oxlint` on every changed file
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/runtime/` — 58 files / 564 tests pass
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/runtime/` — 213 files pass; 6 failures in `orca-runtime-files.test.ts` and `rpc/methods/ai-vault.test.ts` are pre-existing Windows path-separator assertions (`expected [ '\ctor\codex\home\sessions' ] to include '/ctor/codex/home/sessions'`), reproduced identically on the base commit with this change stashed
- [x] All out-of-directory importers of `sync-runtime-graph` — 9 files / 780 tests pass
- [x] Added tests, and verified each fails without the change

`pnpm test` (the wrapper) cannot run in this environment — `windows-native-registry` fails to rebuild — so vitest was invoked directly.

### New tests

`src/renderer/src/runtime/sync-runtime-graph-payload-partition.test.ts`
- withholds every unchanged worktree from a republish
- publishes only the worktree whose snapshot changed
- republishes a changed worktree whose publication threw before it was acknowledged
- resends everything when the first publication throws
- republishes worktrees main reports it dropped after acknowledging them
- drops a removed worktree from both the payload and the withheld list

`src/main/runtime/graph-sync-payload-partition.test.ts`
- keeps a withheld worktree alive instead of pruning it as removed
- still prunes a worktree the renderer omits from both lists
- asks for a republish of a withheld worktree it dropped, and accepts the resend
- leaves the legacy full-payload contract untouched when no list is sent

Negative controls: reverting the ack gate fails 3 of the renderer tests; reverting the `nextWorktrees` seed fails 2 of the main tests.

## AI Review Report

Reviewed the renderer/main contract for staleness risk, since a false "unchanged" silently publishes stale tabs to paired clients.

Checked and resolved:
- **Return-value starvation** — audited every field of `RuntimeSyncWindowGraphResult` and every side effect in the main handler before concluding the call must stay unconditional. This is why the PR shrinks the payload rather than skipping the sync.
- **Prune correctness** — the withheld list is folded into `nextWorktrees` before the prune loop, and a test asserts a worktree omitted from *both* lists is still pruned and still emits its removed frame.
- **Ack ordering** — the memo commit sits after `await`, inside `try`, so a rejection leaves it untouched.
- **Accept-gate interaction** — a resync clears `acceptedRendererMobileSnapshotByWorktree`, otherwise a republish at the same `(epoch, version)` is discarded and the worktree never recovers.
- **Version-skew** — renderer and main ship in one build, so there is no mixed-version path; a renderer reload resets the module memo and forces a full resend. The new fields are optional, and a caller that sends neither (`src/main/index.ts` headless sync) keeps its exact previous behavior, covered by a test.

Cross-platform: no platform-dependent behavior added — no paths, shell invocation, shortcuts, or labels. The change is plain data structures over existing IPC, identical on macOS, Linux, and Windows.

## Security Audit

No new IPC surface: two optional fields on an existing authenticated channel. `unchangedMobileSessionWorktrees` carries worktree ids that main already holds, and it can only ever *retain* an existing entry — it cannot inject, mutate, or resurrect snapshot content, so a malformed list cannot forge tabs. Unknown ids fall through to the resync path rather than creating state. `mobileSessionResyncWorktrees` only deletes renderer-local memo entries, whose worst case is a redundant full republish. No input parsing, command execution, path handling, auth, secrets, or dependency changes.

## Notes

- #12207 is merged into `main`; this follow-up now targets `main` directly and relies on its reference-identical unchanged snapshots.
- The renderer-side rebuild is not addressed here and does not need to be: it measures ~1 ms at this scale after #12207. A separately-proposed micro-optimization of the per-sync tab flatten was dropped as not worth a review cycle at 0.18 ms.

## Final merge-readiness review — 2026-08-03

Reviewed exact stacked range `27d470c2ac645cbcd4cf75c087ce6b3dc771a49d...5081a12ac2f044725939fa7eb63f22a461611658`. The final independent review returned **CLEAN**. GitHub reports a clean merge state and 42 completed checks with zero failures, including Windows packaging. The final focused run covered 18 files and passed 1,158 tests with one skip; typecheck, lint, format, max-lines, diff/reference cleanliness, and negative controls also passed.

### Meaningful fixes found during review

- `6e369e6c37`: withheld publications skipped same-revision preservation reconciliation. Removed serve/SSH tabs could remain stale, while headless hydration could mask a renderer snapshot dropped by main. The fix tracks accepted renderer membership and requests targeted resyncs; regressions cover both paths.
- `5081a12ac2`: accepted membership metadata used raw `snapshot.tabs` even when retirement fencing removed tabs before storage. Every later withheld publication could therefore request another full resync. Metadata now uses `fencedSnapshot.tabs`; a negative control proves two consecutive withheld publications no longer resync.

### Statistical benchmark

Identical SHA-256-verified uncommitted harness at baseline `3658165119`, #12207 at `27d470c2ac`, and the combined stack at `5081a12ac2`. Each target used three independently started, rotated rounds with 50 warmups and 150 measured samples per round: **450 measured samples per target per metric**. Values are `median / p95 [bootstrap 95% CI for median]`.

| Metric | Baseline | #12207 | Combined through #12245 |
| --- | ---: | ---: | ---: |
| Direct snapshot build | 1.382 / 2.053 ms [1.341, 1.421] | **0.135 / 0.189 ms [0.1348, 0.1360]** | 0.138 / 0.218 ms [0.1373, 0.1399] |
| Scheduled publication, excluding the deliberate 16 ms coalescing delay | 2.913 / 3.929 ms [2.860, 2.967] | 1.581 / 1.851 ms [1.575, 1.588] | **0.277 / 0.379 ms [0.2758, 0.2792]** |
| `structuredClone` | 1.369 / 1.873 ms [1.363, 1.381] | 1.329 / 1.436 ms [1.327, 1.332] | **0.0466 / 0.0529 ms [0.0465, 0.0468]** |
| Payload bytes | 933,388 | 933,386 | **49,631** |

The combined stack reduces median scheduled-publication time **90.5%** and payload size **94.7%** versus baseline. Compared with #12207 alone, this PR reduces scheduled time **82.5%**, clone time **96.5%**, and payload **94.7%**. Its 0.003 ms direct-build bookkeeping overhead is dwarfed by the transport reduction.

Every measured iteration changed exactly one status. Baseline and #12207 published all 222 snapshots; the combined stack published the changed worktree and withheld the other 221. Payload content preserved the changed status, and there were zero measured publication failures.

Raw local artifacts remain uncommitted in the dedicated benchmark worktree under `.benchmarks/mobile-session-publication/` (`report.md`, `summary.json`, `execution.json`, and `raw/*.json`). Absolute wall times should be treated cautiously because an unrelated two-day-old Node/Vite repro remained at approximately one logical CPU throughout; rotated commit order, independent processes, warmups, and confidence intervals reduce but cannot eliminate that host-load noise. The relative effects are large and consistent.

### Validation gaps and merge order

- #12207 merged first; #12245 now targets `main`.
- The saved `low spec windows` environment was unavailable: `remote_runtime_unavailable: Could not connect to the remote Orca runtime.` Windows packaging CI passed.
- `$electron`, `$perf`, and `$orca-mobile-emulator-qa` were unavailable. This is not a visual change, so no screenshots were fabricated and no prohibited OS-level UI automation was substituted.
- This PR remains a draft and was not merged.

### DIFF
--- a/src/main/runtime/graph-sync-payload-partition.test.ts
+++ b/src/main/runtime/graph-sync-payload-partition.test.ts
@@ -40,13 +40,27 @@ const storeBase = {
   })
 }
 
-function makeSession(): WorkspaceSessionState {
+function makeSession(overrides: Partial<WorkspaceSessionState> = {}): WorkspaceSessionState {
   return {
     activeRepoId: 'repo-1',
     activeWorktreeId: WT_A,
     activeTabId: null,
     tabsByWorktree: {},
-    terminalLayoutsByTabId: {}
+    terminalLayoutsByTabId: {},
+    ...overrides
+  }
+}
+
+function makeTerminalTab(id: string, ptyId: string) {
+  return {
+    id,
+    ptyId,
+    worktreeId: WT_A,
+    title: `Terminal ${id}`,
+    customTitle: null,
+    color: null,
+    sortOrder: 0,
+    createdAt: 1
   }
 }
 
@@ -65,6 +79,9 @@ function createRuntime() {
   })
   const events: RuntimeMobileSessionTabsResult[] = []
   runtime.onMobileSessionTabsChanged((snapshot) => events.push(snapshot))
+  const setSession = (next: WorkspaceSessionState): void => {
+    session = next
+  }
   const sync = (
     mobileSessionTabs: RuntimeMobileSessionTabsSnapshot[],
     unchangedMobileSessionWorktrees?: string[]
@@ -75,7 +92,7 @@ function createRuntime() {
       mobileSessionTabs,
       ...(unchangedMobileSessionWorktrees ? { unchangedMobileSessionWorktrees } : {})
     })
-  return { runtime, events, sync, internals: runtime as unknown as RuntimeInternals }
+  return { runtime, events, sync, setSession, internals: runtime as unknown as RuntimeInternals }
 }
 
 function makeSnapshot(worktree: string, version: number): RuntimeMobileSessionTabsSnapshot {
@@ -161,6 +178,57 @@ describe('graph-sync mobile payload partition', () => {
     expect(internals.mobileSessionTabsByWorktree.get(WT_B)?.tabs).toEqual([
       expect.objectContaining({ parentTabId: 'tab-1' })
     ])
+  })
+
+  it('asks for a republish when headless hydration masks a dropped renderer snapshot', () => {
+    const { sync, setSession, internals } = createRuntime()
+    setSession(
+      makeSession({
+        tabsByWorktree: { [WT_A]: [makeTerminalTab('serve-tab', 'serve-pty-1')] }
+      })
+    )
+    sync([makeSnapshot(WT_A, 1)])
+    internals.mobileSessionTabsByWorktree.delete(WT_A)
+
+    const result = sync([], [WT_A])
+    expect(result.mobileSessionResyncWorktrees).toEqual([WT_A])
+    expect(internals.mobileSessionTabsByWorktree.get(WT_A)?.publicationEpoch).toMatch(
+      /^headless-hydrated:/
+    )
+
+    const resend = sync([makeSnapshot(WT_A, 1)], [])
+    expect(resend.mobileSessionResyncWorktrees).toBeUndefined()
+    expect(
+      internals.mobileSessionTabsByWorktree
+        .get(WT_A)
+        ?.tabs.map((tab) => (tab.type === 'terminal' ? tab.parentTabId : tab.id))
+    ).toEqual(['tab-1', 'serve-tab'])
+  })
+
+  it('requests the accepted revision again to prune a stale preserved tab', () => {
+    const { sync, setSession, internals } = createRuntime()
+    setSession(
+      makeSession({
+        tabsByWorktree: { [WT_A]: [makeTerminalTab('serve-tab', 'serve-pty-1')] }
+      })
+    )
+    sync([makeSnapshot(WT_A, 1)])
+    expect(
+      internals.mobileSessionTabsByWorktree
+        .get(WT_A)
+        ?.tabs.some((tab) => tab.type === 'terminal' && tab.parentTabId === 'serve-tab')
+    ).toBe(true)
+
+    setSession(makeSession())
+    const result = sync([], [WT_A])
+    expect(result.mobileSessionResyncWorktrees).toEqual([WT_A])
+
+    sync([makeSnapshot(WT_A, 1)], [])
+    expect(
+      internals.mobileSessionTabsByWorktree
+        .get(WT_A)
+        ?.tabs.some((tab) => tab.type === 'terminal' && tab.parentTabId === 'serve-tab')
+    ).toBe(false)
   })
 
   it('leaves the legacy full-payload contract untouched when no list is sent', () => {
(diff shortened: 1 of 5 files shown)


brennanb2025:
## Post-merge Electron validation — 2026-08-03



- Target: squash merge ce8b778d3138cb905c395da51c625d9c34cf96e1, verified through the worktree-scoped merged dev identity.
- Visible check: the full Orca surface remained rendered and usable after two real unchanged graph publications; the page reported 0 console errors.
- Backing payload captured immediately before the real syncWindowGraph IPC: 0 changed snapshots and 4/4 worktrees withheld as unchanged. The partitioned graph was 1,981 bytes versus 48,655 bytes with the full snapshots, a 95.9% reduction.
- Batched structuredClone measurement on that live payload: median 0.003 ms partitioned versus 0.049 ms full; p95 0.005 ms versus 0.064 ms.
- Independent post-merge review returned CLEAN. Integrated checks passed: 13 partition tests, 13 terminal-retirement tests, node/web typechecks, diff-check, lint/format/max-lines gates, and adjacent headless/SSH coverage.

Validation gap: the offered Windows remote was not discoverable from either validation surface. The merged dev profile reported 0 hosts and no runtime environments; the installed runtime exposed one Linux SSH workspace and no Windows host. Windows packaging CI passed, but this run does not claim live Windows-remote validation. No mobile emulator run was warranted because the merged diff contains no mobile/ or mobile UI changes.

PR #12261, opened 2026-08-03
fix(terminal): return IME composition ownership to xterm

nwparker:
## Summary

- restore xterm-owned composition, preedit, and commit delivery
- refuse IME-owned events only at Orca application-action boundaries
- prevent stale PTY generations from receiving xterm data
- delete the parallel composition route, trackers, deferred newline state, native-text forwarding, and related platform machinery

This intentionally supersedes #12259: the held modifier set, modified-enter chord owner, and deferred newline sender no longer exist.

## VS Code parity

VS Code adds no terminal IME commit route. Its `attachCustomKeyEventHandler` handles application shortcuts and otherwise returns `true` so xterm owns input.

Local reference: `/Users/nwparker/projects/vscode/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts`.

VS Code uses xterm `6.1.0-beta.191`; Orca uses beta.287. Their stock `CompositionHelper` and `_inputEvent` composition behavior is identical apart from import-path changes. This PR now removes Orcas remaining keypress/composition xterm patch too. The only xterm patch left is the unrelated `SortedList` decoration cleanup.

## Size

Net change: 393 insertions, 7,772 deletions. No new production IME class, tracker, buffer, adapter, registry, route, or file.

## Evidence

- stock xterm back-to-back Hangul composition emits `가나`
- settled composition followed by physical Enter emits `한\r`
- ordinary typing plus Enter remains `ab\r`
- marked real Enter (`key: Enter`, `keyCode: 13`, `isComposing: true`) is refused before shortcut resolution
- replacement PTY generations reject stale xterm data
- the deterministic IBus replay now preserves the native task boundary required by xterms composition-position timer
- the workflow runs native IBus before deterministic replay so synthetic coverage cannot mask native failure

## Verification

- `pnpm install --frozen-lockfile`
- `pnpm typecheck`
- `pnpm lint`
- 713 affected tests passed after reinstalling the stock xterm patch

## Remaining certification

Draft until the native Linux IBus workflow passes on this head. Windows, macOS, SSH, rendering-scale, and reporter evidence remain separate goalpost work; no issue should be closed from this PR alone.


nwparker:
Closing as superseded by the targeted IME fixes and native coverage now on main, including #12259, #12265, #12273, #12278, #12280, #12281, and #12282. The current branch also conflicts with main and does not preserve the certified native macOS 2-Set Korean leading-vowel sequence. We will continue hardening IME support incrementally from the landed, tested path.

Issue #12118, opened 2026-08-02
[Bug]: Korean IME composition flickers/drops in the session chat input during agent-streaming re-renders (Windows)


DanielJ1025:
### Operating system

Windows

### Orca version

1.4.164

#### Short summary

In the **session chat input** (not the terminal) on the Windows desktop app, the
in-progress Korean (Hangul) composition character **flickers — it disappears and
reappears** while typing. The effect is **strongly correlated with agent streaming**:
it is severe while an agent is generating a response and calm when the session is idle.
Text can also appear to wrap to a new line, and a confirm-Enter during composition is
sometimes swallowed (no send / inserts a newline).

The committed value is preserved, so it is cosmetic, but it disrupts Korean typing badly.

#### What happened?

While an agent turn is streaming, the chat panel re-renders at high frequency (live
token counter / elapsed timers updating, e.g. "…53s ↓1.5k tokens", "Baked for 37s").
Each re-render appears to interrupt the active IME composition in the Chromium webview
input, so the composing 글자 is dropped and re-rendered → visible flicker. When the
session is idle (no re-renders), Korean input is stable.

Likely mechanism: the input node is remounted / re-keyed (or loses+regains focus) on
each streaming re-render, discarding the composition buffer. Enter handling also seems
to lack an `isComposing` guard, so a confirm-Enter during composition can send/newline.

#### How can we reproduce it?

1. Windows desktop app, Korean 2-Set IME (Windows `TextInputHost` active).
2. Start an agent turn so the chat panel is actively streaming (live counters ticking).
3. Simultaneously type a Korean sentence into the **session chat input**.
4. Observe the composing syllable flicker (drop + reappear); occasionally the
   confirm-Enter is swallowed and shows as a newline instead of sending.
5. For contrast: repeat while the session is idle → little/no flicker; type in English
   → no flicker (confirms IME composition cause).

#### Suggested fix

- Preserve/isolate the input's IME composition across streaming re-renders (don't
  remount/re-key the input node while streaming; throttle the streaming counters so they
  don't re-render the input subtree).
- Guard the send-on-Enter path with `isComposing` / `compositionend` so a confirm-Enter
  during composition never triggers send or inserts a newline.

#### Related issues

Appears distinct from existing reports (those are macOS and/or terminal, different
mechanisms), but likely the same problem family — cross-referencing for triage:
- #11332 — Korean IME breaks in **session chat input** (macOS; jamo-split mechanism)
[...]

PR #12263, opened 2026-08-03
fix(chat): preserve browser-owned IME composition

nwparker:
## Summary

- leave active textarea composition owned by Chromium instead of driving it with a controlled React value
- synchronize external drafts only while the browser is idle and reconcile the live DOM value at composition end
- stop IME-owned events at the existing application-action boundary for submit, picker, typing redirect, font scale, and chat toggle
- remove the duplicate parent composition ref and callback plumbing

NativeChatComposerField already existed. This adds no class, component abstraction, tracker module, registry, or production file.

## Native evidence

A physical macOS 2-Set Korean Electron journey types 테스트 into session chat while the agent is working:

- the same textarea retains 테스트 through 1.5 seconds of streaming rerenders
- the recorded Enter shape is key Enter, keyCode 229, isComposing true
- compositionstart and compositionend are both present in the attached Playwright trace
- exact PTY input is 테스트 followed by newline, once
- the paired ordinary path is ordinary followed by newline

## Mutation checks

- deleting the single IME action guard duplicates the composed send in the physical journey while the ordinary path remains ordinary
- changing the textarea back to controlled value draft fails only the active-composition positive; all six idle and ordinary field tests pass
- restoring preventDefault on IME Enter does not change the native macOS result; its removal is a native-ownership simplification, not the trace-proven mechanism

## Verification

- physical macOS Korean headful Electron E2E: 1 passed
- focused native-chat tests: 24 passed
- changed-file Oxlint and formatting
- node and web typechecks
- max-lines ratchet and diff check

## Remaining certification

This remains draft. Windows native Korean evidence, reporter confirmation, and issue-specific classification are still required by the goalpost. Terminal and CLI/TUI reports remain owned by the stacked terminal PR rather than this native-chat path.

### DIFF
--- a/src/renderer/src/components/native-chat/NativeChatComposer.test.tsx
+++ b/src/renderer/src/components/native-chat/NativeChatComposer.test.tsx
@@ -14,7 +14,6 @@ const mocks = vi.hoisted(() => ({
   fieldProps: null as {
     onSend?: () => void
     onStop?: () => void
-    onCompositionStart?: () => void
     onCompositionEnd?: (event: { currentTarget: HTMLTextAreaElement }) => void
     sessionOptionsSurface?: SessionOptionsSurface | null
     sessionOptionsSnapshot?: SessionOptionDescriptor[]
@@ -267,7 +266,6 @@ describe('NativeChatComposer', () => {
     mocks.setDraft.mockClear()
 
     act(() => {
-      mocks.fieldProps?.onCompositionStart?.()
       mocks.fieldProps?.onCompositionEnd?.({ currentTarget: textarea })
     })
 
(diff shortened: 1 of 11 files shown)


nwparker:
Superseded by #13128 (17b3dff3c41). The composer-ownership fix landed as `native-chat-composer-composition-hold.ts`: the composer stays mounted while a composition is in flight, because unmounting mid-composition aborts it at the OS level and commits partial text — typing 아 then ㄴ yields 아ㄴ rather than 안. `display:none`/`visibility:hidden` blur and abort the same way, so the hold has to keep it genuinely mounted. Closes #12118.

Issue #11729, opened 2026-07-31
SSH remote: environment not restored after login on Rocky Linux 9.4


AmethystLiang:
## Summary

On an SSH remote host running **Rocky Linux 9.4**, the remote environment is **not restored after each login**. Still broken as of **Orca 1.4.167** (also reported on **1.4.162**).

Related fix PR (not merged yet): https://github.com/stablyai/orca/pull/12264

## Environment

| Item | Value |
|------|--------|
| Connection | SSH remote (not Orca remote-server) |
| Remote OS | Rocky Linux 9.4 |
| Orca versions | 1.4.162, **1.4.167** (still fails) |

## Expected

After connecting/logging in again to the same SSH remote project, the previous environment (workspace / session state needed to continue work) should be restored — without spawning extra unused terminal panes / relay PTYs.

## Actual

- Environment is not restored after each login.
- User reports **100% reproducible**.
- Still broken on **v1.4.167** after earlier reports on v1.4.162.
- User notes the fix PR has **not merged yet**: https://github.com/stablyai/orca/pull/12264

## Notes

- Confirmed path is **SSH** (not beta remote-server).
- Reporter suspects **Rocky 9.4 compatibility** as a contributing factor.
- PR #12264 describes the engineering root cause as SSH remote-PTY lease identity / reattach fan-out (leases keyed only by `(targetId, ptyId)`, no per-pane retirement → reconnect reattaches all leases and can recreate UI panes). That matches “environment wrong after login / reconnect.”

## Impact

Blocks reliable day-to-day use of Orca over SSH against Rocky 9.4 hosts: every re-login loses or corrupts environment state.

## Request

- Land / verify #12264 against Rocky 9.4 SSH.
- Confirm whether any Rocky-specific follow-up remains after the lease/reattach fix.

rodboev:
This appears addressed by merged PR https://github.com/stablyai/orca/pull/11860, which restores SSH relay ownership after an app restart and includes a Rocky Linux 9.4 reproduction tied to this issue. Please retest the login restore path on a current build; if it still reproduces, leave the issue open with fresh logs.

AmethystLiang:
Closing as fixed.

Merged PR that closes this: https://github.com/stablyai/orca/pull/11860 (`fix(ssh): restore relay ownership after app restart`).

[...]

Issue #11006, opened 2026-07-27
Transient SSH disconnect treats live relay PTYs as dead, causing duplicate agent respawns when a tab is opened


eisen0419:
## Summary

After a transient SSH disconnect, a remote agent process can remain alive under the relay while its Orca terminal view comes back blank. The renderer clears the tab's PTY binding on connection loss and remounts panes after reconnect; if the renderer no longer has a restored leaf/session candidate, opening the affected tab enters cold restore and starts a second agent process instead of attaching the live relay PTY. The main/relay layers already have an authoritative `pty.attach` path with process-liveness and pane-identity checks, but the renderer's no-candidate cold-restore branch does not consult that authority. In our incident this produced two live `pi` processes using the same session file; the same path also attempted a Codex resume against an already active session.

## Environment

- Topology: macOS Orca client/runtime → SSH → remote relay and PTYs.
- Incident relay: `0.1.0+0eee5f4c09ed`. The `0eee5f4c09ed` suffix is the first 12 hex characters of a SHA-256 over the packaged relay artifacts, not a Git commit (`config/scripts/build-relay.mjs:108-123`).
- Source used for this analysis: `334027cfd5b41e032b7d34fc5bfe5c334e39c322`, described locally as `v1.4.150-rc.0-26-g334027cfd`.
- Exact incident client app build: available on request (the relay artifact hash above pins the incident relay; the client Mac is not accessible from where this report is assembled).
- Affected agents observed: pi and Codex.

Because the incident relay content hash cannot be resolved as a Git object, the code references below describe the matching failure mode in the analyzed HEAD; they do not prove byte-for-byte identity with the incident build.

## Minimal reproduction

1. Open an SSH worktree terminal and start a resumable agent such as pi. Record the remote agent PID/start time and its session file.
2. Interrupt only the SSH connection long enough for Orca to enter a disconnected/reconnecting state, while keeping the remote relay, PTY, and agent process alive.
3. Restore SSH connectivity and wait for Orca to report the target as connected.
4. On an affected run, observe that the agent tab is blank even though the original remote PID and session file are still active.
5. Open/click the blank terminal tab.
6. Observe a new process such as `pi --session <the-same-session-file>` starting at the click time while the original process remains alive.
7. For a regression test, assert that reconnect + tab activation performs `pty.attach` to the existing PTY and never calls fresh/cold-resume spawn unless the relay returns a definitive not-found/identity failure.

[...]

PR #12264, opened 2026-08-03
fix(ssh): stop reconnect from grafting ghost terminal panes (STA-3077)

AmethystLiang:
Fixes STA-3077 / #11729.

> **On the line numbers below.** Cites in *Root cause* and *RC0* describe the code **as it was on
> `main`** — they are the diagnosis, and they stop being findable at those offsets once this branch is
> applied. Cites in the later sections are against this branch's head. Every cite names its symbol, so
> a stale offset degrades to a grep rather than to a wrong claim.

## The symptom

> "SSH remote environment not restored after login" — reconnecting an SSH-backed workspace adds
> terminal panes the user never opened, and the remote host accumulates shells nobody is using
> (the reporter's own count went 2 → 19 → 20 relay PTYs across three reconnects).

**In one sentence:** an SSH pane's remote-PTY lease is keyed only by `(targetId, ptyId)`, so nothing
ever retires the lease a pane leaves behind; every reconnect reattaches *all* of them, and the reattach
path writes through a **creating** store method that splits the layout root to graft the missing leaf
back in.

Both halves of the ticket title are literally true at different moments, and they feed each other.
`reattachKnownPtys` is the last blocking gate before a reconnecting session reaches `'ready'`
(`src/main/ssh/ssh-relay-session.ts:633` → `:641`), so every stale lease enlarges the fan-out the next
reconnect must complete. Loss on cycle N becomes duplication on cycle N+1.

## Root cause

**RC1 — the lease store has no pane key.** `Store.upsertSshRemotePtyLease`
(`src/main/persistence.ts:7022-7054`) finds an existing row by identity alone:

```ts
const existingIndex = this.state.sshRemotePtyLeases.findIndex(
  (entry) => entry.targetId === normalizedLease.targetId && entry.ptyId === normalizedLease.ptyId
)
```

`worktreeId`, `tabId` and `leafId` ride on the row (`src/shared/ssh-types.ts:147-158`) but participate
in no uniqueness rule anywhere. Two upserts for the same pane with different `ptyId` produce two
`attached` rows. And there is no per-lease retirement path to undo it: `removeSshRemotePtyLease`
(`persistence.ts:7152`) has exactly one caller — the spawn-persist rollback at
`src/main/ipc/pty.ts:5561` — and the only other removal is the plural
`removeSshRemotePtyLeases(targetId)` used by SSH-target deletion (`src/main/ipc/ssh.ts:138`). Lease
count is monotonically non-decreasing.

**RC2 — reattach fans out over every lease, and the relay's fence structurally cannot help.**
`reattachKnownPtys` (`ssh-relay-session.ts:1848-1918`) filters on state only (`:1853-1854`:
`state !== 'terminated' && state !== 'expired'`), so orphans are included. The relay's identity fence
(`src/relay/pty-handler.ts:1608-1619`) compares the incoming pane key against **the PTY's own recorded
pane key** (`managed.attachIdentity ?? { paneKey: managed.paneKey, ... }`) — and two PTYs minted for
the same pane each recorded *that pane's* key, so both pass. `attachIdentityMismatches`
(`pty-handler.ts:332`, JSDoc `:327-331`) fires on cross-generation id *collisions*, a different
failure. The fence guards relay-id reuse, not pane cardinality, and there is no "at most one live PTY
per pane" check on either side of the wire. Only `Store` can see two leases for one pane.

**RC3 — the reattach binding write creates UI.** `restoreReattachedPtyRuntime`
(`ssh-relay-session.ts:2112-2136`) calls `store.persistPtyBinding(...)` at `:2124`. That method has
four creating branches: mint a tab (`persistence.ts:6666-6685`), mint a root leaf (`:6700-6705`),
**split the root and graft the leaf** (`:6706-6719`), mint a layout (`:6724-6734`). All four are
deliberate and load-bearing *for the spawn path* — the comments cite the Issue #217 force-quit race,
where `pty:spawn` beats the renderer's debounced layout writer. None of it is appropriate on reattach,
where the pane either already exists or is gone for good.

> **RC3's precondition, stated because it is the single most misread fact in this investigation.**
> The graft at `:6707` is gated on `!layoutContainsLeafId(layout.root, args.leafId)`
> (`persistence.ts:1699-1707`, recursive). A stale lease reattaching onto a leaf that **still exists**
> does **not** graft — it falls through to the plain `layout.ptyIdsByLeafId[leafId] = ptyId` write at
> `:6720-6723` and adds no pane. So duplicate leases on a *live* leaf produce a ptyId overwrite, not an
> extra pane. An extra pane requires the leafId to be **absent** from the durable layout. That is why
> the fix has two independent halves: one owns unbounded lease growth, the other owns the pane count,
> and neither substitutes for the other.

**RC4 — a failed reattach mints the duplicate.** The lease stays non-terminated while the renderer
fresh-spawns for the same pane. **RC4b:** when the relay reports an *identity mismatch* — proving the
row names a PTY that now belongs to someone else — `handlePtyReattachFailure` returns without retiring
it. That lease can never attach and is never retired: a permanent zombie re-entering the fan-out on
every future reconnect. RC4b is why the count grows without bound instead of oscillating.

**RC5 — retirement is not durable.** `markSshRemotePtyLease` (`persistence.ts:7123-7150`), the
*singular* setter every retirement path uses, has no terminal-state guard: it short-circuits only on
`lease.state === state` (`:7132`) and otherwise assigns unconditionally at `:7139`. Its plural sibling
`updateSshRemotePtyLeaseStates` guards both directions (`:7097` refuses `attached` over a
terminated/expired row, `:7100` refuses `detached` over anything not `attached`) — so the invariant
exists, it is just not where every caller inherits it. `abandonPtySourceRecovery`
(`ssh-relay-session.ts:2575`) writes `'detached'` unconditionally over whatever it finds, returning a
correctly retired lease to the reattach-eligible set. Observed directly during a SIGSTOP experiment:
`expired` → `detached` on a second reconnect. This is why a fix aimed only at RC4 can ship fully green
and regress silently later.

### The evidence

A maintainer's pre-migration store held 15 rows for one worktree. Two are residue from a Docker e2e
harness and carry a different `targetId`, leaving **13 real leases across four leaves** — three leaves
with four leases each, one with a single lease. Read at millisecond precision, one reconnect resolves
into four beats:

```
wave 1   .274 / .296 / .318           pty-1  pty-2  pty-3          reattach, cycle-1 leases
wave 2   .592 / .617 / .638           pty-4  pty-5  pty-6          reattach, cycle-2 leases
   └─ terminated .746 / .758 / .770                                ← a synchronized sweep, ~12 ms apart
wave 3   .781 / .803 / .824           pty-7  pty-8  pty-9          reattach, cycle-3 leases
wave 4  03.143 / .176 / .209 / .245   pty-10 pty-11 pty-12 pty-13  createdAt == lastAttachedAt: FRESH MINTS
```

Three PTYs minted weeks apart were all reattached onto one leaf by a *single* reconnect — that is the
accumulation, and it is what this artifact proves. Every leaf ended with exactly four leases, though
the mint occasions are **not** aligned across leaves: one leaf's second mint came 27 seconds after its
first, in the same session, while the other two waited 12 days. So this is per-leaf accumulation, not
a synchronized per-reconnect wave.

(Read these at millisecond precision or not at all — at minute granularity all thirteen rows collapse
to "07-06 22:47" and pty-4/5/6 look like leases the `terminated` filter correctly excluded. pty-4 was
*attached* at .592 and terminated at .746: an outcome of the sweep, not an exclusion from it.)

**Two limits on this artifact, stated so nobody over-reads it.** First, wave 4's four fresh mints are
**not** established as RC4's signature: pty-13's leaf held no prior lease, so reattach cannot have
minted it, and once a pane-driven spawn exists in that 102 ms window it explains all four. RC4 rests
instead on code reading plus a second machine's live store, where two leaves that *did* already hold
leases each minted a duplicate 19 s and 63 s later. Second, this table evidences **lease
accumulation**, not **pane multiplication** — waves 1–3 land on a single shared leaf, and a reattach
onto a leaf that already exists overwrites a binding rather than adding a pane. Pane multiplication is
evidenced by RC4b/RC4c and by the graft's absent-leaf path, not here.

## RC0 — one write window, three consequences

**A lease is durable the instant it is written; the layout that gives it meaning is not.**
`upsertSshRemotePtyLease` ends in `this.flush()` (`persistence.ts:7053`) and the spawn path calls it
synchronously (`pty.ts:5511-5519`). The layout write is doubly debounced — 150 ms in the renderer
(`src/renderer/src/lib/session-write-subscriber.ts:163`), then 1 s up to 5 s in main
(`persistence.ts:3819-3820`). `pty.ts:5524-5526` names this window in its own comment.

**Worse than late: during a remote-workspace snapshot apply the write is *dropped*, not deferred.**
The subscriber gates on `!isRemoteWorkspaceSnapshotApplyInProgress()`
(`src/renderer/src/App.tsx:1262`); when the gate is closed it clears the pending change set and returns
(`session-write-subscriber.ts:201-208`, and again at `:227-229`) — *after* advancing its baseline at
`:197` (`prev = next`). So the next comparison sees no change, nothing re-emits it, `persist({ patch })`
at `:237` never runs, and a later `session.flush()` cannot rescue a patch main never received. The gate
is held for a 30 s reconnect deadline plus a 1 s tail
(`src/renderer/src/hooks/remote-workspace-snapshot-apply.ts:21`, `:20`).

The loss is **not unconditional**, and the precise version is more damning than the blanket one: the
timer rebuilds its payload from live state (`:222` `store.getState()`, `:233`
`buildWorkspaceSessionPatch(fresh, changed)`) rather than from a captured diff, so any *later* mutation
of the same field carries the earlier one forward. What is permanently lost is a mutation that
**settles during the freeze and is never touched again** — which is exactly a pane close, because a
closed pane generates no further layout events.

Three consequences follow, and each one misled part of this investigation:

1. **RC3's precondition gets manufactured.** An ungraceful death inside that window leaves a lease
   whose leaf the durable layout never recorded — which is precisely the absent-leaf state the graft
   requires.
2. **A pane close during a reconnect leaves the screen but is never durably recorded.**
3. **"Absent from the durable layout" is therefore never trustworthy evidence for a *destructive*
   action.** This is the governing rule of the whole change: **an absence can justify "do not create";
   it can never justify "destroy."**

Nothing in this PR closes that window. What this PR does is stop acting destructively on its output.

## What each commit does, and what it can destroy

Five commits, ordered by blast radius, so each boundary is independently revertable and the riskiest
code lands last against the safest foundation. Four more follow: the e2e specs, and three fixes found
by the review rounds (each described in *The reap flag has a scope* above).

| # | Commit | What it can destroy |
|---|---|---|
| 1 | `persistPtyBinding` gains `mayCreate: false`; reattach may no longer create pane structure | **nothing** — it can only decline to create |
| 2 | Two `Store` retirement primitives, and no third way to retire a lease | nothing on its own; it constrains everything after |
| 3 | Pane-keyed supersede, terminal-state guard on the singular setter, orphan reaping, load-time migration | a remote shell, via the reaper |
| 4 | Failed-attach exits converge on one retirement decision; a relay generation reset retires that target's leases | a lease row only |
| 5 | Durable close intent recorded before the kill RPC | a remote shell, but only on explicit user intent |
| 6 | The two e2e specs | nothing |
| 7 | Relay incarnation **token**; bulk retirement pays one sync flush instead of N | narrows #4 — retires strictly fewer leases |
| 8 | A revived lease no longer inherits the retired row's reap flag | narrows #3 |
| 9 | The reap drain is armed only inside the incarnation that minted the flags | narrows #3 and #5 |

Commits 7–9 only ever *withhold* a destructive action, so the safety argument for 1–5 is not weakened
by reading them: each one strictly shrinks the set of remote PTYs this change is willing to kill.

**Commit 1 is non-destructive and shippable alone**, which is what makes this ordering worth following
rather than merely tidy. It needs no new predicate: every creating branch in `persistPtyBinding`, and
only the creating branches, already sets `terminalMembershipChanged = true`, so that flag *is* the
pane-existence predicate — already computed, already correct, and maintained by whoever edits the
function next. The function also already owns its rollback (`cloneWorkspaceSessionState` at `:6625`,
`restoreSession()` at `:6640-6649`). Under `mayCreate: false` a would-be-creating write restores,
skips the flush, and returns `'refused'`: skip `registerPty`, leave the lease `detached`, log, retry
next reconnect. **No retirement and no remote shutdown** — refusing to create is the only safe action
on an absence.

Two implementation notes a reviewer should check: there are **two** flush points (`:6690-6695` and
`:6738-6743`) and both need the guard, or there is a hole on exactly the legacy pane ids; and the clone
at `:6625` precedes the incarnation write at `:6650-6660`, so a refusal correctly rolls that back too.

**Commit 3 also heals existing profiles.** Every affected user already has duplicate leases on disk, so
a fix that only prevents future duplication leaves them broken. A load-time reconcile groups
non-terminated leases with a complete pane identity, keeps the newest by `updatedAt` (tiebreaking
`createdAt` then `ptyId` — the snapshot has three rows inside one second, so this must be specified and
tested, not left to sort stability), retires the rest, and flags them for reaping. It is idempotent and
runs unconditionally on every load for every profile, so it is a no-op on clean stores rather than
something gated on a heuristic.

**Both new durable fields must be taught to `normalizeSshRemotePtyLease` (`persistence.ts:1585-1610`).**
That normalizer is a strict allowlist that rebuilds the row field by field, and it fails two ways,
neither loud: an unlisted field is silently dropped on load (works in memory, works in every in-process
test, dead after restart), and a value outside a validated literal union makes it return `null` so the
**whole row** is filtered out at `:3631-3633`. Every test for a new durable field must round-trip
*through normalize* — construct, persist, reload, assert. A set-then-read test passes in both failure
modes and proves nothing.

## Why there are two retirement primitives

Everything that retires a lease goes through exactly one of two `Store` methods; nothing else writes
lease state directly.

| Primitive | Reaps the remote PTY? | Callers |
|---|---|---|
| `retireLeaseAndReap(targetId, ptyId, reason)` | **yes** — sets `pendingRemoteShutdown` | supersede, close intent, migration |
| `retireLeaseSparingPty(targetId, ptyId, reason)` | **structurally cannot** | failed-attach exits, generation reset |

**The constraint that makes this non-negotiable: retiring must never enqueue a relay pty id for
shutdown when the evidence is about the *binding*.** A reset relay reuses `pty-N` —
`src/relay/pty-handler.ts:329` says so outright, and the identity fence at `:1610` exists for it — so
`pty-1` in a new generation may be another pane's **live** shell. The identity-mismatch case is the
sharp one: `pty-handler.ts:1597-1607` runs a liveness probe *first* and routes a dead PTY to a plain
not-found, so reaching the fence at `:1610-1619` **proves** the remote process is alive and belongs to
another pane. Handing its id to the reaper would kill working sessions.

"Structurally cannot" is stronger than "does not" on purpose: `retireLeaseSparingPty` must neither
accept nor expose `pendingRemoteShutdown` in any form, so a future caller cannot opt in by passing a
flag. This design reaches the id-reuse trap at two independent call sites, and prose alone would leave
it to both of them to remember.

The split is by **what the evidence is about**, not by which handler produced it: "the pane is gone,
proven positively" → retire and reap; "the lease is wrong about its PTY" → retire and never touch the
process; "nothing (a timeout, a dead channel)" → no retirement at all, retry next cycle.

### The reap flag has a scope, and it is one relay incarnation

The two-primitive split above is necessary but not sufficient, and review round 3 is what surfaced why.
`pendingRemoteShutdown` asserts *"relay id `pty-N` is ours to kill"* — a claim that is only true of the
relay **process** that minted it, because the next incarnation restarts `pty-N` numbering from scratch.
A flag is minted at connect *N* and drained at connect *N+1*, so a relay restart in between silently
re-points it at whatever holds that id now, which may be another client's live shell. That hazard is
independent of which primitive set the flag: it reaches the load-time migration, the pane-keyed
supersede, and the durable close intent alike, so guarding any one of them individually would have
fixed a third of it.

So the drain is armed only when `relay.status` **proves** the incarnation is unchanged. Unknown,
unparseable, unreachable, changed, or no stored incarnation at all → the flags for that target are
abandoned before the drain runs (`disarmReapDrain`, `ssh-relay-session.ts:1943`). Retirement still
fails *open* — an unavailable RPC retires nothing — but reaping fails *closed*, because the two
failure modes are not symmetric: abandoning a flag leaks a remote shell, honouring a stale one kills a
live pane.

That identity check needs to survive clocks it does not control, so the relay now mints a
`randomUUID()` **incarnation token** at startup and returns it from `relay.status`
(`src/relay/relay.ts`). When both readings carry a token the comparison is exact and needs no clock;
`pid` + derived start time (±30 s) remains only for a relay still running from a pre-token deploy. The
fallback alone was wrong in both directions — an NTP step on either host retires every lease on a relay
that never restarted, and a recycled pid inside the tolerance passes as "same process".

A related one-line rule, from the same round: a lease row is keyed `(targetId, ptyId)`, so a fresh
spawn on a reset relay lands on a **retired row's key**. `upsertSshRemotePtyLease` merges over the
existing row, so retirement metadata was inherited by the new binding — handing a live pane's lease a
`pendingRemoteShutdown` it never earned. A revive to a live state now clears both retirement fields.

## What this does **NOT** fix

- **The loss half, as reproduced, is NOT fixed.** An instrumented run settles this independently:
  SIGTERM the relay while the SSH transport stays up → the replacement relay destroys every remote
  shell. The session reached `'ready'` **promptly**, and the first sample **27 ms after
  `ssh.connect()` returned** already showed 3 SSH panes → 1, unchanged through 45 s, with **zero**
  reattach-failure log lines. No fan-out invariant was violated and the panes went anyway.
- **Relay death destroys the remote shells, and nothing in this PR restores them.** The 300 s grace
  window protects against *Orca* detaching, not against the relay process dying. A destroyed shell
  cannot be brought back, so the honest fixes are a dead-but-respawnable pane presentation or fewer
  relay deaths — both larger than this change, and both requiring a UI decision nobody has made.
  Commit 4 closes a *different* piece of the same repro (the stale leases the dead generation leaves
  behind) without closing the loss itself.
- **The reattach fan-out can still gate `'ready'`.** Deferred deliberately: removing the relay churn
  that currently reaps zombie leases by accident would *extend* the life of leases that can never
  attach, so that follow-up must not ship before commit 4.
- **A pane close is still removed from the screen before it is durable.** The visual teardown is
  unconditional and upstream of the kill, and the kill stays deliberately non-blocking — a close must
  not wait on a wedged transport. What changed is that the *intent* no longer depends on the RPC: main
  retires the lease **before** the shutdown call (`pty.ts`), so a throw on a severed transport can no
  longer leave a live lease behind, and the renderer now at least reports the rejection instead of
  dropping it silently (`pty-transport.ts:982`). Net: the lease is right; the pixels are still ahead of
  the disk.

  **The stranded leaf resurfaces only on mount or remount.** `replayTerminalLayout`
  (`src/renderer/src/components/terminal-pane/layout-serialization.ts:234-273`) is the only site that
  reconstructs panes from a persisted root, and its single call site
  (`use-terminal-pane-lifecycle.ts:1475`) sits in a `useEffect` keyed `[tabId, cwd]`. So the pane comes
  back when the user switches tabs or worktrees away and back, or relaunches — **not** during a
  background reconnect while the pane stays mounted. That matches the ticket title ("not restored after
  login" = reopen-after-reconnect), and it means **any test for this must remount**; asserting pane
  count immediately after a reconnect passes on a broken build.

  This may be the reported symptom's actual cause, but the minimal fix is a renderer-side **ordering**
  change in an otherwise entirely main-process PR, and whether the pane should stay, return with an
  error affordance, or be re-added on reconnect is an open UX call. Sibling PR.
- **PR #11860 is not a duplicate of this.** `ed00ab0f34` restored relay *consumer ownership* after an
  app restart; it touched neither the lease upsert nor the reattach binding, and its oracle
  structurally cannot catch duplication — the restart test uses one tab and asserts a single PTY
  *survives* (`tests/e2e/ssh-cold-activation-restore.spec.ts:213-311`), which passes cheerfully while
  eleven extra PTYs accumulate beside it.
- **The load-time migration cannot heal an existing RC4b zombie.** It collapses *duplicates*: a pane
  with two live leases keeps the newest and retires the rest. A zombie that is the **only** row for its
  triple is the newest for that triple, so it survives the reconcile untouched. Those clean up on the
  first reconnect after this ships, when the failed-attach exits prove the row mismatched — one cycle
  later than the duplicates, and never at all for a target the user does not reconnect.
- **Three remote-shell leaks are accepted, in exchange for never killing a live pane.** (a) A refused
  reattach leaves its remote PTY running — strictly better than the ghost pane it replaces, but the
  shell is still there. (b) The graceful-disconnect tombstone fast-path in `pty:kill` is deliberately
  unchanged, so a close taken through it does not record a reap. (c) Every disarmed drain abandons its
  flags outright. All three follow the same rule: an absence justifies *declining to act*, never
  destroying. With a grace period of 0 (never expire) these shells persist until the relay dies.
- **The reap drain still fans out on the connect path.** Up to 32 concurrent shutdowns are awaited
  before the session settles. The same path already awaits per-lease attaches under the same mux
  timeout, so this adds no new class of stall — but a deadline around the fan-out is a follow-up, not
  something this PR does.
- **Do not "fix" this by raising the timeouts.** A bloated lease set genuinely can prevent a reconnect
  from settling (10 s per attempt, `src/main/providers/ssh-pty-provider.ts:199`; two retries;
  concurrency 8, `ssh-relay-session.ts:117`) — but that is a *rate* argument, not a cause. Raising any
  of the three lengthens time-to-visible-breakage without removing the defect. Nothing in the chain is
  distro-dependent; Rocky 9.4 is incidental.

## Retracted measurement — do not re-run it

An earlier round of this investigation reported durable-state evidence that a closed pane's leaf
survived in the layout ("measured 2/2"). **That reading is void and is retracted here so nobody
re-derives it and reports a false confirmation.**

It read the layout from `raw.workspaceSessionsByHostId['ssh:<targetId>']`. That partition **cannot
contain renderer pane truth for an SSH worktree.** The renderer's session writer resolves an SSH-backed
worktree to the *local* host id — `buildHostIdByWorktreeId` ends in
`return parsed?.kind === 'runtime' ? parsed.id : LOCAL_EXECUTION_HOST_ID`
(`src/renderer/src/lib/workspace-session-host-persistence.ts:204-208`); **only `runtime:*` is
partitioned out.** And main's own store code says where local lives, three times:
`persistence.ts:569` ("'local' (the legacy `workspaceSession` blob) is dropped so the two surfaces
never diverge"), `:581` ("a local/invalid key here is legacy noise that must not shadow the canonical
partition"), `:3620` ("'local' stays in `workspaceSession` for downgrade compat"), with the parser
`continue`-ing on a `local` key at `:582-584`. The local partition is **structurally absent** from
`workspaceSessionsByHostId`.

(The `ssh:*` partition is not *unused*, just not renderer pane truth: four main-side writers feed it —
`pty.ts:4415-4418`, `pty.ts:5542`, `orca-runtime.ts:4610` and `:4668` — and one main-side consumer
reads it, `Store.clearSshRemotePtyBindingsForLeases`, which unions it with the local session to clear
bindings that name a retiring lease. That consumer sits on this PR's retirement path, so retirements
drive binding-clears against this mirror too.)

What that key *does* hold is a spawn-fed mirror written only by main (`pty.ts:5541-5545`;
`orca-runtime.ts:4630-4635`, host resolved at `:4600-4614`). So `afterClose === beforeClose` there is
the expected reading under **every** hypothesis, including a perfectly recorded close. It proves
nothing in either direction — a void measurement is not weak evidence.

**The rule this section originally drew — "always read `raw.workspaceSession`" — is itself too strong,
and is corrected here.** Which partition owns a tab depends on its execution host:
`Store.getWorkspaceSession(hostId)` returns `state.workspaceSession` only for
`LOCAL_EXECUTION_HOST_ID` and `state.workspaceSessionsByHostId[hostId]` for every remote one. Hardcoding
*either* side risks the same vacuous read in the other direction, which is why both e2e specs here go
through `readPersistedTerminalLayout` (`tests/e2e/helpers/persisted-terminal-layout.ts`): it searches
both partitions, reports **which one answered**, and both specs assert non-null before asserting
anything about the contents. A layout read that silently returns `null` forever turns every downstream
assertion into a tautology, so the guard is the load-bearing part.

(That the two resolvers disagree at all is a real latent defect — an unbounded parallel tabs/layout
tree that nothing reads and nothing prunes — and is filed separately. Do not "make the partitions
consistent" here; doing it without first fixing the read side would break the one main-side write that
currently reaches the renderer.)

## Test plan, and the honest state of the e2e

**Unit — `Store` invariants.** Pane-keyed supersede across a different `ptyId`; legacy rows without a
complete identity are exempt and untouched; terminal states are terminal *in fact*, including the
specific `'detached'`-over-`'expired'` write observed resurrecting a lease; `retireLeaseSparingPty`
leaves the reap queue empty; every new durable field round-trips through normalize.

**Unit — `mayCreate: false`.** One test per creating branch asserting `'refused'` and an unchanged
session — covering both flush points and the legacy-`leafId` path — plus the non-creating rebind
asserting `'bound'` with behavior identical to today.

**Unit — retirement dispositions.** Failed attach: transport-ish failure → `detached`, retry next
cycle, no retirement; relay says not-found → `expired`; identity mismatch → `expired` **with the reap
queue asserted empty** — that last one is the branch where the remote process is proven alive and
belongs to someone else. Generation reset: identity match → no-op; identity differs → every lease for
the target retired via the sparing primitive, reap queue empty; RPC unavailable / errored / no prior
identity → **retires nothing**, as three separate cases, because failing open is the entire safety
property.

**Unit — incarnation identity and drain arming.** A matching **token** keeps every lease however far
the derived start has drifted (the NTP-step case the pid+time fallback gets wrong); a new token retires
even when the pid was reused inside the ±30 s tolerance (the recycled-pid case it also gets wrong); a
token-less *reading* against a stored token falls back rather than retiring, since that pairing means an
older deploy, not a new process — and asserts the stored identity does **not** gain an empty token. The
drain is asserted disarmed in all three unproven cases (changed / no previous / `relay.status` failed),
with the fake store answering the claim out of the same flag set it clears — otherwise the drain would
reap from a queue that clearing had already emptied and the ordering would prove nothing. Round-trip
through normalize covers a lease reviving onto a retired row's key, and the per-target scoping of the
flag clear.

**E2E — Docker SSH (`ORCA_E2E_SSH_DOCKER=1`), two specs.**

`ssh-reconnect-no-ghost-pane-graft.spec.ts` drives the close-while-severed case that the seeded shape
below cannot: SIGSTOP the relay, close a pane, disconnect (which tears down the mux and invalidates the
pending shutdown), unfreeze, reconnect, then sample for 20 s at 2 s intervals. It asserts the close
intent went durable **despite the failed RPC**, that the retired lease never reappears as
`attached`/`detached` in any sample, that the closed PTY id never appears in any pane's
`dataset.ptyId`, that the pane count never exceeds the post-close count, and that the id is gone from
`ptyIdsByLeafId`. The freeze is not incidental: a *graceful* disconnect unregisters the provider, so
`pty:kill` takes the tombstone fast-path and never issues the RPC that has to fail for this to be a
test of anything.

`ssh-relay-redeploy-stale-pty-binding.spec.ts` covers the generation-reset half. Both specs resolve the
tab from the **renderer**, not from a lease row — the lease's `tabId` is one of the things under test,
so keying the spec on it could mask the bug — and both read layouts through the partition-searching
helper described above.

**Honest state of the e2e, stated because it changes how much the green means:**

- **The freeze replaced a race, and only because the race was not runnable.** RC3's natural producer is
  an ungraceful death inside a ~1.15–5 s durable-write window; racing that with a SIGKILL was flaky in
  every attempt, and the "disconnect, close a pane, reconnect" shape proposed as an everyday trigger
  came back **negative** (the graceful path takes the tombstone fast-path and terminates the lease
  cleanly). SIGSTOP makes the same condition deterministic — the transport is severed for exactly as
  long as the test needs — so the spec is a real repro rather than a seeded precondition. What it does
  *not* do is reproduce the field's own producer; it reproduces a condition with the same signature.
- **If you seed leases instead, two traps make the test silently vacuous.** Seed the **profile-scoped**
  store (`profiles/<profileId>/orca-data.json`), not `path.join(userDataDir, 'orca-data.json')` — the
  latter is a pre-migration path, and a run that used it returned `"leasesAfter": null`. And seed
  **relay-form** ids (`pty-N`), not app ids: leases store the relay form
  (`getRelayPtyIdForSshLeaseStorage`), so an app-form id in a seeded lease is never matched at attach
  time and the test asserts against a fan-out that never considered it.
- **Failure injection is mandatory.** A clean reconnect loop passes on the *unfixed* build, because
  without a reattach failure no duplicate lease is minted.
- **Any close-related oracle must remount.** Panes are reconstructed from the persisted root only by
  `replayTerminalLayout`, whose one call site is keyed `[tabId, cwd]`, so a stranded leaf is invisible
  until a tab/worktree switch or a relaunch. Asserting pane count straight after a reconnect passes on
  a broken build — a distinct trap from the partition-read one above, and both were live in the same
  spec.
- **The relay-death spec must not assert that panes are restored.** They are genuinely dead; that
  assertion would be false by design. It asserts the in-scope things: the surviving pane does not retain
  a binding to a destroyed PTY, and the dead generation's leases are retired.


dbachko:
Thanks, this was a really bad bug

nwparker:
Your root-cause analysis here is what made #13111 possible — RC1 (lease identity has no pane key, and `removeSshRemotePtyLease` has exactly one caller so lease count is monotonically non-decreasing), RC2 (reattach fans out over every non-terminated lease; the relay fence compares each PTY against its *own* recorded pane key so both duplicates pass), and RC3 (the reattach write goes through a *creating* store method). I verified all three still reproduce on current `main` and cited them in the code comments.

#13111 is a smaller take on the same defect — ~180 production lines, rebased on current `main` — plus a fix for the duplicate-agent family (#12448, #11006, #12699): `restoreRequired` was reported as `SSH_SESSION_EXPIRED`, and the reattach handler treated every error as proof the session was gone, so a transport fault respawned a shell that resumed the same agent session into one transcript.

Flagging it here so we don't land conflicting fixes. Your RC4 — a failed reattach never retiring its lease, leaving a permanent zombie that re-enters every future fan-out — is not addressed by mine and is worth keeping on the table.

PR #12274, opened 2026-08-03
fix(terminal): drop IME preedits abandoned by Backspace

kcfang:
## Summary

Backspace that empties an IME preedit no longer replays the abandoned preedit into the PTY.

On macOS Zhuyin, pressing Backspace on the last symbol of a preedit deleted nothing and instead typed that symbol into the terminal, so the user had to press Backspace a second time to actually remove it. Reproduced on a real macOS Zhuyin session with `stty -icanon -echo; cat -v`, which returned the raw bytes `E3 84 91` / `E3 84 98` (`ㄑ` U+3111, `ㄘ` U+3118) — no CR and no DEL.

Root cause is in our own xterm patch, not upstream. `CompositionHelper._sendPendingComposition` resolves the session payload as:

```ts
textareaInput || pending.endData || pending.compositionData
```

and `_lastCompositionData` is deliberately not cleared by an empty `compositionupdate`. When an IME empties its preedit instead of committing, all observable channels are empty, so the chain falls back to `compositionData` — the preedit the user just deleted — and sends it. Pristine `@xterm/xterm@6.1.0-beta.287` emits nothing for the same event sequence, verified by running the identical sequence against an unpatched build.

The fix adds `installTerminalImeAbandonedPreeditGuard`, which treats a composition as abandoned when `compositionend` carries no data **and** no following `input` event carries text. A real commit always surfaces on one of those two channels, so the discriminator does not depend on the `compositionupdate` sequence, which differs per IME dialect (Sogou/fcitx blank the preedit first; macOS can go straight to `compositionend`).

It is wired into both xterm surfaces:

- `terminal-ime-composition-route.ts` (terminal pane) skips the payload instead of calling `terminal.input`.
- `preview-terminal-ime-bridge.ts` (dashboard popout preview terminal) has no composition route and would otherwise emit through xterm's own data event, so it cancels the session-end instead.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — exit 0; oxlint, native and type-aware code-quality audits all report 0 warnings / 0 errors; reliability gate manifest check passed for 62 gates
- [x] `pnpm test` — 4111 files passed, 16 skipped; 44115 tests passed, 135 skipped, 0 failed
- [x] `pnpm build` — exit 0 (desktop, CLI, web projection, native)

New file `terminal-ime-abandoned-preedit.test.ts` drives real `Terminal` instances with the real composition route, not mocks:

| Case | Expectation |
| --- | --- |
| Backspace empties preedit, IME blanks it with its own `compositionupdate` first | nothing reaches the PTY |
| Backspace empties preedit, IME goes straight to `compositionend` | nothing reaches the PTY |
| Backspace only shrinks a multi-character preedit | still composing, nothing reaches the PTY |
| Sogou-style commit announced by an empty `compositionupdate` | `你` routes exactly once |
| fcitx-style commit whose text only lands on the `input` event after an empty `compositionend` | `你` routes exactly once |
| Routeless dashboard-popout preview terminal, abandoned preedit | nothing reaches the PTY |

The last three are the important ones: they fail if the guard over-suppresses. Forcing the guard off turns the four suppression assertions red while the three commit assertions stay green, so the suite discriminates rather than passing vacuously.

Registered under the existing `terminal-input.ime-and-synthetic-forwarding` reliability gate (commands, testFiles, assertionRefs, evidenceRuns), and narrowed its `knownGaps` entry accordingly.

## AI Review Report

Review focused on the risk that a payload guard drops text the user actually committed.

**Cross-platform.** The guard contains no platform branch — it reads only `CompositionEvent.data` and `InputEvent.data`. No paths, no shell, no keyboard-accelerator or shortcut-label changes, so the `metaKey`/`CmdOrCtrl` and `⌘`/`Ctrl+` rules are not touched. One behavioral note: `installPreviewImeBridge` previously returned `null` on non-darwin and now always returns a bridge, because the abandoned-preedit path is not macOS-specific. Both call sites use `imeBridge?.claimKeyEvent(event) ?? false` and `imeBridge?.dispose()`; the non-darwin bridge returns `claimKeyEvent: () => false`, so observable behavior on Linux and Windows is unchanged and cleanup is now actually performed. Windows MS-IME and Linux fcitx/IBus commits always carry text on `compositionend` or the following `input` event, which the two "must route" tests pin — but that reasoning is covered by unit tests only, not by real-OS IME automation on those platforms.

**SSH / remote / local.** No change to transport, PTY ownership, or process assumptions. The composition route's existing captured-transport and pty-id checks are untouched; the guard only adds one more reason to skip `terminal.input`, evaluated before those checks change any state.

**Agents / integrations / git providers.** Nothing provider- or agent-specific. The change sits below every CLI agent in the renderer input path.

**Performance.** Three passive listeners per terminal element and two booleans. No allocation in the keystroke path, no polling, no IPC. Listeners are removed in `dispose()`, which both surfaces call on teardown.

**UI quality.** No UI surface touched.

**What was flagged and changed.** The first implementation keyed on `compositionupdate('')` preceding `compositionend('')`. That sequence was assumed, not observed, and would have silently failed on IMEs that skip the blanking update — the fix would have looked correct while doing nothing. It was replaced with the dialect-independent discriminator described above, and a second test case was added for the skipped-update dialect. The first implementation also only covered the terminal pane; the dashboard popout preview terminal was found to be independently affected and was fixed in the same change.

## Security Audit

No input validation, command execution, path handling, authentication, secrets, IPC, or dependency surface is touched. The change is a renderer-local DOM event state machine that can only ever *suppress* one already-computed string; it cannot construct, widen, or redirect a payload, and it does not alter which PTY a payload reaches. The residual risk is functional rather than security-related: an over-broad guard would drop legitimate committed text, which the three "must route" test cases exist to catch. No follow-up needed.

## Notes

- The bug is triggered by IMEs that empty a preedit instead of committing (macOS Zhuyin Backspace is the reported case), but the fix is platform-neutral and applies wherever xterm runs.
- This is a regression from our own `config/patches/@xterm__xterm@6.1.0-beta.287.patch`, not an upstream xterm bug, so no upstream PR is needed. The `|| pending.compositionData` fallback is intentionally kept for IMEs that clear the textarea before committing; only the abandoned case is filtered out.
- Real-OS IME automation remains a registered gap for this gate on every platform. Verification here is renderer-unit plus a manual macOS Zhuyin reproduction.

### DIFF
--- a/config/reliability-gates.jsonc
+++ b/config/reliability-gates.jsonc
@@ -5697,10 +5697,12 @@
       "commands": [
         "pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/terminal-ime-native-text-forwarder.test.ts src/renderer/src/components/terminal-pane/terminal-ime-input-source.test.ts src/renderer/src/components/terminal-pane/terminal-paste-runtime.test.ts src/renderer/src/components/terminal-pane/terminal-ime-composition-tracker.test.ts src/renderer/src/components/terminal-pane/terminal-ime-candidate-key-release-guard.test.ts src/renderer/src/components/terminal-pane/xterm-bypass-policy-non-mac.test.ts src/renderer/src/components/terminal-pane/xterm-bypass-policy.test.ts",
         "pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/terminal-ime-linux-candidate-state.test.ts",
+        "pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/terminal-ime-abandoned-preedit.test.ts",
         "pnpm run test:e2e -- tests/e2e/chinese-ime-chat-input-repro.spec.ts"
       ],
       "testFiles": [
         "src/renderer/src/components/terminal-pane/terminal-ime-native-text-forwarder.test.ts",
+        "src/renderer/src/components/terminal-pane/terminal-ime-abandoned-preedit.test.ts",
         "src/renderer/src/components/terminal-pane/terminal-ime-input-source.test.ts",
         "src/renderer/src/components/terminal-pane/terminal-paste-runtime.test.ts",
         "src/renderer/src/components/terminal-pane/terminal-ime-composition-tracker.test.ts",
@@ -5716,6 +5718,15 @@
           "assertions": [
             "native text commits route once to the intended PTY",
             "composition/preedit bookkeeping does not leak premature text"
+          ]
+        },
+        {
+          "file": "src/renderer/src/components/terminal-pane/terminal-ime-abandoned-preedit.test.ts",
+          "assertions": [
+            "a Backspace that empties the preedit sends nothing to the PTY instead of replaying the abandoned preedit, whether or not the IME blanks the preedit with its own compositionupdate first",
+            "shrinking a multi-character preedit keeps composing and stays out of the PTY",
+            "real commits still route once, including a Sogou-style commit announced by an empty compositionupdate and an fcitx-style commit whose text only lands on the input event after an empty compositionend",
+            "the routeless dashboard-popout preview terminal drops the abandoned preedit too"
           ]
         },
         {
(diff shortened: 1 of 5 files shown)


OrcaWin:
Superseded by the cancelled-preedit reconciliation now on main plus #12281's explicit Sogou 
ihao → … → n → empty no-trailing-input regression. Because the fix lives inside the patched xterm CompositionHelper, it also covers routeless terminal surfaces without another listener state machine. Thank you @kcfang for the real macOS Zhuyin byte trace and the cross-dialect analysis.

Issue #10812, opened 2026-07-27
Add Claude 5-family models (Opus 5, Sonnet 5, Fable 5) to the Claude model catalog


ThomasR101:
## Summary

Anthropic's Claude 5 family (Opus 5, Sonnet 5, Fable 5 — plus Haiku 4.5) is not selectable in Orca's Claude model picker. The underlying `claude` CLI already supports these models (confirmed via `/model opus|sonnet|fable|haiku` on CLI v2.1.220), but Orca's model catalog appears to be a static list bundled into the app rather than sourced live from the CLI.

## What I found

Extracted `app.asar` from a local install (v1.4.158) and grepped `out/main/index.js` for Claude model identifiers. The bundled catalog tops out at:

```
claude-opus-4-8 / claude-opus-4.8 / claude-opus-4.8-thinking
claude-sonnet-4-6 / claude-sonnet-4.6
claude-haiku-4-5
```

No `claude-opus-5`, `claude-sonnet-5`, or `claude-fable-5` entries exist anywhere in the bundle.

I also checked whether this was just a stale-install issue on my end:
- `gh release list --repo stablyai/orca` shows `v1.4.158` as the current `Latest` stable release — no newer stable build exists yet.
- The open `v1.4.159-rc.0` pre-release changelog doesn't mention any model-catalog changes either.
- Searched existing issues/PRs for `opus-5`, `sonnet-5`, `fable-5`, `claude-5` — no matches, so this doesn't appear to be tracked yet.

## Request

Add the Claude 5-family model IDs to the model catalog/picker (and any backend/session-orchestration mapping needed to actually route requests to them), ideally sourced dynamically from the installed `claude` CLI's available models rather than a static bundled list, so future model releases don't require an Orca app update to become selectable.

## Environment

- Orca: v1.4.158 (macOS, arm64)
- Claude Code CLI: v2.1.220 (already supports Opus 5 / Sonnet 5 / Fable 5 / Haiku 4.5)

Lunivo:
Update: **#12369** implements exactly what this issue asks for — it discovers the catalog from the installed `claude` CLI per host (via a `list_models` control request) instead of hardcoding it, with the seed as fallback on CLIs too old to answer. That's the "sourced dynamically … so future model releases don't require an Orca app update" part, and it's the PR to follow from here.

#12298 (mine) only refreshes the static list and explicitly does **not** close this issue.

ThomasR101:
[...]

PR #12298, opened 2026-08-03
fix(native-chat): refresh the Claude model catalog and scope fast mode

### LINKED ISSUE #12084: [Bug]: Model picker still labels the "opus" entry as "Opus 4.8"

Lunivo:
Fixes #12084. Refs #10812.

**Related work — please read before reviewing:** #12369 (Track Claude models from the installed CLI per host) solves the same staleness problem properly, by discovering the catalog from the installed CLI instead of hardcoding it. It touches several of the same files. I'd expect that PR to supersede this one long-term; this PR is the small, shippable stopgap for the bug described below. Happy to close in its favor if #12369 lands first — see *Relationship to other open work* at the bottom.

## Why this should ship

On `main`, opening native chat against a current Claude CLI leaves the model picker with **nothing selected**, and the session ends up persisting **Fable 5** as the default — so subsequent chats silently open on Fable. Fable is Anthropic's most expensive tier ($10/$50 per MTok vs Opus 5's $5/$25), so this is a real cost issue, not just a cosmetic one.

The mechanism is the stale labels. `readClaudeSessionOptionsFromTerminalScreen` recovers the live model by substring-matching Claude's TUI header against the catalog labels. A current CLI prints:

```
Claude Code v2.1.220
Opus 5 (1M context) with high effort · Claude Max
```

`main`'s catalog labels are `Fable 5` / `Opus 4.8` / `Sonnet 5` / `Haiku` — none of which appear in that header, so the parse returns `null` and no model is ever tracked. Refreshing the labels so `Opus 5` is one of them is what makes the live model resolve again.

Worth flagging separately: `isDefault` on `CatalogModel` is currently **read by nothing** (`agent-session-option-catalog-types.ts:49` declares it; only the two catalogs set it). So the "default model" today is whatever gets persisted into `nativeChatSessionOptions`, not the cataloged default — which is how Fable ends up sticky. Making that flag load-bearing is out of scope here.

## What this changes

**Model catalog** (`src/shared/agent-session-option-catalog-claude-codex.ts`)

- Labels refreshed to what current CLIs actually print, so header recovery matches again: `Fable 5`, `Opus 5`, `Sonnet 5`, `Haiku`.
- Adds Claude's "More models" submenu entries — Opus 4.8, Opus 4.7, Opus 4.6, Sonnet 4.6 — using full model ids, since the short aliases only resolve to the latest release of each family.
- Opus 5 is marked `isDefault` in place of Sonnet 5. (Inert today, per the note above — it documents intent and takes effect if the flag is ever wired up.)

**Fast mode scoping**

- `/fast` is now only offered on Opus 5 and Opus 4.8, the models Claude gates it to. Previously it appeared where the command is a no-op. Opus 4.7's fast mode was removed upstream and now errors, so it is deliberately not offered there.

**Context window: follow the CLI, don't override it**

Orca no longer composes anything onto the model id — `claude-opus-4-8` is passed through as `claude-opus-4-8`. That means a session inherits whatever Claude Code itself resolves, which on a current CLI is the 1M-context default its own `/model` picker shows (`Opus 5 (1M context) (default)`). Mirroring the CLI is the goal; Orca has no business second-guessing it.

An earlier revision of this PR added an explicit **1M context** toggle that composed a `[1m]` suffix onto the model id. It's been removed. Probing the CLI directly showed the suffix is not uniformly available, so a cataloged toggle would have advertised a capability Orca cannot verify before launch:

| Probe | Result |
| --- | --- |
| `claude --model 'sonnet[1m]'` | accepted → `claude-sonnet-5[1m]` |
| `claude --model 'claude-opus-4-8[1m]'` (and 4.7 / 4.6) | accepted |
| `claude --model 'fable[1m]'` | resolves to plain `claude-fable-5` — suffix silently dropped |
| `claude --model 'haiku[1m]'` | `400 — The long context beta is not yet available for this subscription` |
| `claude --model 'claude-sonnet-4-6[1m]'` | `Usage credits required for 1M context` |

The last two are account-scoped, so the same toggle would work for one user and hard-fail at launch for another. #12369's discovery approach is the right way to surface these — the CLI reports its own accepted values per host, including the bracketed variants.

**Live-state recovery** (`claude-terminal-session-options.ts`)

- Unchanged in behavior; the refreshed labels are what make it work again on current CLIs.

## Visual change

Yes — the Claude model picker's list changes.

Claude model picker showing Fable 5, Opus 5, Sonnet 5, Haiku, Opus 4.8, Opus 4.7, Opus 4.6 and Sonnet 4.6

Captured against a live Claude Code session. The transcript shows real switches landing (`/model opus`, `/model claude-opus-4-8`, `/model fable`) — the previous-generation entries dispatch their full model ids because the short aliases only resolve to the latest release of each family.

## Relationship to other open work

| | Overlap |
| --- | --- |
| #12369 — Track Claude models from the installed CLI per host | **Supersedes this.** Same file (`agent-session-option-catalog-claude-codex.ts`) and several of the same tests. It replaces the hardcoded seed with per-host CLI discovery, which is the durable fix for the staleness this PR patches by hand. It also keeps `sonnet` as `isDefault`, where this PR moves it to Opus 5 — worth reconciling. |
| #12102 — drop stale version from the Claude opus picker label | **Conflicts.** Same label line, same two test files. It makes the label versionless (`Opus`) to avoid future staleness. That can't work with this PR's approach: the labels double as header-match keys, and a bare `Opus` would also substring-match the `Opus 4.8` / `4.7` / `4.6` rows this PR adds. #12369 resolves the tension properly by matching on family prefix instead. |
| #10836 — recognize custom Claude models in the picker | **Conflicts.** Same file (`claude-terminal-session-options.ts`) and test. Complementary in intent — it handles headers naming a model absent from the catalog; this one fixes headers naming a model the catalog has under a stale label. |
| #12084 — model picker still labels "opus" as "Opus 4.8" | Fixed here. |
| #10812 — add the Claude 5-family models | Refs, not fixes. The 5-family models were already reachable (the aliases resolve to them); what that issue actually asks for is dynamic discovery, which is #12369's job, not this PR's. |
| #9085, #9134 (merged) | Background — introduced the per-model session-option pickers this PR edits. |

## Review summary

- **Cross-platform (macOS/Linux/Windows):** no platform-dependent code paths. Changes are a static catalog object and a regex over terminal text. No file paths, keyboard modifiers, or shell assumptions touched.
- **SSH / remote / local:** options resolve into CLI args passed to the agent process wherever it runs; header recovery reads the PTY screen, which behaves identically for local and SSH-hosted sessions. No new local-only assumptions. (Note: which model ids a given host's CLI accepts *is* host-dependent — that's exactly what #12369 addresses and why this PR no longer composes ids.)
- **Agent / integration compatibility:** scoped to the Claude catalog entry. Codex and all other agent catalogs are untouched; no git-provider surface involved.
- **Performance:** the catalog is a module-level constant; header parsing is unchanged. No hot-path work added.
- **UI quality:** picker rows only — no new components, colors, or sizes.
- **Security:** no new process spawning, credential handling, filesystem access, or network calls.

## Tests

- `agent-session-option-catalog.test.ts` — Claude option sets stay model-scoped (`effort`, `fastMode`); previous-generation models launch by full id.
- `claude-terminal-session-options.test.ts` — header recovery for current and previous-generation models, including a `(1M context)` header resolving to the same model id it composes from.
- `native-chat-pty-session-options.test.ts`, `NativeChatComposer.test.tsx`, `NativeChatSessionOptionPickers.test.tsx`, `tui-agent-startup-session-options.test.ts` — updated for the new option ordering and default model.

Local run: `pnpm lint`, `pnpm typecheck`, and `pnpm build` pass. `pnpm test` has 6 failures, none in files this branch touches and all reproducible on `main`:

- `src/relay/agent-exec-handler.test.ts` (2) — asserts on the spawned process env; fails when `GIT_CONFIG_*` is exported in the parent shell, passes with a clean env.
- `config/scripts/check-root-directory-entries.test.mjs` (3) — `.github/scripts/check-root-directory-entries.sh` uses `declare -A`, which needs bash 4+; macOS ships bash 3.2. Passes on CI's Ubuntu runner.
- `src/main/updater.test.ts` (1) — the test expects `"Could not start the update installer. Orca remains open."` but the code sends `"Could not restart to install the update. Quit and reopen Orca, then try again."` Looks like a copy change that landed without the test update; happy to open a separate issue.

### DIFF
--- a/src/renderer/src/components/native-chat/NativeChatComposer.test.tsx
+++ b/src/renderer/src/components/native-chat/NativeChatComposer.test.tsx
@@ -295,7 +295,7 @@ describe('NativeChatComposer', () => {
 
   it('shows the model already selected in the Claude TUI when chat opens', async () => {
     mocks.getMainBufferSnapshot.mockResolvedValue({
-      data: 'Claude Code v2.1.211\r\nOpus 4.8 with medium effort · API Usage Billing',
+      data: 'Claude Code v2.1.211\r\nOpus 5 with medium effort · API Usage Billing',
       cols: 120,
       rows: 40
     })
@@ -330,7 +330,7 @@ describe('NativeChatComposer', () => {
 
   it('reads Claude state from mounted xterm while its alternate screen is active', async () => {
     mocks.getMainBufferSnapshot.mockResolvedValue({
-      data: 'Claude Code v2.1.211\r\nOpus 4.8 with high effort · stale main buffer',
+      data: 'Claude Code v2.1.211\r\nOpus 5 with high effort · stale main buffer',
       cols: 120,
       rows: 40,
       alternateScreen: true
@@ -398,7 +398,7 @@ describe('NativeChatComposer', () => {
     expect(mocks.createClaudeModelSwitchConfirmationObserver).toHaveBeenCalledWith({
       ptyId: 'pty-1',
       settings: {},
-      expectedModelLabel: 'Opus 4.8'
+      expectedModelLabel: 'Opus 5'
     })
     expect(onSwitchToTerminal).not.toHaveBeenCalled()
   })
--- a/src/renderer/src/components/native-chat/NativeChatSessionOptionPickers.test.tsx
+++ b/src/renderer/src/components/native-chat/NativeChatSessionOptionPickers.test.tsx
@@ -134,7 +134,7 @@ function model(overrides: Partial<SessionOptionDescriptor> = {}): SessionOptionD
       type: 'select',
       currentValue: 'opus',
       choices: [
-        { value: 'opus', label: 'Opus 4.8' },
+        { value: 'opus', label: 'Opus 5' },
         { value: 'sonnet', label: 'Sonnet 5' }
       ]
     },
@@ -180,17 +180,15 @@ describe('NativeChatSessionOptionPickers', () => {
         isWorking={false}
       />
     )
-    expect(screen.getByRole('button', { name: 'Model Opus 4.8' }).textContent).toContain('Opus 4.8')
-    expect(screen.getByRole('button', { name: 'Model Opus 4.8' }).textContent).not.toContain(
-      'Model:'
-    )
+    expect(screen.getByRole('button', { name: 'Model Opus 5' }).textContent).toContain('Opus 5')
+    expect(screen.getByRole('button', { name: 'Model Opus 5' }).textContent).not.toContain('Model:')
     expect(screen.getByRole('button', { name: 'Effort High · Fast' }).textContent).toContain(
       'High · Fast'
     )
     expect(
       screen
         .getByRole('button', { name: 'Effort High · Fast' })
-        .compareDocumentPosition(screen.getByRole('button', { name: 'Model Opus 4.8' })) &
+        .compareDocumentPosition(screen.getByRole('button', { name: 'Model Opus 5' })) &
         Node.DOCUMENT_POSITION_FOLLOWING
     ).not.toBe(0)
 
@@ -221,7 +219,7 @@ describe('NativeChatSessionOptionPickers', () => {
     )
     expect(
       screen
-        .getByRole('button', { name: 'Model Opus 4.8' })
+        .getByRole('button', { name: 'Model Opus 5' })
         .parentElement?.getAttribute('data-disabled')
     ).toBe('true')
     expect(
--- a/src/renderer/src/components/native-chat/claude-terminal-session-options.test.ts
+++ b/src/renderer/src/components/native-chat/claude-terminal-session-options.test.ts
@@ -48,6 +48,7 @@ describe('Claude terminal session option detection', () => {
 
     expect(readClaudeSessionOptionsFromTerminalScreen(screen)).toEqual({
       model: 'sonnet',
+      context1m: false,
       effort: 'medium'
     })
   })
@@ -60,6 +61,7 @@ describe('Claude terminal session option detection', () => {
 
     expect(readClaudeSessionOptionsFromTerminalScreen(screen)).toEqual({
       model: 'sonnet',
+      context1m: false,
       effort: 'medium'
     })
   })
--- a/src/renderer/src/components/native-chat/claude-terminal-session-options.test.ts
+++ b/src/renderer/src/components/native-chat/claude-terminal-session-options.test.ts
@@ -5,12 +5,39 @@ describe('Claude terminal session option detection', () => {
   it('reads the current model and effort from Claude header chrome', () => {
     const screen =
       '\u001b[1mClaude Code\u001b[0m v2.1.211\r\n' +
-      '\u001b[38;2;102;102;102mOpus 4.8 with high effort · API Usage Billing\r\n' +
+      '\u001b[38;2;102;102;102mOpus 5 with high effort · API Usage Billing\r\n' +
       '~/Documents/projects/orca'
 
     expect(readClaudeSessionOptionsFromTerminalScreen(screen)).toEqual({
       model: 'opus',
+      context1m: false,
       effort: 'high'
+    })
+  })
+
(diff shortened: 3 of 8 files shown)


Lunivo:
Closing in favor of #12369, which fixes the root cause instead of the symptom: it discovers the Claude catalog from the installed CLI per host rather than hardcoding it, so the labels can't go stale again. This PR was a hand-maintained refresh of the same list — worth having only until that lands.

Handing off the pieces that don't automatically come along with discovery, so they don't get lost:

1. **The bug that motivated this.** On `main`, `readClaudeSessionOptionsFromTerminalScreen` can't match a current CLI header (`Opus 5 (1M context) with high effort`) against the seed label (`Opus 4.8`), so it returns `null`, no model is tracked, and the session persists **Fable 5** as the default — later chats then silently open on the most expensive tier ($10/$50 per MTok vs Opus 5's $5/$25). #12369's family-prefix matching should fix it; it's a silent failure rather than an error, so it's worth an explicit regression test with a real `Opus 5 (1M context)` header.

2. **Fast-mode scoping.** `/fast` is gated to Opus 5 and Opus 4.8. On `main` it's offered on the `opus` row only; Opus 4.7's fast mode was removed upstream and now returns an error, so a discovered 4.7 row must not offer it. #12369 notes that discovered variant rows stay model-only until cataloged, so the seed's fast-mode assignment still matters here.

3. **`isDefault` is inert.** `CatalogModel.isDefault` (`agent-session-option-catalog-types.ts:49`) is set by both catalogs and read by nothing, so the effective default is whatever lands in persisted `nativeChatSessionOptions`. That's the mechanism behind the sticky-Fable behavior above, and it's why "mark Opus 5 as default" is not by itself a fix.

4. **Entitlement-gated `[1m]` variants.** Probed against a real CLI — `sonnet[1m]` and `claude-opus-4-8/4-7/4-6[1m]` are accepted; `fable[1m]` silently resolves to plain `claude-fable-5`; `haiku[1m]` returns `400 The long context beta is not yet available for this subscription`; `claude-sonnet-4-6[1m]` returns `Usage credits required for 1M context`. The last two are account-scoped, not host-scoped, so a variant row discovered from `list_models` can still hard-fail at launch for the same user on the same host. Full table in #12369's thread.

#12084 stays open — #12369 should be the one to close it.

PR #11769, opened 2026-07-31
fix(terminal): trust Pi CSI-u Shift+Enter on Windows (#9703)

itisvincent:
## Summary

Fixes #9703 — on Windows + Git Bash, Shift+Enter in Pi inserts a newline right after start/refocus but **submits the message** immediately after Pi runs a tool that spawns a child process, then recovers on the next pane refocus.

### Root cause

Pi enables the Kitty keyboard protocol at startup and decodes CSI-u, but `TUI_AGENT_CONFIG['pi']` never set `windowsShiftEnterEncoding`. So on Windows, Pi could only get CSI-u via the flaky live-KKP-flag path (`isKittyKeyboardActivePane`). During a tool run, something in Pi's re-render/tool path emits a reset sequence (RIS / DECSTR / alt-screen swap / `CSI < u` pop), the `TerminalKittyKeyboardModeTracker` flags drop to 0, Orca sends `Esc+CR`, and Pi reads it as plain Enter → submit. On refocus, `scanReplay` re-applies Pi's startup KKP push → flags > 0 → Shift+Enter works again. This matches the observed “works after start/refocus, breaks after a tool run” pattern.

### Fix

Set `windowsShiftEnterEncoding: 'csi-u'` for `pi`, mirroring the existing Droid fix (#7668). This routes Shift+Enter through the trusted CSI-u path (`hasTrustedWindowsCsiU`) independently of the live KKP-flag churn from tool subprocesses, so a newline stays a newline for the life of the pane.

The same class of issue was already fixed for Droid (#7620 / #7668 / #8948) and Grok (#8385); Pi was simply never given the config those fixes introduced.

## Test

New case in `terminal-windows-shift-enter.test.ts`: trusted Pi foreground evidence resolves to `'csi-u'`; launch-only evidence (no trusted foreground) stays `'alt-enter'`.

## Verification

- `pnpm typecheck` — clean.
- `pnpm exec vitest run` on the changed suites — 74/74 pass.
- `pnpm exec oxlint` + `pnpm exec oxfmt --check` on changed files — clean.

## Cross-platform / compatibility notes

- Only affects Windows hosts where Pi is the trusted foreground agent. macOS/Linux already send CSI-u and are unaffected.
- Other agents without `windowsShiftEnterEncoding` keep the legacy `Esc+CR` fallback unchanged.
- No UI/visual change. No perf impact (a record property read).

## Relationship to #10511

[...]

PR #12299, opened 2026-08-03
fix(terminal): send LF for Pi Shift+Enter to stop the CSI-u decode freeze (#10203)

### LINKED ISSUE #10203: [Bug]: Pi agent input freezes briefly on Shift+Enter newline (other CLI agents unaffected)

itisvincent:
## Summary

Fixes the **freeze** half of #10203 (and supersedes the `csi-u` encoding #11769 set for Pi). After #11769, Pi's Shift+Enter correctly inserts a newline instead of submitting, but it is **visibly slower than Alt+J** — a brief input freeze before the newline lands. That freeze is the cost of Pi decoding the CSI-u sequence (`\x1b[13;2u`, 6 bytes), versus Alt+J's fast 2-byte `\x1bj` binding.

### Fix

Send a **1-byte LF (`\n`, 0x0A)** for Pi's Shift+Enter instead of CSI-u. Pi binds Ctrl+J (0x0A) as a newline alias by default — so LF is a fast, reliable newline that:
- avoids the CSI-u decode freeze (#10203), and
- stays distinct from Enter's CR (0x0D), so it does **not** submit (#9703, already fixed by #11769's encoding + foreground detection).

Adds a new `'lf'` value to `windowsShiftEnterEncoding`, switches Pi to it, and routes it in the Shift+Enter policy. Droid stays on `csi-u`; agents without an override keep the legacy `Esc+CR` fallback unchanged.

### Why this rides on #11769

The encoding resolver still gates on `routingTrusted`, so `'lf'` only applies when the pane's foreground is trusted as Pi. #11769's Windows foreground detection (`PI_AGENT_NODE_ENTRYPOINT_RE`) is what makes that trust hold — and it covers both the npm path **and** the pnpm virtual-store path (verified below).

## Verification (from source)

- `pnpm typecheck` — clean.
- `pnpm exec vitest run` on the changed suites — 134/134 pass, including a new policy test that Shift+Enter sends `'\n'` when the encoding is `'lf'`, and an updated resolver test that Pi resolves to `'lf'`.
- `pnpm exec oxlint` + `pnpm exec oxfmt --check` on changed files — clean.
- Confirmed the foreground detection recognizes the **reporter's real pnpm Pi command line**:
  `D:\programming\SDK\node\node.exe E:\pnpm/global/5/.pnpm/@earendil-works+pi-coding-a_<hash>/node_modules/@earendil-works/pi-coding-agent/dist/cli.js ...`
  → `recognizeAgentProcessFromCommandLine` returns `{ agent: 'pi', processName: 'pi' }` (the regex matches `node_modules/@earendil-works/pi-coding-agent/dist/cli.js` at the path tail). So `routingTrusted` holds and `'lf'` applies for pnpm-installed Pi, not just npm-installed.

## Relationship to #11778

My earlier #11778 tried to fix the freeze via the IME deferred-newline path, but that was the wrong target — the freeze is CSI-u decode cost, not an IME delay. This PR is the correct fix; #11778 can be closed.

## Cross-platform / compatibility notes

- Windows-only in effect (the defer/encoding path is Windows-gated). macOS/Linux already send CSI-u where negotiated and are untouched.
- Pi-specific (only `pi` uses `'lf'`); other agents unchanged.
- No UI/visual change. No perf risk — it replaces a 6-byte sequence with a 1-byte one.
- The agent encoding lookup is still called once per Shift+Enter (refactored to a single `windowsEncoding` read), preserving the lazy-call-count behavior the existing tests assert.

## Reviewer checklist

- [x] Cross-platform: Windows fix; macOS/Linux untouched
- [x] SSH/remote/local: renderer input path; unaffected by transport
- [x] Agent/integration: Pi only; Droid stays csi-u, others unchanged
- [x] Performance: 6 bytes -> 1 byte; single encoding read
- [x] Security: no new surface

Closes #10203

X handle: @itisvincent

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.test.ts
@@ -140,6 +140,25 @@ describe('resolveTerminalShortcutAction', () => {
     ).toEqual({ type: 'sendInput', data: '\x1b[13;2u' })
   })
 
+  it('sends a literal LF for Windows panes whose agent binds the Ctrl+J newline alias (#10203)', () => {
+    // Why: Pi binds 0x0A as a newline; a 1-byte LF is a fast newline that avoids the
+    // CSI-u decode freeze vs Alt+J and stays distinct from Enter's CR (no submit).
+    expect(
+      resolveTerminalShortcutAction(
+        event({ key: 'Enter', code: 'Enter', shiftKey: true }),
+        false,
+        'false',
+        0,
+        true,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        () => 'lf'
+      )
+    ).toEqual({ type: 'sendInput', data: '\n' })
+  })
+
   it('uses CSI-u for a non-Windows PTY reached from Windows only while Kitty is active', () => {
     const getWindowsShiftEnterEncoding = vi.fn(() => 'csi-u' as const)
     const resolve = (kittyActive: boolean) =>
--- a/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.ts
@@ -182,7 +182,12 @@ export function resolveTerminalShortcutAction(
   ) {
     // Why: negotiated KKP is authoritative everywhere; trusted pane evidence also preserves Droid's Windows encoding without KKP.
     const windowsHost = isWindowsTerminalHost()
-    const hasTrustedWindowsCsiU = windowsHost && getWindowsShiftEnterEncoding?.() === 'csi-u'
+    const windowsEncoding = windowsHost ? getWindowsShiftEnterEncoding?.() : undefined
+    // Why: agents that bind a literal LF newline (e.g. Pi's Ctrl+J/0x0A alias) send a fast 1-byte LF instead of CSI-u - avoids the CSI-u decode freeze vs Alt+J (#10203) while staying distinct from Enter's CR (no submit, #9703).
+    if (windowsEncoding === 'lf') {
+      return { type: 'sendInput', data: '\n' }
+    }
+    const hasTrustedWindowsCsiU = windowsEncoding === 'csi-u'
     // Why: CSI-u is application input, not universal; without trusted Windows evidence, require active KKP negotiation.
     const canSendCsiU = hasTrustedWindowsCsiU || isKittyKeyboardActivePane?.() === true
     return { type: 'sendInput', data: canSendCsiU ? '\x1b[13;2u' : '\x1b\r' }
--- a/src/renderer/src/components/terminal-pane/terminal-windows-shift-enter.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-windows-shift-enter.test.ts
@@ -5,22 +5,23 @@ import {
 } from './terminal-windows-shift-enter'
 
 describe('resolveWindowsShiftEnterEncoding', () => {
-  it('uses CSI-u only for trusted Droid process evidence', () => {
+  it('uses CSI-u for trusted Droid evidence and for a Droid-launched pane when live trust is missing (#9703 reliability)', () => {
     expect(
       resolveWindowsShiftEnterEncoding({
         foreground: { agent: 'droid', routingTrusted: true, shellForeground: false }
       })
     ).toBe('csi-u')
-    expect(resolveWindowsShiftEnterEncoding({ launchAgentType: 'droid' })).toBe('alt-enter')
+    // Why: launch ownership keeps the encoding applied across a transient foreground-scan miss.
+    expect(resolveWindowsShiftEnterEncoding({ launchAgentType: 'droid' })).toBe('csi-u')
   })
 
-  it('uses LF for trusted Pi process evidence (fast newline, no CSI-u decode freeze) (#9703 #10203)', () => {
+  it('uses LF for trusted Pi evidence and for a Pi-launched pane when live trust is missing (fast newline, no CSI-u decode freeze) (#9703 #10203)', () => {
     expect(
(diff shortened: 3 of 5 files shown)


itisvincent:
## Runtime diagnostic confirmed the root cause

Installed the diagnostic build on Windows (v1.4.167 + a `console.warn` in the Shift+Enter policy) and captured the actual encoding resolution at runtime:

```
[S+Enter] {"enc":"csi-u","host":true,"kitty":false}   → CSI-u → newline (with freeze)
[S+Enter] {"enc":"alt-enter","host":true,"kitty":false} → Esc+CR → submit
[S+Enter] {"enc":"csi-u","host":true,"kitty":false}
```

**The encoding flickers between `csi-u` and `alt-enter`.** When `routingTrusted` drops (the foreground scan transiently misses Pi during tool subprocess churn), the resolver returns `alt-enter` → `Esc+CR` → **submit**. When trust holds, it returns `csi-u` → newline (with the CSI-u decode freeze).

`kitty` is always `false` — the Kitty keyboard protocol path never helps; the config path is the only thing that works, and it flickers.

### After applying this PR (LF + launch fallback)

```
[S+Enter] {"enc":"lf","host":true,"kitty":false}   → always LF → always a fast newline
```

The launch fallback makes the encoding always resolve to `lf` (the pane was launched as Pi), regardless of the transient scan miss. The LF encoding sends `\n` (1 byte, Pi binds as the Ctrl+J newline alias) instead of CSI-u — no submit, no freeze.

### Remaining lag vs Alt+J

Confirmed: the remaining ~10ms lag vs Alt+J is **Orca's `sendInput` dispatch overhead** (JS policy → IPC → PTY), not Pi decode cost. Alt+J bypasses the policy entirely (xterm.js writes to the PTY natively). No `[S+Enter DEFER]` appeared — the IME deferred-newline path is NOT the cause. This architectural overhead is not fixable by changing the byte; it would need an xterm-level custom key handler (a separate, deeper PR).

### #11778 (IME defer) — wrong fix, should be closed

The IME deferred-newline path was NOT firing (no `[S+Enter DEFER]` in the diagnostic). The freeze was CSI-u decode cost, not an IME delay. Closing #11778.

itisvincent:
New finding:
Orca only sends the good newline (LF) when it can see that Pi is the program running in your pane right now. When I answer you, I run   
 little helper programs (to read files, run commands). While those helpers are running, Orca sees the helper, not Pi — so it switches to
 the default "alt-enter" (which submits). And after my answer finishes, Orca doesn't look back — it keeps thinking "not Pi" until you
 click the pane again.

itisvincent:
## Closing this PR — LF approach was wrong

After runtime diagnostic + analysis of Pi's `parseKey` (in `@earendil-works/pi-tui/dist/keys.js`):

1. **LF (`\n`) has the same Kitty-flag dependency as `\x1b\r`.** Pi's `parseKey` checks `_kittyProtocolActive` for both `\n` and `\x1b\r`. When Kitty flag is OFF (after tool churn), both parse as `enter` → **submit**. So switching to LF didn't fix the submit — it just changed which byte submits.

2. **The launch-agent fallback was non-functional.** `agentLaunchConfigByPaneKey[paneKey]` is **ephemeral** — cleared on session transitions. After Orca restart (restored panes), the fallback reads `null`. So Patch B never fired.

3. **CSI-u (`\x1b[13;2u`) is the ONLY unconditional byte.** Pi's `parseKey` runs `parseKittySequence(data)` **first and unconditionally** — no Kitty-flag check. CSI-u → `shift+enter` → newline in **every** state (fresh, restored, post-churn). `\n` and `\x1b\r` both depend on the Kitty flag, which flips OFF on tool churn.

The correct fix is **title-based fallback**: when foreground detection fails (returns `alt-enter`), read the pane's OSC title (which Pi sets to `⠸ Pi` and keeps stable across restore/tool-churn/typed-shell states), recover the agent type, and send CSI-u for CSI-u agents (Pi, Droid). Raw shells (no agent title) keep the safe `alt-enter`. Filing as a separate PR.

PR #12309, opened 2026-08-03
fix(claude-accounts): stop path resolution from creating the config directory

### LINKED ISSUE #12181: [Bug]: Orca creates ~/.claude during background usage refresh even when Claude is disabled
### Operating system

macOS

### Orca version

1.4.164

#### Short summary

Orca creates an empty `~/.claude` directory even when Claude is disabled in `disabledTuiAgents`, there are no managed Claude accounts, no active Claude account, and no Claude session has been launched.

Deleting the empty directory is temporary: a later Orca background usage refresh can recreate it.

#### Steps to reproduce

1. Disable Claude in Orca's agent settings.
2. Ensure there is no active/managed Claude account and quit Orca.
3. Remove an existing empty `~/.claude` directory.
4. Start Orca and allow its deferred usage refresh to run, or focus/show/restore the Orca window.
5. Observe that `~/.claude` is created again even though Claude was not launched.

A visible Orca window also runs the rate-limit poll on the normal 15-minute cadence, so the directory can reappear later after deletion.

#### Expected behavior

Reading runtime paths or checking whether Claude usage is available should not create Claude's configuration directory. Orca should create it only when an operation actually needs to write Claude configuration or credentials.

Disabling Claude should also prevent unrelated background usage preparation from materializing Claude state.

#### Actual behavior

The default Claude runtime path resolver creates the directory as a side effect of resolving paths:

```ts
const configDir = inheritedConfigDir || join(homedir(), '.claude')
mkdirSync(configDir, { recursive: true })
```

The background call path is:

```
RateLimitService refresh
  -> ClaudeRuntimeAuthService.prepareForRateLimitFetch()
  -> getPreparation()
  -> ClaudeRuntimePathResolver.getRuntimePaths()
  -> mkdirSync(~/.claude)
```

The main window attaches usage refreshes to focus/show/restore, starts a deferred startup refresh, and runs a 15-minute polling timer. This rate-limit path is not gated by `disabledTuiAgents`.

#### Suggested fix

Make `ClaudeRuntimePathResolver.getRuntimePaths()` side-effect free: return the paths without calling `mkdirSync`. Create the directory only inside the specific credential/config mutation that requires it.

Suggested regression coverage:

1. Calling `getRuntimePaths()` with a nonexistent default config directory returns the expected paths without creating the directory.
2. Preparing a rate-limit fetch with no selected Claude account does not create `~/.claude`.
3. An actual Claude config/credential write still creates its required parent directory.

This is separate from the agent-hook config-directory issue in #11641: that issue also has a local disabled-agent removal side effect, while this `~/.claude` case originates in the Claude usage/auth path resolver.

makoto-developer:
## ELI5

Background Claude usage checks should read account paths without recreating directories. This removes the synchronous `mkdirSync` from path resolution; credential and configuration writers still create their own parent directories.

## What Changed

- Keep `ClaudeRuntimePathResolver.getRuntimePaths()` free of directory writes, including inherited `CLAUDE_CONFIG_DIR` paths.
- Preserve the existing credential path, colocated configuration preference, home configuration fallback, and environment patch.
- Cover background usage preparation, repeated lookups with existing/missing directories, and actual managed-credential creation and recreation after the directory is removed.

## Why

A usage refresh reached the resolver even with Claude disabled and no managed account, repeatedly recreating `~/.claude`. Each path lookup also paid for a redundant synchronous directory operation. The deterministic work-count regression exercises 1,000 lookups with both existing and missing directories: baseline performs 1,000 `mkdirSync` calls; this change performs zero. No timing or cache-freshness policy changes.

## Linked Issue

Fixes #12181.

## Visual Proof

N/A — backend path resolution; no rendered or interaction changes.

## Testing

Verified against `main` at `1a9a5f9bc720`, with `ORCA_BACKGROUND_LAUNCH=1`, on macOS:

- `pnpm test src/main/claude-accounts src/main/rate-limits`: **692 tests passed in 71 files**.
- `pnpm tc:node` with an 8 GB Node heap: passed.
- `pnpm run check:code-quality:changed origin/main`: passed all three changed-file gates.
- Formatting and `git diff --check`: passed.
- Restoring only the original resolver makes all five targeted regression cases fail; the four path-compatibility cases still pass.

The credential materialization test now starts without a runtime directory and removes the whole directory before a second launch. Both writes succeed and retain the expected credential contents. The existing broader suites cover account switching, readback, snapshots, file modes, and WSL selection. These local checks are not a claim of a full all-platform build/test run; PR CI covers the broader matrix.

## Review

All current resolver consumers are in the Claude runtime auth service. Reads handle absent files; credential/config writers create parent directories themselves; Keychain calls use the config path as a namespace. Path construction still uses Node path APIs. Execution-host routing, SSH/WSL commands, folder-workspace behavior, remote wire formats, and refresh intervals are unchanged.

Original implementation by @makoto-developer; maintainer follow-up preserves the contributor commits, updates the branch to current main, and strengthens regression coverage.

## AI Disclosure

The original PR disclosed Claude Code assistance. The maintainer follow-up and verification were assisted by OpenAI Codex.

## Agent skill upstream boundary

- [x] Not applicable; no upstream skill content is copied.

## Independent review follow-up

Preserve canonical macOS Keychain aliases when a missing config directory sits below a symlinked parent. Resolve the nearest existing ancestor without creating directories, and retain raw-only lookup for broken symlinks, inaccessible paths, or missing paths containing parent traversal (`..`), whose canonical credential namespace cannot safely be inferred across symlinks. This fixes a rare credential-discovery regression while keeping redundant directory writes removed. Seven real-filesystem alias cases and the existing account suite pass (224 tests across 27 files); Node typecheck and changed-code quality pass. Keychain calls are mocked; no user credentials were accessed.


Jinwoo-H (approved):
Reviewed exact candidate ff83f4b86cb56187f46a05522eebe5720cb347d4. The background auth-preparation oracle is red on current main and the explicit revert, green on this head, and legitimate credential/config writers still create their parent directories.

Issue #12308, opened 2026-08-03
[Bug] Korean IME composition issue when using Gemini CLI within Orca: characters overlap, abnormal spacing, and last character gets cut off


Gukui0516:
**Description:**
When typing Korean while using Gemini CLI within Orca, there are several issues with how the characters are rendered and processed during IME composition:
1. The newly typed character seems to overlap or cover the previously typed character, causing visual glitches and making it very uncomfortable to type.
2. Abnormal spaces are automatically inserted between characters.
3. If the spacebar is not pressed after typing a word, the last typed Korean character gets cut off or is not fully committed to the prompt.

<img width="335" height="169" alt="Image" src="https://github.com/user-attachments/assets/b6b27db4-2132-4460-94d2-22bbe3763f8d" />

<img width="442" height="167" alt="Image" src="https://github.com/user-attachments/assets/f84bfb48-a2a6-4122-827e-f5fd14d752ec" />

**Steps to Reproduce:**
1. Run Orca in a macOS terminal and start Gemini CLI.
2. Switch to the Korean IME.
3. Attempt to type Korean sentences in the prompt.
4. Observe the overlapping characters and abnormal spacing.
5. Type a Korean word and press enter without pressing the spacebar at the end, and observe the last character getting cut off.

**Environment:**
- OS: macOS (Darwin)

handlecusion:
I'm picking this up. 👋

Starting investigation now on the macOS Korean IME composition path in the terminal pane (`src/renderer/src/components/terminal-pane/terminal-ime-*`), focusing on the three reported symptoms:

1. Overlapping / covered characters during composition
2. Spurious spaces inserted between characters
3. Last Hangul syllable dropped when Enter is pressed without a trailing space

Will follow up here with findings and a PR.

handlecusion:
Update after digging into this.

**What I could not reproduce on `main`:** the three symptoms as filed. #12278, #12280, #12282 and #12284 landed roughly three hours before this issue was opened, and they cover exactly this: syllable N is now flushed to the PTY before syllable N+1 starts composing, so it is no longer stranded behind the opaque composition overlay (that overlay covering the previous syllable is what the first screenshot shows — `현` had not reached the PTY yet, so the preedit `상` was drawn in its cell).

[...]

PR #12313, opened 2026-08-03
fix(terminal): keep the key that commits an IME composition

handlecusion:
## Summary

- Stop xterm dropping the key that commits a Hangul composition, so the word-separating Space and the submitting Enter still reach the PTY when the IME reports its commit through an `insertText` input event.
- Add CJS and ESM regressions for the committing Space and Enter, plus a guard that a commit the textarea already carries is not sent twice.
- Add a headless end-to-end regression that types `한글이 밀리는 현상` and submits it through the terminal pane's real IME surface (bypass policy, composition tracker, native-text forwarder, composition route).

### The bug

`CompositionHelper` defers a keypress that lands while a composition finalizer is pending, so the commit is emitted before the key that caused it. It then reconciles the two observations:

```ts
const observedInput = this._removeAlreadySentData(
  pending.inputData || pending.keypressData,   // <-- either/or
  pending.dataAlreadySent
);
```

When the IME withdraws its preedit before `compositionend` and reports the committed syllable as a separate `insertText` — the recorded IBus Hangul shape already checked in at `tests/e2e/terminal-ime-observed-event-sequences.ts` — `inputData` holds the syllable and `keypressData` holds the committing key. `||` keeps the syllable and discards the key. Nothing else re-sends it: xterm sets `_keyPressHandled`, so the key's own `input` event is suppressed too.

The result is user-typed input silently vanishing at every commit:

| typed | reached the PTY |
| --- | --- |
| `이` + Space | `이` |
| `상` + Enter | `상` |

which reads as words running together and as a prompt that will not submit.

macOS is currently shielded by accident — `terminal-ime-native-text-forwarder.ts` swallows the IME's `insertText` while a composition transaction is pending, so `inputData` stays empty there and the merge keeps the keypress. That forwarder is installed on macOS only, so Linux and Windows IMEs take the lossy path. The fix removes the dependence on that side effect.

### The fix

Re-emit the deferred committing key after the commit it was waiting on, and only when the reconciled text does not already end with it, so the existing paths that already carry the key in the textarea (`이 `) still send it exactly once.

### Relationship to the linked issue

The three macOS symptoms in #12308 — a syllable hidden behind the composition overlay, odd spacing, and a trailing syllable lost on Enter — reproduce on builds from before #12278 / #12280 / #12284, which merged about three hours before the issue was filed. I could not reproduce them against current `main`: `terminal-ime-macos-hangul-sentence.test.ts` drives the reporter's exact sentence through the real pane IME surface, including the final-consonant transfer in `밀`→`미`+`리`, both word spaces and the submitting Enter, and every syllable flushes before the next preedit is drawn. That test is new coverage — the equivalent macOS check today is headful-only and gated behind `ORCA_E2E_NATIVE_MACOS_KOREAN=1`, so CI has never run it.

What is still broken on that path is the committing-key drop above, which produces the same two user-visible symptoms the issue describes (spacing that disappears between words, input not committed to the prompt) on the platforms the macOS forwarder does not cover.

Refs #12308

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

Ran:

- The two new files add 10 cases. 6 of them (the committing Space and Enter, in both the CJS and ESM bundles, plus the follow-up second space) fail on `main` and pass with the patch. The other 4 — the "already carries the committing key" case and the two end-to-end macOS sentence cases — pass with and without the patch, so they pin the over-correction risk and the behaviour #12278 / #12280 established rather than the fix itself.
- 226 terminal-pane test files / 2976 tests, plus the native-chat and dashboard-popout IME suites.
- Full `pnpm test`: 44270 passing. Six failures in `src/relay/agent-exec-handler.test.ts`, `src/main/updater.test.ts` and `config/scripts/check-root-directory-entries.test.mjs` reproduce identically on a clean `main` checkout and are unrelated to this change.
- `pnpm lint` (oxlint native + type-aware, reliability gates, max-lines ratchet, localization verifiers), `pnpm typecheck` (node, cli, web), `pnpm build`.
- Verified the rewritten source maps resolve `_cancelPendingFinalizer` in both bundles back to the correct line of the updated `CompositionHelper.ts`.

## Vendored xterm artifacts

The patch touches four pinned artifacts:

- `src/browser/input/CompositionHelper.ts` — the change itself.
- `lib/xterm.js`, `lib/xterm.mjs` — the same statement transposed into the pinned CJS and ESM bundles (97 generated characters each).
- `lib/xterm.js.map`, `lib/xterm.mjs.map` — re-spliced rather than reformatted: mappings were decoded to absolute segments, shifted by the new generated columns and the 13 added source lines, and re-encoded (verified to round-trip byte-for-byte before editing), then written back by replacing only the two changed JSON string literals so webpack's compact layout and esbuild's pretty layout are both preserved.

## AI Review Report

Reviewed the whole IME commit boundary, from the DOM event stream through `CompositionHelper` reconciliation to PTY bytes.

- **Cross-platform**: the change is in shared composition code with no platform branch. It is a no-op wherever `inputData` is empty, which is every path macOS takes today (the native-text forwarder swallows the IME `insertText`), and every path where the commit arrives as `insertCompositionText`. It only adds bytes on the `insertText`-commit shape that Linux/IBus and Windows IMEs produce. Windows Shift/Ctrl+Enter chords go through `terminal-ime-deferred-newline.ts`, which is untouched.
- **SSH / remote**: composition output still flows through `installTerminalImeCompositionRoute`, so the captured-transport and PTY-identity guards that drop commits after a reconnect or pane swap are unchanged. No new send path.
- **Agents / integrations**: no agent-specific or provider-specific behavior; this is raw terminal input for every CLI.
- **Performance**: two length checks and one `endsWith` per composition commit, on a path that already builds strings. No new timers, listeners or allocations in a hot loop.
- **UI quality**: no rendering change. Both new tests are headless and add ~1s to the suite.
- **Security**: no command execution, IPC, dependency, credential or path handling. The re-emitted text is the same keypress xterm had already accepted; the `endsWith` guard bounds it to at most one extra copy.

Main risk considered was double-sending the committing key when the textarea reconciliation already contains it — covered by the `endsWith` guard and by two tests that pass with and without the fix.

## Notes

X (Twitter): I don't have one to share — happy to be credited as @handlecusion on GitHub.


handlecusion:
Thanks @coderabbitai — all eight were on the tests, none on the fix. Applied four, declined three with reasons, one was already correct.

**Applied** (74d0799):

- `composition()` in the sentence test now takes the `compositionstart | compositionupdate | compositionend` union, so a typo can no longer dispatch an event nothing listens for. This also removes the divergence from the sibling file, as suggested.
- `key()` now throws when a `keypress` is given a key *name* rather than the committed character. That is the `'Enter'` → `69` trap you flagged; the call sites already pass `'\r'`, and now a future one cannot regress silently.
- The input source fixture is pinned with `satisfies MacNativeTextInputSourceFeatures`.
- Added a line stating why the first syllable commits with no tick: it models typing faster than the finalizer, and the later commits deliberately get their tick back. Making the placement uniform would delete that case.
- Also renamed the double-space test and wrote the expectation as `` `이${' '.repeat(2)}` `` so the count is explicit rather than two trailing spaces in a literal.

**Declined:**

- **`as const` on the `describe.each` table.** `pnpm typecheck` passes on all three projects without it — vitest's `describe.each` overload already infers the tuple positionally. The sibling `terminal-ime-hangul-syllable-flush.test.ts` uses the identical untyped table, so adding it here would only make the two files inconsistent.
- **Narrowing the transport stub.** `installTerminalImeCompositionRoute` takes `PtyTransport`, so a `Pick` stub still needs the same double cast to be accepted. `terminal-ime-composition-route.test.ts` already stubs it as `as unknown as PtyTransport`; matching that convention seemed better than a longer expression with no extra checking.
- **Splitting the sentence test per behaviour.** The accumulated composition state *is* the subject — the stale-preedit phase only means anything after three prior syllables and a word space have shaped `_compositionPosition`, `_compositionSuffix` and the pending finalizer. Split tests would each have to replay that prefix, so the same regression would fail in every one of them. The five intermediate `expect`s already pin which phase broke, which is the part that matters when triaging.

Full suite re-run after the changes: 226 files / 2976 tests in `terminal-pane`, `pnpm typecheck`, oxlint native and type-aware all clean.

One note for a maintainer: `PR Checks` is sitting in `action_required` because this is my first PR here, so the real CI has not run yet. Local `pnpm lint`, `pnpm typecheck`, `pnpm test` and `pnpm build` are all green — happy to push again if approving the workflow surfaces anything.

handlecusion:
Rebased onto current `main` (`a025a71`, v1.4.176-rc.0) — the only conflict was the `pnpm-lock.yaml` patch-hash entry, resolved by taking main's lockfile and re-running `pnpm install`. `main` has not touched the `@xterm/xterm` patch since this branch was cut, so the fix applies unchanged.

Re-verified on the new base: 247 terminal-pane test files / 3164 tests, `pnpm typecheck`, `pnpm lint`, `pnpm build` all green.

For context: the reporter of #12308 confirmed on v1.4.167 that the three symptoms they filed are gone and closed the issue, which matches the diagnosis in the PR description — those were fixed by #12278 / #12280 / #12284, before this branch. What is left here is the separate residual defect this PR fixes (the committing Space/Enter dropped when the IME reports its commit as its own `insertText`), plus the headless macOS sentence coverage that pins the behaviour those PRs established, which CI has never been able to run.

Still waiting on workflow approval for `PR Checks`.

handlecusion:
Closing this — the reporter of #12308 confirmed on v1.4.167 that all three symptoms they filed are gone and closed the issue, so the problem this branch set out to solve is already fixed on `main` by #12278 / #12280 / #12282 / #12284.

Leaving a note on what is in here in case a maintainer wants to pick either piece up later.

**The residual defect.** `CompositionHelper._sendPendingComposition` reconciles the deferred keypress with `pending.inputData || pending.keypressData`. When an IME withdraws its preedit before `compositionend` and reports the committed syllable as its own `insertText` — the IBus Hangul shape recorded in `tests/e2e/terminal-ime-observed-event-sequences.ts` — both fields hold *different* text, and `||` discards the key that caused the commit. `이` + Space sends only `이`; `상` + Enter sends only `상` and never submits. macOS is shielded by accident: `terminal-ime-native-text-forwarder.ts` swallows that `insertText` while a transaction is pending, so `inputData` stays empty there. That forwarder is macOS-only, so Linux and Windows IMEs take the lossy path. 6 of the 10 new cases fail on `main` and pass with the one-statement fix.

**The coverage.** `terminal-ime-macos-hangul-sentence.test.ts` types `한글이 밀리는 현상` through the pane's real IME surface — bypass policy, composition tracker, native-text forwarder, composition route — including the final-consonant transfer in `밀`→`미`+`리`, both word spaces and the submitting Enter. It passes on `main` today, so it is not tied to the fix: it pins the behaviour #12278 / #12280 established, which CI has never been able to exercise because the equivalent macOS check is headful and gated behind `ORCA_E2E_NATIVE_MACOS_KOREAN=1`.

Happy to reopen or split the test file out on its own if either is useful. Thanks @coderabbitai for the review.

Issue #11340, opened 2026-07-29
[Bug]: Tab restoration on Windows (cmd.exe) replays launch flags as literal terminal input instead of relaunching with them


asiaenv8053:
### Operating system

Windows (11 Pro, version 25H2, build 10.0.26200.8875, 64-bit, ko-KR locale, cmd.exe shell)

### Orca version

1.4.159.0

### Details

**Environment:**
- OS: Windows 11 Pro, version 25H2, build 10.0.26200.8875, 64-bit
- System/UI locale: ko-KR
- Shell: cmd.exe (Microsoft Windows [Version 10.0.26200.8875])
- Orca: 1.4.159.0 (Windows)

**Short summary:**
On Windows (cmd.exe as the terminal shell), CLI launch flags passed to an agent command (e.g. `claude --dangerously-skip-permissions --resume <session-token>`, or the equivalent for `codex`) apply correctly when a terminal tab is opened fresh. But when tabs are *restored* — e.g. after Orca or Windows restarts — the flags are not reapplied to the relaunch. Instead they appear to get replayed as literal keystrokes typed into the already-running REPL, one flag per submitted line.

**What happened?**
After restarting, a restored tab showed each launch flag echoed back as its own separate chat prompt/input, single-quoted, e.g.:

```
❯ '--resume'

✶ Gallivanting… (6s · thinking with xhigh effort)
────────────────────────────────────────────────
❯
```

Each flag (`'--dangerously-skip-permissions'`, `'--resume'`, `'<session token>'`) is submitted to the agent as if the user typed it as a chat message, instead of being applied as `argv` to the process launch. This is not limited to `--resume`/session restoration specifically — it reproduces for ordinary tab restoration with any launch flags (e.g. `--dangerously-skip-permissions` alone), not just when resuming a prior AI session.

The single-quoting around each token is a strong hint at the cause: it looks like the tab-restore path serializes the original launch command POSIX-style (space-joined, single-quoted argv tokens) and replays that string as terminal *input* rather than storing/reusing structured argv to relaunch the process. `cmd.exe` doesn't treat single quotes as quoting at all, so the literal text just gets typed into the already-started interactive process instead of being parsed as flags.

**How can we reproduce it?**
1. In Orca on Windows 11, with `cmd.exe` as the terminal shell, open a tab and launch e.g. `claude --dangerously-skip-permissions --resume <session-token>` (or the `codex` equivalent).
2. Confirm the flags apply correctly (permissions bypass active, session resumed).
[...]

PR #12321, opened 2026-08-03
fix(windows): quote agent resume for cmd.exe on cold restore

### LINKED ISSUE #12320: [Bug]: Windows cmd.exe agent session resume after reboot fails due to PowerShell-style quoting
### Operating system

Windows 11 (cmd.exe as Orca terminal shell via `terminalWindowsShell=cmd.exe`)

### Orca version

1.4.163

### Details

**Short summary:**
After restarting Orca (or rebooting Windows), auto-resume of sleeping/cold-restored agent tabs fails on `cmd.exe` terminals. The resume command is built with **PowerShell-style single quotes**, which `cmd.exe` treats as literal characters. Agent CLIs (reproduced with Grok; same path applies to Claude/Codex) reject the argv:

```
error: unexpected argument ''019fc272-80fa-7a91-80a2-9c461ef1a9da'' found
```

or:

```
error: unexpected argument ''bypassPermissions'' found
```

Session transcript data is intact under the agent home (e.g. `~/.grok/sessions/...`). Manual resume without single quotes works.

**Root cause:**
`buildAgentResumeStartupPlan` quotes resume argv via `quoteStartupArg(arg, shell)`. When `shell` is omitted on win32, `resolveStartupShell` defaults to `powershell` (single quotes).

Main-process RPC resume (`orca-runtime`) and AI Vault resume already pass `resolveLocalWindowsAgentStartupShell` / `terminalWindowsShell`.

But these **renderer cold-restore paths omit `shell`**:
1. `pty-connection.ts` → `buildColdRestoreAgentResumeStartup`
2. `sleeping-agent-session-launch.ts` → `launchSleepingAgentSession`

Observed broken line after reboot (mixed quoting when `agentCommand` was previously cmd-quoted):

```
grok \"--permission-mode\" \"bypassPermissions\" '--resume' '019fc272-80fa-7a91-80a2-9c461ef1a9da'
```

**How to reproduce**
1. Set Orca Windows terminal shell to `cmd.exe`.
2. Launch Grok (or Claude) with YOLO/permission args so a provider session id is recorded.
3. Fully quit Orca / reboot so PTYs die and tabs cold-restore.
4. Reopen Orca and open the restored worktree tabs.
5. Observe resume command fails with `unexpected argument ''...''`; agent does not resume.

**Expected**
Cold restore / sleeping resume should quote for the configured host shell (cmd → double quotes):

```
grok \"--permission-mode\" \"bypassPermissions\" \"--resume\" \"<session-id>\"
```

**Related**
- #11340 (same cmd.exe + single-quote family; that issue describes flags typed into REPL)
- Fix PR will target the two missing `shell` call sites above

**Workaround**
- Switch Windows terminal shell to PowerShell, or
- Manually run: `grok --permission-mode bypassPermissions --resume <session-id>` (no single quotes)

CountClaw:
## Summary
- Cold restore and sleeping-agent resume typed launch lines into the live terminal **without** passing the host `terminalWindowsShell` family, so win32 defaulted to PowerShell single-quote argv.
- On `cmd.exe` tabs those quotes are literal characters, so agent CLIs reject `--resume` / permission flags after reboot (e.g. `unexpected argument ''<uuid>'' found`).
- Pass `resolveLocalWindowsAgentStartupShell` into `buildAgentResumeStartupPlan` from:
  - `pty-connection.ts` cold restore
  - `sleeping-agent-session-launch.ts`
- Matches existing runtime RPC + AI Vault resume paths.
- Add cmd resume quoting regression tests (including mixed `agentCommand` + resume suffix case).

Fixes #12320
Related #11340

## Test plan
- [x] `npx vitest run src/shared/tui-agent-startup.test.ts src/shared/windows-terminal-shell.test.ts` (73 tests pass)
- [ ] Windows: set terminal shell to `cmd.exe`, launch grok/claude with permission args, fully quit Orca, reopen worktree — auto-resume should start agent TUI (no `unexpected argument ''...''`)
- [ ] Windows: same with `powershell.exe` shell — resume still works (single-quote path)
- [ ] AI Vault / runtime RPC resume paths unchanged (already passed shell)

### DIFF
--- a/src/renderer/src/components/terminal-pane/pty-connection.ts
+++ b/src/renderer/src/components/terminal-pane/pty-connection.ts
@@ -278,6 +278,7 @@ import {
   resolveTuiAgentLaunchArgs,
   resolveTuiAgentLaunchEnv
 } from '../../../../shared/tui-agent-launch-defaults'
+import { resolveLocalWindowsAgentStartupShell } from '../../../../shared/windows-terminal-shell'
 import {
   agentProviderSessionsEqual,
   isResumableTuiAgent,
@@ -4894,6 +4895,14 @@ export function connectPanePty(
         (useLiveEntry && entry ? state.getAgentLaunchConfigForStatusEntry(entry) : undefined) ??
         matchingSleepingLaunchConfig
       const resumePlatform = getColdRestoreAgentResumePlatform()
+      // Why: cold restore types the resume line into the restored tab shell.
+      // Omitting `shell` defaults win32 quoting to PowerShell, which breaks
+      // cmd.exe tabs (single-quoted resume ids become literal argv).
+      const resumeShell = resolveLocalWindowsAgentStartupShell({
+        platform: resumePlatform,
+        isRemote: false,
+        terminalWindowsShell: state.settings?.terminalWindowsShell
+      })
       const startupPlan = buildAgentResumeStartupPlan({
         agent,
         providerSession,
@@ -4910,7 +4919,8 @@ export function connectPanePty(
         ...(launchConfig?.ompResumeFilePath
           ? { ompResumeFilePath: launchConfig.ompResumeFilePath }
           : {}),
-        platform: resumePlatform
+        platform: resumePlatform,
+        shell: resumeShell
       })
       if (!startupPlan) {
         return null
--- a/src/shared/tui-agent-startup.test.ts
+++ b/src/shared/tui-agent-startup.test.ts
@@ -593,6 +593,40 @@ describe('tui agent startup plans', () => {
     expect(plan?.launchCommand).toBe("codex 'resume' 's1'")
   })
 
+  it('quotes Windows resume argv for cmd.exe when shell is cmd', () => {
+    const plan = buildAgentResumeStartupPlan({
+      agent: 'grok',
+      providerSession: { key: 'session_id', id: '019fc272-80fa-7a91-80a2-9c461ef1a9da' },
+      cmdOverrides: {},
+      agentArgs: '--permission-mode bypassPermissions',
+      platform: 'win32',
+      shell: 'cmd'
+    })
+
+    // Why: cmd.exe treats single quotes as literal characters. Resume must use
+    // double quotes (or unquoted tokens) so the CLI receives clean argv.
+    expect(plan?.launchCommand).toBe(
+      'grok "--permission-mode" "bypassPermissions" "--resume" "019fc272-80fa-7a91-80a2-9c461ef1a9da"'
+    )
+  })
+
+  it('keeps cmd-quoted agentCommand aligned with cmd resume suffix', () => {
+    const plan = buildAgentResumeStartupPlan({
+      agent: 'grok',
+      providerSession: { key: 'session_id', id: '019fc272-80fa-7a91-80a2-9c461ef1a9da' },
+      cmdOverrides: {},
+      agentCommand: 'grok "--permission-mode" "bypassPermissions"',
+      platform: 'win32',
+      shell: 'cmd'
+    })
+
+    // Regression: agentCommand from a prior cmd launch + PowerShell-default resume
+    // suffix produced mixed quoting and broke reboot restore on cmd.exe tabs.
+    expect(plan?.launchCommand).toBe(
+      'grok "--permission-mode" "bypassPermissions" "--resume" "019fc272-80fa-7a91-80a2-9c461ef1a9da"'
+    )
+  })
+
   it('honors command overrides when building POSIX resume plans', () => {
     const plan = buildAgentResumeStartupPlan({
       agent: 'codex',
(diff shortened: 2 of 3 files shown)


src/renderer/src/components/terminal-pane/pty-connection.ts
```diff
@@ -4894,6 +4895,14 @@ export function connectPanePty(
         (useLiveEntry && entry ? state.getAgentLaunchConfigForStatusEntry(entry) : undefined) ??
         matchingSleepingLaunchConfig
       const resumePlatform = getColdRestoreAgentResumePlatform()
+      // Why: cold restore types the resume line into the restored tab shell.
+      // Omitting `shell` defaults win32 quoting to PowerShell, which breaks
+      // cmd.exe tabs (single-quoted resume ids become literal argv).
+      const resumeShell = resolveLocalWindowsAgentStartupShell({
+        platform: resumePlatform,
+        isRemote: false,
+        terminalWindowsShell: state.settings?.terminalWindowsShell
```
iamthe0ne23, line 4904: Could we derive resume quoting from the same effective shell used to spawn the PTY? The connection path honors `tab.shellOverride`, while this resolver considers only the global `terminalWindowsShell`. If a live tab explicitly uses `cmd.exe`, or the global default changes after the tab is created, the replacement PTY could launch CMD while the resume command is still quoted for PowerShell.

Perhaps this could use the existing effective-shell resolution—such as `resolveWindowsShellOverride(tab?.shellOverride, state.settings?.terminalWindowsShell)`—and include coverage for a tab-level CMD override. Separately, if tab-specific shell selection is intended to survive a full application restart, `shellOverride` may also need to be included in the persisted terminal-tab schema with a round-trip test.

_Transparency: I’m a human investigating this bug with Codex. Codex found this existing PR and helped surface this concern while we reviewed the proposed fix._

src/shared/tui-agent-startup.test.ts
```diff
@@ -593,6 +593,40 @@ describe('tui agent startup plans', () => {
...
   })
 
+  it('quotes Windows resume argv for cmd.exe when shell is cmd', () => {
+    const plan = buildAgentResumeStartupPlan({
+      agent: 'grok',
+      providerSession: { key: 'session_id', id: '019fc272-80fa-7a91-80a2-9c461ef1a9da' },
+      cmdOverrides: {},
+      agentArgs: '--permission-mode bypassPermissions',
+      platform: 'win32',
+      shell: 'cmd'
```
iamthe0ne23, line 603: These assertions are useful for validating CMD quoting itself, but they call `buildAgentResumeStartupPlan` with `shell: 'cmd'` already supplied. Since the builder supported that before this change, these tests would still pass if the new call-site wiring were removed.

Could we add regression coverage around the changed cold-restore and sleeping-session paths—or extract their shell-selection logic into a small testable helper? Ideally, the test would begin with a CMD terminal setting or persisted tab override and verify that the resulting resume command uses CMD-compatible quoting. That would directly protect the behavior fixed by this PR.

_Transparency: I’m a human investigating this bug with Codex. Codex found this existing PR and helped surface this concern while we reviewed the proposed fix._

nwparker:
Independently reproduced and confirmed your diagnosis on a real Windows 11 host — thank you for the report and the fix, @CountClaw. In a `cmd.exe` pane the line the buggy path emits fails exactly as you described:

```
>codex 'resume' '0199f7a1-0000-7000-8000-000000000001'
error: unexpected argument ''0199f7a1-0000-7000-8000-000000000001'' found
```

and the cmd-quoted form parses and reaches session lookup.

I've opened #12476, which builds on your analysis and adopts your two plan-layer test vectors verbatim (including the mixed `agentCommand` + resume-suffix case). Differences there: the `{ platform, shell }` resolution is extracted into one module so the two call sites can't drift, the per-tab `shellOverride` is honored (a tab pinned to cmd.exe under a PowerShell global setting still mis-quotes otherwise), `isRemote` is computed from the execution host rather than hardcoded `false` (a remote-runtime/serve pane on a Windows client keeps `platform: 'win32'`), and there are regression tests driving the real `launchSleepingAgentSession` and cold-restore paths — your `buildAgentResumeStartupPlan` tests pass on unpatched main (68/68), so they document the plan layer without guarding the defect. Full credit to you in the PR body.

Jinwoo-H:
Thank you, @CountClaw, for isolating the real cmd.exe cold-resume quoting failure and contributing the first focused fix.

We reproduced the failure on native Windows 11 and then ran the same effective-shell oracle across five production topologies: local Git, local folder workspace, remote runtime, SSH folder workspace, and a tab-level shell override. The submitted change fixes the local globally configured cmd.exe cases, but it resolves shell authority from the local client globally: it applies cmd.exe quoting to remote/SSH execution where POSIX quoting may be required, and it misses a tab pinned to cmd.exe when the global shell is PowerShell. The 68 plan-layer tests (and the expanded 73-test slice) remain green with the production call sites disabled, so they do not guard those ownership failures.

PR #12476 now supersedes this branch with a centralized effective-shell/execution-host design, per-tab-over-global precedence, conservative fallback for nonlocal or unresolved ownership, and production-call-site coverage. We reviewed that overlap only to establish supersession; #12476 still needs its own full review before merge.

I’m closing this PR unmerged because this implementation cannot safely land alongside the superseding direction. Reopening would make sense if #12476 is withdrawn and this branch is reworked to resolve the effective per-tab shell and execution host at both production call sites, with baseline/candidate/disabled coverage for all five topologies plus a native Windows restart/restore demonstration. We appreciate the diagnosis and test vectors—they directly moved the fix forward.

PR #11782, opened 2026-07-31
fix(windows): make managed grok-hook.cmd safe when GROK_HOME is unset

Kyou12138:
## Summary

Fixes #9358 and #9941.

Managed `~/.orca/agent-hooks/grok-hook.cmd` fails with **exit code 1/255** and
`The syntax of the command is incorrect` on every Grok hook event
(`SessionStart`, `UserPromptSubmit`, tool lifecycle, …) when `GROK_HOME` is
**unset** — the default for Grok launched outside an Orca-managed terminal.

### Root cause

cmd.exe expands `%VAR:~n,m%` at **parse time**. The generated template did:

```cmd
set "ORCA_GROK_HOME=%GROK_HOME%"
if not "%GROK_HOME:~4096,1%"=="" set "ORCA_GROK_HOME="
if "%ORCA_GROK_HOME:~-1%"=="\" set "ORCA_GROK_HOME=%ORCA_GROK_HOME%."
```

When `GROK_HOME` is undefined:

1. Substring expansion leaves literal `~…` junk instead of empty.
2. The trailing-backslash guard uses `"\"` as the RHS, which **itself** is a
   cmd quote-parser syntax error even when the variable is defined.

So every managed Grok hook aborted before the `curl` POST to `/hook/grok`.

### Fix

1. Default `ORCA_GROK_HOME=` and **`if not defined GROK_HOME goto`** past all
   substring work (not a parenthesized block — those still parse-expand early).
2. Length-cap against `%ORCA_GROK_HOME:~4096,1%` only after assignment.
3. Trailing `\` detection via `findstr /r /c:"\\$"` — never `=="\"`.
4. Extract builder to `windows-grok-hook-script.ts` for testability.

### Why not just the existing open PRs?

Several open PRs guard with `if defined` but still emit
`if "%ORCA_GROK_HOME:~-1%"=="\"`, which fails as soon as `GROK_HOME` **is** set
(e.g. inside Orca, or after the user applies the documented workaround). This
PR removes that comparison entirely.

## Related Issue

Closes #9358  
Closes #9941  

## Type of Change

- [x] Bug fix (non-breaking)
- [x] Tests

## Changes Made

- `src/main/grok/windows-grok-hook-script.ts` — Windows script template
- `src/main/grok/hook-service.ts` — use the safe builder
- `src/main/grok/hook-service.test.ts` — template assertions + Windows `cmd`
  spawn tests (unset `GROK_HOME`, trailing-backslash path)

## How to Test

```bash
[...]

PR #12326, opened 2026-08-03
fix(windows): avoid grok-hook exit 255 when GROK_HOME is unset

CountClaw:
## Summary
- Grok TUI shows `global/orca-status:stop[0].hooks[0] failed with exit code 255` on Windows when `GROK_HOME` is unset (common for Orca-launched agent sessions).
- Root cause: managed `grok-hook.cmd` percent-expanded `%GROK_HOME:~N%` / `%ORCA_GROK_HOME:~-1%` on an empty home. Unset vars become bare `~N` tokens; combined with the trailing-`\` guard's `\"` comparison, cmd rewrites the line into invalid syntax and exits **255**.
- Fix: `setlocal EnableDelayedExpansion` and only apply envelope/trailing-slash handling when `GROK_HOME` is defined, using `!ORCA_GROK_HOME:~-1!` for the slash check.

## Test plan
- [x] Reproduced exit 255 with empty `GROK_HOME` before fix; exit 0 after
- [x] Empty / trailing-slash / normal `GROK_HOME` → hook + runner exit 0
- [x] `vitest run src/main/grok/hook-service.test.ts` (4 pass)
- [ ] In Orca on Windows: run a Grok turn, confirm stop hook no longer shows red exit 255

Related: #12320 (resume quoting is separate); user-visible orca-status failure.

### DIFF
--- a/src/main/grok/hook-service.test.ts
+++ b/src/main/grok/hook-service.test.ts
@@ -89,11 +89,10 @@ describe('GrokHookService', () => {
     expect(script).toContain('/hook/grok')
     if (process.platform === 'win32') {
       expect(script).toContain('%SystemRoot%\\System32\\curl.exe')
-      expect(script).toContain('set "ORCA_GROK_HOME=%GROK_HOME%"')
+      expect(script).toContain('setlocal EnableDelayedExpansion')
+      expect(script).toContain('if defined GROK_HOME set "ORCA_GROK_HOME=%GROK_HOME%"')
       expect(script).toContain('%GROK_HOME:~4096,1%')
-      expect(script).toContain(
-        'if "%ORCA_GROK_HOME:~-1%"=="\\" set "ORCA_GROK_HOME=%ORCA_GROK_HOME%."'
-      )
+      expect(script).toContain('!ORCA_GROK_HOME:~-1!')
       expect(script).toContain('--data-urlencode "grokHome=%ORCA_GROK_HOME%"')
     } else {
       // Why: payload is piped to curl via stdin (`payload@-`) so it never lands
--- a/src/main/grok/hook-service.ts
+++ b/src/main/grok/hook-service.ts
@@ -118,14 +118,14 @@ function getManagedScript(target: 'local' | 'posix' = 'local'): string {
   if (target === 'local' && process.platform === 'win32') {
     return [
       '@echo off',
-      'setlocal',
+      // Why: delayed expansion so trailing-`\` guard never percent-expands an empty
+      // GROK_HOME into invalid `\"` syntax (exit 255 → orca-status:stop red X).
+      'setlocal EnableDelayedExpansion',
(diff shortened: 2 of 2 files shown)


nwparker:
Thanks @CountClaw — independently reproduced your exact failure on a real Windows 11 host and confirmed your diagnosis (undefined `%VAR:~N%` leaves a bare `~N` token, which collides with the trailing-`\` comparison and aborts the batch at parse time with 255).

I built your script verbatim and verified it does fix the crash: unset/empty/normal/trailing-slash/over-envelope all exit 0.

One thing surfaced while cross-checking, which is why I opened #12479 rather than piling onto this branch: `setlocal EnableDelayedExpansion` applies to *every* percent-expanded value on the curl line, so any `!` in a value is consumed as a delayed reference. Both scripts, same host, same HTTP listener:

```
# GROK_HOME=C:\Users\neil\a!b!c
this-pr  exit=0 grokHome=[C:\Users\neil\ac]        <- corrupted
#12479   exit=0 grokHome=[C:\Users\neil\a!b!c]

# ORCA_PANE_KEY=tab!1:pane!2  ORCA_WORKTREE_ID=repo::C:/Users/neil/orca/workspaces/orca/wip!fix
this-pr  exit=0 paneKey=[tabpane:C:/Users/neil/orca/workspaces/orca/wip\Users\neil\.grok] worktreeId=[null]
#12479   exit=0 paneKey=[tab!1:pane!2] worktreeId=[repo::C:/.../wip!fix]
```

`!` is legal in a Windows directory name, so that trades a loud 255 for a silently mangled pane key and a dropped `worktreeId`. #12479 instead jumps *past* the substring lines (`if not defined ORCA_GROK_HOME goto :orca_grok_home_ready`) so no delayed expansion is needed and `!`/`%` values stay byte-exact. Your analysis and repro cases are credited in that PR — it would not exist without this report.

brennanb2025:
Closing this as superseded — with thanks, because your diagnosis was the useful part here.

You correctly identified that `%GROK_HOME:~N%` is percent-expanded while cmd.exe *reads* the line, so an unset `GROK_HOME` turns the guard into a syntax error and the hook dies with exit 255. That analysis is what made the fix possible, and #12479 credits you for it explicitly.

The fix shipped via #11782 and was hardened in #12479, which takes a different route to the same goal: it jumps past the unsafe lines with a `goto` rather than enabling delayed expansion. We verified that rebasing this branch onto current main and resolving correctly leaves a zero-line diff — the behaviour you were after is already on main.

One thing worth knowing for next time, since it is genuinely subtle: enabling delayed expansion has a side effect on this path. `!` is legal in Windows paths, and delayed expansion consumes it in percent-expanded values on the curl line — a `paneKey` like `tab!1:pane!2` posts as garbage and `worktreeId` posts as null, which turns a loud crash into silent misrouting. That is why the merged fix avoids it.

Thanks for the report and the analysis.

PR #2418, opened 2026-05-20
Fix Codex Shift+Enter on Windows

Jinwoo-H:
﻿## Summary

Fixes #2412.

On Windows, Orca now sends `Esc+Enter` (`\x1b\r`) for bare `Shift+Enter` in terminal panes instead of the CSI-u `\x1b[13;2u` sequence that Codex ignores inside PowerShell. macOS and Linux keep the existing CSI-u behavior.

## Screenshots

No visual change.

## Testing

- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`
- [x] Added or updated high-quality tests that would catch regressions, or explained why tests were not needed

Additional verification run:

- `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/components/terminal-pane/terminal-shortcut-policy.test.ts src/renderer/src/lib/pane-manager/pane-lifecycle.test.ts`
- `pnpm exec electron-vite build --mode e2e`
- `SKIP_BUILD=1 pnpm exec playwright test tests/e2e/terminal-shortcuts.spec.ts --config=tests/playwright.config.ts --project=electron-headless -g "Shift\+Enter writes" --reporter=line`
- `pnpm run typecheck:web`
- `git diff --check`
- Manual node-pty repro with real `codex-cli 0.132.0` under Windows PowerShell: old `\x1b[13;2u` produced `helloworld`; new `\x1b\r` rendered `hello` and `world` on separate composer lines.

Note: the broader `all terminal chords reach the PTY or fire their action` Playwright spec still closes the Electron page later during pane action checks in this Windows environment. The new focused regression covers the changed Shift+Enter behavior directly.

## AI Review Report

Reviewed the terminal keyboard routing from the window-level handler through `resolveTerminalShortcutAction` to PTY writes. Main risks checked:

- Windows-only behavior is gated by `navigator.userAgent.includes('Windows')`.
- macOS/Linux keep the previous `\x1b[13;2u` sequence.
- `Cmd/Ctrl+Shift+Enter` pane expand remains separate from bare `Shift+Enter`.
- Remote/SSH input still uses the same PTY send path; only the bytes for this platform-specific chord change.
- Existing kitty keyboard advertisement remains enabled for non-Windows enhanced key sequences.

Added unit and Playwright coverage for the Windows byte sequence and preserved non-Windows expectations.

## Security Audit

[...]

PR #12331, opened 2026-08-03
fix(terminal): stop Ctrl+Enter leaking CSI-u without kitty keyboard

### LINKED ISSUE #12329: [Bug]: Ctrl+Enter leaks literal [13;5u in 1.4.167 Windows ConPTY terminal
## Environment

- Windows 11 Pro, build 22631
- Orca 1.4.167
- Codex CLI 0.146.0
- Microsoft Korean IME (2-beolsik)
- Orca integrated terminal (xterm.js + local Windows ConPTY; not Windows Terminal)

## Reproduction

1. Open a Codex session in Orca's integrated terminal.
2. Focus the actual xterm `Terminal input`.
3. Press Ctrl+Enter.
4. The literal text `[13;5u` is inserted. Repeating the chord repeats that string, and subsequent Korean input can exhibit repeated/unexpected newlines.

## Root cause found in the 1.4.167 renderer bundle

In `out/renderer/assets/terminal-shortcut-policy-BmRK03X0.js`, Shift+Enter checks whether CSI-u/Kitty keyboard handling is active before emitting CSI-u:

```js
data: isWindowsTerminalHost() && getWindowsShiftEnterEncoding?.() === "csi-u" || isKittyKeyboardActivePane?.() === true
  ? "\x1B[13;2u"
  : "\x1B\r"
```

Ctrl+Enter does not perform that check and always emits CSI-u:

```js
data: "\x1B[13;5u"
```

On local native Windows ConPTY, Orca does not have an active Kitty keyboard protocol for this pane, and Codex receives/prints the raw sequence instead of decoding it as Ctrl+Enter.

## Local validation

I changed the Ctrl+Enter branch locally to:

```js
data: isKittyKeyboardActivePane?.() === true ? "\x1B[13;5u" : "\r"
```

After repacking `app.asar`, restarting Orca, and testing the real xterm `Terminal input`, Ctrl+Enter produced a normal CR and no new `[13;5u` leakage. This confirms that the unconditional Ctrl+Enter CSI-u emission is the remaining defect.

## Expected

Ctrl+Enter must not emit CSI-u unless the active terminal application has negotiated/enabled the corresponding keyboard protocol. In a native Windows ConPTY pane without KKP, it should use a safe fallback and never inject raw `[13;5u` into user text.

Related: #8385. My earlier follow-up there incorrectly tested the native chat composer rather than the integrated terminal; I will correct that comment and link this issue.

ShreyanshVaibhaw:
## Summary

Fixes #12329.

Pressing Ctrl+Enter in a local Windows ConPTY pane inserted the literal text `[13;5u` into the prompt instead of delivering the chord.

Ctrl+Enter emitted the kitty CSI-u sequence unconditionally:

```ts
return { type: 'sendInput', data: '\x1b[13;5u' }
```

CSI-u is application input. It only reads as a chord while the kitty keyboard protocol is negotiated for that pane, and on a local Windows ConPTY pane it is not, so the TUI printed the escape verbatim.

Shift+Enter, three lines above, already guards exactly this and falls back to `\x1b\r`:

```ts
const canSendCsiU = hasTrustedWindowsCsiU || isKittyKeyboardActivePane?.() === true
return { type: 'sendInput', data: canSendCsiU ? '\x1b[13;2u' : '\x1b\r' }
```

Ctrl+Enter had no such guard, and its own comment acknowledged the gap: *"no Windows fallback yet (#2418)"*.

This gates the sequence on active kitty keyboard and otherwise sends the bare CR that xterm.js would have collapsed Ctrl+Enter to. That restores the pre-existing default rather than inventing an encoding, and matches the fallback the reporter validated locally.

### Behavior change worth reviewing

The rewritten test replaces an assertion that documented the opposite assumption:

> Windows uses the same kitty sequence for now: no TUI is known to treat the CSI-u Ctrl+Enter form as inert (cf. the Shift+Enter Codex-on-PowerShell case).

#12329 is the counter-example, with Codex 0.146.0 on ConPTY. So this is a deliberate reversal, not an oversight being patched. It narrows Ctrl+Enter only where the chord was never being delivered anyway: with kitty keyboard active the CSI-u path is unchanged.

I used `isKittyKeyboardActivePane` alone rather than also reusing Shift+Enter's `getWindowsShiftEnterEncoding` signal. That setting is specifically about the Shift+Enter encoding, and borrowing it to authorize a different chord would couple two unrelated decisions.

## Screenshots

No visual change.

## Testing

- [x] `pnpm typecheck`
- [x] Added regression tests

- `src/renderer/src/components/terminal-pane/`: **2887 passed, 218 files, 0 failed**
- New `sends a bare CR for Ctrl+Enter when kitty keyboard is not active` covers Windows, non-Windows, and the unset-callback case.
- The existing CSI-u expectation is kept for the kitty-active path on both platforms, so the chord is still pinned where it works.
- **Confirmed to fail against the previous implementation**: reverting the source fails exactly the new case and nothing else.

### Verification caveat

I do not have Codex on a Korean IME Windows setup and did not reproduce the literal `[13;5u` first-hand. The diagnosis in #12329 is unusually precise (the reporter located the branch in the shipped renderer bundle and validated a fallback locally), and it matches the source exactly. Verified here by tests rather than by observation.

## AI Review Report

- **Cross-platform.** No platform branch added. The guard is the kitty-keyboard capability, not an OS check, so macOS and Linux panes with negotiated KKP keep the CSI-u chord and those without get the CR they would have had from xterm.js anyway.
- **Does not disturb Shift+Enter.** That branch is untouched, including its `getWindowsShiftEnterEncoding` trusted-Windows path.
- **Fallback choice.** `\r` rather than Shift+Enter's `\x1b\r`: alt-enter means newline, whereas Ctrl+Enter collapsing to CR is the documented pre-interception xterm.js behavior, so CR is the conservative restore.
- **Flagged and addressed.** The review flagged that the existing test called `resolveTerminalShortcutAction` positionally with 5 args and could not reach the 8th parameter. The test now uses a small helper so the kitty callback is actually exercised instead of left `undefined` in every case.

## Security Audit

- **Input handling.** This strictly reduces what is written to the PTY: an escape sequence is replaced by a single carriage return on the path where the escape was being printed as text. No new input is trusted or forwarded.
- **Command execution, path handling, auth, secrets, IPC, dependencies.** Untouched. No dependency changes.
- Worth noting the pre-fix behavior injected attacker-irrelevant but user-visible escape text into a prompt; narrowing it removes a small correctness hazard where a TUI could receive unintended literal characters.

## Notes

- **Pre-existing lint failure.** `pnpm lint` fails at `verify:skill-bundle-manifest` with stale `resources/skills/*.json`. Reproduces on a clean `main`, so it is not introduced here, and those are generated artifacts I did not want to churn.
- **`pnpm test` and `pnpm build` not completed locally.** A clean `main` on this Windows machine fails 167 tests across 57 files, so a green local suite is not achievable either way. The affected area is fully green.
- **Related but not addressed.** #12329 also mentions subsequent Korean input showing repeated newlines. That symptom belongs to the IME composition path and is handled separately in #12187; this PR deliberately only fixes the literal `[13;5u` leak.
- **#2418** is referenced by the comment this PR replaces and may be closeable as a result. I have not verified its current state.

### DIFF
--- a/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.test.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.test.ts
@@ -270,27 +270,32 @@ describe('resolveTerminalShortcutAction', () => {
     expect(getWindowsShiftEnterEncoding).toHaveBeenCalledTimes(2)
   })
 
-  it('forwards Ctrl+Enter as the kitty CSI-u chord so TUIs can cue instead of send', () => {
-    // Why: xterm.js collapses Ctrl+Enter to a bare CR; intercept upstream and
-    // emit the kitty sequence (modifier code 5 = Ctrl) so probing TUIs receive
-    // the distinct chord on every platform.
-    expect(
-      resolveTerminalShortcutAction(event({ key: 'Enter', code: 'Enter', ctrlKey: true }), true)
-    ).toEqual({ type: 'sendInput', data: '\x1b[13;5u' })
-    expect(
-      resolveTerminalShortcutAction(event({ key: 'Enter', code: 'Enter', ctrlKey: true }), false)
-    ).toEqual({ type: 'sendInput', data: '\x1b[13;5u' })
-    // Windows uses the same kitty sequence for now: no TUI is known to treat the
-    // CSI-u Ctrl+Enter form as inert (cf. the Shift+Enter Codex-on-PowerShell case).
-    expect(
-      resolveTerminalShortcutAction(
-        event({ key: 'Enter', code: 'Enter', ctrlKey: true }),
-        false,
-        'false',
-        0,
-        true
-      )
-    ).toEqual({ type: 'sendInput', data: '\x1b[13;5u' })
+  const ctrlEnter = (isWindows: boolean, kittyActive?: boolean) =>
+    resolveTerminalShortcutAction(
+      event({ key: 'Enter', code: 'Enter', ctrlKey: true }),
+      false,
+      'false',
+      0,
+      isWindows,
+      undefined,
+      undefined,
+      kittyActive === undefined ? undefined : () => kittyActive
+    )
+
+  it('forwards Ctrl+Enter as the kitty CSI-u chord while kitty keyboard is negotiated', () => {
+    // Why: xterm.js collapses Ctrl+Enter to a bare CR; intercept upstream and emit
+    // the kitty sequence (modifier code 5 = Ctrl) so probing TUIs receive the chord.
+    expect(ctrlEnter(false, true)).toEqual({ type: 'sendInput', data: '\x1b[13;5u' })
+    expect(ctrlEnter(true, true)).toEqual({ type: 'sendInput', data: '\x1b[13;5u' })
+  })
+
+  it('sends a bare CR for Ctrl+Enter when kitty keyboard is not active', () => {
+    // Regression: CSI-u is application input. On a local Windows ConPTY pane with no
+    // negotiated KKP the TUI printed the escape literally, leaking `[13;5u` into the
+    // prompt. Fall back to the CR xterm.js would otherwise have sent.
+    expect(ctrlEnter(true, false)).toEqual({ type: 'sendInput', data: '\r' })
+    expect(ctrlEnter(false, false)).toEqual({ type: 'sendInput', data: '\r' })
+    expect(ctrlEnter(true)).toEqual({ type: 'sendInput', data: '\r' })
 
     // Modifier combos that are NOT plain Ctrl+Enter must keep falling through.
     expect(
--- a/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.ts
+++ b/src/renderer/src/components/terminal-pane/terminal-shortcut-policy.ts
@@ -195,8 +195,13 @@ export function resolveTerminalShortcutAction(
     !event.shiftKey &&
     event.key === 'Enter'
   ) {
-    // Why: xterm.js collapses Ctrl+Enter to a bare CR, so forward kitty CSI-u (modifier 5 = Ctrl) so the chord reaches TUIs; no Windows fallback yet (#2418).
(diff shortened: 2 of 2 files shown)


nwparker:
Thanks @ShreyanshVaibhaw — independently reproduced and confirmed your diagnosis is exactly right. I built a fix before seeing this PR and we converged on the identical source change (gate on `isKittyKeyboardActivePane?.() === true`, fall back to `\r`), including the same reasoning for *not* reusing `getWindowsShiftEnterEncoding` to authorize a different chord.

Real-host confirmation on a Windows 11 local ConPTY pane (Git Bash), writing exactly the bytes the policy emitted:

```
$ orca terminal send --terminal term_799db70f-... --text $'\x1b[13;5u'
  "send": { "accepted": true, "bytesWritten": 7 }

# readline prompt:
$ ;5u
# with cat -v running in the same pane:
^[[13;5u
```

So the escape really does arrive as literal application input, as you described. With the fallback byte the same prompt executes cleanly.

One gap worth flagging: this PR updates the unit test but not the two e2e specs that still pin the old bytes — `tests/e2e/terminal-shortcuts.spec.ts:540` (`pressAndExpectWrite(..., 'Control+Enter', '\x1b[13;5u')` in a plain, non-negotiated shell pane) and both Ctrl+Enter fixtures in `tests/e2e/korean-ime-terminal-shift-enter-commit.spec.ts`. Those would fail e2e CI against this source change.

I've opened #12462 with the same fix plus those e2e updates, a test for the dashboard-popout preview call site (which shares the policy), a platform × kitty-state matrix, and the real-host evidence above. Full credit to you for the find and the diagnosis — it's cited in the PR body.

AmethystLiang:
Duplicate of #12462 (fixes #12329).

Same root cause and intended fix: stop Ctrl+Enter from emitting kitty CSI-u into ConPTY when the pane has not negotiated kitty keyboard. Please continue review/landing on #12462. If this branch has a meaningfully different implementation detail worth cherry-picking, call it out on #12462.

PR #12337, opened 2026-08-03
fix(ai-vault): fold trailing-slash and NFC/NFD variants in folder group key

YuriNachos:
## Summary

`groupAiVaultSessions` in `src/shared/ai-vault-session-filters.ts` groups Agent Session History rows by folder, keying a `Map` with `getFolderGroupKey(session.cwd)`. That key was built with `normalizeRuntimePathSeparators(path).toLowerCase()` — a helper that backslash-folds and collapses repeated slashes but **does not trim trailing slashes and does not NFC-fold Unicode**.

So two sessions whose `cwd` differs only by a trailing slash (`/Users/ada/Project` vs `/Users/ada/Project/`) mapped to two distinct keys, while `folderLabel` (which does `split('/').filter(Boolean)`) produced the **identical** label `'ada/Project'` for both — yielding two visually identical duplicate groups that split one folder's sessions between them. The same split hit NFC vs NFD spellings of a non-ASCII folder (macOS file pickers/on-disk names yield NFD while agents record cwd in NFC — the sibling `normalizeRuntimePathForComparison` docstring in `cross-platform-path.ts` cites exactly this as the #10832 cwd-mismatch bug).

## Fix

Route the folder group key through the existing comparison normalizer, keeping the all-platforms case-fold:

```ts
return pathValue ? normalizeRuntimePathForComparison(pathValue).toLowerCase() : 'unknown'
```

`normalizeRuntimePathForComparison` (same `/shared` tree) already does `.normalize('NFC')` + `trimRuntimePathTrailingSlash` + (Windows-only) lowercase + WSL UNC alias fold. The trailing `.toLowerCase()` preserves the case-fold for POSIX paths (the comparison normalizer only lowercases Windows paths). The sibling `isAiVaultSessionInWorkspacePath` in this very file already routes through the same normalizer for comparison — the grouping-key path was simply missed.

## Screenshots

No visual change beyond no-more-duplicate folder groups in Agent Session History.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`

Passed:
- New regression tests in `ai-vault-session-filters.test.ts`: (1) trailing-slash variants of the same folder collapse to one group; (2) NFC and NFD spellings of the same non-ASCII folder collapse to one group.
- **RED-before-GREEN, isolated:** reverting `getFolderGroupKey` to the old bare-separator normalizer makes exactly these 2 new tests fail (`2 failed | 7 passed`); the 7 pre-existing tests stay green (no regression); re-applying the fix → 9/9 green.
- `pnpm typecheck` green across all three node tsconfigs; `pnpm exec oxlint` clean; `pnpm exec oxfmt --check` clean on both files.

## AI Review Report

- **Cross-platform (macOS/Linux/Windows):** the fix reuses the existing cross-platform comparison normalizer (`normalizeRuntimePathForComparison`), which already handles Windows drive/UNC paths, POSIX, and WSL. POSIX case-folding is preserved via the trailing `.toLowerCase()`. Behavior change (intentional, more correct): the two WSL UNC aliases (`//wsl.localhost/X` and `//wsl$/X`) now group together where they previously did not — matching the documented semantics of that normalizer.
- **SSH/remote/local + folder-workspace:** N/A — pure string grouping of already-stored cwd values; no git binary / SSH / folder-workspace path involved.
- **Provider/agent neutrality:** Agent Session History (shared by desktop + mobile).
- **Performance / hot path:** none — runs once per session on group.
- **UI quality:** removes duplicate identical-labeled folder groups.

## Security Audit

Pure string-normalization on already-stored cwd strings. No new input handling, IPC, auth, path, or dependency surface.

## Notes

- Sibling `normalizeRuntimePathForComparison` (`cross-platform-path.ts:25-41`) documents the NFC/NFD cwd mismatch as the #10832 bug it fixed for comparison; `isAiVaultSessionInWorkspacePath` in this same file (~263) already uses it. `getFolderGroupKey` was the missed call site.
- This module lives in `/shared` so the mobile package reuses it via Metro — the fix applies to mobile Agent Session History too; tests live in the shared test file.

### DIFF
--- a/src/shared/ai-vault-session-filters.test.ts
+++ b/src/shared/ai-vault-session-filters.test.ts
@@ -89,6 +89,61 @@ describe('/shared ai-vault-session-filters (lifted core)', () => {
     expect(groups.map((group) => group.label).sort()).toEqual(['packages/ui', 'repo/app'])
   })
 
+  it('groups trailing-slash variants of the same folder as one group', () => {
+    // The folder key must trim trailing slashes so a cwd recorded with and without
+    // one collapse to a single group instead of two visually identical labels.
+    const trailingSlash: AiVaultSession = {
+      ...baseSession,
+      id: 'trailing',
+      cwd: '/Users/ada/repo/app/'
+    }
+    const groups = groupAiVaultSessions([baseSession, trailingSlash], 'folder')
+    expect(groups).toHaveLength(1)
+    expect(groups[0].label).toBe('repo/app')
+    expect(groups[0].sessions.map((session) => session.id).sort()).toEqual(['claude:1', 'trailing'])
+  })
+
+  it('groups NFC and NFD spellings of the same non-ASCII folder as one group', () => {
+    // macOS file pickers/on-disk names yield NFD while agents record cwd in NFC (#10832);
+    // the comparison-key normalizer folds them, so both sessions land in one group.
+    const nfc: AiVaultSession = { ...baseSession, id: 'nfc', cwd: '/Users/ada/repo/Caf\u00E9' }
+    const nfd: AiVaultSession = { ...baseSession, id: 'nfd', cwd: '/Users/ada/repo/Cafe\u0301' }
+    const groups = groupAiVaultSessions([nfc, nfd], 'folder')
+    expect(groups).toHaveLength(1)
+    expect(groups[0].sessions.map((session) => session.id).sort()).toEqual(['nfc', 'nfd'])
+  })
+
+  it('groups case-variant POSIX spellings of the same folder as one group', () => {
+    // getFolderGroupKey lowercases on top of the POSIX-branch normalizer (which does not),
+    // locking the all-platforms case-fold the comment on it promises.
+    const lower: AiVaultSession = { ...baseSession, id: 'lower', cwd: '/Users/ada/repo/app' }
+    const mixed: AiVaultSession = { ...baseSession, id: 'mixed', cwd: '/Users/ada/repo/App' }
+    const groups = groupAiVaultSessions([lower, mixed], 'folder')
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks @YuriNachos — I reproduced this independently and implemented a fix blind to your PR, then compared. We converged on the same root cause and the same core change (route the group key through `normalizeRuntimePathForComparison`), and I pulled in your WSL UNC alias test vector, which I had missed.

Merged result is at #12458. Two reasons I kept mine as the base rather than just endorsing yours:
- the blanket `.toLowerCase()` is retained in yours, which still merges genuinely distinct case-sensitive POSIX folders on Linux/SSH hosts — the rest of `cross-platform-path.ts` case-folds only Windows-flavored paths;
- only one of the two folder-key builders is patched; the renderer's `folderProject()` key stays un-lowercased, so `project`-mode grouping can still split one folder into two identically labelled headers.

Your diff is smaller and cleaner for the reported symptom though — if maintainers prefer that, I'm glad to close mine.

YuriNachos:
Thanks @nwparker — agreed on both points:

- the blanket `.toLowerCase()` does merge genuinely-distinct case-sensitive POSIX folders on Linux/SSH (the rest of `cross-platform-path.ts` only case-folds Windows-flavored paths, so the group key shouldn't either);
- `folderProject()` is the second folder-key builder I missed — project-mode grouping would still split one folder into two identically-labelled headers.

Your #12458 patches both builders and drops the over-broad lowercase, so it's the more complete fix. Closing this in favour of #12458 — glad the WSL UNC alias test vector helped. Thanks again for the independent repro and the cross-check.

Issue #7094, opened 2026-07-02
[Bug]: Android mobile terminal drops characters when typing moderately fast


hopecommon:
## Operating system

Android / Honor Magic 5 / MagicOS 10

## Orca version

- Orca Mobile Android app: likely 0.0.18
- Latest Android release I found: `mobile-android-v0.0.19` pre-release
- Desktop/server runtime observed: 1.4.116 / Linux headless `orca serve`

## Details

Short summary:

On Orca Mobile for Android, terminal input appears to randomly drop characters when typing at a moderate or fast speed, especially with English text/commands.

What happened?

When I type into a terminal from Orca Mobile on Android, if I type only very slowly, the input mostly works. But when I type a little faster, some characters appear to be missing from the terminal. The result looks like random character loss rather than just delayed display.

The SSH/network connection may not be perfect and the remote runtime can have noticeable latency, but I would still expect typed characters to be queued and delivered in order. A slow connection can make input laggy, but it should not silently drop characters.

This makes mobile terminal use unreliable for even short commands, because I have to type slowly and carefully, then visually inspect whether any characters were lost.

How can we reproduce it?

1. Open Orca Mobile on Android.
2. Connect to a remote/headless Linux Orca runtime.
3. Open a terminal session.
4. Type an English command or sentence at a normal/moderate phone typing speed.
5. Compare the text that appears in the terminal with what was typed.
6. Repeat with faster typing; missing characters become more noticeable.

Example affected usage:

- typing shell commands
- typing English prompts into a CLI agent
- editing text in an interactive terminal session

Anything else that might help:

This may be related to mobile terminal input buffering, key event batching, composition handling, or remote terminal transport under latency/backpressure.

Related but not identical issues I found:

- #5250: mobile typing `--` became `-`
- #6513: synthesized/dictated text dropped in terminal panes
- #4500: terminal input silently locks after a mobile driver disconnects
- #7000 / #6995 / #7011: Android mobile terminal IME composition/input issues
- #6972: Android terminal input mode resets on reentry

This issue is specifically about normal typed characters appearing to be dropped when typing a bit faster on Android mobile terminal.

nwparker:
## Repro status: **env-blocked** (no `cannot_repro` label)

Could not device-reproduce: {why}.

Tree: `c3a9a5e8a6`. Code review alone is not conclusive for the user-facing failure. No `cannot_repro` label applied per env-blocked policy.

hopecommon:
[...]

PR #12341, opened 2026-08-03
fix(mobile): derive terminal input from native replacement ranges

nwparker:
## Summary

- derive mobile terminal commits from native replacement text and ranges
- suppress composition-owned submit/accessory actions until native composition ends
- serialize transport writes without optimistic buffer advancement after a rejected send
- delete the Hangul mirror, pending-flush state, accessory commit adapter, and their workaround tests

Net change: 1,216 additions, 1,797 deletions.

## Native boundary

iOS extends React Native's existing `TextInput` change payload with UIKit's marked-text state and `shouldChangeTextInRange` replacement evidence. Android cannot use the source patch because Gradle consumes React Native's precompiled AAR, and public `TextInput.onChange` exposes neither composition state nor replacement ranges.

The Android package therefore contains the five React Native plumbing roles required for one native view—package, manager, `ReactEditText`, `InputConnection`, and non-coalescing event—in one 214-line Kotlin file. There is no locale classification, mutable-buffer diff, timer, recovery tracker, or application composition state machine.

## Evidence

- Native iOS simulator Pinyin trace records preedit, exact `中` commit range, cancellation, ordinary `a`/`b`/`c`, composition-owned Enter suppression, and post-composition accessory Enter.
- Native iOS Japanese Kana records tap, real `か`→`き` flick, `かな`→`カナ` candidate replacement, Backspace cancellation, active-composition accessory Enter suppression, and post-composition Enter.
- Native iOS Japanese Romaji records `k/a/n/a` marked transitions followed by one `かな` commit.
- Exact terminal output contains only the committed Japanese payloads, in order; no preedit reaches the shell.
- English QWERTY `a`/`b`/`c` remains immediate with native insertion ranges.
- Bypassing the composing guard fails the Kana accessory-Enter positive while the English negative passes.
- Android Fcitx Hangul commits exact `한글`, cancellation emits nothing, and the posted unmarked boundary mutation suppresses only the next accessory Enter while ordinary `a` remains exact.
- Android Fcitx Anthy keeps `s+a` preedit native, confirms exact `さ`, cancels without bytes, and sends Enter separately. Marking only `setComposingText` uncomposed leaks exact `sa` while English `a` remains unchanged.
- With 750 ms delayed RPC ACKs, rapid Android `abcd` remains exact and max-one-in-flight; slow `ijkl` and zero-delay rapid `mnop` controls remain exact. Removing the sender wait starts all queued writes concurrently.
- Buffered Android autocorrection is platform-owned: opt-in Gboard correction submits exact `the ` once, while the disabled control submits exact `teh ` once.
- Five production workaround modules and their tests are deleted.

## Verification

- mobile Vitest: 385 files, 2,847 passed, 3 skipped
- focused native-range/mock-contract tests: 13/13 passed
- mobile TypeScript, oxlint, and formatting
- iOS simulator source build succeeds with the patched React Native source
- Expo Android prebuild autolinks `:orca_react-native-terminal-input`
- Android `assembleDebug` succeeds with JDK 21; clean APK SHA-256 is `97035eecc1d5aa64953711bb68a542ac4afc0cde240775d5b2525d1bfce52424`
- frozen mobile lockfile install succeeds

## Remaining certification

This stays draft. Android device behavior, #7094, #7427, and buffered autocorrection now have current-head evidence. Affected-device Korean behavior under #11235 and the Android Fcitx private-preedit accessory-Enter limitation remain uncertified.


nwparker:
Closing my own draft — #7427 is fixed on main by #15007 (`66b5993`), and the iOS half of this is superseded.

Keeping the idea on file though: the Android native ViewManager here is still the only mechanism that could surface composition state on that platform. RN Android has no composing flag at all (`ReactTextChangedEvent` carries text + eventCount), and Android links the prebuilt `react-android` AAR, so a patch in `node_modules` is never compiled. #15007 falls back to holding the trailing non-ASCII run there, which covers Hangul, kana, kanji and Cyrillic but structurally cannot see an ASCII preedit like pinyin.

If Android pinyin turns out to matter, this approach is the answer — but as 28 files on a base 805 commits behind, it wants a fresh start rather than a rebase.

PR #12348, opened 2026-08-03
fix(terminal): stop keyboard effect re-registering on every pane render

### LINKED ISSUE #12269: [Bug]: terminal keyboard effect re-registers on every TerminalPane render (unmemoized createExpandCollapseActions)
### Summary

`TerminalPane` calls `createExpandCollapseActions()` in its render body with no memoization. The factory builds fresh closures on every call, and four of them are dependencies of the terminal keyboard effect, so **that effect tears down and re-registers on every single** `TerminalPane` **render**.

Two consequences:

1. Any state the effect owns is destroyed on every render. `modifiedEnterChordOwner` is created inside the effect body, so its "one claim until release" guarantee silently does not hold — it becomes "one claim per render".
2. Seven `window` listeners (`keydown` x2, `keyup` x2, `keypress`, `beforeinput`, `blur`) are removed and re-added on every render of every mounted terminal pane.

### Where

`src/renderer/src/components/terminal-pane/TerminalPane.tsx` (~line 1246):

```ts
const {
  setExpandedPane,
  restoreExpandedLayout,
  refreshPaneSizes,
  syncExpandedLayout,
  toggleExpandPane
} = createExpandCollapseActions({ ... })
```

`createExpandCollapseActions` (`expand-collapse.ts:115`) is a plain factory, not a hook. It defines five function expressions per call and returns a new object literal, with no caching. Four of its results — `setExpandedPane`, `restoreExpandedLayout`, `refreshPaneSizes`, `toggleExpandPane` — appear in the dependency array of `useTerminalKeyboardShortcuts` (`keyboard-handlers.ts`, deps at ~line 903, entries at 912-916).

`TerminalPane` is wrapped in `forwardRef` only, with no `React.memo`, so it re-renders whenever its parent does.

The other entries in that array are fine: the refs are `useRef` objects, and `persistLayoutSnapshot`, `handleSearchSelectedText`, `handleRequestClosePane`, `clearPaneScrollback`, `handleStartRename`, `handleClearPaneTitleShortcut` are all genuine `useCallback`s. The factory is the only churn source.

### Evidence

Calling the factory twice with an identical state object gives new identities for all five returned functions (`Object.is` false, 5/5).

Driving the real hook in the repo's own harness with a held modifier during composition — one keydown, 50 auto-repeat keydowns, one keyup — isolates the effect of the churn:

```text
no re-render                                    1 send
re-render, dependency identities held stable    1 send
re-render, dependencies rebuilt as TerminalPane rebuilds them   50 sends
```

That third row is what a real pane does today.

### Why it was noticed

While investigating stablyai/orca#11878, holding Shift mid-composition on Windows produced one spurious newline per render rather than one per press, so the injection rate followed the render rate — a few in an idle PowerShell pane, a continuous stream in a busy TUI pane. [#12265](<https://github.com/stablyai/orca/issues/12265>) fixed that symptom upstream by never classifying those events as Enter in the first place, which is the right fix and makes this latent again. But the underlying property is worth correcting on its own: a lock that does not survive a re-render will mislead the next person who reasons about `claim`/`release` as if it bounds anything.

### Suggested fix

Memoize the factory result at the call site, e.g. `useMemo(() => createExpandCollapseActions({ ... }), [...])` over the stable refs and `persistLayoutSnapshot`, or have the factory return stable callbacks. Either way the keyboard effect stops re-registering on unrelated renders.

A regression test in the spirit of the measurement above — assert a single send across N auto-repeats *with* re-renders — would pin it, and would also serve as a general guard that the effect is not re-registering.

### Environment

Found on `v1.4.163` and still present on `main` as of `173abf8b1`. Not platform-specific: the memoization gap is in shared renderer code, though the visible IME consequence was Windows-only.

brennanb2025:
## Summary

`TerminalPane` called `createExpandCollapseActions()` bare in its render body. The factory mints five fresh closures per call, and four of them (`setExpandedPane`, `restoreExpandedLayout`, `refreshPaneSizes`, `toggleExpandPane`) are dependencies of `useTerminalKeyboardShortcuts` — so the terminal keyboard effect tore down and re-registered its seven `window` listeners on **every** `TerminalPane` render, and every piece of effect-owned IME state (`modifiedEnterChordOwner`, `heldImeEnterModifiers`, `nativeOnlyShortcutTracker`, `deferredNewlineSender`, `observedEnterKeydownTimeStamps`, `optionKeyLocation`) was silently reset each time. Any render landing mid-chord or mid-composition wiped tracker state — the class of bug behind the STA-3128 symptom that #12265 patched downstream.

Fix: `useExpandCollapseActions`, a memoized wrapper in `expand-collapse.ts` keyed on field-level identities (all stable per mounted pane), so the actions — and therefore the keyboard effect registration — are stable across renders. The raw factory is unchanged for non-hook callers/tests.

Live measurement (dev app, CDP-instrumented `window` listener add/remove counts, single pane running 50 OSC title updates — title updates are what drives pane renders in a busy TUI; raw output and echo drive none):

| Scenario | Before | After |
|---|---|---|
| Idle 5s | 0 listener ops | 0 |
| Typing 31 chars | 0 | 0 |
| `seq 1 20000` output burst | 0 | 0 |
| 50 OSC title updates | **700 listener ops** (7 listeners × 50 re-registrations, 1:1 with renders) | **0** (all 50 title updates confirmed processed) |

Fixes #12269 (STA-3291)

## Screenshots

No visual change.

## Testing

- [x] `pnpm typecheck` (all three projects, clean `tsbuildinfo`)
- [x] `pnpm test` — targeted: `expand-collapse-render-stability.test.ts` (new), `expand-collapse.test.ts`, `keyboard-handlers.test.ts`, `keyboard-handlers-ime.test.tsx`, `keyboard-handlers-ime-enter-keyup.test.tsx` — 42/42 pass
- [x] `pnpm lint` — oxlint + react-doctor + oxfmt on changed files via pre-commit; `lint-react-doctor-changed` gate exits 0
- [ ] `pnpm build` — not run (renderer-only memoization; no build-surface change)

New regression test pins registration stability with TerminalPane-shaped wiring (dep objects rebuilt every render): action identities stable 5/5 across rerenders, remint on `tabId` change, and exactly one `window` listener registration across 10 rerenders. Verified non-vacuous: wiring it through the raw factory fails with `expected 11 to be 1`.

## AI Review Report

Reviewed the diff for: dependency-array correctness (hook deps enumerate every field the factory consumes; refs are stable, `setExpandedPaneId` is a `useState` setter, `setTabPaneExpanded` a store action, `persistLayoutSnapshot` a `useCallback` on `[tabId, setTabLayout, worktreeId]` — so the memo remints exactly when it should), rules-of-hooks (unconditional call site), and residual churn sources in the keyboard effect's dep array (`keybindings` is a direct store selector; `terminalShortcutPolicy` is a string; remaining entries are refs/`useCallback`s — the factory was the only churn source, confirmed by the zero post-fix measurement). Cross-platform: no shortcuts, labels, paths, shell behavior, or Electron platform branches touched; the memoization gap was shared renderer code affecting macOS/Linux/Windows identically (visible symptom historically Windows IME). SSH/remote and folder-workspace paths unaffected — pane wiring is common to all transports.

## Security Audit

No input handling, command execution, path handling, auth, secrets, dependency, or IPC changes. The change memoizes existing renderer-local closures; listener add/remove surface shrinks. No new attack surface; no follow-up needed.

## Notes

- Follow-up hardening in #12357 (this PR merged while the canary commit was in flight, so it landed separately): a dev-only churn canary (`keyboard-effect-churn-canary.ts`) warns when one mounted keyboard hook re-registers more than 4 times in 2s — catching any future identity-unstable dependency loudly in dev instead of silently re-churning. Keyed per mounted hook instance so tab remounts never false-positive; rate-limited; bounded state; compiled out of production builds (`import.meta.env.DEV`). A wiring-level test asserts the warn actually fires when a dep churns.
- Latent since #12265 removed the visible IME symptom; this corrects the underlying invariant ("one claim until release" now actually bounds a chord's lifetime) and stops 14 listener ops per render per mounted pane.
- A pre-existing react-doctor warning at `TerminalPane.tsx:1642` (`no-adjust-state-on-prop-change`) surfaces when this file is scanned; it is outside this change and left untouched.

### DIFF
--- a/src/renderer/src/components/terminal-pane/TerminalPane.tsx
+++ b/src/renderer/src/components/terminal-pane/TerminalPane.tsx
@@ -56,8 +56,8 @@ import type { TerminalKittyKeyboardModeTracker } from '../../../../shared/termin
 import {
   applyExpandedLayoutTo,
   cancelPendingPaneSizeRefreshFrames,
-  createExpandCollapseActions,
-  restoreExpandedLayoutFrom
+  restoreExpandedLayoutFrom,
+  useExpandCollapseActions
 } from './expand-collapse'
 import { useTerminalKeyboardShortcuts, type SearchState } from './keyboard-handlers'
 import type { MacOptionAsAlt } from './terminal-shortcut-policy'
@@ -1250,7 +1250,7 @@ function TerminalPane(
     refreshPaneSizes,
     syncExpandedLayout,
     toggleExpandPane
-  } = createExpandCollapseActions({
+  } = useExpandCollapseActions({
     expandedPaneIdRef,
     expandedStyleSnapshotRef,
     containerRef,
--- a/src/renderer/src/components/terminal-pane/expand-collapse-render-stability.test.ts
+++ b/src/renderer/src/components/terminal-pane/expand-collapse-render-stability.test.ts
@@ -1,0 +1,126 @@
+// @vitest-environment happy-dom
+import { cleanup, renderHook } from '@testing-library/react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import type { PaneManager } from '@/lib/pane-manager/pane-manager'
+import type { PtyTransport } from './pty-transport'
+import { useExpandCollapseActions } from './expand-collapse'
+import { useTerminalKeyboardShortcuts } from './keyboard-handlers'
+
+type ExpandCollapseHookState = Parameters<typeof useExpandCollapseActions>[0]
+type KeyboardHandlersDeps = Parameters<typeof useTerminalKeyboardShortcuts>[0]
+
+// Field identities are stable across renders (refs, setState, store actions,
+// useCallback); TerminalPane rebuilds only the wrapping object literal.
+function createStableFields(): Omit<ExpandCollapseHookState, 'tabId'> {
+  return {
+    expandedPaneIdRef: { current: null },
+    expandedStyleSnapshotRef: { current: new Map() },
+    containerRef: { current: null },
+    managerRef: { current: null },
+    setExpandedPaneId: vi.fn(),
+    setTabPaneExpanded: vi.fn(),
+    pendingPaneSizeRefreshFrameIdsRef: { current: [] },
+    persistLayoutSnapshot: vi.fn()
+  }
+}
+
+afterEach(() => {
+  cleanup()
+  vi.restoreAllMocks()
+})
+
+describe('useExpandCollapseActions render stability', () => {
+  it('keeps all five action identities across rerenders that rebuild the state object', () => {
+    const fields = createStableFields()
+    const hook = renderHook(() => useExpandCollapseActions({ ...fields, tabId: 'tab-1' }))
+    const first = hook.result.current
+
+    hook.rerender()
+    hook.rerender()
+
+    const actionNames = Object.keys(first) as (keyof typeof first)[]
+    expect(actionNames).toHaveLength(5)
+    for (const name of actionNames) {
+      expect(Object.is(hook.result.current[name], first[name])).toBe(true)
+    }
+  })
+
+  it('remints actions when tabId changes', () => {
+    const fields = createStableFields()
+    const hook = renderHook(({ tabId }) => useExpandCollapseActions({ ...fields, tabId }), {
+      initialProps: { tabId: 'tab-1' }
+    })
+    const first = hook.result.current
+
+    hook.rerender({ tabId: 'tab-2' })
+
+    expect(Object.is(hook.result.current.setExpandedPane, first.setExpandedPane)).toBe(false)
+  })
+})
+
+describe('terminal keyboard effect registration stability', () => {
+  it('registers window listeners once across rerenders with TerminalPane-shaped wiring', () => {
+    const scope = document.createElement('div')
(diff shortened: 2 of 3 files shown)


brennanb2025:
## Electron QA evidence

Validated pushed HEAD `75e5648053deae829909e4329f9bb7cf918cd1a7` in the intended dev instance (`tubesnout @ brennanb2025/issue-3291-user-report`).

- Loaded the real local terminal for this worktree through Electron/CDP.
- Typed 50 distinct OSC title updates through the terminal input surface; the visible tab reached `listener-stability-50`.
- Instrumented the keyboard effect's listener types during that render storm: **0 adds, 0 removes**.
- Typed again after the storm and observed `KEYBOARD_INPUT_OK` in the terminal.
- Renderer console: 0 errors.

Initial loaded state:



After 50 title-driven renders and a post-storm keyboard round trip:



Remaining platform gap: this macOS Electron run does not automate a real Windows OS IME/ConPTY session; simulated Windows keyboard/IME behavior is covered by the focused renderer tests and CI, while the Electron proof here covers the listener-stability invariant and visible terminal input path.

PR #6844, opened 2026-06-30
feat(server): headless, Electron-free @stablyai/orca-server

Jinwoo-H:
## Summary

Adds `@stablyai/orca-server`: an npm-installable, headless Orca runtime server that runs under plain Node without Electron or X11. After install, the supported cross-platform command is:

```sh
npm install -g @stablyai/orca-server
orca-ide serve
```

The npm package intentionally exposes only the `orca-ide` bin. It does not publish `orca`, to avoid shadowing GNOME Orca on Linux or colliding with desktop-app CLI shims on macOS/Windows. `orca-server.js` remains the package `main`/direct bundle entry, not a public global command.

## Screenshots

No visual change.

## Testing

- [x] `pnpm exec oxlint --format github` (0 errors; existing unrelated warnings remain)
- [x] `pnpm typecheck`
- [x] `pnpm vitest run config/scripts/package-electron-runtime-contract.test.mjs src/main/server/node-server-args.test.ts src/main/server/node-app-environment.test.ts src/main/speech/openai-api-key-store.test.ts src/main/providers/local-pty-shell-ready.test.ts src/cli/index.test.ts src/cli/runtime/launch.test.ts src/main/server/node-server-main.test.ts src/main/server/shared-user-data-guard.test.ts src/shared/secure-file.test.ts`
- [x] `pnpm vitest run src/main/persistence.test.ts -t "Kagi session"`
- [x] `pnpm run build:server`
- [x] `node config/scripts/headless-server-smoke-client.mjs "$PWD/out/server/orca-ide.js"`
- [x] Added/updated regression coverage for npm package invariants, serve arg parsing, node app paths, secret persistence, secure-file exclusive writes, and headless smoke behavior

## AI Review Report

Reviewed the PR with CodeRabbit plus an additional AI pass focused on the npm CLI surface and release flow. Main findings addressed:

- Keep the public npm bin as `orca-ide` only; do not publish `orca` or `orca-server` as global bins.
- Sync `@stablyai/orca-server` releases with desktop stable and RC releases: stable publishes to npm `latest`, RC publishes to npm `alpha`.
- Pin Node 24 for host prebuilds and verify node-pty ABI compatibility before copying prebuilts.
- Run Docker images as the non-root `node` user and keep healthchecks aligned with the published bin.
- Harden dynamic imports, serve arg parsing, app path resolution, and release workflow secret scoping.
- Preserve existing node secret-store keys on corrupt/read/race failures instead of regenerating and orphaning ciphertext.

[...]

PR #12352, opened 2026-08-03
fix(remote): distinguish hosts that need a workspace window

### LINKED ISSUE #12350: [Bug]: Reachable remote server is shown as Ready without a workspace window
### Operating system

macOS

### Orca version

Observed with a 1.4.167 client connected to a 1.4.150 host. The runtime protocol versions were compatible.

### Details

**Short summary**

A paired Remote Orca Server can remain reachable after its workspace window/renderer graph becomes unavailable. In that state, project Settings still labels the host setup **Ready**, and the status bar still presents the server as ordinarily connected, even though interactive workspace actions cannot use the missing renderer graph.

**What happened?**

The host runtime remained reachable and returned status successfully, but reported:

- `graphStatus: unavailable`
- `desktopWindowStatus: openable`

Settings > Project > Available Hosts nevertheless showed the remote setup as **Ready**. Attempting graph-dependent terminal actions then failed instead of explaining that Orca needed to be opened on the host. Opening Orca on the host restored the graph immediately.

**How can we reproduce it?**

1. Pair a desktop Orca client with a headed Remote Orca Server.
2. Close the server's workspace window while leaving its main process/runtime reachable.
3. Confirm the server reports an unavailable graph and an openable desktop window.
4. Open the project's **Available Hosts** settings or the status-bar remote-host menu.
5. Observe that the setup is still shown as **Ready/Connected** rather than identifying the unavailable workspace window.

**Expected behavior**

A reachable server without a usable workspace renderer should have a distinct, actionable state—for example, **Open Orca** / **Workspace window closed**—without being mislabeled as network-disconnected.

**Roadmap context**

This report is intentionally narrower than the ongoing work to remove the renderer dependency:

- #8300 keeps serving remote clients by optionally hiding rather than destroying the headed renderer.
- #6844 introduces an Electron-free headless server.

Those changes can prevent this state in more configurations. This issue only tracks truthful status for current, legacy, and degraded headed servers while the runtime explicitly reports that its desktop window is openable but its graph is unavailable.

The separate terminal-surface race present on the older host build was resolved by updating the host and is not the scope of this report.

gatsby74:
## Summary

Reachable Remote Orca Servers no longer look fully ready when their workspace renderer is unavailable. Project host settings now show **Open Orca** with an explanation, while the status-bar host row shows **Workspace window closed** and remains correctly counted as connected with a **Disconnect** action.

This is a status-clarity fix for current, legacy, and degraded headed servers. Graph-ready headless servers are unaffected, so it does not preserve or expand the renderer dependency being removed by #6844.

Fixes #12350.

## Screenshots

Before, the remote setup was presented as **Ready** even though its renderer graph was unavailable (sanitized):

Sanitized screenshot showing a remote host incorrectly labeled Ready

After this change, that badge reads **Open Orca** with explanatory copy, and the remote-host menu reads **Workspace window closed**.

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test` — 44,356 tests passed and 9 unrelated tests failed in updater, relay, native-watcher, and root-directory fixture suites. Under the required Node 24 runtime, five of those failures passed on rerun; the remaining four reproduced unchanged on clean `upstream/main` (`src/main/updater.test.ts` and `config/scripts/check-root-directory-entries.test.mjs`).
- [x] `pnpm build`

Focused verification: 24/24 tests passed across project-host settings, runtime-host classification, status-bar presentation, connected-count semantics, and closed-connection precedence, including disconnected and graph-ready boundary states added from review feedback.

## AI Review Report

An independent final review returned clean after checking the shared state classifier, both UI surfaces, and regression coverage. An earlier simplification pass identified duplicate reachability logic and an inverted UI-to-domain type dependency; the final implementation centralizes classification in a neutral renderer runtime module and reuses it from Settings and the status bar. CodeRabbit requested explicit disconnected and graph-ready Settings boundaries; both were added and revalidated.

Compatibility review explicitly covered macOS, Linux, and Windows. The change adds no shortcuts, platform labels, paths, shell behavior, Git commands, or Electron platform branches. Local hosts remain unchanged; nested SSH setups inherit their Remote Orca Server owner's state; graph-ready remote/headless hosts remain connected; and no agent, integration, or Git-provider-specific behavior changes. The classification is O(1) over already-fetched status and adds no polling, subscriptions, IPC, or hot-path work. UI review confirmed existing badge, color, spacing, and typography tokens are reused.

## Security Audit

The change reads existing validated runtime status fields only. It adds no input handling, command execution, filesystem access, path construction, authentication changes, secrets, dependencies, IPC methods, or network requests. A stale `closed` shared-control connection with an error continues to take precedence as disconnected, so stale window diagnostics cannot make a failed connection appear reachable.

## Notes

- Related: #8300, which can prevent this state by keeping a headed renderer alive when its window closes.
- Related: #6844, the Electron-free headless-server work.
- The separate terminal-surface timeout seen on the older host build was resolved by updating that host and is intentionally outside this PR.

### DIFF
--- a/src/renderer/src/components/settings/RepositoryHostSetupsSection.runtime-window.test.tsx
+++ b/src/renderer/src/components/settings/RepositoryHostSetupsSection.runtime-window.test.tsx
@@ -1,0 +1,99 @@
+// @vitest-environment happy-dom
+
+import { act } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { RUNTIME_PROTOCOL_VERSION } from '../../../../shared/protocol-version'
+import type { Project, ProjectHostSetup, Repo } from '../../../../shared/types'
+import { useAppStore } from '../../store'
+import { RepositoryHostSetupsSection } from './RepositoryHostSetupsSection'
+
+let container: HTMLDivElement
+let root: Root
+
+beforeEach(() => {
+  useAppStore.setState(useAppStore.getInitialState(), true)
+  container = document.createElement('div')
+  document.body.appendChild(container)
+  root = createRoot(container)
+})
+
+afterEach(() => {
+  act(() => root.unmount())
+  container.remove()
+  useAppStore.setState(useAppStore.getInitialState(), true)
+})
+
+describe('RepositoryHostSetupsSection runtime window availability', () => {
+  it('asks users to open Orca when a reachable host has no workspace window', () => {
+    const repo: Repo = {
+      id: 'remote-repo',
+      displayName: 'Orca',
+      path: '/srv/orca',
+      badgeColor: '#737373',
+      addedAt: 100,
+      kind: 'git',
+      executionHostId: 'runtime:hub'
+    }
+    const project: Project = {
+      id: 'github:stablyai/orca',
+      displayName: 'Orca',
+      badgeColor: '#737373',
+      sourceRepoIds: [repo.id],
+      createdAt: 100,
+      updatedAt: 100
+    }
+    const setup: ProjectHostSetup = {
+      id: 'hub-local-setup',
+      projectId: project.id,
+      repoId: repo.id,
+      hostId: 'runtime:hub',
+      executionHostId: 'local',
+      runtimeOwnerEnvironmentId: 'hub',
+      path: repo.path,
+      displayName: repo.displayName,
+      kind: 'git',
+      setupState: 'ready',
+      setupMethod: 'legacy-repo',
+      createdAt: 100,
+      updatedAt: 100
+    }
+    useAppStore.setState({
+      repos: [repo],
+      projects: [project],
+      projectHostSetups: [setup],
+      runtimeStatusByEnvironmentId: new Map([
+        [
+          'hub',
+          {
+            checkedAt: 1,
+            status: {
+              runtimeId: 'runtime-hub',
+              rendererGraphEpoch: 1,
+              graphStatus: 'unavailable',
+              authoritativeWindowId: null,
+              desktopWindowStatus: 'openable',
+              liveTabCount: 0,
+              liveLeafCount: 0,
+              runtimeProtocolVersion: RUNTIME_PROTOCOL_VERSION,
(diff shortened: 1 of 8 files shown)


mmarabel:
> Add boundary-case coverage for the distinct connection states.

Addressed in 2efa78f7ea: the project-host Settings test now also locks the fully disconnected state to **Disconnected** and the graph-ready state to **Ready**, alongside the existing **Open Orca** case. The focused suite now passes 24/24 tests.

nwparker:
Independently reproduced this and confirmed your root cause — thanks for the report and the fix, @gatsby74.

I opened #12477, which is a merge of your PR and an independent fix. Three things came straight from your diff: the broader `graphStatus !== 'ready'` predicate (a window torn down mid-reload is just as unusable, and `desktopWindowStatus === 'openable'` already implies no live renderer window), routing Settings > Available Hosts through the same `runtimeHostConnectionState` the status bar uses instead of only sharing the new predicate, and your regression vectors (unreachable owner stays Disconnected; a closed control channel wins over workspace-window diagnostics). Your single `isConnectedRuntimeHostState` in the status row replaced a near-duplicate helper I had.

Differences, with reasons: the predicate lives in `src/shared/` so main and the CLI can reuse it; the state is named `'workspace-window-closed'` to match the label; and the Settings badge reads "Workspace window closed" (a state, like "Ready"/"Disconnected") with the imperative moved into the explanation line, which names the host.

Full credit is in the PR body.

AmethystLiang:
Duplicate of #12477 (fixes #12350).

Same root cause and intended fix: a reachable remote Orca host with its workspace window closed must not present as fully Ready/Connected for work that needs the renderer. Please continue review/landing on #12477. If this branch has a meaningfully different implementation detail worth cherry-picking, call it out on #12477.

nwparker:
Landed as #12477 → `f968583e958` on `main`, closing #12350. Three things came straight from your diff: the broader `graphStatus !== 'ready'` predicate, routing both surfaces through a single `runtimeHostConnectionState` derivation (including your module placement), and reusing `isConnectedRuntimeHostState` instead of a near-duplicate predicate. Credited as co-author on the squash commit — thanks @gatsby74.

PR #12357, opened 2026-08-03
test(terminal): pin IME chord survival across mid-composition re-renders

brennanb2025:
## Summary

Adds an IME regression test pinning the STA-3291 failure mode behaviorally. The terminal keyboard effect owns the tab's IME chord state; before #12348 it tore down and re-registered on every `TerminalPane` render, so any render landing mid-composition silently reset that state. #12348 fixed the cause and pinned listener-registration stability; this test pins the *user-visible consequence*, so a future regression fails on behavior rather than on an internal counter.

The test wires the hook TerminalPane-shaped (dependency objects rebuilt every render) and interleaves re-renders through modifier hold, composition start, and five auto-repeat `Process`/229 keydowns. It asserts the bare Enter redispatch is still absorbed and that no newline is sent.

Verified non-vacuous: wired through the pre-fix (unmemoized) factory, the redispatch is not absorbed and the test fails.

Also verified on real Windows hardware with the genuine MS Korean 2-Set IME — see the comment below for the raw per-keystroke PTY capture. Five Shift-held double consonants (ㅃㅉㄸㄲㅆ) plus a composed syllable produced exactly one `0d`, from the deliberate Enter, with zero listener churn.

## Screenshots

No visual change.

## Testing

- [x] `pnpm typecheck` — web project, clean `tsbuildinfo`, exit 0
- [x] `pnpm test` — targeted: the IME suites, `keyboard-handlers`, and both expand-collapse suites — 43/43 pass on current `main`
- [x] `pnpm lint` — oxlint + react-doctor + oxfmt via pre-commit, clean
- [ ] `pnpm build` — not run (test-only change; no production code touched)

## AI Review Report

Reviewed for vacuity (confirmed the test fails against pre-fix wiring), for whether the assertion is behavioral rather than structural (it asserts absorbed redispatch and zero newline sends, not listener counts), and for platform assumptions. Cross-platform: the test forces the Windows IME code path via a `userAgent` stub, which is how the neighboring IME suites already work; no production code, shortcuts, labels, paths, or shell behavior are touched, so macOS/Linux/Windows behavior is unchanged.

## Security Audit

Test-only change. No input handling, command execution, path handling, auth, secrets, dependency, or IPC surface is touched. No new attack surface.

## Notes

- Test-only; production behavior is unchanged. The fix itself shipped in #12348.
- This PR previously also added a dev-only "churn canary" that warned when the effect re-registered rapidly. It has been dropped: its payout depended on a human noticing a console warning, which is not a realistic detection path. The CI regression test in #12348 plus this behavioral test cover the same ground without adding a runtime concept to the codebase.
- The branch name (`sta-3291-churn-canary`) is a leftover from that earlier scope.

### DIFF
--- a/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
+++ b/src/renderer/src/components/terminal-pane/keyboard-handlers-ime.test.tsx
@@ -3,6 +3,7 @@ import { cleanup, renderHook } from '@testing-library/react'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import type { PaneManager } from '@/lib/pane-manager/pane-manager'
 import type { PtyTransport } from './pty-transport'
+import { useExpandCollapseActions } from './expand-collapse'
 import { useTerminalKeyboardShortcuts } from './keyboard-handlers'
 import {
   installTerminalImeCompositionRoute,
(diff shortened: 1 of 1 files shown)


brennanb2025:
Windows real-IME verification — **PASS**.

Ran on a Windows host against a QA worktree off current \`main\` (fix present at \`637c7e94c9\`), driving the **real MS Korean 2-Set IME** through the OS input stack (not synthesized DOM events), with the pane going through a real PTY.

Renderer keylog confirms genuine IME events reached the handler: `HangulMode` keyup, then Shift held, then a `Process`/keyCode-229 keydown per consonant with `isComposing: true, shiftKey: true`.

Raw-mode PTY capture (per keystroke, hex + ms):

```
52748 e38583   ㅃ   } all five typed with Shift HELD
52961 e38589   ㅉ   } (the STA-3128 double-consonant shape)
53177 e384b8   ㄸ
53393 e384b2   ㄲ
53928 e38586   ㅆ
54759 eab080   가   composed syllable
54759 0d       CR   <- the single deliberate Enter
```

Zero spurious newlines across the held-modifier composition; exactly one `0d`, from the deliberate Enter. The decisive case is the commit itself: the IME fired `Process`(229) at t=615629 and Windows redispatched a bare `Enter`(13) at t=615631 — 2 ms apart, the classic double-Enter shape — and the PTY still received exactly one CR.

Keyboard-effect listener churn during the entire run: **0** (keyup/keypress/beforeinput/blur registrations all zero). The dev churn canary in this PR stayed silent, i.e. no false positives under real IME input.

Note for anyone reproducing: use `setRawMode(true)` for the stdin probe. A first attempt with a line-buffered probe showed an apparent extra newline after the first consonant, which turned out to be conhost line-editing at a `Terminate batch job (Y/N)?` prompt — not app behavior. Raw mode disambiguates.

QA worktree and all temp artifacts removed from the host.

brennanb2025:
Scope reduced: the dev-only churn canary has been dropped from this PR, leaving only the IME regression test.

Reasoning: the canary's detection path required a developer to notice a `console.warn`. In practice nobody has DevTools open for a warning they aren't looking for, so the insurance would rarely have paid out — while permanently adding a runtime concept that every future reader of the keyboard effect has to understand.

The coverage it was meant to provide is better served by the CI regression test already merged in #12348 (which blocks a merge rather than relying on someone watching), and, if we ever want the stronger guarantee, by a test that mounts the real `TerminalPane` and asserts listener registration count — that would catch newly-added unstable dependencies automatically. That is deliberately not in this PR; it needs the store, PTY transports, and xterm stood up, and should be weighed on its own merits.

The Windows real-IME verification in the comment above still stands: it exercised the shipped fix and this test's exact scenario.

PR #12359, opened 2026-08-03
fix(renderer): drive agent working spinner on the compositor (STA-3328)

brennanb2025:
## Summary

Typing lag on Mac (STA-3328) was root-caused live on a production instance to the shared agent-spinner JS clock: a 12Hz `setInterval` writing `el.style.transform` on every mounted working-spinner. With 41 spinners mounted that is ~490 main-thread style writes/s (97% of all DOM mutation traffic), keeping style perpetually dirty — single style-recalc passes of 96–120ms — and delaying keystroke dispatch.

This PR replaces the clock with a compositor-driven CSS animation (`steps(12, end)`, 30° per tick — the exact cadence the clock used). Each ring's CSS Animation is anchored once to the document timeline at mount (and when recreated after a reduced-motion change), keeping every ring in exact lockstep. `prefers-reduced-motion` still freezes the ring and the existing filled-top-border static marker is unchanged. `agent-spinner-clock.ts` is deleted; the `data-agent-spinner` hook attribute stays.

**Measured impact (live prod instance, user actually typing, A/B by stopping the clock's writes):**

| real-keystroke input delay | JS clock (before) | writes stopped (after-equivalent) |
|---|---|---|
| p50 | 7.3ms | 3.3ms |
| p90 | **363ms** | **19ms** |
| DOM attribute mutations | 453/s | 15/s (0 from spinners) |
| long-task total per 2min | 10.8s | 6.6s |

The residual long tasks are a separate mechanism (per-agent-status synchronous render bursts — follow-up ticket to come) and are unaffected by this PR.

**Verified in a dev build of this branch (CDP-driven):** 43 concurrent spinners → **0 attribute mutations and 0 long tasks over 8s**; computed-transform sampling shows exactly 12 distinct angles/s in 30° steps; all spinners report identical angles at the same instant (phase sync intact).

## Screenshots

Dev instance of this branch — worktree card, agent row, and tab indicator spinners render identically to prod (same ring geometry/colors; only the animation driver changed):

## Testing

- [x] `pnpm lint` — repo-wide oxlint clean, react-doctor changed-lines gate clean, max-lines ratchet OK (localization verifies not run; no string changes)
- [x] `pnpm typecheck` — all three configs pass
- [x] `pnpm test` — targeted: AgentWorkingSpinner, AgentStateDot, StatusIndicator, TerminalTabLeadingIcon (20 tests pass); full suite left to CI
- [ ] `pnpm build` — left to CI
- [x] Added `AgentWorkingSpinner.test.tsx`: pins the component→`main.css` wiring (class present in markup; `.agent-working-spinner` rule exists with `steps(12, end)` and a reduced-motion `animation: none` override) so neither side can be renamed or dropped alone. Updated AgentStateDot/StatusIndicator tests that pinned the old clock-driven contract.

## AI Review Report

Reviewed the change for: dangling imports of the deleted module (none — `registerStaleDocumentVisibilityRecovery` and `window-visibility-interval` have other consumers and are untouched); reduced-motion behavior (media query disables the animation; existing `motion-reduce:border-t-yellow-500` fills the ring, matching the old frozen-state rendering, #9515); phase-sync lifecycle safety (only the matching CSS animation is anchored with `startTime = 0` at mount/recreation, with no retained animation refs, timers, or cleanup); visual cadence parity (measured 30°/12Hz, identical to the clock). Cross-platform: the change is renderer-only CSS/Chromium behavior, identical on macOS, Linux, and Windows; no shortcuts, labels, paths, or shell behavior touched; SSH and folder-workspace surfaces render the same component. The old clock's occluded-window concern is moot: Chromium stops producing frames for occluded windows, so the CSS animation costs nothing when hidden, and unlike the JS clock it needs no stale-visibility recovery wiring to survive macOS occlusion-tracker glitches.

## Security Audit

No input handling, command execution, path handling, auth, secrets, or IPC changes. Net code deletion (a timer module and its DOM writes); one per-creation Web Animations `startTime` assignment; it does not mutate DOM attributes or install a timer. Test file reads `main.css` from the repo only.

## Notes

- Root-cause evidence and measurement methodology (CDP rig, DOM-mutation census, content traces) are recorded in the Linear ticket thread; the second lag mechanism (synchronous per-status render bursts: `applyAgentStatus` → `setAgentStatus` → `renderRootSync`, 200–488ms tasks under heavy orchestration) is intentionally out of scope here and needs its own ticket.
- The deleted clock's design goal (don't keep the frame pipeline awake per element) was measured against compositor CPU; this PR optimizes for input latency instead, which the compositor path cannot affect by construction.


brennanb2025:
## Post-review Electron validation

Validated pushed head `f6d5cabae5` in an isolated Orca dev instance attached to this exact worktree. I selected the workspace through the rendered UI, seeded 40 working child-agent rows through the renderer store to reproduce the production-scale workload, and exercised trusted keyboard input through the real terminal surface.

- 43 real `AgentWorkingSpinner` components: every animation had `startTime = 0`
- 11 distinct 30° steps sampled over 1.26s; 0/14 phase-mismatch samples
- 0 spinner attribute mutations, 0 long tasks, 0 animation-start callbacks after settled mount
- Terminal keydown delay: p50 0.4ms, p90 0.6ms, max 1.3ms (32 samples)
- Reduced motion: 0 active animations and the filled static ring; restoring motion recreated 43 animations exactly once, all re-anchored to `startTime = 0`, with 0 DOM mutations
- Focused component and formerly CI-failing suites: 35/35 pass; renderer/web TypeScript, focused oxlint, and diff check pass
- Final inherited-model subagent review: **CLEAN**

Final-head rendered workspace state (`getAnimations` available; all 3 visible surfaces anchored at `startTime = 0`):



43-spinner agent/subagent scale state:

PR #12362, opened 2026-08-03
fix(mobile): bound terminal viewport resubscribe loop with backoff (STA-3337)

brennanb2025:
## Summary

Fixes STA-3337: with the mobile terminal view active, a host that answers `terminal.subscribe` with an empty scrollback frame and no usable dims sent the client into an unbounded `unsubscribe` → `subscribe` loop (~25 cycles/s measured live) with no backoff and no attempt cap. Each pass re-initialized xterm mid-gesture (reproducibly breaking the tab title's long-press action sheet) and re-rendered the session route, draining battery and bandwidth for as long as the view stayed mounted.

Client (primary fix — must be robust regardless of host version):

- Absent host dims are no longer coerced to 80x24 for the fit comparison. A dimless frame can never converge with the phone viewport, so the stream is now held open instead of torn down and re-armed.
- Viewport-fit resubscribes are budgeted per handle (3 attempts) with escalating backoff (immediate → 750ms → 3s). The live stream stays up through the backoff so input keeps flowing; teardown happens only when the retry fires. The budget refills only when `terminal.list` drops the handle and later lists it again — the same absence-gated bound the chat-side stream uses in `use-mobile-native-chat-terminal-stream.ts`, so a still-listed non-converging handle cannot be re-funded on every list refresh.
- On exhaustion the view degrades visibly instead of hot-looping: the terminal keeps rendering at host dims and the existing toast shows "Couldn't fit the terminal to this screen" once per exhaustion.
- A fresh post-measure match with the host now counts as convergence (the pre-measure mismatch was a stale cached viewport) instead of forcing another teardown.
- `setTerminalModes` keeps the Map identity when the mode didn't change, so same-mode stream frames no longer re-render the whole session route.

Host:

- The `terminal.subscribe` scrollback emission now falls back to the subscriber's own viewport when both the serialized snapshot and the PTY size are unavailable (JSON and binary stream paths), so current hosts report converging dims instead of nothing (JSON path) or a hardcoded 80x24 (binary path).

The decision logic lives in a new pure module `mobile/src/session/mobile-terminal-viewport-resubscribe.ts` (decision function, post-measure convergence check, per-handle budget, and the orchestrated fit pass), which also brings the session route back under the max-lines gate.

Linear: [STA-3337](https://linear.app/stably/issue/STA-3337/mobile-terminal-view-unbounded-terminalsubscribe-re-arm-loop-no)

## Screenshots

No visual change in normal operation. The only new UI behavior is the existing toast component showing "Couldn't fit the terminal to this screen" once if fit attempts exhaust (degraded-host edge case; reproducing requires the STA-3337 mock-host rig, see the ticket).

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [ ] `pnpm build`

`pnpm lint` (full root suite) and `mobile/ pnpm lint` both pass. `pnpm typecheck` run at root and in `mobile/`. `pnpm test` = the full mobile suite from `mobile/` (391 files, 2915 passed) plus the root suites covering `terminal.subscribe` (`terminal-mobile-subscribe-tab-mount`, `terminal-subscribe-lease-only`, `terminal-subscribe-exit-waiter-leak`, `rpc/streaming`); the remaining root suite and `pnpm build` are left to CI. The new unit tests cover the resubscribe decision (absent dims, convergence, bound, backoff ordering), budget semantics (charge, converge-reset, no refill while still listed, absence-gated refill, single exhaustion announcement), and the orchestrated fit pass (stale-generation guards on deferred retries, backoff keeps the stream up until the retry fires, attempts charged only when the resubscribe actually armed).

## AI Review Report

Main risks checked: (1) stale-generation races — every async resume point, including the new backoff timer callback, re-checks `subscribeSeqRef` before tearing anything down, so a tab switch or reconnect during the backoff cannot resubscribe a stale generation; (2) budget re-funding — refills are absence-gated exactly like the chat-side rearm budget, preventing `terminal.list` refreshes from re-arming the loop; (3) interplay with the other resubscribe drivers (foreground recovery, `useTerminalViewportRefit`, chat cover/resume) — the budget scopes only to the scrollback fit pass, so user- and layout-driven refits remain unaffected; (4) input-lease continuity — backoff defers the teardown, not just the retry, so the subscription (and the input floor it carries) stays live between attempts; (5) old-host compatibility — pre-binary desktops that emit dimless JSON scrollback now hold instead of looping, with no host upgrade required.

Cross-platform compatibility was explicitly reviewed for macOS, Linux, and Windows hosts and iOS/Android clients: the change is pure TypeScript state logic plus the route's existing timer (`scheduleDelayedAction`) and toast utilities — no keyboard shortcuts, shortcut labels, file paths, shell behavior, or Electron platform-specific APIs are touched. SSH-hosted and folder (non-worktree) workspaces use the same stream path; the fix is host-agnostic.

## Security Audit

Host-provided cols/rows are accepted only when they are finite numbers greater than zero; anything else is treated as absent. No command execution, no path handling, no auth or secret handling, and no dependency changes. The host-side change only echoes back viewport values the client itself supplied in the schema-validated `terminal.subscribe` params, so nothing new crosses the RPC boundary. The behavioral direction is defensive: the change strictly reduces client-initiated RPC volume (previously unbounded, now capped at 3 fit attempts per handle per lifetime-with-refill).

## Notes

- Old desktops in the field still emit dimless scrollback on the JSON stream path; the client-side hold covers them without a host upgrade.
- A PTY that legitimately cannot converge to phone dims (e.g. pinned desktop display mode reporting desktop cols) now stops probing after 3 attempts and announces the degraded state once — the terminal remains usable at host dims, and the long-press sheet keeps working so the user can leave the view.
- Follow-up candidates (not blockers): the fit pass could prefer the in-place `terminal.updateViewport` RPC (as `useTerminalViewportRefit` already does) over teardown/resubscribe on hosts that support it.

### DIFF
--- a/mobile/app/h/[hostId]/session/[worktreeId].tsx
+++ b/mobile/app/h/[hostId]/session/[worktreeId].tsx
@@ -237,6 +237,7 @@ import { useMobileNativeChatTerminalStream } from '../../../../src/session/use-m
 import { subscribeMobileTerminalSafely } from '../../../../src/session/mobile-terminal-stream-subscribe'
 import {
   TerminalViewportResubscribeBudget,
+  readTerminalViewportDims,
   runTerminalViewportFitPass
 } from '../../../../src/session/mobile-terminal-viewport-resubscribe'
 import { activateMobileSessionTab } from '../../../../src/session/mobile-session-tab-activation'
@@ -1512,14 +1513,7 @@ export default function SessionScreen() {
               return
             }
             updateTerminalCwdFromStreamEvent(handle, data, terminalCwdRef.current)
-            const hostCols =
-              typeof data.cols === 'number' && Number.isFinite(data.cols) && data.cols > 0
-                ? data.cols
-                : null
-            const hostRows =
-              typeof data.rows === 'number' && Number.isFinite(data.rows) && data.rows > 0
-                ? data.rows
-                : null
+            const { hostCols, hostRows } = readTerminalViewportDims(data)
             // Why: absent host dims must not be coerced into a comparable size — 80x24
             // never equals a phone viewport and armed a zero-delay resubscribe loop (STA-3337).
             const cols = hostCols ?? viewportRef.current?.cols ?? 80
@@ -1596,8 +1590,12 @@ export default function SessionScreen() {
           } else if (data.type === 'resized') {
             updateTerminalCwdFromStreamEvent(handle, data, terminalCwdRef.current)
             // Server resize: reinit xterm on a full-buffer snapshot (width reflow rewraps scrollback), else just resize geometry.
-            const cols = (data.cols as number) || 80
-            const rows = (data.rows as number) || 24
+            const viewport = viewportMeasuredRef.current ? viewportRef.current : null
+            const [cols, rows] = viewportResubscribeBudgetRef.current.observeResize(
+              handle,
+              data,
+              viewport
+            )
             const serialized = typeof data.serialized === 'string' ? data.serialized : null
             diagnostics.streamResized(handle, seq, eventSeq, data, getTerminalRef(handle) != null)
             const oscLinks = isTerminalOscLinkRanges(data.oscLinks) ? data.oscLinks : undefined
--- a/mobile/app/h/[hostId]/session/[worktreeId].tsx
+++ b/mobile/app/h/[hostId]/session/[worktreeId].tsx
@@ -235,6 +235,10 @@ import {
 } from '../../../../src/session/mobile-terminal-prune-decision'
 import { useMobileNativeChatTerminalStream } from '../../../../src/session/use-mobile-native-chat-terminal-stream'
 import { subscribeMobileTerminalSafely } from '../../../../src/session/mobile-terminal-stream-subscribe'
+import {
+  TerminalViewportResubscribeBudget,
+  runTerminalViewportFitPass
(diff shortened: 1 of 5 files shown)


brennanb2025:
## QA evidence — post-review head `0e9558112a`

### Electron (macOS)
Attached to the intended `sta-3337-terminal-resub-loop` dev instance. The real terminal accepted input and visibly rendered its output.

Electron host terminal

### iOS Simulator — iPhone 17 Pro
Ran the development build against the STA-3337 `MOCK_NATIVE_CHAT=1` host, which returns `subscribed` plus empty 80×24 scrollback. After settling, mock totals remained flat at 4 subscribes / 3 unsubscribes across a five-second observation window, leaving one live stream instead of hot-looping. The tab long-press action sheet opened from terminal view and exposed “Switch to chat view,” proving the gesture remained usable after the faulty frame.

iOS terminal long-press after bounded resubscribe

Metro bundled successfully with no redbox. Deterministic review verification also passed 46 focused mobile tests and 114 host streaming/multiplex/mobile-integration tests (2 existing skips), plus mobile typecheck, targeted lint/React checks, formatting, and `git diff --check`. Live SSH/relay remains unexercised; the shared provider contract and unit coverage were reviewed.

PR #12363, opened 2026-08-03
Fix terminal corruption after restored snapshot replay

brennanb2025:
## Summary

- Fix three `@xterm/addon-serialize` fidelity defects that can shift, duplicate, or leave filler in restored terminal rows:
  - parenthesize the soft-wrap validity ternary;
  - suppress zero-distance cursor movement (`CSI 0 C/D` is interpreted as one cell);
  - materialize visually meaningful inverse null cells while preserving genuine wide-character wrap padding and effective WebGL styles.
- Bump the persistent daemon protocol to v32 so new PTYs use the corrected serializer.
- Preserve live v30/v31 PTY ownership and its one attach snapshot, but stop treating those lossy snapshots as authoritative for background thinning/parking. This retains existing sessions and shell history while keeping future bytes complete in the current process.
- Refresh snapshot capability once services are ready, including degraded-hydration recovery, so the pre-provider fallback cannot permanently disable v32 parking.
- Remove the broad known-bug allowances from both restore fuzz suites and add focused normal-buffer, scrollback, alternate-screen, Codex truecolor, inverse-null, wide-wrap, and decoration parity coverage.

## Root cause

`v1.4.168-rc.1` includes #12351, which correctly reattaches panes to persistent daemons after an update. That exposed pre-existing serializer defects in the first restored frame; it did not corrupt the live PTY byte stream. The effect is most visible on the first terminal after restart/worktree switching because fresh output or the existing post-reattach `SIGWINCH` repaint normally replaces the bad frame. Renderer throttling under system load can leave it visible much longer.

This was reproduced in the headless xterm restore-parity harness with minimal ANSI streams. The two soft-wrap repros failed before the patch and now run as ordinary regression tests.

## Reliability contract

- **Invariant:** v32 snapshots restore the exact visible terminal state; v30/v31 panes remain mounted and their snapshots are never treated as authoritative.
- **Oracle:** cell/text/style/wrap/cursor parity and fuzz coverage, plus a real hidden/reveal cycle followed by live keyboard input/output.
- **Gate:** `terminal-output.scrollback-restore` is the primary gate, supplemented by cold-restore, capability, parking, serializer, and protocol suites.
- **Provider coverage:** local daemon covered; WSL shares local routing; SSH behavior is unchanged; paired/mobile paths are unaffected.
- **Performance:** no polling or hot-path provider calls were added; capability refresh is one-shot after services become ready.
- **Diagnostics:** snapshot source/sequence and hidden-delivery counters distinguish provider restore from renderer replay.
- **Residual gaps:** live UI validation covered macOS with a local daemon; Linux, Windows, SSH, and WSL live UI were not exercised.

## Compatibility

- Native/WSL: surviving protocol v30/v31 sessions stay alive and mounted instead of relying on unsafe provider snapshots; new v32 sessions keep snapshot-backed parking.
- SSH: unchanged; remote runtime snapshot authority remains independently determined.
- Folder workspaces and git worktrees share the same version-gated local routing.
- No timeout, startup reset, forced PTY restart, or cached-state heuristic is introduced.

## Review fixes

- `1cc3fe9672` keeps legacy v30/v31 snapshot panes mounted.
- `e50f85e3ef` refreshes snapshot capability after startup.
- `f92b6aa01f` applies the same refresh after degraded hydration recovery.
- A fourth same-model review round returned clean.

## Validation

- Full affected reliability gate: 2,075 passed, 1 skipped.
- 340 additional focused tests passed across capability/parking and serializer/protocol suites.
- Serializer fidelity fuzz: 5,000 seeds passed; independent review run: 3,000 seeds passed.
- Hidden/reveal reconciliation fuzz: 2,000 scenarios passed.
- Targeted minimized seeds 45, 2318, and 2476 passed.
- Web and Node typechecks passed.
- Standard and type-aware lint, formatting, max-lines, offline frozen-install, and diff checks passed.

Performance measurements (200x50 terminal):

| Buffer | Snapshot median | Snapshot max | Serialized size |
| --- | ---: | ---: | ---: |
| Empty | 1.15 ms | 4.51 ms | 0 B |
| 1k rows | 13.65 ms | 20.80 ms | 95,497 B |
| 5k-row daemon cap | 73.33 ms | 82.64 ms | 482,674 B |
| 50k renderer scale | 531.22 ms | 691.21 ms | 4,868,799 B |

The existing incremental checkpoint path remains 0.06 ms in the simulated daemon loop versus 63.56 ms for a full snapshot (2.68 ms versus 66.23 ms worst forwarding gap). This patch does not add work to the live renderer/output hot path.

## Visual evidence

Electron QA used the intended `brennanb2025/fix-terminal-snapshot-wrap-restore` worktree and a real tab switch. With 1,241 hidden characters withheld from the renderer, the authoritative headless snapshot restored the exact 135x59 alternate-screen frame on the same PTY, and live keyboard input/output continued afterward.

Hidden terminal snapshot restored

Live input and output after restore


brennanb2025:
Electron QA completed against the intended `brennanb2025/fix-terminal-snapshot-wrap-restore` worktree.

- Real hidden/reveal flow withheld 1,241 characters from renderer delivery.
- An authoritative headless snapshot restored the same live PTY at 135x59, including soft-wrap, inverse-null, wide-margin padding, and style fidelity.
- `RESTORE_OK_PR12363` was visible after reveal; focused terminal input then produced `LIVE INPUT / OUTPUT AFTER RESTORE: OK` on the same PTY.

Hidden terminal snapshot restored

Live input and output after restore

brennanb2025:
Final same-model review and Electron QA completed on head `e4bbb184e6`.

Review fix:
- Fixed a startup race where snapshot-capability refresh could resolve against the fallback local provider before daemon-provider authority settled.
- Local capability lookup now awaits the existing bounded provider-startup promise; SSH/remote-only batches bypass it.
- The follow-up review of the new head returned `CLEAN`.

Validation:
- Snapshot capability/provider gate: 590/590 passed.
- Scrollback restore gate: 540/540 passed.
- Electron attached to `amphipod @ brennanb2025/fix-terminal-snapshot-wrap-restore`.
- The real daemon PTY reported authoritative snapshot support.
- A tab switch withheld 2,633 renderer characters across 3 chunks.
- Main returned a 179x56 alternate-screen snapshot with `source: headless` and the delayed restore marker.
- Reveal visibly restored soft-wrap, wide-glyph, inverse-empty-cell, truecolor, and row fidelity.
- Live terminal input/output continued on the same PTY after restore.

Restored authoritative terminal snapshot

Live input and output after restore

PR #12366, opened 2026-08-03
feat(mobile): native-chat model/session-option picker + shared slash catalog (STA-3332)

brennanb2025:
## Summary

Mobile native chat reaches desktop parity for model/session-option control (STA-3332):

**Slash-command catalog (shared source of truth).** The mobile composer's autocomplete now serves `getVerifiedNativeChatCommands` from `src/shared/native-chat-slash-commands.ts` — agent-aware (Claude vs Codex command sets) with the catalog's description rows — replacing a hardcoded provider-agnostic list that advertised `/model`, `/cost`, `/diff` (not in Claude's catalog) and omitted the Codex set entirely.

**Send classification (fixes the stuck "Queued" bubble).** `classifyNativeChatSend` moved into the shared module (the renderer re-exports it, so desktop import paths are untouched), and mobile's send seam now gates its optimistic echo on it, exactly like desktop's composer. Slash sends dispatch into the agent's TUI and never echo as user turns, so previously every command send left a "Queued" bubble that no transcript match could retire. Commands also skip the ack-lost "delivery unconfirmed" hold, which could never observe a landing.

**Model/session-option pickers.** The composer grows a pill row (current model + option summary) that opens an inline choice card following the proven mobile Ask-card pattern. It reads the shared `agent-session-option-catalog` and serves whatever it provides:

- **Claude**: model select (applies via the catalog's `modelApply` semantics — `/model <value>` through the existing send path), per-model effort select (`/effort <value>`), and fast-mode flip with the desktop's flip-only truth rules (toggle action when the baseline is unknown, absolute On/Off once tracked).
- **Codex**: model/effort are agent-picker entries per the catalog — picking one dispatches `/model` and flips the tab to the terminal view so the agent's own TUI selector is visible (desktop's `onAgentPicker` behavior).
- Applied values render as dispatched ("Sent to the agent — not confirmed") until a hook report confirms; the current model seeds from the live agent status's provider model when derivable and degrades to a neutral "Model" pill when not.
- Typed commands (`/model sonnet`, `/fast`, bare `/model`) update tracked state via the desktop `recordOutgoingCommand` rules, including clearing tracked truth and surfacing the terminal for bare picker commands.

Per-tab option records live in a bounded in-memory scope cache (parity with desktop's composer scope cache), in new well-named modules: `mobile-native-chat-session-option-{state,commands,labels}.ts`, `use-mobile-native-chat-session-options.ts`, `MobileNativeChatSessionOptionPickers.tsx`, `MobileNativeChatComposerSuggestions.tsx`.

Deliberately out of scope: `agent-session-option-catalog-claude-codex.ts` is untouched — STA-3330 owns catalog contents; this picker renders whatever the shared catalog provides.

## Screenshots

No screenshots — rendering requires booting the iOS simulator rig, which is not cheap in this environment. Manual verification path: open a Claude tab's native chat on mobile → the pill row shows "Model" (or the reported model) above the composer; tapping lists the catalog models with descriptions; picking one writes `/model <id>` to the PTY (visible in the terminal view) and the pill shows the picked model as dispatched. On a Codex tab, the model pill offers "Choose in agent picker…", which sends `/model` and flips to the terminal where Codex's TUI selector appears. Typing `/` in the composer lists the agent's real command set with descriptions; sending `/clear` no longer leaves a permanent "Queued" bubble. All of these behaviors are additionally locked by unit tests (see below).

## Testing

- [x] `pnpm lint` — mobile `oxlint` + `oxfmt --check` clean; root `oxlint`/react-doctor/`oxfmt` ran via the pre-commit hook on all changed files
- [x] `pnpm typecheck` — mobile `tsc --noEmit` clean; root `typecheck:web` clean
- [x] `pnpm test` — mobile suite: 395 files / 2938 tests pass; desktop `src/renderer/src/components/native-chat` (62 files / 616 tests) and the shared slash-command/classification suites pass. Pre-existing local failures in `src/shared/remote-runtime-client.test.ts` and similar socket-timing suites are environment reds unrelated to this change (no import path touches the changed modules); CI is authoritative there.
- [ ] `pnpm build` — not run locally; no build-config changes
- [x] Added/updated tests: slash catalog agent-awareness + description rendering (composer test), classification gating incl. the Queued regression and ack-lost hold (`use-mobile-native-chat-message-send.test.ts`), send classification per agent, autocomplete ranking, session-option snapshot/reporting/model matching, command build/parse/recording, hook apply-dispatch + agent-picker + reported seeding, and picker rendering/apply dispatch (component test)

## AI Review Report

Reviewed the change for: (1) optimistic-echo accounting — command sends must not create pending bubbles, and image-echo ordinal reconciliation must be unaffected (gating happens before `acceptSend`, and image-only sends still classify as chat); (2) state-truth honesty — dispatched values never claim confirmation, flip-only toggles never invent a baseline, model switches invalidate destination-model tracked values, and reports supersede local guesses (ported from the desktop apply/recording rules with their tests); (3) send interleaving — picker dispatches go through the existing single send seam (heal → clear → send) and are serialized in the options hook, so applies cannot interleave on one input line; (4) memory growth — per-tab records are capped at 32 scopes. Cross-platform: this is mobile (iOS/Android) + shared-module code; no keyboard shortcuts, file paths, or shell behavior are touched, the shared module is pure data/string logic used identically by the Electron renderer on macOS/Linux/Windows, and the desktop change is a re-export with no behavior delta. SSH/folder-workspace behavior is unaffected: sends ride the existing `terminal.send` RPC, which is host-agnostic.

## Security Audit

No new input surfaces, exec paths, or IPC channels. Picker values come exclusively from the curated shared catalog (never free text), and commands are built by catalog `build()` functions before entering the existing PTY send path with its existing budgets/acks; typed-command parsing (`parseBuiltSessionOptionCommand`) does template prefix/suffix matching only and never executes or re-sends content. No secrets, auth, or dependency changes; the mobile lockfile is untouched.

## Notes

- Mobile cannot read the agent's terminal screen, so two desktop affordances degrade gracefully: the current model/effort are only known when the live agent status reports a provider model (otherwise the pill shows "Model" until the user picks), and Claude's model-switch confirmation prompt is not detected — a dispatched value stays "Sent to the agent — not confirmed" rather than ever claiming applied.
- Selections are not persisted into `nativeChatSessionOptions` defaults (desktop writes app settings; mobile has no equivalent settings write today). Tracked state is per-tab and in-memory.
- Records are scoped per tab, not per agent-session: an agent restart in the same tab keeps prior dispatched state until a hook report corrects it.
- STA-3330 owns catalog contents (e.g. new model seeds); this PR renders whatever the shared catalog provides and does not edit it.

Linear: STA-3332


brennanb2025:
QA evidence for final head c208bc883c

Electron (Playwright/CDP, exact worktree): opened a real Claude native-chat session; the model picker showed Fable 5, Opus 4.8, Sonnet 5, and Haiku. Renderer console: 0 errors.



iOS 26.5 — QA-Lane-B-iPhone (paired mock host):

- Claude showed the same model catalog; selecting Sonnet 5 sent /model sonnet and revealed its Effort option.
- Agent-aware slash autocomplete showed command descriptions.
- /clear was accepted once with no Queued bubble.
- Codex offered Choose in agent picker… and returned to the TUI once.









Final review: CLEAN. Full CI is green.

brennanb2025:
Final-head QA at 3eafcb4500 — CLEAN.

- Electron via CDP: launched this exact worktree with a disposable profile, opened a real Claude native-chat tab, and verified the model picker shows Fable 5, Opus 4.8, Sonnet 5, and Haiku. Console errors: 0.
Electron model picker

- iOS via Orca emulator: paired the final-head development build to the native-chat mock, opened the model picker, selected Sonnet 5, verified its checked state and composer label, and confirmed the server received the /model sonnet command. Metro bundled successfully with no redbox.
Mobile model picker with Sonnet 5 selected

Focused verification: mobile 81/81 tests, desktop native-chat 616/616 tests, web typecheck, max-lines ratchet, and diff whitespace all passed. All current PR checks are green.

brennanb2025:
## Review gate: Electron desktop validation

Ran the dev build from this worktree (isolated `ORCA_DEV_USER_DATA_PATH` profile, own CDP + renderer ports) and confirmed app identity before probing — `devRepoRoot` = `sta-3332-mobile-model-picker`, so these are this branch's renderer, not a neighbouring worktree's.

The desktop-facing half of this PR is the extraction of the session-option logic into `src/shared/native-chat-session-option-{commands,snapshot,state}.ts` with re-export shims in the renderer. The claim is "no behavior delta", so the check is whether the desktop native-chat model/option pickers — the real consumers of that logic — still work end to end. They do:

**1. Native chat composer renders with the Model control**

Desktop native chat

**2. Model menu serves the shared catalog** (Fable 5 / Opus 4.8 / Sonnet 5 / Haiku) via the extracted `buildNativeChatSessionOptionSnapshot`

Desktop model menu

**3. Picking Sonnet 5 dispatches through the catalog's `modelApply`** — the transcript shows `Ran /model sonnet` (dispatched into the TUI, *not* echoed as a user chat turn, which is exactly the `classifyNativeChatSend` behavior this PR now shares with mobile), the pill updates to `Sonnet 5`, and the per-model `Effort` descriptor appears now that a model is tracked

Desktop model applied

**4. Per-model Effort options resolve for the tracked model** (Low / Medium / High / Extra high / Max)

Desktop effort menu

Screenshots are full app windows, uncropped. They live on `evidence/pr-12366-screenshots` (SHA-pinned above) rather than in this PR's diff: `files` in `config/electron-builder.config.cjs` is all-negation, so a `.pr-screenshots/` directory with no `!` entry would ship inside `app.asar`.

`Opus 4.8` rather than Opus 5 in the catalog is the known STA-3330 gap (#12369) — this PR renders whatever the shared catalog provides and deliberately does not edit it.

### Review fixes pushed to this branch

- **`getScopedRecord` evicted by insertion order, not recency** (`use-mobile-native-chat-session-options.ts`). It returned an existing record without re-inserting it, so a long-lived active tab is the oldest key — crossing the 32-scope cap silently dropped that tab's tracked model and reset the pill to "Model". Desktop's `setBoundedScopeCacheEntry` does delete-then-set for exactly this reason, and the PR body claimed parity with it. Now LRU, pinned by a test that fails on the old code and passes on the new.
- **Shared-logic tests only ran in mobile's vitest project.** `mobile-native-chat-session-option-{commands,state}.test.ts` imported nothing from `mobile/` — they exercised `src/shared` logic the Electron renderer consumes — but sat under `mobile/src/session/`, which the root suite's `src/**/*.test.ts` glob never sees, so a desktop-side edit to those shared modules would have shipped green. Moved to `src/shared/native-chat-session-option-{commands,state,snapshot}.test.ts` (20 tests now under the root gate) and named after modules that actually exist.
- Restored two load-bearing "why" comments dropped during the extraction (the flip-only-action truth rule; the model-command-resets-model-scoped-state rule).

Verified after the fixes: root `tsc --noEmit` clean (tsbuildinfo cleared first), mobile `tsc --noEmit` clean, root + mobile `oxlint`/`oxfmt --check` clean, `check:react-doctor:changed` exit 0, desktop + shared native-chat suites 75 files / 692 tests pass, mobile `src/session` 116 files / 916 tests pass.

brennanb2025:
## Review gate: iOS Simulator QA (Orca Mobile)

Driven on a booted `iPhone 17 Pro` against a Metro started from **this** worktree on its own port (8092 — deliberately not the 8081 Metro another worktree owns, which would have bundled the wrong branch), using the mock-server native-chat rig so no desktop pairing runtime was needed.

**1. The pill row exists.** Mobile previously had no model/session-option control at all; `/model` was raw PTY text. The row now sits above the composer in a Claude tab's native chat, showing the neutral `Model` pill because this rig reports no `agentStatus.model` — the documented degradation, not a bug.

Pill row

**2. The choice card serves the same shared catalog as desktop** — Fable 5 / Opus 4.8 / Sonnet 5 / Haiku, every radio unselected because nothing is tracked yet (no invented truth).

Model card

**3. Picking a model dispatches and tracks it honestly.** The card shows **"Sent to the agent — not confirmed"** with Opus 4.8 checked, and the pill row gains a second `Effort` pill now that a model is tracked — the same materialize-on-tracked behavior the desktop screenshots show.

Dispatched not confirmed
Pill row after pick

**4. Per-model options come from the catalog** — Effort (Low → Max, all unselected = unknown) plus a Fast mode section, matching desktop's effort menu.

Options card

**5. The slash catalog is agent-aware with descriptions.** Claude's real set — `/clear`, `/compact`, `/init`, `/review`, `/help` — and **no `/model`, `/cost`, or `/diff`**, which the old hardcoded provider-agnostic list advertised despite them not being in Claude's catalog.

Slash catalog

**6. The stuck-"Queued" regression, proven in both directions.** An ordinary chat send *does* still echo a Queued bubble (positive control, left), and a slash send adds *none* (right — the transcript still holds only the earlier "Hi"). Without the control the second shot would prove nothing, since a globally broken echo looks identical.


| chat send → bubble | slash send → no bubble |
| --- | --- |
| Control | Slash |

### Not exercised

The **stale-report re-entry** case (the second bug I fixed — see below) could not be proven in this rig: the mock reports no `agentStatus.model`, so `reportedModel` is null, the reported-model effect early-returns, and a tab round-trip would show the picked model whether or not the fix is present. That would have been a vacuous pass, so I'm calling it unexercised on-device rather than claiming it. It **is** pinned by a unit test that fails on the pre-fix code and passes after (`use-mobile-native-chat-session-options.test.ts`), and by a second test proving a genuinely *new* report still supersedes a local pick.

### Second fix pushed since the desktop comment

**A re-delivered identical status report reverted a just-dispatched model pick.** The reported-model effect keyed on `agentStatus.model` with `scopeKey` in its deps, and re-applied the value unconditionally. Re-entering a chat tab re-delivers the *same* session-start report — which cannot have observed a `/model` sent after it — so picking a model, switching tabs, and coming back reverted the pill to the reported model. A status-stream reconnect did the same. A report is now only evidence when the matched catalog id *changes* for that scope; a genuinely new report still wins. Mobile has no screen read to confirm a switch against, so "value changed" is the only honest signal available.

CI on `39a8098b`: 38 checks pass, 0 fail.

brennanb2025:
## Review gate: mobile QA addendum — reported-model truth rules

Follow-up to the mobile comment above. After establishing a reported model in the rig (the earlier shots had none, which is why the pill read a neutral `Model`), the truth-source rules are now exercised on-device:

**Seeded from the reported model.** With `agentStatus.model = claude-sonnet-5`, the pill resolves through the catalog to `Sonnet 5` and the per-model `Effort` pill appears — the model-matching path (exact id → label → provider-id containment) working end to end.

Seeded from report

**A reported value carries no "not confirmed" caption** — unlike a dispatched one. Reported *is* the agent's own truth, so claiming confirmation there is correct.

No caption when reported

**A genuinely new report supersedes a dispatched guess** (pill → `Fable 5`). This is the second direction of the repeat-report fix: the guard suppresses only a *repeat* of the same value, never a real change. Worth stating plainly because a one-sided guard would have been the easy way to get this wrong.

New report supersedes

**Effort applies and updates its pill**, and **the pills lock while the agent is working** (desktop parity — a mid-turn `/model` would interleave with the agent's own output).

Effort dispatched
Pills locked

### One thing to read correctly in the Queued pair above

In the control screenshot the "Hi" bubble sits at **Queued and stays there** — that is a property of the mock, not the defect this PR fixes. The mock never echoes a transcript at all, so nothing exists to retire any optimistic echo. What that shot proves is narrower and still exactly what's needed: a *chat* send does create a bubble, so the *absence* of one after a slash send is the classification gate working rather than echoes being globally broken. Retirement-by-transcript-match is unchanged by this PR and is covered by unit tests, not by that screenshot.

### Still not exercised on-device

The **repeat-report re-entry** case — pick a model with a different one reported, leave the tab, return, and confirm the pick survives. The rig didn't reproduce a re-delivered report across a tab round-trip, so I'm leaving it as unexercised rather than inferring a pass from the shots above. It is pinned in both directions by unit tests in `use-mobile-native-chat-session-options.test.ts` (one fails on the pre-fix code; the other proves a changed report still wins).

Rig teardown verified: working tree clean, the throwaway mock-server edit reverted, this run's Metro and mock stopped, and the neighbouring worktree's Metro plus the other QA simulators left untouched.

brennanb2025:
## Review gate: second reviewer pass — four more defects fixed

Two independent reviewer passes landed after the evidence comments above and found real problems the first pass missed. Fixed in `4df6a15e94`, each with a test I confirmed fails on the pre-fix code.

**1. A picker apply could interleave with a composer send.** The composer already blocks a text send while an apply is dispatching (`canSend` includes `!sessionOptionDispatching`), but not the reverse. The host writes a send's body, waits ~500ms, then writes the Enter — so an apply tapped inside that window landed between them, and the agent received one prompt reading `run the tests/model opus` followed by an empty submit. Worse, `setOption` still tracked the model as dispatched, so the pill claimed a change that never ran as a command.

The pickers render inside the composer, so they now take its in-flight state directly and disable on it — the same guard, mirrored. Deliberately scoped: the Ask-card answer and permission-reply paths are also ungated against this, but the right fix there is the shared per-terminal write lock #12373 introduces, and duplicating that mechanism here would conflict in the same files.

**2. An option could be filed under the wrong model.** `setTrackedSessionOption` resolves the owning model when it *commits*, not when the command was built, and the reported-model effect mutates the same record off the apply queue. So a report landing mid-dispatch caused `/effort low` — sent while Sonnet was live — to be recorded against Opus, and the options pill then read "Opus · Low" for a command Opus never received. Desktop guards exactly this (`native-chat-session-option-apply.ts:255-273`, with a comment naming the hazard); mobile had neither check. Ported.

**3. Typed-command values were tracked without validation.** A template's prefix also matches prose that merely starts with it, so `/model is a weird word` parsed out `is a weird word`, tracked it as the current model, rendered it as the pill label, and — matching no catalog model — dropped every per-model option, leaving the Effort pill gone until a report or remount healed it. `/effort of will is required` did the same to effort. Parsed values are now canonicalized against the catalog, and a typed value containing whitespace is treated as prose, so a multi-line paste whose first line reads `/model sonnet` tracks nothing rather than guessing. `/model opus`, `/model claude-sonnet-5` and `/model  sonnet` all still track correctly.

I had reviewed this exact question earlier and concluded no template could prefix-match arbitrary prose. That was wrong, and the second pass was right to refute it.

**4. Perf: `/` on a Codex tab mounted 45 rows.** `CODEX_COMMANDS` is 45 entries and the suggestion cap was 50, rendered into a plain `ScrollView` showing ~5 — so RN mounted every row (~130 views) directly above the transcript FlatList, and re-reconciled them on each 50ms streaming tick. Capped at 12.

### Also: the pickers file was over its line cap

Fix 1 pushed `MobileNativeChatSessionOptionPickers.tsx` to **402** effective lines against the 400 `.tsx` cap — it had been sitting at exactly 400 with zero headroom. Since raising the cap is forbidden, the row primitives moved to `MobileNativeChatSessionOptionRows.tsx`; the card is now 176 effective and the rows file 239.

### Reported and deliberately not fixed

- **Image-echo ordinal after an image+command send** — desktop has the identical gap (`NativeChatComposer.tsx:265,280`), so this is a shared-design question about the retirement contract rather than something this PR regressed.
- **`onAgentPicker` can flip a tab the wrong way** if the user manually switches views mid-dispatch — real, but needs view-mode plumbing outside this PR, and #12500 is reshaping that area.
- **`recordNativeChatSessionOptionCommand` returns `changed: true` for no-op recordings**, and **`applyNativeChatReportedSessionOptions` mutates on its unchanged path** — both accurate; both cost one extra render or an imprecise return with no consumer keying on it. Not worth changing shared code desktop also calls for no user-visible gain.
- **The slash menu is smaller for non-catalog agents** (grok empty, others 2 commands) — that's `getVerifiedNativeChatCommands`' own documented policy, shared with desktop. Mobile previously advertised a hardcoded list that was wrong for every agent; it now matches desktop.
- **A picker dispatch failure toasts "Message not sent"** — wrong copy for a model change. Worth a follow-up.

### Verification after these fixes

Mobile full suite 393 files / 2934 pass (3 skipped); desktop + shared native-chat 75 files / 696 pass; root and mobile `tsc --noEmit` clean (tsbuildinfo cleared); root and mobile `oxlint`/`oxfmt --check` clean; `check:react-doctor:changed` exit 0 — and the react-doctor plugin was confirmed to actually emit on a seeded violation rather than trusted for printing nothing. No `max-lines` disable or per-file bump anywhere in the diff; `mobile/pnpm-lock.yaml` untouched.

Test accounting across the earlier file moves: the two relocated files became 6 mobile files / 64 tests + 3 shared files / 20 tests, so all 17 moved tests still run under the root gate, nothing dropped.

brennanb2025:
## Corrections to the deferral notes above

Two of my "reported and deliberately not fixed" entries were framed in ways that would send a future reader down the wrong path. Correcting them, because the wrong framing is the kind that gets a ticket closed as a non-issue.

**1. The image-echo ordinal is a mobile-only regression, not desktop parity.** I wrote that "desktop has the identical gap." The *echo suppression* is desktop parity (`NativeChatComposer.tsx:265,280`), but the consequence is not: `countImageSourceTurnsAfter` is mobile-only, and desktop tracks a non-chat send by handle with no ordinal that can be stolen. So this **is** a narrow regression against mobile's own pre-PR behavior, where an image+command send created a pending entry with `images:[A]` that contributed to the ordinal. Anyone checking desktop for the mechanism will not find it there.

Still deferred — severity is a transient missing thumbnail during the transcript-lag window, self-healing once the real turn lands — but the fix shape belongs in the ticket: a non-echoing send that carried images must still *reserve* an ordinal (count images sent without an echo since the baseline and add that into the expected ordinal). And a dead end worth recording so nobody burns time on it: classifying image sends as chat regardless of the leading slash does **not** work, because the pending's text would be `/model sonnet`, which reconciles by text match and never lands — reintroducing exactly the stuck-"Queued" bug this PR exists to fix.

**2. The `onAgentPicker` view flip understates its own main path.** I described the worst case as "a background tab's view mode flipping." The primary path is worse and involves the *active* tab: the user sends bare `/model` on tab A, flips A to the terminal themselves to see the picker, and the dispatch's completion flips A back to chat — yanking them away from the TUI selector the feature just told them to look at. The background-tab case exists too (if they switched tabs mid-flight), but the active-tab yank is the one to fix, since it's the same interaction the feature directs the user to perform. Still low severity, still correctly gated behind the view-mode plumbing #12500 is reshaping.

## On the mobile screenshots

Do not read the iOS screenshots as pinned to a specific commit. The branch moved during that run (`fb3e18fcdc` and `39a8098b2e` both landed inside the window) with Fast Refresh active, and bundle identity was not verified per shot. What they show is unaffected by those commits — the pill row, the shared catalog, the agent-aware slash list, and the chat-vs-slash Queued pair — but I'm not claiming any of them demonstrates pre- or post-fix state for the two truth-rule fixes.

The **repeat-report re-entry** case remains unproven on-device, and there is now a second reason beyond the earlier one: `appliedReportByScope` and `recordsByScope` are module-level, so any Fast Refresh re-evaluation resets them, after which re-entering a tab legitimately re-applies the report. On screen that is indistinguishable from the regression, which makes a single dev-mode run unable to tell the fix from the artifact. This is a dev-only measurement problem — production has no Fast Refresh, and a real JS reload only happens on app restart, which resets those maps anyway — but it means the honest status stays "unit-pinned in both directions, not proven on device."

CI on `4df6a15e94`: 42 pass, 0 fail, `MERGEABLE` / `CLEAN`.

brennanb2025:
## The one gap is now closed — repeat-report re-entry proven on iOS, both directions

I twice reported this case as unexercised on-device. It has now been measured on a cleanly-pinned run against **`4df6a15e94`** (Metro started after the pull, HEAD verified, full app reload rather than Fast Refresh, and file mtimes across `mobile/src/session` and `src/shared` compared before and after so any edit landing mid-window would have invalidated the run).

**With the fix — the pick survives.** Reported model establishes the precondition (`Sonnet 5`), the user picks `Opus 4.8`, leaves the chat tab, returns → still `Opus 4.8`.

| precondition (reported) | after picking | after leaving and returning |
| --- | --- | --- |
| precondition | picked | keeps |

**Negative control — the guard removed, and it reverts.** Same steps with the `appliedReportByScope` early return dropped: the return trip puts the pill back to `Sonnet 5`, which is exactly the regression.

| precondition | after picking | after returning (guard removed) |
| --- | --- | --- |
| neg precondition | neg picked | neg reverts |

The A/B is what makes this conclusive rather than suggestive. An earlier run had shown the same revert *with* the fix in place, which looked like a FAIL — but it was correctly withheld, because `appliedReportByScope` and `recordsByScope` are module-level and any Fast Refresh re-evaluation resets them, after which re-entering a tab legitimately re-applies the report. On screen that is indistinguishable from the regression. Only the paired negative control separates the fix from the dev-time artifact. (That reset path is dev-only: production has no Fast Refresh, and a real JS reload only happens on app restart, which resets those maps anyway.)

The source is back to `4df6a15e94` exactly — guard restored, throwaway mock rig reverted, working tree clean, verified against HEAD.

### Corrections to what I published earlier

- "**Unproven on device**" for this fix — **superseded**, it is now proven in both directions.
- The caveat that the mobile screenshots aren't commit-pinned still stands for shots 02–15. Shots 18–23 above **are** pinned to `4df6a15e94`.

CI on `4df6a15e94`: 42 pass, 0 fail, `MERGEABLE` / `CLEAN`.

brennanb2025:
## Scope of the device QA — what is and isn't covered

Closing note so the pinned run above isn't read as broader than it is.

**Device-exercised on `4df6a15e94`:** the repeat-report re-entry case, in both directions (the A/B above). The QA run verified the *running bundle* rather than trusting that the edit took effect — `curl` of the served `entry.bundle` confirmed `appliedReportByScope` present with `.set`/`.delete`/`.clear` but **no** `.get(...) === matched` guard on the negative-control build, so the two runs differ by exactly those three lines and nothing else.

**Device-exercised earlier, but not commit-pinned** (shots 02–15): pill row, shared catalog card, dispatched-not-confirmed caption, per-model options, agent-aware slash catalog, the chat-vs-slash Queued pair, reported-model seeding and supersede, picker lock while the agent works.

**NOT exercised on device — unit-verified only:** the four fixes from the second review round. Specifically:
- pills disabling while a composer send is in flight
- the mid-dispatch commit abort when the model moves
- typed-value canonicalization against the catalog
- the slash-suggestion cap

Each is covered by a test I confirmed fails on the pre-fix code, and two are structural (the suggestion cap is a one-argument change; the row extraction is a pure move), but none has a screenshot behind it. I'm not going to imply otherwise — the honest line is that the second round is test-verified, not QA'd.

Two earlier runs of the re-entry case were quarantined rather than published, in `~/orca-qa/pr-12366-VOID-confounded-do-not-publish/` with a note on why: both showed the revert *with* the fix present, because a Fast Refresh reset of the module-level caches reproduces the pre-fix behavior exactly. They are not evidence in either direction and are deliberately absent from the evidence branch.

Rig teardown verified independently: working tree identical to `4df6a15e94`, mock rig reverted, all QA ports free, the run's serve-sim helper killed, and the neighbouring worktree's Metro plus other agents' simulators and helpers untouched.

brennanb2025:
## Merged main — reconciled with STA-3330 (#12369) and the write lock

#12369 merged while this was open and put the branch in conflict. Resolving it was not mechanical: #12369 **changed the contract of the snapshot builder this PR had just moved into `src/shared`**, so the two dedup efforts had to be reconciled rather than merged.

### What #12369 changed underneath this PR

It stopped the snapshot builder from self-healing an unlisted tracked model into its own choice list, on the premise that *callers* reconcile it first — and added `withTrackedModel` in `native-chat-pty-session-options.ts` to do that. The reason is per-host discovery: a real `list_models` can return `opus[1m]`/`sonnet[1m]` and omit the plain `opus` alias, and without reconciliation the tracked id has no matching choice, which blanks the trigger *and* drops every option that model owns.

This PR's mobile hook was calling the builder with `catalog.models` raw — so post-merge it would have been the one caller not honoring the new contract.

### How I resolved it

**Shared the reconciler instead of leaving a second copy.** `withTrackedModel` is pure logic both platforms need, so it moved to `src/shared/native-chat-session-option-snapshot.ts` as `withTrackedNativeChatModel`. Desktop imports it rather than keeping its own; mobile routes **every** consumer through the same reconciled list — snapshot, command build, per-model option lookup, and typed-command recording — matching what desktop's `activeModels()` feeds. That's the same deduplication direction as the rest of this PR, applied to the piece #12369 had just written.

**Fixed a real interaction with my own typed-value validation.** The canonicalization I added earlier validated a typed `/model <x>` against the active model list. Against a discovered list that dropped the `opus` alias, that would have **rejected a legitimate command** — defeating exactly what #12369's reconciliation exists to preserve. It now falls back to the seed, so an alias the host CLI no longer lists is still trackable, while prose is still rejected. Pinned by a test using a discovered list with `opus` filtered out.

**Took the real write lock for the picker.** Main now carries `mobile-native-chat-terminal-write-lock.ts` from the STA-3333 splits, which `answerQuestion` uses. `dispatchCommand` now acquires it too — so the picker-vs-send interleave is fixed at the write layer rather than by my earlier UI-level guard. The disabled pills stay as the affordance; the lock is what makes it correct. This also closes the Ask-card/permission-reply direction I had explicitly scoped out as needing that lock.

### Test contract updated, not deleted

The shared snapshot test that asserted the old self-healing behavior would have been a silent semantic break. It now pins the new contract in three parts: the builder renders exactly the models it is given; an empty list yields an empty snapshot; and `withTrackedNativeChatModel` restores an unlisted tracked model **preferring the seed row**, so the model keeps its options rather than getting a bare label-only entry.

Worth noting: that break surfaced only because of the earlier fix moving those tests from `mobile/src/session/` into `src/shared/`. In their old location the root suite would never have run them against #12369's change.

### Verification after the merge

Root shared + desktop native-chat 76 files / 729 tests pass; mobile `src/session` 121 files / 999 tests pass; root and mobile `tsc --noEmit` clean (tsbuildinfo cleared); lint and format clean; `check:react-doctor:changed` exit 0.

One note on that last gate: it reported errors while the merge was staged-but-uncommitted, because the diff base skews to include everything main brought in. Every flagged error was in a file this PR never touches (`use-automation-list-search.ts`, `editor.ts`, `use-worktree-issue-link.ts`, …). After committing the merge it is exit 0.

Full-suite reds seen locally (`worktree-base-directory-poller`, `remote-runtime-pty-transport`, and similar socket/fuzz suites) are environment, not this change: none is in this PR's file set, `worktree-base-directory-poller` fails identically on a pristine `origin/main` worktree, and the transport suite passes in isolation and only fails under full-suite load.

brennanb2025:
## Mobile model-picker layout QA

Updated through 875d4fb77d to match the Claude mobile layout while retaining Orca styling and behavior:

- full-width composer input above the action row
- one combined model/options chip
- explicitly inset rounded card around the input and actions
- grouped model choices in Orca’s bottom drawer
- effort and other options exposed as drill-in summary rows

### Focused rounded composer

Focused rounded mobile composer

This shows the text field focused with a typed draft and visible caret. The input, model button, and actions sit inside one rounded card with surrounding inset space, rather than an edge-to-edge bottom bar.

### Model picker drawer

Model picker drawer

### Selected model and effort summary

Selected model with effort summary

Validation: focused mobile composer/session-option tests, mobile lint/typecheck, and live iOS interaction verified through the Orca emulator and Electron/CDP workflows.

PR #12367, opened 2026-08-04
fix(ssh): handle owner displacement and graceful shutdown

AmethystLiang:
## Summary

Fix SSH reconnect lockouts by allowing a reconnect with valid logical-owner proof to displace a half-open incumbent, while serializing overlapping owner publications behind a bounded retry. Also fence reset-created transports once committed app shutdown has started.

## What changed

- Keep `ownerLease` stable for the lifetime of one logical ownership claim.
- Keep `ownerGeneration` as the per-admission delivery fencing epoch.
- Match recovery on lease, client instance, and authenticated principal; an older generation can identify the same claim but receives a fresh generation when admitted.
- Preserve the two-phase publication fence and use coded `-32042` retry while one owner response is settling.
- Revoke and close a displaced incumbent only after the replacement grant publishes.
- Re-check the committed shutdown gate before `ssh:resetRelay` opens a replacement transport.
- Drain active SSH sessions and bounded in-flight work during quit.

## Reliability contract

- **Invariant (`agent-session.provider-ownership`):** one published SSH PTY owner generation has write authority at a time; reconnecting the same logical owner cannot be permanently rejected merely because another grant publication is settling.
- **Failure source:** network loss or sleep can leave the incumbent socket half-open, and closely spaced reconnect attempts can overlap while the first replacement response is still publishing.
- **Oracle:** deterministic session and relay tests hold the first replacement response pending, require the second attempt to receive the coded pending refusal, settle the first response, then require the second attempt to succeed using its original proof with a higher fencing generation. Rollback and connection-close interleavings retain the incumbent.
- **Gate:** `agent-session.provider-ownership` tracks the single-owner authority contract; `terminal-performance.output-backpressure-budget` covers the SSH relay response-settlement, generation-fencing, and reconnect paths. Both gates remain experimental/partial.
- **Provider/platform coverage:** direct SSH/relay is covered by deterministic tests. Local PTY, daemon PTY, WSL routing, folder workspaces, and Git providers are unaffected. The wire shape and persisted schema are unchanged, so older mobile clients see no new required field. Relay builds succeed for macOS, Linux, Windows, x64, and arm64.
- **Performance budget:** matching and admission remain O(1); no polling, scans, subprocesses, or unbounded state were added. The new pending-publication retry is capped at three seconds and stops when attempt authority or the channel is lost.
- **Diagnostics:** pending publication has a stable numeric code; relay loss, displacement, and terminal version failures keep their existing bounded logs and surfaced status.
- **Residual gaps:** no live physical half-open SSH connection is exercised locally. Final quit-snapshot ordering, full SSH-local shutdown budgeting, explicit held-owner taxonomy, and recovery-contract cleanup remain follow-up work.

## Verification

- 719 focused ownership, dispatcher, relay, recovery, shutdown, and persistence tests
- Node typecheck
- Changed-code quality gate
- Max-lines ratchet
- Relay build for all supported target bundles
- Existing PR package and Windows package checks

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -250,7 +250,7 @@ import { buildHeadlessAutomationWorktreeCreateArgs } from './automations/headles
 import { AgentAwakeService } from './agent-awake-service'
 import { registerSystemResumeBroadcast } from './system-resume-broadcast'
 import { settleTeardownWithinDeadline } from './quit-teardown-deadline'
-import { QuitTeardownStartGate } from './quit-teardown-start-gate'
+import { quitTeardownStartGate } from './quit-teardown-start-gate'
 import { detachAllSshSessionsForShutdown } from './ipc/ssh'
 import { PluginService } from './plugins/plugin-service'
 import { PluginKillListService } from './plugins/plugin-kill-list-service'
@@ -2943,7 +2943,6 @@ app.on('before-quit', () => {
 
 // Why: will-quit fires twice — first pass preventDefaults and runs teardown; second pass exits.
 let daemonDisconnectDone = false
-const quitTeardownStartGate = new QuitTeardownStartGate()
 app.on('will-quit', (e) => {
   // Why return instead of re-running teardown: the second pass is Electron re-firing after
   // our own app.quit(), so every step below already ran and every durable write already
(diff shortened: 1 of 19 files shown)


AmethystLiang:
@coderabbitai Fixing. Will be in the next commit

PR #12369, opened 2026-08-04
Track Claude models from the installed CLI per host (STA-3330)

brennanb2025:
## Summary

Fixes [STA-3330](https://linear.app/stably/issue/STA-3330/claude-model-catalog-is-hardcoded-and-missing-opus-5): the Claude model catalog was a hardcoded seed (labeled `opus` as "Opus 4.8", no Opus 5) that drifted on every model release. The catalog now tracks the installed Claude CLI per host:

- **`listModels` for the Claude catalog** (mirroring the existing Cursor pattern): discovery sends one `list_models` control request to `claude -p --input-format stream-json --output-format stream-json --verbose` and parses the control response. This returns the CLI's own `/model` picker catalog — accepted values (e.g. `opus[1m]`, `sonnet`), display names, per-host resolution ("Opus 5 with 1M context …"), effort levels, and fast-mode support — **without starting an API turn** (~1s, exit 0, no inference cost).
- **Seed corrected to never lie on any host.** `--model` aliases resolve to the newest model of each family *on that host's CLI* — verified live: on claude 2.1.220 `opus` resolves to Opus 5 (the CLI's own constants read `OPUS_ID: "claude-opus-5"`), so the old "Opus 4.8" label was wrong on current hosts, and a hardcoded "Opus 5" label would be wrong on older hosts. The seed now uses family labels (Fable / Opus / Sonnet / Haiku) with role descriptions; discovery overlays exact per-host names and descriptions, which is where "Opus 5" becomes visible to users.
- **Fallback verified on a real old CLI.** claude 2.1.100 (probed via a pinned npx run) answers `{"subtype":"error","error":"Unsupported control request subtype: list_models"}` and exits 0 → parse yields nothing → the seed stays. Discovery state is host-scoped through the existing enrichment flow (`hostKey` covers native/WSL/SSH, same principle as `GitCapabilityCache`).
- **Source Control AI benefits too**: the commit-message spec for Claude switches to `modelSource: 'dynamic'` with the same probe (the discovery plan now carries a `stdinPayload`, honored by the local spawner and already supported by the SSH executor). Unknown persisted ids pass through like Codex's.
- **Cross-version detection hardening**: the `/model` switch-success detector now also matches the model-family prefix, because the CLI echoes the *resolved* name ("Set model to Opus 5 (1M context)") which drifts from picker labels across versions. The terminal-header scrape keeps working on both old ("Opus 4.8 …") and new ("Opus 5 …") headers via the family labels.

Default model behavior is unchanged (`sonnet` stays `isDefault`). Model ids persisted by users (`fable`/`opus`/`sonnet`/`haiku`) are unchanged. Desktop and mobile consume the same shared catalog; changes are additive (optional `description` field), and no mobile files are touched.

### CLI capability verification (hard rule)

| Check | Result |
| --- | --- |
| `claude --help` (2.1.220) | `--model` takes "an alias for the latest model (e.g. 'fable', 'opus', or 'sonnet') or a model's full name" — no listing subcommand exists (`claude models` starts a chat session) |
| `list_models` control request on 2.1.220 | Full picker catalog: `default`, `opus[1m]`, `sonnet`, `sonnet[1m]`, `fable`, `haiku` with resolutions, effort levels (`low…max`), fast-mode flags |
| `list_models` on 2.1.100 (npx-pinned) | Control error, exit 0 → seed fallback path |
| `--model 'sonnet[1m]'` live call | Accepted; `modelUsage` reports `claude-sonnet-5[1m]` — bracketed picker values are valid `--model` inputs |
| Invalid model | Fast failure, no enumeration in the error |

## Screenshots

No screenshot (headless worker session). Visible changes in the Claude model picker:

- Before: `Fable 5 / Opus 4.8 / Sonnet 5 / Haiku` (labels wrong on current CLIs where `opus` is Opus 5).
- After (before discovery lands, or when it fails): `Fable / Opus / Sonnet / Haiku` with role descriptions.
- After discovery on a current CLI: matched rows adopt the CLI's names/descriptions (e.g. Sonnet — "Sonnet 5 · Efficient for routine tasks · $2/$10 per Mtok") and account-specific variants are appended (e.g. "Opus (1M context)" — "Opus 5 with 1M context …").

## Testing

- [x] `pnpm lint` (full pipeline: oxlint, native + type-aware audits, reliability gates, max-lines ratchet, skill-guide + localization checks; changed-code quality and React Doctor changed-lines gates both report 0 new findings)
- [x] `pnpm typecheck` (also `typecheck:tsc` with `--composite false` to rule out stale tsbuildinfo passes)
- [ ] `pnpm test` — full suite left to CI (local Node 26 environment produces known unrelated false reds); all affected suites run green with the explicit config: `claude-model-list-probe`, `agent-session-option-catalog`, `commit-message-agent-spec`, `commit-message-text-generation`, `runtime-git-client`, `rpc/methods/git`, and the full `native-chat`/`session-option`/`commit-message`/`source-control` sweep (171 files / 1713 tests)
- [ ] `pnpm build` — not run; no build-affecting changes (no new deps, no packaging changes)
- [x] Added or updated high-quality tests that would catch regressions: parser (live-captured fixtures from 2.1.220 and the 2.1.100 error shape, CRLF/init-noise/dedupe/hostile-nesting cases), catalog parse → options mapping (effort-level subsets, fast mode, `default` row dropped), merge behavior (seed order, label/description overlay, options preservation, appended variants), seed-fallback on unsupported CLIs (catalog, spec, and local-discovery layers), stdin delivery of the probe request, resolved-echo switch detection, and version-agnostic header scraping

## AI Review Report

Self-review focused on:

- **Mixed CLI versions across hosts** — the central risk. Verified live on 2.1.220 and 2.1.100 that both outcomes (catalog / control error) exit 0 and that every failure mode (non-zero exit, empty output, malformed JSON, unsupported subtype) lands on the seed. Discovery is once-per-host via the existing `hostKey`-scoped enrichment.
- **Cross-platform (macOS/Linux/Windows/WSL/SSH)**: the local spawner pipes stdin only when the plan carries a payload (`stdio` stays `ignore` otherwise); Windows keeps the existing `resolveCliCommand`/`getSpawnArgsForWindows` cmd-shim routing (stdin piping is unaffected by the cmd wrapper); the WSL branch passes the same stdio tuple; the SSH executor already forwarded `plan.stdinPayload` for stdin prompt delivery. No paths, no shortcuts, no platform-specific labels. An EPIPE handler covers CLIs that exit before reading stdin (flag rejection on very old versions).
- **Detection regressions**: switch-echo detection previously required the label to be a prefix of the echoed resolved name — enriched labels like "Opus (1M context)" would have made every verified switch report as unverifiable (the `unknown` outcome throws in the UI). The family-prefix fallback closes this; test added with the resolved-echo fixture.
- **RPC/mobile compatibility**: only an additive optional `description` on `CommitMessageModel`/capability types; RPC results are passed through unvalidated on the response path; no mobile files touched and mobile has no imports of the changed modules in this branch.
- Flagged and deliberately deferred: newly discovered variant rows (e.g. `opus[1m]`) stay model-only (no effort/fast sub-options) until cataloged — same accepted behavior as Cursor discovery; noted in Notes.

## Security Audit

- **Command execution**: discovery spawns the agent binary with a fixed argv (no shell string interpolation on the executed path) through the same plumbing as existing generation/discovery; the agent-binary override goes through the existing `planAgentBinary` validation. The `listModels.command` string on the catalog is display/reference only — nothing executes it.
- **Input handling**: stdin payload is a compile-time constant JSON line. Parser output is untrusted: each candidate line is bounds-checked (`assertJsonTextStructureWithinLimits`, 64K structural tokens / depth 16) before `JSON.parse`, entries are shape-checked field-by-field, deduped, and total process output is already capped by `MAX_AGENT_OUTPUT_BYTES` with kill-on-overflow.
- **No secrets, no auth changes, no new IPC surface** (existing `git.discoverCommitMessageModels` route reused), no new dependencies. The probe runs the user's own CLI under their existing credentials, identical trust model to commit-message generation.

## Notes

- On hosts where the account exposes only bracketed variants (e.g. `opus[1m]`), the base `opus` seed row remains available — both are valid `--model` values (verified live).
- Follow-up candidate: catalog effort/fast-mode sub-options for newly discovered variant ids (they currently render model-only, matching Cursor discovery behavior), and piping `supportedEffortLevels` into Source Control AI's per-model effort dropdown for variants.
- The probe creates a throwaway `claude -p` session per host per app run (~1s); if session-file noise ever matters, `--no-session-persistence` could be added once the fleet's minimum CLI version is known to accept it.

### DIFF
--- a/src/main/ipc/filesystem.test.ts
+++ b/src/main/ipc/filesystem.test.ts
@@ -2419,6 +2419,111 @@ describe('registerFilesystemHandlers', () => {
     )
   })
 
+  it('discovers models from an exact repo-less folder workspace root', async () => {
+    const folderPath = path.resolve('/outside-workspace/folder-project')
+    const folderStore = {
+      ...store,
+      getFolderWorkspaces: () => [
+        {
+          id: 'folder-1',
+          projectGroupId: 'group-1',
+          folderPath,
+          connectionId: null
+        }
+      ]
+    }
+    discoverCommitMessageModelsLocalMock.mockResolvedValue({
+      success: true,
+      models: [{ id: 'sonnet', label: 'Sonnet' }],
+      defaultModelId: 'sonnet'
+    })
+
+    registerFilesystemHandlers(folderStore as never)
+
+    await handlers.get('git:discoverCommitMessageModels')!(null, {
+      agentId: 'claude',
+      worktreePath: folderPath
+    })
+
+    expect(discoverCommitMessageModelsLocalMock).toHaveBeenCalledWith(
+      'claude',
+      undefined,
+      undefined,
+      { cwd: folderPath }
+    )
+  })
+
+  it('does not authorize remote-only folder roots as local discovery paths', async () => {
+    const folderPath = path.resolve('/remote-only/folder-project')
+    const folderStore = {
+      ...store,
+      getFolderWorkspaces: () => [
+        {
+          id: 'folder-1',
+          projectGroupId: 'group-1',
+          folderPath,
+          connectionId: 'ssh-1'
+        }
+      ]
+    }
+
+    registerFilesystemHandlers(folderStore as never)
+
+    await expect(
+      handlers.get('git:discoverCommitMessageModels')!(null, {
+        agentId: 'claude',
+        worktreePath: folderPath
+      })
+    ).rejects.toThrow('Access denied')
+    expect(discoverCommitMessageModelsLocalMock).not.toHaveBeenCalled()
+  })
+
+  it('routes a repo-less WSL folder workspace discovery through its distro', async () => {
+    await withPlatform('win32', async () => {
+      const folderPath = '\\\\wsl.localhost\\Ubuntu\\home\\tester\\folder-project'
+      const prepareForClaudeLaunch = vi.fn().mockResolvedValue({
+        configDir: '\\\\wsl.localhost\\Ubuntu\\home\\tester\\.claude',
+        envPatch: { CLAUDE_CONFIG_DIR: '/home/tester/.claude' },
+        stripAuthEnv: true,
+        provenance: 'managed:account-1'
+      })
+      const folderStore = {
+        ...store,
+        getFolderWorkspaces: () => [
+          {
+            id: 'folder-1',
+            projectGroupId: 'group-1',
+            folderPath,
+            connectionId: null
+          }
+        ]
+      }
+      discoverCommitMessageModelsLocalMock.mockResolvedValue({
+        success: true,
+        models: [{ id: 'sonnet', label: 'Sonnet' }],
+        defaultModelId: 'sonnet'
+      })
+
+      registerFilesystemHandlers(folderStore as never, { prepareForClaudeLaunch })
+
+      await handlers.get('git:discoverCommitMessageModels')!(null, {
+        agentId: 'claude',
+        worktreePath: folderPath
+      })
+
+      expect(prepareForClaudeLaunch).toHaveBeenCalledWith({
+        runtime: 'wsl',
+        wslDistro: 'Ubuntu'
+      })
+      expect(discoverCommitMessageModelsLocalMock).toHaveBeenCalledWith(
+        'claude',
+        expect.objectContaining({ CLAUDE_CONFIG_DIR: '/home/tester/.claude' }),
+        undefined,
+        { cwd: path.resolve(folderPath), wslDistro: 'Ubuntu' }
+      )
+    })
+  })
+
   it('routes local WSL project model discovery through the project runtime target', async () => {
     await withPlatform('win32', async () => {
       discoverCommitMessageModelsLocalMock.mockResolvedValue({
--- a/src/main/ipc/filesystem.ts
+++ b/src/main/ipc/filesystem.ts
@@ -409,6 +409,26 @@ function getLocalAgentRuntimeTarget(
     : { runtime: 'host' }
 }
 
+async function resolveModelDiscoveryLocalPath(
+  store: Store,
+  requestedPath: string
+): Promise<string> {
+  try {
+    return await resolveRegisteredWorktreePath(requestedPath, store)
+  } catch (error) {
+    const folderWorkspaces =
+      typeof store.getFolderWorkspaces === 'function' ? store.getFolderWorkspaces() : []
+    const isFolderWorkspaceRoot = folderWorkspaces.some(
+      (workspace) =>
+        comparableLocalPath(workspace.folderPath) === comparableLocalPath(requestedPath)
+    )
+    if (!isFolderWorkspaceRoot) {
+      throw error
+    }
+    return resolveAuthorizedPath(requestedPath, store)
+  }
+}
+
 function getLocalTextGenerationTarget(
   worktreePath: string,
   gitOptions: LocalProjectWorktreeGitOptions,
@@ -1517,16 +1537,17 @@ export function registerFilesystemHandlers(
       let localRuntimeTarget: CommitMessageAgentRuntimeTarget = { runtime: 'host' }
       let localDiscoveryOptions: Parameters<typeof discoverCommitMessageModelsLocal>[3]
       if (args.worktreePath) {
-        const worktreePath = await resolveRegisteredWorktreePath(args.worktreePath, store)
+        const worktreePath = await resolveModelDiscoveryLocalPath(store, args.worktreePath)
         const gitOptions = getLocalGitOptionsForRegisteredWorktree(
           store,
           args.worktreePath,
           worktreePath
         )
-        localRuntimeTarget = getLocalAgentRuntimeTarget(gitOptions)
-        localDiscoveryOptions = gitOptions.wslDistro
-          ? { cwd: worktreePath, wslDistro: gitOptions.wslDistro }
-          : { cwd: worktreePath }
+        const wslDistro = gitOptions.wslDistro ?? parseWslPath(args.worktreePath)?.distro
+        localRuntimeTarget = wslDistro
+          ? { runtime: 'wsl', wslDistro }
+          : getLocalAgentRuntimeTarget(gitOptions)
+        localDiscoveryOptions = wslDistro ? { cwd: worktreePath, wslDistro } : { cwd: worktreePath }
       }
       const localEnv = await prepareLocalCommitMessageAgentEnv(
         agentId,
--- a/src/main/text-generation/commit-message-text-generation.test.ts
+++ b/src/main/text-generation/commit-message-text-generation.test.ts
@@ -311,6 +311,7 @@ describe('discoverCommitMessageModelsLocal', () => {
 
     expect(result).toMatchObject({
       success: true,
+      catalogOrigin: 'spec',
       defaultModelId: 'smart'
     })
     expect(spawnMock).not.toHaveBeenCalled()
@@ -390,6 +391,7 @@ describe('discoverCommitMessageModelsLocal', () => {
 
     await expect(pending).resolves.toMatchObject({
       success: true,
+      catalogOrigin: 'probe',
       defaultModelId: 'sonnet',
       models: [
         { id: 'opus[1m]', label: 'Opus (1M context)' },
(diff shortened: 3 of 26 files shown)


brennanb2025:
### Electron QA — `9ff7764915`

Validated the exact PR worktree in an isolated Orca dev instance (`http://127.0.0.1:5174`, Claude CLI `2.1.220`). Launched a real Claude Code session, opened the native-chat Model picker, and selected `opus[1m]`. The UI confirmed `Ran /model opus[1m]`.

The live CLI catalog shows `Opus (1M context)` with the description `Opus 5 with 1M context`, plus current Sonnet/Haiku metadata:

Claude model catalog

The selected discovered variant exposes all reported effort levels and Fast mode:

Claude variant options

Both screenshots are uncropped full-window captures. No layout clipping, overlap, or console error from the tested interaction was observed.

brennanb2025:
### Electron QA — `da3c171cab`

Validated the exact PR worktree in an isolated CDP-enabled Orca dev instance with Claude Code `2.1.220`. A newly launched real Claude session populated native chat from the installed CLI probe.

The Model picker contains exactly the five live rows returned on this host: `Opus (1M context)`, `Sonnet`, `Sonnet 5 (1M context)`, `Fable`, and `Haiku`. The obsolete hardcoded plain `Opus` fallback is no longer present. The full-window capture shows the complete picker with no clipping or overlap.

Claude live model catalog without hardcoded fallback

Lunivo:
One housekeeping note: the description links STA-3330 but no GitHub issues, so nothing here will auto-close on merge. Adding the keywords to the PR body would let this sweep the backlog it actually clears:

```
Fixes #12084
Fixes #10812
```

- **#12084** — *Model picker still labels the "opus" entry as "Opus 4.8"*. Directly fixed: family labels plus the per-host name overlaid from `list_models` means the row can't display a stale version again.
- **#10812** — *Add Claude 5-family models to the Claude model catalog*. Its actual ask is "ideally sourced dynamically from the installed `claude` CLI's available models rather than a static bundled list, so future model releases don't require an Orca app update" — which is exactly this PR. (The 5-family models themselves were already reachable, since the aliases resolve to them; only the labels lied. So this closes it on the merits, not just the title.)

**PRs superseded by this one:**

- **#12102** — makes the `opus` label versionless. Same line, same two test files. Its goal (never stale again) is what discovery achieves properly; the versionless string on its own would break header recovery once previous-generation rows exist. Worth closing when this lands, with a note to the author — it's their second contribution and the diagnosis in it was sound.

**Not superseded — needs a rebase rather than a close:**

- **#10836** — recognizes *custom* Claude models in the picker (models absent from the catalog entirely). Discovery shrinks that surface but doesn't remove it, since genuinely custom slugs won't appear in `list_models` either. It touches `claude-terminal-session-options.ts`, so it will conflict with this branch.

brennanb2025:
### Review + Electron QA — `f1ef49e82b`

Reviewed with adversarial subagent passes and validated live in an isolated CDP-attached Orca dev instance built from this exact worktree (identity verified as `sta-3330-claude-model-catalog` before every interaction), against Claude CLI `2.1.220`.

#### Two regressions found and fixed in `f1ef49e82b`

**1. Tracked model could fall outside the discovered list.** `readClaudeSessionOptionsFromTerminalScreen` matches the terminal header against the *static* seed, so the header `Opus 5 (1M context) with xhigh effort` yields tracked id `opus` — but this account's live `list_models` returns `opus[1m]`, `sonnet`, `sonnet[1m]`, `fable`, `haiku` and **no plain `opus`**. Combined with `includeTrackedModelChoice: agent !== 'claude'`, the snapshot builder produced:

```
choices:      opus[1m], sonnet, sonnet[1m], fable, haiku
currentValue: "opus"      ← no matching choice → blank trigger
descriptors:  [model]     ← effort and fast mode dropped entirely
```

A session genuinely running Opus showed a blank model trigger and lost its effort control. Fixed by reconciling the tracked id into the active list, preferring the seed row so the model keeps its options.

**2. The picker rendered nothing while pending, and never on failure.** `initialModels: []` plus the empty-list short-circuit meant the whole model/effort/fast-mode group popped in ~1s after mount, and stayed absent permanently when the probe yields nothing — an old CLI without `list_models`, no `claude` on PATH, or an older remote runtime whose response omits the new `catalogOrigin` field (SSH users on an older remote Orca lost the picker outright). Fixed by restoring the version-neutral family seed as the starting list; discovery still replaces it wholesale on success, so a host with a real catalog never shows an obsolete hardcoded row.

Regression tests were confirmed non-vacuous by reverting the fix and observing them fail.

#### Live evidence

Model picker populated from this host's CLI — real per-host names and descriptions, no hardcoded labels:

Claude model picker with live per-host catalog

Selecting the discovered variant dispatches correctly (`Ran /model opus[1m]` in the transcript) and the row carries its reported effort levels:

Selected discovered model with effort control

This also exercises the resolved-echo switch detection: the CLI echoes the resolved name, and the family-prefix matcher confirms the switch rather than reporting it unverifiable.

Both captures are uncropped full-window. No clipping, overlap, or console error from the tested interactions.

#### Checks

- `pnpm typecheck` — pass
- `pnpm lint` — pass (full pipeline)
- `pnpm vitest run --config config/vitest.config.ts` over `native-chat/`, `shared/`, `text-generation/` — 65 files / 682 tests pass
- Probe cost re-verified: one spawn per `(agent, hostKey)` per renderer lifetime; the enrichment guard prevents re-spawn on workspace switch or composer remount, and the existing timeout + 4 MB output cap wrap it.

#### Not covered

- Windows and WSL discovery paths were not exercised live (macOS host only); they are covered by unit tests.
- A settled probe failure is never retried for the app's lifetime. Acceptable now that the seed always renders, but worth knowing.

brennanb2025:
### Windows validation — `f1ef49e82b`

Ran this branch on a real Windows host (not emulated, not WSL) to close the Windows gap I flagged earlier. The branch was checked out there at the exact fix commit, `pnpm install` run, and the Electron dev app launched and CDP-attached, with instance identity verified as the branch worktree before any assertion.

Host: Windows 11, PowerShell 7.6 Core, node 24.18.0, pnpm 10.24.0, git 2.55.0.windows.2, **Claude CLI 2.1.221**.

**Result — the Windows discovery path works.** Calling the real IPC in the running app on that host:

```
git.discoverCommitMessageModels({ agentId: 'claude', worktreePath: <windows worktree> })
→ { ok: true, origin: "probe", ids: ["opus[1m]", "sonnet", "sonnet[1m]", "haiku"] }
```

`origin: "probe"` is the load-bearing assertion. It can only be `probe` if the `list_models` control request was actually written to the CLI's stdin through the Windows `cmd` shim routing (`resolveCliCommand` → `getSpawnArgsForWindows` → `stdio: ['pipe', ...]` → `stdin.end(payload)`) **and** the control response parsed. Had the stdin payload not reached the CLI, this would have fallen back to `origin: "spec"` with the seed ids. The returned ids are the host's real bracketed picker values, which also confirms the parser handles the live Windows output.

Independently, the same control request run directly against the CLI on that host returns the expected catalog, matching macOS:

```
{"type":"control_response","response":{"subtype":"success", ... "value":"opus[1m]",
 "displayName":"Opus (1M context)","description":"Opus 5 with 1M context · ..."}
```

Note this host's account lists no plain `opus` and no `fable` — different entitlements from the macOS host — which is a useful second data point for the reconciliation fix in `f1ef49e82b`: a tracked family alias like `opus` is genuinely absent from real hosts' catalogs, so the picker must reconcile it rather than drop it.

#### Incidental finding (not caused by this PR)

`pnpm install` exits 1 on a Windows host without MSVC build tools: `windows-native-registry` has no prebuilt binary for this ABI and node-gyp fails with "Could not find any Visual Studio installation to use". The dev app still boots, because that module is `require`d lazily and only for reading the user PATH from the registry. Pre-existing, unrelated to this change, but it means a clean Windows dev setup currently needs the VS C++ workload.

#### Not covered

- No Windows UI screenshot. The app ran and was driven headlessly via CDP, but the machine is shared and heavily contended (other Electron instances and automation running concurrently), and the settings/native-chat navigation did not take within a reasonable number of attempts. The functional assertion above covers the Windows-specific code; the UI rendering path is shared with macOS, which is screenshotted in the previous comment.
- The WSL execution branch of the discovery spawn was not exercised.

PR #12374, opened 2026-08-04
fix(mobile): keep relay runtime recovery alive without direct connectivity

OrcaWin:
## Field failure (verified on a physical Android device)

A phone paired **over the relay** whose baked-in direct LAN endpoint is unreachable (a Tailscale IP with Tailscale off on the phone) permanently loses the runtime channel:

- the `[net]` client dials only the LAN endpoint (0.5/1/2/4s backoff, then trickle) and the relay runtime session never starts or never comes back;
- the home card shows **"Connected · Orca Relay"** (overlay data) or **"Can't connect — check Tailscale"** while the host page sits at **0 worktrees** forever;
- background/foreground cycles and app restarts do not recover.

## Root causes

1. **Terminal recovery gates.** `RelayReconnectController`'s `fresh-credential` gate (set on a 4401 rejected outer credential) parked with no timer, no log, and no exit that works without direct connectivity — on a relay-only phone that is a deadlock. The `external-signal` gate (E2EE rejection — which fires transiently right after pairing while the desktop commits credentials) waited indefinitely for a UI nudge.
2. **One-shot credential snapshot.** The supervisor read the credential bundle once at `start()`: a Keychain race killed relay recovery for the process lifetime, and credentials rotated after start left it dialing stale tokens (feeding the 4401 gate).
3. **Zero observability.** The entire relay dial path had no logging, so all of the above was invisible in logcat and the in-app connection log.

## Fix

- Gates arm a slow reprobe instead of parking: **60s escalating ×2 to a 15min ceiling, jittered 0.75–1.25×** (no fleet phase-alignment; a permanently revoked device is not a forever one-minute beacon). Clearing a gate also drops its timer, pending tick, and cadence, so an orphaned reprobe can never swallow the next fast backoff.
- Each gated attempt **re-reads the durable bundle** and adopts it **by outcome, not by version**: the disk bundle is adopted exactly when it yields a dialable (unexpired, non-rejected) credential while the in-memory one does not. Version comparison is unusable here — renewals extend `expiresAt` without bumping `current.version`, and a re-pair restarts the counter (both found in adversarial review, with measured repros). Rejected versions are still never redialed, and a revoked version cannot be resurrected from a stale disk copy.
- A merely missing/expired bundle uses a plain cooldown rather than the `fresh-credential` gate, so link flap cannot force credential rotations on direct reconnects.
- Supervisor `start()` subscribes even when the initial bundle read fails or the bundle is expired.
- Every recovery decision emits `[relay]` lines to logcat and the in-app connection log.
- Structural: direct-return probing → `mobile-direct-return-probe.ts`, credential selection → `mobile-relay-credential-selection.ts`.

## Tests

`mobile-relay-runtime-failover.test.ts` mirrors the device evidence: rejected-outer-credential recovery, unreadable-bundle-at-start, expired-bundle-then-fresh-write, E2EE-rejection-without-UI-nudge, renewal-without-version-bump, re-pair-with-restarted-counter, background/foreground recovery, plus two real-`rpc-client` failover integration tests. Controller suite adds escalating-cadence and orphaned-timer regressions. **All new scenarios verified red on the pre-fix code.** Full mobile suite: 2903 passed; oxlint and tsc clean.

## Review

Adversarially reviewed (independent session with a scratch harness): two measured blockers in the original version-comparison rule were found and fixed; churn analysis confirms the gated steady-state is a keychain read per escalating tick, not a dial loop (no repeat of the reconnect-lane monopolization class).

## Follow-ups (out of scope, noted)

- client-context host entries are built once from a profile snapshot; a host that gains `relay` after its entry opened keeps the direct-upgrade controller until restart.
- migrate-timeout behavior under the real 12s `waitForAuthenticated` is untested.

## Verification plan

CI APK from this branch goes onto the affected physical device (relay-only, Tailscale off) before merge; acceptance is the host page loading real worktrees over "Connected · Orca Relay".

### DIFF
--- a/mobile/src/transport/mobile-direct-return-probe.ts
+++ b/mobile/src/transport/mobile-direct-return-probe.ts
@@ -1,0 +1,83 @@
+import { openAuthenticatedDirectEndpoint } from './mobile-direct-endpoint-probe'
+import type { MobileEndpointHysteresis } from './mobile-endpoint-hysteresis'
+import type { RpcClient } from './rpc-client'
+import type { HostProfile } from './types'
+import type { MobileConnectionPath } from './stable-logical-rpc-client'
+
+const DIRECT_PROBE_INTERVAL_MS = 15_000
+
+// While the runtime channel rides the relay, periodically probe the direct
+// endpoint and migrate back once hysteresis proves it stable.
+export class DirectReturnProbe {
+  private timer: ReturnType<typeof setTimeout> | null = null
+
+  constructor(
+    private readonly deps: {
+      now: () => number
+      setTimer: typeof setTimeout
+      clearTimer: typeof clearTimeout
+      openDirect: (endpoint: string) => RpcClient
+    },
+    private readonly hooks: {
+      hysteresis: MobileEndpointHysteresis
+      host: () => HostProfile
+      canSchedule: () => boolean
+      canAttempt: () => boolean
+      beginOperation: () => void
+      migrate: (client: RpcClient, path: MobileConnectionPath) => Promise<void>
+      onDirectMigrated: () => Promise<void>
+      afterProbe: () => void
+    }
+  ) {}
+
+  schedule(delayMs = DIRECT_PROBE_INTERVAL_MS): void {
+    if (!this.hooks.canSchedule() || this.timer) {
+      return
+    }
+    this.timer = this.deps.setTimer(() => {
+      this.timer = null
+      void this.probe()
+    }, delayMs)
+  }
+
+  clear(): void {
+    if (this.timer) {
+      this.deps.clearTimer(this.timer)
+      this.timer = null
+    }
+  }
+
+  private async probe(): Promise<void> {
+    if (!this.hooks.canAttempt() || !this.hooks.hysteresis.canProbe(this.deps.now())) {
+      this.schedule()
+      return
+    }
+    this.hooks.beginOperation()
+    let successful: Awaited<ReturnType<typeof openAuthenticatedDirectEndpoint>> = null
+    try {
+      successful = await openAuthenticatedDirectEndpoint(
+        this.hooks.host(),
+        this.deps.openDirect,
+        12_000
+      )
+      if (!successful) {
+        this.hooks.hysteresis.recordDirectFailure(this.deps.now())
+        return
+      }
+      if (!this.hooks.hysteresis.recordDirectSuccess(this.deps.now())) {
+        successful.client.close()
+        return
+      }
+      await this.hooks.migrate(successful.client, successful.path)
+      successful = null
+      this.hooks.hysteresis.recordMigration(this.deps.now())
+      await this.hooks.onDirectMigrated()
+    } finally {
+      successful?.client.close()
+      // Why: a relay drop or backoff timer can arrive while the probe owns the
+      // operation mutex; afterProbe releases it and replays deferred recovery.
+      this.hooks.afterProbe()
+      this.schedule()
+    }
+  }
+}
--- a/mobile/src/transport/mobile-endpoint-lifecycle.ts
+++ b/mobile/src/transport/mobile-endpoint-lifecycle.ts
@@ -90,6 +90,7 @@ function createSupervisor(
     readBundle: readMobileRelayCredentialBundle,
     writeBundle: writeMobileRelayCredentialBundle,
     saveHost,
+    onLog,
     now: Date.now,
     randomBytes: ExpoCrypto.getRandomBytes,
     setTimer: setTimeout,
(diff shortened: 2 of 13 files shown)


OrcaWin:
Adversarial review round 3 (final): merge candidate `1302bf540b` verified clean — measured gated attempt ladder is exactly the documented 60/180/420/900/1800/2700/3600s, the resume-latency regression is gone (first post-resume dial at 60s, was 675s), and background holds zero timers on the clean path.

Two pre-existing non-blocking notes for follow-up (deliberately not respun into this PR so the phone-verified binary matches the merge commit):

1. An `external-signal` (E2EE) gate never populates `rejectedCredentialVersions`, so `armCredentialReprobe` with no dialable bundle takes the cooldown branch under a held gate — every second wakeup is a no-op re-arm and the effective attempt interval settles at ~30min instead of 15 at ceiling. Strictly better than before this PR (that branch previously never minted a gate token at all). One-line follow-up: take the token-minting branch when `recoveryGate` is already set.
2. `registerFailure`'s gate branches arm a reprobe timer even when the supervisor passed `scheduleRetry=false` (background mid-dial), so one tick can fire in background / after `stop()`. Measured consequences are benign (token minted but unconsumed, no re-arm, no churn; the leftover token makes the first post-resume dial immediate).

Verification note: with a credential gate held, the first relay dial after app-open lands 45–75s later by design — a quiet minute after resume is expected behavior, not the outage.

OrcaWin:
Live device verification of the failover fix exposed a third masked defect, now fixed in `c3b899451a`:

**Session churn off the attach deadline.** The relay-hello's `leaseExpiresAt` is the cell's attach-reservation deadline (`now + 10s` for resumes), but the supervisor scheduled proactive rotation from it with a 30s margin clamped to 1s — every relay runtime session force-replaced itself ~1s after connecting (measured every ~2.5s on device; 253 dials per 5 simulated minutes in the red test). Any RPC slower than the cycle died mid-flight, which is the "Worktree list unavailable" symptom. The old latch bug had hidden this entirely: sessions rarely lived long enough to churn.

Fix: the session captures `resumeExpiresAt` from the resume hello (kept fresh by the resume confirmation) and rotation keys off it via new `getResumeExpiresAt()`. Test fakes previously used a fictional 120s lease — they now mirror the production 10s attach deadline, plus a churn regression that holds one session across 5 minutes with direct unreachable.

Server-side follow-up worth filing: sending the reservation attach deadline under the name `leaseExpiresAt` in the resume hello invites exactly this misuse — rename or send an actual session lease.

OrcaWin:
Review round 4 caught a blocker in `c3b899451a` before it reached a device: production `resumeTtlMs` is **30 days**, and `30d − 30s` overflows setTimeout's int32 limit — Node clamps the timer to **1ms**, turning the rotation fix into a ~1000 dials/sec loop (measured: 3001 dials + 3001 SecureStore writes in 3 simulated seconds). The churn test had passed only because its fake used 7 days, which fits in int32.

Fixed in `94441e7827`:
- `scheduleFromLease` clamps the delay to **[60s, 6h]** — the ceiling makes overflow unreachable at any server TTL (cost: a harmless re-resume every 6h on long sessions); the floor bounds any bad deadline (including `return-unchanged-grace` resumes whose stored expiry can be arbitrarily near) to one forced rotation per minute instead of a sub-second loop.
- `getLeaseExpiresAt` → `getAttachDeadlineAt` (zero production callers remained; the plausible-sounding name is exactly how the churn bug happened).
- Expired vs missing credential bundles now log distinct strings for field triage.
- Both test fakes use **production constants** (10s attach deadline, 30-day resume TTL). Fictional fake values hid all three transport defects in this PR; the churn regression is red on the pre-clamp timer at production constants.

Also noted from review (pre-existing, non-blocking): a relay-only phone that sleeps past the 30-day resume expiry degrades gracefully to the slow reprobe ladder and recovers on LAN return or re-pair; the 24h grace window cannot bridge it.

OrcaWin:
Final review verification at production constants against `94441e7827` (independent harness): 30-day TTL over 5 simulated minutes = 1 dial/1 credential write (was 3001 dials in 3s); over 7 simulated hours = exactly the initial connect plus one 6h-ceiling re-resume; near-expiry unrenewed deadline = one rotation per minute at the floor (was a 1s loop). Build 5 cleared for device verification.

One residual for a follow-up issue (not this PR): when the cell answers `return-unchanged-grace` (`renewed: false`), the returned deadline never advances, so a phone stuck on a stale grace credential steady-states at ~60 session replacements/hour (an RPC failure roughly once a minute) until it adopts the newer current credential over direct or from a fresh bundle read. Precise remedy: skip proactive rotation entirely when `confirmation?.renewed === false` — a re-resume with the same grace credential provably returns the same decision, so rotation can accomplish nothing there.

PR #12377, opened 2026-08-04
perf(agent-status): coalesce live status bursts into one render pass (STA-3328)

brennanb2025:
## Summary

Third installment of the STA-3328 typing-lag work (after #12359 spinner, #12371 tab-group amplifiers). This addresses the burst **source**: every live agent-status event arrives as its own IPC task, and each one committed the store and paid a full React render pass — profiled live as 200–488ms main-thread tasks (`Receive mojo message → applyAgentStatus → setAgentStatus → renderRootSync`) under heavy orchestration, exactly matching Neil's "happens when using orchestration heavily" report.

The design-defining fact (verified empirically with a real-store/real-React probe before writing any code): **N zustand commits in the same macrotask produce exactly 1 React render pass** — React 19 flushes external-store sync work at the microtask boundary. So no `setAgentStatus` refactor is needed; the fix is dispatch-level in `useIpcEvents`:

- The **leading** event of a burst applies synchronously — single events pay **zero** added latency.
- Followers within a 33ms window buffer and apply together **in one task** — one render pass instead of N. Ordering is unchanged (FIFO; every guard, fold, and side-effect runs per event exactly as before, each seeing the previous event's state).
- Only an **applied** event arms the window: dropped/pending events (startup replay, not-yet-mounted panes) commit nothing and cost no render, so their successors stay immediate — this also preserves the existing buffered-until-mount retry semantics byte-for-byte.
- Both clear paths purge the pane's/connection's queued sets first — without this, a coalesced `working` flushing 33ms after a `clear` would resurrect a removed status (pinned by a dedicated test).
- Effect teardown clears the timer and queue.

Worst-case added status-visibility latency is 33ms — under half the spinner's own 83ms tick.

## Screenshots

No visual change.

## Testing

- [x] `pnpm lint` — oxlint, react-doctor changed-lines gate, oxfmt clean (pre-commit hook re-verified all three)
- [x] `pnpm typecheck` — all three configs pass, 0 errors
- [x] `pnpm test` — full `useIpcEvents.test.ts` suite: 98/98 pass (96 existing + 2 new); full suite left to CI
- [ ] `pnpm build` — left to CI
- [x] New tests verified **non-vacuous** (each fails against the pre-fix source): (1) burst of 3 → leading applies synchronously, followers coalesce into one deferred batch in arrival order; (2) a clear purges queued burst events so a coalesced set cannot resurrect the removed status. One pre-existing test ("buffers ready push events until a mounted tab contains the pane leaf") drove the applied-only-arms-the-window refinement and passes unchanged.

## AI Review Report

Reviewed for: ordering preservation (FIFO queue; leading precedes queued; per-event guards like the `receivedAt` out-of-order rejection see the store state left by the previous event, identical to today's sequential arrival); clear-vs-set races (both the transient watermark path and the direct paneKey clear purge the burst queue using the same raw-paneKey/connectionId comparisons as the existing pending-retry purges); re-entrancy (queue is spliced before applying, mirroring the crash-9fc89529 guard pattern; the existing `isFlushingAgentStatuses` guard is untouched); timer lifecycle (cleared on effect teardown alongside the retry timer; queue emptied); occluded-window behavior (Chromium clamps the 33ms timer to 1s when hidden — status still applies, and a hidden window renders nothing anyway). Cross-platform: renderer-only dispatch change; the SSH/WSL connection-watermark semantics are exercised by the existing suite (98/98). The startup snapshot replay loop was already a single task (batches naturally) and is untouched.

## Security Audit

No input handling, command execution, path handling, auth, secrets, or IPC surface changes. Same events, same validation pipeline, same store actions — only dispatch timing is batched, with a bounded in-memory queue (drained every 33ms, purged on clears and teardown).

## Notes

- Combined effect with #12371: this PR reduces render-pass *count* per burst; #12371 reduces the *cost* of each pass. After both land, re-running the live keystroke-latency rig under agent load will show what's left of the 200–488ms tasks.
- Remaining STA-3328 queue: split-divider drag off the store (drag-only jank), `setTabLayout` structural bailout, audit medium tier.

### DIFF
--- a/src/renderer/src/hooks/useIpcEvents.test.ts
+++ b/src/renderer/src/hooks/useIpcEvents.test.ts
@@ -6236,17 +6236,34 @@ describe('useIpcEvents agent status snapshot integration', () => {
     }
   })
 
-  it('drops queued burst events for a pane its clear removed, preventing resurrection', async () => {
+  it('preserves queued set-clear order for working removal and done retention', async () => {
     vi.useFakeTimers()
-    const setAgentStatus = vi.fn()
-    const removeAgentStatus = vi.fn()
+    let storeState: StoreLike
+    const setAgentStatus = vi.fn(
+      (
+        paneKey: string,
+        payload: { state: string },
+        _title: unknown,
+        timing: { updatedAt: number }
+      ) => {
+        storeState.agentStatusByPaneKey = {
+          ...(storeState.agentStatusByPaneKey as Record<string, unknown>),
+          [paneKey]: { ...payload, updatedAt: timing.updatedAt }
+        }
+      }
+    )
+    const removeAgentStatus = vi.fn((paneKey: string) => {
+      const next = { ...(storeState.agentStatusByPaneKey as Record<string, unknown>) }
+      delete next[paneKey]
+      storeState.agentStatusByPaneKey = next
+    })
     const onSetListenerRef: { current: ((data: AgentStatusSetData) => void) | null } = {
       current: null
     }
     const onClearListenerRef: { current: ((data: { paneKey: string }) => void) | null } = {
       current: null
     }
-    const storeState: StoreLike = buildStoreState({
+    storeState = buildStoreState({
       setAgentStatus,
       removeAgentStatus,
       workspaceSessionReady: true,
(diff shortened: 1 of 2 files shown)


brennanb2025:
## Electron QA — agent-status burst + terminal typing

Validated the intended dev instance for `brennan/sta-3328-agent-status-coalescing` through CDP/playwright. The real local agent-hook HTTP surface delivered 90 status updates while Playwright typed 24 characters into the actual terminal: 90/90 returned HTTP 204, p95/max inter-key gaps were 32.9/35.0 ms (idle: 27.6/27.9 ms), max frame interval was 9.2 ms, and Chromium reported no long animation frame during the burst.

Visible state matched the backing Zustand state: the worktree and terminal showed `Working` during the burst, then `Done` with the final completion message in FIFO order. The deterministic 98/98 hook suite separately covers the exact queued `done` → pane-clear race fixed in `c97559a604`.

### Working during status burst, with terminal input responsive

Working status burst and terminal typing

### Completion applied in FIFO order

Done status after FIFO completion

Not exercised live: hidden-window timer clamping and an SSH/relay connection; those paths were reviewed and covered by the focused suite, but this Electron run used a local workspace.

PR #12380, opened 2026-08-04
feat(sidebar): link Linear issues from Edit Worktree Details

AmethystLiang:
## Problem

The Edit Worktree Details dialog has a single **GH Issue** field. A workspace whose work is tracked in Linear had no way to say so from the UI — the link could only be set when the workspace was created, or out-of-band via `orca worktree set --linear-issue`. The data model already had the slots; the editor didn't.

## What changed

One provider-aware **Issue** row replaces the GitHub-only field. No extra rows, no extra dialog height:

- A chip suffix inside the input selects **GitHub** or **Linear**.
- Pasting a URL auto-detects the provider and flips the chip.
- A **bare key never steers the provider.** `LINEAR_IDENTIFIER_PATTERN` and Jira's issue-key pattern are byte-identical, so shape alone cannot decide one — letting it try would silently override an explicit choice.
- A GitHub *pull* URL is not decisive either: it's a valid GitHub link but not an issue, and must not flip the chip into a state where the field then refuses to save it.

### One issue per workspace

A changed field displaces the other provider's slot, and the helper line names what Save is about to unlink (both, when clearing the field drops both). Only a *dirty* field displaces anything — the dialog opens focused on Comment, and an untouched row leaves every link alone.

**GitLab and Jira links are deliberately left alone.** The row cannot render them, `displacedLinkLabels` has no branch to name one, and there is no CLI flag or card property to restore one from. Displacing what it cannot display would destroy a link the user was never shown.

## Correctness notes

- **Present-but-`undefined` keys erase fields.** `setWorktreeMeta` raw-spreads `{...existing, ...meta}`, so a key that must not be written has to be structurally *absent*, not `undefined`. Every build in `worktree-meta-updates.test.ts` asserts this invariant.
- **Frozen baseline.** Dirtiness compares against a snapshot captured in the open transition, not live store state. A background `orca worktree set --linear-issue` while the dialog is open would otherwise make the untouched field read as dirty, and the next comment-only save would write the stale seed back over the new link.
- **Remote runtimes are gated on key *presence*, not value.** `worktree.set` parses in strip mode, so an older runtime drops an unknown key and applies the rest of the payload. A dropped *clear* is as damaging as a dropped write: the stale link it strands is one the Issue row then hides while the card still badges it. Two capabilities are asserted, because `linkedWorkItem`/`linkedTaskSourceContext` entered the schema six weeks after `task-source-context.v1`.
- **Save failures are now visible.** `updateWorktreeMeta` returns a result instead of swallowing. A failed save refetches, which reverts the optimistic write — closing the dialog there reported success for an edit that undid itself, and discarded the name, comment and PR changes in the same payload. The dialog now stays open with a `role="alert"` line. (Result-return rather than rethrow: ~20 `void updateWorktreeMeta(...)` call sites would otherwise produce unhandled rejections.)
- **Parses are length-bounded.** `matchGitHubItemPath` strips trailing slashes with an unanchored `/\/+$/`, which is quadratic — a 160 KB paste of slashes froze the renderer for ~19s. These parses run on every keystroke, so the field, the hook, and the PR gate all take the same bound; the check short-circuits on length, so it stays O(1) on a huge paste.
- **Stale org key.** The workspace's stored Linear org key describes the *persisted* issue, so it only applies while the typed identifier still is that issue. Reusing it for a newly typed one built a URL in the wrong org and short-circuited the lookup that would have reported the miss.
- **Folder workspaces are read-only here** — their link is creation-time only, and `updateWorktreeMeta` drops link keys for them entirely. The row still *shows* the link (they live outside `worktreesByRepo`, so the indexed lookup alone left it blank).
- **Protocol gate on Linear URLs.** A hostname-only check accepts `file://linear.app/...`; the parsed value is later handed to `shell.openUrl`.

Also fixed while here: a bare identifier now clears `linkedLinearIssueWorkspaceId`/`OrganizationUrlKey` instead of inheriting them. A stale org key pins `resolveLegacyLinearLinkWorkspace` to the wrong org — not-found, or a silently wrong issue on a team-prefix collision.

## Performance

The dialog's worktree lookup goes through the existing owner index. The code it replaced was `Object.values().flat().find()` — an N-element allocation on every store notification, not a plain `Array.find`.

## Testing

- `npm run typecheck` — clean
- `npx oxlint <changed files>` — clean
- `npx oxfmt --check` — clean
- `npx vitest run --config config/vitest.config.ts src/shared src/renderer/src/components/sidebar src/renderer/src/store src/cli` — **811 files / 8592 tests passed**
- All three localization gates pass

New coverage: provider detection and parse (`issue-link-input.test.ts`), payload shape and the erasure invariant (`worktree-meta-updates.test.ts`), dialog behavior incl. seeding, chip flips, displacement notices, the frozen baseline, folder workspaces, and the save-failure surface (`WorktreeMetaDialog.test.tsx`), and the Linear protocol gate (`linear-links.test.ts`).

Two rounds of `/readiness-checklist` were run with sub-agents across functional, performance, security, back-compat and cross-platform seats; round 2 caught three defects introduced by round 1's own fixes.

## Not in scope

Jira. The chip menu lists only GitHub and Linear — no disabled "coming soon" item. `ISSUE_LINK_PROVIDERS` narrows the canonical provider union rather than minting a parallel one, so adding Jira later is a one-entry change.


AmethystLiang:
@coderabbitai[bot] fixed

PR #12383, opened 2026-08-04
perf(runtime): memoize mobile snapshots for mounted terminal worktrees

brennanb2025:
## Summary

Follow-up to #12207, found during a release scan of #12207 + #12245: the per-worktree mobile-snapshot rebuild memo refused to skip any worktree with a mounted TerminalPane, because the builders read live PaneManager/DOM state the memo could not witness. Hidden worktrees stay mounted (8 hot-parked + 12 retained by default; unbounded with `terminalHiddenViewParking` off), so on the always-on hosts the memo targeted, ~20 worktrees still paid a full content build on every publication — plus the new inputs build and deep compare #12207 added.

This PR captures the live reads instead of bypassing the memo. `MountedTerminalSurfaceCapture` snapshots, once per publication per mounted tab: the live pane leaf ids, the live active leaf, the serialized pane DOM tree, and each pane's numeric-id/pty bindings. The builders now read only the capture — `registeredTabs` is no longer reachable from the build path — so live state outside the capture is unrepresentable, preserving the compiler-proof input-set property #12207 established. The memo compares captures by value, so a mounted worktree whose panes did not change reuses its snapshot like any other, and any live change (split, pty rebinding, active-pane move, unmount) still forces a rebuild.

No behavior change to snapshot content: the capture holds exactly the values the builders previously read live, resolved at the same point in the same synchronous build pass.

## Screenshots

No visual change.

## Testing

- [x] `pnpm exec oxlint` on both changed files (plus pre-commit oxlint + React Doctor + oxfmt)
- [ ] `pnpm test` — the wrapper does not run in this environment (`windows-native-registry` rebuild); vitest was invoked directly:
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/renderer/src/runtime/` — 58 files / 569 tests pass
- [x] All out-of-directory importers of `sync-runtime-graph` (`terminal-pane/`, `hooks/`, `store/slices/`, `workspace-activation-terminal-focus`) — 370 files / 5317 tests pass
- [ ] `pnpm build`
- [x] Added 4 tests in `sync-runtime-graph-publication-cost.test.ts`; the skip test (`reuses a mounted worktree snapshot when its live pane state is unchanged`) fails against the pre-fix source (verified via stash), and the three live-change tests pin the safety direction

### New tests

- reuses a mounted worktree snapshot when its live pane state is unchanged (zero content-build reads + object identity)
- publishes a live pty rebinding on a mounted worktree (new ptyId reaches the snapshot; unmounted sibling untouched)
- publishes a live pane split on a mounted worktree (second leaf appears)
- falls back to the persisted pty binding after the surface unmounts (rebuild fires, saved `ptyIdsByLeafId` takes over)

These also close a review gap: `hasMountedTerminalSurface` — the guard this PR replaces — previously had no test that would fail if it were dropped.

## AI Review Report

Reviewed as part of a 4-reviewer release scan of #12207 + #12245 (skip-safety, contract, perf, security), which produced this finding: the mounted-set assumption ("mounted ≈ active worktree") was wrong, making #12207 a small net regression for the mounted tail. Risks checked for this fix:

- **Capture completeness**: the build path's only live-state reads were `registeredTabs.get()` in `getRuntimeLeafIdsForTerminal` and `buildMobileTerminalSurfaceTabs` (`getPanes`, `getActivePane`/`getLeafId`, `getContainer` → `serializePaneTree`, `getNumericIdForLeaf`, `getPtyIdForPane`). All six reads are in the capture; after this change neither builder references the registry, so a future live read must be added to the capture type to compile.
- **Staleness (the P0-shaped risk post-#12245)**: a false "unchanged" would be withheld from main permanently. The capture is rebuilt from live sources on every publication and compared by value, so any live change is witnessed; verified by the pty-rebinding and pane-split tests.
- **Semantics preservation**: the active-leaf fallback chain (`hasLiveActivePane` distinguishes live-null from no-live-pane), the pty resolution chain (live → saved → single-tab fallback), and the persisted-leaf path when `getPanes()` is empty all mirror the pre-change expressions one-for-one.
- **Cost**: capture is O(panes) per mounted tab (a DOM walk the build already did every publication) and is only allocated for worktrees with registered tabs; the 200+ unmounted tail shares the empty-map singleton and compares by reference.
- **Cross-platform (macOS/Linux/Windows)**: renderer-only TypeScript; no paths, shortcuts, shell, or Electron platform APIs touched. The `typeof HTMLElement !== 'undefined'` guard moved intact, so node-side test environments are unaffected. SSH/remote and folder-workspace paths flow through the same store slices and are unaffected.

## Security Audit

No new IPC surface, command execution, path handling, or input parsing. The change is renderer-local memoization: it reads existing in-process PaneManager/DOM state and compares it by value. The IPC payload shape (`RuntimeSyncWindowGraph`) is untouched; main-process code is untouched. No dependencies added. Captured pty ids were already published in snapshots pre-change; the capture holds them at most one publication generation.

## Notes

- The capture also removes the mounted-tail deep-compare/backstop churn flagged in the scan; the remaining P2s from that scan (comparator exhaustiveness enforcement, unmemoized agent-status index, two-generation retention in the backstop) are intentionally not bundled here.
- If `serializePaneTree` output ever jitters (e.g. float pane sizes on window resize), the capture compares unequal and rebuilds — identical behavior to before this PR, never a suppressed publication.

### DIFF
--- a/src/renderer/src/runtime/sync-runtime-graph-publication-cost.test.ts
+++ b/src/renderer/src/runtime/sync-runtime-graph-publication-cost.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { buildMobileSessionTabSnapshots } from './sync-runtime-graph'
+import { buildMobileSessionTabSnapshots, registerRuntimeTerminalTab } from './sync-runtime-graph'
 import type { AppState } from '../store/types'
 
 // Why: getBrowserTabsByWorktree reads this slice once per worktree inside the build loop,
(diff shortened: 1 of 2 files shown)


brennanb2025:
## QA evidence

Validated the branch build at `454bf3bf` using the isolated Electron instance `trumpetfish @ brennanb2025/mounted-surface-capture-memo`.

- Electron: activated the PR worktree, mounted `Terminal 1`, and split it through the real terminal control. The visible app rendered both panes; backing state held two distinct leaf IDs and PTY bindings.
- iOS Simulator: iPhone 17 Pro (iOS 26.5), Orca Mobile dev client, Metro LAN bundle, paired to this same branch runtime. The worktree session rendered `3 tabs`: the two `Terminal 1` split leaves plus `Mobile QA`, and opened live terminal output.
- Automated verification: focused runtime suites `74/74`, full runtime suite `569/569`, `typecheck:web`, changed-file oxlint, and `git diff --check` all passed.

### Electron — mounted split surface

Electron mounted split

### iOS — paired mobile session receives both mounted split leaves

iOS paired mobile session

PR #12393, opened 2026-08-04
fix(pty): require proven absence before retiring a stable pane owner

OrcaWin:
## The defect

`LocalPtyProvider.spawn({ attachOnly })` throws the literal string `Session not found:` for an id it does not own. `isPtyAlreadyGoneError()` regex-matches that string, and `attachStablePaneOwner()` treated it as proof the PTY had died — firing a synthetic exit, deleting the persisted pane→PTY binding, and spawning a duplicate.

In degraded mode any unmapped session id is routed to that local fallback, so a **live daemon-hosted agent got declared dead by a provider that never owned it**. For a single-pane tab the retirement also drops the tab from persisted state, which the renderer cannot restore.

## The fix

Retire the pane only when a provider **authoritatively answers absent** — `probePtyLiveness(...) === false`. `true` (someone owns it and it is alive) and `null` (nobody could answer) do **not** authorize destroying the binding. Providers that do not implement `probePtyLiveness` are their own sole owner, so their refusal stays authoritative and their behavior is unchanged.

That is 9 lines in `attachStablePaneOwner`.

### Closing the one gap the strict gate opens

A one-shot gate means an adapter that can *never* answer makes the probe return `null` forever, so a genuinely dead pane would surface an error instead of respawning — a permanent wedge.

There is one such adapter, and it is pre-existing code (`getSize` + its catch-to-`null` were introduced in `24a2accc3c`, not by this PR):

- `getSize` landed in daemon protocol **v18** (`22b00a7cd2`, added to `daemon-server.ts` without a version bump).
- Legacy adapters are constructed for **every** version in `PREVIOUS_DAEMON_PROTOCOL_VERSIONS` (1..30) whose socket still accepts a connection (`daemon-init.ts:1149`).
- A pre-v18 daemon falls off `routeRequest`'s switch and error-replies `Unknown request type: getSize` (`daemon-server.ts:1078`); the client rejects, and `probePtyLiveness` catches it as `null`.
- One `null` makes `probePtyOwners` return `null` (`[false, null]` → `null`), so the fan-out is unprovable **forever** — `protocolVersion` is fixed at construction and never renegotiated.

`probePtyLiveness` now version-gates `getSize` and asks such a daemon for `listSessions` instead — present since the first daemon commit (`fd4f986c59`), and the same inventory legacy discovery already routes by. It is requested **directly** rather than through `listProcesses`, because `listProcesses` swallows errors into an empty list, which would read as absence.

This never lets "unreachable" mean "absent": if `listSessions` also fails, the answer is still `null`. Only an owner that actually answered can authorize retirement. This mirrors the existing pre-v27 `inspectProcess` fallback in the same file, which uses the identical "request directly so a dead socket still rejects" reasoning.

## Scope

| File | +/- |
|---|---|
| `src/main/ipc/pty.ts` | +9 |
| `src/main/daemon/daemon-pty-adapter.ts` | +13 |
| `src/main/daemon/daemon-protocol-version.ts` | +2 |
| `src/main/ipc/pty.test.ts` | +125 |
| `src/main/daemon/daemon-pty-adapter.test.ts` | +53 |

Purely additive: **+202, −0** against the merge base.

## Dropped follow-up

The second commit (`aa27eb5274`) was reverted — too much surface for a P0, and two pieces did not hold up under review:

- The `hasPty` owner scan added to `DaemonPtyRouter.spawn` and the degraded fresh-spawn router is **inert** for the case it targeted. `hasPty` reads `activeSessionIds`, which only spawn/attach populate — `listProcesses` never does. After a failed discovery it cannot find the owner it was written to find.
- Its two routing tests **could not fail**: the mocks backed `hasPty` and `listProcesses` with one shared array, giving the fake adapter knowledge the real `DaemonPtyAdapter` provably lacks.

Also dropped: the bounded 6s retry loop and its constants, the sentinel symbol, the exported user-facing message constant, the `opts.deadlineMs` widening of `probePtyLiveness` and its threading through the probe and both routers, and the `ensureConnected` change.

**Kept** from the follow-up: the `flushOrThrow` revert. `flushPendingOrThrowAsync` is not a like-for-like twin — it defaults to the drain-to-stable-generation loop, adds active-view and GitHub sidecar writes, re-serializes state per iteration, and rejects when writes are frozen where the sync path no-ops. `retirePersistedStablePaneOwner` stays synchronous.

## Tests

- `keeps a persisted owner whose absence is unproven` (×2: probe says alive / nobody could answer) — proves the pane is not retired, gets no synthetic exit, is not duplicated, and the tab survives.
- `answers a pre-getSize daemon from its session inventory` — with only the source reverted this fails `expected null to be true`, i.e. it reproduces the wedge.
- `still answers unknown when a pre-getSize daemon cannot be reached` — pins the P0 boundary: unreachable stays `null`.

Every test fails with only the source reverted (verified via patch, not stash).

```
npx vitest run --config config/vitest.config.ts src/main/ipc/pty.test.ts src/main/daemon
  → 86 files passed | 1674 tests passed
npx tsc --noEmit -p config/tsconfig.node.json        → clean
node config/scripts/check-changed-code-quality.mjs   → 0 new findings
npx oxlint <touched files>                           → clean
node config/scripts/check-max-lines-ratchet.mjs      → OK, no new bypasses
```

## Follow-up (not in this PR)

The real root cause is that a provider which does **not own** a session reports it with the same string as a provider that owns it and knows it is dead. Both surface as the literal `Session not found`, and `isPtyAlreadyGoneError()` cannot tell "the owner says it is dead" from "a non-owner says it has never heard of it".

The structural fix is for a non-owning provider to raise a **distinct typed "not mine" error** instead of `Session not found`, so `isPtyAlreadyGoneError` stops conflating the two. That would make liveness probing unnecessary here entirely — this PR's guard, the fan-out, and the protocol gate could all be removed — and it would fix every other caller that pattern-matches that string.


OrcaWin:
## Review findings addressed (all four)

**#1 graceful-teardown window (should-fix) — fixed.** Confirmed: `terminal-host-session-create.ts` throws `SessionNotFoundError` while `existing?.isTerminating`, but `getAppliedSize` still answers because `isAlive` is `_state !== 'exited'` and `isTerminating` never flips `_state` — so the probe says alive for up to `KILL_TIMEOUT_MS`. The veto is no longer terminal: `attachStablePaneOwnerUntilProvenGone` re-attaches on a bounded 6s budget (250ms spacing), and the instant any owner answers `false` the pane retires and spawns fresh exactly as on main. Only an owner that stays unprovable for the whole budget fails the spawn.

**#2 unprovable owners (should-fix) — fixed at the root.** `DaemonPtyAdapter.probePtyLiveness` never called `ensureConnected` (unlike `listProcesses`), so a merely disconnected adapter answered `null` forever and one `null` poisons `probePtyOwners`; it also took no deadline, so a wedged daemon burned the client's 30s `REQUEST_TIMEOUT_MS` per attach. It now connects first and takes an absolute `deadlineMs` threaded through `probePtyOwners` → both routers, so the whole unproven wait shares one budget. The thrown error is user-legible English instead of the raw `terminal_pane_owner_unverified` token that reached the toast verbatim.

I did **not** convert "unreachable" into "absent": that would re-open the P0 this PR exists to fix. What changed is that unknown is now rare (reconnect), bounded (deadline), and self-healing (each retry and each later attach re-proves), rather than a permanent one-shot veto.

**#3 durability barrier (should-fix) — reverted, reviewer is right.** `flushPendingOrThrowAsync` routes to `flushCurrentStateAsync` with `drainToStableGeneration` defaulting to `true` — the exact `for(;;)` livelock the sibling `flushPendingAsync` avoids by passing `false` — plus active-view + GitHub-cache sidecar writes, a `buildStateToSave()` per iteration, and a reject on `writesFrozen` where `writeToDiskSync` silently no-ops. Not a twin, and unrelated to the correctness fix. `retirePersistedStablePaneOwner` is synchronous again and calls `flushOrThrow()`, matching the sibling retire paths in `orca-runtime.ts`.

**#4 misrouting root cause (nit) — fixed.** `DegradedDaemonFreshSpawnRouter.spawn` sent an unmapped `sessionId` to the local fallback and `DaemonPtyRouter.spawn` sent it to the current daemon — both fake "Session not found" for a live session. Both now resolve an existing owner first (the same `findProviderForExistingSession` / `hasPty` resolution every non-spawn path already used); minted ids still route to the fallback / current daemon. The pane now **reattaches** instead of merely surviving.

## Verification

New/changed tests, each proven to fail with only the source reverted (`git diff` + `git checkout --`, no stash):
- `pty.test.ts`: retry-until-proven (alive and null), teardown-window retire, misroute-then-reattach, `flushOrThrow` durability.
- `daemon-pty-router.test.ts`: shared probe deadline, undiscovered-session owner routing.
- `degraded-daemon-pty-provider.test.ts`: undiscovered-session owner routing.
- `daemon-pty-adapter.test.ts`: unconnected adapter answers authoritatively, probe request bounded by caller deadline.

Two provider mocks now drop the session id on `emitExit`, mirroring `DaemonPtyAdapter`'s real `activeSessionIds.delete` on the exit event — without that they claimed `hasPty` for exited sessions.

`src/main` 18,840 passed / 1 pre-existing failure (`updater.test.ts` "recovers quit-for-update state on sync quitAndInstall error event", fails identically with every file of this PR reverted). `tsc -p config/tsconfig.node.json`, `tsconfig.cli.json` clean; `tsconfig.web.json` only pre-existing TS6307. `oxlint` + `oxfmt --check` clean; max-lines ratchet OK.

SSH/relay providers still have no `probePtyLiveness`, so their refusal stays authoritative and their behavior is unchanged. No wire or persisted-state change.

OrcaWin:
Follow-up: replaced the protocol-version gate on `getSize` with runtime capability detection.

**Why the version gate was unsound.** `getSize` landed in `22b00a7cd2` while `PROTOCOL_VERSION` was *already* 18, and that commit did not bump it. Verified:

| ref | `PROTOCOL_VERSION` | `getSize` in daemon-server |
|---|---|---|
| `22b00a7cd2^` | 18 | 0 occurrences |
| `22b00a7cd2` | 18 | 1 occurrence |

So version 18 means two different things. A daemon from that build window passes `>= 18`, gets sent `getSize`, error-replies `Unknown request type`, and answers `null` — and one `null` makes the owner fan-out permanently unprovable, so a genuinely dead pane can never retire. That is the same wedge the gate was meant to close, narrowed to one window rather than removed.

**The change.** Attempt `getSize`; fall back to `listSessions` only when the daemon itself says the request type is unknown. The predicate is deliberately narrow (`Unknown request type`, which the server throws and `addNodePtyRecoveryHint` only ever prepends to) so a transient failure stays unproven rather than being mistaken for a missing capability. The rejection is remembered per adapter so later probes skip a round trip that cannot work — relevant on cold restore, where many panes probe at once.

This follows the capability-probe rule `AGENTS.md` already mandates for Git binaries, for the same reason: version numbers overstate capabilities.

**Why no path regresses**
- `getSize` succeeds → unchanged.
- `getSize` rejected as unknown → now answers from `listSessions` instead of `null` (strictly better).
- `getSize` fails any other way → rethrown, still `null` (unchanged).
- Pre-v18 → still goes straight to `listSessions` (unchanged).

Unreachable still never means absent.

**Verification.** `src/main/daemon` + `src/main/ipc/pty.test.ts` → 1676 passed. `tsc -p config/tsconfig.node.json` clean. Changed-code quality gate: 0 new findings across all three checks. oxlint/oxfmt clean. The new fallback test was proven genuine by reverting only the source and watching it fail.

PR #12394, opened 2026-08-04
fix(worktree): tell the truth about live PTYs, and offer force for a wedged sweep

OrcaWin:
Found by the `v1.4.167..v1.4.168-rc.1` production release scan. Three gaps in the #11960 / #12153 force-delete path.

## 1. The delete toast described proven-live terminals as unconfirmed

`describeUnstoppedPtys()` already distinguishes two verdicts — `still live: <ids>` (verification watched them running) and `could not verify these exited: <ids>` (nobody could answer). The toast collapsed both into:

> Orca could not confirm every terminal in this workspace has exited...

Force Delete proceeds in **either** case, so a user waiving "we could not confirm" was in fact killing terminals Orca had just observed running — and any uncommitted work in them. They were making a destructive decision on wrong information.

The live verdict now gets copy that says so, naming the consequence. The unverifiable case keeps the existing wording.

Note the existing test had encoded the bug: its fixture was the `still live:` variant while it asserted the "could not confirm" copy. It is now split into the two real cases.

## 2. A wedged sweep offered no Force Delete at all

When a sweep fails before any per-PTY verdict exists, `classifyWorktreeForceDeleteReason()` did not recognise the message, so it returned `null`, so the UI rendered no Force Delete button. That is precisely the permanent wedge #11960 set out to remove: the workspace could not be deleted and the escape hatch was invisible.

There are two distinct shapes, and both are now covered:

- **The outer deadline expires.** `Timed out waiting for physical PTY teardown: <id>` now carries the shared force hint and classifies as `unstopped-pty`.
- **The provider rejects the sweep** — dropped SSH channel, erroring daemon. `settleBeforeDeadline` rejects with the provider's *original* error (it only substitutes the deadline error on the timer path), so the message carried no marker at all. On the destructive path that rejection is now reworded through the existing `Failed to physically stop every PTY for worktree:` prefix, keeping the provider's own text in the message and the original error as `cause`. Reusing the existing prefix means clients that predate this change classify it too.

## 3. Force could delete files while a sweep was still running

`settleBeforeDeadline` rejects on its timer **without cancelling `run()`**, so `Promise.allSettled` resolved at the deadline with a `shutdown()` still in flight — by construction, the deadline error can only fire while something is still working. The forced path then returned, and the caller went on to delete the worktree directory that a not-yet-dead PTY still held open (`EBUSY` / half-deleted workspace on Windows and WSL) — the exact ordering the removal call site's comment exists to guarantee.

Widening the classifier in §2 makes that path reachable more often, so it is fixed here rather than left: sweeps are tracked, and the forced path waits for the abandoned work bounded by a 2s grace. Force still never wedges — and when the grace expires with work in flight, the warning now says PTY handles may outlive the delete instead of implying the sweep finished.

## Verification

New tests: the sweep-rejection classification (with the provider wording preserved), the wait for a deadline-abandoned `shutdown()`, the grace-expiry warning, and the anchored live-verdict matcher. The toast tests now derive `forceDeleteReason` from `classifyWorktreeForceDeleteReason` the way the store does, instead of passing a literal in. Suites: `src/main/runtime` 3,508 passing; `src/shared` + `src/renderer/src/components/sidebar` + `src/renderer/src/i18n` 5,651 passing; `src/main/ipc/worktrees*` + worktrees store slice 501 passing. Both typechecks, oxlint, oxfmt, and all three localization gates clean.

### DIFF
--- a/src/main/runtime/forced-sweep-settlement.ts
+++ b/src/main/runtime/forced-sweep-settlement.ts
@@ -1,0 +1,122 @@
+import { describeError } from './unstopped-pty-verification'
+
+// Why: an abandoned sweep is one whose provider ignored the deadline it was already given,
+// so this is a second chance, not a second budget — long enough for a shutdown() that is
+// mid-kill to finish releasing handles, short enough that Force Delete still feels forced.
+export const ABANDONED_SWEEP_GRACE_MS = 2_000
+
+export type WorktreeSweepTracker = {
+  /** Wraps a sweep body so the promise stays observable after its deadline abandons it. */
+  track: <T>(run: () => Promise<T>) => () => Promise<T>
+  /** Resolves when every tracked sweep settled, or the grace expired; true if one is still running. */
+  awaitAbandonedSweeps: (graceMs?: number) => Promise<boolean>
+}
+
+/**
+ * Keeps the real sweep promises reachable after `settleBeforeDeadline` stops waiting on them.
+ *
+ * Why: that helper rejects on its timer without cancelling `run()`, so by construction the
+ * deadline error can only fire while work is still in flight — and forced removal deletes
+ * the worktree directory the moment teardown returns.
+ */
+export function createWorktreeSweepTracker(): WorktreeSweepTracker {
+  const running = new Set<Promise<unknown>>()
+  return {
+    track:
+      <T>(run: () => Promise<T>) =>
+      () => {
+        const sweep = run()
+        running.add(sweep)
+        // Why: the tracked copy is observed here too, so a rejection this call abandons
+        // never surfaces as an unhandled rejection in whatever ran next.
+        void sweep.catch(() => undefined).finally(() => running.delete(sweep))
+        return sweep
+      },
+    awaitAbandonedSweeps: async (graceMs = ABANDONED_SWEEP_GRACE_MS) => {
+      const pending = [...running]
+      if (pending.length === 0) {
+        return false
+      }
+      let expiry: ReturnType<typeof setTimeout> | undefined
+      const expired = new Promise<true>((resolve) => {
+        expiry = setTimeout(() => resolve(true), graceMs)
+        expiry.unref?.()
+      })
+      const stillRunning = await Promise.race([
+        Promise.allSettled(pending).then(() => false),
+        expired
+      ])
+      clearTimeout(expiry)
+      return stillRunning
+    }
+  }
+}
+
+export type WorktreeSweepPromises = {
+  runtime: Promise<{ stopped: number }>
+  provider: Promise<number>
+  registry: Promise<number>
+}
+
+export type ForcedSweepSettlement = {
+  stopped: { runtimeStopped: number; providerStopped: number; registryStopped: number }
+  /** A sweep failed outright, so no per-PTY verdict below it can be trusted. */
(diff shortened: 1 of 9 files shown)


OrcaWin:
Addressed the review. Both `should-fix` findings were real — I reproduced each before changing anything — and all three nits are fixed.

**#1 (`worktree-teardown.ts:93`) — only the deadline case was classifiable.** Confirmed: a provider whose `listProcesses` throws made `killAllProcessesForWorktree` reject with `"SSH channel closed while listing processes"`, and `classifyWorktreeForceDeleteReason` returned `null` — no Force Delete button, which is the #11960 wedge itself. `settleBeforeDeadline` substitutes `deadlineError` only on the timer path, so the sweep-rejection shape never carried a marker. On the destructive path (`requirePhysicalStop`) that rejection is now reworded through the **existing** `UNSTOPPED_PTY_REMOVAL_PREFIX` — provider text kept in the message, original error kept as `cause`, deadline errors passed through untouched. Reusing the shipped prefix rather than adding a third one means clients that predate this change classify it too.

**#2 (`worktree-teardown.ts:188`) — force raced its own delete.** Confirmed: with `listProcesses` resolving and `shutdown()` hanging past a 30ms budget, `killAllProcessesForWorktree(..., {allowUnverifiedStop:true})` returned while the shutdown promise was still pending — so the caller's `rmdir` ran against a PTY still holding the directory. The `allSettled` comment claimed the opposite. Root cause is that `settleBeforeDeadline` abandons `run()` without cancelling it, so I made the abandoned work observable: new `forced-sweep-settlement.ts` tracks the real sweep promises, and the forced path waits for them, bounded by a 2s grace. Force still never wedges — and when the grace expires with work in flight, the warning now says handles may outlive the delete instead of implying the sweep finished.

**#3** — the toast tests now derive `forceDeleteReason` from `classifyWorktreeForceDeleteReason(error)` the way `store/slices/worktrees.ts` does, instead of passing the literal in. Reverting the four source files to `origin/main` now makes that test fail (it did not before).

**#4** — `unstoppedPtyLive` added to `en.json`, matching the sibling key's precedent.

**#5** — `isProvenLivePtyRemovalError` now anchors `still live:` to the ` — ` detail separator (extracted as a shared constant used by the formatter, so the two cannot drift). A worktree path containing the phrase no longer flips an unverifiable verdict to the destructive copy.

**Proof each test is genuine:** reverting only the source files and re-running produced exactly these failures — `does not let a worktree path spell out a live verdict`, `offers force delete for a sweep-level failure without losing the provider wording`, `waits for a shutdown the deadline abandoned before force returns`, and `lets force through a sweep that never settles before the deadline` (the grace-expiry warning assertion). Against full `origin/main`, `offers force delete when the teardown sweep itself timed out` now fails too.

**Suites:** `src/main/runtime` 3,508 passed; `src/shared` + `src/renderer/src/components/sidebar` + `src/renderer/src/i18n` 5,651 passed; `src/main/ipc/worktrees*` + worktrees store slice 501 passed. `tsconfig.node` and `tsconfig.web` both clean (web has only pre-existing TS6307 noise), oxlint and oxfmt clean, and the localization catalog / extraction / coverage gates all pass. PR description corrected — it had described the sweep-rejection case while only the timeout case was fixed.

PR #12395, opened 2026-08-04
fix(updater): re-prove the retained Linux package before a privileged retry

OrcaWin:
Found by the `v1.4.167..v1.4.168-rc.1` production release scan.

## The defect

The Linux package recovery card (#12183) offers three actions. **Copy Install Command** and **Show Package** both call `validateTrackedArtifact()` and re-hash the retained `.deb`/`.rpm`. **Try Automatic Install Again** — the one that actually runs a package manager as root — called `quitAndInstall()` with no re-verification at all.

The cached path is user-writable (`$XDG_CACHE_HOME`/`~/.cache/orca-updater/pending/`). electron-updater's `DebUpdater` runs `["dpkg","-i",installerPath]` through `runCommandWithSudoIfNeeded`. So the digest proven when the recovery card rendered said nothing about the bytes a root package manager would read minutes later, after a failed escalation left the card on screen.

## Scope, stated honestly

The privileged-install primitive is **not new** — `v1.4.167` already invoked the same electron-updater path, so this is not a regression introduced by 1.4.168. What #12183 added is a long-lived retry affordance that widens the window, plus a copy-command path that did not exist before.

This change **narrows** the window; it does not close it. A same-UID attacker can still race between the hash and `dpkg -i`. Closing it properly needs an immutable handoff — installing from a held descriptor or a root-owned copy — which is a larger change and deliberately out of scope here.

## The fix

Re-hash the artifact immediately before the install and ahead of any destructive quit prep, so a replaced or vanished package aborts with copy telling the user to download again, instead of being installed as root. The check runs only on the recovery retry path, where the user has already had one failure.

## Also

Fixes a macOS-only failure this suite gained from the platform-conditional pre-commit copy: `updater.test.ts` hardcoded the non-Darwin string, so the unit suite was red on any Mac. The sibling block already had the platform guard.

## Verification

New test drives a **real** staged package in a real temp cache through the real validator: install fails, a local process swaps the file, the retry aborts — no PTY kill, no quit, correct error status and lifecycle event. The existing retry test now also runs against a genuinely valid artifact rather than a path that never existed. `src/main/updater.test.ts` 109/109 on macOS; updater/window/preload suites 363 passing.

### DIFF
--- a/src/main/linux-package-update-recovery.test.ts
+++ b/src/main/linux-package-update-recovery.test.ts
@@ -611,6 +611,31 @@ describePosix('validation coalescing', () => {
     expect(hashPasses.count).toBe(2)
   })
 
+  // Why: a verdict handed to a root package manager must cover the bytes as of the click, not the
+  // bytes a Copy click started streaming seconds earlier.
+  it('never reuses an in-flight pass for a pre-install re-proof', async () => {
+    await writePackage('orca.deb')
+    capture()
+    const artifact = recovery.getTrackedLinuxPackageArtifact()
+    const copyPass = recovery.resolveLinuxPackageInstallInstructions(recoveryFor())
+    const installPass = recovery.revalidateLinuxPackageForInstall(artifact!)
+
+    await expect(installPass).resolves.toEqual({ ok: true })
+    await expect(copyPass).resolves.toMatchObject({ ok: true })
+    expect(hashPasses.count).toBe(2)
+  })
+
+  it('lets a later Copy click join the pre-install pass', async () => {
+    await writePackage('orca.deb')
+    capture()
+    const artifact = recovery.getTrackedLinuxPackageArtifact()
+    const installPass = recovery.revalidateLinuxPackageForInstall(artifact!)
+    const copyPass = recovery.resolveLinuxPackageInstallInstructions(recoveryFor())
+
+    await Promise.all([installPass, copyPass])
+    expect(hashPasses.count).toBe(1)
+  })
+
   it('does not reuse an in-flight pass for a different artifact', async () => {
     await writePackage('orca.deb')
     await writePackage('orca-next.deb')
@@ -626,6 +651,39 @@ describePosix('validation coalescing', () => {
     )
     await Promise.all([first, second])
     expect(hashPasses.count).toBe(2)
+  })
+})
+
+describePosix('revalidateLinuxPackageForInstall', () => {
+  it('proves the retained package still matches its release digest', async () => {
+    await writePackage('orca.deb')
+    capture()
+    const artifact = recovery.getTrackedLinuxPackageArtifact()
+    await expect(recovery.revalidateLinuxPackageForInstall(artifact!)).resolves.toEqual({
+      ok: true
+    })
+  })
+
+  it('rejects a package swapped after the download was verified', async () => {
+    await writePackage('orca.deb')
+    capture()
+    const artifact = recovery.getTrackedLinuxPackageArtifact()
+    await writePackage('orca.deb', 'attacker supplied package')
+    await expect(recovery.revalidateLinuxPackageForInstall(artifact!)).resolves.toEqual({
+      ok: false,
+      reason: 'hash-mismatch'
+    })
+  })
+
+  it('reports a package deleted from the cache as missing', async () => {
+    const filePath = await writePackage('orca.deb')
+    capture()
+    const artifact = recovery.getTrackedLinuxPackageArtifact()
+    await fsp.rm(filePath)
+    await expect(recovery.revalidateLinuxPackageForInstall(artifact!)).resolves.toEqual({
+      ok: false,
+      reason: 'missing'
+    })
   })
 })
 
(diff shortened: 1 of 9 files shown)


OrcaWin:
Addressed the review findings in f657a51.

**#1 stale abort clobbers a newer cycle (fixed).** The abort now captures an install-cycle signature before hashing and drops its status if the cycle changed — the same protection `failLinuxPackageRecovery` had. The artifact is still cleared, but only when it is identity-identical to the one that failed, so a download that landed during the hash is not destroyed. Test: `drops an abort verdict once a newer check replaced the card` (Check for Updates lands mid-hash; the `available` card survives and the install is still abandoned).

**#2 read-failed copy and card teardown (fixed).** Both messages are gone; the abort now reuses `LINUX_PACKAGE_RECOVERY_MESSAGES`, so `read-failed` reads "Orca could not read the downloaded package…" and `not-regular` keeps its own copy. Reasons outside `RECOVERY_CLEARING_REASONS` keep `recovery` on the status, so the card and its Copy/Show actions survive a transient I/O blip. It still fails closed — root can read a file we cannot, so `chmod 000` on a swapped package would otherwise be a one-line bypass. Test: `keeps the recovery card usable when the re-proof cannot read the package`.

**#3 primary "Restart to Update" path unprotected (fixed).** The re-proof moved into `performQuitAndInstall`, keyed on `getTrackedLinuxPackageArtifact()` instead of the recovery status, so the long-window primary install is covered too. It is skipped when the retained artifact belongs to a different version than the pending install, and a synchronous artifact check keeps every non-Linux install on its exact previous timing. Tests: `aborts the first install when the downloaded package was swapped`, `installs normally when the retained package still matches its digest`.

**#4 wedge on rejection (fixed).** The re-proof owns a flag cleared in `finally`, a rejection is recorded and fails closed as `read-failed`, and the timer callback is back to plain `void performQuitAndInstall()`. The new flag also joins the `quitAndInstall()` guard so a second click during the hash cannot schedule a parallel install. Tests: `stays installable after a re-proof that rejects outright`, `ignores a second install request while the digest re-proof runs`. (This was live, not latent: the PR as pushed left `updater-linux-package-recovery-actions.test.ts` red with exactly that wedge because the module mock lacked the new export.)

**#5 in-flight reuse (fixed).** `revalidateLinuxPackageForInstall` takes `{ fresh: true }` and never joins a hash that started before the click; a later Copy click can still join it. Tests in `linux-package-update-recovery.test.ts` assert the pass counts both ways.

**#6 retry affordance (fixed).** The retry action gained `pendingLabel`/`isPending`, matching Copy and Show.

Test fixtures: the whole Linux block now stages real packages in a real updater cache, so a path that never existed would abort every install rather than silently pass.

Verified: 115/115 `updater.test.ts`, 311/311 across updater/recovery/card suites, plus full `src/main` (only an unrelated pty-subprocess timeout under parallel load, green in isolation). `tsc` node clean; web clean apart from pre-existing TS6307. Every finding's test was confirmed to fail with only its source change reverted.

OrcaWin:
Round-2 finding #1 (should-fix, updater.ts:1906) — fixed in `ef447a5`.

**Verified the claim first.** Confirmed the stale-cycle branch in `reportLinuxPackageRevalidationFailure` returns before any `sendInstallFailureStatus`, and that `createUpdaterQuitAbortRelay.handleStatus` only calls `abort()` on `status.state === 'error'` — so that branch abandoned the install while leaving the renderer permanently restart-prepared (`intentionalAppRestartInProgress` stuck true → `Terminal.tsx`/`Settings.tsx` skip their dirty-buffer prompts, `shutdownCheckpoint` stays deduped).

**Fix (4 lines of source):** push the abandon from `performQuitAndInstall`'s single `return false` site, so it can't depend on what the reporter decides about the status text, and relay it to the relay's existing `abort()`:

```ts
if (getTrackedLinuxPackageArtifact() && !(await proveRetainedLinuxPackage(pendingVersion))) {
  mainWindowRef?.webContents.send('updater:quitAndInstallAborted')
  return
}
```

Deliberately **not** done: no new retry/timeout/reorder machinery, no relaxing of the cycle guard (the stale status stays withheld — the round-1 clobber fix is untouched), no new abort trigger in the relay's status handling (broadening it to non-error states would spuriously abort a legit install when a background nudge check pushes `checking` mid-quit). `abort()` is already a no-op unless the renderer armed a restart, so the push is inert on every other path.

**Tests** (all fail with only the source reverted, verified by patch-revert):
- `drops an abort verdict once a newer check replaced the card` — now also asserts the abort push (the exact case the reviewer proved emitted nothing).
- `aborts the first install when the downloaded package was swapped` — asserts the push, pinning the structural placement at the abandon site.
- `installs normally when the retained package still matches its digest` — asserts the push is **absent**, so a committed install can't have its restart flag cleared mid-quit.
- `renderer restart wiring` — covers the new channel end-to-end to `relay.abort()`.

**Verification:** 236/236 across the 7 touched suites, 155/155 across adjacent updater/UpdateCard/beforeunload suites, both tsconfigs clean, oxlint + oxfmt clean, `check-changed-code-quality.mjs` passes (0 new findings incl. react-doctor). Original defect still fixed — the full `updater.test.ts` Linux block, including the swapped-package aborts, still passes.

Issue #9795, opened 2026-07-21
[Feature]: Support syntax highlighting and formatting for JSP (.jsp) files


ChoiHyungKuen:
### Problem or use case

I am working on a project that involves JavaServer Pages (`.jsp` files). Currently, Orca does not officially support or properly recognize `.jsp` file formats. As a result, code rendering (syntax highlighting) and basic editor features do not work well when editing JSP files, making it difficult to develop Java web applications seamlessly.

### Proposed solution

- **Syntax Highlighting:** Add syntax highlighting support for `.jsp` files (supporting the combination of HTML, Java snippets, and JSP tags).

PR #12402, opened 2026-08-04
fix(editor): map .jsp/.jspf to the html language id

taehwanis:
## Summary

`.jsp` and `.jspf` files open as plaintext. `EXT_TO_LANGUAGE` in `language-detect.ts` has no entry for either, so `detectLanguage()` falls through to `'plaintext'` and the file renders with no color at all — including the plain HTML that makes up most of a typical JSP page.

JSP is HTML markup with `<% %>` scriptlets and `${}` expressions embedded in it, so Monaco's built-in `html` id already colors the majority of such a file correctly. This is two lookup entries: no new language registration, no grammar, no worker routing.

**This is a stopgap, not the finished feature.** #10858 by @hwantage adds a dedicated JSP grammar that tokenizes scriptlets and EL properly, which an `html` mapping cannot do. I checked where the boundary actually falls by reading the rendered token classes in a running build rather than guessing:

| Region | Under this change |
| --- | --- |
| HTML tags and attributes, including `<c:forEach>` / `<c:if>` | colored (`mtk14` for the tag, attribute values tokenized) |
| `<style>` blocks | colored (`color:` → `mtk4`, value → `mtk5`) |
| `<script>` blocks | colored (`const` → `mtk6`) |
| Java inside `<% %>` | **not colored** — the whole line is a single `mtk1` span |
| `${...}` EL | **not colored** (`mtk1`; the braces only pick up bracket-pair colorization) |

That boundary is structural, not something to tune later: Monaco's html tokenizer has no `<%` rule at all, so `<` becomes a delimiter and the remainder falls into the `[^<]+` text rule. When #10858 lands, these two entries should become `'jsp'` and this change disappears into it. I am opening this only because the full grammar has been open since July 27 and the no-color case is fixable in two lines today.

Related: #9795 (feature request), #10858 (dedicated grammar), #11309 (my own grammar attempt, closed in favor of #10858).

## Screenshots

Same `demo.jsp` in both images.

**Before** — `.jsp` is not in the extension table, so the file falls back to `plaintext` (image reused from #11309, which showed the identical unmapped state):

JSP file rendered as plaintext

**After** — markup, embedded CSS and embedded JS are colored; scriptlet bodies and EL deliberately are not, per the table above:

Same JSP file with markup, CSS and JS colored; scriptlet bodies left plain

## Testing

- [ ] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm build`

`typecheck` passes on all three projects (node, cli, web). The new test covers `.jsp`, `.jspf`, and an uppercase `.JSP` Windows path; `vitest run src/renderer/src/lib/language-detect.test.ts` is 14/14. `oxlint` is clean on both changed files. The unchecked boxes are explained under Notes.

## AI Review Report

Reviewed with an AI agent across three angles — correctness of the mapping, cross-platform behavior, and whether the change claims more than it delivers.

What it flagged and what changed as a result:

- **The PR description originally overstated the visual result.** The reviewer pushed back on describing this as "JSP highlighting", so I verified the actual rendering instead of reasoning from the diff: I read the computed token classes and colors of the rendered lines in a running dev build. That produced the table above and corrected the framing to markup-only. Scriptlet interiors resolve to a single `mtk1` span, confirming Java and EL get no tokenization.
- **The comment was longer than this repo's style allows.** Trimmed to a single line per AGENTS.md ("BE CONCISE. 1 LINE if possible").
- **Ordering within the table.** Placed next to `.html`/`.htm` so the related entries stay adjacent, matching how `.cts`/`.mts` sit beside `.ts`.

Cross-platform compatibility for macOS, Linux and Windows was explicitly checked. Nothing here branches on platform: the change is two lookup entries, and the existing `extname()` helper already handles both `/` and `\` separators while `detectLanguage()` lowercases before lookup — the added test asserts an uppercase `.JSP` Windows path resolves correctly. No keyboard shortcuts, labels, shell behavior, path utilities, Electron platform differences, SSH/remote paths or git-provider behavior are touched.

## Security Audit

No meaningful risk surface. The change adds two string keys to a lookup table that maps a file extension to a Monaco language id. There is no input handling, command execution, path resolution beyond the existing extension parse, IPC, auth, secret material, or dependency change. The `html` language and its worker are already bundled and routed for `.html`/`.htm`, so this introduces no new runtime capability — it only reuses one that already ships.

## Notes

- `pnpm lint`: `oxlint` and the code-quality audits pass. The chain stops later at `verify:skill-bundle-manifest`, which reports stale generated artifacts under `resources/skills/` — a pre-existing state unrelated to this change, and those files are not in this diff.
- `pnpm test`: the full suite on a Windows machine fails a set of pre-existing tests that require a POSIX shell (`spawn /bin/sh ENOENT`); the count is unchanged by this PR. The test file touched here passes.
- `pnpm build` is unchecked because verification here was done against a running `pnpm dev` build of this branch, which is what the screenshot and the token-class readings come from.
- `mobile/src/session/mobile-file-language.ts` keeps a separate extension table for the mobile surface and is intentionally not touched — it feeds a different highlighter, and #11309 left it alone for the same reason.
- This touches the same two files as #10858. If that PR lands first, this one becomes unnecessary and should be closed rather than rebased.


brennanb2025:
Thanks for this — and for reading the rendered token classes instead of reasoning from the diff. Scoping a stopgap by where the boundary actually falls is the right way to do it.

I pushed one small commit to your branch:

- Reworded the `// Why:` comment above the two entries to name the *intent* as well as the reason — that this is a stopgap until a real JSP grammar, and that `<% %>` and `${}` bodies stay plain. Without that, the next person to read the table can't tell whether the partial coverage was deliberate or an oversight, and might "fix" it in either direction.

Two things worth folding into the description, both verified in a running build rather than reasoned about:

- Monaco's own `html` language registration already lists `.jsp` in its `extensions` array. Orca never benefits from it because we always pass an explicit language id, which is exactly why the file falls to plaintext today. That's the strongest argument for this mapping and it isn't in the PR text yet.
- Picking the `html` id brings more than tokenization along with it. `Cmd+/` now comments a line as `<!-- ... -->`, which inside a scriptlet isn't valid Java and doesn't stop it executing server-side — and it used to be a no-op on plaintext. `Format Document` becomes reachable and collapses a multi-line scriptlet onto one line while rewriting the Java spacing. And an "Open Preview to the Side" button appears in the editor header for these files, which opens the raw source rather than a rendered page. All three are user-initiated and reversible, autosave is off by default, and all of them go away when a dedicated `jsp` id lands — but they're the consequences a reviewer won't predict, so they belong in the body.

I looked at whether a different existing id would dodge those and concluded no: `razor` treats `@` as a transition into expression tokenization, so every `<%@ page %>` directive would mis-tokenize, and `handlebars` targets `{{ }}` rather than `${ }`. Both would also put a false language id in front of every future `language === …` check. Your choice is the right one.

The preview button is the one I'd track separately — the gate keys off the language id alone, which is a pre-existing coupling in that subsystem rather than anything you introduced.

PR #12404, opened 2026-08-04
fix(editor): stop Close All / Close Others freezing the renderer on large tab sets

OrcaWin:
## Defect

`closeAllFiles` (the store action behind editor **Close All** / bulk close) is roughly **O(N^3)** in the number of open editor tabs.

Per closing tab it called `getRecentlyClosedTabPosition(s, worktreeId, f.id)` to record the reopen position. That function was itself quadratic: it rebuilt `groupTabEntityIds` and `tabBarGroupEntityIds` by `map`-ing over `group.tabOrder` / `tabBarOrder` and doing a full `unifiedTabs.find(...)` inside each iteration. One O(N^2) derivation, re-run once per closed tab.

It runs inside a synchronous zustand `set()`, so the renderer is blocked for the whole duration — the user sees the app hang, with no spinner and no way to cancel.

## Failure scenario

Open a workspace with a few hundred editor tabs — easy after a long session, a mirrored host republishing its open files over SSH/relay, or opening a wide folder workspace — then hit Close All. Measured against the **real** store (`createEditorTabsStore`, real `openFile`, real unified tabs/groups), not synthetic objects:

| tabs | before | after |
| --- | --- | --- |
| 125 | 30 ms | 7 ms |
| 250 | 156 ms | 14 ms |
| 500 | **1.20 s** | 44 ms |
| 1000 | **13.0 s** | 142 ms |

At 1000 tabs the renderer was frozen for 13 seconds.

## Fix

Hoist the per-tab rescans into a single precomputed index.

- New `createRecentlyClosedTabPositionIndex(state, worktreeId)` builds tab-bar order, unified-tab (by `id` and by `entityId`), and per-group order/membership lookups in **one linear pass**, then answers each position query in O(1).
- `closeAllFiles` builds that index once before the loop and calls `positionIndex.positionFor(f.id)`.
- `getRecentlyClosedTabPosition` now delegates to the index, so every single-tab caller (close file, preview replace, terminal close, browser close) keeps identical behavior.

Behavior is preserved exactly: first-occurrence wins everywhere, matching the `indexOf`/`find` scans it replaces; the `groupOrderMatchesTabBar` guard, the early `undefined` return, and the emitted `{ tabBarIndex, groupId, groupIndex }` shape are unchanged. Reopen order and the recorded tab positions from #12236 are untouched. Group index is scoped per group rather than globally, matching the original `group.tabOrder.indexOf(...)`.

Position derivation moved into its own module (`recently-closed-tab-position.ts`) so `recently-closed-tabs.ts` stays under the oxlint `max-lines` limit **without** a suppression or ratchet bump; existing importers are unaffected via re-export.

## Verification

- New test in `editor.test.ts`: `captures close-all snapshot positions without rescanning tab state per closed tab`. It counts work rather than wall-clock time — unified tabs are wrapped so every read of `entityId` increments a counter — and asserts that 4x the tabs costs at most 8x the scans.
  - **Reverted the source change and watched it fail** (`expected 275240 to be less than or equal to 41680`), then restored it and watched it pass. The instrumented scan count for N=1000 goes from 503,000,500 to 4,000 (exactly 4N).
  - The same test also asserts the recorded snapshot position is still `{ tabBarIndex: 0, groupId, groupIndex: 0 }` — that assertion passes both before and after, pinning behavior preservation.
- `npx vitest run --config config/vitest.config.ts` on `editor.test.ts`, `recently-closed-tabs.test.ts`, `tabs.test.ts`, `browser.test.ts`, `worktrees.test.ts`, `useIpcEvents.test.ts`, `terminal-pane-detach-agent-retention.test.ts`, `agent-status-drop-ipc.test.ts`, `worktree-removal-maps-leak.test.ts`, `useRetainedAgents.test.ts` — 719 passing.
- `npx tsc --noEmit -p config/tsconfig.tc.web.json` — clean.
- `npx oxlint` on the changed files and `node config/scripts/check-max-lines-ratchet.mjs` — clean, no new suppressions.

## Not in scope

A residual O(N^2) remains outside this path: the trailing `closeUnifiedTab` loop rebuilds tab arrays once per tab, and `notifyHostOfMirroredEditorClose` re-`find`s the file in `openFiles`. Together they account for the remaining 142 ms at 1000 tabs. Fixing that means a bulk-close entry point in the tabs slice with its own activation/MRU semantics — a separate change.


OrcaWin:
Resolved the #12399 / #12404 merge conflict by merging `origin/fix/editor-reopen-phantom-tab` into this branch.

**What conflicted:** both PRs append a new test to the tail of the same describe block (`createEditorSlice recently closed editor tabs`) in `src/renderer/src/store/slices/editor.test.ts`. Because the two new `it` bodies were adjacent and shared a closing brace, git saw one conflicting hunk rather than two independent additions. `editor.ts` auto-merged cleanly — the two source changes are far apart (`openFile` id resolution vs. `closeAllFiles` position capture).

**Resolution is purely additive:** both tests kept as sibling `it` blocks. No assertion, helper, or source line altered. Both PRs' test helpers (`ownedEditorFileId`, `withCountedEntityIdReads`) merged without conflict.

**Verified both fixes are still live and independently guarded** by reverting one source hunk at a time and confirming only the matching test goes red:

| reverted hunk | failing test |
|---|---|
| `const id = existing ? existing.id : ...` (#12399) | `reactivates the live tab when a stale reopen id targets an already-open file` |
| `positionIndex.positionFor(f.id)` → per-tab `getRecentlyClosedTabPosition` (#12404) | `captures close-all snapshot positions without rescanning tab state per closed tab` (25920 reads vs. the 13440 ceiling) |

Restored, then:
- `npx vitest run --config config/vitest.config.ts src/renderer/src/store/slices/editor.test.ts` → **193/193 passed**
- plus `recently-closed-tabs.test.ts` and `tabs.test.ts` → **296/296 passed**
- `npx tsc --noEmit -p config/tsconfig.node.json` clean; `config/tsconfig.web.json` clean apart from pre-existing TS6307 project-membership noise
- `npx oxfmt --check` and `npx oxlint` clean on the touched files

GitHub now reports this PR as `MERGEABLE`. #12404 lands cleanly after #12399.

PR #12405, opened 2026-08-04
fix(runtime): keep the listener on loopback for a "This computer only" pairing link

OrcaWin:
## Defect

`mobile:getRuntimePairingUrl` (`src/main/ipc/mobile.ts`) called `rpcServer.ensureNetworkExposure()` for **every** runtime pairing offer, regardless of the address being advertised. Settings → "Share this Orca server" offers three choices, one of which — **"This computer only"** ("A browser or Orca client on this computer") — pairs against `127.0.0.1` (`RUNTIME_PAIRING_LOOPBACK_ADDRESS`). Picking it still ran the widen path, rebinding the WebSocket listener from `127.0.0.1` to `0.0.0.0`.

The widen is one-way for the life of the process (`ensureNetworkExposure` early-returns once `wsBoundHost === '0.0.0.0'`, and nothing ever narrows it back), so the runtime stayed published on every interface after the user picked the option that exists to prevent exactly that.

## Failure scenario

Fresh desktop, no paired device — the listener is correctly on `127.0.0.1` (STA-2370). The user opens Settings → runtime pairing, selects the radio labelled "This computer only", and clicks generate. The listener moves to `0.0.0.0` and stays there for the rest of the process, exposing the runtime RPC/web-client surface to the whole LAN (and to anything else on the host's networks, e.g. a coffee-shop Wi-Fi) even though the user explicitly chose the loopback-only option and the generated link only ever names `127.0.0.1`.

Reproduced on `origin/main` before the fix: driving the real IPC handler against a real `OrcaRuntimeRpcServer` with `address: '127.0.0.1'` left `resolvedHost === '0.0.0.0'`.

## Fix

Gate the widen on the advertised address instead of running it unconditionally:

```ts
if (classifyRemotePairingHostname(ip) !== 'loopback') {
  try { await rpcServer.ensureNetworkExposure() } catch (error) { /* unchanged */ }
}
```

- **LAN / Tailscale / custom-host offers keep widening exactly as today** — STA-2370 is intentional: a client off this machine cannot reach a loopback-only socket, and the widen-failure path still reports `network_exposure_failed` rather than advertising a dead endpoint.
- A loopback endpoint is already served by the existing loopback listener, so it needs no bind change.
- Reuses the existing shared classifier (`src/shared/remote-pairing-address.ts`), so `localhost`, `::1`, `[::1]`, `127.0.0.0/8` and `*.localhost` typed into the "Custom address" field are all treated as loopback too — no new address parsing, no new dependency.
- The mobile QR path (`createMobilePairingOffer`) is untouched: its addresses are always non-internal interfaces, and its `local-only` connection mode means "LAN direct, no relay", which genuinely needs the wide bind.

No renderer, preload, or IPC signature change, so old paired web/mobile clients meeting a new desktop are unaffected.

## Verification

- `npx vitest run --config config/vitest.config.ts src/main/ipc/mobile.test.ts` → 22 passed.
- Reverted just `src/main/ipc/mobile.ts` (patch/checkout, no stash) and re-ran: **4 tests fail** (`expected '0.0.0.0' to be '127.0.0.1'`, plus the three handler cases); re-applied → all pass.
- New integration test starts a real `OrcaRuntimeRpcServer` (`wsPort: 0`), registers the real IPC handlers against it, and asserts the actual transport bind host: stays `127.0.0.1` after a `127.0.0.1` offer, becomes `0.0.0.0` after a `100.64.1.20` offer — so the LAN path is pinned as still widening.
- New handler cases cover `127.0.0.1`, `localhost` and `::1` (offer still minted, `ensureNetworkExposure` never called).
- `npx vitest run --config config/vitest.config.ts src/main/runtime/runtime-rpc.test.ts -t "bind host"` → 11 passed (STA-2370 suite unchanged).
- `npx tsc --noEmit -p config/tsconfig.node.json` → clean. `oxlint` + `oxfmt --check` clean on both files.

### DIFF
--- a/src/main/ipc/mobile.test.ts
+++ b/src/main/ipc/mobile.test.ts
@@ -1,4 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
+import type * as NodeOs from 'node:os'
 
 const { handleMock, networkInterfacesMock } = vi.hoisted(() => ({
   handleMock: vi.fn(),
@@ -17,12 +18,21 @@ vi.mock('qrcode', () => ({
   }
 }))
 
-vi.mock('os', () => ({
+// Why: only the interface enumeration is faked; the real `os` stays available for the integration
+// test below, which needs tmpdir() for a real runtime's user data directory.
+vi.mock('os', async (importOriginal) => ({
+  ...(await importOriginal<typeof NodeOs>()),
   networkInterfaces: networkInterfacesMock
 }))
 
+import { mkdtempSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
 import { registerMobileHandlers } from './mobile'
 import { NETWORK_EXPOSURE_FAILED_GUIDANCE } from '../runtime/network-exposure-guidance'
+import { OrcaRuntimeService } from '../runtime/orca-runtime'
+import { OrcaRuntimeRpcServer } from '../runtime/runtime-rpc'
+import { WebSocketTransport } from '../runtime/rpc/ws-transport'
 
 describe('registerMobileHandlers', () => {
   const handlers = new Map<string, (...args: unknown[]) => unknown>()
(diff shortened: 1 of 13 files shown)


OrcaWin:
## Review findings addressed (#1, #2 fixed; #3 corrected)

### #1 — the guarantee died at the next launch (should-fix, confirmed)
Reproduced exactly as reported: after a `127.0.0.1` offer the in-process bind stayed loopback, but `updateLastSeen(deviceId)` (what `mobile-socket-wiring.ts` does for **every** authenticated socket, local browser included) + relaunch on the same `userDataPath` printed `0.0.0.0`.

**Root cause fix:** grants now record the reach they were minted for (`DeviceEntry.pairingReach`, persisted), and `resolveInitialWebSocketBindHost()` only counts a connected device whose grant was minted for network reach. A "This computer only" grant's client is a browser on this machine, so its connection can no longer republish the runtime on every interface.
- **Backwards compatible:** registries written before the field normalize to `network` on load, so an already-paired phone still finds a wide listener after upgrading (covered by a test that writes a legacy registry entry with no `pairingReach`).
- **Never narrows:** a pending grant re-advertised at network reach is upgraded in place, so a LAN link issued over a previously local pending token still works after a relaunch.

### #2 — widen gated on address shape, not intent (should-fix, confirmed)
Confirmed the classifier output on raw override strings: `127.0.0.1:8443 -> loopback`, but `localhost:8443`, `[::1]:6768`, `ws://127.0.0.1:6768 -> public`.

**Root cause fix:** the renderer now sends the declared reach (`this-computer` | `network`, derived from the existing intent) and main gates the widen on that. A Custom `127.0.0.1:8443` fronting `ssh -L 8443:<desktop-lan-ip>:6768` widens again, exactly as on main. The address is used only as a mismatch guard — a `this-computer` reach carrying an off-host address still widens rather than minting an unreachable link — and it is resolved through a new `resolveAdvertisedPairingHostname()` so every accepted form (`host`, `host:port`, `[v6]:port`, `ws(s)://`) classifies identically. An undeclared caller (e.g. the E2E helper) keeps main's always-widen behavior.

### #3 — stale invariant comment (nit)
Verified: `ws-transport.stop()` terminates every `wss.clients` entry, and the widen is no longer confined to the first pairing action. The comment above `ensureNetworkExposure` now says so. The rebind cannot avoid dropping sockets (ws cannot move a listener), so the contract that matters is port reuse — added a test that connects a live `ws://127.0.0.1:<port>` client, widens, and asserts the port is unchanged and a client can reconnect on the already-issued endpoint.

### Verification
- Reverting only the source (`mobile.ts`, `runtime-rpc.ts`, `device-registry.ts`, both renderer files) fails **18** of the new/updated tests; restored, all pass. The monotonic reach upgrade was proved separately by removing just that branch (`expected '127.0.0.1' to be '0.0.0.0'`).
- `mobile.test.ts` 33, `runtime-rpc.test.ts` 89, `pairing-endpoint.test.ts` 56, `RuntimePairingUrlGenerator.test.tsx` 4. Full `src/main/runtime` 3528 pass, `src/main/ipc` + `src/preload` + `src/shared` 6449 pass, `src/renderer/.../settings` 930 pass.
- `tsc -p config/tsconfig.node.json` clean; `tsconfig.web.json` only the pre-existing TS6307 noise. oxlint + oxfmt clean.

Issue #6896, opened 2026-06-30
[Bug]: Worktree "Local setup command" is built as a cmd.exe invocation on Windows — fails to run under the configured Git Bash shell


VWforEI:
### Operating system

Windows

### Orca version

1.4.110

### Environment
- Orca 1.4.110
- Windows 11 (10.0.26200)
- Default terminal shell: Git Bash (`terminalWindowsShell: "git-bash"`)

### Summary
When a worktree is created, the interactive terminal correctly opens in Git Bash (my configured shell). However, the **Local setup command** is built as a Windows cmd invocation (`cmd.exe /c "orca/setup-runner.cmd"`) and sent into that Git Bash terminal, ignoring the configured shell. This fails in two ways at once: (1) the runner is a cmd batch file, so setup commands written in shell syntax can't run; and (2) the `cmd.exe /c ...` launch line itself is mangled by Git Bash's path conversion, so it never actually executes.

### Repro
1. Set default Windows shell to Git Bash.
2. Configure a worktree Local setup command using POSIX syntax:
   ```
   cp ~/Documents/test/.env.local .env.local
   pnpm install
   ```
3. Create a new worktree.

### Expected
The setup command runs in the configured shell (Git Bash), so `cp` and `~` work.

### Actual
The Git Bash terminal launches but immediately shells out to cmd, and the setup never runs:
```
$ cmd.exe /c "C:/Users/VW/Documents/test/.git/worktrees/<wt>/orca/setup-runner.cmd"
Microsoft Windows [Version 10.0.26200.8737]
(c) Microsoft Corporation. All rights reserved.

C:\Users\VW\orca\workspaces\test\<wt>>
```

Two distinct failures, same root cause:

**1. The runner does not execute at all.** The command `cmd.exe /c "<path>"` is sent into a Git Bash shell, but Git Bash's MSYS path conversion rewrites the `/c` argument into a Windows path. cmd.exe therefore never receives the `/c` switch and opens *interactively* (hence the version banner and the empty `C:\...>` prompt) instead of running the runner. Reproduced directly in Git Bash:
```
$ cmd.exe /c "echo HELLO"
Microsoft Windows [Version 10.0.26200.8737]
(c) Microsoft Corporation. All rights reserved.

C:\Users\VW\Documents\test>          # "HELLO" never prints; interactive cmd opens

# git-bash-safe forms both work:
$ cmd.exe //c "echo HELLO"           # -> HELLO
[...]

PR #12406, opened 2026-08-04
fix(worktrees): stop silently switching existing Windows setup scripts to Git Bash

OrcaWin:
## Defect

`resolveSetupRunnerShell` (src/main/hooks.ts) derives the native-Windows setup-runner shell from the `terminalWindowsShell` **terminal** preference. When that preference resolves to Git Bash it returns `{ family: 'posix' }`, and `createWorktreeRunnerScript` writes the project's setup script to a `.sh` file launched with `bash`.

That was applied retroactively to setup scripts that already existed. Before #6967 every native-Windows setup script ran under `cmd.exe`, so **every** pre-existing Windows setup script is cmd-authored.

## Concrete failure scenario

A Windows user has `terminalWindowsShell` set to Git Bash — a perfectly reasonable terminal preference — and an `orca.yaml` setup hook written in batch:

```yaml
scripts:
  setup: |
    copy .env.example .env
    set NODE_ENV=test
    if errorlevel 1 exit /b 1
    xcopy /E assets dist
```

They upgrade. The same script is now handed to bash: `copy` and `xcopy` do not exist, `set NODE_ENV=test` sets positional parameters instead of an env var, `if errorlevel 1` is a syntax error, and `assets dist` backslash paths get eaten by bash escaping. Setup breaks with no migration and no warning, and it looks like Orca broke their project.

There is a second, steady-state half of the same defect: because the interpreter came from a *per-user* preference, two contributors on the same repo got different interpreters for the same `orca.yaml`. No project could write a Windows setup script that worked for all of its contributors.

## The fix

The interpreter is a property of the script, not of the terminal the user likes, so the script declares it the standard way — a leading `#!` line.

- Native Windows keeps the historical `.cmd` runner **unless** the script's first line is a `#!` naming a POSIX shell. No existing script changes behavior on upgrade.
- `resolveSetupRunnerShell` keeps its existing role as the *feasibility* gate: a bash runner still requires the configured terminal to resolve to Git Bash, because the launch command is typed into that shell and the runner uses MSYS `/c/...` paths that Cygwin and the WSL shim reject. Its return value now means "a bash runner could launch here", and the per-script `#!` decides whether one is used.
- `buildWindowsRunnerScript` drops a leading `#!` line instead of `call`ing it. A declared-bash script that falls back to cmd (Git Bash not installed) now fails on a real setup line rather than aborting on errorlevel at line one.
- New `src/shared/setup-script-shebang.ts` parses the interpreter line: `#!/usr/bin/env bash`, `#!/bin/sh -e`, `#!/usr/bin/env -S bash -euo pipefail`, and Windows `bash.exe` paths all count; `node`/`python3` and comment-only first lines do not.

Issue-command runners go through the same code path and get the same rule, so an issue command authored for cmd is not flipped either.

### Why this option

The alternatives were weighed:

- *Record the shell a script was authored against.* Requires persistent per-repo state, can't distinguish "upgrading user" from "new user", and freezes a repo on whatever it was first run with — a user who later writes a bash script stays on cmd forever.
- *Detect batch-only idioms.* A heuristic guessing user intent, which is the same class of silent-wrong-behavior this PR is fixing, and it still has to pick a side for the large majority of scripts that are ambiguous (`pnpm install`).
- *A `#!` line.* Explicit, zero state, zero heuristics, standard, per-project, and identical for every contributor who checks the repo out. Chosen.

Unchanged: WSL worktrees, macOS/Linux (always bash), and SSH hosts (runner format follows the remote path format, never local Windows preferences).

## Verification

All tests use runtime platform overrides, so they run on any host.

New/changed tests:
- `src/shared/setup-script-shebang.test.ts` — interpreter-line parsing, including CRLF, `env -S`, non-shell interpreters, and a non-first-line `#!`.
- `src/main/hooks.test.ts` — "keeps batch setup scripts on cmd.exe when the terminal is Git Bash" (the reported regression: `{ family: 'posix' }` + `copy`/`xcopy` script still produces `setup-runner.cmd`), plus the issue-command equivalent, plus the existing POSIX-runner test updated to declare a shebang.
- `src/main/hooks-runner.test.ts` — `buildWindowsRunnerScript` drops a leading shebang instead of `call`ing it.

I reverted `src/main/hooks.ts` (`git stash push src/main/hooks.ts`), confirmed all three new regression tests fail without the fix, then restored it (`git stash pop`) and confirmed they pass.

```
  src/main/hooks.test.ts src/main/hooks-runner.test.ts \
  src/shared/setup-script-shebang.test.ts src/shared/setup-runner-command.test.ts \
  src/main/ipc/worktrees-windows.test.ts src/main/ipc/worktrees.test.ts
# 6 files, 379 tests passed

  src/shared/setup-agent-sequencing.test.ts src/shared/setup-agent-sequencing.windows.test.ts \
  src/main/ipc/worktree-logic.test.ts src/main/ipc/worktree-logic-wsl.test.ts \
  src/main/runtime/orchestration
# 32 files, 406 tests passed

npx tsc --noEmit -p config/tsconfig.node.json   # clean
npx oxlint <touched files>                       # clean
npm run check:max-lines-ratchet                  # OK, no new bypasses
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,7 @@ docs/**
 !docs/reference/headless-linux-server.md
 !docs/reference/linux-glibc-compatibility.md
 !docs/reference/relay-grace-time-reconfiguration.md
+!docs/reference/windows-setup-shell.md
 
 # Stably CLI (only docs/ are tracked)
 .stably/*
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,6 +29,7 @@ Orca targets macOS, Linux, and Windows. Keep all platform-dependent behavior beh
 - **Keyboard shortcuts**: Never hardcode `e.metaKey`. Use a platform check (`navigator.userAgent.includes('Mac')`) to pick `metaKey` on Mac and `ctrlKey` on Linux/Windows. Electron menu accelerators should use `CmdOrCtrl`.
 - **Shortcut labels in UI**: Display `⌘` / `⇧` on Mac and `Ctrl+` / `Shift+` on other platforms.
 - **File paths**: Use `path.join` or Electron/Node path utilities — never assume `/` or `\`.
+- **Windows setup scripts**: the setup/issue-command runner is a `.cmd` batch file unless the script starts with a `#!` line. Never derive it from the user's terminal-shell preference. See [`docs/reference/windows-setup-shell.md`](./docs/reference/windows-setup-shell.md).
 - **Linux native modules**: keep the glibc floor at Ubuntu 20.04 / glibc 2.31. A module compiled from source on a newer runner can reference symbol versions absent on the floor and crash the app on startup. See [`docs/reference/linux-glibc-compatibility.md`](./docs/reference/linux-glibc-compatibility.md); packaging fails if a bundled native binary needs newer glibc.
 
 ## SSH Use Case
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,7 +29,7 @@ Orca targets macOS, Linux, and Windows. Keep all platform-dependent behavior beh
 - **Keyboard shortcuts**: Never hardcode `e.metaKey`. Use a platform check (`navigator.userAgent.includes('Mac')`) to pick `metaKey` on Mac and `ctrlKey` on Linux/Windows. Electron menu accelerators should use `CmdOrCtrl`.
 - **Shortcut labels in UI**: Display `⌘` / `⇧` on Mac and `Ctrl+` / `Shift+` on other platforms.
 - **File paths**: Use `path.join` or Electron/Node path utilities — never assume `/` or `\`.
-- **Windows setup scripts**: the setup/issue-command runner is a `.cmd` batch file unless the script starts with a `#!` line. Never derive it from the user's terminal-shell preference. See [`docs/reference/windows-setup-shell.md`](./docs/reference/windows-setup-shell.md).
+- **Windows setup scripts**: the setup/issue-command runner is a `.cmd` batch file unless the script starts with a `#!` line — never derive that from the user's terminal-shell preference, and never launch a `.cmd` runner with a bare `cmd.exe /c` from a Git Bash pane (MSYS rewrites the `/c`). See [`docs/reference/windows-setup-shell.md`](./docs/reference/windows-setup-shell.md).
 - **Linux native modules**: keep the glibc floor at Ubuntu 20.04 / glibc 2.31. A module compiled from source on a newer runner can reference symbol versions absent on the floor and crash the app on startup. See [`docs/reference/linux-glibc-compatibility.md`](./docs/reference/linux-glibc-compatibility.md); packaging fails if a bundled native binary needs newer glibc.
 
 ## SSH Use Case
--- a/docs/reference/windows-setup-shell.md
+++ b/docs/reference/windows-setup-shell.md
@@ -35,13 +35,45 @@ language a project's setup script is written in. Deriving the runner from it had
 
 A `#!` line is per-project, explicit, and identical for everyone who checks the repo out.
 
+The same rule applies to the per-user setup command in **Settings → repository hooks**
+(`repo.hookSettings.scripts.setup`): it is merged into the same script that reaches the runner, so a
+POSIX one-liner stored there needs its own `#!` line to run under bash on Windows.
+
+## What the `#!` line does and does not select
+
+The generated runner is always executed by bash (`bash <runner>`; Git Bash on native Windows), on
+every platform. The `#!` line therefore does two things:
+
+- It declares the script is written for a POSIX shell, which is what selects the bash runner.
+- Its option flags are replayed with `set`, so `#!/usr/bin/env -S bash -euo pipefail` really does
+  get `pipefail`. Without that replay the flags would be silently dropped, because `bash <runner>`
+  never parses the interpreter line.
+
+The interpreter name itself is not honored beyond "is this a POSIX shell": `#!/bin/sh` and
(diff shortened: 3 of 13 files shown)


OrcaWin:
Adversarial review found the fix was correct at the runner-generation layer but reopened issue #6896 at the launch layer. Addressed:

**#1 (blocking) — `cmd.exe /c` from a Git Bash pane.** The PR redefined `WorktreeSetupLaunch.shell` to mean "the format the runner file was written in", but `resolveSetupRunnerCommand` reads it as "the shell that types the launch command". A Git Bash terminal + batch setup script therefore produced `cmd.exe /c "C:\...\setup-runner.cmd"` typed into a bash pane, where MSYS rewrites `/c` into a drive path — cmd opens interactively and setup never runs. Root cause fixed by restoring the contract: `shell` is the terminal's family again (`src/main/hooks.ts`), the runner file's `.cmd`/`.sh` extension carries the format, and `resolveSetupRunnerCommand` now branches on both — a batch runner launched from a POSIX pane reuses the existing PowerShell `ProcessStartInfo` launcher (`buildWindowsCmdRunnerDelayedLaunchCommand`), which is already the repo's MSYS-safe shape. WSL, PowerShell/cmd panes, SSH remotes, and `.sh` runners keep their exact previous commands.

**#2 (should-fix) — untested composition.** Added coverage for the newly-possible "posix pane + cmd runner" combination at every layer that builds a launch string: `setup-runner-command.test.ts` (no `bash`, no bare `cmd.exe /c`, native path preserved), `hooks.test.ts` end-to-end from `createSetupRunnerScript` through `buildSetupRunnerCommand`, `setup-agent-sequencing.test.ts` (wait-for-setup gate), and `setup-completion-signal.test.ts` (observed setup).

**#3 (should-fix) — declared shebang only used as a boolean.** `#!/usr/bin/env -S bash -euo pipefail` lost `pipefail` because the runner is launched as `bash <path>`, which never parses the interpreter line. The generated posix runner now replays declared option flags through `set` and drops the duplicate interpreter line (all platforms). Docs state plainly what the `#!` line does and does not select — the interpreter name is only read as "is this a POSIX shell", and sh/zsh scripts run under bash exactly as they already did on macOS/Linux.

**#4 (should-fix) — partial execution of a bash body under cmd.** Dropping the `#!` line meant a bash script reaching the cmd runner (PowerShell/cmd terminal, or any SSH-to-Windows host, which always uses `buildWindowsRunnerScript`) executed its interpreter-agnostic prefix before dying mid-way. The generated `.cmd` now prints why and `exit /b 1` without running anything, restoring the pre-PR "abort at line 1" with a message that says what to do.

**#5 (nit) — partial.** `docs/reference/windows-setup-shell.md` now covers the per-user setup command in repository hook settings (the exact configuration #6896 was filed against) and the launch-shell contract. The settings-UI hint is not included: it needs a new i18n key plus translations in 5 locale files, which is out of proportion for a nit — happy to add it separately.

**Verification.** Every new/updated test was proven to fail with only the source reverted (`git checkout -- src/main/hooks.ts src/shared/setup-runner-command.ts src/shared/setup-script-shebang.ts`, restored via patch — no stash). 16 failures without the fix, all green with it. Suites run: hooks, hooks-runner, setup-script-shebang, setup-runner-command, setup-agent-sequencing (+windows), orchestration, worktrees, worktrees-windows, worktree-activation, launch-worktree-background-terminals, orca-runtime (763 + 1016 pass). `tsc -p config/tsconfig.node.json` clean; `tsconfig.web.json` clean except pre-existing TS6307; `oxlint` and `oxfmt --check` clean on touched files. The original fix and its three regression tests are intact.

OrcaWin:
Round-2 findings addressed in def7374658.

**#1 — `set` replayed invocation-only shebang flags (should-fix, confirmed).** Reproduced: `#!/bin/bash -l` produced a runner containing `set -l`, which exits 2 and — under the runner's own `set -e` — aborted setup before the first line. `parseShellOptions` now allowlists only the flags `set` documents (`[--abefhkmnptuvxBCHP] [-o option]`); `-l`/`-s`/`-i`/`-r` are dropped, and a trailing `-o`/`-euo` with no option name is dropped whole instead of dumping the shell-option table into the Setup terminal. Proof is an executable test (`src/main/hooks-runner.test.ts`) that writes the generated runner and runs it under bash.

**#2 — sequencing gate language followed the runner file, not the pane (should-fix, confirmed).** Reproduced: git-bash pane → `quoteStartupArg` posix → `claude 'fix the user'\''s login'`, then a `.cmd` runner drove `createSequencedSetupAgentCommands` into the PowerShell branch, whose `Invoke-Expression` cannot parse that. Fix is one condition in `setup-agent-sequencing.ts`: the gate follows the pane's shell family, since the pane types the gate and already quoted the startup command for itself. The batch runner still launches through `buildWindowsCmdRunnerDelayedLaunchCommand` (the #6896 fix), now wrapped by the bash gate rather than the PowerShell one — so it is never handed to bash, and the marker is both written and read by bash (no PowerShell-CRLF vs `IFS=: read` mismatch).

No new timeouts, retries, or async paths. The `not.toContain('bash -lc')` assertion in the Git-Bash-pane test was replaced with the invariant it was actually protecting (`not.toMatch(/bash\s+\S*setup-runner/)`) plus new assertions that the gate is `bash -lc` and evals the POSIX startup command.

Verification: touched-area vitest (10 files, 229 passed / 1 skipped), `tsc` node + web, and `check-changed-code-quality.mjs` (0 new findings incl. react-doctor). Original defect re-confirmed: reverting only `src/main/hooks.ts` to main fails 7 tests, including `keeps batch setup scripts on cmd.exe when the terminal is Git Bash`. Each new test verified to fail with only its source change reverted.

PR #12407, opened 2026-08-04
fix(codex): do not silently run the previous account when managed auth is unreadable

OrcaWin:
## Defect

`CodexRuntimeHomeService.syncForCurrentSelection` treats a missing/unreadable managed `auth.json` as "not settled yet" and returns early to protect the selection from a rotation race. In the shared runtime-home lane that early return leaves the runtime home exactly as it was — holding the **previously synced account's** credentials — and `prepareForCodexLaunch` then hands that home to the launch.

The credential-absence grace window therefore answered "I can't read account B" with "then run as account A".

## Failure scenario

1. Account A is selected and synced; the shared runtime home holds A's `auth.json`.
2. The user switches to account B. B's `auth.json` cannot be read — home on a disconnected network drive, never provisioned, removed, or held under an `EPERM`/`EACCES` lock by Windows AV.
3. `syncForCurrentSelection` sees `missing`/`unreadable`, the absence is inside the grace window, and it returns without touching the runtime home.
4. Settings (and the account picker) show **B**. Codex launches with the runtime home still containing **A**'s credentials.

Result: wrong quota, wrong identity, work attributed to the wrong account — silently. The readiness gate in `resolveCodexHomeAfterManagedAuthReadiness` cannot catch this, because the shared runtime home is not one of the managed account homes it inspects, so `isCodexHomeAuthReadyForLaunch` reports it ready regardless of whose credentials are inside.

## Fix

Absence of evidence about B is not evidence for A. The grace window now short-circuits **only when Orca can prove the runtime home already holds the selected account** — via the shared-runtime-auth provenance record, falling back to the in-memory `lastSyncedAccountId` on pre-provenance installs that have no record to consult.

When it cannot prove that, the runtime auth is removed and the provenance is fenced:

- the selection survives (the absence may still heal, and a settled read decides later, exactly as before);
- the launch is **logged out rather than logged in as someone else**;
- a warning is logged naming the condition.

Before removing them, the mirrored bytes are read back to whichever managed account owns them (identity + freshness proven, the same helper an account switch uses). On a cold start whose mirror still holds the previous account, no outgoing read-back has run yet, so a token refresh Codex wrote there exists nowhere else — with refresh-token rotation, dropping it would force that account to log in again.

The durable path (absence outliving the grace window → deselect + restore the system default) is unchanged, as is the legitimate case this window was built for: when Codex rewrites the *selected* account's `auth.json` in place, provenance proves the runtime home is already that account, so its credentials survive the race untouched.

## Verification

- New `runtime-home-service.test.ts` cases: with A synced and B selected, `prepareForCodexLaunch()` must not leave A's credentials in the runtime home — parameterized over ENOENT (`auth.json` removed) and EACCES/EPERM (`chmod 000`, skipped on Windows/root).
- New cold-start case: the mirror holds account A's *refreshed* tokens with provenance `managed/account-a` while settings already select B (whose `auth.json` was never written) — the refresh lands in A's managed home and the mirror is cleared.
- New still-settling case: torn mid-write `auth.json` for the *selected* account keeps the runtime credentials and the selection through the grace window.
- `codex-credential-absence-grace.test.ts`: a permission-denied read stays transient inside the window and turns durable after it.
- `managed-codex-auth-readiness.test.ts`: `readStoredCodexCredentialState` separates `missing` (ENOENT) from `unreadable` (permission denied).
- Reverted `runtime-home-service.ts` to `origin/main` and confirmed all three new wrong-account cases fail (`expected A's auth not to be A's auth`), then restored it and confirmed they pass.
- `npx vitest run --config config/vitest.config.ts src/main/codex-accounts/` → 13 files, 285 tests passing.
- `npx tsc --noEmit -p config/tsconfig.node.json` clean; `oxlint` (default + code-quality-native) clean; max-lines ratchet clean (no new suppressions).

Platform notes: no platform-specific behavior added; the EACCES tests self-skip on Windows and as root. The WSL and self-contained per-account-home lanes are untouched (they launch under the account's own home, so they never carry another account's credentials).

### DIFF
--- a/src/main/codex-accounts/codex-auth-identity.ts
+++ b/src/main/codex-accounts/codex-auth-identity.ts
@@ -50,6 +50,26 @@ export function codexAuthMatchesManagedAccount(
   )
 }
 
+// Why: an unreadable managed home cannot be compared byte-for-byte, so only the
+// account record itself can rule that home out as the credential's owner.
+export function codexAuthCouldBelongToManagedAccount(
+  runtimeAuthContents: string,
+  account: CodexManagedAccount
+): boolean {
+  const identity = readIdentityFromAuthContents(runtimeAuthContents)
+  if (!identity) {
+    return true
+  }
+  const accountEmail = normalizeField(account.email)
+  if (accountEmail && identity.email && accountEmail !== identity.email) {
+    return false
+  }
+  return (
+    identityFieldMatches(normalizeField(account.providerAccountId), identity.providerAccountId) &&
+    identityFieldMatches(normalizeField(account.workspaceAccountId), identity.workspaceAccountId)
+  )
+}
+
 // Why: the shared mirror may still hold managed credentials; only the same
 // positively identified system account may ever be read back to ~/.codex.
 export function codexAuthMatchesSystemDefaultIdentity(
--- a/src/main/codex-accounts/codex-auth-identity.ts
+++ b/src/main/codex-accounts/codex-auth-identity.ts
@@ -29,8 +29,17 @@ export function codexAuthMatchesManagedAccount(
     normalizeField(account.workspaceAccountId),
     managedIdentity?.workspaceAccountId
   )
-  const emailMatches = Boolean(selectedEmail && identity.email && selectedEmail === identity.email)
-  if (selectedEmail && identity.email && selectedEmail !== identity.email) {
+  const emailMatches = identityFieldsAgree(selectedEmail, identity.email)
+  if (
+    identityContradictsSelection(
+      {
+        email: selectedEmail,
+        providerAccountId: selectedProviderId,
+        workspaceAccountId: selectedWorkspaceId
+      },
+      identity
+    )
+  ) {
     return false
   }
   if (!identityFieldMatches(selectedProviderId, identity.providerAccountId)) {
@@ -57,16 +66,16 @@ export function codexAuthCouldBelongToManagedAccount(
   account: CodexManagedAccount
 ): boolean {
   const identity = readIdentityFromAuthContents(runtimeAuthContents)
-  if (!identity) {
-    return true
-  }
-  const accountEmail = normalizeField(account.email)
-  if (accountEmail && identity.email && accountEmail !== identity.email) {
-    return false
-  }
   return (
-    identityFieldMatches(normalizeField(account.providerAccountId), identity.providerAccountId) &&
-    identityFieldMatches(normalizeField(account.workspaceAccountId), identity.workspaceAccountId)
+    !identity ||
+    !identityContradictsSelection(
+      {
+        email: normalizeField(account.email),
+        providerAccountId: normalizeField(account.providerAccountId),
+        workspaceAccountId: normalizeField(account.workspaceAccountId)
+      },
+      identity
+    )
   )
 }
 
@@ -259,4 +268,34 @@ function firstNonNull(...values: (string | null | undefined)[]): string | null {
 
 function identityFieldMatches(selectedField: string | null, runtimeField: string | null): boolean {
   return !selectedField || Boolean(runtimeField && selectedField === runtimeField)
+}
+
+// Why: only a claim both sides carry can rule an account out — a credential with
+// no identity claims (api key, PAT, bedrock) contradicts nothing. A ChatGPT
+// rename leaves the record email stale, so a matching account id outranks it.
+function identityContradictsSelection(
+  selected: CodexAuthIdentity,
+  identity: CodexAuthIdentity
+): boolean {
+  if (
+    identityFieldsConflict(selected.providerAccountId, identity.providerAccountId) ||
+    identityFieldsConflict(selected.workspaceAccountId, identity.workspaceAccountId)
+  ) {
+    return true
+  }
+  return (
+    identityFieldsConflict(selected.email, identity.email) &&
+    !identityFieldsAgree(selected.providerAccountId, identity.providerAccountId)
+  )
+}
+
+function identityFieldsAgree(selectedField: string | null, runtimeField: string | null): boolean {
+  return Boolean(selectedField && runtimeField && selectedField === runtimeField)
+}
+
+function identityFieldsConflict(
+  selectedField: string | null,
+  runtimeField: string | null
+): boolean {
+  return Boolean(selectedField && runtimeField && selectedField !== runtimeField)
 }
--- a/src/main/codex-accounts/codex-credential-absence-grace.test.ts
+++ b/src/main/codex-accounts/codex-credential-absence-grace.test.ts
@@ -1,4 +1,4 @@
-import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { chmodSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, beforeEach, describe, expect, it } from 'vitest'
@@ -53,6 +53,21 @@ describe('CodexCredentialAbsenceGrace', () => {
     expect(grace.assess(authPath, 100_000)).toEqual({ state: 'missing', durable: false })
     expect(grace.assess(authPath, 100_000 + CODEX_CREDENTIAL_ABSENCE_GRACE_MS)).toEqual({
       state: 'missing',
+      durable: true
+    })
+  })
+
+  it('keeps a permission-denied read inside the grace window before it turns durable', () => {
+    if (process.platform === 'win32' || process.getuid?.() === 0) {
+      return
+    }
+    const grace = new CodexCredentialAbsenceGrace()
+    writeFileSync(authPath, VALID_AUTH, 'utf-8')
+    chmodSync(authPath, 0o000)
+
+    expect(grace.assess(authPath, 1_000)).toEqual({ state: 'unreadable', durable: false })
+    expect(grace.assess(authPath, 1_000 + CODEX_CREDENTIAL_ABSENCE_GRACE_MS)).toEqual({
+      state: 'unreadable',
       durable: true
     })
   })
--- a/src/main/codex-accounts/managed-codex-auth-readiness.test.ts
+++ b/src/main/codex-accounts/managed-codex-auth-readiness.test.ts
@@ -1,4 +1,4 @@
-import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { chmodSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { afterEach, describe, expect, it, vi } from 'vitest'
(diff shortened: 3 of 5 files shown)


brennanb2025:
## Review + QA

Reviewed on this branch in an isolated worktree. One defect found and fixed (pushed as `af3cfead4d`); two things about the framing worth resolving before merge.

### Fixed: the pre-delete read-back destroyed an uninvolved account's refresh token

`clearRuntimeAuthForUnprovenSelection` reads the mirrored bytes back to their owning account before `rmSync`. That read-back scans every managed account's `auth.json`, and an **unreadable** one threw. The throw is swallowed by the catch in `readBackRefreshedTokensFromPath`, which returns `'rejected'` — so the read-back silently no-opped and the delete still ran.

The trigger is self-inflicted: the only way to reach this delete is a managed `auth.json` that cannot be read, which is exactly the account that aborts the scan. So in the EACCES/EPERM variant this PR targets, a rotated refresh token belonging to a *different, uninvolved* account is destroyed — a forced re-login under refresh-token rotation. It was also cross-contaminating: one unreadable home poisoned the identity match for every other account.

The existing preservation test missed it because it removes `auth.json` (ENOENT) rather than making it unreadable. Fix is a `continue` on unreadable, matching the ENOENT `continue` two lines above; the test is now parameterized over both.

### The description overstates the defect

The stated failure — "Settings show B, Codex launches with A's credentials" — does not appear to be reachable in a shipped build:

- `isCodexSystemDefaultRealHomeEnabled()` returns `true` unconditionally, and its own docstring says no production path sets the override. Every setter in the tree is a test/benchmark/e2e harness pinning `'0'`.
- With it on, `syncForCurrentSelection` returns at the self-contained-lane guard for any non-WSL managed selection, and `prepareForCodexLaunch` returns the account's *own* home. WSL returns even earlier.
- The changed branch requires a managed host account selected **and** `getSelfContainedManagedHostAccount()` to be null — mutually exclusive while the flag is on.

So this is hardening of the flag-off rollback lane, not a live user-facing wrong-account bug. Still worth having; the body should say which it is, since the current framing sets the wrong merge priority.

### Reported, deliberately not fixed

- The `lastSyncedAccountId` fallback is **vacuously true on cold start** — it is initialized from the selection itself, so on the first sync after upgrade (no provenance record yet) an unreadable selected account keeps the other account's credentials. This is the original bug, narrowed to that window. Fixing it soundly needs identity matching, which would wipe credentials for accounts with no strong id — a design call, not a review edit.
- A `system-default`-owned mirror, and a provenance owner whose managed home no longer exists, are both deleted with no read-back. Same data-loss shape as the fixed bug, different entry paths.
- `rmSync` with `{force: true}` only suppresses ENOENT; an EPERM (Windows AV/EBUSY lock) on the mirror propagates into an unguarded launch path. Pre-existing class in that chain.

### Verification

Reverting only `runtime-home-service.ts` to `origin/main` turns the three new wrong-account cases red and restores to `285 passed`, so the tests are not vacuous. The EACCES cases genuinely executed here (macOS, non-root) rather than self-skipping. Full `src/main/codex-accounts/` suite, `tsc`, `oxlint` (default + code-quality-native), the changed-code quality gate, React Doctor, the max-lines ratchet, and `oxfmt` are all clean.

App launches clean on this branch with no `codex-runtime-home` warnings at startup — i.e. the new clear-path does not fire spuriously:



That screenshot is a startup no-regression check only. Live QA cannot exercise the changed branch without setting the rollback env var and fabricating managed account state, so the real evidence here is the unit tests, not the app. Windows, WSL, and network-drive behavior remain unverified.

brennanb2025:
## Correction and follow-up on the read-back finding

Two corrections to my earlier comment, both from A/B testing against `origin/main` rather than reasoning.

**1. The `readFileSync` throw is pre-existing, not introduced here.** `findManagedAccountForRuntimeAuth` already exists on `origin/main` (line 1228) with the same unguarded read and no `try`/`catch`. Since it backs `readBackRefreshedTokensFromPath`, which has ~7 call sites, the "one unreadable home poisons the identity match for every account" defect affects **every** read-back path, not just the new one. This PR added a caller that exposed it. That makes `af3cfead4d` more valuable than I framed it — it repairs a shared helper — but it is not a regression this PR introduced.

**2. The remaining rejection paths do not represent new data loss either.** I suggested the `rmSync` might destroy a rotated refresh token that `main` would have kept. Tested, and that is wrong. In the drive-disconnect shape (whole `codex-accounts` root unreachable, mirror holds A's refreshed tokens, B selected), on `origin/main` the mirror still holds those bytes at t0 — but once the 5s grace expires the durable branch deselects and `restoreSystemDefaultSnapshot` writes over the mirror, so A's rotation is gone from both the mirror and A's home there too. Same end state, reached about five seconds later. If the absence instead heals, `writeRuntimeAuth` overwrites the mirror on `main` for the same result.

So this PR moves an existing destruction earlier by at most the grace window. It does not create a loss that did not already exist, and it should ship as-is.

### Follow-up worth filing separately

`readBackRefreshedTokensFromPath` returns `'rejected'` — persisting nothing — in three cases:

- the owning account's managed `auth.json` is **unreachable** (`:1281`); note this is the *same* incident as the trigger, since both homes live under one `codex-accounts` root;
- `match.kind === 'ambiguous'` (`:1301`, duplicate account records sharing an id);
- restart with `lastWrittenAuthJson === null` where `compareCodexAuthFreshness` returns `null` because the blob carries no `expires_at`/`exp`/`iat` (`codex-auth-identity.ts:105-119`) — the widest of the three.

Four sites then drop those bytes: `clearRuntimeAuthForUnprovenSelection` (`:957`), the `writeRuntimeAuth` overwrite (`:804`), `restoreSystemDefaultSnapshot` (`:1879`/`:1893`/`:1912`/`:1918`/`:1927`), and `clearRuntimeAuthAfterSystemDefaultLogout` (`:1944`).

The decision is between accepting the loss with a log line, or quarantining the unpersistable mirror with a real lifecycle — who reads it back, when it is GC'd, and how long unowned credential material may sit on disk — covering **all four** sites rather than one. Worth recording a dead end: targeting the owner's home via provenance `accountId` fixes the unreadable-owner case but not the unreachable-drive case, since you cannot write to a drive that is gone.

Reachability is the shared-mirror lane only (`ORCA_CODEX_SYSTEM_DEFAULT_REAL_HOME=0`), so this is not a shipped-configuration issue.

OrcaWin:
Addressed the adversarial review findings in ae7904dc (hardening on top of the original fix, which is unchanged in intent).

**What the reviewer found**

1. `fenced` provenance made the grace window log out the account the mirror *correctly* held (a torn provenance file or a crashed pending write is "unknown owner", not "someone else's").
2. The pre-provenance fallback was tautological: `initializeLastSyncedState` seeds `lastSyncedAccountId` with the **selected** account, so `missing && lastSyncedAccountId === id` is true on every cold start — the headline wrong-account launch still reproduced on installs with no provenance file.
3. The pre-delete read-back only rescued *managed*-owned bytes, so a system-default refresh living only in the mirror was destroyed with no path back to `~/.codex`.
4. `rmSync` was unguarded on a branch whose entry condition is "the filesystem refused a read" (Windows AV/EPERM/EBUSY), and the fence was persisted *after* the delete.
5. Swallowing managed-home read errors let a candidate set that would have been rejected as ambiguous collapse to a single match and be written to the wrong home.

**What changed**

- Ownership is now proven from the credential itself: `codexAuthMatchesManagedAccount(runtimeAuth, account, null)` matches the mirror's identity claims against the account record without reading the unreadable home. Provenance is corroboration, not the sole evidence — so a fenced/corrupt provenance file no longer causes a logout (1), and a cold start with no provenance file now clears a mirror that identity-contradicts the selection (2). The tautological `missing && lastSyncedAccountId === id` fallback is replaced by a real proof: `lastSyncedAccountId === id && lastWrittenAuthJson === runtimeAuth` (bytes Orca itself mirrored this run).
- `clearRuntimeAuthForUnprovenSelection` now rescues a mirror-only system-default refresh into `~/.codex` when the managed read-back rejects, reusing the exact ownership rules `syncRuntimeAuthWithSystemDefault` already applies (extracted to `resolveSystemDefaultMirrorClaim`, so the two paths cannot drift) (3).
- The clear fences *before* deleting and is wrapped in try/catch, so a refused delete leaves the surviving bytes marked untrusted and never throws out of `prepareForCodexLaunch` (4).
- An unreadable managed home whose account record cannot rule it out (`codexAuthCouldBelongToManagedAccount`) now makes the no-expectation scan `ambiguous` instead of misattributing the bytes; homes whose record positively contradicts the credential are still skipped, so the previous commit's "don't drop another account's refresh" behavior is preserved (5).

**Disclosure (finding 6):** this whole lane is the legacy shared-mirror path. `isCodexSystemDefaultRealHomeEnabled()` returns `true` in shipped builds, so host managed selections take `syncSelfContainedManagedSelection` and never reach this code; only the `ORCA_CODEX_SYSTEM_DEFAULT_REAL_HOME=0` opt-out/rollback lane does. This PR hardens that fallback lane — it is not a user-visible wrong-quota fix in a released build.

**Verification**

- 5 new tests, one per finding. With only `runtime-home-service.ts` + `codex-auth-identity.ts` reverted to the previous commit, exactly those 5 fail (136 pre-existing pass); with them restored, 141/141 pass.
- `npx vitest run --config config/vitest.config.ts src/main/codex-accounts/ src/main/codex/ src/main/agent-auth-restart-preservation.test.ts src/main/pty/codex-home-wsl-env.test.ts` → 54 files / 954 tests pass, 1 file + 6 tests skipped.
- `npx tsc --noEmit -p config/tsconfig.node.json` clean; `config/tsconfig.web.json` only pre-existing TS6307 noise; `oxlint`, `oxfmt --check`, `check-max-lines-ratchet` and `check-changed-code-quality` all clean (no new suppressions).

OrcaWin:
Round-2 findings addressed in 40c6d9d4f6 — all three reproduced first, each fix has a test that fails without it.

**1 (blocking) — fence cancelled the delete.** Confirmed: with the provenance path unwritable, `persistSharedRuntimeAuthProvenance({owner:'fenced'})` threw and `rmSync` never ran, so the launch kept account A's credentials under a B selection. `rmSync` now runs from a `finally`, so neither the read-back rescue nor the fence can skip it; the fence still goes first, so a delete the OS refuses still leaves the survivors marked untrusted. New test `drops another account from the mirror even when the fence write fails` fails on ae7904dcdd.

**2 (should-fix) — api-key/PAT/bedrock credentials could never prove ownership.** Confirmed. Fixed at the cause rather than by adding another proof: with no *committed* provenance record left, ownership is now **refuted, not proven** — only claims that contradict the account record rule the mirror out, and a claim the credential does not carry contradicts nothing. Committed provenance keeps deciding exactly as before, so `never launches another account on a cold start with no provenance record` and both original-defect tests still pass (a full ChatGPT credential for another account contradicts the selected record). New test: `keeps a mirrored api-key credential no identity claim can attribute`.

**3 (should-fix) — stale record email hard-failed the proof.** Confirmed, and it also silently discarded the renamed account's refresh instead of reading it back. A positive account-id match now outranks a mismatched email, expressed once in a shared `identityContradictsSelection` used by both `codexAuthMatchesManagedAccount` and `codexAuthCouldBelongToManagedAccount` so the two predicates cannot drift. Email alone still rules an account out when no account id agrees. New tests: `reads a renamed account refresh back into its own home despite a stale record email` and `keeps the mirror of a renamed account whose record email is stale`.

No test was weakened or deleted. Original defect still fixed: both `never launches the previously synced account…` tests pass at this commit and fail when only the two source files are reverted to a69ca6d5b7.

Verified: `vitest src/main/codex-accounts src/main/codex` (1022 passed), full `src/main` (one pre-existing unrelated `updater.test.ts` failure, identical with these changes reverted), `tsc` node + web, and the changed-code quality gate (0 new findings, react-doctor clean).

PR #12235, opened 2026-08-03
fix(mobile): surface worktree catalog failures instead of silent 0 worktrees (STA-3123)

brennanb2025:
## Summary

Fixes the "remote Orca server shows 0 worktrees" presentation from STA-3123 (synced GitHub issue #11869): a connected host whose `worktree.ps` request fails is no longer indistinguishable from a genuinely empty host.

- **Host page** (`/h/[hostId]`): when the catalog RPC fails (RPC error, invalid payload, or transport throw) and there are no cached rows, the screen now shows "Could not load workspaces from this host — `worktree.ps failed (<code>) — retrying automatically`" instead of an eternal spinner or a "No worktrees" empty state. The existing 3s poll / reconnect / pull-to-refresh paths clear the state on the next success; cached rows still win when present.
- **Home screen host card**: a failed `worktree.ps` no longer seeds a fabricated `0 worktrees` count. With no prior data the card now says "Worktree list unavailable"; cached counts are still kept through transient failures.
- `WorktreeCatalogSnapshotClient.fetch` now returns a discriminated result (`response` vs `request_failed` with the server error code) instead of collapsing failures to `null`, and preserves the snapshot token across failures.

Investigation notes: on current code the direct-pair catalog path is healthy end-to-end — verified via a fake mobile E2EE client (exact app params: `{limit: 10000, afterSnapshotId: null}`) against an isolated branch-built `orca serve` and against the installed production serve, plus a live iOS-sim pairing (screenshots below). The field "0 worktrees while CLI shows dozens" therefore reaches users through this silent-failure masking, which also made the original report undiagnosable — after this PR the failing host displays the actual RPC error code.

## Screenshots

Host page, paired to a connected mock host whose `worktree.ps` returns an RPC error (previously rendered as an empty host):

host page failure state

Home screen: failing host (Host 4) says "Worktree list unavailable"; healthy host (Host 3) keeps real counts:

home card unavailable

Healthy path unchanged — iOS sim paired directly to an isolated `orca serve` built from this branch lists its catalog:

healthy direct pairing

## Testing

- [x] `pnpm lint` (root, plus mobile `oxlint` — clean)
- [x] `pnpm typecheck` (root, plus mobile `tsc --noEmit` — clean)
- [x] `pnpm test` — targeted: mobile vitest `src/worktree/` (113 tests passing, includes new failure-path and state-precedence tests); changed-code quality gate (`check-changed-code-quality.mjs`): 0 new findings incl. React Doctor
[...]

PR #12408, opened 2026-08-04
fix(mobile): keep cached workspace counts across a transient RPC failure

OrcaWin:
## Defect

The mobile Home host card drops its worktree counts on **any** transient `worktree.ps` failure and never refetches them when the connection comes back.

Two independent bugs, both on `main` today:

1. **Render discards proven counts.** `markHomeWorktreeCatalogUnavailable` deliberately keeps the last successful counts in state, but `app/index.tsx` only passed them to the card when `catalogUnavailable` was unset — so the card fell through to `Worktree list unavailable` while the good numbers sat unused in state.
2. **Nothing re-drives the fetch.** The per-host effect latched `let statsFetched = false` on the first `'connected'`. The logical RPC client survives socket drops (`stable-logical-rpc-client` swaps the underlying session and re-publishes `'connected'`), and the effect is keyed on `hostId:clientKey`, so the latch was never reset. The one-shot never fired again.

## Failure scenario

Phone on Home showing `12 worktrees · 2 active`. The socket dies while the periodic `worktree.ps` is in flight — app backgrounded, Wi-Fi→cellular handoff, sleep/resume. The promise rejects, the card collapses to `Worktree list unavailable`, and the reconnect that follows re-subscribes notifications/accounts but never re-reads the catalog. The card stays wrong until the user navigates away and back. On a phone this is the normal case.

## Fix

- `home-worktree-info.ts`: a failed refresh now flags `staleCounts` alongside `catalogUnavailable`, and the new pure `homeHostWorktreeSummary()` owns the card's wording — `Last known: 12 worktrees · 2 active` when counts are proven-but-stale, `Worktree list unavailable` only when a catalog never loaded (preserves STA-3123: a failed fetch must never render as a healthy host with 0 worktrees).
- `host-connect-refetch-gate.ts` (new): fires on each transition **into** `'connected'`. `app/index.tsx` uses it instead of the `statsFetched` latch, so every reconnect re-reads stats / worktrees / task providers exactly once. No polling timer — it hangs off the existing `onStateChange` machinery, mirroring `useWorktreeResync` on the host screen. Repeated `'connected'` notifications can't storm the desktop (the logical client also dedupes identical states).
- `home-host-worktree-fetch.ts` (new): `fetchHomeHostWorktreeInfo` moved verbatim out of `app/index.tsx` so its rejection path is reachable from tests (mobile vitest only collects `src/**/*.test.ts`).
- `MobileHostCard` now takes a prebuilt `worktreeSummary` string instead of formatting counts itself.

No RPC/protocol change: the reconnect path re-issues existing methods, and `preflight.check` is cached desktop-side, so old desktops see no new load.

## Verification

- `mobile$ vitest run src/worktree src/transport` — 84 files, 557 passed.
- Full mobile suite: 381 passed; the 12 failures are pre-existing in this environment (`terminal-webview-engine.generated` is a postinstall build artifact that isn't present).
- `mobile$ tsc --noEmit` — clean apart from that same pre-existing generated-file error.
- **Reverted each half of the fix and watched the new tests fail, then restored:**
  - dropping `staleCounts` → `keeps showing the last proven counts after a failed refresh`, `keeps the last proven counts when the in-flight read rejects`, `re-reads the catalog on reconnect and clears the stale flag`, and the existing `marks cached counts unavailable without discarding the cached snapshot` all fail.
  - making the gate latch like the old `statsFetched` → `fetches again after a dropped socket comes back` fails (`expected 1 to be 2`) and the reconnect test never sees the refreshed 13/3 counts.

### DIFF
--- a/mobile/app/index.tsx
+++ b/mobile/app/index.tsx
@@ -18,14 +18,14 @@ import AsyncStorage from '@react-native-async-storage/async-storage'
 import { loadHosts } from '../src/transport/host-store'
 import { navigateToMobileHostEdit } from '../src/transport/host-edit-navigation'
 import { removeHostAndCloseClient } from '../src/transport/host-removal-lifecycle'
-import { pickResumeWorktree } from '../src/worktree/resume-worktree'
-import { WORKTREE_PS_FULL_LIMIT } from '../src/worktree/worktree-catalog-snapshot-client'
+import { fetchHomeHostWorktreeInfo } from '../src/worktree/home-host-worktree-fetch'
 import {
-  markHomeWorktreeCatalogUnavailable,
+  homeHostWorktreeSummary,
   type HomeWorktreeSummary,
   type HostWorktreeInfo
 } from '../src/worktree/home-worktree-info'
 import type { RpcClient } from '../src/transport/rpc-client'
+import { createHostConnectRefetchGate } from '../src/transport/host-connect-refetch-gate'
 import { sendSingleFlightRequest } from '../src/transport/request-single-flight'
 import {
   useAllHostClients,
@@ -136,55 +136,6 @@ function fetchStats(
       }
     })
     .catch(() => {})
-}
-
-function fetchWorktreeInfo(
-  client: RpcClient,
-  hostId: string,
-  setInfo: (
-    updater: (prev: Record<string, HostWorktreeInfo>) => Record<string, HostWorktreeInfo>
-  ) => void,
-  disposed: () => boolean
-) {
-  const markUnavailable = () => {
-    setInfo((prev) => {
-      const current = prev[hostId]
-      const next = markHomeWorktreeCatalogUnavailable(current, hostId)
-      return next === current ? prev : { ...prev, [hostId]: next }
-    })
-  }
-
-  sendSingleFlightRequest(client, hostId, 'worktree.ps', { limit: WORKTREE_PS_FULL_LIMIT })
-    .then((response) => {
-      if (disposed()) {
-        return
-      }
-      if (response.ok) {
-        const result = response.result as { worktrees: HomeWorktreeSummary[] }
-        const worktrees = result.worktrees ?? []
-        setCachedWorktrees(hostId, worktrees)
-        const activeStatuses = new Set(['working', 'active', 'permission'])
-        const active = worktrees.filter((w) => w.status && activeStatuses.has(w.status))
-        // Mirror the desktop's focused workspace (see pickResumeWorktree).
-        const lastActive = pickResumeWorktree(worktrees)
-        setInfo((prev) => ({
-          ...prev,
-          [hostId]: {
-            hostId,
-            totalWorktrees: worktrees.length,
-            activeCount: active.length,
-            lastActiveWorktree: lastActive
-          }
-        }))
-      } else {
-        markUnavailable()
-      }
-    })
-    .catch(() => {
-      if (!disposed()) {
-        markUnavailable()
-      }
-    })
 }
 
 function fetchAccountsSnapshot(
@@ -376,7 +327,7 @@ export default function HomeScreen() {
       for (const entry of allClientsRef.current) {
         if (entry.client.getState() === 'connected') {
           fetchStats(entry.client, entry.hostId, setStats, () => stale)
-          fetchWorktreeInfo(entry.client, entry.hostId, setWorktreeInfo, () => stale)
+          void fetchHomeHostWorktreeInfo(entry.client, entry.hostId, setWorktreeInfo, () => stale)
           fetchAccountsSnapshot(entry.client, entry.hostId, setAccountsByHost, () => stale)
           fetchTaskProviders(entry.client, entry.hostId, setTaskProvidersByHost, () => stale)
         }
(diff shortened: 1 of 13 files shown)


OrcaWin:
Addressed the review findings. Both `should-fix` items are fixed at the root, plus both nits.

**#1 — cutover leaves the card latched on "Last known" (`home-host-worktree-fetch.ts`)**
Confirmed: `stable-logical-rpc-client.ts` `migrateTo` rejects every pending request with `LogicalClientCutoverError` and then calls the state listeners *directly*, bypassing `publishState`'s dedupe — connected→connected, so the connect gate never re-arms. Fixed where the evidence points: a cutover rejection is an interrupted read, not an answered one, so `worktree.ps` is re-issued on the authenticated replacement (bounded by `CUTOVER_RETRY_LIMIT = 2`), matching `runtime-capability-probe.ts` and `worktree-create-retry.ts`. Re-arming the gate would have been wrong — a cutover only loses in-flight work; subscriptions are replayed.
New test wires the *real* `createStableLogicalRpcClient` to the real gate: after `migrateTo`, the gate fires 0 times and the card still lands on the fresh 13/3 counts. A second test proves the retry is capped and a genuinely wedged host still reports the failure.

**#2 — "Last known" had no age bound (`home-worktree-info.ts`)**
Confirmed: the home snapshot persists counts across launches and `app/index.tsx` hydrates them before any RPC, so a cold start whose first `worktree.ps` failed rendered day-old counts identically to seconds-old ones — exactly what STA-3123 (#12235) rendered as "Worktree list unavailable". Counts now carry `countsProvenAt`, stamped when the host answers, and `homeHostWorktreeSummary(info, now)` drops back to "Worktree list unavailable" past `HOME_WORKTREE_COUNTS_STALE_MAX_AGE_MS` (10 min). Counts persisted by a build without the stamp have an unknowable age and are treated as expired, so the up/downgrade path degrades to the old wording. This also closes the "connected host with a permanently failing `worktree.ps`" window the review called out — the label ages out on its own. `home-snapshot-cache.ts` now uses the canonical `HostWorktreeInfo` instead of a stripped-down copy, so the stamp keeps round-tripping.

**#3 (nit) — the wiring was untested**
`mobile/vitest.config.ts` only collected `src/**/*.test.ts`, so the existing `WorktreeAgentList.test.tsx` had never run. Widened to `.test.tsx` (that file passes) and moved the summary derivation into `MobileHostCard`, which now takes `HostWorktreeInfo` and calls `homeHostWorktreeSummary` itself — the caller can no longer re-gate the counts away, which was the original defect. `MobileHostCard.test.tsx` renders the card and asserts fresh / "Last known" / "unavailable" / nothing-yet. Verified it catches the regression: re-adding a `catalogUnavailable` gate inside the card fails it. The `if (reconnected)` call site is still untested — covering it needs a HomeScreen render (expo-router + AsyncStorage + a pile of mocks) or extracting the whole subscription lifecycle, which I judged riskier than the nit is worth.

**#4 (nit) — shared `stats` churn**
`stats` is now `Record<hostId, HomeStatsSummary>` summed by `totalHomeStats` for the header row, so a reconnect on any desktop no longer overwrites the globally displayed numbers with that host's. Single-host users see identical values; malformed replies are skipped instead of NaN-ing the row.

**Verification**
`mobile$ npx vitest run` → 384 passed / 2824 tests; the 12 failures are the pre-existing missing `terminal-webview-engine.generated` postinstall artifact (identical without this branch). `mobile$ tsc --noEmit` clean except that same generated file; `oxlint` and `oxfmt --check` clean; no `max-lines` disable added. Each fix was proven by reverting only its source file (`git checkout --` + `git apply`, never `git stash`): dropping the cutover retry fails 4 tests, dropping the age bound fails 2, re-gating the card fails 1. Root `config/tsconfig.node.json` and `config/tsconfig.web.json` both clean (web has only the pre-existing TS6307 noise); nothing outside `mobile/` changed.

OrcaWin:
Both round-2 findings fixed, plus the red `static analysis` check.

**#1 — age bound anchored on proof time (`home-worktree-info.ts`)**
Confirmed, both halves. Probe against the previous HEAD: `countsProvenAt = now-11min` + a failed refresh → `Worktree list unavailable` (the pre-fix string this PR exists to remove), while `countsProvenAt = now-3d` with the refresh merely pending → `12 worktrees · 2 active`, asserted as live.

Root cause was where the bound was applied, not when it was measured from: age gated `Last known` → `unavailable` inside the failure branch only. It now gates live → `Last known` on every path, so the three strings mean one thing each:

- proven and confirmed inside the window → `12 worktrees · 2 active`
- proven but not currently confirmed — failed refresh **or** a rehydrated snapshot older than the window → `Last known: 12 worktrees · 2 active`
- never proven → `Worktree list unavailable`

A failed refresh therefore keeps whatever the host last proved however old it is (finding #1a), and aged counts stop rendering as live even when nothing has failed yet (finding #1b, which the previous bound never covered). `hasFreshEnoughCounts` and its `!staleCounts → false` inversion are gone; the branch is `catalogUnavailable && !staleCounts`, which is exactly "never had counts". Net −2 lines. The round-1 concern is still satisfied: day-old counts remain visibly distinct from seconds-old ones, just as history instead of as nothing.

Two existing tests changed expectation because they encoded the reported bug (`provenAt + MAX + 1` after a failure asserted `Worktree list unavailable`); no test was deleted, skipped or loosened, and the suite gained the pending-refresh case. Reverting only `home-worktree-info.ts` fails 3 of them.

**#2 — `statsByHost` never pruned (`home-stats-total.ts`)**
Confirmed: no `delete` anywhere, and `handleRemove` is not the only path — `app/h/[hostId]/index.tsx:625` removes hosts too, after which Home reloads `hosts` and drops the card but keeps summing the entry. Fixed at the derivation instead of at one call site: `totalHomeStats(byHost, hostIds)` sums the hosts still paired, so every removal path self-corrects and the map needs no mutation. Reverting only `home-stats-total.ts` fails `drops a removed desktop from the total`.

**Also: `static analysis` was failing on this PR** — `react-doctor(effect-needs-cleanup)` on `mobile/app/index.tsx`, whose label spans the whole per-host effect, so the response commit’s added lines inside it turned a pre-existing false positive into a gated one. The effect does own every handle (the closure pushed onto `cleanups` releases all three). An inline `oxlint-disable-next-line` does not work here: the changed-code gate’s first scan runs without the react-doctor plugin and reports the directive as unused. Followed the existing precedent in `useDashboardPopoutBridge.ts` instead — the effect body moved verbatim into `wireHostSubscriptions`, which returns the disposer. Zero behavior change.

**Verification**
`mobile$ npx vitest run` → 384 files / 2826 passed; the 12 failures are the same pre-existing missing `terminal-webview-engine.generated` artifact. The PR’s original tests all pass, including `keeps rendering the last proven counts after a failed refresh` and `fetches again after a dropped socket comes back` — the original defect is still fixed. `mobile$ tsc --noEmit` clean apart from that artifact; both root tsconfigs clean (web has only the pre-existing TS6307 noise); `oxlint` clean; max-lines ratchet OK, no suppression added; `node config/scripts/check-changed-code-quality.mjs` now **passes**.

PR #12416, opened 2026-08-04
perf(worktrees): delete worktree directories after the removal returns

brennanb2025:
## Summary

Deleting a workspace froze on "Deleting…" for 8–35 s. `git worktree remove` deletes the whole checkout — usually a multi-GB `node_modules` — synchronously inside the remove IPC, and Orca holds the file-watcher removal gate and PTY teardown for that entire window (prod traces: `worktree.remove.git_remove` p50 8–14 s, p90 29 s, max 34.7 s over 21 removals; the worst single `git.exec` was 31.6 s).

Local worktree removal now does the standard two-phase delete:

1. Every existing preflight runs unchanged (lock check, clean check, PTY sweep, watcher gate, linked-path/symlink removal).
2. The checkout is renamed into a hidden sibling trash root (`<worktree parent>/.orca-worktree-trash/wt-<epoch>-<nonce>`) — a metadata operation on the same volume.
3. Git's registration for the now-missing path is cleared with `git worktree remove --force <path>` (falls back to `git worktree prune` + a strict registration re-check if that ever fails).
4. The moved tree is deleted in the background, one at a time, after the IPC has returned.
5. Leftovers from a crash or a quit mid-delete are swept at startup, scoped to the trash roots under known local workspace roots and only for entries matching the generated `wt-<digits>-<8 hex>` name.

Anything that cannot be renamed falls back to the previous in-place removal with no behavior change: WSL-hosted checkouts (deleted inside the distro), cross-volume paths (`EXDEV`), Windows open handles (`EPERM`/`EBUSY`), and dirty checkouts. `recoverLocalWindowsWorktreeRemoval` is untouched and still covers the Windows partial-removal path. The SSH/remote provider path and folder workspaces are untouched.

Return payloads, branch deletion (`worktree.remove.branch_delete`), and all error messages are unchanged. One new span stage, `worktree.remove.trash_rename`, was added to `WorktreeRemoveStage`.

### Measured before/after

Fixture: scratch repo + worktree holding a committed 100 000-file tree (200 dirs × 500 files), generated programmatically by `src/main/git/worktree-removal-large-tree.bench.test.ts` (opt-in via `ORCA_WORKTREE_REMOVAL_BENCH=1`). Baseline measured on the parent commit with the same fixture and harness. macOS/APFS.

| | user-visible `removeWorktree` | background deletion |
| --- | --- | --- |
| Before (parent commit) | 6755 ms / 6419 ms | — (inline) |
| After | 272 ms / 256 ms / 160 ms | 3388 ms / 4019 ms / 2287 ms, after the call returned |

~25–40× faster for the user; the recursive delete demonstrably completes afterwards (the bench asserts the trash root is empty once the queue settles). Startup sweep: `sweepStaleWorktreeTrash` is verified against a simulated leftover trash directory in both `src/main/worktree-trash.test.ts` and the real-Git suite.

The ~160–270 ms that remains is the cleanliness probe: `git worktree remove` re-checks the worktree at removal time, so the rename path re-proves the same thing with `git status --porcelain --untracked-files=all` before moving anything. That check is what Git itself was already running, so it is not new work on the fast path — but on a fallback removal (WSL excepted, which returns before the probe) it is now run twice. That was the deliberate trade: dropping it would let a file written between the IPC preflight and the removal be trashed silently.

## Screenshots

No visual change.

## Testing

- [x] `npx oxlint` on all changed files (clean), plus the pre-commit oxlint + react-doctor + oxfmt gates
- [x] `pnpm run tc`
- [x] `pnpm exec vitest run --config config/vitest.config.ts src/main/git/ src/main/worktree-trash.test.ts src/main/ipc/worktrees.test.ts src/main/local-worktree-removal-recovery.test.ts src/main/local-worktree-filesystem.test.ts` — 964 passed
- [x] Real-Git compatibility matrix: `src/shared/git-binary-compatibility.test.ts` run against Git 2.38.1 and 2.49.1 images locally (CI also runs 2.25.5); the new case proves `worktree remove --force` deregisters a renamed-away worktree. Manually verified the same on 2.24.4 and 2.26.3 containers, which bracket the 2.25 baseline.
- [x] `ORCA_WORKTREE_REMOVAL_BENCH=1` benchmark, before and after (numbers above)
- [ ] `pnpm build` — not run (no bundling-relevant change)
- [x] New tests, and verified they fail on the parent commit where behavior differs

New/updated coverage:

- `src/main/git/worktree-deferred-removal-real-git.test.ts` (real Git binary): registration cleared and `git worktree list` no longer shows the path; sibling worktrees stay registered; branch deletion identical; a dirty checkout is never renamed aside; startup sweep removes leftover trash. **The first case fails on the parent commit** (no trash root is ever created there) — verified by stashing the source change and re-running.
- `src/main/git/remove-worktree.test.ts`: rename-then-deregister taken on success; `prune` fallback when deregistration fails; restore + in-place removal when the registration survives pruning; in-place removal when the rename is unavailable (the Windows `EPERM` / cross-volume shape); dirty checkout never renamed; WSL never renamed.
- `src/main/worktree-trash.test.ts`: rename/restore semantics, background deletion, sweep removes only generated entry names (a stray directory, a stray file, live worktrees and deeper trash roots all survive), and sweep-root derivation skips SSH repos and folder workspaces.
- Full `src/main` suite: the only failures are 13 pre-existing local false reds in `pty.test.ts` / `updater.test.ts` / `worktree-base-directory-poller.test.ts` — confirmed identical on `origin/main` with the same command.

`src/main/ipc/worktrees.test.ts` mocks `removeWorktree` wholesale, so it needed no new cases: the change lives entirely below that seam, and its existing 200+ cases (including the removal span-stage assertions) pass unchanged.

## AI Review Report

Self-review focused on: safety-gate preservation, Git-version compatibility, cross-platform path/rename semantics, and blast radius of the sweep.

- **Cleanliness gate.** The rename must not bypass the check Git performs at removal time, so the fast path re-proves cleanliness and declines (falling through to the untouched in-place path) if that check throws for any reason. That keeps orphan-worktree handling working: a corrupt/missing checkout still errors out of `git worktree remove` exactly as before, and the IPC's `isOrphanedWorktreeError` recovery still fires.
- **Git compatibility.** `git worktree remove --force` on a missing directory was verified to succeed and to touch only the targeted entry on 2.24.4, 2.26.3, 2.38.1 and 2.49.1 — it works at and below the 2.25 baseline, so no `GitCapabilityCache` capability was added; a plain try/`prune` fallback with a strict registration re-check covers any binary that behaves differently. The boundary is now asserted in the real-binary CI contract.
- **Cross-platform.** All paths use `path.join`/`dirname`; the trash root is always the worktree's own parent, so the rename never crosses a volume by construction. Windows `EPERM`/`EBUSY` from open handles and `EXDEV` both resolve to "deferral unavailable" → previous behavior; background deletion reuses `removeHostTree`, which already carries the Windows transient-retry contract. WSL returns before the probe so nothing renames a distro-owned path from the Windows side. No shortcuts, labels, or UI touched.
- **Failure atomicity.** If the registration cannot be cleared, the directory is renamed back and the original in-place removal runs, so a partial failure leaves the worktree exactly as Git registered it. If the restore itself fails, the underlying error is thrown rather than a success being reported.
- **Watcher interaction.** The base-directory poller only reacts to `<worktree>/.git` creation and `<worktree>` deletion, so the trash root (kept in place, never deleted) produces no structural rescans, and the background deletion happens under a name no watcher or PTY can be spawned into.

## Security Audit

- **Path handling.** No user input reaches the trash path: the root is derived from the worktree's parent directory and the entry name is `wt-<Date.now()>-<4 random bytes hex>`. Nothing is interpolated into a shell — Git is invoked through the existing `execFile`-based runner with argument arrays.
- **Deletion blast radius.** The background delete only ever targets a path this process just created by rename. The startup sweep is bounded three ways: only the trash roots at, or one level under, workspace roots of local git repos; only entries matching `^wt-\d+-[0-9a-f]{8}$`; at most 200 repo containers per root. Stray files/directories inside a trash root are left alone (asserted in tests). `fs.rm` does not traverse symlinks, so a planted symlink inside a trash root cannot redirect the delete.
- **Privilege/secrets.** No new IPC surface, no new permissions, no credentials or secrets touched; warnings log paths only, matching the surrounding removal logging.
- **Follow-up.** None identified.

## Notes

- A hidden `.orca-worktree-trash` directory now persists (empty) beside worktrees after the first deletion in a repo. Removing it on drain would fire a spurious structural rescan in the workspace-root poller, which is exactly the churn this PR reduces, so it is deliberately left in place.
- The trashed tree occupies disk for a few seconds longer than before, since deletion no longer completes before the IPC returns.
- Windows and WSL were reasoned about and covered by the fallback path and unit tests, but not live-tested on this change; the fallback is the unmodified pre-existing code path.
- The benchmark file is opt-in (`ORCA_WORKTREE_REMOVAL_BENCH=1`) and skipped in CI; it exists so the before/after numbers are reproducible.


brennanb2025:
Electron QA evidence — macOS, CDP, `df09d69e7c`

- Created a disposable local repo/worktree through the running dev app.
- Confirmed deletion through the real **Delete Workspace** dialog.
- Verified the row disappeared, Git no longer listed the worktree, the checkout path was removed, and the deferred trash queue drained to an empty root.
- 100,000-file benchmark: `removeWorktree` returned in **420 ms**; recursive deletion finished in **4,264 ms** afterward.

Before deletion



Confirmation



After deletion

PR #12421, opened 2026-08-04
perf(renderer): index detectedWorktreesByRepo so SSH panes and cards stop walking the full detected catalog on every store write

brennanb2025:
## Summary

SSH-owned terminal panes and sidebar worktree cards resolve their runtime owner inside retained Zustand selectors, so `getExplicitRuntimeEnvironmentIdForWorktree` runs once **per mounted pane and per rendered card, on every store write**. That function opened with an uncached `Object.values(state.detectedWorktreesByRepo ?? {}).some(r => r.worktrees.some(...))` — a full scan of every detected worktree of every repo. For a user with many remote worktrees this is the single largest multiplier on the store-write path: 30 SSH cards × a 2,000-entry detected catalog is 60,000 string comparisons per unrelated store write.

`worktree-runtime-owner-index.ts` already maintains WeakMap-keyed, immutable-slice indexes for `worktreesByRepo`, `repos`, `folderWorkspaces`, and `projectGroups`. `detectedWorktreesByRepo` was the only major collection without one. This PR adds it and converts the walk sites.

**New index** (`worktree-runtime-owner-index.ts`), following the existing idiom exactly:
- `findIndexedDetectedWorktrees(detectedWorktreesByRepo, worktreeId)` → every detected publication of that id, in catalog order.
- `hasIndexedDetectedWorktree(...)` → the existence predicate the three `hasDetectedOwner` sites need.

A **list** index rather than a resolved-owner index is deliberate: rival repos can publish the same worktree id, and `resolveExplicitWorktreeOperationRouteResult` must see *all* of them to fail closed on a conflict. Collapsing to one owner would change routing behavior.

**Converted call sites** (all behavior-preserving):
- `worktree-runtime-owner.ts` — the three `hasDetectedOwner` walks in `getRuntimeEnvironmentIdForWorktree`, `getExplicitRuntimeEnvironmentIdForWorktree`, and `getExecutionHostIdForWorktree`.
- `worktree-operation-route.ts` — the detected half of `ownerRecordsOnHost`, the detected half of `hasKnownWorktree`, and the detected-publication loop in `resolveExplicitWorktreeOperationRouteResult`.
- `worktree-operation-route.ts` `hasKnownWorktree` also stops walking `worktreesByRepo`: it now asks the **already-built** worktree owner index (`resolveIndexedWorktreeOwner(...).kind !== 'missing'`), which that code path populates a few lines earlier. No new index, exact same answer.

Correctness rests on the detected catalog being replaced rather than mutated. Verified: every store write goes through immutable replacement, and `applyDetectedWorktreeUpdates` (`store/slices/worktrees.ts:813`) explicitly returns the *same* object identity when nothing changed — which is precisely what keeps the index warm across unrelated writes.

No user-visible behavior change.

## Screenshots

No visual change.

## Measurement

Micro-bench: 20 repos × 100 detected worktrees (2,000 entries), 200 lookups per sweep, 40 sweeps, 5 warm-up sweeps, each leg given its own never-indexed catalog pool so no leg silently inherits a warm index. Committed as `worktree-runtime-owner-index.detected-perf.test.ts`; it writes its report to `$TMPDIR/orca-detected-worktree-index-bench.json`. Numbers below are the median of 3 isolated local runs (Apple silicon, Node 26).

**Lookup path** (ms per 200-call sweep):

| Scenario | Before (walk) | After (index) | Speedup |
| --- | --- | --- | --- |
| Catalog identity stable across sweeps | 1.14 | **0.0088** | **~130x** |
| Catalog republished every sweep | 1.12 | 0.35 | ~3.2x |

**End-to-end `getExplicitRuntimeEnvironmentIdForWorktree`** (ms per 200-call sweep). Measured as a true A/B: the identical bench was run against the pre-change source (`git stash` of the three source files) and then against this branch, 3 runs each.

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| Detected catalog identity stable | 0.516 | **0.093** | **~5.5x** |
| Detected catalog republished every sweep | 0.518 | 0.496 | ~1.0x (parity) |

Absolute cost after warm-up is **0.0088 ms** on the lookup path and **0.093 ms** end-to-end per 200-call sweep, both well inside the 0.5 ms budget.

**Honest caveat on the republish case.** Building the 2,000-entry index costs roughly what one 200-call walk cost, so a sweep that *starts* with a freshly republished catalog is a wash, not a win — the target of ">=10x" holds on the warm path (~130x) but not on that leg, and I did not want to report a number the change does not deliver. Break-even is ~75–170 lookups per catalog identity. That is comfortably met in practice because a detected-worktree scan republishes the catalog only when its contents actually change (seconds apart), while the store takes many writes per second and each write re-runs the selector for every SSH pane and card. The perf test records both timing legs as informational benchmark evidence without using wall-clock results as a CI gate.

The non-timing CI guarantee is deterministic: the detected records are instrumented with a counting `id` getter, and the test asserts 200 warm `getExplicitRuntimeEnvironmentIdForWorktree` owner lookups perform **0** detected-record reads while the old walk performs **>100,000**. That covers the actual retained-selector hot path and cannot flake under load.

## Testing

- [x] `pnpm lint` — `npx oxlint src/renderer/src/lib/` clean; changed-code quality gate run explicitly: `node config/scripts/check-changed-code-quality.mjs -- a6b14eb04c9a` → `code quality: 0`, `type-aware code quality: 0`, `React Doctor: 0`, across **5 changed files** (non-vacuous — matches the diff).
- [x] `pnpm typecheck` — `typecheck:tsc:web`, `typecheck:tsc:node`, `typecheck:tsc:cli` all clean (`--composite false`, so no stale-tsbuildinfo pass).
- [x] `pnpm test` — targeted: `src/renderer/src/lib/` 372 files / 3,535 passed; `src/renderer/src/store` + `components/terminal-pane` + `components/sidebar` 589 files / 7,131 passed. `oxfmt --check` clean.
- [ ] `pnpm build` — not run; renderer-only pure-function change with typecheck green across all three projects.

New tests in `worktree-runtime-owner-index.detected.test.ts`:
- **Randomized parity, 240 cases × 5 probes each (1,200 assertions per predicate)** against the pre-index walks kept verbatim in the test as oracles. Seeded LCG so any failure reproduces from the printed case index. Covers: id present in first repo / last repo / absent entirely, `undefined` catalog, empty catalog, repos publishing empty worktree arrays, and ids duplicated across repos. Record shapes mirror the store types (`id`, `repoId`, optional `hostId` across local/ssh/runtime, optional `runtimeOwnerEnvironmentId`). The predicate matches on **`id`**, not path — the tests pin that.
- **Ordering and identity**: rival publications come back in catalog order, and the returned records are the same object references (`toBe`), which is what the downstream ambiguity checks depend on.
- **Invalidation**: a new catalog identity with changed content is re-indexed, including the remove-then-re-add sequence, and each prior identity keeps its own correct answer — no bleed between snapshots.

## AI Review Report

I reviewed the diff myself rather than delegating, focusing on the ways an index can silently diverge from the walk it replaces:

- **Staleness / cache invalidation** — the top risk. Audited every write to `detectedWorktreesByRepo` in `store/slices/worktrees.ts` and `store/slices/repos.ts`: all are immutable replacements (`{...s.detectedWorktreesByRepo}`, `Object.fromEntries`, fresh `nextByRepo`). No in-place mutation, so a content change always yields a new identity and a rebuilt index. Pinned by test.
- **Iteration order** — `Object.values` insertion order is preserved because repo ids are non-integer-like strings; the pre-change code used the same `Object.values`, so ordering is parity by construction. This matters because `resolveExplicitWorktreeOperationRouteResult` early-returns on the first ambiguous match, and `ownerRecordsOnHost` must keep published-before-detected order. Both verified in the diff and covered by the ordering test.
- **Over-collapsing duplicates** — rejected a resolved-owner index in favor of a list index precisely so conflicting publications still reach the fail-closed paths.
- **`hasKnownWorktree` rewrite** — confirmed `resolveIndexedWorktreeOwner(...).kind !== 'missing'` is exactly `.some(w => w.id === id)`: the index sets an entry for every id (resolved or ambiguous) and returns `missing` only when absent. Included in the randomized parity test rather than argued.
- **Empty-array aliasing** — `NO_DETECTED_WORKTREES` is a shared constant returned on miss; it is `readonly`-typed and no caller mutates it.
- **Memory** — WeakMap-keyed, so an index is collected with its catalog; no unbounded growth.
- **Cross-platform (macOS / Linux / Windows)** — this change is pure in-memory data-structure work in the renderer. It touches no paths, no shell or command execution, no keyboard shortcuts or accelerators, no shortcut labels, no `path` joining, no Electron main/renderer platform branches, and no Git invocation. There is nothing platform-dependent in the diff. The one platform-adjacent detail is the perf test's report file, which uses `os.tmpdir()` + `path.join` rather than a hardcoded `/tmp`, so it is correct on Windows too. Applies equally to SSH-backed and folder workspaces: folder workspaces short-circuit on the `parseWorkspaceKey` branch before these lookups and are unaffected, and SSH-owned worktrees are the primary beneficiary.

What I changed as a result of review: kept wall-clock measurements report-only after repeated runs showed shared-runner variance, and retained the deterministic record-read count as the CI gate.

## Security Audit

Low risk. No input parsing, command execution, path handling, authentication, secrets, IPC, or dependency changes. The diff adds an in-memory lookup table over data the renderer already holds and reads; no new data crosses a trust boundary and no new surface is exposed. The one security-adjacent property worth stating is that the change is **fail-closed preserving**: `resolveExplicitWorktreeOperationRouteResult` and `getExecutionHostIdForWorktree` deliberately refuse to resolve (→ `ambiguous` / `runtime:unresolved-owner`) when rival hosts publish the same worktree id, which is what stops a remote-owned worktree from being routed to a local PTY. Because the index preserves every duplicate publication and its order, those conflict paths see exactly the same inputs as before — verified by the ordering test and by the randomized parity oracle. No follow-up needed.

## Notes

- Deliberately left out of scope, all still doing full walks, each a separate concern from indexing the detected catalog:
  - `worktree-operation-route.ts` `ownerRecordsOnHost` still walks `worktreesByRepo` (its detected half is now indexed). The existing `findIndexedWorktreeOwnerForHost` is **not** a drop-in: it collapses duplicates and returns `null` on ambiguity, whereas this caller needs every match. Indexing it properly needs a second list index — worth doing, but not in a PR about detected worktrees.
  - `editor-file-operation-owner.ts:123` `isWorktreePublished` runs the identical predicate over both catalogs. It sits on user-initiated file operations, not the store-write path, so it is not part of this hot path; it is now a one-line conversion whenever someone wants it.
  - `WorktreeCard.tsx:377-382` does a linear `s.runtimeEnvironments.find(...)` per card per store write. Real, and flagged in the task, but `runtimeEnvironments` is a small catalog and it lives in a different module family from the worktree owner indexes; folding it in here would have made this two concerns. Left as a follow-up.
  - `runtime-session-mirror-owners.ts:37` and `direct-ssh-target-scope.ts:70` iterate the whole detected catalog rather than looking up one id, so an id-keyed index does not apply.
- The perf test's wall-clock legs are report-only: the numbers land in `$TMPDIR/orca-detected-worktree-index-bench.json` on every run, while deterministic record-read counts provide the CI gate.

### DIFF
--- a/src/renderer/src/lib/worktree-operation-route.ts
+++ b/src/renderer/src/lib/worktree-operation-route.ts
@@ -7,7 +7,11 @@ import {
 import { parseWorkspaceKey } from '../../../shared/workspace-scope'
 import { getRepoIdFromWorktreeId } from '@/store/slices/worktree-helpers'
 import { addRoute, resolveExactWorktreeRoute, routeForOwner } from './worktree-owner-route'
-import { resolveIndexedWorktreeOwner } from './worktree-runtime-owner-index'
+import {
+  findIndexedDetectedWorktrees,
+  hasIndexedDetectedWorktree,
+  resolveIndexedWorktreeOwner
+} from './worktree-runtime-owner-index'
 import {
   findFolderWorkspaceOwner,
   getExecutionHostIdForFolderWorkspace,
@@ -51,11 +55,7 @@ function ownerRecordsOnHost(
   executionHostId: ExecutionHostId
 ): WorktreeOperationOwnerRecord[] {
   const owners: WorktreeOperationOwnerRecord[] = []
-  const catalogs = [
-    ...Object.values(state.worktreesByRepo ?? {}),
-    ...Object.values(state.detectedWorktreesByRepo ?? {}).map((result) => result.worktrees)
-  ]
-  for (const worktrees of catalogs) {
+  for (const worktrees of Object.values(state.worktreesByRepo ?? {})) {
     for (const worktree of worktrees) {
       if (
         worktree.id === worktreeId &&
@@ -63,6 +63,11 @@ function ownerRecordsOnHost(
       ) {
         owners.push(worktree)
       }
+    }
+  }
+  for (const worktree of findIndexedDetectedWorktrees(state.detectedWorktreesByRepo, worktreeId)) {
+    if (parseExecutionHostId(worktree.hostId)?.id === executionHostId) {
+      owners.push(worktree)
     }
   }
   return owners
@@ -135,12 +140,8 @@ export function resolveWorktreeOperationRouteResult(
   }
 
   const hasKnownWorktree =
-    Object.values(state.worktreesByRepo ?? {}).some((worktrees) =>
-      worktrees.some((worktree) => worktree.id === worktreeId)
-    ) ||
-    Object.values(state.detectedWorktreesByRepo ?? {}).some((result) =>
-      result.worktrees.some((worktree) => worktree.id === worktreeId)
-    )
+    resolveIndexedWorktreeOwner(state.worktreesByRepo, worktreeId).kind !== 'missing' ||
+    hasIndexedDetectedWorktree(state.detectedWorktreesByRepo, worktreeId)
   const repoId = getRepoIdFromWorktreeId(worktreeId)
   const hasKnownRepo = state.repos?.some((repo) => repo.id === repoId) === true
   if (!hasKnownWorktree && !hasKnownRepo) {
@@ -216,18 +217,14 @@ export function resolveExplicitWorktreeOperationRouteResult(
       addRoute(exactRoutes, resolution.route)
     }
   }
-  for (const result of Object.values(state.detectedWorktreesByRepo ?? {})) {
-    for (const worktree of result.worktrees) {
-      if (worktree.id === worktreeId) {
-        exactRepoIds.add(worktree.repoId)
-        const resolution = resolveExactWorktreeRoute(state, worktree)
-        if (resolution.kind === 'ambiguous') {
-          return resolution
-        }
-        if (resolution.kind === 'resolved') {
-          addRoute(exactRoutes, resolution.route)
-        }
-      }
+  for (const worktree of findIndexedDetectedWorktrees(state.detectedWorktreesByRepo, worktreeId)) {
+    exactRepoIds.add(worktree.repoId)
+    const resolution = resolveExactWorktreeRoute(state, worktree)
+    if (resolution.kind === 'ambiguous') {
+      return resolution
+    }
+    if (resolution.kind === 'resolved') {
+      addRoute(exactRoutes, resolution.route)
     }
   }
   if (exactRoutes.size > 0) {
--- a/src/renderer/src/lib/worktree-runtime-owner-index.detected-perf.test.ts
+++ b/src/renderer/src/lib/worktree-runtime-owner-index.detected-perf.test.ts
@@ -129,7 +129,7 @@ describe('detected worktree index performance', () => {
     expect(walkedReads).toBeGreaterThan(100_000)
   })
 
-  it('keeps the lookup path an order of magnitude cheaper than the catalog walk', () => {
+  it('records lookup-path timing without making wall clock a CI gate', () => {
     const walkPool = freshCatalogPool()
     const walkMs = measureSweeps((sweep) => {
       const catalog = walkPool[sweep]!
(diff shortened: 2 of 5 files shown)


brennanb2025:
Electron QA on commit 7a5f21c30a:

- Launched the current worktree dev app with Electron CDP.
- Switched to the demo-project/main worktree and opened Workspace board.
- Verified the worktree sidebar and board rendered in the large detected-worktree context (three projects each showing 370 hidden discovered worktrees).
- No renderer page errors were reported.
- No visual change is expected from this in-memory lookup optimization.

Electron QA: workspace board and detected-worktree sidebar

PR #12426, opened 2026-08-04
fix(mobile): stop unrenewed-grace rotation churn and close gate cadence gaps

OrcaWin:
Follow-ups documented on #12374 (all three were review-round residuals deliberately left out of the incident PR so the phone-verified binary matched the merge commit):

1. **Unrenewed-grace rotation churn.** When the cell answers a resume with `renewed: false` (return-unchanged-grace), the returned deadline never advances, so the phone force-rotated once per clamp floor — ~60 session replacements/hour, each killing any in-flight RPC — until a fresh credential arrived. A re-resume with the same grace credential provably returns the same decision, so rotation is now skipped entirely when `confirmation?.renewed === false`. Fresh credentials still arrive via direct rotation or durable bundle reads, exactly as before.

2. **Gate cadence doubling under external-signal gates.** An E2EE (`external-signal`) gate never records rejected credential versions, so `armCredentialReprobe` with no dialable bundle took the plain-cooldown branch and its tick carried no gate pass token — every second wakeup bounced off `shouldDefer`, settling the effective attempt interval at ~30min instead of the intended 15min ceiling. Under a held gate the reprobe now mints the token.

3. **Gate timers armed against `scheduleRetry=false`.** `registerFailure`'s gate branches armed a reprobe tick even when the supervisor was backgrounded/stopped. Measured-benign but wrong lifecycle; the gate branches now honor `scheduleRetry`, and a test proves foreground resume restores the cadence (no permanent stall).

Also extracts `RelayRetryDelays` (all jittered retry delay policy) and the supervisor test fakes into their own modules to satisfy max-lines without weakening the cap.

## Tests

Three new red-before/green-after tests (verified failing on pre-fix sources):
- supervisor: `skips forced rotation for a session resumed without renewal` (red: 2nd dial at the 60s clamp floor)
- controller: `mints the gate pass when arming a no-bundle reprobe under a held gate` (red: `shouldDefer` true after the 120s tick)
- controller: `does not arm a gate reprobe timer when the supervisor declined retries` (red: timer armed after each gated failure)

Full mobile suite: 2969 passed (1 pre-existing unrelated local failure: missing `mermaid-webview-engine.generated` codegen artifact).

### DIFF
--- a/mobile/src/transport/mobile-endpoint-supervisor-test-fakes.ts
+++ b/mobile/src/transport/mobile-endpoint-supervisor-test-fakes.ts
@@ -1,0 +1,169 @@
+import { vi } from 'vitest'
+import type { MobileRelayCredentialBundle } from './mobile-relay-credential-bundle'
+import type { MobileRelayRpcSession } from './mobile-relay-rpc-session'
+import type { MobileEndpointSupervisorDependencies } from './mobile-endpoint-supervisor'
+import type { RpcClient } from './rpc-client'
+import type { MobileConnectionPath, StableLogicalRpcClient } from './stable-logical-rpc-client'
+import type { ConnectionState, HostProfile, RpcResponse } from './types'
+
+export class FakeSession implements RpcClient {
+  readonly sendRequest = vi.fn(
+    async (_method: string, _params?: unknown): Promise<RpcResponse> => ({
+      id: 'rpc-1',
+      ok: true,
+      result: {},
+      _meta: { runtimeId: 'runtime-1' }
+    })
+  )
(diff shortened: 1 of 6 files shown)


Jinwoo-H:
Merging on green CI + red-before/green-after tests + a line-by-line self-review of the final diff. Note for the record: an external adversarial review pass was spawned but did not return; the three behavior changes here implement remedies that were themselves specified during PR #12374's review rounds (grace-path rotation skip, gate-token minting, scheduleRetry honoring), so the design has had adversarial eyes even though this diff has not.

PR #12435, opened 2026-08-04
fix(runtime): isolate same-path folder workspace terminals

dgk-dev:
## Summary

Folder projects intentionally support multiple independent workspace sessions backed by the same directory. Their IDs differ only by the synthetic `::workspace:<uuid>` suffix.

Runtime PTY ownership comparisons used `splitWorktreeIdForFilesystem`, which deliberately strips that suffix for cwd/filesystem operations. As a result, two distinct folder workspace instances compared equal. During controller inventory refresh, an exact target could select or rebind a sibling workspace's PTY; paired/mobile clients could then receive a stale or perpetually pending terminal handle and remain on **Loading terminal**.

This change keeps the suffix in runtime identity comparison and identity-map keys by using `splitWorktreeId`. Filesystem callers continue using `splitWorktreeIdForFilesystem`, so both sessions still launch in the shared physical directory.

## User-visible result

Multiple workspaces under one Folder project can keep separate terminal graphs even though they intentionally share a directory. Opening one workspace from Mobile no longer allows another same-path workspace's PTY to satisfy its terminal refresh.

## Regression coverage

The regression creates two live controller PTYs with the same folder path and different `::workspace:<uuid>` owners, requests the second workspace by exact ID, and proves only its own PTY is selected and recorded. The pre-fix comparison admits both because the UUID suffix is stripped.

## Validation

- `src/main/runtime/orca-runtime.test.ts`: 1,022 passed, 1 existing skip
- Node/CLI/Web TypeScript projects passed
- Focused Oxlint passed
- Focused formatter check passed
- `git diff --check` passed

## Relationship to #11868

This is complementary and substantially narrower. #11868 restores and reattaches persisted folder session snapshots across runtime generations and adds deletion/routing fences. This PR fixes the lower-level synthetic folder-project instance identity (`repo::path::workspace:<uuid>`) that currently collapses before those snapshot/materialization decisions. It does not duplicate the reattachment or tombstone work.

### DIFF
--- a/src/main/runtime/orca-runtime.test.ts
+++ b/src/main/runtime/orca-runtime.test.ts
@@ -3502,6 +3502,50 @@ describe('OrcaRuntimeService', () => {
     expect(terminals.terminals[0]?.worktreePath).toBe(TEST_FOLDER_WORKSPACE_PATH)
   })
 
+  it('keeps same-path folder workspace instance PTYs scoped to their exact ids', async () => {
+    const firstWorktreeId = `${TEST_REPO_ID}::${TEST_FOLDER_WORKSPACE_PATH}${FOLDER_WORKSPACE_INSTANCE_SEPARATOR}11111111-1111-4111-8111-111111111111`
+    const secondWorktreeId = `${TEST_REPO_ID}::${TEST_FOLDER_WORKSPACE_PATH}${FOLDER_WORKSPACE_INSTANCE_SEPARATOR}22222222-2222-4222-8222-222222222222`
+    vi.mocked(listWorktrees).mockClear()
+    vi.mocked(listWorktrees).mockRejectedValue(
+      new Error('folder explicit-id fallback should not rescan worktrees')
+    )
+    const runtime = createRuntime()
+    runtime.setPtyController({
+      write: () => true,
+      kill: () => true,
+      getForegroundProcess: async () => null,
+      listProcesses: async () => [
+        {
+          id: 'first-folder-pty',
+          cwd: TEST_FOLDER_WORKSPACE_PATH,
+          title: 'first',
+          worktreeId: firstWorktreeId
+        },
+        {
+          id: 'second-folder-pty',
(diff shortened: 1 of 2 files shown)


nwparker:
Thanks @dgk-dev — independently reproduced and confirmed your diagnosis: `runtimeWorktreeIdsEqual`/`runtimeWorktreeIdentityKey` were parsing with `splitWorktreeIdForFilesystem`, so every folder workspace on one path (including the suffix-less folder root) compared equal.

I've opened #12474 which carries your regression test over essentially verbatim (it drives the public `listTerminals('id:<exact>')` path and asserts the sibling PTY is never recorded — a better vector than my private-method tests), plus your inline note on `runtimeWorktreeIdentityKey`. Credit for the root cause is yours.

It adds three follow-ons: collapsing the duplicated hand-rolled comparison in `stopTerminalsForWorktree` (the drift that caused this), a `targetWorktreeId` tie-break in `findResolvedWorktreeIdForPath` (without it, the stricter identity comparison silently drops unowned cwd-only PTYs from a folder workspace's inventory — a new regression the identity fix would otherwise introduce), and coverage for the persisted-surface / "Loading terminal" path and the folder root as a third distinct identity.

AmethystLiang:
Duplicate of #12474.

Same root cause and intended fix: isolate same-path folder workspace PTY identity so distinct `::workspace:<uuid>` sessions do not compare equal after filesystem id stripping. Please continue review/landing on #12474. If this branch has a meaningfully different implementation detail worth cherry-picking, call it out on #12474.

nwparker:
Landed as #12474 → `6ea99f66075` on `main`, with your regression test carried over essentially verbatim and your inline note on `runtimeWorktreeIdentityKey`. Credited as co-author on the squash commit. Thanks for finding and diagnosing this one, @dgk-dev — the root cause (`splitWorktreeIdForFilesystem` used for identity) was yours.

PR #12436, opened 2026-08-04
fix(xterm): flush queued application input after IME commit

nwparker:
## Summary

- flush application input queued before a native `insertText` commit
- release application input queued during xterm's late-duplicate window
- delete dead `_dataAlreadySent` bookkeeping

This is separated from #12341 so the mobile native-range work and xterm composition ordering remain independently reviewable. It adds no Orca IME class, public xterm API, tracker, timer, locale rule, or production file.

## Why the xterm seam remains

VS Code, Tabby, Hyper, Electerm, and Wave delegate terminal composition to stock xterm, but several retain open Korean/Chinese failures. Stock beta.287 and beta.292 fail Orca's recorded rapid Korean and empty-end IBus traces. The retained change stays inside xterm's existing `CompositionHelper`, the owner that already orders native commits and `Terminal.input()`.

The IME source delta after this follow-up is 40 additions and 51 deletions. The unrelated existing `SortedList` patch remains unchanged.

## Evidence

- Native `insertText` emits exact event data before queued application input.
- Application input arriving during the one-task late-duplicate window flushes when that window releases.
- Reverting only those flushes strands `\x1b\r` in both positive traces while the other nine stock-xterm tests, including ordinary input, pass.
- Electron/PTTY Shift+Enter and Ctrl+Enter pass for both redispatch-before-keyup and keyup-before-redispatch ordering.

## Verification

- 3 focused Vitest files: 33/33 passed
- Electron/PTTY ordering E2E: 4/4 passed
- root typecheck and targeted oxlint passed
- `pnpm install --frozen-lockfile` passed
- `git diff --check` passed

## Scope

This PR is IME-only. It does not alter or slim unrelated xterm behavior, and it does not claim the full cross-platform IME goalpost is complete.


nwparker:
Keeping this one open — it is **not** superseded by #13128, unlike the rest of the IME stack.

I checked: the `_pendingInput` queue and the post-commit flush this adds to the xterm patch are absent from main. Main's patch has grown past this branch's (355 vs 276 lines) but in a different direction — it reworks the `_inputEvent` / `_keyDownSeen` gate rather than queueing application input across a composition. So this needs a rebase onto the new patch baseline; it will not apply as-is.

One thing worth re-testing rather than assuming, given what we just found: the flush here is `setTimeout(..., 0)`. We removed a 200 ms deferral from the terminal because it fired straight through the pending-composition gate and wrote a Shift+Enter payload to the PTY with no Enter pressed (10/10 reproducible on 1.4.171, landing 203-218 ms after keydown). A 0 ms flush is a much weaker claim than a 200 ms guess and may well be fine, but it is the same class of mechanism, so it deserves a check against a real IME rather than a synthetic event sequence — phrase-level Japanese and Chinese preedits are where the 200 ms version fell apart.

Happy to rebase it if you would rather I take it.

Issue #12106, opened 2026-08-02
i18n: languages with more than two plural forms cannot be translated


smwbev:
## Summary

Languages with more than two plural forms cannot be translated correctly today. Russian needs four (`one` / `few` / `many` / `other`); every plural site in the renderer can express two. This is not a catalog gap — the selection happens in JavaScript, so no translation can fix it.

Two independent causes, both measurable.

### 1. Call sites pick the form in JS, not through i18next

Every plural family in the renderer is resolved with a ternary:

```ts
// ComputerUsePane.tsx
: missingCount === 1
  ? translate('…permissionsRequired_one', '1 permission required…')
  : translate('…permissionsRequired_other', '{{value0}} permissions…', { value0: missingCount })
```

All 7 plural families follow this shape (`found_theme`, `import_theme`, `imported`, `over_limit`, `fullCoverage`, `permissionsRequired`, plus the one in `WarpThemeImportModal`). None passes `count`, so i18next's CLDR selection never runs and `_few` / `_many` are unreachable regardless of what a catalog provides.

This is worth flagging because [`config/i18n-translation-source.md`](https://github.com/stablyai/orca/blob/main/config/i18n-translation-source.md#L144) assumes the opposite:

> Plurals use Orca's existing per-key convention: each i18next plural-suffixed key (`…_one`, `…_other`) is its own PO entry. The profile must not use `msgid_plural`/`msgstr[n]`, which would collide with **i18next's own CLDR plural selection**.

The PO profile is designed around a selector that is currently bypassed. Worth reconciling before PR B lands, since the migration inherits whatever convention the call sites use.

### 2. Plugin language packs never reach CLDR rules

Even after a call site passes `count`, plugin catalogs still get English plural behaviour. They register under the synthetic `plugin<hex>` resource language from `pluginLanguageResourceId()`, which i18next cannot map to CLDR. Measured against a real catalog containing all four Russian forms:

| count | `ru` | `plugin<hex>` |
|---|---|---|
| 1 | 1 разрешение | 1 разрешение |
| 2 | 2 разрешения | 2 разрешения |
| 5 | 5 разрешен**ий** ✅ | 5 разрешен**ия** ❌ |

At `5` the synthetic language falls back to `_other` instead of `_many`.

## Why this matters beyond correctness

The current shape produced real defects in shipped catalogs. Before #11826, `permissionsRequired` used suffix interpolation, and the translations read:

```
[...]

PR #12439, opened 2026-08-04
fix(i18n): localize the status bar resource and host labels

smwbev:
## Summary

The status bar renders the Resource Manager tooltip and the remote-host count from bare literals inside helper functions, so both stay English while everything around them is translated.

`resource-manager-terminal-copy.ts` builds the whole tooltip without touching the catalog:

```ts
export function formatTerminalSessionCount(count: number): string {
  return `${count} terminal session${count === 1 ? '' : 's'}`
}
// …
lines.push('Terminal sessions are grouped by workspace.')
```

`SshStatusSegment.tsx` does the same for the collapsed segment:

```ts
function connectedHostCountLabel(count: number): string {
  return `${count} ${count === 1 ? 'host' : 'hosts'}`
}
```

**Why the coverage gate does not catch it.** `audit-localization-coverage.mjs` inspects JSX attributes and object properties, not values returned by helper functions, so `pnpm verify:localization-coverage` passes while the strings ship untranslated. This is the same class of defect as [#11826](https://github.com/stablyai/orca/pull/11826); that sweep counted 57 such helpers across the renderer, and this PR takes the status bar ones.

### What changed

| Area | Strings |
|---|---|
| Resource Manager tooltip | headline, memory-unavailable fallback, space-scan row, grouped/empty session rows |
| Resource Manager screen-reader label | segment name + session count |
| Remote hosts segment | host count (singular/plural) + `Connecting…` |

`sync:localization-catalog` added **11 keys**; nothing was removed or reworded (11948 → 11959, 0 deletions, 0 value changes). English wording is byte-identical.

### A latent bug this surfaces

The segment tinted one tooltip row by comparing it against the English text:

```tsx
className={line === 'Space scan ready' ? 'text-primary' : ''}
```

That comparison stops matching the moment the copy comes from the catalog, so simply routing the string through `translate()` would have dropped the tint in every non-English build. `getResourceManagerTooltipLines` now returns `{ text, emphasized }` and the segment reads the flag. A test pins it so the text-matching pattern cannot come back.

Notes on the approach:

- Keys are stable and semantic (`sessionsGroupedByWorkspace`, `hostCountOther`) rather than content hashes, matching `config/i18n-translation-source.md`: *"a feature developer adds a stable message ID and current English copy once"*.
- Plural forms use two explicit keys selected by a ternary, matching how the rest of the codebase selects plurals rather than introducing `count`.
- Host helpers moved to `ssh-status-copy.ts` so `SshStatusSegment.tsx` keeps its headroom under the 400-line gate — the same split accepted in #11826.

## Screenshots

**No visual change in English** — every string keeps its exact English wording, so an English build renders identically.

The defect is only visible in a localized build. From a Russian language pack on current `main`, before this PR — the tooltip and `0 hosts` sit in English next to a fully translated status bar:

Status bar

## Testing

- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`

`src/renderer/src/i18n/status-bar-copy-localization.test.ts` pins all 11 keys to their English text, so a helper that reverts to a bare literal fails the suite. `resource-manager-terminal-copy.test.ts` gains a case asserting that the space-scan row is flagged rather than recognized by its wording.

The full suite reports 44520 passed with 20 pre-existing failures across 8 files — `omp-shell-wrapper.node-pty`, `ssh-posix-command-wrapper`, `updater`, `terminal-ime-hangul-syllable-flush`, `terminal-ime-xterm-composition-cancel`, `automation-schedules`, `external-automation-jobs-file`, and `macos-computer-helper-owner-loss-processes`. All of them reproduce on a clean `upstream/main` checkout on this machine (macOS, Node 26), and none touch the status bar or i18n. The status-bar suite specifically is green: 38 files, 268 tests.

## AI Review Report

Reviewed the diff for correctness, localization behavior, and cross-platform impact on macOS, Linux, and Windows.

- **Flagged and fixed:** the `line === 'Space scan ready'` comparison described above. Routing the copy through the catalog without this change would have been a silent visual regression in every translated build — caught before pushing, and now covered by a test.
- **Cross-platform:** the change touches UI copy and its lookup path only. No paths, shell invocations, keyboard shortcuts, path separators, or Electron platform APIs are involved. `Connecting…` uses U+2026, already present in this file. The memory label (`3.69 GB · Σ RSS`) is produced upstream by `memoryMetricCopy` and is passed through unchanged as an interpolation value on all platforms.
- **Verified:** English output is unchanged, confirmed by the pre-existing assertions in `resource-manager-terminal-copy.test.ts`; the full status-bar suite (38 files, 268 tests) passes.

## Security Audit

- **Input handling:** `memoryLabel` originates from the daemon resource snapshot over IPC. It previously reached the DOM via template concatenation and now reaches it as an i18next interpolation value — the same data on the same path, no new source of untrusted input.
- **Escaping:** the renderer sets `interpolation.escapeValue: false`, but these strings render as React text nodes (`{line.text}`), so React escapes them. No `dangerouslySetInnerHTML`, no HTML assembly.
- **Command execution / paths / auth / secrets / dependencies:** untouched. No new IPC channels, no new dependencies, no filesystem or process access.

## Notes

Plural forms are selected with a ternary because that is the existing convention. This keeps two-form languages working and leaves languages with more plural categories approximated, which is the subject of [#12106](https://github.com/stablyai/orca/issues/12106) — no change to that is proposed here.

### DIFF
--- a/src/renderer/src/components/status-bar/ResourceUsageStatusSegment.tsx
+++ b/src/renderer/src/components/status-bar/ResourceUsageStatusSegment.tsx
@@ -1188,11 +1188,8 @@ export function ResourceUsageStatusSegment({
         <TooltipContent side="top" sideOffset={6}>
           <div className="space-y-0.5">
             {resourceManagerTooltipLines.map((line, index) => (
-              <div
-                key={`${index}:${line}`}
-                className={line === 'Space scan ready' ? 'text-primary' : ''}
-              >
-                {line}
+              <div key={`${index}:${line.text}`} className={line.emphasized ? 'text-primary' : ''}>
+                {line.text}
               </div>
             ))}
           </div>
--- a/src/renderer/src/components/status-bar/SshStatusSegment.tsx
+++ b/src/renderer/src/components/status-bar/SshStatusSegment.tsx
@@ -19,6 +19,7 @@ import {
 import { isUserManagedRuntimeEnvironment } from '../../../../shared/runtime-environments'
 import { RuntimeHostStatusRow, type RuntimeHostConnectionState } from './RuntimeHostStatusRow'
 import { SshTargetStatusRow } from './SshTargetStatusRow'
+import { connectedHostCountLabel, connectingHostsLabel } from './ssh-status-copy'
 import type { RemoteRuntimeSharedConnectionDiagnostics } from '../../../../shared/remote-runtime-shared-control-types'
 import { connectRuntimeEnvironmentAndRecordStatus } from './runtime-environment-explicit-connect'
 
@@ -60,10 +61,6 @@ function overallDotColor(
     case 'disconnected':
       return 'bg-muted-foreground/40'
   }
-}
-
-function connectedHostCountLabel(count: number): string {
-  return `${count} ${count === 1 ? 'host' : 'hosts'}`
 }
 
 function sshStatusForOverall(status: SshConnectionStatus): HostStatus {
@@ -343,7 +340,9 @@ export function SshStatusSegment({
                 <span className="text-[11px]">
                   <span className={syncProblem ? 'text-destructive' : 'text-muted-foreground'}>
                     {syncProblemLabel ??
-                      (anyConnecting ? 'Connecting…' : connectedHostCountLabel(connectedHostCount))}
+                      (anyConnecting
+                        ? connectingHostsLabel()
+                        : connectedHostCountLabel(connectedHostCount))}
                   </span>
                 </span>
               )}
--- a/src/renderer/src/components/status-bar/resource-manager-terminal-copy.test.ts
+++ b/src/renderer/src/components/status-bar/resource-manager-terminal-copy.test.ts
@@ -19,8 +19,8 @@ describe('resource manager terminal copy', () => {
         spaceScanReady: false
       })
     ).toEqual([
-      'Resource Manager - 512 MB · Σ RSS - 2 terminal sessions',
-      'Terminal sessions are grouped by workspace.'
+      { text: 'Resource Manager - 512 MB · Σ RSS - 2 terminal sessions', emphasized: false },
+      { text: 'Terminal sessions are grouped by workspace.', emphasized: false }
     ])
   })
 
@@ -32,9 +32,23 @@ describe('resource manager terminal copy', () => {
         spaceScanReady: true
       })
     ).toEqual([
-      'Resource Manager - memory unavailable - 0 terminal sessions',
-      'Space scan ready',
-      'No terminal sessions yet.'
+      { text: 'Resource Manager - memory unavailable - 0 terminal sessions', emphasized: false },
+      { text: 'Space scan ready', emphasized: true },
+      { text: 'No terminal sessions yet.', emphasized: false }
+    ])
+  })
+
+  // Why: the segment used to tint the space-scan row by comparing it against the
+  // English text, so a translated build lost the tint. The flag replaces that.
+  it('flags the space-scan row instead of leaving callers to match its text', () => {
+    const lines = getResourceManagerTooltipLines({
+      memoryLabel: '512 MB',
+      sessionCount: 1,
+      spaceScanReady: true
+    })
+
+    expect(lines.filter((line) => line.emphasized).map((line) => line.text)).toEqual([
+      'Space scan ready'
     ])
   })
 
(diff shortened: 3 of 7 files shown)


nwparker:
Thanks @smwbev — I reproduced this independently and your diagnosis is right, including the point that `audit-localization-coverage.mjs` cannot see values returned from helper functions.

I've opened #12478 with a fix that builds on yours. Credit where it's due: **your `{ text, emphasized }` change caught a real bug in my version.** I had localized `'Space scan ready'` but left `ResourceUsageStatusSegment.tsx:1193` tinting the row via `line === 'Space scan ready'`, so the highlight silently vanished in every translated build. I've adopted your structural fix and your unit test for the flag, and commit `6d1fdbc` is co-authored to you.

Visual proof of exactly that, three states from one harness under a Japanese UI (before / localize-only / after) is in #12478.

What #12478 adds on top:

- **es/ja/ko/zh translations.** This PR adds keys to `en.json` only, so at runtime nothing changes — `translate()` already falls back to the same inline English default. Recent i18n fixes here (`0d38053945`, `8fc892dd02`) ship the target catalogs in the same PR.
- The `'Workspace conflict'` / `'Workspace sync error'` ternary two lines from the `Connecting…` you fixed.
- A narrowing of `classifyStringNode` so `??` / `||` / `&&` ancestors keep walking, which surfaced 4 more genuine unlocalized strings. (I left the helper-return half open — I measured 1056 candidates, too big for a bug fix.)
- Whole-line keys for the tooltip summary and aria label instead of `parts.join(', ')`, since ja/zh repunctuate to `、`.

Happy to close mine in favour of yours if you'd rather land the target catalogs and the sibling strings here instead.

smwbev:
Thanks @nwparker — closing this in favour of #12478, and I appreciate you tracing the tint bug back rather than just landing your own version.

Your PR is the better one to ship, for two reasons that are not close calls:

- **It ships the target catalogs.** This PR only adds keys to `en.json`, so es/ja/ko/zh would have stayed English until a follow-up. I am not in a position to write those four translations to a standard worth shipping, so landing mine first would have meant landing the smaller half.
- **`ariaLabel` / `ariaLabelWithSpaceScan` is a real fix to my version.** I built the screen-reader label with `parts.join(', ')`, which bakes a comma into every locale. Whole-line keys are correct and I had missed it.

Two things that may save you a round trip.

### The plural mechanism may draw a review comment

You use `{{count}}` with `_one`/`_other`. That is what I originally proposed in #11826, and @AmethystLiang rewrote it to a `=== 1 ? … : …` ternary before merging — the codebase has since read as if the ternary were the house style. Worth knowing before review, since the two of you have now landed on different answers for the same question.

Separately, and unrelated to which mechanism you pick: neither one produces CLDR plural categories for **plugin** language packs. Those register under a synthetic `plugin<hex>` resource language, which i18next does not map to a locale, so a Russian pack falls back to English plural rules — two forms where Russian needs three. That is what #12106 is about, and your PR is a good concrete example if you want to weigh in there. It does not affect the built-in catalogs in your diff.

### The helper-return sweep you set aside

You mentioned measuring 1056 candidates and leaving that half open — that number is dominated by noise, and a narrower cut is tractable. Scanning `src/renderer/src` on `13f033f09` for functions that **return a bare UI-shaped string literal** (excluding anything already inside `translate()`, plus Tailwind class lists), then keeping only those whose result reaches JSX:

**87 functions, 214 literals.** Heaviest files:

| Literals | File |
|---|---|
| 10 | `components/automations/AutomationsPage.tsx` |
| 10 | `components/automations/automation-page-parts.tsx` |
| 10 | `components/crash-report/CrashReportDialogSurface.tsx` |
| 10 | `components/github/pr-check-presentation.ts` |
| 10 | `components/right-sidebar/checks-panel-content.tsx` |
| 9 | `components/task-page-github-work-item-status.ts` |
| 8 | `components/AgentStateDot.tsx` |
| 8 | `components/settings/repository-source-control-ai-labels.ts` |
| 8 | `components/status-bar/usage-error-copy.ts` |
| 7 | `components/settings/McpConfigFileRow.tsx` |

Spot-checking the top of that list: `getCheckStatusLabel()` returns `Successful` / `Failed` / `Cancelled` / `Timed out` / …, and `agentStateLabel()` returns `Working` / `Blocked` / `Waiting for input` / … — all user-visible, none reachable by the coverage audit today.

Treat 214 as an upper bound rather than a worklist: a tail of it is proper nouns that should stay English (`getProviderDisplayName()` returns `Claude`, `Codex`, `Grok`). Happy to share the scan script if it is useful, or to take a slice of that list as a separate PR once #12478 lands — your call, and no need to route it through me if you would rather keep it in one pair of hands.

smwbev:
Superseded by #12478, which carries this fix plus the es/ja/ko/zh catalogs.

Issue #12063, opened 2026-08-02
[Bug]: macOS Korean Hangul input lags/offset in integrated terminal (Terminal.app OK)


ryan-uhhg:
## Summary

When typing Korean (Hangul) in **Orca’s integrated terminal**, composition feels **lagged / shifted** (characters appear delayed or visually “pushed” relative to keystrokes). The same input is **normal in macOS Terminal.app**.

This appears specific to Orca’s embedded terminal IME/composition handling, not system-wide macOS input.

## Environment

| Item | Value |
|------|--------|
| Orca | **1.4.163** (`com.stablyai.orca`) |
| OS | **macOS 26.3** (Build 25D125) |
| Arch | **arm64** (Apple Silicon) |
| Input source | System **2-Set Korean** (`com.apple.inputmethod.Korean.2SetKorean`) + ABC |
| Locale | `en_US.UTF-8` |
| Control | **Terminal.app: no issue** |

## Steps to reproduce

1. Open Orca → integrated terminal (shell or agent TUI).
2. Switch to Korean 2-Set (두벌식).
3. Type Hangul continuously at a normal pace, e.g. long sentences / multi-syllable words with ongoing composition.
4. Compare with the same typing in **Terminal.app**.

## Expected

- Hangul composition preview and committed syllables stay aligned with the caret.
- No lag or visual offset relative to keypresses (same feel as Terminal.app).

## Actual

- In **Orca terminal only**, Hangul appears **lagged / offset while composing** (“밀리면서 써짐”).
- **Terminal.app** with the same macOS IME and locale is fine.
- Suggests Orca terminal (xterm/PTY + composition events) rather than OS IME or font configuration alone.

## Notes / related issues

Several open Hangul/IME terminal bugs already exist; this may be related or the same root cause. Filing with a clear **Terminal.app control** and version so it can be deduped:

- #11914 – composition only after Space (macOS)
- #11855 – previous IME text disappears during next composition (macOS, P0)
- #8038 – last Hangul syllable drops onto new line on Enter (macOS)
- #11332 – session chat composition issues (macOS)
- #9803 / #11878 / #11919 – Windows/WSL Hangul terminal issues

Happy to retest on a nightly/build with IME fixes, or provide a short screen recording if useful.

## Workaround

Type Korean in Terminal.app / paste into Orca, or use non-terminal UI fields for longer Hangul prompts.

nwparker:
[...]

Issue #11914, opened 2026-08-01
[Bug]: macOS Korean IME composition text only appears after pressing Space


mrjijn01:
### Operating system

macOS

### Orca version

1.4.163

### Details

Title: macOS Korean IME composition text only appears after pressing Space

Environment:
- macOS: [tahoe 26.5.2]
- Orca version: [Orca 1.4.163]
- Korean input source: Korean – 2-Set

Steps to reproduce:
1. Open an Orca terminal.
2. Switch the macOS input source to Korean.
3. Type “한글”.

Expected behavior:
The full composing Korean text should appear as I type.

Actual behavior:
Only the first syllable (“한”) is visible while composing.
The rest appears only after pressing Space to commit the composition.

The same Korean input works normally in macOS Terminal/iTerm2.

tada-js:
Same symptom on a second machine — confirming this report.

**Environment**
- macOS 26.5.2 (25F84), Apple M1 Pro (arm64)
- Input source: Korean – 2-Set

**What I see**
Identical to the original report. When typing 한글 in an Orca terminal, the
composing syllable is not rendered as I type. The previous syllable only shows
up after I press Space or type the next character, so the text always lags one
character behind what I typed. Korean input behaves normally in other terminal
apps on the same machine, so this is specific to Orca.

**Regression range**
I rolled back and tested both versions on the same machine:

- **1.4.162** — Korean composition renders correctly while typing. No lag.
- **1.4.163** — reproduces exactly as described above.

Only the Orca version differed, so this is a regression introduced in 1.4.163
rather than something specific to my environment.

**Possible starting point**
The v1.4.163 release notes include:

> fix(terminal): reconcile cross-platform IME composition lifecycle

That change may be worth looking at first.

PR #12441, opened 2026-08-04
test(terminal): pin Korean intermediate commit

nwparker:
## Summary

- assert `한` reaches xterm `onData` while `글` remains native 2-Set Korean preedit
- retain the same exact `한글` PTY-byte assertion after physical Enter
- reuse the existing ordinary-input negative and stock-xterm mutation

No production code changes. This is issue-specific evidence for #11914, #12063, STA-3197, and STA-3198.

## Verification

- targeted oxlint and React Doctor lint
- oxfmt
- Electron E2E build and Playwright discovery (5 native macOS cases)

## Remaining certification

The physical test must be rerun on the final restacked landing head when it will not interrupt the active Mac session. Reporter confirmation remains separate.

### DIFF
--- a/tests/e2e/terminal-macos-2set-korean-native.spec.ts
+++ b/tests/e2e/terminal-macos-2set-korean-native.spec.ts
@@ -195,8 +195,9 @@ test.describe('Native macOS 2-Set Korean terminal input @headful', () => {
       testInfo,
       testRepoPath,
       electronApp.process().pid!,
-      [5, 40, 1, 15, 46, 3, 36],
-      '한글'
+      [5, 40, 1, 15, 46, 3],
+      '한글',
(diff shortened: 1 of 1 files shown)


nwparker:
Closing my own draft — the property is already pinned on main.

`terminal-macos-2set-korean-native.spec.ts` has the `preCommit` parameter and a test using it: 'flushes each syllable while the next remains in preedit', keyCodes `[15,40,1,40,14,40]`, `{ committedText: '가나', preeditText: '다' }` — landed as `e08eba674c1` (#12284). This PR would have converted the 한글 case to the same shape, which adds keystrokes but no coverage.

Also sitting on a base 805 commits behind main, so a rebase would cost more than it returns.

